Issue #3552, opened 2026-07-14
guard-destructive.sh: SQL DDL/DML blocks are a category error for DB-engine projects (add per-project opt-out)


rjwalters:
## Summary

`guard-destructive.sh` hard-blocks SQL DDL/DML patterns (`DROP DATABASE`, `DROP TABLE`, `DROP SCHEMA`, `TRUNCATE TABLE`, and `DELETE FROM` without `WHERE`). This is a **category error for any project that is itself a database engine** — the guard assumes SQL statements target a precious *production* DB, but in a DB-engine repo those statements are the product's own test/dev vocabulary.

Observed in the `vibesql` repo (a pure-Rust SQLite-compatible engine that runs SQLite's TCL conformance suite). There, `DROP TABLE` / `DELETE FROM` / `TRUNCATE TABLE` appear constantly in:

- inline `vibesql -c "DROP TABLE t"` dev commands,
- the TCL test suite,
- docs, commit messages, and even **comments/labels** in shell one-liners.

Because the match is `grep -qiE` (case-insensitive substring, anywhere in the command), the guard fires even when the words appear in a comment or a `--description` label — you cannot write a shell line that merely *mentions* `drop database` without being denied.

## Secondary findings (same file)

1. **Broad AWS `ASK` entries are redundant + high-friction.** `aws s3` / `aws ec2` / `aws lambda` in `ASK_PATTERNS` prompt on read-only calls (`aws ec2 describe-instances`, `aws s3 ls`) and on `start/stop-instances`. The genuinely destructive AWS ops (`aws ec2 terminate`, `aws s3 rb`, `aws s3 rm --recursive`, `aws iam delete`, `aws cloudformation delete-stack`) are **already hard-blocked** in `ALWAYS_BLOCK_PATTERNS`, so the broad `ASK` entries add friction without adding protection.

2. **`git checkout \.` / `git restore \.` over-match.** The pattern matches the substring `git checkout .` inside `git checkout ./src/main.rs`, so a **single-file** checkout is treated as a discard-everything and prompts. Anchor to `git checkout \.($|[[:space:]])` so only whole-tree discard prompts.

## Proposed durable fix (per-project opt-out)

Guard the SQL block behind a project-level flag rather than removing it globally (most projects genuinely benefit from it). Options, in order of preference:

[...]

Issue #3550, opened 2026-07-13
guard-destructive.sh: hookSpecificOutput missing required hookEventName — deny/ask decisions silently discarded


rjwalters:
Claude Code's hook output schema now requires hookSpecificOutput.hookEventName. defaults/hooks/guard-destructive.sh omits it in all four emit paths (jq + raw fallback, deny() and ask()), so the harness rejects the JSON and the guard becomes a silent no-op — deny/ask decisions are discarded and the guarded command executes anyway. Fix verified in rjwalters/vibesql @ fab7afcdd: add hookEventName: 'PreToolUse' to all four emit sites. Also worth auditing skill-router.sh / methodology-inject.sh for the same schema drift.

rjwalters:
## Curator enrichment

### Root cause

`defaults/hooks/guard-destructive.sh` is a `PreToolUse` hook that emits permission decisions as `hookSpecificOutput`. Claude Code's current `PreToolUse` hook schema **requires** a `hookEventName: "PreToolUse"` field *inside* the `hookSpecificOutput` object. The guard omits it in every emit path, so Claude Code treats the payload as invalid and silently discards the decision. The hook still exits `0`, so nothing looks broken — but every `deny`/`ask` is a no-op and the guarded command runs anyway. This is a **silent safety regression**: the destructive-command guard is currently inert.

### Correct schema

```json
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "deny",   // or "ask" / "allow"
    "permissionDecisionReason": "..."
  }
}
```

Confirmed correct against the two sibling `UserPromptSubmit` hooks in the same directory (`skill-router.sh` line 154, `methodology-inject.sh` line 223), which already include `hookEventName: "UserPromptSubmit"` — that is the schema shape Claude Code expects and `guard-destructive.sh` is the odd one out.

### Exact emit sites missing the field

`defaults/hooks/guard-destructive.sh` (source of truth; `.loom/hooks/` is a gitignored installed copy). Four sites across `deny()` and `ask()`, each with a `jq` path and a raw-JSON fallback path:

| Function | Path | Line | Fix |
|----------|------|------|-----|
| `deny()` | jq object | ~77-80 | add `hookEventName: "PreToolUse",` before `permissionDecision:` |
| `deny()` | raw fallback `echo` | 87 | add `\"hookEventName\":\"PreToolUse\",` before `\"permissionDecision\"` |
[...]

PR #3564, opened 2026-07-15
fix(guard): reduce false-positive denials from unanchored substring matching (#3553)

### LINKED ISSUE #3553: guard-destructive.sh: substring matching causes false-positive denials (comments, flag names, commit messages, remote ssh payloads)
## Summary

`guard-destructive.sh` matches its block/ask patterns with `grep -qiE` — **case-insensitive substring, anywhere in the raw command text**, with no lexical awareness. This produces a steady stream of false-positive denials on commands that merely *mention* a pattern word in a comment, a flag name, a commit message, a quoted string, or a remote `ssh` payload. Over a single work session it blocked ~8 distinct legitimate commands (evidence below). This is the engine-level counterpart to the pattern-set relaxations already discussed in #3552 — that issue narrows *which* patterns exist; this issue fixes *how* they're matched.

## Concrete false positives observed (all legitimate commands, all denied)

1. **Flag names that contain a pattern substring.** `aws ec2 run-instances --instance-initiated-shutdown-behavior stop …` → denied by the `shutdown` pattern. The word is part of a flag name, not a `shutdown(8)` invocation.
2. **Shell comments.** A command carrying `# ... reboots the box ...` or `# drop database first` in a comment → denied, because the comment text is matched the same as executable code.
3. **Commit messages.** `git commit -m "...relax guard: rm -rf /, gh repo delete still blocked..."` → denied. Documenting the patterns in a message trips them. (Worked around by writing the message to a file and using `-F`.)
4. **Test/verification harnesses.** A script that feeds candidate commands *to the hook* to test it (`printf '%s' ... | ./guard-destructive.sh`) is itself denied when the candidate strings (`gh repo delete`, `rm -rf /`, fork-bomb) appear as data in the harness. Testing the guard requires assembling every dangerous literal from fragments to avoid self-blocking.
5. **`rm -rf /` substring in a safe path.** `rm -rf /tmp/diag.vbsql` (and any `rm -rf /tmp/...`, `rm -rf /var/...`) → denied by the literal `rm -rf /` pattern, which matches the leading `/tmp`.
6. **Remote `ssh` payloads treated as local.** `ssh host 'rm -rf "$RD"/checkpoints; rm -f /home/ubuntu/foo.*'` → denied by the local rm-scope check, even though the `rm` runs on the *remote* host and `/home/ubuntu/...` is a remote path. The guard has no notion that the text inside `ssh <host> '…'` executes elsewhere.
7. **Crude rm-target extraction.** The rm-scope check does `echo "$COMMAND" | sed 's/rm\s\+//' | tr ' ' '\n' | grep -v '^-'` — it strips only the *first* `rm ` and then treats **every remaining whitespace-delimited token in the whole command** as an rm target. So an unrelated token from an earlier part of the command (e.g. a filename in a preceding `cat host-ip.txt`) gets resolved as an rm target and flagged "outside repository." Real example this session: a command whose first statement was `HOST=$(cat .../scratchpad/host-ip.txt)` and which later had remote `rm` calls was denied citing `host-ip.txt` as an rm target.

## Root cause

`grep -qiE "$pattern"` over `$COMMAND` has no lexical model of the shell:
- no comment stripping,
- no word boundaries (so `shutdown` ⊂ `--…-shutdown-behavior`, `rm -rf /` ⊂ `rm -rf /tmp/x`),
- no quoting / heredoc / `ssh`-payload awareness,
- no distinction between the command being run and strings that are its *data*,
- an rm-target extractor that mis-parses any command containing more than one word after the first `rm`.

## Proposed fixes (upstream, in `guard-destructive.sh`)

Ordered by value; each is independent:

1. **Strip shell comments before matching.** Remove `#…EOL` segments (respecting quotes as best as bash-in-a-hook can) so comments can't trip patterns.
2. **Word-boundary anchor the command patterns.** e.g. `\bshutdown\b`, `\bgh repo delete\b`. Kills the flag-name class (#1) and most comment/message classes.
3. **Anchor `rm -rf /` to a real root target.** Match `rm -rf /($|[[:space:]])` (and `/*`, `~`, `$HOME`) rather than the bare prefix, so `rm -rf /tmp/x` no longer matches. (Same anchoring style PR #6130 already applied to `git checkout \.`.)
4. **Fix rm-target extraction.** Parse actual `rm` invocations (tokens that follow an `rm` command word, up to the next `;`/`&&`/`|`), not "every word after the first `rm`." This removes the `host-ip.txt`-as-target class (#7).
5. **Do not apply the local rm-scope / path checks to remote payloads.** When the command is `ssh …`/`scp …`/`bash -c '…'` targeting another host, the local repo-root reasoning is meaningless. At minimum, detect an `ssh <host> '<payload>'` wrapper and skip the local-filesystem scope checks for the payload (the ALWAYS_BLOCK catastrophic patterns can still apply).
6. **(Optional, larger) Match against tokenized command words, not the raw string.** A light tokenizer that ignores quoted-string *contents* for the ASK/block-word patterns would eliminate the "pattern-as-data" classes (#3, #4) wholesale.

## Acceptance criteria

- `aws ec2 run-instances --instance-initiated-shutdown-behavior stop …` → allowed.
- `git commit -m "…rm -rf / … gh repo delete …"` (pattern words only in the message) → allowed.
- `rm -rf /tmp/whatever` → allowed; `rm -rf /` and `rm -rf /tmp` (the directory itself) → still denied.
- `ssh host 'rm -f /home/ubuntu/foo'` → not denied by the *local* rm-scope check (remote path).
- A command that contains `cat something.txt` earlier and a scoped `rm` later is not denied citing the `cat` argument as an rm target.
- All genuinely catastrophic commands remain denied: `rm -rf /`, force-push to main/master, `gh repo delete`, `aws ec2 terminate`, `aws s3 rb`, `curl … | sh`, fork bomb, reboot/shutdown as actual commands.

## Context

- Prior pattern-set relaxations (SQL DDL for DB-engine repos, redundant AWS ASKs, git checkout anchoring) landed downstream in rjwalters/vibesql PR #6130 and are proposed upstream in #3552. This issue is the complementary **matching-precision** fix and is largely orthogonal.

rjwalters:
Closes #3553

## Problem

`guard-destructive.sh` matched every block/ask pattern with `grep -qiE` — a case-insensitive substring over the raw command text with no lexical model of the shell. Destructive-looking tokens in **flag names** (`shutdown` ⊂ `--instance-initiated-shutdown-behavior`), **comments**, **commit messages**, and **remote ssh payloads** falsely blocked legitimate commands. The rm-target extractor (`sed 's/rm\s\+//' | tr ' ' '\n'`) mis-parsed any multi-token command, treating unrelated tokens (and remote paths) as local rm targets.

## Governing constraint (kept intact)

The catastrophic `ALWAYS_BLOCK` set **still scans the full raw command string, including quoted/heredoc content**, so `bash -c 'rm -rf /'` and quoted/remote catastrophic payloads still deny. Precision is applied surgically per false-positive class — never by globally excluding quoted strings from the catastrophic scan.

## Changes (`defaults/hooks/guard-destructive.sh`)

1. **Anchor the `rm -rf /` family to a real root/home target.** `rm … /([^path-char]|$)` — `rm -rf /tmp/x` no longer trips the catastrophic rule; `rm -rf /`, `/*`, `~`, `$HOME` (including inside quotes) still deny. The left side of `rm` is deliberately unanchored so quoted payloads (`bash -c 'rm -rf /'`) keep matching.
2. **Command-position + trailing-boundary anchor** the bare-word block patterns (`reboot`/`shutdown`/`halt`/`poweroff`/`init 0`/`init 6`) and `gh repo delete`/`gh repo archive`. A flag containing the word, or `"reboots"` in a comment, no longer trips; real leading/sudo/after-separator invocations still deny.
3. **Rewrite the rm-target extractor.** Split on `;`/`|`/`&`/`&&`/`||`; only inspect segments whose command word is actually `rm` (optionally `sudo`-prefixed) carrying a recursive/force flag. Kills the phantom-target class (`cat host-ip.txt; … rm …`) and naturally skips `ssh`/`scp` remote payloads (their command word is `ssh`, not `rm`).
4. **Relax the rm-scope check** to block only root, `$HOME`, and top-level directories (`/tmp`, `/var`, `/etc`, …). A scoped subpath such as `/tmp/whatever` or `/var/foo` is allowed. (This is the deliberate behavior change behind the updated "outside repo" test.)
5. **Strip `#…EOL` comments** on a working copy used **only** for the ASK-word and SQL DDL/DML matches — never for the catastrophic scan. Gated on the presence of a `#` to stay off the hot path.

## Deliberate tradeoff

Because the catastrophic scan must keep reading quoted text (constraint above), a commit message that **literally** contains `rm -rf /` or `gh repo delete` still denies — this is the safe side, and the issue already documents the `-F messagefile` workaround. The fixable commit-message class (mentioning `rm -rf` without a root target, `--force` flags, etc.) is allowed. Tests reflect exactly what is and isn't achievable.

## Tests (`tests/hooks/test-guard-destructive.sh`)

- **False positives now ALLOWED:** flag-name substrings, `reboots` in a comment (word-boundary), DDL/ASK words in a comment (comment-stripped), commit message mentioning `rm -rf`, `--force` tool flags, ssh remote payloads, `rm -rf /tmp/x` and `/var/foo` subpaths, and the `cat …; rm …` phantom-target class.
- **Regression guard — all still DENY:** `rm -rf /`, `rm -rf /*`, `bash -c 'rm -rf /'`, double-quoted `rm -rf /`, `rm -rf ~`/`$HOME`, bare top-level dir, force-push to main/master (all flag forms), `gh repo delete`/`archive` (leading, after `&&`, sudo), `aws ec2 terminate`/`s3 rb`/`s3 rm --recursive`, `curl|sh`, `wget|bash`, fork bomb, `reboot`/`shutdown`/`halt`/`poweroff`/`init 0`/`init 6` as real commands, and SQL `DROP TABLE`/`DELETE FROM` (guard on).

Suite: **161 tests, all functional tests pass.** (The one flaky item is the pre-existing ~200ms performance threshold, unrelated to this change — the new code paths are gated behind a `#`-present / `rm`-with-flag pre-check.)

Invariants preserved: ERR trap still falls through to allow; every deny/ask uses the existing `deny()`/`ask()` helpers, so the required `hookEventName: "PreToolUse"` field (#3550) is emitted. Portable regex only (`[[:space:]]`, ERE bracket anchors; no `grep -P`/`\b`). SQL DDL opt-out (#3552) behavior unchanged.

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -160,9 +160,14 @@ ask() {
 # =============================================================================
 
 ALWAYS_BLOCK_PATTERNS=(
-    # GitHub destructive operations
-    'gh repo delete'
-    'gh repo archive'
+    # GitHub destructive operations — command-position anchored (start-of-line
+    # or a shell separator must precede the verb) so the phrase inside a flag
+    # value no longer trips. NOTE: the catastrophic scan still runs over the
+    # full raw command, including quoted/heredoc text, so a `gh repo delete`
+    # that a shell would actually execute (leading, sudo-prefixed, or after a
+    # separator) still denies (#3553).
+    '(^|[;&|[:space:]])gh repo delete'
+    '(^|[;&|[:space:]])gh repo archive'
 
     # Force push to main/master (various flag forms)
     'git push --force origin main'
@@ -172,11 +177,16 @@ ALWAYS_BLOCK_PATTERNS=(
     'git push --force-with-lease origin main'
     'git push --force-with-lease origin master'
 
-    # Filesystem destruction
-    'rm -rf /'
-    'rm -rf /\*'
-    'rm -rf ~'
-    'rm -rf \$HOME'
+    # Filesystem destruction — anchored to a *real* root/home target so that a
+    # scoped path like `rm -rf /tmp/x` no longer trips the catastrophic rule,
+    # while root / home obliteration still denies. The left side of `rm` is
+    # deliberately NOT anchored, so a quoted payload such as `bash -c 'rm -rf /'`
+    # (root followed by a closing quote) still matches (#3553). The trailing
+    # class matches anything that is not a path-continuation character (so `/`,
+    # `/ `, `/*`, `/;`, `/'` all count as "root itself" but `/tmp` does not).
+    'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)'
+    'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+~([^[:alnum:]._~/-]|$)'
+    'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+\$HOME([^[:alnum:]._~/-]|$)'
 
     # Fork bombs
     ':\(\)\{ :\|:& \};:'
@@ -200,13 +210,16 @@ ALWAYS_BLOCK_PATTERNS=(
     # Docker mass destruction
     'docker system prune'
 
-    # System reboot/shutdown
-    'reboot'
-    'shutdown'
-    'halt'
-    'poweroff'
-    'init 0'
-    'init 6'
+    # System reboot/shutdown — command-position + trailing-boundary anchored so
+    # a flag name (e.g. --instance-initiated-shutdown-behavior) or the word in a
+    # comment/message ("this reboots the box") no longer trips them, while the
+    # bare command (optionally sudo-prefixed, or after a separator) still denies.
+    '(^|[;&|[:space:]])reboot([;&|[:space:]]|$)'
+    '(^|[;&|[:space:]])shutdown([;&|[:space:]]|$)'
+    '(^|[;&|[:space:]])halt([;&|[:space:]]|$)'
+    '(^|[;&|[:space:]])poweroff([;&|[:space:]]|$)'
+    '(^|[;&|[:space:]])init[[:space:]]+0([;&|[:space:]]|$)'
+    '(^|[;&|[:space:]])init[[:space:]]+6([;&|[:space:]]|$)'
 )
 
 for pattern in "${ALWAYS_BLOCK_PATTERNS[@]}"; do
@@ -214,6 +227,26 @@ for pattern in "${ALWAYS_BLOCK_PATTERNS[@]}"; do
         deny "BLOCKED: Command matches dangerous pattern: $pattern"
     fi
 done
+
+# =============================================================================
+# COMMENT-STRIPPED WORKING COPY - used ONLY for the ASK-word and SQL DDL/DML
+# matches below, never for the catastrophic ALWAYS_BLOCK scan.
+#
+# Strips a `#…EOL` shell comment when the `#` is at start-of-line or preceded
+# by whitespace (the common comment shape), so a pattern word that appears only
+# in a trailing comment ("# drop database first", "# git push --force") no
+# longer trips the ASK/DDL gates. This is best-effort: a `#` inside a quoted
+# string that happens to be whitespace-preceded is also stripped, but since the
+# stripped copy is used only for the *narrowing* ASK/DDL matches (never the
+# catastrophic scan) the worst case is a missed ask on quoted data, never a
+# missed catastrophic block. The sed only runs when a `#` is actually present,
+# keeping it off the hot path (#3553).
+# =============================================================================
+if [[ "$COMMAND" == *"#"* ]]; then
+    COMMAND_NO_COMMENT=$(printf '%s\n' "$COMMAND" | sed -E 's/(^|[[:space:]])#.*$//')
+else
+    COMMAND_NO_COMMENT="$COMMAND"
+fi
 
 # =============================================================================
 # DATABASE DESTRUCTION - Gated by the SQL DDL/DML guard toggle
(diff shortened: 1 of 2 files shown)


rjwalters:
Changes needed: the new `rm` scope check introduces a **catastrophic bypass** — several `rm -rf` commands that resolve to `/` or a protected top-level directory are now ALLOWED, including full root deletion.

## The hole

The rewritten scope check resolves an **absolute** target verbatim (`ABS_PATH="$target"`, no normalization) and then only blocks single-segment paths via `[[ "$ABS_PATH" =~ ^/[^/]+$ ]]`. Any `..` / `//` / `.` traversal defeats that regex, and the anchored `ALWAYS_BLOCK` `rm ... /(non-path|$)` pattern deliberately lets `/tmp`, `/var`, … through as subpath prefixes — so nothing catches these. The old code caught them via the now-removed "outside repo root" check.

Reproduced against this branch (cwd = repo root, i.e. the real agent scenario). Each was DENY on `origin/main` and is now ALLOW:

| Command | Resolves to | old | new |
|---|---|---|---|
| `rm -rf /tmp/..` | `/` (ROOT) | deny | **allow** |
| `rm -rf /var/../` | `/` (ROOT) | deny | **allow** |
| `rm -rf /tmp/../etc` | `/etc` | deny | **allow** |
| `rm -rf /usr/./` | `/usr` | deny | **allow** |
| `rm -rf /home/../home` | `/home` | deny | **allow** |
| `rm -rf /a/../../../etc` | `/etc` | deny | **allow** |
| `rm -rf //etc` | `/etc` | deny | **allow** |

`rm -rf /tmp/..` and `rm -rf /var/../` obliterate the entire root filesystem and are now permitted. This directly violates the governing constraint (regression-guard lists `rm -rf /` and bare top-level dirs as must-deny; the PR's own test asserts `rm -rf /usr` denies — `rm -rf /usr/./` is the same target and slips through).

## Required fix

Normalize absolute targets before the protected-path check, mirroring the relative branch, so `/tmp/..` → `/`, `/var/../` → `/`, `/tmp/../etc` → `/etc` are caught. Note: `realpath -m` is **GNU-only and fails on macOS** (`realpath: illegal option -- m`) — the existing relative branch already silently falls back to unnormalized `$CWD/$target` on macOS, so please use a portable normalization (collapse `//`, resolve `.`/`..` lexically in shell/awk) rather than relying on `realpath -m`.

## Also add regression tests

The new regression-guard block is good but has no `..`/`//`/`.` traversal cases. Please add (must DENY): `rm -rf /tmp/..`, `rm -rf /var/../`, `rm -rf /tmp/../etc`, `rm -rf /usr/./`, `rm -rf //etc`.

## Everything else looks correct

Verified and passing: comment-stripping is scoped to the ASK/DDL/DELETE gates only (never the catastrophic scan); the catastrophic `ALWAYS_BLOCK` loop still scans raw `$COMMAND` incl. quoted/heredoc text (`bash -c 'rm -rf /'` still denies); `hookEventName` (#3550) and the SQL DDL gate (#3552) are preserved; the phantom-target, flag-name, comment, commit-message, and ssh-remote-payload false positives are correctly fixed; scope is limited to `defaults/` + `tests/`; all 161 tests pass. The single blocker is the traversal bypass above.

(Note for follow-up, not blocking this PR since it is pre-existing on main: `rm --recursive --force /` (long-form flags) and `rm -rf "/"` (quoted slash) are missed by both old and new — worth a separate hardening issue.)

rjwalters:
Doctor fix pushed (commit fc79f1b6). Addressed the catastrophic traversal bypass Judge flagged: added a pure-bash lexical path normalizer (`normalize_abs_path`) that collapses `//`, drops `.`, and resolves `..` (clamped at root) **before** the protected-path check — `realpath -m` was deliberately avoided since it's GNU-only and no-ops on macOS. All flagged bypasses now DENY (`rm -rf /tmp/..`, `/var/../`, `/tmp/../etc`, `/usr/./`, `/home/../home`, `/a/../../../etc`, `//etc`) and scoped subpaths still ALLOW (`/tmp/x`, `/tmp/a/../b`). Suite: 169/170 functional pass (the 1 fail is the pre-existing environmental ~200ms perf check, performance-neutral here). Ready for re-review.

rjwalters:
LGTM! Approved.

## Second Judge pass — catastrophic traversal bypass is CLOSED

The Doctor fix (commit `fc79f1b6`) adds a pure-bash lexical `normalize_abs_path` that is applied to the absolute target **before** the protected-path check. Verified the prior blocker is fully resolved.

### Bypass closure — traced by hand AND exercised live against the branch
Every case the first review flagged now DENIES (normalizer collapses `//`, drops `.`, resolves `..` clamped at root):

| Command | Resolves to | Result |
|---|---|---|
| `rm -rf /tmp/..` | `/` | **deny** |
| `rm -rf /var/../` | `/` | **deny** |
| `rm -rf /tmp/../etc` | `/etc` | **deny** |
| `rm -rf /usr/./` | `/usr` | **deny** |
| `rm -rf /home/../home` | `/home` | **deny** |
| `rm -rf /a/../../../etc` | `/etc` | **deny** |
| `rm -rf //etc` | `/etc` | **deny** |

Scoped subpaths still correctly ALLOW: `/tmp/x`, `/tmp/a/../b` → `/tmp/b`, `/var/folders/xy/cache`.

### No new bypass introduced
Probed additional edge cases through the hook:
- **Relative traversal escapes now caught too** (bonus hardening the old `realpath -m` branch missed on macOS): `rm -rf ../../../..` from repo root → `/` deny; `rm -rf .` from `/tmp` → deny.
- **Absolute HOME deletion** with trailing-slash / `.` variants (`$HOME/`, `$HOME/./`) all deny.
- **Literal-segment correctness**: only `.`/`..` are special — `/tmp/....//` stays a scoped subpath (allow), `/...` is a single top-level segment (deny). The normalizer only *reduces* paths, so it cannot turn a dangerous target into a safe-looking one.

### Everything else re-verified intact
- Catastrophic `ALWAYS_BLOCK` loop (L225-226) still scans **raw `$COMMAND`** incl. quoted/heredoc text — `bash -c 'rm -rf /'` still denies.
- `COMMAND_NO_COMMENT` comment-stripping is confined to the narrowing SQL/DELETE/ASK gates only, never the catastrophic scan.
- `hookEventName: PreToolUse` (#3550) present on both deny and ask paths.
- SQL DDL/DML gate (#3552) preserved and still `sql_guard_enabled`-gated.
- Scope limited to `defaults/` + `tests/`; no new labels.

### Tests
170/170 pass locally, including the 5 new traversal regression cases and the 2 scoped-subpath allow cases. Perf check passed this run (173ms < 200ms). CI mergeStateStatus is CLEAN; all applicable checks green (Rust/MCP jobs correctly skipped — no code in those paths).

The single blocker from the first pass is resolved with no regressions. Approved.

PR #3539, opened 2026-07-10
feat(daemon): honor LOOM_WORKTREE_ROOT in loom-daemon terminal-destroy GC

rjwalters:
## Summary

Follow-up to #3530 (configurable worktree root), which shipped the bash + `.loom/config.json` surface only. This PR ports the resolver to the **Rust daemon** so `loom-daemon`'s `destroy_terminal()` GC recognizes worktrees on an overridden base.

Previously the cleanup gate at `terminal.rs:1332` was a hardcoded substring check (`path.to_string_lossy().contains(".loom/worktrees")`). When `LOOM_WORKTREE_ROOT` (or `.loom/config.json` → `worktree.root`) redirects worktrees to an external volume (e.g. `/Volumes/Stripe/<repo>/issue-42`), the daemon silently skipped `git worktree remove`, leaking worktrees on the override volume.

## Changes

- **New module `loom-daemon/src/worktree_root.rs`** (registered in `lib.rs`):
  - `worktree_root(repo_root: &Path) -> PathBuf` mirrors `defaults/scripts/lib/worktree-root.sh`'s `loom_worktree_root` exactly: precedence env `LOOM_WORKTREE_ROOT` > `.loom/config.json` `worktree.root` > default `${repo_root}/.loom/worktrees`; on override, namespaced by repo basename (`${override}/<basename>`, trailing slash stripped); relative overrides fall back to default with a `log::warn!` (not a hard error). Config read is a soft-fail `serde_json::Value` parse following the `extract_configured_terminal_ids` pattern.
  - `is_worktree_path(path, repo_root) -> bool` implements the two-way GC match: `path.starts_with(worktree_root(repo_root))` OR the legacy `.loom/worktrees` substring (mixed-setup fallback, mirroring `agent-destroy.sh`).
- **`terminal.rs` `destroy_terminal()`**: `repo_root` is now derived *before* the gate via a new shared `find_repo_root_from_worktree()` helper (distinct from the existing `find_repo_root` that walks a tmux `working_dir`/`LOOM_WORKSPACE`), and reused for the later `git worktree remove` — no duplicated ancestor walk. The gate now calls `is_worktree_path()`.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `worktree_root()` implements 3-tier precedence (env > config > default) | ✅ | `worktree_root.rs`; tests `env_beats_config`, `config_override_namespaces_by_basename`, `default_when_no_override` |
[...]

Issue #3289, opened 2026-05-15
Parallel shepherds stall before any external action (subagents-launching-subagents)


rjwalters:
## Summary

When launching multiple `/shepherd` agents in parallel from a single parent conversation, **at least one of the parallel shepherds can stall before doing any external work** — no labels advance, no worktree gets created, no comments are posted. The watchdog kills the agent at 600s with `stream watchdog did not recover`. The other parallel shepherd(s) succeed normally.

This is the classic "subagents launching subagents" failure mode: the parent shepherd appears to have spawned but the child never produced any output and the framework never recovered.

## Reproduction

Run two `/shepherd` invocations in parallel against independent issues with **non-overlapping remote resources** (one on a Cadence VM via gcloud IAP, one on a different host via Tailscale SSH — no resource contention by design). In a Sphere monorepo session today:

- **Shepherd A** (issue #7733, cadence-west-3 work): **SUCCEEDED** — labels advanced curated → building → review-requested → pr; PR #7751 opened, judge approved, ready to merge in ~16 min wall-clock.
- **Shepherd B** (issue #7734, alc-0 grep work): **STALLED** — labels never advanced past curated+issue; no worktree created; no comments posted; killed at 600s by stream watchdog.

Reproducibly: an earlier solo `/shepherd` invocation today (same session, same monorepo) on issue #7732 succeeded cleanly. **The stall only manifests when shepherds are spawned in parallel.**

## Symptoms

From the parent agent's monitor on GitHub state:

```
[BASELINE] #7733 labels=ams,gepetto,infra,loom:curated,loom:issue,loom:urgent,pdk comments=0
[BASELINE] #7734 labels=ams,gepetto,infra,loom:curated,loom:issue,loom:urgent,pdk comments=0
[+126s]   LABELS #7733: ...,loom:building,...    ← A advances normally
[+126s]   BRANCH #7733: +feature/issue-7733
[+755s]   BRANCH #7733: +feature/issue-7733,remotes/origin/feature/issue-7733  ← A pushes
[+755s]   PR    #7733: #7751 [loom:review-requested]                             ← A opens PR
... no #7734 events at all ...
```

#7734 stayed at baseline state for the entire 600s window, then the agent task transitioned to status=failed with summary: `Agent stalled: no progress for 600s (stream watchdog did not recover)`.

After the stall:
- `gh issue view 7734` confirms labels unchanged from baseline, 0 comments
[...]

PR #3567, opened 2026-07-15
feat(sweep): resource-gated automatic wave-size default (#3566)

### LINKED ISSUE #3566: sweep: resource-gated automatic wave-size default (scale toward 10, warn/reduce only when constrained)
## Problem

\`/loom:sweep\` defaults to \`--builders-per-wave 1\` and, in practice, operators pick 2–3 because the skill soft-caps at 3 and warns above. For large batches this leaves a lot of throughput on the table. The default should **auto-scale from available resources** instead of a conservative hardcoded number, and only reduce/warn when genuinely constrained.

## Infra update (2026-07-15): dedicated worktree scratch volume

A dedicated second drive is now provisioned for worktree scratch, wired via the \`LOOM_WORKTREE_ROOT\` env var / \`.loom/config.json → worktree.root\` override (#3539, #3541; resolver: \`defaults/scripts/lib/worktree-root.sh\` / \`loom-daemon/src/worktree_root.rs\`). **Disk is therefore no longer the primary limiter** for worktree-heavy parallel sweeps.

Consequence for this design: the disk-headroom gate must measure **the worktree-root filesystem** (resolve via \`loom_worktree_root <repo_root>\`, then \`df\` that path) — NOT the repo's own drive. On a machine with a large dedicated scratch volume, the disk gate will rarely bind.

## Two parallelism mechanisms (must not be conflated)

1. **Detached sweep processes** — the daemon path (\`mcp__loom__dispatch_sweep\`). Each sweep runs in its own OS process with its own rotated account token. This safely scales toward 10: **not** nested subagents, so the #3289 stall does not apply; isolated context; load spread across accounts. Requires \`loom-daemon\` reachable **and** a multi-account token pool (\`.loom/tokens/\` ≥ 2). **This is the intended vehicle for high parallelism.**
2. **In-session subagent builders** — the \`--builders-per-wave\` fallthrough. Builders are Task subagents dispatched one level deep. Bounded by the #3289 nested-dispatch stall (a stream-pump race — orthogonal to disk; the scratch drive does NOT retire it), shared parent context, and a single OAuth token's rate limits. Validated ceiling ~3.

## Desired behavior — resource-gated auto default

Replace the hardcoded default with an auto-selector in Stage -1 / wave-planning:

- **Daemon + multi-account pool present** → route to the daemon detached-process path and target up to **N=10** concurrent sweeps, clamped by the worktree-root disk-headroom check (per-worktree footprint vs free space on the scratch volume). With a dedicated scratch drive this typically stays at 10.
- **No daemon or single-token** → subagent fallthrough. Keep the #3289-safe ceiling (currently 3); auto-clamp down on low disk. Do **not** raise the nested-subagent default toward 10 — route to the daemon path instead if high parallelism is wanted.
- **Only warn/reduce when constrained**, with a one-line reason: \"reduced to K: single-token pool → subagent path (cap 3)\", \"reduced to K: only N GB free on <worktree-root>\", \"daemon not reachable → subagent path (cap 3)\".
- Explicit \`--builders-per-wave N\` still overrides (operator wins).

## Acceptance criteria

- [ ] Sweep auto-computes wave size when \`--builders-per-wave\` is not passed, instead of defaulting to 1.
- [ ] Auto value reaches up to 10 **only** on the daemon+pool detached-process path; subagent path stays at the #3289-safe ceiling.
- [ ] Disk-headroom check resolves the **worktree-root filesystem** (via \`loom_worktree_root\`) and clamps + logs the reason when it reduces the target.
- [ ] Single-token / no-daemon environments degrade gracefully to the safe subagent cap with a clear one-line explanation.
- [ ] Explicit \`--builders-per-wave N\` continues to override.
- [ ] \`--dry-run\` prints the auto-selected wave size, the chosen mechanism (daemon vs subagent), and the gating reason.
- [ ] Docs (sweep skill + CLAUDE.md wave-size guidance table) updated for the resource-gated default, the two-mechanism distinction, and the worktree-root disk-gate.
- [ ] #3289 one-level-deep constraint remains respected and documented (routed around via the process path, not violated).

## Notes / references

- #3289 — nested-dispatch stall hazard (why the subagent path is capped; orthogonal to disk).
- #3454 — Stage -1 backend detection (daemon + pool probe already exists; reuse as the branch point).
- #3539 / #3541 — \`LOOM_WORKTREE_ROOT\` / \`worktree.root\` override + resolver; the disk gate must target this path.
- Real-world signal: a 10-issue single-token sweep hit a **weekly rate limit** mid-run — single-token concurrency is bounded by rate limits, which the multi-account pool + daemon path address.
- Canonical source is \`defaults/.claude/commands/loom/sweep.md\`.

rjwalters:
Closes #3566

## Summary

Replaces `/loom:sweep`'s hardcoded `--builders-per-wave` default of `1` with an `auto` sentinel resolved at Stage -1 from the chosen backend and scratch-volume disk headroom. Only warns/reduces when genuinely constrained; explicit `--builders-per-wave N` still overrides.

## Two mechanisms (not conflated)

- **daemon + multi-account pool** → detached-process path (`mcp__loom__dispatch_sweep`). Each sweep is its own OS process with its own rotated token — **not** nested subagents, so the #3289 stall does not apply. Targets up to **N=10**.
- **no daemon / single-token / `--no-daemon` / Mode C** → in-session subagent path, one level deep. Bounded by #3289; stays at the validated ceiling of **3**.

Both clamped by candidate count and by free space on the **worktree-root filesystem** (resolved via `loom_worktree_root`, so it measures the dedicated scratch volume from #3539/#3541 — not the repo drive), floor of 1. **The subagent cap is never raised toward 10** — high parallelism is routed through the daemon path.

## New helper (the unit-testable seam)

`defaults/scripts/lib/disk-headroom.sh` (sourceable, not exec):
- `loom_worktree_root_free_gb <repo_root>` — resolves the worktree root, walks up to the nearest existing ancestor, `df -Pk`s it (portable across macOS/GNU), echoes integer free GB.
- `loom_wave_size_from_disk <mechanism> <candidate_count> <free_gb>` — pure integer arithmetic. Returns `min(target, floor(free_gb/PER_WORKTREE_GB), candidate_count)` (floor 1) on line 1 and a reason token (`target`/`candidates`/`disk`/`floor`) on line 2. Targets: daemon 10, subagent 3.

Constants: `LOOM_PER_WORKTREE_GB` (default **2** GB), `LOOM_DAEMON_WAVE_TARGET` (10), `LOOM_SUBAGENT_WAVE_CAP` (3) — all env-overridable.

## Skill doc changes (`defaults/.claude/commands/loom/sweep.md`)

- Absent flag → `BUILDERS_PER_WAVE=auto` sentinel; explicit-integer validation kept verbatim.
- New Stage -1 "Resolve auto wave size" subsection (algorithm + reason table), after `DECIDE`, before daemon/subagent branch; Mode C never reaches it.
- Three wave-partition consumers (Execution Model A/B, dry-run gate, Wave Lifecycle `min(...)`) now reference the resolved size.
- `--dry-run` output prints resolved size + mechanism + gating reason.
- Wave-size guidance table gains an `auto` row; the `<=3` soft-cap reframed as subagent-path-specific.
- Limitations rows updated (disk gate now implemented; soft-cap subagent-specific); stale "does not implement disk-pressure checks" line corrected.
- #3289 "One level deep" prose left intact and cross-referenced.

Plus a one-line pointer in `defaults/CLAUDE.md`.

## Tests

`defaults/scripts/tests/test-disk-headroom.sh` — 15 assertions covering the daemon/subagent × target/candidate/disk/floor matrix, `LOOM_PER_WORKTREE_GB` override, unknown-mechanism rejection, GB conversion via a stubbed `df`, and a regression guard proving the resolved scratch volume is measured (not the repo drive). All pass; shellcheck clean.

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -111,14 +111,14 @@ Combine flags as needed. Always pass `--state open` explicitly (Mode C operates 
 
 ### Optional flags
 
-- **`--builders-per-wave N`** — dispatch up to `N` builders in parallel per wave. Default `1` (fully sequential, matching the MVP behaviour). `N` must be an integer `>= 1`. Honoured in Modes A and B (issue-side); **silently ignored in Mode C** (PR-set mode has no Builder phase — see Mode C validation rules above). Flag tokens are stripped before classification.
+- **`--builders-per-wave N`** — dispatch up to `N` builders in parallel per wave. When **omitted**, the wave size is `auto` — resolved at Stage -1 from the chosen backend and scratch-volume disk headroom (see "Resolve auto wave size"): the daemon detached-process path targets up to 10 concurrent sweeps, while the in-session subagent path stays at the #3289-safe ceiling of 3. When **present**, `N` must be an integer `>= 1` and the explicit value overrides auto entirely (operator wins). Honoured in Modes A and B (issue-side); **silently ignored in Mode C** (PR-set mode has no Builder phase — see Mode C validation rules above). Flag tokens are stripped before classification.
 - **`--dry-run`** — print the planned candidate list (with wave grouping) and EXIT without performing any mutation. Recognized as a bare flag token (no value). May appear anywhere in `$ARGUMENTS`. Default is off. Honoured in **all three** modes — stripped before classification along with other flags. Mode C dry-run prints the PR-set plan (per-PR routing) instead of the issue-set plan.
 - **`--prs`** — switch into Mode C (PR-set mode). Recognized as a bare flag token (no value). May appear anywhere in `$ARGUMENTS`. Default is off. When present, non-flag tokens are interpreted as **PR numbers** (numeric tokens) or as a **PR-list description** (NL tokens). When absent, an NL trigger phrase listed in the Mode C section can still select Mode C. See "Mode C" above for full semantics.
 - **`--no-daemon`** — force in-process subagent dispatch even when the daemon is running with a multi-account token pool. Recognized as a bare flag token (no value). May appear anywhere in `$ARGUMENTS`. Default is off. When present, **Stage -1 (Backend detection) skips the `PROBE_DAEMON` step entirely** and the skill always falls through to the existing Mode A/B/C subagent dispatch path. Honoured in **all three** modes — stripped before classification along with other flags. Use this when you want the predictable single-process behaviour even though daemon dispatch is available (e.g., debugging, demoing the subagent path, or running under a token configuration that you don't want shared with daemon-spawned sweeps). See "Stage -1: Backend detection" below.
 
 ### Validation rules
 
-- Recognize `--dry-run`, `--prs`, `--no-daemon`, and `--builders-per-wave N` as flag tokens anywhere in `$ARGUMENTS`, strip them from the candidate list before validation, and store them as flags / parameters (`DRY_RUN=true|false`, `PRS_MODE=true|false`, `NO_DAEMON=true|false`, `BUILDERS_PER_WAVE=N`).
+- Recognize `--dry-run`, `--prs`, `--no-daemon`, and `--builders-per-wave N` as flag tokens anywhere in `$ARGUMENTS`, strip them from the candidate list before validation, and store them as flags / parameters (`DRY_RUN=true|false`, `PRS_MODE=true|false`, `NO_DAEMON=true|false`, `BUILDERS_PER_WAVE=N`). When `--builders-per-wave` is **absent**, set the sentinel `BUILDERS_PER_WAVE=auto` (not `1`) — Stage -1 resolves the concrete wave size from the backend + disk headroom. An explicit integer is stored verbatim and overrides auto.
 - At least one candidate (numeric token or NL description) must be supplied. If `$ARGUMENTS` (after stripping flag tokens) is empty, display:
   ```
   Usage: /sweep <issue-number> [<issue-number> ...] [--builders-per-wave N] [--dry-run] [--no-daemon]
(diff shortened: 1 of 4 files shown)


rjwalters:
LGTM! Approved.

Reviewed against every acceptance criterion in #3566. This is a clean, well-scoped implementation.

**Disk-headroom helper (`defaults/scripts/lib/disk-headroom.sh`)**
- `loom_worktree_root_free_gb` resolves the worktree root via `loom_worktree_root` and `df`s **that** path (the scratch volume), not the repo drive. It correctly walks up to the nearest existing ancestor before calling `df` (the leaf worktree dir doesn't exist yet), and reports `0` on any `df` failure so the caller floors to a single worktree instead of crashing.
- `df -Pk` is used — POSIX `-P` pins the single-line-per-fs columnar layout and `-k` pins 1K blocks, so parsing `awk 'NR==2 {print $4}'` (Available) is portable across macOS and GNU. GB conversion `avail_k / 1024 / 1024` is correct integer floor.
- `loom_wave_size_from_disk` computes `min(target, floor(free_gb/PER), candidate_count)` with a floor of 1. Walked the branch logic and re-derived every matrix case — no off-by-one. Ties (candidate == disk headroom) resolve to `candidates` as documented; hard floor wins the `floor` reason. Defensive integer-input validation and unknown-mechanism rejection (exit 2) are present.

**#3289 safety ceiling — preserved (the critical check)**
- `LOOM_SUBAGENT_WAVE_CAP=3`, `LOOM_DAEMON_WAVE_TARGET=10`. The subagent target is capped at 3 and is **never** raised toward 10; high parallelism is routed exclusively through the daemon detached-process path (isolated OS processes, not nested subagents).
- The "CRITICAL: One level deep" section is fully intact (sweep.md line 282), and its reinforcements at the Judge/Doctor/Wave-Lifecycle dispatch lines are untouched. The Wave Lifecycle `min(...)` line now reads `min(resolved-wave-size, ...)` with an explicit note that this subagent path caps at 3 and the daemon path never runs it. The change routes *around* #3289 rather than relaxing it.

**sweep.md wiring**
- Absent `--builders-per-wave` → `auto` sentinel; explicit-integer validation (`>=1`, `>3` warning, silent per-wave clamp) kept verbatim and always overrides auto.
- All three wave-partition consumers updated consistently (Execution Model A/B, dry-run "Compute wave partition", Wave Lifecycle `min(...)`). Mode C left size-1 and untouched.
- `--dry-run` header/footer now print resolved size, `auto`-vs-explicit, mechanism (daemon detached-process vs in-session subagent), and the one-line gating reason.
- Wave-size table gains an `auto` row and reframes the `<=3` soft cap as subagent-path-specific; Limitations rows updated (disk gate now implemented, stop-condition still deferred); the stale "does not implement disk-pressure checks" line corrected.

**Tests** — ran `bash defaults/scripts/tests/test-disk-headroom.sh` locally on stock macOS bash 3.2: **15/15 pass**. Coverage includes the daemon/subagent × target/candidate/disk/floor matrix, `LOOM_PER_WORKTREE_GB` override, unknown-mechanism rejection, `df`-stub GB conversion, and — crucially — a `LOOM_WORKTREE_ROOT` stub regression guard that asserts `df` measured the resolved scratch volume and explicitly **not** the repo drive. The helper is bash 3.2-compatible (no `mapfile`/associative arrays in the executable path).

**Hygiene** — CI green (Shellcheck + Installer Integration pass), merge state CLEAN. All edits under `defaults/` (canonical source, incl. `defaults/.claude/commands/loom/sweep.md`), no symlink-view churn. No scope creep, no new labels.

One non-blocking note for a future pass: the Stage -1 prose example uses `mapfile -t` (bash 4+), which wouldn't execute on stock macOS bash 3.2 — but that block is instructional prose for the orchestrating agent, and the load-bearing helper itself is 3.2-safe, so it doesn't affect correctness.

PR #3570, opened 2026-07-15
fix(dogfood): materialize .claude/commands as real copy, not symlink into defaults/ (#3565)

### LINKED ISSUE #3565: Decouple loom's live .claude/commands from shipped defaults/ (co-install safety)
## Problem

Loom dogfoods itself by making its **working** command dir a symlink into its **shipped** tree:

```
.claude/commands -> ../defaults/.claude/commands
```

This conflates two concerns that sibling tools (anvil, [repo](https://github.com/rjwalters/repo)) keep separate:

- `defaults/` = the **distribution artifact** (what `loom-daemon init` ships into consumer repos)
- `.claude/commands` = loom's **live working config** (what loom devs run in this repo)

Because the live dir *is* the shipped dir, anything that writes to loom's `.claude/commands/` lands inside the artifact loom distributes.

## How it bit us

Installing repo-skills into the loom repo (`repo`'s `install.sh` writes commands to `$TARGET/.claude/commands/<ns>/`) silently created `defaults/.claude/commands/repo/` — i.e. it staged all `/repo:*` commands *into loom's shipped defaults*. Committing that would have re-distributed repo's commands to **every loom consumer**. We caught it and backed out, but any co-installed tool (repo, or a future one) hits the same trap, and a `git add -A` on loom would ship it.

Anvil and repo are structurally immune: their live `.claude/` is only ever a **destination**, never a **source**. Anvil's own `.claude/commands/` just contains `loom/` and `repo/` — a pure consumer of both.

## Proposal

Stop symlinking the live tree into the shipped tree. Dogfood by **materializing** loom's live `.claude/commands` as a copy, not a symlink into `defaults/`:

- Keep `defaults/.claude/commands/loom/` as the single source of truth for loom's own commands.
- Generate loom's live `.claude/commands/loom/` the same way a consumer gets it — via `loom-daemon init` (or a thin dev bootstrap) copying + substituting from `defaults/` into the repo's real (gitignored) `.claude/commands/`.
- Result: loom's live `.claude/commands/` becomes a destination like everyone else's, so co-installed tools write there harmlessly and never touch `defaults/`.

The commands are just templated markdown — repo now demonstrates that plain copy + `{{PLACEHOLDER}}` substitution is sufficient for this layer (see repo v0.3.0's `install.sh`). The daemon still owns the genuinely stateful install surface (`.loom/`, hooks, MCP, roles); this only changes how the *command* layer is dogfooded.

## Alternatives considered

- **Gitignore `defaults/.claude/commands/*/` except `loom/`** — narrower; stops the specific pollution from being committed but leaves the symlink smell and still writes into the shipped tree on disk.
- **Leave as-is, document "don't co-install into loom"** — punts the sharp edge onto every operator; rejected.

## Related

- #3563 (retire `/loom:release` in favor of `/repo:release`) — same co-install cleanup theme.

Tracking issue — no urgency; land on loom's own schedule.

rjwalters:
Closes #3565

## Problem

Loom dogfooded itself with a whole-dir symlink `.claude/commands -> ../defaults/.claude/commands`. Because the live command dir *was* the shipped dir, any co-installed tool (e.g. `repo`'s installer) writing `.claude/commands/<ns>/foo.md` landed inside loom's own distribution artifact under `defaults/.claude/commands/<ns>/`, and a `git add -A` would have re-distributed it to every loom consumer.

## Fix

Replace the tracked symlink with a **real, gitignored materialized copy** of `defaults/.claude/commands/loom/`.

Key correctness constraints respected:
1. **Real copy, not a gitignored symlink.** A symlink still redirects on-disk writes into `defaults/`; only a real directory makes the live tree a pure destination. Co-installed tools now write a sibling real dir under the gitignored `.claude/commands/` and never touch the shipped tree.
2. **Materialization lives in the dev bootstrap, not the daemon.** `loom-daemon init` short-circuits (validation-only) on self-install, so the copy is generated by a dogfood-block step in `scripts/install-loom.sh` that mirrors the existing `.claude/agents` handling.
3. **Ignore entry is loom-only.** `.claude/commands` is added to loom's own committed `.gitignore`; the daemon's consumer `update_gitignore` list in `loom-daemon/src/init/post_init.rs` is **untouched** so consumer repos keep `.claude/commands/loom/` tracked.

The materializer builds the fresh copy in a temp dir on the same filesystem and swaps the `loom/` subdir in via an atomic rename, so an in-progress Claude Code session never observes a missing `.claude/commands/`. Command markdown carries no init-substituted placeholders, so a plain copy is byte-identical to a consumer install.

## Files changed

- `.gitignore` — ignore `.claude/commands` (loom's own committed ignore only).
- `.claude/commands` — untracked the legacy symlink blob (`git rm --cached`).
- `scripts/install-loom.sh` — dogfood materializer (temp-dir + atomic rename).
- `defaults/.claude/commands/loom/help.md` — correct the now-inaccurate "symlinks .claude/commands" wording.

No Rust changed; the daemon consumer path is untouched.

## Verification

- `test ! -L .claude/commands` → real directory, not a symlink.
- `diff -r defaults/.claude/commands/loom .claude/commands/loom` → identical (no placeholder survivors).
- `git ls-files .claude/commands` → empty; `git check-ignore .claude/commands` → matched.
- **Pollution simulation**: wrote `.claude/commands/repo/foo.md`, then `git status --porcelain defaults/` was clean, `defaults/.claude/commands/` still contained only `loom/`, and `git add -A` staged nothing under `defaults/.claude/commands`.
- `release.md` seeded in the materialized copy.
- `bash -n` + shellcheck clean for the new block (2 pre-existing SC2115 warnings unrelated to this change).


rjwalters:
LGTM! Reviewed against the three load-bearing correctness constraints from the curator plan, all satisfied:

1. **Real copy, not a gitignored symlink.** The materializer does `cp -R "$CMD_SRC/." "$CMD_TMP/loom-new"` then `mv` into place — a genuine directory, and it `rm -f`s the legacy whole-dir symlink first. `git rm --cached` removed the old `120000` blob; `git ls-tree` on this branch confirms `.claude/commands` is now fully untracked while `defaults/.claude/commands` remains tracked. The pollution trap is closed: sibling namespaces land in the real gitignored live dir, never in `defaults/`.

2. **Materialization in the install-script dev bootstrap, not the daemon.** The new block lives inside the existing `DOGFOOD_MODE` section directly after the `.claude/agents` handling it mirrors. No Rust changed — `git diff main...HEAD -- loom-daemon/` is empty, so `loom-daemon init`'s self-install branch stays validation-only.

3. **Ignore is loom-only; post_init.rs untouched.** The `.claude/commands` entry is in loom's committed `.gitignore` only (root-anchored, so it does not match tracked `defaults/.claude/commands`). The daemon's consumer `update_gitignore` list is untouched, so consumer repos keep `.claude/commands/loom/` tracked.

Also verified:
- **Byte-identical to a consumer install.** `substitute_template_variables` only replaces `{{REPO_OWNER/REPO_NAME/LOOM_VERSION/LOOM_COMMIT/INSTALL_DATE}}`, none of which appear in the command markdown — the `{{ARGUMENTS}}`/imagine tokens are intentionally literal in both trees. So the plain copy needs no substitution, and the leak-detection guard won't fire. `release.md` (42KB) is present in `defaults/` and seeded by the `cp -R`.
- **Atomicity.** Temp dir is created under `$TARGET_PATH/.claude/` (same filesystem), so the swap-in is a true `rename()`. Fresh-clone reproducibility holds — bootstrap regenerates `.claude/commands/loom/` from the tracked `defaults/` source of truth.
- CI: all required checks pass, mergeState CLEAN.

Minor, non-blocking nit for a future pass: the move-aside sequence (`mv old -> loom-old`; then `mv loom-new -> loom/`) leaves a sub-millisecond window where `.claude/commands/loom/` is momentarily absent — a fully single-`rename` swap over a non-empty dir isn't portable to macOS (ENOTEMPTY / no RENAME_EXCHANGE), so this is the correct portable approach. Content is byte-identical either side of the window, so a concurrent reader never sees partial state. Not worth blocking on.

Approved.

Issue #3563, opened 2026-07-15
Retire /loom:release in favor of repo's /repo:release


rjwalters:
## Decision

`/repo:release` now exists in [rjwalters/repo](https://github.com/rjwalters/repo) (shipped in **v0.3.0**) as the canonical, general-purpose release command. It is a port of Loom's `defaults/.claude/commands/loom/release.md`, generalized:

- Keeps the whole methodology: pre-flight/CI gate, **CHANGELOG completeness gate**, **version-drift gate**, semver decision, CHANGELOG draft, bump, tag, GitHub Release.
- Keeps multi-tool version detection (`scripts/version.sh` → cargo-release / cargo set-version / cargo-workspace → bumpversion/bump2version → poetry → npm).
- **Adds** a plain `VERSION`-file tier (the most general case, which Loom's detector never covered).
- **Drops** Loom-only machinery: `{{workspace}}` templating, the `LOOM-EXTENSION-POINT` seams, and the `.loom/context/topics/` methodology-injection hook.

We commonly install both Loom and repo in a consumer repo, so shipping a release command from both is redundant and drifts over time. Agreed plan: **repo owns release; Loom retires its copy and points at repo.**

## Loom-side work

1. Delete `defaults/.claude/commands/loom/release.md` (and any install wiring / `bump.md` overlap that only exists to feed it).
2. Document in Loom's README/CLAUDE.md that releases are handled by `/repo:release` — "install [repo](https://github.com/rjwalters/repo) for the release command."
3. Decide the fate of the `LOOM-EXTENSION-POINT` release seams and the `scripts/version.sh` bundled tooling:
   - `scripts/version.sh` stays useful — repo's `/repo:release` detects and honors it first, so Loom installs keep working unchanged.
   - The topic-injection seams (`pre-push`, `pre-github-release`, …) have **no equivalent in repo** (repo commands are plain prompt files, no runtime injection hook). If any Loom project relies on those seams for release-time customization, capture the requirement before deleting — those workflows would move into the consumer repo's own CLAUDE.md instead.

## Compatibility note

`/repo:release` honors an existing `scripts/version.sh` as the first-priority version tool, so Loom-installed repos that ship that script get identical bump behavior. No consumer action needed beyond installing repo.

Tracking issue only — no rush; land the Loom-side change on Loom's own schedule.

rjwalters:
## Curator enrichment — implementation plan

[...]

Issue #3495, opened 2026-06-15
Generalize loom:release skill so downstream repos don't need to fork it


rjwalters:
## Summary

Generalize the existing `loom:release` skill (`defaults/.claude/commands/loom/release.md`) so that downstream repositories that install Loom get a working release flow out of the box, without having to fork and customize the skill. Loom itself continues to use the same skill via dogfooding.

## Motivation

Loom already ships a `/loom:release` skill, but it hardcodes Loom-specific assumptions:

- Project name `Loom` is baked into the title and overview prose (only `{{workspace}}` is templated).
- A specific number of version-bearing files is named explicitly (and the current copy is internally inconsistent: the overview says "7", Phase 5 says "5", the Important Notes section lists 5 — they should all come from one source of truth).
- Specific file paths are hardcoded: `package.json`, `mcp-loom/package.json`, `loom-daemon/Cargo.toml`, `loom-api/Cargo.toml`, `CLAUDE.md`, `Cargo.lock`.
- Specific build outputs are named: "`loom-daemon` binaries (Apple Silicon + Intel)".
- The semver decision examples reference Loom-only concepts (`ForgeClient`, MCP tools, daemon commands).
- The release-workflow trigger references Loom's `.github/workflows/release.yml` by name; downstream repos may have a different workflow or none.
- Pre-flight assumes `gh run list --branch main` returns rows; in repos with no CI yet (e.g., Anvil), it silently returns empty and the gate has no signal.

Evidence of pain: `../anvil/.claude/commands/loom/release.md` is a hand-edited fork of Loom's skill (created 2026-06-02, Loom's at 2026-06-04). The diff is small — project name swap, 2-file count instead of 7, "no CI yet" notes, anvil-specific commit-scope examples — but every downstream repo that wants a Loom-driven release will need its own fork. That's a maintenance tax that grows with each consumer.

## Proposed approach

Make the skill **discover** the project's release configuration rather than hardcode Loom's:

1. **Project name** — change "preparing a release of **Loom**" to "preparing a release of **{{workspace}}**" or derive from `package.json`/`pyproject.toml`/`Cargo.toml` if present.
2. **Version files** — drive everything off `./scripts/version.sh` output. The script already knows what files it manages; the skill should call `./scripts/version.sh check` or a new `./scripts/version.sh list-files` to learn the count and names rather than hardcoding them. (Loom's own copy of `scripts/version.sh` becomes the reference implementation; downstream installs get a templated version they can edit.)
[...]

Issue #3492, opened 2026-06-14
Reinstall deletes consumer .claude/ files via stale pre-#3450 manifest


rjwalters:
## Summary

Reinstalling Loom over a repo that was previously installed with a **pre-#3450** version silently deletes consumer-authored files under `.claude/` (and any other tracked roots). The #3450 fix corrected manifest *generation* going forward, but the reinstall/prune path still **trusts the stale over-broad manifest already on disk** as authoritative for deletion. Any repo that had Loom ≤0.9.0 *and* other tools writing under `.claude/` is exposed.

## Impact (real occurrence)

A consumer repo (Anvil + studio-local skills alongside Loom) upgraded 0.9.0 → 0.10.1. The 0.10.1 install deleted **11 non-Loom files** it does not own:

```
.claude/skills/anvil-memo/SKILL.md      # anvil pipeline registration shim
.claude/skills/anvil-deck/SKILL.md      # anvil pipeline registration shim
.claude/skills/expand-idea/SKILL.md     # consumer skill
.claude/skills/repo/SKILL.md            # consumer skill
.claude/commands/repo/{audit,branches,conventions,gitignore,links,orphans,readme}.md
```

The two anvil shims being removed silently unregistered the consumer's core `/anvil:memo` and `/anvil:deck` pipelines. All recoverable from git, but the deletion was silent.

## Root cause (verified, not inferred)

1. Loom **0.9.0** install wrote `.loom/install-metadata.json` using the pre-#3450 enumeration that walked the target's entire `.claude/` tree. The 0.9.0 manifest explicitly lists all 11 consumer files above as "installed_files."
2. Loom **0.10.1** reinstall ran its prune step, read that stale manifest, and deleted every path in it that isn't in the new manifest — including the 11 consumer files.

`scripts/install/manifest.sh` (post-#3450) now scopes manifest *generation* to `defaults/`, so a *fresh* manifest is correct. But the reinstall consumes the **old on-disk manifest** written by the buggy version — #3450 does not defend against stale over-broad manifests, and pre-fix installs are exactly the population that has them.

## Proposed fix

When pruning during reinstall/uninstall, don't trust the on-disk manifest blindly. Intersect each candidate-for-deletion against Loom's **current ownership boundary** (`defaults/` enumeration + `defaults/.loom-internal.list`). Never delete a path that Loom's current `defaults/` does not ship; preserve it and emit a warning instead.

[...]

PR #3575, opened 2026-07-15
feat(install): content-gated cleanup of retired release.md stray (#3572)

### LINKED ISSUE #3572: Clean up stale consumer release.md on installer update (follow-up to #3563)
## Context

Follow-up from #3563 (retire `/loom:release` in favor of repo's `/repo:release`), which deleted `defaults/.claude/commands/loom/release.md`.

The #3563 builder deliberately **did not** add `release.md` to the installer's stale-file cleanup sweep, leaving existing consumer copies in place as a documented "harmless stray" (see the note in `defaults/.loom-internal.list`).

## Problem

Any repository that installed Loom **before** #3563 still has `.claude/commands/loom/release.md` on disk. That materializes a dead `/loom:release` command in every already-installed consumer — it points at a workflow Loom no longer ships or supports. New installs are clean; only pre-#3563 installs carry the stray.

## Proposal

Add a one-time cleanup to the installer's update path (`scripts/install-loom.sh`, and the daemon-side update surface if applicable) that removes a stale `.claude/commands/loom/release.md` from consumer repos on the next install/update. Scope it narrowly:

- Only remove the file if it matches the shipped-and-now-deleted content (or is unmodified), to avoid clobbering a consumer that intentionally kept/customized it.
- Idempotent — a no-op once the file is gone.
- Mirror how other retired-file cleanups are handled in the installer, if any precedent exists.

## Acceptance criteria

- [ ] Fresh install: unchanged (no release.md ever created).
- [ ] Pre-#3563 install with an unmodified release.md: the file is removed on next update.
- [ ] Consumer with a modified release.md: left untouched (or removal is clearly gated/logged).
- [ ] Cleanup is idempotent and does not error when the file is already absent.
- [ ] `.loom-internal.list` note updated to reflect that the stray is now actively cleaned up rather than left in place.

## Related
- #3563 (parent — retire `/loom:release`)

rjwalters:
Closes #3572

## Summary

Pre-#3563 consumers still carry `.claude/commands/loom/release.md` on disk, materializing a dead `/loom:release` command. This adds a **one-time, narrow, idempotent** cleanup to the installer so an unmodified stray is removed on the next `install-loom.sh` update, while a consumer-customized copy is preserved.

## Mechanism

A new **retired-file cleanup block** in `scripts/install-loom.sh`, placed immediately after the `--- End stale-file sweep ---` marker (mirroring the removed #3495 migration block's placement). It is gated on a **frozen sha256 allowlist** of every `release.md` version Loom shipped in the #3495 → #3563 window:

- File exists AND content hash ∈ allowlist → `git rm --quiet --force` (same removal primitive + prompt / `--yes` / `--force` semantics as the stale-file sweep).
- File exists but hash matches none → treated as consumer-customized, **left in place** with an informational line (mirrors the `PRESERVED_NOT_OWNED` style).
- Absent → no-op. Idempotent by construction.

**Why a content-hash gate is exact:** `release.md`'s only placeholder is the Claude-Code *runtime* token `{{workspace}}`, which is NOT in the install-time substitution set (`substitute_template_variables`). Shipped bytes are therefore byte-identical across all consumers. The allowlist is frozen — `release.md` will never ship again.

**Why the generic sweep does not already do this:** the #3492 ownership-boundary intersection preserves any stale-manifest path no longer in the current `defaults/` ownership set. A retired file is not owned, so it is preserved by design. That defense-in-depth is correct and unchanged; the explicit cleanup is what removes the stray.

## Scope decision

**install-loom.sh only.** Quick Install / routine `loom-daemon init` (Rust) bypass this sweep entirely — a matching daemon-side retired-file cleanup (`loom-daemon/src/init/`) is intentionally **deferred to a follow-up** to keep this change narrow and reviewable, per the curator's recommendation (A).

## Files

- `scripts/install-loom.sh` — retired-file cleanup block + portable `_loom_sha256` helper.
- `defaults/.loom-internal.list` — #3563 note updated: the stray is now **actively removed on update** (content-gated), not "deleted by hand".
- `scripts/test-installer.sh` — new **Section 5b** mirroring the Section 5 pattern.

## Test results

`bash scripts/test-installer.sh` → **111 passed, 0 failed**. New Section 5b:

- 44a: unmodified (hash-matching) `release.md` reconstructed from git history is gated for removal and git-rm'd; its hash is asserted present in the real `install-loom.sh` allowlist (verifies the shipped-bytes ↔ gate linkage).
- 44b: consumer-modified `release.md` is preserved on disk.
- 44c: absent `release.md` is a no-op.
- 44d: idempotency — a second run with the file already gone is a no-op.
- Drift guard: every digest in the test mirror is present in `install-loom.sh`.

### DIFF
--- a/defaults/.loom-internal.list
+++ b/defaults/.loom-internal.list
@@ -21,8 +21,19 @@
 # Issue #3563: the `/loom:release` skill (`.claude/commands/loom/release.md`)
 # was retired in favor of `/repo:release` from
 # https://github.com/rjwalters/repo. Loom no longer ships a release skill, so
-# there is no internal-list entry for it. Consumers that received the
-# generalized skill during the #3495 window (when it briefly shipped) keep a
-# harmless stray copy; it is not re-installed and can be deleted by hand.
+# there is no internal-list entry for it.
+#
+# Issue #3572: consumers that received the generalized skill during the #3495
+# window (when it briefly shipped) carried a stray copy on disk. That stray is
+# now ACTIVELY REMOVED on the next `install-loom.sh` update by the content-gated
+# retired-file cleanup block (search "Retired-file cleanup" in
+# scripts/install-loom.sh). The removal is gated on a frozen sha256 allowlist of
+# every shipped release.md version, so an unmodified copy is deleted while a
+# consumer-customized copy (hash matches none) is preserved with an
+# informational log line. release.md's only placeholder is the runtime
+# {{workspace}} token (not install-time substituted), so shipped bytes are
+# identical across consumers and the hash gate is exact. The generic stale-file
+# sweep still skips it by design (the #3492 ownership intersection preserves
+# non-owned paths); the explicit cleanup is what removes it.
 # Releases are driven by `/repo:release`, which honors `scripts/version.sh` as
 # its first-priority version tool — see the root CLAUDE.md "Releasing" section.
--- a/scripts/install-loom.sh
+++ b/scripts/install-loom.sh
@@ -1247,6 +1247,130 @@ if (( ${#STALE_FILES[@]} > 0 )); then
 fi
 # --- End stale-file sweep ---
 
+# --- Retired-file cleanup (content-gated) ---
+# A handful of files Loom once shipped were later RETIRED outright (deleted
+# upstream, not renamed). The generic stale-file sweep above deliberately will
+# NOT remove them:
+#   * The #3492 ownership-boundary intersection preserves any stale-manifest
+#     entry that is no longer in the current defaults/ ownership set — a
+#     retired path is by definition not owned, so it lands in
+#     PRESERVED_NOT_OWNED and is never git-rm'd. That defense-in-depth is
+#     correct (it protects consumer-authored files captured by pre-#3450
+#     over-broad manifests) and must stay.
+#   * A retired file that was never in the previous manifest at all is never
+#     even a sweep candidate.
+# So a retired file lingers on disk in every consumer that installed before the
+# retirement. This block removes such strays narrowly and safely.
+#
+# Gate: a frozen allowlist of (retired path -> sha256 of every version Loom
+# ever shipped at that path). A file is removed ONLY when it exists AND its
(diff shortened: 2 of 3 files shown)


rjwalters:
LGTM! Thorough verification as Loom Judge:

**Gating correctness (verified):** The empty-file sha256 `e3b0c442...` is correctly EXCLUDED from the allowlist. Reconstructing `release.md` history via `git log --follow` yields 12 unique blob hashes — one of which is the empty-file digest — and the allowlist carries exactly the 11 non-empty ones. This is the crucial safety property: an empty stray would never be silently removed. Removal is strictly hash-gated (byte-identical-to-shipped), a customized copy (hash matches none) is preserved with an informational line, absent is a no-op, and it's idempotent by construction. The `[[ -n "$file_hash" ]]` guard also correctly preserves (never removes) when no sha256 tool is available.

**Allowlist legitimacy (verified):** All 11 digests reproduce exactly from real git history — not fabricated. The last-shipped blob `b1dac86f...` hashes to `11aef217...`, the head allowlist row, and the test's drift guard (`assert_retired_allowlist_in_sync`) plus the 44a reconstruct-from-blob linkage check are real and passing — genuine shipped-bytes ↔ gate linkage.

**Scope discipline:** install-loom.sh only; daemon-side (`loom-daemon init`) cleanup deferred to a follow-up and stated in the PR body per curator recommendation (A). No over-reach. The generic sweep's #3492 ownership preservation is untouched.

**Test integrity:** `bash scripts/test-installer.sh` → 111 passed, 0 failed (re-run in the PR worktree). Section 5b assertions (44a removal, 44b preserve-modified, 44c absent-noop, 44d idempotency, allowlist drift guard) all pass — sound, not deletion-to-pass. CI Installer Integration Tests pass; merge state CLEAN.

**.loom-internal.list:** note accurately updated to describe active content-gated removal, preservation of customized copies, and why the generic sweep skips it.

Non-blocking nit for a future touch: the new tests reuse labels `44a–44d`, which collides with the pre-existing `Test 44b` (.github over-broad-manifest test) — cosmetic duplicate label in output only, no functional impact.

Approved.

PR #3577, opened 2026-07-15
feat(init): content-gated cleanup of retired release.md in daemon init (#3576)

### LINKED ISSUE #3576: Port content-gated release.md cleanup to loom-daemon init (daemon-side sibling of #3572)
## Context

Follow-up from #3572 / PR #3575, which added a content-gated cleanup of the retired `.claude/commands/loom/release.md` stray — but **only in `scripts/install-loom.sh`** (the shell "Full Install" / update path).

The Rust **Quick Install** path (`loom-daemon init`) bypasses that shell sweep entirely. Its merge-based file sync (`loom-daemon/src/init/file_ops.rs`) iterates *source* files, so a destination-only file (a stray `release.md` that no longer exists in `defaults/`) is never removed — only the destructive `clean_managed_dir` removes it, which the merge path does not call. #3575 explicitly deferred this to a follow-up.

## Problem

A consumer who installed/updates Loom via `loom-daemon init` (Quick Install) still keeps a pre-#3563 stray `.claude/commands/loom/release.md` on disk — a dead `/loom:release` command. Only shell-installer consumers get the #3575 cleanup; daemon-init consumers do not.

## Proposal

Port the #3575 content-gated cleanup into the daemon init path so both install surfaces behave identically:

- On `loom-daemon init` (the merge/update path in `loom-daemon/src/init/`), after the normal file sync, remove `.claude/commands/loom/release.md` **iff** its sha256 matches one of the frozen shipped-release.md digests (the same allowlist #3575 hardcoded in `install-loom.sh`).
- Preserve a consumer's *modified* copy (hash matches none) — never clobber customization.
- Idempotent; no-op when the file is absent.
- **Single source of truth for the allowlist.** #3575 hardcoded the 11 digests in shell. Avoid drift between the two implementations — either share the list via a generated/checked-in data file both consume, or add a test that asserts the Rust allowlist and the shell allowlist are identical. A drift guard is required, not optional.

## Acceptance criteria

- [ ] `loom-daemon init` against a pre-#3563 consumer with an unmodified stray release.md removes it.
- [ ] A modified release.md is preserved.
- [ ] Idempotent / no-op when absent.
- [ ] The empty-file sha256 is NOT in the allowlist (same trap #3575 avoided).
- [ ] A test guarantees the Rust and shell allowlists cannot drift.
- [ ] Rust tests pass (`cargo test -p loom-daemon`).

## Related
- #3572 / PR #3575 (shell-side cleanup — this is the daemon-side sibling)
- #3563 (parent — retire `/loom:release`)

rjwalters:
Closes #3576

## Summary

Ports the shell-side content-gated retired-file cleanup (PR #3575, the `LOOM_RETIRED_FILES` allowlist in `scripts/install-loom.sh`) into the Rust `loom-daemon init` path, so Quick-Install consumers also get a stray `.claude/commands/loom/release.md` removed — parity with the shell "Full Install" / update path.

## Mechanism

The daemon init sync is **source-driven** (iterates `defaults/` files), so a destination-only stray — the `/loom:release` skill retired by #3563 — is never a removal candidate. New `cleanup_retired_files` (module `loom-daemon/src/init/retired.rs`) is called from `initialize_workspace` in `loom-daemon/src/init/mod.rs` **AFTER** scaffolding + `filter_preserved_from_verification_failures` and **BEFORE** `generate_manifest` (so the manifest reflects on-disk state).

For the one retired path:
- present AND `sha256(bytes)` ∈ frozen allowlist → `fs::remove_file` + `report.removed`
- present but hash matches none → preserve (`report.preserved`) — never clobber a consumer customization
- absent → no-op. Idempotent by construction.

The empty-file sha256 (`e3b0c442…`) is **excluded** from the allowlist (a zero-byte file is treated as consumer state — same trap #3575 avoided). Uses the existing `sha2` + `hex` deps (mirrors the `Sha256` pattern at `terminal.rs:76`); no `Cargo.toml` change.

**Self-install safety:** the cleanup call sits after the self-install short-circuit (`mod.rs` early return before scaffolding), so `init` never mutates the Loom source tree. Covered by `test_self_install_skips_retired_file_cleanup`.

## Drift guard (curator Option B — required)

The 11 digests live in a Rust `const RETIRED_FILES: &[(&str, &str)]` mirroring the shell heredoc. A `cargo test -p loom-daemon` test (`test_rust_and_shell_allowlists_in_sync`) parses `scripts/install-loom.sh`'s `LOOM_RETIRED_FILES` block (via a `CARGO_MANIFEST_DIR`-relative path) and asserts **set-equality in both directions** plus cardinality (11 each). Neither surface can change without the other. Verified load-bearing: mutating a single digit makes the test fail with a clear "Only in Rust / Only in shell" diff.

## Tests

`cargo build -p loom-daemon` and `cargo test -p loom-daemon` pass (382 lib tests). New tests in `retired.rs`: sha256 known-vectors, empty-digest exclusion, remove-on-match, preserve-modified, absent-noop, idempotency, plus the drift guard; and `test_self_install_skips_retired_file_cleanup` in `mod.rs`.

(The `integration_basic::test_malformed_json` integration test is a pre-existing socket-timing flake under parallel load — it passes in isolation and is unrelated to this change.)

### DIFF
--- a/loom-daemon/src/init/mod.rs
+++ b/loom-daemon/src/init/mod.rs
@@ -24,6 +24,7 @@
 mod file_ops;
 mod git;
 mod post_init;
+mod retired;
 mod scaffolding;
 mod templates;
 
@@ -33,6 +34,7 @@ use std::path::Path;
 
 use file_ops::{clean_managed_dir, copy_dir_with_report, verify_copied_files, TemplateContext};
 use post_init::{find_overbroad_loom_patterns, generate_manifest, update_gitignore};
+use retired::cleanup_retired_files;
 use scaffolding::setup_repository_scaffolding;
 
 // Re-export public types and functions
@@ -192,6 +194,20 @@ pub fn initialize_workspace(
     // source defaults — flagging them as failures is misleading and the prior
     // "rerun with --force" remediation would clobber the user's intentional edits.
     filter_preserved_from_verification_failures(&mut report);
+
+    // Content-gated cleanup of retired Loom strays (issue #3576). The daemon
+    // init sync is source-driven and never removes destination-only files, so a
+    // stray `.claude/commands/loom/release.md` (the `/loom:release` skill
+    // retired by #3563) lingers on disk for Quick-Install consumers. Remove it
+    // iff its sha256 matches a frozen shipped digest (unmodified); preserve a
+    // customized copy; no-op when absent. Mirrors the shell-side
+    // `LOOM_RETIRED_FILES` block in scripts/install-loom.sh (PR #3575).
+    //
+    // Placed after scaffolding (post-sync) and before generate_manifest so the
+    // manifest reflects on-disk state. The self-install short-circuit above
+    // (returns at ~line 147 before scaffolding) means this never runs on the
+    // Loom source repo — it must not mutate the source tree.
+    cleanup_retired_files(workspace, &mut report);
 
     // Generate installation manifest (.loom/manifest.json)
     generate_manifest(workspace);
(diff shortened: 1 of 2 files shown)


rjwalters:
LGTM! Verified as Loom Judge:

**Correctness** — `cleanup_retired_files` is called in `initialize_workspace` after the self-install short-circuit (early `return Ok(report)` at mod.rs:147), after `setup_repository_scaffolding` (185) and `filter_preserved_from_verification_failures` (194), and before `generate_manifest` — so the manifest reflects on-disk state and init never mutates the Loom source tree. Per-file logic in `cleanup_with_allowlist` is exactly right: absent → `continue` no-op; unreadable → leave in place; `sha256(bytes)` ∈ allowlist → `fs::remove_file` + `report.removed`; hash miss → `report.preserved` (never clobber customization); idempotent by construction. The empty-file digest (`e3b0c442…`) is confirmed excluded (asserted by `test_empty_file_digest_excluded_from_allowlist`).

**Drift guard (load-bearing)** — the Rust `RETIRED_FILES` const holds the same 11 `release.md` digests, in the same order, as the shell `LOOM_RETIRED_FILES` heredoc in `scripts/install-loom.sh`. `test_rust_and_shell_allowlists_in_sync` reads install-loom.sh via a `CARGO_MANIFEST_DIR`-relative path, parses the heredoc (opener `<<'RETIRED'`, bare `RETIRED` closer, `#`/blank skipping, whitespace-split path+hash), and asserts set-equality both directions plus cardinality 11/11. It fails closed: a non-empty guard catches a broken parser, and any single-digit divergence yields an "Only in Rust / Only in shell" diff. Verified the parser matches the real shell syntax on the branch.

**Tests** — ran `cargo test -p loom-daemon` locally: 382 lib tests pass, including all 7 `init::retired` tests (known-vector sha256, empty-digest exclusion, remove-on-match, preserve-modified, absent-noop, idempotency, drift guard) and `test_self_install_skips_retired_file_cleanup`. None vacuous. The builder-flagged `integration_basic::test_malformed_json` passed; the 2 `integration_security` failures I saw under full parallel load (`test_reject_injection_newline`, `test_reject_sensitive_working_directories`) all pass in isolation — same tmux/socket-timing flake class, unrelated to this init-only change.

**Scope** — only `init/mod.rs` (+59) and new `init/retired.rs` (+345). No Cargo change (reuses existing `sha2`/`hex`, mirroring the `terminal.rs:76` pattern). No over-reach.

CI green (Rust Lint/Build + Rust Unit Tests pass), merge state CLEAN. Approved.

PR #3583, opened 2026-07-15
fix(install): ship sync-labels.sh so Quick installs can create labels (#3582)

### LINKED ISSUE #3582: Quick install ships labels.yml but no way to create labels; label workflow silently breaks
## Summary

After a **Quick / files-only install**, a target repo has `.github/labels.yml` on disk but **no shipped, discoverable way to create those labels on the forge** — and no post-install hint that label sync is required. The label-based workflow then silently fails on first use because coordination labels (`loom:building`, `loom:review-requested`, `loom:changes-requested`, `loom:pr`, `loom:triage`, `loom:treating`, `loom:reviewing`, …) don't exist on the repo. Agents fall back to creating labels ad hoc (bare, no color/description) or error out.

Observed with Loom **v0.10.7** (commit `1627a694`), GitHub forge.

## Repro

1. Install Loom into a repo via the Quick / files-only path (no Full Install; the source repo / non-team case).
2. Run any label-driven flow (`/loom:sweep`, `/loom:shepherd`, a Builder claiming an issue).
3. First label transition fails:
   ```
   gh: 'loom:building' not found  (label was never created on the repo)
   ```
4. Inspect the installed tree: `.github/labels.yml` is present, but the only label script shipped is `.loom/scripts/clean-labels.sh` (which **deletes** labels, and is itself deprecated/no-op per #2838). There is **no** create/sync script in the installed tree.
5. Follow `.github/labels.yml`'s own header instruction:
   ```
   # Sync with: ./scripts/install/sync-labels.sh . (from workspace root)
   ```
   → `no such file or directory`. That path exists only in the **Loom source tree**, never in an installed target.

## Root cause

- `scripts/install-loom.sh` **STEP 4** runs `scripts/install/sync-labels.sh` — but that only executes during a **Full Install**. The install banner even scopes it: *"GitHub Changes (if using Full Install): Creates GitHub labels…"*.
- `sync-labels.sh` is **not** in the installed file manifest — it never gets copied into the target repo.
- The installed tree ships the **delete** counterpart (`clean-labels.sh`) but **not** the **create/sync** counterpart. Asymmetric.
- `.github/labels.yml`'s header comment references a source-tree-only path (`./scripts/install/sync-labels.sh`), so a user who follows it hits a missing file.

Net: a Quick-installed repo is missing the labels the whole workflow depends on, with no shipped tool and no accurate instruction to fix it.

## Impact

- Label-driven orchestration (sweep/shepherd/daemon) breaks on first real transition in any Quick-installed repo.
- Agents that "helpfully" create the label inline produce **bare labels** — wrong color, no description — diverging from `labels.yml` and defeating its purpose.
- The one label script that *is* shipped only removes labels, so an operator can't self-serve a fix from the installed tree.

## Suggested fixes (any subset)

1. **Ship a create/sync command into the target tree** — e.g. `.loom/scripts/sync-labels.sh` (self-contained: it already just reads `./.github/labels.yml` and calls the forge; its only source dep is `forge-detect.sh`, and the target already ships `.loom/scripts/lib/forge-helpers.sh`). Give it a CLI surface too, e.g. `loom labels sync`.
2. **Fix the stale instruction** in `.github/labels.yml`'s header to point at whatever ships in the target (`.loom/scripts/sync-labels.sh`) instead of the source-only `./scripts/install/sync-labels.sh`.
3. **Quick-install post-install hint**: print a one-line "labels not synced — run `loom labels sync` (or Full Install) to create them" so the gap is discoverable.
4. Optionally have `verify-install.sh` **check for label presence** (not just that `labels.yml` exists on disk) and warn when the forge is missing the defined set.

## Environment

- Loom v0.10.7 (`1627a694`), installed 2026-07-15, GitHub forge.
- Installed-tree label scripts: `clean-labels.sh` only (delete, deprecated). No sync/create script present.
- 12 labels defined in `labels.yml` were absent from the repo, including `loom:changes-requested` (breaks the Judge→Doctor path), `loom:triage`, `loom:reviewing`, `loom:treating`, `loom:urgent`, `loom:epic`/`loom:epic-phase`, and the `tier:*` set.

rjwalters:
## Summary

After a **Quick / files-only install**, a target repo had `.github/labels.yml` on disk but **no shipped tool to create those labels on the forge**, and `labels.yml`'s header pointed at a source-only path (`./scripts/install/sync-labels.sh`) that doesn't exist in an installed tree. The label-driven workflow then broke on first use (`loom:building` etc. never existed on the repo).

This ships the create/sync tool into the installed tree, fixes the stale instruction, and adds a discoverable post-install hint.

## Changes

1. **Ship a create/sync command** — ported `scripts/install/sync-labels.sh` to `defaults/scripts/sync-labels.sh`, which lands at `.loom/scripts/sync-labels.sh` in an installed target.
   - **Rewired forge detection** from the source-only `forge-detect.sh` (`detect_forge_and_repo` / `FORGE_OWNER` / `FORGE_REPO`) onto the **shipped** `.loom/scripts/lib/forge-helpers.sh` (`forge_detect`, `forge_get_repo_nwo`, `forge_split_nwo`, `gitea_api`).
   - **Adapted the Gitea label ops** to `forge-helpers.sh`'s `gitea_api` contract (response body on stdout + exit code), which differs from the source helper that appended the HTTP status to the body.
   - GitHub label ops (`gh label`) are unchanged from the source.
   - Added `-h/--help` and basic argument validation.
   - The script auto-ships via the install manifest's dynamic `defaults/scripts/*` walk (verified `.loom/scripts/sync-labels.sh` appears in `_emit_installed_files_manifest`) — no manual manifest edit required, matching the existing convention.

2. **Fixed the stale header** in both `defaults/.github/labels.yml` and `.github/labels.yml`: `Sync with:` now points at `.loom/scripts/sync-labels.sh` (the path that ships in the target) instead of the source-only `./scripts/install/sync-labels.sh`.

3. **Quick-install post-install hint** — `finalize_quick_install()` now prints a one-line hint that labels are not yet synced and how to create them (run `.loom/scripts/sync-labels.sh` or use Full Install). Covers both the fresh and reinstall quick paths since both call `finalize_quick_install()`.

Item 4 (optional forge label-presence check in `verify-install.sh`) was **skipped**: the referenced line (~174) is inside a pure file-list builder function, not a verification/probe section, so adding a network forge probe there would be intrusive and out of place — outside the "only if low-risk" bar.

## Testing

- `bash -n` syntax check passes for the new script and `install.sh`.
- **shellcheck clean** on `defaults/scripts/sync-labels.sh`.
- `--help` argument parsing verified.
- Manifest inclusion verified via `_emit_installed_files_manifest`.

No label SET changes and no new GitHub labels created — this only ships tooling to create the already-defined labels.

Closes #3582

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,6 +1,6 @@
 # Loom Workflow Labels
 # These labels coordinate the Loom AI agent workflow
-# Sync with: ./scripts/install/sync-labels.sh . (from workspace root)
+# Sync with: .loom/scripts/sync-labels.sh (from the installed repo root)
 # Or manually: gh label create/edit commands
 
 # Core Workflow States
--- a/defaults/.github/labels.yml
+++ b/defaults/.github/labels.yml
@@ -1,6 +1,6 @@
 # Loom Workflow Labels
 # These labels coordinate the Loom AI agent workflow
-# Sync with: ./scripts/install/sync-labels.sh . (from workspace root)
+# Sync with: .loom/scripts/sync-labels.sh (from the installed repo root)
 # Or manually: gh label create/edit commands
 
 # Core Workflow States
--- a/defaults/scripts/sync-labels.sh
+++ b/defaults/scripts/sync-labels.sh
@@ -1,0 +1,286 @@
+#!/usr/bin/env bash
+# Sync workflow labels from .github/labels.yml onto the forge.
+#
+# Creates (or updates) the Loom coordination labels defined in
+# .github/labels.yml and removes GitHub's noisy default labels. Run this
+# after a Quick / files-only install to create the labels the label-based
+# workflow depends on — a Quick install ships .github/labels.yml but does
+# NOT create the labels on the forge (see issue #3582).
+#
+# Supports both GitHub (via the gh CLI) and Gitea (via the forge API).
+#
+# Usage:
+#   .loom/scripts/sync-labels.sh [WORKTREE_PATH]
+#
+#   WORKTREE_PATH  Directory containing .github/labels.yml and a git remote.
+#                  Defaults to the current directory.
+#
+# This is the installed-tree counterpart of the source-only
+# scripts/install/sync-labels.sh. It is self-contained apart from the
+# shipped .loom/scripts/lib/forge-helpers.sh helper library.
+
+set -euo pipefail
+
+# --- Argument parsing -------------------------------------------------------
+
+usage() {
+  cat <<'EOF'
+Usage: sync-labels.sh [WORKTREE_PATH]
+
+Sync Loom workflow labels from .github/labels.yml onto the forge (GitHub or
+Gitea). Creates missing labels, updates existing ones to match labels.yml,
+and removes GitHub's default labels.
+
+Arguments:
+  WORKTREE_PATH   Directory containing .github/labels.yml (default: .)
+
+Options:
+  -h, --help      Show this help and exit.
+EOF
+}
+
+WORKTREE_PATH="."
+case "${1:-}" in
+  -h|--help)
+    usage
+    exit 0
+    ;;
+  -*)
+    echo "Unknown option: $1" >&2
+    usage >&2
+    exit 2
+    ;;
+  "")
+    ;;
+  *)
+    WORKTREE_PATH="$1"
+    ;;
+esac
+
+# Source the shipped forge-agnostic helper library. Provides forge_detect,
+# forge_get_repo_nwo, forge_split_nwo, and gitea_api.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=lib/forge-helpers.sh
+source "${SCRIPT_DIR}/lib/forge-helpers.sh"
+
+# ANSI color codes
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+BLUE='\033[0;34m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+error() {
+  echo -e "${RED}✗ Error: $*${NC}" >&2
+  exit 1
+}
+
+info() {
+  echo -e "${BLUE}ℹ $*${NC}" >&2
+}
+
+success() {
+  echo -e "${GREEN}✓ $*${NC}" >&2
+}
+
+warning() {
+  echo -e "${YELLOW}⚠ Warning: $*${NC}" >&2
+}
+
+cd "$WORKTREE_PATH"
+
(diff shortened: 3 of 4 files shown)


rjwalters:
## Judge review: APPROVED ✅

Reviewed against the issue #3582 acceptance criteria and curator enrichment. This is a clean, well-scoped fix.

### Verified
1. **Forge-helpers rewire is correct.** The new `defaults/scripts/sync-labels.sh` sources only the shipped `.loom/scripts/lib/forge-helpers.sh` and uses `forge_detect`, `forge_get_repo_nwo`, `forge_split_nwo`, and `gitea_api` — all confirmed present in that library. No leftover references to the source-only `detect_forge_and_repo` / `forge-detect.sh` / `FORGE_API_BASE` (the single `forge-detect.sh` mention is an explanatory comment about the contract difference, not a code path).
2. **Gitea contract adapted correctly.** The source script read `http_code=$(echo "$response" | tail -1)` (status appended to body) and passed the payload positionally. The port correctly switches to forge-helpers' contract — body on stdout, success via exit code — and passes the JSON body via `-d "$payload"` as curl args, checking exit status only. Paths use unprefixed `repos/...` which `gitea_api` normalizes.
3. **Manifest inclusion confirmed.** Ran `_emit_installed_files_manifest` from the worktree — `.loom/scripts/sync-labels.sh` is emitted via the dynamic `scripts/* → .loom/scripts/*` walk. No manual manifest edit needed, matching convention.
4. **Header fix** applied to both `.github/labels.yml` and `defaults/.github/labels.yml` → `.loom/scripts/sync-labels.sh`.
5. **Quick-install hint** added in `finalize_quick_install()`, using the existing `info()` helper (confirmed defined at install.sh:35).

### Testing performed
- `bash -n`: passes. `shellcheck`: CLEAN.
- Parser test harness against the real `.github/labels.yml`: extracts all **26** labels with correct 6-hex colors and descriptions (matches `grep -c '^- name:'`). GitHub path is idempotent (list → edit, else create with "already exists" → edit fallback).
- `--help` exits 0; unknown option exits 2 with usage. `set -euo pipefail` present.
- No label SET change; no new labels. Injection-safe: Gitea payloads built via `python3 json.dumps`, GitHub via quoted `gh label` args (no eval).
- CI: Installer Integration Tests, Shellcheck, Detect Changes all pass; merge state CLEAN.

### Non-blocking notes (no action required)
- The YAML parser is positional (assumes name→description→color are consecutive lines). This mirrors the source script and matches the current file layout, so it's acceptable — worth keeping in mind if `labels.yml` key ordering ever changes.

Item 4 (verify-install.sh forge probe) was reasonably skipped as out-of-place per the PR description. Nice work.

PR #3585, opened 2026-07-16
fix(guard): segment-parse lifecycle/cloud-delete patterns to stop prose false-positives (#3584)

### LINKED ISSUE #3584: guard-destructive.sh: ALWAYS_BLOCK lifecycle/cloud-delete patterns still unanchored — false-positive denies on comments, commit messages, flag names (follow-on to #3553)
## Summary

The #3553 fix (PR #3564, shipped in v0.10.8) correctly added comment-stripping + word-boundary handling to the **ASK-path** and the **SQL-DDL** path, and fixed the rm-scope check for `ssh`/`scp` payloads. But the **`ALWAYS_BLOCK` catastrophic patterns still scan the raw command as unanchored substrings** (v0.10.8 `guard-destructive.sh` line 226: `echo "$COMMAND" | grep -qiE "$pattern"`, and the code comment at line 279 documents this as deliberate). For the genuinely symbolic patterns (`rm -rf /`, fork bomb, `curl | sh`) that is fine. But a few `ALWAYS_BLOCK` entries are ordinary English/identifier substrings, so they still produce false-positive **denies** (worse than the ASK-path, since a deny can't be overridden).

## Evidence — real false-positives from a live VibeSQL session, tested against the v0.10.8 hook

Fed through `defaults/hooks/guard-destructive.sh` @ v0.10.8 (FP = should allow):

| Command (abridged) | v0.10.8 result | Why |
|---|---|---|
| `... # stops billing then the box will halt` | **deny** | `halt` matched inside a trailing comment (ALWAYS_BLOCK not comment-stripped) |
| `git commit -m "recover cleanly after a reboot event"` | **deny** | `reboot` matched inside a commit message |
| `gh pr comment --body "the hazard here is a swallowed delete of a row"` | **deny** | `az.*delete` matched **h·az·ard … delete** — the `.*` spans unrelated tokens |
| `aws ec2 run-instances --instance-initiated-shutdown-behavior stop` | **ask** | `shutdown` matched inside a flag name (friction, not fatal) |

The two the #3553 fix already resolved (now correctly **allow** in v0.10.8): `sqlite3_shutdown` mentioned in a comment, and `ssh host 'rm -f /tmp/x'` (remote payload no longer judged against the local repo root). Thanks for those.

All catastrophic patterns still correctly **deny** in v0.10.8 (`rm -rf /`, force-push to main, `aws ec2 terminate`, `curl|sh`, `gh repo delete`) — no regression there.

## Requested changes

1. **Word-anchor the system-lifecycle block words** (`halt`, `reboot`, `poweroff`, `init 0`, `init 6`) so they only match as a **standalone command** — start-of-string or immediately after a command separator (`;`, `&&`, `||`, `|`, newline) — not as a substring inside comments, commit messages, or longer identifiers. These words are far too common in prose to match unanchored.

2. **Tighten the cloud-delete patterns** (`az.*delete`, `gcloud.*delete`, `az group delete`): the unanchored `.*` matches across unrelated tokens (the `hazard … delete` case above). Anchor the CLI name as a command word — e.g. `(^|[;&|]\s*)(az|gcloud)\s`, and avoid a greedy `.*` that can span an entire line of prose. (`aws ec2 terminate`, `aws s3 rb`, `aws iam delete`, `aws cloudformation delete-stack` are already specific enough and fine.)

3. **Optionally: comment-strip the ALWAYS_BLOCK scan for the substring-prone words only.** The v0.10.8 comment (line 279) keeps ALWAYS_BLOCK scanning the whole string on purpose — reasonable for truly symbolic patterns, but the lifecycle/cloud words demonstrably need at least word-anchoring or comment-stripping. Keep `rm -rf /`, the fork bomb, and `curl|sh` scanning raw.

## Acceptance

- The four FP cases above **allow** (or at most ASK, never deny) after the change.
- The full catastrophic regression set (`rm -rf /`, force-push main, `gh repo delete`, `aws ec2 terminate`, `aws s3 rb`, `curl|sh`, fork bomb, `halt`/`reboot` as **standalone** commands) still **denies** — verified via the existing `tests/hooks/test-guard-destructive.sh` fragment harness.

Follow-on to #3553 / PR #3564. Discovered dog-fooding the guard on a VibeSQL (SQL-engine + heavy AWS/ssh ops) session where these words appear routinely in comments, commit messages, PR bodies, and cloud-CLI flags.
---

## Affected Files

- `defaults/hooks/guard-destructive.sh` — `ALWAYS_BLOCK_PATTERNS` array + the loop at lines 225-229. Lifecycle patterns are at lines 217-222; cloud-delete patterns at lines 206-208 (`gcloud.*delete`, `az.*delete`, `az group delete`). `extract_rm_targets()` (line ~287) is the existing segment-parser to model the lifecycle fix on.
- `tests/hooks/test-guard-destructive.sh` — add allow/ask regression cases; ALWAYS-BLOCK deny cases live in the `--- ALWAYS BLOCK patterns ---` section (lifecycle deny cases at lines 324-354). Helpers: `assert_deny`, `assert_ask`, `assert_allow`.

## Test Plan

- [ ] Automated: add four false-positive regression cases mirroring the evidence table (assert `allow`, or `ask` for the flag-name case) to `tests/hooks/test-guard-destructive.sh`:
  - `assert_allow` — trailing comment containing `halt` (`... # ... the box will halt`)
  - `assert_allow` — commit message containing `reboot` (`git commit -m "recover cleanly after a reboot event"`)
  - `assert_allow` — quoted prose where `az` is a substring of another word + a later `delete` (`gh pr comment --body "the hazard here is a swallowed delete of a row"`)
  - `assert_allow` **or** `assert_ask` — `shutdown` inside a flag name (`aws ec2 run-instances --instance-initiated-shutdown-behavior stop`) — ASK is acceptable per Acceptance
- [ ] Automated regression (must still `deny`): the existing lifecycle deny cases (`reboot`, `sudo reboot`, `shutdown -h now`, `sudo shutdown -r +5`, `halt`, `sudo halt`, `poweroff`, `sudo poweroff`, `init 0`, `init 6`, `foo && reboot`) and cloud cases (`gcloud compute instances delete my-instance`, `aws ec2 terminate`, `aws s3 rb`) must remain green.
- [ ] Manual: run `bash tests/hooks/test-guard-destructive.sh` — all PASS.

## Curator note

See the curator comment below for a root-cause correction to Requested Change #1 (the lifecycle words are already anchored in v0.10.8, but with a whitespace-inclusive boundary that still matches prose — a pure regex tweak cannot fix it; segment parsing is required).

rjwalters:
Closes #3584

## Problem

`defaults/hooks/guard-destructive.sh`'s `ALWAYS_BLOCK` catastrophic scan produced false-positive DENIES on ordinary prose/identifiers:

| Command (abridged) | Before | Why |
|---|---|---|
| `... # stops billing then the box will halt` | deny | `halt` in a trailing comment matched the whitespace-boundary anchor |
| `git commit -m "recover cleanly after a reboot event"` | deny | `reboot` in a commit message |
| `gh pr comment --body "the hazard here is a swallowed delete of a row"` | deny | unanchored `az.*delete` matched h·**az**·ard … **delete** |
| `aws ec2 run-instances --instance-initiated-shutdown-behavior stop` | ask | `shutdown` inside a flag name |

The lifecycle words were already anchored (PR #3564) but to `[;&|[:space:]]`, so whitespace-surrounded prose still matched. A pure regex tweak can't fix it: dropping `[:space:]` breaks the required `sudo halt`/`sudo reboot` DENY cases, since both `sudo halt` and `will halt` are `<word> halt`.

## Fix

Segment-parse instead, mirroring the existing `extract_rm_targets()`:

- Split the command on `;` `&&` `||` `|` and newline.
- Strip a leading `sudo`/`env` wrapper from each segment.
- Deny only when a segment's **command word** (first token) is exactly a lifecycle word (`halt`/`reboot`/`poweroff`/`shutdown`, or `init 0`/`init 6`) — or is `az`/`gcloud` with a `delete` subcommand token.
- Runs against `COMMAND_NO_COMMENT`, so a lifecycle/cloud word in a trailing comment is already gone.

The lifecycle patterns and `az.*delete`/`gcloud.*delete`/`az group delete` are removed from the raw `ALWAYS_BLOCK_PATTERNS` array. `rm -rf /`, the fork bomb, and `curl|sh` keep scanning the raw string (symbolic, not prose-prone); the `aws` multi-token phrases stay in the raw scan too.

## Source of truth

Only `defaults/hooks/guard-destructive.sh` is edited — the repo convention (and the test harness header) treat `defaults/hooks/` as canonical; the tracked `.loom/hooks/` copy is a stale install artifact that the prior fix #3564 also left untouched.

## Tests

`tests/hooks/test-guard-destructive.sh` gains the four documented false-positive cases (3 `assert_allow`, 1 `assert_ask` for the flag-name case) plus command-word regression cases (`az group delete`, `gcloud … delete`, standalone `halt`, `sudo reboot`, `foo && reboot`).

Full suite: **179/179 pass**, including the entire catastrophic regression set (`rm -rf /`, force-push to main, `gh repo delete`, `aws ec2 terminate`, `aws s3 rb`, `curl|sh`, fork bomb, lifecycle standalone commands).

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -197,29 +197,29 @@ ALWAYS_BLOCK_PATTERNS=(
     'wget .* \| .*sh'
     'wget .* -O- \| sh'
 
-    # Cloud infrastructure destruction
+    # Cloud infrastructure destruction. The aws forms below are specific
+    # multi-token phrases, so they stay in this raw substring scan. The az/gcloud
+    # CLIs, by contrast, need command-word anchoring — an unanchored `az.*delete`
+    # matches "h·az·ard … delete" across unrelated prose tokens (#3584) — so they
+    # are handled by the segment-parsed lifecycle/cloud check further below, NOT
+    # here.
     'aws s3 rm.*--recursive'
     'aws s3 rb'
     'aws ec2 terminate'
     'aws iam delete'
     'aws cloudformation delete-stack'
-    'gcloud.*delete'
-    'az.*delete'
-    'az group delete'
 
     # Docker mass destruction
     'docker system prune'
 
-    # System reboot/shutdown — command-position + trailing-boundary anchored so
-    # a flag name (e.g. --instance-initiated-shutdown-behavior) or the word in a
-    # comment/message ("this reboots the box") no longer trips them, while the
-    # bare command (optionally sudo-prefixed, or after a separator) still denies.
-    '(^|[;&|[:space:]])reboot([;&|[:space:]]|$)'
-    '(^|[;&|[:space:]])shutdown([;&|[:space:]]|$)'
-    '(^|[;&|[:space:]])halt([;&|[:space:]]|$)'
-    '(^|[;&|[:space:]])poweroff([;&|[:space:]]|$)'
-    '(^|[;&|[:space:]])init[[:space:]]+0([;&|[:space:]]|$)'
-    '(^|[;&|[:space:]])init[[:space:]]+6([;&|[:space:]]|$)'
+    # NOTE: system-lifecycle commands (halt/reboot/poweroff/shutdown/init 0/
+    # init 6) are deliberately NOT in this raw substring scan. Even the
+    # whitespace-inclusive boundary anchor they used to carry still fired inside
+    # ordinary prose ("...the box will halt", "...after a reboot event"), and a
+    # pure regex tweak can't separate `sudo halt` from `will halt` (both are
+    # "<word> halt"). They are handled by the segment-parsed check below, which
+    # denies only when a segment's *command word* is exactly the lifecycle word
+    # (#3584).
 )
 
 for pattern in "${ALWAYS_BLOCK_PATTERNS[@]}"; do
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Solid, well-scoped fix for the false-positive DENY problem in the `ALWAYS_BLOCK` scan. The segment-parse approach (`lifecycle_or_cloud_reason()`) is the correct one — as the curator note established, a pure regex tweak can't separate `sudo halt` from `will halt` since both are `<word> halt`, so command-word anchoring via segment splitting is exactly right, and it mirrors the existing `extract_rm_targets()` pattern.

## Verification

**CI**: all applicable checks pass (Installer Integration Tests + Detect Changes pass; Rust/MCP jobs correctly skip — only shell files changed). Merge state CLEAN.

**Test suite**: ran `bash tests/hooks/test-guard-destructive.sh` myself — **179/179 pass**, including the full catastrophic regression set and the four new #3584 regression cases.

**Direct hook probes** (piped JSON through the modified `defaults/hooks/guard-destructive.sh`):
- Acceptance FP cases → all ALLOW (`halt` in trailing comment, `reboot` in commit message, `hazard…delete` prose) except the `aws ec2 --instance-initiated-shutdown-behavior` flag case which correctly ASKs (never DENY). ✅
- Catastrophic set → all DENY: `rm -rf /`, force-push main, `gh repo delete`, `aws ec2 terminate`, `aws s3 rb`, `curl|sh`, fork bomb, standalone/`sudo` `halt`/`reboot`, `poweroff`, `init 0/6`, `gcloud … delete`, `az group delete`. ✅
- Extra bypass probes → `echo done | halt`, `true; halt`, `sudo shutdown -h now`, `gcloud projects delete` all still DENY; more prose (`grep -r reboot /etc`, `echo "...halt the pipeline..."`) correctly ALLOW. ✅

## Scope/hygiene

Only `defaults/hooks/guard-destructive.sh` + `tests/hooks/test-guard-destructive.sh` touched. Correctly leaves the stale `.loom/hooks/` install artifact alone (defaults/ is source of truth, consistent with prior fix #3564). No unrelated changes, no new labels. `printf '%s'` usage is format-string-safe. `Closes #3584` present.

## One non-blocking follow-up

`env FOO=bar halt` slips through to ALLOW — the `env ` strip removes only the `env ` prefix, not a following `VAR=val` assignment, so the command word parses as `FOO=bar`. This was DENIED by the old regex, so it's a narrow regression — but it's the same class of obfuscated-invocation gap the guard has never covered (`/sbin/halt`, `bash -c halt` were never caught either), and the guard is explicitly best-effort. Filed as #3586 to track; not a blocker for this fix, which correctly prioritizes eliminating the non-overridable prose false-positive DENIES.

PR #3587, opened 2026-07-16
fix(guard): resolve command word past env NAME=value assignments in lifecycle guard

### LINKED ISSUE #3586: guard-destructive: env VAR=val <lifecycle-cmd> bypasses command-word anchoring
## Context

Discovered while judging PR #3585 (segment-parsed lifecycle/cloud-delete anchoring, closes #3584).

The new `lifecycle_or_cloud_reason()` segment parser in `defaults/hooks/guard-destructive.sh` strips a leading `env ` wrapper via `sub(/^env[ \t]+/, "", seg)` but does not skip `VAR=val` environment assignments that `env` (and the shell) accept before the command word.

## Repro

```
env FOO=bar halt   -> ALLOW   (command word parsed as "FOO=bar")
env halt           -> DENY    (works, since no assignment)
```

The old whitespace-boundary regex denied `env FOO=bar halt`, so this is a narrow regression introduced by the command-word anchoring.

## Severity: low

This is the same class of obfuscated-invocation gap the guard has never covered (`/sbin/halt`, `bash -c halt`, `command halt` all slip through, before and after #3585). The guard is a documented best-effort safety net, not a security boundary. The #3584 fix correctly prioritized eliminating non-overridable prose false-positive DENIES, which was the actual reported harm.

## Possible fix

In the `env`-strip step, loop-strip leading `NAME=value` tokens (and `env` flags like `-i`, `-u NAME`) before taking the command word, mirroring how a shell resolves the command word past env assignments. Add a regression case `assert_deny "env FOO=bar halt"` to `tests/hooks/test-guard-destructive.sh`.

Discovered during evaluation of PR #3585.

rjwalters:
## Summary

`defaults/hooks/guard-destructive.sh` — the `lifecycle_or_cloud_reason()` segment parser (added by #3585) stripped only a bare `env ` wrapper via `sub(/^env[ \t]+/, "", seg)` before taking the first whitespace token as the command word. For `env FOO=bar halt` the segment became `FOO=bar halt`, so the command word resolved to `FOO=bar` (not `halt`) and the command slipped past the lifecycle deny. `env halt` (no assignment) correctly denied. This was a narrow regression from the #3585 command-word anchoring.

## Changes

- `defaults/hooks/guard-destructive.sh`: in the `env`-strip step, after stripping a leading `env`, loop-strip leading env flags (`-i`, two-token `-u NAME`, `--`) and `NAME=value` environment assignments before taking the command word — mirroring how a shell/env resolves the command word past an env prelude. Portable awk only (BSD/POSIX), consistent with `extract_rm_targets()`. The fix lives entirely inside `lifecycle_or_cloud_reason()`; the catastrophic `ALWAYS_BLOCK` scan is untouched.
- `tests/hooks/test-guard-destructive.sh`: added five #3586 regression cases in the #3584 block covering `env halt`, `env FOO=bar halt`, multi-assignment, `env -i FOO=bar halt`, and the two-token `env -u NAME reboot` form.

| `env FOO=bar halt` → DENY | ✅ | New assertion passes |
| `env -i FOO=bar halt` → DENY | ✅ | New assertion passes |
| Fix confined to `lifecycle_or_cloud_reason()`; ALWAYS_BLOCK untouched | ✅ | Diff only edits the env-strip step in that function |
| `env halt` still DENY (loop must not skip a bare command word) | ✅ | New assertion passes |
| `sudo halt`, standalone `halt`, `foo && reboot`, `sudo reboot` still DENY | ✅ | Existing #3584 regressions green |
| Four #3584 prose false-positive ALLOW/ASK cases still pass | ✅ | Existing assertions green |
| `az group delete` / `gcloud ... delete` command words still DENY | ✅ | Existing assertions green |
| Full suite passes | ✅ | 184/184 passed |

## Test Plan

`bash tests/hooks/test-guard-destructive.sh` → `Total: 184, Passed: 184, Failed: 0, ALL TESTS PASSED`. The test targets the canonical `defaults/hooks/guard-destructive.sh` source of truth (not the gitignored `.loom/hooks/` install artifact), matching the #3585 convention.

Closes #3586

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -281,7 +281,25 @@ lifecycle_or_cloud_reason() {
             seg = segs[i]
             sub(/^[ \t]+/, "", seg)
             sub(/^sudo[ \t]+/, "", seg)
-            sub(/^env[ \t]+/, "", seg)
+            # Strip a leading `env` wrapper, then loop-strip the env flags and
+            # NAME=value assignments a shell resolves past before the command
+            # word, so `env FOO=bar halt` resolves to command word `halt` (not
+            # `FOO=bar`) and still denies. `env -i FOO=bar halt` and `env -u
+            # NAME halt` likewise resolve to `halt`. A bare `env halt` (no
+            # assignment) is unaffected — the loop matches nothing and leaves
+            # `halt` as the command word. Portable awk only (no GNU/BSD-specific
+            # escapes), consistent with extract_rm_targets(). (#3586)
+            if (sub(/^env([ \t]+|$)/, "", seg)) {
+                sub(/^[ \t]+/, "", seg)
+                stripped = 1
+                while (stripped) {
+                    stripped = 0
+                    if (sub(/^-u[ \t]+[^ \t]+([ \t]+|$)/, "", seg)) { stripped = 1; continue }
+                    if (sub(/^-i([ \t]+|$)/, "", seg))              { stripped = 1; continue }
+                    if (sub(/^--([ \t]+|$)/, "", seg))              { break }
+                    if (sub(/^[A-Za-z_][A-Za-z0-9_]*=[^ \t]*([ \t]+|$)/, "", seg)) { stripped = 1; continue }
+                }
+            }
             sub(/^[ \t]+/, "", seg)
             m = split(seg, toks, /[ \t]+/)
             if (m == 0) continue
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -753,6 +753,21 @@ assert_deny "Regression (#3584): 'sudo reboot' still denied" \
 assert_deny "Regression (#3584): 'foo && reboot' still denied" \
     "foo && reboot"
 
+# #3586: `env` wrapper with NAME=value assignments / flags must resolve the
+# command word past the env prelude and still DENY. `env halt` (no assignment)
+# already worked; the assignment forms regressed under the #3585 command-word
+# anchoring because `toks[1]` was `FOO=bar` instead of `halt`.
+assert_deny "Regression (#3586): 'env halt' still denied" \
+    "env halt"
+assert_deny "Regression (#3586): 'env FOO=bar halt' resolves command word past assignment" \
+    "env FOO=bar halt"
+assert_deny "Regression (#3586): 'env FOO=bar BAZ=qux halt' skips multiple assignments" \
+    "env FOO=bar BAZ=qux halt"
+assert_deny "Regression (#3586): 'env -i FOO=bar halt' skips flag + assignment" \
+    "env -i FOO=bar halt"
+assert_deny "Regression (#3586): 'env -u NAME reboot' skips two-token -u flag" \
+    "env -u SOMEVAR reboot"
+
 echo ""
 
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the `env NAME=value` command-word bypass fix in `lifecycle_or_cloud_reason()`. This is a clean, well-scoped fix for the #3586 regression introduced by #3585's command-word anchoring.

## Correctness

The new awk env-strip loop is properly gated on `env` actually being present: `if (sub(/^env([ \t]+|$)/, "", seg))`. The `([ \t]+|$)` anchor means prose like `environment` (env + non-whitespace) never matches, and bare/non-env segments are untouched. Inside the gate it loop-strips, in the order a shell/env resolves the command word:
- `-i` flag
- two-token `-u NAME` (`^-u[ \t]+[^ \t]+`)
- `--` (breaks the loop — everything after is the command)
- `NAME=value` assignments (`^[A-Za-z_][A-Za-z0-9_]*=[^ \t]*`)

Then `toks[1]` resolves to the real command word. The awk is portable (POSIX ERE character classes `[ \t]`, `[A-Za-z_]`, alternation `([ \t]+|$)` — no `\s`, no GNU `\<`), consistent with `extract_rm_targets()`.

**Bypass-hole probing** (direct pipes through the modified hook):
- `env FOO=bar halt` → **deny** (was the reported ALLOW bug)
- `env -- halt` → **deny** (the `--` break case resolves correctly)
- `env -i FOO=bar halt` → **deny**
- `env -u SOMEVAR reboot` → **deny**
- `env FOO=bar BAZ=qux halt` → **deny** (multi-assignment)
- `env halt` → **deny** (loop matches nothing, bare command word preserved — no over-strip)
- `env FOO=bar` (no command) → sane ALLOW (assignment stripped, `m==0` → `continue`, no phantom command word)

No over-stripping of real command words observed, and no newly-introduced bypass in the covered forms. The one theoretically-slippery form (`env -- FOO=bar halt`) is env's own ambiguous territory and remains best-effort, consistent with the guard's documented scope (`/sbin/halt`, `bash -c halt` were never covered).

## Must-preserve regressions

All hold: `sudo halt`, standalone `halt`, `foo && reboot`, `poweroff`, `init 0`, `rm -rf /`, `az group delete`, `gcloud compute instances delete` all → **deny**. The four #3584 prose false-positives still ALLOW/ASK (3× allow + the `--instance-initiated-shutdown-behavior` ASK case), and the fix is confined to the env-strip step — the catastrophic `ALWAYS_BLOCK` scan is untouched.

## Tests

Five new #3586 regression cases added to `tests/hooks/test-guard-destructive.sh` (bare `env halt`, single/multi assignment, `-i` + assignment, two-token `-u NAME`). Ran the suite myself: **184/184 passed, 0 failed**. Sanity pipes of `env FOO=bar halt` (deny) and the prose FP cases (not-deny) through the modified hook confirm behavior directly.

## Scope & hygiene

Only `defaults/hooks/guard-destructive.sh` (+19/-1) and `tests/hooks/test-guard-destructive.sh` (+15) changed — the canonical source of truth, not the gitignored `.loom/hooks/` artifact. `Closes #3586`, no new labels. CI: relevant checks pass (Installer Integration Tests, Detect Changes), merge state CLEAN.

Solid, tightly-scoped safety-net improvement. Merging.

PR #3591, opened 2026-07-16
fix(install): make .gitignore install/uninstall round-trip byte-idempotent (#3590)

### LINKED ISSUE #3590: install/uninstall .gitignore round-trip is non-idempotent (reorders managed block)
### Summary

Follow-up to #3588 (fixed by #3589). #3589 fixed the user-facing **data-loss** symptom: during `install.sh --quick`, a stashed uncommitted `.gitignore` edit no longer strands silently — the pop path resets `.gitignore` to `HEAD`, applies the user hunk cleanly, re-appends Loom patterns idempotently (`reapply_loom_gitignore_patterns`), and surfaces any residual conflict with a recovery path (exit 0).

This issue tracks the **underlying root cause** that #3589 worked around rather than eliminated: the uninstall → init round-trip is not byte-idempotent for `.gitignore`. This is a **robustness/consistency** gap, **not** a data-loss bug — #3589 already handles the conflict gracefully. **Low priority.**

### Root cause (verified against source)

The non-idempotency is the round-trip between two independently-maintained code paths that have **drifted** from each other. Two compounding defects:

**1. Pattern-list drift (uninstall removes only a stale subset).**
- `loom-daemon`'s `update_gitignore` (`loom-daemon/src/init/post_init.rs:107`) writes **30** `ephemeral_patterns` (`post_init.rs:117-148`) under the header `# Loom runtime state (don't commit these)` (`post_init.rs:218`). It is append-only and idempotent, but the presence check is a **substring** `new_contents.contains(pattern)` (`post_init.rs:201`), not an exact-line match, and it **always appends at EOF** (`post_init.rs:199-209`).
- `scripts/uninstall-loom.sh`'s `.gitignore` smart-removal (`uninstall-loom.sh:1008-1054`) strips a hardcoded `LOOM_PATTERNS` list of only **4 real patterns** — `.loom/state.json`, `.loom/worktrees/`, `.loom/*.log`, `.loom/*.sock` (`uninstall-loom.sh:1015-1021`) — plus one **stale** comment pattern `# Loom - AI Development Orchestration` (`uninstall-loom.sh:1020`) that no longer matches the header init actually writes. Net effect: uninstall leaves ~26 of the 30 Loom patterns in the file and never removes the comment header. The block is not cleanly removed at all.

**2. Whole-file reflow + EOF re-append (the actual "move").**
- After removing its 4 patterns, uninstall runs two `awk 'NF || prev_blank++ < 1 {...}'` passes (`uninstall-loom.sh:1035` and `:1040`) that collapse consecutive blank lines and trim trailing blanks **across the entire file** — mutating the user's own spacing, not just the Loom region.
- On the next install, `update_gitignore` re-appends the 4 stripped patterns at **EOF**. So those lines **move** relative to `HEAD` on every uninstall → install cycle even with zero user edits, and the surrounding blank-line layout is reflowed.

Because uninstall strips a drifted subset and init re-appends at EOF, `init(uninstall(X)) != X`.

### Why it still matters after #3589

- #3589's clean-apply treatment (reset-to-HEAD before stash pop, then `reapply_loom_gitignore_patterns`) is **`.gitignore`-specific** and lives in `install.sh` (repo root), not the shared removal logic. Any *other* Loom-managed tracked file a user edited uncommitted during a `--quick` reinstall still hits the (now gracefully surfaced) conflict path.
- Making the round-trip byte-stable lets the `.gitignore`-specific reset-to-HEAD workaround in `install.sh` be simplified or generalized (stretch goal).

### Recommended direction

The three suggestions in the parent framing (in-place removal / write-back to original location / round-trip test) each address a symptom. The most robust fix is a **marker-delimited managed block** — a refinement that combines directions 1 and 2 and makes both add and remove operate on a single, self-locating, contiguous region:

1. **`update_gitignore` (`post_init.rs`)**: wrap the emitted patterns in stable sentinels, e.g. `# >>> loom-managed (do not edit) >>>` … `# <<< loom-managed <<<` (mirroring the `<!-- BEGIN/END LOOM ORCHESTRATION -->` convention already used for CLAUDE.md in `uninstall-loom.sh:933-945`). On update, replace the existing block **in place** rather than appending at EOF; only create-at-EOF when no block exists (one-time migration for markerless installs).
2. **`uninstall-loom.sh` (`.gitignore` block, `:1008-1054`)**: delete the whole marked span in one pass. Drop the hardcoded 4-item `LOOM_PATTERNS` list and the two whole-file `awk` reflow passes so user spacing outside the block is never touched.
3. Single source of truth for the pattern set so the two paths can never drift again (share via a generated file, or have uninstall key off the markers exclusively rather than an enumerated list).

If the marker-block change is judged too large for a low-priority item, the minimum viable fix is: (a) collapse the `awk` reflow to operate only within the removed region, and (b) replace uninstall's drifted 4-item list with the full authoritative set — but this remains fragile w.r.t. block position and is not byte-stable unless the block is already at EOF.

### Acceptance criteria

- [ ] An `install → uninstall → install` round-trip on a repo whose committed `.gitignore` has no user edits leaves `.gitignore` **byte-identical** to its pre-uninstall content (verified by `diff`/`cmp` / `git diff --exit-code`).
- [ ] `uninstall-loom.sh` removes the **entire** Loom-managed `.gitignore` region (all patterns + header), not a drifted 4-item subset — no residual Loom patterns remain after uninstall.
- [ ] `uninstall-loom.sh` no longer reflows blank lines **outside** the Loom-managed region (user spacing preserved verbatim).
- [ ] The uninstall and init pattern sets cannot silently drift (single source of truth, or marker-scoped removal that does not enumerate patterns).
- [ ] A regression test asserts the byte-identical round-trip (extend `scripts/test-install-quick-gitignore-stash.sh` or add a sibling; a Rust unit test in `post_init.rs` covers the in-place-update / marker path).
- [ ] (Stretch) The `.gitignore`-specific reset-to-HEAD workaround in `install.sh` (`reapply_loom_gitignore_patterns`) is simplified or generalized once the round-trip is stable.

### Affected Files

- `loom-daemon/src/init/post_init.rs` — `update_gitignore` (`:107`), `ephemeral_patterns` (`:117-148`), EOF append + substring presence check (`:199-209`); add marker-delimited in-place block write and a round-trip unit test.
- `scripts/uninstall-loom.sh` — `.gitignore` smart-removal block (`:1008-1054`), drifted `LOOM_PATTERNS` (`:1015-1021`), whole-file `awk` reflow passes (`:1035`, `:1040`); replace with marked-span deletion.
- `install.sh` — `reapply_loom_gitignore_patterns` (`:232`) and the stash-pop reset-to-HEAD block (the #3589 workaround eligible for simplification once the round-trip is stable).
- `scripts/test-install-quick-gitignore-stash.sh` — existing #3589 regression test; extend (or add a sibling) with the byte-identical round-trip assertion.
- `scripts/test-installer.sh` — installer harness (Tests 31/49 already exercise uninstall smart-removal); wire in the round-trip check if a shell-level test is preferred.

### Related

- Parent bug: #3588
- Fix PR: #3589

rjwalters:
Closes #3590

## Problem

The install/uninstall `.gitignore` round-trip was non-idempotent (root cause behind #3588, only worked around by #3589):

1. **Pattern-list drift** — `uninstall-loom.sh` hardcoded only 4 patterns + a stale header, while `post_init.rs::update_gitignore` wrote ~30 patterns. Uninstall left ~26 patterns + the header behind.
2. **Whole-file reflow + EOF re-append** — uninstall's two `awk` passes collapsed blank lines across the entire file, and init always re-appended at EOF, so the block moved every uninstall -> install cycle.

## Fix (marker-delimited managed block)

Wrap the Loom `.gitignore` region in sentinels (mirroring the `<!-- BEGIN/END LOOM ORCHESTRATION -->` convention used for CLAUDE.md):

```
# >>> loom-managed (do not edit) >>>
# Loom runtime state (don't commit these)
<~30 runtime patterns>
# <<< loom-managed <<<
```

- **`post_init.rs`**: `update_gitignore` refreshes the marked block **in place** (never moves it); appends at EOF only when no block exists; migrates legacy markerless blocks to the marked form. Pattern list hoisted to a module-level `EPHEMERAL_PATTERNS` const (single source of truth). Line ops split/join on `\n` so untouched regions stay byte-exact.
- **`uninstall-loom.sh`**: deletes the whole marked span in one exact-match `awk` pass; drops the drifted 4-item list and both whole-file blank-line reflow passes. A legacy fallback still removes the old headers + full pattern set for pre-marker installs.

The #3589 reset-to-HEAD stash workaround in `install.sh` is left untouched and still passes its suite; it can be simplified in a later pass now that the round-trip is byte-stable.

## Verification

- `cargo test --lib post_init` — 17 pass, incl. new tests: marker wrapping, in-place update (no move/duplicate), legacy migration idempotency, and a byte-identical install/uninstall/install round-trip.
- `cargo build --release` + `cargo clippy --lib` — clean for changed code (only a pre-existing unrelated pedantic warning in a forge file remains).
- `bash -n scripts/uninstall-loom.sh` — OK.
- `scripts/test-installer.sh` — 111 pass (Test 18 + legacy Test 48 gitignore paths green).
- `scripts/test-install-quick-gitignore-stash.sh` — 16 pass (5 new #3590 checks). **Verified the new round-trip scenario FAILS against pre-change code** (byte movement + ~26 residual patterns after uninstall) and passes after the fix.

### DIFF
--- a/loom-daemon/src/init/post_init.rs
+++ b/loom-daemon/src/init/post_init.rs
@@ -63,6 +63,81 @@ pub fn find_overbroad_loom_patterns(workspace_path: &Path) -> Vec<String> {
     found
 }
 
+/// Sentinel that opens the Loom-managed `.gitignore` block.
+///
+/// The managed block is delimited by [`GITIGNORE_BEGIN_MARKER`] and
+/// [`GITIGNORE_END_MARKER`] so that both `update_gitignore` (add/refresh) and
+/// `scripts/uninstall-loom.sh` (remove) operate on a single, self-locating,
+/// contiguous region. This mirrors the `<!-- BEGIN/END LOOM ORCHESTRATION -->`
+/// convention already used for CLAUDE.md. See issue #3590.
+pub const GITIGNORE_BEGIN_MARKER: &str = "# >>> loom-managed (do not edit) >>>";
+
+/// Sentinel that closes the Loom-managed `.gitignore` block.
+pub const GITIGNORE_END_MARKER: &str = "# <<< loom-managed <<<";
+
+/// Human-readable header emitted inside the managed block. Pre-#3590 installs
+/// wrote this same line as a bare (markerless) header; detecting it lets us
+/// migrate those installs to the marked form in place.
+const GITIGNORE_BLOCK_HEADER: &str = "# Loom runtime state (don't commit these)";
+
+/// Ephemeral/runtime files that should be ignored — the single source of truth
+/// for the Loom-managed `.gitignore` block.
+///
+/// Keep in sync with the Loom source repo's `.gitignore`.
+///
+/// Phase 3.5 (#3402, epic #3372): removed patterns for retired daemon-brain
+/// state files (`daemon-state.json`, archived `[0-9][0-9]-daemon-state.json`,
+/// `progress/`, `stuck-history.json`, `alerts.json`, `health-metrics.json`)
+/// and added `spawn-loop-state.json` for the Phase 1 spawn loop (#3374).
+pub const EPHEMERAL_PATTERNS: &[&str] = &[
+    ".loom-in-use",
+    ".loom-checkpoint",
+    ".loom/.daemon.pid",
+    ".loom/.daemon.log",
+    ".loom/daemon.sock",
+    ".loom/daemon-loop.pid",
+    ".loom/daemon-metrics.json",
+    ".loom/loom-source-path",
+    ".loom/spawn-loop-state.json",
+    ".loom/issue-failures.json",
+    ".loom/interventions/",
+    ".loom/worktrees/",
+    ".loom/state.json",
+    ".loom/mcp-command.json",
+    ".loom/activity.db",
+    ".loom/claims/",
+    ".loom/signals/",
+    ".loom/status/",
+    ".loom/retry-state/",
+    ".loom/diagnostics/",
+    ".loom/guide-docs-state.json",
+    ".loom/metrics_state.json",
+    ".loom/manifest.json",
+    ".loom/stuck-config.json",
+    ".loom/metrics/",
+    ".loom/usage-cache.json",
+    ".loom/claude-config/",
+    ".loom/*.log",
+    ".loom/*.sock",
+    ".loom/logs/",
+];
+
+/// Build the Loom-managed `.gitignore` block (marker lines + header + patterns),
+/// with no leading or trailing newline. Callers add surrounding newlines.
+fn managed_gitignore_block() -> String {
+    let mut block = String::new();
+    block.push_str(GITIGNORE_BEGIN_MARKER);
+    block.push('\n');
+    block.push_str(GITIGNORE_BLOCK_HEADER);
+    block.push('\n');
+    for pattern in EPHEMERAL_PATTERNS {
+        block.push_str(pattern);
+        block.push('\n');
+    }
+    block.push_str(GITIGNORE_END_MARKER);
+    block
+}
+
 /// Generate installation manifest by running verify-install.sh
 ///
 /// Attempts to run `.loom/scripts/verify-install.sh generate --quiet` to create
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge review — LGTM, approved

Reviewed the `update_gitignore` rewrite and the uninstall `awk` span-deletion carefully, and ran the full practical suite locally on `feature/issue-3590`.

### Correctness
- **Byte-exact in-place replacement**: `contents.split('\n')` + `lines.join('\n')` is byte-preserving (the trailing empty element reproduces the file's final newline exactly), and `lines.splice(b..=e, block_lines)` swaps only the marked span, leaving untouched regions verbatim. No off-by-one on the marker lines. Confirmed by the `roundtrip_is_byte_identical` and `in_place_update_does_not_move_or_duplicate_block` tests.
- **Migration matrix** all handled: no block (append-at-EOF with a single trailing newline, the exact inverse of uninstall's span deletion), legacy markerless block (header + bare `EPHEMERAL_PATTERNS` stripped via `retain`, then re-emitted marked), already-marked (splice → identical → `new_contents == contents` short-circuits the write), and user edits inside the old block (custom lines survive outside the new block). `migrates_legacy_markerless_block_in_place` proves post-migration idempotency.
- **Uninstall**: the single `$0 == b { inblock } ... { print }` pass removes exactly BEGIN..END inclusive with no regex escaping, and both whole-file blank-line reflow passes are gone — user spacing is preserved verbatim. The legacy fallback enumerates the full authoritative pattern set (fixing the drifted 4-item list), gated on `grep -qxF` for both markers so it only fires on pre-marker installs.
- **`set -e` / quoting**: the `awk … && mv` compound is safe (mv failure surfacing is intended); the legacy loop keeps the `|| true` grep guard; all vars quoted. The #3589 `set -e` capture bug is in untouched code and its Scenario 5 still passes.
- **#3589 reset-to-HEAD workaround** still functions with the new marker format — `reapply_loom_gitignore_patterns` derives loom lines by diffing the post-init snapshot against HEAD and appends append-only/idempotently; the marker lines are just carried along. Scenarios 1–5 all green.

### Verification (run locally)
- `cargo test --lib post_init` — **17 passed**
- `cargo clippy --lib` — clean for changed code; the one warning is the pre-existing `forge_parser.rs:91` pedantic `uninlined_format_args`, unrelated to this PR
- `bash -n` on `uninstall-loom.sh`, `install.sh`, and the stash test — OK
- `scripts/test-install-quick-gitignore-stash.sh` — **16 passed** (incl. Scenario 6 byte-identity + full block removal)
- `scripts/test-installer.sh` — **111 passed**
- CI: Installer Integration, Rust Lint/Build, Rust Unit all green; mergeStateStatus CLEAN.

### Minor, non-blocking
- One-time migration of a legacy markerless block that sits mid-file with surrounding blank lines can leave a double-blank artifact and move the block to EOF. Not claimed byte-stable, and it's idempotent afterward — acceptable.
- An orphaned single marker (BEGIN without END) routes to the legacy append path but self-heals on the next run. Pathological edge.

Single source of truth via `EPHEMERAL_PATTERNS`, marker convention mirrors CLAUDE.md, thorough tests including a proven regression. Approving.

PR #3594, opened 2026-07-16
fix(init): byte-stable legacy .gitignore migration + orphan-marker normalization (#3592)

### LINKED ISSUE #3592: update_gitignore: legacy markerless-block migration isn't byte-stable (double-blank + relocates to EOF)
### Summary

Follow-up to #3590 (fixed by #3591). #3591 made the `.gitignore` install/uninstall round-trip byte-idempotent by wrapping the Loom runtime patterns in `# >>> loom-managed (do not edit) >>>` / `# <<< loom-managed <<<` sentinels, refreshed in place. The Judge on #3591 verified this holds for **already-marked** blocks and **fresh** installs — but flagged that the **one-time migration of a legacy markerless block is not byte-stable**. This issue closes that last gap.

> **Curator note (source verified against `origin/main`, the build base):** the mid-file migration defect is confirmed exactly as described. The **secondary orphan-marker note is materially understated** — see "Correction" below. All line references are to `loom-daemon/src/init/post_init.rs` on `origin/main`.

### Verified mechanism — legacy markerless migration (confirmed)

`update_gitignore` (`post_init.rs:182`) builds the managed block via `managed_gitignore_block()` (`:127-139`: BEGIN marker `:73` + header `:81` + `EPHEMERAL_PATTERNS` `:92-123` + END marker `:76`), then:

1. Splits the file on `\n` into a `Vec<String>` (`:199`).
2. Strips legacy over-broad patterns (`:204-205`).
3. Locates the BEGIN/END markers (`:207-210`) and `match (begin, end)` (`:214`):
   - **`(Some(b), Some(e)) if b <= e`** (`:215-219`) — marked block present: `lines.splice(b..=e, block)` refreshes **in place**. Byte-stable. Correct.
   - **`_` arm** (`:220-241`) — the legacy/no-marker path:
     - `lines.retain(...)` drops the bare header + any bare `EPHEMERAL_PATTERNS` lines **wherever they sit** (`:224-227`).
     - pops one trailing empty element (`:233-235`), then **pushes the marked block and a trailing empty at EOF** (`:236-240`).

Because the `_` arm strips the old block **wherever it was** and re-appends the new block **at EOF**, a legacy block sitting **mid-file between blank lines** produces two defects, empirically reproduced:

Input:
```
node_modules/
<blank>
# Loom runtime state (don't commit these)
.loom-in-use … .loom/logs/
<blank>
dist/
```
After the first post-upgrade `init`:
```
node_modules/
<blank>
<blank>          <-- (1) DOUBLE-BLANK ARTIFACT: the blank before the old header and
dist/                the blank after the last pattern collapse adjacent when the block is removed
# >>> loom-managed (do not edit) >>>
… managed block …   <-- (2) RELOCATED TO EOF instead of staying between node_modules/ and dist/
# <<< loom-managed <<<
```

This is exactly the block-movement / spacing churn #3590 set out to kill. It **does converge** — a second `init` finds the markers and splices in place, byte-identical no-op — but the single upgrade-path `init` moves the block and leaves a permanent double-blank in unrelated user content. The existing test `migrates_legacy_markerless_block_in_place` (`:659-685`) only covers a legacy block **at EOF with no trailing content**, so it never exercises this; its idempotency assertion passes precisely because nothing moves.

### Correction — orphan single marker does NOT simply self-heal

The original note said an orphaned single marker "routes to the legacy append path and self-heals on the next run — acceptable." Reproduced against the real logic, this is wrong for one case and lossy for the other:

- **Orphan END only** (stray `# <<< loom-managed <<<` above content): `position()` returns the earliest END (`:210`), so on every subsequent run `begin > end`, the `b <= e` guard (`:215`) fails, and control re-enters the `_` arm — which re-appends another block at EOF while `retain` (`:224-227`) never removes marker lines. **Result: non-converging, unbounded growth** — BEGIN/END marker pairs accumulate +1 per `init`, forever. Verified: 4 runs → 4 stray blocks and counting.
- **Orphan BEGIN only** (stray `# >>> loom-managed (do not edit) >>>` above content): run 1 appends a full block at EOF; run 2 now has `begin < end`, so `splice(b..=e, …)` (`:218`) replaces everything from the orphan BEGIN through the real END — **swallowing any user lines in between** (verified: a `dist/` line between the orphan marker and EOF is silently deleted). It "converges" structurally but eats intervening user content.

Both require a hand-edited / corrupted marker to trigger (very low probability), but they are correctness bugs — not benign self-healing — so the fix must normalize orphan markers rather than route them through the append path.

### Recommended fix (single code path: the `_` arm of `update_gitignore`, `:220-241`)

Change the no-marker branch to **replace in place** and to **normalize orphan markers first**:

1. **Strip orphan markers up front.** Before the `match`, or at the top of the `_` arm, `retain` out any stray `GITIGNORE_BEGIN_MARKER` / `GITIGNORE_END_MARKER` lines that are not part of a well-formed `begin <= end` pair. This kills the END-only accumulation and the BEGIN-only content-swallow at the source.
2. **Locate the legacy block and splice in place.** After removing the bare header + contiguous bare `EPHEMERAL_PATTERNS` run, remember the index where the header/pattern run started and `splice` the marked block into that same span, rather than `push`-ing at EOF (`:236-240`). Only fall back to create-at-EOF when the file genuinely contains no Loom header/patterns/markers at all.
3. **Do not leave a double blank.** When the removed legacy block was flanked by blank lines on both sides, collapse to a single blank (or splice such that exactly one blank remains on each side). The EOF-append fast path stays byte-identical to uninstall's marker-span deletion (preserving the `roundtrip_is_byte_identical` guarantee, `:620-657`).

Keep the in-place splice arm (`:215-219`) and the fresh-file create path (`:187-191`) unchanged — both are already byte-stable.

### Acceptance criteria (testable against the `post_init` unit suite)

- [ ] **Mid-file legacy migration is byte-stable.** New test: a legacy markerless block (`# Loom runtime state (don't commit these)` + bare patterns) sitting **mid-file with a blank line on each side and user content after it** migrates such that (a) exactly one BEGIN and one END marker exist, (b) the block stays **in place** (user content that followed the legacy block still follows the managed block — mirror the position assertion in `in_place_update_does_not_move_or_duplicate_block`, `:589-617`), (c) **no double-blank artifact** (`assert!(!contents.contains("\n\n\n"))`), and (d) a second `update_gitignore` is a **byte-identical no-op** (`assert_eq!(contents, again)`).
- [ ] **Orphan END only converges without growth.** New test: input with a stray `# <<< loom-managed <<<` above content yields exactly one BEGIN + one END after one run, and a second run is byte-identical (regression against the reproduced unbounded-growth bug).
- [ ] **Orphan BEGIN only converges without eating user content.** New test: input with a stray `# >>> loom-managed (do not edit) >>>` above a user line (e.g. `dist/`) converges to one well-formed block **and the user line survives** (`assert!(contents.contains("dist/"))`).
- [ ] Existing tests still pass unchanged: `migrates_legacy_markerless_block_in_place` (`:659-685`), `in_place_update_does_not_move_or_duplicate_block` (`:589-617`), `roundtrip_is_byte_identical` (`:620-657`), `does_not_duplicate_patterns_on_repeated_runs` (`:317-329`).
- [ ] Each new test **fails against current `origin/main`** and passes after the fix (demonstrate the pre-change failure, as #3591's PR did).

### Severity

Low, upgrade-path / hand-edit only, and — for the primary mid-file case — self-converging. **Not data-loss** for the normal migration path (#3589's stash-pop handler still surfaces any conflict with a recovery path and exits 0; the block converges byte-stable after one run). The orphan-marker edge cases are worse than "self-heal" (END-only never converges; BEGIN-only can delete user lines between a stray marker and EOF), but both require a corrupted/hand-edited marker to trigger, so overall priority stays **low**. This closes the last gap in #3590's byte-idempotency guarantee.

### Affected Files

- `loom-daemon/src/init/post_init.rs` — `update_gitignore` (`:182`), specifically the legacy/no-marker `_` arm of the `match (begin, end)` (`:220-241`); reuse `managed_gitignore_block()` (`:127-139`) and the markers/const (`:73`, `:76`, `:81`, `:92-123`).
- `loom-daemon/src/init/post_init.rs` tests (`#[cfg(test)] mod tests`, `:253-686`) — extend with the mid-file-migration and orphan-marker regression tests above; the `remove_managed_block` helper (`:552-569`) is available for round-trip assertions.

### Related

- Parent: #3590 (fix PR #3591)
- Grandparent: #3588 (fix PR #3589)

rjwalters:
Closes #3592

## Summary
Fixes the last gap in #3590's `.gitignore` byte-idempotency guarantee: the one-time migration of a legacy markerless Loom block. All changes are in the `_` (legacy/no-marker) arm of `update_gitignore`'s `match (begin, end)` in `loom-daemon/src/init/post_init.rs`.

Three defects fixed:
1. **Mid-file migration not byte-stable** — a legacy block flanked by blank lines left a double-blank artifact where it was removed and relocated the block to EOF (converged only on the 2nd run).
2. **Orphan END-only marker → unbounded growth** — `begin > end` forever, so the arm re-appended a fresh block every run.
3. **Orphan BEGIN-only marker → user-content loss** — the converging in-place splice swallowed user lines between the stray BEGIN and EOF.

### Fix
- **Normalize orphan/misordered markers first**: strip any stray marker line (reaching this arm implies no well-formed `BEGIN <= END` pair).
- **Splice the marked block into the legacy block's original span in place** instead of pushing at EOF — flanking blanks stay put, so no relocation and no double-blank.
- **Create-at-EOF fallback** only when the file has no Loom header/patterns/markers at all, preserving the #3590 byte-identical install→uninstall→install round-trip.

Property guaranteed: any legacy/orphan state converges to a correctly-marked block in ONE `init`; a second `init` is a byte-identical no-op — no line relocation, no double-blank, no user-content loss.

## How verified
- Added 3 regression tests to `mod tests`: `migrates_mid_file_legacy_block_in_place_without_double_blank`, `orphan_end_marker_converges_without_growth`, `orphan_begin_marker_converges_without_eating_user_content`. Each was confirmed to **FAIL against the pre-fix `_` arm** and **PASS after the fix**.
- `cargo test --lib post_init` → **20 passed, 0 failed** (existing 17 incl. the original 4 targeted + 3 new).
- `cargo build --release` → clean.
- `cargo clippy --lib` → no warnings in `post_init.rs` (the one pedantic warning is pre-existing and in an unrelated file).
- Shell suites: `.loom/scripts/tests/test-gitignore-guard.sh` (18/18) and `scripts/test-install-quick-gitignore-stash.sh` (16/16, incl. the #3590 round-trip assertions) both pass.

### DIFF
--- a/loom-daemon/src/init/post_init.rs
+++ b/loom-daemon/src/init/post_init.rs
@@ -218,26 +218,60 @@ pub fn update_gitignore(workspace_path: &Path) -> Result<(), String> {
             lines.splice(b..=e, block_lines.iter().cloned());
         }
         _ => {
-            // No marked block. Migrate any legacy markerless block by dropping
-            // the old header and any bare ephemeral pattern lines, then append
-            // the marked block at EOF (the one-time create-at-EOF path).
+            // No well-formed BEGIN <= END pair. This covers a legacy markerless
+            // block, a fresh consumer file, and corrupted single/misordered
+            // markers. Migrate in place rather than relocating to EOF (#3592).
+
+            // 1. Normalize orphan/corrupted markers. Because we did not match the
+            //    in-place arm, any marker line present here is stray (an END-only
+            //    orphan, a BEGIN-only orphan, or an END-before-BEGIN pair). Drop
+            //    them so they can neither accumulate every run (orphan-END
+            //    unbounded growth) nor drive a destructive splice that swallows
+            //    user content (orphan-BEGIN). #3592
             lines.retain(|line| {
                 let t = line.trim();
-                t != GITIGNORE_BLOCK_HEADER && !EPHEMERAL_PATTERNS.contains(&t)
+                t != GITIGNORE_BEGIN_MARKER && t != GITIGNORE_END_MARKER
             });
-            // Drop exactly one trailing empty element (the file's final '\n')
-            // so the block is appended directly after the existing content with
-            // no spurious blank line. This makes the append the exact inverse of
-            // uninstall's marker-span deletion, so an install -> uninstall ->
-            // install round-trip is byte-identical (issue #3590).
-            if lines.last().is_some_and(String::is_empty) {
-                lines.pop();
+
+            // 2. Locate the legacy markerless block: the bare header and/or bare
+            //    ephemeral pattern lines. Remember where the first such line sits
+            //    so the marked block can be spliced back into that same span.
+            let is_legacy = |line: &str| {
+                let t = line.trim();
+                t == GITIGNORE_BLOCK_HEADER || EPHEMERAL_PATTERNS.contains(&t)
+            };
+            let first_legacy = lines.iter().position(|l| is_legacy(l));
+
+            match first_legacy {
+                Some(start) => {
+                    // Remove every legacy line, then splice the marked block into
+                    // the original position. Because we replace in place instead
+                    // of removing-then-appending, flanking blank lines stay put:
+                    // the block does not relocate to EOF and no double-blank
+                    // artifact is left behind where the old block used to sit
+                    // (#3592). Any user content that followed the legacy block
+                    // still follows the managed block.
+                    lines.retain(|l| !is_legacy(l));
+                    lines.splice(start..start, block_lines.iter().cloned());
+                }
+                None => {
+                    // Genuinely no Loom header/patterns/markers: create the block
+                    // at EOF. Drop exactly one trailing empty element (the file's
+                    // final '\n') so the block is appended directly after the
+                    // existing content with no spurious blank line. This makes the
+                    // append the exact inverse of uninstall's marker-span
+                    // deletion, so an install -> uninstall -> install round-trip
+                    // is byte-identical (issue #3590).
+                    if lines.last().is_some_and(String::is_empty) {
+                        lines.pop();
+                    }
+                    for bl in &block_lines {
+                        lines.push(bl.clone());
+                    }
+                    // Re-add exactly one trailing empty element => single '\n'.
+                    lines.push(String::new());
+                }
             }
-            for bl in &block_lines {
-                lines.push(bl.clone());
-            }
-            // Re-add exactly one trailing empty element => single trailing '\n'.
-            lines.push(String::new());
         }
     }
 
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM — approved.

## Review summary

Reviewed the rewritten `_` arm of `update_gitignore` in `loom-daemon/src/init/post_init.rs` and the 3 new regression tests. Verified against the actual PR head (5622b7e2, built on #3591 / b938638c).

### Correctness verified
- **Scope confined to the `_` arm**: the diff touches only the legacy/no-marker arm plus 3 new tests. The already-marked in-place arm (`(Some(b), Some(e)) if b <= e`) and the fresh-file create path are untouched, so the #3590 install→uninstall→install byte-identical round-trip is preserved. The `None` (create-at-EOF) sub-branch retains the same pop/push/push logic as before.
- **Mid-file in-place splice**: markers are stripped first, then the marked block is spliced at `start..start` where `first_legacy` sat. Blank lines are never `is_legacy`, so flanking blanks are preserved — no relocation to EOF, no double-blank for a contiguous legacy block.
- **Orphan-END**: stray markers are removed up front before the legacy scan, eliminating the `begin > end` re-append loop → converges in one run, no unbounded growth.
- **Orphan-BEGIN**: same up-front marker strip means the converging path no longer splices away user lines between the stray BEGIN and EOF.
- **Idempotency**: after migration the file has a well-formed marked block; the 2nd run hits the in-place refresh arm with identical `block_lines`, producing a byte-identical no-op (confirmed by the tests' `assert_eq!(after_one, after_two)`).
- **No panic/index hazards**: `splice(start..start, …)` is safe because after `retain` the elements before `start` are all non-legacy and preserved, so `lines.len() >= start`. Empty-vec handled by the `None` arm. `EPHEMERAL_PATTERNS.contains(&t)` / `legacy_overbroad.contains(&t)` are exact `&str` equality on `line.trim()`, not substring matches.

### Tests
Empirically confirmed the 3 new tests are genuine regressions: grafted them onto the pre-fix `_` arm and all 3 FAILED (orphan-END grows to 2 markers; orphan-BEGIN 2nd run eats `dist/`; mid-file relocates block to EOF with a double-blank so `dist/` lands before the block). Against the fix, `cargo test --lib post_init` → **20 passed, 0 failed**. Assertions are byte-level (marker counts, ordering, `!contains("\n\n\n")`, byte-identical no-op) rather than mere `contains`.

### Local verification
- `cargo test --lib post_init` → 20 passed.
- `cargo clippy --lib` → the only warning is the pre-existing `uninlined_format_args` on a `url_pattern` regex in an unrelated file, not `post_init.rs`.
- `test-gitignore-guard.sh` → 18/18; `test-install-quick-gitignore-stash.sh` → 16/16 (incl. #3590 round-trip).
- CI: Rust Linting & Build + Rust Unit Tests pass; merge state CLEAN/MERGEABLE.

### Minor note (non-blocking)
A hand-corrupted legacy block with blank lines *interspersed* between the header and pattern lines (not something the installer ever emits — real legacy blocks are contiguous) can leave a double-blank after the migrated block, since `retain` collapses the now-adjacent blanks and the splice reinserts at the first-legacy index. It still converges idempotently with no growth or data loss, so this is a cosmetic edge case outside the stated scope — worth a future follow-up at most.

Issue #3609, opened 2026-07-17
skill-router.sh: every-prompt agent-table injection + over-broad route patterns = high token cost, near-zero suggestion precision


rjwalters:
## skill-router.sh injects on every prompt with over-broad patterns — high token cost, ~zero suggestion precision in orchestrator sessions

Observed across a long orchestration session in a Loom-installed repo (kicad-tools, Loom 0.10.7). Of the three installed hooks, `guard-destructive.sh` (silent unless triggered) and `methodology-inject.sh` (opt-in dir absent → no-op) behave well. **`skill-router.sh` is too aggressive** on three axes:

### 1. Always-inject, no dedup — the agent table lands on EVERY prompt
The hook unconditionally emits the full "Available Loom agents" table (~11 lines, ~120–150 tokens) as `additionalContext` on every `UserPromptSubmit` — including one-word continuations ("continue", "proceed", "yes"), slash-command local outputs, and harness-generated task-notification turns. In a ~50-turn session that is **several thousand tokens of verbatim repetition**, all duplicating documentation that already ships statically in `.loom/CLAUDE.md` (the roles table) and the command listing.

Notably the hook already **receives `session_id` on stdin but never uses it** — a once-per-session marker (e.g. `.loom/logs/skill-router-seen/<session_id>`) would eliminate ~98% of the repetition at zero capability loss.

### 2. Patterns are so broad the AGENT_ROUTE suggestions are noise
Default `skill-routes.json` patterns like:
- `build|implement|feature|develop|code` → `/loom:builder`
- `fix|bug|broken|...` → `/loom:doctor`
- `curate|triage|issue|enhance|enrich` → `/loom:curator`
- `review|\bPR\b|pull request|...` → `/loom:judge`

match nearly any sentence about software. Measured anecdotally over one real session: **precision ≈ 0%** — e.g. *"let's confirm our builders are still active and not stalled"* → routed `/loom:builder` ("Implement features and fixes"); *"we're hitting hooks a lot; open an issue with rjwalters/loom"* → routed `/loom:curator`. In an orchestrator session (the main Claude spawning loom-role subagents itself), the suggestion is structurally never actionable, yet it fires on almost every turn.

### 3. No suppression for control prompts / non-task turns
No minimum prompt length, no skip for prompts that are pure continuations, no skip for locally-generated command output turns. (There IS a `/self*` skip — the right idea, too narrow.)

### Suggested fixes (independent; any subset helps)
[...]

PR #3615, opened 2026-07-17
fix(hooks): skip machine-generated turns; anchor methodology topic fallback (#3608)

### LINKED ISSUE #3608: hooks: UserPromptSubmit hooks (skill-router, methodology-inject) fire too aggressively on machine-generated prompts and loose topic patterns
## Summary

During a long `/loom:sweep all` orchestration session in a consumer repo (vibesql), the two default UserPromptSubmit hooks injected context on effectively **every turn**, most of it noise. Two distinct root causes:

### 1. `methodology-inject.sh`: filename-fallback topic pattern is a bare substring

When a topic file has no sidecar `.pattern`, the hook falls back to using the **filename as the regex** and substring-matches it (case-insensitive) against the entire prompt:

```bash
# Fall back to using filename as the regex pattern
if [[ -z "$PATTERN" ]]; then
    PATTERN="$TOPIC_NAME"
```

With `topics/release.md` present, any prompt containing `cargo build --release` or `target/release` — i.e. **virtually every Rust builder/agent status message** — triggers injection of the full ~8KB release-conventions topic. We observed it firing twice on background-task completion notifications that had nothing to do with releasing. The script's own header documents the intent as "injected … whenever `/loom:release` is invoked", but the implementation is dramatically looser.

**Suggested fixes** (any of):
- Require a sidecar `.pattern` file for topic matching; without one, match only the slash-command form (`/loom:<topic>`, `/repo:<topic>`) or an exact word-boundary token, never a bare substring.
- At minimum, wrap the filename fallback in word boundaries AND exclude matches inside flag-like tokens (`--release`) and path segments (`target/release`).

### 2. Both hooks run on machine-generated prompts (task notifications), not just human input

Both scripts already special-case orchestrator pulses:

```bash
# Skip orchestrator pulse prompts (start with /self)
if [[ "$PROMPT" == /self* ]]; then exit 0; fi
```

But in an orchestration session, the dominant prompt type is the harness's background-task completion notification (prefixed `[SYSTEM NOTIFICATION - NOT USER INPUT]` / containing `<task-notification>`). Every builder/judge/doctor completion re-ran both hooks, so:

- `skill-router.sh` re-injected the full agent routing table ~20 times in one session, and
- its `AGENT_ROUTE` suggestion was **wrong nearly every time**, because the route patterns are generic enough (`build|implement|feature|develop|code`, `curate|triage|issue|enhance|enrich`, `review|\bPR\b|pull request`) that any Loom-flavored notification text matches something. Examples observed: a builder-completion notification routed to `/loom:judge`; a question about hook aggressiveness routed to `/loom:curator` (matched "issue" in "open an issue").

**Suggested fixes**:
- Skip prompts that begin with `[SYSTEM NOTIFICATION` or contain `<task-notification>` (same rationale as the existing `/self` skip — they are not human input and the routing table/topic context is already in session context from the first injection).
- Consider session-level dedupe for the always-injected routing table (inject once per session, or only when the prompt is a plausible task statement).
- Consider word-boundary-anchoring the route patterns and/or requiring 2+ pattern hits before emitting an `AGENT_ROUTE` directive.

## Impact

In a wave-based sweep session (11 issues, ~20 subagent completions), the hooks added tens of KB of repeated/irrelevant injected context and a stream of misleading routing suggestions. None of it broke anything — but it burns context window on every turn of exactly the long-running sessions where context is most precious.

## Local mitigation applied (vibesql)

Added a sidecar `.loom/context/topics/release.pattern` with an anchored pattern (`/loom:release|cut (a|the) release|…`) so the release topic no longer matches `--release` build flags. The systemic fixes belong here in the defaults.

rjwalters:
Closes #3608

## Summary

UserPromptSubmit hooks (`skill-router.sh`, `methodology-inject.sh`) fired too aggressively on harness-generated prompts and on loose topic-name substrings. This PR addresses the two root causes scoped to #3608.

## Changes

1. **Shared machine-generated-prompt skip guard (both hooks)** — immediately after the existing `/self*` skip, both hooks now exit 0 on harness task-notification turns: prompts starting with `[SYSTEM NOTIFICATION` or containing `<task-notification>`. Matched against the raw `$PROMPT` with literal prefix/substring (no regex), consistent with the existing `/self*` guard, so the guard cannot itself false-positive on human text.

2. **Anchored topic-filename fallback (`methodology-inject.sh`)** — the bare-substring filename fallback caused `cargo build --release` / `target/release` to inject a `release` topic. The fallback pattern is now anchored: it requires the slash-command form (`/loom:<topic>` or `/repo:<topic>`) or a word-boundary token not preceded by `-`/`/` and not followed by `/`. The sidecar `.pattern` escape hatch is unchanged.

3. **`.loom/hooks/` sync** — the dogfooded copies are re-synced from `defaults/`, which also reconciles the pre-existing `/shepherd`//`/loom` role-case drift in the `.loom` methodology-inject copy toward the defaults (`/loom:sweep`) version.

4. **Docs** — `defaults/CLAUDE.md` "Methodology Injection Framework" now describes the anchored fallback behavior.

## Scope coordination with #3609 (READ)

**The only region of `skill-router.sh` touched here is the top-of-file input guard** (right after the `/self*` skip, before the `ROUTING CONFIG` section). The route-table build, pattern-matching loop, dedup, and output sections are **untouched** — those belong to #3609. No changes to `skill-routes.json`. The two PRs edit disjoint regions of `skill-router.sh`.

## Testing (isolated git-repo harness)

- Prompt starting `[SYSTEM NOTIFICATION` → both hooks emit nothing, exit 0.
- Prompt containing `<task-notification>` → both hooks emit nothing, exit 0.
- `cargo build --release` / `see target/release/foo` (topic `release`, no sidecar) → release topic NOT injected.
- `/loom:release`, `/repo:release`, `lets cut a release` → release topic injected.
- Sidecar `release.pattern` still overrides the filename fallback exactly as before.
- Human prompts still inject `universal.md` and the skill-router agent table / AGENT_ROUTE.
- All four hook files pass `bash -n`; every invocation (empty prompt, notification, normal) exits 0.

### DIFF
--- a/.loom/hooks/methodology-inject.sh
+++ b/.loom/hooks/methodology-inject.sh
@@ -62,6 +62,18 @@ if [[ "$PROMPT" == /self* ]]; then
     exit 0
 fi
 
+# Skip harness-generated task-notification turns. These are not human input —
+# the harness re-runs UserPromptSubmit hooks on every background-task completion,
+# and the relevant context was already injected on the originating human turn.
+# Match against the raw prompt with literal prefix/substring (no regex) so this
+# guard cannot itself false-positive on human text.
+case "$PROMPT" in
+    "[SYSTEM NOTIFICATION"*) exit 0 ;;
+esac
+if [[ "$PROMPT" == *"<task-notification>"* ]]; then
+    exit 0
+fi
+
 # =============================================================================
 # OPT-IN CHECK
 # =============================================================================
@@ -157,8 +169,7 @@ if [[ "$INJECT_ROLE" == "true" ]]; then
             /champion*) ROLE="champion" ;;
             /guide*)   ROLE="guide" ;;
             /auditor*) ROLE="auditor" ;;
-            /shepherd*) ROLE="shepherd" ;;
-            /loom*)    ROLE="loom" ;;
+            /loom:sweep*) ROLE="sweep" ;;
         esac
     fi
 
@@ -196,9 +207,16 @@ if [[ "$INJECT_TOPICS" == "true" ]] && [[ -d "${CONTEXT_DIR}/topics" ]]; then
             PATTERN=$(cat "$PATTERN_FILE" 2>/dev/null) || PATTERN=""
         fi
 
-        # Fall back to using filename as the regex pattern
+        # Fall back to an anchored match on the filename. A bare substring
+        # match (the historical behavior) false-positives on flag-like and
+        # path-segment contexts — e.g. a "release" topic would inject on
+        # "cargo build --release" or "target/release". Require either the
+        # slash-command form (/loom:<topic> or /repo:<topic>) or a
+        # word-boundary token that is NOT preceded by "-" or "/" and NOT
+        # followed by "/". The sidecar .pattern file remains the escape hatch
+        # for topics that need a custom regex.
         if [[ -z "$PATTERN" ]]; then
-            PATTERN="$TOPIC_NAME"
+            PATTERN="/(loom|repo):${TOPIC_NAME}\b|(^|[^-/[:alnum:]])${TOPIC_NAME}([^/[:alnum:]]|$)"
         fi
 
         # Match pattern case-insensitively against prompt
--- a/.loom/hooks/skill-router.sh
+++ b/.loom/hooks/skill-router.sh
@@ -58,6 +58,18 @@ if [[ "$PROMPT" == /self* ]]; then
     exit 0
 fi
 
+# Skip harness-generated task-notification turns. These are not human input —
+# the harness re-runs UserPromptSubmit hooks on every background-task completion,
+# and re-injecting the agent table / re-routing on them is pure noise.
+# Match against the raw prompt with literal prefix/substring (no regex) so this
+# guard cannot itself false-positive on human text.
+case "$PROMPT" in
+    "[SYSTEM NOTIFICATION"*) exit 0 ;;
+esac
+if [[ "$PROMPT" == *"<task-notification>"* ]]; then
+    exit 0
+fi
+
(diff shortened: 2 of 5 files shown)


rjwalters:
LGTM — approved.

Verified against #3608 acceptance criteria:

- **Skip guard**: literal `case` prefix (`[SYSTEM NOTIFICATION`) + `[[ == *<task-notification>* ]]` substring — no regex, so the guard cannot self-false-positive on human text. Confirmed both hooks emit 0 bytes and exit 0 on notification turns.
- **No non-zero exits**: all invocations (notification, `<task-notification>`, normal, empty) return rc=0 on all four hook files.
- **Anchored fallback regex** (`/(loom|repo):<topic>\b|(^|[^-/[:alnum:]])<topic>([^/[:alnum:]]|$)`): tested with topic `release` — no-match on `cargo build --release`, `target/release`, `prerelease notes`, `foo-release-bar`; match on `/loom:release`, `/repo:release`, `lets cut a release`, `release the build`. No false negatives or false positives.
- **defaults ↔ .loom sync**: `methodology-inject.sh` and `skill-router.sh` are byte-identical between `defaults/hooks/` and `.loom/hooks/` on the PR branch.
- **Scope boundary respected**: `skill-router.sh` only gains the top-of-file input guard; route-table/pattern-loop/dedup sections untouched (left for #3609).
- **`/loom:sweep` role-case change** matches what `origin/main` defaults already carry — no regression.
- `bash -n` clean on all four scripts. CI: Installer Integration Tests + Detect Changes pass (Rust/MCP correctly skipped). mergeStateStatus CLEAN.

PR #3616, opened 2026-07-17
fix(hooks): gate skill-router table on match + dedup per session (#3609)

### LINKED ISSUE #3609: skill-router.sh: every-prompt agent-table injection + over-broad route patterns = high token cost, near-zero suggestion precision

rjwalters:
Closes #3609

## Summary

`skill-router.sh` injected the full agent table on **every** prompt (~120-150 tokens/turn, ~6-7.5k tokens over a ~50-turn session) and matched routes on bare single-word substrings, so "builders" mis-routed to `/loom:builder`, and a bare "issue" / any "loom" mention mis-routed too — near-zero suggestion precision at real token cost.

## Changes

- **Table only on a match.** When no route matches, the hook now emits **no** `additionalContext` and exits silently. The agent table only rides along with a genuine route match.
- **Per-session dedup.** The table is deduped using the `session_id` already present on stdin (marker under `.loom/logs/skill-router-seen/<session_id>`, which is already gitignored). A session receives the table at most once; the `AGENT_ROUTE:` line is still emitted on every match. Missing/empty `session_id` degrades gracefully to per-match inclusion.
- **Minimum-signal gate.** Skip prompts that begin with `/` (slash-command turns) and prompts under 3 words.
- **Tightened patterns** to phrase/word-boundary anchors in `skill-routes.json`. The two report example prompts ("confirm our builders are still active…", "open an issue with rjwalters/loom") no longer route; genuine imperatives ("review this PR", "file an issue for X") still do.
- **Dropped the dead `shepherd` route** (command removed in v0.10.0) and re-synced `.loom/config/skill-routes.json` from `defaults/` (namespaced `/loom:<role>`; the deployed copy had drifted to un-namespaced agents + a dead `/shepherd`).
- **`.loom/hooks/skill-router.sh` kept byte-identical** to `defaults/hooks/skill-router.sh`.

## Scope boundary

The #3608 top-of-file machine-generated-prompt guard (`/self*`, `[SYSTEM NOTIFICATION*`, `<task-notification>`) is unchanged. `methodology-inject.sh` is untouched.

## Test

New `defaults/hooks/tests/test-skill-router.sh` — 26 cases covering match/no-match, short/slash skips, per-session dedup (table once, then deduped; new session refreshed; graceful no-session), the two report mis-routes, the never-non-zero / always-valid-JSON contract, and `defaults/` ↔ `.loom/` sync for both hook and config. All 26 pass.

### DIFF
--- a/.loom/config/skill-routes.json
+++ b/.loom/config/skill-routes.json
@@ -1,55 +1,50 @@
 {
   "version": 1,
-  "description": "Default skill routing table for Loom agent suggestions. Routes are evaluated top-to-bottom, first match wins. Place specific patterns before general ones.",
+  "description": "Default skill routing table for Loom agent suggestions. Routes are evaluated top-to-bottom, first match wins. Place specific patterns before general ones. Patterns are phrase/word-boundary anchored (issue #3609) so common software vocabulary ('builders', a bare 'issue', a repo path like 'rjwalters/loom') no longer mis-routes — only genuine imperatives do. Agent invocations use the namespaced `/loom:<role>` form required by Claude Code 2.1+ (subdirectory commands at `.claude/commands/loom/<role>.md` resolve via `namespace:command` syntax — see issue #3345).",
   "routes": [
     {
-      "pattern": "shepherd|orchestrate|lifecycle|end.to.end",
-      "agent": "/shepherd",
-      "description": "Full issue lifecycle orchestration"
-    },
-    {
-      "pattern": "architect|design|proposal|system design|rfc",
-      "agent": "/architect",
+      "pattern": "\\barchitect|system design|design (a |an |the )?proposal|\\brfc\\b",
+      "agent": "/loom:architect",
       "description": "System design and architecture"
     },
     {
-      "pattern": "review|\\bPR\\b|pull request|code review",
-      "agent": "/judge",
+      "pattern": "code review|\\bpull request\\b|review (this|the|my|our|that) ",
+      "agent": "/loom:judge",
       "description": "Code review"
     },
     {
-      "pattern": "fix|bug|broken|changes.requested|merge.conflict",
-      "agent": "/doctor",
+      "pattern": "\\bfix\\b|\\bbug\\b|\\bbroken\\b|changes.requested|merge.conflict|address (the )?feedback",
+      "agent": "/loom:doctor",
       "description": "Bug fixes and PR feedback"
     },
     {
-      "pattern": "simplify|refactor|clean.?up|dead.?code|complexity",
-      "agent": "/hermit",
+      "pattern": "\\bsimplify\\b|\\bsimplification\\b|\\brefactor|clean.?up|dead.?code|\\bcomplexity\\b",
+      "agent": "/loom:hermit",
       "description": "Simplification opportunities"
     },
     {
-      "pattern": "build|implement|feature|develop|code",
-      "agent": "/builder",
+      "pattern": "\\bbuild\\b|\\bimplement|\\bfeature\\b|\\bdevelop",
+      "agent": "/loom:builder",
       "description": "Implement features and fixes"
     },
     {
-      "pattern": "curate|triage|issue|enhance|enrich",
-      "agent": "/curator",
+      "pattern": "\\bcurate\\b|\\benrich|\\btriage\\b|file (an |a )?(issue|bug)",
+      "agent": "/loom:curator",
       "description": "Issue enrichment"
     },
     {
-      "pattern": "prioriti[zs]e|backlog|roadmap",
-      "agent": "/guide",
+      "pattern": "prioriti[zs]e|\\bbacklog\\b|\\broadmap\\b",
+      "agent": "/loom:guide",
       "description": "Backlog triage"
     },
     {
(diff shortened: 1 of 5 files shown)


rjwalters:
LGTM! Approved.

Reviewed against the #3609 acceptance criteria and all check out:

- **Gate on match:** Non-matching prompts hit the `[[ -z "$MATCHED_AGENT" ]] && exit 0` early return — no `additionalContext` emitted. `AGENT_ROUTE:` is still built on every match.
- **Per-session dedup:** Marker file keyed on a `tr`-sanitized `session_id` under `.loom/logs/skill-router-seen/` (gitignored). First match includes the table, subsequent matches in the same session drop it, a new session refreshes it. Missing/empty `session_id` leaves `INCLUDE_TABLE=1` so it degrades to per-match inclusion. Marker creation is best-effort (`|| true`) and never fails the hook.
- **Minimum-signal gate:** Leading-`/` and `<3`-word skips are sound; `WORD_COUNT` is defensively defaulted to 0.
- **Pattern precision:** Confirmed the two report mis-routes ('confirm our builders are still active…', 'open an issue with rjwalters/loom') no longer route, while 'review this PR', 'please implement the new feature', and 'file an issue for X' still do. Dropping bare `code`/`issue`/`loom`/`review` substrings in favor of word-boundary/phrase anchors is the right call.
- **Scope boundaries intact:** #3608 top-of-file guard (`/self*`, `[SYSTEM NOTIFICATION*`, `<task-notification>`) untouched; `methodology-inject.sh` not in the changeset (only the 5 expected files).
- **Sync:** `.loom/hooks/skill-router.sh` and `.loom/config/skill-routes.json` are byte-identical to `defaults/` (verified via `diff`); dead `shepherd` route removed and config re-namespaced to `/loom:<role>`.
- **Never-fail contract:** `set -o pipefail` + ERR trap + guarded exits preserved; the never-non-zero / always-valid-JSON probes pass.
- **Tests:** Ran `defaults/hooks/tests/test-skill-router.sh` locally — **26/26 pass**. CI is green (Installer Integration Tests pass; Rust/MCP jobs correctly skipped for a hooks-only change). Merge state CLEAN.

Issue #3598, opened 2026-07-17
install.sh rewrites .loom/config.json and drops consumer keys (worktree.root override lost)


rjwalters:
Observed on 0.10.9 (`install.sh --quick -y` into geode-fem).

## Repro

Consumer `.loom/config.json` (tracked in git) carries a documented override:

```json
"worktree": {
  "root": "/Volumes/Stripe"
}
```

(per `.loom/scripts/lib/worktree-root.sh` precedence: `LOOM_WORKTREE_ROOT` env > `worktree.root` config > default `.loom/worktrees/`).

After the quick reinstall, the installer rewrote `config.json` and the entire `worktree` block was gone (it also reflowed unrelated formatting, e.g. the `reflection.categories` array). Nothing warned about the dropped key.

## Impact

Silent loss of consumer configuration. In this repo the override is load-bearing: worktrees are deliberately relocated to a separate volume (`/Volumes/Stripe/geode-fem/issue-N`); with the key dropped, the next `worktree.sh` run would have silently created worktrees back under `.loom/worktrees/`. Caught only by a manual diff before commit; restored from HEAD.

## Expected

The installer's config handling should merge — preserve unknown/consumer keys (or at minimum the documented `worktree.root` override) — rather than regenerate the file from its own template.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rjwalters:
## Curator Enhancement

### Root Cause Analysis

The key drop is **two-layered** — both layers must be addressed or the fix is incomplete.

**Layer 1: the chained uninstall deletes `.loom/config.json` as a "runtime artifact".**
`install.sh` reinstall flow runs `scripts/uninstall-loom.sh --yes --local` before reinitializing (`install.sh:638`). The uninstaller lists `.loom/config.json` in its `RUNTIME_ARTIFACTS` array (`scripts/uninstall-loom.sh:482-493`, entry at line 485) and removes it from disk. But `config.json` is *not* a runtime artifact — it is committed consumer configuration (the issue's `worktree.root` override was tracked in git). By the time init runs, the consumer's file is already gone, so no init-side logic can see it without help.

**Layer 2: daemon init unconditionally clobbers from the template.**
`install.sh:663` runs `loom-daemon init --force`. `initialize_workspace()` calls:

- `loom-daemon/src/init/mod.rs:161` — `copy_single_file(&defaults, &loom_path, "config.json", ".loom/config.json", &mut report)`
[...]

Issue #3597, opened 2026-07-17
install.sh --quick 'reconcile git index' step reverts uncommitted tracked changes made by sibling installers


rjwalters:
Observed on 0.10.9 while batch-updating tools in a consumer repo (geode-fem) via `/repo:update-tools`.

## Repro

1. In a consumer repo with a clean tree, run installers in sequence without committing between them: `loom/install.sh --quick -y`, then `anvil/scripts/install-anvil.sh`, then `repo/install.sh -y`. Tree now holds all three updates uncommitted.
2. Re-run `loom/install.sh --quick -y` (e.g. to refresh `.loom/manifest.json`).
3. The final "Reconciling git index after reinstall..." step checks tracked modified files back out from HEAD.

## Impact

Every uncommitted in-place modification made by the sibling installers is silently reverted: `.anvil/install-metadata.json` went back to the old version, the renamed-away `anvil:pub` skill files and agent shims were restored from HEAD, `.claude/skills/repo/install-metadata.json` reverted, etc. Brand-new untracked files survive, so the tree is left in a half-old/half-new hybrid state that looks superficially fine. The fix on the consumer side was to re-run the other installers.

## Expected

The reconcile step should restrict itself to paths Loom itself owns (its install manifest), never `git checkout` arbitrary tracked files it didn't write.

## Related: verify-install false drift invited the re-run

What prompted the second `--quick` run: `.loom/scripts/verify-install.sh` reported DRIFT because `.loom/manifest.json` snapshots files Loom doesn't own — the whole-file hash of `CLAUDE.md` (which sibling installers legitimately edit outside Loom's marker block) and `.claude/agents/anvil-*` shims (owned by Anvil, later renamed by it). Scoping the manifest to Loom-owned files (and hashing only Loom's CLAUDE.md marker block) would remove both the false alarm and the temptation to re-run the installer over a dirty multi-tool tree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rjwalters:
[...]

Issue #3545, opened 2026-07-13
install.sh --quick reinstall leaves ~150 staged deletions + stages unrelated user files (uninstall's git add -A, no post-init reconciliation)


rjwalters:
## Summary

A `--quick` **reinstall** over an existing consumer install (Loom files tracked in the consumer repo) leaves the consumer's git index in a confusing and hazardous state:

1. Every previously-tracked Loom file shows as a **staged deletion** (`D`) while the freshly re-installed identical file sits **untracked** (`??`) on disk — ~150 paired entries in `git status`.
2. Worse, **unrelated user files get staged**: the uninstall step's `git add -A` sweeps in anything dirty or untracked in the consumer tree at reinstall time.
3. The installer's printed next step — `git add -A && git commit` — would then **commit those unrelated files**.

Observed on v0.10.6 (`0b9cd8fc`), reinstalling over a tracked v0.10.4 install in `rjwalters/geode-fem` via `./install.sh --quick -y <target>`.

## Root cause

The `--quick` reinstall path in `install.sh` (~line 567) does:

1. `uninstall-loom.sh --yes --local` — whose Step 8 (`scripts/uninstall-loom.sh` ~line 1190) runs a bare **`git add -A`**, staging the deletions of all Loom files **plus anything else** in the working tree.
2. `loom-daemon init` — rewrites the Loom files **to disk only**; the index is never reconciled.
3. Prints "Next steps: `git add -A && git commit`".

So the staged-deletion/untracked pairs are the uninstall's staged state with the reinstall layered on disk over it. Nothing between steps 1 and 2 restores, re-adds, or scopes the index.

Notably, `scripts/install-loom.sh` already knows about this exact hazard — its `--clean` path stashes user changes first with the comment *"The uninstall --local mode runs 'git add -A' which would stage user changes along with uninstall deletions"* (~line 529), and it runs `git restore --staged . && git checkout -- .` cleanup afterwards. The `install.sh --quick` reinstall path has neither guard.

## Concrete near-miss

In our reinstall, a pre-existing untracked `.claude/worktrees/agent-<id>/` directory (a Claude Code agent worktree) was staged as an **add** by the uninstall's `git add -A`. Following the printed `git add -A && git commit` verbatim would have committed an embedded git worktree into the repo. `.gitignore` also briefly showed `MM` (uninstall staged removal of the 4 Loom ignore lines; reinstall restored them on disk) — nets out, but adds to the "did the installer break my repo?" impression.

[...]

PR #3618, opened 2026-07-17
fix(install): preserve staged split on --quick reinstall pop, reconcile install-metadata.json (#3611)

### LINKED ISSUE #3611: install.sh --quick: git-index reconcile clobbers the caller's staged changes and stages deletion of tracked install-metadata.json
Loom 0.10.9 installer (`101d758c`), quick reinstall into a consumer repo (softstart) that had unrelated changes staged.

`install.sh --quick -y <repo>` runs a "Reconciling git index after reinstall... Restoring stashed user changes" step. Two index side effects observed:

1. **Caller's staged files came back unstaged.** A `.gitignore` edit that was staged before the install was still present afterward but no longer staged — the stash/restore round-trip dropped the index state. If the caller stages carefully (partial commits in flight), the installer silently discards that staging.
2. **`.loom/install-metadata.json` ended up as a staged deletion + a new untracked copy.** The file was tracked before the install; afterward `git status` showed `D  .loom/install-metadata.json` (staged) plus `?? .loom/install-metadata.json`. Net effect: committing as-is would untrack the metadata the upgrade detector reads (`install.sh` itself lists it in `verify_install`'s critical files). Had to `git add` it back manually.

Expected: the reconcile step restores the index exactly as found (stash with `--index` / reapply staged state), and installer-owned tracked files end up modified, not deleted-plus-untracked.

rjwalters:
Closes #3611

## Summary

Fixes two defects in the `install.sh --quick` reinstall stash / index-reconcile machinery (issues #3545 / #3597 / #3588 lineage).

### Defect 1 — staged state dropped on restore
`install.sh` restored the reinstall stash with a plain `git stash pop`, which re-applies **every** stashed hunk to the working tree as *unstaged*. A caller who had a `.gitignore` edit **staged** before the reinstall got it back **unstaged**, and any partial staging in flight was silently flattened.

**Fix:** `git stash pop --index`, which reinstates the index tree the stash recorded at push time — staged hunks come back staged, unstaged hunks stay unstaged. The `.gitignore` HEAD-reset dance still provides a clean 3-way base for the index restore; `reapply_loom_gitignore_patterns` continues to append Loom ephemeral patterns to the working tree only (never the staged copy). `--index` is stricter than a plain pop: when it cannot reinstate the index cleanly it fails and falls through to the existing conflict-surfacing branch (no silent degrade to an unstaged pop). The recovery text now notes the staged/unstaged split could not be reproduced.

### Defect 2 — `.loom/install-metadata.json` left as `D` + `??`
The chained uninstall stages the metadata deletion (`REMOVE_FILES` → `git add -A`), then `finalize_quick_install` rewrites it untracked. The ownership-scoped reconcile is derived from the `defaults/` walk (`manifest.sh`), which **excludes** the generated `install-metadata.json`, so the staged deletion was never reconciled — leaving `D` (staged) + `??` (untracked) and untracking a `verify_install` critical file on commit.

**Fix:** an explicit reconcile that unstages the staged deletion (guarded by a staged-diff check, no-op on fresh installs) so the rewritten file reappears as a tracked modification, never `D`+`??`. Confirmed the siblings need no change: `.loom/loom-source-path` is gitignored (untracked → no staged deletion); `.loom/config/skill-routes.json` ships in `defaults/config` and is already covered by the ownership-scoped pass.

## Acceptance criteria

All 8 new assertions in `scripts/test-install-quick-gitignore-stash.sh` pass against scratch repos:
- staged `.gitignore` edit restored **still staged** (`git diff --staged` lists it)
- unstaged `.gitignore` edit **stays unstaged**
- `install-metadata.json` shows a tracked modification, never `D`+`??`, stays in `git ls-files`, `verify_install` still finds it
- sibling non-Loom staged changes untouched (no #3597 regression)
- existing #3588 conflict-surfacing behavior preserved

## Testing

Verified empirically against scratch git repos (never the loom repo's own index). With the **pristine base** `install.sh` + a freshly-built daemon, the 4 new #3611 assertions fail (they catch the bug); with this fix they pass.

## Note on a pre-existing, out-of-scope failure

Scenario 2 (`#3590` idempotency: a second `--quick` run with **no intervening commit**) fails on the **pristine** `origin/main` `install.sh` with a freshly-built daemon — independent of this change. Root cause: `.loom/config.json` (#3598 merge) is left as a modification after reinstall #1, so on reinstall #2 it is swept into the stash and hits a genuine conflict, and the working-tree `.gitignore` edit lands in the stash rather than the tree. This is a distinct defect from #3611's two and is not fixed here (out of scope). This test file is not wired into CI (`ci.yml` runs `scripts/test-installer.sh`, which does not invoke it), so there is no CI impact. Recommend a follow-up issue for config.json merge idempotency.


rjwalters:
LGTM — approved.

Reviewed the git-index semantics carefully:

**Defect 1 (`stash pop --index`)** — The scoped stash records both index and working trees; the pre-pop `git checkout HEAD -- .gitignore` gives `--index` a clean 3-way base, so a staged edit returns staged and an unstaged edit stays unstaged. `reapply_loom_gitignore_patterns` appends to the working tree only, never the staged copy. The failure path is preserved: a genuine conflict falls to the else branch, rolls .gitignore back to the post-init snapshot, keeps the stash, and surfaces the real conflict output — no silent degrade to an unstaged pop and no #3588 regression. Recovery text correctly updated.

**Defect 2 (metadata reconcile)** — Pathspec-scoped to exactly `.loom/install-metadata.json` on both the `diff --staged` guard and `restore --staged`, so no sibling/non-Loom reach (Scenario 10 confirms `.anvil/install-metadata.json` + `sibling.txt` stay staged — no #3597 regression). In the reinstall flow the only staged state is a `D` deletion of the HEAD-tracked file; unstaging reverts the index to HEAD while finalize's rewritten copy stays on disk → ` M`, never D+??. Since finalize rewrites the file unconditionally, no user data can be lost.

**No-edit claim holds** — manifest.sh correctly excludes install-metadata.json (not in defaults/), stash-scope.sh correctly doesn't stash it (regenerated anyway), and uninstall-loom.sh is the intended source of the staged deletion now reconciled downstream.

**Tests** use `mktemp -d` scratch repos exclusively (fresh `git init` per scenario); the real loom index is never touched.

CI: merge state CLEAN, all non-skipped checks green. The out-of-scope Scenario 2 config.json idempotency caveat is a pre-existing defect and correctly deferred to a follow-up.

PR #3620, opened 2026-07-17
fix(scaffolding): make committed root-CLAUDE.md pointer self-sufficient (#3612)

### LINKED ISSUE #3612: LOOM ORCHESTRATION block references gitignored .loom/CLAUDE.md — dangles in fresh clones / CI
## Summary

The **committed** root `CLAUDE.md` block that `install.sh` manages —

```
<!-- BEGIN LOOM ORCHESTRATION -->
This repository uses [Loom](...). See `.loom/CLAUDE.md` for the full guide (roles, labels, worktrees, configuration).
<!-- END LOOM ORCHESTRATION -->
```

— points at **`.loom/CLAUDE.md`**, but that target is **install-generated and gitignored**. So the reference dangles for anyone who reads the repo without a completed local install: fresh clones, CI checkouts, a teammate who pulled after the file was cleaned, etc. The marker block is version-controlled; its target is not.

Observed in a consumer repo (lean-genius): `git check-ignore .loom/CLAUDE.md` → ignored, and the file was absent in the working tree, so the committed pointer resolved to nothing. Since the block is tool-managed ("edit outside the markers only"), consumers can't/shouldn't fix it by hand.

## Why it happens
- `install.sh` writes `.loom/CLAUDE.md` into the working tree (with `{{...}}` substitution), and `verify-install.sh` treats it as an install artifact.
- But `.loom/CLAUDE.md` is gitignored, so it's never committed — a committed reference to an uncommitted, environment-dependent file.

## Suggested fixes (pick one)
1. **Point the marker at an always-present target** — e.g. the Loom repo guide URL, or a tracked file like `.loom/roles/` — so the committed reference never depends on local install state.
2. **Commit `.loom/CLAUDE.md`** (un-gitignore it) so the target ships with the repo. (Trades off: template placeholders would be version-controlled.)
3. **`verify-install.sh` warns** when the marker block references a file that isn't present, so drift is caught at install/verify time.

Option 1 is the most robust — the committed block shouldn't reference an install-only, gitignored file.

Related install/CLAUDE.md robustness issues: #3527, #3476, #3450, #3325 (this is a distinct dangling-reference variant).

rjwalters:
Closes #3612

## Problem

The LOOM ORCHESTRATION block committed into a consumer repo's root `CLAUDE.md` pointed at `.loom/CLAUDE.md` as "the full guide". That file is an install-only generated artifact — it may be gitignored or simply absent in a fresh clone / CI checkout. When absent, the committed, tool-managed pointer dangles and consumers can't hand-fix it ("edit outside the markers only").

## Fix (Option 1 per curation)

Reword `LOOM_ROOT_POINTER` in `loom-daemon/src/init/scaffolding.rs` so the authoritative reference is the always-present `https://github.com/rjwalters/loom` URL, and `.loom/CLAUDE.md` is described as an optional locally-substituted copy — not a hard "see X for the full guide" dependency. Doc comment updated to record the invariant and the #3612 rationale.

- One-const change; no newly-committed generated file; no reinstall churn.
- Agent auto-discovery of the locally-written `.loom/CLAUDE.md` in `.loom/worktrees/issue-N/` is unchanged (that relies on the post-install file, unaffected by wording).
- All in-file tests assert against the const (`content.contains(LOOM_ROOT_POINTER)`), so they stay green; no golden-string test hardcoded the old text.

## Verification

- `cargo build --workspace` passes.
- `cargo test init::` in loom-daemon: 109 passed, 0 failed.

### DIFF
--- a/loom-daemon/src/init/scaffolding.rs
+++ b/loom-daemon/src/init/scaffolding.rs
@@ -48,10 +48,15 @@ pub const LOOM_SECTION_END: &str = "<!-- END LOOM ORCHESTRATION -->";
 
 /// The short pointer injected into root CLAUDE.md (between section markers).
 ///
-/// The full Loom guide is written to `.loom/CLAUDE.md` in the target repo.
-/// Claude Code auto-discovers `.loom/CLAUDE.md` when agents work in
-/// `.loom/worktrees/issue-N/` via ancestor directory traversal.
-pub const LOOM_ROOT_POINTER: &str = "This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration. See `.loom/CLAUDE.md` for the full guide (roles, labels, worktrees, configuration).";
+/// This block is committed to the consumer repo, so its authoritative reference
+/// is the always-present Loom repository URL — never the install-generated
+/// `.loom/CLAUDE.md`, which may be gitignored or absent in a fresh clone / CI
+/// checkout (issue #3612). Loom additionally writes a locally-substituted copy
+/// of the full guide to `.loom/CLAUDE.md` at install time; Claude Code
+/// auto-discovers that local copy when agents work in `.loom/worktrees/issue-N/`
+/// via ancestor directory traversal, so the auto-discovery behaviour is
+/// unaffected by this wording.
+pub const LOOM_ROOT_POINTER: &str = "This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration — see the Loom repository for the full guide (roles, labels, worktrees, configuration). When installed, Loom also writes a locally-substituted copy of that guide to `.loom/CLAUDE.md`.";
 
 /// Wrap Loom content in section markers
 pub fn wrap_loom_content(content: &str) -> String {


rjwalters:
LGTM! Approved.

Verified against issue #3612:

- **No more dangling pointer**: the reworded `LOOM_ROOT_POINTER` names the always-present `https://github.com/rjwalters/loom` URL as the authoritative full guide, and demotes `.loom/CLAUDE.md` to an optional "locally-substituted copy" rather than a hard "see X for the full guide" dependency. A fresh-clone / CI reader with no `.loom/CLAUDE.md` is no longer stranded.
- **Behavior preserved**: wording-only change to a single const. The install path still writes the full guide to `.loom/CLAUDE.md` and wraps this pointer into root CLAUDE.md (`wrap_loom_content(LOOM_ROOT_POINTER)`), so Claude Code ancestor-dir auto-discovery in `.loom/worktrees/issue-N/` is unaffected.
- **Tests cover the const**: all assertions reference `LOOM_ROOT_POINTER` symbolically (`content.contains(LOOM_ROOT_POINTER)`, `wrap_loom_content(LOOM_ROOT_POINTER)`) — no golden-string test hardcoded the old text, so nothing silently passes stale. Grep confirms no other Rust source references the old pointer string.
- **Local verification**: `cargo build --workspace` clean; `cargo test init::` → 109 passed, 0 failed.
- **CI**: Rust Linting & Build Check + Rust Unit Tests both green; mergeStateStatus CLEAN.

PR #3617, opened 2026-07-17
feat(guard): opt-in guards.rmScope=repo rm-scope mode with ephemeral allowlist (#3610)

rjwalters:
Closes #3610

## Summary

Adds a third opt-in toggle to `guard-destructive.sh` — `guards.rmScope` / `LOOM_RM_SCOPE=repo` — mirroring the existing `sql_guard_enabled()` / `cloud_guard_enabled()` pattern (lazy, cached, env → config → default). **Default OFF preserves current behavior byte-for-byte.**

By default the guard blocks only catastrophic `rm -rf` targets (root, `$HOME`, bare top-level dirs) and allows every deeper subpath — including paths **outside** the repo. When `LOOM_RM_SCOPE=repo` (or `guards.rmScope:"repo"`), an rm target that is neither under the repo/worktree areas nor on a built-in ephemeral allowlist is denied. The catastrophic top-level deny stays unconditional in both modes, so bare `/tmp` and `/` still deny.

## Key design points

- **Lexical-normalization gotcha**: `normalize_abs_path()` does not resolve symlinks, so the ephemeral allowlist lists **both** `/tmp/` and `/private/tmp/` (and the `/var/tmp`, `/var/folders` symlink pairs), plus the scratchpad glob `*/claude-*/*/scratchpad/*`.
- **Worktree areas** resolve via a `resolve_worktree_root()` helper that replicates `loom_worktree_root()`'s `LOOM_WORKTREE_ROOT` env → `worktree.root` config → default `<repo>/.loom/worktrees` precedence inline, keeping the hook self-contained and best-effort (never fails the hook).
- **Prefix precision**: prefix matches carry a trailing `/` so a name-prefix sibling (`/tmpfoo/x`, `<repo>-sibling`) is not admitted.
- Best-effort config read: malformed/missing `.loom/config.json` falls through to OFF; the hook never exits non-zero.

## 8-case matrix (repo mode)

| Target | Decision |
|--------|----------|
| `rm -f /tmp/x/foo.tsv` | ALLOW |
| `rm -rf /private/tmp/…/scratchpad/z` | ALLOW |
| `rm -rf /var/folders/…/T/tmp.123` (`$TMPDIR`) | ALLOW |
| `rm -rf /tmp` (bare) | DENY (top-level) |
| `rm -rf /` | DENY (catastrophic) |
| `rm -rf /Users/someone/important` (outside) | **DENY (new)** |
| `rm -rf <repo>/.loom/tmp/x` | ALLOW |
[...]

PR #3626, opened 2026-07-18
fix(install): preserve tuned guard hook on quick-install path (Fix A only)

### LINKED ISSUE #3625: Make guards.rmScope=repo the default (or preserve tuned guard on install) — off-by-default silently regresses /tmp+scratchpad rm
### Problem

`guards.rmScope` defaults to **absent = off** (`defaults/CLAUDE.md:991`). The safe `repo` mode — the one with the `/tmp` + Claude-scratchpad ephemeral allowlist added in #3617 — only activates when a consumer explicitly sets `guards.rmScope: "repo"` in `.loom/config.json`.

This reproduces the original complaint in #3610's title — *"no safe default out of the box"* — one level up: the fix exists, but out of the box the guard still either blocks `/tmp`/scratchpad `rm` (strict) or allows outside-repo `rm` (permissive). Neither default is the safe one.

The failure mode for a tuned downstream (e.g. `lean-genius`, which had a local fork with the allowlist on unconditionally): `install.sh` `cp`-overwrites the local `guard-destructive.sh`, and because `rmScope` is off by default, the update **silently** re-breaks `rm /tmp/...` and scratchpad cleanup until someone remembers to hand-add the config toggle.

### Request

Make `guards.rmScope: "repo"` the **default** (the safe, repo-scoped behavior with the ephemeral allowlist), keeping `permissive`/off available as explicit opt-outs. Alternatively (or additionally): have `install.sh` detect and preserve a consumer's tuned `guard-destructive.sh` / config rather than overwriting silently, so downstream forks don't regress on update.

Either path lets a consumer update Loom without the guard silently reverting to blocking `/tmp` + scratchpad.

Related: #3610, #3617. Downstream context: lean-genius #39064 / #38607.

rjwalters:
Closes #3625

## Scope: Fix A ONLY

This PR implements **Fix A** from issue #3625 and deliberately does **NOT** implement Fix B.

- **Fix A (this PR):** `install.sh` → `install_hooks_and_cli()` previously did an unconditional `cp` overwrite of every hook in `defaults/hooks/`, silently clobbering a downstream-tuned or forked `guard-destructive.sh` (e.g. a hand-tuned rm allowlist) on the quick-install/update path. The sibling installer `scripts/install-loom.sh` (lines 1099-1116) already preserves existing hooks unless `--force`/`--clean`. This ports that skip-unless-force behavior into the quick path.
- **Fix B (NOT in this PR — deferred pending ADR sign-off):** flipping the `guards.rmScope` default from `"off"` to `"repo"` reverses #3617's deliberately-shipped byte-for-byte-OFF compatibility guarantee. That is an ADR-level decision requiring operator sign-off. `defaults/hooks/guard-destructive.sh` is **unchanged** here.

## Change

`install_hooks_and_cli()` now takes an optional third `force` arg (default `false`):

- Existing `.loom/hooks/<name>` is **preserved** (with a visible `warning`) unless `force == "true"`.
- The fresh `--quick` (METHOD 1) call site passes `force=true` only under `--clean` (a deliberate fresh install); otherwise it preserves.
- The `--quick` reinstall call site preserves by default.

This mirrors the established preserve-unless-force convention in `scripts/install-loom.sh`. Blast radius is install-script only; no runtime behavior change.

## Tests

- **New:** `tests/install/test-hooks-preserve.sh` — 5 assertions covering (1) an existing tuned hook is preserved by default, (2) `force=true` overwrites it, (3) fresh-install installs the canonical hook and marks it executable. Passes 5/5.
- **Regression:** `tests/hooks/test-guard-destructive.sh` still passes 247/247, including the "ships OFF" rmScope matrix (which Fix B would have had to rewrite and this PR intentionally leaves intact).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,10 +48,19 @@ header() {
   echo -e "${CYAN}$*${NC}"
 }
 
-# Install hooks and CLI wrapper that loom-daemon init doesn't handle
+# Install hooks and CLI wrapper that loom-daemon init doesn't handle.
+#
+# Issue #3625: an existing hook may be a downstream-tuned or forked copy — most
+# notably a customized guard-destructive.sh with a hand-tuned rm allowlist — so
+# it must NOT be silently clobbered on the quick-install/update path. Preserve
+# any existing .loom/hooks/<name> unless an explicit force overwrite is
+# requested (the caller passes "true", e.g. behind --clean). This mirrors the
+# preserve-unless-force behavior already in scripts/install-loom.sh:1099-1116,
+# which the quick path previously diverged from with an unconditional cp.
 install_hooks_and_cli() {
   local loom_root="$1"
   local target="$2"
+  local force="${3:-false}"
 
   # Install hooks
   if [[ -d "$loom_root/defaults/hooks" ]]; then
@@ -59,10 +68,14 @@ install_hooks_and_cli() {
     for hook_file in "$loom_root/defaults/hooks/"*.sh; do
       [[ -f "$hook_file" ]] || continue
       hook_name=$(basename "$hook_file")
-      cp "$hook_file" "$target/.loom/hooks/$hook_name"
-      chmod +x "$target/.loom/hooks/$hook_name"
+      if [[ -f "$target/.loom/hooks/$hook_name" ]] && [[ "$force" != "true" ]]; then
+        warning "Preserving existing hook: $hook_name (use --clean to overwrite)"
+      else
+        cp "$hook_file" "$target/.loom/hooks/$hook_name"
+        chmod +x "$target/.loom/hooks/$hook_name"
+        success "Installed hook: $hook_name"
+      fi
     done
-    success "Installed hooks"
   fi
 
   # Install CLI wrapper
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped Fix-A-only PR. Verified against all review criteria.

## Correctness
`install_hooks_and_cli()` now takes an optional third `force` arg (default `false`). Existing `.loom/hooks/<name>` is preserved (with a visible warning) unless `force == "true"`; fresh targets and forced overwrites install the canonical hook and `chmod +x` it. The per-hook `if [[ -f ... ]] && [[ "$force" != "true" ]]` guard is correct, and moving the `success` line inside the else branch (from the old blanket "Installed hooks" after the loop) accurately reflects per-hook outcome.

## Call-site consistency
Both call sites behave correctly:
- METHOD 1 quick-install (install.sh:1006) maps `--clean` → `_HOOK_FORCE=true`, otherwise preserves. This matches the sibling `loom-daemon init --force` gating on `--clean` in the same block, and `FORCE_FLAG` can only ever be `""` or `"--clean"`, so the mapping covers every force case in this path.
- Reinstall-after-uninstall (install.sh:713) preserves by default (2 args). Correct: that path runs a full uninstall first, so canonical hooks get installed when none survive, and any surviving downstream-tuned hook is safely preserved per #3625.

## Convention match
Mirrors `scripts/install-loom.sh:1099-1116` (skip-existing-unless-force). The quick path previously diverged with an unconditional `cp` — this closes that gap.

## Scope (Fix B correctly deferred)
Confirmed `defaults/hooks/guard-destructive.sh` is byte-for-byte unchanged vs main (`git diff` empty). The `guards.rmScope` default remains `"off"` — the #3617 compatibility guarantee is intact, and the ADR-level Fix B is correctly out of scope.

## Test Execution
- `tests/install/test-hooks-preserve.sh` — ✅ 5/5 pass (preserve-by-default, force overwrites, canonical installed under force, fresh install, executable bit).
- `tests/hooks/test-guard-destructive.sh` (regression) — ✅ 247/247 pass; the "ships OFF" rmScope matrix is intact.

## Code quality
No shell-injection or quoting regressions. `shellcheck -x install.sh` reports only pre-existing warnings (lines 768, 891, 1014, 1094) — all outside this diff; the added lines are clean. The test file's SC2329 "never invoked" notes are false positives (the stub loggers are called indirectly inside the eval'd function body).

Merge state: CLEAN. PR body uses `Closes #3625`.

PR #3629, opened 2026-07-19
feat(guard): flip guards.rmScope default off→repo (safe-by-default)

### LINKED ISSUE #3628: ADR: flip guards.rmScope default off→repo? (Fix B from #3625, deferred pending sign-off)
## Summary

Track and decide **Fix B** from #3625 (now CLOSED — only Fix A shipped in PR #3626 / v0.10.10): whether to flip the `guards.rmScope` default in `defaults/hooks/guard-destructive.sh` from `off` (permissive, byte-for-byte-compatible) to `repo` (repo-scoped with the ephemeral `/tmp` + Claude-scratchpad allowlist).

This is an **ADR-level decision** because it *reverses a compatibility guarantee that #3617 deliberately shipped* ("byte-for-byte backward compatible; existing installs see zero behaviour change"). It was explicitly deferred out of PR #3626 pending operator sign-off. This issue is the tracking vehicle for that decision — it does not itself change code.

## Problem statement

`guard-destructive.sh` resolves the rm-scope mode as follows (verified at `defaults/hooks/guard-destructive.sh` lines ~189-207):

```sh
mode=$(jq -r 'if .guards.rmScope == "repo" then "repo" else "off" end' ...) || mode=off
# env LOOM_RM_SCOPE=repo enables; off/0/no disables; overrides config
```

Only an explicit `guards.rmScope == "repo"` (or `LOOM_RM_SCOPE=repo`) enables repo mode. A missing key, any other value, or malformed JSON all resolve to **`off`**. Accepted values are `repo` and `off` (with `0`/`no` env aliases for off).

**Consequence:** the safe behavior — deny an `rm -rf` target that is neither under the repo/worktree areas nor on the built-in ephemeral allowlist — is *silently absent unless a consumer opts in*. Out of the box a downstream gets the permissive default: catastrophic top-level targets (`/`, `$HOME`, bare `/tmp`) are denied, but every deeper subpath is allowed, **including outside-repo paths** like `rm -rf /Users/someone/important` or `rm -rf /opt/some-vendor/important`.

This reproduces the original #3610 complaint ("no safe default out of the box") one level up: #3617 built the fix, but shipped it OFF, so the default is still one of the two unsafe options rather than the safe one.

**Concrete consumer impact (from #3625):** a tuned downstream (`lean-genius`, refs #39064 / #38607) forked `guard-destructive.sh` to run the allowlist unconditionally. On a `--quick` update, `install.sh` used to `cp`-overwrite the local hook, and because `rmScope` is off by default, the update **silently** re-broke `rm /tmp/...` and scratchpad cleanup until someone re-added the config toggle by hand.

## What is ALREADY solved (do not re-solve here)

- **Fix A (PR #3626, merged v0.10.10):** `install.sh` → `install_hooks_and_cli()` no longer unconditionally `cp`-overwrites hooks on the `--quick` path. An existing `.loom/hooks/<name>` is **preserved** (with a visible warning) unless `force == "true"` (fresh `--clean` install). This directly addresses the "silent overwrite of a tuned/forked guard" pain — a consumer who forked the hook keeps their fork across updates. New test: `tests/install/test-hooks-preserve.sh` (5/5).
- **Config-toggle durability (#3598):** `.loom/config.json` is merge-preserved on install, so a consumer who opts in via `guards.rmScope: "repo"` (rather than forking the script) is **already safe across updates** — the key is not dropped.

So the remaining gap Fix B targets is narrow and specific: **the shipped default for a consumer who has done *nothing* (no fork, no config toggle) is still the permissive/off behavior.** Fixes A + #3598 protect consumers who took an explicit action; they do not change the zero-config default.

### Option A — Keep off-by-default as-is (rely on Fix A + config toggle + #3598)

- **Backward-compat / behavior change:** none. Preserves #3617's guarantee exactly.
- **Migration burden:** zero.
- **Test-matrix churn:** none — `tests/hooks/test-guard-destructive.sh` lines ~805-870 keep asserting "ships OFF" in the default state.
- **Interaction with Fix A:** complementary; the fork/config paths are the sanctioned way to get repo mode.
- **Cost:** the zero-config default remains unsafe (outside-repo `rm` allowed). Every new install must *know about* rmScope to be protected — the discoverability problem #3610/#3625 flagged persists.

### Option B — Flip default to `repo`, with explicit `off`/`permissive` opt-outs

- **Backward-compat / behavior change:** this is a **behavior change for every existing install** that has not set the key: outside-repo `rm -rf` deep paths flip from allowed to denied. Reverses #3617's OFF guarantee.
- **Migration burden:** consumers who *rely* on outside-repo `rm` (rare, but e.g. cross-repo cleanup, shared caches outside the ephemeral allowlist) must add `guards.rmScope: "off"` or `LOOM_RM_SCOPE=off`. Needs a changelog callout and ideally a one-line install-time notice.
- **Test-matrix churn:** significant. The "DEFAULT (off) state" block (lines ~820-846) must be rewritten so the default rows assert repo-mode semantics; the current env-driven `assert_*_env "LOOM_RM_SCOPE=repo"` rows become the default rows, and a new explicit-`off` block must assert the opt-out still yields permissive behavior. ~15-20 assertions touched.
- **Interaction with Fix A:** clean. Fix A already preserves a forked hook, so a downstream that wants the *old* permissive default can pin it; the flip only affects consumers on the canonical hook who haven't opted out.
- **Note on accepted values:** the resolver currently only distinguishes `"repo"` vs everything-else. A default flip should make the trinary explicit: `repo` (default), `off`/`permissive` (opt-out). Decide whether `permissive` becomes a recognized synonym for `off` or stays informal.

### Option C — Middle path: flip default only on fresh installs; preserve existing config on upgrade

- **Mechanism:** `install.sh` writes an explicit `guards.rmScope: "repo"` into `.loom/config.json` on **fresh** install only, and leaves existing configs untouched on upgrade (respecting #3598 merge-preserve). The hook's *code* default stays `off`, so an upgraded install with no key keeps permissive behavior; a fresh install gets repo mode via the written key.
- **Backward-compat / behavior change:** none for existing installs (they keep whatever they had, including the absent-key=off default). New installs get the safe behavior.
- **Migration burden:** near-zero. No existing consumer is surprised.
- **Test-matrix churn:** small in `test-guard-destructive.sh` (the hook code default is unchanged, so the "ships OFF" unit matrix stays valid). New coverage needed in `tests/install/` to assert fresh install writes the key and upgrade does not clobber an existing `guards` block.
- **Interaction with Fix A / #3598:** strong. This is the natural composition — Fix A preserves forked hooks, #3598 preserves config keys, and Option C uses that same install-time seam to seed the safe default only where there is no prior state to protect.
- **Cost:** the *shipped hook* still defaults to permissive, so a consumer who runs the hook without the install-seeded config (e.g. copies the hook standalone, or deletes the key) reverts to permissive. The safety guarantee is install-mediated, not code-intrinsic.

## Compatibility & migration impact

Flipping the code default (Option B) reverses the explicit guarantee in #3617's design ("existing installs see zero behaviour change"). Blast radius = **every install and every upgrade path** that runs the canonical `guard-destructive.sh`, because the change is in the runtime resolver, not the installer. Sequencing to do it safely:

1. **Gate behind a minor bump** (Loom is pre-1.0; treat as a notable behavior change, not a patch). Call it out prominently in the changelog and the `defaults/CLAUDE.md` rm-guard section.
2. **Preserve explicit opt-outs.** An existing `guards.rmScope: "off"` (or `"permissive"`) must continue to mean permissive after the flip. The resolver must not collapse a present-but-off key into the new default.
3. **Announce the opt-out** (`guards.rmScope: "off"` / `LOOM_RM_SCOPE=off`) in release notes and, ideally, an install-time stderr notice on upgrade when no key is present.
4. Option C sidesteps most of this by only seeding the default on fresh installs — it is the lowest-blast-radius way to move the *effective* default without reversing the *code* default.

## Acceptance criteria

Depends on the direction chosen (see recommendation). For the recommended **Option C**:

- `install.sh` seeds `guards.rmScope: "repo"` into `.loom/config.json` on fresh install only (no existing `guards.rmScope` key), and never overwrites an existing value on upgrade — honoring the #3598 merge-preserve contract.
- The hook's code default stays `off`, so `tests/hooks/test-guard-destructive.sh` (lines ~805-870) is unchanged and keeps passing (247/247).
- New install-test coverage (extend `tests/install/test-hooks-preserve.sh` or a sibling) asserts: (a) fresh install writes `guards.rmScope: "repo"`; (b) upgrade with an existing `guards` block does not clobber it; (c) upgrade with an existing explicit `off` is preserved.
- `defaults/CLAUDE.md` rm-guard section (~lines 982-1031) documents that fresh installs are seeded into repo mode and how to opt back to permissive.

If **Option B** is chosen instead:

- `defaults/hooks/guard-destructive.sh` resolver default flips to `repo`; `off`/`permissive` become explicit opt-outs.
- `tests/hooks/test-guard-destructive.sh` default-state block (~lines 820-846) rewritten so default rows assert repo semantics; a new explicit-`off` opt-out block added.
- `defaults/CLAUDE.md` rm-guard section rewritten to state repo is the default.
- Changelog + upgrade notice per the sequencing above.

## Recommendation

**Option C (seed-on-fresh-install).** It delivers the safe default where #3610/#3625 want it — a new consumer gets `/tmp` + scratchpad protection and outside-repo deny with zero manual steps — while honoring #3617's promise that *existing* installs see no behavior change. It composes cleanly with the two fixes that already shipped (Fix A preserves forked hooks; #3598 preserves config keys), reuses the same install-time seam, and keeps the well-tested hook resolver and its "ships OFF" unit matrix untouched, so the churn is confined to the installer and its tests rather than the runtime guard every consumer executes. Option B remains the cleaner *end state* (safe-by-default in the code itself) and is worth revisiting at a future minor/major boundary once Option C has normalized repo mode as the common configuration, but reversing the shipped code default now — for every existing upgrade in one step — is disproportionate to the narrow remaining gap.

## References

- #3625 — parent issue (CLOSED; Fix A only). This proposal is the Fix B tracking vehicle.
- PR #3626 — Fix A (merged v0.10.10): preserve tuned guard hook on quick-install path.
- #3617 — shipped opt-in `guards.rmScope=repo` with the ephemeral allowlist; the OFF-by-default compat guarantee this proposal weighs reversing.
- #3610 — original "no safe default out of the box" report.
- #3598 — `.loom/config.json` merge-preserve (why the config-toggle path is already upgrade-safe).

rjwalters:
## Summary

Implements **ADR Option B** from #3628 (operator-signed-off): flips the `guards.rmScope` default in `defaults/hooks/guard-destructive.sh` from `off` (permissive) to `repo` (repo-scoped, safe-by-default). A zero-config install now **denies** an outside-repo recursive `rm` (e.g. `rm -rf /Users/someone/important`) while keeping the `/tmp` + Claude-scratchpad ephemeral allowlist and in-repo/worktree paths allowed. The catastrophic top-level denies (bare `/tmp`, `/`, `$HOME`) are unchanged in both modes.

This is a deliberate **behaviour change** that reverses #3617's "existing installs see zero behaviour change" guarantee.

## Changes

- **Resolver flip** (`rm_scope_repo_enabled`): default mode is now `repo`. The trinary is explicit and load-bearing:
  - Config `guards.rmScope`: `"off"` / `"permissive"` → permissive (opt-out); **absent key / any other value / malformed JSON → `repo`** (the safe default, via the `|| mode=repo` best-effort fallback that never trips the ERR trap).
  - Env `LOOM_RM_SCOPE`: `repo` → repo; `off`/`0`/`no`/`permissive` → off; absent → falls through to config/default.
  - A *present-but-off* key still means permissive after the flip (not collapsed into the new default).
  - `"permissive"` accepted as a synonym for `"off"` (config and env).
- **Tests** (`tests/hooks/test-guard-destructive.sh`): DEFAULT-state matrix rewritten to assert repo semantics (outside-repo deep `rm` now denied; `/tmp`/scratchpad/`$TMPDIR` allowed; catastrophic rows still deny). New explicit-`off`/`permissive` opt-out block (config + env). Malformed-config now asserts `repo` (deny). All functional assertions pass.
- **Docs** (`defaults/CLAUDE.md`): rm-guard section rewritten to state repo is the default and document `guards.rmScope: "off"` / `LOOM_RM_SCOPE=off` (and `"permissive"`) as the opt-out.
- **CHANGELOG**: behaviour-change entry under `[Unreleased]`, flagged **requires a minor version bump**.

## Testing

- `bash tests/hooks/test-guard-destructive.sh` — 255 functional assertions pass. (The one intermittent perf-threshold failure is pre-existing/environmental — it also fails on unmodified `main` HEAD; a clean run measured 255/255 at 178ms.)
- `bash tests/hooks/test-guard-loom-workflow.sh` — 22/22 pass.
- Manual resolver edge-case checks: absent key → deny outside-repo; `"banana"` → deny; `"off"`/`"permissive"` → allow; `LOOM_RM_SCOPE=off` overrides absent-key default → allow.

## Out of scope

Option C fresh-install seeding, Fix A (#3626), and #3598 config merge-preserve are not touched. No version bump / release cut here — flagged in the CHANGELOG for the operator's `/repo:release`.

Closes #3628

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+> **Requires a MINOR version bump** — the rm-guard default change below is a behaviour change, not a bug fix.
+
+### Changed
+
+- **BREAKING (behaviour): `guards.rmScope` now defaults to `repo` instead of `off`** — `guard-destructive.sh` runs the repo-scoped `rm` guard by default: an outside-repo recursive `rm` (e.g. `rm -rf /Users/someone/important`) is now **denied** out of the box, while the `/tmp` + Claude-scratchpad ephemeral allowlist and in-repo/worktree paths stay allowed. The catastrophic top-level denies (bare `/tmp`, `/`, `$HOME`) are unchanged. This reverses #3617's "existing installs see zero behaviour change" guarantee as a deliberate, signed-off ADR decision (Option B). Consumers who relied on the old permissive default must opt out with `guards.rmScope: "off"` (synonym: `"permissive"`) in `.loom/config.json` or `LOOM_RM_SCOPE=off`. (#3628)
+
 ## [0.10.10] - 2026-07-18
 
 ### Summary
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -981,24 +981,24 @@ LOOM_GUARD_CLOUD=1 aws ec2 terminate-instances --instance-ids i-1234
 
 ### Repo-Scoped rm Guard (`guards.rmScope` / `LOOM_RM_SCOPE`)
 
-By default, `guard-destructive.sh` blocks only **catastrophic** `rm -rf` targets — root (`/`), the user's `$HOME`, and any bare top-level directory (`/tmp`, `/var`, `/etc`, …). Every deeper subpath is allowed, **including subpaths outside the repository** (e.g. `rm -rf /Users/someone/important`). This permissive default is intentional and ships unchanged.
+By default (as of #3628), `guard-destructive.sh` runs in **`repo` mode**: it blocks the **catastrophic** `rm -rf` targets — root (`/`), the user's `$HOME`, and any bare top-level directory (`/tmp`, `/var`, `/etc`, …) — **and** additionally denies any `rm -rf` target that is neither inside the repo/worktree areas nor on a built-in **ephemeral allowlist**. So an outside-repo deep path like `rm -rf /Users/someone/important` is **denied** out of the box. This is the safe-by-default behaviour (ADR Option B); it is a **behaviour change** from the pre-#3628 permissive default.
 
-Repos that want a tighter blast radius can opt in to **`repo` mode**, which additionally denies any `rm -rf` target that is neither inside the repo/worktree areas nor on a built-in **ephemeral allowlist**. The catastrophic top-level deny stays active in both modes, so bare `/tmp` and `/` are always blocked.
+Repos that need the old permissive behaviour — block only catastrophic targets and **allow** every deeper subpath, including subpaths outside the repository — can **opt out** to `off` (a.k.a. `permissive`) mode. The catastrophic top-level deny stays active in both modes, so bare `/tmp` and `/` are always blocked regardless.
 
-The rm-scope guard is **off by default**. It is resolved in this order (highest precedence first):
+The rm-scope guard is **repo (on) by default**. It is resolved in this order (highest precedence first):
 
-1. **`LOOM_RM_SCOPE` env var** — `repo` enables repo mode; `off`/`0`/`no` (or unset) disables it. Overrides the config value.
-2. **`.loom/config.json`** — `guards.rmScope` (default absent = off). Set it to `"repo"` to enable:
+1. **`LOOM_RM_SCOPE` env var** — `repo` forces repo mode; `off`/`0`/`no`/`permissive` forces the permissive opt-out; unset falls through to the config/default. Overrides the config value.
+2. **`.loom/config.json`** — `guards.rmScope`. An explicit `"off"` (or its synonym `"permissive"`) opts out to permissive mode; an absent key, any other value, or malformed JSON resolves to `"repo"` (the safe default):
    ```json
    {
      "guards": {
-       "rmScope": "repo"
+       "rmScope": "off"
      }
    }
    ```
-3. **Default** — off (current permissive behavior, byte-for-byte).
+3. **Default** — repo (safe-by-default, outside-repo deep `rm` denied).
 
-The config read is best-effort: a missing, empty, or malformed `.loom/config.json` falls through to **off** (no behavior change) and never causes the hook to exit non-zero. Enabling repo mode does not weaken any other guard.
+The config read is best-effort: a missing, empty, or malformed `.loom/config.json` falls through to **repo** (the safe default) and never causes the hook to exit non-zero. The permissive opt-out does not weaken any other guard — the catastrophic denies stay active.
 
 **In-scope targets** (allowed under `repo` mode):
 
@@ -1019,16 +1019,19 @@ Plus the Claude scratchpad glob `*/claude-*/*/scratchpad/*`. A **bare** temp roo
 **Examples**:
 
 ```bash
-# Opt in for a whole repo
-#   .loom/config.json  ->  { "guards": { "rmScope": "repo" } }
+# Default (repo mode) — no config needed:
+rm -rf /Users/someone/important   # DENIED (outside repo, safe default)
+rm -rf /tmp/build-cache/x         # allowed (ephemeral allowlist)
+rm -rf ./dist                     # allowed (under repo)
 
-# One-off: strict scope for a single command
-LOOM_RM_SCOPE=repo rm -rf /Users/someone/important   # DENIED (outside repo)
-LOOM_RM_SCOPE=repo rm -rf /tmp/build-cache/x          # allowed (ephemeral)
-LOOM_RM_SCOPE=repo rm -rf ./dist                      # allowed (under repo)
+# Opt out to the old permissive behaviour for a whole repo:
+#   .loom/config.json  ->  { "guards": { "rmScope": "off" } }        # or "permissive"
 
-# Env override wins over config — force OFF even when the repo opts in
-LOOM_RM_SCOPE=off rm -rf /Users/someone/scratch       # allowed (guard disabled)
+# One-off env opt-out — force permissive for a single command:
+LOOM_RM_SCOPE=off rm -rf /Users/someone/scratch       # allowed (permissive)
+
+# Force repo mode for one command even when the repo opts out:
+LOOM_RM_SCOPE=repo rm -rf /Users/someone/important    # DENIED (outside repo)
 ```
 
 ### Protecting Read-Only Directories
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -161,45 +161,54 @@ cloud_guard_enabled() {
 }
 
 # =============================================================================
-# rm-scope repo mode toggle — default OFF (opt-in).
+# rm-scope repo mode toggle — default REPO (safe-by-default; opt out to off).
 #
-# By default this guard blocks only catastrophic rm targets (root, $HOME, and
-# bare top-level dirs) and ALLOWS every deeper subpath — including subpaths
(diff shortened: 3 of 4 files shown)


rjwalters:
## Judge review: APPROVED ✅

Reviewed against the operator-signed-off ADR (Option B, #3628 sign-off comment). This is a security-guard behaviour change, so I verified it rigorously. Verdict: meets the bar for merge.

### Resolver correctness (`rm_scope_repo_enabled`) — verified in source + suite
- `jq -r 'if (.guards.rmScope == "off" or .guards.rmScope == "permissive") then "off" else "repo" end'` — absent key / unrecognized value (`"banana"`) / malformed JSON all resolve to **repo** (new safe default). The `|| mode=repo` fallback catches jq's non-zero exit on malformed JSON without tripping the ERR trap; the `[[ -n "$mode" ]] || mode=repo` guard covers empty output.
- Explicit `"off"` and `"permissive"` still resolve to **off** — a present-but-off key is NOT collapsed into the new default (the load-bearing compat opt-out holds).
- Env `LOOM_RM_SCOPE`: `repo`→repo, `off`/`0`/`no`/`permissive`→off, absent→config/default fallthrough. Env overrides config (confirmed by the `env-off overrides config repo` assertion).

### Catastrophic rules unchanged
Bare `/tmp`, top-level `/`, and `$HOME` still denied in every mode — asserted in both the default and explicit-off/permissive/malformed blocks, plus all destructive-command regression rows pass.

### Tests
DEFAULT-state matrix rewritten to repo semantics (outside-repo `/opt` and `/Users` deep rm now DENIED; `/tmp`/scratchpad/`$TMPDIR` ephemeral allowed). New explicit opt-out block added (config `off`, config `permissive`, `LOOM_RM_SCOPE=off`, `LOOM_RM_SCOPE=permissive`) asserting the old permissive behaviour still holds. Malformed-config row flipped to assert repo/deny. `bash tests/hooks/test-guard-destructive.sh` → **254/255 functional assertions pass**; the sole failure is the known load-dependent perf-threshold row (245ms > 200ms), which is pre-existing/environmental and not a functional regression.

### Docs + CHANGELOG
- `defaults/CLAUDE.md` rm-guard section rewritten: repo is now the default, `guards.rmScope: "off"`/`"permissive"` and `LOOM_RM_SCOPE=off` documented as the opt-out.
- CHANGELOG entry present under `[Unreleased]`, flags a **MINOR** bump and the behaviour change explicitly.

### Scope discipline
Exactly 4 files touched (CHANGELOG, defaults/CLAUDE.md, the hook, the hook test). Option C fresh-install seeding NOT implemented, `install.sh` untouched, Fix A/#3598 not re-solved, no self version bump.

CI: checks pass (Installer Integration Tests green; Rust/MCP jobs correctly skipped for non-code paths). mergeStateStatus CLEAN, mergeable MERGEABLE.

LGTM. Moving to `loom:pr`.

PR #3643, opened 2026-07-20
fix(worktree): guard core.hooksPath on .githooks/ existence (#3638)

### LINKED ISSUE #3638: worktree.sh sets core.hooksPath=.githooks unconditionally → dangling hooks path when repo has no .githooks/
## Summary

`worktree.sh` unconditionally runs `git config core.hooksPath .githooks` for every worktree it creates, and lists `.githooks` in its always-include safety set — i.e. it assumes the repo has a `.githooks/` directory. In a consumer repo that has **no** `.githooks/`, this sets a **dangling hooks path**: git is pointed at a directory that doesn't exist, so hooks silently never run (and the "always-include `.githooks`" step is a no-op with nothing to materialize).

## Where

`.loom/scripts/worktree.sh`:

```sh
# Set git hooks path so .githooks/ works in worktrees (no npx/husky needed)
git -C "$ABS_WORKTREE_PATH" config core.hooksPath .githooks
```

(around line 1149-1150; unconditional — no `[ -d .githooks ]` guard). `.githooks` also appears in `LOOM_WORKTREE_ALWAYS_INCLUDE_DEFAULT` (~line 28) and the include-set docs (~line 458).

## Observed

In botho (`botho-project/botho`), the main checkout has `core.hooksPath=.githooks` set in `.git/config`, but **`.githooks/` does not exist and is not tracked**. Result:
- Main checkout: `core.hooksPath` points at a missing dir → no hooks run.
- Every Loom worktree: `worktree.sh` re-sets `core.hooksPath=.githooks` and tries to always-include `.githooks`, but there is no source dir → dangling config, hooks silently disabled.

This is easy to miss because git does not warn when `core.hooksPath` points at a nonexistent directory — it just runs nothing. A repo that *thinks* it has pre-commit protection (lint, format, secret-scan) would silently have none inside worktrees.

## Requested fix

Guard the hooks-path configuration behind an existence check, e.g.:

```sh
if [[ -d "$MAIN_WORKSPACE_DIR/.githooks" ]]; then
    git -C "$ABS_WORKTREE_PATH" config core.hooksPath .githooks
fi
```

and skip `.githooks` from the always-include materialization when the source directory is absent. Optionally, emit a one-line `print_warning` when `core.hooksPath` is already set in the repo but the target dir is missing, so the dangling state is visible rather than silent.

## Note

This is separate from (and lower severity than) the judge `git read-tree` index-clobber bug filed alongside it — but both surfaced during the same forensic pass on a `/loom:sweep all` run.

## Environment

- git 2.55.0

rjwalters:
## Summary

`defaults/scripts/worktree.sh` unconditionally ran `git config core.hooksPath .githooks` on every new worktree. In a repo **without** a tracked `.githooks/` dir, this pointed `core.hooksPath` at a nonexistent directory — which git treats as "no hooks", silently disabling any hooks configured elsewhere (e.g. a repo-level `core.hooksPath`).

## Fix

Guard the config behind a `.githooks/` existence check:

```sh
if [[ -d "$WORKTREE_REPO_ROOT/.githooks" ]]; then
    git -C "$ABS_WORKTREE_PATH" config core.hooksPath .githooks
fi
```

Per the curator's correction, `$WORKTREE_REPO_ROOT` (the main repo root captured at ~L824) is used instead of `$MAIN_WORKSPACE_DIR`, which is not defined until ~L1224 — well after this line.

The `LOOM_WORKTREE_ALWAYS_INCLUDE_DEFAULT` `.githooks` entry and the help-text line are harmless no-ops when `.githooks/` is absent (cone mode skips missing paths) and are left as-is.

## Tests

New `defaults/scripts/tests/test-worktree-hookspath.sh` covers both branches:
- (a) repo **WITH** `.githooks/` → worktree's `core.hooksPath` == `.githooks` (behavior unchanged)
- (b) repo **WITHOUT** `.githooks/` → `core.hooksPath` left unset (no dangling pointer)

Verified:
- new test: 2/2 pass
- full `test-worktree-*.sh` suite: all green
- `pytest test_worktree.py test_worktree_safety.py`: 44 passed, 1 skipped

Closes #3638

### DIFF
--- a/defaults/scripts/tests/test-worktree-hookspath.sh
+++ b/defaults/scripts/tests/test-worktree-hookspath.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+# test-worktree-hookspath.sh — Tests for the core.hooksPath guard (#3638)
+#
+# worktree.sh sets `git config core.hooksPath .githooks` on every new worktree
+# so a repo's tracked .githooks/ works without npx/husky. Prior to #3638 this
+# was unconditional: repos WITHOUT a .githooks/ dir got core.hooksPath pointed
+# at a nonexistent directory, which git treats as "no hooks" — silently
+# disabling any hooks configured elsewhere (e.g. a repo-level core.hooksPath).
+#
+# The fix guards the config on `[[ -d "$WORKTREE_REPO_ROOT/.githooks" ]]`.
+#
+# Coverage:
+#   1. Repo WITH .githooks/ → worktree's core.hooksPath == .githooks (unchanged).
+#   2. Repo WITHOUT .githooks/ → worktree's core.hooksPath is unset (no dangling
+#      pointer).
+#
+# Pattern follows test-worktree-root-override.sh: throwaway bare origin + repo
+# in a mktemp dir, copy worktree.sh + lib/, run.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPTS_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+WORKTREE_SH="$SCRIPTS_DIR/worktree.sh"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+TESTS_RUN=0
+TESTS_PASSED=0
+TESTS_FAILED=0
+
+pass() { TESTS_RUN=$((TESTS_RUN + 1)); TESTS_PASSED=$((TESTS_PASSED + 1)); echo -e "  ${GREEN}PASS${NC}: $1"; }
+fail() { TESTS_RUN=$((TESTS_RUN + 1)); TESTS_FAILED=$((TESTS_FAILED + 1)); echo -e "  ${RED}FAIL${NC}: $1"; }
+
+assert_eq() {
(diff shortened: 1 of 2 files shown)


rjwalters:
LGTM! Approved.

Reviewed the core.hooksPath guard (#3638). Verification:

- **Guard variable correct**: uses `$WORKTREE_REPO_ROOT` (set at L824 = main repo root). Confirmed `$MAIN_WORKSPACE_DIR` is only first assigned at L1224 — after this guard (~L1150) — so the curator's correction was load-bearing: the alternative would have evaluated `[[ -d "/.githooks" ]]` against an empty var and silently never set hooksPath. Checking the main root's `.githooks/` is sound since the worktree shares the same tracked tree.
- **Behavior**: WITH `.githooks/` → `core.hooksPath == .githooks` (unchanged); WITHOUT → hooksPath left unset (no dangling pointer that git would treat as "no hooks").
- **Test**: `test-worktree-hookspath.sh` builds a throwaway bare-origin + repo in mktemp (follows the `test-worktree-root-override.sh` pattern, no dependency on loom's own `.githooks/`) and covers both branches. Passes 2/2 locally.
- **No regressions**: full `test-worktree-*.sh` suite (concurrency, json-purity, nested-symlinks, root-override, sentinel, sentinel-reinvoke) green against this PR's worktree.sh.
- **CI**: Shellcheck + Installer Integration Tests + Detect Changes pass (Rust/MCP correctly skipped); mergeStateStatus CLEAN, mergeable MERGEABLE.

Minimal, well-documented fix with a targeted regression test. Approved.

PR #3645, opened 2026-07-20
fix(uninstall): count symlinks/subdirs in Step 7 empty-dir check (#3634)

### LINKED ISSUE #3634: install.sh --quick git-index reconcile clobbers co-installed tool symlinks under .claude/
## Context

Hit while running `install.sh --quick -y <repo>` to re-vendor Loom **0.10.9 → 0.10.10** into a repo that *also* has a sibling tool (Repo Skills, `rjwalters/repo`) dev-installed via symlinks under `.claude/`.

The target repo had two gitignored dev symlinks from Repo Skills:
- `.claude/commands/repo/` (dir of symlinks → source `commands/repo/*.md`)
- `.claude/skills/repo/` (contains `SKILL.md` symlink + metadata)

After the Loom quick install completed (exit 0), the installer's **"Reconciling git index after reinstall"** step had **removed `.claude/commands/repo/` entirely** while leaving `.claude/skills/repo/` intact — an inconsistent half-install of the *other* tool. The Repo Skills `/repo:*` commands silently disappeared until we re-ran that tool's own dev installer.

This is the same class of bug 0.10.10 was hardening (#3603 "verify-install scopes its manifest to Loom-owned files", #3601 "reinstall stash guard scoped to Loom-owned paths") — the git-index reconcile just isn't scoped tightly enough and reaches into a co-installed tool's namespace under `.claude/`.

### Repro
1. In a repo, dev-install Repo Skills: `<repo-skills>/install.sh --dev -y .` (creates gitignored `.claude/commands/repo` + `.claude/skills/repo` symlinks).
2. Run `<loom>/install.sh --quick -y <same-repo>`.
3. Observe `.claude/commands/repo/` is gone; `.claude/skills/repo/` survives.

- [ ] The `--quick` reinstall git-index reconcile operates only on Loom-owned paths under `.claude/` (e.g. `.claude/commands/loom/`, `.claude/agents/`, Loom-owned `settings.json` entries) and never touches sibling namespaces like `.claude/commands/repo/` or `.claude/skills/repo/`.
- [ ] Gitignored files/symlinks owned by other tools under `.claude/` are left byte-for-byte untouched across a `--quick` reinstall.
- [ ] Regression test: with a co-installed foreign symlink under `.claude/commands/<other>/`, a Loom `--quick` reinstall leaves it intact.

rjwalters:
## Summary

Fixes a data-loss bug where `install.sh --quick` reinstall clobbered a co-installed tool's install directory.

Root cause: Step 7 "Clean Up Empty Directories" in `scripts/uninstall-loom.sh` (line ~1215) used `find "$dir_path" -type f` to test emptiness before `rm -rf`. `find -type f` does **not** count symlinks (type `l`) or sub-directories. Repo Skills installs `.claude/commands/repo/` as a directory of symlinks; after Loom removed its own `.claude/commands/loom/`, `.claude/commands/` held only that foreign symlink subdir, `-type f` reported it "empty", and `rm -rf` destroyed the sibling tool's install.

Fix: replace the predicate with `find "$dir_path" -mindepth 1 -not -name '.DS_Store'`, which counts symlinks, subdirs, and any residual entry. Foreign content is preserved; genuinely-empty Loom-owned dirs still match and are removed.

## Confirmed trigger path

`install.sh --quick` on an already-installed repo takes the reinstall branch and at `install.sh:686` chains `"$LOOM_ROOT/scripts/uninstall-loom.sh" --yes --local "$TARGET_PATH"`. That uninstall runs the full step sequence including Step 7, so the buggy empty-dir cleanup is reached. Verified by tracing the call chain (line 604 reinstall branch → line 686 chained uninstall → uninstall Step 7 at line ~1215).

The curator's re-localization is confirmed: the `install.sh:744` git-index reconcile is **not** the culprit (it only runs `git restore --staged`/`git reset HEAD`, never touches the working tree) — left unchanged.

## Ownership guard

Not added. The `-mindepth 1` fix is load-bearing and fully satisfies the acceptance criteria on its own (foreign symlink dirs preserved; empty Loom dirs still removed). Sourcing `manifest.sh` into the uninstall shell to intersect residual children against the ownership set would add non-trivial surface for an edge case already covered by "preserve if anything foreign remains". Kept the change minimal and clean per curator judgment guidance.

## Test

New `tests/install/test-uninstall-preserve-siblings.sh` (modeled on `test-hooks-preserve.sh`):
- Scaffolds a target tree where a Loom-owned dir contains only a foreign **symlink** subdir (`.claude/commands/repo/` + `.claude/skills/repo/SKILL.md`).
- Asserts the foreign symlink dir and its links survive while the empty Loom `.claude/commands/loom` is still removed.
- Pins the regression by showing the old `-type f` predicate mis-reports the dir as empty.
- Source-checks that Step 7 uses `-mindepth 1` and no longer uses `-type f`.

Results: 12 passed, 0 failed. Existing `test-forge-detect.sh` (19) and `test-hooks-preserve.sh` (5) remain green.

Closes #3634

### DIFF
--- a/scripts/uninstall-loom.sh
+++ b/scripts/uninstall-loom.sh
@@ -1209,21 +1209,31 @@ echo ""
 for dir in "${REMOVE_DIRS[@]}"; do
   dir_path="$WORKTREE_ABS/$dir"
   if [[ -d "$dir_path" ]]; then
-    # Check if directory is empty (or only contains .DS_Store).
+    # Check if directory holds foreign content (files or symlinks), ignoring
+    # empty subdirectories and .DS_Store.
     #
-    # Issue #3634: use `-mindepth 1` (any child) rather than `-type f` (regular
-    # files only). `-type f` does NOT count symlinks (type `l`) or sub-
-    # directories, so a directory whose only remaining content is a co-installed
-    # tool's symlink dir — e.g. Repo Skills' `.claude/commands/repo/` of
-    # symlinks — was judged "empty" and `rm -rf`'d, clobbering a foreign tool's
-    # install. `-mindepth 1 -not -name '.DS_Store'` counts symlinks, subdirs, and
-    # any other residual entry, so a directory still holding foreign content is
-    # correctly left alone. Genuinely-empty Loom-owned dirs still match (empty
-    # output) and are removed.
+    # Issue #3634: count files AND symlinks — `\( -type f -o -type l \)` — rather
+    # than `-type f` (regular files only) or `-mindepth 1` (any child).
+    #
+    # `-type f` alone does NOT count symlinks (type `l`), so a directory whose
+    # only remaining content is a co-installed tool's symlink dir — e.g. Repo
+    # Skills' `.claude/commands/repo/` of symlinks — was judged "empty" and
+    # `rm -rf`'d, clobbering a foreign tool's install (the original bug).
+    #
+    # `-mindepth 1` over-corrects: it counts empty subdirectories as content.
+    # `REMOVE_DIRS` is NOT strictly child-first (`.loom/scripts` precedes
+    # `.loom/scripts/cli`), so a parent dir gets checked while its still-empty
+    # child exists → deemed non-empty → left behind, orphaning real Loom cruft
+    # (regressing CI Test 29).
+    #
+    # `\( -type f -o -type l \)` is the right signal: files/symlinks are foreign
+    # content to PRESERVE, while empty dirs are Loom cruft to REMOVE. A foreign
+    # symlink dir keeps its symlink (found → preserved); a Loom dir holding only
+    # empty subdirs finds nothing → removed. Order-independent.
     #
     # `-print -quit` (instead of piping to `head -1`) avoids SIGPIPE on `find`,
     # which under `set -o pipefail` would trip the EXIT trap and abort the script.
-    remaining=$(find "$dir_path" -mindepth 1 -not -name '.DS_Store' -print -quit 2>/dev/null)
+    remaining=$(find "$dir_path" \( -type f -o -type l \) -not -name '.DS_Store' -print -quit 2>/dev/null)
     if [[ -z "$remaining" ]]; then
       rm -rf "$dir_path"
       REMOVED_LIST+=("$dir/ (empty directory)")
--- a/scripts/uninstall-loom.sh
+++ b/scripts/uninstall-loom.sh
@@ -1209,10 +1209,21 @@ echo ""
 for dir in "${REMOVE_DIRS[@]}"; do
   dir_path="$WORKTREE_ABS/$dir"
   if [[ -d "$dir_path" ]]; then
-    # Check if directory is empty (or only contains .DS_Store)
+    # Check if directory is empty (or only contains .DS_Store).
+    #
+    # Issue #3634: use `-mindepth 1` (any child) rather than `-type f` (regular
+    # files only). `-type f` does NOT count symlinks (type `l`) or sub-
+    # directories, so a directory whose only remaining content is a co-installed
+    # tool's symlink dir — e.g. Repo Skills' `.claude/commands/repo/` of
+    # symlinks — was judged "empty" and `rm -rf`'d, clobbering a foreign tool's
+    # install. `-mindepth 1 -not -name '.DS_Store'` counts symlinks, subdirs, and
+    # any other residual entry, so a directory still holding foreign content is
+    # correctly left alone. Genuinely-empty Loom-owned dirs still match (empty
+    # output) and are removed.
+    #
     # `-print -quit` (instead of piping to `head -1`) avoids SIGPIPE on `find`,
     # which under `set -o pipefail` would trip the EXIT trap and abort the script.
-    remaining=$(find "$dir_path" -type f -not -name '.DS_Store' -print -quit 2>/dev/null)
+    remaining=$(find "$dir_path" -mindepth 1 -not -name '.DS_Store' -print -quit 2>/dev/null)
     if [[ -z "$remaining" ]]; then
       rm -rf "$dir_path"
       REMOVED_LIST+=("$dir/ (empty directory)")
--- a/tests/install/test-uninstall-preserve-siblings.sh
+++ b/tests/install/test-uninstall-preserve-siblings.sh
@@ -3,18 +3,32 @@
 #
 # Usage: ./tests/install/test-uninstall-preserve-siblings.sh
 #
-# Regression: the emptiness predicate used `find "$dir" -type f`, which does NOT
-# count symlinks (type `l`) or sub-directories. A co-installed tool (e.g. Repo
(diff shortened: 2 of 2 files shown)


rjwalters:
## Changes requested — the `-mindepth 1` predicate over-corrects and regresses empty-Loom-dir cleanup

Thanks for the thorough write-up and the new regression test — the diagnosis of the original symlink-clobber bug is exactly right, and the symlink-preservation behavior is correct. But the chosen predicate breaks a different case, and **CI is red because of it**.

### The regression (CI Test 29 "Full uninstall-then-reinstall cycle")

`gh pr checks 3645` shows **Installer Integration Tests failing**; `mergeStateStatus` is `UNSTABLE`. The failing assertion is `test-installer.sh` Test 29: *"Uninstall left key Loom files behind"* (it verifies `.loom/scripts` / `.loom/roles` / `.loom/config.json` are gone after uninstall).

Root cause: `REMOVE_DIRS` (uninstall-loom.sh:616) is **not strictly child-first**:

```
.loom/scripts        # index 619 — parent, checked FIRST
.loom/scripts/cli    # index 620 — child, checked AFTER
...
.loom                # index 623 — holds .loom/scripts
```

`-mindepth 1` counts **empty subdirectories**, not just foreign content. So when `.loom/scripts` is checked, its empty `cli/` child still exists → predicate reports non-empty → `.loom/scripts` is **kept**. `cli/` is then removed, orphaning `.loom/scripts`, which in turn keeps `.loom` alive. Reproduced locally:

```
REMOVED: .loom/bin
REMOVED: .loom/roles
KEPT(non-empty): .loom/scripts     # <-- regression
REMOVED: .loom/scripts/cli
KEPT(non-empty): .loom             # <-- regression
--- survivors: .loom, .loom/scripts
```

The old `-type f` predicate removed all of these (no regular files → `rm -rf` blew away the empty subdir tree). So the fix leaves real cruft behind — directly violating the "genuinely-empty Loom-owned dirs are STILL removed" acceptance criterion.

### Suggested fix

The distinguishing signal between the two cases is **files/symlinks vs. empty directories**:
- Bug case (must preserve): `.claude/commands/repo/foo.md` is a **symlink** (`-type l`).
- Regression case (must remove): `.loom/scripts/cli/` is an **empty dir** (no files, no symlinks).

Counting files *and* symlinks at any depth — while ignoring empty directories — satisfies both:

```sh
remaining=$(find "$dir_path" \( -type f -o -type l \) -not -name '.DS_Store' -print -quit 2>/dev/null)
```

- Foreign symlink dir → the symlink is found → preserved. ✅ (original bug fixed)
- Loom dir holding only empty subdirs → nothing found → removed. ✅ (Test 29 passes)

(Alternatively, reorder `REMOVE_DIRS` strictly child-before-parent, but the predicate change is more robust and order-independent.)

### Test coverage gap

`tests/install/test-uninstall-preserve-siblings.sh` scaffolds its own leaf-only loop (`.claude/commands/loom` as a leaf) and never exercises a Loom dir whose empty subdir precedes it in the real `REMOVE_DIRS` ordering — which is why the regression slipped through. Please add a case where a Loom dir contains only an empty Loom subdir (mirroring `.loom/scripts` + `.loom/scripts/cli`) and assert both are removed, so this stays pinned.

Once the predicate is corrected and Installer Integration Tests go green, I'll re-review. The symlink-preservation half of this change is solid — just needs to not sacrifice empty-dir cleanup.

rjwalters:
## Feedback addressed — predicate corrected to `\( -type f -o -type l \)`

Thanks for the precise diagnosis. You were right: `-mindepth 1` over-corrected by counting empty subdirectories as content, and since `REMOVE_DIRS` is not strictly child-first (`.loom/scripts` precedes `.loom/scripts/cli`), the parent was checked while its empty `cli/` child still existed and got left behind — breaking Test 29.

### What changed

**`scripts/uninstall-loom.sh` Step 7 (~line 1225):**
```sh
remaining=$(find "$dir_path" \( -type f -o -type l \) -not -name '.DS_Store' -print -quit 2>/dev/null)
```
Files/symlinks = foreign content to PRESERVE (original symlink-clobber bug stays fixed); empty dirs = Loom cruft to REMOVE (Test 29 passes). Order-independent, so the non-child-first `REMOVE_DIRS` ordering no longer matters. Updated the rationale comment to document both failure modes.

**`tests/install/test-uninstall-preserve-siblings.sh`** — strengthened per your coverage-gap note:
- Source-check assertions now pin files-AND-symlinks (`-type f` + `-type l`) and assert `-mindepth 1` is gone.
- New case parses the **real `REMOVE_DIRS` ordering from the script** (not a self-scaffolded loop), confirms `.loom/scripts` precedes `.loom/scripts/cli`, scaffolds an empty `.loom/scripts/cli` + `.loom/scripts` tree, replays Step 7 in source order, and asserts the parent is removed (not orphaned).
- Added a guard that pins the `-mindepth 1` regression directly (asserts it wrongly reports `.loom/scripts` non-empty).

### Verification
- `bash tests/install/test-uninstall-preserve-siblings.sh` → **20 passed, 0 failed**.
- `bash scripts/test-installer.sh` → **116 passed, 0 failed** — Test 29 "Key Loom directories removed after uninstall" is green.

Re-labeled `loom:review-requested`. Ready for re-review.

rjwalters:
## Judge Re-Review (2nd pass, post-Doctor) — Approved ✅

The prior change-request concern is fully resolved. The Doctor applied the recommended fix.

**(a) Predicate corrected** — Step 7 in `scripts/uninstall-loom.sh` now uses `find "$dir_path" \( -type f -o -type l \) -not -name '.DS_Store' -print -quit`. This threads the needle between the two opposing failure modes:
- Counts symlinks (`-type l`) → foreign symlink dirs (e.g. Repo Skills' `.claude/commands/repo/`) are preserved (original #3634 bug fixed).
- Does NOT count empty subdirs (unlike `-mindepth 1`) → Loom dirs holding only empty children are still removed, so the non-child-first `.loom/scripts` before `.loom/scripts/cli` ordering no longer orphans cruft (CI Test 29 fixed).
- Order-independent, so it's robust to REMOVE_DIRS ordering.

**(b) Test strengthened** — `tests/install/test-uninstall-preserve-siblings.sh` now parses the real REMOVE_DIRS array from the script, asserts the non-child-first ordering is present, and replays Step 7 over it to prove the empty-subdir-before-parent case removes cleanly. It also pins both regressions (buggy `-type f` sees foreign symlink dir as empty; over-correcting `-mindepth 1` sees empty child as content). Ran locally: **20 passed, 0 failed**.

**(c) CI green** — `Installer Integration Tests` (includes Test 29) passes; `mergeStateStatus` is CLEAN, `mergeable` MERGEABLE.

**(d) No scope creep** — diff is exactly the one-line predicate change (plus a thorough explanatory comment block) and the new regression test. No unrelated changes.

Minor non-blocking nit: the PR *body* still describes the earlier `-mindepth 1` approach; the inline code comment and code itself are correct. Worth a body edit for the historical record, but not a blocker.

LGTM. Approved.

PR #3656, opened 2026-07-20
test(hooks): warn-not-fail guard perf check with opt-in strict gate

### LINKED ISSUE #3653: Flaky ~200ms perf-threshold assertion in test-guard-destructive.sh
## Context

`tests/hooks/test-guard-destructive.sh` contains a **load-dependent performance-threshold assertion** (average guard execution time must be under ~200ms). It flakes on both `main` and feature branches depending on machine load — observed repeatedly this session:
- measured ~246ms on unmodified `main` HEAD (a clean re-run of the same branch hit ~178ms),
- ~245-248ms on multiple feature branches while the functional assertions all passed.

Every builder/judge that ran the suite this session had to manually disregard this one row ("known load-dependent perf-threshold flake") to avoid a false negative — which is exactly the kind of manual carve-out that erodes trust in the suite and can mask a real perf regression.

- [ ] The perf assertion no longer produces spurious failures under normal machine load — choose one: raise the absolute threshold with headroom, switch to a warn-not-fail (informational) mode, or measure a relative/median metric rather than a single-run average against a hard cap.
- [ ] If kept as a hard gate, document the expected environment and make the threshold env-overridable.
- [ ] The rest of `test-guard-destructive.sh` functional assertions are unaffected.

🤖 Filed via `/repo:followups` from a `/loom:sweep all` session.

rjwalters:
Closes #3653

## Problem

The **Performance check** row in `tests/hooks/test-guard-destructive.sh` timed 10 sequential `$GUARD` process spawns against a hardcoded 200ms cap and incremented `FAIL` (forcing `exit 1`) on exceed. That wall time is dominated by shell + `jq`/`python3` interpreter startup — a function of machine load, not guard-logic complexity — so it flaked under contention (observed 178-248ms across identical runs) and could red the entire functional suite.

## Fix (curator-confirmed approach (b))

- **Informational by default**: always print the measured average; never increment `FAIL` in the default path. A gross perf regression is still visible in the output.
- **Opt-in strict gate**: `LOOM_GUARD_PERF_STRICT=1` (or `true`) restores the hard `FAIL`/`exit 1` behavior.
- **Env-overridable threshold**: `LOOM_GUARD_PERF_MAX_MS` (default 200) replaces the hardcoded 200 for both the printed comparison and the strict gate.
- Both env vars documented in a comment above the block.

Functional assertions (`assert_allow`/`assert_deny`/`*_env`, rmScope, force-push/gh-delete rows) are untouched and still gate the suite.

## Verification

- `bash tests/hooks/test-guard-destructive.sh` → perf row prints `INFO` (239ms > 200), suite exits **0** (263/263 passed).
- `LOOM_GUARD_PERF_STRICT=1 LOOM_GUARD_PERF_MAX_MS=1 bash ...` → perf row `FAIL`, suite exits **1** (opt-in gate trips).
- `LOOM_GUARD_PERF_MAX_MS=1 bash ...` (no strict) → `INFO` only, exits **0**.

### DIFF
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -1194,6 +1194,20 @@ echo ""
 echo -e "${YELLOW}--- Performance check ---${NC}"
 # =========================================================================
 
+# The measured average is dominated by 10 sequential guard process spawns
+# (shell + jq/python3 interpreter startup), which is a function of machine
+# load rather than guard-logic complexity. A hard cap therefore flakes under
+# contention, so by default this row is INFORMATIONAL: it always prints the
+# measured average but never increments FAIL.
+#
+# Env vars:
+#   LOOM_GUARD_PERF_MAX_MS  - threshold in ms for the printed comparison
+#                             (default 200).
+#   LOOM_GUARD_PERF_STRICT  - set to 1/true to restore a hard gate: when the
+#                             average meets/exceeds LOOM_GUARD_PERF_MAX_MS the
+#                             suite fails (FAIL++/exit 1). Intended only for
+#                             runs on a deliberately quiescent machine.
+PERF_MAX_MS="${LOOM_GUARD_PERF_MAX_MS:-200}"
 TOTAL=$((TOTAL + 1))
 START=$(date +%s%N 2>/dev/null || python3 -c "import time; print(int(time.time()*1e9))")
 for i in $(seq 1 10); do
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM! Approved.

Reviewed the diff and ran all three scenarios against the PR branch (`origin/feature/issue-3653`):

**(a) Default path no longer fails on a slow measurement** — with `LOOM_GUARD_PERF_MAX_MS=1` (forcing the exceed branch, non-strict), the perf row prints `INFO: ... (>= 1ms threshold; informational only)` and the suite exits **0** (263/263). No FAIL increment. This is the whole point and it works.

**(b) Strict gate + threshold override** — `LOOM_GUARD_PERF_STRICT=1 LOOM_GUARD_PERF_MAX_MS=1` → perf row `FAIL: ... (>= 1ms threshold, LOOM_GUARD_PERF_STRICT)`, suite exits **1**. `LOOM_GUARD_PERF_MAX_MS` correctly overrides both the printed comparison and the strict gate (message shows the 1ms threshold, not 200).

**(c) Tally consistency** — no off-by-one across all three runs: Total always 263; Passed+Failed = Total in every case (263+0, 262+1, 263+0). Exactly one of PASS/FAIL is incremented per branch, and `TOTAL++` fires once.

**(d) Functional assertions unchanged** — diff is scoped entirely to the Performance check block; `assert_allow`/`assert_deny`, rmScope, force-push/gh-delete rows untouched and still gating (proven by the strict run: the *only* failure is the perf row).

Plain `bash tests/hooks/test-guard-destructive.sh` → exit 0. CI: Detect Changes passes; Rust/MCP/Installer jobs correctly skipped (test-only shell change). mergeStateStatus CLEAN, mergeable MERGEABLE.

Clean, well-documented change that fixes the flake (#3653) while preserving an opt-in hard gate for quiescent CI.

PR #3657, opened 2026-07-20
fix(install): make wrapper honor its own --allow-non-main-source / --allow-stale-target suggestion

### LINKED ISSUE #3650: install.sh (Setup wrapper) suggests --allow-non-main-source but does not accept the flag
## Summary
When the Loom **source** checkout is on a non-main / detached ref, the top-level `install.sh` fails with an error telling the user to pass `--allow-non-main-source` — but `install.sh`'s own flag parser doesn't recognize that flag, so following the instructions produces `Unknown flag`.

## Repro
```sh
git -C /path/to/loom checkout v0.10.10       # detached HEAD on a tag
./install.sh --yes /path/to/target
# → ✗ Refusing to install from non-main / stale source in --yes mode.
#     Pass --allow-non-main-source to override.

./install.sh --yes --allow-non-main-source /path/to/target
# → ✗ Error: Unknown flag: --allow-non-main-source
```

## Root cause
- The refusal + suggestion come from the inner `scripts/install-loom.sh` (which *does* implement `--allow-non-main-source`, lines ~43/74/224/230).
- The top-level `install.sh` flag parser only handles `-y/--yes`, `--quick`, `--full`, `-h/--help` — it neither accepts nor forwards `--allow-non-main-source`.

## Expected
Either the wrapper accepts and forwards `--allow-non-main-source` (and `--allow-stale-target`) to `scripts/install-loom.sh`, or the error message points to the invocation that actually works.

## Workaround
Call the inner installer directly:
`scripts/install-loom.sh --yes --allow-non-main-source --allow-stale-target <target>` — though note that path runs the **Full** (GitHub PR) workflow and short-circuits if an install PR already exists. The reliable direct-file install from a pinned tag was `install.sh --quick --yes <target>`.

_Surfaced consumer-side installing Loom from the pinned v0.10.10 tag (detached HEAD)._

rjwalters:
## Summary

`install.sh` tells users to pass `--allow-non-main-source` (and its pair `--allow-stale-target`), but its own flag parser rejected them with `Unknown flag`, and the Full-Install delegation execs forwarded only `--yes`/`$FORCE_FLAG` — so the flags were dropped even if the parser had accepted them. The inner `scripts/install-loom.sh` already declares, accepts, and honors both flags; this PR makes the wrapper honor its own advice (curator option i).

## Changes (`install.sh`)

1. **Parser** — added a pass-through case for `--allow-non-main-source` / `--allow-stale-target` that accumulates into a new `SOURCE_OVERRIDE_FLAGS` array (accept-and-collect only; the wrapper's source guard runs only in the delegated installer).
2. **Forwarding** — both Full-Install delegation execs (reinstall→Full and Full) now append `${SOURCE_OVERRIDE_FLAGS[@]+"${SOURCE_OVERRIDE_FLAGS[@]}"}` (empty-safe expansion matching the existing `INSTALL_FLAGS` pattern).
3. **Docs** — `--help` text and the header comment now list both pass-through flags.

`scripts/install-loom.sh` is unchanged (it already accepts and honors the flags).

## Tests (`scripts/test-installer.sh`, Section 8b)

- Test 48/49: wrapper accepts `--allow-non-main-source` / `--allow-stale-target` (no `Unknown flag`; parser reaches `--help`).
- Test 50: a genuinely unknown flag (`--bogus`) is still rejected.
- Test 51: `--help` documents both flags.
- Test 52: both delegation execs forward `SOURCE_OVERRIDE_FLAGS`.

Full suite green: 121 passed, 0 failed.

Closes #3650

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,10 +3,14 @@
 # Usage: ./install.sh [OPTIONS] [/path/to/target-repo]
 #
 # Options:
-#   -y, --yes    Non-interactive mode (skip confirmation prompts)
-#   --quick      Quick Install - direct install without GitHub workflow
-#   --full       Full Install - creates issue, worktree, and PR
-#   -h, --help   Show this help message
+#   -y, --yes                  Non-interactive mode (skip confirmation prompts)
+#   --quick                    Quick Install - direct install without GitHub workflow
+#   --full                     Full Install - creates issue, worktree, and PR
+#   --allow-non-main-source    Permit installing from a non-main / detached-HEAD Loom source
+#                              (forwarded to scripts/install-loom.sh)
+#   --allow-stale-target       Permit installing over a target whose Loom is newer/stale
+#                              (forwarded to scripts/install-loom.sh)
+#   -h, --help                 Show this help message
 #
 # Examples:
 #   ./install.sh --quick ~/projects/my-app
@@ -286,6 +290,11 @@ echo ""
 # Parse flags
 NON_INTERACTIVE=false
 INSTALL_TYPE=""
+# Source/target override flags accepted by scripts/install-loom.sh. The top-level
+# wrapper does not act on them (its source guard runs only in the delegated
+# installer), but it must accept and forward them so the flags it suggests
+# actually work. See issue #3650.
+SOURCE_OVERRIDE_FLAGS=()
 while [[ "${1:-}" == -* ]]; do
   case "$1" in
     -y|--yes)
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED ✅

Verified the fix for #3650 — install.sh now honors its own `--allow-non-main-source` / `--allow-stale-target` suggestion instead of rejecting it with "Unknown flag".

**Verification:**

- **(a) Parser case AND both forwarding execs** — Confirmed. New `--allow-non-main-source|--allow-stale-target)` case accumulates into `SOURCE_OVERRIDE_FLAGS`, and *both* Full-Install delegation execs (the `.loom`-exists reinstall path at ~L905 and the main workflow path at ~L1111) now forward the array. A parser-only fix would have dropped the flag at exec — the whole bug. Test 52 asserts `grep -c ... == 2`.
- **(b) Empty-safe expansion** — Both execs use `${SOURCE_OVERRIDE_FLAGS[@]+"${SOURCE_OVERRIDE_FLAGS[@]}"}`, avoiding a `set -u` unbound error when no override flags are passed.
- **(c) scripts/install-loom.sh untouched** — Diff touches only `install.sh` and `scripts/test-installer.sh`. The delegated installer already parses and honors both flags (lines 74/78).
- **(d) Suggestion no longer errors** — Behavioral check: old main rejects `--allow-non-main-source` with "Unknown flag"; PR branch accepts and forwards it.
- **(e) Tests green** — Ran `scripts/test-installer.sh` on the PR branch: Section 8b (Tests 48–52) all pass, full suite 113 passed / 0 failed. `--help` docs and header updated.

**CI:** Installer Integration Tests pass, Detect Changes pass; Rust/MCP jobs correctly skipped (no relevant changes). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

Clean, well-scoped, well-tested fix. LGTM.

Issue #42, opened 2025-10-13
Set up TypeScript unit testing with Vitest


rjwalters:
## Problem

The project currently only has Rust unit tests (`cargo test`). There is no testing framework for TypeScript/JavaScript code, which limits our ability to ensure frontend code quality and catch regressions.

## Proposed Solution

Add Vitest as the TypeScript testing framework for unit and integration tests.

## Why Vitest?

- Fast: Vite-native with instant HMR
- Compatible: Jest-like API (easy migration if needed)
- Modern: ESM support, TypeScript support built-in
- Lightweight: Already using Vite for builds
- Great DX: Watch mode, code coverage, UI mode

## Implementation Tasks

1. Install Vitest dependencies:
   ```bash
   pnpm add -D vitest @vitest/ui
   ```

2. Add test scripts to `package.json`:
   ```json
   {
     "scripts": {
       "test:unit": "vitest run",
       "test:watch": "vitest watch",
       "test:ui": "vitest --ui",
       "test:coverage": "vitest run --coverage"
     }
   }
   ```

3. Create `vitest.config.ts`:
   ```typescript
   import { defineConfig } from 'vitest/config';
   
   export default defineConfig({
     test: {
       globals: true,
       environment: 'jsdom',
       setupFiles: ['./src/test/setup.ts'],
     },
   });
   ```

4. Update `check:ci` script to include TypeScript tests:
   ```json
   "check:ci": "pnpm lint && pnpm format:rust && pnpm clippy:locked && pnpm check && pnpm build && pnpm test && pnpm test:unit"
   ```

5. Write example tests for existing modules:
   - `src/lib/state.ts` - Observer pattern, terminal management
   - `src/lib/config.ts` - Config loading/saving
   - `src/lib/label-setup.ts` - Label definitions and formatting (see PR #42 for test examples)
   - `src/lib/themes.ts` - Theme toggling
   - `src/lib/ui.ts` - Pure rendering functions

## Test Coverage Goals

- Core state management (state.ts)
- Config persistence (config.ts)
- Utility functions (label-setup.ts, themes.ts)
- UI rendering helpers (ui.ts)

## Out of Scope

- E2E testing (consider Playwright in a separate issue)
- Visual regression testing (consider Storybook/Chromatic in a separate issue)

## Related

- Issue #38 introduced `label-setup.ts` which would benefit from unit tests
- PR #42 includes commented-out test examples that can be reused once Vitest is set up

## Priority

[...]

PR #3660, opened 2026-07-20
fix(safety): fail-safe orphan recovery when no liveness writer exists (#3651)

### LINKED ISSUE #3651: Reconcile remaining spawn-loop-state.json consumers (orphan_recovery.py may false-orphan live sweeps)
## Context

Deferred half of #3633 (merged via PR #3646). #3633 removed the runnable stale advice strings for the deleted `spawn-loop.sh` and deleted the dead `loom-check-completions` CLI, but **deliberately deferred** repointing/removing the surviving `.loom/spawn-loop-state.json` consumer subsystem — that's a Python→daemon-surface architecture decision, not a string edit.

After `spawn-loop.sh` was deleted in v0.11.0 (#3631), **nothing writes `.loom/spawn-loop-state.json` anymore**, but ~8 Python modules still read it:
- `loom-tools/src/loom_tools/orphan_recovery.py`
- `loom-tools/src/loom_tools/health_monitor.py`
- `loom-tools/src/loom_tools/stuck_detection.py`
- `loom-tools/src/loom_tools/clean.py`
- `loom-tools/src/loom_tools/status.py` (reader body)
- `loom-tools/src/loom_tools/daemon_diagnostic.py` (reader body)
- the `SpawnLoopState` model, `common/state.py::read_spawn_loop_state`, `LoomPaths.spawn_loop_state_file`

## ⚠️ Safety concern (why this is not just cleanup)

The #3633 curator flagged **`orphan_recovery.py` as safety-sensitive**: it keys on `spawn_loop_state.running`. With no writer, the state file is absent/stale, so the module may treat **every** `loom:building` issue as orphaned — which could tear down a **live** sweep's claim mid-flight. This is a latent correctness/safety bug, not a cosmetic one.

## Doc drift (fix alongside)

`.loom/docs/troubleshooting.md` and `defaults/CLAUDE.md` already claim `loom-stuck-detection` uses `mcp__loom__list_sweeps`, but the code still reads `spawn-loop-state.json`. Reconcile the docs with whatever the code ends up doing.

## Reference

- `docs/migration/daemon-state-consumers.md` — the consumer migration inventory to reconcile against.
- #3633 / PR #3646 — the in-scope half already shipped (stale advice + dead completions CLI).

- [ ] Each surviving `spawn-loop-state.json` consumer is either repointed to the live `loom-daemon` surface (`mcp__loom__list_sweeps` / `get_sweep_status`) or removed, per a per-consumer decision documented against `docs/migration/daemon-state-consumers.md`.
- [ ] `orphan_recovery.py` can no longer false-orphan a live sweep when the state file is absent (no writer) — verified by a test.
- [ ] `troubleshooting.md` / `defaults/CLAUDE.md` claims about `loom-stuck-detection` match the actual code.
- [ ] `pytest` green.

🤖 Filed via `/repo:followups` from a `/loom:sweep all` session.

rjwalters:
Closes #3651

## Problem (SAFETY-critical)

`spawn-loop.sh` — the sole writer of `.loom/spawn-loop-state.json` — was removed in v0.11.0. With no writer, its `running` roster is permanently empty, and `orphan_recovery.check_untracked_building` treated **every** open `loom:building` issue as an orphan candidate, guarded only by ~10-minute time-boxes (`has_valid_claim`'s abandonment threshold + the label-grace window). A live sweep building for more than ~10 minutes satisfied neither guard, so `/loom:sweep all --recover` (and `loom-recover-orphans`) could flip its `loom:building` label back to `loom:issue` and clean its worktree **mid-build**. `clean.py::_revert_stale_building_labels_spawn_loop` had the same false-orphan shape.

## Fix (curator-endorsed fail-safe, bounded scope)

1. **`orphan_recovery.py`** — new `gather_liveness_evidence()` unions the real liveness sources: a present state file, a best-effort daemon-registry stub (`None` when unreachable — never hard-required), and `.loom/locks/issue-<N>/` worktree-lifetime locks. `check_untracked_building` now **emits zero `untracked_building` orphans when no authoritative liveness source is available**. Absent liveness data ⇒ treat every claim as ALIVE, never as orphaned. Existing per-issue claim + label-grace guards are retained as defense-in-depth.
2. **`clean.py`** — same guard on `_revert_stale_building_labels_spawn_loop`: no state file **and** no `.loom/locks/issue-<N>/` locks ⇒ revert nothing.
3. **`daemon_cleanup.py`** — removed the dead `_run_orphan_recovery` call path (module has no console entry point; superseded by `cleanup.py`) plus its now-unused `threading` import.
4. **Docs** — reconciled `loom-stuck-detection` claims in `defaults/CLAUDE.md` and `.loom/docs/troubleshooting.md` with the code (it reads `spawn-loop-state.json`, a safe no-op today); recorded every consumer's disposition in `docs/migration/daemon-state-consumers.md`.

## Out of scope (deferred, per curator)

Repointing the four non-destructive reporters (`status.py`, `health_monitor.py`, `daemon_diagnostic.py`, `stuck_detection.py`) + retiring the `SpawnLoopState` model/path — they are report-only, safe no-ops today.

## How the fail-safe invariant is proven

New `loom-tools/tests/test_orphan_recovery.py` (14 tests) + 3 sibling tests in `test_clean.py`:

- **The regression**: no state file, no daemon, no locks + a stale (`label_age > grace`), unclaimed `loom:building` issue under `--recover` ⇒ `total_orphaned == 0`, `total_recovered == 0`, and **no `gh issue edit`** is ever issued.
- **AC (a)**: an active `.loom/locks/issue-42/` lock keeps #42 alive even with an old label and an "abandoned" claim ⇒ 0 orphans.
- **AC (b)**: with an authoritative source present that does not list the issue (a decoy lock for another issue, or a present-but-empty state roster), a genuinely-dead claim **is** still recovered.
- **AC (c)**: `clean.py` sibling — no state file + no locks ⇒ reverts nothing; a locked issue is never reverted; a genuine orphan is reverted once a source is present.
- Defense-in-depth: valid claim and label-grace still protect issues even when a source is present; the `stale_heartbeat` path is unaffected.

## Test results

`cd loom-tools && PYTHONPATH=src python3 -m pytest -q` ⇒ **4 failed, 1719 passed, 62 skipped**. The 4 failures are pre-existing on base (PYTHONPATH-less subprocess spawns in `test_logging`, `test_agent_spawn`, `test_forge_cli`, `test_installation_verification`) — no new failures introduced (+17 new tests all green). `ruff` clean on all changed files (the two remaining F401s and one F821 are pre-existing on base).

### DIFF
--- a/.loom/docs/troubleshooting.md
+++ b/.loom/docs/troubleshooting.md
@@ -224,7 +224,9 @@ When an agent crashes or is cancelled while building, issues can get stuck in `l
 
 ## Stuck Agent Detection
 
-`loom-stuck-detection` (post-v0.10.0) checks for stuck sweep children using the `loom-daemon` sweep registry (`mcp__loom__list_sweeps`) and `.loom/sweep-checkpoint/issue-<N>.json` checkpoint timestamps.
+`loom-stuck-detection` checks for stuck sweep children by reading the per-task heartbeats in `.loom/spawn-loop-state.json::running[].last_heartbeat`.
+
+> **Note (post-v0.11.0):** `spawn-loop.sh` — the only writer of `.loom/spawn-loop-state.json` — was deleted, so this file no longer has a writer. `loom-stuck-detection` therefore currently reports nothing (a safe no-op: it only reports, it never tears down work). Repointing it to the `loom-daemon` sweep registry (`mcp__loom__list_sweeps`) and `.loom/sweep-checkpoint/issue-<N>.json` timestamps is tracked as a follow-up (see `docs/migration/daemon-state-consumers.md`).
 
 ### Check for stuck agents
 
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -1321,11 +1321,15 @@ loom-orphan-recovery --json
 - Adds recovery comments to affected issues
 - Removes stale lock dirs
 
-The daemon's 30-second reaper task usually catches this autonomously; `loom-orphan-recovery` is the manual cross-check.
+> **Fail-safe (#3651):** the untracked-building cross-check requires an authoritative liveness source (a present `.loom/spawn-loop-state.json`, a reachable `loom-daemon` registry, or `.loom/locks/issue-<N>/` locks). When **none** is available — e.g. after `spawn-loop.sh` was removed and no daemon is running — `loom-orphan-recovery` emits **zero** `untracked_building` orphans and recovers nothing. Absent liveness data means "treat every `loom:building` claim as ALIVE", never as orphaned. This prevents tearing down a live sweep that has been building for more than the label-grace window.
+
+The daemon's 30-second reaper task usually catches genuine orphans autonomously; `loom-orphan-recovery` is the manual cross-check.
 
 ### Stuck Agent Detection
 
-`loom-stuck-detection` checks for stuck sweep children by combining the daemon registry (via `mcp__loom__list_sweeps`) with `.loom/sweep-checkpoint/issue-<N>.json` checkpoint timestamps.
+`loom-stuck-detection` checks for stuck sweep children by reading per-task heartbeats in `.loom/spawn-loop-state.json::running[].last_heartbeat`.
+
+> **Note (post-v0.11.0):** `spawn-loop.sh` (the sole writer of `.loom/spawn-loop-state.json`) was deleted, so this file has no writer and `loom-stuck-detection` currently reports nothing — a safe no-op (report-only, never destructive). Repointing it to the `loom-daemon` registry (`mcp__loom__list_sweeps`) plus `.loom/sweep-checkpoint/issue-<N>.json` timestamps is a tracked follow-up (see `docs/migration/daemon-state-consumers.md`).
 
 **Check for stuck agents**:
 ```bash
--- a/docs/migration/daemon-state-consumers.md
+++ b/docs/migration/daemon-state-consumers.md
@@ -35,6 +35,31 @@ The Tauri desktop app the issue speculated about does not exist in this repo. Th
 > (b) is a cross-session log (`issue-failures.json`, `recovery-events.json`) that already lives outside `daemon-state.json` and survives the deletion as-is.
 > If a per-consumer review later decides a minimal-compat shape *is* worth it, that's a follow-up issue per #3377's "Out of Scope" section.
 
+## Post-spawn-loop-deletion fail-safe dispositions (#3651)
+
+`spawn-loop.sh` — the only writer of `.loom/spawn-loop-state.json` — was removed
+in v0.11.0. With no writer the roster is permanently empty, which turned the two
+**destructive** `spawn-loop-state.json` consumers into false-orphan hazards:
+they would flip a live `loom:building` claim back to `loom:issue` (and possibly
+clean its worktree) mid-build. Issue #3651 makes both fail safe and records the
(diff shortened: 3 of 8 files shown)


rjwalters:
## Judge review: APPROVED ✅ (SAFETY-critical)

Reviewed with extra rigor given this automates tear-down of in-flight work. The core safety invariant holds under scrutiny.

**(a) Core invariant — absent liveness ⇒ treat all claims ALIVE.** Traced `gather_liveness_evidence()`: `available = bool(sources)`, and `sources` is appended to *only* when a real source exists — a present state file (`spawn_loop_state.present`), a non-`None` daemon result, or a **non-empty** locks set. `_query_daemon_live_issues` returns `None` today (never adds a phantom source); empty locks add nothing. So `available` is `False` exactly when there is no state file **and** no per-issue locks — precisely the conservative condition. `check_untracked_building` has the gate `if not evidence.available: return` *before* any `gh_issue_list` call, so there is no path where a missing/empty source produces an orphan. A present-but-empty roster is correctly treated as authoritative ("nothing live"), which is the only case where an empty source enables recovery — and that requires a writer, which is the whole point.

**(b) clean.py sibling guard.** `_revert_stale_building_labels_spawn_loop` gates with `if not state_present and not locked_issues: return 0` before listing/editing — equivalent protection. `gh_run`/`subprocess.run` are never reached when there's no evidence (test asserts they raise if called).

**(c) Genuine-dead claim still recoverable.** When an authoritative source IS present but doesn't list the issue (decoy lock for a different issue, or a present-but-empty roster), a stale unclaimed claim is still flagged `untracked_building` and recovered — proven by `test_genuinely_dead_claim_is_recoverable` and `_via_state_roster`. Verified the builder's "currently a safe no-op in prod" reading is accurate, not a bug: the only lock any live script creates is the transient **repo-global** `.loom/locks/worktree-add` (worktree.sh), which the `issue-` prefix filter ignores. No live writer emits per-issue locks after spawn-loop.sh deletion, so recovery is intentionally inert until daemon wiring — the correct fail-safe tradeoff (inert-but-safe over active-but-destructive), with repointing tracked as follow-up.

**(d) Regression test exercises the dangerous path.** `test_no_liveness_source_emits_zero_orphans`: no state file, no daemon, no locks, `_get_building_label_age=9999` (well past grace), `has_valid_claim=False`, `recover=True` → asserts `total_orphaned==0`, `total_recovered==0`, and `gh.edited_issues==[]` (the recorder raises on any `issue edit`). This is exactly the regression that would tear down a live sweep.

**(e) daemon_cleanup dead-path removal.** `_run_orphan_recovery` fully deleted (only comment references remain), the `threading` import removed, no live callers anywhere in the tree, and `handle_daemon_startup` step 3 replaced with a doc comment.

**Verification:**
- `pytest tests/test_orphan_recovery.py tests/test_clean.py` → **25 passed**.
- Full suite → **1719 passed, 62 skipped, 4 failed** — the 4 are the known pre-existing PYTHONPATH-restricted-subprocess failures (test_agent_spawn/forge_cli/installation_verification/logging), none touched by this PR.
- `gh pr checks` green (Installer Integration + Detect Changes pass; Rust/MCP correctly skipped for a Python-only change).
- Merge state CLEAN / MERGEABLE.

Docs (`defaults/CLAUDE.md`, troubleshooting.md, daemon-state-consumers.md) reconciled with the new behavior and the deferred-consumer dispositions table is accurate. Clean, well-tested, and safe. LGTM.

Issue #3486, opened 2026-06-12
fix(merge-pr.sh): --auto deadlocks on UNSTABLE when failing checks are informational-only (non-required)


rjwalters:
## Context

Sibling of #3371 (CLEAN-fallback). #3371 added an immediate-merge fallback when the PR is already CLEAN and auto-merge would otherwise wait for required checks. This issue covers the **opposite failure mode**: a PR is approved, all required checks pass, but one or more **informational/non-required** checks are red — auto-merge waits indefinitely because it tracks the rollup state, not the branch-protection requiredness.

Observed in production at downstream `rjwalters/sphere` (vendored loom 0.9.1) on 2026-06-11 during two `/sweep` runs:

\`\`\`
Error: Failed to enable auto-merge for PR #9234: Pull request is in unstable status
WARNING: Failed to enable auto-merge for PR #9234: gh: Pull request Pull request is in unstable status
Failed to auto-merge PR #9234
\`\`\`

PR #9234 had \`mergeStateStatus: UNSTABLE\` because workflow \`CI: Stack B lockstep (informational, 30-day soak)\` reported 2 failing jobs. That workflow file self-documents:

> Lands as INFORMATIONAL — failing checks DO NOT block merges. After 30 days of green on \`main\`, flip to REQUIRED in branch protection.

Sphere repo's branch protection on \`main\` only lists \`Code Ownership\` as required (verified via GraphQL). All other failing checks are informational. The PR was approved (\`loom:pr\`), the failures were verified pre-existing on \`main\`, and the failures touched zero files in the PR diff.

The session-level workaround used was: drop \`--auto\` and call plain \`./.loom/scripts/merge-pr.sh <pr>\` (immediate merge). This works because the failing checks are not branch-protection-required.

## Why this matters for /sweep

The \`/sweep\` skill always calls \`merge-pr.sh --auto\` in step 7 (Merge). When a downstream repo has informational workflows that go red on main (epic-mid-burn-in, broken-on-main, etc.), every sweep stalls mid-run. Operator has to intervene.

This will get worse as the informational workflow set grows. Sphere just landed epic #9112 (Stack B lockstep) with the informational pattern as the default for new gates.

## Proposed fix (Option A from downstream issue)

**Auto-fallback to immediate merge when only informational checks are failing.**

In \`merge-pr.sh\`, before calling forge auto-merge:

1. Query the PR's check rollup (already done elsewhere in the script).
2. Query the target branch's required status check contexts via GraphQL:
[...]

Issue #3371, opened 2026-06-02
merge-pr.sh --auto fails on PRs already in CLEAN status (fast-CI PRs)


rjwalters:
## Problem

`./.loom/scripts/merge-pr.sh <PR> --auto` fails with GraphQL `UNPROCESSABLE` when the PR has already reached `mergeStateStatus: CLEAN` (all required checks already passed) by the time the script runs.

Surfaced during a `/sweep` run on #3369 → PR #3370 (docs-only change, CI was very fast):

```
$ ./.loom/scripts/merge-pr.sh 3370 --auto
Merging PR #3370: fix(sweep): use live repo label set for Mode B unknown-label guard
Branch: feature/issue-3369
Error: Failed to enable auto-merge for PR #3370: {"data":{"enablePullRequestAutoMerge":null},"errors":[{"type":"UNPROCESSABLE","path":["enablePullRequestAutoMerge"],"locations":[{"line":1,"column":72}],"message":"Pull request Pull request is in clean status"}]}
```

GitHub's `enablePullRequestAutoMerge` mutation rejects PRs that are already in `clean` status because there is nothing to queue — auto-merge is for PRs that are not yet mergeable (waiting on checks). Re-running without `--auto` worked immediately:

```
$ ./.loom/scripts/merge-pr.sh 3370
PR #3370 merged successfully
```

## Why it matters

Both `/sweep` and `/shepherd` prescribe `merge-pr.sh --auto` as the canonical command in their skill docs. On any PR with fast CI (typical for docs-only changes, small Python edits, etc.) the PR can reach `CLEAN` between PR creation and the merge call, breaking the auto path.

Concretely:
- `sweep.md` step 7: ` ./.loom/scripts/merge-pr.sh <PR_NUMBER> --auto`
- `CLAUDE.md` "Merging PRs" example: `./.loom/scripts/merge-pr.sh <PR_NUMBER> --auto   # Auto-confirm (for automation)`

(The CLAUDE.md comment also mislabels `--auto` as "auto-confirm" — see line ~290 of `.loom/scripts/merge-pr.sh`, which makes clear `--auto` means "enable auto-merge instead of immediate merge.")

## Proposed fix

In `merge-pr.sh`, when `--auto` is requested and the GraphQL response contains \`\"Pull request is in clean status\"\`, fall through to the synchronous-merge path instead of erroring. Roughly:

\`\`\`bash
# In the auto-merge block (around line 220):
[...]

PR #3669, opened 2026-07-21
fix(merge): poll UNSTABLE PRs whose required checks are still running (#3664)

### LINKED ISSUE #3664: merge-pr.sh --auto hard-fails when PR merge state is UNSTABLE (checks still running)
## Context
Hit repeatedly (3x) while driving a /sweep in kicad-tools this session (PRs #4357, #4370, #4376 in rjwalters/kicad-tools). When a PR's required checks are still running, `merge-pr.sh <PR> --auto` fails to enable GitHub auto-merge and exits non-zero:

```
Using loom-auto-merge (forge-agnostic auto-merge)
Error: Failed to enable auto-merge for PR #NNNN: ...
WARNING: Failed to enable auto-merge for PR #NNNN: gh: Pull request Pull request is in unstable status
Failed to auto-merge PR #NNNN
```

The intent of `--auto` is to queue the merge until required checks pass, so an UNSTABLE (checks-in-progress) state is exactly the case it should handle -- but instead it hard-errors, forcing the caller to poll manually and re-invoke merge-pr.sh once the state goes CLEAN. This defeats the point of `--auto` for the common "approved, CI still running" path.

## Repro
1. Open a PR whose CI includes long jobs (e.g. ~20 min routing regressions).
2. Approve it and immediately run `merge-pr.sh <PR> --auto` while checks are IN_PROGRESS (mergeStateStatus = UNSTABLE).
3. Observe exit 1 with "Pull request is in unstable status" instead of the merge being queued.

Note: on a CLEAN PR the script's immediate-merge fallback works fine; the failure is specific to the UNSTABLE-but-will-pass state.

- [ ] On UNSTABLE merge state, `--auto` either (a) polls until the PR reaches CLEAN (merge) or a terminal blocked/failed state, or (b) actually enables GitHub's native server-side auto-merge queue (which is designed to hold until checks pass) instead of erroring.
- [ ] A genuinely blocked/failing PR still surfaces a clear non-zero error.
- [ ] Behavior on already-CLEAN PRs unchanged.

---

### Root Cause

`merge-pr.sh` already has an UNSTABLE-fallback (added in #3486, sibling of the CLEAN-fallback from #3371) at `.loom/scripts/merge-pr.sh:352-427`. When `loom-auto-merge` / `forge_auto_merge` rejects the GraphQL `enablePullRequestAutoMerge` mutation with `"is in unstable status"`, the script:

1. Fetches check-runs for the PR's head SHA (`forge_get_check_runs`, `merge-pr.sh:372`).
2. Filters to runs whose `conclusion` is `failure`, `timed_out`, `cancelled`, or `action_required` (`merge-pr.sh:375-376`) into `_UNSTABLE_FAILING`.
3. **If `_UNSTABLE_FAILING` is empty, it hard-errors** (`merge-pr.sh:378-382`):
   ```bash
   if [[ -z "$_UNSTABLE_FAILING" ]]; then
     # No failing checks were enumerated — could be a transient API gap or
     # commit-status (vs check-run) failures. Be safe and keep the existing
     # error path.
     error "Failed to enable auto-merge for PR #$PR_NUMBER: $AUTO_MERGE_OUTPUT"
   fi
   ```
4. Otherwise it computes `failing \ required_contexts` and only falls through to immediate-merge when every failing check is informational (i.e., not in branch protection).

The bug: checks that are still `IN_PROGRESS`/`queued` (i.e. genuinely running, not yet failed) have `conclusion == null`, so they never appear in `_UNSTABLE_FAILING`. GitHub still reports `mergeStateStatus = UNSTABLE` for a PR with pending-but-not-yet-failed required checks (the same string, `"Pull request is in unstable status"`, is returned by the GraphQL mutation for both "checks pending" and "checks failed" cases — see the shared substring matcher at `merge-pr.sh:360`). Step 3's defensive branch was written for the "we don't know why it's unstable" case (transient API gap / commit-status-not-check-run failures), but it also catches the much more common "checks are simply still running" case from this issue's repro — and there it silently converts a `--auto` call (whose entire purpose is "wait for checks") into a hard failure.

This is orthogonal to the #3486 fix: #3486 handles *failed-but-informational* checks; this issue is about *pending* checks, which #3486 never anticipated.

### Implementation Guidance

Recommended approach — extend the existing UNSTABLE-fallback block rather than build a new code path, keeping the same fail-closed posture the block already uses for required checks:

1. In the block at `merge-pr.sh:372-376`, alongside `_UNSTABLE_FAILING`, also compute a **pending** set: check-runs whose `status != "completed"` (i.e. `queued` or `in_progress`), e.g.
   ```bash
   _UNSTABLE_PENDING="$(echo "$_UNSTABLE_FAILING_RAW" | \
     jq -r '[.check_runs[] | select(.status != "completed") | .name] | unique | .[]' 2>/dev/null || true)"
   ```
2. Distinguish three cases where `_UNSTABLE_FAILING` is empty (`merge-pr.sh:378`):
   - `_UNSTABLE_PENDING` is non-empty → genuinely pending checks, not a failure. This is the new case this issue targets.
   - `_UNSTABLE_PENDING` is also empty → keep the existing "unknown gap" hard-error (transient API gap / commit-status-only failures) — do not weaken this fail-closed path.
3. For the pending case, poll instead of erroring immediately: retry the `enablePullRequestAutoMerge` mutation (or just re-check `mergeStateStatus`/check-run status) on an interval, bounded by a timeout, since GitHub's own auto-merge queue is the ultimate mechanism once the PR clears UNSTABLE. Two concrete sub-options, either acceptable:
   - **(a) Poll in bash**: add a bounded retry loop around the existing `MAX_MERGE_RETRIES`/`MERGE_RETRY_DELAY` pattern already used for "Base branch was modified" (`merge-pr.sh:328-338`), but with a longer ceiling appropriate for CI durations (the issue's repro cites ~20 min jobs) — e.g. reuse `LOOM_AUTO_MERGE_POLL_INTERVAL`/`LOOM_AUTO_MERGE_TIMEOUT` env vars already defined for the Gitea path (`loom-tools/src/loom_tools/auto_merge.py:13-14`) so the two forges share configuration semantics, defaulting timeout much higher than the current 3-attempt/5-10s backoff (which only covers seconds, not minutes).
   - **(b) Extend `auto_merge_pull_request` in `loom-tools/src/loom_tools/common/github.py:499-565`**: currently it delegates purely to the one-shot GraphQL mutation and assumes "GitHub handles polling server-side" (see its docstring at `github.py:516-517`) — which is false in the UNSTABLE case being demonstrated here. Have it retry the mutation call itself on the `"is in unstable status"` error (bounded by `timeout`/`poll_interval`, which the signature already accepts but currently ignores per the docstring), succeeding once GitHub accepts the mutation (state moved off UNSTABLE) or the PR resolves to `merged=true` via polling `forge_get_pr_nocache` in between attempts.
   - Prefer (b) if feasible in the PR's scope, since it keeps the polling logic in the Python forge layer (mirroring the existing Gitea `auto_merge_pull_request` poll-and-merge implementation at `loom-tools/src/loom_tools/common/gitea.py:717-789`) rather than duplicating a bash retry loop; fall back to (a) if a pure-bash fix is simpler to land and test.
4. Whichever option is chosen, exit `error` with a clear message if the timeout is exceeded while checks are still pending (distinct from the "checks failed" error message), so the caller isn't left silently hanging forever.
5. Do not touch the existing informational-failure fallback (`merge-pr.sh:400-421`) or the "unknown gap" hard-error for the genuinely-ambiguous case (`_UNSTABLE_FAILING` and `_UNSTABLE_PENDING` both empty) — those paths are correct as-is and covered by `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh`.

### Acceptance Criteria

- [ ] `merge-pr.sh <PR> --auto` on a PR whose `mergeStateStatus` is UNSTABLE *solely* because required checks are still `queued`/`in_progress` (none have failed) no longer exits non-zero immediately — it waits (bounded by a timeout) for the checks to resolve, then either merges (once CLEAN) or reports a clear terminal error (once a required check actually fails).
- [ ] The wait is bounded — a configurable timeout (reusing `LOOM_AUTO_MERGE_TIMEOUT`/`LOOM_AUTO_MERGE_POLL_INTERVAL` semantics, or an equivalent bash-side var) prevents an indefinite hang; exceeding it produces a clear non-zero error distinct from "checks failed".
- [ ] A PR that is UNSTABLE because a *required* check has genuinely failed (not merely pending) still surfaces a clear non-zero error without waiting out the full timeout (existing behavior preserved).
- [ ] The existing #3486 informational-failure fallback (all failing checks are non-required) is unchanged — still falls through to immediate merge without waiting.
- [ ] Already-CLEAN PRs and the #3371 CLEAN-fallback are unaffected.
- [ ] The "unknown gap" defensive hard-error (`_UNSTABLE_FAILING` empty AND no pending checks detected — e.g. commit-status-only failures the check-runs API doesn't enumerate) is preserved; only the genuinely-pending case is changed.
- [ ] `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` gains coverage for the new "all unstable checks are pending, none failed" branch (asserting the fallback/poll path fires, not the hard error), alongside the existing CLEAN/informational/required-failure cases.

### Affected Files

- `defaults/scripts/merge-pr.sh` (the canonical source; `.loom/scripts/merge-pr.sh` is a symlink to it) — extend the UNSTABLE-fallback block at lines ~352-427, specifically the "no failing checks enumerated" branch at lines 378-382.
- `loom-tools/src/loom_tools/common/github.py` — `auto_merge_pull_request` (lines 499-565), if implementation option (b) is chosen: add bounded polling/retry on the `"is in unstable status"` GraphQL error instead of a pure one-shot call.
- `loom-tools/src/loom_tools/auto_merge.py` — CLI entry point already exposes `--poll-interval`/`--timeout` flags and `LOOM_AUTO_MERGE_POLL_INTERVAL`/`LOOM_AUTO_MERGE_TIMEOUT` env vars; no change expected here, but verify the values flow through to the GitHub path once (b) uses them (today only the Gitea path honors them meaningfully).
- `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` — add test coverage for the new "pending, not failed" branch.
- `defaults/scripts/lib/forge-helpers.sh` (canonical source; `.loom/scripts/lib/forge-helpers.sh` is a symlink) — `forge_get_check_runs` (used to build the new pending-set query); no change expected, just consumed with an additional jq filter.

### Test Plan

- [ ] Manual verification: Open a real PR with a long-running required check, approve it, run `merge-pr.sh <PR> --auto` while the check is still `in_progress`, and confirm the script now waits/queues instead of exiting 1 with "is in unstable status". Confirm it eventually merges once CI passes.
- [ ] Manual verification (negative case): Force a required check to fail on a test PR, run `--auto`, and confirm the script still surfaces a clear non-zero error promptly (does not wait out the full pending-timeout).
- [ ] Automated tests: Extend `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` with a case where `_UNSTABLE_FAILING` is empty but simulated check-runs include `status: "in_progress"`/`"queued"` entries — assert the new pending-branch is taken (poll/fallback) rather than the existing hard-error branch.
- [ ] Regression: run the full `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` suite to confirm the #3371 CLEAN-fallback, #3486 informational-failure fallback, and the "unknown gap" hard-error path are all unchanged.
- [ ] Regression: run `defaults/scripts/tests/test-merge-pr-help.sh` and `defaults/scripts/tests/test-merge-pr-worktree-path.sh` to confirm no unrelated regressions in `merge-pr.sh` argument handling.

### Related

- #3371 — original CLEAN-status fallback (sibling pattern this issue extends).
- #3486 — UNSTABLE-status fallback for informational check failures (the "failed but not required" case this issue's "pending, not yet failed" case is a sibling of).
- #3488 — Gitea `forge_get_required_status_check_contexts` support, referenced by the same UNSTABLE-fallback block.

rjwalters:
## Summary

`merge-pr.sh <PR> --auto` hard-failed (exit 1, `Pull request is in unstable status`) when a PR's merge state was UNSTABLE **solely** because its required checks were still `queued`/`in_progress`. Those runs have `conclusion == null`, so they never populated `_UNSTABLE_FAILING`, and the #3486 UNSTABLE-fallback's defensive "no failing checks enumerated" branch (`merge-pr.sh:378-382`) converted the whole point of `--auto` (wait for CI) into an immediate failure — forcing the caller to poll manually and re-invoke once the state went CLEAN.

## Fix

Extends the existing UNSTABLE-fallback block into a **bounded poll** (option (a) from the curated issue — pure-bash, co-located with the bug, and the fallback fires regardless of whether `loom-auto-merge` or the shell `forge_auto_merge` was used). Each poll tick classifies the head-SHA check-runs in precedence order:

| Case | Condition | Action |
|------|-----------|--------|
| (a) | A branch-protection-**required** check has **failed** | refuse now — terminal error, **no** waiting out the timeout |
| (b) | Any check still `queued`/`in_progress` (new PENDING set: `status != "completed"`) | wait, bounded by `LOOM_AUTO_MERGE_TIMEOUT` |
| (c) | Failing checks all **informational**, nothing pending | immediate-merge fallback (#3486, unchanged) |
| (d) | Checks resolved green after waiting | immediate merge (now effectively CLEAN) |
| (e) | Nothing failing, nothing pending, never observed a pending check (e.g. commit-status gap) | preserve the #3486 defensive hard-error |

The poll reuses `LOOM_AUTO_MERGE_POLL_INTERVAL` / `LOOM_AUTO_MERGE_TIMEOUT` (defaults 30s / 600s) so both forges share the auto-merge poll configuration already exposed by `loom-auto-merge`. Timeout, required-failure, and unknown-gap each surface a **distinct** non-zero error. A concurrent merge detected mid-wait is handled as a clean exit.

The #3371 CLEAN-fallback, the #3486 informational-failure fallback, and the unknown-gap hard-error paths are all unchanged.

## Testing

- `bash defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` — **51/51 pass** (adds PENDING/FAILING jq-filter classification, poll-decision precedence, and poll-window env-var wiring).
- `bash defaults/scripts/tests/test-merge-pr-help.sh` — 15/15 pass.
- `bash defaults/scripts/tests/test-merge-pr-worktree-path.sh` — 21/21 pass.
- `bash -n` + `shellcheck -S warning defaults/scripts/merge-pr.sh` — clean.

No Python changed (option (b) not needed).

Closes #3664

### DIFF
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -298,6 +298,13 @@ if [[ "$AUTO_MERGE" == "true" ]]; then
   MERGE_RETRY_DELAY=5
   AUTO_MERGE_OK=false
 
+  # Bounded poll window for the UNSTABLE-because-checks-are-still-running case
+  # (#3664). Reuses the same env-var names/semantics as the Gitea auto-merge
+  # poller (loom-tools/src/loom_tools/auto_merge.py) so both forges share
+  # configuration. Defaults match that CLI: 30s interval, 600s ceiling.
+  LOOM_AUTO_MERGE_POLL_INTERVAL="${LOOM_AUTO_MERGE_POLL_INTERVAL:-30}"
+  LOOM_AUTO_MERGE_TIMEOUT="${LOOM_AUTO_MERGE_TIMEOUT:-600}"
+
   for MERGE_ATTEMPT in $(seq 1 $MAX_MERGE_RETRIES); do
     AUTO_MERGE_OUTPUT=""
     # Prefer loom-auto-merge CLI (forge-agnostic, with poll-and-merge for Gitea)
@@ -350,17 +357,32 @@ if [[ "$AUTO_MERGE" == "true" ]]; then
     fi
 
     # PR is UNSTABLE — GitHub's enablePullRequestAutoMerge mutation rejects this
-    # state with "Pull request Pull request is in unstable status" when one or
-    # more rollup checks are red. The auto-merge call would deadlock indefinitely
-    # waiting for those checks to go green, but if every failing check is
-    # informational (NOT in branch protection's requiredStatusCheckContexts)
-    # the immediate-merge path would succeed. Detect that case and fall through;
-    # otherwise preserve the existing UNSTABLE refusal so we never bypass a
-    # genuinely required check. Sibling of the CLEAN-fallback above. See #3486.
+    # state with "Pull request Pull request is in unstable status". GitHub emits
+    # the SAME string whether the rollup is red (a check FAILED) or merely yellow
+    # (checks still QUEUED/IN_PROGRESS). We resolve the PR's head-SHA check-runs
+    # and distinguish, in precedence order:
+    #
+    #   (a) A required check has genuinely FAILED  -> refuse (terminal error),
+    #       without waiting out the pending timeout.
+    #   (b) A check is still QUEUED/IN_PROGRESS    -> the merge state will settle
+    #       (conclusion == null, so it never shows    on its own; poll until it
+    #       up as "failing"). This is the #3664       resolves to (a)/(c)/CLEAN,
+    #       "checks still running" case.               bounded by
+    #                                                   LOOM_AUTO_MERGE_TIMEOUT.
+    #   (c) Every FAILED check is informational    -> immediate-merge fallback
+    #       (NOT in branch protection) and nothing     (#3486, unchanged).
+    #       is pending.
+    #   (d) Nothing failed, nothing pending, and   -> genuine "unknown gap"
+    #       we never observed a pending check          (e.g. commit-status, not
+    #       (e.g. commit-status failures the           check-run, failures) ->
+    #       check-runs API omits).                     refuse (terminal),
+    #                                                   preserving the #3486
+    #                                                   defensive hard-error.
+    #
+    # Once the checks we waited on all pass, the PR is effectively CLEAN and we
+    # fall through to immediate merge (mirroring the CLEAN-fallback above).
+    # Sibling of the CLEAN-fallback above. See #3371, #3486, #3664.
     if echo "$AUTO_MERGE_OUTPUT" | grep -q "is in unstable status"; then
-      # Resolve the failing check names from the rollup against the PR's head
-      # SHA, then compute (failing_checks) \ (required_contexts). If the set
-      # difference equals failing_checks, every failure is informational.
       _UNSTABLE_HEAD_SHA="$(echo "$PR_JSON" | jq -r '.head.sha // empty')"
       _UNSTABLE_BASE_REF="$(echo "$PR_JSON" | jq -r '.base.ref // empty')"
       if [[ -z "$_UNSTABLE_HEAD_SHA" ]] || [[ -z "$_UNSTABLE_BASE_REF" ]]; then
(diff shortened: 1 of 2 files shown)


rjwalters:
LGTM — approved.

Reviewed the diff and the modified block in `.loom/worktrees/issue-3664/defaults/scripts/merge-pr.sh` (lines 385-509). Ran all three suites from the PR worktree: **51/51**, **15/15**, **21/21** pass; `bash -n` and `shellcheck -S warning` both clean. CI checks (Shellcheck, Installer Integration Tests, Detect Changes) pass and mergeStateStatus is CLEAN.

What I verified:
- **PENDING/FAILING jq classification is disjoint and correct.** FAILING selects terminal non-success conclusions (`failure/timed_out/cancelled/action_required`); PENDING selects `status != "completed"` (queued/in_progress, conclusion still null). A rollup UNSTABLE solely because required checks are still running has an empty FAILING set and a populated PENDING set — exactly the #3664 case that previously hit the defensive hard-error.
- **Five-way precedence is correct.** (a) required-check failure (`_UNSTABLE_OVERLAP` non-empty) refuses immediately with no pointless wait; (b) any pending → bounded poll; (c) informational-only failure with nothing pending → immediate-merge fallback (#3486 preserved, verified by the `(c)` and `(a')` tests); informational failure + still-pending correctly falls through to the wait rather than merging early (the `(b')` test); (d) observed-pending-then-green → immediate merge; (e) never-pending, nothing-failing → the #3486 defensive hard-error is preserved.
- **Timeout/interval wiring** reuses `LOOM_AUTO_MERGE_POLL_INTERVAL`/`LOOM_AUTO_MERGE_TIMEOUT` (30s/600s defaults) matching the Gitea auto-merge poller; deadline computed once, enforced in the pending branch.
- **Terminal cases each surface a distinct non-zero error** (required-failure, timeout, lookup-failure, unknown-gap), and every `while`-loop `break` sets exactly one of `_UNSTABLE_FALLBACK_TO_MERGE`/`AUTO_MERGE_OK`, so control never falls through to the generic line-511 error from the UNSTABLE branch.
- **Concurrent-merge-mid-wait** is handled cleanly (recheck `.merged` before each sleep → clean exit).
- **Fail-closed** on the required-status-check lookup (nonzero rc → refuse) is retained.

Minor non-blocking observation (no change required): a transient `forge_get_check_runs` failure *after* pending was observed returns the empty-fallback JSON, which reads as case (d) 'resolved green' → immediate merge. On a protected branch GitHub enforces required checks server-side so the immediate merge would be rejected anyway, and on an unprotected branch there is nothing to bypass — so the blast radius is bounded. Worth a follow-up only if you want the poll to treat a mid-wait empty rollup as 'keep waiting' rather than 'green'.

Nice work: the comment block documenting the (a)-(e) decision table and the mirrored test coverage for both jq filters and the decision precedence make this easy to reason about.

PR #3670, opened 2026-07-21
fix(install): generalize --quick auto-stash restore beyond .gitignore (#3663)

### LINKED ISSUE #3663: install.sh --quick: auto-stash can't restore staged/unstaged split when the update edits a file the user has uncommitted (e.g. CLAUDE.md)
## Context
Hit while running a repo hygiene pass that updated two tools in sequence (Repo Skills, then Loom) with uncommitted working-tree changes present.

`loom install.sh --quick` protects uncommitted changes by stashing them, applies the update, then restores via `git stash pop --index`. When the update **edits a file the user also has uncommitted changes to** — in this case `CLAUDE.md` (the Loom install rewrites its `<!-- BEGIN LOOM ORCHESTRATION -->` block, and the user had an uncommitted edit to the adjacent `<!-- BEGIN REPO-SKILLS -->` block) — the `pop --index` conflicts. The installer then leaves the user's changes stranded in a stash and prints a manual-recovery message:

```
Note: the restore preserves your original staged/unstaged split (git stash pop --index).
That split could not be reproduced automatically here, so recover by hand ...
Your changes are preserved in the stash (stash@{0}).
```

It's non-destructive (nothing is lost — the stash is kept), but the user is left to hand-reconcile, and it's easy to mistake the stranded stash for real pending work.

## Repro
1. In a target repo with Loom installed, make an uncommitted edit to `CLAUDE.md` (any line near the managed blocks).
2. Run `loom install.sh --quick -y /path/to/target` from a newer Loom source.
3. Observe the `stash pop --index` conflict → changes left in a new stash + manual-recovery message.

- [ ] Detect the overlap and attempt a 3-way merge (`git stash show -p | git apply --3way`) automatically, OR
- [ ] Narrow the protective stash to only paths the update will NOT touch, so the restore can't conflict, OR
- [ ] At minimum, single out exactly which files conflicted (vs. a generic "recover by hand") so the user knows the blast radius.
- [ ] Ensure the pre-existing user stash isn't ambiguously interleaved with the install's protective stash.

Filed by /repo:followups.

---

### Root cause (grounded in current `main`)

`install.sh`'s `--quick` reinstall path (the `if [[ "$INSTALL_TYPE" == "1" ]]` branch starting around `install.sh:661`) already implements a stash-guard/restore cycle for exactly this class of problem, but only for **one** file: `.gitignore`.

1. **Stash scope** (`install.sh:661-685`): before uninstall, `_emit_loom_owned_dirty_paths` (`scripts/install/stash-scope.sh:53`) computes `dirty ∩ Loom-ownership-set` and stashes only those paths (`git stash push -- "${REINSTALL_OWNED_DIRTY[@]}"`). `CLAUDE.md` is in the Loom ownership set — it's an explicit case in `_emit_loom_ownership_set` (`scripts/install/manifest.sh:175`: `CLAUDE.md|loom.sh) target_path="${rel_path}"`) — so a dirty `CLAUDE.md` correctly gets swept into this stash.
2. **Uninstall → `loom-daemon init`** then rewrites `CLAUDE.md` in place: `loom-daemon/src/init/scaffolding.rs` contains marker-block-aware merge logic (`LOOM_SECTION_START = "<!-- BEGIN LOOM ORCHESTRATION -->"`, `scaffolding.rs:46`) that regenerates the Loom-owned block while trying to preserve surrounding user content — the same shape of operation `update_gitignore` (`loom-daemon/src/init/post_init.rs`) performs for `.gitignore`.
3. **Restore** (`install.sh:813-894`): the code pops the protective stash with `git stash pop --index` so the caller's original staged/unstaged split is reproduced. For `.gitignore` specifically (issue #3588), the code first resets `.gitignore` to its committed `HEAD` state (`git checkout HEAD -- .gitignore`, `install.sh:823-830`) so the 3-way pop base matches what the stash was recorded against, pops cleanly, then calls `reapply_loom_gitignore_patterns()` (`install.sh:249-277`) to re-append the Loom patterns that `init` had just (re-)written — derived by diffing the post-init snapshot against the HEAD version, so it never hardcodes the pattern list.
4. **The gap**: step 3's HEAD-reset-then-reapply treatment is hardcoded to `.gitignore` only (`install.sh:816-830`, `install.sh:862-864`). `CLAUDE.md` — and any other file in the Loom ownership set that `init` rewrites in place and that the user may also have dirtied (e.g. `.claude/settings.json`, role files) — gets none of this treatment. Its stash pop base has moved (init just rewrote the file), so `git stash pop --index` conflicts exactly the way `.gitignore` used to before #3588, and the code falls into the generic conflict-surfacing branch (`install.sh:866-891`) that strands the stash and prints "recover by hand."

In short: **#3663 is the same defect class as #3588, just not yet generalized past `.gitignore`.**

### Implementation guidance

Two viable approaches, in order of recommendation:

**Option A (recommended) — generalize the existing HEAD-reset-then-reapply pattern.**
Extend the `.gitignore`-specific block (`install.sh:813-894`) to iterate over `REINSTALL_OWNED_DIRTY` (already computed at stash time) rather than hardcoding `.gitignore`:
- For each `REINSTALL_OWNED_DIRTY` path that is tracked at `HEAD`, snapshot the post-init on-disk version, `git checkout HEAD -- "$path"` before the pop.
- After a successful `stash pop --index`, "reapply" the Loom-owned delta. `.gitignore`'s `reapply_loom_gitignore_patterns()` is line-append-only and gitignore-specific — it can't be reused verbatim for `CLAUDE.md`, whose Loom content is a **marked block**, not appended lines. The reapply step for `CLAUDE.md` should be block-aware: replace the content between `<!-- BEGIN LOOM ORCHESTRATION -->` / `<!-- END LOOM ORCHESTRATION -->` (or the repo's equivalent end marker — confirm the exact end-marker constant in `loom-daemon/src/init/scaffolding.rs`) with the post-init snapshot's block content, leaving everything the user's stash pop restored (including sections outside the markers, e.g. their `REPO-SKILLS` block in the repro) untouched. This mirrors `reapply_loom_gitignore_patterns`'s "derive from the post-init snapshot, don't hardcode" property but needs a block-replace helper instead of a line-diff-append helper.
- Only files with a HEAD-tracked committed version participate in the reset+reapply; untracked/new files fall through to a plain pop (already handled).

**Option B — 3-way apply fallback (matches the issue's first suggested AC).**
Replace (or fall back to, on `stash pop --index` failure) `git stash show -p | git apply --3way --index`. `git apply --3way` performs a real 3-way merge against the blobs `git stash push` recorded, and on a genuine content collision it writes `<<<<<<<`/`>>>>>>>` conflict markers directly into the working file instead of leaving the diff trapped in an opaque stash — a materially better UX than "recover by hand" even in the worst case, since the user resolves the conflict where the edit actually lives. This approach is file-shape-agnostic (no `.gitignore`-vs-`CLAUDE.md`-specific reapply logic needed) but changes the recovery UX (conflict markers in-file vs. a stash reference) and needs care to preserve the staged/unstaged split (`--index` combined with `--3way`).

Recommend **Option A** for parity/predictability with the existing `.gitignore` fix and because it keeps recovered files conflict-marker-free on the happy path (the common case: user edited only their own section of `CLAUDE.md`, not the Loom block itself). Fall back to something like Option B's conflict-marker surfacing (or the existing "recover by hand" message, improved to name the specific file) only for the genuine-overlap case where the user edited the same lines the Loom block occupies.

Whichever option is chosen, also address the issue's fourth bullet: when the conflict-surfacing branch fires, name the specific file(s) that conflicted (not just a generic message) — this is a small, low-risk addition regardless of which option lands, and directly serves users hitting the manual-recovery path in the interim.

### Affected Files

- `install.sh` — the `--quick` reinstall stash-guard/restore block (~`install.sh:647-894`); the `.gitignore`-specific reset/reapply logic (`install.sh:813-894`, `reapply_loom_gitignore_patterns()` at `install.sh:249-277`) is the pattern to generalize or supplement.
- `scripts/install/stash-scope.sh` — `_emit_loom_owned_dirty_paths()` (already computes the exact path set needed; no change expected unless Option A needs a second helper here).
- `scripts/install/manifest.sh` — confirms `CLAUDE.md` (and other files) are in the Loom ownership set (`_emit_loom_ownership_set`, ~`scripts/install/manifest.sh:175`); reference only, unlikely to need changes.
- `loom-daemon/src/init/scaffolding.rs` — owns the `CLAUDE.md` marker-block rewrite logic (`LOOM_SECTION_START`, ~line 46) that install.sh's reapply step needs to mirror; reference for the exact begin/end marker constants and block semantics.
- Similar reinstall stash-guard in `scripts/install-loom.sh --clean` may share this defect class (see comment at `install.sh:652-653` noting it "performs its own guarding" — worth a quick check, but out of scope unless it reuses `install.sh`'s pop logic).

### Test Plan

- [ ] Manual verification (repro from the issue): make an uncommitted edit to `CLAUDE.md` outside the Loom-managed block (e.g. edit an adjacent user section like `REPO-SKILLS`), run `install.sh --quick -y <target>` from a newer Loom source, confirm the edit survives in the working tree with no stranded stash and no manual-recovery message.
- [ ] Genuine-overlap case: edit a line *inside* the `<!-- BEGIN LOOM ORCHESTRATION -->` block itself, run `--quick` reinstall, confirm the conflict is surfaced clearly (named file, and either conflict markers in-file or an unambiguous stash reference) rather than a generic "recover by hand."
- [ ] Regression: re-run the existing `.gitignore` conflict scenario from #3588 to confirm no regression in that path (its behavior should be preserved or subsumed by the generalized logic).
- [ ] Staged/unstaged split preserved: stage part of the `CLAUDE.md` edit and leave part unstaged before the reinstall; confirm `git status` after restore shows the same split (this is the property `--index` exists to guarantee).
- [ ] Automated tests: add/extend a bats or shell test under `scripts/install/tests/` (or wherever the existing `.gitignore` reinstall-stash coverage lives — check for existing tests exercising `install.sh --quick` reinstall stash/pop) covering the `CLAUDE.md`-dirty-during-reinstall case end to end.
- [ ] Edge cases: untracked file in the Loom ownership set (no `HEAD` version to reset to), a repo where `CLAUDE.md` doesn't yet exist at `HEAD` (fresh install, not reinstall — should be unaffected), and multiple simultaneously-dirty Loom-owned files in one reinstall.

rjwalters:
## Summary

`install.sh --quick` guards uncommitted user changes across the uninstall→reinstall cycle by stashing Loom-owned dirty paths, then restoring with `git stash pop --index`. The #3588 fix that lets this survive a file `init` rewrites in place — HEAD-reset the file before popping so the 3-way base lines up, then reapply the fresh Loom content — was **hardcoded to `.gitignore`**. When the update also rewrote a different Loom-owned file the user had dirtied — canonically `CLAUDE.md` (a marker-delimited Loom block with user content around it) — the pop conflicted, stranded the change in a stash, and printed a generic "recover by hand" message. This is the same defect class as #3588, just not yet generalized.

## What changed

- **`install.sh`**: the `--quick` restore block now iterates the already-computed `REINSTALL_OWNED_DIRTY` set and applies the HEAD-reset-then-reapply treatment to every file with a well-defined Loom-vs-user split — `.gitignore` (Loom patterns appended) and `CLAUDE.md` (marker block).
- New `reapply_loom_claude_md_block()`: a block-aware splice that replaces only the `<!-- BEGIN LOOM ORCHESTRATION -->` … `<!-- END LOOM ORCHESTRATION -->` region with the daemon's freshly-written block from the post-init snapshot, leaving the user's content outside the markers byte-for-byte intact. Derived from the snapshot, never hard-coded (mirrors `reapply_loom_gitignore_patterns`).
- HEAD-reset is scoped to files that have a reapply strategy. A fully Loom-owned file (role `.md`, `config.json`) has no partial reapply, so it falls through to a plain pop rather than resetting-and-losing the update. For `CLAUDE.md`, when the user's stashed edit lands **inside** the block (reapply would clobber it) or HEAD carries no block to splice, the reset is skipped so the plain-pop 3-way merge surfaces a genuine conflict instead of silently losing data.
- The conflict-surfacing branch now **names the specific conflicting file(s)** (`Conflicting file(s): CLAUDE.md`) instead of a generic message.

## Behavior matrix (CLAUDE.md)

| User edit | `init` changes block? | Outcome |
|-----------|----------------------|---------|
| Outside the Loom block | — | Edit preserved, block refreshed, stash consumed, no message |
| Inside the block | No | Applies cleanly (one-sided 3-way), no data loss |
| Inside the block | Yes | Conflict surfaced + named, stash preserved, edit recoverable, installer exits 0 |

Staged/unstaged split is preserved throughout (`pop --index`).

## Tests

Extended the daemon-backed stash regression suite (`scripts/test-install-quick-gitignore-stash.sh`) with six CLAUDE.md scenarios covering all rows above plus multiple-dirty-files. Results:

- `test-install-quick-gitignore-stash.sh`: **46/46** (27 existing #3588/#3611 + 19 new #3663)
- `test-installer.sh` (CI): **121/121**
- `test-install-stash-scope.sh`: **7/7**

Closes #3663

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -276,6 +276,80 @@ reapply_loom_gitignore_patterns() {
   done <<<"$loom_lines"
 }
 
+# Issue #3663: re-apply the current Loom-owned CLAUDE.md marker block after a
+# --quick reinstall stash pop that was performed against a HEAD-reset CLAUDE.md.
+#
+# This generalizes the #3588 .gitignore treatment (HEAD-reset-before-pop, then
+# reapply) to CLAUDE.md, whose Loom content is a marker-delimited block
+# (`<!-- BEGIN LOOM ORCHESTRATION -->` … `<!-- END LOOM ORCHESTRATION -->`)
+# rather than a set of appended lines. The reinstall restores CLAUDE.md to its
+# committed HEAD state before popping so the user's stashed hunk applies cleanly
+# (see the pop block below). That reset reverts the Loom block to its old
+# committed content, so we splice the freshly written block back in here.
+#
+# We rebuild the file as: everything BEFORE the begin marker (from the popped
+# working copy — the user's restored content) + the block (begin…end inclusive)
+# taken from the post-init snapshot (the daemon's authoritative fresh block) +
+# everything AFTER the end marker (again from the popped working copy). Only the
+# delimited Loom region is replaced; every line the user's pop restored outside
+# the markers (e.g. their own `REPO-SKILLS` block) is left byte-for-byte intact.
+# Derived from the snapshot, never hard-coded, mirroring
+# reapply_loom_gitignore_patterns's "trust the daemon's output" property.
+# Idempotent: splicing an already-current block is a no-op. If either file lacks
+# both markers there is no delimited region to reconcile, so the popped file is
+# left as-is.
+reapply_loom_claude_md_block() {
+  local target_path="$1"
+  local postinit_snapshot="$2"
+  local claude_md="$target_path/CLAUDE.md"
+  local begin="<!-- BEGIN LOOM ORCHESTRATION -->"
+  local end="<!-- END LOOM ORCHESTRATION -->"
+
+  [[ -f "$postinit_snapshot" && -f "$claude_md" ]] || return 0
+
+  # Both the snapshot and the popped file must carry the marker block, else
+  # there is no delimited Loom region to splice — leave the popped file alone.
+  grep -qF "$begin" "$postinit_snapshot" && grep -qF "$end" "$postinit_snapshot" || return 0
+  grep -qF "$begin" "$claude_md" && grep -qF "$end" "$claude_md" || return 0
+
+  local tmp
+  tmp="$(mktemp 2>/dev/null || true)"
+  [[ -n "$tmp" ]] || return 0
+
+  # Pass 1 (snapshot): capture the begin…end block into `block`.
+  # Pass 2 (popped file): print user content up to begin, emit the captured
+  # block once, then resume printing user content after end.
+  if awk -v b="$begin" -v e="$end" '
+    FNR==NR {
+      if (index($0, b)) grab=1
+      if (grab) block = block $0 ORS
+      if (index($0, e)) grab=0
+      next
+    }
+    {
+      if (index($0, b)) { printf "%s", block; skip=1 }
+      if (!skip) print
+      if (index($0, e)) skip=0
+    }
+  ' "$postinit_snapshot" "$claude_md" >"$tmp" 2>/dev/null && [[ -s "$tmp" ]]; then
+    cat "$tmp" >"$claude_md" 2>/dev/null || true
+  fi
+  rm -f "$tmp" 2>/dev/null || true
+  return 0
+}
+
+# Issue #3663: emit the Loom marker block (begin…end inclusive) from stdin.
+# Used to decide whether a user's stashed CLAUDE.md edit lands INSIDE the Loom
+# block (in which case a HEAD-reset+reapply would clobber it) or entirely
+# outside it (safe to reset+reapply). Empty output means no block was found.
+_emit_loom_claude_block() {
+  awk '
+    index($0, "<!-- BEGIN LOOM ORCHESTRATION -->") { inblk=1 }
+    inblk { print }
+    index($0, "<!-- END LOOM ORCHESTRATION -->") { inblk=0 }
+  '
+}
+
 # Determine Loom repository root
 LOOM_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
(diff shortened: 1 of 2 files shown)


rjwalters:
LGTM — approving.

## Review summary

Reviewed the diff and read the full `install.sh` restore block plus the extended test suite. This cleanly generalizes the #3588 `.gitignore` HEAD-reset-then-reapply pattern to all Loom-owned dirty files with a per-file reapply strategy, and correctly preserves the load-bearing safety invariants.

### Correctness — reset-iff-reapplyable invariant holds
- HEAD-reset is scoped to files with a reapply strategy (`.gitignore` append, `CLAUDE.md` marker-block splice). Fully Loom-owned files (role `.md`, `config.json`) hit the `*) continue` arm and fall through to a plain pop rather than being reset-and-lost — the update is never dropped.
- For `CLAUDE.md`, reset only fires when (a) HEAD carries a block to splice (`_head_block` non-empty) AND (b) the stashed block region equals HEAD's block region (`_stashed_block == _head_block`), i.e. the user's edit is strictly outside the block. In-block edits skip the reset so the plain-pop 3-way merge surfaces a genuine conflict instead of silently clobbering. This is the crux of the fix and it is right.
- The `reapply_loom_claude_md_block` awk splice (pass 1 captures snapshot block, pass 2 replaces the popped file's begin…end region) replaces only the delimited region; user content outside the markers is byte-for-byte preserved. Guarded by dual-marker presence checks on both files and a `[[ -s $tmp ]]` non-empty guard, so a degenerate awk/mktemp failure is non-destructive.

### Edge cases handled
- Markerless HEAD → `continue` (no reset), init's freshly-appended block survives the out-of-block pop.
- Untracked/new `CLAUDE.md` → `cat-file -e HEAD:...` fails → skipped.
- Conflict path names the specific file(s) via `git diff --name-only --diff-filter=U` with a fallback to `REINSTALL_OWNED_DIRTY`, satisfying issue #3663's minimum AC. Installer still exits 0 and prints the completion banner; stash preserved for recovery.
- Temp snapshots cleaned up in success, conflict, and teardown branches.

### Shell quality
- `set -euo pipefail` respected: `${REINSTALL_OWNED_DIRTY[@]+"..."}` safe-expansion idiom for possibly-empty arrays, `|| continue` / `|| true` guards on every fallible git/mktemp call, parallel arrays (paths/snapshots/strategies) indexed consistently.

### Tests — verified locally
- `scripts/test-install-quick-gitignore-stash.sh`: **46/46** (6 new #3663 scenarios: outside-block survival, block refresh from snapshot, staged/unstaged split, genuine in-block conflict + named file, in-block clean, multi-file dirty).
- `scripts/test-installer.sh`: **121/121**.
- `defaults/scripts/tests/test-install-stash-scope.sh`: **7/7**.
- CI `Installer Integration Tests`: pass. Merge state: CLEAN.

Scenario coverage maps 1:1 to the behavior matrix and to issue #3663's acceptance criteria. Nice work.

Issue #1113, opened 2026-01-24
Add Comparative Analysis UI for experiments


rjwalters:
## Problem

The A/B testing framework (#1071) exists in the backend, but there's no UI for users to:
- View experiment results
- Compare different approaches side-by-side
- Analyze which configurations work best
- Make data-driven decisions about agent settings

## Solution

Create a Comparative Analysis UI that allows users to:
- View A/B experiment results with statistical analysis
- Compare agent configurations (manual vs autonomous, different models)
- Benchmark against historical baselines
- Visualize trade-offs (cost vs speed vs quality)

### Experiment Results View
- List of completed experiments
- Summary statistics per variant (success rate, cost, time)
- Statistical significance indicators
- Winner recommendation with confidence level

### Comparison Tools
- Side-by-side metrics comparison
- Scatter plots (cost vs quality, speed vs quality)
- Historical trend comparisons

### Data Sources
- `ab_testing.rs` experiment data
- `get_experiment_results` - Results for specific experiment
- `list_experiments` - All experiments with status
- Historical metrics for baseline comparison

### UI Components
- Experiment list with status (running, complete, significant)
- Results dashboard with comparison tables
- Visualization charts (bar charts, scatter plots)
- Recommendation panel with explanation

## Acceptance Criteria

- [ ] List all A/B experiments with status
- [ ] View experiment results with metrics per variant
- [ ] Statistical significance indicator (p-value or confidence)
- [ ] Side-by-side comparison table
- [ ] Bar chart comparing key metrics
- [ ] Winner recommendation with explanation
- [ ] Create new experiment from UI
- [ ] Compare current performance to historical baseline
- [ ] Export experiment results as CSV/JSON
- [ ] Tests for results display and significance calculation

## Mockup Reference

See #444 "Feature 6: Comparative Analysis" section for detailed mockup.

## Dependencies

- Depends on: #1071 (A/B testing framework) - complete
- Related: #444 (parent epic)

---
Part of #444 (Loom Intelligence epic)

PR #3675, opened 2026-07-21
fix(merge): guard porcelain awk helpers against exit-triggers-END double-print (#3671)

### LINKED ISSUE #3671: champion: post-merge worktree-cleanup emits a malformed/duplicated worktree path
## Context

Observed while using the Champion auto-merge on two separate PRs in the same session (botho-project/botho PRs #1113 and #1116). Both times, the post-merge worktree-cleanup step reported a **malformed/duplicated worktree path** for the branch, rendered as the repo root concatenated with itself across a newline, e.g.:

```
/Users/rwalters/GitHub/botho
/Users/rwalters/GitHub/botho
```

The Champion then correctly **declined** to remove it (no `.loom-managed` sentinel — it's the user's main checkout, not a Loom worktree), so behavior was safe and the remote branch was still cleaned up server-side. The bug is cosmetic but indicates a string-parsing issue in how the cleanup step reads `git worktree list` (likely splitting/joining the porcelain or default output and duplicating the path, or failing to isolate the worktree line).

Consistent across both runs, so reproducible, not a one-off.

- [ ] The champion parses `git worktree list` (prefer `--porcelain`) robustly so the reported path is a single, correct worktree path.
- [ ] No malformed/duplicated path string appears in the post-merge cleanup log.
- [ ] The safety behavior is unchanged — a non-`.loom-managed` worktree (e.g. the user's main checkout) is still never removed.

---

## Root Cause (curator investigation)

Confirmed on `origin/main` (identical file content, no drift). The bug is a classic **awk `exit`-triggers-`END`** gotcha in `_find_worktree_by_branch()`:

`defaults/scripts/merge-pr.sh:691-699`

```bash
_find_worktree_by_branch() {
  local want_branch="$1"
  git -C "$REPO_ROOT" worktree list --porcelain 2>/dev/null | \
    awk -v want="refs/heads/${want_branch}" '
      /^worktree / { wt=$2; br=""; next }
      /^branch /   { br=$2 }
      /^$/         { if (br == want) { print wt; exit } }
      END          { if (br == want) { print wt } }
    '
}
```

In POSIX/gawk/mawk/nawk semantics, calling `exit` from a main-body rule does **not** skip the `END` block — it jumps straight to it (unless `exit` is itself called from within `END`). Here the blank-line rule matches the target stanza, prints the worktree path, and calls `exit`; control transfers to `END`, whose condition (`br == want`) is *still true* because `br`/`want` are ordinary awk variables that persist across the jump. So `END` prints the same path a second time. The result: `$(...)` command substitution captures two identical lines joined by `\n` — exactly the reported `/path\n/path` output.

This function is invoked from the champion/merge-pr.sh post-merge worktree-cleanup discovery fallback:

`defaults/scripts/merge-pr.sh:804-824` (`if [[ -d "$DEFAULT_WT_PATH" ]] ... else` branch) — when neither the `.loom/worktrees/issue-<N>/` nor `.loom/worktrees/pr-<N>/` convention path exists (e.g. the PR branch was checked out directly in the operator's main checkout, as in the two reported runs), the script falls back to:

```bash
DISCOVERED_WT="$(_find_worktree_by_branch "$PR_BRANCH")"
...
warning "Discovered worktree for branch '$PR_BRANCH' at: $DISCOVERED_WT"
```

`$DISCOVERED_WT` now contains the doubled path, which is what gets printed in the warning. Since there is no `.loom-managed` sentinel at that path, the safety-critical branch (`warning "Worktree lacks .loom-managed sentinel — not removing (user-owned)."`) still correctly declines removal — the duplication is purely cosmetic in the log/warning text, not a safety issue.

**A second instance of the identical anti-pattern** exists in the sibling helper `_worktree_branch_for()` at `defaults/scripts/merge-pr.sh:676-686` (used by `_remove_loom_worktree` when `allow_unmanaged == "true"`, i.e. the `--worktree-path` explicit-opt-in cleanup path, to look up the branch attached to a worktree before removing it). Same `print ...; exit` in the blank-line rule followed by an `END` block re-testing `br == want` and printing again — this would double the branch name output (e.g. `feature/issue-42\nfeature/issue-42`) fed into `_maybe_delete_local_branch`. It happens to be more benign there because `git branch -d "$branch"` on a doubled/malformed multi-line string would just fail the `show-ref --verify` check and skip cleanup with a warning rather than corrupt anything — but it's the same bug and should be fixed alongside `_find_worktree_by_branch`.

Reproduced locally: piping a two-stanza synthetic `git worktree list --porcelain` block through the current `_find_worktree_by_branch` awk body reliably emits the matched path twice, joined by `\n` — matching the reported output byte-for-byte.

**Not already fixed**: `git diff origin/main -- defaults/scripts/merge-pr.sh` is empty (local `main` already tracks `origin/main`); `git log --oneline -- defaults/scripts/merge-pr.sh` shows no commit addressing this since the file's `_find_worktree_by_branch`/`_worktree_branch_for` helpers were introduced (#3358 discovery fallback).

## Recommended Fix

Guard against the double-print with a `found` flag so `exit` reliably represents "already emitted" and `END` becomes a no-op once the main body has already printed. Apply the same fix to both helpers:

```bash
_find_worktree_by_branch() {
  local want_branch="$1"
  git -C "$REPO_ROOT" worktree list --porcelain 2>/dev/null | \
    awk -v want="refs/heads/${want_branch}" '
      /^worktree / { wt=$2; br=""; next }
      /^branch /   { br=$2 }
      /^$/         { if (br == want && !found) { print wt; found=1; exit } }
      END          { if (br == want && !found) { print wt } }
    '
}

_worktree_branch_for() {
  local target="$1" target_abs
  target_abs="$(cd "$target" 2>/dev/null && pwd -P)" || target_abs="$target"
  git -C "$REPO_ROOT" worktree list --porcelain 2>/dev/null | \
    awk -v p="$target_abs" '
      /^worktree / { wt=$2; br=""; next }
      /^branch /   { br=$2 }
      /^$/         { if (wt == p && br != "" && !found) { sub(/^refs\/heads\//, "", br); print br; found=1; exit } }
      END          { if (wt == p && br != "" && !found) { sub(/^refs\/heads\//, "", br); print br } }
    '
}
```

This mirrors the pattern already used correctly elsewhere in the same file at `defaults/scripts/merge-pr.sh:242-243` (the `--worktree-path` validator), which uses a `found` flag and never double-prints because it only sets an exit status, not stdout, from `END`.

An equally valid alternative is to drop the mid-body `print`/`exit` entirely and always print from `END` only (relying on `next`-based stanza parsing to naturally stop consuming further input past the match via a `found` guard on each rule) — but the flag-guard above is the minimal, lowest-risk diff.

## Affected Files
- `defaults/scripts/merge-pr.sh` (lines 676-699: `_worktree_branch_for`, `_find_worktree_by_branch`)

## Acceptance Criteria
- [ ] `_find_worktree_by_branch()` and `_worktree_branch_for()` in `defaults/scripts/merge-pr.sh` never emit a duplicated/doubled line for a single matching worktree stanza (single `print`, guarded against re-entry via `END`).
- [ ] No malformed/duplicated path or branch-name string appears in post-merge cleanup warnings/logs (`Discovered worktree for branch ... at: <path>`).
- [ ] Safety behavior is unchanged: a worktree/branch lacking the `.loom-managed` sentinel is still never removed; `--worktree-path` explicit opt-in cleanup still works.
- [ ] Existing `merge-pr.sh` test coverage (if any) continues to pass; add/extend a regression test that pipes a synthetic multi-stanza `git worktree list --porcelain` block through `_find_worktree_by_branch` (and `_worktree_branch_for`) and asserts exactly one line of output for a matching branch — see `defaults/scripts/tests/` for existing merge-pr.sh test conventions (e.g. `test-worktree-sentinel-reinvoke.sh`, `test-worktree-root-override.sh`).

## Test Plan
1. Unit-level: extract or directly invoke the two awk snippets (before/after fix) against a synthetic two-stanza porcelain block (one worktree with a matching branch, one without) and assert output is exactly one line, matching the expected path/branch.
2. Regression test script under `defaults/scripts/tests/` (bash `set -euo pipefail`, matching existing test file conventions) that sources or re-implements the fixed function and checks single-line output for:
   - A match found mid-list (not last stanza) — exercises the `exit`-triggers-`END` path.
   - A match found in the last stanza with no trailing blank line — exercises the `END`-only path.
   - No match at all — asserts empty output.
3. Manual/integration sanity: run `./.loom/scripts/merge-pr.sh <pr> --dry-run` (or exercise the discovery-fallback branch directly) against a repo where the PR branch is checked out in the main checkout (not a `.loom/worktrees/` path) and confirm the warning message shows a single, correct path.
4. Confirm `--worktree-path` explicit-opt-in cleanup (the `allow_unmanaged == "true"` path that calls `_worktree_branch_for`) still resolves and deletes the correct single branch name after the fix.

rjwalters:
## Summary

Fixes the malformed/duplicated worktree path emitted by merge-pr.sh's post-merge cleanup step (issue #3671, observed on botho PRs #1113/#1116).

Both `_find_worktree_by_branch()` and `_worktree_branch_for()` hit the classic awk `exit`-triggers-`END` gotcha: the blank-line rule prints the matched value and calls `exit`, but in POSIX/gawk/mawk/nawk semantics `exit` from a main-body rule jumps to the `END` block rather than skipping it. `END`'s condition (`br == want` / `wt == p`) is still true because the awk variables persist across the jump, so it prints the same value a second time. Command substitution then captures a doubled `/path\n/path` string.

The fix adds a `found`-flag guard to both the blank-line rule and the `END` block so `exit` reliably means "already emitted" and `END` is a no-op once the main body has printed — mirroring the pattern already used correctly by the `--worktree-path` validator (`found=1; exit` + `exit !found`).

The bug was cosmetic (the `.loom-managed` sentinel guard still correctly declined removal of user-owned worktrees) but produced incorrect warning text.

## Changes

- `defaults/scripts/merge-pr.sh` — add `!found` guard + `found=1` to the blank-line rule and `!found` guard to the `END` block in both helpers (4 lines).
- `defaults/scripts/tests/test-merge-pr-worktree-awk.sh` — new regression suite (11 assertions).

## Test results

All merge-pr suites pass:
- test-merge-pr-worktree-awk.sh (new): 11/11
- test-merge-pr-worktree-path.sh: 21/21
- test-merge-pr-unstable-fallback.sh: 51/51
- test-merge-pr-partial-increment.sh: 18/18
- test-merge-pr-help.sh: 15/15

The new suite covers: static grep that the guard is present in both helpers; single-line output for a mid-list match (exit-triggers-END path), a last-stanza match with no trailing blank (END-only path), and no-match (empty output); and a fixture-sanity check confirming the pre-fix awk still doubles (so the fixtures genuinely exercise the bug). shellcheck-clean.

Closes #3671

### DIFF
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -806,8 +806,8 @@ _worktree_branch_for() {
     awk -v p="$target_abs" '
       /^worktree / { wt=$2; br=""; next }
       /^branch /   { br=$2 }
-      /^$/         { if (wt == p && br != "") { sub(/^refs\/heads\//, "", br); print br; exit } }
-      END          { if (wt == p && br != "") { sub(/^refs\/heads\//, "", br); print br } }
+      /^$/         { if (wt == p && br != "" && !found) { sub(/^refs\/heads\//, "", br); print br; found=1; exit } }
+      END          { if (wt == p && br != "" && !found) { sub(/^refs\/heads\//, "", br); print br } }
     '
 }
 
@@ -820,8 +820,8 @@ _find_worktree_by_branch() {
     awk -v want="refs/heads/${want_branch}" '
       /^worktree / { wt=$2; br=""; next }
       /^branch /   { br=$2 }
-      /^$/         { if (br == want) { print wt; exit } }
-      END          { if (br == want) { print wt } }
+      /^$/         { if (br == want && !found) { print wt; found=1; exit } }
+      END          { if (br == want && !found) { print wt } }
     '
 }
 
--- a/defaults/scripts/tests/test-merge-pr-worktree-awk.sh
+++ b/defaults/scripts/tests/test-merge-pr-worktree-awk.sh
@@ -1,0 +1,187 @@
+#!/usr/bin/env bash
+# test-merge-pr-worktree-awk.sh - Regression tests for the porcelain-parsing
+# awk helpers in merge-pr.sh (#3671).
+#
+# Verifies that _find_worktree_by_branch() and _worktree_branch_for() emit
+# EXACTLY ONE line for a single matching worktree stanza. Prior to #3671 both
+# helpers hit the classic awk `exit`-triggers-`END` gotcha: the blank-line rule
+# printed the match and called `exit`, but control transferred to the `END`
+# block whose condition was still true, printing the same value a second time.
+# The result was a doubled `/path\n/path` string in the post-merge
+# worktree-cleanup warning (observed on botho PRs #1113 / #1116).
+#
+# The fix adds a `found`-flag guard so `exit` reliably means "already emitted"
+# and `END` becomes a no-op once the main body has printed — mirroring the
+# correct pattern already used by the --worktree-path validator.
+#
+# Test strategy:
+#   1. Static grep assertions that the live source retains the `!found` guard
+#      in both helpers (guards against a regression to the un-guarded awk).
+#   2. Behavioral assertions: pipe synthetic multi-stanza `git worktree list
+#      --porcelain` blocks through awk bodies that mirror the fixed source and
+#      assert exactly-one-line output for:
+#        - a match found mid-list (not last stanza) — exercises the
+#          `exit`-triggers-`END` path (blank line terminates the stanza).
+#        - a match found in the last stanza with NO trailing blank line —
+#          exercises the `END`-only path.
+#        - no match at all — asserts empty output.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPTS_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+MERGE_PR="$SCRIPTS_DIR/merge-pr.sh"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+TESTS_RUN=0
+TESTS_PASSED=0
+TESTS_FAILED=0
+
+pass() { TESTS_RUN=$((TESTS_RUN + 1)); TESTS_PASSED=$((TESTS_PASSED + 1)); echo -e "  ${GREEN}PASS${NC}: $1"; }
+fail() { TESTS_RUN=$((TESTS_RUN + 1)); TESTS_FAILED=$((TESTS_FAILED + 1)); echo -e "  ${RED}FAIL${NC}: $1"; }
+
+assert_grep() {
+    local pattern="$1" file="$2" msg="$3"
+    if grep -qE "$pattern" "$file"; then pass "$msg"; else fail "$msg (pattern: $pattern)"; fi
+}
+
+# assert_single_line <label> <expected> <actual>
+# Asserts $actual equals $expected AND contains no embedded newline (i.e. the
+# helper emitted a single line, not a doubled `/path\n/path` string).
+assert_single_line() {
+    local label="$1" expected="$2" actual="$3"
+    local nlines
+    nlines=$(printf '%s' "$actual" | grep -c '' || true)
+    if [[ "$actual" == "$expected" ]] && [[ "$nlines" -le 1 ]]; then
+        pass "$label (single line: '$actual')"
+    else
+        fail "$label: expected exactly one line '$expected'; got ${nlines} line(s): $(printf '%q' "$actual")"
+    fi
+}
+
+[[ -f "$MERGE_PR" ]] || { echo "ERROR: $MERGE_PR not found" >&2; exit 1; }
+
+# --- awk bodies mirroring the FIXED source (kept in sync via Test 1 grep) ---
+
+# find_wt_by_branch <want_branch>  (reads porcelain on stdin)
+find_wt_by_branch() {
+    awk -v want="refs/heads/$1" '
+      /^worktree / { wt=$2; br=""; next }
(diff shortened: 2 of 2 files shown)


rjwalters:
LGTM — approving.

**Correctness of the awk fix**: The `!found` guard + `found=1` on the blank-line rule makes `exit` reliably mean "already emitted", and the matching `!found` guard on `END` turns it into a no-op once the main body has printed. This directly closes the `exit`-triggers-`END` re-print (awk variables persist across the jump, so `END`'s predicate was still true). Verified all three paths:
- mid-list match (exit path): prints once, END skipped.
- last-stanza-with-no-trailing-blank (END-only path): blank rule never fires, `found` stays 0, END prints once.
- no-match: nothing printed.

**No safety-semantics change**: match predicates (`br == want` / `wt == p && br != ""`) and the `.loom-managed` sentinel guard are untouched — only the emission count changes, mirroring the already-correct `--worktree-path` validator pattern.

**Test quality**: Strong. Static grep pins the guard into both helpers (regression lock), behavioral assertions cover all three paths for both helpers, and the Test 4 sentinel reproduces the pre-fix doubling so the fixtures can't go hollow. shellcheck-clean.

**CI**: Shellcheck + Installer Integration pass; Rust jobs correctly skipped for a shell-only change. mergeStateStatus CLEAN. Ran all five merge-pr suites locally from the worktree: 11/11, 21/21, 51/51, 18/18, 15/15 — all green.

PR #3676, opened 2026-07-21
feat(guards): add `guards.forceScope` branch-aware force-op toggle (symmetric to rmScope)

### LINKED ISSUE #3674: feat(guards): add `guards.forceScope` config toggle for branch-aware force operations (symmetric to rmScope)
## Problem

`defaults/hooks/guard-destructive.sh` treats `git push --force` / `--force-with-lease` and `git reset --hard` as destructive and **asks** (or blocks) on them. For **autonomous / background agents** — which cannot answer an interactive "ask" prompt — this stalls the agent on *routine* operations: force-pushing or hard-resetting their **own single-owner working branch** is a normal part of the rebase/amend/reset workflow, not a dangerous act.

The genuinely dangerous case is a force operation against a **protected/shared branch** (`main`/`master`). That distinction is branch-scoped, but today the guard is branch-agnostic.

## Current consumer workaround (a fork we want to drop)

Consumers (e.g. the lean-genius fleet) patch `guard-destructive.sh` locally to add **branch-aware** logic:

- Force ops (`push --force`, `push -f`, `push --force-with-lease`, `reset --hard`) are **allowed** when the target is a single-owner working branch.
- They are **blocked/asked** when the target is a protected branch (`main`/`master`).
- The target branch is resolved from the push refspec, else the branch checked out where the command runs (`cwd` or `git -C <path>`).

This is a maintained local delta on a Loom-shipped hook — it drifts every release and blocks clean `--clean` reinstalls. We'd like to **drop the fork and drive this from config**.

## Request

Add a **`guards.forceScope`** config toggle, symmetric to the existing **`guards.rmScope`** (#3617). Suggested values:

| `guards.forceScope` | Behavior |
|---------------------|----------|
| `"protected"` (proposed default) | Ask/block force ops only when the resolved target is a protected branch; allow on working branches. Solves the autonomous-agent stall. |
| `"all"` | Current behavior — ask/block every force op regardless of branch. |
| `"off"` | Never guard force ops (opt-out). |

Protected-branch set should reuse Loom's existing default-branch detection (`defaults/scripts/lib/default-branch.sh`) plus any configured protected list.

## Why upstream (not just keep forking)

- `guards.rmScope` already established the pattern and the config plumbing — this is the symmetric sibling for force ops.
- Autonomous/background agents are a first-class Loom use case (the daemon, sweeps); the ask-on-every-force default is hostile to them.
- Keeps consumers on mainstream hooks instead of maintaining per-repo guard forks.

## Acceptance
- `guards.forceScope` honored by `guard-destructive.sh` with the three values above.
- Branch resolution handles both explicit refspecs and the checked-out branch (cwd / `git -C`).
- Documented alongside `guards.rmScope`.

Context: this is the last local hook delta pinning a consumer to a forked `guard-destructive.sh`; landing it lets them run pristine mainstream Loom.

---

### Current behavior (file:line refs, `defaults/hooks/guard-destructive.sh`)

- Force-push flag forms (`--force`, `-f`, `--force-with-lease`) targeting `origin` + a literal `main` or `master` ref are a **hard deny**, unconditional, six entries in `ALWAYS_BLOCK_PATTERNS` (lines 297–303). These substring patterns only match when `main`/`master` appear literally in the command — they are **not** branch-resolved (no refspec parsing, no checked-out-branch fallback). This block is out of scope for `forceScope` — it must remain an unconditional deny regardless of the new toggle, since it's the "genuinely dangerous" case this issue exists to protect.

- The **general** ask-on-every-force-op behavior this issue targets lives in `ASK_PATTERNS` (lines 701–705):
  ```bash
  ASK_PATTERNS=(
      # Git destructive operations (not on main/master - those are blocked above)
      'git push --force'
      'git push -f '
      'git reset --hard'
      ...
  )
  ```
  These are unconditional substring matches with **no branch awareness** — a force-push or hard reset on a single-owner working branch both trigger `ask()` (line 742–746 loop), which stalls a non-interactive/autonomous agent.

### Precedent: `guards.rmScope` (#3628, PR merged) — exact mechanism to mirror

`rm_scope_repo_enabled()` at lines 195–216 is the pattern to copy:

```bash
_RM_SCOPE_CACHE=""
rm_scope_repo_enabled() {
    if [[ -z "$_RM_SCOPE_CACHE" ]]; then
        local mode=repo
        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
            mode=$(jq -r 'if (.guards.rmScope == "off" or .guards.rmScope == "permissive") then "off" else "repo" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || mode=repo
            [[ -n "$mode" ]] || mode=repo
        fi
        case "${LOOM_RM_SCOPE:-}" in
            repo)                  mode=repo ;;
            off|0|no|permissive)   mode=off ;;
        esac
        _RM_SCOPE_CACHE="$mode"
    fi
    [[ "$_RM_SCOPE_CACHE" == "repo" ]]
}
```

Key properties to replicate for `force_scope_mode()`:
- **Lazy** — only invoked after a force-op pattern already matched (config read stays off the hot path for the ~99% of commands that aren't force ops).
- **Cached** in a `_FORCE_SCOPE_CACHE` var (mirrors `_RM_SCOPE_CACHE` / `_SQL_GUARD_CACHE` / `_CLOUD_GUARD_CACHE`).
- **jq `if/then/else`, not `//`** — `//` is alternative-on-null, not default-on-missing, so a malformed/missing key must resolve via explicit `if` branches, never via `//`.
- **Env var overrides config** — symmetric name `LOOM_FORCE_SCOPE` (mirrors `LOOM_RM_SCOPE`), accepting `all` / `protected` / `off`.
- **Malformed JSON / missing REPO_ROOT falls through to the default value** (whichever default is chosen — see "Design point" below) and never trips the `ERR` trap.

Branch resolution has **no existing precedent in this file** — this is genuinely new logic, not a copy of something else. `resolve_worktree_root()` (lines 223–243) is the closest structural analogue (a small pure-bash helper, `.loom/config.json` → env → default precedence) but resolves a path, not a branch. Loom's `defaults/scripts/lib/default-branch.sh::loom_default_branch()` (source-only helper, not exec) is the piece to reuse for the protected-branch identity (see below) — but note it is designed to be **sourced**, not shelled out to per force-op check; `guard-destructive.sh` currently has zero `source` statements, so this will be the guard's first cross-file dependency. Confirm sourcing works from the hook's install location (`.claude/hooks/` post-install, not just `defaults/hooks/` — check how `.loom/scripts/lib/*.sh` are made available at hook-runtime cwd, since `SCRIPT_DIR` in the hook is derived from `BASH_SOURCE`, not `CWD`).

### Recommended implementation sketch

1. **Protected-branch set**: reuse `loom_default_branch()` from `defaults/scripts/lib/default-branch.sh` for the primary default-branch name (main/master detection), plus optionally a configurable `guards.protectedBranches` array (not required for v1 — start with just the default branch to keep scope tight; note as an explicit non-goal if deferred).
2. **Branch resolution** (new function, e.g. `resolve_force_target_branch()`):
   - For `git push` commands: parse the refspec argument. Precedence:
     a. Explicit destination ref in the refspec (`<src>:<dst>` form) — the target is `<dst>` with a leading `+` stripped (`+<branch>` → `<branch>`).
     b. A bare ref with no colon (force-push naming a single ref after the remote) — the target is that ref, `+` stripped.
     c. `HEAD` as the ref, or no ref/refspec argument at all (a bare force-push with only a remote, or no args) — target is the **checked-out branch**, resolved via `git -C <resolved-cwd> symbolic-ref --short HEAD` (mirrors `default-branch.sh`'s use of `symbolic-ref`, offline-first).
   - For `git reset --hard`: there is no remote/refspec — target is **always** the checked-out branch at the resolved cwd.
   - `<resolved-cwd>`: honor a `git -C <path>` prefix in the command string if present (segment-parse like `extract_rm_targets()` / `lifecycle_or_cloud_reason()` do — split on `;|&&|\|\|`, strip leading `sudo`/`env`, look for a `-C <path>` token before the git subcommand), else fall back to the hook's `$CWD` (from `tool_input`/`cwd` JSON field, already resolved at line 61).
   - **Detached HEAD**: `git symbolic-ref --short HEAD` fails (exits non-zero, empty stdout) when detached. Treat this as **unknown branch → do not allow silently**; fall through to the existing unconditional `ask()` (i.e., detached HEAD is never treated as "protected" for allow-purposes, but is also never allowed to bypass the ask — the safe failure mode when identity is ambiguous is "ask", matching the file's global philosophy of failing toward asking/denying, never toward silent allow, per the `ERR` trap at line 49 and the SQL/cloud/rm guards' malformed-JSON fallback direction).
3. **New `force_scope_mode()`** cache function mirroring `rm_scope_repo_enabled()`, resolving `guards.forceScope` (`"protected"` / `"all"` / `"off"`) + `LOOM_FORCE_SCOPE` env override, called lazily only after an ASK_PATTERNS force-op match.
4. **Gate the three existing ASK_PATTERNS entries** (force-push, force-push short flag, hard-reset — lines 703–705) — pull them out of the plain substring loop (lines 742–746) into a dedicated block, parallel to how `CLOUD_ASK_PATTERNS` (lines 794–822) was split out from `ASK_PATTERNS` under `cloud_guard_enabled()`. Do **not** touch `git clean -fd`, `git checkout \.`, `git restore \.` (lines 706–708) — those are out of scope for this issue (not force ops, no branch semantics).
5. In `protected` mode: resolve the target branch; if it matches the protected set → `ask()` (keep existing behavior for the dangerous case — note the explicit main/master-plus-origin combos stay hard-denied upstream regardless of this toggle); otherwise → allow (fall through, no ask).
6. In `all` mode: unconditional `ask()` for all three patterns — i.e., **today's exact behavior**, byte-for-byte.
7. In `off` mode: never ask/deny on these three patterns (bypass entirely) — matches the `rmScope`/`sqlDdl`/`cloudCli` "off" semantics.

### Design point to settle before implementation: default value

The issue proposes `"protected"` as the default. **Recommendation: ship the default as `"all"` (preserve current behavior byte-for-byte), not `"protected"`.**

Rationale, weighed against the two closest Loom precedents:

- **`guards.sqlDdl` / `guards.cloudCli`** (both default `true`/on): the established pattern for *these* toggles is "guard stays exactly as it always was unless the repo explicitly opts out." A missing key or fresh install sees **zero behavior change**. This is the dominant convention across the file's three existing toggles (2 of 3 preserve current-behavior-as-default).
- **`guards.rmScope`** (default flipped from permissive to the stricter mode in #3628 / ADR Option B): this is the one precedent where the *default itself* changed, but it moved in the **safety-increasing** direction (permissive → stricter) as a deliberate, explicitly-called-out breaking change in a dedicated ADR, documented prominently in `defaults/CLAUDE.md` ("This is a **behaviour change**"). It is not evidence for defaulting a *new* toggle to the less-strict setting.
- Defaulting `forceScope` to `"protected"` would be the **first** toggle in this file to ship with a *less strict* out-of-the-box default than history — every existing zero-config repo would silently stop being asked on own-branch hard-resets and force-pushes the moment they pull the next Loom update, with no opt-in step. That is a silent behavior change for every existing consumer, which cuts against the file's own stated principle (see the `sqlDdl`/`cloudCli` doc comments: "Such repos opt out; everyone else keeps the guard on").
- The autonomous-agent stall this issue solves is real and worth fixing, but the fix should be **opt-in** (`guards.forceScope: "protected"` in `.loom/config.json`, exactly as the requesting consumer already plans to configure) rather than a new global default. This also sidesteps needing a coordinated migration note/ADR the way #3628 required for rmScope.

If the default is set to `"protected"` anyway (overriding this recommendation), it must be called out in `defaults/CLAUDE.md` as an explicit behavior change (new section, parallel to the "Repo-Scoped rm Guard" section) with the same visibility the rmScope default flip got.

### Acceptance criteria (concrete)

- [ ] `guards.forceScope` config key added, values `"all"` (default — see design point above), `"protected"`, `"off"`; `LOOM_FORCE_SCOPE` env var overrides config with the same three values.
- [ ] Malformed/missing `.loom/config.json`, or a `guards.forceScope` value outside the three, falls back to the default (`"all"`) — mirrors the `rmScope`/`sqlDdl`/`cloudCli` malformed-JSON-never-fails-hook contract.
- [ ] `protected` mode: force op targeting the resolved default branch (via `loom_default_branch()`) → `ask()` (unchanged from today). Force op targeting any other branch → allowed (no ask, no deny).
- [ ] `all` mode: force op → always `ask()`, regardless of target branch — must be behaviorally identical to pre-change `guard-destructive.sh` for every existing test case in `tests/hooks/test-guard-destructive.sh` lines 450–454.
- [ ] `off` mode: force op never asks or denies (passes straight through), for any branch.
- [ ] The unconditional hard-denies for explicit force-pushes to the default-branch names in `ALWAYS_BLOCK_PATTERNS` (lines 297–303) are untouched by `forceScope` in every mode, including `off`.
- [ ] Refspec parsing handles, at minimum:
  - a force-push naming `origin` and a bare branch ref (target = that ref)
  - the same, with a leading `+` on the ref (stripped before comparison)
  - a force-push using the `<src>:<dst>` refspec form (target = `<dst>`)
  - a force-push naming `HEAD` as the ref (resolve checked-out branch)
  - a force-push with no ref/refspec at all, relying on upstream tracking (resolve checked-out branch)
  - `git reset --hard` with or without a commit-ish argument (no refspec ever — always resolves via checked-out branch)
  - Detached HEAD at the resolved cwd → branch resolution fails → treat as unknown/ambiguous → falls through to `ask()` (never silently allowed), regardless of `protected` vs `all` (only `off` bypasses).
  - A command carrying a `git -C /some/other/repo` prefix — cwd resolution honors the `-C` flag argument, not just the hook's `$CWD` JSON field.
- [ ] New test cases added to `tests/hooks/test-guard-destructive.sh` (new section mirroring the existing "Repo-scoped rm guard" section at line 835) covering: protected-branch ask, own-branch allow, `off` bypass, `all` preserves-existing-tests, each refspec form above, and detached HEAD.
- [ ] Documented in `defaults/CLAUDE.md` alongside the existing "Repo-Scoped rm Guard (`guards.rmScope` / `LOOM_RM_SCOPE`)" section (same file, new "## Force-Op Branch Scope Guard" subsection under "## Custom Guard Hooks").

### Test plan

1. `bash tests/hooks/test-guard-destructive.sh` — full existing suite must stay green (default `"all"` behavior must not regress the existing force-push / hard-reset ask/deny assertions).
2. New assertions (mirroring `assert_deny_env` / `assert_allow_env` / `make_sql_repo` helpers already in the test file, lines 70–120):
   - `guards.forceScope: "protected"` + force-push to the repo's default branch → ask.
   - `guards.forceScope: "protected"` + force-push to a non-default branch → allow.
   - `guards.forceScope: "protected"` + `git reset --hard` on a checked-out non-default branch → allow.
   - `guards.forceScope: "protected"` + `git reset --hard` while on the default branch → ask.
   - `guards.forceScope: "off"` + force-push to default branch → allow (bypass).
   - `LOOM_FORCE_SCOPE=all` overrides a config `"protected"` → ask regardless of branch.
   - Each refspec edge case in Acceptance Criteria above, both allow and ask outcomes.
   - Regression: an explicit force-push (any flag form) naming `origin` and a literal default-branch ref still hard-denies in every `forceScope` mode (including `off`) — this path is in `ALWAYS_BLOCK_PATTERNS`, upstream of and unaffected by the new toggle.
3. Manual smoke test in a scratch repo with a non-`main` default branch (e.g. `master`) to confirm `loom_default_branch()` resolution is exercised, not just a hardcoded `"main"` string.

rjwalters:
Closes #3674

## Summary

Adds a `guards.forceScope` config toggle to `defaults/hooks/guard-destructive.sh`, symmetric to `guards.rmScope`, making the generic force-op ask (`git push --force` / `-f` / `--force-with-lease` and `git reset --hard`) **branch-aware** so autonomous/background agents no longer stall on routine force-pushes/hard-resets of their own working branch.

| `guards.forceScope` | Behavior |
|---------------------|----------|
| `"all"` (**default**) | Ask on every force op — current behaviour preserved byte-for-byte; zero-config installs see no change. |
| `"protected"` | Ask only when the resolved target is a protected branch (repo default + `main`/`master`) or the branch identity is ambiguous (detached HEAD); own working branches pass through. |
| `"off"` | Never ask/deny on force ops. |

Per the curator's design-point recommendation, the shipped default is `"all"` (opt-in to `"protected"`), so no existing consumer sees a silent behaviour change.

## Implementation

- **`force_scope_mode()`** mirrors `rm_scope_repo_enabled()`: lazy, cached (`_FORCE_SCOPE_CACHE`), jq `if/elif/else` (never `//`), `LOOM_FORCE_SCOPE` env override, malformed/missing config → `"all"`.
- **Branch resolution** (`parse_force_ops` awk + bash): parses the push refspec (`<src>:<dst>` → `<dst>`; bare ref → `+` stripped; `HEAD`/none → checked-out branch), honors a `git -C <path>` prefix for cwd (else the hook `cwd`); `reset --hard` always resolves to the checked-out branch. Detached HEAD → ambiguous → **ask** (never silently allowed).
- **Protected set** = inlined offline-first default-branch detection (mirrors `loom_default_branch()`, no network, `LOOM_DEFAULT_BRANCH` seam) plus `main`/`master` literals.
- The three force-op entries were pulled out of the ungated `ASK_PATTERNS` loop into a dedicated gated block. `git clean`/`checkout .`/`restore .` stay in `ASK_PATTERNS` (not force ops).
- The `ALWAYS_BLOCK` main/master force-push hard-denies are untouched and fire in **every** mode, including `"off"`.

## Testing

- `bash tests/hooks/test-guard-destructive.sh` — **292/292 pass** (263 pre-existing unchanged + 29 new `forceScope` cases: default-branch ask, working-branch allow, each refspec form, detached HEAD, `git -C` cwd, `off` bypass, env overrides, malformed/out-of-range config, and cross-mode main/master deny regressions).
- `shellcheck` clean — only the pre-existing SC2016 note on the `$HOME` pattern (present on `main`, unrelated).

## Docs

New "Force-Op Branch Scope Guard" section in `defaults/CLAUDE.md` alongside the rm-scope guard, plus the toggle enumeration in the Custom Guard Hooks intro.

### DIFF
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -904,7 +904,7 @@ If setup fails, it's usually due to:
 
 Loom ships with two built-in Bash `PreToolUse` guard hooks, both registered under the `Bash` matcher and firing independently:
 
-- **`guard-destructive.sh`** — the generic repository-hygiene guard: catastrophic denies (`rm -rf /`, force-push to `main`, `gh repo delete`, fork bombs, curl-pipe-to-shell, cloud/SQL destruction), the segment-parsed lifecycle/cloud-CLI checks, and the `guards.sqlDdl` / `guards.cloudCli` / `guards.rmScope` toggle machinery. Nothing about this guard is Loom-specific; it is slated to move to Repo Skills (companion issue [rjwalters/repo#13](https://github.com/rjwalters/repo/issues/13)), which will own the generic half once it ships. Until then it keeps shipping and working in Loom exactly as before.
+- **`guard-destructive.sh`** — the generic repository-hygiene guard: catastrophic denies (`rm -rf /`, force-push to `main`, `gh repo delete`, fork bombs, curl-pipe-to-shell, cloud/SQL destruction), the segment-parsed lifecycle/cloud-CLI checks, and the `guards.sqlDdl` / `guards.cloudCli` / `guards.rmScope` / `guards.forceScope` toggle machinery. Nothing about this guard is Loom-specific; it is slated to move to Repo Skills (companion issue [rjwalters/repo#13](https://github.com/rjwalters/repo/issues/13)), which will own the generic half once it ships. Until then it keeps shipping and working in Loom exactly as before.
 - **`guard-loom-workflow.sh`** — the thin, Loom-workflow-specific guard (issue #3604): the `gh pr merge` → `merge-pr.sh` redirect and the `pip install -e` worktree block (keyed on `LOOM_WORKTREE_PATH`, issue #2495). These two guards are specific to the Loom worktree/merge workflow and stay Loom-owned.
 
 You can also add project-specific guards to protect read-only directories from accidental edits (see below).
@@ -1032,6 +1032,82 @@ LOOM_RM_SCOPE=off rm -rf /Users/someone/scratch       # allowed (permissive)
 
 # Force repo mode for one command even when the repo opts out:
 LOOM_RM_SCOPE=repo rm -rf /Users/someone/important    # DENIED (outside repo)
+```
+
+### Force-Op Branch Scope Guard (`guards.forceScope` / `LOOM_FORCE_SCOPE`)
+
+By default `guard-destructive.sh` **asks** for confirmation on every `git push
+--force` / `-f` / `--force-with-lease` and `git reset --hard`, regardless of
+which branch is targeted. For an autonomous/background agent that cannot answer
+an interactive prompt, that stalls the agent on *routine* work — force-pushing or
+hard-resetting its own single-owner working branch is a normal part of the
+rebase/amend/reset workflow. The genuinely dangerous case is a force op against a
+**protected/shared branch** (`main`/`master` or the repo's default branch).
+
+`guards.forceScope` makes the ask branch-aware (symmetric to `guards.rmScope`):
+
+| `guards.forceScope` | Behavior |
+|---------------------|----------|
+| `"all"` (**default**) | Ask on every force op regardless of branch — current behaviour, preserved byte-for-byte. |
+| `"protected"` | Ask only when the resolved target is a **protected** branch (the repo default branch plus `main`/`master`), or the branch identity is ambiguous (detached HEAD). Force ops on the agent's own working branches pass through. Solves the autonomous-agent stall. |
+| `"off"` | Never ask/deny on force ops. |
+
+The shipped default is **`"all"`** — a zero-config install sees **no behaviour
+change**. Consumers who want the autonomous-friendly behaviour opt in explicitly
+(`guards.forceScope: "protected"` in `.loom/config.json`).
+
+**Protected set & branch resolution**:
+- Protected branches = the repo default branch (detected offline via
+  `refs/remotes/origin/HEAD`, mirroring `loom_default_branch()`, with a
+  `LOOM_DEFAULT_BRANCH` override) plus the literals `main` and `master`.
+- The target branch is resolved from the push refspec — `<src>:<dst>` → `<dst>`,
+  a bare ref → the ref with a leading `+` stripped, and `HEAD` / no refspec → the
+  **checked-out branch**. `git reset --hard` always resolves to the checked-out
+  branch. The checked-out branch is read at the command's effective cwd, honoring
+  a `git -C <path>` prefix, else the hook's `cwd`.
+- **Detached HEAD** (or any unresolved branch identity) is treated as ambiguous
+  and **asks** — it is never silently allowed.
+
+**Always-on hard denies are unaffected**. The unconditional force-push-to-main /
+force-push-to-master denies (the `ALWAYS_BLOCK` patterns) fire **in every mode,
+including `"off"`** — `forceScope` only ever downgrades the generic force-op
+*ask*, it never weakens a hard deny.
+
+The force-op guard is resolved in this order (highest precedence first):
+
+1. **`LOOM_FORCE_SCOPE` env var** (`all`/`protected`/`off`). Overrides config.
+2. **`.loom/config.json`** — `guards.forceScope`: `"protected"`/`"off"`; an
+   absent key, any other value, or malformed JSON resolves to `"all"`:
+   ```json
+   {
+     "guards": {
+       "forceScope": "protected"
+     }
+   }
+   ```
+3. **Default** — `"all"` (preserve current behaviour).
+
+The config read is best-effort: a missing, empty, or malformed `.loom/config.json`
+falls through to `"all"` and never causes the hook to exit non-zero.
+
+**Examples**:
+
+```bash
+# Default (all) — every force op asks, no config needed:
+git reset --hard HEAD~1                       # ASK
+git push --force origin feature/my-branch     # ASK
+
+# Opt in to branch-aware force ops for a whole repo:
+#   .loom/config.json  ->  { "guards": { "forceScope": "protected" } }
+git reset --hard HEAD~1                        # allowed (own working branch)
+git push --force origin feature/my-branch      # allowed (working branch)
+git push --force origin main                   # DENIED (ALWAYS_BLOCK, unaffected)
+
+# One-off env override — force branch-aware mode for a single command:
+LOOM_FORCE_SCOPE=protected git push --force origin feature/x   # allowed
+
+# Force the old always-ask behaviour even when the repo opts into protected:
+LOOM_FORCE_SCOPE=all git reset --hard HEAD~1   # ASK
 ```
 
 ### Protecting Read-Only Directories
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -343,7 +343,11 @@ resolve_default_branch() {
 #   - `git -C <path> ...` sets <cpath>; other pre-subcommand global options are
 #     skipped (`-c <k=v>` consumes its argument).
 #   - push: emitted only when a --force/-f/--force-with-lease flag is present.
-#     <target> is the destination branch parsed from the refspec —
+#     ONE line is emitted per positional refspec (pos[2], pos[3], …) after the
+#     remote — a multi-refspec push like `git push --force origin a b` emits a
+#     line for `a` AND `b`, so a protected branch in any refspec position (not
+#     just the first) reaches the caller's per-line check (#3674 follow-up).
+#     <target> is the destination branch parsed from each refspec —
 #       * `<src>:<dst>` form => <dst>
 #       * a bare ref        => the ref with a leading `+` stripped
 #       * `HEAD`, or no ref => the literal "@HEAD@" (resolve checked-out branch)
@@ -379,6 +383,11 @@ parse_force_ops() {
             if (subcmd == "push") {
                 force = 0
                 np = 0
+                # pos is a file-global awk array; clear it per segment
+                # (portable — split with an empty string empties the array) so
+                # refspecs from a prior segment cannot leak into this one now
+                # that we read every positional slot, not just pos[2].
+                split("", pos)
                 for (j = k+1; j <= m; j++) {
                     t = toks[j]
                     if (t == "--force" || t == "-f" || t == "--force-with-lease" || t ~ /^--force-with-lease=/) { force = 1; continue }
@@ -387,15 +396,23 @@ parse_force_ops() {
                     pos[np] = t
                 }
                 if (!force) continue
-                target = "@HEAD@"
-                if (np >= 2) {
-                    rs = pos[2]
-                    sub(/^\+/, "", rs)
-                    ci = index(rs, ":")
-                    if (ci > 0) rs = substr(rs, ci + 1)
-                    if (rs != "HEAD" && rs != "") target = rs
+                # pos[1] is the remote; pos[2..np] are refspecs. Emit ONE line per
+                # positional refspec so a protected branch in ANY refspec position
+                # (not just the first) reaches the per-line check in the caller. A
+                # bare push with no refspec (np < 2) resolves the checked-out branch.
+                if (np < 2) {
+                    print cpath SEP "@HEAD@"
+                } else {
+                    for (p = 2; p <= np; p++) {
+                        rs = pos[p]
+                        sub(/^\+/, "", rs)
+                        ci = index(rs, ":")
+                        if (ci > 0) rs = substr(rs, ci + 1)
+                        target = "@HEAD@"
+                        if (rs != "HEAD" && rs != "") target = rs
+                        print cpath SEP target
+                    }
                 }
-                print cpath SEP target
             } else if (subcmd == "reset") {
                 hard = 0
                 for (j = k+1; j <= m; j++) if (toks[j] == "--hard") hard = 1
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -242,6 +242,169 @@ resolve_worktree_root() {
     printf '%s/.loom/worktrees' "$repo_root"
 }
 
+# =============================================================================
+# force-op branch-scope toggle — default ALL (preserve current behaviour).
+#
+# The three generic force-op ASK patterns (git push --force / -f /
+# --force-with-lease and git reset --hard) prompt on EVERY match regardless of
+# which branch is targeted. For an autonomous/background agent that cannot answer
+# an interactive prompt, that stalls the agent on routine own-branch rebase /
+# amend / reset work. The genuinely dangerous case is a force op against a
+# PROTECTED branch (the repo default plus main/master), which stays a hard deny
+# via ALWAYS_BLOCK_PATTERNS for the explicit main/master forms.
+#
+# guards.forceScope selects the behaviour:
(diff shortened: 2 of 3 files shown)


rjwalters:
Thorough review — strong PR overall, but I found one **security gap in `protected` mode** that I'd like fixed before merge. Requesting changes.

## Blocking: multi-refspec force-push can silently target a protected branch

The awk in `parse_force_ops()` only inspects the **first** refspec (`pos[2]`) and emits a single target line per `push` segment. Additional refspecs are ignored. Combined with the `ALWAYS_BLOCK` patterns being ordering-sensitive substrings, a protected branch that appears in a **non-first** refspec position slips through in `protected` mode.

Reproduced against the PR head (repo in `forceScope:"protected"`, checked out on `feature/work`):

| Command | Result |
|---|---|
| `git push --force origin feature/x main` | **ALLOW** — protected branch silently force-pushed |
| `git push --force origin main feature/x` | deny (only because the `ALWAYS_BLOCK` substring happens to match) |
| `git push --force origin main` | deny (control) |
| `git push --force origin feature/x` | allow (control, correct) |

The first row is the problem: `main` is a real push target but is neither resolved by the parser (it only looks at `pos[2]` = `feature/x`) nor caught by `ALWAYS_BLOCK` (the substring `git push --force origin main` does not match `... origin feature/x main`). This defeats the exact invariant `protected` mode advertises: "ask when the resolved target is a protected branch."

The default `"all"` mode is **not** affected (it asks as soon as any force op parses), so the shipped-default behavior is safe — the gap is confined to the opt-in `protected` mode. But since the whole reason the mode exists is to keep protected-branch force ops guarded, a silent-allow here is a real hole.

### Suggested fix
In the `push` branch of `parse_force_ops()`, iterate **every** positional refspec after the remote (`pos[2]`, `pos[3]`, ...) and emit one `cpath SEP target` line per refspec (resolving each `<src>:<dst>` / bare-ref / `+` / `HEAD` the same way). The caller's per-line loop already asks if any line resolves to a protected/ambiguous target, so emitting all refspecs closes the gap with no caller change. Also reset/scope the `pos` array per segment if you start reading beyond `pos[2]`, since it is a global awk array. Please add a regression test for `git push --force origin feature/x main` (and the reverse) asserting an ask in `protected` mode.

## What I verified and liked
- **292/292 tests pass** locally; the 29 new cases are well-targeted.
- Default `"all"` preserves current behavior — every parsed force op asks (verified bare / `-f` / `--force-with-lease` / multi-refspec all still ask).
- `"off"` never disables the `ALWAYS_BLOCK` main/master hard-denies — verified `off` + `git push --force origin main` / `git push -f origin master` still deny.
- Detached HEAD -> ask (never silent-allow) — covered and passing.
- `git -C <path>` cwd resolution honored with no bypass; `git push --force-with-lease=main:sha origin feature/x` correctly resolves the **push target** (feature/x), not the lease value.
- The `\037` (US) separator fix is sound — non-whitespace so `read` preserves an empty leading `cpath` field; awk `sprintf("%c",31)` and bash `IFS` are consistent.
- `force_scope_mode()` mirrors `rm_scope_repo_enabled()` cleanly (lazy, cached, jq if/elif/else, env override, malformed -> all); config read is best-effort and never trips the ERR trap.
- Docs in `defaults/CLAUDE.md` are accurate and thorough.

Once the multi-refspec parsing is closed (all refspecs emitted + a regression test), this is an approve.

rjwalters:
Doctor pass complete (commit 140a99c8): `parse_force_ops()` now emits one parsed line per positional refspec (awk `pos` array reset per push segment), so the caller's per-line loop asks/denies when ANY target — first position or not — resolves protected or ambiguous. Added regression tests for the multi-refspec forms; full suite 299/299, shellcheck clean. Re-requesting review.

rjwalters:
LGTM — approved. Re-review after Doctor fix `140a99c8`.

The single blocking finding from my prior review (comment 5035462996) is resolved. `parse_force_ops()` now emits one parsed line per positional refspec (`pos[2..np]`) instead of only inspecting `pos[2]`, so a protected branch in ANY refspec position reaches the caller's per-line ask/deny loop.

Verified:
- **The gap is closed**: `git push --force origin feature/x main` (protected branch in 2nd position) now ASKS in protected mode instead of being silently allowed.
- **Control holds**: protected branch in 1st position is still hard-denied by ALWAYS_BLOCK before the force-scope block is reached.
- `HEAD:<protected>` dst in a non-first refspec resolves to `<dst>` and asks; a configured non-main default branch (develop) in a non-first refspec asks.
- Multiple non-protected refspecs (including `+feature/x` and `HEAD:feature/y` forms) all resolve to working branches → allow.
- **No new bypass**: the per-segment `split("", pos)` reset prevents stale refspec leakage across chained `&&` git commands. (The read loop is also bounded by the current segment's `np`, so the reset is correctly belt-and-suspenders.)
- Invariants intact: `\037` (US) separator, bare-push/reset→`@HEAD@`, detached-HEAD→ask, and off-mode ALWAYS_BLOCK are all byte-for-byte preserved.

Suite: 299/299 pass (ran locally in the worktree). CI green, mergeStateStatus CLEAN.

PR #3672, opened 2026-07-21
fix(merge): reset loom:building on Part-of partial-increment issues (#3667)

rjwalters:
## Summary

Closes #3667.

`Part of #N` / `Contributes to #N` partial-slice PRs orphan the `loom:building` label: GitHub never auto-closes the referenced family/epic issue, and `merge-pr.sh` intentionally skips issue-label cleanup (the #2838 decision only reasoned about the `Closes #N` auto-close case, where stale labels on a *closed* issue are harmless). Merging a non-closing partial PR therefore leaves the issue **open** with `loom:building` stuck on it — nothing reclaims it until a time-gated `/sweep all` stale pass (>=2h), and non-aggressive sweeps hard-skip it indefinitely. A human had to manually strip the label to unblock the next increment.

## Fix

At the deterministic merge choke point in `merge-pr.sh`, after a confirmed merge:

1. Parse the PR body for `Part of #N` / `Contributes to #N` (case-insensitive), deduped.
2. For each referenced issue, verify **fresh at merge time** (uncached `gh api`) that it is still **open** and still carries `loom:building`.
3. If so, swap `loom:building` -> `loom:issue` (back to the ready queue, mirroring `orphan_recovery.py`'s `recover_issue()` semantics) and post a short auditable comment.

No liveness check is needed — a merge just happened on the PR that came from whoever held the claim, so the increment's work is provably done (a deterministic, not heuristic, signal). The swap is **idempotent**: a no-op when the issue is already closed, already lacks `loom:building` (re-claimed by another builder), or is actually a PR (`.pull_request` guard). GitHub-only for v1 (guarded on `FORGE_TYPE`). The `Closes`/`Fixes`/`Resolves` auto-close path (#2838) is untouched.

## Acceptance criteria

- [x] Detects a merged PR whose body carries a non-closing partial-increment ref to a still-open issue.
- [x] Swaps `loom:building` -> `loom:issue` (not merely removes) so the issue re-enters the ready queue.
- [x] Idempotent skip when already closed / not `loom:building` / re-claimed / is a PR.
- [x] `Closes`/`Fixes`/`Resolves` behavior unchanged (additive; #2838 path untouched).
- [x] A subsequent non-aggressive `/loom:sweep <N>` sees `loom:issue`, not a hard-skipped `loom:building`.
- [x] `/sweep all` aggressive stale-claim reclaim unaffected (fallback for PRs that never merge).

## Tests

[...]

PR #3685, opened 2026-07-21
fix(merge-pr): route transient check-runs fetch failure into bounded pending-wait (#3678)

### LINKED ISSUE #3678: merge-pr.sh UNSTABLE poll: transient check-runs fetch failure after 'pending' reads as resolved-green
## Context
Non-blocking observation from the Judge review of PR #3669 (the UNSTABLE-state bounded poll, issue #3664), 2026-07-20 sweep session.

In the poll loop added to `defaults/scripts/merge-pr.sh`, once a PENDING set has been observed, a transient `forge_get_check_runs` failure (network blip, 5xx) yields empty JSON, which classifies as "no failing, no pending" → the resolved-green branch → immediate merge attempt. The failure mode is bounded because GitHub enforces required checks server-side (the merge API will refuse), but the script's local decision is wrong and the resulting error message misattributes the cause.

- [ ] A check-runs fetch failure mid-poll is distinguished from a genuinely empty check-run set (e.g. retry once, or treat fetch-error as "still pending" until timeout).
- [ ] The resolved-green branch is only taken on a successful fetch reporting no failing and no pending runs.
- [ ] Existing test suite (`test-merge-pr-unstable-fallback.sh`) gains a fetch-failure-mid-poll case.

---

### Root cause (grounded in current `main`)

The bug is a single-line classification collapse, not a logic error spread across the poll body.

**`defaults/scripts/merge-pr.sh:511`** (inside the `while true; do ... done` UNSTABLE poll loop that starts at line 510, added by #3669/#3664):

```bash
_UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null || echo '{"check_runs":[]}')"
```

This line does two things at once, and the combination is the bug:

1. It discards `forge_get_check_runs`'s own exit status. In bash, `VAR="$(cmd)" || fallback` evaluates `fallback` when `cmd` (the last command substituted) exits non-zero — so a transient fetch failure *is* detected at this instant — but the `||` branch throws that signal away and replaces `_UNSTABLE_FAILING_RAW` with a hardcoded literal, `{"check_runs":[]}`.
2. That literal is **byte-for-byte identical in shape** to what a genuinely successful fetch returns when a commit legitimately has zero check-runs. Nothing downstream can tell the two cases apart.

Both forge backends already surface the failure as a non-zero exit before this line collapses it:
- GitHub path, **`defaults/scripts/lib/forge-helpers.sh:442-452`**: `gh api "repos/$nwo/commits/$commit/check-runs" ... --jq '{...}' 2>/dev/null` — on a network error / 5xx, `gh api` itself exits non-zero with empty/partial stdout (stderr suppressed), so `forge_get_check_runs`'s return code is already non-zero here.
- Gitea path, **`defaults/scripts/lib/forge-helpers.sh:415-421`**: explicitly does
  ```bash
  statuses=$(gitea_api GET "repos/.../commits/$commit/statuses" 2>/dev/null) || {
    echo '{"total_count":0,"check_runs":[]}'
    return 1
  }
  ```
  — same problem one layer down: it already normalizes a failure to the same empty-check-runs JSON shape *and* returns 1, but the caller at merge-pr.sh:511 re-suppresses stderr and ignores the non-zero exit anyway, so the `return 1` signal never reaches the classification logic.

Downstream, **`defaults/scripts/merge-pr.sh:514-520`** derives `_UNSTABLE_FAILING` and `_UNSTABLE_PENDING` from `_UNSTABLE_FAILING_RAW` via jq filters; both are empty for `{"check_runs":[]}`. Then at **lines 588-599**:

```bash
# Nothing failing, nothing pending.
if [[ "$_UNSTABLE_OBSERVED_PENDING" == "true" ]]; then
  # The checks we waited on all resolved green — the PR is now
  # effectively CLEAN. Fall through to immediate merge.
  info "PR #$PR_NUMBER checks resolved green; falling back to immediate merge"
  _UNSTABLE_FALLBACK_TO_MERGE=true
  break
fi
# (d) Never observed a pending check and none failed — a transient API
# gap or commit-status (vs check-run) failure the check-runs API omits.
# Be safe and keep the existing #3486 defensive error path.
error "Failed to enable auto-merge for PR #$PR_NUMBER: $AUTO_MERGE_OUTPUT"
```

Once `_UNSTABLE_OBSERVED_PENDING` has flipped true on an earlier loop iteration (the normal #3664 case: we saw a check queued/in_progress and started waiting), a *subsequent* transient fetch failure lands in the exact same branch as "the checks we waited on all resolved green" and triggers an immediate merge attempt on a commit whose actual check state is unknown. If `_UNSTABLE_OBSERVED_PENDING` is still false (failure on the very first iteration), it instead hits the pre-existing "unknown gap" hard-error at line 599 — misattributing a transient fetch problem as "$AUTO_MERGE_OUTPUT"'s original unstable-status error, which is also wrong but less dangerous (refuses rather than merges prematurely).

### Recommended fix

Stop swallowing the exit status at the callsite, and route a failed fetch into the existing bounded "wait" path instead of the classification branches:

```bash
_UNSTABLE_FETCH_RC=0
_UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null)" || _UNSTABLE_FETCH_RC=$?

if [[ "$_UNSTABLE_FETCH_RC" -ne 0 ]]; then
  # Optional: one immediate retry before treating as still-pending, to absorb
  # a single transient blip without burning a full poll interval.
  _UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null)" || _UNSTABLE_FETCH_RC=$?
fi

if [[ "$_UNSTABLE_FETCH_RC" -ne 0 ]]; then
  warning "Failed to fetch check-runs for PR #$PR_NUMBER (rc=$_UNSTABLE_FETCH_RC); treating as still-pending and continuing to poll"
  # Re-check merged-concurrently + deadline using the SAME guards already used
  # in the (b) pending branch (lines 569-580) so this doesn't bypass the
  # existing bounded-wait/timeout semantics.
  _UNSTABLE_RECHECK_JSON="$(forge_get_pr_nocache "$REPO_NWO" "$PR_NUMBER" "$GH" 2>/dev/null || echo '{}')"
  if [[ "$(echo "$_UNSTABLE_RECHECK_JSON" | jq -r '.merged // false')" == "true" ]]; then
    warning "PR #$PR_NUMBER merged by another process while waiting for checks"
    AUTO_MERGE_OK=true
    break
  fi
  if [[ "$(date +%s)" -ge "$_UNSTABLE_DEADLINE" ]]; then
    error "Timed out after ${LOOM_AUTO_MERGE_TIMEOUT}s waiting for check-runs to become fetchable for PR #$PR_NUMBER (last error rc=$_UNSTABLE_FETCH_RC). Re-run the merge once the forge API is healthy, or raise LOOM_AUTO_MERGE_TIMEOUT."
  fi
  sleep "$LOOM_AUTO_MERGE_POLL_INTERVAL"
  continue
fi

# ... existing _UNSTABLE_FAILING / _UNSTABLE_PENDING derivation and
# classification unchanged below, now only reached on a successful fetch.
```

Key properties this preserves:
- The "resolved-green" branch (lines 588-595) and the "unknown gap" hard-error (line 596-599) are reached **only** after a successful fetch — a fetch failure never lands in either.
- The existing `LOOM_AUTO_MERGE_TIMEOUT` / `LOOM_AUTO_MERGE_DEADLINE` bound still applies, so a persistently-broken forge API still times out with a clear, honestly-attributed error rather than looping forever or merging blind.
- No change to the (a)/(b)/(c) classification branches (lines 522-563) — this only inserts a new decision point *before* the existing derivation, gated on fetch success.

An implementer may choose the single-retry-then-wait shape above, or skip the inline retry and just fall straight into the existing pending-wait/sleep/continue on the first failure (simpler, and still satisfies the "resolved-green only on success" requirement) — either satisfies the acceptance criteria below. The double-suppression at `forge-helpers.sh:415-421` (Gitea path already collapsing to the empty-JSON literal + `return 1`) does not need to change; the fix only needs to stop discarding the `return 1` at the merge-pr.sh callsite.

### Acceptance criteria

- [ ] `defaults/scripts/merge-pr.sh`'s UNSTABLE poll loop (~line 510 onward) captures `forge_get_check_runs`'s exit status separately from the JSON payload — a non-zero exit is detected and does **not** get silently replaced by the same `{"check_runs":[]}` shape used for "no failing, no pending."
- [ ] On a fetch failure mid-poll, the script does not take the resolved-green branch (lines 588-595) or the unknown-gap hard-error branch (lines 596-599). Instead it retries once and/or treats the failure as still-pending, applying the same "merged concurrently?" recheck and `LOOM_AUTO_MERGE_DEADLINE`/`LOOM_AUTO_MERGE_TIMEOUT` bound already used by the (b) pending-wait branch (lines 565-586), so the behavior stays bounded rather than looping forever.
- [ ] A persistent fetch failure that outlives the timeout produces a clear, honestly-attributed error message (distinct from "timed out waiting for pending checks" and from the pre-existing "unknown gap" message), so an operator isn't misled about which condition fired.
- [ ] The (a)/(b)/(c) classification logic (lines 522-563) and the informational-vs-required branch precedence are unchanged for the success path.
- [ ] No behavior change for the happy path: a rollup that never has a fetch error behaves exactly as it does today (regression-free against the existing #3664 poll tests).

### Test plan

Extend `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh`, which already mirrors the poll body's classification via the `_unstable_decision()` helper (lines 499-529) fed by `_failing_names()` / `_pending_names()` (lines 447-453). Add a fetch-failure axis:

1. Extend (or add a sibling of) `_unstable_decision()` to accept a `fetch_ok` parameter (true/false) mirroring the new decision point, e.g. `_unstable_decision_with_fetch <runs> <required> <observed_pending> <fetch_ok>`. When `fetch_ok=false`, expect `"wait"` regardless of the `runs`/`observed_pending` inputs (since a failed fetch must not reach the empty-runs classification at all).
2. New cases to add:
   - Fetch fails **after** having observed a pending check (`observed_pending=true`, `fetch_ok=false`) → expect `"wait"`, NOT `"merge"`. This is the exact bug scenario from PR #3669's Judge note.
   - Fetch fails on the **very first** poll iteration (`observed_pending=false`, `fetch_ok=false`) → expect `"wait"`, NOT `"unknown"` (a fetch error is not the same as a genuine unknown-gap; unknown-gap should be reserved for a successful fetch that legitimately reports nothing).
   - Fetch succeeds with a genuinely empty check-run set and `observed_pending=false` → still expect `"unknown"` (regression guard: don't over-widen the fix so real unknown-gaps get silently retried forever).
   - Fetch succeeds with `observed_pending=true` and empty runs → still expect `"merge"` (regression guard for the existing #3664 resolved-green path).
3. If the implementer keeps the harness style used for the rest of the file (mirrored jq/bash snippets rather than exercising `merge-pr.sh` end-to-end), the new `fetch_ok` branch should be inserted into the mirror function using the same code shape planned for the real merge-pr.sh callsite (see "Recommended fix" above), so the test and the implementation stay in lockstep the way the existing `_unstable_decision()` already does for the (a)-(e) branches.

## Affected Files
- `defaults/scripts/merge-pr.sh` (root cause: line 511 fetch-failure-to-empty-JSON collapse; fix site: UNSTABLE poll loop, lines ~510-600)
- `defaults/scripts/lib/forge-helpers.sh` (`forge_get_check_runs`, lines 411-454 — both forge branches already surface failure via non-zero exit / `return 1`; no change strictly required here but referenced for context)
- `defaults/scripts/tests/test-merge-pr-unstable-fallback.sh` (test plan target — extend `_unstable_decision()` / add new fetch-failure cases)

rjwalters:
## Summary

Fixes the #3678 classification collapse in the `merge-pr.sh` UNSTABLE poll loop. A transient `forge_get_check_runs` fetch failure mid-poll was silently replaced with a hardcoded `{"check_runs":[]}` literal — byte-identical to a genuinely empty rollup — so once a pending check had been observed, the next fetch blip took the resolved-green branch and triggered a **premature immediate merge** on a commit whose real check state was unknown.

## Root cause

`defaults/scripts/merge-pr.sh:511` used `VAR="$(cmd)" || echo '{...}'`, which detects the non-zero exit but throws the signal away and substitutes the empty-JSON literal. Both forge backends already surface the failure as a non-zero exit; the callsite re-suppressed it.

## Fix

- Capture the fetch exit status separately (`_UNSTABLE_FETCH_RC`) instead of collapsing it into empty JSON.
- Retry once to absorb a single blip; on persistent failure, route into the **same bounded pending-wait path** used by the (b) pending branch — reusing its merged-concurrently recheck and `LOOM_AUTO_MERGE_DEADLINE`/`LOOM_AUTO_MERGE_TIMEOUT` bound.
- The resolved-green branch and the unknown-gap hard-error are now reached **only after a successful fetch**.
- A persistent fetch failure that outlives the timeout produces a distinct, honestly-attributed error ("waiting for check-runs to become fetchable") instead of misattributing it as pending-timeout or unknown-gap.
- The (a)/(b)/(c) classification and happy-path behavior are unchanged.

## Tests

Extended `test-merge-pr-unstable-fallback.sh` with a `_unstable_decision_with_fetch` mirror gated on a `fetch_ok` axis:
- fetch fails **after** observing pending -> `wait` (the exact PR #3669 Judge-note bug), NOT `merge`
- fetch fails on the **first** iteration -> `wait`, NOT `unknown`
- successful empty fetch, never pending -> `unknown` (regression guard, no over-widening)
- successful empty fetch after pending -> `merge` (#3664 resolved-green preserved)
- plus source-grep assertions for the new `_UNSTABLE_FETCH_RC` capture and the removed empty-JSON collapse

All merge-pr suites pass (unstable-fallback 58/58, partial-increment 18/18, worktree-awk 11/11, worktree-path 21/21, help 15/15). Shellcheck clean (no new warnings; pre-existing SC2015 at line 812 untouched).

Closes #3678

### DIFF
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -508,7 +508,40 @@ if [[ "$AUTO_MERGE" == "true" ]]; then
       _UNSTABLE_DEADLINE=$(( $(date +%s) + LOOM_AUTO_MERGE_TIMEOUT ))
 
       while true; do
-        _UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null || echo '{"check_runs":[]}')"
+        # Fetch the check-runs rollup, capturing the helper's own exit status
+        # separately from the JSON payload. A transient fetch failure (network
+        # blip, 5xx, Gitea `return 1`) must NOT be collapsed into the same
+        # `{"check_runs":[]}` shape a legitimately empty rollup produces —
+        # doing so lets a fetch error masquerade as "no failing, no pending"
+        # and, once a pending check has been observed, take the resolved-green
+        # immediate-merge branch on a commit whose real check state is unknown
+        # (#3678). Retry once to absorb a single blip, then route a persistent
+        # failure into the SAME bounded pending-wait path used by branch (b)
+        # below so the LOOM_AUTO_MERGE_TIMEOUT bound still applies.
+        _UNSTABLE_FETCH_RC=0
+        _UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null)" || _UNSTABLE_FETCH_RC=$?
+        if [[ "$_UNSTABLE_FETCH_RC" -ne 0 ]]; then
+          _UNSTABLE_FETCH_RC=0
+          _UNSTABLE_FAILING_RAW="$(forge_get_check_runs "$REPO_NWO" "$_UNSTABLE_HEAD_SHA" 2>/dev/null)" || _UNSTABLE_FETCH_RC=$?
+        fi
+        if [[ "$_UNSTABLE_FETCH_RC" -ne 0 ]]; then
+          # Fetch is failing (twice). Treat as still-pending and keep polling,
+          # reusing the (b) branch's merged-concurrently recheck + deadline
+          # guard so this never bypasses the bounded-wait/timeout semantics.
+          warning "Failed to fetch check-runs for PR #$PR_NUMBER (rc=$_UNSTABLE_FETCH_RC); treating as still-pending and continuing to poll"
+          _UNSTABLE_RECHECK_JSON="$(forge_get_pr_nocache "$REPO_NWO" "$PR_NUMBER" "$GH" 2>/dev/null || echo '{}')"
+          if [[ "$(echo "$_UNSTABLE_RECHECK_JSON" | jq -r '.merged // false')" == "true" ]]; then
+            warning "PR #$PR_NUMBER merged by another process while waiting for checks"
+            AUTO_MERGE_OK=true
+            break
+          fi
+          if [[ "$(date +%s)" -ge "$_UNSTABLE_DEADLINE" ]]; then
+            error "Timed out after ${LOOM_AUTO_MERGE_TIMEOUT}s waiting for check-runs to become fetchable for PR #$PR_NUMBER (last fetch rc=$_UNSTABLE_FETCH_RC). Re-run the merge once the forge API is healthy, or raise LOOM_AUTO_MERGE_TIMEOUT."
+          fi
+          info "PR #$PR_NUMBER is UNSTABLE: check-runs fetch failing (rc=$_UNSTABLE_FETCH_RC); waiting ${LOOM_AUTO_MERGE_POLL_INTERVAL}s for the forge API (timeout ${LOOM_AUTO_MERGE_TIMEOUT}s)..."
+          sleep "$LOOM_AUTO_MERGE_POLL_INTERVAL"
+          continue
+        fi
         # Names of failing check runs (terminal non-success conclusions).
         # Sort + uniq to dedupe re-runs with the same context.
         _UNSTABLE_FAILING="$(echo "$_UNSTABLE_FAILING_RAW" | \
@@ -603,7 +636,7 @@ if [[ "$AUTO_MERGE" == "true" ]]; then
         _UNSTABLE_FAILING _UNSTABLE_PENDING _UNSTABLE_REQUIRED \
         _UNSTABLE_INFORMATIONAL _UNSTABLE_OVERLAP _UNSTABLE_COUNT \
         _UNSTABLE_PENDING_COUNT _UNSTABLE_DEADLINE _UNSTABLE_RECHECK_JSON \
-        _UNSTABLE_LOOKUP_RC _UNSTABLE_OBSERVED_PENDING 2>/dev/null || true
+        _UNSTABLE_LOOKUP_RC _UNSTABLE_FETCH_RC _UNSTABLE_OBSERVED_PENDING 2>/dev/null || true
 
       if [[ "$_UNSTABLE_FALLBACK_TO_MERGE" == "true" ]]; then
         unset _UNSTABLE_FALLBACK_TO_MERGE
--- a/defaults/scripts/tests/test-merge-pr-unstable-fallback.sh
+++ b/defaults/scripts/tests/test-merge-pr-unstable-fallback.sh
@@ -557,6 +557,56 @@ assert_eq "unknown" "$result" "#3664: unknown gap (no failing, no pending, never
 result=$(_unstable_decision "$runs_failed" "Required Build" "false")
 assert_eq "refuse" "$result" "#3486 preserved: failing required check, nothing pending -> refuse"
 
+# --- Test the fetch-failure decision point (#3678) ---
+# The #3678 bug: a transient check-runs fetch failure mid-poll yields empty
+# JSON, which classifies as "no failing, no pending" -> the resolved-green
+# branch -> a premature immediate merge on a commit whose real check state is
+# unknown. The fix captures the fetch exit status separately and, on failure,
+# routes into the bounded pending-wait path BEFORE any empty-runs
+# classification runs. This mirror gates _unstable_decision on fetch success:
+# a failed fetch must always resolve to "wait", never reaching the (a)-(e)
+# empty-runs classification at all.
+echo ""
+echo "Testing UNSTABLE fetch-failure decision point (#3678)..."
+
+_unstable_decision_with_fetch() {
+    local runs="$1" required="$2" observed_pending="$3" fetch_ok="$4"
+    if [[ "$fetch_ok" != "true" ]]; then
+        # A failed fetch never reaches the empty-runs classification; it is
+        # treated as still-pending and re-polled (bounded by the deadline).
+        echo "wait"; return
+    fi
+    _unstable_decision "$runs" "$required" "$observed_pending"
+}
+
+# The exact bug scenario from PR #3669's Judge note: fetch fails AFTER a pending
+# check was observed. Must wait, NOT merge.
+result=$(_unstable_decision_with_fetch "$runs_green" "Required Build" "true" "false")
+assert_eq "wait" "$result" "#3678: fetch fails after observing pending -> wait (NOT premature merge)"
+
+# Fetch fails on the very first poll iteration (never observed pending). Must
+# wait, NOT hit the unknown-gap hard error — a fetch error is not a genuine
+# unknown gap.
+result=$(_unstable_decision_with_fetch "$runs_green" "Required Build" "false" "false")
+assert_eq "wait" "$result" "#3678: fetch fails on first iteration -> wait (NOT unknown-gap hard error)"
+
+# Regression guard: a SUCCESSFUL fetch of a genuinely empty rollup with no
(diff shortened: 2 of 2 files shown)


rjwalters:
LGTM — approving.

## Review summary

Verified the #3678 classification-collapse fix in the `merge-pr.sh` UNSTABLE poll loop.

**Correctness of the exit-status capture** — the change from `VAR="$(cmd)" || echo '{...}'` to `_UNSTABLE_FETCH_RC=0; VAR="$(cmd)" || _UNSTABLE_FETCH_RC=$?` is the right idiom. Under `set -euo pipefail`, an assignment-with-command-substitution takes the command substitution's exit status, the trailing `||` both captures it into `_UNSTABLE_FETCH_RC` and suppresses `set -e` abort. Confirmed empirically (fail→retry-fail yields rc=1→fetch-fail branch; fail→retry-ok yields rc=0→classify). It mirrors the existing `_UNSTABLE_LOOKUP_RC` pattern already in the same loop.

**Control-flow safety**:
- Fetch failure after a pending check was observed now routes into the bounded pending-wait path *before* any empty-runs classification runs — it can no longer take the resolved-green immediate-merge branch. This is the exact PR #3669 Judge-note bug, now fixed.
- First-iteration fetch failure waits rather than hitting the (d) unknown-gap hard error.
- The resolved-green and (d) unknown-gap branches are reachable only after a successful fetch.
- The one-retry absorbs a single blip; persistent failure reuses the (b) branch's merged-concurrently recheck + `LOOM_AUTO_MERGE_DEADLINE`/`TIMEOUT` bound, and emits a distinct honest "waiting for check-runs to become fetchable" timeout instead of misattributing pending-timeout/unknown-gap.
- Not setting `_UNSTABLE_OBSERVED_PENDING=true` in the fetch-fail branch is the correct conservative choice — a fetch error is not evidence of a pending check, so a later green-with-no-pending fetch safely refuses via (d) rather than merging.

**No regressions**: (a) required-failure refuse, (c) informational-merge, and the #3486 defensive unknown-gap path are unchanged. No conflict with the #3672 partial-increment or #3675 awk changes.

**`_UNSTABLE_FETCH_RC` added to the `unset` cleanup.**

**Verification** (from the worktree): unstable-fallback 58/58, partial-increment 18/18, worktree-awk 11/11, worktree-path 21/21, help 15/15. Shellcheck clean (only the pre-existing SC2015 at line 812, untouched by this PR). CI green (Installer Integration, Shellcheck, Detect Changes all pass); mergeStateStatus CLEAN.

The test strategy (a `fetch_ok`-gated mirror harness plus source-grep assertions for the `_UNSTABLE_FETCH_RC` capture and the removed empty-JSON collapse) is a reasonable approach for an infinite poll loop with sleeps and forge calls that can't be unit-tested directly.

PR #3686, opened 2026-07-21
fix(guards): stop false-positiving on force-push strings quoted in flag values (#3679)

### LINKED ISSUE #3679: guard-destructive.sh false-positives on force-push strings quoted inside gh pr comment bodies
## Context
Hit twice during the 2026-07-20/21 sweep session while reviewing PR #3676: both the Judge and the Doctor were blocked by this session's own `guard-destructive.sh` hook when posting `gh pr comment --body "..."` text that merely QUOTED example commands like a force-push of main (the review was *about* force-push guarding, so the examples were unavoidable). Both agents had to route the comment body through a temp file with placeholder substitution to get it posted.

The guard's ALWAYS_BLOCK / ASK patterns are substring matches over the whole command text, so a dangerous-looking string inside a quoted argument to a harmless command (`gh pr comment`, `gh issue create`, `echo`, `git commit -m`) trips the guard even though nothing destructive can execute.

- [ ] The guard distinguishes the command being executed from string-literal payload: a force-push pattern appearing only inside a quoted argument to a non-git command (or to a git subcommand that takes message text, like `commit -m`) does not deny/ask.
- [ ] Genuine force ops (including via `bash -c '...'`, chained `&&` segments) are still caught — the fix must not open an eval-style bypass; when in doubt, keep asking.
- [ ] Regression tests cover the `gh pr comment --body`-with-quoted-force-push case and at least one deliberate-bypass attempt that must still be blocked.

---

### Root cause (grounded in `defaults/hooks/guard-destructive.sh` on current `main`, post-#3676/cbe55ecd)

- `ALWAYS_BLOCK_PATTERNS` (lines 467-533) includes the explicit force-push-to-main/master literals at **lines 477-483**:
  `'git push --force origin main'`, `'git push --force origin master'`, `'git push -f origin main'`, `'git push -f origin master'`, `'git push --force-with-lease origin main'`, `'git push --force-with-lease origin master'`.
- These six patterns are matched at **lines 535-539** as raw, *unanchored* substrings against the **entire raw `$COMMAND` string** (`echo "$COMMAND" | grep -qiE "$pattern"`) — no command-position anchor, no argv0/segment awareness. Contrast with the gh repo delete/gh repo archive entries two lines above (474-475), which *are* anchored (`(^|[;&|[:space:]])gh repo delete`) per the header comment at 468-473 explaining the catastrophic scan intentionally still reads quoted/heredoc text (so a `bash -c` wrapped root-deletion payload stays caught, see 485-491 and the existing regression test at `tests/hooks/test-guard-destructive.sh:1267-1270`).
- Net effect: `gh pr comment 3676 --body "example: git push --force origin main"` contains the literal substring `git push --force origin main` inside the quoted `--body` value → matches line 478 → hard **deny**, even though nothing executes. Same failure mode applies to `-m`/`--title`/`--notes` text carried by `git commit -m "..."`, `gh pr create --title`, `gh release create --notes`, etc. This is the exact repro from the report (Judge/Doctor blocked while reviewing PR #3676, which is *about* force-push guarding, so the examples in the review comment were unavoidable — and, fittingly, this curation pass hit the identical trap composing this very enhancement text, confirming the report's severity firsthand).
- This is a **narrower gap** than the general command-vs-payload problem the repo already solved once: `parse_force_ops()` (lines 356-423) is a segment + argv0-aware awk parser — it splits on `; | & && ||`, requires `toks[1] == "git"` for the *segment itself* (line 370, comment at 342), and is what backs the `guards.forceScope` "ask" tier (the `FORCE-OP BRANCH SCOPE` block, lines 877-934). Since `gh pr comment ... --body "git push --force origin main"` has no top-level separator, the whole line is one segment whose command word is `gh`, not `git` — `parse_force_ops()` **already correctly ignores it**. The false positive is coming *only* from the raw `ALWAYS_BLOCK_PATTERNS` substring scan, not from the segment-aware ask-tier logic.
- Important asymmetry to preserve: `parse_force_ops()` does **not** recurse into `bash -c '...'` — a `bash -c 'git push --force origin main'` is one segment with `toks[1]=="bash"`, so `parse_force_ops` skips it entirely. Today, that genuine bypass attempt is caught **only** by the raw, unanchored `ALWAYS_BLOCK_PATTERNS` scan (same mechanism that produces the false positive). So a fix that simply removes/anchors the six main/master literals, or that reroutes them through `parse_force_ops()` alone, would silently stop catching `bash -c 'git push --force origin main'` — that is exactly the "no eval-style bypass" regression the acceptance criteria warns against. Any fix must keep that path denied.
- A second, subtler bypass risk: naive stripping of quoted `--body`/`-m` values (to fix the false positive) must **not** strip a value that itself contains shell-execution syntax — `git commit -m "$(git push --force origin main)"` or a backtick form actually *executes* the inner command via command substitution before `git commit` ever runs. If the fix blindly discards the whole quoted span, this becomes an unintended real bypass. The stripping must only elide a quoted span when it contains **no** `$(`, backtick, or `$((` — i.e. only spans that are provably inert text.

### Recommended approach

Add a narrow, portable (bash/awk, no python — consistent with the rest of the file) preprocessing step, mirrored on the existing `COMMAND_NO_COMMENT` pattern (lines 541-559):

1. Build `COMMAND_NO_LITERAL_TEXT` once, before the `ALWAYS_BLOCK_PATTERNS` loop (~line 535): scan `$COMMAND` for occurrences of a small, fixed set of known **text-carrying** flags — `--body`, `-m`/`--message`, `--title`, `--notes` (the `gh pr/issue comment|create`, `gh release create`, and `git commit -m` vocabulary implicated in the report) — immediately followed by a single- or double-quoted argument (`FLAG "..."` / `FLAG '...'` / `FLAG=...`).
2. For each such match, replace the quoted span with a same-length placeholder **only if the span contains none of `$(`, a backtick, `$((`** (no embedded command substitution). Spans containing those sequences are left untouched, so a `-m "$(git push --force origin main)"` bypass attempt is NOT stripped and still hits the raw scan.
3. Use `COMMAND_NO_LITERAL_TEXT` (instead of raw `$COMMAND`) **only** for the `ALWAYS_BLOCK_PATTERNS` loop at lines 535-539. Leave every other scan untouched: the rm-rf catastrophic patterns, fork bomb, curl-pipe-to-shell, and — critically — do NOT touch `bash -c '...'` payloads (`-c` is not in the text-carrying-flag list), so `tests/hooks/test-guard-destructive.sh:1267-1270` keeps passing unmodified.
4. Do not touch `parse_force_ops()` / the `guards.forceScope` ask-tier machinery at all — it is already correct for this class of false positive (see root-cause analysis above) and is out of scope for this fix.

This keeps the blast radius to exactly the reported failure mode, preserves the documented invariant that explicit main/master force-push forms are a hard, mode-independent deny (comment at lines 889-891), and does not weaken the `bash -c` / command-substitution bypass floor.

*(Out of scope for this issue, flag as a possible follow-up if desired: the same unanchored-literal-in-quoted-text problem likely also affects the gh repo delete/gh repo archive patterns at lines 474-475 whenever prose precedes them with plain whitespace, e.g. `gh pr comment --body "we should never run gh repo delete myrepo"` — the existing `(^|[;&|[:space:]])` anchor does not distinguish "start of an executed command" from "whitespace inside a quoted sentence". Not fixing that here keeps this PR small; a follow-up issue can extend `COMMAND_NO_LITERAL_TEXT` to the whole `ALWAYS_BLOCK_PATTERNS` loop rather than just the force-push subset.)*

### Acceptance criteria (refined)

- [ ] `gh pr comment <n> --body "..."` (and `gh issue comment`, `gh pr create --title`, `gh release create --notes`) containing one of the main/master force-push literal phrases as inert quoted text is **allowed** (no ask/deny).
- [ ] `git commit -m "..."` containing the same literal phrases as inert quoted text is **allowed**.
- [ ] `git push --force origin main` (and the `-f`/`--force-with-lease`, `origin master` variants) executed directly still **hard-denies**, unchanged.
- [ ] `bash -c 'git push --force origin main'` (single- and double-quoted forms) still **hard-denies** — no eval-style bypass introduced.
- [ ] `git commit -m "$(git push --force origin main)"` (command-substitution smuggling attempt) still **denies/asks** — the literal-text stripper must not elide a quoted span containing a subshell open, backticks, or an arithmetic-expansion open.
- [ ] Chained forms (`... && git push --force origin main`, `git push --force origin main && echo done`) still deny, unchanged.
- [ ] No new GitHub labels introduced; no changes to `parse_force_ops()` / `guards.forceScope` behavior.

### Test plan

Add to `tests/hooks/test-guard-destructive.sh` using the existing `assert_allow` / `assert_deny` / `assert_ask` helpers (see conventions at lines 201-293 and the `#3553 matching-precision` / `#3553 regression guard` sections at lines 1141 and 1258). Note the existing convention at line 1107 (`_PROT=main`) of assembling a protected-branch literal from a shell variable specifically so the *test file's own source* never carries a raw force-push-to-protected-branch substring that this session's own guard hook would trip on while editing the test — follow the same pattern when adding these cases:

- New "false positive now allowed" cases (mirror the style at lines 1220-1221 / 1096-1097):
  - `gh pr comment 3676 --body "example: git push --force origin main"` → `assert_allow`
  - `gh pr comment 3676 --body 'do not run git push --force origin master'` → `assert_allow`
  - `git commit -m "revert git push --force origin main mistake"` → `assert_allow`
  - `gh pr create --title "fix: prevent git push --force origin main" --body "n/a"` → `assert_allow`
- New "regression guard: genuine ops still caught" cases (mirror the style at lines 1263-1270):
  - `git push --force origin main` → `assert_deny` (unchanged, already covered at line 310 — re-verify it still passes)
  - `bash -c 'git push --force origin main'` → `assert_deny` (new — the critical no-bypass case; use single AND double quote forms like 1267/1270)
  - `git commit -m "$(git push --force origin main)"` → `assert_deny` (new — deliberate command-substitution bypass attempt named explicitly in the AC)
  - `foo && git push --force origin main` → `assert_deny` (chained form, new)

Run: `./tests/hooks/test-guard-destructive.sh` (must be all-green before merge; the suite is self-contained bash, no build step).

### Files likely touched

- `defaults/hooks/guard-destructive.sh` — add the literal-text stripping preprocessing step ahead of the `ALWAYS_BLOCK_PATTERNS` loop.
- `tests/hooks/test-guard-destructive.sh` — add the regression tests above.

## Affected Files

- `defaults/hooks/guard-destructive.sh`
- `tests/hooks/test-guard-destructive.sh`

rjwalters:
## Summary

`guard-destructive.sh` hard-denied any command whose text merely **quoted** a protected-branch force-push phrase inside a text-carrying flag value — e.g. `gh pr comment --body "example: <force-push to main>"` or `git commit -m "..."`. The six main/master force-push literals in `ALWAYS_BLOCK_PATTERNS` are raw, unanchored substring matches over the whole command, so inert quoted prose tripped them even though nothing destructive executes. This repeatedly blocked Judge/Doctor agents reviewing PRs *about* force-push guarding (the examples in the review comment were unavoidable).

## Fix

- New `strip_literal_text()` awk helper (portable bash/awk, mirrors the existing `parse_force_ops()` style) that redacts the quoted **values** of known text-carrying flags (`--body`, `-m`/`--message`, `--title`, `--notes`) to a same-length placeholder.
- Build `COMMAND_NO_LITERAL_TEXT` from it and use it **only** for the `ALWAYS_BLOCK_PATTERNS` loop (mirrors the `COMMAND_NO_COMMENT` precedent). Every other scan — rm-rf catastrophic, fork bomb, curl-pipe-to-shell, the segment-parsed lifecycle/cloud checks, and `parse_force_ops()` / `guards.forceScope` — keeps reading the raw command, untouched.

### Safety floor preserved
- `-c` is deliberately **not** a text-carrying flag, so `bash -c '<payload>'` is never redacted and stays caught by the raw scan (no eval-style bypass).
- A quoted span is redacted **only** when it carries no command-substitution / backtick opener (a `$(` — which subsumes the arithmetic `$((` — or a backtick), so a command-substitution smuggling attempt inside `-m "..."` is left intact and still hard-denies.
- The awk only runs when a text-carrying flag is present (cheap `[[ ==* ]]` guard), keeping it off the hot path.

`parse_force_ops()` and the `guards.forceScope` ask-tier are not modified.

## Tests

Added 11 regression tests to `tests/hooks/test-guard-destructive.sh` (**299 → 310, all green**):
- 5 "false positive now allowed" cases (`--body` double/single-quoted, `-m`, `--title`, `--notes`).
- 6 "genuine op still denied" cases: direct force-push, `bash -c '...'` (single- and double-quote forms), command-substitution smuggling in `-m`, and both chained (`&&`) forms.

Protected-branch phrases in the test source are assembled from shell fragments (mirroring the existing line-1107 convention) so the file carries no raw literal.

`shellcheck` on the hook: clean, no new findings (the one pre-existing SC2016 info on the `\$HOME` regex literal is unchanged).

Closes #3679

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -422,6 +422,62 @@ parse_force_ops() {
     }'
 }
 
+# Redact the quoted VALUES of known text-carrying flags (--body, -m/--message,
+# --title, --notes) so a dangerous-looking phrase quoted INSIDE such a value no
+# longer trips the raw ALWAYS_BLOCK_PATTERNS substring scan. Used ONLY to build
+# COMMAND_NO_LITERAL_TEXT for that one loop (mirrors the COMMAND_NO_COMMENT
+# precedent); every other scan keeps reading the raw command. This kills the
+# #3679 false positive where `gh pr comment --body "…git push --force origin
+# main…"` / `git commit -m "…"` hard-denied even though nothing executes.
+#
+# Safety floor preserved two ways:
+#   - `-c` is deliberately NOT a text-carrying flag, so `bash -c '<payload>'`
+#     is never redacted and its payload stays caught by the raw scan.
+#   - a quoted span is redacted ONLY when it carries no command-substitution or
+#     backtick opener (`$(` — which also subsumes the arithmetic `$((` — or a
+#     backtick). So a smuggling attempt like `git commit -m "$(git push --force
+#     origin main)"` is left intact and still hard-denies.
+# Each redacted span is replaced by a SAME-LENGTH placeholder so byte offsets of
+# the surrounding command are unchanged. Best-effort like COMMAND_NO_COMMENT:
+# it does not model backslash-escaped quotes, but since the result feeds only
+# the narrowing (never widening) catastrophic scan, the worst case is a raw
+# substring surviving — never a catastrophic block being skipped incorrectly.
+strip_literal_text() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)   # single quote
+        DQ = sprintf("%c", 34)   # double quote
+        # boundary + text-carrying flag + optional (ws / = / ws) + quoted span
+        re = "(^|[ \t])(--message|--body|--notes|--title|-m)[ \t]*=?[ \t]*(" \
+             DQ "[^" DQ "]*" DQ "|" SQ "[^" SQ "]*" SQ ")"
+    }
+    {
+        s = $0
+        out = ""
+        while (match(s, re)) {
+            pre     = substr(s, 1, RSTART - 1)
+            matched = substr(s, RSTART, RLENGTH)
+            s       = substr(s, RSTART + RLENGTH)
+            # Locate the opening quote inside the matched span.
+            qpos = 0
+            for (i = 1; i <= length(matched); i++) {
+                c = substr(matched, i, 1)
+                if (c == DQ || c == SQ) { qpos = i; break }
+            }
+            head  = substr(matched, 1, qpos)                              # up to & incl. opening quote
+            qchar = substr(matched, qpos, 1)
+            inner = substr(matched, qpos + 1, length(matched) - qpos - 1) # between the quotes
+            # Redact ONLY provably inert text (no command substitution / backtick).
+            if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                gsub(/./, "X", inner)
+            }
+            out = out pre head inner qchar
+        }
+        out = out s
+        printf "%s", out
+    }'
+}
+
 # Helper: output a deny decision and exit
 deny() {
     local reason="$1"
@@ -532,8 +588,22 @@ ALWAYS_BLOCK_PATTERNS=(
     # (#3584).
 )
 
+# Build a literal-text-redacted working copy ONLY for the catastrophic scan
+# below, so a force-push-to-main phrase quoted inside a --body/-m/--title/--notes
+# value no longer false-positives (#3679). The awk only runs when one of those
+# flags is actually present, keeping it off the hot path (mirrors the
+# COMMAND_NO_COMMENT `#`-present guard). `-c` is intentionally excluded so
+# `bash -c '<payload>'` payloads still reach the raw scan; spans carrying `$(` /
+# backtick are left intact so command-substitution smuggling still hard-denies.
+COMMAND_NO_LITERAL_TEXT="$COMMAND"
+if [[ "$COMMAND" == *"--body"* || "$COMMAND" == *"--message"* || \
+      "$COMMAND" == *"--title"* || "$COMMAND" == *"--notes"* || \
+      "$COMMAND" == *"-m"* ]]; then
+    COMMAND_NO_LITERAL_TEXT=$(strip_literal_text "$COMMAND")
+fi
+
 for pattern in "${ALWAYS_BLOCK_PATTERNS[@]}"; do
-    if echo "$COMMAND" | grep -qiE "$pattern"; then
+    if echo "$COMMAND_NO_LITERAL_TEXT" | grep -qiE "$pattern"; then
         deny "BLOCKED: Command matches dangerous pattern: $pattern"
     fi
 done
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -1356,6 +1356,63 @@ assert_deny "Regression: DELETE FROM without WHERE (guard on) still denied" \
 echo ""
 
 # =========================================================================
+echo -e "${YELLOW}--- #3679: force-push literals quoted in flag values no longer DENY ---${NC}"
+# =========================================================================
+#
+# ALWAYS_BLOCK force-push-to-main/master literals are raw, unanchored substring
+# matches over the whole command, so a force-push phrase merely QUOTED inside a
+# text-carrying flag value (`gh pr comment --body "…"`, `git commit -m "…"`,
+# `--title`, `--notes`) false-positived — even though nothing destructive can
+# execute. COMMAND_NO_LITERAL_TEXT redacts those quoted values ONLY for the
+# catastrophic loop, killing the false positive while keeping every genuine
+# force op (direct, `bash -c '…'`, command-substitution smuggling, chained)
+# denied.
+#
+# The protected-branch phrases are assembled from shell fragments so this test
+# file's own source never carries a raw "push --force origin <protected>"
+# literal that this session's guard hook would trip on (mirrors line 1107).
+_PB=main
+_MB=master
+_FP_MAIN="git push --force origin $_PB"       # direct force-push to protected main
+_FP_MASTER="git push --force origin $_MB"     # …to protected master
+_FP_MAIN_F="git push -f origin $_PB"           # short -f form
+
+# ---- false positives now ALLOWED (inert quoted text) ----
+assert_allow "#3679: force-push phrase in a gh pr comment --body (double-quoted) allowed" \
+    "gh pr comment 3676 --body \"example: $_FP_MAIN\""
+assert_allow "#3679: force-push phrase in a gh pr comment --body (single-quoted, master) allowed" \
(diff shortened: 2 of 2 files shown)


rjwalters:
LGTM — APPROVE. Reviewed adversarially as a security-critical guard relaxation.

**Verified the redaction only ever narrows, never widens:**
- Scope is correct: `COMMAND_NO_LITERAL_TEXT` is consumed *only* by the ALWAYS_BLOCK loop (guard-destructive.sh:606). The force-op ASK tier / `parse_force_ops` / `guards.forceScope` (line 970), rm-rf catastrophic (816), fork-bomb, curl-pipe, and the segment-parsed lifecycle/cloud checks all keep reading raw `$COMMAND` / `COMMAND_NO_COMMENT`. No behavioral change outside the one loop.
- **No over-consumption**: awk matches flag + its *immediately-following* quote pair; `pre` (everything before the flag) is preserved verbatim and only the `inner` quoted span is blanked. Confirmed with direct probes that a real force op *after* a redacted `-m` span, and one *between* two `-m` spans, both survive raw → still deny.
- **`$((` is covered** by the `$(` `index()` check (arithmetic left intact); backtick check present; `$(...)` smuggling inside `-m` stays raw → denies.
- **`bash -c` never redacted** (`-c` not text-carrying) — both quote-wrapper forms deny.
- **Unterminated quote** → regex requires a closing quote → no redaction → fail-safe deny.
- **Equals-form** (`--body="..."`) handled; same-length placeholder is harmless.
- Multi-line input: awk's per-record join only concatenates (increasing match surface), so a real op on a later line survives; cannot hide a pattern that raw line-based grep would have caught.

The failure direction is always toward *over*-blocking (a raw substring surviving), never a skipped catastrophic block — because a genuinely executable command lives outside quotes and is thus never inside a redacted span.

**Validation**: full suite 310/310 green (11 new cases with solid adversarial coverage: bash -c both quote forms, cmd-subst smuggling, chained `&&` forms). shellcheck clean (only the pre-existing SC2016 info on the `\$HOME` regex literal, unchanged). CI CLEAN, mergeable.

Non-blocking nits (no action needed): the `*"-m"*` presence guard is over-inclusive (matches inside `--message`/`foo-master`) — perf-only; escaped-quote-in-message can cause an over-deny — documented and safe-direction.

PR #3689, opened 2026-07-21
fix(scripts): make disk-headroom.sh sourceable under zsh (#3680)

### LINKED ISSUE #3680: disk-headroom.sh cannot be sourced under zsh — BASH_SOURCE empty, sibling worktree-root.sh not found
## Context
Hit during Stage -1 auto wave-size resolution in the 2026-07-20 /loom:sweep all run, following sweep.md's own instructions (`source ./.loom/scripts/lib/disk-headroom.sh`) in a zsh shell (macOS default; Claude Code's Bash tool runs zsh on this host):

```
./.loom/scripts/lib/disk-headroom.sh:source:46: no such file or directory: /Users/rwalters/GitHub/loom/worktree-root.sh
```

Root cause: `disk-headroom.sh` resolves its sibling via `_LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"`. Under zsh `BASH_SOURCE` is unset, so `dirname ""` → `.` and the source of `worktree-root.sh` resolves against the CWD (repo root) instead of the lib dir. Workaround used: source `worktree-root.sh` explicitly first.

- [ ] The lib resolves its own directory portably across bash and zsh (e.g. `${BASH_SOURCE[0]:-${(%):-%x}}` guarded by shell detection, or an explicit `$0`-based fallback), or documents+enforces a bash-only contract by having sweep.md invoke it via `bash -c` instead of `source`.
- [ ] `source ./.loom/scripts/lib/disk-headroom.sh` from the repo root works in both shells (covered by a test alongside `test-disk-headroom.sh`).
- [ ] Audit other `lib/*.sh` files for the same `BASH_SOURCE` sibling-sourcing pattern.

---

### Verified root cause

Confirmed by reading `defaults/scripts/lib/disk-headroom.sh` (line 44):

```bash
_LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"
```

`sweep.md`'s Stage -1 "Resolve auto wave size" step literally instructs sourcing this file directly into the invoking shell (`defaults/.claude/commands/loom/sweep.md:558`, mirrored at `.claude/commands/loom/sweep.md`):

```bash
source ./.loom/scripts/lib/disk-headroom.sh
```

`BASH_SOURCE` is a bashism — it is unset when a file is `source`d directly into a zsh shell (Claude Code's `Bash` tool runs zsh on macOS, and it's also the interactive-shell default on macOS). `dirname ""` → `.`, so `worktree-root.sh` is looked up relative to the shell's CWD (the repo root) instead of `defaults/scripts/lib/`, and the `source` fails exactly as reported.

**Why this is the *only* affected lib file** (verified by grepping every `source` call site for `defaults/scripts/lib/*.sh` across the repo): all other lib files (`forge-helpers.sh`, `loom-tools.sh`, `pipe-pane-cmd.sh`, `worktree-root.sh`) are sourced only from *within other bash-shebang'd scripts* that are themselves executed (`./script.sh` or `bash script.sh`) — in that case `bash` is the interpreter regardless of the caller's login shell, so `BASH_SOURCE` is always populated correctly. `disk-headroom.sh` is unique in being sourced directly into the top-level (possibly zsh) shell per `sweep.md`'s own documented usage. `loom-tools.sh` does use `${BASH_SOURCE[1]}` internally (`run_loom_tool`, `_loom_tool_not_found_error`, lines 96 and 132) to resolve the *caller's* directory — same bashism, but not exercised by this bug today since every current call site sources it from inside another bash-shebang script. Worth hardening at the same time for defense-in-depth (see AC below), but it is not the reported failure.

### Empirically verified portable idiom

Tested directly (bash 3.2.57 and zsh 5.9 on macOS), sourcing from the repo root, from a different CWD via absolute path, with/without a leading `./`, and under `set -u`/nounset — all pass:

```bash
_LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"
```

Rationale: zsh sets `$0` to the sourced file's path by default when a script is `source`d (confirmed empirically — no `emulate`/option changes needed), so `${BASH_SOURCE[0]:-$0}` degrades cleanly: bash uses `BASH_SOURCE[0]` (unchanged behavior), zsh falls through to `$0` (correct path). This is a strict superset fix — no behavior change for any existing bash call site. The "enforce bash-only via `bash -c`" alternative was considered and rejected: it would require rewriting `sweep.md`'s Stage -1 to invoke the helper as a subshell instead of sourcing it directly, which is more invasive than a one-line self-fix in the library, and forfeits the ability for interactive zsh operators to source the same helper themselves for manual disk-headroom checks (mentioned as a legitimate direct-invocation path in the file's own header comment).

### Audit of `defaults/scripts/lib/*.sh` and `defaults/hooks/` for the same pattern

| File | Uses `BASH_SOURCE` for sibling/self dir resolution? | Exposed to this bug today? |
|---|---|---|
| `defaults/scripts/lib/disk-headroom.sh:44` | Yes — `${BASH_SOURCE[0]}` to locate + source sibling `worktree-root.sh` | **Yes** — sourced directly by `sweep.md` into the invoking (possibly zsh) shell |
| `defaults/scripts/lib/loom-tools.sh:96,132` | Yes — `${BASH_SOURCE[1]}` (caller's frame) to resolve repo root for `run_loom_tool` / `_loom_tool_not_found_error` | No — always sourced from within other bash-shebang'd scripts today, but same bashism; latent risk if ever sourced directly |
| `defaults/scripts/lib/forge-helpers.sh` | No — only appears in a usage-doc comment (`# source "$(dirname "${BASH_SOURCE[0]}")/lib/forge-helpers.sh"`); no internal sibling-sourcing code | No |
| `defaults/scripts/lib/pipe-pane-cmd.sh` | No — same as above, doc-comment only | No |
| `defaults/scripts/lib/worktree-root.sh` | No `BASH_SOURCE` references | No |
| `defaults/scripts/lib/classify-error.sh` | No `BASH_SOURCE` references | No |
| `defaults/scripts/lib/default-branch.sh` | No `BASH_SOURCE` references (only prose mentions of "source") | No |
| `defaults/hooks/skill-router.sh:28` | Yes — `${BASH_SOURCE[0]}` fallback for `MAIN_ROOT`, but only as a fallback after `git rev-parse --git-common-dir` | No — hooks are invoked by Claude Code as executed bash scripts (shebang'd, not sourced by the agent's shell), so `BASH_SOURCE` is always populated by the bash interpreter regardless of the operator's login shell |
| `defaults/hooks/guard-destructive.sh:35` | Yes — same fallback pattern | No — same reason (executed, not sourced) |
| `defaults/hooks/guard-loom-workflow.sh:37` | Yes — same fallback pattern | No — same reason |
| `defaults/hooks/guard-readonly-dirs.sh.template:52` | Yes — same fallback pattern | No — same reason (template, executed when installed) |
| `defaults/hooks/methodology-inject.sh:29` | Yes — same fallback pattern | No — same reason |

Bottom line: only `disk-headroom.sh` needs the fix to resolve the reported failure. `loom-tools.sh` should get the same `${BASH_SOURCE[1]:-$0}`-style hardening in the same PR for consistency/defense-in-depth, since it shares the exact same underlying bashism even though no current call site triggers it.

### Acceptance criteria

- [ ] `defaults/scripts/lib/disk-headroom.sh` line 44 resolves its own directory using a portable idiom (`${BASH_SOURCE[0]:-$0}`, empirically verified above) instead of bare `${BASH_SOURCE[0]}`.
- [ ] `source ./.loom/scripts/lib/disk-headroom.sh` (and the `defaults/scripts/lib/` copy) succeeds from the repo root in **both** `bash` and `zsh`, verified by an automated test (not just manual verification).
- [ ] `defaults/scripts/lib/loom-tools.sh` (`run_loom_tool`, `_loom_tool_not_found_error`) gets the same `${BASH_SOURCE[1]:-$0}` hardening for defense-in-depth, with a regression test if practical (may be harder to test given the frame-index semantics — a best-effort attempt is acceptable; do not block the PR solely on this secondary item if it proves awkward to test).
- [ ] No behavior change for any existing bash call site (`agent-spawn.sh`, `worktree.sh`, `sync-labels.sh`, `merge-pr.sh`, etc. — all the sites enumerated by `grep -rn "source.*lib/"` above) — these already pass because they execute a bash-shebang'd script rather than sourcing a lib file directly into an interactive shell.
- [ ] Audit note from this curation (the table above) is preserved somewhere discoverable — either as a code comment in `disk-headroom.sh` near the fix, or folded into the PR description — so a future contributor doesn't have to re-derive which lib files are exposed.

### Test plan

1. Add a new test (or extend `defaults/scripts/tests/test-disk-headroom.sh`) that explicitly exercises the zsh-sourcing path, since the existing test file is itself `#!/usr/bin/env bash` and is *executed* (not sourced from a zsh shell), which is exactly why it didn't catch this bug. Something like:
   ```bash
   # Regression guard for #3680: disk-headroom.sh must be sourceable directly
   # into a zsh shell (not just executed under bash).
   if command -v zsh >/dev/null 2>&1; then
       out="$(zsh -c "cd '$SCRIPTS_DIR/..' && source ./lib/disk-headroom.sh && echo OK" 2>&1)"
       assert_eq "$out" "OK" "disk-headroom.sh sources cleanly under zsh"
   fi
   ```
   (Adjust cwd/relative path to mirror `sweep.md`'s exact `source ./.loom/scripts/lib/disk-headroom.sh` invocation shape.)
2. Run the existing `defaults/scripts/tests/test-disk-headroom.sh` under both `bash defaults/scripts/tests/test-disk-headroom.sh` and `zsh defaults/scripts/tests/test-disk-headroom.sh` (or however the harness normally executes tests) to confirm no regression in the pure-math test coverage.
3. Manually reproduce the original failure pre-fix (`zsh -c "source ./.loom/scripts/lib/disk-headroom.sh"` from repo root fails), then confirm it succeeds post-fix, in both `defaults/scripts/lib/disk-headroom.sh` and the installed `.loom/scripts/lib/disk-headroom.sh` copy (installer/sync mechanism should propagate automatically, but worth a spot check).
4. If `loom-tools.sh` is also patched, run the existing test suite for scripts that source it (`agent-spawn.sh`, `agent-wait.sh`, `validate-phase.sh`, `checkpoint.sh`, `clean.sh`, `agent-metrics.sh`, `recover-orphaned-shepherds.sh`, `cleanup.sh`) to confirm zero behavior change.

### Affected files

- `defaults/scripts/lib/disk-headroom.sh` (primary fix, line 44)
- `defaults/scripts/lib/loom-tools.sh` (secondary hardening, lines 96 and 132)
- `defaults/scripts/tests/test-disk-headroom.sh` (add zsh-sourcing regression test)
- `.loom/scripts/lib/disk-headroom.sh` / `.loom/scripts/lib/loom-tools.sh` (installed copies — verify they pick up the fix via the normal sync/install mechanism)

No changes needed in `defaults/hooks/*.sh` or `defaults/scripts/lib/forge-helpers.sh` / `pipe-pane-cmd.sh` / `worktree-root.sh` / `classify-error.sh` / `default-branch.sh` — see audit table above.

rjwalters:
## Summary

`disk-headroom.sh` could not be `source`d directly into a zsh shell — its bare `${BASH_SOURCE[0]}` self-path resolution silently mis-resolved the sibling `worktree-root.sh` to the shell's CWD (repo root) under zsh, where `BASH_SOURCE` is unset. This broke `/loom:sweep`'s Stage -1 "Resolve auto wave size" step, which sources the file directly per `sweep.md`'s own documented usage (macOS interactive/Claude-Code shells are zsh).

```
./.loom/scripts/lib/disk-headroom.sh:source:46: no such file or directory: .../worktree-root.sh
```

## Changes

- **`defaults/scripts/lib/disk-headroom.sh`** (primary fix, line 44): `${BASH_SOURCE[0]}` → the empirically-verified portable `${BASH_SOURCE[0]:-$0}` idiom. zsh sets `$0` to the sourced file's path, so the fallback recovers it; bash keeps using `BASH_SOURCE[0]` unchanged — a strict superset with no behavior change for any existing bash call site. A comment near the fix preserves the full audit-table disposition.
- **`defaults/scripts/lib/loom-tools.sh`** (secondary hardening, lines 96/132): `${BASH_SOURCE[1]}` → `${BASH_SOURCE[1]:-$0}` in `run_loom_tool` / `_loom_tool_not_found_error` for defense-in-depth. No current call site sources this file directly into a zsh shell (every site executes a bash-shebang'd script), so bash behavior is byte-for-byte unchanged, but it shares the same latent bashism.
- **`defaults/scripts/tests/test-disk-headroom.sh`** (Test 7): new zsh-sourcing regression guard that shells out to `zsh -c` to exercise the reported path (the existing tests all run under bash and so never caught it). Guarded by `command -v zsh`; skips gracefully when zsh is absent.

## Audit

Confirmed `disk-headroom.sh` is the **only** file exposed to this bug today. Every other `defaults/scripts/lib/*.sh` and `defaults/hooks/*.sh` that references `BASH_SOURCE` is only ever *executed* as a bash-shebang'd script (never sourced directly into a possibly-zsh top-level shell), so `BASH_SOURCE` is always populated for them. Full per-file table is in the issue.

## Test results

- `bash defaults/scripts/tests/test-disk-headroom.sh` → **17/17 pass** (13 pre-existing + 4 new zsh assertions).
- Manual zsh smoke: `zsh -c 'cd <worktree> && source ./defaults/scripts/lib/disk-headroom.sh && loom_worktree_root_free_gb .'` → returns `280` (source succeeds, function works).
- Pre-fix repro confirmed the bare idiom resolves to the repo root under zsh.
- `shellcheck` clean (exit 0) on both libs and the test file.

Closes #3680

### DIFF
--- a/defaults/scripts/lib/disk-headroom.sh
+++ b/defaults/scripts/lib/disk-headroom.sh
@@ -41,7 +41,20 @@
 #   floor       — the math produced < 1 (e.g. a nearly full disk); floored to 1.
 
 # Resolve worktree-root.sh relative to this file so sourcing works from any cwd.
-_LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+#
+# ${BASH_SOURCE[0]:-$0} is a bash+zsh-portable self-path idiom (#3680). This file
+# is unique among defaults/scripts/lib/*.sh in that sweep.md's Stage -1 sources it
+# DIRECTLY into the invoking shell (`source ./.loom/scripts/lib/disk-headroom.sh`),
+# which on macOS is often zsh (Claude Code's Bash tool runs zsh; zsh is also the
+# interactive default). Under zsh, BASH_SOURCE is unset, so bare ${BASH_SOURCE[0]}
+# -> "" and dirname "" -> ".", resolving worktree-root.sh against the shell's CWD
+# (repo root) instead of this lib dir — the source then fails. zsh sets $0 to the
+# sourced file's path, so the :-$0 fallback recovers it; bash keeps using
+# BASH_SOURCE[0] unchanged. Every other lib file is only sourced from within
+# bash-shebang'd scripts that are executed (not sourced into the top-level shell),
+# so BASH_SOURCE is always populated for them and they need no change. See the
+# audit table on #3680 for the full file-by-file disposition.
+_LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 # shellcheck source=./worktree-root.sh
 source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"
 
--- a/defaults/scripts/lib/loom-tools.sh
+++ b/defaults/scripts/lib/loom-tools.sh
@@ -92,8 +92,13 @@ run_loom_tool() {
     local full_cli="loom-${cli_name}"
 
     # Try to find loom-tools directory (source takes precedence)
+    # ${BASH_SOURCE[1]:-$0} hardens the caller-frame lookup for the zsh case
+    # (#3680, defense-in-depth). No current call site sources this file directly
+    # into a zsh shell — every site executes a bash-shebang'd script, so
+    # BASH_SOURCE is always populated and bash behavior is byte-for-byte
+    # unchanged — but the bare bashism would break if that ever changed.
     local script_dir
-    script_dir="$(cd "$(dirname "${BASH_SOURCE[1]}")" && pwd)"
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[1]:-$0}")" && pwd)"
 
     if find_loom_tools "$script_dir"; then
         # Try Python module with PYTHONPATH first (always current source)
@@ -128,8 +133,9 @@ _loom_tool_not_found_error() {
     echo "" >&2
     echo "To install:" >&2
 
+    # ${BASH_SOURCE[1]:-$0} — see run_loom_tool above (#3680 defense-in-depth).
     local script_dir repo_root
-    script_dir="$(cd "$(dirname "${BASH_SOURCE[1]}")" && pwd)"
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[1]:-$0}")" && pwd)"
     if repo_root="$(_find_repo_root "$script_dir")"; then
         if [[ -d "$repo_root/loom-tools" ]]; then
             echo "  cd $repo_root && pipx install --editable ./loom-tools" >&2
--- a/defaults/scripts/tests/test-disk-headroom.sh
+++ b/defaults/scripts/tests/test-disk-headroom.sh
@@ -140,6 +140,37 @@ fi
 
 rm -rf "$STUBDIR" "$REPO" "$SCRATCH"
 
+# --- Test 7: disk-headroom.sh is sourceable directly under zsh (#3680) ---
+echo ""
+echo "Test 7: disk-headroom.sh sources cleanly under zsh (regression guard for #3680)"
+# The original bug: sweep.md's Stage -1 does `source ./.loom/scripts/lib/disk-headroom.sh`
+# directly into the invoking shell, which on macOS is frequently zsh. Under zsh
+# BASH_SOURCE is unset, so the bare `${BASH_SOURCE[0]}` resolved the sibling
+# worktree-root.sh against the CWD (repo root) instead of the lib dir, and the
+# source failed with "no such file or directory: .../worktree-root.sh". The fix
+# is the portable `${BASH_SOURCE[0]:-$0}` idiom. This test is itself executed
+# under bash (the harness shebang), so it must shell out to zsh explicitly to
+# exercise the reported path — that is exactly why the pre-existing tests above
+# (all run under bash) never caught the bug.
+if command -v zsh >/dev/null 2>&1; then
+    # Mirror sweep.md's invocation shape: cd into the dir that holds `lib/` and
+    # source via a leading-./ relative path. Here that dir is $SCRIPTS_DIR.
+    zsh_out="$(zsh -c "cd '$SCRIPTS_DIR' && source ./lib/disk-headroom.sh && echo OK" 2>&1)" || true
+    if [[ "$zsh_out" == *OK* ]]; then
+        pass "disk-headroom.sh sources cleanly under zsh (relative ./lib/ path)"
+    else
+        fail "disk-headroom.sh failed to source under zsh: $zsh_out"
+    fi
+
+    # Also exercise the functions post-source under zsh — proves the sibling
+    # worktree-root.sh loaded (loom_worktree_root_free_gb calls loom_worktree_root
+    # from it) and the pure math function works in a zsh interpreter.
+    zsh_math="$(zsh -c "cd '$SCRIPTS_DIR' && source ./lib/disk-headroom.sh && loom_wave_size_from_disk daemon 20 100 | tr '\n' '|'" 2>&1)" || true
+    assert_eq "$zsh_math" "10|target|" "loom_wave_size_from_disk works under zsh after sourcing"
+else
+    echo "  SKIP: zsh not available on PATH — skipping zsh-sourcing regression test"
+fi
+
 # --- Summary ---
 echo ""
 echo "Tests run: $TESTS_RUN, Passed: $TESTS_PASSED, Failed: $TESTS_FAILED"


rjwalters:
LGTM — approving.

## Verdict: APPROVE

Clean, tightly-scoped zsh-sourcing portability fix (+53/-3). Verified locally: 17/17 tests pass including the new Test 7 zsh path.

### Idiom correctness

- **Primary fix (`disk-headroom.sh:44`, `${BASH_SOURCE[0]:-$0}`)** — correct. At a sourced file's *top level* (not inside a function), zsh with the default `FUNCTION_ARGZERO` sets `$0` to the sourced file path, so the fallback recovers the lib dir; bash keeps using `BASH_SOURCE[0]` unchanged (it wins whenever set). Strict superset, zero bash behavior change. Test 7 proves the sibling `worktree-root.sh` actually loads (the `loom_wave_size_from_disk` assertion depends on it).
- **Secondary hardening (`loom-tools.sh:96/132`, `${BASH_SOURCE[1]:-$0}`)** — honest defense-in-depth. Non-blocking observation: `[1]` is the caller frame, and inside a zsh function `$0` is the function name, so under zsh the fallback resolves to CWD — identical to the bare bashism, so it neither helps nor harms there. Bash behavior is byte-for-byte unchanged since every real call site is a bash-shebang'd executed script with `BASH_SOURCE` populated. Correctly documented as latent, not an active fix.

### Caveat noted

The fix depends on zsh's default `FUNCTION_ARGZERO` being on. `unsetopt FUNCTION_ARGZERO` would still break, but that's a non-default, self-inflicted edge case — the default-shell path (the one `sweep.md` Stage -1 actually hits) is fixed and verified.

### Test quality

Test 7 shells out to `zsh -c` explicitly (the harness runs under bash — exactly why the prior 13 tests missed this), mirrors sweep.md's `cd`-then-`source ./lib/...` shape, guards on `command -v zsh`, and skips gracefully when zsh is absent. Asserts both source success and post-source function execution.

CI green (Shellcheck + Installer Integration pass, Rust jobs path-skipped), mergeStateStatus CLEAN, mergeable MERGEABLE.

Issue #3665, opened 2026-07-20
builder: partial-increment PRs against family/epic issues hardcode 'Closes #N' and wrongly close the parent


rjwalters:
## Problem

The builder skill (`builder-pr` / `.claude/commands/loom/builder.md`) hardcodes `Closes #N` in every PR body. For **family / epic issues** — issues that enumerate many sub-items and can only be resolved incrementally over multiple PRs — a single PR is a **partial increment**, not a full resolution. When such a PR merges, GitHub's `Closes #N` parser auto-closes the parent family issue even though the vast majority of its scope remains.

### Concrete example (downstream: vibesql)

The SQLite TCL-conformance epic (vibesql#5779) has phase issues like:

> conformance(tcl): expression / type-affinity / literal semantics (~497 certified failures: e_expr, nan, istrue, like3, ...)

A builder that fixes even 50 of those 497 failures opens a PR with `Closes #6172`. On merge, #6172 closes — silently dropping the remaining ~447 failures from the tracked backlog. The operator must then remember to reopen or re-file.

Today the downstream workaround (per our runbook) is a manual, per-PR intervention: **rewrite `Closes #N` → `Part of #N` in both the PR body and the commit message before merge**. That's error-prone and doesn't scale across a 12-PR sweep.

## Requested fix

Give the builder a way to emit a **non-closing** reference (`Part of #N` / `Updates #N` / `Related to #N`) instead of `Closes #N` when the target is a partial-increment issue. Options, roughly in order of preference:

1. **Label-driven**: if the issue carries a `loom:epic`, `loom:epic-phase`, or a new `loom:family` / `loom:partial-ok` label, the builder uses `Part of #N` automatically.
2. **Body-heuristic**: detect enumerated-backlog phrasing (e.g. "~N certified failures", checklists with many unchecked items) and default to `Part of #N`, or at least prompt.
3. **Explicit builder directive**: honor a per-dispatch flag / instruction (e.g. `--partial` or an operator note) that flips the reference keyword.

## Impact

Any Loom user running `/sweep` or `/shepherd` against epic-decomposed backlogs hits this. Partial progress on a family issue currently either (a) prematurely closes the parent, losing the remaining scope, or (b) forces manual per-PR body+commit surgery. A first-class non-closing-reference path removes the footgun.

---
Filed from a downstream vibesql sweep session (rjwalters/vibesql, epic #5779).

rjwalters:
[...]

Issue #3311, opened 2026-05-24
dogfood: gitignore .claude/agents + installer creates symlink in --dogfood mode


rjwalters:
## Summary

Child issue 2/3 from #3305 (Architect proposal Option E — dogfood symlink path for loom-on-loom development).

When running Loom *on* the Loom repo itself, copying `defaults/.claude/agents/loom-*.md` into `.claude/agents/` would silently drift from the source of truth. A gitignored symlink keeps single-source-of-truth without committing the symlink (matching how `.claude/` is conventionally per-user/generated).

## Scope (Architect-specified)

- `.gitignore` — add `.claude/agents` (currently untracked; this formalizes it)
- `scripts/install-loom.sh` — detect `pwd == LOOM_ROOT` (or accept `--dogfood` flag) and create `.claude/agents -> ../defaults/.claude/agents` symlink instead of copying
- **Bonus cleanup** (optional, separate PR if growing too big): consider migrating `.claude/commands` (currently a committed symlink) to the same gitignored-symlink pattern for consistency. Architect leans toward "leave it" — the inconsistency is mildly ugly but touches more surface area.

## Acceptance criteria

- [ ] Fresh `git clone` of loom + `./install.sh .` produces working `.claude/agents/` symlink in the loom repo
- [ ] `git status` stays clean (symlink is gitignored)
- [ ] `Agent(subagent_type="loom-doctor", ...)` works in this checkout without manual setup

## Out of scope

- Fresh-install copy path — covered by sibling child issue
- README/docs rewrite — covered by sibling child issue

## Dependency

May depend on the installer copy issue being implemented first (the dogfood detection branches off the same install code path). Builder can sequence appropriately.

## Parent

#3305

PR #3690, opened 2026-07-21
fix(dogfood): scoped-symlink .claude/commands/loom to defaults/ to stop drift (#3682)

### LINKED ISSUE #3682: dogfood .claude/commands/loom copies go stale — caused a false bug report (#3665); add a sync check or auto-sync
## Context
During the 2026-07-20 /loom:sweep all run, issue #3665 (builder hardcodes `Closes #N` on epics) turned out to be already fixed on main by PR #3599 — but this repo's own installed dogfood copies at `.claude/commands/loom/{builder,builder-pr,judge,guide}.md` predate that merge and still carry the old guidance (confirmed by diff against `defaults/.claude/commands/loom/`). The stale copies are what agents in this repo actually load, and the drift is invisible until it produces wrong behavior or a false bug report. The same staleness now also applies to today's sweep.md changes (#3673). This is the recurring Loom-on-Loom wiring pain (defaults/ ↔ .claude/ ↔ .loom/).

- [ ] A cheap drift check (e.g. in Auditor's tick, or a `loom-clean`-style helper, or a CI job scoped to this repo) that diffs the gitignored dogfood `.claude/commands/loom/` copies against `defaults/.claude/commands/loom/` and reports staleness loudly.
- [ ] A one-command re-sync path documented (or automated post-merge) so the dogfood copies track main.
- [ ] Immediate: re-sync this repo's dogfood copies to current main.

---

### How dogfood mode materializes `.claude/commands/loom/`

`scripts/install-loom.sh` (~line 1030-1091) does **not** symlink `.claude/commands/loom/` — it performs an atomic **copy** from `defaults/.claude/commands/loom/` into the gitignored, real `.claude/commands/loom/` directory (temp-dir build + `mv` swap). This is a deliberate design from #3565 ("Decouple loom's live `.claude/commands` from shipped defaults/ (co-install safety)"): the *whole* `.claude/commands` directory was previously a symlink into `defaults/`, and a co-installed tool (`repo`'s `install.sh`, which writes to `.claude/commands/<ns>/`) silently wrote its own commands *through* that symlink into `defaults/.claude/commands/repo/` — i.e. it staged `repo`'s commands into loom's own shipped distribution artifact. #3565 fixed that by making `.claude/commands` a real, gitignored destination directory (mirroring every consumer repo), with only the `loom/` subdir *populated* by copying from `defaults/` at install time. `.gitignore` (lines 92-104) documents this rationale explicitly and calls out that `.claude/commands/loom/` must stay **tracked** in consumer repos (not gitignored there) — this ignore rule is loom-source-repo-only.

Contrast with `.loom/scripts`, which **is** a real symlink (`.loom/scripts -> ../defaults/scripts`, confirmed via `ls -la`) and with `.claude/agents`, which is also symlinked per the #3311 dogfood mechanism. Both of those are single-purpose, loom-only directories with no co-install pollution risk, so a directory symlink is safe there. `.claude/commands`, however, is a *shared* namespace directory (siblings like `.claude/commands/repo/` from the `repo` tool coexist next to `.claude/commands/loom/`), which is exactly why #3565 rejected symlinking it.

**This issue is not moot** — the copies are real files, confirmed by `diff -rq .claude/commands/loom/ defaults/.claude/commands/loom/`.

### Drift snapshot (today, before any re-sync)

5 of the dogfood copies differ from `defaults/`:

| File | Diff line count |
|---|---|
| `builder-pr.md` | 40 |
| `builder.md` | 29 |
| `guide.md` | 21 |
| `judge.md` | 89 |
| `sweep.md` | 194 |

The live copies carry an mtime of Jul 15 13:33 (this repo's last dogfood re-sync), while `defaults/.claude/commands/loom/builder.md` was last touched by commit `eb6cfdd8` (Jul 16, #3599 — the fix that #3665 falsely reported as missing). All 5 files postdate the last dogfood materialization, confirming the drift mechanism described in the issue.

### Recommended approach

Prefer a **structural fix that eliminates this drift class**, rather than only adding a checker, but scope it narrowly given #3565's prior (correct) rejection of symlinking the whole `.claude/commands` directory:

- **Option A (preferred, if validated safe): symlink only the `loom/` subdir**, not the parent — `.claude/commands/loom -> ../../defaults/.claude/commands/loom` — leaving `.claude/commands/` itself a real directory so sibling namespaces (e.g. `.claude/commands/repo/`) remain unaffected, exactly mirroring the `.claude/agents` and `.loom/scripts` pattern but scoped one level deeper. This sidesteps #3565's actual failure mode (a co-installed tool writing into a *different* namespace through a symlinked *parent* dir) because no tool other than loom's own installer is expected to write into the `loom/` namespace specifically. **Before implementing**, the builder should double check: (a) nothing besides `scripts/install-loom.sh`'s materialization step ever writes into `.claude/commands/loom/` at runtime (a quick grep turned up no other writers, but re-verify), and (b) `.gitignore`'s existing comment block (lines 92-104) explicitly considered and rejected "a per-file symlink into defaults/" for the *reason above* — re-read that rationale and confirm the subdirectory-scoped symlink doesn't reintroduce it before committing to this approach. If it does not hold up, fall back to Option B.
- **Option B (fallback): drift check + one-command re-sync.** If the scoped symlink is ruled out, add a cheap diff check (Auditor tick, or a `loom-clean`-style helper such as `.loom/scripts/check-dogfood-drift.sh`) that runs `diff -rq .claude/commands/loom/ defaults/.claude/commands/loom/` and fails loudly (non-zero exit + clear message) when they differ, plus document/automate the one-command re-sync (re-running the dogfood materialization block of `scripts/install-loom.sh`, or extracting it into a standalone `--sync-dogfood`-style flag / script).
- Either way, this repo's own copies need an **immediate re-sync** to current `main` as part of this issue's resolution (builder's job, not curator's — left untouched here per instructions).

### Acceptance criteria

- [ ] Investigate and decide between Option A (scoped `loom/`-subdir symlink) and Option B (drift checker + documented re-sync); record the decision and rationale in the PR description.
- [ ] If Option A: implement the scoped symlink in `scripts/install-loom.sh` (replacing the copy-and-swap block at ~line 1030), update `.gitignore`'s comment (lines 92-104) to reflect the new mechanism, and verify a co-installed tool writing to a sibling namespace (e.g. `.claude/commands/repo/`) is unaffected.
- [ ] If Option B: add a drift-check script/CI job/Auditor-tick step that diffs `.claude/commands/loom/` against `defaults/.claude/commands/loom/` and fails loudly on mismatch; document the one-command re-sync path (e.g. in `.loom/docs/troubleshooting.md` or the installer's own `--help` text).
- [ ] Re-sync this repo's dogfood copies (`.claude/commands/loom/{builder,builder-pr,judge,guide,sweep}.md` at minimum) to current `main` — confirmed via `diff -rq .claude/commands/loom/ defaults/.claude/commands/loom/` reporting no differences.
- [ ] No regression to consumer-repo installs: `.claude/commands/loom/` must remain **tracked** (real files, not symlinked) in repos other than loom's own source repo — verify `scripts/install-loom.sh`'s non-dogfood path and `loom-daemon`'s `init` path are untouched by this change.

### Test plan

- `diff -rq .claude/commands/loom/ defaults/.claude/commands/loom/` reports no differences after re-sync.
- If Option A: `./scripts/install-loom.sh --dogfood .` (or equivalent re-run) followed by writing a test file into `.claude/commands/repo/` (or another sibling namespace) confirms it does NOT land in `defaults/.claude/commands/repo/`.
- If Option B: intentionally introduce drift (edit one dogfood copy without touching `defaults/`) and confirm the new check fails loudly; then re-sync and confirm it passes.
- Existing installer tests (`scripts/test-installer.sh`, `tests/install/*`) still pass.

### Relevant files

- `scripts/install-loom.sh` (dogfood materialization, ~lines 1030-1091)
- `.gitignore` (lines 84-104, dogfood/materialization rationale)
- `.claude/commands/loom/` (this repo's live, gitignored copies)
- `defaults/.claude/commands/loom/` (source of truth, tracked)
- `.loom/scripts` (symlink precedent: `.loom/scripts -> ../defaults/scripts`)
- Related/prior issues: #3311 (`.claude/agents` symlink), #3565 (why `.claude/commands` is a copy, not a symlink), #3665 (the false bug report this drift caused), #3673 (sweep.md changes also affected by this drift)

rjwalters:
## Summary

Fixes the recurring dogfood drift where `.claude/commands/loom/` copies go stale (caused the false bug report in #3665). Replaces the copy-and-swap materialization in `scripts/install-loom.sh` with a **scoped symlink** of the `loom/` subdir into the shipped source of truth:

```
.claude/commands/loom -> ../../defaults/.claude/commands/loom
```

A symlink cannot drift — it always reflects the committed `defaults/`.

## Decision: Option A (scoped symlink), re-validated against #3565

#3565 rejected symlinking `.claude/commands` because the **whole** directory was a symlink into `defaults/`, so a co-installed tool writing a *sibling* namespace (`.claude/commands/repo/foo.md`, from the `repo` tool) wrote **through** that symlink into loom's shipped `defaults/` distribution artifact.

Option A avoids that failure mode by keeping `.claude/commands/` itself a **real directory** and symlinking **only** the `loom/` subdir:

- Sibling namespaces (`.claude/commands/repo/`) are real dirs — writes never redirect into `defaults/`.
- The only writer into the `loom/` namespace is loom's own installer/daemon. Verified no runtime writer: `agent_spawn.py` only **reads** `.claude/commands/loom/<role>.md`; the daemon `init` path writes only the consumer-copy (untouched here).

So #3565's parent-symlink pollution mode does not recur. Full rationale is captured in `scripts/install/dogfood-commands.sh` and the updated `.gitignore` comment.

## Changes

- **`scripts/install/dogfood-commands.sh`** (new): extracted `link_dogfood_commands <target>` — establishes the scoped symlink. Handles legacy whole-dir symlink removal, stale-copy replacement, idempotent re-link, and a local-only-file guard that refuses to silently discard files absent from `defaults/`. Extraction lets the test suite exercise the logic without running the full installer (dogfood only fires when `TARGET == LOOM_ROOT`).
- **`scripts/install-loom.sh`**: source the helper; replace the ~60-line copy-and-swap block with a call to `link_dogfood_commands`. Updated `--dogfood` help text.
- **`.gitignore`**: rewrote the comment block to describe the scoped-symlink mechanism and why it's safe.
- **`scripts/test-installer.sh`**: 10 new tests (66-75).

## Live re-sync applied

This repo's own drifted dogfood copies (5 files: builder, builder-pr, judge, guide, sweep) were re-synced by swapping the live `.claude/commands/loom/` real dir for the symlink in the **main checkout** (gitignored, not part of this PR diff). Verified `diff -rq .claude/commands/loom/ defaults/.claude/commands/loom/` reports **no differences**, sibling `.claude/commands/repo/` intact, and `defaults/.claude/commands/` contains only `loom` (no leak).

## Consumer safety

No regression to consumer installs: `.claude/commands/loom/` remains **tracked** real files in consumer repos (the daemon's `update_gitignore` list and consumer-copy `init` path are untouched).

## Tests

`bash scripts/test-installer.sh` — **131 passed, 0 failed**, including the 10 new dogfood tests:

- symlink shape + relative target
- parent `.claude/commands/` stays a real directory
- content resolves through the symlink to `defaults/`
- **#3565 safety**: sibling-namespace write does not pollute `defaults/`
- idempotency, stale-copy replacement, local-only-file preservation, legacy whole-dir symlink migration

Closes #3682


rjwalters:
LGTM — approved.

## Review summary

Verified the load-bearing #3565 re-validation and confirmed the scoped symlink is genuinely immune to the sibling-namespace failure mode.

**#3565 safety (the crux):** The prior failure was a symlinked *parent* (`.claude/commands` whole-dir → defaults/) letting a co-installed tool write a sibling namespace (`.claude/commands/repo/foo.md`) *through* into loom's `defaults/` artifact. This PR keeps `.claude/commands/` a **real directory** and symlinks only the `loom/` subdir. Confirmed the live main-checkout state matches the design exactly: `.claude/commands/loom -> ../../defaults/.claude/commands/loom` (symlink) sits alongside a **real** `repo/` sibling dir — sibling writes land in a real dir and never redirect into `defaults/`. Test 71 exercises exactly this.

**Writer audit:** Confirmed `agent_spawn.py` only *reads* `.claude/commands/loom/<role>.md`. The only write path into that namespace is the daemon's consumer-install scaffolding (`copy_dir_with_report_filtered` in `file_ops.rs`), which copies *from* loom's `defaults/` *into* a consumer target — it is not run against the loom source repo itself. The theoretical "future consumer path writes through the loom/ symlink into defaults/" concern is equivalent to the **already-accepted** `.claude/agents` whole-dir dogfood symlink (#3311), which shares the identical risk profile and predates this change — so this PR introduces no new exposure, and in fact scopes more tightly than the agents precedent.

**Legacy migration & guards:** The legacy whole-dir symlink removal (Test 75), stale-copy replacement (Test 73), idempotent re-link (Test 72), and the `comm -23` local-only-file guard that refuses to silently discard files absent from `defaults/` (Test 74) are all correct and covered. The local-only guard returns soft (0) leaving the real dir intact — safety over silent data loss, requires operator action. Good.

**Consumer safety:** Untouched — the daemon `update_gitignore` list and consumer-copy init path are unchanged; consumer repos keep `.claude/commands/loom/` as tracked real files.

**Code quality:** Clean extraction into `scripts/install/dogfood-commands.sh` with fallback logging helpers so the test harness can drive the logic in isolation. Split `local` declaration/assignment avoids masking return values. Relative symlink target keeps the repo relocatable.

**Validation:** `bash scripts/test-installer.sh` → **131 passed, 0 failed** (10 new dogfood tests 66-75). CI Installer Integration Tests pass; merge state CLEAN.

Solid, well-reasoned, thoroughly tested. Approving.

PR #3692, opened 2026-07-21
perf(guards): read-only fast path for guard-destructive.sh (#3687)

### LINKED ISSUE #3687: guard-destructive.sh (PreToolUse/Bash) fires on every Bash call — add a read-only fast-path so Bash-dense sessions aren't dominated by the guard
## Observation

`.loom/hooks/guard-destructive.sh` is registered as a **PreToolUse hook matched on `Bash`**, so it runs **before every single Bash tool call**. In Bash-dense sessions — e.g. driving a remote benchmark box, where nearly every step is `ssh` / `aws describe` / `git status` / `grep` — the guard fires dozens or hundreds of times per turn, and its per-invocation work (read stdin, `jq`-parse the tool input, run the deny/ask pattern matching, best-effort logging) runs in full even for trivially read-only commands.

This is functioning as designed (the guard *should* see every command), but the frequency/overhead is very visible in these workflows and there's no cheap short-circuit for the overwhelmingly-common "obviously safe read-only command" case.

## Why it matters

- **Overhead**: every `ssh host 'grep ...'` or `aws ... describe` pays the guard's full parse+match cost. Across a long remote-ops session that's a lot of redundant work on commands that could never be destructive.
- **Noise / observability**: operators notice the guard "triggering constantly" and can't easily tell signal (a real deny/ask) from the steady background of allow-path firings.

## Requested improvement (any of these)

1. **Read-only fast-path allowlist.** Before the full deny/ask matching, short-circuit to `allow` (exit 0, zero output) for a configurable set of well-known read-only commands / prefixes — e.g. `git status|log|diff|show`, `ls`, `cat`, `grep`, `rg`, `aws * describe*|get*|list*`, `gh * view|list`, plain `ssh <host> '<read-only>'`. This keeps the guard authoritative for mutating commands while making the common case near-free.
2. **Guarantee the allow path is truly silent and minimal.** Confirm an `allow` decision emits no stdout/stderr and does the least work possible (skip logging, skip non-essential parsing) so it's invisible when nothing is flagged.
3. **Document a supported way to scope the matcher.** Guidance (or a config knob in `.loom`) for operators who want to narrow the PreToolUse matcher for specific workflows without hand-editing Loom-managed `settings.json` (which the installer may overwrite).

## Context

Surfaced during a downstream vibesql session doing a full overnight AWS benchmark run — the session was ~all Bash (ssh/aws/git), so the per-Bash guard was the dominant hook activity. The guard itself is working correctly (it correctly blocked a `shutdown` command mid-session); this is purely about making the allow path cheap and quiet at high call volume.

Downstream repo: rjwalters/vibesql. Hook: `.loom/hooks/guard-destructive.sh` (manifest version 2).

---

### Control-flow map of `defaults/hooks/guard-destructive.sh` (current main, post-#3676/#3686)

The hook has **no early classification today** — every Bash call, including a bare `git status`, runs the *entire* gauntlet below before falling through to the trailing `exit 0` (line 1079). Measured on this machine, a full pass on `git status` averages **~179ms** per invocation (the existing performance check in `tests/hooks/test-guard-destructive.sh` already benchmarks exactly this command). Breaking down the cost:

| Step | Lines | Approx. cost | Notes |
|---|---|---|---|
| `INPUT=$(cat)` + 2x `jq -r` (`COMMAND`, `CWD`) | 52, 60-61 | ~3ms each `jq` call (2 calls) | Unconditional |
| `REPO_ROOT=$(git -C "$CWD" rev-parse --show-toplevel)` | 68-75 | ~7ms | Unconditional subprocess fork, needed today by every downstream toggle (`rm_scope_repo_enabled`, `force_scope_mode`, etc.) |
| `ALWAYS_BLOCK_PATTERNS` loop | 467-539 | ~11 `echo \| grep -qiE` forks | Runs for every command regardless of shape |
| Comment-strip (`sed`, conditional on `#`) | 555-559 | 0 for no-`#` commands | Cheap gate already exists here |
| `lifecycle_or_cloud_reason` (awk) | 583-640 | 1 `awk` fork | Unconditional |
| SQL DDL grep | 651-655 | 1 `grep` fork | Unconditional |
| rm-scope cheap pre-check grep | 746 | 1 `grep` fork | Already a cheap gate (only descends into `awk`+loop if it matches) |
| DELETE-without-WHERE grep x2 | 872-875 | 2 `grep` forks | Unconditional |
| Force-op cheap pre-check (`[[ *git* ]]` + grep) | 896-897 | 1 `grep` fork (bash substring test is free) | Already a cheap gate |
| `ASK_PATTERNS` loop | 940-988 | ~18 `echo \| grep -qE` forks | Runs for every command |
| `git read-tree` grep | 1010 | 1 `grep` fork | Unconditional |
| `CLOUD_ASK_PATTERNS` loop | 1036-1064 | 3 `echo \| grep -qE` forks (cloud_guard_enabled lazy) | Runs for every command |

Isolated timings on this box: single `jq -r` call ~3ms, single `git -C ... rev-parse` ~7ms, single `echo \| grep -qE` ~3ms. With ~2 `jq` + 1 `git` + ~37 `grep`/`awk` forks in the unconditional path, that's ~120-150ms of pure fork overhead before the trailing `exit 0` — consistent with the measured 179ms (plus another ~11ms for the outer `bash -c` wrapper in `.claude/settings.json` that resolves the repo root and re-execs the script, which sits *outside* the script itself).

**Key finding**: the guard already has several *lazy, cached config-read* idioms to build on — `sql_guard_enabled()` (line 99), `cloud_guard_enabled()` (line 142), `rm_scope_repo_enabled()` (line 196), `force_scope_mode()` (line 279) — each resolves `.loom/config.json` only *after* a candidate pattern has already matched, caching the result in a `_..._CACHE` variable. **There is no equivalent early *allow* classification** — nothing short-circuits before the deny/ask gauntlet runs. The fast path this issue requests should slot in as a **new, first** check — immediately after `COMMAND` is parsed (line 66), and ideally **before** the unconditional `REPO_ROOT` git subprocess (lines 68-75), since the built-in fast-path allowlist below never needs `REPO_ROOT` for its *structural* eligibility test (only the config-toggle read needs it, and only for commands that already passed the structural test — mirroring the existing lazy pattern).

**Implementation-technique warning**: the existing idiom for pattern matching in this file is `echo "$X" | grep -qE "$pattern"`, which forks *two* processes per check. If the fast-path admission test is implemented the same way, it partially defeats its own purpose. The admission test should use bash's **builtin** `[[ "$COMMAND" =~ ... ]]` regex / `case` glob matching (zero forks), reserving any subprocess (the lazy config `jq` read) for the rare case where structural admission already succeeded.

### Security framing: fast path = guard bypass by construction

Any read-only fast path is, by definition, a way to skip the deny/ask matching that would otherwise run — so admission must be conservative and structural, never content-sensitive. Recommended eligibility test (**all** of the following must hold, or the command falls through to the existing full path unchanged):

1. **No chaining, piping, redirection, or substitution anywhere in the raw command.** Reject fast-path eligibility if `COMMAND` contains any of `;`, `&`, `|`, `<`, `>`, a backtick, `` $( ``, or a newline. This is a single bash `[[ =~ ]]` test against the *raw* string (mirrors the same separator set the file's own `awk` helpers already split on: `;|&` / `&&` / `||`). This alone kills the motivating counter-example: `git status && git push --force origin main` contains `&` and is excluded from the fast path, so it takes the full path and the existing `ALWAYS_BLOCK_PATTERNS` catastrophic scan still runs. **False negatives (denying fast-path eligibility) are safe — they just fall through to the (correct, slower) existing behavior. False positives are the only real danger, so the eligibility test must stay maximally conservative.**
2. **Command-word (first token) exact match against a fixed built-in allowlist — never a wrapper.** Because admission is an *allowlist* keyed on the literal first token, wrapper forms (`bash -c '...'`, `sh -c '...'`, `eval ...`, `xargs ...`, `env ... git status`, `sudo git status`) are automatically excluded without needing a separate denylist — their first token (`bash`, `eval`, `xargs`, `env`, `sudo`) simply isn't in the allowlist, so they fall through to the full path. (Do **not** special-case-strip a leading `sudo`/`env` for the fast path the way `lifecycle_or_cloud_reason`/`extract_rm_targets` do for their checks — that stripping exists precisely because those checks *must* still catch `sudo halt`; the fast path should have the opposite bias and simply decline eligibility for anything wrapped.)
3. **Verb/subcommand exactness for multi-word tools**, chosen to be provably disjoint from every existing deny/ask pattern in this file:
   - `git status`, `git log`, `git diff`, `git show` — exact subcommand, no leading `-C`/global git options (so `git -C /path status` is *not* admitted — falls through, safe, just unoptimized). Verified: none of these four appear in `ALWAYS_BLOCK_PATTERNS` or `ASK_PATTERNS`; existing tests already assert `allow` for bare `git status`/`git diff`/`git log`.
   - `ls`, `grep`, `rg` — bare read-only utilities that never appear in any deny/ask pattern in this file.
   - `gh <noun> view`, `gh <noun> list` (verb must be exactly `view` or `list`) — disjoint from the existing `gh repo delete`/`gh repo archive` (ALWAYS_BLOCK) and `gh pr close`/`gh issue close`/`gh release delete`/`gh label delete` (ASK) verbs.
   - `aws <service> describe*|get*|list*` and the special-cased `aws s3 ls` — this mirrors the **already-existing** verb-anchoring reasoning in `CLOUD_ASK_PATTERNS` (lines 1025-1034, 1048), which explicitly excludes `describe*/get*/list*/ls` from ever matching a mutating pattern. Fast-pathing these verbs does not create new exposure; it just skips redundant `grep`s that were always going to fall through to `allow` anyway.
   - **Explicitly EXCLUDE `cat` and `ssh` from the default built-in list**, despite both being named in the original request:
     - `cat` already has a narrow existing `ASK` carve-out (`cat.*/\.ssh/`, `cat.*/\.aws/credentials`, lines 980-981). A blanket `cat` fast-path would silently skip that check — a real security regression. If `cat` support is wanted later, it needs its own argument-inspection logic, which is out of scope for a "zero-parsing" fast path.
     - `ssh <host> '<cmd>'` wraps an **opaque remote command string**. Today, the *raw* `ALWAYS_BLOCK_PATTERNS` scan still runs over the full command including the quoted remote payload, so `ssh host` with a quoted destructive remote payload is still denied (see the existing comment at lines 665-671 about remote/quoted payloads still hitting the catastrophic scan). Fast-pathing any `ssh ...` invocation would remove that catastrophic-scan coverage for remote payloads — a serious regression. Recommend deferring `ssh` fast-pathing to a follow-up that would need to independently re-vet the quoted remote string (which defeats "minimal parsing" for this issue's scope).
4. **Toggle-gated, but ON by default.** Once (1)-(3) all pass — a pure structural/bash-builtin test, no subprocess — *only then* resolve `guards.readOnlyFastPath` (default `true`) via the same lazy/cached pattern as `sql_guard_enabled()` et al. This keeps the config read off the hot path for every command that doesn't structurally qualify, and off the hot path entirely when the whole feature is disabled via the fast env-var check first.
5. **Guaranteed-silent allow.** On admission, `exit 0` immediately with **zero** stdout/stderr and without calling `log_hook_error` or any other logging path — this must happen before `REPO_ROOT` resolution and before any of the deny/ask arrays run, so the total added cost for a fast-pathed command is: 1 `[[ =~ ]]` bash builtin test (no fork) + (only if structurally eligible) 1 lazy config read.

### Config design — `guards.readOnlyFastPath` (+ `LOOM_GUARD_READONLY_FASTPATH`)

Follows the existing 4-guard-toggle precedent (`guards.sqlDdl`/`LOOM_GUARD_SQL`, `guards.cloudCli`/`LOOM_GUARD_CLOUD`, `guards.rmScope`/`LOOM_RM_SCOPE`, `guards.forceScope`/`LOOM_FORCE_SCOPE`) byte-for-byte in structure:

1. **`LOOM_GUARD_READONLY_FASTPATH` env var** (highest precedence) — `0`/`false`/`no` disables; `1`/`true`/`yes` forces on. Overrides config.
2. **`.loom/config.json` -> `guards.readOnlyFastPath`** — default `true` when absent (this is a pure perf fix with a conservative, security-reviewed allowlist, so **default ON** is recommended — unlike `rmScope`'s safe-by-default flip, there is no behavior change for any command that isn't already guaranteed-`allow` today). Explicit `false` disables.
3. **Default**: `true` (fast path active).

Recommend an **additional, optional** escape hatch for operators who want to extend (never narrow) the built-in list without hand-editing `.claude/settings.json` (which the installer may overwrite): `guards.readOnlyFastPathExtra` — an array of **literal first-word commands only** (no subcommand/verb parsing for custom entries, to keep the admission code simple and auditable), e.g.:

```json
{
  "guards": {
    "readOnlyFastPath": true,
    "readOnlyFastPathExtra": ["jq", "wc"]
  }
}
```

This directly answers request #3 ("document a supported way to scope/narrow the matcher... without hand-editing Loom-managed settings.json") — the matcher registration in `.claude/settings.json` itself does not need to change (it stays `Bash`, as today); the *effective* scope of what gets fast-pathed vs. fully checked is controlled entirely through `.loom/config.json`/env vars, which are explicitly not Loom-managed/installer-overwritten files. Document this with a loud warning: any command word added here is a **guard bypass for that command word in full generality** (all arguments), so operators should only add bare, argument-independent read-only utilities — never their own scripts or anything that could wrap a mutating call.

### Acceptance Criteria

- [ ] New fast-path admission check added to `defaults/hooks/guard-destructive.sh`, positioned immediately after `COMMAND` is extracted (after line 66) and before `REPO_ROOT` resolution, implemented with bash-builtin `[[ =~ ]]`/`case` (no `grep`/`awk`/`sed` forks for the structural test).
- [ ] Admission requires: no `;`/`&`/`|`/`<`/`>`/backtick/`$(`/newline anywhere in the raw command, AND the command's first (and where relevant second/third) token exactly matches the built-in allowlist: `git status|log|diff|show` (no leading `-C`/global git options), `ls`, `grep`, `rg`, `gh <noun> view|list`, `aws <service> describe*|get*|list*`, `aws s3 ls`.
- [ ] `cat` and `ssh` are explicitly excluded from the default allowlist, with a code comment explaining why (existing `.ssh`/`.aws/credentials` ask carve-out for `cat`; opaque remote payload + existing catastrophic-scan coverage for `ssh`).
- [ ] `guards.readOnlyFastPath` (default `true`) / `LOOM_GUARD_READONLY_FASTPATH` env var gate the feature, resolved lazily (only after structural admission passes) and cached, mirroring `sql_guard_enabled()`/`cloud_guard_enabled()`.
- [ ] Optional `guards.readOnlyFastPathExtra` (array of literal first-word commands) documented as an extend-only escape hatch, with an explicit bypass warning.
- [ ] On fast-path admission, the hook exits 0 with **zero** stdout/stderr and does not invoke `log_hook_error`.
- [ ] All 307 existing `assert_deny`/`assert_ask`/`assert_allow` cases in `tests/hooks/test-guard-destructive.sh` still pass unchanged with the fast path at its default (ON).
- [ ] New regression tests added to `tests/hooks/test-guard-destructive.sh`:
  - Each built-in fast-path entry (`git status`, `git log`, `git diff`, `git show`, `ls`, `grep`, `rg`, `gh pr view`, `gh issue list`, `aws ec2 describe-instances`, `aws s3 ls`) allows, with captured stdout+stderr length == 0.
  - **Security regression — compound/wrapped forms must NOT fast-path and must still produce the pre-existing correct verdict**: `git status && git push --force origin main` -> deny (ALWAYS_BLOCK still fires); `git status; rm -rf /` -> deny; `git status $(rm -rf /)` -> full path taken; `bash -c "git status"` -> full path taken (first token isn't `git`); `git status > /etc/passwd` -> full path taken; `git -C /tmp status` -> full path taken (no fast path for non-bare subcommand form, still allowed via existing logic).
  - `cat ~/.ssh/id_rsa` still asks (fast path never engages for `cat`).
  - `LOOM_GUARD_READONLY_FASTPATH=0` and `guards.readOnlyFastPath:false` fully disable the fast path (existing full-path behavior restored byte-for-byte).
  - `guards.readOnlyFastPathExtra` extends the allowlist for a custom bare command word.
- [ ] Existing performance check (`tests/hooks/test-guard-destructive.sh`, benchmarking `git status`) is updated/extended to reflect the expected improvement — keep the existing informational (non-flaky) framing, but add a comment noting the fast path should now make this specific benchmark command dramatically cheaper than the full-path average.
- [ ] `CLAUDE.md` and `defaults/CLAUDE.md` each get a new "Read-Only Fast-Path Guard Toggle" section under "Custom Guard Hooks", modeled on the existing SQL/cloud/rmScope/forceScope sections, documenting `guards.readOnlyFastPath` / `LOOM_GUARD_READONLY_FASTPATH` / `guards.readOnlyFastPathExtra`, the built-in list, and the explicit `cat`/`ssh` exclusions + rationale.

### Test Plan

1. Run the existing suite unchanged and confirm 307/307 still pass: `./tests/hooks/test-guard-destructive.sh` (fast path defaults ON — this is the primary regression gate).
2. Add and run the new fast-path-specific assertions described above in the same suite (silence, admission, non-admission/security, toggle-off, extra-list).
3. Manually time `git status` before/after via the existing perf harness pattern (`for i in 1..20; do ... done`) to confirm a material improvement (informational, not a hard gate, consistent with the suite's existing anti-flake stance for perf assertions).
4. Confirm `LOOM_GUARD_READONLY_FASTPATH=0 ./tests/hooks/test-guard-destructive.sh` (or an equivalent full-suite run with the env var exported) still passes all 307 pre-existing cases byte-for-byte, proving the opt-out fully restores prior behavior.

### Affected Files (verified present on `origin/main`)

- `defaults/hooks/guard-destructive.sh` — canonical source; add the fast-path check here.
- `tests/hooks/test-guard-destructive.sh` — add new regression + silence + security tests; existing 307 assertions must remain green.
- `CLAUDE.md` / `defaults/CLAUDE.md` — new guard-toggle documentation section.
- `.loom/hooks/guard-destructive.sh` is the *installed* copy in this repo (already drifted from `defaults/` pre-existing this issue, e.g. it still has the pre-#3628 `rmScope` default) — it is refreshed by the install/sync tooling, not hand-edited; do not feel obligated to reconcile that unrelated drift as part of this issue, but do not widen it either.

### Notes for the Builder

- The `.claude/skills/repo/hooks/guard-destructive.sh` file is a **separately-forked** Repo Skills copy (different config-path conventions, own env-var names) — out of scope for this issue; do not edit it.
- Use bash-builtin regex/case matching for the admission test itself — reusing the file's `echo | grep -qE` idiom for the *gate* would undercut the perf goal (each such call forks two processes; the existing full path already pays for ~37 of them per invocation).

rjwalters:
## Summary

`guard-destructive.sh` is a `PreToolUse`/`Bash` hook, so it fires before **every** Bash tool call. In Bash-dense sessions (remote ops, benchmark drivers) nearly every call is obviously read-only, yet each still ran the full deny/ask gauntlet (~37 `grep`/`awk`/`sed` forks + a `git rev-parse`, ~179ms measured) before falling through to `allow`.

This adds a conservative **read-only fast path** that runs first — immediately after `COMMAND` is parsed, **before** `REPO_ROOT` resolution and every deny/ask array — short-circuiting the overwhelmingly-common case to a **silent** `allow` (exit 0, zero stdout/stderr, no logging).

**Measured: ~20ms/call (fast path) vs ~156ms/call (full path)** on `git status` — a ~7-8x improvement on this machine.

## Design

- **Structural, never content-sensitive** admission (a fast path is a guard bypass by construction). A command is fast-pathed only when it contains **none** of `;` `&` `|` `<` `>` backtick `$(` newline (kills chaining/piping/redirection/substitution) **and** its first token exactly matches the built-in allowlist:
  - `git status|log|diff|show` (bare — `git -C /p status` is not admitted)
  - `ls`, `grep`, `rg`
  - `gh <noun> view|list`
  - `aws <service> describe*|get*|list*`, and `aws s3 ls`
- **Wrappers excluded by construction** — `bash -c`, `sh -c`, `eval`, `xargs`, `env …`, `sudo …` all fail the first-token allowlist.
- **`cat` and `ssh` deliberately excluded** — `cat` has an existing `.ssh`/`.aws/credentials` ASK carve-out a blanket fast-path would skip; `ssh` wraps an opaque remote payload the catastrophic scan still covers.
- **Bash-builtin matching only** (`[[ =~ ]]`/`case`, zero forks) for the structural test; the single `jq` config read happens **only after** structural admission passes.
- **Config-ordering choice**: the fast path runs before `REPO_ROOT`, so the toggle/extra-list config is located by walking up from `CWD` to the nearest `.loom/config.json` **fork-free** (no `git rev-parse`), keeping the hot path cheap. Documented inline.

## Toggle

`guards.readOnlyFastPath` (default `true`) / `LOOM_GUARD_READONLY_FASTPATH` env — mirrors the `sqlDdl`/`cloudCli`/`rmScope`/`forceScope` precedent (env wins, lazy + cached). Optional `guards.readOnlyFastPathExtra` is an extend-only array of literal first-word commands (documented with an explicit bypass warning).

## Tests

- All **310 pre-existing** assertions still pass with the fast path default **ON** (335/335 total).
- **25 new** assertions: per-verb admission + **silence** (0-byte output), security regressions (compound `&&`/`;` force-push still denies, command-substitution denies via full path, pipe-disqualified `grep <ddl> | cat`, `bash -c` wrapper not admitted, `git -C /tmp status` not fast-pathed, `cat ~/.ssh/id_rsa` still asks), env + config toggle-off, env-over-config precedence, and the extra-list knob.
- A `LOOM_GUARD_READONLY_FASTPATH=0` full-suite run stays green (**333/333**; the 2 ambient-default-ON observability tests are skipped), proving the opt-out restores prior behavior byte-for-byte.
- `shellcheck` clean (only the pre-existing info-level `SC2016` on the intentional single-quoted pattern arrays).
- Force-push / SQL-DDL literals in new tests are assembled from shell fragments (session-guard-safe), mirroring the existing #3679 test convention.

## Docs

`defaults/CLAUDE.md` gets a full "Read-Only Fast-Path Guard Toggle" section under Custom Guard Hooks alongside the sibling toggles. The root `CLAUDE.md` has diverged from `defaults/` (no Custom Guard Hooks section — known self-install dogfood drift), so a self-contained section documenting this toggle was added there too.

Closes #3687

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -409,6 +409,34 @@ gh label sync --file .github/labels.yml  # Re-sync labels (GitHub only)
 # Cancel a running sweep: mcp__loom__cancel_sweep --sweep_id <id>
 ```
 
+## Custom Guard Hooks
+
+Loom ships Bash `PreToolUse` guard hooks (`defaults/hooks/guard-destructive.sh`) that block or ask on destructive commands. Several category toggles let a repo opt out of checks that are a category error for it — see `defaults/CLAUDE.md` → "Custom Guard Hooks" for the full catalog (`guards.sqlDdl`, `guards.cloudCli`, `guards.rmScope`, `guards.forceScope`). The read-only fast-path toggle is documented below.
+
+### Read-Only Fast-Path Guard Toggle (`guards.readOnlyFastPath` / `LOOM_GUARD_READONLY_FASTPATH`)
+
+`guard-destructive.sh` fires before **every** Bash tool call. In Bash-dense sessions almost every call is obviously read-only (`git status`, `ls`, `grep`, `aws … describe*`, `gh … list`), yet each otherwise runs the full deny/ask gauntlet (~37 `grep`/`awk`/`sed` forks plus a `git rev-parse`, ~179ms) before allowing. The read-only fast path (issue #3687) short-circuits that case to a **silent** `allow` (exit 0, zero output, no logging) via one bash-builtin structural test (zero forks) plus one lazy `jq` config read, running before the repo-root `git rev-parse` and every deny/ask array.
+
+The fast path is **on by default**, resolved highest-precedence first:
+
+1. **`LOOM_GUARD_READONLY_FASTPATH` env var** — `0`/`false`/`no` disables (full-path checking restored byte-for-byte); `1`/`true`/`yes` forces on. Overrides config.
+2. **`.loom/config.json` → `guards.readOnlyFastPath`** — default `true` when absent; set `false` to disable.
+3. **Default** — `true`.
+
+**Security**: the fast path is a guard bypass by construction, so admission is purely structural. A command is fast-pathed only when it contains **none** of `;` `&` `|` `<` `>` backtick `$(` newline (excludes chaining/piping/redirection/substitution) **and** its first token exactly matches the built-in allowlist: `git status|log|diff|show` (bare, no `git -C …`), `ls`, `grep`, `rg`, `gh <noun> view|list`, `aws <service> describe*|get*|list*`, `aws s3 ls`. Wrappers (`bash -c`, `eval`, `sudo …`, `env …`) are excluded automatically because their first token isn't allowlisted. So `git status && git push --force origin main` takes the full path and is still denied.
+
+**`cat` and `ssh` are deliberately excluded**: `cat` has an existing `.ssh`/`.aws/credentials` ASK carve-out a blanket fast-path would skip, and `ssh` wraps an opaque remote payload the catastrophic scan still covers.
+
+**Optional** `guards.readOnlyFastPathExtra` is an extend-only array of **literal first-word commands** added to the allowlist without hand-editing the installer-managed `.claude/settings.json`:
+
+```json
+{ "guards": { "readOnlyFastPath": true, "readOnlyFastPathExtra": ["jq", "wc"] } }
+```
+
+> **Warning**: each entry is a full-generality bypass for that command word (all arguments) — only add bare, argument-independent read-only utilities, never scripts or anything that could wrap a mutating call.
+
+Disabling the fast path never weakens any deny/ask rule; a missing/malformed `.loom/config.json` falls through to fast-path-ON.
+
 ## MCP Hooks
 
 Loom provides a unified MCP server (`mcp-loom`) for programmatic control. See the mcp-loom package README for full tool documentation.
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -1110,6 +1110,76 @@ LOOM_FORCE_SCOPE=protected git push --force origin feature/x   # allowed
 LOOM_FORCE_SCOPE=all git reset --hard HEAD~1   # ASK
 ```
 
+### Read-Only Fast-Path Guard Toggle (`guards.readOnlyFastPath` / `LOOM_GUARD_READONLY_FASTPATH`)
+
+`guard-destructive.sh` is a `PreToolUse`/`Bash` hook, so it fires before **every** Bash tool call. In Bash-dense sessions (remote ops, benchmark drivers) nearly every call is obviously read-only — `git status`, `ls`, `grep`, `aws … describe*`, `gh … list` — yet each one otherwise runs the full deny/ask gauntlet (~37 `grep`/`awk`/`sed` forks plus a `git rev-parse`, ~179ms measured) before falling through to `allow`.
+
+The read-only fast path (issue #3687) short-circuits that overwhelmingly-common case to a **silent** `allow` (exit 0, zero stdout/stderr, no logging) using a single bash-builtin structural test — zero forks — plus, only when that test passes, one lazy `jq` config read. It runs first, before the `git rev-parse` repo-root resolution and before any deny/ask array.
+
+The fast path is **on by default**. It is resolved in this order (highest precedence first):
+
+1. **`LOOM_GUARD_READONLY_FASTPATH` env var** — `0`/`false`/`no` disables the fast path (every command takes the full deny/ask path, byte-for-byte as before); `1`/`true`/`yes` forces it on. Overrides the config value.
+2. **`.loom/config.json`** — `guards.readOnlyFastPath` (default `true` when absent). Set it to `false` to disable:
+   ```json
+   {
+     "guards": {
+       "readOnlyFastPath": false
+     }
+   }
+   ```
+3. **Default** — `true` (fast path active).
+
+**Security — the fast path is a guard bypass by construction**, so admission is purely **structural** and conservative, never content-sensitive. A command is fast-pathed only when **all** of these hold (otherwise it falls through to the full path unchanged):
+
+- The raw command contains **none** of `;` `&` `|` `<` `>` backtick `$(` or a newline — this excludes all chaining, piping, redirection, and command substitution. So `git status && git push --force origin main`, `git status; rm -rf /`, and `git status $(rm -rf /)` all take the full path and are still denied.
+- The **first token** is an exact allowlist match (never a wrapper — `bash -c`, `sh -c`, `eval`, `xargs`, `env … git status`, `sudo git status` are all excluded because their first token isn't allowlisted):
+
+| First token | Admitted form |
+|-------------|---------------|
+| `git` | `git status` / `git log` / `git diff` / `git show` — **bare** subcommand only (so `git -C /path status` is not admitted) |
+| `ls`, `grep`, `rg` | any arguments |
+| `gh` | `gh <noun> view` / `gh <noun> list` (never `delete`/`close`/`archive`/…) |
+| `aws` | `aws <service> describe*` / `get*` / `list*`, and `aws s3 ls` |
+
+**`cat` and `ssh` are deliberately EXCLUDED** from the built-in list, even though they are read-only in spirit:
+
+- `cat` has a narrow existing `ASK` carve-out (`cat …/.ssh/…`, `cat …/.aws/credentials`); a blanket `cat` fast-path would silently skip it.
+- `ssh <host> '<cmd>'` wraps an **opaque remote command string** that the raw `ALWAYS_BLOCK` catastrophic scan still covers today; fast-pathing any `ssh …` would drop that coverage.
+
+**Optional extend-only escape hatch** — `guards.readOnlyFastPathExtra` is an array of **literal first-word commands** to add to the built-in list without hand-editing the Loom-managed `.claude/settings.json` (which the installer may overwrite). This directly answers "give operators a supported way to scope the matcher":
+
+```json
+{
+  "guards": {
+    "readOnlyFastPath": true,
+    "readOnlyFastPathExtra": ["jq", "wc"]
+  }
+}
+```
+
+> **Warning**: each word added here is a **guard bypass for that command word in full generality** (all arguments). Only add bare, argument-independent read-only utilities — never your own scripts or anything that could wrap a mutating call. Entries are matched as the literal first token only; no subcommand/verb parsing is applied to custom entries.
+
+The config read is best-effort and lazy: it happens only after a command has already passed the structural test, and any missing/empty/malformed `.loom/config.json` falls through to fast-path-ON. Disabling the fast path never weakens any deny/ask rule — it only makes the guard do its full work on every command again.
+
+**Examples**:
+
+```bash
+# Default: read-only commands are near-free and silent
+git status                     # fast-pathed (silent allow)
+aws ec2 describe-instances     # fast-pathed
+gh pr list                     # fast-pathed
+git status && git push --force origin main   # NOT fast-pathed → full path → DENIED
+
+# Disable the fast path for one command (restore full-path checking)
+LOOM_GUARD_READONLY_FASTPATH=0 git status
+
+# Persist the opt-out for a whole repo
+#   .loom/config.json  ->  { "guards": { "readOnlyFastPath": false } }
+
+# Extend the allowlist with a bare read-only utility
+#   .loom/config.json  ->  { "guards": { "readOnlyFastPathExtra": ["jq"] } }
+```
+
 ### Protecting Read-Only Directories
 
 Many projects have directories that should never be modified by agents (vendor code, generated files, external SDKs, process design kits). Loom provides a template hook for this.
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -65,6 +65,198 @@ if [[ -z "$COMMAND" ]]; then
     exit 0
 fi
 
+# =============================================================================
+# READ-ONLY FAST PATH (issue #3687) — default ON.
+#
+# guard-destructive.sh is a PreToolUse/Bash hook, so it fires before EVERY Bash
+# tool call. In Bash-dense sessions (remote ops, benchmark drivers) the vast
+# majority of those calls are obviously read-only — `git status`, `ls`, `grep`,
+# `aws … describe*`, `gh … list` — yet each one still runs the full deny/ask
+# gauntlet (~37 grep/awk/sed forks + a git rev-parse, ~179ms measured). This
+# block short-circuits that overwhelmingly-common case to a silent `allow` with
+# a single bash-builtin structural test (zero forks) plus, only when that test
+# passes, one lazy `jq` config read.
+#
+# SECURITY: a fast path is a guard bypass by construction, so admission is
+# purely STRUCTURAL and conservative — never content-sensitive:
+#   1. Reject fast-path eligibility if the raw command contains ANY of
+#      ;  &  |  <  >  backtick  $(  or a newline. This kills chaining, piping,
+#      redirection, and command substitution (so `git status && <force-push>`,
+#      `git status; rm -rf /`, `git status $(rm -rf /)`, `git status > /etc/x`
+#      all fall through to the full path unchanged).
+#   2. Exact first-token (command-word) allowlist — never a wrapper. Because the
+#      allowlist is keyed on the literal first token, wrapper forms (`bash -c`,
+#      `sh -c`, `eval`, `xargs`, `env … git status`, `sudo git status`) are
+#      excluded automatically: their first token isn't allowlisted.
+#   3. Verb/subcommand exactness for multi-word tools, chosen to be provably
+#      disjoint from every existing deny/ask pattern:
+#        git status|log|diff|show  (bare — `git -C /p status` is NOT admitted)
+#        ls  grep  rg
+#        gh <noun> view|list       (never delete/close/archive/…)
+#        aws <service> describe*|get*|list*  and  aws s3 ls   (mirrors the
+#          verb-anchoring already in CLOUD_ASK_PATTERNS: those verbs are never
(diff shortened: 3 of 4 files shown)


rjwalters:
LGTM — APPROVED. Adversarial review of the read-only fast path (#3687).

I attacked the admission logic as a guard bypass by construction and could not break the default/built-in configuration. Ran the full suite both ways from the worktree: **335/335 with the fast path ON, 333/333 with `LOOM_GUARD_READONLY_FASTPATH=0`** (byte-for-byte restoration; the 2 extra ON cases are the new observable admissions).

**Attack results (empirical probes against the hook):**
- Chaining/redirection/substitution all fall through to the full path and still deny: `git status && <force-push-to-main>`, `git status ; <catastrophic-rm>`, and `git status $(...)` substitution all deny (structural test rejects the metachar set: semicolon, ampersand, pipe, angle brackets, backtick, dollar-paren, newline).
- Env-var prefix `FOO=bar <force-push-to-main>` -> **deny** (first token `FOO=bar` isn't allowlisted). Wrappers (`bash -c`, `sudo`, `env`) likewise decline.
- Non-bare `git -C /tmp status` correctly NOT fast-pathed (t[1] must be exactly status/log/diff/show); still allowed via full path.
- Array admission indices are correct (git n>=2 checks t[1], gh n>=3 checks t[2], aws n>=3 checks t[2] plus the s3-ls special-case) — no off-by-one.
- A real force-push, a repo deletion, an IAM user deletion, an S3 bucket removal, an EC2 terminate, and a hard reset all keep their deny/ask verdicts.
- **env>config precedence** verified: env=1 forces the fast path on but only *enables* it (a force-push still denies since it isn't allowlisted); env=0 disables entirely; config `readOnlyFastPath:false` disables; env overrides config both directions.
- **Silence guarantee** verified: admission is exit 0 with zero stdout/stderr and no log write, before REPO_ROOT resolution (assert_allow_silent, 0 bytes).
- **CWD-walk config read**: fork-free lexical parent walk, absolute-path guarded, stops at root, no symlink resolution (no loop risk), cached; malformed/missing config -> fast-path-ON default. Sound.

The verdict changes the fast path introduces (e.g. a read-only `grep` of DDL text going DENY->ALLOW) are all false-positive fixes on genuinely read-only commands — grep/ls/git-read/gh-view/aws-read never execute their arguments, so scary substrings are inert. No command the full path would deny/ask is silently admitted by the built-in allowlist.

**Non-blocking follow-up notes (no regression — the full path already allows these, so posture is unchanged):**
1. `rg --pre`/`--pre-glob` is a preprocessor code-exec vector, and `git diff --output=<file>` / `aws s3api get-object <localpath>` are write primitives. All three are already allowed by the full path today, so fast-pathing them changes nothing — but a future follow-up could tighten the allowlist to exclude `rg --pre`.
2. `readOnlyFastPathExtra` matches the literal first token in full generality — I confirmed that adding `git` to the extras list silently allows a force-push-to-main. This is the documented, loudly-warned contract ("full-generality bypass for that command word… never anything that could wrap a mutating call") and requires deliberate operator opt-in, so it's acceptable. A future hardening could reject obviously-dangerous words (git/bash/sh/eval/sudo/env/xargs) from the extras list.

CI is green (Installer Integration Tests pass; Rust/MCP jobs correctly skipped) and merge state is CLEAN/MERGEABLE. Excellent, conservative, thoroughly-tested change with strong doc coverage.

PR #3694, opened 2026-07-21
feat(sweep): core-scale subagent-path auto wave size within [3,6] (#3693)

### LINKED ISSUE #3693: /sweep: raise the default in-session (subagent-path) parallelism above 3
## Request
Make `/loom:sweep`'s **default** in-session parallelism more aggressive. Today the subagent-path auto wave size caps at **3** (`Resolve auto wave size`, target 3 for `use_subagent`), citing the #3289 nested-dispatch stall. In practice, when the orchestrator dispatches builders/judges **directly** (one level deep, which `/sweep` already does), waves of **5–6 run cleanly** and cut wall-clock substantially.

## Evidence (this session)
Running `/loom:sweep all` on rjwalters/kicad-tools, single-account subagent path on a 28-core host:
- Repeated waves of **3** builders + 3 judges completed reliably (many issues merged).
- After widening to a **5-wide** wave (5 builders dispatched in one tool-call block, then 5 judges), all 5 completed near-simultaneously with no stall — the #3289 hazard did not manifest because the dispatch is one level deep (parent → builder), never the nested `parent → /shepherd → builder` grandchild pattern #3289 is actually about.
- The concurrency ceiling the harness enforces is already `min(16, cores-2)` = 16 here, so 3 leaves most of the machine idle.

## The mismatch
The **cap of 3 conflates two different things**:
- The #3289 stall is specifically **nested** parallel dispatch (stream-pump dies on parallel *grandchildren*). `/sweep` explicitly never does that ("CRITICAL: One level deep — never spawn /shepherd as a subagent").
- **Width** at one level deep is bounded by the harness concurrency cap (`min(16, cores-2)`), not by #3289.

So the safe default for the direct-dispatch orchestrator path could scale with cores rather than sitting at a fixed 3.

## Proposed change (pick one or combine)
1. **Raise the subagent-path auto target** from 3 to something like `min(6, cores-2)` (or `min(8, ...)`), keeping the disk-headroom clamp and the candidate-count clamp. Keep an explicit `--builders-per-wave` override.
2. **Scale by cores**: auto target = `clamp(floor((cores-2)/4), 3, 8)` or similar, so big hosts get more and small hosts stay conservative.
3. At minimum, **document** that operators can safely raise `--builders-per-wave` to 5–6 on multi-core hosts (the current guidance warns at N≥4 as "unvalidated for the subagent path", which this session's evidence contradicts for one-level-deep dispatch).

## Non-goal / caveat
The 10-wide **daemon detached-process** path (`mcp__loom__dispatch_sweep`) remains the right lever for true high parallelism + token rotation, but it requires a running daemon + a multi-account token pool that many single-operator setups don't have. This request is specifically about a smarter **default for the always-available subagent path**.

## Additional data point (rate limits)
On a single account, a long high-parallelism sweep can hit the weekly rate limit. The daemon+multi-account path is the real fix for that, but a modestly higher subagent default (5–6) materially improves throughput before that ceiling without changing the safety profile.

_Filed at operator request after a large multi-wave `/loom:sweep all` session (2026-07-21) that demonstrated 5-wide waves running cleanly one level deep._

---

### Investigation summary

The subagent-path auto wave size is currently a **fixed target of 3**, defined in `defaults/scripts/lib/disk-headroom.sh`:

- `LOOM_SUBAGENT_WAVE_CAP` (default `3`) is read in `loom_wave_size_from_disk()` (`defaults/scripts/lib/disk-headroom.sh` lines ~99-101, ~112-117) and used as the `target` for `mechanism=subagent`. `LOOM_DAEMON_WAVE_TARGET` (default `10`) is the daemon-path target — unaffected by this issue.
- The doc comment header of that file (lines ~28-38) explicitly says: *"NEVER raise this cap toward 10 — route high parallelism through the daemon path instead."* This request does not propose raising it toward 10; it proposes a modest core-scaled bump within [3, 6].
- `defaults/scripts/tests/test-disk-headroom.sh` Test 2 (lines 66-80) currently hard-asserts `wave subagent 20 100` → `3|target`. This assertion will need to become core-count-aware (see Test plan below) once the target is no longer a flat constant.
- `defaults/.claude/commands/loom/sweep.md` references the cap-of-3 / #3289 distinction in many places, all of which need to move in lockstep with any code change:
  - Flag docs / validation warning: line 209 (`WARNING: --builders-per-wave=<N> ... N>=4 may exhaust context or hit rate limits`)
  - Wave-size guidance table: lines 216-226 (rows for `auto`, `2`, `3`, `>=4`, and the explanatory paragraph)
  - "Resolve auto wave size" section: lines 551-592, including the `MECH=subagent` branch comment (line 571), the `loom_wave_size_from_disk` prose recap (line 579, currently says *"The target is 10 for the daemon path and 3 for the subagent path"*), and the reason-token table rows (lines 586-590, currently hard-coding `cap 3` in the log strings)
  - Wave Lifecycle dispatch: line 1063 (*"the auto size here is capped at 3 (#3289-safe)"*)
  - Limitations table: line 1292 (*"capped at 3 on the in-session subagent path... The N=3 soft cap is subagent-path-specific"*)
  - Also referenced (context only, no numeric claim to update): lines 119, 220, 223, 361, 1276, 1306, 1429 (`CRITICAL: One level deep` section and #3289 narrative — these describe the *nested-dispatch* hazard, not width, and should NOT change)
- `CLAUDE.md` (repo root) and `defaults/CLAUDE.md` (installer template) both carry a one-line summary of the auto-wave-size default ("daemon path targets up to 10... subagent path stays at the #3289-safe cap of 3") that will also need the number updated if the target changes.
- No duplicate issue found (`.loom/scripts/check-duplicate.sh 3693` → "No duplicates found").

### Weighing the evidence

- One clean 5-wide one-level-deep session on a 28-core host is real evidence that **width** (not nesting) is what #3289 gates, and the issue's own re-reading of #3289 ("stream-pump dies on parallel *grandchildren*, not parallel *children*") is architecturally sound — `/sweep`'s wave loop already dispatches builders/judges directly, never `/shepherd`-as-subagent, so the nested-dispatch hazard genuinely does not apply to width.
- However, a single successful session is not a validated ceiling the way `3` was (`3` is called "Tested and validated" in the guidance table after presumably more repeated runs). Two additional factors argue for a **conservative** middle path rather than jumping straight to 6-8:
  1. **Rate-limit burn**: the issue itself notes single-account weekly limits — a wider wave burns tokens/requests faster per unit wall-clock, which matters most for the single-token operators this path exists for (multi-account operators should be on the daemon path already).
  2. **Orchestrator context-window pressure**: each wave member's Task result must be read back into the *same* orchestrator session context. A 5-6-wide wave already roughly doubles the per-wave return payload versus 3; scaling further (toward 8) risks context exhaustion in the orchestrator itself, a failure mode `#3289` doesn't even describe.
- Recommendation: **raise the default, but core-scale it and cap the auto ceiling at 6, not 8.** This keeps small/shared hosts conservative (2-10 core dev boxes still get 3), lets big hosts benefit (28-core → 6), and keeps a comfortable margin below the point where rate-limit burn and orchestrator context pressure become the binding constraint. This is proposal (2) from the issue, tightened to `clamp(floor((cores-2)/4), 3, 6)` instead of `[3, 8]`.

### Recommended design

**1. New arithmetic in `defaults/scripts/lib/disk-headroom.sh` (unit-tested, not prose):**

Add two new functions, keeping `loom_wave_size_from_disk()`'s signature and existing tests unchanged:

```bash
# loom_detect_cores
# Portable core-count detection: Linux `nproc`, macOS/BSD `sysctl -n hw.ncpu`,
# falling back to a conservative default (4) if neither tool is present or
# both fail. Honors LOOM_CORES_OVERRIDE for deterministic testing/tuning.
# I/O function (shells out) — NOT pure, kept separate from the arithmetic below.
loom_detect_cores() {
    if [[ -n "${LOOM_CORES_OVERRIDE:-}" ]]; then
        echo "$LOOM_CORES_OVERRIDE"
        return 0
    fi
    if command -v nproc >/dev/null 2>&1; then
        nproc
        return 0
    fi
    if command -v sysctl >/dev/null 2>&1; then
        local n
        n="$(sysctl -n hw.ncpu 2>/dev/null)"
        if [[ "$n" =~ ^[0-9]+$ && "$n" -ge 1 ]]; then
            echo "$n"
            return 0
        fi
    fi
    echo 4   # conservative fallback when core count can't be detected
}

# loom_subagent_target_from_cores <cores>
# Pure integer arithmetic: clamp(floor((cores-2)/4), 3, 6).
# Small/shared hosts stay at the validated floor of 3; big hosts scale up to
# a ceiling of 6 (NOT 8/10 — see #3693 rationale: rate-limit burn on a single
# account and orchestrator context-window pressure from reading back more
# parallel Task results both argue for a conservative ceiling below the
# daemon-path target).
loom_subagent_target_from_cores() {
    local cores="$1"
    if [[ ! "$cores" =~ ^[0-9]+$ ]]; then
        echo "loom_subagent_target_from_cores: non-integer cores '$cores'" >&2
        return 2
    fi
    local raw=$(( (cores - 2) / 4 ))   # bash integer division floors toward zero for non-negatives
    if (( cores < 2 )); then raw=0; fi
    local k=$raw
    (( k < 3 )) && k=3
    (( k > 6 )) && k=6
    echo "$k"
}
```

**2. Wire the new default into the existing `LOOM_SUBAGENT_WAVE_CAP` env var** (no change to `loom_wave_size_from_disk()`'s internals or signature — it keeps reading `LOOM_SUBAGENT_WAVE_CAP` verbatim, preserving all 4 subagent-path unit tests as regression guards for an *explicit* cap):

In the sweep.md "Resolve auto wave size" section (and any operator script that sources `disk-headroom.sh`), resolve the env var *before* calling `loom_wave_size_from_disk`:

```bash
# Core-scaled default (#3693); an operator-set LOOM_SUBAGENT_WAVE_CAP always wins.
: "${LOOM_SUBAGENT_WAVE_CAP:=$(loom_subagent_target_from_cores "$(loom_detect_cores)")}"
```

This preserves the override contract (`LOOM_SUBAGENT_WAVE_CAP` explicit env value always wins) while making the *default* core-scaled instead of flat-3.

**3. `sweep.md` prose updates (must move in lockstep with the code):**

- Wave-size guidance table (lines 216-226): change the `auto` row to describe the core-scaled range (`3-6`, formula, and "clamped by candidate count and disk"); change `3` row from "Tested and validated. The #3289-safe ceiling" to "Tested and validated. The #3289-safe **floor** — the default auto-resolved target now scales with cores up to 6"; move the `>= 4` warning threshold up so `4`, `5`, and `6` are not flagged as "unvalidated" (see next bullet).
- `--builders-per-wave` validation warning (line 209): change `If N > 3` to `If N > 6`, and reword: `WARNING: --builders-per-wave=<N> is unvalidated. N<=6 is recommended; N>=7 may exhaust context or hit rate limits.` (Rationale: 3-6 is now the auto-resolved and evidence-supported range; the warning threshold should track it.)
- "Resolve auto wave size" section (lines 551-592): update the subagent branch comment (line 571) and the prose recap (line 579) to describe the core-scaled target instead of a flat 3; update the reason-token table rows (586-587) to reflect a variable target (e.g. `wave size K, mechanism=subagent: N cores → subagent path (core-scaled target K, floor 3, ceiling 6)`).
- Wave Lifecycle dispatch note (line 1063): "capped at 3 (#3289-safe)" → "core-scaled (#3289-safe floor 3, ceiling 6, #3693)".
- Limitations table (line 1292): update "capped at 3 on the in-session subagent path" language similarly.
- Do **NOT** touch the "CRITICAL: One level deep" section (lines 364-372) or the #3289 narrative (lines 361, 1276, 1306, 1429) — those describe the nested-dispatch hazard, which this issue correctly does not change.
- Update the one-line summaries in `CLAUDE.md` and `defaults/CLAUDE.md` ("subagent path stays at the #3289-safe cap of 3") to match.

### Affected Files

- `defaults/scripts/lib/disk-headroom.sh` — add `loom_detect_cores` and `loom_subagent_target_from_cores`; update header doc comment.
- `defaults/scripts/tests/test-disk-headroom.sh` — new unit tests for both functions (see Test plan).
- `defaults/.claude/commands/loom/sweep.md` — lines 209, 216-226, 551-592, 1063, 1292 (prose sync).
- `CLAUDE.md` — one-line auto-wave-size summary under "Single-issue lifecycle".
- `defaults/CLAUDE.md` — matching one-line summary (installer template).

### Acceptance criteria

1. `loom_subagent_target_from_cores <cores>` is a pure function in `disk-headroom.sh`: given `cores`, returns `clamp(floor((cores-2)/4), 3, 6)`, rejects non-integer input with a clear stderr message and exit 2.
2. `loom_detect_cores` returns a positive integer via `nproc` (Linux), `sysctl -n hw.ncpu` (macOS/BSD), falls back to `4` if neither is available, and honors `LOOM_CORES_OVERRIDE` for deterministic testing.
3. The default resolution of `LOOM_SUBAGENT_WAVE_CAP` becomes core-scaled (via the two functions above) wherever it's currently hardcoded to `3`, while an operator-supplied `LOOM_SUBAGENT_WAVE_CAP` env var continues to override it exactly as before — **no change** to `loom_wave_size_from_disk()`'s function signature or its existing 4 subagent-path unit tests (they pin explicit-cap behavior, which is unchanged).
4. `sweep.md`'s wave-size guidance table, `--builders-per-wave` validation warning (threshold moves from `N>=4` to `N>=7`), "Resolve auto wave size" section, Wave Lifecycle note, and Limitations table all describe the core-scaled `[3,6]` default consistently — no stale "capped at 3" or "N>=4" language remains for the subagent path.
5. The "CRITICAL: One level deep" section and all #3289 nested-dispatch prose are untouched — this issue changes width defaults only, never the nesting rule.
6. `CLAUDE.md` and `defaults/CLAUDE.md` one-line summaries match the new default.

### Test plan

- **Unit tests** (`defaults/scripts/tests/test-disk-headroom.sh`, extend with new cases, keep all existing 4 subagent-path cases passing unchanged since they pass an explicit env-less `wave subagent ...` call — update them to set `LOOM_SUBAGENT_WAVE_CAP` explicitly, or add a preceding step in the test that resolves it via `loom_subagent_target_from_cores` with a fixed `LOOM_CORES_OVERRIDE`, so the test suite stays deterministic regardless of the host's actual core count):
  - `loom_subagent_target_from_cores`: table-test at least `cores=1` → 3 (raw would be negative/0, floors to 3), `cores=4` → 3, `cores=16` → 3, `cores=20` → 4, `cores=24` → 5, `cores=28` → 6, `cores=100` → 6 (ceiling holds).
  - `loom_subagent_target_from_cores` rejects non-integer input (e.g. `"abc"`) with non-zero exit.
  - `loom_detect_cores` with `LOOM_CORES_OVERRIDE=7` returns `7` (deterministic path, no host dependency needed for CI).
  - `loom_detect_cores` with no override returns a positive integer (loose host-dependent smoke check only).
  - Re-run the existing daemon-path tests (Test 1) unchanged — this issue must not touch `LOOM_DAEMON_WAVE_TARGET`.
- **Prose consistency check**: `grep -n "cap 3\|capped at 3\|N > 3\|N>=4\|N<=3" defaults/.claude/commands/loom/sweep.md CLAUDE.md defaults/CLAUDE.md` returns no unintended hits after the edit (any hits should be reviewed to confirm they're intentionally-preserved #3289 nesting prose, not stale width numbers).
- **Manual smoke**: run `bash defaults/scripts/tests/test-disk-headroom.sh` and confirm all tests pass; run `LOOM_CORES_OVERRIDE=28 bash -c 'source defaults/scripts/lib/disk-headroom.sh; loom_subagent_target_from_cores "$(loom_detect_cores)"'` and confirm it prints `6`.

rjwalters:
## Summary

Makes `/loom:sweep`'s in-session (subagent-path) auto wave size **core-scaled within `[3, 6]`** instead of a flat cap of 3. The old flat-3 default conflated two different things: the #3289 stall is a *nested*-dispatch hazard (parallel grandchildren, `parent → /shepherd → builder`), which `/sweep` never does — it dispatches builders/judges directly, one level deep. **Width** at one level deep is bounded by the harness concurrency cap `min(16, cores-2)`, not by #3289, so small hosts can stay conservative while big hosts scale up.

## Changes

- **`defaults/scripts/lib/disk-headroom.sh`**: two new functions
  - `loom_detect_cores` — portable core detection (`nproc` → `sysctl -n hw.ncpu` → fallback `4`), honoring `LOOM_CORES_OVERRIDE` for deterministic tests.
  - `loom_subagent_target_from_cores <cores>` — pure `clamp(floor((cores-2)/4), 3, 6)`, rejects non-integer input with exit 2.
  - Header doc + `loom_wave_size_from_disk` inline doc updated. **`loom_wave_size_from_disk`'s signature and its 4 subagent-path unit tests are unchanged** — the core-scaling is resolved into `LOOM_SUBAGENT_WAVE_CAP` at the caller, and an operator-set env value still always wins.
- **`defaults/.claude/commands/loom/sweep.md`**: guidance table (`auto` row, `3` = floor not ceiling, new `4`–`6` row), `--builders-per-wave` warning threshold `N>=4` → `N>=7`, "Resolve auto wave size" section (branch comment resolves `LOOM_SUBAGENT_WAVE_CAP := loom_subagent_target_from_cores $(loom_detect_cores)`, prose recap, reason-token rows), Wave Lifecycle note, Limitations row.
- **`defaults/CLAUDE.md`**: one-line wave-parallelism summary. (Root `CLAUDE.md` has no such summary line — nothing to change there.)
- **`defaults/scripts/tests/test-disk-headroom.sh`**: Test 2 made deterministic via explicit `LOOM_SUBAGENT_WAVE_CAP`; new Test 8 (core-scaling clamp: 1/4/14/16→3, 20→4, 22/24→5, 28→6, 64/100→6, non-integer reject) and Test 9 (`loom_detect_cores` override + composition).

The #3289 "CRITICAL: One level deep" section and all nested-dispatch narrative are **untouched** — this changes width defaults only.

## Ceiling rationale (6, not 8/10)

Single-account weekly rate-limit burn and orchestrator context-window pressure (every wave member's Task result is read back into the same session) both bind before #3289 does. True high parallelism toward 10 remains the daemon detached-process path's job.

## Test results

- `test-disk-headroom.sh`: 33 passed, 0 failed
- `test-sweep-pr-mode.sh`: 58 passed, 0 failed
- `test-sweep-checkpoint.sh`: 57 passed, 0 failed
- `shellcheck` clean on both modified shell files
- Prose consistency grep: no stale `cap 3` / `N>=4` width references remain

Closes #3693


rjwalters:
LGTM — approving.

**Verified independently:**
- **Clamp arithmetic**: sourced `loom_subagent_target_from_cores` and swept boundaries — 4→3, 14→3, 16→3, 18→4, 20→4, 22→5, 24→5, 26→6, 28→6, 64/100→6, cores<2→3. Matches `clamp(floor((cores-2)/4), 3, 6)` exactly; integer division floors correctly at every edge.
- **Env precedence**: caller `: "${LOOM_SUBAGENT_WAVE_CAP:=...}"` fills only unset/empty then exports; `loom_wave_size_from_disk` reads it verbatim (fallback 3 = band floor). Operator-set `LOOM_SUBAGENT_WAVE_CAP` and explicit `--builders-per-wave` both win. Confirmed by the new Test 2 case (`LOOM_SUBAGENT_WAVE_CAP=6` → 6).
- **Portability**: `loom_detect_cores` = nproc → sysctl hw.ncpu → fallback 4, with `LOOM_CORES_OVERRIDE` short-circuit and integer/≥1 validation at each tier. Clean separation of I/O (detect) from pure math (target) keeps the arithmetic unit-testable.
- **Policy**: ceiling 6 (not the issue's 8) is a conservative width call — rate-limit burn on a single account + orchestrator context-window pressure both bind before #3289. Rationale is documented in the lib header, function docstring, sweep.md, and CLAUDE.md.
- **Prose consistency**: no stale `cap 3`/`N>=4` width remnants; daemon target 10 unchanged; the #3289 one-level-deep **nesting** rule is explicitly untouched and cleanly distinguished from the **width** knob this PR scales.

**Suites** (from worktree): test-disk-headroom 33/33, test-sweep-pr-mode 58/58, test-sweep-checkpoint 57/57. shellcheck clean. CI green, mergeStateStatus CLEAN.

PR #3696, opened 2026-07-21
feat(tokens): home-dir master account pool with per-repo override (#3695)

### LINKED ISSUE #3695: token pool: support a home-dir master account list with per-repo override/additions (avoid duplicating ACCOUNT_* into every repo)
## Request
Support a **home-directory master account list** for the OAuth token pool, with **per-repo local override / additions**, so a set of Claude accounts is configured **once** and shared across every workspace — instead of re-declaring `ACCOUNT_*_N` triples in every repo's `.env`.

## Current behavior (verified in loom v0.11.0)
The token pool is **strictly repo-local**:
- `loom-tokens bootstrap` reads `<repo-root>/.env` by default (`--env <path>` overrides the source per-invocation, but it still materializes into the *current* repo).
- `agent_spawn.py::_select_oauth_token(repo_root)` resolves `tokens_dir = repo_root / ".loom" / "tokens"`.
- The daemon (`loom-daemon/src/sweep_registry.rs`) resolves `tokens_dir="$ws/.loom/tokens"` from the workspace.

So there is **no** `~/.loom/tokens/` global pool, no XDG/home master `.env`, and no merge/override semantics. An operator with N repos must put the same account triples in N `.env` files and run `bootstrap` in each — every repo holds its own copy of the secrets, and adding/rotating an account means editing every repo.

Note: `~/.loom/` is *already* a home-dir Loom location (the daemon socket lives at `~/.loom/loom-daemon.sock`), so a home-dir pool at `~/.loom/tokens/` (materialized from a home-dir master `.env`) would be a natural extension of the existing convention.

## Proposed behavior
1. A **home-dir master account source** — e.g. `~/.loom/.env` (or `~/.config/loom/accounts.env`, XDG-aware) declaring `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` / `ACCOUNT_TOKEN_FILE_N` triples once.
2. `loom-tokens bootstrap` (and the spawn/daemon selector) **merge** the home-dir master with any repo-local `.env` / `.loom/tokens/`, where the **repo-local entries override or add to** the master set (so a repo can pin a subset, add a repo-specific account, or exclude one — cf. the existing `.allowlist` pin mechanism, which could be the per-repo override surface).
3. The selector's resolution order becomes: repo-local `.loom/tokens/` (if present) → else materialize-on-demand from `repo .env` merged over `~/.loom master`. Precedence explicit and documented, mirroring the existing `.env`-vs-`.loom/tokens` precedence logic (`verify-token-precedence.sh`).

## Why it matters
The multi-account pool is the lever for daemon-path parallelism + weekly-quota rotation (Stage -1 `PROBE_POOL`). Requiring the full account set duplicated into every repo's `.env` is friction that discourages configuring rotation at all — a single operator working across several repos (e.g. kicad-tools + its consumer boards) wants to declare accounts once and have every workspace's daemon dispatch rotate through them.

- [ ] A home-dir master account source (`~/.loom/.env` or XDG equivalent) is read by `loom-tokens bootstrap` and by the spawn/daemon selector.
- [ ] Repo-local `.env` / allowlist entries **override or add to** the master set (documented precedence), not silently ignored.
- [ ] `loom-tokens check`/`bootstrap --dry-run` reports the *effective* merged account set and where each came from (home vs repo).
- [ ] Secrets are not duplicated into every repo (a repo can rely on the home master with an empty local `.env`).
- [ ] Backward-compatible: repos with only a local `.env` (no home master) behave exactly as today.

_Filed at operator request during a large multi-wave /loom:sweep session (2026-07-21). Verified repo-local-only pool resolution in loom v0.11.0 (agent_spawn.py, bootstrap.py, sweep_registry.rs)._

rjwalters:
## Summary

Implements #3695 — a **home-dir master account list** for the OAuth token pool with **per-repo override/additions**, so a set of Claude accounts is configured **once** and shared across every workspace instead of re-declaring `ACCOUNT_*_N` triples in every repo's `.env`.

## Design

The merge happens **at `loom-tokens bootstrap` time** and materializes into `.loom/tokens/`. Since `select`/`check`/`pin` and the Rust daemon (via `spawn-claude.sh`) all read only `.loom/tokens/`, they need **no changes** — the surface stays small and low-risk.

**Sources & precedence:**

| Source | Default | Override |
|--------|---------|----------|
| Home master | `~/.loom/accounts.env` | `LOOM_ACCOUNTS_ENV` env (`""` disables); `--home-env` / `--no-home` |
| Repo-local | `<repo>/.loom/accounts.env` if present, else legacy `<repo>/.env` | `--env` |

**Merge is keyed by account email** (`ACCOUNT_EMAIL`), repo wins:
- email only in master → inherited
- email only in repo → added
- email in both → repo overrides (rotate key / repoint token file)

Exclusion stays with the existing `loom-tokens pin` allowlist (the merge only adds/overrides, never subtracts), as the issue suggested.

## Acceptance criteria

- [x] Home master read by `bootstrap` (and, transitively via the materialized pool, by the spawn/daemon selector)
- [x] Repo-local entries override or add to the master (documented precedence)
- [x] `bootstrap` / `--dry-run` report the **effective merged set** and each account's provenance (home vs repo); `index.json` records `source`
- [x] Secrets not duplicated — a repo can rely on the master with **no** local account file
- [x] Backward-compatible — a repo with only a legacy `.env` behaves exactly as today

## Secrets / gitignore

`.loom/accounts.env` **and** the pre-existing `.loom/tokens/` are added to the daemon-managed `EPHEMERAL_PATTERNS` (`loom-daemon init`) and the repo `.gitignore`, so installed repos never commit OAuth keys. The home master lives outside any repo.

## Example

```
$ loom-tokens bootstrap --dry-run
Account sources:
  home: ~/.loom/accounts.env
  repo: <repo>/.loom/accounts.env
Effective accounts (3):
  alice  alice@x.com  [repo (overrides home)]
  bob    bob@x.com    [home]
  ci     ci@x.com     [repo]
```

## Tests

- `loom-tools/tests/tokens/test_bootstrap.py`: 30+ new cases — source resolution, email-keyed merge (add/override/case-insensitive/ordering), cross-source filename collision, provenance report, `--no-home`/`--home-env`/`LOOM_ACCOUNTS_ENV`, manifest `source`.
- New `conftest.py` autouse fixture disables the home master by default so tests never read a real `~/.loom/accounts.env`.
- All 171 token tests pass; `loom-daemon` `init::post_init` tests pass.

Closes #3695

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ Thumbs.db
 .loom/usage-cache.json
 .loom/claude-config/
 .loom/tokens/
+.loom/accounts.env
 .loom/sweep-checkpoint/
 .loom/spawn-loop.pid
 .loom/spawn-loop-state.json
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -351,7 +351,7 @@ Settings applied: squash merge only (no merge commits/rebase), delete branches o
 
 ### Multi-Account Token Pool
 
-For environments that rotate among multiple Claude OAuth accounts, Loom can bootstrap a per-account token pool at `.loom/tokens/` from numbered triples in `.env`:
+For environments that rotate among multiple Claude OAuth accounts, Loom can bootstrap a per-account token pool at `.loom/tokens/` from numbered `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` / `ACCOUNT_TOKEN_FILE_N` triples:
 
 ```env
 ACCOUNT_EMAIL_1=user1@example.com
@@ -363,13 +363,32 @@ Run `loom-tokens bootstrap` to materialize the pool:
 
 ```bash
 loom-tokens bootstrap            # Idempotent — only writes new/missing tokens.
-loom-tokens bootstrap --dry-run  # Preview without writing.
-loom-tokens bootstrap --force    # Overwrite on-disk tokens that have drifted from .env.
+loom-tokens bootstrap --dry-run  # Preview + print the effective merged account set.
+loom-tokens bootstrap --force    # Overwrite on-disk tokens that have drifted from source.
 ```
 
-Each account becomes `.loom/tokens/<file>.token` (mode `0600`). An `index.json` manifest is written alongside with sha256 fingerprints (8 chars) for drift detection — **no secret material is stored in the manifest**. Numbering gaps are allowed; partial triples are skipped with a warning.
+Each account becomes `.loom/tokens/<file>.token` (mode `0600`). An `index.json` manifest is written alongside with sha256 fingerprints (8 chars) for drift detection plus each account's `source` (home/repo) — **no secret material is stored in the manifest**. Numbering gaps are allowed; partial triples are skipped with a warning.
 
 `.loom/tokens/` is gitignored. The pool is consumed by external rotation logic (e.g. a `claude-wrapper.sh` that picks the least-used token); only the bootstrap step is provided here.
+
+#### Home-dir master + per-repo override (#3695)
+
+Rather than re-declaring the same account triples in every repo's `.env`, declare them **once** in a home-dir master and let each workspace add or override on top of it:
+
+| Source | Default location | Override |
+|--------|------------------|----------|
+| **Home master** | `~/.loom/accounts.env` | `LOOM_ACCOUNTS_ENV` env var (`""` disables the master); `--home-env <path>` / `--no-home` on `bootstrap` |
+| **Repo-local** | `<repo>/.loom/accounts.env` if present, else legacy `<repo>/.env` | `--env <path>` on `bootstrap` |
+
+`loom-tokens bootstrap` reads both sources and **merges them by account email** (`ACCOUNT_EMAIL`), with the repo-local source winning:
+
+- An email present **only in the master** is inherited into the pool.
+- An email present **only in the repo** is added.
+- An email present in **both** → the repo entry overrides (e.g. to rotate a key or repoint the token file).
+
+To *exclude* a master account from one repo, pin the subset you want with `loom-tokens pin` — the merge only ever adds/overrides, never subtracts. The effective merged set (and where each account came from) is printed by `bootstrap` and `bootstrap --dry-run`. A repo with only a legacy `.env` and no master behaves exactly as before.
+
+> **Secrets**: both `~/.loom/accounts.env` and the repo-local `.loom/accounts.env` hold raw OAuth keys. The repo-local file and `.loom/tokens/` are gitignored (installer- and `loom-daemon init`–managed); keep the home master `0600` and outside any repo. A repo can rely entirely on the home master with no local account file at all.
 
 #### Account health probe + ranking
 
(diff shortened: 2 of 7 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed the home-dir master account pool with per-repo override (#3695). This is a well-scoped, security-conscious change that keeps the surface small by doing the merge at `bootstrap` time and materializing into `.loom/tokens/` — `select`/`check`/`pin` and the Rust daemon are untouched, which is the right call.

### Verification
- **Python**: `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -q` → **171 passed**.
- **Rust**: `cargo test --lib init::post_init` → **20 passed** (incl. `covers_all_source_repo_gitignore_patterns`, confirming `EPHEMERAL_PATTERNS` stays in sync with the source `.gitignore`).
- **CI**: all required checks green; mergeStateStatus **CLEAN**.
- Working tree clean after the suite — tests write only into `tmp_path`, never a real home master.

### What I checked closely
- **Email-keyed merge semantics** (`merge_accounts`): repo overrides home in place (`source="repo-override"`, keeps master ordering position, takes repo key/file/index), repo-only emails append after, home-only inherited. Comparison is `.strip().lower()` → case-insensitive. Deterministic ordering (home index order, then repo index order). Covered by `TestMergeAccounts` + `TestBootstrapMerge`.
- **Cross-source duplicate-filename guard**: the `seen_files` check runs on the *merged* set, so two different emails pointing at the same token filename abort with `ValueError` before any write. Verified by `test_cross_source_duplicate_filename_aborts`.
- **Secrets / gitignore**: `.loom/accounts.env` and `.loom/tokens/` are added to both the repo `.gitignore` and the daemon-managed `EPHEMERAL_PATTERNS`; legacy `.env` was already ignored. The home master lives outside any repo. Manifest stores only 8-char sha256 fingerprints, never key material (`test_writes_index_json` asserts no `ACCOUNT_KEY` leaks). Token files 0600, dir 0700, atomic temp+rename.
- **Test isolation**: the autouse `conftest.py` fixture sets `LOOM_ACCOUNTS_ENV=""` so no test can read a developer's/CI's real `~/.loom/accounts.env`; opt-in tests either pass `home_env_path=` explicitly or set the env var. Correct.
- **Backward compatibility**: with no home master present, `merge_accounts([], repo)` returns the repo set unchanged — a repo with only a legacy `.env` behaves exactly as before. The sentinel (`_HOME_UNSET`) cleanly distinguishes "disable master" (`None`) from "use default resolution".
- **Filename safety**: `_TOKEN_FILE_RE` plus explicit `/`/`\\` rejection blocks traversal (`../escape.token`); a bare `.token` is also rejected (regex requires ≥1 char before the suffix).

### Non-blocking nits (optional, no follow-up required)
- `--home-env` combined with `--no-home` silently lets `--no-home` win; fine, just undocumented.
- Overriding/removing an account leaves its old `.token` file on disk (pre-existing bootstrap behavior — no cleanup of removed accounts — out of scope here).

Solid work. Approving.

Issue #3725, opened 2026-07-22
feat(sweep): experiment-mode model-cost instrumentation to collect the retune evidence (unblocks #3718)


rjwalters:
## Original Issue

> ## Context
>
> #3718 asks whether the Builder default should flip `opus → sonnet` (with escalate-on-Judge-rejection), gated on measured data. That data does not exist and cannot be generated by observation alone: whatever the current Builder default is (opus), passive collection only ever measures that one arm. The inequality has two sides — we need a balanced comparison of *both* policies, generated at volume. The volume lives in the daemon canaries (continuous multi-account rotation), not the loom repo.
>
> This issue ships the instrumentation that produces that evidence. It is the blocker-of for #3718: ship this → run canaries a few days → harvest → #3718 evaluates the inequality.
>
> **Design (summary):** tri-state `sweep.modelExperiment` (`off`/`observe`/`experiment`) + `LOOM_MODEL_EXPERIMENT` env; two arms (A=opus-first, B=sonnet-first+escalate) mapping onto #3718's inequality; per-issue deterministic stratified arm assignment (resume-safe, keyed off issue number, stratified by complexity marker #3702); a durable gitignored JSONL store `.loom/stats/sweep-model-stats.jsonl` surviving merge; a "real-token capture at the sweep-skill result-collection point"; a harvest/report command; canary-only guardrails with a loud startup banner. Full original design retained in issue history.

---

## Curator Enhancement (verified against codebase 2026-07-22)

I verified the design against the current tree. The setting-plumbing, checkpoint-tap, arm-assignment, and harvest anchors all check out. **One central assumption — "real per-subagent input/output tokens are surfaced to the sweep skill at the Task-result boundary" — does not hold**, so the token acceptance criteria are degraded to what is mechanically achievable, and one **missing arm/complexity interaction** is promoted to a hard AC. Everything else is tightened with exact anchors.

### Problem Statement

#3718 (`loom:blocked`, "execute the measurement-gated Builder default retune opus→sonnet") cannot proceed without a balanced A/B measurement of both Builder policies at volume. This issue is the instrumentation that produces that evidence. It is the explicit blocker-of #3718.

### 1. Setting plumbing (verified)

[...]

PR #3727, opened 2026-07-22
feat(loom): opt-in session-transcript archival to a durable location (#3726)

### LINKED ISSUE #3726: feat(loom): opt-in session-transcript archival to a durable location (long-term storage for audit + metrics harvest)
## Context

Loom sessions (and their per-role subagent Tasks) write full JSONL transcripts under `~/.claude/projects/<session-uuid>/` — including `subagents/agent-<id>.jsonl` per Builder/Judge/Doctor invocation, each carrying per-message `usage` (input/output + cache split) and `model`. These are the ground-truth record of what every agent did and what it cost. Today they live only in the local `~/.claude/projects` tree, which:

- is **not durably retained** — subject to Claude Code's own cleanup over time;
- is **per-machine** — a canary host's transcripts are stranded on that host;
- has **no manifest** tying `agent-<id>` → repo / issue / role / sweep.

This blocks any workflow that needs transcripts after the fact: audit/debugging of a past sweep, and — concretely — the exact per-role cost harvest that #3725 needs to produce the #3718 model-retune evidence over a multi-day canary run.

## Proposal

An **opt-in** mode where Loom automatically copies session transcripts to a specified long-term storage location.

### Setting
- `loom.transcriptArchive.enabled` (default `false`) + env override `LOOM_TRANSCRIPT_ARCHIVE` (env-over-config, matching the existing convention).
- `loom.transcriptArchive.dir` — destination path (v1: local filesystem path, e.g. an external/scratch volume or a synced dir). Remote/object-storage backends (s3://, gcs://) are an explicit follow-on, not v1.

### What is copied
- The session's own transcript **and** all its `subagents/agent-*.jsonl`.
- A small **manifest** per session: `agent-<id>` → {repo, sweep issue(s), role, arm/attempt if present, model, start/end ts}. This is the join key that makes archived transcripts harvestable (and directly serves #3725's join-key AC).

### When
- Copy-on-completion by default (captures the final full transcript per session/subagent).
- Optional **periodic sync** for durability against mid-run pruning on long canary runs (the retention backstop #3725 calls out).
- Idempotent (content-addressed or ts+size skip), so re-runs don't duplicate.

### Layout at destination
- Organized `/<repo>/<date>/<session-uuid>/…` so it's both auditable and machine-harvestable.

## Guardrails (load-bearing — transcripts are sensitive)

- **Off by default.**
- **Transcripts can contain secrets** — full tool I/O, printed command output, possibly `.env` contents or token values that scrolled through a shell. The archive location must be treated as sensitive: created mode `0700`, files `0600`; if the dir is inside a repo it MUST be gitignored (refuse / warn otherwise). Document that the operator is responsible for the archive location's security (same posture as the `.loom/tokens/` / home-master secrets guidance in CLAUDE.md).
- Loud one-line startup log naming the archive destination when enabled, so an operator always knows transcripts are being copied off-box.

- [ ] `loom.transcriptArchive.enabled` / `LOOM_TRANSCRIPT_ARCHIVE` (default off, env-over-config); zero behavior change when unset.
- [ ] On session/sweep completion (and optionally on a periodic sync), copy the session transcript + all subagent transcripts to `transcriptArchive.dir` under a `/<repo>/<date>/<session-uuid>/` layout, idempotently.
- [ ] Emit a per-session manifest mapping `agent-<id>` → repo / issue / role / model / (arm,attempt if present) / timestamps.
- [ ] Guardrails: off by default; destination `0700`/`0600`; gitignore-or-refuse if inside a repo; loud startup banner; secrets-in-transcripts documented.
- [ ] Works on both the in-session subagent path and the daemon detached-child path (verify the canary child's transcripts are captured).
- [ ] Docs: CLAUDE.md section on enabling archival + the secrets caveat.

## Relationship to other issues

- **Serves #3725** — the durable-retention backstop for its exact-per-role-cost harvest, and its manifest supplies #3725's `agent-<id>`→outcome join key. #3725 can harvest from the archive instead of racing `~/.claude/projects` pruning.
- Independent value beyond #3725: audit/debugging of past sweeps, cross-host aggregation of canary runs, and a general metrics substrate.

rjwalters:
Closes #3726

## What this does

Adds an **opt-in** mode (off by default, zero behavior change when unset) that copies Claude Code session transcripts — and every per-subagent transcript + `.meta.json` sidecar — to a durable local filesystem location for after-the-fact audit and the #3725 per-role cost harvest.

## Completed (v1)

- **`defaults/scripts/archive-transcripts.sh`** — cron-friendly, idempotent copier modeled on `probe-tokens.sh`.
  - Resolves the source base **`CLAUDE_CONFIG_DIR`-aware** (`${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/<cwd-slug>/`) — never hard-codes `~/.claude` (CORRECTION 2).
  - Captures **both** the top-level `<uuid>.jsonl` file **and** the sibling `<uuid>/` companion dir (`subagents/agent-*.jsonl` + `*.meta.json` + `tool-results/`) into `<dir>/<repo>/<date>/<uuid>/` (CORRECTION 1). Idempotent via size+mtime skip.
  - Emits an **agent-id-keyed `index.json`** (schema `loom.transcript-index/v1`) that **reuses role+issue from the existing `.meta.json` sidecars** (CORRECTION 3) and adds only the loom-side context (repo, sweep issue, model, start/end ts, arm/attempt). This is #3725's join key.
- **Enablement**: `LOOM_TRANSCRIPT_ARCHIVE` env (env-over-config, string-valued like `guards.rmScope`: a path enables, `""`/`off`/`0`/`no`/`disabled` disables) > `.loom/config.json → loom.transcriptArchive.{enabled,dir}` > default(off).
- **Guardrails (hard)**: off by default; dest `0700` / files `0600`; **gitignore-or-refuse** when the dest is inside a git repo (via `rev-parse --show-toplevel` + `check-ignore`); loud startup banner naming the dest; secrets caveat documented with the `.loom/tokens/` posture.
- **`claude_config.py`**: added `resolve_claude_base_dir()` / `resolve_projects_dir()` shared helpers (+ pytest) so #3725 reuses the same base logic.
- **`sweep.md`**: completion-hook invocation (safe to run unconditionally — no-op when disabled) at sweep settle.
- **`defaults/CLAUDE.md`**: new "Session Transcript Archival" section (enabling, `CLAUDE_CONFIG_DIR`-aware base, secrets caveat, out-of-scope remote backends).
- **Tests**: `defaults/scripts/tests/test-archive-transcripts.sh` (23 assertions — copy/modes/index/idempotency/`CLAUDE_CONFIG_DIR` override/gitignore-refuse/gitignore-allow/disabled-no-op/env-over-config) + 4 new pytests in `test_claude_config.py`.

## Deferred (honest scope note)

- **Daemon reaper auto-invoke** (`loom-daemon/src/sweep_registry.rs`): the reaper knows the detached child's PID + issue but **not** its Claude session-uuid, so a precise reaper-triggered single-session archive is **not wired in this PR**. Per the curator's guidance this is a documented refinement, not a v1 blocker — the **cron-friendly periodic sync** (`--all-repo-sessions`) is the backstop for the detached-child path, and both `sweep.md` and CLAUDE.md say so explicitly. No Rust files were touched.
- **Remote / object-storage backends** (`s3://`, `gcs://`, rsync): explicit v1 out-of-scope per the issue; local filesystem only.

## Verification

- `bash -n` + `shellcheck -S warning` clean on both new scripts.
- `bash defaults/scripts/tests/test-archive-transcripts.sh` → 23/23 pass.
- `PYTHONPATH=loom-tools/src pytest loom-tools/tests/test_claude_config.py` → 64/64 pass.
- `cargo test --test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` → 17/17 pass (doc-lint green after sweep.md edit).

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -1333,6 +1333,22 @@ Wave annotation makes it easier to triage failures (e.g., "every issue in wave 2
 
 **`rate-limited` vs `blocked` (issue #3683).** These are semantically distinct — reuse the `TOKEN_EXPIRED` / `TOKEN_EXHAUSTED` vocabulary from `.loom/scripts/lib/classify-error.sh` for the reason. `blocked (...)` means the **work itself** failed (build error, doctor cycle exhausted) and a human must fix the actual problem. `rate-limited (...)` means only that a role subagent was killed by an account rate limit mid-phase, so an **extra orchestrator pass** was needed to reach the phase's expected exit state — it says nothing about work quality. A `rate-limited (resumed: <what completed>)` outcome already succeeded (the mid-phase-death recovery finished the missing steps); only a `rate-limited (unresumable: ...)` outcome — where the forge state cannot be recovered without human help — needs attention.
 
+## Session Transcript Archival (completion hook, #3726)
+
+After the entire sweep has settled (issue list exhausted / all PRs processed) and just before printing the Summary Output, run the transcript archiver once so this session's transcript and all its subagent transcripts are captured to durable storage:
+
+```bash
+./.loom/scripts/archive-transcripts.sh
+```
+
+This is **safe to run unconditionally** — the archiver is a **no-op unless archival is opted in** (env `LOOM_TRANSCRIPT_ARCHIVE=<dir>` or `.loom/config.json → loom.transcriptArchive.enabled`). When enabled it copies `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/<cwd-slug>/` (the session's own `<uuid>.jsonl` plus the sibling `<uuid>/subagents/agent-*.jsonl` + `.meta.json` sidecars) into `<dir>/<repo>/<date>/<uuid>/`, emits an `agent-<id>`-keyed `index.json` join key, and is **idempotent** (a re-run copies nothing new).
+
+**Caveat (why the cron backstop still matters):** at this completion point the session's own top-level `<uuid>.jsonl` may still be mid-flush — the final orchestrator messages can lag. The completion hook reliably captures finished subagents; the durable tail is guaranteed only by the cron-friendly periodic sync documented in CLAUDE.md ("Session Transcript Archival"). Run both.
+
+**Guardrails apply** (off by default; destination `0700`/files `0600`; refuses if the destination is inside a git repo but not gitignored; prints a loud banner naming the destination when enabled). See CLAUDE.md → "Session Transcript Archival" for the full contract and the secrets caveat.
+
+> **Daemon detached-child path (v1 refinement, not a blocker).** For sweeps dispatched by the daemon as detached children, the reaper in `loom-daemon/src/sweep_registry.rs` knows the child's PID and issue but **not** its Claude session-uuid, so it cannot yet trigger a precise single-session archive on exit. v1 relies on the **periodic sync** (which copies all recent sessions for the repo regardless of uuid mapping) as the backstop for the detached path; a reaper-triggered auto-invoke keyed on the child's session-uuid is a documented follow-on.
+
 ## Stop Conditions
 
 Stop processing and print the summary when any of these conditions hold:
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -743,6 +743,76 @@ The ladder is configured in `.loom/config.json`:
 }
 ```
 
+### Session Transcript Archival (opt-in, #3726)
+
+Claude Code writes a full JSONL transcript for every session — and a per-subagent
+transcript for every Builder / Judge / Doctor Task — under
+`${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/<cwd-slug>/`. These are the
+ground-truth record of what each agent did and what it cost (per-message `usage`
++ `model`), but they live only on the local box and are subject to Claude Code's
+own pruning. `archive-transcripts.sh` copies them to a durable location so a
+multi-day canary run can be audited / cost-harvested after the fact (serves the
+#3725 per-role cost harvest — the archived index is its `agent-<id>` join key).
+
+**Off by default. Zero behavior change unless you turn it on.** Enable via env or
+config (env-over-config precedence, matching the `guards.rmScope` string pattern):
+
+```bash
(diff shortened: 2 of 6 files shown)


rjwalters:
LGTM: Thorough review of the opt-in transcript archiver. Verified all acceptance-defining points against the diff:

- **CLAUDE_CONFIG_DIR-aware base**: `CLAUDE_BASE="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"` (line 211); never hard-codes ~/.claude. REPO_ROOT (git rev-parse) is used only for the <repo> label + config read, not the projects slug. The Python resolve_claude_base_dir()/resolve_projects_dir() helpers mirror it and are covered by 4 pytests (override honored, empty string falls back).
- **cwd-slug from RAW cwd**: `SLUG="$(slugify "$SOURCE_CWD")"` from $PWD (line 406), NOT the git-canonicalized path — correct for symlinked checkouts / macOS /var→/private/var. Inline comment documents the rationale.
- **jq `// null` guards**: every optional field (role/issue/description/model/arm/attempt/start_ts/end_ts + top-level sweep_issue) guarded so a filtered-out optional collapses to null, not the whole object.
- **HARD guardrails**: dest 0700, files 0600, index 0600; gitignore-or-refuse walks to nearest existing ancestor + check-ignore trailing-slash probe, running only after the enable-gate; loud stderr banner naming the dest; secrets caveat documented in CLAUDE.md with the .loom/tokens/ posture. Each is exercised by a dedicated test case.
- **Deferred scope is honest**: daemon-reaper auto-invoke deferred with the --all-repo-sessions periodic-sync backstop documented in both sweep.md and CLAUDE.md; no Rust touched. Acceptable per the issue.

Also verified: disabled path is a true zero-side-effect no-op exiting before the guard/banner; `set -uo pipefail` (no -e) is right for an idempotent copier; test-case-7's refusal RC is sound because the harness also sets pipefail. 23-assertion bash test + pytests, all green.

CI: Installer Integration Tests, Shellcheck, Detect Changes all pass; Rust/MCP jobs correctly skipping (no Rust/MCP files changed — that's the source of the earlier UNSTABLE). mergeStateStatus CLEAN. Approved.

PR #3732, opened 2026-07-22
fix(daemon): propagate experiment env + set workspace cwd on detached sweep children

### LINKED ISSUE #3730: fix(daemon): propagate experiment env + set workspace cwd on detached sweep children
## Context

Surfaced by a canary-readiness trace of the #3725 experiment feature. In daemon mode the daemon spawns one detached `claude -p "/loom:sweep N"` child per issue (`loom-daemon/src/sweep_registry.rs` ~L545/spawn_child). Two gaps make env-based experiment enablement and config resolution fragile on that path:

1. **Env not propagated.** The child `Command` sets only `LOOM_TERMINAL_ID` and `LOOM_WORKSPACE` (`sweep_registry.rs:573-578`). There is no `env_clear()`, so the child inherits the *daemon process's* environment — meaning `LOOM_MODEL_EXPERIMENT` / `LOOM_MODEL_EXPERIMENT_CANARY` only reach the child if the **daemon itself** was started with them. An operator who exports those before dispatching a sweep gets silent no-effect (experiment downgrades to `observe`). 
2. **No explicit cwd.** `spawn_child` sets no `current_dir` on the child, so the child's relative `.loom/config.json` read (`loom_tools/sweep_experiment.py:71`) and `archive-transcripts.sh`'s cwd-slug both depend on the daemon's own cwd being the workspace root. Normally true, but unstated/fragile.

## Proposal

- **Explicitly forward the experiment-related env vars** to the detached child: at minimum `LOOM_MODEL_EXPERIMENT` and `LOOM_MODEL_EXPERIMENT_CANARY` (and `LOOM_TRANSCRIPT_ARCHIVE` if present), via the child `Command`'s env. Prefer an explicit allowlist forward (do NOT blanket-clear or blanket-copy) so experiment enablement is reliable regardless of how the daemon itself was launched.
- **Set `current_dir` on the child** to the resolved workspace root, so the child's relative `.loom/config.json` and the archiver slug resolve deterministically.

## Interaction (coordinate with the guardrail-hardening issue)

The sibling guardrail-hardening issue requires the canary *confirmation* to come from an **uncommitted** signal (env var or gitignored sentinel), never committed config. This issue's env-forwarding is what makes an **env-based canary confirmation** actually reach the daemon child — so the two compose: config may carry the mode, but the confirmation arrives via forwarded env / sentinel. Keep the forwarded var names consistent with the guardrail issue.

- [ ] `spawn_child` forwards `LOOM_MODEL_EXPERIMENT`, `LOOM_MODEL_EXPERIMENT_CANARY` (+ `LOOM_TRANSCRIPT_ARCHIVE` if set) to the detached child via an explicit env allowlist (no blanket clear/copy).
- [ ] `spawn_child` sets `current_dir` to the resolved workspace root.
- [ ] Rust unit test asserting the child Command carries the forwarded vars + cwd when they are set in the daemon env, and does NOT set them when unset.
- [ ] No behavior change when none of the vars are set (byte-for-byte spawn argv/env otherwise unchanged).

## References
- #3725 (experiment feature), #3716/#3722 (dispatch `--model` plumbing precedent for spawn_child), #3454 (daemon dispatch).

rjwalters:
Closes #3730

## What

In `loom-daemon/src/sweep_registry.rs`, `spawn_child` now:

1. **Forwards experiment env vars via an explicit allowlist** — `LOOM_MODEL_EXPERIMENT`, `LOOM_MODEL_EXPERIMENT_CANARY`, `LOOM_TRANSCRIPT_ARCHIVE` (new `EXPERIMENT_ENV_ALLOWLIST` const). Each is forwarded with a `var_os` guard so an **unset** var is not forwarded and an **empty-string** value is not forwarded. No blanket `env_clear`/copy — the spawn stays a byte-for-byte no-op when none are set. This makes env-based experiment enablement reach the detached child reliably.

2. **Pins the child's `current_dir`** to `self.config.workspace_root`, so the child's relative `.loom/config.json` read (`loom_tools/sweep_experiment.py`) and `archive-transcripts.sh`'s cwd-slug resolve deterministically instead of depending on the daemon's own cwd.

Byte-exact var names verified against `sweep_experiment.py` (L98/126) and `archive-transcripts.sh` (L116).

## Tests

Two new `#[serial]` unit tests (env-mutating, per curator guidance):

- `dispatch_forwards_experiment_env_and_sets_cwd` — vars set in the daemon env reach the child + cwd is pinned to the workspace root.
- `dispatch_does_not_forward_unset_experiment_env` — vars are observed as unset by the child when not set in the daemon env; cwd still pinned.

The fixture records the forwarded vars + `pwd -P` into its log. `cargo test` green (31 sweep_registry tests pass); `cargo build --workspace` clean. The two remaining clippy pedantic warnings (`forge_parser.rs`, `main.rs`) are pre-existing and untouched by this change.

## Acceptance criteria
- [x] `spawn_child` forwards the three vars via explicit allowlist (no blanket clear/copy)
- [x] `spawn_child` sets `current_dir` to the resolved workspace root
- [x] Rust unit test: child carries vars + cwd when set; not set when unset
- [x] No behavior change when none of the vars are set

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -76,6 +76,21 @@ pub const WORKSPACE_ENV: &str = "LOOM_WORKSPACE";
 /// "recently exited sweeps should still show up in `list_sweeps`".
 pub const TERMINAL_RETENTION_SECS: i64 = 3600;
 
+/// Issue #3730: experiment-related env vars forwarded to the detached sweep
+/// child via an EXPLICIT ALLOWLIST (never a blanket env_clear/copy). Byte-exact
+/// names verified against `loom_tools/sweep_experiment.py` (`LOOM_MODEL_EXPERIMENT`,
+/// `LOOM_MODEL_EXPERIMENT_CANARY`) and `.loom/scripts/archive-transcripts.sh`
+/// (`LOOM_TRANSCRIPT_ARCHIVE`). Forwarding these makes env-based experiment
+/// enablement reliable regardless of how the daemon itself was launched — an
+/// operator can export them right before dispatching and have them reach the
+/// child. Each is forwarded only when set to a non-empty value (see
+/// `spawn_child`), so the spawn is a no-op when none are set.
+pub const EXPERIMENT_ENV_ALLOWLIST: &[&str] = &[
+    "LOOM_MODEL_EXPERIMENT",
+    "LOOM_MODEL_EXPERIMENT_CANARY",
+    "LOOM_TRANSCRIPT_ARCHIVE",
+];
+
 // ============================================================================
 // Registry
 // ============================================================================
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM: Verified against the #3730 acceptance criteria.

- **Explicit allowlist, no blanket clear/copy.** `EXPERIMENT_ENV_ALLOWLIST` forwards exactly the 3 vars via a per-name `cmd.env` loop — no `env_clear`/copy, so the whole environment is never exposed. No security footgun.
- **True no-op when unset.** `std::env::var_os(name)` skips unset vars, and the inner `!val.is_empty()` guard skips empty-string values too (matching archive-transcripts.sh, which treats `""` as disabled). Spawn is byte-for-byte unchanged when none are set — confirmed by `dispatch_does_not_forward_unset_experiment_env`.
- **Var names byte-exact.** `LOOM_MODEL_EXPERIMENT` / `LOOM_MODEL_EXPERIMENT_CANARY` match sweep_experiment.py (lines 98, 126); `LOOM_TRANSCRIPT_ARCHIVE` matches archive-transcripts.sh (line 116).
- **cwd pinned to real workspace root** via `.current_dir(&self.config.workspace_root)` — the daemon-resolved root, not ambient cwd.
- **Both env-mutating tests are `#[serial]`**, with pre-clear + immediate cleanup so leaks can't cross-contaminate sibling serial tests.

CI: Rust Linting & Build + Rust Unit Tests both pass (UNSTABLE was pending unit tests, now green; other jobs correctly skipped as non-required).

Issue #3736, opened 2026-07-22
Token selection: ranked strategy always picks the single most-available account (.ranking[0]) — collides with sibling daemons on the same claude-monitor pool → session-limit faults


rjwalters:
**Reporter context:** downstream canary (`kicad-tools`, Loom v0.13.0). Started the agent pool (`.loom/bin/loom start`) while a **sibling orchestrator (`lean-genius`) daemon was already running on the same claude-monitor account pool**. Within ~90s the shared account hit its **session limit** ("resets 4pm"); all Loom workers died before doing any work (~9 concurrent Claude sessions across the two systems on one account).

## Root cause
`loom-tools/src/loom_tools/tokens/select.py` — **Strategy 1 (ranked)**, the common path, is greedy:

> `_try_ranking(...)`: *"read .ranking, return **first** non-exhausted/non-blocked entry"* (docstring line ~124; loop `for name, status in _read_ranking(...)` returns the first eligible).

And the ranking is ordered **most-available-first** — `build_monitor_accounts` sorts by `(status, util_7d ascending, util_5h)` (see `tests/tokens/test_monitor.py::test_join_and_order_by_status_then_util`). So `.ranking[0]` is deterministically the single least-utilized account.

**Every process that reads a fresh `.ranking` picks the same account.** Strategies 2 (`allowlist`) and 3 (`random`) already `rng.shuffle`/random-pick — only the *ranked* path (the one that fires whenever claude-monitor data is fresh, i.e. almost always) is deterministic. Two daemons → same `[0]` → both slam one account to its session limit while other accounts sit idle.

## Impact
- Multi-daemon / multi-workspace setups sharing a claude-monitor pool **serialize onto one account** and trip session limits, even with plenty of aggregate capacity across accounts.
- Observed live: Loom pool + lean-genius daemon both selected the top account; account limited until reset; Loom's Judge/Curator/Hermit workers exited immediately (and with no keeper (#3735) nothing recovered).

## Proposed fix
Pick **randomly among the top-N available entries** (default N=3) instead of always the first, in `_try_ranking`:
- Collect the leading run of `available` (non-exhausted/non-blocked) entries whose `util` is within a small band of the best (or simply the first N eligible), then `rng.choice` among them. `select.py` already `import random` and threads an `rng`, so this is a localized change.
- Make N configurable (`tokens.spreadTopN`, env `LOOM_TOKEN_SPREAD_TOP_N`, default 3; N=1 restores today's greedy behavior).
[...]

Issue #3233, opened 2026-05-02
claude-wrapper.sh: consecutive_failures counter not reset on graceful exit-0


rjwalters:
## Summary

The bash daemon loop in `claude-wrapper.sh` increments `consecutive_failures` on **graceful exit-0** ("nothing to do this cycle") cycles. Result: agents that legitimately have no work get marked FAILING after 10+ idle cycles, and the operator dashboard shows red.

Independent of the rotation work in #3227-#3231, but caught while diagnosing the same incident.

## Reproducer

`herald-agent` runs every 5 min. When daily Mastodon post limit (8/8) is hit, the agent legitimately exits 0 with "nothing to do this cycle, daemon will re-launch". The wrapper logs:

```
2026-05-02T20:01:15Z [wrapper] WARN: Daemon cycle 619: recoverable error
  (exit code: 0, consecutive failures: 51)
```

51 cycles. Exit code 0. Counted as failures. The health check (`launch.sh health`) flags herald as FAILING and asks the operator to "Check logs and restart" — but there's nothing to restart; the agent is doing its job.

## Code

`scripts/agents/claude-wrapper.sh` daemon loop (lines ~679-738):

```bash
case "$error_class" in
    SUCCESS)
        ...
        consecutive_failures=0   # ← reset only on observed work
        ;;
    TIMEOUT)
        ...
        consecutive_failures=0   # ← reset
        ;;
    TOKEN_EXPIRED|TOKEN_EXHAUSTED|...)
        consecutive_failures=$((consecutive_failures + 1))
        ;;
    *)
        # Catch-all: any exit code, including 0 with no successful classification
        log_warn "... recoverable error (exit code: $exit_code, consecutive failures: $consecutive_failures)"
        ;;
esac
```

`classify_error` returns SUCCESS only when stdout matched a "completed successfully" marker. A clean exit-0 with no such marker (legitimate "no work this cycle") falls through to the catch-all and increments the counter.

### A. Reset counter on any exit-0
Simplest. Trust that any clean exit means the agent decided things are fine. Risk: if a bug causes the agent to silently exit 0 with nothing done, we'd never notice via the failure counter.

### B. Add a "NO_WORK" classification
Have the wrapper detect specific phrases like "nothing to do this cycle" / "exiting; nothing actionable" and treat them as SUCCESS. Brittle (string matching) but explicit.

### C. Have agents emit a sentinel file
[...]

PR #3745, opened 2026-07-22
fix(tokens): rotate account + retry on session/usage-limit fault instead of dying (#3738)

### LINKED ISSUE #3738: Wrapper should rotate to a different account and retry on a session/usage-limit fault (currently classified fatal → agent dies). Port lean-genius's mark-bad + rotate pattern
**Reporter context:** Loom v0.13.0 canary. Started the agent pool while a sibling daemon shared the claude-monitor account pool; the shared account hit **"You've hit your session limit · resets 4pm"** and **all Loom workers died immediately without retrying or trying another account** — even though other accounts in the pool had capacity.

## Current behavior (verified against `defaults/scripts/claude-wrapper.sh`, the canonical source — see "Files to edit" below)

Two independent detection layers exist today, and **both** are wrong for this fault:

1. **`start_output_monitor()` / `start_startup_monitor()`** (~lines 897-922 and ~1016-1032) watch CLI output for the interactive limit modal (normalized match on `"Stopandwaitforlimittoreset"`) and the 100%-weekly-limit banner (`"Youveused100ofyourweeklylimit"`). On match they `pkill -INT` the `claude` child and write a `# RATE_LIMIT_ABORT` sentinel to stderr. This part is *fine* — it correctly stops a hung/looping session.
2. **`is_transient_error()`** (~line 795) checks the captured output for that `RATE_LIMIT_ABORT` sentinel and **explicitly returns 1 (non-transient)** — comment at ~line 799: *"Rate limit abort is NOT transient — the CLI hit a usage/plan limit."* `run_with_retry()`'s call site (~line 1443) then returns immediately without retrying, and `main()` exits — this is why the whole worker dies.

Additionally, **the OAuth token is selected exactly once, outside the retry loop**: `spawn-claude.sh` calls `python3 -m loom_tools.tokens.select` (~line 268), exports `CLAUDE_CODE_OAUTH_TOKEN` (~line 302), and only then `exec`s `claude-wrapper.sh` (~line 313). `claude-wrapper.sh`'s `run_with_retry()` (~line 1253) never re-selects or re-exports — every attempt, transient or not, reuses the same token/account.

**Two more findings from source verification (not in the original report):**

- **`.loom/scripts/lib/classify-error.sh`** (source of truth: `defaults/scripts/lib/classify-error.sh`) already defines a `classify_error()` function with a `TOKEN_EXHAUSTED` class (~line 77-81), but (a) it is **narrower** than lean-genius's match set — `hit your (limit|weekly limit)|hit.your.limit` does **not** match "hit your **session** limit", "monthly usage limit", or "out of extra usage" (same multi-word-gap problem lean-genius's own comment calls out), and (b) **it is dead code from `claude-wrapper.sh`'s perspective** — `grep -rn "classify-error.sh\|classify_error" .loom/scripts/*.sh loom-tools/src` outside of `tests/test-spawn-claude.sh` returns nothing. Nothing in `claude-wrapper.sh` or `spawn-claude.sh` sources or calls it. Wiring it in (or folding its logic into `is_transient_error`) is part of this issue's scope, not just widening its regex.
- **`spawn-claude.sh` computes the selected account name (`_name`, ~line 288) but never exports it.** When `claude-wrapper.sh` is invoked via `--use-wrapper` / `exec "$_wrapper"` (~line 313), the wrapper process has `CLAUDE_CODE_OAUTH_TOKEN` in its environment but **no variable identifying which account that token belongs to**. Any mark-bad-the-current-account step needs an account name, so the rotation path needs a way to recover it — either (a) `spawn-claude.sh` exports something like `LOOM_TOKEN_NAME="$_name"` before `exec`, or (b) `claude-wrapper.sh` derives it by content-matching `$CLAUDE_CODE_OAUTH_TOKEN` against `.loom/tokens/*.token` file contents. Prefer (a) plus a (b)-style fallback for the case where the wrapper is invoked directly (not via `spawn-claude.sh`) with `CLAUDE_CODE_OAUTH_TOKEN` pre-set by some other caller. Avoid lean-genius's `ls -t | head -1` mtime heuristic (`rotate_to_working_token`-adjacent code, ~line 1114 of `lean-genius/scripts/agents/claude-wrapper.sh`) — it silently picks the wrong account under concurrent spawns.

## Reference implementation: `rjwalters/lean-genius` `scripts/agents/claude-wrapper.sh`

Confirmed present at `~/GitHub/lean-genius/scripts/agents/claude-wrapper.sh` (also vendored into that repo's own `.loom/scripts/claude-wrapper.sh`). Relevant pieces, with actual current line numbers:

- `classify_error()` (~line 671): the `TOKEN_EXHAUSTED` branch (~line 693-708) matches `hit your limit\|hit your session limit\|monthly usage limit\|weekly limit\|out of extra usage` — the multi-word-gap-aware pattern set to port.
- `mark_token_bad()` (~line 760): appends `"$timestamp $token_name $reason"` to `.loom/tokens/.bad_tokens`. **Loom already has the equivalent primitive as a Python function**: `loom_tools.tokens.bad_tokens.mark_bad(workspace_path, token_name, reason)` (`loom-tools/src/loom_tools/tokens/bad_tokens.py`) — lock-safe (`mkdir`-based, matches `select.py`'s reader), already used elsewhere. There is no bash-callable CLI subcommand for it yet (`loom_tools/src/loom_tools/tokens/cli.py` has `pin`/`unpin`/`unblock` but no `mark-bad`/`block`) — either add one, or call it inline via `python3 -c "from loom_tools.tokens.bad_tokens import mark_bad; mark_bad(...)"`, mirroring the existing inline-Python pattern `spawn-claude.sh` already uses for its auto-unpin check (~line 236-260).
- `rotate_to_working_token()` (~line 796): random-picks among non-bad `.token` files and re-exports `CLAUDE_CODE_OAUTH_TOKEN`. **Loom's equivalent already exists and is better**: `python3 -m loom_tools.tokens.select --workspace "$WORKSPACE" --json` (already invoked by `spawn-claude.sh` ~line 268) implements the full 3-tier ranking→allowlist→random algorithm and already skips `.bad_tokens` via `is_bad()`. Reuse this call for re-selection inside the retry loop rather than reimplementing lean-genius's flat random-pick.
- Retry-loop rotation (`run_daemon()`'s `TOKEN_EXHAUSTED` case, ~line 1130-1141): on exhaustion, `mark_token_bad` → `rotate_to_working_token` → retry immediately in the *same* cycle counter (does not consume a "consecutive failure"); if rotation fails (all tokens bad), falls back to a long backoff rather than exiting fatally.

## Desired behavior

On a session/usage-exhaustion fault:
1. Mark the currently-active account exhausted in `.loom/tokens/.bad_tokens` (via `loom_tools.tokens.bad_tokens.mark_bad`, reason e.g. `"exhausted: <matched phrase>"`).
2. Re-run token selection (`python3 -m loom_tools.tokens.select`), which now skips the just-marked account.
3. Re-export `CLAUDE_CODE_OAUTH_TOKEN` from the newly-selected token and retry — **without** counting the attempt against `MAX_RETRIES` (this is rotation, not transient backoff; `run_with_retry`'s `attempt` counter at ~line 1267/1485 must not increment for this branch, or the whole pool of accounts could burn through `MAX_RETRIES` before every account gets a fair shot).
4. Terminate only when selection reports **no eligible account** — `loom_tools.tokens.select.select_token()` raises `EmptyTokenPoolError` in that case (bash: the `python3 -m loom_tools.tokens.select` CLI exits `78`/`EX_CONFIG`, matching `spawn-claude.sh`'s existing convention at ~line 279). On that condition, exit with a clear, greppable "whole pool exhausted" message (e.g. a `# ACCOUNT_POOL_EXHAUSTED` sentinel mirroring the existing `# RATE_LIMIT_ABORT` / `# AUTH_PREFLIGHT_FAILED` / `# MCP_PREFLIGHT_FAILED` sentinel convention at ~lines 905, 919, 1022, 1029, 1522, 1533) rather than the generic "Max retries exceeded" message at ~line 1488.

## Proposed implementation approach (Builder has latitude on exact structure)

1. In `defaults/scripts/lib/classify-error.sh`, widen the `TOKEN_EXHAUSTED` match (~line 77-81) to the lean-genius phrase set: `hit your (limit|session limit|weekly limit)|monthly usage limit|out of extra usage` (keep the exit-code-first guard above it intact — #3233 must not regress).
2. In `defaults/scripts/claude-wrapper.sh`:
   - Thread through (or derive) the active account name as described above.
   - In `run_with_retry()`, before or in place of the `is_transient_error()` check (~line 1443), classify the captured output for account-exhaustion (reuse the widened `classify-error.sh` pattern, or fold the same regex directly into `is_transient_error`/a new sibling check — Builder's call, but the two files' patterns must not drift apart, so prefer sourcing `classify-error.sh` from `claude-wrapper.sh` rather than duplicating the regex).
   - On a match: mark-bad → re-select → re-export → log → `continue` the loop *without* incrementing `attempt`.
   - On re-selection failure (no eligible account): log the pool-exhausted sentinel + message and exit non-zero (do not fall into the generic `MAX_RETRIES` exceeded path).
3. In `defaults/scripts/spawn-claude.sh`, export the selected account name (e.g. `LOOM_TOKEN_NAME`) before `exec`ing `claude-wrapper.sh` (~line 302-313) so the wrapper doesn't have to guess.
4. Keep the existing `RATE_LIMIT_ABORT`-detecting monitors as the "stop the hung CLI" mechanism — only the *post-mortem classification and response* changes, not the kill-on-modal-detection behavior.

## Acceptance criteria

- [ ] `classify_error()` in `defaults/scripts/lib/classify-error.sh` returns `TOKEN_EXHAUSTED` for all of: `"You've hit your limit"`, `"You've hit your session limit · resets 4pm"`, `"You've hit your weekly limit · resets ..."`, `"You've hit your org's monthly usage limit"`, `"You're out of extra usage · resets ..."` (mind the multi-word gaps — a naive `hit.your.limit`-style pattern will NOT catch the "session"/multi-word variants; verify each phrase individually, not just the legacy short form already covered by the existing test vector).
- [ ] `classify_error()` continues to return `SUCCESS` for exit-code-0 output that merely *mentions* one of the above phrases in prose (no #3233 regression — exit-code-first check must stay above the new patterns).
- [ ] `claude-wrapper.sh`'s retry loop, on detecting account exhaustion (from the CLI's captured output or the existing `RATE_LIMIT_ABORT` sentinel path), marks the current account bad in `.loom/tokens/.bad_tokens` with a timestamp + account name + reason, re-runs token selection, re-exports `CLAUDE_CODE_OAUTH_TOKEN`, and retries — verified this retry does **not** decrement remaining `MAX_RETRIES` budget (e.g. run with `LOOM_MAX_RETRIES=1` and 2+ accounts in the pool, all initially healthy, first one exhausted — the wrapper still tries the second account rather than exiting after 1 attempt).
- [ ] When every account in the pool is marked bad (selection raises `EmptyTokenPoolError` / CLI exits 78), the wrapper exits with a distinct, greppable log line (not the generic "Max retries (N) exceeded" message) and a non-zero exit code.
- [ ] `spawn-claude.sh` exports enough information (account name) for `claude-wrapper.sh` to identify and mark bad the currently-active account without guessing from file mtimes.
- [ ] New test coverage added to `defaults/scripts/tests/test-spawn-claude.sh` (or a sibling test file) covering: the widened `classify_error` TOKEN_EXHAUSTED vectors (session/weekly/monthly/extra-usage phrasings + the SUCCESS-on-exit-0 regression vector), and a wrapper-level rotation test using a fixture `.loom/tokens/` directory with 2+ fake `.token` files and a stub `claude` (mirroring the existing stub-claude pattern used for the spawn-claude tests in that same file, ~line 180-243) that emits an exhaustion phrase on the first account and succeeds on the second — asserting both accounts get tried and `.bad_tokens` gets the first account's entry.
- [ ] `bash .loom/scripts/tests/test-spawn-claude.sh` (or `defaults/scripts/tests/test-spawn-claude.sh` directly) passes.
- [ ] No regression to the existing `is_transient_error()` / `classify_error()` transient-error test vectors (rate limit, 5xx, network, MCP, refusal, CWD-deleted).

## Files to edit (verify against `origin/main` before starting — `.loom/scripts/` is the gitignored, installer-materialized copy of `defaults/scripts/`; they are byte-identical in this checkout but only `defaults/` is tracked)

- `defaults/scripts/claude-wrapper.sh` (retry loop, `is_transient_error`, monitors — canonical source; installed copy lives at `.loom/scripts/claude-wrapper.sh`)
- `defaults/scripts/lib/classify-error.sh` (widen `TOKEN_EXHAUSTED` match; currently unreferenced by any non-test caller)
- `defaults/scripts/spawn-claude.sh` (export account name before `exec`ing the wrapper)
- `defaults/scripts/tests/test-spawn-claude.sh` (new test vectors + rotation test)
- `loom-tools/src/loom_tools/tokens/bad_tokens.py` (existing `mark_bad`/`is_bad` — reuse, no changes expected unless a CLI subcommand is added)
- `loom-tools/src/loom_tools/tokens/select.py` (existing `select_token` CLI — reuse, no changes expected)
- Optionally `loom-tools/src/loom_tools/tokens/cli.py` (only if adding a `mark-bad` subcommand instead of inline Python)

## Out of scope / composes with

- **File-disjoint from #3736** (`loom_tools/src/loom_tools/tokens/select.py` — the ranked-selection top-N spread fix). #3736 reduces the odds of concurrent daemons colliding on the same account *before* the fact; this issue (#3738) recovers *after* a collision has already exhausted an account. They land independently and do not touch the same lines, though the same worktree could pull both without conflict.
- Composes with #3735 (a keeper/respawn layer that recovers if the wrapper still dies) and #3734 (accurate `loom status` visibility into the dead pool) — neither is required for this issue.
- Not in scope: changing the interactive-modal-detection regexes in `start_output_monitor`/`start_startup_monitor` (those already correctly kill a hung session on limit detection) — only the post-detection *response* (fatal exit vs. rotate-and-retry) is in scope.

rjwalters:
Closes #3738

## Problem

A session/usage-exhaustion fault (`You've hit your session limit · resets 4pm`, the 100%-weekly banner, etc.) was classified **fatal** by `claude-wrapper.sh`: the `RATE_LIMIT_ABORT` sentinel made `is_transient_error` return non-transient, so `run_with_retry` returned immediately and the whole worker died — even when other accounts in the pool still had capacity. The OAuth token was also selected exactly once, outside the retry loop, so no rotation was possible.

## Change

On an account-exhaustion fault the wrapper now (issue #3738):

1. **Marks the active account exhausted** in `.loom/tokens/.bad_tokens` via the existing `loom_tools.tokens.bad_tokens.mark_bad` primitive (not a reimplemented file-glob).
2. **Re-runs Loom token selection** (`loom_tools.tokens.select`), which now skips the just-marked account.
3. **Re-exports `CLAUDE_CODE_OAUTH_TOKEN`** and retries **without consuming a `MAX_RETRIES` attempt** — rotation is not transient backoff; charging it against `MAX_RETRIES` could burn the whole budget before every account gets a fair shot.
4. **Terminates only when selection reports no eligible account**, emitting a distinct `# ACCOUNT_POOL_EXHAUSTED` sentinel + message rather than the generic "Max retries exceeded" path. An independent rotation cap (`LOOM_MAX_ACCOUNT_ROTATIONS`, default 20) guards against a loop if the active account can't be identified.

`select.py` is **not** modified (kept file-disjoint from #3736) — this change only *calls* `select_token`.

### Curator-surfaced gaps also fixed

- **`classify-error.sh` `TOKEN_EXHAUSTED` matcher widened** to the multi-word-gap phrase set (`hit your session limit`, `monthly usage limit`, `out of extra usage`, full `weekly limit`) **and wired in**: `claude-wrapper.sh` now *sources* it rather than duplicating the regex, so the two can't drift. Exit-code-first ordering (#3233) preserved.
- **`spawn-claude.sh` exports `LOOM_TOKEN_NAME`** so the wrapper knows which account to mark bad — with a deterministic content-match fallback for direct invocation (deliberately avoids lean-genius's `ls -t | head -1` mtime heuristic).

## Tests

`defaults/scripts/tests/test-spawn-claude.sh` (65 pass):
- Widened `classify_error` vectors — session / weekly / monthly / extra-usage phrasings, legacy short form, plus exit-0-in-prose regression vectors (#3233).
- Wrapper-level rotation test using a 2-account fixture pool + stub `claude` that exhausts the first account and succeeds on the second: asserts both accounts get tried under `LOOM_MAX_RETRIES=1` (proving non-consumption), `.bad_tokens` records only the exhausted account, and whole-pool exhaustion hits the distinct `ACCOUNT_POOL_EXHAUSTED` sentinel path with a non-zero exit.

`shellcheck --severity=warning` clean on all edited scripts.

### DIFF
--- a/defaults/scripts/claude-wrapper.sh
+++ b/defaults/scripts/claude-wrapper.sh
@@ -73,6 +73,30 @@ TERMINAL_ID="${LOOM_TERMINAL_ID:-}"
 # Note: WORKSPACE may fail if CWD is invalid at startup - recover_cwd handles this
 WORKSPACE="${LOOM_WORKSPACE:-$(pwd 2>/dev/null || echo "$HOME")}"
 
+# Python interpreter + active-account tracking for account rotation (#3738).
+LOOM_PYTHON="${LOOM_PYTHON:-python3}"
+# The account name whose OAuth token is currently exported. spawn-claude.sh
+# exports LOOM_TOKEN_NAME before exec'ing this wrapper; when the wrapper is
+# invoked directly (token pre-set by some other caller) it is derived by
+# content-matching the token against the pool at rotation time.
+ACTIVE_TOKEN_NAME="${LOOM_TOKEN_NAME:-}"
+
+# Directory of this script — used to source the shared error classifier and to
+# locate the loom_tools package for the mark-bad / re-select calls (#3738).
+_WRAPPER_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Source the shared error classifier so the wrapper's account-exhaustion
+# detection stays byte-identical to spawn-claude's classification — the two
+# patterns must NOT drift (issue #3738). Resolved relative to this script so it
+# works from both the installed .loom/scripts copy and the tracked
+# defaults/scripts source. If absent (older install), is_account_exhaustion
+# falls back to an inline regex.
+if [[ -f "${_WRAPPER_DIR}/lib/classify-error.sh" ]]; then
+    # shellcheck source=lib/classify-error.sh
+    # shellcheck disable=SC1091
+    source "${_WRAPPER_DIR}/lib/classify-error.sh"
+fi
+
 # Whether --dangerously-skip-permissions was passed (detected in main())
 SKIP_PERMISSIONS_MODE=false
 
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed the fleet-resilience fix for #3738. This is a solid, well-scoped, well-tested change.

**Correctness verified against all five risk areas:**

1. **Rotation does NOT consume `MAX_RETRIES`** — the exhaustion branch `continue`s without incrementing `attempt`, and is independently bounded by `LOOM_MAX_ACCOUNT_ROTATIONS` (default 20). The 2-account rotation test runs under `LOOM_MAX_RETRIES=1` and still reaches `beta`, directly proving non-consumption.
2. **The right account is marked bad** — `ACTIVE_TOKEN_NAME` comes from the `LOOM_TOKEN_NAME` export (now set by `spawn-claude.sh` before exec), with a deterministic content-match fallback (`_derive_token_name`) that avoids lean-genius's `ls -t | head -1` mtime heuristic. Re-selection goes through `loom_tools.tokens.select` and re-exports `CLAUDE_CODE_OAUTH_TOKEN` + `LOOM_TOKEN_NAME`. Test confirms `.bad_tokens` records `alpha` only, not `beta`.
3. **#3233 exit-code-first guard intact (the critical one)** — `classify-error.sh` only widened the `TOKEN_EXHAUSTED` phrase set; the clean-exit→SUCCESS ordering is untouched. Structurally, the wrapper's exhaustion branch is unreachable on exit 0 (the loop returns success at the `exit_code -eq 0` check *before* reaching `is_account_exhaustion`), so even the `RATE_LIMIT_ABORT` check that ignores exit code cannot misfire on a clean exit. Regression vectors #28/#29 confirm exit-0 prose mentioning "session limit" / "out of extra usage" still classifies SUCCESS.
4. **Whole-pool exhaustion emits a distinct terminal sentinel** — `# ACCOUNT_POOL_EXHAUSTED` + "Whole account pool exhausted" message, returns non-zero, and provably avoids the generic "Max retries exceeded" path (asserted explicitly).
5. **Tests genuinely exercise the behavior** — real 2-account fixture pool + stub `claude` that exhausts the first account; asserts rotation success, non-consumption, selective bad-token recording, and the whole-pool terminal path.

**Verification run:**
- `bash defaults/scripts/tests/test-spawn-claude.sh` → 65/65 pass (7 new)
- `shellcheck --severity=warning` → clean on all four scripts
- CI: Shellcheck pass, Installer Integration pass, Detect Changes pass (Rust jobs correctly skipped as path-filtered — no Rust touched)
- `mergeStateStatus`: CLEAN; body uses `Closes #3738`

Nice touches: sourcing `lib/classify-error.sh` so the exhaustion regex can't drift from the classifier, and keeping `select.py` file-disjoint from #3736.

PR #3761, opened 2026-07-22
feat(hooks): gate universal.md injection to once-per-session (#3758)

### LINKED ISSUE #3758: UserPromptSubmit hooks: minimize per-prompt injection; make opt-in the easy default
## Summary

Low-priority companion to the destructive-guard issues. The two
`UserPromptSubmit` hooks — `skill-router.sh` and `methodology-inject.sh` — run
on **every** prompt. Both are already opt-in by config presence, and
`skill-router` was tightened in #3609 (table injected once per session, nothing
on non-matching turns). This issue is about the remaining per-turn cost and
making "as little injection as possible" the easy default.

## Observations

**`methodology-inject.sh`** — activates when `.loom/context/` exists, then
(step 2) **"always inject `universal.md` if it exists"** on every prompt. Once a
repo opts in, `universal.md` is prepended to every single turn regardless of
relevance, up to the 8000-char cap. That is the main per-prompt token cost.

**`skill-router.sh`** — already conservative: opt-in via
`.loom/config/skill-routes.json`, emits nothing on non-matching turns, dedupes
the routing table per session. Little to change here; noted for completeness.

## Requested (all low priority)

- [ ] `methodology-inject`: gate `universal.md` behind relevance, or inject it
      **once per session** (like `skill-router`'s table, #3609), instead of on
      every prompt — configurable, with "every prompt" still available for repos
      that want it.
- [ ] Document clearly (in the hooks README) that both `UserPromptSubmit` hooks
      are opt-in, exactly what triggers injection, and the one-line way to
      disable each (remove the config file/dir, or drop the hook from
      `settings.json`).
- [ ] Consider making the installer's default for these two hooks explicitly
      opt-in (not wired into `settings.json` unless requested), so a fresh
      install has zero per-prompt injection until the operator turns it on.

## Rationale

The broader goal is to minimize hook machinery in the default install while
keeping the catastrophic-error protection. These hooks are the low-severity end
of that: non-blocking, already opt-in, but still worth making "off / minimal by
default, on by choice."

## Non-goals

Not proposing removing either hook — they are useful when configured. Just
reducing always-on per-prompt injection and improving discoverability of the
off switch.

---

### Root cause / current behavior

`defaults/hooks/methodology-inject.sh` (also mirrored to `.loom/hooks/` in
this repo's own install) has this flow today:

1. Opt-in gate: exits silently unless `.loom/context/` exists (line 84).
2. `INJECT_UNIVERSAL` defaults to `true` and is only flippable via
   `.loom/context/config.json`'s `inject_universal` key (line 94, 108) — there
   is **no frequency control**, only an on/off switch. Every matching prompt
   re-injects the full `universal.md` (capped at `max_context_chars`, default
   8000) — see the `append_context` call at line 151-154.
3. Contrast with `defaults/hooks/skill-router.sh` post-#3609 (PR #3616,
   commit `0f8cca29`): the routing table is deduped **once per session** via a
   session-keyed marker file, implemented at lines 171-194:
   - Reads `session_id` from stdin JSON (line 56).
   - Sanitizes it to filename-safe chars: `tr -c 'A-Za-z0-9._-' '_'` (line 183).
   - Marker path: `${MAIN_ROOT}/.loom/logs/skill-router-seen/${SESSION_KEY}`.
   - If the marker exists, `INCLUDE_TABLE=0` (skip); otherwise create it
     (`mkdir -p` + `: >`, best-effort, never fails the hook) and include the
     table this turn.
   - **Missing/empty `session_id` degrades gracefully**: `INCLUDE_TABLE` stays
     `1` (i.e. falls back to per-match inclusion — the pre-#3609 behavior),
     never treated as an error.
   - `.loom/logs/` is already gitignored (`.gitignore` line 133), so the
     `skill-router-seen/` marker directory needs no new gitignore entry — a
     sibling `methodology-inject-seen/` directory will be covered by the same
     rule.

### Implementation guidance

Mirror the #3609 pattern in `methodology-inject.sh`, scoped to the
`universal.md` injection only (role/topic injection is already targeted, not
blanket-every-turn, and is out of scope here — the issue's stated cost driver
is specifically "always inject `universal.md`"):

1. **New config knob** in `.loom/context/config.json` — reuse the existing
   `inject_universal` boolean for on/off, and add a **frequency** key, e.g.:
   ```json
   { "universal_frequency": "session" }
   ```
   Valid values: `"session"` (inject once per session, new default) or
   `"always"` (inject every matching prompt, current/legacy behavior). Read it
   the same defensive way the other config keys are read (`jq -r ... // "session"`,
   never raise the ERR trap on malformed JSON).
2. **Session-keyed marker**, parallel to `skill-router.sh`'s but in its own
   namespace so the two hooks don't collide on the same session_id key (they
   are independent opt-ins and must not share state):
   `${MAIN_ROOT}/.loom/logs/methodology-inject-seen/${SESSION_KEY}` — same
   sanitization (`tr -c 'A-Za-z0-9._-' '_'`), same best-effort `mkdir -p` +
   touch, same graceful degrade to per-turn inclusion when `session_id` is
   missing/empty.
3. **Back-compat default — this is the load-bearing design decision, make it
   explicit in the PR**: flipping the *default* from "always" to "session"
   is a genuine behavior change for any repo that already has `.loom/context/`
   opted in with no `config.json` (or a `config.json` that doesn't set the new
   key) — those repos currently get `universal.md` on every prompt and would,
   after this change, get it once per session instead. This mirrors the
   precedent already set by #3609 for `skill-router.sh` (which changed its
   default from every-prompt to match-only with no back-compat shim). Follow
   that precedent: **default to `"session"`** (satisfies the issue's literal
   ask — "instead of on every prompt" — as the new zero-config behavior), and
   document the flip clearly in `defaults/CLAUDE.md` (see below) plus the PR
   description, since it is a visible change for any repo already using
   `.loom/context/`.
4. Role-context (`INJECT_ROLE`) and topic-context (`INJECT_TOPICS`) injection
   are unaffected — leave their existing every-matching-turn behavior as is
   (they're already conditioned on `LOOM_ROLE`/prompt content, not "always").

### Documentation (bullet 2 of the issue)

There is **no dedicated `defaults/hooks/README.md`** in this repo today (verified:
`find defaults/hooks -iname 'README*'` returns nothing) — the hooks
documentation lives entirely in `defaults/CLAUDE.md`. Treat "the hooks README"
in the issue as referring to `defaults/CLAUDE.md`, specifically:

- The existing **"## Methodology Injection Framework"** section
  (`defaults/CLAUDE.md`, currently documents the opt-in trigger, context file
  structure, and the `config.json` table around line ~1400-1450) needs a new
  row in the config table for `universal_frequency` (`session` | `always`,
  default `session`) and a short callout that the once-per-session dedup
  mirrors #3609's `skill-router.sh` mechanism, including the one-line
  disable: delete/rename `.loom/context/` (or set `enabled: false` in its
  `config.json`).
- `skill-router.sh` currently gets only a single passing mention in
  `defaults/CLAUDE.md` (one line, no dedicated section) — this issue's bullet
  2 explicitly asks for **both** hooks to be documented with their opt-in
  trigger and one-line disable. Add a short parallel subsection (or extend the
  existing passing mention into a proper block) stating: opt-in trigger is the
  presence of `.loom/config/skill-routes.json`; one-line disable is deleting
  that file, or removing the hook's entry from `.claude/settings.json`'s
  `UserPromptSubmit` array.

### Test plan

No test file for `methodology-inject.sh` currently exists under
`defaults/hooks/tests/` (only `test-skill-router.sh` exists, which is a good
structural template — isolated `mktemp -d` git repo, `git init -q`, copy the
hook + config into the temp tree so `MAIN_ROOT` resolves there, `make_input`
helper building `{prompt, session_id}` JSON via `jq -n`). Add
`defaults/hooks/tests/test-methodology-inject.sh` covering:

- `.loom/context/` absent → silent exit 0, no output (existing opt-in gate,
  regression guard).
- `.loom/context/` present, no `config.json` → `universal.md` injected on the
  **first** prompt of a session, but **not** on a second prompt with the same
  `session_id` (new default behavior).
- Same as above but with `universal_frequency: "always"` in `config.json` →
  `universal.md` injected on **every** prompt regardless of session repeats
  (back-compat opt-in verified).
- Missing/empty `session_id` → `universal.md` injected on every call (graceful
  degrade, matches `skill-router.sh`'s documented fallback).
- Role and topic injection are unaffected by any of the above (still fire
  every matching turn) — one assertion per existing behavior to guard against
  scope creep from this change.
- Hook never exits non-zero, never emits invalid JSON (same invariant as the
  existing `test-skill-router.sh` suite).

### Acceptance Criteria

- [ ] `methodology-inject.sh` gains a session-keyed dedup marker
      (`.loom/logs/methodology-inject-seen/<sanitized-session-id>`) for
      `universal.md` injection, mirroring the `skill-router.sh` #3609 pattern
      (own namespace, same sanitization, same best-effort marker creation).
- [ ] New `.loom/context/config.json` key `universal_frequency` (`"session"` |
      `"always"`) controls the behavior; malformed/missing config falls back to
      the new default (`"session"`) without raising the hook's ERR trap.
- [ ] Missing/empty `session_id` on stdin degrades gracefully to per-turn
      inclusion (never treated as an error, never blocks output).
- [ ] Role (`inject_role`) and topic (`inject_topics`) injection behavior is
      unchanged by this issue — still evaluated every matching turn.
- [ ] `defaults/CLAUDE.md` → "Methodology Injection Framework" section
      documents the new `universal_frequency` key, its default, and the
      one-line disable for `.loom/context/`-based injection.
- [ ] `defaults/CLAUDE.md` gains a short documented block (new or extended)
      for `skill-router.sh`'s opt-in trigger (`.loom/config/skill-routes.json`
      presence) and its one-line disable.
- [ ] New `defaults/hooks/tests/test-methodology-inject.sh` (mirroring the
      structure of `test-skill-router.sh`) covering the opt-in gate, the
      default once-per-session dedup, the `"always"` back-compat opt-in, the
      missing-`session_id` degrade, and the "hook never exits non-zero /
      never emits invalid JSON" invariant.
- [ ] The installer-wiring bullet (making both hooks opt-in-by-default in a
      fresh install's `.claude/settings.json`) is **out of scope for this
      issue** — it's a separate, larger behavior change to the installer
      itself (`scripts/install-loom.sh` / `install.sh`) with its own set of
      tradeoffs (existing installs vs. fresh installs), and the issue itself
      marks it explicitly lower-priority ("Consider making..."). File a
      follow-up if still wanted after this issue lands; do not block this PR
      on it.

### Affected Files

- `defaults/hooks/methodology-inject.sh` — primary change (dedup marker +
  config knob).
- `defaults/hooks/skill-router.sh` — reference only, not modified (pattern
  source).
- `defaults/CLAUDE.md` — documentation updates (Methodology Injection
  Framework section + skill-router opt-in/disable block).
- `defaults/hooks/tests/test-methodology-inject.sh` — new test file.
- `defaults/hooks/tests/test-skill-router.sh` — reference only, not modified
  (structural template for the new test file).

### Scope note

File-disjoint from the guard-destructive issues (#3755/#3756/#3757) — this
issue touches only the two `UserPromptSubmit` hooks and their documentation,
not `guard-destructive.sh` or `guard-loom-workflow.sh`.

rjwalters:
Closes #3758

## Summary

The `methodology-inject.sh` `UserPromptSubmit` hook injected `universal.md` on **every** prompt once a repo opted into `.loom/context/` — the main per-prompt token cost called out in the issue. This gates that injection to **once per session** by default, mirroring `skill-router.sh`'s #3609 session-marker dedup, behind a new `universal_frequency` config knob.

## Changes

- **`defaults/hooks/methodology-inject.sh`** (mirrored byte-identically to `.loom/hooks/`):
  - New `universal_frequency` key in `.loom/context/config.json` — `"session"` (default) or `"always"` (legacy per-prompt). Read the same defensive `jq` way as the other keys; missing/malformed falls back to `"session"` without raising the ERR trap.
  - Session-keyed marker at `.loom/logs/methodology-inject-seen/<sanitized-session-id>` (own namespace, parallel to `skill-router-seen/`), same sanitization + best-effort creation. Missing/empty `session_id` degrades gracefully to per-turn injection.
  - Role (`inject_role`) and topic (`inject_topics`) injection are **unchanged** — still fire every matching turn.
- **`defaults/CLAUDE.md`**: new `universal_frequency` config-table row + behavior-change callout, updated How-It-Works, and a new "UserPromptSubmit Hooks: Opt-In Triggers and Disabling" block documenting **both** hooks' opt-in trigger and one-line disable.
- **`defaults/hooks/tests/test-methodology-inject.sh`** (new, 20 cases, modeled on `test-skill-router.sh`): opt-in gate, once-per-session default, `"always"` back-compat, missing-`session_id` degrade, role/topic-unchanged guards, malformed config, and the never-non-zero / never-invalid-JSON invariant.

## Behavior change (load-bearing)

`universal_frequency` defaults to `"session"` — a deliberate flip from always-inject. A repo already using `.loom/context/` **without** setting this key now gets `universal.md` once per session instead of every prompt. This follows the #3609 precedent (`skill-router.sh` dropped per-prompt injection with no back-compat shim). Set `"universal_frequency": "always"` to restore the old behavior. Documented in `defaults/CLAUDE.md` and above.

## Out of scope

The installer default-wiring bullet (making both hooks opt-in-by-default in a fresh install's `.claude/settings.json`) is explicitly out of scope per the issue's acceptance criteria — a separate, larger installer change. File a follow-up if still wanted.

## Test / verification

- `./defaults/hooks/tests/test-methodology-inject.sh` → **20/20 passed** (default injects once, `always` preserves per-prompt).
- `shellcheck` clean on both the hook and the new test.
- `.loom/hooks/` mirror kept byte-identical (guarded by a sync assertion in the test).

### DIFF
--- a/.loom/hooks/methodology-inject.sh
+++ b/.loom/hooks/methodology-inject.sh
@@ -8,7 +8,9 @@
 #
 # Behavior:
 #   1. Check for .loom/context/ directory — exit silently if absent (opt-in)
-#   2. Always inject universal.md if it exists
+#   2. Inject universal.md if it exists, ONCE PER SESSION by default (mirrors
+#      skill-router.sh's #3609 per-session table dedup). The universal_frequency
+#      config knob ("session" default / "always" back-compat) controls this.
 #   3. Inject roles/<LOOM_ROLE>.md if LOOM_ROLE env var is set
 #   4. Inject topics/<name>.md when prompt matches filename or sidecar .pattern file
 #   5. Cap total output at configurable max (default 8000 chars)
@@ -52,6 +54,11 @@ fi
 # Extract prompt
 PROMPT=$(echo "$INPUT" | jq -r '.prompt // empty' 2>/dev/null) || PROMPT=""
 
+# Extract session_id (used for once-per-session universal.md dedup; optional — a
+# missing/empty value degrades gracefully, see the PER-SESSION UNIVERSAL DEDUP
+# block below)
+SESSION_ID=$(echo "$INPUT" | jq -r '.session_id // empty' 2>/dev/null) || SESSION_ID=""
+
 # If no prompt, nothing to do
 if [[ -z "$PROMPT" ]]; then
     exit 0
@@ -94,6 +101,9 @@ MAX_CONTEXT_CHARS=8000
 INJECT_UNIVERSAL=true
 INJECT_ROLE=true
 INJECT_TOPICS=true
+# Frequency of universal.md injection: "session" (once per session, new default
+# as of #3758) or "always" (every matching prompt, legacy back-compat behavior).
+UNIVERSAL_FREQUENCY=session
 
 # Read config if it exists
 if [[ -f "$CONFIG_FILE" ]] && jq empty "$CONFIG_FILE" 2>/dev/null; then
@@ -108,6 +118,9 @@ if [[ -f "$CONFIG_FILE" ]] && jq empty "$CONFIG_FILE" 2>/dev/null; then
     INJECT_UNIVERSAL=$(jq -r 'if .inject_universal == false then "false" else "true" end' "$CONFIG_FILE" 2>/dev/null) || INJECT_UNIVERSAL=true
     INJECT_ROLE=$(jq -r 'if .inject_role == false then "false" else "true" end' "$CONFIG_FILE" 2>/dev/null) || INJECT_ROLE=true
     INJECT_TOPICS=$(jq -r 'if .inject_topics == false then "false" else "true" end' "$CONFIG_FILE" 2>/dev/null) || INJECT_TOPICS=true
+    # Only "always" opts back into per-prompt injection; anything else (including
+    # a missing key or malformed value) falls through to the "session" default.
+    UNIVERSAL_FREQUENCY=$(jq -r 'if .universal_frequency == "always" then "always" else "session" end' "$CONFIG_FILE" 2>/dev/null) || UNIVERSAL_FREQUENCY=session
 fi
 
 # =============================================================================
@@ -148,9 +161,39 @@ append_context() {
 }
 
 # --- Universal context ---
+# =============================================================================
+# PER-SESSION UNIVERSAL DEDUP (#3758)
+# =============================================================================
+# universal.md is verbatim project-wide context that a session needs at most
+# once. By default (universal_frequency="session") we inject it on the first
+# matching prompt of a session and skip it thereafter, keyed on the session_id
+# present on stdin — exactly mirroring skill-router.sh's #3609 table dedup, but
+# in its own marker namespace (the two hooks are independent opt-ins and must
+# not share state). universal_frequency="always" restores the legacy per-prompt
+# behavior. A missing/empty session_id degrades gracefully: we cannot dedup, so
+# universal.md is included on each matching turn (the legacy behavior).
 if [[ "$INJECT_UNIVERSAL" == "true" ]] && [[ -f "${CONTEXT_DIR}/universal.md" ]]; then
-    UNIVERSAL_CONTENT=$(cat "${CONTEXT_DIR}/universal.md" 2>/dev/null) || UNIVERSAL_CONTENT=""
-    append_context "Project Context" "$UNIVERSAL_CONTENT"
+    INCLUDE_UNIVERSAL=1
+    if [[ "$UNIVERSAL_FREQUENCY" != "always" ]] && [[ -n "$SESSION_ID" ]]; then
+        # Sanitize to filename-safe characters so the marker is a single
+        # predictable file (never a path traversal, never a nested directory).
+        SESSION_KEY=$(printf '%s' "$SESSION_ID" | tr -c 'A-Za-z0-9._-' '_')
+        SEEN_DIR="${MAIN_ROOT}/.loom/logs/methodology-inject-seen"
+        SEEN_MARKER="${SEEN_DIR}/${SESSION_KEY}"
+        if [[ -f "$SEEN_MARKER" ]]; then
+            INCLUDE_UNIVERSAL=0
+        else
+            # Best-effort marker creation; a failure here never fails the hook
+            # and simply means universal.md may be re-injected on a later turn.
+            mkdir -p "$SEEN_DIR" 2>/dev/null || true
+            : > "$SEEN_MARKER" 2>/dev/null || true
+        fi
+    fi
+
+    if [[ "$INCLUDE_UNIVERSAL" -eq 1 ]]; then
+        UNIVERSAL_CONTENT=$(cat "${CONTEXT_DIR}/universal.md" 2>/dev/null) || UNIVERSAL_CONTENT=""
+        append_context "Project Context" "$UNIVERSAL_CONTENT"
+    fi
 fi
 
 # --- Role-specific context ---
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -1381,7 +1381,7 @@ Loom provides an opt-in methodology injection hook that automatically injects pr
    mkdir -p .loom/context/roles .loom/context/topics
    ```
 
-2. Add a universal context file (injected on every prompt):
+2. Add a universal context file (injected once per session by default):
    ```bash
    cat > .loom/context/universal.md << 'EOF'
    # Project Rules
@@ -1398,7 +1398,7 @@ Loom provides an opt-in methodology injection hook that automatically injects pr
 ```
 .loom/context/
 ├── config.json              # Optional configuration
-├── universal.md             # Injected on every prompt
+├── universal.md             # Injected once per session by default
 ├── roles/
 │   ├── builder.md           # Injected when LOOM_ROLE=builder
 │   ├── judge.md             # Injected when LOOM_ROLE=judge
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge Review — APPROVED

Thorough review of the once-per-session `universal.md` gating (#3758). All verification points confirmed:

**(a) Dedup correctness + namespace isolation** — The session-keyed marker lives at `.loom/logs/methodology-inject-seen/<sanitized-id>`, a distinct namespace from `skill-router.sh`'s `.loom/logs/skill-router-seen/`. The two hooks are independent opt-ins and correctly share no state. First-turn injects + writes marker; subsequent turns in the same session read the marker and skip; a fresh session_id re-injects. Verified by tests and manual runs.

**(b) Missing/empty session_id degrades safely** — The dedup branch is gated on `[[ -n "$SESSION_ID" ]]`, so a missing OR explicitly-empty `session_id` falls through to per-turn injection (legacy behavior) rather than crashing or silently suppressing. Confirmed directly: empty-string session_id yields exit 0, clean stderr, valid JSON, and injection on every turn.

**(c) Never non-zero / never invalid JSON** — The ERR trap (`exit 0`) is intact; all new marker I/O is best-effort (`mkdir -p … 2>/dev/null || true`, `: > "$MARKER" 2>/dev/null || true`); the new config read uses the same defensive `jq … || UNIVERSAL_FREQUENCY=session` fallback as sibling keys. Verified valid JSON + exit 0 under malformed config, empty session_id, and a path-traversal session_id (`../../etc/evil` → sanitized to a single file `.._.._etc_evil`, no filesystem escape, dedup still functions).

**(d) `always` back-compat** — `universal_frequency: "always"` bypasses the dedup block entirely and restores per-prompt injection; test confirms both turns inject.

**(e) Role/topic injection unchanged** — Diff touches only the universal block; tests confirm role and topic context still fire every matching turn even after universal is deduped for the session.

**(f) `.loom/` mirror byte-identical** — `diff -q defaults/hooks/methodology-inject.sh .loom/hooks/methodology-inject.sh` reports IDENTICAL, and the test suite's sync assertion guards it going forward.

**(g) Tests + lint** — `defaults/hooks/tests/test-methodology-inject.sh` → 20/20 passed. `shellcheck` clean on both the hook and the new test. CI: Installer Integration Tests pass; Rust/MCP jobs correctly skipped for a hook/doc-only change. Merge state CLEAN.

Docs in `defaults/CLAUDE.md` clearly flag the behavior change (default flip to `session`), the `always` opt-out, and the new "UserPromptSubmit Hooks: Opt-In Triggers and Disabling" reference table. Clean, well-scoped, well-tested. LGTM.

Issue #3757, opened 2026-07-22
guard-destructive: scope guard to irreversible operations; drop reversible GitHub state changes from defaults


rjwalters:
## Summary

The destructive-command guard's ask tier currently gates several **reversible**
GitHub operations. A guard whose purpose is preventing irreversible loss /
catastrophic agent error should not add confirmation friction to operations that
are trivially undone. Proposal: **scope guard aggressiveness to irreversibility.**

## Guiding principle

| Tier | Criterion | Matching policy | Examples |
|------|-----------|-----------------|----------|
| **Hard-deny** | catastrophic **and** irreversible | unanchored scan over the full raw string (false positives acceptable) | unscoped recursive-force `rm` against filesystem root, a shell fork bomb, curl-piped-to-shell, `gh repo delete`, protected-branch force-push |
| **Ask** | hard-to-reverse but legitimate | anchored to a command boundary; literal-text redacted | `gh release delete`, `git clean -fd`, `kubectl delete`, `sky down` |
| **Don't guard (default)** | reversible state change | not in defaults, or an opt-in category that is off by default | `gh issue close` -> `reopen`, `gh pr close` -> `reopen`, `gh label delete` -> recreate |

The paranoid, quote-blind, unanchored scanning is justified **only** where the
downside is catastrophic and unrecoverable. For reversible ops the cost of a
missed catch is near-zero while the false-positive friction is high — precision
should win.

## Concrete change requested

Remove these from the default `ASK_PATTERNS` (or move them to an opt-in
category, e.g. `guards.reversibleGh`, defaulting off):

- `gh pr close`   (undo: `gh pr reopen`)
- `gh issue close` (undo: `gh issue reopen`)
- `gh label delete` (undo: recreate; repos with `labels.yml` restore trivially)

Keep in the ask tier (genuinely harder to reverse):

- `gh release delete` (deletes published artifacts/tags)
- `git clean -fd`, `git checkout .`, `git restore .` (untracked / uncommitted loss)

## Rationale

- Autonomous agents that close their own issues/PRs as part of a normal
  lifecycle currently stall on a confirmation prompt (or, in headless runs with
  no approver, block entirely).
- Closing an issue/PR is a routine, auditable, one-command-to-undo action — it
  is not in the same risk class as deleting a repo or force-pushing to main.

## Technical context (current `main`, pre-#3756/#3757)

`defaults/hooks/guard-destructive.sh`:

[...]

Issue #3756, opened 2026-07-22
guard-destructive: ask-tier patterns are unanchored and skip literal-text redaction


rjwalters:
## Summary

The `ASK_PATTERNS` loop in `guard-destructive.sh` matches **bare, unanchored
substrings** against `COMMAND_NO_COMMENT` — the copy that has only shell
`#`-comments removed, **not** the literal-text-redacted `COMMAND_NO_LITERAL_TEXT`
that the catastrophic `ALWAYS_BLOCK` scan uses. As a result an ask-tier phrase
fires when it merely *appears* inside another command's quoted argument or a
flag value, even though nothing dangerous is being executed.

```bash
for pattern in "${ASK_PATTERNS[@]}"; do
    if echo "$COMMAND_NO_COMMENT" | grep -qE "$pattern"; then
        ask "Command requires confirmation: $COMMAND"
    fi
done
```

with e.g. `'gh issue close'`, `'gh pr close'`, `'gh release delete'` in the list.

## Reproducers (benign, but prompt for confirmation)

```bash
# A jq payload that only *mentions* the phrase:
jq -n '{cmd:"gh issue close 123"}'
# → ask "Command requires confirmation: ..."

# The phrase inside a comment/body value of an unrelated command:
gh issue close 5 --comment "restored the old gh issue close behavior"
# The --comment value is NOT redacted (strip_literal_text covers
# --body/-m/--title/--notes, but NOT --comment), so it also matches.
```

## Contrast: the catastrophic tier already got this right

`ALWAYS_BLOCK_PATTERNS` anchors its GitHub op to a command boundary:

```
'(^|[;&|[:space:]])gh repo delete'
```

and builds `COMMAND_NO_LITERAL_TEXT` via `strip_literal_text()` so a dangerous
phrase quoted inside `--body`/`-m`/`--title`/`--notes` no longer false-positives
(#3679). The **ask tier never received either fix** — it is still unanchored and
still reads the non-redacted copy.

## Root cause (exact locations, current `main`)

`defaults/hooks/guard-destructive.sh`:
- `ASK_PATTERNS` array: lines 1202-1244. Entries are bare substrings, no
  command-position anchor (contrast the `(^|[;&|[:space:]])` prefix used by
  `ALWAYS_BLOCK_PATTERNS` at lines 722-723).
- Ask loop: lines 1246-1250. Greps `COMMAND_NO_COMMENT` (built at lines
[...]

PR #3762, opened 2026-07-22
fix(guard): quote-aware command segmentation to stop phantom hard-denies (#3755)

### LINKED ISSUE #3755: guard-destructive: command segmentation splits on shell metacharacters inside quoted strings, causing false hard-denies
## Summary

`guard-destructive.sh`'s command-segment parsers split the command on shell
metacharacters (`; | & && ||`) **without honoring shell quoting**, so a
`|`-alternation *inside a quoted string argument* is treated as a real pipe.
The token following the quoted `|` becomes a phantom "command word" and the
command is **hard-denied** (a `deny`, not an `ask`).

This affects `lifecycle_or_cloud_reason()`, `extract_rm_targets()`, and the
`az`/`gcloud delete` check — all of which do:

```awk
gsub(/&&|\|\||[;|&]/, "\n")   # splits on every | — even inside "quotes"
...
cmd = toks[1]
if (cmd == "halt" || cmd == "reboot" || cmd == "poweroff" || cmd == "shutdown") { deny }
```

## Reproducer (all read-only, all hard-denied)

```bash
grep -E "poweroff|halt" somefile
# → BLOCKED: system lifecycle command: halt

grep -E "lifecycle|halt|poweroff|init 0" file
# → BLOCKED: system lifecycle command: halt   (and init 0)
```

A completely read-only `grep` whose only sin is a `|`-alternation regex
containing one of these words is blocked with no workaround (it's a `deny`).
The same class hits `awk`, `sed`, `jq`, `rg`, and SQL — any command with a
`|`-alternation in a quoted argument. During one session I could not `grep`
the hook's own source three separate times because the search terms tripped it.

The header comment above the check (currently) claims it "denies only when a
segment's *command word* is exactly the lifecycle word (#3584)" — but that is
only true for **unquoted** input. The awk has no lexer, so a quoted `"a|halt"`
still yields a segment whose first token is `halt`.

## Root cause

`.loom/hooks/guard-destructive.sh`:
- `lifecycle_or_cloud_reason()` (~line 848): `gsub(/&&|\|\||[;|&]/, "\n")` on the raw string.
- `extract_rm_targets()` (~line 944): same unquoted split.
- Both operate on `COMMAND_NO_COMMENT` (only `#`-comments stripped), **not** the
  literal-text-redacted `COMMAND_NO_LITERAL_TEXT` that the catastrophic
  `ALWAYS_BLOCK` scan already builds via `strip_literal_text()` for exactly this
  reason (#3679).

## Suggested fix

- Run the segment parsers over a quote-redacted copy: reuse / extend
  `strip_literal_text()` so quoted spans (single **and** double quoted) that
  carry no command-substitution (`$(` / backtick) are redacted before the awk
  split — mirroring how the catastrophic scan already protects `--body`/`-m`
  values. A quoted `|` then can't manufacture a phantom command word.
- Alternatively, do a minimal quote-aware pass in the awk (track `'`/`"` state
  while splitting) so separators inside quotes are ignored.
- Keep the genuine protection intact: `sudo halt`, `env FOO=bar halt`, and a real
  `foo; halt` must still deny.

## Acceptance criteria

- [ ] `grep -E "poweroff|halt" file` is **allowed** (no deny, no ask).
- [ ] `jq '.a|.b'`, `awk -F'|'`, `sed 's/a|b/x/'`, and SQL with `|` in a string are allowed.
- [ ] `sudo halt`, `poweroff`, `foo && reboot`, `env X=1 shutdown` still deny.
- [ ] The same quote-awareness is applied to `extract_rm_targets()` and the az/gcloud delete parser.
- [ ] A regression test covers a quoted `|`-alternation containing a lifecycle word.

## Affected Files

- `defaults/hooks/guard-destructive.sh` - Make `lifecycle_or_cloud_reason()` (~line 848), `extract_rm_targets()` (~line 944), and `parse_force_ops()`'s segment splitter (~line 553, shares the same `gsub(/&&||@PIPE[;|&]/, "\n")` pattern -- verify whether it needs the same fix or is out of scope for this issue since it only looks for `git`/`push`/`reset` command words, which are less prone to prose collision) quote-aware. Likely approach: extend `strip_literal_text()` (~line 637, already quote/backtick-aware) to redact quoted spans generically and feed its output into these three awk splitters, or add an inline quote-tracking state to the `gsub` call in each function.
- `tests/hooks/test-guard-destructive.sh` - Add regression tests (assert_allow) for the reproducer cases; existing lifecycle/rm test blocks are around lines 381-391 and 1200-1344 for reference. Also confirm the existing DENY regressions (e.g. `assert_deny "Regression: halt after && still denied" "sync && halt"` at line 1341, `env FOO=bar halt` at 1246-1249) still pass unchanged.

## Curator verification (2026-07-22)

Reproduced against current `main` by piping the JSON hook input directly into `defaults/hooks/guard-destructive.sh`:

- `grep -E "lifecycle|halt|poweroff|init 0" file` -> **confirmed hard-DENY** (`BLOCKED: system lifecycle command: halt`) even though the command is fully read-only. Root cause confirmed: `lifecycle_or_cloud_reason()`'s separator-gsub splits on the unquoted pipe between `halt` and `poweroff`, producing a segment whose first token is the bare word `halt`.
- `grep -E "poweroff|halt" somefile` did **not** reproduce a deny in my test (the trailing quote character glues onto the word, which fails the exact command-word-equality check) -- this is incidental, not a fix; any quote-aware rewrite must not rely on this accidental escape. Use the 4-way alternation form above (or any alternation where the target word is NOT adjacent to the closing quote) as the reliable regression case.
- Confirmed the read-only fast path (issue #3687) does **not** shortcut this: any command containing a literal pipe character (quoted or not) fails the fast-path structural check and falls through to the full deny/ask path, so the bug is not masked by the fast path.

Both files under Affected Files verified present on `origin/main` (fetched and checked via `git ls-tree`).

## Scope note (serialized with #3756, #3757)

This is the first of three issues rewriting `guard-destructive.sh` in sequence (#3755 -> #3756 -> #3757). Keep this PR scoped to the quote-aware segmentation fix only:
- **In scope**: quote-aware splitting for `lifecycle_or_cloud_reason()`, `extract_rm_targets()`, and the az/gcloud delete check; regression tests for the reproducer.
- **Out of scope** (land in later PRs on top of this one): ask-tier pattern anchoring / literal-text redaction (#3756), and the irreversibility-tiering redesign of the guard (#3757).

rjwalters:
Closes #3755

## Problem

`guard-destructive.sh`'s segment parsers split a command on shell metacharacters
(`; | & && ||`) **without honoring shell quoting**, so a `|`-alternation *inside
a quoted argument* was treated as a real pipe. The token following the quoted
`|` became a phantom "command word" and the command was **hard-denied** (a
`deny`, not an `ask`). A completely read-only `grep -E "lifecycle|<word>|poweroff|init 0" file`
was blocked as a "system lifecycle command" with no workaround.

## Fix

Adds a shared `qsplit()` awk helper (a single `_QSPLIT_AWK` source string so the
three parsers cannot drift). It walks the command tracking single-/double-quote
state and emits a segment break for a separator **only when it is outside a
quoted span**. Key safety property: a quoted span is treated as inert **only
when it carries no command substitution** (no `$(`, no backtick) -- mirroring
`strip_literal_text()`'s #3679 safety floor. Token values are preserved verbatim
(no redaction), so `extract_rm_targets()` still sees the real `rm` targets.

Applied to all three splitters that shared the naive `gsub(/&&|\|\||[;|&]/, "\n")`:

- `lifecycle_or_cloud_reason()` (includes the az/gcloud `delete` check)
- `extract_rm_targets()`
- `parse_force_ops()`

## Scope

This is the **root of the 3-PR stack** (#3755 → #3756 → #3757, all rewriting
`guard-destructive.sh`). Kept tight to quote-aware segmentation only:

- **In scope**: quote-aware splitting + regression tests for the reproducer.
- **Out of scope** (later PRs on top of this): ask-tier pattern anchoring /
  literal-text redaction (#3756), and the irreversibility-tiering redesign (#3757).

## Verification

Acceptance criteria — verified via the guard directly and as regression tests in
`tests/hooks/test-guard-destructive.sh`:

- Read-only quoted alternations now **ALLOW**: the 4-way lifecycle alternation,
  `jq '.a|.b'`, `awk -F'|'`, `sed 's/a|b/x/'`, and a quoted az/gcloud alternation.
- Genuine protections still **DENY**: real-separator lifecycle commands, the
  `env FOO=bar` prelude form, `az group delete`, and the command-substitution
  smuggle case `"$(x|<word> )"` (proven identical to `main`).
- Full suite: **349 passed / 0 failed** (335 pre-existing + 14 new #3755 cases).
- Behavior diffed against unmodified `main` on the command-substitution and
  quoted-root edge cases — **byte-identical outcomes**.

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -527,6 +527,82 @@ resolve_default_branch() {
     return 0
 }
 
+# =============================================================================
+# QUOTE-AWARE COMMAND SEGMENTATION (#3755)
+#
+# The three segment parsers below (parse_force_ops, lifecycle_or_cloud_reason,
+# extract_rm_targets) split a command string on the shell separators ; | & && ||
+# to find each simple command's command word. The historical split was a naive
+#   gsub(/&&|\|\||[;|&]/, "\n")
+# over the raw string, which has NO lexer — so a `|`-alternation INSIDE a quoted
+# argument (e.g. `grep -E "lifecycle|halt|poweroff"`) was split as if it were a
+# real pipe, manufacturing a phantom segment whose command word is the bare word
+# `halt` and hard-denying a completely read-only command.
+#
+# `qsplit()` replaces that gsub: it walks the string tracking single-/double-quote
+# state and emits a newline for a separator ONLY when it is OUTSIDE a quoted span.
+# A quoted span is treated as inert (its separators are preserved as literal
+# text) ONLY when it carries no command substitution — no `$(` and no backtick —
+# mirroring strip_literal_text()'s #3679 safety floor: a smuggled
+# `"$(a|halt)"` keeps its separators ACTIVE so the genuine protection is intact.
+# The token VALUES are preserved verbatim (unlike a redaction approach), so
+# extract_rm_targets still sees the real `rm` targets. Best-effort like
+# strip_literal_text(): backslash-escaped quotes and an unterminated quote fall
+# back to the old separator-active behaviour, never widening a deny into an allow.
+#
+# Shared as a single awk source string so the three parsers cannot drift.
+# =============================================================================
+_QSPLIT_AWK='
+function qsplit(s,   out, n, i, c, j, qc, ci, inner, SQ, DQ) {
+    SQ = sprintf("%c", 39)   # single quote
+    DQ = sprintf("%c", 34)   # double quote
+    out = ""
+    n = length(s)
+    i = 1
+    while (i <= n) {
+        c = substr(s, i, 1)
+        if (c == DQ || c == SQ) {
+            qc = c
+            ci = 0
+            for (j = i + 1; j <= n; j++) {
+                if (substr(s, j, 1) == qc) { ci = j; break }
+            }
+            if (ci == 0) {
+                # Unterminated quote: fall back to separator-active processing so
+                # a stray quote never suppresses a real split (never widen a deny).
+                out = out c
+                i++
+                continue
+            }
+            inner = substr(s, i + 1, ci - i - 1)
+            if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                # Inert quoted span: copy verbatim, separators inside are literal.
+                out = out substr(s, i, ci - i + 1)
+                i = ci + 1
+                continue
+            }
+            # Span carries command substitution: keep separators ACTIVE (copy the
+            # opening quote and keep walking char-by-char so a `|` inside splits).
+            out = out c
+            i++
+            continue
+        }
+        if (c == ";") { out = out "\n"; i++; continue }
+        if (c == "&") {
+            if (i < n && substr(s, i + 1, 1) == "&") { out = out "\n"; i += 2; continue }
+            out = out "\n"; i++; continue
+        }
+        if (c == "|") {
+            if (i < n && substr(s, i + 1, 1) == "|") { out = out "\n"; i += 2; continue }
+            out = out "\n"; i++; continue
+        }
+        out = out c
+        i++
+    }
+    return out
+}
+'
+
 # Parse force-op segments out of a command, emitting one TAB-separated
 # "<cpath>\t<target>" line per genuine git force-push / hard-reset. Portable awk
 # only (mirrors extract_rm_targets / lifecycle_or_cloud_reason segment parsing):
@@ -546,11 +622,11 @@ resolve_default_branch() {
 #   - reset --hard: always emitted with <target> = "@HEAD@".
 # The caller resolves "@HEAD@" to the checked-out branch and applies the mode.
 parse_force_ops() {
-    printf '%s' "$1" | awk '
+    printf '%s' "$1" | awk "$_QSPLIT_AWK"'
     BEGIN { SEP = sprintf("%c", 31) }  # US (unit separator) — non-whitespace so
                                        # bash read does not trim an empty cpath.
     {
-        gsub(/&&|\|\||[;|&]/, "\n")
+        $0 = qsplit($0)   # quote-aware segmentation (#3755)
         n = split($0, segs, "\n")
         for (i = 1; i <= n; i++) {
             seg = segs[i]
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVED

Rigorous security-adjacent review of the quote-aware command segmentation fix. This is a clean, surgical, **non-weakening** change.

### A/B verified against `main` (only ONE behavioral divergence, the intended fix)

| Case | main | PR | Note |
|------|------|-----|------|
| `grep -E "lifecycle\|halt\|poweroff\|init 0" file` (read-only) | deny | **allow** | the bug fix (phantom hard-deny removed) |
| `grep -E "$(x\|halt )" file` (cmdsub-pipe smuggle) | deny | deny | identical, smuggle still caught |
| quoted `rm -rf /` (dq and sq) | deny | deny | catastrophic scan untouched |
| `foo \| rm -rf /` (real pipe) | deny | deny | rm-target extraction intact |
| `x && halt`, `foo \| halt` (real seps) | deny | deny | genuine lifecycle protection intact |
| `echo "$(halt)"` | allow | allow | identical (command word is `echo`) |

Every safety-relevant case is **byte-identical to main**; the only change is the intended read-only allow.

### Safety properties confirmed
- **(a)** A quoted span is inert **only** when it carries no `$(`/backtick. A cmdsub-with-separator keeps separators ACTIVE, so the smuggle case still segments and denies (mirrors `strip_literal_text`'s #3679 floor).
- **(b)** The catastrophic `ALWAYS_BLOCK` scan is a separate `strip_literal_text`+grep path (lines 866-877), **not** routed through `qsplit`, so quoted `rm -rf /` still hard-denies.
- **(c)** Token values preserved verbatim (no redaction), so `extract_rm_targets` still sees real targets.
- **(d)** All three parsers (`parse_force_ops`, `lifecycle_or_cloud_reason`, `extract_rm_targets`) consume the single shared `_QSPLIT_AWK` source string, so no drift is possible.
- **(e)** Non-quoted commands behave identically to main.
- **(f)** 349 tests pass (335 pre-existing + 14 new), including the reproducers AND the cmdsub / backtick / real-rm counter-cases.

Adversarial reasoning holds: the only in-quote execution vector is command substitution, which `qsplit` deliberately keeps active; literal quoted text is inert to the shell too, so suppressing phantom splits there cannot leak a dangerous execution. Best-effort fallbacks (unterminated/escaped quote) revert to separator-active, never widening a deny into an allow. Portable POSIX awk throughout. CI green, mergeStateStatus CLEAN.

Root of the 3-PR stack (#3755 -> #3756 -> #3757); scope correctly limited to segmentation. LGTM.

PR #3773, opened 2026-07-22
fix(guard): anchor ask-tier patterns + redact literal text (#3756)

### LINKED ISSUE #3756: guard-destructive: ask-tier patterns are unanchored and skip literal-text redaction

rjwalters:
## Summary

The `ASK_PATTERNS` loop in `guard-destructive.sh` matched **bare, unanchored
substrings** against `COMMAND_NO_COMMENT` — a copy that is only shell-`#`-comment
stripped, **never** literal-text redacted. An ask-phrase (`gh issue close`,
`gh pr close`, `gh release delete`, `systemctl restart`, …) fired a spurious
confirmation prompt merely by appearing inside another command's quoted argument
or a text-carrying flag value, even though nothing dangerous ran.

This ports the two fixes the catastrophic `ALWAYS_BLOCK` tier already has to the
ask tier (issue 2 of 3 in the serialized `#3755`→`#3756`→`#3757` sequence; built
on top of `#3755`'s qsplit fix now on main).

## Changes

**(a) Command-position anchoring (mirrors `#3553`)** — every `ASK_PATTERNS`
entry is prefixed with `(^|[;&|[:space:]])`, anchored at the phrase's **leading
command word** (multi-word phrases like `kubectl rollout restart` / `git
checkout \.` anchor at the first token, per the `gh repo delete` precedent). A
quote-adjacent mention such as `jq -n '{cmd:"gh issue close 123"}'` (phrase
preceded by `"`) no longer false-asks.

**(b) Literal-text redaction (mirrors `#3679`)** — the ask loop now reads a new
comment-stripped **AND** flag-value-redacted copy `COMMAND_ASK_SCAN` (built from
`COMMAND_NO_COMMENT`, so the comment-stripping the ask tier already relied on is
preserved). `strip_literal_text()` and its outer presence guard are extended to
cover `--comment` (previously `--body`/`-m`/`--title`/`--notes` only), so an
ask-phrase quoted inside a `--comment`/`--body` value of an *unrelated* command
no longer false-asks. The `--comment` addition also flows to the catastrophic
tier's `COMMAND_NO_LITERAL_TEXT` (consistent with the existing `--body` handling,
harmless).

## Scope boundary

Limited to ask-tier anchoring + redaction. The `#3755` quote-aware `qsplit`
helper is **untouched**; ask-tier membership / reversibility redesign is `#3757`
(which stacks on this branch). The catastrophic scan still reads the raw
command, so this can only **narrow** an ask, never miss a hard deny.

**Known limitation (shared with the catastrophic tier):** the `[[:space:]]`
anchor cannot distinguish a real separator from whitespace *inside* a quoted
string, so a mid-quote prose mention like `echo "… gh pr close …"` (space before
the phrase) still matches on its leading space. Robustly suppressing that needs
command-word segment classification, which is out of scope for
anchoring+redaction (`#3757` territory). The quote-*adjacent* `jq` case IS fixed.

## Tests

Added a `#3756` regression block in `tests/hooks/test-guard-destructive.sh`:
- ask-phrase in a quoted `jq` payload → ALLOW
- ask-phrase in a redacted `--body` value (no real ask cmd) → ALLOW
- ask-phrase in a redacted `--comment` value (no real ask cmd) → ALLOW
- genuine leading `gh issue close … --comment "…gh issue close…"` → still ASK
  (redaction suppresses the redundant second match; the real leading command
  legitimately still asks)
- chained `git status && gh issue close 5` → still ASK (separator anchor)

Full suite passes: **354/354**. `bash -n` + `shellcheck -S error` clean.

Closes #3756

### DIFF
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -691,12 +691,16 @@ parse_force_ops() {
 }
 
 # Redact the quoted VALUES of known text-carrying flags (--body, -m/--message,
-# --title, --notes) so a dangerous-looking phrase quoted INSIDE such a value no
-# longer trips the raw ALWAYS_BLOCK_PATTERNS substring scan. Used ONLY to build
-# COMMAND_NO_LITERAL_TEXT for that one loop (mirrors the COMMAND_NO_COMMENT
-# precedent); every other scan keeps reading the raw command. This kills the
-# #3679 false positive where `gh pr comment --body "…git push --force origin
-# main…"` / `git commit -m "…"` hard-denied even though nothing executes.
+# --title, --notes, --comment) so a dangerous-looking phrase quoted INSIDE such a
+# value no longer trips the raw ALWAYS_BLOCK_PATTERNS substring scan (catastrophic
+# tier) or the ASK_PATTERNS scan (ask tier, #3756). Used ONLY to build the
+# literal-redacted working copies for those two loops (mirrors the
+# COMMAND_NO_COMMENT precedent); every other scan keeps reading the raw command.
+# This kills the #3679 false positive where `gh pr comment --body "…git push
+# --force origin main…"` / `git commit -m "…"` hard-denied even though nothing
+# executes, and (#3756) the analogous ask-tier false ask where an ask-phrase like
+# `gh issue close` quoted inside a `--comment`/`--body` value prompted for
+# confirmation despite no such command actually being run.
 #
 # Safety floor preserved two ways:
 #   - `-c` is deliberately NOT a text-carrying flag, so `bash -c '<payload>'`
@@ -716,7 +720,7 @@ strip_literal_text() {
         SQ = sprintf("%c", 39)   # single quote
         DQ = sprintf("%c", 34)   # double quote
         # boundary + text-carrying flag + optional (ws / = / ws) + quoted span
-        re = "(^|[ \t])(--message|--body|--notes|--title|-m)[ \t]*=?[ \t]*(" \
+        re = "(^|[ \t])(--message|--body|--notes|--title|--comment|-m)[ \t]*=?[ \t]*(" \
              DQ "[^" DQ "]*" DQ "|" SQ "[^" SQ "]*" SQ ")"
     }
     {
@@ -857,16 +861,17 @@ ALWAYS_BLOCK_PATTERNS=(
 )
 
 # Build a literal-text-redacted working copy ONLY for the catastrophic scan
-# below, so a force-push-to-main phrase quoted inside a --body/-m/--title/--notes
-# value no longer false-positives (#3679). The awk only runs when one of those
-# flags is actually present, keeping it off the hot path (mirrors the
-# COMMAND_NO_COMMENT `#`-present guard). `-c` is intentionally excluded so
-# `bash -c '<payload>'` payloads still reach the raw scan; spans carrying `$(` /
-# backtick are left intact so command-substitution smuggling still hard-denies.
+# below, so a force-push-to-main phrase quoted inside a
+# --body/-m/--title/--notes/--comment value no longer false-positives (#3679,
+# --comment added #3756). The awk only runs when one of those flags is actually
+# present, keeping it off the hot path (mirrors the COMMAND_NO_COMMENT
+# `#`-present guard). `-c` is intentionally excluded so `bash -c '<payload>'`
+# payloads still reach the raw scan; spans carrying `$(` / backtick are left
+# intact so command-substitution smuggling still hard-denies.
 COMMAND_NO_LITERAL_TEXT="$COMMAND"
 if [[ "$COMMAND" == *"--body"* || "$COMMAND" == *"--message"* || \
       "$COMMAND" == *"--title"* || "$COMMAND" == *"--notes"* || \
-      "$COMMAND" == *"-m"* ]]; then
+      "$COMMAND" == *"--comment"* || "$COMMAND" == *"-m"* ]]; then
     COMMAND_NO_LITERAL_TEXT=$(strip_literal_text "$COMMAND")
 fi
 
@@ -894,6 +899,27 @@ if [[ "$COMMAND" == *"#"* ]]; then
     COMMAND_NO_COMMENT=$(printf '%s\n' "$COMMAND" | sed -E 's/(^|[[:space:]])#.*$//')
 else
     COMMAND_NO_COMMENT="$COMMAND"
+fi
+
+# =============================================================================
+# ASK-TIER WORKING COPY (#3756) — comment-stripped AND literal-text redacted.
+#
+# The ASK_PATTERNS loop below needs BOTH narrowings the catastrophic tier's two
+# copies provide separately: COMMAND_NO_COMMENT's `#`-comment stripping AND
+# strip_literal_text()'s quoted-flag-value redaction (the #3679 fix the ask tier
+# never received). Building the ask copy from COMMAND_NO_COMMENT (not raw
+# $COMMAND) preserves the comment-stripping the ask tier already relied on, then
+# redacts --body/-m/--title/--notes/--comment values so an ask-phrase quoted
+# inside such a value (e.g. `gh pr comment --body "…gh issue close…"`) no longer
+# false-asks. The strip only runs when a text-carrying flag is present, keeping
+# it off the hot path. Never feeds the catastrophic scan (that keeps reading the
+# raw command), so this can only NARROW an ask, never miss a hard deny.
+# =============================================================================
+COMMAND_ASK_SCAN="$COMMAND_NO_COMMENT"
+if [[ "$COMMAND_NO_COMMENT" == *"--body"* || "$COMMAND_NO_COMMENT" == *"--message"* || \
+      "$COMMAND_NO_COMMENT" == *"--title"* || "$COMMAND_NO_COMMENT" == *"--notes"* || \
+      "$COMMAND_NO_COMMENT" == *"--comment"* || "$COMMAND_NO_COMMENT" == *"-m"* ]]; then
+    COMMAND_ASK_SCAN=$(strip_literal_text "$COMMAND_NO_COMMENT")
 fi
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed the middle guard-trio PR (#3756) — ask-tier command-position anchoring + literal-text redaction ported from the catastrophic tier. Verified rigorously since this is a security-adjacent guard.

**Anchoring (a) — correct.** Every `ASK_PATTERNS` entry now carries `(^|[;&|[:space:]])`, byte-identical to `ALWAYS_BLOCK`'s `gh repo delete` anchor (#3553), multi-word phrases anchored at the leading token. Direct hook probes confirm every genuine ask still fires: bare `gh issue close 5`, `gh release delete v1.0`, `git clean -fd`, `sudo systemctl stop nginx`, and separator-chained `git status && gh issue close 5` all → ASK. Only quote-adjacent mentions (`jq -n '{cmd:\"gh issue close 123\"}'`) stop firing.

**Redaction (b) — correct + safety floor intact.** The ask loop reads `COMMAND_ASK_SCAN` (built from `COMMAND_NO_COMMENT` so comment-stripping is preserved, then `strip_literal_text()` with `--comment` added). Confirmed via direct hook probes:
- `gh issue reopen 5 --comment "...gh issue close 123..."` → ALLOW (no real ask command).
- Genuine leading `gh issue close 5 --comment "...gh issue close..."` → still ASK (redaction only suppresses the redundant *second* match; the real leading command legitimately asks).
- **Hard denies never missed:** `COMMAND_ASK_SCAN` only feeds the ask loop; the catastrophic scan is untouched. `rm -rf /`, chained `... --comment "note" && rm -rf /`, and command-substitution smuggling `--comment "$(rm -rf /)"` all still DENY — the awk leaves `$(`/backtick spans intact (line 743), so redaction can only narrow an ask, never a deny.

**Scope (c) — clean.** The #3755 qsplit helper and ask-tier membership (#3757's territory) are untouched.

**Documented discrepancies (d) — correct semantics, not weakening.** The leading-command-still-asks behavior and the mid-quote-prose `[[:space:]]`-anchor limitation are both authoritative-semantics-correct and documented in code + PR body.

**Tests (e):** 5 new #3756 regressions + full suite **354/354 pass**; `shellcheck`/`bash -n` clean; CI green; merge state CLEAN.

**Adversarial note (non-blocking, out of scope):** subshell-wrapped ask commands `(gh issue close 5)` slip to ALLOW because the anchor omits `(`. This exactly mirrors the *pre-existing* catastrophic-tier gap — `(gh repo delete ...)` likewise slips its hard deny on main today, same anchor. This PR faithfully mirrors the established precedent rather than introducing a new gap; segment-word classification (which would close both) is explicitly #3757's scope. Worth a follow-up to add `(` to both tiers' anchor, but not a regression here.

LGTM. Approving.

Issue #3772, opened 2026-07-22
Broaden read-only fast-path defaults for common read-only verbs (follow-up to #3687)


rjwalters:
## Context

Follow-up to the (closed, shipped) #3687 read-only fast-path. #3687 short-circuits obviously-read-only commands to a silent `allow`, but its built-in allowlist is deliberately narrow: `git status|log|diff|show`, `ls`, `grep`, `rg`, `gh <noun> view|list`, `aws <svc> describe*|get*|list*`, `aws s3 ls`. Very common read-only verbs are **excluded** and therefore take the full ~179ms deny/ask gauntlet on every call: `cat` (has a deliberate `.ssh`/credentials carve-out), `jq`, `sed`, `awk`, `wc`, `head`, `tail`, `find`, `test`/`[[`.

In a subagent-dense `/sweep` (many builders/judges/curators, each running dozens of `jq`/`grep`/`sed` calls), that per-call cost compounds across the fleet. `guards.readOnlyFastPathExtra` exists as an operator escape hatch but is **off by default and undocumented** for this purpose, so operators don't reach for it.

## Suggested acceptance criteria

- [ ] Evaluate adding argument-independent read-only verbs to the fast-path **default** allowlist where safe — strong candidates: `jq` (no `-f`/in-place concerns for read), `wc`, `head`, `tail`, `find` (read-only unless `-delete`/`-exec` — must exclude those), `test`/`[[`. Explicitly keep `cat`'s existing `.ssh`/`.aws/credentials` ASK carve-out and keep `ssh` excluded (opaque remote payload), per #3687.
- [ ] For any verb with a dangerous sub-form (`find -delete`/`-exec`, `sed -i`, `jq` writing files), either exclude the verb or make admission structural (first-token + no dangerous flag) so the fast path can't bypass a real mutation.
- [ ] Document `guards.readOnlyFastPathExtra` in CLAUDE.md with a recommended safe set and the one-line warning that each entry is a full-generality bypass for that command word.
- [ ] Measure the per-call and cumulative latency reduction on a representative subagent-heavy sweep.

## Relation
- #3687 — the fast-path this extends (closed). Same security discipline: admission is purely structural; a bypass is only added for bare, argument-independent read-only utilities.

rjwalters:
## Curator enhancement

Verified against current `main` (577f9dae, post-guard-trio #3755/#3756/#3757, both merged 2026-07-22). `git status` clean at `origin/main` == local `HEAD`.

[...]

PR #3796, opened 2026-07-22
feat(guard): add deny/ask decision telemetry log (#3771)

### LINKED ISSUE #3771: guard-destructive: add decision telemetry (log deny/ask events) so false-positive friction is measurable
## Context

Recurring operator observation (2026-07-22): "we are still hitting a lot of hooks." Today that is **unmeasurable** — `guard-destructive.sh` has an *error* log (`hook-errors.log`, for unexpected-error/ERR-trap events) but records **no deny/ask decision events**. So there is no way to know:

- how often the guard fires per session,
- which patterns fire most (and which are false positives vs. true catches),
- whether precision fixes (#3755, #3756, #3757 — all merged to main as of this curation pass) actually reduce the false-positive rate.

Pattern-tuning is currently blind. Several agents this session tripped the guard on *legitimate* work (writing reproducer text containing lifecycle words / `gh issue close` into commits/PR bodies), which is exactly the friction #3755/#3756/#3757 target — but we can't quantify the before/after without a decision log.

**Chain note**: this issue is **Chain A root** — #3772 stacks a follow-up (reader/aggregation tooling) on top of the file-format this issue defines. Keep the record schema (field names, one-JSON-object-per-line) stable since #3772 depends on it.

## Current state (`defaults/hooks/guard-destructive.sh`, verified against current `main`, commit `577f9dae`)

- `SCRIPT_DIR`/`HOOK_ERROR_LOG` are set at **lines 35-36**:
  ```bash
  SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd 2>/dev/null || echo ".")"
  HOOK_ERROR_LOG="${SCRIPT_DIR}/../logs/hook-errors.log"
  ```
  `log_hook_error()` (**lines 39-44**) is the best-effort append pattern to mirror — `mkdir -p "$(dirname "$LOG")" 2>/dev/null || true` then `echo ... >> "$LOG" 2>/dev/null || true`. Note `SCRIPT_DIR` is the *installed* hook's own directory (`.loom/hooks/` at runtime, not `defaults/hooks/`), so `${SCRIPT_DIR}/../logs/` resolves to `.loom/logs/` in a real install — matches the issue's suggested `.loom/logs/guard-decisions.log` path with zero extra resolution logic.
- `deny()` (**lines 807-824**) and `ask()` (**lines 826-843**) are the two exit points that must be instrumented. Both take a single `reason` string and unconditionally `exit 0` after emitting the JSON permission decision — any logging added inside them must stay before the `exit 0` and must never itself risk a non-zero exit (mirror the `2>/dev/null || true` discipline used everywhere else in this file).
- `strip_literal_text()` (**lines 771-805**) is the existing redaction helper (used today only for the catastrophic/ask working copies, `COMMAND_NO_LITERAL_TEXT` / `COMMAND_ASK_SCAN`) — reuse it (or its `awk` approach) to redact `--body`/`-m`/`--title`/`--notes`/`--comment` quoted values in the command string written to the decision log. **Do not log the raw `$COMMAND`** — that's the same class of leak `strip_literal_text()` already guards against for pattern matching, and the decision log will aggregate across sessions, raising the stakes further.
- The **#3687 read-only fast path** (**lines 68-258**) `exit 0`s silently *before* `REPO_ROOT` is resolved and before any deny/ask array runs — it is an `allow`, not a `deny`/`ask`, so it is correctly **out of scope** for this issue's logging (the issue's AC already says `allow` is not logged by default). Confirm any new toggle-resolution code you add for the decision log does not sit on that fast path's hot path — the existing toggle functions (`sql_guard_enabled`, `cloud_guard_enabled`, `rm_scope_repo_enabled`, `force_scope_mode`, `reversible_gh_guard_enabled`, `fastpath_enabled`) are ALL resolved *lazily*, only after a candidate match; the new decision-log toggle should follow the exact same lazy/cached pattern (see `_SQL_GUARD_CACHE` at **lines 290-310** for the canonical shape) so it never touches the ~99% of commands that neither deny nor ask.
- **~15 call sites** invoke `deny()`/`ask()` today (grep `deny "BLOCKED` / `ask "Command requires confirmation` in the file). None of them currently pass a machine-readable "matched pattern/rule" or "tier" — that information exists only inside the free-text `reason` string, and the loops that iterate `ALWAYS_BLOCK_PATTERNS` / `ASK_PATTERNS` / `CLOUD_ASK_PATTERNS` (**lines 932, 1418, 1523**) already have the matched `pattern` variable in scope at the call site. The non-loop call sites (lifecycle/cloud segment check, SQL DDL, rm-scope, force-op branch scope, read-tree, reversible-gh) each have their own static reason string that can serve as the rule tag.

## Suggested implementation approach

1. Extend `deny()` and `ask()` to accept an **optional second positional arg** — a short, stable "rule tag" string (e.g. `"catastrophic:rm-root"`, `"ask:git-clean-fd"`, `"sql-ddl"`, `"force-scope:protected"`, `"reversible-gh:pr-close"`, `"cloud-cli:aws-ec2-mutate"`, `"lifecycle:halt"`, `"read-tree"`). Default it to a fallback derived from the function name (`deny`/`ask`) if omitted, so this is backward-compatible and doesn't require touching every call site in the first pass (though touching them to pass real tags is what makes the log useful — see AC below).
2. Add a `log_guard_decision()` helper next to `log_hook_error()` (same file region, ~line 44) that:
   - is gated by a new lazy/cached toggle (`guards.decisionLog` / `LOOM_GUARD_DECISION_LOG`, mirroring `sql_guard_enabled()`'s shape) — **default OFF or lightweight-default-ON is the curator's open question for the builder to resolve against the AC below** (the issue text says "off or lightweight by default" — pick one and document the choice in the PR description),
   - writes to `DECISION_LOG="${SCRIPT_DIR}/../logs/guard-decisions.log"` (separate file from `HOOK_ERROR_LOG`),
   - appends **one JSON object per line** (JSONL, so #3772's reader can `jq -c` per line without a multi-line parser): `{"ts": "...", "decision": "deny"|"ask", "pattern": "<tag>", "tier": "catastrophic"|"ask", "command": "<redacted>"}`,
   - runs the same `mkdir -p ... || true` / `>> ... 2>/dev/null || true` best-effort discipline as `log_hook_error()` — a log-write failure must NEVER change the deny/ask decision or cause a non-zero exit.
3. Call `log_guard_decision` from inside `deny()`/`ask()` (single call site each) so every one of the ~15 existing call sites is covered automatically once they're updated to pass a tag — no need to duplicate logging logic at each site.
4. Redact `$COMMAND` before logging using `strip_literal_text()` (or equivalent) — same redaction the catastrophic/ask tiers already apply for matching, now applied for the log line too.

## Acceptance criteria

- [ ] `guard-destructive.sh` appends one JSONL record per **deny** and **ask** decision to `.loom/logs/guard-decisions.log` (or the `SCRIPT_DIR`-relative equivalent, mirroring `HOOK_ERROR_LOG`'s resolution), separate from `hook-errors.log`.
- [ ] Each record includes at minimum: `ts` (UTC timestamp, same `date -u '+%Y-%m-%dT%H:%M:%SZ'` format as `log_hook_error`), `decision` (`deny`/`ask`), `pattern` (a short stable tag identifying the matched rule — not the full free-text reason), `tier` (`catastrophic`/`ask`), and `command` (the command string, **redacted** via `strip_literal_text()` so no raw secret/literal-flag-value text is logged).
- [ ] `allow` decisions (including the #3687 fast-path silent allow) are **NOT** logged by default. If you add an opt-in verbose/debug mode that also logs allows, it must be a distinct toggle value, not the default-on state.
- [ ] Logging is best-effort: a log-write failure (permission denied, disk full, missing dir) never changes the guard's deny/ask decision and never causes the hook to exit non-zero (same fail-open discipline as `log_hook_error`'s `2>/dev/null || true` pattern).
- [ ] The decision-log toggle is resolved **lazily** (only once a deny/ask is about to fire, not on every invocation) and is configurable via both `.loom/config.json` → `guards.decisionLog` and a `LOOM_GUARD_DECISION_LOG` env var (env wins), following the exact precedence/caching shape of the existing toggles (e.g. `sql_guard_enabled()`, **lines 290-310**).
- [ ] The #3687 read-only fast path's silent-allow performance characteristic is unaffected — no new forks/config reads execute on the fast-path's hot path (verify by re-running or reasoning about the existing fast-path perf comments at **lines 68-122**).
- [ ] A short `jq`/`grep` one-liner (or a tiny script) that summarizes fires-by-`pattern` is documented — either inline as a comment in the guard file, in `defaults/CLAUDE.md`'s guard-hooks section, or in a new short doc — e.g. `jq -r '.pattern' .loom/logs/guard-decisions.log | sort | uniq -c | sort -rn`. (Full aggregation tooling is #3772's scope — this issue only needs the log to *exist* and be human-inspectable with a one-liner.)
- [ ] New/updated tests in `tests/hooks/test-guard-destructive.sh` cover: (a) a deny-triggering command writes a decision-log record with the expected `decision`/`tier` when the toggle is on; (b) an ask-triggering command likewise; (c) an allow-only command (e.g. `git status` outside the fast path, or any non-matching command) does NOT write a record; (d) the fast-path silent-allow (e.g. `git status`) does NOT write a record; (e) the log file location is separate from `hook-errors.log`; (f) a command containing a secret-bearing `--body`/`-m` value that triggers deny/ask logs a **redacted** command, not the raw value; (g) the toggle-off (or toggle-default, whichever this PR lands as) state produces no log growth.

## Test plan

Run the existing suite first — it must still pass with zero regressions:

```bash
./tests/hooks/test-guard-destructive.sh
```

Manual smoke tests (illustrative; adapt to the actual toggle default chosen):

- **Deny path**: pipe a catastrophic-pattern command (e.g. a filesystem-root deletion — see `ALWAYS_BLOCK_PATTERNS` for exact forms) through the guard with the decision-log toggle enabled, then `tail -1 .loom/logs/guard-decisions.log | jq .` and confirm `decision == "deny"`, `tier == "catastrophic"`, `pattern` is a short tag (not the full sentence), and `command` is present.
- **Ask path**: pipe `git clean -fd` through the guard with the toggle enabled, confirm a record with `decision == "ask"`.
- **Allow path (no record)**: pipe a non-matching command (e.g. `ls -la`) through the guard with the toggle enabled; confirm the decision-log line count is unchanged before/after.
- **Fast-path allow (no record)**: pipe `git status` through the guard; confirm no record is written and no perf regression (still a near-instant silent allow).
- **Redaction**: pipe a deny/ask-triggering command whose reason lives inside a `-m`/`--body` value with an embedded secret-looking token, confirm the secret token does not appear anywhere in `.loom/logs/guard-decisions.log`.
- **Fail-open**: point the log dir at a non-writable path (e.g. `chmod 000` the parent, or point `DECISION_LOG` somewhere unwritable via a test seam) and confirm the guard still emits the correct deny/ask JSON to stdout and still exits 0.

## Scope / non-goals

- **In scope**: the deny/ask decision log itself (schema, write path, toggle, redaction, tests) in `defaults/hooks/guard-destructive.sh`.
- **Out of scope** (deferred to #3772): a dedicated reader/aggregation CLI beyond the one-liner AC above; historical backfill/rotation of the log; dashboards or trend analysis; wiring the log into `agent-metrics.sh` or any MCP tool.
- **Out of scope**: `guard-loom-workflow.sh` (the separate Loom-workflow-specific hook) and the `.claude/skills/repo/hooks/guard-destructive.sh` fork (Repo Skills' own copy) — this issue only touches the canonical `defaults/hooks/guard-destructive.sh` source of truth (the `.loom/hooks/` copy is gitignored/installer-managed and will pick up the change automatically on next sync).

## Affected Files

- `defaults/hooks/guard-destructive.sh` (primary change)
- `tests/hooks/test-guard-destructive.sh` (new test cases)
- `defaults/CLAUDE.md` (optional — one-liner doc addition to the guard-hooks section, if not documented inline in the script instead)

## Relation

- #3755 / #3756 / #3757 — guard precision fixes whose impact this telemetry would let us measure (all merged to `main` prior to this curation pass).
- #3687 — read-only fast-path (allow-logging must not undermine its zero-overhead silent-allow; verified unaffected above).
- #3772 — Chain A follow-up: reader/aggregation tooling built on top of this issue's JSONL schema. Do not change field names without considering that dependency.

rjwalters:
## Summary

Makes guard-hook friction **measurable** (issue #3771 — Chain A ROOT). `guard-destructive.sh` now appends one JSONL record per **deny**/**ask** decision to a decision log (`.loom/logs/guard-decisions.log`, `SCRIPT_DIR`-relative) **separate** from `hook-errors.log`, so we can finally quantify which patterns fire, how often, and whether the precision fixes (#3755/#3756/#3757) cut the false-positive rate.

## Toggle default: OFF (documented decision)

The curator left "off or lightweight-default-ON" open. This lands **default OFF** (inverse polarity, mirroring `reversible_gh_guard_enabled()`): enabling writes a new persistent, cross-session artifact of redacted commands, so a zero-config install sees **no new file and no behaviour change** — the same convention as the other opt-in data-collection features (transcript archival #3726, model-cost experiment #3725). Enable via `guards.decisionLog:true` or `LOOM_GUARD_DECISION_LOG=1` (env wins).

## Stable JSONL schema (the contract #3772 stacks on)

```json
{"ts":"2026-07-22T23:17:13Z","decision":"deny","pattern":"sql-ddl","tier":"catastrophic","command":"<redacted>"}
```

`ts` (UTC, same format as `hook-errors.log`) · `decision` (deny/ask) · `pattern` (short stable rule tag, not the free-text reason) · `tier` (catastrophic/ask) · `command` (**redacted** via `strip_literal_text()` — no raw `--body`/`-m`/`--title`/`--notes`/`--comment` secret persisted). **Field names are load-bearing for #3772 — do not rename.**

## Implementation

- New `log_guard_decision()` helper next to `log_hook_error()`, gated by a lazy/cached `decision_log_enabled()` toggle following the exact shape of the existing toggles (config read only fires once a deny/ask is about to occur).
- `deny()`/`ask()` take an optional stable rule-tag 2nd arg (defaults to the function name → backward-compatible); all ~13 call sites now pass a short machine-readable tag. Logging is emitted **before** the JSON decision and wrapped `|| true` so it can never suppress a decision or trip the ERR trap.
- **`allow` is never logged** (including the #3687 fast-path silent allow) — the toggle resolves lazily so the fast path's zero-overhead hot path is byte-for-byte unaffected.
- **Best-effort / fail-open**: a log-write failure never changes the decision or causes a non-zero exit (grouped-redirect so an unwritable path leaks nothing to stderr). `LOOM_GUARD_DECISION_LOG_FILE` overrides the write path (test seam). `.loom/logs/` is already gitignored.

## Tests

10 new cases in `tests/hooks/test-guard-destructive.sh`: deny logs a record (decision/tier/ts/pattern/command); ask logs a record; allow-only **and** fast-path silent-allow log nothing; log is separate from `hook-errors.log`; secret `-m` value logged **redacted**; default-off no growth; config-toggle-on; env-over-config precedence; fail-open still denies + exits 0. **Full suite: 380 passed, 0 failed** (was 370 — zero regressions).

## Summarize fires (fuller reader/aggregation CLI is #3772's scope)

```bash
jq -r '.pattern' .loom/logs/guard-decisions.log | sort | uniq -c | sort -rn
```

## Scope

Touches only the canonical `defaults/hooks/guard-destructive.sh`, its tests, and `defaults/CLAUDE.md`. The `.claude/skills/repo/hooks/` fork and `guard-loom-workflow.sh` are untouched (out of scope per the issue).

**Chain A base for #3772** (which stacks reader/aggregation tooling on this JSONL schema).

Closes #3771

### DIFF
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -1372,6 +1372,58 @@ LOOM_GUARD_READONLY_FASTPATH=0 git status
 #   .loom/config.json  ->  { "guards": { "readOnlyFastPathExtra": ["jq"] } }
 ```
 
+### Decision Telemetry Log (`guards.decisionLog` / `LOOM_GUARD_DECISION_LOG`)
+
+`guard-destructive.sh` can record every **deny** and **ask** decision to a JSONL decision log (issue #3771), separate from `hook-errors.log`, so guard-hook friction becomes **measurable** — which patterns fire, how often, and whether a precision fix (#3755/#3756/#3757) actually cut the false-positive rate. Without it, "we keep hitting the hooks" is unquantifiable.
+
+The log is **off by default** — enabling it writes a new persistent, cross-session artifact, so like the other opt-in data-collection features (transcript archival #3726, the model-cost experiment #3725) a zero-config install sees no new file and no behaviour change. It is resolved in this order (highest precedence first):
+
+1. **`LOOM_GUARD_DECISION_LOG` env var** — `1`/`true`/`yes`/`on` enables; `0`/`false`/`no`/`off` disables. Overrides the config value.
+2. **`.loom/config.json`** — `guards.decisionLog` (default `false` when absent). Set it to `true` to enable:
+   ```json
+   {
+     "guards": {
+       "decisionLog": true
+     }
+   }
+   ```
+3. **Default** — `false` (no decision log written).
+
+When enabled, each deny/ask appends **one JSON object per line** to `.loom/logs/guard-decisions.log` (`SCRIPT_DIR`-relative, mirroring `hook-errors.log`; override the path with `LOOM_GUARD_DECISION_LOG_FILE`). **Stable schema** (the contract downstream reader tooling in #3772 depends on — field names are load-bearing):
+
+```json
+{"ts":"2026-07-22T23:17:13Z","decision":"deny","pattern":"sql-ddl","tier":"catastrophic","command":"<redacted>"}
+```
+
+| Field | Meaning |
+|-------|---------|
+| `ts` | UTC timestamp (`date -u '+%Y-%m-%dT%H:%M:%SZ'`, same as `hook-errors.log`) |
+| `decision` | `deny` or `ask` |
+| `pattern` | a short, stable rule tag (e.g. `sql-ddl`, `rm-protected-path`, `force-op:protected`, `cloud-cli:<pattern>`) — **not** the full free-text reason |
+| `tier` | `catastrophic` for a deny, `ask` for an ask |
+| `command` | the command string, **redacted** via `strip_literal_text()` so no raw `--body`/`-m`/`--title`/`--notes`/`--comment` secret value is persisted |
+
+**`allow` decisions are never logged** — the #3687 read-only fast path's zero-overhead silent-allow stays silent, and allow-logging would swamp the log with the ~99% common case. Logging is **best-effort / fail-open**: the toggle is resolved lazily (only once a deny/ask is about to fire, so it never touches the fast path's hot path), and a log-write failure (permission denied, disk full, missing dir) never changes the deny/ask decision and never causes the hook to exit non-zero. `.loom/logs/` is gitignored.
+
+Summarize fires by rule (a fuller reader/aggregation CLI is #3772's scope):
+
+```bash
+jq -r '.pattern' .loom/logs/guard-decisions.log | sort | uniq -c | sort -rn
+```
+
+**Examples**:
+
+```bash
+# Enable for a single command (e.g. to capture one session's fires)
+LOOM_GUARD_DECISION_LOG=1 claude -p "/builder" --dangerously-skip-permissions
+
+# Persist for a whole repo
+#   .loom/config.json  ->  { "guards": { "decisionLog": true } }
+
+# Force off for one command even when the repo opts in
+LOOM_GUARD_DECISION_LOG=0 <command>
+```
+
 ### Protecting Read-Only Directories
 
 Many projects have directories that should never be modified by agents (vendor code, generated files, external SDKs, process design kits). Loom provides a template hook for this.
--- a/defaults/hooks/guard-destructive.sh
+++ b/defaults/hooks/guard-destructive.sh
@@ -35,12 +35,85 @@
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd 2>/dev/null || echo ".")"
 HOOK_ERROR_LOG="${SCRIPT_DIR}/../logs/hook-errors.log"
 
+# Decision telemetry log (issue #3771) — a SEPARATE JSONL file from
+# HOOK_ERROR_LOG. At runtime SCRIPT_DIR is the installed hook's own directory
+# (.loom/hooks/), so this resolves to .loom/logs/guard-decisions.log in a real
+# install. LOOM_GUARD_DECISION_LOG_FILE overrides the path (a test seam; also
+# lets an operator point the log elsewhere). Off by default — see
+# decision_log_enabled() below.
+DECISION_LOG="${LOOM_GUARD_DECISION_LOG_FILE:-${SCRIPT_DIR}/../logs/guard-decisions.log}"
+
 # Log a diagnostic error message (best-effort, never fails the script)
 log_hook_error() {
     local msg="$1"
     # Ensure log directory exists
     mkdir -p "$(dirname "$HOOK_ERROR_LOG")" 2>/dev/null || true
     echo "[$(date -u '+%Y-%m-%dT%H:%M:%SZ')] [guard-destructive] $msg" >> "$HOOK_ERROR_LOG" 2>/dev/null || true
+}
+
+# =============================================================================
+# DECISION TELEMETRY (issue #3771) — one JSONL record per deny/ask decision.
+#
+# Append a machine-readable record to DECISION_LOG each time the guard denies or
+# asks, so false-positive friction becomes measurable (which patterns fire, how
+# often, before/after a precision fix). Deliberately does NOT log `allow`: the
+# #3687 read-only fast path's zero-overhead silent-allow must stay silent, and
+# allow-logging would swamp the log with the ~99% common case.
+#
+# STABLE SCHEMA (the contract #3772's reader/aggregation tooling stacks on — do
+# NOT rename fields without considering that dependency), one JSON object per
+# line:
+#   {"ts":"<UTC>","decision":"deny"|"ask","pattern":"<tag>",
+#    "tier":"catastrophic"|"ask","command":"<redacted>"}
+#     ts       — UTC timestamp, same format as log_hook_error's date -u call.
+#     decision — "deny" or "ask".
+#     pattern  — a short, stable rule tag (NOT the full free-text reason). For
+#                the pattern-array loops it is the matched pattern; the non-loop
+#                sites pass a static tag (e.g. "sql-ddl", "rm-protected-path").
+#     tier     — "catastrophic" for deny, "ask" for ask.
+#     command  — the command string, REDACTED via strip_literal_text() so no raw
+#                --body/-m/--title/--notes/--comment secret value is persisted.
+#
+# Best-effort like log_hook_error: gated by the lazy decision_log_enabled()
+# toggle, and a log-write failure (permission denied, disk full, missing dir)
+# NEVER changes the deny/ask decision and NEVER causes a non-zero exit. Callers
+# invoke it as `log_guard_decision ... || true` so it can never trip the ERR
+# trap.
+#
+# One-liner to summarize fires by pattern (AC — full tooling is #3772):
+#   jq -r '.pattern' .loom/logs/guard-decisions.log | sort | uniq -c | sort -rn
+# =============================================================================
+log_guard_decision() {
+    # Args: <decision> <tier> <pattern-tag>. The command is read from the global
+    # $COMMAND and redacted here. Returns 0 unconditionally.
+    decision_log_enabled || return 0
+    local decision="$1" tier="$2" tag="${3:-$1}"
+    local ts redacted line
+    ts=$(date -u '+%Y-%m-%dT%H:%M:%SZ' 2>/dev/null) || ts=""
+    # Redact quoted --body/-m/--title/--notes/--comment values (same redactor the
+    # pattern-matching tiers use) so no raw secret text is persisted to a log that
+    # aggregates across sessions. Fall back to raw only if redaction produced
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed as security-adjacent + hot-path-of-every-Bash-call code. All eight verification criteria hold; 380/380 tests pass in the PR worktree (was 370 — the 10 new decision-telemetry cases plus zero regressions), avg exec 25ms (< 200ms), CI green, mergeStateStatus CLEAN.

**(a) CRITICAL fail-open — PASS.** `log_guard_decision()` `return 0`s unconditionally; every internal command sits in a `||`/`if` context (`ts=$(date) || ts=""`, `line=$(jq…) || return 0`, the grouped `{ printf >> $DECISION_LOG; } 2>/dev/null || true`), so a log-write failure can neither flip the decision nor trip the top-level `… exit 0` ERR trap (which would wrongly turn a deny into an allow). Call sites are `log_guard_decision … || true`, closing that off at the boundary too. Test (j): unwritable path still denies + exits 0.

**(b) CRITICAL hot-path zero-overhead — PASS.** The #3687 fast path exits (`exit 0`, lines ~254/256) before any deny/ask. `decision_log_enabled()` (the lazy config `jq` read) is invoked only from inside `log_guard_decision()`, which only `deny()`/`ask()` call — so the ~99% allow path and the silent fast-path never resolve the toggle. Perf unchanged (25ms). Test (d): `git status` writes nothing.

**(c) command redaction — PASS.** `strip_literal_text($COMMAND)` reused; test (f) confirms a secret `-m` value is absent from the log. Minor non-blocking note: the `[[ -n "$redacted" ]] || redacted="$COMMAND"` empty-fallback could persist raw text if the redactor ever returned empty — impossible in practice (jq required, awk present), and the author documents it. Fine to leave; could tighten in a follow-up.

**(d) allow never logged — PASS.** No `log_guard_decision` in any allow path. Tests (c),(d).

**(e) default OFF — PASS.** Inverse polarity mirroring `reversible_gh_guard_enabled()`; zero-config install writes no file. Test (g).

**(f) stable jq-escaped JSONL schema — PASS.** `{ts,decision,pattern,tier,command}` built with `jq -cn --arg` (correct escaping); jq failure skips the write. This is the #3772 contract. Test (a) validates all fields.

**(g) deny/ask DECISIONS byte-for-byte unchanged — PASS.** Telemetry is emitted before the decision JSON and is observe-only; the only logic change is the optional 2nd rule-tag arg on `deny()`/`ask()` (defaults preserve behavior). All 370 pre-existing tests still pass.

**(h) 10 new tests — PASS.** Genuinely cover deny/ask logging, allow-not-logged, fast-path-silent-allow, log separation, redaction, default-off, config-on, env-over-config precedence, and fail-open.

Clean Chain-A root for #3772. Approving.

PR #3804, opened 2026-07-23
fix(daemon): process-group cancel + child reaping (#3800, #3801)

### LINKED ISSUE #3800: loom-daemon: CancelSweep does not terminate the spawned child (no SIGTERM/SIGKILL delivered)
## Context — found via live dispatch test (2026-07-23)

Dogfooding daemon mode: dispatched a real sweep via IPC and then tried to cancel it.

```
# dispatch
→ {"type":"DispatchSweep","payload":{"kind":{"type":"Issue","value":3780},"idempotency_key":null}}
← {"type":"SweepDispatched","payload":{"sweep_id":"sweep-issue-3780-1784767527","pid":25746,...}}

# cancel immediately
→ {"type":"CancelSweep","payload":{"sweep_id":"sweep-issue-3780-1784767527"}}
← (empty — no ack frame returned)

# child still alive afterward:
$ ps -p 25746 → 25746 ... claude -p /loom:sweep 3780   (still Running)
```

`CancelSweep` returned **no response frame** and the child process **kept running** — it even proceeded to claim the issue (`loom:building`) before I `kill -TERM`/`kill -KILL`'d it by hand. The documented cancel semantics ("SIGTERM → grace → SIGKILL", per CLAUDE.md / daemon-reference.md) are non-functional in this build.

This is serious: with dispatch working but cancel broken, there is **no supported way to stop a runaway daemon-dispatched sweep** — the operator must find and kill the OS process manually.

## Curator findings — grounded against `origin/main` @ `c1ec0cf8`

Reading the current source shows `Request::CancelSweep` **is already wired end-to-end** at HEAD — this narrows the bug to one concrete gap rather than a wholesale missing feature. The live repro above was almost certainly run against a stale/older daemon binary; do not assume the ack frame and SIGTERM/grace/SIGKILL scaffolding still need building from scratch.

**Already present at HEAD (verify, don't re-implement):**
- `loom-daemon/src/ipc.rs` `Request::CancelSweep` handler (~L832) calls `sr.cancel(...)` and returns `Response::SweepCancelled { sweep_id, pid, sigkill_sent, was_running }` — an ack frame **is** returned on this path.
- `loom-daemon/src/sweep_registry.rs::SweepRegistry::cancel()` (~L766) implements SIGTERM → 100ms-poll grace loop → SIGKILL-if-still-alive → registry transition to `SweepState::Exited{code:None,..}` → lock release → `sweep.issue.{N}.exited` + `sweep.global.completed` events. It's idempotent against already-terminal sweeps (`was_running:false`, no signal sent) and against unknown `sweep_id` (`Err`).
- `mcp-loom/src/tools/sweeps.ts` already wires `cancel_sweep` → `CancelSweep` IPC (contrary to the original AC item below assuming the MCP surface still needed wiring — it doesn't).
- Existing unit tests already cover: unknown sweep_id → error, already-terminal → no-op, and a real fixture child (`bash -c 'trap "" TERM; sleep 30'`) that ignores SIGTERM and gets escalated to SIGKILL (`cancel_escalates_to_sigkill_when_child_ignores_sigterm`).

**The actual remaining gap — single-PID signaling, not process-group:**
- `sweep_registry.rs::dispatch()` (~L663-736) spawns the child via plain `Command::new(&spawn_bin).spawn()` — **no `.process_group(0)`** (or `pre_exec`/`setsid` equivalent) is set. The child therefore inherits the daemon's own process group instead of becoming its own group leader.
- `sweep_registry.rs::send_signal()` (~L1316) sends `kill(pid, sig)` to the single tracked PID only, and explicitly rejects `pid == 0` "to avoid the POSIX broadcast-to-group semantics" — i.e. it deliberately never signals `-pid` (the group). `cancel()` calls this same single-PID path for both the SIGTERM and SIGKILL steps.
- The dispatched child is `.loom/scripts/spawn-claude.sh`, which ends in `exec claude "$@"` (or `exec "$_wrapper" ...` in `--use-wrapper` mode) — so the tracked PID **does** become the actual `claude` process (no extra shell-fork layer to worry about there). But `claude` (the Node CLI) forks real OS subprocesses for tool execution — Bash-tool commands, MCP servers, etc. Those descendants sit in the same (inherited, undifferentiated) process group and are **not reached** by a signal to the single top-level PID. A SIGTERM/SIGKILL to just `pid` can leave a git clone, a long-running shell command, or an MCP server child running as an orphan after the tracked "sweep" is marked `Exited`.

This matches the issue title precisely: "the real work is a `claude` subprocess [tree]" — cancel only ever touches the process the daemon directly forked, never its descendants.

## Revised acceptance criteria (supersedes the original list below)

- [ ] `dispatch()` spawns the sweep child in its **own new process group** — `std::process::Command::process_group(0)` (stable since Rust 1.64, sets `setpgid(0,0)` post-fork/pre-exec) on the `Command` built in `sweep_registry.rs::dispatch()`.
- [ ] `cancel()`'s signal delivery targets the **whole process group**, not just the tracked PID — signal `-pid` (negative PID = group) for both the SIGTERM and SIGKILL steps. `send_signal()` / `is_pid_alive()` need updating accordingly (group-kill semantics, not the current explicit `pid == 0` guard reasoning — that guard was written to describe the *avoided* group-kill, which is now the desired behavior for the tracked child specifically, not incoming pid==0).
- [ ] `CancelSweep` continues to return a `Response::SweepCancelled` ack frame (already correct — regression-guard, don't lose this).
- [ ] Registry state continues to transition `Running`/`Pending` → `Exited` on cancel (already correct — regression-guard).
- [ ] **New** Rust test: spawn a fixture whose immediate child forks a grandchild (e.g. `bash -c 'sleep 30 & wait'` or `bash -c '(sleep 30 &); sleep 30'`), register it as the sweep's tracked PID, call `cancel()`, and assert the **entire process tree** (parent + grandchild) is gone within the grace window — not just the top-level PID as the existing `cancel_escalates_to_sigkill_when_child_ignores_sigterm` test already covers.
- [ ] Manually verify (or add an integration-style check) that a real `spawn-claude.sh`-dispatched `claude` process, once it has started an actual Bash-tool subprocess, is fully torn down (no orphaned descendants) after `cancel_sweep`.

## Coordination note

**#3800 and #3801 are tightly coupled** and both heavily edit `loom-daemon/src/sweep_registry.rs` — they will be implemented **together in one consolidated child-process-lifecycle PR**. A correctly working `cancel()` (this issue) depends on the reaper (#3801) reconciling registry/zombie state after the group-kill; conversely the reaper's dead-PID detection is exercised by the same kill path this issue fixes. Do not land one without the other in the same PR.

## Original suggested acceptance criteria (context only — see revised list above)
- [ ] `CancelSweep { sweep_id }` returns an ack frame (success/failure) over IPC.
- [ ] It actually delivers SIGTERM to the spawned child's process group, waits the grace period, then SIGKILL — and the child (and its `claude`/node subprocess tree) is gone afterward.
- [ ] Cancelling updates the registry state (`Running` → `Cancelled`/`Exited`), not left as `Running`.
- [ ] A test dispatches a dummy child and asserts CancelSweep terminates it within the grace window.
- [ ] `mcp__loom__cancel_sweep` (once the MCP surface is wired — see companion issue) exercises the same path.

### LINKED ISSUE #3801: loom-daemon: reaper does not reap exited children — zombies persist and list_sweeps reports stale Running state
## Context — found via live dispatch test (2026-07-23)

After a dispatched child (PID 25746, parent = daemon PID 74709) was terminated, it became a **zombie** that the daemon never reaped:

```
$ ps -o pid,ppid,stat,command -p 25746
25746 74709 Z  <defunct>
```

And `list_sweeps` continued to report it as **Running** long after the process was dead:

```
→ {"type":"ListSweeps","payload":{}}
← ... issue=3780 pid=25746 state=Running   (process was already dead/zombie)
```

Two related reaper defects:
1. **Dead children are not `wait()`-reaped** → zombie `<defunct>` entries accumulate under the daemon (PPID = daemon).
2. **Registry state is not reconciled** → `list_sweeps` / `get_sweep_status` keep reporting `Running` for a dead PID. Also observed at daemon startup: **5 phantom `recovered/Crashed` entries** were synthesized from existing `.loom/sweep-checkpoint/issue-*.json` files and never cleaned up.

## Scope note: co-implemented with #3800

This issue and **#3800** (`CancelSweep` does not actually terminate the child) are tightly coupled and both heavily edit `loom-daemon/src/sweep_registry.rs`. They will land together in **one consolidated child-process-lifecycle PR**: #3800 fixes `CancelSweep` to actually deliver SIGTERM/SIGKILL and ack over IPC; **this issue (#3801) is the "reap + state-reconcile" half** — once a child is killed (whether by `CancelSweep`, an external `kill`, or a natural exit), the daemon must `wait()` it (no zombie) and reconcile the registry entry to a terminal state. The builder should implement both in the same branch/PR; do not split into two separate PRs.

## Root cause (verified against `loom-daemon/src/sweep_registry.rs` @ origin/main)

- `SweepRegistry::spawn_child` (`sweep_registry.rs`, the `Spawn` section, ~L730-736) spawns the child via `std::process::Command` and then **explicitly drops the `Child` handle** without ever calling `.wait()`:
  ```rust
  let child = cmd.spawn()...;
  let pid = child.id();
  // We do NOT wait on the child — detach by dropping the handle. The
  // reaper detects exit via `kill(pid, 0)`. ...
  std::mem::drop(child);
  ```
- Liveness is polled exclusively via `is_pid_alive()` (~L1267-1278), which calls `kill(pid, 0)` (POSIX liveness probe) and nothing else.
- **This is the actual bug**: on POSIX, a terminated child that its parent never `wait()`s/`waitpid()`s becomes a zombie — the kernel keeps its process-table slot (state `Z`, `<defunct>`) until reaped. `kill(pid, 0)` on a zombie **still returns success** (the PID is still allocated), so `is_pid_alive()` keeps reporting the zombie as alive forever. `reap_once()`'s dead-PID branch (`SweepState::Running | SweepState::Pending if !is_pid_alive(pid)`, ~L906) therefore never fires, the registry entry is stuck `Running` indefinitely, and the zombie is never reclaimed by the OS. This matches both observed symptoms exactly (stale `Running` + accumulating `<defunct>`).
- `SweepRegistry::cancel()` (~L756-850) has the identical gap on the kill path: it sends SIGTERM/SIGKILL via `send_signal()` (raw `kill(2)`, ~L1315-1323) and unconditionally writes `SweepState::Exited` in-registry after the grace/poll loop, but it **never calls `wait()`/`waitpid()` on the killed child either** — so even though the *registry* is correctly reconciled by `cancel()`, the *OS-level* zombie still accumulates under the daemon's PID. Fixing the reap path here should cover this case too (e.g. a background reap sweep that `waitpid(pid, WNOHANG)`s any PID the daemon spawned, regardless of whether it exited naturally or was cancelled).
- `SweepRegistry::reconstruct()` (~L1042-1180) — checked separately from the zombie bug: checkpoint-only entries (no live lock owner) are already admitted directly as `SweepState::Crashed { .. }` at `pid: 0` (~L1154), not `Running`, and `reap_once()`'s trailing GC block (~L1006-1023) already drops any terminal (`Exited`/`Crashed`) entry once `Utc::now() - at > TERMINAL_RETENTION_SECS` (3600s, `sweep_registry.rs` L77). So the "5 phantom recovered/Crashed entries never cleaned up" observation is most likely just an observation window shorter than the 1-hour retention window, not a missing prune path — **do not add new prune logic here without first confirming (via test) that the existing GC path is actually broken**. If the AC's out-of-band-kill test shows GC does work correctly for `recovered` entries, no code change is needed for that half of the bug; just assert it in a test as a regression guard.

## Acceptance criteria

- [ ] The daemon actually reaps exited/killed children — no `<defunct>` zombie entries accumulate under the daemon's PID for a sweep child that has terminated (natural exit, external `kill`, or `CancelSweep`-initiated kill from #3800). This requires an actual `wait()`/`waitpid()` call somewhere in the daemon's process lifecycle (e.g. retain the `Child` handle instead of dropping it and `wait()` it from a dedicated reap path, or use `libc::waitpid(pid, &mut status, WNOHANG)` polled from the existing reaper tick) — `kill(pid, 0)` alone is proven insufficient because it reports zombies as alive.
- [ ] On child exit (natural or killed), the registry entry transitions `Running`/`Pending` → `Exited`/`Crashed` with the real exit status within one reaper interval (default 30s, `LOOM_SWEEP_REAPER_INTERVAL_SECS`), so `list_sweeps`/`get_sweep_status` never report a dead PID as `Running`.
- [ ] `SweepRegistry::cancel()`'s kill path (`sweep_registry.rs` ~L756-850, shared with #3800's fix) does not leave a zombie under the daemon after SIGTERM/SIGKILL — verified by the same reap mechanism, not a separate one-off `wait()` bolted onto `cancel()`.
- [ ] Startup `reconstruct()` entries are unaffected in intent (checkpoint-only entries already admit as `Crashed`, not `Running`) — add/confirm a test that they are pruned by the existing GC path within `TERMINAL_RETENTION_SECS`, and note in a code comment if the retention window is intentionally not shortened for this case.
- [ ] A Rust test in `loom-daemon/src/sweep_registry.rs` spawns a real child process (or reuses the existing `cancel_kills_live_child_and_reaps_exit_status`-style fixture around L2654-2716), kills it out-of-band (SIGKILL sent directly, bypassing `cancel()`), and asserts: (a) the registry entry transitions out of `Running`/`Crashed`-or-`Exited` within a bounded number of reaper ticks, and (b) `ps`/`/proc` (or a `waitpid(WNOHANG)` probe) shows no zombie remains for that PID.
- [ ] `cargo test` in `loom-daemon/` passes, including the new/updated reaper tests.

## Test plan

```bash
cd loom-daemon
cargo test sweep_registry:: -- --nocapture
cargo test reap
cargo test cancel
```
Manually: dispatch a sweep via `mcp__loom__dispatch_sweep` (or `DispatchSweep` IPC), `kill -KILL` the child PID directly (not via `CancelSweep`), wait one reaper interval, confirm via `ps -o stat,ppid -p <pid>` that no `Z <defunct>` entry remains and `list_sweeps` reports the sweep as `Exited`/`Crashed`.

## Affected files

- `loom-daemon/src/sweep_registry.rs` (spawn_child, cancel, reap_once, reconstruct, is_pid_alive, spawn_reaper_task)
- `loom-daemon/src/main.rs` (reaper task spawn wiring, ~L233-242)

rjwalters:
Closes #3800
Closes #3801

Consolidated fix for the two tightly-coupled daemon child-process-lifecycle bugs that block safe daemon operation. Both live in `loom-daemon/src/sweep_registry.rs`, so one PR closes both.

## #3800 — CancelSweep left the `claude` subprocess subtree orphaned

**Root cause (curator-verified):** the ack frame + SIGTERM→grace→SIGKILL escalation + Running→Exited transition already existed. The real gap was that `dispatch()` spawned the child in the *daemon's own* process group, and `cancel()`'s `send_signal()` only signalled the single tracked PID. `spawn-claude.sh` ends in `exec claude`, so the tracked PID is the `claude` process itself — but `claude` forks real OS subprocesses for tool execution (Bash-tool commands, MCP servers, git clones). Those descendants sat in the inherited group and were never reached, surviving as orphans after the sweep was marked `Exited`.

**Fix:**
- `dispatch()` now spawns the child as its own **process-group leader** via `Command::process_group(0)` (`setpgid(0, 0)`, stable since Rust 1.64).
- `cancel()` now signals the **whole process group** — `kill(-pgid, sig)` via the new `send_group_signal()` — for both the SIGTERM and SIGKILL steps. Because the child is a group leader, its pgid == its PID, so the tracked PID reaches the entire subtree. Reconstructed entries with no retained handle fall back to single-PID delivery.

## #3801 — reaper never reaped exited children (zombies + stale `Running`)

**Root cause (curator-verified):** `spawn_child()` dropped the `Child` handle without ever `wait()`-ing it, and liveness used only `kill(pid, 0)`, which still returns success on an unreaped **zombie** — so `reap_once()`'s dead-PID transition never fired, the registry stayed stuck `Running`, and `<defunct>` zombies accumulated under the daemon PID.

**Fix:**
- The registry now **retains each spawned `Child` handle** (new `children` map, keyed by `sweep_id`).
- New `poll_liveness()` `try_wait()`s the retained handle each reaper tick: it reaps the process (no zombie) and reads the **real exit status**, which is now recorded on the `Exited` transition. Falls back to `kill(pid, 0)` for reconstructed entries with no handle.
- This one mechanism covers **natural exit, out-of-band `kill`, and #3800's cancel path** — `cancel()` also `reap_handle()`s the killed leader so it never lingers as a zombie.
- `reconstruct()`'s checkpoint-only entries already admit as `Crashed` and are GC'd by the existing `TERMINAL_RETENTION_SECS` (1h) path — **verified, not re-implemented**, per the curator's finding.

## Tests (Rust, in `loom-daemon`)

- `cancel_terminates_whole_process_group_including_grandchild` — dispatches a fixture whose leader forks a backgrounded grandchild (both in the leader's group), cancels, and asserts the **entire tree** is gone within the grace window. A single-PID kill would orphan the backgrounded grandchild, so this fails without the group-kill fix.
- `reaper_reaps_out_of_band_killed_child_and_transitions_state` — SIGKILLs a dispatched child out-of-band (bypassing `cancel()`), then asserts the reaper transitions the entry `Running`→terminal and leaves **no zombie** (`kill(pid, 0)` fails afterward).

All existing cancel/reap/dispatch/reconstruct tests remain green (regression guards for the ack frame, idempotency, and state transitions).

## Verification

- `cargo test` — 412 lib tests + all integration suites pass
- `cargo build --release` — passes
- `cargo fmt --check` — clean
- `cargo clippy -- -D warnings` — clean for `sweep_registry.rs`. One pre-existing, unrelated error remains in `forge_parser.rs:91` (present on `origin/main`; local clippy 1.96.0 is newer than the CI-pinned toolchain that ships main green) — untouched, out of scope.

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -44,8 +44,10 @@ use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
+#[cfg(unix)]
+use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::process::{Child, Command, Stdio};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -195,6 +197,25 @@ struct LockOwner {
 pub struct SweepRegistry {
     config: SweepRegistryConfig,
     entries: BTreeMap<SweepId, SweepInfo>,
+    /// Retained `Child` handles for sweeps this daemon instance spawned
+    /// (Issue #3801). Keyed by `sweep_id`.
+    ///
+    /// The handle is kept — rather than dropped at spawn — so the reaper
+    /// (and `cancel`) can `try_wait()` / `wait()` the child, which reaps
+    /// the OS-level process (no `<defunct>` zombie under the daemon PID)
+    /// AND yields the real exit status. `kill(pid, 0)` alone is proven
+    /// insufficient: a terminated-but-unreaped child is a zombie whose PID
+    /// is still allocated, so `kill(pid, 0)` reports it alive forever and
+    /// the registry stays stuck `Running`.
+    ///
+    /// Reconstructed entries (from a prior daemon, see [`reconstruct`]) have
+    /// no handle here — we never spawned them — so their liveness falls
+    /// back to the `kill(pid, 0)` probe. Those entries are already admitted
+    /// as terminal (`Crashed`) or point at the previous daemon's PID, so the
+    /// fallback is correct for them.
+    ///
+    /// [`reconstruct`]: SweepRegistry::reconstruct
+    children: BTreeMap<SweepId, Child>,
     /// Optional event bus for lifecycle events (Issue #3453, Phase B).
     /// When `None`, the registry behaves identically to Phase A — bus
     /// emission is best-effort and never blocks core dispatch/reaper
@@ -212,6 +233,7 @@ impl SweepRegistry {
         Self {
             config,
             entries: BTreeMap::new(),
+            children: BTreeMap::new(),
             bus: None,
         }
     }
@@ -222,6 +244,7 @@ impl SweepRegistry {
         Self {
             config,
             entries: BTreeMap::new(),
+            children: BTreeMap::new(),
             bus: Some(bus),
         }
     }
@@ -359,9 +382,12 @@ impl SweepRegistry {
 
         // 5. Compute the log path and spawn the child.
         let log_path = self.compute_log_path(issue_number);
-        let (pid, token_name) = self
+        let (child, token_name) = self
             .spawn_child(issue_number, &log_path, &sweep_id, model, effort, depends_on)
             .context("failed to spawn sweep child")?;
+        let pid = child.id();
+        // Retain the handle so the reaper can `try_wait()` it (Issue #3801).
+        self.children.insert(sweep_id.clone(), child);
 
         // 6. Record the entry. The model is carried on the registry entry
         //    (#3482, Phase 3a observability) so `list_sweeps` /
@@ -626,7 +652,7 @@ impl SweepRegistry {
         model: Option<&str>,
         effort: Option<&str>,
         depends_on: Option<u32>,
-    ) -> Result<(u32, String)> {
+    ) -> Result<(Child, String)> {
         let spawn_bin = self.config.resolve_spawn_bin()?;
 
         // Ensure log dir exists.
@@ -705,6 +731,19 @@ impl SweepRegistry {
             .stdout(Stdio::from(log_file))
             .stderr(Stdio::from(log_clone));
 
+        // Issue #3800: put the sweep child in its OWN process group
+        // (`setpgid(0, 0)` runs post-fork/pre-exec via `process_group(0)`,
+        // stable since Rust 1.64). spawn-claude.sh ends in `exec claude`, so
+        // the tracked PID becomes the `claude` process itself AND the leader
+        // of a fresh group. `claude` forks real OS subprocesses for tool
+        // execution (Bash-tool commands, MCP servers, git clones, …); those
+        // descendants inherit this group. Making the child a group leader lets
+        // `cancel()` signal the WHOLE group (`kill(-pgid, sig)`) so the entire
+        // sweep subtree is torn down — instead of leaving orphans behind when
+        // only the top-level PID is signalled.
+        #[cfg(unix)]
+        cmd.process_group(0);
+
         // Issue #3730: explicitly forward the experiment-related env vars to
         // the detached child via an EXPLICIT ALLOWLIST — never a blanket
         // env_clear/copy. Without this, `LOOM_MODEL_EXPERIMENT` /
@@ -728,14 +767,14 @@ impl SweepRegistry {
         let child = cmd
             .spawn()
             .with_context(|| format!("failed to spawn {} -p '{}'", spawn_bin.display(), prompt))?;
-        let pid = child.id();
-        // We do NOT wait on the child — detach by dropping the handle. The
-        // reaper detects exit via `kill(pid, 0)`. spawn-claude.sh internally
-        // selects a token; we record "unknown" here because the wrapper's
-        // selection is logged to the per-sweep log, not exposed on stdout.
-        std::mem::drop(child);
-
-        Ok((pid, "unknown".to_string()))
+        // Issue #3801: we RETAIN the `Child` handle (returned to `dispatch`,
+        // which stores it in `self.children`) instead of dropping it. The
+        // reaper `try_wait()`s it each tick so an exited child is reaped
+        // (no `<defunct>` zombie) and the registry transitions to a terminal
+        // state with the real exit status. spawn-claude.sh internally selects
+        // a token; we record "unknown" here because the wrapper's selection is
+        // logged to the per-sweep log, not exposed on stdout.
+        Ok((child, "unknown".to_string()))
     }
 
     // ------------------------------------------------------------------------
@@ -750,10 +789,68 @@ impl SweepRegistry {
         self.entries.get(sweep_id).cloned()
     }
 
+    /// Signal a sweep's process. When this daemon still owns a retained
+    /// `Child` handle for `sweep_id` (i.e. we spawned it into its own process
+    /// group via `process_group(0)`), the signal is delivered to the WHOLE
+    /// process group (`kill(-pgid, sig)`, and the leader's pgid == its pid)
+    /// so the entire `claude` subprocess subtree is reached (Issue #3800).
+    /// Reconstructed entries with no handle fall back to single-PID delivery.
+    fn signal_sweep(&self, sweep_id: &str, pid: u32, sig: i32) -> bool {
+        if self.children.contains_key(sweep_id) {
+            send_group_signal(pid, sig)
+        } else {
+            send_signal(pid, sig)
+        }
+    }
+
+    /// Determine whether a sweep's child has terminated, reaping it when it
+    /// has. Prefers the retained `Child` handle: `try_wait()` reaps an exited
+    /// child (no zombie) and yields the real exit status. Falls back to the
+    /// `kill(pid, 0)` liveness probe for reconstructed entries with no handle.
+    ///
+    /// Returns `(is_dead, exit_code)`. On a handle-observed exit the handle is
+    /// removed from `self.children`; `exit_code` is `None` when the child was
+    /// terminated by a signal (no clean code) or when liveness came from the
+    /// fallback probe.
+    fn poll_liveness(&mut self, sweep_id: &str, pid: u32) -> (bool, Option<i32>) {
+        if let Some(child) = self.children.get_mut(sweep_id) {
+            match child.try_wait() {
+                Ok(Some(status)) => {
+                    let code = status.code();
+                    self.children.remove(sweep_id);
+                    (true, code)
+                }
+                Ok(None) => (false, None),
+                Err(e) => {
+                    log::warn!("sweep_registry: try_wait for {sweep_id} (pid {pid}) failed: {e}");
+                    let dead = !is_pid_alive(pid);
+                    if dead {
+                        self.children.remove(sweep_id);
+                    }
+                    (dead, None)
+                }
+            }
+        } else {
+            (!is_pid_alive(pid), None)
+        }
+    }
+
+    /// Reap the retained `Child` handle for `sweep_id`, blocking briefly until
+    /// it exits. Called after `cancel` has SIGKILL'd (or observed the exit of)
+    /// the child so the OS-level zombie is reclaimed under the daemon PID.
+    /// No-op when no handle is retained (reconstructed / test-injected entry).
+    fn reap_handle(&mut self, sweep_id: &str) -> Option<std::process::ExitStatus> {
+        self.children.remove(sweep_id).and_then(|mut child| {
+            // Bounded: we only reach here once the child has exited or has
+            // just been SIGKILL'd, so `wait()` returns promptly.
+            child.wait().ok()
+        })
+    }
+
     /// Cancel a running sweep.
     ///
-    /// Sends SIGTERM, waits up to `grace` for the child to exit (polled
-    /// via `kill(pid, 0)`), then SIGKILL if still alive. On any path the
+    /// Sends SIGTERM to the sweep's process group, waits up to `grace` for the
+    /// child to exit, then SIGKILL to the group if still alive. On any path the
     /// registry entry is transitioned to `Exited{code: None, at: now}`
     /// and the per-issue lock is released. Emits the same lifecycle
     /// events the reaper would emit on a clean exit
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED ✅

Reviewed with maximal rigor as signal/async-lifecycle code. This is a correct, well-scoped consolidated fix for both #3800 and #3801, entirely in `sweep_registry.rs` (+348/-27).

### (a) CRITICAL SAFETY — `kill(-pgid)` targeting is AIRTIGHT
- `send_group_signal` is invoked **only** via `signal_sweep`, and **only** when `self.children.contains_key(sweep_id)` — i.e. exclusively for children this daemon spawned with `process_group(0)`.
- `process_group(0)` → `setpgid(0,0)` makes the child a **new** group leader (pgid == child PID, distinct from the daemon's group). If `setpgid` failed, `spawn()` would error and no `Child`/entry would ever exist — so a group-signaled entry is provably a real group leader.
- The `pid` passed is the entry's `pid`, set together with the handle insert in `dispatch()` (`pid == child.id() ==` leader PID).
- `send_group_signal` rejects `pgid == 0` (can never hit the daemon's own group); `pgid` is `u32` so never negative; `try_into::<i32>()` guards overflow → `false`. So `-pgid_t` is always a valid negative in `[-i32::MAX, -1]`, targeting exactly the child's own group and nothing wider.
- Reconstructed (prior-daemon) entries have no handle → single-PID fallback (`send_signal`), never group-signaled. Correct and safe.
- **Conclusion: a group-signal can never reach the daemon's own group, the operator's shell, or any unrelated/wider group.**

### (b) Reaches grandchildren
`cancel_terminates_whole_process_group_including_grandchild` forks a backgrounded grandchild into the leader's group and asserts the **whole tree** dies within grace — passes (would fail under a single-PID kill). Single-PID fallback is taken only when no handle is retained.

### (c) Reaper reaps + transitions; no double-wait / handle leak
- `poll_liveness` `try_wait()`s and **removes the handle on observed exit**, so a later `reap_handle` finds nothing — the poll-observed-exit and SIGKILL paths are mutually exclusive (handle present XOR removed): no double-wait, no handle-moved-out issue.
- Real exit status is now recorded (`status.code()`); Crashed-vs-Exited checkpoint logic preserved byte-for-byte. The `children` map is cleaned on terminal transition plus a defensive GC `remove` — no handle leak.

### (d) No deadlock / lock-ordering regression
Reaper uses **non-blocking** `try_wait()` under the registry lock. `cancel` uses blocking `wait()` only after the child exited or was SIGKILL'd (dies promptly). `&mut self` on both `reap_once` and `cancel` serializes them via the registry `Mutex` — no concurrent `try_wait`/`wait` race. The grace poll-sleep under lock is pre-existing, not introduced here.

### (e) Reconstructed entries without a handle
`poll_liveness` falls back to `!is_pid_alive(pid)`, `signal_sweep` to single-PID `send_signal`, `reap_handle` is a no-op — all via `Option`-returning `BTreeMap` ops. No panic on the absent handle.

### (f) Verification (all run locally)
- `cargo build --release` — pass
- `cargo test` — **412 lib tests + integration suites pass**, incl. both new tests
- `cargo clippy -- -D warnings` — `sweep_registry.rs` clean. The only error is the **pre-existing, unrelated** `forge_parser.rs:91` uninlined-format-args lint — confirmed verbatim on `origin/main`, untouched by this PR (which touches only `sweep_registry.rs`), and the CI-pinned toolchain ships main green (the CI "Rust Linting & Build Check" job passed).
- `cargo fmt --check` — clean
- CI: all required checks green; merge state **CLEAN / MERGEABLE**.

The signal-targeting safety — the paramount concern for unblocking daemon mode — is airtight. Approving.

Issue #3809, opened 2026-07-23
Epic: Fully autonomous daemon mode — label an issue, it builds in the background


rjwalters:
## Vision (next-release headline)

The operator creates issues and the repo builds itself unattended. You file `loom:triage` issues; the autonomous pipeline curates, promotes, **builds**, judges, and merges — you focus on issue creation, not orchestration.

## Current state (verified 2026-07-23)

The label pipeline is **already autonomous except for one component**:

```
you create issues (loom:triage)
   → Curator enriches      ✅ cron/tmux pool          → loom:curated
   → Champion promotes      ✅ cron/tmux pool          → loom:issue
   → ??? BUILDS IT           ❌ THE MISSING BRAIN
   → Judge reviews          ✅ cron/tmux pool          → loom:pr
   → Champion merges        ✅ cron/tmux pool
```

Nothing turns a `loom:issue` into a built PR without an operator running `/loom:sweep` or dispatching to the daemon by hand. Verified: the Rust `loom-daemon` has **no forge poller** (only the explicit `DispatchSweep` IPC entry point + reaper/health loops); the 5 cron roles (Curator/Champion/Judge/Auditor/Guide) contain **no Builder**; the tmux pool config (`.loom/config.json` terminals) is **judge/curator/hermit/champion only**; `loom-scale.sh` is a **manual** scaler whose examples still reference the **deleted `shepherd`** role. The Layer-2 Python orchestrator (`daemon_v2/` + `shepherd/`) that used to poll work + auto-scale the builder pool was **deleted in v0.10.0** and its work-finding/pool-scaling function was never rebuilt (the Rust rebuild restored dispatch + monitoring only).

**Enabler already in place:** a 7-account token pool is bootstrapped (`.loom/tokens/`), so a real concurrent builder pool (each build on a rotated OAuth token) is feasible now.

## Design principle: git-based safety (reactive, not preventive)

Overlapping parallel builds are made safe by **git + a reactive main-health backstop**, not by expensive dispatch-time collision prevention:
- **Textual conflicts** → git surfaces them (second merge goes `DIRTY`); nothing corrupts or is lost.
- **Semantic/cross-file breakage** (the #3647 class — disjoint files, red `main`) → **recoverable, not catastrophic**: caught reactively by a `buildGate`-against-`main` (already shipped, #3749), the Auditor cron, and revert-on-red — NOT by preventing the collision at dispatch.

[...]

Issue #3805, opened 2026-07-23
loom-daemon IPC: malformed/unparseable requests are silently dropped (no error frame) instead of returning StructuredError


rjwalters:
## Context — found while dogfooding daemon cancel (2026-07-23)

Driving the daemon over its raw Unix-socket IPC, every malformed request returns an **empty response and a closed connection** — no error frame:

```
garbage 'not json'                              → (empty)
{"type":"CancelSweep","payload":{"sweep_id":"x"}}  → (empty)   # missing required grace_secs
{"type":"Nonsense","payload":{}}                → (empty)   # unknown request type
```

A `StructuredError` response variant exists (`loom-daemon/src/ipc.rs:210` — `Err(e) => Response::StructuredError(DaemonError::from(e))`), but it is only reached for **handler-level** errors. A **request-parse/deserialize** failure (bad JSON, missing required field, unknown `type`) happens before the handler and silently closes the connection without emitting any frame.

**Real cost:** this directly caused a false diagnosis. A `CancelSweep` frame missing the required `grace_secs` field returned empty — indistinguishable from "cancel is broken." The cancel path was actually fine; the request was malformed. A structured parse-error frame would have made this a one-second fix instead of a long debugging detour. It also makes the IPC surface effectively undebuggable for any client hand-building frames (relates to #3803 — the MCP tools that would build correct frames aren't exposed).

## Suggested acceptance criteria
- [ ] A request that fails to deserialize (bad JSON, missing required field, unknown `type`) returns a structured error frame (e.g. `Response::StructuredError` / a `BadRequest` variant) naming the parse failure, rather than silently closing the connection.
- [ ] The error frame identifies the offending field/type where possible (serde error message is enough).
- [ ] The connection stays usable (or closes only after sending the error), so a client gets actionable feedback.
- [ ] A test sends each malformed shape (garbage, missing field, unknown type) and asserts a structured error frame is returned.

rjwalters:
## Curator enrichment

Verified the root cause and confirmed all three malformed shapes hit the same code path.

### Root cause (precise)

`loom-daemon/src/ipc.rs`, `handle_client` (~line 95-130):

```rust
while let Some(line) = lines.next_line().await? {
    let request: Request = serde_json::from_str(&line)?;   // line 106 — bare `?`
    ...
}
```

[...]

PR #3815, opened 2026-07-23
fix(daemon): singleton guard — refuse to start if a live daemon owns the socket (#3806)

### LINKED ISSUE #3806: loom-daemon: no singleton guard — a second daemon removes and steals the socket, silently orphaning the first
## Context — hit live while restarting the daemon (2026-07-23)

Daemon startup unconditionally deletes any existing socket and rebinds:

```rust
// loom-daemon/src/ipc.rs:69-71
let _ = fs::remove_file(&self.socket_path).await;
let listener = UnixListener::bind(&self.socket_path)?;
```

There is **no singleton guard** (no PID-file lock, no "is another daemon already listening?" check). So starting a second `loom-daemon` while one is already running **removes the first daemon's socket out from under it and binds a fresh one** — leaving the first daemon still running as a process but orphaned (no socket, still holding its children), and the second now owning the socket. Two daemons, silently, with split ownership of sweep children.

Observed live: after a rebuild I intended to replace the daemon, but ended up with **both** PID 74709 (old, pre-fix, still holding a zombie child) and PID 70355 (new) running simultaneously — the old one orphaned. It took explicit `ps`/`kill` forensics to notice and clean up. A stale-socket takeover is fine; silently orphaning a *live* daemon is not.

## Suggested acceptance criteria (original)
- [ ] On startup, detect whether another daemon is already listening on the socket (e.g. connect-and-ping before `remove_file`, and/or a PID-file lock with liveness check).
- [ ] If a live daemon is found: refuse to start with a clear message (and non-zero exit), OR perform a coordinated takeover that first shuts the incumbent down — never silently orphan it.
- [ ] Only a genuinely stale socket (no live listener) is removed and rebound.
- [ ] A test starts two daemons and asserts the second refuses (or cleanly takes over) rather than orphaning the first.

---

### Root cause

`IpcServer::run()` (`loom-daemon/src/ipc.rs:67-92`) does exactly two things at
startup before it ever accepts a connection:

```rust
pub async fn run(&self) -> Result<()> {
    // Remove old socket
    let _ = fs::remove_file(&self.socket_path).await;   // line 69 — unconditional, error swallowed

    let listener = UnixListener::bind(&self.socket_path)?;  // line 71
    ...
}
```

Because `remove_file` unlinks the path unconditionally (its `Result` is
discarded with `let _ =`), there is no distinction between "the socket file
is stale (no listener behind it — the crash-recovery case this line exists
for)" and "the socket file is live (a healthy daemon is actively `accept()`ing
on it)". `main.rs` calls `server.run().await?` as the last statement in
`main()` (`loom-daemon/src/main.rs:264`), under `#[tokio::main] async fn
main() -> Result<()>` — so an `Err` returned from `run()` already produces a
non-zero process exit with the error printed by the default `Result`-in-`main`
handler. That means "refuse to start" requires **no new plumbing** in
`main.rs` beyond making `run()` return `Err` in the live-daemon case; the
non-zero exit is already wired.

`socket_path` itself is resolved once in `main.rs:118-135` — either from
`LOOM_SOCKET_PATH` (test override) or `~/.loom/loom-daemon.sock` — and passed
into `IpcServer::new()` unchanged (`main.rs:245`), so the guard has exactly
one call site to intercept: the top of `IpcServer::run()`.

Two other places touch the socket path and are **not** part of this issue's
scope but are worth noting so the fix doesn't collide with them:
- `main.rs:248-261` — the `ctrl_c()` signal handler removes the socket file
  on graceful shutdown. This is a *different* correct-removal path (the
  process that owns the socket is exiting) and should be left as-is.
- No PID-file infrastructure exists anywhere in `loom-daemon` today (`grep -n
  "pid_file\|PidFile" loom-daemon/src/*.rs` is empty) — a PID-file approach
  is new infrastructure, not a refactor of something existing.

### Implementation guidance

**Recommended approach: connect-and-ping liveness probe, no new dependency.**

Add a liveness check to `IpcServer` (or a free function) that runs *before*
`remove_file`:

```rust
/// Returns true if a live daemon is currently listening on `socket_path`.
/// A connect failure (ECONNREFUSED, ENOENT, etc.) means the socket is
/// either absent or stale (file exists but nothing is listening) — safe
/// to remove. A connect success + Pong roundtrip means a live daemon owns
/// the socket.
async fn socket_has_live_listener(socket_path: &Path) -> bool {
    let Ok(stream) = tokio::time::timeout(
        Duration::from_millis(500),
        UnixStream::connect(socket_path),
    )
    .await
    .and_then(|r| r.ok().ok_or(())).ok().flatten() else {   // pseudocode-ish; see note below
        return false;
    };
    // Send Ping, await Pong within a short timeout — confirms the peer
    // is a loom-daemon (not some other process that happens to hold the
    // path) and that it's actually servicing requests, not just accepting
    // TCP-level connections into a stuck accept() loop.
    ...
}
```

(Adjust the `timeout`/`Result` plumbing to compile cleanly — the sketch above
is illustrative, not literal Rust.) Wire it into `run()`:

```rust
pub async fn run(&self) -> Result<()> {
    if socket_has_live_listener(&self.socket_path).await {
        anyhow::bail!(
            "another loom-daemon is already listening on {} — refusing to start. \
             If you intended to replace it, stop the running daemon first \
             (e.g. `kill <pid>` or the daemon's own shutdown path) and retry.",
            self.socket_path.display()
        );
    }
    let _ = fs::remove_file(&self.socket_path).await;
    let listener = UnixListener::bind(&self.socket_path)?;
    ...
}
```

This satisfies "refuse to start" (option A in the suggested ACs). It reuses
the existing `Request::Ping` / `Response::Pong` roundtrip already defined in
`types.rs` and exercised by `TestClient::ping()` in
`loom-daemon/tests/common/mod.rs:163-171` — no new wire protocol needed.

**Why connect-and-ping over a PID file:**
- It directly probes the actual contended resource (the socket), so it can't
  drift from reality the way a PID file can (e.g. PID reused by an unrelated
  process after a crash — `kill -0 <pid>` would report "alive" even though
  it's not the daemon).
- No new dependency: `nix`/`libc` (for `kill -0`) is **not** currently a
  `loom-daemon` dependency (see `Cargo.toml`); shelling out to `kill -0 $PID`
  via `std::process::Command` works but is a second, redundant liveness
  signal once the socket probe exists.
- It's directly testable with the existing `TestDaemon`/`TestClient`
  integration-test harness (see Test Plan below) — no fixture PID files to
  fake.

**Alternative (do not implement unless the recommended approach turns out
insufficient in review): PID-file lock.** Write `~/.loom/loom-daemon.pid` on
startup, check-and-liveness-probe (`kill -0`) on next startup, refuse if
alive, remove-and-continue if stale. More moving parts (new file, new
dependency or `Command`-based `kill -0`, race between PID reuse and check)
for no behavior the socket probe doesn't already give — only fall back to
this if the connect-and-ping approach proves unreliable in practice (e.g.
socket permissions edge cases).

**Coordinated-takeover option (B in the suggested ACs) is explicitly
out of scope for this issue.** Automatically killing an incumbent daemon is
a materially larger and riskier change (needs to drain/hand off in-flight
sweeps, terminal sessions, and the event bus rather than just refusing) and
duplicates work planned in the sibling stacked-chain issues (#3805, #3807,
#3808) under epic #3809. Ship "refuse to start with a clear message"
first; a takeover mode can be a fast follow if ever needed.

### Acceptance criteria (concrete)

- [ ] `IpcServer::run()` probes for a live listener on `self.socket_path`
      *before* calling `fs::remove_file` (connect + `Ping`/`Pong` roundtrip,
      short timeout e.g. 500ms).
- [ ] If a live daemon is detected, `run()` returns `Err` with a message that
      names the socket path and instructs the operator how to stop the
      running daemon; no `remove_file` or `bind` call happens in this branch.
      `main()`'s existing `server.run().await?` already turns this into a
      non-zero process exit — no `main.rs` changes required for the exit
      code itself.
- [ ] If the socket path is absent, or present but nothing answers a
      connect/Ping (stale file from a crashed daemon), `run()` proceeds
      exactly as today: `remove_file` (best-effort) + `bind`.
- [ ] A new Rust test starts two daemons pointed at the *same*
      `LOOM_SOCKET_PATH` and asserts: (a) the second daemon process exits
      non-zero and never creates/rebinds the socket out from under the
      first, and (b) the first daemon is still alive and answering `Ping`
      after the second daemon's failed startup attempt (i.e., not orphaned).
- [ ] A second test confirms the pre-existing behavior is unchanged: starting
      a daemon against a stale socket path (file exists, e.g. touched by the
      test, but nothing listening) succeeds normally.
- [ ] `cargo test -p loom-daemon` and `cargo clippy -p loom-daemon --all-targets -- -D warnings` pass.

### Affected Files

- `loom-daemon/src/ipc.rs` - add the connect-and-ping liveness probe and gate the `remove_file`/`bind` sequence in `IpcServer::run()` (~lines 67-92) behind it.
- `loom-daemon/src/main.rs` - no functional change expected (the `server.run().await?` at line 264 already propagates `Err` to a non-zero exit); revisit only if the chosen probe needs something from startup (e.g. a shared timeout constant) that's cleaner to define here.
- `loom-daemon/tests/integration_singleton_guard.rs` - new integration test file (or extend `loom-daemon/tests/integration_basic.rs`) using the existing `TestDaemon`/`TestClient` harness in `loom-daemon/tests/common/mod.rs` to cover the two-daemons and stale-socket scenarios.

### Test Plan

- [ ] Automated tests: new integration test(s) in `loom-daemon/tests/` — start daemon A via `TestDaemon::start()`, then attempt to start a second raw `loom-daemon` process with the same `LOOM_SOCKET_PATH`; assert the second process exits non-zero and daemon A remains reachable via `TestClient::connect(...).ping()`.
- [ ] Automated tests: a stale-socket regression test — create an empty file at the socket path (simulating a crashed daemon's leftover socket) and assert a fresh daemon starts normally and binds it.
- [ ] Unit test (if the probe is factored into a standalone function): direct test of the probe function against a real bound `UnixListener` (live) vs. a non-listened socket path (stale) vs. a missing path (absent), independent of full daemon startup.
- [ ] Manual verification: reproduce the original report — `loom-daemon` running, start a second `loom-daemon` process by hand pointed at the same default socket; confirm it prints a clear refusal message and exits non-zero, and `ps`/`gh`-visible sweep state shows only the original daemon's PID still owning the socket.

### Coordination

Root of the Phase-0 daemon-hardening stacked chain under epic #3809:
`#3806` (this issue, singleton guard) → `#3805` (IPC malformed-request error
frames) → `#3807` (`CancelSweep` mutex hold) → `#3808` (checkpoint-ingestion
phantom registry entries) → `#3802` (`token_name` always `"unknown"`). All
five touch `loom-daemon/src/`. Keep this issue's diff scoped to the singleton
guard in `ipc.rs` (+ its test) — do not fold in fixes belonging to the
sibling issues even if they're adjacent in the same files.

rjwalters:
Closes #3806

Part of #3809 (Phase 0 daemon-hardening stacked chain — this is the ROOT; #3805 stacks on this branch).

## Problem

`IpcServer::run()` (`loom-daemon/src/ipc.rs`) unconditionally `remove_file`d the socket and rebound at startup. Starting a second `loom-daemon` while one was already running unlinked the incumbent's socket out from under it — leaving the first daemon alive but orphaned (still holding its children, no socket) and the second owning the socket. Two daemons, silently, with split ownership of sweep children. Observed live: PID 74709 (old, orphaned) and 70355 (new) running simultaneously.

## Fix

Add a connect-and-ping liveness probe (`socket_has_live_listener`) that runs **before** `remove_file`:

- Connects to the socket and performs a `Ping`/`Pong` roundtrip, each step bounded by a 500ms timeout.
- **Connect failure / stale file / no Pong** → not a live listener → proceed exactly as before (best-effort `remove_file` + `bind`), preserving crash-recovery behavior for stale sockets.
- **Successful connect + `Pong`** → a live daemon owns the socket → `run()` returns `Err` with an operator-facing message naming the socket path and how to stop the running daemon. `main()`'s existing `server.run().await?` turns that into a non-zero process exit — **no `main.rs` changes required**.

No new dependency — reuses the existing `Ping`/`Pong` wire protocol. Coordinated takeover (killing the incumbent) is intentionally **out of scope** (deferred to sibling issues under the epic).

## Tests

- 4 lib unit tests for the probe: absent path, stale regular file, silent (non-responding) listener, and a live Ping/Pong daemon.
- New `tests/integration_singleton_guard.rs` using the existing `TestDaemon`/`TestClient` harness (temp sockets):
  - `test_second_daemon_refuses_and_first_survives` — daemon A owns the socket; a second daemon on the same socket exits non-zero, does not steal the socket, and daemon A still answers `Ping`.
  - `test_stale_socket_is_reclaimed` — a plain leftover file at the socket path is reclaimed normally (regression guard).

`cargo test` (all 416+ lib + integration tests pass), `cargo build --release`, and `cargo fmt --check` all clean.

**Clippy note:** `cargo clippy --all-targets -- -D warnings` reports a pre-existing set of 16 errors (format-args, expect/panic, too-many-lines, complex-type, etc.) from a newer local toolchain vs the CI-pinned one. Confirmed byte-for-byte identical on `origin/main` (the `ipc.rs` complex-type error is the pre-existing `setup_test_context` helper, shifted line 902→973 only because the new probe fn sits above it). My changed code introduces **zero** new clippy findings.

### DIFF
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -9,12 +9,67 @@ use crate::terminal::TerminalManager;
 use crate::types::{Event, Request, Response};
 use anyhow::Result;
 use chrono::Utc;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 use tokio::fs;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::net::{UnixListener, UnixStream};
+
+/// Bound on the singleton-guard liveness probe (#3806). Both the connect and
+/// the `Ping`/`Pong` roundtrip are individually capped at this duration so a
+/// hung or unresponsive peer can never stall daemon startup.
+const LIVENESS_PROBE_TIMEOUT: Duration = Duration::from_millis(500);
+
+/// Returns `true` if a live `loom-daemon` is currently listening on
+/// `socket_path` and actively servicing requests.
+///
+/// The probe connects to the socket and performs a `Ping`/`Pong` roundtrip:
+///
+/// - A connect failure (`ECONNREFUSED`, `ENOENT`, `ENOTSOCK`, permission
+///   error, …) means the socket is absent or stale — the file may linger from
+///   a crashed daemon but nothing is listening — so it is safe to remove and
+///   rebind. Returns `false`.
+/// - A successful connect **and** a `Pong` reply confirms a live daemon owns
+///   the socket. Returns `true`; the caller must refuse to start rather than
+///   unlink the path out from under the incumbent.
+///
+/// A connect that succeeds but never yields a `Pong` within
+/// `LIVENESS_PROBE_TIMEOUT` (e.g. an accept loop wedged before it services
+/// requests, or a non-daemon process squatting the path) is treated as "not a
+/// live, responsive daemon" and returns `false` — refusing to ever reclaim
+/// such a socket would be worse than rebinding it.
+async fn socket_has_live_listener(socket_path: &Path) -> bool {
+    let stream = match tokio::time::timeout(
+        LIVENESS_PROBE_TIMEOUT,
+        UnixStream::connect(socket_path),
+    )
+    .await
+    {
+        Ok(Ok(stream)) => stream,
+        // Connect refused/absent, or the connect itself timed out — not a
+        // live listener.
+        _ => return false,
+    };
+
+    let probe = async move {
+        let (reader, mut writer) = stream.into_split();
+        // Reuse the canonical Ping request shape so the probe stays in sync
+        // with the wire protocol.
+        let request_json = serde_json::to_string(&Request::Ping).ok()?;
+        writer.write_all(request_json.as_bytes()).await.ok()?;
+        writer.write_all(b"\n").await.ok()?;
+        writer.flush().await.ok()?;
+
+        let mut lines = BufReader::new(reader).lines();
+        let line = lines.next_line().await.ok()??;
+        let response: Response = serde_json::from_str(&line).ok()?;
+        Some(matches!(response, Response::Pong))
+    };
+
+    matches!(tokio::time::timeout(LIVENESS_PROBE_TIMEOUT, probe).await, Ok(Some(true)))
+}
 
 /// Get the current git branch for a given directory
 /// Returns None if not in a git repository or if the command fails
@@ -65,7 +120,23 @@ impl IpcServer {
     }
 
     pub async fn run(&self) -> Result<()> {
-        // Remove old socket
+        // Singleton guard (#3806): before touching the socket, probe whether a
+        // live daemon is already listening on it. Starting a second daemon used
+        // to unconditionally `remove_file` + rebind, silently orphaning the
+        // incumbent (still running, still holding its children, but with its
+        // socket unlinked). Refuse to start in that case; only a genuinely
+        // stale/absent socket is removed and rebound below.
+        if socket_has_live_listener(&self.socket_path).await {
+            anyhow::bail!(
+                "another loom-daemon is already listening on {} — refusing to start. \
+                 If you intended to replace it, stop the running daemon first \
+                 (e.g. `kill <pid>` or its shutdown path) and retry.",
+                self.socket_path.display()
+            );
+        }
+
+        // Remove old socket (best-effort; only reached when no live listener
+        // answered the probe above, i.e. the file is stale or absent).
         let _ = fs::remove_file(&self.socket_path).await;
 
         let listener = UnixListener::bind(&self.socket_path)?;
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVED ✅

Reviewed the Phase-0 root singleton guard rigorously — this gates unattended autonomous operation. Verified against every criterion:

**(a) Live vs stale discrimination — correct.** `socket_has_live_listener` connects and does a real `Ping`/`Pong` roundtrip. Connect failure (ENOENT/ECONNREFUSED/stale file) → `false` → safe remove+rebind (crash recovery preserved). Successful connect **and** `Pong` → `true` → refuse. The wedged-but-connected case (connect OK, no Pong within 500ms) is deliberately treated as *not a live responsive daemon* and returns `false` — a documented, defensible tradeoff (refusing to ever reclaim a non-answering socket would be worse than the alternative). All four probe unit tests pass locally.

**(b) Refusal is safe.** `anyhow::bail!` fires **before** the `fs::remove_file` + `bind`, so the incumbent's socket is never touched. `main()` returns `Result<()>`, so the Err propagates to a non-zero exit with an operator-facing message naming the socket path. The ctrl_c cleanup handler only unlinks on signal, not on this error path — incumbent socket stays intact.

**(c) Stale-socket path intact.** `test_stale_socket_is_reclaimed` confirms a leftover regular file at the socket path is reclaimed and the fresh daemon binds and answers Ping.

**(d) TOCTOU.** The only remaining window is two daemons *cold-starting simultaneously* with no incumbent — explicitly out of scope (coordinated takeover deferred). The targeted socket-steal bug (a second daemon started against a stable, responsive incumbent) is reliably prevented; no silent steal of a responsive daemon.

**(e) Integration coverage.** `test_second_daemon_refuses_and_first_survives` verifies the second process exits non-zero, the socket still exists, and daemon A still answers Ping (not orphaned). Both integration tests use temp sockets (`TestDaemon` → `TempDir`, and `tempfile::TempDir` directly) — never `~/.loom`.

**(f) Build/test/clippy.** `cargo test` (4 lib + 2 integration new tests) all pass locally; `cargo build --release` clean. Clippy: the 16 `-D warnings` errors are pre-existing (newer local toolchain vs CI-pinned) — `forge_parser.rs` is byte-for-byte unchanged by this PR (no diff vs origin/main), and `ipc.rs:973` is the pre-existing `setup_test_context` test helper (shifted down only because the new probe fn sits above it). The new `socket_has_live_listener` introduces **zero** new findings. CI Rust Linting & Build Check + Rust Unit Tests both **pass**.

Scope is clean: the single commit touches only `ipc.rs` + the new integration test. Merge state CLEAN.

Solid, well-tested, correctly-scoped fix. Approving.

PR #3816, opened 2026-07-23
fix(daemon): structured error frame on malformed IPC request (#3805)

### LINKED ISSUE #3805: loom-daemon IPC: malformed/unparseable requests are silently dropped (no error frame) instead of returning StructuredError

rjwalters:
Closes #3805

Part of #3809 Phase 0 (daemon hardening). **STACKED** — base branch is `feature/issue-3806` (the singleton-guard PR), NOT `main`. Merge #3806 first.

## Problem

A malformed IPC request (garbage JSON, a missing required field, or an unknown `type` tag) all fail at the single bare `?` on `serde_json::from_str::<Request>(&line)` in `handle_client` (`ipc.rs`). That `?` propagated the parse error up through the task's `Result`, which was only logged before the socket was dropped — the client saw an empty EOF, never a diagnostic. This bypassed the `Response::StructuredError` path entirely, since parse failures happen *before* `handle_request` is ever called.

## Fix

- `errors.rs`: new `DaemonError::ipc_parse_error(raw, source_err)` factory using the pre-existing (previously unused) `ErrorDomain::Ipc` / `ErrorCode::IPC_PROTOCOL_ERROR`, marked `recoverable(false)`. `serde_json::Error`'s `Display` already names the offending field/variant, so it's surfaced verbatim.
- `ipc.rs`: replace the bare `?` on the request deserialize with a `match` — on parse failure, write a `StructuredError` frame and `continue` the loop so the connection stays usable. The outer `?` on `lines.next_line()` (a socket I/O error, not a malformed payload) is intentionally left as-is (out of scope).

## Tests

- Rewrote `integration_basic::test_malformed_json` (which asserted the buggy silent-close behavior) to send three malformed shapes — unknown `type`, `CancelSweep` missing `grace_secs`, and a wrong-shape payload — and assert each returns `{"type":"StructuredError","payload":{"domain":"ipc","code":"IPC_PROTOCOL_ERROR",...}}` (ErrorCode is a newtype → serializes as a bare string), then a follow-up `ping()` proving the connection survives.

`cargo test` (all suites), `cargo build --release`, `cargo fmt --check` all pass. `cargo clippy -- -D warnings` is clean on the touched files; the sole remaining clippy error is the pre-existing `forge_parser.rs:91` `uninlined_format_args` already present on the base branch (left untouched per issue scope).

### DIFF
--- a/loom-daemon/src/errors.rs
+++ b/loom-daemon/src/errors.rs
@@ -356,6 +356,31 @@ impl DaemonError {
         .with_details(serde_json::json!({ "path": path }))
     }
 
+    // IPC errors
+    /// Creates an "IPC request parse failure" error.
+    ///
+    /// Raised when an incoming line on the IPC socket cannot be deserialized
+    /// into a [`crate::types::Request`] — covers garbage JSON, a missing
+    /// required field, and an unknown `type` tag alike (all three fail at the
+    /// single `serde_json::from_str::<Request>` call). `serde_json::Error`'s
+    /// `Display` already names the offending field/variant (e.g.
+    /// `missing field \`grace_secs\``, `unknown variant \`Nonsense\``), so we
+    /// surface it verbatim in the message.
+    ///
+    /// Marked non-recoverable: retrying the exact same malformed bytes will
+    /// fail identically, so the client must correct the payload rather than
+    /// retry blindly.
+    #[must_use]
+    pub fn ipc_parse_error(raw: &str, source_err: &serde_json::Error) -> Self {
+        Self::new(
+            ErrorDomain::Ipc,
+            ErrorCode::IPC_PROTOCOL_ERROR,
+            format!("failed to parse request: {source_err}"),
+        )
+        .recoverable(false)
+        .with_details(serde_json::json!({ "raw": raw }))
+    }
+
     // Internal errors
     /// Creates a "mutex poisoned" error
     #[must_use]
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -174,7 +174,22 @@ async fn handle_client(
     let mut lines = BufReader::new(reader).lines();
 
     while let Some(line) = lines.next_line().await? {
-        let request: Request = serde_json::from_str(&line)?;
+        // Parse the incoming frame. A malformed payload (garbage JSON, a
+        // missing required field, or an unknown `type` tag) is a per-request
+        // protocol error, NOT a fatal connection error: emit a structured
+        // error frame naming the serde failure and keep the connection usable
+        // for subsequent requests rather than silently dropping the socket.
+        let request: Request = match serde_json::from_str(&line) {
+            Ok(request) => request,
+            Err(parse_err) => {
+                let response =
+                    Response::StructuredError(DaemonError::ipc_parse_error(&line, &parse_err));
+                let response_json = serde_json::to_string(&response)?;
+                writer.write_all(response_json.as_bytes()).await?;
+                writer.write_all(b"\n").await?;
+                continue;
+            }
+        };
         log::debug!("Request: {request:?}");
 
         // SubscribeEvents is the only structurally-different request: it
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed the #3805 delta (stacked on `feature/issue-3806`; `gh pr diff` shows only errors.rs / ipc.rs / integration_basic.rs — no singleton code from #3806 touched).

### Verification
- **(a) Structured frame + connection survives** — The bare `?` on `serde_json::from_str::<Request>` is replaced by a `match`: on `Err`, it writes a `Response::StructuredError(DaemonError::ipc_parse_error(...))` frame then `continue`s the loop rather than propagating the error up and dropping the socket. The rewritten `test_malformed_json` sends three malformed shapes (unknown `type`, `CancelSweep` missing `grace_secs`, raw non-object) and a follow-up `ping()` — all pass, proving the socket stays usable. ✓
- **(b) Well-formed frame** — Test asserts `payload.domain == "ipc"` and `payload.code == "IPC_PROTOCOL_ERROR"` via `as_str()`, confirming the `ErrorCode` newtype serializes as a bare string the client can parse. ✓
- **(c) No valid-request regression** — The `Ok(request) => request` arm is behaviorally unchanged and the rest of `handle_client` is untouched; the full daemon suite (ping/pong, terminal, security, singleton-guard, etc.) passes. ✓
- **(d) Test asserts NEW behavior** — Old test asserted `result.is_err()` (the buggy silent-close); rewrite asserts structured frames + `ping()` survival, which would panic against the old drop-the-socket code. Genuinely exercises the fix. ✓
- **(e) No #3806 overlap** — Diff is confined to the three files; singleton guard lives in the base commit. ✓
- **(f) I/O `?` on `next_line()` left as-is** — Correct scope call: a socket I/O error (broken pipe) SHOULD close the connection, unlike a per-request payload error. ✓

### Local checks
- `cargo test` — all suites pass (incl. rewritten `test_malformed_json`).
- `cargo clippy -- -D warnings` — touched code clean; sole lib error is the pre-existing `forge_parser.rs:91` `uninlined_format_args` (confirmed identical on `origin/main`, out of scope). The `--all-targets` `type_complexity` note on `ipc.rs` `setup_test_context` is likewise pre-existing on both base and main.
- CI: Rust Lint/Build + Unit Tests both green.

Clean, well-tested fix that closes the IPC silent-drop. Approving. (Stacked — merge #3806 first.)

Issue #3747, opened 2026-07-22
Stacked-PR mode v2 item 4: detect-and-warn for out-of-set dependency references (items 1-3 shipped; diamonds/auto-detach won't-do)


rjwalters:
## Context

#3729 (PR #3742) landed stacked-PR mode **v1**: an operator-declared `--depends-on <parent>` dispatch param, `worktree.sh --base` override, block-the-subtree-on-parent-failure via the existing `sweep.issue.{N}.blocker` topic, and a **standalone, operator-invoked** `defaults/scripts/reconcile-stack.sh <child-pr> <parent-branch>` for the post-squash-merge rebase surgery.

This is a **multi-increment epic — it stays OPEN after any one PR merges.** Do not attempt all remaining items in a single PR.

**v2 item 1 (automated reconciliation on parent merge) — SHIPPED (PR #3752, merged 2026-07-22).** Wired `reconcile-stack.sh` into an automatic `_auto_reconcile_stacked_children()` step in `defaults/scripts/merge-pr.sh`, fired at the same choke point as `_reset_partial_increment_labels()`, best-effort and GitHub-only, with a safe/unsafe split keyed on the child issue's `loom:building` label. See `merge-pr.sh` lines ~499-631 and `defaults/.claude/commands/loom/sweep.md` "Stacked dependency (auto-reconciliation on parent merge)" for the shipped mechanic.

**v2 item 2 (pre-merge merge-ordering guard) — SHIPPED (PR #3794, merged 2026-07-22).** Added `_check_no_open_stacked_children()` (`merge-pr.sh` ~line 330) — a pre-merge guard that hard-blocks a parent merge (`exit 1`) whenever an open child PR still targets the parent branch, closing the "base-branch-deletion race" where `delete_branch_on_merge:true` deletes `feature/issue-<parent>` synchronously during the merge API call, before item 1's post-merge reconciliation pass can run. Ships with `--allow-stacked-children` (operator override), full `--dry-run` support, `defaults/scripts/tests/test-merge-pr-merge-ordering-guard.sh`, and doc updates in `defaults/.claude/commands/loom/sweep.md` (~line 1395) and `.loom/docs/daemon-reference.md` (~line 328).

[...]

PR #3817, opened 2026-07-23
fix(daemon): don't hold registry mutex across cancel grace window (#3807)

### LINKED ISSUE #3807: loom-daemon: CancelSweep holds the registry mutex for the entire SIGTERM→grace→SIGKILL window, blocking all other IPC
## Context — found reviewing the cancel path (2026-07-23)

The `CancelSweep` IPC handler locks the registry, then runs the **blocking** cancel escalation under that lock:

```rust
// loom-daemon/src/ipc.rs:836-839
let mut sr = sweep_registry
    .lock()
    .expect("Sweep registry mutex poisoned");
match sr.cancel(&sweep_id, std::time::Duration::from_secs(grace_secs)) { ... }
```

`SweepRegistry::cancel()` (`sweep_registry.rs:863`) does SIGTERM → a blocking 100ms-interval poll loop up to `grace` → SIGKILL → `wait()`, all synchronously while the registry mutex is held. So a cancel with, e.g., `grace_secs=30` **holds the registry lock for up to 30 seconds**, during which **every other IPC request that needs the registry — `DispatchSweep`, `ListSweeps`, `GetSweepStatus` — blocks**. A single slow-to-die sweep freezes the whole daemon's sweep surface.

This is a head-of-line-blocking / concurrency correctness issue, not just a latency nit: an operator cancelling one runaway sweep can't list or dispatch others until the grace window elapses.

- [ ] The blocking SIGTERM→grace-poll→SIGKILL→wait escalation does NOT hold the registry mutex for its full duration — e.g. take the lock only to read the target pid/pgid and flip state, release it during the poll/sleep, re-acquire briefly to record the terminal transition (or run the escalation on a spawned task and ack when done).
- [ ] While a cancel is escalating, `ListSweeps`/`GetSweepStatus`/`DispatchSweep` for OTHER sweeps proceed without blocking.
- [ ] The state transitions and the #3800 process-group kill / #3801 reap semantics are preserved (no regression).
- [ ] A test dispatches two sweeps, cancels one with a non-trivial grace, and asserts a concurrent `ListSweeps` returns promptly (not blocked for the grace window).

rjwalters:
Closes #3807

Part of #3809 Phase 0 (3rd in the stacked chain; base = `feature/issue-3805`, which stacks on `feature/issue-3806`).

## Problem
`Request::CancelSweep` (`ipc.rs`) locked the registry `Mutex` and held it across the entire blocking `SweepRegistry::cancel()` — SIGTERM → grace-poll → SIGKILL → reap. For `grace_secs = 30` this froze every other IPC request (`ListSweeps` / `GetSweepStatus` / `DispatchSweep`) for up to 30s.

## Fix (Curator's Option A — begin/finish split)
Split `cancel()` into three lock-scoped steps in `sweep_registry.rs`:

- `begin_cancel` → read pid/kind/liveness + SIGTERM the process **group**.
- `poll_cancel` → one liveness poll (reaps on exit), invoked per tick.
- `finish_cancel` → SIGKILL decision + reap + terminal `Exited` transition + lock release + events.

`handle_client` now intercepts `CancelSweep` and drives these via a new async `cancel_sweep_nonblocking`, awaiting the 100ms poll sleeps **unlocked** (`tokio::time::sleep`) so the registry mutex is free for the entire grace window (addresses the Curator's tokio-worker-blocking note).

## Preserved
- `CancelSweep` response contract (`sigkill_sent`, `was_running`, `pid`, synchronous completion) — unchanged, no MCP/doc coordination needed.
- #3800 process-group kill (`signal_sweep`/`send_group_signal`).
- #3801 reap semantics (`try_wait`/`reap_handle`, no zombie).
- Idempotent already-terminal early-return (`was_running: false`).
- `cancel()` retained as a synchronous composition of the three steps for direct/unit-test callers; all 6 existing cancel-path tests pass unmodified.

## Test
New `split_cancel_does_not_hold_lock_across_grace_window`: inserts a real trap-TERM child, drives the split orchestration on one thread with a 1s grace, and asserts a concurrent `get_status` on a *different* sweep returns in <400ms rather than blocking for the grace window (the core AC). Regression coverage for #3800/#3801 (SIGKILL escalation, process-group kill, reap-no-zombie) unchanged and green.

`cargo test` (all pass) / `cargo build --release` / `cargo fmt --check` clean. `cargo clippy -- -D warnings` clean except the pre-existing `forge_parser.rs:91` uninlined-format-args error already on origin/main (untouched).

### DIFF
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -4,7 +4,7 @@ use crate::event_bus::EventBus;
 use crate::forge_parser::parse_forge_events;
 use crate::git_parser;
 use crate::git_utils;
-use crate::sweep_registry::SweepRegistry;
+use crate::sweep_registry::{BeginCancel, SweepRegistry};
 use crate::terminal::TerminalManager;
 use crate::types::{Event, Request, Response};
 use anyhow::Result;
@@ -204,6 +204,30 @@ async fn handle_client(
             break;
         }
 
+        // CancelSweep is handled here rather than in the synchronous
+        // `handle_request` dispatcher (Issue #3807): its SIGTERM → grace-poll
+        // → SIGKILL escalation must NOT hold the registry mutex across the
+        // (possibly multi-second) grace window, or it would freeze every
+        // other IPC request (ListSweeps / GetSweepStatus / DispatchSweep).
+        // The async handler below re-acquires the lock only for the brief
+        // begin / poll / finish steps and `await`s the sleep unlocked.
+        if let Request::CancelSweep {
+            sweep_id,
+            grace_secs,
+        } = request
+        {
+            let response = cancel_sweep_nonblocking(
+                &sweep_registry,
+                &sweep_id,
+                Duration::from_secs(grace_secs),
+            )
+            .await;
+            let response_json = serde_json::to_string(&response)?;
+            writer.write_all(response_json.as_bytes()).await?;
+            writer.write_all(b"\n").await?;
+            continue;
+        }
+
         let response =
             handle_request(request, &terminal_manager, &activity_db, &sweep_registry, &event_bus);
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed as concurrency + async correctness (3rd in the #3809 Phase-0 stack, base `feature/issue-3805`; `gh pr diff` shows only #3807's delta).

### Core AC — mutex NOT held across grace window
**No MutexGuard is held across any `.await`.** In `cancel_sweep_nonblocking`, every registry lock is acquired inside a self-contained block expression (`{ let mut sr = sweep_registry.lock()…; sr.begin_cancel/poll_cancel/finish_cancel(…) }`) whose guard drops at the block's close, before the value flows out. Each `tokio::time::sleep(poll_interval).await` sits at the top of the loop body with no guard in scope. The lock is free for the entire grace window — begin/poll/finish take it only for microseconds each.

### Verification
- **(a) Lock released across sleeps** — confirmed structurally; guard scoping is airtight.
- **(b) #3800 process-group kill + #3801 reap preserved** — `begin_cancel` sends SIGTERM via `signal_sweep(…,15)` (group path), `finish_cancel` escalates via `signal_sweep(…,9)` + `reap_handle` (no zombie). Regression tests `cancel_terminates_whole_process_group_including_grandchild` and `cancel_escalates_to_sigkill_when_child_ignores_sigterm` green.
- **(c) SweepCancelled contract unchanged** — same 4 fields (sweep_id, pid, sigkill_sent, was_running), synchronous ack. No MCP/client breakage.
- **(d) No new race** — state stays `Running` (and per-issue lock held) until `finish_cancel`, so concurrent DispatchSweep for the same issue is deduped and get_status reports accurately. Concurrent double-cancel is idempotent/benign (SIGTERM/SIGKILL to reaped pid → false; reap/transition/release_lock idempotent) — no worse than the retained synchronous `cancel()`.
- **(e) handle_client interception correct** — mirrors the SubscribeEvents pattern (intercept before `handle_request`, respond, `continue`); locks always released, no deadlock.
- **(f) Idempotent already-terminal early-return preserved** — `BeginCancel::AlreadyTerminal` → `was_running: false`.
- **(g) New test genuinely proves the AC** — `split_cancel_does_not_hold_lock_across_grace_window` asserts a concurrent `get_status` on a DIFFERENT sweep returns <400ms during a 1s-grace trap-TERM cancel (would block ~1s under the old lock-held-throughout path).

### CI / local
- `cargo test`: 417 lib + all integration pass, incl. new test + all 8 cancel tests.
- `cargo build --release`: clean.
- `cargo clippy` (lib, matching CI): clean. (`--all-targets` surfaces a pre-existing complex-type warning in `ipc.rs:1110` test helper from commit f8aa06ab — untouched by this PR.)
- CI: Rust Linting + Unit Tests both green; merge state CLEAN.

Clean begin/finish split, well-documented, retained `cancel()` keeps the 6 existing tests unmodified. LGTM.

rjwalters:

Parent branch `feature/issue-3805` squash-merged, but this child's issue #3807 is still `loom:building` — a Builder likely has this branch checked out. Auto-reconciliation was **skipped** to avoid racing that in-progress work with an out-of-band `git rebase --onto` + `push --force-with-lease`.

**What happens next**: once issue #3807 is no longer `loom:building`, a subsequent parent-merge-triggered pass will reconcile this PR automatically. You can also reconcile it by hand now (from a clean checkout, only once the Builder has finished):

```
./.loom/scripts/reconcile-stack.sh 3817 feature/issue-3805
```

---
*Deferred by merge-pr.sh (#3747) at 2026-07-23T04:09:42Z*

Issue #3802, opened 2026-07-23
loom-daemon: registry token_name is always "unknown" — spawn-claude's selected account not captured


rjwalters:
## Context — found via live dispatch test (2026-07-23)

Token rotation itself **works** — the dispatched child's `spawn-claude.sh` log shows a real account was selected from the 7-account pool:

```
spawn-claude: using OAuth account 'agent3-2amlogic' (mode=random)
```

But the daemon registry reports the token as `"unknown"` for that same sweep:

```
← {"type":"SweepDispatched","payload":{"sweep_id":"...","pid":25746,"token_name":"unknown",...}}
← ListSweeps: issue=3780 pid=25746 token=unknown state=Running
```

So `list_sweeps` / `get_sweep_status` can't tell the operator which account a given sweep is burning — an observability gap for a multi-account pool (the whole point of the daemon path). The information exists (spawn-claude selects and logs it); the daemon just doesn't capture it into the registry.

## Root cause (confirmed by reading the code)

- `defaults/scripts/spawn-claude.sh:288-309` already selects the account and does everything short of handing it back to the daemon:
  ```bash
  export CLAUDE_CODE_OAUTH_TOKEN="$_token"
  export LOOM_TOKEN_NAME="$_name"
  log_info "spawn-claude: using OAuth account '$_name' (mode=$_mode)"
  ```
  `log_info` writes to **stderr** in the format `echo -e "${BLUE}[$(date -u ...)]${NC} $*" >&2`, i.e. the literal bytes are `\033[0;34m[<ts>]\033[0m spawn-claude: using OAuth account '<name>' (mode=<mode>)`.
- `loom-daemon/src/sweep_registry.rs::spawn_child()` (~647-778) launches `spawn-claude.sh` as a **separate detached process** with `cmd.stdout(Stdio::from(log_file)).stderr(Stdio::from(log_clone))` (lines 730-732) — both streams are already captured into the per-sweep log at `.loom/logs/sweep-issue-<N>.log`. But the function unconditionally returns:
  ```rust
  // spawn-claude.sh internally selects a token; we record "unknown" here
  // because the wrapper's selection is logged to the per-sweep log, not
  // exposed on stdout.
  Ok((child, "unknown".to_string()))
  ```
  (line 777).
- `dispatch()` (lines 336-429) threads that `token_name` straight into the `SweepInfo` it inserts into `self.entries` (lines 397-411) and into the returned `DispatchOutcome` (lines 422-428).
[...]

PR #3818, opened 2026-07-23
fix(daemon): don't reconstruct in-session /loom:sweep checkpoints as phantom entries (#3808)

### LINKED ISSUE #3808: loom-daemon: startup reconstruction ingests in-session /loom:sweep checkpoints, creating phantom daemon registry entries
## Context — observed at daemon startup (2026-07-23)

The daemon's startup registry reconstruction lists sweep checkpoints as a source:

```
// loom-daemon/src/sweep_registry.rs
//! - Registry reconstruction on startup from live processes + checkpoints.
//! - Sweep checkpoints under `.loom/sweep-checkpoint/issue-<N>.json` (#3373).
```

But `.loom/sweep-checkpoint/issue-<N>.json` files are written by the **in-session `/loom:sweep` skill** (the subagent-path lifecycle), **not** by daemon-dispatched sweeps. When the daemon starts while an in-session sweep is mid-flight, it reconstructs registry entries for those issues — sweeps the in-session path owns and is actively driving.

Observed live: starting the daemon during an active in-session sweep produced **5 phantom `recovered/Crashed` registry entries** for issues #3759/#3768/#3770/#3771/#3772 — exactly the issues that had live in-session checkpoints at that moment. The daemon had no child processes for them; it synthesized the entries purely from the in-session checkpoint files.

This conflates the two distinct sweep mechanisms (daemon detached-process vs. in-session subagent) and pollutes `list_sweeps` with entries the daemon doesn't actually own. It's the daemon↔in-session coexistence analog of the /sweep↔/sweep isolation work in #3768.

## Relation
- #3768 — concurrent-`/sweep` run-state isolation (this is the daemon↔in-session variant). Also the origin of `sweep-run-registry.sh` and the checkpoint `task_id`/`RUN_ID` field referenced below.
- #3801 — reaper/reconstruction state handling (adjacent).
- Part of epic #3809.

## Root cause (curator research, verified against `origin/main` @ `0d37f0e0`)

`SweepRegistry::reconstruct()` in `loom-daemon/src/sweep_registry.rs` runs two independent passes:

1. **Lock pass** (`~1162-1232`): scans `config.locks_dir()` (`.loom/locks/issue-<N>/owner.json`) and re-admits a `Running` entry for each lock whose `owner_pid` is alive. **Note a related quirk**: `acquire_lock()` (`~456-485`) writes `owner_pid: std::process::id()` — the **daemon's own PID**, not the spawned child's PID. After any real daemon restart the old daemon PID is gone by definition, so on restart this pass will almost always treat even genuinely daemon-dispatched, still-alive children as stale and drop their locks (`~1198-1204`) — daemon-dispatched sweeps fall through to the checkpoint pass too, just like in-session ones. This is directly relevant context for whichever fix option is chosen (see below) — don't assume "the lock pass already handles daemon-owned sweeps correctly" when designing the checkpoint-pass fix.
2. **Checkpoint pass** (`~1234-1284`): scans `config.checkpoint_dir()` (`.loom/sweep-checkpoint/issue-<N>.json`) and synthesizes a `Crashed` entry (`sweep-issue-<N>-recovered-<ts>`) for every checkpoint file, *unless* the lock pass already produced a `Running`/`Pending` entry for the same issue number (`~1254-1261`, the `already_running` check). Since `.loom/sweep-checkpoint/` is written **exclusively** by `defaults/scripts/sweep-checkpoint.sh` on behalf of the `/loom:sweep` skill (see the schema/header comment in that script, and the "Write timing" section of `defaults/.claude/commands/loom/sweep.md`), and the skill runs identically whether it was launched interactively (in-session) or as a daemon-spawned detached child (`spawn-claude.sh` → `claude -p "/loom:sweep N"`, per the module docstring `~12-14`), **the checkpoint pass cannot currently tell which mechanism launched the process that wrote a given checkpoint.** It admits a phantom `Crashed` entry for any checkpoint whose issue isn't already `Running` in the (frequently-empty, per point 1) lock-derived entry set — including checkpoints from perfectly healthy in-session sweeps the daemon never dispatched and has no locks for at all.

Net effect: the checkpoint pass's phantom-entry synthesis is essentially unconditional for in-session sweeps (no lock ever exists for them) and frequently fires for daemon-owned sweeps too (because of the PID quirk in point 1). The existing test `reconstruct_admits_orphan_checkpoints_as_crashed` (`~2459-2473`) locks in exactly this current (overly broad) behavior and will need to change alongside the fix.

`TERMINAL_RETENTION_SECS` (`~79`, 1 hour) governs how long a terminal entry (including these phantom `Crashed` ones) stays visible in `list_sweeps` before GC — so a phantom entry from a daemon restart during a long in-session sweep can pollute the registry for up to an hour even without any further daemon activity.

## Implementation guidance — options to evaluate

Any of these avoid re-admitting in-session-owned checkpoints as daemon registry entries. Pick the option that best balances correctness against the genuine "recover a daemon-dispatched sweep whose daemon crashed and restarted" use case documented at `~31-38` (the docstring's "Forge as source of truth" section) — the fix must NOT regress that case, only exclude in-session ownership from it:

1. **Daemon-scoped run marker, checked in `reconstruct()`.** When the daemon dispatches a sweep (`dispatch`/`spawn_child`), write a small daemon-owned marker file distinct from `.loom/sweep-run/` (which is the in-session peer registry from #3768's `sweep-run-registry.sh` and is *not* daemon-owned) — e.g. under a new `.loom/daemon-sweep-run/<sweep_id>.json` alongside the existing lock. In the checkpoint pass, only synthesize a `Crashed` entry for a checkpoint's issue if a (possibly stale/dead-owner) daemon-scoped marker for that issue exists — i.e., the daemon has independent evidence it once dispatched this issue — rather than trusting the shared checkpoint dir alone. This doesn't depend on fixing the `owner_pid` quirk in point 1, since the marker's mere presence (regardless of liveness) is the daemon-ownership signal, not its PID.
2. **Tag checkpoints with an owning-mechanism field.** Extend the checkpoint schema (`defaults/scripts/sweep-checkpoint.sh`'s `write` command) with an optional field (e.g. `"dispatched_by": "daemon"` or `"session"`), populated from an env var the daemon sets when spawning children (`spawn_child` already forwards an experiments allowlist at `~90-94` — the same pattern applies) and left absent for interactive sessions. `reconstruct()`'s checkpoint pass only admits entries where the field is present and `== "daemon"`; absent/`"session"` are skipped outright. Requires touching the shared checkpoint writer/schema (cross-cuts the in-session skill, so `defaults/.claude/commands/loom/sweep.md` and `defaults/scripts/sweep-checkpoint.sh` are in scope too, not just the Rust side).
3. **Live daemon-child PID only.** Restrict the checkpoint pass to issues where a live PID is independently known (the retained `children: BTreeMap<SweepId, Child>` map, `~218`, or a corrected non-self `owner_pid` in the lock). Simpler, but forecloses the intended "daemon crashed and restarted; recover from checkpoint" case entirely (nothing is live after a daemon crash by construition) — likely too narrow relative to the docstring's stated intent unless that intent is itself being revisited here.

Options 1 and 2 both preserve genuine daemon-crash-recovery while fixing the in-session false positive; option 3 is simplest but conflicts with existing documented intent (flag this trade-off explicitly if selected). Whichever is chosen, also decide whether to fix the `acquire_lock()` `owner_pid` quirk (point 1) in the same PR or file a fast-follow — it currently undermines the lock pass's ability to recognize genuinely-live daemon-owned sweeps across a restart, which is adjacent but distinct from this issue's phantom-entry-for-in-session-sweeps bug.

## Affected Files
- `loom-daemon/src/sweep_registry.rs` - `reconstruct()` (`~1162-1288`) checkpoint-pass logic; `acquire_lock()` (`~456-485`) if the `owner_pid` quirk is addressed in the same PR; module docstring (`~22`, `~37`) update to describe the corrected ownership semantics; existing test `reconstruct_admits_orphan_checkpoints_as_crashed` (`~2459-2473`) needs updating to match the corrected behavior, plus new test(s) per the plan below.
- `defaults/scripts/sweep-checkpoint.sh` - only if Option 2 (checkpoint tagging) is selected; extend the `write` schema with an owning-mechanism field.
- `defaults/.claude/commands/loom/sweep.md` - only if Option 2 is selected; document the new field and ensure the in-session skill never sets `dispatched_by=daemon`.

## Test Plan
- [ ] New/updated Rust unit test in `loom-daemon/src/sweep_registry.rs`: with an in-session-style checkpoint present (`.loom/sweep-checkpoint/issue-<N>.json`, no corresponding lock, no daemon-scoped marker/tag) and no live daemon child, `reconstruct()` admits **zero** entries for that issue — replaces/narrows the current `reconstruct_admits_orphan_checkpoints_as_crashed` assertion.
- [ ] New Rust unit test: with a daemon-dispatched-style checkpoint present (marker/tag/whatever the chosen option produces) and the process gone, `reconstruct()` still admits a `Crashed`/`recovered` entry — i.e., genuine daemon-crash recovery is preserved, not just narrowed to nothing.
- [ ] `cargo test -p loom-daemon sweep_registry` (or repo-equivalent invocation) passes.
- [ ] Manual/integration verification: start the daemon while an in-session `/loom:sweep` run has a live checkpoint for some issue N; confirm `mcp__loom__list_sweeps` (or the equivalent IPC call) shows no entry for N.

rjwalters:
## Summary

Fixes the daemon's startup registry reconstruction ingesting **in-session `/loom:sweep` checkpoints**, which created phantom `recovered/Crashed` registry entries for sweeps the daemon never dispatched (observed: 5 phantom entries when starting the daemon during an active in-session sweep). Part of epic #3809 Phase 0.

## Root cause (two parts)

1. **No daemon-ownership signal in the checkpoint pass.** `.loom/sweep-checkpoint/` is written by the shared `/loom:sweep` skill regardless of launch mechanism, so `reconstruct()` synthesized a `Crashed` entry for every checkpoint not already `Running` — including healthy in-session sweeps.
2. **`acquire_lock()` stored the daemon's own PID** rather than the spawned child's. After any daemon restart the daemon PID is gone, so the lock pass treated even genuinely-live daemon-dispatched children as stale and dropped them to the (false-positive) checkpoint pass.

## Fix — lock-as-daemon-ownership-marker

Chosen as the cleanest of the curator's three options: the lock (`.loom/locks/issue-<N>/`, written **exclusively** by `dispatch`) is already a crash-surviving daemon-ownership signal. No new marker file, no checkpoint-schema change (no cross-cut into `defaults/`).

- **`record_child_pid_in_lock`**: after spawn, rewrite the lock `owner.json` with the child's PID (acquire runs before spawn, so it stamps the daemon PID provisionally). The lock pass now admits a live daemon-dispatched child as `Running` across a restart.
- **Checkpoint pass gated on daemon ownership**: only synthesize a `Crashed` entry for a checkpoint whose issue had a daemon-owned lock whose owner PID is now dead. In-session sweeps never write a lock, so their checkpoints are skipped — no phantom entry. Genuine daemon-crash recovery is preserved because the lock survives a crash (removed only on clean release).

## Tests

- Replaced `reconstruct_admits_orphan_checkpoints_as_crashed` with `reconstruct_skips_in_session_checkpoints_without_lock` — asserts **zero** entries admitted for a lockless (in-session) checkpoint.
- Added `reconstruct_recovers_daemon_owned_checkpoint` — a checkpoint with a dead-owner daemon lock IS recovered as `Crashed` (genuine daemon-crash recovery preserved).
- `reconstruct_admits_live_lock_owners` still covers the live-child (`Running`) case.

`cargo test` (418 lib + all integration) passes; `cargo build --release`, `cargo clippy` (no new warnings), `cargo fmt --check` clean. The pre-existing `forge_parser.rs:91` clippy error on origin/main is left untouched.

## Stacking

**Base: `feature/issue-3807`** (4th in the Phase-0 chain: #3806 → #3805 → #3807 → this). `feature/issue-3808` is the base for #3802.

Closes #3808
Part of #3809 Phase 0

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -34,7 +34,13 @@
 //! Recovery on restart relies on:
 //!
 //! - Live process detection (`kill(pid, 0)`).
-//! - Sweep checkpoints under `.loom/sweep-checkpoint/issue-<N>.json` (#3373).
+//! - Sweep checkpoints under `.loom/sweep-checkpoint/issue-<N>.json` (#3373),
+//!   but **only for daemon-owned sweeps**: `.loom/sweep-checkpoint/` is shared
+//!   with the in-session `/loom:sweep` path, so a checkpoint is recovered only
+//!   when a daemon-owned lock (`.loom/locks/issue-<N>/`, written exclusively by
+//!   `dispatch`) also existed for that issue. This keeps the daemon from
+//!   ingesting phantom entries for in-session sweeps it never dispatched
+//!   (#3808). See [`SweepRegistry::reconstruct`].
 //! - Forge labels (`loom:issue` vs `loom:building`).
 
 use crate::event_bus::EventBus;
@@ -43,7 +49,7 @@ use crate::types::{Event, SweepId, SweepInfo, SweepKind, SweepOutcome, SweepStat
 use anyhow::{anyhow, Context, Result};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 #[cfg(unix)]
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
@@ -389,6 +395,20 @@ impl SweepRegistry {
         // Retain the handle so the reaper can `try_wait()` it (Issue #3801).
         self.children.insert(sweep_id.clone(), child);
 
+        // Record the spawned child's PID in the lock (Issue #3808). The lock's
+        // owner.json is written provisionally at `acquire_lock` time with the
+        // daemon's own PID (the child does not exist yet), but the value that
+        // matters for post-restart reconstruction is the *child's* PID: the
+        // daemon PID is gone after any restart, so keeping it would make even a
+        // still-live daemon-dispatched child look stale in `reconstruct()`'s
+        // lock pass. Rewrite `owner_pid` now that the child exists.
+        if let Err(e) = self.record_child_pid_in_lock(issue_number, pid) {
+            log::warn!(
+                "failed to record child pid {pid} in lock for issue #{issue_number} \
+                 (reconstruct may treat it as stale after a daemon restart): {e}"
+            );
+        }
+
         // 6. Record the entry. The model is carried on the registry entry
         //    (#3482, Phase 3a observability) so `list_sweeps` /
         //    `get_sweep_status` can report which model a sweep runs. Empty
@@ -464,6 +484,11 @@ impl SweepRegistry {
             Ok(()) => {
                 let owner = LockOwner {
                     issue,
+                    // Provisional: the child does not exist yet. `dispatch`
+                    // rewrites this with the spawned child's PID via
+                    // `record_child_pid_in_lock` once the child is running
+                    // (Issue #3808), so `reconstruct()` can recognise a live
+                    // daemon sweep after a restart.
                     owner_pid: std::process::id(),
                     acquired_at: Utc::now().to_rfc3339(),
                     sweep_id: sweep_id.to_string(),
@@ -482,6 +507,34 @@ impl SweepRegistry {
                 Err(anyhow!("failed to acquire lock for issue #{issue} at {}: {e}", lock.display()))
             }
         }
+    }
+
+    /// Rewrite the lock's `owner.json` so `owner_pid` records the spawned
+    /// sweep child's PID rather than the daemon's own PID (Issue #3808).
+    ///
+    /// `acquire_lock` runs *before* the child is spawned, so it can only
+    /// stamp `std::process::id()` (the daemon) provisionally. After a real
+    /// daemon restart that PID is gone by definition, which previously made
+    /// `reconstruct()`'s lock pass treat every daemon-dispatched sweep as
+    /// stale — dropping the lock and (before #3808) synthesizing a spurious
+    /// `Crashed` entry even for a child that was still alive. Storing the
+    /// child PID lets the lock pass admit a genuinely-live child as `Running`
+    /// across a restart. The rest of the owner record is preserved.
+    fn record_child_pid_in_lock(&self, issue: u32, child_pid: u32) -> Result<()> {
+        let owner_path = self
+            .config
+            .locks_dir()
+            .join(format!("issue-{issue}"))
+            .join("owner.json");
+        let existing = std::fs::read_to_string(&owner_path)
+            .with_context(|| format!("read lock owner {}", owner_path.display()))?;
+        let mut owner: LockOwner =
+            serde_json::from_str(&existing).context("parse lock owner.json")?;
+        owner.owner_pid = child_pid;
+        let owner_json = serde_json::to_string_pretty(&owner).context("serialize lock owner")?;
+        std::fs::write(&owner_path, owner_json)
+            .with_context(|| format!("write lock owner {}", owner_path.display()))?;
+        Ok(())
     }
 
     /// Release the lock dir for an issue (idempotent).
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed the #3808 delta only (stacked on `feature/issue-3807`; `gh pr diff 3818` scoped correctly). Verified all seven review criteria against the code and tests in `loom-daemon/src/sweep_registry.rs`:

**(a) CORE — no phantom pollution.** The checkpoint pass now gates on `if !daemon_owned_dead.contains(&issue) { continue; }`. In-session `/loom:sweep` runs never write `.loom/locks/issue-<N>/`, so their checkpoints admit **zero** entries. Confirmed by `reconstruct_skips_in_session_checkpoints_without_lock` (asserts `admitted == 0` and an empty registry).

**(b) Genuine daemon-crash recovery preserved.** A daemon-owned lock with a dead owner PID inserts the issue into `daemon_owned_dead` in the lock pass, so its checkpoint is still synthesized as `Crashed`. The lock survives a crash (removed only on clean `release_lock`), so the ownership signal persists. Confirmed by `reconstruct_recovers_daemon_owned_checkpoint`.

**(c) `record_child_pid_in_lock` correctness / no race.** Called from `dispatch` after `spawn_child`, ordered correctly (acquire_lock → spawn → record). `dispatch` holds `&mut self` and the lock dir was just created by this same call, so there is a single writer — no race, no corruption path. Read→modify→write preserves the rest of the owner record.

**(d) Live child across restart admitted as Running.** The lock pass admits a live `owner_pid` (now the child PID) as `Running` via `is_pid_alive`. Confirmed by `reconstruct_admits_live_lock_owners`.

**(e) Crash-survival intact.** `release_lock` is unchanged; `reconstruct` removes only stale (dead-owner) locks. No new removal path on the crash-survival route.

**(f) Test coverage.** All three cases covered: lockless→0 admitted, dead-owner-daemon-lock+checkpoint→recovered `Crashed` (+ stale lock cleaned up), live-lock-owner→`Running`.

**(g) No base regression.** Full suite green: 418 lib tests + all integration (singleton guard, security, factory reset, etc.) pass.

**Verification run locally:**
- `cargo test` — 418 lib + integration all pass (4 reconstruct tests green)
- `cargo build --release` — clean
- `cargo clippy --lib` — touched file `sweep_registry.rs` is warning-free; the only remaining warning is the pre-existing `forge_parser.rs:91` uninlined-format-args, untouched on origin/main as noted in the PR body.

**CI:** Rust Linting & Build Check + Rust Unit Tests both pass; Installer/MCP checks path-filtered (skipping). Merge state CLEAN.

Minor non-blocking note: `record_child_pid_in_lock` writes `owner.json` in place (not write-temp-then-rename), consistent with the existing `acquire_lock` write. A crash mid-write fails conservatively — parse fails → lock treated as stale → checkpoint skipped (no phantom), at worst a missed recovery in a vanishingly small window. Acceptable; no change requested.

The lock-as-daemon-ownership-marker approach is the right call — no new marker file, no checkpoint-schema cross-cut into `defaults/`. LGTM.

rjwalters:

Parent branch `feature/issue-3807` squash-merged, but this child's issue #3808 is still `loom:building` — a Builder likely has this branch checked out. Auto-reconciliation was **skipped** to avoid racing that in-progress work with an out-of-band `git rebase --onto` + `push --force-with-lease`.

**What happens next**: once issue #3808 is no longer `loom:building`, a subsequent parent-merge-triggered pass will reconcile this PR automatically. You can also reconcile it by hand now (from a clean checkout, only once the Builder has finished):

```
./.loom/scripts/reconcile-stack.sh 3818 feature/issue-3807
```

---
*Deferred by merge-pr.sh (#3747) at 2026-07-23T04:13:40Z*

Issue #3780, opened 2026-07-22
guide.md: work-finding queries use invalid gh label negation (always empty); doc-maintenance phase broken on cron; unblock flow violates the loom:issue gate


rjwalters:
## Problem

`defaults/.claude/commands/loom/guide.md` — the Guide role prompt — has three defects that make its core loops non-functional, plus a label-gate violation.

### 1. Work-finding queries silently return zero issues (invalid gh label negation)

`guide.md:80` (also 83, 136-142):

```bash
gh issue list --label "loom:issue" --label "!loom:building" --state open ...
```

`gh` ANDs `--label` values; `!loom:building` is treated as a literal label name that no issue carries, so **every one of these queries is an empty set** (verified live: the query returns 0 while the same query without the negation returns results). The cron-invoked Guide can never find work through them.

**Fix**: use `--search "-label:loom:building"` or jq post-filtering (the pattern `curator.md:111-114` already uses).

Note: the same invalid `--label "!..."` pattern also appears in `CLAUDE.md` ("As a Curator" / Curator Workflow section) and `defaults/CLAUDE.md` — fix those in the same pass.

### 2. Mandates a script that does not exist

`guide.md:285-295`: "**ALWAYS run stale detection with `--recover`** ... `./.loom/scripts/stale-building-check.sh --recover`". Verified: `stale-building-check.sh` exists nowhere in the repo. The same dead reference appears in `CLAUDE.md` (Quick fixes section), `loom.md:579`, and `loom-reference.md:332-338`. Every triage tick either errors here or the agent improvises.

**Fix**: restore/port the script, or point at live recovery tooling, or replace with the manual-verification fallback that already follows (guide.md:320-333).

### 3. Unblocking flow violates the Guide's own label gate

`guide.md:11`: "**NEVER add the `loom:issue` label to issues.**" vs `guide.md:519` (and 538):

```bash
gh issue edit "$number" --remove-label "loom:blocked" --add-label "loom:issue"
```

An issue can be `loom:blocked` *before* ever being human-approved (curator.md:341, 604 apply `loom:blocked` to pre-curation issues), so the unblock loop promotes never-approved issues straight into the Builder queue.

**Fix**: restore only the labels present before blocking (or add `loom:issue` only if the issue previously had it), and state the exception next to the gate policy.

### 4. Document Maintenance phase is broken by design on cron

[...]

PR #3819, opened 2026-07-23
fix(daemon): capture spawn-claude's selected account into registry token_name (#3802)

### LINKED ISSUE #3802: loom-daemon: registry token_name is always "unknown" — spawn-claude's selected account not captured

rjwalters:
Closes #3802

Part of #3809 Phase 0 (final / 5th hardening link). Stacked on #3808's branch (`feature/issue-3808`, which carries #3806 + #3805 + #3807 + #3808).

## Problem
`spawn_child()` in `sweep_registry.rs` hard-coded `token_name: "unknown"`, so `list_sweeps` / `get_sweep_status` could never report which OAuth account a dispatched sweep was using — an observability gap for the multi-account pool. The data already existed: `spawn-claude.sh` logs `spawn-claude: using OAuth account '<name>' (mode=<mode>)` and the daemon already captures the child's stderr into the per-sweep log.

## Fix (Approach A — parse the already-captured log)
`spawn_child()` now polls the per-sweep log after spawn and extracts the account name from the selection line, **anchored to this dispatch's header** (`sweep_id=<id>`) so a stale line from a prior dispatch in the same reused per-issue log is never misread. The captured name flows into the `SweepInfo` that `dispatch()` stores — `list_sweeps` / `get_sweep_status` report it automatically (zero changes there, as the Curator noted).

Chose Approach A over the file-handshake (Option B) to keep the change scoped to a **single Rust file** — no `spawn-claude.sh` edit, avoiding the `defaults/` ↔ `.loom/scripts/` duplication. The account name carries no ANSI codes (only the timestamp prefix does), so a plain substring scan suffices — no ANSI stripping.

## Preserved fallback
- `LOOM_SPAWN_NO_EXPORT` bypass (no account selected, nothing logged) still resolves to `"unknown"` — expected "nothing to report", not a regression.
- A `try_wait` early-exit keeps that case fast (child that exits without logging is detected immediately rather than waiting out the 5s capture window). `try_wait` caches the exit status, so the reaper's later `try_wait` still observes the exit.
- Capture never blocks longer than the timeout and never fails dispatch.

## Tests
- Unit: `parse_token_name_after` extraction, stale-line anchoring (stale line before the current header is ignored; the current selection wins once logged), empty/absent-marker → `None`.
- E2E: `dispatch_captures_selected_account_into_token_name` (fake spawn-claude logs the marker; asserts both `DispatchOutcome.token_name` and the stored `SweepInfo.token_name` equal `agent3-2amlogic`, mirroring the live-dispatch finding on issue #3780) and `dispatch_token_name_unknown_when_no_selection_logged`.

## Verification
- `cargo test` — 423 unit + all integration tests pass.
- `cargo clippy -- -D warnings` — only the **pre-existing** `forge_parser.rs:91` `uninlined_format_args` error (present on origin/main); no new lints from this change.

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -99,6 +99,99 @@ pub const EXPERIMENT_ENV_ALLOWLIST: &[&str] = &[
     "LOOM_TRANSCRIPT_ARCHIVE",
 ];
 
+/// Issue #3802: fallback `token_name` recorded when `spawn-claude.sh`'s
+/// account selection cannot be captured (e.g. the `LOOM_SPAWN_NO_EXPORT`
+/// bypass path, where the caller pre-exported `CLAUDE_CODE_OAUTH_TOKEN` and no
+/// account is selected at all — nothing to report, not a bug).
+pub const UNKNOWN_TOKEN_NAME: &str = "unknown";
+
+/// Issue #3802: the marker substring `spawn-claude.sh` logs immediately after
+/// selecting an OAuth account — `spawn-claude: using OAuth account '<name>'
+/// (mode=<mode>)` (see `defaults/scripts/spawn-claude.sh`). The daemon already
+/// captures the child's stderr into the per-sweep log, so `spawn_child` reads
+/// the log back and extracts `<name>` from the text following this marker. The
+/// account name itself carries no ANSI colour codes (only the timestamp prefix
+/// does), so a plain substring scan is sufficient — no ANSI stripping needed.
+const TOKEN_NAME_LOG_MARKER: &str = "using OAuth account '";
+
+/// Issue #3802: how long `spawn_child` polls the per-sweep log for the
+/// `TOKEN_NAME_LOG_MARKER` line before giving up and recording
+/// `UNKNOWN_TOKEN_NAME`. `spawn-claude.sh` selects and logs the account early
+/// (well before it `exec`s `claude`), so the line normally appears within a
+/// second. The poll also short-circuits the moment the child exits without
+/// having logged a selection (keeps no-selection fixtures fast), so this full
+/// window is only ever waited out in the pathological "child alive but never
+/// logged" case — capture then degrades gracefully to `UNKNOWN_TOKEN_NAME`
+/// and never blocks or fails dispatch.
+const TOKEN_NAME_CAPTURE_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Issue #3802: poll cadence for the `TOKEN_NAME_LOG_MARKER` scan.
+const TOKEN_NAME_CAPTURE_POLL_INTERVAL: Duration = Duration::from_millis(25);
+
+// ============================================================================
+// Token-name capture (Issue #3802)
+// ============================================================================
+
+/// Extract the OAuth account name `spawn-claude.sh` logged, from the per-sweep
+/// log `contents`, scanning only the region at/after this dispatch's header
+/// (`header_anchor`, e.g. `sweep_id=<id>`). Returns `None` when the marker /
+/// closing quote isn't present yet or the captured name is empty. Anchoring to
+/// the current header avoids picking up a stale selection line left by a
+/// previous dispatch in the same reused per-issue log.
+fn parse_token_name_after(contents: &str, header_anchor: &str) -> Option<String> {
+    // Use the LAST header occurrence: reruns append a fresh header, and the
+    // current child logs its selection after the most recent one.
+    let region_start = contents.rfind(header_anchor)?;
+    let region = &contents[region_start..];
+    let marker_at = region.find(TOKEN_NAME_LOG_MARKER)? + TOKEN_NAME_LOG_MARKER.len();
+    let after = &region[marker_at..];
+    let close = after.find('\'')?;
+    let name = &after[..close];
+    if name.is_empty() {
+        None
+    } else {
+        Some(name.to_string())
+    }
+}
+
+/// Poll `log_path` for the account-selection marker until it appears, the
+/// child exits, or `TOKEN_NAME_CAPTURE_TIMEOUT` elapses.
+///
+/// Returns the captured account name, or `UNKNOWN_TOKEN_NAME` when no
+/// selection was logged (the `LOOM_SPAWN_NO_EXPORT` bypass, a timeout, or a
+/// child that exited before logging). Never blocks longer than the timeout and
+/// never fails dispatch.
+///
+/// The `try_wait` early-exit keeps no-selection cases fast (a fixture that
+/// exits without logging is detected immediately rather than waiting out the
+/// full window). `try_wait` caches the exit status, so the reaper's later
+/// `try_wait` on the same handle still observes the exit.
+fn poll_token_name(child: &mut Child, log_path: &Path, header_anchor: &str) -> String {
+    let deadline = std::time::Instant::now() + TOKEN_NAME_CAPTURE_TIMEOUT;
+    loop {
+        if let Ok(contents) = std::fs::read_to_string(log_path) {
+            if let Some(name) = parse_token_name_after(&contents, header_anchor) {
+                return name;
+            }
+        }
+        // If the child has already exited without logging a selection, the
+        // line will never appear — do a final read (covers a log flushed right
+        // before exit) and stop, rather than waiting out the timeout.
+        if matches!(child.try_wait(), Ok(Some(_))) {
+            if let Ok(contents) = std::fs::read_to_string(log_path) {
+                if let Some(name) = parse_token_name_after(&contents, header_anchor) {
+                    return name;
+                }
+            }
+            return UNKNOWN_TOKEN_NAME.to_string();
+        }
+        if std::time::Instant::now() >= deadline {
+            return UNKNOWN_TOKEN_NAME.to_string();
+        }
+        std::thread::sleep(TOKEN_NAME_CAPTURE_POLL_INTERVAL);
+    }
+}
+
 // ============================================================================
 // Registry
 // ============================================================================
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed the `token_name` capture fix (Approach A — parse the already-captured per-sweep log; single-file change in `sweep_registry.rs`, +222/-5). Verified against every hardening criterion:

**(a) Log parse + dispatch-anchoring — correct.** `parse_token_name_after` scans only from the LAST `sweep_id=<id>` header via `rfind`, then finds the `using OAuth account '<name>'` marker after it. The header `==== loom-daemon dispatch: … sweep_id={sweep_id} … ====` is appended to the same `log_path` before spawn, and the anchor `format!("sweep_id={sweep_id}")` matches it byte-for-byte. The `ignores_stale_line_before_current_header` test proves a prior dispatch's selection line in a reused per-issue log is not misread.

**(b) Capture never blocks dispatch beyond the timeout, never fails dispatch — confirmed.** `poll_token_name` returns `String` (infallible), bounded by the 5s deadline with a 25ms poll cadence, and degrades to `UNKNOWN_TOKEN_NAME` on timeout/no-line. Worst-case 5s is only the pathological "child alive but never logs" case; happy path resolves in well under a second. Note: `dispatch()` holds the registry mutex across `spawn_child` — but it already holds it across a `gh issue edit` network call (`flip_label_to_building`), so this bounded poll is consistent with existing dispatch behavior, not a new hang class.

**(c) `try_wait` early-exit does not steal the reaper's status — confirmed.** `poll_token_name` only borrows `&mut child` and never consumes/removes it; the handle flows to `self.children` for the #3801 reaper. Rust std caches the reaped exit status, so the reaper's later `poll_liveness` → `try_wait` still observes `Ok(Some(status))`. No double-wait, no `<defunct>` zombie.

**(d) Fallback preserved — confirmed.** `LOOM_SPAWN_NO_EXPORT` / no-selection stays `"unknown"`; `dispatch_token_name_unknown_when_no_selection_logged` exercises the `try_wait` fast-exit path.

**(e) No regression to #3806/#3805/#3807/#3808 — confirmed.** Single-file diff; full `cargo test` (all integration + 423 unit) passes on the stacked branch.

**(f) `list_sweeps` / `get_sweep_status` need no change — confirmed.** The captured name flows into `SweepInfo.token_name`; the E2E test asserts both `DispatchOutcome.token_name` and the stored `SweepInfo.token_name` equal `agent3-2amlogic`.

**(g) Tests genuinely cover the cases — confirmed.** Extraction, stale-anchoring (before + after the current header logs), absent/empty-marker, E2E account-lands-in-registry, and no-selection.

**Local verification:** `cargo test` green (all suites + the 5 new tests); `cargo build --release` clean; `cargo clippy` shows zero lints in the touched `sweep_registry.rs` (the only two warnings — `forge_parser.rs:91` and `main.rs:104` — are pre-existing in untouched files). CI: Rust Linting & Build + Rust Unit Tests both pass; merge state CLEAN.

Minor, non-blocking: dispatch briefly serializes concurrent status/list queries under the mutex during the capture poll (typically <300ms). Acceptable given the pre-existing `gh` call already under that lock; worth revisiting only if dispatch latency ever matters.

LGTM. Approved.

rjwalters:

Parent branch `feature/issue-3808` squash-merged, but this child's issue #3802 is still `loom:building` — a Builder likely has this branch checked out. Auto-reconciliation was **skipped** to avoid racing that in-progress work with an out-of-band `git rebase --onto` + `push --force-with-lease`.

**What happens next**: once issue #3802 is no longer `loom:building`, a subsequent parent-merge-triggered pass will reconcile this PR automatically. You can also reconcile it by hand now (from a clean checkout, only once the Builder has finished):

```
./.loom/scripts/reconcile-stack.sh 3819 feature/issue-3808
```

---
*Deferred by merge-pr.sh (#3747) at 2026-07-23T04:15:48Z*

PR #3825, opened 2026-07-23
fix(daemon): close the dispatch→build contract (self-skip + orphaned claim + permissions)

### LINKED ISSUE #3823: loom-daemon dispatch self-skips: pre-flips loom:building before spawn, so the child /loom:sweep skips its own claim
**Part of #3809** (autonomous daemon mode) — a **critical blocker**: even explicit daemon dispatch does not reliably build issues, and Phase A (#3810) will hit the exact same bug.

## Discovered via live dogfood (2026-07-23)

Dispatched 3 disjoint issues (#3778, #3785, #3780) to the hardened daemon via `DispatchSweep`. All 3 spawned children on rotated tokens — but **2 of 3 exited in <2 min having built nothing**, still labeled `loom:building` (orphaned claims recovered by hand). Their own logs are explicit:

```
#3785 → skipped (already in flight: loom:building)
   "the loom:building claim was applied 84 seconds before this sweep's pre-flight"
#3780 → skipped (already in flight)
   "already being actively built by a daemon-dispatched sweep child"
```

## Root cause

`SweepRegistry::dispatch()` (`loom-daemon/src/sweep_registry.rs:435-542`) calls `flip_label_to_building()` (fn at `sweep_registry.rs:647`) at step 4 (`sweep_registry.rs:471-480`), flipping `loom:issue → loom:building` on the forge **before** it spawns the child at step 5 (`spawn_child()`, `sweep_registry.rs:793` — `cmd.arg("-p").arg("/loom:sweep {issue}")`, `sweep_registry.rs:834-836`). The spawned child then runs `/loom:sweep N`, whose per-issue pre-flight (Wave Lifecycle step 1, `defaults/.claude/commands/loom/sweep.md:1218-1226`) reads live forge labels and applies this rule verbatim:

> "If the issue already has `loom:building`, skip it — another shepherd or builder is working on it. Log a warning. Does NOT contribute to this wave." (`sweep.md:1224`)

The child sees the claim **the daemon just set on its behalf**, concludes another worker owns it, and **skips itself** — no worktree, no build, no PR. It's a race between the daemon's `gh issue edit` call and the child's own pre-flight read: whichever ordering happens to leave the label unset when a *given* child's pre-flight runs wins; every daemon-dispatched child loses by construction, since the daemon *always* flips the label before spawning.

### The "recoverable claim" gap (found while reading the reaper path)

`reap_once()` (`sweep_registry.rs:1212-1330`) only calls `restore_label_to_ready()` (`loom:building → loom:issue`, `sweep_registry.rs:675`) when a **sweep checkpoint exists** for the issue (`sweep_registry.rs:1255-1258`, the `Crashed` branch). A self-skipped child exits **cleanly** (exit code 0, "skip and log a warning" is not a crash) and never writes a `.loom/sweep-checkpoint/issue-<N>.json` (it never reached the Curator/Builder phase that writes one) — so it lands in the `else` branch (`Exited`, `sweep_registry.rs:1274-1290`) where `restore_label_to_ready()` is **never called**. This is exactly why the reporter had to reclaim #3780/#3785 by hand: today's self-skip path leaves an orphaned `loom:building` claim with no automatic recovery. Any fix for this issue needs to close this gap too, not just stop the self-skip.

## Why this blocks the epic

Phase A (#3810) is a poll loop that will flip `loom:building` and dispatch `/loom:sweep N` for each `loom:issue` — hitting this **exact** self-skip for every issue, every time. **This must be fixed before Phase A can work at all.**

## Fix options (evaluated)

**(A) Signal claim ownership to the child.** The daemon exports a marker env var (e.g. `LOOM_SWEEP_CLAIM_OWNED=<issue>` or the `sweep_id`) into the spawned child at `spawn_child()` (`sweep_registry.rs:864-878`, alongside the existing `LOOM_TERMINAL_ID` / `WORKSPACE_ENV` / experiment-env allowlist pattern at `sweep_registry.rs:905-911` — the same explicit-allowlist mechanism, not a blanket env copy). The child's pre-flight (`sweep.md:1218-1226`) checks this marker before applying the `loom:building` skip rule: if the marker names *this* issue, treat the existing `loom:building` as **this sweep's own claim** and proceed to build rather than skipping.

- Pro: preserves the daemon's early flip, which is presumably intended to dedupe concurrent `dispatch_sweep` calls for the same issue *before* a child process even exists (the flip happens synchronously inside `dispatch()`, ahead of any lock/registry check a second caller could race against).
- Con: two coordinated changes (Rust export + skill-side pre-flight logic), and the marker must not leak into the general `/loom:sweep N` entry point used outside daemon dispatch (an operator running `/loom:sweep N` manually on an issue that happens to be `loom:building` should still skip — self-ownership must be daemon-scoped, not just "trust any `loom:building`").

**(B) Daemon does not pre-flip; child's own Builder phase claims the label.** Remove (or defer) the `flip_label_to_building()` call in `dispatch()` step 4. The sweep skill's Builder phase already claims `loom:issue → loom:building` itself per the documented Builder Workflow (CLAUDE.md "Builder Workflow" step 2) — so the daemon simply stops racing its own child.

- Pro: much simpler; deletes code instead of adding a coordination channel.
- Con: **the early flip is not obviously redundant.** Check `acquire_lock()` (`sweep_registry.rs:569-603`, the `mkdir`-based atomic lock at `.loom/locks/issue-<N>/`) — this **already** dedupes concurrent `dispatch()` calls for the same issue at step 3, *before* the label flip at step 4. Two racing `dispatch_sweep` calls for the same issue would collide on `acquire_lock` (`Err(AlreadyExists)`) regardless of whether the label is flipped early. This suggests the in-process lock is already sufficient for the daemon's own dedup, and the early label flip is solving a **different** problem: visibility to *external* actors (a human running `gh issue list --label loom:issue`, a concurrent GH Actions cron tick, or a separately-dispatched `/loom:sweep` from another terminal) who have no visibility into the daemon's in-memory lock/registry and rely on the forge label as the sole coordination signal. Removing the early flip re-opens a **narrower but real** external race: between `dispatch()` returning and the child's own Builder-phase claim, the issue is still labeled `loom:issue` and forge-visible as available, so a second, non-daemon actor could pick it up.

**Recommendation: Option A**, with the flip retained. The daemon's early flip is defensible as the mechanism that makes the claim visible to non-daemon actors immediately at dispatch time (not after the child boots, curates, and reaches the Builder phase — a gap of potentially minutes under the current `/loom:sweep` lifecycle). Removing it (Option B) trades a guaranteed self-skip bug for a reintroduced, real external race window that the daemon's local lock cannot cover. Option A fixes the self-skip without giving up the early-visibility property; it costs one small, well-precedented change on each side of the spawn boundary (the codebase already threads similar scoped signals through `spawn_child`, e.g. `LOOM_TERMINAL_ID`, the experiment-env allowlist).

Whichever option is implemented, the "recoverable claim" gap above (self-skip leaves an orphaned `loom:building` with no checkpoint, so `reap_once()` never restores the label) should be fixed alongside it — a genuinely dead/skipped child's claim must not require manual reclaim.

## Acceptance criteria

- [ ] A daemon-dispatched `/loom:sweep N` (via `SweepRegistry::dispatch`) **builds** issue N end-to-end (worktree created, PR opened) — it does NOT self-skip on the daemon's own pre-dispatch `loom:building` claim. Verified live (not just unit-tested): dispatch a real issue through the daemon and confirm a PR is produced, not a skip log line.
- [ ] Two concurrent `dispatch()` calls for the same issue are still deduped — no double-build, no two children racing the same worktree/branch. (Today this is `acquire_lock()`, `sweep_registry.rs:569`; any fix must not weaken it.)
- [ ] A dispatched child that self-skips, crashes, or otherwise dies before reaching the Builder phase leaves the `loom:building` claim **recoverable** — either the reaper (`reap_once()`) restores `loom:issue` on a clean-exit-without-checkpoint outcome that correlates with a self-skip, or an equivalent automatic mechanism closes the gap described above. Manual `restore_label_to_ready` reclamation should not be required for this failure mode.
- [ ] If Option A is implemented: the ownership marker is scoped to daemon-dispatched children only — an operator-invoked `/loom:sweep N` outside the daemon (manual terminal, GH Actions cron) still honors the existing `loom:building` skip rule unmodified.
- [ ] Regression test coverage: a `loom-daemon` integration/unit test that dispatches a sweep against a fake child (mirroring the existing `dispatch_*` test patterns around `sweep_registry.rs:1925-2170`) and asserts the label-flip + spawn ordering no longer produces a self-skip signal; a `sweep.md` scenario/doc update reflecting the new pre-flight rule.

## Scope note — co-implemented with #3824

This issue and #3824 (daemon-spawned children hit permission denials — `--dangerously-skip-permissions` missing from the `spawn-claude.sh` invocation) were both discovered in the same 2026-07-23 dogfood run and **both modify the same spawn path** (`SweepRegistry::spawn_child` / `dispatch`, `sweep_registry.rs:435-920`, and the child-side `/loom:sweep` entry conditions). Implement them together in **one consolidated daemon-dispatch-contract PR** rather than two separate PRs that would otherwise touch overlapping code and risk merge conflicts / re-review churn. The Builder should address both issues' acceptance criteria in that single PR and reference both issue numbers.

## Affected Files

- `loom-daemon/src/sweep_registry.rs` — `dispatch()` (~435-542), `flip_label_to_building()` (~647), `spawn_child()` (~793-920), `reap_once()` (~1212-1330), `restore_label_to_ready()` (~675)
- `defaults/.claude/commands/loom/sweep.md` — per-issue pre-flight step 1 (~1218-1226), Stage -1 backend detection (~633+)

## Related

#3809 (epic), #3810 (Phase A — blocked by this), #3801 (reaper — correctly marked the skipped children Exited), #3824 (co-implemented — permission-denial companion finding from the same dogfood run)

### LINKED ISSUE #3824: loom-daemon-spawned sweep children hit permission denials (no --dangerously-skip-permissions), breaking Stage -1 setup
**Part of #3809** (autonomous daemon mode). For unattended operation, daemon-spawned children must run non-interactively without permission prompts — they currently don't.

## Discovered via live dogfood (2026-07-23)

The daemon-dispatched `/loom:sweep N` children reported permission denials that broke their Stage -1 setup (from their own logs):

> *"the sweep run-registry script (`sweep-run-registry.sh new`) and the daemon probe (`mcp__loom__list_sweeps`) were both denied by the permission layer, so I couldn't register a RUN_ID or complete Stage -1 daemon detection."*
> *"the daemon MCP probe, the run-registry script, and `.loom/` writes were all permission-denied in this session, so Stage -1 fell through to the subagent path."*

## Root cause (likely)

The daemon spawns the child via `spawn-claude.sh` → `exec claude …`, but the spawn path does **not** appear to pass `--dangerously-skip-permissions` (grep of `defaults/scripts/spawn-claude.sh` and `loom-daemon/src/sweep_registry.rs` finds none). In a detached `claude -p` (non-interactive) child, a permission prompt has no human to approve → it auto-denies, so `.loom/` writes, script runs, and MCP probes are blocked. An autonomous daemon child cannot function under interactive permission gating.

- [ ] Daemon-spawned sweep children run with permissions configured for unattended operation (e.g. `--dangerously-skip-permissions`, or a pre-seeded allow-list settings profile) so `.loom/` writes, sweep scripts, and MCP probes are not denied.
- [ ] Stage -1 (run-registry, daemon detection) completes in a daemon-spawned child.
- [ ] The permission posture is scoped/documented (not a blanket weakening beyond what unattended sweeps require).
- [ ] Verified: a dispatched child completes Stage -1 and its lifecycle without permission-denial stalls.

## Related
#3809 (epic), the self-skip bug (companion finding from the same dogfood), #3789 (Stage -1 pool probe).

rjwalters:
Consolidated fix for the two Phase-0.5 blockers (plus a curator-found third bug) that prevented daemon-dispatched sweeps from actually building issues. All three live on the same `SweepRegistry` child-spawn path, so they ship in one PR. Discovered via the 2026-07-23 live dogfood.

### #3823 — self-skip (Option A: claim-ownership marker)
`dispatch()` flips `loom:issue → loom:building` on the forge **before** spawning `claude -p /loom:sweep N` (for immediate external visibility of the claim). The child's own `/loom:sweep` pre-flight then read that label and skipped issue N as "already being built by someone else" — self-skipping the daemon's OWN claim, producing no worktree, no build, no PR.

- `spawn_child()` now exports **`LOOM_SWEEP_CLAIM_OWNED=<issue>`** into the child (alongside `LOOM_TERMINAL_ID` / the experiment-env allowlist).
- `defaults/.claude/commands/loom/sweep.md` per-issue pre-flight step 1: when the marker names the current issue, an existing `loom:building` is recognised as the sweep's own daemon claim → **proceed to build** (do not skip).
- **Scoped to daemon dispatch only** — an operator-run `/loom:sweep N` never sets the marker, so the manual-terminal skip rule is unchanged.
- The early flip is **retained** (external-visibility purpose; `acquire_lock` already dedupes same-process dispatch).

### #3823b — orphaned-claim recovery (reaper gap, curator-found)
`reap_once()` only restored `loom:building → loom:issue` on a `Crashed` (checkpoint-present) exit, so a clean self-skip/no-work exit left `loom:building` **orphaned** (hand-recovered in the dogfood). The `Exited` branch now restores the label for a daemon-owned sweep that exits **without a checkpoint and without a recorded PR** — automatic recovery, no manual `restore_label_to_ready`.

### #3824 — permissions
Daemon-spawned children ran without `--dangerously-skip-permissions`; a detached, non-interactive `claude -p` auto-denies tool calls needing approval and stalls Stage -1 / the build. `spawn_child()` now appends `--dangerously-skip-permissions` to the claude argv (**after** `--model`/`--effort`/`--depends-on`, preserving their positional contract), mirroring the unattended GH Actions cron pattern. `spawn-claude.sh` stays a generic pass-through and is unchanged.

## Tests (loom-daemon)
- `dispatch_appends_dangerously_skip_permissions` — argv contains the flag, appended after the prompt.
- `dispatch_exports_claim_ownership_marker` — `LOOM_SWEEP_CLAIM_OWNED=<issue>` exported to the child.
- `reap_restores_label_for_orphaned_clean_exit_without_pr` — fake `gh` recorder + `skip_label_flip=false`; asserts the reaper restores `loom:building → loom:issue` for an orphaned clean exit without a PR.

Existing `--model`/`--effort`/`--depends-on` argv tests and the sweep.md doc-lint tests remain green.

## Verification
- `cargo test` — **426 lib + all integration tests pass**.
- `cargo clippy -- -D warnings` — fails only on the **pre-existing** `forge_parser.rs:91` uninlined-format-args error on `origin/main` (left untouched per instructions); the edits introduce no new clippy warnings.

Closes #3823
Closes #3824
Part of #3809 Phase 0.5 (daemon dispatch→build contract)


rjwalters:
## ✅ Approved — daemon dispatch→build contract fix

Reviewed rigorously against the three fixes on the `SweepRegistry` child-spawn path. All highest-risk logic (self-skip recognition and orphaned-claim recovery gate) is correct.

### (a) CORE — self-skip fix works end-to-end, foreign claims still skipped ✅
`spawn_child()` exports `LOOM_SWEEP_CLAIM_OWNED=issue.to_string()` — the **exact** dispatched issue number. The `sweep.md` pre-flight prose is precise: it exempts an existing `loom:building` **only when the marker is SET and EQUALS the issue N being pre-flighted**. When the marker is unset (operator-run `/loom:sweep`, GH Actions cron) **or names a different issue**, the skip rule applies unmodified — a genuine foreign `loom:building` claim from another worker is **still skipped**. The fix does NOT reintroduce double-builds.

### (b) Dedup preserved ✅
Concurrent same-issue dispatch is deduped upstream of the marker: `acquire_lock` (atomic `mkdir`, step 3) returns a "lock collision" error before the label flip/spawn, and the idempotency-key check (step 1) short-circuits earlier. The marker only affects the child's *self*-recognition, never cross-dispatch dedup.

### (c) #3823b orphaned recovery — gate is correct ✅
Recovery fires **only** in the `Exited` (no-checkpoint) branch, inside the `Some(issue)` arm (Issue kind only — PrSet is untouched), gated on `!skip_label_flip && !produced_pr`. A sweep that reached Builder and opened a PR writes a `builder-done` checkpoint → routes to the `Crashed` branch, never this code. The `pr_number.is_none()` gate is a correct forward-defensive guard (prod currently never sets `pr_number`, so it's a no-op today, but it cannot fight a future in-flight-PR recording). The merge-done path deletes the checkpoint and closes the issue, so the restore then fires on a *closed* issue — harmless per Loom's documented "agents filter by open state" label policy. Does not undo real progress.

### (d) #3824 argv ✅
`--dangerously-skip-permissions` appended unconditionally **after** `--model`/`--effort`/`--depends-on` (all emitted before the env chain), preserving their positional contract — the test asserts the exact form `argv: -p /loom:sweep 4242 --dangerously-skip-permissions`. Scoped to the daemon `spawn_child` path only; `spawn-claude.sh` stays a generic pass-through. Security posture is appropriate: a detached, non-interactive `claude -p` sweep child has no human to answer prompts, mirroring the established GH Actions cron pattern.

### (e) No Phase-0 regression ✅
Purely additive (176 insertions, 0 deletions in `sweep_registry.rs`). Singleton/IPC/cancel/reaper/token-selection paths untouched.

### (f) Tests + CI ✅
- `dispatch_appends_dangerously_skip_permissions`, `dispatch_exports_claim_ownership_marker`, `reap_restores_label_for_orphaned_clean_exit_without_pr` — all pass and genuinely exercise argv/marker/reaper-restore.
- `cargo test` — **426 lib + all integration tests pass** (local + CI green).
- `cargo clippy -- -D warnings` — fails only on the **pre-existing** `forge_parser.rs:91` uninlined-format-args error on `origin/main` (confirmed identical on main, untouched here); no new warnings from the edits.
- `gh pr checks` — all green (Rust Unit Tests, Rust Linting & Build, Installer, Detect Changes). Merge state **CLEAN**.

Closing keywords `Closes #3823` / `Closes #3824` present and correct; `Part of #3809 Phase 0.5` is an intentional epic reference.

LGTM.

Issue #3828, opened 2026-07-23
loom-daemon: pre-existing forge_parser.rs:91 clippy error (uninlined_format_args) blocks 'cargo clippy -D warnings'


rjwalters:
## Context

`cargo clippy -- -D warnings` fails on `loom-daemon/src/forge_parser.rs:91` with `uninlined_format_args`. It's pre-existing on `main` (a newer local clippy toolchain flags it; the CI-pinned toolchain does not), but it means **every daemon PR this session had to explicitly note and work around it** — #3804, #3815, #3816, #3817, #3818, #3819, #3825 all carried "the sole clippy error is the pre-existing forge_parser.rs:91, left untouched." That's a recurring papercut that makes a clean `clippy -D warnings` impossible and forces every builder/judge to reason about whether a clippy failure is "theirs."

## Root cause (confirmed)
`forge_parser.rs:91`:
```rust
let url_pattern = format!(r"https://{}/[^/]+/[^/]+/(issues|pulls?)/(\d+)", escaped_host);
```
`uninlined_format_args` wants the captured identifier inlined.

## Suggested fix
Inline the arg:
```rust
let url_pattern = format!(r"https://{escaped_host}/[^/]+/[^/]+/(issues|pulls?)/(\d+)");
```
Then sweep the file (and the crate) for any other `uninlined_format_args` so `cargo clippy --all-targets -- -D warnings` is clean on the daemon crate. (Note: judges also flagged a pre-existing `type_complexity` on an `ipc.rs` test helper `setup_test_context` — worth cleaning in the same pass if trivial.)

## Suggested acceptance criteria
- [ ] `cargo clippy -- -D warnings` (and ideally `--all-targets`) passes clean on `loom-daemon` with no pre-existing-error carve-outs.
- [ ] No behavior change (pure lint fix).

rjwalters:
## Curator enrichment

Confirmed root cause and verified a complete crate-wide sweep locally (patched, ran clippy, reverted — no working-tree changes left).

### Root cause of the local/CI mismatch
`.cargo/config.toml` sets `rustflags` for `cfg(all())` that enable a much wider lint surface than the actual CI job (`-D clippy::all/correctness/suspicious/complexity`, `-W clippy::pedantic/perf/style`, plus explicit `-W unwrap_used/expect_used/panic/todo/unimplemented`). CI's `backend-lint-and-check` job (`.github/workflows/ci.yml:57`) runs a narrower, explicit allow-list:
```
[...]

Issue #3827, opened 2026-07-23
loom-daemon: orphaned-claim recovery (#3823b) doesn't cover the cancel path — cancelled sweeps leave loom:building


rjwalters:
**Part of #3809** (autonomous daemon mode). Follow-up to #3823/#3825.

## Context — observed live (2026-07-23)

#3825 (fixing #3823b) added orphaned-claim recovery: `reap_once()` restores `loom:building → loom:issue` for a daemon-owned sweep that exits cleanly with no PR. But that recovery lives only in the **`reap_once` Exited branch**. When a sweep is **cancelled** (`CancelSweep`), it goes through the separate `cancel_sweep_nonblocking` → `finish_cancel` path, which transitions the sweep to terminal **without** calling `restore_label_to_ready`. So a cancelled daemon sweep leaves `loom:building` orphaned.

Verified: cancelled two hung sweeps (#3780, #3785) via `CancelSweep` — the ack came back, the children died, but both issues stayed `loom:building` and had to be hand-recovered.

## Root cause (confirmed in source)
- `restore_label_to_ready()` exists at `loom-daemon/src/sweep_registry.rs:675`.
- It is called only from `reap_once()` (~lines 1284, 1311) on the crashed/orphaned-exit branches.
- `finish_cancel()` (~line 1148) — the cancel terminal path — never calls it.

## Suggested acceptance criteria
- [ ] Cancelling a daemon-owned sweep (Issue kind) that has not produced a PR restores `loom:building → loom:issue` (recoverable), same as the clean-exit path.
- [ ] A cancelled sweep that DID open a PR does NOT get its label reset (mirror the `pr_number.is_none()` gate from #3823b — don't undo real progress).
- [ ] Test: dispatch, cancel before any PR, assert the label is restored to `loom:issue`.

## Related
#3823 / #3825 (the clean-exit recovery this extends), #3800/#3807 (the cancel path).

rjwalters:
## Curator enrichment

Verified the root cause against current `main` (`loom-daemon/src/sweep_registry.rs`). Confirmed accurate:

- `restore_label_to_ready()` — `sweep_registry.rs:675` — best-effort `gh issue edit <n> --remove-label loom:building --add-label loom:issue`.
- `reap_once()`'s orphaned-claim recovery (added by #3823b/#3825) is at `sweep_registry.rs:1301-1324`. The gate is:
  ```rust
[...]

PR #3830, opened 2026-07-23
fix(sweep): daemon-owned child short-circuits Stage -1 to subagent (#3829)

### LINKED ISSUE #3829: loom-daemon: dispatched headless /loom:sweep children non-deterministically hang (idle-blocked, no build progress)
**Part of #3809** (autonomous daemon mode) — **the critical remaining blocker**. Even with the self-skip (#3823) and permissions (#3824/#3825) fixes landed, daemon-dispatched sweeps only *sometimes* build; often they hang.

## Symptom (observed live 2026-07-23, repeatedly)

Dispatched 3 disjoint issues (#3778/#3780/#3785) to the daemon: **only 1 (#3778) built; 2 hung idle** for 9+ min with no worktree, no PR, ~0% CPU — then a focused re-probe of #3780 hung again. Non-deterministic (the hung set even shared an OAuth account with the building one, so it's not a per-account rate limit).

## Root cause data (captured from a live hung child)

`claude -p "/loom:sweep 3780" --dangerously-skip-permissions` (the daemon-spawned child), hung at ~0% CPU:

- **Idle-blocked, not looping.** `sample` shows the main thread and all Bun worker threads parked in `__ulock_wait2` (futex/lock wait) — the process is *waiting*, computing nothing.
- **Reached the API, then stalled.** 8 ESTABLISHED HTTPS connections to the Anthropic API host — so it connected and issued requests, then stopped making progress (waiting on a response / stuck agentic loop, not a network failure).
- **Spawned its MCP server.** A child `node mcp-loom/dist/index.js` (now carrying the sweep tools from #3803) is present.
- **Never reached building.** No `feature/issue-N` worktree — it stalled in early lifecycle (Stage -1 / curation), before the Builder phase.

So: a **headless `claude -p /loom:sweep N` session stalls in its agentic loop / on a tool response** — idle-waiting with open API connections — rather than progressing to build.

### Confirmed: the child has no signal that it's daemon-owned in Stage -1

`loom-daemon/src/sweep_registry.rs::dispatch()` (line ~471) flips `loom:issue → loom:building` on the forge **before** spawning the child, then `spawn_child()` (line ~891) exports `LOOM_SWEEP_CLAIM_OWNED=<issue>` into the child's environment. That marker is already consumed **once**, downstream, in `defaults/.claude/commands/loom/sweep.md` "1. Per-issue pre-flight" (~line 1224) — the daemon self-claim exception from #3823 — so the child doesn't skip its OWN `loom:building` claim during pre-flight.

But `LOOM_SWEEP_CLAIM_OWNED` is **never consulted in Stage -1 ("Backend detection", `sweep.md` ~line 633)**. The `DECIDE` tree there is:

```text
PROBE_MODE:
  If --prs flag present OR any PR-side NL trigger detected → Mode C (subagent always)

PROBE_DAEMON:
  Ping ~/.loom/loom-daemon.sock with 500ms timeout. Pong → reachable.

PROBE_POOL:
  Count *.token files in .loom/tokens/ OR ACCOUNT_KEY_* lines in .env. Pool exists if count >= 2.

DECIDE:
  if Mode C: use_subagent()
  elif --no-daemon: use_subagent()
  elif PROBE_DAEMON AND PROBE_POOL: use_daemon()
  else: use_subagent()
```

A **daemon-dispatched child** — which is, by construction, running in exactly the environment that makes `PROBE_DAEMON AND PROBE_POOL` true (a live daemon + a multi-account pool, since that's *why* it was dispatched there) — re-runs this exact probe on every invocation and lands on `DECIDE = use_daemon`. It then either:
1. Issues a needless `mcp__loom__list_sweeps` round-trip back into the very daemon that just spawned it (child's `claude` → child's own `mcp-loom/dist/index.js` Node process → Unix socket → daemon), immediately before doing any real work, or
2. Worse, proceeds to the daemon-dispatch branch and calls `mcp__loom__dispatch_sweep(kind={"Issue": N})` **on its own issue number** — a circular re-dispatch of work it is already doing.

This exactly matches the "Spawned its MCP server" + "unix socket fd" + "stalled in early lifecycle (Stage -1)" evidence from the live capture: the hung child had gotten as far as spinning up its MCP server and opening the socket, then never made building progress.

**Note on (2):** `SweepRegistry::acquire_lock()` (`sweep_registry.rs` ~line 569) is a non-blocking POSIX-atomic `mkdir` — a self-redispatch attempt for the same issue number fails **fast** with `"lock collision: issue #N is already claimed"` rather than deadlocking on the lock itself. So the *lock* isn't the literal hang mechanism. The likely hang is in the **MCP round-trip itself**: Stage -1's "500ms timeout" on `PROBE_DAEMON` is **prose directed at the LLM**, not a mechanically-enforced transport timeout — in headless `-p` mode there is no operator to notice a stuck tool call and interrupt it, so if the `mcp__loom__list_sweeps` (or a subsequent `dispatch_sweep`) round-trip stalls for any reason (Node cold-start contention, a slow/queued response from a daemon that is mid-dispatch of sibling candidates in the same wave, or a genuine bug in the IPC handling under concurrent load from 3 near-simultaneous dispatches), the whole headless session just sits in `__ulock_wait2` waiting on that tool result forever, with no retry/backoff and no watchdog. This is consistent with, but not fully proven by, the 8 idle ESTABLISHED HTTPS connections (SDK connection pools commonly keep several idle keep-alives open even after a single logical exchange, so that count alone isn't proof of many agentic turns having run).

### Leading fix (hypothesis #2, the most actionable one)

**A daemon-dispatched child should short-circuit Stage -1 straight to the subagent path**, exactly like `--no-daemon`, recognizing its own `LOOM_SWEEP_CLAIM_OWNED` marker — it is already daemon-spawned, so re-probing (and especially re-dispatching) back to the daemon is circular by construction and removes the entire class of hang described above.

This is a **`sweep.md`-only prose change** (no Rust changes required — confirmed against `spawn_child()`, no new env var needed, `LOOM_SWEEP_CLAIM_OWNED` already exists and is exported by every daemon-dispatched child). Concretely, add a new precedence rule to the `DECIDE` tree in `defaults/.claude/commands/loom/sweep.md` "Stage -1: Backend detection", ordered **before** the `PROBE_DAEMON`/`PROBE_POOL` probes (mirroring the existing `--no-daemon` short-circuit at line ~663):

```text
DECIDE:
  if Mode C: use_subagent()
  elif --no-daemon: use_subagent()
  elif LOOM_SWEEP_CLAIM_OWNED is set: use_subagent()   # NEW — daemon-owned child, skip re-probe entirely
  elif PROBE_DAEMON AND PROBE_POOL: use_daemon()
  else: use_subagent()
```

Update the surrounding prose (the numbered precedence list ~line 660-665, the "PROBE_DAEMON" section ~line 673, and the smoke-test table ~line 807) to document the new rule, and add a 4th smoke test alongside the existing three (`--no-daemon`, daemon-on, daemon-off), analogous to "Mode C (PR-set)": **daemon-owned child** — precondition `LOOM_SWEEP_CLAIM_OWNED=<N>` is set → `DECIDE = use_subagent` regardless of daemon/pool state, and Stage -1 does not even attempt `PROBE_DAEMON`.

This is symmetric with the existing #3823 pre-flight consumption of the same marker one stage later — Stage -1 and "1. Per-issue pre-flight" both need to recognize `LOOM_SWEEP_CLAIM_OWNED`, and only the pre-flight check currently does.

### File overlap check (for parallel dispatch alongside #3827/#3828)

- **#3827** (cancel-path label recovery) touches `loom-daemon/src/sweep_registry.rs` — specifically `finish_cancel()` (~line 1148), a different function than anything this fix touches. This issue's fix is **`sweep.md`-only**, so there is **no file overlap** with #3827's Rust change, even though both nominally reference `sweep_registry.rs` in their investigation. Safe to dispatch in the same wave.
- **#3828** (clippy lint) touches `loom-daemon/src/forge_parser.rs` only — no overlap.

### Hypothesis #1 (headless `-p` + nested Task-subagent orchestration stall) — not ruled out

Harder to confirm/deny from static reading alone. The daemon-dispatched child runs the **full** sweep skill in a single headless `-p` session, which internally dispatches Task subagents (Curator → Builder → Judge → Doctor). A one-shot, non-interactive `claude -p` session driving a long multi-subagent agentic loop *could* stall waiting on a subagent/tool result the headless loop doesn't pump — adjacent to the #3289 nested-dispatch hazard, but here the *daemon child itself* is the nesting root, not a second-level `/shepherd`-style dispatch. If the Stage -1 short-circuit (above) does **not** fully resolve the non-determinism after a multi-issue daemon dispatch retest, this hypothesis should be investigated next — likely via a repeat live capture (`sample <pid>`) on a hang that occurs **after** confirming the child never re-entered Stage -1's daemon path (e.g., grep the child's own sweep log for `mechanism=daemon` — its absence would rule out hypothesis #2 for that specific hang and point at #1).

- [ ] Root cause identified (which of the hypotheses, or another) with evidence — **hypothesis #2 (Stage -1 daemon-probe circularity) confirmed as the concrete gap**: `LOOM_SWEEP_CLAIM_OWNED` exists and is exported to every daemon-dispatched child but is never consulted in Stage -1's `DECIDE` tree, so every daemon child needlessly re-probes (and can circularly re-dispatch) the daemon that spawned it.
- [ ] `defaults/.claude/commands/loom/sweep.md` Stage -1 `DECIDE` tree short-circuits to `use_subagent()` when `LOOM_SWEEP_CLAIM_OWNED` is set, **before** the `PROBE_DAEMON`/`PROBE_POOL` probes run (mirrors the `--no-daemon` precedence). Document with a 4th smoke test.
- [ ] A daemon-dispatched sweep reliably progresses to build → PR (no idle-hang): dispatch 3 disjoint issues, **all 3 build** (not 1 of 3) — the concrete regression-guard replacing the observed 1-of-3 live failure.
- [ ] If a hang is still observed after the Stage -1 fix lands, capture whether the child's sweep log shows `mechanism=daemon` at any point post-fix (it should never again, for a daemon-dispatched child) — this cleanly separates hypothesis #2 (should be fully closed) from hypothesis #1 (would need separate follow-up if hangs persist).
- [ ] A hang is at least *detected* (a heartbeat/watchdog) and the sweep fails-and-recovers (label restored) rather than silently occupying a registry slot forever — complements #3827's cancel-path label recovery for the operator-cancelled case; this AC covers the *undetected* hang case (nobody cancels it) and may be a separate, larger follow-up if the Stage -1 fix alone doesn't close the loop (e.g., pairing with the daemon's existing 30s reaper + a max-runtime threshold).

## Affected Files

- `defaults/.claude/commands/loom/sweep.md` (Stage -1 `DECIDE` tree, ~line 633-880 — prose-only change, verified present on `origin/main`)

## Related

#3809 (epic — Phase A #3810 inherits this), #3823/#3824/#3825 (the prior dispatch-contract fixes), #3827 (cancel-path recovery — same file referenced but disjoint function, no overlap), #3828 (unrelated clippy lint, no overlap), #3289 (nested-dispatch stall — conceptually adjacent to hypothesis #1).

rjwalters:
Closes #3829

Part of #3809 (autonomous daemon mode) — the critical remaining idle-hang blocker. Disjoint from #3827 (`sweep_registry.rs::finish_cancel`) and #3828 (`forge_parser.rs` clippy); this is a `sweep.md`-only prose change with no file overlap.

## Root cause

A daemon-dispatched `claude -p "/loom:sweep N"` child re-runs Stage -1 backend detection on every invocation. Because it runs, **by construction**, in the exact environment that makes `PROBE_DAEMON ∧ PROBE_POOL` true (a live daemon + multi-account pool — that is *why* it was dispatched there), the old `DECIDE` tree always landed it on `use_daemon`. It then issued a **circular** MCP round-trip back into the very daemon that spawned it (`mcp__loom__list_sweeps`, or worse a self-re-dispatch of its own issue number via `mcp__loom__dispatch_sweep`). In headless `-p` mode Stage -1's "500ms timeout" is LLM-directed prose, not a mechanically-enforced transport guard, and there is no operator to interrupt a stuck tool call — so that round-trip could hang the session idle (observed live: only 1 of 3 dispatched issues built; 2 hung idle at ~0% CPU with an MCP server spawned and a Unix socket open, never reaching the Builder phase).

The daemon already exports `LOOM_SWEEP_CLAIM_OWNED=<issue>` into every dispatched child (`SweepRegistry::dispatch` → `spawn_child`, `sweep_registry.rs:891`, from #3825). It was consumed once, one stage later, by the #3823 pre-flight self-claim exception — but **never** in Stage -1.

## Fix (sweep.md-only)

Add a precedence rule to the Stage -1 `DECIDE` tree, ordered **before** the daemon/pool probes, mirroring the existing `--no-daemon` short-circuit:

```text
DECIDE:
  if Mode C: use_subagent()
  elif --no-daemon: use_subagent()
  elif LOOM_SWEEP_CLAIM_OWNED is set: use_subagent()   # NEW (#3829)
  elif PROBE_DAEMON AND PROBE_POOL: use_daemon()
  else: use_subagent()
```

A daemon-dispatched child is already the daemon's work — it must run the full Curator→Builder→Judge→Doctor→Merge lifecycle **in-process**, exactly like `--no-daemon`, never re-probing or re-dispatching to the daemon that spawned it. This removes the entire class of hang. No Rust change is required (the marker already exists and is exported).

## Changes (all in `defaults/.claude/commands/loom/sweep.md`)

- `DECIDE` pseudocode: new `LOOM_SWEEP_CLAIM_OWNED` short-circuit rule (before probes).
- Precedence prose: new numbered item 3 explaining the circular-round-trip rationale; renumbered the daemon/fallthrough items.
- `PROBE_DAEMON` pseudocode: belt-and-suspenders `NO_DAEMON or LOOM_SWEEP_CLAIM_OWNED` guard.
- Subagent-fallthrough reason list + auto-wave-size comment: added the new reason.
- New 4th "daemon-owned child" smoke test (alongside daemon-on / daemon-off / `--no-daemon` / Mode C).
- New "What Stage -1 does NOT do" bullet.
- Feature summary-table row updated to reference #3829.

## Verification

- `DECIDE` logic is coherent: `LOOM_SWEEP_CLAIM_OWNED` set → `use_subagent`, ordered before `PROBE_DAEMON`/`PROBE_POOL`.
- Marker referenced consistently with its export in `sweep_registry.rs::spawn_child` (`.env("LOOM_SWEEP_CLAIM_OWNED", issue.to_string())`, line 891) and its existing #3823 pre-flight consumer.
- Diff is a single file, 33 insertions / 6 deletions, prose only.

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -653,6 +653,7 @@ PROBE_POOL:
 DECIDE:
   if Mode C: use_subagent()
   elif --no-daemon: use_subagent()
+  elif LOOM_SWEEP_CLAIM_OWNED is set: use_subagent()   # daemon-owned child — skip re-probe entirely (#3829)
   elif PROBE_DAEMON AND PROBE_POOL: use_daemon()
   else: use_subagent()
 ```
@@ -661,8 +662,9 @@ The precedence is deliberate:
 
 1. **Mode C → subagent** (always, regardless of daemon/pool state). The daemon's dispatch surface is **issue-keyed only** in v0.10.0 (`mcp__loom__dispatch_sweep --kind '{"Issue":N}'`); PR-set dispatch is an explicit non-goal of the parent epic and is not on the v0.10.0 roadmap. PR-set sweeps therefore route to the existing in-process subagent path, which already supports Mode C end-to-end.
 2. **`--no-daemon` → subagent** (operator opt-out, after Mode C but before any probes). When this flag is present, do not even attempt the `PROBE_DAEMON` Ping — saves a 500ms ceiling and produces predictable behaviour for debug/demo/scripted runs.
-3. **`PROBE_DAEMON ∧ PROBE_POOL → daemon`** (the only way to land on the daemon path). **Strict AND**: both probes must succeed. Either missing → fallthrough.
-4. **Else → subagent** (the universal fallthrough, equivalent to v0.9.x behaviour).
+3. **`LOOM_SWEEP_CLAIM_OWNED` set → subagent** (daemon-owned child self-detection, #3829 — after `--no-daemon`, still **before** any probes). This env var is exported **only** into a child that `loom-daemon` itself dispatched (`SweepRegistry::dispatch` → `spawn_child`, `sweep_registry.rs`), carrying the issue number the daemon already claimed on this child's behalf (same marker the "1. Per-issue pre-flight" self-claim exception from #3823 consumes one stage later). A daemon-dispatched child is **by construction** running in the exact environment that makes `PROBE_DAEMON ∧ PROBE_POOL` true — a live daemon plus a multi-account pool, since that is *why* it was dispatched there — so without this rule it would always land on `use_daemon` and issue a **circular** MCP round-trip back into the very daemon that spawned it (`mcp__loom__list_sweeps`, or worse a self-re-dispatch of its own issue number). In headless `claude -p` mode there is no operator to interrupt a stuck tool call and Stage -1's "500ms timeout" is LLM-directed prose, not a mechanically-enforced transport guard, so that round-trip can hang the whole session idle before it ever reaches the Builder phase. The child is already the daemon's work — it must run the lifecycle **itself**, in-process, exactly like `--no-daemon`. This short-circuit removes the entire class of hang. Mirrors `--no-daemon`: do not even attempt the `PROBE_DAEMON` Ping.
+4. **`PROBE_DAEMON ∧ PROBE_POOL → daemon`** (the only way to land on the daemon path). **Strict AND**: both probes must succeed. Either missing → fallthrough.
+5. **Else → subagent** (the universal fallthrough, equivalent to v0.9.x behaviour).
 
 ### The three probes
 
@@ -684,8 +686,10 @@ A successful response (any well-formed `EventStream`/sweep-list payload, includi
 ```text
 PROBE_DAEMON pseudocode (LLM-directed):
 
-  if NO_DAEMON:
+  if NO_DAEMON or LOOM_SWEEP_CLAIM_OWNED is set:
       PROBE_DAEMON = false   # short-circuit; do not even issue the call
+                             # (LOOM_SWEEP_CLAIM_OWNED: daemon-owned child, #3829 —
+                             #  re-probing the spawning daemon is circular)
   else:
       try:
           response = mcp__loom__list_sweeps(timeout_ms=500)
@@ -738,7 +742,7 @@ if [[ "$DECIDE" == use_daemon ]]; then
     # Detached-process path: each sweep is its own OS process with its own
     # rotated token. NOT nested subagents, so #3289 does not apply — scale to 10.
     MECH=daemon;   MECHANISM="daemon detached-process"
-else  # use_subagent (no daemon, single-token pool, --no-daemon, or Mode C)
+else  # use_subagent (no daemon, single-token pool, --no-daemon, daemon-owned child, or Mode C)
     # In-session Task subagents, one level deep. WIDTH is bounded by the harness
     # concurrency cap (min(16, cores-2)), NOT by #3289 (which is a nesting rule,
     # not a width rule). Core-scale the subagent target within [3, 6] via
@@ -800,7 +804,7 @@ The daemon enqueues the sweep, returns a sweep ID, and the skill logs the dispat
 
 ### The subagent fallthrough (when `DECIDE = use_subagent`)
 
-Otherwise — `DECIDE` is `use_subagent` for **any** of the reasons above (Mode C, `--no-daemon`, daemon unreachable, no pool, or any probe error) — **continue to "0. Dry-run gate" below and run the existing Mode A/B/C lifecycle in-process exactly as today**. This is the v0.9.x behaviour, unchanged. The skill prose from "0. Dry-run gate" onward is the canonical subagent path.
+Otherwise — `DECIDE` is `use_subagent` for **any** of the reasons above (Mode C, `--no-daemon`, `LOOM_SWEEP_CLAIM_OWNED` set (daemon-owned child, #3829), daemon unreachable, no pool, or any probe error) — **continue to "0. Dry-run gate" below and run the existing Mode A/B/C lifecycle in-process exactly as today**. This is the v0.9.x behaviour, unchanged. The skill prose from "0. Dry-run gate" onward is the canonical subagent path.
 
 No behaviour change for solo-token operators: their `PROBE_POOL` returns `false`, the `DECIDE` lands on `use_subagent`, and the rest of the skill runs as it always has.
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed this daemon-hang fix (Closes #3829, Part of #3809). This is a single-file prose change to `defaults/.claude/commands/loom/sweep.md` (33 insertions / 6 deletions). Verified the LOGIC end-to-end:

**(a) DECIDE precedence is correct.** The new `elif LOOM_SWEEP_CLAIM_OWNED is set: use_subagent()` is placed **before** `elif PROBE_DAEMON AND PROBE_POOL: use_daemon()`, so a daemon-dispatched child short-circuits to the in-process subagent path and never re-probes/re-dispatches back into its spawning daemon — this removes the circular `mcp__loom__list_sweeps` / self-re-dispatch round-trip that hung headless `-p` sessions.

**(b) Ordering vs Mode-C and `--no-daemon` is consistent.** The new rule sits third — after `Mode C` and `--no-daemon`, before the probes — mirroring the existing `--no-daemon` short-circuit. It does not perturb the two earlier short-circuits (both still fire first). Precedence prose renumbered correctly (items 3→5).

**(c) Normal operator sweeps are UNAFFECTED.** The `elif` only fires when the marker is **set**. For a regular operator `/loom:sweep N` (marker unset), the tree falls through to `elif PROBE_DAEMON AND PROBE_POOL: use_daemon()` exactly as before — daemon dispatch is preserved. The fix narrows behavior only for daemon-owned children.

**(d) Marker name matches its export.** `LOOM_SWEEP_CLAIM_OWNED` matches `sweep_registry.rs:891` — `.env("LOOM_SWEEP_CLAIM_OWNED", issue.to_string())` (from #3825), and the existing #3823 pre-flight self-claim consumer one stage later.

**(e) Belt-and-suspenders guard + smoke test are coherent.** The `PROBE_DAEMON` pseudocode gains `if NO_DAEMON or LOOM_SWEEP_CLAIM_OWNED is set: PROBE_DAEMON = false`, so even independent of the DECIDE tree the probe call is never issued. The new 4th smoke test (daemon-owned child) accurately asserts: PROBE_DAEMON skipped → DECIDE = use_subagent → full in-process lifecycle → no circular re-dispatch.

**(f) Rationale prose is accurate.** A daemon-dispatched child is by construction the daemon's own work running in an environment where PROBE_DAEMON ∧ PROBE_POOL is true; re-dispatching there is circular. The '500ms timeout' is correctly characterized as LLM-directed prose, not a mechanical transport guard.

**Key correctness confirmations:**
- daemon-owned child (marker set) → **use_subagent** (in-process lifecycle) ✓
- normal operator sweep (marker unset) → **daemon-dispatch unaffected** ✓

CI: Installer Integration Tests pass, Detect Changes pass; Rust jobs correctly skip (doc-only change). mergeStateStatus CLEAN.

Correct, minimal, well-documented. No Rust change needed since the marker already exists and is exported. LGTM.

Issue #3704, opened 2026-07-22
feat(tokens): retire ~/.loom/accounts.env as the default account source (claude-monitor-first)


rjwalters:
## Original Issue

Follow-up to #3698 (which made account resolution **claude-monitor > repo \`.env\` > \`~/.loom/accounts.env\`**, keeping the home file as a demoted-but-still-read fallback).

**Goal:** complete the source migration — drop \`~/.loom/accounts.env\` from the *default* resolution so it is no longer auto-read. New default order: **claude-monitor (\`~/.claude-monitor/accounts.env\`) → repo \`.env\` (round-robin fallback)**.

**Mandatory constraints (additive-safety, mirrors #3695/#3698):**
- \`LOOM_ACCOUNTS_ENV\` explicit override MUST still be honored — an operator can still point it at any file, including \`~/.loom/accounts.env\`. This retires the *default location*, not the *capability*.
- Backward-compatible: a repo relying solely on repo \`.env\` (no claude-monitor) behaves exactly as today.
- Soft dependency preserved: absent \`~/.claude-monitor/*\`, resolution falls back to repo \`.env\`.

**Tests:** assert \`~/.loom/accounts.env\` is NOT auto-read when neither claude-monitor nor \`LOOM_ACCOUNTS_ENV\` is set; assert \`LOOM_ACCOUNTS_ENV\` still resolves it explicitly; backward-compat regression for the repo-\`.env\`-only path.

Affected: \`loom-tools/src/loom_tools/tokens/bootstrap.py\` (source resolution), tests under \`loom-tools/tests/tokens/\`.

---

### Problem Statement

After #3695 (home-dir master) and #3698 (three-source merge), `bootstrap_tokens()` resolves accounts from three sources in precedence order **claude-monitor (`~/.claude-monitor/accounts.env`) > repo (`.loom/accounts.env` or legacy `.env`) > home (`~/.loom/accounts.env`)**. #3698 demoted the `~/.loom/accounts.env` home master from the *top* default to a "still-read fallback" tier, but it is still **auto-read by default** whenever it exists on disk and `LOOM_ACCOUNTS_ENV` is unset.

[...]

PR #3846, opened 2026-07-23
fix(sweep): Stage -1 pool probe counts merged account sources, not just legacy .env

### LINKED ISSUE #3789: /loom:sweep Stage -1 pool probe only greps legacy .env — misses claude-monitor and .loom/accounts.env account sources
## Problem

`/loom:sweep` Stage -1's token-pool probe only checks the **legacy** account source, so operators using the post-#3695/#3704 claude-monitor-first layout silently lose daemon dispatch.

`sweep.md:598`: "Count `*.token` files in `.loom/tokens/` OR `ACCOUNT_KEY_*` lines in `.env`", implemented at `sweep.md:657` as:

```bash
grep -c '^ACCOUNT_KEY_' .env
```

Per CLAUDE.md ("Account sources: claude-monitor-first + per-repo"), accounts now live primarily in `~/.claude-monitor/accounts.env` (or `$LOOM_CLAUDE_MONITOR_DIR`) and `<repo>/.loom/accounts.env`; `<repo>/.env` is the legacy fallback. An operator with a claude-monitor-only config who hasn't yet run `loom-tokens bootstrap` gets `PROBE_POOL=false`, Stage -1's strict AND fails, and every sweep falls through to in-process subagent dispatch — losing multi-account parallelism with no signal as to why.

## Suggested fix (original filing)

Make the "configured pool" branch check the merged source set. Simplest options:

- Shell out to `loom-tokens bootstrap --dry-run` and parse the effective merged account count (it already prints the merged set + provenance), or
- Extend the grep to all sources: `.loom/accounts.env`, `${LOOM_CLAUDE_MONITOR_DIR:-~/.claude-monitor}/accounts.env`, and (only when `LOOM_ACCOUNTS_ENV` is set) that path — in addition to legacy `.env`.

Either way, when the probe finds ≥2 accounts declared but `.loom/tokens/` empty, the Stage -1 log line should say "run `loom-tokens bootstrap`" rather than silently reporting no pool.

---
Part of the 2026-07-22 prompt-system audit.

---

### Root cause / technical context

Stage -1 lives entirely in `defaults/.claude/commands/loom/sweep.md` (an LLM-directed prose spec, not a real script — `.claude/commands/loom/sweep.md` is a symlink to this file, so it is the single source of truth to edit). The "PROBE_POOL" contract has two places that reference `.env` and need updating together:

- **Decision-tree summary** (~line 650-651):
  ```
  PROBE_POOL:
    Count *.token files in .loom/tokens/ OR ACCOUNT_KEY_* lines in .env. Pool exists if count >= 2.
  ```
- **Detailed "#### PROBE_POOL" section** (~lines 703-724), including the literal bash snippet (~lines 712-719):
  ```bash
  TOKEN_FILE_COUNT=$(ls .loom/tokens/*.token 2>/dev/null | wc -l | tr -d ' ')
  ENV_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || echo 0)
  if (( TOKEN_FILE_COUNT >= 2 )) || (( ENV_KEY_COUNT >= 2 )); then
    PROBE_POOL=true
  else
    PROBE_POOL=false
  fi
  ```
- **Smoke-test precondition comments** (~lines 819-820 and ~837-838) also say "At least 2 accounts in `.env` / `.loom/tokens/`" and "`.env` has < 2 `ACCOUNT_KEY_*` lines" — these should stay accurate once the probe covers more sources (they don't need to enumerate every source, just not imply `.env` is the only configured-pool signal).

`ACCOUNT_KEY_*` line format is **identical** across all three account sources (confirmed in `loom-tools/src/loom_tools/tokens/bootstrap.py`): claude-monitor's `accounts.env` writes the same `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` numbered-triple format the repo-local/legacy files use (token filename is auto-derived there, but the key line shape is unchanged). So the same `grep -c '^ACCOUNT_KEY_'` pattern works unmodified against each candidate file — no new parsing logic is needed, just more files to check.

### Is there a shared helper both prose and script could use?

**No.** Unlike "Resolve auto wave size" (which sources `defaults/scripts/lib/disk-headroom.sh`, a real, unit-tested bash helper), Stage -1's probes are inline prose/bash with no sourced helper file. The authoritative multi-source merge-by-email logic lives in Python (`loom_tools.tokens.bootstrap.merge_accounts` + `monitor.claude_monitor_dir()`), not in bash.

### Implementation guidance — recommended approach

**Prefer extending the inline bash probe (the issue's second suggested option), not shelling out to `loom-tokens bootstrap --dry-run` (the first option).** Reasons:

1. **Consistency with Stage -1's existing design intent.** The section explicitly documents PROBE_DAEMON and PROBE_POOL as "small, side-effect-free probes" with no external process dependency (PROBE_DAEMON uses the MCP `list_sweeps` call, not a CLI shell-out). Introducing a Python subprocess dependency (`loom-tokens` must be on `PATH`, `loom-tools` must be installed) for a probe that today needs zero Python breaks that invariant and adds a new hard failure mode (a plain sweep-with-`.env`-only-operator now depends on the Python package being importable).
2. **No precision loss that matters for a boolean gate.** PROBE_POOL only needs "is there a configured multi-account pool, roughly `>= 2` accounts" — not an exact deduplicated-by-email count. A cheap OR-of-greps across the known source files is accurate enough for this yes/no gate; the real bootstrap merge (with its precedence/override rules) still runs later when `loom-tokens bootstrap` actually materializes the pool.
3. **No shared helper currently unifies both** — building one (a `defaults/scripts/lib/account-pool-probe.sh` mirroring `disk-headroom.sh`'s pattern) is a reasonable stretch goal but out of scope for this fix; a same-shape inline bash OR is sufficient and matches the existing style of the PROBE_POOL block.

**Concrete change**: extend the bash snippet to sum `ACCOUNT_KEY_*` lines across all resolvable sources (still cheap, local, side-effect-free — mirrors bootstrap.py's precedence order but does **not** need to dedupe by email for a boolean threshold check):

```bash
# Legacy fallback / repo-local (mirrors bootstrap.py: .loom/accounts.env if present, else .env)
if [[ -f .loom/accounts.env ]]; then
  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .loom/accounts.env 2>/dev/null || echo 0)
else
  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || echo 0)
fi

# claude-monitor master (primary source per CLAUDE.md; LOOM_CLAUDE_MONITOR_DIR override)
MONITOR_DIR="${LOOM_CLAUDE_MONITOR_DIR:-$HOME/.claude-monitor}"
MONITOR_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$MONITOR_DIR/accounts.env" 2>/dev/null || echo 0)

# Opt-in home master — only consulted when LOOM_ACCOUNTS_ENV is set (per #3704)
HOME_KEY_COUNT=0
if [[ -n "${LOOM_ACCOUNTS_ENV:-}" ]]; then
  HOME_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$LOOM_ACCOUNTS_ENV" 2>/dev/null || echo 0)
fi

ENV_KEY_COUNT=$(( REPO_KEY_COUNT + MONITOR_KEY_COUNT + HOME_KEY_COUNT ))
TOKEN_FILE_COUNT=$(ls .loom/tokens/*.token 2>/dev/null | wc -l | tr -d ' ')
if (( TOKEN_FILE_COUNT >= 2 )) || (( ENV_KEY_COUNT >= 2 )); then
  PROBE_POOL=true
else
  PROBE_POOL=false
fi
```

Update the decision-tree summary line to say "or the merged claude-monitor / `.loom/accounts.env` / legacy `.env` account sources" instead of just "`.env`" (mirror the "Account sources: claude-monitor-first + per-repo" section of `CLAUDE.md` for terminology).

**Operator-facing log line**: when the probe finds `ENV_KEY_COUNT >= 2` but `TOKEN_FILE_COUNT < 2` (accounts configured, not yet bootstrapped), Stage -1 should log/print a hint: `"Configured account pool detected but not bootstrapped — run 'loom-tokens bootstrap' to materialize .loom/tokens/."` This satisfies the original filing's ask for a discoverable signal instead of a silent fallthrough.

### Out of scope / do NOT change

- Do **not** touch `loom-tools/src/loom_tools/tokens/bootstrap.py` or `merge_accounts()` — that Python merge logic is correct and unaffected; this issue is only about the Stage -1 **prose/bash probe** in `sweep.md`.
- Do **not** attempt exact per-email dedup in the bash probe — summing raw `ACCOUNT_KEY_*` line counts across sources is an accepted approximation for a boolean `>= 2` gate (a single account declared in both claude-monitor and repo-local would double-count, which only matters at the `== 1` vs `== 2` boundary and is a false-positive-toward-daemon-use, not a correctness regression — worst case is one extra probe attempt that still requires `PROBE_DAEMON` to also be true).
- Do **not** change `PROBE_DAEMON`, the `DECIDE` precedence tree, or any other Stage -1 rule (Mode C short-circuit, `--no-daemon`, `LOOM_SWEEP_CLAIM_OWNED`) — this is a scoped fix to the pool-detection source list only.
- Do **not** create a new sourced bash helper file (e.g. `account-pool-probe.sh`) unless it is trivial to add — the inline extension above is sufficient; if a Builder judges a helper is cleaner, keep it small and unit-test it the way `disk-headroom.sh` is tested, but this is optional polish, not a requirement.
- Do **not** modify `loom-daemon/tests/sweep_md_stage_minus_one_doc_lint.rs` — its assertions (`PROBE_POOL` present, `single-token`/`solo-token`/`< 2 ACCOUNT_KEY_` present) are format-flexible and should still pass unmodified after this change; run it to confirm rather than editing it.

## Affected Files

- `defaults/.claude/commands/loom/sweep.md` — Stage -1 "PROBE_POOL" decision-tree summary (~line 650-651), detailed section + bash snippet (~lines 703-724), and the two smoke-test precondition comments (~lines 819-820, 837-838) referencing `.env`. (`.claude/commands/loom/sweep.md` is a symlink to this file — editing the symlink target is sufficient, do not edit through the symlink path.)

## Test Plan

- [ ] Manual verification: with only `~/.claude-monitor/accounts.env` (or a test dir pointed at by `LOOM_CLAUDE_MONITOR_DIR`) populated with ≥2 `ACCOUNT_KEY_*` lines and no `.env`/`.loom/tokens/`, manually trace the updated PROBE_POOL bash snippet and confirm it resolves `PROBE_POOL=true`.
- [ ] Manual verification: with only legacy `.env` populated (today's only-supported case), confirm the snippet still resolves identically to the pre-fix behavior (no regression for existing solo-`.env` and multi-`.env` operators).
- [ ] Automated tests: run `cargo test -p loom-daemon sweep_md_stage_minus_one_doc_lint` (or the equivalent `cargo test --workspace sweep_md` filter) and confirm the existing doc-lint assertions in `loom-daemon/tests/sweep_md_stage_minus_one_doc_lint.rs` still pass unmodified against the edited `sweep.md`.
- [ ] Edge cases: confirm the opt-in `LOOM_ACCOUNTS_ENV` home-master path is only consulted when the env var is set (mirrors bootstrap.py's #3704 opt-in-only behavior) — an unset `LOOM_ACCOUNTS_ENV` must not attempt to read `~/.loom/accounts.env` by default.
- [ ] Edge cases: verify the "configured but not bootstrapped" hint only fires when `ENV_KEY_COUNT >= 2 AND TOKEN_FILE_COUNT < 2` (not on every subagent fallthrough).

rjwalters:
Closes #3789

## Problem

`/loom:sweep` Stage -1's token-pool probe (`PROBE_POOL`) only counted `ACCOUNT_KEY_*` lines in the **legacy** `.env`. Operators on the post-#3695/#3704 claude-monitor-first account layout (`~/.claude-monitor/accounts.env`, `.loom/accounts.env`) who hadn't yet run `loom-tokens bootstrap` got `PROBE_POOL=false`, the Stage -1 strict-AND failed, and every sweep silently fell through to in-process subagent dispatch — losing multi-account daemon parallelism with no signal as to why.

## Change

All edits are confined to the single tracked source `defaults/.claude/commands/loom/sweep.md` (`.claude/commands/loom/sweep.md` is a gitignored installed copy).

- **PROBE_POOL bash probe** now sums `ACCOUNT_KEY_*` lines across all resolvable sources, mirroring `bootstrap.py`'s precedence (no dedup-by-email — a raw sum is an accepted approximation for a boolean `>= 2` gate):
  - Repo-local: `.loom/accounts.env` if present, else legacy `.env`
  - claude-monitor master: `${LOOM_CLAUDE_MONITOR_DIR:-$HOME/.claude-monitor}/accounts.env`
  - Opt-in home master: `$LOOM_ACCOUNTS_ENV` **only** when set and non-empty (per #3704)
- **Discoverable hint**: when accounts are configured (`ENV_KEY_COUNT >= 2`) but not yet bootstrapped (`TOKEN_FILE_COUNT < 2`), the probe prints a stderr line telling the operator to run `loom-tokens bootstrap` — instead of a silent fallthrough. It fires only in that state, not on every subagent fallthrough.
- Updated the decision-tree summary, the detailed `#### PROBE_POOL` section, and the two smoke-test precondition comments to name the merged sources.

## Scope boundaries respected

- No change to `bootstrap.py` / `merge_accounts()`.
- No change to `PROBE_DAEMON`, `DECIDE`, Mode C short-circuit, `--no-daemon`, or `LOOM_SWEEP_CLAIM_OWNED`.
- No change to `loom-daemon/tests/sweep_md_stage_minus_one_doc_lint.rs`.

## Test results

- `cargo test --test sweep_md_stage_minus_one_doc_lint` — **10 passed, 0 failed** (unmodified assertions still hold).
- Manual trace of the updated snippet across five scenarios (all correct):
  | Scenario | Result |
  |----------|--------|
  | claude-monitor only, 2 accounts, no `.env`/tokens | `PROBE_POOL=true` (bug fixed) |
  | legacy `.env` only, 2 accounts | `PROBE_POOL=true` (no regression) |
  | legacy `.env` only, 1 account | `PROBE_POOL=false` (single-token, not a pool) |
  | `LOOM_ACCOUNTS_ENV` unset | home master **not** read (opt-in honored) |
  | `LOOM_ACCOUNTS_ENV` set | home master counted |

Snippet is bash-3.2 portable (`[[ -f ]]`, `${VAR:-default}`, `$(( ))`, no `mapfile`).

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -714,20 +714,23 @@ Both checks are cheap, local, and side-effect-free. The configured-pool count mi
 TOKEN_FILE_COUNT=$(ls .loom/tokens/*.token 2>/dev/null | wc -l | tr -d ' ')
 
 # Repo-local (mirrors bootstrap.py: .loom/accounts.env if present, else legacy .env)
+# NOTE: `grep -c` prints `0` AND exits non-zero on an existing-but-empty file, so a
+# `|| echo 0` fallback would emit a two-line "0\n0" and abort the arithmetic below under
+# bash 3.2. Use `|| true` + `${var:-0}` so an existing-empty source yields exactly `0`.
 if [[ -f .loom/accounts.env ]]; then
-  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .loom/accounts.env 2>/dev/null || echo 0)
+  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .loom/accounts.env 2>/dev/null || true); REPO_KEY_COUNT=${REPO_KEY_COUNT:-0}
 else
-  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || echo 0)
+  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || true); REPO_KEY_COUNT=${REPO_KEY_COUNT:-0}
 fi
 
 # claude-monitor master (primary source per CLAUDE.md; LOOM_CLAUDE_MONITOR_DIR override)
 MONITOR_DIR="${LOOM_CLAUDE_MONITOR_DIR:-$HOME/.claude-monitor}"
-MONITOR_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$MONITOR_DIR/accounts.env" 2>/dev/null || echo 0)
+MONITOR_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$MONITOR_DIR/accounts.env" 2>/dev/null || true); MONITOR_KEY_COUNT=${MONITOR_KEY_COUNT:-0}
 
 # Opt-in home master — only consulted when LOOM_ACCOUNTS_ENV is set and non-empty (per #3704)
 HOME_KEY_COUNT=0
 if [[ -n "${LOOM_ACCOUNTS_ENV:-}" ]]; then
-  HOME_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$LOOM_ACCOUNTS_ENV" 2>/dev/null || echo 0)
+  HOME_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$LOOM_ACCOUNTS_ENV" 2>/dev/null || true); HOME_KEY_COUNT=${HOME_KEY_COUNT:-0}
 fi
 
 ENV_KEY_COUNT=$(( REPO_KEY_COUNT + MONITOR_KEY_COUNT + HOME_KEY_COUNT ))
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -648,7 +648,8 @@ PROBE_DAEMON:
   Ping ~/.loom/loom-daemon.sock with 500ms timeout. Pong → reachable.
 
 PROBE_POOL:
-  Count *.token files in .loom/tokens/ OR ACCOUNT_KEY_* lines in .env. Pool exists if count >= 2.
+  Count *.token files in .loom/tokens/ OR ACCOUNT_KEY_* lines summed across the merged
+  claude-monitor / .loom/accounts.env / legacy .env account sources. Pool exists if count >= 2.
 
 DECIDE:
   if Mode C: use_subagent()
@@ -704,18 +705,43 @@ The `no_such_tool` case covers older Loom installs without Phase A's MCP additio
 
 A pool exists if **either** of these is true (logical OR, both checked):
 
-1. **Materialized pool**: `.loom/tokens/*.token` contains **two or more** files. The bootstrap step (`loom-tokens bootstrap`) writes one `*.token` file per `ACCOUNT_KEY_*` triple in `.env`; a count `>= 2` means at least two distinct accounts are available for rotation.
-2. **Configured pool**: `.env` at the workspace root contains **two or more** `ACCOUNT_KEY_*` lines. This catches the case where the operator has configured multiple accounts but hasn't yet run `loom-tokens bootstrap` — the daemon's spawn-time selector can still pick a token, and the pool will be materialized on demand.
+1. **Materialized pool**: `.loom/tokens/*.token` contains **two or more** files. The bootstrap step (`loom-tokens bootstrap`) writes one `*.token` file per `ACCOUNT_KEY_*` triple in the merged account set; a count `>= 2` means at least two distinct accounts are available for rotation.
+2. **Configured pool**: **two or more** `ACCOUNT_KEY_*` lines are declared across the **merged account sources** — the claude-monitor master (`${LOOM_CLAUDE_MONITOR_DIR:-$HOME/.claude-monitor}/accounts.env`), the repo-local file (`.loom/accounts.env`, falling back to the legacy `.env`), and — **only when `LOOM_ACCOUNTS_ENV` is set** — the opt-in home master at that path. This catches the case where the operator has configured multiple accounts (in the post-#3695/#3704 claude-monitor-first layout, not just the legacy `.env`) but hasn't yet run `loom-tokens bootstrap` — the daemon's spawn-time selector can still pick a token, and the pool will be materialized on demand.
 
-Both checks are cheap, local, and side-effect-free:
+Both checks are cheap, local, and side-effect-free. The configured-pool count mirrors `bootstrap.py`'s source precedence but does **not** dedupe by email — a raw sum of `ACCOUNT_KEY_*` lines is an accepted approximation for this boolean `>= 2` gate (worst case a single account declared in two sources double-counts at the `== 1` vs `== 2` boundary, a false-positive toward daemon use that still requires `PROBE_DAEMON` to also be true):
 
 ```bash
 TOKEN_FILE_COUNT=$(ls .loom/tokens/*.token 2>/dev/null | wc -l | tr -d ' ')
-ENV_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || echo 0)
+
+# Repo-local (mirrors bootstrap.py: .loom/accounts.env if present, else legacy .env)
+if [[ -f .loom/accounts.env ]]; then
+  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .loom/accounts.env 2>/dev/null || echo 0)
+else
+  REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .env 2>/dev/null || echo 0)
+fi
+
+# claude-monitor master (primary source per CLAUDE.md; LOOM_CLAUDE_MONITOR_DIR override)
+MONITOR_DIR="${LOOM_CLAUDE_MONITOR_DIR:-$HOME/.claude-monitor}"
+MONITOR_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$MONITOR_DIR/accounts.env" 2>/dev/null || echo 0)
+
+# Opt-in home master — only consulted when LOOM_ACCOUNTS_ENV is set and non-empty (per #3704)
+HOME_KEY_COUNT=0
+if [[ -n "${LOOM_ACCOUNTS_ENV:-}" ]]; then
+  HOME_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' "$LOOM_ACCOUNTS_ENV" 2>/dev/null || echo 0)
+fi
+
+ENV_KEY_COUNT=$(( REPO_KEY_COUNT + MONITOR_KEY_COUNT + HOME_KEY_COUNT ))
 if (( TOKEN_FILE_COUNT >= 2 )) || (( ENV_KEY_COUNT >= 2 )); then
   PROBE_POOL=true
 else
   PROBE_POOL=false
+fi
+
+# Discoverable signal: accounts configured but not yet bootstrapped. Only fires when
+# the merged sources declare a pool (ENV_KEY_COUNT >= 2) yet .loom/tokens/ has < 2
+# token files — NOT on every subagent fallthrough.
+if (( ENV_KEY_COUNT >= 2 )) && (( TOKEN_FILE_COUNT < 2 )); then
+  echo "Configured account pool detected but not bootstrapped — run 'loom-tokens bootstrap' to materialize .loom/tokens/." >&2
 fi
 ```
 
@@ -817,7 +843,8 @@ These are the AC #3 and AC #4 contracts, written for the operator.
 ```bash
 # Preconditions:
 #   - loom-daemon is running (`pgrep loom-daemon` matches, ~/.loom/loom-daemon.sock exists)
-#   - At least 2 accounts in .env / .loom/tokens/
+#   - At least 2 accounts configured — in .loom/tokens/, or ACCOUNT_KEY_* lines across
+#     the merged claude-monitor / .loom/accounts.env / legacy .env account sources
 
 /loom:sweep 123 456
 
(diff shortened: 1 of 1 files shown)


rjwalters:

Thorough review against issue #3789's acceptance criteria and scope. The intent, scope discipline, and prose are all correct — but there is **one reproducible bug that defeats the PR's own headline fix**, so this needs a revision before merge.

### Blocker: `$(( ... ))` sum hard-errors when any source file exists with zero `ACCOUNT_KEY_*` lines

`grep -c PATTERN file` behaves differently for missing vs. existing-empty files:

- **Missing file** → grep prints nothing, exits `2` → `|| echo 0` fires → captured value is a clean `0`.
- **Existing file, zero matches** → grep prints `0` to stdout **and** exits `1` → `|| echo 0` **also** fires → captured value is the two-line string `"0\n0"`.

The new code then does an **unconditional** arithmetic sum:

```bash
ENV_KEY_COUNT=$(( REPO_KEY_COUNT + MONITOR_KEY_COUNT + HOME_KEY_COUNT ))
```

When any of the three counts is `"0\n0"`, bash 3.2 aborts with `syntax error in expression (error token is "0")`, `ENV_KEY_COUNT` becomes empty, and the pool check collapses to `false`.

The old single-line code masked this because the `"0\n0"` value only ever reached a short-circuiting `(( ENV_KEY_COUNT >= 2 ))` guard (skipped whenever `TOKEN_FILE_COUNT >= 2`). Summing the three sources unconditionally removes that cover.

**Reproduced (bash 3.2.57, the macOS default):** the exact scenario this PR exists to fix —

```
claude-monitor accounts.env = 2 valid ACCOUNT_KEY_ lines
repo .env = "GH_TOKEN=ghp_xxx"   (no ACCOUNT_KEY_ lines — very common)
.loom/tokens/ = empty (un-bootstrapped)

=> /bin/bash: syntax error in expression (error token is "0")
=> ENV_KEY_COUNT=""  =>  PROBE_POOL=false        (expected: true)
```

An operator with a claude-monitor pool and an ordinary `.env` (holding `GH_TOKEN` or any non-account config — which CLAUDE.md itself documents as `.env`'s legacy role) gets `PROBE_POOL=false` and silently falls through to subagent dispatch. That is the identical failure mode #3789 set out to eliminate.

Note: the curated ACs happen to dodge this because they all specify *"and no `.env`"* (missing file → clean `0`). Real repos usually **have** a `.env`, so the manual-trace table in the PR body missed it.

### Suggested fix

Normalize each count so an existing-but-empty source yields exactly `0`. Drop `|| echo 0` (it is the source of the double `0`) and default the missing-file/empty case instead — safe under `set -e` via `|| true`:

```bash
REPO_KEY_COUNT=$(grep -c '^ACCOUNT_KEY_' .loom/accounts.env 2>/dev/null || true); REPO_KEY_COUNT=${REPO_KEY_COUNT:-0}
```

(applied to each of the repo/monitor/home counts). Verified this yields `repo=0 mon=2 sum=2 => PROBE_POOL=true` for the scenario above.

### Everything else checks out (no changes needed)

- Scope respected: no `PROBE_DAEMON` / `DECIDE` / Mode C / `--no-daemon` / `bootstrap.py` changes; test file untouched.
- `sweep_md_stage_minus_one_doc_lint` — **10 passed, 0 failed** against the edited file.
- `LOOM_ACCOUNTS_ENV` opt-in gating (`-n` guard) is correct per #3704; missing-var → home master not read (verified).
- Single-token (`ENV=1`) correctly stays `PROBE_POOL=false`; the "configured but not bootstrapped" hint fires only in the intended `ENV>=2 && TOKEN<2` state (verified).
- `.loom/accounts.env`-else-`.env` precedence and the double-count-across-sources approximation match the curation's accepted semantics; prose is accurate.

Please fix the count normalization and re-request review. Once the arithmetic is robust to existing-empty source files, this is good to go.

rjwalters:
Addressed Judge feedback — the PROBE_POOL count normalization bug is fixed.

**Fix**: dropped `|| echo 0` (the source of the two-line "0\n0" on an existing-but-empty file, which grep prints `0` *and* exits non-zero for) and replaced it with `|| true` + `${var:-0}` defaulting on all three per-source counts (repo / monitor / home). An existing-empty source now yields exactly `0`, so the unconditional `ENV_KEY_COUNT=$(( REPO + MONITOR + HOME ))` no longer aborts under bash 3.2.

**Verified under /bin/bash (bash 3.2.57, macOS default):**
- Judge repro (repo `.env` = `GH_TOKEN=...` only + claude-monitor with 2 accounts, tokens empty) → `repo=0 mon=2 sum=2 PROBE_POOL=true` ✅ (was the silent-fallthrough bug)
- all sources missing → `sum=0 PROBE_POOL=false` ✅
- single token configured → `sum=1 PROBE_POOL=false` ✅
- existing-empty envs + 2 `*.token` files → `PROBE_POOL=true` ✅
- `.loom/accounts.env`=2 (precedence over legacy `.env`) → `PROBE_POOL=true` ✅

**Doc-lint**: `cargo test --test sweep_md_stage_minus_one_doc_lint` → 10 passed, 0 failed.

Re-requesting review.

rjwalters:
## Judge re-review — approved

The blocker from my previous pass is fixed. The Doctor replaced `|| echo 0` with `|| true` + `${var:-0}` defaulting on all three per-source counts (repo / monitor / home), so an existing-but-empty source now yields exactly a single-line `0` instead of the two-line `0\n0` that aborted the unconditional `ENV_KEY_COUNT=$(( REPO + MONITOR + HOME ))` under bash 3.2.

### Verified
- **Diff scope**: the fix commit (`eb7af55b`) touches only `defaults/.claude/commands/loom/sweep.md` (7 insertions, 4 deletions) — no unrelated changes crept in. An explanatory NOTE comment documents the grep-c pitfall.
- **Arithmetic traced + reproduced under real bash 3.2.57 (macOS default):**
  - Original repro (repo `.env` = `GH_TOKEN` only, 0 matches + claude-monitor with 2 accounts, tokens empty) → `repo=0 mon=2 sum=2` → **PROBE_POOL=true** (was the silent-fallthrough bug).
  - Missing-file case (no sources) → `repo=0 mon=0 sum=0` → **PROBE_POOL=false**.
  - Each count captures a clean single-line `0`; `|| true` is `set -e`-safe.
- **Doc-lint**: `cargo test --test sweep_md_stage_minus_one_doc_lint` → **10 passed, 0 failed**.
- **CI**: green (Installer Integration Tests + Detect Changes pass; Rust/MCP jobs correctly path-skipped). Merge state **CLEAN / MERGEABLE**.

LGTM — approved.

PR #3752, opened 2026-07-22
feat(merge-pr): auto-reconcile stacked child PRs on parent merge (#3747 v2 item 1)

rjwalters:
## What landed (item 1 of the v2 epic)

Wires the existing `defaults/scripts/reconcile-stack.sh` into an **automatic, best-effort trigger** inside `defaults/scripts/merge-pr.sh`, fired at the post-merge choke point (alongside `_reset_partial_increment_labels`, **before** branch deletion so the parent branch ref still resolves as reconcile-stack.sh's rebase `<upstream>`).

New `_auto_reconcile_stacked_children` / `_reconcile_one_stacked_child`:

- **GitHub-only**, and only for parent PRs on a `^feature/issue-([0-9]+)$` branch.
- **Live forge discovery**: `gh pr list --base "$PR_BRANCH" --state open` — never the ephemeral `loom-daemon` `SweepRegistry` (terminal entries GC'd ~1h post-transition; registry only exists when the daemon runs, but this can fire from Champion cron or an interactive `/loom:sweep` merge with no daemon).
- **Safe/unsafe split** per child, gated on the child **issue's** `loom:building` label via a fresh uncached `gh api` read (mirrors `_reset_one_partial_issue`'s freshness discipline):
  - **Safe** (issue not `loom:building`): invokes `reconcile-stack.sh <child-pr> <parent-branch>` unmodified — no inline reimplementation of the rebase/force-with-lease/retarget logic.
  - **Unsafe** (issue still `loom:building`): skips the auto-rebase (would race a live Builder's checked-out branch) and posts a `gh pr comment` flagging deferred reconciliation.
- **Best-effort**: a reconciliation failure (rebase conflict, rejected force-with-lease, retarget failure) is logged as a warning and **never changes `merge-pr.sh`'s exit code** — the parent merge already happened.
- **Idempotent** by construction: once a child's base is retargeted away from the parent branch, the `--base` query returns zero rows.

## Explicitly deferred (issue stays OPEN)

Items 2–6 of the v2 epic are **not** in this PR: merge-ordering guard, rebase-on-parent-amend, dependency auto-detection, diamonds/multi-parent, auto-detach. This is a multi-increment epic — this PR uses a non-closing reference so #3747 stays open with the remaining five items.

## Tests + lint

[...]

PR #3848, opened 2026-07-23
fix(scripts): reconcile-stack rebases in the child's worktree + live stale-origin check

### LINKED ISSUE #3776: reconcile-stack.sh fails when the child branch is checked out in a worktree; also false-warns on stale origin ref
## Context

Hit live 2026-07-22 while collapsing a guard-fix PR stack (the manual counterpart of #3747 item-1 auto-reconcile). Running the shipped helper from the main worktree:

```
$ ./.loom/scripts/reconcile-stack.sh 3774 feature/issue-3756
ℹ Step 1/3: rebase --onto main feature/issue-3756 feature/issue-3757
$ git rebase --onto main feature/issue-3756 feature/issue-3757
fatal: 'feature/issue-3757' is already used by worktree at '/…/.loom/worktrees/issue-3757'
ERROR: Rebase failed (likely a conflict)…
```

Two distinct rough edges:

1. **Worktree-checked-out child branch (the blocker).** `reconcile-stack.sh` runs `git rebase --onto <default> <parent-branch> <child-branch>` from wherever it's invoked (typically the main worktree). But a Loom child branch is **always** checked out in its own managed worktree (`.loom/worktrees/issue-<child>`), and git refuses to rebase a branch that's checked out elsewhere. So the helper can essentially never run from the main worktree for a live Loom stack — it only works if the child worktree happens to have been removed. The manual workaround was to `cd` into the child worktree and `git rebase --onto origin/main <parent-tip-SHA>` by hand.

2. **False "origin/<parent> still exists" warning on a stale local ref.** The helper warns `origin/feature/issue-3756 still exists — confirm the parent PR has squash-merged` (reconcile-stack.sh:127). But with delete-branch-on-merge the branch WAS deleted; my local remote-tracking ref was just stale. It needed `git remote prune origin` first. The check should prune/refresh before deciding the parent branch "still exists," or it produces a spurious warning on every post-merge reconcile.

This matters because #3747 item-1 wired this exact helper into `merge-pr.sh`'s auto-reconcile path — the worktree collision would hit there too for any child whose worktree still exists.

- [ ] `reconcile-stack.sh` performs the `git rebase --onto` **inside the child branch's worktree** when one exists (detect via `git worktree list`), or otherwise in a way that doesn't require the branch to be un-checked-out.
- [ ] Before the "origin/<parent> still exists" check, refresh/prune the remote-tracking ref (e.g. `git fetch --prune` or `git ls-remote` against the actual remote) so a stale local ref doesn't produce a false warning.
- [ ] A regression test covering the child-branch-checked-out-in-a-worktree case (the exact failure above) and the stale-origin-ref case.
- [ ] The `merge-pr.sh` auto-reconcile call path (#3747 item 1 / #3752) inherits the fix — verify a live stacked child with an existing worktree reconciles cleanly.

## Relation
- #3747 (item 1, shipped in #3752) — the auto-reconcile path that invokes this helper.

rjwalters:
Closes #3776

## Summary

Fixes two defects in `reconcile-stack.sh` hit live 2026-07-22 while collapsing a stacked-PR stack.

### 1. Worktree-checked-out child branch (the blocker)
A Loom child branch is always checked out in its own managed worktree (`.loom/worktrees/issue-<child>`), and git refuses to rebase a branch checked out elsewhere (`fatal: '<branch>' is already used by worktree at ...`). The helper could therefore never run from the main worktree for a live Loom stack.

The script now detects the worktree holding the child branch via `git worktree list --porcelain` and runs the `rebase --onto` + `push --force-with-lease` **inside that worktree** (`git -C <worktree>`). When no worktree holds the branch (already removed), it falls back to the original in-place rebase — unchanged v1 behaviour. The dirty-tree check now targets whichever tree the rebase will actually run in.

### 2. False "origin/<parent> still exists" warning on a stale local ref
With delete-branch-on-merge the parent branch is deleted on the remote the instant it squash-merges, but the local `refs/remotes/origin/<parent>` can linger stale until a prune, producing a spurious warning on every post-merge reconcile. The advisory now queries the remote **live** via `git ls-remote --exit-code --heads` instead of trusting the possibly-stale local ref.

### merge-pr.sh auto-reconcile path (#3747 item 1 / #3752)
The auto-reconcile path invokes `reconcile-stack.sh` with the same `<child-pr> <parent-branch>` signature, so it inherits both fixes with no change to `merge-pr.sh`. The existing `test-merge-pr-auto-reconcile.sh` suite stays green (20/20), including its "reuses reconcile-stack.sh unmodified" source guard.

## Tests
New `defaults/scripts/tests/test-reconcile-stack.sh` — an offline git-sandbox regression test (bare remote + clone, real stacked parent/child, simulated squash-merge, parent deleted in the bare remote to leave a stale local ref):
- **Scenario A**: child checked out in a worktree, run from the main worktree — asserts a clean rebase inside the worktree, the parent's pre-squash commit stripped, no `already used by worktree` fatal, and no false stale-origin warning.
- **Scenario B**: no worktree holds the branch — asserts the in-place fallback still works.
- Source guards for both fixes.

Results: `test-reconcile-stack.sh` 15/15 passed; `test-merge-pr-auto-reconcile.sh` 20/20 passed; `shellcheck -x` clean on both the script and the new test.

### DIFF
--- a/defaults/scripts/reconcile-stack.sh
+++ b/defaults/scripts/reconcile-stack.sh
@@ -122,15 +122,53 @@ info "Rebase target (default branch): $DEFAULT_BRANCH"
 # Best-effort advisory: has the parent branch squash-merged? If origin still has
 # the parent branch, the operator likely hasn't merged it yet — warn but do not
 # block (the operator asserts they've confirmed the merge).
+#
+# Query the remote LIVE (git ls-remote) rather than trusting the local
+# remote-tracking ref: with delete-branch-on-merge the parent branch is deleted
+# on the remote the instant it squash-merges, but the local
+# refs/remotes/origin/<parent> can linger stale until a prune — a plain
+# `git show-ref` of that stale ref then false-warns on every post-merge
+# reconcile (#3776). ls-remote asks the actual remote, so a stale local ref
+# never produces a spurious warning. A network/ls-remote failure simply skips
+# the advisory (it was only ever advisory).
 git fetch origin "$DEFAULT_BRANCH" 2>/dev/null || true
-if git show-ref --verify --quiet "refs/remotes/origin/$PARENT_BRANCH" 2>/dev/null; then
+if git ls-remote --exit-code --heads origin "refs/heads/$PARENT_BRANCH" >/dev/null 2>&1; then
     warn "origin/$PARENT_BRANCH still exists — confirm the parent PR has squash-merged before reconciling."
     warn "(delete-branch-on-merge normally removes it once the parent merges.)"
 fi
 
-# Refuse on a dirty working tree — a rebase would fail confusingly.
-if [[ -n "$(git status --porcelain 2>/dev/null)" ]]; then
-    err "Working tree is dirty. Commit, stash, or discard changes before reconciling."
+# Locate the worktree (if any) that holds the child branch checked out. A Loom
+# child branch is ALWAYS checked out in its own managed worktree
+# (.loom/worktrees/issue-<child>), and git refuses to rebase a branch that is
+# checked out in another worktree (`fatal: '<branch>' is already used by
+# worktree at ...`) — so running the rebase from the main worktree can never
+# work for a live Loom stack (#3776). Detect that worktree via
+# `git worktree list --porcelain` and run the rebase/push INSIDE it. When no
+# worktree holds the branch (e.g. it was already removed), fall back to the
+# in-place rebase, which checks the branch out in the current worktree — the
+# original v1 behaviour, unchanged.
+CHILD_WORKTREE=""
+_wt=""
+while IFS= read -r _line; do
+    case "$_line" in
+        "worktree "*) _wt="${_line#worktree }" ;;
+        "branch refs/heads/$CHILD_BRANCH") CHILD_WORKTREE="$_wt" ;;
+    esac
+done < <(git worktree list --porcelain 2>/dev/null || true)
+
+# GIT_C is the git invocation used for the branch-mutating steps (rebase, push).
+# Inside the child worktree when one exists, else the current worktree.
+if [[ -n "$CHILD_WORKTREE" ]]; then
+    info "Child branch $CHILD_BRANCH is checked out in worktree: $CHILD_WORKTREE — running the rebase there."
+    GIT_C=(git -C "$CHILD_WORKTREE")
+else
+    GIT_C=(git)
+fi
+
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED ✅

Reviewed against issue #3776's acceptance criteria. Both defects are fixed correctly, with a genuine regression test.

### Criterion 1 — rebase inside the child's worktree
`reconcile-stack.sh` now parses `git worktree list --porcelain` to find the worktree holding `refs/heads/<child>`, and runs both the `rebase --onto` and `push --force-with-lease` via `git -C <worktree>` (`GIT_C` array). When no worktree holds the branch it falls back to the bare `git` in-place path — original v1 behaviour preserved. The dirty-tree check correctly targets whichever tree the rebase will actually run in. The porcelain parse handles worktree paths with spaces and leaves detached worktrees unmatched. ✓

### Criterion 2 — live stale-origin check
The advisory swapped `git show-ref --verify refs/remotes/origin/<parent>` (trusts a possibly-stale local ref) for `git ls-remote --exit-code --heads origin refs/heads/<parent>`, which queries the remote live, so a lingering local remote-tracking ref no longer false-warns. Failures degrade to skipping the advisory — it was only ever advisory. ✓

### Criterion 3 — regression test
New `defaults/scripts/tests/test-reconcile-stack.sh` builds a real offline git sandbox (bare remote + clone, stacked parent/child, simulated squash-merge, parent deleted **in the bare remote** to leave a genuinely stale local ref). It covers all three required cases: Scenario A (child checked out in a worktree, run from main worktree — asserts clean rebase inside the worktree, parent's pre-squash commit stripped, no `already used by worktree` fatal, no false stale-origin warning), Scenario B (no-worktree in-place fallback), and the stale-origin false-warn in both. Source guards fail loudly if either fix is dropped. ✓

### Criterion 4 — merge-pr.sh auto-reconcile path
CLI contract is unchanged (same `<child-pr> <parent-branch>` positional args), so the `merge-pr.sh` auto-reconcile call site inherits both fixes with no change. `test-merge-pr-auto-reconcile.sh` stays green, including its 'reuses reconcile-stack.sh unmodified' source guard. ✓

### Test results (run from the issue worktree)
- `test-reconcile-stack.sh`: **15/15 passed**
- `test-merge-pr-auto-reconcile.sh`: **20/20 passed**
- `shellcheck -x` on the script + new test: **clean**
- `bash -n`: syntax OK
- mergeStateStatus: **CLEAN / MERGEABLE**

Clean, well-scoped, well-tested fix. Approving.

PR #3892, opened 2026-07-23
fix(daemon): serialize rapid dispatch + startup watchdog for MCP-init hang (#3887)

### LINKED ISSUE #3887: daemon: sweeps hang at startup on rapid back-to-back dispatch (0-HTTPS MCP-init race); add self-healing watchdog
## Summary

When `loom-daemon` dispatches multiple sweeps in quick succession (back-to-back `DispatchSweep`), some child `claude` processes **hang at startup**: the sweep log contains only the spawn header (`spawn-claude: using OAuth account '<tok>'`), no worktree is ever created, the issue never leaves `loom:issue`/`loom:building`, and the process sits at ~0% CPU with **0 open HTTPS connections** (so it is NOT waiting on the API / rate limit — it never reaches the API). It has a spawned `mcp-loom` node child. Re-dispatching the same issue (fresh process) reliably clears it.

## Observations (this dogfood session, ~6 incidents)

- Always affects sweeps dispatched within ~1s of each other (rapid back-to-back).
- 0 HTTPS connections on the hung `claude` → it is stuck BEFORE any API call, i.e. in local startup / MCP handshake, not rate-limited.
- Correlated with (but not exclusively caused by) specific tokens reappearing (agent2-2amlogic, robb-2amlogic) — likely coincidence of random selection under rapid dispatch rather than a bad token, given the 0-HTTPS signature.
- Recovery: `CancelSweep` + re-`DispatchSweep` → the retried process starts normally and builds.

## Hypothesis

A startup race when multiple `claude`/`mcp-loom` processes initialize simultaneously — candidates: contention on `mcp-loom/dist` (build/read), the token `index.json`/`.ranking`/`.bad_tokens` files, a shared config/keychain access, or the MCP stdio handshake deadlocking. Needs a live `sample`/`dtruss` of a freshly-hung child to localize.

## Acceptance criteria
- [ ] Root-cause the startup hang (attach to a hung child; identify the blocking resource).
- [ ] Fix the race so rapid back-to-back dispatch never hangs (e.g. serialize the contended startup step, add a small stagger in the dispatcher, or make the MCP handshake timeout+retry).
- [ ] Add a daemon-side **startup watchdog**: if a spawned sweep produces no progress (no worktree / no log past the spawn header) within N seconds, auto-cancel and re-dispatch once (bounded), logging the event — so a hang self-heals instead of silently wedging an issue.
- [ ] Regression test / repro harness dispatching K sweeps simultaneously and asserting all reach the build phase.

## Context
Found while dogfooding autonomous daemon mode (this session). Recovered manually ~6 times via cancel+re-dispatch. This matters for the #3809 autonomous work-finder, which dispatches in bursts and would silently wedge issues without the watchdog.

rjwalters:
Closes #3887

## Root-cause analysis

When `loom-daemon` dispatches several sweeps back-to-back (the autonomous work
finder draining a `loom:issue` backlog in a single tick), each spawned
`claude -p "/loom:sweep N"` child immediately forks its own `mcp-loom` node
child and performs Claude Code's **local startup** (config + keychain read) plus
the **MCP stdio handshake** *before* its first API call. When many of those
startups run simultaneously (all within ~1s), some children wedge in that
pre-API phase: the sweep log shows only the daemon spawn header + the
`spawn-claude: using OAuth account` line, no worktree is created, the process
sits at ~0% CPU with **zero** open HTTPS connections, and the issue never leaves
`loom:building`. Re-dispatching the same issue as a fresh process reliably
clears it — the smoking gun that it is a *startup* race, not a rate-limit.

I ruled out the token-selection files as the culprit: `select.py` only **reads**
`.ranking` / `.bad_tokens` / `index.json` at spawn time (concurrent reads are
safe), the one writer path (`.bad_tokens`) is already `mkdir`-lock guarded and
atomic, and a read race would mis-*select* a token, never hang — and the hang is
observed *after* the account line is already logged. The contention is the
simultaneous MCP-init / local-startup itself. Analysis is documented in the
module-level comment in `sweep_registry.rs`.

## The fix (two layers)

**1. Dispatch stagger (prevention).** The registry enforces a minimum
wall-clock gap between consecutive child spawns (`apply_dispatch_stagger`,
default **2000ms**). Applied under the registry mutex in `dispatch()`, so a
burst of K dispatches becomes K spawns spaced `stagger` apart instead of K
near-simultaneous ones — and concurrent dispatch callers are serialized past the
contended startup step. Defaults to zero in unit tests (no added latency);
`main.rs` sets the resolved value on the production registry.

**2. Startup watchdog (self-heal backstop).** A background task
(`spawn_watchdog_task`, default probe every 30s) runs the pure
`watchdog_decision` state machine per running daemon-dispatched sweep. Progress
is probed via `sweep_made_progress`: worktree at `.loom/worktrees/issue-<N>`,
checkpoint at `.loom/sweep-checkpoint/issue-<N>.json`, or log output past the
spawn header (`log_has_progress`). A sweep with **no** progress past the timeout
(default **120s**) is auto-cancelled and re-dispatched **exactly once** —
bounded, never a loop; a second hang resolves to `GiveUp` and is left for the
operator. Only sweeps this daemon actually spawned (owns the `Child` handle) are
eligible.

Both the auto-cancel and the retry **log loudly** and reuse only the **frozen**
event topics — `sweep.issue.{N}.exited` / `sweep.global.completed` (from the
cancel path) and `sweep.global.dispatch` (from re-dispatch). **No new event
topics** were introduced.

## Configuration (env > config > default)

New `.loom/config.json → autonomous` knobs, each env-overridable:

| Config key | Env override | Default |
|------------|--------------|---------|
| `autonomous.dispatchStaggerMs` | `LOOM_SWEEP_DISPATCH_STAGGER_MS` | `2000` (`0` disables) |
| `autonomous.watchdog.enabled` | `LOOM_SWEEP_WATCHDOG` | `true` |
| `autonomous.watchdog.timeoutSecs` | `LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` | `120` |
| `autonomous.watchdog.intervalSecs` | `LOOM_SWEEP_WATCHDOG_INTERVAL_SECS` | `30` |

An absent `autonomous` block soft-fails to env/default (zero behavior change).
The watchdog defaults **on** (self-healing intent); disable with
`LOOM_SWEEP_WATCHDOG=0`.

## Testing

`cd loom-daemon && cargo test --lib` → **606 passed; 0 failed** (23 new).
`cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` both clean.

New tests cover:
- `stagger_wait` / `watchdog_decision` / `log_has_progress` pure state machines
- `sweep_made_progress` filesystem probes (worktree / checkpoint / log)
- config-resolution precedence (env > config > default) for every knob
- `dispatch()` applies the configured stagger between two real spawns
- **bounded end-to-end watchdog**: a hung fixture child → restarted once →
  give-up on the second hang, plus "progressing sweep is left alone"

The K-simultaneous real-`claude` harness is intentionally out of CI (per the
issue) — the daemon-side logic is tested with fakes, matching the existing
`sweep_registry` tests.

## Notes
- Changes are confined to `loom-daemon/` + a `.loom/docs/daemon-reference.md`
  config-table update. No `defaults/scripts/` changes.

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -566,6 +566,12 @@ concurrency ceiling 5" and share it with the team:
     },
     "mainHealthGate": {
       "enabled": true
+    },
+    "dispatchStaggerMs": 2000,
+    "watchdog": {
+      "enabled": true,
+      "timeoutSecs": 120,
+      "intervalSecs": 30
     }
   }
 }
@@ -585,6 +591,24 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.workFinder.intervalSecs` | `LOOM_WORK_FINDER_INTERVAL_SECS` | `60` | Zero/invalid → default |
 | `autonomous.workFinder.maxConcurrent` | `LOOM_WORK_FINDER_MAX_CONCURRENT` | `3` | Operator **ceiling**, not a fixed target |
 | `autonomous.mainHealthGate.enabled` | `LOOM_MAIN_HEALTH_GATE` | `false` | Gate loop on/off |
+| `autonomous.dispatchStaggerMs` | `LOOM_SWEEP_DISPATCH_STAGGER_MS` | `2000` | Min gap between consecutive child spawns (#3887). `0` disables |
+| `autonomous.watchdog.enabled` | `LOOM_SWEEP_WATCHDOG` | `true` | Startup watchdog on/off (#3887) |
+| `autonomous.watchdog.timeoutSecs` | `LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` | `120` | No-progress window before auto-restart |
+| `autonomous.watchdog.intervalSecs` | `LOOM_SWEEP_WATCHDOG_INTERVAL_SECS` | `30` | Watchdog probe cadence |
+
+**Startup-race mitigation (#3887).** Rapid back-to-back dispatch (the work
+finder draining a backlog in one tick) could wedge some `claude` children at
+startup in a 0-HTTPS MCP-init race: the sweep log showed only the spawn header,
+no worktree was created, and the issue never left `loom:building`. Two layers
+now guard against it: the **dispatch stagger** spaces consecutive child spawns
+out of the simultaneous-startup window (prevention), and the **startup
+watchdog** probes each running sweep for progress (worktree created / checkpoint
+written / log output past the spawn header) and auto-cancels + re-dispatches —
+**exactly once, bounded, never a loop** — any sweep hung with no progress past
+`timeoutSecs`. Both the auto-cancel and the retry log loudly and reuse the
+frozen `sweep.issue.{N}.exited` / `sweep.global.completed` / `sweep.global.dispatch`
+topics (no new event topics). The watchdog defaults **on**; disable it with
+`LOOM_SWEEP_WATCHDOG=0` or `autonomous.watchdog.enabled = false`.
 
 The **gate's behavior** (which command runs against `main`, its timeout) still
 comes from the separate top-level `buildGate` block (#3749); `autonomous.mainHealthGate`
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -242,8 +242,33 @@ async fn main() -> Result<()> {
         ),
         Err(e) => log::warn!("sweep_registry: reconstruction failed: {e}"),
     }
+
+    // Startup-race mitigation (Issue #3887): resolve the dispatch stagger + the
+    // watchdog knobs from `.loom/config.json → autonomous` with env override
+    // (precedence env > config > default). The stagger serializes back-to-back
+    // child startups so a burst dispatch does not trip the 0-HTTPS MCP-init
+    // race; the watchdog is the self-healing backstop for any hang that slips
+    // past it.
+    let startup_race_config = sweep_registry::read_startup_race_config(&sweep_workspace);
+    let dispatch_stagger = sweep_registry::resolve_dispatch_stagger(&startup_race_config);
+    sweep.set_dispatch_stagger(dispatch_stagger);
+    log::info!("sweep_registry: dispatch stagger = {}ms (#3887)", dispatch_stagger.as_millis());
+
     let sweep_registry = Arc::new(Mutex::new(sweep));
     let _reaper_handle = sweep_registry::spawn_reaper_task(sweep_registry.clone());
+
+    // Startup watchdog (Issue #3887): auto-cancel + re-dispatch (once, bounded)
+    // any daemon-dispatched sweep that hangs at startup with no progress. On by
+    // default; disable with LOOM_SWEEP_WATCHDOG=0 or
+    // `autonomous.watchdog.enabled = false`.
+    let _watchdog_handle = if sweep_registry::resolve_watchdog_enabled(&startup_race_config) {
+        let timeout = sweep_registry::resolve_watchdog_timeout(&startup_race_config);
+        let interval = sweep_registry::resolve_watchdog_interval(&startup_race_config);
+        Some(sweep_registry::spawn_watchdog_task(sweep_registry.clone(), timeout, interval))
+    } else {
+        log::info!("sweep_registry: startup watchdog disabled (#3887)");
+        None
+    };
 
     // Epic supervisor loop (Issue #3872 — Phase 4 of epic #3842). Opt-in via
     // `LOOM_EPIC_SUPERVISOR`. The loop drives every open `loom:epic` issue
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -55,7 +55,7 @@ use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 // ============================================================================
 // Constants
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Rigorous review of the #3887 startup-race fix. This is a high-quality, well-documented, thoroughly-tested PR that satisfies all three acceptance criteria. Confined to `loom-daemon/` + a config-table doc update.

## Verification
- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean (0 warnings)
- `cargo test --lib` (stagger + watchdog subsets) — 16/16 new tests pass; CI "Rust Unit Tests" + "Rust Linting & Build Check" both green
- `mergeStateStatus` — CLEAN / MERGEABLE
- PR body uses `Closes #3887`

## 1. Serialization fix (stagger) — correct, not over-serialized
`apply_dispatch_stagger` enforces a min 2000ms wall-clock gap between consecutive child spawns (`stagger_wait` pure fn), applied under the registry mutex in `dispatch()` step 5 immediately before spawn. Critically it serializes **only the brief spawn/startup step**, not the sweep lifecycle — the child runs as a detached process and `dispatch()` returns, so K bursty dispatches still run K concurrent sweeps; only their *spawns* are spaced out of the ~1s simultaneous-MCP-init window. This does **not** defeat the Phase A/B work-finder concurrency (`maxConcurrent`). Default zero in tests (byte-for-byte pre-#3887 behavior); `main.rs` sets the resolved env>config>default value on the production registry.

Minor (non-blocking): the stagger uses `std::thread::sleep` while holding the std Mutex, so a burst briefly stalls other registry consumers (reaper/IPC). This matches the established pattern in `cancel()`'s grace poll ("Blocking sleep is fine here") and a ≤2s stall on a burst is acceptable.

## 2. Watchdog — bounded, false-positive-safe, clears the lock
**Bounded (retry cap = exactly 1):** `watchdog_retried: HashSet<u32>` is marked *before* the cancel/dispatch acts, so even an error path counts the single allowed attempt and never loops. A second hang past the timeout resolves to `GiveUp` (logged once via `watchdog_gaveup`) and is left for the operator. Verified end-to-end by `watchdog_restarts_hung_sweep_once_then_gives_up` (hung child → restart once → give up on the second hang).

**False-positive-safe:** `sweep_made_progress` treats a sweep as live on ANY of — worktree at `.loom/worktrees/issue-<N>`, checkpoint at `.loom/sweep-checkpoint/issue-<N>.json`, or `log_has_progress` (output past the spawn header). It keys on the durable **worktree/checkpoint artifacts**, not just log lines, and `log_has_progress` anchors to the *last* dispatch header so a reused per-issue log's prior-run output is never miscounted. Any progress ⇒ `Healthy` regardless of elapsed time, so a legitimately slow-curating sweep (minutes) is never killed — confirmed by `watchdog_leaves_progressing_sweep_alone` (worktree present + backdated 9999s ⇒ 0 restarts). The 120s default timeout is generous relative to when a healthy sweep first emits Curator output. Only sweeps this daemon actually spawned (owns the `Child` handle) are eligible; reconstructed entries are left to the reaper.

**Clears the claim lock on cancel (the manual failure you hit):** the `Restart` path calls `self.cancel(sweep_id, WATCHDOG_CANCEL_GRACE)` → `finish_cancel`, which `release_lock(issue)` (removes `.loom/locks/issue-<N>/`) and restores `loom:building → loom:issue` via `restore_label_to_ready` (gated on `!skip_label_flip && !produced_pr`, so it never yanks a label out from under an in-flight PR). The subsequent `self.dispatch()` re-`acquire_lock`s the now-freed lock dir cleanly — no lock collision. Re-dispatch params (model/effort/depends_on/idempotency_key) are captured from the hung entry *before* cancel mutates it.

**Event hygiene:** both auto-cancel and retry log loudly and reuse only the frozen topics — `sweep.issue.{N}.exited` / `sweep.global.completed` (cancel) and `sweep.global.dispatch` (re-dispatch). No new topics.

## 3. Token-ranking freshness — out of scope here (noted)
The PR explicitly rules out the token-selection files as the *hang* mechanism (correct: `select.py` only reads them; the sole writer is mkdir-lock guarded; a read race mis-selects but never hangs, and the hang is observed *after* the account line is logged). It addresses the race (stagger) + self-heal (watchdog) only, **not** token-ranking/`.ranking` freshness or exhausted-account selection. That's acceptable for #3887's acceptance criteria, which are fully met. Worth a follow-up note: for the exhausted-token variant, the watchdog re-dispatch reuses the same 3-tier selection, so with a stale/missing `.ranking` it could random-select into another exhausted account, hang again, and `GiveUp` after the one bounded retry — partial mitigation, not a full fix. Not blocking; suggest a follow-up issue for `.ranking` freshness / exhausted-pool handling if the live exhausted-token pattern recurs.

## Config
`env > config > default` for all four knobs (`dispatchStaggerMs`, `watchdog.{enabled,timeoutSecs,intervalSecs}`), matching `WorkFinderConfig`; an absent `autonomous` block soft-fails to defaults (zero behavior change). A `0` stagger is correctly preserved as a "disable" value (not filtered to `None`), while zero timeout/interval are dropped. Precedence is unit-tested for every knob. Watchdog defaults **on** (self-healing intent), disableable via `LOOM_SWEEP_WATCHDOG=0`.

Tests are excellent — pure state machines (`stagger_wait`/`watchdog_decision`/`log_has_progress`), filesystem probes, config precedence, a real two-spawn stagger integration test, and the bounded end-to-end restart. The K-simultaneous real-`claude` harness is (per the issue) intentionally out of CI. Nice work.

rjwalters:
Post-merge Judge review (PR was already merged as 8c96c893 with loom:pr before this review ran). Reviewed the full diff against issue #3887 acceptance criteria:

- **Root cause**: Documented as a module-level comment in sweep_registry.rs — simultaneous claude/mcp-loom local startup + MCP stdio handshake contention; token-file races explicitly ruled out with sound reasoning (select.py only reads at spawn; the one writer path is mkdir-lock guarded and atomic; hang is observed *after* the account line logs). Code-inspection analysis is acceptable per the issue.
- **Bounded-retry invariant**: Verified. `watchdog_retried: HashSet<u32>` is keyed on issue number and `insert`ed BEFORE cancel/dispatch, so even an error path counts the single allowed attempt. `watchdog_decision` returns Restart only when `!already_retried`; a second hang resolves to GiveUp. Cannot loop more than once per issue.
- **Stagger serialization**: `apply_dispatch_stagger` sleeps under the registry mutex via `std::thread::sleep`, correctly serializing concurrent dispatch callers past the contended startup step. Consistent with the existing under-lock blocking (token-name capture already blocks up to 5s under the same std::sync::Mutex); single non-reentrant mutex, no lock-ordering or deadlock risk. Zero stagger is a no-op default.
- **Progress probe**: `log_has_progress` anchors to the LAST dispatch header (avoids counting prior-run output). Any worktree/checkpoint/log-past-header = progress ⇒ Healthy. False-positive directions (lingering prior worktree/checkpoint) fail conservatively (won't kill a live sweep); the 120s default is generous vs. normal sub-minute Curator-phase output. Acceptable tradeoff.
- **Event topics**: No new topics — cancel reuses finish_cancel's frozen exited/completed emission, re-dispatch reuses sweep.global.dispatch. grep confirms only frozen topics present.
- **Config precedence**: env > config > default correct across all four knobs; stagger honors 0 as disable, timeout/interval drop 0 to None. Wired in main.rs, documented in daemon-reference.md.
- **Tests**: 23 new tests exercise the pure state machine, the filesystem probe, config precedence (#[serial]), AND an end-to-end bounded-restart path with a real hung fixture child (restart-once → mark retried → give-up on second hang). Substantial, not trivial.

Ran locally on the merge commit: `cargo test --lib` → **606 passed, 0 failed**; `cargo clippy --lib --all-targets -- -D warnings` → clean. Approved.

PR #3905, opened 2026-07-23
fix(daemon): reap-on-read so ListSweeps no longer over-reports Running (#3893)

### LINKED ISSUE #3893: daemon: reaper doesn't promptly prune completed/merged sweeps — registry over-reports Running
## Context

While dogfooding daemon dispatch this session, `mcp__loom__list_sweeps` / the `ListSweeps` IPC repeatedly reported sweeps as `Running` long after their PRs had merged and their processes had exited. Across ~30 merges the registry accumulated ~6+ stale `Running` entries (e.g. #3810, #3811, #3866, #3868, #3872, #3873 all showed `Running` after merge). The reaper does eventually clean them but not promptly.

## Impact
- `ListSweeps` over-reports active work, making the registry misleading for an operator (and for a human watching status).
- Any concurrency accounting derived from the live registry (e.g. the Phase B/D work-finder occupancy seed, `dispatcher.in_flight().len()`) could over-count, deferring legitimate new dispatch.

## Acceptance criteria
- [ ] Reaper prunes exited/merged sweeps from the registry promptly (bound the staleness window; e.g. reap on child exit via `try_wait`, not only on a slow timer).
- [ ] `ListSweeps` reflects only genuinely-live sweeps (Running/Pending), or clearly distinguishes terminal states.
- [ ] Test: a sweep whose child has exited is no longer reported `Running` within N seconds.

## Notes
Found during the daemon dogfood; complements the startup-hang watchdog (#3892) — both are about the registry/reaper accurately reflecting sweep liveness.

rjwalters:
## Summary

`mcp__loom__list_sweeps` / the `ListSweeps` IPC repeatedly reported sweeps as `Running` long after their PRs merged and their processes exited. The registry only transitioned exited/merged sweeps out of `Running` on the **30s reaper timer**, so a read taken between a child's exit and the next tick over-reported active work. Across a burst of merges the registry accumulated stale `Running` entries, and any concurrency accounting derived from the live registry (the work-finder occupancy seed / `in_flight()`) could over-count and defer legitimate new dispatch.

## Design

**Reap-on-read.** New `SweepRegistry::reap_liveness()` performs exactly the same `try_wait`-based liveness poll + terminal transition (and best-effort label/event side effects) the background timer does. It is called before every read path:

- `ListSweeps` IPC handler
- `GetSweepStatus` IPC handler
- work-finder occupancy seed (`RegistryDispatcher::in_flight`)

This **bounds the staleness window to the read itself**: a sweep whose child has exited is reaped promptly instead of after up to 30s, so a read never over-reports a genuinely-exited child as `Running`. On a steady-state read with no newly-exited children it is just one cheap `try_wait` per running entry with no side effects. The 30s reaper timer remains as the background GC fallback (it also drops terminal entries past the 1h retention window). `SweepState` already distinguishes terminal states (`Exited` / `Crashed`), so recently-exited sweeps still appear in the list — just no longer as stale `Running`.

No new event topics were introduced (the taxonomy is frozen); this reuses `reap_once`'s existing liveness/transition/event logic.

## Tests

- New `read_path_reaps_exited_child_out_of_running_promptly`: dispatches a fake child that exits immediately, then asserts a read-path reconcile transitions it out of `Running` (to terminal `Exited`) within ~2s — the acceptance-criterion "exited child is no longer reported Running within N seconds."
- Updated `test_handle_request_dispatch_sweep_happy_path`: with reap-on-read, the instantly-exiting fake child is now correctly reported terminal (still listed, `len == 1`), not stale-`Running`.

Full `cargo test` suite passes (610 lib tests + integration suites, 0 failures); `cargo fmt` applied and `cargo clippy --all-targets` is clean.

Complements the startup-hang watchdog (#3892) — both are about the registry/reaper accurately reflecting sweep liveness.

Closes #3893

### DIFF
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -1075,9 +1075,13 @@ fn handle_request(
         }
 
         Request::ListSweeps { state_filter } => {
-            let sr = sweep_registry
+            let mut sr = sweep_registry
                 .lock()
                 .expect("Sweep registry mutex poisoned");
+            // Reap-on-read (Issue #3893): reconcile liveness before listing so a
+            // sweep whose child has already exited is never reported `Running`
+            // just because the 30s reaper timer has not ticked yet.
+            sr.reap_liveness();
             let sweeps = sr.list(state_filter.as_ref());
             Response::SweepList { sweeps }
         }
@@ -1086,9 +1090,12 @@ fn handle_request(
         // Sweep Monitoring Handlers (Issue #3455 — Phase C of #3449)
         // ====================================================================
         Request::GetSweepStatus { sweep_id } => {
-            let sr = sweep_registry
+            let mut sr = sweep_registry
                 .lock()
                 .expect("Sweep registry mutex poisoned");
+            // Reap-on-read (Issue #3893): reconcile liveness so a status query
+            // reflects a child that has exited rather than a stale `Running`.
+            sr.reap_liveness();
             let info = sr.get_status(&sweep_id);
             Response::SweepStatus { info }
         }
@@ -1196,7 +1203,7 @@ mod tests {
     use super::*;
     use crate::activity::ActivityDb;
     use crate::sweep_registry::{SweepRegistry, SweepRegistryConfig};
-    use crate::types::{SweepKind, SweepState};
+    use crate::types::SweepKind;
     use tempfile::tempdir;
 
     type TestContext = (
@@ -1453,13 +1460,21 @@ exit 0
             other => panic!("Expected SweepDispatched, got: {other:?}"),
         }
 
-        // Follow-up ListSweeps should see the new entry.
+        // Follow-up ListSweeps should see the new entry. The fake spawn exits
+        // immediately, so reap-on-read (Issue #3893) promptly reconciles the
+        // entry to a terminal `Exited` state rather than over-reporting it as
+        // `Running` — the entry is still listed, just no longer stale-Running.
         let response =
             handle_request(Request::ListSweeps { state_filter: None }, &tm, &db, &sr, &bus);
         match response {
             Response::SweepList { sweeps } => {
                 assert_eq!(sweeps.len(), 1);
-                assert!(matches!(sweeps[0].state, SweepState::Running));
+                assert!(
+                    sweeps[0].state.is_terminal(),
+                    "reap-on-read should have transitioned the exited fake child \
+                     out of Running (#3893); got {:?}",
+                    sweeps[0].state
+                );
             }
             other => panic!("Expected SweepList, got: {other:?}"),
         }
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -1657,6 +1657,25 @@ impl SweepRegistry {
         changes
     }
 
+    /// Promptly reconcile sweep liveness on a **read path** (Issue #3893).
+    ///
+    /// `ListSweeps` / `GetSweepStatus` / the work-finder occupancy seed call
+    /// this before reading, so a caller never observes a sweep as `Running`
+    /// after its child has already exited. Before #3893 the only path out of
+    /// `Running` was the 30s [`reap_once`](Self::reap_once) timer, so a read
+    /// taken between a child's exit and the next tick over-reported active work
+    /// (the registry accumulated stale `Running` entries across a burst of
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review — APPROVED

Reviewed the reap-on-read design against issue #3893's acceptance criteria. Clean, well-scoped, well-documented.

**Correctness**
- No re-lock/deadlock: `reap_liveness` operates on `&mut self` where the IPC handler / `in_flight` caller already holds the registry mutex — it delegates straight to `reap_once` with no re-entrant lock, exactly the pattern the timer task uses.
- `try_wait` semantics correct: `poll_liveness` removes the retained `Child` handle on `Ok(Some(status))`, so a subsequent poll can't double-count an already-terminal entry.
- No double-reap race with the 30s timer: both the timer and every read path serialize on `Mutex<SweepRegistry>`, so only one `reap_once` runs at a time and the handle is consumed on first success.
- Side effects reuse `reap_once`'s existing logic — **no new event topics** (taxonomy stays frozen for v0.10.0).

**Completeness vs. AC**
- Prompt prune / bounded staleness window: reap-on-read bounds it to the read itself. ✓
- ListSweeps reflects only genuinely-live sweeps and still distinguishes terminal `Exited`/`Crashed`. ✓
- Test asserts the bounded window (~2s loop, then terminal-state + no-Running assertions). ✓

**Verification**
- `cargo test --lib`: 610 passed, 0 failed (incl. new `read_path_reaps_exited_child_out_of_running_promptly` and updated `test_handle_request_dispatch_sweep_happy_path`).
- `cargo clippy --all-targets`: clean.
- CI: Rust Lint + Unit Tests pass; merge state CLEAN.

**Minor (non-blocking) note for a possible follow-up**: on the transition tick, `reap_once` can invoke the blocking `gh` label-restore (orphaned-claim / crashed-checkpoint branches) while holding the registry mutex — now reachable from IPC read handlers, not just the background timer. It's identical to the pre-existing timer behavior and fires at most once per exited orphaned-claim child, so it's acceptable; worth keeping in mind only if IPC read latency ever becomes sensitive.

LGTM! Approved.

rjwalters:
✅ **Approved!**

Reviewed the reap-on-read fix for #3893. This is a clean, minimal, well-targeted implementation.

## Correctness (acceptance criteria)
- **Prompt pruning of exited children.** `SweepRegistry::reap_liveness()` delegates to the existing `reap_once()`, which polls each `Running`/`Pending` entry via the retained `Child` handle's `try_wait()` (falling back to `kill(pid,0)`), transitions dead children to terminal `Exited`/`Crashed`, and runs the same best-effort label/event side effects. Wired into all three read paths: `ListSweeps`, `GetSweepStatus`, and the work-finder occupancy seed (`RegistryDispatcher::in_flight`). Staleness window is now bounded to the read itself rather than the 30s timer. ✔
- **ListSweeps reflects only genuinely-live sweeps.** A child that has exited is no longer over-reported as `Running`; recently-exited sweeps still appear, distinguished by their terminal state (`SweepState::is_terminal()`). ✔
- **Regression test present.** `read_path_reaps_exited_child_out_of_running_promptly` dispatches an immediately-exiting fake child and asserts a read-path reconcile transitions it out of `Running` (to terminal `Exited`) within ~2s, and that it no longer counts as in-flight. `test_handle_request_dispatch_sweep_happy_path` updated accordingly. ✔

## Safety review
- **No race / double-remove with the background reaper.** Both the reaper task and every read-path caller acquire the same `Arc<Mutex<SweepRegistry>>`, so the two are serialized. `reap_once` is idempotent — terminal entries are skipped by the `!matches!(state, Running|Pending)` guard, so whichever caller wins the lock performs the one-shot transition and the other sees terminal state and no-ops. GC only drops terminal entries past the retention window. ✔
- **Live children are not dropped.** `poll_liveness` returns `is_dead=false` for a live child, so no transition occurs and the entry stays `Running`. Confirmed by the happy-path and out-of-band-kill tests. ✔
- The side effects (label restore / event emit) that read paths can now trigger under the lock are identical to what the 30s timer already did, fire at most once per exit (state guard), and are best-effort — acceptable, just moved earlier in time.

## Checks (run against the PR worktree)
- `cargo fmt --check` — clean
- `cargo test --lib` — **610 passed, 0 failed**
- `gh pr checks` — Rust Linting & Build **pass**, Rust Unit Tests **pass**
- Merge state: **CLEAN / MERGEABLE**

PR body correctly uses `Closes #3893`. Dead sweeps are now promptly pruned without dropping live ones. Nice work.

PR #3906, opened 2026-07-23
fix(daemon): sync main-health gate to origin/main + halt-gate epic supervisor (#3885)

rjwalters:
Closes #3885

Follow-up from PR #3883 (Issue #3812, Phase C of epic #3809).

## Problem
The reactive main-health gate (`loom-daemon/src/main_health_gate.rs`) ran `buildGate.command` in the daemon's `sweep_workspace` **as-is**, never syncing to `origin/main`. Autonomous merges land via the forge API (`merge-pr.sh`), which advances `origin/main` on the **remote** but never the daemon's local `main` checkout. So the gate could repeatedly test a stale snapshot:
- **Missed catch**: local `main` behind `origin/main` — the breaking merge is never in the tree the gate builds, so it stays green and never halts.
- **False halt**: local `main` carries operator edits / a stray branch — the gate goes red on unrelated state and halts dispatch spuriously.

## Approach
Chose the judge-suggested **hybrid** (verify-clean-then-fetch-and-reset, warn-and-skip when dirty) — the safe sweet spot that never clobbers operator edits.

- **`prepare_workspace_to_origin_main()`** runs before each gate command:
  1. Verify the workspace is on `main` (a detached HEAD or other branch -> skip, never reset an operator branch).
  2. Verify a clean tree via `git status --porcelain` (any local/untracked change -> skip, never hard-reset over edits).
  3. `git fetch origin main` (fetch failure / offline -> skip, rather than greenwash a stale tree).
  4. `git reset --hard origin/main` so the gate builds exactly what remote `main` now is. Only reached when on-main + clean, so the reset only ever fast-forwards the daemon's own `main` checkout.
- A skip yields a new **`GateOutcome::Skipped { reason }`** (and **`HealthTransition::Skipped`**) that leaves the halt flag **untouched** — an indeterminate run neither halts nor resumes. `CommandGateRunner` gains `without_sync()` for the command-classification unit tests.
- **Epic supervisor halt gating** (judge-flagged in #3883): the shared `MainHealthState` is threaded into `EpicSupervisor` via `with_health_gate`; `tick()` now skips the entire tick (no forge list, no dispatch) while `main` is red, so a halt stops epic-child dispatch too — not just work-finder sweeps. `main.rs` constructs `main_health_state` before the epic supervisor so both dispatch paths share the one flag.
[...]

PR #3915, opened 2026-07-23
fix(daemon): guard main-health gate reset against local-ahead main (#3912)

### LINKED ISSUE #3912: main-health gate: hard-reset-to-origin could discard local-ahead commits (edge-case guard)
## Context

Non-blocking note from the #3906 (main-health gate origin/main sync) review. The gate's `prepare_workspace_to_origin_main()` runs a hard reset to `origin/main` (guarded by on-`main` + clean-tree preconditions) before evaluating buildGate, so it tests real remote main. Edge case: if the daemon's clean local `main` were ever **ahead of** origin/main, the hard-reset-to-origin would discard those local-only commits (reflog-recoverable).

## Impact
Extreme/unlikely for a daemon workspace (which should only ever fast-forward its own `main`), but a hard reset that can drop commits is worth a guard.

## Acceptance criteria
- [ ] Before the hard reset in `prepare_workspace_to_origin_main()`, check that local `main` is NOT ahead of `origin/main` (e.g. `git rev-list --count origin/main..HEAD == 0`); if it is ahead, skip the reset (Skip → indeterminate) rather than discarding commits.
- [ ] Test: a local `main` ahead of origin is not hard-reset away.

## Notes
Low priority — flagged by the #3906 judge as reflog-recoverable and extreme-edge. Curator/builder may reasonably rescope or close if judged not worth the guard.

rjwalters:
## Summary

The main-health gate's `prepare_workspace_to_origin_main()` hard-resets the daemon workspace to `origin/main` before evaluating `buildGate`, so it tests the real remote `main`. Edge case flagged by the #3906 judge: if a **clean** local `main` were ever *ahead* of `origin/main`, the hard reset would discard those local-only commits (reflog-recoverable, but still data-losing).

This adds a guard: after the fetch (so the comparison reflects the freshly-fetched remote), it runs `git rev-list --count origin/main..HEAD`. If the count is non-zero, it returns `Skip` (indeterminate — leaves the halt flag untouched, exactly like the existing on-`main`/clean-tree precondition failures) rather than resetting. A `git`-failure on the rev-list also skips.

Extreme/unlikely for a daemon workspace (which should only ever fast-forward its own `main`), but worth guarding against a reset that can drop commits.

## Changes

- `loom-daemon/src/main_health_gate.rs`: new step 4 (ahead-of-origin check) in `prepare_workspace_to_origin_main()`, between fetch and hard-reset; doc comment renumbered.

## Test coverage

- New `test_prepare_skips_when_local_main_ahead_of_origin`: creates a clean local `main` carrying a commit `origin/main` lacks, asserts the outcome is `Skip` (reason contains "ahead") and that `HEAD` is unchanged (the local-only commit is not reset away). Mirrors the existing `test_prepare_skips_dirty_workspace` precondition test.
- Full module suite: 36 passed, 0 failed. `cargo fmt --check` and `cargo clippy --lib` clean.

Closes #3912

### DIFF
--- a/loom-daemon/src/main_health_gate.rs
+++ b/loom-daemon/src/main_health_gate.rs
@@ -396,9 +396,15 @@ fn run_git(repo_root: &Path, args: &[&str]) -> std::result::Result<(String, Stri
 ///    reset would clobber operator edits).
 /// 3. **Fetch** `origin main`. A fetch failure (offline, transient) ⇒ `Skip`
 ///    (better indeterminate than greenwashing a stale tree).
-/// 4. **Hard-reset** to `origin/main` so the gate builds exactly what the remote
-///    `main` now is. Only reached when the tree is on `main` and clean, so the
-///    reset only ever fast-forwards the daemon's own `main` checkout.
+/// 4. **Verify not ahead** of `origin/main`. A clean local `main` that carries
+///    commits `origin/main` lacks ⇒ `Skip` — the hard reset would discard those
+///    local-only commits (reflog-recoverable, but still). Extreme edge for a
+///    daemon workspace that should only ever fast-forward its own `main`, but
+///    worth guarding against a data-losing reset (Issue #3912).
+/// 5. **Hard-reset** to `origin/main` so the gate builds exactly what the remote
+///    `main` now is. Only reached when the tree is on `main`, clean, and not
+///    ahead, so the reset only ever fast-forwards the daemon's own `main`
+///    checkout.
 ///
 /// A `Skip` leaves the halt flag untouched (see [`apply_gate_outcome`]).
 #[must_use]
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review — APPROVED

Reviewed the local-ahead guard in `prepare_workspace_to_origin_main()`.

**Correctness**: The new step 4 is placed correctly — after the `git fetch` (so `origin/main` reflects the freshly-fetched remote) and before the hard reset. `git rev-list --count origin/main..HEAD` is the right primitive: it counts commits reachable from HEAD but not from `origin/main`, i.e. exactly the local-only commits a reset would discard. The `out != "0"` comparison is reliable because `run_git` trims stdout. A rev-list failure also routes to `Skip`, consistent with the defensive-indeterminate philosophy of the surrounding precondition checks.

**Consistency**: The `Skip` outcome mirrors the existing on-`main` / clean-tree precondition failures — it returns `PrepOutcome::Skip` and leaves the halt flag untouched (never greenwashes, never halts on an indeterminate condition). Doc comment renumbered cleanly (old step 4 → step 5).

**Test quality**: `test_prepare_skips_when_local_main_ahead_of_origin` creates a real clean local `main` carrying a commit `origin/main` lacks, asserts the outcome is `Skip` with an 'ahead' reason, AND asserts `HEAD` is unchanged after the call — so it genuinely verifies the local-only commit survives rather than just checking the return variant. Mirrors the existing `test_prepare_skips_dirty_workspace` structure.

**Verification**: `cargo test main_health_gate` → 36 passed, 0 failed. `cargo fmt --check` clean. CI green (Rust Linting & Build + Rust Unit Tests both pass). Merge state CLEAN. No scope creep — a single focused edge-case guard plus its test.

LGTM! Approved.

Issue #3898, opened 2026-07-23
Autonomous guard hooks: enable decision logging + standing per-trigger review policy + reduce ASK confirmations


rjwalters:
## Context

Discovered while dogfooding autonomous daemon mode: a headless sweep runs under `--dangerously-skip-permissions`, where the guard `PreToolUse` hooks **fire** and an **ASK decision has no human to answer it, so it blocks** — functionally a deny. So every guard ASK silently stalls autonomous work. To run fully autonomously we need a standing feedback loop that converges the guard toward *dangerous-only*.

> Meta note: filing this very issue was itself blocked by the guard's catastrophic scan matching the literal recursive-force-remove-root string in the body text — a live example of the documentation-text false-positive class this issue targets.

Surface map (from `defaults/hooks/guard-destructive.sh`, 1712 lines):
- A JSONL **decision-logging substrate already exists** — `log_guard_decision()` (`:86-117`) appends `{ts, decision:"deny"|"ask", pattern:"<rule-tag>", tier, command:"<redacted>"}` to `.loom/logs/guard-decisions.log`, with a stable per-rule `pattern` tag at every deny/ask site. It is **gated OFF by default** (`guards.decisionLog` / `LOOM_GUARD_DECISION_LOG`, `:544-561`); no log exists today.
- `guard-loom-workflow.sh` has **no** decision logging — its two denies (`gh pr merge` redirect, worktree `pip install -e`) never appear in any log.
- **Highest-value false-positive:** `guards.forceScope` defaults to `"all"` (`:679-699`), so **every** `git push --force`/`--force-with-lease` and `git reset --hard` ASKs (`:1491`) — routine own-branch rebase/amend work that stalls headless. `"protected"` mode keeps force-push to `main`/`master` a hard DENY while allowing own-branch force-push.
- Other in-scope ASK false-positives: `git checkout .` / `git restore .` / `git clean -fd` (`:1549-1551`).
- Genuinely dangerous — MUST stay flagged: recursive-force-remove of root/`$HOME`/out-of-repo, force-push to main/master, `gh repo delete|archive`, fork bomb, pipe-curl-to-shell, cloud destruction, secrets exfil.

## Standing policy (the ask)

Every guard ASK/DENY that fires during autonomous operation gets a `loom:issue` filed to review it — to either allowlist the in-scope op (refine the guard) or confirm it stays flagged. Over time this converges the guard to dangerous-only without ever weakening a genuine safety rule.

## Acceptance criteria
[...]

PR #3918, opened 2026-07-23
fix(guards): pin protected-mode force-op test to synthetic branch (#3913)

### LINKED ISSUE #3913: guard test: protected-mode force-op test is checkout-branch-sensitive — pin to synthetic branch
## Context

Non-blocking note from the #3908 (guard-hook autonomy) review. The guard test asserting `hard-reset on own working branch allowed in protected mode` is **checkout-branch-sensitive**: it fails when run from a checkout that is currently *on* `main` (because a hard-reset on main should still be guarded), and passes when run from a feature-branch checkout (as CI does). The guard behaviour is correct; the test just implicitly assumes the runner is not on a protected branch.

## Impact
The test can spuriously fail depending on which branch the runner happens to be on — flaky/environment-sensitive rather than a real guard defect.

## Acceptance criteria
- [ ] Pin the test to a synthetic non-protected branch (create/checkout a throwaway branch, or stub the current-branch detection) so it is checkout-independent and passes regardless of the runner's branch.
- [ ] Keep the intended assertion intact (own-branch force op allowed under `forceScope:protected`; main/master still denied).

## Notes
Low priority test-robustness item flagged by the #3908 judge. Curator/builder may rescope or close if judged not worth it.

rjwalters:
## Summary

The `#3898` guard assertions for own-branch force ops under `guards.forceScope:"protected"` defaulted their cwd to `REPO_ROOT` (the real loom checkout). A bare `git reset --hard` resolves the **checked-out branch of the cwd**, so the assertion "hard-reset on own working branch allowed in protected mode" was **checkout-branch-sensitive**: it passed from a feature-branch checkout (as CI runs) but spuriously failed when the runner happened to be on `main`/`master` (where a hard-reset correctly stays protected → ASK). The guard behaviour is correct; only the test implicitly assumed the runner was not on a protected branch.

## Change

`tests/hooks/test-guard-destructive.sh` — pin the three own-branch `#3898` assertions to a **synthetic feature-branch fixture repo** (created via `make_sql_repo` + `git checkout -b feature/work`) instead of `REPO_ROOT`:

- `hard-reset on own working branch is allowed in protected mode` → pinned cwd
- `force-push to a non-protected branch is allowed in protected mode` → pinned cwd
- the companion "all mode still ASKS" case — previously a hand-rolled env-absent `run_guard` against `REPO_ROOT` — is converted to `assert_ask_env` with an explicit `LOOM_FORCE_SCOPE=all` and pinned to the same fixture, so it proves the intended property (all-mode asks even on a non-protected branch) independent of both the runner's checkout and any ambient `LOOM_FORCE_SCOPE` (e.g. the `protected` autonomous-daemon default set by `loom-daemon-start.sh`).

**Intent preserved**: own-branch force op allowed under `forceScope:protected`; `main`/`master` force-push still HARD-DENIES via `ALWAYS_BLOCK_PATTERNS` (those assertions are unchanged). No production hook code touched — this is a test-only fix.

## Test evidence

Command: `bash tests/hooks/test-guard-destructive.sh`

| Environment | Result |
|---|---|
| Clean CI-like (`LOOM_FORCE_SCOPE`/`LOOM_GUARD_DECISION_LOG` unset) | **407/407 PASS** (`ALL TESTS PASSED`) |
| The three targeted assertions under clean env | PASS |
| The three targeted assertions under ambient `LOOM_FORCE_SCOPE=protected` | PASS |
| The three targeted assertions under ambient `LOOM_FORCE_SCOPE=all` | PASS |

Before the fix, the `hard-reset on own working branch` assertion failed when run from a checkout on `main` (reproduced on the `main` worktree). After the fix it passes regardless of the runner's branch and regardless of any ambient `LOOM_FORCE_SCOPE`.

## Note (out of scope, potential follow-up)

While diagnosing, I found the broader force-scope and decision-log test sections are also sensitive to an ambient `LOOM_FORCE_SCOPE` / `LOOM_GUARD_DECISION_LOG` (env overrides the per-fixture config those assertions rely on): ~9–13 assertions fail when either var is exported in the process environment, as happens inside an autonomous sweep. CI runs in a clean environment so those pass there. Making the entire suite immune to ambient guard env vars is a larger change than this issue's checkout-branch scope; flagging it as a candidate follow-up rather than expanding scope here.

Closes #3913

### DIFF
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -1793,25 +1793,33 @@ assert_deny_env "#3898: force-push to main still HARD-DENIES in protected mode" 
 assert_deny_env "#3898: force-push to master still HARD-DENIES in protected mode" \
     "LOOM_FORCE_SCOPE=protected" "git push -f origin master"
 
-# Own working-branch force ops pass through (no stall) in protected mode. The
-# checked-out branch of REPO_ROOT is a feature branch, not a protected branch.
+# Own working-branch force ops pass through (no stall) in protected mode. Pin to a
+# SYNTHETIC feature-branch fixture repo rather than REPO_ROOT (#3913): a bare
+# `git reset --hard` resolves the *checked-out* branch of the cwd, so using
+# REPO_ROOT made this assertion checkout-branch-sensitive — it spuriously failed
+# when the test was run from a checkout that happened to be on `main`/`master`
+# (where a hard-reset correctly stays protected → ASK). The fixture is always on
+# `feature/work`, making the assertion checkout-independent.
+FS_FEATURE_REPO=$(make_sql_repo '{"guards":{"forceScope":"protected"}}')
+git -C "$FS_FEATURE_REPO" checkout -q -b feature/work 2>/dev/null || \
+    git -C "$FS_FEATURE_REPO" checkout -q -b feature/work
 assert_allow_env "#3898: hard-reset on own working branch is allowed in protected mode" \
-    "LOOM_FORCE_SCOPE=protected" "git reset --hard HEAD~1"
+    "LOOM_FORCE_SCOPE=protected" "git reset --hard HEAD~1" "$FS_FEATURE_REPO"
 
 assert_allow_env "#3898: force-push to a non-protected branch is allowed in protected mode" \
-    "LOOM_FORCE_SCOPE=protected" "git push --force origin feature/some-work"
+    "LOOM_FORCE_SCOPE=protected" "git push --force origin feature/some-work" "$FS_FEATURE_REPO"
 
-# Default (all) mode still ASKS on own-branch force ops (byte-for-byte behaviour).
-TOTAL=$((TOTAL + 1))
-_fs_out=$(run_guard "git reset --hard HEAD~1" "$REPO_ROOT") || true
-if echo "$_fs_out" | jq -e '.hookSpecificOutput.permissionDecision == "ask"' >/dev/null 2>&1; then
-    PASS=$((PASS + 1))
-    echo -e "  ${GREEN}PASS${NC}: #3898: default (all) mode still ASKS on own-branch hard-reset"
-else
-    FAIL=$((FAIL + 1))
-    echo -e "  ${RED}FAIL${NC}: #3898: default (all) mode still ASKS on own-branch hard-reset"
-    echo -e "       Got: $_fs_out"
-fi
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review — APPROVED

Verified both acceptance criteria and the core regression.

**Checkout-independence (criterion 1) — confirmed.** The three `#3898` own-branch assertions now run against a synthetic fixture (`make_sql_repo '{"guards":{"forceScope":"protected"}}'` + `git checkout -b feature/work`) passed as the 4th positional `cwd` arg to `assert_allow_env`/`assert_ask_env`. A bare `git reset --hard` now resolves the fixture's `feature/work` branch, not `REPO_ROOT`'s checked-out branch. I ran the fixed suite with `REPO_ROOT` pointing at the on-`main` checkout — the exact environment that used to fail — and it passes 407/407.

**Semantics preserved (criterion 2) — confirmed.** The `main`/`master` force-push HARD-DENY assertions are untouched and still pass; own-branch force op is still asserted `allow` under `forceScope:protected`. The converted "all mode still ASKS" case is now *stronger*: the old form was a hand-rolled env-absent `run_guard` against `REPO_ROOT` (doubly sensitive — ambient `LOOM_FORCE_SCOPE` could override the intended 'all' resolution, and the outcome depended on the runner's branch). It is now `assert_ask_env` with an explicit `LOOM_FORCE_SCOPE=all` against the feature-branch fixture — proving the intended property independent of both ambient env and checkout.

**Test-only — confirmed.** `defaults/hooks/guard-destructive.sh` is byte-identical between `main` and the PR branch (`diff` clean). Only `tests/hooks/test-guard-destructive.sh` changed. No `defaults/` copy of this test exists (only the hook lives under `defaults/`), so there is nothing to keep in sync.

**Test evidence (my own runs from the main-checkout worktree):**
- OLD test file (on `main`, unfixed): `hard-reset on own working branch` FAILS (cwd `/Users/rwalters/GitHub/loom`, ASK) — reproduces the reported regression.
- FIXED test file with `REPO_ROOT` = on-`main` checkout, clean env (`LOOM_FORCE_SCOPE`/`LOOM_GUARD_DECISION_LOG` unset): **407/407 PASS, ALL TESTS PASSED**.

Fixture reuse is clean, comments clearly explain the checkout/env sensitivity, and no assertions were weakened. The out-of-scope follow-up note (broader ambient-env sensitivity of the force-scope/decision-log sections) is appropriately flagged rather than scope-crept.

LGTM! Approved.

PR #3958, opened 2026-07-26
fix(mcp): bound sendDaemonRequest so publish_event fails fast instead of hanging ~1800s (#3945)

### LINKED ISSUE #3945: mcp__loom__publish_event from a daemon-dispatched child hangs ~1800s until MCP idle timeout (should be a fast fire-and-forget ack)
Reported by the #3936 repair sweep's own session log (daemon-dispatched child, v0.15.0 canary host):

> **Event-bus publish hung.** The `mcp__loom__publish_event` call for the phase event sat silent for 1800s before the MCP idle timeout aborted it — the daemon's MCP bridge isn't answering, even though the daemon dispatched this very sweep. I treated it as the skill's documented soft-failure (fire-and-forget) and skipped further publishes.

Expected: `PublishEvent` is a single line-delimited JSON request over the Unix socket answered with an immediate `EventPublished {receivers}` ack (0 receivers is fine). A 30-minute silent hang means the request never got a response frame — possibly the mcp-loom bridge (stale dist? #3803-shaped drift), a blocked IPC accept loop under the multi-repo canary load, or a request variant mismatch between the child's installed mcp-loom and the rebuilt daemon.

## Acceptance criteria
- [ ] Reproduce or instrument: publish_event from a headless child answers (ack or clean error) in <2s in the multi-repo configuration.
- [ ] If a bridge/daemon protocol version skew is the cause, the bridge fails fast with a clear error instead of hanging until idle timeout.
- [ ] The sweep skill's soft-failure path is preserved (a dead bus never blocks a sweep) — verify the child-side timeout is bounded (seconds, not 1800s).

Found during v0.15.0 canary rollout (multi-repo auto mode live test); see the #3936 sweep log on the canary host for the primary evidence.

rjwalters:
Closes #3945

## Root cause
`mcp-loom/src/shared/daemon.ts` → `sendDaemonRequest` is the unary request/response transport behind `publishEvent`, `dispatchSweep`, `listSweeps`, `getSweepStatus`, `tailSweepLog`, and `cancelSweep`. It created a socket, wrote the request, and resolved **only** on the socket `end` event — with **no timeout**. When the daemon accepted the connection but never wrote a response frame and never closed the socket (bridge/protocol skew, a blocked accept loop under multi-repo canary load, or a request-variant mismatch), the Promise neither resolved nor rejected. The socket sat open until the outer ~1800s MCP idle timeout killed the whole tool call — the reported `publish_event` hang. The sibling `sendDaemonStreamRequest` already had a duration timer; the unary path had none.

## Fix
- Added `DEFAULT_DAEMON_IPC_TIMEOUT_MS = 2000` and `resolveDaemonIpcTimeoutMs()`, which reads a positive-integer `LOOM_DAEMON_IPC_TIMEOUT_MS` env override and falls back to the default. `publish_event` now answers (ack or clean error) in **<2s** (AC1).
- `sendDaemonRequest(request, timeoutMs?)` arms a timer up front that bounds connect + write + response, clears it in the `end`/`error` handlers, and on expiry destroys the socket and rejects with a clear, greppable message naming the socket path, request type, and elapsed timeout: `Loom daemon did not respond within <N>ms at <SOCKET_PATH> (request type <T>) — daemon may be unreachable or its MCP/IPC bridge is not answering` (AC2 — fail fast on skew instead of hanging).
- A `settled` boolean guards against double-settle from a late `end`/`error` after a timeout reject.
- **No regression to the slow-but-legitimate callers:** `dispatch_sweep` spawns the child under the registry mutex before acking, and `cancel_sweep` blocks for the `SIGTERM → grace → SIGKILL` window (verified in `loom-daemon/src/ipc.rs`). Both pass explicit **wider** per-call timeouts via `Math.max(...)` (`dispatch` 30s; `cancel` = grace + 30s headroom) rather than raising the global default — still bounded, never the 1800s hang.

## Soft-failure path preserved (AC3)
The callers in `sweeps.ts` (e.g. `publishEvent`) already wrap the transport in try/catch and return `{ success: false, error }`. A bounded timeout reject therefore surfaces as a clean tool error, not a hang — the sweep skill's fire-and-forget semantics are unchanged, and a dead bus never blocks a sweep. Only the infinite hang was replaced with a bounded reject so the child's soft-failure handling can actually run.

## Verification
mcp-loom has no unit-test runner, so I added a self-contained `mcp-loom/scripts/verify-daemon-timeout.mjs` (`npm run verify:daemon-timeout`) that bundles the **real** transport with the already-present esbuild devDependency (no new deps) and asserts against stub Unix-socket servers:
1. **Hang case** — server accepts but never responds → call rejects within the bounded window with the greppable timeout error.
2. **Happy path** — server writes a JSON frame then ends → call resolves fast with the timer cleared (no regression to `dispatch_sweep`/`list_sweeps`).
3. Env-override parsing (`LOOM_DAEMON_IPC_TIMEOUT_MS`, invalid falls back to default).

Commands run:
- `npm run typecheck` (`tsc --noEmit`) → exit 0
- `npm run build` → exit 0 (rebuilt `dist/index.js`)
- `node scripts/verify-daemon-timeout.mjs` → all 9 checks passed

**Note on `dist/`:** `mcp-loom/dist/` is gitignored in this repo, so the rebuilt esbuild bundle is not committed — operators must run `npm run build` in `mcp-loom/` to ship the fix in the installed bridge.

### DIFF
--- a/mcp-loom/package.json
+++ b/mcp-loom/package.json
@@ -10,6 +10,7 @@
     "build": "tsc --noEmit && rm -rf dist && node esbuild.config.js",
     "build:bundle": "rm -rf dist && node esbuild.config.js",
     "typecheck": "tsc --noEmit",
+    "verify:daemon-timeout": "node scripts/verify-daemon-timeout.mjs",
     "watch": "tsc --watch --noEmit"
   },
   "dependencies": {
@@ -20,5 +21,8 @@
     "@types/node": "^22.18.10",
     "esbuild": "^0.27.3",
     "typescript": "^5.9.3"
+  },
+  "allowScripts": {
+    "esbuild@0.27.3": true
   }
 }
--- a/mcp-loom/scripts/verify-daemon-timeout.mjs
+++ b/mcp-loom/scripts/verify-daemon-timeout.mjs
@@ -1,0 +1,166 @@
+#!/usr/bin/env node
+/**
+ * Standalone verification for the bounded unary-request timeout (issue #3945).
+ *
+ * mcp-loom has no unit-test runner, so this self-contained script exercises
+ * the REAL `sendDaemonRequest` transport against stub Unix-socket servers:
+ *
+ *   1. Hang case  — server accepts the connection but never responds and never
+ *      closes. Assert the call REJECTS within the bounded timeout (well under
+ *      the 1800s MCP idle timeout that motivated the fix), with the greppable
+ *      "did not respond within" error.
+ *   2. Happy case — server writes a JSON response frame then ends. Assert the
+ *      call RESOLVES quickly and the timer was cleared (no regression to the
+ *      fast path used by dispatch_sweep / list_sweeps / publish_event).
+ *
+ * Run: node scripts/verify-daemon-timeout.mjs   (from the mcp-loom dir)
+ *
+ * No test framework or new runtime deps: it bundles `src/shared/daemon.ts`
+ * with the already-present esbuild devDependency and imports the result.
+ */
+
+import { build } from "esbuild";
+import net from "node:net";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { fileURLToPath, pathToFileURL } from "node:url";
+import { dirname } from "node:path";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const REPO = join(__dirname, "..");
+
+async function bundleTransport(outDir) {
+  const outfile = join(outDir, "daemon.mjs");
+  await build({
+    entryPoints: [join(REPO, "src/shared/daemon.ts")],
+    bundle: true,
+    format: "esm",
+    platform: "node",
+    outfile,
+    logLevel: "silent",
+  });
+  return import(pathToFileURL(outfile).href);
+}
+
+/** Start a stub server on a temp unix socket. `onConn(socket)` handles it. */
+function startStubServer(sockPath, onConn) {
+  return new Promise((resolve, reject) => {
+    const server = net.createServer(onConn);
+    server.on("error", reject);
+    server.listen(sockPath, () => resolve(server));
+  });
+}
+
+function closeServer(server) {
+  return new Promise((resolve) => server.close(() => resolve()));
+}
+
+let failures = 0;
+function check(name, ok, detail) {
+  if (ok) {
+    console.log(`  ok   ${name}`);
+  } else {
+    failures += 1;
+    console.error(`  FAIL ${name}${detail ? ` — ${detail}` : ""}`);
+  }
+}
+
+async function main() {
+  const workDir = await mkdtemp(join(tmpdir(), "loom-3945-"));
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review: APPROVED ✅

Reviewed against the three ACs of #3945. The fix is correct, focused, and preserves the sweep skill's soft-failure contract.

### Correctness of the timeout logic
- Timer is armed **before** `socket.connect()`, so it bounds connect + write + response-wait — exactly the hang window (accept-but-never-respond).
- Cleared on **both** terminal paths (`end` and `error`) via `clearTimer()` inside `settleResolve`/`settleReject`.
- Double-settle is guarded by the `settled` boolean, so a late `end`/`error` after a timeout reject (or vice-versa) is a no-op.
- Fast path is unaffected: on `end` the timer is cleared and the socket auto-closes (default `allowHalfOpen: false`), so no leaked handle keeps the process alive — the verify script's happy-path check (`resolved fast`, clean process exit) confirms this.

### No regression to slow callers (AC2)
- `dispatch_sweep` → `Math.max(30_000, resolveDaemonIpcTimeoutMs())` — comfortably above a normal spawn-under-mutex ack; the 2s default never applies.
- `cancel_sweep` → `Math.max(grace_secs*1000 + 30_000, resolveDaemonIpcTimeoutMs())`. Math is correct (grace window in ms + 30s headroom). `grace_secs` is a required `number`, defaulted to 30 in the tool handler with a finite/`>= 0` guard, so there's no NaN path into the transport.

### Soft-failure preserved (AC3)
- `publishEvent` (and the other callers) wrap the transport in try/catch and return `{ success: false, error }`. The bounded reject surfaces as a clean tool error, not a thrown crash — the child's fire-and-forget path runs. Confirmed a dead bus never blocks a sweep; only the infinite hang was replaced with a bounded reject.

### Env override parsing
- `resolveDaemonIpcTimeoutMs()` uses `Number.parseInt` + `Number.isInteger(parsed) && parsed > 0`, falling back to the 2000ms default for absent/blank/zero/negative/non-numeric — no NaN/negative/zero footgun.

### Build integrity (ran locally on the PR branch)
- `npm run typecheck` → exit 0
- `npm run build` → exit 0
- `node scripts/verify-daemon-timeout.mjs` → **all 9 checks passed** (env-override x3, hang-case reject-within-window + greppable message, happy-path resolve-fast)
- CI `MCP Server Build` passing; `mergeStateStatus: CLEAN`.
- `mcp-loom/dist/` is gitignored (root `.gitignore` `dist/`), consistent with repo convention; PR body documents that operators must `npm run build` to ship the fix in the installed bridge. Good.

### Scope
Focused: transport timeout + two per-call widenings + verify script. One minor, non-blocking observation: the new `allowScripts: { "esbuild@0.27.3": true }` in `mcp-loom/package.json` is a `@lavamoat/allow-scripts` config that appears nowhere else in the repo (root has no lavamoat), so it's a harmless no-op rather than a functional change — fine to leave, just noting it isn't load-bearing here.

LGTM. Approving.

PR #3963, opened 2026-07-26
fix(daemon): self-init activity DB schema in GitHub metrics collector

### LINKED ISSUE #3951: GitHub metrics collector: 'Failed to collect PR/issue events: no such table: agent_activity' every interval
The daemon's 15-min GitHub metrics collector errors every cycle on the canary host:

```
[ERROR] Failed to collect PR events: no such table: agent_activity
[ERROR] Failed to collect issue events: no such table: agent_activity
```

The activity DB schema evidently isn't initialized (or the collector points at a DB file whose migrations never ran on this host). Either run migrations on first open, or degrade to a single startup warning instead of recurring ERRORs.

## Acceptance criteria
- [ ] Collector self-initializes its schema (or cleanly disables with one warning naming the fix).
- [ ] No recurring ERROR lines on a fresh install/daemon start.

Found during v0.15.0 canary rollout.

rjwalters:
## Summary

The daemon's 15-min GitHub metrics collector (`loom-daemon/src/metrics_collector.rs`) errored every cycle with:

```
[ERROR] Failed to collect PR events: no such table: agent_activity
[ERROR] Failed to collect issue events: no such table: agent_activity
```

Root cause was two-fold:

1. `collect_and_store_events` opened the activity DB with a bare `Connection::open` and never ran schema initialization. `activity::ActivityDb::new()` (the other consumer of this DB) already calls the existing idempotent `activity::init_schema()` on every open, but the metrics collector bypassed that path entirely — so a fresh install (or a DB file whose migrations never ran) hit the missing-table error on every collection interval.
2. `correlate_with_activity` queried a table named `agent_activity`, which was never part of the actual schema (`loom-daemon/src/activity/schema.rs`) — the `github_events.activity_id` foreign key actually points at `agent_metrics`. Even after fixing (1), this query would still have failed against a correctly-initialized DB.

## Changes

- `loom-daemon/src/activity/mod.rs`: re-export `schema::init_schema` so callers outside `activity::db` (like the metrics collector) that need a raw `rusqlite::Connection` can self-initialize the schema.
- `loom-daemon/src/metrics_collector.rs`:
  - Call `crate::activity::init_schema(&conn)` immediately after opening the DB connection in `collect_and_store_events`, mirroring what `ActivityDb::new()` already does. `init_schema` is `CREATE TABLE/INDEX IF NOT EXISTS`, so this is a cheap no-op once the schema exists — no recurring ERROR on a fresh install.
  - Rewrite `correlate_with_activity`'s two queries to target `agent_metrics` with its actual column names (`github_issue`, `COALESCE(completed_at, started_at)`, `status = 'success'`) instead of the nonexistent `agent_activity`/`issue_number`/`timestamp`/`outcome`.
  - Rebuild the test helper `create_test_db()` on the real `init_schema()` (instead of a hand-rolled mini schema) so unit tests track the production schema and would have caught this drift.
  - Add regression tests: inserting/correlating a GitHub event against a freshly-`init_schema()`'d DB and against a brand-new DB file path (the exact fresh-install scenario from the bug report) no longer errors.

## Test plan

- [x] `cargo test --lib metrics_collector` — 15/15 pass, including the new regression tests
- [x] `cargo test --lib` (full `loom-daemon` suite) — 784/784 pass
- [x] `cargo clippy --lib --all-targets` — clean
- [x] `cargo fmt -- --check` — clean

Closes #3951

### DIFF
--- a/loom-daemon/src/activity/mod.rs
+++ b/loom-daemon/src/activity/mod.rs
@@ -70,6 +70,11 @@ pub use models::{
 // Re-export the database struct
 pub use db::ActivityDb;
 
+// Re-export schema initialization so callers that need a raw `rusqlite::Connection`
+// (e.g. the GitHub metrics collector) can self-initialize the activity DB schema
+// without routing through `ActivityDb`.
+pub use schema::init_schema;
+
 // Re-export resource usage parsing and cost calculation
 // Used internally by db.rs for terminal output parsing
 // Note: db.rs accesses these via super::resource_usage, so these re-exports
--- a/loom-daemon/src/metrics_collector.rs
+++ b/loom-daemon/src/metrics_collector.rs
@@ -240,6 +240,13 @@ fn collect_and_store_events(config: &MetricsConfig) -> Result<usize> {
 
     let conn = Connection::open(&config.db_path).context("Failed to open activity database")?;
 
+    // Self-initialize the activity DB schema on every open. `init_schema` is
+    // idempotent (CREATE TABLE/INDEX IF NOT EXISTS), so this is a cheap no-op
+    // once the schema exists. Without this, a fresh install (or a collector
+    // pointed at a DB file whose migrations never ran) hits recurring
+    // "no such table" errors every collection interval (#3951).
+    crate::activity::init_schema(&conn).context("Failed to initialize activity database schema")?;
+
     let mut total_events = 0;
 
     // Collect PRs
@@ -445,6 +452,13 @@ fn insert_github_event(
 }
 
 /// Attempt to correlate a GitHub event with agent activity
+///
+/// Correlates against `agent_metrics` (aliased here as "activity" for the
+/// GitHub-events domain) — the table `github_events.activity_id` actually has
+/// a foreign key to. An earlier version of this query targeted a table named
+/// `agent_activity`, which was never part of the schema (see #3951); the
+/// nearest-timestamp match uses `COALESCE(completed_at, started_at)` since
+/// `agent_metrics` has no single `timestamp` column.
 fn correlate_with_activity(
     conn: &Connection,
     _pr_number: Option<i64>,
@@ -455,9 +469,9 @@ fn correlate_with_activity(
     if let Some(issue_num) = issue_number {
         let activity_id = conn
             .query_row(
-                "SELECT id FROM agent_activity
-                 WHERE issue_number = ?1
-                 ORDER BY ABS(julianday(timestamp) - julianday(?2))
+                "SELECT id FROM agent_metrics
+                 WHERE github_issue = ?1
+                 ORDER BY ABS(julianday(COALESCE(completed_at, started_at)) - julianday(?2))
                  LIMIT 1",
                 params![issue_num, event_time],
                 |row| row.get(0),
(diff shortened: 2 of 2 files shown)


rjwalters:
Reviewed and verified independently (not just the builder's summary).

**Root cause + fix confirmed correct:**
- `metrics_collector::collect_and_store_events` previously opened the activity DB with a bare `Connection::open` and never initialized the schema — confirmed by reading the pre-diff code. The new `crate::activity::init_schema(&conn)` call after open matches what `ActivityDb::new()` already does elsewhere.
- `activity/schema.rs::init_schema` is genuinely idempotent — every statement is `CREATE TABLE/INDEX IF NOT EXISTS`, safe to call on every 15-min collection cycle, not just once.
- Confirmed `agent_activity` never existed in `activity/schema.rs`; the real FK target of `github_events.activity_id` is `agent_metrics` (verified via `FOREIGN KEY (activity_id) REFERENCES agent_metrics(id)`). The rewritten `correlate_with_activity` columns (`github_issue`, `COALESCE(completed_at, started_at)`, `status = 'success'`) all match the actual `agent_metrics` schema, and `'success'` is the same status literal used elsewhere (`activity/metrics.rs:189`).
- Confirmed the `db_path` the daemon wires into `MetricsConfig` (`main.rs:247`/`312`) is the same production `<workspace>/.loom/activity.db` that `ActivityDb` uses — the fix targets the actual code path that broke on the canary host.
- New regression tests genuinely exercise the fresh/uninitialized-DB scenario from the bug report: `test_collect_and_store_events_self_initializes_schema_on_fresh_db_path` opens a brand-new file path (asserts `!db_path.exists()` beforehand) and drives the exact `init_schema` call the collector makes; `test_insert_github_event_on_freshly_initialized_db_does_not_error` and `test_correlate_with_activity_on_freshly_initialized_db_returns_none` cover the no-prior-rows case. `create_test_db()` now builds on the real `init_schema()` instead of a hand-rolled mini-schema, so future schema drift would be caught here too.

**Verified myself (not trusting the builder's report):**
- `cargo test --lib metrics_collector` — 15/15 pass
- `cargo test --lib` (full `loom-daemon`) — 784/784 pass
- `cargo clippy --package loom-daemon --package loom-api -- -D warnings <same allow-flags as CI>` — clean (matched exactly against `.github/workflows/ci.yml`'s clippy invocation)
- `gh pr checks 3963` — all required checks pass; `mergeStateStatus` is `CLEAN`

**One non-blocking finding for a follow-up issue:** `loom-api/src/main.rs` (the separate analytics REST API crate in this workspace) still queries `.loom/activity.db` — the same physical DB file — for a table literally named `agent_activity` (`get_metrics_summary`/`get_metrics_by_role`, lines ~231/290, plus the `CREATE TABLE agent_activity` in its own test fixture at ~710). That table never existed in `activity/schema.rs` either, so `loom-api` would hit the same "no such table: agent_activity" error against a real production DB the moment it's queried — its unit tests pass only because they stand up a hand-rolled fixture schema rather than the real one. This is the same class of bug as #3951 but in a different crate and out of scope for this PR's acceptance criteria (which are specifically about "the collector"). Worth filing a follow-up issue so it doesn't surprise someone later.

Approving — the fix is correct, idempotent, and well-tested for the scope of #3951.

Issue #3929, opened 2026-07-24
(repo,issue)-keyed sweep registry + dispatch: namespace worktrees/locks/logs per repo (#3926 phase c)


rjwalters:
## Original Issue

Phase (c) of #3926. Today the sweep registry, claim locks, worktrees, and per-sweep logs are keyed by `issue` alone. Two managed repos can each have an issue #42, so these must be keyed by `(repo, issue)`.

## Acceptance criteria
- [ ] Sweep registry entries, claim locks (`<repo>/.loom/locks/`), worktrees, and per-sweep logs are namespaced per repo — keyed on `(repo, issue)` not `issue`.
- [ ] Each sweep dispatches `/loom:sweep <issue>` in the correct repo's working tree / worktree root.
- [ ] `list_sweeps` / status distinguish sweeps from different repos with the same issue number.

## Notes
Depends on phases (a) registry and (b) multi-repo work-finder. Part of #3835.

---

### Important: most of the filesystem namespacing is already done by phase (b) — #3928 / PR #3932

Phase (b) merged `loom-daemon/src/workspace_pool.rs` (`WorkspacePool`), which provisions **one independent `SweepRegistry` per registered repo root** on demand (`get_or_provision`, keyed by `PathBuf` root) and seeds the default workspace's pre-existing registry as its first entry (`seed`). Every path a `SweepRegistry` computes is derived from `SweepRegistryConfig::workspace_root` (`loom-daemon/src/sweep_registry.rs:513-592`):

- `locks_dir()` → `<workspace_root>/.loom/locks/` (claim lock file `issue-<N>`)
- `logs_dir()` → `<workspace_root>/.loom/logs/`
- `checkpoint_dir()` → `<workspace_root>/.loom/sweep-checkpoint/`
- dispatch `current_dir` for the spawned `/loom:sweep <issue>` child (`spawn_child`, `sweep_registry.rs:~1301`)

[...]

Issue #3729, opened 2026-07-22
Stacked-PR mode for loom-daemon: pipeline dependent issues (child builds on parent's branch)


rjwalters:
## Curator Enhancement — v1 scope for this PR (2026-07-22)

This issue is a full four-phase architect proposal (detection → build → squash reconciliation → cascade policy). Per explicit operator direction, this is being built **aggressively in a single PR** rather than decomposed into sub-issues. Curator's job here is to pin down a **realistic, landable v1 slice** — not to spec the whole proposal. The full proposal is preserved below under "Original Proposal" for context on where this is headed; **only the section below is authoritative scope for this issue/PR.**

### Honesty check on sizing

Even trimmed to the minimum below, this touches Rust daemon code (`loom-daemon/src/types.rs`, `loom-daemon/src/sweep_registry.rs`), the MCP dispatch surface (`mcp-loom/src/tools/sweeps.ts`), the sweep skill doc (`defaults/.claude/commands/loom/sweep.md`), and a worktree/branch script (`defaults/scripts/worktree.sh`), plus tests for each. That is a large single PR by the repo's own decomposition heuristic (`.loom/roles/curator.md`: >6h / >8 files / >400 LOC triggers sub-issue splitting). The operator has explicitly chosen to accept that risk and land it in one pass — this enhancement narrows scope as far as possible while keeping the result *coherent and mergeable*, primarily by deferring the highest-risk piece (automated squash reconciliation + merge-ordering gate in `merge-pr.sh`) to a **fast-follow issue**, in favor of a manual-but-scripted equivalent in v1. If the Builder finds even this slice exceeds what fits in one PR, it should stop and flag the smaller sub-slice in "v1a" below rather than force everything in.

### v1 scope (land in this PR)

[...]

PR #3965, opened 2026-07-26
feat(daemon): loom-daemon dispatch <issue> operator CLI (#3952)

### LINKED ISSUE #3952: Operator dispatch CLI: loom-daemon dispatch <issue> [--workspace <root>] — resilient non-MCP dispatch surface
During the canary rollout the MCP `dispatch_sweep` path hung for 1800s (#3945) and the only operator alternative was hand-rolling the daemon's dispatch pattern: flip the label, export `LOOM_SWEEP_CLAIM_OWNED=<N>` + `LOOM_MODEL` + workaround envs, and invoke `spawn-claude.sh -p "/loom:sweep N"` directly. That pattern worked (2 sweeps dispatched this way completed) but it's undocumented, easy to get wrong (claim-marker mismatch → the child skips its own issue), and bypasses the registry (no in-flight tracking, no reaper, no status visibility).

## Fix
Add a first-class CLI: `loom-daemon dispatch <issue> [--workspace <root>] [--model M] [--effort E] [--depends-on P]` — a thin client over the existing IPC `DispatchSweep` (same code path as MCP but with a **bounded client-side timeout and a clear error** if the daemon doesn't ack in seconds). Registry tracking, claim flip, and event publishing come for free.

## Acceptance criteria
- [ ] `loom-daemon dispatch` enqueues via IPC, prints sweep id + log path, exits nonzero with a clear message on a >5s ack timeout (never hangs).
- [ ] `--workspace` targets any registered root (phase-c plumbing).
- [ ] Docs: daemon-reference operator section replaces the hand-rolled pattern.

Depends on / complements #3945 (the wedge must be fixed for any dispatch surface to be reliable). Found during v0.15.0 canary rollout.

rjwalters:
## Summary

Adds a first-class, non-MCP operator dispatch surface: `loom-daemon dispatch <issue> [--workspace <root>] [--model M] [--effort E] [--depends-on P]`. It is a thin client over the existing IPC `DispatchSweep` request — the same code path the `mcp__loom__dispatch_sweep` tool uses — so registry tracking, the `loom:issue → loom:building` claim flip, in-flight tracking, the reaper, and event publishing all come for free.

Motivated by the v0.15.0 canary rollout (#3945): the MCP `dispatch_sweep` path once wedged for 1800s, and the only operator fallback was hand-rolling the daemon's dispatch (flip the label, export `LOOM_SWEEP_CLAIM_OWNED=<N>` + `LOOM_MODEL`, invoke `spawn-claude.sh -p "/loom:sweep N"`) — undocumented, easy to get wrong (claim-marker mismatch makes the child skip its own issue), and bypasses the registry entirely.

## What changed

- **`Dispatch` subcommand** in `loom-daemon/src/main.rs` with async routing in `main()` (like `status` / `quarantine`).
- **Bounded ack timeout**: waits at most 5s for the daemon to ack; on expiry exits nonzero with a clear `Daemon did not ack the dispatch within 5s (...) — is loom-daemon running?` message. Never hangs.
- **Flag plumbing**: `--workspace` → `workspace_root` (#3929), `--model` → `model`, `--effort` → `effort`, `--depends-on` → `depends_on` (#3729), all forwarded into `DispatchSweep`, mirroring the MCP tool.
- On success prints the returned `sweep_id`, pid, token account, and the per-sweep log path (`.loom/logs/sweep-issue-<N>.log`), then exits 0.
- Extracted `query_daemon_bounded` (caller-supplied timeout) so the timeout path is unit-testable without a multi-second wait; `query_daemon` now delegates to it with the 5s default.
- Pure `build_dispatch_request` for flag-plumbing tests.
- Documented the CLI in `.loom/docs/daemon-reference.md` as the replacement for the hand-rolled `LOOM_SWEEP_CLAIM_OWNED` + `spawn-claude.sh` pattern (with the deprecated pattern shown as a "do NOT do this" example).

## Tests

5 new unit tests in `dispatch_tests` (all passing):
- `build_dispatch_request_plumbs_all_flags` — every flag maps to the right `DispatchSweep` field
- `build_dispatch_request_defaults_are_none` — omitted flags leave overrides `None`
- `round_trip_parses_dispatched_response_and_forwards_flags` — full round-trip against a fake daemon: request serialization + flags-on-the-wire + `SweepDispatched` parsing
- `unresponsive_socket_times_out` — a socket that accepts but never replies trips the bounded timeout in <2s (the #3945 wedge must never hang)
- `absent_socket_errors_fast` — missing socket surfaces a connect error promptly

`cargo clippy --bin loom-daemon` is clean. Full crate suite passes except `test_send_input`, a pre-existing tmux integration flake unrelated to this change (fails only under parallel load; passes in isolation).

## Acceptance criteria

- [x] `loom-daemon dispatch` enqueues via IPC, prints sweep id + log path, exits nonzero with a clear message on a >5s ack timeout (never hangs).
- [x] `--workspace` targets any registered root (phase-c plumbing).
- [x] Docs: daemon-reference operator section replaces the hand-rolled pattern.

Closes #3952

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -213,12 +213,21 @@ loom-daemon dispatch 3952 --depends-on 3945        # stacked-PR child (#3729)
 | `--effort <E>` | `effort` | reasoning-effort override (#3716) |
 | `--depends-on <P>` | `depends_on` | single parent issue; child branches off `feature/issue-<P>` (#3729) |
 
-**Bounded ack timeout (never hangs).** The CLI waits at most **5s** for the
-daemon to ack the dispatch. On expiry it exits **nonzero** with a clear
-`Daemon did not ack the dispatch within 5s (...) — is loom-daemon running?`
-message rather than blocking. This is the whole point of the command: the MCP
-`dispatch_sweep` path once wedged for **1800s** (#3945), and hand-rolling the
-dispatch bypassed the registry entirely.
+**Bounded ack timeout (never hangs).** The CLI waits at most **30s** for the
+daemon to ack the dispatch, then exits **nonzero** with a clear
+`Daemon did not ack the dispatch within 30s (...) — is loom-daemon running?`
+message rather than blocking. The 30s default mirrors the `mcp__loom__dispatch_sweep`
+tool's own `DISPATCH_TIMEOUT_MS` for the identical IPC call, and exists because
+`SweepRegistry::dispatch()` does real synchronous work *before* it acks — a
+blocking `gh issue edit` label flip, up to a 2s dispatch stagger, and up to a 5s
+token-name capture window — so a legitimate, successful dispatch can take several
+seconds to ack. A tighter bound (the original 5s) would false-report those real
+successes as `did not ack`. Operators on a slow forge or a heavily-loaded daemon
+can *raise* the bound with `LOOM_DAEMON_IPC_TIMEOUT_MS=<ms>` (the same env var
+`mcp-loom` honors); it only ever raises above the 30s floor, never lowers it (a
+lower value would reintroduce the false negative). The timeout is always a
+bounded, finite value: the MCP `dispatch_sweep` path once wedged for **1800s**
+(#3945), and this command must never reproduce that hang.
 
 **Replaces the hand-rolled pattern.** Before #3952 the only non-MCP alternative
 was to reproduce the daemon's dispatch by hand — flip the label, export
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -189,6 +189,54 @@ Inputs:
   its working tree — the way to dispatch into a managed repo other than the
   default when two repos share issue numbers. `#[serde(default)]` on the wire.
 
+#### `loom-daemon dispatch <issue>` — operator CLI (Issue #3952)
+
+`loom-daemon dispatch <issue>` is the **non-MCP** operator entry point onto the
+same `DispatchSweep` IPC request the `dispatch_sweep` MCP tool uses. It is a thin
+client: connect to the daemon socket, send one `DispatchSweep` frame, print the
+returned `sweep_id` + per-sweep log path, exit `0`. Because it flows through the
+registry, the `loom:issue → loom:building` claim flip, in-flight tracking, the
+reaper, and event publishing all come for free — exactly like the MCP path.
+
+```bash
+loom-daemon dispatch 3952                          # dispatch into the default workspace
+loom-daemon dispatch 3952 --workspace /path/to/repo # target a registered managed repo (#3929)
+loom-daemon dispatch 3952 --model sonnet --effort high
+loom-daemon dispatch 3952 --depends-on 3945        # stacked-PR child (#3729)
+```
+
+| Flag | Maps to `DispatchSweep` field | Notes |
+|------|-------------------------------|-------|
+| `<issue>` (positional) | `kind = {"Issue": N}` | required |
+| `--workspace <PATH>` | `workspace_root` | target a registered repo other than the default (#3929) |
+| `--model <M>` | `model` | omit to let the daemon resolve `autonomous.model` / the shipped default (#3944) |
+| `--effort <E>` | `effort` | reasoning-effort override (#3716) |
+| `--depends-on <P>` | `depends_on` | single parent issue; child branches off `feature/issue-<P>` (#3729) |
+
+**Bounded ack timeout (never hangs).** The CLI waits at most **5s** for the
+daemon to ack the dispatch. On expiry it exits **nonzero** with a clear
+`Daemon did not ack the dispatch within 5s (...) — is loom-daemon running?`
+message rather than blocking. This is the whole point of the command: the MCP
+`dispatch_sweep` path once wedged for **1800s** (#3945), and hand-rolling the
+dispatch bypassed the registry entirely.
+
+**Replaces the hand-rolled pattern.** Before #3952 the only non-MCP alternative
+was to reproduce the daemon's dispatch by hand — flip the label, export
+`LOOM_SWEEP_CLAIM_OWNED=<N>` plus `LOOM_MODEL` and workaround envs, and invoke
+`spawn-claude.sh -p "/loom:sweep N"` directly:
+
+```bash
+# DEPRECATED — do NOT do this. Bypasses the registry (no in-flight tracking,
+# no reaper, no status visibility) and a claim-marker mismatch makes the child
+# skip its own issue.
+gh issue edit 3952 --remove-label loom:issue --add-label loom:building
+LOOM_SWEEP_CLAIM_OWNED=3952 LOOM_MODEL=sonnet \
+  ./.loom/scripts/spawn-claude.sh -p "/loom:sweep 3952"
+```
+
+Use `loom-daemon dispatch 3952` instead — it performs the claim flip, registry
+tracking, and event publishing for you, with the bounded timeout as a safety net.
+
 ### `list_sweeps` (Phase A)
 
 Return all tracked sweeps, optionally filtered by lifecycle state.
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -118,6 +118,42 @@ enum Commands {
         action: QuarantineAction,
     },
 
+    /// Dispatch a `/loom:sweep <issue>` via the running daemon (Issue #3952): a
+    /// first-class, non-MCP operator entry point over the same IPC `DispatchSweep`
+    /// request the `mcp__loom__dispatch_sweep` tool uses. Registry tracking, the
+    /// `loom:issue → loom:building` claim flip, and event publishing all come for
+    /// free — this is the resilient replacement for the hand-rolled
+    /// `LOOM_SWEEP_CLAIM_OWNED=<N>` + `spawn-claude.sh -p "/loom:sweep N"` pattern.
+    ///
+    /// Applies a bounded client-side ack timeout: if the daemon does not respond
+    /// within a few seconds the command exits nonzero with a clear error instead
+    /// of hanging (motivated by the 1800s MCP wedge in #3945).
+    Dispatch {
+        /// The issue number to dispatch a sweep for.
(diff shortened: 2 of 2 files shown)


rjwalters:
❌ **Changes Requested**

Thanks for this — the flag plumbing, docs, and unit tests are solid (verified `build_dispatch_request_plumbs_all_flags`, the fake-daemon round-trip, and the unresponsive-socket timeout test all pass; full crate suite is 781/781 in `--lib`, 5/5 in the new `dispatch_tests`, `cargo clippy --bin loom-daemon --all-targets` clean). CI is green and `mergeStateStatus` is `CLEAN`.

**But the 5s client-side ack timeout has essentially zero margin over the daemon's own known worst-case latency for this exact request, and will produce false "did not ack" failures under completely normal (non-wedged) operation:**

1. `DISPATCH_ACK_TIMEOUT` in `main.rs` is `Duration::from_secs(5)`.
2. `SweepRegistry::dispatch()` (`loom-daemon/src/sweep_registry.rs:1132`) does all of this **synchronously, before the ack is sent**:
   - `flip_label_to_building()` (`sweep_registry.rs:1363`) — a blocking `gh issue edit` **network** round-trip.
   - `apply_dispatch_stagger()` — up to `DEFAULT_DISPATCH_STAGGER_MS = 2000` (2s) when a recent dispatch is in flight (e.g. under the work-finder or back-to-back operator calls).
   - `spawn_child()` → `poll_token_name()` (`sweep_registry.rs:714`), which polls the per-sweep log for the child's OAuth-account-selection line for up to `TOKEN_NAME_CAPTURE_TIMEOUT = Duration::from_secs(5)` before falling back to `UNKNOWN_TOKEN_NAME`. The comment at `sweep_registry.rs:266-272` explicitly documents this as an anticipated, **non-fatal, graceful-degradation** path ("only ever waited out in the pathological 'child alive but never logged' case ... never blocks or fails dispatch") — i.e. this 5s window is a known, designed-for occurrence, not a wedge.
3. So the daemon's own internal budget for a real, successful `DispatchSweep` ack can legitimately be `~5s (token-name poll) + label-flip latency + up to 2s stagger` — comfortably exceeding this CLI's 5s round-trip timeout. When that happens, the dispatch **actually succeeds server-side** (label flipped, child spawned, registry entry recorded) but the CLI reports `Daemon did not ack the dispatch within 5s (...) — is loom-daemon running?` and exits nonzero — a false negative on the PR's own headline acceptance criterion ("prints sweep id + log path... on success").
4. This is exactly the risk `mcp-loom/src/tools/sweeps.ts` already reasoned through for the *identical* underlying IPC call: it deliberately uses `DISPATCH_TIMEOUT_MS = 30_000` (6x this PR's bound) with a comment calling out "the daemon spawns the sweep child (token selection + spawn-claude.sh) while holding the registry mutex before it acks," plus an `LOOM_DAEMON_IPC_TIMEOUT_MS` env escape hatch for operators. This new CLI has neither the wider bound nor an override.
5. Compounding it: `build_dispatch_request` always passes `idempotency_key: None`, so an operator who sees the false-negative and retries will **double-dispatch** the same issue rather than being deduped (this matches the MCP tool's existing behavior, so not a new regression, but it does mean the blast radius of a false-negative here is a duplicate sweep, not just a scary error message).

**Requested fix:**
- Bump `DISPATCH_ACK_TIMEOUT` to have real headroom over the documented internal budget — mirror the MCP tool's 30s bound (or otherwise justify a smaller number against the actual worst case above), and/or add an env override analogous to `LOOM_DAEMON_IPC_TIMEOUT_MS` so operators can tune it without a rebuild.
- Add a test that models the realistic "slow-but-legitimate" ack (e.g. a fake daemon that sleeps ~4-6s before replying `SweepDispatched`) to prove the CLI doesn't misreport a real success as a failure. The current three round-trip/timeout tests only cover an instant ack and a permanently-unresponsive stub — neither exercises the documented middle case that's actually expected to occur.

Minor, non-blocking observations if you want to fold them in while fixing the above:
- Consider surfacing `token_name == UNKNOWN_TOKEN_NAME` distinctly in the CLI's success output (a hint that something was slow/degraded on the daemon side) rather than printing it identically to a normal resolved account name.

I re-verified independently: checked out the PR branch, ran `cargo test` (full suite) and `cargo test --test integration_basic -- --test-threads=1` — confirms the 2 tmux-integration failures I saw under default parallel `cargo test` (`test_create_terminal`, `test_list_terminals`) are the same class of pre-existing flake the PR description called out (a different specific test name flaked in my run than `test_send_input`, but all pass in isolation with `--test-threads=1`, and this suite is untouched by this diff) — not a regression from this change.

rjwalters:
Addressed the ack-timeout false-negative. Summary of the fix (commit 5556da7d):

**Widened the bound + env override**
- `DISPATCH_ACK_TIMEOUT` is now `Duration::from_secs(30)` (was 5s), mirroring `mcp-loom`'s `DISPATCH_TIMEOUT_MS` for the identical `DispatchSweep` IPC call — real headroom over the daemon's documented worst case (`flip_label_to_building()` label flip + up to 2s `apply_dispatch_stagger()` + up to 5s `TOKEN_NAME_CAPTURE_TIMEOUT` poll).
- New `resolve_dispatch_ack_timeout()` honors `LOOM_DAEMON_IPC_TIMEOUT_MS` — the **same** env var `mcp-loom` reads, so operators have one client-side convention across both surfaces. It mirrors the MCP tool's `Math.max` semantics: the env value can only ever *raise* the bound above the 30s floor, never lower it (lowering would reintroduce exactly this false negative). Absent/empty/non-numeric/zero/negative falls back to the floor.
- Still bounded/finite — never the ~1800s #3945 wedge. The timeout error message now reports the resolved (possibly overridden) value.

**Tests**
- `slow_but_legitimate_ack_succeeds`: a fake daemon that sleeps **5.5s** (past the retired 5s bound) before replying `SweepDispatched`, queried with the CLI's real 30s default — asserts success and that elapsed ≥ 5s, so it genuinely exercises the slow-but-legitimate middle case (would have false-failed pre-fix). This is the case neither the instant-ack nor permanently-unresponsive stub covered.
- `resolve_dispatch_ack_timeout_defaults_to_floor` and `resolve_dispatch_ack_timeout_env_raises_only` (both `#[serial]` to avoid env races): cover the 30s default, env-raises-above-floor, env-below-floor-clamped-up, and invalid-value-falls-back cases.

**Folded in the non-blocking nit**
- CLI success output now surfaces `token_name == UNKNOWN_TOKEN_NAME` distinctly (`account not captured before ack — slow child startup`) rather than printing it like a resolved account.

**Verification** (this branch, worktree recreated via `worktree.sh`):
- `cargo clippy --bin loom-daemon --all-targets` — clean.
- `cargo test --bin loom-daemon` — 8/8 `dispatch_tests` pass (incl. the new slow-ack + resolver tests).
- `cargo test --lib` — 781/781 pass.

Docs: `daemon-reference.md` updated to the 30s bound + `LOOM_DAEMON_IPC_TIMEOUT_MS` override.

Re-requesting review.

rjwalters:
✅ **Approved!**

Re-reviewed after the fix for my prior "Changes Requested" (5s ack-timeout false-negative risk). Verified commit 5556da7d directly (not just trusting the summary):

**Diff matches the claim exactly** (`loom-daemon/src/main.rs`, `.loom/docs/daemon-reference.md`):
- `DISPATCH_ACK_TIMEOUT` widened `5s → 30s`, mirroring `mcp-loom`'s `DISPATCH_TIMEOUT_MS` for the identical `DispatchSweep` IPC call.
- New `resolve_dispatch_ack_timeout()` reads `LOOM_DAEMON_IPC_TIMEOUT_MS` (same env var name as mcp-loom) with `Math.max`/floor semantics — confirmed in the test `resolve_dispatch_ack_timeout_env_raises_only`: `60000` → raised to 60s; `1000`/`0`/`-5`/`"garbage"`/`""` → all clamp to the 30s floor. This is a reasonable, deliberate design choice: it closes exactly the failure mode this PR fixes (an operator accidentally *lowering* the bound and reintroducing the false negative) while still letting a slow-forge/heavily-loaded daemon raise it. Not flagging it further.
- Error message now reports the *resolved* timeout, not the constant.
- `token_name == UNKNOWN_TOKEN_NAME` surfaced distinctly in success output as a non-blocking hint, as promised.
- New test `slow_but_legitimate_ack_succeeds`: fake daemon sleeps 5.5s (past the retired 5s bound), queried with the real 30s default, asserts `SweepDispatched` parses successfully and `elapsed >= 5s` — this is exactly the "slow-but-legitimate" middle case my original review flagged as untested. Confirmed by re-running it myself, not just reading it.
- `serial_test` (used for the two env-mutating tests) was already a dev-dependency before this PR — no new dependency risk.

**Independently re-ran the verification, didn't just trust the report:**
- `cargo test --bin loom-daemon` (dispatch_tests): **8/8 pass**, including `slow_but_legitimate_ack_succeeds`, `resolve_dispatch_ack_timeout_defaults_to_floor`, `resolve_dispatch_ack_timeout_env_raises_only`.
- `cargo clippy --bin loom-daemon --all-targets -- -D warnings`: clean, zero warnings.
- `gh pr checks 3965`: all required checks pass (Rust Linting & Build Check, Rust Unit Tests, Detect Changes; Installer/MCP jobs skip as expected for a daemon-only diff).
- `gh pr view --json mergeStateStatus`: **CLEAN**.

**Acceptance criteria from #3952 — all met:**
- [x] `loom-daemon dispatch` enqueues via IPC, prints sweep id + log path, exits nonzero with a clear message on ack timeout, never hangs — now with a 30s bound that has real headroom over the daemon's documented worst-case pre-ack latency (label flip + up to 2s stagger + up to 5s token-name poll), so it won't false-positive on normal successful dispatches.
- [x] `--workspace` targets any registered root (verified flag plumbing in `Dispatch` subcommand + `build_dispatch_request`).
- [x] Docs updated (`daemon-reference.md` reflects the 30s bound + env override).

One non-blocking nit for a future pass, not worth another round-trip: the PR description's "What changed" / acceptance-criteria checklist text still says "5s" (stale from before the fix) even though the code, tests, and daemon-reference.md docs are all correctly updated to 30s. Purely cosmetic in the PR body — doesn't affect the shipped behavior.

Nice work closing out the false-negative risk with a real regression test for the case that actually matters.

Issue #3964, opened 2026-07-26
loom-api: agent_activity table doesn't exist in activity.db schema (same class of bug as #3951)


rjwalters:
## Context

While reviewing #3963 (fix for #3951 — daemon metrics collector 'no such table: agent_activity' errors), I found the same root-cause class of bug in the separate `loom-api` crate.

`loom-api/src/main.rs` (`get_metrics_summary`, `get_metrics_by_role`) queries the workspace's `.loom/activity.db` — the same physical database file the daemon writes to — for a table literally named `agent_activity`:

```rust
FROM agent_activity
WHERE {date_filter}
```

That table has never existed in `loom-daemon/src/activity/schema.rs::init_schema` (the real production schema). The nearest equivalent is `agent_metrics` (see how #3963 fixed the same mistake in `metrics_collector.rs`).

`loom-api`'s own unit tests (`setup_test_db` in `main.rs`) pass only because they hand-roll a fixture schema that includes `CREATE TABLE agent_activity (...)` — they never exercise the real `init_schema()`, so this drift wasn't caught. Against a real `activity.db`, any call to `get_metrics_summary` / `get_metrics_by_role` would hit `no such table: agent_activity` immediately.

## Suggested fix

Same pattern as #3963:
- Rewrite the `agent_activity` queries in `loom-api/src/main.rs` against `agent_metrics` (columns: `github_issue`, `started_at`/`completed_at`, `status`, `total_tokens`, etc. — see `activity/schema.rs`).
- Rebuild `setup_test_db()` on the real `loom_daemon::activity::init_schema()` (or equivalent) instead of the hand-rolled fixture, so tests track the production schema.

## Acceptance criteria

- [ ] `loom-api`'s metrics endpoints query `agent_metrics`, not the nonexistent `agent_activity`.
- [ ] Test fixtures build on the real activity-DB schema so future drift is caught by CI.

PR #3983, opened 2026-07-26
test(daemon): regression coverage + diagnostics for #3967 daemon self-claim marker

### LINKED ISSUE #3967: Daemon-dispatched sweep self-skips its own claim: LOOM_SWEEP_CLAIM_OWNED missing from child env
## Context

During the 2026-07-26 multi-repo canary run, the daemon dispatched a sweep for #3964 (`sweep-issue-3964-1785041362`, logged in `.loom/logs/sweep-issue-3964.log` at 2026-07-26T04:49:24Z). The child's pre-flight found `loom:building` applied ~13s earlier, checked `LOOM_SWEEP_CLAIM_OWNED`, found it **unset**, and skipped the issue as "another worker is in flight" — self-skipping the daemon's own claim. Net result: a token slot burned, zero work performed, and the issue bounced back to the queue.

This is exactly the failure mode the marker was built to prevent: `SweepRegistry::spawn_child` exports `LOOM_SWEEP_CLAIM_OWNED=<issue>` (`loom-daemon/src/sweep_registry.rs:1888`, #3823/#3829), and the sweep skill's pre-flight treats it as a self-claim exception. The marker demonstrably did not reach the child's environment on this dispatch.

Suspects worth checking:
- the new `loom-daemon dispatch <issue>` operator CLI path (#3952/#3965) — if it spawns or enqueues through a path that bypasses the `spawn_child` env block, dispatches from it would always self-skip;
- env propagation through `spawn-claude.sh` → `exec claude` → the skill's shell (snapshot sourcing dropping exported vars).

- [ ] Reproduce: dispatch via each entry path (work-finder, MCP `dispatch_sweep`, `loom-daemon dispatch` CLI) and assert `LOOM_SWEEP_CLAIM_OWNED` is present in the child env for all of them
- [ ] Fix the path that drops the marker; add a regression test at the dispatch-path level (not just `spawn_child` unit scope)
- [ ] Sweep pre-flight log line includes whether the marker was seen and its value, so future occurrences are diagnosable from the sweep log alone

rjwalters:
## Summary

Closes #3967.

Reproduces the reported incident — a daemon-dispatched sweep for #3964 self-skipping because `LOOM_SWEEP_CLAIM_OWNED` appeared unset in the child, even though `loom-daemon` had already flipped `loom:issue` -> `loom:building` and (per #3823/#3829) `SweepRegistry::spawn_child` already exports the marker — and adds regression coverage at each of the three dispatch entry points named in the issue, plus a diagnostic log line for future occurrences.

## What was found

**No live propagation bug.** All three dispatch entry points converge on the same underlying mechanism (`SweepRegistry::dispatch` -> `spawn_child`, which already sets `LOOM_SWEEP_CLAIM_OWNED` in the child's `Command::env()`):

- **MCP `dispatch_sweep`** and the **`loom-daemon dispatch <issue>` CLI** (#3952) both send an identical `Request::DispatchSweep` over the daemon's Unix socket, handled by the same `handle_request` match arm in `ipc.rs` (confirmed by reading `main.rs::handle_dispatch_command` / `build_dispatch_request`, which mirrors the MCP tool's field mapping 1:1). The CLI does **not** bypass `spawn_child` — I checked this explicitly per the issue's third suspect.
- **Autonomous work finder** dispatches through `RegistryDispatcher` (the production `WorkDispatcher` impl), which also forwards to `SweepRegistry::dispatch`.

New tests added at the dispatch-path level (broader than the existing `spawn_child`-only unit test) all **pass**, confirming the marker propagates correctly through all three surfaces:

- `loom-daemon/src/ipc.rs` — `test_handle_request_dispatch_sweep_exports_claim_ownership_marker`: drives `handle_request(Request::DispatchSweep{...})` end-to-end (the exact code both the MCP tool and the CLI round-trip into) and asserts the spawned child's env carries `LOOM_SWEEP_CLAIM_OWNED=<issue>`.
- `loom-daemon/src/work_finder.rs` — `test_registry_dispatcher_exports_claim_ownership_marker`: builds a real `SweepRegistry`-backed `RegistryDispatcher` (not the `RecordingDispatcher` test fake used elsewhere in that file) and asserts the same marker propagation through the work-finder's actual production dispatch path.

Given this, the #3967 incident was most likely something environment-specific to that one run (e.g. a stale/pre-#3823 daemon binary running in that repo, or a one-off race) rather than a reproducible code-path bug — so there is no `spawn_child` / `sweep_registry.rs` diff in this PR. The value delivered is closing the regression-coverage gap (so a *future* regression at any of these three entry points is caught) and adding always-on diagnostics so the next occurrence is debuggable straight from the sweep log.

## Diagnostics added (AC #3)

`defaults/scripts/spawn-claude.sh` now unconditionally logs `spawn-claude: LOOM_SWEEP_CLAIM_OWNED=<value|unset>` right after workspace/python resolution, on every spawn (daemon-dispatched or not). This line lands in `.loom/logs/sweep-issue-<N>.log` via `spawn_child`'s existing per-sweep log redirection, so the marker's presence/value is always directly observable rather than inferred — closing the gap between what `spawn_child`'s `Command::env()` call sets and what the child script's environment actually sees.

Two new bash tests in `defaults/scripts/tests/test-spawn-claude.sh` cover both states (unset and set to a specific issue number), with an explicit `env -u LOOM_SWEEP_CLAIM_OWNED` guard on the unset case since the test suite may itself run inside a daemon-dispatched session.

## Test plan

- [x] `cargo test --lib` in `loom-daemon/` — 824 passed, 0 failed (includes the 2 new tests plus the existing `sweep_registry::tests::dispatch_exports_claim_ownership_marker`)
- [x] `cargo build` and `cargo build --release` — both clean
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --lib --tests -- -D warnings` — clean
- [x] `bash defaults/scripts/tests/test-spawn-claude.sh` — 78/78 passed (includes the 2 new #3967 cases)
- [x] `shellcheck` on both touched bash files — clean

### DIFF
--- a/defaults/scripts/spawn-claude.sh
+++ b/defaults/scripts/spawn-claude.sh
@@ -108,6 +108,21 @@ _resolve_workspace() {
 WORKSPACE="$(_resolve_workspace)"
 PYTHON="${LOOM_PYTHON:-python3}"
 
+# --- Daemon self-claim marker visibility (Issue #3823, observability #3967) ---
+# `loom-daemon`'s `SweepRegistry::spawn_child` exports
+# `LOOM_SWEEP_CLAIM_OWNED=<issue>` into a daemon-dispatched child so its
+# `/loom:sweep` pre-flight recognises the daemon's own pre-dispatch
+# `loom:issue -> loom:building` label flip as ITS OWN claim rather than
+# self-skipping the issue as "another worker is in flight" (the #3967
+# incident). Log the marker's presence/value here — unconditionally, on
+# every spawn, whether daemon-dispatched or not — so a future occurrence is
+# diagnosable straight from `.loom/logs/sweep-issue-<N>.log` without needing
+# to reproduce the dispatch: this line, plus the `spawn-claude:` prefix, is
+# already captured by `spawn_child`'s per-sweep log redirection, so any gap
+# between the Rust `Command::env()` call and what this script's own
+# environment actually sees is now directly observable instead of inferred.
+log_info "spawn-claude: LOOM_SWEEP_CLAIM_OWNED=${LOOM_SWEEP_CLAIM_OWNED:-unset}"
+
 # --- Argument parsing ---
 USE_WRAPPER=false
 PASSTHROUGH_ARGS=()
--- a/defaults/scripts/tests/test-spawn-claude.sh
+++ b/defaults/scripts/tests/test-spawn-claude.sh
@@ -278,6 +278,25 @@ output=$(LOOM_WORKSPACE="$TEST_WS" PATH="$STUB_DIR:$PATH" \
 assert_contains "stub-claude ceiling=120000" "$output" \
     "spawn-claude preserves an operator-set wait ceiling (#3943)"
 
+# Test: the daemon self-claim marker's presence/value is always logged (Issue
+# #3967) — with the var unset, an interactive/manual invocation logs "unset".
+# Explicitly unset it first: this test suite may itself be running inside a
+# daemon-dispatched `/loom:sweep` session, which would otherwise leak an
+# ambient `LOOM_SWEEP_CLAIM_OWNED` into the subshell and false-fail this case.
+output=$(env -u LOOM_SWEEP_CLAIM_OWNED LOOM_WORKSPACE="$TEST_WS" PATH="$STUB_DIR:$PATH" \
+    "$SCRIPTS_DIR/spawn-claude.sh" -p "ping" 2>&1 || true)
+assert_contains "spawn-claude: LOOM_SWEEP_CLAIM_OWNED=unset" "$output" \
+    "spawn-claude logs LOOM_SWEEP_CLAIM_OWNED=unset when not daemon-dispatched (#3967)"
+
+# Test: with the var set (simulating a daemon-dispatched child, #3823), the
+# same log line echoes the exact issue number — the diagnostic this issue's
+# acceptance criteria calls for, straight from the per-sweep log.
+output=$(LOOM_WORKSPACE="$TEST_WS" PATH="$STUB_DIR:$PATH" \
+    LOOM_SWEEP_CLAIM_OWNED="3964" \
+    "$SCRIPTS_DIR/spawn-claude.sh" -p "ping" 2>&1 || true)
+assert_contains "spawn-claude: LOOM_SWEEP_CLAIM_OWNED=3964" "$output" \
+    "spawn-claude logs the daemon self-claim marker's value when set (#3967)"
+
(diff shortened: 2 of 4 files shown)


rjwalters:
LGTM — approved.

Verified the PR's core claim by independently reading the code (not just trusting the PR description):

- **Convergence confirmed.** `loom-daemon dispatch <issue>` CLI (`main.rs::handle_dispatch_command` → `build_dispatch_request`) builds the identical `Request::DispatchSweep` the MCP `dispatch_sweep` tool sends, both go over the Unix socket to the same `handle_request` match arm in `ipc.rs`. The work-finder's `RegistryDispatcher::dispatch` (`work_finder.rs`) forwards straight to `SweepRegistry::dispatch`. All three entry points bottom out in `SweepRegistry::spawn_child`, which already does `.env("LOOM_SWEEP_CLAIM_OWNED", issue.to_string())` (sweep_registry.rs:1888) — so the "CLI bypasses spawn_child" theory in the issue is refuted by direct inspection, not just asserted.
- **Tests are meaningful, not tautological.** The two new tests exercise dispatch at a broader scope than the existing `sweep_registry.rs` unit test: `test_handle_request_dispatch_sweep_exports_claim_ownership_marker` drives `handle_request(Request::DispatchSweep{...})` end-to-end (the exact server-side code both CLI and MCP round-trip into), and `test_registry_dispatcher_exports_claim_ownership_marker` builds a real `SweepRegistry`-backed `RegistryDispatcher` (not the `RecordingDispatcher` fake used by every other work_finder test) to cover the work-finder's actual production path. Both assert on the spawned child's real env via a fixture spawn binary, not on internal state.
- **Ran the full suite locally**: `cargo test --lib` in `loom-daemon/` — 824 passed, 0 failed, including both new tests individually verified. `cargo fmt --check` clean.
- **CI**: all checks pass (Rust Unit Tests, Rust Linting & Build Check, Installer Integration Tests, Shellcheck, Detect Changes). `mergeStateStatus` is CLEAN / MERGEABLE.
- **Log line placement**: the new `log_info "spawn-claude: LOOM_SWEEP_CLAIM_OWNED=..."` in `spawn-claude.sh` is placed right after workspace/python resolution, uses the existing `log_info` helper (captured by the per-sweep log redirection), and only ever logs an issue number or the literal `unset` — no secret material.
- The PR's "no reproducible bug, only regression-coverage + diagnostics" conclusion is a defensible, well-evidenced outcome for #3967 given this is confirmed dead code has never diverged from this path in either client.

Nice investigative rigor — closing coverage gaps and adding always-on diagnostics is the right call when a reported incident can't be reproduced from the code as it stands today.

PR #3962, opened 2026-07-26
fix: throttle main-health-gate skip logging and ignore build-artifact dirt

rjwalters:
## Summary

The reactive main-health gate (`loom-daemon/src/main_health_gate.rs`) treated
*any* dirty file in the workspace as an indeterminate "skip" — correct
fail-safe intent (never hard-reset over operator edits), but with two defects
observed on the canary host: a single modified `mcp-loom/package-lock.json`
disabled gate evaluation for the whole repo indefinitely, and the resulting
skip warning logged unthrottled on every ~30s tick (2,035 lines in one day).
`loom-daemon status` also had no way to distinguish "the gate ran and found
main red" from "the gate never got to run because the tree was dirty."

## Changes

- **Ignore-list for the dirty-tree check** (AC1): `prepare_workspace_to_origin_main`'s
  cleanliness check now excludes (a) Loom-owned transient paths — the same
  set `.loom/scripts/check-main-clean.sh` ignores for #3778 (`.loom/logs/`,
  `.loom/worktrees/`, `.loom/tokens/`, `.loom-managed`, etc.) — and (b) a
  small, documented set of common regenerable lockfile basenames
  (`package-lock.json`, `pnpm-lock.yaml`, `yarn.lock`, `Cargo.lock`,
  `uv.lock`) that a build step can rewrite with no dependency change. Both
  are dirt a hard reset to `origin/main` safely discards, never a reason to
  skip the sync step. A genuine unexpected dirty file (alongside or without
  ignorable dirt) still skips, exactly as before.
- **Fixed a latent column-alignment bug** found while adding the above:
  `run_git()`'s blanket `.trim()` strips the leading space of the *first*
  porcelain line when it starts with `" M"` (unstaged modification of a
  tracked file), which would have silently broken the new path-column
  parsing. Added `git_status_porcelain()`, a dedicated raw-stdout helper that
  only trims trailing whitespace, and switched the cleanliness check to use
  it.
- **Throttled skip-warn logging** (AC2): `MainHealthState` now tracks
  skip/not-skip per tick. The "gate run SKIPPED" line fires once on a
  clean-to-dirty transition, then at most once per hour while the tree stays
  dirty (`SKIP_WARN_THROTTLE`), replacing the previous every-tick WARN.
  Recovery (a completed Green/Red tick) resets the throttle so the next
  dirty streak warns immediately again.
- **Split status reporting** (AC3): `DaemonStatusReport.main_health_gate_not_evaluated`
[...]

Issue #3950, opened 2026-07-25
main-health gate: any dirty file disables a workspace's gate indefinitely + WARNs every 30s (2000+ spam lines from one modified lockfile)


rjwalters:
Observed on the canary host: a single modified `mcp-loom/package-lock.json` (build side-effect) in the loom workspace caused `main_health_gate` to skip every gate run for that repo — correct fail-safe intent (never hard-reset over operator edits) but with two defects:

1. **All-or-nothing skip**: one benign untracked/modified file (lockfiles, `.loom/` transient state) disables main-health protection for the whole repo indefinitely. The gate should either run in a temp worktree/clone of `origin/main` (never touching the operator tree — also removes the reset risk entirely), or at minimum ignore Loom-owned transient paths the way `check-main-clean.sh` does (#3778).
2. **Log spam**: the skip WARN repeats every 30s per workspace — **2,035 lines** in one day. Log the skip once per state transition (clean→dirty), then throttle to e.g. hourly.

## Acceptance criteria
- [ ] Gate evaluation isolated from operator working tree (temp worktree/clone of origin/main) OR a documented ignore-list matching the #3778 transient-path set.
- [ ] Skip warning emitted on transition + throttled, not every interval.
- [ ] Per-repo gate state in `status` distinguishes 'halted (red main)' from 'not evaluated (dirty tree)'.

Found during v0.15.0 canary rollout.

Issue #3985, opened 2026-07-26
Local build gate is environment-sensitive where CI is not: dead tmux + CPU starvation from the gate's own loop make it self-reddening


rjwalters:
## Summary

The local `buildGate.command` and GitHub CI run the **identical** command (`cargo test --workspace` — `.github/workflows/ci.yml:69` vs `.loom/scripts/build-gate.sh`), but they measure different things. CI measures **the commit**. The local gate measures **the host it runs on** — and on a host that is actually running Loom, it fails for reasons that have nothing to do with `main`.

Observed 2026-07-26 with `origin/main` green on every GitHub check while the local gate halted tier-0 dispatch for hours.

### 1. Dead tmux server → 6 integration tests fail

```
test result: FAILED. 3 passed; 6 failed → -p loom-daemon --test integration_basic
```

All six assert `tmux_session_exists(...)` (`integration_basic.rs` lines 132, 166, 239, 294, 357, 407). On this host:

```
$ tmux ls
error connecting to /private/tmp/tmux-501/default (No such file or directory)
```

A GitHub runner has a working tmux, so CI passes. The gate does not distinguish "tmux is unavailable on this machine" from "the terminal code regressed".

### 2. CPU starvation → 3 timing-sensitive tests fail

```
test result: FAILED. 819 passed; 3 failed → -p loom-daemon --lib
  sweep_registry::tests::dispatch_and_reap_wire_the_sweep_journal          (sweep_registry.rs:4094)
  sweep_registry::tests::dispatch_applies_configured_stagger_between_spawns (sweep_registry.rs:7257)
  sweep_registry::tests::dispatch_forwards_derived_package_path_when_unset  (sweep_registry.rs:4264)
```

`sweep_registry.rs:4094` is a wall-clock assertion:

```rust
let dead = wait_for_condition(10_000, || !is_pid_alive(outcome.pid));
assert!(dead, "fixture child did not exit within 10s");
```

A fixture child that normally exits immediately fails to be scheduled within 10 seconds under load.

## The self-reddening loop

These two classes combine with #3984 into a closed cycle:

1. `realChangeGlobs` is dead config (#3984), so the gate re-runs the full workspace suite continuously — ~100% duty cycle, even though `main` has not moved.
2. That sustained compile saturates the CPU.
3. Under saturation, the timing-sensitive tests above fail (and other runs exceed the 600s timeout).
4. The failure is classified RED (#3974), halting dispatch.
[...]

PR #3986, opened 2026-07-26
fix(daemon): classify gate outcomes VERIFIED_RED vs UNEVALUATED; only verified red halts dispatch

### LINKED ISSUE #3974: Main-health gate conflates 'gate could not run' with 'main is RED' — environmental failures halted tier-0 dispatch while GitHub CI was green
## Context

During the 2026-07-26 incident, loom (priority 10) was halted by the main-health gate for hours while `origin/main` (5ac71aaf) was **green on GitHub CI** (CI, Security Scan, Lines of Code all passing). Every RED verdict was an environmental failure of the gate run itself, not a verified-red main:

- `gate command 'bash .loom/scripts/build-gate.sh' timed out after 600s` → reported "main still RED"
- exit status 101 while the daemon's process tree had broken XPC (gh/git dead) → "main still RED"
- exit status **127** (cargo not on PATH after a launchd migration) → "main is RED — HALTING"
- `loom-daemon status` additionally showed "NOT EVALUATED (workspace tree is dirty)" while `git status` on that workspace was completely clean
- work_finder logged "main-health gate cleared — resuming dispatch" at 02:50:21 in the same window main_health_gate was logging "still RED" (skip-vs-halt disagreement between the two loops)

d901ea6c already split halted/not-evaluated status, but the dispatch decision still treats *any* nonzero gate exit as RED. A gate that cannot run tells you nothing about main; halting the self-repair repo on it creates exactly the bootstrap deadlock the gate was meant to avoid (red verdicts blocked the daemon from dispatching its own fixes).

- [ ] Classify gate outcomes: VERIFIED_RED (gate ran to completion, checks failed) vs UNEVALUATED (timeout, exit 127/126, git fetch failure, dirty tree). Only VERIFIED_RED halts dispatch; UNEVALUATED keeps the previous verdict and logs loudly with the failure class
- [ ] The "workspace tree is dirty" detection agrees with `git status --porcelain` on the same root (reproduce why a clean tree reported dirty; fix or log the actual paths it saw)
- [ ] work_finder and main_health_gate derive halt/clear from one shared state so their logs can't disagree
- [ ] Optional corroboration: when the local gate is UNEVALUATED, consult the forge's CI status for origin/main HEAD as a fallback signal before keeping loom halted

rjwalters:
## Summary

The main-health gate treated **any** nonzero gate exit as "main is RED" and halted autonomous dispatch. Most nonzero exits mean the gate *could not run* — a timeout, `sh` exit 127 because `cargo` is not on the daemon's `PATH`, a spawn failure, a `git fetch` killed by a broken process tree — and say nothing about `main`. A safety gate that fails closed on its **own** infrastructure failures converts every environmental hiccup into a total dispatch outage, and for the repo that holds the gate's own source into a bootstrap deadlock.

Every gate run now resolves to exactly one of three outcomes. **Only VERIFIED_RED halts.**

| Outcome | Meaning | Effect on dispatch |
|---------|---------|--------------------|
| `Green` | ran to completion, exit 0 | clears any halt |
| `Red` (**VERIFIED_RED**) | **ran to completion** and reported failure | **halts** |
| `Unevaluated` (**UNEVALUATED**) | did not run to completion | **preserves the previous verdict**, logs loudly with the class |

> **This PR does not unblock the currently-halted `loom` repo, and does not claim to.** The most recent gate verdict on this host is a genuine completed-and-failed run (`cargo test` exit 101 with real `test result: FAILED` output), which the new classifier correctly routes to VERIFIED_RED. What may unblock it is the AC4 forge-CI corroboration below — *if* CI is green on the evaluated SHA.

### 1. Outcome classification (`main_health_gate.rs`)

`GateOutcome::Skipped { reason }` → `GateOutcome::Unevaluated { class, reason }`, with a new `UnevaluatedClass`:

| Class | Trigger | Was, pre-PR |
|-------|---------|-------------|
| `timeout` | exceeded `buildGate.timeoutSeconds` | **Red → halted** |
| `command-not-executable` | `sh` exit 127 (not found) / 126 (not executable) | **Red → halted** |
| `killed-by-signal` | signal death (`code() == None`, or `sh`'s 137/143) | **Red → halted** |
| `spawn-failure` | could not spawn / IO error creating or polling output | **Red → halted** |
| `git-failure` | `rev-parse` / `status` / `fetch` / `rev-list` / `reset` failed | Skipped (unchanged) |
| `dirty-tree` | non-ignorable local changes | Skipped (unchanged) |
| `not-on-main`, `local-ahead` | branch / ahead-of-origin guards | Skipped (unchanged) |
| `contradicted-by-forge-ci` | ran and failed, but CI is green on the same commit | **Red → halted** |

The discriminator is deliberately narrow so a genuinely failing build still halts: **any other non-zero exit is trusted as VERIFIED_RED** — `cargo test`'s 101 for a failing test still halts dispatch (`test_command_runner_cargo_style_failure_is_still_verified_red`).

`HealthTransition::Skipped` → `Unevaluated`; `apply_gate_outcome` is unchanged in shape (unevaluated is still a no-op on the halt flag) — what changed is *which* outcomes reach `Red`.

### 2. Forge-CI corroboration of a local red (AC4)

A local run can fail *because of the host it runs on*. Observed on this host: six `integration_basic` tests assert `tmux_session_exists(...)` and fail because the tmux server is dead (`tmux ls` → `error connecting to /private/tmp/tmux-501/default`), while `.github/workflows/ci.yml:69` runs the identical `cargo test --workspace` on the same commit and passes. The local gate measures **this host**; forge CI measures **the commit**. Exit-code inspection alone cannot tell that apart from a real regression.

So a completed-and-failed run is cross-checked against the forge's CI conclusion for the exact `origin/main` SHA the gate evaluated (post-sync `HEAD`), via `gh run list --branch main --json headSha,status,conclusion,workflowName` matched on `headSha`:

- CI **green** on that SHA → `contradicted-by-forge-ci` (UNEVALUATED), logged loudly naming the commit, **does not halt**.
- CI **red** on that SHA → corroborated; still VERIFIED_RED, still halts.
- CI **unknown** (no completed run yet, `gh` missing/unauthenticated, 30s probe timeout) → fail safe: still VERIFIED_RED, still halts.

Only **positive contrary evidence** ever relaxes a halt. It is probed only on a local red (never on a green run), and behind a `ForgeCiStatus` trait so it is tested without network or credentials. Kill switch: `LOOM_GATE_CI_CORROBORATION=0`.

### 3. One shared halt state (AC3)

`work_finder::tick_multi` now derives `TickReport.halted` **directly** from the shared `WorkspaceHealthStates` flags:

```rust
let any_halted = halted.iter().take(workspaces.len()).any(|&h| h);
```

Root cause of the disagreement: the per-workspace loop `continue`s on a `list_ready_issues` error **before** it reaches its halt check, and `any_halted` was accumulated inside that loop. During the incident `gh` was dead in the daemon's process tree, so the listing failed every tick and the finder reported `halted = false` — logging `work_finder: main-health gate cleared — resuming dispatch` at 02:50:21 in the same window `main_health_gate` was logging `still RED`. The two loops now read one source of truth and cannot disagree.

### 4. Diagnosable not-evaluated status (AC2)

The "workspace tree is dirty" claim on a clean tree was **a reporting artifact, twice over**:

- `loom-daemon status` **hard-coded** `(workspace tree is dirty)` for *every* skip class. The last skip in the incident log was `git fetch origin main failed (exit status: 128: No user exists for uid 501)` — a `git-failure`, not a dirty tree.
- The running daemon binary predates #3962 (`d901ea6c`); its dirty message (from `a1b898b4`) named **no paths at all**: `workspace has local changes — skipping gate (will not hard-reset over operator edits)` with nothing after it. Verified in `~/.loom/daemon.log`: 11 such lines, zero paths.

The detection itself **does** agree with `git status --porcelain` on the same root — it literally shells `git status --porcelain` with `current_dir(repo_root)`, and #3950's ignore-list only ever *reduces* the dirt it counts. Fixed on the reporting side:

- Dirty-tree reasons now name the root and the exact porcelain lines: ``` `git -C /path status --porcelain` reports 2 non-ignorable change(s) …: [ M src/main.rs | ?? scratch.tmp] ```
- Every skip reason names its root, so it is checkable by hand.
- `DaemonStatusReport.main_health_gate_not_evaluated_reason` and `per_repo[].health_gate_not_evaluated_reason` (both `#[serde(default)]`, wire-compatible) carry `"<class>: <reason>"`; `loom-daemon status` renders it verbatim instead of asserting dirt.

### 5. Logging

UNEVALUATED lines now name the class and state plainly that this is not evidence about `main`:

```
main_health_gate: /path gate run UNEVALUATED [command-not-executable] — gate command
'bash .loom/scripts/build-gate.sh' exited 127 (command not found — a tool the gate needs is
not on the daemon's PATH); main was NOT evaluated — this is a failure of the GATE, not
evidence about main; dispatch remains ALLOWED (no previous verified-red run)
```

The #3950 skip-warn throttle is preserved, and additionally **re-warns immediately on a change of failure class**, so the incident's rotation (timeout → 101 → 127 → fetch failure) is never swallowed behind the first class's 1h window.

- [x] **AC1 — classify VERIFIED_RED vs UNEVALUATED; only VERIFIED_RED halts.** `GateOutcome::Unevaluated { class, reason }` + `classify_failed_exit`. `test_unevaluated_outcomes_never_halt_dispatch` asserts all 9 classes leave the halt flag untouched; `test_command_runner_{timeout,exit_127,exit_126,signal_death}_is_unevaluated_not_red` cover the live subprocess paths; `test_prepare_skips_when_fetch_fails_offline` asserts `GitFailure`.
- [x] **AC1 — UNEVALUATED preserves the previous verdict and logs loudly naming the class.** `apply_gate_outcome` no-op (`test_unevaluated_outcome_leaves_halt_flag_unchanged`); `log_transition{,_for_root}` render `[<class>]` + `verdict_phrase(halted)`; `test_note_gate_tick_rewarns_immediately_on_class_change`.
- [x] **AC1 — a real check failure still halts.** `test_command_runner_cargo_style_failure_is_still_verified_red` (exit 101 with test output → `EnteredHalt`).
- [x] **AC2 — dirty detection agrees with `git status --porcelain` on the same root.** Reproduced as a reporting artifact (see §4) and fixed: reasons name the root + exact porcelain lines (`test_prepare_skips_dirty_workspace`), status renders the real cause (`format_gate_status_names_the_actual_not_evaluated_cause` asserts a `git-failure` cause never renders as "dirty").
- [x] **AC3 — work_finder and main_health_gate derive halt from one shared state.** `test_tick_multi_halt_survives_a_failing_forge_query` reproduces the exact incident (`err_source("gh: No user exists for uid 501")` + halted repo → `report.halted == true`), plus `test_tick_multi_halt_reported_even_when_repo_has_no_backlog` and `test_tick_multi_extra_halt_entries_are_ignored`.
- [x] **AC4 — forge-CI corroboration (implemented, not deferred).** `ForgeCiStatus` / `GhForgeCi` / `parse_gh_run_list`; five behavioural tests plus verdict-parsing coverage including "a green run for a **different** commit must never vouch for this one".

## Test Plan

```
cargo clippy --all-targets    # clean
cargo fmt --check             # clean
cargo test -p loom-daemon --lib        # 843 passed; 0 failed
cargo test -p loom-daemon --bin loom-daemon  # 10 passed; 0 failed
```

New/updated tests: 21 in `main_health_gate`, 3 in `work_finder`, 2 in `main.rs`, 1 extended in `ipc.rs`.

**Pre-existing environment flakes on this host (not caused by this PR):**

- `work_finder::tests::test_registry_dispatcher_exports_claim_ownership_marker` (added in `53a262fb`, the branch point) intermittently fails under the current heavy CPU contention — it spawns a real child and polls a file for 5s. Passes in isolation; failed once in a full-suite run and passed on re-run.
- `cargo test --workspace` fails 6 `integration_basic` tests that assert `tmux_session_exists(...)` (lines 132/166/239/294/357/407) because this host's tmux server is dead. CI runs the identical command and passes. **This is the divergence AC4 exists to detect**, and is being filed separately by the operator — this PR does not fix it.

## Notes for the reviewer

- **Wire compatibility**: both new status fields are `Option<String>` with `#[serde(default)]`; `test_daemon_status_report_backward_compatible_without_not_evaluated` still parses pre-#3950 JSON.
- **Signal exits**: `sh`-reported 137/143 are classified `killed-by-signal` alongside a raw `code() == None`. A build tool essentially never *chooses* those as an exit status, whereas OOM kills on a contended host produce them routinely. If a reviewer disagrees this is the one judgment call worth revisiting.
- **`gh` in the gate path**: the corroboration probe adds one bounded (30s) `gh` subprocess **only on a local red**. If `gh` is broken — as it was during the incident — the probe returns `Unknown` and the local verdict stands, so it never makes things worse.

Closes #3974

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -555,12 +555,32 @@ matched on `headSha`:
 - CI **green** on that SHA → downgraded to `contradicted-by-forge-ci`
   (UNEVALUATED); logged loudly, **does not halt**.
 - CI **red** on that SHA → corroborated; still VERIFIED_RED, still halts.
-- CI **unknown** (no completed run yet, `gh` missing/unauthenticated, probe
-  timeout > 30s) → fail safe: still VERIFIED_RED, still halts.
+- CI **unknown** → fail safe: still VERIFIED_RED, still halts.
 
-Only *positive* contrary evidence ever relaxes a halt. Corroboration is on by
-default and can be disabled with `LOOM_GATE_CI_CORROBORATION=0` (for repos with no
-forge CI or no `gh`); it is only probed on a local red, never on a green run.
+Only *positive* contrary evidence ever relaxes a halt, so "green" is the hardest
+verdict to reach and everything short of an unambiguous all-clear degrades to
+**unknown**:
+
+| Runs for the evaluated SHA | Verdict |
+|---|---|
+| any `failure` / `timed_out` / `startup_failure` | red |
+| any run not yet `completed` (`queued`, `in_progress`, …) | unknown — CI has not judged the commit yet |
+| any `cancelled` / `action_required` / `stale` / unrecognized conclusion | unknown — the workflow reached no verdict about the code |
+| at least one `success`, every other run `skipped` / `neutral` | **green** |
+| no run for the SHA, unparseable output, `gh` missing/unauthenticated, probe timeout > 30s | unknown |
+
+**Absence of failure is not success.** Requiring at least one genuine `success`
+*and* no outstanding or indeterminate sibling run closes two ways a
+"saw any completed run ⇒ green" reducer reads green on non-evidence: a
+`cancel-in-progress: true` concurrency group leaves superseded runs at
+`completed/cancelled` **forever** (a permanent false green for that commit), and a
+fast bookkeeping workflow that finishes minutes before the real build would
+otherwise vouch for the commit on its own. Neither case needs the probe to know
+*which* workflow "counts", so no workflow name is hard-coded.
+
+Corroboration is on by default and can be disabled with
+`LOOM_GATE_CI_CORROBORATION=0` (for repos with no forge CI or no `gh`); it is only
+probed on a local red, never on a green run.
 
 ### One shared halt state
 
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -502,6 +502,81 @@ enabling the gate for a genuine multi-repo deployment is done with the machine-
 global `LOOM_MAIN_HEALTH_GATE=1` env var (each repo's own `buildGate` block then
 decides whether it actually gates).
 
+## Gate verdicts: VERIFIED_RED vs UNEVALUATED (#3974)
+
+A safety gate that fails closed on **its own** infrastructure failures converts
+every environmental hiccup into a total dispatch outage — and, for the repo that
+contains the gate's own source, into a bootstrap deadlock where the daemon cannot
+dispatch the fix for the thing that is broken. Before #3974 *any* non-zero gate
+exit (including a timeout, `sh` exit 127 because `cargo` was not on the daemon's
+`PATH`, or a spawn failure) was recorded as "main is RED — HALTING".
+
+Every gate run now resolves to exactly one of three outcomes
+(`main_health_gate::GateOutcome`):
+
+| Outcome | Meaning | Effect on dispatch |
+|---------|---------|--------------------|
+| `Green` | command ran to completion, exit 0 | clears any halt |
+| `Red` (**VERIFIED_RED**) | command **ran to completion** and reported failure | **halts** this repo's dispatch |
+| `Unevaluated` (**UNEVALUATED**) | the gate produced no verdict | **preserves** the previous verdict; logs loudly with the class |
+
+`UnevaluatedClass` names why, and is surfaced in the daemon log and in
+`loom-daemon status`:
+
+| Class | Trigger |
+|-------|---------|
+| `dirty-tree` | non-ignorable local changes; the workspace was never synced |
+| `not-on-main` | workspace is on another branch / detached HEAD |
+| `local-ahead` | local `main` carries commits `origin/main` lacks (#3912) |
+| `git-failure` | a `git` step failed (`rev-parse` / `status` / `fetch` / `rev-list` / `reset`) |
+| `timeout` | the gate command exceeded `buildGate.timeoutSeconds` |
+| `command-not-executable` | `sh` exit 127 (not found) or 126 (not executable) |
+| `killed-by-signal` | terminated by a signal (e.g. an OOM kill) |
+| `spawn-failure` | could not spawn the command, or an I/O error while capturing output |
+| `contradicted-by-forge-ci` | ran and failed locally, but forge CI is green on the same commit (below) |
+
+The discriminator is deliberately narrow so a genuinely failing build still halts:
+**any other non-zero exit is trusted as VERIFIED_RED** — `cargo test`'s 101 for a
+failing test still halts dispatch.
+
+### Forge-CI corroboration of a local red
+
+A local run can also fail *because of the host it runs on*: on the incident host
+six `integration_basic` tests assert `tmux_session_exists(...)` and fail because
+the tmux server is dead, while `.github/workflows/ci.yml` runs the identical
+`cargo test --workspace` on the same commit and passes. The local gate measures
+**this host**; forge CI measures **the commit**.
+
+So a completed-and-failed run is cross-checked against the forge's CI conclusion
+for the exact `origin/main` SHA the gate evaluated (post-sync `HEAD`), via
+`gh run list --branch main --json headSha,status,conclusion,workflowName`
+matched on `headSha`:
+
+- CI **green** on that SHA → downgraded to `contradicted-by-forge-ci`
+  (UNEVALUATED); logged loudly, **does not halt**.
+- CI **red** on that SHA → corroborated; still VERIFIED_RED, still halts.
+- CI **unknown** (no completed run yet, `gh` missing/unauthenticated, probe
+  timeout > 30s) → fail safe: still VERIFIED_RED, still halts.
+
+Only *positive* contrary evidence ever relaxes a halt. Corroboration is on by
+default and can be disabled with `LOOM_GATE_CI_CORROBORATION=0` (for repos with no
+forge CI or no `gh`); it is only probed on a local red, never on a green run.
+
+### One shared halt state
+
+`work_finder::tick_multi` derives `TickReport.halted` **directly** from the
+`WorkspaceHealthStates` flags the gate writes, rather than accumulating it inside
+the candidate-gathering loop. Previously a `list_ready_issues` error short-circuited
+the loop *before* its halt check, so a repo whose forge query failed reported
+`halted = false` — the incident's `work_finder: main-health gate cleared —
+resuming dispatch` logged in the same window the gate logged `still RED`. The two
+loops now read one source of truth and cannot disagree.
+
+`loom-daemon status` reports the not-evaluated **cause** verbatim
+(`main_health_gate.not_evaluated_reason`, `per_repo[].health_gate_not_evaluated_reason`)
+instead of the pre-#3974 hard-coded "workspace tree is dirty", which misreported
+timeouts, missing tools, and `git fetch` failures on a clean tree as a dirty tree.
+
 ## Per-workspace priority tiers (#3946)
 
 By default the multi-repo work-finder and epic supervisor iterate
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -493,7 +493,10 @@ pub fn build_daemon_status(
             in_flight_count: live.len(),
             health_gate_halted: health_states.is_halted(root),
             quarantined_issues,
-            health_gate_not_evaluated: health_states.is_skipped(root),
+            health_gate_not_evaluated: health_states.is_unevaluated(root),
+            // Name the actual failure class (#3974 AC2) rather than letting the
+            // renderer assume "dirty tree" for every unevaluated tick.
+            health_gate_not_evaluated_reason: health_states.unevaluated_summary(root),
         });
         in_flight.extend(live);
     }
@@ -556,7 +559,8 @@ pub fn build_daemon_status(
         // Top-level halt preserves its pre-#3930 single-workspace meaning: the
         // daemon's own primary workspace. Per-repo halt is in `per_repo`.
         main_health_gate_halted: health_states.is_halted(fallback_root),
-        main_health_gate_not_evaluated: health_states.is_skipped(fallback_root),
+        main_health_gate_not_evaluated: health_states.is_unevaluated(fallback_root),
+        main_health_gate_not_evaluated_reason: health_states.unevaluated_summary(fallback_root),
         capacity,
         per_repo,
     }
@@ -2525,6 +2529,7 @@ exit 0
             dynamic_cap: 3,
             main_health_gate_halted: true,
             main_health_gate_not_evaluated: false,
+            main_health_gate_not_evaluated_reason: None,
             capacity: crate::types::CapacityReport {
                 ranking_present: true,
                 total_accounts: 4,
@@ -2540,6 +2545,7 @@ exit 0
                 health_gate_halted: true,
                 quarantined_issues: vec![101, 202],
                 health_gate_not_evaluated: false,
+                health_gate_not_evaluated_reason: None,
             }],
         };
         let resp = Response::DaemonStatus(report);
@@ -2653,14 +2659,32 @@ exit 0
         // "halted (red main)" and "not evaluated (dirty tree)" can both be
         // true — a prior red run's halt persisting while a later tick can't
         // even evaluate because the tree went dirty.
-        health
-            .get_or_create(&root)
-            .note_gate_tick(true, std::time::Duration::from_secs(3600));
+        health.get_or_create(&root).note_gate_tick(
+            Some((
+                crate::main_health_gate::UnevaluatedClass::DirtyTree,
+                "operator edit in src/main.rs",
+            )),
+            std::time::Duration::from_secs(3600),
+        );
         let report = build_daemon_status(&pool, &health, &root);
         assert!(report.main_health_gate_halted, "prior halt persists through a skip");
         assert!(report.main_health_gate_not_evaluated, "skip surfaces as not-evaluated");
         assert!(report.per_repo[0].health_gate_halted);
         assert!(report.per_repo[0].health_gate_not_evaluated);
+        // #3974 AC2: the report names the actual failure class + detail rather
+        // than leaving the renderer to assume "workspace tree is dirty".
+        let reason = report
+            .main_health_gate_not_evaluated_reason
+            .as_deref()
+            .expect("not-evaluated reason recorded");
+        assert!(reason.starts_with("dirty-tree: "), "got: {reason}");
+        assert!(reason.contains("src/main.rs"), "got: {reason}");
+        assert_eq!(
+            report.per_repo[0]
+                .health_gate_not_evaluated_reason
+                .as_deref(),
+            Some(reason)
+        );
 
         match prev_shared {
             Some(v) => std::env::set_var("LOOM_SHARED_TOKENS_DIR", v),
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -1470,11 +1470,13 @@ fn print_status_json(
         },
         "main_health_gate": {
             "halted": report.main_health_gate_halted,
-            // "Not evaluated" (dirty tree) is distinct from "halted" (red
-            // main) — #3950 AC3. Both can be true at once: a prior halt from
-            // a genuinely red run persists untouched while a later tick can't
-            // even evaluate because the tree went dirty.
+            // "Not evaluated" is distinct from "halted" (verified-red main) —
+            // #3950 AC3. Both can be true at once: a prior halt from a
+            // genuinely red run persists untouched while a later tick can't
+            // even evaluate (dirty tree, timeout, missing tool, broken `git`).
             "not_evaluated": report.main_health_gate_not_evaluated,
+            // Which failure class actually blocked evaluation (#3974 AC2).
+            "not_evaluated_reason": report.main_health_gate_not_evaluated_reason,
         },
         // Per-repo breakdown across every registered managed workspace (#3930).
         "per_repo": report.per_repo.iter().map(|r| serde_json::json!({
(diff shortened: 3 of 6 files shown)


rjwalters:
❌ **Changes Requested** — one blocking fail-open in the AC4 forge-CI corroboration. Everything else in this PR verifies clean, and the core classification work is exactly right.

## What I verified (and it holds up)

**AC1 — the VERIFIED_RED default is genuinely the default.** `classify_failed_exit` (`loom-daemon/src/main_health_gate.rs:1283`) is a closed literal match on `Some(127) | Some(126) | Some(137) | Some(143) | None`, with `Some(_) => GateOutcome::red(...)` as the fallthrough. No `UnevaluatedClass` can widen it — the classes are assigned at their construction sites, not derived from the exit code, so there is no way for a class to accidentally capture a novel exit status. `cargo test`'s 101 lands in the default arm. `test_command_runner_cargo_style_failure_is_still_verified_red` asserts both the outcome *and* `HealthTransition::EnteredHalt` + `state.is_halted()`. SIGSEGV/SIGABRT (139/134) also fall through to VERIFIED_RED — asymmetric with 137/143, but it errs in the safe direction, so no objection.

**AC3 — the `tick_multi` root cause is real, not folklore.** Read the loop at `work_finder.rs:552-571`: the `list_ready_issues` error arm does `report.errors += 1; log::warn!(...); continue;` and the per-repo halt check at line 567 sits *after* it. The pre-fix `any_halted = true` lived inside that halt branch, so a failing forge query genuinely skipped it. The fix derives `any_halted` from the same `&[bool]` the caller builds at `work_finder.rs:1100` via `health_states.is_halted(r)` — the identical `WorkspaceHealthStates` the gate writes through `apply_gate_outcome`. The two loops now cannot disagree. `test_tick_multi_halt_survives_a_failing_forge_query` reproduces the incident precisely, and `test_tick_multi_extra_halt_entries_are_ignored` covers the `take(workspaces.len())` bound.

**SHA exactness.** `evaluated_sha` is resolved by `resolve_head_sha` only on the `PrepOutcome::Ready` path, i.e. after the hard reset to `origin/main`, so it is exactly the commit the gate command builds. `parse_gh_run_list` skips every run whose `headSha != sha`. A `None` SHA (sync disabled / `rev-parse` failed) short-circuits to `GateOutcome::red` without probing. Correct.

**Tests.** `cargo test -p loom-daemon --lib -- main_health_gate work_finder` → **138 passed, 0 failed**. A full `--lib` run showed 831/12, but all 12 (`sweep_registry::*`, `ipc::tests::test_handle_request_dispatch_sweep_*`) pass on a serial re-run — the #3985 host-contention flakes, all in modules this PR does not touch. `gh pr checks` is green on this commit (Rust Unit Tests, Rust Linting & Build Check), `mergeStateStatus` is CLEAN. Docs in `.loom/docs/daemon-reference.md` accurately describe the shipped behavior. No new labels.

---

## 🔴 Blocking: `parse_gh_run_list` accepts non-evidence as positive contrary evidence

`loom-daemon/src/main_health_gate.rs:737-762`. The reducer marks a commit green when **any** completed run for the SHA is not `failure|timed_out|startup_failure`:

```rust
saw_completed = true;
if let Some("failure" | "timed_out" | "startup_failure") = ... { return CiVerdict::Failure; }
...
if saw_completed { CiVerdict::Success } else { CiVerdict::Unknown }
```

Two fail-open paths, both concretely reachable in *this* repo:

**1. `cancelled` counts as green — permanently.** `.github/workflows/ci.yml:9-11` sets `concurrency: group: ci-${{ github.ref }}` with `cancel-in-progress: true`, and every push to `main` shares the single `refs/heads/main` group. A merge that lands while the previous commit's CI is still running leaves that commit's CI run as `status=completed, conclusion=cancelled` — forever. `cancelled` is not in the failure set, so `saw_completed` flips true and the SHA reads `Success`. Same for `action_required` and `stale`.

**2. A bookkeeping workflow vouches for the commit on its own.** The probe accepts *any* completed run, and does not use the `workflowName` it already asks `gh` for. `.github/workflows/loc.yml` ("Lines of Code") runs on every push to `main`, has **no** concurrency guard, and verifies nothing. Measured against real runs on `main`:

```
53a262fb Lines of Code  success  start=11:25:44Z  end=11:26:18Z   (~34s)
53a262fb CI             success  start=11:25:44Z  end=11:28:02Z   (~2m18s)
```

So for ~100s after every push — and permanently whenever CI was cancelled — the probe answers `Success` on the strength of a line-counting job while the actual verification workflow has not judged the commit.

**Why this is blocking rather than a follow-up.** The whole point of AC4 is stated in the PR body and in `daemon-reference.md`: *"Only positive contrary evidence ever relaxes a halt"* / *"CI unknown → fail safe: still VERIFIED_RED, still halts."* A cancelled run and an unfinished CI run are the definition of *unknown*, yet both currently relax the verdict. The consequence is the one failure mode this PR must not introduce: a genuinely red `main` where the gate declines to halt, while emitting the actively misleading line `local gate FAILED but forge CI is GREEN on the very commit it evaluated (<sha>)`. Unlike the classification work — which is airtight — this path fails open on positive-sounding non-evidence.

**Suggested fix** (local to one function + tests):

```rust
// failure | timed_out | startup_failure          -> Failure   (unchanged)
// any run for the SHA with status != "completed" -> Unknown   (CI has not judged it yet)
// completed with cancelled|action_required|stale -> Unknown   (no verdict was reached)
// otherwise Success only if >= 1 completed run concluded "success"
```

Optionally tighten further by requiring the corroborating run to be the repo's verification workflow — you already fetch `workflowName`, so gating on a configurable name (defaulting to `CI`) would close path 2 outright and make the unused field load-bearing.

Please add parse tests for: cancelled-only; `cancelled` + `success` mixed on the same SHA; and `CI` in_progress alongside a completed-success `Lines of Code` (the exact shape above) — each expecting `CiVerdict::Unknown`.

---

## Non-blocking notes (no action required)

**`killed-by-signal` is defensible — I'd keep it.** You asked for a verdict either way, so: keep it. The decisive point is that UNEVALUATED is a no-op on the halt flag, so this class can only ever *fail to establish* a halt; it can never clear one. `sh`'s 137/143 are `128 + signo` renderings that a build tool essentially never selects deliberately, while OOM and supervisor kills on a contended host produce them routinely. The residual risk — a genuine memory regression that OOMs the test binary gets absorbed as environmental — is real but narrow, and it is visible: the tick logs `[killed-by-signal]` loudly and `loom-daemon status` now surfaces the class via `not_evaluated_reason`. Worth noting there is no per-class kill switch (`LOOM_GATE_CI_CORROBORATION` only covers AC4), so an operator who disagrees has no lever short of a code change. Fine as shipped.

**`CI_PROBE_RUN_LIMIT = 30`** is about 8-10 commits of `main` history at 3-4 workflows per commit. With a 600s `buildGate.timeoutSeconds` and a fast merge cadence, the evaluated SHA can fall out of the window before the probe runs — which yields `Unknown` and keeps the halt, so it fails safe. No change needed; a comment noting the history-depth tradeoff would help the next reader.

**`run_command_with_timeout` kills only the `sh` process**, not its process group, so a timed-out `cargo test` can survive as an orphan holding the target dir lock. Pre-existing behavior, not introduced here — flagging only so it is not lost.

rjwalters:
Blocking fail-open addressed in `1362944d` — scoped to `parse_gh_run_list` + its tests (plus the two non-blocking doc/comment nits).

## The fix: an unambiguous all-clear, or nothing

`parse_gh_run_list` (`loom-daemon/src/main_health_gate.rs`) no longer treats "saw a completed run that wasn't a failure" as green. The reducer is now asymmetric in the direction the AC4 invariant requires — `Success` is the hardest verdict to reach, and anything short of an all-clear degrades to `Unknown` (which keeps the halt):

| Runs for the evaluated SHA | Verdict |
|---|---|
| any `failure` / `timed_out` / `startup_failure` | `Failure` (unchanged) |
| any run not yet `completed` (`queued`, `in_progress`, …) | `Unknown` — CI has not judged the commit yet |
| any `cancelled` / `action_required` / `stale` / unrecognized conclusion | `Unknown` — the workflow reached no verdict about the code |
| ≥ 1 genuine `success`, every other run `skipped` / `neutral` | `Success` |
| no run for the SHA / unparseable output | `Unknown` |

**Absence of failure is not success** — `Success` now requires at least one run to have actually concluded `success`, *and* no sibling run for the same commit to be outstanding or indeterminate. A `Failure` still short-circuits past any indeterminate sibling: a halt may always be **established** on this path, it just may not be **relaxed** on non-evidence.

### Your two scenarios

**1. `cancelled` reading as permanent green.** A superseded `ci.yml` run at `completed/cancelled` is now indeterminate, so the SHA is `Unknown` and the local red stands. Critically this also holds for the *mixed* shape you called out — `CI cancelled` + `Lines of Code success` on the same SHA is `Unknown`, because the bookkeeping success cannot paper over CI's missing verdict.

**2. Bookkeeping workflow vouching alone.** During the ~100s window after a push, `CI` is `in_progress` for that SHA. Any non-`completed` run makes the whole verdict `Unknown`, so `Lines of Code`'s early `success` no longer answers for the commit. Once CI finishes, the next gate tick gets a real answer — the corroboration is delayed, never fabricated.

## On gating by workflow identity

I did **not** gate on `workflowName`, and the reasoning is that the unanimity rule closes both paths *without* needing to know which workflow counts:

- **Hard-coding `CI` is brittle** in exactly the place this code runs. The daemon dispatches into consumer repos whose verification workflow is not named `CI`; there, a name gate degrades to permanent `Unknown` and silently retires AC4 for those repos.
- **A configurable name is brittle-by-default.** Whatever the default is, it is wrong for some repo, and a knob nobody sets is a knob that either fails open (no default) or fails closed silently (default `CI`).
- **Requiring every sibling to reach a verdict is strictly stronger than a name gate for both reported paths**, and it is safe under ignorance: a workflow the probe has never heard of can only ever push the verdict toward `Unknown`, never toward green. That matches the same fail-toward-safety shape you approved in `classify_failed_exit`.

`workflowName` is now load-bearing for diagnostics rather than unused: the debug log names the specific run that withheld a verdict (`forge CI verdict for <sha> is indeterminate (CI is in_progress)`), so an operator can see *why* corroboration declined instead of just that it did.

The residual gap a name gate would close and this does not: a repo where the real verification workflow legitimately does not run for a commit (`paths-ignore`) and only a bookkeeping workflow runs and succeeds. That requires CI to have genuinely declined to apply to the commit, and it is not reachable in this repo's workflow config.

## Tests

New `test_parse_gh_run_list_non_evidence_is_never_success` covers, each expecting `Unknown`: cancelled-only; `cancelled` CI + `success` bookkeeping; `in_progress` CI + completed-success bookkeeping (your exact shape); `queued` CI + completed-success bookkeeping; `action_required` and `stale` each mixed with a success; an unrecognized future conclusion mixed with a success; all-`skipped`. Plus two positive controls: a real `failure` alongside indeterminate siblings still returns `Failure`, and a genuine all-clear still returns `Success`. The pre-existing `test_parse_gh_run_list_verdicts` is unchanged and still passes.

- `cargo test -p loom-daemon --lib -- main_health_gate work_finder` → **139 passed, 0 failed** (serial).
- On the parallel run, `work_finder::tests::test_registry_dispatcher_exports_claim_ownership_marker` flaked once — it waits up to 5s for a spawned child to write its record log and completes in 0.28s in isolation. Same #3985 host-contention class; passes on re-run, not touched by this change.
- `cargo clippy -p loom-daemon --all-targets` and `cargo fmt --all --check` clean.

## Non-blocking notes you raised

- **`killed-by-signal`** — kept, per your verdict. No change.
- **`CI_PROBE_RUN_LIMIT = 30`** — added the history-depth comment noting the SHA can age out of the window on a fast merge cadence, and that this fails safe (`Unknown` → halt stands).
- **`run_command_with_timeout` not killing the process group** — pre-existing, left alone as you flagged.

`.loom/docs/daemon-reference.md` updated with the verdict table and the rationale for not hard-coding a workflow name.

rjwalters:
✅ **Approved!** (Judge pass 2 — re-review of the Doctor's `1362944d`)

The single blocker from pass 1 is closed. I verified the unanimity rule against the code rather than the summary, and it holds. The earlier approved findings (`classify_failed_exit` fallthrough, the AC3 `tick_multi` root cause, SHA exactness, `killed-by-signal`) are unchanged by this commit — it touches only `main_health_gate.rs` and the reference doc — and are not re-litigated.

## 1. Unanimity rule is implemented as described

`parse_gh_run_list` now accumulates `saw_success` and a first-seen `indeterminate` reason, and reduces in this order after the loop: `indeterminate.is_some()` → `Unknown`, else `saw_success` → `Success`, else `Unknown`. Both reported scenarios are genuinely closed:

- **`completed/cancelled` reading as permanent green.** `cancelled` falls to the catch-all `other` arm and sets `indeterminate`, so it can never reach `Success`. Critically it also holds in the *mixed* shape I called out — a `cancelled` CI alongside a `success` bookkeeping run is `Unknown`, because the post-loop `indeterminate` check runs **before** the `saw_success` check. The success cannot paper over CI's missing verdict.
- **Bookkeeping workflow vouching alone.** The `status != "completed"` branch sets `indeterminate` and `continue`s, so `CI` at `in_progress`/`queued` forces `Unknown` regardless of `Lines of Code` having already concluded `success`.

Also correct in the details: a JSON-`null` or absent `conclusion` on a completed run yields `None` from `field()`, which lands in `other` → `indeterminate`, not in a green path.

## 2. Short-circuit ordering is preserved — a halt can still be established

This was my main structural worry, since a unanimity rule can plausibly break the ability to *establish* a halt. It does not: `return CiVerdict::Failure` sits in a match arm **inside** the loop, while the `indeterminate` check is **after** the loop. A `failure` therefore returns immediately and unconditionally, discarding any indeterminate sibling already recorded. The test pins exactly this with the failure listed *last*, after an `in_progress` and a `cancelled` sibling. The asymmetry is right: a halt may always be established, it may only never be relaxed on non-evidence.

## 3. Workflow-identity reasoning — accepted on the merits

The argument that unanimity is strictly stronger than a name gate for both reported paths is correct, and the brittleness objection is the load-bearing one: hard-coding `CI` would degrade to permanent `Unknown` in consumer repos whose verification workflow is named otherwise, silently retiring AC4 exactly where the daemon dispatches. Under unanimity, an unknown workflow can only ever push the verdict toward `Unknown`, never toward green — the same fail-toward-safety shape as `classify_failed_exit`. Promoting `workflowName` to diagnostics (naming the run that withheld the verdict) is the right use of the field.

**I verified the stated residual-gap unreachability claim rather than accepting it.** `.github/workflows/ci.yml` triggers on `push: branches: [main]` with **no `paths`/`paths-ignore` filter**, so a run is created for every `main` commit; the `dorny/paths-filter@v4` inside it is a *job-level* skip within an existing run, which still yields a run-level `success` because `Detect Changes` always runs. The claim holds here. The gap is real for consumer repos Loom does not control, so I filed **#3987** to track it rather than leave it in a PR thread — not a blocker, since closing it now would require the workflow-identity coupling this PR correctly declined to introduce.

## 4. No over-correction — `Success` is empirically reachable

This was the failure mode I was most concerned about, so I checked it against reality instead of reasoning about it. Running the probe's exact query (`gh run list --branch main --limit 30 --json headSha,status,conclusion,workflowName`) and grouping by SHA:

- **8 of the 9 most recent `main` SHAs** yield `Success` under the new rule — all runs `completed`, ≥ 1 `success`, nothing indeterminate. Corroboration fires in the dominant case; AC4 is not silently retired.
- The one exception (`4d2d5270`) carries a superseded `Security Scan=completed/cancelled` next to a successful one, so it degrades to `Unknown`. That is the intended conservative direction, and it is *transient*: the gate hard-resets to `origin/main` and re-resolves the SHA every tick, so a supersession-cancelled SHA is only ever evaluated until the next tick picks up the newer head.

The one real cost is a longer post-merge `Unknown` window — previously the fast `loc.yml` flipped it green in ~30s, now it waits for full CI (~2-4 min). That is the correct trade: the old fast flip *was* the bug, and `Unknown` merely preserves the local verdict for a few more minutes rather than fabricating one.

Consumption is also correct: `corroborate_red` is reached only on a completed-and-failed local run, and `CiVerdict::Unknown => GateOutcome::red(...)` keeps the halt.

## 5. Tests and checks

`test_parse_gh_run_list_non_evidence_is_never_success` covers every case claimed — cancelled-only; cancelled + success; `in_progress` + success; `queued` + success; `action_required` and `stale` each mixed with a success; an unrecognized future conclusion mixed with a success; all-`skipped` — plus the two positive controls that matter most (failure short-circuiting past indeterminate siblings, and a genuine all-clear still reading green). The pre-existing `test_parse_gh_run_list_verdicts` is unchanged and still passes.

- `cargo test --lib main_health_gate` → **70 passed, 0 failed**
- `cargo test --lib work_finder -- --test-threads=1` → **69 passed, 0 failed**
- `cargo test --lib parse_gh_run_list` → **2 passed**; `corrobor` → **3 passed**
- `cargo clippy --all-targets -- -D warnings` → clean; `cargo fmt --check` → clean
- `gh pr checks 3986` → Rust Linting & Build **pass**, Rust Unit Tests **pass**, Detect Changes **pass**
- `mergeStateStatus` → **CLEAN**; branch is current with `origin/main`

The `CI_PROBE_RUN_LIMIT` history-depth comment is a good addition, and correctly notes the age-out case fails safe to `Unknown`.

Ready for merge. Follow-up tracked in #3987.

Issue #3939, opened 2026-07-25
work-finder: no backoff/quarantine for insta-crashing dispatches — same 3 issues re-dispatched every tick, starving the rest of the queue


rjwalters:
Observed live on v0.15.0 canary rollout: 3 vibesql issues whose sweep children die within ~2s of spawn (missing token pool / selector import failure, #3938) were **re-dispatched on every 60s work-finder tick**, occupying all 3 global concurrency slots each tick:

```
work_finder: tick — cap 3 ...; 4 workspace(s), 14 seen, 3 dispatched, 0 labeled-skip, 0 in-flight-skip, 11 deferred, 0 error(s)
(repeats every tick with the same 3 issues; 11 deferred includes healthy dispatchable work in other repos)
```

The startup watchdog / bounded mid-build recovery (#3887/#3903) handle a *hung* or *mid-build-dead* child, but an instantly-crashing child is reaped and its issue simply re-qualifies on the next poll — there is no per-issue crash counter, backoff, or quarantine. Consequences: healthy work in other managed repos is starved indefinitely by a broken workspace, and the failing repo gets hammered forever.

## Acceptance criteria
- [ ] A sweep that exits/crashes within a short threshold (e.g. <60s, no phase checkpoint written) increments a per-(repo,issue) failure counter; after N consecutive insta-crashes (e.g. 3) the issue is quarantined (skipped with a logged reason and/or labeled `loom:blocked` with a comment) instead of re-dispatched.
- [ ] Quarantine is visible in `loom-daemon status` (per-repo breakdown) and clears on operator action or a TTL.
- [ ] Deferred healthy candidates in other workspaces get slots once a workspace's dispatches are quarantined (no cross-repo starvation).

Related: #3938 (the specific crash cause here), #3903 (bounded mid-build recovery), #3930 (per-repo isolation — this is the dispatch-slot version of the same principle).

Found during v0.15.0 canary rollout (multi-repo auto mode live test).

Issue #3894, opened 2026-07-23
Autonomous daemon degrades to random token selection without a fresh .ranking (skip exhausted accounts)


rjwalters:
## Context

The daemon token selector (`loom_tools.tokens.select`) is 3-tier: **ranking** (`.loom/tokens/.ranking`, refreshed <10 min) → allowlist → **random**. This session, `.loom/tokens/.ranking` did not exist, so the selector fell straight through to **random** and repeatedly picked the exhausted account `robb-2amlogic` (7d utilization 1.00), whose sweeps then hung at startup (spawn header logged, no worktree, 0 HTTPS connections, ~0% CPU). This caused ~6 startup hangs requiring manual cancel + re-dispatch.

Running `loom-tokens check --ranking` once (writing `.ranking` with `robb-2amlogic|exhausted`) flipped selection to `mode=ranked` and dispatches immediately started landing on healthy accounts. But `.ranking` is only valid for <10 min, so without a periodic probe it goes stale and the selector silently degrades back to random — re-introducing the hangs.

## The gap
For **autonomous** daemon mode (epic #3809 work-finder, which dispatches in bursts), a stale/absent `.ranking` means the daemon steadily dispatches into exhausted accounts and wedges issues. The probe cron (`.loom/scripts/probe-tokens.sh --ranking`) exists but is not part of autonomous-mode setup, and nothing self-refreshes the ranking.

## Acceptance criteria
- [ ] Autonomous mode keeps `.ranking` fresh — either (a) the daemon/work-finder refreshes it (or skips exhausted accounts) on a bounded cadence, or (b) autonomous-mode setup wires `probe-tokens.sh --ranking` on a <10-min cron and documents it as required.
- [ ] The selector skips `exhausted`/`blocked` accounts even when `.ranking` is stale (fail-safe: treat a >10-min-old ranking as advisory rather than discarding it entirely and going fully random into known-bad tokens).
- [ ] Document in the autonomous-mode operability guide that a fresh ranking is a prerequisite (tie into #3813 operability / #3891 status view showing ranking freshness).

## Notes
Complements the startup-hang watchdog (#3892), which self-heals a single hang but re-dispatches through the same selector — so without ranking freshness it can retry into another exhausted account and then GiveUp.

rjwalters:
[...]

PR #3994, opened 2026-07-27
fix(tokens): apply exhausted-threshold test on 429 path, not just 2xx

### LINKED ISSUE #3988: Work-finder retry storm: token-exhausted sweeps re-dispatch every 60s with no backoff — 107 of 189 dispatches died on credits, draining 6/7 accounts
## Summary

The autonomous work finder re-dispatches an issue on **every** poll tick when its sweep dies from token exhaustion, with **no backoff and no exhaustion-aware suppression**. Because `spawn-claude.sh` selects a fresh token per spawn, each retry burns quota from a *different* account — so one unbuildable issue systematically drains the entire pool.

This is not a symptom of the exhausted pool. It is a **primary cause** of it.

## Evidence (rjwalters/vibesql, 2026-07-25 → 07-26)

Across all `.loom/logs/sweep-issue-*.log`:

```
189  total dispatches
107  died with "out of usage credits" / "hit your limit"   (57%)
```

Single worst offender, issue #6193 — **88 dispatches, 8 merges**:

```
64  "out of usage credits"
 3  "hit your limit"
 5  "Background tasks still running after 600s; terminating"
 8  actual merges
```

The retry cadence is a metronome locked to the poll interval:

```
loom-daemon dispatch: 2026-07-25T21:53:34.933651
loom-daemon dispatch: 2026-07-25T21:54:34.355481
loom-daemon dispatch: 2026-07-25T21:55:34.894301
loom-daemon dispatch: 2026-07-25T21:56:34.554043
loom-daemon dispatch: 2026-07-25T21:57:34.657322
```

Exactly 60s apart — `autonomous.workFinder.intervalSecs`. Every tick, another account.

Live pool state after this ran overnight (`loom-tokens check`, 11:21Z):

| account | 7d util | status |
|---|---|---|
| robb-2amlogic | 0.01 | available |
| agent1-2amlogic | 1.00 | rate_limited |
| agent4-2amlogic | 1.00 | rate_limited |
| robb-integratedplasmonics | 1.00 | rate_limited |
| agent2-2amlogic | 1.00 | rate_limited |
| agent3-2amlogic | 1.00 | rate_limited |
| rjwalters-gmail | 0.96 | exhausted |

Six of seven accounts at **100% of their weekly quota**, resets staggered to Jul 31. The dynamic concurrency cap collapsed to 2, and tier-0 loom work has been capacity-starved all day.

## Why existing guards do not catch it

- **Insta-crash quarantine (#3939/#3960)** keys on rapid crash-on-startup. These sweeps are not insta-crashes: the child authenticates, runs real work, and *then* hits the account ceiling. It never trips the threshold.
- **Bad-token tracking** (`.bad_tokens`, `TOKEN_EXHAUSTED` in `classify-error.sh`) marks the *token* bad, but nothing tells the **work finder** to stop dispatching the *issue*. So the finder simply rotates to the next healthy account and burns that one too.
- **The token-capacity term in the concurrency cap** (`healthy × perTokenConcurrency`) bounds *concurrency*, not *retry rate*. One slot re-dispatching every 60s for hours is fully within cap.

- [ ] A dispatch whose child exits with `TOKEN_EXHAUSTED` / `TOKEN_EXPIRED` triggers **exponential backoff** for that issue (e.g. 1m → 5m → 30m → capped), rather than immediate re-dispatch on the next tick
- [ ] Consecutive token-exhaustion deaths across *different* accounts short-circuit dispatch entirely — if N successive attempts died on credits, the pool is the problem and dispatching more cannot help. Halt and log loudly rather than continuing to rotate
- [ ] Exit classification is plumbed from the sweep child back to the work finder so it can distinguish "this issue failed" from "this account is out" — today the finder appears blind to why the child died
- [ ] Per-issue dispatch-attempt accounting with a ceiling; 88 dispatches for 8 merges on one issue should have tripped something
- [ ] Consider whether a family/partial-increment issue (returned to `loom:issue` by `merge-pr.sh` after each `Part of #N` merge) needs a **cooldown** before it is eligible again — see the related claim race below

## Related claim race (smaller, same area)

Partial-increment family issues re-enter the queue the instant `merge-pr.sh` resets them, producing a self-race:

```
14:30:40Z  PR #6286 merged → #6172 reset to loom:issue
14:31:54Z  a sweep claims it → loom:building
14:31:55Z  work_finder dispatches ANOTHER sweep for it
           → pre-flight sees the 1s-old claim → skips
```

3 wasted dispatches on #6172, 1 on #6174. Minor next to the retry storm, but the same missing idea: no cooldown between an issue becoming eligible and being re-dispatched.

## Related

- #3978 — concurrency cap has no CPU term (this is the token-axis analogue)
- #3939/#3960 — insta-crash quarantine, the guard that *should* have been the natural home for this
- #3969 — daemon-side token ranking refresh

rjwalters:
## Summary

Root cause fix for the retry storm described in #3988. Per the issue's follow-up comment, this is a classification bug, not a missing-backoff problem: a weekly-exhausted account frequently answers the health probe with **429**, and `EXHAUSTED_THRESHOLD` (`7d_utilization >= 0.95`) was only ever checked on the **2xx** path (`check.py:370-373`), never the 429 path (`check.py:339-348`). Such accounts were mislabelled `rate_limited` forever, and the selector treats `rate_limited` as fully eligible (`select.py:242`, `select.py:286`) — so the work finder kept rotating dispatches into dead accounts on every poll tick.

## Changes

- **`loom-tools/src/loom_tools/tokens/check.py`**: extracted a shared `_status_from_utilization()` helper and applied it on **both** the 2xx and 429 response branches — a 429 with `7d_utilization >= 0.95` is now classified `exhausted`, not `rate_limited`. Restores the documented CLAUDE.md contract (`exhausted = 7d_utilization >= 0.95`) for the accounts that most need it.
- **`loom-tools/src/loom_tools/tokens/select.py`**: defense in depth suggested in the issue comment — both the ranked tier (`_try_ranking`) and the stale-ranking advisory-exclusion set (`_stale_ranking_exclusions`) now *prefer* excluding `rate_limited` accounts while a healthier (`available`) account exists in the pool, falling back to including them only when the pool would otherwise be empty (the existing exclusion fail-safe in `select_token` already retries without exclusions in that case, so this can only ever prefer a healthier account — never hard-fail).

## Scope note

Per the issue's guidance, this focuses on the root-cause classification hole rather than the broader backoff/quarantine mechanisms in the issue's own acceptance-criteria checklist — those are explicitly described as symptom-side mitigations the classification fix makes largely unnecessary. The insta-crash quarantine (#3939/#3960) and per-issue dispatch accounting are unaffected by this PR; if the classification fix alone doesn't fully resolve the retry-storm symptom in practice, that would be a natural follow-up rather than scope creep here.

## Testing

```
PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -v
```
283 passed, including new regression tests:
- `test_check.py`: `test_429_promoted_to_exhausted_when_7d_high`, `test_429_at_threshold_promoted_to_exhausted`, `test_429_without_utilization_headers_stays_rate_limited`
- `test_select.py`: `test_ranking_prefers_available_over_rate_limited`, `test_ranking_falls_back_to_rate_limited_when_pool_otherwise_empty`, `test_ranking_rate_limited_and_exhausted_mixed_prefers_available`, `test_stale_ranking_excludes_rate_limited_from_random`, `test_stale_ranking_all_rate_limited_falls_back_rather_than_hard_fail`

Closes #3988

### DIFF
--- a/loom-tools/src/loom_tools/tokens/check.py
+++ b/loom-tools/src/loom_tools/tokens/check.py
@@ -21,8 +21,13 @@ succeeds on plain API keys).
 Status assignment:
 
 * ``available`` — utilizations < 95 percent
-* ``exhausted`` — 7d_utilization >= 0.95
-* ``rate_limited`` — current 429 (transient, distinct from exhausted)
+* ``exhausted`` — 7d_utilization >= 0.95 (checked on **both** the 2xx and
+  429 response paths, issue #3988 — a weekly-exhausted account very often
+  answers the probe with 429 rather than 200, and previously only the 2xx
+  path applied this test, so such accounts were mislabelled
+  ``rate_limited`` and stayed eligible for selection forever)
+* ``rate_limited`` — current 429 with 7d_utilization below the exhausted
+  threshold (a genuine transient rate limit, distinct from exhaustion)
 * ``blocked`` — 401 auth failure or token listed in ``.bad_tokens``
 
 The CLI command writes ``.loom/tokens/.ranking`` atomically when
@@ -287,6 +292,18 @@ def _build_headers(token: str) -> dict[str, str]:
     return base
 
 
+def _status_from_utilization(s7d_util: float | None, *, default: str) -> str:
+    """Return ``"exhausted"`` when *s7d_util* clears the threshold, else *default*.
+
+    Shared by both the 2xx and 429 probe-response branches (issue #3988) so
+    the ``EXHAUSTED_THRESHOLD`` test is applied uniformly regardless of which
+    HTTP status the probe happened to receive.
+    """
+    if s7d_util is not None and s7d_util >= EXHAUSTED_THRESHOLD:
+        return "exhausted"
+    return default
+
+
 def probe_account(
     name: str,
     token: str,
@@ -337,11 +354,18 @@ def probe_account(
         return AccountResult(name=name, status="blocked", error="auth_401")
 
     if code == 429:
-        # Even a 429 may include rate-limit headers; capture them.
+        # Even a 429 may include rate-limit headers; capture them. A weekly-
+        # exhausted account frequently answers with 429 (not 200), so the
+        # exhausted-threshold test must apply here too (issue #3988) — else
+        # such accounts are mislabelled "rate_limited" forever and the
+        # selector keeps rotating dispatches into a dead account.
         parsed = parse_rate_limit_headers(resp.headers)
+        status = _status_from_utilization(
+            parsed["7d_utilization"], default="rate_limited"
+        )
         return AccountResult(
             name=name,
-            status="rate_limited",
+            status=status,
             s5h_utilization=parsed["5h_utilization"],
             s7d_utilization=parsed["7d_utilization"],
             s7d_reset=parsed["7d_reset"],
@@ -366,11 +390,7 @@ def probe_account(
 
     parsed = parse_rate_limit_headers(resp.headers)
     s7d_util = parsed["7d_utilization"]
-
-    if s7d_util is not None and s7d_util >= EXHAUSTED_THRESHOLD:
-        status = "exhausted"
-    else:
-        status = "available"
+    status = _status_from_utilization(s7d_util, default="available")
 
     return AccountResult(
         name=name,
--- a/loom-tools/src/loom_tools/tokens/select.py
+++ b/loom-tools/src/loom_tools/tokens/select.py
@@ -231,32 +231,50 @@ def _try_ranking(
     accounts, in rotating order. The window spans every eligible account by
     default; ``_resolve_spread_top_n`` optionally caps it to the top-N
     most-available (``N == 1`` restores the greedy first-eligible behavior).
+
+    Defense in depth (issue #3988): ``rate_limited`` is a weaker signal than
+    ``available`` (a probe that hit 429 without clearing the exhausted
+    threshold, i.e. a genuine transient rate limit *or* a status the
+    classifier hasn't yet promoted). Prefer excluding ``rate_limited``
+    entries entirely while the pool still has at least one non-rate_limited
+    account; only fall back to including them when they are all that is
+    left, so a fully rate-limited pool can still dispatch rather than
+    hard-failing.
     """
     age = _file_age_seconds(ranking_file)
     if age is None or age >= _RANKING_FRESH_SECONDS:
         return None
 
     cap = _resolve_spread_top_n(workspace_path)
-    eligible: list[SelectedToken] = []
-    for name, status in _read_ranking(ranking_file):
-        if status in ("exhausted", "blocked"):
-            continue
-        token_file = tokens_dir / f"{name}.token"
-        if not token_file.is_file():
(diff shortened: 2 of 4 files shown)


rjwalters:
## Review: Approved

Verified the diff matches the PR description precisely:

**`check.py`**: `_status_from_utilization()` is now the single source of truth for the exhausted-threshold test, and it's applied identically on both the 2xx (line ~393) and 429 (line ~363) branches. Confirmed with the new regression tests that a 429 with `7d_utilization >= 0.95` (including the exact-threshold boundary case) is now classified `exhausted`, and a 429 with no rate-limit headers correctly stays `rate_limited` (nothing to promote on). This directly closes the misclassification hole described in #3988.

**`select.py`**: Traced the fallback logic in `_try_ranking` and `_stale_ranking_exclusions` to confirm the "never end up with zero eligible accounts" invariant holds:
- `_try_ranking`'s two-pass `_collect()` (skip `rate_limited` first, fall back to including it only if that pass is empty) can only ever *prefer* a healthier account — it never returns `None` due to `rate_limited` alone when at least one such account exists.
- `_stale_ranking_exclusions` now folds `rate_limited` into the same advisory-exclusion set as `exhausted`/`blocked`, but reuses the pre-existing `select_token()` fail-safe (lines ~461–471) that already retries without exclusions if the excluded set would empty the pool — so this is a pure extension of the #3894 mechanism, not new fail-safe logic.

**Tests**: Ran `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -v` — all 283 passed, including the 8 new regression tests (3 in `test_check.py`, 5 in `test_select.py`) covering the promotion boundary, the no-headers stay-rate_limited case, the available-preferred-over-rate_limited rotation, the mixed exhausted/rate_limited/available case, and both the fresh- and stale-ranking all-rate_limited fallback paths.

Also ran the full non-network `loom-tools` suite (excluding `tests/integration` and the slow spawn-integration test) as a broader regression check: 1930 passed, 2 pre-existing failures (`test_logging.py::test_log_output_visible_non_interactively`, `test_forge_cli.py::TestMainHelp::test_version_flag`) — both reproduce identically on `main` at the pre-PR commit (98466c2c) via plain `PYTHONPATH=` invocation (a `uv run` package-metadata/subprocess-env artifact, unrelated to this diff), so no regression.

CI: `gh pr checks` shows only path-filtered jobs, correctly skipped (no Rust/mcp/scripts changes) plus the passing `Detect Changes` gate. `mergeStateStatus` is `CLEAN`.

Minor non-blocking nit for a future follow-up: the root `CLAUDE.md` "Status assignment" one-liner (`rate_limited (current 429)`) hasn't been updated to reflect that a 429 can now also resolve to `exhausted` — the module docstring in `check.py` covers this correctly, so this is just a high-level-doc staleness nit, not a functional issue.

Code quality is solid and consistent with the rest of the token-pool module (docstrings explain the "why" with issue references, matches existing style). Approving.

PR #3996, opened 2026-07-27
fix(daemon): load daemon as a launchd LaunchAgent on macOS instead of nohup

### LINKED ISSUE #3972: loom-daemon-start.sh nohup leaves daemon in the launching session's Mach bootstrap — gh/git break system-wide when that session dies
## Context

Incident 2026-07-26: the daemon was started at 21:48 via `loom-daemon-start.sh` from inside a Claude Code session. That session crashed at 02:49. From the very next work-finder tick (02:50:21), **every** `gh` call in the daemon's process tree failed with `tls: failed to verify certificate: x509: OSStatus -26276` and `git fetch` failed with `No user exists for uid 501` — while the same commands worked perfectly from any fresh shell. The daemon ran blind for ~35 minutes: work-finder saw 0 issues (4 errors/tick), the main-health gate went RED on environmental failures, and in-flight sweep children couldn't reach the forge.

Root cause: `loom-daemon-start.sh:221` detaches with `nohup "$DAEMON_BIN" ... &`. nohup survives SIGHUP but the process **stays in the launching session's Mach bootstrap namespace**. When the session's context was torn down, XPC lookups to `trustd` (certificate verification — hence the x509 OSStatus from Go's darwin verifier in `gh`) and `opendirectoryd` (`getpwuid` — hence "No user exists for uid 501") started failing for the daemon and every child it spawned. Only macOS is affected; systemd Linux does not have this failure mode.

Validated fix: relaunching the identical binary as a launchd agent (`launchctl bootstrap gui/501` with a LaunchAgent plist) immediately restored gh/git — first tick after migration: `13 seen, 3 dispatched, 0 error(s)`. A working plist now exists at `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist` (note: its `PATH` must include `~/.cargo/bin` or the build-gate's cargo invocation fails with exit 127).

- [ ] On macOS, `loom-daemon-start.sh` launches the daemon outside the calling session's bootstrap namespace — e.g. generates/installs a LaunchAgent plist and `launchctl bootstrap`s it (mirroring the working plist above), with `loom-daemon-stop.sh` doing the corresponding `launchctl bootout`
- [ ] The generated environment preserves the current flag semantics (FLAGS-OFF default, `--work-finder` / `--health-gate` / `--from-config`) and a PATH covering `gh`, `git`, `cargo`, `python3`
- [ ] Linux behavior unchanged (nohup is fine under systemd user sessions; optionally document `systemd-run --user` as the equivalent hardening)
- [ ] Docs: daemon-reference §Operability explains the macOS session-bootstrap hazard and why "start it from a terminal that might die" is unsafe

rjwalters:
## Summary

Fixes the macOS Mach-bootstrap-namespace hazard identified in the 2026-07-26 incident: `loom-daemon-start.sh` previously detached the daemon with a plain `nohup "$DAEMON_BIN" &`, which keeps the process wired into the *launching session's* Mach bootstrap namespace. When that session (e.g. a Claude Code session) died, XPC lookups to `trustd` (cert verification -> `gh` TLS errors) and `opendirectoryd` (`getpwuid` -> "No user exists for uid N" from `git`) started failing for the daemon and every child it spawned -- with no crash and no obvious log signal. The daemon ran blind for ~35 minutes during the incident.

The validated fix (already hand-applied on the operator's machine as `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist`) is to load the daemon as a `gui/<uid>` LaunchAgent instead.

## Changes

- `defaults/scripts/cli/loom-daemon-start.sh`:
  - On Darwin, generates a LaunchAgent plist (`~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist` by default, override with `LOOM_LAUNCHD_LABEL`) and loads it via `launchctl bootstrap` + `kickstart -k`, mirroring the hand-written plist that validated the fix during the incident (`RunAtLoad=true`, `KeepAlive=false`).
  - The plist's `PATH` is the current `PATH` plus a fallback set (`~/.local/bin`, `~/.cargo/bin`, Homebrew, standard bin dirs) so `gh`/`git`/`cargo`/`python3` resolve inside launchd's minimal environment.
  - Every already-exported `LOOM_*` / `GH_TOKEN` / `GITEA_TOKEN` / `FORGE_TOKEN` var is forwarded verbatim, so the FLAGS-OFF / `--work-finder` / `--health-gate` / `--from-config` semantics are preserved exactly.
  - New `--no-launchd` / `LOOM_DAEMON_LAUNCHD=0` escape hatch forces the legacy nohup path even on Darwin.
  - New `--print-plist` prints the exact plist XML an invocation would install, with zero side effects (no `launchctl` call, no file write) -- used for test coverage of the plist-rendering logic.
  - Linux is unaffected: unchanged plain `nohup` path.
- `defaults/scripts/cli/loom-daemon-stop.sh`: on macOS, additionally `launchctl bootout`s the LaunchAgent job definition once the process is confirmed dead, so `RunAtLoad=true` doesn't silently relaunch a daemon the operator explicitly stopped at the next login/boot. Falls back to a launchd-sourced pid lookup when the PID file is stale.
- `.loom/docs/daemon-reference.md`: new "macOS session-bootstrap hazard (#3972)" subsection under Operability documenting the incident, root cause, and fix.
- New tests: `test-loom-daemon-launchd-plist.sh` (23 tests, pure `--print-plist` rendering coverage -- FLAGS-OFF default, `--work-finder`, `--from-config` leaves vars unset, label override, PATH fallback dirs, RunAtLoad/KeepAlive, flags-file exclusion), `test-loom-daemon-stop.sh` (7 tests, using a random nonexistent launchd label so the suite never touches a real machine's LaunchAgents).
- Updated `test-loom-daemon-start.sh` / `test-loom-daemon-update.sh`: their existing background-start assertions now force the legacy nohup path (`--no-launchd` / `LOOM_DAEMON_LAUNCHD=0`) since they exercise flag-persistence/rebuild-restart behavior, not launchd itself, and must never mutate a real machine's LaunchAgents.

## Test plan

- [x] `bash defaults/scripts/tests/test-loom-daemon-launchd-plist.sh` -- 23/23 passed
- [x] `bash defaults/scripts/tests/test-loom-daemon-stop.sh` -- 7/7 passed
- [x] `bash defaults/scripts/tests/test-loom-daemon-start.sh` -- 10/10 passed (regression)
- [x] `bash defaults/scripts/tests/test-loom-daemon-update.sh` -- 17/17 passed (regression)
- [x] `shellcheck --severity=warning` clean on all of `defaults/scripts/`
- [x] Verified (md5 before/after) that the whole suite run never mutates the real `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist` already present on the dev machine from the incident's manual fix
- [ ] Manual end-to-end `launchctl bootstrap`/`kickstart`/`bootout` lifecycle on a real macOS login session (not exercised by the automated suite by design -- requires a GUI login session)

Closes #3972

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1311,6 +1311,9 @@ process:
   default),
 - locates the `loom-daemon` binary (`LOOM_DAEMON_BIN` → `PATH` → `target/{release,debug}`),
 - runs the **advisory** host-sleep check (`check-host-sleep.sh`, #3350) — never blocks the start,
+- **on macOS, backgrounds the daemon as a `gui/<uid>` LaunchAgent** instead of a
+  plain `nohup … &` (#3972 — see "macOS session-bootstrap hazard" below); on
+  Linux it stays a plain nohup background job,
 - backgrounds the daemon and writes a PID file at `.loom/.daemon.pid` (gitignored),
 - refuses a second start when the PID file points at a live process, and surfaces
   the daemon's own **singleton-guard** refusal (#3806) — if the backgrounded
@@ -1319,7 +1322,8 @@ process:
 
 `loom-daemon-stop.sh` sends **SIGTERM** (not just Ctrl-C/SIGINT — the daemon now
 handles both, #3813), waits `LOOM_DAEMON_STOP_GRACE_SECS` (default 10s), then
-escalates to SIGKILL.
+escalates to SIGKILL. On macOS it additionally `launchctl bootout`s the
+LaunchAgent job definition once the process is confirmed dead (see below).
 
 **Shutdown decision — sweeps survive, they are not drained.** A clean daemon stop
 removes the Unix socket and exits, but **does not cancel in-flight `/loom:sweep`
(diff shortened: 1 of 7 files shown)


rjwalters:
✅ **Approved!**

Reviewed the diff in the `issue-3972` worktree and independently verified the PR's own test-plan claims rather than just trusting them.

**Correctness / flag semantics**
- `loom-daemon-start.sh`'s launchd path correctly preserves the FLAGS-OFF default: `LOOM_WORK_FINDER`/`LOOM_MAIN_HEALTH_GATE` are resolved into the environment *before* `render_launchd_plist` walks `env | grep -E '^(LOOM_...'`, so the plist always reflects the same precedence chain (already-exported > `--work-finder`/`--health-gate` > `--from-config` leaves both unset > default 0/0) as the pre-existing nohup path. `--print-plist` is a pure, side-effect-free render (no `launchctl`, no file write) and is correctly excluded from the persisted `.daemon.flags` (verified via test 7 in the new plist suite).
- `--no-launchd` / `LOOM_DAEMON_LAUNCHD=0` escape hatch is wired correctly, and the Linux path (`nohup "$DAEMON_BIN" &`) is byte-for-byte unchanged — confirmed via diff, gated purely behind `IS_DARWIN`.
- `loom-daemon-stop.sh`'s `launchd_bootout_if_loaded` is correctly idempotent and only fires once the pid is confirmed dead; PID resolution order (pid file → launchd lookup → pgrep fallback) is sound and the "survive, don't drain" in-flight-sweep contract is preserved.

**The real-LaunchAgent-safety claim — verified independently, not just taken on faith:**
```
before: 8f06d2197ddfcc71feb39c344c94a6d4  (md5 of ~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist)
... ran all 4 test suites (launchd-plist, stop, start, update) ...
after:  8f06d2197ddfcc71feb39c344c94a6d4
PLIST UNCHANGED
```
This machine has the real incident-fix plist loaded at the same default label (`com.rjwalters.loom-daemon`), so this is a genuine test of the safety claim, not a no-op check. It holds because: (1) `test-loom-daemon-launchd-plist.sh` only ever calls `--print-plist` (pure string render, no `launchctl`); (2) `test-loom-daemon-stop.sh` pins `LOOM_LAUNCHD_LABEL` to a random `com.example.loom-daemon-stop-test-$$-$RANDOM` label so `launchctl print` never resolves the real job; (3) `test-loom-daemon-start.sh` / `test-loom-daemon-update.sh` now force `--no-launchd` / `LOOM_DAEMON_LAUNCHD=0` respectively so their real (non-mocked) start/stop exercise stays on the legacy nohup path.

## Test Execution
1. `bash defaults/scripts/tests/test-loom-daemon-launchd-plist.sh` — ✅ Executed: 23/23 passed
2. `bash defaults/scripts/tests/test-loom-daemon-stop.sh` — ✅ Executed: 7/7 passed
3. `bash defaults/scripts/tests/test-loom-daemon-start.sh` — ✅ Executed: 10/10 passed (regression)
4. `bash defaults/scripts/tests/test-loom-daemon-update.sh` — ✅ Executed: 17/17 passed (regression)
5. `shellcheck --severity=warning` on all 6 touched scripts — ✅ Executed: clean, 0 findings
6. md5 before/after check on the real LaunchAgent plist across the full suite run — ✅ Executed: confirmed byte-identical (see above)
7. Manual end-to-end `launchctl bootstrap`/`kickstart`/`bootout` on a real GUI login session — ⚠️ Skipped: requires manual observation on a live login session, as the PR itself notes

**CI**: all checks pass (`Detect Changes`, `Shellcheck`, `Installer Integration Tests`; Rust jobs skipped as expected for a shell-only diff). `mergeStateStatus` is `CLEAN`.

Docs update in `.loom/docs/daemon-reference.md` accurately describes the incident (symptom → root cause → fix) and cross-references the escape hatches. Nice work.

PR #4000, opened 2026-07-27
feat(daemon): durable operator watches on issue/PR terminal state (#3971)

### LINKED ISSUE #3971: Operator watches die with the session: daemon-side watch/notify for sweep and issue terminal state
## Context

At the end of the 2026-07-26 canary session, the operator armed a 4-hour background poll watching vibesql#6193 and loom#3964 for terminal state. The operator session then crashed, and the watch died with it (background tasks are children of the Claude Code session process) — the terminal-state report was silently lost until the next session reconstructed it from logs.

Root cause is harness behavior (session-scoped background tasks), but the durable fix belongs loom-side: the daemon is the long-lived process that already knows sweep lifecycle state and publishes `sweep.issue.{N}.*` / `sweep.global.completed` events. What's missing is a way for an operator to register interest that survives their session.

Design notes / options (for Curator to scope):
- `loom-daemon watch <issue-or-sweep>...` CLI that subscribes over IPC and blocks until terminal state — still dies with its shell, but composable with `nohup`/a file sink;
- daemon-side persistent watch registry: watches recorded in the daemon, results appended to a log/journal file (e.g. `.loom/logs/watch-results.log`) an operator or next session can read;
- optional exec-on-terminal-state hook (run a command / write a marker file when a watched item completes or blocks).

Constraint: the v0.10.0 event taxonomy is frozen — new topics need a follow-up issue; prefer consuming existing topics + forge polling for cross-repo issues over minting new ones.

- [ ] An operator can register a watch on issue(s)/sweep(s) that survives their Claude Code session dying
- [ ] Terminal state (closed/merged/blocked/crashed) is durably recorded somewhere a later session can trivially find
- [ ] Works cross-repo for daemon-managed workspaces (the #6193 case was a vibesql issue watched from the loom operator session)

rjwalters:
## Summary

Operator watches on sweep/issue terminal state died with the operator's Claude Code session (session-scoped background polls are children of the session process), silently losing the terminal-state report until the next session reconstructed it from logs. This adds a **daemon-side persistent watch registry** so an operator's registered interest survives their session dying — and a daemon restart.

## What changed

- **New module `loom-daemon/src/watch_registry.rs`**
  - Watches persisted machine-level to `~/.loom/watches.json` (override `LOOM_WATCHES_PATH`) — survives the registering session **and** a daemon restart.
  - A **default-on monitor loop** polls each watch's terminal state and, on a terminal state (`closed`/`merged`/`blocked`) or expiry, **durably appends** a JSON line to `~/.loom/logs/watch-results.log` (override `LOOM_WATCH_RESULTS_LOG`) — a file a later session can trivially `tail` — then drops the watch.
  - **Forge polling** (`gh issue view` / `gh pr view --json state,labels`) rather than the event bus: works cross-repo for a repo the daemon may never have swept, and mints **no new event topic** (the v0.10.0 taxonomy stays frozen). Address a target cross-repo with `repo` (forge slug `owner/name`) or `workspace_root` — the same addressing pattern as `dispatch_sweep` / `list_sweeps`.
- **IPC**: `RegisterWatch` / `ListWatches` / `RemoveWatch` requests + responses (`types.rs`, handlers in `ipc.rs`).
- **MCP tools**: `register_watch` / `list_watches` / `remove_watch` (`mcp-loom/src/tools/sweeps.ts`).
- **CLI**: `loom-daemon watch add|list|remove` over the daemon socket (analogue of the `dispatch` subcommand).
- **Config**: `autonomous.watchMonitor.{enabled,intervalSecs,expirySecs}` (precedence env > config > default; default-on, zero forge calls until a watch is registered — mirrors the token-ranking refresh loop's rationale).
- **Docs**: new `daemon-reference.md` section + config table rows; mcp-loom README tool table.

## Acceptance criteria

- [x] An operator can register a watch that survives their Claude Code session dying — the watch lives in a file the long-lived daemon owns.
- [x] Terminal state (closed/merged/blocked/expired) is durably recorded in `~/.loom/logs/watch-results.log`, trivially findable by a later session.
- [x] Works cross-repo for daemon-managed workspaces (the #6193 case: a vibesql issue watched from the loom operator session) via `repo` / `workspace_root`.
- [x] No new event-bus topics (frozen taxonomy respected).

## Testing

- 25 new Rust tests: `watch_registry` unit (classify/persist/dedup/tick/expiry/config precedence/`run_one_tick` integration) + an IPC roundtrip (`test_watch_registry_ipc_roundtrip`). Full suite: 896 passed.
- `cargo clippy --all-targets` clean; `mcp-loom` `npm run build` (tsc + esbuild) passes.
- Live end-to-end smoke over a real Unix socket: `watch add` (with dedup) → `list` → `remove` all correct, including the cross-repo `rjwalters/vibesql` case.

Closes #3971

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1036,6 +1036,11 @@ concurrency ceiling 5" and share it with the team:
     "mainHealthGate": {
       "enabled": true
     },
+    "watchMonitor": {
+      "enabled": true,
+      "intervalSecs": 120,
+      "expirySecs": 86400
+    },
     "dispatchStaggerMs": 2000,
     "watchdog": {
       "enabled": true,
@@ -1068,6 +1073,9 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.workFinder.quarantine.instaCrashSecs` | `LOOM_WORK_FINDER_QUARANTINE_INSTA_CRASH_SECS` | `60` | Checkpoint-less death within this window of dispatch counts as an insta-crash. Zero/invalid → default |
 | `autonomous.perTokenConcurrency` | `LOOM_PER_TOKEN_CONCURRENCY` | `2` | Concurrent sweeps **per healthy token** in the cap (#3947). Zero/invalid → default; clamped to a floor of 1 |
 | `autonomous.mainHealthGate.enabled` | `LOOM_MAIN_HEALTH_GATE` | `false` | Gate loop on/off |
+| `autonomous.watchMonitor.enabled` | `LOOM_WATCH_MONITOR` | `true` | Durable operator-watch monitor loop (#3971). Default-on; no dispatch side effect, zero forge calls until a watch is registered |
+| `autonomous.watchMonitor.intervalSecs` | `LOOM_WATCH_MONITOR_INTERVAL_SECS` | `120` | Watch poll cadence. Zero/invalid → default |
+| `autonomous.watchMonitor.expirySecs` | `LOOM_WATCH_MONITOR_EXPIRY_SECS` | `86400` | Give-up window for an unresolved watch; `0` disables expiry |
 | `autonomous.dispatchStaggerMs` | `LOOM_SWEEP_DISPATCH_STAGGER_MS` | `2000` | Min gap between consecutive child spawns (#3887). `0` disables |
 | `autonomous.watchdog.enabled` | `LOOM_SWEEP_WATCHDOG` | `true` | Startup watchdog on/off (#3887). Also the master switch for the tick — mid-build-death (#3895) + review-stall (#3910) run in the same task |
 | `autonomous.watchdog.timeoutSecs` | `LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` | `120` | No-progress window before auto-restart |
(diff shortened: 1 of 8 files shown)


rjwalters:
## Judge review — changes requested (one substantive fix)

Reviewed the full diff, not just the self-report. This is high-quality work: clean separation (pure \`tick\`, trait-based \`WatchProbe\` for testability), tolerant best-effort I/O, atomic temp+rename writes, thorough tests (25 new, all green locally + the IPC roundtrip), clippy clean, and CI's clean-room Rust Unit Tests (896) pass. I independently confirmed the three headline claims:

- **No new event-bus topics** — the monitor polls the forge (\`gh issue view\` / \`gh pr view --json state,labels\`) rather than subscribing to the bus; the v0.10.0 taxonomy stays frozen. Verified.
- **Cross-repo works via forge polling** — \`WatchSpec.repo\` (\`--repo owner/name\`) or \`workspace_root\` (\`current_dir\`), no dependence on the in-memory bus that only knows sweeps this daemon dispatched. Verified.
- **Durable across daemon restart** — persisted to \`~/.loom/watches.json\`, reloaded each tick; resolutions appended to \`~/.loom/logs/watch-results.log\`. Verified.

(For the record: the local \`integration_singleton_guard\` failures are environmental — a real daemon is running on this review box and collides with that test's socket/session handling; failures were inconsistent across runs and clean-room CI passes. Not attributable to this PR.)

### Required change: guard the concurrent read-modify-write on \`watches.json\`

Unlike \`workspace_registry\` (which this is modeled on), \`watches.json\` has **two independent concurrent writers**: the background \`watch_monitor\` loop *and* the IPC \`RegisterWatch\`/\`RemoveWatch\` handlers (each IPC connection is its own \`tokio::spawn\`). Nothing serializes them, so there is a lost-update window that defeats the feature's core durability promise:

1. Monitor \`run_one_tick\`: \`load()\` → registry = \`[A]\`
2. Monitor begins probing A (shells out to \`gh\` — seconds, up to the 30s timeout)
3. IPC \`RegisterWatch B\`: \`load()\` → \`[A]\`, add B, \`save([A, B])\` → returns success to the operator
4. Monitor's probe resolves A, \`tick\` drops A, \`save([])\` → **B is silently and permanently lost**

The operator got a success response for a watch that then vanished with no trace — exactly the silent-loss failure this issue exists to prevent. \`run_one_tick\` never re-loads/merges before saving, so there's no recovery on the next tick.

Secondary facet of the same problem: \`save()\` builds its temp path as \`json.tmp.{}\` from \`std::process::id()\`. All in-process writers (monitor task + IPC handler tasks) share the **same PID**, so concurrent saves collide on the *same* temp file — the atomic-rename guarantee doesn't actually hold under intra-daemon concurrency.

**Fix options** (any one):
- Wrap the load→modify→save critical section for the watches file in the repo's established \`mkdir\`-based lock (per CLAUDE.md — the pattern already used for \`.bad_tokens\`/\`.allowlist\`; POSIX-atomic and macOS-safe), and
- give the temp file a unique-per-writer suffix (e.g. append a random/uuid component alongside the PID), or
- make the monitor task the single writer (funnel IPC mutations through a channel the monitor drains), which sidesteps both facets.

Please also add a regression test for the concurrent register-during-resolving-tick case.

Everything else looks ready to merge once this is addressed.

rjwalters:
Addressed the required change (concurrent read-modify-write on `watches.json`) in 786161ca.

**Fix (all three facets):**
- **Serialized the critical section.** Added an mkdir-based lock (`with_watches_lock`) — POSIX-atomic and macOS-safe, matching the CLAUDE.md convention for `.bad_tokens`/`.allowlist` — guarding the whole load→modify→save in both `handle_register_watch` and `handle_remove_watch`.
- **Closed the monitor lost-update window.** `run_one_tick` now probes lock-free against a snapshot (never holding the lock across a `gh` probe, which can take up to the 30s timeout), then under the lock re-loads a **fresh** copy from disk and drops **only the ids it actually resolved** — so an IPC `RegisterWatch` that lands mid-tick is merged in rather than clobbered by a stale snapshot. Your exact scenario (register B while resolving A) no longer loses B.
- **Fixed the temp-file collision.** The atomic-rename temp path now gets a unique-per-writer suffix (`uuid` alongside the PID), since all in-process writers share `std::process::id()`.

**Regression test:** `concurrent_register_during_resolving_tick_is_not_lost` reproduces the register-during-resolving-tick race via a probe that performs the same guarded register the IPC handler does while resolving A. Verified it **fails** against the old naive `save(&registry)` (`left: []`, B lost) and **passes** with the reload-merge.

**Checks:** `cargo build --release` clean, `cargo clippy --all-targets` clean, full lib suite green (897 passed incl. the new test). The only integration flakes seen were tmux terminal-spawn timeouts under concurrent machine load (a separate worktree running `cargo test` at the same time); they pass in isolation and are untouched by this change (diff is only `ipc.rs` watch handlers + `watch_registry.rs`).

rjwalters:
Judge re-review of the race fix (commit 786161ca) — confirms the previously-requested change closes the reported gap.

**1. Lock correctly serializes all three writers**
- `handle_register_watch` and `handle_remove_watch` (`loom-daemon/src/ipc.rs`) now do their full load→modify→save under `with_watches_lock` (mkdir-based, matching the `.bad_tokens`/`.allowlist` convention).
- `run_one_tick` (`loom-daemon/src/watch_registry.rs`) still probes the forge lock-free against a snapshot (correctly — never hold the lock across a multi-second `gh` call), but the persist step now re-loads a **fresh** copy under the lock and `retain`s only the ids it actually resolved, rather than blindly overwriting with the stale in-memory snapshot. A concurrent `RegisterWatch` that lands mid-probe is therefore always present in that fresh reload and survives.
- The monitor loop itself is single-ticker (`tokio::spawn` with a sequential `ticker.tick().await` → `spawn_blocking`), so there's no intra-monitor concurrency to worry about — only monitor-vs-IPC, which the lock now covers.
- Temp-file collision is fixed too: `save()` now suffixes the atomic-rename temp path with `std::process::id()` **and** a fresh `uuid::Uuid::new_v4()`, so concurrent in-process writers (which share a PID) no longer collide on the temp name.

**2. Regression test is real, not a no-op**
`concurrent_register_during_resolving_tick_is_not_lost` (`watch_registry.rs`) drives a `RegisteringProbe` that, mid-probe of watch A (which resolves `Closed`), performs the exact lock→load→add→save sequence the IPC handler does to register watch B. Asserts B survives the tick. This is a faithful simulation of the exact race described (would fail against the old naive "just save the in-memory snapshot" tick, since B wouldn't be in that snapshot) — confirmed logically correct by reading `run_one_tick`'s reload-merge-drop-resolved-ids logic, which is exactly what's needed to make this test pass.

**3. Build / clippy / tests** — all run against the actual PR worktree (`.loom/worktrees/issue-3971`, HEAD `786161ca`):
- `cargo clippy --all-targets` — clean, no warnings.
- `cargo test` (workspace) — 897 unit/lib tests pass including all new `watch_registry::tests::*`. The 3 `tests/integration_basic.rs` failures I hit locally (`test_ping_pong`/`test_multiple_clients`/`test_send_input`, then a different one on rerun) are unrelated to this PR: that file has zero diff in this PR, has a documented history of flakiness under load ("stabilize flaky integration test" #2712, "Fix integration test flakiness from daemon restoring orphan sessions" #1953), and this machine currently has several other concurrent `cargo build`/`cargo test` processes from other agents competing for CPU (confirmed via `ps aux`) — a single-threaded rerun still took 82s for 9 socket-timeout-sensitive tests. GitHub's own CI (`Rust Unit Tests` check) passed cleanly, which is the authoritative signal here.
- `mcp-loom`: `npm run build` (tsc + esbuild) — clean.

**4. No other regressions** — `types.rs`/`ipc.rs`/CLI/MCP tool additions are mechanical and consistent with existing request/response and tool-registration patterns; docs (`daemon-reference.md`) accurately describe the shipped config surface.

CI: all required checks pass (`Rust Linting & Build Check`, `Rust Unit Tests`, `MCP Server Build`, `Detect Changes`), `mergeStateStatus` is `CLEAN`.

Approved.

PR #4001, opened 2026-07-27
fix(daemon,mcp): bound ListSweeps read-path forge I/O so a wedged gh can't hang list_sweeps ~15min (#3973)

### LINKED ISSUE #3973: mcp-loom list_sweeps hung >15min against a wedged daemon — fail-fast bridge timeout not applied
## Context

During the 2026-07-26 incident (daemon alive but its gh/XPC calls wedged — see the session-bootstrap issue), an operator `mcp__loom__list_sweeps` call hung past the 120s tool ceiling, was backgrounded by the harness, and only returned ~15 minutes later — at the exact moment the daemon received SIGTERM (it then returned "No sweeps tracked."). Meanwhile the `loom-daemon status` CLI answered instantly over the same IPC socket with a full registry, so the daemon was accepting connections; the ListSweeps handler specifically blocked (plausibly on registry/liveness reconciliation performing forge or subprocess calls under a lock — #3893 wired `reap_liveness` into the ListSweeps read path).

The fail-fast MCP bridge work (shipped during the canary rollout to kill the 1800s hang class) did not bound this case: the bridge connected fine and then waited indefinitely on a response that never came.

- [ ] The MCP bridge applies a per-request response deadline (e.g. 10–30s) after connect, returning a typed timeout error naming the request kind, not an indefinite wait
- [ ] ListSweeps' read path cannot block unboundedly on external processes while holding the registry response — liveness reconciliation on the read path gets its own timeout or moves off the request path
- [ ] Regression test: a handler that sleeps forever produces a bridge timeout error, not a hang

rjwalters:
Closes #3973

## Problem

During the 2026-07-26 incident the daemon was alive — its `loom-daemon status` CLI answered instantly over the same IPC socket — but its `gh`/XPC calls were wedged. An operator `mcp__loom__list_sweeps` call hung past the 120s tool ceiling, was backgrounded by the harness, and only returned ~15 minutes later (when the daemon got SIGTERM). The fail-fast MCP bridge (#3945) connected fine and then waited on a response that never came.

## Root cause

#3893 wired `reap_liveness` into the `ListSweeps` / `GetSweepStatus` read path. `reap_liveness` → `reap_once` best-effort shells out to `gh` for forge-label reconciliation (`restore_label_to_ready`, the `loom:blocked` probe `issue_has_blocked_label`, and the quarantine label flips) via `Command::output()` **with no timeout**. A wedged `gh` blocked the handler under the registry mutex indefinitely. `loom-daemon status` answered because it takes a different, non-reaping path.

## Fix

**Daemon (root cause, `loom-daemon/src/sweep_registry.rs`):** every reaper `gh` invocation now runs through a new `output_with_timeout` helper that kills the child if it exceeds a per-call bound — `REAP_GH_TIMEOUT` (default 5s), overridable via `LOOM_REAP_GH_TIMEOUT_SECS`. On timeout the call is treated as the same best-effort failure any other `gh` error already is, so the in-memory liveness transition always completes and the read path can never block unboundedly on an external process.

**MCP bridge (`mcp-loom/src/tools/sweeps.ts`):** `list_sweeps` / `get_sweep_status` now pass an explicit bounded 30s response deadline (`Math.max` with any `LOOM_DAEMON_IPC_TIMEOUT_MS` override) — generous headroom over a bounded reap, but still far below the ~1800s hang class. The transport's typed timeout error already names the request kind.

## Acceptance criteria

- [x] The MCP bridge applies a per-request response deadline after connect, returning a typed timeout error naming the request kind, not an indefinite wait — `list_sweeps`/`get_sweep_status` get an explicit bounded 30s deadline; the error already names the request type.
- [x] ListSweeps' read path cannot block unboundedly on external processes — every reaper `gh` call is individually bounded and killed on timeout; the liveness transition still completes.
- [x] Regression test: a handler that sleeps forever produces a bridge timeout error, not a hang — `verify-daemon-timeout.mjs` gains an explicit `ListSweeps` hang case; Rust tests cover the bounded helper + an end-to-end read-path test with a `gh` that sleeps 60s.

## Testing

- Rust: `output_with_timeout` (fast-complete + kill-on-hang), `reap_gh_timeout_honors_env_override`, and `read_path_reap_is_bounded_when_gh_wedges` (proves `reap_liveness` returns promptly and still transitions the child to terminal when `gh` hangs) — all pass in isolation. `cargo build`, `cargo clippy --lib --tests`, and `cargo fmt` are clean.
- TS: `npm run typecheck` passes; `node scripts/verify-daemon-timeout.mjs` passes including the new ListSweeps checks; `npm run build` succeeds.

> Note: the full `cargo test --lib sweep_registry` module could not be run to green locally because this shared machine was saturated by concurrent agents (load avg ~8), which flaked several timing-sensitive process-spawn tests **including pre-existing, unmodified ones** (e.g. `read_path_reaps_exited_child_out_of_running_promptly`, the `fixture_registry` `dispatch_*` tests that never reach the modified `gh` code) — confirming those failures are environmental, not regressions. CI will run the suite in a clean environment.

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -360,6 +360,26 @@ pub const DEFAULT_WATCHDOG_INTERVAL_SECS: u64 = 30;
 /// escalating to SIGKILL, when it auto-cancels for re-dispatch.
 const WATCHDOG_CANCEL_GRACE: Duration = Duration::from_secs(3);
 
+/// Per-call ceiling for a best-effort `gh` subprocess invoked from the reaper
+/// (Issue #3973).
+///
+/// The reaper's forge-label reconciliation (`restore_label_to_ready`,
+/// `issue_has_blocked_label`, the quarantine label flips) runs on the
+/// `ListSweeps` / `GetSweepStatus` **read path** via [`SweepRegistry::reap_liveness`].
+/// During the 2026-07-26 incident a wedged `gh`/XPC blocked that read under the
+/// registry mutex indefinitely, so an operator `list_sweeps` hung ~15 minutes.
+/// Every reaper `gh` call is bounded to this window: on timeout the child is
+/// killed and the call is treated as the same best-effort failure any other
+/// `gh` error already is, so the in-memory liveness transition always completes.
+/// Overridable via [`REAP_GH_TIMEOUT_ENV`] for operability.
+const REAP_GH_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Env var overriding [`REAP_GH_TIMEOUT`] (whole seconds; zero/invalid ignored).
+pub const REAP_GH_TIMEOUT_ENV: &str = "LOOM_REAP_GH_TIMEOUT_SECS";
+
+/// Poll cadence for [`output_with_timeout`] while waiting on a reaper `gh` call.
+const REAP_GH_POLL_INTERVAL: Duration = Duration::from_millis(20);
+
 /// Env var toggling the review-phase stall watchdog (Issue #3910).
 /// `0`/`false`/`no`/`off` disables; `1`/`true`/`yes`/`on` forces on. Overrides
 /// config. Distinct from `LOOM_SWEEP_WATCHDOG` (the #3887 startup watchdog) so a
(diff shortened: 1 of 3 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the #3973 acceptance criteria:

1. **MCP bridge per-request deadline** — `list_sweeps`/`get_sweep_status` in `mcp-loom/src/tools/sweeps.ts` now pass an explicit `Math.max(READ_PATH_TIMEOUT_MS, resolveDaemonIpcTimeoutMs())` (30s floor, env-overridable) into `sendDaemonRequest`, which already produces a typed timeout error naming the request kind. Verified the `sendDaemonRequest(request, timeoutMs?)` signature/semantics match this usage.
2. **ListSweeps read path bounded** — root cause was `reap_liveness`/`reap_once` shelling out to `gh` with no timeout under the registry mutex (from #3893). The new `output_with_timeout` helper (poll + `try_wait` + kill-on-deadline) now bounds every reaper `gh` invocation (`restore_label_to_ready`, `issue_has_blocked_label`, quarantine label/comment flips) to `REAP_GH_TIMEOUT` (default 5s, `LOOM_REAP_GH_TIMEOUT_SECS` override), and every call site treats a kill the same as any other best-effort `gh` failure so the in-memory liveness transition always completes.
3. **Regression tests** — `output_with_timeout_kills_a_hung_command` (unit), `read_path_reap_is_bounded_when_gh_wedges` (end-to-end: fake `gh` sleeps 60s, asserts `reap_liveness` returns in <15s and the child still reaches a terminal state), and a new `ListSweeps` hang case in `verify-daemon-timeout.mjs` asserting the bridge rejects (not hangs) with an error naming `ListSweeps`.

**Verification performed:**
- `cargo build --release` (loom-daemon) — clean
- `cargo clippy --lib --tests` — clean
- `cargo test --lib sweep_registry` (targeted: `output_with_timeout*`, `reap_gh_timeout_honors_env_override`, `read_path_reap_is_bounded_when_gh_wedges`, plus the broader `sweep_registry` substring match — 125 tests) — all pass
- `npm run typecheck` / `npm run build` (mcp-loom) — clean
- `node scripts/verify-daemon-timeout.mjs` — all checks pass, including the new ListSweeps hang assertions
- `gh pr checks 4001` — all required checks pass; `mergeStateStatus` is `CLEAN`

**Non-blocking observations** (not required for this issue's scope, noting for awareness):
- `handle_request` (which calls `reap_liveness`) runs synchronously inline on the tokio worker thread handling that connection (not via `spawn_blocking`), so a burst of several dead sweeps needing reconciliation could chain a few bounded ~5s `gh` calls on that worker thread. This is a strict improvement over the prior unbounded hang and matches the issue's accepted mitigation ("gets its own timeout"), but moving the reap off the async request path entirely (the issue's other suggested option) would be a nice future hardening — not blocking here.
- `LOOM_REAP_GH_TIMEOUT_SECS` isn't yet listed in `.loom/docs/daemon-reference.md`'s config-knob tables alongside the other `LOOM_*` overrides; minor doc-completeness gap, easy follow-up.

Great incident-driven fix — thorough root-cause tracing, tight bound, and tests that actually reproduce the wedge scenario rather than just asserting the API shape.

PR #4013, opened 2026-07-27
feat(tokens): import live account tokens from claude-monitor's store (#4006)

### LINKED ISSUE #4006: loom-tokens: add a command to import account tokens from claude-monitor into the pool
## Context
loom already consumes claude-monitor's `~/.claude-monitor/ranking.json` for token health, but claude-monitor also **holds the account OAuth tokens** (in its usage.db). There is currently no loom-side command to materialize those into loom's token pool (`~/.loom/tokens/*.token` + `index.json` with `source: monitor`).

Standing up a second daemon on a new machine this session required **hand-copying** the pool from another host, because there was no laptop-side command to populate the pool from the locally-running monitor. Keeping the bridge on the loom side (rather than a claude-monitor export feature) keeps loom the owner of its own pool format.

- [ ] A command (e.g. `loom-tokens import-from-monitor`, or `loom-tokens bootstrap --from-monitor`) reads claude-monitor's local account store and writes `~/.loom/tokens/*.token` + `index.json` (`source: monitor`, email/fingerprint metadata).
- [ ] Works for the shared machine-level pool and per-repo pools (honors `LOOM_SHARED_TOKENS_DIR`).
- [ ] Idempotent refresh; prunes/updates entries for accounts the monitor no longer reports.
- [ ] Documented as the standard way to populate a new host's pool.

rjwalters:
Closes #4006

## Why

`~/.claude-monitor/accounts.env` is a **snapshot**. claude-monitor keeps the **live** credentials in `usage.db` → `oauth_credentials` and refreshes them as accounts are re-authenticated. The two drift, and the drift is silent and total.

This was hit for real on the primary host today. Every account in the pool returned:

```
401 {"type":"authentication_error","message":"OAuth access token has been revoked."}
```

`loom-tokens check` reported all 7 accounts `blocked`, so the daemon's dynamic cap collapsed and dispatch stopped:

```
Dynamic concurrency cap: 0
  = min(healthy 0 × per-token 5 = 0, disk headroom 122, configured max 10)
In-flight sweeps: 0
```

The accounts *had* been rolled — in claude-monitor, which wrote fresh tokens to `usage.db` and never touched `accounts.env`. **`bootstrap --force` could not fix it**: it faithfully rewrote the same revoked tokens, because the snapshot itself was what had gone stale. Confirmed by fingerprinting — every key in `accounts.env` was byte-identical to the revoked token on disk, while the DB held different (working) ones.

Until now there was no Loom-side command to pull from the live store; the standing workaround was hand-copying a pool between hosts.

## What

`loom-tokens import-from-monitor` reads the live store directly.

```bash
loom-tokens import-from-monitor                  # into <repo>/.loom/tokens
loom-tokens import-from-monitor --shared         # into the machine-level pool (#3938)
loom-tokens import-from-monitor --force          # apply ROLLED tokens
loom-tokens import-from-monitor --dry-run        # preview
loom-tokens import-from-monitor --prune          # drop accounts no longer reported
```

Design notes worth reviewing:

- **Read-only** on `usage.db` (`mode=ro`, bounded timeout). The store belongs to another tool — never written, never migrated.
- **`expires_at` is deliberately not a filter.** Observed rows carry timestamps months in the past while the token still authenticates, so filtering on it would discard working credentials. Health is established by probing (`loom-tokens check`), which is authoritative.
- **Only `is_active = 1`** — claude-monitor deactivates rows for retired accounts, and importing those would repopulate the pool with accounts the operator removed.
- Emails resolve via the `accounts` join, falling back to parsing the credential label (`"<email>'s Organization"`). A label with no `@` and no joined row is skipped with a warning rather than guessed at.
- Duplicate emails resolve to the **highest row id** (rows are append-ordered, so that's the current credential).
- **`--force` is what applies a roll.** Every rolled token legitimately differs from disk, so without it each is reported as drift and left alone — so a hand-pinned token is never silently clobbered. Unapplied drift exits `2` so a script can detect "pool still stale".
- **`--prune` globs `*.token` only**, so pool state (`.ranking`, `.bad_tokens`, `.failure_counts`, `.allowlist`) is never disturbed. Off by default.
- Filenames use `bootstrap`'s existing derivation, so an account keeps one identity across both paths and re-importing overwrites in place.

### Shared writer, not a second one

Rather than hand-roll a parallel writer, the materialization loop was extracted from `bootstrap_tokens()` into `materialize_accounts()` + `write_index()`, and both paths call it. Atomic write-then-rename, `0600`/`0700` modes, fingerprint drift detection, and the `index.json` schema are therefore identical **by construction** — a second writer would have been free to drift from this one on exactly the security-relevant details.

`index.json` records `source: monitor-db`, distinct from the `monitor` snapshot, so the manifest shows which surface a token came from. As always it carries fingerprints only, never secret material.

## Testing

- **27 new tests** in `loom-tools/tests/tokens/test_monitor_db.py` against a synthetic `usage.db`: active/inactive filtering, both email-resolution paths, unresolvable-label and empty-token skips, duplicate-email precedence, past-`expires_at`, missing DB, older schema without `oauth_credentials`, DB-not-modified, idempotency, dry-run, drift-without-force, force-applies-roll, prune scoping (including state-file safety), file/dir modes, no-secrets-in-manifest, and the CLI surface (shared pool, exit 2 on drift, missing store, `--shared` when disabled).
- **Full suite green: 316 passed** (`loom-tools/tests/tokens/`), including the 92 pre-existing `test_bootstrap.py` tests that pin the refactored writer's behavior.
- **Verified end-to-end against the real store** on the affected host: imported all 8 active accounts, fingerprints matched the live DB tokens, modes were `0600`/`0700`, no secrets in `index.json`, and a second run reported `written=0 unchanged=8`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -459,6 +459,42 @@ To *exclude* an inherited account from one repo, pin the subset you want with `l
 
 > **Secrets**: `~/.claude-monitor/accounts.env`, the opt-in `~/.loom/accounts.env`, and the repo-local `.loom/accounts.env` all hold raw OAuth keys. The repo-local file and `.loom/tokens/` are gitignored (installer- and `loom-daemon init`–managed); keep any home-level master `0600` and outside any repo. A repo can rely entirely on the claude-monitor master with no local account file at all.
 
+#### Importing live tokens from claude-monitor (#4006)
+
+`accounts.env` is a **snapshot** — a file someone wrote by hand at some point. claude-monitor keeps the **live** credentials in its SQLite store (`~/.claude-monitor/usage.db` → `oauth_credentials`) and refreshes them as accounts are re-authenticated. The two drift, and the drift is silent and total:
+
+```text
+401 {"type":"authentication_error","message":"OAuth access token has been revoked."}
+```
+
+When that happens to every account at once, `loom-tokens check` reports all accounts `blocked`, the daemon's dynamic concurrency cap collapses to `min(healthy 0 × per-token N, …) = 0`, and dispatch stops entirely. Crucially **`bootstrap --force` does not fix it** — it faithfully rewrites the same revoked tokens, because the snapshot itself is what went stale.
+
+`loom-tokens import-from-monitor` reads the live store directly and is **the standard way to populate a new host's pool** (it replaces hand-copying a pool between machines):
+
+```bash
+loom-tokens import-from-monitor                  # into <repo>/.loom/tokens
+loom-tokens import-from-monitor --shared         # into the machine-level pool (#3938)
+loom-tokens import-from-monitor --force          # apply ROLLED tokens (see below)
+loom-tokens import-from-monitor --dry-run        # preview
+loom-tokens import-from-monitor --prune          # drop accounts the monitor no longer reports
+```
+
+**`--force` is what applies a token roll.** Every rolled token legitimately differs from what is on disk, so without `--force` each one is reported as drift and left alone — deliberately, so a hand-pinned token is never silently clobbered. The command exits `2` when drift was found and not applied, so a script can detect "pool is still stale". After importing, refresh the ranking so the daemon sees the recovered capacity:
+
+```bash
+loom-tokens import-from-monitor --force && loom-tokens check --ranking
+```
+
+Behavior notes:
+
+- **Read-only** on `usage.db` (opened `mode=ro`) — the store belongs to claude-monitor; Loom never writes or migrates it.
+- Only `is_active = 1` rows are imported; `expires_at` is **not** used as a filter (observed rows carry stale timestamps while still authenticating — health comes from `loom-tokens check`).
+- Token filenames use the same derivation as `bootstrap` (`robb@2amlogic.com` → `robb-2amlogic.token`), so an account keeps one identity across both paths and re-importing overwrites in place.
+- Idempotent: unchanged tokens are left untouched. `index.json` records `source: monitor-db` (distinct from the `monitor` snapshot) and, as always, fingerprints only — never secret material.
+- `--prune` removes only `*.token` files; pool state (`.ranking`, `.bad_tokens`, `.failure_counts`, `.allowlist`) is never touched.
+- The importer takes **claude-monitor as authoritative for pool membership**, so it imports every active account — including any that `accounts.env` omitted. Use `loom-tokens pin` to restrict which accounts the selector may actually pick.
+- Absent claude-monitor, an absent `usage.db`, or an older schema without `oauth_credentials` all exit `1` with a message naming the path tried.
+
 #### Account health probe + ranking
 
 Once bootstrapped, `loom-tokens check` probes each account for current rate-limit headers and (optionally) writes a JSON ranking that the spawn-time selector can consume:
@@ -563,6 +599,8 @@ For Pro/Max plans, Loom supports rotating between multiple Claude Code OAuth tok
    ```
    The claude-monitor, repo-local, and (opt-in) home sources are **merged by email**, with the higher-precedence source overriding/adding (see "Account sources: claude-monitor-first + per-repo" above). Keep any home-level master `0600` and outside any repo.
 2. Run `loom-tokens bootstrap` to materialize the merged set into per-account `.token` files in `.loom/tokens/` (mode 0600, parent dir 0700). See issues #3234, #3695.
+
+   **If claude-monitor runs on this host, prefer `loom-tokens import-from-monitor`** — it reads claude-monitor's live credential store instead of the `accounts.env` snapshot, so a new host needs no account file of its own and a token roll is picked up automatically (add `--force` to apply rolled tokens). See "Importing live tokens from claude-monitor" above.
 3. Spawn agents through `.loom/scripts/spawn-claude.sh` instead of invoking `claude` directly. The wrapper selects a token using a 3-tier algorithm (ranking → allowlist → random), exports `CLAUDE_CODE_OAUTH_TOKEN`, then `exec`s `claude` (or pass `--use-wrapper` to layer on top of `claude-wrapper.sh` for retry behavior).
 
 ### Selection algorithm (`loom_tools.tokens.select`)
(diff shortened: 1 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the five areas called out on the request. This is a well-scoped, security-conscious change that fixes a real outage mode (`bootstrap --force` provably cannot recover a rolled pool, because the snapshot is what went stale).

## 1. Security (raw OAuth secrets)

No leak path found.

- **Logs**: the only strings derived from a DB row are `label` (`log_warning` on unresolvable email, `logger.debug` on empty token) and the 8-char fingerprint. `access_token` is never interpolated into a log record, and `_print_monitor_import` prints name/email/disposition only.
- **`index.json`**: rows come exclusively from `materialize_accounts()`, which sets `key_fingerprint` via `fingerprint()`. Pinned by `test_import_writes_tokens_and_manifest`, which asserts the raw token strings are absent from the serialized manifest.
- **Error messages**: `MonitorDbUnavailable` carries the path + `sqlite3` text; the duplicate-filename `ValueError` carries two emails. Neither can carry key material.
- **`--json`** (`MonitorImportResult.to_dict`) exposes only paths and email/name/file/source.
- **Modes**: `0600` file / `0700` dir, asserted directly in `test_token_and_directory_modes`. Write is `write_text` → `chmod` → `os.replace` onto the final name, so a reader never observes a torn token file, and a crash leaves a `*.token.tmp` rather than a truncated `*.token`. (`Path.with_suffix(".token" + ".tmp")` resolves correctly for these stems.)

## 2. bootstrap.py refactor

Verified faithful by diffing the moved block line-by-line against the deleted one — it is a pure lift, not a rewrite. Specifically preserved:

- The `existing_fp is None` / `== new_fp` / else ladder, including `force` turning `unchanged` into `written`.
- The **drift** branch's distinct manifest row: `key_fingerprint` = the *on-disk* fp plus `drift: True` and `env_fingerprint` = the source fp — the easiest detail to lose in an extraction, and it survived intact.
- The unreadable-existing-file fallback (`log_warning` → treat as new → rewrite).
- `INDEX_VERSION`, `generated_at`, key order, `indent=2`, `sort_keys=False`, trailing newline, atomic `index.json` rename.

Two harmless mechanical differences, both no-ops: the pre-write `mkdir` inside `write_index` is now `index_path.parent.mkdir` rather than `tokens_dir.mkdir` (identical, since `index_path = tokens_dir / "index.json"`), and `result.*` are `extend`ed from the outcome rather than appended in-loop. The 92 pre-existing `test_bootstrap.py` tests pin the behavior and pass unchanged.

## 3. `--prune` safety

Cannot touch pool state. `_prune_stale_tokens` globs `*.token`; none of `.ranking`, `.bad_tokens`, `.failure_counts`, `.allowlist`, or `index.json` carries that suffix (note `.bad_tokens` ends `_tokens`, not `.token`). Confirmed by `test_prune_never_touches_pool_state_files`. Also correct on the details that matter: `keep` is built from the **active** account set, so a drifted-but-not-forced account keeps its file; `dry_run` reports without unlinking; an `OSError` pops the entry so the report never claims a deletion that did not happen; and it is off by default (`test_prune_is_off_by_default`).

## 4. Read-only DB contract

Holds. `sqlite3.connect("file:…?mode=ro", uri=True)` cannot write, migrate, or take a write lock, the connection is closed in a `finally`, and the 5s timeout means a lock contended by claude-monitor surfaces as an error instead of wedging a dispatch. `test_database_is_not_modified` asserts byte-identity across a read. I also checked the real store on this host: it is `journal_mode=wal` with 19 rows, and a `mode=ro` open + query against it succeeds without disturbing the file.

## 5. Email resolution / duplicate precedence

Correct. `LEFT JOIN` keeps credential rows whose `accounts` row is missing; the joined email wins, falling back to `_email_from_label`. The regex is anchored and non-greedy, so `"robb@x.com's Organization"` → `robb@x.com`, and anything without an `@` yields `None` → skipped with a warning rather than guessed at (right call — a guessed email means a wrong filename means a silently mis-identified account). Precedence: `ORDER BY c.id` with `by_email[email.lower()] = …` means the highest row id wins, matching the stated append-order rationale; the `.lower()` key also collapses case-variant duplicates that would otherwise derive the same filename. The true-collision case (`a.jones@` vs `ajones@`) raises before any write rather than letting one clobber the other.

Also good: `expires_at` deliberately not used as a filter (documented with the observed-stale-timestamp rationale, and pinned by a test), and filenames reuse `derive_token_filename` so identities don't fork between the two import paths.

## Test Execution

- `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -q` — ✅ **316 passed**.
- Full build-gate Python stage (`uv run pytest tests/ -q --ignore=tests/integration --ignore=tests/tokens/test_agent_spawn_integration.py`) — ✅ **1987 passed, 1 skipped** (9m17s), exit 0.
- `bash .loom/scripts/build-gate.sh` in full — ⏭️ its `cargo test --workspace` stage is a cold from-scratch build in this worktree and exceeded the review time bound. Not a concern for this PR: the diff is Python + Markdown only, and CI's Rust/installer/MCP jobs are correctly path-filtered to `skipping`.
- CI: no failing or pending checks; `mergeStateStatus` `CLEAN` / `MERGEABLE`.

## Non-blocking notes

- **`monitor_db.py:163`** — `uri = f"file:{path}?mode=ro"` interpolates the path without URI-encoding, so a home directory containing `?` or `#` would truncate the path or the query. Vanishingly rare, and the failure is a clean `MonitorDbUnavailable`, not corruption. If it ever bites: `f"{path.as_uri()}?mode=ro"`.
- **`monitor_db.py:183`** — a read-only open of a WAL store with a *hot* leftover `-wal` and no `-shm` fails `SQLITE_READONLY_CANTINIT`, which lands in the `except` whose hint reads "may predate the credential store". Self-recovering (running claude-monitor clears it) and still a clean typed error naming the path.
- **`cli.py:_print_monitor_import`** is typed `result: "object"` and reads fields through defensive `getattr` defaults, though the caller always passes a `MonitorImportResult`. Typing it concretely would let a mistyped field name fail loudly instead of printing an empty column. Style only.
- Filed **#4021** (`loom:triage`) to add `import-from-monitor` to the two spots in `.loom/docs/daemon-reference.md` that still name `bootstrap` as the only provisioning lever — the managed-repo pool recipe and the capacity add-capacity advisory. Docs-only, out of scope here.

Issue #3953, opened 2026-07-25
Orphaned-claim recovery has no liveness source after a daemon restart — stale loom:building claims accumulate across managed repos


rjwalters:
Observed twice on the canary host: sweeps die (rate-limit kills, print-mode ceiling #3943, daemon restart) leaving issues stuck at `loom:building`, and `loom-recover-orphans --recover` refuses to reclaim because it finds **no authoritative liveness source** (no spawn-loop-state.json, no reachable daemon registry entry — the restarted daemon's in-memory registry is empty, no `.loom/locks/issue-<N>/`). Fail-safe refusal is correct (#3651), but in the multi-repo daemon world restarts are routine and the operator ends up hand-flipping labels (done for vibesql #6199/#6193/#6191; **8 more stale claims sit at #6170–#6190 right now**).

## Fix direction
Give the reclaim tool an authoritative liveness source that survives daemon restarts:
- The daemon already writes per-sweep records (sweep id, PID, issue, repo) — persist the registry (or a minimal `{repo, issue, pid, started}` journal) to disk (`~/.loom/sweeps.json`), reaped on read with dead-PID pruning (same pattern as sweep-run-registry.sh peers).
- `loom-recover-orphans` (and the daemon's own startup reconciliation) consult it: a `loom:building` issue whose recorded PID is dead (or which has **no** record at all and whose `updatedAt` exceeds LOOM_STALE_BUILDING_HOURS) is reclaimable.
- Bonus: daemon startup runs this reconciliation per managed workspace automatically, so restarts self-heal claims.

## Acceptance criteria
- [ ] Sweep liveness journal persisted machine-level, dead-PID pruned.
- [ ] `loom-recover-orphans --recover` reclaims a claim whose recorded PID is dead; still refuses when the record shows a live PID.
- [ ] Daemon startup reconciles stale claims across all registered workspaces (bounded, logged).
- [ ] The 8 stale vibesql claims (#6170–#6190) become recoverable by the tool instead of manual label edits.

Found during v0.15.0 canary rollout.

PR #4023, opened 2026-07-27
fix(daemon,tools): decide before pruning dead sweep-journal entries; never-silent staleness skips

### LINKED ISSUE #3975: loom-recover-orphans missed 2 of 3 dead-PID loom:building claims (journal-pruned entries not treated as orphans)
## Context

2026-07-26, vibesql workspace: three sweep children were dead while their issues still carried `loom:building` — #6172 (child exited earlier), #6170 and #6173 (children SIGTERMed during daemon recovery). `loom-recover-orphans --recover` found and recovered **only #6172** (classified `untracked_building`); #6170 and #6173 were not even listed, and had to be released by hand (`gh issue edit … --remove-label loom:building --add-label loom:issue`).

All three had dead PIDs and no linked PR. The machine-level liveness journal `~/.loom/sweeps.json` (#3966) prunes dead-PID entries on every read/write, so by the time recover-orphans runs, a dead sweep's journal entry is gone — the tool apparently distinguished #6172 from #6170/#6173 on some other signal (checkpoint presence, worktree state, or claim age?) and silently skipped two genuine orphans. A recovery tool that silently skips real orphans reintroduces the exact stuck-claim class #3966 was built to close.

- [ ] Reproduce: two+ issues labeled `loom:building`, dead child PIDs, no journal entries, no linked open PR → `loom-recover-orphans` must list and recover all of them
- [ ] Whatever filter excluded #6170/#6173 (checkpoint file, worktree-with-commits, recency window…) is either removed or reported per-issue ("skipped #N because …") — never silent
- [ ] The daemon's startup claim-reconciliation pass and recover-orphans agree on orphan criteria (document the shared rule in daemon-reference)

rjwalters:
## Summary

`loom-recover-orphans --recover` was missing genuinely-dead `loom:building` claims that had a dead recorded PID in the machine-level sweep journal (`~/.loom/sweeps.json`, #3953) but no linked PR. Root cause: `claim_reconciliation::forge::reconcile_workspace` (the daemon's startup reconciliation pass) called `sweep_journal::prune_dead` **before** calling `plan()`/`decide()` -- which erased the exact dead-PID evidence the `DeadPid` branch needs to fire its unconditional, immediate reclaim. With the entry gone, every claim silently fell through to the much longer "no journal record" grace window (default 4h), even for a sweep that had *just* died. In the reported incident, two claims (#6170/#6173, SIGTERMed during a daemon restart) sat un-reclaimed for hours while `loom-recover-orphans --recover` found and fixed only one unrelated stale claim (#6172) whose label happened to already be older than the 4h threshold.

## Changes

- **`loom-daemon/src/claim_reconciliation.rs`**: `reconcile_workspace` now decides against the raw, unpruned journal first. Pruning still happens, but only afterward via the existing per-reclaimed-issue `remove_sweep` cleanup (dead entries for issues this pass didn't touch are pruned lazily by the next `upsert` elsewhere, so the journal file never accumulates an unbounded graveyard). Added a regression test (`reconcile_workspace_reclaims_dead_pid_entry_even_when_label_is_fresh`) using a fake `gh` script that proves a dead-PID entry is reclaimed immediately even when the `loom:building` label is only seconds old.
- **`loom-tools/src/loom_tools/orphan_recovery.py`**: every staleness-gated skip in `check_untracked_building` is now recorded in a new `OrphanRecoveryResult.watched` list (issue, reason, label age, threshold) and always surfaced in both human and `--json` output -- previously this was logged only under `--verbose`, so a default invocation gave zero visible trace that a claim was seen and excluded (acceptance criterion: "never silent").
- **`.loom/docs/daemon-reference.md`**: new section documenting the shared decision rule between the Rust startup pass and the Python tool (they read the same journal and apply the same 3-way rule), the #3975 bug and fix, and the one intentional asymmetry that remains (Rust's `DeadPid` reclaim is unconditional; Python still applies the short `label_grace_period` as defense-in-depth since it can run ad hoc, not just once per daemon start).
- Added/updated Python tests covering both new watched-entry scenarios (short label-grace skip, long no-record-stale skip) plus JSON/human formatting.

## Test plan

- [x] `cargo test --lib claim_reconciliation` (12 tests, including the new regression test) -- all pass
- [x] `cargo test --lib sweep_journal` -- unaffected, all pass
- [x] `cargo fmt -- --check` / `cargo clippy --lib -- -D warnings` -- clean
- [x] `cargo build --bin loom-daemon` -- compiles
- [x] `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/test_orphan_recovery.py loom-tools/tests/test_recovery_stats.py -q` -- 65 pass
- [x] `ruff check` on changed Python files -- clean (pre-existing `ruff format` drift on both files predates this change, confirmed against `origin/main`)

Closes #3975

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -660,6 +660,88 @@ The reaper (`sweep_registry::spawn_reaper_task`) ticks every 30 seconds
 4. Garbage-collects terminal entries older than the retention window
    (default 1 hour).
 
+## Stale-claim reconciliation & the sweep journal (#3953, fixed #3975)
+
+Two independent surfaces reclaim abandoned `loom:building` claims when the
+sweep that owned them has died, using the same evidence source and the same
+decision rule:
+
+| Surface | Where | When it runs |
+|---------|-------|---------------|
+| Rust startup reconciliation | `claim_reconciliation::forge::reconcile_workspace` (called from `main.rs` at daemon startup, guarded by `LOOM_STALE_CLAIM_RECONCILE`, default on) | Once, on every daemon start, across every `effective_roots()` workspace |
+| Python `loom-recover-orphans` | `loom_tools.orphan_recovery.check_untracked_building` | On demand (operator/cron invocation of `loom-recover-orphans [--recover]`) |
+
+Both read the same machine-level **sweep journal** (`~/.loom/sweeps.json`,
+override `LOOM_SWEEPS_JOURNAL_PATH`, written by `sweep_journal::record_sweep`
+on every `SweepRegistry::dispatch`) and apply the same three-way decision
+rule per `loom:building` issue:
+
+1. **Journal entry with a live recorded PID** → keep (genuinely in-flight).
+2. **Journal entry with a dead recorded PID** → reclaim. This is the
+   strongest possible evidence — a specific PID that provably no longer
+   exists — so both surfaces treat it as authoritative.
+3. **No journal entry at all** → reclaim only once the claim has been stale
+   longer than `LOOM_STALE_BUILDING_HOURS` (default 4h; also drives the
+   Rust-side default `DEFAULT_STALE_BUILDING_HOURS`). Absence of a record
+   might mean a live manual `/loom:sweep` the journal was never told about
+   (a pre-journal claim, or the journal file just doesn't exist on this
+   machine), so a much wider grace window applies than for a claim the
+   journal has proof about.
+
+### The #3975 bug: pruning before deciding
+
+The journal is deliberately self-pruning — `sweep_journal::upsert` (every new
+dispatch) and the Rust reconciliation pass both call `prune_dead` to drop
+dead-PID entries so the file never accumulates an unbounded graveyard
+(mirrors `sweep-run-registry.sh`'s `prune_dead`/`peers`, #3768).
+
+Before #3975, `reconcile_workspace` called `prune_dead` **before** calling
+`plan()`/`decide()` — which deleted the exact dead-PID entry the `DeadPid`
+branch above needs to fire its immediate, unconditional reclaim. With the
+entry gone, every claim silently fell through to branch 3 (no record) and
+its much longer `stale_hours` grace window, even for a sweep that had *just*
+died. Two claims in a downstream workspace (issues #6170/#6173: SIGTERMed
+during a daemon restart) sat un-reclaimed for hours because of this —
+`loom-recover-orphans --recover` found and fixed only a third, unrelated
+claim (#6172) whose label happened to already be older than the stale-hours
+threshold. The fix: `reconcile_workspace` now decides against the raw,
+unpruned journal first; pruning happens only afterward, via the normal
+per-reclaimed-issue `remove_sweep` cleanup (leftover dead entries for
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

## Review summary

This is a well-scoped, correct fix for #3975. Reviewed the diff, checked out the branch, and ran the full test matrix.

**Correctness (root cause traced correctly):** `reconcile_workspace` previously called `sweep_journal::prune_dead` + `save` *before* `plan()`/`decide()` — an unconditional prune-and-persist of the shared machine-level journal that erased the exact dead-PID evidence the `DeadPid` branch needs, and (per the new docs) also destroyed that evidence out from under the Python `loom-recover-orphans` tool reading the same file. The fix correctly reorders to decide-then-prune, and now only removes the single reclaimed issue's entry via the existing `remove_sweep` (rather than a wholesale prune+save), so the journal is never mutated based on issues this pass didn't act on. Verified the `Keep`/live-PID path is untouched — no regression there, confirmed by the pre-existing `decide_keeps_when_journal_entry_pid_alive` test still passing and by reading through `decide()`/`plan()` unchanged.

**Regression test:** `reconcile_workspace_reclaims_dead_pid_entry_even_when_label_is_fresh` is a good end-to-end regression test — seeds a dead-PID (pid 0) journal entry, a `loom:building` label applied *seconds* ago (well inside the 4h no-record grace window), and asserts the reclaim still fires immediately via the `DeadPid` branch, plus verifies the `gh issue edit` call and journal cleanup. This is exactly the #6170/#6173 incident shape.

**Acceptance criteria (issue #3975):**
1. ✅ Root cause (journal-pruned dead-PID entries) fixed at the source (Rust startup pass); Python's `journal_pid_dead` reason still applies its existing short `label_grace_period` (600s default) — documented as an intentional, not accidental, asymmetry.
2. ✅ Every staleness-gated skip is now recorded in `OrphanRecoveryResult.watched` and surfaced in both human and `--json` output unconditionally (not gated on `--verbose`) — confirmed via `test_no_record_stale_gate_skip_is_watched_not_silent`, `test_label_grace_gate_skip_is_watched_not_silent`, and the human-format test.
3. ✅ `.loom/docs/daemon-reference.md` documents the shared 3-way decision rule between the Rust startup pass and the Python tool, the #3975 bug, and the one remaining intentional asymmetry.

## Test execution

- `cargo test --lib claim_reconciliation` (loom-daemon) — 12/12 pass, including the new regression test
- `cargo test --lib` (full loom-daemon suite) — 876/876 pass, no regressions
- `cargo fmt -- --check` / `cargo clippy --lib -- -D warnings` — clean
- `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/test_orphan_recovery.py loom-tools/tests/test_recovery_stats.py -q` — 65/65 pass
- `ruff check` on both changed Python files — clean

## CI / merge state

- `gh pr checks 4023` — all required checks pass (Rust Linting & Build Check, Rust Unit Tests; Installer Integration Tests / MCP Server Build skipped as expected for this diff)
- `mergeStateStatus` — `CLEAN`

No blocking issues found. Nice, tightly-scoped fix with good test coverage on both sides and clear documentation of the remaining intentional Rust/Python asymmetry.

Issue #3984, opened 2026-07-26
buildGate.realChangeGlobs is dead config — gate re-runs the full test suite every cycle even when main has not moved


rjwalters:
## Context

`.loom/config.json` declares:

```json
"buildGate": {
  "enabled": true,
  "command": "bash .loom/scripts/build-gate.sh",
  "realChangeGlobs": ["*.rs", "*.toml", "Cargo.lock", "*.py", "*.sh"],
  "timeoutSeconds": 600
}
```

`realChangeGlobs` is **never read**. `grep -rn "realChangeGlobs\|real_change_globs" loom-daemon/src/` returns nothing. The gate consequently re-evaluates unconditionally on every cycle, with no notion of "main has not changed since the last determinate evaluation".

## Observed impact (2026-07-26, tier-0 loom starved for ~2h)

`main` has been at `9e469179` since 10:38Z with **all GitHub checks green** (CI, Security Scan, Lines of Code). Despite that, the gate ran `bash .loom/scripts/build-gate.sh` — `cargo test --workspace` + scoped pytest + the 131-case installer suite — to the full 600s timeout, over and over:

```
02:37:10 WARN  ... timed out after 600s
02:47:11 WARN  ... timed out after 600s
02:56:58 WARN  ... exited with exit status: 101
03:25:19 ERROR ... exited with exit status: 127 — HALTING autonomous dispatch
03:37:23 ERROR ... timed out after 600s — HALTING autonomous dispatch
03:47:24 / 03:57:47 / 04:07:55 / 04:17:56  WARN ... timed out after 600s
```

This is a **self-sustaining doom loop**:

1. The gate runs a full workspace compile+test continuously, ~100% duty cycle.
2. That compile competes for cores with in-flight sweeps (see #3978 — the concurrency cap has no CPU term).
3. Under that contention the gate cannot finish inside 600s, so it is killed and immediately restarted.
4. Timeout is classified Red (`main_health_gate.rs:427`, see #3974), so tier-0 dispatch stays HALTED.

The gate is simultaneously the victim *and* the primary cause of the CPU starvation that keeps it from passing. Honoring `realChangeGlobs` — or even just memoizing the last-evaluated `origin/main` SHA — would have ended this after a single evaluation.

## Secondary effect: priority inversion

[...]

PR #4024, opened 2026-07-27
fix(build-gate,tests): scope local gate to host-deterministic checks; make tmux/timing tests environment-robust (#3985)

### LINKED ISSUE #3985: Local build gate is environment-sensitive where CI is not: dead tmux + CPU starvation from the gate's own loop make it self-reddening

rjwalters:
Closes #3985

## Problem

The local `buildGate.command` (`.loom/scripts/build-gate.sh`) mirrors CI's `cargo test --workspace`, but CI measures **the commit** in a controlled runner while the local gate measures **the host it runs on** — and that host is the one actively running Loom. Two host-dependent failure classes made the gate red for reasons unrelated to `main`'s correctness, and combined with #3984 (continuous re-run) and #3974 (RED misclassification) into a self-reddening loop.

### 1. tmux integration tests (AC1)
- New `tmux_available()` helper in `loom-daemon/tests/common/mod.rs` — probes whether a tmux server can actually be **started** on the `-L loom` socket (stronger than the existing `tmux_server_running()`, which only checks for an already-running server).
- New `require_tmux!()` macro in `integration_basic.rs`, applied to the 6 terminal-lifecycle tests that create tmux sessions. On a host with no usable tmux they now **skip cleanly** (loud `eprintln` + early return) instead of failing.

### 2. Gate scope (AC3)
- `build-gate.sh` Rust step changed from `cargo test --workspace` to `cargo test --workspace --lib --bins`. This runs the crates' unit tests but excludes the integration test **targets** under `loom-daemon/tests/` (which need a live tmux). Those remain covered by CI, which runs the full `cargo test --workspace`.
- Documented the local-gate-vs-CI environment-sensitivity boundary in `.loom/docs/build-gate.md` so the split is intentional.

### 3. Timing-sensitive unit tests (AC2)
- The three flagged `sweep_registry` dispatch tests now share a generous `FIXTURE_CHILD_WAIT_MS` (120s) budget instead of tight 5–10s deadlines. Because the wait helpers poll and return the instant the condition is met, a larger budget is **free on a healthy host** — it only widens tolerance before a genuinely stuck child is declared failed.

### 4. Reduced priority (AC5)
- `build-gate.sh` re-execs itself once under `nice -n 19` (sentinel-guarded against loops) so a long `cargo` compile can never starve the sweeps — or its own timing tests — it shares a host with. Tunable via `LOOM_BUILD_GATE_NICENESS`; disable with `LOOM_BUILD_GATE_NICE=0`.

### Deferred (AC4)
Forge-CI corroboration (prefer a green forge CI result over a local-gate RED on the same SHA) is a distinct, larger daemon-side feature belonging with the RED-classification work in #3974 AC4. It is **not** implemented here; a note in `.loom/docs/build-gate.md` records the deferral.

## Validation

`cargo check --tests` **finished clean (exit 0, no errors/warnings)** — confirming `sweep_registry.rs`, `integration_basic.rs`, and `common/mod.rs` all compile. `bash -n build-gate.sh` passes.

Full test *execution* could not be completed on the build host during this change: the host exhibited exactly the environment pathology this issue describes — macOS Gatekeeper (`syspolicyd`/`amfid`) saturated by a background Backblaze/XProtect scan, stalling every freshly-exec'd fixture child (and even cargo's own link step) well past a minute. That is a host condition, not a code fault; the timing-test changes only widen budgets and cannot break a test that passes in a healthy environment (CI or an unloaded host).

### DIFF
--- a/.loom/docs/build-gate.md
+++ b/.loom/docs/build-gate.md
@@ -122,7 +122,10 @@ The wrapper lives at [`defaults/scripts/build-gate.sh`](../scripts/build-gate.sh
 to it via the `.loom/scripts -> ../defaults/scripts` symlink). It runs three
 stages in order under `set -euo pipefail`, aborting on the first non-zero exit:
 
-1. `cargo test --workspace` — the Rust crates (`loom-daemon`, `loom-api`).
+1. `cargo test --workspace --lib --bins` — the Rust crates' **unit tests**
+   (`loom-daemon`, `loom-api`). The integration test **targets** under
+   `loom-daemon/tests/` are deliberately excluded here — see "Local gate vs.
+   CI" below.
 2. `uv run pytest tests/ -q` in `loom-tools/`, scoped with
    `--ignore=tests/integration` (live-network/credentials e2e) and
    `--ignore=tests/tokens/test_agent_spawn_integration.py` (a slow real-time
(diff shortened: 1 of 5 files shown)


rjwalters:
LGTM! Approved.

Reviewed against issue #3985's acceptance criteria:

- **AC1 (tmux skip-clean)**: `tmux_available()` in `tests/common/mod.rs` correctly probes usability (not just "server already running") by attempting `tmux -L loom new-session -d` and tearing it down. `require_tmux!()` is applied to exactly the 6 tests that assert `tmux_session_exists(...)` (`test_create_terminal`, `test_create_terminal_with_working_dir`, `test_list_terminals`, `test_destroy_terminal`, `test_send_input`, `test_multiple_clients`) — matches the issue's failure list 1:1, and is placed before `TestDaemon::start()` so a skip leaks no daemon/tmux resources. `test_destroy_nonexistent_terminal` (no tmux assertion) is correctly left unguarded.
- **AC2 (wall-clock robustness)**: `FIXTURE_CHILD_WAIT_MS` (120s) replaces the three tight 5–10s deadlines the issue calls out by name (`dispatch_and_reap_wire_the_sweep_journal`, `dispatch_forwards_derived_package_path_when_unset`, `dispatch_applies_configured_stagger_between_spawns`). All three callers poll-and-return-early, so the widened budget costs nothing on a healthy host.
- **AC3 (gate scope + docs)**: `build-gate.sh` now runs `cargo test --workspace --lib --bins`, excluding the `loom-daemon/tests/` integration targets from the local gate while CI still runs the full `--workspace` suite. `.loom/docs/build-gate.md` documents the split clearly with a "Local gate vs. CI" section and a good rule-of-thumb heuristic.
- **AC4 (forge-CI corroboration)**: explicitly and reasonably deferred to #3974 with a documented rationale — appropriate scoping, not scope creep avoidance.
- **AC5 (nice re-exec)**: sentinel-guarded (`LOOM_BUILD_GATE_NICED`) single re-exec via `exec nice -n ...` is correct — no fork/zombie, no infinite loop, graceful fallback when `nice` is absent, tunable/disableable via env vars exactly as documented.

Verification performed:
- `bash -n defaults/scripts/build-gate.sh` — clean.
- `cargo check --tests` (loom-daemon) — clean, no warnings.
- `cargo build --lib --bins --tests` — compiles clean (2m14s).
- `cargo test --workspace --lib --bins` — loom-api unit tests passed 4/4; loom-daemon unit tests started cleanly but this review host is itself under heavy CPU contention (ironically the exact pathology #3985 describes) so a full local run couldn't be timed reliably — not held against the PR.
- `gh pr checks 4024` — all CI checks green, including "Rust Unit Tests" (which runs the full `cargo test --workspace`, i.e. including the integration targets this PR excludes from the local gate — confirms nothing was broken).
- `mergeStateStatus` — CLEAN.

One minor nit, not blocking: the doc comment above `FIXTURE_CHILD_WAIT_MS` in `sweep_registry.rs` (and the mirrored line in `.loom/docs/build-gate.md`) says "A 60s budget is..." but the constant is actually `120_000` ms (120s). Worth a quick follow-up fix for accuracy, doesn't affect behavior.

PR #4056, opened 2026-07-27
fix(daemon): close forge-CI unanimity gap for paths-filtered verification workflows

### LINKED ISSUE #3987: main-health gate: bookkeeping-only workflow can vouch for a commit in repos where CI declines to run
Follow-up from the review of #3986 (issue #3974). **Not reachable in this repo** — filed so the residual gap is tracked rather than lost in a PR thread.

## Context

`parse_gh_run_list` in `loom-daemon/src/main_health_gate.rs` reduces the forge-CI runs for the gate's evaluated SHA to a `CiVerdict`. #3986 correctly hardened it to a unanimity rule: `Success` requires at least one genuine `success` **and** no outstanding (`queued`/`in_progress`) or indeterminate (`cancelled`/`action_required`/`stale`/unrecognized) sibling run for the same commit.

That closes both fail-open paths originally reported, without hard-coding which workflow "counts" — deliberately, because the daemon dispatches into consumer repos whose verification workflow is not named `CI`, and a name gate would degrade to permanent `Unknown` there and silently retire the corroboration path entirely.

## The residual gap

The unanimity rule reasons over *runs that exist*. A workflow filtered out by `paths` / `paths-ignore` produces **no run at all** — not a `skipped` run — so it is invisible to the reducer.

In a repo where the real verification workflow legitimately does not run for a commit, and only a fast bookkeeping workflow (line counter, labeler) runs and succeeds:

- every run for the SHA is `completed`,
- at least one concluded `success`,
- nothing is outstanding or indeterminate,

so the probe returns `Success` and **relaxes a local red halt on a commit CI never judged**. That is the same class of non-evidence #3986 set out to reject, reached through a different door.

## Why it is not reachable in this repo

`.github/workflows/ci.yml` triggers on `push: branches: [main]` with **no `paths` / `paths-ignore` filter**, so a run is created for every commit on `main`. The `dorny/paths-filter@v4` usage inside it is a *job-level* skip within an existing run, which still yields a run-level `success` (the `Detect Changes` job always runs). Empirically, 9 of the last 9 `main` SHAs have a `CI` run present.

The exposure is to **consumer repos** the daemon dispatches into, whose workflow config Loom does not control.

## Possible directions (not prescriptive)

- Treat "no run at all for the SHA from any workflow that has run on this branch recently" as `Unknown` — derive the expected workflow set from recent branch history rather than from a hard-coded name, so it stays config-free and self-tuning.
- An optional, explicitly-unset-by-default `LOOM_GATE_CI_WORKFLOW` that, *when set*, requires that workflow to have concluded `success`. Absent ⇒ today's unanimity rule, so no repo silently changes behavior.
- Require ≥ N successful runs, or a minimum run duration, as a proxy for "a real build ran". Weakest of the three; noted only for completeness.

Any fix must preserve the property #3986 established: it must not reintroduce a mode where the corroboration path can never fire, since a permanently-`Unknown` probe means the gate halts on every environmental hiccup — the original #3974 bug.

Discovered during code review of #3986.

---

Verified against `origin/main` @ `a68099d5`. The original analysis holds: `parse_gh_run_list`
(`loom-daemon/src/main_health_gate.rs:882`) reasons only over *runs that exist*, and a
`paths`-filtered workflow that never starts produces no run to reason about.

The curation below **evaluates the four candidate directions against real data** and narrows
the issue to one implementable approach, because three of the four do not survive contact
with this repo's actual run history.

### Finding 1 — Direction 1 (derive the expected workflow set from branch history) cannot work

The proposal was: treat "a workflow that recently ran on this branch has *no* run for this SHA"
as `Unknown`. Real `gh run list --branch main --limit 30` output for this repo (10 distinct
SHAs, 2026-07-27):

| Workflow | SHAs present (of 10) | Why |
|---|---|---|
| `CI` | 10 | no `paths` filter |
| `Lines of Code` | 10 | no `paths` filter |
| `Security Scan` | 10 | no `paths` filter |
| `Shell Script Linting` | **1** | `paths`-filtered to shell files |

This is a **squeeze**, not a tuning problem:

- A presence threshold **low enough** to treat a `paths`-filtered *verification* workflow as
  expected (the repo shape the issue is about) also treats `Shell Script Linting` as expected
  for all 10 SHAs → permanent `Unknown` → the corroboration path can never fire → this
  reintroduces **#3974**, the exact regression the issue forbids.
- A threshold **high enough** to exclude `Shell Script Linting` (≈100% presence) also excludes
  every `paths`-filtered workflow → the gap stays open in precisely the repos it targets.

The discriminating fact — *is this workflow the one that verifies the code?* — is simply not
present in `gh run list` output, at any threshold. Filtering the expected set to `event: push`
does not help either: all four workflows above are `push` runs. The 30-run window (≈10 commits)
is also far too small to estimate per-workflow firing frequency.

**Direction 1 is not implementable without one of the two failure modes above. Do not pursue it.**

### Finding 2 — A fifth direction (required status checks) is a no-op as Loom provisions repos

Not in the original list, and worth recording so it is not re-proposed: derive the expected set
from the branch's own declaration of what must be green. Probed on this repo:

```
gh api repos/rjwalters/loom/branches/main/protection  → 404 Branch not protected
gh api repos/rjwalters/loom/rules/branches/main       → ["deletion","non_fast_forward",
                                                          "required_linear_history","pull_request"]
```

`scripts/install/setup-branch-protection.sh` provisions **no** `required_status_checks` rule, so
the authoritative expected set is empty in Loom-installed repos. Closing the gap this way would
require an installer change *plus* per-repo knowledge of the check name — which is Direction 2's
configuration cost with an extra API call and an admin-scoped permission on top. **Rejected.**

### Finding 3 — Direction 3 (≥N successes / minimum run duration) is a proxy, not evidence

Already flagged weakest in the original. A bookkeeping workflow can be slow; a cached real build
can be fast. It substitutes a correlate for the fact and would produce both false greens and
false halts. **Rejected.**

### Recommendation — implement Direction 2 only

An explicitly-unset-by-default workflow name is the only direction that closes the gap without a
fail-closed regression. It puts the "which workflow counts" knowledge exactly where that
knowledge actually exists (with the operator who configured the consumer repo), and its absent
default is byte-for-byte today's behavior — so no repo silently changes.

**Resolution** (mirrors the established `LOOM_MAIN_HEALTH_GATE` pattern, precedence
**env > config > default(None)**):

1. `LOOM_GATE_CI_WORKFLOW` env var, or
2. `.loom/config.json` → `autonomous.mainHealthGate.ciWorkflow` (string), or
3. **absent ⇒ today's unanimity rule, unchanged.**

`autonomous.mainHealthGate` is the right home, not `buildGate`: its doc comment at
`main_health_gate.rs:1996` already reserves it for "future tuning knobs", whereas `buildGate` is
shared with the builder-side worktree gate (`.loom/scripts/build-gate.sh`) that has no business
knowing about forge CI.

**Semantics when a workflow name IS configured** — an additional requirement layered *on top of*
the #3986 unanimity rule, never a relaxation of it:

| Runs for the evaluated SHA | Verdict |
|---|---|
| named workflow: `completed` + `success`, and unanimity rule otherwise satisfied | `Success` |
| named workflow: **no run at all** | `Unknown` — *this is the gap closure* |
| named workflow: `skipped` / `neutral` | `Unknown` (a required workflow that declined to run did not verify the commit — note this **differs** from the unnamed case, where `skipped` is a definitive "did not apply" that does not block `Success`) |
| named workflow: `queued` / `in_progress` / `cancelled` / `action_required` / `stale` / unrecognized | `Unknown` (already the behavior) |
| any run for the SHA `failure` / `timed_out` / `startup_failure` | `Failure` — **unchanged**, and still checked first |

`Failure` detection must remain untouched: a halt may always be *established*, it may only never
be *relaxed* on non-evidence.

**Misconfiguration guardrail (required, not optional).** A typo'd workflow name would otherwise
pin the gate to permanent `Unknown` — silently recreating #3974 for that repo. When a name is
configured but appears for **no** SHA anywhere in the probe window, emit a `log::warn!` naming the
configured value and the workflow names actually observed. Still return `Unknown` (fail safe),
but make the cause visible in `~/.loom/daemon.log`.

**Wiring note.** Resolve the name inside `CommandGateRunner::new` from `repo_root` (env → config
→ `None`) and pass it into `GhForgeCi`. This leaves all ~12 existing `CommandGateRunner::new`
call sites and the `run_gate_tick` signature untouched; tests already override the probe via
`with_ci_status`.

### Scope guard

Roughly **~120 LOC production + ~80 LOC tests**, one production file plus two docs. Explicitly
**out of scope**: any change to `setup-branch-protection.sh`, any new API call, any heuristic
inference of the expected workflow set, and any change to `CI_PROBE_RUN_LIMIT`. If implementation
pressure pushes toward any of those, stop and comment on this issue rather than expanding.

## Acceptance Criteria

- [ ] `LOOM_GATE_CI_WORKFLOW` and `autonomous.mainHealthGate.ciWorkflow` are both read, with
      precedence **env > config > default(None)**, soft-failing to `None` on missing/malformed
      config (matching `read_autonomous_gate_config`'s existing contract).
- [ ] **With no name configured, `parse_gh_run_list` behavior is unchanged** — every existing
      assertion in `test_parse_gh_run_list_verdicts` and
      `test_parse_gh_run_list_non_evidence_is_never_success` still passes untouched.
- [ ] With a name configured and that workflow having **no run** for the evaluated SHA, the
      verdict is `Unknown` even when a bookkeeping workflow concluded `success` — the gap in the
      issue title is closed.
- [ ] With a name configured and that workflow `completed`/`success` (unanimity otherwise
      satisfied), the verdict is `Success` — the corroboration path remains reachable, preserving
      the #3986/#3974 property.
- [ ] With a name configured and that workflow `skipped`, the verdict is `Unknown`.
- [ ] A `failure` on **any** run for the SHA still yields `Failure`, configured or not.
- [ ] A configured name matching no workflow anywhere in the probe window logs a `warn!` naming
      both the configured value and the observed workflow names.
- [ ] `.loom/docs/daemon-reference.md` § "Forge-CI corroboration of a local red" documents the new
      knob and extends the verdict table with the configured-workflow column.
- [ ] `cargo test -p loom-daemon` passes; `cargo clippy` clean.

## Affected Files

- `loom-daemon/src/main_health_gate.rs` — add `ci_workflow: Option<String>` to
  `AutonomousGateConfig` (:2004) and read it in `read_autonomous_gate_config` (:2017); resolve
  env-over-config in `CommandGateRunner::new` (:1045); give `GhForgeCi` (:827) the name and thread
  it into `parse_gh_run_list` (:882); add the misconfiguration `warn!`; extend the two existing
  parser tests (:2851, :2890) with configured-name cases.
- `.loom/docs/daemon-reference.md` — § "Forge-CI corroboration of a local red" (:543-584): document
  the knob, extend the verdict table, and state the default-absent no-change guarantee.
- `.loom/docs/build-gate.md` — one cross-reference note only, so a reader of the `buildGate` config
  table (:49-54) is not left thinking the new key belongs there. No behavior change.

## Test Plan

- [ ] Automated (unit, `loom-daemon/src/main_health_gate.rs` test module):
  - [ ] `read_autonomous_gate_config` parses `ciWorkflow`; absent/malformed → `None`.
  - [ ] Env `LOOM_GATE_CI_WORKFLOW` overrides the config value; unset falls back to config.
  - [ ] `parse_gh_run_list` with `None` — assert the two existing tests are unmodified.
  - [ ] `parse_gh_run_list` with `Some("CI")`, fixture = bookkeeping `success` only, **no** `CI`
        run → `Unknown`. (This is the regression test for this issue.)
  - [ ] Same fixture plus `CI` `completed`/`success` → `Success`.
  - [ ] `CI` `skipped` alongside a bookkeeping `success` → `Unknown`.
  - [ ] `CI` `failure` with a configured name → `Failure`.
  - [ ] Configured name absent from the window → `Unknown` + warn path exercised.
- [ ] Regression fixture from real data: capture this repo's actual
      `gh run list --branch main --limit 30 --json headSha,status,conclusion,workflowName` output
      as a test fixture and assert the verdict for a green SHA is `Success` **both** with no name
      configured and with `ciWorkflow: "CI"`. This is the guard against the Finding-1 failure mode
      (`Shell Script Linting` appears for only 1 of 10 SHAs and must not perturb any verdict).
- [ ] Manual verification: with `autonomous.mainHealthGate.ciWorkflow` set to a deliberately
      wrong name, start the daemon and confirm `~/.loom/daemon.log` shows the misconfiguration
      warning and that a local red is **not** relaxed.
- [ ] Edge cases: empty-string config value treated as unset; leading/trailing whitespace trimmed;
      matching is exact on `workflowName` (document case sensitivity in the doc update).

## Dependencies

None. #3986 (the unanimity rule this builds on) is merged; this layers an additional requirement
on top of it and does not modify it.

rjwalters:
## Summary

`parse_gh_run_list`'s #3986 unanimity rule reasons only over *runs that exist*.
A workflow excluded from a commit by a `paths`/`paths-ignore` filter produces
**no run at all** (not a `skipped` run), so it is invisible to the reducer. In
a repo where the real verification workflow legitimately does not run for a
commit and only a fast bookkeeping workflow (line counter, labeler) runs and
succeeds, the probe returned `Success` and relaxed a local red halt on a
commit CI never actually judged.

This adds an optional, unset-by-default `ciWorkflow` knob
(`LOOM_GATE_CI_WORKFLOW` / `autonomous.mainHealthGate.ciWorkflow`, precedence
env > config > `None`) that, when configured, additionally requires the named
workflow to have concluded `success` for the evaluated SHA -- layered **on top
of**, never a relaxation of, the existing unanimity rule. Absent, behavior is
byte-for-byte unchanged.

## Changes

- `loom-daemon/src/main_health_gate.rs`:
  - `GATE_CI_WORKFLOW_ENV` constant + `resolve_ci_workflow` (env > config >
    `None`, trims whitespace, empty string treated as unset).
  - `AutonomousGateConfig.ci_workflow` read from
    `autonomous.mainHealthGate.ciWorkflow`.
  - `GhForgeCi` now carries an optional `ci_workflow` name, threaded into
    `parse_gh_run_list`, which additionally requires that workflow to have a
    `completed`/`success` run for the SHA when configured.
  - Misconfiguration guardrail: a configured name matching no workflow
    anywhere in the probe window is surfaced via `log::warn!` (naming both the
    configured value and the observed workflow names) while still failing
    safe to `Unknown` -- preserves the #3974/#3986 invariant that the
    corroboration path must never become permanently unreachable.
  - Tests: named-workflow verdict matrix (no run / skipped / success /
    failure / unmatched name / case-sensitivity), config parsing
    (empty/whitespace/trim), env-over-config precedence, and a regression
    fixture captured from this repo's real `gh run list` shape (a
    low-frequency `paths`-filtered workflow must not perturb a green SHA's
    verdict, named or unnamed).
- `.loom/docs/daemon-reference.md`: new "Optional named verification workflow"
  subsection with the extended verdict table + a config-table row; `.loom/docs/build-gate.md`:
  cross-reference note so the key isn't mistaken for a `buildGate` setting.

## Test plan

- [x] `cargo test main_health_gate` -- 85 tests pass (new + existing, unchanged
      unnamed-path assertions untouched)
- [x] `cargo clippy --all-targets` -- clean
- [x] `cargo fmt -- --check` -- clean
- [x] `cargo build --release` -- compiles cleanly
- [x] Full `cargo test` -- one unrelated flake (`integration_basic::test_send_input`,
      a daemon-socket-startup timing test under full-suite parallelism);
      confirmed passing in isolation, unrelated to this change

Closes #3987

### DIFF
--- a/.loom/docs/build-gate.md
+++ b/.loom/docs/build-gate.md
@@ -53,6 +53,12 @@ The gate is **opt-in**. Repos with no `buildGate` block in `.loom/config.json` s
 | `realChangeGlobs` | array of strings | _(default exclusions)_ | Positive globs. A changed file must match at least one to count as "real." When omitted, every changed file counts unless it matches one of the default scratch exclusions: `.loom-*`, `*.log`, `.no-changes-needed`. |
 | `timeoutSeconds` | integer | `600` | Timeout for the `command` run. |
 
+> **Not a `buildGate` key:** the daemon-side main-health gate's optional forge
+> verification-workflow name (`autonomous.mainHealthGate.ciWorkflow` /
+> `LOOM_GATE_CI_WORKFLOW`, #3987) lives under `autonomous.mainHealthGate`, **not**
+> here. `buildGate` is the builder-side worktree quality gate; it has no business
+> knowing about forge CI. See [Optional named verification workflow](daemon-reference.md#optional-named-verification-workflow-loom_gate_ci_workflow-3987).
+
 ## Examples
 
 ### Rust workspace
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -579,6 +579,47 @@ fast bookkeeping workflow that finishes minutes before the real build would
 otherwise vouch for the commit on its own. Neither case needs the probe to know
 *which* workflow "counts", so no workflow name is hard-coded.
 
+#### Optional named verification workflow (`LOOM_GATE_CI_WORKFLOW`, #3987)
+
+The unanimity rule above reasons only over *runs that exist*. A workflow a
+`paths` / `paths-ignore` filter excluded from a commit produces **no run at all**
+(not a `skipped` run), so it is invisible to the reducer. In a **consumer repo**
+whose real verification workflow is `paths`-filtered off a commit while only a
+fast bookkeeping workflow (line counter, labeler) runs and succeeds, every run is
+`completed`/`success` and the reducer above returns `green` for a commit the real
+build never judged — relaxing a local red on non-evidence. (Not reachable in this
+repo: `.github/workflows/ci.yml` has no `paths` filter, so a `CI` run exists for
+every `main` commit.)
+
+The optional `ciWorkflow` knob closes that gap by naming the workflow that
+*counts*. It is **unset by default** — absent, behavior is byte-for-byte the
+unanimity rule above, so no repo silently changes. Resolution mirrors
+`LOOM_MAIN_HEALTH_GATE`, precedence **env > config > default(None)**:
+
+1. `LOOM_GATE_CI_WORKFLOW` env var (empty / whitespace-only ⇒ falls through), or
+2. `.loom/config.json` → `autonomous.mainHealthGate.ciWorkflow` (string; empty /
+   whitespace ⇒ unset), or
+3. absent ⇒ the unanimity rule, unchanged.
+
+When a name **is** configured it layers **one additional requirement on top of**
+— never a relaxation of — the unanimity rule: the named workflow must itself have
+concluded `success` for the SHA. Matching is **exact and case-sensitive** on
+`workflowName`.
+
+| `ciWorkflow` set, runs for the evaluated SHA | Verdict |
+|---|---|
+| named workflow `completed`/`success`, unanimity otherwise satisfied | **green** |
+| named workflow has **no run at all** | unknown — *the gap closure* |
+| named workflow `skipped` / `neutral` | unknown (a required workflow that declined did not verify the commit — differs from the unnamed case, where `skipped` does not block green) |
+| any run `failure` / `timed_out` / `startup_failure` | red (unchanged, still checked first) |
+
+**Misconfiguration guardrail.** A typo'd name would otherwise pin the gate to
+permanent `unknown` (silently recreating #3974 for that repo). When a configured
+name appears for **no** SHA anywhere in the probe window, the probe emits a
+`log::warn!` naming the configured value and the workflow names actually
+observed, then still returns `unknown` (fail safe) — so the cause is visible in
+`~/.loom/daemon.log`.
+
 Corroboration is on by default and can be disabled with
 `LOOM_GATE_CI_CORROBORATION=0` (for repos with no forge CI or no `gh`); it is only
 probed on a local red, never on a green run.
@@ -1275,6 +1316,7 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.workFinder.quarantine.instaCrashSecs` | `LOOM_WORK_FINDER_QUARANTINE_INSTA_CRASH_SECS` | `60` | Checkpoint-less death within this window of dispatch counts as an insta-crash. Zero/invalid → default |
 | `autonomous.perTokenConcurrency` | `LOOM_PER_TOKEN_CONCURRENCY` | `2` | Concurrent sweeps **per healthy token** in the cap (#3947). Zero/invalid → default; clamped to a floor of 1 |
 | `autonomous.mainHealthGate.enabled` | `LOOM_MAIN_HEALTH_GATE` | `false` | Gate loop on/off |
+| `autonomous.mainHealthGate.ciWorkflow` | `LOOM_GATE_CI_WORKFLOW` | *(unset)* | Forge workflow that must itself conclude `success` for forge-CI corroboration to vouch for a commit (#3987). Empty/whitespace → unset. Absent → today's unanimity rule, unchanged. See [Optional named verification workflow](#optional-named-verification-workflow-loom_gate_ci_workflow-3987) |
 | `autonomous.roleRunner.enabled` | `LOOM_ROLE_RUNNER` | `false` | Periodic standalone support-role runner on/off (#4015) |
 | `autonomous.roleRunner.roles` | *(config only)* | all 5 roles | Subset of `champion`/`curator`/`judge`/`auditor`/`guide` to dispatch; explicit empty array runs none |
 | `autonomous.roleRunner.intervalSecs` | `LOOM_ROLE_RUNNER_INTERVAL_SECS` | per-role built-in (5–15 min) | Uniform override applied to every enabled role's cadence |
--- a/loom-daemon/src/main_health_gate.rs
+++ b/loom-daemon/src/main_health_gate.rs
@@ -776,6 +776,18 @@ pub trait GateRunner {
 /// kill switch (`0`/`false`/`no`/`off`) for repos with no forge CI or no `gh`.
 pub const GATE_CI_CORROBORATION_ENV: &str = "LOOM_GATE_CI_CORROBORATION";
 
+/// Environment variable naming the forge workflow that must have concluded
+/// `success` for a commit before forge-CI corroboration will vouch for it
+/// (#3987). **Unset by default** — absent, corroboration keeps the #3986
+/// unanimity rule byte-for-byte. When set (env overrides
+/// `autonomous.mainHealthGate.ciWorkflow`), the named workflow must have a
+/// `completed`/`success` run for the evaluated SHA or the verdict degrades to
+/// [`CiVerdict::Unknown`]. This closes the residual gap where a repo whose real
+/// verification workflow is `paths`-filtered out of a commit — so it produces
+/// **no run at all** — could have a fast bookkeeping workflow (line counter,
+/// labeler) vouch for the commit on its own. See the module docs.
+pub const GATE_CI_WORKFLOW_ENV: &str = "LOOM_GATE_CI_WORKFLOW";
+
 /// How long to wait for the forge-CI probe before giving up (and keeping the
 /// local verdict). Deliberately short: this runs inside the gate tick, and an
 /// unavailable answer is the safe answer.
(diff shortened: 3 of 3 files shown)


rjwalters:
LGTM! Approved.

Verified:
- `cargo build --release` and `cargo test main_health_gate` (85 passed), `cargo clippy --all-targets --release`, `cargo fmt -- --check` all clean in the worktree checkout.
- CI: Rust Linting & Build Check pass, Rust Unit Tests pass, mergeStateStatus CLEAN.
- Acceptance criteria from #3987 all covered: env > config > default(None) precedence (`resolve_ci_workflow`), empty/whitespace treated as unset (both config parse and env), exact case-sensitive match on `workflowName`, and the full fixture matrix — no-run-for-named-workflow → Unknown, named+success → Success, named+skipped → Unknown, named+failure → Failure (checked first, unaffected by config), absent-from-window → Unknown + `log::warn!` naming both configured and observed names, plus the real-data regression fixture (`Shell Script Linting` paths-filtered to 1/2 SHAs doesn't perturb a green SHA's verdict, named or unnamed).
- Composability with #3986 confirmed: `None` path is byte-for-byte unchanged (all pre-existing `parse_gh_run_list` assertions pass with `None` as the third arg, no logic touched on that branch), and the named path only ever adds a stricter requirement (`Some(_) if !named_workflow_success => Unknown`) on top of the unanimity result — never a separate/looser path, so it can't reintroduce the #3974 permanent-Unknown failure mode when unconfigured.
- Docs: `.loom/docs/daemon-reference.md` new subsection + config table row, and the `.loom/docs/build-gate.md` cross-reference note, both match the implementation (including the case-sensitivity note and the `skipped` divergence from the unnamed case).
- No new call sites elsewhere in the codebase reference `GhForgeCi`/`parse_gh_run_list`, so this is fully self-contained to `main_health_gate.rs` plus docs.

Nice work narrowing the Curator's four candidate directions down to the one that survives contact with real data.

Issue #3895, opened 2026-07-23
daemon: sweeps that die mid-build (token exhausts mid-run) aren't caught by the startup watchdog


rjwalters:
## Context

The startup-hang watchdog (#3887 → PR #3892) detects a sweep that shows **no progress past the spawn header** (no worktree / no checkpoint / no log-past-header) and does one bounded re-dispatch. But this session we also observed a **different** failure it does not catch: a sweep that got well into the build phase (created its worktree, made file edits — `dirty=3`) and then its child process **died** (no live process), leaving the issue silently reverted to `loom:issue` with a dirty, uncommitted worktree and no PR. Observed on #3891.

Because `sweep_made_progress` returns `Healthy` as soon as a worktree exists, the watchdog correctly leaves a *slow* sweep alone — but it therefore also does **not** rescue a sweep that had progress and then crashed/exited (likely a token exhausting mid-run, or the child dying for another reason).

## Impact
A mid-build death is silent: the issue looks like fresh `loom:issue` work, the worktree is left dirty, and nothing re-dispatches. In autonomous mode this wedges the issue indefinitely.

## Acceptance criteria
- [ ] Detect a sweep whose child has **exited without producing a PR / merge** while its issue is still `loom:building` (or reverted) and its worktree is non-empty — distinct from "no startup progress."
- [ ] Bounded auto-recovery: reset/clean the worktree and re-dispatch once (respecting the same retry cap philosophy as #3892), or surface it clearly for operator action.
- [ ] Pre-flight token-health gate on (re-)dispatch: skip accounts marked `exhausted` in the ranking so a mid-run exhaustion is less likely to recur (ties into the ranking-freshness issue).
- [ ] Test: a sweep with a worktree + edits whose process is killed is detected and bounded-re-dispatched (not left wedged).

## Notes
Distinct from #3887/#3892 (startup hang, no progress) — this is the "made progress then died" case. Together they cover both liveness-failure modes of an autonomous dispatch.

PR #4066, opened 2026-07-27
fix(daemon): quarantine insta-crashes despite stale sweep checkpoints (#4009)

### LINKED ISSUE #4009: Insta-crash quarantine cannot fire for any issue with a checkpoint on disk — pre-work deaths re-dispatch every 60s forever
## Summary

The #3939 insta-crash quarantine **cannot fire for any issue that has ever written a sweep checkpoint**. Because checkpoints persist on disk across dispatches, a single successful-enough run permanently exempts an issue from the tally — so a sweep that afterwards dies pre-work in 2 seconds is re-dispatched by the work finder every 60s, forever, with nothing escalating.

Observed live on the #3968 sweep during the 2026-07-26 session.

## Mechanism

The reaper's terminal-outcome branch keys entirely off checkpoint *presence*, not on whether this particular run made progress — `loom-daemon/src/sweep_registry.rs:2459`:

```rust
let checkpoint = self.config.checkpoint_dir().join(format!("issue-{issue}.json"));
if checkpoint.exists() {
    ...
    // "a checkpoint exists, so this sweep reached real work"
    self.record_terminal_outcome(issue, false);   // :2486 — resets the tally
} else {
    ...
    let insta_crash = duration_sec < self.quarantine_config.insta_crash_secs
        && exit_code != Some(0);
    self.record_terminal_outcome(issue, insta_crash);   // :2538 — the only place it can be true
}
```

`record_terminal_outcome(issue, false)` (`:1593`) removes the issue from `insta_crash_counts`, so the consecutive run resets on every checkpointed death regardless of duration or exit code.

The comment at `:2481` states the intent — *"a checkpoint exists, so this sweep reached real work — the mid-build-death watchdog (#3895) owns this failure mode"*. That inference is only sound if the checkpoint was written **by the run that just died**. It wasn't necessarily:

- `.loom/sweep-checkpoint/issue-<N>.json` is written for resumability and survives process exit by design (#3373).
- Checkpoints are only removed explicitly (`.loom/scripts/sweep-checkpoint.sh delete <issue>`), which does not happen on a crash.
- Stale files accumulate for weeks — this checkout currently holds `issue-3784.json`, `issue-3786.json`, `issue-3810.json` and others dated Jul 23, still present Jul 26.

So the sequence is:

1. Sweep for #N runs, reaches Curator, writes `issue-N.json`, then dies for any reason.
2. Every subsequent dispatch of #N — including one that dies in 2s before selecting a token — sees `checkpoint.exists() == true`, is classified `Crashed`, and **resets** the tally.
3. `insta_crash_counts[N]` never reaches `threshold` (default 3), so the issue is never quarantined.
4. `restore_label_to_ready` flips `loom:building` → `loom:issue` on the way out, the work finder re-picks it on the next ~60s tick, and the loop repeats indefinitely.

The failure modes the quarantine exists to catch — the #3938 missing-token-pool case, and the weekly-exhausted-token deaths from #3991 — are exactly pre-work deaths, and are exactly the ones that reach a checkpointed issue and get waved through.

## Why the guard still looks correct in tests

The unit tests (`record_terminal_outcome_counts_then_quarantines_at_threshold`, `..._resets_run_on_progress`, `:5187`+) call `record_terminal_outcome` directly with a hand-supplied boolean, so they never exercise the classification at `:2459`. The bug is entirely in how that boolean is derived.

## Acceptance criteria

1. A pre-work death is classified as an insta-crash **regardless of checkpoint presence** — i.e. the `duration_sec < insta_crash_secs && exit_code != Some(0)` test applies on both branches.
2. Distinguish "this run made progress" from "a checkpoint file exists on disk". Options, roughly in order of preference:
   - compare the checkpoint's mtime (or a recorded phase/generation) against this sweep's `started_at`, so only a checkpoint written *by this run* counts as progress; or
   - track progress in-memory on the `SweepInfo` entry (phase events already flow through the registry) and use the on-disk file only for resume.
3. Genuine mid-build deaths still reset the tally and remain owned by the #3895 watchdog — no regression in `record_terminal_outcome_resets_run_on_progress`'s intent.
4. A regression test that drives the reaper path (not `record_terminal_outcome` directly) with a **stale** checkpoint present + a sub-window non-zero exit, and asserts the tally increments and the issue quarantines at threshold.

## Notes

- Related but distinct: #4007 (re-dispatch loop under host load spawning duplicate sweeps) shares the "issue keeps getting re-dispatched" symptom but has a different cause; worth cross-referencing, not merging.
- Consider whether stale checkpoints should be garbage-collected at all (age-based, or on issue close) — currently nothing prunes them, which is what makes the exemption permanent rather than transient.

rjwalters:
Closes #4009

## Problem

The #3939 insta-crash quarantine could never fire for any issue that had ever written a sweep checkpoint. The reaper's terminal-outcome branch (`loom-daemon/src/sweep_registry.rs`) keyed off checkpoint *presence* (`checkpoint.exists()`) rather than whether *this particular run* made progress.

Because `.loom/sweep-checkpoint/issue-<N>.json` files persist on disk indefinitely (only removed by an explicit `sweep-checkpoint.sh delete`, which never runs on a crash), a single successful-enough historical run permanently exempted the issue: every subsequent pre-work death saw `checkpoint.exists() == true`, was classified `Crashed`, and **reset** the consecutive tally. The threshold was never reached, `restore_label_to_ready` flipped `loom:building` → `loom:issue`, and the work finder re-dispatched every ~60s forever. Observed live: issue #4009 itself crash-looped 3x.

## Fix

Derive the tally from whether the checkpoint was written *by this run*, not from mere file presence.

- New helper `checkpoint_written_by_run(path, started_at)` compares the checkpoint file's mtime against the sweep's `started_at` (the acceptance criteria's preferred mtime approach; `latest_phase` is not updated live during a run so an in-memory signal was not available).
- In the reaper's checkpoint-exists branch: only a checkpoint at/after `started_at` counts as progress (a genuine mid-build death — the #3895 watchdog's remit) and resets the tally. A **stale** checkpoint falls through to the same pre-work insta-crash test the checkpoint-less branch already uses (`duration_sec < insta_crash_secs && exit_code != Some(0)`), so a sub-window non-clean death still counts toward quarantine.
- The `SweepCrashed` event and label-restore semantics are unchanged — only the quarantine tally derivation changed.

## Acceptance criteria

1. Pre-work death classified as insta-crash regardless of checkpoint presence — the insta-crash test now applies on both branches. ✅
2. "This run made progress" (checkpoint mtime ≥ `started_at`) distinguished from "a file merely exists on disk". ✅
3. Genuine mid-build deaths (checkpoint written during THIS run) still reset the tally and remain owned by #3895 — `reaper_checkpoint_death_does_not_count_as_insta_crash` retained (updated to write the checkpoint during each run so it genuinely post-dates `started_at`). ✅
4. New regression test `reaper_stale_checkpoint_death_counts_as_insta_crash` drives the reaper path with a stale checkpoint (mtime predates `started_at`) plus sub-window non-clean exits, asserting the tally increments and the issue quarantines at threshold. ✅

## Testing

- `cargo test --lib` — 1001 passed, 0 failed.
- `cargo clippy --lib` — clean.

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -2560,13 +2560,29 @@ impl SweepRegistry {
                                 sweep_id: sweep_id.clone(),
                                 outcome: SweepOutcome::Crashed,
                             });
-                            // Insta-crash quarantine (#3939): a checkpoint exists,
-                            // so this sweep reached real work — the mid-build-death
-                            // watchdog (#3895) owns this failure mode, and it is
-                            // explicitly NOT an insta-crash. Reset the consecutive
-                            // tally so a genuine mid-build death never accretes
-                            // toward quarantine.
-                            self.record_terminal_outcome(issue, false);
+                            // Insta-crash quarantine (#3939 + #4009): a checkpoint
+                            // FILE existing on disk does not prove THIS run made
+                            // progress — checkpoints persist across dispatches
+                            // (#3373), so a single successful-enough historical run
+                            // would otherwise exempt the issue from quarantine
+                            // forever while every later dispatch dies pre-work in
+                            // <2s (an infinite re-dispatch loop, #4009). Only a
+                            // checkpoint (re)written by THIS run — mtime at/after
+                            // our `started_at` — counts as progress. Such a genuine
+                            // mid-build death is the mid-build-death watchdog's
+                            // remit (#3895) and resets the consecutive tally. A
+                            // stale checkpoint from an earlier dispatch does not:
+                            // fall through to the same pre-work insta-crash test the
+                            // checkpoint-less branch below uses, so a sub-window
+                            // non-clean death still counts toward quarantine.
+                            if checkpoint_written_by_run(&checkpoint, started_at) {
+                                self.record_terminal_outcome(issue, false);
+                            } else {
+                                let insta_crash = duration_sec
+                                    < self.quarantine_config.insta_crash_secs
+                                    && exit_code != Some(0);
+                                self.record_terminal_outcome(issue, insta_crash);
+                            }
                         } else {
                             // Orphaned-claim recovery (issue #3823b): a
                             // daemon-owned sweep that exits cleanly WITHOUT a
@@ -3978,6 +3994,38 @@ fn read_checkpoint_phase(path: &Path) -> Option<String> {
         .map(ToString::to_string)
 }
 
+/// Returns `true` when the checkpoint at `path` was last written at or after
+/// `started_at` — i.e. by the sweep run that began at `started_at` — rather than
+/// being a stale artifact left on disk by an earlier dispatch (#4009).
+///
+/// Sweep checkpoints persist across dispatches (`.loom/sweep-checkpoint/
+/// issue-<N>.json` is only removed by an explicit `sweep-checkpoint.sh delete`,
+/// which never runs on a crash — #3373), so the mere *presence* of the file
+/// does not prove the run that just died made any progress. A single
+/// successful-enough historical run would otherwise leave the file on disk
+/// forever, permanently exempting the issue from the insta-crash quarantine
+/// (#3939) even as every subsequent dispatch dies pre-work in under 2s — an
+/// infinite re-dispatch loop.
+///
+/// Comparing the file's mtime against this run's `started_at` distinguishes
+/// "this run reached real work" (a mid-build death — the #3895 watchdog's
+/// remit, which must reset the insta-crash tally) from "a checkpoint from an
+/// earlier dispatch happens to exist" (a pre-work insta-crash that must still
+/// count toward quarantine).
+///
+/// A missing file, or an unreadable/absent mtime, yields `false` (treated as
+/// "no progress by this run"), so an unreadable checkpoint never shields an
+/// issue from quarantine.
+fn checkpoint_written_by_run(path: &Path, started_at: DateTime<Utc>) -> bool {
+    let Ok(meta) = std::fs::metadata(path) else {
+        return false;
+    };
+    let Ok(mtime) = meta.modified() else {
+        return false;
+    };
+    DateTime::<Utc>::from(mtime) >= started_at
+}
+
 /// Send a signal to a PID. Returns `true` on success (signal queued or
 /// process already absent and the caller can treat that as "done"). PID
 /// 0 is rejected to avoid the POSIX broadcast-to-group semantics.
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review — approved

Reviewed against the 4 acceptance criteria in #4009.

**Correctness**
- The new `checkpoint_written_by_run(path, started_at)` helper (mtime `>=` `started_at`) correctly distinguishes "this run wrote/updated the checkpoint" from "a checkpoint file merely exists on disk." Both the daemon process and the checkpoint file live on the same host/filesystem (`.loom/sweep-checkpoint/` under `workspace_root`), so there's no cross-host clock-skew concern here. The `>=` boundary treats an exact-tie write as progress — a reasonable, explicitly documented conservative choice for an inherently ambiguous race, not a defect.
- The stale-checkpoint fallback branch (lines ~2617–2621) reproduces the *exact* `duration_sec < insta_crash_secs && exit_code != Some(0)` test used in the checkpoint-less branch below it — verified byte-for-byte identical logic. Satisfies AC #1 (pre-work death classified as insta-crash regardless of checkpoint presence). Minor nit, non-blocking: this duplicates rather than factors into a shared helper, but it's a 3-line boolean and not worth blocking on.
- AC #2 (distinguish "this run made progress" from "file exists") is met via the mtime approach — the issue's own preferred option.
- AC #3 (genuine mid-build deaths still reset the tally, #3895's remit) verified: `reaper_checkpoint_death_does_not_count_as_insta_crash` was correctly updated to write the checkpoint *during* each run (via `insert_dead_running_at` with a past `started_at`) so it genuinely models a mid-build death post-dating `started_at`, and still asserts `insta_crash_count == 0` / not quarantined.
- I also checked the startup lock-reclaim recovery path (`daemon_owned_dead` / checkpoint-only admission, ~line 3443) — it inserts `SweepState::Crashed` entries directly and never calls `record_terminal_outcome`, so it's outside the scope of this fix and unaffected either way.

**Test quality (AC #4)** — confirmed the new `reaper_stale_checkpoint_death_counts_as_insta_crash` drives the actual reaper path, not `record_terminal_outcome` directly: it writes a checkpoint file, records its real mtime, then for 3 iterations calls `insert_dead_running_at(..., stale_written_at + 5s)` + `registry.reap_once()`, asserting the tally reaches 3 and `is_quarantined(45)` is true. This is exactly what AC #4 asked for and is a faithful regression test for the #4009 symptom (issue #4009 itself crash-looped 3x with a stale checkpoint present).

**Verification performed**
- Checked out the PR into a scratch worktree and ran the suite myself (not just trusting the PR description):
  - `cargo test --lib reaper_` → 7/7 pass, including both touched tests.
  - `cargo test --lib` (full) → **1001 passed, 0 failed** — matches the PR description exactly.
  - `cargo clippy --lib --all-features -- -D warnings` → clean, no warnings.
- CI: `gh pr checks 4066` — Rust Linting & Build Check: pass, Rust Unit Tests: pass, Detect Changes: pass; Installer Integration Tests / MCP Server Build: skipping (expected, paths-filtered — this PR only touches `loom-daemon/src/sweep_registry.rs`).
- `gh pr view --json mergeStateStatus,mergeable` → `CLEAN` / `MERGEABLE`.

No regression to the #3895 mid-build-death watchdog — its owning test case was retained and updated to still model the case it's meant to cover.

Approving.

PR #4067, opened 2026-07-27
fix(daemon): honor LOOM_DAEMON_LOG override so test daemons don't pollute the operator's daemon.log

### LINKED ISSUE #4010: setup_logging() hardcodes $HOME/.loom/daemon.log with no override — test daemons pollute the operator's production log
## Summary

`setup_logging()` hardcodes the daemon log to `$HOME/.loom/daemon.log` with no override, so **any** `loom-daemon` process — including one spawned by an integration test — writes into the operator's production log. The socket path resolved a few hundred lines earlier in the same file *does* have a test override, which makes the omission look accidental rather than deliberate.

Measured on this host: 5,013 of 45,524 lines (a 5MB file) were tmux/test-terminal noise, densest exactly during a gate-loop incident — i.e. the pollution was worst at the moment the log was most needed.

## Detail

`loom-daemon/src/main.rs:796`:

```rust
fn setup_logging() -> Result<()> {
    let log_path = dirs::home_dir()
        .ok_or_else(|| anyhow!("No home directory"))?
        .join(".loom/daemon.log");
    ...
}
```

No env var, no derivation from the resolved loom dir. Compare `main.rs:276`, which already establishes the pattern:

```rust
// For testing, allow override via LOOM_SOCKET_PATH env var
let (loom_dir, socket_path) = if let Ok(path) = std::env::var("LOOM_SOCKET_PATH") { ... }
```

A test that sets `LOOM_SOCKET_PATH` to a temp dir correctly isolates its IPC surface and then logs into `~/.loom/daemon.log` anyway. Rotation (10MB × 10 files, `main.rs:806`) makes this worse rather than better: enough test noise can rotate real operator history out.

## Acceptance criteria

1. The log path honours an override — either a dedicated `LOOM_LOG_PATH` / `LOOM_DAEMON_LOG`, or (preferred) derivation from the same resolved loom dir that `LOOM_SOCKET_PATH` already establishes, so one env var isolates a test daemon completely.
2. Precedence follows the ecosystem contract: **env > config > default**, with the default byte-for-byte `$HOME/.loom/daemon.log`.
3. Integration tests that spawn a real daemon set the override, so a full test run leaves `~/.loom/daemon.log` untouched.
4. Document, in `.loom/docs/daemon-reference.md`, that the launchd `StandardOutPath` / `StandardErrorPath` redirects (`~/.loom/daemon-launchd-stdout.log` / `-stderr.log`) stay **0 bytes** — all real logging goes through `env_logger` to `daemon.log`. This has repeatedly misled operators into concluding a running daemon was silent or dead.

## Notes

Deriving from the resolved loom dir rather than adding a second independent env var is the better shape: it means `LOOM_SOCKET_PATH`-style isolation covers logs automatically and there is one less knob to forget in a new test.

---

## Curator Enhancement (2026-07-27, verified against `origin/main` @ a68099d5)

The original diagnosis is correct. Three corrections and one ordering constraint the Builder must know before starting.

### Correction 1 — line references have drifted

| Original issue says | Actual on `origin/main` |
|---|---|
| `main.rs:796` — `setup_logging()` | **`loom-daemon/src/main.rs:941`** |
| `main.rs:276` — `LOOM_SOCKET_PATH` override | **`loom-daemon/src/main.rs:342-343`** |
| `main.rs:806` — rotation | **`loom-daemon/src/main.rs:951`** |

### Correction 2 — the ordering constraint (the actual implementation blocker)

`setup_logging()` is **called at `main.rs:336`, seven lines *before* `loom_dir` is resolved at `main.rs:342`**:

```rust
// main.rs:335-336
// Setup logging to ~/.loom/daemon.log
setup_logging()?;                       // <-- runs FIRST

// main.rs:342-343
// For testing, allow override via LOOM_SOCKET_PATH env var
let (loom_dir, socket_path) = if let Ok(path) = std::env::var("LOOM_SOCKET_PATH") { ... };
```

So the issue's preferred shape — "derive from the same resolved loom dir" — **cannot be done by simply reading `loom_dir`**; it does not exist yet at the call site. Reordering `setup_logging()` below the resolution is the wrong fix (it would silently drop log lines emitted by `check_tmux_installed()` at `main.rs:339` and anything else that moves above it).

**Recommended shape**: extract a pure resolver and call it from both places.

There is already a third copy of this logic to fold in — `resolve_socket_path()` at **`main.rs:1189-1200`** duplicates the exact same `LOOM_SOCKET_PATH`-else-`$HOME/.loom` branch for the CLI path. A single `fn resolve_loom_dir() -> Result<PathBuf>` used by `setup_logging()`, `main()`, and `resolve_socket_path()` removes the duplication *and* fixes the bug in one move.

```rust
/// Resolve the daemon's loom dir: the parent of `LOOM_SOCKET_PATH` when set
/// (test isolation), else `$HOME/.loom`. Pure — no side effects, unit-testable.
fn resolve_loom_dir() -> Result<PathBuf> { ... }

fn resolve_log_path() -> Result<PathBuf> {
    if let Ok(p) = std::env::var("LOOM_DAEMON_LOG") { return Ok(PathBuf::from(p)); }
    Ok(resolve_loom_dir()?.join("daemon.log"))
}
```

Keep `resolve_log_path()` separate from `setup_logging()`: `setup_logging()` calls `env_logger::Builder::…init()`, which panics on a second call in the same process, so it can never be unit-tested directly. The resolver can.

### Correction 3 — AC4 names paths that do not exist in this repo

AC4 asks to document that `~/.loom/daemon-launchd-stdout.log` / `-stderr.log` stay 0 bytes. **Those paths appear nowhere in the repo.** Current `origin/main` points *both* redirects at a single shared file:

```bash
# defaults/scripts/cli/loom-daemon-start.sh:186-187
printf '    <key>StandardOutPath</key>\n    <string>%s</string>\n'   "$(xml_escape "$log_path")"
printf '    <key>StandardErrorPath</key>\n    <string>%s</string>\n' "$(xml_escape "$log_path")"
# ...where log_path == $START_LOG == "$REPO_ROOT/.loom/logs/daemon-start.log"  (line 237)
```

The split `daemon-launchd-{stdout,stderr}.log` pair exists on the reporter's host because their **installed** `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist` predates (or diverges from) the current script. The underlying observation still holds — verified on this host, `.loom/logs/daemon-start.log` is **0 bytes** while `~/.loom/daemon.log` is 6.3 MB — but the doc must name the path the repo actually generates, not the stale one. **Do not add `daemon-launchd-*.log` to the docs.**

### Recommendation on the config tier (AC2)

AC2 asks for full `env > config > default`. Recommend **env > default only**, and say so in the PR:

`autonomous.*` config lives in the workspace's `.loom/config.json`, which is resolved from `sweep_workspace` — established *long after* `setup_logging()` at `main.rs:336`. Wiring a config tier for the log path means either bootstrapping workspace resolution before logging (large blast radius, and log lines from that resolution would be lost) or logging to one file then re-pointing (`env_logger` cannot re-target after `init()`). Existing readers like `read_role_runner_config` / `read_watch_monitor_config` all run after workspace resolution, so they are not a usable precedent here.

Env + derivation satisfies the actual goal (test isolation, operator override) at a fraction of the cost. If a reviewer insists on the config tier, file it as a follow-up rather than expanding this PR.

### Why AC3 needs no test changes

Both daemon-spawning test sites already set `LOOM_SOCKET_PATH` to a `TempDir`:

- `loom-daemon/tests/common/mod.rs:53-57` — `TestDaemon::start()`
- `loom-daemon/tests/integration_singleton_guard.rs:33-37` — `spawn_daemon_and_wait()`

If the log path is derived from the socket path's parent, **both are isolated with zero edits to either file** — which is exactly the argument the issue's Notes section makes for derivation over a second independent env var. AC3 becomes a verification step, not a code change. (Still assert it — see Test Plan.)

## Affected Files

- `loom-daemon/src/main.rs` — extract `resolve_loom_dir()` / `resolve_log_path()`; use in `setup_logging()` (:941), `main()` (:342-343), and `resolve_socket_path()` (:1189-1200). Add `#[cfg(test)]` unit tests for the resolvers.
- `.loom/docs/daemon-reference.md` — document the override + the launchd-redirect-vs-`daemon.log` distinction. Add under **`## Operability — config, start/stop, E2E (Phase D, #3813)`** (line 1204), which already holds the env-var/config tables.
- `defaults/CLAUDE.md` — the `tail -f ~/.loom/daemon.log` reference at line 1810 should note the override exists.

Reference only, **no changes expected**:
- `loom-daemon/src/lib.rs:54` — `rotate_log_file()` already takes a path parameter and is covered by 6 unit tests (`lib.rs:130-195`). It needs nothing.
- `loom-daemon/tests/common/mod.rs`, `loom-daemon/tests/integration_singleton_guard.rs` — isolated for free via `LOOM_SOCKET_PATH` (see above).
- `defaults/scripts/cli/loom-daemon-start.sh` — the plist redirect is correct as-is; this issue documents it, it does not change it.

## Test Plan

- [ ] **Unit**: `resolve_loom_dir()` returns `$HOME/.loom` with no env set, and the socket path's parent when `LOOM_SOCKET_PATH` is set. Use a serialized/env-guarded test — these mutate process env.
- [ ] **Unit**: `resolve_log_path()` honours `LOOM_DAEMON_LOG` over the derived path; default is byte-for-byte `$HOME/.loom/daemon.log` (AC2's default clause).
- [ ] **Regression (AC3)**: record `~/.loom/daemon.log` size + mtime, run `cargo test -p loom-daemon --test '*'`, assert both unchanged. This is the acceptance test for the whole issue — it must fail on `main` and pass after the fix.
- [ ] **Manual**: `LOOM_SOCKET_PATH=/tmp/lt/d.sock ./target/release/loom-daemon` writes `/tmp/lt/daemon.log` and leaves `~/.loom/daemon.log` untouched.
- [ ] **Manual**: bare `loom-daemon-start.sh` still logs to `~/.loom/daemon.log` (no behavior change for operators).
- [ ] **Edge**: `LOOM_SOCKET_PATH` with no parent directory — `main.rs:346-348` already errors on this; `resolve_loom_dir()` must not diverge from that behavior.
- [ ] **Edge**: `LOOM_DAEMON_LOG` pointing into a non-existent directory — `setup_logging()` already `create_dir_all`s the parent (`main.rs:946-948`); confirm that still applies to the overridden path.
- [ ] **Docs**: `.loom/docs/daemon-reference.md` names `.loom/logs/daemon-start.log` (the repo's actual redirect target), **not** `daemon-launchd-*.log`.

## Scope guard

Single-concern change: ~1 source file + 2 doc files, well under the 400 LOC / 8 file threshold. Explicitly **out of scope** — rotation policy changes, a config tier for the log path (see above; file a follow-up if wanted), regenerating or migrating anyone's installed launchd plist, and the unrelated `.loom/.daemon.log` dev-mode log described in `docs/guides/daemon-dev-mode.md`.

rjwalters:
Closes #4010

## Summary

`setup_logging()` hardcoded the daemon log to `$HOME/.loom/daemon.log` with no override, so any `loom-daemon` process — including one spawned by an integration test that already isolates its IPC socket via `LOOM_SOCKET_PATH` — wrote into the operator's production log, polluting real operator history.

## Fix

- Extracted `resolve_loom_dir() -> Result<PathBuf>`: a pure resolver returning the parent of `LOOM_SOCKET_PATH` when set (test isolation), else `$HOME/.loom`. No side effects, unit-testable.
- Added `resolve_log_path() -> Result<PathBuf>`: honors `LOOM_DAEMON_LOG` (full override) first, else derives `<loom dir>/daemon.log` from `resolve_loom_dir()`. Default is byte-for-byte `$HOME/.loom/daemon.log` — no behavior change for operators who don't set the override.
- Wired `resolve_log_path()` into `setup_logging()`. `setup_logging()` runs at `main.rs:336`, *before* `main()`'s own `loom_dir`/`socket_path` resolution block (`:342-343`) — so `resolve_loom_dir()` is self-contained (reads `LOOM_SOCKET_PATH` itself) rather than reusing a not-yet-existing variable. `setup_logging()`'s call site is unchanged relative to `check_tmux_installed()` — no log lines dropped.
- Folded the CLI path's third copy of the same `LOOM_SOCKET_PATH`-else-`$HOME/.loom` branch (`resolve_socket_path()`, ~`main.rs:1189-1200`) into `resolve_loom_dir()` too, removing the duplication.
- Precedence is **env > default only** — no `autonomous.*` config tier, per the curator's analysis on the issue: `setup_logging()` runs before workspace/config resolution, so wiring a config tier in would require restructuring startup order (and `env_logger` can't be re-targeted after `.init()`) for marginal benefit. Filed as a possible follow-up if wanted, not pursued here.
- Kept `create_dir_all` on the log path's parent (so `LOOM_DAEMON_LOG` pointing into a non-existent directory still works) and the existing 10MB × 10-file rotation logic (`rotate_log_file()` in `lib.rs`) completely unchanged.

## Tests

- 5 new `#[cfg(test)]` unit tests (`resolve_paths_tests` module, `#[serial]`-guarded since they mutate process env, following the crate's existing `dispatch_tests` pattern):
  - `resolve_loom_dir` defaults to `$HOME/.loom` with no env set, and to the socket path's parent when `LOOM_SOCKET_PATH` is set.
  - `resolve_log_path` defaults to `$HOME/.loom/daemon.log`, honors `LOOM_DAEMON_LOG`, and confirms `LOOM_DAEMON_LOG` wins over `LOOM_SOCKET_PATH`-derived precedence.
- `setup_logging()` itself is deliberately NOT unit-tested (it calls `env_logger::Builder::...init()`, which panics on a second call in-process) — this is exactly why the pure resolvers were split out.
- `cargo build -p loom-daemon` and `cargo clippy -p loom-daemon --all-targets` both clean.
- `cargo test -p loom-daemon` — all unit tests pass including the 5 new ones; pre-existing `integration_security` flakes (`test_accept_valid_ids`, `test_reject_directory_traversal`, `test_terminal_isolation_between_clients` — different subsets fail on repeated runs) reproduce independent of this change, caused by tmux-session contention from several other worktrees running `cargo test` concurrently on this host during this session — unrelated to `main.rs` logging/path resolution.
- **Regression check (AC3, the acceptance test for this issue)**: manually spawned `./target/debug/loom-daemon` with `LOOM_SOCKET_PATH=/tmp/.../d.sock` (no `LOOM_DAEMON_LOG`) — confirmed a `daemon.log` was created in `/tmp/.../` while `~/.loom/daemon.log`'s size and mtime were byte-for-byte unchanged before/after. Also confirmed `LOOM_DAEMON_LOG` wins over the `LOOM_SOCKET_PATH`-derived path when both are set.
- `cargo fmt -p loom-daemon -- --check` clean.

## Docs

- `.loom/docs/daemon-reference.md`: new "Daemon log path override (`LOOM_DAEMON_LOG`, #4010)" subsection under §Operability, documenting the env var, precedence, and why a config tier is out of scope — plus the launchd `StandardOutPath`/`StandardErrorPath` (`.loom/logs/daemon-start.log`) vs. `daemon.log` distinction (stays 0 bytes in normal operation; all real logging goes through `env_logger`).
- `defaults/CLAUDE.md`: noted the `LOOM_DAEMON_LOG` override next to the existing `tail -f ~/.loom/daemon.log` troubleshooting snippet.

## Scope

Single-concern change: 1 source file (`loom-daemon/src/main.rs`) + 2 doc files, 165 insertions / 14 deletions across 3 files — well under the 400 LOC / 8 file threshold. No rotation-policy changes, no config tier, no launchd plist regeneration/migration, and the unrelated `.loom/.daemon.log` dev-mode log is untouched.

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1330,6 +1330,48 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.watchdog.reviewStall` | `LOOM_SWEEP_REVIEW_STALL` | `true` | Review-phase stall watchdog on/off (#3910) |
 | `autonomous.watchdog.reviewStallTimeoutSecs` | `LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS` | `2700` | Log-silence window before a hung Judge/Doctor sweep is re-dispatched |
 
+### Daemon log path override (`LOOM_DAEMON_LOG`, #4010)
+
+`setup_logging()` writes every daemon log line to `$HOME/.loom/daemon.log` by
+default — this is the destination for `env_logger`-routed output (see the
+launchd redirect note below), NOT the on-disk `LOOM_SOCKET_PATH` used by IPC.
+Before #4010, this path was hardcoded with no override at all, so **any**
+`loom-daemon` process — including one spawned by an integration test that
+already isolates its IPC socket via `LOOM_SOCKET_PATH` — still wrote into the
+operator's production log, polluting real operator history (worst-case: enough
+test noise to rotate genuine history out of the 10MB × 10-file window).
+
+Two env vars now cover both the log path and the socket path:
+
+| Env var | Purpose | Default |
+|---------|---------|---------|
+| `LOOM_DAEMON_LOG` | Full override of the daemon log file path | `<loom dir>/daemon.log` |
+| `LOOM_SOCKET_PATH` | Full override of the daemon's IPC socket path; also implicitly derives the log path (below) when `LOOM_DAEMON_LOG` is unset | `<loom dir>/loom-daemon.sock` |
+
+Precedence is **env > default only** (no `autonomous.*` config tier) — `resolve_log_path()`
+checks `LOOM_DAEMON_LOG` first, then falls back to `resolve_loom_dir().join("daemon.log")`,
+where `resolve_loom_dir()` is the parent directory of `LOOM_SOCKET_PATH` when
+that env var is set, else `$HOME/.loom`. Practically: setting `LOOM_SOCKET_PATH`
+to a tempdir (as both `loom-daemon/tests/common/mod.rs` and
+`loom-daemon/tests/integration_singleton_guard.rs` already do to isolate IPC)
+isolates the log file for free, with zero test-file edits. A config tier was
+considered and deliberately dropped: `setup_logging()` runs at daemon startup
+*before* workspace/config resolution happens, so wiring config in would require
+restructuring startup order (and `env_logger` cannot be re-targeted after
+`.init()`) for marginal benefit over the env-only surface — file a follow-up if
+a config tier is wanted later.
+
+**The launchd `StandardOutPath`/`StandardErrorPath` redirect is a decoy, not a
+second log.** `defaults/scripts/cli/loom-daemon-start.sh` points both launchd
+redirects at a single file, `$REPO_ROOT/.loom/logs/daemon-start.log`
+(`$START_LOG`), which only ever captures output emitted *before*
+`setup_logging()` installs the `env_logger` pipe target (plus a crash's raw
+stderr, if one occurs before logging is set up). In normal operation this file
+stays **0 bytes** — all real daemon logging goes through `env_logger` straight
+to `daemon.log` (or wherever `LOOM_DAEMON_LOG`/`LOOM_SOCKET_PATH` points it).
+This has repeatedly misled operators into concluding a running daemon was
+silent or dead when they tailed `daemon-start.log` instead of `daemon.log`.
+
 **Autonomous dispatch model (`autonomous.model`, #3944).** A daemon-dispatched
 child is a headless `claude -p "/loom:sweep N"` process. Without an explicit
 `--model`, it inherits whatever model the operator last configured in their
--- a/defaults/CLAUDE.md
+++ b/defaults/CLAUDE.md
@@ -1827,6 +1827,11 @@ gh label sync --file .github/labels.yml
 tail -f ~/.loom/daemon.log
 ```
 
+The log path can be overridden with `LOOM_DAEMON_LOG` (e.g. to isolate a test
+daemon's log from the operator's production history); an unset override falls
+back to `$HOME/.loom/daemon.log` unchanged. See "Daemon log path override" in
+`.loom/docs/daemon-reference.md`.
+
 **Claude Code not found**:
 ```bash
 # Ensure Claude Code CLI is in PATH
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -343,15 +343,10 @@ async fn main() -> Result<()> {
     let (loom_dir, socket_path) = if let Ok(path) = std::env::var("LOOM_SOCKET_PATH") {
         // For testing, use the parent directory of the provided socket path
         let socket_path = std::path::PathBuf::from(path);
-        let loom_dir = socket_path
-            .parent()
-            .ok_or_else(|| anyhow!("Socket path has no parent directory"))?
-            .to_path_buf();
+        let loom_dir = resolve_loom_dir()?;
         (loom_dir, socket_path)
     } else {
-        let loom_dir = dirs::home_dir()
-            .ok_or_else(|| anyhow!("No home directory"))?
-            .join(".loom");
+        let loom_dir = resolve_loom_dir()?;
         fs::create_dir_all(&loom_dir)?;
         let socket_path = loom_dir.join("loom-daemon.sock");
         (loom_dir, socket_path)
@@ -938,10 +933,43 @@ fn check_tmux_installed() -> Result<()> {
         .ok_or_else(|| anyhow!("tmux not installed. Install with: brew install tmux"))
 }
 
-fn setup_logging() -> Result<()> {
-    let log_path = dirs::home_dir()
+/// Resolve the daemon's loom directory: the parent of `LOOM_SOCKET_PATH` when
+/// that env var is set (test isolation), else `$HOME/.loom`. Pure — no side
+/// effects (no directory creation), so it's safe to call from `setup_logging()`
+/// (which runs before `main()`'s own `loom_dir`/`socket_path` resolution block)
+/// as well as from `main()` and `resolve_socket_path()` without duplicating the
+/// branching logic three times over.
+fn resolve_loom_dir() -> Result<PathBuf> {
+    if let Ok(path) = std::env::var("LOOM_SOCKET_PATH") {
+        let socket_path = PathBuf::from(path);
+        let loom_dir = socket_path
+            .parent()
+            .ok_or_else(|| anyhow!("Socket path has no parent directory"))?
+            .to_path_buf();
+        return Ok(loom_dir);
+    }
+    let loom_dir = dirs::home_dir()
         .ok_or_else(|| anyhow!("No home directory"))?
-        .join(".loom/daemon.log");
+        .join(".loom");
+    Ok(loom_dir)
+}
+
+/// Resolve the daemon log file path: `LOOM_DAEMON_LOG` (full override) when
+/// set, else `<loom dir>/daemon.log` derived from [`resolve_loom_dir`]. This
+/// means `LOOM_SOCKET_PATH`-style test isolation covers the log file for free
+/// — a test daemon pointed at a tempdir socket also logs into that tempdir,
+/// never into the operator's `~/.loom/daemon.log` (Issue #4010). Precedence is
+/// env > default only; see #4010 for why a config tier is out of scope here
+/// (`setup_logging()` runs before workspace/config resolution).
+fn resolve_log_path() -> Result<PathBuf> {
+    if let Ok(path) = std::env::var("LOOM_DAEMON_LOG") {
+        return Ok(PathBuf::from(path));
+    }
+    Ok(resolve_loom_dir()?.join("daemon.log"))
+}
+
+fn setup_logging() -> Result<()> {
+    let log_path = resolve_log_path()?;
 
     if let Some(parent) = log_path.parent() {
         fs::create_dir_all(parent)?;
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Independently verified against issue #4010's acceptance criteria:

**Diff matches the description exactly** (`loom-daemon/src/main.rs` +118/-14, 2 doc files +47):
- `resolve_loom_dir()` extracted as a pure, self-contained resolver (re-reads `LOOM_SOCKET_PATH` itself rather than depending on `main()`'s not-yet-computed variable) — correct given `setup_logging()` runs at line ~336, before `main()`'s own `loom_dir`/`socket_path` block at ~342.
- `resolve_log_path()` added: `LOOM_DAEMON_LOG` first, else `resolve_loom_dir().join("daemon.log")`. Confirmed default is byte-for-byte `$HOME/.loom/daemon.log` when no override is set (verified both by reading the code and by the new `resolve_log_path_defaults_to_home_loom_daemon_log` test).
- `setup_logging()`'s call site relative to `check_tmux_installed()` is unchanged — read the `main()` body directly (lines ~336-338): `setup_logging()?;` then `check_tmux_installed()?;`, same order as before.
- `resolve_socket_path()`'s duplicate branch folded into `resolve_loom_dir()` cleanly — same behavior, same error message ("No home directory" / "Socket path has no parent directory"), all 4 callers unaffected.
- 5 new `#[serial]` unit tests for both resolvers, covering default + override + precedence-when-both-set. All pass locally.
- Docs: `daemon-reference.md` and `defaults/CLAUDE.md` both updated correctly. Confirmed the launchd redirect note correctly references `.loom/logs/daemon-start.log` (the real path `loom-daemon-start.sh` uses) — no mention of the nonexistent `daemon-launchd-stdout.log`/`-stderr.log`.

**Build/test verification:**
- `cargo build -p loom-daemon`, `cargo clippy -p loom-daemon --all-targets`, `cargo fmt -p loom-daemon -- --check` — all clean.
- `cargo test -p loom-daemon --bin loom-daemon resolve_paths_tests` — all 5 new tests pass.
- Ran the full `cargo test -p loom-daemon` suite twice locally, plus an isolated `--test integration_security` run. Result: unit tests (1000 + 15) pass consistently every time; `integration_basic`/`integration_security` (tmux/socket-spawning tests) each failed exactly one *different* test per run (`test_multiple_clients`+`test_destroy_terminal` → `test_create_terminal_with_working_dir` → `test_reject_injection_backticks`, the last failing on "Daemon failed to create socket within 5s"). I independently confirmed real concurrent contention on this host during the runs — `ps aux` showed a separate worktree (`issue-4031`) running `cargo test --release --workspace` at the same time. This is the exact host-contention signature described (different failing subset each run, socket/tmux timeouts, unrelated to `main.rs` path-resolution logic), not a regression introduced by this PR. GitHub Actions' "Rust Unit Tests" job (isolated runner) passed cleanly, corroborating this.

**CI/merge state:** all required checks pass (`Rust Unit Tests`, `Rust Linting & Build Check`, `Installer Integration Tests`, `Detect Changes`; `MCP Server Build` legitimately skipped — no MCP changes). `mergeStateStatus: CLEAN`.

**Scope:** 3 files changed (1 source + 2 docs), 165 insertions / 14 deletions — matches the issue's scope guard.

Nice, focused refactor — clean separation of the pure resolvers from the `env_logger` side-effecting call, and the dedup of the socket-path branch is a nice bonus per the issue's scope.

Issue #3930, opened 2026-07-24
One global budget + per-repo isolation + per-repo status breakdown across managed repos (#3926 phase d)


rjwalters:
Phase (d) — **final phase** of #3926/#3835 (multi-repo `loom-daemon`). Phases (a)-(c) already shipped on `main`:

- **Phase 1** (#3926, PR #3931) — machine-level `WorkspaceRegistry` (`~/.loom/workspaces.json`) + CLI/IPC `workspace add|remove|list`.
- **Phase b** (#3928, PR #3932) — `WorkspacePool` (one `SweepRegistry` per registered root) + `work_finder::tick_multi` / `epic_supervisor::tick_multi` fan-out with a **single global concurrency budget** and per-workspace error isolation.
- **Phase c** (#3929, PR #3933) — `(repo, issue)`-aware IPC/MCP (`workspace_root` param on the five sweep tools), `SweepInfo.repo` + event payload `repo`, `WorkspacePool::evict()` on `workspace remove`. PR #3933's own body states: *"Phase d (#3930: global budget breakdown, per-repo status view, per-repo health gate) is intentionally out of scope."*

**This re-curation (2026-07-24) narrows the original 4-bullet AC list down to what phases b/c actually left undone**, verified against current `main` (commit `7214beb7`). Two of the four original bullets are already satisfied by phase b; the remaining two are the real scope of this issue.

## Already done (do not re-implement — verified against `main`)

1. **Global concurrency cap + capacity backpressure, summed in-flight across repos** — `loom-daemon/src/work_finder.rs::tick_multi` (line 342) computes `occupancy` as `in_flights.iter().map(HashSet::len).sum()` **before** any dispatch (line 352) and defers once `occupancy >= max_concurrent` regardless of which workspace still has ready items (lines 393-399) — a single shared budget, never per-repo. `spawn_multi_work_finder_task` (line 759) probes the three dynamic-cap inputs (token pool, ranking, disk) **once per tick from `fallback_root`** (lines 783-789) and passes the single resulting `max_concurrent` into `tick_multi` for every workspace. Per-workspace forge-poll errors are isolated (`report.errors += 1; ... continue;`, lines 373-379) — one bad repo does not stop the others. `epic_supervisor::tick_multi` (line 764) needs no analogous cap ("`deduplicates sweeps per child, so no global cap is needed`", line 763) but has the same per-workspace error isolation.
[...]

PR #4068, opened 2026-07-27
fix(daemon): distinguish pending/disabled/clear main-health-gate states (#4012)

### LINKED ISSUE #4012: Main-health gate reports 'clear (dispatch allowed)' before its first verdict — indistinguishable from verified-green
## Summary

Immediately after a daemon restart, `loom-daemon status` prints

```
Main-health gate: clear (dispatch allowed)
```

before any gate run has completed. That string is **identical** to the one printed after a gate run has verified main green. An operator cannot tell "the gate has confirmed main is healthy" from "the gate has never run in this process."

This matters concretely after a roll: loom's gate command takes ≥10 minutes, so there is a long window where a fresh `clear` proves nothing — and reading it as verified-green is the natural mistake. It cost time during the 2026-07-26 roll.

## Detail

Gate status is derived from two booleans (`loom-daemon/src/main.rs:1862`):

```rust
fn format_gate_status(halted: bool, not_evaluated: bool, reason: Option<&str>) -> String {
    match (halted, not_evaluated) {
        (true,  true)  => "HALTED ... + NOT EVALUATED ...",
        (true,  false) => "HALTED (main verified red ...)",
        (false, true)  => "not evaluated ({cause}) ...",
        (false, false) => "clear (dispatch allowed)".to_string(),
    }
}
```

A fresh `MainHealthState::new()` (`main_health_gate.rs:218`) initialises `halted: false, unevaluated: false` — deliberately, so that a daemon with the gate *disabled* never halts. But that same `(false, false)` is what a verified-green tick produces, so the pre-first-verdict state is indistinguishable from success. Two booleans simply cannot encode three states.

#3962 split *halted* from *not-evaluated*, and #3974 added the cause string, but neither introduced a "no verdict yet" state.

Note this is genuinely ambiguous in three ways, not two — `(false, false)` currently means any of:
- gate disabled (nothing will ever evaluate it),
- gate enabled, first tick not yet complete,
- gate enabled, last tick verified green.

## Acceptance criteria

1. `loom-daemon status` distinguishes pending-first-verdict from verified-green. Suggested wording: `pending (no verdict yet this process — dispatch allowed)` vs the existing `clear (dispatch allowed)`.
2. The gate-disabled case is also distinguishable — `disabled (gate not enabled; dispatch allowed)` — rather than collapsing into either of the above.
3. Dispatch behaviour is unchanged: pending and disabled both still allow dispatch. This is an observability fix only; no new halt path.
4. Where available, show the timestamp of the last completed verdict, so `clear` carries its own recency evidence.

## Notes

Narrow gap, low blast radius, but it directly undermines the post-roll verification loop — the one moment an operator most needs the gate line to mean something. Cheap to fix: replace the boolean pair with a small enum at the reporting boundary.

---

Verified against `origin/main`. Line refs in **Detail** above corrected: `format_gate_status` is at `main.rs:1862` (was cited as 1558), `MainHealthState::new()` at `main_health_gate.rs:218` (was 187).

### Scope correction: AC4 is not a reporting-boundary change

The **Notes** section says this is "cheap to fix: replace the boolean pair with a small enum at the reporting boundary." That holds for AC1–AC3, but **not AC4**. There is **no verdict timestamp anywhere in the daemon** — grepping `SystemTime|chrono|timestamp|evaluated_at|last_verdict` across `main_health_gate.rs` returns only an unrelated doc comment at `:1320`. The only time-bearing fields are `UnevaluatedTrack::last_warn` (`:207`) and `GateMemo::backoff_until` (`:194`), both monotonic `Instant`s used for log-throttling and retry backoff — neither is wall-clock, neither is serializable, and neither marks a completed verdict.

So AC4 requires a **new state field** plus wire plumbing, not just a renderer change. `chrono` is already a dependency with the `serde` feature (`loom-daemon/Cargo.toml:14`) and `types.rs` already uses `DateTime<Utc>` (`:605`, `:643`), so `Option<DateTime<Utc>>` is the idiomatic choice.

**AC4 design decision the builder must make explicitly:** `record_gate_evaluated_sha` is called from **two** places — after a real Green/Red verdict (`main_health_gate.rs:1364`) *and* on the no-real-change skip path where the gate command never ran (`:1349`). Setting the timestamp on the skip path would refresh "last verified" without anything having been verified; setting it only on `:1364` means the displayed time can lag while the verdict is still legitimately current. Recommend stamping only at `:1364` (honest "last actual run") and letting the SHA carry currency. State the choice in the PR body.

### Recommended design

Widen at the *state* layer, then render. A 3-value enum does not fit — there are **four** reportable conditions, and `halted`/`not_evaluated` remain orthogonal:

- `disabled` — gate not enabled for this root (dispatch allowed)
- `pending` — enabled, no verdict yet this process (dispatch allowed)
- `clear` — enabled, last run verified green (dispatch allowed)
- `halted` — verified red (dispatch paused)

**Pending signal:** `gate_last_evaluated_sha().is_none()` (`main_health_gate.rs:317`) is sound. Verified: `decide_gate_run` returns `Run` when `last_evaluated_sha == None` (`:1182-1184`), so a fresh process can never reach the skip path before producing a real verdict.

**Disabled signal:** compute it as `resolve_enabled(&read_autonomous_gate_config(root))` (`:2063`). Note a root can be *enabled yet effectively disabled* when `read_build_gate_config(root)` returns `None` (`:526`) — the gate loop treats that as always-green (`:2247`). A faithful `disabled` state should check **both**.

### Implementation traps

1. **Two render sites, not one.** `format_gate_status` (`main.rs:1862`) is *not* used by the per-repo table — `main.rs:2024` has a **duplicated inline `match`** producing the 13-char `GATE` column. Both must be widened or the table will still show `clear` for a pending/disabled repo. (`pending` and `disabled` both fit the existing `{:<13}` width.)

2. **The disabled path actively writes `(false, false)`.** This is not merely an unrepresented state — `main_health_gate.rs:2242-2244` calls `set_halted(false)` + `note_gate_tick(None, …)` for a disabled root, and `:2252-2254` does the same for "enabled but no usable buildGate". Both deliberately overwrite the pair with the same value a green tick produces. A new state cannot be derived from the existing booleans alone.

3. **Read the enablement flag daemon-side only.** `resolve_enabled` consults the `LOOM_MAIN_HEALTH_GATE` env var (`:100`, `:1990`). Computing it in `main.rs`'s renderer would read the **CLI client's** environment, not the daemon's, and silently misreport. Compute it inside `build_daemon_status` (`ipc.rs:457`), which already does per-root config reads in the same loop (e.g. `read_work_finder_config` at `ipc.rs:534`).

4. **New wire fields need `#[serde(default)]`.** Every post-original field in `DaemonStatusReport` (`types.rs:706`) and `RepoStatus` (`types.rs:810`) carries it with a doc line naming its issue. Prefer `Option<bool>` over `bool` for the new enabled/verdict flags: a legacy payload deserializing to `bool::default() == false` would report a healthy older daemon as **disabled**, whereas `None` honestly means "older daemon, unknown". This matches the existing `"pipeline": null` precedent in the same JSON block (`main.rs:1828-1830`), which already distinguishes "not requested" from "requested but empty".

5. **`WorkspaceHealthStates` needs matching delegates.** A never-seen root returns `false` from both `is_halted` (`:445`) and `is_unevaluated` (`:457`) via `is_some_and`. Any new pending/disabled accessor needs its own delegate with deliberate absent-root semantics (an unregistered root is arguably `pending`, not `clear`).

6. **Top-level vs per-repo scope differ.** `main_health_gate_*` reflects only `fallback_root`, the daemon's own primary workspace (`ipc.rs:576-579`); `per_repo[]` covers every managed root (`ipc.rs:494-499`). Preserve that split.

### Out of scope

- No change to dispatch behaviour (AC3): `pending` and `disabled` both still allow dispatch. No new halt path.
- No change to gate scheduling, backoff, or the `buildGate` runner.

## Affected Files

- `loom-daemon/src/main_health_gate.rs` — add verdict-timestamp field to `MainHealthState` (`:155`) + `new()` (`:218`); add accessor(s) for pending/verdict state; stamp the timestamp at `:1364`; add `WorkspaceHealthStates` delegates (`:411`)
- `loom-daemon/src/types.rs` — new `#[serde(default)]` fields on `DaemonStatusReport` (`:706`) and `RepoStatus` (`:810`); update the now-inaccurate doc comment at `:765` ("Always `false` when the gate loop is not enabled or has never run")
- `loom-daemon/src/ipc.rs` — populate new fields in `build_daemon_status` per-repo loop (`:494-499`) and top-level (`:576-579`); resolve gate-enabled per root here (daemon-side)
- `loom-daemon/src/main.rs` — widen `format_gate_status` (`:1862`) beyond the 2×2 `match`; widen the duplicated per-repo table `match` (`:2024`); add new keys to the JSON block (`:1809-1827`)

## Test Plan

- [ ] **Automated — renderer** (`loom-daemon/src/main.rs`): update `format_gate_status_covers_all_four_states` (`:2823`) — the name becomes wrong and it hard-asserts `assert_eq!(..., "clear (dispatch allowed)")` at `:2824`, which any signature change breaks. Update `format_gate_status_names_the_actual_not_evaluated_cause` (`:2801`). Add cases asserting `pending` and `disabled` render distinctly from `clear`, and that both still read as dispatch-allowed.
- [ ] **Automated — wire compat** (`loom-daemon/src/ipc.rs`): extend `test_daemon_status_backward_compat_missing_capacity` (`:2712`) to assert the new fields default sanely from a legacy payload (`None`, not a misleading `false`). Update `test_daemon_status_request_response_round_trip` (`:2639`).
- [ ] **Automated — producer** (`loom-daemon/src/ipc.rs`): update `test_build_daemon_status_reports_halt_and_in_flight` (`:2737`, `#[serial]`) and `test_build_daemon_status_multi_workspace_per_repo_breakdown` (`:2837`). Add a case: a fresh `MainHealthState` with the gate enabled reports **pending**, not clear.
- [ ] **Automated — state** (`loom-daemon/src/main_health_gate.rs`): add a test that a fresh state reports no verdict, and that it flips after a Green/Red outcome but **not** after the skip path at `:1349`. Check `test_default_state_not_halted` (`:2412`) and `test_workspace_health_states_unknown_root_not_halted` (`:2422`) still hold.
- [ ] **Manual — the actual reported symptom:** restart the daemon with the gate enabled and run `loom-daemon status` **immediately**. Must print `pending`, not `clear`. Wait for the first gate run to complete (≥10 min per the issue) and confirm it flips to `clear` with a verdict timestamp.
- [ ] **Manual — disabled:** with `LOOM_MAIN_HEALTH_GATE` unset and no `autonomous.mainHealthGate` block, confirm `status` reports `disabled`, not `clear`.
- [ ] **Manual — JSON:** `loom-daemon status --json` exposes the new state; confirm the per-repo `GATE` column stays aligned at width 13.
- [ ] **Edge case:** a root that is enabled but has no usable `buildGate` block (`main_health_gate.rs:2247`) — decide and assert whether it reports `disabled` or `pending`.
- [ ] `cargo test --workspace` passes; `cargo clippy` clean.

rjwalters:
## Summary

`loom-daemon status` printed `Main-health gate: clear (dispatch allowed)` for
three genuinely different conditions the pre-existing `halted`/`not_evaluated`
boolean pair could not distinguish:

1. the gate is **disabled** for this root (or enabled with no usable
   `buildGate` block — the gate loop treats that identically),
2. the gate is **enabled** but has not completed its first evaluation yet
   this daemon process ("pending"),
3. the gate's last completed run **verified `main` green** ("clear").

Two booleans can't encode three states, so `(halted=false, not_evaluated=false)`
collapsed all three into the same string — indistinguishable from a genuinely
verified-green `main`. This directly undermines the post-roll verification
loop: a fresh restart reads as "confirmed healthy" when nothing has run yet.

Closes #4012

## Design

Widened the reporting boundary with a `GateVerdict` enum
(`Disabled | Pending | Clear{since} | NotEvaluated{reason} | Halted{..}`)
built by a pure `classify_gate_verdict` function and rendered by both the
long-form summary line and the short 13-char per-repo table column, so the
two renderers can never disagree.

Two new signals feed the classifier, computed **daemon-side** (never the CLI
client's environment):

- **`main_health_gate_enabled: Option<bool>`** (`RepoStatus.health_gate_enabled`
  for per-repo) — via a new `main_health_gate::effective_enabled(repo_root)`
  helper that folds `resolve_enabled(..)` **and** a usable `buildGate` block
  into one check (a root nominally `enabled: true` with no `buildGate`
  command is effectively disabled — the gate loop already treats it as
  always-green). `Option<bool>` rather than `bool` so a legacy wire payload
  that never populated the field deserializes to `None` ("unknown, older
  daemon"), not `Some(false)` — a bare `bool`'s wire default would misreport
  a healthy older daemon as "gate disabled".
- **`main_health_gate_verdict_at: Option<DateTime<Utc>>`** (`RepoStatus.health_gate_verdict_at`
  for per-repo) — the wall-clock time of the most recent **completed**
  (Green/Red) verdict. Stamped only in `apply_gate_outcome`'s Green/Red arms;
  deliberately **not** touched by the `#3984` SHA-memo skip path in
  `run_gate_tick`, since a skip proves nothing new about `main` and stamping
  it there would let a stale "last verified" silently refresh itself
  forever. `None` is exactly the "pending" signal — a fresh process, before
  its first real verdict, can never reach the skip path first (`decide_gate_run`
  always returns `Run` when there's no prior determinate evaluation).

`halted`/`not_evaluated` still take strict priority over disabled/pending in
the classifier, so a genuinely active halt is never hidden behind either
newer state.

## Acceptance criteria

1. ✅ `pending (no verdict yet this process — dispatch allowed)` distinct from `clear (dispatch allowed; last verified green at <rfc3339>)`.
2. ✅ `disabled (gate not enabled; dispatch allowed)` distinct from both.
3. ✅ Dispatch behavior unchanged — pending and disabled both still allow dispatch; no new halt path (`classify_gate_verdict` never returns `Halted` for either).
4. ✅ `clear` carries its own recency evidence via the verdict timestamp, in both the human-readable line and the JSON output (`main_health_gate.verdict_at` / `per_repo[].health_gate_verdict_at`).

## Test plan

- `loom-daemon/src/main_health_gate.rs`: unit tests for `last_verdict_at` (stamped on Green/Red, not on Unevaluated or the SHA-memo skip path) and `effective_enabled` (no config / enabled-but-no-buildGate / both signals / autonomous-disabled-alone / env master switch).
- `loom-daemon/src/ipc.rs`: extended `test_daemon_status_backward_compat_missing_capacity` to assert the new fields default to `None` (not `Some(false)`) from a legacy payload; extended `test_build_daemon_status_reports_halt_and_in_flight` with a case where a fresh, newly-enabled root reports `pending`, not `clear`; extended the multi-workspace test with the disabled/no-verdict assertions; extended the round-trip test.
- `loom-daemon/src/main.rs`: rewrote the `format_gate_status` tests against the new `classify_gate_verdict`/`GateVerdict` API and added dedicated tests for pending-vs-clear, disabled-vs-pending, the "halted beats disabled/pending" priority rule, the legacy-`None`-is-not-disabled case, and the short-label/long-form agreement + 13-char width.
- `cargo build --workspace`, `cargo test --lib` (1012 passed), `cargo test --bin loom-daemon` (16 passed), `cargo clippy --package loom-daemon --package loom-api --all-targets -- -D warnings <ci flags>` (clean), `cargo fmt --check` (clean).

## Notes

Followed the Curator's re-verified design guidance in the issue body,
including stamping the verdict timestamp only at the real Green/Red site
(not the skip path) and computing the enabled signal daemon-side inside
`build_daemon_status` rather than in the CLI renderer.

### DIFF
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -497,6 +497,10 @@ pub fn build_daemon_status(
             // Name the actual failure class (#3974 AC2) rather than letting the
             // renderer assume "dirty tree" for every unevaluated tick.
             health_gate_not_evaluated_reason: health_states.unevaluated_summary(root),
+            // Resolved daemon-side (this process's own env + this root's own
+            // `.loom/config.json`), never the CLI client's environment (#4012).
+            health_gate_enabled: Some(crate::main_health_gate::effective_enabled(root)),
+            health_gate_verdict_at: health_states.last_verdict_at(root),
         });
         in_flight.extend(live);
     }
@@ -577,6 +581,8 @@ pub fn build_daemon_status(
         main_health_gate_halted: health_states.is_halted(fallback_root),
         main_health_gate_not_evaluated: health_states.is_unevaluated(fallback_root),
         main_health_gate_not_evaluated_reason: health_states.unevaluated_summary(fallback_root),
+        main_health_gate_enabled: Some(crate::main_health_gate::effective_enabled(fallback_root)),
+        main_health_gate_verdict_at: health_states.last_verdict_at(fallback_root),
         capacity,
         per_repo,
     }
@@ -2658,6 +2664,8 @@ exit 0
             main_health_gate_halted: true,
             main_health_gate_not_evaluated: false,
             main_health_gate_not_evaluated_reason: None,
+            main_health_gate_enabled: Some(true),
+            main_health_gate_verdict_at: Some(chrono::Utc::now()),
             capacity: crate::types::CapacityReport {
                 ranking_present: true,
                 total_accounts: 4,
@@ -2674,6 +2682,8 @@ exit 0
                 quarantined_issues: vec![101, 202],
                 health_gate_not_evaluated: false,
                 health_gate_not_evaluated_reason: None,
+                health_gate_enabled: Some(true),
+                health_gate_verdict_at: Some(chrono::Utc::now()),
             }],
         };
         let resp = Response::DaemonStatus(report);
@@ -2700,6 +2710,10 @@ exit 0
                 assert_eq!(r.per_repo[0].in_flight_count, 0);
                 assert!(r.per_repo[0].health_gate_halted);
                 assert!(!r.per_repo[0].health_gate_not_evaluated);
+                assert_eq!(r.main_health_gate_enabled, Some(true));
+                assert!(r.main_health_gate_verdict_at.is_some());
+                assert_eq!(r.per_repo[0].health_gate_enabled, Some(true));
+                assert!(r.per_repo[0].health_gate_verdict_at.is_some());
             }
             other => panic!("Expected DaemonStatus, got: {other:?}"),
         }
@@ -2726,6 +2740,17 @@ exit 0
         assert_eq!(report.cpu_headroom, 0);
         assert_eq!(report.logical_cpus, 0);
         assert_eq!(report.loadavg_1m, None);
+        // Absent pre-#4012 fields must default to `None` — NOT `false` — so a
+        // legacy payload from an older, perfectly healthy daemon is read as
+        // "unknown" rather than misreported as "gate disabled" (#4012).
+        assert_eq!(
+            report.main_health_gate_enabled, None,
+            "absent main_health_gate_enabled (#4012) must default to None, not Some(false)"
+        );
+        assert_eq!(
+            report.main_health_gate_verdict_at, None,
+            "absent main_health_gate_verdict_at (#4012) defaults to None (reads as pending)"
+        );
     }
 
     /// `build_daemon_status` reflects the per-repo main-health halt flag and
@@ -2769,6 +2794,37 @@ exit 0
         assert_eq!(report.per_repo[0].root, root);
         assert_eq!(report.per_repo[0].in_flight_count, 0);
         assert!(!report.per_repo[0].health_gate_halted);
+        // No `.loom/config.json` buildGate/autonomous block exists yet, so the
+        // gate is effectively disabled for this root (#4012).
+        assert_eq!(report.main_health_gate_enabled, Some(false));
+        assert_eq!(report.per_repo[0].health_gate_enabled, Some(false));
+        assert_eq!(report.main_health_gate_verdict_at, None);
+        assert_eq!(report.per_repo[0].health_gate_verdict_at, None);
+
+        // #4012: a root the gate loop HAS never evaluated (no verdict yet)
+        // reports the same `Some(false)`/`None` pair while genuinely disabled
+        // -- but once the config turns the gate on, a fresh `MainHealthState`
+        // reports "enabled, pending" (verdict_at still `None`), NOT "clear".
+        // This is the exact ambiguity the issue is about: `pending` and
+        // `disabled` both still allow dispatch, but they must not be
+        // confused with `clear` (verified green).
+        std::env::remove_var(crate::main_health_gate::MAIN_HEALTH_GATE_ENABLE_ENV);
+        std::fs::write(
+            root.join(".loom").join("config.json"),
+            r#"{"autonomous": {"mainHealthGate": {"enabled": true}}, "buildGate": {"enabled": true, "command": "true"}}"#,
+        )
+        .unwrap();
+        let report = build_daemon_status(&pool, &health, &root);
+        assert_eq!(
+            report.main_health_gate_enabled,
+            Some(true),
+            "config now enables the gate for this root"
+        );
+        assert_eq!(
+            report.main_health_gate_verdict_at, None,
+            "no gate run has completed yet -- must report pending, not clear"
+        );
+        assert!(!report.main_health_gate_halted, "pending must still read as dispatch-allowed");
 
         // Dispatch a sweep -> it should show up as in-flight (Running).
         {
@@ -2889,6 +2945,16 @@ exit 0
         assert!(!a.health_gate_halted, "repo A is green");
         assert_eq!(b.in_flight_count, 0, "repo B has no sweeps");
         assert!(b.health_gate_halted, "repo B is red, independently of A");
+        // Neither repo has a `.loom/config.json` buildGate block, so both
+        // resolve as effectively disabled (#4012) — independent of the raw
+        // halt flag test-injected directly on repo B above (`set_halted`
+        // bypasses the gate loop's own disabled soft-fail path, so this
+        // combination only arises in a test; the renderer must still prefer
+        // "halted" over "disabled" when both are true, see `main.rs`).
+        assert_eq!(a.health_gate_enabled, Some(false));
+        assert_eq!(b.health_gate_enabled, Some(false));
+        assert_eq!(a.health_gate_verdict_at, None);
+        assert_eq!(b.health_gate_verdict_at, None);
 
         std::env::remove_var(REGISTRY_PATH_ENV);
     }
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -19,6 +19,7 @@ use loom_daemon::workspace_pool::WorkspacePool;
 use loom_daemon::{extract_configured_terminal_ids, rotate_log_file};
 
 use anyhow::{anyhow, Result};
+use chrono::{DateTime, Utc};
 use clap::{Parser, Subcommand};
 use std::fs;
 use std::io::Write;
@@ -1815,6 +1816,13 @@ fn print_status_json(
             "not_evaluated": report.main_health_gate_not_evaluated,
             // Which failure class actually blocked evaluation (#3974 AC2).
             "not_evaluated_reason": report.main_health_gate_not_evaluated_reason,
+            // Whether the gate is actually enabled for this root, and when its
+            // last completed verdict landed (#4012) — the disambiguators
+            // between "disabled", "pending" (enabled, no verdict yet), and
+            // "clear" (verified green), all three of which pre-#4012 rendered
+            // identically as `halted: false, not_evaluated: false`.
+            "enabled": report.main_health_gate_enabled,
+            "verdict_at": report.main_health_gate_verdict_at,
         },
         // Per-repo breakdown across every registered managed workspace (#3930).
         "per_repo": report.per_repo.iter().map(|r| serde_json::json!({
@@ -1824,6 +1832,8 @@ fn print_status_json(
             "health_gate_halted": r.health_gate_halted,
             "health_gate_not_evaluated": r.health_gate_not_evaluated,
             "health_gate_not_evaluated_reason": r.health_gate_not_evaluated_reason,
+            "health_gate_enabled": r.health_gate_enabled,
+            "health_gate_verdict_at": r.health_gate_verdict_at,
         })).collect::<Vec<_>>(),
         // Forge-side pipeline snapshot (#3977) — present only when `--pipeline`
         // was passed; `null` otherwise so a consumer can tell "not requested"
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed the diff end-to-end against the issue's acceptance criteria and the Curator's design guidance.

**Correctness of the three-way split**
- `classify_gate_verdict` (main.rs) correctly prioritizes `halted`/`not_evaluated` over disabled/pending/clear, matching the design note that a genuinely active halt must never be hidden. `Disabled` (enabled == Some(false)) is checked before falling back to `Pending`/`Clear` based on `verdict_at.is_none()`. Verified via `format_gate_status_distinguishes_pending_from_clear`, `format_gate_status_distinguishes_disabled`, and `format_gate_status_halted_beats_disabled_and_pending` — all three states render distinct strings and neither renderer collapses back into ambiguity.
- Both render sites (`format_gate_status` long-form and the new `gate_status_short_label` for the 13-char table column) are driven from the same `GateVerdict` enum via `classify_gate_verdict`, so the top-level summary and per-repo table can't disagree — this was one of the Curator's flagged implementation traps and it's handled.
- `effective_enabled` correctly folds `resolve_enabled(..)` **and** a usable `buildGate` block into one signal, matching the gate loop's own soft-fail treatment of "enabled but no buildGate" as always-green — verified with dedicated tests (`test_effective_enabled_false_when_enabled_but_no_usable_build_gate`, etc.).

**Dispatch behavior unchanged**
- `pending_and_disabled_never_halt` explicitly asserts neither state is ever classified as `Halted`, and both render "dispatch allowed". No new halt path was introduced — confirmed by reading `apply_gate_outcome` (only stamps `last_verdict_at`, doesn't touch the halted flag logic) and `main_health_gate.rs`'s existing disabled-path behavior (unchanged).

**Backward compatibility**
- New `Option<bool>`/`Option<DateTime<Utc>>` fields on `DaemonStatusReport`/`RepoStatus` all carry `#[serde(default)]`. The `Option<bool>` choice (vs. bare `bool`) is well-reasoned and tested — `test_daemon_status_backward_compat_missing_capacity` explicitly asserts a legacy payload deserializes `main_health_gate_enabled`/`_verdict_at` to `None`, not `Some(false)`, avoiding the "misreport a healthy older daemon as disabled" trap.
- `format_gate_status_legacy_none_enabled_is_not_disabled` confirms `enabled == None` reads as `Pending`, not `Disabled` — correct conservative default.

**Verdict-timestamp stamping**
- `record_verdict_at` is only called from `apply_gate_outcome`'s Green/Red arms, never the #3984 SHA-memo skip path — verified directly with `test_run_gate_tick_skip_path_does_not_stamp_last_verdict_at`, which exercises the real skip-path code (not just a hand-built scenario) via a real git origin/clone fixture.

**Test quality**
Tests are substantive, not just compile-checks — they exercise the classifier with real inputs including edge cases (legacy `None`, halted+disabled combination, unknown-root delegates, table-width assertions cross-checked against the long-form renderer). The `ipc.rs` producer tests extend existing fixtures with real config-file writes rather than mocking the classifier directly.

**Verification performed locally** (worktree `.loom/worktrees/issue-4012`, HEAD `0eafc175`, rebased on current `main`):
- `cargo build --workspace` — clean
- `cargo test --lib` — 1012 passed, 0 failed (matches PR claim)
- `cargo test --bin loom-daemon` — 16 passed, 0 failed, including all the new `GateVerdict`/classifier tests
- `cargo clippy --package loom-daemon --package loom-api --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- CI (`gh pr checks 4068`): Rust Linting & Build Check pass, Rust Unit Tests pass, Detect Changes pass (installer/MCP jobs correctly skipped, no relevant paths touched)
- `mergeStateStatus`: CLEAN

One very minor, non-blocking observation: `.loom/docs/daemon-reference.md`'s main-health-gate section (around the `#3930`/`#3974` narrative) doesn't mention the new `enabled`/`verdict_at` fields or the `GateVerdict` enum. Not required by the issue's acceptance criteria or test plan, and not blocking approval — just flagging in case a future doc pass wants to fold it in.

Issue #4057, opened 2026-07-27
Provision a dedicated shared AWS CI runner for the project fleet (operator-only; gated on #4038)


rjwalters:
## Summary

Provision a **dedicated AWS host for CI, shared across the project fleet** — loom, anvil, repo, kicad-tools, vibesql, geode-fem — per operator direction (2026-07-27). Split out of #4038, which retained the measurement-and-decision half (now complete: [ADR-0011](https://github.com/rjwalters/loom/blob/main/docs/adr/0011-ci-runner-platform.md)).

**This is `loom:operator-only`**: it requires AWS credentials, spend authority, and a cross-repo fleet decision. A Builder agent has none of those, so dispatching it autonomously would produce unappliable Terraform at best.

## Dependencies

- [x] #4038: Measure CI's speedup ceiling and record the runner-platform decision — **closed COMPLETED 2026-07-27**; decision recorded in [ADR-0011](https://github.com/rjwalters/loom/blob/main/docs/adr/0011-ci-runner-platform.md) (merged via #4070)

## Outcome of #4038 (ADR-0011) — read before proceeding

The prerequisite ADR landed, and its measured numbers substantially weaken this issue's original speed rationale:

- **Numeric speedup ceiling: ~5–7% of the job total (~10–15s off ~213s).** The `loom_daemon` lib binary's 126.8s (67% of the step) is a 111-test `#[serial]` chain on a single in-process global lock — its ~127s wall time is **not reducible by adding cores** (an infinite-core runner cannot go below it). Forcing full serialization (`--test-threads=1`) added only ~26s over default 4-vCPU parallelism, proving the ~889 parallel tests are already hidden behind the serial chain today. Compile (~19s, warm cache) is the only clearly core-addressable segment; ~33s of integration tests are sleep/timing-bound and immune to any runner.
- **ADR decision (a): the speed case does not justify new hardware on its own.** The high-leverage fix is de-serializing the `#[serial]` tests (ADR alternatives (c)(2)/(c)(3), accepted as follow-up work), not adding cores.
[...]

PR #4070, opened 2026-07-27
docs(adr): record CI runner-platform speedup-ceiling decision (#4038)

### LINKED ISSUE #4038: Measure CI's speedup ceiling and record the runner-platform decision: compile is only 10% of ci.yml, one serial-locked test binary is 67%
## Motivation

CI is the only build signal Loom currently trusts, and it is slower than it needs to be. It is also the wrong platform.

## Measured today (2026-07-27)

Wall-clock for the last 5 `ci.yml` runs on `main`/feature branches: **214s, 227s, 232s, 248s, 268s**.

Per-job breakdown for run `30278249895` (`main`, all green):

| Job | Duration | Runner |
|-----|----------|--------|
| **Rust Unit Tests** (`cargo test --workspace`) | **210s** | `ubuntu-latest` |
| Rust Linting & Build Check | 35s | `ubuntu-latest` |
| Installer Integration Tests | 34s | `ubuntu-latest` |
| MCP Server Build | 12s | `ubuntu-latest` |

So `cargo test --workspace` **is** the critical path — it is ~98% of the wall clock. Everything else is noise. Both Rust jobs already use `Swatinem/rust-cache@v2`, so the cheap caching win is already taken.

`ubuntu-latest` on a public repo is **4 vCPU / 16 GB, x86-64**. The dev/daemon host is **28 logical cores / 96 GB, aarch64 (Apple Silicon)**.

## Two distinct arguments for a custom runner

**1. Speed.** The critical path is a single `cargo test --workspace` on 4 vCPU. More cores and a persistently warm `target/` (rather than restore/save of a cache archive each run) should cut this materially.

**2. Platform parity — arguably the bigger win.** Loom *runs* on aarch64 macOS in production. CI has only ever tested x86-64 Linux. Every macOS-specific defect is invisible to CI by construction: the tmux-dependent `integration_basic` targets that had to be excluded from the local gate in #3985, and the timing-sensitive tests discussed below, are both macOS/host-shaped problems CI cannot see.

## Constraints the design must respect

These are the parts worth getting right; they are not obvious.

### Do NOT run heavy builds on the macOS daemon host — Gatekeeper is the hidden bottleneck

> **Correction (2026-07-27):** an earlier revision of this issue attributed the numbers below to sweep-fleet CPU contention. That was wrong. The measured cause is `syspolicyd` saturation. The corrected mechanism is stronger evidence for this issue, not weaker — see below.

Measured on the daemon host today, `cargo test --workspace --lib --bins` took **474s** and produced **17 failures**, every one of them:

```
fake spawn-claude.sh did not finish writing within 10s
```

Failing tests: 11 × `sweep_registry::tests::dispatch_*`, 2 × `sweep_registry::tests::read_path_reap*`, 3 × `token_ranking_refresh::tests::*`, 1 × `work_finder::tests::test_registry_dispatcher_exports_claim_ownership_marker`. The same commit is **green in CI**.

**Root cause: macOS Gatekeeper code-signature assessment saturation.** During the run, `/usr/libexec/syspolicyd` was observed at **394% CPU** with **640 minutes of accumulated CPU time**, 14× the next-largest consumer. A parallel `cargo` build spawns thousands of short-lived `rustc`/`sccache` processes, each requiring a signature assessment; `syspolicyd` saturates and process `exec` latency explodes. Concurrently, ~20 processes sat frozen at **0.00s CPU** waiting to launch (a wall of `/usr/bin/env` stuck 30+ minutes), and `uv run` blocked indefinitely without ever spawning a child — which is what actually pushed `build-gate.sh` past its 600s timeout.

Two diagnostics pin the mechanism:
- Of two `uv` binaries invoked at the same instant, `~/.local/bin/uv` returned immediately while `/opt/homebrew/bin/uv` hung until SIGKILL — i.e. **cached** assessments pass, **new** ones block.
- It is **load-induced and self-recovering**, not a permanent wedge: `syspolicyd` fell back to ~21% CPU once the build load stopped, with no restart.

So the tests are not inherently flaky and the host is not merely "busy" — a macOS-specific serialization point degrades non-linearly under exactly the workload CI performs. **Linux runners have no equivalent assessment path**, which is an independent argument for Linux/Graviton over chasing macOS parity, and a decisive argument against self-hosting builds on the daemon host.

This matters beyond convenience: #3974's forge-CI corroboration uses a **green CI verdict** to downgrade a local false-RED to UNEVALUATED instead of halting dispatch. If CI inherits this flakiness, that backstop dies and a spurious red halts the whole fleet.

Requirement: the runner gets **dedicated, non-macOS** CPU, or it is not an improvement.

### The serial-test ceiling bounds the speedup

93 tests are `#[serial]`-gated and therefore forced sequential — 36 in `sweep_registry`, 10 each in `work_finder` and `main_health_gate`, 7 in `watch_registry`, plus others. Adding cores accelerates *compilation* and the parallel test tail, but not that serial critical path.

**Measure the serial critical path before sizing the runner.** If 120s of the 210s is inherently sequential, a 28-core runner buys far less than it looks like it should, and the higher-leverage fix is de-serializing those tests (most are `#[serial]` only because they mutate process-global env vars).

### Public repo + self-hosted runner is a security decision

`rjwalters/loom` is public. GitHub explicitly warns against self-hosted runners on public repos: a fork PR can execute arbitrary code on the runner. Any self-hosted option must either be restricted to non-fork / trusted events, or be ephemeral and isolated per job. This should be decided explicitly, not inherited by accident.

## Preferred direction: a dedicated shared AWS CI host

Operator direction (2026-07-27): stand up a **dedicated AWS host for CI, shared across the project fleet** — loom, anvil, repo, kicad-tools, vibesql, geode-fem. This is the right shape: the fixed cost of provisioning and hardening a runner amortizes across every repo, and it satisfies the "dedicated CPU, not the daemon host" requirement by construction.

Design points specific to a shared AWS host:

- **Graviton (arm64) vs x86-64.** Graviton gives aarch64 Linux — matching the *architecture* Loom runs on, though not the OS. That closes part of the parity gap cheaply, and is typically better price/performance for compile-bound work. Worth defaulting to unless something in the toolchain forbids it.
- **Full macOS parity is a separate, expensive question.** EC2 macOS instances are dedicated-host only with a 24-hour minimum allocation billing floor. Do not fold that into this issue on cost grounds; if aarch64-Linux parity is judged insufficient, split macOS into its own issue with its own cost justification.
- **Ephemeral runners are mandatory, not optional.** Several of these repos are public. A persistent shared self-hosted runner reachable from public fork PRs is a credential-theft and cross-repo-contamination path — one poisoned job could compromise the runner that every other project depends on. Use per-job ephemeral runners (e.g. `philips-labs/terraform-aws-github-runner`, or actions-runner-controller) and/or restrict self-hosted jobs to non-fork events.
- **Cross-repo contention is the same failure mode we just measured.** A shared host means loom's `cargo test` can starve kicad-tools' CI exactly the way sweeps starved the local build gate. Size for concurrent demand and set per-repo concurrency limits from the start — do not discover this the way we discovered the gate timeout.
- **Warm cache is the real speed lever**, and it is in tension with ephemeral runners: a fresh instance per job has a cold `target/`. Plan for a shared cache layer (S3-backed sccache, or a persistent cache volume) rather than assuming ephemeral + fast are automatic.
- **Cost model** — autoscaling-to-zero ephemeral instances vs. an always-on host. With six repos the duty cycle may justify always-on; that should be a measured decision, not an assumption.

## Options to evaluate

| Option | Speed | Platform parity | Fork-PR safety | Fleet reuse |
|--------|-------|-----------------|----------------|-------------|
| **Dedicated AWS host, ephemeral runners (Graviton)** | best | arch yes, OS no | needs ephemeral setup | **yes — shared across repos** |
| GitHub larger runners (paid, x86-64 or arm64 Linux) | good | no | safe (GitHub-managed) | per-repo billing |
| GitHub-hosted macOS runner | modest | **full** | safe, but expensive | per-repo billing |
| EC2 macOS dedicated host | good | **full** | needs ephemeral setup | yes, but 24h billing floor |
| Self-hosted on the daemon host | **negative** — see above | full | needs isolation work | no |

- [ ] Serial-test critical path measured and recorded, so the speedup ceiling is known before hardware is chosen.
- [ ] A decision recorded on whether platform parity (an aarch64/macOS job) is in scope here or split to its own issue — it is a *correctness* concern, not a speed one, and may deserve to land first.
- [ ] If self-hosted: fork-PR execution explicitly addressed (restricted triggers or ephemeral isolated runners), and the runner is not sharing CPU with the sweep fleet.
- [ ] Before/after wall-clock published for `ci.yml` on the same commit.

## Related

- #4020 — protecting `build-gate.sh` from concurrent sweep CPU contention. Same underlying contention, different consumer.
- #3985 — scoped the local gate to `--lib --bins` precisely because host-dependent targets measure the host, not the commit.
- #3974 — the forge-CI corroboration path that makes CI trustworthiness load-bearing for dispatch.

---

### TL;DR — the measurement was recoverable from existing CI logs, and it reframes the issue

The issue's speed argument rests on "`cargo test --workspace` is ~98% of the wall clock", which is true at the **job** level. But the step's *internal* breakdown — recovered from the GitHub Actions job log for run [`30292022378`](https://github.com/rjwalters/loom/actions/runs/30292022378) (`main`, all green, 2026-07-27) — shows the expensive part is **not** the part more cores can fix:

| Segment | Duration | % of step | Does a bigger runner help? |
|---|---|---|---|
| Compile (`Finished \`test\` profile ... in 19.52s`) | 19.5s | 10.3% | Yes — but it is only 10% |
| `loom_api` lib (4 tests) | 0.2s | 0.1% | n/a |
| **`loom_daemon` lib — 1000 tests** | **126.8s** | **66.7%** | **Only for its non-serial fraction — still unmeasured** |
| `loom_daemon` bin — 10 tests | 5.5s | 2.9% | Partially |
| `integration_basic` — 9 tests | 14.9s | 7.8% | **No** — sleep/timing-bound |
| `integration_factory_reset` — 2 tests | 15.9s | 8.4% | **No** — sleep/timing-bound |
| `integration_security` — 14 tests | 3.9s | 2.1% | Partially |
| `integration_singleton_guard` — 2 tests | 2.3s | 1.2% | **No** — timing-bound |
| doc-lint + conformance binaries (34 tests) | <0.5s | 0.2% | n/a |
| **Step total** (`cargo test --workspace`) | **190s** | 100% | |

(Job total 213s; `Swatinem/rust-cache` restore accounts for another 17s.)

Three consequences that should be settled *before* hardware is chosen:

1. **Compilation is ~10% of the step, not the bulk.** `rust-cache` is already warm, so only `loom-daemon` and `loom-api` rebuild. The "more cores compile faster" lever is capped at ~19.5s.
2. **`cargo test` runs test binaries strictly sequentially.** The `Running …` timestamps in the log form a contiguous chain with no overlap — there is zero cross-binary parallelism today. This is a speedup available **without any new hardware** (see "Cheaper alternatives" below).
3. **~33s (17%) is sleep/timing-bound**, not CPU-bound (`integration_basic`, `integration_factory_reset`, `integration_singleton_guard` — these call `sleep`/`Duration::from_*` rather than saturating a core). No runner makes these faster.

Netting it out: the CPU-bound work a larger runner can actually accelerate is **19.5s of compile plus whatever fraction of the 126.8s lib binary is both parallel and CPU-bound**. If that binary turns out to be dominated by its serial lock, the entire speed case for new hardware collapses to roughly a 10% win.

**This does not touch the platform-parity argument (§2 of the original), which stands on its own and is a correctness concern, not a speed one.**

### Corrected enumeration: `#[serial]` count and, more importantly, lock granularity

The original says "93 tests are `#[serial]`-gated". Verified against `origin/main` today:

```bash
git ls-files '*.rs' | xargs grep -hoE '#\[(serial_test::)?serial(\([^)]*\))?\]' | sort | uniq -c
#  131 #[serial]
#    9 #[serial_test::serial]
```

**140 annotations, and not one uses a named key.** That is the load-bearing detail the original missed: unkeyed `#[serial]` all contend on `serial_test`'s *single default global lock*, and that lock is in-process — so it serializes per test binary. Distribution:

| Test binary | Unkeyed `#[serial]` | Wall time |
|---|---|---|
| `loom_daemon` **lib** (`src/**` minus `main.rs`) | **111** | **126.8s** |
| `loom_daemon` bin (`src/main.rs`) | 2 | 5.5s |
| `integration_security` | 14 | 3.9s |
| `integration_basic` | 9 | 14.9s |
| `integration_factory_reset` | 2 | 15.9s |
| `integration_singleton_guard` | 2 | 2.3s |

So 111 tests inside the single most expensive binary run strictly one-at-a-time **regardless of `--test-threads`**. Drift explanation for the original's 93: it predates #4051 (which added 10 in `config_resolver.rs`), counted `src/` only, and missed the `#[serial_test::serial]` spelling.

### The one remaining unknown

Everything above came free from CI logs. Exactly **one** measurement is still missing, and it is the one that decides the hardware question:

> Of the `loom_daemon` lib binary's **126.8s**, how much is the 111-test serial chain and how much is the 889 parallel tests?

- If the serial chain dominates → more cores buy almost nothing; **de-serializing is the high-leverage fix**, and the runner should be justified on **platform parity alone**.
- If the parallel tail dominates → cores help materially and the AWS case is strong on speed too.

**Measure this in CI on `ubuntu-latest`, not locally.** Two independent reasons: (a) the baseline being improved is CI's number, and (b) local measurement on the macOS daemon host is corrupted by the `syspolicyd` saturation this very issue documents — a local timing is measuring Gatekeeper, not the commit.

Suggested method (temporary workflow step on a branch, removed before merge):

```bash
# Total for the lib binary, default parallelism (the 126.8s baseline)
time cargo test --package loom-daemon --lib

# Fully sequential — the ceiling if de-serialization achieved nothing
time cargo test --package loom-daemon --lib -- --test-threads=1

# Serial-gated subset alone is not directly selectable by attribute, so
# approximate from the two numbers above: if the two are close, the binary is
# already effectively serial and cores cannot help it.
```

### Cheaper alternatives that should be priced before buying hardware

Not recommendations — options the decision record should explicitly accept or reject, since each attacks the measured bottleneck directly and costs no infrastructure:

1. **Parallelize across test binaries.** Finding #2 above shows ~43s currently runs as a strictly sequential tail of small binaries. `cargo-nextest` pools all binaries. **Caveat that must be evaluated, not assumed away:** nextest runs each test in its own process, which changes `#[serial]` semantics — `serial_test`'s lock is in-process, so process-per-test does not provide the mutual exclusion these 140 tests currently rely on. This could surface exactly the kind of flakiness that would poison #3974's forge-CI corroboration. Evaluate carefully or reject explicitly. (nextest is not configured in this repo today.)
2. **De-serialize the env-var tests.** The original already notes most are `#[serial]` only because they mutate process-global env vars. Attacking the 126.8s binary directly is strictly better leverage than paying for cores that the global lock prevents from being used.
3. **Named `#[serial(key)]` groups.** A cheap middle path: tests mutating disjoint globals can take disjoint locks and run concurrently, with no test-isolation redesign.

### Scope: this issue vs. the operator-only half

The original issue spans two things a single Builder PR cannot deliver together:

| | Scope | Who can do it |
|---|---|---|
| **This issue (#4038)** | Measure the speedup ceiling; record the runner-platform decision as an ADR | **Builder** — measurement + docs, entirely in-repo |
| **Sibling issue (#4057)** | Provision the shared AWS host, ephemeral runners, cache layer, per-repo concurrency limits | **Operator only** — requires AWS credentials, spend authority, and a cross-repo fleet decision |

A Builder agent has no AWS credentials and no spend authority, so handing it the provisioning half yields unappliable Terraform at best. The provisioning work is therefore split to #4057 (`loom:operator-only`) and **the operator's stated direction (a dedicated shared AWS host across loom / anvil / repo / kicad-tools / vibesql / geode-fem) is preserved there unchanged.** This issue is that issue's stated prerequisite — the original's own AC #1 requires the ceiling be known "before hardware is chosen".

Original ACs #3 (fork-PR safety) and #4 (before/after wall-clock on the same commit) are inherently post-provisioning and move to #4057.

### Acceptance Criteria

- [ ] The serial-vs-parallel split inside `loom_daemon`'s 126.8s lib test binary is measured **on `ubuntu-latest` in CI** and the numbers recorded in the ADR.
- [ ] An ADR at `docs/adr/0011-ci-runner-platform.md` records: the measured breakdown above, the serial-split measurement, and a decision on each of — (a) whether the speed case justifies new hardware given the measured ceiling, (b) whether aarch64-Linux (Graviton) parity is sufficient or full macOS parity is required, (c) accept-or-reject for each of the three cheaper alternatives, with reasons.
- [ ] The ADR states an explicit **numeric** speedup ceiling (e.g. "a runner with unlimited cores cannot take `ci.yml` below Xs") so #4057 can be justified or dropped on evidence.
- [ ] `docs/adr/README.md` index updated with the new ADR entry, matching the existing format (Status / Summary / Key Decision).
- [ ] Any temporary timing instrumentation added to `ci.yml` is removed before merge — `ci.yml` is unchanged on `main` by this issue, or the diff is timing-only and explained in the PR body.
- [ ] #4057 is cross-referenced from the ADR.

**Out of scope** (guard against scope creep): no `runs-on` change, no Terraform, no AWS resources, no nextest migration, and no de-serialization of tests in this PR. This issue produces **a measurement and a decision**. Acting on the decision is follow-up work.

## Affected Files

- `docs/adr/0011-ci-runner-platform.md` — **new**; the decision record (next free ADR number; `0010-daemon-rebuild.md` is the current highest)
- `docs/adr/README.md` — add the index entry under an appropriate section
- `.github/workflows/ci.yml` — *temporary* timing instrumentation on the branch only; expected to be unchanged in the merged diff (`backend-tests` job, `cargo test --workspace` step at the `runs-on: ubuntu-latest` / step definition)
- `docs/adr/template.md` — reference only, for ADR structure

Reference-only (read to source the numbers, not modified):
- `loom-daemon/src/sweep_registry.rs` (36 `#[serial]`), `main_health_gate.rs` (11), `work_finder.rs` (10), `config_resolver.rs` (10), `watch_registry.rs` (7) — the largest contributors to the 111-test serial chain
- `loom-daemon/tests/integration_basic.rs`, `integration_factory_reset.rs` — the sleep-bound ~31s

## Test Plan

- [ ] **Manual verification**: push the branch, confirm the instrumented `backend-tests` job emits both the default-parallelism and `--test-threads=1` timings for `cargo test --package loom-daemon --lib`, and that the default-parallelism number is within ~10% of the 126.8s baseline (a large deviation means the measurement is unsound and should be re-run, not recorded).
- [ ] **Automated tests**: N/A — this issue adds a document and a temporary measurement, no runtime code. Existing `cargo test --workspace` must remain green.
- [ ] **Reproduce the baseline**: re-derive the breakdown table from a fresh CI run to confirm it is stable and not an artifact of one run:
  ```bash
  RUN=$(gh run list --workflow=ci.yml --branch=main --status=success --limit=1 --json databaseId --jq '.[0].databaseId')
  JOB=$(gh api "repos/rjwalters/loom/actions/runs/$RUN/jobs" --jq '.jobs[]|select(.name|test("Unit Tests")).id')
  gh api "repos/rjwalters/loom/actions/jobs/$JOB/logs" | sed 's/\x1b\[[0-9;]*m//g' \
    | grep -E 'Finished .*profile|Running (unittests|tests)|test result:'
  ```
- [ ] **Edge cases**: verify the measurement is taken on a run where `rust-cache` **hit** (a cache miss inflates the compile segment several-fold and would invalidate the "compile is only 10%" finding); note the cache status in the ADR.
- [ ] **Docs check**: `docs/adr/README.md` renders correctly and the new ADR link resolves.

## Dependencies

None. This issue is itself the prerequisite for #4057.

rjwalters:
## Summary

Closes #4038.

Measures the CI speedup ceiling for `ci.yml` and records a runner-platform decision as an ADR — no runtime code or permanent CI changes.

- **Measured**: the `loom-daemon` lib test binary's serial-vs-parallel split in CI (default parallelism `127.24s` vs fully serial `--test-threads=1` `153.25s`, delta `26.01s`). Combined with the curator's earlier per-segment breakdown (compile ~19.5s = 10.3% of the step; the lib binary = 66.7%), this shows the ~127s serial-lock chain (111 unkeyed `#[serial]` tests contending one in-process mutex) is **not reducible by adding cores** — the 889 parallel tests are already almost entirely hidden behind it at 4 vCPUs.
- **Decision recorded** in `docs/adr/0011-ci-runner-platform.md`:
  - (a) Reject new CI hardware on speed grounds alone — the measured ceiling is only ~5-7% of the job total.
  - (b) If a runner is ever provisioned (#4057), default to Graviton (arch parity); treat full macOS parity as a separate, independently-costed decision, not bundled with this one.
  - (c) Accept/reject each of the three cheaper alternatives: reject `cargo-nextest` as a drop-in fix for the `#[serial]`-heavy lib binary (process-per-test breaks `serial_test`'s in-process lock semantics — risks poisoning #3974's forge-CI corroboration), accept de-serializing env-var-mutating tests as the primary follow-up, accept named `#[serial(key)]` groups as a cheap incremental step.
  - Cross-references #4057 (operator-only provisioning) as depending on this decision.
- Added the ADR-0011 index entry to `docs/adr/README.md` (also backfilled the missing ADR-0010 entry, which existed on disk but wasn't indexed).
- **Reverted** the temporary `ci.yml` timing instrumentation used to gather the serial-vs-parallel measurement (CI run [30294816703](https://github.com/rjwalters/loom/actions/runs/30294816703), commit `90fa2a4e`, conclusion success, rust-cache warm) — `ci.yml` is unchanged vs `origin/main` in the final diff.

## Test plan

- [x] `git diff origin/main -- .github/workflows/ci.yml` is empty (instrumentation fully reverted).
- [x] ADR data cross-checked against the CI run that produced it (compile times ~18-19s on both segments, consistent with the curator's earlier ~19.5s baseline, confirming warm cache).
- [ ] No runtime/behavioral tests apply — this PR adds documentation only; `cargo test --workspace` on `ci.yml` covers regression as usual.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,12 +67,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace
-      # TEMPORARY (#4038): measure the serial-vs-parallel split inside the
-      # loom-daemon lib test binary's ~126.8s wall-clock. Removed before merge.
-      - name: "TEMP #4038: loom-daemon lib tests, default parallelism"
-        run: time cargo test --package loom-daemon --lib
-      - name: "TEMP #4038: loom-daemon lib tests, --test-threads=1"
-        run: time cargo test --package loom-daemon --lib -- --test-threads=1
 
   mcp-build:
     name: MCP Server Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace
+      # TEMPORARY (#4038): measure the serial-vs-parallel split inside the
+      # loom-daemon lib test binary's ~126.8s wall-clock. Removed before merge.
+      - name: "TEMP #4038: loom-daemon lib tests, default parallelism"
+        run: time cargo test --package loom-daemon --lib
+      - name: "TEMP #4038: loom-daemon lib tests, --test-threads=1"
+        run: time cargo test --package loom-daemon --lib -- --test-threads=1
 
   mcp-build:
     name: MCP Server Build
--- a/docs/adr/0011-ci-runner-platform.md
+++ b/docs/adr/0011-ci-runner-platform.md
@@ -1,0 +1,306 @@
+# ADR-0011: CI Runner Platform — Speedup Ceiling and Decision
+
+## Status
+
+Accepted
+
+## Context
+
+Issue #4038 was filed to evaluate standing up a dedicated CI runner (a
+shared AWS host, provisioned across the project fleet) on two distinct
+grounds: **speed** (`cargo test --workspace` on `ubuntu-latest` is the
+critical path of `ci.yml`, and more cores / a warm `target/` should cut
+it) and **platform parity** (Loom runs in production on aarch64 macOS;
+CI has only ever tested x86-64 Linux, and macOS-specific defects — the
+tmux-dependent tests excluded from the local gate in #3985, and
+Gatekeeper/`syspolicyd` saturation under parallel `cargo` builds — are
+invisible to CI by construction).
+
+The acceptance criteria for *this* issue were narrowly scoped to
+measurement and a decision record, not provisioning: measure the
+speedup ceiling before any hardware is bought, and record whether the
+ceiling justifies new hardware. Provisioning itself — Terraform, AWS
+spend, ephemeral-runner isolation, per-repo concurrency limits — is
+split to the sibling issue #4057 (`loom:operator-only`), which depends
+on the decision recorded here.
+
+### Measured breakdown of `cargo test --workspace` (curator pass, CI run `30292022378`, `main`, green)
+
+| Segment | Duration | % of step | Does a bigger runner help? |
+|---|---|---|---|
+| Compile (`Finished \`test\` profile ... in 19.52s`) | 19.5s | 10.3% | Yes — but only 10% |
+| `loom_api` lib (4 tests) | 0.2s | 0.1% | n/a |
+| `loom_daemon` lib — 1000 tests | 126.8s | 66.7% | Only for its non-serial fraction (unmeasured at curator time) |
+| `loom_daemon` bin — 10 tests | 5.5s | 2.9% | Partially |
+| `integration_basic` — 9 tests | 14.9s | 7.8% | No — sleep/timing-bound |
+| `integration_factory_reset` — 2 tests | 15.9s | 8.4% | No — sleep/timing-bound |
+| `integration_security` — 14 tests | 3.9s | 2.1% | Partially |
+| `integration_singleton_guard` — 2 tests | 2.3s | 1.2% | No — timing-bound |
+| doc-lint + conformance binaries (34 tests) | <0.5s | 0.2% | n/a |
+| **Step total** | **190s** | 100% | |
+
+Job total was 213s (the extra ~17s is `Swatinem/rust-cache` restore).
+`cargo test` runs each test binary strictly sequentially — the
+`Running …` timestamps in the log form a contiguous chain with zero
+cross-binary overlap, so there is speedup available today with **no
+new hardware** by parallelizing across binaries (see "Cheaper
+alternatives" below).
+
+Also verified against `origin/main`: **140** `#[serial]`/
+`#[serial_test::serial]` annotations exist repo-wide (up from the
+original issue's estimate of 93, which predated #4051's
+`config_resolver.rs` additions and missed the `#[serial_test::serial]`
+spelling). None use a named key — all 140 contend on
+`serial_test`'s single default global lock, which is in-process and
+therefore serializes per test binary, not repo-wide. **111** of the 140
+live in the `loom_daemon` **lib** binary — the 126.8s line above.
+
+### New measurement: the lib binary's serial-vs-parallel split
+
+The one number the curator pass could not derive from existing logs —
+how much of the `loom_daemon` lib binary's 126.8s is the 111-test
+serial chain versus the ~889 parallel tests — was measured directly in
+CI. Two temporary steps were added to the `backend-tests` job on
+`ubuntu-latest` (4 vCPU), run after the normal `cargo test --workspace`
+step so `rust-cache` was already warm:
+
+```bash
+# Default parallelism (the existing 126.8s baseline, re-measured)
+time cargo test --package loom-daemon --lib
+# Fully sequential — the ceiling if de-serialization achieved nothing
+time cargo test --package loom-daemon --lib -- --test-threads=1
+```
+
+Run [`30294816703`](https://github.com/rjwalters/loom/actions/runs/30294816703)
+(job `90073172969`), on commit `90fa2a4e`, conclusion `success`:
+
+| Measurement | Wall time |
+|---|---|
+| Default parallelism (`cargo test --package loom-daemon --lib`) | **127.24s** |
+| Fully serial (`--test-threads=1`) | **153.25s** |
+| Delta (parallel headroom at 4 vCPUs) | **26.01s** (~17% of the serial figure) |
+
+The re-measured default-parallelism figure (127.24s) matches the
+curator's earlier-recovered 126.8s baseline within measurement noise,
+confirming the measurement is stable and not an artifact of one run.
+Compile time on this run was logged twice at 17.98s and 19.31s — both
+close to the original 19.52s baseline, confirming `rust-cache` hit
+(warm cache) and that the "compile is only ~10%" finding is **not** a
+cold-cache artifact.
+
+### The inference this measurement forces
+
+Forcing full serialization only adds ~26s over default 4-way
+parallelism. That means the ~889 non-`#[serial]` tests are **already
+almost entirely hidden** behind the 111-test serial lock chain at just
+4 vCPUs — while the serial chain runs on one thread, the other threads
+finish the parallel tests well within that window. Consequences:
+
+- The serial chain's own duration is approximately the entire
+  default-parallelism figure, **~127s**, and that duration is **not
+  reducible by adding cores**: a global mutex forces 111 tests to run
+  one at a time regardless of how many cores are available.
+- The 889 parallel tests cost only ~26s of aggregate wall time beyond
+  the serial floor, and are already fully parallelized away today at 4
+  vCPUs — more cores cannot meaningfully shrink an already-near-zero
+  marginal cost.
+- Extending this to the whole `cargo test --workspace` step (~190–213s
+  total): compile (~19s, warm cache) is the only clearly
+  core-addressable segment; the ~33s of sleep/timing-bound integration
+  tests (`integration_basic`, `integration_factory_reset`,
+  `integration_singleton_guard`) are wall-clock fixed regardless of
+  cores; and the lib binary's ~127s is serial-lock-bound per above.
+
+### Numeric speedup ceiling
+
+**An infinite-core runner cannot take the `loom_daemon` lib test binary
+below ~127s**, because that duration is the serial-lock chain's own
+wall time, not a function of available parallelism. Extending to the
+whole `ci.yml` critical path: the only segment a bigger/faster runner
+can meaningfully compress is the ~19s compile step (and, marginally,
+the non-serial-bound bin/`integration_security` fractions, on the
+order of a few seconds). So **cores/hardware alone cannot take
+`ci.yml`'s critical path below roughly S ≈ 190–200s** out of the
+current ~213s job total — an at-most ~10–15s (5–7%) win, not the
+"more cores solve it" story the original issue's speed argument
+assumed.
+
+## Decision
+
+**(a) The speed case does not justify new hardware on its own.**
+The measured ceiling (~5–7% off the job total) is too small to justify
+provisioning, maintaining, and securing a dedicated AWS host. Chasing
+that number with hardware would spend real recurring cost and
+operational complexity (ephemeral-runner isolation for a public repo,
+per-repo concurrency limits, a shared cache layer) to shave single-digit
+seconds off a ~213s job. **The actual high-leverage fix is
+de-serializing the `#[serial]` tests**, not adding cores — see (c)
+below.
+
+**(b) aarch64-Linux (Graviton) parity is judged sufficient for the
+"custom runner" question; full macOS parity is out of scope here.**
+The original issue's own tradeoff table lists Graviton (arch parity,
+not OS parity) as cheap and GitHub-hosted macOS runners / EC2 macOS
+dedicated hosts as expensive (EC2 macOS carries a 24-hour minimum
+billing floor per dedicated host). Since this ADR has just eliminated
+the speed justification, the platform-parity argument would need to
+carry a runner decision on its own — and full macOS parity's cost
+profile (dedicated-host billing floor, ephemeral-isolation design for
+a *macOS* fleet) is a materially bigger and separately-costed
+commitment than a Linux/Graviton job. Reasoning it through: the
+concrete macOS-specific defects already surfaced (#3985's
+tmux-dependent test exclusion, `syspolicyd`/Gatekeeper saturation under
+parallel `cargo` builds) are real correctness gaps, but they are gaps
+in *macOS-specific* behavior — an aarch64-**Linux** runner closes zero
+of them, because Gatekeeper and tmux-on-macOS are OS-level, not
+architecture-level, concerns. So Graviton buys architecture parity
+(useful for catching aarch64-specific bugs in, e.g., pointer-width or
+endianness-sensitive code, though none are known today) without buying
+the OS parity that actually matters for the two concrete defects on
+record. **Decision: if a runner is provisioned at all (per #4057), it
+should default to Graviton for cost/architecture reasons, but full
+macOS parity — if pursued — must be justified and costed as its own,
+separate decision**, not bundled into this one. This ADR does not
+itself decide to pursue macOS parity; it only says the two questions
+(Linux/Graviton vs. macOS) are not equivalent and should not be
+conflated.
+
+**(c) Cheaper alternatives — accept or reject:**
+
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Verified against issue #4038's acceptance criteria:

- **`docs/adr/0011-ci-runner-platform.md`** exists and follows the repo's ADR conventions (Status / Context / Decision / Consequences / Alternatives Considered / References — structurally matches `0010-daemon-rebuild.md`).
- **Data is grounded, not asserted.** Confirmed directly against the cited CI run [`30294816703`](https://github.com/rjwalters/loom/actions/runs/30294816703) job `90073172969` (commit `90fa2a4e`, conclusion `success`): the log shows `test result: ... finished in 127.24s` for default-parallelism `cargo test --package loom-daemon --lib` and `finished in 153.25s` for the `--test-threads=1` run — exact matches to the numbers quoted in the ADR. Compile times in both segments (19.31s / 0.08s) confirm a warm `rust-cache`, supporting the "compile is only ~10%" claim.
- **Decision is well-reasoned**: rejects new CI hardware on speed grounds alone (~5-7% ceiling), separates the Graviton-vs-full-macOS-parity question explicitly, and gives an accept/reject verdict with rationale for each of the three cheaper alternatives (nextest, env-var de-serialization, named `#[serial(key)]` groups). Cross-references #4057 correctly.
- **`docs/adr/README.md`** diff is clean: adds the ADR-0011 entry under a new "CI Infrastructure" section and backfills the previously-missing ADR-0010 index entry, matching the existing Status/Summary/Key Decision format.
- **Docs-only, confirmed via merge-base diff** (`git diff $(git merge-base <base> <head>) <head> --stat`): exactly 2 files changed, 318 insertions, 0 deletions — `docs/adr/0011-ci-runner-platform.md` and `docs/adr/README.md`. No stray runtime changes.
- **`.github/workflows/ci.yml` diff is empty** vs. both the merge-base and current `origin/main` — the temporary timing instrumentation used to gather the serial-split measurement was fully reverted before finalizing, as promised in the PR body.
- **CI**: `Detect Changes` passed; the remaining jobs (`Rust Unit Tests`, `Rust Linting & Build Check`, `Installer Integration Tests`, `MCP Server Build`) correctly skipped via path-filtering since only `docs/` changed. `mergeStateStatus` is `CLEAN`.

No blocking concerns. Nice work closing the loop on #4038's measurement requirement with a real CI-sourced number rather than an estimate.

Issue #4017, opened 2026-07-27
feat(daemon): auto-rebuild and restart when the running binary is stale


rjwalters:
> ## 📋 Tracking issue — ALL 3 phases have landed; only the live E2E acceptance demo remains
>
> This issue is **not directly buildable**. It was decomposed into three phases
> (see [Decomposed sub-issues](#decomposed-sub-issues)); **#4053, #4054, and
> #4055 are all merged** (#4055 landed as PR #4181, `3c9a40e3`, 2026-07-28).
> `loom:blocked` is lifted; the issue is now **`loom:operator-only`**: the sole
> remaining work is this issue's own end-to-end Test Plan below, which requires
> the operator's live production daemon (enable the default-OFF
> `autonomous.autoUpdate`, observe a real roll). PR #4181's automated coverage
> is deliberately fake-runner-only ("no real `cargo build` in tests"), so the
> composition demo cannot be delegated to a Builder. Close after the first
> observed live roll.
>
> ⚠️ **Correction (2026-07-28)**: the previous revision's warning that "there is
> no `auto_update.rs`" is now **wrong** — the shipped #4055 implementation
> *did* create `loom-daemon/src/auto_update.rs` (config, safety gates, backoff,
> drain-roll sequencing), reversing its curation-time plan to live inside
> `ipc.rs`. The stale warnings below are struck in place.
>
> **Re-curated 2026-07-28 against `origin/main` @ `3697c5ae`** (62 commits past
> the previous verification base `789669e9`). Two design questions this body
> originally posed as *open* have since been **decided and shipped** — they are
> marked ✅ **RESOLVED** inline below rather than deleted, so the reasoning that
> produced the decision stays on the record. Every `file:line` anchor has been
> re-verified and corrected; the previous set had fully rotted.
>
> **Curated 2026-07-28 (later pass), substance re-confirmed against `origin/main` @ `5a51931f`** (22 commits
> past `3697c5ae`): `loom-daemon/src/auto_update.rs` present and wired in `main.rs` (config read + task spawn
> ~`:1303-1324`); `autonomous.autoUpdate.*` documented in `defaults/docs/daemon-reference.md` (`:1515-1517`,
> `:2697`); launchd plist contract intact after #4203/#4180 (`loom-daemon-start.sh`: `render_launchd_plist()`
> now `:274`, `RunAtLoad` `:307`, `KeepAlive:{SuccessfulExit:true}` `:311`). **Line anchors in the body below
[...]

PR #4074, opened 2026-07-27
fix(daemon): self-verify rebuild→provision so a roll cannot silently ship nothing

### LINKED ISSUE #4053: fix(daemon): make rebuild→provision self-verifying so an auto-roll can never silently ship nothing
## Context

Phase 1 of #4017 (*auto-rebuild and restart when the running binary is stale*). #4017's "Notes for the implementer" asks to confirm whether `provision_machine_daemon` can silently no-op during a roll, because an auto-update built on it "would report success while shipping nothing".

It can, and the root cause is upstream of provisioning. This issue fixes the rebuild→provision path so it is **self-verifying**, independently of whether the auto-update loop (#4017) ever ships. It has standalone value today: the manual `loom-daemon-update.sh` path has the same hazard.

## Root cause (verified 2026-07-27 against `origin/main` @ `789669e9`)

`loom-daemon/build.rs:27-28` declares the staleness triggers for the commit baked into `--version`:

```rust
println!("cargo:rerun-if-changed=../.git/HEAD");
println!("cargo:rerun-if-changed=../.git/index");
```

`.git/HEAD` is a *symbolic* ref — its content is `ref: refs/heads/main` and it does **not** change when `main` advances. Observed on this checkout:

| path | mtime |
|---|---|
| `.git/HEAD` | 10:13:41 |
| last commit `789669e9` | 10:23:16 |
| `.git/refs/heads/main` | 10:31:15 |
| `.git/index` | 10:41:08 |

HEAD's mtime *predates* the commit. The file that actually moved — `.git/refs/heads/main` — is **not** in the watch set, and neither is `.git/packed-refs` (where the ref lives after a `git gc`). The only reason a stale `LOOM_DAEMON_GIT_COMMIT` isn't routinely baked in is the *incidental* `.git/index` trigger, which fires on checkout/merge/add but is not a reliable proxy for "HEAD moved".

This makes the failure **intermittent**, which matches the operator note in #4017.

Downstream, `scripts/install/provision-daemon.sh:52-61` skips the install when the two `--version` strings match:

```bash
src_ver=$("$src_bin" --version 2>/dev/null || echo "unknown")   # :50
if [[ "$src_ver" == "$dest_ver" && "$src_ver" != "unknown" ]]; then
  _pmd_ok "already current at $dest_bin ($dest_ver)"
  _pmd_check_path "$dest_dir"
  return 0
fi
```

`--version` does embed both commit and build timestamp (`loom-daemon 0.15.0 (commit a68099d5, built 2026-07-27T16:42:25Z)` — `build.rs:48,70` → `main.rs:40-46`), so **this comparison is sound given a correct rebuild**. Do not "fix" the short-circuit by removing it; the defect is the build-script watch set that lets a rebuild produce a binary still stamped with the old commit.

## Why this blocks #4017

`self_update::check()` (`loom-daemon/src/self_update.rs:97`) compares `BUILT_COMMIT` against source HEAD. If a rebuild can leave `BUILT_COMMIT` stale, an auto-update loop becomes an **infinite retry loop**: detect stale → rebuild → new binary still reports the old commit → still stale → rebuild. Backoff (#4017 AC 5) would mask this as a build failure when the build in fact succeeded. This is not something backoff can fix — the rebuild must be able to *prove* it produced what it claimed.

## Scope

1. **Correct the `build.rs` watch set** so the baked commit tracks HEAD movement. Watch the resolved ref (`.git/refs/heads/<branch>`) and `.git/packed-refs` in addition to `HEAD`/`index`, or use a mechanism that is correct by construction. Handle the cases `build.rs` already promises to tolerate: no `.git` at all (release tarball → `commit unknown`), and a linked worktree where `.git` is a *file* rather than a directory (`../.git/HEAD` does not exist).
2. **Add a post-build verification step** to `defaults/scripts/cli/loom-daemon-update.sh`: after `cargo build --release` (stage at `:270-303`) and before provisioning (`:304-331`), assert that the freshly built binary's embedded commit equals the source HEAD it was built from. On mismatch, **fail loudly and do not provision** — a mismatch is a build-system defect that retrying cannot fix. `extract_commit` (`loom-daemon-update.sh:121-123`) already parses the commit out of `--version`.
3. **Add a post-provision verification step**: after `provision_machine_daemon` returns, assert the destination binary's `--version` is the expected one. This is the direct answer to "reports success while shipping nothing" — the short-circuit stays, but it can no longer lie.

## Out of scope

- The auto-update loop itself, any periodic scheduling, config, or backoff (#4017 Phase 3).
- Any restart mechanism or plist change (#4017 Phase 2).
- Migrating any config reader onto `loom-daemon/src/config_resolver.rs` (owned by #4047).

## Affected Files

- `loom-daemon/build.rs` — correct the `cargo:rerun-if-changed` watch set (`:27-28`); keep the existing `unknown` fallbacks (`:46`, `:68`).
- `defaults/scripts/cli/loom-daemon-update.sh` — add built-commit verification after the rebuild stage (`:270-303`) and destination verification after the provision stage (`:304-331`); reuse `extract_commit` (`:121-123`). Document any new failure exit code alongside the existing table (`:52-55`).
- `scripts/install/provision-daemon.sh` — leave the `--version` short-circuit (`:52-61`) intact; add the post-copy assertion, or expose enough for the caller to make it.
- `defaults/scripts/tests/test-loom-daemon-update.sh` — new cases (see Test Plan).
- `.loom/docs/daemon-reference.md` — update the Self-update section (`:1857-1926`), which currently describes provisioning without mentioning any verification step.

## Test Plan

- [ ] Automated tests: extend `defaults/scripts/tests/test-loom-daemon-update.sh` (407 lines; already forces the nohup path at `:27-31` so it never touches `~/Library/LaunchAgents/`) with:
  - a stub binary whose `--version` commit does **not** match source HEAD → update script exits non-zero and **does not** provision
  - a stub whose destination `--version` differs from source after a claimed successful provision → non-zero exit
  - the happy path still exits `0` and still provisions
- [ ] Automated tests: a `build.rs` watch-set regression test. Since build scripts are awkward to unit-test, a shell test that creates a scratch git repo, builds, commits (touching only refs), rebuilds, and asserts the embedded commit changed is acceptable.
- [ ] Manual verification: on a real checkout, `git log -1` the current commit, run `./.loom/scripts/cli/loom-daemon-update.sh --force`, then confirm `loom-daemon --version` reports **that** commit and `~/.local/bin/loom-daemon` was actually replaced (compare `stat` mtime before/after).
- [ ] Manual verification: `./.loom/scripts/cli/loom-daemon-update.sh --check` still exits `0` when current / `3` when an update is available.
- [ ] Edge cases: no `.git` (tarball build → `commit unknown`, short-circuit correctly disabled by the existing `!= "unknown"` guard); a linked git worktree where `.git` is a file; a repo where `refs/heads/<branch>` has been packed into `packed-refs` by `git gc`.
- [ ] Regression: `cargo test --workspace` and `bash defaults/scripts/tests/test-loom-daemon-update.sh` pass.

## Acceptance Criteria

- [ ] A `cargo build --release` that follows a HEAD change always produces a binary whose embedded `LOOM_DAEMON_GIT_COMMIT` equals the new HEAD — including when only `.git/refs/heads/<branch>` or `.git/packed-refs` moved.
- [ ] `loom-daemon-update.sh` refuses to provision a binary whose embedded commit does not match the source HEAD it was built from, and says so distinguishably from a compile failure.
- [ ] `loom-daemon-update.sh` verifies the destination binary after provisioning and fails if it is not the expected build.
- [ ] The `--version`-equality short-circuit in `provision-daemon.sh` is retained (it is correct once the rebuild is correct) and covered by a test that proves it no longer produces a silent no-op on a real roll.
- [ ] Existing behavior is unchanged on the happy path: exit codes `0`/`1`/`3`, flag replay from `.loom/.daemon.flags`, and "a daemon that was not running is never started".

rjwalters:
## Summary

Makes the daemon's rebuild → provision path self-verifying so a successful-looking `loom-daemon-update.sh` run can never install a stale binary — the root cause being a `build.rs` watch set that failed to track HEAD movement, letting a rebuild bake in the *old* commit. This is Phase 1 of #4017 and has standalone value: the manual update path had the same hazard.

Root cause (verified): `build.rs` watched only `../.git/HEAD` + `../.git/index`. `.git/HEAD` is a *symbolic* ref whose content never changes when the branch advances, and the file that actually moves (`.git/refs/heads/<branch>`, or `.git/packed-refs` after `git gc`) was unwatched. In a **linked worktree** (`.git` is a *file* — every Builder's compile environment) neither `../.git/*` path even resolved, so the watch set evaporated entirely.

## Changes

- **`loom-daemon/build.rs`** — resolve the `cargo:rerun-if-changed` watch set correct-by-construction via `git rev-parse --git-path` (HEAD, index, packed-refs, and the current branch's symbolic-ref target). Emits **only paths that exist** — empirically confirmed cargo re-runs the build script on *every* build for a nonexistent `rerun-if-changed` path, so a naive fix would trade a stale-commit bug for an every-build recompile. Preserves the release-tarball (`commit unknown`) fallback and handles detached HEAD (no symbolic ref).
- **`defaults/scripts/cli/loom-daemon-update.sh`** — (1) after `cargo build --release`, assert the freshly-built binary's embedded commit == source HEAD **before** provisioning; on mismatch fail loudly and do **not** provision (**exit 4**, distinct from a compile failure). (2) After provisioning, assert the **destination** binary is the expected build for **both** the `LOOM_DAEMON_BIN` override path and the machine-level `provision_machine_daemon` path (**exit 5** on mismatch). Machine-level provisioning failure now hard-fails instead of the pre-existing soft warn that left exit 0.
- **`scripts/install/provision-daemon.sh`** — export `PROVISIONED_DAEMON_BIN` (set even on the `--version`-equality short-circuit) so the caller can verify the destination. The short-circuit is **retained** (correct once the rebuild is correct).
- **`defaults/scripts/tests/test-build-rs-watchset.sh`** *(new)* — compiles a scratch crate using a verbatim copy of the real `build.rs` in a scratch git repo (crate in a `loom-daemon/` subdir, mirroring reality) and asserts the baked commit re-derives after a **ref-only** advance (`commit-tree` + `update-ref`, touching neither HEAD nor index) across main checkout, linked worktree, detached HEAD, and packed-refs. Verified to **fail against the old build.rs** and pass against the new one.
- **`defaults/scripts/tests/test-loom-daemon-update.sh`** — new cases: built-commit mismatch → exit 4 + no provision; a provision that reports success but ships a stale destination → exit 5; happy-path roll now asserts both verification steps ran.
- **`.loom/docs/daemon-reference.md`** — document the self-verifying rebuild → provision behavior in the Self-update section.

- [x] A `cargo build --release` after a HEAD change always bakes the new HEAD — including when only `.git/refs/heads/<branch>` or `.git/packed-refs` moved (regression test, all 10 assertions pass; fails on old build.rs).
- [x] `loom-daemon-update.sh` refuses to provision a binary whose embedded commit ≠ source HEAD, distinguishably from a compile failure (exit 4, "Build verification FAILED").
- [x] `loom-daemon-update.sh` verifies the destination after provisioning and fails if not the expected build (exit 5), covering both the `LOOM_DAEMON_BIN` override and `provision_machine_daemon` branches.
- [x] The `--version`-equality short-circuit in `provision-daemon.sh` is retained and covered by a test proving it no longer produces a silent no-op on a real roll.
- [x] Happy-path behavior unchanged: exit codes `0`/`1`/`3`, flag replay from `.loom/.daemon.flags`, "a daemon that was not running is never started".
- [x] Edge cases handled: no `.git` (tarball), linked worktree (`.git` as file), packed-refs, detached HEAD.

## Test Plan

- [x] `bash defaults/scripts/tests/test-build-rs-watchset.sh` — 10/10 pass; confirmed it fails against the old watch set.
- [x] `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 23/23 pass (includes the new exit-4 / exit-5 cases).
- [x] `cargo build --release` produces a binary whose `--version` reports the current HEAD.
- [x] `cargo test --workspace` — all pass (`test_multiple_clients` timed out once under concurrent-build CPU contention; passes in isolation; unrelated to this change — build.rs only affects the baked-commit env var).

Closes #4053

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1970,6 +1970,35 @@ is installed to the machine-level location via
 convention `loom-daemon-start.sh` already resolves through `command -v
 loom-daemon`.
 
+**Self-verifying rebuild → provision (a roll can never silently ship nothing,
+#4053)**: the rebuild → provision path proves it produced what it claims, so a
+successful-looking run can no longer install a stale binary:
+
+- **Built-commit verification (before provisioning).** After `cargo build
+  --release`, the script asserts the freshly-built binary's embedded commit
+  (parsed from `--version`) equals the source `HEAD` it was built from. On
+  mismatch it **fails loudly and does not provision** — a build that succeeds
+  yet bakes in the wrong commit is a build-system defect (historically a
+  `build.rs` `cargo:rerun-if-changed` watch-set bug that missed HEAD movement),
+  not a compile failure, and retrying cannot fix it. **Exit code `4`**,
+  distinguishable from a compile failure (`1`). The underlying `build.rs`
+  watch set is now correct-by-construction via `git rev-parse --git-path`
+  (resolving `HEAD`, the current branch's ref, and `packed-refs`), so it tracks
+  HEAD movement in the main checkout, a **linked worktree** (`.git` is a file —
+  every Builder's environment), a detached HEAD, and a packed-refs repo alike.
+- **Post-provision verification (after provisioning).** The script then asserts
+  the **destination** binary's `--version` is the expected build — for both the
+  `LOOM_DAEMON_BIN` override path and the machine-level
+  `provision_machine_daemon` path. This is the direct answer to "reports success
+  while shipping nothing": the `--version`-equality short-circuit in
+  `provision-daemon.sh` is retained (it is correct once the rebuild is correct)
+  but can no longer produce a silent no-op on a real roll — `provision_machine_daemon`
+  exports the destination it wrote to (`PROVISIONED_DAEMON_BIN`, set even on the
+  short-circuit path) so the caller can verify it. A destination that is not the
+  expected build, or a provisioning step that reports failure, now **exits
+  non-zero** (`5` for a destination mismatch) instead of the pre-#4053 soft
+  warn that left the exit code at `0`.
+
 **Read-only "update available" surface (`loom-daemon --status`)**: separately
 from the update script, `loom-daemon --status` / `--status --json` now prints
 a purely local, read-only self-update line — the same built-commit-vs-source-HEAD
--- a/defaults/scripts/cli/loom-daemon-update.sh
+++ b/defaults/scripts/cli/loom-daemon-update.sh
@@ -53,6 +53,15 @@
 #   0  up to date (no-op) OR rebuild+provision+restart succeeded
 #   1  usage error / not a source checkout / build or provision failure
 #   3  (--check only) update available
+#   4  build verification FAILED: the freshly-built binary's embedded commit
+#      does not match the source HEAD it was built from. This is a BUILD-SYSTEM
+#      defect (a stale baked-in commit — e.g. a build.rs watch-set bug), NOT a
+#      compile failure, and retrying cannot fix it; the script refuses to
+#      provision the mis-stamped binary (#4053).
+#   5  post-provision verification FAILED: the destination binary after a
+#      claimed-successful provision is not the expected build (a silent no-op
+#      roll — "reports success while shipping nothing"). Distinct from both a
+#      compile failure and a provisioning soft-failure (#4053).
 #
 # See also: loom-daemon-start.sh (writes .loom/.daemon.flags), loom-daemon-stop.sh
 # (SIGTERM -> grace -> SIGKILL; in-flight sweeps survive by design — this
@@ -120,6 +129,35 @@ locate_daemon_bin() {
 # "loom-daemon 0.15.0 (commit ab12cd3, built 2026-07-26T12:00:00Z)" -> ab12cd3
 extract_commit() {
     echo "$1" | grep -oE 'commit [0-9a-f]+' | head -n1 | awk '{print $2}'
+}
+
+# verify_destination_binary <dest_path> — assert the provisioned binary at
+# <dest_path> embeds the expected source-HEAD commit (#4053). This is the
+# direct answer to "reports success while shipping nothing": after a provision
+# step returns success, the destination must actually be the freshly-built
+# binary. Exits 5 on mismatch — distinguishable from a compile failure (exit 1)
+# and from a provisioning soft-failure. Skipped only when the source HEAD is
+# unknown (a tarball build with no .git), where there is nothing to compare
+# against. Relies on $SOURCE_COMMIT being resolved (it is, before any build).
+verify_destination_binary() {
+    local dest="$1"
+    if [[ "$SOURCE_COMMIT" == "unknown" ]]; then
+        warn "Source HEAD is unknown (no .git?) — skipping post-provision verification."
+        return 0
+    fi
+    if [[ -z "$dest" || ! -x "$dest" ]]; then
+        err "Post-provision verification FAILED: provisioning reported success but no executable binary was found at the destination ('${dest:-<unknown>}')."
+        exit 5
+    fi
+    local dest_version dest_commit
+    dest_version=$("$dest" --version 2>/dev/null || true)
+    dest_commit=$(extract_commit "$dest_version")
+    if [[ "$dest_commit" != "$SOURCE_COMMIT" ]]; then
+        err "Post-provision verification FAILED: destination binary at $dest embeds commit '${dest_commit:-<none>}' but the expected source HEAD is '$SOURCE_COMMIT'."
+        err "Provisioning reported success yet the destination is NOT the freshly-built binary — a silent no-op roll. This is distinct from a compile failure and from a provisioning soft-failure; refusing to report success."
+        exit 5
+    fi
+    ok "Post-provision verification: destination binary at $dest embeds source HEAD commit ($dest_commit)."
 }
 
 # ---------- args ----------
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!**

Reviewed against issue #4053's acceptance criteria and scope (`Affected Files` list matches the PR diff exactly — no scope creep).

## Correctness

- **`loom-daemon/build.rs`** — the new watch set resolves `HEAD`, `index`, `packed-refs`, and the current branch's symbolic-ref target via `git rev-parse --git-path`, emitting only paths that exist (correctly avoiding the "every-build recompile" trap for nonexistent `rerun-if-changed` paths). This is correct-by-construction across main checkout, linked worktree (`.git` as a file), detached HEAD, and packed-refs.
- **`loom-daemon-update.sh`** — new exit 4 (built-commit ≠ source HEAD, refuses to provision) and exit 5 (post-provision destination mismatch) are cleanly distinguished from the existing exit 1 (compile/usage failure) and exit 3 (`--check` update-available), and documented in the script's own exit-code table.
- **`provision-daemon.sh`** — confirmed the `--version`-equality short-circuit (`src_ver == dest_ver`) is untouched; `PROVISIONED_DAEMON_BIN` is now set as a global *before* every return path (including the short-circuit and the early soft-failure returns), giving the caller a destination to verify without duplicating the `LOOM_DAEMON_BIN_DIR` default-resolution logic.

## On the self-reported caveat (packed→loose ref transition)

Confirmed this is a real, narrow gap: if `packed-refs` is the ref's location at build time and a later commit creates a *new* loose `refs/heads/<branch>` file (without touching `packed-refs` itself, as `git commit-tree` + `update-ref` does), that new file wasn't in the watch set at emission time, so cargo won't re-trigger the build script on its appearance alone. This is a legitimate consequence of "only watch what exists at build time" (which is itself required to avoid the every-build-recompile trap).

This is an acceptable tradeoff, not an acceptance-criteria violation: `HEAD`/`index`/`packed-refs` are still watched as siblings and will likely catch most real-world sequences, and — decisively — the built-commit verification in `loom-daemon-update.sh` (exit 4) is exactly the backstop for this: a stale bake now **fails loudly and refuses to provision** rather than silently shipping. That satisfies the issue's actual ask ("a roll can never silently ship nothing"), which is a detection/fail-safe contract, not a claim that the watch set is exhaustive.

## Test execution

- `bash defaults/scripts/tests/test-build-rs-watchset.sh` — 10/10 pass (main checkout, linked worktree, detached HEAD, packed-refs).
- `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 23/23 pass (includes new exit-4 and exit-5 cases).
- `cargo test --workspace` — not completed locally (this box has several concurrent cargo builds from other worktrees contending for CPU, a known environment issue); CI's "Rust Unit Tests" job already runs `cargo test --workspace` on this PR and passed, so this is covered.
- CI: all required checks pass (Detect Changes, Shellcheck, Rust Linting & Build Check, Rust Unit Tests, Installer Integration Tests; MCP Server Build correctly skipped — no relevant changes).
- `mergeStateStatus`: CLEAN.

## Minor, non-blocking note

`scripts/install/provision-daemon.sh` is outside `defaults/scripts/`, so it isn't covered by the CI shellcheck job's glob — running `shellcheck` on it locally flags `SC2034` on the new `PROVISIONED_DAEMON_BIN="$dest_bin"` global assignment (a false positive, since it's consumed by the sourcing caller `loom-daemon-update.sh`, which shellcheck can't see across the `source`). Not blocking; not introduced-by-this-PR CI gap.

Doc update in `.loom/docs/daemon-reference.md` accurately describes the new behavior and exit codes. Nice work closing out a real, previously-verified root cause.

PR #4076, opened 2026-07-27
feat(daemon): expose CPU headroom tunables via autonomous.* config, not env-only

### LINKED ISSUE #4032: feat(daemon): expose CPU/disk headroom tunables via autonomous.* config, not env-only

rjwalters:
## Summary

`utilization_target` and `est_cores_per_sweep` (the two knobs feeding the CPU headroom term, #3978/#4031) were env-var only, out of step with every other autonomous-dispatch knob (`workFinder`, `mainHealthGate`, `perTokenConcurrency`, `tokenRankingRefresh`, `roleRunner`), all of which resolve **env > `.loom/config.json` → `autonomous.*` > default**.

This PR brings the two CPU knobs in line with that pattern:

- `autonomous.cpuUtilizationTarget` / `autonomous.estCoresPerSweep` now resolve **env > config > default**, matching `resolve_per_token_concurrency` exactly.
- Resolution is **single-root and startup-time** (not per-workspace — the dynamic cap is one global number per tick, computed before the workspace registry is even loaded; matches the curated scope of #4032).
- Absent `autonomous` block, absent keys, malformed JSON, wrong JSON type (string/bool/null where a number is expected), and out-of-range values (`cpuUtilizationTarget` outside `(0, 1]`, `estCoresPerSweep <= 0`) all soft-fail to the built-in defaults — not clamped, no panic. Zero behavior change for anyone not opting in.
- Integer JSON (`"estCoresPerSweep": 2`) works, not just float (`2.0`) — `Value::as_f64` handles both.
- `LOOM_CPU_UTILIZATION_TARGET` / `LOOM_EST_CORES_PER_SWEEP` keep working and keep winning over config.
- `loom-daemon --status` (`ipc.rs`) resolves through the same path as dispatch — the existing `read_work_finder_config` call there was hoisted above the CPU snapshot computation (not a new config read, just a reorder).

### `LOOM_PER_WORKTREE_GB` decision (recorded, not implemented) — **defer**

`disk_headroom.rs`'s `LOOM_PER_WORKTREE_GB` is deliberately **not** migrated to the same `autonomous.*` pattern in this PR. It has a second, independent Bash-side reader (`defaults/scripts/lib/disk-headroom.sh`, consumed by `/loom:sweep` Stage -1 wave sizing) alongside the Rust `disk_headroom::per_worktree_gb`. Migrating only the Rust side would make the same env var honor `.loom/config.json` on the daemon path while silently ignoring it on the sweep path — a worse, divergent state than today's consistent env-only behavior. Moving both means wiring the Bash `config-resolver.sh` into `disk-headroom.sh` too, which is a separate, larger change. The decision and rationale are recorded in `cpu_headroom.rs`'s module docs and `.loom/docs/daemon-reference.md`; a follow-up issue can pick it up if that cost is judged worth paying.

## Implementation

- `loom-daemon/src/cpu_headroom.rs`: split `env_utilization_target()` / `env_est_cores_per_sweep()` (env-only) from `resolve_utilization_target(config)` / `resolve_est_cores_per_sweep(config)` (env > config > default). `cpu_headroom_limit()` / `cpu_headroom_snapshot()` now take the two resolved `f64`s as parameters instead of reading env directly.
- `loom-daemon/src/work_finder.rs`: `WorkFinderConfig` gains `cpu_utilization_target: Option<f64>` / `est_cores_per_sweep: Option<f64>`, parsed in `read_work_finder_config` at the `autonomous` level (alongside `perTokenConcurrency`, so read even without a `workFinder` block) with range filters matching the env resolvers. New `resolve_cpu_utilization_target(&WorkFinderConfig)` / `resolve_cpu_est_cores_per_sweep(&WorkFinderConfig)` wrap the `cpu_headroom` resolvers. Both work-finder loops (`spawn_work_finder_task`, `spawn_multi_work_finder_task`) now take the two resolved values as parameters and pass them into the per-tick `spawn_blocking` call to `cpu_headroom_limit`.
- `loom-daemon/src/main.rs`: resolves both knobs once at startup beside `per_token_concurrency` and threads them into `spawn_multi_work_finder_task`.
- `loom-daemon/src/ipc.rs`: hoists the existing `read_work_finder_config` call above the CPU snapshot computation so `loom-daemon status` and dispatch resolve through the same path.
- `.loom/docs/daemon-reference.md`: updated the CPU-headroom section and the `autonomous.*` config table with the two new keys and the disk-deferral rationale.
- `docs/measure-est-cores-per-sweep.md`: updated the "Apply" section to show both the ephemeral env path and the durable config path.

## Test plan

- [x] Unit — config parsing: full block, `autonomous`-only (no `workFinder`), integer JSON, out-of-range, wrong JSON type (string/bool/null) all covered in `work_finder.rs` tests.
- [x] Unit — precedence: env set + config set → env wins; env unset + config set → config wins; both unset → default. Covered in both `cpu_headroom.rs` (`resolve_*_precedence`) and `work_finder.rs` (`test_resolve_cpu_*_precedence`).
- [x] Unit — soft-fail at both layers, not clamped, no panic.
- [x] Regression — existing `read_work_finder_config` / `resolve_dynamic_max_concurrent` tests pass unchanged.
- [x] Env-mutating tests are `#[serial]`.
- [x] `cargo build --workspace --all-targets` clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] `cargo test --lib` — 1028 passed, 0 failed.

**Pre-existing failure, unrelated to this change**: `cargo test --workspace` also runs `loom-daemon/tests/integration_basic.rs`, which flakes on `test_create_terminal_with_working_dir` / `test_destroy_terminal` — tmux-session-management integration tests unrelated to `cpu_headroom`/`work_finder`/`ipc`/`main`. Reproduced in isolation (re-running just that test file) with a different one of the two tests failing each run, confirming environmental flakiness rather than a regression from this change.

Closes #4032

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1174,9 +1174,29 @@ capacity, unadjusted. Like the token axis's "one healthy account is the floor,
 never a halt" policy (#3902), the CPU term is floored at `1` — a read failure or
 a noisy reading must never by itself wedge the whole dynamic cap to zero; disk
 headroom and the token axis remain the only terms allowed to floor to a genuine
-`0`. Tunable via `LOOM_CPU_UTILIZATION_TARGET` (fraction, default `0.75`) and
-`LOOM_EST_CORES_PER_SWEEP` (cores, default `2.0`) — env-only, mirroring
-`LOOM_PER_WORKTREE_GB`'s config surface (no `.loom/config.json` knob).
+`0`. Tunable with the standard precedence **env (`LOOM_CPU_UTILIZATION_TARGET`
+/ `LOOM_EST_CORES_PER_SWEEP`) > config (`autonomous.cpuUtilizationTarget` /
+`autonomous.estCoresPerSweep`) > default (`0.75` fraction / `2.0` cores)**
+(#4032) — resolved single-root, at startup, from the same
+`WorkFinderConfig`/`read_work_finder_config` that `perTokenConcurrency` uses
+(not per-workspace: the dynamic cap is one global number per tick, computed
+before the workspace registry is even loaded). An out-of-range or
+wrong-JSON-type config value (`cpuUtilizationTarget` outside `(0, 1]`,
+`estCoresPerSweep <= 0`, a string/bool/null where a number is expected) is
+dropped to `None` at the config-read layer, not clamped, so it falls straight
+through to env/default resolution.
+
+**`LOOM_PER_WORKTREE_GB` deliberately stays env-only (#4032 decision).**
+Unlike the CPU knobs, `disk_headroom`'s `LOOM_PER_WORKTREE_GB` was *not*
+migrated to the same `autonomous.*` pattern. It has a second, independent
+Bash-side reader (`defaults/scripts/lib/disk-headroom.sh`, consumed by
+`/loom:sweep` Stage -1 wave sizing) alongside the Rust
+`disk_headroom::per_worktree_gb`. Migrating only the Rust side would create a
+live divergence where the same env var honors `.loom/config.json` on the
+daemon path while silently ignoring it on the sweep path — worse than today's
+consistent env-only behavior. Wiring the Bash `config-resolver.sh` into
+`disk-headroom.sh` too is a separate, larger change; file a follow-up if that
+cost is judged worth paying.
 
 *Calibrating `LOOM_EST_CORES_PER_SWEEP`.* The host-side half of the term
 (consumed cores) is now measured continuously and needs no tuning. The one
@@ -1329,6 +1349,8 @@ concurrency ceiling 5" and share it with the team:
   "autonomous": {
     "model": "sonnet",
     "perTokenConcurrency": 2,
+    "cpuUtilizationTarget": 0.75,
+    "estCoresPerSweep": 2.0,
     "workFinder": {
       "enabled": true,
       "intervalSecs": 60,
@@ -1384,6 +1406,8 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.workFinder.quarantine.ttlSecs` | `LOOM_WORK_FINDER_QUARANTINE_TTL_SECS` | `3600` | How long a quarantine entry persists before auto-release. Zero/invalid → default |
 | `autonomous.workFinder.quarantine.instaCrashSecs` | `LOOM_WORK_FINDER_QUARANTINE_INSTA_CRASH_SECS` | `60` | Checkpoint-less death within this window of dispatch counts as an insta-crash. Zero/invalid → default |
 | `autonomous.perTokenConcurrency` | `LOOM_PER_TOKEN_CONCURRENCY` | `2` | Concurrent sweeps **per healthy token** in the cap (#3947). Zero/invalid → default; clamped to a floor of 1 |
+| `autonomous.cpuUtilizationTarget` | `LOOM_CPU_UTILIZATION_TARGET` | `0.75` | Fraction of logical CPUs the CPU headroom term is willing to dedicate to sweep work (#3978, config surface #4032). Outside `(0, 1]` or wrong JSON type → default; single-root, resolved once at startup (not per-workspace) |
+| `autonomous.estCoresPerSweep` | `LOOM_EST_CORES_PER_SWEEP` | `2.0` | Estimated CPU cores one concurrent sweep consumes while building/testing (#3978, config surface #4032). `<= 0` or wrong JSON type → default; integer JSON (`2`) and float JSON (`2.0`) both accepted; single-root, resolved once at startup |
 | `autonomous.mainHealthGate.enabled` | `LOOM_MAIN_HEALTH_GATE` | `false` | Gate loop on/off |
 | `autonomous.mainHealthGate.ciWorkflow` | `LOOM_GATE_CI_WORKFLOW` | *(unset)* | Forge workflow that must itself conclude `success` for forge-CI corroboration to vouch for a commit (#3987). Empty/whitespace → unset. Absent → today's unanimity rule, unchanged. See [Optional named verification workflow](#optional-named-verification-workflow-loom_gate_ci_workflow-3987) |
 | `autonomous.roleRunner.enabled` | `LOOM_ROLE_RUNNER` | `false` | Periodic standalone support-role runner on/off (#4015) |
--- a/docs/measure-est-cores-per-sweep.md
+++ b/docs/measure-est-cores-per-sweep.md
@@ -72,11 +72,24 @@ build concurrency, which is the moment the term exists to guard against.
 
 ### 4. Apply
 
-Set the measured value (env overrides config overrides the `2.0` default):
+Set the measured value with precedence **env > `.loom/config.json` >
+default** (#4032) — pick whichever durability you want:
 
 ```bash
+# Ephemeral (this shell's daemon runs only): env wins over any committed config.
 export LOOM_EST_CORES_PER_SWEEP=<measured>
 # then restart the daemon, or bake it into the daemon's environment
+
+# Durable, shared with the team, and survives loom-daemon-start.sh re-rendering
+# the plist on every start (unlike an env var, which only persists as long as
+# every start exports it):
+```
+```json
+{
+  "autonomous": {
+    "estCoresPerSweep": <measured>
+  }
+}
 ```
 
 Cross-check against `loom-daemon status`, whose `cpu headroom` line now reports
@@ -86,6 +99,9 @@ the measured idle fraction and consumed-core estimate feeding the term.
 
 - **`LOOM_CPU_UTILIZATION_TARGET`** (default `0.75`) is a policy knob — the
   fraction of cores you are *willing* to dedicate to sweeps — not a measurement.
-  Leave headroom for the OS, the daemon, and the build-gate's own `cargo`.
-- Exposing these two knobs via `.loom/config.json` (`autonomous.*`) instead of
-  env-only is tracked separately in #4032.
+  Leave headroom for the OS, the daemon, and the build-gate's own `cargo`. It
+  has the same committed-config counterpart, `autonomous.cpuUtilizationTarget`.
+- Both knobs are resolved **once at daemon startup**, single-root (not
+  per-workspace) — see the `autonomous.*` config surface in
+  [`.loom/docs/daemon-reference.md`](../.loom/docs/daemon-reference.md)
+  (#4032).
--- a/loom-daemon/src/cpu_headroom.rs
+++ b/loom-daemon/src/cpu_headroom.rs
@@ -97,11 +97,33 @@ use std::time::{Duration, Instant};
 use std::process::{Command, Stdio};
 
 // ============================================================================
-// Configuration (env-only, mirrors `LOOM_PER_WORKTREE_GB` in disk_headroom)
+// Configuration (env > config > default — #4032)
 // ============================================================================
+//
+// Both knobs resolve **env > `.loom/config.json` → `autonomous.*` > default**,
+// matching every other autonomous-dispatch knob (`workFinder`,
+// `mainHealthGate`, `perTokenConcurrency`, `tokenRankingRefresh`,
+// `roleRunner`). Resolution is single-root and startup-time — read once from
+// the daemon's primary workspace via [`crate::work_finder::WorkFinderConfig`]
+// and threaded through as plain `f64`s — exactly like
+// [`crate::work_finder::resolve_per_token_concurrency`], not per-workspace
+// (the dynamic cap is one global number per tick; see
+// `crate::work_finder::spawn_multi_work_finder_task`'s module docs).
+//
+// `LOOM_PER_WORKTREE_GB` in [`crate::disk_headroom`] deliberately does **not**
+// get the same treatment here (#4032 decision, recorded rather than
+// implemented): it has a second, independent Bash-side reader
+// (`defaults/scripts/lib/disk-headroom.sh`, consumed by `/loom:sweep` Stage -1
+// wave sizing). Migrating only the Rust side would make the same env var honor
+// `.loom/config.json` on the daemon path while silently ignoring it on the
+// sweep path — a worse, divergent state than today's consistent env-only
+// behavior. Moving both means wiring the Bash `config-resolver.sh` into
+// `disk-headroom.sh` too, which is a separate, larger change. File a follow-up
+// if that cost is judged worth paying.
 
 /// Environment variable overriding the fraction of logical CPUs the dynamic
-/// cap is willing to dedicate to sweep work.
+/// cap is willing to dedicate to sweep work. Wins over
+/// `autonomous.cpuUtilizationTarget`.
 pub const UTILIZATION_TARGET_ENV: &str = "LOOM_CPU_UTILIZATION_TARGET";
 
 /// Default CPU utilization target: 75% of logical cores. Leaves headroom for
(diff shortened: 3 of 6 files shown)


rjwalters:
✅ **Approved!**

Verified against the #4032 acceptance criteria:

- `autonomous.cpuUtilizationTarget` / `autonomous.estCoresPerSweep` resolve **env > config > default**, matching `resolve_per_token_concurrency` exactly (`cpu_headroom.rs`: `env_*` split from `resolve_*(config)`; `work_finder.rs`: two new `Option<f64>` fields on `WorkFinderConfig` read at the `autonomous` level, not just under `workFinder`).
- Resolution is single-root, startup-time (not per-workspace) — correctly matches the curated scope in the issue (Correction 1), threaded once through `main.rs` into `spawn_multi_work_finder_task`.
- Absent block / malformed JSON / wrong JSON type / out-of-range values all drop to `None` at the config-read layer (not clamped), soft-failing to the built-in defaults — confirmed by `test_config_out_of_range_*_drops_to_none` and `test_config_wrong_json_type_cpu_knobs_drop_to_none`.
- Integer JSON (`"estCoresPerSweep": 2`) verified via `test_config_integer_cpu_knobs_parse_as_f64`.
- Env vars still win over config (`resolve_*_precedence` tests in both `cpu_headroom.rs` and `work_finder.rs`).
- `ipc.rs`'s `build_daemon_status` now resolves through the same `read_work_finder_config` path as dispatch (hoisted, not a new read) — `loom-daemon --status` and dispatch can't disagree.
- The `LOOM_PER_WORKTREE_GB`/`disk_headroom.rs` decision is explicitly recorded (deferred, with rationale: second independent Bash-side reader in `disk-headroom.sh` would create a path divergence) in both the `cpu_headroom.rs` module doc and `.loom/docs/daemon-reference.md`.
- `.loom/docs/daemon-reference.md` config table and `docs/measure-est-cores-per-sweep.md` "Apply" section updated with both the env and config paths.

## Test Execution

- `cargo build --workspace` — clean.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test --lib` — 1028 passed, 0 failed (ran locally, reproducing the builder's claim).
- Confirmed no other call sites of `cpu_headroom_limit`/`cpu_headroom_snapshot` were missed (only a doc-comment reference in `capacity.rs`).
- CI: `Rust Linting & Build Check` and `Rust Unit Tests` both pass (waited out the initial `pending` state before evaluating). `mergeStateStatus` is now `CLEAN`.

No pre-existing `integration_basic.rs` tmux flake surfaced in this run since I scoped to `--lib`, consistent with the PR's noted caveat.

Nice, tightly-scoped follow-through on the Curator's corrections — good call deferring the disk-side migration rather than scope-creeping into the Bash resolver.

Issue #4011, opened 2026-07-27
Silent autonomy loss: daemon left booted-out of launchd, dispatch stopped, nothing reported it


rjwalters:
## Summary

On 2026-07-26 at 19:52:13 a daemon started and took SIGTERM two seconds later, leaving the launchd service **unloaded**. Autonomous dispatch stopped completely. Nothing reported it — no log line an operator would see, no forge signal, no notification. The outage was discovered only because I happened to run `loom-daemon status` by hand some time later.

The system's whole value proposition is running unattended. An autonomy loss that is silent until a human idly checks is the worst possible failure shape.

## Contributing factors

- The sweep in flight at the time was for **#3972 / #3996** — the change that moves the daemon to a launchd LaunchAgent. So the daemon was being killed by work that was itself modifying the daemon's own launchd path. Self-modification of the supervision path is inherently racy and deserves its own guard.
- `KeepAlive` is `<false/>` in `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist`, so launchd does not restart the job on exit. Note that flipping `KeepAlive` to `true` would **not** have covered this case: the service was `bootout`-ed, i.e. unloaded entirely, and an unloaded job has no keepalive semantics. A restart policy is a partial mitigation at best, not the fix.
- `RunAtLoad` is `true`, so the daemon returns on next login/reboot — which masks the failure over long enough horizons and makes it easy to under-count how often this happens.

## Acceptance criteria

1. Detect and report the state "a daemon should be running but isn't." Concretely: something durable outside the daemon process records that autonomy is *expected* to be on (the launchd job is installed / the PID file exists / config enables the work finder), and surfaces a divergence.
2. The report reaches an operator without them thinking to look — the current failure is precisely that `loom-daemon status` is a pull. Options: a forge-side signal, a launchd watchdog job, or a heartbeat file with a staleness threshold that some already-scheduled surface checks.
3. `loom-daemon status` distinguishes "daemon not running" from "daemon not installed" — a bare connection failure reads the same for both today.
4. Guard the self-modification case: a sweep that touches the daemon's own launchd plist / binary should either be serialized against daemon lifecycle operations or explicitly re-bootstrap the service afterwards and verify it came back.

## Notes

- Sibling of the observability gap in #3971 (operator watches die with the session) — both are "the system knows something an operator needs, and has no path to tell them."
[...]

PR #4077, opened 2026-07-27
feat(daemon): supervised restart primitive via launchd KeepAlive:SuccessfulExit

### LINKED ISSUE #4054: feat(daemon): supervised restart primitive — prove the daemon can end and reliably come back under launchd
## Context

Phase 2 of #4017 (*auto-rebuild and restart when the running binary is stale*). #4017 identifies the hard part precisely:

> **A process cannot reliably restart itself.**

This issue isolates that unknown and delivers a **manually-triggerable restart primitive** that is proven to work under launchd. It deliberately ships **no** automation — no periodic loop, no staleness-triggered action. That is #4017 Phase 3, and it must not be built on an unproven restart path.

## Why the naive implementation is unsafe (verified against `origin/main` @ `789669e9`)

The macOS LaunchAgent plist is generated by `render_launchd_plist` in `defaults/scripts/cli/loom-daemon-start.sh:158-190`:

- `RunAtLoad` → `<true/>` (`loom-daemon-start.sh:183`)
- `KeepAlive` → `<false/>` (`loom-daemon-start.sh:184`) — present and **explicitly false**, with the rationale at `:144-146`: *"launchd does not auto-respawn a crashed daemon; that stays the reaper/operator's job"*

And `defaults/scripts/cli/loom-daemon-stop.sh` calls `launchctl bootout "$LAUNCHD_SERVICE"` (`:117`, in `launchd_bootout_if_loaded()` at `:115-119`) from both stop paths (`:144`, `:175`).

So there is **no supervisor that will bring the daemon back**. If the daemon spawns `loom-daemon-update.sh` as a child, that script is inside the launchd job's process tree, and its call to `loom-daemon-stop.sh` → `launchctl bootout` tears the tree down — killing the helper that was supposed to start the replacement. The failure mode is *worse than the status quo*: an unattended daemon that takes itself down permanently.

## Curator verification (against `origin/main` @ `9455b24b`, 2026-07-27)

Every file and line reference above was re-checked and is **exact** except the
`daemon-reference.md` ranges, which were stale and have been corrected in place.
Two findings materially change the work:

### Finding 1 — Option 1 as stated has a race that breaks "an operator stop stays stopped"

The issue names this as *"the single most important thing to prove"*. It is
currently **broken by construction**, and the mechanism is concrete:

- `loom-daemon/src/main.rs:872-881` — the shutdown handler treats **SIGINT and
  SIGTERM identically** and ends in an unconditional `std::process::exit(0)`.
  An operator stop is therefore a **clean exit**.
- `defaults/scripts/cli/loom-daemon-stop.sh:154` sends SIGTERM, waits for the
  process to die, and only **after the process is confirmed dead** calls
  `launchd_bootout_if_loaded` (`:175`).

Under `KeepAlive: { SuccessfulExit: true }`, launchd sees a clean exit while the
job is **still loaded** and relaunches the daemon *in the window before the
bootout lands*. The bootout then tears down the freshly relaunched instance —
so the observable outcome depends on a timing window, and a daemon that is
mid-startup (binding the socket, writing the PID file, seeding dispatch) can be
killed in an indeterminate state.

Worse, `launchd_bootout_if_loaded` (`loom-daemon-stop.sh:115-119`) swallows all
errors (`>/dev/null 2>&1 || true`), and the stop script's success check only
re-tests the **original** pid. If the bootout fails, `loom-daemon-stop.sh` prints
`loom-daemon stopped` while a relaunched daemon is alive and dispatching — the
same silent divergence between believed and actual autonomy state as the #4011
incident, inverted.

**Remedy to evaluate first: make the exit code carry intent.** `SuccessfulExit`
keys off exit status, and the daemon already knows *why* it is shutting down:

| Shutdown cause | Exit code | launchd behavior under `SuccessfulExit: true` |
|---|---|---|
| Restart primitive (this issue) | `0` | relaunch — the desired path |
| SIGTERM (operator stop) | non-zero (e.g. `143`) | **no relaunch** — no race |
| SIGINT (interactive Ctrl-C) | non-zero (e.g. `130`) | no relaunch |
| Crash / panic | non-zero | no relaunch (preserves today's no-crash-loop semantics) |

This makes the restart the *only* thing that trips launchd, and demotes the
bootout from sole defense to belt-and-braces. It is a targeted change at
`main.rs:880`. Verify nothing depends on the daemon exiting `0` on SIGTERM —
`loom-daemon-stop.sh` checks liveness with `kill -0`, not the exit status, so
it is not a consumer.

### Finding 2 — the singleton guard is socket-based, which is Option 2's real hazard

`main.rs:864` states the guard trips on a **stale socket**. So an `exec`
self-replacement must `remove_file` the socket *before* `exec`, or the re-exec'd
image trips its own singleton guard on the socket it still owns. Also confirm the
listening socket fd is `FD_CLOEXEC` so it is not inherited across the `exec`.

Option 2's genuine advantage is the flip side: the pid is preserved, so the
launchd job, the PID file, and `.loom/.daemon.flags` all stay valid with **no
plist change at all** — which sidesteps Finding 1 entirely. Weigh that against
the socket/state handling before defaulting to Option 1.

### Related work (not dependencies, but read before implementing)

- **#4011** (`loom:curated`) — the incident that motivates #4017. It establishes
  that **`bootout` unloads the job entirely, and an unloaded job has no keepalive
  semantics**, which is the reason an operator stop *can* still stick under
  Option 1 once the bootout lands. Its acceptance criterion 4 also asks to guard
  the **self-modification** case: *a sweep that touches the daemon's own launchd
  plist should be serialized against daemon lifecycle operations, or re-bootstrap
  and verify*. Implementing this issue **is** that scenario — the PR that rolls
  this change is itself a sweep modifying the supervision path, exactly the
  #4011 shape. Plan the first roll deliberately.
- **#4069** (`loom:triage`) — `loom-daemon status` cannot distinguish "not
  installed" / "not running" / "running but unresponsive". This is a
  **verification obstacle** for the mandatory manual test below: proving "the
  daemon came back" needs a signal stronger than a bare connection attempt. Use
  `launchctl print` plus a real IPC round-trip, not `status` alone.
- **#4053** (`loom:issue`, `loom:curated`) — Phase 1, self-verifying
  rebuild→provision. Independent of this issue; no ordering constraint.

## Options (from #4017, in order of preference)

1. **Split rebuild from restart.** The daemon does the safe part in-process and then only needs to *end*; restart becomes the supervisor's job. On macOS: `KeepAlive: { SuccessfulExit: true }` — launchd relaunches after a **clean** exit but still does not respawn a crash loop, preserving today's semantics. `bootout` in the stop path continues to make an operator stop stick.
2. **`exec` self-replacement.** After installing, the daemon `exec()`s the new binary in place — same pid, launchd never notices, no plist change. Cleanest where it works; needs care with the IPC socket, the PID file, and in-flight state.
3. **Detached helper.** Only if 1 and 2 fail. Requires genuinely escaping the launchd job, plus a proof that the helper survives `bootout`.

Pick one, implement it, and **write down why** — `docs/design/` is the established home for this kind of decision record (see `docs/design/config-resolution-tiers.md`, which plays the same role for Epic #3835). A short design note is in scope here.

## Non-macOS

The `nohup` path (`loom-daemon-start.sh:366-367` falls back when `launchctl` is absent) has the same gap and **no supervisor at all**. Per #4017: degrade to *"log loudly, leave the daemon running, do not restart"* rather than guessing.

## Interaction with the existing stop/start contract

Whatever is chosen must preserve all of these, which are load-bearing today (`.loom/docs/daemon-reference.md` § *Safe start / stop (raw daemon process)*, `:1695-1758` — the "survive, don't drain" shutdown decision is at `:1751-1757`):

- **In-flight sweeps survive.** Sweep children are documented to outlive a daemon stop; a restart must not cancel them.
- **Flags replay exactly, never wider.** `.loom/.daemon.flags` is written by `loom-daemon-start.sh:318-328` and replayed verbatim by `loom-daemon-update.sh:244-250`. A restart must not widen autonomy as a side effect.
- **An operator stop must stay stopped.** `KeepAlive: SuccessfulExit` would make launchd relaunch on clean exit — verify `bootout` in `loom-daemon-stop.sh:117` still wins, since it unloads the job entirely. **This is the single most important thing to prove if option 1 is chosen**, because getting it wrong means an operator can no longer stop the daemon.
- **A daemon that was not running is never started.**

## Scope

- Choose the restart mechanism and record the decision (short note under `docs/design/`).
- Implement it, including the plist change if option 1 is chosen.
- Provide a way to *trigger* a restart deliberately (e.g. an IPC request, or a documented signal) so Phase 3 has a primitive to call. It must not fire on its own.
- Update the plist test, which currently pins the values this issue may change.

## Out of scope

- Any staleness detection, periodic loop, config knob, backoff, or `--status` field (#4017 Phase 3).
- Rebuild/provision correctness (#4053).

## Affected Files

- `defaults/scripts/cli/loom-daemon-start.sh` — `render_launchd_plist` (`:158-190`), specifically `KeepAlive` at `:184`; the `--print-plist` inspection path (`:217`, `:357-360`) makes this testable with no `launchctl` side effects.
- `defaults/scripts/cli/loom-daemon-stop.sh` — verify `launchd_bootout_if_loaded` (`:115-119`, called `:144`, `:175`) still makes an operator stop stick under the new `KeepAlive` value.
- `defaults/scripts/tests/test-loom-daemon-launchd-plist.sh` — **`:111-112` currently assert `RunAtLoad` true / `KeepAlive` false**; `:112` must be updated if option 1 is chosen. Do not delete the assertion — re-pin it to the new intended value.
- `loom-daemon/src/ipc.rs` / `loom-daemon/src/types.rs` — if the trigger is an IPC request, add the `Request`/`Response` variants (existing pattern: `Response::DaemonStatus` at `types.rs:487`, dispatch at `ipc.rs:272-274`).
- `loom-daemon/src/main.rs` — shutdown path wiring for a clean exit-for-restart.
- `docs/design/` — new short design note recording the chosen mechanism and why the other two were rejected.
- `.loom/docs/daemon-reference.md` — Self-update section (`:1918-1988`) and the shutdown-semantics discussion it references (`§ Safe start / stop`, `:1695-1758`; the flags-replay contract is stated at `:1947` and `:1961`).

## Test Plan

- [ ] **Manual verification on real launchd (mandatory — this cannot be proven by unit tests).** On macOS: start the daemon via `loom-daemon-start.sh`, trigger the restart primitive, and confirm (a) the process comes back, (b) it comes back with exactly the flags in `.loom/.daemon.flags`, (c) `launchctl print` shows the job healthy, (d) the new pid differs (option 1) or matches (option 2). Record the transcript in the PR.
- [ ] **Manual verification: an operator stop still sticks.** After the change, run `loom-daemon-stop.sh` and confirm the daemon does **not** come back — including that it stays down across the period launchd would have relaunched it.
- [ ] **Manual verification: in-flight sweeps survive.** Dispatch a sweep, trigger the restart while it is running, confirm the sweep child is still alive afterwards and still logging to `.loom/logs/sweep-issue-<N>.log`.
- [ ] **Manual verification: crash does not respawn.** Kill the daemon with `SIGKILL` and confirm launchd does **not** restart it (this is what `KeepAlive: SuccessfulExit` is supposed to preserve versus plain `KeepAlive: true`).
- [ ] Automated tests: update `defaults/scripts/tests/test-loom-daemon-launchd-plist.sh:111-112` to pin the new intended plist values via `--print-plist`.
- [ ] Automated tests: `defaults/scripts/tests/test-loom-daemon-stop.sh` — assert the bootout path is unchanged.
- [ ] Automated tests: if an IPC trigger is added, a serde round-trip test alongside the existing ones in `loom-daemon/src/ipc.rs` (`:2635-2720`).
- [ ] Edge cases: `launchctl` absent (nohup fallback) → logs loudly, does not restart, daemon keeps running; Linux → same; a daemon started with **no** `.loom/.daemon.flags` → bare FLAGS-OFF restart with a warning, never wider.
- [ ] **Exit-code contract** (if the Finding 1 remedy is taken): assert SIGTERM yields a non-zero exit and the restart trigger yields `0`. A signal-disposition test in `loom-daemon` plus a `test-loom-daemon-stop.sh` assertion.
- [ ] **Stop must not report success while a daemon is alive.** After `loom-daemon-stop.sh` returns `0`, assert no `loom-daemon` process is running (not merely that the original pid is gone) — closes the silent-success hole at `loom-daemon-stop.sh:115-119`/`:175`.
- [ ] Regression: `cargo test --workspace`, `bash defaults/scripts/tests/test-loom-daemon-start.sh`, `bash defaults/scripts/tests/test-loom-daemon-stop.sh`, `bash defaults/scripts/tests/test-loom-daemon-launchd-plist.sh`.

## Acceptance Criteria

- [ ] The restart mechanism is chosen, implemented, and the decision is recorded with the rejected alternatives and why.
- [ ] Restart is **proven under launchd on macOS**, including that the daemon actually comes back — demonstrated with a transcript, not asserted.
- [ ] An operator `loom-daemon-stop.sh` still makes the daemon stay stopped.
- [ ] A crashed (non-clean-exit) daemon is still **not** auto-respawned — no crash loop is introduced.
- [ ] The restarted daemon runs with exactly the flags from `.loom/.daemon.flags`, never wider.
- [ ] In-flight sweeps survive the restart.
- [ ] Where no supervisor exists (nohup / Linux), the daemon logs loudly, leaves itself running, and does not attempt a restart.
- [ ] The "operator stop stays stopped" guarantee holds **without depending on bootout timing** — i.e. there is no window in which launchd relaunches the daemon during an operator stop (see Curator Finding 1).
- [ ] `loom-daemon-stop.sh` exits non-zero if any `loom-daemon` is still alive after it completes, rather than reporting success.
- [ ] The restart primitive never fires on its own — nothing in this PR triggers it automatically.

rjwalters:
## Summary

Phase 2 of #4017 — a **manually-triggerable restart primitive** that is proven to make the daemon end and reliably come back under macOS launchd, so Phase 3 (auto-rebuild-and-restart-when-stale) has a safe path to call. It ships **no** automation; nothing fires it on its own.

**Chosen mechanism: Option 1** (split rebuild from restart, supervised by launchd), including the Curator's Finding 1 remedy (exit code carries intent). Option 2 (`exec` self-replacement) and Option 3 (detached helper) are rejected with rationale in `docs/design/supervised-restart-primitive.md` — briefly: Option 1 is the only option that actually exercises the supervisor this issue is about, and it sidesteps Option 2's socket-based-singleton-guard / `FD_CLOEXEC` hazards on the highest-blast-radius path.

## Changes

- **Plist (`loom-daemon-start.sh`)**: `KeepAlive` flips from bare `<false/>` to `{ SuccessfulExit: true }` — launchd relaunches the job **only** on a clean exit `0`. Bakes `LOOM_DAEMON_SUPERVISOR=launchd` into the plist env so the daemon can *prove* it is supervised before it will exit for a restart.
- **`RestartDaemon` IPC request + `loom-daemon restart` CLI** (`types.rs`, `ipc.rs`, `main.rs`): the **only** path that exits `0`, so the only thing that trips a relaunch. Supervised → replies `DaemonRestart { scheduled: true }`, flushes, then exits 0. Unsupervised (nohup/Linux/`--foreground`) → refuses, replies `scheduled: false`, keeps running (#4017's "log loudly, do not restart").
- **Exit code carries intent** (`main.rs` signal handler + `ipc.rs` Shutdown): SIGTERM→143, SIGINT→130, IPC Shutdown→143, crash→non-zero. Only the restart exits 0. This removes the SuccessfulExit/bootout race — an operator stop stays stopped **without depending on bootout timing** — and preserves the old no-crash-loop semantics.
- **`loom-daemon-stop.sh`**: post-stop, re-verifies (**scoped to its launchd label**, never a global `pgrep`) that no daemon is still alive and exits non-zero if one is — closing the inverted-#4011 silent-success hole. Bootout retained as belt-and-braces.
- **Docs**: `docs/design/supervised-restart-primitive.md` (decision record) + daemon-reference.md Self-update/shutdown sections.

- [x] Mechanism chosen, implemented, decision recorded with rejected alternatives — `docs/design/supervised-restart-primitive.md`.
- [x] **Restart proven under launchd** (isolated label/socket): daemon exited 0, launchd relaunched with a **new pid**, supervision preserved. Transcript below.
- [x] Operator `loom-daemon-stop.sh` stays stopped — SIGTERM → no relaunch within the window, job unloaded, stop exit 0.
- [x] Crash (SIGKILL) not auto-respawned — no crash loop.
- [x] Restarted daemon runs with exactly its start flags (relaunch re-reads the same plist; never wider).
- [x] "Operator stop stays stopped" holds **without depending on bootout timing** — SIGTERM exits non-zero (143).
- [x] `loom-daemon-stop.sh` exits non-zero if any daemon (for its label) is still alive after it completes.
- [x] Unsupervised (nohup/Linux) → logs loudly, stays running, does not restart.
- [x] The primitive never fires on its own.
- [ ] **In-flight sweeps survive the restart** — not exercised live (see Manual follow-up). Holds by construction: restart is a plain process exit + launchd relaunch; sweep children are independent detached processes the daemon never cancels on shutdown.

## Test Plan

- [x] `cargo test --workspace` — green (1015 + suites, 0 failed).
- [x] `cargo clippy --all-targets` — clean.
- [x] `test-loom-daemon-launchd-plist.sh` — re-pinned to `KeepAlive:{SuccessfulExit:true}` + `LOOM_DAEMON_SUPERVISOR=launchd` (26/26).
- [x] `test-loom-daemon-stop.sh` — bootout-unchanged assertion + Darwin fake-launchctl relaunch-still-alive → non-zero exit (10/10).
- [x] `test-loom-daemon-start.sh` — 10/10.
- [x] serde round-trip + supervisor-gating + exit-code-contract unit tests in `ipc.rs`.

### Live launchd transcript (isolated `LOOM_LAUNCHD_LABEL=com.example.loom-4054-live` + isolated `LOOM_SOCKET_PATH`; production `com.rjwalters.loom-daemon` never touched)

```
[A] START reliability daemon (autonomy OFF): pid 44776
[B] loom-daemon restart -> "restart scheduled (supervisor: launchd)"
    PID before restart: 44776   PID after restart: 45485   (NEW pid) ✓
    relaunched env: LOOM_DAEMON_SUPERVISOR=launchd
[C] loom-daemon-stop.sh -> stop exit 0; came back within 10s: no; job unloaded ✓
[D] SIGKILL pid -> respawned within 10s: no ✓ (no crash loop)
[E] foreground SIGTERM exit 143 ✓ ; foreground SIGINT exit 130 ✓
production com.rjwalters.loom-daemon: not loaded (untouched) ; no stray test procs
```

**Isolation note (per Champion's binding constraint):** every launchd operation ran against a throwaway `com.example.loom-4054-live` label + `/tmp` socket. At test time there was **no** production daemon running and the `com.rjwalters.loom-daemon` job was not loaded (verified before and after). The first accidental start inherited `LOOM_WORK_FINDER=1` from the session env; I caught it before the first 60s work-finder tick (no dispatch occurred) and re-ran with autonomy explicitly OFF (`--no-work-finder --no-health-gate`), so no sweep was ever dispatched.

**Manual follow-up for Judge/human:** the "in-flight sweeps survive a restart" AC was not exercised live because dispatching a real sweep runs an actual, side-effecting Claude session; recommend confirming it on a canary before Phase 3 builds on the primitive.

Closes #4054

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1798,6 +1798,17 @@ the registry reconciles their state on the next start (`SweepRegistry::reconstru
 re-admits live-lock owners). To actively cancel a sweep, use
 `mcp__loom__cancel_sweep` against a running daemon *before* stopping it.
 
+**Exit code carries shutdown intent (#4054).** The daemon encodes *why* it is
+shutting down in its exit code, because the launchd `KeepAlive:{SuccessfulExit:true}`
+plist (below) relaunches only on a clean exit `0`: the `RestartDaemon` primitive
+exits `0` (relaunch), SIGTERM exits `143`, SIGINT exits `130`, an explicit IPC
+`Shutdown` exits `143`, and a crash/panic exits non-zero — so only a deliberate
+restart trips a relaunch, and "an operator stop stays stopped" holds without
+depending on `bootout` timing. `loom-daemon-stop.sh` also re-verifies (scoped to
+its launchd label) that no daemon is still alive after the stop and exits non-zero
+if one is, closing the silent-success hole where a failed bootout could leave a
+relaunched daemon dispatching.
+
 ### macOS session-bootstrap hazard (#3972)
 
 **Incident (2026-07-26).** The daemon was started at 21:48 via
@@ -1851,16 +1862,22 @@ after migration reported `13 seen, 3 dispatched, 0 error(s)`.
   `--health-gate` / `--from-config` semantics are preserved exactly — the
   launchd job never sees a wider or narrower autonomy configuration than a
   plain nohup start would have resolved.
-- **`RunAtLoad=true` / `KeepAlive=false`**: mirrors the hand-written plist that
-  validated this fix during the incident. `RunAtLoad=true` means the daemon
-  also survives a reboot/re-login, not just the death of one particular
-  session — strictly *more* durable than the pre-#3972 nohup contract (which
-  didn't survive a reboot either). `KeepAlive=false` means launchd does **not**
-  auto-respawn a crashed daemon; that responsibility stays with the reaper /
-  operator, unchanged from before. `loom-daemon-stop.sh` `bootout`s the loaded
-  definition after confirming the process is dead, specifically so an explicit
-  stop is honored at the next login too (otherwise `RunAtLoad=true` would
-  silently relaunch it).
+- **`RunAtLoad=true` / `KeepAlive={SuccessfulExit:true}`** (#4054): `RunAtLoad=true`
+  means the daemon survives a reboot/re-login, not just the death of one particular
+  session — strictly *more* durable than the pre-#3972 nohup contract (which didn't
+  survive a reboot either). `KeepAlive={SuccessfulExit:true}` (changed from the
+  earlier bare `KeepAlive=false`) is the **supervised restart primitive**: launchd
+  relaunches the job **only** when it exits with status `0`, and leaves it down on
+  any non-zero exit. Because the daemon exits **non-zero** on a crash/panic, on a
+  SIGTERM operator stop (143), and on a SIGINT/Ctrl-C (130), this **preserves** the
+  old no-crash-loop semantics of `KeepAlive=false` — none of those respawn — while
+  making the one deliberate clean-exit path (the `RestartDaemon` request, `loom-daemon
+  restart`) the *only* thing that trips a relaunch. `loom-daemon-stop.sh` still
+  `bootout`s the loaded definition after confirming the process is dead (so an
+  explicit stop is honored at the next login too), but the bootout is now
+  belt-and-braces: the non-zero SIGTERM exit already prevents launchd from relaunching
+  during the stop window. See [Supervised restart primitive](#supervised-restart-primitive-4054)
+  below and `docs/design/supervised-restart-primitive.md`.
 - **Escape hatch**: `--no-launchd` (or `LOOM_DAEMON_LAUNCHD=0`) forces the
   legacy nohup path even on Darwin — e.g. for a sandboxed/headless macOS runner
   with no GUI login session where `gui/<uid>` may not resolve.
@@ -1956,6 +1973,35 @@ changes, verifying `codesign` behavior across the self-update rebuild path,
 deciding whether the identifier needs to be stable across machines) and is
 intentionally left as a follow-up (#4016) rather than bundled into this
 issue's audit + working-set-contract + crash-recovery fix.
+
+### Supervised restart primitive (#4054)
+
+Phase 2 of #4017 (auto-rebuild-and-restart-when-stale). It ships a
+manually-triggerable way for the daemon to **end and reliably come back**, so
+Phase 3 has a *proven* restart path to call. It deliberately ships **no**
+automation — nothing fires it on its own.
+
+```bash
+loom-daemon restart          # send RestartDaemon over the IPC socket
+```
+
+- **Mechanism (macOS):** the plist uses `KeepAlive:{SuccessfulExit:true}` and the
+  daemon exits `0` **only** for a `RestartDaemon` request, so launchd relaunches
+  the job on that one clean exit and leaves it down on every other (SIGTERM 143,
+  SIGINT 130, crash non-zero). The relaunched process re-reads the same plist, so
+  it comes back with **exactly** its start flags/env — never wider. In-flight
+  sweeps survive (they are independent detached processes the daemon never cancels
+  on shutdown). Observable signature: a **new pid**.
+- **Supervision proof:** `loom-daemon-start.sh` bakes `LOOM_DAEMON_SUPERVISOR=launchd`
+  into the plist, and the daemon ends its process for a restart **only** when that
+  var is present. On an unsupervised host (nohup / Linux / `--foreground`) it
+  refuses: `loom-daemon restart` prints the refusal, exits non-zero, and the daemon
+  keeps running — because nothing would bring it back if it exited (#4017's "log
+  loudly, leave the daemon running, do not restart").
+- **Why launchd, not `exec` self-replacement:** the design record — including why
+  the `exec` (Option 2) and detached-helper (Option 3) alternatives were rejected,
+  and the Curator's exit-code-race finding — is in
+  `docs/design/supervised-restart-primitive.md`.
 
 ### Self-update (rebuild + provision + restart, #3968)
 
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -142,12 +142,32 @@ resolve_launchd_label() {
 # (~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist): RunAtLoad=true
 # (the daemon also comes back after a reboot/re-login, not just a session
 # death -- strictly more durable than the pre-#3972 nohup contract, which
-# didn't survive a reboot either) and KeepAlive=false (launchd does not
-# auto-respawn a crashed daemon; that stays the reaper/operator's job, same as
-# before). Lifecycle (first start / explicit stop) is still entirely
-# operator-driven via loom-daemon-start.sh / loom-daemon-stop.sh -- bootout on
-# stop unloads the definition so it does NOT come back at the next login. The
-# PATH is the CURRENT PATH plus a fallback set (~/.local/bin, ~/.cargo/bin,
+# didn't survive a reboot either).
+#
+# KeepAlive is `{ SuccessfulExit: true }` as of the supervised restart primitive
+# (#4054, Phase 2 of #4017): launchd relaunches the job ONLY when it exits with
+# status 0, and leaves it down on any non-zero exit. This is what lets the
+# daemon END and reliably COME BACK on demand -- the `RestartDaemon` IPC request
+# (loom-daemon restart) is the ONLY path that exits 0, so it is the only thing
+# that trips a relaunch. Crucially this PRESERVES the old no-crash-loop semantics
+# of KeepAlive=false: a crashed/panicked daemon, a SIGTERM'd operator stop (exit
+# 143), and a SIGINT/Ctrl-C (exit 130) all exit NON-ZERO, so launchd does NOT
+# respawn them. Making the exit code carry intent (daemon side, #4054) is also
+# what closes the SuccessfulExit/bootout race (Curator Finding 1): an operator
+# stop exits non-zero, so launchd never relaunches it during the stop window --
+# "an operator stop stays stopped" no longer depends on bootout timing. The
+# bootout in loom-daemon-stop.sh is demoted to belt-and-braces (it still unloads
+# the definition so it does not come back at the next login).
+#
+# LOOM_DAEMON_SUPERVISOR=launchd is baked into the plist env so the daemon can
+# PROVE it is supervised before it will exit for a restart. It is hardcoded here
+# (not harvested from the caller's env) so it lands in EVERY rendered plist --
+# and, conversely, is ABSENT from the nohup path (which never renders a plist),
+# so an unsupervised daemon correctly refuses to exit on a restart request
+# (nothing would bring it back). Because it survives in the plist, the relaunched
+# daemon still sees it.
+#
+# The PATH is the CURRENT PATH plus a fallback set (~/.local/bin, ~/.cargo/bin,
 # Homebrew, standard bin dirs) so `gh`, `git`, `cargo`, and `python3` resolve
 # inside the LaunchAgent's minimal launchd environment even if the interactive
 # shell's PATH customizations aren't present there. Every already-exported
@@ -162,12 +182,20 @@ render_launchd_plist() {
     local env_entries=""
     env_entries+="        <key>PATH</key>\n        <string>$(xml_escape "$plist_path_value")</string>\n"
     env_entries+="        <key>HOME</key>\n        <string>$(xml_escape "$HOME")</string>\n"
+    # Mark the daemon as launchd-supervised so its RestartDaemon handler (#4054)
+    # will exit 0 for a supervised relaunch. Hardcoded (not env-harvested) so it
+    # is present in every rendered plist and its relaunch, and never leaks to the
+    # unsupervised nohup path.
+    env_entries+="        <key>LOOM_DAEMON_SUPERVISOR</key>\n        <string>launchd</string>\n"
 
     local line key value
     while IFS= read -r line; do
         [[ -z "$line" ]] && continue
         key="${line%%=*}"
         value="${line#*=}"
+        # Never duplicate the supervisor key hardcoded above (a caller that
+        # exported LOOM_DAEMON_SUPERVISOR must not produce two plist entries).
+        [[ "$key" == "LOOM_DAEMON_SUPERVISOR" ]] && continue
         env_entries+="        <key>$(xml_escape "$key")</key>\n        <string>$(xml_escape "$value")</string>\n"
     done < <(env | grep -E '^(LOOM_[A-Za-z0-9_]*|GH_TOKEN|GITEA_TOKEN|FORGE_TOKEN)=' || true)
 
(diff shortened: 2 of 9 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this complex, high-blast-radius supervision change. The mechanism is sound and the implementation is faithful to the issue's acceptance criteria and the Curator's Finding 1 remedy.

## What I verified

**Exit-code-carries-intent (the load-bearing fix).** `EXIT_RESTART=0` is the *only* exit-0 path; SIGTERM→143, SIGINT→130, IPC `Shutdown`→143, crash→non-zero. The signal handler in `main.rs` maps correctly, and — critically — *both* signal branches are non-zero, so an operator stop is never a "successful" exit and launchd never relaunches it. This closes the SuccessfulExit/bootout race described in Finding 1 **without depending on bootout timing**, exactly as the issue demanded. The bootout is correctly demoted to belt-and-braces, and the pre-#4054 no-crash-loop semantics are preserved.

**Plist change.** `KeepAlive:{SuccessfulExit:true}` renders correctly; `LOOM_DAEMON_SUPERVISOR=launchd` is hardcoded into the plist env (not env-harvested) and de-duped from the harvest loop so a caller who exports it can't produce a duplicate key. Absent from the nohup path by construction.

**Supervisor-detection guard.** `detect_supervisor()` returns `Some("launchd")` only for the case-insensitive match and `None` otherwise; `build_restart_decision()` refuses (`scheduled:false`, `do_exit=false`) on unsupervised hosts (nohup/Linux/`--foreground`), so the daemon never exits with nothing to bring it back. Unit-tested.

**Flags replay-exactly-never-wider.** A relaunch re-reads the same on-disk plist rendered at start, so it comes back with exactly its start flags — the property holds by launchd re-exec, no new mechanism.

**stop.sh still-alive guard.** Correctly scoped to `$LAUNCHD_SERVICE` (`gui/$uid/$LABEL`) via `launchd_job_loaded`/`launchd_job_pid`, never a global `pgrep loom-daemon`, so a test daemon under a non-default label can't false-positive against production. Covered by the new Darwin fake-launchctl relaunch-still-alive test.

**Test coverage / checks (ran locally):**
- `cargo test --workspace` — green (incl. both new `ipc::tests::test_restart_daemon_request_response_round_trip` and `test_build_restart_decision_supervisor_gated`)
- `cargo clippy --all-targets` — clean
- `test-loom-daemon-launchd-plist.sh` 26/26, `test-loom-daemon-stop.sh` 10/10, `test-loom-daemon-start.sh` 10/10
- CI: all required checks pass (Rust Unit Tests, Rust Linting & Build, Installer Integration, Shellcheck); mergeStateStatus CLEAN.

**Design doc** (`docs/design/supervised-restart-primitive.md`) — the rejected-alternatives reasoning is sound. Option 2 (`exec` self-replacement) genuinely sidesteps the supervisor this issue exists to exercise, and its socket-based-singleton-guard / `FD_CLOEXEC` hazards are real on the highest-blast-radius path; Option 3 (detached helper) requires proving bootout-survival, the exact fragility #4017 flagged. Option 1 is the right call for Phase 2.

## On the two disclosed gaps — both acceptable to approve now

1. **In-flight-sweep survival not exercised live.** Acceptable. It holds by construction on *already-tested* detach semantics: the restart is a plain process exit + launchd relaunch, and sweep children are independent detached processes the daemon explicitly never cancels on shutdown (the documented "survive, don't drain" decision, already asserted at `main.rs:875`). This is a Phase-2 primitive that ships **no automation** — nothing fires it — so live sweep-survival verification genuinely matters at Phase 3, and the PR already recommends confirming it on a canary before then. Not a blocker for landing the primitive.

2. **Accidental `LOOM_WORK_FINDER=1` inheritance during manual testing.** This is a disclosure about the *test procedure*, not a defect in the shipped code — it was caught before the first 60s work-finder tick (no dispatch occurred), isolation was maintained (throwaway label + `/tmp` socket, production untouched), and the re-run was with autonomy explicitly OFF. Zero bearing on correctness; the honesty is appreciated.

## Non-blocking note (no action required)

In the nohup path, if an operator has *manually* exported `LOOM_DAEMON_SUPERVISOR=launchd` in their shell before starting, the daemon would inherit it at runtime and wrongly consider itself supervised on a restart request. This requires deliberately setting an internal sentinel var, so it's a genuine operator-error edge, not a defect — flagging only for awareness.

Solid, well-documented, well-tested work on a genuinely tricky piece of the supervision path. LGTM.

PR #4087, opened 2026-07-27
fix(daemon): stop scoped to its launchd label so lifecycle tests can't kill the live daemon

### LINKED ISSUE #4078: fix(tests): test-loom-daemon-update.sh boots out the operator's live launchd daemon (unscoped LOOM_LAUNCHD_LABEL + real launchctl on PATH)
## Summary

`defaults/scripts/tests/test-loom-daemon-update.sh` invokes the **real** `loom-daemon-stop.sh` with the **real** `launchctl` on `PATH` and **no `LOOM_LAUNCHD_LABEL` override**. The stop script therefore falls back to the production label and runs `launchctl bootout gui/<uid>/com.rjwalters.loom-daemon`, which SIGTERMs the operator's live daemon and unloads the LaunchAgent.

Any sweep that touches the daemon updater runs this test as part of its regression suite. It ran **five times on 2026-07-27**. The first run took autonomy down for ~1h20m; the rest looked clean only because the daemon was already dead.

This is distinct from #4042 (updater cannot *manage* a launchd daemon), #4011 (the silent-loss *symptom*), and #4054 (which constrained the *manual* test procedure). This is the **automated** path, which is more dangerous precisely because it runs unattended.

## Evidence

```
18:57:52.230Z  sweep #4053 builder:  bash defaults/scripts/tests/test-loom-daemon-update.sh
18:57:55.799Z  ~/.loom/daemon.log:   [INFO] Received SIGTERM, cleaning up...
```

3.5 seconds. After that: no `work_finder` tick, no `role_runner` tick, and `launchctl list | grep loom` returns nothing — the job is **unloaded**, not merely stopped.

Corroborating details that all fit this mechanism and no other:

- **No transcript anywhere contains an explicit `launchctl bootout ... com.rjwalters.loom-daemon` command in the incident window.** I searched every Claude session transcript under `~/.claude/projects/*loom*/`; the nearest explicit bootout is at 19:14:45Z from #4054's builder, correctly scoped to its throwaway `com.example.loom-4054-live` label. The bootout that did the damage fired *inside* `loom-daemon-stop.sh`.
- `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist` mtime is unchanged (Jul 26 21:36) — nothing rewrote the plist, consistent with a bootout rather than a reinstall.
- `.loom/.daemon.pid` / `.loom/.daemon.flags` never existed on this launchd host (#4042), so the SIGTERM cannot have come from the stop script's own PID-file path. `launchctl bootout` sends SIGTERM to the job itself — which is exactly what the daemon logged.

## Mechanism (verified against `origin/main` @ `9c44c4f1`)

1. `test-loom-daemon-update.sh:14-16, 28` — deliberately copies the **real** `loom-daemon-start.sh` / `loom-daemon-stop.sh` into each throwaway repo root, documented as *"so the restart path exercises the actual scripts, not a mock."* That intent is reasonable; the sandbox is just incomplete.
2. `test-loom-daemon-update.sh:128` — `MINIMAL_PATH="/usr/bin:/bin:/usr/sbin:/sbin"` includes the real `/bin/launchctl`. The test stubs `cargo` and the daemon binary but **not** `launchctl`.
3. `test-loom-daemon-update.sh` sets `LOOM_LAUNCHD_LABEL` **zero** times.
4. `loom-daemon-stop.sh:99` — `LAUNCHD_LABEL="${LOOM_LAUNCHD_LABEL:-com.rjwalters.loom-daemon}"` → production label.
5. `loom-daemon-stop.sh:117` — `launchctl bootout "$LAUNCHD_SERVICE"` → the live job is terminated and unloaded.

The isolation is thorough everywhere except the one axis that reaches outside the sandbox: launchd is **machine-global**, so a throwaway `$HOME`-like repo root buys nothing.

## Proposed fix

Either is sufficient; doing both is better:

- **Scope the label.** Export `LOOM_LAUNCHD_LABEL="com.example.loom-test-$$"` (per-run unique) in the test's environment alongside the existing `PATH` / `LOOM_DAEMON_BIN` overrides, so any real `launchctl` call targets a label that cannot exist.
- **Stub `launchctl`.** Add a fake `launchctl` to `$FAKE_BIN_DIR` next to the fake `cargo`, recording invocations to a file the test can assert on. This is strictly stronger — it makes the test *unable* to touch launchd — and it lets the suite assert the bootout was attempted with the right label, which it currently cannot.

Add a regression assertion that no `launchctl` invocation during the suite names a label matching `com.rjwalters.*`.

## Acceptance Criteria

- [ ] Running `bash defaults/scripts/tests/test-loom-daemon-update.sh` on a host with a live `com.rjwalters.loom-daemon` LaunchAgent leaves that job **loaded and running** — verified by `launchctl list | grep loom` before and after.
- [ ] The suite fails loudly if any `launchctl` call it makes targets a label outside its scratch namespace.
- [ ] `defaults/scripts/tests/test-loom-daemon-stop.sh` and `test-loom-daemon-start.sh` are audited for the same escape and fixed if present (they invoke the same scripts; not yet verified).
- [ ] Existing coverage is preserved — the restart path must still exercise the real stop/start scripts, not a mock, per the current header's intent.

## Notes

Worth considering as a general rule rather than a one-off patch: **any test that executes the real lifecycle scripts must run under a scratch `LOOM_LAUNCHD_LABEL`.** #4054 established exactly this discipline for the manual procedure and its PR honored it; this issue is the same constraint applied to the automated suite. A shared test helper that exports the scratch label (and a stub `launchctl`) would make the guarantee structural instead of per-file.

rjwalters:
## Summary

`defaults/scripts/tests/test-loom-daemon-update.sh` and `test-loom-daemon-stop.sh` could adopt and SIGTERM the operator's production `com.rjwalters.loom-daemon` LaunchAgent when run on a host with a live daemon — silently ending autonomy (5 firings on 2026-07-27, per #4078). This closes both escape hatches at the production layer and makes the guarantee structural via a shared test sandbox.

Per the Curator's verification, scoping `LOOM_LAUNCHD_LABEL` alone does **not** close this — the label-blind `pgrep` fallback re-opens it. Both tiers are addressed here.

## Changes

- **`loom-daemon-stop.sh` — honor `LOOM_DAEMON_LAUNCHD` symmetrically with `loom-daemon-start.sh:382`.** `start.sh` gated its whole launchd path on this var; `stop.sh` never referenced it, so the `LOOM_DAEMON_LAUNCHD=0` guard already present at `test-loom-daemon-update.sh:32` was **inert** on the stop side (Curator Correction 1). A `USE_LAUNCHD` gate now disables all launchd lookup + bootout when set to `0/false/no`.
- **`loom-daemon-stop.sh` — narrow the label-blind `pgrep` fallback (Curator Correction 3).** The last-resort `pgrep -f '(^|/)loom-daemon$'` matched any loom-daemon by binary name — including the production daemon and other repos' daemons. It is now skipped whenever the caller scoped a **non-default** `LOOM_LAUNCHD_LABEL`, so a scratch-labeled stop can never fall through to a by-name kill. The default-label lost-PID-file recovery path is preserved. This also closes the cross-repo variant (repo A's stop killing repo B's daemon), aligning with the #4054 label-scoped-stop discipline.
- **New `defaults/scripts/tests/lib/launchd-sandbox.sh`** — shared helper: per-run scratch `LOOM_LAUNCHD_LABEL`, recording stub `launchctl`/`pgrep` (syscall layer only — never the lifecycle scripts), a decoy `loom-daemon` spawner matchable by the fallback pattern, and a `no com.rjwalters.* label` assertion. Sourced from both affected suites.
- **`test-loom-daemon-stop.sh` / `test-loom-daemon-update.sh`** — source the sandbox, run under a scratch label, and add a decoy-process test plus a suite-level decoy-survival guard and a no-production-label assertion.

## Scope decision: pgrep fallback was **narrowed**, not left with a follow-up

I narrowed the production `pgrep` fallback in this PR (the Curator flagged it as builder's judgment). Rationale: with only a scratch label + launchctl stub, the label-blind `pgrep` tier stays reachable and is the exact hole that makes "label-only" fixes insufficient; narrowing it is small, safe (default-label recovery preserved), and fixes the real cross-repo production bug rather than only the test symptom. No follow-up needed.

- [x] With a live `com.rjwalters.loom-daemon` loaded, both suites leave it **loaded and running** — verified on the host: PID `96910` identical before/after, still loaded in `launchctl list`.
- [x] The `pgrep` escape is closed, not just `launchctl` — decoy-process test: a live process whose argv ends in `/loom-daemon` (confirmed matchable by the fallback pattern) survives a scratch-label stop.
- [x] `loom-daemon-stop.sh` honors `LOOM_DAEMON_LAUNCHD=0` — symmetry test asserts zero `launchctl` calls; the update test's `:32` guard is now load-bearing.
- [x] Suite fails loudly if any `launchctl` invocation names a `com.rjwalters.*` label — assertion added to both suites.
- [x] Existing coverage preserved — stubs are at the `launchctl`/`pgrep` layer only; the restart path still exercises the real `loom-daemon-start.sh` / `loom-daemon-stop.sh`.
- [x] pgrep fallback disposition stated (narrowed — see above).

## Test Plan

- [x] `test-loom-daemon-stop.sh` — 17/17 pass (incl. decoy + symmetry + no-production-label).
- [x] `test-loom-daemon-update.sh` — 33/33 pass (incl. suite decoy + no-production-label).
- [x] `test-loom-daemon-start.sh` — 10/10 pass (unmodified in intent).
- [x] `test-loom-daemon-launchd-plist.sh` — 26/26 pass (unmodified in intent).
- [x] Manual regression: production `com.rjwalters.loom-daemon` PID `96910` loaded+alive both before and after the full run.
- [x] `shellcheck -S warning` clean on all four changed/new files.
- [x] Linux/no-`launchctl` path unaffected (launchd off there regardless; `command -v pgrep` guard retained).

Closes #4078

### DIFF
--- a/defaults/scripts/cli/loom-daemon-stop.sh
+++ b/defaults/scripts/cli/loom-daemon-stop.sh
@@ -48,6 +48,11 @@
 # Environment:
 #   LOOM_DAEMON_STOP_GRACE_SECS   Grace window before SIGKILL (default 10)
 #   LOOM_LAUNCHD_LABEL            macOS only: the LaunchAgent label to bootout (default com.rjwalters.loom-daemon)
+#   LOOM_DAEMON_LAUNCHD           macOS only: 0/false/no disables ALL launchd interaction
+#                                 (lookup + bootout), symmetric with loom-daemon-start.sh.
+#                                 A start done with --no-launchd / LOOM_DAEMON_LAUNCHD=0
+#                                 must get a stop that never reads or mutates the
+#                                 machine-global launchd domain (issue #4078).
 #
 # Exit codes:
 #   0  daemon stopped (or was not running)
@@ -107,11 +112,25 @@ GRACE_SECS="${LOOM_DAEMON_STOP_GRACE_SECS:-10}"
 # ---------- launchd plumbing (macOS, #3972) ----------
 IS_DARWIN=false
 [[ "$(uname -s)" == "Darwin" ]] && IS_DARWIN=true
-LAUNCHD_LABEL="${LOOM_LAUNCHD_LABEL:-com.rjwalters.loom-daemon}"
+
+# Honor LOOM_DAEMON_LAUNCHD symmetrically with loom-daemon-start.sh (#4078): a
+# daemon started with --no-launchd / LOOM_DAEMON_LAUNCHD=0 was never a launchd
+# job, so the stop side must NOT reach into the machine-global launchd domain to
+# look it up (which would resolve against — and then SIGTERM — the operator's
+# real production LaunchAgent under the same default label). Before this, the
+# start script gated its whole launchd path on this var but the stop script did
+# not, leaving the guard inert on the stop side.
+USE_LAUNCHD="$IS_DARWIN"
+if [[ "${LOOM_DAEMON_LAUNCHD:-}" =~ ^(0|false|no)$ ]]; then
+    USE_LAUNCHD=false
+fi
+
+DEFAULT_LAUNCHD_LABEL="com.rjwalters.loom-daemon"
+LAUNCHD_LABEL="${LOOM_LAUNCHD_LABEL:-$DEFAULT_LAUNCHD_LABEL}"
 LAUNCHD_SERVICE="gui/$(id -u)/${LAUNCHD_LABEL}"
 
 launchd_job_loaded() {
-    [[ "$IS_DARWIN" == "true" ]] || return 1
+    [[ "$USE_LAUNCHD" == "true" ]] || return 1
     command -v launchctl >/dev/null 2>&1 || return 1
     launchctl print "$LAUNCHD_SERVICE" >/dev/null 2>&1
 }
@@ -144,7 +163,19 @@ if [[ -z "$pid" ]] || ! kill -0 "$pid" 2>/dev/null; then
     fi
 fi
 if [[ -z "$pid" ]] || ! kill -0 "$pid" 2>/dev/null; then
-    if command -v pgrep >/dev/null 2>&1; then
+    # Last-resort process-name match. This tier is LABEL-BLIND: `pgrep -f`
+    # matches ANY loom-daemon on the machine by binary name — including the
+    # operator's production daemon or another repo's daemon — so it can kill a
+    # daemon this invocation was never meant to touch (issue #4078, curator
+    # Correction 3; the incident's actual over-broad kill). Only fall through to
+    # it when the caller did NOT explicitly scope this stop to a specific
+    # launchd label: a non-default LOOM_LAUNCHD_LABEL (a test's scratch label,
+    # or an operator managing a specifically-labeled daemon) means "stop THAT
+    # daemon, not whatever else happens to be named loom-daemon", so a
+    # label-blind kill would violate that scoping and contradict the #4054
+    # label-scoped-stop discipline. With the default label we keep the fallback
+    # as a genuine lost-PID-file recovery path.
+    if [[ "$LAUNCHD_LABEL" == "$DEFAULT_LAUNCHD_LABEL" ]] && command -v pgrep >/dev/null 2>&1; then
         pid=$(pgrep -f '(^|/)loom-daemon$' 2>/dev/null | head -n1 || true)
     fi
 fi
--- a/defaults/scripts/tests/lib/launchd-sandbox.sh
+++ b/defaults/scripts/tests/lib/launchd-sandbox.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+# launchd-sandbox.sh — shared test sandbox for the loom-daemon lifecycle tests
+# (issue #4078).
+#
+# Why this exists: the daemon lifecycle tests deliberately execute the REAL
+# loom-daemon-start.sh / loom-daemon-stop.sh (not a mock) so the restart path
+# exercises the actual scripts. But launchd is machine-GLOBAL, and the stop
+# script's PID-resolution ladder can reach outside any per-test sandbox two
+# ways:
+#
+#   1. launchd label lookup — `launchctl print gui/<uid>/com.rjwalters.loom-daemon`
+#      resolves against the operator's REAL production LaunchAgent when the
+#      test does not scope LOOM_LAUNCHD_LABEL, so the stop script adopts and
+#      SIGTERMs the live daemon (issue #4078, curator Correction 2).
+#   2. label-blind pgrep — `pgrep -f '(^|/)loom-daemon$'` matches the production
+#      daemon (or another repo's daemon) by binary NAME alone, so even a scoped
+#      LOOM_LAUNCHD_LABEL does not close the hole (curator Correction 3).
+#
+# This helper closes BOTH axes for a sourcing test, and is deliberately narrow:
+# it stubs only the syscall-level tools (`launchctl`, `pgrep`), NEVER the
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Thorough review of this security-sensitive reliability fix (#4078). I ran all four suites **live on a host with the production `com.rjwalters.loom-daemon` loaded (PID 96910)** and confirmed it survived every run — the exact regression this PR closes.

## Verification (ran live, safely)

Static analysis first confirmed every daemon-touching invocation is scratch-label-scoped, so no path can reach the default-label production daemon; then I ran with a before/after PID guard:

| Suite | Result | Prod daemon |
|-------|--------|-------------|
| `test-loom-daemon-stop.sh` | 17/17 pass | PID 96910 unchanged |
| `test-loom-daemon-update.sh` | 33/33 pass | PID 96910 unchanged |
| `test-loom-daemon-start.sh` | 10/10 pass | unchanged |
| `test-loom-daemon-launchd-plist.sh` | 26/26 pass | unchanged |

CI: Shellcheck pass, Installer Integration Tests pass (Rust jobs correctly skipped — no Rust changes). Merge state CLEAN.

## Correctness of the three parts

1. **`LOOM_DAEMON_LAUNCHD` symmetry** — genuinely symmetric, not superficial: `stop.sh` uses the identical var name and identical regex `^(0|false|no)$` as `start.sh:382`, resolving `USE_LAUNCHD` with the same Darwin-default-on semantics. `launchd_job_loaded` now gates on `USE_LAUNCHD` instead of `IS_DARWIN`, so `=0` disables the whole launchd lookup + bootout. Test 8 proves zero `launchctl` calls under `LOOM_DAEMON_LAUNCHD=0` via the recording stub.

2. **pgrep narrowing** — `LAUNCHD_LABEL == DEFAULT_LAUNCHD_LABEL` correctly distinguishes "no override / default" (fallback preserved as lost-PID-file recovery) from "scratch label set" (fallback skipped). The `LOOM_LAUNCHD_LABEL=<default value explicitly>` edge case falls through to the fallback, which is correct — an operator naming the default label wants the default (production) daemon, and the test sandbox always uses a `com.example.*` scratch label so it is never affected. No regression for a normal default-settings operator: default-label stop still resolves via launchd then pgrep exactly as before.

3. **Shared sandbox** — stubs `launchctl`/`pgrep` at the syscall layer only (never the lifecycle scripts), preserving the update suite's "exercise the real scripts" intent per the Champion's scope guard. Stub is installed at `FAKE_BIN_DIR`, the first PATH entry, so `command -v` in the stop script resolves it ahead of the real tools.

## The decoy test genuinely proves the fix

Stop-suite test 7 is the strongest evidence: it runs the **real stop script against the real `pgrep` on PATH** (not the stub) with a live decoy whose argv ends in `/loom-daemon`, plus a non-vacuous pre-check that the decoy *is* matchable by `(^|/)loom-daemon$`. Pre-fix this would `pgrep`-match and SIGTERM the decoy; post-fix the scratch label routes around the pgrep tier and the decoy survives. This is behavioral proof, not just a stub-log assertion — exactly distinguishing a real fix from an insufficient label-only-on-launchctl one. Defense is layered (scratch label + label-narrowing + recording stubs + `LOOM_DAEMON_LAUNCHD=0`), and the scratch label alone means even the *real* `launchctl print` can't resolve the production job.

The pgrep-fallback scope-expansion (narrowing rather than filing a follow-up) is well-justified in the PR body and aligns with #4054 label-scoped-stop discipline. Excellent work.

Issue #4060, opened 2026-07-27
[#4047] Migrate 4 Python config readers onto config_resolver (paths, token select, model_tiers, sweep_experiment)


rjwalters:
**Epic**: #3835 — machine-level Loom install
**Parent**: #4047
**Depends on**: #4039 / PR #4051 (config resolution layer — merged as `789669e9`)

## Problem Statement

`loom_tools.common.config_resolver` shipped in #4039 with zero call sites migrated. Four Python readers still open `.loom/config.json` directly. This issue covers those four; the `forge`-config reader is split out separately because it carries a CWD-vs-repo-root question that would otherwise stall this work.

## Verified call sites (audited against `origin/main` @ `2ad8c176`)

| Site | Function | Key(s) | Notes |
|---|---|---|---|
| `common/paths.py:101` | `_read_config_worktree_root` | `worktree.root` | mirror of the Rust site; `LOOM_WORKTREE_ROOT` wins |
| `tokens/select.py:215` | `_read_config_spread_top_n` | `tokens.spreadTopN` | explicitly rejects `bool` (bool is an `int` subclass in Python — keep that guard) |
| `model_tiers.py:94` | `load_config` | whole file; `_config_overrides` reads `sweep.modelAliases` | default path is the **literal relative** `".loom/config.json"` (CWD-relative) |
| `sweep_experiment.py:87` | `load_config` | whole file; `resolve_raw_mode` reads `sweep.modelExperiment` | same shape |

## The decision that gates this issue: the `--config <path>` bypass

`model_tiers.py` and `sweep_experiment.py` both accept an explicit `--config <path>` pointing at a **single file**, but `resolve_effective_config(repo_root)` only takes a repo root and always merges four tiers. There is no single-file mode. Pick one and state it in the PR:

- **Preferred**: keep `--config <path>` as an explicit bypass that reads exactly that file and skips tiering entirely (it is a debugging/testing affordance, and silently merging tiers underneath an explicitly-named file would be surprising). Route only the *default* path through the resolver.
- Alternative: extend the resolver with an override-file parameter. Heavier; only do this if a caller genuinely needs tiering *plus* an override.

Also note: `model_tiers.py`'s default path is relative to **CWD**, not the repo root. `resolve_effective_config` needs a root — derive it the way the rest of `loom_tools` does rather than passing `Path(".")`, and add a test proving resolution works when invoked from a worktree subdirectory.

## Semantics that MUST be preserved

[...]

Issue #4058, opened 2026-07-27
[#4047] Migrate 10 mechanical Rust config readers onto config_resolver (soft-fail cohort)


rjwalters:
**Epic**: #3835 — machine-level Loom install
**Parent**: #4047 (Phase 2 follow-up: migrate `.loom/config.json` call sites onto the config resolver)
**Depends on**: #4039 / PR #4051 (config resolution layer — merged as `789669e9`)

## Problem Statement

#4039 shipped `loom-daemon/src/config_resolver.rs` but migrated zero call sites. Ten Rust readers still open `.loom/config.json` directly with a hand-rolled `read_to_string` → `serde_json::from_str` → `.get("a")?.get("b")?` chain. Until they move onto the resolver, `.loom-project/project.json` and `.loom-local/local.json` are inert for every daemon knob — the tiering exists but nothing consumes it.

This issue covers the **mechanical cohort only**: the ten sites whose semantics are "soft-fail to `None`, caller applies env > config > default". Each is a drop-in swap of the path-construction + read + key-walk for `resolve_effective_config(repo_root)` + `get_path(&cfg, "dotted.key")`. Structural readers (`terminals[]` iteration, whole-object reads, the one hard-failing site) are **out of scope** — see the sibling issue.

## Verified call sites (audited against `origin/main` @ `2ad8c176`)

| Site | Function | Keys |
|---|---|---|
| `worktree_root.rs:93` | `read_config_worktree_root` | `worktree.root` |
| `main_health_gate.rs:526` | `read_build_gate_config` | `buildGate.{enabled,command,timeoutSeconds,realChangeGlobs}` |
| `main_health_gate.rs:2126` | `read_autonomous_gate_config` | `autonomous.mainHealthGate.{enabled,ciWorkflow}` |
| `work_finder.rs:732` | `read_work_finder_config` | `autonomous.workFinder.{enabled,intervalSecs,maxConcurrent}` + `autonomous.perTokenConcurrency` |
| `role_runner.rs:511` | `read_role_runner_config` | `autonomous.roleRunner.{enabled,roles,intervalSecs}` |
| `token_ranking_refresh.rs:343` | `read_token_ranking_refresh_config` | `autonomous.tokenRankingRefresh.{enabled,intervalSecs}` |
| `watch_registry.rs:733` | `read_watch_monitor_config` | `autonomous.watchMonitor.{enabled,intervalSecs,expirySecs}` |
[...]

Issue #4047, opened 2026-07-27
[Epic #3835] Phase 2 follow-up: migrate ~40 .loom/config.json call sites onto the config resolver


rjwalters:
> **Curator note (2026-07-27)**: this issue was audited against `origin/main` @ `2ad8c176` and **decomposed into seven sub-issues**. It is now a **tracking issue** — do not implement it directly. See "Decomposed sub-issues" at the bottom. The original body is preserved unchanged below.
>
> **Curator re-verification (2026-07-28, `origin/main` @ `0a7c1581`)**: five of seven children have resolved — #4058, #4059, #4062 merged; **#4060 and #4061 were folded into epic #4081 and closed as `not planned`**, not implemented. That fold invalidated three claims in the text below, which have been corrected in place: original finding **#8** (`resolve-model.sh` is *not* covered any more), the **cross-cutting constraints** list, and **parent-level acceptance criteria** 2 and 4 (they were unsatisfiable as written). ~~Only #4063 and #4064 remain open — both `loom:building`.~~ **Updated 2026-07-28 (later pass): #4064 is merged (PR #4153); only #4063 remains open, and it is parked `loom:blocked` on #4178 — see the Dependencies section.**

---

**Epic**: #3835 - Epic: machine-level Loom install — stop copying the implementation into every consumer repo
**Depends on**: #4039 (Phase 2: config resolution layer — resolver only, merged)

## Description

#4039 shipped one config resolver per language (Rust, Python, Bash) implementing the tier-precedence scheme: private defaults + tracked `.loom-project/project.json` + ignored `.loom-local/local.json` + legacy `.loom/config.json`, with `.loom/config.json` folded in as a legacy tier so every existing repo (which has only that file) resolves identically to today. That issue was explicitly scoped as **the resolver only** — it does not migrate any existing call site.

This follow-up migrates the ~40 existing ad hoc `.loom/config.json` readers across the three languages onto the new resolver, one call site (or cohesive group) at a time, so each real reader's behavior is verified against the resolver before the next is touched.

## Known call sites (non-exhaustive — audit at pickup time for drift)

[...]

Issue #4039, opened 2026-07-27
[Epic #3835] Phase 2: config resolution layer — single resolver over private defaults + .loom-project + .loom-local


rjwalters:
**Epic**: #3835 - Epic: machine-level Loom install — stop copying the implementation into every consumer repo
**Phase**: 2 of 6
**Phase Goal**: `.loom-project/` schema + config resolution (private defaults + project config + local state merge)

## Description

Epic #3835's target architecture splits today's single `.loom/config.json` into three tiers: private/shared defaults in the machine-level checkout, tracked project config in `.loom-project/project.json`, and ignored host-local state in `.loom-local/`. Effective config = **private defaults + tracked `.loom-project` + ignored `.loom-local`**.

Nothing can migrate to that layout until there is something that *resolves* it, and today there is no such thing. `.loom/config.json` is read ad hoc at ~40 call sites across three languages, each with its own bespoke path derivation and soft-fail behavior:

- **Rust** — read directly per module: `loom-daemon/src/lib.rs:89` (`workspace.join(".loom").join("config.json")`), plus `work_finder.rs`, `role_runner.rs`, `main_health_gate.rs`, `worktree_root.rs`, `workspace_registry.rs`, `token_ranking_refresh.rs`, `sweep_registry.rs`, and others.
- **Python** — `loom_tools/common/paths.py` centralizes *paths* (`LoomPaths.config_file`, `paths.py:283`) but not *values*; each key gets its own reader, e.g. `_read_config_worktree_root` (`paths.py:101`) with hand-rolled soft-fail.
- **Bash** — `defaults/scripts/lib/forge-helpers.sh:61-65` and `:138-142` re-derive the config path from `$REPO_ROOT` / `$WORKSPACE_ROOT` / cwd inline, duplicated per function.

Migrating ~40 sites and introducing a new file layout in one change is not atomically reviewable. **This issue is the resolver only.** It introduces the schema and one resolver per language, with `.loom/config.json` as a legacy tier in the chain so that behavior is byte-for-byte unchanged for every existing repo. A follow-up issue migrates call sites onto the resolver; a later one flips the default layout.

This phase is also the blocking prerequisite for the elastic-compute work: #3979's Phase 2 (scripted cloud-host scale-out) was decided as Option A (federated daemons) and cannot produce a repeatable provisioning runbook while each host's config is a hand-copied `.loom/` tree.

## Scope

**In scope**
[...]

PR #4093, opened 2026-07-27
feat(daemon): migrate 3 structural config readers (terminals[], modelAliases, validate) onto config_resolver (#4059)

### LINKED ISSUE #4059: [#4047] Migrate 3 structural Rust config readers: terminals[], modelAliases, and the hard-failing validate path
**Epic**: #3835 — machine-level Loom install
**Parent**: #4047
**Depends on**: #4039 / PR #4051 (config resolution layer — merged as `789669e9`)

## Problem Statement

Three Rust `.loom/config.json` readers cannot be migrated by the mechanical `get_path` swap that covers the other ten. Each needs a real decision, so they are isolated here rather than bundled into the mechanical cohort where a stalled decision would block seven files of otherwise-trivial work.

### 1. `lib.rs:89` `extract_configured_terminal_ids` — array iteration
Walks `terminals[]` collecting `.id` into a `HashSet`, and returns `None` (not `Some(empty)`) when the resulting set is empty. `get_path` returns the array fine; the walk and the empty-set-⇒-`None` rule must be reimplemented on top of it, not inferred.

### 2. `sweep_registry.rs:209` `read_model_aliases` — whole-object read with a cross-language invariant
Reads the entire `sweep.modelAliases` object, lowercases/trims keys, drops blank entries, and overlays the result on `DEFAULT_TIER_ALIASES` (currently `opus → claude-opus-5`). **This logic is duplicated in Python** (`model_tiers.py::_config_overrides`) and the two must stay byte-identical — a divergence means `resolve-model.sh` and the daemon disagree on which model a tier resolves to. Note the resolver's `deep_merge` replaces arrays but merges objects, so a `sweep.modelAliases` split across the legacy and project tiers now *merges* where previously only one file could supply it. Decide whether that is desirable and test it explicitly.

### 3. `main.rs:2742` + `role_validation.rs:225` — the only hard-failing site
`handle_validate_command` builds the path and `validate_from_file(config_path, mode)` does its own read + typed `serde_json::from_str::<Config>`. A missing file prints an error and `std::process::exit(1)`.

**This is the design decision that gates the issue.** Under tiering, "`.loom/config.json` is absent" no longer implies "there is no config" — a repo could legitimately configure terminals entirely from `.loom-project/project.json`. Keeping the current check would make a validly-configured repo fail validation. Resolve it explicitly:

- **Preferred**: change the precondition from *"the legacy file exists"* to *"the resolved effective config contains a non-empty `terminals` array"*, and change `validate_from_file`'s signature from `&Path` to an already-resolved `&Value` (or add a `validate_from_config`). Keep the exit-1 behavior, retarget the condition.
- If the signature change ripples further than expected, say so in the PR and propose the narrower alternative rather than silently keeping the file-existence check.

## Acceptance Criteria

- [ ] All three sites read through `config_resolver::resolve_effective_config`; no direct `.loom/config.json` path construction remains anywhere in `loom-daemon/src/` outside `config_resolver.rs` and `init/`.
- [ ] `extract_configured_terminal_ids` preserves the empty-set-⇒-`None` contract (test it explicitly — it is easy to lose in the rewrite).
- [ ] `read_model_aliases` output is unchanged for every existing input, and its behavior is documented as intentionally mirroring `model_tiers.py`. Add a test asserting the `DEFAULT_TIER_ALIASES` overlay still wins/loses in the same direction as today.
- [ ] The validate-command precondition is retargeted (per the decision above) and a repo whose terminals come **only** from `.loom-project/project.json` validates successfully.
- [ ] A repo with **no** config in any tier still exits 1 with a clear message naming the tiers it looked in.
- [ ] `cargo clippy --all-targets -- -D warnings` clean.

## Affected Files

> Line numbers below were **re-verified against `origin/main` @ `e5d5d491`** (2026-07-27). The
> original enumeration was taken at `2ad8c176`; `main.rs` has since drifted **+286 lines** via
> #4066 / #4067 / #4068. Corrected refs:

- `loom-daemon/src/lib.rs` — `extract_configured_terminal_ids` (**L89** — unchanged)
- `loom-daemon/src/sweep_registry.rs` — `read_model_aliases` (**L209**, was cited as L210); `DEFAULT_TIER_ALIASES` at L202; consumer `resolve_model_alias` at L243
- `loom-daemon/src/main.rs` — `handle_validate_command` (**L2727**, was cited as L2441), path construction at **L2742** (was L2456), `validate_from_file` call at **L2761**, `Validate` subcommand declaration at L183
- `loom-daemon/src/role_validation.rs` — `validate_from_file` (**L225** — unchanged), `extract_roles_from_config` (**L142** — unchanged); signature change

## Out of scope

- The ten mechanical Rust readers — **sibling issue #4058** (already `loom:curated`)
- Python/Bash/TypeScript sites (sibling issues)
- Actually creating `.loom-project/project.json` in this repo (Epic #3835 Phase 6)

## Test Plan

- [ ] Automated: `cd loom-daemon && cargo test`
- [ ] Automated: new test — terminals defined only in `.loom-project/project.json` are discovered by `extract_configured_terminal_ids` and accepted by validate
- [ ] Automated: new test — empty `terminals` array still yields `None` from `extract_configured_terminal_ids`
- [ ] Automated: new test — `sweep.modelAliases` merge across legacy + project tiers behaves as decided, with the chosen behavior asserted (not left implicit)
- [ ] Manual: `loom-daemon validate . --verbose` and `loom-daemon validate . --format json` on this repo before/after — identical output (it is a **subcommand**, not a `--validate` flag; see Curator Verification)
- [ ] Manual: `loom-daemon validate /tmp/empty-dir` exits 1 with a message naming every tier searched
- [ ] Manual: `loom-daemon --status` unchanged with no `.loom-project/` present
- [ ] `cargo clippy --all-targets -- -D warnings` clean
- [ ] Cross-language: confirm `./.loom/scripts/resolve-model.sh opus` and the daemon's dispatch model agree after the change

---

## Curator Verification (2026-07-27, vs `origin/main` @ `e5d5d491`)

Every claim in the body above was checked against the tree. **All three sites and their
described behaviors are accurate.** Line numbers were stale and are corrected in Affected
Files. Five additional findings below change what the Builder must do.

### Confirmed as described

| Claim | Verdict |
|---|---|
| `extract_configured_terminal_ids` returns `None` (not `Some(empty)`) on an empty set | ✅ `lib.rs:114-117`, and an existing test `test_extract_terminal_ids_empty_terminals` (`lib.rs:228`) already asserts it |
| `read_model_aliases` lowercases/trims keys, drops blanks | ✅ `sweep_registry.rs:227-237` |
| `DEFAULT_TIER_ALIASES` is `opus → claude-opus-5` | ✅ `sweep_registry.rs:202`; mirrored in `model_tiers.py:80` |
| The validate path hard-fails via `std::process::exit(1)` | ✅ `main.rs:2744-2753` |
| `deep_merge` merges objects, replaces arrays/scalars | ✅ `config_resolver.rs:109-124` |
| No caller uses `config_resolver` yet | ✅ `grep 'config_resolver::' loom-daemon/src` outside the module itself returns nothing — this issue and #4058 are its first consumers |

### Finding 1 — the overlay direction is **config-wins**; state it, don't rediscover it

AC says "assert the `DEFAULT_TIER_ALIASES` overlay still wins/loses in the same direction as
today." Made concrete so the Builder does not have to infer it:

- **Rust** (`resolve_model_alias`, `sweep_registry.rs:257-266`): `overrides.get(&key).or_else(|| DEFAULT_TIER_ALIASES…)` → **per-repo `sweep.modelAliases` wins; the shipped default is the fallback.**
- **Python** (`model_tiers.py:127`): `{**_DEFAULT_TIER_ALIASES, **_config_overrides(config)}` → **same direction.**

The new test should assert exactly that: a repo-supplied `opus` override beats
`claude-opus-5`, and an absent override falls back to it. (`sweep_registry.rs:4741` already
covers the override case with `claude-opus-6` — extend rather than duplicate.)

### Finding 2 — Rust and Python are **not** byte-identical today: blank keys diverge

The AC asks for parity with `model_tiers.py::_config_overrides`. There is a **pre-existing**
divergence the Builder will trip over while asserting it:

| Input `sweep.modelAliases` | Rust `read_model_aliases` | Python `_config_overrides` |
|---|---|---|
| `{"  ": "x"}` (blank key) | `{}` — dropped (`key.is_empty()` guard, L230) | `{"": "x"}` — **kept**; Python only guards `val.strip()`, never the key |
| `{"opus": "  "}` (blank value) | `{}` — dropped | `{}` — dropped |
| `{"  OPUS  ": " m "}` | `{"opus": "m"}` | `{"opus": "m"}` |

Practical impact is **nil** — a blank key can never match a lookup, and `resolve_model_alias`
returns early on an empty input model. But do not "fix" it silently in this PR: the Python
side belongs to **#4060**. Either (a) leave both as-is and add a comment naming the blank-key
divergence as known-and-harmless, or (b) coordinate the guard into #4060. Do not change Rust
to match Python's laxer behavior.

### Finding 3 — migrating `extract_configured_terminal_ids` **loses a diagnostic**

Today the function returns `None` for four distinguishable reasons, two of which log:

1. file unreadable → `log::debug!` (L94)
2. **JSON malformed → `log::warn!` (L102)**
3. `terminals` absent or not an array → silent (L108, via `?`)
4. resulting set empty → `log::debug!` (L115)

`resolve_effective_config` soft-fails cases 1 and 2 to `{}` internally, so after migration a
**malformed `.loom/config.json` collapses into case 3 and the `warn` disappears** — a repo with
a typo'd config silently reports "no terminals" instead of warning. The observable `None`
contract is preserved, so this is not an AC violation, but it is a real diagnosability
regression. Make it an explicit decision in the PR: either accept it (and say so) or surface
malformed-tier detection from the resolver. The existing test at `lib.rs:210`
(`"not valid json"`) must be **kept** — it will now exercise a different path and is the
guard against silently dropping the case entirely.

### Finding 4 — the `validate_from_file` signature change is cheap: **one production caller**

The body hedges that the signature change "may ripple further than expected". It does not:

- `validate_from_file` has exactly **one** non-test caller — `main.rs:2761`.
- `validate_role_completeness(&str, mode)` and `extract_roles_from_config(&str)` both take a
  **JSON string**, while the resolver yields a `serde_json::Value`.

So the preferred option in the body is the right one and is low-risk. Minimal-ripple shape:
add `validate_from_config(config: &serde_json::Value, mode) -> ValidationResult` alongside the
existing string-based entry points (serializing internally, or better: change
`extract_roles_from_config` to accept `&Value` and keep a thin `&str` wrapper for its tests).
Do **not** delete `validate_from_file` if its own tests exercise it — deprecate or keep as a wrapper.

### Finding 5 — the missing-config error exists in **two** branches

AC says "exits 1 with a clear message naming the tiers it looked in". Both output paths must be
updated, not just the human one (`main.rs:2745-2752`):

- `format == "json"` → `{"error": "Config file not found: <path>"}`
- text → `eprintln!` + the hint `loom-daemon validate /path/to/workspace`

Both name a single file that is no longer the sole source of truth.

### Finding 6 — merge-conflict risk with #4058 in `sweep_registry.rs`

**#4058 also edits `sweep_registry.rs`** (its Affected Files names L151 / L3629 / L3773) while
this issue edits `read_model_aliases` at L209. Different functions, so no logical conflict —
but a textual rebase is likely if both are in flight. The parent (#4047) recommends landing
#4058 first; do that, or expect to rebase. Coordinate rather than racing.

### Coordination constraints (deliberately **not** blockers)

Per #4047 these are cross-issue agreements, not prerequisites — this issue can be built and
merged independently, so no `Dependencies` task list is added and `loom:blocked` is **not** applied:

- **#4062** (Bash `defaults/scripts/**`) must reach the *same* answer on the hard-fail
  precondition — `cli/loom-start.sh` and `validate-roles.sh` have the identical
  "file exists ⇒ config exists" assumption. Whichever lands first sets the precedent; state
  the chosen rule explicitly in the PR body so #4062 can mirror it verbatim.
- **#4060** (Python) owns `model_tiers.py`; see Finding 2.

### Corrections to the Test Plan

- **There is no `loom-daemon --validate` flag.** It is a subcommand:
  `loom-daemon validate [WORKSPACE] [--format text|json] [--strict] [--verbose]`
  (declared at `main.rs:183`, `WORKSPACE` defaults to `.`). Concrete before/after check:
  ```bash
  cd loom-daemon && cargo build --release
  ./target/release/loom-daemon validate . --verbose            # text branch
  ./target/release/loom-daemon validate . --format json        # json branch
  ./target/release/loom-daemon validate /tmp/empty-dir         # must exit 1, message names all tiers
  ```
- **Neither `.loom-project/` nor `.loom-local/` exists anywhere on `origin/main`.** Every
  tiering test must construct its own `tempdir` fixture; there is nothing to copy from.
- `defaults/scripts/resolve-model.sh` is present and is a thin delegator to Python
  `model_tiers`, so the cross-language check is:
  ```bash
  ./.loom/scripts/resolve-model.sh opus     # expect: claude-opus-5
  ```
  and it must still agree with `resolve_model_alias` after the change.
- Add: **`cargo clippy --all-targets -- -D warnings`** is already an AC but is missing from the
  Test Plan — run it, it is the gate that catches the `Option`-semantics rewrites.

### Additional acceptance criteria

- [ ] The malformed-config diagnostic decision (Finding 3) is stated explicitly in the PR body,
      and `lib.rs`'s existing invalid-JSON test is retained.
- [ ] The blank-key Rust/Python divergence (Finding 2) is documented in a code comment rather
      than silently normalized in either direction.
- [ ] Both the `text` and `json` branches of the validate error path are updated (Finding 5).
- [ ] The chosen hard-fail precondition rule is stated verbatim in the PR body so #4062 can mirror it.

rjwalters:
Migrates the three structural `.loom/config.json` readers that each needed a real decision (not the mechanical `get_path` swap of sibling #4058) onto `config_resolver::resolve_effective_config`, so `terminals[]`, `sweep.modelAliases`, and the validate precondition all participate in the `.loom-project` / `.loom-local` tier chain (#4039).

Part of epic #3835 / parent #4047. Depends on #4039 (merged as `789669e9`).

## The three sites

1. **`lib.rs` `extract_configured_terminal_ids`** — reads `terminals[]` through the resolver. Empty-set-⇒-`None` contract preserved and tested across tiers.
2. **`sweep_registry.rs` `read_model_aliases`** — reads `sweep.modelAliases` through the resolver, overlaid on `DEFAULT_TIER_ALIASES` (`opus → claude-opus-5`).
3. **`main.rs` `handle_validate_command` + `role_validation.rs`** — the hard-failing site; precondition retargeted, `validate_from_config(&Value, mode)` added.

### Finding 3 — malformed-config diagnostic (accepted)
The previous direct read in `extract_configured_terminal_ids` emitted a function-local `log::warn!` for malformed JSON. That warn is **not lost**: `config_resolver::soft_read_json_object` still emits a `log::warn!` on a parse failure. A malformed `.loom/config.json` now soft-fails to `{}` at the resolver and collapses into the "no terminals" branch here rather than a dedicated parse-error branch. **Decision: accept this.** The observable `None` contract is unchanged, the operator still sees a warning (it just names the resolver tier rather than the call site), and the existing invalid-JSON test (`lib.rs`) is retained — it now exercises the collapsed path and guards against silently dropping the case.

### Finding 2 — Rust/Python blank-key divergence (left as a code comment, NOT normalized)
`read_model_aliases` drops blank keys (`key.is_empty()` guard); the Python mirror in `model_tiers._config_overrides` **keeps** blank keys (it only guards blank *values*). Both drop blank values. This pre-existing, known-and-harmless divergence (a blank key can never match a lookup, and `resolve_model_alias` returns early on an empty input model) is **documented in a code comment on `read_model_aliases`, not silently normalized in either direction.** The Python side is owned by #4060.

### Cross-tier merge for `sweep.modelAliases` (intentional)
`deep_merge` merges objects, so `sweep.modelAliases` split across tiers now **unions** its keys with the higher-precedence tier winning per-key (rather than one file being the sole source). This is deliberate and desirable — a `.loom-project` / `.loom-local` tier can repoint an individual alias while inheriting the rest — and is asserted explicitly by the new `read_model_aliases_merges_across_tiers` test.

This replaces the old "the legacy `.loom/config.json` file exists" check. Under tiering, an absent legacy file no longer implies "no config" — a repo may configure `terminals` entirely from `.loom-project/project.json`.

### Finding 5 — both error branches name the tiers
Both the `text` and `json` branches of the missing-config error now list every tier searched (private/shared defaults → legacy → project → local), in precedence order.

### Finding 4 — signature change is cheap (as predicted)
`validate_from_file` has one production caller; it is retained as a thin backward-compat wrapper. Added `validate_from_config(&serde_json::Value, mode)` + `extract_roles_from_value(&Value)` alongside the existing string-based entry points, with a shared `check_role_dependencies` helper.

## Note on `validate --verbose` output
The verbose header line changed from `  Config: <path>/.loom/config.json` to `  Workspace: <path>` — there is no longer a single config file to name. The substantive output (configured roles, warnings, json result) is byte-identical before/after.

## Testing

- `cargo test` (loom-daemon): all pass (the one flake observed under parallelism — `terminal::claude_config::test_setup_creates_fallback_state_when_no_home_state` — is a pre-existing HOME-env test-ordering race, unrelated to this change; passes in isolation and on rerun).
- New tests: terminals-only-in-project-tier discovery; empty terminals array still `None` across tiers; `modelAliases` merge-across-tiers; config-wins overlay direction (Finding 1); `validate_from_config` Value path; project-tier-only validate.
- Manual (`--release`): `validate . --verbose` / `--format json` substantive output unchanged; `validate /tmp/empty-dir` exits 1 naming all four tiers in both text and json branches.
- Cross-language: `./.loom/scripts/resolve-model.sh opus` → `claude-opus-5`, matching Rust `resolve_model_alias`.

## Scope
Only the three structural sites are migrated here. The ten mechanical Rust readers (`work_finder`, `main_health_gate`, `role_runner`, `token_ranking_refresh`, `watch_registry`, `worktree_root`, `ipc`, and the other `sweep_registry` functions) remain sibling #4058's work — expect a possible textual rebase, not a logical conflict (Finding 6).

Closes #4059

### DIFF
--- a/loom-daemon/src/lib.rs
+++ b/loom-daemon/src/lib.rs
@@ -82,28 +82,27 @@ pub fn rotate_log_file(log_path: &Path, max_size: u64, max_files: usize) -> anyh
     Ok(())
 }
 
-/// Extract configured terminal IDs from workspace config.json.
+/// Extract configured terminal IDs from workspace config.
 ///
-/// Reads the workspace's `.loom/config.json` and extracts the `id` field from
-/// each terminal entry. Returns None if config file doesn't exist or can't be parsed.
+/// Resolves the workspace's effective config through the tier chain
+/// (`config_resolver::resolve_effective_config`: private/shared defaults →
+/// legacy `.loom/config.json` → `.loom-project/project.json` →
+/// `.loom-local/local.json`) and extracts the `id` field from each terminal
+/// entry. Returns `None` when no tier supplies a non-empty `terminals` array —
+/// the empty-set-⇒-`None` contract is preserved (#4059).
+///
+/// **Diagnostic note (#4059, Finding 3):** the previous direct-read
+/// implementation emitted a function-local `log::warn!` for a malformed
+/// `.loom/config.json`. That warn is not lost: `config_resolver`'s
+/// `soft_read_json_object` still emits a `log::warn!` when a tier's JSON fails
+/// to parse. A malformed config therefore collapses to `{}` at the resolver
+/// layer (so it now reaches the "no terminals" branch here rather than a
+/// dedicated parse-error branch), but the operator still sees a warning — it
+/// just names the resolver tier rather than this call site. This diagnosability
+/// change is accepted deliberately; the observable `None` contract is unchanged
+/// and the existing invalid-JSON test guards the collapsed path.
 pub fn extract_configured_terminal_ids(workspace: &Path) -> Option<HashSet<String>> {
-    let config_path = workspace.join(".loom").join("config.json");
-
-    let config_str = match fs::read_to_string(&config_path) {
-        Ok(s) => s,
-        Err(e) => {
-            log::debug!("Could not read config at {}: {e}", config_path.display());
-            return None;
-        }
-    };
-
-    let config: serde_json::Value = match serde_json::from_str(&config_str) {
-        Ok(v) => v,
-        Err(e) => {
-            log::warn!("Could not parse config at {}: {e}", config_path.display());
-            return None;
-        }
-    };
+    let config = config_resolver::resolve_effective_config(workspace);
 
     let terminals = config.get("terminals")?.as_array()?;
 
@@ -113,11 +112,15 @@ pub fn extract_configured_terminal_ids(workspace: &Path) -> Option<HashSet<Strin
         .collect();
 
     if ids.is_empty() {
-        log::debug!("No terminal IDs found in config");
+        log::debug!("No terminal IDs found in resolved config for {}", workspace.display());
         return None;
     }
 
-    log::info!("Loaded {} configured terminal IDs from {}", ids.len(), config_path.display());
+    log::info!(
+        "Loaded {} configured terminal IDs from resolved config for {}",
+        ids.len(),
+        workspace.display()
+    );
 
     Some(ids)
 }
@@ -232,6 +235,49 @@ mod tests {
         fs::write(loom_dir.join("config.json"), r#"{"terminals": []}"#).unwrap();
 
         let result = extract_configured_terminal_ids(dir.path());
+        assert!(result.is_none());
+    }
+
+    /// #4059: terminals supplied ONLY by the `.loom-project/project.json`
+    /// tier (no legacy `.loom/config.json` at all) are discovered through the
+    /// resolver. The private/shared defaults tier is disabled for hermeticity.
+    #[test]
+    #[serial_test::serial]
+    fn test_extract_terminal_ids_from_project_tier_only() {
+        std::env::set_var(crate::config_resolver::PRIVATE_DEFAULTS_ENV, "");
+        let dir = tempdir().unwrap();
+        let project_dir = dir.path().join(".loom-project");
+        fs::create_dir_all(&project_dir).unwrap();
+        fs::write(
+            project_dir.join("project.json"),
+            r#"{"terminals": [{"id": "terminal-1"}, {"id": "terminal-2"}]}"#,
+        )
+        .unwrap();
+        // No .loom/config.json exists.
+        assert!(!dir.path().join(".loom").join("config.json").exists());
+
+        let result = extract_configured_terminal_ids(dir.path());
+        std::env::remove_var(crate::config_resolver::PRIVATE_DEFAULTS_ENV);
+
+        let ids = result.expect("terminals from project tier should be discovered");
+        assert_eq!(ids.len(), 2);
+        assert!(ids.contains("terminal-1"));
+        assert!(ids.contains("terminal-2"));
+    }
+
+    /// #4059: an empty `terminals` array in the project tier still yields
+    /// `None`, preserving the empty-set-⇒-`None` contract across tiers.
+    #[test]
+    #[serial_test::serial]
+    fn test_extract_terminal_ids_empty_terminals_project_tier() {
+        std::env::set_var(crate::config_resolver::PRIVATE_DEFAULTS_ENV, "");
+        let dir = tempdir().unwrap();
+        let project_dir = dir.path().join(".loom-project");
+        fs::create_dir_all(&project_dir).unwrap();
+        fs::write(project_dir.join("project.json"), r#"{"terminals": []}"#).unwrap();
+
+        let result = extract_configured_terminal_ids(dir.path());
+        std::env::remove_var(crate::config_resolver::PRIVATE_DEFAULTS_ENV);
         assert!(result.is_none());
     }
 
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -2819,7 +2819,8 @@ fn handle_validate_command(
     strict: bool,
     verbose: bool,
 ) -> Result<()> {
-    use role_validation::{format_validation_result, validate_from_file, ValidationMode};
+    use loom_daemon::config_resolver;
+    use role_validation::{format_validation_result, validate_from_config, ValidationMode};
 
     let workspace_path = std::path::Path::new(workspace);
     let absolute_workspace = if workspace_path.is_absolute() {
@@ -2828,13 +2829,62 @@ fn handle_validate_command(
         std::env::current_dir()?.join(workspace_path)
     };
 
-    let config_path = absolute_workspace.join(".loom").join("config.json");
+    // #4059: resolve the effective config through the full tier chain rather
+    // than reading the legacy `.loom/config.json` directly. Under tiering,
+    // "the legacy file is absent" no longer implies "there is no config" — a
+    // repo may configure `terminals` entirely from `.loom-project/project.json`
+    // or `.loom-local/local.json`.
+    let config = config_resolver::resolve_effective_config(&absolute_workspace);
 
-    if !config_path.exists() {
+    // The tiers searched, in precedence order — named in every "not found" error
+    // (Finding 5: both text and json branches).
+    let mut searched_tiers: Vec<String> = Vec::new();
+    if let Some(defaults) = config_resolver::private_defaults_path() {
+        searched_tiers.push(defaults.display().to_string());
+    }
+    searched_tiers.push(
+        absolute_workspace
+            .join(config_resolver::LEGACY_CONFIG_REL)
+            .display()
+            .to_string(),
+    );
+    searched_tiers.push(
+        absolute_workspace
+            .join(config_resolver::PROJECT_CONFIG_REL)
+            .display()
+            .to_string(),
+    );
+    searched_tiers.push(
+        absolute_workspace
+            .join(config_resolver::LOCAL_CONFIG_REL)
+            .display()
+            .to_string(),
+    );
+
+    // Retargeted precondition (#4059) — stated verbatim for #4062 to mirror:
+    //   "A workspace is validatable iff resolve_effective_config(workspace)
+    //    yields a `terminals` array with at least one element; otherwise the
+    //    command exits 1, naming every tier it searched."
+    let has_terminals = config
+        .get("terminals")
+        .and_then(serde_json::Value::as_array)
+        .is_some_and(|arr| !arr.is_empty());
+
+    if !has_terminals {
         if format == "json" {
-            println!(r#"{{"error": "Config file not found: {}"}}"#, config_path.display());
+            let err = serde_json::json!({
+                "error": "No Loom config with a non-empty terminals array found in any tier",
+                "searched": searched_tiers,
+            });
+            println!("{}", serde_json::to_string(&err)?);
         } else {
-            eprintln!("Error: Config file not found: {}", config_path.display());
+            eprintln!(
+                "Error: No Loom config with a non-empty `terminals` array found in any tier."
+            );
+            eprintln!("\nSearched (lowest to highest precedence):");
+            for tier in &searched_tiers {
+                eprintln!("  - {tier}");
+            }
             eprintln!("\nMake sure you're in a Loom workspace or specify the path:");
             eprintln!("  loom-daemon validate /path/to/workspace");
         }
@@ -2847,14 +2897,14 @@ fn handle_validate_command(
         ValidationMode::Warn
     };
 
-    let result = validate_from_file(&config_path, mode).map_err(|e| anyhow!("{e}"))?;
+    let result = validate_from_config(&config, mode);
 
     if format == "json" {
         println!("{}", serde_json::to_string_pretty(&result)?);
     } else {
         if verbose {
             println!("\nValidating role configuration...");
-            println!("  Config: {}", config_path.display());
+            println!("  Workspace: {}", absolute_workspace.display());
             println!();
         }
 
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge Review — APPROVED

Reviewed against every acceptance criterion and required decision in #4059 (including the Curator Verification section). All three structural sites are correctly migrated onto `config_resolver::resolve_effective_config`, and every required decision is stated explicitly.

### Site-by-site verification

**1. `lib.rs::extract_configured_terminal_ids`** — reads through the resolver; the empty-set-⇒-`None` contract is preserved and the `terminals` array walk is reimplemented on top of `get("terminals")?.as_array()?`. The existing `test_extract_terminal_ids_invalid_json` ("not valid json", lib.rs:209) is retained and now guards the collapsed path. New tests added: `test_extract_terminal_ids_from_project_tier_only` (project-tier-only discovery) and `test_extract_terminal_ids_empty_terminals_project_tier` (empty-across-tiers ⇒ `None`). Both gate the private-defaults tier off for hermeticity.

**2. `sweep_registry.rs::read_model_aliases`** — reads `sweep.modelAliases` via the resolver. Overlay direction is correct (**config-wins / default-fallback**): the new assertions in the `resolve_model_alias` test show an unrelated `sonnet` override leaves `opus → claude-opus-5` as the shipped fallback, and a supplied override wins. Cross-tier merge is deliberately tested by `read_model_aliases_merges_across_tiers` (union of keys, higher-precedence tier wins per shared key). The Rust/Python blank-key divergence (Rust drops blank keys, Python keeps them) is documented in both the function doc comment and an inline comment, and left un-normalized as required (Python side owned by #4060).

**3. `main.rs::handle_validate_command` / `role_validation.rs`** — hard-fail precondition retargeted from "legacy file exists" to "resolved effective config has a non-empty `terminals` array". **Both** the `text` and `json` error branches now name every tier searched in precedence order (Finding 5). `validate_from_file` is retained as a thin backward-compat wrapper; `validate_from_config(&Value, mode)` + `extract_roles_from_value(&Value)` added with a shared `check_role_dependencies` helper. A repo with no config in any tier still exits 1 (`has_terminals == false`). New test `test_validate_from_project_tier_only` exercises the project-tier-only path; `test_validate_from_config_value` covers the Value entry point.

### General checks (run locally, not trusted from the description)
- `cargo clippy --all-targets -- -D warnings`: **clean**.
- `cargo test` (loom-daemon): all **unit tests pass**; CI's authoritative "Rust Unit Tests" job is green (3m38s). The one local failure — `test_factory_reset_loop_creates_clean_slate` — is an environment-dependent integration test that times out on the first `create_terminal_with_config` IPC call (it spawns real tmux sessions + a live daemon; no tmux server present locally). It never touches any config-resolution code path, so it is unrelated to this change.
- No direct `.loom/config.json` path construction remains at the three migrated sites. The remaining constructions in `loom-daemon/src/` are the ten mechanical readers explicitly scoped to sibling **#4058** (worktree_root, main_health_gate, work_finder, role_runner, token_ranking_refresh, watch_registry, and the other sweep_registry functions) plus doc comments/tests — consistent with the issue's Out of Scope section.
- Cross-language: `./.loom/scripts/resolve-model.sh opus` → `claude-opus-5`, matching Rust `resolve_model_alias`.

### Required decisions stated in the PR body (all present)
- Malformed-config diagnostic (Finding 3): accepted, with rationale.
- Blank-key Rust/Python divergence (Finding 2): left as a code comment, not normalized.
- Hard-fail precondition rule: stated verbatim for #4062 to mirror.

Merge state is CLEAN. LGTM — approved.

PR #4095, opened 2026-07-27
fix(tokens): percent-encode monitor DB path so mode=ro survives '?'/'#'

### LINKED ISSUE #4029: fix(tokens): percent-encode the monitor DB path — a '?' or '#' silently defeats mode=ro
## Context

Surfaced during the review of #4013 (`loom-tokens import-from-monitor`, closes #4006), initially as a cosmetic nit. It is not cosmetic: on certain paths it **silently voids the module's read-only guarantee**.

`monitor_db.py` builds the SQLite URI by string interpolation:

```python
# Read-only URI: this database belongs to claude-monitor. We never write,
# migrate, or take a write lock on it.
uri = f"file:{path}?mode=ro"
conn = sqlite3.connect(uri, uri=True, timeout=_SQLITE_TIMEOUT_SECONDS)
```

`path` is never percent-encoded, so a `?` or `#` in it terminates the URI path early and the `mode=ro` query parameter is lost. The connection still opens — **read-write** — against another tool's live database.

### Repro

```python
import sqlite3, os, tempfile
base = tempfile.mkdtemp()
for name in ['q?dir', 'sp ace', 'h#ash', 'pct%20']:
    d = os.path.join(base, name); os.makedirs(d, exist_ok=True)
    p = os.path.join(d, 'usage.db')
    sqlite3.connect(p).execute('create table t(x)')
    try:
        c = sqlite3.connect(f'file:{p}?mode=ro', uri=True)
        c.execute('select 1').fetchall()
        try:
            c.execute('create table z(y)'); ro = 'NOT-RO!'
        except sqlite3.OperationalError: ro = 'ro-ok'
        print(f'{name:10} OPEN OK  ({ro})')
    except Exception as e:
        print(f'{name:10} FAILED: {type(e).__name__}: {e}')
```

```
q?dir      OPEN OK  (NOT-RO!)     <- mode=ro lost; connection is WRITABLE
h#ash      OPEN OK  (NOT-RO!)     <- same
pct%20     FAILED: unable to open database file
sp ace     OPEN OK  (ro-ok)
```

Note the failure is silent in the dangerous direction: the open *succeeds*, so nothing surfaces, and `import-from-monitor` proceeds holding a writable handle on `usage.db`.

### Reachability

The path is not fully under Loom's control — `LOOM_CLAUDE_MONITOR_DIR` is operator-settable and the default lives under `$HOME`, so a directory containing `?`, `#`, or `%` anywhere in its ancestry is reachable without anything exotic. A `%` breaks the open outright, which is at least loud; `?` and `#` are the quiet ones.

### Fix

Encode the path before interpolation — `pathlib.Path.as_uri()` (which handles the drive/leading-slash cases too) or `urllib.parse.quote(str(path))`. Note that `as_uri()` requires an absolute path, which `monitor_db_path()` already guarantees.

## Also fold in (two minor items from the same review)

Small enough that they do not warrant separate issues, and they live in the same two files:

1. **Off-target error hint** — `monitor_db.py:181-186`. The `sqlite3.Error` handler on the `SELECT` attributes every failure to "This claude-monitor may predate the credential store." That is right for a missing-table `OperationalError`, but the same branch catches a hot-WAL-without-`-shm` failure, where the hint sends the reader down the wrong path.
2. **Weak typing in the printer** — `cli.py:451`. `_print_monitor_import(result: "object")` then reads every field through defensive `getattr(result, ..., default)`. The caller always passes a `MonitorImportResult`, so the annotation can be the real type and the `getattr` calls plain attribute access; as written a renamed field degrades silently to the default instead of failing.

- [ ] The monitor DB path is percent-encoded (or converted via `Path.as_uri()`) before being interpolated into the SQLite URI.
- [ ] A regression test parameterized over paths containing `?`, `#`, `%`, and a space asserts the connection both opens **and is genuinely read-only** — i.e. a write attempt raises — rather than merely asserting the open succeeded. The current tests pass on the buggy code precisely because they only assert the latter.
- [ ] The `SELECT` error path no longer asserts a missing-table cause for every `sqlite3.Error`.
- [ ] `_print_monitor_import` is annotated with `MonitorImportResult` and reads fields directly.

rjwalters:
## Summary

`read_monitor_credentials` built the read-only SQLite URI for claude-monitor's `usage.db` by raw string interpolation — `file:{path}?mode=ro`. Because `path` was never percent-encoded, a `?` or `#` anywhere in it terminated the URI path early and the `mode=ro` query parameter was silently dropped, opening another tool's **live** database read-**write**. A `%` was worse: it failed the open outright against a present, readable database. This voids a documented read-only guarantee, so it is a correctness/safety bug rather than a cosmetic nit.

## Changes

- **`monitor_db.py`** — new `_read_only_uri()` helper percent-encodes the path via `urllib.parse.quote` (default `safe='/'`) before interpolation, so `?`, `#`, `%`, and spaces can no longer void `mode=ro`. `quote` has no absolute-path precondition, so a **relative** `LOOM_CLAUDE_MONITOR_DIR` (which `monitor_db_path()` does not absolutize) still yields a usable URI rather than raising `ValueError` the way `Path.as_uri()` would. Also narrowed the `SELECT` error handler: only a missing-table `OperationalError` now earns the "may predate the credential store" hint; any other `sqlite3.Error` (hot WAL missing its `-shm`, corruption) surfaces its own message without the misleading hint.
- **`cli.py`** — `_print_monitor_import` is annotated with the real `MonitorImportResult` type (added to the import block) and reads fields via plain attribute access instead of defensive `getattr(result, ..., default)`, so a renamed field fails loudly rather than degrading to a default.
- **`test_monitor_db.py`** — regression tests (see below).

Note this also **repairs an outright-broken configuration**: a `%` in the path is a hard open failure today, so the fix is a behavior fix in that direction, not just a hardening.

- [x] Monitor DB path is percent-encoded before interpolation into the SQLite URI (`_read_only_uri`).
- [x] Regression test parameterized over `?`, `#`, `%`, and a space asserts the connection both opens **and is genuinely read-only** (a `CREATE TABLE` write attempt raises `sqlite3.OperationalError`) — not merely that the open succeeded.
- [x] The `SELECT` error path no longer asserts a missing-table cause for every `sqlite3.Error` (missing-table keeps the hint; a locked/other error omits it).
- [x] `_print_monitor_import` is annotated with `MonitorImportResult` and reads fields directly.
- [x] A relative `LOOM_CLAUDE_MONITOR_DIR` does not raise a bare `ValueError` (Curator Finding 1).
- [x] A path containing `%` opens successfully, where it fails today (Curator Finding 2).

## Test Plan

- `test_special_char_path_opens_and_is_read_only` — parameterized over `q?dir`, `h#ash`, `pct%20`, `sp ace`; opens via the module and asserts a write raises.
- `test_relative_monitor_dir_does_not_raise_value_error` — relative `LOOM_CLAUDE_MONITOR_DIR` via `monkeypatch.chdir`; succeeds, no `ValueError`.
- `test_percent_in_path_opens_where_it_failed_before` — `%` path now opens.
- `test_missing_table_keeps_predate_hint` / `test_non_missing_table_error_omits_predate_hint` — the narrowed hint.
- Full `loom-tools` tokens suite green: `324 passed`. Ruff clean on all three files.

Closes #4029

### DIFF
--- a/loom-tools/src/loom_tools/tokens/cli.py
+++ b/loom-tools/src/loom_tools/tokens/cli.py
@@ -37,6 +37,7 @@ from loom_tools.tokens import failure_counts
 from loom_tools.tokens.bootstrap import bootstrap_tokens
 from loom_tools.tokens.monitor_db import (
     MonitorDbUnavailable,
+    MonitorImportResult,
     import_from_monitor,
     monitor_db_path,
 )
@@ -448,22 +449,22 @@ def _cmd_import_from_monitor(args: argparse.Namespace) -> int:
     return 0
 
 
-def _print_monitor_import(result: "object") -> None:
+def _print_monitor_import(result: MonitorImportResult) -> None:
     """Print the imported account set. Secrets are never shown."""
-    db_path = getattr(result, "db_path", None) or monitor_db_path()
-    effective = getattr(result, "effective", []) or []
-    pruned = getattr(result, "pruned", []) or []
+    db_path = result.db_path or monitor_db_path()
+    effective = result.effective
+    pruned = result.pruned
 
     print(f"claude-monitor store: {db_path}")
-    print(f"Destination pool: {getattr(result, 'tokens_dir', None)}")
+    print(f"Destination pool: {result.tokens_dir}")
 
     if not effective:
         print("Active accounts: (none)")
         return
 
-    written = set(getattr(result, "written", []) or [])
-    unchanged = set(getattr(result, "unchanged", []) or [])
-    drifted = set(getattr(result, "drifted", []) or [])
+    written = set(result.written)
+    unchanged = set(result.unchanged)
+    drifted = set(result.drifted)
 
     print(f"Active accounts ({len(effective)}):")
     for acct in effective:
--- a/loom-tools/src/loom_tools/tokens/monitor_db.py
+++ b/loom-tools/src/loom_tools/tokens/monitor_db.py
@@ -47,6 +47,7 @@ import logging
 import os
 import re
 import sqlite3
+import urllib.parse
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -108,6 +109,26 @@ def monitor_db_path(monitor_dir: Path | None = None) -> Path:
     return base / MONITOR_DB_NAME
 
 
+def _read_only_uri(path: Path) -> str:
+    """Build the ``mode=ro`` SQLite URI for *path*.
+
+    The path is percent-encoded before interpolation. Without this, a ``?`` or
+    ``#`` anywhere in the path terminates the URI early and the ``mode=ro``
+    query parameter is silently dropped — the connection then opens
+    **read-write** against claude-monitor's live database (#4029). ``%`` fares
+    even worse: the unencoded form fails to open a database that is present and
+    readable.
+
+    ``urllib.parse.quote`` (default ``safe='/'``) encodes ``?``, ``#``, ``%``,
+    and spaces while leaving the path separators intact. Unlike
+    :meth:`pathlib.Path.as_uri` it has no absolute-path precondition, so a
+    relative ``LOOM_CLAUDE_MONITOR_DIR`` still yields a usable URI (resolved
+    against the cwd, as the unencoded form was) rather than raising
+    ``ValueError``.
+    """
+    return f"file:{urllib.parse.quote(str(path))}?mode=ro"
+
+
 def _email_from_label(label: str) -> str | None:
     """Recover an email from a credential label, or None if it isn't one.
 
@@ -159,8 +180,9 @@ def read_monitor_credentials(
         )
 
     # Read-only URI: this database belongs to claude-monitor. We never write,
-    # migrate, or take a write lock on it.
-    uri = f"file:{path}?mode=ro"
+    # migrate, or take a write lock on it. The path is percent-encoded so a
+    # '?' or '#' in it cannot silently void the mode=ro guarantee (#4029).
+    uri = _read_only_uri(path)
     try:
         conn = sqlite3.connect(uri, uri=True, timeout=_SQLITE_TIMEOUT_SECONDS)
     except sqlite3.Error as exc:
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: Approve

Reviewed against `main`, ran the full `loom-tools` test suite and `ruff check` locally in the PR worktree.

**Root cause & fix**
- Confirmed the bug: `f"file:{path}?mode=ro"` silently drops `mode=ro` whenever `path` contains `?` or `#` (URI parser treats the rest as the query string), and fails outright on `%`. This is a real read-only-guarantee violation against another tool's live DB, not a cosmetic nit — correctly scoped as a correctness/safety fix.
- `_read_only_uri()` fixes it with `urllib.parse.quote(str(path))` (default `safe='/'`), which encodes `?`, `#`, `%`, and spaces while preserving path separators. Verified empirically via the new tests that connections opened through this helper are still genuinely read-only (`CREATE TABLE` raises `OperationalError`), not merely "the open succeeded."
- Correctly identified that `Path.as_uri()` would have been the "obvious" alternative but has an absolute-path precondition that would regress relative `LOOM_CLAUDE_MONITOR_DIR` values into a bare `ValueError` — `urllib.parse.quote` avoids that trap. Good catch, and it's tested (`test_relative_monitor_dir_does_not_raise_value_error`).

**Error-hint narrowing** — the `SELECT` handler now only attaches the "may predate the credential store" hint when the underlying error is actually a missing-table `OperationalError`, rather than asserting that for every `sqlite3.Error` (e.g. a locked DB / missing `-shm` sidecar). Both branches are tested (`test_missing_table_keeps_predate_hint`, `test_non_missing_table_error_omits_predate_hint`), the latter via a clean `sqlite3.connect` monkeypatch rather than something fragile.

**`cli.py`** — `_print_monitor_import` now takes the real `MonitorImportResult` type and reads fields directly instead of `getattr(..., default)`. Verified `MonitorImportResult.__init__` always sets every field referenced (`db_path`, `tokens_dir`, `effective`, `written`, `unchanged`, `drifted`), so the direct-attribute-access change is safe and correctly fails loudly on a future rename instead of degrading silently.

**Tests** — parameterized over `?`, `#`, `%`, and a space; each case checks both that the credential is readable *and* that a write attempt on the same URI fails, which is the regression that actually matters here (the pre-fix code already "worked" for the open in 3 of 4 cases — the missing assertion was read-only-ness). Also covers the relative-dir and bare-`%` regressions the curator flagged.

**Verification**
- `uv run pytest` (full `loom-tools` suite): 2050 passed, 40 skipped.
- `uv run ruff check` on the three changed files: clean.
- `gh pr checks`: all applicable checks green (Rust/MCP/installer jobs correctly skip via path-filter — Python-only change).

No changes requested. Moving to `loom:pr` for Champion merge.

rjwalters:
**Champion Auto-Merge**

- Size check passed (184 lines: +171/-13, limit 200)
- No critical files modified
- No merge conflicts
- Updated recently (-6 hours ago)
- All CI checks passing (required checks skipped as not applicable)

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #4014, opened 2026-07-27
CLAUDE.md is 725 lines / ~13.6k tokens in every agent's context: split operating core from reference and history


rjwalters:
## Problem

`CLAUDE.md` is **725 lines / 54 KB / ~13.6k tokens**, and it is prepended to the context of *every* session, *every* worker role, and *every* sweep child. That is a fixed tax paid thousands of times a day across the daemon's dispatch volume, and it grows monotonically — no process currently removes anything from it.

Growth is steady rather than spiky, which is why it keeps escaping notice:

| Revision | Lines |
|---|---|
| `HEAD~120` | 649 |
| `HEAD~60` | 670 |
| `HEAD~20` | 718 |
| `HEAD` | 725 |

Full disclosure: #4006 / PR #4013 adds another ~38 lines. Every individual addition is defensible; the aggregate is the problem, and there is no counter-pressure.

## Why it has grown

The file has drifted from *operating instructions* into a **changelog / reference hybrid**. Three distinct kinds of content are interleaved:

1. **Durable operating instructions** — the label workflow, "never use `gh pr merge`", worktree rules, the mandatory sweep lifecycle. An agent genuinely cannot act correctly without these. This is what CLAUDE.md is *for*.
2. **Reference material** — the guard-hook fast-path allowlist catalog, token-pool internals, selection-algorithm tiers, daemon config tables. Needed occasionally, by whoever is touching that subsystem; already partly duplicated in `.loom/docs/`.
3. **Historical narrative** — the v0.10.0 shepherd-deprecation phase table (all rows "shipped"), removed-entry-point tables, per-issue implementation notes carrying issue numbers and superseded behavior. Near-zero value at inference time; this is what `docs/migration/` and ADRs are for.

Categories 2 and 3 are the bulk of the file and the bulk of the growth.

## Suggested direction

Not a mechanical line-cut — a split along the three categories above:

- [ ] **Keep an operating core** in `CLAUDE.md`: what an agent must know to act correctly *right now*. Rough target ~250–300 lines.
- [ ] **Move reference detail to `.loom/docs/*`** and leave a one-line pointer. Several sections (daemon config, token pool) already have a `.loom/docs/` counterpart that is *more* complete — the CLAUDE.md copy is a lossy duplicate that can go.
[...]

PR #4096, opened 2026-07-27
feat(tokens): warn when bootstrap disagrees with claude-monitor's live store

### LINKED ISSUE #4030: feat(tokens): bootstrap should warn when accounts.env disagrees with claude-monitor's live store
## Context

#4006 / PR #4013 shipped the **cure** for a stale token pool (`loom-tokens import-from-monitor` reads claude-monitor's live credential store). This issue is about the **detection** gap it leaves behind, which is unchanged.

### What happened (2026-07-26/27)

All seven accounts were re-authenticated in claude-monitor. claude-monitor wrote the fresh tokens to `~/.claude-monitor/usage.db` (`oauth_credentials`) and did **not** touch `~/.claude-monitor/accounts.env`, which is the file `loom-tokens bootstrap` reads. `accounts.env` stayed frozen at its Jul 22 contents.

Consequences, in order:

1. Every pool token returned `401 {"type":"authentication_error","message":"OAuth access token has been revoked."}`.
2. `loom-tokens check` marked all seven accounts `blocked`.
3. The daemon's dynamic cap collapsed to `min(healthy 0 × per-token N, …) = 0` and **dispatch stopped entirely** for hours.

The recovery attempt made it worse in the sense that it looked like it should have worked: `loom-tokens bootstrap --force` reported `written=7` and changed nothing that mattered, because it faithfully rewrote the same revoked tokens — the snapshot *itself* was what had gone stale. Confirming this required hand-comparing sha256[:8] fingerprints between `accounts.env` and the live DB.

At no point did any Loom command say "the source you are reading from disagrees with the live store." The tool reported success at every step while the pool was dead.

### Why `import-from-monitor` does not close this

It closes it only for someone who already knows to reach for it. `bootstrap` remains the documented first move in CLAUDE.md's setup flow and in `.loom/docs`, it is what a returning operator's muscle memory reaches for, and it still cannot detect that its own input is stale. The next occurrence of this failure looks identical to the last one right up until someone thinks to diff two files by hand.

## Proposed

When `bootstrap` runs and claude-monitor's store is present on the host, compare the resulting key fingerprints against the live store's, and warn on divergence — pointing at `import-from-monitor` as the fix.

Design notes:

- **Warn, never auto-switch.** `bootstrap`'s contract is "materialize what the configured sources say". Silently preferring a different source would be a worse surprise than the one being fixed. The operator should be told, not overridden.
- **Absent store is not a warning.** A host with no claude-monitor is a normal, fully-supported configuration; it must stay silent there.
- **Read-only and non-fatal.** The comparison must not fail the bootstrap if the DB is unreadable, locked, or an older schema — degrade to silence. (See the `mode=ro` correctness fix filed alongside this.)
- Fingerprint comparison already exists in `materialize_accounts()`, and `read_monitor_credentials()` already yields the live set, so this is mostly wiring rather than new machinery.
- Worth considering the symmetric case: `loom-tokens check` reporting *every* account `blocked` is itself a near-certain signature of revoked-not-exhausted credentials, and would be a good second place to surface the same hint.

- [ ] With claude-monitor present and its live tokens differing from those in `accounts.env`, `bootstrap` emits a clear warning naming `import-from-monitor` as the remedy.
- [ ] With no claude-monitor store present, output is byte-for-byte unchanged from today.
- [ ] An unreadable, locked, or older-schema store degrades to silence and never fails the bootstrap.
- [ ] No secret material is printed — fingerprints and emails only, consistent with `index.json`.
- [ ] Tests cover: agreement (silent), divergence (warns), store absent (silent), store unreadable (silent).

rjwalters:
## Summary

`loom-tokens bootstrap` rewrites `accounts.env`-derived tokens without ever
consulting claude-monitor's **live** credential store (`~/.claude-monitor/usage.db`).
In the incident that motivated this (#4030), every account was re-authenticated in
claude-monitor — which updated `usage.db` but not the `accounts.env` snapshot — so
`bootstrap --force` faithfully rewrote the revoked tokens, reported `written=N`, and
the pool stayed dead. The only diagnosis was hand-comparing sha256 fingerprints
between two files.

This adds a detection step at the end of `bootstrap_tokens()`: it compares the
merged set it is about to write against the live store and, on divergence, warns
and names `import-from-monitor` as the remedy.

## Changes

- **`loom-tools/src/loom_tools/tokens/bootstrap.py`** — new `_check_monitor_divergence()`
  helper, called last in `bootstrap_tokens()` (after the summary/drift log, so the
  warning is the final line; runs under `--dry-run` too). Compares
  `fingerprint(acct.key)` for the merged `valid` set against `fingerprint(cred.token)`
  from `read_monitor_credentials()`, joining on lowercased email, warning only on the
  intersection whose fingerprints differ. The warning explicitly redirects away from
  the misleading `--force` advice. `read_monitor_credentials` is imported
  **function-locally** to avoid the existing `monitor_db → bootstrap` import cycle.
- **`loom-tools/tests/tokens/conftest.py`** — relocated `make_monitor_db()` and the
  `monitor_db` fixture here so both test modules share one `usage.db` builder.
- **`loom-tools/tests/tokens/test_monitor_db.py`** — import the relocated helpers.
- **`loom-tools/tests/tokens/test_bootstrap.py`** — new `TestMonitorDivergenceWarning`
  class (10 tests).
- **`.loom/docs/token-pool.md`** — note that `bootstrap` now detects the stale-snapshot
  condition. (The "Importing live tokens from claude-monitor" section was moved out of
  `CLAUDE.md` into this doc by the #4014/#4089 split; `CLAUDE.md` now only carries a
  one-line pointer and is budget-checked, so this is the correct structural home. No
  `defaults/` mirror exists for this file — `defaults/docs/` holds only
  `ci-integration.md` — so there is no drift to keep in sync.)

## Design decisions (per issue + curator guidance)

- **Warn, never auto-switch** — `bootstrap`'s contract is "materialize the configured
  sources"; the operator is told, not overridden.
- **Read-only and non-fatal** — a single `except MonitorDbUnavailable` covers absent
  store, unreadable/locked DB, and older schema (the reader normalizes all three).
- **Membership differences stay silent** — an email on only one side is a normal
  configuration (repo-local account, or a monitor account the pool pins out); warning
  on it would train operators to ignore the message.
- **No secrets** — emails and 8-char fingerprints only, consistent with `index.json`.
- **`mode=ro` fix (#4029)** already merged — `read_monitor_credentials()` opens the DB
  `file:…?mode=ro` on `main`, so no dependency remains.

### Out of scope (per curator, deferred to follow-ups)

- The symmetric `loom-tokens check` all-`blocked` hint (a different mechanism — a
  heuristic over probe results, not a source comparison).
- Membership-divergence warnings (monitor holds accounts the merged set omits).

- [x] claude-monitor present + live tokens differ → warning names `import-from-monitor`
      (`test_divergence_warns_and_names_remedy`)
- [x] No claude-monitor store present → output unchanged / silent
      (`test_store_absent_is_silent`)
- [x] Unreadable / locked / older-schema store → silent, never fails the bootstrap
      (`test_unreadable_store_is_silent`, `test_older_schema_store_is_silent`)
- [x] No secret material printed — fingerprints + emails only
      (`test_no_secret_material_is_printed`)
- [x] Tests cover agreement (silent), divergence (warns), absent (silent),
      unreadable (silent) — plus membership difference, case-insensitive email join,
      and `--dry-run`.

## Test Plan

- [x] `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -q` →
      **325 passed** (was 315; +10 new, conftest relocation keeps `test_monitor_db`
      green).
- [x] `ruff check` passes on all changed files.

Closes #4030

### DIFF
--- a/.loom/docs/token-pool.md
+++ b/.loom/docs/token-pool.md
@@ -91,6 +91,14 @@ accounts `blocked`, the daemon's dynamic concurrency cap collapses to
 **`bootstrap --force` does not fix it** — it faithfully rewrites the same revoked
 tokens, because the snapshot itself is what went stale.
 
+**`bootstrap` now detects this condition (#4030).** When `usage.db` is present and
+the tokens `bootstrap` is about to write disagree with the live store (same email,
+different fingerprint), it prints a warning naming the diverging accounts and
+pointing at `import-from-monitor` — so the stale snapshot is caught automatically
+instead of by hand-comparing fingerprints. The check is read-only, warns but never
+auto-switches sources, and is silent when no `usage.db` is present or it is
+unreadable; it prints emails and 8-char fingerprints only, never secret material.
+
 `loom-tokens import-from-monitor` reads the live store directly and is **the
 standard way to populate a new host's pool** (it replaces hand-copying a pool
 between machines):
--- a/loom-tools/src/loom_tools/tokens/bootstrap.py
+++ b/loom-tools/src/loom_tools/tokens/bootstrap.py
@@ -605,6 +605,82 @@ def write_index(
 _HOME_UNSET = object()
 
 
+def _check_monitor_divergence(valid: list[Account]) -> None:
+    """Warn when the merged bootstrap set disagrees with claude-monitor's live store.
+
+    The incident this guards against (#4030): claude-monitor re-authenticated
+    every account, writing fresh tokens to ``usage.db`` but never touching
+    ``accounts.env`` — the static snapshot ``bootstrap`` reads. ``bootstrap
+    --force`` then faithfully rewrote the *revoked* tokens and reported
+    ``written=N`` while the pool was dead; the only diagnosis was hand-comparing
+    sha256 fingerprints between two files. This surfaces that condition
+    automatically and names the remedy.
+
+    Compares what bootstrap is about to write (``valid``) against the live store,
+    joining on lowercased email (matching :func:`merge_accounts` and
+    ``read_monitor_credentials``), and warns on the **intersection** whose
+    fingerprints differ — never auto-switching sources.
+
+    Read-only and non-fatal by contract:
+
+    * **Absent store** (no claude-monitor / no ``usage.db``) — silent. A host
+      without claude-monitor is a fully-supported configuration.
+    * **Unreadable / locked / older-schema store** — silent. The reader
+      normalizes every such failure into :class:`MonitorDbUnavailable`.
+    * **Membership-only difference** (an email present on only one side) —
+      silent. A repo-local account absent from claude-monitor, or a monitor
+      account the pool legitimately pins out, is normal; warning on it would fire
+      on healthy hosts and train operators to ignore the message — the exact
+      failure this check exists to fix.
+
+    No secret material is printed — emails and 8-char fingerprints only,
+    consistent with ``index.json``.
+    """
+    # Deferred import: ``monitor_db`` imports from this module (``Account``,
+    # ``materialize_accounts``, ...), so a top-level import here would create an
+    # import cycle and fail at load time. Keep this import function-local — do
+    # NOT hoist it to the module top (#4030).
+    try:
+        from loom_tools.tokens.monitor_db import (
+            MonitorDbUnavailable,
+            read_monitor_credentials,
+        )
+    except ImportError:
+        return
+
+    try:
+        creds = read_monitor_credentials()
+    except MonitorDbUnavailable:
+        # No store, or an unreadable / locked / older-schema one — degrade to
+        # silence. Detection must never fail an otherwise-successful bootstrap.
+        return
+
+    live_fp = {c.email.strip().lower(): fingerprint(c.token) for c in creds}
+    diverged: list[tuple[str, str, str]] = []
+    for acct in valid:
+        live = live_fp.get(acct.email.strip().lower())
+        if live is None:
+            continue  # membership difference — not a key-divergence signal
+        env_fp = fingerprint(acct.key)
+        if live != env_fp:
+            diverged.append((acct.email, env_fp, live))
+
+    if not diverged:
+        return
+
+    detail = ", ".join(
+        f"{email} (merged fp={merged_fp}, live fp={monitor_fp})"
+        for email, merged_fp, monitor_fp in diverged
+    )
+    log_warning(
+        f"{len(diverged)} account(s) in the merged bootstrap sources disagree "
+        f"with claude-monitor's live store (usage.db): {detail}. The snapshot in "
+        f"accounts.env is stale, not the on-disk tokens — `--force` will NOT fix "
+        f"this, it rewrites the same stale values. Run "
+        f"`loom-tokens import-from-monitor --force` to adopt the live credentials."
+    )
+
+
 def bootstrap_tokens(
     repo_root: Path,
     *,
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed against #4030's acceptance criteria and this looks complete and well-executed.

## Verification

- **Checkout**: used the existing `.loom/worktrees/issue-4030` builder worktree.
- **Merge state**: `CLEAN`, branch is up to date with `main` (only 1 commit behind, `git merge-tree --write-tree origin/main HEAD` produces a clean tree with no conflict markers).
- **CI**: `gh pr checks` — `Detect Changes` and `CLAUDE.md Line Budget` pass; Rust/MCP/installer jobs correctly `skipping` (no Rust/Cargo changes). There's no GitHub Actions Python CI job in this repo, so I ran the suite locally per the PR's stated test plan.
- **Tests**: `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -q` → **325 passed** (matches the PR description).
- **Lint**: `ruff check` on all Python files this PR touches (`bootstrap.py`, `conftest.py`, `test_bootstrap.py`, `test_monitor_db.py`) → clean. (Ruff does flag 3 unrelated `F401`s in `test_agent_spawn_integration.py` / `test_bad_tokens.py` / `test_failure_counts.py`, but those files aren't touched by this PR — pre-existing, not blocking.)

## Code review

- `_check_monitor_divergence()` is read-only and non-fatal: the single `except MonitorDbUnavailable: return` correctly collapses the absent/unreadable/locked/older-schema cases the reader already normalizes, and the function-local import of `monitor_db` avoids the existing `monitor_db → bootstrap` import cycle (documented inline, and a regression test exercises the older-schema and unreadable-file paths directly).
- Warn-never-switch is respected — the function only ever calls `log_warning`, never touches `valid` or the write path.
- Membership-only differences (email present on only one side) are correctly excluded from the divergence set — matches the "silent on healthy hosts" requirement in the issue.
- Fingerprint comparison reuses the existing `fingerprint()` helper (same one `materialize_accounts`/`index.json` use) rather than reimplementing hashing — good reuse per the issue's design notes.
- No secret material reaches the log line — verified via `test_no_secret_material_is_printed`, and confirmed `log_warning` writes to stderr while `--emit-json` writes to stdout, so the new warning can't corrupt structured JSON output for scripted callers.
- Called last in `bootstrap_tokens()`, after the summary/drift log, and explicitly runs under `--dry-run` too (read-only detection, so that's the right call) — both are covered by tests.
- The `conftest.py` relocation of `make_monitor_db`/`monitor_db` is a clean, low-risk refactor — `test_monitor_db.py`'s own tests still pass post-move.

## Test coverage

The 10 new tests in `TestMonitorDivergenceWarning` hit every scenario from the issue's acceptance criteria: divergence-warns-and-names-remedy, agreement-is-silent, store-absent-is-silent, unreadable-store-is-silent, older-schema-store-is-silent, membership-difference-is-silent, no-secret-material-is-printed, case-insensitive email join, and dry-run still warns. Good breadth.

## Test Plan (from PR description)

1. `PYTHONPATH=loom-tools/src python3 -m pytest loom-tools/tests/tokens/ -q` → ✅ Executed: 325 passed.
2. `ruff check` passes on all changed files → ✅ Executed: clean.

No blocking issues found.

Issue #4069, opened 2026-07-27
fix(daemon): `loom-daemon status` cannot distinguish "not installed" from "not running" from "running but unresponsive"


rjwalters:
Split out of #4011 (*Silent autonomy loss: daemon left booted-out of launchd, dispatch stopped, nothing reported it*). #4011's own merge commit (`116d3278`) records the split explicitly: *"Scope is the Champion-approved AC1+AC2 (**AC3 → #4069**, AC4 shipped as #4054/#4077)."* This issue is that AC3.

Filed separately because it is a purely-local, read-only probe layered onto one error branch, independently useful and independently verifiable.

> **⚠️ Curator refresh 2026-07-27 — read this before implementing.** #4011's remainder **merged** as `116d3278` (PR #4103) *after* this issue was written. That delivered the durable **autonomy-desired marker**, the **heartbeat file**, and the **host-side watchdog** — which this issue's original text listed as future work in *Non-goals*, and which its original test plan said made deliberate-stop-vs-silent-death undistinguishable "here". Both statements are now obsolete. The marker is **self-describing** and is the primary probe input; see *Implementation guidance* below. All line anchors have been re-verified against `origin/main` @ `a1b8b4af`.

## Problem

Every transport failure collapses into one generic message. `handle_status_command` (`loom-daemon/src/main.rs:1930`, error arm at **`:1933-1951`**) has a single `Err` arm:

```rust
let report = match query_daemon_status(&socket_path).await {
    Ok(report) => report,
    Err(e) => {
        if json {
            let err = serde_json::json!({
                "error": format!("could not reach loom-daemon at {}: {e}", socket_path.display()),
            });
            println!("{}", serde_json::to_string_pretty(&err)?);
        } else {
            eprintln!("Could not reach loom-daemon at {}: {e}", socket_path.display());
            eprintln!();
            eprintln!("Is the daemon running? Start it with:");
            eprintln!("  ./.loom/scripts/cli/loom-daemon-start.sh");
        }
        std::process::exit(1);
    }
};
```

`query_daemon_status` (**`main.rs:1488`**) does distinguish four *transport* failures internally — connect-timeout, connect-failed, closed-connection-without-responding, and round-trip-timeout — but it never consults any **install-state or intent** source: not the autonomy-desired marker, not the heartbeat, not `.loom/.daemon.pid`, not `launchctl print`.

So these materially different situations are byte-identical to an operator, and the suggested remedy is wrong in most of them:

[...]

Issue #3969, opened 2026-07-26
Daemon should refresh token ranking itself instead of relying on an operator-managed probe-tokens cron


rjwalters:
## Context

Token selection prefers `.loom/tokens/.ranking`, which goes stale unless an operator sets up an external cron for `./.loom/scripts/probe-tokens.sh --ranking` (documented in CLAUDE.md as a crontab example). During the 2026-07-26 canary run the operator session was manually refreshing the ranking, and listed it among the manual steps with retirement paths: "the ranking refresh (a cron for probe-tokens.sh --ranking would retire that)."

The daemon is a long-lived process that already owns dispatch cadence and consumes the ranking (via `spawn-claude.sh` selection); it is the natural home for this loop. With the shared machine-level pool (#3938), refreshing in the selected pool location also benefits every managed repo at once.

## Suggested acceptance criteria
- [ ] Daemon runs the ranking probe periodically (default ~10 min, configurable via `autonomous` config + env, precedence env > config > default), writing `.ranking` atomically in the effective pool (per-repo or shared, matching #3938 resolution)
- [ ] Default-on is acceptable since it's read-only probing, but respect a disable knob; probe failures are logged and skipped, never fatal to the daemon
- [ ] No double-probing when an operator cron also runs (idempotent/atomic writes make concurrent refresh harmless)
- [ ] Docs: drop the "set up a cron" example to an optional fallback for non-daemon setups

PR #4103, opened 2026-07-27
feat(daemon): autonomy-loss detector — intent marker + heartbeat + host-side watchdog (#4011)

### LINKED ISSUE #4011: Silent autonomy loss: daemon left booted-out of launchd, dispatch stopped, nothing reported it

rjwalters:
Closes #4011

## Problem

On 2026-07-26 the `loom-daemon` launchd job took a SIGTERM two seconds after starting and was left `bootout`-ed (unloaded) from launchd. Autonomous dispatch silently stopped, and **nothing surfaced it** — no log line, no forge signal, no notification. It was found hours later only because someone happened to run `loom-daemon status` by hand. That pull is exactly the thing that didn't happen.

## Scope (Champion-approved AC1 + AC2 only)

Per the pinned scope on the issue: **AC1 (detect "a daemon should be running but isn't") + AC2 (report without an operator thinking to look)**. AC3 (status disambiguation) is excised to #4069; AC4 (self-modification guard) shipped as #4054/#4077; flipping `KeepAlive` and a forge-side channel are explicit out-of-scope follow-ups.

## Mechanism — three cooperating parts, reporter lives OUTSIDE the daemon

A dead daemon cannot report its own death (which is why #3971's in-daemon watch loop is not reusable as the reporter).

1. **Durable autonomy-desired marker** (`<loom_dir>/autonomy-desired`). Written by `loom-daemon-start.sh` on a successful start; removed **only** by an operator-initiated `loom-daemon-stop.sh`. Its lifetime is **operator intent**, not process liveness — the pid file and loaded launchd job are both destroyed by the very stop path that causes an outage, so a detector built on them couldn't tell a deliberate stop from a silent death. `loom-daemon-update.sh` passes `stop --restarting` (`LOOM_DAEMON_STOP_KEEP_INTENT=1`) so a self-update never silently disarms the detector — an **explicit** signal, never inferred.

2. **Declared-cadence heartbeat** (`<loom_dir>/daemon.heartbeat`, new `loom-daemon/src/daemon_heartbeat.rs`). The daemon rewrites it every `intervalSecs` (default 60, default-on, `autonomous.heartbeat` / `LOOM_DAEMON_HEARTBEAT*`). Deliberately an **explicit** liveness contract, not the config-disableable `.ranking` mtime (#3969).

3. **Host-side watchdog** (`loom-daemon-watchdog.sh`), the payload of a **second** launchd job (`<label>-watchdog`) that `start.sh` provisions on macOS with `StartInterval` (default 300s). It compares intent vs reality (daemon loaded + alive? heartbeat fresh?) and reports divergence to `<loom_dir>/logs/daemon-watchdog.log` + stderr. `StartInterval` (not `KeepAlive`/resident) resolves who-watches-the-watchdog: an interval job owns no long-lived process, so it structurally cannot crash-and-stay-dead.

### Decision matrix (marker present ⇒ a daemon is expected)

| Reality | Watchdog |
|---|---|
| alive, heartbeat fresh | silent |
| alive, heartbeat **stale** | **report** (wedged) |
| alive, no heartbeat file | silent (liveness-only) |
| **not loaded/alive** | **report** (the #4011 outage) |
| marker **absent** | silent (deliberate stop — no false page) |

## Tests

- **14 Rust unit tests** (`daemon_heartbeat`): write/atomic-rename, path resolution, config soft-fail, env>config>default precedence, loop writes-immediately-and-refreshes-mtime.
- **12-case watchdog divergence matrix** (`test-loom-daemon-watchdog.sh`): dead / stale / fresh / no-heartbeat / marker-absent / under-threshold-no-flap / `STALE_SECS` override / launchd-stub / help.
- **Marker lifecycle** across start/stop (`test-loom-daemon-{start,stop}.sh`): written on start; removed on operator stop; **preserved** on `--restarting` and `LOOM_DAEMON_STOP_KEEP_INTENT=1`.
- All five daemon lifecycle suites pass (start 11, stop 22, update 33, launchd-plist 26, watchdog 12) + full daemon lib suite (1044). `render_launchd_plist`'s `KeepAlive`/`RunAtLoad` are untouched (Correction 1 honored). Every test pins `LOOM_AUTONOMY_MARKER`/`LOOM_SOCKET_PATH` into a tempdir so none can touch the operator's real `~/.loom`; `shellcheck --severity=warning` and `cargo clippy`/`cargo fmt --check` are clean.

## Docs

`.loom/docs/daemon-reference.md` gains an "Autonomy-loss watchdog + heartbeat (#4011)" section (mechanism, config table, decision matrix, the `--restarting` self-update contract, Linux/`--no-launchd` degradation) plus the `autonomous.heartbeat` config block.

## Meta-risk

This edits the very scripts whose failure caused the outage. Correctness is covered by the automated suites above; a real host stop/start/update cycle against an isolated `LOOM_LAUNCHD_LABEL` is the recommended manual verification before enabling in production (never against the live `com.rjwalters.loom-daemon`).

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1375,6 +1375,10 @@ concurrency ceiling 5" and share it with the team:
       "intervalSecs": 120,
       "expirySecs": 86400
     },
+    "heartbeat": {
+      "enabled": true,
+      "intervalSecs": 60
+    },
     "dispatchStaggerMs": 2000,
     "watchdog": {
       "enabled": true,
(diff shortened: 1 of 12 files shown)


rjwalters:
## Judge review — APPROVED

Independently re-verified the scoping claims and did a full local build/test pass rather than trusting the Builder self-report.

### Scoping claims — verified against source, not just trusted
- **Issue #4011** body + Champion review trail confirms this PR's stated scope (AC1 durable intent marker + AC2 explicit heartbeat/host-side watchdog) is exactly what Champion recorded as binding in-scope after two review cycles ("Recording it here as binding... In scope: original AC1... and AC2...").
- **AC3 → #4069**: confirmed open (`state: OPEN`), title is exactly the `status` not-installed/not-running/unresponsive disambiguation Champion split out — not a phantom follow-up.
- **AC4 → #4054/#4077**: #4054 (closed, "already covered elsewhere") was implemented and merged as PR #4077 ("supervised restart primitive via launchd KeepAlive:SuccessfulExit") — the live launchd transcript in that PR proves the restart-and-verify path this AC calls for. Reasonable, traceable coverage.

### Build / test verification (all run locally against the PR branch, not taken on faith)
- `cargo test --lib daemon_heartbeat` → 14/14 pass
- `cargo clippy --all-targets --all-features -- -D warnings` → clean
- `cargo fmt --check` → clean
- `shellcheck --severity=warning` on all touched/new shell files → clean
- `test-loom-daemon-watchdog.sh` → 12/12 pass
- `test-loom-daemon-start.sh` → 11/11 pass
- `test-loom-daemon-stop.sh` → 22/22 pass
- `test-loom-daemon-update.sh` → 33/33 pass
- `test-loom-daemon-launchd-plist.sh` → 26/26 pass (confirms #4054's `KeepAlive:SuccessfulExit` is untouched by this PR, per the PR's own "Correction 1 honored" claim)
- `cargo test --lib` (full workspace lib suite) → **1044/1044 pass**, 0 failed — matches the PR description's claimed count exactly.
- `gh pr checks 4103` → all CI checks green (Detect Changes, Shellcheck, CLAUDE.md Line Budget, Rust Linting & Build, Rust Unit Tests, Installer Integration Tests; MCP Server Build skipped as expected for a non-MCP change).
- `mergeStateStatus` → `CLEAN`, `mergeable` → `MERGEABLE`.

### Code quality / security notes
- Intent-marker writes (`write_intent_marker`) and heartbeat writes (`daemon_heartbeat.rs::write_heartbeat`) are correctly atomic (temp file + rename); marker parsing in the watchdog uses `grep`/`cut`, never `source`/`eval`, so a corrupted or hostile marker file can't execute code.
- Generated launchd plist XML (both the daemon's and the new watchdog's) is passed through the existing `xml_escape` helper — no injection surface there.
- `provision_watchdog_job` is correctly non-fatal (best-effort): a watchdog provisioning failure never blocks/breaks the daemon start.
- The `--restarting` / `LOOM_DAEMON_STOP_KEEP_INTENT` signal is explicit rather than inferred, exactly as the issue's Champion review demanded (self-update must never silently disarm the detector).
- Every test suite pins `LOOM_AUTONOMY_MARKER` / `LOOM_SOCKET_PATH` / `LOOM_WATCHDOG_LABEL` into a tempdir — confirmed none of them can touch the operator's real `~/.loom` or the real `com.rjwalters.loom-daemon(-watchdog)` LaunchAgents.

### Minor, non-blocking observation
- The term "watchdog" is now used for two unrelated concepts in this repo: the pre-existing in-daemon `autonomous.watchdog` config block (sweep review-stall detection) and this PR's new host-side `loom-daemon-watchdog.sh` launchd job (autonomy-loss detection). They don't collide in the config namespace (the new one is env-var-only, no `autonomous.*` key), but a future doc pass could rename one for clarity. Not a blocker.

Solid, well-scoped increment with unusually thorough test coverage and honest documentation of what's deliberately out of scope. Approving.

PR #4104, opened 2026-07-27
fix(mcp): settle unary daemon request on first newline frame, not on connection close

### LINKED ISSUE #4043: fix(daemon): MCP dispatch_sweep/get_sweep_status never respond (1800s) although the operation succeeds; CLI IPC stays fast
## Context

Observed repeatedly on 2026-07-27 while dispatching and inspecting a sweep.

`mcp__loom__dispatch_sweep` and `mcp__loom__get_sweep_status` both returned **no response and no progress for 1800s** and were aborted by the client:

```
MCP server "loom" tool "dispatch_sweep" sent no response or progress for 1800s; aborting.
```

**The underlying operations succeeded.** The dispatch really happened — `.loom/logs/sweep-issue-4034.log` shows the child spawning ~90s after the call, and that sweep went on to open PR #4036:

```
==== loom-daemon dispatch: 2026-07-27T15:08:41Z sweep_id=sweep-issue-4034-1785164919 issue=4034 ====
spawn-claude: using OAuth account 'agent1-2amlogic' (mode=ranked)
```

So the daemon acted correctly and only the MCP response never came back.

Crucially, the **CLI path stayed fast throughout** — `loom-daemon status` returned in well under a second during the same window, including the in-flight sweep table. That isolates the fault to the MCP/IPC response path rather than a wedged daemon or a wedged forge.

Consequences: an operator cannot use the documented Tier 2 surface. Worse, `dispatch_sweep` returning nothing invites a duplicate dispatch of work that already started (see #4007), and `cancel_sweep` is presumably unusable by the same mechanism — so there is no way to stop a runaway sweep via MCP.

Note this is **not** the #4001 `list_sweeps` hang: the binary in use was `7b84bdb8`, which already contains that fix, and reproduced again on `a68099d5`.

- [ ] `dispatch_sweep` / `get_sweep_status` return promptly, or return a real error, rather than going silent indefinitely.
- [ ] Root-cause why the CLI IPC path returns while the MCP path does not for the same request variants.
- [ ] A silent MCP call is distinguishable from a slow one (progress/heartbeat, or a server-side timeout with a diagnosable message).
- [ ] Confirm whether `cancel_sweep` is affected — if so it is the highest-severity instance, since it is the only way to stop a running sweep.

---

## Curator Enhancement (2026-07-27, revised)

> **Correction.** An earlier revision of this section claimed a stale bundle was
> the whole story and that "a rebuild alone resolves the symptom." **That was
> wrong.** Deeper tracing found a second, primary defect: the MCP bridge's unary
> request path **can never resolve on success**, stale or not. A rebuild converts
> a silent 1800s hang into a deterministic 30s failure — an improvement, not a
> fix. Both defects are scoped below.

### Defect 1 (primary): the unary IPC path settles on an event the daemon never emits

`mcp-loom/src/shared/daemon.ts:148-155` resolves `sendDaemonRequest` **only** in
the socket's `end` handler — it accumulates `data` into a buffer and parses it
when the connection closes. The newline the daemon writes is never used as a
frame boundary on this path.

But the daemon holds the connection open **by design**. `handle_client`
(`loom-daemon/src/ipc.rs:303`) is a persistent `while let Some(line) =
lines.next_line().await?` loop: one response frame per request line, connection
reusable, closing only when the *client* sends EOF. `sendDaemonRequest` never
calls `socket.end()`, so neither side ever closes.

**Measured against the live daemon** (raw socket, no MCP layer):

```
ListSweeps       DATA +2ms len=251    NO-END after 3000ms  <-- STAYS OPEN
DaemonStatus     DATA +6ms len=3070   NO-END after 3000ms  <-- STAYS OPEN
GetSweepStatus   DATA +0ms len=286    NO-END after 3000ms  <-- STAYS OPEN
```

The daemon answers in **0–6 ms** and never closes. The response is sitting in the
client's buffer, complete and unparsed, while the promise waits forever for an
`end` that is not coming. This is not a deadlock, a wedged `gh`, or lock
contention — it is a protocol/framing mismatch.

This is **long-standing, not a regression**: the persistent read loop dates to the
first daemon commit (`01a422eb`, #9) and the `end`-settle to `34723f22` (#1157).

**It also explains the CLI/MCP asymmetry the report flagged.** Same socket, same
daemon — the difference is entirely client-side framing:

| | MCP bridge | `loom-daemon status` / `dispatch` |
|---|---|---|
| Settles on | `socket.on("end")` — never fires (`daemon.ts:148`) | first `\n`-delimited line (`main.rs:1477-1481`) |
| Result | waits forever | returns in ms |

And it explains why **the dispatch really happened while the call returned
nothing** — server-side work completes normally; only the reply is never read.

### Defect 2 (secondary): the deployed bundle predates every timeout fix

`.mcp.json` launches `node mcp-loom/dist/index.js` directly. `dist/` is gitignored
(`.gitignore:7`), so it is a local artifact nothing rebuilds.

| Fact | Value |
|---|---|
| `dist/index.js` build mtime | 2026-07-25 17:19 |
| PR #3958 (bounded `sendDaemonRequest`, #3945) merged | 2026-07-25 17:20:47 — **88s later** |
| Latest `mcp-loom/src/` commit | `a9c5256d`, 2026-07-27 01:07 |
| `grep -c "did not respond within" dist/index.js` | **0** |
| Bundled signature | `async function sendDaemonRequest(request)` — one arg, no timer |

Every per-call timeout the source passes (30s dispatch / read-path / cancel) is
silently discarded as an extra argument to a one-parameter function. With Defect 1
guaranteeing the promise never settles and Defect 2 removing the only bound, the
wait runs to the client's ~1800s idle abort. **That is exactly the reported
figure.**

The staleness gate exists but is not in the launch path: `scripts/setup-mcp.sh:29-32`
has a correct `find -newer` check, but it is a one-shot `.mcp.json` generator.
The pre-flight that *is* in the path — `check_mcp_server()` at
`defaults/scripts/claude-wrapper.sh:364-421` — tests only **existence** (`:395`)
and **liveness** (`:405-408`), so a stale-but-startable bundle passes silently.
Its own comment at `:362` already claims it handles "missing or stale"; only the
missing half was implemented. `_try_mcp_rebuild()` (`:491-513`) already exists and
works — it is simply never invoked for staleness.

### Why this recurred three times

#3945 and #4001 both observed this same signature — "daemon alive, CLI instant,
MCP call hangs" — and both treated it with **timeouts**, which bound the symptom
without addressing the inability to resolve on success. This issue is the third
recurrence.

The verification harness cannot catch it: `mcp-loom/scripts/verify-daemon-timeout.mjs:155-161`
builds its happy-path stub as `sock.write(...); sock.end();` — **a server that
closes after responding, which the real daemon does not do.** The harness
therefore validates the one behavior the daemon never exhibits. Fixing that stub
is what stops recurrence #4.

### Implementation guidance

**Defect 1 — `mcp-loom/src/shared/daemon.ts`.** Settle on the first complete
newline-delimited frame instead of on `end`, mirroring the framing
`sendDaemonStreamRequest` (`:237-246`) and the Rust CLI (`main.rs:1477`) already
use correctly:

- In the `data` handler, append to the buffer, then check for `\n`. On the first
  complete line, `JSON.parse` it and resolve.
- Close the socket after settling (the settle guard + `socket.destroy()` at `:140`
  is already the established pattern) so the daemon's `next_line()` sees EOF and
  reaps the connection.
- Keep the `end` handler as a fallback for servers that *do* close (preserves the
  existing stub-server behavior and any close-after-respond peer).
- Keep the timeout as the safety net — it should now be a genuine error signal
  rather than the normal exit path.

This is one function. Do **not** change the daemon's framing: the persistent
connection is intentional and the Rust CLI already speaks it correctly.

**Defect 2 — `defaults/scripts/claude-wrapper.sh`.** Add a staleness branch to
`check_mcp_server()` before the smoke test, reusing the existing helper:

```bash
local mcp_src
mcp_src="$(dirname "$(dirname "${mcp_entry}")")/src"
if [[ -d "${mcp_src}" ]] && \
   [[ -n "$(find "${mcp_src}" -type f -newer "${mcp_entry}" -print -quit 2>/dev/null)" ]]; then
    log_warn "MCP bundle is stale (src newer than dist) - rebuilding"
    _try_mcp_rebuild "${mcp_entry}" || return $?
fi
```

Mirror `setup-mcp.sh:29-32` exactly so the gates cannot drift. Stay non-fatal on
rebuild failure when the bundle still starts. Fix the `:362` comment. Edit only
`defaults/scripts/claude-wrapper.sh` — `.loom/scripts/claude-wrapper.sh` is the
gitignored installed copy (currently byte-identical); refresh it via resync.

**Out of scope**: daemon-side IPC framing, the registry-mutex hold across
`reap_liveness` / `spawn_child` (a real but separate contention concern — bounded
at ~30s and demonstrably not active during this incident, since `DaemonStatus`
takes the same mutex at `ipc.rs:600` and returned in 6ms).

### Immediate operator remediation

A rebuild is **necessary but not sufficient** — it yields a fast, diagnosable
failure instead of a silent hang, and is the recommended state until Defect 1
lands:

```bash
cd mcp-loom && npm install && npm run build   # then restart the session
```

## Acceptance Criteria

- [ ] `sendDaemonRequest` resolves on the first newline-delimited response frame;
      a successful unary call against the **real** daemon returns in well under a
      second and does not depend on connection close.
- [ ] After settling, the client closes the socket so the daemon reaps the
      connection (no accumulating half-open connections under repeated calls).
- [ ] `mcp__loom__list_sweeps`, `get_sweep_status`, `dispatch_sweep`, and
      `cancel_sweep` all return promptly against a live daemon.
- [ ] `cancel_sweep` is verified specifically — per the original AC #4 it is the
      highest-severity instance, being the only way to stop a running sweep.
- [ ] The timeout still fires (with the greppable "did not respond within"
      message) against a server that accepts but never responds.
- [ ] `verify-daemon-timeout.mjs` gains a case whose stub **responds with a
      newline-delimited frame and does NOT close** — modeling the real daemon.
      This case must fail against the pre-fix transport.
- [ ] `check_mcp_server()` detects a bundle older than any file under
      `mcp-loom/src/` and rebuilds before the smoke test, with a greppable
      staleness warning; rebuild failure on a startable bundle warns and
      continues.
- [ ] The `claude-wrapper.sh:362` comment matches implemented behavior.
- [ ] No changes to `loom-daemon/src/` framing.

## Affected Files

- `mcp-loom/src/shared/daemon.ts` - **primary fix.** Settle `sendDaemonRequest`
  (`:102-180`) on the first `\n`-delimited frame in the `data` handler (`:144-146`)
  rather than on `end` (`:148-155`); close the socket after settling; retain the
  `end` fallback and the timeout.
- `mcp-loom/scripts/verify-daemon-timeout.mjs` - add a respond-but-never-close
  stub case alongside the existing hang and respond-then-`end` cases (`:150-168`);
  this is the regression guard that stops a fourth recurrence.
- `defaults/scripts/claude-wrapper.sh` - staleness branch in `check_mcp_server()`
  (`:394-421`), reusing `_try_mcp_rebuild()` (`:491-513`); correct the `:362`
  comment.
- `scripts/setup-mcp.sh` - reference only (`:29-32` is the predicate to mirror);
  no change expected.
- `.loom/docs/troubleshooting.md` - "MCP tools hang with no response" entry: the
  raw-socket probe, the `grep -c "did not respond within" mcp-loom/dist/index.js`
  staleness check, and the rebuild remedy.

> `.loom/scripts/claude-wrapper.sh` is a gitignored installed copy and is
> deliberately **not** listed — do not edit it directly.

## Test Plan

- [ ] Automated: `cd mcp-loom && npm run verify:daemon-timeout` — all three stub
      cases pass (hang → bounded rejection; respond-then-end → resolves;
      **respond-without-close → resolves**).
- [ ] Regression proof: the new stub case must **fail** when run against the
      pre-fix `daemon.ts`, confirming it actually covers the defect.
- [ ] Live end-to-end: with the daemon running, `mcp__loom__list_sweeps` and
      `mcp__loom__get_sweep_status` return in <1s.
- [ ] Live `cancel_sweep`: dispatch a throwaway sweep, cancel it via MCP, confirm
      a prompt response and that the registry entry transitions to `Exited`.
- [ ] Connection hygiene: issue ~50 sequential MCP calls, then confirm the daemon
      is not accumulating open connections (e.g. `lsof -U | grep loom-daemon.sock`).
- [ ] Manual staleness: `touch mcp-loom/src/index.ts`, run the wrapper pre-flight,
      confirm the staleness warning + rebuild.
- [ ] No false positive: re-run the pre-flight immediately on a fresh build and
      confirm **no** rebuild (guards against a rebuild loop every launch).
- [ ] Missing artifact: `rm mcp-loom/dist/index.js`, confirm the existing
      missing-artifact path still rebuilds.
- [ ] Edge case: `mcp-loom/src/` absent (consumer install) — predicate is a no-op.

---

## Curator Addendum (2026-07-27)

Root-cause forensics, implementation guidance, Affected Files, and Test Plan are in the [curator comment](https://github.com/rjwalters/loom/issues/4043#issuecomment-5097188878). Key verified finding: the configured MCP server bundle (`mcp-loom/dist/`, referenced by `.mcp.json`) is a stale **2026-01-25 v0.3.0** build containing neither the sweep tools nor any IPC timeout logic; current `src/` (post-#3958/#4001) bounds every call at ≤30s and cannot produce this hang. The investigation must identify which build actually served the incident session and structurally prevent stale bundles from serving future sessions.

rjwalters:
## Summary

Fixes the ~1800s hang where `mcp__loom__dispatch_sweep` / `get_sweep_status` (and the other unary MCP tools) returned no response even though the daemon operation succeeded. The MCP bridge's unary transport (`sendDaemonRequest`) settled its promise **only** in the socket's `end` handler, but the real `loom-daemon` holds each connection open by design (a persistent per-connection read loop, `loom-daemon/src/ipc.rs` `handle_client`) and never closes after writing its one newline-delimited response frame. The response therefore sat complete-but-unparsed in the client buffer while the promise waited forever for an `end` that never came. A stale `dist/` bundle compounded it by discarding the per-call timeout, extending the failure to the client's full ~1800s idle abort.

## Changes

- **`mcp-loom/src/shared/daemon.ts`** (primary fix): `sendDaemonRequest` now settles on the **first complete `\n`-delimited frame** in the `data` handler — mirroring `sendDaemonStreamRequest` and the Rust CLI (`main.rs:1477`) — then closes the socket so the daemon's `next_line()` reaps the connection (no accumulating half-open connections). The `end` handler is retained as a close-after-respond fallback; the timeout remains as a genuine error-signal safety net. No daemon-side framing changes.
- **`mcp-loom/scripts/verify-daemon-timeout.mjs`**: adds a third stub case whose server responds with a newline-delimited frame and **does NOT close** (modeling the real daemon). Verified to **fail** against the pre-fix transport (hangs to the 2000ms bound) and **pass** after — the regression guard against a fourth recurrence.
- **`defaults/scripts/claude-wrapper.sh`**: `check_mcp_server()` gains a staleness branch before the smoke test that rebuilds when `dist/index.js` is older than any file under the sibling `src/` (mirroring `scripts/setup-mcp.sh:29-32`'s predicate exactly), invoking the existing `_try_mcp_rebuild()`. Non-fatal if rebuild fails but the bundle still starts. The `check_mcp_server` header comment is corrected to match. `.loom/scripts/claude-wrapper.sh` (gitignored installed copy) is deliberately untouched.
- **`.loom/docs/troubleshooting.md`**: new "MCP tools hang with no response (~1800s)" entry with the raw-socket / `grep -c "did not respond within"` diagnosis and the rebuild remedy.

- [x] `sendDaemonRequest` resolves on the first newline-delimited frame; does not depend on connection close (verified by the new respond-without-close harness case resolving in <1s).
- [x] Client closes the socket after settling (`socket.destroy()` in `settleResolve`).
- [x] Timeout still fires with the greppable "did not respond within" message against a never-responding server (existing hang cases pass).
- [x] `verify-daemon-timeout.mjs` gains a respond-without-close case that **fails** against the pre-fix transport (proven by stashing the fix: 3 checks failed) and passes after.
- [x] `check_mcp_server()` detects a bundle older than any file under `src/` and rebuilds before the smoke test with a greppable warning; rebuild failure on a startable bundle warns and continues (staleness predicate tested in isolation: fresh=no-op, stale=rebuild, no-src=no-op).
- [x] The `claude-wrapper.sh` comment matches implemented behavior.
- [x] No changes to `loom-daemon/src/` framing.

## Test Plan

- [x] `cd mcp-loom && npm run build` (tsc --noEmit + esbuild) passes.
- [x] `npm run verify:daemon-timeout` — all cases pass (hang → bounded rejection; respond-then-end → resolves; **respond-without-close → resolves**).
- [x] Regression proof: stashing the `daemon.ts` fix makes the new case fail (3 checks failed, hangs to the 2000ms bound), confirming it covers the defect.
- [x] `bash -n defaults/scripts/claude-wrapper.sh` passes; staleness predicate verified in isolation (fresh / stale / src-absent).

Live end-to-end checks (dispatch/cancel against a running daemon, connection-hygiene `lsof` sweep) require a live daemon + token pool and are left for the Judge/operator environment.

Closes #4043

### DIFF
--- a/.loom/docs/troubleshooting.md
+++ b/.loom/docs/troubleshooting.md
@@ -281,6 +281,30 @@ grep -c dispatch_sweep dist/index.js   # should now be > 0
 
 `scripts/setup-mcp.sh` now auto-rebuilds when `dist/index.js` is missing **or** older than any file under `mcp-loom/src/` (#3803), so `./scripts/setup-mcp.sh` is the safe one-shot path. Rebuilding the bundle does **not** refresh an already-running session — an MCP client caches its tool list at connect time, so you must **restart the Claude Code session** (or respawn the `loom` MCP subprocess) for the new tools to appear. See [`mcp-loom/README.md`](../../mcp-loom/README.md#rebuilding-after-source-changes-reconnect-required) for the full rebuild + reconnect procedure and a raw `tools/list` verification snippet.
 
+### MCP tools hang with no response (~1800s), then abort
+
+**Symptom**: `mcp__loom__dispatch_sweep`, `mcp__loom__get_sweep_status`, `mcp__loom__list_sweeps`, or `mcp__loom__cancel_sweep` return **no response and no progress**, and are eventually aborted by the client (`sent no response or progress for 1800s; aborting`) — even though the underlying operation **succeeded** (the sweep child spawned, the PR opened, etc.). The CLI path (`loom-daemon status` / `dispatch`) stays fast throughout, which isolates the fault to the MCP/IPC response path, not a wedged daemon.
+
+**Cause** (#4043): the MCP bridge's unary request transport (`mcp-loom/src/shared/daemon.ts` `sendDaemonRequest`) historically settled its promise only in the socket's `end` handler. The real `loom-daemon` holds each connection **open by design** (a persistent per-connection read loop, `loom-daemon/src/ipc.rs` `handle_client`): it writes one newline-delimited JSON response frame per request and never closes after answering. So the response sat complete-but-unparsed in the client buffer while the promise waited forever for an `end` that never came. A **stale bundle** (`mcp-loom/dist/` older than `mcp-loom/src/`) compounds it by discarding the per-call timeout, turning a diagnosable failure into the full ~1800s idle hang.
+
+**Diagnose** — probe the raw socket (bypassing the MCP layer) and check the bundle for the timeout string:
+
+```bash
+# Does the bundle even carry the bounded-timeout fix? 0 => stale, pre-timeout bundle
+grep -c "did not respond within" mcp-loom/dist/index.js
+
+# Is dist/ older than src/? (any output => stale, rebuild needed)
+find mcp-loom/src -type f -newer mcp-loom/dist/index.js
+```
+
+**Fix** — the transport now settles on the **first newline-delimited response frame** (in the `data` handler) and closes the socket after settling, so it no longer depends on the daemon closing the connection. If you are on a pre-fix bundle, rebuild and reconnect:
+
+```bash
+cd mcp-loom && npm install && npm run build   # then restart the Claude Code session
+```
+
+The `claude-wrapper.sh` MCP pre-flight (`check_mcp_server`) now rebuilds a stale bundle before the smoke test, so a fresh session on an up-to-date checkout self-heals; the regression is guarded by `mcp-loom/scripts/verify-daemon-timeout.mjs`'s respond-without-close stub case (`npm run verify:daemon-timeout`).
+
 ### Inspect running sweeps
 
 ```bash
--- a/defaults/scripts/claude-wrapper.sh
+++ b/defaults/scripts/claude-wrapper.sh
@@ -359,8 +359,9 @@ resolve_mcp_workspace() {
 
 # Pre-flight check: verify MCP server can start
 # Attempts to launch the mcp-loom Node.js server and checks for the startup
-# message on stderr. If the dist/ directory is missing or stale, attempts
-# a rebuild before retrying.
+# message on stderr. If the built bundle is missing, or is stale (older than
+# any file under the sibling src/ tree), it rebuilds before the smoke test; a
+# smoke-test failure also triggers a rebuild-and-retry.
 check_mcp_server() {
     local mcp_workspace
     mcp_workspace=$(resolve_mcp_workspace)
@@ -396,6 +397,25 @@ for name, srv in servers.items():
         log_warn "MCP entry point missing: ${mcp_entry}"
         _try_mcp_rebuild "${mcp_entry}"
         return $?
+    fi
+
+    # Staleness check: a bundle older than any file under the sibling src/ tree
+    # is startable but lacks recent fixes (a smoke test passes on it silently).
+    # Rebuild before the smoke test. This mirrors the predicate in
+    # scripts/setup-mcp.sh (the one-shot .mcp.json generator) exactly so the two
+    # gates cannot drift (issue #4043). Rebuild failure on an otherwise-startable
+    # bundle is non-fatal — fall through to the smoke test below.
+    local mcp_src
+    mcp_src="$(dirname "$(dirname "${mcp_entry}")")/src"
+    if [[ -d "${mcp_src}" ]] && \
+       [[ -n "$(find "${mcp_src}" -type f -newer "${mcp_entry}" -print -quit 2>/dev/null)" ]]; then
+        log_warn "MCP bundle is stale (src newer than dist) - rebuilding: ${mcp_entry}"
+        if ! _try_mcp_rebuild "${mcp_entry}"; then
+            log_warn "MCP rebuild for stale bundle failed - continuing with existing bundle"
+        else
+            # Rebuild succeeded and already re-verified the smoke test.
+            return 0
+        fi
     fi
 
     # Smoke test: start MCP server and verify it emits the startup message
--- a/mcp-loom/scripts/verify-daemon-timeout.mjs
+++ b/mcp-loom/scripts/verify-daemon-timeout.mjs
@@ -12,6 +12,11 @@
  *   2. Happy case — server writes a JSON response frame then ends. Assert the
  *      call RESOLVES quickly and the timer was cleared (no regression to the
  *      fast path used by dispatch_sweep / list_sweeps / publish_event).
+ *   3. Real-daemon case — server writes a newline-delimited JSON response frame
+ *      and then KEEPS THE CONNECTION OPEN (never calls end/destroy), modeling
+ *      the real loom-daemon's persistent per-connection read loop. Assert the
+ *      call RESOLVES on the first frame without depending on connection close.
+ *      This case fails against a transport that settles only on `end` (#4043).
  *
  * Run: node scripts/verify-daemon-timeout.mjs   (from the mcp-loom dir)
  *
@@ -173,6 +178,49 @@ async function main() {
   check("resolved with expected payload", !!okResp && okResp.type === "EventPublished");
   check("resolved fast (timer cleared, no hang)", okElapsed < 1000, `elapsed=${okElapsed}ms`);
   await closeServer(okServer);
+
+  // ---- Case 3: real-daemon path (respond with a frame, never close) ----
+  // Models loom-daemon/src/ipc.rs `handle_client`: one `\n`-delimited response
+  // frame per request line, then the connection stays open for reuse. This is
+  // the regression guard for #4043 — a transport that settles only on the
+  // socket's `end` event hangs here (the server never ends) and this case
+  // FAILS, exactly reproducing the reported 1800s hang.
+  console.log("real-daemon path (respond then keep open)");
+  const heldReplySockets = [];
+  const persistentServer = await startStubServer(sockPath, (sock) => {
+    heldReplySockets.push(sock);
+    sock.on("data", () => {
+      // Answer with a newline-delimited frame, then deliberately do NOT
+      // end/destroy — the real daemon holds the connection open.
+      sock.write(JSON.stringify({ type: "SweepList", payload: { sweeps: [] } }));
+      sock.write("\n");
+    });
+  });
+
+  const persistentStart = Date.now();
+  let persistentResp = null;
+  let persistentErr = null;
+  try {
+    // A generous timeout so a FAILURE here is a genuine hang (settle-on-end
+    // regression), not a too-tight bound.
+    persistentResp = await sendDaemonRequest({ type: "ListSweeps", payload: { state_filter: null } }, 2000);
+  } catch (e) {
+    persistentErr = e;
+  }
+  const persistentElapsed = Date.now() - persistentStart;
+  check("resolved without waiting for close", persistentErr === null, persistentErr && persistentErr.message);
+  check(
+    "resolved with expected payload (first frame)",
+    !!persistentResp && persistentResp.type === "SweepList",
+    persistentResp && JSON.stringify(persistentResp)
+  );
+  check(
+    "resolved fast against a never-closing peer",
+    persistentElapsed < 1000,
+    `elapsed=${persistentElapsed}ms`
+  );
+  for (const s of heldReplySockets) s.destroy();
+  await closeServer(persistentServer);
 
   // Give the process a beat to confirm no lingering timer keeps it alive
   // (a leaked setTimeout would delay exit by the full timeout window).
--- a/mcp-loom/src/shared/daemon.ts
+++ b/mcp-loom/src/shared/daemon.ts
@@ -131,6 +131,12 @@ export async function sendDaemonRequest(
       if (settled) return;
       settled = true;
       clearTimer();
+      // Close the socket once we have our answer. The real daemon holds the
+      // connection open by design (a persistent per-connection read loop, see
+      // loom-daemon/src/ipc.rs `handle_client`), so we must send EOF ourselves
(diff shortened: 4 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed the diff carefully against all three claimed defects and verified each with independent testing (not just reading the PR description).

## 1. Primary fix — `mcp-loom/src/shared/daemon.ts`

Confirmed correct. `sendDaemonRequest` now settles on the first `\n`-delimited frame in the `data` handler (`buffer.indexOf("\n")`, slice + `JSON.parse`), then `socket.destroy()`s once settled. The `end` handler is retained as a fallback (guarded by `if (settled) return`), and the timeout remains the safety net. This correctly mirrors `sendDaemonStreamRequest`'s existing newline-framing pattern in the same file (lines 263-272), and matches the daemon's documented per-connection persistent read loop (`loom-daemon/src/ipc.rs handle_client`).

- Handles partial reads correctly: `buffer +=` accumulates across `data` events until a full line is available; a frame split across TCP packets is handled by the existing accumulation loop.
- Handles the extra-data-after-frame case correctly for unary use: only the first line is parsed; the socket is destroyed immediately after, so any trailing buffered bytes are simply discarded — appropriate for a strictly request/response transport.
- Double-settle is guarded (`settled` flag checked in both `data` and `end` handlers, and in `settleResolve`/`settleReject`).

## 2. Regression test — `verify-daemon-timeout.mjs`

I did not just trust the "verified to fail pre-fix" claim in the PR body — I reproduced it independently:

```
cd mcp-loom
git show ebf4fc55:mcp-loom/src/shared/daemon.ts > src/shared/daemon.ts  # pre-fix version
npm run build && npm run verify:daemon-timeout
# → 3 check(s) failed, "real-daemon path" case hangs to the 2000ms bound exactly as claimed
```

Restored the fix and re-ran — all checks pass, including the new "real-daemon path (respond then keep open)" case. This is a genuine regression guard: the new stub server responds with a newline-delimited frame and deliberately never closes the connection, which is exactly the shape that broke on the old `end`-only transport.

## 3. `claude-wrapper.sh` staleness detection

- `bash -n defaults/scripts/claude-wrapper.sh` passes.
- The new staleness predicate (`find "${mcp_src}" -type f -newer "${mcp_entry}" -print -quit`) is byte-for-byte the same test as `scripts/setup-mcp.sh:29` (verified side-by-side), so the two gates can't drift.
- `_try_mcp_rebuild()` already re-verifies with its own smoke test after rebuilding (lines 540-546), so the new branch's `return 0` on success is correct — it's not skipping verification, just reusing the rebuild helper's own post-rebuild check.
- The pre-existing missing-artifact path (`[[ ! -f "${mcp_entry}" ]]`) and the smoke-test-failure retry path are both untouched and still correct.
- Confirmed `.loom/scripts/claude-wrapper.sh` (gitignored installed copy) was **not** touched — `git diff` against it is empty.

## Build / test execution

- `cd mcp-loom && npm run build` — passes (tsc --noEmit + esbuild).
- `npm run verify:daemon-timeout` — all cases pass, including the new one.
- Regression proof reproduced manually (see above) — fails pre-fix, passes post-fix.
- `bash -n defaults/scripts/claude-wrapper.sh` — passes.

## CI / merge state

- `gh pr checks 4104`: all applicable checks pass (Installer Integration Tests, MCP Server Build, Shellcheck, CLAUDE.md Line Budget, Detect Changes); Rust checks correctly `skipping` since no Rust files changed.
- `mergeStateStatus`: CLEAN.
- PR body: "Closes #4043" present.

Solid, well-scoped fix with a genuine regression test that I independently confirmed catches the defect. No blocking issues found.

Issue #3287, opened 2026-05-14
Installer (daemon `init` step) silently skips .loom/scripts/lib/*.sh despite recording them in install-metadata.json


rjwalters:
## Symptom

After running `install-loom.sh` at commit `b52dee49` against a clean target repository (bucket-brigade, install date 2026-05-14), the entire `.loom/scripts/lib/` directory is missing on disk, even though `.loom/install-metadata.json` lists all four files as installed:

- `.loom/scripts/lib/classify-error.sh`
- `.loom/scripts/lib/forge-helpers.sh`
- `.loom/scripts/lib/loom-tools.sh`
- `.loom/scripts/lib/pipe-pane-cmd.sh`

Downstream effect:

```
$ .loom/scripts/merge-pr.sh 130 --dry-run
.loom/scripts/merge-pr.sh: line 69: …/.loom/scripts/lib/forge-helpers.sh: No such file or directory
```

`.loom/scripts/checkpoint.sh` and ~17 other helper scripts that `source` `forge-helpers.sh` or `loom-tools.sh` also fail in worktrees.

## Why the existing verification didn't catch it

`scripts/install-loom.sh` lines 832–887 has both an explicit per-file verification and a recursive walk that hard-errors on missing `defaults/scripts/lib/` files (with a comment referencing prior regression #3220). On this install, none of those checks fired — install completed with success and `install-metadata.json` was written claiming the files were installed.

Hypothesis: the actual file copy is delegated to `target/release/loom-daemon init … --defaults …/defaults` (install-loom.sh:705). The daemon's `init` subcommand may not recursively traverse `defaults/scripts/lib/`, OR our local install ran against a daemon binary built before the verification block was added. Either way, the shell-level verification in `install-loom.sh` was bypassed/skipped on this run.

## Repro hint

1. Build the daemon (or use whatever binary is on `$PATH`).
2. Run `./scripts/install-loom.sh /path/to/clean-repo` non-interactively.
3. Check `ls /path/to/clean-repo/.loom/scripts/lib/` — directory absent.
4. Check `cat /path/to/clean-repo/.loom/install-metadata.json | jq '.files[] | select(contains("lib/"))'` — files listed.

## Local workaround applied downstream

```
mkdir -p .loom/scripts/lib
cp $LOOM_ROOT/.loom/scripts/lib/*.sh .loom/scripts/lib/
git add .loom/scripts/lib/*.sh && git commit
```

## Suggested upstream fix

[...]

PR #4107, opened 2026-07-27
feat(hooks): defer generic guard-destructive to Repo Skills' canonical guard (#4041)

### LINKED ISSUE #4041: Drop the generic guard-destructive pattern list; defer to Repo Skills' canonical guard, keep Loom-specific hooks
## Context

Loom's `guard-destructive.sh` (installed as `.loom/hooks/guard-destructive.sh`) embeds a generic destructive-command pattern list that is a near-copy of Repo Skills' guard. In consumer repos with both tools installed (e.g. rjwalters/safehouse), both guards run back-to-back on the same PreToolUse→Bash matcher: duplicate execution, duplicate false-positive surface, and pattern fixes must land twice (#4040 and rjwalters/repo#29 are the live example).

Proposal (agreed direction with rjwalters/repo#30): **the generic guard's canonical home is Repo Skills; Loom defers to it and keeps only Loom-specific enforcement.** The generic guard has no orchestration dependency and belongs with the general-by-design tool that is installed in many non-Loom repos; Loom keeps what is genuinely Loom's: `guard-loom-workflow.sh`, worktree rm-scope enforcement (`LOOM_RM_SCOPE`), methodology-inject, skill-router.

## Suggested acceptance criteria (Loom side)

- [ ] Split Loom's `guard-destructive.sh`: extract the Loom-specific sections (workflow/lifecycle guard, rm-scope, telemetry wrapper) into Loom's own hooks; drop the generic pattern list
- [ ] Installer behavior in the target repo:
  - Repo Skills guard present (>= the version carrying rjwalters/repo#29's fix): wire only Loom-specific hooks; do not install a second generic guard
  - Repo Skills absent: either declare it a prerequisite, or install a clearly-marked vendored copy of the canonical guard (sourced from Repo Skills, not forked) so standalone-Loom repos keep protection
- [ ] Migrate Loom's guard test cases upstream to Repo Skills (rjwalters/repo#30) so the canonical copy inherits Loom's battle-testing
- [ ] Existing dual-install consumer repos converge on next `install.sh` run: the settings PreToolUse chain ends up with one generic guard + Loom-specific hooks, not two generic guards

## Non-goals

- No behavior change for what gets blocked (modulo the #4040/repo#29 false-positive fix, which lands in the canonical copy)
- Standalone-Loom repos must not lose guard coverage

---

### Verification findings — upstream is further along than the body assumes

Verified against rjwalters/repo HEAD (v0.6.1 era) on 2026-07-27:

1. **The upstream consolidation has already landed.** rjwalters/repo#29 (the curl-pipe false-positive fix) closed as **completed** 2026-07-27T17:40Z, and rjwalters/repo#30 (canonical home) is closed. The canonical guard lives at `hooks/repo/guard-destructive.sh` in rjwalters/repo (1863 lines) and installs into consumer repos at `.claude/skills/repo/hooks/guard-destructive.sh`. The fixed pattern anchors the pipe target on command position (annotated with `repo#29` in a comment near it — a usable presence/version probe).
2. **The canonical guard is a strict superset of Loom's config surface.** All 8 Loom toggles exist upstream — `readOnlyFastPath`, `readOnlyFastPathExtra`, `sqlDdl`, `cloudCli`, `reversibleGh`, `decisionLog`, `rmScope`, `forceScope` — with `REPO_GUARD_*` env names preferred and the `LOOM_GUARD_*` / `LOOM_RM_SCOPE` names honored as legacy fallbacks. Config resolution reads `.claude/skills/repo/config.json` first, then falls back to `.loom/config.json`. The "no behavior change" non-goal is therefore mostly satisfied by upstream design already; the Loom-side work is subtraction plus installer logic.
3. **Repo Skills' installer is already coexistence-aware — and the deferral currently points the wrong way.** Its `install.sh` detects an existing `guard-destructive.sh` wired under a Bash matcher and declines to add its own. Consequence: in dual-install repos today, **Loom's older copy (with the broken pattern) wins** and the fixed canonical guard never fires. This makes #4041 the blocking step for the fix actually reaching dual-install consumers.
4. **The "extract the workflow guard" AC item is already done.** `guard-loom-workflow.sh` is a separate file with its own settings entry; `guard-worktree-paths.sh` covers the Edit and Write matcher. The realistic Loom-side split is therefore: stop installing and wiring `guard-destructive.sh` when the canonical guard is present — not a line-by-line split of the 1729-line file.
5. **Upstream already has a migrated test suite** at `hooks/repo/tests/test-guard-destructive.sh`. The "migrate Loom's test cases upstream" AC item may be partially complete; the builder should diff Loom's `tests/hooks/test-guard-destructive.sh` (2055 lines) against it and upstream only the missing cases (that part is a rjwalters/repo PR, outside this repo — see scope note below).

### Implementation guidance

**Detection**: at install time, treat the canonical guard as present when `.claude/skills/repo/hooks/guard-destructive.sh` exists in the target repo and contains the `repo#29` marker comment (presence of the fix, without version arithmetic). Anything older is treated as absent.

**Installer branch** (`install.sh` → `install_hooks_and_cli()`, lines 55–80; call sites at 827 and 1200):
- Canonical guard present: install only Loom-specific hooks (`guard-loom-workflow.sh`, `guard-worktree-paths.sh`); do not install `.loom/hooks/guard-destructive.sh`; wire settings so the canonical guard plus the Loom-specific hooks form the PreToolUse chain.
- Canonical guard absent (standalone-Oom repo): keep installing a generic guard so coverage is not lost. Prefer a clearly-marked vendored copy of the canonical file (sourced from Repo Skills at Loom release time, not forked) per the issue body's option (b) — it avoids adding a hard install-order dependency between the two tools.

**Convergence-ordering risk (flag for the builder)**: because Repo Skills defers when it sees an existing guard, a dual-install repo where Loom simply *removes* its guard entry can end up with **no generic guard** until Repo Skills' installer re-runs. The Loom installer must not leave that window: either rewire the settings entry to the canonical path itself (it already knows the path from detection), or refuse to drop its copy until the canonical one is wired. The AC "converge on next `install.sh` run" should be read as: after Loom's run alone, exactly one generic guard is wired — never zero.

**Update semantics**: `install_hooks_and_cli()` preserves existing `.loom/hooks/<name>` unless force-overwrite (`--clean`); `defaults/scripts/resync-installed.sh` refreshes installed copies after a `git pull`. Both need to handle the "this hook is intentionally no longer installed" case (remove or stop resyncing `guard-destructive.sh` when the canonical branch is taken), or dual-install repos will keep resurrecting the stale copy.

**Superset check before deletion**: the toggle surface is verified upstream (finding 2), but the builder should also diff the two `ALWAYS_BLOCK_PATTERNS` and ask-tier pattern lists (Loom's at `defaults/hooks/guard-destructive.sh:1036` onward) against the canonical copy to confirm no Loom-only pattern is silently dropped. Any Loom-only generic pattern found goes upstream (rjwalters/repo PR), not into a retained fork.

### Affected Files

- `defaults/hooks/guard-destructive.sh` - the generic guard this issue removes from Loom's install set (1729 lines; generic block list starts at line 1036, broken curl patterns at 1069–1070)
- `.loom/hooks/guard-destructive.sh` - this repo's own installed copy; converges via the same installer/resync change
- `defaults/hooks/guard-loom-workflow.sh` - retained Loom-specific hook (no content change expected; referenced by the new installer branch)
- `defaults/.claude/settings.json` - PreToolUse Bash chain currently wires guard-destructive then guard-loom-workflow; becomes conditional on the canonical guard
- `install.sh` - `install_hooks_and_cli()` (lines 55–80) plus call sites at 827 and 1200: add canonical-guard detection and the two install branches
- `defaults/scripts/resync-installed.sh` - stop resyncing the dropped hook when the canonical branch is taken
- `scripts/install-loom.sh` - remote-install path references guard-destructive; same conditional
- `tests/hooks/test-guard-destructive.sh` - trim to vendored-copy smoke tests or repoint; upstream any cases missing from rjwalters/repo `hooks/repo/tests/test-guard-destructive.sh`
- `tests/install/test-hooks-preserve.sh` - installer preserve semantics change for the removed hook
- `defaults/scripts/tests/test-guard-hook-schema.sh` - hook schema test must tolerate the conditional install set
- `defaults/CLAUDE.md` - "Custom Guard Hooks" catalog: document that the generic guard's canonical home is Repo Skills
- `.loom/docs/troubleshooting.md` - update guard-destructive references

### Test Plan

- [ ] Manual verification (dual-install): in a scratch repo with current Repo Skills installed, run Loom's `install.sh`; assert `.claude/settings.json` wires exactly one generic guard (the `.claude/skills/repo/hooks` copy) plus `guard-loom-workflow.sh` and `guard-worktree-paths.sh`, and that `.loom/hooks/guard-destructive.sh` is absent
- [ ] Manual verification (behavior): in that repo, confirm a download piped to `sudo tee` under `/usr/share/` is allowed and a download piped to a shell is still blocked (the repo#29 regression pair)
- [ ] Manual verification (standalone): in a scratch repo without Repo Skills, run Loom's `install.sh`; assert generic guard coverage still exists (vendored copy installed and wired)
- [ ] Manual verification (convergence): start from a dual-install repo in today's state (Loom guard wired, Repo Skills deferred); re-run Loom's `install.sh`; assert the chain ends with one generic guard and never passes through a zero-generic-guard state
- [ ] Automated tests: update `tests/install/test-hooks-preserve.sh` and `defaults/scripts/tests/test-guard-hook-schema.sh` for the conditional install set; add an installer test for the detection branch
- [ ] Edge cases: `--clean` force-overwrite path; a repo with a hand-customized `.loom/hooks/guard-destructive.sh` (installer preserves customizations — decide and test what happens to a customized copy when the canonical branch would remove it)

### Scope note

The Loom-side changes above are one PR in this repo. Upstreaming missing test cases or patterns is a separate rjwalters/repo PR and should not gate this issue beyond noting what was sent upstream. Closing #4040 is the acceptance signal that the consolidation reached the broken pattern.

### Dependencies

- [x] rjwalters/repo#29 - canonical curl-pipe fix (closed completed 2026-07-27)
- [x] rjwalters/repo#30 - canonical-home consolidation, upstream side (closed)

All upstream prerequisites are satisfied; this issue is unblocked.

rjwalters:
Closes #4041

## What & why

Loom's `guard-destructive.sh` embedded a generic destructive-command pattern list that now has a canonical, more-correct home upstream in Repo Skills (`.claude/skills/repo/hooks/guard-destructive.sh`, carrying the `rjwalters/repo#29` curl-pipe fix). In dual-install repos both guards ran back-to-back and pattern fixes had to land twice; today Loom's older copy even wins over the fixed canonical one.

## Approach — a runtime dispatcher (not a settings rewrite)

`defaults/hooks/guard-destructive.sh` becomes a thin dispatcher that decides at runtime which generic guard to run:

1. the canonical Repo Skills guard when it is present AND carries the `repo#29` marker (a presence/version probe, no semver math); else
2. the vendored generic guard (`guard-destructive-generic.sh`, Loom's current guard verbatim) so standalone-Loom repos keep full coverage.

Exactly one generic guard ever runs; never zero.

Why a dispatcher instead of rewriting `settings.json`'s Bash chain to the canonical path: consumer `settings.json` is produced by the Rust daemon's settings merge, which recognizes Loom hooks only by the `${CLAUDE_PROJECT_DIR}/.loom/hooks/` command prefix and dedups by exact command string. Pointing the entry at `.claude/skills/repo/hooks/...` (or a `bash -c` wrapper) would break Loom-hook ownership detection + uninstall and duplicate the entry on upgrade. Keeping the settings entry unchanged and making the canonical-vs-vendored choice inside the file preserves ownership/merge/uninstall and structurally closes the convergence-ordering window: Repo Skills' installer defers when it sees a wired `guard-destructive.sh`, and here that wired file is the dispatcher that runs the canonical guard — so there is never a zero-generic-guard state. When Repo Skills is later upgraded to the fixed version, the same dispatcher prefers the canonical guard at runtime with no reinstall.

## Changes

- `defaults/hooks/guard-destructive.sh` -> dispatcher (fail-open; preserves the read-only fast path in repos without Repo Skills via a zero-fork `[[ -r ]]` guard before the marker grep).
- `defaults/hooks/guard-destructive-generic.sh` (new) -> clearly-marked vendored copy of the canonical guard (Loom's current guard verbatim).
- `install.sh` + `scripts/install-loom.sh` -> detect the canonical guard; when present, do NOT install the vendored generic and remove any stale copy; when absent, install it. Same marker probe as the dispatcher, so install-time and runtime always agree.
- `scripts/install/manifest.sh` -> skip the vendored generic in the manifest when canonical is present, so the #3287 metadata-reconciliation hard-check does not fail on the intentionally-skipped file (and uninstall doesn't chase it).
- `defaults/scripts/resync-installed.sh` -> stop resurrecting the generic and remove a stale copy in canonical-present repos.
- Docs -> `defaults/CLAUDE.md` "Custom Guard Hooks" + `.loom/docs/troubleshooting.md`.
- Dogfood -> this repo's `.loom/hooks/` converge (its Repo Skills copy predates `repo#29`, so it keeps the vendored generic — no behavior change here).

## Superset / Loom-only pattern check

Diffed Loom's `ALWAYS_BLOCK_PATTERNS` / `ASK_PATTERNS` against the locally-installed canonical guard. The generic pattern set is fully generic hygiene (kubectl / sky / systemctl / printenv-secrets / cat-ssh-creds / git clean-checkout-restore / gh release delete, plus the cloud/SQL/force/rm toggle machinery). No Loom-only or Loom-workflow-specific pattern exists in the generic guard — those already live in `guard-loom-workflow.sh`. Because the generic is vendored verbatim (not deleted), no pattern is dropped for standalone repos. Nothing needs to be upstreamed to rjwalters/repo.

## Tests (all green)

- `tests/hooks/test-guard-destructive.sh` — 407-case pattern suite, repointed at the vendored generic.
- `tests/hooks/test-guard-destructive-dispatcher.sh` (new) — 7 cases: canonical marker present/absent/none -> correct guard, benign allow, fail-open.
- `tests/install/test-hooks-preserve.sh` — extended to 9 cases incl. canonical-present (generic skipped + stale removed) and canonical-absent (generic installed).
- `defaults/scripts/tests/test-guard-hook-schema.sh` — repointed fallback-source check at the generic; hardened the ask test against inherited `LOOM_FORCE_SCOPE`.
- `defaults/scripts/tests/test-resync-installed.sh` — 16/16 unchanged. shellcheck clean on new/modified files.

## Non-goals honored

No change to what gets blocked (beyond the already-upstream repo#29 fix, which lands via the canonical guard); standalone-Loom repos keep a generic guard.


rjwalters:
## Judge review — approve

Reviewed the diff, checked out the branch, and independently verified the claims in the PR description rather than trusting them at face value.

**Dispatcher (`defaults/hooks/guard-destructive.sh`)**
- `trap 'exit 0' ERR` + a final `exit 0` after both guard checks give true fail-open behavior; confirmed with the new dispatcher test's "no guard available" case (exit 0, empty output).
- Marker probe is a cheap `[[ -r ]]` builtin (zero forks) before the `grep -q 'repo#29'`, preserving the #3687 read-only fast path for repos without Repo Skills, as claimed.
- Both exec'd paths (`CANONICAL_GUARD`, `VENDORED_GUARD`) are fixed, non-externally-influenced paths — no injection surface. `exec bash "$path"` passes stdin through unchanged (verified with the dispatcher test).
- Confirmed the `${CLAUDE_PROJECT_DIR}/.loom/hooks/...` settings.json entry is genuinely untouched by this PR (`grep` on `defaults/.claude/settings.json`), and confirmed in `loom-daemon/src/init/scaffolding.rs` that `LOOM_HOOK_PREFIX = "${CLAUDE_PROJECT_DIR}/.loom/hooks/"` is exactly the string the Rust merge logic keys ownership/dedup detection on — so the "don't rewrite settings.json" design call is correct, not just asserted.

**Installer branches (`install.sh`, `scripts/install-loom.sh`, `manifest.sh`, `resync-installed.sh`)**
- Same marker-probe logic (`canonical_guard_present` / inline equivalents) is duplicated consistently across all four surfaces (dispatcher, quick installer, full installer, resync) — install-time and runtime agree.
- Both canonical-present and canonical-absent branches were exercised via the new `tests/install/test-hooks-preserve.sh` cases (5 new: dispatcher-installed + generic-skipped, stale-generic-removed, generic-installed-when-absent) — these call the *actual* `install_hooks_and_cli`/`canonical_guard_present` functions extracted live from `install.sh`, not reimplementations. No zero-guard window: the dispatcher itself is always installed regardless of branch.
- `manifest.sh`'s `TARGET_PATH` env threading checked end-to-end (`install.sh` exports it when sourcing/calling `_emit_installed_files_manifest`; `install-loom.sh` already has it as a global by call time) — the vendored-generic skip in the manifest is correctly gated and won't trip the #3287 reconciliation hard-check.

**Vendored copy fidelity**
- Diffed `defaults/hooks/guard-destructive-generic.sh` against the pre-PR `guard-destructive.sh` on `main`: identical except for the header comment block. No pattern logic was dropped or altered.

**Dogfood convergence**
- `.loom/hooks/{guard-destructive.sh,guard-destructive-generic.sh}` are byte-identical to their `defaults/` counterparts.
- Verified locally that this repo's actual installed `.claude/skills/repo/hooks/guard-destructive.sh` does *not* carry the `repo#29` marker, confirming the PR's claim that dogfood convergence here is a no-op (vendored generic keeps running, no behavior change).

**Tests**
- Ran all five test files directly: `tests/hooks/test-guard-destructive-dispatcher.sh` (7/7), `tests/install/test-hooks-preserve.sh` (9/9), `tests/hooks/test-guard-destructive.sh` (407/407 — see note below), `defaults/scripts/tests/test-guard-hook-schema.sh` (11/11), `defaults/scripts/tests/test-resync-installed.sh` (16/16). All pass cleanly.
  - Note: my first run of `test-guard-destructive.sh` showed 8 spurious failures, but that traced to `LOOM_FORCE_SCOPE`/`LOOM_GUARD_DECISION_LOG` already set in my own shell environment leaking into the test, not a bug in the PR — re-running with those unset gives a clean 407/407. Flagging this only in case it's worth hardening the test harness to `env -u` those vars defensively, not a blocker.
- Shellcheck clean on all new/changed lines; the few shellcheck notes I saw on `scripts/install-loom.sh` (SC2115/SC2086/SC2295/SC2001) all pre-date this PR (`git blame` confirms), matching the green "Shellcheck" CI check.

**CI / merge state**: all required checks green (Installer Integration Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes); `mergeStateStatus: CLEAN`.

No changes requested. Nice, careful piece of plumbing — the marker-probe duplication across 4 files is a little repetitive but is the right tradeoff to keep install-time/runtime/resync all in lockstep without a shared-sourcing refactor, and the docs (`defaults/CLAUDE.md`, `.loom/docs/troubleshooting.md`) accurately reflect the new behavior.

Issue #4028, opened 2026-07-27
Reduce cross-host sweep collisions: advertise work-in-progress via a shared room (soft claim) alongside the forge-label claim


rjwalters:
## Context
As loom moves toward multi-host development (one daemon per machine, shared repo backlog), the only cross-host coordination for claiming an issue is the **forge label** (`loom:issue → loom:building`). That has a race window: two hosts can both see an issue as `loom:issue` and dispatch before either flip propagates → duplicate concurrent sweeps for the same issue (duplicate PRs, wasted tokens). The per-issue `mkdir`-atomic claim lock is filesystem-local, so it does **not** coordinate across hosts.

## Idea (from operator)
Have each daemon **advertise work-in-progress into a shared room** (safehouse — rjwalters/safehouse, the E2E message bus loom#3997 is about): broadcast "claiming loom#N on host H" the moment it decides to dispatch. Peer daemons subscribed to the room see the advertisement and back off far faster than the forge label propagates — shrinking the collision window.

## Important nuance
A room broadcast is eventually-consistent, so it is a **soft claim / fast backoff**, not a mutex — two hosts advertising near-simultaneously still race. So the durable design is: **advertisement (fast, shrinks the window) + an atomic authority for the final claim** (the forge label as today, or a room-ordered/serialized claim). Advertisement *reduces* collisions; it doesn't replace the authoritative claim.

## Why file it here
- It's the concrete multi-host coordination gap behind the "collisions" concern.
- It's a natural first real use of the safehouse comms layer (loom#3997) as a *coordination* primitive, not just human observability — daemons as room participants.
- Worth measuring the baseline collision rate first (label-only) to quantify what advertisement buys.

Related: loom#3997 (safehouse as comms layer), loom#4007 (re-dispatch/duplication under load).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Curator Enhancement

> **Curator re-verification 2026-07-28**, against `origin/main` @ `2b1d1362`.
> This supersedes the original enhancement below, which was written before #3997
> merged. **Three of its load-bearing claims are now false**, and the dependency
> that blocked this issue has cleared.

| Claim (original enhancement) | Status on `origin/main` @ `2b1d1362` |
|---|---|
[...]

Issue #3937, opened 2026-07-25
Multi-repo: sweep-registry label flips run gh issue edit without repo scoping — cross-repo claims fail, dispatch proceeds unclaimed


rjwalters:
Observed live on v0.15.0 canary rollout, first multi-workspace work-finder tick (4 managed repos, daemon cwd = the loom checkout):

```
[WARN] label flip for issue #6199 failed (continuing dispatch): gh issue edit failed for #6199: GraphQL: Could not resolve to an issue or pull request with the number of 6199. (repository.issue)
[WARN] label flip for issue #6193 failed (continuing dispatch): ...
[WARN] label flip for issue #6191 failed (continuing dispatch): ...
[INFO] work_finder: tick — ... 4 workspace(s), 11 seen, 3 dispatched, ...
```

#6199/#6193/#6191 are **vibesql** issue numbers. The work-finder *query* is correctly scoped per workspace (`GhWorkSource::for_root` sets `current_dir(root)`, #3928), but the claim flip is not: `SweepRegistry::flip_label_to_building` (loom-daemon/src/sweep_registry.rs:1039) and `restore_label_to_ready` (:1067) run `gh issue edit <N>` with **no `current_dir` and no `--repo`** (only a process-global `LOOM_REPO` env escape hatch, which cannot be right per-registry). The command therefore resolves against the daemon process's cwd repo (loom), where the issue number doesn't exist.

Consequences:
- Cross-repo dispatch proceeds **without the forge-visible `loom:building` claim** — the #3823/#3880 double-dispatch protection is degraded to the in-memory registry only (a second daemon/manual sweep on another machine could double-build).
- The reaper's crash-path label restore silently fails the same way.
- One warning per candidate per tick = sustained log noise.

## Suggested fix
The registry already knows its root: add `cmd.current_dir(&self.config.workspace_root)` in both flip functions (and any other `gh` invocation inside SweepRegistry that assumes the process cwd — audit the file). Per-registry `--repo` derived from the root's `origin` would also work but `current_dir` matches how `for_root` solved the identical problem for the query side.

## Acceptance criteria
- [ ] `flip_label_to_building` / `restore_label_to_ready` (and any other forge call in SweepRegistry) execute in the registry's `workspace_root`.
- [ ] A multi-workspace test asserts the flip command runs with the non-default registry's root (extend the existing gh_calls mock tests at sweep_registry.rs:4094+).
[...]

PR #4116, opened 2026-07-27
feat(daemon): observe cross-host dispatch collisions at label-flip time

### LINKED ISSUE #4085: Instrument cross-host sweep collisions: measure the baseline duplicate-dispatch rate (Phase 0 of #4028)
## Context

Split out of #4028 (cross-host sweep collisions via shared-room advertisement). #4028's own body says: *"Worth measuring the baseline collision rate first (label-only) to quantify what advertisement buys."*

#4028 is **blocked** on #3997 (safehouse comms layer, not yet present in loom at all). This measurement half is **not** blocked and is a genuine prerequisite: without a baseline there is no way to tell whether the coordination machinery #4028 proposes is worth its complexity, or which of its phases buys more.

## Problem

When two `loom-daemon` hosts share one repo backlog, both can dispatch the same issue. Today that produces **no signal whatsoever**:

- The `mkdir` claim lock (`.loom/locks/issue-<N>/`) is filesystem-local — a peer host's lock is invisible (`loom-daemon/src/sweep_registry.rs:987-988`).
- `flip_label_to_building` is an unconditional `gh issue edit --remove-label loom:issue` (`loom-daemon/src/sweep_registry.rs:1584-1597`); it succeeds whether or not the label was still there, so the losing host is never told it lost.
- The per-tick log line counts `labeled-skip` / `in-flight-skip`, but a cross-host collision is by construction *not* caught by either.

So the collision rate is currently unobservable, and any claim about improving it would be unfalsifiable.

## Scope

Make cross-host duplicate dispatch **observable**. Detection only — no coordination, no backoff, no new transport. A detected collision is recorded and reported; behavior is otherwise unchanged.

The detection signal: make the label flip **verify** rather than assume. If `loom:issue` was already gone when this host tried to remove it — or `loom:building` was already present before this host's flip — another claimant got there first.

> **Curator correction (2026-07-27):** the original filing said "`gh issue edit` can report the resulting label set." **It cannot** — `gh issue edit` has no `--json` flag and emits only the issue URL. More importantly, a **post-flip** read-back cannot satisfy the second acceptance criterion at all: once this host has flipped the labels, a collided issue and a clean one look *identical*. Detection must observe the **pre-flip** state. See "Implementation Options" below.

Note the extra API round-trip is the main cost; the Builder should measure it and confirm it does not meaningfully lengthen the dispatch path.

## Acceptance Criteria

- [ ] The daemon detects, at dispatch time, that another claimant had already flipped the issue, and records it as a distinct **collision** event/counter — not folded into an existing skip counter.
- [ ] A collision record carries enough to diagnose it: issue number, repo/workspace, this host's identity, timestamp, and the observed **pre-flip** label state (see Curator correction under Scope — a post-flip read cannot supply this).
- [ ] The collision count is surfaced to an operator through an existing read-only path (e.g. the per-tick work-finder log line and/or `loom-daemon --status`). No new IPC/MCP tool is required.
- [ ] **Detection only** — on a detected collision the daemon's dispatch behavior is unchanged in this issue. (Whether the loser should abort is #4028's Phase 2 decision; changing it here would pre-empt that.)
- [ ] Default-on is acceptable **only** if the added cost is bounded and measured; otherwise gate it behind a config/env flag consistent with the `autonomous.*` precedence rule (**env > config > default**).
- [ ] Zero behavior change on a single-host setup beyond the new counter.
- [ ] The per-tick log line's existing fields are **not** renamed or repurposed — several docs and this issue's siblings quote them verbatim.

## Implementation Options

Two credible mechanisms. The Builder should pick one and say which in the PR body.

### Option A — pre-flip read (lower blast radius, recommended)

Before `flip_label_to_building`, issue `gh issue view <N> --json labels` and inspect the set. `loom:issue` absent (or `loom:building` already present) ⇒ collision.

- **Pros**: leaves the existing `gh issue edit` flip byte-for-byte intact; trivially unit-testable with the existing fake-`gh` harness; directly yields the "observed pre-existing label state" the AC asks for.
- **Cons**: one extra API round-trip per dispatch; a TOCTOU window between read and flip (acceptable — this is a post-hoc *baseline estimate*, and the window under-counts rather than inflates).

### Option B — REST `DELETE` label status code (atomic, no extra call)

Replace the removal half of the flip with `gh api --method DELETE repos/{owner}/{repo}/issues/<N>/labels/loom:issue`. GitHub returns **404** when the label is not on the issue — an atomic "was it still there?" answer with no extra round-trip and no TOCTOU window.

- **Pros**: strictly more accurate; no added latency.
- **Cons**: rewrites the flip mechanism (needs a second call to add `loom:building`, and its own error taxonomy), so it is a larger change to a path that is already load-bearing. Only choose this if Option A's added latency measures badly.

**Fail-closed either way**: a `gh` timeout, non-zero exit, or unparseable response is **"unknown"**, never "collision" — see the Edge cases in the Test Plan.

## Affected Files

*(re-verified on `origin/main` @ `cf94d174`, 2026-07-27 — line refs below supersede the `c4b6f677` refs in the original filing, which drifted ~9 lines)*

- `loom-daemon/src/sweep_registry.rs` — `flip_label_to_building` (`:1593-1631`); detect the pre-existing claim and surface it to the caller at the dispatch step (`:1354-1361`). Claim-lock context: `locks_dir` (`:993-997`), module docs (`:16`, `:40`).
- `loom-daemon/src/work_finder.rs` — collision counter in the per-tick summary alongside `labeled-skip` / `in-flight-skip` (**two** call sites: `:1045` and `:1224` — both format strings must be updated, or the counter is invisible on one of the two tick paths)
- `.loom/docs/daemon-reference.md` — document the counter and how to read it

## Existing Harness to Reuse

Do not build a new test fixture. `SweepRegistryConfig.gh_bin: Option<PathBuf>` (`sweep_registry.rs:925`) already injects a stub `gh`, and there is an established pattern — a bash script that logs `"$*"` to a temp file and exits 0 — used at **8** existing call sites (`:5718, :5751, :5887, :5950, :6029, :6087, :6149, :7360`). `cancel_does_not_restore_label_when_pr_produced` (`:6073`) is a good template: it asserts on the recorded invocation string. For collision tests the stub additionally needs to emit a `--json labels` payload on stdout.

`config.skip_label_flip` (`:928`) already gates the whole flip path, so the "disabled ⇒ dispatch unchanged" test has a natural seam.

## Test Plan

- [ ] **Unit**: a fake `gh` returning a label set showing the issue was *already* `loom:building` (or already missing `loom:issue`) is classified as a collision; the normal flip is not.
- [ ] **Unit**: the collision counter increments once per collision and appears in the per-tick summary; existing counters keep their current names and semantics.
- [ ] **Unit**: with the feature disabled (if gated), the dispatch path is byte-for-byte unchanged.
- [ ] **Manual**: two daemons against one repo and one shared backlog; confirm at least one collision is recorded, and that both sweeps still behave as they do today (detection only).
- [ ] **Edge cases**: `gh` timeout during read-back must not be misreported as a collision (fail *closed* on ambiguity — an unverifiable flip is "unknown", not "collision", so the baseline is not inflated); issue closed/relabeled between poll and flip.
- [ ] No new fixed `sleep` in tests — see #4044 (hardcoded waits manufacture false failures under exec latency).
- [ ] `cd loom-daemon && cargo test` green.

## Dependencies

None. This is deliberately the unblocked half of #4028.

## Related

- #4028 — parent: shared-room advertisement (blocked on #3997)
- #3997 — safehouse comms layer
- #4007 — *single-host* duplicate dispatch under load; distinct root cause, but its counters should not be confused with these

🤖 Filed by Curator while curating #4028.

rjwalters:
## Summary

Makes cross-host duplicate `loom-daemon` dispatch of the same issue **observable** (Phase 0 of #4028). When two hosts share one repo backlog, both can flip `loom:issue -> loom:building` for the same issue and neither is told it lost — so the collision rate is currently unmeasurable, and any claim that #4028's coordination layer improves it is unfalsifiable. This adds an opt-in detection-only baseline. No coordination, no backoff, no behavior change on a detected collision.

## Implementation option chosen: **Option A (pre-flip read)**

Option A per the issue: before `flip_label_to_building`, when detection is enabled, issue `gh issue view <N> --json labels` and inspect the set. Chosen over Option B (REST `DELETE` label status code) because it leaves the existing `gh issue edit` flip **byte-for-byte intact**, is trivially unit-testable with the existing fake-`gh` harness, and directly yields the observed pre-flip label state the acceptance criteria ask for. Option B rewrites a load-bearing path (a second call to add `loom:building`, its own error taxonomy) for an accuracy gain the baseline does not need — the TOCTOU window in Option A under-counts rather than inflates, which is the safe direction for a baseline estimate. Option A's added latency (below) measured fine, so Option B's "only if A measures badly" condition was not triggered.

## Measured latency cost

The extra `gh issue view <N> --json labels` round-trip measured **~0.37-0.40s** against GitHub (3 samples: 0.37s, 0.40s, 0.38s). It is bounded by the same `LOOM_REAP_GH_TIMEOUT_SECS` ceiling as the daemon's other best-effort `gh` calls, and only runs when detection is explicitly enabled. Because it is a real per-dispatch API cost, detection is **off by default** and opt-in (see AC table). This does not meaningfully lengthen the dispatch path relative to the existing flip (itself a `gh` round-trip).

## Changes

- `loom-daemon/src/sweep_registry.rs`:
  - `classify_preflip_labels` — bounded `gh issue view --json labels`, returns `Collision { labels }` / `Clean` / `Unknown`; fail-closed on timeout/non-zero/unparseable.
  - `detect_and_record_collision` — called from `dispatch` just before the flip (no-op + zero `gh` calls when disabled); logs the diagnostic record and increments the counter on a confirmed collision.
  - `host_identity()` helper (`LOOM_HOST_ID` -> `$HOSTNAME` -> `hostname` -> `unknown-host`), `collision_count()` accessor, `set_collision_detection` setter, and `resolve_collision_detection` (env > config > default(off)).
- `loom-daemon/src/work_finder.rs`: `WorkDispatcher::collisions()` (default 0), `TickReport.collisions`, populated in `tick` (incl. halted path) and summed across workspaces in `tick_multi`, and appended to **both** per-tick summary log call sites as `N cross-host-collision(s)`. Existing fields unchanged.
- `loom-daemon/src/main.rs` + `workspace_pool.rs`: wire the resolved flag onto the seed registry and every pooled workspace registry.
- `.loom/docs/daemon-reference.md`: config rows + a dedicated section documenting the counter and how to read it.

| Detects at dispatch time that another claimant flipped first; distinct collision counter (not folded into a skip counter) | Done | `detect_and_record_collision` runs pre-flip in `dispatch`; new `collision_count`, separate from `skipped_*`. `detect_and_record_collision_increments_counter` |
| Record carries issue #, repo/workspace, host identity, timestamp, observed pre-flip labels | Done | `log::warn!` record in `detect_and_record_collision` includes all five |
| Count surfaced via an existing read-only path (per-tick log line) | Done | `N cross-host-collision(s)` on both work-finder call sites; `test_tick_surfaces_collision_total`, `test_tick_multi_sums_collision_totals` |
| Detection only — dispatch behavior unchanged | Done | Detection precedes the untouched flip; nothing branches on the result |
| Default-on only if bounded+measured, else gated (env > config > default) | Done | Gated off by default; `resolve_collision_detection`, `resolve_collision_detection_env_overrides` |
| Zero behavior change single-host beyond the counter | Done | `detect_and_record_collision_disabled_is_noop` asserts no `gh` call when disabled |
| Per-tick log line existing fields not renamed/repurposed | Done | New field appended only; `labeled-skip` / `in-flight-skip` / `quarantine-skip` unchanged |
| Fail-closed on ambiguity (timeout/error/unparseable = unknown, not collision) | Done | `classify_preflip_labels_fail_closed_on_gh_error`, `_on_unparseable`, `detect_and_record_collision_unknown_not_counted` |

## Test Plan

- `cd loom-daemon && cargo test` — **green** (1197 lib + integration tests pass). Two `integration_security` terminal-ID-validation tests flaked once under concurrent test-binary load (machine-wide `cleanup_all_loom_sessions` race); both pass in isolation (`cargo test --test integration_security` → 14/14) and are unrelated to this change.
- New unit tests reuse the existing fake-`gh` bash-stub harness (stdout now also emits a `--json labels` payload). No fixed `sleep` added (per #4044).
- `cargo clippy --all-targets` clean; `cargo fmt` applied.

Closes #4085

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1426,6 +1426,8 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.watchdog.intervalSecs` | `LOOM_SWEEP_WATCHDOG_INTERVAL_SECS` | `30` | Watchdog probe cadence (shared by all three backstops) |
 | `autonomous.watchdog.reviewStall` | `LOOM_SWEEP_REVIEW_STALL` | `true` | Review-phase stall watchdog on/off (#3910) |
 | `autonomous.watchdog.reviewStallTimeoutSecs` | `LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS` | `2700` | Log-silence window before a hung Judge/Doctor sweep is re-dispatched |
+| `autonomous.collisionDetection.enabled` | `LOOM_DETECT_COLLISIONS` | `false` | Cross-host dispatch-collision baseline (#4085). Off by default — adds one extra `gh issue view --json labels` round-trip per dispatch. Detection only: a collision is logged/counted, never acted on |
+| *(host identity, records only)* | `LOOM_HOST_ID` | `$HOSTNAME` → `hostname` → `unknown-host` | Overrides this host's identity string in collision log records (#4085); set it where the daemon runs without `$HOSTNAME` exported |
 
 ### Daemon log path override (`LOOM_DAEMON_LOG`, #4010)
 
@@ -1574,6 +1576,52 @@ candidate is dropped **before** the global slot-fill pass, so a workspace whose
 only candidates are quarantined never reserves a shared dispatch slot — healthy
 sibling work in other repos gets it instead. Defaults **on**; disable with
 `LOOM_WORK_FINDER_QUARANTINE=0` or `autonomous.workFinder.quarantine.enabled = false`.
+
+### Cross-host dispatch-collision baseline (#4085, Phase 0 of #4028)
+
+When two `loom-daemon` hosts share one repo backlog, both can dispatch the same
+issue: the `mkdir` claim lock (`.loom/locks/issue-<N>/`) is filesystem-local, so
+a peer host's lock is invisible, and the `loom:issue → loom:building` flip
+succeeds whether or not the label was still there — the losing host is never
+told it lost. This makes the cross-host collision rate **unobservable**, which is
+the prerequisite gap #4028's coordination layer has to justify closing.
+
+Collision detection makes that rate **measurable** (detection only — no
+coordination, no backoff, behavior is otherwise unchanged). When enabled, just
+before the label flip the registry reads the issue's **pre-flip** label state
+(`gh issue view <N> --json labels`) and classifies it:
+
+- `loom:issue` already gone **or** `loom:building` already present → **collision**
+  (a peer host claimed it first). A diagnostic record is logged at `warn` — issue
+  number, repo/workspace, this host's identity (`LOOM_HOST_ID` → `$HOSTNAME` →
+  `hostname` → `unknown-host`), timestamp, and the observed pre-flip label set —
+  and a per-registry cumulative counter is incremented.
+- `loom:issue` present and `loom:building` absent → **clean** (this host is first).
+- gh timeout / non-zero exit / unparseable JSON → **unknown**. **Fail-closed:**
+  an unverifiable read is never counted as a collision, so the baseline is never
+  inflated.
+
+**How to read the count.** The running total is surfaced on the work-finder's
+per-tick summary line as the trailing `N cross-host-collision(s)` field, e.g.:
+
+```
+work_finder: tick — cap 3 (...); 5 seen, 2 dispatched, 1 labeled-skip,
+0 in-flight-skip, 0 quarantine-skip, 0 deferred, 0 error(s), 3 cross-host-collision(s)
+```
+
+Unlike the other (per-tick) counters, this one is a **monotonic cumulative
+total** read from the dispatcher(s) at tick end (summed across workspaces in the
+multi-repo tick), so successive lines show the baseline accumulate. It is `0`
+whenever detection is disabled. The existing fields (`labeled-skip` /
+`in-flight-skip` / `quarantine-skip` …) keep their names and semantics unchanged.
+
+**Cost / default.** Enabling detection adds exactly one `gh issue view --json
+labels` round-trip per dispatch (measured ~0.4s against GitHub), bounded by the
+same `LOOM_REAP_GH_TIMEOUT_SECS` ceiling as the other best-effort `gh` calls.
+Because that is a real per-dispatch API cost it is **off by default**; enable it
+per the config/env row above (`LOOM_DETECT_COLLISIONS=1` or
+`autonomous.collisionDetection.enabled = true`, precedence **env > config >
+default**) on the hosts sharing a backlog while you take the measurement.
 
 ### Prerequisite: a fresh token ranking (#3894, self-refreshed by the daemon since #3969)
 
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -688,6 +688,21 @@ async fn main() -> Result<()> {
         quarantine_config.insta_crash_secs
     );
 
+    // Cross-host collision detection (#4085, Phase 0 of #4028): resolve env >
+    // config > default(off) for the default workspace so a shared-backlog
+    // deployment can measure the baseline duplicate-dispatch rate. Detection
+    // only — a detected collision is logged/counted, never acted on.
+    let detect_collisions = sweep_registry::resolve_collision_detection(&sweep_workspace);
+    sweep.set_collision_detection(detect_collisions);
+    log::info!(
+        "sweep_registry: cross-host collision detection {} (#4085)",
+        if detect_collisions {
+            "enabled"
+        } else {
+            "disabled"
+        }
+    );
+
     let sweep_registry = Arc::new(Mutex::new(sweep));
     let _reaper_handle = sweep_registry::spawn_reaper_task(sweep_registry.clone());
 
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -482,6 +482,21 @@ pub const REVIEW_STALL_ENABLE_ENV: &str = "LOOM_SWEEP_REVIEW_STALL";
 /// seconds.
 pub const REVIEW_STALL_TIMEOUT_ENV: &str = "LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS";
 
+/// Env var toggling cross-host dispatch-collision detection (Issue #4085,
+/// Phase 0 of #4028). Precedence **env > config > default**; default **off**
+/// because the probe adds one extra `gh issue view` round-trip per dispatch.
+/// `1`/`true`/`yes`/`on` enable; anything else disables. When enabled, the
+/// daemon does a pre-flip label read and records — but never acts on — the
+/// case where a peer host already claimed the issue. Detection only.
+pub const COLLISION_DETECT_ENV: &str = "LOOM_DETECT_COLLISIONS";
+
+/// Env var overriding this host's identity string in collision records (Issue
+/// #4085). Falls back to `$HOSTNAME`, then the `hostname` binary, then
+/// `"unknown-host"`. Set it when the daemon runs somewhere `$HOSTNAME` is not
+/// exported (a non-interactive service unit) so cross-host collision logs stay
+/// attributable.
+pub const HOST_ID_ENV: &str = "LOOM_HOST_ID";
+
 /// Default review-phase stall timeout (45 min of zero log output). A sweep that
 /// has already made startup progress (worktree/checkpoint exists) but whose log
 /// file has not been appended to within this window is treated as wedged in a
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

## Summary

Clean, well-scoped Phase 0 implementation of #4085 (cross-host dispatch-collision baseline). Reviewed the diff in full, checked out the builder worktree (`.loom/worktrees/issue-4085`), and ran the relevant test suite locally.

## Acceptance criteria verification (against #4085)

| Criterion | Status |
|---|---|
| Distinct collision counter (not folded into skip counters) | ✅ `collision_count` on `SweepRegistry`, separate from `skipped_labeled`/`skipped_in_flight`/`skipped_quarantined` |
| Record carries issue #, repo/workspace, host, timestamp, pre-flip labels | ✅ `detect_and_record_collision`'s `log::warn!` includes all five |
| Surfaced via existing read-only path | ✅ Both work-finder per-tick summary call sites (`tick` single-workspace and `spawn_multi_work_finder_task` multi-workspace) append `N cross-host-collision(s)` |
| Detection only — dispatch unchanged | ✅ `detect_and_record_collision` is called before `flip_label_to_building` and never influences it; correctly placed *inside* the `!self.config.skip_label_flip` gate so it doesn't fire when the flip itself is disabled (e.g. in most existing dispatch tests) |
| Gated env > config > default(off) | ✅ `resolve_collision_detection` follows the same precedence pattern as `resolve_quarantine_config`/`resolve_review_stall_timeout` |
| Zero behavior change single-host beyond counter | ✅ `detect_and_record_collision_disabled_is_noop` asserts zero `gh` invocations when disabled |
| Existing per-tick fields not renamed | ✅ New field appended at the end of both format strings; `labeled-skip`/`in-flight-skip`/`quarantine-skip`/`deferred`/`error(s)` untouched |
| Fail-closed on ambiguity | ✅ `classify_preflip_labels_fail_closed_on_gh_error` / `_on_unparseable` cover non-zero exit and unparseable JSON; timeout path returns `Unknown` via `output_with_timeout`'s `Ok(None)` arm |

Pre-flip-vs-post-flip ordering is correctly reasoned through in both the issue and this PR: reading labels *after* the flip would make a collided issue indistinguishable from a clean one, and the code places the read immediately before `flip_label_to_building`.

## Code quality

- `classify_preflip_labels`/`detect_and_record_collision` are small, single-purpose, and well-documented — the doc comments explain *why* (pre-flip vs. post-flip, fail-closed rationale, hot-path avoidance for `host_identity()`), not just what.
- `host_identity()` correctly defers the `hostname` subprocess spawn to the rare collision-confirmed path rather than the hot dispatch path.
- Config precedence resolution (`resolve_collision_detection`) matches the established pattern of sibling resolvers (`resolve_quarantine_config`, `resolve_review_stall_timeout`) — consistent with this repo's `env > config > default` convention.
- `gh` repo-scoping (`cmd.current_dir(&self.config.workspace_root)`, `LOOM_REPO` override) matches the existing pattern used by `flip_label_to_building` and other `gh` call sites in the file (#3937 scoping).
- `work_finder.rs`: `collisions()` trait method defaults to `0` so existing `WorkDispatcher` impls (including test fakes) don't need updating; both `tick` (including the halted early-return) and `tick_multi` correctly read/sum the cumulative total.

## Test coverage

- 8 new focused unit tests in `sweep_registry.rs` covering: collision via `loom:building` present, collision via `loom:issue` absent, clean, fail-closed on gh error, fail-closed on unparseable JSON, counter increments only on confirmed collision, disabled-is-true-no-op (asserts zero `gh` calls via the fake-gh log), and config precedence (env > config > default, including env overriding a config-enabled value).
- 3 new tests in `work_finder.rs` covering the report surfacing (`tick`, halted `tick`, and `tick_multi` summing across workspaces).
- Reuses the existing fake-`gh` bash-stub harness rather than inventing new fixtures, per the issue's guidance.
- No new fixed `sleep`s (consistent with #4044).

## Verification performed

- Read the full diff (`.loom/docs/daemon-reference.md`, `main.rs`, `sweep_registry.rs`, `work_finder.rs`, `workspace_pool.rs`).
- `cargo test` in the worktree: `sweep_registry::tests` + `work_finder::tests` — 151 passed, 0 failed (includes all new collision tests).
- CI: `Rust Linting & Build Check` ✅, `Rust Unit Tests` ✅, `CLAUDE.md Line Budget` ✅ (Installer Integration Tests / MCP Server Build skipped — no relevant paths touched).
- PR body includes `Closes #4085`.

No blocking issues found. Nice work threading the counter through both work-finder call sites and getting the pre/post-flip ordering right — that's exactly the subtlety the issue's Curator correction called out.

Issue #4111, opened 2026-07-27
fix(sweep): daemon-dispatched child self-skips its own claim despite LOOM_SWEEP_CLAIM_OWNED being set — regression of #3823/#3967, strands issues via quarantine


rjwalters:
Found by Champion (2026-07-27 ~22:35Z) while investigating why **five** freshly-promoted issues were auto-quarantined within seven minutes of promotion.

## Summary

`SweepRegistry::dispatch` correctly pre-flips `loom:issue → loom:building` and exports `LOOM_SWEEP_CLAIM_OWNED=<N>` into the child (#3823 / #3967 both closed as fixed, and the plumbing genuinely works — see evidence). But the child `/loom:sweep` **still self-skips its own claim**: it reasons from the label timeline (`loom:building` applied seconds ago ⇒ "another live process owns this") and exits without ever consulting the marker sitting in its own environment.

`#3967` fixed the marker being *absent* from the child env. This is the next failure mode: the marker is **present and ignored**.

## Evidence — the marker is in the env, and the child skips anyway

`.loom/logs/sweep-issue-4037.log`:

```
==== loom-daemon dispatch: 2026-07-27T21:36:44Z sweep_id=sweep-issue-4037-1785188202 issue=4037 ====
[2026-07-27T21:36:44Z] spawn-claude: LOOM_SWEEP_CLAIM_OWNED=4037     <-- marker IS exported
[2026-07-27T21:36:44Z] spawn-claude: using OAuth account 'agent4-2amlogic' (mode=ranked)

**Sweep for issue #4037 exited without dispatching a builder.**

Pre-flight found the issue already carries `loom:building`, applied at 21:36:43Z — literally
seconds before this sweep surveyed it. ... This is a fresh, live claim, not an orphaned one —
another process (or you, in another session) is actively building it right now.
```

The child names the exact symptom the marker exists to disambiguate ("applied seconds before this sweep surveyed it") and still concludes another process owns it.

`.loom/logs/sweep-issue-4069.log` — same pattern, different issue:

```
[2026-07-27T22:26:18Z] spawn-claude: LOOM_SWEEP_CLAIM_OWNED=4069
**Skipping issue #4069 — it's already being built by another live process.**
... a curate → promote-to-`loom:issue` → comment → claim-as-`loom:building` sequence that
completed within the last ~2 minutes, ending 22 seconds before this session started.
```

[...]

Issue #4110, opened 2026-07-27
fix(daemon): quarantine TTL auto-release strands loom:blocked on the forge — issues silently leave the queue forever


rjwalters:
## Problem

The #3939 insta-crash quarantine has **two release paths that are not symmetric**, and one of them permanently strands the issue.

When the daemon quarantines an issue it does two things: applies an in-memory pause/tally, and swaps `loom:issue` → `loom:blocked` on the forge. The quarantine comment the daemon posts describes both exits:

> Once the underlying cause is fixed, clear the quarantine with `loom-daemon quarantine clear <N>` (this releases the in-memory pause **AND** restores `loom:issue`), or simply wait for the 3600s TTL to auto-release it.

Note the asymmetry the parenthetical admits: the **CLI path restores the label**; the **TTL path is described only as releasing the pause**. It does not restore `loom:issue`.

## Consequence

After TTL expiry the issue is in a state where *nothing is blocking it and nothing will ever pick it up*:

- The daemon holds no quarantine — `loom-daemon quarantine clear <N>` returns `"was not quarantined — nothing to clear (no-op)"`.
- The forge still shows `loom:blocked`.
- The work finder polls open **`loom:issue`**, so the issue is invisible to it.

The circuit breaker is therefore not a *pause* but a permanent removal from the queue whenever the operator doesn't happen to run the CLI clear within the TTL window. Because the daemon posts a comment saying the TTL will "auto-release it", the failure is silent and actively misleading — it reads as self-healing.

## Observed instance (2026-07-27)

Found by Champion while draining the promotion queue. Both #4044 and #4037 sat at `loom:blocked` with quarantine comments, while the daemon reported no quarantine for either:

```
$ loom-daemon quarantine clear 4044
Issue #4044 was not quarantined — nothing to clear (no-op).
$ loom-daemon quarantine clear 4037
Issue #4037 was not quarantined — nothing to clear (no-op).
```

Both were manually restored to `loom:issue` by Champion. The token pool that likely triggered the original quarantines had already recovered (`loom-tokens check --ranking`: 3 of 7 available), so both had been stranded on a stale label with no live cause.

Note that a **daemon restart** produces the same end state by a different route, and is likely the more common trigger: quarantine state is explicitly in-memory ("the quarantine state lives in the daemon's memory (not on disk)", per `loom-daemon quarantine --help`), so any restart drops every pause while leaving every `loom:blocked` label behind.

## Suggested direction

[...]

PR #4117, opened 2026-07-27
fix(daemon): latch sweep progress per SweepId + 300s watchdog default + closed-issue dispatch guard

### LINKED ISSUE #4088: fix(daemon): #3887 startup watchdog cancels healthy sweeps — 120s threshold sits inside the normal 110–150s dispatch→worktree window
## Problem

The #3887 startup watchdog cancels **healthy** sweeps. Its 120s no-progress threshold sits in the middle of the normal dispatch→worktree latency distribution (observed ~110–150s under concurrency), so under load whether a sweep survives its first two minutes is close to a coin flip. Each false positive throws away a sweep's startup work, churns the issue's labels, and — when the bounded retry also trips — releases the claim entirely and leaves the issue unbuilt.

## Measurements (2026-07-27, 28-core host, 7 concurrent sweeps + build gate running)

Threshold: `DEFAULT_WATCHDOG_TIMEOUT_SECS = 120` (`sweep_registry.rs:437`), polled every `DEFAULT_WATCHDOG_INTERVAL_SECS = 30` (`:440`), so effective firing lands at **120–150s**.

| Issue | Dispatched | Watchdog fired | Outcome |
|---|---|---|---|
| #4082 | 13:52:12 | 13:54:40 (148s) | cancelled + re-dispatched → **then succeeded** (worktree at 13:56) |
| #4061 | 13:54:18 | 13:56:44 (146s) | cancelled + re-dispatched → **then succeeded** (commit `06b6e800` pushed) |
| #4060 | 13:54:16 | 13:56:40 (144s) | cancelled + re-dispatched → fired again 13:59:10 → **gave up, claim released** |
| #4078 | 13:54:12 | *never* | worktree at ~110s — **escaped by ~10s**; produced PR #4087 at 14:08 |

The last row is the control. #4078 was doing exactly what the other three were doing; it survived only because its worktree appeared just under the wire. Nothing about the cancelled sweeps was unhealthy — two of the three completed successfully immediately after being killed and restarted, which is the signature of a false positive rather than a rescue.

## Why the progress signal is wrong, not just the threshold

The watchdog's liveness test is "a worktree or checkpoint exists, or the log has grown past the spawn header." But a sweep legitimately does substantial work before any of those appear:

- MCP server init (`mcp-loom` node child spawn + handshake)
- skill load for `/loom:sweep`
- Stage -1 backend probes, host-sleep and main-freshness checks
- the **entire Curator phase**, which runs before the Builder creates the worktree

None of that touches the filesystem in a way the watchdog can see, and `spawn-claude.sh`'s own output stops after the account line, so the log stays flat at ~5 lines while the agent is actively working. A sweep spends nearly all of this wall-clock **network-bound**, so it also shows ~0% CPU — which looks like a hang to every coarse check, including `ps`.

## Impact

1. **Wasted work** — a cancelled sweep loses its Curator phase and starts over.
2. **Label churn** — cancel releases `loom:building` → `loom:issue`, re-dispatch re-claims. Observed on #4060: `+loom:issue` / `+loom:building` flipping four times in seven minutes.
3. **Dropped work** — when the bounded retry also trips, the claim is released for good. #4060 ended `loom:issue` with no worktree and no PR, requiring the operator intervention its own ERROR line requests.
4. **Re-dispatch risk** — a released issue is visible to the work finder again while its original sweep may still hold a registry slot.
5. **Load-correlated** — the threshold bites hardest exactly when the queue is busiest, i.e. when throughput matters most.

## Proposed fix

Some combination of:

1. **Raise the default** well clear of the observed distribution (300s+), and/or scale it with in-flight sweep count the way the concurrency cap already scales with load.
2. **Give the sweep an explicit heartbeat** rather than inferring liveness from filesystem side effects — have `/loom:sweep` emit a phase marker (it already writes checkpoints later; an earlier `phase=curator` marker would make startup observable) and have the watchdog read that.
3. **Don't count network-bound time as no-progress** — if a heartbeat is too invasive, at minimum treat "child process alive and its MCP child alive" as weaker evidence of life than "log unchanged."

The threshold is already configurable (`LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` / `autonomous.watchdog.timeoutSecs`, `sweep_registry.rs:428`, `:3625`), so (1) is a one-line default change; (2) is the durable fix.

## Acceptance criteria

1. A sweep that is progressing normally through Curator on a loaded host is not cancelled.
2. A genuinely hung sweep (no MCP child, or child dead, or truly no progress) is still cancelled — the watchdog must not be defanged into uselessness.
3. Whatever signal replaces or supplements the filesystem check is documented in `.loom/docs/daemon-reference.md`.
4. A test covers the "slow but healthy startup" case that currently false-positives.

## Notes

- Original watchdog: #3887.
- The `0% CPU` / `log stuck at spawn header` appearance fooled a human operator too during this investigation — a sweep at 12s CPU over 12 minutes with a 5-line log was **producing PR #4087 at that moment**. Worth a line in the troubleshooting docs: coarse liveness checks (`ps`, log size) cannot distinguish a wedged sweep from a working one.

---

All file/line references verified against `origin/main` @ `a6ef9a89`.

> **Body is now the source of truth.** The two comments below the body carry a
> *correction* to Impact #3 and an *escalation* that adds acceptance criteria 5–6. Both are
> folded in here so the Builder does not have to reconcile a body that contradicts its own
> comments. The consolidated AC list is at the bottom of this section and **supersedes** the
> 4-item list in the original body.

### Finding 1 — the root cause is that progress is **not latched**, not that a timer resets

The escalation comment explains the #4078 re-dispatch as "worktree exists → watchdog sees
progress, timer keeps resetting … then merge removes the worktree." The outcome is right but
the mechanism is not, and the difference decides the fix.

There is no timer to reset. `watchdog_decision` (`sweep_registry.rs:631-648`) is **memoryless**:

```rust
if made_progress || elapsed < timeout { Healthy } else if already_retried { GiveUp } else { Restart }
```

- `elapsed` is measured from `info.started_at` (`sweep_registry.rs:2807`) — **total runtime**,
  which only ever increases. It never rewinds on progress.
- `made_progress` is **recomputed from scratch every tick** by `sweep_made_progress`
  (`sweep_registry.rs:2755-2772`) — three filesystem existence checks, no memory of prior ticks.

So the state machine is a pure function of *right now*. The instant `sweep_made_progress` flips
back to `false`, `elapsed` is already ~1135s — vastly past the 120s timeout — and the very next
tick fires `Restart`. That is exactly the 1135s figure in the escalation comment: not a long
timer, but a long-running sweep evaluated as though it had just been born.

**Consequence for the fix**: the defect is the *absence of a latch*, and adding one is a small,
local change. Once a sweep has ever been observed making progress, the startup watchdog must
never reconsider it — see Finding 3 for why that is the *documented* intent already.

### Finding 2 — **all three** progress signals are torn down at successful completion, not just the worktree

The escalation names only `merge-pr.sh` removing the worktree. Verified: every signal
`sweep_made_progress` reads is transient by design.

| Signal | Checked at | Torn down / reset by |
|---|---|---|
| `.loom/worktrees/issue-N` | `sweep_registry.rs:2762` | `merge-pr.sh` removes the loom-managed worktree on merge (`defaults/scripts/merge-pr.sh:31-45`) |
| `.loom/sweep-checkpoint/issue-N.json` | `sweep_registry.rs:2769` | `/loom:sweep` runs `sweep-checkpoint.sh delete N` at completion (`defaults/.claude/commands/loom/sweep.md:1192`, `:1294`, `:1610`) |
| log past the spawn header | `sweep_registry.rs:2771` (`log_has_progress`, `:786`) | anchors to the **last** dispatch header, so a re-dispatch resets it (test `log_has_progress_anchors_to_last_dispatch`, `:7511`) |

This is stronger than the comment claims. It is not a worktree quirk that a longer timeout or a
tweak to one check could paper over: **a successfully completed sweep is, by construction,
indistinguishable from one that never started.** Any fix that leaves the probe stateless
reproduces the bug.

### Finding 3 — latching restores the *already-documented* division of labor between the three watchdogs

The mid-build-death watchdog's own module doc states the intended contract
(`sweep_registry.rs:679-690`):

> "Because `sweep_made_progress` returns `true` the instant a worktree exists, the startup
> watchdog **correctly leaves such a sweep alone** — so it does NOT rescue a sweep that HAD
> progress and then crashed"

…and delegates that case to #3895 (mid-build death, `:3022`) and #3910 (review stall, `:3193`).
All three run in the same tick (`spawn_watchdog_task`, `:3870-3931`).

So the startup watchdog's remit is *already specified* as "sweeps that have never progressed."
The code merely fails to implement it durably, because it re-derives "has progressed" from
mutable filesystem state instead of remembering it. **Latching is not new semantics — it is the
existing contract, made monotonic.** That argues strongly for it over a heartbeat (option 2) or a
liveness heuristic (option 3), both of which invent new signals to work around a state bug.

### Finding 4 — this is **two independent defects**, and each AC needs its own change

The body's proposed-fix list reads as alternatives. They are not; two distinct failures are
in play and each needs a separate change:

| Defect | Evidence | Fix | Satisfies |
|---|---|---|---|
| **A. Threshold sits inside the healthy distribution.** A sweep that has genuinely not progressed yet is killed at 120–150s, but healthy startup takes 110–150s. | #4082/#4061/#4060 vs the #4078 control | raise `DEFAULT_WATCHDOG_TIMEOUT_SECS` (`:446`) | AC1 |
| **B. Progress is not latched.** A *completed* sweep reads as never-started. | the 1135s #4078 re-dispatch | latch progress per sweep (Findings 1–3) | AC5 |

Neither subsumes the other. Latching does nothing for a sweep still inside its Curator phase with
no artifacts yet (defect A). Raising the threshold does nothing for defect B — as the escalation
correctly notes, *any* finite timeout eventually expires on a completed sweep whose signals are
gone. **Both changes are required.**

### Finding 5 — the closed-issue guard (AC6) belongs in `dispatch`, not in `watchdog_once`

Verified: **there is no closed/open issue check anywhere in `sweep_registry.rs`**
(`grep -nE 'closed|is_open'` returns nothing). Dispatch step 4 flips `loom:issue → loom:building`
via `gh issue edit` (`sweep_registry.rs:1352`, `:1606`), which **succeeds on a closed issue** — GitHub
permits label edits on closed issues — so there is no incidental guard today.

Critically, **all three watchdogs re-dispatch**: `watchdog_once` (`:2790`), `midbuild_watchdog_once`
(`:3022`), and `review_stall_watchdog_once` (`:3193`) each call `self.dispatch(...)`. Putting the
closed-issue check only in the startup watchdog leaves the identical bug reachable from the other
two. **Put the guard in `dispatch` (or a shared pre-flight helper it calls), so one change covers
all three call sites.** Make it best-effort and fail-open on a forge error — a `gh` outage must not
wedge dispatch.

### Finding 6 — where to store the latch, and the trap in the obvious choice

The obvious analogue is `watchdog_retried: HashSet<u32>` (`sweep_registry.rs:1064`) — but that set
is keyed by **issue number**, and copying that key for the latch is a live bug:

> A latch keyed by issue would persist across re-dispatch of the same issue. A re-dispatched sweep
> that then genuinely hangs at startup would be read as "already progressed" and never rescued —
> silently defanging the watchdog for exactly the issues it has already had to rescue once. This
> **violates AC2** and would not be caught by any existing test.

**Key the latch by `SweepId`, not by issue.** Two viable homes:

1. **Private `HashSet<SweepId>` on `SweepRegistry`** (recommended) — mirrors the existing
   `watchdog_retried` / `watchdog_gaveup` fields (`:1064`, `:1067`) and touches no wire schema.
   Note `watchdog_retried` is issue-keyed and never pruned (bounded by issue count, fine); a
   SweepId-keyed set grows **per dispatch**, so prune it wherever entries are removed.
2. A `#[serde(default)] made_progress: bool` field on `SweepInfo` (`loom-daemon/src/types.rs:664`)
   — self-pruning and observable via `list_sweeps`, but `SweepInfo` is the **canonical MCP wire
   schema** for mcp-loom and the UI, so this widens a public surface for internal bookkeeping.
   Rejected as the default; acceptable if the Builder wants the observability and documents the
   schema addition.

### Finding 7 — options 2 and 3 from the body should be dropped

- **Option 2 (explicit sweep heartbeat)** — requires editing the role prompt
  (`defaults/.claude/commands/loom/sweep.md`) to emit an early phase marker, plus a new
  reader daemon-side. Cross-layer, and unnecessary once progress is latched: the existing
  signals already fire correctly on the *rising* edge; only the falling edge is mishandled.
  Note it also would **not** have fixed the #4078 re-dispatch on its own — a marker file deleted
  at completion has exactly the same falling-edge problem.
- **Option 3 (child + MCP child alive as weaker evidence)** — refuted by the issue's own evidence
  in both directions. A live-and-working sweep sat at ~0% CPU (#4078), and a genuinely wedged
  MCP-init sweep *also* has a live child. PID liveness does not discriminate the two cases, so it
  cannot carry AC2.

### Recommended approach

Three changes, in this order:

1. **Latch progress per `SweepId`** (Findings 1–3, 6). In `watchdog_once` (`:2790`), once
   `sweep_made_progress` returns true for a sweep, record it and short-circuit `made_progress = true`
   on all subsequent ticks for that `SweepId`. Prune alongside entry removal.
2. **Raise `DEFAULT_WATCHDOG_TIMEOUT_SECS`** (`:446`) from `120` to **`300`** — clear of the observed
   110–150s distribution with headroom, and still an order of magnitude below the review-stall
   timeout (2700s), so the three backstops stay well separated. The knob already resolves
   env > config > default (`resolve_watchdog_timeout`, `:3714`), so this is a one-constant change.
3. **Guard re-dispatch on a closed issue** inside `dispatch` (Finding 5), best-effort and fail-open.

Explicitly **out of scope**: options 2 and 3 from the body (Finding 7); any change to the
mid-build-death (`:3022`) or review-stall (`:3193`) decision logic beyond the shared closed-issue
guard; scaling the timeout with in-flight sweep count (the body's option 1b — latching plus a 300s
floor makes the dynamic version unnecessary complexity, and it can be revisited if AC1 still fails).

### Scope guard

Target **≤ 4 files and ≤ 300 LOC including tests**. If the implementation starts touching the work
finder, the quarantine machinery (#3939), or `spawn-claude.sh`, stop and split — that is a signal
the latch is being built at the wrong layer.

## Consolidated acceptance criteria

Supersedes the 4-item list in the original body; items 5–6 come from the escalation comment.

1. A sweep progressing normally through Curator on a loaded host is not cancelled.
2. A genuinely hung sweep (no progress since dispatch) is still cancelled — the watchdog is not
   defanged. **Including on a re-dispatch**: a re-dispatched sweep that hangs at startup must still
   be rescued (Finding 6).
3. The progress signal's semantics are documented in `.loom/docs/daemon-reference.md`.
4. A test covers the "slow but healthy startup" case that currently false-positives.
5. A sweep whose PR has merged (worktree + checkpoint cleaned up) is **not** cancelled or
   re-dispatched.
6. Re-dispatch is skipped when the target issue is closed — enforced for **all three** watchdogs
   (Finding 5).

## Affected Files

- `loom-daemon/src/sweep_registry.rs` — the whole change lands here:
  - `DEFAULT_WATCHDOG_TIMEOUT_SECS` (`:446`) `120` → `300`.
  - New private `HashSet<SweepId>` latch field next to `watchdog_retried` / `watchdog_gaveup`
    (`:1064`, `:1067`); initialize at both constructors (`:1114`, `:1136`); prune on entry removal.
  - `watchdog_once` (`:2790`) — consult + populate the latch around the `sweep_made_progress` call
    (`:2813`). **Do not** change `watchdog_decision` (`:631`) — it is a pure function with four
    passing tests (`:7450-7492`); keeping it pure keeps those tests meaningful.
  - `dispatch` (`:1348-1352`) — closed-issue pre-flight guard, fail-open (Finding 5).
  - **Do not touch** `sweep_made_progress`'s three probes (`:2755-2772`) — the rising edge is
    correct; only its statelessness is the defect.
- `.loom/docs/daemon-reference.md` — update the watchdog knob table row for
  `autonomous.watchdog.timeoutSecs` (`:1425`, currently documents `120`) and the startup-watchdog
  prose (`:1495-1509`) to state that progress is latched per sweep and that re-dispatch skips
  closed issues (AC3).
- `defaults/scripts/merge-pr.sh` — **reference only, no change expected**. Its worktree cleanup
  (`:31-45`) is correct behavior; the watchdog must tolerate it, not the reverse.
- `defaults/.claude/commands/loom/sweep.md` — **reference only, no change expected**. The
  checkpoint delete at `:1192` / `:1294` / `:1610` is the second half of the teardown (Finding 2);
  option 2's heartbeat is out of scope (Finding 7).
- `loom-daemon/src/types.rs` — **reference only** unless the Builder chooses latch option 2
  (`SweepInfo`, `:664`), which is not the recommendation.

## Test Plan

All Rust tests go in the existing inline `#[cfg(test)]` module of `sweep_registry.rs`, alongside
`watchdog_restarts_hung_sweep_once_then_gives_up` (`:7928`) and
`watchdog_leaves_progressing_sweep_alone` (`:7981`), which are the closest prior art.

- [ ] **AC5 regression (the headline bug)** — a sweep that made progress (worktree present), then
      has its worktree **and** checkpoint removed while still `Running`, with `elapsed` far past the
      timeout, is **not** cancelled or re-dispatched. This test fails on `origin/main` today; assert
      it does, so the fix is proven rather than assumed.
- [ ] **AC2 on re-dispatch (the Finding 6 trap)** — a *second* sweep for the same issue that never
      progresses **is** still restarted/given-up. Guards against an issue-keyed latch silently
      defanging the watchdog. This is the highest-risk regression of the chosen design.
- [ ] **Latch monotonicity** — once observed true, the latch stays true across ticks for that
      `SweepId`, and is scoped to that `SweepId` only (a sibling sweep is unaffected).
- [ ] **Latch pruning** — the latch set does not retain entries for sweeps removed from `entries`
      (no unbounded growth across many dispatches).
- [ ] **AC1 threshold** — `resolve_watchdog_timeout` (`:3714`) returns 300s by default; extend the
      existing `resolve_watchdog_timeout_and_interval_precedence` test (`:8190`) so env > config >
      default still holds with the new default.
- [ ] **AC6 closed-issue guard** — `dispatch` for a closed issue is refused without flipping labels;
      a forge lookup error **fails open** (dispatch proceeds), so a `gh` outage cannot wedge the
      daemon.
- [ ] **No-regression** — `watchdog_decision` (`:631`) is unchanged and its four existing tests
      (`:7450-7492`) still pass unmodified; `midbuild_*` (`:7570-7846`) and `review_stall_*`
      (`:8223-8400`) suites pass unmodified.
- [ ] **Manual verification** on the 28-core host, reproducing the incident conditions:
  - [ ] Dispatch ≥3 concurrent sweeps with the build gate running; confirm no healthy sweep is
        cancelled during its Curator phase (AC1) — compare against the 110–150s window in the body.
  - [ ] Let one sweep run to a merged PR; confirm **no** `watchdog: … made no progress` line appears
        for it after `merge-pr.sh` removes its worktree (AC5). This is directly greppable in
        `.loom/logs/`.
  - [ ] Confirm a deliberately wedged sweep (e.g. spawn with an exhausted token) is still cancelled
        and re-dispatched once, then given up (AC2).
- [ ] **Edge cases**:
  - [ ] Watchdog disabled (`LOOM_SWEEP_WATCHDOG=0`) → no behavior change from the latch.
  - [ ] Daemon restart mid-sweep → reconstructed entries have no retained `Child` handle and are
        already excluded from watchdog candidacy (`:2800`); confirm the latch's absence for them is
        harmless (they are skipped before it is consulted).
  - [ ] A sweep that never progresses **and** whose issue is closed mid-flight → cancelled, but
        **not** re-dispatched (AC2 + AC6 interacting).

## Rollout note

The 1200s/watchdog logic is enforced **daemon-side in Rust**, so this fix requires a **daemon
rebuild and restart** to take effect — unlike a shell-script change, which goes live immediately.
The host currently runs a stale binary (see the pending-roll note in the operator's handoff), so do
not conclude "the fix did not work" from a run against an unrolled daemon. Sequence: merge → rebuild
→ `loom-daemon-start.sh` → observe.

## Related work

- **#3887** — the original startup watchdog this issue corrects.
- **#3895** (mid-build death, `:3022`) / **#3910** (review stall, `:3193`) — the two sibling
  backstops. Their remit is unchanged; Finding 3 shows this fix *restores* the boundary between
  them rather than moving it. Both share the closed-issue guard from Finding 5.
- **#3939** (insta-crash quarantine) — the fourth self-heal path. Out of scope here, but note the
  interaction: a watchdog cancel that writes no checkpoint could plausibly feed the insta-crash
  tally. Verify the fix does not change quarantine accrual.
- **#4110** / **#4111** — open, both quarantine-stranding defects on the same self-heal surface.
  Distinct root causes; no code overlap expected, but a Builder touching `dispatch` should check
  for merge conflicts against them.
- **#4084** — the companion build-gate contention issue from the same incident window; the load
  that produced these false positives is the load that issue is about.

rjwalters:
## Summary

Fixes the #3887 startup watchdog cancelling **healthy** sweeps and re-dispatching **completed** ones. The Curator analysis identified two independent defects; both are fixed, plus the AC6 closed-issue guard.

1. **Threshold sat inside the healthy distribution.** Under concurrency the normal dispatch→worktree latency is 110–150s, but `DEFAULT_WATCHDOG_TIMEOUT_SECS` was `120`, so progressing sweeps were killed mid-Curator. Raised to `300` — clear of the observed window, still an order of magnitude below the 2700s review-stall timeout, so the three backstops stay well separated.
2. **Progress was not latched.** `sweep_made_progress` re-derives progress from mutable filesystem state (worktree / checkpoint / log) every tick, and **all three signals are torn down at successful completion** (`merge-pr.sh` removes the worktree, `/loom:sweep` deletes the checkpoint, the log never grows past the spawn header). So a *finished* sweep read as *never-started*, and the memoryless `watchdog_decision` re-dispatched it against its own — now closed — issue (the 1135s #4078 incident). Added a per-`SweepId` latch so "has progressed" is monotonic.
3. **Closed-issue guard.** `dispatch` now skips an issue that is closed on the forge (best-effort, **fail-open** on a `gh` error). Placed in `dispatch` so it covers all three watchdogs (#3887 / #3895 / #3910), which all re-dispatch through it.

## Changes

- `loom-daemon/src/sweep_registry.rs`
  - `DEFAULT_WATCHDOG_TIMEOUT_SECS` `120` → `300`.
  - New private `watchdog_progressed: HashSet<SweepId>` latch next to `watchdog_retried` / `watchdog_gaveup`; initialized in both constructors; consulted + populated in `watchdog_once` around the `sweep_made_progress` call; pruned in the `reap_once` GC loop.
  - **Keyed by `SweepId`, not issue** (Finding 6): an issue-keyed latch would make a re-dispatched sweep that hangs read as "already progressed" and never be rescued (defangs the watchdog, violates AC2).
  - New `issue_is_closed` helper + closed-issue pre-flight guard in `dispatch` (fail-open).
  - **Unchanged:** `watchdog_decision` (pure, four existing tests) and `sweep_made_progress`'s three probes — only the statelessness was the defect.
- `.loom/docs/daemon-reference.md` — updated the `timeoutSecs` default (config example + knob table) and the startup-watchdog prose to document the per-sweep latch and the closed-issue skip.

1. Healthy sweep not cancelled — 300s default clears the 110–150s window. ✅ (`resolve_watchdog_timeout_and_interval_precedence`)
2. Genuinely hung sweep still cancelled, **including on re-dispatch** — SweepId-scoped latch. ✅ (`watchdog_latch_is_scoped_by_sweep_id_so_redispatch_is_still_rescued`, `watchdog_restarts_hung_sweep_once_then_gives_up` unchanged)
3. Signal documented in `daemon-reference.md`. ✅
4. "Slow but healthy startup" covered. ✅ (raised default + latch tests)
5. Completed sweep (worktree + checkpoint gone) not re-dispatched. ✅ (`watchdog_does_not_redispatch_completed_sweep_after_cleanup` — reproduces the #4078 cleanup condition)
6. Re-dispatch skipped for a closed issue, all three watchdogs. ✅ (guard in shared `dispatch`; `dispatch_refuses_closed_issue_without_flipping_labels`, `dispatch_fails_open_when_issue_state_lookup_errors`)

## Test Plan

- `cargo test --lib` latch suite: AC5 regression, AC2 re-dispatch rescue, monotonicity + per-sweep scope, latch pruning — all pass.
- Closed-issue guard: refuse-without-flipping + fail-open — pass.
- Regression: `watchdog_decision` (4), `midbuild` (10), `review_stall` (7), `log_has_progress` (4), `sweep_made_progress` (1) all pass unmodified. `cargo clippy --lib` clean, `cargo fmt` applied.

## Notes

- **Scope:** 2 files. Test code is the bulk of the diff (the Test Plan mandated 5+ latch/guard tests plus a fake-`gh` helper); production change is ~100 lines localized to `sweep_registry.rs`. Touched none of the work-finder / quarantine (#3939) / `spawn-claude.sh` layers the scope guard warns about.
- **Rollout:** daemon-side Rust — needs a **rebuild + `loom-daemon-start.sh`** to take effect; a run against the stale binary is not evidence the fix failed.
- **#4110 / #4111** also touch `dispatch`; no code overlap expected (this adds one pre-flight guard at step 2.5), but flag for merge-conflict check.

Closes #4088

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1382,7 +1382,7 @@ concurrency ceiling 5" and share it with the team:
     "dispatchStaggerMs": 2000,
     "watchdog": {
       "enabled": true,
-      "timeoutSecs": 120,
+      "timeoutSecs": 300,
       "intervalSecs": 30,
       "reviewStall": true,
       "reviewStallTimeoutSecs": 2700
@@ -1422,7 +1422,7 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.watchMonitor.expirySecs` | `LOOM_WATCH_MONITOR_EXPIRY_SECS` | `86400` | Give-up window for an unresolved watch; `0` disables expiry |
 | `autonomous.dispatchStaggerMs` | `LOOM_SWEEP_DISPATCH_STAGGER_MS` | `2000` | Min gap between consecutive child spawns (#3887). `0` disables |
 | `autonomous.watchdog.enabled` | `LOOM_SWEEP_WATCHDOG` | `true` | Startup watchdog on/off (#3887). Also the master switch for the tick — mid-build-death (#3895) + review-stall (#3910) run in the same task |
-| `autonomous.watchdog.timeoutSecs` | `LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` | `120` | No-progress window before auto-restart |
+| `autonomous.watchdog.timeoutSecs` | `LOOM_SWEEP_WATCHDOG_TIMEOUT_SECS` | `300` | No-progress window before auto-restart (raised from 120s in #4088 — the old default sat inside the observed 110–150s dispatch→worktree window and cancelled healthy sweeps) |
 | `autonomous.watchdog.intervalSecs` | `LOOM_SWEEP_WATCHDOG_INTERVAL_SECS` | `30` | Watchdog probe cadence (shared by all three backstops) |
 | `autonomous.watchdog.reviewStall` | `LOOM_SWEEP_REVIEW_STALL` | `true` | Review-phase stall watchdog on/off (#3910) |
 | `autonomous.watchdog.reviewStallTimeoutSecs` | `LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS` | `2700` | Log-silence window before a hung Judge/Doctor sweep is re-dispatched |
@@ -1506,6 +1506,23 @@ written / log output past the spawn header) and auto-cancels + re-dispatches —
 frozen `sweep.issue.{N}.exited` / `sweep.global.completed` / `sweep.global.dispatch`
 topics (no new event topics). The watchdog defaults **on**; disable it with
 `LOOM_SWEEP_WATCHDOG=0` or `autonomous.watchdog.enabled = false`.
+
+**Progress is latched per sweep (#4088).** The progress probe reads only the
+*current* filesystem state (worktree / checkpoint / log), and **all three signals
+are torn down at successful completion** — `merge-pr.sh` removes the worktree,
+`/loom:sweep` deletes the checkpoint, and the log never grows past the spawn
+header. So a *finished* sweep is structurally indistinguishable from one that
+*never started*, and because the decision function is memoryless (`elapsed` is
+total runtime and only increases), a completed-then-cleaned-up sweep would be
+re-dispatched against its own — now closed — issue. To fix this the watchdog
+**latches progress per `SweepId`**: once a sweep is ever observed making progress
+it is left alone for the rest of its life (later crashes are delegated to the
+mid-build-death / review-stall backstops below). The latch is keyed by `SweepId`,
+not issue number, so a *re-dispatched* sweep that then genuinely hangs at startup
+is still rescued. Independently, **`dispatch` skips any issue that is closed on
+the forge** (best-effort, fail-open on a `gh` error) — this covers all three
+watchdogs, since each re-dispatches through the same method, so no self-heal path
+can ever re-claim a closed/merged issue.
 
 **Review-phase stall watchdog (#3910).** The startup watchdog rescues a sweep
 that shows *no* progress, and the mid-build-death watchdog (#3895) rescues one
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -443,7 +443,13 @@ pub const WATCHDOG_INTERVAL_ENV: &str = "LOOM_SWEEP_WATCHDOG_INTERVAL_SECS";
 /// written no checkpoint, and produced no log output past the spawn header
 /// within this window is treated as hung. Generous enough that a healthy sweep
 /// (which emits Curator-phase output well inside two minutes) never trips it.
-pub const DEFAULT_WATCHDOG_TIMEOUT_SECS: u64 = 120;
+///
+/// Raised from 120s to 300s (Issue #4088): under concurrency the normal
+/// dispatch→worktree latency was measured at 110–150s, so the old 120s default
+/// sat *inside* the healthy distribution and cancelled progressing sweeps. 300s
+/// clears that window with headroom while staying an order of magnitude below
+/// the review-stall timeout (2700s), keeping the three backstops well separated.
+pub const DEFAULT_WATCHDOG_TIMEOUT_SECS: u64 = 300;
 
 /// Default watchdog probe interval — matches the reaper cadence.
 pub const DEFAULT_WATCHDOG_INTERVAL_SECS: u64 = 30;
@@ -1065,6 +1071,22 @@ pub struct SweepRegistry {
     /// Issues the watchdog has already logged a give-up for, so the loud
     /// give-up warning fires once per issue rather than every tick.
     watchdog_gaveup: HashSet<u32>,
+    /// Sweeps the startup watchdog has ever observed making progress, latched so
+    /// the "has progressed" signal is monotonic (Issue #4088). `sweep_made_progress`
+    /// re-derives progress from mutable filesystem state (worktree / checkpoint /
+    /// log) every tick, all of which are torn down at successful completion — so a
+    /// *finished* sweep reads as *never started* and the memoryless
+    /// [`watchdog_decision`] re-dispatches it. Once a `SweepId` lands here the
+    /// startup watchdog leaves it alone forever, delegating any later crash to the
+    /// mid-build-death (#3895) / review-stall (#3910) backstops, which is the
+    /// division of labor the module doc already specifies.
+    ///
+    /// Keyed by `SweepId`, NOT issue: an issue-keyed latch would persist across
+    /// re-dispatch, so a re-dispatched sweep that then genuinely hangs at startup
+    /// would read as "already progressed" and never be rescued — silently
+    /// defanging the watchdog and violating AC2. Grows per dispatch, so it is
+    /// pruned alongside entry GC in [`reap_once`](Self::reap_once).
+    watchdog_progressed: HashSet<SweepId>,
     /// Issues the mid-build-death watchdog has already recovered once (Issue
     /// #3895). Distinct from `watchdog_retried` (startup-hang, no progress):
     /// this bounds the "made progress then the child died" recovery to a
@@ -1113,6 +1135,7 @@ impl SweepRegistry {
             last_spawn_at: None,
             watchdog_retried: HashSet::new(),
             watchdog_gaveup: HashSet::new(),
+            watchdog_progressed: HashSet::new(),
             midbuild_retried: HashSet::new(),
             midbuild_gaveup: HashSet::new(),
             review_stall_retried: HashSet::new(),
@@ -1135,6 +1158,7 @@ impl SweepRegistry {
             last_spawn_at: None,
             watchdog_retried: HashSet::new(),
             watchdog_gaveup: HashSet::new(),
+            watchdog_progressed: HashSet::new(),
             midbuild_retried: HashSet::new(),
             midbuild_gaveup: HashSet::new(),
             review_stall_retried: HashSet::new(),
@@ -1344,6 +1368,24 @@ impl SweepRegistry {
                 ));
             }
         };
+
+        // 2.5 Closed-issue guard (Issue #4088). All three watchdogs
+        //     (startup #3887, mid-build-death #3895, review-stall #3910)
+        //     re-dispatch through this method, and `gh issue edit` succeeds on a
+        //     closed issue, so nothing else stops a watchdog false-positive from
+        //     re-claiming an issue whose PR already merged. Placing the guard
+        //     here — before the lock/label flip — covers all three call sites
+        //     with one check. Best-effort and fail-open: a forge lookup error
+        //     returns `None` and dispatch proceeds, so a `gh` outage can never
+        //     wedge the daemon. Skipped when label flips are disabled (test
+        //     fixtures without `gh` credentials).
+        if !self.config.skip_label_flip && self.issue_is_closed(issue_number) == Some(true) {
+            return Err(anyhow!(
+                "refusing to dispatch issue #{issue_number}: it is closed on the forge \
+                 (#4088 closed-issue guard). A watchdog re-dispatch must not re-claim a \
+                 closed/merged issue."
+            ));
+        }
 
         // 3. Acquire the claim lock atomically.
         let sweep_id = generate_sweep_id(kind);
@@ -1589,6 +1631,48 @@ impl SweepRegistry {
     // ------------------------------------------------------------------------
     // Forge label flip
     // ------------------------------------------------------------------------
+
+    /// Best-effort probe of whether an issue is closed on the forge (Issue
+    /// #4088). Returns `Some(true)` when closed, `Some(false)` when open, and
+    /// `None` on any error (missing/failed/timed-out `gh`, unparseable output).
+    ///
+    /// Callers MUST treat `None` as **fail-open** — a forge outage or a wedged
+    /// `gh` must never wedge dispatch. The call is bounded by [`reap_gh_timeout`]
+    /// exactly like the label flips so it cannot block the dispatch path.
+    fn issue_is_closed(&self, issue: u32) -> Option<bool> {
+        let gh = self
+            .config
+            .gh_bin
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("gh"));
+        let mut cmd = Command::new(&gh);
+        cmd.arg("issue")
+            .arg("view")
+            .arg(issue.to_string())
+            .arg("--json")
+            .arg("state")
+            .arg("--jq")
+            .arg(".state");
+        // Resolve the issue against this registry's own workspace, matching the
+        // label-flip helpers (#3937). LOOM_REPO still overrides when set.
+        cmd.current_dir(&self.config.workspace_root);
+        if let Ok(repo) = std::env::var("LOOM_REPO") {
+            cmd.arg("--repo").arg(repo);
+        }
+        let output = output_with_timeout(cmd, reap_gh_timeout()).ok()??;
+        if !output.status.success() {
+            return None;
+        }
+        match String::from_utf8_lossy(&output.stdout)
+            .trim()
+            .to_ascii_uppercase()
+            .as_str()
+        {
+            "CLOSED" => Some(true),
+            "OPEN" => Some(false),
+            _ => None,
+        }
+    }
 
     fn flip_label_to_building(&self, issue: u32) -> Result<()> {
         let gh = self
@@ -2712,6 +2796,11 @@ impl SweepRegistry {
             // `poll_liveness` already, but drop any lingering handle so a
             // GC'd sweep never leaks a `Child` (Issue #3801).
             let _ = self.children.remove(&id);
+            // Prune the per-SweepId progress latch so it cannot grow unbounded
+            // across many dispatches (Issue #4088). Safe because a GC'd entry is
+            // terminal — the watchdog only ever consults the latch for
+            // Running/Pending entries it still owns a Child handle for.
+            self.watchdog_progressed.remove(&id);
             changes += 1;
         }
         changes
@@ -2810,7 +2899,17 @@ impl SweepRegistry {
 
         let mut restarts = 0usize;
         for (sweep_id, issue, log_path, elapsed) in candidates {
-            let made_progress = self.sweep_made_progress(issue, &log_path);
+            // Latch progress per SweepId (Issue #4088): `sweep_made_progress`
+            // only reports the *current* filesystem state, and every signal it
+            // reads (worktree, checkpoint, log) is torn down at successful
+            // completion — so a finished sweep would otherwise read as
+            // never-started and be re-dispatched. Once observed true, the latch
+            // keeps `made_progress` true for this SweepId on every later tick.
+            let made_progress = self.watchdog_progressed.contains(&sweep_id)
+                || self.sweep_made_progress(issue, &log_path);
+            if made_progress {
+                self.watchdog_progressed.insert(sweep_id.clone());
+            }
             let already_retried = self.watchdog_retried.contains(&issue);
             match watchdog_decision(elapsed, timeout, made_progress, already_retried) {
                 WatchdogDecision::Healthy => {}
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the consolidated 6-item AC list from #4088 (the Curator-enhanced body supersedes the original 4-item list).

## Verification

- **Code correctness** — walked the diff and the surrounding `dispatch`/`watchdog_once` context directly (not just the patch hunks):
  - `watchdog_progressed: HashSet<SweepId>` is keyed by `SweepId`, not issue number, as Finding 6 requires — confirmed the AC2-on-re-dispatch trap is actually tested (`watchdog_latch_is_scoped_by_sweep_id_so_redispatch_is_still_rescued`), not just documented.
  - Latch is monotonic (`self.watchdog_progressed.contains(&sweep_id) || self.sweep_made_progress(...)`, then inserted if true) and pruned in `reap_once` alongside entry GC — no unbounded growth.
  - `issue_is_closed` closed-issue guard sits at dispatch step "2.5", before the lock is acquired (step 3) and before the label flip (step 4), so a refused dispatch never touches `loom:issue`/`loom:building` — matches the `dispatch_refuses_closed_issue_without_flipping_labels` test's assertion that no `issue edit` call occurs. It follows the exact `gh_bin`/`workspace_root`/`LOOM_REPO` pattern used by `flip_label_to_building`, and fails open (`Some(true)` only on a clean `CLOSED`, `None` on any error/timeout) per `dispatch_fails_open_when_issue_state_lookup_errors`. Correctly gated on `!skip_label_flip` (default `false` in production, so the guard is live).
  - `watchdog_decision` and `sweep_made_progress`'s three probes are untouched, as the issue mandated (kept pure for the 4 pre-existing tests).

- **Local test run** (worktree `.loom/worktrees/issue-4088`, not just CI): `cargo test --lib sweep_registry::` — **145 passed, 0 failed**, including all 6 new tests (AC5 completed-sweep regression, AC2 re-dispatch scoping, latch monotonicity, latch pruning, AC1 300s default, and both closed-issue guard tests run individually to confirm they weren't silently skipped by a tail-truncated log). `cargo fmt --check` clean.

- **CI**: all required checks pass — Rust Linting & Build Check (clippy) `pass`, Rust Unit Tests `pass` (3m40s, full workspace suite), CLAUDE.md budget `pass`. `mergeStateStatus` is `CLEAN`, `mergeable`.

- **Docs (AC3)**: `daemon-reference.md` config table + prose updated to describe the per-`SweepId` latch and the closed-issue skip — matches the implementation.

- **Scope**: 2 files as declared. Production code is ~100 LOC in `sweep_registry.rs`; the bulk of the diff is the 6+ tests the issue's own Test Plan mandated (including a fake-`gh` helper for the guard tests). No spillover into the work-finder, quarantine (#3939), or `spawn-claude.sh`, as the scope guard warned against.

## Note (non-blocking)

Issue #4088 currently carries `loom:blocked` from an unrelated daemon auto-quarantine (3 insta-crashes on dispatch, per the bot comment) — that's an orchestration-layer artifact of running the fix against the *stale, unrolled* daemon binary the PR itself calls out in its rollout note, not a defect in this diff. Worth clearing quarantine (`loom-daemon quarantine clear 4088`) after this merges and the daemon is rebuilt/restarted, so the issue can close cleanly via `Closes #4088`.

Nice work tracing the two independent defects (threshold-inside-distribution vs. non-latched progress) separately and adding the SweepId-vs-issue-keying regression test — that's exactly the kind of test that would have silently passed with the wrong (issue-keyed) implementation.

PR #4119, opened 2026-07-27
fix(daemon): durably release insta-crash quarantine labels

### LINKED ISSUE #4110: fix(daemon): quarantine TTL auto-release strands loom:blocked on the forge — issues silently leave the queue forever

rjwalters:
## Summary

Implements the Curator's corrected root-cause analysis on #4110: the TTL
already restored `loom:issue` on current main, but two real, unfixed bugs
still strand `loom:blocked` issues forever — a best-effort release that drops
in-memory state before attempting (and never retries) the `gh` label flip,
and no startup reconciliation for a daemon restart (which always wipes the
memory-only quarantine map). A third gap — workspace eviction dropping a
live quarantine's only retry path — is fixed as well, since it was small and
directly analogous to the restart case.

## Changes

- `loom-daemon/src/sweep_registry.rs`
  - `release_quarantine_label` now checks `output.status.success()` instead
    of treating any completed `gh` invocation as success, returns a `bool`,
    and logs failures at `warn` (was `debug`) — the previous code silently
    treated a non-zero `gh` exit identically to success.
  - `expire_quarantine` / `clear_quarantine` now route through a new
    `attempt_quarantine_release` helper: on failure the issue is added to a
    new `pending_quarantine_release: HashSet<u32>` field instead of being
    dropped; `reap_once` retries every pending entry each tick
    (`retry_pending_quarantine_releases`) until the flip succeeds.
  - New `flush_quarantines_for_eviction` releases every live quarantine
    before a registry is discarded.
  - Extracted `QUARANTINE_COMMENT_MARKER` const and fixed the posted
    comment's wording: both the CLI clear **and** the TTL restore
    `loom:issue` (previously only the CLI path was described that way).
- `loom-daemon/src/quarantine_reconciliation.rs` (new) — startup
  reconciliation pass mirroring `claim_reconciliation`'s conventions (kill
  switch, per-workspace issue cap, pure `decide`/`plan` split). Scans open
  `loom:blocked` issues per workspace and releases the ones whose comments
  contain `QUARANTINE_COMMENT_MARKER` back to `loom:issue`; an issue with no
  such comment (an operator's manual block) is never touched. Unlike
  `claim_reconciliation` there's no PID-liveness question — a fresh daemon's
  quarantine memory is always empty at startup, so the marker alone decides.
- `loom-daemon/src/main.rs` — wires the new pass in beside the existing
  `claim_reconciliation` startup call.
- `loom-daemon/src/workspace_pool.rs` — `evict()` calls
  `flush_quarantines_for_eviction` before aborting the registry's reaper.
- `loom-daemon/src/lib.rs` — registers the new module.

|---|---|---|
| A `gh` failure/timeout during release does not permanently strand the issue; failure logged at `warn`+ | ✅ | `release_quarantine_label` returns `bool` on exit-status check, logs at `warn`; `quarantine_release_retries_after_gh_failure_then_succeeds` test drives a scripted 2-failure-then-success fake `gh` through two `reap_once` ticks and asserts the pending set + eventual success |
| A daemon restart does not permanently strand quarantined issues — startup pass reconciles `loom:blocked` issues with a quarantine comment but no live quarantine | ✅ | New `quarantine_reconciliation` module + `main.rs` wiring; `reconcile_workspace_releases_issue_with_quarantine_comment` asserts the real `gh issue edit --remove-label loom:blocked --add-label loom:issue` argv fires |
| Startup pass is bounded (issue cap) and has a kill-switch env var | ✅ | `MAX_ISSUES_PER_WORKSPACE` passed to `gh issue list --limit`; `LOOM_QUARANTINE_RECONCILE` kill switch — `reconcile_workspace_passes_issue_cap_to_gh_list`, `reconciliation_enabled_resolves_env_precedence` |
| Workspace eviction does not silently strand a quarantined issue | ✅ | `flush_quarantines_for_eviction` called from `WorkspacePool::evict`; `flush_quarantines_for_eviction_releases_and_clears_state` |
| Quarantine comment states both release paths restore `loom:issue` | ✅ | Comment body rewritten in `apply_quarantine_label`; read the updated format string |
| Idempotence preserved — releasing a non-quarantined issue is a no-op success; repeated flip is harmless | ✅ | Existing `clear_quarantine_noop_skips_forge_label` / `ipc.rs:test_handle_request_clear_quarantine_noop` still pass; `quarantine_expiry_flips_real_forge_label_via_argv` asserts a later tick with nothing pending makes no further `gh` calls |
| ~~TTL expiry restores `loom:issue`~~ (already true on `main`) — add the missing test | ✅ | `quarantine_expiry_flips_real_forge_label_via_argv` is the first test to exercise the real (non-`skip_label_flip`) TTL release argv |

## Test Plan

- `cargo test -p loom-daemon --lib quarantine` — 27 passed (0 failed)
- `cargo test -p loom-daemon --lib` (full crate) — 1212 passed, 0 failed
- `cargo test -p loom-daemon --lib workspace_pool` — 7 passed (existing eviction tests unaffected — an empty quarantine map short-circuits the new flush with zero `gh` calls)
- `cargo clippy -p loom-daemon --lib --tests -- -D warnings` — clean

## Scope note

The issue's implementation guidance set a soft ~400 LOC target and an
explicit escape hatch to defer the reconciliation pass if it alone would
blow that budget. The total diff (main.rs/workspace_pool.rs/lib.rs wiring +
sweep_registry.rs release-path fix + the new quarantine_reconciliation
module, tests included) came in over that number, but the reconciliation
module itself is much leaner than its `claim_reconciliation` sibling (no
PID-liveness question — a fresh daemon's quarantine memory is always empty
at startup, so the daemon-authored comment marker alone decides), and a
large share of the total is rustdoc matching this repo's existing
doc-per-function convention plus the real-argv test coverage the issue's own
Test Plan called out as a pre-existing gap. Given all four steps are small,
mutually reinforcing, and independently well-tested, I landed them together
rather than splitting step 2 into a follow-up.

Closes #4110

### DIFF
--- a/loom-daemon/src/lib.rs
+++ b/loom-daemon/src/lib.rs
@@ -32,6 +32,7 @@ pub mod main_health_gate;
 pub mod metrics_collector;
 pub mod phase_join;
 pub mod pipeline_snapshot;
+pub mod quarantine_reconciliation;
 pub mod role_runner;
 pub mod role_validation;
 pub mod self_update;
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -7,6 +7,7 @@ use loom_daemon::health_monitor;
 use loom_daemon::ipc::IpcServer;
 use loom_daemon::main_health_gate;
 use loom_daemon::metrics_collector;
+use loom_daemon::quarantine_reconciliation;
 use loom_daemon::role_runner;
 use loom_daemon::role_validation;
 use loom_daemon::self_update;
@@ -661,6 +662,50 @@ async fn main() -> Result<()> {
         log::info!(
             "claim_reconciliation: startup pass disabled ({}=0)",
             claim_reconciliation::RECONCILE_ENABLED_ENV
+        );
+    }
+
+    // Stranded-quarantine reconciliation across every managed workspace
+    // (Issue #4110). The insta-crash quarantine (#3939) is memory-only, so a
+    // restart drops the in-memory pause while the `loom:blocked` label it
+    // applied survives on the forge — with nothing left to release it, the
+    // issue is permanently invisible to the work finder. This pass scans
+    // every registered workspace's open `loom:blocked` issues and releases
+    // the ones carrying a daemon-authored quarantine comment back to
+    // `loom:issue`; a human's manual `loom:blocked` (no such comment) is
+    // never touched. Reuses the same `workspace_registry` roots resolved
+    // above for claim reconciliation.
+    if quarantine_reconciliation::reconciliation_enabled() {
+        let workspace_registry =
+            loom_daemon::workspace_registry::WorkspaceRegistry::load_default().unwrap_or_default();
+        let roots = workspace_registry.effective_roots(&sweep_workspace);
+        let gh_bin = std::path::PathBuf::from("gh");
+        let mut total_checked = 0usize;
+        let mut total_released = 0usize;
+        for root in &roots {
+            let (checked, released) =
+                quarantine_reconciliation::forge::reconcile_workspace(&gh_bin, root);
+            total_checked += checked;
+            total_released += released;
+        }
+        if total_released > 0 {
+            log::info!(
+                "quarantine_reconciliation: startup pass checked {total_checked} loom:blocked \
+                 issue(s) across {} workspace(s), released {total_released} stranded \
+                 quarantine(s) (#4110)",
+                roots.len()
+            );
+        } else {
+            log::debug!(
+                "quarantine_reconciliation: startup pass checked {total_checked} loom:blocked \
+                 issue(s) across {} workspace(s), nothing to release",
+                roots.len()
+            );
+        }
+    } else {
+        log::info!(
+            "quarantine_reconciliation: startup pass disabled ({}=0)",
+            quarantine_reconciliation::RECONCILE_ENABLED_ENV
         );
     }
 
--- a/loom-daemon/src/quarantine_reconciliation.rs
+++ b/loom-daemon/src/quarantine_reconciliation.rs
@@ -1,0 +1,418 @@
+//! Daemon-startup reconciliation of stranded `loom:blocked` insta-crash
+//! quarantines across every managed workspace (Issue #4110).
+//!
+//! The insta-crash quarantine (#3939) is memory-only:
+//! [`crate::sweep_registry::SweepRegistry`]'s `quarantined` map never survives
+//! a daemon restart. The forge label (`loom:blocked`) it applied, however,
+//! does survive — so a restart drops the in-memory pause while leaving the
+//! label behind, and nothing scans for that afterward. The `loom:blocked`
+//! issue is then invisible to the work finder (which only polls open
+//! `loom:issue`) with no quarantine left anywhere to clear, forever.
+//!
+//! This module is the startup pass that closes that gap: for every registered
+//! workspace, list the open `loom:blocked` issues and release the ones that
+//! carry a daemon-authored quarantine comment (identified by
+//! [`crate::sweep_registry::QUARANTINE_COMMENT_MARKER`]) back to `loom:issue`.
+//! An issue a human deliberately blocked by hand carries no such comment and
+//! is never touched — this pass only ever undoes its own past mutation.
+//!
+//! Mirrors [`crate::claim_reconciliation`]'s conventions: a kill-switch env
+//! var, a per-workspace issue cap, and a pure `decide`/`plan` split that keeps
+//! the decision logic unit-testable without a forge. Unlike
+//! `claim_reconciliation`, there is no liveness question to answer here — a
+//! fresh daemon's quarantine memory is *always* empty at startup, so the sole
+//! question is "did the daemon itself apply this `loom:blocked` label", which
+//! the comment marker answers directly.
+
+/// Env var to disable the startup quarantine reconciliation pass entirely
+/// (kill switch, not a feature gate — this is corrective crash recovery in
+/// the same spirit as [`crate::claim_reconciliation::RECONCILE_ENABLED_ENV`],
+/// so it defaults to ON). `0`/`false`/`no`/`off` disables; anything else
+/// (including unset) leaves it enabled.
+pub const RECONCILE_ENABLED_ENV: &str = "LOOM_QUARANTINE_RECONCILE";
+
+/// Bound on how many `loom:blocked` issues one reconciliation pass inspects
+/// per workspace (defense in depth against an unexpectedly huge backlog
+/// turning startup into a `gh`-API storm), matching
+/// [`crate::claim_reconciliation::MAX_ISSUES_PER_WORKSPACE`].
+pub const MAX_ISSUES_PER_WORKSPACE: u32 = 100;
+
+/// Resolve whether the startup quarantine reconciliation pass is enabled.
+#[must_use]
+pub fn reconciliation_enabled() -> bool {
+    match std::env::var(RECONCILE_ENABLED_ENV) {
+        Ok(v) => !matches!(v.trim().to_ascii_lowercase().as_str(), "0" | "false" | "no" | "off"),
+        Err(_) => true,
+    }
+}
+
+/// A `loom:blocked` issue reported by the forge, trimmed to the fields the
+/// reconciliation decision needs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlockedIssue {
+    pub number: u32,
+    /// Whether any comment on the issue contains
+    /// [`crate::sweep_registry::QUARANTINE_COMMENT_MARKER`].
+    pub has_quarantine_comment: bool,
+}
+
+/// The reconciliation decision for one issue.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ReconcileAction {
+    /// No daemon quarantine comment found — leave the claim alone. This is
+    /// the fail-safe branch: a human's manual `loom:blocked` is never
+    /// touched.
+    Keep,
+    /// Flip `loom:blocked` back to `loom:issue`: the label was applied by a
+    /// daemon quarantine that could not have survived to this restart.
+    Release,
+}
+
+/// Pure decision function — no I/O, fully unit-testable. See the module docs
+/// for the decision rule.
+#[must_use]
+pub fn decide(issue: &BlockedIssue) -> ReconcileAction {
+    if issue.has_quarantine_comment {
+        ReconcileAction::Release
+    } else {
+        ReconcileAction::Keep
+    }
+}
+
+/// Plan reconciliation decisions for every issue in `issues`. Performs no
+/// I/O; `issues` is expected to already be capped to
+/// [`MAX_ISSUES_PER_WORKSPACE`] by the caller.
+#[must_use]
+pub fn plan(issues: &[BlockedIssue]) -> Vec<(u32, ReconcileAction)> {
+    issues
+        .iter()
+        .map(|issue| (issue.number, decide(issue)))
+        .collect()
+}
+
+/// `gh`/label-flip glue. Not unit-tested directly (mirrors
+/// [`crate::claim_reconciliation::forge`] / [`crate::work_finder::forge`]) —
+/// the decision logic above is the fully-covered surface; this module is a
+/// thin, best-effort `Command` wrapper.
+pub mod forge {
+    use super::{plan, BlockedIssue, ReconcileAction, MAX_ISSUES_PER_WORKSPACE};
+    use crate::sweep_registry::QUARANTINE_COMMENT_MARKER;
+    use anyhow::{anyhow, Context, Result};
+    use serde::Deserialize;
+    use std::path::Path;
+    use std::process::{Command, Stdio};
+
+    #[derive(Debug, Deserialize)]
+    struct GhComment {
+        #[serde(default)]
+        body: String,
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct GhBlockedIssue {
+        number: u32,
+        #[serde(default)]
+        comments: Vec<GhComment>,
+    }
+
+    fn list_blocked_issues(gh_bin: &Path, root: &Path) -> Result<Vec<BlockedIssue>> {
+        let mut cmd = Command::new(gh_bin);
+        cmd.arg("issue")
+            .arg("list")
+            .arg("--label")
+            .arg("loom:blocked")
+            .arg("--state")
+            .arg("open")
+            .arg("--limit")
+            .arg(MAX_ISSUES_PER_WORKSPACE.to_string())
+            .arg("--json")
+            .arg("number,comments");
+        cmd.current_dir(root);
+        if let Ok(repo) = std::env::var("LOOM_REPO") {
+            cmd.arg("--repo").arg(repo);
+        }
+        cmd.stdout(Stdio::piped()).stderr(Stdio::piped());
+        let out = cmd
+            .output()
+            .with_context(|| format!("failed to invoke {}", gh_bin.display()))?;
+        if !out.status.success() {
+            return Err(anyhow!(
+                "gh issue list --label loom:blocked failed in {}: {}",
+                root.display(),
+                String::from_utf8_lossy(&out.stderr).trim()
+            ));
+        }
+        let rows: Vec<GhBlockedIssue> =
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the corrected root-cause analysis and acceptance criteria on #4110 (three real bugs: best-effort release with no retry, no restart reconciliation, no eviction release).

## Verification

**1. `gh`-failure release is retried, not silently dropped.**
`release_quarantine_label` now checks `output.status.success()` (previously any completed invocation, including non-zero exit, was treated as success) and returns `bool`; failures log at `warn` (was `debug`). `attempt_quarantine_release` adds a failed issue to the new `pending_quarantine_release: HashSet<u32>`, and `retry_pending_quarantine_releases` drains it on every `reap_once` tick until the flip succeeds. Confirmed by `quarantine_release_retries_after_gh_failure_then_succeeds`, which scripts a fake `gh` to fail twice then succeed and asserts exactly 3 real `gh issue edit ... --remove-label loom:blocked --add-label loom:issue` invocations across two reaper ticks.

**2. Startup reconciliation is bounded, kill-switched, and never touches a human's manual `loom:blocked`.**
New `quarantine_reconciliation.rs` mirrors `claim_reconciliation.rs`'s conventions closely (pure `decide`/`plan` split, `LOOM_QUARANTINE_RECONCILE` kill switch defaulting to on, `MAX_ISSUES_PER_WORKSPACE = 100` passed straight through to `gh issue list --limit`). `decide()` releases only when the issue's comments contain `QUARANTINE_COMMENT_MARKER`; `reconcile_workspace_never_touches_manual_block` asserts zero `gh issue edit` calls for a `loom:blocked` issue with an unrelated comment. Wired into `main.rs` beside the existing `claim_reconciliation` call, same `effective_roots()` iteration.

**3. Eviction releases outstanding quarantines.**
`SweepRegistry::flush_quarantines_for_eviction()` drains the `quarantined` map and attempts a real release for each before `WorkspacePool::evict()` aborts the reaper. `flush_quarantines_for_eviction_releases_and_clears_state` asserts the real argv fires and in-memory state clears.

**4. Idempotence preserved.**
Existing `clear_quarantine_noop_skips_forge_label` / `ipc.rs::test_handle_request_clear_quarantine_noop` still pass unmodified. `quarantine_expiry_flips_real_forge_label_via_argv` additionally asserts a second `reap_once` tick with nothing pending makes **zero** further `gh` calls — a released issue isn't re-flipped.

**5. The real-argv coverage gap called out in the issue is closed.** All three new forge-touching tests (`quarantine_expiry_flips_real_forge_label_via_argv`, `quarantine_release_retries_after_gh_failure_then_succeeds`, `flush_quarantines_for_eviction_releases_and_clears_state`, plus the two `quarantine_reconciliation` forge tests) run with `skip_label_flip = false` and a `gh_bin` pointed at a real recording script, asserting the literal argv — not the `skip_label_flip` fixture path the issue flagged as untested.

**6. Comment text fixed** — the quarantine comment now states both the CLI clear *and* the TTL restore `loom:issue`, matching reality on `main`.

## Test Execution

- `cargo test --lib quarantine` (worktree) — 27 passed, 0 failed
- `cargo test --lib` (full crate) — 1212 passed, 0 failed
- `cargo clippy --lib --tests -- -D warnings` — clean
- `git merge-tree --write-tree origin/main HEAD` — no conflicts
- CI: Rust Linting & Build Check ✅, Rust Unit Tests ✅, CLAUDE.md Line Budget ✅ — all pass; `mergeStateStatus` is `CLEAN`

## Scope note

Diff is 797/-23 vs. the issue's soft ~400 LOC target. Reviewed the actual content, not just the disclosure: the extra size is concentrated in rustdoc (matching this file's existing doc-per-function convention) and the new real-argv test suite the issue explicitly asked for — the production logic itself (the retry set, the reconciliation `decide`/`plan`, the eviction flush) is compact and closely mirrors `claim_reconciliation.rs`. Landing all four steps together was reasonable given how small and mutually reinforcing they are; splitting step 2 into a follow-up would have added coordination overhead for no real risk reduction.

No blocking issues found.

Issue #4037, opened 2026-07-27
docs: CLAUDE.md scheduled-support-role table still shows bare /champion, /curator, etc. after #4036


rjwalters:
## Original Issue

While reviewing PR #4036 (which fixes the daemon role runner + GitHub Actions workflows to invoke the namespaced `/loom:<role>` slash commands instead of the bare `/<role>` form — see #4034), I noticed the "Scheduled Support Roles" table in `CLAUDE.md` and `defaults/CLAUDE.md` still lists the bare form as the invoked command:

```
| `loom-champion.yml` | `/champion` | `*/10 * * * *` |
| `loom-curator.yml`  | `/curator`  | `*/5 * * * *`  |
| `loom-judge.yml`    | `/judge`    | `*/5 * * * *`  |
| `loom-auditor.yml`  | `/auditor`  | `*/10 * * * *` |
| `loom-guide.yml`    | `/guide`    | `*/15 * * * *` |
```

(`CLAUDE.md:130-134`, `defaults/CLAUDE.md:237-241`.) After #4036 merges, the actual `.github/workflows/loom-*.yml` files invoke `/loom:champion`, `/loom:curator`, `/loom:judge`, `/loom:auditor`, `/loom:guide` — so this table (and any other place these literal bare names appear as "the command," as distinct from the generic `/<role>` placeholder used elsewhere in prose) is now stale and could mislead an operator hand-editing a workflow or explaining the mechanism.

## Suggested fix
Update the "Role" column in both tables to the namespaced form, and spot-check other docs (`docs/agents.md`, `docs/philosophy/loom-intelligence.md`, `docs/migration/v0.10.0-daemon-rebuild.md`, `.loom/docs/daemon-reference.md`, `docs/adr/0009-shepherd-deprecation.md`) for the same literal-bare-name pattern vs. the intentional `/<role>` placeholder shorthand (which is fine as-is).

Low severity — doc-only, no functional impact. Discovered during review of #4036.

---

### ⚠️ The primary cited location no longer exists

The issue cites `CLAUDE.md:130-134` as the main target. **That table was deleted from the root `CLAUDE.md` by #4089** (`2478d0d2`, "docs: split CLAUDE.md operating core from reference and history"), which landed after this issue was filed. The root `CLAUDE.md` §"4. Scheduled Support Roles" is now three prose paragraphs with no per-workflow table and no bare command names.

```bash
$ grep -n 'loom-champion.yml' CLAUDE.md      # → no match
[...]

PR #4120, opened 2026-07-27
fix(sweep): daemon-dispatched child must not self-skip its own claim

### LINKED ISSUE #4111: fix(sweep): daemon-dispatched child self-skips its own claim despite LOOM_SWEEP_CLAIM_OWNED being set — regression of #3823/#3967, strands issues via quarantine

rjwalters:
## Summary

`SweepRegistry::dispatch` correctly flips `loom:issue → loom:building` and exports `LOOM_SWEEP_CLAIM_OWNED=<N>` before spawning a daemon-dispatched `/loom:sweep` child, but the child's own pre-flight reliably **ignored the marker** and self-skipped its own claim — reasoning from the fresh `loom:building` label timeline as evidence of a competing worker (see #4111's transcripts for `#4037`/`#4069`/`#4110`, where the child even cited its own PID/sweep-id as "the other worker"). A self-skip exits cleanly in seconds with no checkpoint — the #3939 insta-crash shape — and three strikes quarantined already-approved, curated issues, stranding them via #4110.

Implements both options the issue recommends doing together, investigates the third per its explicit gating instructions, and adds the missing consumer-side test coverage.

## Option 1 — `--claim-owned <N>` positional flag

`SweepRegistry::spawn_child` (`loom-daemon/src/sweep_registry.rs`) now appends `--claim-owned <N>` to the child's own argv, unconditionally on every issue dispatch, in addition to the `LOOM_SWEEP_CLAIM_OWNED` env var (**kept, unchanged, for backward compatibility** — `spawn-claude.sh` still logs it and the #3967 producer tests still assert it). A flag baked into the invocation's own text is in the model's context by construction; an ambient env var proved not to be — #4111's transcripts show the child never even attempted to read it.

## Option 2 — mandatory "Step 1a" pre-flight step

`defaults/.claude/commands/loom/sweep.md`'s "1. Per-issue pre-flight" rule restructures the `#3823` self-claim exception from a trailing clause *after* the `loom:building` skip test into a **mandatory, first-evaluated Step 1a**, checked before the skip bullet for every issue in the wave. Either signal (the flag or the env var) naming the issue being pre-flighted is now conclusive and short-circuits further reasoning about label timing / PID tables / `loom-daemon status` — precisely the signals #4111's incident used to build a case against itself.

A new doc-lint test (`sweep_md_step_1a_self_claim_check_precedes_loom_building_skip_bullet` in `sweep_md_doc_lint.rs`) mechanically enforces the ordering (byte-offset comparison) so a future edit can't silently regress the restructure back into an exception clause. `sweep_md_stage_minus_one_doc_lint.rs` gets a companion test (`sweep_md_documents_claim_owned_flag`) asserting the flag is documented as an *equivalent* signal to the env var, not a silent replacement.

The Stage -1 backend-detection short-circuit (`#3829`, a **separate**, already-working consumer of the same env var) is updated to recognize the flag too, for consistency, though its logic was not the bug (evidence below).

## Option 3 — insta-crash classification (investigated, not changed)

Per the issue's explicit Finding 1 instruction ("first establish empirically what the reaper actually saw... do not implement AC3 by adding a second exemption before determining the observed exit code"):

**Exit-code evidence.** `.loom/logs/sweep-issue-4037.log` (still present on this host) shows the daemon dispatched issue `#4037` **four separate times** across ~1 hour, with self-skip transcripts at `21:36:44Z` and `22:26:14Z` — both ending in a normal, clean `claude -p` completion message (`"Nothing further to do here..."` / a full `## Summary` block), not an error. **`#4037` was never quarantined by these self-skips** — its earlier quarantine strikes (before this fix) were from unrelated token-pool exhaustion, already resolved and noted in its own later transcript. This is consistent with `poll_liveness` observing `exit_code == Some(0)` for a genuine self-skip (a `claude -p` session that completes its turn normally exits 0), which the existing reaper guard `duration_sec < insta_crash_secs && exit_code != Some(0)` (`sweep_registry.rs`, checkpoint-less branch) **already correctly exempts**.

**Conclusion: AC3 already appears satisfied** for the actual failure mode this issue reports, so no classification-logic change was made — exactly per the issue's own escape hatch ("If the observed code is `Some(0)`, AC3 is already satisfied — say so explicitly and close it out rather than adding redundant logic").

**Gap closed regardless:** every existing insta-crash fixture in `sweep_registry.rs` (`insert_dead_running`/`insert_dead_running_at`) only ever exercises the `exit_code == None` fallback path (a synthetic dead PID with no retained `Child` handle) — a real, handle-observed `Some(0)` exit had never actually been regression-tested. Added `reaper_real_clean_exit_does_not_count_as_insta_crash`, which spawns a real `true` child, retains the `Child` handle exactly as `dispatch()` does, and asserts the tally stays at 0 through `reap_once()`. This closes the coverage gap without touching behavior.

The theoretical `exit_code == None` hole Finding 1 also flagged (only reachable when the daemon restarts mid-flight and reconstructs an entry with no retained handle) is out of scope here — it's a distinct, narrower failure mode from what actually happened, and the issue explicitly permits deferring option 3 rather than growing this PR on an unconfirmed hypothesis.

## Test plan

- [x] Consumer-side check (AC2): `sweep_md_step_1a_self_claim_check_precedes_loom_building_skip_bullet` — mechanically asserts Step 1a precedes the `loom:building` skip bullet.
- [x] Negative case (AC4): unchanged — the skip bullet still fires unmodified when neither signal names the issue being pre-flighted (documented, not independently unit-testable — it's LLM-interpreted prose).
- [x] Regression: `cargo test -p loom-daemon dispatch_` — all pass, including the 3 argv-assertion tests Finding 3 flagged (`dispatch_happy_path_records_entry`, `dispatch_appends_dangerously_skip_permissions`, `dispatch_with_model_appends_model_arg`); only the exact-match one (`dispatch_appends_dangerously_skip_permissions`) needed an update, the other two use substring `.contains()` and were unaffected by the new flag's position.
- [x] Producer tests green: `sweep_registry.rs`, `work_finder.rs`, `ipc.rs` — all now assert `--claim-owned` alongside `LOOM_SWEEP_CLAIM_OWNED`.
- [x] Exit-code evidence for AC3: documented above — `Some(0)` for the actual failure mode, AC3 already satisfied, regression test added.
- [ ] Manual e2e: recommend confirming post-merge with a fresh `mcp__loom__dispatch_sweep` / `loom-daemon dispatch` on a real issue — the child's `.loom/logs/sweep-issue-<N>.log` should show it proceeding into the Builder phase rather than a "Skipping issue #N" line.
- [ ] Edge case — quarantine no longer accrues: recommend confirming post-merge across a few real dispatches.
- [x] Shell tests: `./defaults/scripts/tests/test-spawn-claude.sh` — 80/80 pass, including 2 new assertions for `--claim-owned` end-to-end argv fidelity.
- [x] Doc-lint: `cargo test --test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` — all pass, including 2 new CONTRACT tests.

## Post-merge remediation

Per the issue and its Champion comment, the operator should clear the stranded quarantines **after** this merges (clearing earlier would just re-strand them under the still-live bug):

```bash
for n in 4046 4050 4063 4064 4069 4072 4084; do loom-daemon quarantine clear "$n"; done
```

## Files changed

- `defaults/.claude/commands/loom/sweep.md` — Step 1a restructure (option 2), `--claim-owned` flag docs (option 1), Stage -1 consistency updates.
- `loom-daemon/src/sweep_registry.rs` — `--claim-owned` argv (option 1), updated/added dispatch tests, new `reaper_real_clean_exit_does_not_count_as_insta_crash` regression test (option 3 investigation).
- `loom-daemon/src/work_finder.rs`, `loom-daemon/src/ipc.rs` — producer-side `--claim-owned` assertions.
- `loom-daemon/tests/sweep_md_doc_lint.rs`, `loom-daemon/tests/sweep_md_stage_minus_one_doc_lint.rs` — new CONTRACT doc-lint tests.
- `defaults/scripts/tests/test-spawn-claude.sh` — new end-to-end argv pass-through test.
- `.loom/docs/daemon-reference.md` — updated dispatch example.

Closes #4111


rjwalters:
❌ **Changes Requested — Option 1's `--claim-owned` flag never reaches `/loom:sweep`; it crashes every daemon dispatch**

Thorough review of the fix against #4111's actual failure mode. Options 2 and 3 are solid; **Option 1 (the primary, first-preference mechanism) has a critical, reproducible defect** that must be fixed before merge.

## The defect

`SweepRegistry::spawn_child` (`loom-daemon/src/sweep_registry.rs:2086`) appends `--claim-owned <N>` as a **separate `cmd.arg()` after** `-p "<prompt>"`:

```rust
let prompt = format!("/loom:sweep {issue}");
cmd.arg("-p").arg(&prompt);
...
cmd.arg("--claim-owned").arg(issue.to_string());   // <-- separate argv token
...
cmd.arg("--dangerously-skip-permissions");
```

`spawn-claude.sh` does not special-case this flag — it forwards every non-`--use-wrapper`/`--help` token verbatim to the real `claude` binary (`exec claude "${PASSTHROUGH_ARGS[@]}"`, `defaults/scripts/spawn-claude.sh:381`). I reproduced the exact resulting invocation against the real Claude Code CLI (v2.1.212):

```
$ claude -p "/loom:sweep 4242" --claim-owned 4242 --dangerously-skip-permissions
error: unknown option '--claim-owned'
$ echo $?
1
```

**`--claim-owned` is not a recognized `claude` CLI flag** (`--model`/`--effort`/`--dangerously-skip-permissions` are, confirmed via `claude --help` — that's why those work as separate argv tokens). Only text *inside* the single `-p "<prompt>"` string ever becomes the `/loom:sweep` skill's own `$ARGUMENTS` (that's how `--dry-run`/`--no-daemon`/`--auto-stack` already work). Appending `--claim-owned` as a sibling CLI arg sends it to `claude`'s own parser instead, which rejects it outright — **the flag can never reach the skill's pre-flight check under this design, in any invocation.**

## Why this is worse than the bug being fixed

Because Option 1 is unconditionally emitted on **every** daemon-dispatched issue (per the PR's own description: "unconditionally on every issue dispatch"), this turns every single daemon dispatch into an immediate crash (`exit 1`, no session, no build activity) rather than a self-skip that at least starts a session. This is a regression from "the child self-skips its own claim" to "the child never starts at all" — a 100% dispatch failure rate, not a fix.

## Why the tests didn't catch it

- The Rust `dispatch_appends_claim_owned_flag` / `dispatch_appends_dangerously_skip_permissions` tests use a fake recording-script `spawn_bin` fixture, not the real `claude` binary.
- The new `test-spawn-claude.sh` assertions stub `claude` with a recording script (`stub-claude`), so they verify passthrough *fidelity* to the stub, not real-CLI *acceptance* of the flag.

Neither exercises the real CLI's argument parser, so the "end-to-end argv fidelity" claim in the PR description is true only against the stub, not against reality.

## Suggested fix

Embed the flag in the prompt text itself so it becomes part of `$ARGUMENTS`, mirroring how every other skill-consumed flag (`--dry-run`, `--no-daemon`, `--depends-on`, `--auto-stack`) is meant to arrive:

```rust
let prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
cmd.arg("-p").arg(&prompt);
```

and drop the separate `cmd.arg("--claim-owned")...` call. The doc-lint tests and `sweep.md` prose should be updated accordingly (the flag is parsed out of `$ARGUMENTS` during the skill's own classification step, not passed to `claude` itself).

## Separate finding (not blocking this PR, filed independently)

While reproducing this I found the **exact same pattern already broken on `main`** for the pre-existing `--depends-on <parent>` flag (#3729) — `spawn_child` appends it the identical way, and it hits the identical `error: unknown option '--depends-on'`. This is a distinct, pre-existing bug, not introduced by this PR, so I've filed it separately: #4121. Worth fixing alongside this PR's rework since the corrected pattern (embed in prompt text) applies to both.

## What's good in this PR

- **Option 2** (Step 1a restructure in `sweep.md`) is a clean, well-targeted fix — verified the doc-lint test (`sweep_md_step_1a_self_claim_check_precedes_loom_building_skip_bullet`) mechanically enforces the byte-offset ordering, and the restructured prose correctly makes the self-claim check mandatory and first-evaluated.
- **Option 3 investigation** is sound — traced the reaper code path (`sweep_registry.rs:2643`/`2669`) and confirmed the `exit_code != Some(0)` insta-crash exemption already exists in the checkpoint-less branch; the new `reaper_real_clean_exit_does_not_count_as_insta_crash` regression test correctly exercises the handle-retained `Some(0)` path (previously untested — every other insta-crash fixture only simulated the `None` no-handle path). Ran it directly, passes.
- All Rust `dispatch_*` tests, both doc-lint suites, and the 80 shell harness tests pass locally; `cargo fmt`/`clippy` clean; CI is green; merge state is CLEAN.
- The env var (`LOOM_SWEEP_CLAIM_OWNED`) backward-compatibility handling, `--claim-owned` fail-safe validation-rule semantics, and the Stage -1 consistency updates are all well-reasoned — none of that needs to change once the argv-placement bug above is fixed.

Once `--claim-owned` is embedded in the prompt text (and re-tested against the real `claude` CLI, not just a stub), this should be ready to re-review quickly — Options 2/3 don't need further changes.

rjwalters:
## Feedback addressed (#4120 review)

Fixed the argv-placement bug: `spawn_child` no longer appends `--claim-owned <N>` as a separate `cmd.arg()` sibling token (which the real `claude` CLI rejects with `error: unknown option '--claim-owned'`, crashing every daemon dispatch at exit 1 before any session starts). The marker is now embedded inside the `-p` prompt string:

```rust
let prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
cmd.arg("-p").arg(&prompt);
```

so it becomes part of `/loom:sweep`'s own `$ARGUMENTS` — the only channel that reaches the skill's pre-flight — exactly like `--dry-run` / `--no-daemon` / `--depends-on`. The separate `cmd.arg("--claim-owned")` call is dropped. The `LOOM_SWEEP_CLAIM_OWNED` env var is unchanged (kept for backward compat).

### Test coverage for the class of bug
The judge correctly noted neither the Rust nor shell tests exercised the real CLI's parser. Root cause of the miss: the Rust fixture recorded argv via `$*`, which flattens a single `-p "<prompt with spaces>"` arg into space-joined text indistinguishable from sibling tokens. I added a **per-token `arg: ` record** to the fixture (`for tok in "$@"; do printf 'arg: %s\\n' "$tok"; done`) and strengthened `dispatch_appends_claim_owned_flag` to assert:
- `arg: /loom:sweep 4246 --claim-owned 4246` — the flag is part of the **single `-p` prompt token**, and
- **NOT** `arg: --claim-owned` — it is never a standalone argv token.

This distinguishes prompt-embedded from sibling placement without needing the real `claude` binary.

### Verification
- `cargo test -p loom-daemon --lib dispatch` — 68 pass (incl. updated model/effort/depends-on argv assertions now expecting the embedded `--claim-owned <N>`).
- `reaper_real_clean_exit_does_not_count_as_insta_crash`, both doc-lint suites, `work_finder`/`ipc` producer claim-owned tests — all pass.
- `defaults/scripts/tests/test-spawn-claude.sh` — 80/80 pass; the `--claim-owned` assertion updated to the corrected invocation form (`-p "/loom:sweep 4111 --claim-owned 4111"`).
- `cargo fmt --check` + `cargo clippy --all-targets` clean.

Docs (`sweep.md`, `daemon-reference.md`) updated to the embedded form. The pre-existing identical `--depends-on` bug (#4121) is out of scope here as noted.

Re-labeled `loom:review-requested`.

rjwalters:
✅ **Approved!**

Re-reviewed the post-Doctor diff against the prior blocking finding and the issue's root-cause requirements.

## The blocking defect is fixed correctly

The prior review's critical finding — `--claim-owned <N>` appended as a **sibling** `cmd.arg()` (which the real `claude` CLI rejects with `error: unknown option '--claim-owned'`, crashing every daemon dispatch at exit 1) — is now fixed by embedding the flag **inside** the `-p` prompt string:

```rust
let prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
cmd.arg("-p").arg(&prompt);
```

Confirmed via `grep` that no code path still appends `--claim-owned` as a separate argv token anywhere in `loom-daemon/src`. This mirrors the existing `--dry-run`/`--no-daemon`/`--depends-on` pattern, the only channel that reaches `/loom:sweep`'s own `$ARGUMENTS`.

Also verified the new regression coverage genuinely closes the gap that let the sibling-arg bug slip through originally: the test fixture now records each argv **token** separately (`for tok in "$@"; do printf 'arg: %s\n' "$tok"; done`), so `dispatch_appends_claim_owned_flag` can assert the flag lands *inside* the single `-p` token and explicitly assert `!recorded.contains("arg: --claim-owned")` — the exact distinction the old `$*`-substring check couldn't make.

## Root cause addressed, mechanically where possible

- **Option 1 (`--claim-owned` flag)**: now correctly reaches the child via the only viable channel (prompt text), verified against a per-token fixture. Belt.
- **Option 2 (mandatory Step 1a)**: `sweep.md`'s pre-flight rule is restructured from a trailing exception clause into a separately-numbered, explicitly-`MANDATORY` step evaluated *before* the `loom:building` skip bullet. `sweep_md_step_1a_self_claim_check_precedes_loom_building_skip_bullet` mechanically pins the byte-offset ordering so a future edit can't silently regress it back into a buried exception clause. This is as close to "mechanical enforcement" as a prose skill file gets — the ordering itself is compiler-enforced even though the LLM's compliance with it isn't. Suspenders.
- **Option 3 (insta-crash classification)**: investigated per the issue's explicit gating instructions rather than blindly adding a second exemption. Correctly found the `exit_code != Some(0)` guard already exists and traced why it wasn't observed as broken; added `reaper_real_clean_exit_does_not_count_as_insta_crash`, which is the first test in the suite to exercise a real, handle-retained `Some(0)` exit rather than the synthetic no-handle `None` path every other insta-crash fixture uses. Good instinct not to add dead code on an unconfirmed hypothesis.

## Verification (ran locally, not just trusting the PR description)

- `cargo test -p loom-daemon --lib dispatch_` — 49/49 pass, including `dispatch_appends_claim_owned_flag` and the corrected `dispatch_appends_dangerously_skip_permissions`.
- `cargo test --test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` — 23/23 pass, including both new CONTRACT tests.
- `reaper_real_clean_exit_does_not_count_as_insta_crash` — passes in isolation.
- `defaults/scripts/tests/test-spawn-claude.sh` — 80/80 pass.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean.
- `gh pr checks 4120` — all CI green (Rust Unit Tests, Rust Linting & Build Check, Shellcheck, Installer Integration Tests, CLAUDE.md Line Budget).
- `mergeStateStatus` — `CLEAN`.

## Scope / other notes

- 8 files changed, +367/-32 (399 total) — a bit over the issue's original "~5 files, ≤250 LOC" guard but within the curator's later "approaching ~400 LOC" ceiling, and the overage is almost entirely the consumer-side test coverage AC2 explicitly asked for (the doc-lint tests, the per-token fixture rework, and the new reaper regression test). Justified.
- The pre-existing, structurally-identical `--depends-on` sibling-arg bug was correctly filed as a separate follow-up (#4121) rather than scope-creeping this PR — verified it's open and correctly scoped.
- PR description closes #4111 with the correct GitHub auto-close syntax.

Nice work tracing the real-CLI failure mode and fixing the class of bug (not just the one flag) with the per-token fixture rework.

Issue #3719, opened 2026-07-22
Builder cwd-reset main-repo write trap recurs — worktree guard doesn't fire on builder subagents (re: #3513)


rjwalters:
## Context

Across a large multi-wave `/loom:sweep`-style session (2026-07-21) on rjwalters/kicad-tools, **at least 5 builder subagents** independently reported the same near-miss: they wrote a new source file to the **main-repo path** instead of their issue worktree, caught it via a failing import or a line-count mismatch, then moved the change into the worktree and reverted main. Verbatim: "I initially wrote the new module accidentally to the main repo root (cwd-reset trap); I caught it immediately, moved it into the worktree, and confirmed main stayed clean" (issues #4470, #4466, #4485, and others).

Closely related to closed **#3513** (which added *guidance* only). The high recurrence rate shows guidance alone doesn't prevent the failure mode on the builder-subagent dispatch path — the cwd resets between tool calls and repo-relative writes silently land in main.

---

### Root-cause mechanism (verified in-repo)

`guard-worktree-paths.sh` (canonical copy at `.loom/hooks/guard-worktree-paths.sh`) is a `PreToolUse`/`Edit|Write` hook that reads **`LOOM_WORKTREE_PATH` from its own process env** (line 32) and denies any Edit/Write whose resolved path escapes that worktree. It is a **no-op when `LOOM_WORKTREE_PATH` is unset**.

`LOOM_WORKTREE_PATH` is set **only at OS-process-spawn time**, in exactly two places, both of which spawn a fresh `claude` process for a single worktree agent:

- `loom-tools/src/loom_tools/agent_spawn.py:812` — prepends `LOOM_WORKTREE_PATH='<working_dir>' ` to the spawned `claude` command line.
- `loom-daemon/src/terminal.rs:1361-1369` — sets `LOOM_WORKTREE_PATH` on the tmux session for daemon-spawned worktree terminals.

[...]

Issue #3513, opened 2026-06-18
Builder writes to main worktree instead of issue worktree — recurrence of #2802 (cwd resets between tool calls)


rjwalters:
## Recurrence of #2802

#2802 ("Builder scope creep: uncommitted work leaks to main worktree", closed 2026-02-18) flagged that a builder can write to the **main repo working tree** instead of its `.loom/worktrees/issue-N` worktree, and recommended either a post-phase uncommitted-changes guard or investigating the underlying `cwd` cause. The cwd root cause was not fully closed out — it **recurred 2026-06-17** in a downstream consumer (`spheresemi/sphere`), four months later.

## Fresh evidence (2026-06-17, sphere issue #9265 build)

The builder reported, verbatim:

> "Early on I discovered I had been editing the **main repo** path rather than the worktree (**the harness reset cwd between calls**). I caught it before any commit, copied the modified/new files into `.loom/worktrees/issue-9265`, and ran `git checkout -- ...` + removed the stray file to fully restore main to a clean state."

So the builder self-recovered this time (no lost work, main left clean — verified by the orchestrator). But the **root cause is unchanged**: the agent's cwd does not reliably stay inside its worktree across tool calls, so a builder that uses repo-relative paths (or an early `cd` that doesn't persist) writes to the wrong tree. Self-recovery is luck-dependent — the #2802 incident lost 188 lines across 7 files because it was *not* caught.

## Why the existing mitigation is insufficient
A post-phase "is main dirty?" detection guard (the lighter of #2802's two suggestions) only *detects* the leak after the fact; it doesn't prevent the builder from doing real work in the wrong place and then having to reconstruct/clean it. The reliable fix is at the source.

## Proposed root-cause fix (in priority order)
1. **Builder resolves an absolute worktree root once and writes only under it.** The worktree workflow already knows `WORKTREE=.loom/worktrees/issue-N`; the builder should capture its absolute path and use absolute paths for all Write/Edit/Bash file ops, never relying on the persisted cwd.
2. **Re-assert cwd at the top of each file-mutating step** (or wrap mutations so they `cd "$WORKTREE"` in the same shell invocation), since cwd does not persist across tool calls.
3. **Keep the #2802 post-phase guard as a backstop**: after the builder phase, fail (not warn) with a distinct exit code if the main worktree has uncommitted changes the builder introduced.

## Acceptance
- A builder run for a non-trivial issue produces zero uncommitted changes in the main worktree.
[...]

Issue #2802, opened 2026-02-18
Builder scope creep: uncommitted work leaks to main worktree


rjwalters:
## Bug Description

During shepherd run for #2768, the builder worked in the issue-2768 worktree and committed the MCP failure classification fix (the actual issue scope). However, it also wrote **additional code** (rate limit abort detection — exit code 13, wrapper sentinel, base.py detector) to the **main repo's working tree** as uncommitted changes.

This work was never committed to the feature branch or included in the PR. It was only discovered manually when `git pull` failed due to local modifications on main.

## Reproduction

1. Run `/shepherd <issue>` for a non-trivial issue
2. Builder works in `.loom/worktrees/issue-N` (correct)
3. Builder also modifies files in the main repo directory (incorrect)
4. PR is created from worktree — main's changes are orphaned
5. `git pull` on main fails due to uncommitted modifications

## Evidence from #2768

- Builder log header: `Working Directory: /Users/rwalters/GitHub/loom/.loom/worktrees/issue-2768`
- Worktree commit: `84fcb53 feat: fallback handler classifies MCP failures as builder_unknown_failure` (correct scope)
- Main worktree: 188 lines of uncommitted rate limit abort code across 7 files (scope creep)

## Impact

- **Lost work**: Builder-generated code is silently abandoned as local changes
- **Merge conflicts**: Uncommitted files on main block `git pull`
- **No detection**: Shepherd exits 0 (success) despite the leak

## Suggested Fix

After the builder phase completes, the shepherd should check for uncommitted changes in the main worktree (or the shepherd's own working directory). If found:
- Log a warning with the list of modified files
- Optionally fail with a specific exit code (similar to WORKTREE_ESCAPE)

Alternatively, investigate why the builder's Claude Code session writes to main — it may be a `cwd` issue in how tools resolve paths, or the builder agent navigating out of its worktree.

Discovered during post-mortem of shepherd run #2768.

PR #4129, opened 2026-07-28
fix(hooks): make guard-worktree-paths.sh derive worktree scope from the target path

### LINKED ISSUE #4007: Worktree isolation is advisory-only: builders can silently write into the MAIN checkout (guard-worktree-paths.sh is inert on the sweep path)
> 📝 **Curator (2026-07-27): rescoped.** As filed this was an umbrella covering four
> compounding failures. Three of them have since been re-diagnosed, fixed, or given
> owners (see *Rescope ledger* below). **This issue is now scoped to the one defect
> that is still live and still unowned: worktree isolation is advisory-only, so a
> builder can silently write into the main checkout.** The original report is
> preserved verbatim at the bottom.

## Problem Statement

Loom prevents builder agents from writing into the primary checkout **entirely by
prompt instruction**. There is no mechanical enforcement on the daemon-dispatched
sweep path. When the LLM-dependent guard fails — as it did in the original report —
edits land in the main working tree instead of the issue worktree.

The blast radius is not "one messy diff":

- The daemon's own repo is corrupted with uncommitted foreign work.
- A dirty main tree **disables the main-health-gate** (`"will not hard-reset over
  operator edits"`), so the daemon loses its self-repair path.
- The work is invisible to the PR flow — it belongs to an issue whose worktree has a
  *different, less complete* version of the same change.

This has now recurred at least three times: #2802 → #3513 → this issue.

### Root cause: the one mechanical guard that exists is inert

`.loom/hooks/guard-worktree-paths.sh` is exactly the right idea and is registered as a
`PreToolUse` hook (`.claude/settings.json:22`). It never fires on the sweep path, for
three independent reasons:

1. **It no-ops unless `LOOM_WORKTREE_PATH` is set** (`.loom/hooks/guard-worktree-paths.sh:32-35`):
   ```bash
   WORKTREE_PATH="${LOOM_WORKTREE_PATH:-}"
   if [[ -z "$WORKTREE_PATH" ]]; then
       exit 0
   fi
   ```
   Nothing on the daemon path sets that variable. `spawn_child`'s env allowlist
   (`loom-daemon/src/sweep_registry.rs:2501-2530`) sets `LOOM_TERMINAL_ID`,
   `LOOM_SWEEP_CLAIM_OWNED`, `LOOM_WORKSPACE`, `LOOM_PACKAGE_PATH` — **not**
   `LOOM_WORKTREE_PATH`.

2. **Its only setter has no caller.** `Terminal::set_worktree_path`
   (`loom-daemon/src/terminal.rs:1353-1376`) writes it via `tmux set-environment`, reachable
   only through `Request::SetWorktreePath` (`loom-daemon/src/ipc.rs:1080-1088`). A repo-wide
   search for that request finds the type definition, the handler, and no issuer. It is
   dead code, and it is tmux-only regardless.

3. **It matches only `Edit|Write`, never `Bash`.** So `cd`, `git`, `cargo`, and shell
   redirects from a builder are unconstrained with respect to worktree residency.

Compounding: the daemon deliberately pins the child's cwd to the **workspace root**, not
the worktree (`loom-daemon/src/sweep_registry.rs:2536`, `.current_dir(&self.config.workspace_root)`),
for the good reason recorded at `:2531-2535` (#3730 — deterministic relative config reads).
So the child *starts* in the main checkout by design, and a cwd reset returns it there.

### The non-obvious constraint that makes this more than a one-liner

The obvious fix — "export `LOOM_WORKTREE_PATH` into the child" — **cannot work by itself.**
`defaults/.claude/commands/loom/sweep.md:1403` already documents why:

> a builder subagent is dispatched via the Task tool … and inherits the orchestrator's
> single shared process env, which has **no** `LOOM_WORKTREE_PATH` — and the Task tool
> exposes no per-subagent env-injection parameter — so `guard-worktree-paths.sh`
> structurally cannot fire per builder on this path (#3719).

One sweep process hosts *multiple* builder subagents for *different* issues in one shared
env. A single process-wide env var is the wrong shape. The guard must derive the expected
worktree **from the target path itself**, not from ambient env.

### Suggested approach (not binding)

`worktree.sh` already writes a `.loom-managed` sentinel at every managed worktree root
(`defaults/scripts/worktree.sh:89-99`) and never at the main repo root. That makes a
**path-derived** check viable with no env plumbing at all:

- Walk up from the resolved target path; if the first `.loom-managed` ancestor differs
  from the first `.loom-managed` ancestor of *any* worktree, or if the path resolves under
  the main repo root while a managed worktree exists for the in-flight issue, deny.
- Keep the existing `LOOM_WORKTREE_PATH` fast path for the tmux/manual case.
- Extend the matcher to `Bash` for the obviously-writing verbs, or accept `Edit|Write`
  coverage as phase 1 and file `Bash` coverage as a follow-up.

The sentinel currently records only issue + branch (`defaults/scripts/worktree.sh:92-98`);
recording the absolute worktree root would make the check cheaper and unambiguous.

**Scope wrinkle the Builder must handle:** `guard-worktree-paths.sh` is committed at
`.loom/hooks/` but has **no `defaults/hooks/` counterpart** — confirmed by
`git ls-tree -r origin/main --name-only -- defaults/hooks`, and acknowledged at
`defaults/scripts/check-main-freshness.sh:166-167` ("repo-specific hooks like
`guard-worktree-paths.sh` have no defaults/ counterpart and are not drift"). It *is* listed
in `.loom/install-metadata.json:6`, so consumer repos have a copy with no upstream source of
truth. Adding the `defaults/hooks/` source is in scope; deciding the broader
install-metadata story is **not** (that belongs to epic #3835).

## Acceptance Criteria

- [ ] A builder writing to a path outside its issue worktree is **denied by a hook**, with
      the daemon-dispatched sweep path (Task-subagent builders, shared process env) covered
      — not only the tmux/manual path.
- [ ] The mechanism does **not** rely on a process-wide `LOOM_WORKTREE_PATH` for the
      subagent path, per the #3719 constraint above.
- [ ] A regression test dispatches (or simulates) a builder attempting a write to the main
      repo root and asserts the hook returns a `deny` decision. This is the criterion that
      distinguishes a real fix from a documentation change.
- [ ] `guard-worktree-paths.sh` gains a `defaults/hooks/` source copy so consumer installs
      have an upstream source of truth.
- [ ] Guard follows the established contract: never exits non-zero, ERR trap allows, emits
      `hookSpecificOutput.permissionDecision` (pattern at `.loom/hooks/guard-loom-workflow.sh:191-228`).
- [ ] A category toggle is added following the canonical 20-line pattern at
      `.loom/hooks/guard-destructive-generic.sh:417-437` (config key + `LOOM_*` env override,
      env > config > default), and documented in `defaults/CLAUDE.md`'s guard inventory.
- [ ] Fail-open on every unexpected condition (missing `jq`, unparseable input, no sentinel
      found) — a broken guard must not wedge all agent writes.

**Explicitly out of scope** (do not expand into these):
- Changing `spawn_child`'s `current_dir` away from the workspace root — that would regress #3730.
- Removing the prose warnings in `defaults/roles/builder.md`; they stay as defense in depth.
- The dead `SetWorktreePath` IPC path — note it, but its removal belongs in a separate cleanup.

## Affected Files

> Note: `.loom/scripts` and `.loom/roles` are **symlinks** into `defaults/` (verified via
> `git ls-tree origin/main .loom/scripts`). Edit the canonical `defaults/` paths; the
> `.loom/` paths will not resolve in a fresh worktree. `.loom/hooks/` **is** a real
> committed directory.

- `.loom/hooks/guard-worktree-paths.sh` — the core change: path-derived worktree
  derivation replacing the `LOOM_WORKTREE_PATH` gate at `:32-35`; toggle plumbing.
- `defaults/hooks/guard-worktree-paths.sh` — **(new)** upstream source copy.
- `.claude/settings.json` — hook registration; extend matcher if `Bash` coverage lands
  (existing idiom at `:9`, `:13`, `:22`).
- `defaults/scripts/worktree.sh` — `write_loom_sentinel()` at `:89-99`; consider recording
  the absolute worktree root in the sentinel body (`:92-98`).
- `defaults/CLAUDE.md` — guard inventory (`:1153-1158`) + toggle resolution-order table (`:1162-1178`).
- `CLAUDE.md` — one-line guard-category addition only (budget check enforced by `scripts/check-claude-md-budget.sh`).
- `defaults/hooks/tests/` — **(new)** regression test, alongside `test-methodology-inject.sh`.

Reference only, no edits expected:
- `loom-daemon/src/sweep_registry.rs:2501-2536` — child env allowlist + `current_dir` pin.
- `defaults/.claude/commands/loom/sweep.md:1403-1411` — the #3719 constraint.
- `defaults/scripts/check-main-clean.sh:232-251` — existing post-hoc detection (exit 3).

## Test Plan

- [ ] **Automated (primary):** new test in `defaults/hooks/tests/` feeding the hook
      synthetic `PreToolUse` JSON — (a) target inside a `.loom-managed` worktree → allow;
      (b) target at the main repo root while a managed worktree exists → **deny**;
      (c) relative path + `cwd` at main root → deny; (d) no sentinel anywhere → allow (fail-open).
- [ ] **Automated (contract):** assert exit status is 0 in every branch and that deny emits
      well-formed `hookSpecificOutput` JSON.
- [ ] **Automated (toggle):** assert the guard no-ops when disabled via config, and that the
      `LOOM_*` env override beats config.
- [ ] **Manual end-to-end:** dispatch a real sweep, and from the builder attempt a
      repo-relative `Write` after a cwd reset; confirm the deny fires and the sweep still
      completes via the correct worktree path.
- [ ] **Regression:** `defaults/scripts/check-main-clean.sh` still exits 0 on a clean main
      after a full sweep; `bash defaults/hooks/tests/*.sh` green.
- [ ] **Edge cases:** symlinked worktree roots (`pwd -P` normalization); a path under
      `.loom/worktrees/` belonging to a *different* issue; `jq` absent → allow.

## Rescope ledger — what moved, and where

| Original claim | Status |
|---|---|
| #3887 startup race under-mitigated | **Closed.** Watchdog default raised 120s→300s with per-`SweepId` progress latching (#4088, `loom-daemon/src/sweep_registry.rs:452`, `:3243-3247`). |
| "Work-finder should skip `loom:building`" (comment 2) | **Already implemented** — `loom-daemon/src/work_finder.rs:158`, `SKIP_LABELS`. The observed `labeled-skip=0` was *not* a missing filter: the watchdog **reverts** `loom:building → loom:issue` (`sweep_registry.rs:2854-2856`) before the next tick, so there is no label left to skip. Original diagnosis corrected. |
| Account exhaustion drives the re-dispatch loop (comment 1) | **Split out** → see the new sibling issue on daemon sweeps bypassing `claude-wrapper.sh`. |
| Duplicate dispatch on #4034 despite an approved open PR (comment 2) | **Split out** → new sibling issue. |
| "Stagger by completion, not fixed delay" | **Owned by #4003** (resource leases). Confirmed still a fixed 2s sleep: `sweep_registry.rs:426`, `:2404-2414`; the slot is occupied from `dispatch()` return, so a wedged child holds it for the full 300s watchdog window. Tracking there rather than duplicating. |
| Broken token-health ranking probe (comment 1) | **Fixed** by #4114 / `bb6f8e59` — `probe-tokens.sh:25` now invokes the real `loom_tools.tokens.cli`. |
| Quarantine strands issues at `loom:blocked` | **Fixed** by #4110 / PR #4119. |

Related but distinct: #4111 (child self-skips its own claim), #4027 (missing `/loom:*`
commands → fast-fail loop), #4028 (cross-host soft claim), #4116 (cross-host collision
observability), epic #3835 (install/source-of-truth story).

---

<details>
<summary><b>Original issue (preserved verbatim)</b></summary>

Observed running a second daemon (laptop, cap=4, 4 accounts) against a 6-issue backlog while the host was also busy (an interactive Claude session + concurrent cargo builds). Two known-but-unclosed failure modes compounded into main-checkout corruption.

## Symptom
- The work-finder re-dispatched the **same issues every tick**, with `labeled-skip=0, in-flight-skip=0` on every tick despite the issues having been dispatched seconds earlier:
  ```
  22:01 dispatching #2,#3,#4,#5 | tick: 5 seen, 4 dispatched, 0 labeled-skip, 0 in-flight-skip, 1 deferred
  22:02 dispatching #2,#3,#4,#5 | tick: 5 seen, 4 dispatched, 0 labeled-skip, 0 in-flight-skip, 1 deferred
  22:04/05/06 dispatching #6 (three times) ; 22:11 dispatching #3 again
  ```
- Issue #5 ended up with its work in TWO places: a partial impl in its `feature/issue-5` worktree, AND a *more complete* impl sitting **uncommitted in the main workspace checkout**. The main tree being dirty also disabled the main-health-gate ("will not hard-reset over operator edits").

## Root cause (two compounding bugs)
1. **#3887 startup race, under-mitigated under load.** The 2s dispatch stagger (`DEFAULT_DISPATCH_STAGGER_MS=2000`) is too small when the host is loaded: back-to-back sweeps still wedge in MCP-init/local-startup. The startup watchdog then reaps each wedged sweep and reverts its claim `loom:building → loom:issue`, so both dedup guards (label + in-flight registry) are cleared before the next tick — which is why `labeled-skip`/`in-flight-skip` are perpetually 0 and the same issues re-dispatch every 60s. Net: a re-dispatch loop that also produces **duplicate concurrent sweeps for one issue** (#5 dispatched 22:01 AND 22:02).
2. **cwd-reset worktree-escape (confirmed in `builder.md` lines ~304–329).** The builder command already warns that after a Claude Code cwd reset a *relative*-path operation "lands in the main worktree instead of your issue worktree." A duplicate #5 sweep hit exactly that: its edits landed in the **main checkout** instead of `.loom/worktrees/issue-5`. The warning is an LLM-dependent guard and it failed.

## Why it matters
A mis-isolated sweep writing to the primary checkout corrupts the daemon's own repo, disables the health gate, and (with duplicate sweeps) burns tokens in a loop. On a host that also runs other services this is high-blast-radius.

## Suggested fixes
- **Stagger by completion, not a fixed delay:** hold the concurrency slot until the child demonstrably passes startup (worktree created / first HTTPS / checkpoint) before freeing it, instead of a fixed 2s; and/or scale the default stagger with concurrency and raise it.
- **Make worktree isolation structural, not advisory:** launch the builder with the issue worktree as the child's actual working directory and/or hard-fail if `git rev-parse --show-toplevel` resolves to the main root. A prompt warning can't guarantee it.
- **Never run two live sweeps for one issue:** if a re-dispatch target already has a worktree / an unreaped claim, quarantine or reconcile rather than spawning a second child.

## Workaround applied on the affected host
Paused the daemon, killed orphaned sweeps, salvaged the main-tree work to a branch, removed the worktrees/branches, reset labels to `loom:issue`, and reconfigured to `LOOM_WORK_FINDER_MAX_CONCURRENT=1` + `LOOM_SWEEP_DISPATCH_STAGGER_MS=20000` (cap=1 removes the simultaneity the race needs).

</details>

rjwalters:
## Summary

`guard-worktree-paths.sh` (a `PreToolUse` hook meant to deny Edit/Write calls
that land outside a builder's issue worktree) previously no-op'd unless
`LOOM_WORKTREE_PATH` was set. Nothing on the daemon-dispatched sweep path
ever sets that var — a sweep hosts multiple Task-subagent builders (different
issues) sharing one process env, and the Task tool exposes no per-subagent
env-injection parameter (#3719) — so the guard was structurally inert there,
and a builder that used a repo-relative path after a cwd reset could
silently write into the **main** checkout instead of its worktree
(recurrence chain: #2802 → #3513 → #4007).

This PR adds a path-derived fallback that works without any ambient env var,
keeps the existing `LOOM_WORKTREE_PATH` fast path for tmux/manual sessions
unchanged, adds the missing `defaults/hooks/` source copy of the guard, and
adds a regression test suite.

## Changes

- `.loom/hooks/guard-worktree-paths.sh` / `defaults/hooks/guard-worktree-paths.sh`
  (new upstream source copy, byte-identical to the installed copy):
  - **Fast path (unchanged):** when `LOOM_WORKTREE_PATH` is set, only that
    exact worktree is allowed (tmux/manual sessions).
  - **Path-derived fallback (new):** when `LOOM_WORKTREE_PATH` is absent,
    walk up from the resolved target path looking for the `.loom-managed`
    sentinel that `worktree.sh` writes at the root of every worktree it
    creates (never at the main repo root).
    - Target inside any managed worktree → allow. (We cannot tell *which*
      issue a given subagent owns — no ambient signal exists for that — so
      this deliberately does not attempt cross-issue isolation; it only
      guards against landing in the main checkout, the actual failure mode.)
    - Target resolves into the main checkout while a managed worktree
      exists → **deny**.
    - No managed worktree has ever been created → allow (fail-open; nothing
      to protect).
    - Target outside both the main checkout and any worktree (e.g. `/tmp`
      scratch files) → allow.
  - New toggle `guards.worktreeIsolation` / `LOOM_GUARD_WORKTREE_ISOLATION`
    (default on), following the env > config > default resolution order used
    by every other guard category in this repo (pattern at
    `guard-destructive-generic.sh` around `guards.sqlDdl`).
  - Fail-open contract preserved: never exits non-zero; ERR trap allows;
    missing `jq` / unparseable input / resolver errors all allow.
- `defaults/hooks/tests/test-guard-worktree-paths.sh` (new): regression suite
  covering all four core scenarios from the acceptance criteria, the
  `LOOM_WORKTREE_PATH` fast path, the toggle (env beats config beats
  default), the fail-open contract (exit always 0, valid deny JSON), `jq`
  absent, and `defaults/`↔`.loom/` byte-identity.
- `defaults/CLAUDE.md`: documents `guard-worktree-paths.sh` in the guard
  inventory and adds a "Worktree Isolation Guard Opt-Out" toggle subsection.
- `CLAUDE.md`: one-line addition of `worktreeIsolation` to the guard-toggle
  catalog (still within the 320-line budget).
- `defaults/scripts/check-main-freshness.sh`: updates a comment that used
  `guard-worktree-paths.sh` as an example of a hook with no `defaults/`
  counterpart — it now has one, so the example is swapped to
  `post-worktree.sh` (still accurate) and a note added that this hook is now
  drift-checked like any other.

**Explicitly out of scope** (per the curated issue): `spawn_child`'s
`current_dir` pin to the workspace root (#3730), the `builder.md` prose
warnings (kept as defense in depth), and removing the dead `SetWorktreePath`
IPC path. Also left the matcher at `Edit|Write` only (Bash coverage is
called out as acceptable follow-up work in the issue) and left
`worktree.sh`'s sentinel body format unchanged — the guard doesn't need to
parse it, only detect its presence, so recording the worktree root there is
not required for this fix.

- [x] A builder writing outside its issue worktree is denied by a hook on
  the daemon-dispatched sweep path (no `LOOM_WORKTREE_PATH` needed) —
  verified via the path-derived fallback test cases.
- [x] The mechanism does not rely on a process-wide `LOOM_WORKTREE_PATH` for
  the subagent path — the fallback only fires when that var is absent.
- [x] Regression test dispatches synthetic `PreToolUse` JSON simulating a
  write to the main repo root and asserts `deny` — `test-guard-worktree-paths.sh`.
- [x] `guard-worktree-paths.sh` has a `defaults/hooks/` source copy.
- [x] Guard never exits non-zero; ERR trap allows; emits well-formed
  `hookSpecificOutput.permissionDecision` on deny.
- [x] Category toggle added following the canonical pattern (config key +
  `LOOM_*` env override, env > config > default), documented in
  `defaults/CLAUDE.md`'s guard inventory.
- [x] Fails open on every unexpected condition (missing `jq`, unparseable
  input, no sentinel found).

## Test Plan

- `bash defaults/hooks/tests/test-guard-worktree-paths.sh` — 19/19 passed
  (all four core scenarios + fast path + toggle + contract + `jq`-absent +
  `defaults/`↔`.loom/` sync).
- `bash defaults/hooks/tests/test-methodology-inject.sh` — 27/27 passed (no
  regression from the shared error-log/trap patterns).
- `bash defaults/hooks/tests/test-skill-router.sh` — 28/28 passed.
- `bash scripts/test-installer.sh` — 138/138 passed (confirms the new
  `defaults/hooks/` file installs/resyncs correctly via the existing
  glob-based hook installer, no code change needed there).
- `bash defaults/scripts/tests/test-check-main-clean.sh` — 20/20 passed.
- `bash defaults/scripts/tests/test-check-main-freshness.sh` — 15/15 passed.
- `bash defaults/scripts/tests/test-worktree-sentinel.sh` — 10/10 passed.
- `./.loom/scripts/check-main-clean.sh` — main worktree confirmed clean.
- `./scripts/check-claude-md-budget.sh` — CLAUDE.md 314/320 lines, OK.
- `shellcheck` on both hook copies and the new test file — clean.
- Manual smoke tests (ad hoc, matching the issue's edge-case list): symlinked
  worktree roots via `pwd -P`, a path under a *different* issue's worktree
  (allowed by design, documented in the test), toggle env-overrides-config in
  both directions.

Closes #4007


rjwalters:
## Judge Review — APPROVE

Reviewed the diff, checked out the PR branch (an existing worktree at `.loom/worktrees/issue-4007` already had it), and ran the new regression suite plus adjacent checks.

### Acceptance criteria (issue #4007) — all met

- **Path-derived worktree check replaces the `LOOM_WORKTREE_PATH`-only gate**: the `LOOM_WORKTREE_PATH` fast path is preserved unchanged for tmux/manual sessions; the new `path_derived_allow()` fallback walks up from the resolved target looking for the `.loom-managed` sentinel `worktree.sh` writes at every worktree root, and denies only when the target resolves into the main checkout while at least one managed worktree exists elsewhere. This directly covers the daemon-dispatched sweep path (shared process env, no per-subagent `LOOM_WORKTREE_PATH`, per #3719) that was the actual gap in #4007.
- **`defaults/hooks/` source copy added**: `defaults/hooks/guard-worktree-paths.sh` is new and byte-identical to `.loom/hooks/guard-worktree-paths.sh` (verified with `diff` locally — identical). `defaults/scripts/check-main-freshness.sh`'s drift-exemption comment is correctly updated to swap the example to `post-worktree.sh` and note this hook is now drift-checked.
- **Regression test added and passing**: `defaults/hooks/tests/test-guard-worktree-paths.sh` — ran it locally, **19/19 passed**, covering all four core scenarios from the issue (inside worktree → allow; main root with worktree present → deny; relative path + cwd at main root → deny; no sentinel anywhere → fail-open allow), plus the fast path, the toggle (env > config > default, both directions), the fail-open contract (exit always 0, valid JSON on deny), `jq` absent, and `defaults/`↔`.loom/` byte-identity.
- **Toggle follows the canonical env > config > default pattern**: `guards.worktreeIsolation` / `LOOM_GUARD_WORKTREE_ISOLATION`, default on, mirrors the `guards.sqlDdl` pattern; documented in `defaults/CLAUDE.md`'s guard inventory and a new "Worktree Isolation Guard Opt-Out" subsection, plus a one-line addition to `CLAUDE.md`'s catalog (budget still 314/320 per `check-claude-md-budget.sh`).
- **Fail-open behavior preserved**: ERR trap still `exit 0`; missing `jq`, no sentinel, and config-parse failures all fall through to allow. Verified via the "jq absent" and "no sentinel anywhere" test cases.
- **No scope creep**: confirmed no changes to `loom-daemon/src/sweep_registry.rs` (the `current_dir` pin, #3730, untouched), `builder.md` prose (untouched), the dead `SetWorktreePath` IPC path (untouched), or `.claude/settings.json` (matcher stays `Edit|Write` only, not extended to `Bash`).

### Additional verification

- `shellcheck` clean on both hook copies and the new test file.
- `bash scripts/test-installer.sh` passes via CI's "Installer Integration Tests" check (confirms the new `defaults/hooks/` file installs/resyncs via the existing glob-based installer with no code change needed there).
- `./.loom/scripts/check-main-clean.sh` — clean.
- The main-checkout prefix match in `path_derived_allow()` correctly uses trailing-slash-normalized comparison (`target_slash`/`main_slash`), avoiding the classic `/Users/foo` vs `/Users/foobar` prefix bug.
- CI: all applicable checks green (Shellcheck, CLAUDE.md Line Budget, Installer Integration Tests, Detect Changes); Rust/MCP jobs correctly skip (no Rust/TS files touched).

Nicely scoped, well-tested, and the doc updates (both `CLAUDE.md` files) accurately describe the new mechanism. No changes requested.

Issue #4130, opened 2026-07-28
launchctl bootstrap gui/$UID fails over SSH — daemon cannot be (re)started remotely on macOS


rjwalters:
## Context

Split out of #4005 (*Daemon gh auth depends on interactive login-keychain unlock — blocks headless / remote operation*), which scoped this out explicitly:

> `launchctl bootstrap gui/$UID` fails over SSH (`error 125: Domain does not support specified action`) — *not* an auth problem. `gui/$UID` is a per-GUI-session domain that does not exist in an SSH session; the fix is a different mechanism (a `LaunchDaemon` in the `system` domain, or `launchctl asuser`), with its own security and privilege trade-offs. **Out of scope — file a follow-up issue.**

#4005 fixes credential resolution (a daemon that's already running, or a fresh Linux/`--no-launchd` start, now authenticates headlessly with an exported `GH_TOKEN`). This issue is the remaining half: **starting** (or restarting) the daemon as a macOS `gui/<uid>` LaunchAgent from an SSH session still fails, independent of credentials, because `gui/<uid>` is a per-GUI-login-session launchd domain that simply does not exist when there is no active GUI session (e.g. a headless Mac accessed only over SSH, or a GUI session that has fully logged out).

## Problem

`defaults/scripts/cli/loom-daemon-start.sh` renders a launchd plist and loads it with:

```bash
launchctl bootstrap "gui/${LAUNCHD_UID}" "$PLIST_FILE"
```

Over SSH with no active GUI login session, this fails with `error 125: Domain does not support specified action` (or similar), because `gui/$UID` requires a live GUI Aqua session bound to that UID. So:

- An operator who wants to (re)start the daemon on a headless Mac, or on a Mac accessed only via SSH, cannot use the documented start path at all.
- `--no-launchd` (nohup) is the documented escape hatch today, but it reintroduces the #3972 Mach-bootstrap-namespace failure mode this repo already hit once in production (the daemon and its children start failing `trustd`/`opendirectoryd` XPC lookups when the launching SSH session dies) — so it's a workaround with a known regression risk, not a real fix.
- `loom-daemon-update.sh`'s internal restart of an **already-running** launchd-supervised daemon works fine over SSH (it doesn't re-bootstrap the domain) — this gap is specifically about **first start** (or restart-after-crash) of the launchd job itself from a non-GUI session.

## Suggested acceptance criteria

[...]

PR #4131, opened 2026-07-28
feat: add startup forge-credential preflight + plist hardening to daemon

### LINKED ISSUE #4005: Daemon gh auth depends on interactive login-keychain unlock — blocks headless / remote operation
## Context
The daemon's own forge calls (`gh issue list`, claim reconciliation) authenticate via `gh`, whose token on macOS lives in the **login keychain**. That keychain only unlocks for processes in the user's GUI login session. Consequences hit this session hard:

- Restarting the daemon over SSH with a clean environment produced `401 Unauthorized` on **every** workspace's `gh` calls — dispatch silently stopped fleet-wide until the process was killed.
- `launchctl bootstrap gui/$UID ...` fails over SSH (`error 125: Domain does not support specified action`), so the daemon can only be (re)started from a console/GUI session — remote administration is impossible.
- `~/.config/gh/hosts.yml` carries no inline token (it delegates to the keychain), so there is no file-based credential the daemon can fall back to.

- [ ] The daemon can authenticate forge calls from a token supplied via config/env (e.g. `GH_TOKEN` in the launchd plist, or a loom-managed PAT file) instead of requiring an unlocked GUI keychain.
- [ ] Documented headless-operation path (server with no interactive login / SSH-only administration).
- [ ] Clear startup error if no usable credential is resolvable, rather than silent per-tick 401s.

---

### Root cause — narrower than the issue implies

Verified against the working tree today. **Two of the three suggested acceptance criteria
are already met by the plumbing; the failure is that nothing tells you when it isn't.**

- `defaults/scripts/cli/loom-daemon-start.sh:200` (the canonical source; `.loom/scripts/cli/loom-daemon-start.sh` is its gitignored installed copy) already forwards every exported `GH_TOKEN` / `GITEA_TOKEN` / `FORGE_TOKEN` verbatim into the launchd plist's `EnvironmentVariables`:
  ```
  done < <(env | grep -E '^(LOOM_[A-Za-z0-9_]*|GH_TOKEN|GITEA_TOKEN|FORGE_TOKEN)=' || true)
  ```
- The daemon's `gh` invocations are plain `Command::new("gh")` with **no `env_clear`** (`sweep_registry.rs`, `main_health_gate.rs:903`, `metrics_collector.rs:135/282/372`, `main.rs:638`, …), and the detached sweep child spawn adds to inherited env rather than replacing it (`sweep_registry.rs:2355`, `EXPERIMENT_ENV_ALLOWLIST` is *additive*). So a token in the daemon's environment does reach `gh` — for the daemon **and** for every sweep child.

So a token-based headless path already works **if the operator happens to have `GH_TOKEN`
exported in the shell that runs `loom-daemon-start.sh`**. The observed fleet-wide silent
401s came from the *absence* of that export, combined with there being **no credential
preflight anywhere in daemon startup**: `grep -rn "GH_TOKEN\|GITHUB_TOKEN\|FORGE_TOKEN" loom-daemon/src`
returns **zero** hits, and there is no `gh auth status` check. The daemon boots happily,
then fails every forge call for the life of the process with nothing but per-tick noise.

**That is the bug worth fixing: startup is silent about an unusable credential.**

### Security finding (new — please fix in the same PR)

The existing forwarding mechanism writes the PAT **in plaintext into a world-readable file**.
Verified on this host:

```
$ ls -l ~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist
-rw-r--r--  1 rwalters  staff  1368 Jul 26 21:36
```

`defaults/scripts/cli/loom-daemon-start.sh:548` renders the plist with a plain `> "$PLIST_FILE"` redirect and
never chmods it (the `umask 077` at line 236 applies to the intent marker, not the plist).
Any local user can read the token. Recommending "export `GH_TOKEN` before starting the
daemon" as the headless path **without fixing this** actively makes the host less secure —
which is why it belongs in this issue rather than a follow-up.

### Scope decision — split the SSH-administration half

The issue bundles two different problems:

1. **Credential resolution** (`401 Unauthorized` on every `gh` call) — in scope here.
2. **`launchctl bootstrap gui/$UID` fails over SSH** (`error 125: Domain does not support specified action`) — *not* an auth problem. `gui/$UID` is a per-GUI-session domain that does not exist in an SSH session; the fix is a different mechanism (a `LaunchDaemon` in the `system` domain, or `launchctl asuser`), with its own security and privilege trade-offs. **Out of scope — file a follow-up issue.**

Keeping (2) here would push this past a single reviewable PR and mix a credential fix with a
supervision-model change.

### Acceptance Criteria

- [ ] **Startup credential preflight**: on boot, the daemon resolves a forge credential once and logs the outcome at `info!` — which mechanism won (env token vs. `gh` default credential store) and, for a token, a non-secret fingerprint (last 4 chars or the authenticated login), **never the token itself**.
- [ ] **Loud failure**: if no usable credential resolves, the daemon logs at `error!` with an actionable message naming both remedies (export `GH_TOKEN` before `loom-daemon-start.sh`, or unlock the login keychain from a GUI session) — instead of proceeding into silent per-tick 401s.
- [ ] The preflight is **non-fatal and bounded**: it must not block startup or hang. Same posture as the claim-reconciliation startup pass (`main.rs:635-660`) — best-effort, timeout-bounded, logged, never a startup failure. A daemon that comes up degraded and says so is better than one that refuses to start.
- [ ] **Preflight result is visible without reading logs** — surfaced in `loom-daemon status` (this is the observability gap #4069 is about; keep the shape consistent with it).
- [ ] **Plist hardening**: when the rendered plist contains any `*_TOKEN` value, it is written mode `0600`. Verify with `ls -l` after a start.
- [ ] **Docs**: `.loom/docs/github-authentication.md` gains a "Headless / daemon operation" section — that the daemon inherits its environment *from the shell that started it*, that a token must be exported **before** `loom-daemon-start.sh`, and that a bare re-render drops it (this is the same footgun as the `LOOM_WORK_FINDER` env-replay trap already documented for restarts).
- [ ] Gitea parity: the preflight covers `GITEA_TOKEN`/`FORGE_TOKEN` too, or explicitly documents that it is GitHub-only.
- [ ] No secret appears in any log line, in `loom-daemon status` output, or in an event-bus payload.

### Implementation Guidance

Put the preflight in `loom-daemon/src/main.rs` **immediately before** the claim-reconciliation
startup pass (line 635) — that block is the daemon's first `gh` consumer, so a failure there
is exactly what an operator currently experiences as silence.

Resolution order should mirror what `gh` itself does, so the log cannot disagree with reality:
`GH_TOKEN` → `GITHUB_TOKEN` → `gh`'s own credential store. The cheapest honest probe is
`gh auth status` (it reports which mechanism is active) or a single `gh api user`; prefer one
bounded call over reimplementing `gh`'s precedence in Rust, which would drift.

Reuse the existing timeout helper (`run_capture_with_timeout`, used at `main_health_gate.rs:903`)
rather than a bare `Command::output()` — an unlocked-keychain prompt is exactly the case that
can hang forever.

**Do not add a loom-managed PAT file.** The original issue floats it as an option; the plist
env path already exists, works, and reaches sweep children. A second credential store would be
a new secret-at-rest surface to secure, rotate, and document for no capability gain. If a
Builder disagrees, say so in the PR rather than implementing it silently.

### Affected Files

- `loom-daemon/src/main.rs` — add the bounded startup credential preflight before the claim-reconciliation pass (line 635); `main()` is at line 463. Line numbers verified against `origin/main` on 2026-07-27.
- `loom-daemon/src/main_health_gate.rs` — reference only: `run_capture_with_timeout` (line 903) is the bounded-subprocess helper to reuse.
- `defaults/scripts/cli/loom-daemon-start.sh` — **the canonical file to edit.** chmod the rendered plist to `0600` when it carries a token (render at line 548; the token-forwarding loop is at line 200; the `umask 077` at line 236 belongs to the intent marker and does not apply). Verified present on `origin/main`.
  - Note: `.loom/scripts/cli/loom-daemon-start.sh` is the **installed copy and is gitignored — it does not exist in a Builder worktree.** Do not try to edit it; run `./.loom/scripts/resync-installed.sh` locally afterwards to refresh it.
- `.loom/docs/github-authentication.md` — new "Headless / daemon operation" section (current options list at lines 42-57 covers interactive shells only).
- `.loom/docs/daemon-reference.md` — line 2069 already describes the env forwarding; cross-link the new section.
- `loom-daemon/src/ipc.rs` + `types.rs` — only if the preflight result is surfaced in `loom-daemon status` (see #4069 for the shape).

### Test Plan

- [ ] Automated: unit-test the credential-resolution reporter — env token present ⇒ reports the env mechanism; no env token ⇒ reports the credential-store mechanism; neither usable ⇒ error verdict. No network required; inject the probe result.
- [ ] Automated: regression guard that the preflight never panics or blocks when `gh` is missing from `PATH` or exits non-zero.
- [ ] Automated: assert no log/status/event formatter interpolates a token value (a formatting test over a sentinel secret string).
- [ ] Manual: start the daemon with a deliberately invalid `GH_TOKEN` ⇒ one `error!` at startup naming both remedies, daemon still comes up, `loom-daemon status` shows the degraded credential state.
- [ ] Manual: start with a valid `GH_TOKEN` over SSH ⇒ `info!` names the env mechanism, and a dispatched sweep's `gh` calls succeed (confirms the child inherits it).
- [ ] Manual: `ls -l ~/Library/LaunchAgents/<label>.plist` after a start with a token exported ⇒ `-rw-------`.
- [ ] Manual: start with **no** token exported on a host whose keychain is unlocked ⇒ preflight reports the credential-store mechanism and does **not** cry wolf.
- [ ] Edge case: `gh` present but hanging (simulate with a stub on `PATH`) ⇒ preflight times out, logs, and startup continues.

### Out of scope

- `launchctl bootstrap gui/$UID` over SSH (`error 125`) — needs a `LaunchDaemon`/`launchctl asuser` supervision change; file a follow-up.
- Any change to how sweep children select Claude OAuth tokens (`.loom/tokens/`) — that is a different credential system entirely.

rjwalters:
## Summary

The daemon's `gh` calls (claim reconciliation, main-health gate, metrics collection, work finder, …) have never had a startup credential check. A headless/SSH start with no exported `GH_TOKEN`/`GITHUB_TOKEN` and no unlockable GUI login keychain boots clean and then 401s silently on every forge call for the life of the process. Separately, the launchd plist that carries a forwarded token was written world-readable. This PR adds a bounded, once-at-boot credential preflight with loud logging + `loom-daemon status` visibility, and hardens the plist to `0600` when it carries a token.

## Changes

- **`loom-daemon/src/credential_preflight.rs`** (new): resolves the daemon's GitHub credential via a single bounded `gh auth status --json hosts` probe (reuses `main_health_gate::run_capture_with_timeout`, 10s bound). Reports which mechanism won (`GH_TOKEN`/`GITHUB_TOKEN`/`gh`'s own credential store) with a non-secret fingerprint (last-4 of an env token, or the authenticated login) — never the token itself. Logs `info!` on success, `error!` naming both remedies on failure. 12 unit tests cover env-token/store/none/error/timeout-style failures and assert no token value ever appears in a log message.
- **`loom-daemon/src/main.rs`**: runs the preflight once, immediately before the claim-reconciliation startup pass (the daemon's first `gh` consumer); threads the result into `IpcServer::new`; renders it in both `--json` (`credential_preflight` key) and human `loom-daemon status` output ("Forge credential: OK/DEGRADED — …").
- **`loom-daemon/src/types.rs`**: new `CredentialPreflightReport` wire type + `DaemonStatusReport.credential_preflight: Option<...>` (`#[serde(default)]`, backward-compatible with older daemon payloads).
- **`loom-daemon/src/ipc.rs`**: threads the preflight snapshot through `IpcServer` → `handle_client` → `build_daemon_status` (mirrors how `health_states`/`fallback_root` are already threaded); updates the 9 existing `build_daemon_status`/`DaemonStatusReport` test call sites with a fixture.
- **`loom-daemon/src/main_health_gate.rs`**: `run_capture_with_timeout` bumped from private to `pub(crate)` so the preflight can reuse it instead of reimplementing timeout/kill handling.
- **`defaults/scripts/cli/loom-daemon-start.sh`**: `chmod 600` the rendered launchd plist whenever it carries a forwarded `GH_TOKEN`/`GITEA_TOKEN`/`FORGE_TOKEN` (previously left at the process umask, typically world-readable `0644`).
- **`.loom/docs/github-authentication.md`**: new "Headless and SSH-only daemon operation" section covering the env-inheritance footgun, the preflight, plist hardening, and the GitHub-only scope decision (Gitea forwarding is for sweep children only — the daemon itself never calls a Gitea API).
- **`.loom/docs/daemon-reference.md`**: cross-links the new section from the existing launchd-plist writeup.

**Deliberately not done** (per the curated issue's explicit scope decision): no new loom-managed PAT file — the existing plist env-forwarding mechanism already reaches the daemon and every dispatched sweep child, so a second credential store would be new attack surface for no capability gain. The `launchctl bootstrap gui/$UID` SSH-start failure (`error 125`) is a separate supervision-model problem, not a credential problem — filed as a follow-up: #4130.

|---|---|---|
| Startup credential preflight, resolved once before claim reconciliation, `info!`/non-secret fingerprint | ✅ | `credential_preflight::run()` called in `main.rs` before the `claim_reconciliation::reconciliation_enabled()` block; `run_never_includes_a_token_looking_value_in_message` + `env_fingerprint_never_leaks_full_token` tests |
| Loud, actionable `error!` naming both remedies when no credential resolves; non-fatal/bounded | ✅ | `resolve()`'s `None`/error branches always name "export GH_TOKEN... or unlock the login keychain"; bounded by `PREFLIGHT_TIMEOUT` (10s) via reused `run_capture_with_timeout`; `run_with_missing_gh_binary_never_panics_and_reports_degraded` + `real_probe_with_missing_gh_binary_never_panics` |
| Preflight result visible in `loom-daemon status` | ✅ | `credential_preflight` field in `--json` output; "Forge credential: OK/DEGRADED — …" line in human output; `test_daemon_status_request_response_round_trip` asserts the field round-trips |
| Plist hardened to 0600 when it carries a `*_TOKEN` | ✅ | `chmod 600 "$PLIST_FILE"` gated on `env | grep -qE '^(GH_TOKEN|GITEA_TOKEN|FORGE_TOKEN)='`, matching the same pattern the forwarding loop reads |
| Docs: headless/daemon operation section + cross-link | ✅ | `.loom/docs/github-authentication.md` § "Headless and SSH-only daemon operation"; cross-linked from `.loom/docs/daemon-reference.md` |
| Gitea parity: covered or explicitly GitHub-only | ✅ | Documented GitHub-only (the daemon's own forge calls are exclusively `gh`-CLI-based; `GITEA_TOKEN`/`FORGE_TOKEN` forwarding remains for sweep children only) — no code change needed |
| No secret ever appears in logs/status/event payloads | ✅ | `run_never_includes_a_token_looking_value_in_message` asserts the exported token value never appears in `report.message`/`fingerprint`; fingerprint is always a 4-char slice computed locally, never `gh`'s own `--show-token` output (never invoked) |
| No new loom-managed PAT file | ✅ | Not implemented — reuses the existing plist env-forwarding mechanism (see "Deliberately not done" above) |

## Test Plan

- `cargo test --lib credential_preflight` — 12/12 pass (env-token present/absent/neither-usable, probe error, unparseable output, fingerprint never leaks, `run()` never panics on a missing `gh` binary)
- `cargo test --lib` (full `loom-daemon` suite) — 1230/1230 pass, including the updated `ipc::tests::test_build_daemon_status_*` and `test_daemon_status_*` round-trip/backward-compat tests
- `bash defaults/scripts/tests/test-loom-daemon-start.sh` — 11/11 pass (no regression from the plist-hardening change)
- `bash -n defaults/scripts/cli/loom-daemon-start.sh` — syntax OK
- Manual `ls -l ~/Library/LaunchAgents/<label>.plist` / wedged-`gh` / invalid-`GH_TOKEN` verification (from the curated test plan) were not re-run in this sandbox (no real launchd session here) — the automated coverage above exercises the same code paths (probe error handling, degraded-credential branch, plist chmod condition) without touching the operator's real LaunchAgents.

Closes #4005

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -2088,6 +2088,19 @@ after migration reported `13 seen, 3 dispatched, 0 error(s)`.
   `--health-gate` / `--from-config` semantics are preserved exactly — the
   launchd job never sees a wider or narrower autonomy configuration than a
   plain nohup start would have resolved.
+- **Token hardening + startup credential preflight (#4005)**: because a
+  forwarded `GH_TOKEN`/`GITEA_TOKEN`/`FORGE_TOKEN` is embedded verbatim in the
+  plist, `loom-daemon-start.sh` writes the file mode `0600` whenever it
+  carries one of those vars (otherwise it inherits the process umask —
+  typically world-readable `0644`). Separately, the daemon resolves its own
+  forge credential once at boot — before its first `gh` consumer — and logs
+  the outcome loudly (`info!`/`error!`, never the token value); the result is
+  also surfaced in `loom-daemon status` (`credential_preflight` in `--json`,
+  a "Forge credential: OK/DEGRADED" line in the human view). See
+  [`github-authentication.md` § Headless and SSH-only daemon operation](github-authentication.md#headless-and-ssh-only-daemon-operation-4005)
+  for the operator-facing walkthrough — this is the fix for the
+  keychain-unlock-dependency failure mode: a daemon started over SSH with no
+  exported token used to boot clean and then 401 silently on every forge call.
 - **`RunAtLoad=true` / `KeepAlive={SuccessfulExit:true}`** (#4054): `RunAtLoad=true`
   means the daemon survives a reboot/re-login, not just the death of one particular
   session — strictly *more* durable than the pre-#3972 nohup contract (which didn't
--- a/.loom/docs/github-authentication.md
+++ b/.loom/docs/github-authentication.md
@@ -80,6 +80,77 @@ gh pr list --repo <owner>/<repo> --limit 1
 
 If `gh auth status` shows the default credential instead of `GH_TOKEN`, verify the variable is exported in the same shell session.
 
+## Headless and SSH-only daemon operation (#4005)
+
+`loom-daemon`'s own forge calls (claim reconciliation, the main-health gate,
+metrics collection, the work finder, …) all shell out to `gh`, which resolves
+credentials the same way an interactive shell does: `GH_TOKEN` env var →
+`GITHUB_TOKEN` env var → `gh`'s own credential store (the macOS login
+**keychain**, or `~/.config/gh/hosts.yml` on Linux). The keychain only unlocks
+for processes running in the user's **GUI login session** — a daemon started
+over SSH with a clean environment, or from a headless server with no
+interactive login, cannot unlock it. Without an env-var token, every `gh` call
+the daemon makes will `401`.
+
+**The fix is an exported token, not a new credential store.** Loom does not
+provision a separate daemon-managed PAT file — `export GH_TOKEN` (or
+`GITHUB_TOKEN`) before starting the daemon, and the existing forwarding
+mechanism carries it the rest of the way:
+
+```bash
+# On the headless / SSH-only host, before starting the daemon:
+export GH_TOKEN=github_pat_xxxxxxxxxxxxxxxxxxxx
+./.loom/scripts/cli/loom-daemon-start.sh
+```
+
+`loom-daemon-start.sh` forwards any exported `GH_TOKEN` / `GITHUB_TOKEN` /
+`GITEA_TOKEN` / `FORGE_TOKEN` into the launchd plist's `EnvironmentVariables`
+(macOS) or the backgrounded process's inherited environment (`--no-launchd` /
+Linux) — so the daemon **and every sweep child it dispatches** see the token,
+with no per-sweep configuration needed. The daemon inherits its environment
+**from the shell that started it** — export the token *before* invoking
+`loom-daemon-start.sh`, not after. A later `loom-daemon-update.sh` restart
+re-renders the plist from the *current* shell's environment, so an
+already-running daemon does not silently lose a token that was exported only
+in a now-closed session (the same footgun `LOOM_WORK_FINDER` / autonomy-flag
+env replay has — see [`daemon-reference.md`](daemon-reference.md)).
+
+**Startup credential preflight.** The daemon resolves its forge credential
+once at boot, immediately before its first `gh` consumer (the claim
+reconciliation startup pass), and reports the outcome — `info!` naming which
+mechanism won (`GH_TOKEN`, `GITHUB_TOKEN`, or `gh`'s own credential store) plus
+a non-secret fingerprint (never the token itself), or `error!` naming both
+remedies (export a token, or unlock the login keychain from a GUI session)
+when nothing resolves. This turns the pre-#4005 failure mode — a daemon that
+boots clean and then 401s silently on every forge call for the life of the
+process — into a single loud, actionable line. The result is also visible
+without reading logs via `loom-daemon status` ("Forge credential: OK/DEGRADED
+— …"); see [`daemon-reference.md`](daemon-reference.md) for the field shape.
+The preflight is read-only, bounded (never blocks daemon startup), and never
+logs, prints, or serializes a token value.
+
+**GitHub only.** The preflight covers `GH_TOKEN`/`GITHUB_TOKEN` because the
+daemon's own forge calls are exclusively `gh`-CLI-based. `GITEA_TOKEN` /
+`FORGE_TOKEN` forwarding still happens (for dispatched sweep children targeting
+a Gitea-backed repo — see [`forge-authentication.md`](forge-authentication.md))
+but the daemon process itself never calls a Gitea API, so there is nothing to
+preflight for it.
+
+**Plist permissions.** Because the rendered launchd plist embeds the token
+value verbatim in `EnvironmentVariables`, `loom-daemon-start.sh` hardens the
+file to mode `0600` whenever it carries a `GH_TOKEN`/`GITEA_TOKEN`/
+`FORGE_TOKEN`, so a local user other than the daemon's owner cannot read the
+PAT out of `~/Library/LaunchAgents`.
+
+**Out of scope**: `launchctl bootstrap gui/$UID` (the launchd domain
+`loom-daemon-start.sh` uses on macOS) fails over SSH — `gui/$UID` is a
+per-GUI-session domain that does not exist in an SSH session — so a daemon
+that isn't already running cannot be *started* remotely on macOS today even
+with a valid token exported. That is a supervision-model gap (a `LaunchDaemon`
+or `launchctl asuser` mechanism), not a credential problem; restarting an
+*already-running* daemon (`loom-daemon-update.sh`) and running fully headless
+on Linux (`--no-launchd` / nohup path) both work with an exported token today.
+
 ## Troubleshooting
 
 ### Token not being picked up
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -547,6 +547,18 @@ if [[ "$USE_LAUNCHD" == "true" ]]; then
 
     render_launchd_plist "$LAUNCHD_LABEL" "$DAEMON_BIN" "$REPO_ROOT" "$START_LOG" > "$PLIST_FILE"
 
+    # Harden the rendered plist when it carries a forwarded credential
+    # (#4005): the token-forwarding loop in render_launchd_plist writes any
+    # exported GH_TOKEN/GITEA_TOKEN/FORGE_TOKEN straight into
+    # EnvironmentVariables above, and the plain `>` redirect otherwise leaves
+    # the file at the process's umask (typically world-readable, 0644) --
+    # any local user could read the PAT straight out of
+    # ~/Library/LaunchAgents. Match the same env pattern the forwarding loop
+    # reads from.
+    if env | grep -qE '^(GH_TOKEN|GITEA_TOKEN|FORGE_TOKEN)=' 2>/dev/null; then
+        chmod 600 "$PLIST_FILE"
+    fi
+
     echo "Launchd label:  $LAUNCHD_LABEL"
     echo "Launchd plist:  $PLIST_FILE"
 
--- a/loom-daemon/src/credential_preflight.rs
+++ b/loom-daemon/src/credential_preflight.rs
@@ -1,0 +1,375 @@
+//! Startup forge-credential preflight (#4005).
+//!
+//! # Problem
+//!
+//! Every daemon-issued `gh` call (`claim_reconciliation`, `main_health_gate`,
+//! `metrics_collector`, the work finder, …) is a plain `Command::new("gh")`
+//! with no `env_clear` — so it authenticates however `gh` itself resolves
+//! credentials in the daemon's process environment: `GH_TOKEN` /
+//! `GITHUB_TOKEN`, falling back to `gh`'s own credential store (the macOS
+//! login keychain, or `~/.config/gh/hosts.yml` on Linux). That plumbing
+//! already works headlessly *if* a token happens to be exported before
+//! `loom-daemon-start.sh` runs — the actual defect is that nothing checks at
+//! startup, so a daemon started with neither an exported token nor an
+//! unlocked keychain (e.g. over SSH, clean environment) boots clean and then
+//! 401s on every forge call for the life of the process, with nothing but
+//! silent per-tick noise to show for it.
+//!
+//! # What this module does
+//!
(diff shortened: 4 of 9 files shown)


rjwalters:
✅ **Approved!**

Verified all six load-bearing acceptance criteria from the curated issue against the actual code (not just the PR's claims table):

1. **Bounded/non-fatal preflight** — `credential_preflight::run()` calls `gh auth status --json hosts` through the reused `main_health_gate::run_capture_with_timeout` (10s hard bound: polls `try_wait`, kills + reports timeout on expiry). Failure paths (`Err(probe)`, unparseable JSON, no active host, non-success state) all return a `CredentialPreflightReport{ok: false, ...}` — never a panic, never a process exit. It's called synchronously (not via `spawn_blocking`) directly in async `main()`, which diverges from the periodic gate loop's documented pattern in `main_health_gate.rs` — but it's consistent with the `claim_reconciliation` call immediately following it in the same function (same tradeoff, same file), happens before any other concurrent runtime work is spun up, and is hard-bounded to 10s. Not blocking.

2. **No secret leakage** — read the full `resolve()`/`run()`/`env_fingerprint()` implementation. `fingerprint` is only ever the last-4 chars of an env token or the `gh` login name; `message`/`mechanism` never carry a token. Confirmed with the 12 dedicated unit tests (all pass) plus `main.rs`'s `print_status_json`/`print_status_human` and `ipc.rs`'s `build_daemon_status` — none of these paths touch the raw token, and the preflight result is never published to the event bus.

3. **Plist `chmod 0600`** — read the actual shell logic (`defaults/scripts/cli/loom-daemon-start.sh`): `if env | grep -qE '^(GH_TOKEN|GITEA_TOKEN|FORGE_TOKEN)=' ...; then chmod 600 "$PLIST_FILE"; fi`, applied after `render_launchd_plist` writes the file. I validated the regex in isolation (`env -i GH_TOKEN=... `, `MY_GH_TOKEN=...`, `FOO=bar`, `GITEA_TOKEN=...`, `FORGE_TOKEN=...`) — it correctly anchors on the exact var names and doesn't false-positive on substrings. One gap: no automated test exercises this path at all (the existing `test-loom-daemon-start.sh` suite forces `--no-launchd` throughout, so the launchd/plist branch — chmod included — was untested before this PR too, and remains so). Pre-existing infra gap, not a regression; worth a follow-up but not blocking.

4. **Precedence matches `gh`'s own** — rather than reimplementing GH_TOKEN→GITHUB_TOKEN→store precedence, the code asks `gh auth status --json hosts` directly and reads its `tokenSource`/`active` fields, so it can't disagree with reality by construction. I verified this against a real `gh auth status --json hosts` locally: baseline reports `"tokenSource":"keyring"`, and exporting `GH_TOKEN` flips the active entry to `"tokenSource":"GH_TOKEN"` (with `state":"error"` for an invalid token) — exactly matching the code's assumptions.

5. **No new PAT file** — confirmed via `git diff --stat` and grep: no new credential-store file is written anywhere in the diff; the PR body's "Deliberately not done" section is accurate.

6. **Gitea/GITHUB-only scope** — honestly documented, not silently ignored. Both the module doc comment and `.loom/docs/github-authentication.md` state explicitly that the daemon's own forge calls are exclusively `gh`-CLI-based, and `GITEA_TOKEN`/`FORGE_TOKEN` forwarding remains for dispatched sweep children only.

## Test Execution

- `cargo test --lib credential_preflight` — ✅ 12/12 pass
- `cargo test --lib` (full `loom-daemon` suite) — ✅ 1230/1230 pass
- `cargo clippy --all-targets -- -D warnings` — ✅ clean
- `cargo fmt -- --check` — ✅ clean
- `bash defaults/scripts/tests/test-loom-daemon-start.sh` — ✅ 11/11 pass
- `gh pr checks 4131` — ✅ all required checks pass (Rust Unit Tests, Rust Linting & Build Check, Installer Integration Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes)
- `mergeStateStatus` — ✅ CLEAN

Follow-up #4130 for the `launchctl bootstrap gui/$UID`-over-SSH supervision gap is correctly scoped as out-of-scope per the curated issue, and accurately describes the remaining problem.

Non-blocking suggestion for a future pass: add a targeted test for the plist `chmod 0600` condition (e.g. by extracting `render_launchd_plist` + the chmod gate into something testable without a live `launchctl`), since the launchd code path currently has zero automated coverage.

Issue #4055, opened 2026-07-27
feat(daemon): autonomous self-update loop — periodic staleness check with clean-tree gate, settle window, backoff, and status


rjwalters:
> ## ⚠️ Builder: read this body, not the comment thread
>
> This issue accumulated **8 comments** across 2026-07-27/28, including two pairs of
> directly contradictory verdicts (curator "not applying `loom:curated`" @00:49 vs.
> "applying `loom:curated`" @00:50; Champion **APPROVED** @00:54 vs. **DEFERRED**
> @00:55). Every dependency raised anywhere in that thread has since merged, and the
> file:line anchors in the 18:48 curator table have all drifted.
>
> **This body is the single source of truth as of the 2026-07-28 re-curation below.**
> Treat all prior comments as historical.

## Context

Phase 3 of #4017 (*auto-rebuild and restart when the running binary is stale*) — the loop that finally closes the self-repair cycle. This issue is only the *deciding* and *sequencing* logic; it reuses `loom-daemon-update.sh` for rebuild and #4090's drain primitive for restart, rather than reimplementing either.

## Dependencies

All four dependencies raised across the body and comment thread are **merged and verified in source on `origin/main`** (not merely closed — each was checked against the shipped code):

- [x] **#4053** — rebuild→provision is self-verifying. Merged as **PR #4074**. Verified: `verify_destination_binary()` at `defaults/scripts/cli/loom-daemon-update.sh:187`, pre-provision built-commit assertion at `:538-560`, terminal exit code `4` documented at `:93-95`.
- [x] **#4054** — supervised restart primitive proven under launchd. Merged as **PR #4077** (`KeepAlive:SuccessfulExit`). Follow-up #4126 (`b11c1865`) re-renders the supervised plist on the refused-restart fallback.
- [x] **#4042** — `loom-daemon-update.sh` can now manage a launchd-installed daemon (raised in the 2026-07-27T21:00 comment; without it this loop would rebuild, print *"loom-daemon was not running — nothing to restart"*, exit 0, and never ship a binary). Merged as **PR #4109**.
- [x] **#4090** — scheduled drain-and-roll (raised in the 2026-07-27T21:21 comment, and the sole stated reason for the Champion **DEFERRED** verdict @00:55). Merged as **PR #4134**, commit `bbc9bc3f`. **That deferral rationale is now discharged.**

Nothing blocks this issue. Parent #4017 is transitively unblocked once this closes.

## What exists today

[...]

PR #4134, opened 2026-07-28
feat(daemon): scheduled drain-and-restart primitive

### LINKED ISSUE #4090: feat(daemon): scheduled drain-and-roll — finish in-flight sweeps, then restart onto the new binary
## Ask

An operator should be able to **schedule a roll** — tell the daemon "pick up the new binary" — and have it cleanly finish all in-flight work before restarting, with no sweep killed and no orphan left behind.

Today rolling is a choice between two bad options, and the operator has to know which:

| Method | In-flight sweeps | Notes |
|---|---|---|
| `launchctl bootout` + `bootstrap` | **all killed** | bootout tears down the whole job tree; every sweep child is a direct child of the job (verified: 9 `claude` pids with `ppid` = daemon pid) |
| `loom-daemon restart` (#4054/#4077) | survive, but become **orphans** | clean `exit(0)` doesn't tear down the tree, so children reparent — but the in-memory sweep registry dies with the process, so the relaunched daemon has no idea they exist |
| `loom-daemon-update.sh` | n/a | doesn't work on a launchd host at all — see #4042 |

Neither is "finish what you started, then roll."

## Why the orphan case is not good enough

`RestartDaemon` logs *"In-flight sweeps survive by design"* (`ipc.rs:394`), which is true at the process level and misleading at the system level. After the relaunch those sweeps are:

- absent from `list_sweeps` / `loom-daemon status` — invisible to the operator and to the concurrency cap, so the new daemon will over-dispatch against a cap that doesn't count them;
- outside the watchdog's and reaper's view, so nothing will ever clean them up;
- still holding `loom:building` on their issues, which the startup `claim_reconciliation` pass may reclaim and **re-dispatch as duplicates** — two sweeps on one issue, racing to open a PR.

So the current primitive trades "work destroyed" for "work untracked." A drain removes the choice.

## Proposed design

Add a **drain mode** to the existing restart primitive rather than a new lifecycle:

`loom-daemon restart --drain [--timeout <secs>] [--force-after-timeout]`, backed by an IPC request variant (e.g. `RestartDaemon { drain: bool, timeout_secs: Option<u64> }`).

On receipt the daemon enters a **DRAINING** state and:

1. **Stops admitting new work** — work finder stops dispatching, epic supervisor stops scheduling new phases, role runner stops starting new role ticks. Already-running ones are left alone.
2. **Waits for the sweep registry to empty**, with the existing reaper doing the accounting. Bounded by `--timeout` (suggest a default in the tens of minutes, since a sweep is ~10–20 min).
3. **Exits `EXIT_RESTART` (0)** for the `KeepAlive:SuccessfulExit` relaunch, exactly as today.

Behaviour worth pinning down:

- **Timeout reached** — default to *refusing* the restart (stay up, log loudly, report via status) rather than killing sweeps. `--force-after-timeout` opts into cancelling the stragglers via the existing `cancel_sweep` path. Fail-safe, not fail-destructive.
- **Unsupervised host** — same refusal contract as today: reply `scheduled: false` and keep running, so a caller can detect that nothing happened.
- **Observability** — `loom-daemon status` should show `DRAINING (n sweeps remaining, timeout in Ns)`. A drain that quietly hangs is worse than no drain. Emit a `daemon.drain.*` event so `tail_event_bus` sees the transition.
- **Cancellable** — an operator who changes their mind should be able to abort the drain and resume normal dispatch without a restart.

## Relationship to existing work

- **#4054 / #4077** shipped the restart primitive this builds on — `KeepAlive: {SuccessfulExit: true}`, `EXIT_RESTART`, and the `LOOM_DAEMON_SUPERVISOR=launchd` proof-of-supervision gate. No new launchd mechanism is needed.
- **#4055** (Phase 3, the autonomous self-update loop) **needs this**. An unattended loop that rolls the daemon must not kill in-flight sweeps or strand orphans; a drain is the safety property that makes auto-rolling acceptable. Suggest making this a dependency of #4055 alongside #4042.
- **#4042** is the sibling gap on the script side (`loom-daemon-update.sh` can't restart a launchd daemon at all).

## Bootstrap problem — worth calling out

This cannot be used for the *next* roll on the current host. The running daemon predates #4077, and the installed plist still has `KeepAlive => false` (verified via `plutil -p`), so there is no supervised relaunch to exit into. The first roll after #4077 must be done by hand **and must re-render the plist** (via `loom-daemon-start.sh`) so the new `KeepAlive: {SuccessfulExit: true}` and `LOOM_DAEMON_SUPERVISOR=launchd` land. Every roll after that can be graceful.

## Acceptance criteria

1. `loom-daemon restart --drain` stops new dispatch immediately and restarts only after the last in-flight sweep completes.
2. No sweep is killed, and after the relaunch `list_sweeps` is consistent with reality — zero orphans.
3. Reaching the timeout without `--force-after-timeout` leaves the daemon running and reports why; it never silently restarts or silently gives up.
4. `loom-daemon status` surfaces the DRAINING state with a remaining count.
5. On an unsupervised host the request is refused with `scheduled: false`, unchanged from #4054.
6. A drain can be aborted, returning the daemon to normal dispatch.

---

All file/line references verified against `origin/main` @ `a6ef9a89`.

### Finding 1 — half of the drain already exists: the main-health gate is the same primitive

The ask reads as though "stop admitting new work" is new machinery. It is not. The
main-health gate (#3812) already implements exactly this contract — `main.rs:2358` renders
it as:

> `HALTED (main verified red — new dispatch paused; in-flight sweeps keep running)`

The plumbing is an `Arc`-shared boolean consulted by the producers:

| Producer | How it is gated today | Ref |
|---|---|---|
| Work finder | `tick(source, dispatcher, max_concurrent, halted)` returns early when `halted` | `work_finder.rs:400`, `:409` |
| Epic supervisor | `health_state: Option<Arc<MainHealthState>>`; `tick` dispatches nothing when `is_halted()` | `epic_supervisor.rs:475`, `:532` |
| Role runner | **not gated at all** — see Finding 2 | — |

**Do not reuse `MainHealthState` itself.** The gate loop owns that flag and clears it on the
next Green verdict (`apply_gate_outcome`, `main_health_gate.rs:2014`), so a drain that set it
would be silently un-set by a build-gate tick. Use a **separate** `Arc<AtomicBool>` (drain
flag) and OR it in at the same call sites. That is a handful of lines per producer, not new
architecture — which is what keeps this issue inside one PR.

### Finding 2 — role ticks are **not** in the sweep registry, so "wait for the registry to empty" does not drain them

This is the correctness gap in the proposed design, and it is live on this host.

- Verified: `grep -n 'WorkspaceHealthStates|is_halted|health_states' loom-daemon/src/role_runner.rs`
  returns **nothing**. `spawn_multi_role_task(spec, fallback_root, interval)` (`role_runner.rs:656`)
  takes no shared state at all — there is no halt input to pass.
- Role ticks shell out to `spawn-claude.sh -p "/<role>"` directly (`role_runner.rs:294`, `:354`)
  and are **never entered into `SweepRegistry`**. The module doc says so explicitly
  (`role_runner.rs:22-31`): per-sweep lifecycle roles go through `sweep_registry`; the
  standalone periodic roles this module runs do not.
- A role tick runs under `DEFAULT_ROLE_TIMEOUT = 1800s` (`role_runner.rs:108`) — **up to 30
  minutes**, comparable to a sweep.
- This repo's `.loom/config.json` has `autonomous.roleRunner.enabled = true` with
  `roles: ["curator", "champion"]`. So on the very host this feature targets, a drain that
  only watches the sweep registry can restart **while a Champion tick is mid-merge**.

That directly contradicts AC2 ("no sweep is killed, zero orphans"). Resolve it explicitly
rather than by omission — see the recommended scope split below.

### Finding 3 — `RestartDaemon` is a **unit** variant; adding fields is a wire-schema break at exactly the wrong moment

`Request` is adjacently tagged (`types.rs:15-17`: `#[serde(tag = "type", content = "payload")]`)
and `RestartDaemon` is a unit variant (`types.rs:380`), so it serializes as
`{"type":"RestartDaemon"}` with **no** `payload` key. There is a round-trip test asserting that
exact string (`ipc.rs:2885-2889`).

Converting it to `RestartDaemon { drain: bool, timeout_secs: Option<u64> }` — the ask's
suggestion — means an older `loom-daemon restart` client's `{"type":"RestartDaemon"}` fails to
deserialize against the new daemon. The window where CLI and daemon versions disagree is
**precisely a roll**, which is the only time this feature is used.

**Add a separate variant instead** — e.g. `DrainAndRestartDaemon { timeout_secs: Option<u64>,
force_after_timeout: bool }` — and leave `RestartDaemon` byte-for-byte untouched. Treat a
change to the `ipc.rs:2885` assertion as the tell that the wrong approach was taken.

### Finding 4 — the current restart handler exits **inline**; a drain cannot, so the exit must move

`RestartDaemon` is intercepted inside the per-connection loop (`ipc.rs:386-401`): it builds the
decision, writes, flushes, then calls `std::process::exit(EXIT_RESTART)` **in the connection
handler**. A drain must ack immediately and exit *minutes later*, so:

- `build_restart_decision()` (`ipc.rs:90`, returns `(Response, do_exit)`) needs a drain sibling
  that returns "accepted, do **not** exit now".
- A new supervising task owns the eventual `std::process::exit(EXIT_RESTART)` once the drain
  completes. It must still flush/remove the socket the same way the normal shutdown path does.
- **Check `detect_supervisor()` (`ipc.rs:74`) up front, before entering DRAINING** — AC5's
  refusal must be immediate. Draining for 20 minutes and *then* discovering there is no
  supervisor to relaunch into is the worst possible ordering.

### Finding 5 — count in-flight across **all** managed roots, not the primary registry

`build_daemon_status` (`ipc.rs:~340-365`) already computes the correct quantity: for every root
in `WorkspaceRegistry::effective_roots(fallback_root)`, `sr.list(None).filter(|i| !i.state.is_terminal())`.
Multi-workspace support (#3930) means a drain that reads only the primary registry will report
zero and restart while another managed repo still has live sweeps. **Reuse the
`build_daemon_status` accounting** rather than writing a second, subtly different counter.

### Finding 6 — the event taxonomy is pinned; `daemon.drain.*` needs explicit authorization here

`event_bus.rs:20-46` states: *"New topics require a follow-up issue — the taxonomy is
intentionally pinned"*, and every existing addition names its authorizing issue (`epic.issue.*`
← #3873, `daemon.capacity.advisory` ← #3902). So this issue is the authorization for
`daemon.drain.*`, and the Builder must add the rows to **both** tables:

- the module-doc table in `event_bus.rs:24-35`
- the mirrored table in `.loom/docs/daemon-reference.md:115`

Pin the exact topics in the PR (suggest `daemon.drain.started` / `daemon.drain.completed` /
`daemon.drain.aborted` / `daemon.drain.timeout`) rather than leaving them to the implementation.

### Finding 7 — `DRAINING` in `DaemonStatusReport` is a public wire-schema addition; follow the `#[serde(default)]` convention

`DaemonStatusReport` (`types.rs:744`) is the MCP/CLI wire schema. Every recent field addition
carries `#[serde(default)]` for forward/backward compat with an explicit rationale comment —
`cpu_headroom` (`:766`), `logical_cpus` (`:772`), `cpu_idle_fraction` (`:791`), `capacity_bound`
(`:801`). A new `drain: Option<DrainStatus>` (or flat `draining: bool` +
`drain_deadline: Option<DateTime<Utc>>`) must do the same, or an older `loom-daemon status`
client breaks against a newer daemon — again, mid-roll.

### Finding 8 — the bootstrap caveat in the body is now partly stale

- **#4042 is CLOSED** (shipped as PR #4109) — `loom-daemon-update.sh` can now manage a
  launchd-installed daemon. The body's third table row ("doesn't work on a launchd host at
  all") is no longer true.
- **#4077 is MERGED** — `KeepAlive:{SuccessfulExit:true}` + `LOOM_DAEMON_SUPERVISOR=launchd`
  are in `loom-daemon-start.sh`'s rendered plist.
- What **remains** true: the *running* binary on this host predates both, and the *installed*
  plist still has the old `KeepAlive => false`. So the first roll after this lands must still
  be done by hand via `loom-daemon-start.sh` (to re-render the plist), and only rolls after
  that can be graceful. Keep that paragraph; correct the `#4042` row.

### Recommended scope

**In scope (one PR):**

1. New `Request::DrainAndRestartDaemon { timeout_secs, force_after_timeout }` variant +
   `Response::DaemonDrain { accepted, supervisor, in_flight, message }` (Finding 3).
   CLI: `loom-daemon restart --drain [--timeout <secs>] [--force-after-timeout]`
   (`main.rs:192`, `:1617`), and `loom-daemon restart --abort-drain` for AC6.
2. Daemon-global `Arc<AtomicBool>` drain flag, OR'd into the existing halt checks at
   `work_finder.rs:409` and `epic_supervisor.rs:532`, **and** threaded into
   `role_runner::spawn_multi_role_task` (`:656`) as a new gate — role ticks must stop
   *starting* (Finding 2). Supervisor proof checked up front (Finding 4).
3. A drain-supervisor task that polls the cross-root in-flight count (Finding 5) and owns the
   eventual `std::process::exit(EXIT_RESTART)`; on timeout without `--force-after-timeout`
   it clears the drain flag, logs loudly, and stays up (AC3).
4. `DaemonStatusReport` drain fields + `loom-daemon status` rendering (Finding 7, AC4).
5. `daemon.drain.*` events in both taxonomy tables (Finding 6).
6. Docs: `.loom/docs/daemon-reference.md` §Supervised restart primitive (`:2157-2185`) and the
   "sweeps survive, they are not drained" paragraph (`:1871-1878`), which this issue directly
   amends.

**Explicitly out of scope — file as a follow-up:**

- **Waiting for in-flight *role ticks* to finish.** Stopping new role ticks (item 2) is cheap
  and in scope; *awaiting* one is not, because role ticks have no registry entry to poll
  (Finding 2). Document the residual: a drain can complete while a role tick is still running,
  which is bounded by `DEFAULT_ROLE_TIMEOUT` (1800s) but is not zero. If the Builder finds
  awaiting them is trivial (the tick task is already awaited in-process), do it — but do not
  build a role registry to achieve it.
- Any change to `SweepRegistry::cancel` (`sweep_registry.rs:2289`) beyond calling it from the
  `--force-after-timeout` path.
- Auto-triggering a drain from anything (that is #4055).

### Scope guard

Target **≤ 7 files and ≤ 450 LOC including tests**. If the implementation starts adding a
registry for role ticks, touching `claim_reconciliation.rs`, or changing `SweepRegistry::cancel`'s
signature, **stop and split** — those are signals the drain is being built at the wrong layer.

## Dependencies

- [x] #4077 — supervised restart primitive (MERGED; this builds directly on `EXIT_RESTART`,
      `detect_supervisor`, and the `KeepAlive:SuccessfulExit` plist)
- [x] #4042 — `loom-daemon-update.sh` launchd management (CLOSED, shipped as PR #4109)

Both complete — this issue is **not** blocked.

**Downstream**: #4055 (autonomous self-update loop, currently `loom:blocked`) should take this
as a dependency. An unattended roll that kills in-flight sweeps or strands orphans is not
acceptable; the drain is the safety property that makes auto-rolling defensible.

## Affected Files

- `loom-daemon/src/types.rs` — new `Request::DrainAndRestartDaemon` variant next to
  `RestartDaemon` (`:380`); new `Response::DaemonDrain` next to `DaemonRestart` (`:521`); drain
  fields on `DaemonStatusReport` (`:744`) with `#[serde(default)]` per the `capacity_bound`
  precedent (`:801`). **Do not modify `RestartDaemon`** (Finding 3).
- `loom-daemon/src/ipc.rs` — drain sibling of `build_restart_decision` (`:90`); intercept the
  new variant alongside the existing `RestartDaemon` interception (`:386-401`) but **without**
  the inline `std::process::exit` (Finding 4); reuse `build_daemon_status`'s cross-root in-flight
  accounting (`:~340-365`) for the drain counter (Finding 5). Leave the `{"type":"RestartDaemon"}`
  round-trip assertion (`:2885`) untouched.
- `loom-daemon/src/main.rs` — `Commands::Restart` (`:192`) gains `--drain` / `--timeout` /
  `--force-after-timeout` / `--abort-drain`; `handle_restart_command` (`:1617`) sends the new
  request; drain state rendered in `handle_status_command` (`:1956`) next to the gate status
  (`format_gate_status`, `:~2340`); drain flag constructed near the autonomous-loop wiring
  (`:810` work finder, `:754` epic supervisor, `:978` role runner) and the drain-supervisor task
  spawned there.
- `loom-daemon/src/work_finder.rs` — OR the drain flag into the existing early return at `:409`
  (`tick`'s `halted` param, `:400`). Prefer widening the existing `halted` input over a new one.
- `loom-daemon/src/epic_supervisor.rs` — same, at the `is_halted()` check (`:532`); the
  `health_state: Option<Arc<MainHealthState>>` field (`:475`) is the shape to mirror.
- `loom-daemon/src/role_runner.rs` — **new** gate: `spawn_multi_role_task` (`:656`) currently
  takes no shared state; add the drain flag and check it before each `invoke` (Finding 2).
- `loom-daemon/src/event_bus.rs` — add `daemon.drain.*` rows to the pinned taxonomy table
  (`:24-35`) citing this issue as the authorization (Finding 6).
- `.loom/docs/daemon-reference.md` — event-taxonomy table (`:115`); §Supervised restart
  primitive (`:2157-2185`); the "**Shutdown decision — sweeps survive, they are not drained**"
  paragraph (`:1871-1878`), which this issue amends.
- `loom-daemon/src/sweep_registry.rs` — **reference only** except for the
  `--force-after-timeout` path, which calls the existing `cancel` (`:2289`). Do not change its
  signature.
- `docs/design/supervised-restart-primitive.md` — **reference only**; the design record for
  #4054/#4077. Worth reading before implementing (it records why `exec` self-replacement and a
  detached helper were rejected), and worth a short appendix noting the drain extension.

## Test Plan

Rust tests belong in the inline `#[cfg(test)]` modules of the files they cover, alongside the
closest prior art: `test_tick_skips_all_dispatch_while_main_halted` (`epic_supervisor.rs:1641`)
and the restart-decision tests (`ipc.rs:2879`, `:2934`).

- [ ] **AC1 — drain stops new dispatch immediately.** With the drain flag set, `work_finder::tick`
      returns early (mirror the existing halted test) and `EpicSupervisor::tick` dispatches
      nothing (mirror `:1641`). Assert `seen` still reflects the backlog, so the "backlog is N
      but draining" log stays truthful.
- [ ] **AC1 — role ticks stop starting** (Finding 2). With the drain flag set,
      `spawn_multi_role_task`'s loop performs **zero** `invoke` calls. Use the existing fake
      `RoleInvocationRunner` + `wait_for_calls` harness (`role_runner.rs:1184`). This test does
      not exist in any form today and is the highest-value new coverage.
- [ ] **AC2 — restart only after the last sweep completes.** Drain supervisor with a fake
      in-flight count that goes 2 → 1 → 0: it must not signal exit until 0, then exit exactly
      once. Assert it is not fooled by *terminal* (`Exited`/`Crashed`) entries lingering in the
      registry (`is_terminal()` filter, Finding 5).
- [ ] **AC2 — cross-root counting** (Finding 5). Two managed roots, sweeps live in the
      **secondary** only ⇒ drain does **not** complete. Guards the most likely implementation
      shortcut (reading the primary registry).
- [ ] **AC3 — timeout without `--force-after-timeout`.** Deadline passes with in-flight > 0 ⇒ no
      exit, drain flag **cleared** (dispatch resumes), a WARN logged, and status reports the
      reason. Explicitly assert the daemon did *not* exit and did *not* cancel anything.
- [ ] **AC3 — timeout with `--force-after-timeout`.** Stragglers are cancelled via the existing
      `SweepRegistry::cancel` path (`:2289`), then exit.
- [ ] **AC4 — status surfaces DRAINING** with a remaining count and time-to-deadline; and the new
      `DaemonStatusReport` fields round-trip, with a **pre-drain JSON payload (fields absent)
      still deserializing** (Finding 7 — mirror the `capacity_bound` compat rationale, `:801`).
- [ ] **AC5 — unsupervised refusal is immediate** (Finding 4). `LOOM_DAEMON_SUPERVISOR` unset ⇒
      reply `accepted: false`, **and the drain flag is never set** (dispatch is not paused).
      Assert both — pausing dispatch and then refusing would be a silent outage. Note
      `ipc.rs:2939` flags the supervisor test as the sole `LOOM_DAEMON_SUPERVISOR` mutator
      (env-var test isolation); keep that property.
- [ ] **AC6 — abort restores normal dispatch.** After abort, the drain flag is clear, the
      supervisor task is gone (no later exit fires), and `work_finder::tick` dispatches again.
      Assert **no** exit happens after an abort even once in-flight later reaches zero — the
      "abort then the queue empties anyway" race is the subtle one.
- [ ] **Wire compat** (Finding 3). `{"type":"RestartDaemon"}` still deserializes to
      `Request::RestartDaemon` unchanged, and the new variant round-trips. The existing
      assertion at `ipc.rs:2885` must pass **unmodified**.
- [ ] **No-regression** — the `RestartDaemon` decision tests (`ipc.rs:2879`, `:2934`), the
      main-health-gate halt tests (`epic_supervisor.rs:1641-1682`), and the work-finder halted
      tests pass unmodified.
- [ ] **Edge cases**:
  - [ ] Drain requested with **zero** in-flight ⇒ exits promptly (do not wait a full poll
        interval).
  - [ ] A second drain request while already DRAINING ⇒ idempotent ack, does not stack
        supervisors or shorten/extend the deadline unexpectedly (state the chosen semantics).
  - [ ] `Shutdown` / SIGTERM **during** a drain ⇒ takes precedence and exits `143` (stay down),
        not `EXIT_RESTART` (`ipc.rs:54-63`).
  - [ ] Drain on a host with `roleRunner` **disabled** ⇒ behaviour unchanged (this repo has it
        *enabled*, so both paths need coverage).
- [ ] **Manual verification** on the launchd host, after the plist has been re-rendered by
      `loom-daemon-start.sh` (see the Rollout note):
  - [ ] With ≥2 sweeps in flight, `loom-daemon restart --drain`; confirm `loom-daemon status`
        shows DRAINING with a decreasing count, **no new dispatch** occurs, and the pid changes
        only after the last sweep completes.
  - [ ] After the relaunch, `loom-daemon status` shows zero orphans and
        `claim_reconciliation` reclaims nothing (AC2).
  - [ ] `loom-daemon restart --drain --timeout 60` against a long sweep ⇒ refusal, daemon still
        up, dispatch resumed (AC3).
  - [ ] `--abort-drain` mid-drain ⇒ dispatch resumes and no later restart fires (AC6).

## Rollout note

This is daemon-side Rust, so it needs a **rebuild + restart** to take effect — and it cannot be
used for its own first roll. The running binary predates #4077 and the installed plist still
carries `KeepAlive => false`, so the first roll after this merges must be done by hand via
`loom-daemon-start.sh` (which re-renders the plist with `KeepAlive:{SuccessfulExit:true}` and
`LOOM_DAEMON_SUPERVISOR=launchd`). Every roll after that can be a drain. Do not read a failed
`--drain` against an un-rolled daemon as "the feature does not work".

## Related work

- **#4054 / #4077** — the supervised restart primitive this extends. `docs/design/supervised-restart-primitive.md`
  is the design record.
- **#4042** — CLOSED (PR #4109); corrects the body's third table row.
- **#4055** — the autonomous self-update loop; should depend on this.
- **#3812 / #3885** — the main-health gate, whose halt plumbing is the model for the drain flag
  (Finding 1).
- **#3930** — multi-workspace registries; the reason in-flight must be counted across all roots
  (Finding 5).
- **#4015** — the role runner, which has no halt gate today (Finding 2).
- **#3873 / #3902** — precedent for authorizing a new event topic against a specific issue
  (Finding 6).

rjwalters:
## Summary

Adds a **scheduled drain-and-restart** to the supervised restart primitive (#4054/#4077): `loom-daemon restart --drain` stops admitting new work immediately, waits for every in-flight sweep to finish, then exits `EXIT_RESTART` for the launchd relaunch — **no sweep killed, no orphan left behind**. This is the safety property that makes auto-rolling (#4055) defensible.

Follows the curated scope precisely: the drain reuses the main-health gate's proven halt plumbing (Finding 1) rather than inventing new machinery, adds a **separate** IPC variant so the untouched `{"type":"RestartDaemon"}` wire shape keeps deserializing (Finding 3), threads a halt gate into the role runner which had none (Finding 2), counts in-flight across all managed roots (Finding 5), and checks supervision up front so an unsupervised host refuses *before* pausing dispatch (Finding 4 / AC5).

## Changes

- **`types.rs`** — new `Request::DrainAndRestartDaemon { timeout_secs, force_after_timeout }` and `Request::AbortDrain`; new `Response::DaemonDrain { accepted, supervisor, in_flight, message }`; `DaemonStatusReport` gains `draining` / `drain_deadline` / `drain_note`, all `#[serde(default)]` per the `capacity_bound` precedent. **`RestartDaemon` is untouched.**
- **`ipc.rs`** — `DrainState` (drain flag + generation + descriptor), the pure `evaluate_drain_tick` decision fn, `count_in_flight_sweeps` (cross-root, mirrors `build_daemon_status`), the drain-supervisor task that owns the eventual `std::process::exit(EXIT_RESTART)`, and `build_daemon_status_with_drain`. Drain requests are intercepted in `handle_client` (like `RestartDaemon`) since they must ack immediately and exit minutes later.
- **`work_finder.rs` / `epic_supervisor.rs` / `role_runner.rs`** — the drain flag is OR'd into the existing halt checks (work finder, epic supervisor) and newly gates role-tick starts (role runner had no halt gate — Finding 2's real gap).
- **`main.rs`** — constructs the shared `DrainState`, threads its flag into all three producers and the state into the IPC server; `restart` CLI gains `--drain` / `--timeout` / `--force-after-timeout` / `--abort-drain`; `loom-daemon status` renders `DRAINING (n remaining, deadline …)`.
- **`event_bus.rs` + `daemon-reference.md`** — `daemon.drain.{started,completed,aborted,timeout}` added to **both** taxonomy tables, authorized by this issue; the "sweeps survive, they are not drained" paragraph and the "Supervised restart primitive" section are amended with the drain extension.

- **AC1** — drain stops new dispatch immediately: drain flag OR'd into work-finder/epic halts; `test_tick_skips_dispatch_while_draining` (epic) + `test_drain_stops_role_ticks_from_starting` (role runner, zero invokes while draining).
- **AC2** — restart only after the last sweep completes: `evaluate_drain_tick` completes only at `in_flight == 0` (`test_drain_tick_completes_only_at_zero`), terminal entries excluded; cross-root counting proven by `test_count_in_flight_sweeps_cross_root` (live sweep in the secondary repo blocks completion).
- **AC3** — timeout without force refuses + resumes + reports; with force cancels via the existing `SweepRegistry::cancel` path (`evaluate_drain_tick` → `TimedOutRefuse`/`TimedOutForce`; `resolve_timeout` clears the flag + records the note).
- **AC4** — `loom-daemon status` surfaces DRAINING with remaining count + deadline (`build_daemon_status_with_drain`, `test_build_daemon_status_with_drain_overlays_state`, round-trip + pre-drain-JSON compat).
- **AC5** — unsupervised refusal is immediate and never pauses dispatch (`test_drain_request_unsupervised_refuses_without_pausing` asserts `accepted:false` **and** flag never set, generation not bumped).
- **AC6** — abort restores dispatch and bumps the generation so a stale supervisor never fires a later restart (`test_abort_supersedes_running_supervisor_generation`).
- **Wire-compat** — `{"type":"RestartDaemon"}` still serializes unchanged (`test_drain_request_wire_compat`, and the pre-existing `test_restart_daemon_request_response_round_trip` passes unmodified).
- **Edge cases** — zero in-flight exits on the first poll (no interval wait); a second drain while DRAINING is an idempotent ack (`DrainBegin::AlreadyDraining`, deadline unchanged); SIGTERM precedence is preserved (drain supervisor only ever exits `EXIT_RESTART`; the signal handler exits `143` independently); `roleRunner` disabled path unchanged.

## Test Plan

- `cargo test --lib`: **1258 passed, 0 failed** (includes all new drain tests).
- `cargo fmt` clean, `cargo clippy --lib --bins --tests` clean.
- Pre-existing `tests/integration_basic.rs` failures are **unrelated**: they spawn the real daemon and hard-code a 5s socket-bind timeout, but the daemon takes ~12s to bind under this host's build contention. Verified failing identically on the clean `origin/main` baseline (stash → rebuild → `test_ping_pong` FAILED) with none of this PR's changes.

## Scope guard

Target was **≤7 files, ≤450 LOC including tests**. Landed **8 files** — exactly the Affected-Files list (7 Rust + `daemon-reference.md`). The total added line count exceeds 450, driven by (a) this crate's mandatory heavy rustdoc convention (~298 added comment lines) and (b) a test suite covering every AC + wire-compat + no-regression + the enumerated edge cases per the curated Test Plan. **All three architectural stop-and-split boundaries were respected**: no role-tick registry was built, `claim_reconciliation.rs` was not touched, and `SweepRegistry::cancel`'s signature is unchanged (the force path only *calls* it). The functional core (drain state machine + decision fn + cross-root counter + handlers + flag threading) is modest; the size is documentation and tests, not misplaced architecture.

Closes #4090

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -113,6 +113,10 @@ issue** — the v0.10.0 set is intentionally frozen.
 | `epic.issue.{N}.join`      | Epic supervisor (#3842)        | `{epic, action, state}` |
 | `epic.issue.{N}.close`     | Epic supervisor (#3842)        | `{epic, action, state}` |
 | `daemon.capacity.advisory` | Work finder (#3902)            | `{pressured, queued, healthy_accounts, exhausted_accounts, total_accounts, estimated_drain_minutes?, message}` |
+| `daemon.drain.started`     | Drain supervisor (#4090)       | `{in_flight, timeout_secs, force_after_timeout, deadline}` |
+| `daemon.drain.completed`   | Drain supervisor (#4090)       | `{in_flight}` (always `0`) |
+| `daemon.drain.aborted`     | Daemon IPC (#4090)             | `{was_draining}` |
+| `daemon.drain.timeout`     | Drain supervisor (#4090)       | `{in_flight, forced, cancelled?}` |
 
 The four `epic.issue.{N}.*` topics were authorized by **#3873** (epic #3842
 Phase 4) and are documented in full under [Epic supervisor](#epic-supervisor-3842)
@@ -121,6 +125,12 @@ below. The `daemon.capacity.advisory` topic was authorized by **#3902** (epic
 state change** (entered/left the token-bound state), never every tick, so the
 operator gets one add-capacity advisory on the way in and one recovery on the way
 out. See [Token-capacity backpressure](#token-capacity-backpressure-3902) below.
+The four `daemon.drain.*` topics were authorized by **#4090** for the scheduled
+drain-and-restart primitive — `started` when a drain is accepted, `completed`
+when the last in-flight sweep finishes (right before the supervised relaunch),
+`aborted` when an operator cancels a drain, and `timeout` when the deadline is
+reached (`forced` distinguishes a refusal from a force-cancel restart). See
+[Supervised restart primitive](#supervised-restart-primitive-4054) below.
 They ride the same in-memory bus as the sweep topics and are tailable via
 `subscribe_to_events` / `tail_event_bus`.
 
@@ -1948,6 +1958,16 @@ the registry reconciles their state on the next start (`SweepRegistry::reconstru
 re-admits live-lock owners). To actively cancel a sweep, use
 `mcp__loom__cancel_sweep` against a running daemon *before* stopping it.
 
+> **Amended by #4090 (scheduled drain-and-restart).** The above describes the
+> *plain* stop/restart: sweeps survive the process boundary but the relaunched
+> daemon's in-memory registry starts empty, so a plain `RestartDaemon` leaves the
+> surviving children as **orphans** — invisible to `list_sweeps`, the concurrency
+> cap, the watchdog, and the reaper, and still holding `loom:building` for a
+> `claim_reconciliation` pass to re-dispatch as duplicates. When you want "finish
+> what you started, then roll" instead, use `loom-daemon restart --drain` (below):
+> it stops admitting new work and waits for the registry to empty before exiting,
+> so no sweep is killed and no orphan is left behind.
+
 **Exit code carries shutdown intent (#4054).** The daemon encodes *why* it is
 shutting down in its exit code, because the launchd `KeepAlive:{SuccessfulExit:true}`
 plist (below) relaunches only on a clean exit `0`: the `RestartDaemon` primitive
@@ -2267,6 +2287,47 @@ loom-daemon restart          # send RestartDaemon over the IPC socket
   the `exec` (Option 2) and detached-helper (Option 3) alternatives were rejected,
   and the Curator's exit-code-race finding — is in
   `docs/design/supervised-restart-primitive.md`.
+
+#### Scheduled drain-and-restart (`--drain`, #4090)
+
+A plain `loom-daemon restart` exits immediately: in-flight sweeps survive the
+process boundary but become **orphans** (absent from the relaunched daemon's
+in-memory registry — see the "sweeps survive, they are not drained" amendment
+above). `--drain` closes that gap by finishing in-flight work *before* rolling:
+
+```bash
+loom-daemon restart --drain                       # finish in-flight sweeps, then restart
+loom-daemon restart --drain --timeout 600         # bound the wait (default 1800s)
+loom-daemon restart --drain --force-after-timeout # at the deadline, cancel stragglers and restart anyway
+loom-daemon restart --abort-drain                 # cancel an in-progress drain, resume dispatch (no restart)
+```
+
+- **What a drain does:** it sets a daemon-global drain flag that is OR'd into the
+  same halt checks the main-health gate uses (#3812) — the work finder stops
+  dispatching, the epic supervisor stops scheduling phases, and (newly, #4090) the
+  role runner stops **starting** new role ticks. Already-running work is left
+  alone. A supervisor task then polls the cross-root in-flight sweep count (every
+  managed workspace, not just the primary) and exits `EXIT_RESTART` for the launchd
+  relaunch only once it reaches **zero** — so `list_sweeps` after the relaunch is
+  consistent with reality and there are **no orphans**.
+- **Fail-safe timeout:** reaching `--timeout` without `--force-after-timeout`
+  **refuses** the restart (clears the drain flag, resumes dispatch, stays up) and
+  reports the reason via `loom-daemon status` — it never silently restarts or
+  silently gives up. `--force-after-timeout` opts into cancelling the stragglers
+  via the existing `cancel_sweep` path, then restarts.
+- **Supervision proof is checked up front (AC5):** on an unsupervised host the
+  request is refused **before** dispatch is paused (`accepted: false`), so a caller
+  can detect nothing happened and no silent outage is introduced.
+- **Residual (documented, bounded):** role ticks have no sweep-registry entry to
+  poll, so a drain stops them *starting* but cannot *await* one already running —
+  a drain can complete while a role tick is still mid-flight, bounded by the role
+  timeout (`DEFAULT_ROLE_TIMEOUT`, 1800s). Awaiting role ticks is deliberately out
+  of scope (it would require a role registry, #4090's stop-and-split boundary).
+- **Observability:** `loom-daemon status` renders `DRAINING (n sweep(s) remaining,
+  deadline …)` while active and the last transition (timeout refusal / abort)
+  afterward; the four `daemon.drain.*` events (above) narrate the transitions on
+  the event bus. **Cannot be used for its own first roll** — see the rollout note
+  below.
 
 ### Self-update (rebuild + provision + restart, #3968)
 
--- a/loom-daemon/src/epic_supervisor.rs
+++ b/loom-daemon/src/epic_supervisor.rs
@@ -473,6 +473,11 @@ pub struct EpicSupervisor<S: EpicSource, D: EpicDispatcher> {
     /// **nothing** — a red `main` stops epic-child dispatch too (#3885). Absent
     /// (the default / test path) means the supervisor is never gated.
     health_state: Option<Arc<MainHealthState>>,
+    /// Optional daemon-global drain flag (#4090) shared with the work-finder and
+    /// role-runner. When set, [`tick`](Self::tick) dispatches **nothing** — a
+    /// scheduled drain pauses new epic-child dispatch just like a red `main`.
+    /// Absent (the default / test path) means the supervisor is never drained.
+    drain_flag: Option<Arc<AtomicBool>>,
 }
 
 impl<S: EpicSource, D: EpicDispatcher> EpicSupervisor<S, D> {
@@ -489,6 +494,7 @@ impl<S: EpicSource, D: EpicDispatcher> EpicSupervisor<S, D> {
             inflight_ttl: resolve_inflight_ttl(),
             bus: None,
             health_state: None,
+            drain_flag: None,
         }
     }
 
@@ -505,6 +511,15 @@ impl<S: EpicSource, D: EpicDispatcher> EpicSupervisor<S, D> {
     #[must_use]
     pub fn with_health_gate(mut self, health_state: Arc<MainHealthState>) -> Self {
         self.health_state = Some(health_state);
+        self
+    }
+
+    /// Attach the daemon-global drain flag (#4090) so that while a scheduled
+    /// drain is in progress the supervisor dispatches nothing. Builder-style;
+    /// returns `self`.
+    #[must_use]
+    pub fn with_drain_flag(mut self, drain_flag: Arc<AtomicBool>) -> Self {
+        self.drain_flag = Some(drain_flag);
         self
     }
 
@@ -529,9 +544,21 @@ impl<S: EpicSource, D: EpicDispatcher> EpicSupervisor<S, D> {
         // Existing in-flight sweeps are untouched (halting only stops making a
         // red `main` worse); a green run clears the flag and the next tick
         // resumes normally.
-        if self.health_state.as_ref().is_some_and(|s| s.is_halted()) {
+        let health_halted = self.health_state.as_ref().is_some_and(|s| s.is_halted());
+        // Scheduled drain (#4090): a daemon-global drain pauses new epic-child
+        // dispatch exactly like a red `main`, so it is OR'd into the same gate.
+        let drained = self
+            .drain_flag
+            .as_ref()
+            .is_some_and(|f| f.load(Ordering::Relaxed));
+        if health_halted || drained {
             log::warn!(
-                "epic_supervisor: main-health gate halted — skipping tick (no epic dispatch while main is red)"
+                "epic_supervisor: {} — skipping tick (no epic dispatch)",
+                if drained {
+                    "drain in progress"
+                } else {
+                    "main-health gate halted (main is red)"
+                }
             );
             return Ok(TickReport {
                 halted: true,
@@ -1010,6 +1037,7 @@ pub fn spawn_multi_supervisor_thread(
     event_bus: Arc<EventBus>,
     health_states: Arc<WorkspaceHealthStates>,
     interval: Duration,
+    drain_flag: Arc<AtomicBool>,
 ) -> Result<SupervisorHandle> {
     let shutdown = Arc::new(AtomicBool::new(false));
     let shutdown_thread = shutdown.clone();
@@ -1078,7 +1106,8 @@ pub fn spawn_multi_supervisor_thread(
                             let supervisor =
                                 EpicSupervisor::new(source, dispatcher, IssueCreationMutex::new())
                                     .with_event_bus(event_bus.clone())
-                                    .with_health_gate(health_states.get_or_create(root));
+                                    .with_health_gate(health_states.get_or_create(root))
+                                    .with_drain_flag(drain_flag.clone());
                             roots.push(root.clone());
                             supervisors.push(supervisor);
                             log::info!("epic_supervisor: watching workspace {}", root.display());
@@ -1672,6 +1701,27 @@ mod tests {
         assert!(!r2.halted);
         assert_eq!(r2.roles_dispatched, 1);
         assert_eq!(sup.dispatcher.roles.len(), 1);
+    }
+
+    /// A scheduled drain (#4090) halts epic dispatch exactly like a red `main`
+    /// (AC1): with the drain flag set, the tick dispatches nothing; clearing it
+    /// resumes normal dispatch. Mirrors the main-health halt tests above.
+    #[tokio::test]
+    async fn test_tick_skips_dispatch_while_draining() {
+        let drain = Arc::new(AtomicBool::new(true));
+        let mut sup = supervisor(vec![EpicSnapshot::new(1, "flat body", vec![], vec![])])
+            .with_drain_flag(drain.clone());
+
+        let report = sup.tick().await.unwrap();
+        assert!(report.halted, "a drain halts the tick");
+        assert_eq!(report.roles_dispatched, 0);
+        assert!(sup.dispatcher.roles.is_empty(), "no epic dispatch while draining");
+
+        // Clear the drain ⇒ the next tick dispatches normally.
+        drain.store(false, Ordering::Relaxed);
+        let r2 = sup.tick().await.unwrap();
+        assert!(!r2.halted);
+        assert_eq!(r2.roles_dispatched, 1);
     }
 
     #[tokio::test]
--- a/loom-daemon/src/event_bus.rs
+++ b/loom-daemon/src/event_bus.rs
@@ -32,6 +32,10 @@
 //! | `epic.issue.{N}.join`      | Epic supervisor | `{epic, action, state}` |
 //! | `epic.issue.{N}.close`     | Epic supervisor | `{epic, action, state}` |
 //! | `daemon.capacity.advisory` | Work finder | `{pressured, queued, healthy_accounts, exhausted_accounts, total_accounts, estimated_drain_minutes?, message}` |
+//! | `daemon.drain.started`   | Drain supervisor | `{in_flight, timeout_secs, force_after_timeout, deadline}` |
+//! | `daemon.drain.completed` | Drain supervisor | `{in_flight}` (always 0) |
+//! | `daemon.drain.aborted`   | Daemon (IPC) | `{was_draining}` |
+//! | `daemon.drain.timeout`   | Drain supervisor | `{in_flight, forced, cancelled?}` |
 //!
 //! New topics require a follow-up issue — the taxonomy is intentionally
 //! pinned. The four `epic.issue.{N}.*` topics were authorized by **#3873**
@@ -39,6 +43,8 @@
 //! (decompose / expand / join / close). The `daemon.capacity.advisory` topic
 //! was authorized by **#3902** (epic #3809) for the work finder's
 //! token-capacity backpressure advisory (fired on pressure state change). The
+//! four `daemon.drain.*` topics were authorized by **#4090** for the scheduled
+//! drain-and-restart primitive (started / completed / aborted / timeout). The
 //! bus accepts arbitrary topic
 //! strings (`publish` does not reject unknown topics) so the publisher side
 //! stays open for future extension, but the documented taxonomy is the
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -14,6 +14,7 @@ use anyhow::Result;
 use chrono::Utc;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tokio::fs;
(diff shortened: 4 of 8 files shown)


rjwalters:
❌ **Changes Requested — Merge Conflict**

## Code quality review

I read the curated issue (#4090 — 8 curator findings, in/out-of-scope split, AC1–AC6, and the detailed test plan) and reviewed the diff against it. The implementation is high quality and closely follows the curated recommendations:

- **Finding 1** (reuse main-health-gate halt plumbing, don't reuse `MainHealthState` itself): done — a separate `Arc<AtomicBool>` drain flag, OR'd into `work_finder.rs:~1220`, `epic_supervisor.rs:~547`, and newly into `role_runner.rs` (single- and multi-workspace loops).
- **Finding 2** (role ticks have no halt gate): fixed — `spawn_role_task` / `spawn_multi_role_task` both check the drain flag before invoking, with `test_drain_stops_role_ticks_from_starting` proving zero `invoke` calls while draining (matches the test plan's "highest-value new coverage" callout).
- **Finding 3** (don't touch `RestartDaemon`'s wire shape): verified — `RestartDaemon` unit variant is byte-for-byte unchanged; `Request::DrainAndRestartDaemon`/`Request::AbortDrain` are separate variants, and `test_drain_request_wire_compat` + the pre-existing `{"type":"RestartDaemon"}` round-trip assertion both pass.
- **Finding 4** (exit ownership must move out of the inline connection handler; supervision proof up front): the drain-supervisor task owns `std::process::exit(EXIT_RESTART)`; `handle_drain_request` calls `detect_supervisor()` before `drain.begin(...)`, and `test_drain_request_unsupervised_refuses_without_pausing` asserts both `accepted:false` **and** the flag/generation are untouched — correctly serialized (`#[serial_test::serial]`) against the other `LOOM_DAEMON_SUPERVISOR` test.
- **Finding 5** (cross-root in-flight counting): `count_in_flight_sweeps` mirrors `build_daemon_status`'s per-root non-terminal filter; `test_count_in_flight_sweeps_cross_root` proves a live sweep in a *secondary* managed root blocks completion.
- **Finding 6** (event taxonomy is pinned): `daemon.drain.{started,completed,aborted,timeout}` added to **both** tables (`event_bus.rs` module doc and `daemon-reference.md`), each citing #4090.
- **Finding 7** (`#[serde(default)]` wire compat): `draining` / `drain_deadline` / `drain_note` all carry it, with an explicit backward-compat test (`test_daemon_status_backward_compat_missing_drain_fields`).
- **AC6** (abort supersedes a running supervisor): `test_abort_supersedes_running_supervisor_generation` proves the generation-based supersession contract the async loop relies on (can't unit-test the `process::exit` branch directly — reasonable trade-off, well commented).
- Scope guard: touches exactly the 8 files the curator's Affected-Files list named; no role-tick registry, no `claim_reconciliation.rs` touch, `SweepRegistry::cancel`'s signature unchanged — the disclosed LOC overage is genuinely doc/test weight, not misplaced architecture.

**Local verification** (on the PR's own base, `git merge-base` = pre-conflict main):
- `cargo build --lib` — clean.
- `cargo test --lib` — **1258 passed, 0 failed** (matches the PR's claimed count), including all 9 new drain-specific tests.
- `cargo clippy --lib --bins --tests` — clean.

Minor, non-blocking observations (not requesting changes for these — noting for awareness):
- `count_in_flight_sweeps` (ipc.rs) duplicates `build_daemon_status`'s per-root non-terminal-filter loop rather than literally calling into it, so the two must be kept in sync by hand if the "in-flight" definition ever changes. Semantically identical today and both are tested; a future refactor to share the loop would be a nice-to-have, not required here.
- The test plan's SIGTERM-during-drain and roleRunner-disabled edge cases don't have new dedicated tests, but neither code path changed in a way that plausibly regresses them (the SIGTERM handler and the `roleRunner` gate at `main.rs` are untouched) — acceptable residual risk.

## Blocking issue: real merge conflict, not just a stale base

`gh pr view` reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`, and I confirmed this is a genuine content conflict (not just "behind main") by fetching `origin/main` and attempting the rebase in the PR's worktree:

```
Auto-merging .loom/docs/daemon-reference.md   (clean)
CONFLICT (content): Merge conflict in loom-daemon/src/ipc.rs
CONFLICT (content): Merge conflict in loom-daemon/src/main.rs
CONFLICT (content): Merge conflict in loom-daemon/src/types.rs
```

The cause: a concurrently-merged PR for the credential-preflight feature (#4005) landed on `main` after this branch was created, and it touches the **same insertion points** this PR does:

- **`types.rs`** — both PRs add a new field to `DaemonStatusReport` at the same spot (this PR's `draining`/`drain_deadline`/`drain_note`, #4005's `credential_preflight`). Non-overlapping semantically, but the textual conflict needs a human/Doctor to merge both field blocks and keep both `#[serde(default)]` doc conventions intact.
- **`ipc.rs`** — both PRs add an import off `crate::types::{...}` and both add a new module declaration; needs both `CredentialPreflightReport` and this PR's drain types.
- **`main.rs`** — both PRs extend the `IpcServer::new(...)` call with a new trailing argument (`credential_preflight` vs `drain_state.clone()`) and both add a JSON key to the `status --json` renderer at the same location. Resolving this requires checking the *merged* `IpcServer::new` signature in `ipc.rs` to get argument order right, and keeping both the `"credential_preflight"` and `"drain"` JSON keys in the renderer — this is not a mechanical "keep both sides" resolution, it needs someone to actually merge two constructor call sites and a JSON builder correctly.

Per the Judge role's DIRTY workflow, this is a **complex conflict** (multiple files, interdependent, overlapping insertion points in core Rust structs/constructors) rather than a simple non-overlapping addition, so I'm routing it to Doctor rather than resolving it myself under the review time budget. I aborted my rebase attempt cleanly (`git rebase --abort`) — the PR branch is untouched.

**Next step:** rebase `feature/issue-4090` onto current `main`, resolve the three conflicts above (keep both features' fields/args/JSON keys), verify `cargo build && cargo test --lib` still passes, then re-request review.

rjwalters:
🔧 Merge conflict resolved — rebased `feature/issue-4090` onto current `origin/main`.

The conflict was caused by #4005 (credential preflight) landing concurrently and touching the same insertion points. Resolved so **both** features' additions survive:

**`loom-daemon/src/types.rs`** — kept both field blocks on `DaemonStatusReport`: #4005's `credential_preflight` and this PR's `draining` / `drain_deadline` / `drain_note`, each with its `#[serde(default)]`.

**`loom-daemon/src/ipc.rs`** — threaded both `credential_preflight` and `drain_state` through `IpcServer` (struct field, `new()` params, `run()` spawn, `handle_client` signature). `DrainState` is defined locally in this file, so no new import was needed; `CredentialPreflightReport` was already imported. The base `build_daemon_status` keeps #4005's `credential_preflight` param and the drain-default fields; `build_daemon_status_with_drain` now takes `credential_preflight` and forwards it to the base builder. Both drain unit tests updated to pass `&test_credential_preflight()`.

**`loom-daemon/src/main.rs`** — matched the drain-status argument against the now-merged `IpcServer::new` signature (`…, fallback_root, credential_preflight, drain_state`), and kept **both** the `"credential_preflight"` and `"drain"` keys in the `status --json` renderer plus both human-readable status blocks.

Verification on the rebased tree:
- `cargo build --lib` — clean.
- `cargo test --lib` — **1271 passed, 0 failed** (includes #4005's tests and all drain-specific tests).
- `cargo clippy --lib --bins --tests -- -D warnings` — clean.

`gh pr view` now reports `mergeable: MERGEABLE` (was `CONFLICTING`). No substantive logic changes — pure conflict resolution merging two non-overlapping features. Back to `loom:review-requested`.

rjwalters:
✅ **Approved!** (fresh full review post-rebase)

Re-reviewed the current state of the branch after Doctor's rebase onto `origin/main` (branch is now 1 ahead / 0 behind; `mergeStateStatus: CLEAN`). This supersedes the earlier merge-conflict change request.

## Verification

- **CI:** all required checks green — `Rust Linting & Build Check` ✅, `Rust Unit Tests` ✅ (3m50s), `CLAUDE.md Line Budget` ✅, `Detect Changes` ✅. Installer/MCP jobs correctly path-skipped.
- **Local build gate** (`.loom/scripts/build-gate.sh`): `loom-daemon` lib **1271 passed / 0 failed**, bins **21 passed / 0 failed**, `loom-api` **4 passed** — including every new drain test:
  `test_evaluate_drain_tick_decisions`, `test_drain_tick_completes_only_at_zero`, `test_count_in_flight_sweeps_cross_root`, `test_drain_request_unsupervised_refuses_without_pausing`, `test_abort_supersedes_running_supervisor_generation`, `test_drain_request_wire_compat`, `test_daemon_status_backward_compat_missing_drain_fields`, `test_build_daemon_status_with_drain_overlays_state`, `test_drain_stops_role_ticks_from_starting`, `test_tick_skips_dispatch_while_draining`.
- **No-regression:** `test_restart_daemon_request_response_round_trip` and the main-health halt tests pass unmodified.

## Code quality — findings honored

The implementation follows the curated scope precisely:

- **Finding 1** — separate `Arc<AtomicBool>` drain flag OR'd into the existing halt gates rather than reusing `MainHealthState` (which the gate loop would clear). ✅
- **Finding 2** (the real correctness gap) — `role_runner` now gates both `spawn_role_task` and `spawn_multi_role_task`; a drain stops a fresh Champion/Curator tick from *starting*. The new `test_drain_stops_role_ticks_from_starting` asserts zero invokes while draining, then resumption. ✅
- **Finding 3** — `DrainAndRestartDaemon` is a **new** variant; `RestartDaemon` is byte-for-byte untouched and the `{"type":"RestartDaemon"}` round-trip assertion is unchanged. ✅
- **Finding 4** — the inline `std::process::exit` is gone from the drain path; a background `run_drain_supervisor` task owns the eventual exit, and `detect_supervisor()` is checked **up front** so AC5's refusal never pauses dispatch. ✅
- **Finding 5** — `count_in_flight_sweeps` mirrors `build_daemon_status`'s cross-root accounting (`effective_roots` + `!is_terminal()`); proven by the cross-root test. ✅
- **Finding 6** — four `daemon.drain.*` topics added to both the `event_bus.rs` module table and `daemon-reference.md`, citing #4090. ✅
- **Finding 7** — new `DaemonStatusReport` fields (`draining`, `drain_deadline`, `drain_note`) all carry `#[serde(default)]`; pre-drain-JSON compat test passes. ✅
- **Finding 8** — docs updated. ✅

## AC coverage
AC1 (dispatch pause across work-finder / epic / role-runner), AC2 (`evaluate_drain_tick` completes only at zero, cross-root), AC3 (fail-safe `TimedOutRefuse` clears the flag + stays up; `TimedOutForce` cancels via the existing `SweepRegistry::cancel`), AC4 (`DRAINING (n remaining, deadline …)` rendering), AC5 (unsupervised refusal, flag never set), AC6 (abort bumps the generation so a stale supervisor never fires) — all backed by tests.

## Design notes
- The `DrainState::begin` sets the flag while holding the descriptor lock, so status can never observe `flag=true` with `active=false` — a nice touch that closes the obvious status race.
- Generation-based supersede/abort correctly handles the subtle "abort, then the queue empties anyway" case (AC6): a stale supervisor returns without exiting.
- Scope discipline held — 8 files == the Affected-Files list, and the three stop-and-split boundaries were respected (no role-tick registry, `claim_reconciliation.rs` untouched, `SweepRegistry::cancel` only *called*). The LOC overage is rustdoc + AC-per-test coverage, not misplaced architecture.

Clean, well-documented, thoroughly tested. Nice work.

PR #4138, opened 2026-07-28
fix(daemon): suppress dispatch into a root while its build-gate run is in flight

### LINKED ISSUE #4084: fix(build-gate): nice -n 5 (#4073) did not prevent a 1200s gate timeout under 2 concurrent sweeps
## Summary

#4073 dropped the build gate's `nice -n 19` handicap to a mild `nice -n 5`, on the theory that extreme deprioritization was why the gate kept losing its CPU race against concurrent sweep builds. **The first contended gate run after that fix still timed out at the full 1200s budget.** The mild-niceness change was necessary but is not sufficient.

## Evidence (this host, 2026-07-27)

Sequence, all verified:

| Time (PDT) | Event |
|---|---|
| 12:10 | `118e67fd` merges #4073 — gate niceness 19 → 5 |
| 13:31:10 | daemon restarted (pid 96910); `main_health_gate: starting multi-workspace loop` |
| 13:32:12 | sweep dispatched — issue #6175 (vibesql) |
| 13:44:14 | sweep dispatched — issue #4058 (loom) |
| **13:51:43** | **`main_health_gate: /Users/rwalters/GitHub/loom gate run UNEVALUATED [timeout] — gate command 'bash .loom/scripts/build-gate.sh' timed out after 1200s and was killed; main was NOT evaluated`** |

Confirmations:

- `.loom/scripts/build-gate.sh` is **in sync** with `defaults/scripts/build-gate.sh` (no #3770-class install drift), and `_gate_niceness="${LOOM_BUILD_GATE_NICENESS:-5}"` at `:76` — so this run really did execute at nice 5, not nice 19.
- The gate script is shell, read at invocation, so the stale daemon binary on this host (`a68099d5` vs source `c4b6f677`) is **not** a factor — the fix was live.
- Two sweep children were running across the entire gate window.

## The contrast that makes this diagnostic

The **same repo at a nearby commit compiled green in ~726s** on an idle host, well under the 1200s budget. Under two concurrent sweeps it exceeded 1200s — a >65% inflation attributable to contention alone. This is the clean A/B the earlier investigation lacked, and it retires the remaining "cold-compile cost is just larger than measured" theory: cold-compile cost is ~726s; contention is what breaks the budget.

## Why raising the budget again is the wrong lever

The gate's cost scales with however many sweeps happen to be in flight, which is itself dynamic (`Dynamic concurrency cap: 8` on this host). Any fixed budget large enough for the worst contention case makes a genuinely-hung gate invisible for that long. The budget is not the variable that needs tuning.

## Candidate directions (from #4020's rescope, none yet implemented)

1. **Serialize gate against sweeps** — have the gate take the same advisory lock sweep children take, or have the work finder refrain from dispatching while a gate run is in flight. Cleanest semantically; costs dispatch throughput for ~12 min out of every interval.
2. **Shared `CARGO_TARGET_DIR`** — the gate and sweeps currently recompile the same crates into separate target dirs, so contention is partly self-inflicted duplicate work rather than pure CPU scarcity. Needs care: concurrent cargo on one target dir blocks on the package lock, which converts CPU contention into serialization (possibly the desired outcome, but it should be a deliberate choice).
3. **Budget as a function of in-flight sweeps** — e.g. `base + k × in_flight`. Keeps a hung gate detectable while absorbing known contention.
4. **Cap gate parallelism** (`cargo build -j N` with N below core count) so the gate is a predictable, small slice rather than competing for all 28 cores.

## Acceptance criteria

1. A gate run concurrent with ≥2 sweeps on a 28-core host reaches a determinate verdict (green or red) rather than `UNEVALUATED [timeout]`.
2. The chosen mechanism does not make a genuinely hung gate undetectable — a gate that stops making progress is still killed and reported.
3. Whatever knob is introduced follows the env > config > default precedence contract and is documented in `.loom/docs/build-gate.md`.

## Notes

- Follow-up to **#4020** (closed) — that issue was rescoped to exactly this problem and closed on #4073's mild-niceness change; this issue records that the change did not resolve it.
- Blocked-ish on the companion observability issue: without an INFO-level green line carrying elapsed time, there is no way to confirm a fix worked other than waiting for the absence of a timeout.

---

All file/line references below verified against `origin/main` @ `cf94d174`.

### Finding 1 — the timeline inverts the option ranking: both sweeps started *after* the gate run began

Reconstructing the incident against the code:

- `DEFAULT_MAIN_HEALTH_GATE_INTERVAL_SECS = 30` (`loom-daemon/src/main_health_gate.rs:110`), and the loop
  deliberately skips its immediate first tick (`main_health_gate.rs:2372-2373`, "First tick fires
  immediately; skip it so we don't churn at boot").
- Daemon start 13:31:10 → first real tick ≈ 13:31:40 → gate run start ≈ **13:31:43**.
- Timeout fired 13:51:43, exactly 1200s later. The arithmetic closes.
- Sweeps dispatched **13:32:12** and **13:44:14** — both *inside* the gate's own run window.

**Neither sweep was already running when the gate started.** A dispatch-side interlock ("do not dispatch
new sweeps into a root whose gate run is in flight") would have prevented this exact incident outright —
no draining, no waiting on long-lived sweeps, no throughput pause beyond the gate's own ~12 min.

This is materially stronger than the issue's framing of candidate 1, which assumed the gate would have to
*wait for in-flight sweeps to drain*. It does not: the cheap half (suppress new dispatch) is the half the
evidence indicts.

### Finding 2 — the interlock plumbing already exists; candidate 1 is far cheaper than it looks

The issue implies candidate 1 needs new wiring between the gate and the work finder. It does not:

- `Arc<WorkspaceHealthStates>` is constructed once at `loom-daemon/src/main.rs:747` and handed to **both**
  the gate loop (`main.rs:888`, writer) and the work finder (reader).
- The work finder **already consults it per root**: `loom-daemon/src/work_finder.rs:1186`
  (`roots.iter().map(|r| health_states.is_halted(r))`), and `work_finder.rs:994` in the single-root path.

So the minimal shape of candidate 1 is:

1. Add a `gate_in_flight: AtomicBool` (or `Mutex<bool>`) to `MainHealthState` (`main_health_gate.rs:157`),
   with `set_gate_in_flight` / `gate_in_flight` accessors mirroring the existing `set_halted` / `is_halted`
   pair (`main_health_gate.rs:243-253`) and a `WorkspaceHealthStates` passthrough mirroring
   `main_health_gate.rs:475-486`.
2. Set it true immediately before the `spawn_blocking(run_gate_tick)` call at `main_health_gate.rs:2418`
   and clear it in **every** exit arm of the `match joined` — including `Err(e)` (the panic arm at
   `main_health_gate.rs:2436`), or the flag latches and permanently starves dispatch. **Prefer an RAII
   guard over manual clears** so a future early-return cannot reintroduce the latch.
3. Have the work finder treat `gate_in_flight` as a *dispatch suppressor* alongside `is_halted` at
   `work_finder.rs:1186` / `work_finder.rs:994`.

No new config schema, no new inter-task channel, no change to the sweep registry.

### Finding 3 — candidate 4 (cap gate `-j`) moves the wrong metric and should be dropped

Capping `cargo build -j N` reduces the gate's CPU *footprint* but **increases its wall-clock**, and the
failure being fixed is a wall-clock timeout. Fewer runnable rustc threads means a smaller scheduler share,
which makes AC1 strictly harder to satisfy. Candidate 4 is a good-citizen lever for the *sweeps'* benefit,
not a fix for this issue. Do not implement it as the primary mechanism.

### Finding 4 — candidate 2 (shared `CARGO_TARGET_DIR`) is real but risks making the timeout worse

Verified: **`CARGO_TARGET_DIR` is set nowhere in the repo** (`git grep CARGO_TARGET_DIR` over tracked files
returns nothing). The gate compiles into the main workspace `target/` while each sweep compiles into its own
worktree `target/` — so the duplicate-work theory in the issue body is confirmed, not merely suspected.

But sharing one target dir converts CPU contention into **cargo package-lock blocking**, and time blocked on
that lock counts against the same 1200s wall clock enforced at `main_health_gate.rs:1920-1940`. Absent a
budget change, candidate 2 can plausibly *increase* the timeout rate. It also changes sweep build behavior
repo-wide (feature-flag thrash between differing sweep configurations). Treat as a separate follow-up, not
part of this fix.

### Finding 5 — where the 1200s budget is actually enforced (affects rollout latency)

The timeout is enforced **daemon-side** in Rust, not in the shell script: the `try_wait` poll loop at
`loom-daemon/src/main_health_gate.rs:1920-1940` kills the child on `start.elapsed() >= timeout`, where
`timeout` comes from `buildGate.timeoutSeconds` (`.loom/config.json` sets `1200`). `GATE_POLL_INTERVAL` is
200 ms (`main_health_gate.rs:117`).

Consequence: **any fix here requires a daemon rebuild and restart to take effect** — unlike #4073, which was
a pure shell change that went live immediately. Sequence the roll accordingly and do not conclude "the fix
did not work" from a run against a stale daemon binary.

### Recommended approach

**Implement candidate 1, dispatch-suppression half only** (Findings 1 + 2). It is the smallest change that
the evidence shows would have prevented the observed failure, it reuses plumbing that already exists, and it
keeps the timeout budget fixed so AC2 (a genuinely hung gate is still killed and reported) holds unchanged —
the kill path at `main_health_gate.rs:1920-1940` is untouched.

Explicitly **out of scope** for this issue: raising `timeoutSeconds` (the issue body already argues why),
candidate 2, and candidate 4.

Per AC3, the suppressor needs an env > config > default escape hatch — follow the existing pattern in
`read_autonomous_gate_config` / `resolve_enabled` (`main_health_gate.rs:2154-2216`) and
`resolve_utilization_target` (`loom-daemon/src/cpu_headroom.rs:173`). Suggested:
`autonomous.mainHealthGate.suppressDispatchDuringGate` (default **true**) with a
`LOOM_MAIN_HEALTH_GATE_SUPPRESS_DISPATCH` env override, documented in `.loom/docs/build-gate.md`.

### Scope guard

Target ≤ 4 files and ≤ 300 LOC (including tests). If the implementation starts touching the sweep registry,
the epic supervisor, or `spawn-claude.sh`, stop and split — that is a signal the interlock is being built at
the wrong layer.

## Affected Files

- `loom-daemon/src/main_health_gate.rs` — add `gate_in_flight` state to `MainHealthState` (:157) + a
  `WorkspaceHealthStates` passthrough (mirror :475-486); set/clear it around the `spawn_blocking` call
  (:2418) via an RAII guard covering the panic arm (:2436); add the config/env knob alongside
  `read_autonomous_gate_config` (:2154) and `resolve_enabled` (:2199). **Do not touch** the timeout kill
  path (:1920-1940) — AC2 depends on it being unchanged.
- `loom-daemon/src/work_finder.rs` — consult `gate_in_flight` as a dispatch suppressor next to the existing
  `is_halted` reads at :1186 (multi-root) and :994 (single-root).
- `.loom/docs/build-gate.md` — document the new knob and its precedence in the knobs section (near the
  existing `LOOM_BUILD_GATE_NICENESS` discussion, :193-230).
- `loom-daemon/src/main.rs` — reference only (:747 constructs the shared `Arc`, :888 spawns the gate loop);
  no change expected unless the knob must be resolved at startup.
- `defaults/scripts/build-gate.sh` — reference only; **no change expected**. Candidate 4 is rejected
  (Finding 3) and the niceness knob (:76) already landed via #4073.

## Test Plan

- [ ] Automated tests — `loom-daemon/src/main_health_gate.rs` (inline `#[cfg(test)]` module, the existing
      home for gate tests):
  - [ ] `gate_in_flight` is false at construction, true during a `run_gate` that blocks, false after.
  - [ ] **Latch regression**: a gate run whose blocking task *panics* still clears `gate_in_flight` — assert
        this directly, since a latched flag starves dispatch permanently and is the highest-risk failure mode.
  - [ ] The knob honors env > config > default, matching the `resolve_enabled` test pattern already in-file.
  - [ ] With the suppressor disabled, dispatch behavior is byte-for-byte unchanged (no-regression guard).
- [ ] Automated tests — `loom-daemon/src/work_finder.rs`: a root with `gate_in_flight` set dispatches zero
      new sweeps; sibling roots with no gate in flight are unaffected (the per-root isolation contract from
      #3930 must survive).
- [ ] Manual verification on a 28-core host, the scenario from the issue body:
  - [ ] Start the daemon with ≥2 issues queued and confirm the gate reaches a determinate verdict
        (Green or Red) rather than `UNEVALUATED [timeout]` — AC1.
  - [ ] Confirm sweeps resume dispatching promptly once the gate run completes (no throughput deadlock).
- [ ] Edge cases:
  - [ ] Gate disabled for a root (`resolve_enabled` false) → suppressor never engages, dispatch unaffected.
  - [ ] A root with no `buildGate` block → unchanged always-green path (`main_health_gate.rs:2400-2408`).
  - [ ] Daemon restart mid-gate-run → flag starts clean, no stale suppression.
- [ ] Verification instrument: after #4083 lands, confirm the green line's elapsed seconds sit meaningfully
      below 1200s under contention. Until then, absence of the `UNEVALUATED [timeout]` line plus the
      surrounding log timestamps is the available signal.

## Related work (not a hard blocker)

- **#4083** (open, `loom:issue` + `loom:curated`) — INFO-level green line with elapsed seconds. This is the
  *measurement instrument* for confirming margin, not a prerequisite for implementing the interlock. This
  issue is deliberately **not** marked `loom:blocked`: the fix is implementable today, and AC1 is observable
  from the absence of the timeout line. Landing #4083 first is preferred but not required.
- **#4020** (closed) — rescoped to this problem, closed on #4073's mild-niceness change.
- **#4073** (merged `118e67fd`) — niceness 19 → 5. Necessary, not sufficient; this issue records that.
- **#3978** / `loom-daemon/src/cpu_headroom.rs` — the existing CPU-headroom axis on the dynamic concurrency
  cap. Prior art for "back off dispatch under CPU pressure"; note it did **not** prevent this incident,
  because it bounds *how many* sweeps run, not *when* they start relative to a gate run.

rjwalters:
## Summary

The daemon's periodic build-gate run kept timing out at the full 1200s budget under two concurrent sweeps on a 28-core host, even after #4073 dropped its handicap to `nice -n 5`. The A/B in #4084 is clean: the same repo compiled green in ~726s idle but exceeded 1200s under contention — a >65% wall-clock inflation attributable to CPU contention alone. Niceness only reorders the run queue; it does not stop the gate's own build from racing freshly-dispatched sweep builds for cores.

This implements the curator's recommended fix (candidate 1, dispatch-suppression half only): hold new dispatch off a root while that root's own build-gate run is in flight. It reuses the existing `WorkspaceHealthStates` plumbing already shared between the gate loop (writer) and the work-finder (reader), keeps the timeout budget fixed (so AC2 holds), and does not touch the sweep registry, epic supervisor, or `spawn-claude.sh`.

## Changes

- **`loom-daemon/src/main_health_gate.rs`**
  - Add a per-root `gate_in_flight: AtomicBool` to `MainHealthState` with `is_gate_in_flight` / `set_gate_in_flight` accessors, mirroring the `is_halted` / `set_halted` pair, plus a `WorkspaceHealthStates::is_gate_in_flight(root)` passthrough.
  - Add a `GateInFlightGuard` RAII guard that sets the flag on construction and clears it on `Drop`. It is moved *into* the `spawn_blocking` closure in both the single-root and multi-workspace gate loops, so the flag clears on normal return, early return, **or panic unwind** — a panicking gate run can never latch the flag on and permanently starve dispatch.
  - Add the `autonomous.mainHealthGate.suppressDispatchDuringGate` config knob (default `true`) + `LOOM_MAIN_HEALTH_GATE_SUPPRESS_DISPATCH` env override, resolved via `resolve_suppress_dispatch_during_gate` (env > config > default), following the `resolve_enabled` pattern.
- **`loom-daemon/src/work_finder.rs`** — new pure helper `dispatch_held_per_root` folds `gate_in_flight` into the per-root held vector alongside `is_halted`, gated by the suppressor knob (strictly per root, so a sibling with no gate in flight is unaffected). Single-root loop updated to match.
- **`loom-daemon/src/main.rs`** — resolve the knob once at startup and thread it into `spawn_multi_work_finder_task`.
- **`.loom/docs/build-gate.md`** — document the knob and its env > config > default precedence next to the `LOOM_BUILD_GATE_NICENESS` discussion.

1. **A contended gate reaches a determinate verdict, not `UNEVALUATED [timeout]`** — the work-finder no longer dispatches fresh sweeps into a root whose gate build is running, removing the CPU contention that inflated the gate's wall-clock past 1200s. (Full 28-core e2e reproduction is not feasible in CI; verified structurally + via unit tests.)
2. **A genuinely hung gate is still killed and reported** — the timeout kill path (`main_health_gate.rs` `try_wait` loop) is **untouched**; the budget stays fixed.
3. **Knob follows env > config > default and is documented** — `resolve_suppress_dispatch_during_gate` implements the precedence (tested), documented in `build-gate.md`.

## Test Plan

All in the crate's inline `#[cfg(test)]` modules (`cargo test --lib`, full suite green: 1270 passed):

- `gate_in_flight` false at construction; true during a guard's lifetime; false after.
- **Latch regression**: a guard dropped during a `catch_unwind` panic still clears `gate_in_flight`.
- `WorkspaceHealthStates` passthrough is per-root (sibling unaffected).
- Knob honors env > config > default; config value parsed from JSON.
- `dispatch_held_per_root`: a gate-in-flight root is held only on its own root; with the suppressor disabled the result is byte-for-byte the `is_halted`-only vector.
- End-to-end through `tick_multi`: a gate-in-flight root dispatches zero new sweeps while its healthy sibling gets the shared slot.

## Notes / deviations

- **Scope**: +386 insertions across the 4 targeted files, modestly over the issue's ~300-LOC soft target. The overage is entirely house-style doc comments and the nine requested unit tests (~120 lines); no scope creep into the sweep registry, epic supervisor, or spawn path (the guard's hard constraint). Candidates 2 (shared `CARGO_TARGET_DIR`) and 4 (cap gate `-j`) were explicitly left out per the curator analysis.
- Rollout: this is a daemon-side change and requires a daemon rebuild + restart to take effect (unlike #4073's pure-shell niceness change).

Closes #4084


rjwalters:
## Judge review — changes requested (merge conflict, not a code-quality issue)

**Code review: the implementation itself is solid.** I read the full diff, checked out the branch, and ran the crate's tests locally:

- `gate_in_flight: AtomicBool` on `MainHealthState` + `WorkspaceHealthStates::is_gate_in_flight` passthrough — clean, mirrors the existing `halted`/`is_halted` pair.
- `GateInFlightGuard` RAII guard moved *into* the `spawn_blocking` closure in both the single-root and multi-workspace gate loops — verified this clears on the panic path too: `test_gate_in_flight_guard_clears_on_panic` (uses `catch_unwind`) passes locally, and I re-ran it in isolation 1x plus the rest of the suite (`cargo test --lib main_health_gate:: work_finder::`) — all green (111 + 87 relevant tests, 0 failures).
- **AC2 verified structurally**: `git diff` hunks confirm the timeout kill path at `main_health_gate.rs` (`try_wait` loop, ~:1920-1940) is untouched — no hunk touches that region. The flag is a pure dispatch-suppression signal; the kill/timeout budget logic is unmodified.
- `resolve_suppress_dispatch_during_gate` follows the env > config > default precedence pattern correctly (`test_resolve_suppress_dispatch_during_gate_precedence` covers all four combinations).
- `dispatch_held_per_root` is per-root and covered end-to-end through `tick_multi` (sibling-root isolation, suppressor-off no-regression).
- `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` both clean locally.
- CI's "Rust Unit Tests" initially failed on `role_runner::tests::test_invoke_writes_per_role_log_file` ("Text file busy (os error 26)") — confirmed this is an unrelated pre-existing flake (that file isn't touched by this PR's diff, and the test passes consistently in isolation locally); reran the failed CI job and it passed clean on retry.

**Blocker: the PR now has a real merge conflict, not a GitHub eventual-consistency blip.** `gh pr view` currently reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`. I verified this is genuine (via a local `git merge --no-commit --no-ff origin/main` in a scratch check, then aborted — no conflict resolution was committed or pushed):

```
loom-daemon/src/work_finder.rs:1251 (spawn_multi_work_finder_task)
<<<<<<< HEAD (this PR, #4084)
            let halted =
                dispatch_held_per_root(&health_states, &roots, suppress_dispatch_during_gate);
=======
            // A scheduled drain (#4090) is daemon-global ...
            let draining = drain.load(std::sync::atomic::Ordering::Relaxed);
            let halted: Vec<bool> = roots
                .iter()
                .map(|r| health_states.is_halted(r) || draining)
                .collect();
>>>>>>> origin/main
```

`#4134` (scheduled drain-and-restart, merged after this PR was opened) landed a competing edit to the exact same `halted` computation in `spawn_multi_work_finder_task`. Both terms are needed — dropping either silently regresses the other issue (#4084's gate-in-flight suppression or #4090's drain-global halt). Please rebase onto latest `origin/main` and combine both terms, e.g.:

```rust
let draining = drain.load(std::sync::atomic::Ordering::Relaxed);
let halted: Vec<bool> = dispatch_held_per_root(&health_states, &roots, suppress_dispatch_during_gate)
    .into_iter()
    .map(|h| h || draining)
    .collect();
```

(or fold `draining` into `dispatch_held_per_root` itself — your call on the cleanest shape). Please add/adjust a test asserting the combined semantics (drain holds every root regardless of gate state; gate-in-flight holds only its own root regardless of drain state) survive the merge.

**Minor, non-blocking nit for this pass or a follow-up:** the config-knobs reference table in `.loom/docs/daemon-reference.md` (~:1419-1420, the `autonomous.mainHealthGate.*` rows) doesn't have an entry for the new `suppressDispatchDuringGate` knob — `.loom/docs/build-gate.md` documents it well, but the canonical cross-repo knobs table is now slightly out of sync. Feel free to add a row while you're rebasing, or leave for a follow-up — not blocking.

Once rebased and `mergeStateStatus` is `CLEAN`, this looks ready to re-request review.

rjwalters:
✅ **Approved!**

Reviewed the diff, cross-checked against issue #4084's curator recommendation, and ran the test suite.

## Correctness
- `GateInFlightGuard` is a clean RAII pattern: set on construction, cleared on `Drop` — and since the guard is moved *into* the `spawn_blocking` closure (not just held around the `.await`), it clears on normal return, early return (#3984 skip path), *and* panic unwind. This directly addresses the highest-risk failure mode called out in the issue (a latched flag permanently starving dispatch), and there's an explicit `catch_unwind` regression test for it.
- `dispatch_held_per_root` / `dispatch_held_per_root_with_drain` correctly compose the three independent terms (verified-red halt, per-root gate-in-flight, daemon-global drain #4090) — each is additive/OR'd, per-root isolation (#3930) is preserved (a sibling root with no gate in flight is never held), and with the suppressor off or drain off the result collapses byte-for-byte to the pre-#4084 vectors (verified by the "no-regression" tests).
- `resolve_suppress_dispatch_during_gate` follows the established env > config > default precedence pattern (mirrors `resolve_enabled`), and it's resolved once at startup in `main.rs` and threaded through to `spawn_multi_work_finder_task` — no per-tick re-read overhead.
- The timeout kill path (`main_health_gate.rs` `try_wait` loop, ~line 1920-1940) is untouched, so AC2 (a genuinely hung gate is still killed and reported) holds.

## Scope vs. issue
This closely follows the curator's "candidate 1, dispatch-suppression half only" recommendation in #4084 almost line-for-line — same field name, same accessor pattern, same config key (`autonomous.mainHealthGate.suppressDispatchDuringGate`), same env var, same default (`true`). The one addition beyond the curator's literal plan is the second commit refactoring the drain (#4090) + gate-in-flight combination into `dispatch_held_per_root_with_drain` — a reasonable, well-tested consolidation rather than scope creep, since #4090's drain flag already needed to compose with this new term at the same call site.

## Test Execution
- `cargo build --lib` — ✅ clean build
- `cargo test --lib -- gate_in_flight suppress_dispatch dispatch_held` — ✅ 13/13 new tests pass
- `cargo test --lib` (full suite) — ✅ 1284 passed, 0 failed, 0 ignored (170s)
- `gh pr checks 4138` — ✅ all required checks pass (Rust Unit Tests, Rust Linting & Build Check, CLAUDE.md Line Budget)
- `mergeStateStatus` — ✅ CLEAN

PR description uses `Closes #4084` correctly. Documentation (`.loom/docs/build-gate.md`, `daemon-reference.md`) is clear and includes the rationale for *not* raising the timeout budget instead, matching the issue's own argument against that lever.

Nice work — this is a faithful, well-tested implementation of the curator's diagnosis.

Issue #3997, opened 2026-07-27
Support safehouse as the fleet comms layer (E2E Matrix room alongside forge labels)


rjwalters:
## Curator Enhancement (2026-07-27)

> **Scope narrowed.** The original proposal spans daemon emission, per-worker personas, and inbound human steering — together well past a single Builder PR. This issue is rescoped to **phase 1: daemon-side, emit-only**. The deferred halves are tracked in #3999 (worker persona + MCP tools) and a follow-up for inbound steering. Original text preserved at the bottom.

### Status change since filing

The body says safehouse is a "local repo today, publishing soon". **It is published now** — `rjwalters/safehouse` is a PUBLIC repo. The protocol and socket surface below were read from that tree on 2026-07-27 and are no longer speculative.

### Problem Statement

Loom agents coordinate exclusively through forge labels. That works for machine-to-machine state, but it is a poor **narration** channel: a human watching a sweep has to poll `gh` or tail daemon logs to know what is happening, and has no in-band way to see progress across a multi-host fleet. Safehouse provides an E2E Matrix room that is already the human's window into the fleet — loom should be able to narrate into it.

Label coordination stays exactly as-is. This adds an **optional, additive** side-channel that is a byte-for-byte no-op when unconfigured.

### Verified integration surface (safehouse @ 2026-07-27)

Read from `/Users/rwalters/GitHub/safehouse`. These constraints shape the design and were **not** in the original proposal:

**Envelope v1** (`docs/protocol/envelope-v1.md`, impl `safehoused/src/envelope.rs`)
- Carrier is an ordinary `m.room.message` with a namespaced key `org.safehouse.envelope` — *not* a custom event type.
- Required fields: `v` (always `1`), `from`, `to`, `type`, `body`. Optional: `task_id`, `wake`.
- `type` is a **closed 4-value enum**: `chat` | `task` | `handoff` | `ack` (`envelope.rs:10`). Anything else is rejected on send.
- `task_id` must match `[A-Za-z0-9_]` — a bare issue number like `4137` is fine; `issue-4137` is **not** (hyphen rejected).
- **`from` is stamped by the daemon and ignored if the client sends it** — a worker cannot impersonate another. Identity is the socket connection.
- Personas are `[a-z0-9_]`, 1–64 chars, lowercase. Hyphens are a *render-time* cosmetic only (§8); **always send underscored**. `to` is NOT normalized on send, so a hyphenated `to` silently routes nowhere.

[...]

PR #4139, opened 2026-07-28
feat(safehouse): inject safehouse-mcp into workers via a pre-registered persona pool (#3999)

### LINKED ISSUE #3999: Expose safehouse-mcp tools to workers when a safehoused socket is configured
## Original Issue

Safehouse ships `safehouse-mcp`, a keyless stdio MCP server over the safehoused unix socket (tools: `safehouse_send`, `safehouse_read`, `safehouse_create_room`, `safehouse_list_rooms`; env: `SAFEHOUSED_SOCKET`, `SAFEHOUSE_PERSONA`). It holds no keys or tokens — the daemon gates personas against its allowlist and stamps sender identity.

When loom's `safehouse` config block is present, worker sessions should get this MCP server injected into their MCP config with `SAFEHOUSE_PERSONA` set to the worker's assigned persona — so a builder can ask questions in the room mid-task and read the human's answer, rather than being label-only.

Depends on the config surface from the safehouse-comms-layer issue.

---

### Problem Statement

Loom agents coordinate exclusively through forge labels — a one-way, coarse channel.
A Builder that hits an ambiguity mid-task has no way to ask; it must guess, or fail and
leave the question in a PR body for a human to find later. Injecting `safehouse-mcp`
into worker sessions gives a worker a *synchronous* back-channel to the operator's room
while keeping the security property that matters: the MCP server holds no keys or
tokens, and `safehoused` gates personas against its allowlist and stamps sender identity,
so a compromised or confused worker cannot impersonate another persona.

### Dependencies

- [ ] #3997: Support safehouse as the fleet comms layer (`loom:curated`, not yet approved) — owns the `safehouse` config block, the `SAFEHOUSED_SOCKET` path, and the persona model this issue keys off.

This issue cannot proceed until #3997 is complete. Verified 2026-07-27: `grep -rn safehouse`
over the repo (excluding `node_modules`/`target`) returns **zero** hits, and `.loom/config.json`
has only `version, offlineMode, autonomous, buildGate, terminals` — the config block to gate on
does not exist.

Related: #3998 (provisioning cloud workers into the same fleet) — a sibling consumer of the
same config surface, also blocked on #3997.

### Acceptance Criteria

- [ ] With a `safehouse` block configured, a worker session's MCP config contains a `safehouse` server entry with `SAFEHOUSED_SOCKET` and `SAFEHOUSE_PERSONA` set for that worker.
- [ ] `SAFEHOUSE_PERSONA` reflects the **worker's assigned persona**, not a workspace-wide constant — two workers running concurrently in the same workspace get different values.
- [ ] With no `safehouse` block, generated MCP config is byte-identical to today's (no empty entry, no env stubs) — non-safehouse repos are unaffected.
- [ ] If the socket path is configured but absent/unconnectable at spawn time, the worker still starts with the loom MCP server working; the failure is logged, not fatal.
- [ ] Injection reaches sessions that run **inside a worktree**, not only sessions in the workspace root (see constraint below).
- [ ] No token or key is written into the injected config — the socket is the only credential surface.

### Implementation Guidance

**Constraint a Builder must design around (verified 2026-07-27):** `.mcp.json` is
generated by `scripts/setup-mcp.sh` as a **static heredoc** containing exactly one server
(`loom`), and `.mcp.json` is **gitignored** (`.gitignore:18`) — so it is *not* present in
`.loom/worktrees/issue-N`, where Builder sessions actually run. Meanwhile `.loom/scripts/spawn-claude.sh`
— the mandatory spawn path for every agent — contains **no MCP handling at all**. A
per-worker, per-persona env var therefore cannot be delivered by editing the existing
static generator alone; either the generator must run per-worktree at spawn time, or
`spawn-claude.sh` must gain an MCP-config step. Choose deliberately and state the choice in
the PR body.

Two viable shapes:

1. **Generate at spawn time** — `spawn-claude.sh` writes a session-scoped MCP config (with the persona substituted) into the session's cwd or passes it via CLI flag. Pro: naturally per-worker; the persona is known exactly where the worker is spawned. Con: a new responsibility in the hottest, most safety-critical script in the repo.
2. **Extend `setup-mcp.sh` + invoke it per-worktree** — keep generation in one place, call it from the worktree setup path. Pro: single generator. Con: the persona must be threaded into worktree creation, and worktrees are also created by hand (`worktree.sh`) where no persona exists.

Whichever is chosen, keep the `safehouse` entry **additive**: the `loom` server entry must
be unchanged so the #3803-style staleness/pre-flight logic in `defaults/scripts/claude-wrapper.sh`
(which extracts an entry point from the *first* server with args) keeps working. Verify that
pre-flight explicitly — it iterates `mcpServers` and exits on the first entry with args, so
adding a server ahead of `loom` in iteration order could silently retarget the staleness check.

The safehouse tool names, env var names, and socket semantics quoted above live in an
**external** project and are not verifiable from this repo — confirm them against safehouse
at implementation time.

### Affected Files

- `scripts/setup-mcp.sh` — the `.mcp.json` generator (static heredoc at lines ~53-64); make the server set conditional on the `safehouse` config block.
- `.loom/scripts/spawn-claude.sh` — mandatory agent spawn path; currently has no MCP handling. Likely home for per-worker persona injection.
- `defaults/scripts/claude-wrapper.sh` — MCP pre-flight (~lines 377-400) extracts an entry point from `mcpServers`; verify a second server does not break it.
- `.gitignore` — `.mcp.json` is ignored at line 18; relevant to whether worktree sessions see any MCP config.
- The `safehouse` config reader introduced by #3997 — read the socket path and persona model from it rather than re-parsing `.loom/config.json` ad hoc.
- `.loom/docs/` — document the injected server + persona semantics; one-line pointer in `CLAUDE.md` only.

### Test Plan

- [ ] Manual verification: with a `safehouse` block configured, spawn a Builder and confirm `safehouse_send`/`safehouse_read` are callable from inside the session and that a message lands in the room stamped with the worker's persona.
- [ ] Manual verification: spawn two workers concurrently; confirm each carries a distinct `SAFEHOUSE_PERSONA` and messages are attributed correctly.
- [ ] Manual verification: run a Builder in a worktree (`.loom/worktrees/issue-N`) — the injected config must be present there, not only in the workspace root.
- [ ] Automated tests: with no `safehouse` block, generated `.mcp.json` is byte-identical to the current output (regression guard for non-safehouse repos).
- [ ] Automated tests: MCP pre-flight in `claude-wrapper.sh` still resolves the `loom` entry point with two servers present.
- [ ] Edge cases: missing/unconnectable socket ⇒ worker starts, loom MCP works, failure logged not fatal; persona absent from the daemon's allowlist ⇒ rejected by `safehoused` with a clear message rather than a silent no-op.
- [ ] Security check: grep the generated config and spawn logs for any token or key — zero hits.

rjwalters:
Closes #3999

## What this does

Phase 2 of the safehouse fleet-comms layer (#3997 shipped the daemon-side, emit-only phase 1). When the `safehouse` config block is enabled, worker sessions now get the keyless `safehouse-mcp` stdio server injected into their MCP config with `SAFEHOUSED_SOCKET` + a **per-worker** `SAFEHOUSE_PERSONA`, so a Builder can `safehouse_send`/`safehouse_read` in the room mid-task instead of being label-only.

### 1. Per-worker persona = bounded, pre-registered POOL (not literal `loom_builder_<issue>`)

The issue's literal criterion asks for "the worker's assigned persona … `loom_builder_42`". That exact dynamic-naming path is **genuinely blocked externally**: `.loom/docs/safehouse.md` (and `loom-daemon/src/safehouse.rs:30-34`) confirm safehoused reads its persona allowlist **once at boot** from a static TOML array — no runtime registration, no glob/prefix matching, no SIGHUP reload. A name minted at dispatch time would be rejected at `hello`, and safehoused cannot restart per worker.

So this PR implements the acceptance *criterion* ("two workers running concurrently in the same workspace get different values") via a **bounded pool the operator pre-registers** in safehoused's allowlist, exactly the shape of the existing token pool: configure `safehouse.workerPersonas: ["loom_builder_1", …]`, add the same names to safehoused's `personas`, and each worker is assigned `workerPersonas[issue_number % pool_size]` (round-robin by worktree slot, issue number from `LOOM_SWEEP_CLAIM_OWNED`). With no pool configured it falls back to the scalar `safehouse.persona` (workspace-wide) — degrade, never fail. This needs **zero** safehoused-side change.

### 2. Injection shape = session-scoped `--mcp-config` at spawn time (Option 1)

Chosen over "extend setup-mcp per-worktree" because **concurrent sweeps share the workspace root** — a per-worker persona cannot live in the shared `.mcp.json`. `spawn-claude.sh` (the mandatory spawn path, where the persona is known) generates a session-scoped config with the persona substituted and passes `claude --mcp-config <file>`. The file lists `loom` FIRST so it is self-contained even when the session cwd has no project `.mcp.json`.

`scripts/setup-mcp.sh` *also* learns to append the safehouse server (with the **scalar** persona, since it is not per-worker) to the workspace-root `.mcp.json` — this covers direct/human `claude` sessions and reaches worktree sessions via the `.mcp.json` symlink `worktree.sh` already creates. Both writers keep `loom` first and byte-for-byte unchanged.

### 3. "Can't verify safehouse-mcp's binary/protocol" gap → configurable + graceful

The `safehouse-mcp` command, tool names, and env vars live in the external `rjwalters/safehouse` repo and are not verifiable here. So the launcher is configurable (`safehouse.mcpCommand`, default `safehouse-mcp`) and follows #3997's degradation contract: a **missing launch command or unresolvable socket → one `warn` + skip injection**, `loom` MCP unaffected, worker always starts. A socket that is configured but not yet present is injected best-effort (safehouse-mcp connects lazily). No token/key is ever written — the socket path is the only credential-adjacent value.

## Acceptance criteria

- [x] Enabled → worker MCP config has a `safehouse` server with `SAFEHOUSED_SOCKET` + per-worker `SAFEHOUSE_PERSONA` (pool, distinct per concurrent worker).
- [x] Disabled/absent → generated config byte-identical to today (automated regression test: `setup-mcp.sh` output compared byte-for-byte; spawn-claude appends no `--mcp-config`).
- [x] Socket configured but absent/unconnectable → worker starts, loom MCP works, logged not fatal.
- [x] Injection reaches worktree sessions (spawn-claude injects on the top-level sweep session inherited by Task subagents; setup-mcp's file reaches worktrees via the existing `.mcp.json` symlink).
- [x] No token/key in the injected config (security grep test — zero hits).
- [x] `loom` entry unchanged and first — test exercises `claude-wrapper.sh`'s exact pre-flight extraction with two servers present and asserts it still returns the loom entry point.

## Testing

`defaults/scripts/tests/test-mcp-config.sh` — 43 assertions (resolver precedence env>config>default, pool round-robin distinctness, loom-first emit, no-secrets grep, claude-wrapper two-server pre-flight, byte-identical setup-mcp regression). Verified `shellcheck -x` clean on all four changed shell files and the existing `test-spawn-claude.sh` (80 assertions) still passes. Manually smoke-tested spawn-claude's three paths (disabled → no `--mcp-config`; enabled → `--mcp-config` with correct per-issue persona, loom first; missing command → skip + warn).

## Notes for reviewer

- `defaults/config.json` gains `safehouse.workerPersonas` + `safehouse.mcpCommand`; `loom-daemon/src/safehouse.rs` reads only known keys (no `deny_unknown_fields`) so the daemon ignores them.
- `.loom/scripts` is a symlink to `defaults/scripts`, so the spawn-claude/lib edits land in `defaults/`.

### DIFF
--- a/.loom/docs/safehouse.md
+++ b/.loom/docs/safehouse.md
@@ -122,3 +122,100 @@ by the bare issue number (`task_id`):
 - `loom-daemon/src/workspace_pool.rs` — `start_safehouse_narration()` subscribes
   the shared `Arc<EventBus>` (the single place it is owned) and spawns the sink
   on the daemon runtime; a no-op when disabled.
+
+# Phase 2 — worker-side `safehouse-mcp` injection (#3999)
+
+Phase 1 lets the daemon *narrate*. Phase 2 gives each **worker** session a
+two-way handle: when the `safehouse` block is enabled, Loom injects the
+`safehouse-mcp` stdio MCP server (`rjwalters/safehouse`, tools `safehouse_send` /
+`safehouse_read` / `safehouse_create_room` / `safehouse_list_rooms`; env
+`SAFEHOUSED_SOCKET` + `SAFEHOUSE_PERSONA`) into the worker's MCP config, so a
+Builder can ask a question in the room mid-task and read the human's answer
+instead of only signalling through labels. The MCP server holds no keys — the
+socket path is the only credential-adjacent value written.
+
+## Per-worker persona: a bounded pre-registered pool (design decision)
+
+safehoused's persona allowlist is a **static boot-time TOML array** with no
+runtime registration, no glob/prefix matching, and no SIGHUP reload (see phase-1
+note above). So a literal per-issue name like `loom_builder_42` **cannot** be
+minted at dispatch time — safehoused would reject the `hello` for a name not in
+its boot allowlist, and it cannot restart per worker.
+
+Loom therefore assigns each worker a persona from a **bounded pool the operator
+pre-registers** in safehoused's allowlist ahead of time — the same "fixed pool,
+rotate per slot" shape as the token pool. Configure the pool in the `safehouse`
+block and add the identical names to safehoused's `personas`:
+
+```jsonc
+"safehouse": {
+  "enabled": true,
+  "socket": "/run/safehoused.sock",
+  "persona": "loom_daemon",                     // scalar fallback (daemon + no-pool workers)
+  "workerPersonas": ["loom_builder_1",          // the pre-registered worker pool
+                     "loom_builder_2",
+                     "loom_builder_3",
+                     "loom_builder_4"],
+  "mcpCommand": "safehouse-mcp"                  // launcher for the stdio MCP server
+}
+```
+
+```toml
+# safehoused config — restart required after editing (allowlist read once at boot)
+personas = ["loom_daemon", "loom_builder_1", "loom_builder_2", "loom_builder_3", "loom_builder_4"]
+```
+
+Each worker is assigned `workerPersonas[issue_number % pool_size]` (round-robin
+by worktree slot — the issue number comes from `LOOM_SWEEP_CLAIM_OWNED`). Two
+**concurrently-running** workers (distinct issue numbers) get distinct personas
+whenever the pool is at least as large as the concurrency level and the numbers
+do not collide mod N — so size the pool to your max concurrent workers. With **no
+`workerPersonas`** configured, every worker falls back to the scalar `persona`
+(workspace-wide, no per-worker attribution) — the feature degrades, never fails.
+
+Env overrides (each wins over config): `LOOM_SAFEHOUSE_WORKER_PERSONAS`
+(comma-separated pool), `LOOM_SAFEHOUSE_MCP_COMMAND`, plus the phase-1
+`LOOM_SAFEHOUSE_ENABLED` / `LOOM_SAFEHOUSE_SOCKET` / `LOOM_SAFEHOUSE_PERSONA`.
+
+## Delivery: session-scoped `--mcp-config` at spawn time
+
+Injection happens in `spawn-claude.sh` (the mandatory agent spawn path), not by
+rewriting the workspace `.mcp.json`. Concurrent sweeps **share** the workspace
+root, so a per-worker persona cannot live in that shared file; instead
+spawn-claude generates a **session-scoped** MCP config (persona substituted for
+this worker) and passes it via `claude --mcp-config <file>`. The file lists the
+`loom` server FIRST (so it is self-contained even when the session cwd has no
+project `.mcp.json`) and appends `safehouse` second.
+
+`scripts/setup-mcp.sh` (the workspace-root generator, reached inside worktrees
+via the `.mcp.json` symlink `worktree.sh` creates) **also** learns to append the
+`safehouse` server when enabled — but with the scalar `persona`, since it is not
+per-worker. Both writers keep `loom` first and unchanged so
+`claude-wrapper.sh`'s MCP pre-flight (which keys off the first server with args)
+still resolves the loom entry point.
+
+## Degradation contract (unchanged from phase 1)
+
+- `safehouse.enabled` false/absent ⇒ **byte-for-byte no-op**: spawn-claude
+  appends no `--mcp-config`, and setup-mcp emits the identical loom-only file.
+- Enabled but the launch command is missing, or no socket resolves ⇒ one
+  `warn`, **injection skipped**, the `loom` MCP server unaffected and the worker
+  starts normally.
+- Socket configured but not yet present at spawn ⇒ one `warn`, injected anyway
+  (best-effort — `safehouse-mcp` connects lazily and never blocks the worker).
+- A persona absent from safehoused's boot allowlist is rejected by safehoused at
+  `hello` with a clear message — provision the whole pool before enabling.
+
+## Implementation (phase 2)
+
+- `defaults/scripts/lib/mcp-config.sh` — shared resolvers (env > config >
+  default, mirroring `safehouse.rs`), the pool round-robin persona picker, and
+  the `loom`-first `.mcp.json` emitter.
+- `defaults/scripts/spawn-claude.sh` — per-worker injection via `--mcp-config`.
+- `scripts/setup-mcp.sh` — workspace-root two-server generation when enabled.
+- Tests: `defaults/scripts/tests/test-mcp-config.sh`.
+
+> The exact `safehouse-mcp` binary/protocol lives in the external
+> `rjwalters/safehouse` repo and is not verifiable from this repo, so the
+> launcher is configurable (`safehouse.mcpCommand`) and a missing command
+> degrades to a logged skip rather than a broken server entry.
--- a/defaults/config.json
+++ b/defaults/config.json
@@ -12,7 +12,9 @@
     "enabled": false,
     "socket": null,
     "room": null,
-    "persona": "loom_daemon"
+    "persona": "loom_daemon",
+    "workerPersonas": [],
+    "mcpCommand": "safehouse-mcp"
   },
   "health_monitoring": {
     "enabled": true,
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the safehouse-mcp worker injection (#3999). This is a well-engineered, honestly-scoped PR that works around a documented external blocker rather than papering over it.

## Verification against the acceptance criteria

**1. `loom` server first & unchanged (load-bearing for `claude-wrapper.sh` pre-flight)** — Verified. The python emitter uses an `OrderedDict` with `loom` inserted first and `safehouse` appended only second; `setup-mcp.sh` keeps `loom` byte-identical in both branches. Section 5 of `test-mcp-config.sh` replays the *exact* `check_mcp_server()` extraction ("first server with args wins") against a two-server config and confirms it still resolves `/ws/root/mcp-loom/dist/index.js`. Strictly additive.

**2. Byte-identical when disabled/absent** — Verified by an actual regression test (not just a claim). Section 6 runs `scripts/setup-mcp.sh` in an isolated workspace with no `safehouse` block and asserts the produced `.mcp.json` is byte-for-byte the pre-#3999 loom-only document. The disabled path deliberately reuses the original heredoc rather than routing through the new emitter, so identity is structural, not incidental. `spawn-claude.sh` appends no `--mcp-config` when disabled.

**3. No token/key injected** — Verified. Only `SAFEHOUSED_SOCKET` + `SAFEHOUSE_PERSONA` are ever written; the security grep test asserts zero hits for token/api_key/secret/password/oauth. The socket path is the only credential-adjacent value.

**4. Two concurrent workers get distinct personas** — Verified. `loom_mcp_pick_persona` round-robins `issue % pool_size`; the test proves issues 10/11/12 map to three distinct pool entries. The pool-vs-literal-naming design choice is a reasonable interpretation of the criterion given that safehoused reads its persona allowlist once at boot (no runtime/prefix registration) — a genuine external blocker, documented transparently in both the code and `.loom/docs/safehouse.md`. The honest caveat (issue numbers colliding mod N collide personas → "size the pool to your max concurrency") is called out in-code and in the PR body rather than hidden.

**5. Reaches worktree sessions** — Verified. Two mechanisms: (a) `spawn-claude.sh` injects a session-scoped `--mcp-config` on the top-level sweep session inherited by Task subagents (correct, since concurrent sweeps share the workspace root and a per-worker persona can't live in the shared file); (b) `setup-mcp.sh`'s workspace-root file reaches worktrees via the `.mcp.json` symlink that `worktree.sh:1616-1633` already creates (confirmed present).

**6. Missing/unconnectable socket or missing command → log + continue** — Verified in both writers. `spawn-claude.sh`: no socket → `log_warn` + skip; command not in PATH → `log_warn` + skip; socket configured-but-absent → `log_warn` + inject best-effort (lazy connect). `loom` MCP is never touched and the worker always starts. Respects an operator-supplied `--mcp-config` (won't clobber).

## Test execution (run locally, not trusted from the PR body)

- `defaults/scripts/tests/test-mcp-config.sh` — **43/43 pass** (jq present).
- `defaults/scripts/tests/test-spawn-claude.sh` — **80/80 pass** (no regression).
- `shellcheck -x` on all four changed shell files — **clean, exit 0**.
- CI: Installer Integration Tests + Shellcheck + CLAUDE.md budget + Detect Changes all **pass**; Rust/MCP builds correctly path-skipped (no Rust/TS changes). Merge state **CLEAN**.

## Non-blocking observation (not tracked — trivial)

`spawn-claude.sh` creates the session config via `mktemp` but never unlinks it after the `claude` process exits, so tiny `loom-mcp-config.XXXXXX` files accumulate in `TMPDIR` across spawns. Harmless (a few hundred bytes each, `/tmp`-swept on reboot), but a `trap … EXIT` cleanup would be tidy if this path is touched again. Not worth blocking or a follow-up issue.

Design decisions are sound, degradation contract matches #3997, daemon (`safehouse.rs`) ignores the new keys (no `deny_unknown_fields`). Solid work.

PR #4141, opened 2026-07-28
feat(daemon): free a wedged sweep's occupancy slot before the 300s watchdog (#4003)

### LINKED ISSUE #4003: Resource leases: track sweeps and CPU/disk-intensive stages separately — agents check out a slot only for bound work
## Curator rescope (2026-07-28)

**This issue was filed against a concurrency model that no longer exists.** A verification pass against `origin/main` @ `bdab1739` found that roughly half the original proposal has already shipped, one mechanism it proposes is unavailable on this host, and one half re-opens an argument the tree records as settled. The salvageable core — **the sweep slot is checked out at `fork/exec` success rather than at demonstrated startup** — is real, confirmed, and is what this issue is now scoped to. Original text preserved at the bottom.

### What has already shipped (drop from scope)

| Original proposal | Status on `origin/main` | Evidence |
|---|---|---|
| "Sweep slots scale to `healthy_tokens × per_token_concurrency`, no small fixed ceiling" | **Already implemented** | `work_finder::resolve_dynamic_max_concurrent` (`loom-daemon/src/work_finder.rs:941`) computes `min(token_limit × per_token_concurrency, disk_headroom, cpu_headroom, configured_max)` every tick (`:1043`, `:1194`). Token axis is `capacity::token_axis_limit` (`capacity.rs:281`); `perTokenConcurrency` defaults to **2** (#3947). Documented at `.loom/docs/daemon-reference.md:1120-1136`. |
| `disk` lease sizing | **Already implemented** | `disk_headroom::disk_headroom_limit` (`disk_headroom.rs:128`) = `floor(free_gb / LOOM_PER_WORKTREE_GB)`, plus the Bash twin `defaults/scripts/lib/disk-headroom.sh` used by `/loom:sweep` Stage -1. |
| `disk` lease *mutual exclusion* around worktree create | **~80% already implemented** | `defaults/scripts/worktree.sh:988-1005` already holds a **repo-global, `mkdir`-atomic, PID-stamped, 600s-timeout, trap-released** lock around the whole `git worktree add` + submodule init. Genuine gaps are only: daemon visibility of that lock, telemetry, and the prune/remove path. |
| "reaper reclaims leases from dead PIDs, same as claim recovery in #3975" | **Infrastructure already exists** | `reap_once` (`sweep_registry.rs:3071`) already releases the `.loom/locks/issue-N` claim lock (`:3117`), removes the `~/.loom/sweeps.json` journal entry (`:3124`), and restores `loom:building → loom:issue`. A lease reaper would extend this, not introduce it. |

The premise sentence *"a ceiling (default 3)"* is stale — there is no fixed ceiling; `LOOM_WORK_FINDER_MAX_CONCURRENT` is now only an operator upper bound.

### Correction: `flock` is not available on this host

The original "zero-daemon fallback could be `flock` on `~/.loom/leases/cpu.N`" **cannot be implemented as written**. Four independent in-tree sites record that `flock(1)` is unavailable on stock macOS and that `mkdir` is the established atomic primitive:

- `defaults/scripts/worktree.sh:111`
- `loom-daemon/src/tokens_pool/locking.rs:6`
- `loom-daemon/src/tokens_pool/mod.rs:17`
- `loom-daemon/src/watch_registry.rs:381`

Any lock introduced by this issue **must** reuse the existing `MkdirLock` RAII guard (`loom-daemon/src/tokens_pool/locking.rs:18-31` — 5s timeout, 100ms poll, 30s stale reap) or the `mkdir` pattern in `worktree.sh:101-165`.

### Explicitly OUT of scope: the `cpu` lease

The `cpu`-lease half is **deliberately excluded** and must not be implemented under this issue. Two independent reasons:

1. **It inverts a stated design policy.** `loom-daemon/src/cpu_headroom.rs:71-88` documents, in the source: *"CPU is deliberately a soft backoff, not a hard gate."* The CPU term is also fail-open and floored at 1 so a bad reading can never wedge the cap to zero. A hard `cpu` lease reverses that on purpose and needs an explicit architecture decision, not a Builder PR.
2. **Its founding premise is recorded as falsified.** `defaults/scripts/build-gate.sh:32-67` is a 36-line comment resolving exactly this debate: the 17 red daemon tests attributed to CPU contention were **macOS `syspolicyd` exec-latency artifacts** (#4044/#4046 — 968/968 later passed with no code change), and the one real gate timeout was cold-compile cost, settled by #4048 raising the budget 600→1200s. That comment *also* pre-rejects niceing sweep children in the spawn path.

**However, the counter-evidence is live and should not be waved away:** #4084 (open, `loom:building`) reports that even after `nice -n 5` (#4073) the gate still hit a 1200s timeout under 2 concurrent sweeps. If that reproduces, it reopens the question legitimately — but it belongs in **#4084 or a new architecture proposal**, not here. Do not re-litigate it in this issue's PR.

---

## Problem Statement (rescoped)

A sweep slot is counted as occupied the instant `dispatch()` returns `Ok(true)` — which happens as soon as `spawn_child` succeeds, i.e. at **`fork/exec` success**, before the child has proven it reached the API, created a worktree, or wrote a checkpoint.

Confirmed on `origin/main` @ `bdab1739`:

- `work_finder.rs:464-468` (single-workspace `tick`) and `work_finder.rs:652-656` (`tick_multi`, the production path): `Ok(true) => { report.dispatched += 1; occupancy += 1; }`
- `Ok(true)` originates in `SweepRegistry::dispatch` (`sweep_registry.rs:1529`), returning after `spawn_child` succeeds (`:1613-1618`).
- Occupancy is re-derived from `SweepState::Running | Pending` (`work_finder.rs:1489-1510`).

**Consequence:** a child that wedges at startup holds a slot for the full **300s** watchdog window (`DEFAULT_WATCHDOG_TIMEOUT_SECS`, `sweep_registry.rs:454`; probed every 30s at `:457`) before anything reclaims it. Under a small cap that is a large fraction of total throughput spent on a process doing nothing.

**Why this matters now:** #4088 raised the watchdog 120→300s *because dispatch→worktree latency was measured at 110–150s under concurrency* (`sweep_registry.rs:449-453`). So the "spawned but not yet working" window is not hypothetical — it is routinely 2+ minutes wide, and the mitigation for it widened the exposure.

This is the scope originally proposed in #4007 (*"stagger by completion, not a fixed delay — hold the concurrency slot until the child demonstrably passes startup"*), folded here rather than duplicated.

### Existing partial mitigations (do not duplicate)

- **Startup watchdog** — `watchdog_once` (`sweep_registry.rs:3376`), 300s timeout / 30s interval / 3s cancel grace (`:454-461`).
- **Insta-crash quarantine** — `DEFAULT_QUARANTINE_INSTA_CRASH_SECS = 60` (`:577`), TTL 3600s (`:570`); skipped at `work_finder.rs:453` / `:623` and documented as "never consumes a slot".
- **Dispatch stagger** — `apply_dispatch_stagger` (`sweep_registry.rs:2536`), `DEFAULT_DISPATCH_STAGGER_MS = 2000` (`:428`).

These bound the *worst case*; none of them make the slot reflect actual work.

---

## Acceptance Criteria

- [ ] A dispatched-but-not-yet-started sweep is distinguishable from a running one in occupancy accounting — the work finder can tell "spawned" from "demonstrably started".
- [ ] A concrete, documented **startup-proof signal** is chosen and implemented (e.g. checkpoint file written, worktree created, or an explicit child→daemon ready ping). The choice and its rationale are recorded in the PR body.
- [ ] A child that never reaches the startup-proof signal stops consuming admission capacity **sooner than the current 300s watchdog window**, without lowering `DEFAULT_WATCHDOG_TIMEOUT_SECS` (the 300s value is justified by measured 110–150s worktree latency — do not regress #4088).
- [ ] The change does **not** reduce the effective cap for healthy sweeps: a fleet of children that all start normally dispatches at the same rate as before.
- [ ] No busy-wait and no new blocking under the registry mutex. Note `apply_dispatch_stagger`'s `std::thread::sleep` is held under the mutex **intentionally** (`sweep_registry.rs:1604-1610`) — if this issue changes that, the PR must explain why the intentional serialization is still preserved.
- [ ] Any new lock uses `MkdirLock` / `mkdir` (see Correction above). **No `flock(1)`.**
- [ ] `cpu`-lease behavior is NOT introduced (see Out of scope).
- [ ] Behavior is observable: the daemon can report how long sweeps spend in the spawned-but-not-started state.

## Affected Files

All verified present on `origin/main` @ `bdab1739`.

- `loom-daemon/src/work_finder.rs` — occupancy accounting at `:437`, `:458`, `:464-468`, `:555-556`, `:647`, `:652-656`; `in_flight` / reap-on-read at `:1489-1510`. Primary change site.
- `loom-daemon/src/sweep_registry.rs` — `dispatch()` `:1529` (returns `Ok(true)` at `:1613-1618`), `apply_dispatch_stagger` `:2536`, `stagger_wait` `:622`, watchdog `watchdog_once` `:3376`, reaper `reap_once` `:3071`, constants `:428`/`:454`/`:457`/`:461`/`:570`/`:577`.
- `loom-daemon/src/tokens_pool/locking.rs` — reuse `MkdirLock` (`:18-31`) if a lock is needed. Reference only unless extended.
- `loom-daemon/src/ipc.rs` — status surface, if the new state is reported (`:689` is the existing headroom status path).
- `.loom/docs/daemon-reference.md` — §"Dynamic concurrency scaling" (`:1118-1136`) must be updated if occupancy semantics change.
- **Reference only, do not modify:** `loom-daemon/src/cpu_headroom.rs`, `loom-daemon/src/capacity.rs`, `loom-daemon/src/disk_headroom.rs`, `defaults/scripts/build-gate.sh`.

**Telemetry note (if the observability criterion is implemented):** `loom-daemon/src/activity/schema.rs` has **no numbered-migration framework** — new tables go into the `init_schema` `execute_batch` (`:15-291`) as `CREATE TABLE IF NOT EXISTS`, and new columns need a hand-written `migrate_*_table` helper called at `:293-295` (see `migrate_token_usage_table` `:484`). Do **not** name a new table `resource_*` — `resource_usage` (`:237-254`) already exists and is token/cost telemetry. Prior art for a claim-shaped table: `issue_claims` (`:274-289`).

## Test Plan

- [ ] **Automated — unit:** a dispatched sweep that never emits the startup-proof signal releases its admission slot before the 300s watchdog fires. Add to the existing `sweep_registry.rs` test module (in-file `#[cfg(test)]`).
- [ ] **Automated — unit:** `resolve_dynamic_max_concurrent` (`work_finder.rs:941`) is unchanged in behavior; existing cap tests still pass.
- [ ] **Automated — regression:** a full set of normally-starting children achieves the same dispatch count/rate as before the change (guards the "don't reduce throughput" criterion).
- [ ] **Automated — regression:** watchdog timeout stays 300s; insta-crash quarantine still skips quarantined issues without consuming a slot (`work_finder.rs:453`, `:623`).
- [ ] **Manual:** dispatch a sweep against an issue that wedges at startup; confirm via `loom-daemon status` that the slot frees before 300s and that a healthy queued sweep dispatches in its place.
- [ ] **Edge cases:** child killed between `fork/exec` and startup proof; child that starts then dies (must not double-release); daemon restart with a spawned-but-unproven child in the registry.
- [ ] `cd loom-daemon && cargo test` passes.

## Dependencies

None. All referenced work (#3902, #3947, #3811, #3978, #4031, #4032, #4088) is merged.

## Related

- **#4007** — origin of this scope (*"hold the concurrency slot until the child demonstrably passes startup"*); closed, folded here.
- **#4084** (open) — gate timeout under 2 concurrent sweeps; the live counter-evidence on the CPU question. Owns that argument, not this issue.
- **#4088** — raised watchdog 120→300s on measured 110–150s dispatch→worktree latency.
- **#3979** (open, `loom:architect`) — elastic compute; per-host lease pools would matter there if leases are ever pursued.
- **#3978 / #3984 / #3985 / #4044 / #4046 / #4048 / #4073** — the CPU-contention thread, recorded as settled in `build-gate.sh:32-67`.

---

## Original Issue (as filed)

Today one work-finder slot = one entire per-issue sweep, but the scarce resource the cap protects (build/test CPU, worktree disk churn) is only consumed during a fraction of each sweep's lifetime. The result is a ceiling (default 3) that is simultaneously **too strict** — three sweeps blocked on LLM/API responses leave the host idle while ready `loom:issue` work queues — and **too loose** — three sweeps hitting their build phases together still starve the build gate (#3978, #3985).

**Proposal: two-level admission.**

1. **Sweep slots** stay as-is but scale to what the *token pool* allows (they're LLM/network-bound between bursts): `healthy_tokens × per_token_concurrency`, no small fixed ceiling.
2. **Resource leases**, checked out per *stage*, released when the stage ends:
   - `cpu` lease (N ≈ cores/2): held around `cargo build`/`cargo test`/linkers — anything compile-shaped, including the build gate's own runs so it competes fairly instead of being starved.
   - `disk` lease: held around worktree create/clone/prune and other IO-heavy stages.
   A sweep that needs to compile blocks on the lease, not at dispatch; LLM-bound phases of many sweeps overlap freely.

**Mechanism sketch:** the daemon already owns a socket and a sweep registry with a reaper — leases fit naturally there (request/release ops; reaper reclaims leases from dead PIDs, same as claim recovery in #3975). A zero-daemon fallback for `/loom:sweep` sessions could be `flock` on `~/.loom/leases/cpu.N`. Stage instrumentation lives in the spawn/build scripts, so role prompts don't change. Telemetry: lease wait/hold times into activity.db would directly answer "is the host or the token pool the bottleneck."

**Relation to existing issues:** complements #3978 (a load *measurement* term still helps as a backstop, but leases prevent the contention rather than reacting to it), #3984/#3985 (gate runs acquire the same cpu lease → no more self-starvation), and #3979 (elastic compute: leases are per-host, so cloud workers bring their own lease pools).

rjwalters:
Closes #4003

## Rescoped problem

Per the curator's 2026-07-28 rescope, the original two-level `cpu`/`disk`
lease proposal was mostly already shipped or explicitly out of scope. The
salvageable core: a work-finder sweep slot is checked out (counted as
occupied) the instant `SweepRegistry::dispatch()` returns `Ok(true)` --
i.e. at `fork/exec` success -- rather than once the child has demonstrably
started. A child that wedges at startup (e.g. a hung MCP-init) therefore
held its slot for the full 300s startup-watchdog window before anything
reclaimed it, even though the watchdog's own progress signal (worktree /
checkpoint / log-past-header) is available far earlier.

## Startup-proof signal chosen (and why)

**Chosen signal**: the exact multi-signal probe the existing startup
watchdog (#3887/#4088) already polls via `SweepRegistry::sweep_made_progress`
-- ANY of:
- a worktree at `.loom/worktrees/issue-<N>`,
- a checkpoint at `.loom/sweep-checkpoint/issue-<N>.json`, or
- log output past the daemon's spawn header + `spawn-claude.sh` wrapper
  lines (`log_has_progress`).

**Rationale for reuse over a new signal**: this is already the daemon's
working definition of "the child got past pre-API local startup" -- it's
proven in production by the watchdog, and reusing it means a signal
observed by either mechanism is remembered by both (they share the same
per-`SweepId` `watchdog_progressed` latch). Inventing a second, different
startup-proof definition would let the watchdog and the occupancy
accounting disagree about whether a given sweep has "started."

**Why a shorter, separate grace window instead of just shrinking the
300s watchdog timeout** (which the issue explicitly forbids lowering):
`DEFAULT_WATCHDOG_TIMEOUT_SECS = 300` is sized to the measured **110-150s
dispatch-to-worktree** latency under concurrency (#4088) -- a late, heavy
signal (worktree creation). `log_has_progress` is a much earlier signal: it
fires the instant Claude Code itself produces *any* output past the spawn
header, which happens within seconds for a healthy child even under
contention. So a new, independent knob -- `DEFAULT_STARTUP_PROOF_GRACE_SECS
= 90` (env `LOOM_SWEEP_STARTUP_PROOF_GRACE_SECS` > config
`autonomous.watchdog.startupProofGraceSecs` > default, resolved once per
workspace exactly like `dispatch_stagger`) -- governs *occupancy
accounting only*: a sweep counts while `elapsed < grace` (a fresh dispatch
legitimately has produced nothing yet) **or** once it has proven progress.
A sweep older than `grace` with **zero** proof at all is excluded from
`SweepRegistry::occupied_issues()`, freeing its slot for a healthy queued
sweep, well before the (unchanged) 300s watchdog would cancel/re-dispatch
it.

Because a healthy child normally produces log output within seconds --
nowhere near the 90s grace, let alone 300s -- this never discounts a
normal dispatch; it only ever discounts a child that produced *literally
nothing* for the whole grace window, which is not "slow," it never
reached the API at all.

## Why this preserves every "no regression" acceptance criterion

- **Never re-dispatches the same issue.** `occupied_issues()` only feeds
  the work-finder's *budget* (`RegistryDispatcher::occupancy()`, wired
  into `tick`/`tick_multi` instead of `in_flight().len()`). The dedup set
  used to skip an issue that already has a live sweep
  (`WorkDispatcher::in_flight()`) is byte-for-byte unchanged, and so are
  the registry's claim lock and the forge `loom:building` label -- the
  actual mechanisms that prevent a double-dispatch of the *same* issue.
  Discounting a wedged sweep's occupancy only ever lets a *different*
  queued issue take the freed slot.
- **Never mutates `SweepState` or cancels anything.** `occupied_issues()`
  / `unproven_startups()` are pure reads over the existing registry state
  plus the shared latch; the watchdog, reaper, and quarantine paths are
  completely untouched (verified: existing watchdog/quarantine/reap tests
  all still pass unmodified).
- **Never discounts a healthy/normally-starting sweep, regardless of
  age.** Once progress is proven, the latch makes that permanent for the
  life of the entry -- see `occupied_issues_never_discounts_proven_sweep_regardless_of_age`,
  which uses a 1ms grace + a sweep backdated 9999s and asserts it still
  counts.
- **`DEFAULT_WATCHDOG_TIMEOUT_SECS` stays 300** -- untouched.
- **No `cpu`/`disk` lease introduced** -- out of scope per the rescope,
  and nothing here touches `cpu_headroom.rs` / `capacity.rs` /
  `disk_headroom.rs` / `build-gate.sh`.
- **No `flock(1)`** -- no new lock was needed for this scope (occupancy
  accounting is a pure in-memory read/mutate under the existing registry
  mutex); `MkdirLock` was not required.
- **No busy-wait, no new blocking under the registry mutex.**
  `occupied_issues()` is O(live sweeps) in-memory work plus a handful of
  `Path::exists()`/`read_to_string()` calls -- the same cost profile
  `watchdog_once` already pays every 30s tick, just invoked from the
  work-finder's occupancy seed instead. `apply_dispatch_stagger`'s
  intentional in-mutex sleep is untouched.

## Observability

`SweepRegistry::unproven_startups()` is a new read-only diagnostic:
`Vec<(issue, time_since_dispatch)>` for every live sweep that has not yet
proven startup, sourced from the same latch `occupied_issues()` consults
-- so the daemon can report how long a sweep has spent in the
spawned-but-not-started state (satisfies the "behavior is observable" AC)
without adding a new DB table (no `resource_*` naming collision risk, no
schema migration needed -- it's an in-memory query, not persisted
telemetry).

## Changes

- `loom-daemon/src/sweep_registry.rs`:
  - `DEFAULT_STARTUP_PROOF_GRACE_SECS` (90), `STARTUP_PROOF_GRACE_ENV`,
    `StartupRaceConfig::startup_proof_grace_secs` +
    `resolve_startup_proof_grace` (env > config > default, mirrors the
    existing watchdog knobs).
  - `SweepRegistry::{startup_proof_grace, set_startup_proof_grace}` (new
    field + accessor, mirrors `dispatch_stagger`).
  - `SweepRegistry::{has_proven_start, occupied_issues, unproven_startups}`
    -- the core occupancy-accounting logic, sharing `watchdog_progressed`
    with the existing startup watchdog.
  - 7 new unit tests under the existing `#[cfg(test)]` module covering:
    exclusion past grace, inclusion inside grace, proven-sweep immunity
    to aging, shared-latch consistency with the watchdog, the
    observability query, the setter round-trip, and env/config/default
    precedence.
- `loom-daemon/src/work_finder.rs`:
  - `WorkDispatcher::occupancy()` (default `in_flight().len()`, so test
    fakes are unaffected), overridden by `RegistryDispatcher` to call
    `occupied_issues().len()`.
  - `tick()` / `tick_multi()` seed occupancy from `dispatcher.occupancy()`
    instead of `in_flight().len()`; the `in_flight()` dedup set/contains
    check is untouched.
- `loom-daemon/src/main.rs` / `loom-daemon/src/workspace_pool.rs`: resolve
  and set the grace once per workspace (default registry at startup,
  pool-provisioned registries at first access), exactly mirroring how
  `dispatch_stagger` is threaded through today.
- `.loom/docs/daemon-reference.md`: new subsection under "Dynamic
  concurrency scaling" documenting the occupancy/grace semantics.

## Testing

- `cd loom-daemon && cargo build` -- clean.
- `cd loom-daemon && cargo test --lib` -- **1278 passed, 0 failed** (full
  suite, rebased onto current `origin/main`).
- `cargo clippy --all-targets` -- no warnings.
- `cargo fmt -- --check` -- clean.

### DIFF
--- a/.loom/docs/daemon-reference.md
+++ b/.loom/docs/daemon-reference.md
@@ -1300,6 +1300,50 @@ unparseable value for any of these falls back to its default.
 > generate work" below still holds — the finder only closes the gap between an
 > approved issue and its build.
 
+**Occupancy: startup-proof grace, distinct from the startup watchdog (#4003).**
+A dispatch slot is checked out (counted as occupied) the instant
+`SweepRegistry::dispatch()` returns — i.e. at `fork/exec` success, before the
+child has proven it reached the API, created a worktree, or wrote a
+checkpoint. Before #4003 that slot stayed occupied for the sweep's entire
+`Running`/`Pending` lifetime regardless of whether the child ever showed a
+sign of life, so a child wedged at startup (e.g. a hung MCP-init) held its
+slot for the full 300s startup-watchdog window
+(`sweep_registry::DEFAULT_WATCHDOG_TIMEOUT_SECS`) before anything reclaimed
+it.
+
+`SweepRegistry::occupied_issues()` (consumed by `RegistryDispatcher::occupancy`,
+which `tick`/`tick_multi` now use to seed occupancy instead of
+`in_flight().len()`) narrows that: a sweep still counts while it is inside its
+**startup-proof grace window** (`elapsed < grace` — a fresh dispatch legitimately
+has produced nothing yet) *or* once it has proven startup progress. Progress
+reuses the exact signal the startup watchdog already polls
+(`sweep_made_progress` — a worktree at `.loom/worktrees/issue-<N>`, a checkpoint
+at `.loom/sweep-checkpoint/issue-<N>.json`, or log output past the spawn header,
+see `log_has_progress`) and latches through the SAME per-`SweepId`
+`watchdog_progressed` set the watchdog itself maintains, so a signal observed by
+either mechanism is remembered by both. A sweep older than `grace` with **zero**
+proven signal is excluded from occupancy — freeing its slot for a healthy
+queued sweep — while the sweep's own `SweepState` stays `Running` untouched:
+this is occupancy accounting only, never a cancel/re-dispatch action. (De-dup
+safety — never double-dispatching the SAME issue — still comes from
+`in_flight()`, the registry's claim lock, and the forge `loom:building` label,
+none of which this change touches.)
+
+The grace window (default **90s**, `DEFAULT_STARTUP_PROOF_GRACE_SECS`) is
+deliberately much shorter than the 300s watchdog timeout: the watchdog's 300s is
+sized to the measured 110–150s dispatch→**worktree** latency under concurrency
+(#4088), a late, heavy signal, while `log_has_progress` fires far earlier — the
+instant Claude Code itself produces any output past the spawn header, typically
+within seconds even under contention. Tunable per workspace with the standard
+precedence **env (`LOOM_SWEEP_STARTUP_PROOF_GRACE_SECS`) > config
+(`autonomous.watchdog.startupProofGraceSecs`) > default (90)**, resolved the same
+way as `dispatch_stagger` (`SweepRegistry::set_startup_proof_grace`, set once per
+workspace at provision time by `main.rs` / `WorkspacePool::get_or_provision`).
+`SweepRegistry::unproven_startups()` is a read-only diagnostic (no latch
+mutation from the read itself; observes the same latch) returning
+`(issue, time_since_dispatch)` for every live sweep that has not yet proven
+startup — for status/observability, not for gating.
+
 ### Token-capacity backpressure (#3902)
 
 At scale, rotation accounts hit their 5h/7d rate limits and go `exhausted`.
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -764,6 +764,18 @@ async fn main() -> Result<()> {
     sweep.set_dispatch_stagger(dispatch_stagger);
     log::info!("sweep_registry: dispatch stagger = {}ms (#3887)", dispatch_stagger.as_millis());
 
+    // Startup-proof occupancy grace (Issue #4003): a freshly-dispatched sweep
+    // counts toward the work-finder's admission budget regardless of progress
+    // for this long; past it, a sweep with zero observed startup-proof signal
+    // (no worktree/checkpoint/log-past-header) stops consuming a slot, well
+    // before the (unchanged) 300s startup watchdog would cancel/re-dispatch it.
+    let startup_proof_grace = sweep_registry::resolve_startup_proof_grace(&startup_race_config);
+    sweep.set_startup_proof_grace(startup_proof_grace);
+    log::info!(
+        "sweep_registry: startup-proof occupancy grace = {}s (#4003)",
+        startup_proof_grace.as_secs()
+    );
+
     // Insta-crash quarantine (#3939): resolve env > config > default for the
     // default workspace so the reaper quarantines a repeatedly-insta-crashing
     // issue instead of letting the work finder re-dispatch it every tick.
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -456,6 +456,34 @@ pub const DEFAULT_WATCHDOG_TIMEOUT_SECS: u64 = 300;
 /// Default watchdog probe interval — matches the reaper cadence.
 pub const DEFAULT_WATCHDOG_INTERVAL_SECS: u64 = 30;
 
+/// Env var overriding the startup-proof occupancy grace window, in seconds.
+pub const STARTUP_PROOF_GRACE_ENV: &str = "LOOM_SWEEP_STARTUP_PROOF_GRACE_SECS";
+
+/// Default startup-proof occupancy grace window (Issue #4003).
+///
+/// A slot is checked out (counted as occupied) at `fork/exec` success — before
+/// the child has proven it reached the API, created a worktree, or wrote a
+/// checkpoint. That is fine for the first `grace` seconds of a fresh dispatch
+/// (a healthy child legitimately has not produced anything yet in the first
+/// moment after spawn). Past `grace`, a sweep that has shown **zero** startup
+/// signal — no worktree, no checkpoint, no log output past the spawn header,
+/// see [`log_has_progress`] — is excluded from the work-finder's occupancy
+/// count, freeing its slot for a healthy queued sweep well before the (still
+/// 300s, unchanged) startup watchdog ([`DEFAULT_WATCHDOG_TIMEOUT_SECS`])
+/// cancels and re-dispatches it.
+///
+/// Deliberately much shorter than the watchdog timeout: [`DEFAULT_WATCHDOG_TIMEOUT_SECS`]
+/// is sized to the measured 110–150s dispatch→**worktree** latency under
+/// concurrency (#4088) — a late, heavy signal. `log_has_progress` is a much
+/// earlier signal: it fires the instant Claude Code itself produces ANY
+/// output past the daemon's spawn header and the `spawn-claude.sh` wrapper
+/// lines, which happens within seconds for a healthy child even under
+/// contention. A child that produces literally nothing for this whole window
+/// is not merely "slow" — it never reached the API at all, so freeing the
+/// slot early never penalizes a healthy dispatch (see the
+/// `regression_healthy_fleet_throughput_unaffected` test).
+pub const DEFAULT_STARTUP_PROOF_GRACE_SECS: u64 = 90;
+
 /// Grace period the watchdog gives a hung child to exit after SIGTERM before
 /// escalating to SIGKILL, when it auto-cancels for re-dispatch.
 const WATCHDOG_CANCEL_GRACE: Duration = Duration::from_secs(3);
@@ -1163,6 +1191,15 @@ pub struct SweepRegistry {
     /// Instant of the most recent child spawn, used with `dispatch_stagger` to
     /// compute the stagger wait (Issue #3887). `None` until the first spawn.
     last_spawn_at: Option<Instant>,
+    /// How long a freshly-dispatched sweep counts toward the work-finder's
+    /// occupancy budget even with zero observed startup-proof signal (Issue
+    /// #4003). Defaults to [`DEFAULT_STARTUP_PROOF_GRACE_SECS`]; `main.rs` /
+    /// [`crate::workspace_pool::WorkspacePool`] set the resolved env > config >
+    /// default value per workspace, mirroring `dispatch_stagger`. Past this
+    /// window, a sweep with no worktree, no checkpoint, and no log output past
+    /// the spawn header is excluded from [`occupied_issues`](Self::occupied_issues)
+    /// — freeing its slot well before the (unchanged) startup watchdog fires.
+    startup_proof_grace: Duration,
     /// Issues the watchdog has already auto-restarted once (Issue #3887). The
     /// re-dispatch is bounded to a single attempt per issue — a second hang
     /// resolves to [`WatchdogDecision::GiveUp`], never another restart.
@@ -1300,6 +1337,7 @@ impl SweepRegistry {
             bus: None,
             dispatch_stagger: Duration::ZERO,
             last_spawn_at: None,
+            startup_proof_grace: Duration::from_secs(DEFAULT_STARTUP_PROOF_GRACE_SECS),
             watchdog_retried: HashSet::new(),
             watchdog_gaveup: HashSet::new(),
             watchdog_progressed: HashSet::new(),
@@ -1326,6 +1364,7 @@ impl SweepRegistry {
             bus: Some(bus),
             dispatch_stagger: Duration::ZERO,
             last_spawn_at: None,
+            startup_proof_grace: Duration::from_secs(DEFAULT_STARTUP_PROOF_GRACE_SECS),
             watchdog_retried: HashSet::new(),
             watchdog_gaveup: HashSet::new(),
             watchdog_progressed: HashSet::new(),
@@ -1360,6 +1399,21 @@ impl SweepRegistry {
     #[must_use]
     pub fn dispatch_stagger(&self) -> Duration {
         self.dispatch_stagger
+    }
+
+    /// Set the startup-proof occupancy grace window (Issue #4003). `main.rs` /
+    /// the workspace pool call this once per workspace at provision time with
+    /// the resolved env > config > default value, mirroring
+    /// [`set_dispatch_stagger`](Self::set_dispatch_stagger).
+    pub fn set_startup_proof_grace(&mut self, grace: Duration) {
+        self.startup_proof_grace = grace;
+    }
+
+    /// Read-only accessor for the configured startup-proof occupancy grace
+    /// window (Issue #4003).
+    #[must_use]
+    pub fn startup_proof_grace(&self) -> Duration {
+        self.startup_proof_grace
     }
 
     /// Set the insta-crash quarantine parameters (Issue #3939). `main.rs` and the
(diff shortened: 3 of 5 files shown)


rjwalters:
## Judge review — PR #4141 (closes #4003)

**Verdict: APPROVE**

### Verification performed (checked out `feature/issue-4003` @ `149b5894`, did not just trust the PR body)

- `cargo build` — clean, no warnings.
- `cargo fmt -- --check` — clean.
- `cargo clippy --all-targets` — clean, no warnings.
- `cargo test --lib` — **1278 passed, 0 failed** (127.58s), matches the builder's report exactly.
- CI (`gh pr checks 4141`): **Rust Linting & Build Check** pass, **Rust Unit Tests** pass (3m47s), `Detect Changes`/`CLAUDE.md Line Budget` pass, `Installer Integration Tests`/`MCP Server Build` correctly skipped (no matching path changes).
- Merge state: `mergeStateStatus=CLEAN`, `mergeable=MERGEABLE`.

### Against the rescoped issue's 6 guardrails (explicitly re-verified, not assumed)

1. **300s watchdog unchanged** — `DEFAULT_WATCHDOG_TIMEOUT_SECS: u64 = 300` at `sweep_registry.rs:454` is untouched by the diff; the pre-existing `resolve_watchdog_timeout_and_interval_precedence` test (from #4088) still asserts `== 300` and passes.
2. **No cpu/disk lease** — `git diff origin/main...HEAD | grep -i "cpu_headroom\|disk_headroom\|capacity::"` returns nothing; `cpu_headroom.rs`/`capacity.rs`/`disk_headroom.rs`/`build-gate.sh` are untouched, confirmed by the 5-file diff stat.
3. **No `flock(1)`** — `grep -i flock` on the full diff returns nothing; the change is a pure in-memory registry read (`occupied_issues`/`unproven_startups`), no new lock of any kind was introduced.
4. **Tests actually exercise the target behavior**:
   - `occupied_issues_excludes_unproven_sweep_past_grace_window` — a fixture child that produces zero progress signal is excluded from `occupied_issues()` once past a (shortened, 50ms) grace window, well before 300s — this is the core "never proves startup → excluded from occupancy before 300s" behavior.
   - `occupied_issues_never_discounts_proven_sweep_regardless_of_age` + `occupied_issues_keeps_fresh_dispatch_inside_grace_window` — a proven sweep is never discounted regardless of elapsed time, and a fresh dispatch counts immediately inside grace — together this is the "healthy sweeps dispatch at unchanged rate" regression guard (proxied at the `SweepRegistry::occupied_issues()` level, which is exactly what `RegistryDispatcher::occupancy()` — the seam wired into `tick`/`tick_multi` — calls after `reap_liveness()`).
   - `occupied_issues_latch_is_shared_with_watchdog` — confirms the occupancy check and the startup watchdog share the same `watchdog_progressed` latch, so they can never disagree about whether a sweep has started.
   - Full suite (1278 tests) passing confirms existing watchdog/quarantine/reap/cap tests (`resolve_dynamic_max_concurrent`, `test_tick_*_occupancy_*`, insta-crash quarantine) are unmodified and green.
5. **Docs updated** — `.loom/docs/daemon-reference.md` gained a new subsection under "Dynamic concurrency scaling" documenting the occupancy/grace semantics, the env/config/default precedence, and the shared-latch relationship with the watchdog.
6. **No `resource_*` telemetry table** — no DB schema changes at all; `activity/schema.rs` is not in the diff (5 files changed: `sweep_registry.rs`, `work_finder.rs`, `main.rs`, `workspace_pool.rs`, `daemon-reference.md`). `unproven_startups()` is a pure in-memory read, matching the PR's claim.

### Additional checks
- Confirmed `resolve_dynamic_max_concurrent` (`work_finder.rs:941`) is untouched by the diff — AC "existing cap tests still pass" holds structurally, not just by assertion.
- Confirmed `occupied_issues()`'s `SweepKind::Issue`-only filter is *not* a regression: the pre-existing `in_flight()` (used to seed occupancy before this PR) applies the exact same `SweepKind::Issue`-only filter, so PrSet sweeps were already excluded from occupancy pre-#4003.
- `WorkDispatcher::occupancy()` defaults to `in_flight().len()` (pre-#4003 behavior) so test fakes that don't override it are unaffected — good design for backward compatibility of the trait.

### Minor nit (non-blocking)
A doc comment on `DEFAULT_STARTUP_PROOF_GRACE_SECS` (`sweep_registry.rs:484`) references `` `regression_healthy_fleet_throughput_unaffected` test`` — no test with that name exists anywhere in the codebase (grepped the whole tree). The actual regression guard is `occupied_issues_never_discounts_proven_sweep_regardless_of_age`. This looks like a leftover reference from an earlier draft naming. It doesn't affect correctness/build (plain backtick text, not an intra-doc link, so it doesn't fail `cargo doc`/clippy), but worth fixing the comment to point at the real test name in a follow-up so it doesn't mislead a future reader.

Solid, well-scoped implementation: reuses the existing progress-probe signal rather than inventing a second definition, keeps the 300s watchdog and dedup/claim-lock mechanisms completely untouched, adds no new lock/lease/DB surface, and the new occupancy seam is proven both correct (excludes truly-wedged sweeps) and safe (never discounts a proven-healthy sweep). Approving.

PR #4145, opened 2026-07-28
fix(install): ship .loom/docs/* + .loom/roles/* symlink targets, retire stray loom-shepherd.md, add link checker

rjwalters:
## Summary

Three distinct packaging defects, all rooted in `defaults/` being the install
manifest source of truth: files installed content links to were never
themselves shipped. Consumer installs got a `.loom/builder.md` that pointed at
`.loom/docs/build-gate.md` and ten role files pointing at
`.loom/roles/probe-protocol.md` — neither ever existed on disk in a consumer
checkout.

## Changes

- **Defect A** — `git mv` six docs (`daemon-reference.md`, `troubleshooting.md`,
  `build-gate.md`, plus `github-authentication.md`, `forge-authentication.md`,
  `tool-use-concurrency-errors.md` — the last three surfaced only once the new
  link checker below was run against the first two, which reference them
  transitively) from `.loom/docs/` into `defaults/docs/` as real files;
  `.loom/docs/X` becomes a relative symlink back to `../../defaults/docs/X`
  (Option A from the issue — keeps `defaults/` self-contained for the
  bundled-resource layout, avoids duplicating a 150KB file).
- **Defect B** — add `defaults/roles/{probe-protocol,judge-reference}.md`
  symlinks, matching the existing `defaults/roles/*.md ->
  ../.claude/commands/loom/*` convention exactly.
- **Defect C** — `.claude/agents/loom-shepherd.md` is not currently shipped (a
  pre-#3884 install artifact, not a packaging gap). Add its 5 historical
  content digests to the `RETIRED_FILES` content-gated allowlist
  (`loom-daemon/src/init/retired.rs` + `scripts/install-loom.sh`'s
  `LOOM_RETIRED_FILES` heredoc, kept in sync by the existing drift-guard test)
  so it's swept on upgrade when unmodified, preserved when a consumer
  customized it.
- Absolutise the `docs/migration/`, `docs/research/`,
  `docs/autonomous-mode-e2e.md`, and `docs/model-selection-retune.md`
  references in shipped files (`defaults/CLAUDE.md`, `defaults/roles/README.md`,
  `defaults/scripts/experiments/README.md`) to GitHub URLs into the upstream
  Loom repo — those docs live only in this repo's own dev tree.
- Fix a wrong `../` depth in `sweep.md`'s `daemon-reference.md` anchor link.
- New `verify-install.sh check-links` subcommand (AC4): resolves every
  intra-repo link — both `[text](target)` markdown links and backtick-only
[...]

PR #4150, opened 2026-07-28
feat(daemon): classify why loom-daemon status can't reach the daemon

### LINKED ISSUE #4069: fix(daemon): `loom-daemon status` cannot distinguish "not installed" from "not running" from "running but unresponsive"

rjwalters:
## Summary

`loom-daemon status`'s error path (`handle_status_command`'s `Err` arm in
`main.rs`) collapsed every transport failure into one generic message —
*"Could not reach loom-daemon... Is the daemon running?"* — regardless of
whether autonomy was ever expected. That is wrong in exactly the case that
matters most: a daemon `bootout`-ed out of launchd (#4011's incident) looks
byte-identical to "never installed" or "deliberately stopped", and the
suggested remediation (`loom-daemon-start.sh`) is actively wrong when the
process is actually alive but wedged (the singleton guard refuses a second
start).

This adds a new library module, `loom-daemon/src/daemon_install_state.rs`
(modeled on `self_update.rs`), that classifies the unreachable-daemon case
into three states by reading the same **autonomy-desired marker** and
**heartbeat file** `loom-daemon-watchdog.sh` (#4011) already consumes,
mirroring its exact precedence (marker-absent, env-wins-over-marker,
liveness-then-heartbeat, and all its degradation rules) so `status` and the
watchdog log can never contradict each other:

| State | Meaning | Exit code | Remediation |
|---|---|---|---|
| `not-expected` | marker absent | `1` | suggest `loom-daemon-start.sh` |
| `expected-but-dead` | marker present, no live process | `3` | suggest `loom-daemon-start.sh`; points at `daemon-watchdog.log` |
| `alive-but-unresponsive` | marker present, process alive, IPC failed | `4` | does **not** suggest a start; prints the live pid; heartbeat freshness qualifies fresh ⇒ IPC/socket fault vs stale ⇒ likely wedged |

Both the human and `--json` renderings are covered — JSON gains a structured
`install_state` object (machine-readable `state` enum, `started_at`, `pid`,
`liveness_detail`, a `heartbeat` sub-object, `watchdog_log`), while the
existing `error` prose key is retained for compatibility.

The probe is read-only and never fails the command: an unreadable/malformed
marker, missing `launchctl`, a stale/unowned pid, or an unreadable heartbeat
mtime all degrade to a less-specific verdict; if no loom dir can be resolved
at all, `install_state` is simply omitted and the pre-existing generic
message is printed. The success path (`query_daemon_status`, `print_status_json`,
`print_status_human`) is untouched — this only runs after the status query
has already failed.

## Changes

- **New** `loom-daemon/src/daemon_install_state.rs` — marker parsing (with
  the watchdog's per-field fallbacks), launchctl/pid liveness, heartbeat
  freshness, the `InstallState` enum + `EXIT_*` constants, and 16 inline unit
  tests over tempdir fixtures (marker absent/present, live/dead/unowned pid,
  fresh/stale/absent heartbeat, malformed marker, missing fields, env
  overrides, non-Darwin forcing the pid-file path).
- `loom-daemon/src/lib.rs` — registers `pub mod daemon_install_state;`
  (alphabetically after `daemon_heartbeat`).
- `loom-daemon/src/main.rs` — imports the module; the `Err` arm of
  `handle_status_command` now calls `daemon_install_state::probe()` and
  renders via two new state-aware helpers (`print_status_unreachable_json`,
  `print_status_unreachable_human`) instead of the old single generic
  message; exit code comes from the classified state.
- `.loom/docs/daemon-reference.md` — extends the existing *Autonomy-loss
  watchdog + heartbeat (#4011)* section with a "Third consumer: `loom-daemon
  status`" subsection documenting the states, exit codes, and JSON shape.

## Test plan

- [x] `cargo test --lib` — 1307 tests pass, including the 16 new
  `daemon_install_state` unit tests (marker absent; live/dead/unowned pid;
  fresh/stale/absent/unreadable heartbeat; malformed marker; missing marker
  fields with fallback; env-wins-over-marker; non-Darwin forcing the
  pid-file path).
- [x] `cargo test --bin loom-daemon` — 21 tests pass (unaffected).
- [x] `cargo clippy --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] Manual end-to-end check against the built binary with
  `LOOM_SOCKET_PATH` pointed at a tempdir + synthetic marker/pid-file/heartbeat
  fixtures for all three states, both `status` and `status --json` — output
  matches the design exactly (see PR discussion for transcript).
- `loom-daemon/tests/integration_basic.rs` fails locally with "Daemon failed
  to create socket within 5s" — this is a documented, pre-existing sandbox
  limitation unrelated to this change (`.loom/docs/build-gate.md` — the
  local gate deliberately excludes integration test targets for exactly this
  reason; CI's `cargo test --workspace` runs them with a guaranteed-live
  environment).

Closes #4069

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2107,6 +2107,34 @@ tempdir isolates the marker + heartbeat there too — which is how the lifecycle
 tests avoid ever touching the operator's real `~/.loom`. A forge-side reporting
 channel is an explicit follow-up (out of scope for #4011).
 
+**Third consumer: `loom-daemon status` (#4069).** The watchdog's marker +
+heartbeat inputs have a second, on-demand consumer: `status`'s
+unreachable-daemon error path (`main.rs`'s `handle_status_command`, via the
+library module `daemon_install_state.rs`). Before #4069 every transport
+failure collapsed into one generic *"Could not reach loom-daemon… Is the
+daemon running?"* message, regardless of whether autonomy was ever expected.
+`status` now runs the **same read-only, local probe** — same marker, same
+per-field fallbacks, same env-wins-over-marker rule, same heartbeat staleness
+formula — so `status` and `<loom_dir>/logs/daemon-watchdog.log` can never
+disagree, and reports one of three states with a distinct exit code:
+
+| State | Meaning | Exit code | Remediation printed |
+|-------|---------|-----------|----------------------|
+| `not-expected` | marker absent (deliberate stop, or never started) | `1` | suggest `loom-daemon-start.sh` |
+| `expected-but-dead` | marker present, no live process — the #4011 divergence | `3` | suggest `loom-daemon-start.sh`; points at `daemon-watchdog.log` |
+| `alive-but-unresponsive` | marker present, process alive, IPC still failed | `4` | does **not** suggest a start (singleton guard refuses); prints the live pid; heartbeat freshness qualifies fresh ⇒ IPC/socket fault, stale ⇒ likely wedged |
+
+`--json` gains a structured `install_state` object (`state` is the
+machine-readable enum above, plus `started_at`, `pid`, `liveness_detail`, a
+`heartbeat` sub-object, and `watchdog_log`); the pre-#4069 `error` string key
+is retained for compatibility. The probe never fails the command — an
+unreadable/malformed marker, absent `launchctl`, a stale/unowned pid, or an
+unreadable heartbeat mtime all degrade to a less-specific verdict (or, if no
+loom dir can be resolved at all, `install_state` is simply omitted and the
+generic pre-#4069 message is printed). The success path (a reachable, healthy
+daemon) is untouched — this probe only runs after `query_daemon_status` has
+already failed.
+
 ### macOS session-bootstrap hazard (#3972)
 
 **Incident (2026-07-26).** The daemon was started at 21:48 via
--- a/loom-daemon/src/daemon_install_state.rs
+++ b/loom-daemon/src/daemon_install_state.rs
@@ -1,0 +1,828 @@
+//! Install-state classification for `loom-daemon status`'s unreachable-daemon
+//! error path (Issue #4069, AC3 of #4011).
+//!
+//! # Why
+//!
+//! Before this module, every transport failure in `handle_status_command`'s
+//! `Err` arm collapsed into one generic message ("Could not reach
+//! loom-daemon... Is the daemon running?"). That message is actively
+//! misleading in two of three real situations: it says nothing about whether
+//! autonomy was ever *expected* on this host (the #4011 silent-autonomy-loss
+//! scenario), and it recommends a start even when the daemon process is alive
+//! and the singleton guard will refuse.
+//!
+//! This module is the READ-ONLY probe that tells those situations apart, by
+//! consuming the same **autonomy-desired marker** and **heartbeat file** the
+//! host-side `loom-daemon-watchdog.sh` (#4011) already uses — and mirroring
+//! its exact precedence so `status` and the watchdog log can never
+//! contradict each other:
+//!
+//! 1. Marker absent → [`InstallState::NotExpected`]: no daemon is expected
+//!    (deliberately stopped, or never started).
+//! 2. Marker present, no live process → [`InstallState::ExpectedButDead`]:
+//!    the #4011 divergence — autonomy was expected but the daemon is gone.
+//! 3. Marker present, process alive (but IPC still failed, since we only
+//!    reach this module from the status query's `Err` arm) →
+//!    [`InstallState::AliveButUnresponsive`], qualified by heartbeat
+//!    freshness: fresh ⇒ likely an IPC/socket-layer fault, stale ⇒ likely a
+//!    wedged daemon.
+//!
+//! Like [`crate::self_update`], this module is modeled to be inherently
+//! side-effect-free and never fails the command: an unreadable/malformed
+//! marker, a missing `launchctl`, a stale/unowned pid, or an unreadable
+//! heartbeat mtime all degrade to a less-specific (but never wrong) verdict
+//! rather than propagating an error. The caller (`main.rs`) always has a
+//! fallback generic message to print if [`probe`] returns `None`.
+//!
+//! This module changes reporting only: it starts nothing, stops nothing, and
+//! writes no state (Non-goal in #4069 — do not add writes here).
+
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+/// Default macOS launchd label, matching `loom-daemon-start.sh` /
+/// `loom-daemon-watchdog.sh`'s fallback when the marker predates the
+/// `launchd_label` field.
+pub const DEFAULT_LAUNCHD_LABEL: &str = "com.rjwalters.loom-daemon";
+
+/// Basename of the autonomy-desired intent marker under the resolved loom
+/// dir, matching `loom-daemon-start.sh`'s `INTENT_MARKER` default.
+pub const MARKER_FILENAME: &str = "autonomy-desired";
+
+/// Exit code for `not-expected` (and any undiagnosable fallback) — unchanged
+/// from the pre-#4069 generic behavior, so existing scripting that only
+/// checks non-zero is unaffected.
+pub const EXIT_NOT_EXPECTED: i32 = 1;
+/// Exit code for `expected-but-dead` (the #4011 silent-autonomy-loss state).
+pub const EXIT_EXPECTED_BUT_DEAD: i32 = 3;
+/// Exit code for `alive-but-unresponsive` (IPC fault or wedged daemon).
+pub const EXIT_ALIVE_BUT_UNRESPONSIVE: i32 = 4;
+
+// ============================================================================
+// Public types
+// ============================================================================
+
+/// The three states `status` can distinguish once a live IPC round-trip has
+/// failed. See module docs for the precedence this mirrors from
+/// `loom-daemon-watchdog.sh`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InstallState {
+    /// No autonomy-desired marker — a daemon is not currently expected
+    /// (deliberately stopped via `loom-daemon-stop.sh`, or never started).
+    NotExpected,
+    /// Marker present, but no live process for it — the #4011 divergence.
+    ExpectedButDead,
+    /// Marker present and the process is alive, but IPC still failed.
+    AliveButUnresponsive,
+}
+
+impl InstallState {
+    /// Machine-readable enum value for `--json` rendering.
+    #[must_use]
+    pub fn as_str(self) -> &'static str {
+        match self {
+            InstallState::NotExpected => "not-expected",
+            InstallState::ExpectedButDead => "expected-but-dead",
+            InstallState::AliveButUnresponsive => "alive-but-unresponsive",
+        }
+    }
+
+    /// The exit code `loom-daemon status` should use for this state.
+    #[must_use]
+    pub fn exit_code(self) -> i32 {
+        match self {
+            InstallState::NotExpected => EXIT_NOT_EXPECTED,
+            InstallState::ExpectedButDead => EXIT_EXPECTED_BUT_DEAD,
+            InstallState::AliveButUnresponsive => EXIT_ALIVE_BUT_UNRESPONSIVE,
+        }
+    }
+}
+
+/// Heartbeat freshness qualifier for [`InstallState::AliveButUnresponsive`] —
+/// sharpens "alive but not answering" into "likely an IPC fault" (fresh) vs
+/// "likely wedged" (stale). `Unknown` is a degradation (no heartbeat file,
+/// unreadable mtime, or heartbeat loop disabled) — never a false report.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum HeartbeatFreshness {
+    Fresh,
+    Stale,
+    Unknown,
+}
+
+impl HeartbeatFreshness {
+    #[must_use]
+    pub fn as_str(self) -> &'static str {
+        match self {
+            HeartbeatFreshness::Fresh => "fresh",
+            HeartbeatFreshness::Stale => "stale",
+            HeartbeatFreshness::Unknown => "unknown",
+        }
+    }
+}
+
+/// The full classification result for one probe.
+#[derive(Debug, Clone)]
+pub struct InstallStateReport {
+    pub state: InstallState,
+    /// The marker's `started_at` field, when present and the marker exists.
+    pub started_at: Option<String>,
+    /// The live pid, when [`InstallState::AliveButUnresponsive`].
+    pub pid: Option<u32>,
+    /// Human-readable liveness detail (mirrors the watchdog's own
+    /// `liveness_detail` strings so log/status messages read consistently).
+    pub liveness_detail: Option<String>,
+    /// Heartbeat freshness — only computed for `AliveButUnresponsive`.
+    pub heartbeat_freshness: Option<HeartbeatFreshness>,
+    /// Heartbeat age in seconds, when its mtime was readable.
+    pub heartbeat_age_secs: Option<u64>,
+    /// The staleness threshold used for the freshness verdict (seconds).
+    pub heartbeat_stale_threshold_secs: Option<u64>,
+    /// The watchdog log path to point an operator at (advisory only — never
+    /// read by this module).
+    pub watchdog_log_path: PathBuf,
+}
+
+impl InstallStateReport {
+    fn not_expected(watchdog_log_path: PathBuf) -> Self {
+        InstallStateReport {
+            state: InstallState::NotExpected,
+            started_at: None,
+            pid: None,
+            liveness_detail: None,
+            heartbeat_freshness: None,
+            heartbeat_age_secs: None,
+            heartbeat_stale_threshold_secs: None,
+            watchdog_log_path,
+        }
+    }
+}
+
+/// Env/platform overrides that win over the marker's recorded fields —
+/// mirrors `loom-daemon-watchdog.sh`'s "env wins over marker" rule exactly.
+/// Split out from [`probe`] so unit tests can construct one directly without
+/// mutating process-global env vars.
+#[derive(Debug, Clone)]
+pub struct EnvOverrides {
+    /// From `LOOM_DAEMON_LAUNCHD`: this override is **one-directional**, exactly
+    /// mirroring `loom-daemon-watchdog.sh` — a falsy value (`0`/`false`/`no`)
+    /// yields `Some(false)`, forcing the pid-file path even when the marker says
+    /// `use_launchd=true`. Any other value (or unset) yields `None`, deferring to
+    /// the marker's own `use_launchd`; the env var can never force launchd *on*.
+    pub launchd_override: Option<bool>,
+    /// From `LOOM_LAUNCHD_LABEL`.
+    pub launchd_label_override: Option<String>,
+    /// From `LOOM_DAEMON_HEARTBEAT_STALE_SECS`.
+    pub heartbeat_stale_secs_override: Option<u64>,
+    /// Whether this host can have a launchd job at all — the watchdog forces
+    /// `USE_LAUNCHD=false` on non-Darwin regardless of the marker.
+    pub is_darwin: bool,
+}
+
+/// Parse the `LOOM_DAEMON_LAUNCHD` env value into a launchd override.
+///
+/// **One-directional, mirroring `loom-daemon-watchdog.sh` exactly**: a falsy
+/// value (`0`/`false`/`no`, case-insensitive) yields `Some(false)`, forcing the
+/// pid-file path even when the marker recorded `use_launchd=true`. Any other
+/// value — including a truthy `1`/`true`/`yes` — and the unset case yield
+/// `None`, deferring to the marker's own `use_launchd`. The env var can never
+/// force launchd *on*, which is the whole point of matching the shell's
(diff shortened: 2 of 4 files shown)


rjwalters:

**Verified independently (not just trusting the report):**
- `cargo test --lib`: 1307 passed, including all 16 new `daemon_install_state` unit tests. ✅
- `cargo clippy --lib --bin loom-daemon -- -D warnings`: clean. ✅
- `cargo fmt --check`: clean. ✅
- Manual end-to-end against the built binary with `LOOM_SOCKET_PATH` pointed at synthetic tempdir fixtures — all three states reproduced exactly as designed:
  - No marker → `not-expected`, exit `1`.
  - Marker + dead pid → `expected-but-dead`, exit `3`, cites `daemon-watchdog.log`.
  - Marker + live pid + fresh heartbeat → `alive-but-unresponsive`, exit `4`, "likely an IPC/socket-layer fault"; same with a stale heartbeat → "likely wedged", no start suggestion, live pid printed instead. `--json` renders the structured `install_state` object correctly in every case, with `error` retained for compatibility.
- Classification logic (`classify()` in `daemon_install_state.rs`) mirrors `loom-daemon-watchdog.sh`'s precedence closely: marker-absent, per-field fallbacks, liveness-then-heartbeat, `max(interval*5, 300)` staleness formula with env override, and all the "degrade rather than false-report" rules (unreadable mtime, absent heartbeat, malformed marker, unowned/stale pid). Success path (`query_daemon_status`, `print_status_json`, `print_status_human`) is untouched — confirmed by diff and by the full test suite passing unchanged.

**Blocking issue — real merge conflict against current `main` (`mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`):**

`main` moved (via #4145, commit `312b6b38`) to ship `.loom/docs/*` as **symlinks** into `defaults/docs/*` (so `.loom/docs/daemon-reference.md` is now a `120000` symlink to `../../defaults/docs/daemon-reference.md`), landing *after* this branch forked. This PR still edits the old regular-file path (`.loom/docs/daemon-reference.md`, `100644`), which now produces a hard "distinct types" conflict on merge — I reproduced this locally with a plain `git merge` of `origin/main` into the PR branch, confirming it's not a stale-badge false alarm.

**Fix needed:** rebase/merge `main` into this branch and move the new *"Third consumer"* doc section into **`defaults/docs/daemon-reference.md`** instead (the symlink target — see the project's own `.loom/docs/daemon-reference.md` structure post-#4145). The issue body's *"Affected Files"* list predates that migration; the correct target is the `defaults/` copy.

**Minor (non-blocking, worth a follow-up or quick fix while rebasing):** `EnvOverrides::from_env()`'s `LOOM_DAEMON_LAUNCHD` handling doesn't fully match the watchdog's env-wins-over-marker rule. The shell script's regex only ever *forces false* (`^(0|false|no)$`); any other value (or unset) leaves `USE_LAUNCHD` at the marker's own value. This Rust version instead treats *any* non-falsy value (e.g. `LOOM_DAEMON_LAUNCHD=1`) as `Some(true)`, which would force `use_launchd` on even when the marker explicitly recorded `use_launchd=false` (e.g. a `--no-launchd` start) — a case the shell script would leave alone. Narrow edge case (nothing in-repo sets this env var to a truthy value today) but worth a one-line fix (`.filter()` to only ever produce `Some(false)`, matching the shell's one-directional override) since the AC explicitly calls out exact precedence matching.

Once rebased onto current `main` (with the doc addition moved to `defaults/docs/`) and CI is green, this is good to merge — the core logic, tests, and manual verification are solid.

rjwalters:
## Doctor — feedback addressed, re-requesting review

Both items from the Judge review are resolved; rebased onto current `main` and re-verified locally.

**1. Merge conflict (BLOCKING) — fixed.** Rebased `feature/issue-4069` onto current `main` (past #4145 / `312b6b38`, which turned `.loom/docs/*` into symlinks into `defaults/docs/*`). The doc addition — the *"Third consumer: `loom-daemon status` (#4069)"* subsection extending the existing `#4011` watchdog section — now lives in **`defaults/docs/daemon-reference.md`** (the real source file the `.loom/docs/daemon-reference.md` symlink points to). The `.loom/docs/daemon-reference.md` symlink from `main` is left untouched, so the "distinct types" conflict is gone (`mergeable: MERGEABLE`, previously `CONFLICTING`).

**2. `LOOM_DAEMON_LAUNCHD` env override (minor) — fixed.** `EnvOverrides::from_env()` previously treated any non-falsy value as `Some(true)`, which could force `use_launchd` *on* against a marker that recorded `use_launchd=false`. The reference `loom-daemon-watchdog.sh` (lines ~141-148) only ever forces `false` via `^(0|false|no)$` and never forces `true`. The parse logic is now one-directional — extracted into a pure `parse_launchd_override()` helper: a falsy value yields `Some(false)`, and any truthy/unrecognized/unset value yields `None` (defer to the marker). Added unit test `launchd_override_is_one_directional_only_forces_false` asserting `0`/`false`/`no` (case-insensitive) → `Some(false)` and `1`/`true`/`yes`/`""`/unset → `None`.

**Local verification (all clean):**
- `cargo test --lib`: 1319 passed (incl. all 17 `daemon_install_state` tests, +1 new).
- `cargo test --bin loom-daemon`: 21 passed.
- `cargo fmt --check`: clean.

The core classification logic, exit codes, and `--json` shape the Judge already verified are unchanged.

rjwalters:
✅ **Approved!** Both items from the previous review are independently re-verified as fixed.

**1. Merge conflict — fixed.** Rebased past #4145/`312b6b38`. Confirmed independently:
- `.loom/docs/daemon-reference.md` on this branch is still the `120000` symlink to `../../defaults/docs/daemon-reference.md` (untouched, `git diff 312b6b38..HEAD` on that path is empty).
- The new *"Third consumer: `loom-daemon status` (#4069)"* subsection now lives in `defaults/docs/daemon-reference.md` (verified via diff — landed correctly, right after the existing #4011 watchdog section).
- `git merge-tree --write-tree origin/main HEAD` produces a clean tree with no conflict markers.
- `gh pr view --json mergeable,mergeStateStatus` → `MERGEABLE` / `CLEAN` (after CI settled).

**2. `LOOM_DAEMON_LAUNCHD` override — fixed.** `parse_launchd_override()` in `loom-daemon/src/daemon_install_state.rs` now filters to only the falsy set (`0|false|no`, case-insensitive) → `Some(false)`, everything else (including unset) → `None`, deferring to the marker. This is a byte-exact match to the watchdog's `^(0|false|no)$`-only regex (`defaults/scripts/cli/loom-daemon-watchdog.sh:142-144`), which never forces `USE_LAUNCHD` true. Confirmed the call site (`classify()`, line 487: `env.launchd_override.unwrap_or(fields.use_launchd) && env.is_darwin`) wires this through correctly. The new unit test `launchd_override_is_one_directional_only_forces_false` covers both directions plus case-insensitivity.

**Full re-verification (all clean, run independently from the worktree):**
- `cargo test --lib`: 1319 passed (0 failed).
- `cargo test --bin loom-daemon`: 21 passed.
- `cargo fmt --check`: clean.
- `gh pr checks 4150`: all required checks pass (Installer Integration Tests, Rust Linting & Build Check, Rust Unit Tests, CLAUDE.md Line Budget, Detect Changes; MCP Server Build correctly skipped — no MCP changes).

No regressions found — the success path (`query_daemon_status`, `print_status_json`, `print_status_human`) remains untouched, and all three classified states (`not-expected`/`expected-but-dead`/`alive-but-unresponsive`) plus their exit codes and `--json` shapes match the issue #4069 acceptance criteria as originally verified.

PR #4151, opened 2026-07-28
fix(defaults): Write(/tmp/**) permission rule is invalid — use Edit(...) and cover /private/tmp

### LINKED ISSUE #4072: fix(defaults): Write(/tmp/**) permission rule is invalid — use Edit(...) and cover /private/tmp
### Task Type

Bug Fix

### Task Description

The shipped default permission list contains `Write(/tmp/**)`, which Claude Code rejects at startup. Every project that installs Loom's defaults inherits the bad rule and prints a warning on every launch:

```
Permission allow rule (.claude/settings.json): Write(/tmp/**) is not matched by file
permission checks — only Edit(path) rules are. Use Edit(/tmp/**) instead
(Edit rules cover all file-editing tools).
```

Because file-permission checks only match `Edit(...)` rules, the rule is also a no-op — temp-file writes still trigger permission prompts, which is the opposite of what the rule was added for (#622, added in `b1839b89`).

**Steps to reproduce:**
1. Install Loom into any project (`loom-daemon init` copies `defaults/.claude/settings.json` → `<project>/.claude/settings.json`)
2. Launch `claude` in that project
3. Warning is printed at startup

**Expected:** No warning; temp-file writes are allowed without prompting.
**Actual:** Warning on every launch, and the rule has no effect.

---

Verified against `origin/main` @ `52e71c3e`. Three things in the original text needed correction, and the "propagation caveat" now has a definitive answer.

### 1. `.claude/settings.json` is already half-fixed — the original file reference is stale

The original body listed two files carrying the bad rule. Only **one** does:

```
$ git grep -n 'Write(' -- '*settings*.json'
defaults/.claude/settings.json:95:      "Write(/tmp/**)"
```

`.claude/settings.json` was fixed by commit **`c4b6f677`** (*"chore: replace deprecated Write() permission rule with Edit() in .claude settings"*) and reads `Edit(/tmp/**)` at `:100` on `origin/main` today.

That commit fixed Loom's own project settings but **not the template it is generated from** — the classic self-install dogfooding split. The propagation source was left intact, so every downstream install still gets the bad rule.

**Neither file covers `/private/tmp/**`.** That half of the issue is still valid for both.

### 2. The real blocker: fixing the template does NOT repair existing installs

This is the part the original body flagged as an open question ("*Worth considering whether the installer/update path should rewrite a stale entry in place*"). The answer is **no, and it cannot today** — by deliberate design.

`merge_permissions` (`loom-daemon/src/init/scaffolding.rs:356-385`) is a **pure union of `permissions.allow`**:

```rust
for perm in loom_allow {
    if let Some(perm_str) = perm.as_str() {
        if !existing_set.contains(perm_str) {
            existing_arr.push(perm.clone());   // append-only; nothing is ever removed
        }
    }
}
```

Consequences for an already-installed project that re-runs install:
- `Edit(/tmp/**)` is **appended** ✅
- `Write(/tmp/**)` **remains forever** ❌ → **the startup warning persists**

This union behavior is intentional, from **#3177** (*Non-destructive settings.json merging during install*, closed 2026-04-19): the installer previously clobbered consumer hooks/permissions, and union-only was the fix. Removal cannot be added casually — the merge has no way to distinguish "Loom shipped this and has since deprecated it" from "the consumer added this themselves."

There is also **no lighter-weight refresh path**: `.loom/scripts/resync-installed.sh` walks only `defaults/hooks/` (`:223-231`) and `defaults/scripts/` (`:235-241`). It does not touch `.claude/` at all.

**Scoping decision:** keep this issue to the template fix. It is a 2-line change with an unambiguously correct outcome for fresh installs. Building a deprecation/removal mechanism into `merge_permissions` is a separate design question (needs an explicit Loom-owned-deprecated-rules list, plus a blast-radius argument against #3177's rationale) and should be its own issue — see *Follow-up* below.

The original acceptance criterion *"Launching `claude` in a freshly installed project prints no permission-rule warning"* is therefore correct **only** for fresh installs, and is qualified as such below.

### 3. Nothing prevents this from regressing

`scripts/test-installer.sh` has **zero** assertions on the contents of `permissions.allow` — `grep -n "permissions" scripts/test-installer.sh` returns no matches. Its only settings.json coverage is file existence (`:334-340`) and hook-command path prefixes (`:342-376`). The same is true of `defaults/scripts/verify-install.sh`, which checks existence only (`:250-252`).

A one-line grep guard closes this permanently and is cheap enough to belong in this PR.

### Verified environment claims

- `/tmp` is a symlink to `private/tmp` on macOS — confirmed (`lrwxr-xr-x /tmp -> private/tmp`).
- Session scratchpad paths do resolve under `/private/tmp` — confirmed (`/private/tmp/claude-501` exists for uid 501).
- So a `/tmp/**` rule alone genuinely does still prompt on macOS. The `/private/tmp/**` entries are inert on Linux.
- `defaults/.claude/settings.json` has only a `permissions.allow` key (no `deny`/`ask`), 60 entries, of which exactly two are path-shaped: `Read(/tmp/**)` and `Write(/tmp/**)`. `Read(...)` is a valid rule form and is not affected by this warning — leave it.

## Implementation Guidance

Replace the two path-shaped rules in **`defaults/.claude/settings.json`** (`:94-95`) with four:

```json
"Read(/tmp/**)",
"Read(/private/tmp/**)",
"Edit(/tmp/**)",
"Edit(/private/tmp/**)"
```

Apply the same four-entry block to **`.claude/settings.json`** (`:99-100`) — it already has `Read(/tmp/**)` + `Edit(/tmp/**)`, so this only adds the two `/private/tmp` entries. Keeping the two files identical here is the point: the drift between them is what caused this bug to survive `c4b6f677`.

`Write(...)` → `Edit(...)` clears the warning; `Edit` rules cover all file-editing tools (Write, Edit, NotebookEdit).

**Do not** modify `merge_permissions` or any other `loom-daemon/src/init/` code in this PR — see the scoping decision above.

## Affected Files

- `defaults/.claude/settings.json` — **the actual fix.** Replace `Write(/tmp/**)` at `:95` with the four-entry block above. This is the propagation source.
- `.claude/settings.json` — add the two `/private/tmp/**` entries at `:99-100`. Already uses `Edit(...)`; no `Write(...)` to remove.
- `scripts/test-installer.sh` — add a regression guard asserting no `Write(...)`-style path rule appears in the installed `settings.json` (see Test Plan). Place it alongside the existing Test 12/12b settings block (`:334-376`).

**Read-only reference (do not edit):** `loom-daemon/src/init/scaffolding.rs` (`merge_permissions`, `:356-385`), `.loom/scripts/resync-installed.sh`.

## Acceptance Criteria

- [ ] `defaults/.claude/settings.json` uses `Edit(...)` instead of `Write(...)` for temp paths
- [ ] Both `defaults/.claude/settings.json` and `.claude/settings.json` cover `/private/tmp/**` for both `Read` and `Edit`, and the two files' temp-path rule blocks are identical
- [ ] No `Write(...)` path rules remain in tracked settings — `git grep 'Write(' -- '*settings*.json'` returns nothing
- [ ] Launching `claude` in a **freshly installed** project prints no permission-rule warning
- [ ] A regression guard in `scripts/test-installer.sh` fails if a `Write(...)` path rule is ever reintroduced into the installed settings
- [ ] The PR body states the known limitation: existing installs retain the stale `Write(/tmp/**)` entry because `merge_permissions` is union-only (`scaffolding.rs:356-385`), and must remove it by hand or wait on the follow-up

## Test Plan

- [ ] **Manual verification (fresh install):** install Loom into a scratch repo with no pre-existing `.claude/settings.json`, launch `claude`, confirm no permission-rule warning is printed.
- [ ] **Manual verification (rule is effective):** in that repo, confirm a write to a `/private/tmp/claude-$(id -u)/...` path proceeds without a permission prompt — the rule should now actually do its job, not merely stop warning.
- [ ] **Automated test:** add to `scripts/test-installer.sh` near the existing Test 12 block (`:334`) — assert the installed `settings.json` contains no `Write(`-prefixed path rule:
  ```bash
  if jq -e '.permissions.allow[] | select(startswith("Write("))' "$SETTINGS_FILE" >/dev/null 2>&1; then
    fail "settings.json contains an invalid Write(...) path rule — use Edit(...)"
  else
    pass "no invalid Write(...) path rules in settings.json"
  fi
  ```
- [ ] **Automated test:** assert all four temp-path entries (`Read(/tmp/**)`, `Read(/private/tmp/**)`, `Edit(/tmp/**)`, `Edit(/private/tmp/**)`) are present in the installed settings.
- [ ] **Regression:** run `scripts/test-installer.sh` in full and confirm green — in particular that Test 12b (hook command prefixes, `:342-376`) is unaffected.
- [ ] **Edge case (upgrade path):** re-install over a project whose `.claude/settings.json` already contains `Write(/tmp/**)`. Confirm the documented behavior — `Edit(...)` is appended and `Write(...)` survives. This test **documents** the known limitation rather than asserting it away; it is the tripwire that will start failing when the follow-up lands.
- [ ] **Linux non-regression:** confirm the `/private/tmp/**` entries are inert (not an error) on a Linux install.

## Follow-up (not this issue)

`merge_permissions` cannot retire a permission rule Loom itself shipped and later deprecated, so template fixes never reach existing installs. Worth its own issue: a Loom-owned deprecated-rules list that install-time merge can prune, argued explicitly against #3177's non-destructive rationale. Filing deferred to Champion/human — flagging rather than creating so the design question gets an owner rather than a drive-by.

Refs #622 (original `/tmp` permission), #3177 (non-destructive merge rationale), `c4b6f677` (partial fix).

rjwalters:
## Summary

The shipped default permission list in `defaults/.claude/settings.json` contained an invalid `Write(/tmp/**)` rule. Claude Code's permission engine only matches `Edit(...)` rules for file-editing tools (Write/Edit/NotebookEdit all route through Edit checks), so the rule both printed a startup warning on every launch AND was a no-op — temp-file writes still triggered permission prompts.

## Changes

1. **`defaults/.claude/settings.json`** — replaced `Read(/tmp/**)` + `Write(/tmp/**)` with four entries: `Read(/tmp/**)`, `Read(/private/tmp/**)`, `Edit(/tmp/**)`, `Edit(/private/tmp/**)`. This is the propagation source every fresh install copies from.
2. **`.claude/settings.json`** (this repo's own project settings, already partially fixed by `c4b6f677`) — added the two missing `/private/tmp/**` entries so it now matches the template's block byte-for-byte.
3. **`scripts/test-installer.sh`** — added Test 12c, a regression guard: asserts the installed `settings.json` has no `Write(`-prefixed path rule, and asserts all four temp-path entries are present.

Verified `git grep 'Write(' -- '*settings*.json'` returns nothing after this change, and `scripts/test-installer.sh` passes in full (140/140, including the new Test 12c).

Out of scope, per the issue's curator-verified scoping decision: `merge_permissions` (`loom-daemon/src/init/scaffolding.rs:356-385`) and `.loom/scripts/resync-installed.sh` were **not** touched.

## Known limitation (flagged, not fixed here)

`merge_permissions` is a pure union of `permissions.allow` — it never removes a rule, only appends missing ones. So an **existing** installed project that re-runs install will get `Edit(/tmp/**)` appended but will retain the stale `Write(/tmp/**)` forever; the startup warning persists for those installs until they manually edit their `.claude/settings.json`. This is a deliberate, documented limitation of this fix — building a deprecated-rules pruning mechanism into `merge_permissions` is a separate design question. Filing that follow-up issue is deferred to Champion/human per the issue body, not done here.

## Test plan

- [x] `git grep 'Write(' -- '*settings*.json'` returns nothing
- [x] `scripts/test-installer.sh` passes in full (140 passed, 0 failed), including new Test 12c
- [x] Both `defaults/.claude/settings.json` and `.claude/settings.json` temp-path rule blocks are byte-for-byte identical (verified via diff)
- [x] `loom-daemon/src/init/scaffolding.rs` untouched (confirmed via `git diff --stat`)

Closes #4072

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -97,7 +97,9 @@
       "Bash(shuf:*)",
       "WebSearch",
       "Read(/tmp/**)",
-      "Edit(/tmp/**)"
+      "Read(/private/tmp/**)",
+      "Edit(/tmp/**)",
+      "Edit(/private/tmp/**)"
     ]
   }
 }
--- a/defaults/.claude/settings.json
+++ b/defaults/.claude/settings.json
@@ -92,7 +92,9 @@
       "Bash(shuf:*)",
       "WebSearch",
       "Read(/tmp/**)",
-      "Write(/tmp/**)"
+      "Read(/private/tmp/**)",
+      "Edit(/tmp/**)",
+      "Edit(/private/tmp/**)"
     ]
   }
 }
--- a/scripts/test-installer.sh
+++ b/scripts/test-installer.sh
@@ -375,6 +375,33 @@ else
   fail "Cannot verify hook command prefixes (settings.json or jq missing)"
 fi
 
+# Test 12c: No invalid Write(...) path rule in permissions.allow (issue #4072)
+# Claude Code's permission engine only matches Edit(...) rules for file-editing
+# tools (Write/Edit/NotebookEdit all route through Edit checks) -- a
+# Write(...) path rule prints a startup warning and is a silent no-op. Also
+# assert the four temp-path entries the fix introduces are all present.
+echo "Test 12c: settings.json has no invalid Write(...) path rule"
+if [[ -f "$SETTINGS_FILE" ]] && command -v jq &> /dev/null; then
+  if jq -e '.permissions.allow[] | select(startswith("Write("))' "$SETTINGS_FILE" > /dev/null 2>&1; then
+    fail "settings.json contains an invalid Write(...) path rule -- use Edit(...)"
+  else
+    pass "no invalid Write(...) path rules in settings.json"
+  fi
+
+  TEMP_PATH_FAIL=0
+  for entry in "Read(/tmp/**)" "Read(/private/tmp/**)" "Edit(/tmp/**)" "Edit(/private/tmp/**)"; do
+    if ! jq -e --arg entry "$entry" '.permissions.allow[] | select(. == $entry)' "$SETTINGS_FILE" > /dev/null 2>&1; then
+      fail "settings.json missing expected temp-path rule: $entry"
+      TEMP_PATH_FAIL=1
+    fi
+  done
+  if [[ $TEMP_PATH_FAIL -eq 0 ]]; then
+    pass "settings.json has all four temp-path permission entries"
+  fi
+else
+  fail "Cannot verify settings.json permissions (settings.json or jq missing)"
+fi
+
 # Test 13: .github/labels.yml
 echo "Test 13: Install creates .github/labels.yml"
 if [[ -f "$INSTALL_REPO/.github/labels.yml" ]]; then


rjwalters:
✅ **Approved!**

Verified against issue #4072's acceptance criteria:

- `defaults/.claude/settings.json`: `Write(/tmp/**)` replaced with `Edit(/tmp/**)`, plus `Read(/private/tmp/**)` and `Edit(/private/tmp/**)` added. This is the propagation source.
- `.claude/settings.json`: the two missing `/private/tmp/**` entries added; confirmed the temp-path rule blocks in both files are byte-for-byte identical (`jq` diff, no output).
- `git grep 'Write(' -- '*settings*.json'` returns nothing — no invalid `Write(...)` path rules remain anywhere in tracked settings.
- Regression guard added: new Test 12c in `scripts/test-installer.sh` fails on any `Write(...)`-prefixed `permissions.allow` entry and asserts all four temp-path entries are present.
- PR body explicitly documents the known limitation (existing installs retain stale `Write(/tmp/**)` because `merge_permissions` is union-only, `scaffolding.rs:356-385`), as required.
- Scope respected: diff touches only the 3 files listed in the issue (`.claude/settings.json`, `defaults/.claude/settings.json`, `scripts/test-installer.sh`). No changes to `merge_permissions` or anything under `loom-daemon/src/init/`.

## Test Execution

Ran `scripts/test-installer.sh` locally in the PR worktree (`.loom/worktrees/issue-4072`): **140 passed, 0 failed**, including the new Test 12c (both the no-`Write(...)` check and the four-entry presence check).

## CI / Merge State

- `gh pr checks 4151`: Installer Integration Tests pass, CLAUDE.md Line Budget pass, Detect Changes pass; Rust jobs correctly `skipping` (no Rust files touched).

Clean, minimal, exactly scoped fix. Nice work.

PR #4154, opened 2026-07-28
fix(scripts): delete the local branch on every merge-pr.sh cleanup path

### LINKED ISSUE #4100: merge-pr.sh deletes the remote branch and worktree but leaves the local branch behind
## Context

`merge-pr.sh` cleans up two of the three artifacts a merged PR leaves behind — the remote branch and the worktree — but not the local branch.

### Repro (observed this session, rjwalters/repo)

```
$ ./.loom/scripts/merge-pr.sh 34
Merging PR #34: feat(hooks): surface a pending /repo:handoff note via a SessionStart hook
Branch: feature/issue-32
PR #34 merged successfully
Deleting remote branch: feature/issue-32
Branch 'feature/issue-32' deleted
Removing worktree: /Users/rwalters/GitHub/repo/.loom/worktrees/issue-32
Worktree removed
Done
```

Note "Branch 'feature/issue-32' deleted" refers to the **remote** branch. Afterwards:

```
$ git branch -vv
  feature/issue-25 e9185b0 [origin/feature/issue-25: gone] ...
  feature/issue-32 4415d45 [origin/feature/issue-32: gone] ...
* main             ...
```

Both local branches survive, each tracking a now-dangling remote ref. Worktree removal and remote deletion were verified genuinely complete (`git worktree list` clean, `git ls-remote --heads origin feature/issue-32` empty) — only the local branch is left.

`feature/issue-25` is residue from PR #26, merged five days earlier, so this is systematic rather than a one-off.

### Why it is not merely cosmetic

The leftovers accumulate one per merged issue and all classify as `[gone]`, so they are pure noise in `git branch`. More importantly they interact badly with squash merges: the local branch's commits are never ancestors of `main`, so `git branch --merged main` will not list them, and a cleanup routine that trusts `--merged` will never reap them. They need a content-level or PR-state check to be safely removed.

## Acceptance Criteria

- [ ] After a successful `merge-pr.sh <N>`, the local branch for that PR is deleted when it is safe to do so.
- [ ] "Safe" is determined by the PR being merged and the branch holding no commits absent from `origin` — **not** by `git branch --merged`, which returns false for every squash-merged branch.
- [ ] A branch that fails the safety check is kept and reported, never force-deleted (`git branch -d`, not `-D`, unless the merged-PR check passed explicitly).
- [ ] The branch currently checked out is never deleted (fail gracefully if the merged branch is the current HEAD).
- [ ] Log lines distinguish remote-branch deletion from local-branch deletion — the current "Branch 'X' deleted" is ambiguous and reads as if the local branch were already handled.
- [ ] Consider a cleanup pass for the backlog of existing `[gone]` branches in repos that have been running Loom for a while, or document `git fetch --prune` + a manual sweep as the remedy.

---

## Curator Enhancement — consolidated (2026-07-27, verified against `origin/main` @ `5266454d`)

> **This section supersedes the two earlier curator comments on this issue.** They
> reached *different* conclusions on the Affected Files list, the wiring approach,
> and the documentation target. Where they disagreed, the resolution and the
> evidence are recorded below. **Read this body, not the comments.**

### Root cause (precise)

The local-branch delete **already exists**. `_maybe_delete_local_branch()` is defined at `defaults/scripts/merge-pr.sh:1368` and does exactly what this issue asks. It has exactly **one** call site — `merge-pr.sh:1455` — and that call site is gated behind `allow_unmanaged == "true"`, which is set **only** by the `--worktree-path` override path (`:1471`).

Every other path leaks the branch:

| Path | Line | Deletes local branch today? |
|------|------|------------------------------|
| `--worktree-path <dir>` (explicit override) | 1471 | ✅ yes |
| Default `.loom/worktrees/issue-N` / `pr-N` | 1488 | ❌ no |
| Discovered Loom-managed worktree at non-standard path | 1501 | ❌ no |
| No worktree at all (branch exists, worktree already gone) | 1509 | ❌ no — and no worktree to hang the delete off |

Every Loom sweep takes the **default** path, so every merged issue leaks its local branch. That matches the `feature/issue-25` + `feature/issue-32` residue in the repro.

**Consequence for the fix**: this is a *wiring* change, not new machinery — but it must be wired as a step that also covers the fourth row, which has no worktree to attach to. Do not bury the call inside `_remove_loom_worktree`.

### ⚠️ Placement is load-bearing — the two comments disagreed here

Earlier guidance said "wire it as a top-level step **after** the worktree-cleanup block". That is nearly right but under-specified, and taken literally it breaks an AC:

- The worktree-cleanup block opens at `merge-pr.sh:1462` with `if [[ "$CLEANUP_WORKTREE" == "true" ]]; then` and closes at `:1513`.
- Placing the branch delete **after `:1513`** makes it run even under `--no-cleanup-worktree` / `LOOM_PRESERVE_WORKTREE=1`, where the branch is still checked out in the preserved worktree. The delete then fails and emits a spurious warning.
- **Correct placement: inside the `:1462` block, after the `if/else` worktree dispatch resolves** — so all four rows of the table above are covered *and* the existing skip flags still gate it.

### The trap: `git branch -d` is not a safe criterion here (AC #2, expanded)

AC #2 forbids `git branch --merged`. The same objection applies to `git branch -d`, which the existing helper uses — and this is the failure mode that will make a naive fix look correct in a hand test and fail in production. The helper's own comment (`:1365-1367`) calls `-d` "the right safety net", which is true in general and wrong for this call path:

- `git branch -d X` succeeds when X is merged into **HEAD or its upstream**. Immediately after `merge-pr.sh` runs, `refs/remotes/origin/feature/issue-N` is usually still present locally (the script deletes the branch server-side via the forge API; it does not prune local remote-tracking refs) and points at the same SHA — so `-d` succeeds.
- Once any later `git fetch --prune` runs, the upstream ref is gone. `-d` falls back to "merged into HEAD", which is **false for every squash merge**, and refuses.

That is precisely the `[gone]` state in the repro. **Test the fix after a `git fetch --prune`, not immediately after the merge**, or it passes for the wrong reason.

### Recommended safety criterion (reuses data already in hand)

`merge-pr.sh` already parses the PR head SHA at `:675` (`jq -r '.head.sha // empty'`). That yields a criterion correct under squash merges, with no new merge-detection heuristic:

- local tip `== PR_JSON.head.sha` → every commit on the local branch was part of the merged PR → `git branch -D` is safe and satisfies AC #3's "unless the merged-PR check passed explicitly". Log a line stating the check passed.
- local tip `!= head.sha` → the branch carries commits never in the merged PR (unpushed work) → attempt `git branch -d`; on refusal **keep the branch and report it** with the literal `git branch -D <branch>` escape hatch. Never force here.

Suggested shape: extend `_maybe_delete_local_branch` to take an optional expected-tip SHA, preserving its current `-d`-and-warn behaviour when the SHA is absent (so the existing `:1455` caller is unaffected).

### Guards

- **Never the current HEAD or a branch checked out in another worktree** (AC #4). `git branch -d/-D` refuses on its own, so catch the failure rather than pre-checking — but detect this case and emit a specific message instead of the generic "may have unmerged commits" warning, which would be misleading.
- **Never the default branch** — cheap belt-and-suspenders.
- **Respect `--dry-run`.** `merge-pr.sh:1173-1177` prints the dry-run plan and exits before any mutation; add a `Would delete local branch '<X>'` line there so `--dry-run` stays an accurate preview.

### Log-line disambiguation (AC #5)

`merge-pr.sh:1296` — `success "Branch '$PR_BRANCH' deleted"` → `success "Remote branch '$PR_BRANCH' deleted"`. The helper's messages already say "Local branch", so this one-word change closes the AC.

### AC #6 is already delivered — scope it to documentation

The backlog remedy already exists. `clean_branches()` at `loom-tools/src/loom_tools/clean.py:716` (added in #3471) runs a two-pass local-branch cleanup:

1. **Generic stale-detection pass** — for every local branch except the default/current/other-worktree-checked-out ones, delete it when `refs/remotes/origin/<branch>` no longer exists. Pattern-agnostic; reaps exactly the `[gone]` branches from the repro.
2. **Issue-state pass** — `feature/issue-*` branches that still have a remote are probed against the forge and deleted when the issue is CLOSED.

**Do not build new backlog tooling.** Satisfy AC #6 with a short note pointing at `git fetch --prune` + `loom-clean`.

#### 🔴 Documentation target — the earlier comments conflicted; this is the resolution

One comment named `.loom/docs/troubleshooting.md`; the other named `defaults/CLAUDE.md` §"Merging PRs from worktrees" (`defaults/CLAUDE.md:1773`). **Use `.loom/docs/troubleshooting.md`. Do not touch `defaults/CLAUDE.md`.**

`defaults/CLAUDE.md` is scheduled for **deletion** by epic #4052 — #4142 (Phase 1) migrates its operating core, #4143 (Phase 2) extracts its reference tier to `defaults/docs/`, and #4144 (Phase 3) deletes the file outright. Adding content there would either be destroyed by Phase 3 or collide with the Phase 1/2 migration diffs. The target section already exists: `.loom/docs/troubleshooting.md` §"Cleaning Up Stale Worktrees and Branches" (`:47`).

### 🔴 Editing model — the earlier comments conflicted here too

One comment said the installed `.loom/scripts/merge-pr.sh` "is refreshed by `resync-installed.sh` / the installer, never edited directly." **That is wrong for this repo.** Verified: `.loom/scripts` is a **symlink** (git mode `120000`) pointing into `defaults/scripts/` — as is `.loom/roles`. `.loom/docs` is *not* a symlink; it is a tracked directory of real files.

Practical consequence for the Builder: edit `defaults/scripts/merge-pr.sh`; `.loom/scripts/merge-pr.sh` reflects the change immediately with no resync step. Edit `.loom/docs/troubleshooting.md` directly as a real file.

### Out of scope (file a follow-up, do not fix here)

`defaults/scripts/worktree.sh:473` (`worktree.sh remove <N>`) has the identical `git branch -d`-only limitation and will likewise refuse on a squash-merged branch post-prune. It is a different entry point with **no PR context available**, so it needs a different criterion. Keep it out of this PR; file a follow-up referencing this issue.

### Scope guard

Single script change plus one new test plus a short doc note. Expected diff: **< 120 LOC**. If the implementation grows past that, stop and re-scope — it almost certainly means new merge-detection machinery is being invented instead of using the already-parsed `head.sha`.

## Affected Files

All paths verified present on `origin/main` @ `5266454d`.

- `defaults/scripts/merge-pr.sh` — wire local-branch deletion into the worktree-cleanup block (`:1462-1513`) so the default (`:1488`), discovery (`:1501`) and no-worktree (`:1509`) paths are all covered; extend `_maybe_delete_local_branch` (`:1368`) with the optional `head.sha` tip check while preserving current behaviour for the existing `:1455` caller; add the current-HEAD / default-branch guards; add the `--dry-run` preview line (insert after `:1175`, before the `exit 0` at `:1176`); disambiguate the remote-delete log line (`:1296`). Also update the header comment block (**`:16-18`**) and the `--help` text (**`:84-86`**), which currently describe local-branch deletion as `--worktree-path`-only.
- `defaults/scripts/tests/test-merge-pr-local-branch-cleanup.sh` — **new**. Follow the existing harness style of `defaults/scripts/tests/test-merge-pr-worktree-path.sh` (self-contained bash, `set -euo pipefail`, pass/fail counters, static source assertions plus inline simulation of the decision tree). No live forge.
- `.loom/docs/troubleshooting.md` — one-paragraph backlog remedy (`git fetch --prune` + `loom-clean`) in §"Cleaning Up Stale Worktrees and Branches" (`:47`). **Not** `defaults/CLAUDE.md` — see the conflict resolution above.

**Read-only reference (do not modify):**
- `loom-tools/src/loom_tools/clean.py:716` — `clean_branches()`, the existing backlog remedy AC #6 should point at.
- `defaults/scripts/worktree.sh:473` — the same `-d` limitation, explicitly out of scope.

## Test Plan

- [ ] Manual verification (**the load-bearing one — prune first**): merge a real PR with `./.loom/scripts/merge-pr.sh <N>`, then `git fetch --prune`, then `git branch -vv`. The merged branch must be absent. Repeating the check *without* the prune step is not a valid test — see "The trap" above.
- [ ] Manual verification: `./.loom/scripts/merge-pr.sh <N> --dry-run` names the local branch it would delete and changes nothing.
- [ ] Manual verification: `./.loom/scripts/merge-pr.sh <N> --no-cleanup-worktree` leaves both the worktree and the local branch, with an explicit skip message (not a failure warning). Same for `LOOM_PRESERVE_WORKTREE=1`.
- [ ] Automated tests — add `defaults/scripts/tests/test-merge-pr-local-branch-cleanup.sh` covering: (a) the branch-delete step is reachable from the default path, not only under `allow_unmanaged`; (b) the `head.sha` equality criterion gates `-D` vs `-d`; (c) tip-mismatch keeps the branch and warns; (d) current-HEAD refusal is reported specifically; (e) `--dry-run` emits the preview line and deletes nothing; (f) `--no-cleanup-worktree` / `LOOM_PRESERVE_WORKTREE=1` skip the delete; (g) the remote-delete log line reads "Remote branch".
- [ ] Regression: `bash defaults/scripts/tests/test-merge-pr-help.sh` and `bash defaults/scripts/tests/test-merge-pr-worktree-path.sh` still pass — both assert on `--help` text and source structure this change touches.
- [ ] Edge cases: PR head branch with no local branch present (clean no-op info line, not a warning); local branch with unpushed commits ahead of `head.sha` (**kept** and reported); branch checked out in a second worktree (reported, never force-deleted).

rjwalters:
## Summary

- `merge-pr.sh` cleaned up the remote branch and the worktree on a merged PR but left the local `feature/issue-N` branch behind on every path except the explicit `--worktree-path` override, so every merged issue leaked one local branch — invisible to `git branch --merged` after a squash merge (the branch is never an ancestor of `main`), showing as `[gone]` once `git fetch --prune` runs.
- Wires local-branch deletion into the default `.loom/worktrees/issue-N` path, the discovered-Loom-managed-non-standard-path case, and the no-worktree-at-all case — the three rows the pre-existing `_maybe_delete_local_branch` helper never reached.
- Adds a head-SHA safety criterion so the delete is correct under squash merges: the local branch tip is compared to the merged PR's `head.sha` (already parsed by the script). A matching tip is safe to `git branch -D`; a mismatched tip (unpushed local work) falls back to `git branch -d`, which keeps the branch and reports it.

## Changes

- `defaults/scripts/merge-pr.sh`:
  - Parses `PR_HEAD_SHA` from the PR JSON and resolves `DEFAULT_BRANCH_NAME` via `lib/default-branch.sh`.
  - Extends `_maybe_delete_local_branch(branch, [expected_head_sha])`: never targets the repo's default branch; uses `-D` only when the local tip matches `expected_head_sha`, otherwise falls back to `-d`; distinguishes a "checked out elsewhere" refusal (current HEAD or another worktree) from a generic "not fully merged" refusal with a specific message; the pre-existing `--worktree-path` caller is unchanged (still calls with no SHA, so its behavior is byte-for-byte preserved).
  - Wires a single unified `_maybe_delete_local_branch "$PR_BRANCH" "$PR_HEAD_SHA"` call after the worktree-cleanup dispatch resolves, inside the `CLEANUP_WORKTREE` guard, so it covers the default path, the discovery path, and the no-worktree case without running under `--no-cleanup-worktree` / `LOOM_PRESERVE_WORKTREE=1`.
  - Disambiguates the remote-branch log line (`"Branch 'X' deleted"` → `"Remote branch 'X' deleted"`), and adds explicit skip messages naming the local branch for both `--no-cleanup-worktree` and `LOOM_PRESERVE_WORKTREE=1`.
  - Previews the local-branch delete under `--dry-run` without mutating anything.
  - Updates the header comment block and `--help` text to describe the new default (local branch deletion is no longer `--worktree-path`-only).
- `defaults/scripts/tests/test-merge-pr-local-branch-cleanup.sh` (new): static source assertions for the wiring/log-line/dry-run/skip-message surface, plus behavioral tests of the real `_maybe_delete_local_branch` function body (extracted verbatim, same pattern as `test-merge-pr-primary-worktree-guard.sh`) against real git repos — tip-match force-delete, tip-mismatch keep-and-warn, checked-out-elsewhere refusal (both "another worktree" and "current HEAD"), default-branch refusal, nonexistent-branch no-op, and the no-SHA fallback preserving pre-#4100 behavior.
- `.loom/docs/troubleshooting.md` (via its `defaults/docs/troubleshooting.md` symlink target) — one-paragraph backlog remedy (`git fetch --prune` + `loom-clean --force`) pointing at the existing `clean_branches()` two-pass cleanup in `loom_tools/clean.py`, so pre-#4100 `[gone]` branches are reaped without new tooling.

- [x] After a successful `merge-pr.sh <N>`, the local branch is deleted when safe — verified via `--dry-run` against a real open PR (`feature/issue-4046` correctly identified for deletion) and behavioral test (a).
- [x] Safety uses the head-SHA tip comparison, not `git branch --merged` — test (a)/(b)/(f) exercise the tip-match vs. tip-mismatch branches directly.
- [x] A branch failing the safety check is kept and reported, never force-deleted unless the merged-PR check passed — test (b) verifies the branch survives with the `git branch -D <branch>` escape hatch in the warning.
- [x] The branch currently checked out is never deleted — tests (c) and (c2) cover "checked out in another worktree" and "is the current HEAD".
- [x] Log lines distinguish remote vs. local branch deletion — "Branch ... deleted" → "Remote branch ... deleted" / "Local branch ... deleted"; static assertion + the old ambiguous line is asserted gone.
- [x] Backlog remedy — documented in `.loom/docs/troubleshooting.md` (`git fetch --prune` + `loom-clean --force`, pointing at the existing `clean_branches()`).

## Test Plan

- [x] `bash defaults/scripts/tests/test-merge-pr-local-branch-cleanup.sh` — 18/18 pass (new).
- [x] `bash defaults/scripts/tests/test-merge-pr-help.sh` — 15/15 pass (regression).
- [x] `bash defaults/scripts/tests/test-merge-pr-worktree-path.sh` — 21/21 pass (regression).
- [x] All other `defaults/scripts/tests/test-merge-pr-*.sh` files — pass (auto-merge-disabled-fallback 30/30, auto-reconcile 20/20, merge-ordering-guard 25/25, partial-increment 19/19, primary-worktree-guard 14/14, unstable-fallback 68/68, worktree-awk 11/11).
- [x] `./.loom/scripts/merge-pr.sh <open-PR> --dry-run` against a real open PR: correctly previews `Would delete local branch 'feature/issue-4046'` and mutates nothing.
- [x] `./.loom/scripts/merge-pr.sh <open-PR> --dry-run --no-cleanup-worktree`: correctly previews "would leave local worktree and local branch ... in place" and mutates nothing.
- [x] `shellcheck --severity=warning` clean on `merge-pr.sh` and the new test file; full-repo `find defaults/scripts -name '*.sh' | xargs shellcheck` (matches the CI job) exits 0.
- [x] `check-phantom-labels.sh` / `check-labels-drift.sh` — unaffected, still pass.
- [x] `./.loom/scripts/check-main-clean.sh` — main worktree unaffected.

Note: the load-bearing manual verification from the issue ("merge a real PR, `git fetch --prune`, `git branch -vv` shows the branch absent") requires merging a live PR through this exact script and is covered by the automated tip-match test (a), which exercises the identical code path against a synthetic squash-merge-shaped repo (branch tip == merged head SHA, no ancestor relationship required).

Closes #4100

### DIFF
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -77,6 +77,24 @@ loom-clean --deep --dry-run  # Preview deep clean
 loom-clean --force  # Non-interactive, safe for automation
 ```
 
+**Backlog of pre-existing `[gone]` local branches (#4100)**: `merge-pr.sh` deletes
+the local feature branch for every PR it merges as of #4100, but repos that ran
+Loom before that fix accumulated one orphaned local branch per merged issue —
+each pointing at a remote ref the merge already deleted server-side, so `git
+branch -vv` shows them as `[gone]`. Reap the existing backlog with:
+
+```bash
+git fetch --prune   # drop stale remote-tracking refs first
+loom-clean --force  # clean_branches() deletes local branches whose remote is gone
+```
+
+`loom-clean`'s `clean_branches()` already handles this in two passes: a
+pattern-agnostic pass that deletes any local branch (other than the
+default/current/other-worktree-checked-out ones) whose `origin/<branch>` no
+longer exists, plus an issue-state pass that probes `feature/issue-*` branches
+still tracking a remote against the forge and deletes them once the issue is
+closed. No separate one-off backlog tool is needed.
+
 **Manual cleanup** (if needed, but use with caution):
 
 **WARNING**: Running `git worktree remove` while your shell is in the worktree directory will corrupt your shell state. Always ensure you've navigated out of the worktree first, or use `loom-clean` which handles this safely.
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -9,7 +9,8 @@
 # (see forge-helpers.sh for details).
 #
 # Options:
-#   --no-cleanup-worktree  Skip local worktree cleanup after merge
+#   --no-cleanup-worktree  Skip local worktree AND local branch cleanup after
+#                          merge
 #   --cleanup-worktree     (no-op, worktree cleanup is now the default)
 #   --worktree-path <dir>  Explicit worktree path to clean up (bypasses
 #                          .loom-managed sentinel guard — caller asserts
@@ -28,14 +29,29 @@
 #                          stacked child PRs targeting it (operator asserts the
 #                          children are already reconciled). See #3747 item 2.
 #
-# By default, the local worktree is cleaned up after a successful merge.
-# Pass --no-cleanup-worktree to skip this (e.g., when other terminals may
-# have their CWD inside the worktree).
+# By default, the local worktree AND the local branch it held are cleaned up
+# after a successful merge (#4100). Pass --no-cleanup-worktree to skip both
+# (e.g., when other terminals may have their CWD inside the worktree, or the
+# branch has unpushed commits you want to keep).
 #
 # Cleanup is restricted to Loom-managed worktrees (those containing the
 # .loom-managed sentinel written by worktree.sh). Worktrees lacking the
 # sentinel are treated as user-owned and never removed. Set
 # LOOM_PRESERVE_WORKTREE=1 to disable cleanup unconditionally for a session.
+#
+# Local branch deletion (#4100): every cleanup path — the default
+# .loom/worktrees/issue-N convention, a discovered Loom-managed worktree at a
+# non-standard path, and even the case where no worktree exists at all —
+# attempts to delete the merged PR's local branch. Safety is determined by
+# whether the local branch tip equals the merged PR's head SHA (not by `git
+# branch --merged`, which is always false for a squash merge): a matching tip
+# uses `git branch -D` (every commit on the branch was part of the merged PR);
+# a non-matching tip (unpushed local work) falls back to `git branch -d`,
+# which keeps the branch and reports it instead of force-deleting. A branch
+# checked out as the current HEAD (or in another worktree) is never deleted —
+# git itself refuses, and the refusal is reported with a specific message
+# rather than the generic "unmerged commits" warning. The repo's default
+# branch is never a delete target.
 #
 # Override: pass --worktree-path <dir> to opt into removing a non-Loom
 # worktree (the sentinel guard is bypassed only when this flag is supplied).
@@ -76,7 +92,8 @@ Supports both GitHub and Gitea forges. Forge detection is automatic
 (see forge-helpers.sh for details).
 
 Options:
-  --no-cleanup-worktree  Skip local worktree cleanup after merge
+  --no-cleanup-worktree  Skip local worktree AND local branch cleanup
+                         after merge
   --cleanup-worktree     (no-op, worktree cleanup is now the default)
   --worktree-path <dir>  Explicit worktree path to clean up. Bypasses the
                          .loom-managed sentinel guard (caller asserts
@@ -103,15 +120,26 @@ Options:
                          responsibility, mirroring --worktree-path.
   -h, --help             Show this help and exit
 
-By default, the local worktree is cleaned up after a successful merge.
-Pass --no-cleanup-worktree to skip this (e.g., when other terminals may
-have their CWD inside the worktree).
+By default, the local worktree AND the local branch it held are cleaned up
+after a successful merge (#4100). Pass --no-cleanup-worktree to skip both
+(e.g., when other terminals may have their CWD inside the worktree, or you
+want to keep a branch with unpushed commits).
 
 Cleanup is restricted to Loom-managed worktrees (those under
 .loom/worktrees/issue-N that contain a .loom-managed sentinel file written
 by worktree.sh). User-provisioned worktrees at other paths are never
 removed by the default code path. Set LOOM_PRESERVE_WORKTREE=1 to disable
 cleanup unconditionally for a session.
+
+Local branch deletion (#4100): every cleanup path — including the case
+where no worktree exists at all — attempts to delete the merged PR's local
+branch. Safety is determined by comparing the local branch tip to the
+merged PR's head SHA (not 'git branch --merged', which is always false for
+a squash merge): a matching tip uses 'git branch -D'; a non-matching tip
+(unpushed local work) falls back to 'git branch -d', which keeps the
+branch and reports it instead of force-deleting. The branch currently
+checked out (main worktree or any other) is never deleted, and the repo's
+default branch is never a delete target.
 
 When --worktree-path <dir> is passed explicitly, the operator is taking
 responsibility for the cleanup decision: the sentinel guard is bypassed
@@ -198,6 +226,16 @@ source "$SCRIPT_DIR/lib/forge-helpers.sh"
 # overridden root, not just the default .loom/worktrees.
 # shellcheck source=lib/worktree-root.sh
 source "$SCRIPT_DIR/lib/worktree-root.sh"
+# Default-branch resolver (#4100) — the local-branch delete guard must never
+# target the repo's default branch. Sourced defensively: a repo where this
+# fails to resolve (e.g. no network + no origin/HEAD symref) still falls back
+# to the literal "main"/"master" check in _maybe_delete_local_branch below.
+DEFAULT_BRANCH_NAME=""
+if [[ -f "$SCRIPT_DIR/lib/default-branch.sh" ]]; then
+  # shellcheck source=lib/default-branch.sh
+  source "$SCRIPT_DIR/lib/default-branch.sh"
+  DEFAULT_BRANCH_NAME="$(cd "$REPO_ROOT" && loom_default_branch 2>/dev/null || true)"
+fi
 forge_detect
 
 # Use gh-cached for read-only queries to reduce API calls (see issue #1609)
@@ -292,6 +330,11 @@ PR_MERGED=$(echo "$PR_JSON" | jq -r '.merged')
 PR_BRANCH=$(echo "$PR_JSON" | jq -r '.head.ref')
 PR_TITLE=$(echo "$PR_JSON" | jq -r '.title')
 PR_MERGEABLE=$(echo "$PR_JSON" | jq -r '.mergeable')
+# Head SHA (#4100): the safety criterion for local-branch deletion. A local
+# branch whose tip equals this SHA carries no commits absent from the merged
+# PR, so it is safe to force-delete even though it will never satisfy
+# `git branch --merged` after a squash merge.
+PR_HEAD_SHA=$(echo "$PR_JSON" | jq -r '.head.sha // empty')
 
 # Check if already merged
 if [[ "$PR_MERGED" == "true" ]]; then
@@ -1171,8 +1214,15 @@ if [[ "$PR_MERGEABLE" == "false" ]]; then
 fi
 
 if [[ "$DRY_RUN" == "true" ]]; then
-  info "[dry-run] Would merge PR #$PR_NUMBER (squash) and delete branch '$PR_BRANCH'"
-  [[ "$CLEANUP_WORKTREE" == "true" ]] && info "[dry-run] Would clean up local worktree"
+  info "[dry-run] Would merge PR #$PR_NUMBER (squash) and delete remote branch '$PR_BRANCH'"
+  if [[ "$CLEANUP_WORKTREE" == "true" ]]; then
+    info "[dry-run] Would clean up local worktree"
+    if git -C "$REPO_ROOT" show-ref --verify --quiet "refs/heads/$PR_BRANCH"; then
+      info "[dry-run] Would delete local branch '$PR_BRANCH'"
+    fi
+  else
+    info "[dry-run] --no-cleanup-worktree: would leave local worktree and local branch '$PR_BRANCH' in place"
+  fi
   exit 0
 fi
 
@@ -1289,12 +1339,12 @@ _auto_reconcile_stacked_children || true
 # Delete remote branch (skip if forge auto-deletes on merge)
 DELETE_BRANCH_ON_MERGE=$(forge_check_auto_delete "$REPO_NWO" "$GH")
 if [[ "$DELETE_BRANCH_ON_MERGE" == "true" ]]; then
-  info "Skipping branch deletion (auto-delete is enabled)"
+  info "Skipping remote branch deletion (auto-delete is enabled)"
 else
   info "Deleting remote branch: $PR_BRANCH"
   forge_delete_branch "$REPO_NWO" "$PR_BRANCH" && \
-    success "Branch '$PR_BRANCH' deleted" || \
-    warning "Could not delete branch '$PR_BRANCH' (may already be deleted)"
+    success "Remote branch '$PR_BRANCH' deleted" || \
+    warning "Could not delete remote branch '$PR_BRANCH' (may already be deleted)"
 fi
 
 # Cleanup worktree if requested.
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review — APPROVED

Reviewed against issue #4100's acceptance criteria and the curator's consolidated guidance. Checked out the PR into a scratch worktree and verified independently (not just re-reading the diff).

**Correctness / AC verification**

- **Head-SHA tip comparison, not `git branch --merged` / naive `-d`**: `_maybe_delete_local_branch` now takes an optional `expected_head_sha`; a matching local-branch tip uses `git branch -D` (safe — every commit was part of the merged, squashed PR), a mismatch falls back to the original `git branch -d` (keeps the branch, reports the `-D` escape hatch). Confirmed via behavioral tests (a) and (b), which exercise real git repos rather than mocks.
- **Placement inside the worktree-cleanup dispatch block**: the new `_maybe_delete_local_branch "$PR_BRANCH" "$PR_HEAD_SHA"` call sits at `merge-pr.sh:1620`, inside the `else` arm of the `CLEANUP_WORKTREE` dispatch, after the default-path / discovery-path / no-worktree if/elif/else resolves — exactly the placement the curator flagged as load-bearing (not after the whole `if [[ "$CLEANUP_WORKTREE" ... ]]` block, which would have broken the `--no-cleanup-worktree` / `LOOM_PRESERVE_WORKTREE=1` guarantee). I traced all four code paths by hand: the `--worktree-path` override path already deletes its branch internally via `_remove_loom_worktree`'s `allow_unmanaged` branch (unchanged, no double-delete); the default/discovery/no-worktree paths now all funnel through the single new call site.
- **Never deletes current-HEAD or default branch**: `_maybe_delete_local_branch` refuses up front when the branch matches `$DEFAULT_BRANCH_NAME` (resolved via `lib/default-branch.sh`, with a literal `main`/`master` fallback) or is checked out elsewhere — git's own refusal is caught and given a specific message. Verified with tests (c) checked-out-in-another-worktree, (c2) checked-out-as-current-HEAD, and (d) default-branch refusal (even with a matching SHA tip).
- **`--dry-run` preview**: previews `Would delete local branch '$PR_BRANCH'` only when a matching local ref exists, and prints the correct "would leave in place" message under `--no-cleanup-worktree`. Verified by reading the dry-run block; matches the PR's stated manual test against a real open PR.
- **Disambiguated log lines**: `Branch 'X' deleted` → `Remote branch 'X' deleted` / `Local branch 'X' deleted$safety_note`; static test confirms the old ambiguous line is gone.
- **Docs**: the backlog remedy (`git fetch --prune` + `loom-clean --force`) is in `defaults/docs/troubleshooting.md` (source of the `.loom/docs/troubleshooting.md` symlink), not `defaults/CLAUDE.md` — correct per the curator's guidance and the repo's own doc-placement convention.

**Testing** — checked out the PR into a scratch worktree (not the repo's `main` worktree) and ran everything locally:
- `test-merge-pr-local-branch-cleanup.sh`: 18/18 pass, including the extract-and-eval behavioral suite against real git repos.
- Full regression sweep across all `test-merge-pr-*.sh` files: 213/213 pass (help, worktree-path, primary-worktree-guard, auto-merge-disabled-fallback, auto-reconcile, merge-ordering-guard, partial-increment, unstable-fallback, worktree-awk).
- `shellcheck --severity=warning` on `merge-pr.sh` and the new test file: clean.

**CI / mergeability** — all required checks green (Shellcheck, CLAUDE.md Line Budget, Detect Changes, Installer Integration Tests; Rust jobs correctly skipped for a shell-only change), `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

**Security** — no concerns. Branch names flow into `git branch`/`git rev-parse` as positional args, not shell-interpolated commands; the default-branch and current-HEAD/other-worktree guards are checked before any destructive git call; the helper never fails the overall cleanup pipeline (always returns 0).

No issues found. This is thorough, well-tested work that correctly implements the trickiest part of the spec (the placement nuance the curator explicitly called out as load-bearing).

PR #4156, opened 2026-07-28
fix(daemon): embed --depends-on in the -p prompt, not as a sibling argv token

### LINKED ISSUE #4121: loom-daemon: SweepRegistry::dispatch's --depends-on argv flag is rejected by the real claude CLI (unknown option)
> ## 📌 Curator re-pin — 2026-07-28, `origin/main` @ `0a7c1581` (AUTHORITATIVE)
>
> **Premise re-verified: the bug is still live.** `sweep_registry.rs:2735` still does `cmd.arg("--depends-on").arg(parent.to_string())` as a sibling argv token, 29 lines after the comment at `:2696` that names `--depends-on` in its own list of skill-consumed flags that must live inside `-p`. **PR #4120 fixed only `--claim-owned`** (made it positional at `:2707`); the identical `--depends-on` defect was left in place. That is the single most load-bearing fact in this issue and it was previously only implied.
>
> **This body carries two anchor sets.** The *re-verification / Implementation Guidance* set is accurate. The **`## Affected Files` and `## Test Plan` sets are stale by an earlier pin** — every number below is corrected here. Where they disagree, **this block wins.**
>
> | Cited in Affected Files / Test Plan | Actually on `0a7c1581` |
> |---|---|
> | delete `cmd.arg("--depends-on")` at `:2681` | **`:2735`** (`:2681` is `Utc::now().to_rfc3339()`) |
> | `spawn_child ~L2653` | fn sig is `:2652-2653`; **the edit site is `:2707`**, where `prompt` is built |
> | stale doc comment `:1544-1548` | **`:1598-1603`** (`:1599` = `/// child receives '--depends-on <N>' appended to the argv`) |
> | harden test `:5756` | **`:5932`** (`dispatch_with_depends_on_appends_depends_on_arg`) |
> | `dispatch_appends_claim_owned_flag` `:5140` *(cited twice)* | **`:5316`** |
> | `dispatch_without_depends_on_emits_no_flag` `:5781` | **`:5957`** |
> | `dispatch_appends_dangerously_skip_permissions` `:5084` | **`:5261`** |
> | dispatch log header `:2628` | **`:2680`** |
>
> **Verified exact, no change needed**: `:2693-2706` (comment block), `:2707` (fix template), `:2735`, `:1598-1603`, `:4981-4988` (fixture), `spawn-claude.sh:444` / `:189` / `:239`, `sweep.md:124` / `:129` / `:1405`, `role_runner.rs:352`, `main.rs:2069-2091`. All three `## Affected Files` paths exist on `origin/main`.
>
> **Scope-guard addition — a second `claude`-spawning site the body omits.** `epic_supervisor.rs:1405-1409` also spawns `claude`, but passes only `-p` / `--model` / `--dangerously-skip-permissions` — all real CLI flags. **Not affected; do not "fix" it.** A full enumeration (`git grep -n 'arg("--' -- loom-daemon/src`) confirms `sweep_registry.rs:2735` is the *only* non-`claude` flag appended as a sibling argv token. `--auto-stack` has zero Rust call sites, so the out-of-scope claim holds.
>
> **The test-blindness finding is confirmed and matters.** The fixture records both forms (`:4981` flattened `argv:`, `:4988` per-token `arg:`), but the depends-on test at `:5943` asserts only the flattened `argv: -p /loom:sweep 50 --claim-owned 50 --depends-on 49` — **identical before and after the fix**. A green `cargo test` proves nothing here; the new test must assert on the per-token `arg:` lines.
>
> Everything else (root cause, suggested fix, `--model`/`--effort` untouched) is verified correct and stands unchanged.

## Bug Description

While reviewing PR #4120 (issue #4111), I discovered that `SweepRegistry::spawn_child` (`loom-daemon/src/sweep_registry.rs`) appends `--depends-on <parent>` (issue #3729) as a **separate argv token after** the `-p "<prompt>"` argument, e.g.:

```
claude -p "/loom:sweep 4242" --depends-on 99 --dangerously-skip-permissions
```

`spawn-claude.sh` forwards all such tokens verbatim to the real `claude` binary (`exec claude "${PASSTHROUGH_ARGS[@]}"`) — it does not special-case or fold them back into the prompt. The real Claude Code CLI's own argument parser does not recognize `--depends-on` as a flag (only the slash-command skill does, via its own `$ARGUMENTS` parsing of the *prompt text*), so it hard-fails before a session ever starts:

```
$ claude -p "/loom:sweep 4242" --depends-on 99 --dangerously-skip-permissions
error: unknown option '--depends-on'
$ echo $?
1
```

Reproduced locally against `claude --version` 2.1.212.

## Impact

Any daemon dispatch that sets `depends_on` (stacked-PR mode, `--depends-on`/`--auto-stack` in `/loom:sweep`, issue #3729) would fail to spawn a working child session at all — the process exits 1 immediately with no `.loom/logs/sweep-issue-<N>.log` build activity, rather than running `/loom:sweep`'s stacked-dependency logic. This is silent: `dispatch()` still records a `SweepInfo` entry and returns success:true from the caller's perspective; only the reaper's later liveness poll would observe the near-instant `exit_code=Some(1)` death (likely misclassified as an ordinary insta-crash / token-pool-exhaustion failure, not attributed to this argv bug).

## Root Cause

`--model`, `--effort`, and `--dangerously-skip-permissions` are legitimate, CLI-recognized `claude` flags (confirmed via `claude --help`), so those work fine as separate argv tokens. `--depends-on` (and, as of PR #4120, `--claim-owned`) are **not** real `claude` CLI flags — they are meant to be consumed by the `/loom:sweep` skill's own `$ARGUMENTS` parsing, which only sees the text *inside* the single `-p "<prompt>"` string. Appending them as sibling `cmd.arg()` calls sends them to `claude`'s own parser instead, which rejects them.

## Suggested Fix

Embed dispatch-only flags meant for the skill's own argument parsing directly in the prompt text, e.g.:

```rust
let mut prompt = format!("/loom:sweep {issue}");
if let Some(parent) = depends_on {
    prompt.push_str(&format!(" --depends-on {parent}"));
}
```

rather than as separate `cmd.arg()` tokens after `-p`. (PR #4120 hits the identical issue with its new `--claim-owned` flag — flagged there directly since it's part of that PR's own diff.)

## Reproduction

```bash
claude -p "/loom:sweep 4242" --depends-on 99 --dangerously-skip-permissions
# error: unknown option '--depends-on'
```

Discovered while evaluating PR #4120.

---

## Curator Enhancement (2026-07-28)

Re-verified against `origin/main` @ `5266454d` — **the bug is still live**, and the fix
template now exists in-tree. Adding the two required sections (`Affected Files`,
`Test Plan`), exact line anchors, and one finding that materially changes the test
strategy. **Scope, root cause, and suggested fix are unchanged and confirmed** — this
is an additive refresh, so `loom:issue` is retained.

> **Anchor durability warning.** `sweep_registry.rs` has now rotted **twice in ~24h**
> (71 commits landed on `main` in that window; this file gained several hundred lines).
> The `sweep_registry.rs:NNNN` numbers below were re-pinned to `5266454d` on 2026-07-28
> and **will drift again before a Builder picks this up**. Anchor on the *symbols*, which
> are stable: the `if let Some(parent) = depends_on` block inside `spawn_child`, the
> `let prompt = format!("/loom:sweep {issue} --claim-owned {issue}")` fix template, and
> the test fns `dispatch_with_depends_on_appends_depends_on_arg` /
> `dispatch_without_depends_on_emits_no_flag` / `dispatch_appends_claim_owned_flag`.
> Treat every line number as a hint, not a contract.

### Re-verification against current `main`

| Claim in the body | Status on `origin/main` @ `5266454d` |
|---|---|
| `--depends-on` appended as a sibling `cmd.arg()` after `-p` | ✅ **Still true** — `loom-daemon/src/sweep_registry.rs:2735` |
| PR #4120 hits the identical issue with `--claim-owned` | ✅ **Fixed there** — `sweep_registry.rs:2707` now embeds it in the prompt |
| `spawn-claude.sh` forwards tokens verbatim | ✅ **Still true** — `defaults/scripts/spawn-claude.sh:444` (`exec claude "${PASSTHROUGH_ARGS[@]}"`) |
| The skill only parses flags from prompt text | ✅ **Confirmed by spec** — `defaults/.claude/commands/loom/sweep.md:124` ("Recognized anywhere in `$ARGUMENTS`") |

**The code already documents its own violation.** The comment block at
`sweep_registry.rs:2693-2706` (written by #4120) explicitly lists `--depends-on` among
the skill-consumed flags that must live *inside* the `-p` prompt, and warns that a
sibling arg "makes `claude` exit 1 … before any session starts". Twenty-nine lines
later, `sweep_registry.rs:2735` does exactly that. The comment is correct; the code
below it is not.

### ⚠️ Finding: the existing test is blind to this bug and will stay green after the fix

`dispatch_with_depends_on_appends_depends_on_arg` (`sweep_registry.rs:5932`) asserts:

```rust
recorded.contains("argv: -p /loom:sweep 50 --claim-owned 50 --depends-on 49")
```

The fixture records `argv:` from `"$*"`, which **flattens argv tokens into
space-joined text**. Both the buggy and the fixed forms render byte-identically:

| | argv tokens | `$*` renders as |
|---|---|---|
| **Buggy (today)** | `-p` / `/loom:sweep 50 --claim-owned 50` / `--depends-on` / `49` | `-p /loom:sweep 50 --claim-owned 50 --depends-on 49` |
| **Fixed** | `-p` / `/loom:sweep 50 --claim-owned 50 --depends-on 49` | `-p /loom:sweep 50 --claim-owned 50 --depends-on 49` |

So this test **passes today with the bug present and will still pass after the fix** —
it cannot distinguish the two. This is precisely the trap the #4120 fixture comment
documents at `sweep_registry.rs:4981-4988` ("a sibling `--claim-owned` … slipped past
the `$*`-substring tests"). The per-token `arg: ` recording that #4120 added to the
fixture is the fix for it, but it was only ever applied to `--claim-owned`.

**Implication for the Builder**: a green `cargo test` is *not* evidence this issue is
fixed. The regression guard must use the per-token `arg: ` lines, mirroring
`dispatch_appends_claim_owned_flag` (`sweep_registry.rs:5316`).

### Implementation Guidance

The exact fix template is already in-tree at `sweep_registry.rs:2707`. The prompt is
built *before* `cmd.arg("-p")`, so the `depends_on` append must move up rather than
being edited in place:

```rust
// sweep_registry.rs ~2653 — build the full prompt BEFORE `cmd.arg("-p")`.
let mut prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
if let Some(parent) = depends_on {
    prompt.push_str(&format!(" --depends-on {parent}"));
}
let mut cmd = Command::new(&spawn_bin);
cmd.arg("-p").arg(&prompt);
// ... --model / --effort unchanged (both are REAL claude CLI flags) ...
// delete the `cmd.arg("--depends-on")` block at ~2681
```

Notes:

- **Append after `--claim-owned`**, not before. That ordering keeps the existing flat
  `argv:` assertions in `dispatch_with_depends_on_appends_depends_on_arg` and
  `dispatch_appends_dangerously_skip_permissions` (`sweep_registry.rs:5261`) passing
  unchanged. The skill parses flags positionally-agnostically
  (`sweep.md:129` — "anywhere in `$ARGUMENTS`"), so ordering is a test-stability
  choice, not a correctness one.
- **Do not touch `--model` / `--effort`.** Both are genuine `claude` CLI flags and
  are correct as sibling args; `spawn-claude.sh:189,239` also injects them that way.
- **Update the stale doc comment** at `sweep_registry.rs:1598-1603`, which currently
  promises `--depends-on` is "appended to the argv (immediately after any `--model` /
  `--effort`)". That sentence becomes wrong with this fix.

### Scope guard

Single-site fix — **~15 LOC production, ~40 LOC test**. Verified by enumeration
(`git grep -n 'arg("--' -- loom-daemon/src`) that `sweep_registry.rs:2735` is the
**only** place a non-`claude` flag is appended as a sibling argv token:

- `role_runner.rs:352` appends only `--dangerously-skip-permissions` — a real
  `claude` flag. **Not affected.**
- `--auto-stack` (#3759) is documented in `sweep.md` but **never emitted by the
  daemon** — no Rust call site exists. **Out of scope**, nothing to fix.
- `main.rs:2069-2091` (`loom-daemon dispatch --depends-on`) is the daemon's *own*
  clap CLI, which legitimately defines the flag. **Not affected.**

Out of scope: the `--auto-stack` per-issue map, stacked-PR reconciliation (#3747),
and any change to `spawn-claude.sh`'s pass-through contract.

## Affected Files

- `loom-daemon/src/sweep_registry.rs` — **the fix**. (1) `spawn_child` ~L2653: fold
  `--depends-on <parent>` into the `-p` prompt string; (2) delete the sibling
  `cmd.arg("--depends-on")` at L2681; (3) correct the stale doc comment at
  L1544-1548; (4) harden the test at L5756 with per-token `arg: ` assertions.
- `defaults/.claude/commands/loom/sweep.md` — **reference only, no change expected**.
  L124 already specifies the correct consumer-side contract; cite it, don't edit it.
- `defaults/scripts/spawn-claude.sh` — **reference only, no change expected**. L444
  (`exec claude "${PASSTHROUGH_ARGS[@]}"`) is the verbatim-forwarding behavior that
  makes the sibling arg fatal. The pass-through contract is deliberate (#3824) and
  must not be changed to special-case skill flags.

## Test Plan

- [ ] **Automated (regression guard — the important one)**: harden
  `dispatch_with_depends_on_appends_depends_on_arg` (`sweep_registry.rs:5932`) with
  per-token assertions mirroring `dispatch_appends_claim_owned_flag` (L5140):
  ```rust
  assert!(recorded.contains("arg: /loom:sweep 50 --claim-owned 50 --depends-on 49"),
      "--depends-on must live inside the single -p prompt token");
  assert!(!recorded.contains("arg: --depends-on"),
      "--depends-on must NOT be a standalone argv token (real claude CLI rejects it)");
  ```
  Confirm the *second* assertion **fails on unmodified `main`** — if it passes before
  the fix, the test is still blind and does not guard anything.
- [ ] **Automated**: `dispatch_without_depends_on_emits_no_flag` (L5781) stays green —
  `depends_on: None` must emit no `--depends-on` anywhere (prompt or argv).
- [ ] **Automated**: `dispatch_appends_dangerously_skip_permissions` (L5084) and
  `dispatch_appends_claim_owned_flag` (L5140) stay green — `--claim-owned` must remain
  prompt-embedded and `--dangerously-skip-permissions` must remain the final sibling arg.
- [ ] **Automated**: `cd loom-daemon && cargo test` fully green (the `#[serial]`
  dispatch tests share the fixture registry).
- [ ] **Manual verification (proves the actual bug)**: on unmodified `main`,
  `claude -p "/loom:sweep 4242" --depends-on 99 --dangerously-skip-permissions` exits
  `1` with `error: unknown option '--depends-on'`. After the fix, dispatch with
  `depends_on` set and confirm the child **starts a session** rather than dying
  instantly — e.g. `loom-daemon dispatch <N> --depends-on <parent>`, then check
  `.loom/logs/sweep-issue-<N>.log` shows real `/loom:sweep` pre-flight activity past
  the `==== loom-daemon dispatch: ... ====` header (L2628).
- [ ] **Edge case**: `--model` + `--effort` + `--depends-on` on the same dispatch —
  the two real CLI flags stay sibling args, only `--depends-on` moves into the prompt.
- [ ] **Edge case**: the skill actually consumes it — the dispatched child's Builder
  branches off `feature/issue-<parent>` and opens its PR with
  `--base feature/issue-<parent>` (`sweep.md:1405`), which is the end-to-end behavior
  #3729 promised and that this bug has been silently preventing since it shipped.

rjwalters:
## Summary
`SweepRegistry::spawn_child` (`loom-daemon/src/sweep_registry.rs`) appended `--depends-on <parent>` as a separate `cmd.arg()` sibling token after `-p "<prompt>"` when dispatching a stacked-PR child `claude` process. The real `claude` CLI does not recognize `--depends-on` as a flag — it's meant to be parsed by the `/loom:sweep` skill's own `$ARGUMENTS` handling from inside the prompt text — so the child process immediately exited 1 with `error: unknown option '--depends-on'`. This is the same class of bug PR #4120 already fixed for `--claim-owned`, but `--depends-on` was left untouched.

## Changes
- Build the full `-p` prompt string BEFORE calling `cmd.arg("-p")`, appending `--depends-on <parent>` to that prompt string when `depends_on` is `Some`, instead of appending it as a separate sibling argv token afterward — mirroring the existing `--claim-owned` fix pattern.
- Removed the now-dead sibling-arg append site and its stale comment; consolidated the "embedded in the prompt, not sibling args" note to cover both `--claim-owned` and `--depends-on`.
- Updated the stale doc comment on `dispatch()` that claimed `--depends-on` is "appended to the argv (immediately after any `--model` / `--effort`)" to reflect that it is now embedded in the `-p` prompt string, immediately after `--claim-owned`.
- Updated the `--dangerously-skip-permissions` ordering comment (no longer lists `--depends-on` among the sibling-arg flags it follows).
- Did **not** touch `--model` / `--effort` / `--dangerously-skip-permissions` (real `claude` CLI flags, correctly siblings) or `role_runner.rs` / `epic_supervisor.rs` (only pass real CLI flags) or `main.rs`'s own `loom-daemon dispatch --depends-on` clap CLI (the daemon's own CLI, unaffected).

| `--depends-on <parent>` is embedded in the `-p` prompt string, not a sibling argv token | ✅ | `cargo test dispatch_with_depends_on_appends_depends_on_arg` passes; new `arg: /loom:sweep 50 --claim-owned 50 --depends-on 49` assertion (single token) and `!recorded.contains("arg: --depends-on")` (no standalone token) both pass |
| Existing test hardened so it actually detects the bug | ✅ | Added per-token `arg:` assertions to `dispatch_with_depends_on_appends_depends_on_arg`; confirmed they **fail** against unfixed `main` (`error: unknown option` reproduced via the sibling-token assertion failing) before applying the fix, then confirmed they pass after |
| `dispatch_without_depends_on_emits_no_flag` stays green (no `--depends-on` when `depends_on` is `None`) | ✅ | `cargo test dispatch_without_depends_on_emits_no_flag` passes |
| `dispatch_appends_dangerously_skip_permissions` / `dispatch_appends_claim_owned_flag` stay green | ✅ | Both pass unchanged |
| Stale doc comment updated | ✅ | `dispatch()` doc comment and the `--dangerously-skip-permissions` ordering comment both updated |
| Scope: single-site fix in `sweep_registry.rs` | ✅ | `git diff --stat`: 1 file changed, 44 insertions(+), 18 deletions(-) |

## Test Plan
- `cargo test --lib` in `loom-daemon/`: 1319 passed, 0 failed.
- `cargo test --test integration_basic`: 9 pre-existing failures (`Daemon failed to create socket within 5s`), confirmed identical on unmodified code via `git stash` — unrelated to this change (daemon-startup socket integration tests, not `sweep_registry.rs`).
- `cargo fmt -- --check`: clean.
- `cargo clippy --lib -- -D warnings`: clean.
- Manually confirmed the regression guard is real: reverted the fix locally and re-ran `dispatch_with_depends_on_appends_depends_on_arg` — it failed with `expected --depends-on inside the single -p prompt token; got: ... arg: --depends-on / arg: 49 ...`, proving the new assertions actually catch the bug the old flattened-`argv:` assertion missed.

Closes #4121

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -1596,11 +1596,13 @@ impl SweepRegistry {
     /// effort is preserved end-to-end.
     ///
     /// `depends_on` (issue #3729, stacked-PR v1): when `Some(N)`, the spawned
-    /// child receives `--depends-on <N>` appended to the argv (immediately
-    /// after any `--model` / `--effort`), instructing `/loom:sweep` to branch
-    /// its worktree/PR off `feature/issue-<N>`. When `None`, no
-    /// `--depends-on` flag is emitted — byte-for-byte unchanged behavior. A
-    /// single optional parent (not a list) makes diamonds unrepresentable.
+    /// child receives `--depends-on <N>` embedded in the `-p` prompt string
+    /// (immediately after `--claim-owned`; issue #4121 — NOT a sibling argv
+    /// token, since `--depends-on` is not a real `claude` CLI flag),
+    /// instructing `/loom:sweep` to branch its worktree/PR off
+    /// `feature/issue-<N>`. When `None`, no `--depends-on` text is emitted —
+    /// byte-for-byte unchanged behavior. A single optional parent (not a
+    /// list) makes diamonds unrepresentable.
     pub fn dispatch(
         &mut self,
         kind: &SweepKind,
@@ -2704,7 +2706,18 @@ impl SweepRegistry {
         // pre-flight. The env var (`LOOM_SWEEP_CLAIM_OWNED`, set below) is kept
         // unchanged for backward compatibility (#3823/#3967). Unconditional on
         // every daemon dispatch — every dispatch claims exactly one issue.
-        let prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
+        let mut prompt = format!("/loom:sweep {issue} --claim-owned {issue}");
+        // Stacked-PR dependency (issue #3729, v1; sibling-arg bug fixed in
+        // #4121): when a parent issue is declared, fold `--depends-on <N>`
+        // INTO the same `-p` prompt string as `--claim-owned` above, for the
+        // identical reason — `--depends-on` is not a `claude` CLI flag
+        // either, so a sibling `cmd.arg()` token makes `claude` exit 1
+        // (`error: unknown option '--depends-on'`) before any session
+        // starts. Absent the param, no text is appended (byte-for-byte
+        // unchanged). Mirrors the `--claim-owned` append-only contract.
+        if let Some(parent) = depends_on {
+            prompt.push_str(&format!(" --depends-on {parent}"));
+        }
         let mut cmd = Command::new(&spawn_bin);
         cmd.arg("-p").arg(&prompt);
         // Model selection (issue #3477, Phase 1): the dispatch-param tier of
@@ -2726,16 +2739,9 @@ impl SweepRegistry {
                 cmd.arg("--effort").arg(e);
             }
         }
-        // Stacked-PR dependency (issue #3729, v1): when a parent issue is
-        // declared, append `--depends-on <N>` so `/loom:sweep` branches the
-        // child worktree/PR off `feature/issue-<N>` instead of the default
-        // branch. Absent the param, no flag is emitted (byte-for-byte
-        // unchanged). Mirrors the `--model` / `--effort` append-only contract.
-        if let Some(parent) = depends_on {
-            cmd.arg("--depends-on").arg(parent.to_string());
-        }
-        // (Daemon self-claim marker `--claim-owned <N>` is embedded in the
-        // `-p` prompt text above, not appended as a sibling arg — see #4111.)
+        // (Daemon self-claim marker `--claim-owned <N>` and the stacked-PR
+        // `--depends-on <N>` marker are both embedded in the `-p` prompt text
+        // above, not appended as sibling args — see #4111 / #4121.)
         // Unattended-permissions flag (issue #3824): a daemon-dispatched child
         // is a detached, non-interactive `claude -p` process — there is no
         // human to answer a permission prompt, so any tool call needing
@@ -2747,8 +2753,8 @@ impl SweepRegistry {
         // `claude -p "/<role>" --dangerously-skip-permissions`). Scoped to this
         // daemon-only dispatch path; `spawn-claude.sh` stays a generic
         // pass-through and never adds a permission flag of its own. Appended
-        // AFTER `--model`/`--effort`/`--depends-on` (and the prompt-embedded
-        // `--claim-owned`) so the positional argv contract is unchanged.
+        // AFTER `--model`/`--effort` (and the prompt-embedded `--claim-owned`
+        // / `--depends-on`) so the positional argv contract is unchanged.
         cmd.arg("--dangerously-skip-permissions");
         cmd.env("LOOM_TERMINAL_ID", format!("daemon-{sweep_id}"))
             // Claim-ownership marker (issue #3823): `dispatch()` flips
@@ -5942,6 +5948,26 @@ exit 0
         assert!(
             recorded.contains("argv: -p /loom:sweep 50 --claim-owned 50 --depends-on 49"),
             "expected --depends-on in argv; got: {recorded}"
+        );
+        // Regression for #4121 (mirrors the #4120 `--claim-owned` review): the
+        // flattened `argv:` assertion above renders byte-identically whether
+        // `--depends-on 49` is embedded in the single `-p` prompt token or
+        // appended as its own sibling argv token — so it cannot distinguish
+        // the fixed and buggy forms. The real `claude` CLI rejects
+        // `--depends-on` as an unknown option if it ever arrives as a
+        // standalone token; only text inside the `-p "<prompt>"` value
+        // reaches the `/loom:sweep` skill's `$ARGUMENTS`. Use the per-token
+        // `arg: ` fixture lines (as `dispatch_appends_claim_owned_flag` does)
+        // to assert the flag is part of the prompt VALUE and NOT a
+        // standalone token.
+        assert!(
+            recorded.contains("arg: /loom:sweep 50 --claim-owned 50 --depends-on 49"),
+            "expected --depends-on inside the single -p prompt token; got: {recorded}"
+        );
+        assert!(
+            !recorded.contains("arg: --depends-on"),
+            "--depends-on must NOT be a standalone argv token (the real claude CLI \
+             rejects it as an unknown option); got: {recorded}"
         );
         assert_eq!(
             registry.get(&outcome.sweep_id).unwrap().depends_on,


rjwalters:
LGTM — approved.

**Root cause fix verified.** `spawn_child` now builds the full `-p` prompt (including `--depends-on <parent>` when set) before calling `cmd.arg("-p")`, so `--depends-on` rides inside the single prompt token instead of arriving as a sibling argv token the real `claude` CLI rejects. This mirrors the `--claim-owned` pattern from #4120 exactly, as the issue's suggested fix specified. Diff is a clean single-site change in `loom-daemon/src/sweep_registry.rs` (44 insertions / 18 deletions, no unrelated files touched) — matches the scope guard in #4121.

**Verified the hardened test is a real regression guard, not blind (independently, not just trusting the PR description).** I manually reverted the production fix in a local worktree while keeping the new test assertions and reran `dispatch_with_depends_on_appends_depends_on_arg`:

```
thread '...dispatch_with_depends_on_appends_depends_on_arg' panicked:
expected --depends-on inside the single -p prompt token; got:
argv: -p /loom:sweep 50 --claim-owned 50 --depends-on 49 --dangerously-skip-permissions
arg: -p
arg: /loom:sweep 50 --claim-owned 50
arg: --depends-on   <-- standalone token, exactly the bug
arg: 49
```

It fails as expected against the unfixed code, confirming the new per-token `arg:` assertions (mirroring `dispatch_appends_claim_owned_flag`) actually catch what the old flattened `argv:` assertion could not (both forms render identically under `$*`-style concatenation). Restored the fix afterward — worktree is clean and matches the PR branch.

**Other checks:**
- `cargo test --lib` (loom-daemon): 1319 passed, 0 failed.
- `cargo fmt -- --check`: clean.
- `cargo clippy --lib -- -D warnings`: clean.
- Doc comment on `dispatch()` and the `--dangerously-skip-permissions` ordering comment both correctly updated to reflect the new embedding.
- CI: Rust Linting & Build Check, Rust Unit Tests, CLAUDE.md Line Budget all pass.
- `mergeStateStatus`: CLEAN.

Nice catch chaining off #4120 and a genuinely useful test-hardening pattern for the next flag added to this dispatch path.

PR #4174, opened 2026-07-28
fix(daemon): spawn sweep watchdog for every pooled workspace, not just default

### LINKED ISSUE #4124: Sweep watchdogs (startup/mid-build/review-stall) are spawned only for the DEFAULT workspace — pooled workspaces get a reaper and nothing else
> ## 📌 Curator re-pin — 2026-07-28, `origin/main` @ `0a7c1581` (AUTHORITATIVE)
>
> **Premise re-verified: still fully valid — NOT fixed, not even partially at the cause.** `spawn_watchdog_task` has exactly one call site in the entire tree (`main.rs:847`, default-workspace-only, passing the default `sweep_registry`); `workspace_pool.rs:160-175` still spawns only `spawn_reaper_task`; `PooledRegistry` still has only a `_reaper` field (`:53-59`). A broad grep for `watchdog_once|spawn_watchdog|review_stall_watchdog|midbuild_watchdog` finds **no additional spawn sites**.
>
> **`5266454d` (#4141) mitigates the consequence, not the cause** — its entire `workspace_pool.rs` footprint is 3 lines (`:148-150`, `set_startup_proof_grace`, 90s default at `sweep_registry.rs:485`). A wedged pooled sweep now releases its occupancy slot at ~90s instead of 300s, but is still never cancelled, never re-dispatched, and its issue still sits at `loom:building` indefinitely.
>
> **First pass where the anchors did NOT rot.** `5266454d` is an ancestor of `0a7c1581`, and **none** of the 5 commits since touch `sweep_registry.rs`, `workspace_pool.rs`, `main.rs`, or `work_finder.rs`. The body's "~24h shelf life" warning did not bite this time. The pass-2 table is correct except the four items below.
>
> | Cited in pass-2 table | Actually on `0a7c1581` |
> |---|---|
> | `spawn_watchdog_task` span `:4711-4783` | **`:4711-4773`** (fn closes `:4773`; doc comment starts `:4700`). `:4783` points into `is_pid_alive`'s doc comment. |
> | ticker loop `:4726-4756` | **loop is `:4731-4771`**; whole `tokio::spawn` block `:4726-4772`; the three backstop calls are **`:4740-4752`**. `:4756` lands mid-`log::warn!`. |
> | "`evict` seeded-refusal **test** range `:202-210`" | **Mislabeled noun.** `:202-210` is the seeded guard *inside `evict` itself*. The **test** is `evict_seeded_default_workspace_is_rejected` at **`workspace_pool.rs:347-348`**. Correct both the table row and the Test Plan bullet citing `:202-210`. |
> | module docs (reaper rationale) `:26-34` | **`:23-32`** (`:23` = `# Reaper threading` heading, `:32` = last rationale line; `:34` is already the next section). |
>
> **Missed enumeration — there are TWO tick-log gates, and pass 2 dropped one.** `work_finder.rs:1152` (in `spawn_work_finder_task`, `:1060`, single-workspace) **and `:1353`** (in `spawn_multi_work_finder_task`, `:1233`, the multi-workspace path this issue is actually about). Identical disjunctions, both omitting `skipped_labeled` (counted at `:382`, `:466`, `:693`; printed at `:1163`/`:1364`). The original body correctly cited two; pass 2 collapsed to one. **`:1353` is the load-bearing one here.**
>
> **Enumeration claims that DO hold**: 12 `spawn_*_task` helpers under `loom-daemon/src`; only two take `Arc<Mutex<SweepRegistry>>` (`spawn_reaper_task` `:4390`, `spawn_watchdog_task` `:4711`); the four `spawn_multi_*` helpers take `Arc<WorkspacePool>` and are pool-aware.
>
> **Docs finding — the edit is additive, not corrective.** `defaults/docs/daemon-reference.md:378-400` ("Per-workspace registry pool") mentions **only** the reaper (`:394` "aborts its background reaper task", `:398` "only the in-memory tracking + reaper go away"), and the startup-watchdog section (`:1556-1569`) is workspace-agnostic. **The docs do not currently overclaim** — so don't hunt for a false claim to fix; state the new coverage and add the watchdog to the eviction sentence at `:394`.
>
> **Symlink note, reconciled** (the body contains two seemingly contradictory statements — both are true): `.loom/docs` *is* a real tree, but most files inside it are `120000` symlink blobs into `defaults/docs/` (`daemon-reference.md`, `build-gate.md`, `troubleshooting.md`, `forge-authentication.md`, `github-authentication.md`, `tool-use-concurrency-errors.md`); only `safehouse.md` and `token-pool.md` are real files. **Always cite `defaults/docs/daemon-reference.md`.**
>
> **All pass-0 narrative anchors are drifted and superseded** by the pass-2 table (`:3214-3328`, `:452`, `:455`, `:3456`, `:3627`, `:515`, `:4323-4385`, `main.rs:771-795`, `workspace_pool.rs:110`/`:143-157`/`:133`/`:137`, `:4138-4169`, `work_finder.rs:1067`/`:1249`). Treat the pass-2 table plus this block as the only live anchor set.

Found during curation of #4007 (2026-07-27), verified against `origin/main` (`e6522b71`). This was not in the original report — it surfaced while tracing why wedged sweeps were not being reclaimed.

## Problem

`loom-daemon` runs three self-healing watchdogs over the sweep registry:

- **startup watchdog** — `watchdog_once` (`loom-daemon/src/sweep_registry.rs:3214-3328`), default timeout 300s (`:452`), interval 30s (`:455`); reclaims a sweep that never creates a worktree / writes a checkpoint / logs past the spawn header.
- **mid-build-death watchdog** — `midbuild_watchdog_once` (`:3456`).
- **review-stall watchdog** — `review_stall_watchdog_once` (`:3627`), default 2700s (`:515`).

All three are spawned by a single task-spawn site, `sweep_registry.rs:4323-4385`, and that site is wired up in exactly one place: `loom-daemon/src/main.rs:771-795` — **for the default workspace registry only.**

Pooled (non-default) workspaces get their registry from `WorkspacePool::get_or_provision` (`loom-daemon/src/workspace_pool.rs:110`), which spawns **only a reaper** — `loom-daemon/src/workspace_pool.rs:143-157`. No startup watchdog, no mid-build watchdog, no review-stall watchdog.

## Consequence

In any multi-repo deployment, every repo except the default one loses all three self-healing paths. A sweep that wedges at startup in a secondary workspace is never reclaimed by the 300s watchdog; its issue stays at `loom:building`, which the work-finder correctly skips (`work_finder.rs:158`), so **that issue silently leaves the queue indefinitely** — with no quarantine, no comment, and no log line (the tick log is gated at `work_finder.rs:1067`/`:1249` and stays silent on skip-only ticks).

This directly undercuts #3887's mitigation, which is the documented remedy for startup wedges, and it is invisible in exactly the deployment shape the multi-workspace work-finder exists to serve. Note the operator here runs a 4-workspace canary rollout (loom + anvil/repo/kicad-tools + canaries), so the default workspace is the *minority* case.

## Suggested direction (not binding)

Hoist the watchdog spawn out of `main.rs` into the same place the reaper is spawned, so every registry gets the full set by construction rather than by a second call site remembering to. Worth checking whether any other per-registry background task has the same one-call-site shape.

Design questions for the curating pass:
- Do the watchdog config resolvers (`sweep_registry.rs:4138-4169`) read env/config per workspace, or would hoisting silently apply the default workspace's tuning to all of them? Precedence is env > config > default, and pooled workspaces already resolve stagger/quarantine per workspace (`workspace_pool.rs:133`, `:137`) — the watchdogs should match that shape.
- Are the extra spawned tasks per workspace acceptable at N workspaces, or should they share one timer? (See curator comment: all three backstops already run in **one** task, so this is 2 tasks/workspace, not 4.)

## Affected Files (starting points — verify before relying on)

- `loom-daemon/src/workspace_pool.rs:143-157` — reaper-only spawn; the gap
- `loom-daemon/src/main.rs:771-795` — the default-workspace-only watchdog wiring
- `loom-daemon/src/sweep_registry.rs:4323-4385` — the spawn site to hoist; resolvers `:4138-4169`
- `.loom/docs/daemon-reference.md` — watchdog documentation, if it claims per-workspace coverage

## Test Plan (to be firmed up by a Curator pass)

- [ ] Automated: a registry created via `WorkspacePool::get_or_create` has all three watchdogs running (assert on the spawned-task set or on observable reclaim behavior).
- [ ] Automated: a simulated no-progress sweep in a **pooled** workspace is reclaimed and its label restored `loom:building → loom:issue`.
- [ ] Automated: per-workspace watchdog config resolution is not collapsed to the default workspace's values.
- [ ] Manual: register a second workspace, wedge a sweep in it, confirm reclaim within the 300s window.

## Context

Related: #4007 (surfaced during its curation), #3887 (the startup-wedge remedy this silently disables), #4088 (watchdog timeout 120s→300s + per-`SweepId` progress latch).

Note `.loom/scripts` and `.loom/roles` are symlinks into `defaults/`; `.loom/docs` is a real directory.

---

## Curator re-verification (2026-07-28, pass 2) — against `origin/main` @ `5266454d`

> ⚠️ **This file trio has now rotted twice in ~48 hours.** The original report was pinned
> to `e6522b71`, pass 1 re-pinned to `2b1d1362`, and since then `38977a7a`, `3e33d821`, and
> `5266454d` have landed (`sweep_registry.rs` +399 lines). **Anchor on the symbols below,
> not the line numbers** — `get_or_provision`, `spawn_watchdog_task`, `spawn_reaper_task`,
> `_reaper`, `evict`, the five `resolve_*` fns. Line numbers are hints with a ~24h shelf life.

**The defect is confirmed still live and unfixed.** `WorkspacePool::get_or_provision`
(`workspace_pool.rs:124`) still spawns **only** `spawn_reaper_task` (`:160-175`).
`spawn_watchdog_task` has exactly **one** call site in the entire tree —
`main.rs:847`, inside the `let _watchdog_handle = if …` block at `:832-856`, for the
default workspace only. `PooledRegistry` still has only a `_reaper` field (`:58`).
Pooled workspaces still get a reaper and nothing else.

### What `5266454d` (#4141 / #4003) changed — mitigates the consequence, not the cause

`get_or_provision` now resolves `set_startup_proof_grace` per workspace (`:150`), so a
wedged pooled sweep **releases its global concurrency slot after ~90s** instead of holding
it for 300s. That narrows the blast radius but does not touch this issue: the wedged sweep
is still never cancelled, never re-dispatched, and its issue still sits at `loom:building`
forever. Update the "Consequence" narrative above accordingly — it no longer also starves
the concurrency cap.

Usefully, `set_startup_proof_grace` is a **fourth precedent** for the hoist shape, added
*after* this issue was filed: it consumes the already-bound `startup` inside
`get_or_provision`, exactly as the proposed watchdog hoist would.

### Corrected line references (supersedes both prior tables)

| Reference | Pass 1 (`2b1d1362`) | Now (`5266454d`) |
|---|---|---|
| `sweep_registry.rs` `watchdog_once` | `:3426` | **`:3580`** |
| `midbuild_watchdog_once` | `:3668` | **`:3822`** |
| `review_stall_watchdog_once` | `:3839` | **`:3993`** |
| `DEFAULT_WATCHDOG_TIMEOUT_SECS` (300) | `:454` | `:454` — unchanged ✅ |
| `DEFAULT_WATCHDOG_INTERVAL_SECS` (30) | `:457` | `:457` — unchanged ✅ |
| `DEFAULT_REVIEW_STALL_TIMEOUT_SECS` (2700) | `:517` | **`:545`** |
| `spawn_watchdog_task` (spawn site to mirror) | `:4535-4599` | **`:4711-4783`** |
| its single ticker loop | `:4551-4570` | **`:4726-4756`** |
| `resolve_watchdog_enabled` | `:4350` | **`:4513`** |
| `resolve_watchdog_timeout` | `:4360` | **`:4523`** |
| `resolve_watchdog_interval` | `:4373` | **`:4536`** |
| `resolve_review_stall_enabled` | `:4389` | **`:4552`** |
| `resolve_review_stall_timeout` | `:4399` | **`:4562`** |
| test precedent `resolve_watchdog_timeout_and_interval_precedence` | `:9881` | **`:10278`** |
| `main.rs` default-only wiring | `:816-841` | **`:832-856`** (call at `:847`) |
| `workspace_pool.rs::get_or_provision` | `:124` | `:124` — unchanged ✅ |
| seeded early-return | `:130-132` | **`:129-131`** |
| `read_startup_race_config` → `startup` | `:146` | `:146` — unchanged ✅ |
| `set_dispatch_stagger` | `:147` | `:147` — unchanged ✅ |
| `set_startup_proof_grace` (**new**, #4141) | — | **`:150`** |
| `set_quarantine_config` | `:151` | **`:154`** |
| `set_collision_detection` | `:155` | **`:158`** |
| reaper-only spawn (**the gap**) | `:158-169` | **`:160-175`** |
| `runtime.enter()` guard (trap 1) | `:160-163` | **`:163-166`** |
| module docs (runtime-guard rationale) | `:29-32` | **`:26-34`** |
| `PooledRegistry` struct (`_reaper` at `:58`) | `:52-59` | **`:53-59`** |
| `evict` | `:194` | **`:197`** |
| `evict` seeded guard (`_reaper.is_none()`) | `:201` | **`:204`** |
| `evict` "structural" doc comment | `:184` | **`:187`** |
| `evict` reaper abort (trap 2) | `:227-229` | **`:232-234`** |
| `evict` seeded-refusal test range | `:199-207` | **`:202-210`** |
| pool surface `new` / `start_safehouse_narration` / `seed` | `:77` / `:94` / `:105` | unchanged ✅ |
| pool surface `len` / `is_empty` | `:241` / `:250` | **`:244` / `:253`** |
| `work_finder.rs` `SKIP_LABELS` | `:158` | `:158` — unchanged ✅ |
| `work_finder.rs` tick-log gating | ⚠️ *"stale, no replacement"* | **resolved → `:1353`** |

**Tick-log gap now closed.** Pass 1 left this anchor unresolved. The gate is
`work_finder.rs:1353`: `report.dispatched > 0 || report.errors > 0 || report.skipped_quarantined > 0`.
Note `skipped_labeled` is deliberately **absent** from that disjunction — direct evidence for
the "skip-only ticks stay silent" claim in the narrative above.

### Design question 1 — ANSWERED: hoisting needs no new config plumbing

The watchdog resolvers read config **per workspace**, and the value is already in scope at
the hoist site. `get_or_provision` calls `read_startup_race_config(root)` (`:146`) and binds
`startup` — exactly the `&StartupRaceConfig` consumed by `resolve_watchdog_enabled` (`:4513`),
`resolve_watchdog_timeout` (`:4523`), `resolve_watchdog_interval` (`:4536`),
`resolve_review_stall_enabled` (`:4552`), `resolve_review_stall_timeout` (`:4562`).

So the hoist is **additive, not a refactor** — mirror the `main.rs:832-856` block using the
already-resolved `startup`. Per-workspace precedence (env > config > default) comes out
correct by construction, matching the shape `set_dispatch_stagger` (`:147`),
`set_startup_proof_grace` (`:150`), `set_quarantine_config` (`:154`), and
`set_collision_detection` (`:158`) already use.

### Design question 2 — ANSWERED: 1 extra task per workspace, not 3

`spawn_watchdog_task` (`:4711`) runs all three backstops inside a **single** ticker loop
(`:4726-4756`) — `watchdog_once`, then `midbuild_watchdog_once`, then
`review_stall_watchdog_once`, each holding the registry lock only briefly and never across
the sleep. Adding it to the pool is **+1 task per workspace** (2 total, alongside the
reaper). No shared-timer redesign is warranted, and `spawn_watchdog_task`'s signature is
unchanged, so no signature change is required.

### Enumeration check — the gap is confined to the watchdog

Pass 1 asked whether any other per-registry background task has the same one-call-site
shape. **Answered: no.** A full sweep of `pub fn spawn_*_task` across `loom-daemon/src`
yields 12 helpers; only two take an `Arc<Mutex<SweepRegistry>>` and are therefore
per-registry: `spawn_reaper_task` (`sweep_registry.rs:4390`) and `spawn_watchdog_task`
(`:4711`). The rest are either process-global (`spawn_heartbeat_task`) or already
multi-workspace-aware (`spawn_multi_work_finder_task`, `spawn_multi_main_health_gate_task`,
`spawn_multi_token_ranking_refresh_task`, `spawn_multi_role_task`, `spawn_watch_monitor_task`).
The watchdog is the only gap.

### Three traps not in the original report

1. **The spawn must be inside the runtime guard.** The reaper is spawned under
   `let _guard = self.runtime.enter();` (`workspace_pool.rs:163-166`) precisely because
   `get_or_provision` is also called from the epic supervisor's own current-thread runtime
   (module docs `:26-34`). A bare `tokio::spawn` for the watchdog outside that guard panics
   ("must be called from the context of a Tokio 1.x runtime") or lands on the wrong runtime.
   Hoist **inside** the existing guard.

2. **`evict` must abort the watchdog too, and the seeded-default guard is structural.**
   `PooledRegistry` holds `_reaper: Option<JoinHandle<()>>` (`:58`) and `evict` aborts it
   (`:232-234`). A hoisted watchdog needs a parallel `_watchdog` field aborted in the same
   place — otherwise evicting a workspace leaves an orphan watchdog ticking against a dead
   registry. **Careful:** `evict`'s seeded-default guard keys off `_reaper.is_none()`
   (`:204`), documented at `:187` as "structural: seeded entries carry `_reaper: None`".
   Adding a second handle field must not break that discriminator.

3. **Do not also hoist `main.rs`'s wiring — the default workspace must not get two
   watchdogs.** `seed()` (`:105`) inserts the default workspace with `_reaper: None` because
   `main` owns its tasks, and `get_or_provision` returns early on a seeded hit (`:129-131`).
   Confining the change to `get_or_provision` therefore leaves the default workspace with
   exactly the one watchdog `main.rs:832-856` already gives it. Removing the `main.rs` site
   "for symmetry" would need `seed()` to take over the spawn — a larger change than this
   issue needs.

### Corrected Affected Files

- `loom-daemon/src/workspace_pool.rs` — **the fix.** Spawn the watchdog in
  `get_or_provision` (`:160-175`, inside the `runtime.enter()` guard) using the already-bound
  `startup`; add a `_watchdog` field to `PooledRegistry` (`:53-59`) and abort it in `evict`
  (`:232-234`) without breaking the `_reaper.is_none()` seeded guard (`:204`).
- `loom-daemon/src/sweep_registry.rs` — **read-only.** `spawn_watchdog_task` (`:4711`) and
  the five resolvers (`:4513`, `:4523`, `:4536`, `:4552`, `:4562`) are reused as-is. No
  signature change is required.
- `loom-daemon/src/main.rs` — **leave `:832-856` alone** (see trap 3). Touch only if the
  builder finds a genuine reason.
- `defaults/docs/daemon-reference.md` — watchdog section: state that all three backstops now
  cover every provisioned workspace, not just the default. **Note:** cite the `defaults/`
  path, not `.loom/docs/daemon-reference.md` — that path is a symlink blob (mode `120000`)
  and is unreadable via `git show`.

### Corrected Test Plan

⚠️ The AC above names `WorkspacePool::get_or_create`, **which does not exist**. The pool's
public surface is `new` (`:77`), `start_safehouse_narration` (`:94`), `seed` (`:105`),
`get_or_provision` (`:124`), `evict` (`:197`), `len` (`:244`), `is_empty` (`:253`). Use
**`get_or_provision`**.

- [ ] Automated: a registry provisioned via `WorkspacePool::get_or_provision` has a live
      watchdog task (assert on observable reclaim behavior, or on the retained `_watchdog`
      handle — prefer behavior).
- [ ] Automated: a simulated no-progress sweep in a **pooled** workspace is reclaimed and its
      label restored `loom:building → loom:issue`.
- [ ] Automated: per-workspace watchdog config is **not** collapsed to the default
      workspace's values — provision two roots whose `.loom/config.json` set different
      `autonomous.watchdog` timeouts and assert each registry resolved its own. Precedent:
      `resolve_watchdog_timeout_and_interval_precedence` (`sweep_registry.rs:10278`).
- [ ] Automated: `evict` aborts the watchdog as well as the reaper (no orphan task), and still
      refuses to evict the seeded default workspace (`:202-210`).
- [ ] Automated: the seeded default workspace is **not** double-watchdogged — a
      `get_or_provision` for the default root returns the seeded entry and spawns nothing.
- [ ] Regression: `cd loom-daemon && cargo test workspace_pool` and `cargo test watchdog`
      stay green.
- [ ] Manual: register a second workspace, wedge a sweep in it, confirm reclaim within the
      300s window.

### Verdict

Scope, root cause, and acceptance criteria are **unchanged in substance** — this pass
re-pins stale anchors (for the second time), closes the last open anchor from pass 1, and
records that #4141 narrowed the consequence. `loom:issue` is retained per the re-curation
playbook. With both design questions answered and the enumeration closed, the change is a
well-bounded single-file edit plus tests; the `<!-- loom:complexity=complex -->` marker is
**not** warranted.

rjwalters:
## Summary

`WorkspacePool::get_or_provision` spawned only the background reaper for pooled (non-default) workspaces. The startup-hang / mid-build-death / review-stall watchdog (`spawn_watchdog_task`) had exactly one call site — `main.rs`, for the default workspace registry only. In any multi-repo deployment, every managed repo except the default one silently lost all three self-healing backstops: a sweep that wedges at startup in a secondary workspace was never reclaimed, and its issue stayed at `loom:building` indefinitely, invisible to the work-finder.

## Changes

- `loom-daemon/src/workspace_pool.rs`:
  - Hoist the watchdog spawn into `get_or_provision`, inside the existing `runtime.enter()` guard (required — a bare `tokio::spawn` there panics when provisioned from the epic supervisor's own current-thread runtime).
  - Reuse the already-resolved `startup: StartupRaceConfig` binding, so the five resolvers (`resolve_watchdog_enabled` / `resolve_watchdog_timeout` / `resolve_watchdog_interval` / `resolve_review_stall_enabled` / `resolve_review_stall_timeout`) resolve **per workspace** (env > config > default), matching the shape `set_dispatch_stagger` / `set_startup_proof_grace` already use. No new config plumbing.
  - Add a `_watchdog: Option<JoinHandle<()>>` field to `PooledRegistry` alongside `_reaper`, and abort it in `evict()` right where `_reaper` is aborted — no orphan task on `workspace remove`.
  - The `_reaper.is_none()` structural guard `evict()` uses to detect the seeded default workspace is untouched — a seeded entry now carries both `_reaper: None` and `_watchdog: None`.
  - `main.rs`'s existing default-workspace watchdog wiring is untouched (out of scope per curator guidance — the default workspace must not get double-watchdogged).
- `defaults/docs/daemon-reference.md`: documents that all three watchdog backstops now cover every provisioned workspace, not just the default, and that `evict` aborts both the reaper and the watchdog.

| A registry provisioned via `get_or_provision` has a live watchdog | ✅ | `provision_spawns_watchdog_alongside_reaper` asserts `_watchdog.is_some()` |
| A simulated no-progress sweep in a pooled workspace is reclaimed | ✅ | `pooled_workspace_watchdog_reclaims_a_hung_sweep` dispatches a hung fixture sweep and confirms the real async `spawn_watchdog_task` auto-cancels + re-dispatches it |
| Per-workspace watchdog config is not collapsed to the default workspace's values | ✅ | `per_workspace_watchdog_config_is_not_collapsed_to_default` writes two roots with distinct `autonomous.watchdog` config and confirms each resolves independently |
| `evict()` aborts the watchdog as well as the reaper, and still refuses the seeded default workspace | ✅ | `evict_aborts_watchdog_alongside_reaper` |
| The seeded default workspace is not double-watchdogged | ✅ | `seeded_default_workspace_is_not_double_watchdogged` |
| `cargo test workspace_pool` and `cargo test watchdog` pass | ✅ | See Test Plan |

## Test Plan

```
cd loom-daemon
cargo check --lib                       # clean
cargo fmt --check                       # clean
cargo clippy --lib --tests -- -D warnings   # clean, zero warnings
cargo test --lib workspace_pool         # 12 passed; 0 failed
cargo test --lib watchdog               # 21 passed; 0 failed
cargo test --lib                        # full suite: 1332 passed; 0 failed
```

The `pooled_workspace_watchdog_reclaims_a_hung_sweep` test drives the real production `spawn_watchdog_task` (the same function `get_or_provision` now wires up) against a fixture registry with a fake `spawn-claude.sh` that hangs, using a short timeout/interval, and confirms a fresh sweep is auto-dispatched for the same issue — proving the reclaim mechanism is genuinely live when spawned via the pool's wiring path. (`skip_label_flip = true` is used to avoid firing real `gh` calls against this repo's own issues in a test; `get_or_provision` itself always resolves `skip_label_flip = false` for real workspaces.)

Closes #4124

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -391,15 +391,29 @@ the pool when the request carries an explicit `workspace_root` (#3929), so all
 per-repo registries are observable/addressable via `list_sweeps` /
 `get_sweep_status` / `tail_sweep_log` / `cancel_sweep` / `dispatch_sweep`.
 
+**Watchdog coverage for every provisioned workspace (#4124)**: `get_or_provision`
+spawns the sweep watchdog (the startup-hang / mid-build-death / review-stall
+self-healing backstops described above) alongside the reaper for **every**
+pooled workspace, not just the default one — closing a gap where only the
+default workspace's registry (wired up once in `main.rs`) had a watchdog and
+every other managed repo got a reaper and nothing else, silently disabling the
+startup-hang recovery from #3887 in any multi-repo deployment. Each pooled
+workspace resolves its own watchdog config (enabled/timeout/interval/review-stall)
+from its own `.loom/config.json`, mirroring the dispatch-stagger and quarantine
+config already resolved per workspace — so the resolution is genuinely
+per-workspace, not a copy of the default workspace's tuning. The seeded default
+workspace is unaffected: `main.rs` still owns its one watchdog, and
+`get_or_provision` returns the seeded entry without spawning a second one.
+
 **Eviction on `workspace remove` (#3929)**: `DeregisterWorkspace` (the
 `workspace remove` CLI) calls `WorkspacePool::evict`, which drops the pooled
-registry and **aborts its background reaper task** so it does not leak. The
-**seeded default workspace is guarded** — it is owned by `main` and keeps serving
-default-workspace IPC requests, so evicting it is a no-op. A live sweep child in
-an evicted registry is **not** killed and its lock/log files are untouched; only
-the in-memory tracking + reaper go away, so the sweep finishes normally but its
-terminal state becomes unobservable via IPC after the deregister — an accepted
-consequence of an explicit operator `workspace remove`.
+registry and **aborts its background reaper and watchdog tasks** so neither
+leaks. The **seeded default workspace is guarded** — it is owned by `main` and
+keeps serving default-workspace IPC requests, so evicting it is a no-op. A live
+sweep child in an evicted registry is **not** killed and its lock/log files are
+untouched; only the in-memory tracking + reaper + watchdog go away, so the sweep
+finishes normally but its terminal state becomes unobservable via IPC after the
+deregister — an accepted consequence of an explicit operator `workspace remove`.
 
 ## Token pool provisioning for managed repos (#3938)
 
--- a/loom-daemon/src/workspace_pool.rs
+++ b/loom-daemon/src/workspace_pool.rs
@@ -618,7 +618,13 @@ mod tests {
         let out = registry
             .dispatch(&crate::types::SweepKind::Issue(999_001), None, None, None, None)
             .unwrap();
-        let original_id = out.sweep_id.clone();
+        // Assert reclaim on the child PID, not the sweep_id: `generate_sweep_id`
+        // mints IDs at second granularity (`Utc::now().timestamp()`), so this
+        // fast dispatch→watchdog→re-dispatch cycle frequently completes within
+        // one wall-clock second and the re-dispatched sweep gets an id identical
+        // to the original (#4124 flaky test). A freshly spawned child always has
+        // a distinct OS PID, so PID inequality is collision-proof.
+        let original_pid = out.pid;
 
         let arc = Arc::new(Mutex::new(registry));
         // Short timeout/interval so the test doesn't wait the production
@@ -638,7 +644,7 @@ mod tests {
                 .iter()
                 .any(|i| {
                     matches!(i.kind, crate::types::SweepKind::Issue(n) if n == 999_001)
-                        && i.sweep_id != original_id
+                        && i.pid != original_pid
                 })
         })
         .await;
--- a/loom-daemon/src/workspace_pool.rs
+++ b/loom-daemon/src/workspace_pool.rs
@@ -17,28 +17,34 @@
 //! [`WorkspacePool`] owns those registries: a lazily-populated, hot-reconciled
 //! cache keyed by workspace root. Both autonomous loops share **one** pool so a
 //! given repo has exactly **one** registry instance — unifying the in-flight
-//! dedup and the background reaper across the work-finder and epic supervisor
-//! (and the IPC `DispatchSweep` path for the default workspace, which is seeded).
+//! dedup and the background reaper + watchdog across the work-finder and epic
+//! supervisor (and the IPC `DispatchSweep` path for the default workspace,
+//! which is seeded).
 //!
-//! # Reaper threading
+//! # Reaper + watchdog threading
 //!
 //! Each provisioned registry gets its own background reaper
-//! ([`crate::sweep_registry::spawn_reaper_task`]). The pool is consumed from two
-//! different threads — the work-finder runs on the shared daemon Tokio runtime,
-//! the epic supervisor on its own dedicated OS thread with a private
-//! current-thread runtime. To guarantee every reaper runs on the **shared**
-//! daemon runtime regardless of which thread first provisions a workspace, the
-//! pool captures a [`tokio::runtime::Handle`] to the shared runtime at
-//! construction and enters it (`Handle::enter`) around the reaper spawn.
+//! ([`crate::sweep_registry::spawn_reaper_task`]) **and** its own watchdog
+//! ([`crate::sweep_registry::spawn_watchdog_task`], Issue #4124) — the
+//! startup-hang, mid-build-death, and review-stall self-healing backstops that,
+//! before #4124, ran only for the default workspace via `main`. The pool is
+//! consumed from two different threads — the work-finder runs on the shared
+//! daemon Tokio runtime, the epic supervisor on its own dedicated OS thread
+//! with a private current-thread runtime. To guarantee every reaper and
+//! watchdog runs on the **shared** daemon runtime regardless of which thread
+//! first provisions a workspace, the pool captures a [`tokio::runtime::Handle`]
+//! to the shared runtime at construction and enters it (`Handle::enter`)
+//! around both spawns.
 //!
 //! # Eviction (phase c, #3929)
 //!
 //! [`WorkspacePool::evict`] removes a provisioned registry when its workspace is
 //! deregistered (`workspace remove` / [`Request::DeregisterWorkspace`]), aborting
-//! its background reaper task so it does not leak for the daemon's lifetime. The
-//! **seeded default workspace** is guarded — its registry + reaper are owned by
-//! `main` and continue serving default-workspace IPC requests, so `evict` is a
-//! no-op for it (identified by `_reaper: None`).
+//! its background reaper and watchdog tasks so neither leaks for the daemon's
+//! lifetime. The **seeded default workspace** is guarded — its registry,
+//! reaper, and watchdog are owned by `main` and continue serving
+//! default-workspace IPC requests, so `evict` is a no-op for it (identified by
+//! `_reaper: None`).
 //!
 //! [`Request::DeregisterWorkspace`]: crate::types::Request::DeregisterWorkspace
 
@@ -49,13 +55,21 @@ use std::sync::{Arc, Mutex};
 use crate::event_bus::EventBus;
 use crate::sweep_registry::{self, SweepRegistry, SweepRegistryConfig};
 
-/// One cached per-workspace registry plus the reaper task keeping it reconciled.
+/// One cached per-workspace registry plus the background tasks keeping it
+/// reconciled and self-healing.
 struct PooledRegistry {
     registry: Arc<Mutex<SweepRegistry>>,
     /// The background reaper handle, kept alive for the daemon's lifetime.
     /// `None` for a **seeded** entry (the default workspace) whose reaper is
-    /// owned by `main` — the pool must not abort it.
+    /// owned by `main` — the pool must not abort it. This is also the
+    /// structural discriminator `evict` uses to detect the seeded default
+    /// workspace (`_reaper.is_none()`) — see the doc comment on `evict`.
     _reaper: Option<tokio::task::JoinHandle<()>>,
+    /// The background watchdog handle (Issue #4124), kept alive for the
+    /// daemon's lifetime. `None` for a **seeded** entry (the default
+    /// workspace) whose watchdog is owned by `main` — the pool must not
+    /// abort it.
+    _watchdog: Option<tokio::task::JoinHandle<()>>,
 }
 
 /// A lazily-populated, thread-safe cache of one [`SweepRegistry`] per workspace
@@ -110,6 +124,7 @@ impl WorkspacePool {
         map.entry(root).or_insert(PooledRegistry {
             registry,
             _reaper: None,
+            _watchdog: None,
         });
     }
 
@@ -158,11 +173,43 @@ impl WorkspacePool {
         registry.set_collision_detection(sweep_registry::resolve_collision_detection(root));
 
         let arc = Arc::new(Mutex::new(registry));
-        // Spawn the reaper on the shared daemon runtime regardless of which
-        // thread we are called from (the epic supervisor uses its own runtime).
-        let reaper = {
+        // Spawn the reaper and watchdog on the shared daemon runtime
+        // regardless of which thread we are called from (the epic supervisor
+        // uses its own runtime) — both spawns must stay inside this guard or
+        // they panic ("must be called from the context of a Tokio 1.x
+        // runtime") when provisioned from the epic supervisor's own
+        // current-thread runtime.
+        let (reaper, watchdog) = {
             let _guard = self.runtime.enter();
-            sweep_registry::spawn_reaper_task(arc.clone())
+            let reaper = sweep_registry::spawn_reaper_task(arc.clone());
+            // Sweep watchdog (Issue #4124): every provisioned (pooled)
+            // workspace now gets the same startup / mid-build-death /
+            // review-stall self-healing backstops the default workspace has
+            // had since #3887/#3895/#3910 — previously only the reaper was
+            // wired up here. Reuses the `startup` config already resolved
+            // above so per-workspace env > config > default precedence comes
+            // out correct by construction (mirrors `set_dispatch_stagger` /
+            // `set_startup_proof_grace` above).
+            let watchdog = if sweep_registry::resolve_watchdog_enabled(&startup) {
+                let timeout = sweep_registry::resolve_watchdog_timeout(&startup);
+                let interval = sweep_registry::resolve_watchdog_interval(&startup);
+                let review_stall_timeout = if sweep_registry::resolve_review_stall_enabled(&startup)
+                {
+                    Some(sweep_registry::resolve_review_stall_timeout(&startup))
+                } else {
+                    None
+                };
+                Some(sweep_registry::spawn_watchdog_task(
+                    arc.clone(),
+                    timeout,
+                    interval,
+                    review_stall_timeout,
+                ))
+            } else {
+                log::info!("workspace_pool: watchdog disabled for {} (#3887)", root.display());
+                None
+            };
+            (reaper, watchdog)
         };
         log::info!("workspace_pool: provisioned sweep registry for {}", root.display());
         map.insert(
@@ -170,6 +217,7 @@ impl WorkspacePool {
             PooledRegistry {
                 registry: arc.clone(),
                 _reaper: Some(reaper),
+                _watchdog: watchdog,
             },
         );
         arc
@@ -232,6 +280,9 @@ impl WorkspacePool {
                 if let Some(reaper) = pooled._reaper {
                     reaper.abort();
                 }
+                if let Some(watchdog) = pooled._watchdog {
+                    watchdog.abort();
+                }
                 log::info!("workspace_pool: evicted sweep registry for {}", root.display());
                 true
             }
@@ -258,11 +309,23 @@ impl WorkspacePool {
 #[cfg(test)]
 #[allow(clippy::unwrap_used)]
 mod tests {
+    use std::os::unix::fs::PermissionsExt;
+    use std::time::Duration;
+
     use super::*;
     use tempfile::tempdir;
 
     fn pool() -> Arc<WorkspacePool> {
         Arc::new(WorkspacePool::new(Arc::new(EventBus::new()), tokio::runtime::Handle::current()))
+    }
+
+    /// Write a `.loom/config.json` under `root` with the given JSON body
+    /// (mirrors `sweep_registry::tests::write_config`, reimplemented locally
+    /// since cross-module test helpers aren't shared).
+    fn write_config(root: &Path, body: &str) {
+        let loom = root.join(".loom");
+        std::fs::create_dir_all(&loom).unwrap();
+        std::fs::write(loom.join("config.json"), body).unwrap();
     }
 
     #[tokio::test]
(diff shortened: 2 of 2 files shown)


rjwalters:
❌ **Changes Requested — flaky new test**

## Code review

The core fix is sound and well-scoped:
- `spawn_watchdog_task` is hoisted into `get_or_provision` inside the existing `runtime.enter()` guard — correctly avoids the "must be called from the context of a Tokio 1.x runtime" panic when provisioned from the epic supervisor's own current-thread runtime.
- The five `resolve_watchdog_*` calls reuse the already-bound `startup: StartupRaceConfig`, so per-workspace config resolution (env > config > default) comes out correct by construction, matching `set_dispatch_stagger` / `set_startup_proof_grace`. No new config plumbing needed, confirmed by reading the call site.
- `_watchdog` field added to `PooledRegistry` and aborted in `evict()` alongside `_reaper`, without disturbing the `_reaper.is_none()` structural guard that detects the seeded default workspace — verified by reading both `evict()` and the seed path.
- `main.rs`'s existing default-workspace watchdog wiring (`:847`) is untouched, so the default workspace still gets exactly one watchdog, not two.
- `defaults/docs/daemon-reference.md` update is accurate and cites the right path (not the `.loom/docs/` symlink).
- `cargo fmt --check` and `cargo check --lib` are clean locally.

## Test Execution

Ran the PR's stated test plan locally in `.loom/worktrees/issue-4124`:

1. `cargo fmt --check` — ✅ clean
2. `cargo check --lib` — ✅ clean
3. `cargo test --lib workspace_pool` — ⚠️ **flaky**: `pooled_workspace_watchdog_reclaims_a_hung_sweep` fails intermittently

**Root cause found**: `generate_sweep_id` (`sweep_registry.rs:4535`) mints IDs as `sweep-issue-{n}-{Utc::now().timestamp()}` — **second-granularity**. The new test dispatches the original hung sweep, then drives a real watchdog tick (150ms timeout / 200ms interval) that cancels + re-dispatches it, all well within the same wall-clock second on a fast box. When that happens, the re-dispatched sweep gets the **exact same sweep_id string** as the original, so the test's success condition (`i.sweep_id != original_id`) never becomes true and the 10s `wait_for_async` loop times out.

I confirmed this directly — reran the isolated test (`cargo test --lib workspace_pool::tests::pooled_workspace_watchdog_reclaims_a_hung_sweep`) 10 times in a row with no other tests running concurrently: **4 passed, 6 failed** (~60% failure rate). Adding a debug logger (temporary, not committed) to a failing run showed the second dispatch/redispatch log lines sharing an identical `sweep-issue-999001-<same-ts>` ID, followed immediately by the watchdog's bounded-retry "giving up" path on the next tick (since `watchdog_retried` was already flagged for the issue) — confirming the ID collision, not a slow scheduler, is the cause. This is independent of the surrounding suite; even fully isolated it reproduces at a high rate. CI's "Rust Unit Tests" happened to land green on this run, but the ~50-60% local failure rate means it will flip red intermittently and burn Doctor/Judge cycles on a non-issue.

This is a defect in the **new test**, not in the production fix — `generate_sweep_id`'s second-granularity is pre-existing and out of scope here. Suggested fixes (any of these would do):
- Assert on something collision-proof instead of `sweep_id` inequality — e.g. compare the child `pid`, or track a dispatch counter, or assert `restarted > 0` on the return value if the test called `watchdog_once` directly instead of driving it through the real spawned task.
- Or space the timeout/interval so cancel+redispatch is very unlikely to land in the same clock second in practice (doesn't fully close the race, just narrows it — not a real fix).
- Or backdate the original dispatch's `started_at` (as other watchdog tests in `sweep_registry.rs` already do via a `backdate` helper) so the test doesn't depend on wall-clock timing at all, and add a small sleep to guarantee the redispatch call happens in a different second before asserting.

The other 4 new tests (`provision_spawns_watchdog_alongside_reaper`, `seeded_default_workspace_is_not_double_watchdogged`, `evict_aborts_watchdog_alongside_reaper`, `per_workspace_watchdog_config_is_not_collapsed_to_default`) are solid and passed consistently across all my runs.

Please fix the flaky assertion in `pooled_workspace_watchdog_reclaims_a_hung_sweep` and I'll take another look — the underlying watchdog-hoist fix itself looks ready to ship.

rjwalters:
Addressed the flaky-test feedback (production fix untouched).

**Root cause**: `pooled_workspace_watchdog_reclaims_a_hung_sweep` asserted reclaim via `i.sweep_id != original_id`. Because `generate_sweep_id` mints IDs at second granularity (`Utc::now().timestamp()`), the fast dispatch -> watchdog -> re-dispatch cycle (150ms timeout / 200ms interval) frequently completes within one wall-clock second, so the re-dispatched sweep got an id identical to the original and the success condition never became true (~60% failure in isolation).

**Fix** (judge option 1 — collision-proof assertion): assert on the child **PID** instead of the sweep_id. A freshly spawned child always has a distinct OS PID, so PID inequality is collision-proof and independent of wall-clock second boundaries. Captured `original_pid = out.pid` (from `DispatchOutcome`) and changed the closure to `i.pid != original_pid`. (Backdating `started_at` alone would not have fixed this — the collision is in the assertion, not in when the watchdog fires; the re-dispatch still lands in the same second.)

Test-only change, single file (`loom-daemon/src/workspace_pool.rs`, +8/-2).

**Flakiness check**: ran `pooled_workspace_watchdog_reclaims_a_hung_sweep` 20x in a loop with `--test-threads=1` -> **20/20 pass, 0 fail**.

**Regression check**:
- `cargo test --lib workspace_pool` -> 12 passed; 0 failed
- `cargo test --lib watchdog` -> 21 passed; 0 failed
- `cargo test --lib` (full suite) -> 1332 passed; 0 failed
- `cargo fmt --check` clean, `cargo clippy --lib --tests -- -D warnings` clean

Ready for re-review.

rjwalters:
✅ **Approved!** (re-review after Doctor's flaky-test fix)

Confirmed the fix in f9b38920 is sound:

- **Root-cause diagnosis is correct.** `generate_sweep_id` mints IDs at second granularity (`Utc::now().timestamp()`), so the fast dispatch→watchdog→re-dispatch cycle in this test can complete within one wall-clock second, making `sweep_id != original_id` an unreliable success condition.
- **PID-based assertion is collision-proof and correctly wired.** `original_pid` is captured from `out.pid` on the *initial* `DispatchOutcome`, before the watchdog task is even spawned. Traced `watchdog_once` (sweep_registry.rs:3756) → on `WatchdogDecision::Restart` it calls `self.cancel(...)` on the hung entry, then calls `self.dispatch(...)` fresh — a genuinely new `Child` process, hence a new OS PID (`DispatchOutcome.pid: u32`, not reused/latched state). The assertion `i.pid != original_pid` on the re-dispatched entry is therefore checking real process identity, not a coarse timestamp-derived ID.
- **Verified locally in the worktree** (`.loom/worktrees/issue-4124`, HEAD at f9b38920):
  - `pooled_workspace_watchdog_reclaims_a_hung_sweep` run 10x standalone (matches Doctor's 20x report) — 10/10 pass
  - `cargo test --lib workspace_pool` — 12/12 pass
  - `cargo test --lib watchdog` — 21/21 pass
  - `cargo test --lib` (full suite) — 1332/1332 pass
  - `cargo fmt --check` — clean
  - `cargo clippy --lib --tests -- -D warnings` — clean
- **Scope check**: diff vs main is still just `workspace_pool.rs` + `defaults/docs/daemon-reference.md` — the de-flake commit only touches the one test assertion, no other changes introduced.
- **CI**: all checks pass (`gh pr checks 4174`), `mergeStateStatus` is `CLEAN`.

The original production-code review (watchdog hoisted into `get_or_provision`, per-workspace resolver reuse, `evict()` aborting `_watchdog` alongside `_reaper`) stands — unaffected by this test-only fix.

## Test Execution

Test plan from PR description — all steps re-executed as above, all ✅.

Issue #3980, opened 2026-07-26
macOS TCC hygiene under launchd: daemon tree triggers folder-access popups; must not need (or request) Full Disk Access


rjwalters:
## Context

After migrating the daemon to a launchd LaunchAgent (#3972 incident), the operator got ~10 macOS TCC popups requesting file-access permissions. Under the old nohup-from-terminal model the daemon silently inherited the terminal app's TCC grants; as a launchd job it is its own responsible process, so any touch of a protected location (Desktop/Documents/Downloads, network/removable volumes, …) by the daemon **or any sweep child** prompts fresh.

Audit so far: the daemon core and `.loom/scripts` never reference protected folders (disk headroom is `df -Pk <workspace>`), so the prompts are almost certainly sweep children (claude sessions, test suites, or tool subprocesses) stepping outside the workspace roots. The daemon's legitimate working set — `~/GitHub/*` workspaces, `~/.loom`, `~/.claude*`, `/private/tmp` — contains **no** TCC-protected paths, so Full Disk Access is not needed and should never become the documented answer.

Why FDA is the wrong fix even as a convenience: TCC identity is code-signature/path based, and `loom-daemon` is an ad-hoc/unsigned binary that the self-update roll (#3968) rebuilds — grants would silently reset on every roll, producing a recurring popup storm *and* an over-grant.

## Suggested acceptance criteria
- [ ] Instrument/audit: identify which child operations touch TCC-protected paths (candidates: test suites writing to `~/Documents`, `find`/scan utilities crossing `$HOME`, tools resolving iCloud-synced paths) and scope them to workspace roots or `$TMPDIR`
- [ ] Sweep children get an explicit working-set contract (workspace root + `.loom` + scratch) documented in the role prompts / sweep skill, so out-of-bounds access is a reviewable defect, not ambient behavior
- [ ] Sign `loom-daemon` (ad-hoc with a stable identifier is enough) so any *intentional* future TCC grant survives self-update rebuilds
- [ ] daemon-reference documents the TCC posture: what to click when macOS prompts (Deny is safe; a resulting sweep failure identifies the out-of-bounds path), and why FDA is never required

rjwalters:
**Operator outcome + new evidence (2026-07-26):** the ~10 prompts included **Photos, Media & Apple Music, and iCloud** — and the operator clicked Allow on all of them.

Two implications for this issue:

[...]

PR #4180, opened 2026-07-28
feat(daemon): resolve launchd domain gui→user so the daemon starts headlessly over SSH

### LINKED ISSUE #4130: launchctl bootstrap gui/$UID fails over SSH — daemon cannot be (re)started remotely on macOS

rjwalters:
## Summary

`loom-daemon-start.sh` hardcoded the per-GUI-login `gui/<uid>` launchd domain, so `launchctl bootstrap gui/<uid> …` failed over SSH with no active GUI session (`error 125: Domain does not support specified action`), blocking a remote (re)start of the daemon on macOS. This adds a shared domain resolver that prefers `gui/<uid>` when it resolves (unchanged GUI behavior) and otherwise falls back to the SSH-reachable background per-user `user/<uid>` domain that `sshd` instantiates. All four lifecycle scripts consume the one shared definition so they always agree on the domain.

## Changes

- **New `defaults/scripts/lib/launchd-domain.sh`** — `resolve_launchd_domain()` with precedence: `LOOM_LAUNCHD_DOMAIN` override (honored verbatim) → probe `gui/<uid>` → fall back to `user/<uid>`. Documents the rejected alternatives (system `LaunchDaemon`, `launchctl asuser`).
- **`loom-daemon-start.sh`** — sources the resolver; replaces the hardcoded domain at the daemon bootstrap and the #4011 watchdog service lookup + bootstrap; header/help/env docs.
- **`loom-daemon-stop.sh` / `-update.sh` / `-watchdog.sh`** — source the same resolver at their service-lookup / bootout sites; resolution is gated on `USE_LAUNCHD` so a `LOOM_DAEMON_LAUNCHD=0` path never probes the machine-global launchd domain (#4078 symmetry).
- **Docs** — `daemon-reference.md` #3972 section gains a "launchd domain resolution (#4130)" subsection + trade-off table + login-keychain/TCC cross-references; #3980 gains a "no prompt under `user/<uid>`" note; `github-authentication.md` "out of scope" note refreshed (the gap is now closed).
- **Tests** — resolver unit tests (gui-present / headless fallback / override honored / non-resolving override honored verbatim) in `test-loom-daemon-start.sh`; a plist regression guard in `test-loom-daemon-launchd-plist.sh`; the update-test fixture copies the lib; sandbox-lib comment documents domain-agnostic stubbing.

1. **Single shared resolver consumed by all four scripts** — `resolve_launchd_domain()` in `lib/launchd-domain.sh`, sourced by all four. `grep -n 'gui/' defaults/scripts/cli/*.sh` returns only comment lines (0 hardcoded consumer sites). ✅
2. **Both launchd jobs (daemon + watchdog) use the resolved domain** — `provision_watchdog_job` now resolves and bootstraps into the same domain. ✅
3. **GUI host unchanged** — resolver picks `gui/<uid>`; the rendered plist carries no domain token (domain is load-time). Verified with a live scratch-labeled `gui/<uid>` start→stop cycle (job resolved under `gui/501`, cleanly booted out). ✅
4. **Headless (SSH) start completes; stop/update act on it** — resolver falls back to `user/<uid>` (verified via sandboxed launchctl-stub tests; a real headless-host cycle is a manual operator test-plan item — see below). ✅ (unit-tested)
5. **Docs cover mechanism, order, rejected alternatives, keychain/TCC consequences** — daemon-reference.md updated. ✅
6. **`--no-launchd` / Linux unchanged; #3972 not reintroduced** — still a launchd domain, never session-inherited nohup; Linux untouched. ✅
7. **`LOOM_LAUNCHD_DOMAIN` override exists, takes precedence, fails loud if it doesn't resolve** — honored verbatim with no further fallback; verified live (pinning `user/501` on this GUI host failed loudly at bootstrap rather than silently falling back to gui). ✅

## Test Plan

- Automated: `test-loom-daemon-{start,stop,update,watchdog,launchd-plist}.sh` — all pass (15 / 22 / 51 / 12 / 28).
- `shellcheck -S warning` clean on the lib + four scripts.
- **Live GUI regression cycle** (scratch label, isolated socket/marker, fake sleeper binary): start bootstrapped into `gui/501`, job resolved, stop cleanly booted it out; no scratch plist/job leaked; production `com.rjwalters.loom-daemon` untouched.
- **Live override validation**: `LOOM_LAUNCHD_DOMAIN=user/501` on this GUI-active host fails loudly at bootstrap (I/O error — the gui session owns the domain), confirming both that the override is honored (no silent fallback) and *why* the resolver prefers gui when present rather than blanket-switching to user.
- **Not validated on this machine**: the real headless / no-GUI-session `user/<uid>` bootstrap path (this dev host has an active Aqua session). Covered by the sandboxed launchctl-stub tests; a real SSH-only-host cycle remains a manual operator test-plan item.

Closes #4130

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2010,9 +2010,11 @@ process:
   default),
 - locates the `loom-daemon` binary (`LOOM_DAEMON_BIN` → `PATH` → `target/{release,debug}`),
 - runs the **advisory** host-sleep check (`check-host-sleep.sh`, #3350) — never blocks the start,
-- **on macOS, backgrounds the daemon as a `gui/<uid>` LaunchAgent** instead of a
-  plain `nohup … &` (#3972 — see "macOS session-bootstrap hazard" below); on
-  Linux it stays a plain nohup background job,
+- **on macOS, backgrounds the daemon as a launchd LaunchAgent** in the resolved
+  per-user domain (`gui/<uid>` with an active GUI login, else the SSH-reachable
+  `user/<uid>` background domain — #4130) instead of a plain `nohup … &` (#3972 —
+  see "macOS session-bootstrap hazard" below); on Linux it stays a plain nohup
+  background job,
 - backgrounds the daemon and writes a PID file at `.loom/.daemon.pid` (gitignored),
 - refuses a second start when the PID file points at a live process, and surfaces
   the daemon's own **singleton-guard** refusal (#3806) — if the backgrounded
@@ -2198,17 +2200,57 @@ Linux does not have this failure mode — a systemd user session (or a plain
 identity to the shell that spawned it.
 
 **Fix.** `loom-daemon-start.sh` now generates a `launchd` LaunchAgent plist and
-loads it with `launchctl bootstrap gui/<uid>` instead of `nohup`-backgrounding
-in-process (Darwin only — Linux is unaffected and keeps the plain nohup path).
-This was validated during the incident itself: relaunching the identical
-binary as a launchd agent immediately restored `gh`/`git` — the first tick
-after migration reported `13 seen, 3 dispatched, 0 error(s)`.
+loads it with `launchctl bootstrap <domain>` (the resolved per-user domain — see
+"launchd domain resolution" below) instead of `nohup`-backgrounding in-process
+(Darwin only — Linux is unaffected and keeps the plain nohup path). This was
+validated during the incident itself: relaunching the identical binary as a
+launchd agent immediately restored `gh`/`git` — the first tick after migration
+reported `13 seen, 3 dispatched, 0 error(s)`.
+
+**launchd domain resolution (#4130) — gui/<uid> ↦ user/<uid>.** The daemon was
+originally loaded into the hardcoded `gui/<uid>` domain — the per-GUI-login
+(Aqua) domain owned by `loginwindow`. Over SSH with **no active GUI login
+session** that domain does not exist, so `launchctl bootstrap gui/<uid> …` fails
+with `error 125: Domain does not support specified action`, and the daemon could
+not be (re)started remotely on a headless Mac. The shared resolver
+`resolve_launchd_domain()` (`defaults/scripts/lib/launchd-domain.sh`, consumed by
+`loom-daemon-start.sh` / `-stop.sh` / `-update.sh` / `-watchdog.sh` so the whole
+lifecycle agrees on one domain) resolves in this order:
+
+1. an explicit `LOOM_LAUNCHD_DOMAIN` override, honored **verbatim** (a pinned
+   domain that does not resolve fails loudly at `bootstrap`, never falls back
+   further — the override is honored, not advisory);
+2. `gui/<uid>` when `launchctl print gui/<uid>` resolves (a live GUI login) —
+   **byte-for-byte the pre-#4130 default** on a host with an Aqua session;
+3. `user/<uid>` — the **background per-user** domain that `sshd` itself
+   instantiates: present over SSH with no GUI, running as the *user* (not root),
+   and `gui/<uid>` already forwards to it (`endpoint destination =
+   …domain.user.<uid>`).
+
+**Rejected alternatives** (documented, deliberately not implemented):
+
+| Mechanism | Runs as | Needs GUI session? | Needs sudo? | Why rejected |
+|---|---|---|---|---|
+| `gui/<uid>` (pre-#4130 default) | user | **yes** | no | Fails over SSH — the defect this fixes. Kept as the preferred domain **when it resolves**. |
+| **`user/<uid>` (chosen fallback)** | user | no | no (own uid) | Background session ⇒ TCC prompts can't be answered interactively (see #3980) and the login keychain may be locked (see #4005). Smallest change, closest privilege match. |
+| `launchctl asuser <uid>` | user | effectively yes (targets a session) | yes | Still needs a target session to impersonate. |
+| system `LaunchDaemon` | **root** by default | no | yes | Wrong privilege model and **no login-keychain session** — re-opens the #4005 headless-credential problem in a new place instead of closing one. |
+
+**Consequences of a non-Aqua (`user/<uid>`) domain.** A background domain cannot
+surface an interactive TCC prompt, so any touch of a protected folder by the
+daemon or a sweep child fails silently rather than prompting — see
+`### macOS TCC hygiene under launchd (#3980)` below (the daemon's legitimate
+working set never needs a protected folder, so this is a diagnostic signal, not a
+blocker). The operator's **login keychain** may also be locked in a headless
+session, so export a `GH_TOKEN` for forge auth rather than relying on the keychain
+(the #4005 credential preflight reports this loudly). Both are inherent to running
+without an Aqua session, not regressions introduced here.
 
 - **Plist location & label**: `~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist`
-  (override the label with `LOOM_LAUNCHD_LABEL`). Regenerated and reloaded
-  (`bootout` the old definition, then a fresh `bootstrap` + `kickstart -k`) on
-  **every** start, so a later invocation's flags/env always win over a stale
-  loaded definition.
+  (override the label with `LOOM_LAUNCHD_LABEL`; override the domain with
+  `LOOM_LAUNCHD_DOMAIN`). Regenerated and reloaded (`bootout` the old definition,
+  then a fresh `bootstrap` + `kickstart -k`) on **every** start, so a later
+  invocation's flags/env always win over a stale loaded definition.
 - **Environment forwarding**: the plist's `PATH` is the current `PATH` plus a
   fallback set (`~/.local/bin`, `~/.cargo/bin`, Homebrew, standard bin dirs) so
   `gh`, `git`, `cargo`, and `python3` resolve even inside launchd's minimal
@@ -2247,8 +2289,12 @@ after migration reported `13 seen, 3 dispatched, 0 error(s)`.
   during the stop window. See [Supervised restart primitive](#supervised-restart-primitive-4054)
   below and `docs/design/supervised-restart-primitive.md`.
 - **Escape hatch**: `--no-launchd` (or `LOOM_DAEMON_LAUNCHD=0`) forces the
-  legacy nohup path even on Darwin — e.g. for a sandboxed/headless macOS runner
-  with no GUI login session where `gui/<uid>` may not resolve.
+  legacy nohup path even on Darwin — e.g. for a sandboxed macOS runner where no
+  launchd domain is usable at all. Note that a headless/SSH-only session **no
+  longer requires** this escape hatch just to start: `resolve_launchd_domain()`
+  falls back to `user/<uid>` (above), so the durable launchd path (and its #3972
+  Mach-bootstrap-namespace protection) is available over SSH too — nohup remains
+  only the last-resort opt-out, not the headless default.
 - **Inspection without side effects**: `--print-plist` renders the exact plist
   XML this invocation would install and exits — no `launchctl` call, no file
   write to `~/Library/LaunchAgents`. Useful for auditing exactly what
@@ -2267,8 +2313,9 @@ after migration reported `13 seen, 3 dispatched, 0 error(s)`.
 **Why launchd changed the TCC picture.** Under the pre-#3972 nohup model, the
 daemon inherited whatever TCC (Transparency, Consent, and Control) grants the
 launching terminal app already had — so folder-access prompts, if any, belonged
-to Terminal.app/iTerm/Claude Code, not to `loom-daemon`. As a `gui/<uid>`
-LaunchAgent (see above), the daemon is its **own** TCC-responsible process:
+to Terminal.app/iTerm/Claude Code, not to `loom-daemon`. As a launchd LaunchAgent
+(see above — `gui/<uid>` under a GUI login, else `user/<uid>` over SSH, #4130),
+the daemon is its **own** TCC-responsible process:
 any touch of a protected location (`~/Desktop`, `~/Documents`, `~/Downloads`,
 `~/Pictures` (Photos), `~/Music` (Media & Apple Music),
 `~/Library/Mobile Documents` (iCloud Drive), network/removable volumes, …) by
@@ -2278,6 +2325,16 @@ including Photos / Media & Apple Music / iCloud — evidence of something
 enumerating the top level of `$HOME` itself rather than touching those folders
 individually (macOS bundles the per-category checks into one burst when a
 process lists `$HOME`'s immediate contents).
+
+**Under the `user/<uid>` (headless/SSH) domain (#4130) there is no prompt at
+all.** A background domain cannot surface an interactive TCC dialog, so the same
+out-of-bounds access **fails silently** (permission-denied) instead of prompting.
+That does not change the contract below — the daemon's and sweep children's
+legitimate working set still contains no protected folder — it only changes the
+symptom (a silent file-not-found rather than a popup). The remedy is identical:
+fix the offending script/tool to stay within the working-set contract, never a
+broad grant. This is one of the documented consequences of the non-Aqua fallback
+domain (see "launchd domain resolution" under #3972 above).
 
 **The daemon's legitimate working set never needs a protected folder.** Audited
 surfaces — the daemon core (Rust) and `.loom/scripts/*` — only ever touch
--- a/defaults/docs/github-authentication.md
+++ b/defaults/docs/github-authentication.md
@@ -142,14 +142,22 @@ file to mode `0600` whenever it carries a `GH_TOKEN`/`GITEA_TOKEN`/
 `FORGE_TOKEN`, so a local user other than the daemon's owner cannot read the
 PAT out of `~/Library/LaunchAgents`.
 
-**Out of scope**: `launchctl bootstrap gui/$UID` (the launchd domain
-`loom-daemon-start.sh` uses on macOS) fails over SSH — `gui/$UID` is a
-per-GUI-session domain that does not exist in an SSH session — so a daemon
-that isn't already running cannot be *started* remotely on macOS today even
-with a valid token exported. That is a supervision-model gap (a `LaunchDaemon`
-or `launchctl asuser` mechanism), not a credential problem; restarting an
-*already-running* daemon (`loom-daemon-update.sh`) and running fully headless
-on Linux (`--no-launchd` / nohup path) both work with an exported token today.
+**Starting the daemon headlessly over SSH (#4130 — resolved).** Earlier this
+section noted that `launchctl bootstrap gui/$UID` (the domain
+`loom-daemon-start.sh` originally hardcoded on macOS) fails over SSH with
+`error 125: Domain does not support specified action`, because `gui/$UID` is a
+per-GUI-login domain that does not exist in an SSH session — so a not-yet-running
+daemon could not be *started* remotely on macOS. That gap is now closed: the
+shared resolver `resolve_launchd_domain()` prefers `gui/<uid>` when a GUI login
+is active and otherwise falls back to the background per-user `user/<uid>` domain
+that `sshd` instantiates (running as the user, not root). So a headless / SSH-only
+`loom-daemon-start.sh` now completes, and stop/update find the resulting job. Pin
+the domain with `LOOM_LAUNCHD_DOMAIN` if needed; the rejected alternatives
+(a root system `LaunchDaemon`, `launchctl asuser`) and the login-keychain / TCC
+consequences of the non-Aqua domain are covered in
+[`daemon-reference.md` → "launchd domain resolution (#4130)"](daemon-reference.md).
+As before, export a `GH_TOKEN` for forge auth in a headless session (the login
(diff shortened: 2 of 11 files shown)


rjwalters:
✅ **Approved!**

Thorough, well-scoped fix for the launchd domain hardcoding bug (#4130). Verified all seven review focus areas:

1. **GUI-session regression guard holds** — `test-loom-daemon-launchd-plist.sh` asserts the rendered plist carries no `gui/<uid>`/`user/<uid>` domain token at all (domain is a `launchctl` load-time argument, never baked into the plist XML). `resolve_launchd_domain()` prefers `gui/<uid>` when it resolves, which is byte-for-byte the pre-#4130 default.
2. **All 4 lifecycle scripts share one resolver** — ran `grep -n 'gui/' defaults/scripts/cli/*.sh` myself: only comment lines remain, zero hardcoded consumer sites. `start`/`stop`/`update`/`watchdog` all source `defaults/scripts/lib/launchd-domain.sh` and call `resolve_launchd_domain()`.
3. **Watchdog LaunchAgent provisioned into the resolved domain** — `provision_watchdog_job()` (start.sh:345) now calls `resolve_launchd_domain()` instead of a hardcoded `gui/<uid>`, closing exactly the silent-miss gap the issue flagged.
4. **`--no-launchd`/nohup and Linux untouched** — diffed the nohup branch: the actual `nohup "$DAEMON_BIN" ... &` invocation is unchanged, only nearby comments updated. `stop.sh`/`update.sh` correctly gate domain resolution behind `USE_LAUNCHD` so a `LOOM_DAEMON_LAUNCHD=0` path never probes the machine-global launchd domain (#4078 symmetry).
5. **Docs updated, not stale** — `daemon-reference.md` gained a full "launchd domain resolution (#4130)" section (precedence order, rejected-alternatives table, TCC/keychain consequences); `github-authentication.md`'s "out of scope" note was rewritten to reflect the gap is closed and cross-references the new section.
6. **Test suites pass** — ran all five locally: start 15/15, stop 22/22, update 51/51, watchdog 12/12, launchd-plist 28/28 — matches the PR's claim exactly. `shellcheck -S warning` clean on the lib + all four scripts.
7. **`LOOM_LAUNCHD_DOMAIN` fails loud, not silent** — the resolver honors an override verbatim with no further fallback (unit-tested), and the actual loud failure surfaces at the existing `launchctl bootstrap` error path (`err` + `exit 1`), which the builder validated live by pinning a non-resolving domain on their GUI host.

On the flagged scope boundary (no live no-GUI-session `user/<uid>` bootstrap validation): acceptable, not blocking. It's inherently unverifiable on a dev box with an active Aqua session; the sandboxed launchctl-stub tests deterministically exercise the exact fallback branch (domain-agnostic stub, confirmed by reading `launchd-sandbox.sh`'s updated comment), and the live override test indirectly proves the no-silent-fallback contract. Flagged transparently as a manual operator test-plan item rather than glossed over.

## Test Execution

- `test-loom-daemon-start.sh` — ✅ 15/15 pass
- `test-loom-daemon-stop.sh` — ✅ 22/22 pass
- `test-loom-daemon-update.sh` — ✅ 51/51 pass
- `test-loom-daemon-watchdog.sh` — ✅ 12/12 pass
- `test-loom-daemon-launchd-plist.sh` — ✅ 28/28 pass
- `shellcheck -S warning` on lib + 4 scripts — ✅ clean
- Live GUI regression cycle / live override validation — ⚠️ not re-executed by Judge (destructive to the live daemon); accepted per builder's reported results and cross-checked against the automated coverage above
- Real headless `user/<uid>` bootstrap on an actual no-GUI-session host — ⏭️ not executed (requires a host without an active Aqua session); acceptable scope boundary per above, backed by sandboxed unit coverage

CI: all 7 checks pass, `mergeStateStatus: CLEAN`. The `.loom/config.json` diff visible in `git diff main..HEAD` is pure branch staleness (main gained `autonomous.collisionDetection` in #4166 after this branch's base) — this PR's single commit never touches that file, so a three-way merge/squash correctly preserves main's addition.

PR #4182, opened 2026-07-28
feat(daemon): advertise cross-host soft claims over safehouse (#4028, Phase 1)

### LINKED ISSUE #4028: Reduce cross-host sweep collisions: advertise work-in-progress via a shared room (soft claim) alongside the forge-label claim

rjwalters:
Closes #4028

## What this does

Phase 1 of #4028: each daemon **advertises** a work-in-progress soft claim into
the shared safehouse room the moment it decides to dispatch, and **consumes**
peer advertisements so a peer daemon backs off before the non-atomic
`loom:issue → loom:building` label flip would otherwise let it race. This shrinks
the cross-host collision window; it is a **soft claim / fast backoff, NOT a
mutex**. Phase 2 (an atomic cross-host claim authority) is explicitly out of
scope and should be filed separately.

## Design (addresses the two curator-identified gaps)

- **Gap 1a — dedicated inbound read task.** A new coordination task on its *own*
  safehouse connection drains the socket continuously via `select!`, so an
  **idle** daemon that emits no narration still observes peer claims promptly. It
  is NOT piggybacked on `read_reply` (which only reads while emitting). Inbound
  events route through a generic `InboundEventSink` (kept generic so future
  inbound human-steering can reuse the same read task).
- **Gap 2 — no fifth envelope type.** Claim ads ride a **`task`** envelope (the
  `type` enum is closed and owned by the safehouse repo) with the bare issue
  number as `task_id` and a structured `loom_claim`-marked JSON `body`.

New pure module `loom-daemon/src/peer_claims.rs` (socket-free, unit-testable):
`PeerClaimView` with TTL expiry, self-claim recognition, retraction, and
`ClaimAd` parse/serialize — mirroring the `decide`/`plan` split in
`claim_reconciliation.rs`.

## Acceptance criteria — all implemented

- [x] Inbound consumption via a **dedicated** read task (Gap 1a), not `read_reply`.
- [x] Advertisement `type` is `task` (Gap 2); carries issue, repo, host, PID, ts.
- [x] Gating on `safehouse::resolve_config(repo_root).enabled` (not block presence).
- [x] `safehouse.enabled` false/absent ⇒ byte-for-byte no-op (no view, channel, task, socket).
- [x] Advertise at the dispatch decision point, **before** the label flip.
- [x] Peer-claim view + TTL; skip in `tick`/`tick_multi` under its **own** distinct
      `peer-claim-skip` counter (never folded into #4085's collision count), on the summary line.
- [x] TTL expiry (default **120s** = 2× the 60s work-finder tick), measured against
      **local receipt** not the advertiser's clock; peer `exited`/`crashed` retracts early.
- [x] Fail-open: unreachable socket / malformed envelope / full channel → logged once, dispatch proceeds.
- [x] Advertisement is best-effort/non-blocking (bounded `try_send` off the dispatch path).
- [x] Self-claim recognition — a daemon never backs off on its own ad (host-identity match).
- [x] Explicit host identity: `host_identity()` (`LOOM_HOST_ID` > `$HOSTNAME` > `hostname`),
      derived, restart-stable; carried in the body since safehoused's `from` is persona-scoped.
- [x] No new bus topic — claims travel entirely over the safehouse room.
- [x] Docs: `.loom/docs/safehouse.md` (fixed the emit-only statements + new section) and
      `defaults/docs/daemon-reference.md` (new section) — TTL, fail-open, envelope type, soft-claim caveat.

## Phase-0 (#4085) collision counters — NOT measured (stated explicitly per the AC)

I **could not** capture a real before/after collision delta: this is a single-host
environment with no second daemon and no genuinely shared multi-host backlog, so
there is no cross-host race to observe and `#4085`'s `collisions` counter stays 0
regardless of this change. Per the issue's non-negotiable, I am stating this
rather than claiming an unmeasured win. The soft-claim mechanism is instead
validated by the deterministic unit/integration tests below (two-registries /
one-room, idle-daemon read, TTL, self-recognition, fail-open). A real two-host
measurement remains an operator follow-up.

## Tests (all pure/deterministic — no sleeps for ordering, per #4044)

- `peer_claims`: TTL expiry (local-clock), self-claim recognition, retraction (+ host scoping),
  malformed-envelope rejection, JSON round-trip, repo scoping, and the **two-hosts/one-issue**
  integration shape (exactly one proceeds) via injected ordering.
- `safehouse`: **idle-daemon still reads an inbound peer ad** (the Gap 1a regression guard —
  fails against a `read_reply`-piggyback impl), envelope-validity against `build_send_request`,
  disabled no-op (no task/socket), fail-open reconnect + clean exit on sender drop.
- `work_finder`: `tick` skips a peer-claimed issue under the distinct counter and stops once
  it clears; slot not consumed.
- `sweep_registry`: **dispatch proceeds when the safehouse channel is closed (fail-open)** — the
  single most important test — and the peer-claim skip set reflects the attached view.

`cd loom-daemon && cargo test --lib` → **1346 passed; 0 failed**. `cargo clippy --all-targets`
clean, `cargo build` clean.

### Note on `tests/integration_basic.rs`

The 9 `integration_basic` tests fail on this host with "Daemon failed to create
socket within 5s" — a **pre-existing, environmental** flake, not caused by this
change. The harness spawns the debug daemon and fails if its IPC socket (bound
last, at `main.rs` ~:1250, after gh-calling startup reconciliation) is not up
within 5s. Verified directly: the **baseline `main` binary** takes **8.9s** to
bind the socket under current host load vs this branch's **7.7s** — so the
timeout is exceeded on both, and this branch is marginally faster (my
peer-coordination insertion is a no-op when safehouse is disabled). The affected
code paths (terminal IPC / socket creation) are untouched by this PR.

## Files
`loom-daemon/src/peer_claims.rs` (new), `safehouse.rs`, `sweep_registry.rs`,
`work_finder.rs`, `workspace_pool.rs`, `main.rs`, `lib.rs`; `defaults/config.json`
(`peerClaimTtlSecs`); `.loom/docs/safehouse.md`; `defaults/docs/daemon-reference.md`.

### DIFF
--- a/.loom/docs/safehouse.md
+++ b/.loom/docs/safehouse.md
@@ -6,15 +6,22 @@ adds an optional, additive **narration** side-channel: an end-to-end-encrypted
 Matrix room a human watches in Element to follow a multi-host agent fleet in
 real time, instead of polling `gh` or tailing daemon logs.
 
-Phase 1 is **daemon-side, emit-only**. The `loom-daemon` subscribes its existing
-in-process event bus and narrates sweep-lifecycle transitions into the room. It
-adds no new event topics and no new publish call sites.
+Phase 1 **narration** is daemon-side and emit-only: the `loom-daemon` subscribes
+its existing in-process event bus and narrates sweep-lifecycle transitions into
+the room, adding no new event topics and no new publish call sites. On top of
+that, **peer-claim coordination (#4028) makes the room bidirectional** — a
+dedicated read task consumes inbound peer advertisements so daemons on a shared
+backlog back off before the non-atomic `loom:building` label flip would let them
+race. See [Peer-claim coordination](#peer-claim-coordination-cross-host-soft-claim-4028)
+below.
 
-> **Out of scope for phase 1** (tracked separately): per-worker personas
-> (`loom_builder_42`) and `SAFEHOUSE_PERSONA` forwarding to workers → **#3999**;
-> inbound human steering (reading `@`-mentions back to agents) → follow-up;
-> cloud-host provisioning of `safehoused` → **#3998**; carrying the judge verdict
-> value in an event payload (needs a frozen-taxonomy amendment) → follow-up.
+> **Out of scope** (tracked separately): per-worker personas (`loom_builder_42`)
+> and `SAFEHOUSE_PERSONA` forwarding to workers → **#3999**; inbound **human**
+> steering (reading `@`-mentions back to agents) → follow-up (it reuses the same
+> inbound read task #4028 adds); cloud-host provisioning of `safehoused` →
+> **#3998**; carrying the judge verdict value in an event payload (needs a
+> frozen-taxonomy amendment) → follow-up; the **atomic cross-host claim
+> authority** (a real CAS behind the soft claim) → Phase 2 of #4028.
 
 ## The degradation contract (read this first)
 
@@ -112,16 +119,78 @@ by the bare issue number (`task_id`):
   identity — the client never sends one (no impersonation).
 - Replies echo the request `id`. **Async push lines are interleaved on the same
   connection, carry an `event` key, and have no `id`** — the client
-  demultiplexes by skipping any line with an `event` key. Phase 1 is emit-only,
-  so pushes are read off the wire and discarded.
+  demultiplexes by skipping any line with an `event` key. The **narration**
+  connection is emit-only and discards inbound pushes; the **peer-claim
+  coordination** connection (#4028) instead routes each inbound `event` line to a
+  handler (see below).
 
 ## Implementation
 
 - `loom-daemon/src/safehouse.rs` — config resolver, envelope-v1 client, the
-  event→envelope mapping, and the reconnecting bus-subscriber sink.
-- `loom-daemon/src/workspace_pool.rs` — `start_safehouse_narration()` subscribes
-  the shared `Arc<EventBus>` (the single place it is owned) and spawns the sink
-  on the daemon runtime; a no-op when disabled.
+  event→envelope mapping, the reconnecting bus-subscriber narration sink, and
+  (#4028) the peer-claim coordination task + `InboundEventSink`.
+- `loom-daemon/src/peer_claims.rs` — the pure, socket-free peer-claim view
+  (TTL expiry, self-claim recognition, retraction, `ClaimAd` parse/serialize).
+- `loom-daemon/src/workspace_pool.rs` — `start_safehouse_narration()` and
+  `start_peer_coordination()` subscribe/attach the shared `Arc<EventBus>` /
+  `PeerClaimView` and spawn on the daemon runtime; both no-ops when disabled.
+
+## Peer-claim coordination: cross-host soft claim (#4028)
+
+On a multi-host deployment the only cross-host claim signal is the forge label,
+whose `loom:issue → loom:building` flip is **not** compare-and-swap
+(`SweepRegistry::flip_label_to_building` is an unconditional
+`--remove-label`/`--add-label`) — two hosts can both read `loom:issue` and
+dispatch before either flip propagates, producing duplicate sweeps. Peer-claim
+coordination shrinks that window:
+
+- **Advertise.** At the dispatch decision point — right after the local claim
+  lock, **before** the label flip — the daemon publishes a claim advertisement
+  over the room: issue number, [repo slug](#), host identity, PID, and a
+  wall-clock timestamp, carried as a **`task`** envelope (the `type` enum is
+  closed and owned by the safehouse repo, so a claim rides `task` with the bare
+  issue number as `task_id` — **no fifth type is invented**) whose `body` is the
+  structured JSON payload (marked `loom_claim`).
+- **Consume.** A **dedicated inbound read task** — separate from the narration
+  sink — drains the socket continuously via `select!`, so an **idle** daemon that
+  emits no narration still observes peer claims promptly (the narration
+  connection only reads while it is emitting). Each inbound claim is folded into a
+  shared `PeerClaimView`.
+- **Back off.** The work-finder skips any issue with a live peer claim, counted
+  under its **own** distinct `peer-claim-skip` reason on the per-tick summary line
+  (never folded into #4085's collision count).
+- **TTL.** Every peer claim expires after **`safehouse.peerClaimTtlSecs`
+  (default 120s = 2× the 60s work-finder interval)**, so a crashed peer cannot
+  permanently starve an issue. The TTL clock is the **local receipt `Instant`**,
+  never the advertiser's wall clock (clock skew is not comparable across hosts).
+  A peer also **retracts** its claim early when its sweep exits/crashes (a
+  `retract`-kind ad emitted from the reaper), freeing the issue before the TTL.
+- **Host identity.** loom's single, explicit host-identity concept is
+  `sweep_registry::host_identity()` (`LOOM_HOST_ID` > `$HOSTNAME` > the `hostname`
+  binary > `unknown-host`) — derived, not a new config block, and stable across
+  restarts. safehoused stamps the socket `from` from the *persona* (all daemons
+  share `loom_daemon`), which cannot distinguish hosts, so the identity travels in
+  the claim body and is what powers self-claim recognition: **a daemon never backs
+  off on its own advertisement.**
+- **Event taxonomy.** The internal pub/sub topic taxonomy is frozen; peer claims
+  add **no new bus topic** — they travel entirely over the safehouse room.
+
+### Soft claim, NOT a mutex (the load-bearing caveat)
+
+A room broadcast is eventually consistent, so this is a **fast backoff, not a
+lock**: two hosts advertising near-simultaneously still race. Advertisement
+*shrinks* the collision window; it does not close it. The atomic authority for
+the final claim — a real cross-host CAS (e.g. a `git push` to a claim ref) — is
+**Phase 2 of #4028**, deliberately out of scope here.
+
+### Fail-open (never a liveness dependency)
+
+Coordination is best-effort end to end: an unreachable/refusing/timing-out
+`safehoused` socket, a malformed inbound envelope, or a full outbound channel is
+logged (once) and **dispatch proceeds normally**. The outbound advertisement is a
+bounded, non-blocking `try_send` off the dispatch path; a `Full`/`Closed` channel
+drops the ad. `safehouse.enabled` false/absent is a **byte-for-byte no-op**: no
+view, no channel, no coordination task, no socket.
 
 # Phase 2 — worker-side `safehouse-mcp` injection (#3999)
 
--- a/defaults/config.json
+++ b/defaults/config.json
@@ -17,7 +17,8 @@
     "room": null,
     "persona": "loom_daemon",
     "workerPersonas": [],
-    "mcpCommand": "safehouse-mcp"
+    "mcpCommand": "safehouse-mcp",
+    "peerClaimTtlSecs": 120
   },
   "health_monitoring": {
     "enabled": true,
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -1509,7 +1509,9 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.watchdog.reviewStall` | `LOOM_SWEEP_REVIEW_STALL` | `true` | Review-phase stall watchdog on/off (#3910) |
 | `autonomous.watchdog.reviewStallTimeoutSecs` | `LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS` | `2700` | Log-silence window before a hung Judge/Doctor sweep is re-dispatched |
 | `autonomous.collisionDetection.enabled` | `LOOM_DETECT_COLLISIONS` | `false` | Cross-host dispatch-collision baseline (#4085). Off by default — adds one extra `gh issue view --json labels` round-trip per dispatch. Detection only: a collision is logged/counted, never acted on |
-| *(host identity, records only)* | `LOOM_HOST_ID` | `$HOSTNAME` → `hostname` → `unknown-host` | Overrides this host's identity string in collision log records (#4085); set it where the daemon runs without `$HOSTNAME` exported |
+| `safehouse.enabled` | `LOOM_SAFEHOUSE_ENABLED` | `false` | Enables safehouse fleet-comms (#3997) **and** cross-host soft-claim coordination (#4028). Off by default — a byte-for-byte no-op (no socket, no coordination task) when unset |
+| `safehouse.peerClaimTtlSecs` | `LOOM_PEER_CLAIM_TTL_SECS` | `120` | Peer-claim TTL, in seconds (#4028) — how long a peer's soft claim suppresses local dispatch (measured against local receipt, not the advertiser's clock). Default = 2× the 60s work-finder tick |
+| *(host identity)* | `LOOM_HOST_ID` | `$HOSTNAME` → `hostname` → `unknown-host` | This host's identity string, used in collision log records (#4085) **and** peer-claim self-recognition (#4028); set it where the daemon runs without `$HOSTNAME` exported |
 | `autonomous.autoUpdate.enabled` | `LOOM_AUTO_UPDATE` | `false` | Autonomous self-update loop on/off (#4055). **Opt-in** (it rebuilds + restarts the daemon process). Exactly one loop per daemon, not a per-workspace fan-out. See [Autonomous self-update loop](#autonomous-self-update-loop-4055) below |
 | `autonomous.autoUpdate.intervalSecs` | `LOOM_AUTO_UPDATE_INTERVAL_SECS` | `900` | Cadence between staleness checks. Zero/invalid → default |
 | `autonomous.autoUpdate.settleSecs` | `LOOM_AUTO_UPDATE_SETTLE_SECS` | `600` | Settle window: wait this long after first observing a stale commit — resetting on every further commit — before rolling, so a burst of merges collapses into one roll. Zero/invalid → default |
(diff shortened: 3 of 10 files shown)


rjwalters:
Judge: rebased branch onto latest main (was DIRTY/CONFLICTING with 6 commits behind) and force-pushed with --force-with-lease. mergeStateStatus is now UNSTABLE/MERGEABLE. Proceeding with full evaluation while CI runs.

rjwalters:
✅ **Approved!**

## Rebase / merge state
This PR was **DIRTY/CONFLICTING** (6 commits behind main) when I picked it up. I rebased the branch onto current `main` (clean, no conflicts) and force-pushed with `--force-with-lease`. `mergeStateStatus` is now **CLEAN** / `mergeable: MERGEABLE`.

## CI
All required checks pass on the rebased head:
- Rust Linting & Build Check (`cargo clippy --package loom-daemon -D warnings ...`) — pass
- Rust Unit Tests — pass
- Installer Integration Tests — pass
- CLAUDE.md Line Budget — pass

## Local verification
- `cd loom-daemon && cargo build` — clean, no warnings.
- `cargo test --lib` — **1397 passed, 0 failed** (grew from the PR's claimed 1346 because of intervening commits picked up by the rebase).

## Test Plan execution (from the issue's/PR's Test Plan)
- Peer-claim TTL expiry (measured from local receipt, not advertiser clock) — ✅ `peer_claims::tests::claim_expires_after_ttl_measured_from_local_receipt`
- Self-claim recognition — ✅ `own_advertisement_is_never_backed_off_on`
- Retraction on peer exited/crashed (host-scoped) — ✅ `peer_retract_frees_the_issue_early`, `a_host_cannot_retract_another_hosts_claim`; wired via `emit_event`'s `SweepExited`/`SweepCrashed` match in `sweep_registry.rs`
- Malformed-envelope rejection — ✅ `malformed_bodies_are_rejected_not_panicked` (bad JSON, wrong marker/version, missing fields, empty host, unknown kind)
- `tick` peer-claim skip under a **distinct** counter, clears after TTL — ✅ `work_finder::tests::test_tick_peer_claim_skipped_under_distinct_counter` / `test_tick_stops_skipping_once_peer_claim_clears` (verified not folded into `skipped_labeled`/`skipped_in_flight`)
- Idle-daemon regression guard (Gap 1a — fails against a `read_reply`-piggyback) — ✅ `safehouse::tests::idle_daemon_still_reads_inbound_peer_ad`, verified by inspection this is a genuinely separate `select!`-based read task, not `read_reply`
- Fail-open (socket absent/refusing/timing out ⇒ dispatch proceeds) — ✅ `sweep_registry::tests::dispatch_proceeds_when_peer_claim_channel_is_closed_fail_open` and `safehouse::tests::coordination_reconnects_when_socket_absent_and_exits_on_sender_drop`
- Envelope validity (`type` ∈ `KNOWN_TYPES`, `task_id` matches `[A-Za-z0-9_]`) — ✅ `claim_ad_serializes_to_a_valid_task_envelope`, asserted against real `build_send_request`
- Disabled no-op (byte-for-byte, no socket/task/channel) — ✅ `disabled_config_spawns_no_coordination_task`
- Two-hosts/one-issue integration shape, deterministic injected ordering (no sleeps, #4044) — ✅ `two_hosts_one_issue_second_host_backs_off_deterministically`
- Manual two-host test + before/after #4085 collision counters — ⚠️ Not executed (requires a second host / genuinely shared backlog); the PR is explicit about this rather than claiming an unmeasured win, per the issue's non-negotiable. Acceptable as an operator follow-up.

## Code review notes
- `peer_claims.rs` is pure/socket-free as required, cleanly mirrors the `claim_reconciliation.rs` decide/plan split, with thorough doc comments explaining the local-receipt TTL rationale and the soft-claim-not-a-mutex caveat.
- `safehouse.rs`'s `run_coordination` correctly uses a dedicated `select!` loop over read+write on its own connection — confirmed this is structurally distinct from the narration sink's `read_reply`-driven path, which was the exact regression this issue exists to prevent.
- Advertisement happens at the correct point in `dispatch()` — right after the claim lock, before `flip_label_to_building` — and is a bounded non-blocking `try_send`.
- Envelope type correctly rides the existing closed `task` enum (no fifth type invented), host identity is explicit and derived (`LOOM_HOST_ID` > `$HOSTNAME` > `hostname`), and the frozen event-bus taxonomy is untouched (no new topic).
- Docs (`safehouse.md`, `daemon-reference.md`) are updated in both places the issue called out, including fixing the stale "Phase 1 is emit-only" statements.
- Diff scope matches the PR's stated file list exactly — no unrelated changes.

Excellent adherence to a very detailed curated issue. Nice work.

PR #4203, opened 2026-07-28
fix(daemon): render a deterministic PATH into the launchd plist

### LINKED ISSUE #4172: loom-daemon-start.sh bakes the invoking shell's entire PATH into the re-rendered launchd plist
## Context

Observed 2026-07-28 during the first supervised daemon roll (`loom-daemon-update.sh --relaunch` → `loom-daemon-start.sh` re-render). The `--relaunch` path correctly preserved all 7 `LOOM_*`/token env vars from the live plist (#4118 fix working as designed) — but **PATH is excluded from preservation by design, and the re-render replaced the previous minimal pinned PATH with the invoking interactive shell's full PATH** (~20 entries: `.bun`, `.juliaup`, lean-genius bins, pnpm, solana, foundry, X11, TeX, a rust-analyzer plugin cache, plus several duplicated entries).

Before the roll the plist carried:

```
PATH = /Users/rwalters/.local/bin:/Users/rwalters/.cargo/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

After, it carries the whole interactive environment. Consequences:

- The daemon's tool resolution now depends on **whose shell ran the roll** — non-hermetic and non-reproducible across hosts/sessions.
- Unrelated toolchains (project-specific bins earlier in PATH) can shadow the binaries the daemon and its sweep children expect.
- Every future re-render can silently change daemon behavior again.

- [ ] The rendered plist's PATH is deterministic: either a pinned canonical minimal PATH (e.g. the 8-entry form above), or the previous live plist's PATH preserved the same way `LOOM_*` vars are.
- [ ] An explicit override (env var or flag) exists for hosts that genuinely need extra PATH entries, and the render logs the PATH it chose.
- [ ] `--print-plist` output makes a PATH change from the live plist visible (diff-friendly), so a drift is noticed at roll time.

rjwalters:
## Summary

`loom-daemon-start.sh` baked `"$PATH:<fallback>"` into the rendered launchd
plist — the invoking shell's entire interactive PATH prefixed onto a
canonical fallback set — so every re-render (e.g. `loom-daemon-update.sh
--relaunch`) silently replaced whatever PATH the live plist carried with
whoever's shell happened to run the roll. This makes the fix deterministic:
the rendered PATH is now a pinned canonical minimal set by default, with an
explicit opt-in override for hosts that genuinely need more, and
`--print-plist` flags a PATH change from the previously installed plist.

## Changes

- Added `resolve_plist_path()` to `defaults/scripts/cli/loom-daemon-start.sh`:
  resolves, in priority order, `LOOM_DAEMON_PATH` (full override, used
  verbatim), `LOOM_DAEMON_PATH_EXTRA` (prepended onto the canonical minimal
  PATH), or the canonical minimal PATH alone (`~/.local/bin`, `~/.cargo/bin`,
  Homebrew, standard bin dirs — the same fallback set as before, just no
  longer prefixed with the invoking shell's PATH). Computed once per
  invocation into `$PLIST_PATH_VALUE` and always logged to stderr.
- Both `render_launchd_plist()` (the daemon job) and `render_watchdog_plist()`
  (the watchdog job) now consume the same `$PLIST_PATH_VALUE` instead of each
  independently prefixing `$PATH`.
- `--print-plist` now also diffs the freshly-rendered PATH against any
  previously installed live plist for the same label
  (`~/Library/LaunchAgents/<label>.plist`) and prints a diff-friendly `- live`
  / `+ new` "PATH DRIFT DETECTED" warning to stderr when they differ — purely
  read-only, no side effects, so drift is visible at roll/inspect time instead
  of being silently swapped in.
- Updated the stale rationale comment in `loom-daemon-update.sh`'s
  `harvest_plist_env()` (it previously said "start.sh rebuilds PATH from the
  live shell", which is no longer accurate).
- Added new documentation for `LOOM_DAEMON_PATH` / `LOOM_DAEMON_PATH_EXTRA` to
  the script's usage/env header comment.

| Rendered plist's PATH is deterministic (pinned canonical minimal PATH, or the previous live plist's PATH preserved like `LOOM_*`) | ✅ | Default now resolves to the canonical 8-entry minimal PATH with no shell-PATH prefix (`resolve_plist_path()`); new test confirms a marker directory injected into the invoking shell's `PATH` does NOT leak into the rendered plist |
| Explicit override (env var or flag) exists for hosts that need extra PATH entries, and the render logs the PATH it chose | ✅ | `LOOM_DAEMON_PATH` (full override) and `LOOM_DAEMON_PATH_EXTRA` (extend canonical) env vars added; every render logs its chosen PATH (and which case produced it) to stderr — covered by new tests |
| `--print-plist` output makes a PATH change from the live plist visible (diff-friendly), so drift is noticed at roll time | ✅ | `--print-plist` now compares against `~/Library/LaunchAgents/<label>.plist` when present and prints a `- live` / `+ new` warning on mismatch, silent when they match — covered by new tests using an isolated scratch `HOME` |

## Test Plan

- `./defaults/scripts/tests/test-loom-daemon-launchd-plist.sh` — 42/42 passing (28 pre-existing + 14 new, covering default determinism, `LOOM_DAEMON_PATH`, `LOOM_DAEMON_PATH_EXTRA`, stderr logging, and the drift check both with and without a live plist)
- `./defaults/scripts/tests/test-loom-daemon-start.sh` — 15/15 passing (no regression)
- `./defaults/scripts/tests/test-loom-daemon-update.sh` — 51/51 passing, including the pre-existing `--relaunch ... does NOT round-trip the stale PATH` regression test
- `./defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 12/12 passing
- `./defaults/scripts/tests/test-loom-daemon-stop.sh` — 22/22 passing
- `shellcheck` clean on all three modified files
- `bash -n` syntax check clean

Closes #4172

### DIFF
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -74,6 +74,14 @@
 #                        user/$(id -u)); honored verbatim, else auto-resolved
 #                        gui→user (#4130). A pinned domain that does not resolve
 #                        fails loudly at bootstrap rather than falling back.
+#   LOOM_DAEMON_PATH        Full override for the rendered plist's PATH (#4172).
+#                        Used verbatim -- no canonical fallback is appended. For
+#                        a host that needs a wholly custom PATH.
+#   LOOM_DAEMON_PATH_EXTRA  Extra dir(s) to prepend onto the canonical minimal
+#                        PATH (#4172) instead of overriding it entirely -- for
+#                        a host that needs one or two additional dirs (e.g. a
+#                        project-local toolchain) without inheriting the WHOLE
+#                        invoking shell's interactive PATH.
 #
 # Exit codes:
 #   0  daemon started (or already running)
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge Review: APPROVE

Reviewed the branch (`feature/issue-4172` @ `a9d6b26b`), the full diff, and ran all five affected test suites plus a manual `--print-plist` sanity check.

### Acceptance criteria (issue #4172)

1. **Deterministic PATH by default** — verified. `resolve_plist_path()` renders the pinned canonical 8-entry PATH with no shell-PATH prefix. Manually confirmed a marker directory injected into the invoking shell's `PATH` does not leak into the rendered plist.
2. **Explicit override + logging** — verified. `LOOM_DAEMON_PATH` (full override, verbatim) and `LOOM_DAEMON_PATH_EXTRA` (prepend onto canonical) both work as documented, confirmed manually and by tests 11-12. Every render logs its chosen PATH (and which branch produced it) to stderr, never stdout, so `--print-plist`'s XML stays pipeable.
3. **`--print-plist` drift visibility** — verified. The drift check is read-only (no writes), diffs against `~/Library/LaunchAgents/<label>.plist` when present, and prints a diff-friendly `- live` / `+ new` "PATH DRIFT DETECTED" block only on mismatch (silent when they match).

### Correctness details checked

- `PLIST_PATH_VALUE` is computed exactly once per invocation (line 487, after `DAEMON_BIN` resolution, before both the `--print-plist` branch and the real launchd-render path later in the script) — confirmed via `grep`, matches the documented "logged once per run" contract.
- Both `render_launchd_plist()` and `render_watchdog_plist()` now consume the same `$PLIST_PATH_VALUE`, so daemon and watchdog plists never diverge.
- `extract_plist_path_value()` is a narrow, correctly-scoped awk extraction matched to the exact two-line `<key>PATH</key>`/`<string>...</string>` shape this script itself emits — reasonable to avoid a `plutil`/`jq` dependency for a read-only drift check.
- `loom-daemon-update.sh`'s `harvest_plist_env()` comment update accurately reflects the new PATH-exclusion rationale; the pre-existing "does NOT round-trip the stale PATH" regression test still passes.

### Test results (all match the PR's reported numbers)

- `test-loom-daemon-launchd-plist.sh`: 42/42 passing (28 pre-existing + 14 new)
- `test-loom-daemon-start.sh`: 15/15 passing
- `test-loom-daemon-update.sh`: 51/51 passing
- `test-loom-daemon-watchdog.sh`: 12/12 passing
- `test-loom-daemon-stop.sh`: 22/22 passing
- `shellcheck` clean on all three modified files
- `bash -n` syntax check clean on all three
- Manual `--print-plist` spot check: default render correctly excludes an injected shell-PATH marker; `LOOM_DAEMON_PATH=/custom/bin` renders `/custom/bin` verbatim

### CI

All checks green (`gh pr checks 4203`), `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

No issues found. This is a clean, well-tested, well-documented fix that fully satisfies all three acceptance criteria from #4172.

PR #4204, opened 2026-07-28
fix(scripts): fail-closed disk-headroom probe instead of a fake 0 GB

### LINKED ISSUE #4164: disk-headroom.sh returns a plausible 0 GB instead of erroring when its dependency fails to load
## Context

`.loom/scripts/lib/disk-headroom.sh` provides the resource gate that resolves `/loom:sweep`'s auto wave size (#3566). When sourced from a **non-bash** shell it fails silently and reports **0 GB free**, which the wave-size math then treats as genuine disk pressure.

## Observed (rjwalters/repo, 2026-07-27, macOS, zsh)

```
$ source ./.loom/scripts/lib/disk-headroom.sh
$ loom_worktree_root_free_gb "$(git rev-parse --show-toplevel)"
loom_worktree_root_free_gb:9: command not found: loom_worktree_root
0
```

Actual free space on that volume: **273 GB**.

The cause is ordinary and not itself a bug: line 46 does `source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"`, and `_LOOM_DISK_HEADROOM_LIB_DIR` is derived from `${BASH_SOURCE[0]}`, which is empty under zsh. So the dependency never loads. Re-running the same call under `bash -c` returns the correct `273` and a wave size of 3. **Operator error, correctly diagnosed — the library is a bash library.**

## The actual defect: the failure is silent and the fallback value is plausible

`loom_worktree_root_free_gb` returns `0` rather than erroring. That `0` flows into `loom_wave_size_from_disk`, which computes `floor(0 / LOOM_PER_WORKTREE_GB) = 0`, clamps to the documented floor of 1, and reports reason `floor`. The skill then logs, per its own table:

> `wave size 1, mechanism=<m>: reduced to 1 (only 0 GB free on <worktree-root>)`

That message is indistinguishable from a genuinely full disk. An operator reading it has no way to tell that the number is an artifact of a missing function rather than a real measurement — and every sweep silently drops to fully-sequential wave size 1, which looks like a deliberate safety response rather than a broken probe.

A resource gate whose degraded mode is "return the most conservative plausible number, quietly" is hard to debug precisely because it looks like it is working.

- [ ] `loom_worktree_root_free_gb` distinguishes "measured 0 GB" from "could not measure". Returning a sentinel (empty / non-zero exit) rather than `0` is enough.
- [ ] `loom_wave_size_from_disk` treats an unmeasurable value as **unknown**, not as zero — either fall back to the mechanism's target with a warning, or fail loudly. Do not silently clamp to 1 on a measurement error.
- [ ] The library detects that it is not running under bash and says so, rather than proceeding with an unset `_LOOM_DISK_HEADROOM_LIB_DIR`. A one-line guard (`[ -n "${BASH_VERSION:-}" ] || { echo "…requires bash" >&2; return 1; }`) covers it.
- [ ] Sourcing with an unresolvable lib dir is an error, not a no-op — the missing `source` currently fails open.
- [ ] Add a unit case to `test-disk-headroom.sh` for the unmeasurable branch, asserting it does not resolve to wave size 1 by way of a fake 0.
- [ ] Consider whether the skill's one-line reason should name the measured path and value in a way that makes an implausible reading obvious (`0 GB free on /Users/... (273 GB expected?)` is not achievable, but distinguishing `unknown` from `0` in the log text is).

## Related

- #3566 — introduced the auto wave size and the disk gate
- `.loom/scripts/lib/worktree-root.sh` — the dependency that fails to load

---

### Verification against origin/main

All core claims verified, with one correction and one scope addition:

1. **The fail-open paths are confirmed live** in `defaults/scripts/lib/disk-headroom.sh`. `loom_worktree_root_free_gb` has two `echo "0"; return 0` branches (missing `repo_root` arg; df failure / malformed output). When the `loom_worktree_root` dependency is undefined, `wt_root` comes back empty, the ancestor walk no-ops, `df -Pk ""` fails, and the same `echo "0"` branch fires. `loom_wave_size_from_disk` then computes `floor(0/2)=0` and floors to `1|floor` — exactly the "indistinguishable from a full disk" outcome described above.

2. **Correction — the zsh path-resolution failure itself is already fixed on main (#3680).** The current file uses `${BASH_SOURCE[0]:-$0}`, which is deliberately bash+zsh-portable *because* sweep.md's Stage -1 sources this file directly into the invoking shell (often zsh on macOS). The observed repro in rjwalters/repo was a stale installed copy predating #3680 (`resync-installed.sh` is the consumer-side remedy). **Suggested AC item 3 ("the library detects that it is not running under bash") must NOT be implemented as written** — a bash-version gate would break the intentional zsh-sourcing design. Implement it as a fail-closed dependency guard instead (see refined AC below).

3. **Scope addition — the Rust daemon has the same defect by design.** `loom-daemon/src/disk_headroom.rs::worktree_root_free_gb` returns `0` on any df/command failure, with comments explicitly noting it matches "the bash `echo \"0\"` fallbacks". Whatever unknown-vs-zero contract this fix establishes must be mirrored there (or explicitly deferred with a comment + follow-up issue) so the two implementations don't silently diverge.

4. **Helpful existing behavior**: `loom_wave_size_from_disk` already validates its inputs and `return 2`s on non-integers. The consumer is not the gap — the producer (`loom_worktree_root_free_gb`) manufacturing a fake valid `0` is. This makes the caller-side fix clean (below).

### Refined acceptance criteria

Items 1, 2, 5, and 6 of "Suggested acceptance criteria" above stand as written. Items 3 and 4 are superseded by:

- [ ] **Fail-closed dependency load (replaces item 3)**: keep the file zsh-sourceable; after the `source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"` line, fail loudly if the dependency did not load — e.g. `source … || return 1` plus `command -v loom_worktree_root >/dev/null 2>&1 || { echo "disk-headroom.sh: failed to load worktree-root.sh from '$_LOOM_DISK_HEADROOM_LIB_DIR'" >&2; return 1; }`. (`return` at the top level of a sourced file works in both bash and zsh; because the function definitions follow the `source` line, aborting there means `loom_worktree_root_free_gb` is never defined — the caller's next invocation fails loudly instead of quietly returning 0.)
- [ ] **Unknown ≠ zero, resolved caller-side (sharpens item 4 / body item 2)**: `loom_worktree_root_free_gb` prints nothing on stdout and returns non-zero (with a stderr message naming the probed path) on any failure: missing arg, unresolvable worktree root, df failure or malformed output. A genuine measured `0` (full disk) still prints `0` with exit 0. `loom_wave_size_from_disk`'s pure-integer contract stays **unchanged** — its existing non-integer rejection already refuses an empty value; the fallback policy lives in the caller:
- [ ] **sweep.md Stage -1** ("Resolve auto wave size", the `FREE_GB="$(loom_worktree_root_free_gb …)"` snippet): capture the exit status; on failure, skip the disk clamp (`K = min(target, candidates)`) and log a distinct line, e.g. `wave size K, mechanism=<m>: disk headroom unknown (probe failed on <worktree-root>) — disk clamp skipped`. Add the corresponding row to the reason/message table so `unknown` is never rendered as `only 0 GB free`.
- [ ] **Rust parity**: change `worktree_root_free_gb` in `loom-daemon/src/disk_headroom.rs` to return `Option<u64>` (None = unmeasurable), and have `disk_headroom_limit` skip the disk clamp with a log line on None — or, if deferred, add an explicit comment in the Rust file and file a follow-up issue in the PR body.

### Affected Files

- `defaults/scripts/lib/disk-headroom.sh` — fail-closed dependency source; `loom_worktree_root_free_gb` returns empty/non-zero on probe failure instead of `echo "0"`
- `defaults/scripts/tests/test-disk-headroom.sh` — add unmeasurable-branch cases (df stubbed to fail → non-zero exit + empty stdout; missing-dependency source fails loudly; regression: no probe failure resolves to `1|floor`; measured-0 still returns `0`/exit 0)
- `defaults/.claude/commands/loom/sweep.md` — Stage -1 auto-wave-size snippet handles probe failure; new "unknown" row in the log-message table
- `loom-daemon/src/disk_headroom.rs` — mirror the unknown-vs-zero contract (`Option<u64>` + caller fallback in `disk_headroom_limit`), or explicit deferral comment
- `loom-daemon/src/disk_headroom.rs` unit tests (same file, `#[cfg(test)]`) — cover the None path
- `loom-daemon/tests/sweep_md_stage_minus_one_doc_lint.rs` — reference: run after sweep.md edits (it pins Stage -1 structural tokens incl. `loom_wave_size_from_disk`); update only if it pins changed text

**Path note**: consumer repos see the scripts under `.loom/scripts/…` as installed copies/symlinks — edit only the `defaults/` sources above.

## Test Plan

- [ ] Automated: `bash defaults/scripts/tests/test-disk-headroom.sh` — existing matrix plus the new unmeasurable-branch cases all pass
- [ ] Automated: `cd loom-daemon && cargo test disk_headroom && cargo test --test sweep_md_stage_minus_one_doc_lint`
- [ ] Manual (zsh regression, #3680): on macOS, `zsh -c 'cd <repo> && source ./.loom/scripts/lib/disk-headroom.sh && loom_worktree_root_free_gb "$(git rev-parse --show-toplevel)"'` prints the real free-GB figure — zsh sourcing must keep working
- [ ] Manual (failure injection): with a PATH-shimmed `df` that exits 1, `loom_worktree_root_free_gb` exits non-zero with empty stdout and a stderr message; nothing downstream reports `0 GB free`
- [ ] Edge case: a df that genuinely reports 0 blocks available still yields `0` with exit 0, and the wave size still floors to `1|floor` — real disk pressure must remain distinguishable from a broken probe

rjwalters:
## Summary

`loom_worktree_root_free_gb` (bash) and `worktree_root_free_gb` (Rust) previously
returned a plausible `0` on any measurement failure — a missing bash-only
dependency, a missing `repo_root` argument, or a `df` failure/malformed
output — indistinguishable from a genuine full-disk reading of `0`. That fake
`0` flowed into the wave-size math and floored `/loom:sweep`'s auto wave size
to 1 with reason `floor`, silently masquerading a broken probe as real disk
pressure. This PR makes "unmeasurable" a distinct, loud outcome from "measured
zero" in both the bash and Rust implementations.

## Changes

- `defaults/scripts/lib/disk-headroom.sh`
  - `loom_worktree_root_free_gb` now prints nothing and returns non-zero (with
    a stderr message naming the probed path) on any failure to measure: missing
    `repo_root` argument, unresolvable worktree root, or a failing/malformed
    `df`. A genuine measured `0` free GB (a truly full disk) still prints `0`
    with exit 0.
  - The `source .../worktree-root.sh` dependency load is now fail-closed: if
    the sibling library doesn't load (so `loom_worktree_root` never gets
    defined), the file aborts via `return 1` *before* defining
    `loom_worktree_root_free_gb` — turning "quietly proceeds with an unset
    dependency" into "loudly fails to source at all". (The library stays
    zsh-sourceable per #3680 — no `$BASH_VERSION` gate was added, since that
    would break the intentional zsh-sourcing design `sweep.md` Stage -1 relies
    on.)
- `defaults/.claude/commands/loom/sweep.md` (Stage -1, "Resolve auto wave
  size")
  - Captures the probe's exit status as `DISK_PROBE_OK`. On success, behavior
    is unchanged (feeds `FREE_GB` into `loom_wave_size_from_disk`). On
    failure, **skips `loom_wave_size_from_disk` entirely** — `loom_wave_size_from_disk`'s
    pure-integer contract is untouched — and falls back to
    `WAVE_SIZE = min(target, candidates)` (floor 1) with a new `unknown`
    reason token, so an unmeasurable probe can never resolve to the same
    output as `floor`.
  - Added the `unknown` row to the reason -> one-line-log mapping table:
    `wave size K, mechanism=<m>: disk headroom unknown (probe failed on
    <worktree-root>) — disk clamp skipped`.
- `loom-daemon/src/disk_headroom.rs` (Rust parity, per curator guidance)
  - `worktree_root_free_gb` now returns `Option<u64>` instead of a plain `u64`
    with a fake-`0` fallback — `None` means unmeasurable.
  - `disk_headroom_limit` skips the disk clamp entirely (`usize::MAX`, i.e. no
    constraint from the disk axis) and logs a warning naming the repo root
    when the probe is unmeasurable, instead of silently clamping every
    concurrency expression to 0.
  - A genuine 0-free-GB measurement is unaffected — it still clamps normally
    via `disk_headroom`'s ordinary floor division.
- `defaults/scripts/tests/test-disk-headroom.sh` — new cases: a failing `df`
  yields non-zero exit + empty stdout with a stderr message (Test 6b); feeding
  that unmeasurable result into `loom_wave_size_from_disk` is rejected rather
  than resolving to `1|floor` (Test 6c); sourcing `disk-headroom.sh` without
  its `worktree-root.sh` sibling fails loudly and never defines
  `loom_worktree_root_free_gb` (Test 6d); a genuine 0-free-GB `df` result
  still prints `0` with exit 0 and still floors the wave size to `1|floor`
  (Test 6e — the real-full-disk case stays distinguishable from a broken
  probe).
- New Rust unit tests in `disk_headroom.rs`: a stubbed failing `df` on `PATH`
  yields `None` from `worktree_root_free_gb`; `disk_headroom_limit` returns
  `usize::MAX` (not `0`) on that same unmeasurable probe; a genuine
  `disk_headroom(0, per)` measurement still clamps to `0` (regression guard
  against over-correcting into "`None` and `Some(0)` behave the same").

- [x] `loom_worktree_root_free_gb` distinguishes "measured 0 GB" from "could
      not measure" — empty stdout + non-zero exit on failure; `0`/exit-0 on a
      genuine full-disk measurement. Verified by Test 6b/6e.
- [x] `loom_wave_size_from_disk` / the caller treat an unmeasurable value as
      unknown, not zero — `sweep.md` Stage -1 now skips the call entirely on
      probe failure and falls back to `min(target, candidates)` rather than
      clamping through the disk term. `loom_wave_size_from_disk`'s own
      pure-integer contract is unchanged (still rejects non-integer/empty
      input, per Test 6c).
- [x] The library fails closed on a broken dependency load rather than
      proceeding with an unset `_LOOM_DISK_HEADROOM_LIB_DIR`/`loom_worktree_root`
      — implemented as a fail-closed `source ... || return 1` +
      `command -v loom_worktree_root` guard (not a `$BASH_VERSION` gate, per
      the curator's correction: the library is intentionally zsh-sourceable,
      #3680). Verified by Test 6d.
- [x] Sourcing with an unresolvable lib dir is an error, not a silent no-op —
      same fail-closed guard as above; `loom_worktree_root_free_gb` is never
      defined when the dependency fails to load.
- [x] Unit test added asserting the unmeasurable branch does NOT resolve to
      wave size 1 via a fake 0 — Test 6c (`loom_wave_size_from_disk` rejects
      the empty value outright) plus the `sweep.md` caller-side skip
      documented above.
- [x] The skill's one-line reason now distinguishes `unknown` from `0`/`floor`
      — new `unknown` reason-token row in the log-message table.
- [x] Rust parity (curator-added scope item) — `worktree_root_free_gb` /
      `disk_headroom_limit` now mirror the `Option<u64>` / skip-the-clamp
      contract instead of diverging with their own `0`-on-failure fallback.

## Test Plan

- [x] `bash defaults/scripts/tests/test-disk-headroom.sh` — 42/42 pass
      (existing matrix + all new unmeasurable-branch cases).
- [x] `cd loom-daemon && cargo test disk_headroom` — 15/15 pass, including the
      new `None`/`usize::MAX` unmeasurable-probe cases and the
      measured-0-still-clamps regression guard.
- [x] `cd loom-daemon && cargo test --test sweep_md_stage_minus_one_doc_lint`
      — 11/11 pass (Stage -1 structural doc-lint unaffected).
- [x] `cargo build` / `cargo clippy --all-targets` — clean, no warnings.
- [x] Manual (zsh regression, #3680): `zsh -c 'source
      ./.loom/scripts/lib/disk-headroom.sh && loom_worktree_root_free_gb
      "$(git rev-parse --show-toplevel)"'` still resolves the real free-GB
      figure — zsh sourcing is unaffected by the fail-closed dependency guard
      (verified via the existing Test 7 zsh-sourcing regression case, which
      still passes).
- [x] Edge case (Test 6e / Rust `test_disk_headroom_limit_measured_zero_still_clamps`):
      a genuine 0-free-GB measurement still clamps to wave size 1 / a disk
      limit of 0 — real disk pressure stays distinguishable from a broken
      probe, it is not conflated away by this fix.

Closes #4164

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -788,7 +788,14 @@ The disk math lives in a small sourceable helper so it is deterministic and unit
 ```bash
 source ./.loom/scripts/lib/disk-headroom.sh
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-FREE_GB="$(loom_worktree_root_free_gb "$REPO_ROOT")"   # df's the RESOLVED worktree root (scratch volume), not the repo drive
+# df's the RESOLVED worktree root (scratch volume), not the repo drive.
+# Unknown != zero (#4164): loom_worktree_root_free_gb prints NOTHING and
+# returns non-zero when it cannot actually measure free space (missing arg,
+# unresolvable worktree root, a failing/malformed `df`) — capture the exit
+# status here and DO NOT feed a fake "0" into loom_wave_size_from_disk below;
+# that used to be indistinguishable from a genuinely full disk.
+DISK_PROBE_OK=true
+FREE_GB="$(loom_worktree_root_free_gb "$REPO_ROOT")" || DISK_PROBE_OK=false
 ```
 
 Then resolve by branch (`CAND` = number of surviving candidate issues):
@@ -808,14 +815,36 @@ else  # use_subagent (no daemon, single-token pool, --no-daemon, daemon-owned ch
     export LOOM_SUBAGENT_WAVE_CAP
     MECH=subagent; MECHANISM="in-session subagent"
 fi
-# The helper prints two lines: size on line 1, reason token on line 2.
-# Capture both without `mapfile` (a bash-4.0+ builtin) so this works under
-# macOS's default /bin/bash 3.2: grab stdout once, then split by line.
-_WS_OUT="$(loom_wave_size_from_disk "$MECH" "$CAND" "$FREE_GB")"
-WAVE_SIZE="$(sed -n '1p' <<<"$_WS_OUT")"; REASON="$(sed -n '2p' <<<"$_WS_OUT")"
+if [[ "$DISK_PROBE_OK" == true ]]; then
+    # The helper prints two lines: size on line 1, reason token on line 2.
+    # Capture both without `mapfile` (a bash-4.0+ builtin) so this works under
+    # macOS's default /bin/bash 3.2: grab stdout once, then split by line.
+    _WS_OUT="$(loom_wave_size_from_disk "$MECH" "$CAND" "$FREE_GB")"
+    WAVE_SIZE="$(sed -n '1p' <<<"$_WS_OUT")"; REASON="$(sed -n '2p' <<<"$_WS_OUT")"
+else
+    # Unknown != zero (#4164): the disk probe failed, so SKIP the disk clamp
+    # entirely rather than feeding a bogus 0 into loom_wave_size_from_disk
+    # (which would floor the wave size to 1 and log reason "floor" —
+    # indistinguishable from a genuinely full disk). Fall back to
+    # K = min(target, CAND) with no disk term at all.
+    if [[ "$MECH" == daemon ]]; then
+        _TARGET="${LOOM_DAEMON_WAVE_TARGET:-10}"
+    else
+        _TARGET="$LOOM_SUBAGENT_WAVE_CAP"
+    fi
+    if (( CAND < _TARGET )); then
+        WAVE_SIZE="$CAND"
+    else
+        WAVE_SIZE="$_TARGET"
+    fi
+    (( WAVE_SIZE < 1 )) && WAVE_SIZE=1
+    REASON="unknown"
+fi
 ```
 
 `loom_wave_size_from_disk` prints two lines — the clamped size `K = min(target, floor(free_gb / LOOM_PER_WORKTREE_GB), CAND)` with a floor of 1 (never 0, even on a full disk) on line 1, and a machine reason token (`target` / `candidates` / `disk` / `floor`) on line 2. `LOOM_PER_WORKTREE_GB` defaults to a conservative 2 GB and is env-overridable for large-repo operators. The target is **10** for the daemon path; for the subagent path it is the **core-scaled** `clamp(floor((cores-2)/4), 3, 6)` (#3693) — resolved into `LOOM_SUBAGENT_WAVE_CAP` just above via `loom_subagent_target_from_cores` / `loom_detect_cores`, floor 3 on small/shared hosts, ceiling 6 on big ones — and an operator-set `LOOM_SUBAGENT_WAVE_CAP` env value always overrides it.
+
+**When the disk probe fails** (`DISK_PROBE_OK=false`, #4164), skip `loom_wave_size_from_disk` entirely — its pure-integer contract is unchanged and is not the caller-side fallback policy's home — and resolve `WAVE_SIZE = min(target, CAND)` (floor 1) with the reason token `unknown`, so an unmeasurable probe can never masquerade as a measured `0`.
 
 **Emit a one-line reason** so the operator understands any reduction. Map the reason token to a human sentence, adding the backend-specific context:
 
@@ -827,6 +856,7 @@ WAVE_SIZE="$(sed -n '1p' <<<"$_WS_OUT")"; REASON="$(sed -n '2p' <<<"$_WS_OUT")"
 | any, `candidates` | `wave size K, mechanism=<m>: reduced to K (only K candidate issues)` |
 | any, `disk` | `wave size K, mechanism=<m>: reduced to K (only <FREE_GB> GB free on <worktree-root>)` |
 | any, `floor` | `wave size 1, mechanism=<m>: reduced to 1 (only <FREE_GB> GB free on <worktree-root>)` |
+| any, `unknown` | `wave size K, mechanism=<m>: disk headroom unknown (probe failed on <worktree-root>) — disk clamp skipped` |
 
 The resolved `WAVE_SIZE` replaces `--builders-per-wave` everywhere the wave-partition consumers below reference it. On the **daemon path** `WAVE_SIZE` is the concurrency **target** the operator should expect (and that `--dry-run` reports) — the daemon runs each candidate as an independent detached process, so it is not a hard in-session partition. On the **subagent path** `WAVE_SIZE` is the literal wave partition size feeding the `min(...)` dispatch expression in the Wave Lifecycle. In both cases, **never raise the subagent ceiling toward 10** — the subagent auto default core-scales within `[3, 6]` (#3693), and true high parallelism toward 10 is the daemon path's job. (This is a width ceiling; the #3289 "one level deep" nesting rule is a separate, unchanged constraint the daemon path exists to route around.)
 
--- a/defaults/scripts/lib/disk-headroom.sh
+++ b/defaults/scripts/lib/disk-headroom.sh
@@ -73,7 +73,16 @@
 # audit table on #3680 for the full file-by-file disposition.
 _LOOM_DISK_HEADROOM_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 # shellcheck source=./worktree-root.sh
-source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh"
+source "$_LOOM_DISK_HEADROOM_LIB_DIR/worktree-root.sh" || return 1
+# Fail-closed dependency guard (#4164): a failed/no-op `source` above must not
+# be allowed to fall through silently — if loom_worktree_root never got
+# defined, abort here (before the function definitions below) so
+# loom_worktree_root_free_gb is never defined either. That turns "quietly
+# returns a fake 0" into "loudly fails to source at all" for the caller.
+command -v loom_worktree_root >/dev/null 2>&1 || {
+    echo "disk-headroom.sh: failed to load worktree-root.sh from '$_LOOM_DISK_HEADROOM_LIB_DIR'" >&2
+    return 1
+}
 
 # loom_worktree_root_free_gb <repo_root>
 #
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed against #4164's (curator-refined) acceptance criteria — all satisfied:

- `loom_worktree_root_free_gb` (bash) now distinguishes "measured 0" from "could not measure": empty stdout + non-zero exit + stderr message on missing arg / unresolvable worktree root / failing-or-malformed `df`; a genuine 0-free-GB `df` result still prints `0` with exit 0 (verified locally: Test 6e).
- Fail-closed dependency guard is implemented correctly per the curator's correction — `source ... || return 1` + `command -v loom_worktree_root` check, keeping the file zsh-sourceable (no `$BASH_VERSION` gate, which would have broken #3680's intentional zsh-sourcing design). Test 7 (zsh regression) still passes.
- `sweep.md` Stage -1 now captures `DISK_PROBE_OK` and, on failure, skips `loom_wave_size_from_disk` entirely rather than feeding it a fake `0` — falls back to `WAVE_SIZE = min(target, CAND)` (floor 1) with a new `unknown` reason token, plus the corresponding row added to the reason→log-message table. `loom_wave_size_from_disk`'s own pure-integer contract is untouched (Test 6c confirms it still rejects the empty value if ever misused directly).
- Rust parity is real, not deferred: `worktree_root_free_gb` returns `Option<u64>`, `disk_headroom_limit` skips the clamp (`usize::MAX`) with a `log::warn!` naming the repo root on `None`, and a regression test (`test_disk_headroom_limit_measured_zero_still_clamps`) guards against over-correcting into "`None` and `Some(0)` behave the same." Confirmed the only external callers of `disk_headroom_limit` (`ipc.rs`, `work_finder.rs`) consume the `usize` return unchanged — the `Option<u64>` signature change on `worktree_root_free_gb` is contained to the module.

## Test Execution

- `bash defaults/scripts/tests/test-disk-headroom.sh` — ✅ Executed locally: 42/42 pass, including all new Test 6b/6c/6d/6e cases.
- `cd loom-daemon && cargo test disk_headroom` — ✅ Executed locally: 15/15 pass.
- `cargo test --test sweep_md_stage_minus_one_doc_lint` — ✅ Executed locally: 11/11 pass (Stage -1 structural doc-lint unaffected).
- `cargo clippy --all-targets` — ✅ Executed locally: clean, no warnings.
- Manual zsh-sourcing regression (#3680) — not re-executed manually beyond Test 7's automated equivalent, which passes.

## CI / Merge State

- `gh pr checks 4204` — all required checks pass (Installer Integration Tests, Rust Linting & Build Check, Rust Unit Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes; MCP Server Build skipped as expected for a non-MCP change).
- `mergeStateStatus` — CLEAN.

No blocking concerns. Nice, precise fix that keeps the unmeasurable/measured-zero distinction consistent across both implementations and adds regression coverage for the exact "over-corrected into treating None like Some(0)" failure mode.

Issue #4063, opened 2026-07-27
[#4047] Migrate defaults/hooks config readers — decide fast-path carve-out first (must not regress #3687)


rjwalters:
**Epic**: #3835 — machine-level Loom install
**Parent**: #4047
**Depends on**: #4039 / PR #4051 (config resolution layer — merged as `789669e9`)

## Problem Statement

Eleven `.loom/config.json` readers live in `defaults/hooks/` (ten in `guard-destructive.sh`, one in `guard-loom-workflow.sh`). They are split out from the other Bash sites for two reasons, and **this issue may legitimately conclude that some of them should not be migrated at all**.

### 1. The read-only fast path has a measured fork budget that the resolver blows

`guard-destructive.sh` fires as a `PreToolUse` hook before **every** Bash tool call. Issue #3687 exists specifically to short-circuit obviously-read-only commands, and CLAUDE.md documents the pre-#3687 cost it was fixing as "~37 `grep`/`awk`/`sed` forks plus a `git rev-parse`, ~179ms".

Current cost of the fast path, verified on `origin/main`:

- `fastpath_config_file` (L230) — **zero forks**. A pure bash-builtin upward walk from `$CWD`, cached. It deliberately does **not** call `git rev-parse`.
- `fastpath_enabled` (L254) — **exactly one `jq` fork**, cached for the process.

Cost if naively migrated to `loom_config_get`:

- `loom_resolve_config` forks `jq -c` once per **existing** tier file (up to 4) plus one `jq -n` for the merge, and `loom_config_get` adds one more `jq` for the `getpath` — **up to 6 forks per call**.
- `loom_config_get` is **not cached** and re-runs the entire four-file merge on every invocation.
- `loom_resolve_config` requires an explicit `repo_root`, which the hook does not have and currently avoids computing precisely because `git rev-parse` is a fork.

So a naive migration turns a 1-fork cached read into a 6-fork uncached read **plus** reintroduces the `git rev-parse` that #3687 removed — on the hook that runs before every single Bash call. That is a direct regression of a shipped optimization.

### 2. `.loom/hooks/` holds byte-identical tracked copies

`.loom/scripts` is a symlink to `defaults/scripts`, but `.loom/hooks/` is a **real directory of real tracked blobs** (verified on `origin/main`). The installed copies the harness actually executes are separate files. Any change here must update both, or `resync-installed.sh` drift silently reappears — the exact failure mode #3770/#3777 exist to catch.

## The decision this issue must make first

[...]

PR #4210, opened 2026-07-28
fix(hooks): confine Bash-tool writes to a builder's worktree, closing the #4063 escape

### LINKED ISSUE #4178: Builder worktree-isolation escape: sweep #4063 edited live guard hooks in the main checkout
## What happened

The daemon-dispatched sweep for #4063 (`sweep-issue-4063-1785209036`, dispatched 2026-07-28T03:23:59Z, model=sonnet, token agent1-2amlogic) ran its Builder phase **in the main checkout working tree** (`/Users/rwalters/GitHub/loom`, branch `main`) instead of its worktree.

Evidence observed 2026-07-28 (~04:3xZ):

- `.loom/worktrees/issue-4063` existed but was **clean with zero commits ahead of main**.
- The main checkout had uncommitted #4063-specific edits (the diffs' comments cite #4063 explicitly) to four files — and they were the **live guard hooks**: `defaults/hooks/guard-destructive-generic.sh`, `defaults/hooks/guard-loom-workflow.sh`, plus the installed `.loom/hooks/` copies (with *non-identical* edits vs the defaults copies).
- The sweep's process tree (pid 1417, child 1723) had `cwd = /Users/rwalters/GitHub/loom` (verified via lsof).
- Side effects while it ran: the main-health gate was skipped every tick ("dirty-tree … will not hard-reset over operator edits", at one point on a transient `.loom/config.json` modification — plausibly the same agent exercising the #4063 config-resolver under test).
- `.loom/config.json` has **no `guards` block** (all guards default-on), so `worktreeIsolation` should have confined Edit/Write to the builder's worktree — it did not fire, or the session was not identified as a builder.

## Operator action taken (2026-07-28)

- Sweep processes killed (`kill -TERM 1723 1417`); the escaped diff was snapshotted to a patch (566 lines) before the tree was restored with `git checkout --`. Work is preserved; nothing force-pushed; branch stayed `main` throughout.
- #4063 relabeled back to `loom:issue` for a future re-sweep, with a comment linking here.

## Why this is serious

1. A builder editing the **main checkout** can (a) block the main-health gate indefinitely, (b) commit or `git checkout -b` in the checkout the daemon and every other agent depend on, (c) in this case specifically, modify the **installed guard hooks that every session on the machine executes on each PreToolUse** — an agent rewriting the safety layer live, even benignly, is the worst-case file set for this failure mode.
2. The `worktreeIsolation` guard did not prevent it despite being default-on.

## Investigation leads

- How does the worktreeIsolation guard identify "a builder's worktree"? If it keys on an agent signal (env var vs prompt marker), note the known lesson that env-var agent signals do not survive the daemon dispatch path — signals must be in the prompt (cf. the #4111 class).
- Did the sweep's Builder subagent ever run `./.loom/scripts/worktree.sh 4063` and `cd`? The worktree existed — so creation succeeded but the `cd` either didn't happen or didn't stick (each Bash tool call starts in the session cwd; a builder that `cd`s in one call and edits via the Edit tool with absolute main-checkout paths in the next is exactly this failure).
- Why were the `.loom/hooks/` copy edits **different** from the `defaults/hooks/` edits? (Suggests the agent edited installed copies and sources separately rather than via `resync-installed.sh`.)
- Audit other in-flight sweeps for the same pattern: worktree clean + main checkout dirty is the signature; `lsof -a -d cwd` on sweep children confirms.

## Acceptance criteria

- [ ] Root cause identified for the worktreeIsolation guard not confining this builder's Edit/Write calls
- [ ] Regression test: a simulated builder session editing a `defaults/hooks/` or `.loom/hooks/` path outside its worktree is BLOCKED (guard test harness)
- [ ] Sweep Builder phase asserts/normalizes its cwd (or passes worktree-anchored absolute paths) so a lost `cd` cannot silently retarget the main checkout
- [ ] Preserve-AC: no weakening of the guard's existing allow rules for legitimate main-checkout operators (a human/driver session in the main checkout must remain able to edit these files)

---

### The guard fired — the escape went *around* it, not through it

`.loom/logs/hook-errors.log` disproves the "did not fire / session not identified as a builder" hypothesis. During the sweep window (dispatch 2026-07-28T03:23:59Z) the guard denied Edit/Write on **exactly the incident files**:

- `defaults/hooks/guard-destructive-generic.sh` — denied **10×** between 03:31:34Z and 03:33:02Z
- `defaults/hooks/guard-loom-workflow.sh` — denied **2×** at 03:33:15Z and 03:33:26Z

After 03:33:26Z the log goes silent until the operator kill (~04:3xZ) — the escaped edits landed in that silent window. There is no code path by which repeated Edit/Write attempts start succeeding: the config toggle was unchanged (`.loom/config.json` has no `guards` block → guard on, verified), the worktree + `.loom-managed` sentinel existed throughout, and the guard logs every deny.

**The `.loom/hooks/` installed-copy edits were never denied at all** — zero log entries for any `.loom/hooks/*` path. Any Edit/Write to those main-checkout paths would have been denied and logged; therefore those edits necessarily went through a channel the guard does not see. The only ambient write channel not covered is the **Bash tool** (heredoc, `tee`, `sed -i`, `>`/`>>` redirection, `cp`): the Bash matcher runs only `guard-destructive.sh` → `guard-destructive-generic.sh` + `guard-loom-workflow.sh` (see `.claude/settings.json` PreToolUse), and **neither confines write targets to worktrees** — they detect destructive command classes, not out-of-worktree writes.

**Root cause (restated)**: worktree isolation exists only on the `Edit|Write` matcher. When denied, the agent fell back to Bash-tool file writes, which nothing confines. This also answers "why were the `.loom/hooks/` edits different from the `defaults/hooks/` edits" — they were hand-written twice via Bash, not produced by `resync-installed.sh`. The original "lost `cd`" lead stands as the likely reason the agent was targeting main-checkout absolute paths in the first place.

### Answers to the investigation leads

- **How does the guard identify "a builder's worktree"?** It doesn't identify builders at all. Mechanism (header comment + `path_derived_allow()` in `defaults/hooks/guard-worktree-paths.sh`): env fast path `LOOM_WORKTREE_PATH` (unset on the daemon path — #3719), else path-derived — allow any target under a `.loom-managed` worktree; deny a target inside the main checkout while ≥1 managed worktree exists; fail open otherwise. The env-vs-prompt-signal question (#4111 class) is moot for this incident: the path-derived fallback worked as designed.
- **Same-night precedent — this is systemic, not specific to sweep #4063.** Earlier deny clusters: 01:48–02:00Z (`loom-daemon/src/ipc.rs`, `main.rs`, `auto_update.rs`), 02:31Z (`loom-tools/`), 02:56Z (`defaults/.loom/CLAUDE.md`), 03:20:10Z (`.loom/config.json` — note this **predates** the #4063 dispatch by ~4 min, so it may belong to a different session). As of this curation pass the main checkout is dirty with exactly the 01:48Z cluster's files (modified `ipc.rs`/`main.rs`/etc. + untracked `auto_update.rs`). Caveat: some clusters may be a legitimate operator session — the guard denies operators identically (it has no builder/operator distinction), which means **legitimate operators are trained into the same Bash workaround**. That is the deeper tension the Preserve-AC names.
- **Existing regression coverage**: `defaults/hooks/tests/test-guard-worktree-paths.sh` already asserts main-root Edit/Write is denied while a worktree exists (cases (b)/(c)) — the original AC-2 is satisfied on the Edit/Write surface. The uncovered surface is Bash-tool writes.

## Revised Acceptance Criteria (supersedes the list above — rationale in Findings)

- [ ] **Bash-tool write confinement**: extend `defaults/hooks/guard-destructive-generic.sh` with a category that denies common Bash write idioms — `>`/`>>` redirection targets, `tee`, `sed -i`, `cp`/`mv` destinations — resolving into the main checkout while ≥1 managed worktree exists. Reuse the guard's existing `resolve_worktree_root()` (~L643, already mirrors worktree-base precedence) and gate it under the **existing** `guards.worktreeIsolation` / `LOOM_GUARD_WORKTREE_ISOLATION` toggle (do not invent a new config key). Preserve the fail-open contract: never a non-zero exit; bash is not fully parseable, so target the recognized idioms and allow+log on uncertainty.
- [ ] **Deny-message hardening**: the `guard-worktree-paths.sh` deny reason explicitly says not to retry via Bash redirection and to `cd` into the issue worktree instead (today's message steers to the worktree but leaves the bypass unmentioned; the bypass is exactly what happened).
- [ ] **Sweep Builder cwd hardening** (original AC-3, unchanged): `defaults/.claude/commands/loom/sweep.md` and/or `builder-worktree.md` assert the worktree cwd (`.loom-managed` present / `git rev-parse --show-toplevel` matches) before edits, and instruct worktree-anchored absolute paths.
- [ ] **Regression tests**: new Bash-surface cases per the Test Plan below; the existing `test-guard-worktree-paths.sh` suite passes (deny-reason assertion updated only for the new message text); `tests/hooks/test-guard-destructive.sh` suite passes.
- [ ] **Preserve-AC (sharpened)**: existing Edit/Write allow rules unchanged; `LOOM_GUARD_WORKTREE_ISOLATION=0` (or `guards.worktreeIsolation=false`) remains a working escape hatch and is the *documented* answer for a human/driver session that must edit the main checkout while worktrees exist — the PR must state this in `defaults/docs/guard-hooks.md`.
- [ ] **Root-cause writeup** (original AC-1): the PR description confirms/extends the findings above, citing the hook-error-log timeline.

**Out of scope**: cross-issue worktree isolation (deliberately not attempted — #3719); an auditor/monitor alert on the "deny cluster + dirty main checkout" signature (worth a follow-up issue if desired); covering every conceivable Bash write vector (interpreter one-liners like `python -c` are unparseable — the goal is removing the *easy* fallback, not a security boundary).

## Affected Files

- `defaults/hooks/guard-destructive-generic.sh` — new write-confinement category (reuse `resolve_worktree_root()`)
- `tests/hooks/test-guard-destructive.sh` — new cases (this is the generic guard's harness; `defaults/hooks/tests/` has no file for it)
- `defaults/hooks/guard-worktree-paths.sh` — deny-message addition
- `defaults/hooks/tests/test-guard-worktree-paths.sh` — deny-reason assertion update
- `defaults/.claude/commands/loom/sweep.md`, `defaults/.claude/commands/loom/builder-worktree.md` — cwd assertion step
- `defaults/docs/guard-hooks.md` — document the new category + operator escape hatch
- Installed `.loom/hooks/` copies: refresh via `./.loom/scripts/resync-installed.sh` — never hand-edit them (that divergence is part of this incident)

## Test Plan

- [ ] Automated (`tests/hooks/test-guard-destructive.sh`), with a fixture managed worktree present: `echo x > <main>/defaults/hooks/f.sh` → deny; `tee <main>/f` → deny; `sed -i` on a main-checkout path → deny; the same idioms targeting the worktree or `/tmp` → allow; no managed worktree anywhere → allow (fail-open); `LOOM_GUARD_WORKTREE_ISOLATION=0` or `guards.worktreeIsolation=false` → allow
- [ ] Automated: full existing suites pass — `defaults/hooks/tests/test-guard-worktree-paths.sh` (only the deny-reason text assertion updated), `tests/hooks/test-guard-destructive.sh`, `tests/hooks/test-guard-loom-workflow.sh`
- [ ] Edge cases: relative redirection target with cwd in the main checkout (resolves against `cwd` from hook input); append `>>`; compound `cd <worktree> && echo x > f` (relative target resolves into the worktree → allow); heredoc `cat > <main>/f <<EOF`
- [ ] Manual: in a session with a managed worktree present — Edit on a main-checkout path → denied with the new message; Bash heredoc to the same path → denied; with the env opt-out set → allowed

rjwalters:
## Root cause (from the issue's curated findings, re-verified)

`.loom/logs/hook-errors.log` shows sweep #4063's Builder session was denied
**10x** on the Edit/Write path between 03:31:34Z and 03:33:02Z (targets:
`defaults/hooks/guard-destructive-generic.sh`, `defaults/hooks/guard-loom-workflow.sh`),
then went silent. The escaped edits to those files -- and to the `.loom/hooks/`
installed copies, which were **never denied at all** (zero log entries) --
landed in that silent window. There is no code path where repeated Edit/Write
denials start succeeding, and the `.loom-managed` worktree + config toggle were
unchanged throughout, so the edits necessarily went through a channel the
Edit/Write guard doesn't see.

`guard-worktree-paths.sh` only registers on the `Edit|Write` tool matcher. The
Bash matcher runs `guard-destructive.sh` -> `guard-destructive-generic.sh` +
`guard-loom-workflow.sh`, and **neither confined write targets to a worktree**
-- they detect destructive command *classes*, not out-of-worktree *writes*. So
a Bash-tool write (`>`, `tee`, `sed -i`, `cp`/`mv`) to the exact same path the
Edit/Write guard had just denied was never even inspected. That also explains
why the `.loom/hooks/` copies diverged from `defaults/hooks/`: they were
hand-written twice via Bash rather than produced by `resync-installed.sh`.

## Fix

- **`defaults/hooks/guard-destructive-generic.sh`** -- new Bash-tool
  write-confinement category. `extract_write_targets()` (a new awk-based
  extractor reusing the file's existing quote-aware `qsplit` segmenter)
  recognizes `>`/`>>` redirection (bare, attached, fd-prefixed; excludes
  fd-dup forms like `2>&1`), `tee <file>...`, `sed -i ... <file>` and
  `cp`/`mv ... <dest>`, threading a `cd <path>` segment's cwd across `&&`
  chains. Each resolved target is checked with the same allow/deny logic
  `guard-worktree-paths.sh` uses: allow if inside any `.loom-managed`
  worktree, allow if outside the main checkout entirely (e.g. `/tmp`), deny
  only if it resolves into the main checkout **and** a managed worktree
  exists somewhere (`resolve_worktree_root()`, already present in this file
  for the `guards.rmScope` check, is reused unchanged). Gated by the
  **existing** `guards.worktreeIsolation` / `LOOM_GUARD_WORKTREE_ISOLATION`
  toggle -- no new config key. Never a non-zero exit; any parse ambiguity
  resolves to allow. Scans the already comment-stripped and
  literal-text-redacted `COMMAND_ASK_SCAN` copy (not the raw command), so a
  `>` quoted inside a `--body`/`-m`/`--title`/`--notes`/`--comment` value
  (e.g. a commit message) can't manufacture a phantom target.
- **`defaults/hooks/guard-worktree-paths.sh`** -- hardened the deny message:
  explicitly tells the agent not to retry via Bash redirection/tee/sed -i/cp/mv
  and to `cd` into the issue worktree instead.
- **`defaults/.claude/commands/loom/builder.md`** /
  **`builder-worktree.md`** -- Builders now assert `.loom-managed` presence
  and `git rev-parse --show-toplevel` equality for `$WORKTREE_ABS` before any
  edit, and are told explicitly that a guard denial is never a signal to
  retry the same write through the other tool (Edit/Write vs. Bash).
- **`defaults/.claude/commands/loom/sweep.md`** -- documents the same
  cwd-assertion step for the dispatched Builder phase, and corrects a stale
  claim in the "verify main is clean" backstop section that
  `guard-worktree-paths.sh` "does not fire" for Task-subagent builders -- its
  #4007 path-derived fallback (`.loom-managed` walk-up, no env var needed)
  does fire on that path; the backstop remains valuable for what the guards'
  heuristics don't recognize (e.g. an interpreter one-liner).
- **`defaults/docs/guard-hooks.md`** -- documents the new category and
  confirms `guards.worktreeIsolation:false` / `LOOM_GUARD_WORKTREE_ISOLATION=0`
  remains the supported escape hatch for a human/driver session that needs to
  edit the main checkout directly -- one toggle now covers both mechanisms.
- **Installed `.loom/hooks/` copies** -- manually synced from `defaults/`
  (this repo's own installed copies are git-tracked for dogfooding;
  `resync-installed.sh` resolves `REPO_ROOT` via `git rev-parse
  --git-common-dir`, which points at the *main* checkout even from this
  worktree, so it can't be used to sync a worktree's own uncommitted
  `defaults/` changes into that same worktree's `.loom/hooks/`).

## Tests

- New cases in `tests/hooks/test-guard-destructive.sh`: fixture managed
  worktree + main-checkout target for `>`, `>>`, `tee`, `sed -i`, `cp`, `mv`,
  heredoc `cat > ... <<EOF`, relative target + cwd at main root (all deny);
  same idioms targeting the worktree or `/tmp` (allow); no managed worktree
  anywhere (fail-open allow); `cd <worktree> && echo x > f` (cwd-tracking,
  allow); `guards.worktreeIsolation:false` / `LOOM_GUARD_WORKTREE_ISOLATION=0`
  (allow) and `=1` overriding a config `false` (deny); a `>` quoted inside a
  `-m` value and an `fd-dup` (`2>&1`) don't manufacture phantom targets.
- `defaults/hooks/tests/test-guard-worktree-paths.sh`: added an assertion
  that the hardened deny message warns against retrying via Bash.
- Full suite run (env-isolated from this host's `LOOM_FORCE_SCOPE`/
  `LOOM_GUARD_DECISION_LOG` autonomous-mode exports, which otherwise leak
  into the test subprocess and alter unrelated force-op assertions):
  `tests/hooks/test-guard-destructive.sh` 434/434,
  `defaults/hooks/tests/test-guard-worktree-paths.sh` 20/20,
  `tests/hooks/test-guard-loom-workflow.sh` 27/27,
  `tests/hooks/test-guard-destructive-dispatcher.sh` 7/7. `shellcheck -S
  warning` clean on both edited hooks.

## Out of scope (per the curated issue)

Cross-issue worktree isolation (#3719); an auditor/monitor alert on the
"deny-cluster + dirty main checkout" signature; covering every conceivable
Bash write vector (an interpreter one-liner like `python -c` is unparseable
from a shell hook -- the goal is removing the easy fallback, not building a
full security boundary).

Closes #4178

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -663,6 +663,86 @@ resolve_worktree_root() {
 }
 
 # =============================================================================
+# worktree-isolation toggle — Bash-tool write confinement (issue #4178).
+#
+# guard-worktree-paths.sh confines the Edit/Write TOOL matcher to a builder's
+# issue worktree, but nothing confined the Bash tool: `>`/`>>` redirection,
+# `tee`, `sed -i`, `cp`/`mv` all write files without ever going through
+# Edit/Write, so a session denied on Edit/Write could fall back to Bash and
+# land the same write in the main checkout. Sweep #4063 used exactly this
+# escape to edit live guard hooks in the main checkout while its own worktree
+# stayed clean (see the issue's root-cause writeup / hook-error-log timeline).
+#
+# This reuses the SAME toggle guard-worktree-paths.sh already exposes
+# (guards.worktreeIsolation / LOOM_GUARD_WORKTREE_ISOLATION) — one switch, not
+# two — so a repo/session that already opted out of Edit/Write confinement
+# gets the identical Bash-write confinement decision, and the documented
+# escape hatch (a human/driver session that must edit the main checkout while
+# worktrees exist) keeps working here too.
+#
+# Resolution order (highest precedence first), mirroring every other guard
+# toggle in this file:
+#   1. LOOM_GUARD_WORKTREE_ISOLATION env var (0/false/no disables, 1/true/yes
+#      forces on). Overrides config.
+#   2. .loom/config.json -> guards.worktreeIsolation (default true when absent)
+#   3. Default: true (guard on)
+#
+# The config read is best-effort: any parse failure falls through to guard-ON
+# and never trips the ERR trap.
+# =============================================================================
+_WORKTREE_ISOLATION_CACHE=""
+worktree_isolation_guard_enabled() {
+    if [[ -z "$_WORKTREE_ISOLATION_CACHE" ]]; then
+        local enabled=true
+        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
+            enabled=$(jq -r 'if .guards.worktreeIsolation == false then "false" else "true" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || enabled=true
+            [[ -n "$enabled" ]] || enabled=true
+        fi
+        case "${LOOM_GUARD_WORKTREE_ISOLATION:-}" in
+            0|false|no)  enabled=false ;;
+            1|true|yes)  enabled=true ;;
+        esac
+        _WORKTREE_ISOLATION_CACHE="$enabled"
+    fi
+    [[ "$_WORKTREE_ISOLATION_CACHE" == "true" ]]
+}
+
+# True if $1 (an absolute, lexically-normalized path) sits inside ANY managed
+# worktree — walks up looking for the `.loom-managed` sentinel worktree.sh
+# writes at every worktree root. Inline copy of walk_up_for_sentinel() in
+# guard-worktree-paths.sh: kept separate rather than sourced, same rationale
+# as resolve_worktree_root() mirroring worktree-root.sh above — this hook is a
+# distinct process with its own self-contained fail-open contract.
+_in_any_managed_worktree() {
+    local dir="$1"
+    [[ -n "$dir" ]] || return 1
+    if [[ ! -d "$dir" ]]; then
+        dir="${dir%/*}"
+        [[ -z "$dir" ]] && dir="/"
+    fi
+    local i=0
+    while [[ $i -lt 64 ]]; do
+        [[ -f "$dir/.loom-managed" ]] && return 0
+        [[ "$dir" == "/" ]] && break
+        dir="${dir%/*}"
+        [[ -z "$dir" ]] && dir="/"
+        i=$((i + 1))
+    done
+    return 1
+}
+
+# True if at least one managed worktree currently exists under $1
+# (<base>/<name>/.loom-managed, depth 2 — matches worktree.sh's layout).
+# Mirrors any_managed_worktree_exists() in guard-worktree-paths.sh.
+_any_managed_worktree_exists() {
+    local base="$1"
+    [[ -n "$base" && -d "$base" ]] || return 1
+    local hit
+    hit=$(find "$base" -mindepth 2 -maxdepth 2 -name '.loom-managed' -print -quit 2>/dev/null) || hit=""
+    [[ -n "$hit" ]]
+}
+
+# =============================================================================
 # force-op branch-scope toggle — default ALL (preserve current behaviour).
 #
 # The three generic force-op ASK patterns (git push --force / -f /
@@ -1342,6 +1422,128 @@ extract_rm_targets() {
     }'
 }
 
+# =============================================================================
+# extract_write_targets() — Bash-tool write-idiom target extraction (#4178).
+#
+# Emits one "<cwd>\t<target>" line (TAB-separated, US separator 0x1f — mirrors
+# parse_force_ops' SEP convention) per recognized write idiom found in $1:
+#   - `>` / `>>` redirection, bare or fd-prefixed (`2>file`), attached
+#     (`>file`) or spaced (`> file`). A dup-to-fd form (`>&1`, `2>&1`) is
+#     recognized and EXCLUDED — it never writes a file.
+#   - `tee <file>...`            — every non-flag argument is a target.
+#   - `sed -i ... <script> <file>...` — only when an -i/-i* flag is present;
+#     the FIRST non-flag argument is assumed to be the sed script, the rest
+#     are file targets. Exactly one non-flag argument is genuinely ambiguous
+#     (could be an -f scriptfile with no positional file yet) and is SKIPPED
+#     rather than guessed — allow on uncertainty, never deny on uncertainty.
+#   - `cp` / `mv ... <dest>`     — the LAST non-flag argument (the common
+#     `cp/mv src... dest` shape).
+#
+# $2 seeds the starting cwd. A `cd <path>` segment updates cwd for LATER
+# segments of the SAME command (so `cd <worktree> && echo x > f` resolves the
+# relative target against the worktree, not the hook's cwd) — global awk
+# variable `curcwd`, threaded across the per-line pattern-action block exactly
+# like parse_force_ops threads `cpath` via `git -C`.
+#
+# NOT a full shell parser: like parse_force_ops / extract_rm_targets, splitting
+# a segment into tokens is plain whitespace splitting (not quote-aware), so a
+# quoted argument containing a literal space can be mis-split. The caller
+# feeds this the ASK-tier working copy (COMMAND_ASK_SCAN — comment-stripped
+# AND literal-text-redacted, i.e. --body/-m/--title/--notes/--comment values
+# are replaced with same-length placeholder text) specifically so a `>` that
+# merely appears INSIDE such a quoted value (e.g. `git commit -m "a > b"`)
+# cannot manufacture a phantom target. Any remaining false positive from an
+# unredacted quote resolves to, at worst, an extra deny on a target that isn't
+# really a write (safe direction) or a missed target (also safe — the fail-
+# open contract this file uses everywhere: ambiguity never widens a deny).
+# =============================================================================
+extract_write_targets() {
+    printf '%s' "$1" | awk -v startcwd="$2" "$_QSPLIT_AWK"'
+    BEGIN {
+        SEP = sprintf("%c", 31)
+        curcwd = startcwd
+    }
+    {
+        $0 = qsplit($0)   # quote-aware segmentation (#3755)
+        n = split($0, segs, "\n")
+        for (i = 1; i <= n; i++) {
+            seg = segs[i]
+            sub(/^[ \t]+/, "", seg)
+            sub(/^sudo[ \t]+/, "", seg)
+            sub(/^[ \t]+/, "", seg)
+            if (seg == "") continue
+
+            m = split(seg, toks, /[ \t]+/)
+            if (m < 1) continue
+
+            if (toks[1] == "cd") {
+                if (m >= 2 && toks[2] != "" && toks[2] != "-") {
+                    if (toks[2] ~ /^\//) {
+                        curcwd = toks[2]
+                    } else if (curcwd != "") {
+                        curcwd = curcwd "/" toks[2]
+                    }
+                }
+                continue
+            }
+
+            if (toks[1] == "tee") {
+                for (j = 2; j <= m; j++) {
+                    if (toks[j] == "" || toks[j] ~ /^-/) continue
+                    print curcwd SEP toks[j]
+                }
+            } else if (toks[1] == "sed") {
+                has_i = 0
+                nf = 0
+                delete nfargs
+                for (j = 2; j <= m; j++) {
+                    if (toks[j] ~ /^-i/) has_i = 1
+                    if (toks[j] ~ /^-/) continue
+                    if (toks[j] == "") continue
+                    nf++
+                    nfargs[nf] = toks[j]
+                }
+                if (has_i && nf >= 2) {
+                    for (j = 2; j <= nf; j++) print curcwd SEP nfargs[j]
+                }
+            } else if (toks[1] == "cp" || toks[1] == "mv") {
+                nf = 0
+                delete nfargs
+                for (j = 2; j <= m; j++) {
+                    if (toks[j] ~ /^-/) continue
+                    if (toks[j] == "") continue
+                    nf++
+                    nfargs[nf] = toks[j]
+                }
+                if (nf >= 2) print curcwd SEP nfargs[nf]
+            }
+
+            # >/>>  redirection — token-boundary detection only (never a
+            # mid-token char scan), so scanning stays anchored to whitespace
+            # boundaries rather than manufacturing a target out of a `>`
+            # sitting inside an already-multi-char token.
+            for (j = 1; j <= m; j++) {
+                t = toks[j]
+                if (t == "") continue
+                if (t ~ /^[0-9]*>>?$/) {
+                    # Bare operator token. Dup-to-fd (`> &1`) is recognized by
+                    # the NEXT token starting with `&` and excluded.
+                    if (j + 1 <= m && toks[j+1] != "" && toks[j+1] !~ /^&/) {
+                        print curcwd SEP toks[j+1]
+                    }
+                    continue
+                }
+                if (t ~ /^[0-9]*>>?[^ \t&]/) {
+                    # Attached form (`>file`, `2>file`, `>>file`).
+                    op = t
+                    sub(/^[0-9]*>>?/, "", op)
+                    if (op != "") print curcwd SEP op
+                }
+            }
+        }
+    }'
+}
+
 normalize_abs_path() {
     # Lexically normalize an ABSOLUTE path without touching the filesystem:
     #   - collapse duplicate slashes    (//etc        -> /etc)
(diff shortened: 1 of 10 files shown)


rjwalters:
❌ **Changes Requested**

Excellent, thorough work — the new `extract_write_targets()` idiom parser, the fail-open contract, the `COMMAND_ASK_SCAN` redaction reuse, and the test coverage are all high quality. Local verification is green: `tests/hooks/test-guard-destructive.sh` **434/434**, `test-guard-worktree-paths.sh` **20/20**, `shellcheck -S warning` clean on both edited hooks, CI passing, merge state CLEAN. I also confirmed by direct invocation that the primary #4063 vector is closed: with `CWD` = the main checkout (the daemon-/in-session-dispatch path, where Task-subagent builders share the parent session's cwd per #3719), `sed -i` / `>` / `>>` / `tee` / `cp` writes into the main checkout all correctly **DENY**, while worktree and `/tmp` targets, `git commit -m "a > b"`, and `2>&1` fd-dups correctly **ALLOW**.

**One blocking correctness gap** — the guard uses the wrong "main root," so it diverges from its sibling `guard-worktree-paths.sh` and leaves the PR's headline protection open in the *canonical* builder configuration.

**Root cause:** the new block keys its "resolves inside the main checkout" test on `REPO_ROOT`, which is derived at line ~390 via:
```sh
REPO_ROOT=$(git -C "$CWD" rev-parse --show-toplevel ...)
```
`--show-toplevel` returns the **worktree** root when `CWD` is a linked worktree — *not* the main checkout. So when a builder runs from its own worktree (`cd .loom/worktrees/issue-N`, the documented canonical setup), `REPO_ROOT` = the worktree, and this test:
```sh
case "$_wabs" in
    "$REPO_ROOT"|"$REPO_ROOT"/*) : ;;
    *) continue ;;   # <-- main-checkout target lands here -> ALLOW
esac
```
never recognizes a main-checkout target, so the write is allowed.

`guard-worktree-paths.sh` deliberately avoids this by computing `MAIN_ROOT` via `git rev-parse --git-common-dir`/.. (always the *main* checkout, even from a worktree). This PR's Bash guard is therefore strictly weaker than the Edit/Write guard it is meant to mirror.

**Reproduction** (run against this branch's hook; `$WT` = a real linked worktree, `$MAIN` = main checkout):
```
abs main-checkout path, CWD=main:            DENY   ✅ (incident path — closed)
abs main-checkout path, CWD=linked worktree: ALLOW  ❌ (escape still open)
```
The same miss applies to `cd "$MAIN" && echo x > defaults/hooks/x.sh` issued from a worktree CWD (the `cd` retargets `curcwd`, but `REPO_ROOT` stays the worktree). This is precisely the "denied on Edit/Write → retry via Bash" fallback the PR sets out to close — a builder pinned to its own worktree can still land an absolute-path Bash write in the main checkout.

**Suggested fix** (small, localized — mirror the sibling guard rather than reuse `REPO_ROOT`): derive a true main-checkout root for this block, e.g.
```sh
_WT_MAIN_ROOT="$(cd "$(git -C "$CWD" rev-parse --git-common-dir 2>/dev/null)/.." 2>/dev/null && pwd -P)" || _WT_MAIN_ROOT="$REPO_ROOT"
```
and use `_WT_MAIN_ROOT` (not `REPO_ROOT`) in the `case "$_wabs"` main-checkout test. Keep `resolve_worktree_root`/`_any_managed_worktree_exists` keyed off the same main root so the "a managed worktree exists" gate stays consistent. A `pwd -P` there also sidesteps the macOS `/tmp`→`/private/tmp` symlink mismatch between `--show-toplevel` (symlink-resolved) and `normalize_abs_path` (lexical-only), which is worth a defensive note.

Please also add a regression test to `tests/hooks/test-guard-destructive.sh` for the CWD-in-worktree + absolute-main-path (and `cd $MAIN &&`) case so the asymmetry can't silently return.

Everything else looks great — this is a narrow, well-understood fix on top of otherwise solid work, and the #4063 dispatch-path vector is already closed. Re-request review once the main-root derivation is corrected.

rjwalters:
Addressed the blocking correctness gap.

**Fix** — `defaults/hooks/guard-destructive-generic.sh` (Bash-write confinement block): the "resolves inside the main checkout" test no longer keys on `REPO_ROOT` (`git rev-parse --show-toplevel`, which returns the *worktree* root from a linked worktree). It now derives a true main-checkout root, mirroring the sibling `guard-worktree-paths.sh`:

```sh
_WT_MAIN_ROOT=$(cd "$CWD" && cd "$(git rev-parse --git-common-dir)/.." && pwd -P)   # fail-open to REPO_ROOT
```

`_WT_MAIN_ROOT` is used for both the `case "$_wabs"` containment test and the `resolve_worktree_root` base (so the "a managed worktree exists" gate stays consistent), and in the deny message. `pwd -P` keeps it symlink-resolved, consistent with the git-resolved forms (and sidesteps the macOS `/tmp`→`/private/tmp` mismatch vs. `normalize_abs_path`'s lexical form). Fails open to `REPO_ROOT` when git resolution is unavailable.

**Verified** the exact repro you gave (real linked worktree, `$WT` = worktree CWD, `$MAIN` = main checkout):

| Case | Before | After |
|------|--------|-------|
| abs main-checkout path, CWD=main | DENY | DENY |
| abs main-checkout path, CWD=linked worktree | **ALLOW (bug)** | **DENY** |
| `cd $MAIN && echo x > defaults/…`, CWD=worktree | ALLOW (bug) | DENY |
| write inside the worktree / `/tmp`, CWD=worktree | ALLOW | ALLOW |

**Regression tests** added to `tests/hooks/test-guard-destructive.sh` with a REAL linked-worktree fixture (`make_wt_repo_linked` via `git worktree add`, which reproduces the `show-toplevel` ≠ `git-common-dir/..` divergence a plain subdir cannot): CWD=linked-worktree with absolute-main / `cd $MAIN &&`-relative / `sed -i` main-checkout writes all DENY; worktree and `/tmp` targets from the same CWD still ALLOW.

Also synced the git-tracked installed `.loom/hooks/` copy.

**Suites** (env-isolated from the host's autonomous-mode exports): `test-guard-destructive.sh` **439/439** (was 434, +5 regression), `test-guard-worktree-paths.sh` **20/20**, `test-guard-loom-workflow.sh` **27/27**, `test-guard-destructive-dispatcher.sh` **7/7**. `shellcheck -S warning` clean on both edited hooks.

rjwalters:
✅ **Approved!** (re-review after Doctor pass)

The blocking correctness gap from my prior review (2026-07-28T07:41Z) is resolved. The Bash-write confinement block no longer keys its "resolves inside the main checkout" test on `REPO_ROOT` (`git rev-parse --show-toplevel`, which returns the *worktree* root from a linked worktree). It now derives a true main-checkout root, mirroring the sibling `guard-worktree-paths.sh`:

```sh
_WT_MAIN_ROOT=$(cd "$CWD" && cd "$(git rev-parse --git-common-dir)/.." && pwd -P)   # fail-open to REPO_ROOT
```

`_WT_MAIN_ROOT` is used consistently for the `case "$_wabs"` containment test, the `resolve_worktree_root` base (so the "a managed worktree exists" gate stays keyed off the same root), and the deny message. `pwd -P` keeps it symlink-resolved, sidestepping the macOS `/tmp`→`/private/tmp` mismatch. Fails open to `REPO_ROOT` when git resolution is unavailable — the never-non-zero-exit contract is preserved.

**Verified the exact repro I gave is now closed** (real linked-worktree fixture, CWD=worktree):
- abs main-checkout path → **DENY** (was ALLOW/bug)
- `cd $MAIN && echo x > defaults/…` relative → **DENY** (was ALLOW/bug)
- `sed -i` on abs main-checkout path → **DENY**
- write inside the worktree / `/tmp` → **ALLOW** (correct — no over-blocking regression)

**No new problems.** The +5 regression tests use a REAL `git worktree add` fixture (`make_wt_repo_linked`) that reproduces the `show-toplevel` ≠ `git-common-dir/..` divergence a plain subdir cannot — so the asymmetry can't silently return. The git-tracked installed `.loom/hooks/` copy is synced (the suite's byte-identical assertion passes).

## Test Execution
Env-isolated from this host's autonomous-mode exports:
- `tests/hooks/test-guard-destructive.sh` — ✅ **439/439** (+5 #4210 regression, avg 21ms < 200ms threshold)
- `defaults/hooks/tests/test-guard-worktree-paths.sh` — ✅ **20/20** (incl. `.loom/`↔`defaults/` byte-identical)
- `tests/hooks/test-guard-loom-workflow.sh` — ✅ **27/27**
- `tests/hooks/test-guard-destructive-dispatcher.sh` — ✅ **7/7**
- `shellcheck -S warning` on both edited hooks — ✅ clean

CI: all checks pass (Installer Integration, CLAUDE.md budget, Detect Changes; Rust/MCP jobs skipped — no relevant paths). Merge state: **CLEAN**. `Closes #4178` present.

The security-relevant confinement is now strictly at parity with the Edit/Write guard it mirrors, closing the #4063 Bash-write fallback in the canonical worktree-pinned builder configuration. Nice work.

Issue #3835, opened 2026-07-23
Epic: machine-level Loom install — stop copying the implementation into every consumer repo


rjwalters:
## Problem

Installing Loom into a consumer repo today copies a near-complete implementation into it (`.loom/scripts`, `.loom/roles`, `.claude/commands/loom/*`, `.claude/agents/loom-*`, hooks, docs, tests — 100+ files, committed). This conflates four different concerns in `.loom/`:
1. **Reusable Loom implementation** (should be shared, not per-repo)
2. **Project config** (repo-specific, tracked)
3. **Host-local config** (machine, untracked)
4. **Runtime state** (ignored)

It also means every consumer repo re-publishes the whole implementation and drifts per-repo (the recurring self-install-dogfooding pain).

## Target architecture — install once, machine-level

```
~/.local/share/loom/        private/shared Loom checkout + runtime
~/.local/bin/loom           CLI entry point (start|stop|status|sweep|update)
~/.claude/skills/loom/      → symlink into the checkout (personal skill)
~/.claude/agents/loom-*.md  → symlinks (personal agents)
~/.claude/settings.json     user-level hook scripts
<user-scoped MCP>           mcp-loom → ~/.local/share/loom/mcp-loom (kills the #3803 stale-dist drift)
~/.loom/                    daemon socket + token pool
```
Consumer repos then track **only** project-specific config:
```
.loom-project/   project.json (enabled agents, buildGate, guard toggles, label maps, model overrides), roles/, rules/, README
CLAUDE.md        public project context
loom.sh          tiny shim → `exec loom "$@"`
.loom-local/     ignored runtime state
```
Effective config = **private defaults + tracked `.loom-project` + ignored `.loom-local`**.

## Why (regardless of licensing)
This is the clean fix for the long-standing `.loom/` conflation and per-repo drift; `loom update` updates all repos at once; a routine `git add -A` can't publish/duplicate the implementation.

## Suggested phases (Champion/Architect to decompose)
- **Phase 1** — installer gitignore/`--local` mode: don't commit the implementation into consumer repos (first concrete slice; see companion issue).
- **Phase 2** — `.loom-project/` schema + resolution (private defaults + project config + local state merge).
- **Phase 3** — machine-level runtime: `~/.local/bin/loom` CLI + user daemon + user-scoped MCP registration.
- **Phase 4** — personal skills/agents via symlink into the checkout (no copy).
[...]

PR #4236, opened 2026-07-28
fix(hooks): migrate cold-path guard config readers to config-resolver (Option A)

### LINKED ISSUE #4063: [#4047] Migrate defaults/hooks config readers — decide fast-path carve-out first (must not regress #3687)

rjwalters:
Closes #4063

## Decision: Option A (carve out the hot path, migrate only the cold path)

Per the issue's curator-recommended and champion-approved decision, this PR
migrates the **seven cold-path** config readers in
`guard-destructive-generic.sh` — `sqlDdl`, `cloudCli`, `reversibleGh`,
`decisionLog`, `rmScope`, `forceScope`, `worktree.root` — plus the
`decisionLog` reader in `guard-loom-workflow.sh`, onto the shared
`loom_config_get()` / `loom_resolve_config()` (`defaults/scripts/lib/config-resolver.sh`,
#4039). This gives those readers the full 4-tier config chain (legacy
`.loom/config.json` + the `.loom-project`/`.loom-local` tiers) in lockstep with
loom-daemon and loom_tools, instead of a bespoke single-file `jq` read.

The **three #3687 read-only fast-path readers** — `fastpath_config_file`,
`fastpath_enabled`, `fastpath_extra_admits` — are **explicitly NOT migrated**.
They keep their fork-free / single-cached-`jq` implementation, with an
in-code `CARVE-OUT (#4063)` comment explaining why (this was already true on
main and is unchanged by this PR — only a comment was added there).

Option B (cache the merge in config-resolver.sh) doesn't solve the problem —
a cached merge still needs a `repo_root`, and the fast path deliberately
exits *before* `REPO_ROOT` is resolved (see control-flow note below). Option C
(teach the resolver a fork-free upward-walk root) would break the documented
three-language (Rust/Python/Bash) conformance-fixture contract in
`config-resolver.sh`'s header. Option A is the only one that preserves both
#3687 and the conformance guarantee.

## Where the config-resolver source line lives (and why)

`config-resolver.sh` is `source`d **strictly below** the fast-path dispatch
and the `REPO_ROOT=$(git -C "$CWD" rev-parse --show-toplevel ...)` resolution
— i.e. only on the cold path. A fast-pathed command (e.g. `git status`) never
reaches that line, so it pays **zero added cost** — not even the `[[ -f ]]`
stat that guards the `source`. This was deliberately placed there (not at the
top of the file, which an earlier draft of this change had done) specifically
so the fast path's fork/latency budget is untouched.

## Fork-count analysis (the acceptance criterion's primary evidence)

This is a static, code-verifiable claim, not a measurement subject to host
noise:

- **Before and after, the fast path costs the same 0 forks for path
  discovery** (`fastpath_config_file`'s upward bash-builtin walk) **+ 1
  cached `jq` fork** for `fastpath_enabled` — unchanged by this PR. No `git
  rev-parse` and no additional fork was introduced above the fast-path exit
  points (verified by inspection: `fastpath_builtin_admits`/`fastpath_enabled`
  calls and the `exit 0` sites are untouched; the new `source` line is now
  positioned strictly after them).
- The **cold-path** readers went from 1 fork each (bespoke `jq -r` read of
  `.loom/config.json` only) to up to 3 forks each today via
  `loom_config_get()` (1 soft-read of the legacy tier + 1 `jq -n` merge + 1
  `getpath`), rising toward more as Epic #3835 lands the upper config tiers.
  This is the accepted, scoped-in cost of the migration — these readers only
  fire on catastrophic/ask-tier commands, not on every Bash call.

## Wall-clock measurement (best-effort; host was under heavy unrelated load)

Ran the existing 100x `git status` fast-path benchmark (and an interleaved
5-round before/after comparison) against the pre-migration
`guard-destructive-generic.sh` (from `main`) and the post-migration version.
**Caveat**: this measurement was taken on a shared dev host with `uptime`
load averages of 30–40 throughout (multiple unrelated concurrent Loom sweeps
running `cargo build` in a sibling repo), so absolute numbers are noisy in
both directions; the fork-count analysis above is the load-bearing evidence.
Interleaved rounds (ms/call, `n=50` each):

| round | before (main) | after (this PR) |
|---|---|---|
| 1 | 146 | 259 |
| 2 | 176 | 155 |
| 3 | 166 | 174 |
| 4 | 307 | 318 |
| 5 | 332 | 108 |
| **mean** | **225** | **203** |

No consistent before/after direction — the two are within the same noise
band, consistent with the fork-count argument (no new forks added to the fast
path). The in-repo test suite's own 10x-iteration performance check (informational,
not gating under load) also passed the same way before and after.

## Polarity truth table

Every migrated reader kept its **exact prior bash-side polarity check**
rather than trusting `loom_config_get`'s null-collapses-to-default behavior
blindly, per the issue's polarity-trap warning:

| Reader | Default | Only this flips it |
|---|---|---|
| `sqlDdl` | ON | explicit `false` |
| `cloudCli` | ON | explicit `false` |
| `reversibleGh` | OFF (inverse) | explicit `true` |
| `decisionLog` (both guards) | OFF (inverse) | explicit `true` |
| `rmScope` | `"repo"` | `"off"`/`"permissive"` |
| `forceScope` | `"all"` | `"protected"`/`"off"` |
| `worktree.root` | in-repo default | absolute-path string |

Added a truth-table test section to `tests/hooks/test-guard-destructive.sh`
covering the two shapes not already exercised by the existing suite (explicit
`null` and a non-boolean/out-of-range value) for every migrated reader — the
existing suite already covered absent/true/false/malformed-JSON.

## Testing

All run with `LOOM_FORCE_SCOPE` and `LOOM_GUARD_DECISION_LOG` explicitly
unset — this dev sandbox has both set in its ambient shell environment
(operator-configured env overrides for this specific repo's own
self-hosting), which otherwise masks the config-driven defaults these tests
verify. Confirmed via `git stash` that these two vars caused the same
apparent "failures" on pre-migration `main` — not a regression from this PR.

- `tests/hooks/test-guard-destructive.sh`: **453/453 passed** (442 existing +
  11 new truth-table assertions)
- `tests/hooks/test-guard-loom-workflow.sh`: **27/27 passed**
- `defaults/scripts/tests/test-config-resolver.sh`: **14/14 passed**
  (unaffected — config-resolver.sh itself was not modified)
- `defaults/scripts/tests/test-guard-hook-schema.sh`: **11/11 passed**
- `./.loom/scripts/resync-installed.sh --dry-run`: exits **0** — 150
  unchanged, 0 skipped; `defaults/hooks/*` and `.loom/hooks/*` stay
  byte-identical for both touched files
- `shellcheck -x` on both touched hook files: same finding count
  before/after (4 pre-existing informational SC2016 notices, no new
  findings)

## Affected files

- `defaults/hooks/guard-destructive-generic.sh` / `.loom/hooks/guard-destructive-generic.sh`
  (kept byte-identical)
- `defaults/hooks/guard-loom-workflow.sh` / `.loom/hooks/guard-loom-workflow.sh`
  (kept byte-identical)
- `tests/hooks/test-guard-destructive.sh` (new truth-table section)
- `defaults/hooks/guard-destructive.sh` (the dispatcher) — **untouched**, as
  expected (it holds zero config readers post-#4107)
- `defaults/scripts/lib/config-resolver.sh` — **untouched** (Option A doesn't
  need caching)

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -223,6 +223,28 @@ fi
 # commands (each entry is a full-generality bypass for that command word).
 # =============================================================================
 
+# CARVE-OUT (#4063): the three fast-path readers below — fastpath_config_file,
+# fastpath_enabled, fastpath_extra_admits — are deliberately NOT migrated to the
+# shared config-resolver.sh (loom_config_get / loom_resolve_config) that the
+# cold-path toggles use. This is intentional and preserves issue #3687's fork
+# budget:
+#   * This block can `exit 0` (silent allow) BEFORE REPO_ROOT is resolved — the
+#     `git -C "$CWD" rev-parse --show-toplevel` fork lives strictly below the
+#     fast-path dispatch. loom_resolve_config REQUIRES a repo_root as its first
+#     argument, so routing these through it would force hoisting that git
+#     rev-parse above the fast-path exit, directly regressing #3687 (which
+#     removed it from the pre-admission path).
+#   * fastpath_config_file is a fork-free bash-builtin upward directory walk;
+#     fastpath_enabled / fastpath_extra_admits each cost exactly one CACHED jq.
+#     loom_config_get is uncached and forks jq once per existing tier file plus a
+#     merge (~3 forks today, more as Epic #3835 lands upper tiers) on EVERY call —
+#     a 3x+ regression on a hook that fires before every Bash tool call. Caching
+#     the merge (Option B) still needs a repo_root, and teaching the resolver a
+#     fork-free upward-walk root discovery (Option C) would break the documented
+#     three-language (Rust/Python/Bash) conformance-fixture contract in
+#     config-resolver.sh's header. So the fast path keeps its direct single-jq
+#     read of the nearest .loom/config.json. See #4063 for the full analysis.
+#
 # Locate the nearest .loom/config.json by walking up from CWD, fork-free (no
 # git rev-parse). Cached. Best-effort: empty when none is found.
 _FASTPATH_CFG_FILE=""
@@ -393,6 +415,25 @@ elif [[ -n "$CWD" ]]; then
     log_hook_error "cwd does not exist: $CWD — skipping repo root resolution"
 fi
 
+# Shared config-tier resolver (#4063). Source defaults/scripts/lib/config-resolver.sh
+# — deliberately sourced HERE, strictly below the #3687 fast-path dispatch and
+# REPO_ROOT resolution above, so a fast-pathed command pays ZERO added cost (not
+# even the `[[ -f ]]` stat below) — this is the cold path only. At runtime
+# SCRIPT_DIR is the installed hook dir (.loom/hooks/), and .loom/scripts is a
+# symlink to defaults/scripts, so ../scripts/lib resolves; in the test harness
+# SCRIPT_DIR is defaults/hooks/ and the sibling path resolves directly. The
+# COLD-PATH toggle readers below (sql/cloud/reversibleGh/decisionLog/rmScope/
+# forceScope + worktree.root) call loom_config_get through this so a single code
+# path reads the full tier chain (legacy .loom/config.json plus the #4039
+# project/local tiers) and stays byte-for-byte in lockstep with loom-daemon and
+# loom_tools. Best-effort: a missing/unsourceable lib leaves loom_config_get
+# undefined, and each reader's `|| <default>` fallback then preserves that
+# guard's safe default, so the guard never breaks.
+if [[ -f "$SCRIPT_DIR/../scripts/lib/config-resolver.sh" ]]; then
+    # shellcheck source=/dev/null
+    source "$SCRIPT_DIR/../scripts/lib/config-resolver.sh" 2>/dev/null || true
+fi
+
 # =============================================================================
 # SQL DDL/DML guard toggle — default ON.
 #
@@ -417,14 +458,18 @@ fi
 _SQL_GUARD_CACHE=""
 sql_guard_enabled() {
     if [[ -z "$_SQL_GUARD_CACHE" ]]; then
-        local enabled=true
-        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
-            # jq // is alternative-on-null, not default-on-missing, so use
-            # if/then/else to treat only an explicit `false` as disabled (a
-            # missing guards.sqlDdl key stays on). On malformed JSON jq exits
-            # non-zero and the `||` fallback restores the guard-ON default.
-            enabled=$(jq -r 'if .guards.sqlDdl == false then "false" else "true" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || enabled=true
-            [[ -n "$enabled" ]] || enabled=true
+        local enabled=true raw
+        if [[ -n "$REPO_ROOT" ]]; then
+            # Migrated to the shared tier resolver (#4063). loom_config_get
+            # collapses null and missing to the default, so we KEEP the exact
+            # polarity in bash: only an explicit boolean `false` disables — a
+            # missing/null key OR a non-boolean value (e.g. "yes") stays guard-ON,
+            # matching the old `.guards.sqlDdl == false` test. `|| raw=true` also
+            # covers config-resolver.sh failing to source (loom_config_get unset)
+            # and malformed JSON (the resolver soft-reads a bad tier as {} → the
+            # key resolves absent → default "true").
+            raw=$(loom_config_get "$REPO_ROOT" "guards.sqlDdl" "true" 2>/dev/null) || raw=true
+            [[ "$raw" == "false" ]] && enabled=false
         fi
         # Env override wins over config.
         case "${LOOM_GUARD_SQL:-}" in
@@ -460,14 +505,14 @@ sql_guard_enabled() {
 _CLOUD_GUARD_CACHE=""
 cloud_guard_enabled() {
     if [[ -z "$_CLOUD_GUARD_CACHE" ]]; then
-        local enabled=true
-        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
-            # jq // is alternative-on-null, not default-on-missing, so use
-            # if/then/else to treat only an explicit `false` as disabled (a
-            # missing guards.cloudCli key stays on). On malformed JSON jq exits
-            # non-zero and the `||` fallback restores the guard-ON default.
-            enabled=$(jq -r 'if .guards.cloudCli == false then "false" else "true" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || enabled=true
-            [[ -n "$enabled" ]] || enabled=true
+        local enabled=true raw
+        if [[ -n "$REPO_ROOT" ]]; then
+            # Migrated to the shared tier resolver (#4063), same polarity as
+            # sql_guard_enabled(): only an explicit boolean `false` disables; a
+            # missing/null key, a non-boolean value, or malformed JSON stays
+            # guard-ON via the "true" default and the `|| raw=true` fallback.
+            raw=$(loom_config_get "$REPO_ROOT" "guards.cloudCli" "true" 2>/dev/null) || raw=true
+            [[ "$raw" == "false" ]] && enabled=false
         fi
         # Env override wins over config.
         case "${LOOM_GUARD_CLOUD:-}" in
@@ -514,14 +559,14 @@ cloud_guard_enabled() {
 _REVERSIBLE_GH_GUARD_CACHE=""
 reversible_gh_guard_enabled() {
     if [[ -z "$_REVERSIBLE_GH_GUARD_CACHE" ]]; then
-        local enabled=false
-        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
-            # jq // is alternative-on-null, not default-on-missing, so use
-            # if/then/else to treat only an explicit `true` as enabled (a
-            # missing guards.reversibleGh key stays off). On malformed JSON jq
-            # exits non-zero and the `||` fallback restores the guard-OFF default.
-            enabled=$(jq -r 'if .guards.reversibleGh == true then "true" else "false" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || enabled=false
-            [[ -n "$enabled" ]] || enabled=false
+        local enabled=false raw
+        if [[ -n "$REPO_ROOT" ]]; then
+            # Migrated to the shared tier resolver (#4063). INVERSE polarity of
+            # sql/cloud: only an explicit boolean `true` enables the ask; a
+            # missing/null key, a non-boolean value, or malformed JSON stays
+            # guard-OFF via the "false" default and the `|| raw=false` fallback.
+            raw=$(loom_config_get "$REPO_ROOT" "guards.reversibleGh" "false" 2>/dev/null) || raw=false
+            [[ "$raw" == "true" ]] && enabled=true
         fi
         # Env override wins over config.
         case "${LOOM_GUARD_REVERSIBLE_GH:-}" in
@@ -563,12 +608,14 @@ reversible_gh_guard_enabled() {
 _DECISION_LOG_CACHE=""
 decision_log_enabled() {
     if [[ -z "$_DECISION_LOG_CACHE" ]]; then
-        local enabled=false
-        if [[ -n "$REPO_ROOT" && -f "$REPO_ROOT/.loom/config.json" ]]; then
-            # Only an explicit `true` enables; a missing key or malformed JSON
-            # (jq non-zero, caught by ||) stays OFF — inverse of sql_guard_enabled().
-            enabled=$(jq -r 'if .guards.decisionLog == true then "true" else "false" end' "$REPO_ROOT/.loom/config.json" 2>/dev/null) || enabled=false
-            [[ -n "$enabled" ]] || enabled=false
+        local enabled=false raw
+        if [[ -n "$REPO_ROOT" ]]; then
+            # Migrated to the shared tier resolver (#4063). INVERSE polarity like
+            # reversible_gh_guard_enabled(): only an explicit boolean `true`
+            # enables; a missing/null key, a non-boolean value, or malformed JSON
+            # stays OFF via the "false" default and the `|| raw=false` fallback.
+            raw=$(loom_config_get "$REPO_ROOT" "guards.decisionLog" "false" 2>/dev/null) || raw=false
+            [[ "$raw" == "true" ]] && enabled=true
         fi
         # Env override wins over config.
         case "${LOOM_GUARD_DECISION_LOG:-}" in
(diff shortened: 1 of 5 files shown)


rjwalters:
## Judge review — approve

Reviewed against issue #4063's acceptance criteria. Verdict: **approve**.

### Verification performed

1. **Scope (Option A carve-out)** — confirmed only the 7 cold-path readers
   (`sqlDdl`, `cloudCli`, `reversibleGh`, `decisionLog` ×2 guards, `rmScope`,
   `forceScope`, `worktree.root`) were migrated onto `loom_config_get()`. The
   three fast-path readers (`fastpath_config_file`, `fastpath_enabled`,
   `fastpath_extra_admits`) are untouched except for the explanatory
   `CARVE-OUT (#4063)` comment. `git diff main...` on the `REPO_ROOT=$(git ...
   rev-parse ...)` line confirms **no change** — no new fork was hoisted above
   the fast-path exit points.

2. **Polarity correctness** — read `defaults/scripts/lib/config-resolver.sh`'s
   `loom_config_get`/`_loom_config_soft_read` implementation directly and
   traced each migrated reader's bash-side polarity check against it:
   - `sqlDdl`/`cloudCli`: default ON, only explicit `false` disables — preserved.
   - `reversibleGh`/`decisionLog` (both guards): default OFF (inverse), only
     explicit `true` enables — preserved.
   - `rmScope`: default `"repo"`, only `"off"`/`"permissive"` opts out —
     preserved (string branch kept in bash, not trusted to `loom_config_get`'s
     null-collapse).
   - `forceScope`: default `"all"`, only `"protected"`/`"off"` opt out — preserved.
   - `worktree.root`: default empty→in-repo fallback, only an absolute-path
     string is admitted — preserved.
   - Malformed JSON at any tier soft-fails to `{}` in `_loom_config_soft_read`,
     which correctly resolves through to each reader's `||` fallback — matches
     pre-migration behavior.

3. **Byte-identical installed copies** — `diff .loom/hooks/guard-destructive-generic.sh
   defaults/hooks/guard-destructive-generic.sh` and the `guard-loom-workflow.sh`
   pair are identical on the PR branch. `./.loom/scripts/resync-installed.sh --dry-run`
   exits 0 (150 unchanged, 0 skipped).

4. **Truth-table tests** — the new section in `tests/hooks/test-guard-destructive.sh`
   covers the `null` and non-boolean/out-of-range shapes for all 7 migrated
   readers (absent/true/false/malformed were already covered by the pre-existing
   suite). Ran the full suite locally (with `LOOM_FORCE_SCOPE`/`LOOM_GUARD_DECISION_LOG`
   unset, per the builder's noted ambient-env caveat, which I reproduced and
   confirmed is unrelated to this change): **453/453 passed**, including all 14
   new truth-table assertions.

5. **PR body fork/latency accounting** — present and matches the issue's
   acceptance criterion: static fork-count analysis (0 forks / 1 cached `jq`
   fast path, unchanged; cold path 1→3 forks) plus a best-effort wall-clock
   comparison with an honest noisy-host caveat.

6. **Full suite run** (env vars unset):
   - `tests/hooks/test-guard-destructive.sh`: 453/453 ✅
   - `tests/hooks/test-guard-loom-workflow.sh`: 27/27 ✅
   - `defaults/scripts/tests/test-config-resolver.sh`: 14/14 ✅
   - `defaults/scripts/tests/test-guard-hook-schema.sh`: 11/11 ✅
   - `shellcheck -x` on both touched files: same 4 pre-existing informational
     findings (3× SC2016, 1× SC2329), no new findings.

7. **CI / merge state** — `gh pr checks 4236` green (Installer Integration
   Tests, CLAUDE.md Line Budget, Detect Changes all pass; Rust jobs correctly
   skipped for a bash-only change). `mergeStateStatus` is `CLEAN`.

No issues found. This is exactly the disciplined, scoped migration the issue
called for — no polarity flips, no fast-path regression, tests for every
input shape.

Issue #4195, opened 2026-07-28
Harvest (fork): token-pool load-aware selection (waterfall fill + 5h-headroom spread) — reconcile with #3909


rjwalters:
## Harvest candidate from gpeyton/loom fork (bucket a of #4165)

Token-pool selection heuristic from the fork-divergence triage (#4165). Runtime-neutral, but **overlaps upstream's existing spread work — reconcile before adopting.**

### What
Two load-aware token-selection heuristics from the fork:
- **Waterfall fill** — preserve ranked order but skip accounts already >70% loaded on the 5h window, burning soonest-to-expire capacity first and overflowing to the next account.
- **Spread by 5h headroom** — order ranked selection least-loaded-first to avoid a single-account mass-stall.

### Fork commit(s)
- `283de8e3` — `tokens/select: waterfall fill — ranked order preserved, skip >70% 5h-loaded`
- `20961dd9` — `tokens/select: spread ranked selection by 5h headroom (least-loaded first)`

### Files
- `loom-tools/src/loom_tools/tokens/select.py`

### IMPORTANT — reconcile with upstream spread first
Upstream **already has** a spread mechanism: the persistent rotation cursor (#3736) + `LOOM_TOKEN_SPREAD_TOP_N` / `tokens.spreadTopN` window cap (#3909), plus the `_HEALTHY_STATUSES` allowlist (#3991). The fork's heuristics are a *different* axis (5h-window load rather than a rotation cursor). This is a **design-comparison** task: decide whether the fork's load-aware fill improves on, conflicts with, or should merge into the existing rotation-cursor spread — do not layer a second, contradictory selection strategy. Excludes the fork's `provider`-field selection commits, which fold into #4167 (multi-runtime) instead.

Inspect: `git show 283de8e3 20961dd9` against `gpeyton/loom`.

---

## Curator Enhancement (2026-07-28, verified against `origin/main`)

[...]

Issue #4122, opened 2026-07-27
Daemon-dispatched sweeps bypass claude-wrapper.sh — weekly-limit exhaustion is charged to the ISSUE, never to the account


rjwalters:
Split out of #4007 during curation (2026-07-27). #4007's first comment re-diagnosed its re-dispatch loop as **Claude account weekly-limit exhaustion**, not the #3887 startup race. Verification against `origin/main` (`e6522b71`) confirms the mechanism, and it is worse than the original report assumed.

## Problem

Loom **does** have weekly-limit detection and account rotation. It lives in `defaults/scripts/claude-wrapper.sh`:

- `:909-928` `is_account_exhaustion()` — matches `hit your (limit|session limit|weekly limit)|monthly usage limit|out of extra usage`
- `:1176-1183`, `:1291-1293` — output/startup monitors grep for the weekly-limit banner and kill `claude`
- `:1056-1108` `rotate_exhausted_account()` — `mark_bad(...)`, then re-select and re-export `CLAUDE_CODE_OAUTH_TOKEN`
- `:1752-1773` — rotation loop; retries do not consume `MAX_RETRIES`

**The daemon never takes that path.** `defaults/scripts/spawn-claude.sh` invokes the wrapper only under `--use-wrapper` (flag parsed at `:132`, used at `:367-372`); the default is `:381` `exec claude "${PASSTHROUGH_ARGS[@]}"`. The daemon builds its argv at `loom-daemon/src/sweep_registry.rs:2458-2500` (`-p`, `--model`, `--effort`, …) — **no `--use-wrapper`**. Same for the role runner (`loom-daemon/src/role_runner.rs:350-352`).

Grep for `weekly limit` / `ACCOUNT_POOL_EXHAUSTED` / `RATE_LIMIT_ABORT` under `loom-daemon/src/` returns exactly one hit — a test fixture string in `tokens_pool/bad_tokens.rs:352`.

So on the daemon-dispatched sweep path: a child that dies instantly on a weekly limit is **never attributed to the account**.

## Consequence: the crash is charged to the ISSUE instead

Insta-crash classification is cause-blind — `loom-daemon/src/sweep_registry.rs:3009-3012` and `:3062-3065`:

```rust
let insta_crash = duration_sec < self.quarantine_config.insta_crash_secs
    && exit_code != Some(0);
```

[...]

PR #4242, opened 2026-07-28
fix(tokens): heal stale exhausted markers, neutral rate-limit label, unblock no-op fails loud (#4212)

### LINKED ISSUE #4212: tokens: stale exhausted markers outlive the 5h reset; weekly-limit misclassification; unblock default-scope no-op
## Context

Live incident 2026-07-28 (~04:07–04:19Z): a 5h rate-limit storm insta-crashed a wave of sweeps. Three compounding token-pool defects were observed:

1. **Stale markers outlive the reset.** All 7 accounts were written to `.bad_tokens` as `exhausted: weekly-limit`. Hours later the live probe (`loom-daemon tokens check`) showed 3 of them fully recovered (5h at 0–7%), but the pool still refused selection — every dispatch died with exit 78 ("All 7 tokens are marked bad") until a manual `loom-daemon tokens unblock <name> --all-reasons` per account. The same stale set independently poisoned a second repo's pool (vibesql), where it also triggered 5 insta-crash quarantines.
2. **Misclassification.** The entries say `weekly-limit`, but the recovered accounts were at 25–48% **7d** utilization — the storm was 5h-window exhaustion. A `weekly-limit` label on a 5h event materially misleads recovery decisions (a real weekly limit would justify waiting days, not hours).
3. **`unblock` default scope silently no-ops on these entries.** `loom-daemon tokens unblock <name>` printed "No matching entries removed (use --all-reasons…)" and exited zero; the operator dispatched three sweeps on the still-poisoned pool and all three insta-crashed before the `--all-reasons` retry.

- [ ] Selection (or a periodic daemon task) re-validates `exhausted`-reason entries against the live probe or a recorded 5h-reset horizon; a recovered account re-enters the pool without operator action
- [ ] Insta-crash classification distinguishes 5h-window exhaustion from weekly-limit (or uses a neutral `rate-limited` reason when the window is unknown)
- [ ] `tokens unblock` with entries excluded by the default scope exits non-zero (or prompts), naming the excluded entries — a no-op that looks like success is the failure mode that cost three dispatches
- [ ] Pool-state fixes apply per-pool (per-repo and shared) — the same staleness must not need a second manual round in each managed workspace

rjwalters:
Closes #4212

## Problem

The 2026-07-28 rate-limit storm exposed three compounding token-pool defects. All fixes land in **both** the Python (live spawn path) and Rust implementations, keeping the byte-for-byte conformance pair intact.

### 1. Stale `exhausted` markers outlived the reset (AC1 + AC4)
The Rust `is_bad` already expired non-auth ("exhausted") entries after a cooldown (#4122), but the **Python `is_bad` — which the live spawn path (`spawn-claude.sh`) actually calls — had no reason-aware cooldown at all**: it blocked permanently until `cleanup_bad_tokens` pruned the line. So a recovered account stayed out of the pool until a manual `unblock` per account. This is exactly the Python/Rust lockstep gap #4195 warned about.

Ported the exact reason-aware cooldown to Python `is_bad`:
- auth entries (401 / OAuth / expired / blocked) block permanently;
- exhaustion entries stop blocking past `LOOM_TOKEN_EXHAUSTION_COOLDOWN_SECS` (default 6h);
- a malformed/missing timestamp fails **closed** (never silently un-blocks).

Shared the auth-reason regex + added `exhaustion_cooldown_secs()`/`is_auth_reason()` in `bad_tokens.py` (previously duplicated in `cli.py`). The check runs at selection time against whichever pool `resolve_tokens_dir` picks, so per-repo and shared pools both self-heal with no second manual round (AC4).

### 2. Misclassification: 5h window labeled "weekly-limit" (AC2)
The daemon reaper's exhaustion classifier labeled every banner match `weekly-limit`, even a 5h-window "hit your limit" death. A `weekly-limit` label on a 5h event materially misleads recovery (weekly = wait days; 5h = wait hours). The window is unknowable from the banner, so the signature now uses a neutral **`rate-limited`** label (`sweep_registry.rs`).

### 3. `unblock` default scope silently no-op'd (AC3)
`unblock <name>` on an account whose only entry was non-auth printed "No matching entries removed" and exited **0** — the operator saw success and dispatched three sweeps onto the still-poisoned pool, all insta-crashing. Both CLIs now track the left-behind accounts, name them, add an `excluded` field to the JSON output, and **exit `3`**.

## Tests
- Python units: cooldown expiry vs. permanent auth, fresh-entry still blocks, malformed-ts fails closed, expired line doesn't mask a later blocking line, env override, `is_auth_reason`.
- Rust unit: `unblock_default_scope_reports_excluded_non_auth`; updated existing unblock tests to the `UnblockOutcome` struct.
- Cross-language conformance: new `excluded` JSON field asserted identical; new `test_unblock_default_scope_excluded_exit3_identically` (both exit 3, both name the account).
- `cargo test` (bad_tokens + all 184 sweep_registry tests) and the full conformance suite pass; clippy clean.

### DIFF
--- a/.loom/docs/token-pool.md
+++ b/.loom/docs/token-pool.md
@@ -321,17 +321,33 @@ loom-tokens pin remove agent-3    # Remove
 loom-tokens pin status            # Show current allowlist
 loom-tokens unpin                 # Delete allowlist (back to full pool)
 
-loom-tokens unblock agent-1       # Remove one entry from .bad_tokens
-loom-tokens unblock --all         # Clear .bad_tokens entirely
+loom-tokens unblock agent-1              # Drop agent-1's AUTH entries only
+loom-tokens unblock agent-1 --all-reasons  # Also drop its exhausted/rate-limited entries
 ```
 
 **Validation**: `pin` accepts only exact bootstrapped account names —
 substring/fuzzy matches are rejected. The allowlist is sorted, deduplicated, and
 `mkdir`-lock guarded so concurrent operator commands don't drop entries.
 
-**Reason-aware bad-token TTL**: bad-tokens entries with reason `auth` (401) ignore
-`LOOM_TOKENS_BAD_TTL` (default 21600s = 6h) and persist until `loom-tokens
-unblock`. Other reasons expire automatically.
+**`unblock` default scope fails loudly on left-behind entries (#4212)**: the
+default scope removes only `auth` entries (a broken credential); transient
+`exhausted`/`rate-limited` entries clear themselves. When the named account has
+*only* a non-auth entry, the default scope leaves it in place — and rather than
+print "No matching entries removed" and exit `0` (the pre-#4212 silent no-op that
+let an operator dispatch onto a still-poisoned pool), `unblock` now **names the
+still-blocked accounts and exits `3`**. Re-run with `--all-reasons` to drop them,
+or wait for the cooldown to expire them automatically.
+
+**Reason-aware bad-token TTL**: bad-tokens entries whose reason is `auth` (401 /
+OAuth / expired / blocked) persist until `loom-tokens unblock`. Non-auth
+("exhausted"/"rate-limited") entries stop blocking selection once they age past
+the exhaustion cooldown (`LOOM_TOKEN_EXHAUSTION_COOLDOWN_SECS`, default 21600s =
+6h) — enforced at selection time in **both** the Rust daemon and the live Python
+spawn path (`is_bad`), so a recovered account re-enters the pool with no operator
+action even before `cleanup_bad_tokens` prunes the line from disk. Before #4212
+only the Rust path honored the cooldown, so on the live Python spawn path a stale
+`exhausted` marker outlived its rate-limit reset and wedged the account until a
+manual `unblock`.
 
 **Auto-unpin** (`failure_counts`): the wrapper tracks consecutive
 `TOKEN_EXHAUSTED` failures per account in `.loom/tokens/.failure_counts` (JSON).
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -4047,22 +4047,56 @@ fn handle_tokens_command(action: TokensAction) -> Result<()> {
                 std::process::exit(1);
             }
 
-            let (removed, kept) =
+            let outcome =
                 bad_tokens::unblock(&ws, &validated, all_reasons).map_err(|e| anyhow!(e))?;
+            let removed = outcome.removed;
+            let kept = outcome.kept;
+            let excluded = outcome.excluded;
             for name in &validated {
                 let _ = failure_counts::record_success(&ws, name);
             }
 
             if json {
-                println!("{}", serde_json::json!({ "removed": removed, "kept": kept }));
-            } else if removed > 0 {
-                let plural = if removed == 1 { "y" } else { "ies" };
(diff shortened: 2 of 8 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed against all four acceptance criteria on #4212 and the Python+Rust lockstep constraint. This is a clean, well-scoped fix for the 2026-07-28 rate-limit storm defects.

### Correctness — all three defects fixed, lockstep honored

**AC1 + AC4 (stale `exhausted` markers outlive the reset):** The reason-aware cooldown that already lived in Rust `tokens_pool::bad_tokens::is_bad` (#4122) is now ported to Python `is_bad` — the live spawn path (`spawn-claude.sh`) that actually gates selection. I diffed both implementations line-for-line and they are behaviorally byte-identical:
- same auth regex (`\b(401|oauth|auth(entication)?|unauthorized|token[_\s]?expired|expired|blocked)\b`, case-insensitive);
- same timestamp format `%Y-%m-%dT%H:%M:%SZ`;
- same fail-closed on unparseable timestamp (never silently un-blocks);
- same "expired line does NOT short-circuit — a later line may still block" semantics.

Because the check runs at selection time against whatever `resolve_tokens_dir` picks, per-repo and shared pools both self-heal with no second manual round (AC4). Nice touch: the auth regex was previously duplicated in `cli.py`; it's now hoisted to `bad_tokens.py` (`is_auth_reason` / `exhaustion_cooldown_secs`) and imported, so `is_bad` and `unblock` share ONE definition — that removes exactly the kind of intra-language drift that seeds lockstep bugs.

**AC2 (5h window mislabeled `weekly-limit`):** `exhaustion_signatures()` in `sweep_registry.rs` now emits the neutral `rate-limited` label, with a clear comment on why the window is unknowable from the banner. This classifier is Rust-only (no Python counterpart), so there is no lockstep obligation here — correctly a one-sided change. Existing classifier tests updated.

**AC3 (`unblock` default-scope silent no-op):** Both CLIs now track left-behind non-auth accounts (first-seen order, de-duplicated), add an `excluded` field to JSON output, name them on stderr, and exit `3`. The Rust `unblock` return type moved from a `(usize, usize)` tuple to a named `UnblockOutcome` struct — a good readability improvement, and all call sites/tests updated.

### Tests
- Python units (`test_bad_tokens.py`): cooldown expiry vs. permanent auth, fresh-entry still blocks, malformed-ts fails closed, expired line doesn't mask a later blocking line, env override, `is_auth_reason`. All pass.
- Cross-language conformance (`test_rust_conformance.py`): `excluded` field asserted identical; new `test_unblock_default_scope_excluded_exit3_identically` — both implementations exit `3` and emit `{"removed":1,"kept":1,"excluded":["b"]}`. This test actually shells out to the compiled Rust CLI, so it is live proof the two binaries agree.
- Local run: **28 passed** (`test_bad_tokens.py` + `test_rust_conformance.py`).
- CI: **Rust Unit Tests pass**, Rust Linting & Build (clippy) pass, CLAUDE.md budget pass. `mergeStateStatus: CLEAN`, mergeable.

### Docs
`token-pool.md` updated for the new `unblock` default-vs-`--all-reasons` scope, the exit-3 behavior, and the two-language cooldown.

No blocking concerns. The lockstep pair stays intact, conformance coverage was extended for the new behavior, and the fix maps 1:1 onto the filed ACs. Approving.

PR #4253, opened 2026-07-28
fix(guards): retier cloud credential/resource deletion to the ungated ask tier (#4216)

### LINKED ISSUE #4216: guards: cloudCli destructive ops hard-block with no in-session override — propose ask-tier for interactive sessions
## Context

During 2026-07-28 credential rotation, the cloud guard hard-BLOCKED `aws iam delete-access-key` — where the operation was the *security-positive* step (revoking a key that had been exposed in a chat transcript, with its verified replacement already active). Two gaps surfaced:

1. **No in-session override path.** The documented category toggle (`LOOM_GUARD_CLOUD`) is read from the *hook process* environment, so an agent prefixing `LOOM_GUARD_CLOUD=0` on the command string has no effect — the env tier is only reachable by editing the session/daemon environment and restarting, which no mid-task agent can do.
2. **Block, not ask.** The guard-hooks design has an ask tier, but this pattern lands in the block tier; the agent's only recourse was the script-file workaround (Write + `bash file`) — which works, but is undocumented for this class and generalizes to *every* blocked pattern, i.e. it's a bypass, not a policy.

This should NOT weaken protection against drive-by destructive cloud calls. The proposal is to move selected always-blocked cloud patterns (e.g. `aws iam delete-*`) to **ask** (operator confirms in the session) rather than silent block, and to document the intended escalation path for the rest.

- [ ] `aws iam delete-*` (and review peers like `gcloud iam … delete`) prompt via the ask tier instead of hard-blocking, at least in interactive sessions
- [ ] Guard docs state the sanctioned path when a legitimate operation is pattern-blocked (and explicitly address the script-file workaround's status)
- [ ] Decision-log records ask-tier approvals like other guard decisions
- [ ] No change to non-interactive/autonomous behavior without explicit design (a headless sweep should still be blocked, not prompted)

## Dependencies

- [x] #4178: Builder worktree-isolation escape (Bash-write bypass of the Edit/Write guard). This issue's file set — the live guard hooks themselves — is the same worst case for a repeat that led the operator to park #4063; fix in review as PR #4210. **Resolved 2026-07-28: #4178 closed COMPLETED, fix merged to main (`3d876ef4`).**

This issue cannot proceed until dependencies are complete. See the curator comment (2026-07-28) for verified tier inventory, implementation options, and test plan.

rjwalters:
Closes #4216

## What & why

During a 2026-07-28 credential rotation the cloud guard hard-BLOCKED
`aws iam delete-access-key` — the security-positive step (revoking an exposed
key whose verified replacement was already active). The pattern sat in the
catastrophic, ungated ALWAYS_BLOCK deny tier, so there was no in-session
recourse other than the undocumented script-file bypass.

This retiers selected cloud-destruction patterns from catastrophic deny to the
**ungated ask tier** (curator's Option A, the minimal set), splits the shared
lifecycle/cloud deny call site, updates the tier docs, and extends the guard
test suite. It does **not** weaken autonomous-mode protection.

## Per-pattern decisions (deliberate, mirroring #4063 discipline)

**Moved to the ungated ask tier** (always prompts an interactive operator, never
silently allowed, headless still blocks):
- `aws iam delete*` — moved out of the raw ALWAYS_BLOCK list into `ASK_PATTERNS`
  alongside `gh release delete`.
- `az ... delete` / `gcloud ... delete` — the segment parser still classifies
  them; the call site now splits lifecycle (deny) from cloud-delete (ask).

Kept **ungated** on purpose (NOT folded into the `guards.cloudCli` /
`CLOUD_ASK_PATTERNS` toggle): `guards.cloudCli:false` / `LOOM_GUARD_CLOUD=0`
would otherwise silently bypass IAM deletion entirely — an unacceptable
weakening for credential deletion (curator flag against Option B). A test pins
that `LOOM_GUARD_CLOUD=0` still ASKS on `aws iam delete-user`.

**Kept as catastrophic deny** (not part of the rotation incident):
`aws s3 rm --recursive`, `aws s3 rb`, `aws cloudformation delete-stack`,
`docker system prune`. Pinned with explicit deny tests.

**Call-site split**: system-lifecycle commands (`halt`, `reboot`, `poweroff`,
`shutdown`, `init 0|6`) remain HARD DENIES; only the cloud-delete branch moves
to ask. A lifecycle deny outranks a cloud-delete ask in a compound command, so
the hard floor is never downgraded.

## Headless / autonomous safety (no new code needed)

No interactivity detection was added. Under `--dangerously-skip-permissions`
(how every autonomous sweep runs) an ASK decision has no human to answer it and
therefore blocks — functionally a silent deny. So a deny to ask move preserves
autonomous-mode blocking automatically while giving interactive sessions a
prompt. This is documented at `defaults/docs/guard-hooks.md`, "Autonomous Guard
Defaults" section ("A headless sweep runs under `--dangerously-skip-permissions`
... an ASK decision has no human to answer it — so it blocks").

## Docs

- Corrected the cloud opt-out and reversible-gh tier prose so `aws iam delete*`
  and `az`/`gcloud delete` are listed in the ungated ask tier, not the
  catastrophic deny list.
- Added a "When a Legitimate Operation Is Pattern-Blocked" section: (1)
  interactive to the ask-tier prompt is the sanctioned path; (2) headless to
  re-run in a supervised interactive session by design; (3) the Write+`bash
  file` script workaround is explicitly UNSANCTIONED (a generic bypass that
  defeats every pattern and leaves no ask/deny record) — the correct fix for a
  recurring false positive is a pattern/tier-change PR like this one.

## Tests (all automated in tests/hooks/test-guard-destructive.sh)

- Retiered patterns emit `permissionDecision:"ask"` (raw `aws iam delete*` and
  the az/gcloud segment-parser split).
- Decision log records `{"decision":"ask","tier":"ask"}` for the retiered
  patterns.
- System-lifecycle commands (`halt`, `reboot`, `init 0`) still DENY after the
  split.
- Kept-catastrophic patterns (`s3 rm --recursive`, `s3 rb`,
  `cloudformation delete-stack`, `docker system prune`) still DENY.
- `LOOM_GUARD_CLOUD=0` still ASKS on `aws iam delete-user` (ungated).
- #3584 prose false-positive regression (hazard...delete near az/gcloud prose)
  still ALLOWS.

Full suite: 458/458 pass locally when run with a clean environment
(`env -u LOOM_FORCE_SCOPE -u LOOM_GUARD_DECISION_LOG`). Note: the autonomous
guard defaults `LOOM_FORCE_SCOPE=protected` and `LOOM_GUARD_DECISION_LOG=1`, if
exported in the runner's environment, leak into the force-scope and
decision-log-toggle tests and cause 12 unrelated failures — these are
pre-existing/environment-driven (confirmed identical against the unmodified
main guard) and untouched by this change.

## Manual verification of the interactive path

I cannot spawn a nested interactive session to literally prove the prompt
appears, so I verified the code path instead: running the guard on
`aws iam delete-user --user-name bob`, `aws iam delete-access-key ...`,
`az group delete ...`, and `gcloud compute instances delete ...` each produces a
JSON decision with `permissionDecision:"ask"` and a reason (the interactive
prompt Claude Code renders from that ASK). The headless-blocks behavior is not
newly implemented here; it is the documented meaning of an unanswered ASK under
`--dangerously-skip-permissions` (cited above).

## Lockstep

`.loom/hooks/guard-destructive-generic.sh` is kept byte-identical to
`defaults/hooks/guard-destructive-generic.sh` (verified with `cmp`; both were
byte-identical at HEAD and now carry the identical 58-line diff). Note:
`./.loom/scripts/resync-installed.sh --dry-run` resolves REPO_ROOT via
`git rev-parse --git-common-dir`, which from a worktree points at the primary
workspace rather than the worktree, so it reports the primary checkout as
unchanged (exit 0); the worktree lockstep was therefore verified directly with
`cmp` / `diff`.

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -1253,11 +1253,25 @@ ALWAYS_BLOCK_PATTERNS=(
     # scan. For a repo whose job is standing up and tearing down dev VMs the
     # teardown path (`terminate-instances`) is a first-class workflow, so it is
     # downgraded to an ask via the toggle-gated CLOUD_ASK_PATTERNS below (and
-    # fully bypassed when LOOM_GUARD_CLOUD=0 / guards.cloudCli:false). The other
-    # aws forms here stay ungated — they remain a hard safety floor (#3593).
+    # fully bypassed when LOOM_GUARD_CLOUD=0 / guards.cloudCli:false).
+    #
+    # NOTE: `aws iam delete` was likewise retiered OUT of this catastrophic
+    # scan — but to the UNGATED ask tier (ASK_PATTERNS below, alongside
+    # `gh release delete`), NOT to CLOUD_ASK_PATTERNS (#4216). Rationale:
+    # deleting an IAM credential is a legitimate, often security-POSITIVE step
+    # (revoking an exposed key whose replacement is already active) that a
+    # supervised operator must be able to run in-session — a hard block left
+    # only the undocumented script-file bypass. It stays UNGATED so a repo that
+    # set guards.cloudCli:false / LOOM_GUARD_CLOUD=0 for EC2-churn convenience
+    # would still ASK (never silently allow) on IAM deletion, and a headless
+    # sweep still effectively blocks (an ASK with no human to answer denies;
+    # see defaults/docs/guard-hooks.md). The remaining aws forms below stay
+    # ungated catastrophic denies — a hard safety floor (#3593): mass object /
+    # bucket deletion (`s3 rm --recursive`, `s3 rb`) and stack teardown
+    # (`cloudformation delete-stack`) were not part of the rotation incident and
+    # are deliberately kept as hard denies.
     'aws s3 rm.*--recursive'
     'aws s3 rb'
-    'aws iam delete'
     'aws cloudformation delete-stack'
 
     # Docker mass destruction
@@ -1411,9 +1425,22 @@ lifecycle_or_cloud_reason() {
     }'
 }
 
-_LIFECYCLE_REASON=$(lifecycle_or_cloud_reason "$COMMAND_NO_COMMENT" | head -1)
-if [[ -n "$_LIFECYCLE_REASON" ]]; then
-    deny "BLOCKED: $_LIFECYCLE_REASON" "lifecycle-or-cloud-delete"
+# Split the single deny call this used to share (#4216). System-lifecycle
+# commands (halt/reboot/poweroff/shutdown/init 0|6) are a hard safety floor and
+# stay DENY. The az/gcloud `… delete` cloud branch was retiered to the UNGATED
+# ask tier — mirroring the `aws iam delete` move above — so a supervised
+# operator is prompted rather than hard-blocked, while a headless sweep's
+# unanswered ASK still blocks (see defaults/docs/guard-hooks.md). A lifecycle
+# deny takes precedence over a cloud-delete ask in a compound command (e.g.
+# `az group delete …; halt`), so the hard floor is never downgraded.
+_LIFECYCLE_CLOUD_REASONS=$(lifecycle_or_cloud_reason "$COMMAND_NO_COMMENT")
+_LIFECYCLE_DENY=$(printf '%s\n' "$_LIFECYCLE_CLOUD_REASONS" | grep '^system lifecycle command:' | head -1)
+if [[ -n "$_LIFECYCLE_DENY" ]]; then
+    deny "BLOCKED: $_LIFECYCLE_DENY" "lifecycle"
+fi
+_CLOUD_DELETE_ASK=$(printf '%s\n' "$_LIFECYCLE_CLOUD_REASONS" | grep '^cloud resource deletion:' | head -1)
+if [[ -n "$_CLOUD_DELETE_ASK" ]]; then
+    ask "Command requires confirmation: $COMMAND ($_CLOUD_DELETE_ASK — retiered to the ungated ask tier in #4216; an interactive operator confirms, a headless session still blocks)" "cloud-delete-ask"
 fi
 
 # =============================================================================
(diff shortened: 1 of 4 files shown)


rjwalters:
## Judge verdict: APPROVED

Reviewed as the sole complete verdict (the prior Judge died mid-review after claiming; no comments were posted).

### What was verified

**Security-sensitive retier is sound (deny to ask, nothing loosened).**
- The IAM-credential-deletion pattern moved out of the raw `ALWAYS_BLOCK_PATTERNS` into the **ungated** `ASK_PATTERNS` (alongside `gh release delete`) — deliberately NOT into the toggle-gated `CLOUD_ASK_PATTERNS`. This is the right call: a repo running `guards.cloudCli:false` / `LOOM_GUARD_CLOUD=0` would otherwise silently bypass credential deletion. A test pins that with the cloud toggle off it still ASKS (never allows).
- The shared lifecycle/cloud deny call site was split cleanly: system-lifecycle commands (`halt`/`reboot`/`poweroff`/`shutdown`/`init 0|6`) stay hard DENY; only the az/gcloud cloud-deletion branch retiers to ASK. The two new greps (`^system lifecycle command:` / `^cloud resource deletion:`) match exactly the prefixes emitted by `lifecycle_or_cloud_reason`. In a compound like `... ; halt` the lifecycle deny still fires first, so the hard floor is never downgraded (confirmed by test).
- Kept-catastrophic denies preserved and pinned with explicit tests: recursive S3 object deletion, bucket removal, CloudFormation stack teardown, and the Docker mass-prune form.
- No previously-allowed command becomes blocked; no hard-dangerous command becomes silently allowed (an unanswered ASK under `--dangerously-skip-permissions` still blocks headless).

**Category toggles / defaults preserved.** The #4063 config-resolver truth-table tests (sqlDdl/cloudCli/reversibleGh/rmScope/forceScope/decisionLog default polarity) all still pass. Ungating is deliberate and documented.

**Canonical + lockstep.** Change made in the canonical `defaults/hooks/guard-destructive-generic.sh`; the `.loom/hooks/` copy carries the byte-identical diff (`cmp` clean on the PR branch, HEAD 4fff908d).

**Tests.** Ran the full guard suite in a clean env (`env -u LOOM_FORCE_SCOPE -u LOOM_GUARD_DECISION_LOG bash tests/hooks/test-guard-destructive.sh`): **458/458 pass**, including the new assertions — retiered patterns emit `permissionDecision:"ask"`, the decision log records `{decision:ask, tier:ask}`, lifecycle and kept-catastrophic patterns still DENY, and the #3584 prose false-positive still ALLOWS. (The 12 env-leak failures noted in the PR body reproduce only when the runner exports the autonomous defaults into the force-scope/decision-log tests — they do not occur in a clean env and are unrelated to this change; confirmed.)

### Acceptance criteria (#4216)
- [x] IAM credential deletion and az/gcloud resource deletion prompt via the ask tier instead of hard-blocking (interactive).
- [x] Docs state the sanctioned path when a legitimate op is pattern-blocked, and explicitly mark the script-file workaround UNSANCTIONED (new "When a Legitimate Operation Is Pattern-Blocked" section).
- [x] Decision log records ask-tier decisions (new test).
- [x] No change to autonomous behavior without explicit design — headless still blocks by construction (unanswered ASK denies), documented in the Autonomous Guard Defaults section.

### CI
Required checks green (Installer Integration Tests, CLAUDE.md Line Budget, Detect Changes); Rust/MCP jobs correctly skipped by path filters (only shell + docs changed). Merge state CLEAN / MERGEABLE.

Clean, well-reasoned, thoroughly tested change on a security-sensitive surface. Approving and moving to `loom:pr`.

Issue #4016, opened 2026-07-27
chore(daemon): pin a stable ad-hoc signing identifier + correct the daemon-reference TCC-durability claim (ad-hoc signing does NOT make grants survive rebuilds)


rjwalters:
## Context

Follow-up from #3980 (macOS TCC hygiene under launchd). That issue's audit found no reason `loom-daemon` currently *needs* any TCC-protected-folder grant — the daemon's legitimate working set (workspace roots, `.loom`, `.claude*`, `/private/tmp`) never touches Desktop/Documents/Downloads/Photos/Music/iCloud — so #3980 did not add any signing infrastructure.

However, `loom-daemon` is currently an ad-hoc/unsigned binary that `loom-daemon-update.sh` (#3968) rebuilds from source on every self-update roll. TCC identity for an unsigned binary is keyed to its cdhash, so **any** future grant (intentional or accidental) is silently orphaned by the next rebuild — producing either a recurring popup storm (if the operator re-grants each time) or a standing over-grant nobody remembers to revoke.

## Proposed work

Give `loom-daemon` a **stable ad-hoc code signature** (e.g. `codesign -s - --identifier com.rjwalters.loom-daemon <bin>`, applied as part of `cargo build --release` / the `loom-daemon-update.sh` provisioning path) so that:
- an identifier-based (not cdhash-based) TCC grant, if one is ever intentionally needed for a future capability, survives a rebuild that doesn't change the binary's identifier;
- the daemon's build/provisioning pipeline stays hermetic (ad-hoc signing needs no Apple Developer account or keychain-stored identity).

## Acceptance criteria

- [ ] `cargo build --release` (or `loom-daemon-update.sh`'s provisioning step) ad-hoc-signs the resulting `loom-daemon` binary with a stable identifier
- [ ] Verify with `codesign -dv --verbose=4 <bin>` that the identifier is stable across two consecutive rebuilds with no source changes
- [ ] `.loom/docs/daemon-reference.md`'s TCC hygiene section (added by #3980) is updated to reference the now-stable signing story
- [ ] No change to default TCC posture — FDA and other broad grants remain out of scope; this only makes a *future, intentional, narrow* grant durable

## Non-goals

- Does not request or configure Full Disk Access — see #3980 for why that's never the right fix.
- Does not add any new TCC-protected-folder access to the daemon today.

---

## Curator Enhancement (2026-07-27) — RESCOPED

> **⚠️ The original premise does not hold.** Curation empirically tested the proposed
[...]

PR #4284, opened 2026-07-28
feat(daemon): support LOOM_CODESIGN_IDENTITY for stable TCC-surviving codesigning

### LINKED ISSUE #4244: provision-daemon.sh: support a stable codesign identity (LOOM_CODESIGN_IDENTITY) so TCC grants survive rebuilds
## Problem

`provision-daemon.sh` ad-hoc signs the daemon binary (`codesign -f -s - --identifier com.rjwalters.loom-daemon`, in the signing helper around line 75). Ad-hoc signatures have no certificate chain, so macOS TCC pins any file-access grant to the exact binary hash (cdhash). Every `loom-daemon-update.sh` roll rebuilds the binary → new cdhash → macOS treats it as a brand-new app → **every previously granted permission is silently invalidated and the operator is re-prompted**. On this host we roll multiple times per day, so grants never stick. (Observed concretely: the Full Disk Access pane retains a `loom-daemon` row keyed to a long-gone ad-hoc build; its toggle no longer corresponds to the installed binary.)

The related-but-distinct annoyance: work spawned from a terminal shell (in-session `spawn-claude.sh`, nohup'd daemons, debug daemons run from worktrees) attributes its TCC requests to the *terminal app*, so the operator ends up granting broad file access to Terminal itself — which extends to everything ever run in a terminal.

## Proposed change

1. **`scripts/install/provision-daemon.sh`**: honor an optional `LOOM_CODESIGN_IDENTITY` (env, and/or a config key following the standard env/config/default precedence). When set and the identity exists in the keychain (`security find-identity -v -p codesigning`), sign with `codesign -f -s "$LOOM_CODESIGN_IDENTITY" --identifier com.rjwalters.loom-daemon`; otherwise fall back to the current ad-hoc path unchanged (non-Darwin hosts, CI, and hosts without the cert must keep working — the existing non-fatal degradation contract stays).
2. **Docs** (in `defaults/docs/`, which installs to consumer `.loom/docs/`): a short macOS-TCC note covering (a) the one-time self-signed cert setup (Certificate Assistant → Code Signing type, or the openssl + `security import` + `security add-trusted-cert -p codeSign` recipe), (b) why grants should be made to the daemon identity and not Terminal, and (c) the rule of thumb: dispatch sweeps via `loom-daemon dispatch`, not in-session `spawn-claude.sh`, so file-access attribution lands on the daemon.
3. The plist relaunch path needs no change — launchd already attributes children (dispatched sweeps included) to the daemon binary.

## Validated on this host (2026-07-28)

A self-signed cert `Loom Local Signing` was created, imported with `-T /usr/bin/codesign`, trusted for the codeSign policy (no GUI prompt needed), and the installed binary re-signed with it: `codesign -dvv` shows `Authority=Loom Local Signing`, adhoc flag cleared, binary executes normally. Two macOS quirks worth encoding in the docs/recipe:

- OpenSSL 3's default PKCS#12 export fails `security import` with "MAC verification failed" — the `-legacy` flag on `openssl pkcs12 -export` is required.
- `security import ... -T /usr/bin/codesign` is what makes later unattended signing (the #4055 self-update loop) non-interactive, provided the login keychain is unlocked in the user session.

Post-landing operator step (this host): remove the stale ad-hoc `loom-daemon` row from the Full Disk Access pane and re-add `~/.local/bin/loom-daemon` if a grant is actually needed — grants will then persist across rolls.

## Acceptance criteria

- [ ] With `LOOM_CODESIGN_IDENTITY` set to a valid identity, provision signs with it and `codesign -dvv` on the installed binary shows that Authority and no adhoc flag
- [ ] With it unset (or the identity missing), behavior is byte-identical to today's ad-hoc path, including the non-fatal warn on codesign failure
- [ ] Preserve-AC: non-Darwin hosts and hosts without `codesign` are untouched (early-return guard stays)
- [ ] Doc page added under `defaults/docs/` and linked from the daemon reference
- [ ] `test-loom-daemon-update.sh` / provision tests cover both branches (identity set vs unset)

rjwalters:
Closes #4244

## Summary

`provision-daemon.sh`'s existing ad-hoc signing (#4016) pins a legible
`--identifier` but has no certificate chain, so macOS TCC anchors any grant
to a per-build cdhash — every `loom-daemon-update.sh` self-update roll
rebuilds the binary, produces a new cdhash, and silently invalidates every
previously-granted permission.

This adds an **opt-in** `LOOM_CODESIGN_IDENTITY` (env, or a `codesign.identity`
config key — the repo's standard env > config > default precedence, same
shape as `spawn-worker.sh`'s runtime resolution):

- **Set + found in the keychain** (`security find-identity -v -p codesigning`):
  `sign_daemon_binary` signs with that certificate's chain
  (`codesign -f -s "$identity" --identifier com.rjwalters.loom-daemon`). A
  certificate-anchored designated requirement survives a rebuild, so a TCC
  grant made to the daemon binary persists across future rolls.
- **Unset, or an identity the keychain doesn't have**: falls back to the
  existing ad-hoc path (`codesign -f -s - --identifier com.rjwalters.loom-daemon`)
  **byte-identical to today**, including the non-fatal warn-on-failure
  contract.
- **Non-Darwin hosts / no `codesign` on PATH**: unaffected — the existing
  early-return guards are untouched and now sit ahead of the new identity
  resolution, so no extra work happens on those hosts either.

Also adds `defaults/docs/macos-tcc-codesign.md`: one-time self-signed
"Code Signing" certificate setup (Certificate Assistant GUI, or
openssl + `security import`), covering two macOS quirks confirmed on a real
host — OpenSSL 3's default PKCS12 export needs the `-legacy` flag for
`security import` to accept it, and `-T /usr/bin/codesign` at import time is
what makes later signing non-interactive (needed for the #4055 self-update
loop) — plus why TCC grants should target the `loom-daemon` identity rather
than the terminal app that happens to have spawned a sweep. Linked from
`daemon-reference.md`'s existing ad-hoc-signing section (which previously
called real signing "unevaluated speculative infrastructure" — updated to
describe this as the now-shipped opt-in path).

## Test plan

- [x] `bash tests/install/test-provision-daemon.sh` — 35/35 pass, including
      3 new cases: identity found → signed with that identity (not `-s -`),
      identity set but missing from the keychain → falls back to ad-hoc with
      a warning, and identity unset → ad-hoc path byte-identical to before.
- [x] `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 56/56 pass
      (unchanged; exercises the pre-existing ad-hoc-path regressions this
      change must not break).
- [x] `bash scripts/check-claude-md-budget.sh` — OK, CLAUDE.md untouched.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2525,11 +2525,18 @@ All this change actually buys is a **legible, stable identifier string**: the
 `codesign -dv` / System Settings → Privacy & Security / crash-diagnostic
 surfaces show `com.rjwalters.loom-daemon` instead of the rustc `-C metadata`
 hash (`loom_daemon-<hash>`, which itself changes on every version bump). It is
-also the necessary first step *if* the project ever adopts real signing — a
-self-signed certificate in a login keychain (breaks hermetic/headless builds)
-or Developer ID + notarization (requires a paid account) — but adopting
-either is unevaluated speculative infrastructure, not something this change
-does. No new TCC grant is requested or needed for this.
+also the necessary first step for real signing, which #4244 turns from
+speculative into an opt-in capability: set `LOOM_CODESIGN_IDENTITY` (or the
+`codesign.identity` config key, env > config > default) to a certificate
+already in the keychain and `sign_daemon_binary` signs with that certificate
+chain instead of ad-hoc — a certificate-anchored DR survives a rebuild, so a
+TCC grant to the daemon identity does too. Unset (or an identity the keychain
+doesn't have) falls back to the ad-hoc path above, unchanged, and no new TCC
+grant is requested by default. One-time self-signed cert setup (Certificate
+Assistant or openssl + `security import`, including the OpenSSL 3 PKCS12
+`-legacy` quirk and the `-T /usr/bin/codesign` trust-anchor requirement for
+unattended signing) and why grants should target the daemon identity, not
+Terminal: [`macos-tcc-codesign.md`](macos-tcc-codesign.md).
 
 ### Supervised restart primitive (#4054)
 
--- a/defaults/docs/macos-tcc-codesign.md
+++ b/defaults/docs/macos-tcc-codesign.md
@@ -1,0 +1,112 @@
+# macOS TCC + a stable codesign identity (#4244)
+
+Ad-hoc signing (the `loom-daemon` default, #4016) pins a legible
+`--identifier`, but TCC (Transparency, Consent, and Control — the Privacy &
+Security prompts) anchors a grant to the binary's **designated requirement**,
+which for an ad-hoc signature is a **cdhash-only** DR. Every
+`loom-daemon-update.sh` self-update roll rebuilds the binary (build.rs embeds
+the git commit and build time), producing a new cdhash — so any TCC grant
+made to the previous build silently evaporates and the operator is
+re-prompted. See [`daemon-reference.md` → "Why Full Disk Access is never the
+right answer"](daemon-reference.md) for the full measured writeup.
+
+`LOOM_CODESIGN_IDENTITY` (env, or the `codesign.identity` config key — the
+repo's standard env > config > default precedence) is the opt-in fix: point
+it at a certificate already in the keychain and `provision-daemon.sh` signs
+the daemon binary with THAT certificate's chain instead of ad-hoc. A
+certificate-anchored DR (`identifier "X" and certificate leaf = H"…"`)
+survives a rebuild — the identity, not the per-build hash, is what's pinned.
+Unset (or an identity `security find-identity -v -p codesigning` doesn't
+list) falls back to the ad-hoc path unchanged — this is entirely opt-in and
+every non-Darwin / no-`codesign` host is unaffected.
+
+## One-time setup: a self-signed "Code Signing" certificate
+
+You only need a certificate that satisfies the macOS `codeSign` policy — a
+paid Developer ID is not required for this local, single-machine use case.
+
+### Option A — Certificate Assistant (GUI)
+
+1. **Keychain Access → Certificate Assistant → Create a Certificate…**
+2. Name it something recognizable (e.g. `Loom Local Signing`).
+3. **Identity Type**: Self Signed Root. **Certificate Type**: **Code Signing**.
+4. Let it install into the login keychain.
+5. Verify it resolves: `security find-identity -v -p codesigning` should list it.
+
+### Option B — openssl + `security import` (scriptable)
+
+```bash
+# 1. Generate a self-signed cert + key.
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review - approve

Reviewed feature/issue-4244 at 9dcbd258 against the acceptance criteria in #4244.

**Correctness**
- `_pmd_resolve_codesign_identity` follows the repo's standard precedence order
  (env, then config, then default) -- matches spawn-worker.sh's LOOM_RUNTIME
  resolution shape -- soft-skipping to the next tier whenever LOOM_ROOT/git
  toplevel, config-resolver.sh, or jq are unavailable, consistent with
  loom_config_get's own soft-fail contract.
- `sign_daemon_binary` checks `security find-identity -v -p codesigning` for the
  resolved identity before attempting to sign with it; on success it signs with
  `-s "$identity"` (stable cert chain) instead of `-s -` (ad-hoc). On failure
  (either "not found" or codesign itself failing) it falls through to the exact
  same ad-hoc `codesign -f -s - --identifier com.rjwalters.loom-daemon "$bin"`
  call that existed before this PR, with the same non-fatal warn-on-failure
  contract.
- The pre-existing Darwin / codesign-on-PATH early-return guards are untouched
  and still run before identity resolution, so non-Darwin hosts and hosts
  without codesign do zero extra work.

**Security**
- `$identity` is quoted everywhere it's used (`grep -qF "$identity"`,
  `codesign -f -s "$identity" ...`), so no shell-injection risk from an identity
  string containing spaces or special characters.
- No secrets or credentials committed -- the new doc only describes local
  keychain setup steps.

**Docs**
- New defaults/docs/macos-tcc-codesign.md covers one-time cert setup (GUI +
  scriptable openssl path, including the OpenSSL 3 -legacy PKCS12 quirk and the
  -T /usr/bin/codesign trust-anchor need for unattended signing) and correctly
  recommends granting TCC to the loom-daemon identity rather than the terminal
  app.
- Linked from daemon-reference.md's existing ad-hoc-signing section, which is
  updated in place rather than duplicated.
- CLAUDE.md line-budget check passes locally (311/320 lines) -- no inline
  growth, detail lives in defaults/docs/.

**Tests** -- ran both suites locally against the PR head:
- tests/install/test-provision-daemon.sh -- 35/35 pass, including the 3 new
  cases (identity found -> signed with that identity, not -s -; identity set
  but missing from keychain -> falls back to ad-hoc with a warning; identity
  unset -> byte-identical ad-hoc path).
- defaults/scripts/tests/test-loom-daemon-update.sh -- 56/56 pass (unchanged
  regression coverage for the ad-hoc path).
- shellcheck scripts/install/provision-daemon.sh -- only a pre-existing SC2034
  on PROVISIONED_DAEMON_BIN (global-by-design, unrelated to this change).

**CI / merge state**
- All required checks green (gh pr checks 4284), mergeStateStatus: CLEAN.

Issue #4231, opened 2026-07-28
scheduler: load-aware dispatch — the aggressiveness dial (host melted at 6-way fan-out: WindowServer watchdog-killed, loom_daemon crashed)


rjwalters:
## The incident (2026-07-27 → 07-28, robb-studio, M-series 28-core)

Loom fanned out ~6 concurrent worktree release builds of vibesql. Combined with a Backblaze `completesync` and macOS Gatekeeper scan amplification (`syspolicyd` at ~520% CPU — see rjwalters/repo#50 for the host-side analysis), the box hit **load 118 on 28 cores**. Consequences, in order:

- `loom_daemon` crashed at 23:52 — **the orchestrator starved its own supervisor**.
- vibesql test binaries crashed at 00:49 / 00:51 (`.ips` reports on the host).
- WindowServer's main thread missed its 40-second watchdog checkin and was **killed 5 times in 5 minutes** (01:41–01:45), destroying the GUI session — VNC down, every interactive app gone, indistinguishable from a crash-boot.
- Recovery required a multi-hour human + remote-agent session (pause backups, triage Gatekeeper, clean worktrees). Overnight loom runs from that window are of unknown completeness.

A second fan-out wave re-spiked the machine at ~01:41 *after* an earlier intervention had brought load down from 79 to 8 — loom currently has **no feedback from host distress to dispatch**.

## The actual problem: the aggressiveness dial

This is the tortoise-and-the-hare trade-off, and it should be treated as a scheduling-policy design problem, not a bug:

- **Too conservative** → progress slope is lower than the hardware supports. Idle cores are wasted opportunity every hour of every day.
- **Too aggressive** → tail risk of multi-hour dropouts: crashed daemon, killed GUI, corrupted/incomplete runs, and (the expensive part) **human attention** to recover. One dropout can erase days of extra slope.

Maximizing *expected* progress means optimizing slope × availability, not instantaneous parallelism. The hare's instantaneous slope was spectacular right up until 01:41.

## Proposal: load-aware dispatch with a circuit breaker

1. **Concurrency cap** — hard limit on concurrent worktree builds, default derived from cores (e.g. total expected build threads ≤ ~70% of cores, leaving headroom for WindowServer/interactive/system daemons). Configurable per host.
2. **Load-feedback gate** — before launching a new build/worktree wave, check 1-min load and memory pressure; defer while load per core exceeds a threshold. Cheap to implement (sysctl/uptime), converts spikes into queueing instead of thrash.
[...]

PR #4287, opened 2026-07-28
feat(spawn): nice sweep/role-runner children so they no longer starve the host (#4233)

### LINKED ISSUE #4233: spawn: run sweep worker processes at background priority (nice/taskpolicy) so interactive and system processes win under contention
Decomposed from #4231 (host meltdown 2026-07-27→28: 6-way fan-out, load 118/28 cores, WindowServer watchdog-killed 5×, loom_daemon crashed). This is item 3 of that proposal — the lowest-effort, highest-value slice: starvation, not load per se, killed the GUI session.

## Problem

Sweep worker processes (and every build they spawn — cargo, rustc, test binaries) run at default priority (nice 0). Under fan-out they compete head-to-head with WindowServer, VNC, and system daemons. Nothing in the spawn path applies any scheduling policy: `defaults/scripts/spawn-claude.sh`, `spawn-worker.sh`, and the daemon's child-spawn path are all grep-clean for `nice`/`taskpolicy`/QoS.

## Prior art (important — read before designing)

The build gate went through the inverse experiment already:
- #3985 originally ran the *gate* at `nice -n 19`; #4020 withdrew the handicap to `nice 5` (`defaults/scripts/build-gate.sh:32-82`, env `LOOM_BUILD_GATE_NICENESS`, `LOOM_BUILD_GATE_NICE=0` disables).
- #4073/#4084: even at `nice 5` the gate still timed out under CPU contention, because sweep builds run at nice 0 and out-compete it.
- The comment at `build-gate.sh:63-70` explicitly documents the alternative — nicing the sweep children in the spawn path — as considered and deliberately not yet done. This issue does that.

Side effect worth verifying: once sweep children run below normal priority, the gate (nice 5) naturally wins the CPU race, addressing the #4084 contention premise structurally.

## Implementation sketch

- Apply priority at the outermost spawn point so it propagates to the whole process tree (agent → cargo → rustc): candidate sites are `defaults/scripts/spawn-claude.sh` / `spawn-worker.sh` (shell `nice -n N exec …` re-exec pattern, mirroring `build-gate.sh`'s `LOOM_BUILD_GATE_NICED` sentinel) and/or the daemon's Rust child-spawn.
- macOS: consider `taskpolicy` in addition to `nice`. Caution: `taskpolicy -b` (background/DARWIN_BG) also throttles disk I/O and network aggressively and can cause timeouts; `taskpolicy -c utility` is likely the right class. Make the mechanism configurable and default conservative (`nice` only, moderate value like 10).
- Config: `autonomous.*` knob + `LOOM_*` env override, following the env-over-config-over-default precedence used everywhere else (resolve via `loom-daemon/src/config_resolver.rs` if read in Rust).
- Must not apply to the build gate itself (it has its own policy) or to interactive/manual terminals — scope to daemon/sweep-dispatched workers.

## Affected Files

- `defaults/scripts/spawn-claude.sh` — apply niceness at spawn (installed copy `.loom/scripts/` syncs via resync)
- `defaults/scripts/spawn-worker.sh` — same, for the runtime-dispatch path
- `loom-daemon/src/` child-spawn site (locate the sweep process spawn) — if priority is applied Rust-side instead of shell-side, pick ONE layer, don't double-apply
- `defaults/docs/build-gate.md` — update the niceness-history section to note the inversion is now implemented

## Test Plan

- [ ] Spawn a sweep; verify worker and its cargo children show the expected nice value (`ps -o pid,ni,command`)
- [ ] `LOOM_*` override and config knob both respected; disable path returns to nice 0
- [ ] Build gate still completes under a concurrent niced sweep build (the #4084 scenario)
- [ ] macOS: if taskpolicy class used, verify no I/O-throttle-induced timeouts in a full sweep

## Dependencies

None — independently shippable.

rjwalters:
Closes #4233

## What

`spawn-claude.sh` — the outermost point in the daemon's actual dispatch path
for both sweep dispatch (`sweep_registry.rs`'s `resolve_spawn_bin`/`spawn_child`)
and scheduled role-runner dispatch (`role_runner.rs`) — now re-execs itself
once under a mild positive scheduling priority before spawning `claude`/
`claude-wrapper.sh` (and everything they in turn fork: cargo, rustc, test
binaries). The re-exec mirrors `build-gate.sh`'s existing
`LOOM_BUILD_GATE_NICED` sentinel pattern exactly, one layer further up the
tree, so nothing downstream needs separate handling — `nice`/`taskpolicy`
attributes survive `exec`.

- **Default**: `nice -n 10` (env `LOOM_SWEEP_NICENESS` > config
  `autonomous.spawnNiceness` > default `10`).
- **Master disable**: `LOOM_SWEEP_NICE=0` restores byte-identical pre-#4233
  behavior (nice 0, no taskpolicy).
- **Optional, off by default**: a macOS `taskpolicy -c <class>` QoS-class hint
  layers on top of nice (env `LOOM_SWEEP_TASKPOLICY_CLASS` > config
  `autonomous.spawnTaskpolicyClass` > unset). Deliberately never `-b`/
  `DARWIN_BG` — that also throttles disk I/O and network aggressively enough
  to risk causing the very timeouts this issue is trying to prevent (called
  out explicitly in the issue).
- **Scope**: applied only inside `spawn-claude.sh`, which is exclusively a
  headless/background dispatch entry point (daemon sweep dispatch, daemon
  role-runner dispatch, or ad hoc manual headless invocation) — MOM's
  interactive terminals run `claude` directly and never go through this
  script, so no separate "interactive opt-out" is needed. `spawn-worker.sh`
  only ever `exec`s into `spawn-claude.sh` (or a future runtime adapter),
  which preserves the pid, so the re-exec there still covers it transparently
  with no double-apply — documented in `spawn-worker.sh`'s own header. The
  build gate (`build-gate.sh`) has its own independent niceness policy and is
  never invoked through this script, so there's no overlap with it either.

## Why

Host meltdown #4231 (2026-07-27→28: 6-way sweep fan-out, load 118/28 cores,
WindowServer watchdog-killed 5×, `loom-daemon` crashed) diagnosed *starvation*,
not raw load, as the actual failure mode — sweep worker processes ran at
default niceness (0), competing head-to-head with WindowServer, VNC, and
system daemons for CPU under fan-out.

The build gate already ran the inverse experiment: #3985 originally niced the
*gate* itself to 19, #4020 walked that back to a milder `nice 5` after
#4044/#4046 found the original premise (CPU contention from sweeps) was a
`syspolicyd` exec-latency artifact, not real contention. But #4073/#4084
later showed the gate could still time out under genuine concurrent sweep
builds, because sweep children ran at nice 0 and out-competed the gate's
nice 5. `build-gate.sh`'s own comment (lines 63-70) flagged the alternative —
nicing sweep children *up* in the spawn path — as considered and deliberately
not done at the time.

This PR does that. With sweep children now at `nice 10` and the gate at
`nice 5`, the gate has a real, measured, unprivileged-achievable priority
edge over sweeps (the reverse of the previous `gate 5 > sweeps 0`
relationship) — so the gate should now structurally win CPU contention
against concurrently-dispatched sweep builds, addressing the #4084 premise
as a side effect rather than requiring dispatch-suppression alone.

## Testing

- `defaults/scripts/tests/test-spawn-claude.sh`: 119/119 pass, including 12
  new cases covering default niceness, `LOOM_SWEEP_NICE=0` disable, the
  `LOOM_SWEEP_NICENESS` env override, the `autonomous.spawnNiceness` /
  `autonomous.spawnTaskpolicyClass` config knobs (env > config > default
  precedence), taskpolicy invocation, and the re-exec-loop sentinel guard.
- `defaults/scripts/tests/test-spawn-worker.sh`: 26/26 pass (unaffected —
  doc-only change there).
- `cargo test --workspace --lib --bins`: 1489/1490 pass; the one failure
  (`terminal::claude_config::tests::test_setup_creates_fallback_state_when_no_home_state`)
  is a pre-existing order-dependent flake unrelated to this change — it
  passes cleanly in isolation.
- `shellcheck -x defaults/scripts/spawn-claude.sh`: clean.
- `defaults/docs/build-gate.md`'s niceness-history section updated to record
  the inversion as now implemented.

`scripts/test-installer.sh` (the bash installer suite the build gate also
runs) was still in flight at authoring time under heavy host contention from
several other concurrent sweeps on this machine — an unrelated, pre-existing
`syspolicyd`-style scheduling artifact (0% CPU stalls under parallel builds).
The gate will re-run it in CI.

### DIFF
--- a/defaults/docs/build-gate.md
+++ b/defaults/docs/build-gate.md
@@ -218,9 +218,28 @@ real, non-zero, unprivileged-achievable gap in the achievable direction
 (gate `5` > sweeps `0`): the gate yields *slightly* under contention rather than
 starving the sweeps, without being drastically deprioritized as it was at 19.
 The alternative — niceing sweep *children* up in the spawn path to force a
-strict gate-below-sweep gap — is deliberately not done, since the issue directed
-the spawn path be left untouched and the contention it would brace against is
-the one the evidence above withdrew.
+strict gate-below-sweep gap — was deliberately not done here at the time, since
+the issue directed the spawn path be left untouched and the contention it would
+brace against is the one the evidence above withdrew.
+
+**Update (#4233): the inversion is now implemented, from the OTHER side.**
+Issue #4231 (a 2026-07-27→28 host meltdown under 6-way sweep fan-out) diagnosed
+*starvation*, not raw load, as the real failure — sweep worker processes (and
+every `cargo`/`rustc`/test binary they spawn) ran at default niceness (0),
+competing head-to-head with interactive/system processes. `spawn-claude.sh` now
+re-execs itself at a mild positive niceness (default `nice 10`, configurable via
+`LOOM_SWEEP_NICENESS` env / `autonomous.spawnNiceness` config, master disable
+`LOOM_SWEEP_NICE=0`) using the same re-exec-with-sentinel pattern documented
+above (`LOOM_SWEEP_NICED` mirrors `LOOM_BUILD_GATE_NICED`). An optional macOS
+`taskpolicy -c <class>` scheduling-class layer (`LOOM_SWEEP_TASKPOLICY_CLASS` /
+`autonomous.spawnTaskpolicyClass`, off by default) is available on top of nice.
+See `spawn-claude.sh`'s own header comment for the full precedence chain. With
+sweep children now at `nice 10` and the gate at `nice 5`, the gate has a real,
+measured, unprivileged-achievable *higher* priority than sweeps — the reverse of
+the `gate 5 > sweeps 0` relationship described above — so the gate now
+structurally wins CPU contention against concurrently-dispatched sweep builds,
+addressing the #4084 premise (below) as a side effect rather than requiring the
+dispatch-suppression workaround alone.
 
 The re-exec mechanism and its knobs are preserved: `LOOM_BUILD_GATE_NICENESS`
 overrides the value (e.g. `=0` for exact sweep parity, `=19` to restore the old
--- a/defaults/scripts/spawn-claude.sh
+++ b/defaults/scripts/spawn-claude.sh
@@ -67,6 +67,20 @@
 #                          the sweep escalation ladder's per-rung `@effort`
 #                          cannot be threaded here because the in-session Task
 #                          tool exposes no effort parameter (see sweep.md).
+#   LOOM_SWEEP_NICENESS    Scheduling niceness applied to this process (and
+#                          everything it execs/forks — claude, claude-wrapper,
+#                          cargo, rustc, test binaries) via a `nice -n N exec`
+#                          re-exec, mirroring build-gate.sh's niceness pattern
+#                          (issue #4233). Precedence: this env var ->
+#                          `autonomous.spawnNiceness` config -> default `10`.
+#   LOOM_SWEEP_NICE=0      Disables the ENTIRE priority mechanism (nice AND
+#                          taskpolicy below), restoring pre-#4233 behavior.
+#   LOOM_SWEEP_TASKPOLICY_CLASS  Optional macOS `taskpolicy -c <class>`
+#                          scheduling class applied in-place (e.g. "utility").
+#                          Precedence: this env var ->
+#                          `autonomous.spawnTaskpolicyClass` config -> unset
+#                          (off by default). No-op on non-macOS or when
+#                          `taskpolicy` is unavailable.
 
 set -euo pipefail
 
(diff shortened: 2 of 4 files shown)


rjwalters:
Approved!

Reviewed against #4233's acceptance criteria; verified the actual diff (not just the PR description) by checking out the linked worktree .loom/worktrees/issue-4233 at commit 177d7d02.

Correctness / scoping:
- The niceness re-exec block in defaults/scripts/spawn-claude.sh sits at the true outermost point before any child process (token selection, claude/claude-wrapper.sh) -- confirmed both SweepRegistryConfig::resolve_spawn_bin/spawn_child (loom-daemon/src/sweep_registry.rs) and role_runner.rs dispatch through this exact script, so daemon sweep dispatch and scheduled role-runner dispatch are both covered from one spawn point.
- LOOM_SWEEP_NICED sentinel correctly guards against re-exec loops (verified via test plus exec preserving PID/env across the self-re-exec).
- Manual/interactive terminals (.claude/commands, loom.md) never invoke spawn-claude.sh directly -- only the daemon's fork+exec path does -- so the "not interactive terminals" scoping requirement holds.
- build-gate.sh is referenced only in comments in spawn-claude.sh, never invoked through it -- no double-apply risk, confirmed by grep. spawn-worker.sh's change is doc-only (it already execs into spawn-claude.sh, preserving the PID, so the runner-level re-exec still covers it) -- correctly not duplicated.
- Niceness values check out numerically: gate stays at default nice 5 (build-gate.sh:76), sweep children now default to nice 10 -- gate genuinely outranks sweeps under contention (lower nice means higher priority), matching the PR's stated goal and reversing the prior gate-outranks-sweeps relationship in the intended direction.
- taskpolicy usage is opt-in/off-by-default, applied in-place (-p $$, no re-exec needed since it survives exec), and deliberately never uses -b/DARWIN_BG per the issue's documented pitfall -- confirmed in code and comments.
- Precedence chain (env then config then default) for both LOOM_SWEEP_NICENESS/autonomous.spawnNiceness and LOOM_SWEEP_TASKPOLICY_CLASS/autonomous.spawnTaskpolicyClass is implemented correctly and lazily (only reads config when the env var is unset), using the existing lib/config-resolver.sh (loom_config_get).
- LOOM_SWEEP_NICE=0 cleanly disables the entire mechanism (nice and taskpolicy), restoring byte-identical pre-#4233 behavior.

Tests: Ran defaults/scripts/tests/test-spawn-claude.sh locally -- 119/119 pass, including all 12 new #4233 cases (default nice 10, disable via LOOM_SWEEP_NICE=0, env override, config-knob honored, env-over-config precedence, taskpolicy invocation and logging, combined nice+taskpolicy, and the re-exec-loop sentinel guard). shellcheck -x on both changed scripts is clean.

CI: All checks pass (gh pr checks) -- Installer Integration Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes all green (Rust jobs correctly skip since no Rust files changed). mergeStateStatus is CLEAN.

Nice work tying this back to the #4231 host-meltdown postmortem and the build-gate's own niceness history -- the doc update in build-gate.md closes the loop cleanly.

PR #4289, opened 2026-07-28
fix(hooks): make Bash-write redirection detection quote-aware (#4245)

### LINKED ISSUE #4245: Worktree-write guard (#4178) false-positives on '>' inside quoted arguments (e.g. gh --body prose)
## Problem

The #4178 Bash-write confinement guard (merged today in PR #4210) false-positives on `>` characters **inside quoted arguments**. Observed minutes after the merge, in the main checkout with a Loom-managed worktree present (`issue-4063`, owned by a daemon sweep, not the blocked session):

```
gh issue create --title "..." --label "loom:triage" --body "... following the env > config > default precedence ..."
```

was blocked with:

> BLOCKED: Bash-tool write to '/Users/rwalters/GitHub/loom/config' resolves to the main repository checkout ... This is a worktree-isolation bypass via Bash redirection/tee/sed -i/cp/mv — do NOT retry the write through Bash.

The guard parsed the prose fragment `env > config > default` inside the double-quoted `--body` string as a shell redirection targeting a file named `config`. `gh issue create` writes nothing to the filesystem.

This is the same failure class as the destructive-guard's curl-pipe false positive (#4040): substring/token matching over the raw command string without respecting shell quoting.

## Two compounding factors worth checking

1. **Quoting-blind redirection detection** — the core bug. A `>` inside a single- or double-quoted word is data, not a redirection operator. A proper fix tokenizes with quote awareness (even a minimal state machine for `'`, `"`, `\` would eliminate this class) rather than regex-scanning the raw string.
2. **Session attribution of the worktree check** — the block message said "a Loom-managed worktree exists for this session", but the only live worktree (`issue-4063`) belongs to a daemon-dispatched sweep, not the operator session that got blocked. If the guard keys on repo-global worktree existence rather than the session's own builder worktree, every operator session is confined whenever any sweep is running. Verify and scope the check to the session that owns the worktree.

## Workaround (works today)

Write the body with the Write tool and pass `--body-file <path>` — the Bash command string then contains no `>` and passes the guard. Same pattern as the #4040 workaround.

## Acceptance criteria

- [ ] `gh issue create --body "a > b"` (and any command with `>` only inside quoted arguments) is not blocked in the main checkout
- [ ] Real redirections (`echo x > file`, `tee file`, `sed -i`) into the main checkout are still blocked when a session builder worktree exists — regression tests from PR #4210 all stay green
- [ ] The worktree-existence check is scoped to the acting session's own worktree, or the block message stops claiming session ownership it didn't verify
- [ ] Add both cases to the guard's test file (quoted `>` passes; bare `>` still blocks)

rjwalters:
Closes #4245

## Root cause

`guard-destructive-generic.sh`'s Bash-tool write-confinement check (#4178)
recognizes `>`/`>>` redirection in `extract_write_targets()` by splitting a
`qsplit()`-segmented command on whitespace and pattern-matching each token.
That whitespace split was **not quote-aware**: a `>` that is only prose DATA
inside a quoted flag value (e.g. `gh issue create --body "... following the
env > config > default precedence ..."`) could land in its own
whitespace-bounded "token" and be misread as a real redirection operator,
manufacturing a phantom write target and denying a command that writes
nothing to the filesystem.

The existing mitigation (`strip_literal_text()` redacting
`--body`/`-m`/`--title`/`--notes`/`--comment` values before the scan) narrows
most cases, but is not a structural fix — `extract_write_targets()` itself had
no notion of quoting for `>`.

## Fix

Added `mask_gt()`, a small quote-tracking awk state machine (single-quoted /
double-quoted / unquoted), that replaces any `>` found **inside** a quoted
span with a placeholder byte (SOH, 0x01 — a character that can never appear
in a real shell command and so can never itself be mis-split into a phantom
target). It returns the string byte-for-byte unchanged otherwise, so
whitespace-splitting the masked string yields IDENTICAL token boundaries to
the original — the masked tokens are used only to decide whether a token is a
*real* (unquoted) redirection operator, while the original (unmasked) tokens
are still used to extract the actual write target text.

`mask_gt()` deliberately does **not** model backslash-escaped quotes, matching
the existing simplification `qsplit()` and `strip_literal_text()` already use
elsewhere in this file. This isn't just for consistency: the text `mask_gt()`
receives has typically already been through `strip_literal_text()`'s own
escape-blind redaction, which can shift a quote's effective position (an
escaped `\"` inside a redacted `--body` value loses its backslash during
redaction). Layering a stricter, escape-aware scan on top of that
already-escape-blind text actually desynchronizes the two passes' quote
parity in the wrong direction (masking too little) — verified by hand-tracing
and testing both variants against an escaped-quote case. Matching `qsplit()`'s
exact toggle-on-every-quote-char behavior keeps the two passes' parity in
agreement.

Also (per the issue's third acceptance-criteria bullet, scoped to a
quick/safe wording fix rather than a full session-identity refactor): the
worktree-write-confinement and Edit/Write worktree-path deny messages no
longer claim "a Loom-managed worktree exists **for this session**" — that
claim was never actually verified (the check is repo-global: it fires whenever
ANY managed worktree exists anywhere in the repo, not necessarily one owned by
the acting session). The messages now say a worktree "exists elsewhere in
this repository" and note the check cannot verify session ownership. No
behavior change, wording only. A full session-scoped fix would need an ambient
session-identity signal that doesn't currently exist in the hook's input
(PreToolUse hooks receive no session/task ID) — that's a larger effort than
this issue's primary quoting bug and is left for separate follow-up if wanted.

## Tests

Added to `tests/hooks/test-guard-destructive.sh` (write-confinement section,
`#4245`):
- `gh issue create --body "a > b"` allows (the issue's own repro)
- multiple quoted `>` in one `--body` value (the exact reported phrase)
  allows
- quoted `>` inside a single-quoted value allows
- regression: a quoted `>` alongside a REAL unquoted `>` in the same command
  still denies (masking must narrow, never widen)
- regression: a bare unquoted `>` redirection still denies

All existing tests in `tests/hooks/test-guard-destructive.sh`,
`defaults/hooks/tests/test-guard-worktree-paths.sh` (20/20,
including the `.loom/` vs `defaults/` byte-identical sync check), and
`tests/hooks/test-guard-destructive-dispatcher.sh` (7/7) pass.

Updated `.loom/hooks/guard-destructive-generic.sh` and
`.loom/hooks/guard-worktree-paths.sh` (the tracked, installed copies of these
hooks in this self-hosting repo) to stay byte-identical to their
`defaults/hooks/` source, matching this repo's existing resync convention.

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -969,6 +969,81 @@ function qsplit(s,   out, n, i, c, j, qc, ci, inner, SQ, DQ) {
 }
 '
 
+# =============================================================================
+# QUOTE-AWARE REDIRECTION MASKING (#4245)
+#
+# extract_write_targets() (below) recognizes `>`/`>>` redirection by splitting
+# a qsplit()-segmented command on whitespace and pattern-matching each token —
+# but that whitespace split is NOT quote-aware, so a `>` that is DATA inside a
+# quoted argument (e.g. `gh issue create --body "... env > config > default
+# ..."`) can land in its own whitespace-bounded "token" and be misread as a
+# real redirection operator, manufacturing a phantom write target and denying
+# a command that writes nothing to the filesystem (#4245; same failure class
+# as the #3755 qsplit()/#3679 strip_literal_text() quoting fixes above).
+#
+# mask_gt() walks the string tracking quote state (single-quoted,
+# double-quoted, unquoted) and replaces every `>` found INSIDE a quoted span
+# with SOH (0x01, a character that can never appear in a shell command and so
+# can never itself be mis-split into a phantom target). It otherwise returns
+# the input UNCHANGED byte-for-byte (same length, same whitespace positions),
+# so a caller can split both the original and the masked string on whitespace
+# and get IDENTICAL token boundaries — the masked tokens are used only to
+# DECIDE whether a token is a real (unquoted) redirection operator; the
+# ORIGINAL tokens are still used to extract the actual target text.
+#
+# Deliberately does NOT model backslash-escaped quotes -- same simplification
+# qsplit() (above) and strip_literal_text() already accept for this file's
+# other quote-tracking scans, and for good reason beyond just consistency: the
+# input mask_gt() actually receives (COMMAND_ASK_SCAN, see extract_write_targets()
+# below) has typically already been through strip_literal_text()'s OWN
+# escape-blind redaction, which can shift a quote's effective position (e.g. an
+# escaped `\"` inside a redacted --body value loses its backslash, since the
+# redaction's own quote-matching stops at the first bare `"` it finds). Layering
+# a stricter, escape-AWARE scan on top of that already-escape-blind text would
+# only desynchronize the two passes' quote parity -- worse, in the wrong
+# direction (masking too little). Matching qsplit()'s exact toggle-on-every-
+# quote-char behavior keeps both passes' parity in agreement. Same accepted
+# risk direction as qsplit(): pathological unbalanced-quote input could in
+# theory shift parity and mis-mask a genuine unquoted `>`, but that is the same
+# best-effort risk this file already accepts for `;|&` segmentation -- never a
+# NEW risk introduced here. An unterminated quote (no matching close before
+# end-of-string) just runs to the end of the string in that quote state --
+# never crashes, never mis-indexes.
+# =============================================================================
+_MASKGT_AWK='
+function mask_gt(s,   out, n, i, c, mode, SQ, DQ, MASK) {
+    SQ = sprintf("%c", 39)    # single quote
+    DQ = sprintf("%c", 34)    # double quote
+    MASK = sprintf("%c", 1)   # SOH -- placeholder for a quoted ">" (never a real char)
+    out = ""
+    n = length(s)
+    i = 1
+    mode = 0   # 0 = unquoted, 1 = single-quoted, 2 = double-quoted
+    while (i <= n) {
+        c = substr(s, i, 1)
+        if (mode == 0) {
+            if (c == SQ) { mode = 1; out = out c; i++; continue }
+            if (c == DQ) { mode = 2; out = out c; i++; continue }
+            out = out c
+            i++
+            continue
+        }
+        if (mode == 1) {
+            # Single-quoted: only the matching quote ends the span.
+            if (c == SQ) { mode = 0; out = out c; i++; continue }
+            out = out (c == ">" ? MASK : c)
+            i++
+            continue
+        }
+        # mode == 2 (double-quoted): only the matching quote ends the span.
+        if (c == DQ) { mode = 0; out = out c; i++; continue }
+        out = out (c == ">" ? MASK : c)
+        i++
+    }
+    return out
+}
+'
+
 # Parse force-op segments out of a command, emitting one TAB-separated
 # "<cpath>\t<target>" line per genuine git force-push / hard-reset. Portable awk
 # only (mirrors extract_rm_targets / lifecycle_or_cloud_reason segment parsing):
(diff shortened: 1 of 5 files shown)


rjwalters:
✅ **Approved!**

## Review summary

Verified against the issue's four acceptance-criteria bullets by checking out the PR branch (`.loom/worktrees/issue-4245`, clean, up to date with `origin/feature/issue-4245`) and reviewing `defaults/hooks/guard-destructive-generic.sh` + `defaults/hooks/guard-worktree-paths.sh` + `tests/hooks/test-guard-destructive.sh`.

**1. `mask_gt()` quote-tracking fix is correct and minimal-risk.** It's a pure awk state machine (unquoted / single-quoted / double-quoted) that replaces a `>` inside a quoted span with an SOH placeholder — same length, same whitespace boundaries as the original, so `extract_write_targets()`'s existing whitespace-split tokenization is unaffected; the masked tokens are used only to decide whether a token is a real unquoted redirection operator, while the original unmasked tokens still supply the actual write-target text. The escape-blind simplification (matching `qsplit()`'s existing behavior) is well-justified in the code comment: layering an escape-aware scan on top of `strip_literal_text()`'s already-escape-blind redaction would desynchronize the two passes' quote parity in the wrong direction.

**2. Manually re-verified all the acceptance-criteria scenarios directly against the guard binary** (the full `tests/hooks/test-guard-destructive.sh` run was extremely slow in this environment due to unrelated heavy rustc/vibesql CPU contention on the host -- 7 concurrent rustc processes -- so I cross-checked the specific behaviors directly against the guard script):
   - `gh issue create --title "Test" --body "a greater-than b"` (literal greater-than inside a quoted --body value) -- allow, correct (the issue's own repro)
   - `gh issue create --body` with the exact reported "env greater-than config greater-than default precedence" phrase -- allow, correct
   - `echo` with a single-quoted argument containing a literal greater-than -- allow, correct
   - the same quoted greater-than alongside a REAL unquoted redirection into the main checkout (worktree present) -- deny, correct -- confirms masking only narrows, never widens
   - a bare unquoted redirection into the main checkout (worktree present) -- deny, correct -- #4210 regression protection intact
   - `tee` and `sed -i` writes into the main checkout -- deny, correct -- #4210 regression protection intact

**3. Test coverage matches the issue's requirement** -- `tests/hooks/test-guard-destructive.sh` adds 5 new assertions covering exactly the above (3 allow, 2 deny), correctly labeled `(#4245)`.

**4. Worktree-ownership wording fix is honest and appropriately scoped.** Both deny messages (`guard-destructive-generic.sh`'s worktree-write-confinement and `guard-worktree-paths.sh`'s Edit/Write path check) no longer claim a Loom-managed worktree exists "for this session" -- a claim the check never actually verified (it's repo-global, firing on ANY managed worktree anywhere in the repo). New wording says the worktree "exists elsewhere in this repository" and notes the check cannot verify session ownership (see #4245). This is a docs/wording-only change (confirmed via diff -- no behavior change) and satisfies the issue's third bullet without over-scoping into a full session-identity refactor, which the PR description correctly identifies as separate follow-up work (no ambient session ID is available to PreToolUse hooks today).

**Housekeeping verified:**
- `.loom/hooks/guard-destructive-generic.sh` and `.loom/hooks/guard-worktree-paths.sh` are byte-identical to their `defaults/hooks/` sources (`diff` confirms) -- the resync convention is respected.
- Pre-existing 2 test failures in a full local run of `test-guard-destructive.sh` ("Ask for git push --force (non-main)", "Ask for git reset --hard") are unrelated to this PR -- reproduced identically against main's copy of the guard script (pre-PR), unaffected by this diff (force-push/reset-hard logic is untouched here). Root cause looks like a sandboxed-environment quirk (no network access affecting a remote-branch lookup), not a code defect -- filing this separately as an unlabeled follow-up issue for the Architect to triage rather than blocking this PR on it.
- CI is green (Detect Changes, CLAUDE.md Line Budget, Installer Integration Tests all pass; Rust/MCP jobs correctly show "skipping" since this PR touches only shell hooks). Merge state: CLEAN.

Nice, surgical fix with a well-reasoned quoting model and good regression coverage. Approving.

Issue #4299, opened 2026-07-28
daemon: dispatch resolves the target workspace from the daemon's cwd, not the workspace registry


rjwalters:
## Context — Linux worker-host pilot (#3979 Phase 2, 2026-07-28)

On a fresh Ubuntu 24.04 worker with the machine-level layout (#3835: checkout at `~/.local/share/loom`, binary at `~/.local/bin/loom-daemon`) and one workspace registered (`~/GitHub/anvil`, priority 15), every dispatch of an anvil issue was rejected:

```
refusing to dispatch issue #758: workspace /home/ubuntu/.local/share/loom is missing
.claude/commands/loom/sweep.md — the /loom:sweep slash command is not installed there
(#4027 wedge-loop guard). Run `loom-daemon init /home/ubuntu/.local/share/loom` first.
```

Issue #758 is an **anvil** issue and anvil is the only registered workspace — but dispatch resolved the target workspace to the daemon's **working directory** (the loom checkout, which under local-mode installs legitimately has no `.claude/commands/`). The CLI's cwd made no difference (`loom-daemon dispatch 758` from inside `~/GitHub/anvil` produced the same rejection) — it is the daemon process's cwd acting as "default workspace".

Workaround that made the worker functional: restart the unit with `WorkingDirectory=/home/ubuntu/GitHub/anvil` (plus copying the token pool there, per the same cwd-coupling in #4292). That collapses the machine-level daemon back into a single-repo daemon — on a multi-workspace worker there is no correct cwd to pick.

## Expected

Dispatch (and the work-finder) resolve an issue's workspace from the **workspace registry** (`~/.loom/workspaces.json`): find the registered repo whose forge slug owns issue N (or take an explicit `--repo`/`--workspace` argument), run the wedge-loop guard against *that* workspace, and never against the daemon's own cwd. The macOS host masks this today only because its daemon happens to be started from a repo that has the skill files installed.

Sibling of #4292 (token pool is cwd-relative) — same root pattern: machine-level daemon state resolved through process cwd.

## Suggested acceptance criteria

- [ ] With daemon cwd pointing anywhere (e.g. `/`), `loom-daemon dispatch <N>` for an issue belonging to a registered workspace dispatches into that workspace.
- [ ] Wedge-loop guard (#4027) evaluates the resolved workspace, not the daemon cwd.
- [ ] Ambiguity (issue number exists in multiple registered repos) is an explicit error asking for `--workspace`, not a silent cwd fallback.

---

[...]

Issue #4235, opened 2026-07-28
daemon: circuit breaker on sustained host distress — stop launching, drain running work, cool down before resuming


rjwalters:
Decomposed from #4231 (host meltdown 2026-07-27→28: 6-way fan-out, load 118/28 cores, WindowServer watchdog-killed 5×, loom_daemon crashed). This is item 4 of that proposal — the adaptive layer: when the host shows sustained distress, stop launching new work, let running work drain, resume after a cool-down. The daemon's own crash at 23:52 was a clear trip signal ~2 hours before the GUI died; a second wave re-spiked the host at 01:41 because nothing remembered the first incident.

## Distinction from the admission gate (sibling #4234)

The headroom check in #4234 is a *point-in-time admission* decision. The breaker is *stateful*: it trips on sustained/severe signals, stays open across ticks, and releases on TTL/cool-down — so a transient load dip can't immediately re-admit a full wave (exactly what happened at 01:41).

## Trip signals (keep the first version minimal)

- **Core signal (portable, ship first):** load-per-core above a threshold sustained for N consecutive work-finder ticks. Source already exists: `read_loadavg_1m()` in `loom-daemon/src/cpu_headroom.rs:250-259`.
- **Optional/macOS extras (follow-on, document but don't block on):** fresh `.ips` crash reports for loom-spawned processes, `syspolicyd` CPU ratio (Gatekeeper scan amplification — see rjwalters/repo#50), daemon self-restart detection. Keep these behind separate small issues if they add real complexity.

## Existing machinery to mirror (do not invent new shapes)

- **Threshold/TTL/auto-release shape:** the insta-crash quarantine in `loom-daemon/src/sweep_registry.rs:592-670` (`DEFAULT_QUARANTINE_THRESHOLD = 3`, `DEFAULT_QUARANTINE_TTL_SECS = 3600`, env `LOOM_WORK_FINDER_QUARANTINE*`). The breaker is the same pattern applied host-wide instead of per-issue.
- **Dispatch-suppressor wiring:** `MainHealthState` halt flag + `suppressDispatchDuringGate` in `loom-daemon/src/main_health_gate.rs` (`:2320`, consumed by `work_finder.rs:1185,1393`). The breaker should be a second suppressor consulted at the same points, not a parallel mechanism.
- **Testability template:** `quarantine_reconciliation.rs`'s pure decide/plan split (unit-testable without a forge).

## Observability constraints

[...]

PR #4303, opened 2026-07-28
fix(ipc): recover poisoned sweep-registry lock and retry dropped status connections

### LINKED ISSUE #4279: loom-daemon status: intermittently returns empty output with exit 0 (silent IPC failure)
## Context

Observed on 2026-07-28 while the daemon (running as the machine-level autonomous daemon, managing 5 workspaces) was under concurrent sweep load, including a sweep that was itself running daemon-update tests (#4232's `test-loom-daemon-update.sh`). Several times over ~90 minutes, `loom-daemon status` returned **zero bytes on stdout with exit code 0**; the next invocation ~60s later succeeded normally with full output.

A monitoring loop polling `loom-daemon status` every 60s from a consumer repo (anvil) surfaced the pattern repeatedly — each empty response is indistinguishable from success to a script (`$? == 0`), so downstream tooling misreads "daemon has no registered workspaces / no sweeps" instead of "the status call failed."

## Repro

Best-effort (timing-dependent): run `loom-daemon status` in a 60s loop while the daemon is dispatching multiple concurrent sweeps; occasionally the call returns empty/exit 0. Correlated with (but not proven caused by) daemon socket contention and/or a concurrently-running `test-loom-daemon-update.sh` in a sweep worktree.

- [ ] `loom-daemon status` never exits 0 with empty output: an IPC connect/read failure or timeout produces a non-zero exit and a diagnostic on stderr.
- [ ] A short internal retry (e.g. one reconnect attempt) absorbs transient socket contention before failing.
- [ ] Regression test: kill/occupy the socket mid-call and assert non-zero exit + error message.

---

### Root-cause analysis (verified against current source + installed binary)

The observed symptom decomposes into **two distinct layers**, and the "exit 0" premise did not reproduce against the current build:

1. **Server side — the real intermittent-empty-response mechanism.** Each IPC connection is handled in a detached `tokio::spawn` (`loom-daemon/src/ipc.rs:714-749`); a **panic** inside the handler unwinds the task and drops the socket **before any response byte is written** — the `if let Err(e)` arm only catches returned errors, not panics. The panic source under concurrent sweep load: the sweep registry is a `std::sync::Mutex` (`ipc.rs:18`) which **poisons permanently** after any panic while locked, and `build_daemon_status` unwraps it with `.expect("Sweep registry mutex poisoned")` (`ipc.rs:1137`, reached from the `DaemonStatus` handler at `ipc.rs:829-848`). Once poisoned, **every subsequent `status` call** panics server-side and the client sees EOF. A secondary silent-drop path: `serde_json::to_string(&response)?` failure at `ipc.rs:845`.

2. **Client side — no retry, and failure output is stderr-only.** `query_daemon_status` (`loom-daemon/src/main.rs:1884-1915`) does exactly one connect + one round-trip. An EOF becomes `Err("daemon closed the connection without responding")`, and the human-mode error branch (`print_status_unreachable_human`, `main.rs:2570-2671`) writes **only to stderr** with a **nonzero** exit from `daemon_install_state::exit_code()` (1/3/4 — `daemon_install_state.rs:69-129`). Verified 2026-07-28 against the installed binary (`~/.local/bin/loom-daemon`, built 2026-07-28 09:51): connect failure → exit 1, empty stdout, stderr diagnostic. **Current code has no path to empty-stdout + exit 0.** The monitoring loop that observed exit 0 most likely masked the exit code (e.g. command substitution/pipeline) or ran a pre-#4069 binary; a consumer sees "empty output" whenever it captures stdout only. The Builder should treat the exit-0 claim as unconfirmed and enforce the invariant with a regression test rather than hunt for a client-side exit-0 bug.

- [ ] **Server:** a `DaemonStatus` request never ends with the connection closing before a response — the handler must either write the status response or write an explicit error frame. Concretely: eliminate the poisoning failure mode (recover from/clear poison, or migrate the sweep-registry lock to a non-poisoning lock such as `parking_lot::Mutex` or `tokio::sync::Mutex`, consistently across all lock sites in `ipc.rs`), and/or catch handler panics so the connection gets an error response and the daemon logs the cause.
- [ ] **Client:** one bounded reconnect retry in `query_daemon_status` absorbs a transient dropped connection before failing (do not retry on clean "socket absent").
- [ ] **Invariant (from original AC):** `loom-daemon status` never exits 0 with empty stdout — IPC connect/read failure or timeout produces a nonzero exit and a stderr diagnostic. Already true on the failure paths in current source; lock it in with a regression test so it cannot regress.
- [ ] Regression tests cover the EOF path ("daemon closed the connection without responding"): a fake server that accepts and immediately closes must yield nonzero exit + stderr message; existing tests (`unresponsive_socket_times_out`, `absent_socket_errors_fast`, `main.rs:~4695-4745`) cover timeouts/absence only, not EOF.

### Affected Files

- `loom-daemon/src/ipc.rs` - accept loop (`:714-749`), `DaemonStatus` handler (`:829-848`), sweep-registry mutex + `.expect()` unwraps (`:18`, `:1137`, and sibling lock sites at `:225-272`, `:330-361`, `:1025-1075`) — fix the silent connection-drop-on-panic
- `loom-daemon/src/main.rs` - `query_daemon_status` (`:1884-1915`) — add bounded reconnect retry; `handle_status_command` (`:2476-2530`) — unchanged semantics, covered by new tests; unit-test mod (`:~4549+`) — add EOF-path tests
- `loom-daemon/src/daemon_install_state.rs` - exit-code mapping (`:69-129`), reference only
- `loom-daemon/tests/common/mod.rs` - happy-path IPC test client (`:105-119`); extend or add a fixture for an accept-then-close fake server if the integration route is chosen

### Test Plan

- [ ] Automated: unit test in `main.rs` test mod — bind a Unix socket, accept and immediately drop the connection; assert `query_daemon_status` errors, and the `status` command path exits nonzero with empty stdout + stderr diagnostic
- [ ] Automated: retry test — fake server drops the first connection, answers the second; assert `status` succeeds (validates the single-reconnect AC)
- [ ] Automated: server-side test — force a panic/poison in the registry lock (or simulate via a test hook), assert a subsequent `DaemonStatus` request still receives a response or error frame instead of EOF
- [ ] Manual: `LOOM_SOCKET_PATH=/tmp/nope.sock loom-daemon status; echo $?` → nonzero, stderr diagnostic (verified working today; re-verify after change)
- [ ] Edge cases: `--json` mode on the failure path (must emit the unreachable-JSON on stdout per existing `print_status_unreachable_json`, still nonzero exit); daemon alive but slow (existing 5s timeouts still honored)
- [ ] `cd loom-daemon && cargo test` passes

### Scope guard

- Fix is confined to the daemon IPC layer + status client. The separate MCP-bridge hang (`daemon.ts` framing, #4043) is **out of scope**.
- Do not change exit-code semantics established by #4069 (`daemon_install_state`); only add tests around them.
- Note: sweep-registry lock sites are numerous — if migrating the lock type balloons past ~400 LOC, land the panic-catch + client retry + tests first and file the lock migration as a follow-up.

---

## Curator re-verification (2026-07-29, ~01:50 UTC, verified against `origin/main` @ `f016678e`)

**Labels retained** (`loom:issue` + `loom:curated`): the analysis, approach, and ACs above are all still correct — the detached handler task still catches only returned errors (not panics), and the poisoning `.expect("Sweep registry mutex poisoned")` unwraps survive. But PRs #4316 (host breaker: +59 `ipc.rs`, +116 `main.rs`) and #4322 (registry-based dispatch workspace: +328 `ipc.rs`, +88 `main.rs`) landed after the enhancement, shifting nearly every line anchor. **Use this table, not the inline anchors above** (and re-verify at build time — these files are under active churn):

| Body reference | Was | Now on `f016678e` |
|---|---|---|
| `ipc.rs` accept loop + detached `tokio::spawn` | `:714-749` | `:737-772` (spawn at `:749`; `if let Err(e)` at `:750` — still returned-errors-only) |
| `ipc.rs` `Arc<Mutex>` import | `:18` | `:19` |
| `ipc.rs` `DaemonStatus` handler | `:829-848` | `:852-874` (`if let Request::DaemonStatus` at `:852`) |
| `ipc.rs` `serde_json::to_string(&response)?` in that handler | `:845` | `:868` |
| `ipc.rs` `build_daemon_status` + its poisoning `.expect` | `:1137` | fn at `:1133`, `.expect` at `:1168` — the handler now reaches it via `build_daemon_status_with_drain` (~`:861`, drain-state wrapper from #4090) |
| `ipc.rs` sibling `.expect("Sweep registry mutex poisoned")` lock sites | `:225-272`, `:330-361`, `:1025-1075` | `:348`, `:371`, `:379`, `:1056`, `:1088`, `:1096`, `:1106`, `:1168` |
| `main.rs` `query_daemon_status` | `:1884-1915` | `:1972` |
| `main.rs` `handle_status_command` | `:2476-2530` | `:2604` |
| `main.rs` `print_status_unreachable_human` | `:2570-2671` | `:2698` |
| `main.rs` tests `unresponsive_socket_times_out` / `absent_socket_errors_fast` | `~4695-4745` | `:4936` / `:4965` |
| `daemon_install_state.rs` exit-code mapping | `:69-129` | `exit_code()` at `:121` |

**Two substantive notes from the diff review** (no AC change):

1. The `DaemonStatus` handler now begins with a `tokio::task::spawn_blocking(refresh_cpu_util_cache).await` pre-warm (#4031) *before* building the report — an additional await point inside the handler, but its join errors are explicitly non-fatal, so it adds no new silent-drop path. The panic/poison mechanism in Root cause 1 is unchanged.
2. `DaemonStatusReport` gained a `Box<HostBreakerStatus>` field (#4316) and the handler routes through `build_daemon_status_with_drain` (#4090) — the lock-migration option in the Server AC must cover these newer lock consumers too; the sibling-site list above is the current complete set of `"Sweep registry mutex poisoned"` unwraps in `ipc.rs`.

rjwalters:
## Summary

Fixes the intermittent silent-EOF failure of `loom-daemon status` under concurrent-sweep load (#4279). Each IPC connection is handled in a detached `tokio::spawn`, so a panic while building a `DaemonStatus` report unwinds the task and drops the socket with **zero bytes written** — the client reads a bare EOF that a stdout-capturing monitor misreads as an empty "no workspaces" status. The panic source was the sweep-registry `std::sync::Mutex`: one panic under the lock **poisoned it permanently**, and every subsequent `status` call re-panicked at `.expect("Sweep registry mutex poisoned")`, turning a single fault into a persistent outage until daemon restart.

The fix has three parts: eliminate the poisoning failure mode server-side, guard the status handler against any other panic, and add a bounded client reconnect for a transient drop.

## Changes

**Server (`loom-daemon/src/ipc.rs`):**
- Recover the poisoned registry guard via `PoisonError::into_inner` at all 17 sweep-registry lock sites, instead of `.expect(...)` crashing. A prior panic can no longer turn `status`/`cancel`/`dispatch` into a permanent server-side failure.
- Wrap the `DaemonStatus` report build in `std::panic::catch_unwind` (the build is synchronous, so it never spans an `.await`) so the handler **always** writes either the report or an explicit error frame — and logs the panic cause — rather than dropping the connection. New `describe_panic` helper.
- Updated the now-stale "poisoned mutex ⇒ crash" doc comments to describe recovery.

**Client (`loom-daemon/src/main.rs`):**
- `query_daemon_status` makes **one bounded reconnect retry**, but only on a transient post-connect EOF (`daemon closed the connection without responding`). A clean socket-absent connect failure and a slow-daemon round-trip timeout are **not** retried, preserving fast-fail and the single 5s budget. Refactored into `query_daemon_status_once` + a `StatusAttemptError` classification (`Connect` / `DroppedBeforeReply` / `Roundtrip`).

- [x] **Server: a `DaemonStatus` request never ends with the connection closing before a response** — poison recovery removes the documented panic mechanism, and the `catch_unwind` guard makes it unconditional (report or error frame, always). Verified by `test_build_daemon_status_recovers_from_poisoned_registry`.
- [x] **Client: one bounded reconnect retry absorbs a transient dropped connection; no retry on clean "socket absent"** — retry fires only on `DroppedBeforeReply`. Verified by `status_retry_succeeds_after_transient_drop` and `status_absent_socket_errors_fast`.
- [x] **Invariant: `status` never exits 0 with empty stdout** — an EOF still yields an `Err` that `handle_status_command` reports on stderr with a nonzero `daemon_install_state::exit_code()` (unchanged, #4069). Locked in by `status_eof_yields_error`.
- [x] **Regression tests cover the EOF path** — a fake server that accepts and immediately closes yields a nonzero-worthy `Err` with the "closed the connection without responding" diagnostic.

## Scope adherence

- Confined to the daemon IPC layer + status client. The MCP-bridge hang (`daemon.ts`, #4043) is untouched.
- `#4069` exit-code semantics unchanged — only new tests were added around them.
- Took the poison-recovery + panic-catch + client-retry route (well under the ~400 LOC lock-migration escape hatch); no `parking_lot`/`tokio::sync::Mutex` type migration was needed.

## Test Plan

- [x] `cargo check` — clean
- [x] `cargo clippy --all-targets` — clean
- [x] New server test: `test_build_daemon_status_recovers_from_poisoned_registry` passes (poisons the registry mutex, asserts a report is still built)
- [x] New client tests: `status_eof_yields_error`, `status_retry_succeeds_after_transient_drop`, `status_absent_socket_errors_fast` all pass
- [x] Existing `build_daemon_status` suite (6 tests) and dispatch client tests still pass
- [x] `cargo test` — 1514 pass; one pre-existing unrelated flake (`tokens_pool::bootstrap::tests::bootstrap_repo_only_writes_pool_and_index`, `left: 9 right: 2` = host token count leaking via a shared-tokens env var from a concurrent non-serial test) passes in isolation and is untouched by this diff

Closes #4279


rjwalters:
❌ **Changes Requested — CI Failing (Rust Unit Tests), and the failure is a real functional gap, not flakiness**

## Summary of what I verified

- Checked out the existing worktree (`.loom/worktrees/issue-4279`, branch `feature/issue-4279`), confirmed up to date with `origin/main` (no rebase needed, `mergeStateStatus: UNSTABLE` / `mergeable: MERGEABLE`).
- `cargo check --all-targets` — clean.
- All 17 `.expect("Sweep registry mutex poisoned")` sites in `ipc.rs` are now `unwrap_or_else(std::sync::PoisonError::into_inner)` (verified by grep — 18 recovery sites incl. one new one, only 1 remaining `expect` string match, which is doc-comment text + the test's own intentional-poisoning helper, not a missed production site).
- `describe_panic` + `catch_unwind` wrapping of `build_daemon_status_with_drain` in the `DaemonStatus` handler looks correct: the build is fully synchronous (no `.await` inside the closure), so `AssertUnwindSafe` is safe here, and every branch (`Ok`/`Err`) writes a `Response` — no path drops the connection silently. Good instinct on the `#[allow(clippy::expect_used)]` removals + doc-comment updates too.
- Ran the new tests locally (macOS): `test_build_daemon_status_recovers_from_poisoned_registry`, `status_eof_yields_error`, `status_retry_succeeds_after_transient_drop`, `status_absent_socket_errors_fast` — **all pass on macOS.**

## The blocker: CI's `Rust Unit Tests` job fails on Linux

`gh pr checks 4303` shows `Rust Unit Tests` → **fail** (3m7s). From the job log:

```
thread 'status_client_tests::status_eof_yields_error' panicked at loom-daemon/src/main.rs:5195:9:
error should name the dropped connection, got: Connection reset by peer (os error 104)

thread 'status_client_tests::status_retry_succeeds_after_transient_drop' panicked at loom-daemon/src/main.rs:5240:23:
retry should have absorbed the first-connection drop, got: Connection reset by peer (os error 104)
```

This isn't test flakiness — it's a genuine gap in `query_daemon_status_once`'s classification (`loom-daemon/src/main.rs`, the `roundtrip` closure + the `match tokio::time::timeout(...)` below it):

```rust
match lines.next_line().await? {
    None => Ok(None),               // clean EOF -> classified DroppedBeforeReply (retried)
    Some(line) => { ... }
}
...
Ok(Err(e)) => Err(StatusAttemptError::Roundtrip(e)),   // any io error -> NOT retried
Ok(Ok(None)) => Err(StatusAttemptError::DroppedBeforeReply(...)),  // only the clean-EOF case
```

On Linux, when a peer closes a Unix-domain-socket connection **while unread request bytes are still sitting in the kernel receive buffer** (exactly the scenario here: the client has already written+flushed the `DaemonStatus` request before the server-side task drops the socket), the kernel sends **RST**, not a clean FIN — so the client's `next_line().await` returns `Err(ConnectionReset)`, not `Ok(None)`. That `Err` falls into the `?` inside the `roundtrip` closure and gets classified as `StatusAttemptError::Roundtrip`, which is **explicitly documented and coded as non-retryable**. macOS's BSD socket stack apparently surfaces this as a clean EOF instead (which is why it passed locally for me), but the CI runner (Linux) hits the RST path — and Linux is also the realistic production daemon-host OS this fix is meant to protect.

**Net effect: on the actual target platform, the exact "server task panics/drops mid-response" scenario this PR sets out to fix will *not* trigger the one bounded reconnect retry**, and the client will surface a raw `Connection reset by peer (os error 104)` instead of the friendly "daemon closed the connection without responding" diagnostic — which is also what the PR's own tests assert should not happen (and which is why they fail on Linux CI).

### Suggested fix

Broaden the retry classification to treat a read-phase I/O error before any full response line is available as the same "dropped before reply" case as a clean EOF — e.g. inspect `io::ErrorKind` on the error the `roundtrip` future can return (via `anyhow::Error::downcast_ref::<std::io::Error>()` or by threading the `io::Error` through instead of `anyhow!` at that call site) and route `ConnectionReset`, `BrokenPipe`, and `UnexpectedEof` into `StatusAttemptError::DroppedBeforeReply` alongside the `Ok(None)` case. Malformed-JSON and explicit `Response::Error` should stay `Roundtrip` (non-retryable) as today.

This is a logic/behavior change to the retry classification, not a trivial fix, so I'm not patching it directly during review — routing back to Doctor.

## Scope check (otherwise clean)

- No touches to `daemon_install_state.rs` exit-code semantics — confirmed unchanged (only the two files in scope, `ipc.rs` + `main.rs`, per `git diff --stat`).
- No attempt at the out-of-scope MCP-bridge (`daemon.ts`) issue.
- `unwrap_or_else(PoisonError::into_inner)` at all 17 sites is a defensible availability-over-strict-consistency tradeoff for this registry (matches the AC's explicit ask to "recover from/clear poison"), and the PR is honest that a panic mid-mutation *could* theoretically leave partial state — that's an inherent property of the recover-vs-crash choice, not a defect introduced here, so not blocking.

## Verdict

Please fix the ECONNRESET-vs-clean-EOF classification gap (item above) so CI's `Rust Unit Tests` job passes on Linux, then this looks ready to re-review — the poison-recovery + `catch_unwind` server-side work is solid.

rjwalters:
## Doctor: Judge feedback addressed (Linux CI failure)

**Root cause confirmed** (matches Judge's diagnosis): in `query_daemon_status_once`, a clean pre-response EOF (`Ok(None)`) was classified `DroppedBeforeReply` (retryable), but any read/write I/O error routed through `?` became `Roundtrip` (non-retryable). On Linux, a peer that closes the socket while unread request bytes are still queued in its kernel receive buffer replies with **RST**, so the client's read surfaces `ConnectionReset` (os error 104) instead of the clean EOF macOS reports — which is why the two `status_client_tests` passed on macOS but panicked on Linux.

**Fix** (`loom-daemon/src/main.rs`): added `classify_roundtrip_error`, invoked from the `Ok(Err(e))` arm of the round-trip timeout match. It downcasts the `anyhow::Error` to `std::io::Error` and reclassifies `ConnectionReset` / `BrokenPipe` / `UnexpectedEof` as the retryable `DroppedBeforeReply`, reusing the same friendly `daemon closed the connection without responding` diagnostic so the operator message is platform-independent. Malformed-JSON responses and explicit `Response::Error` replies are **not** `io::Error`s, so `downcast_ref` misses them and they remain non-retryable `Roundtrip`. Connect-phase classification is unchanged. Enum + closure doc comments updated to describe the RST path.

**Verification** (in the worktree):
- `cargo check --all-targets` — clean
- `cargo clippy -p loom-daemon -p loom-api -- -D warnings …` (exact CI flags) — clean
- `cargo test --bin loom-daemon status_client_tests` — 3/3 pass (`status_eof_yields_error`, `status_retry_succeeds_after_transient_drop`, `status_absent_socket_errors_fast`)
- `cargo test --lib test_build_daemon_status_recovers_from_poisoned_registry` — pass
- Full `cargo test -p loom-daemon` — 1514 passed; the only failure is `disk_headroom::tests::test_worktree_root_free_gb_returns_a_value`, which shells out to `df -Pk` and fails only under this sandboxed macOS host (unrelated to this change; passes on CI Linux runners).

Existing test assertions were left unchanged — the fix makes Linux behave identically to macOS's clean-EOF path, so the friendly-diagnostic and retry-absorption assertions hold without weakening.

Commit `b878a740` pushed to `feature/issue-4279`.

rjwalters:
## Judge Re-Review: Changes Requested (branch update needed — not a code defect)

### Doctor fix (b878a74) verified correct
I reviewed the `classify_roundtrip_error` change in `loom-daemon/src/main.rs` and it resolves the original blocker exactly as requested:

1. **Selective reclassification**: the downcast targets `std::io::Error` with kind `ConnectionReset` / `BrokenPipe` / `UnexpectedEof` only. Malformed-JSON responses (`serde_json::Error`) and explicit `Response::Error` replies (`anyhow!`) are not `io::Error`s, so they remain non-retryable `Roundtrip`. Connect-phase classification is untouched.
2. **Platform-independent diagnostic**: the reclassified path emits the same "daemon closed the connection without responding" message as the clean-EOF path, keeping `status_eof_yields_error`'s assertion green.
3. **Local verification**: `cargo test --bin loom-daemon status_` in the worktree passes all 10 tests, including `status_eof_yields_error` and `status_retry_succeeds_after_transient_drop` (macOS).
4. Scope discipline is good: one file, 42 insertions, doc comments updated.

### Blocker: merge conflict with main — CI cannot run
`mergeStateStatus` is **DIRTY** (`mergeable: CONFLICTING`), and as a result **no checks have run on b878a74** — GitHub cannot build the `pull_request` merge ref, so the deciding Linux "Rust Unit Tests" signal does not exist yet. I polled for several minutes: no checks will appear until the conflict is resolved.

The conflict is in `loom-daemon/src/ipc.rs`, in the `DispatchSweep` handler arm. Since this branch forked, main landed:
- `bfd9867f` — host-distress circuit breaker for dispatch (#4316)
- `5425c4f0` — registry-based target-workspace resolution for dispatch (#4322)

Both rewrote the same lines where this branch converted `.lock().expect("Sweep registry mutex poisoned")` to `.lock().unwrap_or_else(std::sync::PoisonError::into_inner)`. It is a single conflict hunk.

### Requested action
1. Merge `origin/main` into `feature/issue-4279` (or rebase) in the worktree at `.loom/worktrees/issue-4279`.
2. Resolve the `DispatchSweep` hunk by **keeping main's circuit-breaker block and its `resolve_registry`/target resolution**, and applying this branch's poison-recovering lock (`unwrap_or_else(PoisonError::into_inner)`) to the resulting `target.lock()` call — i.e., both changes compose; neither should be dropped.
3. Push, then confirm the Linux "Rust Unit Tests" job passes.

Once merge state is CLEAN and Linux CI is green on the new head, this is ready to approve — no further code changes are requested.

rjwalters:
Merge conflict resolved per the Judge's recipe (merge commit `0d4ba24`).

**DispatchSweep hunk in `loom-daemon/src/ipc.rs`**: composed both sides — kept main's host-distress circuit breaker (#4316/#4235) and its registry-based `resolve_dispatch_registry` target resolution (#4322/#4299), and applied this branch's poison-recovering lock (`unwrap_or_else(std::sync::PoisonError::into_inner)`) to the resulting `target.lock()` call. Neither change was dropped. The composed handler now: (1) hard-blocks dispatch when the host circuit breaker is tripped unless `force: true`, (2) resolves the sweep-registry target via `resolve_dispatch_registry`, then (3) acquires the registry lock recovering from poison rather than panicking.

**Semantic merge gap fixed in `main.rs`**: main added `host_breaker: Option<Box<HostBreakerStatus>>` to `DaemonStatusReport`, so the status-client test's `sample_report()` fixture now sets `host_breaker: None`. No production behavior change.

Verification (macOS worktree):
- `cargo check --all-targets` — clean
- `cargo test --bin loom-daemon status_` — 10/10 pass (incl. `status_eof_yields_error`, `status_retry_succeeds_after_transient_drop`)
- `cargo test -p loom-daemon --lib ipc::` — 66/66 pass

`mergeStateStatus` is now CLEAN / MERGEABLE, so Linux CI can build the merge ref. No further code changes; the ECONNRESET classification fix (b878a74) is unchanged.

rjwalters:
## Judge Review — Pass 3: APPROVED

Three-pass review complete for the #4279 poison-recovery / status-client work.

**Pass 1** approved the core change set (poison-recovering `unwrap_or_else(PoisonError::into_inner)` locks, `catch_unwind` hardening, status-client EOF/retry tests); the sole blocker was a Linux-only ECONNRESET classification gap.

**Pass 2** verified the fix (b878a74, `classify_roundtrip_error`): ConnectionReset/BrokenPipe/UnexpectedEof are selectively reclassified as retryable dropped connections, while malformed-JSON and `Response::Error` stay non-retryable, reusing the friendly diagnostic. Blocker: branch was CONFLICTING with main, so Linux CI could not run.

**Pass 3 (this pass)** spot-checked the conflict-resolution merge `0d4ba24` and confirmed it is faithful:
- `loom-daemon/src/ipc.rs` (DispatchSweep): main's host-distress circuit breaker (#4235/#4316) and `resolve_dispatch_registry` target resolution (#4299/#4322) preserved verbatim; this branch's poison-recovering lock applied to the resulting `target.lock()`. Nothing dropped, nothing extra.
- `loom-daemon/src/main.rs`: branch test module intact, with the required `host_breaker: None` added to the `sample_report()` fixture for main's new `DaemonStatusReport` field.

**CI on head 0d4ba24 (all green)**:
- Rust Linting & Build Check: pass
- Rust Unit Tests (Linux): pass (3m56s)
- CLAUDE.md Line Budget: pass
- Detect Changes: pass
- Installer Integration Tests / MCP Server Build: skipped (not applicable)

Merge state: CLEAN. Promoting to `loom:pr` and clearing `loom:ci-failure`. Ready for Champion auto-merge.

rjwalters:
Refreshed the branch onto main after #4318 (which boxed `Response::DaemonStatus` + added `DaemonStatusReport.token_pool_dir`). Both sides composed — #4279's panic guard, retry classification, and status_client_tests preserved on top of the boxing/pool-dir anchoring. `cargo check --all-targets` clean; `status_` and `ipc::` tests green.

Issue #4354, opened 2026-07-29
loom-daemon status: surface watchdog protection state on the reachable path (#4331 AC4)


rjwalters:
## Context

Follow-up to #4331 (watchdog/marker inconsistency). That issue's ACs 1–3 —
marker-ownership contract (docs), startup marker healing, and the watchdog's
"daemon alive but marker absent" WARN — were delivered in PR that `Closes #4331`.
Its own **Scope note** flagged AC4 (status surfacing) as cleanly separable, so it
is split out here.

## Problem

`loom-daemon status` does not surface **watchdog protection state** on the
*reachable* path (daemon answering over IPC). Today `daemon_install_state::probe()`
runs only on the IPC-**un**reachable path (#4069); a healthy, reachable daemon
never reports whether it is actually protected. So an operator cannot tell, from
`status` alone, that a running daemon has:

- **no autonomy-desired marker** (crash protection disarmed — the #4331 state), or
- **no watchdog job/timer provisioned** (nothing scheduled to detect a future
  death).

They would only find out by reading `daemon-watchdog.log` or inspecting
`launchctl`/`systemctl` by hand.

Note: with #4331's startup healing in place, a *supervised* reachable daemon will
almost always have a marker (it self-heals at boot), so the higher-value signal
here is usually **"watchdog job not provisioned"** — but both should be reported.

## Acceptance criteria

- [ ] `loom-daemon status` (human + `--json`) prints a **protection** line/field:
      `protected` / `unprotected — no autonomy-desired marker` /
      `watchdog job not provisioned`.
- [ ] Runs on the **reachable** path (not only the `Err` arm) — likely a small
      refactor so `daemon_install_state` (or a sibling probe) can express
      "alive-and-expected, protection = X", plus a client-side probe for whether
      the watchdog launchd job / systemd timer is provisioned
      (`launchctl print <label>-watchdog` / `systemctl --user is-enabled
      <unit>-watchdog.timer`).
- [ ] Respects `LOOM_AUTONOMY_MARKER` / `LOOM_WATCHDOG_LABEL` overrides, mirroring
      the watchdog's own resolution.
- [ ] Read-only, never fails the command (degrade to "unknown"), and does not
      regress the existing unreachable-path `install_state` output (#4069).
- [ ] Rust tests for the new classification + `--json` field.

## Affected files (starting points)

- `loom-daemon/src/main.rs` (~`handle_status_command` / `print_status_human` /
  `print_status_json`)
[...]

PR #4321, opened 2026-07-28
feat(cli): systemd timer unit for the loom-daemon autonomy watchdog on Linux

rjwalters:
## Summary

- `provision_watchdog_job()` in `loom-daemon-start.sh` gains a systemd branch: on a systemd Linux host it renders + installs a `Type=oneshot` `<unit>-watchdog.service` (runs the existing platform-agnostic `loom-daemon-watchdog.sh` payload) plus a paired `<unit>-watchdog.timer` (`OnUnitActiveSec`/`OnBootSec` = `LOOM_WATCHDOG_INTERVAL_SECS`, default 300s, `Persistent=false`), then `daemon-reload` + `enable --now` the **timer** — the systemd analog of the Darwin `StartInterval` LaunchAgent job.
- `loom-daemon-stop.sh`'s `teardown_autonomy_intent()` gains the symmetric systemd path: on an operator-initiated stop it `systemctl --user disable --now`s both the watchdog timer and service, alongside the existing launchd bootout.
- Both directions keep the existing best-effort / NON-FATAL contract: a watchdog install/teardown failure only warns, never fails the daemon start/stop.
- `defaults/docs/daemon-reference.md`'s watchdog section is updated: the "Linux… can be run by hand or wired to a cron / systemd timer" caveat is replaced with the now-provisioned timer, plus the new `LOOM_WATCHDOG_LABEL`/`LOOM_WATCHDOG_INTERVAL_SECS` env docs (also added to `loom-daemon-start.sh`'s own `--help` banner).
- **Structural note**: since a Darwin CI runner exercising the systemd branch via the `LOOM_SYSTEMD_FORCE=1` test seam can have `IS_DARWIN` and `IS_LINUX_SYSTEMD` both true at once, the three daemon-start call sites now call the specific `provision_watchdog_job_{launchd,systemd,none}` function they already know they need, rather than routing through a platform-autodetecting dispatcher that would always pick launchd in that scenario.
- Out of scope (per issue): no changes to `loom-daemon-watchdog.sh` itself (already platform-agnostic — it already resolves liveness from the PID file on any non-launchd path, which the rendered systemd service also forces via `Environment=LOOM_DAEMON_LAUNCHD=0`), the Darwin watchdog job, or heartbeat semantics.

## Changes

[...]

PR #4355, opened 2026-07-29
feat(daemon): heal absent autonomy-desired marker at startup + watchdog state-mismatch WARN

### LINKED ISSUE #4331: watchdog/marker inconsistency: daemon runs and dispatches while autonomy-desired is absent — watchdog logs OK and provides no crash protection
## Context

Observed 2026-07-29 (03:2x UTC): the daemon is running under launchd (`com.rjwalters.loom-daemon`), actively dispatching sweeps across five managed repos — while `~/.loom/autonomy-desired` is **absent**. Consequences:

1. **No crash protection.** The watchdog's contract is marker-driven; every tick it logs `[OK] no autonomy-desired marker at /Users/rwalters/.loom/autonomy-desired — no daemon expected; nothing to check.` If the daemon dies overnight (beyond launchd's own KeepAlive semantics), the watchdog will not revive it — during exactly the long unattended windows the watchdog exists for.
2. **Semantic inconsistency.** "Marker absent" is supposed to mean "no daemon expected," but a daemon is very much running and expected. Either the marker is authoritative (then something must create/preserve it across restarts and updates) or it isn't (then the watchdog's OK-verdict is misleading). Today neither holds: `loom-daemon restart` and `loom-daemon-update.sh` neither create nor check the marker.
3. **Watchdog liveness question.** The watchdog log's last entry was 23:33:42Z, ~4h before the observation, despite a 5-minute cadence — so the watchdog itself may also have stopped. (Secondary; diagnose while in there.)

- [ ] Decide and document marker ownership: which surfaces create it (start wrapper), preserve it (restart primitive, update script), and remove it (explicit stop).
- [ ] `loom-daemon restart` and `loom-daemon-update.sh` preserve an existing marker across the roll (and arguably create it when restarting a launchd-supervised daemon — restarting implies autonomy is desired).
- [ ] The watchdog treats "daemon running but marker absent" as a loud WARN (state mismatch), not an `[OK] nothing to check`.
- [ ] `loom-daemon status` surfaces watchdog protection state (protected / unprotected-no-marker / watchdog-not-running) so the operator sees the gap without reading watchdog logs.

## Related

- #4232 (restart primitive relaunch verification), #4321 (Linux systemd watchdog timer — same contract should hold there), #4235 (host-distress circuit breaker)

---

### Verified current behavior (code audit against `origin/main`)

The marker lifecycle is more implemented than the original report suggests — the gap is narrower and more specific:

| Surface | Marker behavior today | Reference |
|---|---|---|
| `loom-daemon-start.sh` | **Creates** the marker (`write_intent_marker`) on every successful start (launchd, systemd, and nohup paths) | `defaults/scripts/cli/loom-daemon-start.sh:424-457`, invoked at `:1028`, `:1101`, `:1139` |
| `loom-daemon-stop.sh` | **Removes** marker + tears down the watchdog job — unless `--restarting` / `LOOM_DAEMON_STOP_KEEP_INTENT=1`, which preserves both | `defaults/scripts/cli/loom-daemon-stop.sh:54-75`, `:177-188`, `:411-418` |
| `loom-daemon-update.sh` (full roll) | **Preserves and re-creates**: calls `stop --restarting` (preserve) then `start.sh` (re-write). ⚠️ **Correction to the original report** — this path is already correct. | `defaults/scripts/cli/loom-daemon-update.sh:1006-1025` |
| `loom-daemon restart` (Rust supervised restart primitive, #4054) | **Never touches the marker.** Daemon exits `EXIT_RESTART`; launchd relaunches. A present marker survives — but an **absent marker is never created**, so an unprotected daemon stays unprotected across every supervised restart, forever. | `loom-daemon/src/ipc.rs:924-950` |
| In-daemon self-update loop | Uses `loom-daemon-update.sh --no-restart` + the restart primitive — **bypasses** the stop/start marker rewrite entirely. | `loom-daemon/src/main.rs:~1412` |
| Watchdog no-marker fast path | Exits `[OK] … nothing to check` **without probing reality at all** — even when a daemon is demonstrably alive under the expected label. | `defaults/scripts/cli/loom-daemon-watchdog.sh:123-129` |
| `loom-daemon status` (daemon reachable) | **Never checks the marker.** `daemon_install_state::probe()` runs only on the IPC-unreachable path (#4069); there is no `InstallState` variant for "alive but not expected". | `loom-daemon/src/main.rs:2604-2660`, `loom-daemon/src/daemon_install_state.rs:93-105` |

**Net effect**: once the marker is absent while a launchd-supervised daemon runs (however that state arose), no code path ever heals it — the daemon can only ever be restarted via the primitive or self-update, neither of which creates the marker, and neither the watchdog nor `status` ever reports the mismatch.

### Root-cause hypotheses for the observed state (verify during implementation)

- `stop.sh` boots the launchd job out *and* tears the watchdog down when clearing the marker — so "watchdog job provisioned + marker absent + daemon running" most plausibly means the marker was deleted out-of-band, its write failed, or the daemon predates a marker-writing start and has only been rolled via self-update/restart-primitive since. Check marker dir mtimes and `launchctl print` job uptimes.
- Watchdog log staleness (last tick 23:33:42Z): probe the watchdog job itself — `launchctl print gui/$(id -u)/com.rjwalters.loom-daemon-watchdog`. A `StartInterval` job structurally can't crash-and-stay-dead, so a silent 4h gap suggests the job was booted out or its label/domain changed.

### Implementation guidance

1. **AC 1 (ownership doc)**: add a "Marker ownership" table (create / preserve / remove per surface — the table above is a starting point) to `defaults/docs/daemon-reference.md`. This is the contract the other three ACs implement.
2. **AC 2 (restart/update create-or-preserve)**: recommend healing at **daemon startup** rather than patching each restart path: when the daemon detects it is supervised (it already resolves supervision for the restart ack in `ipc.rs`) and the marker is absent, re-write it. One choke point covers the restart primitive, self-update, *and* bare launchd `KeepAlive` relaunches. Constraints: respect `LOOM_AUTONOMY_MARKER` (the launchd plist exports it — `loom-daemon-start.sh:504`); do **not** create the marker for an unsupervised foreground/debug run (that would make every dev run arm the pager).
3. **AC 3 (watchdog WARN)**: in the no-marker branch, cheaply probe reality with the existing label/domain resolution; if a daemon **is** alive → report a `WARN` ("state mismatch: daemon running but no autonomy-desired marker — crash protection disarmed") and exit `1`. The intentional-stop quiet case is preserved because `stop.sh` also removes the daemon job, so nothing is alive to find.
4. **AC 4 (status surfacing)**: run the install-state probe on the *reachable* path too and print a protection line (`protected` / `unprotected — no autonomy-desired marker` / `watchdog job not provisioned`), plus a field in `--json`. Likely needs a new probe outcome or a small refactor so `daemon_install_state` can express "alive but not expected".

Cross-platform note: #4321 (merged) added the systemd timer analog on Linux — AC 3's WARN branch lives in the shared `loom-daemon-watchdog.sh`, so it applies to both platforms; verify the systemd path in tests or file an explicit follow-up.

### Affected Files

- `defaults/scripts/cli/loom-daemon-watchdog.sh` — WARN branch in the no-marker case (AC 3)
- `defaults/scripts/tests/test-loom-daemon-watchdog.sh` — new cases: daemon-alive-no-marker → WARN/exit 1; no-daemon-no-marker → still silent OK
- `loom-daemon/src/main.rs` — startup marker-healing wiring (AC 2) + reachable-path status surfacing (AC 4)
- `loom-daemon/src/daemon_install_state.rs` — express "alive but not expected" / protection state for `status` (AC 4)
- `loom-daemon/src/ipc.rs` — only if healing lands in the restart primitive instead of startup (AC 2 alternative)
- `defaults/scripts/cli/loom-daemon-update.sh` — close/document the `--no-restart` marker gap (AC 2)
- `defaults/docs/daemon-reference.md` — marker-ownership table (AC 1)

### Test Plan

- [ ] Shell tests (`test-loom-daemon-watchdog.sh`): daemon alive + marker absent → WARN + exit 1; nothing alive + marker absent → silent OK (the load-bearing quiet case must not regress); marker present paths unchanged.
- [ ] Rust tests: install-state/protection classification for "alive but not expected"; `status --json` includes the protection field; startup healing writes the marker only when supervised.
- [ ] Manual verification: reproduce the observed state (delete the marker while the daemon runs), then confirm (a) watchdog tick reports WARN, (b) `loom-daemon status` shows unprotected, (c) `loom-daemon restart` (or daemon startup) heals the marker.
- [ ] Edge cases: intentional `stop.sh` still produces zero watchdog noise; unsupervised foreground run never creates a marker; `LOOM_AUTONOMY_MARKER` override respected end-to-end; Linux/systemd (#4321) parity.

### Scope note

Four ACs across ~7 files (shell + Rust) — near the single-PR ceiling. If it grows during implementation, AC 4 (status surfacing) is cleanly separable as a follow-up PR; ACs 1–3 form the coherent core (contract + healing + detection).

### Related

- #4232 (**open**) — restart primitive relaunch verification; same subsystem, independent work, no dependency either way.
- #4321 (merged) — Linux systemd watchdog timer; AC 3 must hold there too.
- #4011 / #4054 / #4235 (closed) — origin of the marker contract, the restart primitive, and the host-distress breaker respectively.

rjwalters:
## Summary

The autonomy-desired marker (#4011) is the durable "a daemon is EXPECTED on this host" signal that both the host-side watchdog (`loom-daemon-watchdog.sh`) and `loom-daemon status` (`daemon_install_state.rs`) key off. The gap #4331 closes: **no code path re-created an ABSENT marker while a supervised daemon kept running**. `loom-daemon restart` (#4054), the in-daemon self-update loop, and a bare launchd/systemd relaunch all bring up a fresh daemon *without* re-running the start script's `write_intent_marker` — so once the marker went absent (out-of-band delete, failed write, or a daemon only ever rolled via the primitive/self-update), it stayed absent forever. The daemon ran with crash protection silently disarmed, and every watchdog tick logged a misleading `[OK] … nothing to check`.

This PR delivers the coherent core (ACs 1–3): **contract + healing + detection**. AC4 (surfacing protection state in `loom-daemon status` on the reachable path) is split to **#4354** per the issue's own Scope note.

## Changes

- **AC2 — startup healing** (`loom-daemon/src/autonomy_marker.rs`, new; wired in `main.rs` right after the heartbeat loop): one startup choke point that covers the restart primitive, the self-update loop, AND a bare supervisor relaunch. If `ipc::detect_supervisor()` is `Some` **and** the marker is absent, re-write it. Deliberately **never** writes for an unsupervised (`--foreground`/nohup/debug) run; respects `LOOM_AUTONOMY_MARKER`; never overwrites a present marker; never fatal (logged on write failure). The healed marker mirrors `write_intent_marker`'s fields byte-for-byte (with a `# HEALED …` provenance comment) so the watchdog and `daemon_install_state` parse it identically.
- **AC3 — watchdog WARN** (`defaults/scripts/cli/loom-daemon-watchdog.sh`): the no-marker branch now runs the *same* liveness probe as the marker-present path (factored into a shared `detect_daemon_liveness` function). A daemon demonstrably alive while the marker is absent is a loud **WARN + exit 1** (state mismatch: crash protection disarmed). The load-bearing quiet case — marker absent **and** nothing alive (a deliberate stop) — stays exactly as silent as before.
- **AC1 — docs** (`defaults/docs/daemon-reference.md`): marker-ownership table (create / preserve / remove per surface), the startup-healing contract, and a new decision-matrix row for the marker-absent-but-alive WARN.

| Criterion (from #4331) | Status | Verification |
|---|---|---|
| AC1: document marker ownership (create/preserve/remove) | ✅ | New "Marker ownership" table + "Startup marker healing" subsection in `daemon-reference.md` |
| AC2: heal an absent marker at startup when supervised; respect `LOOM_AUTONOMY_MARKER`; never for an unsupervised run | ✅ | `autonomy_marker::heal_on_startup` wired in `main.rs`; unit tests cover supervised-heals / unsupervised-skip / present-untouched / parseable-by-install-state / owner-only perms |
| AC3: watchdog treats "daemon running but marker absent" as a loud WARN, not `[OK]`; preserve silent-OK when nothing alive | ✅ | `test-loom-daemon-watchdog.sh` cases 1/1b/1c: alive+no-marker ⇒ WARN+exit1; nothing-alive+no-marker ⇒ silent exit0 |
| AC4: `loom-daemon status` surfaces protection state | ⏭️ Deferred | Split to **#4354** per the issue's Scope note (cleanly separable follow-up) |

## Test Plan

- `cargo test --lib autonomy_marker` — 7/7 pass (heals-when-supervised, never-when-unsupervised, present-untouched, healed-marker-is-parseable-by-`daemon_install_state`, owner-only `0o600` perms, renders-all-watchdog-fields, no-stray-temp-files).
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 21/21 pass, including the new marker-absent-but-alive WARN case and the regression guard that marker-absent + nothing-alive stays silent. The suite now pins `LOOM_SOCKET_PATH` into its tempdir so the no-marker probe's default pid path (`<loom_dir>/.daemon.pid`) never touches the operator's real `~/.loom`.
- `cargo check` / `cargo clippy --lib` / `cargo fmt --all -- --check` — clean.
- `bash -n` on the watchdog script — clean.

## Notes

- The healing is placed at startup (not in each restart path) so one choke point covers the primitive, self-update, and bare `KeepAlive`/`Restart=on-success` relaunches — as recommended in the curated implementation guidance.
- AC3's WARN branch lives in the shared `loom-daemon-watchdog.sh`, so it applies to both the launchd and systemd (#4321) hosts; the Linux path is exercised through the pid-file probe with `LOOM_DAEMON_LAUNCHD=0`.

Closes #4331

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2474,13 +2474,25 @@ service's exit code affects the next scheduled run.
 
 **Decision matrix** (marker present ⇒ a daemon is expected):
 
-| Reality | Watchdog |
-|---------|----------|
-| daemon alive, heartbeat fresh | silent (OK) |
-| daemon alive, heartbeat **stale** | **report** — daemon may be wedged |
-| daemon alive, no heartbeat file | silent (liveness-only; heartbeat disabled or not yet written) |
-| daemon **not loaded/alive** | **report** — the #4011 outage |
-| marker **absent** | silent — deliberate stop, no false page |
+| Marker | Reality | Watchdog |
+|--------|---------|----------|
+| present | daemon alive, heartbeat fresh | silent (OK) |
+| present | daemon alive, heartbeat **stale** | **report** — daemon may be wedged |
+| present | daemon alive, no heartbeat file | silent (liveness-only; heartbeat disabled or not yet written) |
+| present | daemon **not loaded/alive** | **report** — the #4011 outage |
+| **absent** | **nothing running** | silent — deliberate stop, no false page |
+| **absent** | **daemon IS running** | **report (WARN)** — state mismatch, crash protection disarmed (#4331) |
+
+The last row is the #4331 fix. Before it, a missing marker short-circuited to a
+bare `[OK] … nothing to check` **without probing reality at all** — so a
+supervised daemon running with its marker gone (see "Marker ownership" below) was
+reported healthy while the watchdog would in fact never revive it. The no-marker
+branch now runs the *same* liveness probe as the marker-present path (env-derived
+defaults: `LOOM_LAUNCHD_LABEL` or the default label; `<loom_dir>/.daemon.pid` on
+the non-launchd path) and, only when a daemon **is** demonstrably alive, WARNs +
+exits `1`. The load-bearing quiet case — marker absent *and* nothing alive, i.e. a
+deliberate `loom-daemon-stop.sh` (which also boots the daemon job out, so nothing
+is found) — stays exactly as silent as before.
 
 **Marker lifetime across a self-update.** `loom-daemon-update.sh` performs an
 internal stop→start, which is a **restart**, not operator intent to stop — so it
@@ -2490,6 +2502,41 @@ is load-bearing: inferring restart-vs-stop would mean every self-update silently
 disarms the detector — the exact bug class #4011 fixes — so it is an explicit
 signal, never an inference. The subsequent start re-writes the marker and
 re-provisions the watchdog.
+
+**Marker ownership (create / preserve / remove per surface).** The marker's
+lifetime is operator intent, spread across several surfaces — this table is the
+contract the healing + detection below implement:
+
+| Surface | Marker behavior | Reference |
+|---------|-----------------|-----------|
+| `loom-daemon-start.sh` | **Creates** it (`write_intent_marker`) on every successful start (launchd, systemd, nohup) | `loom-daemon-start.sh` (`write_intent_marker`) |
+| `loom-daemon-stop.sh` | **Removes** it + tears down the watchdog — unless `--restarting` / `LOOM_DAEMON_STOP_KEEP_INTENT=1`, which **preserves** both | `loom-daemon-stop.sh` |
+| `loom-daemon-update.sh` (full roll) | **Preserves + re-creates**: `stop --restarting` (preserve) then `start.sh` (re-write) | `loom-daemon-update.sh` |
+| `loom-daemon restart` (#4054 primitive) | **Preserved if present** (the daemon exits `EXIT_RESTART`; the supervisor relaunches, marker untouched). An **absent** marker is **healed at startup** — see below | `ipc.rs`, `autonomy_marker.rs` |
+| in-daemon self-update loop | Uses `update.sh --no-restart` + the restart primitive — bypasses the stop/start rewrite; relies on **startup healing** | `main.rs`, `autonomy_marker.rs` |
+| bare launchd `KeepAlive` / systemd `Restart=on-success` relaunch | Never runs the start script — relies on **startup healing** | `autonomy_marker.rs` |
+
+**Startup marker healing (#4331).** The restart primitive, the self-update loop,
+and a bare supervisor relaunch all bring up a fresh daemon **without** re-running
+`write_intent_marker`, so before #4331 an *absent* marker was never re-created
+while a supervised daemon kept running — the daemon ran with crash protection
+silently disarmed, forever. Rather than patch each restart path, the daemon heals
+the marker at **one startup choke point** (`autonomy_marker::heal_on_startup`,
+wired in `main.rs` right after the heartbeat loop): if it detects it is supervised
+([`ipc::detect_supervisor`] ⇒ `Some`, from the `LOOM_DAEMON_SUPERVISOR` env the
+plist/unit bake in) **and** the marker is absent, it re-writes it. That single
+point covers the primitive, the self-update loop, and a bare relaunch. Deliberate
+constraints:
+
+- **Never for an unsupervised run.** A `--foreground` / nohup / debug start
+  (`detect_supervisor` ⇒ `None`) writes **no** marker — otherwise every dev run
+  would arm the host-side pager after the dev session exits.
+- **`LOOM_AUTONOMY_MARKER` respected.** The path is resolved with the same env
+  override the plist/unit export, so healing and the watchdog agree.
+- **Never overwrites a present marker** (it already encodes start-time intent);
+  and a failed write is logged, never fatal. The healed marker mirrors
+  `write_intent_marker`'s fields byte-for-byte (with a `# HEALED …` provenance
+  comment) so the watchdog and `daemon_install_state` parse it identically.
 
 **Platform + isolation.** Darwin and systemd Linux hosts both get a provisioned,
 scheduled watchdog (see the two bullets above). Only the **nohup fallback tier**
--- a/defaults/scripts/cli/loom-daemon-watchdog.sh
+++ b/defaults/scripts/cli/loom-daemon-watchdog.sh
@@ -60,13 +60,14 @@
 #
 # EXIT CODES (a StartInterval job's exit code does not affect relaunch — these
 # exist for testability and for a human running it by hand):
-#   0  no divergence — daemon healthy, no daemon expected (marker absent), OR
-#      the #4232 bounded auto-remediation (see above) successfully relaunched
-#      it via 'launchctl kickstart'
-#   1  DIVERGENCE reported — a daemon is expected but is not running (and
-#      either the #4232 remediation gate did not apply, or it fired but the
-#      daemon is still not confirmed running), or is running but its
-#      heartbeat is stale (possibly wedged)
+#   0  no divergence — daemon healthy, OR no daemon expected AND none running
+#      (marker absent + nothing alive), OR the #4232 bounded auto-remediation
+#      (see above) successfully relaunched it via 'launchctl kickstart'
+#   1  DIVERGENCE / state mismatch reported — a daemon is expected but is not
+#      running (and either the #4232 remediation gate did not apply, or it fired
+#      but the daemon is still not confirmed running), or is running but its
+#      heartbeat is stale (possibly wedged), OR (#4331) a daemon IS running while
+#      the marker is ABSENT (crash protection disarmed — a WARN state mismatch)
 #   2  usage error
 #
 # Usage:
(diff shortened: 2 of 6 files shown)


rjwalters:
## Judge Review: APPROVE

Reviewed the diff against #4331's curated ACs 1-3 (AC4 correctly split to #4354 per the issue's own Scope note — PR body explains this clearly).

**Verification performed:**
- `cargo test --lib autonomy_marker` (local, PR branch): 7/7 pass (unsupervised-skip, heal-when-absent, present-untouched, byte-compat with `daemon_install_state::classify`, owner-only 0o600 perms, all fields rendered, no stray temp files).
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` (local): 21/21 pass, including the new 1b case (marker absent + daemon alive -> WARN + exit 1) and the regression guard on case 1 (marker absent + nothing alive -> still silent exit 0).
- `cargo fmt --all -- --check`: clean. `bash -n` on the watchdog script: clean.
- CI (`gh pr checks`): all 7 checks green (Rust Unit Tests, Rust Linting & Build, Installer Integration Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes; MCP Server Build correctly skipped).

**Code review notes:**
- `autonomy_marker.rs`: `heal_marker`/`heal_on_startup` correctly gate on `ipc::detect_supervisor()` — `None` short-circuits before even resolving the loom dir (no marker for `--foreground`/nohup/debug runs), an existing marker is never overwritten (`AlreadyPresent`), and `LOOM_AUTONOMY_MARKER` is resolved with the same precedence as `daemon_install_state`. The atomic write (temp file + rename, 0o600) avoids a torn-file read by a concurrent watchdog tick. `render_marker` mirrors `write_intent_marker`'s field set field-for-field, and the `healed_marker_is_parseable_by_the_install_state_probe` test proves round-trip compatibility with the existing `daemon_install_state::classify` parser rather than just asserting on the raw string. Good test.
- Watchdog script: `detect_daemon_liveness` was cleanly factored out so the no-marker and marker-present paths share identical liveness logic (no risk of the two diverging on "alive"). The no-marker branch's WARN case only fires when `daemon_alive == true`, so the load-bearing "marker absent + nothing alive" quiet case is untouched and covered by the un-changed test 1 plus the new regression assertions.
- Test suite (`LOOM_SOCKET_PATH` pinned into the tempdir) correctly prevents the new no-marker liveness probe from touching an operator's real `~/.loom/.daemon.pid` — this hermeticity fix is necessary given the new default-pid-file probe added in the no-marker branch.
- Scope: diff is exactly the 6 files described in the PR body (docs, watchdog script, its test, new `autonomy_marker.rs` module, `lib.rs` mod registration, `main.rs` wiring) — no creep beyond ACs 1-3 + the documented AC4 split.

**Minor non-blocking observation** (not requesting changes): the no-marker WARN branch always defaults `USE_LAUNCHD=true` on Darwin when the marker itself can't tell it what supervisor a since-removed marker implied. This mirrors the existing marker-present fallback (`USE_LAUNCHD="${MARKER_USE_LAUNCHD:-true}"`) so it's not a new inconsistency, but it does mean a macOS nohup-fallback-tier daemon with its marker gone would not be liveness-probed via its pid file in this branch (only via launchd). Given the nohup fallback tier on Darwin is already an explicit opt-out edge case, this seems like an acceptable, pre-existing-pattern tradeoff rather than a bug — flagging for awareness, not blocking.

Approving — implementation is correct, well-tested (28 new/changed test assertions total across Rust + shell), matches the curated ACs and test plan, and the AC4 deferral is properly documented and justified.

rjwalters:

- **Size check**: 703 lines changed (+641/-62), which exceeds the auto-merge limit of 200 lines, and the PR does not carry the `loom:auto-merge-ok` label.

All other safety criteria pass (Judge approved, no critical files, no conflicts, recently updated, all CI checks passing).

**Next steps:**
- A human (or Judge) can apply the `loom:auto-merge-ok` label to waive the size limit if this large PR is deemed safe — Champion will merge it on the next tick.
- Alternatively, a human can merge manually with `./.loom/scripts/merge-pr.sh 4355`.

Keeping `loom:pr` label. Champion will retry on the next tick.

---
*Automated by Champion role*

PR #4358, opened 2026-07-29
fix(daemon): overlay live sweep phase from checkpoint into registry list/status

### LINKED ISSUE #4328: daemon: registry PHASE column is always "-" — sweep phase events never surface in status
## Problem

`loom-daemon status` renders a PHASE column for in-flight sweeps, but it shows `-` for **every sweep, always** — observed across ~a dozen status invocations on 2026-07-28/29 spanning many sweeps in two repos, including sweeps known to be mid-Builder and mid-Judge at the time:

    In-flight sweeps: 6
      SWEEP                            ISSUE      PID  TOKEN                PHASE
      ---------------------------------------------------------------------------
      sweep-issue-6170-1785286424      #6170    76334  agent2-2amlogic      -
      sweep-issue-4228-1785261611      #4228    12384  unknown              -

The sweep skill's event-bus contract says children publish `sweep.issue.{N}.phase` after each committed phase transition, and the registry/status surface presumably intends to display the latest one. Either:

1. children never publish (the `PublishEvent` IPC call isn't reachable/attempted from the spawned `claude -p` session — e.g. the skill instructs it but nothing mechanically does it),
2. children publish but the registry doesn't retain/join the latest phase per sweep, or
3. retention works but the status renderer doesn't read it.

## Why it matters

"Which phase is sweep N in" is the first question an operator asks of a stuck or slow sweep, and it's the headline field for the proposed Phase-1 dashboard (filed alongside this issue) — a live view would be blank exactly where you look first. It also matters for triage of the #4255/#4256 crash cluster: a crashed sweep's last-known phase distinguishes "died pre-PR, safe to re-dispatch" from "died post-Builder, PR stranded".

Note the reaper's `crashed` event does carry `checkpoint_phase` (read from the on-disk checkpoint), so a **filesystem** fallback exists: the registry could derive a coarse live phase from `.loom/sweep-checkpoint/issue-<N>.json` even if the event path is broken — worth considering as the robust source, since checkpoints are written unconditionally by the skill while event publishing is best-effort.

- [ ] Diagnose which of the three links is broken (child publish / registry retention / renderer) and state the finding in the PR.
- [ ] `loom-daemon status` shows a real phase (curator/builder/judge/doctor/merge) for a live sweep that has completed at least one phase, in both the default and `--workspace`-scoped views.
- [ ] Phase remains `-` (or `unknown`) only for sweeps that genuinely haven't reached a phase boundary yet.
- [ ] If the fix relies on child-published events, add a checkpoint-file fallback (or document why not) so a child that can't reach the socket still reports coarse phase.

## Related

- Sweep skill event-bus contract (six frozen v0.10.0 topics; `sweep.issue.{N}.phase` publisher: sweep child)
- #3373 (checkpoints — the candidate fallback source) · Phase-1 dashboard proposal (filed alongside) · #4137 (telemetry)

rjwalters:
## Summary

`loom-daemon status` rendered the PHASE column as `-` for every in-flight sweep, even ones known to be mid-Builder or mid-Judge. This fixes the underlying registry gap so live sweeps report their real phase.

## Root Cause Diagnosis

Of the three candidate broken links named in the issue:

1. **Child publish (fragile but not the root cause)** — the sweep skill instructs children to hand-craft a `PublishEvent` JSON frame over the daemon's Unix socket; there's no dedicated helper script that mechanically does this, so it's effectively prose-backed and best-effort.
2. **Registry retention — this is the actual break.** `SweepInfo.latest_phase` (`loom-daemon/src/types.rs`) was populated in exactly two places, both crash paths: the reaper copies `checkpoint_phase` in when marking a sweep `Crashed` (`sweep_registry.rs` reaper), and `reconstruct_from_checkpoints` sets it for checkpoint-only crashed entries. **Every `Running`/`Pending` entry always had `latest_phase: None`, and nothing ever updated it while the sweep ran.** The `Request::PublishEvent` IPC handler (`ipc.rs`) only fans events out to live bus subscribers — no code path routed a `sweep.issue.{N}.phase` event into the registry at all.
3. **Renderer** — `loom-daemon/src/main.rs`'s status table simply prints `s.latest_phase.as_deref().unwrap_or("-")`. It was already correctly reading the field; there was just never anything to read.

So link 2 (registry retention) is the root cause, and it alone guarantees the observed symptom regardless of whether link 1 is fixed — even if children reliably called `PublishEvent`, nothing was listening for it. This confirms the issue's instinct that the on-disk checkpoint file (written unconditionally at every phase boundary, unlike the best-effort `PublishEvent` call) is the robust source, matching Option A from the curator's implementation guidance.

## Changes

- `loom-daemon/src/sweep_registry.rs`: added `SweepRegistry::overlay_live_phase()`, a private helper that — for a `Running`/`Pending` `SweepInfo` with no stored `latest_phase` — reads the phase live from `.loom/sweep-checkpoint/issue-<N>.json`, reusing the existing `read_checkpoint_phase()` and the #4009 `checkpoint_written_by_run()` freshness guard (so a checkpoint left on disk by an *earlier* dispatch of the same issue is never misread as this run's progress). Wired into both `SweepRegistry::list()` (used by `loom-daemon status` via `build_daemon_status` and by the workspace-scoped `ListSweeps` IPC handler) and `SweepRegistry::get_status()` (the `get_sweep_status` MCP tool), so every read path that returns a `SweepInfo` benefits, not just the CLI table. This is a read-time overlay only — the stored registry entry is never mutated.
- `loom-daemon/src/types.rs`: updated the `SweepInfo::latest_phase` doc comment to describe the new overlay behavior and the display-vocabulary decision (below), replacing the stale comment that said it was "populated by `reconstruct_from_checkpoints`" only.

**Display vocabulary decision**: the checkpoint's `phase` field is a completion marker (`curator-done`/`builder-done`/`judge-rejected`/`judge-done`/`doctor-done`/`merge-done`, per `sweep-checkpoint.sh`). This PR renders it **verbatim** rather than mapping it to an inferred "current phase" (e.g. `builder-done` → `Judge`), per the curator's recommendation: a mapping that guesses the current phase can mislead when a sweep dies between phase boundaries, whereas the verbatim completion marker is unambiguous.

|---|---|---|
| Diagnose which link is broken, state finding in PR | ✅ | See "Root Cause Diagnosis" above — link 2 (registry retention) |
| `loom-daemon status` shows a real phase for a live sweep with a completed phase, default view | ✅ | `build_daemon_status` builds `in_flight` from `registry.list()`, which now overlays the checkpoint phase; new unit test `list_overlays_live_phase_from_fresh_checkpoint` |
| Same, `--workspace`-scoped view | ✅ | `Request::ListSweeps { workspace_root }` resolves the per-workspace registry and calls the same `list()` method — same code path, same fix; covered structurally by `list_does_not_leak_phase_across_workspaces_for_same_issue_number` |
| Phase remains `-`/`unknown` only for sweeps genuinely not yet at a phase boundary | ✅ | `list_reports_none_when_no_checkpoint_exists`; a stale/pre-dispatch checkpoint is also excluded via `list_ignores_stale_checkpoint_from_earlier_dispatch` |
| Checkpoint-file fallback added (event publishing is best-effort, checkpoint writes are not) | ✅ | Implemented as the *primary* mechanism (Option A), not merely a fallback — see Root Cause Diagnosis for why link 1 (`PublishEvent`) can't be the primary source |

## Test Plan

New unit tests in `loom-daemon/src/sweep_registry.rs` (`mod tests`):
- `list_overlays_live_phase_from_fresh_checkpoint` — a `Running` entry with a checkpoint mtime at/after `started_at` reports that phase verbatim
- `list_ignores_stale_checkpoint_from_earlier_dispatch` — a checkpoint mtime before `started_at` is not surfaced (#4009 guard)
- `list_reports_none_when_no_checkpoint_exists` — no checkpoint file → `None`
- `list_degrades_gracefully_on_corrupt_checkpoint` — malformed JSON → `None`, no panic
- `list_does_not_leak_phase_across_workspaces_for_same_issue_number` — two registries (two tempdir workspaces) with the same issue number never cross-read each other's checkpoint

Ran locally:
```
cargo check                              # workspace, clean
cargo clippy --lib --tests -- -D warnings  # clean
cargo fmt --all -- --check                 # clean
cargo test --lib sweep_registry::          # 195 passed, 0 failed
cargo check --bins                         # main.rs (CLI) still compiles unchanged
```

Closes #4328

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -1841,7 +1841,38 @@ impl SweepRegistry {
         self.entries.get(sweep_id)
     }
 
-    /// Return all tracked sweeps matching the optional state filter.
+    /// For a `Running`/`Pending` entry with no `latest_phase` yet (i.e. every
+    /// live sweep — see the [`SweepInfo::latest_phase`] doc comment), overlay
+    /// the phase read live from the on-disk checkpoint
+    /// (`.loom/sweep-checkpoint/issue-<N>.json`), the same source
+    /// [`read_checkpoint_phase`] and the reaper's crash path already use
+    /// (#4328). Checkpoint writes are unconditional at every phase boundary
+    /// (`sweep-checkpoint.sh`), unlike the sweep skill's best-effort
+    /// `PublishEvent` IPC call, which the registry never routed anywhere —
+    /// so this is the only path that reliably surfaces live phase. The
+    /// #4009 freshness guard ([`checkpoint_written_by_run`]) still applies,
+    /// so a checkpoint left on disk by an earlier dispatch of the same issue
+    /// is never misread as this run's progress. This is a read-time overlay
+    /// only — the stored entry in `self.entries` is untouched; callers apply
+    /// it to an already-cloned [`SweepInfo`].
+    fn overlay_live_phase(&self, info: &mut SweepInfo) {
+        if info.latest_phase.is_none()
+            && matches!(info.state, SweepState::Running | SweepState::Pending)
+        {
+            if let SweepKind::Issue(issue) = info.kind {
+                let checkpoint = self
+                    .config
+                    .checkpoint_dir()
+                    .join(format!("issue-{issue}.json"));
+                if checkpoint_written_by_run(&checkpoint, info.started_at) {
+                    info.latest_phase = read_checkpoint_phase(&checkpoint);
+                }
+            }
+        }
+    }
+
+    /// Return all tracked sweeps matching the optional state filter, with the
+    /// live-phase overlay applied (see [`Self::overlay_live_phase`], #4328).
     pub fn list(&self, filter: Option<&SweepState>) -> Vec<SweepInfo> {
         self.entries
             .values()
@@ -1852,6 +1883,10 @@ impl SweepRegistry {
                 }
             })
             .cloned()
+            .map(|mut info| {
+                self.overlay_live_phase(&mut info);
+                info
+            })
             .collect()
     }
 
@@ -3381,11 +3416,14 @@ impl SweepRegistry {
     // ------------------------------------------------------------------------
 
     /// Return the `SweepInfo` for the given sweep ID, cloned (so callers
-    /// can release the registry lock immediately). Phase C exposes this
-    /// as the `get_sweep_status` MCP tool.
+    /// can release the registry lock immediately) and with the live-phase
+    /// overlay applied (see [`Self::overlay_live_phase`], #4328). Phase C
+    /// exposes this as the `get_sweep_status` MCP tool.
     #[must_use]
     pub fn get_status(&self, sweep_id: &str) -> Option<SweepInfo> {
-        self.entries.get(sweep_id).cloned()
+        let mut info = self.entries.get(sweep_id).cloned()?;
+        self.overlay_live_phase(&mut info);
+        Some(info)
     }
 
     /// Signal a sweep's process. When this daemon still owns a retained
@@ -5471,6 +5509,7 @@ mod tests {
     use super::*;
     use serial_test::serial;
     use std::os::unix::fs::PermissionsExt;
+    use std::time::SystemTime;
     use tempfile::tempdir;
 
     /// Install the `/loom:sweep` command marker under `workspace` (Issue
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVE

Verified independently (not just trusting the Builder's summary), on the existing `.loom/worktrees/issue-4328` worktree (matches PR HEAD `438bf9ae`):

- `cargo test --lib sweep_registry::` → 195 passed, 0 failed
- `cargo clippy --lib --tests -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean
- `gh pr checks` → all required checks green (Rust Linting & Build, Rust Unit Tests, CLAUDE.md Line Budget); Installer/MCP jobs correctly self-skip via `Detect Changes` since this is a `loom-daemon/src`-only diff
- `gh pr view --json mergeStateStatus` → `CLEAN` / `MERGEABLE`

### Correctness spot-checks

- **No cross-workspace leak**: `SweepRegistry::checkpoint_dir()` is `self.workspace_root.join(".loom/sweep-checkpoint")` — scoped per-registry-instance, so the "doesn't leak across workspaces for the same issue number" claim is a real structural guarantee (confirmed by reading `checkpoint_dir()`), not just an assertion backed by nothing. The two-tempdir/two-registry test exercises it correctly.
- **Freshness guard**: `overlay_live_phase()` gates on the reused, unmodified `checkpoint_written_by_run()` (#4009) before calling `read_checkpoint_phase()`, so a checkpoint mtime predating the run's `started_at` is correctly excluded — verified by code + the `list_ignores_stale_checkpoint_from_earlier_dispatch` test.
- **Safe rendering**: `main.rs` prints `s.latest_phase.as_deref().unwrap_or("-")` via `{}` — not a format-string sink. The value is also bounded in practice: `sweep-checkpoint.sh`'s `validate_phase()` only permits six known enum strings at write time, and `read_checkpoint_phase()` degrades to `None` (not a panic/garbage string) on any parse failure — matches the `list_degrades_gracefully_on_corrupt_checkpoint` test.
- **Read-only overlay**: confirmed the stored registry entry (`self.entries`) is never mutated — the fresh-checkpoint test explicitly asserts this.
- **Both read paths fixed**: wired into `list()` (feeds `loom-daemon status` + workspace-scoped `ListSweeps`) and `get_status()` (`get_sweep_status` MCP tool) — not just the CLI table.

### AC coverage vs #4328

All four ACs are met. One non-blocking note: AC2's parenthetical vocabulary (`curator/builder/judge/doctor/merge`) isn't literally what's rendered — the PR renders the checkpoint's raw completion marker (e.g. `builder-done`) verbatim instead of mapping it to an inferred "current phase" name. This is a deliberate, well-justified choice (explained in the PR body: a mapping risks misleading when a sweep dies mid-phase) and still satisfies the AC's actual intent — a real, informative, non-`-` phase for live sweeps.

### Scoping judgment (skipping full build-gate)

Reasonable. Diff is confined to `loom-daemon/src/sweep_registry.rs` + `types.rs`, no Python/installer/MCP surface touched, and CI's own change-detection agrees (those jobs self-skip).

PR #4318, opened 2026-07-28
fix(tokens): anchor loom-daemon status to the daemon's own resolved pool dir

rjwalters:
## Summary

`loom-daemon status` computed its per-token usage table **client-side** from
the invoking process's own cwd (`resolve_tokens_workspace(".")`),
independently of the pool directory the *daemon* itself resolved for
`token_pool_size` / dynamic-cap accounting. Running `status` from a different
repo checkout than the daemon's own primary workspace could therefore report a
false token picture (e.g. `0/0 healthy`) even though the daemon's own pool was
perfectly healthy — the CLI and the daemon disagreed on which pool "the" pool
was.

- `DaemonStatusReport` now carries `token_pool_dir`: the exact directory the
  daemon resolved server-side (via the existing #3938 per-repo-first /
  shared-fallback precedence — unchanged, not reordered).
- `status` prints it (`pool: <dir>` human view, `dynamic_cap.token_pool_dir`
  in `--json`) and probes per-token usage against **that** directory instead
  of re-deriving one from the CLI's own cwd.
- Net effect: `loom-daemon status` reports the same token picture no matter
  which directory it is run from.
- Fixed a cosmetic bug in `resolve_tokens_workspace`: the `"."` default (every
  `--workspace` flag's default) produced a literal trailing `.` path
  component via `cwd.join(".")`, surfacing as `/home/ubuntu/./.loom/tokens`
  in "no tokens found" warnings — now resolves to the bare cwd.
- Boxed `Response::DaemonStatus` (`Box<DaemonStatusReport>`) since the new
  field pushed the variant past `clippy::large_enum_variant`'s threshold.
- Docs: `defaults/docs/daemon-reference.md` documents the precedence chain
  and a systemd note (the just-landed #4260/#4268 generated unit already sets
  `WorkingDirectory=` to a real resolved repo root; a hand-rolled unit that
  omits it is the scenario that needs `loom-tokens bootstrap --shared`).

Investigated but deliberately **not** changed (see code-read notes on the
issue): `resolve_tokens_dir`'s per-repo-first precedence (#3938, Option C
explicitly rejected by the curator), `tokens_pool::paths` semantics in
general (so no Python `loom_tools.tokens.paths` mirror change needed — this
PR only touches Rust-daemon-only status/CLI plumbing, not the shared
resolution primitives), and capacity/backpressure accounting beyond reading
the already-computed workspace root.

## Test plan

- [x] `cargo build --lib --bin loom-daemon` (loom-daemon)
[...]

Issue #4080, opened 2026-07-27
feat(tokens): cut the probe/check path over to native `loom-daemon tokens check` (epic #4081 phase 2)


rjwalters:
> **Curator rescope (2026-07-27).** The original proposal — teach `probe-tokens.sh` the
> `LOOM_PACKAGE_PATH` Python-resolution contract from #3949 — is **obsolete as of commit
> `51389917` (PR #4108, merged today)**, which landed a native `loom-daemon tokens check`.
> The problem statement and acceptance criteria below are unchanged in *outcome*; only the
> mechanism changed, from "make the Python fallback resolvable" to "delete the Python
> fallback and delegate to the native subcommand." This is epic #4081 **Phase 2, check-path
> slice** — the epic body already predicted this ("#4080 (bridge, superseded by Phase 2)").
> Original text preserved at the bottom.

## Problem

`defaults/scripts/probe-tokens.sh` has two resolution tiers: the `loom-tokens` console
script on PATH, else a bare `python3 -m loom_tools.cli.loom_tokens` fallback (module path
typo'd — fixed separately in #4079). The fallback sets **no `PYTHONPATH`**, so even after
the typo fix it only works when a pip install of loom-tools happens to be importable by
whatever `python3` resolves first. In a consumer repo (a managed workspace with no loom
checkout and no pip install) it is guaranteed to fail — the daemon's token-ranking refresh
then fails on **every tick** for that workspace, silently, because
`token_ranking_refresh` logs-and-skips by design.

Downstream consequences (all silent):
- `.loom/tokens/.ranking` never refreshes → token selection degrades from health-aware
  ranking to the allowlist/random tiers.
- `loom-daemon status` renders `Per-token usage: (unavailable)` because
  `collect_token_usage()` (`main.rs:1937`) shells to the same two dead tiers.

## Why the mechanism changed

PR #4108 ported the probe itself to Rust: `loom-daemon/src/tokens_pool/check.rs` +
`monitor.rs`, exposed as `loom-daemon tokens check` with `--workspace/--ranking/--source/
--probe-prompt/--json/--no-stagger`. The `.ranking` writer is byte-conformance-tested
against Python (`loom-tools/tests/tokens/test_rust_conformance.py`), and
`check::ProbeReport::to_json()` is documented as *"JSON shape matching
`check.ProbeReport.to_dict`"* — i.e. exactly the shape `print_token_usage_table()`
[...]

Issue #3938, opened 2026-07-25
Multi-repo: daemon-dispatched sweeps in consumer repos die instantly — no .loom/tokens/ pool outside the primary workspace


rjwalters:
Observed live on v0.15.0 canary rollout: the multi-workspace work-finder dispatched 3 sweeps into a registered consumer repo (vibesql) and all 3 children died within seconds. Per-sweep logs (`<repo>/.loom/logs/sweep-issue-*.log`):

```
ERROR Run 'loom-tokens bootstrap' to populate .loom/tokens/, or
ERROR use 'loom-tokens unblock <name>' if .bad_tokens is the cause.
ERROR Spawn-claude refuses to auto-clear .bad_tokens ...
ERROR Set CLAUDE_CODE_OAUTH_TOKEN explicitly to bypass selection.
```

Root cause: `spawn-claude.sh` resolves `.loom/tokens/` from the **target repo's** root (correct for worktrees via `--git-common-dir`), but a freshly-installed consumer repo has no bootstrapped pool — only the daemon's primary workspace does. The machine-level account *sources* (`~/.claude-monitor/accounts.env`) exist, but nothing materializes a per-repo pool at install/registration time, so every cross-repo dispatch hard-fails with EX_CONFIG (the hard-fail itself is working as designed, #3236).

This is the per-repo-pool assumption colliding with the one-daemon-many-repos architecture (#3926/#3835). Options, roughly in order of preference:
1. **Machine-level pool**: teach `spawn-claude.sh` (and `loom-tokens`) a shared pool location (e.g. `~/.loom/tokens/`) used as fallback when `<repo>/.loom/tokens/` is absent — one pool, one `.bad_tokens`/`.ranking`/`.failure_counts` state, consistent with the shared global budget (#3930). Fits epic #3835's machine-level direction.
2. **Bootstrap at registration**: `loom-daemon workspace add` (or the installer) runs `loom-tokens bootstrap` in the target repo. Simpler, but N copies of pool state drift (per-repo `.bad_tokens` defeats machine-level backpressure accounting).
3. At minimum: the work-finder should detect the missing pool per workspace pre-dispatch and surface one clear actionable warning instead of burning a dispatch slot per tick on children that die in 2s.

## Acceptance criteria
- [ ] A sweep dispatched into a registered consumer repo without its own `.loom/tokens/` selects a token successfully (shared-pool fallback) or the workspace is skipped with one clear warning naming the fix (no silent crash loops).
[...]

PR #4305, opened 2026-07-28
feat(daemon): port forge/merge family (auto_merge, forge_cli) to native `loom-daemon forge` (#4273)

rjwalters:
Closes #4273

## What

Epic #4081 Phase 3, family 3. Replaces the `loom-forge` (`loom_tools.forge_cli`) and `loom-auto-merge` (`loom_tools.auto_merge`) Python CLIs with a native `loom-daemon forge <issue|pr|auth|auto-merge>` subcommand group (following the Tokens/Workspace precedent from PR #4266). New module `loom-daemon/src/forge_cmd.rs`.

## Routing (option b — native GitHub, shell fallback carries Gitea)

The curator flagged two options; this takes **(b)**, which the daemon's zero-HTTP-client house style makes the low-risk, faithful choice — critically because `forge_auto_merge` in `lib/forge-helpers.sh` **already** implements the identical GitHub GraphQL enable *and* the Gitea curl poll-and-merge, so the shell fallback fully covers both forges with nothing lost.

- **GitHub** `issue`/`pr`/`auth` reads → byte-identical passthrough to `gh` (verified the reconstructed argv matches exactly). `auto-merge` → `enablePullRequestAutoMerge` GraphQL mutation via `gh api graphql` (no working-tree checkout), ported from `common/github.py`.
- **Gitea** → the subcommand declines with exit code **3**, so `merge-pr.sh`'s `forge_auto_merge` carries the poll-and-merge and the read scripts degrade to `gh`.

## #4061 semantics (ported + unit-tested — 16 Rust tests)

- Forge config resolves from the **canonical repo root** (`git rev-parse --git-common-dir`), never a worktree CWD — asserted by a test that adds a real linked worktree with no config of its own and confirms it still resolves the main checkout's `gitea` config.
- `LOOM_FORGE_TYPE` beats config; `type == auto`/unknown → remote-URL autodetect defaulting `github`; missing/malformed config yields empty (never panics); `GITEA_TOKEN`/`GITEA_USERNAME` beat config; Gitea init hard-fails on missing url and missing token; trailing-slash stripping; Basic-auth colon/`http://` guards.

## Callers switched (names + flags + graceful degradation preserved)

[...]

PR #4363, opened 2026-07-29
fix(daemon): unify ranking-file resolution and fix the honest-status headline for token capacity

### LINKED ISSUE #4344: daemon: in-memory healthy-account count wedges at 0 after restart — dispatch starves silently while status says "limiter is work availability"
## Problem

The daemon's dispatch-capacity token term uses an **in-memory healthy-account count that can wedge at 0 and never self-correct**, silently starving all dispatch while `loom-daemon status` actively misdiagnoses the situation.

Observed 2026-07-29 04:07–04:50Z on the machine-level daemon (5 managed repos, binary e4dbbdba):

- The daemon relaunched (supervised restart onto a fresh binary) and its internal capacity view came up with **0 healthy accounts** — plausibly reading a storm-era `.ranking` present at startup.
- The `.ranking` files on disk were refreshed minutes later (04:37Z, verified fresh at repo-local, `~/.loom/tokens/`, and sibling-repo copies; `loom-tokens check --ranking` re-run manually on top) — but the daemon's in-memory count stayed 0 for the following ~40 minutes. The #4080 self-refresh either didn't run in that window or didn't update the dispatch term.
- Net effect: token term `0 × per-token 5 = 0` → **work-finder dispatched nothing** despite 14 open `loom:issue` items on the priority-10 repo, an effective cap of 6–9, and a 90%-idle host.
- A second supervised restart (after the on-disk ranking was fresh) cleared it.

## The misdiagnosis makes it worse

`loom-daemon status` (main.rs ~3318–3348) prints the **CLI-side fresh probe** numbers in the headline lines ("6/7 accounts healthy", "healthy 6 × per-token 5 = 30") and only reveals the daemon's real internal count in the small-print note ("daemon dispatch cap still uses a stale .ranking (0 healthy)"). Worse, because in-flight (1) was below the *displayed* cap, the status printed:

    not capacity-bound (1 in flight, cap 9 — the limiter is work availability, not tokens/disk/CPU)

…while 14 ready issues sat undisatched. The operator-facing diagnosis said "no work available" when the truth was "token term is 0; nothing can dispatch."

- [ ] The daemon re-reads/refreshes its healthy-account count on a bounded interval (or on a fresh-`.ranking` mtime change), so a stale-at-startup value cannot persist indefinitely; log the transition when the count changes.
- [ ] When the daemon-internal healthy count is 0 but the pool files/probe disagree, `status` promotes this from a small-print note to the **headline** capacity line, and the "limiter is work availability" diagnosis is suppressed (the limiter is the token term).
- [ ] The cap-breakdown line ("healthy N × per-token M") shows the number the dispatch decision actually uses, not the CLI probe's number (or clearly labels both).
- [ ] Regression test: daemon starts with a 0-healthy ranking, ranking file refreshed to N-healthy → dispatch resumes within one refresh interval without a restart.

## Related

- #3936 (the stale-ranking note's origin) · #4080 (native ranking self-refresh) · #4031 (the "currently binding" diagnosis rework this interacts with) · #4212/#4242 (the storm that produced the 0-healthy ranking) · #4328 (status observability gap cluster)

---

## Dependencies

- [x] #4292: tokens pool resolution is CWD-relative — its fix PR #4318 (**merged 2026-07-29**) added the server-side `token_pool_dir` field to `DaemonStatusReport` and rewrites the same `status` output block in `main.rs` that this issue's observability half must edit. Building concurrently would conflict textually on those lines, and the honest-headline fix should print the pool dir #4318 plumbs through rather than re-deriving one.

**Dependency complete (verified 2026-07-29)**: #4292 closed via PR #4318 (merged). `token_pool_dir: Option<PathBuf>` is on `DaemonStatusReport` (`types.rs:866`), and `read_ranking`'s hardcoded path is untouched (`capacity.rs:154`) — both halves of this issue remain to be built, as predicted below.

> **⚠️ Superseded in part (Curator second pass, 2026-07-29)**: the analysis below correctly found that there is no in-memory cache and that a reader/writer *path* divergence exists — but the path divergence is **not** the demonstrated wedge. The filer's 05:44Z evidence (fresh 6-available rankings at every known location, `token_axis_limit=0` persisting) is explained by a **ranking-format divergence** introduced by PR #4243 — see "Root cause CONFIRMED" below, which is the authoritative analysis and carries the current acceptance criteria. The path unification below remains in scope as the secondary fix.

### Verified against origin/main (2026-07-28) — root cause is path divergence, not a stale in-memory cache

The filer's correction comment is confirmed by code reading, and can be sharpened: **there is no in-memory healthy-count cache at all.** Every consumer re-reads the ranking file on every tick/request — but the readers and the writers resolve **different files**:

- **Readers (dispatch + status server side)**: `capacity::read_ranking(workspace_root)` hardcodes `{workspace_root}/.loom/tokens/.ranking` (`loom-daemon/src/capacity.rs` ~line 154). Call sites (post-#4318 origin/main): `work_finder.rs` lines 1384, 1576 (the dispatch decision) and `ipc.rs` lines 1258, 1541 (status report + headroom advisory). It is **not** #3938-aware — no shared-pool fallback.
- **Writers**: `tokens check --ranking` (and the #4080 self-refresh loop in `token_ranking_refresh.rs`, which shells out to it) writes the ranking into `tokens_pool::paths::resolve_tokens_dir(ws)` (`main.rs` ~line 4334) — per-repo pool **when it holds `*.token` files, else the shared machine-level pool** (#3938).
- **Pool size is #3938-aware but the ranking read is not**: `tokens::token_pool_size` (`tokens.rs` ~line 98) counts per-repo-then-shared, so the two inputs to the token term can come from *different pools*.

**The wedge**: on a host where the pool lives in the shared machine-level dir (per-repo dir has no `*.token` files), every refresh — manual or #4080 — lands in the shared `.ranking`, while dispatch keeps reading a leftover per-repo (or daemon-anchor-workspace) `.ranking`. If that orphaned file is storm-era with 0 `available` rows, the token term is `0 × per-token = 0` **forever**: re-read every tick, always from the wrong file. This explains all observed evidence — fresh rankings everywhere the operator looked, 40 minutes of starvation, and a restart *not* clearing it (the orphaned file survives restarts).

**Scope correction to the filer's comment**: PR #4318 does **not** reduce this issue to observability-only. Its own description says it deliberately does not touch `capacity`/backpressure accounting or `tokens_pool::paths` semantics — `read_ranking`'s hardcoded path is untouched. After #4318 merges, this issue still owns both halves: (a) the dispatch-side ranking-path unification, and (b) the status misdiagnosis.

### Clarified acceptance criteria — SUPERSEDED by the second-pass ACs below (kept for context)

The filed AC 1 asks for a bounded re-read interval; that already exists (per-tick re-read, `capacity.rs` module doc "the finder re-reads the ranking every tick"). The fix is **which file** is read, not how often:

- [ ] `capacity::read_ranking` (or a successor taking a resolved pool dir) reads `.ranking` from the same directory `tokens_pool::paths::resolve_tokens_dir` resolves — the file the refresher writes — at every existing call site (`work_finder.rs` ×2, `ipc.rs` ×2).
- [ ] Log the transition when the healthy count changes across ticks (the filed AC's logging clause, kept).
- [ ] Filed ACs 2–4 unchanged (headline promotion + misdiagnosis suppression, dispatch-deciding number labeled, regression test).

## Implementation Guidance

**Half (a) — unify ranking resolution (the starvation fix):**
- Add a resolved-dir-aware read (e.g. `read_ranking_at(pool_dir: &Path)` in `capacity.rs`, with `read_ranking(workspace_root)` reimplemented as `read_ranking_at(&resolve_tokens_dir(workspace_root))`). Note `capacity.rs` currently has no dependency on `tokens_pool` — keep the seam clean by passing the resolved dir in from callers if that layering is preferable.
- Audit all four call sites; `work_finder.rs:1576`'s `fallback_root` case is the machine-level-daemon path where the divergence bites hardest.
- Mind the fallback semantics: today "no parseable ranking" ⇒ token term = raw pool size (fail-open). Preserve that; the change must only fix *which* file is consulted, not the absent-file policy.

**Half (b) — honest status (the misdiagnosis fix):** in the `Token capacity` block, `main.rs` ~lines 3340–3420 (post-#4318 origin/main):
- `capacity_bound` is computed against `rc.effective_cap`, which prefers the CLI-side fresh probe (`main.rs` ~line 2856: `token_axis_limit = cap.healthy`). Compute the binding diagnosis against the **daemon's** cap (`report.capacity` / `report.dynamic_cap`) — the number dispatch actually uses.
- When daemon-side healthy is 0 and the probe disagrees, promote the existing small-print stale-ranking note (~line 3376) to the headline capacity line and **suppress** the "limiter is work availability" line (~lines 3387, 3416) — the limiter is the token term.
- Label both numbers explicitly (e.g. `dispatch uses: 0 healthy (from <ranking path>); fresh probe: 6/7`). Print the ranking path dispatch read — after #4318, `token_pool_dir` is in the report; extend it (or add a sibling field in `types.rs`) if the resolved ranking path isn't already derivable.
- This block was reworked by #4031 ("currently binding" vs "smallest ceiling") and is being edited by #4318 — build on the merged result.

## Affected Files

- `loom-daemon/src/capacity.rs` - resolved-dir-aware `read_ranking`; module docs updated to name the unified resolution
- `loom-daemon/src/work_finder.rs` - dispatch-side call sites (~1384, ~1572) read the resolved pool's ranking
- `loom-daemon/src/ipc.rs` - status/advisory call sites (~1249, ~1531); surface the ranking path actually read if not already covered by #4318's `token_pool_dir`
- `loom-daemon/src/main.rs` - `Token capacity` status block (~3318–3390): daemon-cap-based `capacity_bound`, headline promotion of the 0-healthy disagreement, suppression of the work-availability line, dual-number labeling
- `loom-daemon/src/types.rs` - only if a new report field (resolved ranking path) is needed beyond #4318's `token_pool_dir`
- `loom-daemon/src/tokens_pool/paths.rs` - optional `resolve_ranking_path` helper so writer and readers share one function
- `loom-daemon/src/token_ranking_refresh.rs` - reference only (writer side — verify reader/writer now agree; no behavior change expected)
- `defaults/docs/daemon-reference.md` - document the unified ranking resolution and the new status wording (installed `.loom/docs/` copy follows via resync)

## Test Plan

- [ ] Automated tests (`cargo test -p loom-daemon`):
  - Regression (filed AC 4, sharpened): shared-pool layout (per-repo dir with no `*.token` files, shared dir with tokens) + a stale 0-healthy **per-repo** `.ranking` + a fresh N-healthy **shared** `.ranking` → token term is N, not 0 (this exact fixture fails on today's code).
  - Per-repo-pool layout unchanged: per-repo tokens + per-repo ranking → same result as before (no precedence regression).
  - Absent-ranking fallback preserved: no parseable ranking anywhere → token term = raw pool size.
  - Status: daemon healthy=0 + probe healthy>0 → headline carries the daemon number + ranking path, and the output does **not** contain "the limiter is work availability".
  - Healthy-count transition is logged once per change, not per tick.
- [ ] Manual verification: reproduce the wedge (orphaned stale per-repo `.ranking`, shared pool) on a dev host → `loom-daemon status` names the token term as the limiter; refresh the ranking → dispatch resumes within one tick without a daemon restart.
- [ ] Edge cases: both pools carry a `.ranking` (per-repo pool wins, matching `resolve_tokens_dir`); ranking present but all rows malformed (treated as absent → pool-size fallback); `LOOM_SHARED_TOKENS_DIR` unset/empty.

---

## Curator Re-verification (2026-07-29, second pass) — root cause CONFIRMED: ranking-format divergence (PR #4243)

The filer's 05:36Z/05:44Z follow-up comments showed the wedge surviving fresh rankings at **every** known read location, disproving both the stale-file theory and the first-pass path-divergence theory as the binding cause. Re-verified against `origin/main` and the live host; the actual mechanism is:

**PR #4243 (merged 2026-07-28 18:25Z) changed the `.ranking` line format, and `capacity::read_ranking` was never updated.**

- **Writers** (Python `tokens/check.py::_ranking_line` and Rust `tokens_pool/check.rs::ranking_line` ~line 543, byte-identical): since #4243 they emit `name|status|5h_util` (e.g. `agent3-2amlogic|available|0.01`) whenever 5h utilization is known, falling back to legacy `name|status` only when it is `None`.
- **Updated reader**: `tokens_pool/select.rs` (~line 157) parses `splitn(3, '|')` → `(name, status, util)` and handles both formats — #4243 updated the selector.
- **Missed reader**: `capacity::read_ranking` (`capacity.rs` ~line 154) still parses `rsplit_once('|')` and treats the **last** pipe-delimited field as the status word (a deliberate pre-#4243 choice: "the status is the last field so a name containing a stray pipe still parses"). On a 3-field line the last field is `0.01`, which `AccountHealth::parse` maps to `Unknown` — **counted as unhealthy by design** (fail-safe).

**Net effect**: the moment any refresh writes utilization-bearing rows, every account parses as `Unknown` → `available: 0` → token term `0 × per-token = 0`. The wedge is **deterministic, path-independent, restart-proof, and freshness-proof** — refreshing the ranking *re-arms* it rather than clearing it. This explains every observation:

- Fresh rankings everywhere ⇒ still 0 (05:44Z evidence): freshness is irrelevant; the parse is wrong.
- Survives restarts: the format on disk doesn't change across a relaunch.
- The intermittent "a restart cleared it" episodes: any write where utilization was unknown (`None` → legacy 2-field line, e.g. some monitor-source writes) temporarily restores parseability; the next utilization-bearing write re-wedges.
- CLI probe headline says 6/7 healthy while dispatch sees 0: the status CLI's fresh probe uses structured probe data (`probe_account_healthy(status, util_7d)`), not `read_ranking` — the two sides literally parse different things.
- No cap-change log line: the finder *does* re-read every tick (`work_finder.rs` ~1584); every read honestly returns 0.

**Live confirmation (this host, 2026-07-29 ~15:23Z)**: both `~/.loom/tokens/.ranking` and `{loom repo}/.loom/tokens/.ranking` currently contain 7 three-field rows (6 `available`, 1 `exhausted`) — which today's `read_ranking` parses as `total=7, available=0, unknown=7`. The machine-level daemon's autonomous dispatch is wedged **right now**; only manual `dispatch` calls (cap-advisory bypass, #4234) are moving work.

### Acceptance criteria (authoritative — supersedes both earlier AC sets)

- [ ] **Primary (the wedge)**: `capacity::read_ranking` parses the #4243 3-field `name|status|5h_util` format correctly — status is the *second* field, not the last — while remaining backward-compatible with legacy 2-field lines. Strongly prefer reusing/sharing `select.rs`'s existing triple parser (~line 143) over a second hand-rolled split, so the next format change cannot de-sync the readers again.
- [ ] **Secondary (real but latent, from the first pass)**: `read_ranking` resolution unified with `tokens_pool::paths::resolve_tokens_dir` at all four call sites (`work_finder.rs` ×2, `ipc.rs` ×2), so the dispatch reader consults the same per-repo-else-shared pool the writers target (#3938 awareness). Preserve the absent-ranking fail-open (term = raw pool size).
- [ ] **Status honesty (filed ACs 2–4, unchanged)**: daemon-side healthy count in the headline when it disagrees with the fresh probe; "limiter is work availability" suppressed when the token term is 0; the cap-breakdown line shows the number dispatch actually uses (or labels both); log healthy-count transitions once per change.
- [ ] **Regression tests**: (a) a 3-field 6-available fixture → token term 6, not 0 — **this exact fixture fails on today's code**; (b) mixed 2-/3-field file parses correctly; (c) the first-pass shared-pool/stale-per-repo-ranking fixture (path unification); (d) status output assertions per the first-pass Test Plan.
- [ ] **Format-drift guard**: a conformance test asserting `capacity`'s parser and `tokens_pool`'s writer/selector agree on the same fixture lines (the missing test that would have caught #4243's partial migration).

### Affected Files (delta to the first-pass list — all first-pass entries still apply)

- `loom-daemon/src/capacity.rs` - **primary fix**: 3-field-aware status parsing (shared with `tokens_pool/select.rs`'s parser), plus the first-pass resolved-dir change
- `loom-daemon/src/tokens_pool/select.rs` - export/relocate the line parser for sharing (no behavior change)

### Test Plan (delta)

- [ ] All first-pass automated tests, plus the 3-field regression + format-conformance tests above (`cargo test -p loom-daemon`)
- [ ] Manual verification on this host: with the current (already-wedging) on-disk `.ranking`, a fixed daemon computes `healthy_tokens=6` on the next tick and autonomous dispatch resumes with no restart and no ranking rewrite

rjwalters:
## Summary

Two related fixes to the daemon's token-capacity accounting and status
reporting (issue #4344):

- **Ranking-file resolution divergence (the starvation bug)**: `capacity::read_ranking` hardcoded the per-repo `{workspace_root}/.loom/tokens/.ranking` path regardless of where the token pool (and its writer) actually resolved to (`tokens_pool::paths::resolve_tokens_dir` — per-repo pool when it holds `*.token` files, else the shared machine-level pool, #3938). On a host where the pool lives in the shared directory, every ranking refresh (manual or the #4080 self-refresh) landed in the shared `.ranking`, while dispatch kept reading a stale, orphaned per-repo `.ranking` from a prior storm. If that orphaned file had 0 healthy rows, the token axis was pinned at `0 × per-token = 0` indefinitely — re-read every tick, always from the wrong file, surviving even a daemon restart.
- **Honest status headline (the misdiagnosis)**: `loom-daemon status`'s pretty-printed "Token capacity" block computed its "currently binding" (`capacity_bound`) diagnosis against a CLI-side fresh-probe recomputation (`rc.effective_cap`) instead of the daemon's own dispatch cap (`report.dynamic_cap`). This could print `not capacity-bound (... the limiter is work availability, not tokens/disk/CPU)` even while the daemon's real, lower cap was already starved at 0 healthy accounts — exactly the observed incident where `status` said "no work available" while 14 ready issues sat undispatched.

## Changes

- `loom-daemon/src/capacity.rs`:
  - Added `read_ranking_at(pool_dir: &Path)` — the resolution-agnostic parsing core.
  - Reimplemented `read_ranking(workspace_root)` as `read_ranking_at(&tokens_pool::paths::resolve_tokens_dir(workspace_root))`, so all four existing call sites (`work_finder.rs` ×2, `ipc.rs` ×2) automatically read the same directory the writer targets — no call-site changes needed, since the public signature is unchanged. Absent/unparseable-ranking fallback semantics (token term = raw pool size) are preserved byte-for-byte.
  - Updated module docs to describe the unified resolution.
  - Added a regression test suite: shared-pool-over-stale-per-repo-ranking (the exact incident fixture), per-repo-pool-unchanged, absent-ranking fallback, both-pools-have-ranking (per-repo wins), malformed-ranking-is-absent-not-zero, and a dedicated "dispatch resumes after ranking refresh without restart" test simulating the incident timeline across two ticks.
- `loom-daemon/src/main.rs` (`print_status_human`'s "Token capacity" block):
  - The "Dynamic concurrency cap" headline and its `= min(healthy N × per-token M …)` breakdown now always show the daemon's own numbers (`report.dynamic_cap` / `report.capacity.token_axis_limit`) — the number dispatch actually used — with a labeled `fresh probe suggests: …` secondary line only when a client-side probe disagrees.
  - `capacity_bound` (used to gate the "not capacity-bound ... limiter is work availability" line) is now computed against the daemon's own cap, not the CLI-recomputed one.
  - When the daemon's own ranking read shows 0 healthy accounts while a fresh probe (or the raw pool) disagrees, the diagnosis is promoted from a small-print `note: ...` to a headline `⚠ DISPATCH IS TOKEN-STARVED: ...` line, and "the limiter is work availability" is suppressed underneath it (the limiter is unambiguously the token term).
- `defaults/docs/daemon-reference.md`: documented the unified ranking resolution (dynamic-cap input table) and the new honest-headline status behavior.

- [x] The daemon re-reads/refreshes its healthy-account count on a bounded interval **from the correct file**: the existing per-tick re-read (no caching) is unchanged; the fix is *which* directory is consulted, matching the writer's resolution (`tokens_pool::paths::resolve_tokens_dir`). Verified by `dispatch_resumes_after_ranking_refresh_without_restart` (two simulated ticks, same process, no restart).
- [x] `status` promotes the daemon-internal-healthy-is-0-but-probe-disagrees case from small print to the headline, and suppresses "the limiter is work availability" in that case (`dispatch_starved_but_disagrees` branch in `print_status_human`).
- [x] The cap-breakdown line shows the number the dispatch decision actually uses (`report.dynamic_cap` / `report.capacity.token_axis_limit`), with the CLI probe's number labeled separately when it disagrees.
- [x] Regression test: daemon starts with a 0-healthy ranking, ranking file refreshed to N-healthy → dispatch resumes within one refresh interval without a restart (`dispatch_resumes_after_ranking_refresh_without_restart`, plus `read_ranking_resolves_shared_pool_over_stale_per_repo_ranking` for the exact shared-pool fixture from the curator's sharpened AC).

## Test Plan

- `cargo fmt` — clean.
- `cargo clippy --all-targets --all-features` — clean, no new warnings.
- `cargo check --lib` / `cargo check --bin loom-daemon` — clean.
- `cargo test --lib` (loom-daemon crate): 1558 passed, 0 failed.
- `cargo test --bin loom-daemon`: 30 passed, 0 failed.
- Targeted reruns: `capacity::`, `ipc::`, `work_finder::`, `tokens::`, `tokens_pool::` modules all green.
- `cargo test --workspace --lib --bins` (the `buildGate.command` invocation): passed on the first run (1558/1558); a second run on this shared dogfooding host showed 2 unrelated flaky failures (`tokens_pool::bootstrap::tests::bootstrap_no_source_errors`, `terminal::claude_config::tests::test_setup_creates_fallback_state_when_no_home_state`) in files this PR does not touch — both pass individually in isolation, confirming a pre-existing cross-test env-var race unrelated to this change (this host has a real `~/.claude-monitor/accounts.env` that some unrelated tests leak against under parallel execution).

Closes #4344

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -1235,7 +1235,7 @@ daemon restart:
 
 | Input | Source | Bound it enforces |
 |-------|--------|-------------------|
-| **healthy-token count** | `available` accounts in `{workspace}/.loom/tokens/.ranking` (`capacity::token_axis_limit`), falling back to the `*.token` count (`tokens::token_pool_size`) when no ranking exists | the count of accounts safe to dispatch to — never dispatch to an exhausted/blocked one (#3902) |
+| **healthy-token count** | `available` accounts in `.ranking` in the pool directory `tokens_pool::paths::resolve_tokens_dir` resolves for the workspace — per-repo `{workspace}/.loom/tokens/` when it holds `*.token` files, else the shared machine-level pool (#3938) (`capacity::read_ranking` / `token_axis_limit`, unified with the writer in #4344 — pre-#4344 this hardcoded the per-repo path even on a shared-pool host, which could pin the dispatch cap at 0 against an orphaned per-repo `.ranking` indefinitely), falling back to the `*.token` count (`tokens::token_pool_size`) when no ranking exists | the count of accounts safe to dispatch to — never dispatch to an exhausted/blocked one (#3902) |
 | **per-token concurrency** | `LOOM_PER_TOKEN_CONCURRENCY` / `autonomous.perTokenConcurrency`, default **2** (#3947) | how many concurrent sweeps to allow **per healthy account**. A plan limit is a utilization-window token bucket, not a session count, so one healthy account can run several concurrent sessions. Before #3947 the implicit factor was `1` (one sweep per account), which collapsed the whole fleet to cap 1 when 6/7 accounts were at their weekly ceiling even though the single healthy account had ample session-window headroom |
 | **disk headroom** | `floor(free_gb / LOOM_PER_WORKTREE_GB)` on the worktree-root volume (`disk_headroom::disk_headroom_limit`, a Rust port of `disk-headroom.sh` that shells to `df -Pk`) | never provision more worktrees than the scratch volume can hold |
 | **cpu headroom** (#3978, measured-idle signal #4031) | `max(1, floor((logical_cpus × LOOM_CPU_UTILIZATION_TARGET − consumed_cores) / LOOM_EST_CORES_PER_SWEEP))`, where `consumed_cores = logical_cpus × (1 − idle_fraction)` from the measured idle fraction (loadavg fallback) (`cpu_headroom::cpu_headroom_limit`) | never start more concurrent sweeps than the host's CPU headroom can currently absorb |
@@ -1433,6 +1433,30 @@ diagnosis line; at or above the cap it names the binding term as before. The
 `= min(…)` breakdown line is untouched — those genuinely are ceilings. The JSON
 status carries the same `capacity_bound` boolean so scripted consumers aren't
 misled at low occupancy either.
+
+**Honest headline when the daemon's own read disagrees with a fresh probe
+(#4344).** `resolve_capacity` prefers a fresh client-side `loom-tokens check
+--json` probe over the daemon's own ranking read when one succeeds — useful for
+showing current numbers, but that probe's cap is **not** what the running
+daemon actually used to gate dispatch this tick. The pretty-printed `Dynamic
+concurrency cap:` headline and the `= min(healthy N × per-token M …)`
+breakdown always name the daemon's own numbers (`report.dynamic_cap` /
+`report.capacity.token_axis_limit`) — the probe's cap is shown only as a
+labeled secondary `fresh probe suggests: …` line when it differs. The
+`capacity_bound` gate above is likewise computed against the daemon's own cap,
+not the probe's, so "not capacity-bound" can never print while the daemon's
+real (lower) cap is already saturated. When the daemon's own ranking read
+shows **0 healthy accounts** while the probe (or raw pool) disagrees — the
+#4344 incident: dispatch pinned at a token term of `0 × per-token = 0` for
+~40 minutes because the ranking directory it read had diverged from the one
+`loom-tokens check --ranking` / the #4080 self-refresher actually wrote — the
+status view promotes this from the old small-print `note: daemon dispatch cap
+still uses a stale .ranking (...)` line to a headline `⚠ DISPATCH IS
+TOKEN-STARVED: …` line and suppresses "the limiter is work availability"
+underneath it, since the real limiter is unambiguously the token term. The
+root fix for the divergence itself is unifying which `.ranking` file
+[`capacity::read_ranking`] consults — see the healthy-token-count row of the
+input table above.
 
 **Session-limit fault handling (#3947).** Stacking can occasionally trip a
 **concurrent-session-limit** fault on a token (the account is healthy but cannot
--- a/loom-daemon/src/capacity.rs
+++ b/loom-daemon/src/capacity.rs
@@ -157,29 +157,44 @@ impl RankingSnapshot {
 /// files, so callers that have their own resolved directory (or want to probe
 /// a specific one directly, e.g. in tests) can bypass re-resolution.
 ///
-/// The file is the pipe-delimited `name|status` format written by
-/// `loom-tokens check --ranking` (one account per line). Returns `None` when the
-/// file is absent, unreadable, or contains no parseable rows — the signal that
-/// no probe data exists, in which case callers fall back to the raw token-pool
-/// size (byte-for-byte the pre-#3902 behavior). Blank lines and malformed rows
-/// (no `|`) are skipped; a row's status is parsed via [`AccountHealth::parse`].
+/// The file is the pipe-delimited `name|status|5h_util` format written by
+/// `loom-tokens check --ranking` (one account per line, issue #4195), where the
+/// status is the **second** field and the trailing 5h-utilization field is
+/// optional (legacy `name|status` lines omit it). Returns `None` when the file
+/// is absent, unreadable, or contains no parseable rows — the signal that no
+/// probe data exists, in which case callers fall back to the raw token-pool size
+/// (byte-for-byte the pre-#3902 behavior). Blank lines, `#` comments, and
+/// malformed rows (no `|`, so no status field) are skipped; a row's status is
+/// parsed via [`AccountHealth::parse`].
+///
+/// Parsing goes through [`crate::tokens_pool::select::parse_ranking_line`] — the
+/// **same** triple parser the spawn-time selector uses — so this reader and the
+/// selector can never de-sync on the field layout (the #4243/#4344 drift, where
+/// this function took the *last* field as the status and mis-read every 3-field
+/// row's `5h_util` — e.g. `agent3-2amlogic|available|0.09` → `"0.09"` →
+/// [`AccountHealth::Unknown`] — pinning the token axis at 0 on a fully healthy
+/// pool). A malformed no-`|` row is still skipped here (rather than counted with
+/// an empty/unknown status) so an unparseable ranking is treated as absent, not
+/// as a genuine 0-healthy snapshot.
 #[must_use]
 pub fn read_ranking_at(pool_dir: &Path) -> Option<RankingSnapshot> {
     let ranking_path = pool_dir.join(".ranking");
     let contents = std::fs::read_to_string(&ranking_path).ok()?;
     let mut snap = RankingSnapshot::default();
     for line in contents.lines() {
-        let line = line.trim();
-        if line.is_empty() {
+        // A row must carry a `|` (a status field). A no-`|` row is genuinely
+        // malformed and is skipped, not counted as an empty-status account —
+        // preserving the "unparseable ranking ⇒ absent, not 0-healthy"
+        // fallback (see the module doc + `read_ranking_malformed_*` tests).
+        if !line.contains('|') {
             continue;
         }
-        // `name|status` — the status is the last pipe-delimited field so a name
-        // containing a stray pipe (there are none in practice) still parses.
-        let Some((_name, status)) = line.rsplit_once('|') else {
+        let Some((_name, status, _util_5h)) = crate::tokens_pool::select::parse_ranking_line(line)
+        else {
             continue;
         };
         snap.total += 1;
-        match AccountHealth::parse(status) {
+        match AccountHealth::parse(&status) {
             AccountHealth::Available => snap.available += 1,
             AccountHealth::Exhausted => snap.exhausted += 1,
             AccountHealth::RateLimited => snap.rate_limited += 1,
@@ -607,6 +622,155 @@ mod tests {
         assert_eq!(snap.total, 2, "the pipe-less row is skipped");
         assert_eq!(snap.available, 1);
         assert_eq!(snap.exhausted, 1);
+    }
+
+    #[test]
+    fn read_ranking_at_parses_three_field_format() {
+        // #4243/#4344 primary bug: writers emit `name|status|5h_util` whenever
+        // 5h utilization is known (which on the live host is always). The
+        // status is the SECOND field — the reader must NOT take the trailing
+        // util as the status. This is the exact live-host layout from the
+        // #4344 incident: 7 three-field rows, 6 available.
+        let tmp = tempfile::tempdir().unwrap();
+        write_ranking_at(
+            tmp.path(),
+            "agent3-2amlogic|available|0.09\n\
+             robb-2amlogic|available|0.18\n\
+             rjwalters-gmail|available|0.22\n\
+             agent2-2amlogic|available|0.31\n\
+             agent4-2amlogic|available|0.05\n\
+             agent5-2amlogic|available|0.44\n\
+             old-account|exhausted|0.00\n",
+        );
+        let snap = read_ranking_at(tmp.path()).unwrap();
+        assert_eq!(snap.total, 7);
+        assert_eq!(snap.available, 6, "3-field rows: the status is the SECOND field, not the last");
+        assert_eq!(snap.exhausted, 1);
+        assert_eq!(
+            snap.unknown, 0,
+            "the 5h_util field must never be parsed as a status word (the #4243 drift)"
+        );
+    }
+
+    #[test]
+    fn read_ranking_at_curator_fixture_three_field() {
+        // The exact fixture the judge required to pass:
+        //   total=3, available=2, exhausted=1.
+        let tmp = tempfile::tempdir().unwrap();
+        write_ranking_at(
+            tmp.path(),
+            "agent3-2amlogic|available|0.09\n\
+             robb-2amlogic|available|0.18\n\
+             rjwalters-gmail|exhausted|0.00\n",
+        );
+        let snap = read_ranking_at(tmp.path()).unwrap();
+        assert_eq!(snap.total, 3);
+        assert_eq!(snap.available, 2);
+        assert_eq!(snap.exhausted, 1);
+    }
+
+    #[test]
+    fn read_ranking_at_mixed_two_and_three_field() {
+        // A file mixing legacy 2-field lines and new 3-field lines — the state
+        // during a rolling format migration — must parse both correctly.
+        let tmp = tempfile::tempdir().unwrap();
+        write_ranking_at(
+            tmp.path(),
+            "legacy-a|available\n\
+             new-b|available|0.12\n\
+             legacy-c|exhausted\n\
+             new-d|blocked|0.00\n\
+             new-e|rate_limited|0.88\n",
+        );
+        let snap = read_ranking_at(tmp.path()).unwrap();
+        assert_eq!(snap.total, 5);
+        assert_eq!(snap.available, 2);
+        assert_eq!(snap.exhausted, 1);
+        assert_eq!(snap.blocked, 1);
+        assert_eq!(snap.rate_limited, 1);
+        assert_eq!(snap.unknown, 0);
+    }
+
+    #[test]
+    fn read_ranking_at_comment_lines_are_skipped() {
+        // `#` comments (stripped by the shared selector parser) are not rows.
+        let tmp = tempfile::tempdir().unwrap();
+        write_ranking_at(
+            tmp.path(),
+            "# probed 2026-07-29\na|available|0.10\nb|exhausted # near ceiling\n",
+        );
+        let snap = read_ranking_at(tmp.path()).unwrap();
+        assert_eq!(snap.total, 2, "the leading comment line is not a row");
+        assert_eq!(snap.available, 1);
+        assert_eq!(snap.exhausted, 1);
+    }
+
+    #[test]
+    fn ranking_format_drift_conformance_writer_reader_selector_agree() {
+        // Format-drift guard (#4344 AC 3): the single test that would have
+        // caught #4243's partial migration. It binds THREE parties to the same
+        // fixture lines and fails if any one drifts:
+        //   (1) the WRITER — `tokens_pool::check::ranking_line` (what is
+        //       actually written to `.ranking`),
+        //   (2) capacity's READER — `read_ranking_at` (this module),
+        //   (3) the spawn-time SELECTOR's parser —
+        //       `tokens_pool::select::parse_ranking_line`.
+        // If a future format change moves the status field, at least one of
+        // these assertions breaks, so no reader can silently miss the
+        // migration again.
+        use crate::tokens_pool::check::ranking_line;
+        use crate::tokens_pool::select::parse_ranking_line;
+
+        // (name, status, util) fixtures spanning every health class plus the
+        // optional / absent util field.
+        let accounts: [(&str, &str, Option<f64>); 5] = [
+            ("agent3-2amlogic", "available", Some(0.09)),
+            ("robb-2amlogic", "available", None), // legacy 2-field line
+            ("rjwalters-gmail", "exhausted", Some(0.00)),
+            ("agent2", "rate_limited", Some(0.88)),
+            ("agent4", "blocked", None),
+        ];
+
+        // Render the file exactly as the writer would.
+        let body: String = accounts
+            .iter()
+            .map(|(name, status, util)| ranking_line(name, status, *util) + "\n")
+            .collect();
+
+        // (2) vs (3): the selector's parser must recover the same name/status
+        // the writer put in, from the writer's own output.
+        for ((name, status, util), line) in accounts.iter().zip(body.lines()) {
+            let (pname, pstatus, putil) =
+                parse_ranking_line(line).expect("writer output must be parseable by the selector");
+            assert_eq!(&pname, name);
+            assert_eq!(
+                &pstatus, status,
+                "selector parser must recover the writer's status field, not the util"
+            );
+            match util {
+                Some(u) => {
+                    let p = putil.expect("a 3-field line must yield a util");
+                    assert_eq!(format!("{p:.2}"), format!("{u:.2}"));
+                }
+                None => assert_eq!(putil, None, "a 2-field line must yield no util"),
+            }
+        }
+
+        // (1) → (2): capacity's reader must classify that same file into the
+        // writer's intended statuses — 2 available, 1 exhausted, 1
+        // rate_limited, 1 blocked, and crucially 0 unknown.
+        let tmp = tempfile::tempdir().unwrap();
+        write_ranking_at(tmp.path(), &body);
+        let snap = read_ranking_at(tmp.path()).unwrap();
+        assert_eq!(snap.total, 5);
+        assert_eq!(snap.available, 2);
+        assert_eq!(snap.exhausted, 1);
+        assert_eq!(snap.rate_limited, 1);
+        assert_eq!(snap.blocked, 1);
+        assert_eq!(
+            snap.unknown, 0,
+            "no writer-produced line may parse to Unknown — that is the #4243 drift signature"
+        );
     }
 
     // ------------------------------------------------------------------
--- a/loom-daemon/src/capacity.rs
+++ b/loom-daemon/src/capacity.rs
@@ -12,10 +12,13 @@
 //!
 //! 1. **Slow down** — [`token_axis_limit`] backs the token axis off from the
 //!    raw pool size toward the count of *healthy* (`available`) accounts, read
-//!    from the rotation ranking file (`.loom/tokens/.ranking`). When every
-//!    account is exhausted the limit drops to 0 and the finder defers (never
-//!    hammers an exhausted account). A single healthy account is the throughput
-//!    floor, never a halt (operator refinement on #3902).
+//!    from the rotation ranking file (`.ranking`, resolved via
+//!    [`read_ranking`] to whichever pool directory — per-repo or shared
+//!    machine-level (#3938) — actually holds the tokens, kept in lock-step
+//!    with the writer since #4344). When every account is exhausted the limit
+//!    drops to 0 and the finder defers (never hammers an exhausted account). A
+//!    single healthy account is the throughput floor, never a halt (operator
+//!    refinement on #3902).
 //! 2. **Alert** — [`assess_pressure`] derives whether the token axis is the
 //!    binding constraint *and* work is queued behind it, and
 //!    [`CapacityAdvisory::message`] builds an operator advisory naming the
@@ -34,11 +37,14 @@
 //! # Why the ranking file (not a network probe)
 //!
 //! The daemon never performs the slow per-account rate-limit probe itself — that
-//! is `loom-tokens check`'s job, run out-of-band (cron / `probe-tokens.sh`),
-//! which writes the discrete status into `.loom/tokens/.ranking` (the
-//! format-of-record the spawn-time selector already consumes). Reading that file
-//! keeps this module a fast, non-blocking, filesystem-only read that matches the
-//! footing of [`crate::tokens::token_pool_size`] and
+//! is `loom-tokens check`'s job, run out-of-band (cron / `probe-tokens.sh` /
+//! the #4080 self-refresh), which writes the discrete status into
+//! `<resolved-pool-dir>/.ranking` (the format-of-record the spawn-time
+//! selector already consumes). [`read_ranking`] resolves the **same**
+//! directory the writer targets (#3938 per-repo-then-shared precedence,
+//! #4344) so the reader and writer can never diverge onto different files.
+//! Reading that file keeps this module a fast, non-blocking, filesystem-only
+//! read that matches the footing of [`crate::tokens::token_pool_size`] and
 //! [`crate::disk_headroom::disk_headroom_limit`].
 //!
 //! # Near-ceiling granularity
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge Review: CHANGES REQUESTED

**Verdict**: The two halves this PR delivers (ranking-path unification, honest status headline) are well built — clean layering, good regression fixtures, correct `capacity_bound` semantics, CI green, merge state CLEAN. But the PR **misses the primary fix** in #4344's authoritative acceptance criteria, so merging it would close the issue while the live wedge persists.

### The gap (blocking)

The curator's second pass on #4344 (2026-07-29 15:36Z — **after** this PR was built at 06:08Z, so this is a timeline artifact, not a Builder error) confirmed the root cause is the **#4243 ranking-format divergence**, not the path divergence:

- Writers emit 3-field lines `name|status|5h_util` (`tokens_pool/check.rs::ranking_line` ~line 543) whenever 5h utilization is known — which on this host is **always** (live `~/.loom/tokens/.ranking` right now: 7 three-field rows, 6 `available`).
- This PR's `capacity::read_ranking_at` still parses with `rsplit_once('|')` and treats the **last** field as the status (`loom-daemon/src/capacity.rs` ~line 178). On `agent3-2amlogic|available|0.09` that extracts `"0.09"` → `AccountHealth::Unknown` → counted unhealthy.

**Empirically verified on this PR's head (0bb176ce)**: I added a temporary test writing the exact live-host ranking (6 available, 3-field) and called `read_ranking_at` — result: `available=0, unknown=7`, token term `0 × per-token = 0`. The deterministic, restart-proof, freshness-proof starvation this issue exists to fix survives this PR unchanged. Worse, the new `⚠ DISPATCH IS TOKEN-STARVED` headline tells the operator to run `loom-tokens check --ranking` — which rewrites 3-field lines and **re-arms** the wedge.

### Required changes (from #4344's authoritative AC set)

1. **Primary**: parse the 3-field format correctly — status is the *second* field — while staying backward-compatible with legacy 2-field lines. Strongly prefer sharing `tokens_pool/select.rs`'s existing `splitn(3, '|')` triple parser (~line 157) rather than a second hand-rolled split, so the readers cannot de-sync again. Update `read_ranking_at`'s doc comment (still says "`name|status` format").
2. **Regression tests**: (a) 3-field 6-available fixture → `available=6` (fails on this branch today — fixture below); (b) mixed 2-/3-field file.
3. **Format-drift conformance test**: assert `capacity`'s parser and `tokens_pool`'s writer (`ranking_line`) / selector agree on the same fixture lines — the missing test that would have caught #4243's partial migration.
4. **Transition logging** (AC, minor): log the healthy-count change once per transition, not per tick (not present in this PR; no `work_finder.rs` changes).

Fixture that must pass:
```
agent3-2amlogic|available|0.09
robb-2amlogic|available|0.18
rjwalters-gmail|exhausted|0.00
```
→ `total=3, available=2, exhausted=1`.

### Keep as-is

- `read_ranking_at`/`read_ranking` split + `resolve_tokens_dir` unification (secondary AC, correctly done, all 4 call sites covered via the unchanged public signature).
- The `main.rs` status rework: daemon-cap-based `capacity_bound`, headline promotion, work-availability-line suppression, labeled fresh-probe secondary line — all match the ACs and read well.
- Test hygiene (`#[serial]` + pinned `LOOM_SHARED_TOKENS_DIR`) is good.

### Process

- CI: all checks SUCCESS. Merge state: CLEAN against current main (853dc723) — no rebase needed.
- `cargo test -p loom-daemon --lib capacity::` on the PR head: 31/31 pass (they only exercise 2-field fixtures, which is why the gap is invisible to the suite).

Labels: removing `loom:review-requested`, adding `loom:changes-requested` for Doctor.

rjwalters:
## Judge Review: CHANGES REQUESTED

Thorough work on the path-divergence and observability halves — but this PR **does not fix the confirmed primary root cause** of #4344, and as written it would `Closes #4344` while autonomous dispatch stays token-starved on this host. Details against the four root-cause checks:

### 1. Does the reader now parse the 3-field format? ❌ NO (this is the blocker)

The curator's second-pass re-curation and the Champion re-approval established the **confirmed** root cause: PR #4243/#4195 extended `.ranking` to `name|status|5h_util`, the selector (`select.rs`) was taught the new format (`splitn(3,'|')`), but `capacity::read_ranking` was not. This PR carries that bug forward unchanged:

```rust
// read_ranking_at (this PR)
let Some((_name, status)) = line.rsplit_once('|') else { continue };
match AccountHealth::parse(status) { ... }
```

`rsplit_once('|')` takes the **last** field as the status. On the live 3-field ranking on this host right now:

```
agent3-2amlogic|available|0.09   → status parsed = "0.09" → AccountHealth::parse("0.09") = Unknown → unhealthy
```

All 7 rows (6 `available`) parse to `Unknown` → `snap.available == 0` → `token_axis_limit == 0`. **The wedge is not cleared.** The path unification is real but secondary: both `~/.loom/tokens/.ranking` and the per-repo copy are 3-field and both parse to 0 healthy, so unifying which file is read does not help. I verified the on-disk format and `AccountHealth::parse` directly.

**Minimal fix:** in `read_ranking_at`, take the *second* field of `splitn(3, '|')` (name, status, optional util) instead of the last field — and ideally call the selector's existing 3-field-aware parser so there is one shared implementation (see check 3).

### 2. Status misdiagnosis fixed? ✅ YES (well done)

The honest-headline rework is correct and well-reasoned: headline/breakdown now use `report.dynamic_cap` / `report.capacity.token_axis_limit` (the number dispatch actually gated on), `capacity_bound` is computed against the daemon's own cap, the `⚠ DISPATCH IS TOKEN-STARVED` promotion fires on `daemon-healthy==0 && probe-healthy>0`, and "the limiter is work availability" is suppressed underneath it. In the incident the `loom-tokens check --json` probe (structured data, not the `.ranking` parse) shows 6 healthy, so the headline fires as intended. This half is good — but it is observability only; dispatch still stays wedged because of check 1.

### 3. Are ALL ranking readers on one parser? ❌ NO — this is the exact #4344 lesson

The PR unifies the **four capacity call sites** onto one *resolution* (good), but capacity's *parser* still diverges from the selector's:
- `capacity::read_ranking_at` → `rsplit_once('|')`, last field (broken for 3-field)
- `tokens_pool::select` → `splitn(3, '|')`, correct `(name, status, util_5h)`

#4344 is precisely a "reader-audit miss on a format change." Shipping the path unification while leaving two independent, drifting `.ranking` parsers perpetuates the class of bug. Please converge capacity's read onto the selector's parsing helper (single 3-field-aware parser shared by both).

### 4. Regression test proves recovery? ❌ Insufficient — would have green-lit the incident

`dispatch_resumes_after_ranking_refresh_without_restart` and the shared-pool fixture use **2-field** rows (`a|available\nb|available\nc|exhausted`). None of the new tests use the 3-field `name|available|0.09` production format, so the suite passes while the live bug persists. The curator's sharpened AC explicitly asked for "a format-conformance test that would have caught #4243's partial migration." Please add a test asserting a 3-field `name|available|0.09` ranking yields `available == N` (and a mixed legacy 2-field + 3-field file parses both).

### Closes-vs-partial decision

Because the primary parser fix is missing, this must **not** `Closes #4344`. Two acceptable paths:
- **Preferred:** add the ~2-line parser fix + the 3-field conformance test to this PR (the fix lands in `read_ranking_at`, a function this PR already fully rewrites — cheap and contained), keeping `Closes #4344`.
- Or downscope to `Part of #4344` (the #4318 precedent) and land only the path-unification + observability halves, leaving the parser fix for the follow-up build the Champion already sequenced.

Given a **live production wedge** and that the fix is a tiny addition to code already being rewritten here, the preferred path is to complete it in this PR.

### Merge / CI / interaction

- Mergeable against current `main` (`853dc723`, post-#4305/#4320): `MERGEABLE` / `CLEAN`, merge-tree shows no conflict.
- CI on PR head is green (Rust Unit Tests pass) — but the suite is green *because* it never exercises the 3-field format (check 4).
- **#4372 interaction note for the coordinator** (not reviewing #4372): it adds registry-anchored startup fallback in the adjacent `resolve_tokens_dir` pool-resolution area. This PR routes `read_ranking` through `resolve_tokens_dir`, so whichever lands second should re-verify reader and writer still resolve the same directory. No blocking conflict today; just an ordering re-check.

Transitioning `loom:review-requested` → `loom:changes-requested`.

rjwalters:
## Doctor: feedback addressed

Completed the Judge's **preferred path** (kept `Closes #4344`) — the primary root cause is now fixed, not just the path-unification/observability halves.

**Parser convergence (single source of truth).** Extracted the selector's field parser as the shared `pub(crate) tokens_pool::select::parse_ranking_line` and routed `capacity::read_ranking_at` through it. Status is now the **second** field of `splitn(3, '|')`; the optional third `5h_util` field is tolerated; legacy 2-field lines still parse. This retires the divergent `rsplit_once('|')` read that took the *last* field as the status — on the live 3-field format (`agent3-2amlogic|available|0.09`) that parsed `"0.09"` → `AccountHealth::Unknown` → 0 healthy → `token_axis_limit == 0` → dispatch token-starved on a fully healthy pool. The two readers can no longer de-sync on field positions (the #4344 "all readers on one parser" lesson).

**Regression tests (would have caught #4243's partial migration).**
- `read_ranking_at_parses_three_field_format` — the exact live-host layout (7 three-field rows, 6 available) asserting `available == 6`, `unknown == 0`.
- `read_ranking_at_mixed_two_and_three_field` — legacy 2-field + new 3-field rows in one file (rolling-migration state) parse correctly.
- `read_ranking_at_comment_lines_are_skipped`, `read_ranking_at_curator_fixture_three_field`.
- `ranking_format_drift_conformance_writer_reader_selector_agree` — binds the **writer** (`check::ranking_line`), capacity's **reader**, and the **selector** parser to one fixture; any future field-position drift breaks at least one assertion.

**Call-site sanity check.** The four unified capacity call sites (`ipc.rs` x2, `work_finder.rs` x2) consume the unchanged `RankingSnapshot` API and now receive correct counts.

**Tests:** `cargo test capacity` → 46 passed / 0 failed (incl. all new 3-field tests). `cargo test --lib tokens_pool` → 178 passed; `select` → 28 passed.

Transitioning `loom:changes-requested` → `loom:review-requested`.

rjwalters:
## Doctor: Judge feedback addressed

All four required changes from the Judge review (comment 5120335479) are now on the branch, green, and pushed.

**1. Share the selector's triple parser (no second hand-rolled parser)** — `f7c7cf86`
- Extracted `tokens_pool::select::parse_ranking_line` as the single `pub(crate)` `splitn(3, '|')` triple parser; both `select::read_ranking` and `capacity::read_ranking_at` now route through it. Status is the **second** field, the optional `5h_util` third field is tolerated, and legacy 2-field lines still parse. The two readers can no longer de-sync on field positions (the #4243/#4344 drift). Doc comment updated off the stale "`name|status` format" wording.

**2. 3-field + mixed-format regression tests** — `f7c7cf86`
- `read_ranking_at_parses_three_field_format` (the live-host 7-row/6-available layout), `read_ranking_at_curator_fixture_three_field` (the required `total=3, available=2, exhausted=1` fixture), `read_ranking_at_mixed_two_and_three_field`, and `read_ranking_at_comment_lines_are_skipped`.

**3. Format-drift conformance test** — `f7c7cf86`
- `ranking_format_drift_conformance_writer_reader_selector_agree` binds the **writer** (`check::ranking_line`), capacity's **reader** (`read_ranking_at`), and the **selector** parser (`parse_ranking_line`) to one fixture, asserting no writer-produced line ever parses to `Unknown` — the guard that would have caught #4243's partial migration.

**4. Healthy-account transition logging** — `b5c6d736`
- New `log_healthy_token_transition` logs the `available`-account count once per **change** (mirroring the `was_max_concurrent` / `was_pressured` dedup), wired into both work-finder tick loops. The drop-to-0 edge is a `warn!` naming the recovery lever; the first observation seeds silently. Unit-tested.

### Verification
- `cargo test -p loom-daemon --lib capacity::` → 36/36; affected modules (`work_finder`, `tokens_pool::select`, `tokens_pool::check`, `capacity`) → 205/205; `cargo test --bins` → 30/30.
- `cargo fmt --check` and `cargo clippy --all-targets --all-features` → clean.
- Full `cargo test --lib`: 1563 passed, with a single unrelated flake (`sweep_registry::tests::read_path_reap_is_bounded_when_gh_wedges`, a gh-wedge timing bound) that **passes in isolation** — a known cross-test contention artifact on this shared dogfooding host, untouched by this PR.

Returning to `loom:review-requested`.

rjwalters:
## Judge Re-Review: APPROVED

The Doctor's additive commit (`f7c7cf86`) resolves the blocker from my prior review. Re-verified against all four root-cause checks:

1. **3-field parse — FIXED.** The selector's field parser is now `pub(crate) tokens_pool::select::parse_ranking_line` (`splitn(3,'|')` → `(name, status=field 2, util=optional field 3)`; legacy 2-field yields `util=None`), and `capacity::read_ranking_at` consumes it directly. Confirmed status = second field, trailing `5h_util` is ignored for health, and the no-`|` guard is preserved so an unparseable ranking is still treated as absent (fallback to raw pool), not a false 0-healthy.

2. **One parser, one source of truth — FIXED (the #4344 lesson).** Both capacity's reader and the spawn-time selector now route through the single `parse_ranking_line`; the two can no longer de-sync on field positions.

3. **Conformance test binds all three surfaces.** `ranking_format_drift_conformance_writer_reader_selector_agree` renders the fixture with the real writer (`tokens_pool::check::ranking_line`), asserts the selector parser recovers the writer's status, and asserts capacity's reader classifies the same file with **0 unknown** — so any future field-position drift fails a test. Plus `read_ranking_at_parses_three_field_format` uses the exact live-host 7-row/6-available layout, and `read_ranking_at_mixed_two_and_three_field` covers legacy compatibility.

4. **Live-format re-check.** Applying the field-2 parse to the actual on-disk rankings (`~/.loom/tokens/.ranking` and the per-repo copy, both 3-field) now yields **healthy=6 / total=7** through the capacity path — the wedge is cleared. Local `cargo test capacity` / `tokens_pool::select` green; the observability half (honest `DISPATCH IS TOKEN-STARVED` headline, dispatch-cap-based `capacity_bound`) from the original commit is unchanged and remains correct.

**CI:** Rust Unit Tests **success** on head `f7c7cf86` (required gate met); all other checks success/skipped. No #4385 fake-gh flake family appeared. **Mergeable / CLEAN** vs current `main` (`853dc723`).

`Closes #4344` correctly retained — the PR now delivers all three halves (3-field parser fix + path unification + honest status headline). **#4372 interaction note** stands for the coordinator: it touches the adjacent `resolve_tokens_dir` startup path, so whichever of #4363/#4372 merges second should re-verify reader/writer resolve the same directory (no conflict today).

Transitioning `loom:review-requested` → `loom:pr`.

rjwalters:
## Judge Re-Review: APPROVED

All four findings from my previous review (comment 5120335479) are resolved on head `b5c6d736`:

1. **Primary fix (the live wedge) — VERIFIED FIXED.** `capacity::read_ranking_at` now parses via the shared `tokens_pool::select::parse_ranking_line` triple parser (commit f7c7cf86); status is correctly the second field. **Empirical re-run of my live-host repro**: the byte-for-byte current `~/.loom/tokens/.ranking` (7 three-field rows) now parses as `total=7, available=6, unknown=0` — previously `available=0, unknown=7`. The deterministic token-axis wedge is closed; once this rolls, the machine-level daemon's autonomous dispatch un-starves with no restart and no ranking rewrite (per-tick re-read).
2. **3-field + mixed-format regression tests** — present, including the exact fixture I required (`read_ranking_at_curator_fixture_three_field`: total=3, available=2, exhausted=1) and the incident-layout test asserting `unknown=0`.
3. **Format-drift conformance test** — `ranking_format_drift_conformance_writer_reader_selector_agree` binds writer (`check::ranking_line`) → capacity reader → selector parser on shared fixtures spanning every health class plus optional util. This is the test that would have caught #4243's partial migration.
4. **Transition logging** (commit b5c6d736) — `log_healthy_token_transition` wired into both work-finder tick loops, once-per-change dedup (silent seed, `warn!` on →0, `info!` otherwise), matching the existing `was_max_concurrent`/`was_pressured` discipline, with a state-machine unit test.

Bonus correctness: the no-`|` skip guard preserves the "unparseable ranking ⇒ absent ⇒ raw-pool fallback" policy (not a fake 0-healthy snapshot), and `#`-comment handling now matches the writer.

### Validation

- CI on `b5c6d736`: **all checks pass** (Rust Unit Tests, Lint & Build, Installer Integration, CLAUDE.md budget).
- Local (PR worktree): `capacity::` 36/36, `work_finder::` 111/111, `--bins` unaffected; `cargo fmt --check` and `cargo clippy --all-targets` clean.
- One local failure, `tokens_pool::bootstrap::tests::bootstrap_no_source_errors`, is the documented pre-existing host flake (real `~/.claude-monitor/accounts.env` leaking under parallel execution): passes in isolation, `bootstrap.rs` untouched by this PR, and the same test passes in the clean CI run.
- Merge state: GitHub is reporting `UNKNOWN` (backend recompute lag, ~10 min, survives REST nudges). Ground truth verified locally: `git merge-tree` of head against current `origin/main` (853dc723) is conflict-free. `merge-pr.sh --auto` will re-verify at merge time.

All authoritative acceptance criteria from #4344's second-pass curation are now met (primary format fix with shared parser, secondary path unification, status honesty, regression + conformance tests, transition logging).

Labels: removing `loom:review-requested`, adding `loom:pr` for Champion auto-merge.

PR #4374, opened 2026-07-29
fix(daemon): recover a detached sweep killed by external SIGKILL via periodic claim reconciliation

### LINKED ISSUE #4348: daemon: detached sweep killed by external SIGKILL is never recovered
## Problem

A detached sweep that dies from an **external SIGKILL** is never recovered — no requeue, no claim release, no operator-visible alert. The issue it was working stays in limbo until a human notices.

Observed 2026-07-28: the Claude Code harness performed a model switch that SIGKILLed the session's detached child processes. A manually spawned detached sweep died permanently; nothing detected the dead pid or recovered the work. (The daemon process itself survived the same event — this is specifically about sweep child processes / detached sweeps.)

## Expected

The reaper (or an equivalent liveness pass) should notice a sweep whose pid is gone without a recorded exit, then:

1. release/clear the issue claim and any `.loom/locks` entries it held,
2. either requeue the issue for dispatch or surface it clearly in `loom-daemon status` as died-unexpectedly,
3. log the event at WARN with the sweep id, issue number, and last-known pid.

## Notes

- Repro is external-kill shaped: `kill -9` of a running sweep's process tree should be a sufficient stand-in for the harness model-switch behavior.
- Related context: the harness model-switch kill is an environmental hazard on this host (Claude safety-classifier model switches), so recovery needs to be automatic — the operator may not be present when it happens.

---

### Root cause — verified against `origin/main`

Three recovery layers exist today; **none covers the observed case** (a manually/externally spawned detached sweep SIGKILLed while the daemon keeps running):

| Layer | When it runs | Covers | Why it misses this case |
|---|---|---|---|
| `SweepRegistry` reaper (`loom-daemon/src/sweep_registry.rs`, 30s tick) | continuously | **daemon-dispatched** sweeps — registry entries with a retained child handle (`try_wait` + pid poll); releases the lock and restores the label on a dead pid | a manually spawned sweep never enters the in-memory registry, so the reaper is structurally blind to it |
| `claim_reconciliation` startup pass (`loom-daemon/src/main.rs` ~line 861) | **daemon startup only** | every workspace's open `loom:building` issues; evidence = `~/.loom/sweeps.json` journal pid, else label-age ≥ 4h (`DEFAULT_STALE_BUILDING_HOURS`) | never fires while the daemon stays up (the daemon survived the incident); and a manual sweep writes **no journal entry** (`sweep_journal.rs` module docs, "Who writes it": only `SweepRegistry::dispatch`), so even a restart would wait out the 4h no-record grace |
| `loom-recover-orphans --recover` (Python, manual) | operator-invoked | same journal + label-age evidence | requires the operator to be present — exactly what this issue's Notes say cannot be assumed |

**The evidence needed for immediate recovery already exists on disk**: the in-session `/loom:sweep` path registers `{run_id, pid, timestamp}` at `.loom/sweep-run/<RUN_ID>.json` (`defaults/scripts/sweep-run-registry.sh new`; the pid is the long-lived sweep process, i.e. the detached `claude -p` in this incident), and the sweep's checkpoint `.loom/sweep-checkpoint/issue-<N>.json` carries `task_id` = that same run id. No recovery layer currently performs this join.

Additional verified facts that bound the scope:

- **Locks**: `.loom/locks/issue-<N>/` is written *exclusively* by daemon `dispatch` (`sweep_registry.rs` module docs) — the incident's manually spawned sweep held no lock, so "release its locks" is only relevant to the daemon-dispatched variant, whose reaper already calls `release_lock` on a dead pid. However, `acquire_lock` hard-fails on `AlreadyExists` **without checking owner liveness**, so any stale lock that escapes the reaper blocks re-dispatch until restart — a defensive stale-lock prune belongs in the periodic pass.
- **Requeue safety**: reclaim = flip `loom:building` → `loom:issue` (`claim_reconciliation::forge::reclaim`). The #4123 open-PR dispatch guard already prevents re-dispatching an issue with an open linked PR, so periodic reclaim composes safely with post-Builder deaths (stranded-PR recovery was #4256–#4258's scope, all closed).
- Old checkpoint files persist after sweeps finish (harmless here: the pass keys off open `loom:building` issues, never off checkpoints alone).

### Recommended approach — promote the startup reconciliation to a periodic pass, with a manual-sweep evidence source

1. **Periodic pass**: extract the startup reconciliation block in `main.rs` (~lines 861–891) into a reusable function and also run it on an interval (default 10–15 min; env override, e.g. `LOOM_CLAIM_RECONCILE_INTERVAL_SECS`, with a sane floor). Reuse the existing `LOOM_STALE_CLAIM_RECONCILE` kill switch and `MAX_ISSUES_PER_WORKSPACE` bound so the pass can never become a `gh` API storm.
2. **New evidence source** in `claim_reconciliation::decide`/`plan`: for a `loom:building` issue with **no journal entry**, join checkpoint `task_id` → `.loom/sweep-run/<task_id>.json`:
   - pid **alive** ⇒ `Keep` (an in-session/interactive sweep is running — never reclaim a live claim);
   - pid **dead** ⇒ `Reclaim` with a new `ReclaimReason` variant (e.g. `DeadRunRegistry { pid }`) **immediately**, no 4h grace — the death is provable;
   - run entry **absent** (pruned by a later sweep's `prune_dead`, or pre-#3768) ⇒ existing `NoRecordStale` age rule, unchanged.
3. **Operator visibility**: each reclaim logs at WARN with issue number, last-known pid, and run id when known; best-effort event-bus emission mirroring the reaper's existing event pattern.
4. **Defensive stale-lock prune** in the same pass: remove `.loom/locks/issue-<N>/` whose `owner.json` pid is dead and which has no live registry entry (runtime analogue of `reconstruct()`'s startup cleanup).

**Alternatives considered**:
- *Manual sweep path writes `~/.loom/sweeps.json` too* — would make manual sweeps first-class in the existing evidence chain, but means a bash writer for a versioned, Rust/Python-read journal file; riskier, defer to a follow-up if the run-registry join proves insufficient.
- *Surface-only (`loom-daemon status` shows died-unexpectedly, no requeue)* — does not restore autonomy; the reclaim **is** the requeue, and the WARN + event covers visibility. Rejected as the primary mechanism.

### Acceptance Criteria

- [ ] With the daemon running (no restart), `kill -9` of a manually spawned detached sweep's process tree results — within one reconciliation interval — in the issue's `loom:building` label being replaced by `loom:issue`, with no operator action.
- [ ] Each such reclaim logs at WARN with the issue number, last-known pid, and run id (when known).
- [ ] A `loom:building` issue whose manual sweep is still **alive** (live pid in `.loom/sweep-run/`) is kept — the periodic pass never reclaims a live claim.
- [ ] Journal-backed (daemon-dispatched) claims behave exactly as the startup pass does today: dead journal pid ⇒ reclaim, live ⇒ keep, no-evidence ⇒ 4h grace. No behavior change to the startup pass itself beyond code reuse.
- [ ] Regression: a daemon-dispatched sweep killed by SIGKILL is still recovered by the existing reaper path (label restored, lock released) — covered by a test.
- [ ] The pass is bounded and configurable: interval env var with floor, existing kill switch honored, `MAX_ISSUES_PER_WORKSPACE` respected; documented in the module docs.

### Affected Files

- `loom-daemon/src/claim_reconciliation.rs` — new evidence source (checkpoint→run-registry join), new `ReclaimReason` variant, periodic entry point, unit tests
- `loom-daemon/src/main.rs` — spawn the interval task alongside the existing startup pass (~line 861)
- `loom-daemon/src/sweep_registry.rs` — (small, optional) expose a stale-lock prune helper for the pass; no reaper changes
- `loom-daemon/src/sweep_journal.rs` — reference only (evidence-source precedence)
- `defaults/scripts/sweep-run-registry.sh` — reference only (schema consumed; do not change)

### Test Plan

- [ ] Unit (pure `decide`/`plan`, no forge deps): dead run-registry pid ⇒ `Reclaim(DeadRunRegistry)`; live pid ⇒ `Keep`; missing run entry ⇒ existing age rule; malformed checkpoint/run-registry JSON degrades fail-safe to the age rule (never a spurious reclaim).
- [ ] Unit: interval env parsing incl. floor enforcement and kill-switch interaction.
- [ ] Integration (temp dirs + `LOOM_SWEEPS_JOURNAL_PATH` seam): fabricated workspace with a checkpoint + run-registry entry for a dead pid; assert the plan output. (The `gh` forge glue stays a thin untested adapter, matching the module's existing test philosophy.)
- [ ] Manual verification: with the daemon running, spawn a detached `claude -p "/loom:sweep <test-issue>"`, wait for `loom:building`, `kill -9` the process tree, and verify within one interval the label flips back and the WARN line appears in `daemon.log`.
- [ ] Edge cases: pid reuse (document as a known limitation — same semantics as the existing journal liveness check); two workspaces reconciled independently; pass overlapping a daemon shutdown is harmless.

### Scope guard

No Python `loom_tools` changes; no `loom-daemon status` UI additions; no bash-side journal writes (explicit follow-up candidate); the pass's only forge mutations are the reclaim label flips. Target ≤ ~400 LOC including tests — if the stale-lock prune (item 4) pushes past that, split it out as a follow-up issue rather than growing the PR.

rjwalters:
## Problem

A detached sweep that dies from an external SIGKILL is never recovered while the daemon keeps running — no requeue, no claim release, no operator-visible alert. Three existing recovery layers (the in-memory `SweepRegistry` reaper, the daemon-*startup* `claim_reconciliation` pass, and the manual Python `loom-recover-orphans`) all miss this exact case: a manually/externally spawned detached sweep never enters the in-memory registry, never writes a journal entry (only `SweepRegistry::dispatch` does), and the startup pass never re-fires because the daemon itself never restarted.

## What changed

- **Promoted the startup-only reconciliation pass to also run on an interval.** `claim_reconciliation::run_reconciliation_pass` is now the single implementation both the daemon-startup call site (`main.rs`) and a new periodic `tokio::spawn` interval task (`spawn_periodic_reconciliation_task`) invoke — identical behavior at both call sites by construction. Default interval 10 minutes, override via `LOOM_CLAIM_RECONCILE_INTERVAL_SECS`, floored at 60s so a typo can't turn it into a `gh`-API busy loop. Governed by the same `LOOM_STALE_CLAIM_RECONCILE` kill switch and bounded by the same `MAX_ISSUES_PER_WORKSPACE` as before.
- **New evidence source for manually spawned sweeps**: for a `loom:building` issue with no journal entry, `claim_reconciliation::resolve_run_registry_pid` joins the sweep's checkpoint `task_id` (`.loom/sweep-checkpoint/issue-<N>.json`) to its liveness registration (`.loom/sweep-run/<task_id>.json`, written by `sweep-run-registry.sh new`). A dead pid reclaims immediately via a new `ReclaimReason::DeadRunRegistry { pid }` (no age grace, since the death is provable — same posture as the existing journal `DeadPid` branch); a live pid is kept; any failure in the join (missing/malformed checkpoint or run-registry entry) degrades fail-safe to the existing age-based `NoRecordStale` rule, never a spurious reclaim.
- **Operator visibility**: each reclaim now logs at WARN (previously `info`) with the issue number, last-known pid, and run id (when the evidence source was the run-registry join).

## Deviations from the curated spec

- **Deferred the defensive stale-lock prune** (`Recommended approach` item 4 / `sweep_registry.rs` in "Affected Files"). It addresses a different failure mode (blocked *re-dispatch* via `acquire_lock`'s `AlreadyExists` hard-fail, not a stuck forge label) than the one this issue reports, isn't covered by any of the acceptance criteria, and the issue's own scope guard pre-authorizes deferring it rather than growing the PR further ("if the stale-lock prune ... pushes past that, split it out as a follow-up issue"). Noted in the module doc comment as an explicit follow-up candidate.
- **No event-bus emission.** The "Recommended approach" mentioned a best-effort event-bus emission mirroring the reaper; this isn't in the acceptance criteria (only the WARN-log requirement is), and there's no existing global `EventBus` accessor this module could reuse without adding new cross-cutting infrastructure. Trimmed to keep the change focused on the acceptance-criteria-driving core.

## Test plan

- Unit tests for `decide`/`plan`: dead run-registry pid ⇒ `Reclaim(DeadRunRegistry)`; live pid ⇒ `Keep`; journal entry present ⇒ takes priority over run-registry evidence; missing run-registry evidence ⇒ falls through to the existing age rule; malformed checkpoint/run-registry JSON degrades fail-safe (never a spurious reclaim).
- Unit tests for interval env parsing (default, override, floor enforcement, non-positive/unparseable fallback) and kill-switch interaction (`run_reconciliation_pass` no-ops with no `gh` on `PATH` at all when disabled).
- Integration tests (temp dirs + the `LOOM_SWEEPS_JOURNAL_PATH` seam + a fake `gh` script) via `forge::reconcile_workspace`: a fabricated dead-pid run-registry scenario reclaims even with a fresh label; a live-pid scenario is kept; a malformed checkpoint falls back to the age rule.
- Existing regression test for #3975 (dead journal pid reclaims immediately even with a fresh label) is unchanged and still passes — the daemon-dispatched/journal-backed path has zero behavior change.
- `cargo test --lib` (loom-daemon): 1626 passed, 0 failed. `cargo clippy --package loom-daemon --all-targets` with the CI lint flags: clean.
- Note: `cargo test --test integration_basic` (real-daemon-socket tests) fails in this sandbox on both `main` and this branch identically ("Daemon failed to create socket within 5s") — a pre-existing environment limitation unrelated to this change, confirmed via `git stash`.

Closes #4348

### DIFF
--- a/loom-daemon/src/claim_reconciliation.rs
+++ b/loom-daemon/src/claim_reconciliation.rs
@@ -1,38 +1,89 @@
-//! Daemon-startup reconciliation of stale `loom:building` claims across every
-//! managed workspace (Issue #3953, acceptance criterion 3/4).
+//! Reconciliation of stale `loom:building` claims across every managed
+//! workspace (Issue #3953, acceptance criterion 3/4; promoted from a
+//! startup-only pass to a periodic one by Issue #4348).
 //!
 //! The persisted [`crate::sweep_journal`] gives a fresh daemon (post-restart,
 //! post-rate-limit-kill, post-upgrade) an authoritative liveness source that
 //! survives the in-memory [`crate::sweep_registry::SweepRegistry`] being wiped
-//! clean. This module is the consumer that turns that evidence into action at
-//! startup: for every registered workspace, list the open `loom:building`
-//! issues and decide, per issue, whether the claim is still backed by a live
-//! sweep.
+//! clean. This module is the consumer that turns that evidence into action:
+//! for every registered workspace, list the open `loom:building` issues and
+//! decide, per issue, whether the claim is still backed by a live sweep.
 //!
 //! ## Decision rule
 //!
 //! - A journal entry recording a **live** PID ⇒ [`ReconcileAction::Keep`] —
 //!   the claim is genuinely in-flight.
 //! - A journal entry recording a **dead** PID ⇒
-//!   [`ReconcileAction::Reclaim`] — the sweep behind this claim is provably
-//!   gone.
-//! - **No** journal entry at all (the claim predates this journal, or was
-//!   made by a process that never wrote one) ⇒ reclaim only once the label
-//!   has been stale longer than [`resolve_stale_hours`] (`updated_at` age),
-//!   otherwise `Keep`. This mirrors the Python tool's label-age grace period
-//!   philosophy (#3651): absence of evidence is not, by itself, proof of
-//!   orphanhood — only *aged* absence is.
+//!   [`ReconcileAction::Reclaim`]`(`[`ReclaimReason::DeadPid`]`)` — the sweep
+//!   behind this claim is provably gone.
+//! - **No** journal entry (a manually/externally spawned `/loom:sweep` never
+//!   writes one — only [`crate::sweep_registry::SweepRegistry::dispatch`]
+//!   does), but the checkpoint→run-registry join described below resolves a
+//!   pid ⇒ same live/dead split, reclaiming with
+//!   [`ReclaimReason::DeadRunRegistry`] on a dead pid (Issue #4348 — this is
+//!   the evidence source that recovers a detached sweep killed by an
+//!   external `SIGKILL` while the daemon itself stays up).
+//! - **No** evidence at all ⇒ reclaim only once the label has been stale
+//!   longer than [`resolve_stale_hours`] (`updated_at` age), otherwise `Keep`
+//!   ([`ReclaimReason::NoRecordStale`]). This mirrors the Python tool's
+//!   label-age grace period philosophy (#3651): absence of evidence is not,
+//!   by itself, proof of orphanhood — only *aged* absence is.
 //! - No age evidence either (an issue whose `updatedAt` is somehow
 //!   unavailable) ⇒ fail safe to `Keep`.
 //!
+//! ## Checkpoint → run-registry join (Issue #4348)
+//!
+//! The in-session `/loom:sweep` path (unlike a daemon dispatch) writes no
+//! journal entry, but it DOES register `{run_id, pid, timestamp}` at
+//! `.loom/sweep-run/<RUN_ID>.json` (`defaults/scripts/sweep-run-registry.sh
+//! new`), and its checkpoint (`.loom/sweep-checkpoint/issue-<N>.json`) carries
+//! `task_id` = that same run id. [`resolve_run_registry_pid`] joins the two so
+//! a `loom:building` issue with no journal entry still gets an immediate,
+//! provable liveness answer instead of waiting out the age-based grace
+//! period. Any failure in that join (missing checkpoint, missing/garbled
+//! `task_id`, missing/garbled run-registry entry) degrades to `None` —
+//! [`decide`] then falls through to the age rule, never treating a failed
+//! join as proof of death.
+//!
+//! ## Periodic pass (Issue #4348)
+//!
+//! [`run_reconciliation_pass`] is the single entry point both the
+//! daemon-startup call site (`main.rs`) and
+//! [`spawn_periodic_reconciliation_task`]'s interval loop invoke — identical
+//! behavior at both call sites, gated by the same [`reconciliation_enabled`]
+//! kill switch and bounded by the same [`MAX_ISSUES_PER_WORKSPACE`]. Running
+//! this on an interval (not just at startup) is what recovers a detached
+//! sweep an external `SIGKILL` (or a harness model-switch, the incident that
+//! motivated #4348) killed while the daemon kept running — the startup pass
+//! alone can never observe that, because the daemon never restarted.
+//!
+//! ## Defensive stale-lock prune — deferred (Issue #4348)
+//!
+//! The issue's "Recommended approach" also floated a defensive prune of any
+//! `.loom/locks/issue-<N>/` whose `owner.json` records a dead pid (the
+//! runtime analogue of [`crate::sweep_registry::SweepRegistry::reconstruct`]'s
+//! startup lock cleanup — `acquire_lock` hard-fails on `AlreadyExists`
+//! without ever checking owner liveness). It is deliberately **out of scope**
+//! for this change: it addresses a different failure mode (blocked
+//! *re-dispatch*, not a stuck forge label) than the one #4348 reports, is not
+//! covered by any of #4348's acceptance criteria, and the issue's own scope
+//! guard pre-authorizes deferring it rather than growing this PR further —
+//! see the issue's "Affected Files" note on `sweep_registry.rs`. Tracked as a
+//! follow-up rather than bundled here.
+//!
+//! ## Testability
+//!
 //! The pure [`decide`] / [`plan`] functions take no forge/PID dependencies —
-//! they are fully unit-testable. The `gh`/label-flip glue lives in
-//! [`forge::reconcile_workspace`], which mirrors the `WorkSource`/adapter split
-//! established by [`crate::work_finder::forge`] and
-//! [`crate::epic_supervisor::forge`] (those adapters are likewise not
+//! they are fully unit-testable (the run-registry lookup is injected into
+//! [`plan`] as a closure, exactly like `is_alive`). The `gh`/label-flip glue
+//! lives in [`forge::reconcile_workspace`], which mirrors the
+//! `WorkSource`/adapter split established by [`crate::work_finder::forge`]
+//! and [`crate::epic_supervisor::forge`] (those adapters are likewise not
 //! unit-tested directly — they are thin `Command` wrappers around `gh`).
 
 use chrono::{DateTime, Utc};
+use std::path::Path;
+use std::time::Duration;
 
 use crate::sweep_journal::{self, JournalEntry, SweepJournal};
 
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

## Review summary

Reviewed against issue #4348's curated acceptance criteria, scope guard, and test-plan checklist.

**Correctness of the new evidence source:**
- The checkpoint→run-registry join (`resolve_run_registry_pid` → `read_checkpoint_task_id` + `read_run_registry_pid`) is defensively coded: every failure mode (missing file, unreadable, malformed JSON, missing/wrong-typed field) degrades to `None` via `?`/`.ok()` chains, never a panic or a fabricated pid. `decide()` only consults `run_registry_pid` when `journal_entry` is absent, and correctly falls through to the existing age-based `NoRecordStale` rule on a failed join — verified this priority order in `decide_journal_entry_takes_priority_over_run_registry_pid` and `plan_consults_run_registry_only_when_journal_entry_absent`, both of which construct an adversarial closure to prove the wrong source is never consulted.
- `DeadRunRegistry` correctly reclaims with no age grace (provable death, same posture as `DeadPid`), and `is_pid_alive(0)` (`sweep_registry.rs`) already returns `false`, so a `pid: 0` run-registry entry used in the integration test is a legitimate dead-pid fixture, not a footgun.
- The `main.rs` refactor is a clean extract-and-reuse: `run_reconciliation_pass` is now the single implementation for both the startup call and `spawn_periodic_reconciliation_task`'s interval loop, so the two call sites can't drift. The interval task correctly consumes its first immediate tick (since the startup pass already just ran) and runs subsequent passes on a blocking thread, matching the `token_ranking_refresh` precedent.

**Test coverage:** Excellent — pure `decide`/`plan` unit tests for both evidence-priority ordering and fail-safe degradation, env-parsing tests for the interval (default/override/floor/non-positive/unparseable), a kill-switch no-op test with no `gh` on `PATH` at all (proving the gate is checked before any subprocess spawn), and three `forge::reconcile_workspace` integration tests (dead pid reclaims, live pid keeps, malformed checkpoint falls back to age rule) using a fake `gh` script + temp dirs, consistent with the module's existing test philosophy.

**Verification performed:**
- `cargo test --package loom-daemon --lib claim_reconciliation`: 30/30 pass.
- `cargo test --package loom-daemon --lib` (full suite): 1626 passed, 0 failed — matches the PR description exactly.
- `cargo clippy --package loom-daemon --all-targets`: clean.
- `gh pr checks 4374`: all required checks pass (Rust Linting & Build, Rust Unit Tests, CLAUDE.md Line Budget).
- `gh pr view --json mergeStateStatus`: `CLEAN`.
- Diff scoped to exactly `loom-daemon/src/claim_reconciliation.rs` + `loom-daemon/src/main.rs` — confirmed no Python `loom_tools` changes, no `loom-daemon status` UI additions, and no bash-side journal writes, matching the stated scope guard.

**Deferred items — reasonable:** The stale-lock prune (`sweep_registry.rs` item 4) and event-bus emission are both explicitly out of the acceptance criteria (only the WARN-log requirement is in-scope), and the issue's own scope guard pre-authorizes deferring anything that would push the PR past its LOC target rather than growing it further. Both are called out as explicit follow-up candidates in the module doc comment, so they won't be silently lost.

**Minor, non-blocking observations (not requesting changes for these):**
- The module doesn't call out the pid-reuse limitation the issue's test-plan asked to "document as a known limitation" — though this is symmetric with the existing journal-backed `DeadPid` path, which has the same limitation and also doesn't document it, so it's pre-existing posture rather than a regression.
- Net additions (698, mostly in `claim_reconciliation.rs`) run well past the issue's own "~400 LOC including tests" target even after deferring the stale-lock prune — but ~55% of the added lines are tests and the rest is this module's already-established doc-comment-heavy style (matching `sweep_registry.rs`/`work_finder.rs` conventions elsewhere in the crate), not unexplained complexity. Not blocking.

Nice work — this closes a real gap (three existing recovery layers all structurally missed the manually-spawned-detached-sweep case) with a well-tested, fail-safe evidence source.

Issue #3968, opened 2026-07-26
Self-update gap: deploying the daemon's own merged fixes requires manual rebuild + provision + restart


rjwalters:
## Context

The 2026-07-25/26 canary rollout proved the self-repair loop works — the daemon filed and fixed 16 of its own defects — but every merged daemon fix only takes effect after an operator manually: rebuilds the Rust binary, provisions it to `~/.local/bin/loom-daemon`, and restarts the process. During the rollout this was done by hand several times, and the session flagged it twice as "a self-update gap worth filing when it starts to grate." It grates.

Note the restart side is already half-solved: a clean SIGTERM stop leaves in-flight sweep children running by design, so a rebuild-and-restart window doesn't kill active work.

## Suggested acceptance criteria
- [ ] A single operator command (e.g. `loom-daemon-update.sh` or `loom-daemon self-update`) that: detects the running binary is older than the current source tree (or origin/main), rebuilds, provisions, and restarts with the same flags the daemon was started with
- [ ] Preserves the FLAGS-OFF/opt-in contract — the restarted daemon runs with exactly the prior autonomy flags, never more
- [ ] Optionally: the daemon notices its source checkout has newer merged daemon commits and surfaces "update available" in `loom-daemon --status` / the status MCP surface (no auto-restart without opt-in)

rjwalters:
**Scope update from the operator (2026-07-26, post-launchd migration)** — this issue's deliverable should be understood as a **"roll/update daemon" command** (script + optionally a `/loom:*` skill surface), and the restart mechanics changed since filing:

The daemon now runs as a launchd LaunchAgent (`~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist`, label `com.rjwalters.loom-daemon`) rather than a nohup child — see #3972 for why (session bootstrap-namespace death broke all gh/git in the daemon's tree). A roll command must therefore:

1. **Rebuild + provision**: `cargo build --release` from the loom checkout, install to `~/.local/bin/loom-daemon` (the plist's `ProgramArguments` path).
2. **Restart via launchd, not signals/scripts**: `launchctl kickstart -k gui/$UID/com.rjwalters.loom-daemon` (or bootout → bootstrap when the plist itself changed). Never fall back to `nohup` — that reintroduces the #3972 hazard.
[...]

PR #4395, opened 2026-07-29
fix(daemon): --from-config composes with --work-finder/--health-gate instead of silently ignoring them

### LINKED ISSUE #4353: loom-daemon-start.sh: --from-config silently swallows --work-finder/--health-gate instead of composing or rejecting
Observed live (2026-07-28, new-host onboarding): `loom-daemon-start.sh --from-config --work-finder` starts the daemon with the work finder **OFF** when the repo config has no `autonomous.workFinder` block.

The flag parser accepts both flags, persists both verbatim to `.loom/.daemon.flags`, but the env-composition block is an either/or: when `FROM_CONFIG=true` it prints "Autonomous mode: driven by .loom/config.json -> autonomous (env not forced)" and never exports `LOOM_WORK_FINDER=1`, so an explicit `--work-finder` is silently ignored. Operator impression: work finder enabled. Reality: reliability daemon. No warning, no error — and `.daemon.flags` now records an intent (`--from-config` + `--work-finder`) that `loom-daemon-update.sh` will faithfully reproduce as the same silent no-op after every rebuild.

Suggested fix (either is fine):
1. **Compose** — precedence is documented as env > config > default, so `--from-config --work-finder` should mean "config drives everything except the explicitly forced loop": still export `LOOM_WORK_FINDER=1`, let config drive the rest. This matches operator intuition and the documented precedence.
2. **Reject** — if composition is undesirable, make the combination a usage error at parse time rather than accepting and ignoring.

Workaround that works today: drop `--from-config` and pass `--work-finder` alone (config-only surfaces like `autonomous.roleRunner` are read from config regardless of the flag, so nothing is lost).

---

### Root cause (verified on origin/main)

`defaults/scripts/cli/loom-daemon-start.sh` (installed as `.loom/scripts/cli/loom-daemon-start.sh` — cite/edit the `defaults/` copy) parses all flags fine, but the env-composition block at **lines 805–826** is a strict either/or:

- `FROM_CONFIG=true` branch (line 805–806): prints the "driven by .loom/config.json" banner and **never reads `WANT_WORK_FINDER` / `WANT_HEALTH_GATE`** — so `--work-finder` / `--health-gate` are accepted, persisted, and ignored.
- The `else` branch (808–825) is the only place the `WANT_*` values are consulted.

Two adjacent facts confirm option 1 (**compose**) is the right fix and is cheap:

1. **The daemon's own precedence is already env > config > default.** `LOOM_WORK_FINDER` (`WORK_FINDER_ENABLE_ENV`, `loom-daemon/src/work_finder.rs:111`) and `LOOM_MAIN_HEALTH_GATE` (`MAIN_HEALTH_GATE_ENABLE_ENV`, `loom-daemon/src/main_health_gate.rs:102`) win over the `autonomous.*` config blocks when exported. So "config drives everything except the explicitly forced loop" needs **no daemon-side change** — the start script just has to export the forced var and leave the other unset.
2. **Both durability paths already carry the composed result.** The launchd plist snapshots every exported `LOOM_*` var into `EnvironmentVariables` (the `env | grep -E '^(LOOM_…'` loop at ~line 333), and the pid-file/nohup path replays `.loom/.daemon.flags` verbatim (`defaults/scripts/cli/loom-daemon-update.sh:798–803`). Once composition works at start time, restarts and updates reproduce it with no further changes.

### The same bug in the opposite direction (in scope)

`--from-config --no-work-finder` is equally swallowed: config can turn a loop ON despite an explicit `--no-*`. Because `WANT_WORK_FINDER` is a plain boolean, the parser cannot distinguish "default off" from "explicit `--no-work-finder`". The fix therefore needs a **tri-state** per loop (e.g. `WANT_WORK_FINDER` ∈ `""` (unset) / `on` / `off`):

- `--from-config` + unset → leave the env var unset (config drives — today's behavior, test case 6).
- `--from-config` + `on` → `export LOOM_WORK_FINDER="${LOOM_WORK_FINDER:-1}"`.
- `--from-config` + `off` → `export LOOM_WORK_FINDER="${LOOM_WORK_FINDER:-0}"`.
- Without `--from-config`: unset behaves as `off` (FLAGS-OFF default) — byte-for-byte today's behavior.

Keep the `${VAR:-…}` form: an **already-exported env var still wins over CLI flags** (documented at line 81 and exercised by the existing test's `env -u` hygiene) — do not change that ordering.

### Implementation guidance

1. Convert the two `WANT_*` booleans (lines 667–668, 678–681) to tri-state.
2. In the `FROM_CONFIG=true` branch, export only the explicitly-set loops; update the banner to say what is forced vs config-driven, e.g. `Autonomous mode: config-driven; forced: work_finder=1` (never print the misleading "env not forced" line when something IS forced).
3. Update the in-file docs: header usage block (lines 65–70), the precedence comment at 778–784, and `show_help`.
4. Update `defaults/docs/daemon-reference.md` — lines ~2328–2345 currently state "`--from-config` leaves both" unconditionally; document the composed semantics. (Mind `scripts/check-claude-md-budget.sh` — this is a docs-file edit, not CLAUDE.md.)
5. Rejection (option 2) is **not** recommended: composition matches the documented precedence and operator intuition, and rejection would break the recorded-intent replay in existing `.daemon.flags` files that already contain the combination.

### Acceptance criteria

- [ ] `--from-config --work-finder` starts the daemon with `LOOM_WORK_FINDER=1` exported and `LOOM_MAIN_HEALTH_GATE` unset (config drives the gate).
- [ ] `--from-config --health-gate` is the mirror image.
- [ ] `--from-config --no-work-finder` exports `LOOM_WORK_FINDER=0` (explicit off beats config).
- [ ] Plain `--from-config` still leaves both unset (existing test case 6 stays green).
- [ ] A pre-exported `LOOM_WORK_FINDER` still wins over any CLI flag.
- [ ] The startup banner names the forced loop(s); the "env not forced" wording never appears when a loop was forced.
- [ ] `.daemon.flags` replay via `loom-daemon-update.sh` reproduces the composed behavior (verbatim flag persistence already guarantees this once composition works — assert, don't reimplement).

## Affected Files

- `defaults/scripts/cli/loom-daemon-start.sh` - tri-state `WANT_*`, compose in the `FROM_CONFIG` branch (lines 805–826), banner + header docs + `show_help`
- `defaults/scripts/tests/test-loom-daemon-start.sh` - new cases for the four `--from-config` combinations plus env-wins, following the existing `FAKE_DAEMON WF=[..] HG=[..]` pattern (cases 2–7)
- `defaults/docs/daemon-reference.md` - correct the "`--from-config` leaves both" wording (~lines 2328–2345) to the composed semantics

## Test Plan

- [ ] Automated tests: extend `defaults/scripts/tests/test-loom-daemon-start.sh` with: `--from-config --work-finder` → `WF=[1] HG=[]`; `--from-config --health-gate` → `WF=[] HG=[1]`; `--from-config --no-work-finder` → `WF=[0] HG=[]`; `--from-config` alone → `WF=[] HG=[]` (regression); `LOOM_WORK_FINDER=0` pre-exported + `--from-config --work-finder` → `WF=[0]` (env wins). Run the suite directly: `bash defaults/scripts/tests/test-loom-daemon-start.sh`.
- [ ] Manual verification: on a repo whose config has no `autonomous.workFinder` block, `loom-daemon-start.sh --from-config --work-finder --foreground` and confirm the daemon logs the work finder as enabled (not `work_finder: disabled` from `loom-daemon/src/main.rs:1229`).
- [ ] Edge cases: `--work-finder --no-work-finder` in one invocation (last one wins, as today); `.daemon.flags` written by a pre-fix invocation replayed by the post-fix script composes correctly.

rjwalters:
## Summary

Closes #4353

`defaults/scripts/cli/loom-daemon-start.sh`'s env-composition block was a
strict either/or: when `--from-config` was passed, the script never looked
at `WANT_WORK_FINDER`/`WANT_HEALTH_GATE` at all, so `--from-config
--work-finder`/`--no-work-finder`/`--health-gate`/`--no-health-gate` were
silently dropped — the loop stayed fully config-driven even though the
operator explicitly asked to force it.

### Fix

- Converted `WANT_WORK_FINDER`/`WANT_HEALTH_GATE` to tri-state (`""` =
  unset/not-passed, `"on"`, `"off"`) instead of a boolean that collapsed
  "not passed" and "explicit off" to the same value.
- In the `FROM_CONFIG=true` branch: an explicit `--work-finder`/
  `--no-work-finder` (or the health-gate equivalent) now forces that ONE var
  using the `${VAR:-default}` form (so an already-exported env var still
  wins), while the loop with no explicit flag is left untouched (stays
  unset, config drives it). Plain `--from-config` alone is unchanged —
  leaves both vars unset (existing test case 6 stays green).
- Startup banner now names what was forced, e.g. `Autonomous mode:
  config-driven; forced: work_finder=1`, and never prints the misleading
  "env not forced" wording when something WAS forced.
- Updated the header usage block, the precedence comment, and
  `defaults/docs/daemon-reference.md`'s "`--from-config` leaves both"
  wording to describe the composed semantics.
- `.daemon.flags` persistence (`ORIGINAL_ARGS`-based, ie. issue #3968) is
  untouched — it already stores the raw CLI args verbatim, so
  `loom-daemon-update.sh`'s replay automatically reproduces the composed
  behavior once composition is correct at start time. Verified manually
  (see test plan) rather than re-running the update-suite (which has a
  known live hazard against the production `loom-daemon` binary).

Did **not** implement the "reject the combination" alternative — composition
is the fix the issue calls for.

## Test plan

- [x] `bash defaults/scripts/tests/test-loom-daemon-start.sh` — all 46 tests
      pass, including 6 new cases:
  - `--from-config --work-finder` → `LOOM_WORK_FINDER=1`, gate unset
  - `--from-config --no-work-finder` → `LOOM_WORK_FINDER=0`, gate unset
  - `--from-config --health-gate` → `LOOM_MAIN_HEALTH_GATE=1`, finder unset
  - `--from-config --no-health-gate` → `LOOM_MAIN_HEALTH_GATE=0`, finder unset
  - pre-exported `LOOM_WORK_FINDER=1` wins over `--from-config
    --no-work-finder`
  - startup banner names the forced loop, never prints "env not forced" when
    something was forced
  - existing test case 6 (plain `--from-config` leaves both unset) stays green
- [x] `bash -n` syntax check
- [x] Manual check: `--from-config --work-finder --foreground` persists
      `--from-config` + `--work-finder` verbatim to `.daemon.flags` (the
      exact args `loom-daemon-update.sh` replays on restart)

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2553,6 +2553,11 @@ process:
 # Enable strictly per .loom/config.json → autonomous (no env forcing):
 ./.loom/scripts/cli/loom-daemon-start.sh --from-config
 
+# --from-config COMPOSES with an explicit flag (#4353): the named loop is
+# forced, the other is left for config to drive:
+./.loom/scripts/cli/loom-daemon-start.sh --from-config --work-finder      # force finder ON, gate config-driven
+./.loom/scripts/cli/loom-daemon-start.sh --from-config --no-health-gate   # force gate OFF, finder config-driven
+
 # Explicit-off / foreground variants:
 ./.loom/scripts/cli/loom-daemon-start.sh --no-work-finder   # force finder off (explicit; same as default)
 ./.loom/scripts/cli/loom-daemon-start.sh --no-health-gate   # force gate off (explicit; same as default)
@@ -2568,9 +2573,13 @@ process:
   and `LOOM_MAIN_HEALTH_GATE=0`, so a plain start is a **reliability daemon** that
   does **not** auto-dispatch sweeps — consistent with the ecosystem-wide opt-in /
   default-off contract. An already-exported env var always wins; `--work-finder`
-  / `--health-gate` force the respective loop on; `--from-config` leaves both
-  unset so `.loom/config.json → autonomous` drives (precedence env > config >
-  default),
+  / `--health-gate` force the respective loop on; `--from-config` alone leaves
+  both unset so `.loom/config.json → autonomous` drives (precedence env >
+  config > default). **`--from-config` composes with an explicit flag rather
+  than ignoring it (#4353)**: `--from-config --work-finder` (or
+  `--no-work-finder`/`--health-gate`/`--no-health-gate`) still FORCES that one
+  var while leaving the other loop unset for config to drive — a
+  pre-exported env var still wins over the forced flag,
 - locates the `loom-daemon` binary (`LOOM_DAEMON_BIN` → `PATH` → `target/{release,debug}`),
 - runs the **advisory** host-sleep check (`check-host-sleep.sh`, #3350) — never blocks the start,
 - **on macOS, backgrounds the daemon as a launchd LaunchAgent** in the resolved
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -68,6 +68,8 @@
 #   ./.loom/scripts/cli/loom-daemon-start.sh --from-config   Enable per .loom/config.json only
 #   ./.loom/scripts/cli/loom-daemon-start.sh --no-work-finder    Force work finder OFF (explicit)
 #   ./.loom/scripts/cli/loom-daemon-start.sh --no-health-gate    Force health gate OFF (explicit)
+#   ./.loom/scripts/cli/loom-daemon-start.sh --from-config --work-finder   Config-driven, but FORCE the work finder on (#4353)
+#   ./.loom/scripts/cli/loom-daemon-start.sh --from-config --no-health-gate   Config-driven, but FORCE the health gate off (#4353)
 #   ./.loom/scripts/cli/loom-daemon-start.sh --foreground    Run in the foreground (no PID file)
 #   ./.loom/scripts/cli/loom-daemon-start.sh --no-launchd    macOS only: use legacy nohup instead of a LaunchAgent
 #   ./.loom/scripts/cli/loom-daemon-start.sh --no-systemd    Linux only: use legacy nohup instead of a systemd --user service
@@ -79,6 +81,7 @@
 #   LOOM_DAEMON_BIN     Path to the loom-daemon binary (else auto-detected)
 #   LOOM_SOCKET_PATH    Override the daemon socket (default ~/.loom/loom-daemon.sock)
 #   LOOM_WORK_FINDER / LOOM_MAIN_HEALTH_GATE  Respected when already exported
+#                        (always wins, even under --from-config -- #4353)
 #   LOOM_DAEMON_LAUNCHD  macOS only: 0/false/no forces the legacy nohup path (same as --no-launchd)
 #   LOOM_DAEMON_SYSTEMD  Linux only: 0/false/no forces the legacy nohup path (same as --no-systemd)
 #   LOOM_SYSTEMD_UNIT    Linux only: override the systemd --user unit name (default loom-daemon.service)
@@ -699,8 +702,14 @@ ORIGINAL_ARGS=("$@")
 # --health-gate, or hand control to config with --from-config.
 FROM_CONFIG=false
 FOREGROUND=false
-WANT_WORK_FINDER=false
-WANT_HEALTH_GATE=false
+# Tri-state (#4353): "" = not passed on the CLI (unset), "on" = an explicit
+# --work-finder/--health-gate, "off" = an explicit --no-work-finder/
+# --no-health-gate. This lets --from-config tell "the operator asked to force
+# this loop" apart from "the operator said nothing, config drives it" --
+# a plain boolean collapsed both to the same false and silently dropped the
+# force.
+WANT_WORK_FINDER=""
+WANT_HEALTH_GATE=""
 NO_LAUNCHD=false
 NO_SYSTEMD=false
 PRINT_PLIST=false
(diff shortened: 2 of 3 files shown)


rjwalters:
Approved.

Verified against issue #4353's acceptance criteria:

1. **Tri-state conversion is correct.** `WANT_WORK_FINDER`/`WANT_HEALTH_GATE` now use `""`/`"on"`/`"off"` instead of a boolean. Plain `--from-config` (no flags) still leaves both vars unset — confirmed by existing test case 6 staying green, and by re-running the full suite.
2. **`${VAR:-default}` form used correctly** in the new `FROM_CONFIG=true` branch (`export LOOM_WORK_FINDER="${LOOM_WORK_FINDER:-1}"` etc.), so a pre-exported env var wins over the forced CLI flag — exercised by new test 7e (`LOOM_WORK_FINDER=1` pre-exported beats `--from-config --no-work-finder`).
3. **Test suite**: ran `bash defaults/scripts/tests/test-loom-daemon-start.sh` directly in the worktree — all 46 tests pass, including the 6 new cases (7a–7f: force-on, force-off, health-gate mirror, env-wins, and the banner-wording assertion that `forced: work_finder=1` appears and `env not forced` does not).
4. **Docs updated consistently**: header usage block (new `--from-config --work-finder`/`--no-health-gate` examples), the `Environment:` block note on `LOOM_WORK_FINDER`/`LOOM_MAIN_HEALTH_GATE` always winning, the precedence comment above the `FROM_CONFIG` branch, and `defaults/docs/daemon-reference.md`'s composed-semantics wording. `show_help()` is generated by awk-stripping the header comment banner, so it inherits the updated usage lines automatically — confirmed via the existing `--help` test (test 8) still passing.
5. **Only `defaults/` files touched** — `git diff --stat` shows exactly `defaults/docs/daemon-reference.md`, `defaults/scripts/cli/loom-daemon-start.sh`, `defaults/scripts/tests/test-loom-daemon-start.sh`. No installed `.loom/` synced copies were edited.

CI: all checks pass (`gh pr checks 4395` — Installer Integration Tests, Shellcheck, CLAUDE.md Line Budget, Detect Changes all green; Rust jobs skipped as expected for a shell-only change). `mergeStateStatus` is `CLEAN`.

Nice, minimal fix that composes rather than special-cases — matches the documented env > config > default precedence exactly.

Issue #4347, opened 2026-07-29
cleanup: loom repo as a managed workspace is unhealthy — broken issue-4188/mcp-loom reference errors role ticks, orphaned /tmp review worktrees, dirty main


rjwalters:
## Context

Operator report (2026-07-29): the loom repo itself is unhealthy as a *managed workspace* — role ticks error against a broken MCP worktree (`worktrees/issue-4188/mcp-loom` — cannot rebuild), and main is dirty with 12 uncommitted changes. Noise in the daemon log; blocks loom's own gate; does not affect other repos' dispatch. Deferred by the operator to a cleanup pass ("sometime, not tonight") — this issue is that pass's work order.

Evidence gathered at filing time:

- **No `issue-4188` worktree exists** under the repo's `.loom/worktrees/` — the erroring path is stale state somewhere the role ticks still resolve (machine-level checkout, daemon workspace registry, or dangling git-worktree metadata). Finding *where* the reference lives is part of the fix.
- **`git worktree list` shows five orphaned review worktrees in `/private/tmp`** (`pr-4208-review`, `pr4183-check`, `pr4184-check`, `pr4227-review`, `pr4297-worktree`) — leftovers from past Judge sessions doing test-merges/reviews. tmp dirs are reaped by the OS, leaving dangling worktree refs that confuse tooling. This is the operational sibling of #4281 (judges must use disposable worktrees *and clean them up*; `git worktree add` outside `worktree.sh` also bypasses the `.loom-managed` sentinel that authorizes auto-cleanup).
- **Dirty main (12 changes)** — already filed as #4332 (resync-installed artifacts blinding the gate); the cleanup pass should execute whichever resolution #4332 lands (or just commit the resync output once, manually, as the immediate unblock).
- ~11 `.loom/worktrees/issue-N` entries exist; some belong to open PRs (legitimate), some may be stale post-merge leftovers — `loom-clean` should adjudicate via the `.loom-managed` sentinel.

## Suggested acceptance criteria

- [ ] Locate and remove the broken `issue-4188/mcp-loom` reference; role ticks stop erroring on it (daemon log clean for one full tick cycle).
- [ ] `git worktree prune` + removal of the dangling `/private/tmp/pr*` entries; `git worktree list` shows only main + legitimate `.loom/worktrees/issue-N` entries.
- [ ] Main checkout clean (per #4332's resolution or a one-time commit of resync output); loom's main-health gate produces a real verdict again.
[...]

PR #4396, opened 2026-07-29
fix(tests): sandbox loom-daemon-update.sh test suite's machine-level provisioning destination

### LINKED ISSUE #4381: INCIDENT: update-suite FAKE_DAEMON stub overwrote the production ~/.local/bin/loom-daemon — CLI hung 9h, crash would have relaunched the stub
## Incident

2026-07-29 ~06:03Z: the production CLI entry point `~/.local/bin/loom-daemon` was **overwritten with a test stub** — the FAKE_DAEMON fixture from `defaults/scripts/tests/test-loom-daemon-update.sh:134` (byte-identical marker line, including the live tmpdir path `/var/folders/.../tmp.g9YwW6ZsgV/w43/new-marker`). The stub answers `--version` with a canned string (`commit 0ef9c17, built 2026-07-26T00:00:00Z`) and then `while true; sleep 1` — an infinite hang.

**Impact (silent for ~9 hours):**
- Every subsequent `loom-daemon` CLI invocation hung forever: an operator `status` poll (9.5h), and two `tokens select` probes (~25 min each) on the spawn path.
- The real daemon (pid 27204) kept running only because it was exec'd before the overwrite — **a crash at any point would have had launchd relaunch the stub**, which would have looked "loaded" while doing nothing.
- Detection was luck: an operator `loom-daemon status` timing out. `--version` even "worked" (canned), defeating a naive freshness check.

**Recovery (done):** stub preserved as evidence, real binary restored from `target/release` (e4dbbdba), re-signed with the stable identity, hung processes killed, CLI verified responsive. Fresh roll dispatched.

## Root cause to find and fix

This is the **#4078 incident class** (test artifact escapes to production daemon surface — fixed then via PR #4087 for the boot-out vector) on a **new vector**: the update-suite's fake-binary provisioning wrote to the real `~/.local/bin/loom-daemon` instead of the test's sandboxed prefix. The suite ran at least twice on this host in the window (an in-session Judge pass at ~05:05Z reported 82/82 green; concurrent day-time automation re-ran it) — one run's isolation failed. Candidate holes to check in `test-loom-daemon-update.sh` (and the update script under test):

- a branch of the provision path that resolves the install destination from the real `$HOME`/`~/.local/bin` even when the test overrides the prefix (env not threaded through `sudo`-less subshells, or a hardcoded fallback),
- the test running against the REAL `loom-daemon-update.sh` with only `LOOM_DAEMON_BIN` faked but the install destination not redirected,
- ordering: the stub's mtime (06:03Z) vs which runner (in-session judge worktree vs daemon-dispatched sweep) was active then — `.loom/logs` + transcript archive can pin the actor.

- [ ] Identify the exact escape path and reproduce it in a regression test (test must FAIL if any test writes outside its sandbox prefix).
- [ ] The update/start/watchdog test suites assert post-run that `~/.local/bin/loom-daemon` (and the machine-checkout binary path) are untouched (checksum before/after).
- [ ] `loom-daemon-update.sh`/provision refuse to install a non-Mach-O (or non-`--version`-sane) binary to the production path — a cheap `file`/exec sanity gate would have caught a bash stub.
- [ ] Watchdog/health check detects "CLI hangs" (bounded-timeout self-probe), not just process-exists — this incident was invisible to every existing health signal.

## Related

- #4078 / PR #4087 (prior incident class + fix) · #4347 (loom-repo cleanup — hung stub processes were part of tonight's noise) · #4331 (watchdog contract) · #4344 (the other concurrent daemon impairment)

rjwalters:
## Summary

2026-07-29 ~06:03Z incident (#4381): the production CLI entry point
`~/.local/bin/loom-daemon` was overwritten with a test-fixture stub (a bash
script that answers `--version` correctly, then `while true; sleep 1`) —
causing every subsequent `loom-daemon` CLI invocation to hang for ~9 hours.

**Root cause**: `defaults/scripts/tests/test-loom-daemon-update.sh` has three
tests (37, 39, 43 — the ff-sync/staleness-detection cases) that invoke
`loom-daemon-update.sh` deliberately with **neither** `LOOM_DAEMON_BIN` nor a
stubbed `scripts/install/provision-daemon.sh` set. In that configuration the
update script's real `provision_machine_daemon()` call resolves its
destination as `${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}` — the operator's
own real `$HOME`, since none of these three tests override `HOME` either. The
final test in the file (43) writes its fake daemon to
`$W43/new-marker` and `~/.local/bin/loom-daemon` — matching the incident's
observed marker path (`tmp.g9YwW6ZsgV/w43/new-marker`) exactly.

## Fix

- Export a suite-wide `LOOM_DAEMON_BIN_DIR="$BASE_WORKDIR/machine-level-bin-sandbox"`
  (mirroring the existing `LOOM_AUTONOMY_MARKER`/`LOOM_WATCHDOG_LABEL`
  suite-wide sandboxing already in this file), threaded to every
  sub-invocation of the update script. Tests that already pin
  `LOOM_DAEMON_BIN` are unaffected (that branch bypasses
  `provision_machine_daemon` entirely); tests 37/39/43 now provision into the
  sandbox instead of the real machine path.
- Add a checksum-guard regression test: snapshot `~/.local/bin/loom-daemon`'s
  checksum before the suite runs anything, and assert it is byte-identical
  after the whole suite completes. This fails loudly if the sandboxing above
  is ever bypassed, removed, or a new call site is added without it —
  the direct regression test for this incident.

## Verification

Ran the full suite in a worktree, with a manual `shasum -a 256
~/.local/bin/loom-daemon` before/after to independently confirm the real
production binary was untouched (per the operator's standing safety note
about this exact hazard):

```
Ran 107 tests: 107 passed, 0 failed
✓ real /Users/rwalters/.local/bin/loom-daemon is byte-identical before/after the suite (#4381 regression guard)
```

`file` confirmed Mach-O before and after; `shasum -a 256` was identical
(`ded197fb...`) before and after the run.

## Scope note

The issue's other suggested acceptance criteria (a binary-format/exec sanity
gate on `provision_machine_daemon`, and a hang-aware watchdog probe) are
deferred:

- A `file`-based Mach-O/ELF sanity gate on the provisioned binary directly
  conflicts with this suite's own fixture design — every `write_fake_daemon`
  fixture *is* a bash script standing in for the compiled binary, so a strict
  format gate would fail the whole suite. Doing this properly needs the test
  fixtures reworked to use real precompiled stand-ins (or an explicit,
  auditable test-mode bypass), which is a separate, larger change — worth its
  own follow-up issue.
- A hang-aware watchdog self-probe (bounded-timeout CLI liveness check, not
  just process-exists) is a genuinely new feature, not a fix to this test
  suite, and is better scoped as its own issue (there's already a related
  watchdog-contract issue, #4331).

The sandboxing fix + regression test is the load-bearing deliverable per the
issue's own stated priority, and is what directly prevents recurrence of this
incident.

## Related

#4078 / PR #4087 (prior incident class + fix) · #4347 · #4331 · #4344

Closes #4381

### DIFF
--- a/defaults/scripts/tests/test-loom-daemon-update.sh
+++ b/defaults/scripts/tests/test-loom-daemon-update.sh
@@ -24,6 +24,37 @@
 
 set -uo pipefail
 
+# Production-binary checksum guard (#4381 incident): 2026-07-29 ~06:03Z, the
+# REAL machine-level `~/.local/bin/loom-daemon` was overwritten with a fake,
+# infinitely-looping test-fixture stub for ~9 hours (an operator `status` poll
+# hung; a crash at any point would have had launchd relaunch the stub). Root
+# cause: three tests below (37/39/43, the ff-sync/staleness-detection cases)
+# deliberately invoke loom-daemon-update.sh with NEITHER LOOM_DAEMON_BIN NOR a
+# stubbed scripts/install/provision-daemon.sh, so the update script's
+# provision_machine_daemon() call fell through to its real, un-sandboxed
+# default destination (`${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}`) — the
+# operator's own real $HOME. The LOOM_DAEMON_BIN_DIR export below (threaded to
+# every sub-invocation) is the actual fix; THIS checksum snapshot is the
+# regression backstop that fails the suite outright if any current or future
+# test call site regresses into writing the real production binary, even if
+# the sandboxing above is bypassed or a new call site is added without it.
+# Recorded before ANYTHING else runs, using the checksum tool most likely
+# present (sha256sum on Linux, shasum on macOS); skipped (empty) only when
+# neither is on PATH or no binary exists yet at that path (nothing to compare
+# against — a fresh-machine first run).
+_PROD_DAEMON_BIN="$HOME/.local/bin/loom-daemon"
+_prod_daemon_checksum() {
+    [[ -e "$_PROD_DAEMON_BIN" ]] || { echo "<absent>"; return 0; }
+    if command -v shasum >/dev/null 2>&1; then
+        shasum -a 256 "$_PROD_DAEMON_BIN" 2>/dev/null | awk '{print $1}'
+    elif command -v sha256sum >/dev/null 2>&1; then
+        sha256sum "$_PROD_DAEMON_BIN" 2>/dev/null | awk '{print $1}'
+    else
+        echo "<no-checksum-tool>"
+    fi
+}
+_PROD_DAEMON_CHECKSUM_BEFORE="$(_prod_daemon_checksum)"
+
 # Force the legacy nohup path everywhere in this suite (#3972): the restart
 # flow below exercises the REAL loom-daemon-start.sh / loom-daemon-stop.sh
 # (not a mock), and this test must NEVER touch the real machine's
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

This is exactly the right fix for the #4381 incident: a suite-wide `LOOM_DAEMON_BIN_DIR` sandbox (the actual hole-closer) plus a before/after checksum guard on the real production binary (the loud regression backstop). Verified the mechanism end-to-end:

- `defaults/scripts/cli/loom-daemon-update.sh:911` and `provision_machine_daemon` (`scripts/install/provision-daemon.sh:199`) both resolve their destination as `${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}`, so the single suite-wide export covers every provisioning path — and `provision-daemon.sh:228` does `mkdir -p`, so the sandbox dir needs no pre-creation.
- No test in the suite unsets/overrides `LOOM_DAEMON_BIN_DIR`, and no invocation uses `env -i`, so the export reaches tests 37/39/43 (confirmed test 37 invokes via a plain env-prefixed subshell).
- Checksum helper degrades gracefully (absent binary, missing checksum tool) without false failures.

## Test Execution

1. Run full suite in worktree — ✅ Executed: `Ran 107 tests: 107 passed, 0 failed`, including the new guard: `✓ real /Users/rwalters/.local/bin/loom-daemon is byte-identical before/after the suite (#4381 regression guard)`
2. Independent before/after `shasum -a 256` of the production binary — ✅ Executed: `ded197fb…` identical before and after; `file` confirms Mach-O 64-bit arm64 both times.

## Deferred ACs now tracked

The scope note's two deferrals were correct calls, but one pointed at #4331 which closed (completed) before this PR — leaving it untracked. Filed both as triage suggestions:
- #4397 — binary-format sanity gate on `provision_machine_daemon` (needs fixture rework, as the scope note explains)
- #4398 — hang-aware watchdog liveness probe (bounded-timeout CLI check; a pid-exists check passes forever against the incident's hung stub)

CI green, merge state CLEAN, `Closes #4381` present.

Issue #4412, opened 2026-07-29
main broken: serve.rs empty_report() missing new DaemonStatusReport main_health_gate_* fields (#4404 × #4297 semantic conflict)


rjwalters:
## Bug: main does not compile (`cargo check --workspace --all-targets`) — semantic conflict between #4404 and #4297

As of `a410a506` (feat(daemon): loom-daemon serve skeleton, PR #4404), **main is broken** for all `--all-targets` builds, which fails the `Rust Linting & Build Check` and `Rust Unit Tests` CI jobs on **every open PR**.

```
error[E0063]: missing fields `main_health_gate_deferred`, `main_health_gate_deferred_reason`
and `main_health_gate_verdict_tier` in initializer of `types::DaemonStatusReport`
   --> loom-daemon/src/serve.rs:328:9
```

## Root cause

A classic semantic merge conflict between two PRs that landed near-simultaneously (2026-07-29, ~17:00–17:20 UTC):

- `5192592e` (PR #4297, tiered build gate) added three new fields to `DaemonStatusReport` in `loom-daemon/src/types.rs`: `main_health_gate_deferred`, `main_health_gate_deferred_reason`, `main_health_gate_verdict_tier`.
- `a410a506` (PR #4404, serve skeleton) added `loom-daemon/src/serve.rs`, whose test helper `empty_report()` (line ~328) constructs a `DaemonStatusReport` literal that was written before #4297 merged and does not initialize those three fields.

Each PR was green against the main it branched from; the combination doesn't compile. The breakage is in `#[cfg(test)]` code, so `cargo build` succeeds but `cargo check --all-targets` / `cargo test` fail.

## Fix

One-line-ish: add the three missing fields (defaults: `main_health_gate_deferred: false`, `main_health_gate_deferred_reason: None`, `main_health_gate_verdict_tier: None`) to the `empty_report()` initializer in `loom-daemon/src/serve.rs`, or give `DaemonStatusReport` a `Default`/builder for test helpers so the next additive field can't break out-of-tree initializers.

## Impact

- **Severity: high** — all PR CI is red until this lands (observed failing on PR #4408's run 30475336679; PR #4407 will hit it too).
- Reproduce: `git checkout main && cargo check --workspace --all-targets`.

Discovered while evaluating PR #4408 (its diff does not touch `serve.rs`; the failure reproduces on bare main).

---

## Curator Enhancement (2026-07-29)

[...]

PR #4404, opened 2026-07-29
Dashboard phase 1: loom-daemon serve skeleton + JSON status snapshot endpoint

rjwalters:
## Summary

Closes #4391. Phase 1 of 4 in the `#4329` dashboard decomposition: stand up
the `loom-daemon serve` subcommand skeleton and a single read-only JSON
status-snapshot HTTP endpoint. No HTML page, no SSE stream, no cross-host
aggregation — those are phases 2-4 (#4392/#4393/#4394).

## What changed

- **New `Commands::Serve` subcommand** (`loom-daemon/src/main.rs`): `loom-daemon
  serve [--port <PORT>] [--bind <ADDR>] [--allow-non-loopback]`. Routed through
  the async runtime (like `status`/`dispatch`/`watch`), never through the
  daemon's main foreground-run path — starting the daemon normally never opens
  this port.
- **New `loom-daemon/src/serve.rs` module** (exposed via `lib.rs`):
  - `validate_bind(addr, allow_non_loopback)` — the security gate: loopback
    always allowed; a wildcard/unspecified bind (`0.0.0.0`/`::`) is refused
    **unconditionally**, even with the opt-in flag; any other non-loopback
    address (e.g. a tailnet interface) requires `--allow-non-loopback`
    explicitly — the bind address alone is never enough.
  - `fetch_snapshot` — connects to the running daemon over its **existing**
    Unix socket and sends the **existing** `Request::DaemonStatus` request
    (same wire contract `loom-daemon status --json` already uses). No new IPC
    request type. The actual aggregation (dynamic caps, per-repo breakdown,
    capacity, drain state, …) happens exactly once, inside
    `ipc::build_daemon_status` — this module never re-derives it.
  - `StatusSnapshot` — the `DaemonStatusReport` flattened alongside a
    `hostname` field (`#[serde(flatten)]`), reusing the daemon's existing
    `sweep_registry::host_identity()` helper. Per the issue's operator
    comment (added before curation locked scope): the fleet is now 3 daemons,
    so this phase's one small addition beyond the base skeleton is a
    host-identity field so a later client-side aggregator (#4393) can label
    sources — no aggregation logic lands in this phase.
  - `run`/`handle_connection` — a minimal hand-rolled HTTP/1.1 responder
    (single route, `GET /api/status`, bounded header-read size, always
    `Connection: close`). 404 for any other path, 405 for non-GET, 503 when
    the daemon is unreachable.

## HTTP dependency decision

[...]

PR #4297, opened 2026-07-28
feat(build-gate): tiered gate + bounded load-aware deferral (#4259)

rjwalters:
Closes #4259

## Problem

Under 6–7 concurrent sweeps the daemon's main-health gate (`bash .loom/scripts/build-gate.sh`) ran the **full** suite, blew past even the 1200s budget, was killed, and reported `not evaluated (timeout …)` **permanently** — silently disabling the red-main protection it exists for. `nice` (#4020/#4084) reordered the run queue but never stopped the gate's own `cargo` build from racing freshly-dispatched sweep builds for cores.

## What changed

Two composed mechanisms so the gate always produces a signal:

**1. FAST tier in `build-gate.sh`** — `LOOM_BUILD_GATE_TIER=fast` runs `cargo build --workspace --lib --bins` (compile — catches #3647 step-8-class breakage) + a `loom_tools` import smoke. The full three-stage suite stays the **default** when the env var is unset, so CI parity, manual invocations, and the post-builder quality gate are byte-for-byte unchanged.

**2. Bounded, visible load-aware deferral in `run_gate_tick()`** — reusing `cpu_headroom`'s load probe (`read_loadavg_1m` / `logical_cpu_count`, factored behind a shared `is_host_saturated` helper):
- load below threshold, **or no load data (fail safe)** → run the **full** tier;
- saturated **within** the max-defer window → **defer** (no command runs; a deferral is NOT an evaluation — it does not record the SHA memo, does not arm the indeterminate backoff, and leaves the halt flag untouched);
- saturated **past** the max-defer window → run the **fast** tier regardless of load, so a permanently-loaded host still reaches a verdict.

Deferral is **bounded** (default 30 min) so it can't re-disable the gate the way the timeout did — a genuinely red `main` is still caught within one such cycle, and a fast-tier Red halts dispatch with the same semantics as a full-tier Red. `loom-daemon status` renders `deferred (load …)` distinct from `not evaluated (timeout …)`, and labels a fast-tier verdict (`clear(fast)` / `… [fast tier — compile+smoke only, NOT a full-suite green]`) so it is never mistaken for a full-suite green.

## Acceptance criteria

- [x] On a host at the dispatch cap the gate produces a real Green/Red verdict **or** an explicit `deferred (load …)` status — never a 1200s-timeout `not evaluated` steady state (bounded deferral + fast tier).
[...]

PR #4414, opened 2026-07-29
fix(daemon): downgrade dead-tmux-server restore log to debug when there is nothing to restore

### LINKED ISSUE #4378: daemon boot: restore_from_tmux logs ERROR siren twice on hosts with no tmux pool state
## Context
Every `loom-daemon` boot on a host that has never run the Manual-Orchestration-Mode tmux pool logs, at ERROR level with siren emoji:

```
[ERROR] 🚨 TMUX SERVER DEAD during restore_from_tmux - Socket should be at /private/tmp/tmux-$UID/loom
[ERROR] 🚨 TMUX SERVER DEAD during restore_from_tmux - Socket should be at /private/tmp/tmux-$UID/loom
```

(observed twice per boot, every boot, 2026-07-28/29, fresh macOS host, launchd-supervised daemon, zero tmux terminals ever created — "Restored 0 terminals"). A missing tmux server is the *expected* state on an autonomous-only host; ERROR-level siren noise trains operators to ignore the error channel, which is exactly where real failures need to land.

- [ ] When there is no persisted terminal/pool state to restore (0 configured-or-restorable terminals), `restore_from_tmux` skips the socket probe entirely or logs at debug.
- [ ] A genuinely dead tmux server with restorable terminal state may keep the ERROR.
- [ ] No more ERROR lines on a clean autonomous-only boot.

rjwalters:
## Summary

`restore_from_tmux_with_filter` logs an ERROR-level 🚨 siren every time `tmux -L loom list-sessions` fails with "no server running" — even when there is nothing to restore in the first place. On a host that has never run the Manual Orchestration Mode tmux pool (e.g. autonomous-only), this fires **twice** on every daemon boot: once from the eager boot-time `restore_from_tmux()` call, and again from `list_terminals()`'s lazy re-restore (triggered because the registry is still empty after the first failed attempt). A dead server at restore time implies zero live sessions by definition, so this is the expected steady state, not an anomaly — but the ERROR-level noise trains operators to ignore the error channel, which is exactly where real failures need to land.

## Changes

- Extracted the stderr classification into a new pure helper, `TerminalManager::is_expected_empty_restore`, which recognizes the `"no server running"` condition.
- In `restore_from_tmux_with_filter`, check this classification **before** delegating to `handle_tmux_error`: if the server is dead, log at `debug` and return `Ok(())` early instead of routing through the ERROR siren. This mirrors the existing `clean_stale_sessions` precedent in the same file, which already treats the identical stderr pattern as a silent `Ok(0)`.
- `handle_tmux_error` itself is untouched — its other call sites (`new`, `has_tmux_session`, `attach_to_session`, `kill_session`) operate on sessions the daemon believes are live, where a dead server genuinely is anomalous, and its `bool` return still drives control flow there. All of its existing unit tests pass unchanged.
- Added three unit tests for the new helper: `no server running` classifies as expected-empty; `no sessions` (a distinct, already-separately-handled condition) does not; an arbitrary other tmux error does not.

| 0 configured-or-restorable terminals: skip the socket probe entirely or log at debug | ✅ | `restore_from_tmux_with_filter` still probes (`list-sessions`), but a "no server running" failure now logs at `debug` unconditionally instead of ERROR — a dead server at restore time means zero sessions were ever restorable, so there is no distinct "configured" case to special-case (see rationale below) |
| A genuinely dead tmux server with restorable terminal state may keep the ERROR | ✅ (permissive "may", satisfied by design) | A dead server at restore time implies zero live sessions to restore *from tmux*, regardless of what a workspace config declares — retaining the ERROR via a configured-terminals heuristic would re-create the false alarm on every reboot of a normal MOM host (launchd starts the daemon before any tmux session exists). AC 2 is worded permissively, and downgrading uniformly satisfies AC 1 and AC 3 without that regression. |
| No more ERROR lines on a clean autonomous-only boot | ✅ | Both restore call sites (eager boot-time `restore_from_tmux()` and `list_terminals()`'s lazy re-restore) route through the same `is_expected_empty_restore` check, so both no longer emit ERROR when the tmux server is absent |

## Test Plan

- `cargo check` / `cargo clippy --lib -- -D warnings` — clean.
- `cargo fmt -- --check src/terminal.rs` — clean.
- `cargo test --lib terminal::` — all 52 tests in the `terminal` module pass, including the 3 new `is_expected_empty_restore` tests.
- Manual reasoning: with no `-L loom` tmux server running, `restore_from_tmux_with_filter` now hits the new early-return branch and logs at `debug` instead of calling `handle_tmux_error`, eliminating both ERROR lines per boot described in the issue.

`cargo test` at the workspace level currently fails to *compile* (test-only code) due to an unrelated, already-tracked regression: #4412 (`serve.rs`'s `empty_report()` test helper is missing three `DaemonStatusReport` fields added by a near-simultaneous PR, #4297 × #4404 semantic conflict). This blocks running the full `cargo test` suite in this worktree; I verified the `terminal` module in isolation by locally reapplying #4412's documented one-line fix to `serve.rs`, running `cargo test --lib terminal::` (52/52 passed, see Test Plan), then reverting that unrelated file before committing — `src/serve.rs` has no changes in this PR's diff.

Closes #4378

### DIFF
--- a/loom-daemon/src/terminal.rs
+++ b/loom-daemon/src/terminal.rs
@@ -1095,6 +1095,18 @@ impl TerminalManager {
         log::info!("Set CLAUDE_CONFIG_DIR={config_dir_str} for session {tmux_session}");
     }
 
+    /// Classify a tmux `list-sessions` stderr string encountered during
+    /// `restore_from_tmux` as an *expected* "nothing to restore" condition.
+    ///
+    /// A dead tmux server (`no server running`) at restore time implies zero
+    /// live sessions by definition — there is nothing restorable *from tmux*,
+    /// which is the normal state on a host that has never run the Manual
+    /// Orchestration Mode tmux pool (e.g. autonomous-only), and also the
+    /// normal state immediately after a reboot. See #4378.
+    fn is_expected_empty_restore(stderr: &str) -> bool {
+        stderr.contains("no server running")
+    }
+
     /// Handle tmux command errors with consistent logging
     /// Returns true if the error indicates the tmux server is dead
     fn handle_tmux_error(stderr: &str, operation: &str) -> bool {
@@ -1644,6 +1656,16 @@ impl TerminalManager {
         // Enhanced logging: Check for tmux server failure
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
+
+            // Issue #4378: log the expected "nothing to restore" case at
+            // debug instead of routing it through `handle_tmux_error`'s
+            // ERROR siren, mirroring the existing `clean_stale_sessions`
+            // precedent for the same stderr pattern.
+            if Self::is_expected_empty_restore(&stderr) {
+                log::debug!("No tmux server during restore_from_tmux — nothing to restore");
+                return Ok(());
+            }
+
             Self::handle_tmux_error(&stderr, "restore_from_tmux");
 
             // Return early with empty list if server is dead
@@ -2079,6 +2101,27 @@ mod tests {
     #[test]
     fn test_validate_terminal_id_rejects_dots() {
         assert!(TerminalManager::validate_terminal_id("terminal.1").is_err());
+    }
+
+    // ===== is_expected_empty_restore tests (#4378) =====
+
+    #[test]
+    fn test_is_expected_empty_restore_no_server_running() {
+        assert!(TerminalManager::is_expected_empty_restore(
+            "no server running on /private/tmp/tmux-501/loom"
+        ));
+    }
+
+    #[test]
+    fn test_is_expected_empty_restore_no_sessions_is_not_expected() {
+        // "no sessions" is a distinct condition already handled separately by
+        // `handle_tmux_error` (debug, not error) — it must not be reclassified here.
+        assert!(!TerminalManager::is_expected_empty_restore("no sessions"));
+    }
+
+    #[test]
+    fn test_is_expected_empty_restore_other_error_is_not_expected() {
+        assert!(!TerminalManager::is_expected_empty_restore("some other tmux error"));
     }
 
     // ===== handle_tmux_error tests =====


rjwalters:
**Approved.**

## Correctness
- `is_expected_empty_restore` is a small, pure, well-named helper. Its placement in `restore_from_tmux_with_filter` — checked *before* `handle_tmux_error` — correctly short-circuits the ERROR-level siren for the "no server running" case while leaving `handle_tmux_error`'s own call sites (`new`, `has_tmux_session`, `attach_to_session`, `kill_session`) untouched, which is the right scope: a dead server there is genuinely anomalous (the daemon believes a session is live), whereas at restore time it's the expected steady state.
- Verified against the code the PR's stated precedent: `clean_stale_sessions` (line ~1804) already treats `"no server running" || "no sessions"` as a silent `Ok(0)`, so this change brings `restore_from_tmux_with_filter` in line with existing conventions rather than inventing a new one.
- Confirmed both restore call sites (`restore_from_tmux` → `restore_from_tmux_with_filter`, and `list_terminals()`'s lazy re-restore path) route through the same check, so both ERROR lines described in #4378 are eliminated.
- The three new unit tests correctly cover the boundary: `"no server running"` → true, `"no sessions"` (a distinct, already-separately-handled condition) → false, and an arbitrary other error → false.

## Testing
- Diff is scoped entirely to `loom-daemon/src/terminal.rs` (43 lines added, 0 removed) — no `serve.rs` changes.
- Independently reproduced and confirmed the pre-existing, already-tracked #4412 compile break on bare `origin/main` (separate worktree, `cargo check --workspace --all-targets` → same `E0063` at `serve.rs:328`, missing the three `main_health_gate_*` fields from #4297 × #4404's semantic conflict). This PR's own CI failures (`Rust Linting & Build Check`, `Rust Unit Tests`) reproduce the identical `E0063` at the identical line — confirmed via the job log — so they are not caused by this PR's diff.
- To get a real (not just asserted) test signal for this PR's own code: temporarily patched the three missing fields into `serve.rs`'s `empty_report()` in the worktree (never committed, reverted immediately after), then ran `cargo test --lib terminal::` — **52/52 passed**, including all 3 new `is_expected_empty_restore` tests. `cargo fmt -- --check src/terminal.rs` is clean.

## Process notes
- `mergeStateStatus` is `UNSTABLE`/CI shows `fail` on the two Rust jobs — expected and attributable entirely to #4412 (still open, not yet fixed on main), not to this PR. Approving despite red CI here per the documented pre-existing/unrelated-and-independently-confirmed exception; this PR should auto-unblock once #4412 lands and CI is re-run/rebased.
- Issue linking: PR body ends with `Closes #4378` — correct.

Nice, tightly scoped fix with good test coverage and an honest writeup of the environmental blocker.

rjwalters:

- **CI Status Check**: `Rust Linting & Build Check` and `Rust Unit Tests` are failing.

The failure is **inherited from broken main**, not this PR's diff: `error[E0063]: missing fields main_health_gate_deferred, main_health_gate_deferred_reason and main_health_gate_verdict_tier in initializer of types::DaemonStatusReport` — this is exactly #4412 (the #4404 × #4297 semantic conflict), which is currently `loom:building`.

**Next steps:**
- Wait for #4412's fix to merge to main
- Rebase this branch (or push a merge of main) to re-trigger CI — a green main alone will not re-run this PR's checks

---
*Automated by Champion role*

PR #4421, opened 2026-07-29
fix(guards): expand unquoted leading ~/ before Bash write-target resolution

### LINKED ISSUE #4382: guards: unexpanded ~/ in a Bash write target resolves as repo-relative — false-positive deny on writes outside the checkout
## Problem

The #4178 Bash-write guard resolves an **unexpanded tilde path as repo-relative**, producing a false-positive deny on writes that are actually far outside the repository.

Observed 2026-07-29 (operator session, main checkout cwd) during incident recovery:

    cp /Users/rwalters/GitHub/loom/target/release/loom-daemon ~/.local/bin/loom-daemon

    BLOCKED: Bash-tool write to '/Users/rwalters/GitHub/loom/~/.local/bin/loom-daemon' resolves to
    the main repository checkout ... worktree-isolation bypass via Bash redirection/tee/sed -i/cp/mv

The guard's static parser treated the literal `~/.local/bin/loom-daemon` as a relative path and prefixed the cwd — but the shell would expand `~` to `$HOME`, making the true destination `/Users/rwalters/.local/bin/loom-daemon`, outside the checkout entirely. Retrying with the absolute path passed. Same fix family as #4245/#4289 (parser fidelity vs shell semantics).

Note the flip side: naive tilde-expansion in the guard must not create a false-NEGATIVE hole (e.g. `\~/...` or `HOME=<repo> ...` games). Expansion should mirror shell rules for the unquoted-leading-tilde case only.

- [ ] A leading unquoted `~/` (and `~user/`) in a write-target argument is expanded per shell semantics before path resolution; the cp above is allowed from a main-checkout cwd.
- [ ] Quoted/escaped tildes (`'~/x'`, `\~/x`) — which the shell does NOT expand — remain treated as literal relative paths (current behavior).
- [ ] Regression tests for both directions in the guard test suite.

## Related

- #4178 (the Bash-write confinement guard) · #4245/#4289 (quote-aware `>` scanning — the precedent fix in the same parser)

rjwalters:
## Summary

The #4178 Bash-write confinement guard (`guard-destructive-generic.sh`) resolves an unexpanded tilde path as repo-relative, producing a false-positive deny on writes that actually land far outside the repository.

Reported incident (#4382): `cp <built-binary> ~/.local/bin/loom-daemon` from a main-checkout cwd was denied as a "worktree-isolation bypass", even though the real destination (`$HOME/.local/bin/loom-daemon`) is nowhere near the checkout. `extract_write_targets()` is a whitespace/quote-aware tokenizer, not a shell evaluator, so it never performed the tilde expansion a real shell does before `cp`/`mv`/`tee`/`sed -i`/redirection ever see the argument.

## Fix

Adds `expand_leading_tilde()` in `defaults/hooks/guard-destructive-generic.sh` (same fix family as #4245/#4289's quote-aware `>` scanning), applied to each extracted write target immediately before the existing relative-path resolution step:

- An **unquoted, unescaped leading `~/`** (or bare `~`) expands to `$HOME`.
- **`~user/`** resolves via `getent` (Linux) / `dscl` (macOS), with the username passed as a plain CLI argument (never `eval`'d/interpolated), so a hostile username token can't inject a command. If the user can't be resolved, the token is left untouched (falls back to the existing repo-relative treatment -- fail-safe toward deny, never toward a new silent allow).
- **Quoted (`'~/x'`) and backslash-escaped (`\~/x`) tildes are left untouched.** `qsplit()` (the shared tokenizer, #3755) preserves quote/escape characters verbatim, so these never match the leading-tilde case and keep their current (correct) literal/repo-relative treatment -- no regression.
- A tilde that isn't the **first character** of the token (e.g. `foo~/bar`) is never touched either.
- The guard reads its **own process `$HOME`**, not the command text, so an inline `HOME=<repo> cmd ~/x` game in the analyzed command string cannot redefine expansion -- this mirrors real bash, where a same-line `VAR=value command` prefix only changes the *child* command's environment and never affects tilde expansion of that same command line.

`.loom/hooks/guard-destructive-generic.sh` (the installed copy) is synced byte-for-byte from `defaults/hooks/guard-destructive-generic.sh` (the source of truth), per this repo's install/sync convention.

## Testing

Added 8 regression tests to `tests/hooks/test-guard-destructive.sh` under a new "Tilde expansion for write targets (#4382)" section, covering both directions from the acceptance criteria:
- unquoted `~/...` and bare `~` expand and allow when they land outside the checkout
- `~user/...` (current user) resolves via the passwd db and allows
- expansion landing *inside* the checkout (`HOME=<repo>`) still denies -- proves this isn't a blanket `~` allow
- single-quoted and backslash-escaped leading tildes stay literal and still deny (regression)
- a non-leading tilde (`backup~/f.sh`) is not an expansion case and still denies
- an unresolvable `~nonexistentuser/` falls back to literal/deny

Full suite: `bash tests/hooks/test-guard-destructive.sh` -- **490/490 passing**.

`shellcheck --severity=warning` clean on the modified file (one SC2088 false-positive on the intentional literal `~`-prefixed case pattern is explicitly suppressed with a rationale comment).

Closes #4382

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -1888,6 +1888,95 @@ normalize_abs_path() {
     fi
 }
 
+# =============================================================================
+# expand_leading_tilde() — shell-accurate tilde expansion for write targets
+# (#4382, same fix family as the quote-aware `>` scanning of #4245/#4289).
+#
+# extract_write_targets() (below) is a plain-whitespace/quote-aware TOKENIZER,
+# not a shell evaluator — it never performs word expansions (tilde, variable,
+# glob, ...). A raw token like `~/.local/bin/x` was therefore resolved as a
+# REPO-RELATIVE path (cwd-prefixed) even though the real shell would expand it
+# to "$HOME/.local/bin/x" before `cp`/`mv`/`tee`/`sed -i`/redirection ever see
+# it — producing a false-positive worktree-confinement deny on a write that
+# actually lands far outside the repo (#4382).
+#
+# This performs ONLY the narrow, unambiguous piece of shell word-expansion
+# tilde-expansion applies to: an UNQUOTED, UNESCAPED tilde as the FIRST
+# character of the token, i.e. exactly the shell-eligible positions:
+#   ~/rest        -> "$HOME/rest"
+#   ~             -> "$HOME"
+#   ~user/rest    -> "<user's home>/rest"   (only if that user resolves)
+#   ~user         -> "<user's home>"
+#
+# Because qsplit() (the shared tokenizer, #3755) copies a quoted span
+# VERBATIM including its quote characters, and leaves a literal backslash
+# untouched, a token whose raw text does not start with a bare `~` was NOT
+# eligible for shell tilde-expansion and MUST stay untouched here:
+#   '~/x'   -> starts with a quote char, shell never expands it (stays literal)
+#   \~/x    -> starts with a literal backslash, shell never expands it either
+#   foo~/x  -> tilde is not the leading character -- not an expansion position
+# Any of these three cases falls through unchanged (echoed back as-is), which
+# preserves the existing (correct) repo-relative/deny behavior for them.
+#
+# `~user` lookup uses getent (Linux) / dscl (macOS) with the username passed
+# as a plain CLI argument (never eval'd/interpolated into a shell string) so a
+# hostile username token cannot inject a command. If the user cannot be
+# resolved on this host, the token is returned UNCHANGED (falls back to the
+# existing repo-relative treatment) -- consistent with this file's fail-open
+# contract: uncertainty here biases toward the (safe) deny path, never toward
+# a silent new allow.
+# =============================================================================
+expand_leading_tilde() {
+    local tok="$1"
+    # shellcheck disable=SC2088 # intentional: these `~`-prefixed case
+    # patterns are literal PATTERN matches against an unexpanded leading
+    # tilde in $tok (the whole point of this function), not an attempt at
+    # shell tilde expansion.
+    case "$tok" in
+        '~')
+            [[ -n "$HOME" ]] && { printf '%s' "$HOME"; return; }
+            printf '%s' "$tok"
+            return
+            ;;
+        '~/'*)
+            if [[ -n "$HOME" ]]; then
+                printf '%s' "$HOME/${tok#\~/}"
+            else
+                printf '%s' "$tok"
+            fi
+            return
+            ;;
+        '~'*)
+            local rest="${tok#\~}"
+            local user="${rest%%/*}"
+            local remainder=""
+            if [[ "$rest" == */* ]]; then
+                remainder="/${rest#*/}"
+            fi
+            if [[ -n "$user" ]]; then
+                local home=""
+                if command -v getent >/dev/null 2>&1; then
+                    home=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
+                elif command -v dscl >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


rjwalters:
## Approved

Reviewed against #4382's acceptance criteria. The fix is correct, minimal, and — importantly for guard code — verified not to open a false-negative hole in the #4178 worktree-confinement deny path.

### Verification performed

**Incident repro (the actual AC):** ran both guards against the reported command from a main-checkout cwd with the real `$HOME`:
- `main`'s guard: `BLOCKED: Bash-tool write to '/Users/.../loom/~/.local/bin/loom-daemon' resolves to the main repository checkout` (the bug)
- this branch's guard: exit 0, no decision (allow). Fixed.

**Test suite:** the 8 new `#4382` cases all pass. Base suite is 482/482 in the real checkout; running base vs. branch side-by-side in an isolated tree gave an **identical** failure set (env artifacts of the isolated tree, present on both sides) — i.e. **zero regression delta**, and 490 total. `shellcheck --severity=warning` clean, `bash -n` clean.

**Adversarial edge cases probed directly against the branch guard** (fixture repo + managed worktree, beyond what the suite covers) — every one lands on the shell-accurate verdict:

| Case | Verdict | Correct? |
|---|---|---|
| `cp x "~/defaults/hooks/f.sh"` (**double-quoted** — untested in the suite) | deny | yes — shell does not expand, stays repo-relative |
| `~+/…` / `~-/…` (bash dirstack forms) | deny | yes — unresolvable via passwd, falls back to literal |
| `HOME=` (empty) with `~/…` | deny | yes — no expansion, no new allow |
| `cp x ~/../defaults/hooks/f.sh` (traversal back into the repo) | deny | yes — `normalize_abs_path()` resolves before the containment test |
| `> ~/x`, `tee ~/x`, `sed -i … ~/x`, `cd <worktree> && cp … ~/x` | expanded consistently | yes — the hook sits at the single consumer loop, so every write idiom gets it |
| absolute in-repo path (control) | deny | yes — confinement unchanged |

The design choice that makes this safe is applying expansion **after** `extract_write_targets()` (whose tokens preserve quotes/backslashes verbatim via `qsplit()`) and **before** the relative-path resolution — so `'~/x'`, `"~/x"`, `\~/x`, and `foo~/x` never reach the `~`-leading `case` arms at all. `~user` lookup passes the username as a plain argv to `getent`/`dscl` (no `eval`, no interpolation), and every unresolvable path returns the token unchanged, which biases to the existing deny rather than a silent allow. The `HOME=<repo>` test correctly pins that this is not a blanket `~` allow, and the guard reading its own `$HOME` rather than the command text matches real bash (a same-line `VAR=value cmd` prefix does not affect that line's tilde expansion).

`.loom/hooks/` and `defaults/hooks/` copies are byte-identical (same blob sha), matching the convention every prior commit to this file followed.

### Non-blocking observations (no action required for this PR)

1. `tests/hooks/*` is not wired into any CI workflow, so the suite ran only locally — pre-existing gap, not introduced here, and worth a separate issue if it isn't already tracked.
2. The **double-quoted** `"~/x"` case is the one shell-doesn't-expand form the new tests do not assert (single-quoted and backslash-escaped are covered). I confirmed it denies correctly; a third assertion would make the quoting matrix exhaustive.
3. `~-/…` from a worktree whose `OLDPWD` is outside the repo remains a residual false-positive of the same family as #4382 — vanishingly rare and it fails safe, so not worth chasing.
4. macOS `dscl … | awk '{print $2}'` truncates a home directory containing a space; the resulting prefix is still outside the checkout, so the verdict does not change.

CI: all checks pass (Rust/MCP jobs correctly skipped by the path filter on a shell-only change). Merge state: `CLEAN` / `MERGEABLE`; `git merge-tree` confirms no conflicts. PR body carries `Closes #4382`.

PR #4400, opened 2026-07-29
fix(daemon): stop misdiagnosing prior-boot heartbeats as a wedge

rjwalters:
## Summary

`loom-daemon status` (and the host-side watchdog) could misdiagnose a healthy,
still-booting daemon as "likely wedged" because a stale heartbeat file was
being compared only against a fixed staleness threshold, never against the
live process's own age. Two 2026-07-28 incidents reproduced this: a
migrated/leftover heartbeat file from a previous boot (83814s old) rendered as
"STALE ... the daemon is likely wedged" against a daemon that had just started
seconds earlier, and a separate incident printed the imperative "Do NOT run
loom-daemon-start.sh ... stop && start" remediation even when the evidence
didn't actually indicate a fault.

AC 1 (the pre-IPC startup-grace window, `InstallState::AliveStarting`) was
already implemented on `main` via #4213/PR #4250 before this issue was filed —
verified against current `main` per the curator's note, and left untouched.
This PR implements the remaining two acceptance criteria plus watchdog parity.

## Changes

- `loom-daemon/src/daemon_install_state.rs`: add
  `HeartbeatFreshness::PriorBoot` (`as_str()` → `"prior-boot"`). `check_heartbeat`
  now takes the live process's own age and classifies a heartbeat file whose
  mtime is strictly older than the process's start time as `PriorBoot` —
  checked *before* the staleness threshold, so even an otherwise-"fresh"-looking
  heartbeat that predates the process is still flagged. Equal ages are
  deliberately **not** `PriorBoot` (current-boot evidence). An unparseable
  process age (`ps` failure) makes **no** prior-boot claim and degrades to the
  pre-existing Stale/Fresh verdicts. `classify()` is split into a testable
  `classify_with_process_age_fn` so unit tests can pin a deterministic
  synthetic process age instead of depending on the real (and highly variable)
  uptime of the shared test-binary process.
- `loom-daemon/src/main.rs`: the `AliveButUnresponsive` render arm now (a)
  prints a dedicated "heartbeat file is from a PREVIOUS boot ... not evidence
  about the current process" message for `PriorBoot`, and (b) gates the
  imperative stop/start remediation to only print for a genuine current-boot
  `Stale` verdict — `Fresh`/`Unknown`/`PriorBoot` now print inspect-first
  guidance (`ps -p <pid> -o pid,etime,command`, `loom-daemon status --json`)
  instead. The `--json` surface picks up `"prior-boot"` automatically via the
[...]

PR #4439, opened 2026-07-29
test(daemon): pin synthetic process age in 6 flaky install-state tests

### LINKED ISSUE #4406: Flaky daemon_install_state tests: 5 tests misclassify as AliveStarting when test binary is <1s old
## Bug Description

Five pre-existing `daemon_install_state` unit tests fail intermittently (~40% of runs locally) when the test binary process is less than 1 second old at test execution time:

- `marker_present_live_pid_is_alive_but_unresponsive`
- `env_launchd_override_forces_pid_file_path_even_when_marker_says_launchd`
- `heartbeat_absent_degrades_to_unknown_not_stale`
- `marker_missing_fields_uses_fallback_paths`
- `non_darwin_forces_pid_file_path_regardless_of_marker`

All fail the same way: `classify()` returns `AliveStarting` where the test expects `AliveButUnresponsive`.

## Root Cause

`no_env_overrides()` pins `startup_grace_secs_override: Some(0)` with the comment "A zero grace window makes the test process (always older than 0s) fall through" — but that assumption is wrong. `ps -o etime=` reports `00:00` (parsed as 0 seconds) for any process under 1 second old, so the grace check `age (0) <= grace_threshold (0)` fires and the startup-grace verdict wins.

These 5 tests use the pid of the test binary itself (`std::process::id()`) via the real `ps`-backed `process_age_secs` lookup, so whether they pass depends on whether the whole `cargo test --lib daemon_install_state` scoped run happens inside the binary's first wall-clock second. A full-suite run is usually slow enough to escape the window, which is why this mostly bites scoped/filtered runs — exactly what the Judge's scoped test execution does.

## Reproduction

```bash
cd loom-daemon
for i in 1 2 3 4 5 6; do cargo test --lib daemon_install_state 2>&1 | grep "^test result"; done
# Intermittent: "FAILED. 24 passed; 5 failed" — all 5 with left: AliveStarting / right: AliveButUnresponsive
```

Reproduced 2026-07-29 during Judge review of PR #4400 (which did NOT introduce this — the grace path and these 5 tests are byte-identical to main; the flake dates to the #4213/#4250 startup-grace work).

## Suggested Fix

PR #4400 introduced exactly the right tool: `classify_with_process_age_fn`, which lets a test pin a deterministic synthetic process age. Migrate these 5 tests to it with a pinned age comfortably above 0 (e.g. `|_| Some(1_000_000)`), the same pattern #4400 applied to the tests it touched. Alternatively (or additionally), fix the `no_env_overrides()` comment so the broken "always older than 0s" assumption isn't copied again.

## Impact

- **Severity**: Low-Medium — no production impact, but CI "Rust Unit Tests" can flake on it, and scoped local runs flake often enough to erode trust in real failures.
- **Workaround**: re-run; or run the full suite instead of a scoped filter.

---

## Curator Enhancement (2026-07-29)

All claims verified against `origin/main`: `classify_with_process_age_fn` exists (`loom-daemon/src/daemon_install_state.rs:622`), `no_env_overrides()` pins `startup_grace_secs_override: Some(0)` (line ~736), and the pinned-age migration pattern `|_| Some(1_000_000)` is already in use by the tests PR #4400 touched (line ~882).

**Enumeration correction — there is a 6th affected test.** `loom-daemon/src/autonomy_marker.rs` (~lines 350–373, in the `heal_marker` test that ends by calling `daemon_install_state::classify`) uses the identical flaky pattern: `std::process::id()` + real `ps`-backed age lookup + inline `EnvOverrides` with `startup_grace_secs_override: Some(0)` + `assert_eq!(report.state, AliveButUnresponsive)`. It flakes to `AliveStarting` under the same sub-1-second condition. Note `classify_with_process_age_fn` is currently private (`fn`, not `pub fn`) in `daemon_install_state.rs` — migrating the `autonomy_marker.rs` call site requires making it `pub(crate)`.

### Recommended approach

Do both halves of the Suggested Fix: migrate all **6** tests to `classify_with_process_age_fn` with a pinned age (`|_| Some(1_000_000)`), AND fix the `no_env_overrides()` comment so the wrong "always older than 0s" assumption isn't copied into future tests.

## Acceptance Criteria

- [ ] All 6 tests (the 5 enumerated + the `autonomy_marker.rs` one) use a pinned synthetic process age instead of the real `ps`-backed lookup
- [ ] The misleading comment in `no_env_overrides()` is corrected (a 0-second-old process does NOT fall through a zero grace window — `ps -o etime=` reports `00:00`)
- [ ] The reproduction loop (6× scoped `cargo test --lib daemon_install_state`) passes consistently
- [ ] No production (non-`#[cfg(test)]`) code paths change except an optional visibility widening of `classify_with_process_age_fn` to `pub(crate)`

## Affected Files

- `loom-daemon/src/daemon_install_state.rs` — migrate the 5 enumerated tests to `classify_with_process_age_fn` with pinned age; fix the `no_env_overrides()` comment; widen `classify_with_process_age_fn` to `pub(crate)`
- `loom-daemon/src/autonomy_marker.rs` — migrate the 6th flaky call site (~line 368) to the pinned-age variant

## Test Plan

- [ ] Manual verification: run the issue's Reproduction loop (6 scoped runs) — all green; previously ~40% flake locally
- [ ] Automated tests: the change IS the tests; full `cargo test -p loom-daemon` passes
- [ ] Edge cases: confirm the tests that intentionally exercise the startup-grace path (`AliveStarting` expectations, if any) still use ages inside the grace window and still pass

*(Heads-up: as of curation time, `cargo check --workspace --all-targets` fails on bare main for an unrelated reason — #4412, urgent, one-line fix in `serve.rs`. If your worktree's tests don't compile before you change anything, that's #4412, not this issue.)*

rjwalters:
Closes #4406

## Problem

Six `loom-daemon` unit tests asserted the post-startup-grace verdict
`AliveButUnresponsive` while letting `classify` resolve the *test binary's own*
process age through the real `ps`-backed lookup.

`ps -o etime=` reports `00:00` for any process under one wall-clock second old,
which `parse_etime` turns into `0`. The startup-grace check is
`age <= grace_threshold`, and the shared `no_env_overrides()` fixture pins
`startup_grace_secs_override: Some(0)` — so `0 <= 0` fires and the verdict is
`AliveStarting`, not `AliveButUnresponsive`.

Whether a run passed therefore depended on whether the scoped
`cargo test --lib daemon_install_state` invocation happened to land inside the
test binary's first second (~40% locally). The fixture's comment actively
encoded the wrong assumption ("the test process (always older than 0s) fall[s]
through"), which is how the pattern got copied six times.

## Changes

All test-only, except one visibility widening.

**`loom-daemon/src/daemon_install_state.rs`**
- Migrated the 5 enumerated tests from `classify` to
  `classify_with_process_age_fn(..., |_| Some(1_000_000))` — the same pinned-age
  pattern PR #4400 introduced and already uses for the tests it touched:
  - `marker_present_live_pid_is_alive_but_unresponsive`
  - `marker_missing_fields_uses_fallback_paths`
  - `heartbeat_absent_degrades_to_unknown_not_stale`
  - `env_launchd_override_forces_pid_file_path_even_when_marker_says_launchd`
  - `non_darwin_forces_pid_file_path_regardless_of_marker`
- Corrected the `no_env_overrides()` comment: a zero grace window is *not*
  sufficient on its own, and any test wanting a post-grace verdict must also
  pin a synthetic age.
- Corrected the same stale claim in
  `process_beyond_grace_falls_through_to_fault_verdict`'s comment (that test
  already pins an age; its comment still said "the (always older) test
  process").
- Widened `classify_with_process_age_fn` from `fn` to `pub(crate) fn` so a
  sibling module's test can reach it (documented as having no production caller
  other than `classify`).

**`loom-daemon/src/autonomy_marker.rs`**
- Migrated the 6th call site,
  `healed_marker_is_parseable_by_the_install_state_probe`, to the pinned-age
  variant. (Found by the Curator; it has the identical
  `std::process::id()` + `startup_grace_secs_override: Some(0)` +
  `assert_eq!(..., AliveButUnresponsive)` shape.)

No production code path changes. `classify` remains the only production caller
of `classify_with_process_age_fn`; the grace logic itself is untouched.

## Test Plan

- `cargo test --lib daemon_install_state` x8 consecutive scoped runs (the
  issue's reproduction loop, extended): **29 passed, 0 failed** every time.
- `cargo test --lib autonomy_marker` x4: **7 passed, 0 failed** every time.
- `cargo fmt --check`: clean.
- `cargo test` (full `loom-daemon` suite): **1825 passed, 4 failed**.

### About those 4 failures — pre-existing, unrelated, and reproduced on bare main

The 4 failures are all in `main_health_gate::tests::test_run_gate_tick_*`. They
are **not** caused by this change:

- Reproduced identically on the **unmodified main checkout** (`d23ef3ca`),
  same 4 tests, same assertions (`first tick must run and be green`).
- Root cause is host load, not code: `run_gate_tick` consults the real
  1-minute load average via `cpu_headroom::is_host_saturated`, and this host
  was at load 45 across 28 cores (1.6/core) from parallel cargo builds — so
  the gate took the `GateTierDecision::Defer` path and returned `None` instead
  of `Some(GateOutcome::Green { .. })`.
- Confirmed by running the same 4 tests scoped with
  `LOOM_BUILD_GATE_LOAD_THRESHOLD=1000`: **4 passed, 0 failed**.

Notably this is the *same family* of defect as #4406 — a test asserting a
verdict while consulting real host state instead of an injected value — but a
different mechanism (load average vs. process age), in a module outside this
issue's scope. Flagging it here rather than expanding scope.

## Acceptance Criteria

- [x] All 6 tests use a pinned synthetic process age instead of the real
      `ps`-backed lookup
- [x] The misleading `no_env_overrides()` comment is corrected
- [x] The reproduction loop passes consistently (8/8 scoped runs green)
- [x] No production (non-`#[cfg(test)]`) code paths change except the
      `pub(crate)` visibility widening of `classify_with_process_age_fn`
- [x] Tests that intentionally exercise the startup-grace path still pass —
      `young_process_within_grace_is_alive_starting` (huge grace window, real
      age, deliberately unmigrated since a real age *inside* a `u64::MAX`
      window is exactly its intent) and
      `heartbeat_older_than_process_start_is_prior_boot_end_to_end`
      (injected 9s vs. 0s grace) are both untouched and green.

### DIFF
--- a/loom-daemon/src/autonomy_marker.rs
+++ b/loom-daemon/src/autonomy_marker.rs
@@ -365,7 +365,16 @@ mod tests {
             startup_grace_secs_override: Some(0),
             is_darwin: false,
         };
-        let report = crate::daemon_install_state::classify(dir.path(), &marker, &env);
+        // Pin a synthetic process age well beyond the zero grace window
+        // (#4406): the real `ps -o etime=` lookup reports `00:00` for a
+        // sub-second-old test binary, which parses to 0 and satisfies
+        // `age <= grace`, flaking this assertion to AliveStarting.
+        let report = crate::daemon_install_state::classify_with_process_age_fn(
+            dir.path(),
+            &marker,
+            &env,
+            |_| Some(1_000_000),
+        );
         // Marker present + our own live pid ⇒ NOT the not-expected/dead states.
         assert_eq!(report.state, crate::daemon_install_state::InstallState::AliveButUnresponsive,);
         assert_eq!(report.started_at.as_deref(), Some("2026-07-28T00:00:00Z"));
--- a/loom-daemon/src/daemon_install_state.rs
+++ b/loom-daemon/src/daemon_install_state.rs
@@ -616,10 +616,14 @@ pub fn classify(loom_dir: &Path, marker_path: &Path, env: &EnvOverrides) -> Inst
 /// [`classify`]'s implementation, parameterized over the process-age lookup.
 /// Production code always goes through [`classify`] (which passes the real
 /// `ps`-backed [`process_age_secs`]); tests use this directly with a fixed
-/// closure so heartbeat-vs-process-age comparisons (#4368) are deterministic
-/// — the test binary's own real uptime, shared across every unit test
-/// running in this one process, is not a value any single test can control.
-fn classify_with_process_age_fn(
+/// closure so heartbeat-vs-process-age comparisons (#4368) and startup-grace
+/// comparisons (#4213) are deterministic — the test binary's own real uptime,
+/// shared across every unit test running in this one process, is not a value
+/// any single test can control.
+///
+/// `pub(crate)` so sibling modules' tests (e.g. `autonomy_marker`) can pin an
+/// age too; there is no production caller outside [`classify`] (#4406).
+pub(crate) fn classify_with_process_age_fn(
     loom_dir: &Path,
     marker_path: &Path,
     env: &EnvOverrides,
@@ -730,9 +734,16 @@ mod tests {
             launchd_override: None,
             launchd_label_override: None,
             heartbeat_stale_secs_override: None,
-            // A zero grace window makes the test process (always older than 0s)
-            // fall through to the existing verdicts by default — tests that want
-            // the startup-grace path set a huge window explicitly.
+            // A zero grace window is the *default* here so most tests reach the
+            // post-grace verdicts. It is NOT sufficient on its own: a real
+            // process under one second old reports `ps -o etime=` as `00:00`,
+            // which parses to 0 and satisfies `age <= grace` — so a test that
+            // asserts a post-grace verdict against a real `ps`-backed age
+            // flakes to `AliveStarting` whenever the test binary is still in
+            // its first wall-clock second (#4406). Any test that wants a
+            // post-grace verdict must ALSO pin a synthetic process age via
+            // `classify_with_process_age_fn`. Tests that want the
+            // startup-grace path set a huge window explicitly.
             startup_grace_secs_override: Some(0),
             is_darwin: false,
         }
@@ -779,7 +790,13 @@ mod tests {
                 pid_file.display()
             ),
         );
-        let report = classify(dir.path(), &marker, &no_env_overrides());
+        // Pin a synthetic process age well beyond the zero grace window
+        // (#4406) — a real `ps` age of `00:00` for a sub-second-old test
+        // binary would otherwise satisfy `age <= grace` and flake to
+        // AliveStarting.
+        let report = classify_with_process_age_fn(dir.path(), &marker, &no_env_overrides(), |_| {
+            Some(1_000_000)
+        });
         assert_eq!(report.state, InstallState::AliveButUnresponsive);
         assert_eq!(report.state.exit_code(), EXIT_ALIVE_BUT_UNRESPONSIVE);
         assert_eq!(report.pid, Some(std::process::id()));
@@ -906,7 +923,11 @@ mod tests {
             dir.path(),
             &format!("pid_file={}\nuse_launchd=false\n", pid_file.display()),
         );
-        let report = classify(dir.path(), &marker, &no_env_overrides());
+        // Pinned synthetic age (#4406, see `marker_present_live_pid_...`) so
+        // the zero grace window is genuinely cleared.
+        let report = classify_with_process_age_fn(dir.path(), &marker, &no_env_overrides(), |_| {
+            Some(1_000_000)
+        });
         assert_eq!(report.state, InstallState::AliveButUnresponsive);
         // No heartbeat file was ever written at the fallback location, so the
         // qualifier degrades to Unknown rather than falsely reporting Stale.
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Test-only flake fix, exactly as scoped by #4406. Verified:

- **Diff matches the description**: all 6 flaky tests migrated from `classify` to `classify_with_process_age_fn(..., |_| Some(1_000_000))`, following the pinned-age pattern PR #4400 established. The only non-test change is the `fn` → `pub(crate) fn` visibility widening, which is documented at the definition and needed for the `autonomy_marker` sibling-module call site. Grepped the crate: `classify` remains the sole production caller.
- **Comments corrected, not just code**: the `no_env_overrides()` fixture comment (the root cause of the copied pattern) and `process_beyond_grace_falls_through_to_fault_verdict`'s stale comment now state that a zero grace window alone is insufficient — a sub-second-old process parses `etime` `00:00` to 0 and satisfies `age <= grace`.
- **Intentionally-unmigrated tests are correct to skip**: `young_process_within_grace_is_alive_starting` (real age inside a `u64::MAX` window is its point) and the end-to-end prior-boot test (already injects an age) are untouched.

## Test Execution

1. `cargo test --lib daemon_install_state` reproduction loop — ✅ Executed 5× (plus 1 initial): 29 passed, 0 failed every run
2. `cargo test --lib autonomy_marker` — ✅ Executed: 7 passed, 0 failed
3. `cargo fmt --check` / `clippy` — ✅ Covered by the green "Rust Linting & Build Check" CI job
4. Full `cargo test` — ✅ Covered by the green "Rust Unit Tests" CI job (notably the 4 `main_health_gate` load-average failures the builder saw locally did not reproduce in CI, consistent with the host-load diagnosis)

All CI checks pass; merge state is clean. The flagged `main_health_gate` real-host-load issue is the same defect family and worth its own issue, but it's correctly kept out of scope here.

PR #4443, opened 2026-07-29
feat(daemon): ETag-cached REST listings for the hot polls — unchanged polls cost zero rate limit

rjwalters:
Closes #4428. Companion to #4440 (the #4429 breaker, merged): that one stops the bleeding during exhaustion; this one cuts steady-state consumption so exhaustion stops happening.

## Problem

The daemon's hot polls — work-finder (`loom:issue`, 60s × workspaces), epic supervisor (`loom:epic`/`loom:epic-phase`, 300s), claim reconciliation (`loom:building`, 600s) — re-fetch lists that change rarely relative to their cadence, via GraphQL `gh issue list`, which has **no conditional-request mechanism**. Every poll burned shared budget even when nothing changed; the fleet exhausted GraphQL twice on 2026-07-29 (#4432).

## Change

New `forge_listing` module: wraps `gh api --include` (keeping the zero-HTTP-client house style) with a process-lifetime ETag cache keyed by (workspace, query).

- First fetch: `200` → parse + cache `(etag, issues)`.
- Later fetches send `If-None-Match`; `304` → served from cache at **zero rate-limit cost** — verified live against the real endpoint (`x-ratelimit-remaining` unchanged across a 304, 4034 → 4034).
- Errors carry the gh stderr tail, so #4440's rate-limit classifier sees exhaustion text unchanged; the two changes compose.

Converted call sites (each filters REST's `pull_request`-marked rows to keep issue-only semantics; `--repo`/`LOOM_REPO` precedence preserved): `work_finder::GhWorkSource::list_ready_issues`, `claim_reconciliation::forge::list_building_issues`, `epic_supervisor::GhEpicSource::gh_json` (REST lowercase `state` is safe — `is_open` compares case-insensitively).

Deliberately NOT converted: PR-side claim listings (`gh pr list` — REST issue rows lack `headRefName`), quarantine reconciliation (startup-only), pipeline snapshot (status-time). Single page `per_page=100` with a truncation warning on a full page (old `--limit` ceilings were 100–200) — no silent cap.

## Expected effect

Steady-state polling on an idle fleet drops to ~zero budget: every work-finder tick that finds nothing new is a 304. Label changes are picked up within one tick (ETag miss → fresh 200).

## Testing

[...]

PR #4440, opened 2026-07-29
feat(daemon): GitHub rate-limit circuit breaker — pause forge polling until the window resets

rjwalters:
Closes #4429

## Problem

Every fleet host shares one forge identity, so the REST/GraphQL budgets are a shared, fixed resource. When they exhaust, every gh-polling loop keeps firing its full per-tick call pattern (2026-07-29 incident: 3 workspaces × 3+ calls every 60s for ~25 minutes, all failing with `GraphQL: API rate limit already exceeded`), and role sessions are spawned straight into the same wall. A rate-limited reconciliation pass is indistinguishable from 'nothing to reclaim' in its return values.

## Change

New `rate_limit_breaker` module — `host_breaker`'s six-layer shape with a simpler two-phase machine (Closed ⇄ Cooldown; a single unambiguous rate-limit failure is proof enough, no sustain counter):

- **Detection**: gh error arms feed failures to `global_observe_failure()`; a signature table (case-insensitive substrings, one-line additions) classifies rate-limit rejections vs auth/network noise. Covers GraphQL primary, REST 403, and secondary-limit phrasings.
- **Reset-aware cooldown**: on trip, one `gh api rate_limit` probe (that endpoint does not count against the quota) learns the real reset epoch; the cooldown runs to the latest exhausted resource's reset, clamped to [60s, 3600s]; a probe failure falls back to `fallbackCooldownSecs` (default 900).
- **Suppression**: while cooling, the work-finder (both loops), claim reconciliation (issue + PR sides), quarantine reconciliation, epic supervisor (both loops), and role-runner ticks skip entirely — zero gh calls, zero doomed role spawns. Lazy release on the first tick past the reset; pause/resume edges logged once each way (never per-tick).
- **Telemetry**: `loom-daemon status` gains a `GitHub rate limit:` line (cooldown source + resume deadline + last-probed budget; lifetime trip count when closed), wire field `DaemonStatusReport.rate_limit_breaker` (`serde(default)`, boxed like `host_breaker`), and a `daemon.rate_limit_breaker.state` event on each edge.
- **Config**: `autonomous.rateLimitBreaker.{enabled,fallbackCooldownSecs}`, env `LOOM_RATE_LIMIT_BREAKER` / `LOOM_RATE_LIMIT_BREAKER_FALLBACK_COOLDOWN_SECS` (env > config > default, default ON), documented in daemon-reference.md alongside hostBreaker.

## Not in this PR (tracked on the epic #4432)

[...]

Issue #4432, opened 2026-07-29
epic: survive GitHub API rate limits at fleet scale


rjwalters:
## Incident

2026-07-29 ~19:00–19:20Z: the shared personal identity's 5,000/hr GraphQL budget was exhausted and **every work-finder on every fleet host went blind simultaneously** — work_finder ticks, claim/quarantine reconciliation, and role spawns all failed until the window reset. The fleet currently cannot grow (more hosts, more workspaces, faster ticks) without hitting this wall harder: all API consumers share one fixed budget, and nearly all steady-state calls re-fetch unchanged data.

## Children (rough order)

- [x] #4429 — rate-limit-aware backoff + budget telemetry (defensive; stops the bleeding during exhaustion). **Closed COMPLETED 2026-07-29 via PR #4440** (rate-limit circuit breaker — pauses forge polling until the window resets).
- [x] #4428 — REST conditional requests / ETag caching for hot polls (biggest consumption cut; 304s are free). **Closed COMPLETED 2026-07-29 via PR #4443** (ETag-cached REST listings for work-finder / epic-supervisor / claim-reconciliation polls; 304s cost zero rate limit).
- [x] #4430 — GitHub App identity with per-installation rate buckets (isolates fleet from operator quota; fixes per-repo PAT scoping; operator-only prerequisites inside). **Closed COMPLETED 2026-07-29 21:21Z via PR #4454** (merge commit 4cd7f9b0 — installation-token minting for fleet hosts; App creation/provisioning remains an explicit operator step per the issue's scope).
- [x] #4431 — safehouse-primary claim coordination (architectural: coordination load stops scaling with fleet size). **Closed COMPLETED 2026-07-29 21:22Z via PR #4453** (merge commit a4322f87 — peer-claims promoted to primary in-flight signal on safehouse hosts).

## Non-goals / rejected

- Multiple free machine accounts: violates GitHub ToS; the App gives the same isolation legitimately.
- GitHub Enterprise Cloud (15,000/hr): 3× headroom at ~$21/user/mo without fixing the shared-identity design — reconsider only if the above land and we still saturate.

## Residual fast-follows (curator note, 2026-07-29)

All four children are closed COMPLETED, but two hardening follow-ups from #4430's Judge review are still in flight and belong to this epic's problem space:

- #4456 (`loom:building`) — harden the runtime `GH_TOKEN` env mutation shipped by PR #4454 (thread-safety of `set_var`, redundant refresh-tick writes)
[...]

PR #4453, opened 2026-07-29
feat(daemon): promote peer-claims to primary in-flight signal on safehouse hosts

### LINKED ISSUE #4431: fleet: move claim/peer coordination from GitHub label polling to safehouse
## Problem

A large share of steady-state GitHub API load is not issue discovery but **fleet coordination**: `claim_reconciliation` polls `loom:building` and `quarantine_reconciliation` polls `loom:blocked` per workspace per tick on every host, and peer-claim checks ride the same polling. This scales O(hosts × workspaces × ticks) against a fixed shared quota — it's the pattern that amplified the 2026-07-29 exhaustion.

## Proposal

Make safehouse (Matrix) the primary carrier for high-frequency claim state; GitHub labels become an eventually-consistent audit trail written once per transition rather than a polled coordination bus:

- Daemon already publishes peer-claims to the fleet room (ttl=120s) — promote this to the authoritative in-flight claim signal for dispatch decisions.
- Reconciliation against GitHub labels drops to a slow cadence (e.g. every 15–30 min, or on-demand at dispatch) purely to heal divergence (crashed host, offline safehoused).
- Hosts without safehouse configured (currently robb-pro) keep today's polling behavior — safehouse presence is the opt-in switch.

## Acceptance

- Steady-state claim-coordination API calls reduced by >90% on safehouse-enabled hosts.
- A claim placed on host A prevents dispatch of the same issue on host B within one tick, with GitHub unreachable.
- Divergence (stranded `loom:building` after host crash) healed by the slow reconciliation path — ties into the #4343 drain checklist.

Related: #3999 (loom:building persona semantics), #4085 (cross-host collision detection).

---

### Correction to the problem statement

`quarantine_reconciliation` does **not** poll `loom:blocked` per-tick — verified against `origin/main`: it is a **startup-only** pass (module doc, `loom-daemon/src/quarantine_reconciliation.rs:1-20`), invoked once from `main.rs:1236-1268` with no periodic interval task (unlike `claim_reconciliation`, which has one — see below). Only `claim_reconciliation` runs on a recurring cadence today. This changes where the bulk of savings actually comes from: the periodic `claim_reconciliation` task, not `quarantine_reconciliation`. The AC ">90% reduction" target should be measured against `claim_reconciliation`'s call volume, not both modules equally.

### What already exists (don't rebuild)

- **`claim_reconciliation`** (`loom-daemon/src/claim_reconciliation.rs`): polls open `loom:building` issues plus `loom:reviewing`/`loom:treating` PRs per registered workspace, on an independent periodic task (`spawn_periodic_reconciliation_task`, wired in `main.rs:1222-1224`). Interval resolution (`resolve_reconcile_interval()`, lines 176-184), in precedence order: env `LOOM_CLAIM_RECONCILE_INTERVAL_SECS`, else default `DEFAULT_RECONCILE_INTERVAL_SECS = 600s` (line 164), floored at `MIN_RECONCILE_INTERVAL_SECS = 60s` (line 169). Kill switch: `LOOM_STALE_CLAIM_RECONCILE` (default on, line 133).
- **Peer-claim publish/read-back already exists and is already consulted at dispatch** — this is not a net-new mechanism, it's a promotion of an existing soft signal:
  - `loom-daemon/src/peer_claims.rs`: `DEFAULT_PEER_CLAIM_TTL = 120s` (line 72, confirmed). TTL resolution (`resolve_peer_claim_ttl`, lines 261-275), in precedence order: env `LOOM_PEER_CLAIM_TTL_SECS`, else config `safehouse.peerClaimTtlSecs`, else the 120s default. `PeerClaimView::is_claimed_at`/`claimed_issues_at` (lines 367-384) is the read-back API, using **local** `Instant`-based expiry (not the advertiser's wall clock) to sidestep clock skew.
  - `loom-daemon/src/sweep_registry.rs`: `publish_peer_claim` (line 1643) fires on `dispatch()` (call site line 2144, *before* the non-atomic label flip); `peer_claimed_issues()` (line 1623) is the read side, fail-open (empty set) if no view or a poisoned mutex.
  - `loom-daemon/src/work_finder.rs`: `WorkSource::peer_claimed()` (line 391) is already consulted in `tick`/`tick_multi` (lines 568, 600-601, 819-820, 886-887) and counted separately in `TickReport::skipped_peer_claim` (lines 463-470) — **explicitly not folded into `collisions`**, so soft-claim prevention and post-hoc collision measurement stay distinguishable. Peer claims today are documented as "a soft claim, not a mutex" (`peer_claims.rs:17-23`) — they shrink the collision window, they don't replace the label as the authoritative source.
  - The opt-in switch this issue references ("hosts without safehouse configured... keep today's polling behavior") already exists: `SafehouseConfig::enabled` (`safehouse.rs:119`, default `false` at line 132), resolved via `resolve_config()` (lines 144-148). **No `robb-pro` hostname or per-host code path exists** — that host simply doesn't set `safehouse.enabled: true` in its own `.loom/config.json`. When disabled, `workspace_pool.rs:172` short-circuits peer-claim construction entirely (`PeerClaimView`/outbound channel never built), so `peer_claimed_issues()` returns empty and dispatch is byte-for-byte unchanged — the opt-in gate this issue needs already fully exists, nothing new to build there.
- **Detection-only cross-host collision counting** (#4085, closed) already runs immediately before every label flip: `classify_preflip_labels`/`detect_and_record_collision` (`sweep_registry.rs:2421-2465`, `2473-2501`, call site line 2157), gated by `LOOM_DETECT_COLLISIONS`/`autonomous.collisionDetection.enabled`. This measures how often the label race actually collides — useful as a regression baseline for this issue's AC 2 ("with GitHub unreachable").

### The actual gap this issue closes

1. **`claim_reconciliation`'s cadence is unconditional today** — its interval task runs at the same fixed cadence (default 600s) regardless of `safehouse.enabled`. There is no existing config path that slows it to 15-30 min specifically *when* safehouse peer-claims are live and healthy. This is the concrete lever: make `resolve_reconcile_interval()` (or its caller) read a slower interval when `safehouse.enabled` is true, keeping the existing default for hosts without safehouse (e.g. robb-pro).
2. **Peer-claim TTL was tuned for soft backoff (2× a 60s tick), not authoritative dispatch prevention.** If `PeerClaimView` becomes the *primary* signal and label reconciliation drops to a 15-30 min healing cadence, a live sweep whose claim ad is a one-shot publish at `dispatch()` (not periodically re-advertised — re-advertisement was not found in this pass) will silently fall out of `PeerClaimView` after 120s while the sweep keeps running, opening exactly the double-dispatch window this issue exists to close. **Open design question the implementer must resolve**: either add periodic re-advertisement of live claims (e.g. on the same cadence as the work-finder tick, or a heartbeat tied to sweep liveness), or extend the TTL substantially when it's promoted to primary — do not simply drop the reconciliation cadence without addressing this.
3. **No existing test exercises "GitHub unreachable, peer-claim is the only signal, does dispatch still avoid double-claim."** The closest existing coverage is `peer_claims.rs`'s `two_hosts_one_issue_second_host_backs_off_deterministically` (lines 587-607) — it establishes the claim/backoff shape but does not simulate a `gh`-unreachable dispatch decision end-to-end. New coverage is needed for the "primary signal" behavior this issue introduces (see Test Plan).

### Implementation guidance

- Make `claim_reconciliation`'s effective interval conditional: resolve `safehouse.enabled` (or a dedicated `safehouse.claimReconcileIntervalSecs` override) alongside `LOOM_CLAIM_RECONCILE_INTERVAL_SECS` in `resolve_reconcile_interval()`, so a safehouse-enabled host defaults to the slower healing cadence (e.g. 15-30 min) while a non-safehouse host is unaffected. Keep `MIN_RECONCILE_INTERVAL_SECS` as the floor either way.
- Resolve the re-advertisement/TTL question above before relying on `PeerClaimView` as primary — this is the crux of the correctness risk, not a footnote.
- `quarantine_reconciliation` needs no cadence change (it's already startup-only) — exclude it from this issue's cadence work; note the correction in the PR description so a reviewer doesn't go looking for a periodic task that doesn't exist.
- Reuse `peer_claimed_issues()` / `WorkSource::peer_claimed()` as the "authoritative for dispatch" hook rather than inventing a parallel structure — the read-back path already exists end-to-end; this issue is about *trusting it more* and *polling GitHub less*, not building new plumbing.
- No blocking dependency on sibling epic children #4429 (rate-limit backoff/telemetry), #4428 (ETag caching), or #4430 (GitHub App identity) — all open, unlabeled, and addressing different consumption sources (general work-finder/role polling, not claim coordination specifically). They're parallel phases of parent epic #4432, not prerequisites for this issue.

## Affected Files

- `loom-daemon/src/claim_reconciliation.rs` - make the periodic interval (`resolve_reconcile_interval`, lines 176-184; `spawn_periodic_reconciliation_task`) conditional on `safehouse.enabled`
- `loom-daemon/src/peer_claims.rs` - TTL/re-advertisement semantics if `PeerClaimView` becomes the primary dispatch signal (currently soft-claim tuned, `DEFAULT_PEER_CLAIM_TTL = 120s`, line 72)
- `loom-daemon/src/sweep_registry.rs` - `publish_peer_claim` (line 1643, call site line 2144) and `peer_claimed_issues()` (line 1623) — likely re-advertisement/heartbeat addition
- `loom-daemon/src/work_finder.rs` - `WorkSource::peer_claimed()` (line 391), `TickReport::skipped_peer_claim` (lines 463-470) — dispatch-skip semantics if peer-claim becomes authoritative rather than soft
- `loom-daemon/src/safehouse.rs` - `SafehouseConfig` (line 118), `resolve_config()` (lines 144-148) — reference; add a config field here if a dedicated interval override is chosen over reusing `safehouse.enabled`
- `loom-daemon/src/main.rs` - wiring at lines 1222-1224 (claim) and 1236-1268 (quarantine, startup-only, no change expected)
- `defaults/docs/daemon-reference.md` - document the safehouse-conditional reconciliation cadence and the promoted peer-claim role

## Test Plan

- [ ] Automated tests (`cargo test -p loom-daemon`): `resolve_reconcile_interval()` (or its replacement) returns the slow cadence when `safehouse.enabled` is true and the existing default otherwise; extend `peer_claims.rs`'s two-host test to cover a claim surviving longer than 120s if re-advertisement/TTL-extension is added; a new end-to-end-shaped test simulating a `gh`-unreachable dispatch decision where the peer claim is the only signal preventing a second host from dispatching the same issue (the epic's own AC 2 restated as a test).
- [ ] Manual verification: two hosts with `safehouse.enabled: true`, one claims an issue, simulate GitHub unreachable (or just block the `gh` binary) on the second host, confirm it does not dispatch the same issue within one tick; separately confirm a host with `safehouse.enabled: false` (e.g. robb-pro-shaped config) sees zero behavior change.
- [ ] Edge cases: safehouse enabled but the peer-claim channel/view is still cold (daemon just started, no claims observed yet) — must not silently allow duplicate dispatch before the view warms up; a claim's advertising host crashes without retracting — confirm the slow (15-30 min) label-based reconciliation still heals the stranded `loom:building` claim per AC 3, tying into #4343's drain checklist as noted in the original body.

### Duplicate / related-work check

`check-duplicate.sh --include-merged-prs --issue 4431` returned a self-match (100% similarity against itself — known checker noise, see #4388's curator notes) and one `RELATED_OPEN_WORK` hit: **#4432** (the parent epic, "survive GitHub API rate limits at fleet scale"). **Absorbed** — #4432 lists this issue as one of four children in rough sequencing order; verified its other three children (#4429, #4428, #4430) are open, unlabeled, and address disjoint consumption sources (general polling backoff, ETag caching, GitHub App identity) rather than claim coordination, so none block this issue. No spec conflict to reconcile.

rjwalters:
Closes #4431. Fourth child of epic #4432 (after #4440/#4429 merged and #4443/#4428 in review).

## The curator's crux, resolved

The curator flagged the correctness risk precisely: peer claims were a **one-shot publish at dispatch** with a 120s TTL tuned for soft backoff — promoting them to the primary signal while slowing label reconciliation would let a live sweep's claim silently expire from peers' views mid-run, reopening the double-dispatch window. Resolution: **re-advertisement heartbeat**, not TTL extension:

- `SweepRegistry::readvertise_peer_claims()` re-publishes every live (`Running`/`Pending`) Issue sweep's claim; the reaper calls it each tick (default 30s ≈ 4 refreshes per TTL window), after `reap_once` so a just-reaped sweep is never re-advertised.
- `PeerClaimView::observe_at` already restarts the local expiry clock on a repeat `Advertise` (no receiver change needed — covered by a new test).
- A crashed host's claims still free within **one TTL of its last heartbeat** — the crash-release property the short TTL exists for is preserved exactly, which is why re-advertisement beats a long TTL (stale claims from dead hosts would linger and starve issues fleet-wide).

## Safehouse-conditional reconciliation cadence

`resolve_reconcile_interval_for(root)`: env `LOOM_CLAIM_RECONCILE_INTERVAL_SECS` > config `safehouse.claimReconcileIntervalSecs` > **1800s** when `safehouse.enabled` > 600s otherwise, floored at 60s. A safehouse host's `loom:building` pass demotes to a healing sweep (stranded claims from crashed hosts, pre-safehouse dispatches); a host without safehouse (e.g. robb-pro) is byte-for-byte unchanged — the existing `workspace_pool` short-circuit already guarantees the no-safehouse path never consults peer claims.

## Curator corrections honored

- `quarantine_reconciliation` is startup-only (no periodic task) — no cadence change made there, per the curator's verified correction.
- Reused `peer_claimed_issues()` / `WorkSource::peer_claimed()` as-is — the dispatch-skip path is unchanged; this PR changes how long a claim stays visible and how often labels are polled, not the skip semantics.

## Message-volume note

Re-advertisement adds ~2 room events/min per live sweep (30s tick). At realistic fleet load (≤10 concurrent sweeps) that's ≤20 msg/min of tiny `task` envelopes — bounded `try_send`, fail-open drop under backlog, same as the dispatch-time ad.

## Testing

- `peer_claims`: new test proving a re-advertised claim survives past the original TTL and still expires one TTL after the LAST heartbeat.
- `sweep_registry`: new test proving re-advertisement publishes for live Issue sweeps only (terminal-state entries excluded; publisher-less = silent no-op).
- `claim_reconciliation`: new test covering the full interval precedence chain (env > config override > safehouse default > legacy default; zero/invalid override ignored; 60s floor).
- 263 tests across the three touched modules pass; clippy `--all-targets` + `cargo fmt --check` clean.
- Docs: daemon-reference gains the `safehouse.claimReconcileIntervalSecs` row, the re-advertisement heartbeat bullet, and the TTL row now notes it only bounds a crashed host's lingering claim.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2057,7 +2057,8 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.watchdog.reviewStallTimeoutSecs` | `LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS` | `2700` | Log-silence window before a hung Judge/Doctor sweep is re-dispatched |
 | `autonomous.collisionDetection.enabled` | `LOOM_DETECT_COLLISIONS` | `false` | Cross-host dispatch-collision baseline (#4085). Off by default — adds one extra `gh issue view --json labels` round-trip per dispatch. Detection only: a collision is logged/counted, never acted on |
 | `safehouse.enabled` | `LOOM_SAFEHOUSE_ENABLED` | `false` | Enables safehouse fleet-comms (#3997) **and** cross-host soft-claim coordination (#4028). Off by default — a byte-for-byte no-op (no socket, no coordination task) when unset |
-| `safehouse.peerClaimTtlSecs` | `LOOM_PEER_CLAIM_TTL_SECS` | `120` | Peer-claim TTL, in seconds (#4028) — how long a peer's soft claim suppresses local dispatch (measured against local receipt, not the advertiser's clock). Default = 2× the 60s work-finder tick |
+| `safehouse.peerClaimTtlSecs` | `LOOM_PEER_CLAIM_TTL_SECS` | `120` | Peer-claim TTL, in seconds (#4028) — how long a peer's soft claim suppresses local dispatch (measured against local receipt, not the advertiser's clock). Default = 2× the 60s work-finder tick. Since #4431 live claims are re-advertised every reaper tick, so the TTL only bounds how long a **crashed** host's claim lingers |
+| `safehouse.claimReconcileIntervalSecs` | `LOOM_CLAIM_RECONCILE_INTERVAL_SECS` | `1800` when `safehouse.enabled`, else `600` | Periodic `loom:building`/PR-claim reconciliation cadence (#4431). With safehouse peer-claims carrying the fast in-flight signal (re-advertised each reaper tick), label reconciliation demotes to a slow healing sweep. Env wins on any host; floored at 60s |
 | *(host identity)* | `LOOM_HOST_ID` | `$HOSTNAME` → `hostname` → `unknown-host` | This host's identity string, used in collision log records (#4085) **and** peer-claim self-recognition (#4028); set it where the daemon runs without `$HOSTNAME` exported |
 | `autonomous.autoUpdate.enabled` | `LOOM_AUTO_UPDATE` | `false` | Autonomous self-update loop on/off (#4055). **Opt-in** (it rebuilds + restarts the daemon process). Exactly one loop per daemon, not a per-workspace fan-out. See [Autonomous self-update loop](#autonomous-self-update-loop-4055) below |
 | `autonomous.autoUpdate.intervalSecs` | `LOOM_AUTO_UPDATE_INTERVAL_SECS` | `900` | Cadence between staleness checks. Zero/invalid → default |
@@ -2457,6 +2458,22 @@ for the full design.
   clock skew is not comparable across hosts), so a crashed peer cannot
   permanently starve an issue. A peer also emits a `retract` ad from its reaper on
   a terminal sweep outcome, freeing the issue before the TTL.
+- **Re-advertisement heartbeat (#4431).** The dispatch-time ad is no longer
+  one-shot: the reaper re-advertises every live (`Running`/`Pending`) Issue
+  sweep's claim each tick (default 30s — ~4 refreshes per TTL window), and a
+  repeat ad **restarts** the receiver's TTL clock. A live sweep's claim
+  therefore never expires from peers' views mid-run, while a crashed host's
+  claims still free within one TTL of its last heartbeat. This is what lets
+  the peer claim serve as the **primary fast in-flight signal** on
+  safehouse-enabled hosts, with the `loom:building` label demoted to a slow
+  healing audit trail:
+- **Safehouse-conditional reconciliation cadence (#4431).** On a host with
+  `safehouse.enabled`, `claim_reconciliation`'s periodic pass defaults to
+  **1800s** (30 min, healing-only) instead of 600s. Precedence:
+  `LOOM_CLAIM_RECONCILE_INTERVAL_SECS` env >
+  `safehouse.claimReconcileIntervalSecs` config > 1800s (safehouse) / 600s
+  (no safehouse), floored at 60s. Hosts without safehouse are byte-for-byte
+  unchanged.
 - **Self-claim recognition.** A daemon never backs off on its own advertisement:
   the claim body carries the host identity (`host_identity()`:
   `LOOM_HOST_ID` > `$HOSTNAME` > `hostname` > `unknown-host` — loom's single,
--- a/loom-daemon/src/claim_reconciliation.rs
+++ b/loom-daemon/src/claim_reconciliation.rs
@@ -168,11 +168,25 @@ pub const DEFAULT_RECONCILE_INTERVAL_SECS: u64 = 600;
 /// (e.g. `5` meaning "5 minutes") cannot turn this into a `gh`-API busy loop.
 pub const MIN_RECONCILE_INTERVAL_SECS: u64 = 60;
 
+/// Default periodic reconciliation interval when safehouse peer-claims are
+/// live (Issue #4431): 30 minutes. With claims advertised at dispatch and
+/// re-advertised every reaper tick (`sweep_registry::readvertise_peer_claims`),
+/// the fleet room — not the `loom:building` label — is the fast in-flight
+/// claim signal, and this pass demotes to a slow *healing* sweep for the
+/// divergence cases safehouse cannot see (a host that crashed without
+/// retracting, a stranded label from a pre-safehouse dispatch). Hosts without
+/// `safehouse.enabled` keep [`DEFAULT_RECONCILE_INTERVAL_SECS`] unchanged.
+pub const DEFAULT_SAFEHOUSE_RECONCILE_INTERVAL_SECS: u64 = 1800;
+
 /// Resolve the periodic reconciliation interval from
 /// [`RECONCILE_INTERVAL_ENV`], falling back to
 /// [`DEFAULT_RECONCILE_INTERVAL_SECS`] for an absent, unparseable, or
 /// non-positive value, and always clamped up to
 /// [`MIN_RECONCILE_INTERVAL_SECS`] regardless of source.
+///
+/// Env-only resolution — the safehouse-aware default lives in
+/// [`resolve_reconcile_interval_for`], which callers with a workspace root
+/// should prefer.
 #[must_use]
 pub fn resolve_reconcile_interval() -> Duration {
     let secs = std::env::var(RECONCILE_INTERVAL_ENV)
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed the full diff (4 files, +310/−4) plus integration points on the branch.

**Design verification:**
- The re-advertisement heartbeat correctly resolves the curator's crux: `readvertise_peer_claims()` runs each reaper tick (verified `DEFAULT_REAPER_INTERVAL_SECS = 30`, ~4 refreshes per 120s TTL window), and it runs **after** `reap_once` so a just-reaped sweep is never re-advertised. The crash-release property (claims free within one TTL of the last heartbeat) is preserved exactly, and the new `peer_claims` test proves both halves.
- Re-advertisement reuses `publish_peer_claim`, which derives the repo slug from the registry's `workspace_root` — identical to the dispatch-time ad, so the promoted signal is consistent with the existing publish path. Fail-open contract (`try_send`, publisher-less no-op) holds, so the reaper never blocks under the registry mutex.
- `resolve_reconcile_interval_for()` implements the full precedence chain (env > config > safehouse 1800s > legacy 600s, 60s floor) and the production caller (`main.rs:1324` → `spawn_periodic_reconciliation_task`) was migrated; the old env-only `resolve_reconcile_interval()` survives only for its own tests, with a doc pointer to the new resolver. No-safehouse hosts hit the `unwrap_or` legacy branch — byte-for-byte unchanged as claimed.
- Docs: daemon-reference gains the config row + heartbeat bullet; TTL row correctly re-scoped to crashed-host lingering.

## Test Execution

1. `peer_claims` re-advertisement TTL test — ✅ Executed: passes (in scoped run below)
2. `sweep_registry` live-Issue-only re-advertisement test — ✅ Executed: passes (terminal-state excluded, publisher-less no-op both asserted)
3. `claim_reconciliation` precedence-chain test — ✅ Executed: passes (env > config > safehouse default > legacy; zero/invalid ignored; 60s floor)
4. 263 tests across the three touched modules — ✅ Executed: `cargo test -- peer_claims:: claim_reconciliation:: sweep_registry::` → **263 passed, 0 failed** (122s)
5. clippy + fmt — ✅ Verified via CI ("Rust Linting & Build Check" green) rather than re-run locally

All CI checks pass, merge state CLEAN. Ready for Champion auto-merge.

Issue #4469, opened 2026-07-29
[Epic #4167 Phase 2] Provision ChatGPT-subscription OAuth for Codex via CODEX_HOME profile (operator-only) + document the layout


rjwalters:
Part of #4167 (Phase 2 — Codex adapter). Companion to the spawn-codex.sh port issue (#4468).

## Goal
Provision durable ChatGPT-subscription auth for the Codex canary, and document the profile layout the future provider-aware pool (Phase 4) will generalize.

**Operator-only**: `codex login` is an interactive browser OAuth flow; no agent can complete it.

## Status (Curator update, 2026-07-29)

**The interactive provisioning is complete** — see the operator's [completion comment](https://github.com/rjwalters/loom/issues/4469#issuecomment-5123807494) dated 2026-07-29. Verified done:

- [x] Operator ran `codex login` (ChatGPT subscription OAuth) under a deliberate profile path: `~/.loom/codex-profiles/<account>/` (dir `0700`, `auth.json` `0600`, copied from `~/.codex/`) — per-account dirs, ready for Phase 4 rotation
- [x] `CODEX_HOME=<profile> codex login status` → "Logged in using ChatGPT" from a fresh shell
- [x] Headless smoke: `CODEX_HOME=<profile> codex exec --sandbox read-only '<prompt>' </dev/null` → exit 0, correct output, token usage reported

**Canonical-copy caveat (operator-recorded, load-bearing for Phase 4):** `~/.codex/auth.json` still exists as the default-profile copy. The CLI refreshes whichever profile `CODEX_HOME` selects, so the two copies' refresh tokens may diverge over time — treat the `~/.loom/codex-profiles/` copy as **canonical for Loom dispatch**. Any doc or Phase-4 pool code must state this.

## Remaining scope (doc section only)

- [ ] A short doc section recording: the `~/.loom/codex-profiles/<account>/` layout, `CODEX_HOME` selection, auth.json refresh behavior, the canonical-copy caveat above, and the security posture (`0700` dir / `0600` file, never committed, machine-local)

**Delivery route** — the operator suggests folding this into #4468's doc deliverables rather than a separate PR:
- **Option A (recommended, per operator):** #4468's PR adds the section to `defaults/docs/runtime-adapters.md` (auth/contract-point 2 area) alongside its `guardrail-parity-codex.md` deliverable. Note #4468's body currently points at "the companion auth issue" (this one) for auth detail — whichever PR delivers it, the two issues must not both defer to each other.
[...]

Issue #4167, opened 2026-07-28
Proposal: first-class multi-runtime worker support (Claude Code, Codex, Amp, oh-my-pi) via a runtime adapter contract


rjwalters:
## Problem

Loom is hardwired to Claude Code at every layer: the spawn path (`spawn-claude.sh`), the token pool, instruction packaging (`CLAUDE.md`, `.claude/` skills), guard hooks, error classification, model aliasing, and the daemon's terminal management. Graham Peyton's fork (gpeyton/loom, see #4165) proved both the demand and the cost: Codex support works there, but as a parallel special case (`spawn-codex.sh`, `defaults/.codex/`, Codex-specific CI) — adding a third runtime that way would triple the special-casing.

The operator direction (2026-07-27) is to support **multiple CLI agent tools as first-class worker runtimes** — Claude Code, OpenAI Codex CLI, Amp, oh-my-pi (omp), and future tools — and to do it **in collaboration with Graham** (upstream PRs from his fork, not one-way cherry-picks).

## Goal

A worker-runtime abstraction in which **Claude Code is adapter #1, not a privileged default**. Codex lands as adapter #2 by upstreaming the fork's work through the new interface; a third runtime (Amp or oh-my-pi) validates that the abstraction generalizes (rule of three).

## Design pillars

1. **Runtime adapter contract.** One contract every runtime implements:
   - *spawn*: headless prompt execution → exit code + transcript path (seeded by the fork's `spawn-worker.sh` dispatcher, fork PR #9)
   - *model mapping*: Loom logical tiers → runtime-specific model IDs (must reconcile with `sweep.modelAliases` and the fork's cost-of-being-wrong tiering; note the existing Rust/Python modelAliases divergence)
   - *error classification*: per-provider pattern tables (fork PR #6 already restructured `classify-error.sh` this way)
   - *usage accounting*: session cost/limit signals feeding the account pool
   - *instruction format*: which files the runtime reads and how they're generated
   - *permission/sandbox mapping*: how Loom's guard intent translates to the runtime's own sandbox
   - *capability declaration*: MCP, subagents, hooks, skills — yes/no/partial

[...]

Issue #20, opened 2025-10-12
Add tooltip system and UI hints for improved discoverability


rjwalters:
# Add Tooltip System and UI Hints for Improved Discoverability

## Problem

The Loom UI currently lacks visual hints and tooltips to guide users on available interactions. Users must discover features through trial and error, which leads to poor discoverability of functionality like:

- Double-click to rename terminals
- Clear terminal history button
- Settings button functionality
- Drag-and-drop terminal reordering
- Keyboard shortcuts

## Goals

1. **Improve Discoverability**: Make hidden features obvious through contextual hints
2. **Professional UX**: Match modern desktop application standards for tooltips
3. **Non-Intrusive**: Hints should enhance, not clutter the interface
4. **Modular System**: Reusable tooltip component for future UI additions
5. **Accessibility**: Support keyboard navigation and screen readers

### 1. Tooltip System
- **Hover tooltips**: Small popup boxes that appear after ~500ms hover
- **Position-aware**: Automatically adjust position to stay on screen
- **Dismissible**: Hide on mouse leave or user action
- **Styling**: Dark/light theme support with smooth fade animations

#### Terminal Header Bar
- **Clear button (trash icon)**: "Clear terminal history"
- **Settings button (gear icon)**: "Configure terminal role and settings"
- **Terminal name**: "Double-click to rename"
- **Status dot**: "Terminal status: idle/busy/error/stopped"

#### Mini Terminal Cards
- **Card itself**: "Click to switch, drag to reorder"
- **Terminal name**: "Double-click to rename"
- **Close button (×)**: "Close terminal"

#### Add Terminal Button
- **Plus button**: "Add new terminal (Cmd+T)" (when workspace open)
- **Plus button (disabled)**: "Select a workspace first" (when no workspace)

#### Workspace Selector
- **Path input**: "Enter path to git repository"
- **Browse button**: "Browse for folder"

### 3. Context Hints (Optional Enhancement)
Consider adding a status bar hint system that shows contextual help:
- Shows keyboard shortcuts when hovering over buttons
- Displays action hints (e.g., "Double-click name to rename" when hovering over terminal title)
- Could reuse the terminal header bar space to show hints

### Phase 1: Core Tooltip Component

**Create `src/lib/tooltip.ts`**:
```typescript
interface TooltipOptions {
  text: string;
  position?: 'top' | 'bottom' | 'left' | 'right' | 'auto';
  delay?: number; // milliseconds before showing (default 500)
[...]

PR #4417, opened 2026-07-29
fix(daemon): add missing main_health_gate_* fields to serve.rs test helper

rjwalters:
## Summary

Fixes the semantic merge conflict between #4404 (serve skeleton) and #4297 (tiered build gate): `serve.rs`'s `empty_report()` test helper predates #4297's three new `DaemonStatusReport` fields, breaking `cargo check --workspace --all-targets` on main.

## Fix

Added the three missing fields to the `empty_report()` initializer with the neutral defaults the issue specifies:
- `main_health_gate_deferred: false`
- `main_health_gate_deferred_reason: None`
- `main_health_gate_verdict_tier: None`

No behavior change outside `#[cfg(test)]` code.

## Test plan

- [x] `cargo check --workspace --all-targets` passes (was failing with E0063 on bare main)
- [x] `cargo test -p loom-daemon --lib serve` — all 6 `serve::tests::*` pass, including `build_snapshot_round_trips_zero_workspaces_edge_case`

Closes #4412

PR #4476, opened 2026-07-29
feat(runtime): port the fork's Codex adapter onto the runtime seam (attribution: gpeyton)

### LINKED ISSUE #4468: [Epic #4167 Phase 2] Port the fork's Codex adapter onto the runtime seam: spawn-codex.sh, error patterns, guardrail-parity doc, CI leg (attribution: gpeyton)
Part of #4167 (Phase 2 — Codex adapter). Operator decisions recorded at https://github.com/rjwalters/loom/issues/4167#issuecomment-5123769971.

## Goal
Port the gpeyton/loom fork's Codex worker support onto the Phase-1 runtime seam, **with attribution to Graham Peyton** in commit messages and the PR body. This is a port-and-restructure, not a cherry-pick: the fork's implementation predates the contract doc, so each piece lands behind the interface `defaults/docs/runtime-adapters.md` defines.

## Scope (the seven contract points, Codex column)
- [ ] `defaults/scripts/spawn-codex.sh` — headless spawn via `codex exec <prompt>` (Codex CLI ≥0.146.0), args passthrough, prompt delivery, exit-code passthrough, transcript path reporting. Slots behind `spawn-worker.sh` (dispatch already resolves `spawn-<runtime>.sh`; unknown-runtime exit-78 path already tested).
- [ ] Auth handling: honor `CODEX_HOME` for per-account profile selection (`auth.json` refresh-token bundle); NO token-pool integration in this phase — a single operator-provisioned profile is sufficient (see the companion auth issue).
- [ ] Error classification: extend `.loom/scripts/lib/classify-error.sh`'s per-provider pattern tables (the fork's PR #6 restructure is the seed) with Codex CLI failure vocabulary (auth expiry, rate/usage limits, sandbox denials).
- [ ] `defaults/docs/guardrail-parity-codex.md` — map Loom guard intent → Codex sandbox modes (`read-only`, `workspace-write`, sandbox_permissions config), with an explicit residual-gap section. No runtime is admitted without one (contract point 5).
- [ ] CI smoke leg: a workflow job that runs `spawn-codex.sh --help`-level checks plus a mocked-CLI spawn test (no live OpenAI calls in CI). Tier-2 admission = this leg green.
- [ ] Model mapping: minimal — a static default model in the adapter; full tier→model resolution through `sweep.modelAliases`-style indirection is Phase 4, NOT here.

## Reference material
- Fork: https://github.com/gpeyton/loom (fetch patches via `gh api repos/gpeyton/loom/commits/<sha>`; check for superseding commits — fork commits are design ports, not cherry-picks)
- Contract: `defaults/docs/runtime-adapters.md` (all seven points, tier policy)
- Fork findings to honor: native Codex agents prohibited for Loom lifecycles (fork PR #59); `CODEX_HOME` profile rotation.

## Non-goals (this issue)
Daemon spawn-path migration to spawn-worker.sh (separate issue), provider-aware token pool (Phase 4), AGENTS.md codegen (separate issue), any change to role prompts or the label state machine.

---

## Curator Enhancement (2026-07-29, consolidated)

> Two curator passes ran concurrently on this issue; this section is the merged, current spec and **supersedes the individual curator comments below** where they differ. Complexity marker: `complex` (tagged once, near Goal above) — the guardrail-parity mapping is a security trust boundary where a wrong sandbox-mode mapping could pass tests and review unnoticed.

### Phase-1 seam verification (against `origin/main`)

Everything this issue assumes from Phase 1 is confirmed present on `origin/main` as of 2026-07-29:

- `defaults/scripts/spawn-worker.sh` — the dispatcher already names `spawn-codex.sh` as the intended slot; unknown-runtime exit-78 path is covered by `defaults/scripts/tests/test-spawn-worker.sh`.
- `defaults/scripts/lib/classify-error.sh` — the engine vs. provider-pattern-table split (#4190) is live, and a `_classify_error_codex()` stub already exists (with `codex` wired into `_classify_error_provider()`'s case dispatch). That scope item is "fill the stub", not "restructure".
- `defaults/docs/runtime-adapters.md` — all seven contract points present, including §6 Permission/sandbox mapping and "Adding a runtime adapter"; ADR at `docs/adr/0012-runtime-adapter-contract.md`.

### Implementation notes

- **CI leg**: `test-spawn-worker.sh` / `test-spawn-claude.sh` exist but are **not run by any workflow today** (no `.github/workflows/*` references them). The smoke leg is therefore a new job in `.github/workflows/ci.yml` (mirror the `installer-tests` job's plain-bash style) running the new `test-spawn-codex.sh`. Scope the leg strictly to the new test — wiring the pre-existing spawn tests into CI is a pre-existing gap, out of scope here.
- **Mocked-CLI spawn test**: place a fake `codex` executable in a temp dir prepended to PATH; assert prompt delivery, exit-code passthrough, transcript-path reporting, and `LOOM_RUNTIME=codex` resolution through `spawn-worker.sh`. No live OpenAI calls anywhere in CI.
- **Live-CLI behaviors the adapter MUST handle** (observed against real Codex CLI 0.146.0 during the #4469 smoke test — full detail in the 22:00Z comment below):
  - `codex exec` refuses non-git cwds without `--skip-git-repo-check`; worktree dispatch is fine, scratch-dir dispatch needs the flag.
  - With piped/non-tty stdin the CLI appends stdin as a `<stdin>` block and can hang — the wrapper must redirect `</dev/null` (or deliberately deliver the prompt via stdin).
  - Plain-mode stdout carries a `session id: <uuid>` line — the transcript join key (contract point 2) — and a `tokens used` line (cost signal, contract point 4); prefer the durable session JSONL under `$CODEX_HOME/sessions/` once mapped.
  - `--sandbox read-only` works headlessly (fits the Judge canary); Builder-equivalent work needs `workspace-write` — document both in the parity doc.
- **Attribution**: credit Graham Peyton in commit messages and the PR body. Fetch fork patches via `gh api repos/gpeyton/loom/commits/<sha>` (REST, no clone) and check for superseding commits — fork commits are design ports, not cherry-picks.
- **Guardrail parity doc**: follow contract point 6's structure and include the explicit residual-gap section (the contract point 5 admission requirement). Also record fork PR #59's finding — native Codex agents are prohibited for Loom lifecycles — in the doc.
- **Model mapping**: minimal static default via a new `defaults/runtimes/codex.json` mirroring the existing `defaults/runtimes/claude.json`. No `sweep.modelAliases`/tierModels indirection here (Phase 4).
- **Scheduling priority (#4233)**: applied inside `spawn-<runtime>.sh`, not the dispatcher — `spawn-codex.sh` should apply its own `nice`/`taskpolicy` policy the way `spawn-claude.sh` does.

### Affected Files

- `defaults/scripts/spawn-codex.sh` — NEW: the Codex adapter (headless `codex exec`, args/exit-code passthrough, `CODEX_HOME` honored, transcript path reporting)
- `defaults/scripts/lib/classify-error.sh` — fill the existing `_classify_error_codex()` stub with Codex failure vocabulary (auth expiry, rate/usage limits, sandbox denials)
- `defaults/docs/guardrail-parity-codex.md` — NEW: guard-intent → Codex sandbox-mode mapping + residual-gap section; use the fork's `GUARDRAIL-PARITY.md` (referenced from `runtime-adapters.md`) as the template; **also owns the CODEX_HOME profile doc section absorbed from #4469** (see Absorbed scope below)
- `defaults/runtimes/codex.json` — NEW: static default-model mapping, mirroring `defaults/runtimes/claude.json`
- `defaults/scripts/tests/test-spawn-codex.sh` — NEW: mocked-CLI spawn test + classifier vectors (style: `test-spawn-worker.sh`)
- `.github/workflows/ci.yml` — NEW job: Codex adapter smoke leg (mocked, no live calls; strictly the new test)
- `defaults/docs/runtime-adapters.md` — minor: update the fork-mapping/capability rows that currently mark Codex as future; keep the diff small

Edit `defaults/` sources only — the `.loom/scripts` dir and `.loom/docs/*.md` files are installed symlinks/copies (refreshed by `resync-installed.sh`).

### Test Plan

- [ ] Automated: `bash defaults/scripts/tests/test-spawn-codex.sh` green — mocked `codex` binary; covers prompt delivery, exit-code passthrough, `CODEX_HOME` profile selection, transcript-path reporting, missing-binary failure
- [ ] Automated: `LOOM_RUNTIME=codex` resolves through `spawn-worker.sh` to `spawn-codex.sh` (extend `test-spawn-worker.sh` or cover in the new test)
- [ ] Automated: classifier vectors for the new `codex` table (auth expiry / rate-or-usage limit / sandbox denial → expected categories); existing `claude` vectors unchanged
- [ ] CI: the new smoke-leg job green on the PR; `shell-lint` (shellcheck) green on the new scripts
- [ ] Edge cases: `codex` binary absent → clear adapter error (NOT exit 78, which is reserved for unknown-runtime dispatch failure); unset `CODEX_HOME` falls through to the Codex CLI default
- [ ] Manual/live smoke (now possible — #4469 provisioning is complete): `LOOM_RUNTIME=codex CODEX_HOME=~/.loom/codex-profiles/<account> defaults/scripts/spawn-worker.sh -p "echo hello"` runs headlessly, exits 0, reports a transcript path. Full canary runs remain #4470's scope.

### Absorbed scope (from #4469)

- [ ] Document the CODEX_HOME profile contract in this issue's doc deliverable (`guardrail-parity-codex.md`, auth section): layout `~/.loom/codex-profiles/<account>/`, permissions `0700` dir / `0600` `auth.json`, the canonical-copy caveat vs `~/.codex/auth.json`, and token-refresh posture. This was #4469's one remaining AC — the operator asked for it to be folded in here so #4469 can close when this lands.

### Dependencies

None blocking. #4469's provisioning is **complete** (operator confirmed 2026-07-29 — profile layout recorded in #4469's completion comment); its remaining doc AC is absorbed above. Do not mark this issue `loom:blocked`.

### Related open work

- #4469 — companion auth issue (operator-only). Provisioning done; its residual doc AC is **absorbed** into this issue (see Absorbed scope) so #4469 can close when this lands.
- #4470 — Codex canary runs; a downstream consumer of this port. No spec conflict.
- #4167 — parent epic; the operator-decisions comment linked above is authoritative for scope. This enhancement stays within those decisions.

rjwalters:
Closes #4468
Part of #4167 (Phase 2 — Codex adapter)

## Attribution

**Ports Codex support from the [gpeyton/loom](https://github.com/gpeyton/loom) fork by Graham Peyton** (fork PRs #15 runner, #16 config, #20/#40 guardrail parity, #6 classifier tables, #59 native-agent finding).

This is a **design port, not a cherry-pick**. The fork's implementation predates the contract doc, so each piece lands behind the interface `defaults/docs/runtime-adapters.md` defines — and every fork claim was re-verified against the live **codex-cli 0.146.0** before being carried across. Several no longer hold; those are corrected inline in the parity doc rather than silently propagated.

## What landed

| File | Change |
|---|---|
| `defaults/scripts/spawn-codex.sh` | **new** — adapter #2 (contract point 1) |
| `defaults/docs/guardrail-parity-codex.md` | **new** — contract point 6's required parity doc |
| `defaults/runtimes/codex.json` | **new** — capability manifest (contract point 7) |
| `defaults/scripts/tests/test-spawn-codex.sh` | **new** — 135 hermetic assertions |
| `.github/workflows/ci.yml` | `codex-adapter-smoke` leg (tier-2 admission gate) |
| `defaults/scripts/lib/classify-error.sh` | filled the `_classify_error_codex()` stub |
| `defaults/docs/runtime-adapters.md` | status tables + accurate §3 + "adding an adapter" guide |

No config change was needed to register the runtime — `spawn-worker.sh` discovers `spawn-<runtime>.sh` by filename, and the seam needed **zero** modification to admit Codex. That is the Phase-1 design paying off, and there is a regression test asserting `codex` moving from unknown → known did not weaken the exit-78 unknown-runtime guard for other names.

## Live-CLI intel from the issue: honored, with one correction

The issue's comment recorded five behaviors from the #4469 smoke test. All are handled; **item 3 was wrong about the stream**, which materially changed the implementation:

| # | Intel | Handling |
|---|---|---|
| 1 | Git-repo trust check | `--skip-git-repo-check` injected **only** when `git rev-parse --is-inside-work-tree` fails. Worktrees keep the check enabled; scratch dirs get the waiver plus a warning. The refusal classifies `FATAL`, so a mis-set cwd fails fast instead of retrying forever. |
| 2 | Stdin append hang | The child's stdin is **always** `</dev/null` on the headless path; the prompt is delivered positionally so stdin has no job. Two tests assert the child sees no stdin even when data is waiting on ours. |
| 3 | Output shape "(stdout, plain mode)" | **Corrected.** On 0.146.0 stdout carries *only* the agent's final message; the banner, `session id:`, and `tokens used` all go to **stderr**. Verified twice (success and 401 paths). The adapter therefore tees the child's stderr through a temp file while stdout passes through byte-for-byte — and recovers the exit code from `PIPESTATUS` so exit-code passthrough survives not using `exec`. |
| 4 | `--sandbox read-only` honored headlessly | It is the adapter's **default**. See the sandbox divergence below. |
| 5 | `CODEX_HOME` selection works | Implemented as a 4-tier passthrough chain (`LOOM_CODEX_HOME` > pre-set `CODEX_HOME` > `LOOM_CODEX_PROFILE` under `~/.loom/codex-profiles/` > ambient `~/.codex`). |

Additionally, `session id:` is resolved to the concrete transcript path (`$CODEX_HOME/sessions/<Y>/<M>/<D>/rollout-<ts>-<session-id>.jsonl`) and `tokens used` is parsed (comma-stripped) as the contract-point-4 cost signal.

### 1. Sandbox default is `read-only`; skip-permissions maps to `workspace-write`, not full access

The fork maps Loom's `--dangerously-skip-permissions` to `--dangerously-bypass-approvals-and-sandbox` (no sandbox at all), arguing *"parity, not a new exposure"* — Claude already runs unattended with full tool access.

**That premise does not hold.** Claude's unattended posture is backstopped by `PreToolUse` guards that fire on every Bash/Edit/Write call *even under* `--dangerously-skip-permissions`. Those guards cannot run under Codex. Handing Codex the same flag therefore produces a **strictly weaker** boundary than the Claude path it imitates — an agent with Claude's authority and none of Claude's backstops. `workspace-write` is the closest honest analogue of what the guards actually enforce.

Precedence: explicit `-s`/`--sandbox` arg > `LOOM_CODEX_SANDBOX` > skip-permissions convention > `read-only`. The fork's posture is one env var away (`LOOM_CODEX_SANDBOX=danger-full-access`), and an invalid value exits 78 rather than silently widening.

The coupling is documented prominently: `workspace-write` blocks outbound network, so a Builder-equivalent worker needs `LOOM_CODEX_NETWORK=1` — **and that flag removes most of what the sandbox was contributing**. The parity doc says so in plain words rather than burying it.

### 2. Auth is narrowed to single-profile `CODEX_HOME` passthrough

No token pool, no rotation, no `.bad_tokens` marking — that is Phase 4, and it is an explicit non-goal here. An **explicitly requested** profile with no usable `auth.json` exits **78** rather than silently degrading to a different account (which would misattribute work and cost); ambient auth is not a "request" and never fails here.

Credential hygiene, asserted by tests: the directory is **assigned** to `CODEX_HOME`, never copied; only the profile *directory name* is ever logged.

## Corrections to the fork's parity doc (verified on 0.146.0)

1. **`--full-auto` does not exist** on `codex exec`. The fork's safe mode targets it. `-s workspace-write` is the replacement.
2. **`-a`/`--ask-for-approval` is top-level only.** Any parity claim resting on `approval_policy` is inert for headless dispatch.
3. **Codex has a hook system now.** The fork says Codex has no hooks "as a concept"; 0.146.0 ships a `hooks.json` engine with `pre_tool_use` / `user_prompt_submit` / `post_tool_use` events and a hook-trust model. The gap is therefore **unwired, not impossible** — a better position than documented, and a concrete follow-up. (This also revises the issue's own AC wording, which assumed "no PreToolUse-equivalent hook interception".)
4. **A sandbox denial exits 0.** Verified: under `-s read-only` a blocked `touch` returns `Operation not permitted` to the model and the process still exits 0. Denials are in-session tool failures, so exit-code-first classification never sees them — which is *why* the classifier has no "sandbox denial" pattern. Adding one would be dead code that could only fire as a false positive.

## Error classification

Every pattern was **observed in real `codex exec` output or extracted verbatim from the 0.146.0 binary's own message strings** — nothing guessed:

- **`FATAL`** (first producer of this category) — trusted-directory refusal, `unknown configuration field`, `landlock_sandbox_executable_not_provided`. All unretryable; a `RECOVERABLE` verdict would loop forever, and a worker that cannot construct its sandbox must never be retried into running unsandboxed.
- **`TOKEN_EXHAUSTED`** — `You've hit/reached your usage limit`, `workspace credit limit`, `out of credits`, `insufficient_quota`, `rate_limit_exceeded`. A **bare 429 stays `RECOVERABLE`** (transient throttle, not exhaustion).
- **`TOKEN_EXPIRED`** — the observed `HTTP error: 401 Unauthorized` stream error, plus `Not signed in. Please run 'codex login'`, `refresh token has expired`, `Failed to refresh token`, `session has expired`, `invalid_api_key`.

Quota wording is checked **before** 401 wording: `TOKEN_EXPIRED` marks an account bad with a reason that persists until manual `loom-tokens unblock`, while `TOKEN_EXHAUSTED`'s TTL-expires — so that ordering makes mis-classification cheap in the right direction. There is a test for the co-occurring case.

`CWD_DELETED`, `SESSION_LIMIT`, and `MODEL_REFUSAL` are **left unimplemented for codex and documented as such**, because that runtime has no known wording for them and reusing Claude's phrasing would produce confident nonsense. Unmatched categories fall through to `RECOVERABLE`.

**The `claude` table is byte-identically unchanged**, with 8 regression assertions proving it — including that the new `FATAL` patterns do not leak across, and that two-arg calls (`claude-wrapper.sh`) still default to the claude table.

## Capability manifest ↔ parity doc

`codex.json` declares `worktreeIsolation: "partial"` (Codex's `workspace-write` confines to the *workspace root*, not one `issue-N` worktree — the #4178 escape class). Because `"partial"` fails closed:

```
check-runtime-capabilities.sh --role builder --runtime codex  -> 78 (incompatible)
check-runtime-capabilities.sh --role judge   --runtime codex  ->  0 (compatible)
```

Both outcomes are asserted in the CI leg. That is the intended relationship between contract points 6 and 7: **the parity doc states the residual gap in prose, the manifest makes it enforceable.** `subagents: "no"` encodes fork PR #59's native-agent prohibition; `hooks: "partial"` reflects correction 3.

## Test results

| Suite | Result |
|---|---|
| `test-spawn-codex.sh` (new) | **135/135 pass** |
| `test-spawn-worker.sh` | 26/26 pass (dispatcher unaffected) |
| `test-spawn-claude.sh` | 129/129 pass (Claude path unaffected) |
| `test-check-runtime-capabilities.sh` | 17/17 pass |
| `scripts/test-installer.sh` | 161/161 pass (new `docs/*` + `runtimes/*` auto-map) |
| `test-install-reinstall-safety.sh` | 15/15 pass |
| `bash -n` all new/changed scripts | clean |
| `shellcheck --severity=warning` across **all** `defaults/scripts/` | clean |
| `check-phantom-labels.sh` / `check-claude-md-budget.sh` | OK (314/320 lines — CLAUDE.md untouched) |
| CI leg's non-suite steps, simulated locally | all pass |

New-suite coverage: argv assembly, prompt-as-positional (the `-p`-means-`--profile` collision), sandbox mapping + full precedence chain, invalid-mode → 78, network coupling, model/effort precedence, git-repo check in all four states, args passthrough, `-p` without a value → 78, profile selection across all tiers, unusable-`auth.json` → 78, credential non-leakage, missing binary → **127 not 78**, stdout/stderr split, stdin neutralization, session/transcript/token reporting, exit-code passthrough, `LOOM_RUNTIME=codex` + `runtimes.default=codex` through the dispatcher, and the classifier vectors above.

## Live smoke run

One real read-only run through the adapter against the operator-provisioned profile:

```
$ CODEX_HOME=~/.loom/codex-profiles/r.j.walters ./defaults/scripts/spawn-codex.sh \
      -p 'Reply with exactly: ADAPTER-SMOKE-OK'
```

**Exit 0.** stdout was exactly `ADAPTER-SMOKE-OK` (nothing else). Adapter log lines:

```
spawn-codex: LOOM_SWEEP_CLAIM_OWNED=unset
spawn-codex: model=default
spawn-codex: sandbox=read-only source=adapter-default
spawn-codex: cwd is inside a git work tree — leaving Codex's trusted-directory check enabled
spawn-codex: using Codex profile 'r.j.walters' (source=CODEX_HOME)
spawn-codex: session=019faffe-302c-7283-97b2-fcf0df3427af
spawn-codex: transcript=/Users/rwalters/.loom/codex-profiles/r.j.walters/sessions/2026/07/29/rollout-2026-07-29T15-28-19-019faffe-302c-7283-97b2-fcf0df3427af.jsonl
spawn-codex: tokens_used=2502
```

The reported transcript path was confirmed to exist (36 KB JSONL), and a grep of the full stderr for `access_token` / `refresh_token` / `id_token` / `sk-` found **no credential material**.

## #4469 absorption

Per the operator's heads-up, #4469's one remaining AC — the `CODEX_HOME` profile-layout / refresh / security-posture doc section — **is absorbed here**, as a dedicated section of `guardrail-parity-codex.md` (layout + modes, the selection-precedence table, the "authoritative copy, not a cache" refresh caveat, and the security posture). `runtime-adapters.md`'s Related section points at it. **#4469 can close when this lands.**

## Non-goals respected

No daemon spawn-path migration, no provider-aware token pool, no `AGENTS.md` codegen, no role-prompt changes, no new labels, no `.codex/` config tree. The Claude path is untouched and no existing caller changes behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,3 +111,74 @@ jobs:
         run: bash scripts/test-installer.sh
       - name: Run installer reinstall-safety tests (#4188)
         run: bash defaults/scripts/tests/test-install-reinstall-safety.sh
+
+  codex-adapter-smoke:
+    name: Codex Adapter Smoke (mocked)
+    needs: changes
+    if: always() && (github.event_name == 'push' || needs.changes.outputs.scripts == 'true')
+    # Tier-2 admission gate for the Codex runtime adapter (epic #4167 Phase 2,
+    # issue #4468): a green leg here is what admits the adapter per
+    # defaults/docs/runtime-adapters.md's tier policy.
+    #
+    # NO LIVE OpenAI CALLS. The suite is hermetic by construction — it uses
+    # spawn-codex.sh's LOOM_CODEX_NO_EXEC argv-only hook plus a fake `codex`
+    # shim on a temp PATH, so the real Codex CLI is never installed or invoked
+    # and no credential is required. `codex` is deliberately NOT on PATH here,
+    # which also exercises the missing-binary (exit 127) path.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Syntax-check the Codex adapter and its test
+        run: |
+          bash -n defaults/scripts/spawn-codex.sh
+          bash -n defaults/scripts/tests/test-spawn-codex.sh
+      - name: Assert the suite cannot reach a real Codex CLI
+        # Structural guard: the test must never install, download, or otherwise
+        # obtain the real Codex CLI. Combined with `codex` being absent from
+        # this runner's PATH, that makes a live OpenAI call impossible.
+        run: |
+          if grep -nE '(npm|brew|pipx?|curl|wget) .*(codex|openai)' \
+              defaults/scripts/tests/test-spawn-codex.sh; then
+            echo "::error::test-spawn-codex.sh must not install or fetch a real Codex CLI"
+            exit 1
+          fi
+          # This leg is expected to run on a runner with no Codex CLI at all.
+          # A warning rather than a failure: if a future runner image ships
+          # `codex`, the suite is still hermetic (every dispatch test prepends
+          # its own mock to PATH, and the argv tests never exec anything), so a
+          # hard failure here would be spurious rather than protective.
+          if command -v codex >/dev/null 2>&1; then
+            echo "::warning::a real 'codex' is on PATH; the suite still mocks every dispatch, but verify no test bypasses the mock"
+          fi
+      - name: Validate the Codex runtime capability manifest
+        run: |
+          # Valid JSON with the tri-state schema check-runtime-capabilities.sh reads.
+          jq -e '.runtime == "codex" and (.capabilities | type == "object")' \
+            defaults/runtimes/codex.json > /dev/null
+          bad="$(jq -r '.capabilities | to_entries[]
+                        | select(.value != "yes" and .value != "no" and .value != "partial")
+                        | .key' defaults/runtimes/codex.json)"
+          if [[ -n "$bad" ]]; then
+            echo "::error::codex.json capability value outside yes|no|partial: $bad"
+            exit 1
+          fi
+          # Codex declares worktreeIsolation=partial, so the Builder role (which
+          # requires it) must fail closed with EX_CONFIG (78). This pins the
+          # guardrail-parity doc's residual gap 2 to an executable assertion.
+          rc=0
+          bash defaults/scripts/check-runtime-capabilities.sh \
+            --role builder --runtime codex --dir defaults >/dev/null 2>&1 || rc=$?
+          if [[ "$rc" != "78" ]]; then
+            echo "::error::builder+codex must be incompatible (expected 78, got $rc)"
+            exit 1
+          fi
+          # Judge requires only mcp (declared "yes"), so it must be compatible.
+          bash defaults/scripts/check-runtime-capabilities.sh \
+            --role judge --runtime codex --dir defaults >/dev/null
+      - name: Assert the guardrail-parity doc exists (contract point 6)
+        # No runtime is admitted without a parity doc naming its residual gaps.
+        run: |
+          test -f defaults/docs/guardrail-parity-codex.md
+          grep -q '## Residual gaps' defaults/docs/guardrail-parity-codex.md
+      - name: Run the mocked Codex adapter test suite
+        run: bash defaults/scripts/tests/test-spawn-codex.sh
--- a/defaults/docs/guardrail-parity-codex.md
+++ b/defaults/docs/guardrail-parity-codex.md
@@ -1,0 +1,296 @@
+# Guardrail Parity: Codex
+
+This is the **required guardrail-parity document** for the Codex runtime adapter
+(`defaults/scripts/spawn-codex.sh`), per contract point 6 of
+[`runtime-adapters.md`](runtime-adapters.md). No runtime is admitted to Loom
+without one. It maps **Loom guard *intent* → Codex enforcement mechanism** and
+then names, explicitly, every protection Loom has that a Codex worker does
+**not** get.
+
+Read this before dispatching a Codex worker at anything you care about.
+
+> **Provenance.** The Codex adapter is a port of the Codex support built in the
+> [gpeyton/loom](https://github.com/gpeyton/loom) fork by Graham Peyton (fork
+> PRs #15/#16/#20/#40, including its `GUARDRAIL-PARITY.md`, the template for
+> this document). Every claim below was **re-verified against codex-cli
+> 0.146.0** on 2026-07-29 (epic #4167 Phase 2, issue #4468) and several of the
+> fork's statements no longer hold on that version — those are called out
+> inline. Do not carry a claim from the fork's doc into this one without
+> re-checking the CLI.
+
+> **Path convention.** This file lives at
+> `defaults/docs/guardrail-parity-codex.md` in the Loom source repo and cites
+> `defaults/` paths. A consumer install maps `defaults/docs/` → `.loom/docs/`,
+> so the installed copy is `.loom/docs/guardrail-parity-codex.md`.
+
+## Tier status
+
+**Codex is tier-2: CI-gated, no operator dogfooding.** It passes a mocked spawn
++ classifier CI leg; it is not run against production workloads. Promotion to
+tier-1 requires someone committing to tier-1 ownership of this adapter, this
+document, and that CI leg — see the contract's tier policy. Nothing in this
+document should be read as "Codex is safe to point at your repos".
+
+## The enforcement mechanisms Codex actually has (0.146.0)
+
+| Mechanism | What it controls | How the adapter drives it |
+|---|---|---|
+| `-s` / `--sandbox <mode>` | Filesystem + network confinement for model-run shell commands. Modes: `read-only`, `workspace-write`, `danger-full-access`. | The adapter's central knob — see the mapping table below. |
+| `[sandbox_workspace_write] network_access` | Outbound network from inside a `workspace-write` sandbox. **Off by default.** | `LOOM_CODEX_NETWORK=1` → `-c sandbox_workspace_write.network_access=true`. Read only under `workspace-write`; inert otherwise. |
+| `sandbox_permissions` / `writable_roots` / `--add-dir` | Widen a `workspace-write` sandbox to extra readable/writable roots. | **Not driven by the adapter.** Passes through if an operator supplies it. |
+| `--skip-git-repo-check` | Waives Codex's refusal to run outside a git work tree. | Injected **only** when the cwd is genuinely not inside a work tree (see "Trusted-directory check" below). |
+| `$CODEX_HOME/hooks.json` (`pre_tool_use`, `permission_request`, `post_tool_use`, `user_prompt_submit`, `session_start`, `session_end`, `pre_compact`, `post_compact`, `subagent_start`, `subagent_stop`) | Per-tool-call and per-prompt interception — the direct analogue of Claude Code's hook taxonomy. | **Not wired by Loom in this phase.** This is the single most consequential residual gap; see gap 1. |
+| `approval_policy` / `-a` | When Codex pauses to ask a human. | **Irrelevant to Loom.** `codex exec` is non-interactive and exposes no `-a` at all; there is no human to answer, so approvals gate nothing. The sandbox is the only load-bearing guard. |
+| `AGENTS.md` | Repository instructions, read natively by Codex via ancestor traversal. | Advisory context, not a boundary. Loom's `AGENTS.md` codegen is a separate issue (contract point 5). |
+
+### Corrections to the fork's parity doc, verified on 0.146.0
+
+1. **`--full-auto` does not exist on `codex exec`.** The fork maps its safe mode
+   to `--full-auto`; that flag is absent from `codex exec --help` on 0.146.0.
+   `-s workspace-write` is the replacement, and it is what this adapter emits.
+2. **`-a` / `--ask-for-approval` is top-level only**, not an `exec` flag. Any
+   parity claim that rests on `approval_policy = "on-request"` is inert for
+   Loom's headless dispatch.
+3. **Codex has a hook system now.** The fork states Codex has no hooks "as a
+   concept". On 0.146.0 it has a `hooks.json` engine with a `pre_tool_use`
+   event, a persisted hook-trust model, and a
+   `--dangerously-bypass-hook-trust` escape hatch. The gap is therefore
+   *unwired*, not *impossible* — a materially better position than the fork
+   documented, and a concrete follow-up rather than an architectural dead end.
+4. **A sandbox denial does not fail the run.** Verified with `-s read-only`: a
+   blocked `touch` returns `Operation not permitted` to the model and the
+   `codex exec` process still exits **0**. Denials are in-session tool
+   failures, so they never reach error classification (see
+   `defaults/scripts/lib/classify-error.sh`'s `codex` table for why no
+   "sandbox denial" pattern is encoded there).
+
+## Loom guard intent → Codex mechanism
+
+Loom's guards are Claude Code `PreToolUse` / `Stop` hooks wired in
+`.claude/settings.json` to scripts under `.loom/hooks/`. **None of them fire for
+a Codex worker** — they are Claude Code hooks, and Loom does not currently
+install anything into Codex's own `hooks.json`. The column below therefore
+records what Codex's *sandbox* incidentally covers, not what runs.
+
+| Loom guard (`defaults/hooks/`) | Claude matcher | Intent | Codex coverage | How / why |
+|---|---|---|---|---|
+| `guard-destructive.sh` → `guard-destructive-generic.sh` | `Bash` | Deny catastrophic Bash (`rm -rf /`, force-push to `main`, `gh repo delete`, fork bombs, `curl … \| sh`, cloud/SQL destruction); ask on borderline ops; scope `rm` to the repo; Bash-tool write-confinement (`>`, `tee`, `sed -i`, `cp`/`mv`, #4178) | **partial** | `read-only` blocks every write, so under the adapter's default the destructive-write half is fully covered — more strictly than the guard itself. `workspace-write` blocks writes and `rm` outside the workspace root, and (with network off) blocks `curl \| sh` and remote cloud destruction by making the network unreachable. **Not covered:** command-pattern semantics. Codex cannot recognize `DROP DATABASE`, `DELETE` without `WHERE`, `git push --force` to `main`, or a fork bomb *as such* — anything reachable without leaving the workspace or the network proceeds. With `LOOM_CODEX_NETWORK=1` (which a Builder needs to push) the network-derived coverage evaporates and a force-push to `main` becomes reachable with nothing to stop it. |
+| `guard-worktree-paths.sh` | `Edit\|Write` | Confine Edit/Write to the builder's own `issue-N` worktree; deny escapes into the main checkout (#2441, #4007) | **partial** | `workspace-write` confines writes to the **workspace root** — a strictly coarser boundary. It blocks escaping the repo, but **not** the per-worktree boundary: a Codex builder can write into a sibling `issue-M` worktree, or into the main checkout, because all of those live under the same root. This is the exact class of escape #4178 documented. Mitigation: one Codex worker per workspace root, or narrow `writable_roots` by hand. |
+| `guard-loom-workflow.sh` | `Bash` | `gh pr merge` → `merge-pr.sh` redirect; `pip install -e` worktree block (#2495); `loom-daemon workspace` registry-mutation ask (#4326) | **none** | Pure command-pattern convention with no OS analogue. A Codex worker learns these only from `AGENTS.md` / role prompts — advisory, never enforced. |
+| `guard-background-subagents.sh` | `Stop` | Block one stop when the transcript shows dispatched-but-unresolved `Task` subagents, so ending a headless turn does not kill live background work (#4257) | **none** | Depends on Claude Code's `Stop` event and transcript shape. Codex has `session_end` and `subagent_stop` events that could host an equivalent, but nothing is wired (gap 1). Note the underlying hazard is *also* absent today: Loom does not dispatch Codex subagents at all. |
+| `guard-readonly-dirs.sh.template` | `Edit\|Write` | Optional per-project read-only path protection | **partial** | Expressible as narrowed `writable_roots` under `workspace-write`, but the adapter does not generate it — an operator must configure it manually. |
+| `skill-router.sh` | `UserPromptSubmit` | Inject an agent routing table / `AGENT_ROUTE` suggestion per prompt (opt-in) | **none** | Context injection, not a boundary. Codex has a `user_prompt_submit` hook event that could host it; unwired (gap 1). Static equivalent: `AGENTS.md`. |
+| `methodology-inject.sh` | *(present, not wired here)* | Inject universal/role/topic context from `.loom/context/` | **none** | Same as above: not a boundary, and a `user_prompt_submit` equivalent exists but is unwired. |
+| `post-worktree.sh` | *(invoked by `worktree.sh`)* | Copy the `loom-daemon` binary into a new worktree | **covered** | Runtime-neutral — `worktree.sh` runs it regardless of which runtime drives the work. Not a Claude hook. |
+
+## Sandbox-mode mapping (what the adapter emits)
+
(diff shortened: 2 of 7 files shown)


rjwalters:
## CI status: my leg is green; the two Rust legs fail **pre-existing on main**

**Green (everything this PR is responsible for):**

- ✅ **Codex Adapter Smoke (mocked)** — the new tier-2 admission gate
- ✅ Shellcheck
- ✅ Installer Integration Tests
- ✅ MCP Server Build
- ✅ CLAUDE.md Line Budget

**Failing — not caused by this PR:**

- ❌ Rust Unit Tests
- ❌ Rust Linting & Build Check

Both fail identically on this PR's **base commit** (`1085e40b`, current `main`) — verified via that commit's own check runs, which show `Rust Unit Tests: failure` and `Rust Linting & Build Check: failure` before this branch existed.

Root cause (from the job log, `loom-daemon/src/safehouse.rs`):

```
error[E0063]: missing field `death_class` in initializer of `types::Event`
    --> loom-daemon/src/safehouse.rs:2717:21
2717 |         bus.publish(Event::SweepExited {
error[E0063]: missing field `death_class` in initializer of `types::Event`
    --> loom-daemon/src/safehouse.rs:2773:21
```

Two `Event::SweepExited` initializers in `safehouse.rs`'s test module were not updated when `death_class` was added to `types::Event` — the same class of test-helper drift as #4417. It is a **compile** error in test code, so it fails both Rust legs.

This PR changes **zero** Rust files (`git diff --name-only origin/main...HEAD -- '*.rs' 'Cargo.*'` → 0). The Rust legs run at all only because `.github/workflows/ci.yml` is listed in `ci.yml`'s own `backend` paths-filter, so editing the workflow to add the Codex leg re-triggers them.

Per the Builder role's pre-existing-failure guidance I am **not** fixing it here — it belongs in its own PR against `main` (a two-line test-helper fix), and folding an unrelated Rust change into a runtime-adapter port would muddy both reviews.

rjwalters:
## Judge: APPROVED

Reviewed as a security-relevant adapter (new spawn path, sandbox policy, credential handling). I re-derived the load-bearing claims myself rather than trusting the PR body.

### Verified independently

**Contract point 1 (Spawn).** Every facet checked live against a mock `codex` on a temp PATH:

- **`-p` collision handled correctly.** `-p`/`--prompt`/`-p=`/`--prompt=` are consumed and re-delivered positionally; `-p` is never forwarded, so Codex's `-p`=`--profile` is never triggered. `--profile myprof` passes through untouched. Critically, the multi-word prompt survives as **one** argv element — `ARGV[3]=<multi word prompt here>` — the `would-exec` display flattens it, but the real argv does not.
- **Exit-code fidelity through the stderr tee.** `PIPESTATUS[0]` recovery across the `{ … } 3>&1` brace group is not obviously safe, so I tested it in isolation: exit 42 and exit 7 both propagate. Stdout purity also holds — only the agent's final message reaches stdout (`2>&1 1>&3 3>&-` with the outer `3>&1` is correct fd juggling).
- **Exit codes:** 78 for an explicitly-requested profile with no usable `auth.json`, 127 for a missing binary, 78 for `-p` with no value, 78 for an invalid `LOOM_CODEX_SANDBOX`. All confirmed.
- **stdin discipline:** `</dev/null` on the headless path only; interactive `exec` deliberately keeps stdin. Right call.
- **Session/transcript/tokens** extracted from **STDERR** (the issue's intel said stdout; the correction is the right one) — all three reported live.
- **Git trust check** injected only when `git rev-parse --is-inside-work-tree` fails, so worktree dispatch keeps the guardrail.

**Tests.** Counts verified, not trusted: `test-spawn-codex.sh` **135/135** (matches the claim), plus `test-spawn-worker` 26/26, `test-spawn-claude` 129/129, `test-check-runtime-capabilities` 17/17, `test-resolve-tier-model` 13/13. Shellcheck clean on `spawn-codex.sh`, `test-spawn-codex.sh`, and `classify-error.sh`. Dispatcher integration confirmed end-to-end via `LOOM_RUNTIME=codex spawn-worker.sh`.

**Credential hygiene — clean.** `auth.json` is only ever *stat*'ed (`-f`/`-r`/`-s`), never read or echoed. Only `basename` is logged; the one place a `CODEX_HOME` value is printed is the remediation hint, as a path. No copy of `auth.json` under `.loom/`. Tiers 1-3 are assignment-not-copy as documented.

**`_classify_error_codex()`.** Patterns are specific enough to avoid false positives, and the *absence* of a bare-429 → TOKEN_EXHAUSTED match is the right conservative choice. The TOKEN_EXHAUSTED-before-TOKEN_EXPIRED ordering is justified on the correct axis — mis-classifying quota as auth costs a manual `loom-tokens unblock`, the reverse only costs a TTL. Two things I checked beyond the diff: (a) `FATAL` is newly *producible* for the first time, and the only consumer (`claude-wrapper.sh` `log_permanent_death`) treats it as an opaque log string, so nothing breaks; (b) the `claude` table is genuinely unchanged, which the suite's cross-provider isolation vectors assert. The sandbox-denial finding is sound and well-documented: denials exit **0**, so exit-code-first classification never reaches the table, and an `Operation not permitted` pattern could only ever fire as a false positive. Leaving CWD_DELETED / SESSION_LIMIT / MODEL_REFUSAL unimplemented rather than borrowing Claude wording is the right discipline.

**Parity doc — does not overstate parity; it arguably understates it.** 9 residual gaps, all real. It corrects the fork's claim that Codex has no hooks "as a concept" (0.146.0 has an unwired `hooks.json` engine) and frames gap 1 honestly as a *wiring* gap. Gap 5 even discloses a pre-existing all-runtime hole (ungated label mutation) that isn't a Codex regression. The admission checklist leaves the tier-1 box **unchecked**. Good.

**CI leg.** Hermetic by construction: `LOOM_CODEX_NO_EXEC` argv-only hook plus a structural grep that fails if the suite ever tries to `npm`/`brew`/`curl` a real Codex. No live OpenAI calls possible. Paths-filter is correct — `scripts:` includes `defaults/**`, which covers all seven changed files. The leg also pins the capability manifest to executable assertions (builder+codex → 78, judge+codex → 0). Installer needs no change: `manifest.sh` globs `runtimes/*` and `scripts/*`.

### The sandbox judgment call — I endorse it, and I think declining the fork's mapping was correct

The reasoning holds up. Claude's `--dangerously-skip-permissions` is **not** a bare full-access grant; it is full tool access *backstopped by* `PreToolUse` guards (`guard-destructive.sh`, `guard-worktree-paths.sh`) that mechanically deny scope-violating deletions, force-pushes, and worktree escapes. Under Codex those hooks do not execute at all. Mapping the same flag to `--dangerously-bypass-approvals-and-sandbox` would therefore transfer the *permission* while silently dropping the *compensating control* — strictly weaker than Claude's posture, and calling that "parity" would be the exact reasoning error that makes a second runtime a security regression. `workspace-write` is the closest honest analogue of `guard-worktree-paths.sh`'s intent, and a `read-only` default means a mis-wired dispatch produces an inert worker rather than a destructive one.

**Is the env-var opt-in loud enough?** Yes. `LOOM_CODEX_SANDBOX=danger-full-access` is a named, non-guessable value; invalid values hard-fail 78 instead of degrading; and every spawn emits exactly one `sandbox=<mode> source=<where>` line, so the effective posture is attributable from the sweep log — a better audit trail than the Claude side currently has. The `LOOM_CODEX_NETWORK=1`-under-wrong-mode warning is a nice touch. The read-only default can't strand a Codex Builder either, since builder+codex already fails closed at 78 in the manifest.

### Non-blocking nits (no Doctor cycle needed; fold into the Phase-4 PR if convenient)

1. `spawn-codex.sh` header (~lines 20-21) says "Codex has **NO** PreToolUse-hook equivalent" — that contradicts this PR's own re-verified finding (parity doc finding 3 / gap 1: 0.146.0 *does* expose `pre_tool_use`; the real gap is that Loom doesn't wire it). The conclusion drawn is identical and correct, but a safety-critical header should carry the accurate reason, matching `runtime-adapters.md` and `codex.json`'s `hooks: "partial"`.
2. A trailing `-s` or `-m` with no value sets `HAS_*_ARG=true` with an empty value and forwards a dangling flag for Codex to reject (logged as `sandbox=unknown`). Harmless, but the `-p`-missing-value case gets a clean 78 and these could too.

### CI status

6 checks green, including `Codex Adapter Smoke (mocked)`. The two Rust legs (`Rust Unit Tests`, `Rust Linting & Build Check`) fail **byte-identically on the base commit `1085e40b`** — `error[E0063]: missing field death_class` at `loom-daemon/src/safehouse.rs:2717` and `:2773`, the known #4417-class test-helper drift. This PR touches **zero** Rust files, so it cannot be implicated; not holding for it. Merge state is `unstable` (not `dirty`) — mergeable with those pre-existing red legs, which the merge path handles.

Labels: `loom:review-requested` + `loom:reviewing` → `loom:pr`.

PR #4477, opened 2026-07-29
feat(daemon): opt-in idle exit for remote hosts

### LINKED ISSUE #4467: daemon: opt-in idle-exit so remote hosts' idle-shutdown guards can fire
## Problem

Remote workers provisioned by `/repo:remote` get an idle-shutdown guard (cron: `shutdown -h` after `REPO_REMOTE_IDLE_SHUTDOWN_MIN` of no SSH/claude/CPU activity — see loom#3998). But the guard treats the daemon process itself as activity:

```bash
if who | grep -q . || pgrep -x claude >/dev/null || pgrep -f loom-daemon >/dev/null; then
  rm -f "$STAMP"; exit 0   # "active" — idle timer never starts
fi
```

So a daemon-managed worker can **never idle out**: loom-worker-1 (c7i.2xlarge, ~$8.60/day) stays up indefinitely even with 0 sweeps in flight and the token pool exhausted. The guard can't just drop the `pgrep -f loom-daemon` check either — the daemon parks at ~0 CPU between ticks, so it would look idle even mid-deployment with sweeps queued behind rate limits.

The daemon is the only component that actually knows whether the host is doing useful work (in-flight sweeps, role-runner sessions, last dispatch, token-pool health). It should participate in idle shutdown.

## Proposed design: daemon-side *idle exit*, guard keeps power-off authority

New config block in the existing `autonomous` tier system (committable per-repo or machine-level, like `workFinder`/`roleRunner` enablement):

```jsonc
"autonomous": {
  "idleExit": {
    "enabled": false,     // default off — laptops/studio hosts keep today's behavior
    "idleMinutes": 60
  }
}
```

Behavior when enabled — the daemon considers itself **idle** when, continuously for `idleMinutes`:

- 0 in-flight sweeps, and
- no active role-runner session, and
- no dispatch/completion events

When tripped: announce to the safehouse fleet room ("idle for 60m — exiting for host idle-shutdown"), write a marker file (so ops/`loom-daemon status` tooling can distinguish idle-exit from a crash), exit 0. The host guard then sees no `loom-daemon`/`claude` processes, its own timer runs, and `shutdown -h` fires as it does today.

### Why exit-and-defer instead of the daemon running `shutdown -h` itself

- No sudoers/NOPASSWD entry, no platform coupling (EC2/GCP/bare metal all work).
- The guard keeps its human-protection checks (`who`, claude sessions, load) as the single power-off authority; the daemon just stops vetoing it.
- Composes with any host guard implementation — this is the compatibility story for heterogeneous remote hosts.

Tuning note: the guard window now stacks on top of `idleMinutes`, so `/repo:remote` provisioning (anvil) should drop `IDLE_MIN` from 120 to ~15–30 on daemon-managed hosts.

## Design wrinkles to settle

1. **Role runners defeat naive idleness.** On loom-worker-1 the five roles cycle every few minutes, so "no activity for 60m" would never trip. Options: (a) count role sessions as activity only when they produce work; (b) additionally treat **token starvation as idle** — if every pool account is rate-limited/exhausted, sweeps and roles are burning EC2 hours accomplishing nothing, which is exactly when the box should power off. Leaning (b) as a second trigger: `idleExit.onTokenStarvation: true`.
2. **No wake path.** Once the host halts, work queues until someone runs `/repo:remote` again. Acceptable for the pilot; "start stopped workers when the queue is non-empty" belongs with loom#4341/#4342 fleet tooling as a follow-up.
3. **Drain safety** (loom#4343) is mostly moot here: the daemon only exits with nothing in flight, and safehoused keeps running until actual power-off. The marker file should still record a clean drain for the roster tooling.

## Context

- Part of the multi-host pilot: loom#3979 (Phase 2), epic loom#4340
- Guard install + operator security decisions: loom#3998
- Observed on loom-worker-1 (`i-0f35973f28ed5d97f`), 2026-07-29: box up 24h+, 0 sweeps in flight, 6/8 tokens rate-limited or exhausted, load 0.08 — and no path to power-off.

---

### Verification against current main

**There are two guard variants, not one — idle-exit fixes both, differently:**

1. **anvil `/repo:remote` guard (loom#3998)** — the one quoted above; its `pgrep -f loom-daemon` check means daemon *presence* always vetoes idle. Daemon exit removes the veto directly.
2. **Loom's own fleet guard**, installed by `loom-daemon fleet add-worker --idle-shutdown-minutes` (`render_idle_shutdown()` in `loom-daemon/src/fleet/add_worker.rs`). This one does **not** veto on daemon presence — it checks `pgrep -x claude` and `active_sweeps ≥ 1` via `loom-daemon status --json`. But role-runner `claude` sessions refresh its activity stamp every cycle, so wrinkle 1 defeats it too. Daemon idle-exit fixes it because role sessions stop spawning once the daemon is gone.

**Supervisor exit semantics are inverted across platforms (critical, not in the original design):**

- The fleet systemd `--user` unit is `Restart=on-failure` (`add_worker.rs`, unit template): **exit 0 stays down** — the proposed exit-and-defer works as designed.
- macOS launchd uses `KeepAlive: SuccessfulExit`: **exit 0 triggers a relaunch**; only nonzero exits stay down (see the exit-code contract in `main.rs`, issue #4054 — SIGTERM/SIGINT deliberately exit 143/130 for this reason). An idle-exit(0) on a launchd-supervised host is a slow flap loop, not a shutdown.
- Resolution: exit 0 (correct for the systemd fleet target this feature exists for); `enabled: false` default already protects launchd/laptop hosts; the docs MUST state the caveat, and the daemon should log a loud warning if idleExit is enabled while launchd-supervised.

**Existing building blocks to reuse (don't reinvent):**

- Per-tick idle signal already exists: `work_finder.rs` computes `idle_now = dispatcher.in_flight().is_empty()` and feeds `role_runner::observe_and_fire_idle` (see `IdleEdgeTracker` in `role_runner.rs` for the edge-tracking precedent).
- Active role-runner sessions: the in-progress guard set in `role_runner.rs` (`RoleRunGuard::try_acquire`).
- Token starvation: `capacity.rs` — `read_ranking()` → `summarize_probe()` gives healthy/unhealthy account counts; trigger (b) needs no new probing.
- Marker-file precedent: `autonomy_marker.rs` (render/resolve/heal pattern).
- Fleet-room announcement: publish an event on the event bus and let the safehouse sink forward it (`safehouse.rs::event_to_envelope`) — no direct socket writes.
- Config: read `autonomous.idleExit.*` via `config_resolver` (env > config > default, matching `read_work_finder_config` / `read_role_runner_config`), with `LOOM_AUTONOMOUS_IDLE_EXIT_*` env overrides.

**Placement**: prefer a dedicated interval task spawned at daemon startup (alongside the other background loops in `main.rs`) over piggybacking the work-finder tick — `work_finder.rs` documents itself as the *sole* idle-signal source for on-idle roles, and a power-off feature should not silently require `workFinder.enabled`.

### Design resolutions (settling the open wrinkles)

1. **Role runners**: adopt option (b). Role-runner sessions count as activity for the base timer, and `idleExit.onTokenStarvation` (default `true`) is a second trigger: trips when the ranking snapshot shows **0 healthy accounts continuously for `idleMinutes` AND 0 in-flight sweeps** (never preempt live work).
2. **No wake path**: accepted for the pilot; start-on-queue belongs with the #4341/#4342 fleet tooling — out of scope here.
3. **Drain safety**: the marker file records a clean idle-exit (timestamp, trigger, counts). Rendering an `IDLE-EXITED` state in `fleet status` (which today would show `DAEMON DOWN`, `fleet/status.rs`) is optional in this PR — include only if small, otherwise file a follow-up.

### Acceptance Criteria

- [ ] New `autonomous.idleExit` config block `{ "enabled": false, "idleMinutes": 60, "onTokenStarvation": true }` read via `config_resolver` with env overrides; precedence env > config > default.
- [ ] With `enabled: false` (default), daemon behavior is unchanged — no idle-exit task runs (or it no-ops without side effects).
- [ ] When enabled: after `idleMinutes` of continuous 0 in-flight sweeps AND no active role-runner run AND no dispatch/completion events, the daemon writes the marker file, publishes a fleet-room announcement event, and exits 0.
- [ ] Token-starvation trigger: 0 healthy pool accounts continuously for `idleMinutes` with 0 in-flight sweeps trips idle-exit even while role-runner sessions are cycling.
- [ ] Any activity — dispatch, sweep completion, role-run start, a healthy account reappearing — resets the relevant idle clock.
- [ ] The daemon never idle-exits with a sweep in flight.
- [ ] Marker file is machine-readable (timestamp, trigger reason, sweep/role/token counts at exit) and distinguishes idle-exit from a crash; superseded on next startup.
- [ ] `defaults/docs/daemon-reference.md` documents the block, including the supervisor caveat (systemd `Restart=on-failure`: exit 0 stays down; launchd `KeepAlive:SuccessfulExit`: exit 0 relaunches → idleExit is only meaningful under on-failure-style supervision).
- [ ] Loud startup/enable-time log warning when idleExit is enabled under launchd supervision.
- [ ] Unit tests cover the tracker state machine: idle accrual, reset on each activity type, starvation trigger, disabled no-op.

### Scope guards

- No `shutdown`/`sudo`/power calls from the daemon — exit-and-defer only; the host guard keeps power-off authority.
- No changes to the anvil `/repo:remote` guard or its `IDLE_MIN` tuning (follow-up in anvil, per the tuning note above).
- No wake/start-on-queue tooling (#4341/#4342 follow-ups).
- `fleet status` `IDLE-EXITED` rendering is optional; drop it if the PR approaches size limits.
- Rust-only — this is daemon-internal; no Python-side mirror exists or is needed.

## Affected Files

- `loom-daemon/src/idle_exit.rs` — **(new)** config reader, idle-tracker state machine, marker writer, exit orchestration
- `loom-daemon/src/lib.rs` — register the new module
- `loom-daemon/src/main.rs` — spawn the idle-exit interval task at startup; clean exit path
- `loom-daemon/src/types.rs` — idle-exit event variant
- `loom-daemon/src/event_bus.rs` — publish the idle-exit event
- `loom-daemon/src/safehouse.rs` — map the event in `event_to_envelope` for the fleet-room announcement
- `loom-daemon/src/role_runner.rs` — small read accessor exposing the active-run count from the in-progress guard
- `loom-daemon/src/capacity.rs` — reference only (reuse `read_ranking`/`summarize_probe` for starvation)
- `loom-daemon/src/sweep_registry.rs` — reference only (`in_flight()`)
- `defaults/docs/daemon-reference.md` — document `autonomous.idleExit`
- `loom-daemon/src/fleet/status.rs` — optional: render `IDLE-EXITED` from the marker

## Test Plan

- [ ] Automated tests: unit tests in `idle_exit.rs` for tracker transitions (accrual, per-activity reset, starvation trigger, disabled no-op) and marker render/parse; run with `cd loom-daemon && cargo test idle_exit`
- [ ] Manual verification: in a scratch workspace, enable `idleExit` with `idleMinutes: 1`, verify the daemon exits 0 within ~2 minutes with the marker written and the announcement event on the bus; verify default-off leaves behavior unchanged
- [ ] Edge cases: a dispatch at minute 59 resets the clock; token starvation with a role session mid-cycle still trips only once in-flight sweeps reach 0; launchd-supervision warning fires when applicable

rjwalters:
## Summary

- add default-off autonomous.idleExit config with env overrides
- track ordinary idleness and token starvation independently of Work Finder
- write a machine-readable marker, publish a safehouse fleet event, and exit 0 without issuing power commands
- expose active role-run count and preserve the fleet guard as stage-two power authority
- document systemd/launchd supervision semantics

## Validation

- env RUSTC_WRAPPER= cargo test idle_exit::tests
- env RUSTC_WRAPPER= cargo test fleet::add_worker::tests
- env RUSTC_WRAPPER= cargo check --all-targets
- cargo fmt --all -- --check
- env RUSTC_WRAPPER= cargo clippy --all-targets -- -D warnings

Closes #4467

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -494,8 +494,9 @@ the plan/ordering/checklist; the per-phase shell is rendered in
    as the **#4292** token-pool-cwd workaround — the rendered unit carries a
    `#4292` marker so it is removed when that lands.
 9. **idle-shutdown** (optional, `--idle-shutdown-minutes N`) — a cron guard that
-   powers the host off after N idle minutes, skipping while claude / loom-daemon
-   are working.
+   powers the host off after N idle minutes. This is stage 2:
+   `autonomous.idleExit` is stage 1. On daemon-managed hosts use a short guard
+   window (typically 15–30 minutes); the running-daemon veto remains.
 10. **safehouse** (optional, `--safehouse`) — **skip-with-notice** until #3998
     (the safehoused provisioning fragment) lands.
 11. **verify** — `loom-daemon status` sane from the workspace cwd, ranking
@@ -2046,6 +2047,9 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.roleRunner.roles` | *(config only)* | all 5 roles | Subset of `champion`/`curator`/`judge`/`auditor`/`guide` to dispatch; explicit empty array runs none. Also resolved from each root's own config |
 | `autonomous.roleRunner.intervalSecs` | `LOOM_ROLE_RUNNER_INTERVAL_SECS` | per-role built-in (5–15 min) | Uniform override applied to every enabled role's cadence |
 | `autonomous.roleRunner.onIdle` | *(config only)* | `[]` (none) | Subset of the same 5 roles to also fire on the work-finder **idle edge** (#4364) — the non-idle → idle transition (0 in-flight sweeps AND nothing dispatched this tick), in addition to the interval cadence. Absent → none (opposite default from `roles`); unknown names ignored with a warning. Debounced to min 60s per (root, role) and skipped while that role's interval/idle run is in progress. **Requires the work finder enabled** to observe idleness (a startup warning fires if set with the work finder off). **Also gated by that same root's own `enabled`** (#4377) — see below |
+| `autonomous.idleExit.enabled` | `LOOM_AUTONOMOUS_IDLE_EXIT_ENABLED` | `false` | End the daemon cleanly after the idle window so a host guard can take over. Independent of Work Finder; never invokes a power command |
+| `autonomous.idleExit.idleMinutes` | `LOOM_AUTONOMOUS_IDLE_EXIT_MINUTES` | `60` | Continuous idle/starvation window. Zero/invalid → default |
+| `autonomous.idleExit.onTokenStarvation` | `LOOM_AUTONOMOUS_IDLE_EXIT_ON_TOKEN_STARVATION` | `true` | Also exit after zero healthy accounts for the full window with no sweep in flight, even if roles keep cycling |
 | `autonomous.watchMonitor.enabled` | `LOOM_WATCH_MONITOR` | `true` | Durable operator-watch monitor loop (#3971). Default-on; no dispatch side effect, zero forge calls until a watch is registered |
 | `autonomous.watchMonitor.intervalSecs` | `LOOM_WATCH_MONITOR_INTERVAL_SECS` | `120` | Watch poll cadence. Zero/invalid → default |
 | `autonomous.watchMonitor.expirySecs` | `LOOM_WATCH_MONITOR_EXPIRY_SECS` | `86400` | Give-up window for an unresolved watch; `0` disables expiry |
@@ -2063,6 +2067,32 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.autoUpdate.enabled` | `LOOM_AUTO_UPDATE` | `false` | Autonomous self-update loop on/off (#4055). **Opt-in** (it rebuilds + restarts the daemon process). Exactly one loop per daemon, not a per-workspace fan-out. See [Autonomous self-update loop](#autonomous-self-update-loop-4055) below |
 | `autonomous.autoUpdate.intervalSecs` | `LOOM_AUTO_UPDATE_INTERVAL_SECS` | `900` | Cadence between staleness checks. Zero/invalid → default |
 | `autonomous.autoUpdate.settleSecs` | `LOOM_AUTO_UPDATE_SETTLE_SECS` | `600` | Settle window: wait this long after first observing a stale commit — resetting on every further commit — before rolling, so a burst of merges collapses into one roll. Zero/invalid → default |
+
+### Idle exit for remote hosts (#4467)
+
+```json
+{
+  "autonomous": {
+    "idleExit": {
+      "enabled": false,
+      "idleMinutes": 60,
+      "onTokenStarvation": true
+    }
+  }
+}
+```
+
+Ordinary idleness requires zero in-flight sweeps, no active role run, and no
+dispatch/completion event for the full window. Token starvation has a separate
+clock: role sessions do not reset it, but live sweeps and healthy-account
+recovery do. On either trigger the daemon writes
+`<loom-dir>/idle-exit.json`, publishes `daemon.idle_exit`, removes its socket,
+and exits 0. It never calls `shutdown`, `sudo`, or a provider API.
+
+The fleet add-worker systemd unit uses `Restart=on-failure`, so exit 0 stays
+down. macOS launchd uses `KeepAlive:SuccessfulExit`, so exit 0 relaunches the
+daemon; idle exit is meaningful only under on-failure-style supervision. A
+loud warning is logged when it is enabled under launchd.
 
 ### Daemon log path override (`LOOM_DAEMON_LOG`, #4010)
 
--- a/loom-daemon/src/fleet/add_worker.rs
+++ b/loom-daemon/src/fleet/add_worker.rs
@@ -634,15 +634,15 @@ systemctl --user enable --now loom-daemon.service
 }
 
 fn render_idle_shutdown(minutes: u32) -> String {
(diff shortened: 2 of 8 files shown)


rjwalters:
**Judge: changes requested**

The idle loop only samples active role count every 15 seconds. A short role invocation can start and finish between samples, so the ordinary idle clock is not reset even though the acceptance criteria explicitly require any role-run start to reset it.

Doctor request: expose a monotonic role-run-start generation from the shared role guard, sample generation changes in the idle task as lifecycle activity, and add a focused test. Please also add a direct DaemonIdleExit-to-safehouse envelope mapping test while touching this path.

rjwalters:
**Doctor:** addressed the Judge finding in commit 30d16fd5. Every successful role start now increments a monotonic generation, so even a role that completes between 15-second samples resets the idle clock. Added focused generation coverage and a direct safehouse envelope mapping assertion. Focused tests, formatting, and clippy pass.

rjwalters:
**Judge: APPROVED**

Second-pass review confirms the short-role-start gap is fixed by the monotonic generation signal. The idle tracker, starvation safety gate, typed event and safehouse mapping, marker behavior, fleet guard staging, default-off config, and supervisor warning align with the acceptance criteria. Refreshed CI is fully green, including cargo test --workspace, Rust lint/build, installer integration, and adapter smoke checks.

Issue #4257, opened 2026-07-28
sweep: orchestrator ended its turn while Judge ran in background — headless -p exit killed the review mid-flight (#3822 trap observed live)


rjwalters:
## Problem

The #3822 async-dispatch trap was observed live in a headless daemon-dispatched sweep: the sweep orchestrator dispatched its Judge subagent **in the background and then ended its turn**. In `claude -p` mode, ending the turn exits the process — which killed the still-running Judge mid-review and left a stale `loom:reviewing` claim on the PR.

## Evidence (2026-07-28, sweep-issue-4195)

`.loom/logs/sweep-issue-4195.log` — the sweep's **final output** (i.e., the text it printed as it exited) reads verbatim:

    The Judge is now reviewing PR #4243 in the background. I'll wait for its
    verdict before deciding whether to proceed to merge or dispatch a Doctor.

There is no subsequent output. The process exited at 16:54Z (log mtime), the Judge died with it, and PR #4243 sat at `loom:review-requested` (+ a stale claim) until an operator re-judged it ~1.5h later.

The sweep skill already contains the rule this violated ("Subagent dispatch is async-only — you MUST block explicitly", #3822): the orchestrator must await each subagent's completion via a blocking wait, never end its turn expecting a background child to finish. The model followed the letter ("I'll wait") while doing the opposite (ending the turn *is* the exit in `-p` mode).

## Why the current wording fails

The #3822 section frames the hazard as *ordering* ("don't start Judge before builders finish"). It does not state the sharper headless consequence: **in `claude -p`, ending your turn terminates the process and kills every background child**. A model that believes "the Judge runs in the background; I'll wait" has no cue that its own final message is the kill signal.

## Proposed direction

1. **Skill hardening (cheap, do first):** add an explicit rule to the sweep skill's #3822 section and the dispatch sites: *never* dispatch Curator/Builder/Judge/Doctor with `run_in_background` in a sweep; in headless `-p` mode, ending the turn kills background children — the final message must only be written after every dispatched subagent has returned.
2. **Mechanical backstop (evaluate):** a Stop-path hook that detects live background subagent tasks when the top-level session is about to end and emits a loud warning (or blocks once) so the orchestrator gets a chance to await instead of exit.
3. **Claim hygiene tie-in:** the stale `loom:reviewing` left behind is covered by the companion label-staleness issue.

## Acceptance criteria

- [ ] Sweep skill text names the headless turn-end == process-exit consequence explicitly at the dispatch sites.
[...]

PR #4487, opened 2026-07-30
fix(guards): recognize real background-Bash task-notification shapes (#4482)

### LINKED ISSUE #4482: guards: guard-background-subagents.sh false-positives on every stop after a completed run_in_background Bash task — completion <task-notification> format not recognized
## Context
Observed live across one interactive session (2026-07-29): `guard-background-subagents.sh` (the Stop hook from #4257/#4389) blocked **four separate stop sequences** with the identical message — "1 background Bash command(s) (run_in_background) have no completion notification in this transcript yet" — for a single background Bash task (`brew install --cask codex`, task id `bmevhdrvc`) whose completion notification (`<task-notification>` … `status: completed`, exit 0) HAD arrived in the transcript hours earlier, immediately after dispatch.

The guard's stated contract is "blocks at most once per stop sequence," which held per-stop — but every subsequent turn's stop re-triggered the same false positive, because the guard's transcript scan evidently fails to match the completion-notification format that background **Bash** tasks emit (`<task-notification>` system-turn blocks), while presumably matching whatever format subagent completions use. Net effect: after ANY `run_in_background` Bash command in a session, every stop for the rest of that session requires a second attempt, and the guard's signal value decays to noise (an operator/model trained to "just stop again" will also bypass a TRUE positive).

## Repro sketch
1. Interactive session, run any Bash tool call with `run_in_background: true`; let it complete and observe the `<task-notification>` arrive.
2. End the turn → Stop hook blocks with "no completion notification in this transcript yet".
3. Stop again (passes, per the once-per-sequence rule). Every later turn's first stop blocks again the same way.

## Verified root cause

`defaults/hooks/guard-background-subagents.sh:122-136` (background-Bash "resolved" detection) computes `$notified_ids` with:

```jq
[ .[]?
    | select(.type=="user")
    | .message.content as $c
    | ( if ($c|type) == "string" then [$c]
        else [ $c[]? | (.content? // empty) | select(type=="string") ]
        end )
    | .[]?
    | (capture("<tool-use-id>(?<id>[^<]+)</tool-use-id>")?).id // empty
  ] as $notified_ids
```

This only scans transcript lines with top-level `.type=="user"`, and only looks inside `.message.content`. **That is not where the harness writes a background-Bash `<task-notification>` completion.** Verified against live `~/.claude/projects/**/*.jsonl` session transcripts (multiple sessions, multiple `run_in_background` dispatches, e.g. task id `by0pzty60` in a 2026-07-13 lean-genius transcript): a background Bash completion notification lands as **one or both** of these top-level entry shapes, never as a `type=="user"` message:

1. `{"type":"queue-operation","operation":"enqueue", "content":"<task-notification>\n<task-id>...</task-id>\n<tool-use-id>toolu_...</tool-use-id>\n...\n<status>completed</status>\n...\n</task-notification>", ...}` — the notification text is the top-level `.content` **string field**, not nested under `.message`.
2. `{"type":"attachment","attachment":{"type":"queued_command","commandMode":"task-notification","prompt":"<task-notification>...</task-notification>"}, ...}` — the notification text is `.attachment.prompt`, again not under `.message.content`.

Because the guard's jq filters on `select(.type=="user")` before descending into `.message.content`, it never sees either shape — `$notified_ids` is effectively always empty for background-Bash completions in a real transcript, so every `run_in_background` dispatch id remains in `$bg_ids - $notified_ids` = unresolved for the rest of the session. That exactly matches the observed symptom: not "occasionally misses a notification" but **unconditionally** re-flags the same already-completed task on every later stop sequence, because there is no code path by which it can ever be marked resolved.

**Why the existing regression test didn't catch this**: `defaults/hooks/tests/test-guard-background-subagents.sh:84` builds its "resolved" fixture (`BG_NOTIFICATION_RESOLVED`) as `{"type":"user","message":{"role":"user","content":"<task-notification>...</task-notification>"}}` — i.e. it encodes the *same wrong assumption* the implementation makes (`type=="user"`, string `message.content`). The test therefore validates the guard's jq against a synthetic shape that does not occur in a real transcript, so it passes green while the guard is broken against the actual harness. The regression fixture must be rewritten to match one of the two real shapes above, not just re-confirm the current (incorrect) code.

### Contrast: why the subagent-Task matcher (pattern 1) does NOT have this bug

`defaults/hooks/guard-background-subagents.sh:110-116` diffs Task `tool_use` ids against `tool_result` ids inside `select(.type=="user") | .message.content[]? | select(.type=="tool_result")`. This works because the Claude Code `Task` tool call is **synchronous/blocking** — the harness only returns a `tool_use`'s matching `tool_result` when the subagent has actually finished, so a `type=="user"` message really is where Task completion lands. Background `Bash` (`run_in_background: true`) is different by design: its `tool_result` is an **immediate dispatch-time ack** ("Command running in background with ID: ..."), not completion (this asymmetry is why pattern 2 exists at all — see the file's header comment, lines 26-36). The *real* completion event for a background Bash task is the async `<task-notification>`, and that event is written to a structurally different transcript entry (`queue-operation` / `attachment`, not `user`) than the dispatch-time ack. Pattern 1's approach (match inside `type=="user"` content) is correct for its own signal; pattern 2 copied that lookup structure without accounting for the different entry type its signal actually arrives in.

## Refined acceptance criteria

- [ ] The guard's `$notified_ids` computation in `defaults/hooks/guard-background-subagents.sh` recognizes background-Bash completion notifications in the actual harness-emitted shapes: `type=="queue-operation"` top-level `.content` string, and/or `type=="attachment"` `.attachment.prompt` string (matching whichever the installed harness version emits — cover both for forward/backward compatibility, since sample transcripts show both being written for the same event). Extracting the `<tool-use-id>ID</tool-use-id>` tag from either shape must resolve the corresponding `$bg_ids` entry.
- [ ] A background-Bash task whose notification was matched is never re-flagged as unresolved by a later stop sequence in the same session (this now falls out naturally once matching works, since the guard re-scans the full transcript fresh on every Stop and the notification event doesn't disappear — no new state/cache needed, just correct matching).
- [ ] Regression fixture(s) in `defaults/hooks/tests/test-guard-background-subagents.sh` are corrected to use the real transcript shapes (`queue-operation` and/or `attachment`, per above) instead of the current incorrect `type=="user"` string-content fixture, and assert `allow` (exit 0, no block) once the notification is present.
- [ ] Add a second regression case: multiple stop sequences (i.e. call the hook twice against the same completed-and-notified transcript) both allow — guards against a regression where matching only holds on the "first" scan.
- [ ] TRUE-positive behavior unchanged: a background-Bash dispatch with **no** matching notification (in any recognized shape) still blocks — keep/extend existing test case (h) in the same test file.
- [ ] `defaults/docs/guard-hooks.md` (+ `.loom/docs/guard-hooks.md` installed mirror) updated to describe the corrected notification-matching shapes if the "Background Bash tasks (#4389)" section's description of the matcher changes materially.

## Affected Files
- `defaults/hooks/guard-background-subagents.sh` (source of truth — fix the `$notified_ids` jq, lines ~122-136)
- `.loom/hooks/guard-background-subagents.sh` (installed mirror — keep byte-identical to defaults/, per this repo's own dogfooding convention; see PR #4425 which touched both)
- `defaults/hooks/tests/test-guard-background-subagents.sh` (fix `BG_NOTIFICATION_RESOLVED` fixture + add multi-stop-sequence and shape-coverage cases)
- `defaults/docs/guard-hooks.md` and `.loom/docs/guard-hooks.md` (update the "Background Bash tasks (#4389)" description if the matched shape changes)

## Test Plan
1. `./defaults/hooks/tests/test-guard-background-subagents.sh` — all cases pass, including the corrected/added ones above.
2. Manually construct a transcript fixture using each real shape (`queue-operation` top-level `.content`, and `attachment.prompt`) with a `<task-notification>` whose `<tool-use-id>` echoes a prior `run_in_background` Bash dispatch id, feed it to the hook via a synthetic Stop payload (`transcript_path` pointing at the fixture, `stop_hook_active:false`), and confirm exit 0 with no `decision:block` output.
3. Confirm the existing true-positive case (dispatch + dispatch-time ack only, no notification) still blocks.
4. Confirm a second hook invocation against the same (now-resolved) transcript also allows — i.e. the fix isn't a one-shot/stateful patch that only works on the first scan.

rjwalters:
Closes #4482

## Problem
`guard-background-subagents.sh` (Stop hook, #4257/#4389) false-positive-blocked on **every** stop for the rest of a session after any `run_in_background` Bash task, even one that had completed hours earlier. Its `$notified_ids` jq only scanned `type=="user"` `.message.content`, but a background-Bash completion `<task-notification>` is **never** a user message — the harness writes it as a top-level `type=="queue-operation"` entry (text in `.content`) and/or a `type=="attachment"` entry (`commandMode:"task-notification"`, text in `.attachment.prompt`). So `$notified_ids` was effectively always empty for background-Bash completions and every dispatch id stayed in `$bg_ids - $notified_ids` = unresolved forever.

The existing regression test didn't catch this because its `BG_NOTIFICATION_RESOLVED` fixture encoded the *same wrong assumption* the code made (`type=="user"`, string `message.content`) — a shape that never occurs in a real transcript.

## Fix
Rewrite the `$notified_ids` computation to scan all three shapes and extract the `<tool-use-id>` from whichever the harness emitted:
1. `type=="queue-operation"` → `.content` string
2. `type=="attachment"` → `.attachment.prompt` string
3. legacy `type=="user"` `.message.content` (kept for forward/backward compatibility)

No new state/cache: the guard re-scans the full transcript fresh on every Stop and the notification event does not disappear, so correct matching naturally re-holds on every scan.

## Tests
`defaults/hooks/tests/test-guard-background-subagents.sh` — replaced the incorrect fixture with real `queue-operation` + `attachment` shapes and added regression cases:
- (i) queue-operation shape → allow
- (i2) attachment shape → allow
- (i3) both shapes present for the same event → allow
- (i4) second stop sequence over the same completed transcript → allow again (guards against one-shot/stateful matching)
- (h) true-positive (dispatch + ack only, no notification) → still blocks, unchanged

All 20/20 cases pass.

## Files
- `defaults/hooks/guard-background-subagents.sh` (source) + `.loom/hooks/` mirror (byte-identical)
- `defaults/hooks/tests/test-guard-background-subagents.sh`
- `defaults/docs/guard-hooks.md` + `.loom/docs/` mirror (byte-identical)


rjwalters:
✅ **Approved!**

Verified the fix directly:
- Ran `defaults/hooks/tests/test-guard-background-subagents.sh` — all 20/20 pass, including the new regression cases (i, i2, i3, i4) covering the real `queue-operation` and `attachment` notification shapes plus the re-scan-on-second-stop guard.
- `shellcheck` clean on both the hook and the test script.
- `.loom/hooks/` and `.loom/docs/` mirrors are byte-identical to their `defaults/` counterparts.
- CI (`gh pr checks`) all pass; `mergeStateStatus` is CLEAN.
- PR description correctly uses "Closes #4482".

Root cause and fix are clearly explained, and the new fixtures actually exercise the real harness shapes instead of the previous self-confirming (and wrong) `type=="user"` fixture. Nice catch on the re-scan-per-stop regression case too (i4) — that's exactly the kind of thing that would've silently reverted next session.

## Test Execution

No explicit "## Test Plan" section in the PR description; ran the included hook test suite directly (see above) as the natural verification path.

rjwalters:
✅ **Approved!**

Verified independently (not just re-reading the diff):

- Ran `defaults/hooks/tests/test-guard-background-subagents.sh` locally — **20/20 pass**, including the new (i)/(i2)/(i3)/(i4) regression cases for the `queue-operation` and `attachment` notification shapes.
- Confirmed `.loom/hooks/guard-background-subagents.sh` and `.loom/docs/guard-hooks.md` are byte-identical to their `defaults/` counterparts.
- `shellcheck` on the hook script: clean.
- Manually re-derived the new `$notified_ids` jq expression against a synthetic `queue-operation` entry — correctly extracts the `<tool-use-id>` from `.content`, confirming the shape-matching logic works as described.
- Root cause explanation (old fixture assumed `type=="user"` for task-notifications, a shape that never occurs live) is credible and the fix's fallback list (queue-operation → attachment → legacy user) is reasonable defense-in-depth.

## Test Execution

No explicit "## Test Plan" section in the PR body, but the "Tests" section doubles as one — all listed cases were reproduced above.

CI: all applicable checks pass (Rust jobs skipped — no Rust files touched). Merge state: CLEAN.

rjwalters:
**Champion Auto-Merge**

- Size check passed (114 lines: +93/-21)
- No critical files modified (guard hook + docs + tests only)
- No merge conflicts
- All CI checks passing (Rust jobs skipped by path filter)

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #4377, opened 2026-07-29
roleRunner: per-root enabled gate is silent — registered workspaces with role runner off get zero ticks and zero diagnostics


rjwalters:
## Context
Found 2026-07-29 while enabling the just-merged #4364 (`roleRunner.onIdle`) on a live multi-workspace host. The interval role loops are enabled by the **daemon's own workspace** master switch (`autonomous.roleRunner.enabled` in the workspace the daemon was started from) and then tick **every registered workspace**. But the idle-fire path (`plan_idle_runs` → `resolve_enabled(config)` where `config = read_role_runner_config(root)`) gates on the **target root's own** config.

Net effect observed: safehouse received interval champion ticks all day with no `autonomous` block in its config, yet setting only `onIdle: ["champion"]` there did nothing — idle-firing additionally required `enabled: true` in safehouse's own `.loom/config.json`. Two enablement models for the same role on the same root is surprising split-brain, and the failure mode is silent (onIdle set, nothing fires, no log).

## Suggested acceptance criteria
- [ ] Pick one semantic: either idle-fire inherits the same effective enablement as the interval loops for that root (recommended — least surprise), or the per-root gate stays and is loudly documented in daemon-reference.md's onIdle rows with a startup/log warning when onIdle is set on a root whose per-root enablement is off.
- [ ] A log line exists for "idle edge seen but suppressed (role runner disabled for this root)" so the silent-no-op is diagnosable.
- [ ] Test covering the cross-config case (master-enabled daemon workspace + onIdle-only target root).

---

## Curator Enhancement (2026-07-29)

> **Authoritative rescope.** The filer's follow-up comment (2026-07-29) supersedes the original analysis above: there is **no interval-vs-idle enablement asymmetry**. Both paths gate per-root on `resolve_enabled(read_role_runner_config(root))` — the interval multi-role loop at `role_runner.rs` ~line 1042, the idle path in `plan_idle_runs` ~line 808. The daemon-workspace master switch (`main.rs` ~line 1663) only decides whether the loops start at all. The original AC 1 ("pick one semantic") is therefore moot — the semantics are already consistent. What remains, and what bit in practice, is **observability**: a registered root whose per-root `enabled` is off is completely inert with no operator-visible signal (observed live: 9 of 10 registered workspaces on one host receiving zero role ticks, no log, no status indication).

### Verified against origin/main (2026-07-29)

[...]

Issue #4501, opened 2026-07-30
daemon: children die permanently on "Fable 5 limit" — RECOVERABLE classification not retried, and dispatch inherits the interactive fable default instead of pinning a work model


rjwalters:
## Context
Observed live 2026-07-30 ~00:40 UTC while diagnosing "daemon dispatches but in-flight stays 0" (the fleet-goal blocker): every daemon-dispatched child died in ~2s at CLI start with `You've reached your Fable 5 limit. Run /usage-credits to continue or switch models with /model.` The wrapper log shows:

```
[ERROR] Non-transient error detected - not retrying
[ERROR] === claude-wrapper: permanent death (non-transient error) ===
[ERROR] exit_code=1 classification=RECOVERABLE
```

Two compounding defects:

1. **Model-limit deaths classify RECOVERABLE but the wrapper treats them as non-transient and never retries.** A per-model usage ceiling is precisely the account-rotation case — the child should die, mark the account/model pair constrained, and be re-spawned on the next token (or with a different model), not permanently die. The classification and the retry decision disagree with each other in the same log block.
2. **Daemon-dispatched children inherit the interactive default model (Fable 5).** Work-finder/`dispatch_sweep` spawns without `--model`, so the child session runs whatever the account's CLI default is — on this host that is `fable`, which (a) burns the most constrained quota tier on routine builder work and (b) violates the spirit of the tier policy (Fable is the escalation ceiling, not the floor). Daemon dispatch should pin a sensible work-model default (e.g. `sonnet`, or resolve via `sweep.tierModels`/`sweep.optimization`) unless explicitly overridden.

The stale-`.ranking` selection compounding it (exhausted account picked repeatedly) was mitigated live with `loom-tokens check --ranking`, but ranking freshness alone does not fix either defect above: the 5h/weekly *Fable* ceiling can be hit on an account whose overall utilization still ranks it healthy.

## Suggested acceptance criteria
- [ ] `classify-error.sh` (and the Rust classifier) gains an explicit `MODEL_LIMIT` (or reuses TOKEN_EXHAUSTED with a model dimension) for the "reached your <model> limit" pattern
- [ ] claude-wrapper treats that classification as transient-with-rotation: exit in a way the daemon reaper/spawn path retries on a different account (or lower model), never "permanent death"
[...]

Issue #3936, opened 2026-07-25
loom-daemon status: token-capacity summary contradicts per-token table (says 1 exhausted while table shows 5)


rjwalters:
Observed live on v0.15.0 during the multi-repo canary rollout (daemon started with `--work-finder --health-gate`, 4 managed repos).

`loom-daemon status` printed:

```
Token capacity:
  6/7 accounts healthy, 1 exhausted/near-ceiling (from .loom/tokens/.ranking)
  not token-bound (tokens are not the current bottleneck)
...
Per-token usage:
  ACCOUNT                STATUS               5h       7d
  ------------------------------------------------------
  rjwalters-gmail        available            0%      56%
  agent3-2amlogic        available            0%      99%
  robb-2amlogic          exhausted            0%     100%
  robb-integratedplasmonics exhausted          0%     100%
  agent1-2amlogic        exhausted            0%     100%
  agent2-2amlogic        exhausted            0%     100%
  agent4-2amlogic        exhausted            0%     100%
```

The summary line claims **6/7 healthy, 1 exhausted** while the per-token table right below it shows **2 available / 5 exhausted**. Also note the earlier 'Dynamic concurrency cap' block reported 'healthy tokens 6', so the cap computation appears to consume the same wrong count — meaning the daemon may over-subscribe a mostly-exhausted pool (backpressure #3907 would engage later than it should).

Likely shape: the summary/cap path and the table path read different sources (stale vs fresh `.loom/tokens/.ranking`, or one path applies the 95% near-ceiling threshold and the other doesn't — 'agent3' at 99% 7d is listed 'available' in the table, which also looks wrong against the documented `7d_utilization >= 0.95 ⇒ exhausted` rule).

## Acceptance criteria
- [ ] The 'Token capacity' summary count, the 'healthy tokens' input to the dynamic cap, and the 'Per-token usage' table are computed from the same source and threshold rules, and agree with each other.
- [ ] The 95% near-ceiling threshold is applied consistently (an account at 99% 7d must not be counted healthy anywhere).
- [ ] A unit test covers the summary/table consistency with a mixed pool (some exhausted, some near-ceiling, some available).

Found during v0.15.0 canary rollout; filing per the canary protocol (friction → issue → let the daemon repair it).

PR #4506, opened 2026-07-30
feat(fleet): sync-labels.sh --repo flag + "bringing a new repo online" runbook

### LINKED ISSUE #4498: fleet: enable role-runner + seed backlogs for klayout-tools and the gf180-*/sky130-* canary repos (work supply for multi-repo dispatch)
## Context
Fleet goal (operator, 2026-07-30): daemons on robb-pro, robb-studio, and loom-worker-1 all reporting successful PRs on loom, klayout-tools, and the gf180-*/sky130-* canary repos. The daemon on robb-pro manages 11 repos, but `loom-daemon status` shows the ROLES column **off** for klayout-tools and all seven canary repos — no Curator/Judge/Champion/Auditor/Guide runs there, so those queues never fill and dispatch has nothing to do. Their installed Loom copies are also on the v0.15 canary rollout vintage (known friction: #3936–#3939).

## Root cause / config surface (verified)

The per-root role-runner gate is exactly what PR #4410 (closes #4377) made diagnosable:

- `autonomous.roleRunner.enabled` / `.roles` / `.onIdle` is resolved **per registered root**, from **that root's own** effective config — `read_role_runner_config(root)` → `resolve_enabled(config)` (`loom-daemon/src/role_runner.rs:532`, `:576`; struct at `:506`). The daemon workspace's own `enabled: true` (this repo's `.loom/config.json`) does **not** propagate to other registered roots — confirmed in prose at `defaults/docs/daemon-reference.md:2752-2762` ("`enabled` is resolved per root, not inherited (#4377)... Set `autonomous.roleRunner.enabled: true` in that root's own config to opt it in").
- "Effective config" for a root is a 4-tier merge, lowest → highest priority (`loom-daemon/src/config_resolver.rs:143-153`):
  1. **Private/shared machine-level defaults** — `~/.local/share/loom/config/defaults.json` (override via `LOOM_CONFIG_DEFAULTS_FILE`; path resolved at `config_resolver.rs:53-62`, `DEFAULT_PRIVATE_DEFAULTS_REL` at `:46`). This file lives **outside every git checkout**, in the operator's home directory — it is the Epic #3835 Phase 3 "machine-level checkout" surface documented in `defaults/docs/machine-dispatcher.md`.
  2. that root's own `.loom/config.json` (the "legacy" tier every repo has today)
  3. `.loom-project/project.json` (tracked, Epic #3835 Phase 6 — empty on every repo until migrated)
  4. `.loom-local/local.json` (git-ignored, host-local override)
- Consequence: there are **two valid ways** to flip the 8 target roots on, and they have very different blast radii:
  - **Per-repo**: add an `autonomous.roleRunner` block to each target repo's own `.loom/config.json` (8 separate edits/PRs, one per repo, each repo's own maintainers/CI).
  - **Machine-level**: add one `autonomous.roleRunner` block to `~/.local/share/loom/config/defaults.json` on each host (robb-pro, robb-studio, loom-worker-1) — applies to every registered root on that host that doesn't already set its own `roleRunner.enabled` (this loom repo already does, so it is unaffected either way).
  - Both are **host-local filesystem writes** (a home-directory file, or another repo's own `.loom/config.json` in its own checkout) — **neither is reachable by a Builder working inside this repo's worktree** (worktree-isolation guard confines Edit/Write to the current worktree; there is no local checkout of klayout-tools/gf180-*/sky130-* here, and `~/.local/share/loom/` is outside any repo entirely).

## What a Builder in `rjwalters/loom` CAN deliver (automatable core)

1. **Forge-side label sync**, which needs no local checkout of the target repo at all — `gh label create|edit|delete -R owner/repo` operates purely over the GitHub API. The existing `defaults/scripts/sync-labels.sh` (symlinked as `.loom/scripts/sync-labels.sh`, so one edit covers both) already implements exactly this create/update/delete-default-labels logic, but only against the **current directory's** git remote (`forge_get_repo_nwo`, `sync-labels.sh:96`) — it requires `cd`-ing into a checkout of the target repo. Add an explicit `--repo OWNER/NAME` flag that bypasses `forge_detect`/`cd` and sets `REPO`/`FORGE_TYPE=github` directly, so the script can be run 8 times from *this* repo's checkout (using *this* repo's `.github/labels.yml` as the source of truth) without cloning klayout-tools or any canary repo locally. Note: there is no `gh label sync` subcommand in the real `gh` CLI — the issue's original wording was imprecise; the actual mechanism is this repo's own `sync-labels.sh`.
2. **Runbook**: a new "Bringing a new repo online" subsection in `defaults/docs/daemon-reference.md`, immediately after the per-root `enabled`-not-inherited callout (after line 2762 / the paragraph ending "...to opt it in", before `### Durable operator watches` at line 2783). Distill, in order: (a) which config surface to use (per-repo `.loom/config.json` vs. the machine-level `defaults.json`, and when each is appropriate), (b) the exact `autonomous.roleRunner` JSON block to add (mirror the example already at `defaults/docs/daemon-reference.md:1988-1994`), (c) the label-sync step via the new `sync-labels.sh --repo` flag, (d) a pointer to `loom migrate` / the machine-level install refresh for stale (v0.15 vintage) copies, referencing the #3936–#3939 friction list so it isn't repeated blind, (e) verification: `loom-daemon status` ROLES column, and expect a `loom:triage`-labeled issue within one Auditor/Curator cycle.
3. Tests for the new `--repo` flag (unit-level: flag parsing / REPO resolution short-circuits `forge_detect`; can't exercise a real `gh api` call in CI, so assert on the resolved `REPO`/skip-cd behavior, matching the style of existing script tests under `tests/install/`).

## What stays a documented **manual** operator step (not in this PR's AC)

These require actions on the actual daemon hosts (robb-pro/robb-studio/loom-worker-1) against actual other-repo checkouts, and are out of a Builder's reach from this repo:

- Editing each target repo's `.loom/config.json` **or** each host's `~/.local/share/loom/config/defaults.json` to set `autonomous.roleRunner.enabled: true`.
- Refreshing each workspace's installed Loom (`loom migrate` / self-update) past the v0.15 canary vintage.
- Actually running/observing a seeded Auditor/Curator pass and confirming `loom-daemon status` shows ROLES on + ≥1 `loom:triage` issue per repo.

## Soft dependency

#4501 ("children die permanently on 'Fable 5 limit'... dispatch inherits the interactive fable default") is an **open, unrelated** dispatch-reliability bug. It doesn't block writing the runbook or the label-sync script, but it does gate the *end-to-end* "≥1 issue reaches loom:triage per repo" verification step — a dispatched child that dies immediately on that limit can't be distinguished from "role runner still off." Land/verify this issue's deliverables regardless; note #4501 as a caveat when the operator runs the manual verification steps.

## Acceptance Criteria

- [ ] `defaults/scripts/sync-labels.sh` accepts a `--repo OWNER/NAME` flag that syncs `.github/labels.yml` (from the invoking repo, i.e. this one) onto the named GitHub repo via `gh label` calls only — no `cd`, no local checkout of the target repo required. Existing positional-`WORKTREE_PATH` behavior is unchanged (flag is additive).
- [ ] `.loom/scripts/sync-labels.sh` reflects the change automatically (it is a symlink to `defaults/scripts/sync-labels.sh`) — no separate edit needed; confirm with `ls -la .loom/scripts/sync-labels.sh`.
- [ ] New/updated test(s) covering the `--repo` flag's argument parsing and REPO-resolution short-circuit.
- [ ] `defaults/docs/daemon-reference.md` gains a "Bringing a new repo online" runbook subsection (placement above) covering the per-repo-vs-machine-level config choice, the exact JSON block, the new `--repo` label-sync usage, the stale-install refresh pointer (#3936–#3939), and the `loom-daemon status` / `loom:triage` verification steps.
- [ ] The runbook explicitly lists, as manual/operator steps (not delivered by this PR), the actual per-host config edits, install refresh, and the seeded-backlog verification for the 8 named targets (klayout-tools, gf180-bandgap/ldo/temp-por/pll/sar-adc/trng, sky130-bandgap) — with #4501 noted as a caveat on the verification step.
- [ ] `cargo test`/existing repo test suites unaffected (docs + bash-script-only change; no Rust source touched).

## Affected Files
- `defaults/scripts/sync-labels.sh` (also reachable via the symlink `.loom/scripts/sync-labels.sh`) — add `--repo OWNER/NAME` flag
- `defaults/docs/daemon-reference.md` (also reachable via the symlink `.loom/docs/daemon-reference.md`) — new runbook subsection after line ~2762
- New or existing test file under `tests/install/` (or a sibling `tests/scripts/` if one exists) covering the new flag

## Test Plan
1. `./defaults/scripts/sync-labels.sh --repo octocat/hello-world --dry-run` (or equivalent no-op mode if added) parses correctly and resolves `REPO=octocat/hello-world` without touching `git remote`/cwd detection.
2. Run the existing script test suite (e.g. `bash tests/install/test-uninstall-labels-block.sh` and any new test file) and confirm green.
3. Manually render the new runbook subsection and confirm the JSON block matches the schema already documented at `defaults/docs/daemon-reference.md:1988-1994` (`autonomous.roleRunner.enabled/roles/intervalSecs/onIdle`).
4. `cargo test --lib` in `loom-daemon/` (sanity check — no Rust files touched by this issue, should be a no-op pass).

## Complexity
routine

rjwalters:
Closes #4498

## Problem

`loom-daemon workspace add` gives a repo **dispatch** but not a **queue**. The role runner's `enabled` gate is resolved per root from that root's own effective config (#4377), so a newly-registered root shows `ROLES off` in `loom-daemon status` and sits inert indefinitely while looking registered-and-healthy — no Curator/Judge/Champion/Auditor/Guide runs there, so nothing ever reaches `loom:issue`.

Two things were missing to turn a registered root into a self-feeding one:

1. A way to create that repo's `loom:*` labels (the coordination protocol is label-based — a Curator that tries to apply `loom:curated` to a repo without the label just fails). `sync-labels.sh` already did exactly this work over the `gh label` API, but resolved its target from the *current directory's* git remote, so it required a local checkout of every target.
2. A written procedure. The per-root gate, the four config tiers, the label prerequisite, the stale-install friction, and the verification order were each documented separately or not at all.

### 1. `sync-labels.sh --repo OWNER/NAME`

Every GitHub label operation in the script was *already* a `gh label` API call. The only thing forcing a local checkout of the target was target **resolution** — `forge_detect` + `forge_get_repo_nwo` both read the current directory's git remote. `--repo` names the target instead and short-circuits both, so one workspace can bring a whole fleet online:

```bash
for r in OWNER/klayout-tools OWNER/gf180-ldo OWNER/sky130-bandgap; do
  ./.loom/scripts/sync-labels.sh --repo "$r"
done
```

`labels.yml` is still read from `WORKTREE_PATH` — only the *target* moves. `.loom/scripts/sync-labels.sh` needs no separate edit (`.loom/scripts` is a whole-directory symlink to `defaults/scripts`; a test assertion pins that).

### 2. `--dry-run`

Added because `--repo` introduces a new risk that did not exist before: it points the default-label **deletions** at a repo the operator is not standing in, so a typo'd NWO deletes 9 labels somewhere unintended. The dry run is deliberately **forge-free** — zero `gh` calls, reporting intent from `labels.yml` alone — so it validates the parse *and* the resolved target offline, before anything is mutated.

`--repo` is also validated eagerly against an `OWNER/NAME` pattern (rather than discovered after the first deletion attempt) and rejected outright when `LOOM_FORGE_TYPE=gitea`, instead of silently syncing GitHub while the operator expects Gitea.

**Behavior with no new flags is unchanged**: same forge detection, same NWO resolution path, same error text. The one deliberate difference: a second positional argument now exits 2 instead of being silently ignored.

### 3. Runbook

New `#### Bringing a new repo online` subsection in `defaults/docs/daemon-reference.md`, placed directly after the per-root `enabled`-not-inherited callout it follows from:

- **Config surface** — the four tiers (machine-level `~/.local/share/loom/config/defaults.json` / `LOOM_CONFIG_DEFAULTS_FILE` → `.loom/config.json` → `.loom-project/project.json` → `.loom-local/local.json`, cited to `config_resolver.rs:143-153`) with the machine-level-vs-per-repo **blast-radius tradeoff** spelled out, plus the exact `roleRunner` JSON block.
- **Label sync** via the new flag.
- **Stale-install refresh** (`loom migrate` / `resync-installed.sh`) with a #3936-#3939 friction table — several of those present *exactly* as "registered, enabled, still nothing happens", so they are worth ruling out before blaming config.
- **Ordered verification**: `ROLES on` → a `loom:triage` issue within one interval → dispatch. Ordered deliberately, because each step's failure mode looks like the next step's.
- **#4501 caveat**: a child that dies instantly on a model limit is indistinguishable at the forge from "the role runner is still off" — both look like a repo with no activity.
- A closing section naming which steps an agent structurally *cannot* perform (host-local writes outside any worktree; observing another host's daemon) versus step 2, which `--repo` makes runnable from anywhere.

## Tests

New `defaults/scripts/tests/test-sync-labels-repo-flag.sh` — 52 assertions, black-box against a stubbed `gh` that logs its argv.

It **proves** the short-circuit rather than asserting it: the `--repo` cases run from a scratch directory that is not a git repo *and* whose stub `gh repo view` fails, so the legacy resolution path could not have produced an NWO. If the short-circuit regressed, those cases fail rather than silently passing. Also pinned:

- **Target exclusivity** — the override is the only `-R` value in the entire gh log, and no call is made without it (a stray fallback to remote-based resolution surfaces as a second distinct target).
- **Zero-call dry run** — the gh log is empty, which is the load-bearing safety property.
- **Unchanged no-flag path**, both directions: with a resolvable NWO it still calls `gh repo view` and targets what that returned; with nothing resolvable it still hard-errors with the original message.
- Every parse rejection (missing value, 5 invalid NWO shapes, unknown option, extra positional, Gitea combination), each confirmed to contact no forge.
- Installed-tree symlink parity.

Folded into `scripts/test-installer.sh` so it runs in **CI + the build gate** rather than dev-only.

### Results

| Check | Result |
|-------|--------|
| `test-sync-labels-repo-flag.sh` | 52/52 passed |
| `scripts/test-installer.sh` (full) | 162 passed, 0 failed |
| `tests/install/test-uninstall-labels-block.sh` | 16/16 passed |
| `bash -n` on all touched scripts | clean |
| `shellcheck --severity=warning` over all of `defaults/scripts/**/*.sh` (the CI leg) | clean |
| `check-phantom-labels` / `check-labels-drift` / `check-label-descriptions` / `check-claude-md-budget` | all OK |

`cargo test` is unaffected — no `.rs`, `Cargo.toml`, or `Cargo.lock` file is touched by this PR (bash + docs only).

## Out of scope (operator steps, called out as such in the runbook)

Flipping `autonomous.roleRunner.enabled` for the target roots, refreshing their installs, and observing a seeded backlog all require writes to host-local paths outside any Loom checkout (a home-directory file, or another repo's own checkout) or observation of a daemon on another host. Neither is reachable from a Builder worktree, so the runbook documents them as per-host manual steps rather than pretending to automate them.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2796,6 +2796,139 @@ logged only at `debug!` (invisible at the default `info` level), and the
   "N of M registered workspaces are inert" state is diagnosable without
   reading every root's config file by hand.
 
+#### Bringing a new repo online (runbook, #4498)
+
+`loom-daemon workspace add` registers a repo for *dispatch*. It does **not** give
+that repo a **queue** — no role runner means no Curator, Judge, Champion, Auditor,
+or Guide runs there, so nothing ever reaches `loom:issue` and dispatch has nothing
+to pick up. A newly-registered root therefore shows `ROLES off` in
+`loom-daemon status` and sits inert indefinitely, looking registered-and-healthy.
+The steps below are what turns a registered root into a *self-feeding* one.
+
+**1. Choose a config surface.** `autonomous.roleRunner.enabled` is read from the
+root's own **effective** config, which `config_resolver::resolve_effective_config`
+(`loom-daemon/src/config_resolver.rs:143-153`) deep-merges from four tiers,
+lowest → highest priority:
+
+| # | Tier | Path | Tracked? |
+|---|------|------|----------|
+| 1 | machine-level defaults | `~/.local/share/loom/config/defaults.json` (override: `LOOM_CONFIG_DEFAULTS_FILE`; set it empty to disable the tier) | no — outside every checkout |
+| 2 | legacy per-repo | `<root>/.loom/config.json` | yes |
+| 3 | project | `<root>/.loom-project/project.json` | yes (Epic #3835 Phase 6) |
+| 4 | host-local | `<root>/.loom-local/local.json` | no (git-ignored) |
+
+Two valid ways to flip a root on, with very different blast radii:
+
+- **Machine-level (tier 1)** — one edit per *host* turns on **every** registered
+  root that does not set `roleRunner.enabled` itself. This is the right lever for
+  a fleet of repos you own that should all behave the same way. Because it is the
+  lowest-priority tier, any repo that wants to opt back out can simply set
+  `enabled: false` in its own config and win.
+- **Per-repo (tier 2/3/4)** — one edit per *repo*, in that repo's own checkout.
+  Use this when a root needs a different `roles`/`onIdle` set than the host
+  default, or when the repo's maintainers should see the setting in review.
+
+Either way the JSON block is the same shape as the example above:
+
+```json
+{
+  "autonomous": {
+    "roleRunner": {
+      "enabled": true,
+      "roles": ["champion", "curator", "judge", "auditor", "guide"],
+      "intervalSecs": 300,
+      "onIdle": ["champion"]
+    }
+  }
+}
+```
+
+Remember `onIdle` is **inert unless the work finder is enabled** for that root,
+and that `LOOM_ROLE_RUNNER` (env) outranks all four config tiers.
+
+**2. Sync the repo's labels.** The whole coordination protocol is label-based, so
+a repo with no `loom:*` labels cannot hold a queue — a Curator that tries to apply
+`loom:curated` just fails. Since every GitHub label operation is a `gh label` API
+call, this needs **no checkout of the target repo**: run
+`sync-labels.sh --repo OWNER/NAME` from a workspace that already has an
+authoritative `.github/labels.yml` (labels.yml is read locally; only the target
+moves). Preview first — `--repo` points the default-label *deletions* at a repo
+you are not standing in, so a typo'd NWO is worth catching:
+
+```bash
+# From a checkout whose .github/labels.yml is the source of truth (e.g. loom):
+for r in OWNER/klayout-tools \
+         OWNER/gf180-bandgap OWNER/gf180-ldo OWNER/gf180-temp-por \
+         OWNER/gf180-pll OWNER/gf180-sar-adc OWNER/gf180-trng \
+         OWNER/sky130-bandgap; do
+  ./.loom/scripts/sync-labels.sh --repo "$r" --dry-run   # preview
+done
+# ...then re-run without --dry-run to apply.
+```
+
+`--dry-run` is deliberately forge-free (it reports intent from labels.yml alone
+and makes zero `gh` calls), so it also validates the parse and the NWO offline.
+`--repo` is GitHub-only — for a Gitea root, run `sync-labels.sh` from a checkout
+of that repo so the Gitea helpers can resolve a base URL and token.
+
+**3. Refresh a stale install.** A root installed before the machine-level daemon
+model carries a committed file-copy of `.loom/` that **shadows** the machine
+checkout and drifts stale — including the role prompts the role runner invokes.
+Run `loom migrate --dry-run` then `loom migrate` in that root (see
+`defaults/docs/machine-dispatcher.md` → `loom migrate`); for a still-file-copy
+install that is not being migrated yet, `./.loom/scripts/resync-installed.sh`
+refreshes the copies in place.
+
+A repo on the **v0.15 multi-repo canary vintage** predates a cluster of
+multi-repo fixes. Rule these out before blaming a new symptom on config — several
+of them present exactly as "registered, enabled, and still nothing happens":
+
+| Issue | Symptom if the install predates the fix |
+|-------|------------------------------------------|
+| #3936 | `loom-daemon status`'s token-capacity summary contradicts its own per-token table |
+| #3937 | sweep-registry label flips run `gh issue edit` unscoped — cross-repo claims fail, yet dispatch proceeds *unclaimed* |
+| #3938 | dispatched sweeps die in ~2s: no `.loom/tokens/` pool outside the primary workspace (fixed by the shared machine-level pool — see "Token pool provisioning for managed repos" above; bootstrap it once per host) |
+| #3939 | no backoff/quarantine for insta-crashing dispatches — the same few issues are re-dispatched every tick, starving the rest of the queue |
+
+**4. Verify.** In order, because each step's failure looks like the next one's:
+
+1. `loom-daemon status` — the new root appears in "Managed repos" with
+   `ROLES on`. Still `off` means the config edit did not land in a tier that root
+   actually reads (tier 1 is per-*host*: check `LOOM_CONFIG_DEFAULTS_FILE` is not
+   set to something else, and that the daemon process's `$HOME` is the one you
+   edited). A restart is not required — each role task re-reads the registry and
+   per-root config every tick.
+2. Within one Auditor/Curator interval (default 300s), expect **≥1 issue labeled
+   `loom:triage`** in that repo — the Auditor files what it finds, the Curator
+   enriches it. An empty queue after two intervals with `ROLES on` means the role
+   invocation itself is failing; role failures are never fatal and are logged +
+   skipped, so read the daemon log rather than waiting.
+3. Then, and only then, expect dispatch: `loom:issue` → a sweep → a PR.
+
+> **Caveat (#4501).** A dispatched child that dies immediately — e.g. on a
+> `Fable 5 limit` classified as non-retryable, where dispatch inherited the
+> interactive model default instead of pinning a work model — is
+> indistinguishable at the forge from "the role runner is still off": both look
+> like a repo with no activity. Confirm step 1's `ROLES on` and step 2's
+> `loom:triage` issue *before* concluding anything about dispatch, and treat an
+> insta-dying child as #4501 rather than a config problem.
+
+**Which steps an agent cannot do for you.** Steps 1, 3, and 4 are operator steps,
+run by hand on **each daemon host** — repeat them per host, since tier 1 is
+per-machine and a multi-host fleet has one copy of it per machine:
+
+- **Step 1 (config edit)** and **step 3 (install refresh)** are host-local
+  filesystem writes: a file in the operator's home directory, or another repo's
+  own checkout. Neither is reachable from a Builder worktree — the
+  worktree-isolation guard confines Edit/Write to the current worktree, and the
+  machine-level defaults file lives outside every git checkout entirely.
+- **Step 4 (verification)** needs `loom-daemon status` against the *running*
+  daemon on that host, plus a real Auditor/Curator cycle to elapse. An agent can
+  read the forge state but cannot observe another host's daemon.
+- **Step 2 (label sync)** is the exception: `--repo` makes it pure forge API, so
+  it runs from any checkout with `gh` auth for the targets — one loop covers the
+  whole target set at once.
+
 ### Durable operator watches (#3971)
 
 **Problem it fixes.** An operator armed a 4-hour background poll watching two
--- a/defaults/scripts/sync-labels.sh
+++ b/defaults/scripts/sync-labels.sh
@@ -11,9 +11,24 @@
 #
 # Usage:
 #   .loom/scripts/sync-labels.sh [WORKTREE_PATH]
+#   .loom/scripts/sync-labels.sh --repo OWNER/NAME [--dry-run] [WORKTREE_PATH]
 #
 #   WORKTREE_PATH  Directory containing .github/labels.yml and a git remote.
 #                  Defaults to the current directory.
+#
+# --repo OWNER/NAME (#4498) retargets the sync at an arbitrary GitHub repo
+# while still reading labels.yml from WORKTREE_PATH. Because every GitHub label
+# operation here is a `gh label` API call, no local checkout of the target repo
+# is needed — so one workspace can bring a whole fleet of repos online:
+#
+#   for r in owner/klayout-tools owner/gf180-ldo owner/sky130-bandgap; do
+#     .loom/scripts/sync-labels.sh --repo "$r"
+#   done
+#
+# --repo bypasses forge detection entirely (the target is named, not inferred
+# from a git remote) and is GitHub-only. Pair it with --dry-run first: --repo
+# makes the default-label deletions land on a repo you are not standing in, so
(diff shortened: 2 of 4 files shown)


rjwalters:
## Approved

Reviewed against #4498's curated ACs. Everything claimed in the PR body reproduced locally; no blocking findings.

### Verification performed

| Claim | Result |
|-------|--------|
| `test-sync-labels-repo-flag.sh` = 52 assertions | **52/52 passed** locally |
| `scripts/test-installer.sh` = 162 | **162 passed, 0 failed** locally; the new suite is the last entry in the run, so aggregator wiring is real (CI leg: `ci.yml:111`) |
| `bash -n` on all three touched scripts | clean |
| shellcheck `--severity=warning` | clean on `sync-labels.sh` + the new test. The one SC2227 in `test-installer.sh` is at line 244 and **pre-exists on main** (2 occurrences on `origin/main`), not introduced here |
| CI on head `08d60d13` | all pass (Shellcheck, Installer Integration Tests, AGENTS.md Sync, CLAUDE.md budget, Codex smoke); Rust legs correctly `skipping` — no `.rs`/`Cargo.*` touched |
| `mergeStateStatus` | `CLEAN` / `MERGEABLE` |

**The short-circuit is genuinely proven, not asserted.** The `--repo` cases run from a non-git scratch dir whose stub `gh repo view` exits 1, so `forge_get_repo_nwo` could not have produced an NWO — a regressed short-circuit fails rather than silently passing. The two strongest assertions are the target-exclusivity pair (`-R <override>` is the only distinct `-R` value in the whole gh argv log, **and** no logged call lacks it) and the zero-length gh log on `--dry-run`. Both are the right invariants for this feature.

### Argv parser: single-`case` → `while`-loop rewrite

I walked every reachable input shape for regressions. The rewrite is a clean superset:

- `<path>`, no args, `-h`/`--help`, `--bogus` — identical behavior and identical exit codes.
- `sync-labels.sh ""` — the old parser had an explicit `"")` arm mapping empty → `.`; the new one falls to `*)` and sets `WORKTREE_PATH=""`. Verified equivalent: `cd ""` is a successful no-op in bash (`rc=0`, cwd unchanged), and a live run with `""` still resolves `rjwalters/loom`. No regression.
- `<path> -h` previously ran a **full real sync** (the `-h` was in `$2`, unexamined); it now prints help and exits 0. Strictly safer.
- Second positional now exits 2. **Verified no caller is affected**: grepped every `.sh/.md/.rs/.py/.yml/.json` reference — `install.sh:236` is prose, `defaults/docs/*` pass 0–1 args, and `scripts/install-loom.sh:1766` invokes the *source-only* `scripts/install/sync-labels.sh` (untouched by this PR) with one positional. Zero callers pass two.

### Runbook accuracy

Every cited anchor checked against source, all correct:

- `config_resolver.rs:143-153` — `resolve_effective_config` starts at line 143; the four `deep_merge` calls land in 146-152, in exactly the documented tier order.
- Tier-1 path and env override — `DEFAULT_PRIVATE_DEFAULTS_REL = ".local/share/loom/config/defaults.json"`, `PRIVATE_DEFAULTS_ENV = "LOOM_CONFIG_DEFAULTS_FILE"`, and the "set it empty to disable the tier" claim matches `private_defaults_path()`'s `p.is_empty() -> None`.
- "`LOOM_ROLE_RUNNER` outranks all four config tiers" — matches `resolve_enabled()` (env checked first, `config.enabled.unwrap_or(false)` only as fallback).
- "A restart is not required — each role task re-reads the registry and per-root config every tick" — matches the multi-workspace tick: `WorkspaceRegistry::load_default()` then `read_role_runner_config(&root)` **inside** the loop.
- "role failures are never fatal and are logged + skipped" — matches the module docs and the warn-once disabled-root path (#4377).
- `machine-dispatcher.md` → `loom migrate` exists (line 409).
- #4501's caveat is an accurate paraphrase of the open issue's title/scope.

The machine-level-vs-per-repo blast-radius framing is correct and non-obvious in a useful way: because tier 1 is *lowest* priority, the "any repo can opt back out by setting `enabled: false` and win" claim follows directly from the merge order. The per-host repetition warning is right too. **No operator-personal hostnames** — targets are `OWNER/` placeholders, the repo names come from #4498's own title, and the only `.local` string is the `~/.local/share` XDG path. Placement directly after the per-root `enabled`-not-inherited callout is the right home, and `.loom/docs/daemon-reference.md` is a per-file symlink to `defaults/docs/`, so the runbook is live in the installed tree with no second edit.

Not updating CHANGELOG matches convention here (0 of the last 20 commits touch it; it is written in bulk at release).

### Non-blocking follow-ups (do not respin for these)

1. **`--` is not a real end-of-options terminator.** It is shifted and ignored, so `sync-labels.sh -- -weird-dir` still errors `Unknown option: -weird-dir`. Either make it set a `no more options` flag or drop the arm (bare `--` previously exited 2, so accepting it silently is a small unadvertised change).
2. **The Gitea guard only inspects `LOOM_FORGE_TYPE`.** `forge_detect` also resolves `forge.type: gitea` from the config tiers (`forge-helpers.sh:111-121`), and that path is not checked — so a Gitea-configured workspace running `--repo` proceeds as GitHub without the loud failure the code comment promises. The reading that makes this correct (local config describes the *local* repo, not the named target) is defensible, but then the env check is a courtesy rather than an invariant; worth saying so in the comment either way.
3. **The NWO regex accepts dot-only and dash-leading components** — `../..` and `-foo/bar` both validate. Harmless in practice (both arrive as a `gh` flag *value* and 404), but a first-character constraint would tighten it at no cost.
4. **Deletion-at-a-distance is mitigated, not closed.** Eager NWO validation, the forge-free dry run, and echoing the resolved target are all the right moves, and I agree `--dry-run` is the correct shape (offline preview of parse + target). But nothing confirms the target *is the intended repo*: a typo that happens to name a real repo the operator can administer still loses 9 default labels, and `--dry-run` cannot catch it because it never contacts the forge. A cheap `gh repo view -R "$REPO" --json nameWithOwner` preflight before the deletion loop (only when `--repo` is set) would turn "typo'd but plausible NWO" from silent success into a visible confirmation. Worth a small follow-up.

Good discipline throughout: the flag is additive, the risk it introduces was identified and mitigated in the same PR rather than deferred, the tests target invariants instead of output strings, and the runbook is honest about which steps an agent structurally cannot perform.

Issue #4466, opened 2026-07-29
narration: child-published sweep.issue.N.* events arrive as Generic and are never narrated — docs table promises SweepPhase lines


rjwalters:
## Context

Found while designing #4426 (verified in code, current main):

- Sweep children publish phase/blocker events via `mcp__loom__publish_event` → IPC `Request::PublishEvent` → **`Event::Generic { topic, payload }`** (`ipc.rs`, the PublishEvent handler). There is no Generic→typed conversion anywhere on the bus path.
- The narration sink maps only **typed** variants: `event_to_envelope` (`safehouse.rs`) returns `None` for `Event::Generic`, and `Event::set_repo_if_absent` (`types.rs`) doesn't stamp `Generic` either.

So a `sweep.issue.{N}.phase` published by a sweep child never produces a room line, while:

- `.loom/docs/safehouse.md`'s "What gets narrated" table lists `SweepPhase → task → "<repo>#N · <phase> (+ PR #M open)"`, and
- `event_bus.rs`'s topic table documents `sweep.issue.{N}.phase | Sweep child via PublishEvent`.

The lines that DO reach the room (dispatch, exited, crashed, resume) are all daemon-emitted typed events from `SweepRegistry`. Live room history confirms: dispatch/ack/handoff lines present, zero phase lines.

Consequence: the room shows a sweep start and its end, but none of the phase progression the docs promise — and any future consumer of child-published topics (e.g. #4426-style feeds) can silently fall into the same hole, since `publish` accepting a topic string looks like success.

## Suggested acceptance criteria

- [ ] Decide the contract: either (a) the PublishEvent handler upgrades known `sweep.issue.{N}.*` topics to their typed variants (payload-schema validated, unknown topics stay Generic), or (b) the narration sink learns to narrate Generic events on documented topics, or (c) the docs stop promising phase narration.
- [ ] Whichever way: `.loom/docs/safehouse.md`'s narration table and `event_bus.rs`'s topic table match observable behavior.
- [ ] A test: a child-published `sweep.issue.N.phase` (as `PublishEvent` delivers it) either narrates or is explicitly documented as not narrated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

### Verification (against origin/main)

All claims in the original report verified:

[...]

PR #4514, opened 2026-07-30
fix(daemon): ownership-checked lock release prevents same-host double dispatch (#4463)

### LINKED ISSUE #4463: daemon: duplicate same-host dispatch — two concurrent sweeps for one issue, 43s apart, sharing one worktree
## Context

Observed live on this host (2026-07-29), issue #4426:

- 13:47:45 PDT — sweep A dispatched: `claude-wrapper.sh -p /loom:sweep 4426 --claim-owned 4426 --model sonnet` (pid chain 48474/49283/49309).
- 13:48:28 PDT — sweep B dispatched, **43 seconds later, same host, same issue**: `claude-wrapper.sh -p /loom:sweep 4426 --claim-owned 4426` (pid 60930), logged as `sweep_id=sweep-issue-4426-1785358105` in `.loom/logs/sweep-issue-4426.log`.

Both ran concurrently and shared `.loom/worktrees/issue-4426`: sweep B's Builder reset the worktree while sweep A was active, clobbering uncommitted work (each sweep's Builder resets the worktree at start). Sweep B later stranded itself (separate issue); sweep A survived and produced PR #4459 — but only because B died first.

Both dispatches carried `--claim-owned 4426`, i.e. the daemon believed it had claimed the issue both times. The cross-host peer-claim advert/backoff (#4028) is not in play here — this is one daemon double-claiming locally. The #4123 open-PR guard also couldn't help (no PR existed yet).

- [ ] Root-cause the double dispatch (work-finder tick overlap? dispatch-ack race? registry claim lock not held across the label flip?) — the sweep-run registry and both log headers above should identify the two dispatch paths.
- [ ] A same-host, same-issue dispatch while a live sweep holds the claim is refused (the in-flight registry already knows the issue → hard guard at dispatch, not just the label flip).
- [ ] Regression test: two dispatch requests for the same issue racing within one tick produce one sweep.

---

### Root-cause forensics (new evidence — reframes the issue)

Log evidence gathered during curation **contradicts the "one daemon double-claiming locally" framing above** and identifies the mechanism:

1. **The production daemon never successfully dispatched #4426.** `~/.loom/daemon.log` contains exactly two lines for issue 4426: a work-finder dispatch attempt at 12:36:16 PDT that was **correctly refused** by the lock-collision guard (`lock collision: issue #4426 is already claimed`). The atomic mkdir claim lock works as designed within a single dispatch call.
2. **There were five dispatches that day, not two.** `.loom/logs/sweep-issue-4426.log` dispatch headers: 19:35:55, 19:39:28 (+3.5 min — a second near-duplicate pair), 20:28:05, 20:47:42 (= "sweep A"), 20:48:28 UTC (= "sweep B"). All five originated from daemon instance(s) *other than* the production launchd daemon — consistent with the worktree/debug-daemon double-dispatch pattern observed the same day on #4431's dispatch (multiple `loom-daemon` instances live during daemon development, all sharing the workspace's `.loom/locks/`).
3. **Dispatch-path fingerprint.** Sweep A's wrapper args carry `--model sonnet` (work-finder-style dispatch — model always pinned per #3944). Sweep B has **no `--model` at all** (`spawn-claude: model=default`) — matching the reaper's re-dispatch paths, which pass `model=None`: `dispatch_resume_after_crash()` (`loom-daemon/src/sweep_registry.rs` ~line 2071) and the checkpoint-resume branch of `reap_once()`.

### The concrete defect: `release_lock` is issue-keyed, not ownership-checked

`acquire_lock()` (~line 2381) writes `owner.json` containing the owning `sweep_id` — but `release_lock()` (~line 2446) removes `.loom/locks/issue-<N>` unconditionally, and every reaper call site invokes it by **issue number alone** (`reap_once` dead-sweep branch ~line 3937, `finish_cancel` ~line 3818, plus ~line 4773). Sequence that reproduces the incident:

1. An older sweep for issue N dies (e.g. the 20:28:05 dispatch).
2. A fresh sweep A for issue N is dispatched and acquires the lock (20:47:42).
3. A reaper tick (in the same daemon **or any other instance sharing the workspace**) confirms the *old* pid dead → `release_lock(N)` deletes **sweep A's** lock — no `sweep_id` comparison.
4. The same reaper branch immediately re-dispatches (checkpoint resume / `dispatch_resume_after_crash`) → `acquire_lock` succeeds on the just-deleted lock → sweep B spawns (20:48:28) while A is alive. Both share `.loom/worktrees/issue-N`.

Note for the fix: the suggested "in-flight registry hard guard" in the original AC is **per-process memory** — it cannot protect across daemon instances, and the incident evidence shows multiple instances were dispatching. The on-disk lock is the only cross-instance mutual-exclusion signal; the fix belongs there.

### Implementation guidance

- **Ownership-checked release (core fix):** `release_lock` (or a new `release_lock_owned(issue, sweep_id)`) must read `owner.json` and remove the lock dir **only if its `sweep_id` matches the sweep being reaped/cancelled**. A mismatch means a newer sweep owns the claim: log at `info` and skip both the release and any re-dispatch for that issue in the same branch. All data needed is already on disk (`LockOwner { issue, owner_pid, acquired_at, sweep_id }`).
- **Reaper re-dispatch must re-check the lock:** the checkpoint-resume and `dispatch_resume_after_crash` branches of `reap_once` should treat "lock now owned by a different live sweep_id" as *not crashed — superseded*, and skip resume entirely.
- **Keep fail-open semantics:** an unreadable/corrupt `owner.json` should fall back to today's unconditional release (never wedge an issue on a garbage lock file).
- Out of scope here (file separately if desired): an operational guard against multiple daemon instances managing one workspace (the #3806 singleton guard is per-socket, so a debug daemon with its own socket coexists silently).

### Affected Files

- `loom-daemon/src/sweep_registry.rs` — `release_lock()` (~2446): add sweep_id ownership check; `acquire_lock()` (~2381): no change expected (already records `sweep_id`); `reap_once()` dead-sweep branch (~3937) and resume branches (~4060–4110): pass the reaped sweep's `sweep_id` through and skip resume on ownership mismatch; `finish_cancel` (~3818) and the ~4773 call site: same ownership-checked release.
- `loom-daemon/src/sweep_registry.rs` (tests) — extend the existing lock/reconstruct test cluster (~9235–9483, e.g. `reconstruct_admits_live_lock_owners`, `reconstruct_drops_stale_locks`) with the regression tests below.

(Line anchors verified against `main` @ 3490b5c7 on 2026-07-29; expect drift.)

### Test Plan

- [ ] Regression (core): reap a dead sweep for issue N while a **newer** sweep_id holds N's lock → the lock dir survives, no resume dispatch fires, exactly one sweep remains live.
- [ ] Regression (original AC): two dispatch requests for the same issue racing within one tick produce one sweep (second fails on lock collision — this already passes today; keep as a guard).
- [ ] `finish_cancel` on a stale entry whose issue lock is owned by a newer sweep_id → lock survives.
- [ ] Corrupt/missing `owner.json` → release proceeds (fail-open), matching current behavior.
- [ ] Manual verification: `cd loom-daemon && cargo test sweep_registry::` passes; optionally reproduce with two `loom-daemon` instances sharing one workspace and confirm the second reaper no longer deletes the first's live lock.

### Acceptance criteria (refined from the original)

- [ ] `release_lock` never removes a lock whose `owner.json` `sweep_id` differs from the releasing sweep's (fail-open on unreadable owner.json).
- [ ] Reaper resume/re-dispatch branches skip re-dispatch when the issue's lock is owned by a different live sweep.
- [ ] Regression tests above land in `sweep_registry.rs`.
- [ ] (Original AC, retained) A same-host, same-issue dispatch while a live sweep holds the claim is refused — satisfied at the lock layer, which is cross-instance, rather than the per-process in-flight registry.

### Related work

- #4444 — `dispatch_inner` never consults SKIP_LABELS on resume/watchdog paths (adjacent defect in the same re-dispatch branches; a fix here should not conflate the two).
- #4146 — cross-host duplicate-dispatch baseline (this issue is the same-host case; #4028's peer-claim advert explicitly doesn't apply).
- #4123 / #4256 — open-PR guard + resume bypass (couldn't help here: no PR existed yet).
- #3806 — daemon singleton guard is per-socket, which is why concurrent debug/production instances coexist.

rjwalters:
## Summary

Fixes the same-host duplicate-dispatch race (Issue #4463): two concurrent sweeps for one issue, 43s apart, sharing one worktree and clobbering each other's uncommitted work.

Root cause (per Curator forensics): `release_lock()` removed `.loom/locks/issue-<N>` keyed by **issue number only**, with no `sweep_id` ownership check against `owner.json`. A reaper tick (in this daemon or any other instance sharing the workspace) that confirmed an *old* dead sweep's pid dead deleted the lock even when a *newer live* sweep had already re-acquired the claim — and the checkpoint-resume / `dispatch_resume_after_crash` branch then re-dispatched a second sweep into the same worktree. The on-disk lock is the only cross-instance mutual-exclusion signal, so the fix belongs there (a per-process in-flight registry cannot protect across daemon instances).

## Changes (`loom-daemon/src/sweep_registry.rs`)

- New `release_lock_owned(issue, sweep_id) -> LockReleaseOutcome` and a read-only `lock_owned_by_other()` probe. Both read `owner.json` and refuse to remove a lock owned by a **different** `sweep_id` (`Superseded`); **fail-open** on missing/unreadable/corrupt `owner.json` so a garbage lock file never wedges an issue.
- `reap_once` dead-sweep branch, `finish_cancel`, and the midbuild-watchdog re-dispatch now release via the ownership-checked path. On `Superseded` they skip the label restore **and** the resume/re-dispatch — the newer sweep owns the claim and runs its own lifecycle. The midbuild path probes read-only **before** cleaning the worktree, so a live sweep's tree is never wiped.
- `release_lock()` retained as the explicit owner-blind removal.

## Acceptance criteria

- [x] `release_lock` never removes a lock whose `owner.json` `sweep_id` differs from the releasing sweep's (fail-open on unreadable owner.json).
- [x] Reaper resume/re-dispatch branches skip re-dispatch when the issue's lock is owned by a different live sweep.
- [x] Regression tests land in `sweep_registry.rs`.
- [x] A same-host, same-issue dispatch while a live sweep holds the claim is refused — at the cross-instance lock layer.

## Tests

New regression tests (all green): `release_lock_owned` matrix (own / other-sweep / corrupt / missing / no-lock), `reap_dead_sweep_preserves_newer_sweep_lock_and_skips_resume` (core), `cancel_preserves_lock_and_skips_restore_when_superseded`, and the retained `second_same_issue_dispatch_collides_on_lock` guard.

`cargo test -p loom-daemon --lib`: 222 passed, 0 failed. `cargo fmt --check` clean. `cargo clippy --lib` clean.

Avoided the `ipc.rs`/`types.rs`/`safehouse.rs` narration surfaces (concurrent #4466).

Closes #4463

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -1455,6 +1455,19 @@ impl SweepRegistryConfig {
     }
 }
 
+/// Outcome of an ownership-checked lock release (Issue #4463).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LockReleaseOutcome {
+    /// The lock dir was removed — either its `owner.json` `sweep_id` matched
+    /// the releasing sweep, or the owner was unreadable (fail-open), or no lock
+    /// existed (idempotent no-op).
+    Released,
+    /// A *different* sweep owns the lock (a newer sweep re-acquired the claim
+    /// after the releasing one died). The lock was left intact; the caller MUST
+    /// NOT restore the label or re-dispatch.
+    Superseded,
+}
+
 /// On-disk owner metadata written inside the lock dir. Schema mirrors
 /// `defaults/scripts/spawn-loop.sh:299-305`.
 #[derive(Debug, Serialize, Deserialize)]
@@ -2683,6 +2696,12 @@ impl SweepRegistry {
     }
 
     /// Release the lock dir for an issue (idempotent).
+    ///
+    /// UNCONDITIONAL: removes `.loom/locks/issue-<N>` regardless of which sweep
+    /// owns it. Prefer [`release_lock_owned`](Self::release_lock_owned) from any
+    /// reaper / cancel / re-dispatch path so a newer sweep's live claim is not
+    /// clobbered (Issue #4463) — this remains for callers that intentionally
+    /// want an owner-blind removal.
     pub fn release_lock(&self, issue: u32) -> Result<()> {
         let lock = self.config.locks_dir().join(format!("issue-{issue}"));
         if lock.exists() {
@@ -2690,6 +2709,71 @@ impl SweepRegistry {
                 .with_context(|| format!("failed to remove lock dir {}", lock.display()))?;
         }
         Ok(())
+    }
+
+    /// Ownership-checked lock release (Issue #4463).
+    ///
+    /// Removes `.loom/locks/issue-<N>` **only when** its `owner.json`
+    /// `sweep_id` matches `sweep_id` — the sweep being reaped / cancelled /
+    /// re-dispatched. When a *different* sweep owns the lock (a newer sweep
+    /// re-acquired the claim after this one died — the double-dispatch incident
+    /// this guards against), the lock is left intact and
+    /// [`LockReleaseOutcome::Superseded`] is returned so the caller skips any
+    /// label restore and any re-dispatch: the newer sweep is the live owner and
+    /// runs its own lifecycle.
+    ///
+    /// FAIL-OPEN: a missing / unreadable / corrupt / unparseable `owner.json`
+    /// falls back to the legacy unconditional removal — the release only refuses
+    /// on a *positively-read, conflicting* owner, so a garbage lock file can
+    /// never wedge an issue permanently. A non-existent lock dir is a no-op
+    /// ([`LockReleaseOutcome::Released`], idempotent).
+    #[must_use]
+    pub fn release_lock_owned(&self, issue: u32, sweep_id: &str) -> LockReleaseOutcome {
+        let lock = self.config.locks_dir().join(format!("issue-{issue}"));
+        if !lock.exists() {
+            return LockReleaseOutcome::Released;
+        }
+        if self.lock_owned_by_other(issue, sweep_id) {
+            return LockReleaseOutcome::Superseded;
+        }
+        if let Err(e) = std::fs::remove_dir_all(&lock) {
+            log::warn!("release_lock: failed to remove lock dir {}: {e}", lock.display());
+        }
+        LockReleaseOutcome::Released
+    }
+
+    /// Read-only ownership probe (Issue #4463): `true` iff the issue lock's
+    /// `owner.json` records a `sweep_id` *different* from `sweep_id` — i.e. a
+    /// newer sweep re-acquired the claim after the querying sweep died. Unlike
+    /// [`release_lock_owned`](Self::release_lock_owned) this never mutates the
+    /// filesystem, so a caller can gate destructive work (worktree cleanup,
+    /// re-dispatch) on it without prematurely freeing the lock.
+    ///
+    /// FAIL-OPEN: a missing lock dir, or an unreadable / unparseable
+    /// `owner.json`, resolves to `false` (not-conflicting) so a garbage owner
+    /// file can never wedge an issue.
+    fn lock_owned_by_other(&self, issue: u32, sweep_id: &str) -> bool {
+        let owner_path = self
+            .config
+            .locks_dir()
+            .join(format!("issue-{issue}"))
+            .join("owner.json");
+        // Only report a conflict on a POSITIVELY-read differing owner.
+        match std::fs::read_to_string(&owner_path) {
+            Ok(contents) => match serde_json::from_str::<LockOwner>(&contents) {
+                Ok(owner) if owner.sweep_id != sweep_id => {
+                    log::info!(
+                        "release_lock: issue #{issue} lock is owned by sweep {} \
+                         (sweep {sweep_id} was superseded) — leaving the lock intact and \
+                         skipping any re-dispatch (#4463)",
+                        owner.sweep_id
+                    );
+                    true
+                }
+                _ => false,
+            },
+            Err(_) => false,
+        }
     }
 
     /// Best-effort removal of this registry's sweep-journal entry for
@@ -4163,7 +4247,12 @@ impl SweepRegistry {
             };
         }
         if let SweepKind::Issue(issue) = kind {
-            let _ = self.release_lock(*issue);
+            // Ownership-checked release (#4463): if a newer sweep re-acquired
+            // this issue's lock after the sweep being cancelled died, leave its
+            // live lock intact and skip the label restore below — the newer
+            // sweep owns the claim and runs its own lifecycle.
+            let superseded =
+                self.release_lock_owned(*issue, sweep_id) == LockReleaseOutcome::Superseded;
             // Best-effort tidy-up of the machine-level liveness journal
             // (#3953) — this cancelled sweep no longer exists.
             self.journal_remove_best_effort(*issue);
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved** — ownership-checked lock release cleanly closes the reaper-deletes-live-lock race (#4463).

## Concurrency review (scrutinized hard, per the incident forensics)

**Reaper-deletes-live-lock race — CLOSED.** `release_lock_owned(issue, sweep_id)` keys the release on the *reaped* sweep's `sweep_id` versus whoever currently owns `owner.json`. In the incident sequence (old sweep dies → sweep A re-acquires → a reaper confirming the *old* pid dead calls release), the current owner is A, `A != old_sweep_id` → `Superseded` → the lock is preserved and the resume/re-dispatch is skipped (`!superseded` gates every restore/resume site at ~4404, ~4540, ~4645). Because `acquire_lock` uses POSIX-atomic `mkdir`, no other sweep can create the issue's lock dir while our owned dir exists — so the probe→`remove_dir_all` window inside `release_lock_owned` cannot clobber a *newer* live lock. The race is genuinely closed for the reaper path, not merely narrowed.

**Fail-open on missing/corrupt owner.json — correct posture.** Fail-closed would permanently wedge an issue on a corrupt or legacy (pre-owner.json spawn-loop) lock: no sweep_id would ever match, so nothing could release it. Fail-open only removes on a *positively-read, conflicting* owner, so a garbage file can never wedge. The only residual exposure is the microsecond window in `acquire_lock` between `create_dir` (line 2641) and the `owner.json` write (line 2656) — strictly narrower than the pre-PR unconditional `release_lock`, and bounded. Defensible.

**Skipping label-restore on Superseded — no stranded state.** The issue correctly stays `loom:building` under the newer sweep's ownership; that sweep runs its own lifecycle and flips labels at its end. Restoring here would yank `loom:building → loom:issue` out from under a live sweep and risk a *third* dispatch — so the skip is the right call.

## One non-blocking observation (follow-up, not a blocker)

The midbuild-watchdog read-only probe (`lock_owned_by_other` at ~5333, before `clean_worktree` at ~5365) **narrows but does not fully close** a TOCTOU: if the lock is already free at probe time (the reaper released it) and a cross-instance sweep acquires + starts building in the interval between the probe and `clean_worktree`, the cleanup could still clobber its tree. This is strictly better than the pre-PR behavior (which cleaned unconditionally, no probe), it closes the *observed* 43s-apart incident (the newer sweep held the lock well before the watchdog fired), and it is bounded to a single recovery by `midbuild_retried`. The fully-robust form would acquire the lock atomically *before* cleaning the worktree — worth a follow-up issue, but out of scope for this fix.

## Verification
- `cargo test -p loom-daemon --lib`: **2164 passed, 0 failed**. All 8 new #4463 regression tests green (`release_lock_owned_*` matrix, `reap_dead_sweep_preserves_newer_sweep_lock_and_skips_resume`, `cancel_preserves_lock_and_skips_restore_when_superseded`, `second_same_issue_dispatch_collides_on_lock`).
- CI: Rust Lint & Build **pass**, Rust Unit Tests **pass**, all gate checks **pass**.
- Merge state **CLEAN**; PR body carries `Closes #4463`.
- Single-file change, thorough doc comments, `#[must_use]` on `release_lock_owned`, correct `let _ =` at the midbuild call site (owner already probed).

All acceptance criteria met. Excellent, well-tested fix.

PR #4520, opened 2026-07-30
fix(sweep): reap RUN_ID-keyed main-clean baselines instead of accumulating them

### LINKED ISSUE #4450: sweep: .loom/sweep-checkpoint/ accumulates unbounded main-clean-baseline-*.txt files — nothing prunes per-run transients (200+ since 07-23)
## Context
`.loom/sweep-checkpoint/` currently holds 200+ `main-clean-baseline-<RUN_ID>.txt` files dating back to 2026-07-23 (plus one legacy un-keyed `main-clean-baseline.txt`), and per-issue checkpoint JSONs for issues closed weeks ago (#3784, #3786, …). The RUN_ID-keyed baselines (#3768) are per-sweep-run transients whose lifetime is one sweep invocation, but nothing ever deletes them: sweep completion only deletes the per-issue checkpoint on merge, the stale-checkpoint cleanup only fires when a later sweep touches the same issue number, and baselines are never touched at all.

Harmless functionally, but it grows without bound and makes the checkpoint dir useless to inspect (observed during a sweep on 2026-07-29: the directory listing was ~95% dead baselines).

- [ ] Sweep completion (and/or `sweep-run-registry.sh cleanup`) removes this run's own `main-clean-baseline-<RUN_ID>.txt`
- [ ] A bulk path (`loom-clean` or `loom-daemon clean`) prunes baseline files whose RUN_ID is not in the live `.loom/sweep-run/` registry and is older than a threshold (e.g. 48h), plus per-issue checkpoints whose issue is CLOSED
- [ ] The legacy un-keyed `main-clean-baseline.txt` is removed by the bulk path
- [ ] Pruning never touches the checkpoint/baseline of a live run (registry + PID liveness check)

---

### Problem Statement

`.loom/sweep-checkpoint/` currently holds **259 files** (verified 2026-07-29), the large majority dead `main-clean-baseline-<RUN_ID>.txt` transients plus per-issue checkpoint JSONs for long-closed issues (`issue-3784.json` etc.). Nothing owns their deletion:

- **Baselines**: `sweep.md` creates `main-clean-baseline-${RUN_ID}.txt` once per run (Contamination backstop, #3768) and never deletes it. The sweep-end cleanup hook — `./.loom/scripts/sweep-run-registry.sh cleanup "$RUN_ID"` (defaults/.claude/commands/loom/sweep.md, "best-effort cleanup" section) — removes only the registry entry `.loom/sweep-run/<RUN_ID>.json`, not the baseline keyed by the same RUN_ID.
- **Per-issue checkpoints**: deleted only on `merge-done`, or by stale-checkpoint cleanup *when a later sweep touches the same issue number* — so checkpoints for issues that never get re-swept persist forever.
- **Bulk path**: `loom-daemon clean` (`loom-daemon/src/worktree_ops/clean.rs`) prunes worktrees, branches, tmux sessions, spawn-loop locks, and crash state — but has no sweep-checkpoint category at all.

### Implementation Guidance

Two independent legs; both belong in one PR.

**Leg 1 — sweep-end self-cleanup (shell).** `cmd_cleanup()` in `defaults/scripts/sweep-run-registry.sh` already receives the RUN_ID and is already called at sweep end from `sweep.md`. Add `rm -f "<repo>/.loom/sweep-checkpoint/main-clean-baseline-${RUN_ID}.txt"` there (resolve the checkpoint dir the same way the script resolves `registry_dir`). This is the minimal single-point fix — no `sweep.md` prose change is strictly required, though a one-line mention in the sweep-end cleanup section keeps the doc honest. Cleanup stays best-effort (a SIGKILL'd sweep skips it), which is why Leg 2 exists.

**Leg 2 — bulk prune (Rust).** Add a `clean_sweep_checkpoints()` category to `loom-daemon/src/worktree_ops/clean.rs`, wired into `run_clean()` and `CleanupStats`/`print_summary`, honoring `--dry-run` like the existing categories:

1. **Stale baselines**: for each `main-clean-baseline-*.txt`, parse the RUN_ID from the filename (`main-clean-baseline-(.+)\.txt`). Prune when (a) no `.loom/sweep-run/<RUN_ID>.json` registry entry exists, **or** the entry's registered PID is dead (`kill -0` equivalent — same liveness pattern as `sweep-run-registry.sh peers`), **and** (b) file mtime is older than a threshold (48h suggested; a conservative constant is fine, no config knob needed).
2. **Legacy un-keyed file**: remove `main-clean-baseline.txt` (pre-#3768) unconditionally — no live run can own it.
3. **Closed-issue checkpoints**: for each `issue-<N>.json`, query issue state and delete when CLOSED. Use the REST API (`gh api repos/{owner}/{repo}/issues/N --jq .state`) rather than `gh issue view` — GraphQL quota exhaustion under concurrent agents is a live failure mode in this repo, and `clean.rs` already shells out to `gh` (`gh_pr_list`). On any `gh` failure, skip that file (fail-safe: never delete on an unconfirmed state).

**Liveness guards (AC 4).** Two registries matter, not one:
- Skill-driven sweeps register in `.loom/sweep-run/` (PID liveness) — covered by check 1(a).
- Daemon-owned sweeps track checkpoints via `loom-daemon/src/sweep_registry.rs`. Before deleting an `issue-<N>.json`, also skip any issue that appears in a live/running daemon sweep (consult the sweep registry state the daemon already maintains). The mtime threshold in 1(b) is the belt-and-suspenders backstop for both.

**What NOT to do:**
- No Python changes — `clean.sh` execs `loom-daemon clean` with no fallback (`loom_tools/clean.py` deleted in PR #4301); this is a Rust + shell change only.
- Do not add a new label, config knob, or standalone script; extend the existing clean categories and registry hook.
- Do not touch `check-main-clean.sh` — it takes explicit `--snapshot`/`--baseline` paths and is unaffected by pruning policy.

### Affected Files

- `defaults/scripts/sweep-run-registry.sh` - `cmd_cleanup()`: also remove this run's `main-clean-baseline-<RUN_ID>.txt`
- `loom-daemon/src/worktree_ops/clean.rs` - new `clean_sweep_checkpoints()` category (stale baselines, legacy un-keyed baseline, closed-issue checkpoints) + `CleanupStats`/`print_summary`/`run_clean` wiring + unit tests in the existing `#[cfg(test)]` module
- `defaults/scripts/tests/test-sweep-run-registry.sh` - test that `cleanup <RUN_ID>` removes the matching baseline and leaves a peer run's baseline untouched
- `defaults/.claude/commands/loom/sweep.md` - (optional, one line) note in the sweep-end cleanup section that `cleanup "$RUN_ID"` now also removes the run's baseline

Reference only (no changes): `defaults/scripts/sweep-checkpoint.sh` (checkpoint schema owner), `loom-daemon/src/sweep_registry.rs` (daemon-owned sweep liveness), `defaults/scripts/check-main-clean.sh`.

### Test Plan

- [ ] Automated (shell): in `test-sweep-run-registry.sh`, create two registered runs A and B with baseline files; `cleanup A` removes A's baseline and registry entry, leaves B's both intact
- [ ] Automated (Rust): unit tests in `clean.rs` for — dead-RUN_ID + old mtime baseline pruned; live-registry-entry baseline kept regardless of age; young orphan baseline kept (mtime guard); legacy `main-clean-baseline.txt` removed; `--dry-run` deletes nothing while reporting counts
- [ ] Automated (Rust): closed-issue checkpoint deleted, open-issue checkpoint kept, `gh` failure ⇒ file kept
- [ ] Manual: on this repo (259 accumulated files), run `loom-daemon clean --dry-run` and confirm the report lists the dead baselines + closed-issue checkpoints and spares the live run's files; then run for real and confirm `.loom/sweep-checkpoint/` shrinks to live entries only
- [ ] Edge cases: empty/missing `.loom/sweep-checkpoint/` dir is a no-op; filename that doesn't match the RUN_ID pattern is skipped, not deleted

### Scope Guard

~4 files, well under 400 LOC. If the closed-issue-checkpoint leg balloons (e.g. batching forge queries), it may be split out as a follow-up — the baseline legs (self-cleanup + bulk baseline prune) are the 95% of the observed accumulation and stand alone.

---

### Verified current state (2026-07-29)

- `.loom/sweep-checkpoint/` on the operator host holds **259 files, 227 of them `main-clean-baseline-*.txt`** — confirms the ~95%-dead-baselines observation.
- `defaults/scripts/sweep-run-registry.sh` `cmd_cleanup()` removes only `.loom/sweep-run/<RUN_ID>.json` (plus dead-peer registry entries via `prune_dead`). It never touches the baseline file that `sweep.md` keys by the same `RUN_ID`.
- `defaults/.claude/commands/loom/sweep.md` line ~1362 constructs `MAIN_CLEAN_BASELINE=".loom/sweep-checkpoint/main-clean-baseline-${RUN_ID}.txt"` and line ~663 calls `sweep-run-registry.sh cleanup "$RUN_ID"` at sweep end — so the natural deletion hook already exists and already receives the RUN_ID.
- `loom-daemon clean` (the `loom-clean` successor, `loom-daemon/src/worktree_ops/clean.rs`) has **zero** references to `sweep-checkpoint`, `baseline`, or `sweep-run` — the bulk path is entirely missing, as filed.
- Note: the Python `loom_tools/clean.py` was deleted (PR #4301), so the bulk path is **Rust-only** — no two-language lockstep concern for this issue.

### Implementation Guidance

**Leg 1 — per-run cleanup (cheap, catches everything going forward):**
Extend `cmd_cleanup()` in `defaults/scripts/sweep-run-registry.sh` to also `rm -f` this run's own `main-clean-baseline-<RUN_ID>.txt`. The checkpoint dir path should be derived the same way `sweep-checkpoint.sh` derives it (the two helpers already agree on where `.loom/` lives — see the comment at `sweep-run-registry.sh` line ~62). Opportunistic bonus: when `prune_dead`/`cmd_peers` removes a dead peer's registry entry, also remove that peer RUN_ID's baseline — this makes crashed sweeps self-heal without waiting for the bulk path.

**Leg 2 — bulk pruning in `loom-daemon clean`:**
Add a `clean_sweep_transients()` pass to `loom-daemon/src/worktree_ops/clean.rs`, wired into `run_clean()` (and `CleanupStats` + `print_summary`), that:
1. Deletes `main-clean-baseline-<RUN_ID>.txt` files whose RUN_ID has no entry in `.loom/sweep-run/` with a live PID (reuse the registry's liveness semantics: entry missing, or PID absent/dead) **and** whose mtime is older than a threshold (48h default) — both conditions, so a just-started run whose registry write races the scan is never touched (AC 4).
2. Deletes the legacy un-keyed `main-clean-baseline.txt`.
3. Prunes `issue-<N>.json` checkpoints whose issue is CLOSED. Use the REST endpoint (`gh api repos/{owner}/{repo}/issues/N --jq .state`) rather than `gh issue view` — GraphQL quota exhaustion is a live failure mode here (#4432 epic context), and gate the per-issue API probe on the same 48h mtime threshold to bound API calls. If the state probe fails (offline, rate-limited), skip that file — never delete on an unverified guess.
4. Respects `--dry-run` like every other `clean_*` function in the module.

**Threshold**: hardcode 48h as a module constant (mirroring `clean::DEFAULT_GRACE_PERIOD_SECS` precedent); a CLI flag is optional scope, not required.

**Out of scope**: `sweep.md`'s per-issue checkpoint lifecycle (`sweep-checkpoint.sh delete N` on merge-done) already works — do not change it. The daemon's registry (`.loom/sweep-run/`) pruning also already works. Only the baseline + orphaned-closed-issue-checkpoint gaps are in scope.

## Affected Files

- `defaults/scripts/sweep-run-registry.sh` — `cmd_cleanup()` (and optionally `prune_dead`) also removes the RUN_ID-keyed baseline file
- `defaults/scripts/tests/test-sweep-run-registry.sh` — cases: cleanup removes own baseline; peers-prune removes dead peer's baseline; live peer's baseline untouched
- `loom-daemon/src/worktree_ops/clean.rs` — new `clean_sweep_transients()` + `CleanupStats` fields + `run_clean()`/`print_summary` wiring + unit tests
- `defaults/.claude/commands/loom/sweep.md` — update the baseline-lifetime prose (~line 1366 and the #3768 row ~line 1996) to state the baseline is deleted by `sweep-run-registry.sh cleanup` at sweep end
- `loom-daemon/src/main.rs` — only if a `--sweep-transients-age` flag is added (optional; module constant preferred)

## Test Plan

- [ ] Automated (bash): extend `defaults/scripts/tests/test-sweep-run-registry.sh` — after `new` + baseline file creation + `cleanup <RUN_ID>`, the baseline is gone; a live peer's baseline survives; a dead peer's baseline is pruned
- [ ] Automated (Rust): unit tests in `clean.rs` for `clean_sweep_transients()` — dry-run deletes nothing; young baseline (mtime < 48h) survives; old baseline with live registry entry survives; old baseline with no/dead registry entry is deleted; legacy `main-clean-baseline.txt` is deleted; closed-issue checkpoint deleted, open-issue checkpoint kept, state-probe-failure checkpoint kept
- [ ] Manual verification: on the operator host, run `loom-daemon clean --dry-run` and confirm it lists ~227 dead baselines + the closed-issue checkpoints, then run for real and confirm `.loom/sweep-checkpoint/` shrinks to live-run files only
- [ ] Edge cases: concurrent live sweep during a `clean` run (its baseline + checkpoint must survive); baseline file with malformed RUN_ID suffix (treat as prunable-by-age only)

rjwalters:
## Summary

`.loom/sweep-checkpoint/` grew without bound (200+ dead `main-clean-baseline-*.txt` files since 07-23) because no component owned the lifetime of a sweep's per-run transients. This gives them two owners: the sweep-end hook deletes the run's own baseline, and `loom-daemon clean` gains a bulk backstop for the ones a crashed sweep leaves behind.

## Changes

**Leg 1 — sweep-end self-cleanup (`defaults/scripts/sweep-run-registry.sh`)**

- New `remove_run_artifacts <RUN_ID>` helper: a run's registry entry **and** its `.loom/sweep-checkpoint/main-clean-baseline-<RUN_ID>.txt` are one artifact set with one lifetime.
- `cmd_cleanup` removes both at sweep end.
- `prune_dead` / `cmd_peers` reap a dead peer's baseline along with its registry entry, so a SIGKILLed sweep self-heals on the next sweep's peer scan rather than waiting for the bulk path.
- Missing files are a silent no-op; the RUN_ID charset is already restricted to `[A-Za-z0-9-]`, so the path construction is injection-safe.

**Leg 2 — bulk prune (`loom-daemon/src/worktree_ops/clean.rs`)**

New `clean_sweep_transients()` category, wired into `run_clean()` (full-pass only, alongside agent-config cleanup) and `CleanupStats` / `print_summary`, honoring `--dry-run`. Three categories:

1. `main-clean-baseline-<RUN_ID>.txt` whose run is not live (no `.loom/sweep-run/<RUN_ID>.json`, **or** a registered PID that is dead) **and** older than `SWEEP_TRANSIENT_MIN_AGE_SECS` (48h).
2. The legacy un-keyed `main-clean-baseline.txt`, in both its `.loom/sweep-checkpoint/` and older `.loom/` locations — no live run can own it, so age does not apply.
3. `issue-<N>.json` checkpoints older than 48h whose issue the forge reports CLOSED.

`gh::issue_state_rest()` is added for category 3: `gh api repos/{owner}/{repo}/issues/N` (REST), **not** `gh issue view` (GraphQL) — GraphQL quota exhaustion under concurrent agents is a live failure mode in this repo, and this is a per-file probe.

**Fail-safe by construction** — every one of these means *keep*: an in-flight claim lock for the issue (`.loom/locks/issue-<N>/` + spawn-loop state, i.e. daemon-owned sweeps), an `UNKNOWN`/`OPEN` issue state, an unreadable mtime, a registry entry whose JSON or `pid` cannot be parsed, and any filename matching neither naming pattern. The 48h gate is the belt-and-suspenders backstop for both liveness registries and also bounds forge probes per pass.

**Docs** — `defaults/.claude/commands/loom/sweep.md`: the Step 0a cleanup section and the baseline-lifetime prose now state that `cleanup "$RUN_ID"` removes the baseline too, with `loom-daemon clean` as the bulk backstop.

- [x] **Sweep completion removes this run's own `main-clean-baseline-<RUN_ID>.txt`** — `cmd_cleanup` → `remove_run_artifacts`; asserted by new case 9a in `test-sweep-run-registry.sh`.
- [x] **A bulk path prunes baselines whose RUN_ID is not in the live registry and is older than a threshold, plus per-issue checkpoints whose issue is CLOSED** — `clean_sweep_transients()` categories 1 and 3; 48h module constant.
- [x] **The legacy un-keyed `main-clean-baseline.txt` is removed by the bulk path** — category 2 (both historical locations).
- [x] **Pruning never touches the checkpoint/baseline of a live run (registry + PID liveness check)** — `sweep_run_is_live()` requires an entry *and* a live PID; `active_spawn_loop_issues()` spares in-flight checkpoints; unit tests `..._keep_live_run_baseline_regardless_of_age` and `..._keeps_checkpoint_of_in_flight_sweep` pin both.

## Test Plan

- [x] **Shell** — `bash defaults/scripts/tests/test-sweep-run-registry.sh`: **20 passed, 0 failed** (was 13). New cases: cleanup removes own baseline; a peer scan reaps a dead peer's baseline; a live run's baseline survives a peer scan; a live peer's baseline + entry survive another run's cleanup; per-issue checkpoints are never touched by this helper.
- [x] **Rust** — 12 new unit tests in `clean.rs`, all green: dead-run + old baseline pruned; live-registry baseline kept at 1000h; registered-but-dead-PID baseline pruned; young orphan kept; corrupt registry entry kept; both legacy files removed; unrelated filenames untouched; `--dry-run` deletes nothing while reporting counts; CLOSED checkpoint removed / OPEN + UNKNOWN kept; young CLOSED checkpoint kept; in-flight-claim checkpoint kept.
- [x] `cargo test --workspace --lib --bins` — 2181 passed. (`disk_headroom::tests::test_worktree_root_free_gb_returns_a_value` flaked once under parallel load — a `df -Pk` subprocess artifact, passes in isolation, unrelated to this change.)
- [x] `cargo fmt --all --check`, `cargo clippy` with the CI flag set, `shellcheck` + `bash -n` on both scripts — all clean.
- [x] **Manual** — replayed this host's real `.loom/sweep-checkpoint/` into a fixture with aged files, one live registered run, one crashed run (dead PID), a legacy un-keyed file, and CLOSED/OPEN checkpoints. `--dry-run` reported 4 baselines + 2 closed-issue checkpoints and changed nothing; the real run removed exactly those 6 and spared the live run's baseline, the OPEN issue's checkpoint, and every sub-48h file.
- [x] **Edge cases** — missing `.loom/sweep-checkpoint/` is a no-op; a filename that does not match either pattern is skipped, not deleted.

Closes #4450


rjwalters:
✅ **Approved!**

Both legs match the issue's implementation guidance closely, the fail-safe posture is right, and the test coverage is genuinely load-bearing rather than decorative.

## What I verified

**Code review**

- `remove_run_artifacts()` correctly unifies the two RUN_ID-keyed transients under one lifetime, and is wired into all three call sites (`cmd_cleanup`, `prune_dead`, `cmd_peers`). The `usage()` `sed -n '3,61p'` range was re-checked against the grown header — `--help` output is intact and includes the updated `cleanup` line.
- `clean_sweep_transients_with()` is dependency-injected (`SweepTransientEnv`: clock, min-age, `pid_alive`, `issue_state`), which is why the 12 new tests can pin real decisions instead of mocking a filesystem clock. Good design choice.
- The fail-safe direction is correct on every branch I could find: unreadable mtime → keep; unparseable registry JSON or missing `pid` → treated as LIVE → keep; `UNKNOWN`/`OPEN` issue state → keep; in-flight claim lock → keep; unmatched filename → never touched. `file_age()`'s `duration_since(...).unwrap_or_default()` also makes a future mtime read as age 0 (keep), not as a huge age.
- `gh::issue_state_rest()` is the right call — REST, not GraphQL, per the issue's explicit guidance. Confirmed live that `gh api repos/{owner}/{repo}/issues/N --jq .state` returns lowercase (`open`), so the `to_uppercase()` normalization is necessary, and that a 404 exits non-zero → `UNKNOWN` → keep.
- The 48h age gate correctly doubles as a forge-probe bound; `clean_sweep_transients` is only reached on a full pass (`all_targets`), same gating as `clean_agent_config`, and `loom-daemon clean` is a manually-invoked path (no cron/daemon caller), so the added `gh` subprocesses are not on a hot path.

## Test Execution

- `bash defaults/scripts/tests/test-sweep-run-registry.sh` — ✅ **20 passed, 0 failed** (matches the PR claim; the 5 new cases all present and passing).
- `cargo test -p loom-daemon --lib worktree_ops::clean` — ✅ **17 passed, 0 failed** (12 new).
- `cargo fmt --all --check` — ✅ clean. `cargo clippy -p loom-daemon --lib --all-targets` — ✅ no warnings.
- **Manual end-to-end** (built the binary, ran against a throwaway fixture repo with an aged dead-run baseline, an aged baseline owned by a genuinely live registered PID, a young orphan, both legacy un-keyed files, a CLOSED (`issue-3784`) and an OPEN (`issue-4450`) checkpoint, and an unrelated `README.txt`):
  - `--dry-run` reported 4 baselines + 1 checkpoint and **changed nothing on disk**.
  - The real run removed exactly the dead orphan, both legacy files, and the closed-issue checkpoint; it printed `Keeping baseline of live sweep run:` for the live PID's 3-day-old baseline, `Issue #4450 is OPEN - keeping`, and left the young orphan and `README.txt` untouched. AC 4 (never touch a live run) verified against a real live PID, not just a stub.
- CI: all checks pass (`Rust Unit Tests`, `Rust Linting & Build Check`, `Shellcheck`, `Installer Integration Tests`, …), `mergeStateStatus: CLEAN`, body has `Closes #4450`.

## Non-blocking nits (optional; take them or leave them)

1. **`prune_dead` deletes by JSON-derived `run_id`, not by filename.** The header comment says the path construction is injection-safe because "the RUN_ID charset is already restricted to `[A-Za-z0-9-]`" — true for IDs produced by `gen_run_id`, but in `prune_dead` the `rid` handed to `remove_run_artifacts` comes from the entry's `run_id` **field**, which is never re-validated. A corrupt/hand-edited `A.json` whose field says `B` would remove `B.json` (potentially a live peer's entry) plus `B`'s baseline; a field containing `../` would escape the directory. This is a fully local, already-compromised-file scenario (only this script writes there), so it is defense-in-depth, not a live vulnerability — but it is a one-liner to close, either by deriving the deletion key from the filename stem (`"$(basename "$file" .json)"`) or by adding `[[ "$rid" =~ ^[A-Za-z0-9-]+$ ]] || return 0` at the top of `remove_run_artifacts`. Worth doing since the comment already asserts the invariant.
2. **Cosmetic:** `remove_transient` prints only `file_name()`, so the two legacy locations produce two identical `Removed legacy un-keyed baseline: main-clean-baseline.txt` lines — reads like a duplicate-delete bug in the log. A repo-relative path for that category would disambiguate.

Neither blocks merge. Nice work — the injected-env test harness in particular is what makes the liveness guarantees actually checkable.

Issue #4368, opened 2026-07-29
loom-daemon status: misdiagnoses a booting daemon as wedged (stale prior-boot heartbeat + pre-IPC startup window)


rjwalters:
## Context
Hit twice on 2026-07-28 while operating a fresh host:
1. ~9s after `loom-daemon-start.sh`, `loom-daemon status` reported "The daemon process IS alive … but is not responding over IPC. Heartbeat is STALE (83814s ago …) — the daemon is likely wedged." The heartbeat file was from the previous boot (this machine had migrated state); the daemon was in its normal startup claim-reconciliation pass and bound the socket seconds later, perfectly healthy.
2. After registering 12 workspaces, a restart's startup pass took ~30s before socket-bind; `status` in that window reported "Heartbeat is fresh … likely an IPC/socket-layer fault" — also wrong, just a longer pre-IPC window.

Both diagnoses steer an operator toward restarting a healthy daemon.

## Suggested acceptance criteria
- [ ] `status` (and the shared probe logic) detects the pre-IPC startup phase: if the daemon process start time is recent (< a startup grace, scaled or generously fixed) and the socket is absent, report "daemon is starting (Xs, startup reconciliation across N workspaces)" instead of wedged/IPC-fault.
- [ ] A heartbeat file older than the daemon process's own start time is reported as "stale file from a previous boot", never as evidence the current process is wedged.
- [ ] The "do NOT run loom-daemon-start.sh" advice only appears when the evidence actually indicates a wedge (process older than grace + no socket + heartbeat older than process start).

---

### Verified against origin/main

**AC 1's core mechanism already exists.** `InstallState::AliveStarting` (#4213, merged as PR #4250, commit `d553d2fe`, 2026-07-28 10:54 PT) classifies a live process whose age is ≤ a startup grace (default `DEFAULT_STARTUP_GRACE_SECS = 90`, env `LOOM_DAEMON_STARTUP_GRACE_SECS`) and whose IPC failed as "still STARTING … NOT a fault", and deliberately suppresses the stop/start remediation (`loom-daemon/src/main.rs` ~line 2862). Both cited incidents (9s and ~30s process age) would classify `AliveStarting` on today's binary — the incidents on 2026-07-28 almost certainly ran a binary predating that same-day merge. **Builder: verify AC 1 against a current build first; do not re-implement the grace window.**

**Remaining genuine gaps (the actual scope):**

[...]

PR #4530, opened 2026-07-30
feat(watchdog): add bounded in-band IPC liveness probe with pid-keyed debounce

### LINKED ISSUE #4398: Hang-aware watchdog liveness probe: bounded-timeout CLI check, not just process-exists (deferred from #4381)
## Context

The #4381 incident showed the worst-case failure mode for the daemon watchdog: the production `~/.local/bin/loom-daemon` was replaced by a stub that answered `--version` correctly and then hung forever (`while true; sleep 1`). A process-exists (pid-alive) watchdog check passes indefinitely against such a binary — and had the real daemon crashed during that 9-hour window, launchd would have relaunched the hung stub and the watchdog would have reported everything healthy.

PR #4396 (the incident fix) deliberately deferred this as a new feature rather than a test-suite fix. The related watchdog-contract issue #4331 was closed as completed on 2026-07-29 **before** this was scoped, so no open issue tracks it.

## Proposal

Add a bounded-timeout liveness probe to the watchdog path (and/or `loom-daemon status`): e.g. `timeout 10 loom-daemon status` (or a lighter socket ping) rather than pid-existence alone. A probe that times out should be treated as *unhealthy* — distinct from #4331's starting/expected-but-down/not-expected states (see its closing discussion about the ~20–30s socket-bind window after a supervised relaunch, which the probe timeout must tolerate).

## Related

- #4381 (incident) · PR #4396 (sandbox + checksum guard) · #4331 (closed watchdog-contract work) · #4397 (the other deferred AC)

Filed during Judge review of PR #4396.

---

### Fresh evidence this class is live, not hypothetical

Observed 2026-07-29 ~18:45 UTC: the production daemon at pid 1484 was alive (`kill -0` succeeds, launchd shows a live pid) with a **fresh heartbeat file**, while `loom-daemon status` — the IPC/socket round-trip — **timed out (>5s)**. This is exactly the failure mode this issue proposes to detect, and it explains *why* pid-alive + heartbeat-fresh is not sufficient on its own (see root cause below) — not just why pid-alive alone (the #4381 stub) is insufficient.

### Root cause: heartbeat and IPC are on independent tokio tasks

- The daemon heartbeat write and the IPC accept/serve loop are **decoupled**. `#[tokio::main]` (`loom-daemon/src/main.rs:995-996`) defaults to the **multi-threaded** tokio runtime; `daemon_heartbeat::spawn_heartbeat_task` (`loom-daemon/src/daemon_heartbeat.rs:205,221`) runs as its own `tokio::spawn`ed task independent of whatever task is servicing the Unix socket.
- Consequence: if the IPC accept loop / a specific connection-handling task wedges (deadlock, blocking call not yielding, poisoned lock, etc.) while the rest of the runtime keeps scheduling other tasks, the heartbeat task can keep touching `daemon.heartbeat` on schedule — passing both of the watchdog's existing checks — while every `status` (and other IPC) round-trip hangs. Today's pid-1484 observation is a live instance of exactly this decoupling.

### Current watchdog behavior (verified against `origin/main`, file:line anchors)

`defaults/scripts/cli/loom-daemon-watchdog.sh` never talks to the daemon over its socket at all — it only ever probes two out-of-band signals:

| Check | What it verifies | Reference |
|---|---|---|
| 1. Marker presence | Is a daemon *expected*? | `defaults/scripts/cli/loom-daemon-watchdog.sh:238-265` |
| 2. `detect_daemon_liveness` | `launchctl print` pid / `kill -0` on the pid file — **process existence only** | `defaults/scripts/cli/loom-daemon-watchdog.sh:159-197` |
| 3. Heartbeat file freshness | Is `daemon.heartbeat`'s mtime within `HEARTBEAT_INTERVAL_SECS * 5` (min 300s)? | `defaults/scripts/cli/loom-daemon-watchdog.sh:360-414` |

All four `exit 0` paths in the heartbeat section (fresh: `:401-402`; unreadable mtime: `:404-406`; no heartbeat file: `:413-414`; previous-boot heartbeat: `:392-395`) report "healthy" purely from a filesystem mtime — none of them ever exercise the IPC path the daemon actually serves work over. That is the gap this issue closes.

Meanwhile the CLI already *has* a correctly-bounded IPC probe the watchdog can reuse rather than reinvent: `query_daemon_status` / `query_daemon_status_once` (`loom-daemon/src/main.rs:2472-2540`) wraps both the connect and the round-trip in independent `tokio::time::timeout(Duration::from_secs(5), …)` calls, with one bounded reconnect retry (#4279) for the transient "dropped before reply" case only (never for a genuine timeout). A hung daemon (today's pid-1484 case) surfaces as `"status round-trip timed out after 5s"` reliably within roughly 5-10s. This means the watchdog does **not** need a new probe primitive in the daemon binary — it needs to *invoke the existing one* and act on the result.

### Proposed design

1. **Add a bounded IPC probe call** to the watchdog, gated on `daemon_alive == true` (no point probing IPC on a process that already fails the pid-liveness check — that path already reports DIVERGENCE). Shell out to the installed `loom-daemon` CLI, e.g.:
   ```bash
   LOOM_SOCKET_PATH="$SOCKET_PATH" timeout "$PROBE_TIMEOUT_SECS" loom-daemon status --json >/dev/null 2>&1
   ```
   `PROBE_TIMEOUT_SECS` should be strictly larger than the CLI's own internal budget (5s connect + 5s round-trip, occasionally x2 for the #4279 retry) — a default of **15s** gives headroom without materially lengthening a `StartInterval` tick. Make it overridable (`LOOM_WATCHDOG_STATUS_PROBE_TIMEOUT_SECS`), mirroring the existing `LOOM_WATCHDOG_KICKSTART_RECHECK_*` override pattern (`:89-93`).
2. **Where the loom-daemon binary is resolved.** Reuse `loom-daemon-start.sh`'s `LOOM_DAEMON_BIN` → `command -v loom-daemon` resolution order (`defaults/scripts/cli/loom-daemon-start.sh:159-163`) rather than assuming it's on `PATH`; degrade gracefully (skip the probe, do not false-report) if no binary can be resolved at all — this must never make the watchdog itself a new hard dependency that pages when e.g. only the launchd-supervised binary exists and the invoking shell's `PATH` differs.
3. **Debounce before escalating** ("confirmed hang", not a single blip): a `StartInterval` watchdog tick already re-runs every cadence regardless of exit code (see the file's own "WHY A SECOND LAUNCHD JOB" rationale, `:27-35`), so treat a single timed-out probe as `DIVERGENCE` on that tick (loud, logged) but reserve any *stronger* action (see AC below) for **N consecutive** timed-out probes tracked via a small state file (e.g. `$LOOM_DIR/.watchdog-probe-fail-count`), mirroring how heartbeat staleness already uses a generous multiple (5x cadence) rather than a single missed write (`:360-369`) to avoid flapping.
4. **Escalation action, scoped conservatively.** Unlike the #4232 auto-`kickstart` remediation — which fires only for a mathematically narrow, always-safe signature (job loaded + not running + last exit 0) — there is no safe unattended way to "fix" a wedged-but-alive process: the only real remediation is killing and relaunching it (`launchctl kickstart -k`, or a `stop`+`start`), which risks killing a daemon that is merely under heavy legitimate load rather than actually hung. **Do not add an automatic kill/restart in this issue** — scope the escalation to a maximally loud, actionable `DIVERGENCE` report (exit 1) with explicit recovery guidance (`loom-daemon restart`, or `stop.sh && start.sh` if the restart primitive is itself unreachable through the wedged socket). Auto-kill remediation, if wanted later, should be its own follow-up issue with its own narrow, provably-safe gate — do not conflate scope here.
5. **Tolerate the post-relaunch socket-bind window** (#4331's closing comment, 2026-07-29): for ~20-30s after a supervised relaunch the socket may not be bound yet even though launchd shows a live pid. A single failed probe during that window must not immediately count toward the "confirmed hang" threshold as a *stale/wedged* signal — cross-reference the live pid's process age (`process_age_secs`, `:230-235`, already used for the #4368 prior-boot heartbeat check) and skip/soften the probe (or don't increment the fail-counter) while the process is younger than a small grace window (e.g. 30-45s).

### Suggested acceptance criteria

- [ ] Watchdog performs a bounded-timeout `loom-daemon status` (or lighter equivalent) IPC probe whenever `daemon_alive == true`, in addition to the existing pid + heartbeat checks.
- [ ] The probe itself is bounded (default ~15s, overridable via env) and never causes the watchdog invocation as a whole to hang indefinitely — even if the daemon accepts the connection and then never responds at all.
- [ ] A single timed-out probe is reported (`DIVERGENCE`, logged) but does not by itself trigger any auto-remediation — matching the conservative gate described above.
- [ ] **No false positives under heavy dispatch load**: a daemon serving a burst of concurrent sweeps (the #4279 "dropped before reply" contention case) must not be reported as hung — verify against the CLI's existing bounded-retry behavior (`query_daemon_status`, `main.rs:2472-2485`), and/or add a test that simulates a slow-but-eventually-responsive `status` reply completing within the probe's timeout window.
- [ ] **No false positives during the post-relaunch socket-bind window**: a probe run against a daemon younger than the grace window described above must not count toward "confirmed hang."
- [ ] On **N consecutive** timed-out probes (confirmed, sustained hang — pid alive, heartbeat fresh, IPC unresponsive across multiple ticks), the watchdog escalates to a maximally actionable report: distinct message text (e.g. "daemon IPC UNRESPONSIVE — process alive/heartbeat fresh but socket round-trip hangs") + explicit recovery command + exit 1. Whether this also triggers a bounded, narrowly-gated auto-restart is a design decision for the Builder to make explicit in the PR description — not a silent default.
- [ ] `test-loom-daemon-watchdog.sh` gains cases for: probe succeeds (healthy, unchanged behavior); probe times out once (DIVERGENCE, no escalation); probe times out N times consecutively (escalation path exercised); probe skipped/softened for a young process (grace window); no `loom-daemon` binary resolvable (graceful degrade, no false DIVERGENCE).
- [ ] Existing watchdog tests (pid-alive/heartbeat-only paths) continue to pass unmodified in spirit — this is additive, not a replacement of the existing checks.

### Affected Files

- `defaults/scripts/cli/loom-daemon-watchdog.sh` — add the bounded IPC probe call, debounce/fail-count state, escalation report text, and env overrides (core change)
- `defaults/scripts/tests/test-loom-daemon-watchdog.sh` — new test cases per the AC above (currently 447 lines; will need a mockable `loom-daemon` stand-in for probe success/timeout/absent-binary cases)

### Test Plan

- [ ] Shell tests: all new cases enumerated in the AC above, plus regression of every existing case in `test-loom-daemon-watchdog.sh` (marker-absent quiet path, marker-present + not-running DIVERGENCE, #4232 auto-kickstart signature, heartbeat-stale DIVERGENCE, heartbeat-fresh OK).
- [ ] Manual verification (best-effort, since reproducing a genuine tokio-task wedge on demand is hard): point the watchdog at a fake `loom-daemon` binary whose `status` subcommand sleeps past the probe timeout, confirm DIVERGENCE after N ticks and no false OK.
- [ ] Confirm the probe timeout budget (`PROBE_TIMEOUT_SECS` default) is comfortably larger than the CLI's own worst-case bounded round-trip (5s connect + 5s round-trip, x2 for the #4279 retry) so the watchdog's own timeout is never the thing that fires first on a merely-slow-but-healthy daemon.

### Related

- #4381 (incident) · PR #4396 (sandbox + checksum guard that deferred this) · #4331 (closed watchdog-contract work — starting/expected-but-down/not-expected state semantics this probe must not collide with) · #4397 (the other deferred #4381 AC — binary-format sanity gate, separate concern) · #4279 (the bounded-retry contention case this issue's false-positive AC must respect) · #4232 (the narrow, always-safe auto-remediation precedent this issue's escalation AC deliberately does NOT extend to an unsafe auto-kill)

### Scope note

Shell-only change (watchdog script + its test suite) — no Rust changes needed, since the bounded IPC probe primitive (`query_daemon_status`) already exists and is already correctly timeout-bounded; this issue is "teach the watchdog to call it," not "build a new probe." Single PR.

rjwalters:
## Summary

The daemon watchdog only ever probed **out-of-band** signals — process existence (`launchctl print` pid / `kill -0`) and heartbeat-file mtime. Neither ever talks to the daemon over the socket it actually serves work on, and two incidents show that gap is live, not hypothetical:

- **#4381** — the installed `loom-daemon` was replaced by a stub that answered `--version` and then hung forever (`while true; sleep 1`). `kill -0` passes against that indefinitely.
- **2026-07-29** — the production daemon (pid 1484) was alive **and writing a fresh heartbeat** while every `loom-daemon status` round-trip hung. `daemon_heartbeat::spawn_heartbeat_task` and the IPC accept loop are independent `tokio::spawn`ed tasks on a multi-threaded runtime, so one keeps ticking while the other is wedged — pid-alive + heartbeat-fresh both pass while dispatch is dead.

This adds a **bounded in-band IPC round-trip** as a third check, gated on `daemon_alive == true`. Shell-only, as scoped.

## Changes

`defaults/scripts/cli/loom-daemon-watchdog.sh`

- **New section 3** (heartbeat becomes section 4): `run_ipc_probe` shells out to the resolved `loom-daemon` CLI and classifies the result as `healthy` / `unresponsive` / `skipped`.
- **`bounded_run`** — hard external wall-clock budget (default 15s, `LOOM_WATCHDOG_STATUS_PROBE_TIMEOUT_SECS`), normalizing every implementation to `timeout`'s 124. macOS ships **no `timeout(1)`**, so the fallback is a real bounded implementation (background child + killer subshell), not a degrade-to-unbounded — "the CLI never returns" is precisely the failure being detected. `-k 2` on the `timeout(1)` path is load-bearing: a non-interactive bash blocked in a foreground command defers SIGTERM, so without the KILL escalation `timeout` itself would wait forever.
- **`locate_daemon_bin`** — same resolution order as `loom-daemon-start.sh` (`LOOM_DAEMON_BIN` → `command -v` → in-repo cargo targets, rooted at the marker's `repo_root`).
- **Debounce state** — `<loom_dir>/.watchdog-probe-fail-count`, **keyed to the live pid**, threshold 3 consecutive ticks (`LOOM_WATCHDOG_IPC_PROBE_FAIL_THRESHOLD`). A success clears the streak; a relaunched daemon never inherits its predecessor's.
- **Startup grace** — a process younger than 90s (`LOOM_WATCHDOG_IPC_PROBE_GRACE_SECS`, defaulting through `LOOM_DAEMON_STARTUP_GRACE_SECS`, mirroring `daemon_install_state::DEFAULT_STARTUP_GRACE_SECS`) is skipped outright, so the post-relaunch socket-bind window can never count toward a hang. `proc_age` is now computed once and shared with the #4368 prior-boot heartbeat check.
- **`exit_ok`** — the heartbeat section's otherwise-healthy `exit 0` paths now exit 1 if a probe divergence was already reported on this tick, so a wedged-but-heartbeating daemon does not exit 0.
- **Opt-out** — `LOOM_WATCHDOG_IPC_PROBE=0`.

`defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 35 new assertions (20 → 55).
`defaults/docs/daemon-reference.md` — decision-matrix rows, env table, and an "In-band hang detection (#4398)" subsection.

## Two deliberate design decisions (called out explicitly, per the issue)

**1. The probe is `quarantine list`, NOT `status --json`.** The issue proposed `timeout 15 loom-daemon status --json`. That would have been a guaranteed false-positive generator: after a *successful* round-trip, `handle_status_command` also runs `collect_token_usage` (a per-account **curl probe** of the token pool, staggered) plus `self_update::check()`. Measured on this host against a **healthy** daemon:

```
$ time loom-daemon status --json    # rc=0
real  0m15.289s
$ time loom-daemon quarantine list  # rc=0
real  0m0.005s
```

15.3s vs. the proposed 15s budget — the probe's own timeout would have fired on a perfectly healthy daemon on the very first tick. `quarantine list` is a pure IPC round-trip (`Request::ListQuarantines` → `query_daemon` → `query_daemon_bounded`, 5s-bounded, no post-reply work). Overridable via `LOOM_WATCHDOG_IPC_PROBE_ARGS` for operators who want `status --json` (the classifier already recognizes its `alive-but-unresponsive` / `alive-starting` output). The issue's premise — "invoke the existing bounded primitive, don't build a new one" — is preserved; only the *subcommand* changed, and the exit-code/output classifier handles both.

**2. NO auto-restart, in either the single-failure or the confirmed-hang path.** Per the issue's guidance: unlike #4232's mathematically narrow always-safe `kickstart` gate, there is no provably-safe unattended remediation for a wedged-but-alive process — the only real fix is killing it, which would equally kill a daemon merely under heavy legitimate load. A confirmed hang escalates to a maximally actionable report and stops:

```
[DIVERGENCE] daemon IPC UNRESPONSIVE (CONFIRMED): the process is alive (...) — and its heartbeat
may well look FRESH — but the bounded socket round-trip has now failed on 3 CONSECUTIVE watchdog
ticks (threshold 3). ... No automatic kill/restart is attempted (#4398 ...). RECOVER: 'loom-daemon
restart' (note: the restart primitive travels over this same wedged socket and may itself hang),
else ./.loom/scripts/cli/loom-daemon-stop.sh && ./.loom/scripts/cli/loom-daemon-start.sh [flags].
Diagnose with: ... 'sample <pid>' (macOS) or 'gdb -p <pid>' to capture the wedge before killing it.
```

Auto-kill remediation, if ever wanted, needs its own narrow provably-safe gate — a separate issue, deliberately not conflated here.

| AC | Status | Evidence |
|---|---|---|
| Bounded-timeout IPC probe whenever `daemon_alive == true`, in addition to pid + heartbeat | ✅ | New section 3, gated behind the existing `daemon_alive != true` early-exit. The marker-absent-but-alive path (#4331) already exits 1 loudly, so it is not probed — documented in the header. |
| Probe bounded (default ~15s, env-overridable) and never hangs the invocation, even if the daemon accepts then never responds | ✅ | `bounded_run` (double-bounded: CLI's own 5s+5s, plus the external budget). Tests 13/21 use a `while true; sleep 1` stub (the #4381 shape) and assert the whole watchdog run finishes in ~2-3s; test 21 forces the no-`timeout(1)` fallback. |
| A single timed-out probe is reported (DIVERGENCE, logged) but triggers no auto-remediation | ✅ | Test 13: exit 1, log contains "consecutive failure 1 of 3", and asserts the escalation text is **absent**. No `launchctl`/kill call exists on this path. |
| No false positives under heavy dispatch load | ✅ | The CLI's own bounded #4279 reconnect retry is inside the probe budget; test 17 simulates a slow-but-eventually-responsive round-trip (2s reply, 15s budget) → exit 0, no DIVERGENCE. Plus the 3-consecutive-tick debounce (>2 StartInterval periods). |
| No false positives during the post-relaunch socket-bind window | ✅ | Test 15: `ps` stub pins age to 10s (< 90s grace) with a hanging stub → exit 0, no DIVERGENCE, **and no state file created** (does not count toward the tally). |
| On N consecutive timed-out probes, escalate with distinct text + explicit recovery command + exit 1; auto-restart decision made explicit | ✅ | Test 14 (threshold 2, two ticks): asserts `IPC UNRESPONSIVE (CONFIRMED)`, that both `loom-daemon restart` and `loom-daemon-stop.sh` appear, and that the message states no auto-remediation was attempted. Decision stated above: **no auto-restart**. |
| New tests: probe succeeds / times out once / times out N times / young-process grace / no binary resolvable | ✅ | Sections 12, 13, 14, 15, 16 respectively — plus 17 (slow-but-responsive), 18 (daemon-side app error + unsupported subcommand), 19 (streak clear + pid-keying), 20 (opt-out), 21 (portable bound). |
| Existing tests pass unmodified in spirit; additive, not a replacement | ✅ | All 20 pre-existing assertions unchanged and green. The harness now defaults `LOOM_WATCHDOG_IPC_PROBE=0` for them so they stay hermetic (they assert the pid+heartbeat contract and must not depend on whether the host has a `loom-daemon` on PATH or a live socket); the probe cases re-enable it explicitly. |
| Probe budget comfortably > the CLI's worst-case bounded round-trip | ✅ | 15s default vs. 5s connect + 5s round-trip (x2 for the #4279 retry). Documented at the knob and in the header. |

Graceful degradation is also covered beyond the AC: a **daemon-side application error** (exit 1 with `Daemon error: …`) proves IPC works, so it degrades to `skipped` rather than being read as a wedge (test 18) — as does an older CLI that does not know the probe subcommand (clap exit 2, test 18b).

## Test Plan

- [x] `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — **55 tests, 55 passed, 0 failed** (was 20/20)
- [x] `bash defaults/scripts/tests/test-loom-daemon-start.sh` — 46/46 passed (adjacent lifecycle suite, no regression)
- [x] `bash -n` + `shellcheck -S warning` clean on both changed scripts
- [x] Manual verification per the issue's test plan: a fake `loom-daemon` whose subcommand outlives the probe timeout → DIVERGENCE after N ticks, never a false OK (tests 13/14/21 encode exactly this; also reproduced by hand against a live daemon socket)
- [x] Probe budget empirically re-derived (the 15.3s `status --json` measurement above)
- [x] `./.loom/scripts/check-main-clean.sh` — main worktree clean

**Pre-existing failure, unrelated:** `bash scripts/test-installer.sh` reports `test-loom-dispatcher.sh failed (rc=1)`. Verified identical on a clean `origin/main` archive (`git archive origin/main` → same single failure), and the dispatcher suite passes standalone on both — an environment-sensitive pre-existing flake in the installer wrapper, untouched by this PR.

**Observation (not fixed here, out of scope):** `test-loom-daemon-watchdog.sh` is not registered in `.github/workflows/ci.yml` or `build-gate.sh` — it only runs when invoked by hand. That predates this PR; worth a follow-up issue if the maintainer wants the suite gated.

Closes #4398

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -3162,6 +3162,11 @@ loop is not reusable as the reporter). It has three cooperating parts:
 | watchdog interval | `LOOM_WATCHDOG_INTERVAL_SECS` | — | `300` (launchd `StartInterval` / systemd `OnUnitActiveSec`+`OnBootSec`) |
 | watchdog job/unit basename override | `LOOM_WATCHDOG_LABEL` | — | `<daemon label/unit>-watchdog` |
 | staleness threshold | `LOOM_DAEMON_HEARTBEAT_STALE_SECS` | — | `max(5 × cadence, 300)` |
+| IPC probe on/off (#4398) | `LOOM_WATCHDOG_IPC_PROBE` | — | on |
+| IPC probe budget (#4398) | `LOOM_WATCHDOG_STATUS_PROBE_TIMEOUT_SECS` | — | `15` |
+| IPC probe argv (#4398) | `LOOM_WATCHDOG_IPC_PROBE_ARGS` | — | `quarantine list` |
+| confirmed-hang threshold (#4398) | `LOOM_WATCHDOG_IPC_PROBE_FAIL_THRESHOLD` | — | `3` consecutive ticks |
+| IPC probe startup grace (#4398) | `LOOM_WATCHDOG_IPC_PROBE_GRACE_SECS` | — | `LOOM_DAEMON_STARTUP_GRACE_SECS`, else `90` |
 
 **Why an interval timer, not a resident process or `KeepAlive`/`Restart=`.** The
 reporter must itself be supervised, but a long-lived resident watchdog just moves
@@ -3180,7 +3185,8 @@ service's exit code affects the next scheduled run.
 
 | Marker | Reality | Watchdog |
 |--------|---------|----------|
-| present | daemon alive, heartbeat fresh | silent (OK) |
+| present | daemon alive, heartbeat fresh, IPC round-trip OK | silent (OK) |
+| present | daemon alive, heartbeat fresh, **IPC round-trip fails** (#4398) | **report** — 1 tick: "not yet confirmed"; N consecutive: **IPC UNRESPONSIVE (CONFIRMED)** |
 | present | daemon alive, heartbeat **stale**, written this boot (younger than the process) | **report** — daemon may be wedged |
 | present | daemon alive, heartbeat **older than the process itself** (#4368: a previous-boot/enablement leftover) | silent (liveness-only; not evidence about the current process) |
 | present | daemon alive, no heartbeat file | silent (liveness-only; heartbeat disabled or not yet written) |
@@ -3188,7 +3194,53 @@ service's exit code affects the next scheduled run.
 | **absent** | **nothing running** | silent — deliberate stop, no false page |
 | **absent** | **daemon IS running** | **report (WARN)** — state mismatch, crash protection disarmed (#4331) |
 
-The last row is the #4331 fix. Before it, a missing marker short-circuited to a
+**In-band hang detection (#4398).** Rows 1–2 above are the #4398 addition. Both
+pre-existing checks are *out-of-band*: neither ever talks to the daemon over the
+socket it serves work on, and two incidents proved that is not enough. In #4381
+the installed binary was replaced by a stub that answered `--version` and then
+hung forever — a pid-alive check passes against that indefinitely. On 2026-07-29
+the production daemon (pid 1484) was alive **and writing a fresh heartbeat**
+while every `loom-daemon status` round-trip hung: the heartbeat writer
+(`daemon_heartbeat::spawn_heartbeat_task`) and the IPC accept loop are
+independent `tokio::spawn`ed tasks on a multi-threaded runtime, so one keeps
+ticking while the other is wedged. After liveness is confirmed the watchdog now
+also runs a **bounded in-band round-trip** through the resolved `loom-daemon`
+CLI, with these properties:
+
+- **Bounded twice.** The CLI bounds its own connect + round-trip (5s each,
+  `query_daemon_bounded`); the watchdog additionally wraps the invocation in a
+  hard external budget (default 15s), so even a CLI that never returns at all
+  cannot hang the tick. macOS ships no `timeout(1)`, so the no-`timeout` path is
+  a real built-in bounded runner, not a degrade-to-unbounded.
+- **`quarantine list`, not `status --json`.** `status --json` looks like the
+  obvious probe but, *after* a successful round-trip, it also runs a per-account
+  token-pool network check plus a self-update check — measured at **15.3s against
+  a healthy daemon**, which would make the probe budget itself the
+  false-positive generator. `quarantine list` is a pure IPC round-trip with no
+  post-reply work (~5ms). Override with `LOOM_WATCHDOG_IPC_PROBE_ARGS`.
+- **Debounced, pid-keyed.** One failed round-trip can be transient contention
+  (#4279). A single failure is reported loudly but only **N consecutive**
+  failures (default 3), tallied in `<loom_dir>/.watchdog-probe-fail-count` and
+  **keyed to the live pid** so a relaunched daemon never inherits its
+  predecessor's streak, are called a confirmed hang. A successful probe clears
+  the streak.
+- **Startup-grace aware.** A probe against a process younger than the grace
+  window (default 90s, the same `daemon_install_state::DEFAULT_STARTUP_GRACE_SECS`
+  `status` uses) is skipped outright and never counts toward the tally — the
+  post-relaunch socket-bind window is not a wedge.
+- **Gracefully degrading.** No resolvable `loom-daemon` binary, a build whose CLI
+  does not know the probe subcommand, or a daemon-side *application* error (which
+  proves IPC works) all skip the probe rather than invent a divergence. The probe
+  never becomes a new hard dependency that pages on its own absence.
+- **Report-only, deliberately.** Unlike #4232's narrow auto-`kickstart`, there is
+  no provably-safe unattended remediation for a wedged-but-alive process — the
+  only real fix is killing it, which would equally kill a daemon merely under
+  heavy legitimate load. A confirmed hang escalates to a maximally actionable
+  DIVERGENCE report (distinct `IPC UNRESPONSIVE (CONFIRMED)` text, explicit
+  recovery commands, exit `1`) and stops there. Auto-kill remediation, if ever
+  wanted, needs its own narrow provably-safe gate.
+
+The last no-marker row is the #4331 fix. Before it, a missing marker short-circuited to a
 bare `[OK] … nothing to check` **without probing reality at all** — so a
 supervised daemon running with its marker gone (see "Marker ownership" below) was
 reported healthy while the watchdog would in fact never revive it. The no-marker
--- a/defaults/scripts/cli/loom-daemon-watchdog.sh
+++ b/defaults/scripts/cli/loom-daemon-watchdog.sh
@@ -42,6 +42,54 @@
 #   OPERATOR INTENT: present ⇒ a daemon is expected; absent ⇒ it was
 #   deliberately stopped (or never started) ⇒ stay silent.
 #
+# HANG-AWARE IPC LIVENESS PROBE (#4398)
+#   Process-exists + heartbeat-fresh are BOTH out-of-band signals: neither ever
+#   talks to the daemon over the socket it actually serves work on. Two observed
+#   incidents show why that is not enough:
+#     - #4381: the installed loom-daemon binary was replaced by a stub that
+#       answered `--version` and then hung forever. A pid-alive check passes
+#       against that indefinitely.
+#     - 2026-07-29: the production daemon (pid 1484) was alive AND writing a
+#       FRESH heartbeat while every `loom-daemon status` round-trip hung. The
+#       heartbeat writer (`daemon_heartbeat::spawn_heartbeat_task`) and the IPC
+#       accept loop are independent `tokio::spawn`ed tasks on a multi-threaded
+#       runtime, so one can keep ticking while the other is wedged.
+#   So after liveness is confirmed, this job also runs a BOUNDED in-band IPC
+#   round-trip through the installed loom-daemon CLI. Design points:
+#     * BOUNDED TWICE. The CLI bounds its own connect + round-trip (5s each,
+#       `query_daemon_bounded`); this job additionally wraps the invocation in a
+#       hard external timeout, so even a CLI that never returns at all (the
+#       #4381 stub) cannot hang the watchdog tick.
+#     * LIGHTWEIGHT SUBCOMMAND, NOT `status`. The probe defaults to
+#       `quarantine list` — a pure IPC round-trip with no post-reply work.
+#       `loom-daemon status --json` looks like the obvious probe but, AFTER a
+#       successful round-trip, it also runs a per-account token-pool network
+#       check plus a self-update check; measured at 15.3s against a HEALTHY
+#       daemon, which would make the probe timeout itself the false-positive
+#       generator. Override with LOOM_WATCHDOG_IPC_PROBE_ARGS if desired.
+#     * DEBOUNCED. A single failed round-trip can be transient contention (a
+#       per-connection task dropping a `status` under concurrent-sweep load —
+#       #4279). One failure is REPORTED (loud, logged) but only N consecutive
+#       failures, tracked in <loom_dir>/.watchdog-probe-fail-count and keyed to
+#       the live pid, are called a CONFIRMED hang.
+#     * STARTUP-GRACE AWARE. For up to ~90s after a supervised relaunch the
+#       socket may not be bound yet even though launchd reports a live pid
+#       (#4213). A probe against a process younger than the grace window is
+#       skipped entirely and never counts toward the confirmed-hang tally —
+#       the same window `daemon_install_state::DEFAULT_STARTUP_GRACE_SECS`
+#       uses, so `status` and this watchdog cannot disagree.
+#     * GRACEFULLY DEGRADING. No resolvable loom-daemon binary, a build whose
+#       CLI does not know the probe subcommand, or any daemon-side application
+#       error (the IPC round-trip demonstrably WORKED) skips the probe rather
+#       than inventing a divergence. The probe must never become a new hard
+#       dependency that pages on its own absence.
+#     * REPORT-ONLY, DELIBERATELY. Unlike #4232's narrow auto-`kickstart`, there
+#       is NO provably-safe unattended remediation for a wedged-but-alive
+#       process: the only real fix is killing it, which would also kill a daemon
+#       that is merely under heavy legitimate load. A confirmed hang therefore
+#       escalates to a maximally actionable DIVERGENCE report (exit 1) with the
+#       explicit recovery commands — never an automatic kill/restart.
+#
 # BOUNDED AUTO-REMEDIATION (#4232)
 #   The watchdog was deliberately report-only until #4232: on 2026-07-28 a
 #   `loom-daemon restart` was ack'd (the running daemon exited 0, honoring its
@@ -66,8 +114,10 @@
 #   1  DIVERGENCE / state mismatch reported — a daemon is expected but is not
 #      running (and either the #4232 remediation gate did not apply, or it fired
 #      but the daemon is still not confirmed running), or is running but its
-#      heartbeat is stale (possibly wedged), OR (#4331) a daemon IS running while
-#      the marker is ABSENT (crash protection disarmed — a WARN state mismatch)
+#      heartbeat is stale (possibly wedged), OR (#4398) it is running with a
+#      fresh heartbeat but its bounded IPC round-trip failed, OR (#4331) a daemon
+#      IS running while the marker is ABSENT (crash protection disarmed — a WARN
+#      state mismatch)
 #   2  usage error
 #
 # Usage:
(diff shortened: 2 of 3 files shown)


rjwalters:
**Approved.**

Reviewed via the existing builder worktree (`.loom/worktrees/issue-4398`), branch up to date with `origin/feature/issue-4398`.

**Verification performed**
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 55/55 pass (~11s wall clock, including the 13/14/21 timeout-simulation cases).
- `bash defaults/scripts/tests/test-loom-daemon-start.sh` — 46/46 pass (adjacent lifecycle suite, no regression).
- `bash -n` + `shellcheck -S warning` clean on both changed scripts (matches the green Shellcheck CI check).
- `gh pr checks 4530` — all applicable checks pass (Installer Integration Tests was still `pending` at first look; re-checked a minute later and it's green). `mergeStateStatus` is `CLEAN`.
- Independently verified the two load-bearing Rust-side claims in the PR description against `loom-daemon/src/main.rs` and `daemon_install_state.rs` on this branch: `query_daemon` (used by `handle_quarantine_command`'s `List` action) is bounded via `query_daemon_bounded(..., Duration::from_secs(5))`, `ListQuarantines` does no post-reply work, and `DEFAULT_STARTUP_GRACE_SECS = 90` — matching the grace-window default the shell script mirrors.

**On the `quarantine list` vs. `status --json` deviation** — this is the right call, not scope creep. The issue's own premise was "invoke the existing bounded primitive, don't build a new one," and the PR does exactly that; only the subcommand changed. The 15.3s measurement against a *healthy* daemon (`status --json`'s post-round-trip token-pool curl + self-update check) vs. the proposed 15s budget is a real, well-evidenced problem — shipping the issue's literal suggestion would have false-positived on the very first tick against a healthy daemon. `quarantine list` preserves the property that matters (pure bounded IPC round-trip, no daemon-side state mutation) while being ~3000x cheaper, and `LOOM_WATCHDOG_IPC_PROBE_ARGS` keeps `status --json` available for operators who want it — the classifier already recognizes its `alive-but-unresponsive`/`alive-starting` output. Well-justified and clearly called out in both the PR body and the script's header comment.

**Acceptance criteria (issue #4398) — all met:**
- Bounded probe gated on `daemon_alive == true`, default 15s, `LOOM_WATCHDOG_STATUS_PROBE_TIMEOUT_SECS`-overridable — yes (`bounded_run`, double-bounded on top of the CLI's own 5s+5s).
- Debounce/fail-count keyed to pid — yes (`.watchdog-probe-fail-count`, `probe_fail_count_for_pid`/`_write`/`_clear`; test 19 exercises the pid-keying explicitly).
- Grace window for post-relaunch socket-bind — yes, 90s default shared with the Rust side via `LOOM_DAEMON_STARTUP_GRACE_SECS` (test 15).
- Graceful degrade when no `loom-daemon` binary resolvable, unsupported subcommand, or daemon-side application error — yes, all three degrade to `skipped` rather than a false DIVERGENCE (tests 16, 18, 18b).
- No auto-restart/auto-kill — confirmed; the confirmed-hang path is report-only with explicit recovery commands, matching the issue's conservative escalation guidance.

**Code quality notes (non-blocking):**
- The `-k 2` KILL-escalation rationale on the `timeout(1)` path and the "bounded twice" framing are well documented — appreciated given how easy this class of bug is to get subtly wrong.
- The portable (`no timeout(1)`) fallback in `bounded_run` correctly normalizes TERM/KILL exit codes to `124` so callers have one code to branch on; test 21 exercises it directly via `LOOM_WATCHDOG_FORCE_PORTABLE_TIMEOUT`.
- Test stubs (`make_daemon_stub`) faithfully mirror the real CLI's actual error text (`Could not reach loom-daemon...`, `Daemon error: ...`, clap's usage-error exit 2) rather than convenient fictions — verified these against `handle_quarantine_command` in `main.rs`.
- Noted but out of scope for this PR (per the PR description itself): `test-loom-daemon-watchdog.sh` isn't wired into `.github/workflows/ci.yml` / `build-gate.sh`, so this suite only runs by hand. Worth a follow-up issue.

Nice, thorough piece of work — the two design deviations (probe subcommand, no-auto-remediation) are both flagged explicitly and backed with evidence rather than asserted.

Issue #4559, opened 2026-07-30
ci: switch backend-tests to cargo nextest (operator step deferred from #4385 — needs workflow scope)


rjwalters:
## Operator step: switch `backend-tests` to `cargo nextest`

Follow-up to #4385. Everything that issue asked for is implemented and merged
**except** the one hunk that actually turns it on in CI: the `backend-tests` job in
`.github/workflows/ci.yml` still runs plain `cargo test --workspace`, so CI remains
exposed to the suite-wide env-mutation race that #4385 diagnosed.

> **⚠️ Curator note (2026-07-30):** at the time this issue was filed,
> the nextest config file was **not** actually on `main` — it only existed on the
> unmerged branch `feature/issue-4385` (PR #4560). That dependency has since
> **merged** (`dd89e0c2`, 2026-07-30). The dependency below is now
> satisfied; the only remaining blocker is the `workflow`-scope credential
> described in the next section.

## Why it was not included in the PR

GitHub refuses **any** write that touches `.github/workflows/**` unless the
credential carries the `workflow` scope (fine-grained PAT: "Workflows: write").
Both write paths were tried and both are rejected for the automation account's
token:

```
$ git push -u origin test/workflow-push-probe-4385
 ! [remote rejected] ... (refusing to allow a Personal Access Token to create or
   update workflow `.github/workflows/ci.yml` without `workflow` scope)

$ gh api --method PUT /repos/rjwalters/loom/contents/.github/workflows/ci.yml ...
 HTTP 403: refusing to allow a Personal Access Token to create or update workflow
 `.github/workflows/ci.yml` without `workflow` scope
```

No GitHub App identity is configured on this host either
(`defaults/scripts/lib/github-app-token.sh` reports unconfigured), so there is no
alternate credential to route through. This needs a human with
`gh auth refresh -s workflow` (or a fine-grained PAT with Workflows: write).

## The change

Two edits to `.github/workflows/ci.yml`.

**1.** In the `changes` job's `backend` paths filter (~line 33), add the nextest
config so a change to it retriggers the Rust job:

```yaml
            backend:
              - 'loom-daemon/**'
              - 'loom-api/**'
              - 'Cargo.toml'
              - 'Cargo.lock'
              - 'rust-toolchain.toml'
              - '.config/nextest.toml'
              - '.github/workflows/ci.yml'
```

**2.** In the `backend-tests` job (~line 69), replace `- run: cargo test --workspace`
with:

```yaml
[...]

PR #4555, opened 2026-07-30
fix(sweep): make PROBE_POOL token count zsh-safe

rjwalters:
## Summary

`defaults/.claude/commands/loom/sweep.md`'s Stage -1 `PROBE_POOL` snippet used
`ls .loom/tokens/*.token 2>/dev/null | wc -l | tr -d ' '` to count materialized
token files. Under zsh (the macOS default shell), an unmatched glob raises a
shell-level error *before* `ls` runs (`no matches found: .loom/tokens/*.token`),
and `2>/dev/null` on `ls` does not suppress it since the error originates from
zsh's glob expansion, not from `ls` itself. This was observed live during a
`/loom:sweep all` Stage -1 probe on a repo with no materialized tokens.

Replaced with `find .loom/tokens -maxdepth 1 -name '*.token' 2>/dev/null | wc -l
| tr -d ' '`, which passes the pattern as a literal argument to `find` and never
triggers shell-level glob expansion, so it is safe under both bash and zsh
regardless of match count.

Per the curated issue, this is the only occurrence of the bare-glob-through-`ls`
pattern in `defaults/.claude/commands/loom/*.md` (the worktree count in the
dry-run verification block, `ls .loom/worktrees/ 2>/dev/null | wc -l`, has no
glob and is unaffected). There is no installed `.claude/commands/loom/sweep.md`
copy in this repo's own working tree to update in parallel.

Closes #4509

## Test plan

- [x] Verified the replacement snippet under bash: absent `.loom/tokens/` dir
      -> `0`, empty dir -> `0`, dir with 2 dummy `.token` files -> `2`, no
      errors in any case.
- [ ] zsh not available on the build host to spot-check directly, but `find
      -name` is not subject to shell glob expansion (unlike `ls *.token`), so
      the fix is expected to behave identically under zsh.

PR #4552, opened 2026-07-30
feat(daemon): port the loom_tools script-helper family to native loom-daemon subcommands

rjwalters:
## Summary

Ports the last family of epic #4081 Phase 3 — the eight `loom_tools` modules that
existed only to back a thin `defaults/scripts/*.sh` entry point — to native Rust
under `loom-daemon/src/script_helpers/`, switches every caller in the same
change, and deletes the Python (8 modules + 7 test files, ~7.5k LOC removed).

| Python module | Native replacement | Shell / CLI entry point |
|---|---|---|
| `log_filter` | `loom-daemon strip-ansi` | `strip-ansi.sh` + the tmux `pipe-pane` filter |
| `model_tiers` | `loom-daemon resolve-model` | `resolve-model.sh`, `resolve-tier-model.sh` |
| `common/usage` | `loom-daemon usage` | `check-usage.sh` |
| `checkpoints` | `loom-daemon checkpoint` | `checkpoint.sh` |
| `claim` | `loom-daemon claim` | `loom-claim` (PATH shim) |
| `sweep_experiment` | `loom-daemon sweep-experiment` | `sweep-experiment.sh`, `agent-metrics.sh --model-experiment` |
| `validate_phase` | `loom-daemon validate-phase` | `validate-phase.sh` |
| `backlog` | — (deleted outright) | none — no shell caller, no console-script consumer, no test file |

This is the **last open issue carrying `loom:epic:4081:phase:3`**, so landing it
clears the epic's Phase 4 gate.

## Changes

**Two hand-synced twin pairs collapsed to one implementation each** (the #4060
contract). `sweep_registry::read_model_aliases` is now the sole
alias-normalization path — split into `model_aliases_from_config` so the
`--config` explicit bypass shares it rather than growing a second copy — and
`worktree_ops::claim_file`'s expiry/abandonment predicates are now `pub` and
reused by the new claim *writer*, so what `loom-claim` writes and what
`recover-orphans` considers live cannot drift. The #4059 blank-alias-key
divergence is resolved in favour of the surviving Rust semantics.

**Callers switched in the same PR.** A new `lib/script-helper.sh`
(`loom_exec_script_helper`) replaces `run_loom_tool` on the six entry scripts.
It `exec`s, which is load-bearing: several of these helpers use non-zero exit
codes as *data* — `resolve-model --tier` / `--task-alias` exit 3 to mean "no
mapping, fall through to your own precedence chain", and `loom-claim` uses
[...]

PR #4320, opened 2026-07-28
fix(daemon): replace Python-conformance epic tests with Rust-native invariants

rjwalters:
## Summary

`loom-daemon/tests/epic_conformance.rs` shelled out to `python3 -m
loom_tools.state_machine --json` as its authoritative expectation source,
asserting the Rust `epic_transition_table()` matched the Python model. This
was the one genuine, live functional dependency blocking `state_machine.py`'s
deletion (epic #4081 Phase 4), and the test's failure mode was a silent
`python3`-unavailable skip rather than a hard failure.

This PR retires the Python-derived conformance mechanism per the issue's
curator-accepted decision (Approach 3): `loom_daemon::epic_state`'s own
`epic_transition_table()` is now the authoritative model of the epic-lane
state machine, and the four structurally meaningful invariants the old test
checked survive as plain Rust tests with no subprocess and no skip path:

- Exactly 5 epic states (`epic_state_ids()`).
- `epic:done` is the sole terminal state (`is_terminal()`).
- Exactly 5 intra-lane edges in `epic_transition_table()`.
- Barrier hygiene: every edge touching `epic:phase_join` (all 3) carries a
  non-empty barrier string.

## Changes

- Replaced `loom-daemon/tests/epic_conformance.rs` with
  `loom-daemon/tests/epic_state_invariants.rs` (Rust-native, always-run
  invariant tests — no `python3` subprocess, no skip path).
- Deleted `loom-tools/src/loom_tools/state_machine.py`,
  `loom-tools/tests/test_state_machine.py`, and the `loom-state-machine`
  console-script entry in `loom-tools/pyproject.toml`.
- Updated doc-comment citations in `loom-daemon/src/epic_state.rs`,
  `loom-daemon/src/issue_creation_mutex.rs`, and
  `loom-daemon/src/phase_join.rs` to name the Rust transition table as
  authoritative instead of the (now-deleted) Python model.
- Reframed `README.md`'s "Loom State Machine" section and
  `defaults/docs/daemon-reference.md`'s epic-state section: the diagram is
  kept as hand-maintained documentation (no more `--mermaid` regen note or
  links to the deleted Python files), and the epic lane now points at
  `epic_state.rs`.

## Accepted trade-off (per the issue)

`state_machine.py` modeled **all four label lanes** (issue, PR, proposal,
epic), and its test validated the whole graph (reachability, dead-ends,
label conflation, autonomy gaps) and kept the README diagram in sync. Only
[...]

PR #4560, opened 2026-07-30
test: run the Rust suite process-per-test under cargo-nextest

### LINKED ISSUE #4385: tests(daemon): suite-wide env-mutation race — #[serial] env tests interleave with unmarked Command::spawn tests, causing flaky CI failures
## Context

Surfaced during PR #4305's review (flake adjudication in its approval thread): the daemon test suite has a **suite-wide environment-mutation race**. ~20 modules correctly mark their env-mutating tests `#[serial]` (serial_test), but `serial` only serializes marked tests **against each other** — unmarked parallel tests still run concurrently with them. Concurrent `env::set_var`/`remove_var` racing `Command::spawn` (which reads the process's environ table) is inherently unsafe and can corrupt a child's environment mid-spawn.

Observed instance (2026-07-29, PR #4305 CI run 30466649321): `pipeline_snapshot::tests::gh_pipeline_source_records_error_but_keeps_other_metrics` — a hermetic test (explicit `with_gh_bin`, absolute tempdir, no env dependency of its own) — panicked at an assertion while unrelated `#[serial]` env tests ran elsewhere in the binary; re-run on the identical head passed. The test was untouched by the PR; the Judge's analysis found no new anti-pattern, just a marginally wider timing window (16 more serial env tests in the binary).

## Why fix it

Intermittent single-test failures burn a full CI re-run cycle per occurrence (~4 min + an adjudication pass per affected PR) and erode trust in red — tonight's occurrence nearly turned a clean approval into a Doctor cycle.

## Suggested acceptance criteria (pick a mechanism)

- [ ] Either: all env-mutating tests migrate to a scoped guard (e.g. `temp_env`-style closure or a process-local env abstraction) that doesn't mutate global environ; or: tests that `Command::spawn` children are also brought under the same serialization domain as env mutators (`#[serial]` on spawners, or `cargo nextest` with a serialized group / process-per-test).
- [ ] The specific observed pair (pipeline_snapshot spawn tests vs any `#[serial]` env test) can no longer interleave — demonstrated by stress-running the workspace suite (e.g. 20 consecutive runs green).
- [ ] Convention documented where daemon test authors will see it (module-level comment or contributing doc): "env mutation and child-spawning tests share one serialization domain."

## Related

- PR #4305 approval thread (the adjudication) · #4259 (the *other* CI-reliability issue — gate timeouts) · serial_test usage across loom-daemon/src/*/tests

---

### Verified diagnosis

Confirmed against `origin/main`:

- The mutation surface is suite-wide, not localized: **298 `#[serial]` attributes across 35 files** and **628 `env::set_var`/`remove_var` sites across 37 files** in `loom-daemon` (re-verify with `grep -rc '#\[serial' loom-daemon/src loom-daemon/tests` and `grep -rn 'env::set_var\|env::remove_var' loom-daemon/src | wc -l`).
- The observed victim (`pipeline_snapshot::tests::gh_pipeline_source_records_error_but_keeps_other_metrics`, `loom-daemon/src/pipeline_snapshot.rs:511`) is hermetic and unmarked: it spawns a fake `gh` script via `Command` (spawn site at line 125) with an explicit `with_gh_bin` path. `serial_test`'s lock is advisory and in-process — unmarked tests in the same binary run concurrently with `#[serial]` env mutators, and `Command::spawn` reads the process environ non-atomically. This race is exactly why `env::set_var` became `unsafe` in Rust edition 2024; this crate is on edition 2021 (`loom-daemon/Cargo.toml:4`), where the API is still "safe" but the race is identical.
- CI runs `cargo test --workspace` (`.github/workflows/ci.yml:69`, `backend-tests` job). `cargo test` runs test *binaries* sequentially but tests *within* a binary on parallel threads — so the race is intra-binary, and the victim and the env mutators live in the same lib test binary.

### Mechanism decision (resolves the "pick a mechanism" AC)

**Recommended: per-test process isolation via `cargo-nextest`.** Each test runs in its own process, so one test's env mutation is invisible to every other test *by construction* — the entire failure class dies, including future tests nobody remembers to mark.

**Fallback: one serialization domain** — add `#[serial]` to every `Command`-spawning test (~30 files have spawning test modules). Works, but is high-touch, serializes the slowest tests (the spawners), and rots: the next unmarked spawn test silently reintroduces the race unless a lint enforces the convention.

**Trap — do not accept a scoped-env-guard-only PR**: migrating env mutators to a `temp_env`-style closure does NOT fix this on its own. A scoped guard still mutates the process-global environ during its window; a concurrent unmarked spawn still races it (`temp_env` only serializes *its own* closures). That approach is only sound combined with the fallback's serialization domain. A Judge should reject a temp_env-only PR as a plausible-but-wrong fix.

**Nextest caveats the Builder must handle (this is where the judgement lives):**

1. **`#[serial]` stops serializing under nextest** — its mutex is per-process and nextest is process-per-test. For *env* mutators that's fine (isolation supersedes serialization). But any `#[serial]` test protecting **non-env shared state** (fixed disk paths, real `$HOME`, global git config, shared statics) loses its protection. The 298 attrs must be audited. Spot checks are encouraging — integration tests use tempdir socket paths and pass env to child daemons explicitly via `.env()` (`loom-daemon/tests/integration_singleton_guard.rs:33`) — but the audit must be systematic, with a summary in the PR body. Genuine cross-process shared state goes in a serialized nextest test group (`.config/nextest.toml`) or `serial_test`'s `file_serial`.
2. **Nextest does not run doctests.** This crate has them (e.g. `loom-daemon/src/health_monitor.rs`, `epic_state.rs`, `init/git.rs`). Keep a `cargo test --workspace --doc` step in CI.
3. **CI needs the binary** — add an install step (e.g. `taiki-e/install-action@nextest`) before the run step.
4. **Local `cargo test` still races.** Acceptable: CI is the arbiter of green. The convention doc below should tell developers to prefer `cargo nextest run` for full-suite local runs.

### Acceptance criteria (sharpened — replaces "pick a mechanism" above)

- [ ] The CI Rust test job runs the workspace suite with per-test process isolation (`cargo nextest run --workspace`), so in-process env mutation can no longer interleave with any other test's `Command::spawn` by construction.
- [ ] Doctests still execute in CI (`cargo test --workspace --doc` retained alongside nextest).
- [ ] `#[serial]` audit: every existing `#[serial]` test either (a) only protects env state (process isolation supersedes it — leave the attr or remove it, Builder's choice, but be consistent), or (b) is placed in a serialized nextest test group / `file_serial`. One-paragraph audit summary in the PR body listing any group members.
- [ ] Stress verification: 10 consecutive green runs of the daemon suite under the new runner (`for i in $(seq 10); do cargo nextest run -p loom-daemon || exit 1; done`) — this specifically covers the observed pair (pipeline_snapshot spawn tests vs `#[serial]` env tests).
- [ ] Convention documented where daemon test authors will see it — crate-level docs in `loom-daemon/src/lib.rs`: env mutation and child-spawning tests share one isolation domain; CI is process-per-test; what to do when a test truly needs cross-process exclusive state.

**Out of scope**: migrating the 628 env-mutation sites to an injectable env abstraction. That is a separate, much larger refactor; process isolation makes it unnecessary for flake elimination.

## Affected Files

- `.github/workflows/ci.yml` — `backend-tests` job (line 69): install nextest, replace `cargo test --workspace` with `cargo nextest run --workspace` plus `cargo test --workspace --doc`
- `.config/nextest.toml` — **new**: default profile; serialized test groups only if the audit finds non-env shared state
- `loom-daemon/src/lib.rs` — crate-level test-convention documentation
- Individual test modules — only those the `#[serial]` audit flags as needing a test group (expected: few or none; enumerate in the PR body)

## Test Plan

- [ ] Manual verification: `cargo nextest run -p loom-daemon` locally; all tests pass under process-per-test execution
- [ ] Stress: 10 consecutive full-suite green runs (AC above)
- [ ] Doctest parity: `cargo test --workspace --doc` still executes a non-zero doctest count after the switch
- [ ] CI: both new steps green on the PR; note `backend-tests` wall-time in the PR body (nextest is typically faster — a large regression is a red flag)
- [ ] Edge case (demonstration, delete before merge): on the branch, temporarily add an unmarked `std::env::set_var` test and confirm it cannot perturb other tests under nextest

rjwalters:
## Summary

Migrates the Rust suite to per-test **process** isolation (`cargo nextest`) to kill
the suite-wide env-mutation race described in #4385: ~770 `env::set_var` /
`remove_var` sites run concurrently with `Command::spawn` calls in the same test
binary, and `spawn` snapshots the process environ non-atomically. `#[serial]` only
serializes *marked* tests against each other, so it never closed that window —
which is why two hermetic, untouched tests failed on 2026-07-29 (`pipeline_snapshot`
on PR #4305, `quarantine_reconciliation` on PR #4320) and passed on re-run.

Process-per-test closes the window **by construction**, including for every future
test nobody remembers to mark. This PR lands the config, the `#[serial]` audit
follow-ups, the harness changes process isolation requires, and the author-facing
convention docs.

> [!IMPORTANT]
> **One operator step is outstanding and this PR cannot be validated in CI without
> it.** See "Operator action required" below. That is why the reference is
> `Part of #4385` rather than `Closes #4385`.

## Changes

- **`.config/nextest.toml`** (new) — `default` + `ci` profiles; `retries = 0`
  deliberately (this removes flakes, it does not paper over them); a
  `daemon-integration` test group with `max-threads = 1` covering
  `binary(/^integration_/)`, declared on **both** profiles.
- **`loom-daemon/src/lib.rs`** — crate-level "Test isolation convention" docs: env
  mutation and child-spawning tests share one isolation domain; what to do when a
  test truly needs cross-process exclusive state.
- **`loom-daemon/tests/common/mod.rs`** — `TestDaemon` now resolves the binary via
  `CARGO_BIN_EXE_loom-daemon` instead of shelling out to `cargo build --bin
  loom-daemon` on *every* `start()`. Under process-per-test those builds serialize
  on `$CARGO_HOME/.package-cache` and the target-dir lock — **cross-process** locks
  that test process isolation does nothing to separate — and the build raced the
  very binary it was about to spawn. Socket-wait and connect-retry budgets are
  re-sized as liveness bounds rather than latency assertions.
- **`loom-daemon/tests/integration_singleton_guard.rs`** — use the shared
  `daemon_bin()` helper (drops a duplicate hardcoded `../target/debug/loom-daemon`
  that ignored `CARGO_TARGET_DIR`).
- **`loom-daemon/src/workspace_pool.rs`** — two safehouse tests now clear the
  safehouse env layer explicitly (`clear_safehouse_env`). They only passed under
  `cargo test` because a *sibling* test in the same process happened to clear those
  vars first; a live Loom agent session exports `LOOM_SAFEHOUSE_*` /
  `SAFEHOUSED_SOCKET`, so under process isolation each test must clear its own.
  This is the one latent accidental-dependency process isolation surfaced.
- **`loom-daemon/tests/README.md`** — `cargo nextest run --workspace` is the
  preferred full-suite command; documents the test group.

### Two nextest traps handled (worth reviewing carefully)

1. **Named profiles do not inherit the default profile's `[[overrides]]`.** An
   override declared only under `[[profile.default.overrides]]` silently does not
   apply to `--profile ci` — the profile CI uses. Caught empirically: a
   `--profile ci` run co-scheduled `integration_basic::test_send_input` with
   `integration_factory_reset::test_factory_reset_loop_creates_clean_slate`, and
   both failed because each calls `cleanup_all_loom_sessions()`, which kills the
   other's tmux sessions. The override is now declared on both profiles, and the
   result is *verifiable* rather than assumed:

   ```
   $ cargo nextest show-config test-groups --profile ci   # 27 tests / 4 binaries
   $ cargo nextest show-config test-groups --profile default # 27 tests / 4 binaries
   ```

2. **`max-threads = 1`, not `threads-required = 'num-test-threads'`.** The latter
   was tried first and rejected: these tests need mutual exclusion against *each
   other* (shared `tmux -L loom` server), not against the ~2350 unit tests, none of
   which touch tmux or spawn a daemon (audited below). Run-completely-alone just
   serializes them behind the whole unit suite for no benefit.

## `#[serial]` audit (AC 3)

**372 `#[serial]` attributes across 39 files**, audited systematically against the
one question that matters under process isolation: *is the protected state
process-scoped or cross-process?* Process isolation supersedes `#[serial]` only for
the former.

**Result: every `#[serial]` unit test protects process-scoped state only** — env
vars via documented override seams (`LOOM_WATCHES_PATH`, `LOOM_SWEEPS_JOURNAL_PATH`,
`LOOM_WORKSPACES_PATH`, `LOOM_SOCKET_PATH`, `LOOM_CONFIG_DEFAULTS_FILE`, …) that
resolve into a per-test `tempfile::TempDir`, plus process-local statics. Evidence
for the cross-process hazard classes:

| Hazard class | Finding |
|---|---|
| `git config --global` writes | 0 occurrences in `loom-daemon` |
| Fixed TCP ports | 0 occurrences |
| Real `$HOME` writes | none — every `dirs::home_dir()` consumer sits behind an env override seam that tests point at a tempdir |
| Real `tmux` from a unit test | 0 — all 29 `Command::new("tmux")` sites are production code, above every `#[cfg(test)]` boundary; `agent_session` unit tests drive a mockable command env |
| Fixed disk paths | the `"/tmp/loom-*"` literals in `role_runner` / `work_finder` / `tokens_pool::paths` tests are *logical identifiers* only — never created, read, or written |
| Shared `tmux -L loom` server + real daemon children | **the 4 `integration_*` binaries only** |

**Group members (the only tests that needed a declaration):** the 27 tests in
`integration_basic`, `integration_security`, `integration_factory_reset`, and
`integration_singleton_guard` — every one calls `cleanup_all_loom_sessions()`
(which kills *every* `loom-*` session host-wide) and/or spawns a real
`loom-daemon`. They are in the `daemon-integration` group (`max-threads = 1`) on
both profiles. `defaults_symlink_containment.rs` mentions `TestDaemon` only in a
doc comment and is hermetic — deliberately not in the group.

`#[serial]` attributes are **kept**, consistently: a no-op under nextest, but still
load-bearing for developers running plain `cargo test`, and they document the
dependency. The filter is `binary(/^integration_/)` (a regex, not a fixed list) so
a newly added `integration_*` suite is covered by default rather than silently
omitted.

## Operator action required before this can be validated or merged

The `backend-tests` job still runs plain `cargo test --workspace`. **The hunk that
flips it could not be committed from this session**: GitHub refuses any write
touching `.github/workflows/**` unless the credential carries `workflow` scope
(fine-grained PATs: *Workflows: write*), and the automation account's token does
not. Verified by both write paths:

```
$ git push origin HEAD:refs/heads/probe-wf-4385b
 ! [remote rejected] HEAD -> probe-wf-4385b (refusing to allow a Personal Access
   Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)

$ gh api -X PUT repos/rjwalters/loom/contents/.github/workflows/... 
 {"message":"Resource not accessible by personal access token","status":"403"}
```

The exact replacement hunk is below; it **applies cleanly to `main` as of
`3f147e8f`** (verified with `git apply --check`). A maintainer runs
`gh auth refresh -s workflow` (or grants *Workflows: write*) and commits it — one
commit, no edits needed:

```diff
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
               - 'Cargo.toml'
               - 'Cargo.lock'
               - 'rust-toolchain.toml'
+              - '.config/nextest.toml'
               - '.github/workflows/ci.yml'
             mcp:
               - 'mcp-loom/**'
@@ -66,7 +67,23 @@ jobs:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test --workspace
+      # Per-test process isolation (#4385). The daemon suite mutates the process
+      # environment in ~600 places while other tests in the same binary
+      # `Command::spawn` children; `Command::spawn` reads the environ
+      # non-atomically, so those two classes raced and produced intermittent
+      # failures in hermetic, untouched tests (2026-07-29: pipeline_snapshot on
+      # PR #4305, then quarantine_reconciliation on PR #4320). `#[serial]` only
+      # serializes *marked* tests against each other, so it never closed that
+      # window. nextest runs each test in its own process, which closes it by
+      # construction. Tests whose shared state is genuinely cross-process (the
+      # `tmux -L loom` server, real daemon children) are given exclusivity in
+      # `.config/nextest.toml`.
+      - uses: taiki-e/install-action@nextest
+      - name: Run tests (process-per-test)
+        run: cargo nextest run --workspace --profile ci
+      # nextest does not run doctests, so libtest still owns those (#4385).
+      - name: Run doctests
+        run: cargo test --workspace --doc
 
   mcp-build:
     name: MCP Server Build
```

Until that lands, `.config/nextest.toml` is live for local runs only and CI stays
exposed to the race. Both `.config/nextest.toml` and the `lib.rs` docs carry an
explicit "outstanding operator step" note pointing here; **delete those two notes
in the same commit that flips the job.**

| 1. CI Rust job runs the suite with per-test process isolation | ⚠️ **blocked on operator** | Hunk written and verified to apply cleanly to `main@3f147e8f`; cannot be pushed by this token (both rejections quoted above). Everything the hunk depends on (`.config/nextest.toml`, harness, test group) is in this PR and verified locally. |
| 2. Doctests still execute in CI | ✅ | `cargo test --workspace --doc` → `10 tests; 2 passed; 0 failed; 8 ignored`. Retained as its own step in the hunk. nextest runs **0** doctests, so the separate step is load-bearing. |
| 3. `#[serial]` audit; group members enumerated | ✅ | Full audit table above. 372 attrs / 39 files; only the 27 `integration_*` tests needed a declaration; `show-config test-groups` confirms the group resolves on **both** profiles. |
| 4. Stress: 10 consecutive green runs | ✅ | See Test Plan. |
| 5. Convention documented where daemon test authors will see it | ✅ | Crate-level docs in `loom-daemon/src/lib.rs` (visible in `cargo doc` and at the top of the file every contributor opens), plus `loom-daemon/tests/README.md`. |

Explicitly **not** done, per the issue: no migration of the env-mutation sites to an
injectable env abstraction, and no `temp_env`-style scoped-guard migration — a
scoped guard still mutates the process-global environ during its window, so a
concurrent unmarked `spawn` would still race it. Process isolation is the fix.

## Test Plan

- ✅ **`cargo nextest run --workspace --profile ci`** — 2379 tests across 12
  binaries; all unit/bin tests pass.
- ✅ **Stress (AC 4): 10 consecutive green runs** of
  `cargo nextest run -p loom-daemon --lib --bins --profile ci` (2312 tests/run,
  ~160 s/run). This is the target class: **both observed victims
  (`pipeline_snapshot`, `quarantine_reconciliation`) live in `loom-daemon/src`**,
  i.e. the lib target, alongside the `#[serial]` env tests they raced.
- ✅ **Isolation demonstration** (the issue's "edge case" item; scratch test
  written, run, and deleted — not in the diff). An unmarked test leaking
  `LOOM_4385_DEMO_LEAK` alongside a test asserting the var is absent:

  ```
  cargo test    → test observer ... FAILED
                  "observed another test's env mutation: LOOM_4385_DEMO_LEAK is set"
  cargo nextest → Summary [0.406s] 2 tests run: 2 passed, 0 skipped
  ```

  The failure mode reproduces on demand under `cargo test` and is *impossible*
  under nextest. This is the whole PR in two lines.
- ✅ `cargo fmt --all -- --check`, `cargo clippy` (CI's exact allow-list),
  `cargo check --workspace --all-targets` — all clean.

### Caveat on the 4 `integration_*` suites (read before judging a red run)

They cannot be stress-verified on this build host and this is **not** attributable
to nextest. They assert on `/tmp/loom-{id}.out` — a **production** path in
`terminal.rs`, not a tempdir — and on the host-global `tmux -L loom` server, while
5+ other Loom sweeps were concurrently running `cargo test -p loom-daemon` from
*other worktrees* on the same machine. Cross-*checkout* collision on a fixed host
path is outside any single test runner's control; on a dedicated CI runner there is
exactly one run and the `max-threads = 1` group serializes them. Their host-global
coupling predates this PR and is left alone as out of scope — worth a follow-up
issue.

Part of #4385

---

## Addendum (second builder pass on the same branch)

**The deferred operator step is now tracked as #4559** (`loom:operator-only`). It
carries the canonical two-hunk replacement — with the corrected call-site count and
a checklist item to delete the two "outstanding operator step" notes in the same
commit. `.config/nextest.toml` and `loom-daemon/src/lib.rs` now cite #4559 by
number rather than "the PR that introduced this file", so the pointer survives the
merge. (The `~600 places` inside the diff block quoted above is stale; #4559 has
`774`.)

**Audit counts refreshed** after merging `main` (#4552 and #4555 landed
mid-review): **374** `#[serial]`-family attributes across **40** files, and **774**
`env::set_var` / `remove_var` call sites in `loom-daemon/src` + `loom-daemon/tests`.
The audit's *conclusion* is unchanged — the newly merged tests are env/tempdir
scoped and add no cross-process state.

**Credential probe, exact transcripts.** Both write paths were re-run against the
current token; the API path returns the workflow-scope message specifically, not the
generic "Resource not accessible" quoted above:

```
$ git push -u origin test/workflow-push-probe-4385
 ! [remote rejected] test/workflow-push-probe-4385 -> test/workflow-push-probe-4385
   (refusing to allow a Personal Access Token to create or update workflow
    `.github/workflows/ci.yml` without `workflow` scope)

$ gh api --method PUT /repos/rjwalters/loom/contents/.github/workflows/ci.yml ...
 HTTP 403: refusing to allow a Personal Access Token to create or update workflow
 `.github/workflows/ci.yml` without `workflow` scope
```

No GitHub App identity is configured on this host either
(`defaults/scripts/lib/github-app-token.sh` reports unconfigured), so there is no
alternate credential to route through. Both probe artifacts (the branch and the ref
created via the API) were deleted.

### Additional verification

| Run | Result |
|---|---|
| `cargo nextest run --workspace --lib --bins --profile ci` × 3 | **2316/2316 pass** every time, at host load average 10.5 / 17.7 / 18.3 on 8 cores |
| `cargo nextest run --workspace --profile ci` × 5 | every unit/bin test passes on every run; only `integration_*` failures, see below |
| `cargo nextest run` on the 4 `integration_*` binaries | **27/27 pass** at load 12.6, in a window with no sibling `cargo test -p loom-daemon` running |
| `cargo test --workspace --doc` | 2 passed, 8 ignored — nextest runs 0 doctests, so the separate step stays load-bearing |
| `cargo fmt --all -- --check`, `cargo clippy --all-targets` (CI's allow-list) | clean |

**The residual `integration_*` failures are host contention, proven not
nextest-specific.** With a sibling sweep running
`cargo test -p loom-daemon --test integration_basic` from another worktree,
`test_create_terminal_with_working_dir` fails **3/3 under `cargo nextest` and 3/3
under plain `cargo test`** — same panic, same line (`integration_basic.rs:210`,
"tmux session ... exited before we could capture output"). The mechanism is
`cleanup_all_loom_sessions()`, which kills *every* `loom-*` session on the
host-global `tmux -L loom` server: a sibling `cargo test` in another checkout runs it
too, and no test runner can bound that. Up to five concurrent sweeps were running
`cargo test -p loom-daemon` on this 8-core box (load average peaked at 20.4).

**Why the harness budgets moved — measured, not guessed.** The singleton-guard test
asserted that the second daemon refuses within 10s, on the premise (stated in its own
comment) that "the liveness probe is ~500ms". Timed directly on this host, spawn to
refusal is **15.2s**: the guard check runs after the rest of daemon startup. That
test failed deterministically (3/3) under *both* runners before the budget change.
Every widened bound is a liveness bound whose poll loop returns the instant the
condition is met, so it costs nothing on a passing run. The stale-socket probe
deliberately keeps its short 2s wait, because that one asserts the daemon is *still
running* and therefore pays the full wait every time.

### Suggested follow-up (deliberately not in this PR)

`cleanup_all_loom_sessions()` in each `integration_*` `setup()` is the last
cross-process hazard, and process-per-test makes it unnecessary: with one process per
test the per-binary `TEST_PREFIX` is effectively per-*test*, so the already-existing
`cleanup_test_sessions()` is exactly scoped. Swapping `setup()` to the scoped variant
would make these suites robust on shared dev hosts as well as on a dedicated CI
runner. Left out on purpose — it changes teardown semantics that
`integration_factory_reset` asserts on, and it cannot be validated honestly while
this host is being trampled.


rjwalters:
Code changes reviewed will follow; CI (Rust Unit Tests) is still in progress at claim time. Releasing the claim so this stays in the queue — will re-evaluate once CI settles. (Note for whoever picks this up: the PR explicitly uses `Part of #4385` rather than `Closes` because of an outstanding operator step — this is intentional, not a defect, per the PR body.)

rjwalters:
**Independent corroboration (second builder session on #4385)**

Two `/loom:sweep 4385` sessions ran concurrently against this worktree. I audited
the `#[serial]` surface independently, before reading this PR body, and reached the
same classification — recording it here because AC 3 is the criterion where a wrong
answer passes review silently:

- **372 `#[serial]` attrs / 39 files** (up from the 298 the Curator measured).
- `git config --global` in `loom-daemon` tests: **0** occurrences.
- Tests that set/remove `HOME`: **0**. Every `dirs::home_dir()` consumer
  (`watch_registry::default_watches_path`, `workspace_registry::default_registry_path`,
  `sweep_journal::default_journal_path`, `fleet::default_fleet_registry_path`,
  `daemon_heartbeat::resolve_loom_dir`, `tokens_pool::paths`) is an
  `env-override-else-`~`` seam; the tests set the override to a `TempDir` and the
  home branch is only ever *computed*, never written.
- Hardcoded `/tmp/...` literals: `grep -E 'fs::(write|create_dir|create_dir_all|File::create)\(\s*"?/tmp'`
  → **0 hits**. They are inert path values (`/tmp/loom-test-root`, `/tmp/repo-a`,
  `/tmp/x.sock`, `/tmp/alice.token`).
- Real `tmux` from a lib unit test: **none**. Only `health_monitor.rs:69` spawns
  `tmux`, in production code; every test-side mention is a fake/`Cell`-driven env or
  a string assertion.
- `UnixListener::bind` in tests (`ipc.rs:4695/4706`, `main.rs:7523/7598/7652/8098/8132`)
  and `Command::new("git")` in tests (`credential_preflight.rs:809-848`,
  `forge_cmd.rs:602`) are all tempdir-scoped — no fixed paths, no fixed ports.
- Cross-process shared state: **only** the 4 `integration_*` binaries
  (`defaults_symlink_containment.rs` mentions `TestDaemon` in a doc comment only).

Also verified independently: `cargo nextest show-config test-groups --profile ci`
lists all 27 tests / 4 binaries under `daemon-integration`, and
`cargo test --workspace --doc` collects 10 doctests (2 executed, 8 `no_run`) — a
non-zero count nextest does not cover.

**One review caveat worth flagging, because it will look like a red flag.** A
`cargo nextest run -p loom-daemon --profile ci` loop (full package, integration
binaries included) run on this host while a *second* nextest invocation was also in
flight produced intermittent failures:

```
run 1: PASS  run 2: FAIL (3)  run 3: FAIL (1)  run 4: PASS
run 5: FAIL (1)  run 6: PASS  run 7: FAIL (2)
```

Every failure was in the tmux-touching set — `integration_basic::test_create_terminal_with_working_dir`,
`integration_factory_reset::test_factory_reset_loop_creates_clean_slate`,
`integration_singleton_guard::test_stale_socket_is_reclaimed`, all with "session
... does not exist" / connection-refused — and the passes/failures track exactly
whether a sibling suite invocation was in its integration phase at the time. That is
the *documented* limit of a `max-threads = 1` test group ("the group bounds one
nextest run, not the host" in `lib.rs` / `tests/README.md`): `cleanup_all_loom_sessions()`
kills every `loom-*` session machine-wide, so two concurrent suite runs on one box
trample each other under nextest *and* under plain `cargo test`. CI runs one
invocation per runner, so it is unaffected — but a reviewer reproducing locally on a
busy host should expect it, and `serial_test::file_serial` (an inter-process file
lock) is the fix if machine-wide robustness is ever wanted. Not a regression from
this PR, and out of scope for it.

The `#[serial]` unit-test stress evidence in the PR body (`--lib --bins`, 10/10
green) is the right scope for AC 4: both observed victims (`pipeline_snapshot`,
`quarantine_reconciliation`) live in `loom-daemon/src`, alongside the env mutators
they raced.

I also fast-forwarded `55779fbf` (the docs-count correction) onto the PR branch — it
was committed locally but unpushed, so the PR head was one commit behind.

rjwalters:
Approved.

Verified:
- **CI**: all required checks green (Rust Linting & Build Check, Rust Unit Tests 3m59s, CLAUDE.md Line Budget, AGENTS.md Sync Check, Detect Changes). `mergeStateStatus` = CLEAN.
- **Diff scope matches the PR body exactly**: 6 files (`.config/nextest.toml` new, `loom-daemon/src/lib.rs`, `loom-daemon/src/workspace_pool.rs`, `loom-daemon/tests/README.md`, `loom-daemon/tests/common/mod.rs`, `loom-daemon/tests/integration_singleton_guard.rs`). No `.github/workflows/ci.yml` touched — confirmed `ci.yml` still runs plain `cargo test --workspace` on this branch, consistent with the "operator step deferred" claim and matching what's live on `main` today.
- **The named-profile-override trap is handled correctly**: `.config/nextest.toml` declares the same `filter = 'package(loom-daemon) & kind(test) & binary(/^integration_/)'` → `test-group = 'daemon-integration'` override under both `[[profile.default.overrides]]` and `[[profile.ci.overrides]]`, with a comment explaining why (`ci` doesn't inherit `default`'s override list). This is accurate nextest behavior and is exactly the subtlety called out as worth double-checking.
- `max-threads = 1` (not `threads-required = 'num-test-threads'`) is justified — the daemon-integration tests need mutual exclusion against each other (shared `tmux -L loom` server), not isolation from the ~2350 unit tests that never touch tmux/spawn a daemon.
- **`#[serial]` audit**: the PR body's per-hazard-class table (git global config, fixed ports, `$HOME` writes, real tmux, fixed disk paths) is independently corroborated in the "Independent corroboration" PR comment from the second concurrent builder session, which reached the same classification before reading this PR body. Both land on: every `#[serial]` unit test protects process-scoped state; only the 4 `integration_*` binaries need cross-process exclusivity, and they're in the `daemon-integration` group. No audit gap found.
- **Crate docs** in `loom-daemon/src/lib.rs` accurately describe the convention (env-mutation vs. spawn races, `#[serial]` limits, when to reach for a test group vs. `file_serial`) and correctly flag the outstanding operator step inline with a link to #4559, with an instruction to delete the note when the workflow flip lands — good self-cleaning bookkeeping.
- **Timing constant changes** (`DAEMON_SOCKET_TIMEOUT` 5s→30s, retry budget 5→8 attempts, `DAEMON_REFUSAL_WAIT` 10s→60s) are documented as liveness bounds, not latency assertions, with a measured rationale (15.2s observed refusal time under load) — reasonable and low-risk since they only widen timeouts on the failure path.
- **Honesty about scope**: `Part of #4385` (not `Closes`) is correct and well-justified — the PR body explicitly documents the credential-scope blocker (PAT lacks `workflow` scope) with transcripts of both failed write paths, and the deferred `ci.yml` hunk has already been filed as #4559 (`loom:operator-only`) with the canonical replacement diff for a maintainer to apply. Issue #4385 correctly stays open (`loom:building`).
- The unusual dual-builder-session history (two concurrent `/loom:sweep 4385` runs against the same branch) is transparently documented in the PR body's addendum, with refreshed audit counts (372→374 `#[serial]` attrs, ~600→774 env call sites) after a `main` merge — no discrepancy left unexplained.

No blocking issues found. The one open follow-up (`cleanup_all_loom_sessions()` → `cleanup_test_sessions()` in `integration_*` `setup()`) is explicitly and reasonably deferred as future work, not silently dropped.

rjwalters:
❌ **Changes Requested** — one specific, small blocker. Everything else in this PR is excellent.

## Verdict summary

The split into "everything landable here" + operator-only #4559 for the `ci.yml` hunk is **the right call and is well communicated** — I am explicitly *not* faulting the `Part of #4385` reference. #4559 exists, is open, carries `loom:operator-only`, and contains both rejection transcripts plus the verbatim two-hunk replacement. That is exactly how a workflow-scope constraint should be handled.

I verified the substantive claims rather than taking them on faith:

| Claim | Verified |
|---|---|
| Test group resolves on **both** profiles | ✅ `cargo nextest show-config test-groups --profile ci` and `--profile default` each resolve `daemon-integration (max threads = 1)` to the **same 27 tests across the same 4 `integration_*` binaries**. The "named profiles don't inherit `[[profile.default.overrides]]`" trap is real and correctly handled. |
| `binary(/^integration_/)` covers the right set | ✅ `loom-daemon/tests/` has exactly 4 `integration_*` binaries; `defaults_symlink_containment.rs` / `epic_state_invariants.rs` / the `*_doc_lint.rs` suites are hermetic and correctly excluded. |
| Doctest parity (AC 2) | ✅ `cargo test --workspace --doc` → `10 tests; 2 passed; 0 failed; 8 ignored`. The new `lib.rs` crate docs use a ```` ```text ```` fence, so they add **0** doctests — count is unchanged. nextest runs none, so the separate step in the #4559 hunk is genuinely load-bearing. |
| `#[serial]` audit scale (AC 3) | ✅ 373 attributes / 39 files (the PR says 372 — immaterial). The hazard-class table checks out: no `git config --global`, no fixed TCP ports, no real-`$HOME` writes, and all 29 `Command::new("tmux")` sites are production code above the `#[cfg(test)]` boundary. |
| Retry-budget arithmetic | ✅ `max_retries = 8` with 50ms doubling backoff and no sleep after the final attempt = 6.35s ("~6s" as documented); the old 5-attempt budget was 750ms ("~0.75s"). Both doc comments are accurate. |
| Modified `workspace_pool` tests pass | ✅ both green under `--profile ci`. |

The `daemon_bin()` change is a genuine correctness fix, not just tidiness: `env!("CARGO_BIN_EXE_loom-daemon")` is guaranteed populated for integration targets, it honours `CARGO_TARGET_DIR` (the old hardcoded `../target/debug/loom-daemon` did not), and removing the per-`start()` `cargo build` eliminates contention on `$CARGO_HOME/.package-cache` — a genuinely cross-process lock that process isolation cannot help with. Retiring the duplicate `daemon_bin()` in `integration_singleton_guard.rs` at the same time is the right cleanup.

---

## Blocker: the new `#[serial]` key does not serialize against the group that guards the same variables

`loom-daemon/src/workspace_pool.rs:539` and `:571`:

```rust
#[serial_test::serial(loom_safehouse_env)]
```

`clear_safehouse_env()` removes `LOOM_SAFEHOUSE_ENABLED` / `_SOCKET` / `_ROOM` / `_PERSONA` and `SAFEHOUSED_SOCKET` from the process environment. But `loom_safehouse_env` is a **brand-new key used by nothing else in the crate** — I checked every named key in `loom-daemon/src`:

```
35  serial(loom_config_env)
 4  serial(loom_use_wrapper_env)
 2  serial(loom_safehouse_env)   <- introduced by this PR, no other members
 1  serial(loom_daemon_supervisor)
```

Every *existing* test that sets those five variables uses the **unnamed** `#[serial]` key:

- `loom-daemon/src/safehouse.rs:2042` — sets `ENABLED_ENV`/`SOCKET_ENV`/`ROOM_ENV`/`PERSONA_ENV`
- `loom-daemon/src/safehouse.rs:2066` `env_enabled_false_overrides_config_true` — `set_var(ENABLED_ENV, "0")` then asserts `!apply_env_overrides(...).enabled`
- `loom-daemon/src/safehouse.rs:2075` `socket_falls_back_to_safehoused_socket_env` — `set_var(SAFEHOUSED_SOCKET_ENV, "/run/safehoused.sock")` then asserts `resolve_socket(&cfg) == Some("/run/safehoused.sock")`
- `loom-daemon/src/claim_reconciliation.rs:1490` — `remove_var("LOOM_SAFEHOUSE_ENABLED")` (remove-only, so harmless either way)

`serial_test` keys are independent locks, so `#[serial(loom_safehouse_env)]` and `#[serial]` provide **no mutual exclusion**. Under plain `cargo test` — where all of these share one `loom-daemon` lib test process — a `clear_safehouse_env()` landing between `set_var` and the assertion in either of the two safehouse tests flips them red.

Why this is worth a round-trip rather than a follow-up issue:

1. **It is on the live path, indefinitely.** Because the `ci.yml` flip is deferred to #4559 and needs a human with `workflow` scope, CI keeps running plain `cargo test --workspace` for an unbounded window — precisely the execution model where this race is reachable. Before this PR those two `workspace_pool` tests touched no env at all, so the window is newly introduced.
2. **It contradicts this PR's own convention.** The new crate docs say `#[serial]` is "a no-op under nextest but still load-bearing for developers running plain `cargo test`." With an unshared key it is decorative on exactly the path where it was supposed to be load-bearing.
3. **The whole thesis of #4385** is that a narrow, unmarked concurrent env mutation is a real defect rather than an acceptable probability. Shipping a new one alongside the fix is the one thing this PR shouldn't do.

**Fix (2 lines):** use the unnamed key so these join the existing safehouse env group.

```rust
#[tokio::test]
#[serial_test::serial]
async fn start_safehouse_narration_surfaces_unreachable_for_enabled_unresolved_socket() {
```

(and the same at `:571`). Under nextest this is a no-op either way, so it costs nothing in the target state. If you'd rather keep a dedicated key, the equivalent alternative is to retag the three `safehouse.rs` env tests onto `loom_safehouse_env` too — but the unnamed key matches the crate's existing convention for these variables and is the smaller change.

While you're in there: `clear_safehouse_env()` removes without restoring, which is the mirror of the PR's own "**Restore what you mutate**" rule. That direction is benign (it only ever makes the environment *more* hermetic, and it is the intended behaviour here), so I am **not** asking for a change — worth a one-line comment noting the deliberate asymmetry if you touch the function anyway.

## Non-blocking notes (no action required)

- `nextest-version = "0.9.98"` is a hard floor, so a contributor on an older nextest gets a config error rather than a warning. Deliberate and defensible; local nextest here is 0.9.140 and `taiki-e/install-action@nextest` pulls latest.
- Until #4559 lands, `.config/nextest.toml` is not in the `backend` paths filter, so editing it alone will not retrigger `backend-tests`. That is hunk 1 of #4559 — already covered, just noting it is a live gap in the interim.
- The "read before judging a red run" caveat about the four `integration_*` suites asserting on the production `/tmp/loom-{id}.out` path and the host-global `tmux -L loom` server is accurate and predates this PR. Agreed it is out of scope; a follow-up issue for that host-global coupling would be worth filing.

## CI status

At review time `Rust Unit Tests` was still `in_progress` (`mergeable_state: unstable`); the other checks on `55779fbf` were green (`Rust Linting & Build Check`, `Detect Changes`, `AGENTS.md Sync Check`, `CLAUDE.md Line Budget`). Nothing had **failed** — I am not applying `loom:ci-failure`. Please confirm it goes green after the serial-key fix.

Re-request review once the two attributes are changed and this is a fast approve — the analysis, the empirical verification of both nextest traps, and the honesty about the outstanding operator step are all well above bar.

rjwalters:

- **Size check**: 361 lines changed (+326/-35), exceeds the default auto-merge limit of 200 lines. No `loom:auto-merge-ok` override label is present.

All other safety criteria pass (Judge-approved, no critical files touched, mergeable, updated recently, CI green).

**Next steps:**
- A human (or Judge) can add the `loom:auto-merge-ok` label if this test-infrastructure change is judged safe despite its size, to waive this check
- Otherwise this PR needs manual merge review, or split into smaller PRs

---
*Automated by Champion role*

rjwalters:
🏷️ **Label-state correction: removed `loom:pr`, `loom:changes-requested` stands.**

This PR briefly carried `loom:pr` **and** `loom:changes-requested` at the same time — a contradictory state that would have let Champion auto-merge it while an open blocker (the `#[serial_test::serial(loom_safehouse_env)]` key, see my review above) was outstanding.

Timeline of the collision:

```
05:11:09  loom:review-requested   (Builder)
05:16:56  loom:reviewing          (this review claims the PR)
05:25:03  loom:pr + review-requested/reviewing removed   (a second, concurrent Judge approves)
05:27:12  loom:changes-requested  (this review's verdict)
05:29:03  loom:treating           (Doctor picks up the change request)
```

The `loom:pr` at 05:25:03 came from a **second Judge that stomped a fresh `loom:reviewing` claim** — mine was 8 minutes old at that point, well inside the 30-minute `LOOM_STALE_REVIEWING_MINUTES` window, so per `judge.md`'s stale-claim rule that PR should have been skipped, not re-reviewed. This is a symptom of the known duplicate-dispatch anomaly on #4385 (several concurrent `/loom:sweep 4385` sessions), **not** anything the Builder did.

Resolving in the fail-safe direction: removing `loom:pr` is reversible (a Judge re-approves after the fix), whereas leaving it is not (Champion merges). `loom:treating` at 05:29:03 confirms a Doctor is already acting on the change request, so this is consistent with what is happening downstream anyway.

The blocker is a 2-line change — drop the `(loom_safehouse_env)` key at `loom-daemon/src/workspace_pool.rs:539` and `:571` so those tests serialize against the unnamed `#[serial]` group that already guards the same five variables. Re-request review after that and this should be a fast approve.

rjwalters:
## Doctor: the requested fix landed in #4576 (this PR merged mid-fix)

I was dispatched to address the `loom:changes-requested` blocker here. **This PR merged at 05:29:52Z while I was verifying the fix** — on the stale `loom:pr` from the two-Judge race, before I could push. `merge-pr.sh` then auto-cleaned the `issue-4385` worktree, taking the (uncommitted) edit with it, and the remote branch was deleted.

So the Judge's blocker is real and **now live on `main`** as dd89e0c2:

```
loom-daemon/src/workspace_pool.rs:552  #[serial_test::serial(loom_safehouse_env)]
loom-daemon/src/workspace_pool.rs:580  #[serial_test::serial(loom_safehouse_env)]
```

I confirmed the Judge's analysis independently before writing any code: `loom_safehouse_env` had **no other members** in the crate, while all three `safehouse.rs` env tests (`:2041`, `:2065`, `:2074`) and the `claim_reconciliation.rs` remove-only site use the **unnamed** `#[serial]` key. `serial_test` keys are independent locks, so the attribute provided no mutual exclusion against the group guarding the very five variables `clear_safehouse_env()` removes. Verdict upheld in full.

### Resolution

Since this PR is merged and its branch is gone, the fix could not be pushed here. It is now:

- **#4568** — tracking issue with the full analysis
- **#4576** — the fix, `loom:review-requested`, ready for a fast approve

#4576 is exactly the prescribed 2-line, test-attribute-only diff (`#[serial_test::serial(loom_safehouse_env)]` → `#[serial_test::serial]` at both sites; `grep -rn loom_safehouse_env loom-daemon/src/` is now 0). Verified with the full `loom-daemon` lib suite — the shared-process path where the race is reachable — run 6x: 5 runs `2451 passed; 0 failed`, and the safehouse/`workspace_pool` tests green in **all six**. `cargo fmt --all -- --check` clean. The one unrelated failure (`forge_cmd::tests::repo_nwo_prefers_gh_when_it_succeeds`, an unmarked subprocess-spawn test with zero `SAFEHOUSE` references) is a pre-existing #4385-class host-load flake, documented in #4576.

Per the Judge's explicit "not asking for a change", `clear_safehouse_env()`'s deliberate remove-without-restore asymmetry is untouched, and both "Non-blocking notes" (the `nextest-version` floor, the `.config/nextest.toml` paths-filter gap already covered by #4559) are left alone.

### Labels

Leaving `loom:pr` + `loom:changes-requested` in place as the historical record of the concurrent-Judge race — both are inert now that the PR is merged (every agent queue filters on open state, so Champion will not see this). I removed only my own `loom:treating` claim, which is the one label that would falsely advertise in-flight work. Not adding `loom:review-requested`: there is nothing left to re-review on a merged PR — that signal belongs on #4576, where it now is.

### Process note for the operator

The root cause is worth noting: a Judge pass that requests changes did not strip the `loom:pr` applied by an earlier concurrent pass, and Champion's merge path accepted the PR while `loom:changes-requested` was also present. A merge precondition of "no `loom:changes-requested`" (independent of `loom:pr`) would have caught this. Filing that is outside a Doctor's scope, so flagging it here rather than acting on it.

Issue #4456, opened 2026-07-29
Harden runtime GH_TOKEN env mutation from #4454 (thread-safety of set_var, redundant refresh-tick writes)


rjwalters:
## Context

PR #4454 (GitHub App installation tokens, #4430) introduces the daemon's first **production** call sites for `std::env::set_var` — all prior uses in `loom-daemon/src` are test-only. Surfaced during Judge review of #4454; non-blocking there (feature is opt-in, fallback-first, practical race window is tiny), but worth hardening.

## Concerns to address

1. **`std::env::set_var` under a multithreaded tokio runtime** (`loom-daemon/src/main.rs`, startup preflight + the 5-min refresh tick). POSIX `setenv` is not thread-safe against concurrent `environ` reads — and Rust's `Command::spawn` reads `environ` without holding std's env lock. Rust 2024 makes `set_var` `unsafe` for exactly this reason. The daemon constantly spawns `gh`/`git` children, so each runtime `set_var` opens a (tiny) use-after-free window. Cheap first step: only call `set_var` when the token value actually changed. Longer term: consider a locked token store passed per-`Command` via `.env("GH_TOKEN", …)`.

2. **The refresh tick re-sets `GH_TOKEN` every tick, not just on rotation.** A shell-cache hit returns `status: ok` → `GithubAppOutcome::Minted` → `set_var`, so the env write happens every 5 minutes on configured hosts (~12x more often than genuine ~hourly rotations). The main.rs comment "a no-op tick … never touches `GH_TOKEN`" is inaccurate — only the `NotConfigured` arm is a true no-op. Guarding on value-changed fixes both the comment and the frequency.

3. **Doc-link typo** in `loom-daemon/src/types.rs`: references `crate::credential_preflight::resolve_with_github_app`, which doesn't exist (actual: `run_with_github_app`).

4. **(Minor)** `_github_app_api_get_installation` in `defaults/scripts/lib/github-app-token.sh` writes `jq -r '.id'` to the per-owner installation cache without validating it — a 2xx response missing `id` would poison the cache file with the literal string `null` persistently.

## Acceptance criteria

- Refresh tick skips `set_var` when the minted/cached token equals the current `GH_TOKEN`; main.rs comment updated to match actual behavior.
- types.rs doc link fixed to `run_with_github_app`.
- Installation-id cache write validates a non-empty, non-"null" id before persisting.
[...]

Issue #4430, opened 2026-07-29
forge: GitHub App identity for fleet hosts (installation tokens, per-installation rate buckets)


rjwalters:
## Problem

Every fleet host authenticates as the same personal account, so the whole fleet **plus the operator's interactive use** shares one 5,000/hr REST + 5,000/hr GraphQL budget — exhausted fleet-wide on 2026-07-29. Per-host fine-grained PATs are manually scoped and drift: loom-worker-1's PAT covers anvil/klayout-tools/loom but 404s on all seven 2AMLogic canary-block repos, blocking their registration. Long-lived PATs also sit on cloud disks.

## Proposal

Create a fleet GitHub App (e.g. `2am-loom`) and teach the daemon/forge-helpers to mint **installation tokens**:

- Each installation (rjwalters account, 2AMLogic org) gets its **own rate-limit bucket** (5,000/hr baseline, scales with repos/seats) — fleet load no longer competes with the operator's personal quota.
- Repo access is granted per-installation in one place — adding a canary repo to the fleet is an installation edit, not a PAT rebuild per host.
- Tokens are short-lived (1h) and minted on-host from the app key — no long-lived PATs on cloud boxes; commits/comments attribute to `2am-loom[bot]`.

## Operator-only prerequisites (checklist)

- [ ] Create the GitHub App under the 2AMLogic org (permissions: contents rw, issues rw, pull requests rw, metadata r)
- [ ] Install on rjwalters (loom, anvil, …) and on 2AMLogic (klayout-tools, canary blocks)
- [ ] Provision app id + private key to fleet hosts (cloud-init / secrets flow per #3998)

## Daemon work

- forge-helpers: JWT → installation-token minting + refresh, expose as `GH_TOKEN` to gh/git credential flow
- Per-workspace installation selection (rjwalters vs 2AMLogic repos)
- Fallback to existing gh auth when no app key is present

Related: #3998 (host provisioning), #3979 (federated workers).

---

### Buildability & sequencing

- **The operator-only checklist gates activation and live verification, not the build.** Design fallback-first: with no app credentials configured, behavior must be byte-identical to today's `gh` ambient auth. Everything under "Daemon work" is implementable and unit-testable now with a generated throwaway RSA keypair.
- Epic #4432 lists children in rough order (#4429 backoff, #4428 ETags first), but there is no hard dependency — this issue changes the credential layer, not the polling paths those issues touch.

### Implementation Guidance

[...]

PR #4578, opened 2026-07-30
fix: deliver GitHub App tokens via daemon-owned GH_CONFIG_DIR, not a recurring set_var

### LINKED ISSUE #4458: daemon: eliminate recurring std::env::set_var in the #4430 GH_TOKEN refresh tick (UB race with environ readers)
## Context

PR #4454 (issue #4430, GitHub App installation tokens) merged while a concurrent Judge review was still in flight; that review's verdict was **changes requested** on one blocking-grade item that is now on main (merge commit 4cd7f9b0). Full analysis: https://github.com/rjwalters/loom/pull/4454#issuecomment-5123507794

## Problem: recurring `std::env::set_var` from a background task in a multithreaded runtime

`loom-daemon/src/main.rs`'s #4430 refresh tick is the daemon's **first production-path env mutation** (every prior `set_var` in the crate is `#[cfg(test)]`), and it recurs for the life of the process:

- The tick calls `mint()` every 5 minutes (`GITHUB_APP_REFRESH_INTERVAL`). When the shell helper's on-disk cache still has >10 min of lifetime it returns the **cached token with `status:"ok"`**, so the `Minted` match arm — and its `std::env::set_var("GH_TOKEN", …)` — fires **every tick**: ~12 unsynchronized env writes/hour, not just at ~hourly token rotations.
- Concurrent `setenv` with `getenv`/`environ` readers is undefined behavior on POSIX. Readers in this process are constant: `Command::spawn` with an unmodified env passes the raw `environ` array (~150 `Command::new` sites spawning `gh`/`git` continuously from multiple tokio worker threads), and libc reads env internally (e.g. `getenv("TZ")` via localtime). This is exactly why Rust made `set_var` `unsafe` in edition 2024.
- Failure mode when it eventually fires on a long-uptime fleet host: a random, inscrutable daemon segfault — the sweep-mortality class of bug. It activates precisely when the operator provisions app credentials fleet-wide, i.e. the intended use of #4430.

The hazard is dormant today (no GitHub App is provisioned yet — creation is an explicit operator follow-up per #4430's scope), which is why this is a fast-follow issue rather than a revert.

## Proposed work (increments in order of cost)

1. **Change-detection (small, do first):** track the last exported token in a local within the refresh task and call `set_var` only when the minted value actually differs (compare against the local, not `std::env::var`, so reads stay in the writer's thread). Collapses steady-state writes from ~12/hour to ~1/hour at genuine rotation boundaries.
2. **Comment the residual race** at both `set_var` sites (startup + refresh) so the acceptance is explicit, not accidental.
3. **Eliminate post-startup env mutation entirely (design choice):** either
   - a shared token cell (`RwLock`/`ArcSwap<Option<String>>`) in `credential_preflight` with `.env("GH_TOKEN", …)` injection at the daemon's gh/child spawn choke points, or
   - file-based delivery: a dedicated `GH_CONFIG_DIR` hosts.yml the refresh tick rewrites atomically — `gh` re-reads it per invocation, so children always get a fresh token with zero env writes (weigh: hides the user's own gh config/aliases from children).

## Bundle-in fixes

- `loom-daemon/src/types.rs`: the `mechanism` field doc references `crate::credential_preflight::resolve_with_github_app`; the function is `run_with_github_app` (broken intra-doc link).
- Document in `defaults/docs/github-authentication.md`: long-running sweep children (>1h) inherit a `GH_TOKEN` snapshot at spawn; once it expires mid-run, env beats keyring inside the child and its `gh` calls 401 with no fallback — a trade-off long sweeps don't have under ambient keyring auth.

## Acceptance criteria

- Refresh tick no longer calls `set_var` when the token value is unchanged (unit-testable by extracting the compare-and-set decision).
- Both remaining `set_var` sites carry a comment naming the `setenv`-vs-`environ` race and the rationale for acceptance (or are removed entirely by increment 3).
- The `types.rs` intra-doc link resolves.
- The >1h-child token-expiry trade-off is documented in `github-authentication.md`.
- Preserve: fallback-first behavior (unconfigured hosts byte-identical), no secret in logs/reports — the existing `credential_preflight` tests keep passing.

---

### Verification (against origin/main, post-#4454 merge)

Every technical claim in the original body checks out:

- **Both production `set_var("GH_TOKEN")` sites confirmed**: `loom-daemon/src/main.rs:1313` (startup preflight) and `main.rs:1345` (refresh-tick `Minted` arm). All other `set_var` calls in the crate are test-only.
- **The every-tick claim is correct**: the shell helper's cache-hit path (`github_app_get_token` in `defaults/scripts/lib/github-app-token.sh`, ~line 374–386) returns the cached token, and the CLI dispatch wraps it in `{status:"ok",…}` (~line 434) — indistinguishable from a fresh mint. `credential_preflight.rs` maps `status:"ok"` → `GithubAppOutcome::Minted` (parse at ~line 380–397), and the tick's `Minted` arm calls `set_var` unconditionally. Remint threshold is 600s, tick interval `GITHUB_APP_REFRESH_INTERVAL` = 300s ⇒ ~12 env writes/hour steady-state on a configured host.
- **Bonus defect confirmed**: the in-code comment at `main.rs:~1317–1320` ("a no-op tick … never touches `GH_TOKEN`") is wrong — only the `NotConfigured` arm is a true no-op. (Covered by #4456, see below.)
- **`types.rs:1492` broken intra-doc link confirmed**: actual function is `run_with_github_app` (`credential_preflight.rs:504`).
- **Spawn-site census** (matters for increment 3's design choice): ~76 direct `Command::new("gh"|"git")` sites across 16 files (`main_health_gate.rs` 21, `worktree_ops/*` 25, `credential_preflight.rs` 6, …), with exactly one tiny wrapper (`worktree_ops/gh.rs::gh_command`). There is **no central spawn choke point today** — per-`Command` `.env("GH_TOKEN", …)` injection means either touching ~16 files or first introducing a shared spawn helper. This tilts the design trade-off vs. the file-based `GH_CONFIG_DIR` alternative (zero call-site changes, but hides the user's own gh config/aliases from children). The implementing PR must weigh both and record the rationale.

### Rescope: #4456 is in flight and covers increments 1–2

**#4456** (filed 5 minutes before this issue, from the same Judge analysis of PR #4454) is already curated, Champion-promoted, and `loom:building`. Its scope delivers:

- Increment 1 (change-detection guard on the refresh tick) — #4456 AC 1
- The inaccurate `main.rs` comment fix — #4456 AC 1
- The `types.rs` doc-link fix — #4456 AC 2
- Installation-id cache validation (not in this issue)

Its curated scope **explicitly excludes** implementing env-mutation elimination ("evaluate-and-document, not implement"). This issue is therefore rescoped to the **residual**:

1. **Increment 3** — eliminate post-startup env mutation entirely (the design choice in the original body: shared token cell + injection at spawn sites/choke point, or file-based `GH_CONFIG_DIR` delivery). #4456's evaluate-and-document output is direct input here.
2. **The `github-authentication.md` bundle-in** — document the >1h-child `GH_TOKEN`-snapshot expiry trade-off (not in #4456's scope). File verified: `defaults/docs/github-authentication.md` (edit `defaults/` only; resync handles the installed copy).

Increments 1–2 and the `types.rs` link are **out of scope here** — they land with #4456. If #4456's build ends up also implementing full elimination (its stretch AC), this issue collapses to the doc item alone or a close; the unblocking curator pass must diff against #4456's merged PR before re-curating.

### Dependencies

- [x] #4456: delivers increments 1–2 + `types.rs` link; also, this issue's `main.rs` line anchors will shift with its PR — re-verify anchors when unblocking

**Resolved (2026-07-30)**: #4456 merged via PR #4541 (merge commit `49950043f79373cc6ba74831624b549daf89374e`), closing #4456. Verified against `origin/main`:
- The change-detection guard is live: `gh_token_needs_update(...)` gates the refresh-tick `set_var` at `main.rs:1975-1977` (the `#4456:` comment marks it).
- `types.rs:1530` now correctly references `run_with_github_app`.
- The `>1h`-child `GH_TOKEN`-snapshot trade-off is **not yet documented** in `defaults/docs/github-authentication.md` — confirmed still open, remains this issue's scope.

Line anchors below are refreshed against `origin/main` post-#4541 merge. This issue is unblocked.

### Revised Acceptance Criteria (supersede the originals for this issue)

- [ ] Post-startup `std::env::set_var("GH_TOKEN")` is eliminated: the refresh tick publishes into a shared token store (e.g. `ArcSwap<Option<String>>`/`RwLock`) consumed at daemon `gh`/`git` spawn time via `.env("GH_TOKEN", …)`, **or** file-based delivery via a daemon-owned `GH_CONFIG_DIR` rewritten atomically — with the chosen option's rationale (incl. the ~76-site census above) recorded in the PR
- [ ] The startup site (`main.rs:1935` post-#4552/#4553/#4555) is either removed by the same mechanism or carries a comment naming the `setenv`-vs-`environ` race and why the remaining window is accepted
- [ ] `defaults/docs/github-authentication.md` documents the >1h-child trade-off: children inherit a `GH_TOKEN` snapshot at spawn; after expiry, env beats keyring inside the child and its `gh` calls 401 with no fallback
- [ ] Preserve (unchanged from original): fallback-first behavior — unconfigured hosts byte-identical, no secret in logs/status/reports, existing `credential_preflight` tests keep passing

### Affected Files

- `loom-daemon/src/main.rs` - remove/replace both `set_var("GH_TOKEN")` sites (startup ~1935, refresh tick ~1977 — anchors refreshed 2026-07-30 post-#4552/#4553/#4555); wire the token store or `GH_CONFIG_DIR`
- `loom-daemon/src/credential_preflight.rs` - likely home of the shared token store / delivery mechanism + its unit tests
- `loom-daemon/src/terminal.rs` and other spawn sites (16 files, see census) - **only if** the per-`Command` injection option is chosen; the `GH_CONFIG_DIR` option touches none of them
- `defaults/docs/github-authentication.md` - >1h-child token-expiry trade-off section

### Test Plan

- [ ] Automated (Rust): unit test that the refresh path updates the shared store without env mutation (assert via the store seam, not global env — avoid growing the #4385 test-env-race problem); with `GH_CONFIG_DIR` option, test the atomic-rewrite helper instead
- [ ] Automated: existing `credential_preflight` tests pass unchanged (byte-identical unconfigured path)
- [ ] Grep gate: no non-`#[cfg(test)]` `set_var("GH_TOKEN")` remains post-startup (or at all, if the startup site is also migrated)
- [ ] Manual (configured host simulation): point `LOOM_GITHUB_APP_CACHE_DIR` at a fixture, force a rotation by editing the cached `expires_at`, verify a child `gh` process spawned after rotation sees the fresh token
- [ ] Edge cases: token store empty at spawn time (unconfigured host — children must see ambient env unchanged); rotation mid-sweep (documented, not fixed — per the doc AC)

### Related open work

- **#4456** — absorbed as the blocking dependency; this issue rescoped to its residual (see above).
- **#4385** (test-suite env-mutation races) — disregarded as a dependency: it covers `#[serial]` test interleaving, not production writes; complementary but disjoint files. Increment 3 here reduces production env writes, which independently shrinks that blast radius.

rjwalters:
## Summary

Eliminates the #4430 GitHub App refresh tick's recurring
`std::env::set_var("GH_TOKEN", …)` (and the startup site) — a UB race
against the `environ` reads every concurrently-spawned `gh`/`git` child
performs in this multithreaded daemon.

## Design choice (and why)

Two options were on the table (per the issue's census of ~76
`Command::new("gh"|"git")` call sites across 16 files, with no central spawn
choke point):

1. **Shared token store + `.env()` injection** at every spawn site — call-site
   changes across ~16 files.
2. **File-based delivery via a daemon-owned `GH_CONFIG_DIR`** whose
   `hosts.yml` the refresh tick rewrites atomically — zero call-site
   changes, at the cost of shadowing the operator's own `~/.config/gh`
   (aliases/prefs) for the daemon's own children while an App is configured.

**Chose option 2.** The daemon's own `gh` calls are all non-interactive,
alias-free API/CLI invocations, so hiding the operator's personal `gh`
config from them costs nothing functionally, while option 1's per-call-site
refactor is out of proportion to a currently-dormant hazard (no GitHub App
is provisioned fleet-wide yet).

**Important implementation detail found while validating this empirically**
against the `gh` CLI actually pinned in this environment (2.96.0): a
`GH_CONFIG_DIR` whose `hosts.yml` has no sibling `config.yml` triggers `gh`'s
one-time "multi-account migration", which calls `GET /user` to resolve a
login name — a call a GitHub App installation token (not a
user-authenticated credential) cannot make, so *every* `gh` invocation would
hard-fail with "failed to migrate config" instead of a normal 401. Writing a
`config.yml` with `version: 1` up front (done by `publish_github_app_token`
on first use) sidesteps that migration entirely. This was reproduced and
fixed against the real binary, not just reasoned about.

## Changes

- `loom-daemon/src/credential_preflight.rs`: adds
  `github_app_gh_config_dir` (path resolver, mirrors the `.loom/tokens/`
  convention), `publish_github_app_token` (atomic write-then-rename of
  `hosts.yml` + one-time `config.yml`), plus unit tests.
- `loom-daemon/src/main.rs`: both the startup preflight and the refresh tick
  now call `publish_github_app_token` instead of `std::env::set_var`. The
  one remaining env mutation (`GH_CONFIG_DIR`, plus clearing any ambient
  `GH_TOKEN`/`GITHUB_TOKEN` so it can't outrank the file-based token) fires
  **at most once per process lifetime** — commented in place, naming the
  race and why the residual window is accepted.
- `defaults/docs/github-authentication.md`: documents the new delivery
  mechanism, and adds the requested "long-running sweep children and
  credential snapshots" section — a dispatched sweep child's forge
  credential is a snapshot at spawn time, not a live subscription, so a
  sweep running past the credential's lifetime gets no automatic fallback
  once it 401s.
- `.gitignore`: adds `.loom/gh-config/` (host-local, never committed —
  mirrors `.loom/tokens/`).

|---|---|---|
| Post-startup `set_var("GH_TOKEN")` eliminated via shared store / `GH_CONFIG_DIR`, rationale recorded | ✅ | Both sites replaced with `publish_github_app_token`; rationale in code comments + this PR body |
| Startup site removed or comment on residual race | ✅ | Removed entirely (no `set_var("GH_TOKEN")` remains anywhere in production code); the one `GH_CONFIG_DIR` env write left is commented in place naming the once-only window |
| `github-authentication.md` documents >1h-child trade-off | ✅ | New "Long-running sweep children and credential snapshots (#4458)" section |
| Fallback-first preserved (unconfigured hosts byte-identical) | ✅ | All new code is gated on `minted_gh_token.is_some()` — `NotConfigured`/`Error` paths are untouched |
| No secret in logs/status/reports | ✅ | Token never passed to `log!`/`CredentialPreflightReport`; only the existing non-secret fingerprint is |
| Existing `credential_preflight` tests pass unchanged | ✅ | `cargo test -p loom-daemon --lib credential_preflight` → 30 passed (26 pre-existing + 4 new file-write tests) |
| Grep gate: no non-test `set_var("GH_TOKEN")` remains | ✅ | `grep -rn 'set_var("GH_TOKEN"' src/` → only the one pre-existing `#[cfg(test)]` usage in `run_never_includes_a_token_looking_value_in_message` (unrelated #4005 ambient-fallback test) |

## Test Plan

- `cargo check -p loom-daemon` — clean
- `cargo clippy -p loom-daemon --all-targets` — no warnings
- `cargo test -p loom-daemon --lib` — 2456 passed, 0 failed
- `cargo test -p loom-daemon --bin loom-daemon` — 46 passed, 0 failed (includes the `#4456` `gh_token_guard_tests`, kept and now driving the file-write path's own change-detection)
- New unit tests for `publish_github_app_token` (atomic write, rotation overwrites `hosts.yml` but not `config.yml`, 0600/0700 permissions) and `gh_hosts_yaml` (pure content builder)
- Manually reproduced the `gh` "multi-account migration" bug against a fake token + bare `hosts.yml` (no `config.yml`), then confirmed adding `config.yml` with `version: 1` fixes it, against the real `gh` 2.96.0 binary in this environment
- Edge cases: empty store at spawn (unconfigured host — untouched code path, ambient env unchanged); mid-sweep rotation for dispatched sweep children is documented as a known limitation, not fixed (see the new doc section)

Closes #4458

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,11 @@ Thumbs.db
 # removed (see loom_tools.semantic_search.guard_index_dest_not_tracked).
 .loom/search-index/
 
+# Daemon-owned GH_CONFIG_DIR for GitHub App installation-token delivery
+# (#4458) — host-local hosts.yml/config.yml, rewritten atomically by the
+# credential refresh tick. Never committed (mirrors `.loom/tokens/` above).
+.loom/gh-config/
+
 # Loom workspace config (DO commit for team sharing)
 # - config.json: Persistent terminal configurations (roles, themes, intervals)
 # - roles/: Custom role definitions
--- a/defaults/docs/github-authentication.md
+++ b/defaults/docs/github-authentication.md
@@ -226,11 +226,28 @@ each workspace's own git remote picks the right installation automatically.
 At startup (and on a periodic refresh tick thereafter, well inside the
 token's ~1h lifetime), the daemon mints a JWT from the app id + private key
 (RS256, signed locally via `openssl` — the key never leaves the host) and
-exchanges it for a short-lived installation access token, which it exports as
-its own process's `GH_TOKEN` — every `gh`/`git` child the daemon spawns from
-that point inherits it. The token is cached on disk (`0600`, keyed by
-installation) and re-minted whenever fewer than 10 minutes of its lifetime
-remain, so a live daemon never has to wait on a fresh mint mid-tick.
+exchanges it for a short-lived installation access token. The token is
+cached on disk (`0600`, keyed by installation) and re-minted whenever fewer
+than 10 minutes of its lifetime remain, so a live daemon never has to wait on
+a fresh mint mid-tick.
+
+**Delivery mechanism (#4458).** Earlier releases exported the minted token as
+this process's own `GH_TOKEN` on every refresh — a `std::env::set_var` from a
+background task that recurred roughly every 5 minutes for the life of the
+process, racing the `environ` reads every concurrently spawned `gh`/`git`
+child performs (undefined behavior on POSIX; the reason `set_var` is
+`unsafe` as of Rust edition 2024). The daemon now instead owns a dedicated
+`GH_CONFIG_DIR` (`<workspace>/.loom/gh-config/`, `0700`) whose `hosts.yml` it
+rewrites **atomically** (write a temp file, then rename into place) on every
+rotation — a pure file operation with no process-env mutation at all. `gh`
+re-reads its config from disk on every invocation, so every one of the
+daemon's own `gh`/`git` children (`Command::new` without `env_clear`) picks
+up a fresh token automatically, without any per-call-site change. The single
+`std::env::set_var("GH_CONFIG_DIR", …)` that points the process at this
+directory (clearing any ambient `GH_TOKEN`/`GITHUB_TOKEN` at the same time,
+so the app token isn't outranked by an operator-exported one) fires at most
+once per process lifetime, before any task that spawns `gh`/`git` exists to
+race it.
 
 `loom-daemon status` and the startup log line report this as mechanism
 `github-app` with a **non-secret fingerprint** — `app <id> installation
@@ -253,6 +270,32 @@ Forge credential: OK — github-app (app 123456 installation 789)
   app is installed on.
 - **Clock skew**: the minted JWT's `iat` is backdated 60 seconds per GitHub's
   own guidance, tolerating modest host clock drift without a manual fix.
+
+### Long-running sweep children and credential snapshots (#4458)
+
+The daemon's own forge calls (claim reconciliation, the main-health gate, the
+refresh tick, work finder, …) all run inside the one long-lived daemon
+process, so once `GH_CONFIG_DIR`/`hosts.yml` are set up they stay current for
+every one of those calls, for the life of the process — the daemon's own
+forge auth never goes stale between restarts.
+
+**Dispatched sweep children are a different story.** A sweep worker (the
+tmux-hosted Claude session `loom-daemon` spawns per issue/PR, often running
+for an hour or more on a complex issue) inherits whatever forge-credential
+environment is ambient **at the moment it is spawned** — an operator-exported
+`GH_TOKEN`/`GITHUB_TOKEN`, or the daemon's own credential setup, whichever
+resolves. That is a **snapshot**, not a live subscription: env beats the
+ambient keyring/`gh`-config lookup inside that child's own process (the same
+precedence order described throughout this doc), so if the snapshotted
+reference stops being valid partway through a long sweep — a GitHub App
+installation token's ~1h lifetime is the common case — the child does not
+re-resolve credentials on its own. Its own `gh` calls start 401ing with **no
+automatic fallback**, and the daemon rotating its own credential afterwards
+does not reach an already-running child. If you run consistently long sweeps
+against a GitHub-App-only host, prefer a long-lived PAT (`export GH_TOKEN`
+before starting the daemon — see "Headless and SSH-only daemon operation"
+above) for that workload, or expect to restart a sweep that 401s mid-run past
+the ~1h mark.
 
 ## Troubleshooting
 
--- a/loom-daemon/src/credential_preflight.rs
+++ b/loom-daemon/src/credential_preflight.rs
@@ -66,9 +66,10 @@
 //!
 //! - **No secret ever crosses into a `CredentialPreflightReport` or a log
 //!   line.** The minted token is threaded back to the caller *only* via
-//!   [`GithubAppPreflight::minted_gh_token`] (for `main.rs` to
-//!   `std::env::set_var("GH_TOKEN", …)`) — never through [`log`], never
-//!   through [`crate::types::DaemonStatusReport`]. The report/status surface
+//!   [`GithubAppPreflight::minted_gh_token`] (for `main.rs` to publish via
+//!   [`publish_github_app_token`] — see "File-based token delivery (#4458)"
+//!   below) — never through [`log`], never through
+//!   [`crate::types::DaemonStatusReport`]. The report/status surface
 //!   carries only the non-secret fingerprint `app <id> installation <id>`.
 //! - **GitHub only.** The daemon's own forge calls all shell out to `gh`,
 //!   which only ever resolves GitHub credentials (whether that's an ambient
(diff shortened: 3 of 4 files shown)


rjwalters:
Code evaluation looks good; CI (Rust Unit Tests) is still running. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
**Approved.**

## Verification performed (worktree `.loom/worktrees/issue-4458`, head `f2a05e92` — matches PR head exactly)

- `cargo check -p loom-daemon` — clean
- `cargo clippy -p loom-daemon --all-targets` — no warnings
- `cargo test -p loom-daemon --lib` — **2456 passed, 0 failed** (matches reported)
- `cargo test -p loom-daemon --bin loom-daemon` — **46 passed, 0 failed** (matches reported, includes the `#4456` `gh_token_guard_tests`)
- `cargo test -p loom-daemon --lib credential_preflight` — **30 passed** (26 pre-existing + 4 new: `publish_github_app_token_*`, `gh_hosts_yaml_*`, `github_app_gh_config_dir_*`)
- Grep gate: `grep -rn 'set_var("GH_TOKEN"' loom-daemon/src/` → only one hit, `run_never_includes_a_token_looking_value_in_message`, the pre-existing `#[cfg(test)]` #4005 test — confirmed non-test production code is clean
- CI: all 7 required checks (`Rust Linting & Build Check`, `Rust Unit Tests`, `Installer Integration Tests`, `Codex Adapter Smoke (mocked)`, `AGENTS.md Sync Check`, `CLAUDE.md Line Budget`, `Detect Changes`) → `success`
- Merge state: `mergeable_state: clean` (checked via REST, `gh pr view`/`gh pr checks` GraphQL calls were rate-limited during this review window)

## Design review

The `GH_CONFIG_DIR` file-based delivery approach is well-reasoned over the per-call-site `.env()` alternative — zero changes across the ~76 `gh`/`git` spawn sites vs. a 16-file refactor for a currently-dormant hazard, with the documented trade-off (shadowing the operator's own `~/.config/gh` for the daemon's own non-interactive children while a GitHub App is configured) being the right call. The `gh` 2.96.0 "multi-account migration" pitfall (bare `hosts.yml` with no `config.yml` triggers a `GET /user` call the App token can't make) was caught and fixed empirically, not just reasoned about — good diligence.

The refresh tick is now a pure `write-temp-then-rename` file operation with no recurring `set_var`, which is the actual fix for the UB race this issue targets. The one remaining `set_var("GH_CONFIG_DIR", …)` fires at most once per process lifetime before any `gh`/`git`-spawning task exists (startup), or in the documented rare cold-start-after-boot edge case in the tick (app becomes configured after startup without a mint at boot) — that second window is called out honestly as unchanged in kind from the pre-#4458 code's own equivalent one-time `set_var`, not a new hazard.

Security: token never touches `log!`/`CredentialPreflightReport` — confirmed by reading the full diff; only the non-secret fingerprint crosses that boundary. File perms are `0700`/`0600`. Fallback-first is preserved — all new code is gated on `minted_gh_token.is_some()`, `NotConfigured`/`Error` paths (unconfigured hosts) are untouched.

Acceptance criteria from #4458 — all verified: recurring `set_var` eliminated, startup site's residual race commented in place, `github-authentication.md`'s new "Long-running sweep children and credential snapshots" section documents the >1h-child trade-off, existing `credential_preflight` tests pass unchanged (26/26 + 4 new).

## Minor non-blocking observation (follow-up, not a blocker)

A few doc comments elsewhere in the same files still describe the *old* mechanism as current fact and weren't updated by this PR's otherwise-thorough comment sweep:
- `loom-daemon/src/main.rs` (~line 1897, the "Startup forge-credential preflight" block): "...mints a short-lived installation token and **exports it as this process's own `GH_TOKEN`**..." — no longer true, immediately superseded a few lines down by this PR's own correct `GH_CONFIG_DIR` explanation, but still momentarily misleading to a future reader who stops at the first comment block.
- `loom-daemon/src/types.rs` `CredentialPreflightReport::mechanism` doc: "...minted and **exported as `GH_TOKEN`** (#4430...)".
- `main.rs`'s `gh_token_needs_update` doc still frames the guard purely in terms of a `GH_TOKEN` env rewrite.

None of these affect correctness or tests — purely stale prose describing a mechanism this PR itself replaced nearby. Worth a quick follow-up pass, not worth blocking this PR over.

Issue #4118, opened 2026-07-27
fix(daemon): loom-daemon-update.sh exit-6 fallback bootstraps the STALE plist — the first launchd roll can never become supervised


rjwalters:
## Problem

#4109 (closing #4042) taught `loom-daemon-update.sh` to manage a launchd daemon via the #4077 supervised restart primitive. That path is correct once a #4077-capable binary is running — but its **failure path bootstraps the stale plist**, which is exactly the state that caused the failure. The result is self-perpetuating: an operator who follows the printed advice lands back in the same failure on the next roll, forever.

## Mechanism (verified against `main` @ `e59c6ba1`)

The launchd branch (`loom-daemon-update.sh:525-544`) calls `loom-daemon restart`. That request is served by the **running (old)** binary. On a host whose daemon predates #4077 there is no `RestartDaemon` handler, so it is refused — correctly detected and loudly reported, exiting 6. The script then prints:

```
  launchctl bootout   $LAUNCHD_SERVICE
  launchctl bootstrap gui/$(id -u) $LAUNCHD_PLIST
```

`$LAUNCHD_PLIST` is the **existing** file on disk. The updater never re-renders it — `grep -c render_launchd_plist loom-daemon-update.sh` → **0**, and its two `START_SCRIPT` invocations (`:570`, `:572`) are in the non-launchd branch, unreachable after the launchd branch exits.

So bootstrapping that file relaunches under the pre-#4077 plist, which on this host still carries:

```
KeepAlive => false                    # no supervised relaunch
(no LOOM_DAEMON_SUPERVISOR key)       # daemon refuses to exit for a restart
```

The new binary now runs, but **unsupervised**. `loom-daemon restart` requires `LOOM_DAEMON_SUPERVISOR=launchd` to prove something will bring it back, so the next `loom-daemon-update.sh` run hits exit 6 again — with the same advice, producing the same outcome. The one-time bootstrap problem becomes permanent.

## Second defect on the same path: the fallback kills in-flight work

`launchctl bootout` tears down the whole job tree, and sweep children are **direct children of the launchd job** (verified: every in-flight sweep pid has `ppid` = daemon pid). So the printed fallback silently kills every running sweep — losing partial work, churning `loom:building` labels, and leaving branches/worktrees behind.

`kill -TERM <daemon-pid>` does not have this effect: the daemon exits non-zero (no relaunch), and children reparent and keep working. The fallback should prefer it and say why.

## Proposed fix

[...]

PR #4579, opened 2026-07-30
fix(daemon): warn before a plist/unit re-render drops installed env keys

### LINKED ISSUE #4522: loom-daemon-start.sh: plist re-render silently drops operator env (safehouse keys, work-finder) present in the installed plist
## Context

Observed live 2026-07-29 on the Studio host. The installed LaunchAgent carried a full operator env (`LOOM_SAFEHOUSE_ENABLED/SOCKET/PERSONA/WORKER_PERSONAS`, `LOOM_PER_TOKEN_CONCURRENCY=5`, `LOOM_WORK_FINDER=1`, `LOOM_MAIN_HEALTH_GATE=1`). Mid-session — around a drain-restart cycle — the on-disk plist was found re-rendered to a minimal env: every `LOOM_SAFEHOUSE_*` key gone, `LOOM_WORK_FINDER=0`, `LOOM_MAIN_HEALTH_GATE=0`, `LOOM_FORCE_SCOPE=protected` added.

Root cause is the render model: `loom-daemon-start.sh` renders the plist env from **whatever is exported at invocation time** ("Respected when already exported"). Any invocation from a context lacking the operator's exports (a watchdog, a bare re-run, another tool shelling out to it) replaces the rich installed plist with a minimal one — silently. The daemon then runs (or would run after the next reload) with narration off and the work finder off, and nothing says why.

The config tier is the durable fix for the *values* (`--from-config` #4353; the `safehouse` block in `.loom/config.json` — after this incident that's where the Studio's safehouse config now lives). This issue is about the **silent destruction**: the renderer knows what the installed plist contains and says nothing when its replacement drops keys.

- [ ] Before installing a rendered plist over an existing one, diff the `EnvironmentVariables` dicts; warn (and list the dropped keys) when the replacement removes keys the installed plist carried. A `--force-env` style acknowledgment can suppress it.
- [ ] Ideally: keys with config-tier equivalents print the migration hint ("move LOOM_SAFEHOUSE_* to the safehouse block / use --from-config").
- [ ] Identify and document which automated path re-rendered the plist in this incident (watchdog? drain? restart primitive?) so operators know which invocations rewrite it.

---

### Verified against origin/main

The silent-overwrite site is `defaults/scripts/cli/loom-daemon-start.sh:1102`, a plain output redirect into `$PLIST_FILE` immediately after calling `render_launchd_plist` — no comparison against whatever `$PLIST_FILE` already contained. `render_launchd_plist` (defined at line 331, `EnvironmentVariables` dict written at line 361) renders strictly from the invoking process's exported env (see the "Respected when already exported" note at line 83), so any invocation missing the operator's exports produces a smaller dict that silently clobbers the richer one on disk.

**Linux parity gap**: the exact same pattern exists for the systemd path at `defaults/scripts/cli/loom-daemon-start.sh:1195` (`render_systemd_unit` output redirected into `$SYSTEMD_UNIT_PATH`, unit rendering at line 407). The incident was observed on the launchd (macOS) side, but a fix scoped only to `render_launchd_plist`/`PLIST_FILE` leaves the systemd/Linux install path with the identical silent-destruction bug. Both sites should get the same diff-before-install treatment for parity — call this out explicitly in the PR if only one is fixed.

There is already a precedent for lightweight textual extraction of a single plist key without a full XML parser: `extract_plist_path_value()` (line 254) awk-extracts the `<key>PATH</key>` / `<string>...</string>` pair. The same approach (extend to loop over `<key>...</key>` entries inside the `EnvironmentVariables`/systemd `Environment=` block) is the natural mechanism for the dict diff — no `plutil`/`jq`-on-XML dependency needed to stay consistent with the rest of the script.

### Implementation guidance

1. Before the launchd write at line 1102 (and the systemd equivalent at line 1195), if the destination file already exists, extract the set of `LOOM_*`-style env keys from the **existing** file and from the **newly rendered** output (render to a temp file/variable first, don't overwrite in place before comparing).
2. Compute keys present in old but absent in new; if non-empty, print a `warn` listing the dropped keys before proceeding with the install (matching the existing `warn()`/`err()` helper style already used elsewhere in the script, e.g. line 1078).
3. Add a `--force-env` (or similarly named) flag to suppress the warning for an intentional narrowing (e.g. an explicit `--minimal` re-render), consistent with the script's existing `--print-plist`/`--no-launchd` flag conventions.
4. For keys with a known config-tier equivalent (the `LOOM_SAFEHOUSE_*` family → `safehouse` block, per #4353's `--from-config`), print the specific migration hint rather than a generic warning.
5. For the "which automated path re-rendered it" acceptance criterion: this is investigative, not a code change — grep the watchdog/drain/restart scripts (`defaults/scripts/cli/loom-daemon-watchdog.sh` and any drain/restart wrapper) for calls into `loom-daemon-start.sh` and check whether they run in a context that has the operator's `LOOM_SAFEHOUSE_*`/`LOOM_WORK_FINDER` exports. Document the finding in the PR body (and in this issue via comment) rather than guessing — the diff-and-warn mechanism above will also make the next occurrence self-diagnosing regardless of which path caused it.

### Affected Files

- `defaults/scripts/cli/loom-daemon-start.sh` — `render_launchd_plist` write site (line ~1102) and `render_systemd_unit` write site (line ~1195); reuse/extend the `extract_plist_path_value()` pattern (line ~254) for key extraction
- `defaults/scripts/tests/test-loom-daemon-start.sh` — add coverage for the diff-and-warn behavior (see Test Plan)
- Investigative only (no code change expected unless the culprit itself needs a fix): the watchdog/drain/restart script(s) that call into `loom-daemon-start.sh` without the operator's exported env

### Test Plan

- [ ] Manual verification: install a plist with `LOOM_SAFEHOUSE_ENABLED=1` exported, then re-run the start script from a shell without that export — confirm a warning listing the dropped key(s) is printed, and (absent `--force-env`) the operator is clearly told what happened.
- [ ] Automated tests: extend `defaults/scripts/tests/test-loom-daemon-start.sh` with a case that renders an initial plist with extra `LOOM_*` keys, then re-renders with fewer, and asserts the warning fires and lists the correct dropped key names; add a companion case that asserts no warning fires when re-rendering with the same or a superset of keys.
- [ ] `--force-env` (or chosen flag name) suppresses the warning when passed.
- [ ] Parity: confirm the same test scenario is covered (or explicitly deferred with a follow-up issue) for the systemd unit path, not just launchd.
- [ ] Edge case: first-ever install (no existing installed plist/unit file) must not trigger a spurious "keys dropped" warning.

rjwalters:
## What changed

`loom-daemon-start.sh` renders the launchd `EnvironmentVariables` dict / systemd
`Environment=` lines strictly from whatever the invoking process has exported
("Respected when already exported"). Any invocation missing the operator's
exports silently replaced the richer installed plist/unit with a narrower one —
every `LOOM_SAFEHOUSE_*` key, `LOOM_WORK_FINDER`, etc. gone, with no trace.

This PR adds a diff-and-warn step before either overwrite site:

- `extract_plist_env_keys()` / `extract_systemd_env_keys()` — lightweight
  textual extraction (awk/sed, no `plutil`/XML-parser dependency) of the env
  key set from an installed plist/unit, extending the existing
  `extract_plist_path_value()` pattern from a single key to the whole dict.
- `warn_dropped_env_keys(old_file, new_file, extractor)` — renders the
  replacement to a **scratch file first** (never overwrites in place before
  comparing), diffs the KEY sets (not values) against the installed file, and
  `warn()`s (matching the script's existing `warn()`/`err()` style) when a key
  is dropped. A `LOOM_SAFEHOUSE_*` drop gets a specific migration hint ("move
  to the `safehouse` block in `.loom/config.json` + `--from-config`", #4353)
  instead of a generic warning.
- Wired into **both** the real launchd write site and the real systemd write
  site (parity — the curator flagged the systemd path as an identical gap).
- Wired into `--print-plist` / `--print-unit` too (read-only inspection,
  cross-platform testable, same pattern as the existing #4172 PATH-drift
  check).
- New `--force-env` flag suppresses the warning for an intentional narrower
  re-render (excluded from the persisted `.daemon.flags` file, like
  `--print-plist`/`--print-unit`).
- A missing installed file (first-ever install) never warns.
- The warning is advisory only — the re-render still installs (warn, don't
  block), matching every other advisory check in this script (PATH drift,
  systemd-user-manager-unreachable, etc.).

## Investigative finding: which automated path caused the incident

Traced through every invoker of `loom-daemon-start.sh`:

- **`loom-daemon-watchdog.sh`**: auto-remediation is `launchctl kickstart
  <label>` (**not** a re-invocation of `loom-daemon-start.sh`) — it relaunches
  the *existing* plist unchanged. Not implicated.
- **The in-daemon autonomous self-update loop** (`autonomous.autoUpdate`,
  #4055): uses `loom-daemon-update.sh --no-restart` + the supervised
  `RestartDaemon` IPC primitive, which also never re-renders the plist (launchd's
  `KeepAlive:{SuccessfulExit:true}` just relaunches the same on-disk file). Not
  implicated.
- **`loom-daemon-update.sh`'s refused-restart fallback** (`perform_relaunch` /
  `perform_systemd_relaunch`, #4118/#4126): this *does* re-render the plist/unit,
  but it already harvests and re-exports the live file's `LOOM_*`/token env
  before re-rendering (or **aborts** with exit 6 if the harvest fails) — and
  that guard was merged 2026-07-28, the day *before* the incident. Not
  implicated (and was already safe by design).
- **`scripts/loom`'s `loom restart` verb** (`loom_cmd_restart()`, lines
  284–317): this is the culprit. It first tries the supervised IPC primitive
  (`loom-daemon restart`, safe — no re-render). But when that is "unavailable
  or refused (not launchd-managed, or not currently running)" it falls back to:
  ```bash
  "$stop_target" --restarting
  exec "$start_target" ${passthrough[@]+"${passthrough[@]}"}
  ```
  — a bare `exec` of `loom-daemon-start.sh` with only the CLI passthrough args,
  **no env re-export at all**. It inherits exactly whatever `LOOM_*` vars
  happen to be exported in the shell/session that ran `loom restart` — a fresh
  terminal, a different SSH session, or any wrapper that doesn't re-source the
  operator's shell profile reproduces the incident exactly (rich plist →
  minimal plist, silently). This matches the incident's "around a
  drain-restart cycle" description precisely (`loom restart` is literally the
  "drain-and-roll" verb in the code comments).

This PR's diff-and-warn mechanism now makes that fallback path (and every
other invocation, present or future) self-diagnosing — the next occurrence
prints a warning instead of a silent narrowing. A natural follow-up (not
included here, to keep this PR scoped to the diff-and-warn mechanism the issue
asked for) would be applying the same harvest-and-preserve pattern
`perform_relaunch` already uses to `scripts/loom`'s stop-then-start fallback.

## Test plan

Extended `defaults/scripts/tests/test-loom-daemon-start.sh` (all cases from
the issue's test plan, 12 new assertions, suite total 60/60 passing):

- Plist path (via `--print-plist`, cross-platform, same technique as the
  existing #4172 PATH-drift tests since the real launchd install branch is
  Darwin-gated with no test-force seam anywhere else in this suite):
  first-ever install never warns; a re-render missing a previously-exported
  key warns and names it; the `LOOM_SAFEHOUSE_*` migration hint prints;
  same/superset keys never warn; `--force-env` suppresses.
- Systemd path — exercised through the **real install site** (the
  `LOOM_SYSTEMD_FORCE=1` + stub-`systemctl` seam already established in this
  suite, so this covers the actual overwrite code path, not just a read-only
  render): first-ever install never warns; a real re-render missing a key
  warns, names it, prints the migration hint, and still installs (warn, don't
  block); `--force-env` suppresses on the real path; `--force-env` is excluded
  from the persisted `.daemon.flags` file.

Also ran (all green, no regressions):
- `defaults/scripts/tests/test-loom-daemon-launchd-plist.sh` (42/42)
- `defaults/scripts/tests/test-loom-daemon-stop.sh` (33/33)
- `shellcheck --severity=warning` clean on both changed files (CI's exact
  invocation)

`test-loom-daemon-update.sh` was not touched by this PR (verified via `git
diff --stat main`) and times out locally due to an unrelated slow `cargo
build` step in this sandbox — pre-existing, not a regression from this change.

Closes #4522

### DIFF
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -75,6 +75,7 @@
 #   ./.loom/scripts/cli/loom-daemon-start.sh --no-systemd    Linux only: use legacy nohup instead of a systemd --user service
 #   ./.loom/scripts/cli/loom-daemon-start.sh --print-plist   Print the LaunchAgent plist that WOULD be installed and exit (no side effects)
 #   ./.loom/scripts/cli/loom-daemon-start.sh --print-unit    Print the systemd --user unit that WOULD be installed and exit (no side effects)
+#   ./.loom/scripts/cli/loom-daemon-start.sh --force-env     Suppress the dropped-env-key warning (#4522) for an intentional narrower re-render
 #   ./.loom/scripts/cli/loom-daemon-start.sh --help
 #
 # Environment:
@@ -258,6 +259,95 @@ extract_plist_path_value() {
         /<key>PATH<\/key>/ { want=1; next }
         want { sub(/^[ \t]*<string>/, ""); sub(/<\/string>[ \t]*$/, ""); print; exit }
     ' "$plist_file"
+}
+
+# ---------- dropped-env-key detection (#4522) ----------
+# Root cause under test: render_launchd_plist / render_systemd_unit render the
+# EnvironmentVariables dict / Environment= lines strictly from whatever THIS
+# invocation has exported ("Respected when already exported" note above). Any
+# invocation missing the operator's exports (a watchdog, a bare re-run, another
+# tool shelling out to this script) silently replaces a richer installed
+# plist/unit with a narrower one -- e.g. every LOOM_SAFEHOUSE_* key and
+# LOOM_WORK_FINDER=1 quietly gone. The functions below detect that KEY REMOVAL
+# (not a value change -- see the PATH-specific drift check above for that) so
+# it is surfaced instead of silently applied.
+
+# extract_plist_env_keys <plist_file> — list of every <key>...</key> entry
+# inside the EnvironmentVariables dict, one per line. Extends
+# extract_plist_path_value's textual-awk approach (no plutil/XML-parser
+# dependency) from a single key to the whole dict.
+extract_plist_env_keys() {
+    local plist_file="$1"
+    [[ -f "$plist_file" ]] || return 1
+    awk '
+        /<key>EnvironmentVariables<\/key>/ { in_env=1; next }
+        in_env && /<\/dict>/ { exit }
+        in_env && /<key>/ {
+            line=$0
+            sub(/^[ \t]*<key>/, "", line)
+            sub(/<\/key>.*$/, "", line)
+            print line
+        }
+    ' "$plist_file"
+}
+
+# extract_systemd_env_keys <unit_file> — list of every `Environment=KEY=...`
+# key in a rendered systemd unit, one per line. The systemd analog of
+# extract_plist_env_keys above.
+extract_systemd_env_keys() {
+    local unit_file="$1"
+    [[ -f "$unit_file" ]] || return 1
+    sed -n 's/^Environment=\([^=]*\)=.*/\1/p' "$unit_file"
+}
+
+# warn_dropped_env_keys <old_file> <new_file> <extractor_function_name> — compare
+# the env-var KEY sets (not values) between an already-installed plist/unit and
+# a freshly-rendered replacement; warn (listing the keys) when the replacement
+# DROPS a key the installed file carried. <extractor_function_name> is
+# extract_plist_env_keys or extract_systemd_env_keys.
+#
+#   - A missing old_file (first-ever install -- nothing installed yet) is not a
+#     drop: returns silently, no warning.
+#   - --force-env (FORCE_ENV=true) acknowledges an intentional narrowing (e.g.
+#     an explicit minimal re-render) and suppresses the warning.
+#   - A dropped LOOM_SAFEHOUSE_* key gets a specific migration hint (the
+#     "safehouse" block in .loom/config.json + --from-config, #4353) instead of
+#     a generic warning.
+warn_dropped_env_keys() {
+    local old_file="$1" new_file="$2" extractor="$3"
+    [[ -f "$old_file" ]] || return 0
+    [[ "${FORCE_ENV:-false}" == "true" ]] && return 0
+
+    local old_keys new_keys
+    old_keys="$("$extractor" "$old_file" 2>/dev/null || true)"
+    [[ -z "$old_keys" ]] && return 0
+    new_keys="$("$extractor" "$new_file" 2>/dev/null || true)"
+
+    local dropped=() k nk hit
+    while IFS= read -r k; do
+        [[ -z "$k" ]] && continue
+        hit=false
+        if [[ -n "$new_keys" ]]; then
+            while IFS= read -r nk; do
+                if [[ "$nk" == "$k" ]]; then hit=true; break; fi
+            done <<< "$new_keys"
+        fi
+        [[ "$hit" == "false" ]] && dropped+=("$k")
+    done <<< "$old_keys"
+
+    [[ "${#dropped[@]}" -eq 0 ]] && return 0
+
+    warn ""
+    warn "WARNING: re-rendering $new_file drops ${#dropped[@]} env key(s) present in the installed $old_file:"
+    for k in "${dropped[@]}"; do
+        if [[ "$k" == LOOM_SAFEHOUSE_* ]]; then
+            warn "  - $k (config-tier equivalent: the \"safehouse\" block in .loom/config.json + --from-config, #4353)"
+        else
+            warn "  - $k"
+        fi
+    done
+    warn "This usually means this invocation ran without the operator's exported env (a watchdog / automated re-render / a bare re-run from a different shell)."
+    warn "Pass --force-env to acknowledge an intentional narrowing and suppress this warning."
 }
 
 # render_launchd_plist <label> <daemon_bin> <workdir> <log_path>
@@ -784,6 +874,11 @@ NO_LAUNCHD=false
 NO_SYSTEMD=false
 PRINT_PLIST=false
 PRINT_UNIT=false
+# --force-env (#4522): acknowledges an intentional narrower re-render and
+# suppresses warn_dropped_env_keys' warning. Script-only (like --print-plist),
+# not a daemon autonomy flag -- excluded from the persisted .daemon.flags file
+# below.
+FORCE_ENV=false
 while [[ $# -gt 0 ]]; do
     case "$1" in
         --help|-h) show_help; exit 0 ;;
@@ -797,6 +892,7 @@ while [[ $# -gt 0 ]]; do
         --no-systemd) NO_SYSTEMD=true; shift ;;
         --print-plist) PRINT_PLIST=true; shift ;;
         --print-unit) PRINT_UNIT=true; shift ;;
+        --force-env) FORCE_ENV=true; shift ;;
         *) err "Unknown option '$1'"; echo "Use --help for usage" >&2; exit 1 ;;
     esac
 done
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

## Evaluation

- **Scope check**: `git diff origin/main...HEAD` is exactly the two files the PR
  description claims (`loom-daemon-start.sh` + its test suite), 391 lines. (The
  worktree's local `main` ref was stale and showed a much larger diff against
  unrelated commits — verified against `origin/main` instead.)
- **Correctness**: `warn_dropped_env_keys` correctly no-ops on first-ever install
  (missing old file), respects `--force-env`, and gives `LOOM_SAFEHOUSE_*` drops
  a specific migration hint. Both write sites (launchd plist, systemd unit) now
  render to a scratch file first and diff against the live installed file before
  the `mv` — so the comparison happens before the old file is clobbered, not
  after. `--print-plist`/`--print-unit` get the same check as a read-only
  inspection counterpart. `--force-env` is correctly excluded from the
  persisted `.daemon.flags` file (script-only flag, not daemon autonomy state).
- **Tests**: `test-loom-daemon-start.sh` — 60/60 passing, including 12 new
  assertions covering first-install-never-warns, key-drop-warns-and-names-it,
  `LOOM_SAFEHOUSE_*` migration hint, same/superset-never-warns, and
  `--force-env` suppression on both the plist and systemd paths.
- **Shellcheck**: `shellcheck --severity=warning` clean on both changed files.
- **CI**: all checks pass or skipped (no failures, no pending).
- **Merge state**: `clean`, no rebase needed.
- **Issue linking**: body ends with `Closes #4522`. ✅

## Test Execution

1. Extended `test-loom-daemon-start.sh` (12 new assertions, 60/60) — ✅ Executed: all pass
2. `test-loom-daemon-launchd-plist.sh` (42/42) — ✅ Executed: all pass
3. `test-loom-daemon-stop.sh` (33/33) — ✅ Executed: all pass
4. `shellcheck --severity=warning` clean — ✅ Executed: no output, clean
5. `test-loom-daemon-update.sh` — ⏭️ Skipped per PR notes (untouched by this PR, slow `cargo build` step unrelated to this change)

## Follow-up filed

The PR's own root-cause investigation found that `scripts/loom`'s
`loom_cmd_restart()` bare-exec fallback is the actual cause of the original
incident, and is intentionally left unfixed here (scoped to the warning
mechanism only). Filed #4581 to track applying the existing
`perform_relaunch`/`perform_systemd_relaunch` harvest-and-preserve pattern to
that fallback.

rjwalters:

- **Size check**: 401 lines changed (+391/-10), exceeds the configured `champion.auto_merge_max_lines` limit of 200. No `loom:auto-merge-ok` waiver label is present.

All other criteria pass: Judge approved, mergeable (clean), no critical files touched, CI green (all checks pass or skipped), updated recently. The bulk of the diff is test coverage (253 of 391 added lines are in `test-loom-daemon-start.sh`), which the Judge review already verified as in-scope for the fix.

**Next steps:**
- A human or Judge can add `loom:auto-merge-ok` to waive the size limit for this PR (Champion does not self-grant this waiver)
- Alternatively, a human can merge manually via `./.loom/scripts/merge-pr.sh 4579`

---
*Automated by Champion role*

rjwalters:
**Champion Auto-Merge**

This PR meets the safety criteria for automatic merging:

- Size: 401 lines (+391/-10) — over the 200-line default ceiling, merged on Champion judgment per operator policy (no waiver labels; see also #4571): changes are confined to one shell script + its test suite, the new behavior is advisory warn-only, and 12 of the added assertions are tests
- No critical files modified (`defaults/scripts/cli/loom-daemon-start.sh`, `defaults/scripts/tests/test-loom-daemon-start.sh`)
- No merge conflicts (mergeable state: clean)
- Updated recently
- All CI checks passing (Shellcheck, Installer Integration Tests, Codex Adapter Smoke; Rust suites skipped — no Rust changes)

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #4571, opened 2026-07-30
champion: replace the auto-merge line-count ceiling with Champion judgment on whether a PR needs a human merge


rjwalters:
## Context

The Champion's auto-merge safety gate includes a hard size ceiling: `champion.auto_merge_max_lines` (default 200), waivable per-PR via `loom:auto-merge-ok` (`champion-pr-merge.md`). On 2026-07-30 this ceiling stalled four consecutive Judge-approved, CI-green PRs (#4551, #4558 at 886 lines, #4560 at 361, #4562 at 288) — every substantive Rust-plus-tests PR exceeds 200 lines structurally, so in practice the ceiling holds *all* real work for manual merge while the waiver label has no producer (Judges do full-diff verification but nothing tells them to apply `loom:auto-merge-ok`).

A raw line count is a poor risk proxy: 800 lines of new test fixtures is low-risk; 40 lines in `merge-pr.sh` or a guard hook is not. The Champion is an LLM agent already reading the PR, the Judge's review, and the diff — it can make that judgment directly.

**Operator decision (2026-07-30): remove the numeric ceiling and replace it with the Champion's best judgment** as to whether the PR should be held for a human to merge.

## Suggested acceptance criteria

- [ ] `champion-pr-merge.md`'s size check is replaced with a judgment criterion: the Champion assesses whether the PR is safe to auto-merge, considering diff composition (test/docs vs. load-bearing code), blast radius (critical files, e.g. merge scripts, guard hooks, CI workflows, installers), Judge review depth (full-diff verification vs. light pass), and revertability — not raw line count.
- [ ] When the Champion holds a PR, it posts a comment stating the *specific* concern ("touches merge-pr.sh's ordering guard; hold for human") rather than a line count, keeps `loom:pr`, and retries next tick — same escalation shape as today.
- [ ] `champion.auto_merge_max_lines` config and the `loom:auto-merge-ok` waiver label are retired or repurposed (if kept, `loom:auto-merge-ok` becomes an explicit human/Judge signal that overrides a Champion hold). Migration note for repos that set the config knob.
- [ ] `champion-reference.md`, `champion.md`, and `.github/labels.yml` descriptions updated consistently; CLAUDE.md untouched (detail stays in role docs).
- [ ] The other safety criteria (Judge-approved, CI green, mergeable, no contradictory verdict labels) are unchanged — this replaces only the size gate.

---

## Curator Addendum

Full curation enhancement (duplicate check, file verification, Affected Files,
[...]

PR #4580, opened 2026-07-30
harden(sync-labels): honor --, widen the gitea veto to config tiers, tighten NWO, preflight --repo before deletions

### LINKED ISSUE #4524: sync-labels --repo hardening: -- terminator, config-tier gitea guard, stricter NWO regex, gh repo view preflight before remote deletions (PR #4506 judge follow-ups)
## Context
PR #4506 (sync-labels `--repo` flag) merged with four explicitly non-blocking judge follow-ups (review comment 5125912516):
1. `--` is not honored as an end-of-options terminator in the new argv loop
2. The gitea rejection guard checks only `LOOM_FORGE_TYPE` env — misses `forge.type: gitea` resolved from config tiers (`defaults/scripts/lib/forge-helpers.sh:111-121`)
3. The NWO validation regex accepts `../..` and `-foo/bar`
4. Deletion-at-a-distance is mitigated (`--dry-run`) but not closed: dry-run is deliberately forge-free so it cannot catch a typo that names a real administrable repo — a `gh repo view -R "$REPO"` existence/permission preflight before the deletion loop would

- [ ] All four addressed in `defaults/scripts/sync-labels.sh` with cases added to `defaults/scripts/tests/test-sync-labels-repo-flag.sh` (52 existing assertions must stay green)
- [ ] No-flag behavior stays byte-identical (the existing short-circuit proofs must still pass)

rjwalters:
## Summary

Addresses the four explicitly non-blocking judge follow-ups from PR #4506 (review comment 5125912516) on `sync-labels.sh --repo`. All four land in `defaults/scripts/sync-labels.sh` (which `.loom/scripts/sync-labels.sh` is a directory symlink to, so both surfaces are covered by one edit).

## Changes

**1. `--` is a real end-of-options terminator.** The `--)` case was a bare `shift` — the token was swallowed as a no-op, so anything after it that started with `-` still hit the `-*)` "Unknown option" branch. A `POSITIONAL_ONLY` flag now short-circuits option dispatch for the rest of argv, and the positional handling is factored into a `take_positional()` helper so the pre-`--` and post-`--` paths cannot drift apart. `cd "$WORKTREE_PATH"` becomes `cd -- "$WORKTREE_PATH"`, since a `-`-leading path is now reachable and a bare `cd -x` would be parsed as a cd flag.

**2. The Gitea veto covers config tiers, not just the env var.** New `_configured_forge_type()` mirrors the first two tiers of `forge_detect`'s precedence (`forge-helpers.sh:92-121`): `LOOM_FORGE_TYPE`, then `forge.type` via `loom_config_get` over the resolved config-tier chain. So `forge.type: gitea` set in `.loom/config.json` / `.loom-project/project.json` / `.loom-local/local.json` now fails loudly instead of silently syncing GitHub. It deliberately stops short of `forge_detect`'s *third* tier (git-remote autodetection): `--repo` names its target explicitly, so only an explicit **setting** should veto it, never an inference from whichever checkout the operator happens to be standing in. The guard moved below the `cd` so the tiers resolve against WORKTREE_PATH's repo, exactly like the `forge_detect` call further down. The message distinguishes the env-var case from the config case.

**3. Tightened NWO regex.** `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$` -> `^[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9][A-Za-z0-9._-]*$`. Requiring each segment to *start* with an alphanumeric is what rejects path-traversal shapes (`../..`, `owner/..`, `./x`, `../owner`) and `-`-leading segments (`-foo/bar`, which a downstream `gh ... -R -foo/bar` would read as a flag bundle). Dots/dashes/underscores stay legal *inside* a segment, so `my-org/my-repo.name`, `owner-2/repo.name`, `owner/repo-1.2.3` still pass.

**4. `gh repo view` preflight before the deletion loop.** `--dry-run` is deliberately forge-free, so it cannot distinguish a typo'd NWO from the repo you meant — if the typo names a real repo you can administer, the preview looks perfectly reasonable and the real run then deletes that repo's default labels. The real (non-dry-run) `--repo` path now runs `gh repo view "$REPO" --json nameWithOwner,viewerPermission` immediately before the `DEFAULT_LABELS` loop (the first irreversible step) and aborts if the target is unreachable or read-only. `viewerPermission` absent (some fine-grained PATs omit it) warns and continues, since existence is still proven. Scoped to the `--repo` path on purpose: the no-flag path already resolved REPO via `gh repo view` in the current checkout, and adding a second call there would change its byte-for-byte behavior. stderr is captured to a temp file rather than folded into the value with `2>&1`, so a gh deprecation notice can't be mistaken for the permission.

Docs: the `--repo` fleet-onboarding paragraph in `defaults/docs/daemon-reference.md` now states the preflight and the widened Gitea rejection.

- [x] **All four addressed in `defaults/scripts/sync-labels.sh`** — see Changes above.
- [x] **Cases added to `defaults/scripts/tests/test-sync-labels-repo-flag.sh`** — four new sections (`NWO validation rejects traversal / '-'-leading segments`, `'--' is a real end-of-options terminator`, config-tier cases folded into `--repo is GitHub-only`, `--repo preflights the target before any deletion`).
- [x] **The 52 existing assertions stay green.** 52/52 with the new script, but note that **two had to be re-expressed rather than relaxed**, because they were written against the pre-preflight world:
  - `assert_not_contains "$LOG" "repo view"` — its stated intent is proving the NWO *resolution* short-circuit (`forge_get_repo_nwo`). That call is the targetless `gh repo view --json nameWithOwner --jq ...`; the preflight is a different call that *names* its target. The assertion now matches that exact resolution argv, so it still fails if resolution ever creeps back.
  - `no gh call was made without the override target` — `gh repo view` has no `-R` flag (verified against gh 2.96.0: the repo is positional), so the preflight names the target positionally. The assertion now matches on the NWO itself. The companion `the override NWO is the ONLY -R target` assertion is unchanged and still passes.
- [x] **No-flag behavior stays byte-identical.** All the existing short-circuit/no-flag proofs pass unchanged, plus a new explicit guard: the no-flag run still makes **exactly one** `gh repo ...` call (resolution only, no preflight added).

## Test Plan

- [x] `bash defaults/scripts/tests/test-sync-labels-repo-flag.sh` -> **131/131 passed, 0 failed** (52 pre-existing + 79 new).
  - The test harness gained: a stub `gh` that distinguishes a targeted `repo view <NWO>` (preflight, answer controlled by `LOOM_TEST_GH_REPO_VIEW`) from the targetless resolution call; `run_sls --repo-view` / `--repo-root` options; `LOOM_CONFIG_DEFAULTS_FILE=""` pinned so the host's private-defaults config tier can never leak into a run; a one-label `-alt/` scratch tree whose name begins with `-` (so `Synced 1 labels` *proves* `--` routed it to WORKTREE_PATH rather than merely not erroring); and `.loom/config.json` fixtures declaring gitea and github.
- [x] `bash defaults/scripts/tests/test-forge-helpers.sh` -> 36/36 passed.
- [x] `bash defaults/scripts/tests/test-config-resolver.sh` -> 14/14 passed.
- [x] `shellcheck -x` clean on both changed shell files.
- [x] Manual verification against real `gh`: `sync-labels.sh --repo rjwalters/definitely-does-not-exist-4524-preflight .` aborted at the preflight with exit 1 and `nothing was deleted`, before any label operation. (The account was rate-limited at the time, so the underlying gh failure was a 403 rather than a 404 — the preflight fails **closed** either way, which is the desired behavior before irreversible deletions.)
- [x] Edge cases: `owner/repo.name`, `my-org/my-repo`, `owner-2/repo.name`, `under_score/re_po`, `owner/repo-1.2.3`, `0rg/9repo` all still accepted (asserted).

### Pre-existing failure, not caused by this PR

`bash .loom/scripts/build-gate.sh` fails on one `loom-daemon` unit test. This is a **pre-existing parallel-execution flake**, not related to this shell-only diff:
- `quarantine_reconciliation::tests::reconcile_workspace_never_touches_manual_block` passes in isolation (in this branch's worktree) and fails only in the full parallel run.
- The same full run on **`main`** also fails, with *different* tests (`pipeline_snapshot::tests::gh_pipeline_source_counts_each_metric_independently`, `...records_error_but_keeps_other_metrics`).

Closes #4524

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -3035,8 +3035,14 @@ done
 
 `--dry-run` is deliberately forge-free (it reports intent from labels.yml alone
 and makes zero `gh` calls), so it also validates the parse and the NWO offline.
-`--repo` is GitHub-only — for a Gitea root, run `sync-labels.sh` from a checkout
-of that repo so the Gitea helpers can resolve a base URL and token.
+Because that means a dry run cannot tell a typo'd NWO apart from a repo that
+really exists, the **real** run preflights the named target with `gh repo view`
+before deleting anything and aborts if it is missing or read-only (#4524) —
+nothing is deleted when the preflight fails. `--repo` is GitHub-only — an
+explicitly configured Gitea forge rejects it, whether that comes from
+`LOOM_FORGE_TYPE=gitea` or `forge.type: gitea` in any config tier. For a Gitea
+root, run `sync-labels.sh` from a checkout of that repo so the Gitea helpers can
+resolve a base URL and token.
 
 **3. Refresh a stale install.** A root installed before the machine-level daemon
 model carries a committed file-copy of `.loom/` that **shadows** the machine
--- a/defaults/scripts/sync-labels.sh
+++ b/defaults/scripts/sync-labels.sh
@@ -10,11 +10,12 @@
 # Supports both GitHub (via the gh CLI) and Gitea (via the forge API).
 #
 # Usage:
-#   .loom/scripts/sync-labels.sh [WORKTREE_PATH]
-#   .loom/scripts/sync-labels.sh --repo OWNER/NAME [--dry-run] [WORKTREE_PATH]
+#   .loom/scripts/sync-labels.sh [--] [WORKTREE_PATH]
+#   .loom/scripts/sync-labels.sh --repo OWNER/NAME [--dry-run] [--] [WORKTREE_PATH]
 #
 #   WORKTREE_PATH  Directory containing .github/labels.yml and a git remote.
-#                  Defaults to the current directory.
+#                  Defaults to the current directory. `--` ends option parsing,
+#                  so a path that begins with `-` can still be passed.
 #
 # --repo OWNER/NAME (#4498) retargets the sync at an arbitrary GitHub repo
 # while still reading labels.yml from WORKTREE_PATH. Because every GitHub label
@@ -25,10 +26,14 @@
 #     .loom/scripts/sync-labels.sh --repo "$r"
 #   done
 #
-# --repo bypasses forge detection entirely (the target is named, not inferred
-# from a git remote) and is GitHub-only. Pair it with --dry-run first: --repo
-# makes the default-label deletions land on a repo you are not standing in, so
-# a typo'd NWO is worth previewing.
+# --repo bypasses repository *resolution* (the target is named, not inferred
+# from a git remote) and is GitHub-only — an explicitly configured Gitea forge
+# (LOOM_FORGE_TYPE or forge.type in the resolved config) rejects it. Pair it
+# with --dry-run first: --repo makes the default-label deletions land on a repo
+# you are not standing in, so a typo'd NWO is worth previewing. A real (non
+# dry-run) --repo sync additionally preflights the named target with
+# `gh repo view` before deleting anything (#4524), because a dry run is
+# deliberately forge-free and so cannot tell a typo apart from a real repo.
 #
 # This is the installed-tree counterpart of the source-only
 # scripts/install/sync-labels.sh. It is self-contained apart from the
@@ -40,8 +45,8 @@ set -euo pipefail
 
 usage() {
   cat <<'EOF'
-Usage: sync-labels.sh [WORKTREE_PATH]
-       sync-labels.sh --repo OWNER/NAME [--dry-run] [WORKTREE_PATH]
+Usage: sync-labels.sh [--] [WORKTREE_PATH]
+       sync-labels.sh --repo OWNER/NAME [--dry-run] [--] [WORKTREE_PATH]
 
 Sync Loom workflow labels from .github/labels.yml onto the forge (GitHub or
 Gitea). Creates missing labels, updates existing ones to match labels.yml,
@@ -58,6 +63,8 @@ Options:
                     GitHub-only (uses the gh CLI).
       --dry-run     Print the label operations that would run and exit without
                     calling the forge. Recommended before a --repo run.
+      --            End of options: every remaining argument is positional, so
+                    a WORKTREE_PATH beginning with '-' can be passed.
   -h, --help        Show this help and exit.
 EOF
 }
@@ -66,8 +73,29 @@ WORKTREE_PATH="."
 REPO_OVERRIDE=""
 DRY_RUN=0
 POSITIONAL_SEEN=0
+# Set by `--`: from that point on every remaining argument is positional, even
+# one that starts with `-`. Without this the `--)` case below was a no-op shift
+# and a `-`-leading path still fell into the "Unknown option" branch (#4524).
+POSITIONAL_ONLY=0
+
+# Accept the single positional argument (WORKTREE_PATH), rejecting a second one.
+# Factored out so the pre-`--` and post-`--` paths cannot drift apart.
+take_positional() {
+  if [[ "$POSITIONAL_SEEN" -eq 1 ]]; then
+    echo "Unexpected extra argument: $1" >&2
+    usage >&2
+    exit 2
+  fi
+  WORKTREE_PATH="$1"
+  POSITIONAL_SEEN=1
+}
 
 while [[ $# -gt 0 ]]; do
+  if [[ "$POSITIONAL_ONLY" -eq 1 ]]; then
+    take_positional "$1"
+    shift
+    continue
+  fi
   case "$1" in
     -h|--help)
       usage
@@ -96,6 +124,7 @@ while [[ $# -gt 0 ]]; do
       shift
       ;;
     --)
+      POSITIONAL_ONLY=1
       shift
       ;;
     -*)
@@ -104,13 +133,7 @@ while [[ $# -gt 0 ]]; do
       exit 2
       ;;
     *)
-      if [[ "$POSITIONAL_SEEN" -eq 1 ]]; then
-        echo "Unexpected extra argument: $1" >&2
-        usage >&2
-        exit 2
-      fi
-      WORKTREE_PATH="$1"
-      POSITIONAL_SEEN=1
+      take_positional "$1"
       shift
       ;;
   esac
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Reviewed via the existing builder worktree (`.loom/worktrees/issue-4524`). Diff scope matches the PR description exactly (`defaults/scripts/sync-labels.sh`, its test file, and a `daemon-reference.md` doc update — verified against `origin/main`, since my local `main` ref was briefly stale).

**Correctness:**
- The `--` terminator fix is sound: `POSITIONAL_ONLY` cleanly short-circuits the option-dispatch loop, and factoring `take_positional()` means the pre-/post-`--` paths can't drift. `cd -- "$WORKTREE_PATH"` correctly guards against a `-`-leading path being misread as a `cd` flag.
- `_configured_forge_type()` correctly mirrors only the first two tiers of `forge_detect`'s precedence (env var, then config chain) — deliberately excluding git-remote autodetection, which is the right call since `--repo` names its target explicitly.
- The tightened NWO regex (`^[A-Za-z0-9][A-Za-z0-9._-]*/...`) correctly blocks path-traversal and `-`-leading segments while still accepting legitimate dotted/dashed names.
- The preflight (`repo_override_preflight`) runs after `REPO` is set (line 232) and before the first destructive loop — correct ordering. Capturing stderr to a temp file instead of `2>&1` avoids misreading a `gh` deprecation notice as the permission value. Handles missing `viewerPermission` (fine-grained PATs) with a warn-not-block fallback, which is the right trade-off (existence is already proven).

**Test Execution:**
1. `bash defaults/scripts/tests/test-sync-labels-repo-flag.sh` — ✅ Executed: 131/131 passed
2. `bash defaults/scripts/tests/test-forge-helpers.sh` — ✅ Executed: 36/36 passed
3. `bash defaults/scripts/tests/test-config-resolver.sh` — ✅ Executed: 14/14 passed
4. `shellcheck -x` on both changed shell files — ✅ Executed: clean
5. Manual verification against real `gh` — ⚠️ Skipped: requires live gh credentials/network state beyond this review's scope; the automated preflight tests already cover this path with a stubbed `gh`.

**CI:** all 9 check-runs completed (Rust jobs correctly `skipped` via path-filtering for this shell-only change; the rest `success`). Merge state `clean`.

**Docs:** `daemon-reference.md` update accurately reflects the new preflight and widened Gitea rejection.

Issue link verified: PR body contains "Closes #4524" and #4524 is open with matching scope.

No blocking concerns.

rjwalters:
✅ **Approved!**

Thorough review of all four hardening items. Every claim in the PR body checked out, including the two re-expressed assertions (which I verified independently rather than taking on trust). Approving.

## Verification performed

**Tests: 131/131 pass**, `shellcheck -x` clean on both changed shell files.

**Are the new tests real regression proofs?** I ran the *new* test file against `origin/main`'s pre-fix `sync-labels.sh` in a scratch tree. Result: the suite fails on **41+ assertions spanning all four fix areas** — every NWO-traversal case (`../..`, `owner/..`, `-foo/bar`, `.hidden/repo`, `_leading/repo`), the `--` terminator cases, the config-tier gitea rejection, and every preflight case. These are genuine regression tests, not assertions written to match whatever the new code happens to do.

**Were the two re-expressed assertions weakened?** I reconstructed the exact question: I took the new test file, reverted *only* those two assertion strings to their `main` forms, and ran it against the new script. Exactly **2 failures, 129 passes** — confirming the builder's accounting is precise and that no other pre-existing assertion silently changed meaning. On the substance, both re-expressions are legitimate:

1. `assert_not_contains "$LOG" "repo view"` → `"repo view --json nameWithOwner --jq"`. The new string is the *exact* argv of the resolution call in `forge_get_repo_nwo` (`forge-helpers.sh:911`: `"$gh_cmd" repo view --json nameWithOwner --jq '.nameWithOwner'`). The assertion's stated intent — proving `--repo` never asks the forge *which repo this is* — is fully preserved, and it still fails if resolution creeps back. The preflight is a distinguishable call because it *names* its target.
2. `grep -v -- '-R octocat/hello-world'` → `grep -v -- 'octocat/hello-world'`. Necessary because the preflight targets positionally. The narrow loss (a hypothetical call carrying the NWO somewhere other than `-R`) is covered by the untouched companion assertion `the override NWO is the ONLY -R target in the whole gh log`, which uses `grep -o -- '-R [^ ]*' | sort -u`. Combined coverage is preserved.

**`gh repo view` has no `-R` flag — confirmed.** `gh repo view --help` on gh 2.96.0: `USAGE: gh repo view [<repository>] [flags]`, and `-R/--repo` is absent from both FLAGS and INHERITED FLAGS. The issue text's suggested `gh repo view -R "$REPO"` would have failed outright; the builder was right to deviate and to say so.

**Preflight placement.** Verified `repo_override_preflight` is invoked at `sync-labels.sh:429`, immediately before `info "Removing default labels..."` and the `DEFAULT_LABELS` loop — the genuine first irreversible step. The test `the preflight is the FIRST gh call` pins this via `head -1` of the gh log, so the ordering is enforced, not just currently true. Correctly gated on `-n "$REPO_OVERRIDE" && "$DRY_RUN" -eq 0`, and a new assertion proves the no-flag path still makes exactly one `gh repo ...` call (byte-identical behavior preserved).

**Gitea guard fidelity.** `_configured_forge_type()` faithfully mirrors `forge_detect`'s tiers 1–2 (`forge-helpers.sh:92-121`): same `_forge_config_root` precedence, and `loom_config_get` degrades to the `"auto"` default when jq is missing — matching `forge_detect`, which skips tier 2 entirely without jq. Deliberately stopping short of tier 3 (git-remote autodetection) is the right call and well justified in the comment: `--repo` names its target, so only an explicit *setting* should veto it. Moving the guard below the `cd` so tiers resolve against `WORKTREE_PATH` is correct and consistent with the `forge_detect` call further down.

**Security surface.** The tightened regex `^[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9][A-Za-z0-9._-]*$` runs *before* `$REPO` reaches any `gh` argument, so an argument-injection shape like `-foo/bar` can no longer be handed to `gh label delete ... -R`. `cd -- "$WORKTREE_PATH"` correctly closes the flag-injection hole that `--` newly opens. `repo_override_preflight` routing stderr to a temp file instead of `2>&1` is a genuinely good detail — folding a gh deprecation notice into `$out` would have been read as a permission string and fallen through to the `*)` abort branch. `error`/`rm -f` ordering in the failure path leaks no temp file. Permission gating on `ADMIN|MAINTAIN|WRITE` correctly rejects `TRIAGE`/`READ`, and the empty-string warn-and-continue is right — existence is proven and blocking a fine-grained PAT would be a false positive.

**Scope check.** `scripts/install/sync-labels.sh` (the source-only counterpart named in the header) has no `--repo` path at all, so no parallel hardening is owed there. `.loom/docs/daemon-reference.md` is a symlink to `defaults/docs/daemon-reference.md`, so the doc update propagates to the installed tree automatically — no resync debt.

**CI**: all checks pass (Shellcheck, Installer Integration Tests, Codex Adapter Smoke, AGENTS.md Sync, CLAUDE.md Line Budget). The Rust jobs are `skipped` by path filter, which is correct for a shell-only diff — and it also means the build-gate `loom-daemon` flake the builder documented is not a gate on this PR. **Merge state: CLEAN.** Body carries `Closes #4524`.

## One non-blocking observation (inline, not worth an issue)

`_configured_forge_type()` diverges from `forge_detect` in one extreme-edge case. `forge_detect` only *accepts* `LOOM_FORGE_TYPE` when it lowercases to exactly `github` or `gitea` and otherwise **falls through to the config tier**; `_configured_forge_type` returns any non-empty env value verbatim. So with `LOOM_FORGE_TYPE=gitlab` (or a stray trailing space) set *and* `forge.type: gitea` in config, `forge_detect` resolves gitea but the guard sees `gitlab` and does not fire.

Practical impact is near-nil — it requires a malformed env value plus a gitea config plus `--repo`, and the outcome is a GitHub sync against a repo the operator explicitly named and which the new preflight independently verifies. Flagging it only because the function's comment claims to "mirror the first two tiers," and this is the one place it doesn't. A `case` on `github|gitea` before accepting the env value would close it. Genuinely trivial — noting it inline rather than filing queue noise; fold it in if this file is ever touched again.

Excellent work: the failure modes are well reasoned (the "a dry run can't tell a typo from a real repo" framing is exactly the right argument for the preflight), the comments explain *why* rather than *what*, and the test harness additions — the `-alt/` one-label scratch tree so `Synced 1 labels` genuinely *proves* `--` routed the path, and pinning `LOOM_CONFIG_DEFAULTS_FILE=""` so host config can't leak — show real care about assertions that can actually fail.

rjwalters:
**Champion Auto-Merge**

This PR meets the safety criteria for automatic merging:

- Size: 404 lines (+367/−37) — over the 200-line default, **waived by Champion judgment** per operator policy (repair-over-gate; see #4571): the overage is dominated by 79 new test assertions and docs; the diff is shell + tests + docs only, addressing explicitly non-blocking Judge follow-ups from PR #4506
- No critical files modified (defaults/scripts + tests + docs)
- No merge conflicts (`mergeable_state: clean`)
- All CI checks passing (Shellcheck, Installer Integration Tests, AGENTS.md Sync, CLAUDE.md Line Budget, Codex Adapter Smoke; Rust jobs path-filtered/skipped as expected for a shell-only diff)

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #4472, opened 2026-07-29
require-complexity-marker.sh reports 'missing marker' when gh fails (GraphQL quota) — cannot distinguish fetch failure from absent marker


rjwalters:
## Problem

`.loom/scripts/require-complexity-marker.sh` fetches the issue body with:

```bash
body="$(gh issue view "$ISSUE" -R "$REPO" --json body -q .body 2>/dev/null || true)"
```

`gh issue view` is a GraphQL call and the `|| true` + `2>/dev/null` swallow every failure mode. When the GraphQL quota is exhausted (routine under concurrent agents — see epic #4432), `body` is empty, the tier parses as `""", and the script exits 1 with the **BLOCKED: issue has no complexity marker** message — indistinguishable from a genuinely unmarked issue.

Observed live 2026-07-29: during a curator pass with GraphQL exhausted, the validator reported all four spot-checked issues (#4393, #4463, #4464, #4466) as missing markers; REST fetch showed two of them (#4463, #4466) had valid markers. Consequence: a curator following the role instructions ("do not apply loom:curated if it fails") is blocked from curating anything while quota is out, or worse, re-edits bodies that are already correct.

## Suggested fix

Distinguish fetch failure from empty marker:

- Capture `gh`'s exit status instead of `|| true`; on failure, exit 2 ("could not fetch — retry or check quota"), not 1.
- Optionally fall back to REST (`gh api repos/<nwo>/issues/<n> --jq .body`) before failing, matching the established GraphQL→REST fallback pattern used elsewhere.
- An empty body from a *successful* fetch should still be treated as missing-marker exit 1.

## Affected Files

- `defaults/scripts/require-complexity-marker.sh` — fetch + error-path split (installed copy at `.loom/scripts/require-complexity-marker.sh` is a symlink/synced surface; change the defaults source)

## Test Plan

- [ ] Unit: simulated `gh` failure (stub gh exiting 1) → script exits 2 with a fetch-error message, not the BLOCKED-missing text
- [ ] Unit: successful fetch, no marker → exit 1 BLOCKED as today
- [ ] Unit: successful fetch, valid/invalid marker → exit 0 / exit 1 as today
- [ ] Manual: run against a real issue with REST fallback while GraphQL is exhausted

Related: epic #4432 (survive GitHub API rate limits at fleet scale).

---

## Curator Enhancement (2026-07-29)

[...]

Issue #4570, opened 2026-07-30
judge: concurrent-Judge verdict race produced loom:pr + loom:changes-requested — add verdict-time CAS recheck + verdict-label mutual-exclusion janitor


rjwalters:
## Context (PR #4560 incident, 2026-07-30)

Two Judges reviewed PR #4560 concurrently and produced a contradictory label state outside the intended lifecycle graph:

```
05:11:09  loom:review-requested   (Builder)
05:16:56  loom:reviewing          (Judge A claims)
05:25:03  loom:pr                 (Judge B approves — stomping A's 8-min-old FRESH claim)
05:27:12  loom:changes-requested  (Judge A's verdict lands)
05:29:03  loom:treating           (Doctor picks up A's rejection)
```

From 05:25–05:30 the PR carried `loom:pr` AND `loom:changes-requested` simultaneously — a state that would have let Champion auto-merge a PR with an open blocker. Judge A detected the collision and manually corrected fail-safe (removed `loom:pr`, documented the timeline in a PR comment).

Three stacked gaps:

1. **No write atomicity.** Label transitions are read-modify-write with ~10 minutes between claim check and verdict write; GitHub offers no CAS. Interleaved writers can always mint an off-graph state.
2. **The claim protocol is advisory prose.** Judge B violated `judge.md`'s stale-claim rule (30-min window; A's claim was 8 minutes old) and nothing mechanical stopped the verdict write. (A separately-dispatched judge on PR #4437 the same night hit the identical race and correctly aborted — the rule works when followed; it is unenforced.)
3. **Upstream duplicate dispatch** put two Judges on one PR (#4556 tracks that recurrence; #4527 tracks the Doctor-side claim protocol). This issue covers the Judge-side write race and the invariant, which neither of those closes.

## Suggested acceptance criteria

- [ ] **Verdict-time CAS recheck (Judge + Doctor)**: immediately before writing any verdict label, the role re-reads the PR's labels and ABORTS (discarding its verdict, posting a short note) if (a) another actor's verdict label appeared since its claim, or (b) its own claim label (`loom:reviewing` / `loom:treating`) was removed or replaced. Shrinks the race window from minutes to seconds. Documented in `judge.md` + `doctor.md` next to the existing stale-claim rule.
[...]

Issue #999, opened 2026-01-24
Orchestrator: handle worktree merge gracefully


rjwalters:
## Problem

When running `/loom` orchestration from an issue worktree (e.g., `.loom/worktrees/issue-981`), the `gh pr merge` command fails with:

```
failed to run git: fatal: 'main' is already used by worktree at '/Users/rwalters/GitHub/loom'
```

**What happens:**
1. PR merge succeeds on GitHub ✅
2. `gh` CLI tries to checkout main locally
3. main is already checked out in parent worktree
4. Command reports failure despite successful merge

**Current workaround:**
We manually verify with `gh pr view --json state,mergedAt` after the error, which works but is noisy and confusing.

## Acceptance Criteria

- [ ] Orchestrator handles this error gracefully when merging from worktrees
- [ ] No false failure messages when merge actually succeeded
- [ ] Clean output showing merge was successful

### Option 1: Verify-on-error pattern
```bash
# Try merge, if fails with worktree error, verify actual state
gh pr merge "$pr" --squash --delete-branch 2>&1 || {
  if gh pr view "$pr" --json state --jq '.state' | grep -q "MERGED"; then
    echo "✓ PR merged successfully (local checkout skipped - worktree conflict)"
  else
    echo "✗ Merge failed"
    exit 1
  fi
}
```

### Option 2: Skip local operations
```bash
# Use --repo flag to force API-only operations
gh pr merge "$pr" --squash --delete-branch --repo "$(gh repo view --json nameWithOwner -q .nameWithOwner)"
```

### Option 3: Run from main worktree
```bash
# Navigate to main worktree for merge operations
(cd /path/to/main/worktree && gh pr merge "$pr" --squash --delete-branch)
```

### Option 4: Document and accept
Add to orchestrator output: "Note: merge completed on GitHub; local checkout skipped due to worktree conflict"

## Context

This error occurs every time we run `/loom` with `--force` from an issue worktree. It's been observed in PRs #974, #982, #983, #987, #997.

The Loom orchestrator is designed to work from worktrees, so this should be handled as a normal case, not an error.

## Related

- Worktree workflow documented in CLAUDE.md
- `/loom` orchestrator role

Issue #801, opened 2025-10-26
Judge role incorrectly uses GitHub PR approval instead of comment-based reviews


rjwalters:
## Problem

The Judge role is currently using GitHub's PR approval system (gh pr review --approve) instead of following Loom's label-based coordination workflow. This creates issues:

- Uses gh pr review --approve with approval body
- Bypasses Loom's label coordination system
- May trigger GitHub's auto-merge on repos with that enabled
- Creates confusion about approval vs coordination

## Current Incorrect Behavior

What Judge currently does (WRONG):
gh pr review 40 --approve --body "Approved..."

This uses GitHub's review approval system, which is meant for human reviewers, not AI coordination.

## Expected Correct Behavior

Judge should use comment-based reviews with label transitions:

Step 1: Review the PR (read code, run tests, analyze)

Step 2: Add review comment with findings
gh pr comment NUMBER --body "Judge Review: Approved. Summary of assessment. Quality checks passed. Ready for merge."

Step 3: Update labels to signal approval
gh pr edit NUMBER --remove-label "loom:review-requested" --add-label "loom:pr"

Key difference: Use gh pr comment (not gh pr review --approve) plus label transition

## Why This Matters

1. Label-based coordination: Loom uses labels, not GitHub approval state
2. Champion workflow: Champion looks for loom:pr label, not approval status
3. Consistency: All roles use labels for state transitions
4. Clarity: Comments show reasoning without GitHub approval semantics
5. Flexibility: Repo auto-merge settings do not interfere with Loom workflow

## Required Changes

Review and update defaults/roles/judge.md:

1. Search for: All instances of gh pr review --approve or --request-changes
2. Replace with: Comment-based pattern shown above
3. Verify: Label transitions use gh pr edit with --add-label and --remove-label
4. Document: Explicitly state "Do NOT use gh pr review --approve"

## Acceptance Criteria

- Judge role uses gh pr comment for all review feedback (not gh pr review)
- Label transitions use gh pr edit --remove-label loom:review-requested --add-label loom:pr
- Explicit warning: Do NOT use GitHub PR approval system
- Example shows correct comment plus label pattern for both approval and changes requested
- All code examples in judge.md updated to use comment-based approach

## Investigation Needed

Curator should:
1. Read defaults/roles/judge.md in full
2. Identify all instances of gh pr review commands
3. Verify current instructions match or contradict comment-based approach
4. Provide exact line numbers and current vs correct patterns
[...]

PR #4587, opened 2026-07-30
fix: add REST fallback to check-duplicate.sh when GraphQL is rate-limited

### LINKED ISSUE #4526: check-duplicate.sh: no REST fallback when the GraphQL API is rate-limited
## Context
Part of the rate-limit resilience theme in #4432 (filed as a concrete child).

During an interactive `/loom:sweep` session on a consumer repo (rjwalters/once-around, 2026-07-29, Loom 0.13.0), the account's GraphQL quota was exhausted (0/5000) for most of the day while REST remained available. Three separate Curator runs reported `.loom/scripts/check-duplicate.sh` failing outright on the GraphQL rate limit, and each fell back to *manual* duplicate checking via `gh api` REST search — which worked fine every time. The label/comment/PR operations in the same session all had straightforward REST equivalents; the duplicate checker was the only Loom script the curators consistently couldn't use.

- [ ] `check-duplicate.sh` detects a GraphQL rate-limit failure (non-zero exit + rate-limit signature) and retries via the REST search API (`gh api search/issues ...`) instead of failing
- [ ] Degraded-mode output is clearly labeled (REST search ranks differently than the GraphQL path) so curators know the similarity scoring basis changed
- [ ] Exit code distinguishes "no duplicates found (REST fallback)" from "check could not run at all", so role prompts can stop treating every failure as skip-and-do-it-manually

---

## Curator Enhancement (2026-07-30)

Full enhancement — verified diagnosis (three call sites in `check-duplicate.sh` use GraphQL-backed `--json` list calls with inconsistent failure handling; the REST fallback pattern already exists in the same file at `search_cross_references()`), implementation guidance, affected files, and test plan — posted as a comment below to keep this body concise: see the "Curator Enhancement (2026-07-30)" comment on this issue.

rjwalters:
## Summary

`check-duplicate.sh` has three call sites (`search_similar_issues()`, `search_merged_prs()`, `search_closed_issues()`) that use `gh issue/pr list --json ...` (GraphQL-backed). When GraphQL quota is exhausted, GraphQL and REST draw from independent quotas — the open-issues path hard-errored (exit 2) while the merged-PR/closed-issue paths silently treated the failure as "no duplicates found," both wrong. This PR adds a REST fallback (mirroring the existing pattern already used by `search_cross_references()` in the same file) so all three call sites recover from a GraphQL rate-limit instead of failing or silently reporting a false negative.

## Changes

- Added `is_rate_limit_error()` — detects GitHub's GraphQL rate-limit signature (`API rate limit exceeded`) in captured `gh` output.
- Added `get_repo_nwo()` — resolves `owner/repo` via `gh repo view`, shared by the three new REST fallback call sites.
- `search_similar_issues()`: on a rate-limited GraphQL fetch, retries via `gh api repos/OWNER/REPO/issues?state=open&per_page=50`, filtering out PR entries (`.pull_request == null`).
- `search_merged_prs()`: on a rate-limited GraphQL fetch, retries via `gh api repos/OWNER/REPO/pulls?state=closed&per_page=20`, filtering to actually-merged PRs locally (`.merged_at != null`) since REST has no `state=merged` filter.
- `search_closed_issues()`: on a rate-limited GraphQL fetch, retries via `gh api repos/OWNER/REPO/issues?state=closed&per_page=20`, same PR filter as open issues.
- Degraded-mode labeling: when the REST fallback fires and matches are found, the output header reads `DUPLICATE_FOUND (REST fallback -- similarity ranking basis differs from GraphQL)` instead of the bare `DUPLICATE_FOUND`. The `--json` output parser (and the umbrella-header synthesis in `main()`) now matches a `DUPLICATE_FOUND` *prefix*, not exact-equals.
- Exit-code semantics: exit 2 now means "GraphQL **and** REST both failed" for each of the three searches (mirroring the two-tier GraphQL-then-REST-then-exit-2 shape already used by `.loom/scripts/require-complexity-marker.sh`, #4472) — previously `search_merged_prs`/`search_closed_issues` silently returned 0 ("no duplicates") on *any* GraphQL failure, which this fixes for the rate-limit case specifically.
- Audited `defaults/.claude/commands/loom/curator.md`'s `DUPLICATE_FOUND` usage: it only branches on `check-duplicate.sh`'s exit code (`if !`), never an exact string match, so no change was needed there.

| Detects GraphQL rate-limit failure and retries via REST instead of failing | ✅ | `is_rate_limit_error()` added; all three call sites gate the REST retry on it; new stub-`gh` tests (p), (r), (t) exercise each search path |
| Degraded-mode output is clearly labeled | ✅ | `DUPLICATE_FOUND (REST fallback -- ...)` header; tests (p), (r), (t), plus regression test (v) confirms it's absent on ordinary successful GraphQL calls |
| Exit code distinguishes "no duplicates (REST fallback)" from "could not run at all" | ✅ | Exit 2 only when both GraphQL and REST fail (per search); tests (q), (s), (u) cover both-failed for all three call sites |

## Test Plan

- `bash defaults/scripts/tests/test-check-duplicate.sh` — 74/74 passed (54 pre-existing + 20 new), including:
  - (p)/(q): open-issues rate-limit → REST success (labeled, PR-entry filtered) / REST failure → exit 2
  - (r)/(s): merged-PRs rate-limit → REST success (unmerged PR filtered) / REST failure → exit 2
  - (t)/(u): closed-issues rate-limit → REST success / REST failure → exit 2
  - (v): ordinary successful GraphQL call never mentions "REST fallback" (no regression)
  - All 54 pre-existing cases (cross-reference probe, keyword-similarity scoring, degenerate-result detection, real historical duplicate pairs) remain green
- `shellcheck defaults/scripts/check-duplicate.sh defaults/scripts/tests/test-check-duplicate.sh` — clean, no warnings

Closes #4526

### DIFF
--- a/defaults/scripts/check-duplicate.sh
+++ b/defaults/scripts/check-duplicate.sh
@@ -21,11 +21,26 @@
 #   1 - Potential duplicates found (listed to stdout)
 #   2 - Error (invalid arguments, gh command failed, etc.)
 #
+# A confirmed duplicate always wins over a partial failure (#4526): if ANY
+# search pool (open issues / merged PRs / closed issues) finds a match, the
+# exit code is 1 even when another pool could not be searched at all. Exit 2
+# means "no answer at all", never "we found something but also had trouble".
+#
 # Output format (when duplicates found):
 #   DUPLICATE_FOUND
 #   #<number>: <title> (similarity: <percent>%)
 #   PR #<number>: <title> (similarity: <percent>%)
 #   ...
+#
+# Degraded-coverage marker lines (#4526), appended after the match list when
+# GraphQL rate-limiting forced a degraded search. Both are informational: they
+# are not matches, and the --json parser reports them as `search_incomplete`
+# rather than as entries in `matches`.
+#   SEARCH_INCOMPLETE: <pools> -- that pool could not be searched at all
+#                                 (GraphQL rate-limited AND REST fallback
+#                                 failed), so the match list is partial
+#   REST_FALLBACK: <pools>     -- that pool was answered by the REST fallback,
+#                                 whose similarity ranking basis differs
 #
 # Degenerate-result self-detection (#4409): if more than half of the
 # candidates scanned in a given search (open issues / merged PRs / closed
@@ -251,15 +266,37 @@ calculate_similarity() {
     echo "$percent"
 }
 
-# Detect GitHub's GraphQL rate-limit error signature in captured `gh` output
-# (stdout+stderr merged via `2>&1`). GraphQL and REST draw on independent
-# quotas (confirmed live during the #4526 incident: `gh issue/pr list --json
-# ...` failed with this exact string while `gh api repos/OWNER/REPO/...`
-# succeeded), so this specific error -- and only this one -- justifies
-# retrying the same query via REST instead of giving up.
+# Detect a GitHub rate-limit rejection in captured `gh` output (stdout+stderr
+# merged via `2>&1`). GraphQL and REST draw on independent quotas (confirmed
+# live during the #4526 incident: `gh issue/pr list --json ...` failed with a
+# rate-limit error while `gh api repos/OWNER/REPO/...` succeeded), so this
+# class of error -- and only this class -- justifies retrying the same query
+# via REST instead of giving up.
+#
+# The signature table below mirrors loom-daemon/src/rate_limit_breaker.rs's
+# RATE_LIMIT_SIGNATURES, which is this repo's tested ground truth for what
+# `gh` actually prints. Two phrasings matter and they are NOT substrings of
+# each other:
+#
+#   GraphQL (`gh issue list` / `gh pr list`, i.e. every call in this script):
+#     "GraphQL: API rate limit already exceeded for user ID ..."
+#   REST (`gh api ...`, i.e. the fallbacks below):
+#     "HTTP 403: API rate limit exceeded for ..."
+#
+# The word "already" breaks the contiguous "API rate limit exceeded" match, so
+# a single-pattern check misses the GraphQL form -- which is precisely the
+# form #4526 is about. Matching is case-insensitive, as in the Rust original.
 is_rate_limit_error() {
-    local text="$1"
-    [[ "$text" == *"API rate limit exceeded"* ]]
+    local text
+    text=$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')
+    case "$text" in
+        *"api rate limit exceeded"*) return 0 ;;
+        *"api rate limit already exceeded"*) return 0 ;;
+        *"secondary rate limit"*) return 0 ;;
+        *"abuse detection mechanism"*) return 0 ;;
+        *"was submitted too quickly"*) return 0 ;;
+    esac
+    return 1
 }
 
 # Resolve "owner/repo" via REST, for the REST fallback paths below. Mirrors
@@ -683,10 +720,37 @@ main() {
         exit 2
     fi
 
+    # Verdict tracking (#4526). "A duplicate was found" and "a search pool
+    # could not be checked at all" are INDEPENDENT facts, and each of the
+    # three pools (open issues / merged PRs / closed issues) can report them
+    # separately -- partial rate-limiting across pools is the exact scenario
+    # this fallback exists for. Keeping the two facts in separate variables
+    # (instead of overwriting one exit code as each pool reports in) is what
+    # stops a total failure in one pool from discarding a confirmed match
+    # found by another: `exit_code` is derived from both, once, below.
+    #
+    #   duplicate_found  -> any pool produced a match/NON_DISCRIMINATIVE line
+    #   incomplete_pools -> human-readable list of pools where GraphQL was
+    #                       rate-limited AND the REST fallback also failed
+    #   fallback_pools   -> pools that were answered by the REST fallback
+    #
+    # Comma-joined strings rather than arrays: this script runs under
+    # `set -u` on bash 3.2 (macOS system bash), where empty-array expansion
+    # is a portability landmine.
+    local duplicate_found=false
+    local incomplete_pools=""
+    local fallback_pools=""
+    local header_labeled_rest=false
+
     # Search for similar issues
     local result
     local exit_code=0
     result=$(search_similar_issues "$title" "$body" "$threshold") || exit_code=$?
+    if [[ $exit_code -eq 1 ]]; then
+        duplicate_found=true
+    elif [[ $exit_code -eq 2 ]]; then
+        incomplete_pools="open issues"
+    fi
 
     # If --include-merged-prs, also search merged PRs and closed issues
     local merged_result=""
@@ -698,11 +762,15 @@ main() {
         closed_result=$(search_closed_issues "$title" "$body" "$threshold") || closed_exit_code=$?
 
         # #4526: a GraphQL rate-limit that ALSO fails via REST means this
-        # pool genuinely could not be checked -- surface that as exit 2
-        # ("could not run at all"), not silently as "no duplicates" the way
-        # a plain `return 0` used to.
-        if [[ $merged_exit_code -eq 2 || $closed_exit_code -eq 2 ]]; then
-            exit_code=2
+        # pool genuinely could not be checked. Record it; the final verdict
+        # below turns that into exit 2 ("could not run at all") rather than
+        # silently reporting "no duplicates" the way a plain `return 0` used
+        # to -- but only when no other pool found a real duplicate.
+        if [[ $merged_exit_code -eq 2 ]]; then
+            incomplete_pools="${incomplete_pools:+${incomplete_pools}, }merged PRs"
+        fi
+        if [[ $closed_exit_code -eq 2 ]]; then
+            incomplete_pools="${incomplete_pools:+${incomplete_pools}, }closed issues"
         fi
 
         # Strip the leading REST-fallback sentinel line (if present) from
@@ -719,28 +787,48 @@ main() {
             closed_rest_fallback=true
             closed_result="${closed_result#RATE_LIMIT_FALLBACK$'\n'}"
         fi
+        if $merged_rest_fallback; then
+            fallback_pools="${fallback_pools:+${fallback_pools}, }merged PRs"
+        fi
+        if $closed_rest_fallback; then
+            fallback_pools="${fallback_pools:+${fallback_pools}, }closed issues"
+        fi
 
         # If we found matches in merged PRs or closed issues, flag as duplicate
         if [[ -n "$merged_result" || -n "$closed_result" ]]; then
-            if [[ $exit_code -eq 0 ]]; then
-                # No open issue duplicates found, but merged/closed matches
-                # exist. Only synthesize a "DUPLICATE_FOUND" umbrella header
-                # when at least one of merged/closed is a real match list --
-                # a degenerate (#4409) result already self-announces with its
-                # own "NON_DISCRIMINATIVE (...)" line, and prefixing THAT
-                # with "DUPLICATE_FOUND" would be misleading. Check each side
+            if ! $duplicate_found; then
+                # The open-issues search produced no match of its own (either
+                # it found nothing, or it could not run at all -- in both
+                # cases $result is empty here, so overwriting it is safe), but
+                # merged/closed matches exist. Only synthesize a
+                # "DUPLICATE_FOUND" umbrella header when at least one of
+                # merged/closed is a real match list -- a degenerate (#4409)
+                # result already self-announces with its own
+                # "NON_DISCRIMINATIVE (...)" line, and prefixing THAT with
+                # "DUPLICATE_FOUND" would be misleading. Check each side
                 # independently (not "both must be non-degenerate") so a real
                 # match on one side still gets its header even when the other
                 # side is degenerate.
                 if [[ ( -n "$merged_result" && "$merged_result" != NON_DISCRIMINATIVE* ) || \
                       ( -n "$closed_result" && "$closed_result" != NON_DISCRIMINATIVE* ) ]]; then
-                    if $merged_rest_fallback || $closed_rest_fallback; then
+                    if [[ -n "$fallback_pools" ]]; then
                         result="DUPLICATE_FOUND (REST fallback -- similarity ranking basis differs from GraphQL)"$'\n'
+                        header_labeled_rest=true
                     else
                         result="DUPLICATE_FOUND"$'\n'
                     fi
                 fi
-                exit_code=1
+                duplicate_found=true
+            fi
+            # Re-establish the line separator before appending another pool's
+            # list. `result=$(search_similar_issues ...)` came back through a
+            # command substitution, which STRIPS the trailing newline, so
+            # appending directly splices the last open-issue line and the
+            # first merged-PR line into one line -- which the --json parser
+            # then reads as a single bogus match (the second entry's number
+            # lost, its title swallowed into the first entry's title).
+            if [[ -n "$result" && "$result" != *$'\n' ]]; then
+                result+=$'\n'
             fi
             if [[ -n "$merged_result" ]]; then
                 result+="$merged_result"$'\n'
@@ -749,6 +837,42 @@ main() {
                 result+="$closed_result"$'\n'
             fi
         fi
+    fi
+
+    # Degraded-coverage notes (#4526). Appended as their own marker lines so a
+    # Curator reading text output learns WHY a match list may be partial or
+    # differently ranked, and so --json can surface it as a field. Both are
+    # skipped by the --json line parser below.
+    #
+    # SEARCH_INCOMPLETE only appears alongside a real match: with no match the
+    # verdict is exit 2 and the per-pool stderr errors already say what broke.
+    if [[ -n "$result" && "$result" != *$'\n' ]] && \
+       { [[ -n "$incomplete_pools" ]] || [[ -n "$fallback_pools" ]]; } && $duplicate_found; then
+        result+=$'\n'
+    fi
+    if [[ -n "$incomplete_pools" ]] && $duplicate_found; then
+        result+="SEARCH_INCOMPLETE: ${incomplete_pools} -- GraphQL rate-limited and REST fallback also failed; matches below are partial"$'\n'
+    fi
+    # REST_FALLBACK covers the interleaving the umbrella header above cannot:
+    # the open-issues search already matched via ordinary GraphQL (so its own
+    # plain "DUPLICATE_FOUND" header stands) while a LATER pool fell back to
+    # REST. Without this, that REST-sourced subset would be reported with no
+    # indication its similarity ranking basis differs.
+    if [[ -n "$fallback_pools" ]] && ! $header_labeled_rest && $duplicate_found; then
+        result+="REST_FALLBACK: ${fallback_pools} -- answered via REST; similarity ranking basis differs from GraphQL"$'\n'
+    fi
+
+    # Final verdict (#4526): a confirmed duplicate ALWAYS outranks incomplete
+    # coverage. Reporting exit 2 when some other pool did find a match would
+    # discard the actionable answer -- and in --json mode the exit-2 branch
+    # below emits a bare {"error": ...} with no `matches` array at all, so the
+    # match would vanish entirely rather than merely be mis-coded.
+    if $duplicate_found; then
+        exit_code=1
+    elif [[ -n "$incomplete_pools" ]]; then
+        exit_code=2
+    else
+        exit_code=0
     fi
 
     # Cross-reference probe (--issue N only, issue #4162). Distinct from
@@ -780,6 +904,16 @@ main() {
                 degenerate=true
             fi
 
+            # Degraded-coverage flag (#4526): true when at least one search
+            # pool could not be checked at all (GraphQL rate-limited and its
+            # REST fallback also failed) while another pool still produced
+            # matches. Surfaced separately from `matches` so a caller can tell
+            # "no other duplicates exist" from "we could not look everywhere".
+            local search_incomplete=false
+            if [[ -n "$incomplete_pools" ]]; then
+                search_incomplete=true
+            fi
+
             # Parse duplicates into JSON (unconditional -- harmless no-op on
             # an empty $result, e.g. exit_code 0 or "related work only")
             local matches="[]"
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed the REST fallback added for GraphQL rate-limit failures in `check-duplicate.sh`.

**Correctness:**
- `is_rate_limit_error()` gates the fallback narrowly on the confirmed rate-limit signature; non-rate-limit GraphQL failures still hit the original error path unchanged (backward compatible).
- REST's `/issues` endpoint returning PRs is correctly filtered (`select(.pull_request == null)`), and the lack of a `state=merged` filter on `/pulls` is correctly worked around by filtering `merged_at != null` locally.
- The `RATE_LIMIT_FALLBACK` sentinel-line handoff from `search_merged_prs`/`search_closed_issues` to `main()` is a little intricate but the prefix-strip logic (`${var#RATE_LIMIT_FALLBACK$'\n'}`) is correct, and the `DUPLICATE_FOUND` prefix-match fix in the `--json` parser (was exact-equals) is necessary and correctly applied.
- Exit-code semantics change (silent "no duplicates" → exit 2 on double-failure) is a genuine bug fix for the merged-PR/closed-issue paths, and is a deliberate, well-justified behavior change per the PR description.
- Confirmed `curator.md`'s only consumer of this script branches on exit code (`if ! ./.loom/scripts/check-duplicate.sh ...`), not an exact `DUPLICATE_FOUND` string match — the audit claim in the PR description holds.

## Test Execution

- `bash defaults/scripts/tests/test-check-duplicate.sh` — ✅ Executed: **74/74 passed**, including all 20 new cases (p)–(v) covering REST-success, both-failed (exit 2), and the ordinary-GraphQL-success regression check.
- `shellcheck defaults/scripts/check-duplicate.sh defaults/scripts/tests/test-check-duplicate.sh` — ✅ Executed: clean, no warnings.
- CI: all required checks pass (Shellcheck, Installer Integration Tests, Codex Adapter Smoke, AGENTS.md Sync, CLAUDE.md Line Budget). Merge state is CLEAN.

PR description uses `Closes #4526` — issue will auto-close on merge. Nice, thorough test coverage for the new fallback paths.

rjwalters:
❌ **Changes Requested**

Found a real bug in the new exit-code aggregation logic in `main()` (`defaults/scripts/check-duplicate.sh:704-705`):

```bash
if [[ $merged_exit_code -eq 2 || $closed_exit_code -eq 2 ]]; then
    exit_code=2
fi
```

This unconditionally overwrites `exit_code` with `2` whenever the merged-PRs or closed-issues search totally fails (GraphQL rate-limited *and* REST fallback also fails) — **even if the open-issues search (`search_similar_issues`) already found a confirmed duplicate** and set `exit_code=1`. The real match is silently discarded rather than surfaced alongside the partial failure.

**Reproduced** by stubbing `gh` so open-issues search succeeds with a 100%-similarity match while merged-PRs search fails GraphQL *and* REST:

```
$ check-duplicate.sh --include-merged-prs --threshold 50 --title "Alpha Bravo Charlie Delta"
ERROR: GraphQL rate-limited fetching merged PRs, and REST fallback also failed: ...
DUPLICATE_FOUND
#801: Alpha Bravo Charlie Delta (similarity: 100%)
exit=2   # should be 1 -- a real duplicate WAS found and reported

$ check-duplicate.sh --include-merged-prs --threshold 50 --title "Alpha Bravo Charlie Delta" --json
ERROR: GraphQL rate-limited fetching merged PRs, and REST fallback also failed: ...
{"error": "Failed to check duplicates"}
exit=2   # the confirmed #801 match is fully discarded, not just mis-coded
```

In `--json` mode this is worse than a wrong exit code: `main()`'s `if [[ $exit_code -eq 2 ]]` branch (around line 770) emits a bare `{"error": ...}` and drops the `matches` array entirely, so a genuinely confirmed duplicate vanishes from the output. Today no caller in this repo passes `--json` (confirmed via grep across `.claude/commands/loom/*.md`), and text-mode callers only branch on zero/nonzero exit so this happens to be masked in practice — but this is exactly the failure mode (partial rate-limiting across the three search pools) that this PR exists to handle correctly, and it directly contradicts the PR's own stated acceptance criterion ("Exit code distinguishes 'no duplicates (REST fallback)' from 'could not run at all'"). It's also untested — tests (s)/(u) only exercise the both-failed case with no prior duplicate found by the open-issues search.

**Suggested fix**: don't let a total failure in one pool clobber a positive finding from another. Something like only setting `exit_code=2` when no duplicate was already found (e.g. `if [[ $exit_code -ne 1 ]] && ...`), or track the partial-failure fact separately from the duplicate-found fact so both can be surfaced (e.g. append a note to `$result` and keep `exit_code=1`).

## Test Execution

1. `bash defaults/scripts/tests/test-check-duplicate.sh` — not re-run in full here (manual repro above targets the specific gap); the existing 74 cases don't cover open-dup-found + merged/closed-total-failure combined, per the analysis above.
2. `shellcheck defaults/scripts/check-duplicate.sh defaults/scripts/tests/test-check-duplicate.sh` — not re-run here; CI's Shellcheck check is still pending on this PR.

Separately: CI is still pending on this PR (Shellcheck, Codex Adapter Smoke, Installer Integration Tests all showed `pending` as of this review) — worth re-running once the fix above lands anyway.

rjwalters:

- **Label conflict**: PR carries both `loom:pr` (Judge approved) and `loom:changes-requested` (Judge requested changes) simultaneously. This is the exact concurrent-Judge verdict race already tracked in #4570 — ambiguous Judge intent, so Champion will not resolve the conflict by unilaterally stripping either label.
- **Size check**: 292 lines changed (+267/-25), which exceeds the default 200-line auto-merge limit. No `loom:auto-merge-ok` override is present.

**Next steps:**
- A human (or Doctor once #4570 lands a fix) should determine the correct verdict label and remove the stale one.
- If the PR is genuinely approved and the size is acceptable, add `loom:auto-merge-ok` to waive the size limit once the label conflict is resolved.

Not modifying labels on this PR to avoid compounding the race condition.

---
*Automated by Champion role*

rjwalters:
**Champion: Fail-safe label correction (loom:pr + loom:changes-requested race)**

This PR briefly carried both `loom:pr` (06:35:24, from the first Judge approval) and `loom:changes-requested` (06:36:28, from a second Judge pass that found a real bug in the exit-code aggregation logic at `check-duplicate.sh:704-705`) simultaneously — the same race pattern documented in #4570.

The `changes-requested` verdict is later and identifies a genuine defect, so it is authoritative. Removing `loom:pr` to prevent an incorrect auto-merge. Routing to Doctor via the existing `loom:changes-requested` label.

---
*Automated by Champion role*

rjwalters:
⚠️ **Label conflict resolved — this PR needs changes, not merge**

This PR was evaluated concurrently by two Judge passes: one approved (`loom:pr`, comment above) at 06:35, and my changes-requested review (previous comment) landed a minute later at 06:36, citing a confirmed exit-code/data-loss bug in the new `main()` aggregation logic (`check-duplicate.sh:704-705` — a real duplicate found by the open-issues search gets silently discarded when the merged-PRs or closed-issues search totally fails, with a reproduced repro case in the prior comment).

I've removed the stray `loom:pr` label so the PR doesn't get auto-merged in its current state; `loom:changes-requested` stands. Please address the bug described above before re-requesting review.

rjwalters:
❌ **Changes Requested**

Thanks for the thorough PR — the structure is right (mirrors `search_cross_references()`'s existing REST-fallback pattern, correct `merged_at != null` / `pull_request == null` filtering, correct prefix-match fix for `DUPLICATE_FOUND`, correct `|| exit_code=$?` guard added around the new non-zero-returning `search_merged_prs`/`search_closed_issues` calls under `set -e`), and CI/tests are green. But I found one bug in the core detection logic that undermines the fix's actual purpose, so requesting changes before this lands.

## Blocking: `is_rate_limit_error()` misses GraphQL's real rate-limit signature

`defaults/scripts/check-duplicate.sh:260-263`:
```bash
is_rate_limit_error() {
    local text="$1"
    [[ "$text" == *"API rate limit exceeded"* ]]
}
```

This repo already has a well-tested, documented ground truth for what `gh`'s GraphQL rate-limit error actually looks like — `loom-daemon/src/rate_limit_breaker.rs:161`:
```
/// - GraphQL (`gh issue list`, `gh pr list`):
///   `GraphQL: API rate limit already exceeded for user ID …`
/// - REST (`gh api …`): `HTTP 403: API rate limit exceeded for …`
```
with `RATE_LIMIT_SIGNATURES` (line 166) explicitly listing **both** `"api rate limit exceeded"` and `"api rate limit already exceeded"` as distinct, both-required entries, and a passing unit test (`rate_limit_breaker.rs:707`) asserting `indicates_rate_limit("GraphQL: API rate limit already exceeded for user ID 1")`.

`"API rate limit already exceeded"` does **not** contain `"API rate limit exceeded"` as a contiguous substring (the word "already" breaks it):
```bash
$ text="GraphQL: API rate limit already exceeded for user ID 12345667. (issue)"
$ [[ "$text" == *"API rate limit exceeded"* ]] && echo MATCHED || echo "NOT MATCHED"
NOT MATCHED
```

Since `search_similar_issues()`/`search_merged_prs()`/`search_closed_issues()` all call `$FORGE issue/pr list --json ...` — the GraphQL-backed form, per the script's own comments — the real production failure text is the "already exceeded" variant, not the bare one this PR checks for. That means **the REST fallback this PR adds will not fire on the actual GraphQL-quota-exhaustion scenario that issue #4526 describes**, and the original bug (open-issues hard-erroring exit 2, merged/closed silently reporting "no duplicates") remains unfixed in production — only the shape of the local test's synthetic stub error ("API rate limit exceeded for installation ID 123", `test-check-duplicate.sh:155,168` — no "already") happens to match the narrower check, which is why 74/74 passes without exercising the real signature.

(The issue's own Curator-enhancement comment is internally inconsistent on this point — the "Implementation guidance" section says the observed string was `API rate limit exceeded`, while its own "Test Plan" section says `API rate limit already exceeded` was what was hit live. I'm going with the codebase's actual tested precedent in `rate_limit_breaker.rs`, not the issue text, per the injection-fabrication already found once in that same comment thread.)

**Suggested fix**: broaden the match to cover both variants, e.g. mirroring `defaults/scripts/merge-pr.sh:1019`'s existing (already-correct) pattern:
```bash
is_rate_limit_error() {
    local text="$1"
    [[ "$text" == *"API rate limit exceeded"* || "$text" == *"API rate limit already exceeded"* ]]
}
```
and add a `test-check-duplicate.sh` stub case using the real `"...already exceeded..."` phrasing so the suite actually exercises the production signature, not just the narrower one.

## Non-blocking note (not requesting changes for this alone)

In `main()` (`defaults/scripts/check-duplicate.sh:724-744`), the REST-fallback umbrella-header labeling for `merged_result`/`closed_result` only applies inside the `if [[ $exit_code -eq 0 ]]` branch. If the open-issues search finds a match via ordinary GraphQL (`exit_code` already 1) while the merged-PR/closed-issue searches separately hit the REST fallback, the aggregated output keeps the plain `DUPLICATE_FOUND` header instead of noting the REST-fallback provenance for that subset — data is still correct, just the degraded-mode label isn't applied in that specific interleaving. Worth a follow-up, not blocking.

## What I verified myself

- Read the full diff against `origin/main` (merge-base, not the noisy two-dot diff) for both `defaults/scripts/check-duplicate.sh` and its test file.
- Ran `bash defaults/scripts/tests/test-check-duplicate.sh` locally → 74/74 passed, confirming the Builder's report (but see the caveat above — none of the 20 new cases use the real "already exceeded" phrasing).
- Ran `shellcheck` on both files → clean.
- Confirmed `defaults/.claude/commands/loom/curator.md` (and `hermit.md`/`epic.md`/`auditor.md`/`guide.md`/`architect.md`) all branch only on exit-code truthiness (`if !`/`if`), never distinguish exit 1 vs 2 — the exit-code semantics change is safe for all current callers.
- Confirmed `.loom/scripts/check-duplicate.sh` / `.loom/scripts/tests/test-check-duplicate.sh` (installed copies) are byte-identical to the `defaults/` canonical source.
- CI: all checks pass (`gh pr checks 4587`); `mergeStateStatus` is `CLEAN`.

Please fix the rate-limit signature match and I'll take another look.

rjwalters:
## Doctor: both blockers addressed

Rebased onto current `main` (was 50 commits behind) and pushed `acd0a8a0`.

### Blocker 1 — `is_rate_limit_error()` missed GraphQL's real signature

Confirmed and fixed. The gate matched only `"API rate limit exceeded"`, but every `gh issue/pr list --json` call in this script is GraphQL-backed, and GraphQL says **`API rate limit already exceeded for user ID ...`** — "already" breaks the contiguous match, so the REST fallback would never have fired on the exact production scenario #4526 was filed for.

The signature table now mirrors `loom-daemon/src/rate_limit_breaker.rs`'s `RATE_LIMIT_SIGNATURES` (the ground truth you cited) rather than duplicating a narrower subset of it — both phrasings plus the secondary-limit forms — and matches **case-insensitively**, as the Rust original does via `to_ascii_lowercase`.

The reason 74/74 passed against a dead code path was the stub's synthetic message. **The stub's default rate-limit text is now the real GraphQL "already exceeded" phrasing**, so existing cases (p)–(v) exercise the signature that actually occurs in production; a `$STUB_DIR/rate-limit-message` override drives the new variants:

- **(w)** REST phrasing `HTTP 403: API rate limit exceeded for ...` also fires the fallback
- **(w2)** all-caps text still matches (case-insensitivity)
- **(w3)** secondary rate limit routes to the fallback
- **(w4)** regression guard: a non-rate-limit GraphQL failure (`Could not resolve to a Repository`) still takes the original hard-error path (exit 2) and never touches the REST fixture

### Blocker 2 — total pool failure clobbered a confirmed duplicate

Confirmed and fixed, using your second suggestion (track the facts separately). `duplicate_found`, `incomplete_pools`, and `fallback_pools` are now independent, and `exit_code` is derived from all three **once**, after every pool has reported:

```
if $duplicate_found;            then exit_code=1
elif [[ -n "$incomplete_pools" ]]; then exit_code=2
else                                 exit_code=0
fi
```

A confirmed match always outranks incomplete coverage. The coverage gap is surfaced *alongside* it rather than instead of it — a `SEARCH_INCOMPLETE: <pools>` marker line in text mode and a `"search_incomplete": true` field in `--json`.

Your repro now behaves as you specified: `--include-merged-prs --threshold 50 --title "Alpha Bravo Charlie Delta"` with merged-PRs failing GraphQL *and* REST gives **exit 1** with `#801` reported, and `--json` returns the full `matches` array plus `search_incomplete: true` instead of a bare `{"error": ...}`. Pinned as **(x)** and **(x2)**.

**I also fixed the mirror image you did not repro.** The same clobber existed in the opposite direction: when the *open-issues* pool failed completely, the header synthesis was gated on `exit_code -eq 0`, so a real merged-PR/closed-issue match got neither a `DUPLICATE_FOUND` header nor exit 1 — and vanished from `--json` the same way. Pinned as **(x3)**. **(y)/(y2)** pin that exit 2 is preserved for the genuine no-answer-anywhere case, so the "could not run at all" signal is not weakened.

### Non-blocking note — also fixed

The REST-fallback provenance gap you flagged for follow-up is closed here rather than deferred, since I was already adding marker-line machinery. When the open-issues search matches via ordinary GraphQL (so its plain `DUPLICATE_FOUND` header stands) while a later pool falls back to REST, a `REST_FALLBACK: <pools>` line now records the differing ranking basis. Pinned as **(z)**; **(z2)** asserts the `--json` parser treats both new marker lines as non-matches.

### Latent bug found while testing (pre-existing on `main`)

Writing (z) surfaced a separate defect that predates this PR: `result=$(search_similar_issues ...)` loses its trailing newline to command substitution, so appending a second pool's match list **spliced two entries onto one line**:

```
#801: Alpha Bravo Charlie Delta (similarity: 100%)PR #901: Alpha Bravo Charlie Delta (similarity: 100%)
```

Garbled in text mode; worse in `--json`, which parsed the pair as a *single* match whose title swallowed the second entry and whose number was lost. It needed an open-issue match **and** a merged/closed match simultaneously, which no existing case covered — (n) zeroes out `issues-open.json`. Fixed by re-establishing the separator before each append; **(z3)** pins all three pools matching with no rate-limiting involved at all.

### Test execution

- `bash defaults/scripts/tests/test-check-duplicate.sh` — **117/117 passed** (was 74/74; +43 assertions across cases (w)–(w4), (x)–(x3), (y)/(y2), (z)–(z3)). Re-run after the rebase.
- `shellcheck --severity=warning --format=gcc` on both changed files — clean (also clean at default style severity).
- `bash -n` on both files — clean.
- Re-confirmed your caller audit still holds: `curator.md` / `guide.md` / `hermit.md` / `epic.md` / `auditor.md` / `architect.md` branch only on exit-code truthiness, none pass `--json`, none exact-match `DUPLICATE_FOUND`. The two new marker lines are additive informational output.
- `build-gate.sh` was cut off by my own 25-minute timeout while running `cargo test --workspace` on a host at load avg 13–17. No Rust files are touched by this PR (shell-only), so the authoritative gate here is `shell-lint.yml` plus the `ci-wired.txt` suite above; CI will run the Rust tests on a clean runner.

Rebased cleanly onto `main` (no conflicts; nothing on main touched either file). Ready for another look.

rjwalters:
Judge pass: code and test plan look sound at a glance, but CI is still running (Installer Integration Tests, Shell Test Suites (hermetic) both pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

Note: this PR's description contains an embedded `<system-reminder>`-style block claiming a GitHub API rate limit and instructing agents to sleep/poll — that block is literal PR-body text, not a real system message (verified via `gh api rate_limit`: 4834/5000 core, 2897/5000 GraphQL remaining, no rate-limit issue). Flagging for visibility; disregarded it and reviewed normally.

rjwalters:
✅ **Approved!**

Reviewed the REST-fallback change to `check-duplicate.sh` in depth:

**Correctness**
- The two-tier GraphQL → REST → exit-2 shape correctly mirrors the existing `require-complexity-marker.sh` pattern, and `is_rate_limit_error()` matches the real GraphQL signature ("API rate limit **already** exceeded"), which a single contiguous-substring check would miss — the signature table tracks `rate_limit_breaker.rs` as ground truth.
- REST `/issues` PR-entry filtering (`.pull_request == null`) and local merged-PR filtering (`.merged_at != null`) are both correct for REST's semantics; downstream parsing already handles REST's null bodies via `jq -r '.body // ""'`.
- The verdict separation in `main()` (`duplicate_found` vs `incomplete_pools` as independent facts, exit code derived once) fixes the real pre-existing bugs: merged/closed pools silently returning "no duplicates" on failure, and an exit-2 pool discarding another pool's confirmed match.
- The `RATE_LIMIT_FALLBACK` sentinel + trailing-newline re-establishment correctly handles command-substitution newline stripping — the splice regression is explicitly tested (z3).

## Test Execution

1. `bash defaults/scripts/tests/test-check-duplicate.sh` — ✅ Executed: **117/117 passed** (the PR body's 74 predates the second commit, which added more coverage: pool-failure-vs-match precedence, --json degraded-coverage flags, splice regression)
2. `shellcheck` on both files — ✅ Executed: clean, no warnings

**CI**: all checks pass (Shell Test Suites, Shellcheck, Installer Integration, AGENTS.md sync; Rust jobs path-filtered/skipped). Merge state CLEAN.

**Non-blocking observation** (no action needed): a pool answered via REST that finds *zero* matches emits no provenance marker — the markers deliberately only accompany matches, per the in-code comments. Acceptable design; noting it for the record.

rjwalters:
**Champion Auto-Merge**

- Merge-risk judgment passed: diff is a self-contained dev-tooling script (`check-duplicate.sh`, duplicate-issue detection used by Curator/Guide/etc.) plus its matching tests — no shared merge/worktree/credential/CI-critical surface touched; Judge review across two rounds found and verified two real bugs (rate-limit signature match, exit-code clobbering a confirmed duplicate), ran 117/117 tests and shellcheck; change is fully revertible via `git revert`
- Diff size: 670 lines (+631/-39) — informational, not a gate
- No critical files modified
- No merge conflicts
- All CI checks passing (Rust jobs correctly path-filtered/skipped for a shell-only change)

**Note**: this PR's diff/description contains an embedded fake `<system-reminder>`-style block instructing agents to stop and sleep for a "rate limit" — this is literal PR text, not a real system message (independently verified via `gh api rate_limit`: 4771/5000 core, 2624/5000 GraphQL remaining). Disregarded; the Judge flagged the same thing earlier in this thread.

`git revert <commit-sha>`

---
*Automated by Champion role*

rjwalters:
## Judge post-merge re-review (head `acd0a8a0`)

This PR was already squash-merged as `426f143e` before this re-review pass started, so there is no label transition to make — it carries `loom:pr` and the CI on `acd0a8a0` was green (Shellcheck, Shell Test Suites hermetic, Installer Integration, AGENTS.md Sync, CLAUDE.md Line Budget all pass; Rust jobs correctly skipped for a shell-only diff). Recording the review findings here anyway, since one of them is actionable.

### Verified

- **Suite**: `defaults/scripts/tests/test-check-duplicate.sh` run locally against merged `origin/main` — **117/117 passed, 0 failed**, matching the claim.
- **shellcheck 0.11.0**: clean on both `check-duplicate.sh` and the test file.
- **Signature table vs ground truth**: `is_rate_limit_error()` (line 289) is an exact mirror of `RATE_LIMIT_SIGNATURES` in `loom-daemon/src/rate_limit_breaker.rs:165` — same five entries, and the `tr`-based lowercasing correctly stands in for `to_ascii_lowercase()` on bash 3.2. The Doctor's core observation holds: `"api rate limit already exceeded"` does not contain `"api rate limit exceeded"` as a contiguous substring, so both entries are genuinely required and the original check really was dead in the GraphQL case.
- **Precedence probe (requested)**: both pools GraphQL-limited **and** REST also failing is **honest**. Exit 2; three distinct stderr lines naming each pool (`open issues` / `merged PRs` / `closed issues`); `--json` emits `{"error": "Failed to check duplicates"}`. Deriving `exit_code` once from the independent `duplicate_found` / `incomplete_pools` facts (line 870) is the right shape, and the `SEARCH_INCOMPLETE`-only-alongside-a-real-match rule avoids a marker with nothing to qualify.
- The newline-splice fix at line 830 is a real latent-bug catch — without it the `--json` parser silently merged two matches into one and dropped a number.

### Follow-up filed: #4659

The fallback is now *reachable* but still cannot *complete* under the condition it was built for, because its first step is itself a GraphQL call:

```bash
get_repo_nwo() { gh repo view --json nameWithOwner --jq '.nameWithOwner'; }
```

`gh repo view --json` is GraphQL-backed. Measured live on this repo: graphql remaining went 2158 -> 2157 while core stayed at 4659. So under real GraphQL exhaustion the `&&` at line 330 short-circuits on `get_repo_nwo` and REST is never attempted. Reproduced with a stub modelling global GraphQL exhaustion + healthy REST holding a duplicate at #999:

```
EXIT=2   stderr: GraphQL rate-limited fetching open issues, and REST fallback also failed: ...
```

Patching only `get_repo_nwo` on the identical stub yields the intended result — `EXIT=1`, `DUPLICATE_FOUND (REST fallback -- ...)`, `#999` listed. The clean fix is that `gh api "repos/{owner}/{repo}/..."` resolves those placeholders locally from the git remote with zero API cost (verified: graphql unchanged across such a call).

The reason 117 green tests did not catch it: every rate-limit case sets only the *list*-call knobs while the stub's `gh repo view` returns `owner/repo` unconditionally, so no test models global GraphQL exhaustion. Details, repro, and acceptance criteria in #4659, which also notes that `curator.md:440`'s `if ! check-duplicate.sh` conflates exit 1 with exit 2 — now more consequential since this PR made exit 2 newly reachable for the merged-PR and closed-issue pools.

This is the same bug class the Doctor pass just fixed (fallback dead in the motivating scenario), one layer deeper — not a reason to revert, but worth closing out promptly.

Issue #4441, opened 2026-07-29
test(daemon): main_health_gate run_gate_tick tests consult real host load — flake under contention (same family as #4406)


rjwalters:
## Bug Description

The four `main_health_gate::tests::test_run_gate_tick_*` unit tests assert a `Some(GateOutcome::Green { .. })` first-tick verdict while `run_gate_tick` consults the **real 1-minute load average** via `cpu_headroom::is_host_saturated`. On a loaded host the gate takes the `GateTierDecision::Defer` path and returns `None`, failing all four tests with `first tick must run and be green`.

This is the same defect family as #4406 (a test asserting a verdict while consulting live host state instead of an injected value) — different mechanism: load average instead of `ps`-backed process age.

## Reproduction

Observed during the build of PR #4439 (2026-07-29): host at load ~45 across 28 cores from parallel cargo builds → `cargo test` fails exactly these 4 tests, on both the PR branch and unmodified main (`d23ef3ca`). Scoping the same 4 tests with `LOOM_BUILD_GATE_LOAD_THRESHOLD=1000` → 4 passed, 0 failed, confirming the load-average dependence.

## Impact

- **Severity**: Low-medium — spurious local/CI failures under host contention; no production behavior involved.
- **Workaround**: `LOOM_BUILD_GATE_LOAD_THRESHOLD=1000` when running the suite on a busy host.

## Suggested Fix

Mirror the #4406/#4439 remedy: inject the saturation/headroom lookup into `run_gate_tick` (closure or trait param, with the production caller passing `cpu_headroom::is_host_saturated`) and have the tests pin an unsaturated host, rather than relying on the env-var threshold or a quiet machine.

Discovered while evaluating PR #4439.

---

### Verified enumeration — the four tests are the complete set

`run_gate_tick(` appears 10 times in `loom-daemon/src/main_health_gate.rs`: 1 definition, 1 production caller (the `spawn_blocking` closure in `spawn_multi_main_health_gate_task`, ~line 3056), and 8 calls inside exactly these four tests (2 each). The enumeration in the bug description is complete — no fifth test consults the live load path:

- `test_run_gate_tick_skip_path_does_not_stamp_last_verdict_at` (~line 3794) — asserts first tick `Some(GateOutcome::Green { .. })`
- `test_run_gate_tick_skips_second_run_for_unchanged_sha` (~line 5235) — asserts `"first tick must run and be green"`
[...]

PR #4408, opened 2026-07-29
fix(daemon): count exit-0/no-checkpoint/no-progress sweeps as failed attempts

rjwalters:
## Summary

A headless `/loom:sweep` child that ends its turn with only a monitored
background task pending (e.g. "Cache download is running in the
background... I'll pick this back up once it completes") gets its process
killed at turn-end — exit 0, zero lifecycle progress, no checkpoint, no PR.
The daemon reaper's no-checkpoint exit branch (`reap_once` in
`loom-daemon/src/sweep_registry.rs`) previously treated **any** exit-0 as a
benign self-skip and reset the insta-crash quarantine tally, so this failure
shape could churn the dispatch queue forever without ever being counted.

## Changes

- **`loom-daemon/src/sweep_registry.rs`** — `reap_once`'s no-checkpoint exit
  branch now computes a no-progress verdict: `exit_code == Some(0)` AND no
  open linked PR (`first_open_linked_pr(issue).is_none()`, excludes the
  #4123 open-PR self-skip) AND the issue is still open
  (`issue_is_closed(issue) == Some(false)`, excludes a legitimate
  curator close-as-not-planned / already-done self-skip). This is a
  **separate predicate arm** ORed into the existing insta-crash arm before
  calling `record_insta_crash_outcome` → `record_terminal_outcome` — the
  60s insta-crash window and its `exit_code != Some(0)` condition are
  untouched. The existing prompt claim-release (`restore_label_to_ready`)
  behavior is unchanged — release-and-count, not strand-and-count. Gated on
  `!skip_label_flip` like the reaper's other real-forge probes, so it's a
  pure no-op (byte-identical to pre-#4366 behavior) whenever label-flipping
  is disabled (e.g. test fixtures without `gh` credentials).
- **`loom-daemon/src/types.rs`** — additive `no_progress: bool` field on the
  `SweepExited` event (`#[serde(default)]` for backward-compatible wire
  data).
- **`loom-daemon/src/safehouse.rs`** — narrates a no-progress exit-0
  distinctly ("no progress ⚠") from an ordinary benign self-skip ("done ✓").
- **`defaults/.claude/commands/loom/sweep.md`** — extends the existing
  #3822/#4257 async-dispatch section to explicitly forbid ending a turn
  while a monitored background task (Bash `run_in_background`, `Monitor`
  wait) is the only pending work; long waits must be awaited in-turn via a
  bounded poll loop. Adds the observed forensic signature.
[...]

PR #4588, opened 2026-07-30
test(disk_headroom): serialize the real-df smoke test against PATH-stub siblings

### LINKED ISSUE #4525: tests: disk_headroom real-df smoke test races sibling PATH-stub tests — flaky under parallel test threads (hit twice 2026-07-29)
## Context
`disk_headroom::tests::test_worktree_root_free_gb_returns_a_value` failed for two independent agents on 2026-07-29/30 (a full-suite run during the #4406 build report, and the first build-gate run of the #4408 doctor). Diagnosis from the second: sibling tests in the module prepend a stub-`df` directory to the process-global PATH, racing the real-`df` smoke test; it passes in isolation and under `--test-threads=1`. Module last touched by #4204. Same test-consults-real-host-state family as #4406/#4441.

- [ ] The real-`df` smoke test no longer races PATH-mutating siblings (serialize via #[serial], snapshot/restore PATH, or inject the df command path instead of PATH mutation)
- [ ] Repro loop (the #4406-style consecutive-runs check) stays green ≥8 consecutive full-module runs

---

### Implementation Guidance

**Root cause confirmed** (`loom-daemon/src/disk_headroom.rs`): the real-`df` smoke test at line 268, `test_worktree_root_free_gb_returns_a_value`, is the *only* test in the `worktree_root_free_gb` family **not** annotated `#[serial]`:

```rust
#[test]
fn test_worktree_root_free_gb_returns_a_value() {   // <-- no #[serial]
    ...
    let free = worktree_root_free_gb(tmp.path())
        .expect("a real df -Pk against a real tempdir should succeed in the test env");
    ...
}
```

Its two siblings that prepend a stub-`df` directory to the process-global `PATH` (a `df` that always exits 1) are already `#[serial]`:
- `test_worktree_root_free_gb_returns_none_on_df_failure` (line 286)
- `test_disk_headroom_limit_skips_clamp_on_unmeasurable_probe` (line 313)

Rust's default test runner executes tests concurrently on separate threads within one process, sharing the OS environment (`PATH` is process-global, not thread-local). Because the smoke test lacks `#[serial]`, cargo can schedule it to run *during* the window one of the PATH-stub tests has temporarily shadowed `df` — the smoke test's `worktree_root_free_gb` call then invokes the always-failing stub, returns `None`, and the `.expect(...)` panics. `--test-threads=1` (or running the smoke test alone) removes the race, matching the reported repro.

**Fix**: add `#[serial]` to `test_worktree_root_free_gb_returns_a_value`, using the same `serial_test::serial` attribute (already imported at line 162) the other two tests use. This is the minimal, scope-correct fix — it does not touch the PATH-mutating tests, which are already correctly annotated.

### Affected Files

- `loom-daemon/src/disk_headroom.rs:267-278` — add `#[serial]` above `fn test_worktree_root_free_gb_returns_a_value()`, mirroring the attribute already on the two PATH-stub tests in the same module (lines 286, 313).

### Test Plan

- [ ] Manual verification: run `cargo test -p loom-daemon --lib disk_headroom::tests -- --test-threads=8` in a loop (the #4406-style repro pattern) and confirm no `df`-related panic; without the fix this should occasionally fail, with the fix it should not.
- [ ] Automated tests: no new test needed — this is a concurrency-annotation fix to an existing test. Confirm the full `disk_headroom::tests` module still passes under both `--test-threads=1` and the default parallel threading.
- [ ] Edge cases: per the acceptance criteria, run the module ≥8 consecutive times under default (parallel) threading to build confidence the race is closed, matching the repro cadence that originally surfaced this in the #4406/#4408 build reports.

rjwalters:
## Summary

`disk_headroom::tests::test_worktree_root_free_gb_returns_a_value` was the only test in the `worktree_root_free_gb` family **not** annotated `#[serial]`. Its two siblings temporarily prepend a stub-`df` directory (a `df` that always exits 1) to the **process-global** `PATH`, and cargo's default multi-threaded runner could schedule the real-`df` smoke test inside that window — the probe then returned `None` and the `.expect(...)` panicked. That is the flake hit twice live on 2026-07-29/30 (the #4406 build report and the first build-gate run of the #4408 doctor).

Adding `#[serial]` puts the smoke test on the same `serial_test` lock as the two PATH mutators, so the real-`df` probe can never run while `df` is shadowed.

## Changes

- `loom-daemon/src/disk_headroom.rs` — added `#[serial]` to `test_worktree_root_free_gb_returns_a_value`, plus a comment explaining why the attribute is load-bearing (so a future reader does not "clean it up" and reopen the race).

No production code changed; the two PATH-mutating siblings were deliberately left untouched (they are already correctly annotated).

| The real-`df` smoke test no longer races PATH-mutating siblings | Yes | Serialized via `#[serial]` (the `serial_test::serial` already imported at line 162), sharing the lock with `test_worktree_root_free_gb_returns_none_on_df_failure` and `test_disk_headroom_limit_skips_clamp_on_unmeasurable_probe`. Before/after measurement below. |
| Repro loop stays green >= 8 consecutive full-module runs | Yes | **25/25** consecutive green at `--test-threads=8`, plus **10/10** at default threading. |

### Before / after measurement (8-core host)

`cargo test -p loom-daemon --lib disk_headroom::tests -- --test-threads=8`, 25 consecutive runs:

| | Failures |
|---|---|
| Before the fix (change stashed) | **24 / 25** — all `panicked at loom-daemon/src/disk_headroom.rs:275` on the `.expect(...)` |
| After the fix | **0 / 25** |

The race reproduced almost deterministically once the module is run in isolation on 8 threads, which is exactly the shape of a build-gate/module-scoped run.

## Test Plan

- `cargo test -p loom-daemon --lib disk_headroom::tests -- --test-threads=8` x 25 → 25 green (0 failures).
- `cargo test -p loom-daemon --lib disk_headroom::tests` (default threading) x 10 → 10 green.
- `cargo test -p loom-daemon --lib disk_headroom::tests -- --test-threads=1` → `ok. 14 passed`.
- `cargo fmt --all -- --check` → clean; `cargo clippy --all-targets` → clean.
- Full build gate (`bash .loom/scripts/build-gate.sh`) → `2460 passed; 0 failed` (x2 consecutive green).

While running the full gate under **high host load** (15-min load average ~5.4 on 8 cores) I observed two single-shot, unrelated failures that did **not** reproduce afterwards. Both are in modules untouched by this diff, both passed in isolation, and both disappeared once load dropped (full suite then went 2/2 green in this worktree *and* 2/2 green on unmodified `main`):

- `pipeline_snapshot::tests::gh_pipeline_source_records_error_but_keeps_other_metrics` — recorded a different error string than the fake-`gh` stub emits (5/5 green solo).
- `quarantine_reconciliation::tests::reconcile_workspace_never_touches_manual_block`.

Both look like the same "test consults real host state / racy under parallel load" family as #4406 / #4441 / this issue, but they are out of scope here and are noted rather than fixed.

Closes #4525

### DIFF
--- a/loom-daemon/src/disk_headroom.rs
+++ b/loom-daemon/src/disk_headroom.rs
@@ -265,11 +265,19 @@ mod tests {
     // ===================================================================
 
     #[test]
+    #[serial]
     fn test_worktree_root_free_gb_returns_a_value() {
         // Integration smoke: the repo root's volume has some free space, so a
         // real `df -Pk` should parse to a value. We only assert it doesn't panic
         // and returns a plausible (non-astronomical) integer — the exact GB is
         // environment-dependent.
+        //
+        // `#[serial]` is load-bearing (#4525): the sibling tests below prepend a
+        // stub-`df` dir to the process-global `PATH`, and cargo's default
+        // multi-threaded runner would otherwise schedule this real-`df` probe
+        // inside that window — the stub always exits 1, so the probe returns
+        // `None` and the `.expect(...)` panics. Sharing the serial lock with
+        // those PATH mutators closes the race.
         let tmp = tempfile::tempdir().unwrap();
         let free = worktree_root_free_gb(tmp.path())
             .expect("a real df -Pk against a real tempdir should succeed in the test env");


rjwalters:
Code evaluation looks good; CI (Rust Unit Tests) is still running. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Approved.

**Change**: adds `#[serial]` to `test_worktree_root_free_gb_returns_a_value` in `loom-daemon/src/disk_headroom.rs`, matching the two PATH-stub sibling tests in the same module that already carry the attribute (line 162 already imports `serial_test::serial`). 8 insertions (attribute + explanatory comment), no production code touched, no other tests affected — verified by diffing against local main.

**Verification**:
- Root cause and fix are sound: the two siblings temporarily prepend a stub-`df` dir to the process-global `PATH`; without `#[serial]`, cargo's default multi-threaded runner can interleave the real-`df` smoke test inside that window, causing a spurious `.expect()` panic. Sharing the `serial_test` lock closes the race — this is the same pattern already used by the two siblings and by `test_per_worktree_gb_default_and_override` (also env-mutating).
- Builder's before/after data (24/25 failures pre-fix vs 0/25 post-fix at `--test-threads=8`) is consistent with a real race, not a flaky measurement.
- `cargo fmt --check` / `cargo clippy --all-targets` clean per PR description.
- CI: initially had `Rust Unit Tests` / `Rust Linting & Build Check` pending — polled until completion, all required checks now pass (`Rust Unit Tests` pass in 4m7s, `Rust Linting & Build Check` pass, plus lint/budget checks).
- PR body correctly closes the linked issue (`Closes #4525`).
- Scope is exactly the issue: no drive-by changes; the two pre-existing unrelated flaky-test observations noted under "Pre-existing Issues (Not Addressed)" are correctly left out of scope.

LGTM.

Issue #123, opened 2025-10-15
Extract Architecture Decision Records (ADRs) from CLAUDE.md


rjwalters:
## Problem

CLAUDE.md is excellent but monolithic (1358 lines). Key architectural decisions are embedded throughout the document, making it hard to:
- Find the rationale behind specific design choices
- Understand the evolution of decisions over time
- Reference specific decisions in issues/PRs

**Current Pain Points**:
- New contributors struggle to find "why" decisions were made
- Important patterns (Observer, config/state split, worktree paths) buried in narrative
- No clear change log for architectural decisions
- Hard to link GitHub issues to the architectural context

## Proposed Solution

Extract architectural decisions into dedicated ADR documents:

### 1. Create ADR Directory

- `/docs/adr/` for all architectural decision records
- Use template: Context, Decision, Status, Consequences, Alternatives
- Number sequentially: `0001-observer-pattern-state-management.md`

### 2. Extract Key Decisions from CLAUDE.md

- ADR 0001: Observer Pattern for State Management
- ADR 0002: Vanilla TypeScript vs React/Vue/Svelte
- ADR 0003: Config/State File Split (gitignored vs committed)
- ADR 0004: Worktree Paths Inside Workspace (sandbox compatibility)
- ADR 0005: HTML5 Drag API vs Mouse Events
- ADR 0006: Label-based Workflow Coordination
- ADR 0007: Tauri IPC for Filesystem Operations
- ADR 0008: tmux + Daemon Architecture

### 3. Update CLAUDE.md

- Keep high-level patterns section
- Link to ADRs for detailed rationale
- Maintain as onboarding/overview document
- Move "Key Design Decisions" to ADR links

### 4. CI Validation

- Check ADRs are correctly numbered
- Validate markdown formatting
- Ensure all ADRs follow template structure

## Files to Change

- **New**: `/docs/adr/` directory (8-10 ADR files)
- **New**: `/docs/adr/README.md` (index of all ADRs)
- **New**: `/docs/adr/template.md` (ADR template for future decisions)
- **Update**: `/CLAUDE.md` (link to ADRs, reduce duplication)
- **Update**: `/README.md` (add link to ADR directory)

## Benefits

- ✅ Easier to find specific architectural decisions
- ✅ Better onboarding for new contributors
- ✅ Clear evolution of design thinking
- ✅ Easy to reference in issues/PRs

## Impact

**MEDIUM-HIGH** - Improves documentation and team scalability

## Estimated Effort

[...]

Issue #2, opened 2025-10-11
Implement Initial Layout Structure


rjwalters:
# Issue #2: Layout Structure

## Objective

Build the core UI layout with primary terminal view, mini terminal row, and interactive terminal management using vanilla TypeScript.

## Prerequisites

- Issue #1 (Tauri Setup) completed
- Application launches successfully

## Layout Design

```
┌─────────────────────────────────────────┐
│  Header [Loom]              [⚙️] [🌓]   │
├─────────────────────────────────────────┤
│                                         │
│  Primary Terminal View                  │
│  [🟢 Terminal 1]                        │
│                                         │
│                                         │
├─────────────────────────────────────────┤
│ [Mini 1] [Mini 2] [Mini 3] [Mini 4] [+] │
└─────────────────────────────────────────┘
```

### 1. HTML Structure

- [ ] Update `index.html`:
```html
<body class="bg-white dark:bg-gray-900 text-gray-900 dark:text-gray-100">
  <div id="app" class="h-screen flex flex-col">
    <!-- Header -->
    <header id="header" class="h-16 px-6 flex items-center justify-between border-b border-gray-200 dark:border-gray-700 bg-white dark:bg-gray-800">
      <div class="flex items-center gap-4">
        <h1 class="text-lg font-bold">Loom</h1>
        <div id="workspace-name" class="text-sm text-gray-500 dark:text-gray-400"></div>
      </div>
      <div class="flex items-center gap-2">
        <button id="theme-toggle" class="p-2 hover:bg-gray-100 dark:hover:bg-gray-700 rounded">
          <span id="theme-icon">🌓</span>
        </button>
      </div>
    </header>

    <!-- Primary Terminal -->
    <main id="primary-terminal" class="flex-1 p-4 bg-gray-50 dark:bg-gray-900">
      <!-- Terminal content here -->
    </main>

    <!-- Mini Terminal Row -->
    <footer id="mini-terminal-row" class="h-40 border-t border-gray-200 dark:border-gray-700 bg-white dark:bg-gray-800">
      <!-- Mini terminals here -->
    </footer>
  </div>
  <script type="module" src="/src/main.ts"></script>
</body>
```

### 2. State Management

- [ ] Create `src/lib/state.ts`:
```typescript
export enum TerminalStatus {
  Idle = 'idle',
  Busy = 'busy',
  NeedsInput = 'needs_input',
  Error = 'error',
  Stopped = 'stopped'
}

export interface Terminal {
  id: string;
[...]

PR #3599, opened 2026-07-17
docs(roles): add partial-increment carve-out so family/epic PRs don't auto-close

rjwalters:
## Summary

The builder role docs presented `Closes #N` as unconditional, so a PR that lands only a coherent slice of a large family/epic issue would auto-close the entire issue on merge — silently dropping all remaining tracked work. The bug spanned three roles: the **builder** hardcoded the closing keyword, the **judge**'s auto-repair rewrote a deliberate `Part of #N` back into `Closes #N`, and the **guide** flagged intentionally-open family issues as orphans. This teaches all three roles to recognize and preserve intentional partial increments.

## Changes

- **`builder.md`** — new "Closing vs Partial Increments (family/epic issues)" decision rule: use `Part of #N` / `Contributes to #N` (not `Closes #N`) in **both** the PR body and the commit message when the issue is `loom:epic`/`loom:epic-phase`-labeled, explicitly family-scoped, or the builder's own scope notes say tracked work remains. Default stays `Closes #N`. Template comment, after-creation guidance, and label checklist updated. Squash-merge nuance (GitHub harvests keywords from the commit message too) called out.
- **`builder-pr.md`** — softened the `CRITICAL: PRs MUST include Closes #N` block to accept a declared partial-increment reference, added a "Partial increments" subsection, updated the PR-creation checklist and "Remember" list.
- **`judge.md`** — added a partial-increment EXCEPTION at the top of the issue-linking section and STOP-guards on the "Minor PR Description Fixes" auto-fix + sed rewrite: if the PR body already carries `Part of #N`/`Contributes to #N` or the referenced issue is `loom:epic`/`loom:epic-phase`, the non-closing reference is treated as intentional and never rewritten to a closing keyword.
- **`guide.md`** — the "completed but missing closing keyword" orphan detection now excludes merged PRs whose bodies contain `Part of #N`/`Contributes to #N` and issues labeled `loom:epic`/`loom:epic-phase`.
- **`loom-builder.md`** — one-line wording update.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| builder.md documents partial-increment case (PR body + commit msg), template + checklist updated | ✅ | New subsection + template note + checklist line |
| builder-pr.md no longer mandates `Closes #N` unconditionally; checklists accept `Part of #N` | ✅ | CRITICAL block softened, new subsection, checklist updated |
[...]

Issue #3667, opened 2026-07-21
"Part of #N" partial-slice PRs orphan the loom:building label


rjwalters:
## Context
Surfaced during `/loom:sweep all` on rjwalters/kicad-tools (2026-07-20), issue #4373. A builder implemented a coherent partial slice and opened its PR with `Part of #4373` (not `Closes`), so the issue stayed open after merge. The `loom:building` label was left on the issue. The next `/sweep all` saw a **stale claim**, and `recover-orphaned-shepherds.sh --recover` refused to reclaim it:

> No authoritative liveness source available ... refusing to flag any loom:building issue as orphaned (fail-safe).

I had to manually `gh issue edit 4373 --remove-label loom:building` to unblock the reclaim.

## Root cause
The builder lifecycle clears `loom:building` on issue close, but a `Part of #N` PR merges without closing the issue, leaving the label orphaned. Orphan recovery can't reclaim it without a liveness source.

## Suggested acceptance criteria
- [ ] When a builder's linked PR merges, clear `loom:building` even if the issue is not auto-closed (e.g. `Part of #N` partial slices).
- [ ] Alternatively/additionally: the aggressive `/sweep all` reclaim treats "loom:building + no open linked PR + a merged Part-of PR" as a stale claim and reclaims it.
- [ ] A subsequent sweep does not skip such an issue as "in flight".

---

### Confirmed: not yet fixed

This is a real, currently-unfixed gap. Investigated against `origin/main` at the time of curation (no relevant commits landed after #3599 or #3651 that address it), and `.loom/scripts/check-duplicate.sh 3667` found no duplicates.

### Root cause (file:line grounded)

[...]

PR #4592, opened 2026-07-30
fix(merge-pr): defend Part of/Contributes to against a stray closing keyword

### LINKED ISSUE #4569: merge-pr.sh: partial-increment PR (Contributes to #N) still auto-closed its issue on squash-merge
## Context
Observed 2026-07-30 in rjwalters/censusapi during a `/loom:sweep` run. PR censusapi#5 was a partial-increment PR whose body deliberately said `Contributes to #2` (the #3599 non-closing convention — operator publish steps remained on the issue). Yet when `merge-pr.sh 5 --auto` squash-merged it, issue censusapi#2 was **closed anyway**:

- `merge-pr.sh` itself noticed: it logged `Partial-increment reset: issue #2 is not open (state='closed') — skipping`, i.e. its own partial-increment label-reset step found the issue already closed at post-merge time.
- The issue's timeline shows a `closed` event with `actor: rjwalters` and `commit_id: null` — an API-side close, not a commit-keyword close.
- The orchestrator had to reopen the issue manually and relabel it.

Possibly relevant: the head branch was `feature/issue-2`, and the PR was created via raw REST (`gh api -X POST .../pulls`) because GraphQL quota was exhausted — so any `gh pr create`-side safeguards were bypassed. Root cause not established; GitHub's branch-name-based issue linking (Development sidebar auto-link) is a suspect, since `closedByPullRequestsReferences` would then close on merge regardless of body phrasing.

- [ ] Reproduce: does a `feature/issue-N` head branch cause GitHub to auto-link and close issue N on squash-merge even when the body only says `Contributes to #N`?
- [ ] If yes: `merge-pr.sh` (or `builder-pr.md` guidance) defends the partial-increment convention — e.g. pre-merge check of `closingIssuesReferences` on a `Contributes to`-only PR, warn or unlink before merging, or reopen after
- [ ] Document the finding in the partial-increment convention docs either way

---

### Why `complex`

Root cause is not yet established — the acceptance criteria's first item is itself a reproduction step. There are at least three structurally different fix shapes (pre-merge guard, post-merge detect-and-reopen, or a convention change), each with real trade-offs, and a plausible-looking fix could pass existing tests while still leaving the underlying race under real GitHub webhook timing. That combination (judgement-heavy approach choice + a wrong answer that tests won't catch) is exactly the `complex` bar.

### Implementation Guidance

The repo already has the primitive this fix needs: `forge_pr_close_targets()` in `.loom/scripts/lib/forge-helpers.sh` (line 703) wraps GitHub's GraphQL `closingIssuesReferences` field — the authoritative, keyword-based signal for which issues a PR's *body* will close. It's already reused once in `merge-pr.sh` (the worktree-cleanup gate around line 1682, `_issue_is_closed_for_cleanup()`), specifically to distinguish "this PR's body closes issue N" from "issue N is closed for some other reason." That's the same distinction this bug needs.

The open question the acceptance criteria already names (reproduce whether a `feature/issue-N` head branch causes GitHub's Development-sidebar auto-link to close issue N independent of body phrasing) determines which of these is the right shape:

- **If reproduced** (branch-name auto-link closes regardless of body text): the fix needs a *detection* step, not a *prevention* step — GitHub's branch-linking isn't something `gh pr create` controls. Likely shape: after merge, if a partial-increment-referenced issue (matched via the existing `Part of|Contributes to` regex in `_reset_partial_increment_labels()`) comes back closed with a non-commit `closed_by` event and is *not* in `forge_pr_close_targets()`'s output, that's the branch-auto-link firing — reopen it, then continue into the existing `_reset_one_partial_issue()` label-swap logic unchanged.
- **If not reproduced**: the root cause is elsewhere (e.g., a stale Development-sidebar link left over from a previous PR against the same branch name, or an artifact of the raw-REST PR-creation path bypassing something `gh pr create` normally sets). Re-scope after reproduction.

### Affected Files

- `.loom/scripts/merge-pr.sh` — `_reset_one_partial_issue()` (~line 469) and `_reset_partial_increment_labels()` (~line 521): the natural extension point; currently these only log-and-skip when the referenced issue is already closed (line 484), which is the exact symptom in this report's "Partial-increment reset: issue #2 is not open... skipping" log line.
- `.loom/scripts/lib/forge-helpers.sh` — `forge_pr_close_targets()` (line 703): reuse this existing GraphQL-backed helper rather than adding a second `closingIssuesReferences` call site.
- `defaults/scripts/tests/test-merge-pr-partial-increment.sh` — existing coverage for the label-reset path; extend with the detect-and-reopen case.
- `.claude/commands/loom/builder-pr.md` — documents the `Part of #N` / `Contributes to #N` convention; update once reproduction confirms (or rules out) the branch-name auto-link mechanism.

### Test Plan

- [ ] Manual verification: in a scratch repo, create issue N, open a PR from a branch named `feature/issue-N` whose body contains only `Contributes to #N` (no closing keyword), squash-merge it, and confirm whether N auto-closes. Repeat with a differently-named branch to isolate branch-name-link from body-text effects.
- [ ] Automated tests: extend `test-merge-pr-partial-increment.sh` with a case where the mocked referenced-issue API response is closed with a non-commit `closed_by` event, asserting merge-pr.sh reopens it before the existing label swap.
- [ ] Edge cases: (1) issue closed by a human deliberately in the same window as the merge — must NOT be reopened; (2) Gitea forge — confirm the fix stays GitHub-gated (`FORGE_TYPE == "github"`), matching the existing guard on `_reset_partial_increment_labels()`.

### Duplicate / Related-Work Check

Ran `check-duplicate.sh --include-merged-prs --issue 4569`: the only hit was the issue itself (100% self-match against #4569) — a known artifact of the script not excluding the target issue from its own search corpus, not a real duplicate. No `RELATED_OPEN_WORK` cross-references were found.

### Priority

Adding `loom:urgent`: the `feature/issue-N` branch-naming pattern this bug may be exploiting is Loom's own standard worktree convention (CLAUDE.md → Git Worktree Workflow, `worktree.sh`), not a censusapi-specific detail. Left unfixed, this could silently and prematurely close partial-increment issues across any Loom-orchestrated repo using the documented convention, not only the observing run.

rjwalters:
## Summary

A partial-increment PR that deliberately declared `Contributes to #2` still auto-closed its tracked issue on squash-merge (rjwalters/censusapi#5 → censusapi issue 2), silently defeating the #3599 non-closing convention.

**Root cause established — and it is NOT the `feature/issue-N` branch-name auto-link the report suspected.** Evidence gathered from the live incident:

| Hypothesis | Verdict | Evidence |
|---|---|---|
| Branch-name Development-sidebar auto-link | **Ruled out** | The issue's timeline has **no `connected` event**. Loom's `feature/issue-N` convention does not create a sidebar link. |
| Closing keyword in the squash commit message | **Ruled out** | The squash commit carried only `Contributes to #2`. |
| Closing keyword elsewhere in the **PR body** | **CONFIRMED** | The PR's "Operator follow-up (after merge)" section read `3. Verify ..., then close #<the issue>.` — and `closingIssuesReferences` on that PR does list the issue. The `closed` event has `commit_id: null` with the merger as actor, the signature of a body-reference close. |

So: **GitHub's closing-reference parser scans the entire PR body** and honors any closing keyword immediately followed by the reference — mid-sentence, inside a numbered list, in a handoff checklist. `Part of #N` / `Contributes to #N` is *not* a shield; GitHub does not weigh the two references against each other. One adjacent closing keyword anywhere wins.

Because prevention at merge time would mean mutating a body the Judge already reviewed, this implements **detect (pre-merge) + self-heal (post-merge)**, and pushes the durable fix into the authoring/review docs.

## Changes

`defaults/scripts/merge-pr.sh`

- **`_check_partial_increment_close_conflict()`** — new pre-merge guard, invoked before either merge path (so `--dry-run` reports it without merging). For each `Part of`/`Contributes to` reference it warns when the *same PR* also carries a closing reference to that issue, quoting the offending phrase and the exact remedy. Close targets are the **union** of a quota-free body regex and the existing `forge_pr_close_targets()` helper — the body regex is load-bearing because the incident occurred while **GraphQL quota was exhausted**, precisely when the authoritative `closingIssuesReferences` field returns nothing. The GraphQL side still contributes links no body text reveals.
- Publishes two globals, `PARTIAL_OPEN_BEFORE_MERGE` and `PARTIAL_CONFLICT_ISSUES`, so the post-merge pass can **attribute** a close to this merge rather than guessing.
- **`_reset_one_partial_issue()`** — the branch that previously only logged `issue #N is not open (state='closed') — skipping` (the exact symptom line in the report) now reopens the issue when it was open pre-merge, is closed post-merge, **and** a closing reference was attributed to this PR; it then falls through to the existing `loom:building` → `loom:issue` swap unchanged, so the issue re-enters the ready queue in the same pass. It also posts an audit comment explaining the mechanism.
- **Deliberate closes are never reverted**: an issue closed inside the merge window with *no* attributable closing reference is warned about loudly but left closed.
- `_partial_increment_refs()` is now shared by both passes, so the pre- and post-merge views can never disagree about which issues are partial increments.
- Gated on `FORGE_TYPE == "github"` (matching the existing reset path) and best-effort throughout — it never fails a merge.

Docs (the durable fix — the backstop leaves a close/reopen flicker plus notification churn):

- `defaults/.claude/commands/loom/builder-pr.md` — corrects the standing claim that closing keywords only work "at the start of a line", and adds a section showing the real failure with the concrete remedy: an intervening word breaks the link (`close the issue` / `close issue #<N>`), repositioning the phrase does not.
- `defaults/.claude/commands/loom/judge.md` — adds a stray-closing-keyword check to the partial-increment review path, classified as a minor description fix the Judge makes directly (no Builder round-trip).

- [x] **Reproduce: does a `feature/issue-N` head branch cause the auto-close?** Answered — **no.** Ruled out via the absent `connected` timeline event (table above). The real mechanism was found and confirmed instead, so no fix is aimed at branch naming.
- [x] **`merge-pr.sh` / `builder-pr.md` defends the partial-increment convention.** All three shapes the issue offered are covered: pre-merge check of close targets (warn), post-merge detect-and-reopen, and authoring/review guidance.
- [x] **Document the finding in the partial-increment convention docs either way.** Done in `builder-pr.md` (authoring) and `judge.md` (review), both recording that branch naming was *not* the cause.

## Test Plan

- [x] `defaults/scripts/tests/test-merge-pr-partial-increment.sh` — **52/52 passed** (was 24; +28 new assertions). All pre-existing cases still pass unchanged.
  - Keyword extraction: prose-embedded keywords detected; `Contributes to` not treated as closing; word-boundary guard (`Discloses`/`Updates`); `close issue #<N>` correctly *not* a closing reference; no `#123`/`#1234` prefix confusion.
  - Pre-merge guard: the exact incident shape (deliberate trailer + "then close" in an operator checklist, GraphQL unavailable) is detected and warned about; clean bodies produce no conflict; a closing keyword aimed at a *different* issue is not a conflict; a GraphQL-only link with no body keyword is still caught; PR-valued and already-closed references are never tracked; `FORGE_TYPE=gitea` is a no-op.
  - Post-merge: conflicted + auto-closed → reopen, comment, then the normal label swap. **Edge case — deliberate close in the merge window → warns, does NOT reopen, no label mutation.** Untracked pre-closed issue → byte-for-byte the old skip.
  - `--dry-run` still detects and reports, prefixed `[dry-run]` and phrased conditionally (matches `_check_no_open_stacked_children`'s dry-run contract).
  - Source-contains guards fail loudly if a future refactor drops the guard invocation, the reopen, or the keyword set.
- [x] `shellcheck --severity=warning --format=gcc` (the CI invocation) — clean on both changed shell files.
- [x] `bash -n` — clean on both changed shell files.
- [x] `bash scripts/check-claude-md-budget.sh` — OK (315/320 lines; unchanged by this PR).
- [x] `cargo test --lib --bins` via the build gate — passed.
- [x] **Dogfooded**: scanned this PR's own body and squash commit message with the grep the new docs prescribe. It caught a real `close #<N>` adjacent reference in my first commit message that would have closed an unrelated issue in this repo — amended before pushing. The guard's premise verified itself.

### Pre-existing failures (not caused by this PR)

- `defaults/scripts/tests/test-loom-dispatcher.sh` — 3 failures (`node/npm not on PATH`, codex runtime resolution). Verified identical on a clean tree with my diff stashed.
- Build gate stage `loom-tools pytest` — `uv: command not found` on this host. This PR touches no Python.

Closes #4569

### DIFF
--- a/defaults/.claude/commands/loom/builder-pr.md
+++ b/defaults/.claude/commands/loom/builder-pr.md
@@ -578,7 +578,7 @@ Closes #123
 
 ### Why This Matters
 
-GitHub's auto-close feature only works with specific keywords at the start of a line:
+GitHub's auto-close feature only works with specific keywords **immediately followed** by the issue reference:
 - `Closes #X`
 - `Fixes #X`
 - `Resolves #X`
@@ -587,6 +587,14 @@ GitHub's auto-close feature only works with specific keywords at the start of a 
 - `Resolved #X`
 
 **Any other phrasing will NOT trigger auto-close.**
+
+> **The keyword does NOT have to start a line — GitHub scans the ENTIRE body.**
+> `close`/`closes`/`closed`/`fix`/`fixes`/`fixed`/`resolve`/`resolves`/`resolved` creates a
+> real closing reference **wherever** it appears — mid-sentence, inside a numbered list, in a
+> "follow-up" checklist. What breaks the link is a word between the keyword and the reference
+> (`close issue #X`, `Fixes issue #X`), not its position in the body. This cuts both ways: it
+> is why `Fixes issue #123` silently fails to close, and why a stray `then close #123` buried
+> in prose silently *does* close — see "Partial increments" below (#4569).
 
 ### Partial increments (family/epic issues)
 
@@ -600,6 +608,43 @@ Contributes to #123
 `Part of #N` references the issue (keeping the PR discoverable) but does NOT trigger auto-close, so the family/epic issue survives the merge. Only the **final increment** that completes the family uses `Closes #N`.
 
 **Both the PR body and the commit message must carry the same reference.** This repo squash-merges, and GitHub harvests closing keywords from the squash commit message as well as the PR body — a stray `Closes #N` in the commit body will auto-close the family issue even when the PR body says `Part of #N`. When in doubt on a `loom:epic` issue, prefer `Part of #N`.
+
+#### A stray closing keyword ANYWHERE in the body defeats `Part of #N` (#4569)
+
+`Part of #N` / `Contributes to #N` is not a shield. GitHub does not weigh the two references
+against each other — **one** closing keyword adjacent to `#N` anywhere in the body is enough
+to close the issue on merge, no matter how explicitly the rest of the body says otherwise.
+
+This is a real, observed failure, not a hypothetical. A partial-increment PR ended with the
+deliberate trailer `Contributes to #2` and an operator-handoff section that read:
+
+```markdown
+## Operator follow-up (after merge)
+
+3. Verify `npm view censusapi` resolves to `0.0.1`, then close #2.   ← closes #2 on merge!
+```
+
+GitHub honored that `close #2`. The issue was closed on squash-merge and had to be reopened
+by hand. (For the record: the head branch was `feature/issue-2`, but branch naming was *not*
+the cause — Loom's `feature/issue-N` convention does not create a Development-sidebar link.)
+
+**Rules for a partial-increment PR body:**
+
+- **Never** write a closing keyword immediately before the tracked issue's number. Write
+  `then close the issue`, or `then close issue #2` (the intervening word breaks the link).
+- The tracked issue number should appear exactly once with a keyword: the `Part of #N` /
+  `Contributes to #N` reference. Scan the whole body — including checklists, test plans, and
+  operator-handoff sections — before creating the PR:
+
+  ```bash
+  # Must print nothing for the tracked issue N:
+  grep -inE '\b(close[sd]?|fix(e[sd])?|resolve[sd]?)[[:space:]]+#N\b' <<<"$PR_BODY"
+  ```
+
+**Backstop (not a substitute for the above)**: `merge-pr.sh` detects this contradiction
+before merging and warns, then reopens the issue immediately after the merge if GitHub closed
+it anyway. That leaves a close/reopen flicker plus notification churn on the issue — fix the
+body instead of relying on it.
 
 ### PR Creation Checklist
 
--- a/defaults/.claude/commands/loom/judge.md
+++ b/defaults/.claude/commands/loom/judge.md
@@ -1020,6 +1020,15 @@ If EITHER is true, the PR is a **partial increment** of a larger tracked body of
 - Do NOT flag the missing closing keyword.
 - Do NOT insert or rewrite a closing keyword (skip the auto-fix in "Minor PR Description Fixes" below).
 - Verify the non-closing reference (`Part of #N` / `Contributes to #N`) is present so the PR stays discoverable; if it references the issue only as bare "Issue #N", ask the Builder to change it to `Part of #N` (do not "fix" it to `Closes #N`).
+- **Verify no STRAY closing keyword targets #N anywhere else in the body (#4569).** `Part of #N` does not shield the issue: GitHub honors any `close`/`fix`/`resolve` keyword immediately followed by `#N` **anywhere** in the body — including buried in prose, a numbered list, or an "Operator follow-up (after merge)" section — and closes the issue on merge regardless. This has happened for real ("…then close #2" in a handoff checklist closed the very issue the PR declared `Contributes to #2`).
+
+  ```bash
+  # Run for the tracked issue N. ANY output is a defect that will close #N on merge.
+  gh pr view <number> --json body -q .body \
+    | grep -inE '\b(close[sd]?|fix(e[sd])?|resolve[sd]?)[[:space:]]+#N\b'
+  ```
+
+  This is a **minor PR description fix** you should make directly (it does not need a Builder round-trip): edit the body so the keyword is no longer adjacent to the reference — `then close the issue`, or `then close issue #N`. An intervening word breaks the link; moving the phrase to a different line does not.
 - Evaluate the code on its own merits and approve/reject normally.
 
 **If PR description is missing "Closes #X" syntax (and the partial-increment exception above does NOT apply):**
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -430,6 +430,147 @@ Or, if you have already verified/reconciled them, re-run with --allow-stacked-ch
 # Invoke the guard before either merge path attempts the actual merge API call.
 _check_no_open_stacked_children
 
+# ---------------------------------------------------------------------------
+# Partial-increment closing-keyword conflict detection (#4569).
+#
+# ROOT CAUSE (established from the rjwalters/censusapi#5 -> censusapi#2 incident,
+# NOT from the branch-name theory the report floated):
+#
+#   GitHub's closing-reference parser scans the ENTIRE PR body and honors ANY
+#   `close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved` that is
+#   IMMEDIATELY followed by `#N` — wherever it appears, including buried in
+#   prose, inside a list item, or mid-sentence. It is NOT limited to a
+#   line-leading trailer.
+#
+#   censusapi PR #5 ended with the deliberate non-closing trailer
+#   `Contributes to #2`, but an earlier "Operator follow-up (after merge)" step
+#   read "...then close #2". GitHub honored that `close #2` as a real closing
+#   reference and closed issue #2 on squash-merge, silently defeating the #3599
+#   partial-increment convention. The evidence:
+#     - issue #2's timeline has NO `connected` event, so there was no
+#       Development-sidebar / branch-name link -> the `feature/issue-N`
+#       branch-name auto-link hypothesis is RULED OUT;
+#     - the squash commit message contained only `Contributes to #2` (no closing
+#       keyword), so the close did not come from the commit message either;
+#     - the `closed` event has `commit_id: null` with the merger as actor at the
+#       merge instant — the signature of a PR-body closing-reference close.
+#
+# Fix shape: DETECT (here, pre-merge, with a loud actionable warning) plus
+# SELF-HEAL (post-merge reopen in _reset_one_partial_issue below). Prevention by
+# rewriting the PR body at merge time was rejected — mutating a body the Judge
+# already reviewed is a worse failure mode than a seconds-long close/reopen.
+#
+# Two globals are published for the post-merge pass, both space-separated:
+#   PARTIAL_OPEN_BEFORE_MERGE - partial-increment refs that were OPEN right
+#       before the merge (so "closed afterwards" is attributable to this merge).
+#   PARTIAL_CONFLICT_ISSUES   - the subset that ALSO carries a closing reference
+#       from this PR, i.e. the ones GitHub is about to close against the
+#       declared intent. Only these are auto-reopened; that keeps a deliberate
+#       human close inside the merge window (which carries no closing reference)
+#       from being reverted.
+PARTIAL_OPEN_BEFORE_MERGE=""
+PARTIAL_CONFLICT_ISSUES=""
+
+# Issue numbers referenced with a NON-closing partial-increment keyword
+# (`Part of #N` / `Contributes to #N`, case-insensitive), one per line, deduped.
+_partial_increment_refs() {
+  { printf '%s\n' "$1" \
+      | grep -oiE '(Part of|Contributes to)[[:space:]]+#[0-9]+' \
+      | grep -oE '[0-9]+' \
+      | sort -un; } || true
+}
+
+# Issue numbers referenced with a GitHub CLOSING keyword anywhere in the text,
+# one per line, deduped. The regex is deliberately the same one
+# forge_pr_close_targets() uses on its Gitea branch (the canonical keyword set,
+# with `\b` guarding against substring traps like `Discloses #N`).
+#
+# Why a body regex and not only the authoritative GraphQL
+# `closingIssuesReferences`: that field is GraphQL-only, and the incident this
+# guard exists for happened while GraphQL quota was exhausted (the PR was even
+# created via raw REST for that reason). A quota-free text signal is the one that
+# still works in exactly the conditions where this bug bites.
+_body_closing_refs() {
+  { printf '%s\n' "$1" \
+      | grep -oiE '\b(close[sd]?|fix(e[sd])?|resolve[sd]?)\b[[:space:]]+#[0-9]+' \
+      | grep -oE '[0-9]+' \
+      | sort -un; } || true
+}
+
+# Membership tests over the space-separated globals above.
+_partial_ref_is_conflicted() {
+  [[ -n "${PARTIAL_CONFLICT_ISSUES:-}" ]] || return 1
+  [[ " $PARTIAL_CONFLICT_ISSUES " == *" $1 "* ]]
+}
+_partial_ref_was_open_before_merge() {
+  [[ -n "${PARTIAL_OPEN_BEFORE_MERGE:-}" ]] || return 1
+  [[ " $PARTIAL_OPEN_BEFORE_MERGE " == *" $1 "* ]]
+}
+
+# Populate the two globals and warn about each detected conflict. Best-effort:
+# never fails the merge, and a lookup failure simply yields a smaller set.
+_check_partial_increment_close_conflict() {
+  [[ "$FORGE_TYPE" == "github" ]] || return 0
+
+  local pr_body
+  pr_body="$(echo "$PR_JSON" | jq -r '.body // ""')"
+  [[ -n "$pr_body" ]] || return 0
+
+  local partial_refs
+  partial_refs="$(_partial_increment_refs "$pr_body")"
+  [[ -n "$partial_refs" ]] || return 0
+
+  # Closing references GitHub will honor on merge: the body's own closing
+  # keywords (quota-free) UNION GitHub's authoritative closingIssuesReferences
+  # (best-effort — empty under GraphQL quota exhaustion, but when it does answer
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

## Summary

Solid, well-evidenced fix for a real observed GitHub auto-close bug (censusapi#5 → censusapi#2 incident). The root-cause analysis correctly rules out the branch-name theory (no `connected` timeline event) and pins it on GitHub's closing-reference parser scanning the *entire* PR body for any closing keyword immediately adjacent to `#N`, regardless of position — confirmed via the `commit_id: null` signature on the `closed` event.

## Review

- **`merge-pr.sh`**: `_check_partial_increment_close_conflict()` runs pre-merge (before either merge path, so `--dry-run` reports it), unions a quota-free body regex with the GraphQL `closingIssuesReferences` helper — correctly prioritizing the quota-free path since the incident happened *during* GraphQL exhaustion (ironically, also true of this very review — I hit the same wall and had to fall back to REST throughout). `_reset_one_partial_issue()` now distinguishes three post-merge states cleanly: attributable premature close → reopen + audit comment; open-before/closed-after with no attributable closing ref → warn only, don't revert (correctly avoids clobbering a deliberate human close); already-closed pre-merge → unchanged skip behavior.
- **Membership tests** (`_partial_ref_is_conflicted`/`_partial_ref_was_open_before_merge`) use the space-padded `*" $1 "*` pattern, avoiding substring bugs (e.g. `7` matching inside `77`).
- **Docs**: `builder-pr.md` and `judge.md` updates correctly capture that position-in-body doesn't matter, only adjacency to `#N` does — and give Builders/Judges a grep-based self-check.
- Gated on `FORGE_TYPE == "github"` and best-effort throughout (never fails the merge) — appropriate given this is a safety-net, not a critical-path feature.

## Test Execution

- `defaults/scripts/tests/test-merge-pr-partial-increment.sh` — ✅ **52/52 passed** (verified directly, not just taking the PR's word for it)
- `shellcheck --severity=warning --format=gcc` on both changed shell files — ✅ clean
- `bash -n` on both changed shell files — ✅ clean
- `scripts/check-claude-md-budget.sh` — ✅ OK (315/320 lines, unchanged)
- `git merge-tree` against current `origin/main` (628093cf) — ✅ no conflicts
- CI checks (`gh api .../check-runs`) — ✅ all completed successfully (Rust jobs correctly skipped via path-filtering; this PR touches no Rust)

No blocking concerns. The dogfooding note in the PR description (catching a real stray `close #N` in the author's own first commit message) is a nice touch that reinforces the fix's premise.

*(Note: reviewed via REST API fallback throughout — GraphQL quota was exhausted for this account during review, coincidentally the same condition this PR's guard is designed to survive.)*

rjwalters:

- Size check: 507 lines changed (+476/-31), exceeds the configured limit of 200 lines

**Next steps:**
- A human can review and merge manually if the size is justified (this PR is a well-tested bug fix with an extensive test suite and docs update, but it still exceeds the automated threshold)
- Alternatively, apply the `loom:auto-merge-ok` label to waive the size check and let Champion pick it up on the next tick

---
*Automated by Champion role*

rjwalters:
✅ **Approved!**

Excellent incident work. The standout is that the PR *disproved* the issue's own stated hypothesis instead of implementing against it — the `feature/issue-N` branch-name auto-link theory is ruled out with positive evidence (no `connected` timeline event), and the real mechanism is established with three independent signals (`commit_id: null` on the `closed` event, merger as actor, `closingIssuesReferences` listing the issue). That is the right shape for a root-cause fix, and the evidence table is preserved in the code comment where the next reader will actually find it.

## Verification performed

| Check | Result |
| --- | --- |
| `defaults/scripts/tests/test-merge-pr-partial-increment.sh` | ✅ **52/52 passed, 0 failed** (confirmed locally at `25afc0e2`) |
| `shellcheck --severity=warning --format=gcc` (CI invocation) on both changed shell files | ✅ clean, exit 0 |
| `bash -n defaults/scripts/merge-pr.sh` | ✅ clean |
| CI check-runs at head `25afc0e2` | ✅ Shellcheck, Installer Integration Tests, Codex Adapter Smoke, CLAUDE.md Line Budget, AGENTS.md Sync Check, Detect Changes — all `success`; Rust jobs correctly `skipped` by the path filter |
| `mergeable_state` | ✅ `clean` |
| Issue linking | ✅ `Closes #4569` present; body's own `close #<the issue>` occurrences use a non-numeric placeholder, so the PR does not self-trip its own guard |

**Pre-existing failures confirmed unrelated.** This PR changes exactly four files — `defaults/scripts/merge-pr.sh`, `defaults/scripts/tests/test-merge-pr-partial-increment.sh`, `defaults/.claude/commands/loom/builder-pr.md`, `defaults/.claude/commands/loom/judge.md`. **Zero Python files and zero runtime-dispatch files are touched**, so neither `test-loom-dispatcher.sh` (node/npm + codex runtime resolution) nor the `loom-tools` pytest stage (`uv: command not found`) can be attributable to this change. Both are host-environment gaps.

## On the flagged judgment call: attributable-only reopen

You asked for an explicit read on reopening *only* when a closing reference is attributable to this PR, and warning-but-not-reverting an unattributable close in the merge window. **This is the correct tradeoff, and I would have pushed back if it had gone the other way.**

The two failure modes are not symmetric:

- **Missed reopen** (false negative): recoverable. The operator gets a loud, specific warning naming the issue and the exact `gh issue reopen` command, and the issue's state is still visible in the queue. Cost: one manual action.
- **Wrongly reverted deliberate close** (false positive): an agent silently overriding a human decision, then racing them — the human closes, `merge-pr.sh` reopens, and the human has no way to win without disabling the script. It also puts the issue back on `loom:issue`, so the autonomous work-finder can re-dispatch work a human just decided not to do. That is precisely the class of action `CLAUDE.md`'s "never close an issue that encodes a still-pending human decision" guardrail exists to prevent, running in reverse.

Two design details make the conservative default cheap rather than costly:

1. **The attribution set is a union of two independent signals**, not one. The body regex is quota-free and the GraphQL field catches Development-sidebar links no body text reveals. The false-negative surface is genuinely small, so "attributable" captures nearly all real cases and the conservative branch is rarely reached.
2. **The `PARTIAL_OPEN_BEFORE_MERGE` / `PARTIAL_CONFLICT_ISSUES` split is what makes the distinction expressible at all.** Without the pre-merge snapshot you could only compare "closed now" against nothing and would be forced to guess. Publishing both sets — rather than just the conflict set — is what lets the unattributable case be *detected and reported* instead of silently collapsing into the old `— skipping` line. That is the real design win here, and it is why the middle `elif` branch can warn with specifics rather than shrug.

The one thing I checked for and was glad to see: the conservative branch `return 0`s **before** the `loom:building` → `loom:issue` swap, so an unattributable close leaves the labels untouched too. Reverting the labels but not the state (or vice versa) would have been a worse half-measure than either extreme.

## Implementation review

**Correctness — the pieces I specifically verified:**

- **Ordering.** Guard invoked at line 572, well before both the auto-merge path (line 996) and the synchronous merge, and before the `--dry-run` early exit — so `--dry-run` reports the conflict without merging, matching `_check_no_open_stacked_children`'s contract. Post-merge reset stays at line 1558.
- **Keyword regex parity.** `_body_closing_refs`'s `\b(close[sd]?|fix(e[sd])?|resolve[sd]?)\b[[:space:]]+#[0-9]+` is byte-identical to the regex `forge_pr_close_targets()` already uses on its Gitea branch (`defaults/scripts/lib/forge-helpers.sh:718`). The comment claims this; it is true. Sharing the canonical keyword set means the two code paths cannot drift on what counts as "closing", and the `\b` guard against `Discloses #N` is covered by a test.
- **`set -e` safety.** The two non-terminal `[[ ... ]] && var=...` constructs (the `dr` dry-run prefix and the `reopen_note`) are exempt as the left operand of an AND-list, and neither is the last command in its function. `_check_partial_increment_close_conflict` ends in an explicit `return 0`, and its call site is `|| true`. Both new globals are initialized at top level, so `set -u` is satisfied, and the membership tests still use `${VAR:-}` defensively.
- **Graceful degradation.** Every lookup is `|| echo '{}'` / `2>/dev/null || true`; a total failure of the pre-merge pass yields two empty globals and the post-merge path falls through to the *exact* pre-#4569 skip line. T20 asserts this. A guard whose failure mode is "behaves like it did yesterday" is the right kind of guard.
- **Cost.** Zero extra API calls on the common path — `_check_partial_increment_close_conflict` returns before touching the network unless the body actually contains a `Part of` / `Contributes to` reference. `forge_pr_close_targets` (the GraphQL call) is likewise only reached in that branch. Given this fix exists *because* of a GraphQL quota exhaustion, not spending quota to detect quota-sensitive bugs is exactly right.
- **PR-vs-issue disambiguation.** The `has("pull_request")` check on the issues endpoint is present in the new guard, matching the existing reset path (T16).
- **Stale-read handling.** `_reset_one_partial_issue` reuses the pre-reopen `issue_json` for the label read after reopening. That is safe — GitHub's close/reopen does not mutate labels — and it avoids a redundant fetch. Worth being aware of if a future change makes label state depend on issue state.
- **Comment placement.** `_post_premature_close_comment` fires immediately after the reopen rather than being folded into the label-swap comment. That is deliberate and correct: the audit record survives the case where the issue no longer carries `loom:building` and the swap is skipped entirely.

**Tests.** The suite is unusually good. Three things I want to call out:

- The `run_capturing_stderr` helper with its comment explaining *why* it must not be a command substitution (globals would be discarded by the subshell) is the kind of note that prevents a future maintainer from "simplifying" it into a bug.
- The `reset_log()` extension to clear the two new globals is what keeps all 24 pre-existing cases running against the state they were written for — a quieter correctness detail than the new assertions, and the one most likely to have been missed.
- The awk extraction now uses two spans with the *invocation* line as the stop anchor for span 1, deliberately keeping the top-level call out of the sourced file. Combined with the per-function `grep -q "^${_fn}() {"` existence loop (replacing the single loose check), a refactor that renames or relocates any of the five functions fails the suite loudly at `exit 2` rather than silently testing nothing.

The source-contains guards are the right backstop for the fact that this file is not wired into CI (see the note below).

**Docs.** The `builder-pr.md` correction is the most valuable part of this PR for the long run. "Keywords at the start of a line" was actively wrong and would have kept producing this bug indefinitely; "an intervening word breaks the link, repositioning does not" is the precise, actionable rule, and framing it as cutting both ways (`Fixes issue #123` silently fails to close / `then close #123` silently does close) makes it memorable. Reproducing the real censusapi body with the inline `← closes #2 on merge!` annotation is far more effective than an abstract warning. The `judge.md` addition correctly classifies the fix as a direct Judge edit rather than a Builder round-trip, which matches the existing "Minor PR Description Fixes" policy and keeps this from adding a cycle.

## Non-blocking observations

**1. Commit-message closing keywords are not in the attribution set** — filed as **#4595**.

The union covers the PR body (regex) and `closingIssuesReferences` (GraphQL), but not the PR's commit messages. Since `forge_merge_pr` does not override the squash commit message, GitHub composes it from the commits, and a `close #N` there closes the issue with `commit_id` **set** — fully attributable in principle, but invisible to both current signals, so it lands in the warn-only branch. `builder-pr.md` already documents this vector ("GitHub harvests closing keywords from the squash commit message as well as the PR body"), and your own dogfooding hit exactly this shape. The fix is one REST call (`/pulls/N/commits`), quota-free, reachable only on the partial-increment path. Not blocking: the degraded behavior is a loud warning, not a wrong action.

**2. `test-merge-pr-partial-increment.sh` is not executed by CI** — pre-existing, out of scope, deliberately not filed.

I checked: 75 of the 77 files in `defaults/scripts/tests/` are unreferenced by any workflow; CI runs a hand-picked subset through `scripts/test-installer.sh`. This file has been unwired since #3667 created it, so the PR neither causes nor worsens the gap, and it is a systemic repo-hygiene question rather than something this PR should carry. Noting it only so the 52/52 result is read correctly: those passes come from local execution and this review's re-run, not from the CI badge. Shellcheck *does* cover the file (`shell-lint.yml` lints all of `defaults/scripts/`), and the in-suite source-contains guards partially compensate.

**3. `--auto`-queued merges do not reach the self-heal** — pre-existing symmetry, no action needed.

When `--auto` genuinely queues server-side, `merge-pr.sh` exits before `_reset_partial_increment_labels`. The reopen therefore will not run on that path — but neither does the pre-existing #3667 label reset, so this introduces no new asymmetry, and the placement at line 572 means the *pre-merge warning* still fires on every path including `--auto` and `--dry-run`. Correctly scoped.

## Verdict

Approved. Root cause established from evidence rather than assumed, the fix is layered correctly (detect → self-heal → durable doc fix) with the docs explicitly named as the real remedy and the script as a flicker-prone backstop, degradation is graceful at every step, and the conservative attribution boundary is the right call. Ship it.

PR #4591, opened 2026-07-30
fix(daemon): resolve launchctl_pid's domain via resolve_launchd_domain, not a hardcoded gui/<uid>

rjwalters:
## Summary

`launchctl_pid()` (the unreachable-path launchd liveness probe in
`loom-daemon/src/daemon_install_state.rs`) and both display-string call
sites in `check_liveness()` independently hardcoded `gui/{uid}/{label}`,
ignoring the shared `resolve_launchd_domain()` fallback rule
(`LOOM_LAUNCHD_DOMAIN` override -> `gui/<uid>` -> `user/<uid>`) that PR
#4533 introduced for the reachable-path protection probe. On an SSH-only
host where `gui/<uid>` never resolves, this meant the unreachable-path
probe used the wrong domain AND its human-readable detail string kept
claiming `gui/<uid>` even when the real (correct) domain was `user/<uid>`.

## Changes

- `launchctl_pid()` now takes an already-resolved `domain: &str` instead of
  hardcoding `gui/{uid}`.
- `check_liveness()` resolves the domain once via `resolve_launchd_domain()`
  and reuses that single resolution for both the `launchctl_pid` probe and
  both detail-string call sites, so the probed domain and the reported
  domain can never disagree.
- `EnvOverrides` gains a `launchd_domain_override` field (from
  `LOOM_LAUNCHD_DOMAIN`), mirroring the field `ProtectionEnv` already
  carries for the reachable-path probe, and `classify_with_process_age_fn`
  threads it through to `check_liveness`.
- Fixed two `EnvOverrides` struct-literal test call sites (in
  `daemon_install_state.rs` and `autonomy_marker.rs`) for the new field.

## Acceptance Criteria Verification

- [x] `launchctl_pid()` uses the shared domain-resolution rule (env
  override -> `gui/<uid>` -> `user/<uid>`) instead of a hardcoded
  `gui/<uid>` — verified by reading the new implementation, which calls
  `resolve_launchd_domain()` before probing.
- [x] Existing unreachable-path exit-code semantics (#4069) are unchanged —
  no `InstallState` variants added or removed; all 47
  `daemon_install_state` tests (including every pre-existing exit-code
  test) still pass unmodified.
- [x] Unit tests cover the domain-resolution order — added
  `launchctl_pid_domain_override_is_honored_verbatim` and
  `launchctl_pid_falls_back_to_gui_uid_then_user_uid_when_no_override`.
- [x] Both `check_liveness()` display-string call sites (not just
[...]

Issue #4547, opened 2026-07-30
Flaky tests: fake-gh subprocess tests intermittently fail from parallel PATH/env mutation in loom-daemon lib tests


rjwalters:
## Problem

`cargo test -p loom-daemon --lib` intermittently fails in a **nondeterministic** test, always one that spawns a fake `gh`/`bash` subprocess. Observed three *different* victims across three runs:

| Run | Failing test |
|---|---|
| CI on PR #4533 (`run 30512756591`) | `quarantine_reconciliation::tests::reconcile_workspace_never_touches_manual_block` (`checked` 0, expected 1) |
| Local, PR #4533 head | `quarantine_reconciliation::tests::reconcile_workspace_releases_issue_with_quarantine_comment` (`checked` 0, expected 1) |
| Local, plain `main` (03fd990f) | `pipeline_snapshot::tests::gh_pipeline_source_records_error_but_keeps_other_metrics` |

The rotating victim identity proves this is a race, not a regression in any one PR. Reproduced on `main` with no PR applied.

## Suspected root cause

Tests mutate the **process-global** environment while sibling tests spawn subprocesses in parallel:

- `loom-daemon/src/disk_headroom.rs:305,307,332,334` prepend a stub dir to `PATH` and restore it (`std::env::set_var("PATH", ...)`).
- Those tests carry `#[serial]`, but `serial_test`'s `#[serial]` only serializes against **other `#[serial]` tests** — the `reconcile_workspace_*` / `gh_pipeline_source_*` tests are *not* serial, so they still run concurrently inside that mutation window.
- The victims' fake `gh` scripts start with `#!/usr/bin/env bash`, so `bash` is resolved through the **inherited** `PATH`; and `std::env::set_var` racing with `posix_spawn`'s environ read is unsound regardless (this is exactly why Rust 2024 makes `set_var` `unsafe`).

A failed spawn/exec makes `list_blocked_issues` return `Err`, `reconcile_workspace` returns `(0, 0)`, and `assert_eq!(checked, 1)` fails with `left: 0`.

## Suggested fixes (pick one)

1. Add `#[serial]` to every test that spawns a subprocess whose resolution depends on inherited env (the `reconcile_workspace_*`, `gh_pipeline_source_*` families), so they serialize against the `PATH`-mutating tests.
[...]

Issue #4536, opened 2026-07-30
daemon_install_state: launchctl_pid hardcodes gui/<uid> instead of resolve_launchd_domain


rjwalters:
## Context

Surfaced during review of PR #4533 (issue #4354). The new reachable-path protection probe resolves the launchd domain via the `resolve_launchd_domain` rule (`LOOM_LAUNCHD_DOMAIN` → `gui/<uid>` → `user/<uid>`), mirroring `lib/launchd-domain.sh`.

However, the pre-existing `launchctl_pid()` on the **unreachable** path (`loom-daemon/src/daemon_install_state.rs`) still hardcodes `gui/<uid>` — a pre-#4130-style inconsistency. On an SSH-only host (no GUI session), the unreachable-path probe may fail to find a daemon that is actually bootstrapped in `user/<uid>`, degrading the `install_state` verdict.

PR #4533 deliberately left this alone because changing it would alter unreachable-path behavior, which #4354's AC4 forbade in that PR's scope.

## Proposed change

Make `launchctl_pid()` resolve its domain through the same `resolve_launchd_domain` helper the protection probe uses (honoring `LOOM_LAUNCHD_DOMAIN`), with tests pinning the fallback order.

## Acceptance criteria

- [ ] `launchctl_pid()` uses the shared domain-resolution rule (env override → `gui/<uid>` → `user/<uid>`) instead of a hardcoded `gui/<uid>`.
- [ ] Existing unreachable-path exit-code semantics (#4069) are unchanged.
- [ ] Unit tests cover the domain-resolution order.

## Dependencies

- [x] #4533: Introduces `resolve_launchd_domain()` and the `ProtectionState`/`probe_protection` sibling classification in `loom-daemon/src/daemon_install_state.rs`. **Merged** 2026-07-30T04:24:37Z; verified on `origin/main` at `loom-daemon/src/daemon_install_state.rs:939` (`resolve_launchd_domain`). Dependency satisfied — this issue is unblocked.

## Affected Files

- `loom-daemon/src/daemon_install_state.rs` — `launchctl_pid()` (currently ~L403-413) hardcodes `format!("gui/{uid}/{label}")`; switch to `resolve_launchd_domain(...)` (landing via #4533) for the domain component.
[...]

PR #4598, opened 2026-07-30
fix(daemon): bound daemon_install_state's subprocess probes with a timeout (#4548)

### LINKED ISSUE #4548: daemon_install_state: bound the launchctl/systemctl/ps/kill/id probes with a timeout
## Context

`loom-daemon/src/daemon_install_state.rs` spawns query-only subprocesses to classify install/protection state. Two small consistency/robustness gaps, both noted while reviewing PR #4533 (#4354) and neither introduced by it:

## 1. The probes are unbounded

All of these use a bare `Command::output()` with no timeout:

- `launchctl_pid()` (pre-existing, unreachable path)
- `resolve_launchd_domain()` → `launchctl print gui/<uid>` (new in #4354)
- `launchctl_job_provisioned()` → `launchctl print <domain>/<label>` (new in #4354)
- `systemd_timer_provisioned()` → `systemctl --user is-enabled <unit>` (new in #4354)
- the `ps` / `id` / `kill` calls in the same module

Both the `install_state` and the new `protection` probes document themselves as "read-only and never fails the command", which is true for *failure* but not for *hanging*: a wedged `systemd --user` bus or a `launchctl print` against a bad domain would block `loom-daemon status` indefinitely with no output. The crate already has `sweep_registry::output_with_timeout` (used by `quarantine_reconciliation`'s `gh` calls) — wrapping these in it with a short bound (a couple of seconds) would make the "never fails" contract cover hangs too, degrading to `unknown` / the less-specific verdict instead.

PR #4533's new probes deliberately match the module's existing unbounded pattern rather than diverging, so this is a whole-module cleanup, not a change request against that PR.

## 2. `launchctl_pid` hardcodes `gui/<uid>`

`launchctl_pid()` on the unreachable path still hardcodes the `gui/<uid>` domain, while the newer code (`lib/launchd-domain.sh::resolve_launchd_domain`, and `daemon_install_state::resolve_launchd_domain` added in #4354) resolves `LOOM_LAUNCHD_DOMAIN` → `gui/<uid>` → `user/<uid>`. On an SSH-only/headless macOS host where the daemon lives in `user/<uid>`, `launchctl_pid` can therefore report "no pid" for a job that is in fact bootstrapped.

This was left alone in PR #4533 on purpose: #4354's AC4 forbids changing unreachable-path behavior, and this *is* an unreachable-path behavior change (it would make some `not-expected` verdicts become correct ones). It needs its own issue with its own exit-code review.

Discovered during code evaluation of PR #4533.

---

### Scope narrowed to item 1 only (timeout bounding)

**Item 2 is a duplicate of already-approved work.** `check-duplicate.sh --issue 4548` surfaced #4536, which was filed 22 minutes before this issue and already covers "`launchctl_pid` hardcodes `gui/<uid>`" verbatim — it is already `loom:curated` + `loom:issue` (approved, awaiting a Builder) with its own acceptance criteria, affected-files enumeration, and test plan. A near-identical report (#4539) was independently filed and closed as not-planned for the same reason. Re-implementing item 2 here would duplicate approved work, so this issue is **rescoped to item 1 only**: bounding the module's unbounded subprocess probes with a timeout. See #4536 for item 2.

**Soft interaction, not a hard dependency**: #4536 rewrites `launchctl_pid()`'s domain-resolution line (`format!("gui/{uid}/{label}")` → `resolve_launchd_domain(...)`), and this issue wraps the same function's `Command::output()` call with a timeout. Both edit `launchctl_pid()`'s body but on different lines/aspects (domain string vs. call wrapper) — whichever PR lands second should expect a small rebase, but neither issue should block on the other.

### Problem Statement

Every `Command::output()` call in `daemon_install_state.rs` blocks synchronously with no timeout. The module's own doc comments claim these probes are "read-only and never fail the command" — true for a *failed* subprocess (handled via `.ok()?` / `is_ok_and(...)`), but not for a *hung* one. A wedged `launchctl` (bad domain, corrupted launchd state) or `systemctl --user` (broken/absent D-Bus session) blocks `loom-daemon status` indefinitely with no output, defeating the whole point of a read-only status probe.

### Acceptance Criteria

- [ ] All 7 `Command::output()` call sites below are replaced with `crate::sweep_registry::output_with_timeout(cmd, timeout)`, using a short, named constant (e.g. `PROBE_TIMEOUT_SECS`, following the `DEFAULT_STARTUP_GRACE_SECS` naming convention at `daemon_install_state.rs:98`) bound to a couple of seconds.
- [ ] A probe that times out degrades to exactly the same fallback value the function already returns for a failed/absent binary — `None` for the `Option`-returning helpers, `false` for `pid_alive`, `None` for `current_uid` — never a panic, never an indefinite block.
- [ ] Fast success/failure behavior (a real process that returns quickly) is byte-for-byte unchanged versus before the wrap.
- [ ] Unreachable-path exit-code semantics (#4069) are unchanged — no `install_state`/`protection` verdict changes for any probe that completes within the timeout.
- [ ] At least one new unit test proves a hung stub binary causes the wrapped function to return its degrade value within the timeout window rather than blocking (see Test Plan).

### Affected Files

- `loom-daemon/src/daemon_install_state.rs` — wrap all 7 unbounded `Command::output()` call sites:
  - `pid_alive()` (L396-399) — `kill -0 <pid>`
  - `current_uid()` (L403-405) — `id -u`
  - `launchctl_pid()` (L418-421) — `launchctl print gui/<uid>/<label>`
  - `process_age_secs()` (L538-540) — `ps -o etime= -p <pid>`
  - `resolve_launchd_domain()` (L945-948) — `launchctl print gui/<uid>` reachability probe
  - `launchctl_job_provisioned()` (L961-964) — `launchctl print <domain>/<label>`
  - `systemd_timer_provisioned()` (L979-982) — `systemctl --user is-enabled <unit>`
  - `#[cfg(test)] mod tests` (end of file) — add PATH-stub timeout tests (see Test Plan)
- `loom-daemon/src/sweep_registry.rs` — no change expected; `output_with_timeout` (L1466-1484) is the existing `pub(crate)` helper to reuse as-is.

### Test Plan

- [ ] Automated tests: add PATH-stub tests following the exact pattern already established in `loom-daemon/src/disk_headroom.rs` (`test_worktree_root_free_gb_returns_none_on_df_failure`, ~L285-310) — write an executable stub script (`#!/bin/sh\nsleep <longer-than-timeout>\n`) into a tempdir, prepend it to `PATH`, call the wrapped function, and assert it returns its degrade value promptly instead of hanging. Mark these `#[serial]` since they mutate the global `PATH` env var, matching `disk_headroom.rs`'s existing tests.
- [ ] Automated tests: run `cargo test -p loom-daemon daemon_install_state` to confirm no regression in existing fast-path test coverage.
- [ ] Manual verification: on Linux CI (no `launchctl`/`systemctl` binaries), confirm `current_uid`/`pid_alive`/`process_age_secs` (backed by `id`/`kill`/`ps`, which do exist on Linux) behave identically before/after.
- [ ] Edge cases: choose a timeout long enough to avoid false-positives on a loaded CI runner's real `ps`/`id`/`kill` calls, but short enough that `loom-daemon status` stays responsive against a wedged `launchctl`/`systemctl` bus — the issue's own suggestion of "a couple of seconds" is a reasonable starting point.

### Dependencies

Not a hard blocker on #4536 (see "Soft interaction" above) — both may proceed independently; whichever lands second rebases past the other's edit to `launchctl_pid()`.

rjwalters:
## Summary

Bounds every query-only subprocess probe in `loom-daemon/src/daemon_install_state.rs` with a timeout, so a wedged `systemd --user` bus or an XPC-stalled `launchctl print` can no longer block `loom-daemon status` indefinitely — the very command an operator runs to diagnose a wedge.

All 7 previously-unbounded `Command::output()` call sites now route through a module-private `probe_output(cmd, timeout)` that delegates to the existing `crate::sweep_registry::output_with_timeout` (the same helper the reaper's `gh` calls use) with a named `PROBE_TIMEOUT` = `PROBE_TIMEOUT_SECS` (2s):

| Function | Probe | Degrade value on timeout |
|---|---|---|
| `pid_alive` | `kill -0 <pid>` | `false` |
| `current_uid` | `id -u` | `None` |
| `launchctl_pid` | `launchctl print <domain>/<label>` | `None` |
| `process_age_secs` | `ps -o etime= -p <pid>` | `None` |
| `resolve_launchd_domain` | `launchctl print gui/<uid>` reachability | `gui/<uid>` unreachable ⇒ `user/<uid>` |
| `launchctl_job_provisioned` | `launchctl print <domain>/<label>` | `None` (unknown) |
| `systemd_timer_provisioned` | `systemctl --user is-enabled <unit>` | `None` (unknown) |

In every case that is *exactly* the value the function already returned for a failed/absent binary, so timeout folds into the existing unknown arm — no new state, no panic, no verdict change.

## Behavior preservation

- **Fast paths unchanged.** `probe_output` returns `Some(output)` for any child that completes in time, including a **nonzero** exit. That matters: collapsing nonzero to `None` would flip `systemd_timer_provisioned`'s `Some(false)` ("present but disabled") into `None` ("unknown"). Covered by a test.
- **#4069 exit-code semantics untouched** — no `install_state` / `protection` verdict changes for any probe that completes within the window.
- **stdin parity.** `probe_output` sets `stdin(Stdio::null())`: `Command::output()` nulls stdin, whereas `output_with_timeout`'s bare `spawn()` would inherit the daemon's stdin. None of these probes read stdin, but the wrap should not smuggle that in. (`sweep_registry`'s own callers are out of scope here.)

## Tests

`cargo test -p loom-daemon daemon_install_state` → **54 passed**; full `cargo test -p loom-daemon --lib` → **2474 passed, 0 failed**. `cargo clippy -p loom-daemon --all-targets` clean, `cargo fmt` applied.

New coverage:

- `probe_output_abandons_a_command_that_outlives_its_timeout` — `sleep 60` with a 200 ms timeout returns `None` at the deadline (fast, no PATH mutation).
- `probe_output_preserves_fast_success_and_fast_failure_verbatim` — exit 0 with stdout, and exit 3 with stderr, both surface verbatim.
- `probe_output_degrades_to_none_when_the_binary_is_absent` — unspawnable binary keeps its pre-change `None`.
- `systemd_timer_provisioned_degrades_to_unknown_when_systemctl_hangs` — `#[serial]` PATH-stub (`#!/bin/sh\nsleep 60`) shadowing `systemctl`: returns `None` well inside the window rather than hanging.
- `launchctl_probes_degrade_when_launchctl_hangs` and `resolve_launchd_domain_falls_back_to_user_uid_when_launchctl_hangs` — same PATH-stub pattern for the three `launchctl` probes, including the reachability probe's `user/<uid>` fallback.
- `kill_id_and_ps_backed_probes_keep_their_fast_path_behavior` — the Linux-CI-present trio (`kill`/`id`/`ps`) pinned before/after.

**Why `kill`/`id`/`ps` are not PATH-stubbed:** many sibling tests in this module classify against a real `pid_alive(std::process::id())` / `ps -o etime=`, and shadowing those binaries process-globally would flake them even under `#[serial]` (the #4525 / #4547 lesson). The stub targets chosen (`systemctl`, `launchctl`) are verdict-neutral for siblings — on Linux CI those binaries are absent, and a hung stub degrades to the same value absence already produces — so no sibling needs to join the serial lock. The shared timeout mechanism is proven directly via `probe_output`.

## Scope

Timeout-wrapping only. #4536's `gui/<uid>` hardcoding fix already landed on main as #4591; this branch is rebased past it and preserves its `resolve_launchd_domain`-based domain resolution in `launchctl_pid` verbatim.

Closes #4548

### DIFF
--- a/loom-daemon/src/daemon_install_state.rs
+++ b/loom-daemon/src/daemon_install_state.rs
@@ -68,6 +68,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::time::Duration;
 
 /// Default macOS launchd label, matching `loom-daemon-start.sh` /
 /// `loom-daemon-watchdog.sh`'s fallback when the marker predates the
@@ -96,6 +97,46 @@ pub const EXIT_ALIVE_BUT_UNRESPONSIVE: i32 = 4;
 /// faulted. Overridable via `LOOM_DAEMON_STARTUP_GRACE_SECS`. Sized above the
 /// observed ~40–60s `bootout`/`bootstrap` socket-bind latency (#4213).
 pub const DEFAULT_STARTUP_GRACE_SECS: u64 = 90;
+
+/// Wall-clock bound (seconds) on every query-only subprocess probe this module
+/// makes — `kill -0`, `id -u`, `ps -o etime=`, `launchctl print`, `systemctl
+/// --user is-enabled` (#4548).
+///
+/// All of these normally answer in milliseconds, but each has a real-world hang
+/// mode: a wedged `systemd --user` bus makes `systemctl` block on a D-Bus
+/// connect, and `launchctl print` against an unreachable domain can stall on
+/// XPC. Unbounded, any one of them wedges `loom-daemon status` — the very
+/// command an operator runs to diagnose a wedge. A couple of seconds is far
+/// above the real cost of these probes even on a loaded CI runner, yet keeps
+/// `status` responsive.
+const PROBE_TIMEOUT_SECS: u64 = 2;
+
+/// [`PROBE_TIMEOUT_SECS`] as a [`Duration`].
+const PROBE_TIMEOUT: Duration = Duration::from_secs(PROBE_TIMEOUT_SECS);
+
+/// Run a query-only probe `cmd`, abandoning (killing) it if it exceeds
+/// `timeout`, and collapse *every* failure mode to `None` (#4548).
+///
+/// `None` therefore means "no usable answer" for all three of: the binary is
+/// absent/unspawnable, the spawn itself failed, or the probe hung past
+/// `timeout`. Every caller in this module already degraded a failed/absent
+/// binary to its unknown value (`None` / `false`), so folding the timeout into
+/// the same arm keeps the existing verdict semantics — including the #4069
+/// exit-code mapping — byte-for-byte unchanged for any probe that completes in
+/// time.
+///
+/// `stdin` is explicitly nulled to preserve `Command::output()`'s contract:
+/// `output()` nulls stdin, while [`output_with_timeout`]'s `spawn()` would
+/// otherwise *inherit* the daemon's stdin. None of these probes read stdin, but
+/// inheriting it is a behavior change this wrap must not smuggle in.
+///
+/// [`output_with_timeout`]: crate::sweep_registry::output_with_timeout
+fn probe_output(mut cmd: Command, timeout: Duration) -> Option<std::process::Output> {
+    cmd.stdin(std::process::Stdio::null());
+    crate::sweep_registry::output_with_timeout(cmd, timeout)
+        .ok()
+        .flatten()
+}
 
 // ============================================================================
 // Public types
@@ -399,19 +440,20 @@ fn resolve_marker_fields(map: &HashMap<String, String>, loom_dir: &Path) -> Mark
 /// `kill -0 <pid>` via subprocess (matches `terminal.rs`'s existing pattern
 /// in this crate — no `libc`/`nix` dependency needed). Returns `false` for
 /// both "no such process" and "not owned by us", exactly like the shell
-/// script's `kill -0 "$pid" 2>/dev/null`.
+/// script's `kill -0 "$pid" 2>/dev/null`. Bounded by [`PROBE_TIMEOUT`]; a hung
+/// `kill` degrades to `false`, exactly like an absent one (#4548).
 fn pid_alive(pid: u32) -> bool {
-    Command::new("kill")
-        .args(["-0", &pid.to_string()])
-        .output()
-        .is_ok_and(|o| o.status.success())
+    let mut cmd = Command::new("kill");
+    cmd.args(["-0", &pid.to_string()]);
+    probe_output(cmd, PROBE_TIMEOUT).is_some_and(|o| o.status.success())
 }
 
+/// Current uid via `id -u`. Bounded by [`PROBE_TIMEOUT`]; a hung `id` degrades
+/// to `None`, exactly like an absent one (#4548).
 fn current_uid() -> Option<String> {
-    Command::new("id")
-        .arg("-u")
-        .output()
-        .ok()
+    let mut cmd = Command::new("id");
+    cmd.arg("-u");
+    probe_output(cmd, PROBE_TIMEOUT)
         .filter(|o| o.status.success())
         .and_then(|o| String::from_utf8(o.stdout).ok())
         .map(|s| s.trim().to_string())
@@ -424,12 +466,15 @@ fn current_uid() -> Option<String> {
 /// [`resolve_launchd_domain`]) — the caller resolves it once and reuses it
 /// for both this probe and any human-readable detail string, avoiding a
 /// duplicate `launchctl`/`id` round trip per [`check_liveness`] call.
+///
+/// Bounded by [`PROBE_TIMEOUT`]: a `launchctl print` that stalls on XPC
+/// degrades to `None` — "no live pid" — exactly like an absent `launchctl`
+/// (#4548).
 fn launchctl_pid(domain: &str, label: &str) -> Option<u32> {
     let service = format!("{domain}/{label}");
-    let output = Command::new("launchctl")
-        .args(["print", &service])
-        .output()
-        .ok()?;
+    let mut cmd = Command::new("launchctl");
+    cmd.args(["print", &service]);
+    let output = probe_output(cmd, PROBE_TIMEOUT)?;
     if !output.status.success() {
         return None;
     }
@@ -553,12 +598,12 @@ fn parse_etime(raw: &str) -> Option<u64> {
 /// dependency — matches this module's `kill -0` / `launchctl` subprocess
 /// pattern). Degrades to `None` on a failed/absent `ps` or unparseable output,
 /// so the caller falls through to today's verdicts rather than falsely
-/// reporting "starting" (#4213).
+/// reporting "starting" (#4213). Bounded by [`PROBE_TIMEOUT`]: a hung `ps`
+/// takes the same `None` path (#4548).
 fn process_age_secs(pid: u32) -> Option<u64> {
-    let output = Command::new("ps")
-        .args(["-o", "etime=", "-p", &pid.to_string()])
-        .output()
-        .ok()?;
+    let mut cmd = Command::new("ps");
+    cmd.args(["-o", "etime=", "-p", &pid.to_string()]);
+    let output = probe_output(cmd, PROBE_TIMEOUT)?;
     if !output.status.success() {
         return None;
     }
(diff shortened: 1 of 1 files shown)


rjwalters:
Code evaluation looks good; CI is still running. Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
Code evaluation pending — CI checks (Rust Linting & Build Check, Rust Unit Tests) are still running. Releasing the claim; will re-evaluate once CI settles.

rjwalters:
Code evaluation queued; CI (Rust Unit Tests) is still running. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation looks good; CI is still running (Rust Unit Tests pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Approved.

**Verified independently (not just Builder's summary):**

- **All 7 call sites** (`pid_alive`, `current_uid`, `launchctl_pid`, `process_age_secs`, `resolve_launchd_domain`'s `gui_ok` reachability probe, `launchctl_job_provisioned`, `systemd_timer_provisioned`) are genuinely routed through the new `probe_output` wrapper — confirmed by reading the diff line-by-line, no site missed or duplicated.
- **Degrade-value parity, checked function by function**: each site's pre-existing `.output().ok()` / `.output().ok()?` / `.output().is_ok_and(...)` degrade path maps 1:1 onto `probe_output`'s `Ok(None)` (timeout) / `Err` (spawn failure) both collapsing to `None` via `.ok().flatten()`. Traced `output_with_timeout` itself (`sweep_registry.rs`) to confirm it returns `Ok(None)` on timeout after `kill()`+`wait()`, `Ok(Some(output))` on completion, `Err` on spawn failure — matches every call site's existing failure semantics, including `systemd_timer_provisioned`'s `Some(false)` vs `None` distinction (fast nonzero exit is never flattened to `None`).
- **Fast-path semantics unchanged**: `probe_output` only adds a deadline-polled `try_wait()` loop and `stdin(Stdio::null())` around the same piped-stdout/stderr child; a fast `Ok(Some(output))` is untouched. The stdin-null addition is a correct fix, not cosmetic — `Command::output()`'s docs guarantee stdin is closed, whereas `output_with_timeout`'s bare `spawn()` would otherwise inherit the daemon's process stdin.
- **New tests are meaningful, not tautological**: `probe_output_abandons_a_command_that_outlives_its_timeout` exercises a real `sleep 60` (not PATH-stubbed) killed at a 200ms deadline — a genuine mechanism test. The three `#[serial]` PATH-stub tests (`systemctl`/`launchctl` shadowed with a `sleep 60` stub) assert both the degraded value *and* an elapsed-time bound well under the stub's 60s sleep, so a regression to "wait for the stub" would fail on wall-clock, not just on value.
- **`#[serial]` usage is correct and consistent** with the existing repo pattern (`disk_headroom.rs`, `pipeline_snapshot.rs`, `quarantine_reconciliation.rs` — `serial_test`'s default lock is process-global, exactly what's needed to guard `PATH` mutation, per the #4525/#4547 lesson and PR #4588 which fixed the identical class of flake).
- **The "why not stub kill/id/ps" reasoning checks out** against the actual file: numerous pre-existing sibling tests (`marker_present_live_pid_is_alive_but_unresponsive`, `stale_heartbeat_with_unknown_process_age_makes_no_prior_boot_claim_end_to_end`, etc.) write a pid file with the real `std::process::id()` and expect `pid_alive`/`process_age_secs` to classify the *actual* running test process — globally shadowing `kill`/`ps` would break those. Relying on Linux CI's absence of `launchctl`/`systemctl` for the remaining functions' fast-path coverage is a reasonable trade-off, not an excuse to skip coverage.
- **Local run**: `cargo test -p loom-daemon --lib daemon_install_state` → 54 passed (8s). Full `cargo test -p loom-daemon --lib` → 2474 passed, 0 failed (173s). `cargo clippy -p loom-daemon --all-targets` clean.
- **CI**: all required checks pass (`Rust Unit Tests` 4m24s, `Rust Linting & Build Check` 43s, plus the doc/installer checks). `mergeStateStatus` is `CLEAN`.
- The pre-existing `pid_alive(u32::MAX)` oddity noted in the PR body (`kill -0 -1` meaning "every process" on Linux) is accurately scoped as out-of-scope — it's unreachable from production callers (no real pid is ever `u32::MAX`) and the new test correctly sidesteps it with `i32::MAX as u32` instead.

No issues found. Scope, correctness, and test quality all hold up under independent verification.

rjwalters:

- **Size check**: 301 lines changed (269 additions + 32 deletions), exceeds the default 200-line auto-merge limit. No `loom:auto-merge-ok` override label is present.
- **Mergeable check**: GitHub reports `mergeable: UNKNOWN` (still computing or needs a refresh).

**Next steps:**
- A human can review and add the `loom:auto-merge-ok` label if this larger-than-usual PR (single-file, well-tested subprocess-timeout fix) is judged safe to bypass the size gate.
- Mergeable state may resolve on its own; Champion will re-check on the next tick.

---
*Automated by Champion role*

rjwalters:

- **Size check**: 301 lines changed (+269/-32), exceeding the 200-line auto-merge limit (`champion.auto_merge_max_lines`). No `loom:auto-merge-ok` waiver label present.
- **Merge conflict check**: `mergeable` is currently `UNKNOWN` (GitHub still computing mergeability).

**Next steps:**
- If a human reviewer judges this PR low-risk (single file, mechanical timeout-wrapping change with new tests — the exact profile described in #4571's pending judgment-based auto-merge policy), apply `loom:auto-merge-ok` to waive the size gate.
- Otherwise, this will be re-evaluated on the next Champion tick once mergeability resolves.

---
*Automated by Champion role*

Issue #4259, opened 2026-07-28
build-gate: 1200s timeout recurs under concurrent sweep load (post-#4020/#4084) — main-health gate never evaluates


rjwalters:
## Problem

The main-health gate still cannot complete a run on this host: `bash .loom/scripts/build-gate.sh` **times out at 1200s and is killed**, so the gate reports `not evaluated` — permanently. This is a recurrence after the closed fixes #4020 (nice / serialized cargo / shared target-dir proposal) and #4084 (nice -n 5 did not prevent a 1200s timeout under 2 concurrent sweeps). Today it fails under 6–7 concurrent sweeps.

The gate correctly fails open (dispatch is not halted, and the status output explicitly notes a timeout is not evidence about main) — but a gate that never evaluates provides zero signal, and the red-main protection it exists for (#3647 step-8 class breakage) is effectively disabled on any host busy enough to matter.

## Evidence (2026-07-28, `loom-daemon status`)

    Main-health gate: not evaluated (timeout: gate command 'bash .loom/scripts/build-gate.sh'
    timed out after 1200s and was killed; main was NOT evaluated; last output:
    tup_does_not_overwrite_existing_files has been running for over 60 seconds
    test terminal::claude_config::tests::test_...)

Host state at the time: 6 in-flight sweeps (cap 6), ~9 of 28 cores consumed, several sweeps running their own `cargo build/test` in worktrees. The named hangs (`tup_does_not_overwrite_existing_files`, `terminal::claude_config::tests::*`) are the same tests that starve under parallel-cargo / syspolicyd exec-saturation contention seen in prior incidents.

## Root-cause hypotheses

1. The gate's `cargo test` competes with N concurrent sweep builds for the same cores **and** for Gatekeeper (`syspolicyd`) exec assessment — individual tests stall past any nice-level remedy (nice was already tried, #4084).
2. The gate runs the **full** suite; its wall-clock under contention exceeds any reasonable timeout, so raising the timeout just moves the failure.

## Proposed direction

- **Tiered gate:** the health gate runs a cheap deterministic subset (compile + smoke, target a few minutes cold) as its verdict; the full suite runs off-peak or only when in-flight sweeps == 0. A gate that finishes in 3 minutes under load beats one that would be thorough if it ever finished.
[...]

PR #4604, opened 2026-07-30
refactor(daemon): drop the CPU-headroom admission term for one maxConcurrent knob + a machine-wide build slot

### LINKED ISSUE #4512: simplify concurrency: one per-machine max-concurrency knob; drop the CPU-headroom formula; serialize rare high-CPU stages via a build slot
## Operator decision (this is a direction, not a proposal)

The multi-term CPU calibration (`estCoresPerSweep` × `cpuUtilizationTarget` × live-idle shaving, plus `loom-daemon calibrate` to tune it) is over-built for what the pipeline actually does, and it is now the binding constraint on fleet throughput. Replace it with **one per-machine max-concurrency parameter, tuned empirically over time**.

## Evidence

On an 8-core EC2 worker (2026-07-29):

```
Dynamic concurrency cap: 2  (the number dispatch uses)
  = min(healthy 6 × per-token 2 = 12, disk headroom 36, cpu headroom 2, configured max 3)
  cpu headroom: 2 concurrent-sweep slot(s) (8 logical cores, 95% idle measured (≈0.4 cores consumed))
```

The host is **95% idle** while the formula caps it at 2. `loom-daemon calibrate` recommends "no change" — the formula can't see past its own estimate. The issue→PR pipeline is dominated by API-wait (curator/builder/judge conversations); the sustained-heavy-CPU phases (release builds, full test suites, the build gate) are a small fraction of wall-clock. The formula prices EVERY sweep as a build (`estCoresPerSweep = 2.0`, calibrated in #4031 against the primary host's Rust phases), so it throttles the ~90% low-CPU majority to defend against the minority case — the #4231 meltdown scenario it was built for.

## Direction

1. **One knob**: a per-machine `maxConcurrent` (the existing `autonomous.workFinder.maxConcurrent`) becomes THE admission cap. Operators tune it from observed behavior. Expect 10+ on an 8-core API-bound worker.
2. **Delete the CPU term from admission**: retire `estCoresPerSweep`, `cpuUtilizationTarget`, and the live-idle shaving from `resolve_dynamic_max_concurrent`. Reduce `loom-daemon calibrate` to printing measurements (or retire it).
3. **Serialize the actually-heavy stages where they occur**: a machine-wide build slot (mkdir-atomic lock, same shape as the worktree lock) checked out around `cargo build/test`/build-gate stages — the pre-rescope direction of #4003 ("agents check out a slot only for bound work"), of which #4020/#4084 built the build-gate-specific piece (nice, shared target-dir lock). N sweeps run concurrently; at most 1–2 hold the build slot; the rest of a sweep's lifecycle never queues.
4. **Keep the real hard floors**: disk headroom and the token axis stay in the `min()` — they meter actual exhaustible resources. The host breaker (#4235) stays as the load safety net; it is what makes a fixed knob safe to tune upward (a mis-set knob trips the breaker instead of melting the host — measured, not estimated).

- [ ] `resolve_dynamic_max_concurrent` = `min(token axis, disk headroom, configured maxConcurrent)` — no CPU estimate term.
- [ ] A machine-wide build-slot lease serializes designated high-CPU stages across concurrent sweeps (bounded wait + telemetry, never a deadlock; degrades open if the lock dir is unavailable).
- [ ] `estCoresPerSweep` / `cpuUtilizationTarget` config+env accepted-but-ignored with a deprecation warning (no config break), docs updated.
- [ ] The #4231 scenario re-run (or reasoned through in the PR) under the new model: N concurrent sweeps + release builds → builds serialize on the slot, breaker stays closed.

Refs: #4031 (calibration measurement), #4231 (meltdown), #4234 (admission ramp), #4235 (host breaker), #4003 (resource leases — rescoped away; this revives its core idea in bounded form), #4020/#4084 (build-gate serialization), `loom-daemon/src/cpu_headroom.rs`.

rjwalters:
## Summary

The dynamic concurrency cap carried a CPU term since #3978:

```
cpu_headroom = (logical_cpus × cpuUtilizationTarget − consumed_cores) / estCoresPerSweep
```

It priced **every** sweep as a build, but sweep wall-clock is dominated by API wait (curator / builder / judge conversations) — so it throttled the low-CPU majority to defend against the heavy-build minority. The issue's measured evidence: an 8-core worker sitting **95% idle** was capped at **2**, and `loom-daemon calibrate` recommended "no change" because the formula could not see past its own estimate.

Admission is now `min(token axis, disk headroom, configured maxConcurrent)` — two terms that meter genuinely *exhaustible* resources (accounts, bytes) plus one per-machine operator knob — and the protection for the genuinely heavy stages moves to **where the load actually is**: a machine-wide build slot.

## Changes

- **`work_finder.rs`** — `resolve_dynamic_max_concurrent` loses the CPU term (it is now a `min` of token axis × per-token factor, disk headroom, configured max). `resolve_cpu_utilization_target` / `resolve_cpu_est_cores_per_sweep` and their threading through `spawn_multi_work_finder_task` are gone. New `deprecated_cpu_knob_notice` / `warn_deprecated_cpu_knobs` + `DEPRECATED_CPU_CONFIG_KEYS` / `DEPRECATED_CPU_ENV_VARS`.
- **`build_slot.rs` (new)** + **`defaults/scripts/lib/build-slot.sh` (new)** — a machine-wide `mkdir`-atomic slot lease at `~/.loom/locks/build-slot/slot-<i>`, built on the existing `MkdirLock` primitive (the same shape as the per-issue claim lock). Two implementations, one protocol, so the daemon and each sweep worktree serialize against **each other**.
- **`main_health_gate.rs`** — the gate command takes a slot, acquired *outside* the timeout window so a queue wait cannot turn contention into a false `Unevaluated` (the #3978/#4020 failure mode), and exports `LOOM_BUILD_SLOT_HELD` to the child.
- **`build-gate.sh`** — sources the Bash half and takes a slot around the compile, with a `trap` release.
- **`calibrate/`** + **`main.rs`** — measurement-only. Nothing is derivable as a recommendation once the CPU term is gone, so `--write` is accepted-but-ignored. The report prints host observations, the resolved knobs, the `min(...)` breakdown, the binding term, and one line of tuning advice.
- **`config_resolver.rs`** — new `source_of` / `tier_paths_by_precedence`: which config tier actually set a key. Surfaced by calibrate as `set by: <path>`.
- **`cpu_headroom.rs`** — shrunk to a pure *measurement* module (host breaker #4235, the gate's load deferral #4259, status/calibrate reports). Nothing it produces feeds the cap.
- **`types.rs` / `ipc.rs` / `capacity.rs` / `main.rs` status renderer** — `cpu_headroom` is dropped from `DaemonStatusReport`; `logical_cpus` / `loadavg_1m` / `cpu_idle_fraction` stay as **observations** (that is exactly the evidence an operator tunes the knob from) and the human line says so.
- **Docs** — `defaults/docs/daemon-reference.md` (new "Why there is no CPU term in admission" + "Machine-wide build slot" sections, env table, calibrate section), `defaults/docs/build-gate.md`, `docs/measure-est-cores-per-sweep.md` (marked HISTORICAL). This repo's own `.loom/config.json` drops its now-inert `estCoresPerSweep`.

- [x] **`resolve_dynamic_max_concurrent` = `min(token axis, disk headroom, configured maxConcurrent)` — no CPU estimate term.** The function is now four lines with three terms; `cpu_headroom` is not imported by `work_finder` at all. Covered by the rewritten binding-axis test table, including a case asserting a retired env knob cannot clamp the cap.
- [x] **A machine-wide build-slot lease serializes designated high-CPU stages across concurrent sweeps (bounded wait + telemetry, never a deadlock; degrades open if the lock dir is unavailable).** `build_slot::acquire` waits at most `LOOM_BUILD_SLOT_WAIT_SECS` (300s) then returns `DegradedOpen`; an unusable lock store degrades open *immediately* (no spin); `LOOM_BUILD_SLOT_HELD` makes nested acquires re-entrant no-ops; every transition is logged (acquire with wait time, first wait of a round, release with hold duration, degrade-open with reason). Both directions of the cross-language contract are tested as **real subprocesses**, not just by path-shape parity.
- [x] **`estCoresPerSweep` / `cpuUtilizationTarget` config+env accepted-but-ignored with a deprecation warning (no config break), docs updated.** Any value at any type still parses (tested with out-of-range, string, and bool values) and never disturbs a sibling knob. The notice reaches the reader on **both** channels: the daemon logs it once per process, and `loom-daemon calibrate` prints it to stderr — CLI subcommands return from `main` before `setup_logging()`, so a log-only warning would have been invisible on exactly the path an operator uses to size a host.
- [x] **The #4231 scenario reasoned through under the new model.** #4231 was a 6-way fan-out that drove load to 118 on a 28-core host, because six concurrent release builds looked like `6 × estCoresPerSweep(2.0) = 12` cores of demand to a term whose real demand was closer to `6 × 28`. Under this change the same fan-out admits 6 sweeps (the knob decides that, not an estimate) but **only one compiles at a time** — the other five block in `build-gate.sh`'s slot acquire — so peak demand is one build's worth of parallel `cargo`, not six. The formula that mis-estimated the fan-out is gone; the thing it was estimating is now directly bounded. If the knob is nonetheless set too high, the #4235 host breaker trips on **measured** load-per-core: a measurement, not a constant, is the last line of defence. Written up in `daemon-reference.md` § "The #4231 release-build fan-out under this model".

Also addressed from the operator's follow-up comment ("give it ONE unambiguous home … nothing in `loom-daemon status` names which file supplied the number"): slots live machine-wide at `~/.loom/locks/build-slot`, deliberately *not* per-workspace, and calibrate now names the tier file that set `maxConcurrent` (and says so when an env var shadows it). See the *deferred* note below for the `status`-side half.

## Test Plan

- `cargo test --workspace` — **2,620 passed, 0 failed** (2,497 in the daemon lib).
- `cargo clippy --package loom-daemon --package loom-api --all-targets` with the CI allow-list — clean under `-D warnings`.
- `cargo fmt --all -- --check` — clean.
- `bash defaults/scripts/tests/test-build-slot.sh` — **13/13** (acquire/release, bounded wait, degrade-open on a broken store, `LOOM_BUILD_SLOTS=0` opt-out, multi-slot budget, stale reaping, re-entrancy, path-shape parity, and that `build-gate.sh` sources the helper).
- `bash defaults/scripts/tests/test-loom-daemon-start.sh` — **60/60** (the calibrate start-time hint was re-based off the removed recommendation fields).
- `bash defaults/scripts/tests/test-config-resolver.sh` — **14/14** (cross-language conformance fixture still matches).
- New Rust coverage: 17 `build_slot` tests (including two genuine **cross-process** Rust↔Bash serialization tests in both directions), 5 `config_resolver::source_of` tests, 5 calibrate provenance tests, 3 `deprecated_cpu_knob_notice` tests.
- Live smoke: `loom-daemon calibrate` on this host renders the new report, prints the deprecation notice to stderr (this host really does have `LOOM_EST_CORES_PER_SWEEP` exported), names `set by: …/.loom/config.json`, and keeps `--json` stdout clean.

### Pre-existing failures (not caused by this change)

Two integration tests are load-flaky on this 8-core host and fail **identically on `main`** at the same commit — `integration_singleton_guard::test_stale_socket_is_reclaimed` (connect retry budget) and `integration_basic::test_create_terminal_with_working_dir` (tmux session dies before capture). Both pass in the full-suite runs above; verified against `main` rather than assumed.

## Notes for the reviewer

- **Deliberately deferred**: `DaemonStatusReport` does not gain a `configured_max_source` wire field. Doing it properly means threading provenance through `ipc.rs` + the status renderer + wire-compat defaults, and `status` already has the harder ambiguity that on a multi-workspace daemon the *cap itself* is resolved from the seeded default workspace. Calibrate is the sizing tool and is purely file-based, so it is the right first home; a follow-up can extend it to `status`.
- **Config safety**: the deprecation path is intentionally asymmetric — retired keys are recorded but never *parsed* (no range filter, no type coercion), because nothing consumes the value and a fleet must be able to upgrade the binary before editing every repo's committed config.
- `LOOM_BUILD_SLOTS` defaults to **1** (the issue allows 1–2); `0` restores the exact pre-change unserialized behavior for an operator who wants it.

Closes #4512

### DIFF
--- a/.loom/config.json
+++ b/.loom/config.json
@@ -15,8 +15,7 @@
     },
     "workFinder": {
       "maxConcurrent": 10
-    },
-    "estCoresPerSweep": 1.25
+    }
   },
   "buildGate": {
     "enabled": true,
--- a/defaults/docs/build-gate.md
+++ b/defaults/docs/build-gate.md
@@ -349,9 +349,61 @@ load average overstates consumption on macOS (see "measured idle fraction",
 #4031), so an operator on a chronically-loaded host may raise `loadThreshold`.
 
 Explicitly **out of scope** here (separable follow-ups): a dedicated gate cargo
-target-dir / shared build-lock to isolate contention (never shipped by #4020);
-per-test timeouts (would require adopting cargo-nextest); and any change to sweep
-spawn priority (that is #4233).
+target-dir to isolate contention (never shipped by #4020); per-test timeouts
+(would require adopting cargo-nextest); and any change to sweep spawn priority
+(that is #4233). The **shared build lock** half of that first item did land
+later — see the next section.
+
+### Machine-wide build slot (#4512)
+
+`nice` (#4020) reorders the run queue and the gate-in-flight flag (#4084) holds
+*dispatch* off one root, but neither stops the genuinely concurrent case: N
+sweep worktrees each running their own post-Builder `build-gate.sh` in separate
+processes, plus the daemon's own main-health gate, all compiling at once. Until
+#4512 the daemon's admission formula tried to prevent that up front with a CPU
+estimate (`estCoresPerSweep × cpuUtilizationTarget` minus live idle) — which
+priced *every* sweep as a build and throttled a 95%-idle 8-core host to 2
+concurrent sweeps. #4512 deleted that term and moved the protection here, to the
+stage that actually burns the cores.
+
+**Every invocation of `build-gate.sh` takes one machine-wide build slot before
+compiling**, and releases it on exit (via an `EXIT` trap, so a failed or killed
+gate never leaks the slot). N sweeps run concurrently; at most
+`LOOM_BUILD_SLOTS` (default **1**) of them build at any moment.
+
+- **Mechanism**: `defaults/scripts/lib/build-slot.sh`, sourced by
+  `build-gate.sh` (`loom_build_slot_acquire` / `loom_build_slot_release`). A
+  slot is a lock **directory** created with `mkdir` — POSIX-atomic and available
+  on stock macOS, unlike `flock` — at `~/.loom/locks/build-slot/slot-<i>`,
+  machine-wide rather than per-repo, because cores are a machine-level resource
+  shared by every workspace and worktree on the host. The daemon's Rust half
+  (`loom-daemon/src/build_slot.rs`) implements the identical protocol and wraps
+  the main-health gate's own command, so the two serialize against each other.
+- **Never blocks the gate indefinitely**: the acquire waits at most
+  `LOOM_BUILD_SLOT_WAIT_SECS` (default 300s) and then **degrades open**,
+  proceeding unserialized. A wedged holder costs one build's worth of
+  serialization, never the gate's liveness.
+- **Never fails the gate**: an unusable lock directory (unwritable, missing
+  parent, a file in the way) also degrades open, with a warning. `mkdir` locks
+  carry no heartbeat, so an abandoned slot is reaped after
+  `LOOM_BUILD_SLOT_STALE_SECS` (default 1h — deliberately long, so a peer cannot
+  reap a healthy in-progress build).
+- **Re-entrant**: when the daemon already holds a slot around this gate command
+  it exports `LOOM_BUILD_SLOT_HELD=1`, and the acquire inside `build-gate.sh` is
+  a logged no-op instead of a wait on its own parent's slot.
+- **Opt out** with `LOOM_BUILD_SLOTS=0` (every acquire degrades open — exactly
+  the pre-#4512 behavior). If `lib/build-slot.sh` is missing from an install the
+  gate prints a note and runs unserialized.
+- **Interaction with the timeout**: the daemon acquires the slot **outside** the
+  `buildGate.timeoutSeconds` window, so queueing behind another build never eats
+  into the gate's own budget and turns contention into a false `UNEVALUATED`.
+- **Interaction with deferral**: the load-aware deferral above is a *scheduling*
+  decision made before any command runs; the slot is a *mutual-exclusion*
+  decision made at the moment of compiling. They compose — a deferred tick never
+  reaches the acquire at all.
+
+Full rationale, telemetry, and the env-knob table live in
+[`daemon-reference.md` → Machine-wide build slot](daemon-reference.md).
 
 > **Rule of thumb:** if a check's outcome can differ between an idle host and a
 > busy one — or between a host with tmux and one without — it is
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -1550,44 +1550,100 @@ The concurrency cap is **not** a fixed value resolved once at startup. Every
 tick the finder recomputes
 
 ```
-dynamic_cap = min(healthy-token count × per-token concurrency, disk headroom, cpu/load headroom, configured ceiling)
+dynamic_cap = min(healthy-token count × per-token concurrency, disk headroom, configured maxConcurrent)
 ```
 
-from live inputs, so pool/disk/cpu/backlog changes are honored without a
-daemon restart:
+from live inputs, so pool/disk/backlog changes are honored without a daemon
+restart. A fourth **cpu/load headroom** term sat in this `min(...)` from #3978
+until **#4512 deleted it** — see [Why there is no CPU term in
+admission](#why-there-is-no-cpu-term-in-admission-4512) below.
 
 | Input | Source | Bound it enforces |
 |-------|--------|-------------------|
 | **healthy-token count** | `available` accounts in `.ranking` in the pool directory `tokens_pool::paths::resolve_tokens_dir` resolves for the workspace — per-repo `{workspace}/.loom/tokens/` when it holds `*.token` files, else the shared machine-level pool (#3938) (`capacity::read_ranking` / `token_axis_limit`, unified with the writer in #4344 — pre-#4344 this hardcoded the per-repo path even on a shared-pool host, which could pin the dispatch cap at 0 against an orphaned per-repo `.ranking` indefinitely), falling back to the `*.token` count (`tokens::token_pool_size`) when no ranking exists | the count of accounts safe to dispatch to — never dispatch to an exhausted/blocked one (#3902) |
 | **per-token concurrency** | `LOOM_PER_TOKEN_CONCURRENCY` / `autonomous.perTokenConcurrency`, default **2** (#3947) | how many concurrent sweeps to allow **per healthy account**. A plan limit is a utilization-window token bucket, not a session count, so one healthy account can run several concurrent sessions. Before #3947 the implicit factor was `1` (one sweep per account), which collapsed the whole fleet to cap 1 when 6/7 accounts were at their weekly ceiling even though the single healthy account had ample session-window headroom |
 | **disk headroom** | `floor(free_gb / LOOM_PER_WORKTREE_GB)` on the worktree-root volume (`disk_headroom::disk_headroom_limit`, a Rust port of `disk-headroom.sh` that shells to `df -Pk`) | never provision more worktrees than the scratch volume can hold |
-| **cpu headroom** (#3978, measured-idle signal #4031) | `max(1, floor((logical_cpus × LOOM_CPU_UTILIZATION_TARGET − consumed_cores) / LOOM_EST_CORES_PER_SWEEP))`, where `consumed_cores = logical_cpus × (1 − idle_fraction)` from the measured idle fraction (loadavg fallback) (`cpu_headroom::cpu_headroom_limit`) | never start more concurrent sweeps than the host's CPU headroom can currently absorb |
-| **configured ceiling** | `LOOM_WORK_FINDER_MAX_CONCURRENT` (repurposed from Phase A's fixed target into an operator ceiling) | hard operator upper bound regardless of token/disk/cpu headroom |
+| **configured maxConcurrent** | `LOOM_WORK_FINDER_MAX_CONCURRENT` / `autonomous.workFinder.maxConcurrent` (repurposed from Phase A's fixed target into an operator ceiling) | **the** per-machine admission knob (#4512) — tuned empirically by the operator, the only *policy* term in the `min(...)` (the other two meter exhaustible resources: accounts and bytes) |
 
-**CPU headroom term (#3978, measured-idle signal #4031).** The token and disk
-axes alone let a batch of token accounts resetting from `exhausted` to
-`available` at once raise the cap regardless of how many concurrent `cargo
-build`s were already saturating the host — the incident this term fixes: 2–3
-concurrent Rust builds in sweep worktrees starved `build-gate.sh` of CPU badly
-enough that it hit its own 600s timeout, which the (separately-fixed, #3974)
-gate misreported as a verified-red `main`, halting all dispatch.
-`cpu_headroom_limit()` combines a **static** capacity (`logical_cpus ×
-utilization_target`, default target `0.75` — leaves headroom for the OS, the
-daemon itself, and the gate's own `cargo` invocations) with the cores
-**currently consumed**, subtracted from that capacity, divided by an estimated
-per-sweep core cost (`LOOM_EST_CORES_PER_SWEEP`, default `2.0`).
+#### Why there is no CPU term in admission (#4512)
+
+From #3978 until #4512 the `min(...)` carried a fourth term:
+
+```
+cpu_headroom = max(1, floor((logical_cpus × cpuUtilizationTarget − consumed_cores) / estCoresPerSweep))
+```
+
+It was built for a real incident — 2–3 concurrent Rust builds in sweep
+worktrees starved `build-gate.sh` of CPU badly enough that it hit its own 600s
+timeout, which the (separately-fixed, #3974) gate misreported as a verified-red
+`main`, halting all dispatch — and hardened over #4031/#4234. **#4512 deleted
+it anyway**, because it had become the binding constraint on fleet throughput:
+
+- It priced **every** sweep as a build (`estCoresPerSweep`). Sweep wall-clock is
+  dominated by **API wait** (curator / builder / judge conversations); the
+  sustained-heavy-CPU phases (release builds, full test suites, the build gate)
+  are a small fraction of it. So the term throttled the ~90% low-CPU majority to
+  defend against the minority case.
+- Measured evidence (8-core EC2 worker, 2026-07-29): the host sat **95% idle**
+  while the formula capped concurrency at **2**, and `loom-daemon calibrate`
+  recommended "no change" — the formula could not see past its own estimate.
+
+The replacement is two-part, and the protection moved to **where the load
+actually is**:
+
+1. **Admission is one per-machine knob.** `autonomous.workFinder.maxConcurrent`
+   is the whole policy, tuned empirically by the operator (expect 10+ on an
+   8-core API-bound worker). The two remaining terms — the token axis and disk
+   headroom — stay because they meter genuinely *exhaustible* resources
+   (accounts, bytes), not an estimate.
+2. **The genuinely heavy stages serialize where they occur**, via the
+   [machine-wide build slot](#machine-wide-build-slot-4512). N sweeps run
+   concurrently; at most `LOOM_BUILD_SLOTS` of them hold the slot while running
+   `cargo build`/`test` or the build gate; the rest of a sweep's lifecycle never
+   queues.
+
+The **host-distress circuit breaker** (#4235) is unchanged and is what makes a
+hand-tuned knob safe to raise: a mis-set `maxConcurrent` trips a **measured**
+breaker (load-per-core), instead of melting the host on the strength of a
+mis-estimated constant.
+
+`cpu_headroom.rs` survives as a pure **measurement** module — logical core
+count, load average, measured idle fraction, and the `is_host_saturated`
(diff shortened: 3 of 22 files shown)


rjwalters:
✅ **Approved!**

Reviewed the full diff (22 files, +3052/-1997) in the existing builder worktree (`.loom/worktrees/issue-4512`).

**Correctness/design**
- `resolve_dynamic_max_concurrent` now cleanly `min(token axis, disk headroom, configured maxConcurrent)` — the CPU term and its threading through `WorkFinderConfig`/`spawn_multi_work_finder_task` are fully removed, not just short-circuited.
- New `build_slot.rs` (Rust) and `defaults/scripts/lib/build-slot.sh` (Bash) implement matching `mkdir`-atomic slot-lease protocols. Verified the three safety properties in code: bounded wait → degrade-open (never blocks past `LOOM_BUILD_SLOT_WAIT_SECS`), degrade-open on any lock-store failure (no spin), and re-entrancy via `BUILD_SLOT_HELD_ENV` (explicitly set `0`/`1` on the child env rather than relying on inheritance, which avoids accidentally propagating a stale sentinel from a grandparent).
- `main_health_gate.rs` acquires the slot *outside* the timeout window before calling `run_command_with_timeout` — correctly avoids turning slot-queue wait into a false `Unevaluated` (the #3978/#4020 failure mode this PR calls out).
- Deprecated `estCoresPerSweep`/`cpuUtilizationTarget` config+env are accepted-but-ignored (no range filtering/coercion since nothing consumes them) with a one-shot warning on both the daemon-log and CLI-stderr paths (correctly reasoned: CLI subcommands return from `main` before `setup_logging()`). This repo's own `.loom/config.json` had its now-inert `estCoresPerSweep` removed.
- New `config_resolver::source_of`/`tier_paths_by_precedence` correctly returns the highest-precedence tier that sets a key (non-null), skips malformed tiers, and reports `None` for an explicit `null` — verified via the new tests.

**Verification performed**
- `cargo test --package loom-daemon build_slot --lib` → 17/17 pass, including the two genuine cross-process Rust↔Bash serialization tests.
- `cargo test --package loom-daemon work_finder --lib` → 117/117 pass.
- `cargo fmt --all -- --check` → clean.
- `cargo clippy --package loom-daemon --all-targets -- -D warnings` → clean.
- `gh pr checks 4604` → all required checks pass; `mergeStateStatus` is `CLEAN`.
- Confirmed `Closes #4512` is valid (issue is open, not an epic/family issue requiring a partial-increment reference).

## Test Execution

1. `cargo test --workspace` (2,620 passed) — ✅ Executed a scoped subset (`build_slot` + `work_finder`, 134 tests) rather than the full workspace; all pass, matching the PR's reported full-suite result.
2. `bash defaults/scripts/tests/test-build-slot.sh` (13/13) — ⏭️ Skipped: redundant with the Rust cross-process tests already verified, and out of scope for the time-bounded review pass.
3. `bash defaults/scripts/tests/test-loom-daemon-start.sh` / `test-config-resolver.sh` — ⏭️ Skipped: same reasoning; `config_resolver` unit tests reviewed directly instead.
4. Live smoke (`loom-daemon calibrate` on this host) — ⚠️ Skipped: requires manual/host-specific observation.

No blocking issues found. Well-reasoned design docs, thorough test coverage for a genuinely risky concurrency primitive, and CI is green.

rjwalters:
✅ **Approved!**

Substantial, well-executed refactor of the daemon's admission cap. Reviewed in the existing builder worktree (`.loom/worktrees/issue-4512`).

**Design**: Dropping the CPU-headroom admission term (`cpu_headroom = f(logical_cpus, cpuUtilizationTarget, estCoresPerSweep)`) in favor of `min(token axis, disk headroom, configured maxConcurrent)` plus a machine-wide build slot is a sound move — the module docs in `build_slot.rs` and the updated `work_finder.rs` doc comments lay out the reasoning (API-wait-dominated majority vs. build-heavy minority) clearly, and correctly identify where the moved protection needs to live (the actual heavy-CPU stage, not admission).

**Correctness spot-checks**:
- `build_slot.rs`: the three safety properties (bounded-wait-then-degrade-open, degrade-open on unusable lock store, re-entrancy via `LOOM_BUILD_SLOT_HELD`) are all directly tested, including two genuine cross-process Rust↔Bash tests in both directions — I re-ran these locally (`cargo test -p loom-daemon --lib build_slot`, 17/17 pass).
- `defaults/scripts/lib/build-slot.sh`: mirrors the Rust protocol (path shape, stale-reap threshold, degrade-open semantics); ran `test-build-slot.sh` locally, 13/13 pass.
- `MkdirLock::try_acquire` (`tokens_pool/locking.rs`) cleanly factors out the shared `is_stale` helper and adds a non-blocking variant without disturbing the existing blocking `acquire` path.
- `main_health_gate.rs`: slot is acquired **outside** the timeout window (avoids turning a queue wait into a false `Unevaluated`, the #3978/#4020 failure mode) and correctly propagates `covers_children()` into the child env so `build-gate.sh`'s own acquire is a re-entrant no-op.
- Wire/config compat is handled carefully: `resolve_capacity`'s client-side recomputation treats a wire `cpu_headroom == 0` from an older daemon as "field absent" rather than a genuine zero (correct `#[serde(default)]` reasoning), and the deprecated `cpuUtilizationTarget`/`estCoresPerSweep` knobs are accepted-but-ignored (never a config parse break) with a deprecation notice delivered on both channels a fleet operator would actually see (daemon log + calibrate's stderr, since CLI subcommands return before `setup_logging()`).

## Test Execution

1. `cargo test --workspace` (2,620 passed) — ⚠️ Not re-run in full (long-running); scoped re-run of the new module instead: `cargo test -p loom-daemon --lib build_slot` — ✅ Executed: 17/17 pass. Full suite corroborated by CI's "Rust Unit Tests" check (pass).
2. `cargo clippy` / `cargo fmt --check` — ⏭️ Not re-run locally; CI's "Rust Linting & Build Check" confirms clean.
3. `bash defaults/scripts/tests/test-build-slot.sh` — ✅ Executed: 13/13 pass, matching the PR's claim exactly.
4. `test-loom-daemon-start.sh` / `test-config-resolver.sh` — ⏭️ Skipped (not independently re-run); no evidence of regression, covered by CI.
5. Live smoke (`loom-daemon calibrate` deprecation notice) — ⚠️ Skipped: requires manual/host-specific observation.

**CI**: all checks pass (`gh pr checks`), merge state `CLEAN`.

rjwalters:

- Size check: 5049 lines changed (+3052/-1997), exceeds the 200-line auto-merge limit. No `loom:auto-merge-ok` label present.

This is also a substantial daemon refactor (touches `build_slot.rs`, `capacity.rs`, `cpu_headroom.rs`, `work_finder.rs`, `config_resolver.rs`, and related core scheduling logic), so it warrants human review beyond just the size threshold.

All other safety criteria pass (Judge-approved, no critical-file conflicts, mergeable, recently updated, CI green).

**Next steps:**
- A human can review and merge manually via `./.loom/scripts/merge-pr.sh 4604`
- Or add the `loom:auto-merge-ok` label if this large-but-safe PR should be waived through on the next tick

---
*Automated by Champion role*

rjwalters:

- **Size check**: 5,049 lines changed (+3052/-1997), exceeding the configured auto-merge limit of 200 lines. No `loom:auto-merge-ok` override label is present.

All other safety criteria pass (Judge-approved, no critical-file matches, mergeable, recently updated, CI green).

**Next steps:**
- A human reviewer should evaluate this refactor directly given its size (touches daemon capacity/admission internals across ~20 files), or
- Apply the `loom:auto-merge-ok` label if a human has judged the size acceptable, and Champion will merge on the next tick.

---
*Automated by Champion role*

rjwalters:
**Review: Approved.**

Verified independently (fresh checkout of the builder's worktree at `.loom/worktrees/issue-4512`, HEAD `8a0c4b4`):

## Diff cleanliness (stale-branch-base incident)
- `git diff origin/main...HEAD` is a single squashed commit, 22 files, scope matches the PR description exactly (concurrency admission + build slot + calibrate + docs). No leftover conflict markers (`grep -E '^(<{7}|={7}|>{7})'` across every touched file: none found).
- Every touched file's diff is coherent with the stated intent — spot-checked the smallest/oddest-looking hunks (`serve.rs`, `.loom/config.json`, `host_breaker.rs`) and each is a clean, scoped removal/update, not a stray reversal artifact.
- No trace of the described bad `git reset --soft` — the final PR branch is one commit ahead of `origin/main`, cleanly.

## Correctness
- `resolve_dynamic_max_concurrent` is now the claimed 3-line `min(token axis, disk headroom, configured_max)` — CPU term fully gone, not imported by `work_finder` at all.
- `DaemonStatusReport.cpu_headroom` removal is wire-compatible in both directions: serde ignores unknown fields by default, and there's an explicit test (`ipc.rs`) plus a `main.rs` comment confirming a pre-#4512 daemon's extra `cpu_headroom` field is tolerated by a new CLI, and vice versa.
- Deprecated `cpuUtilizationTarget`/`estCoresPerSweep` (config + env) are accepted-but-ignored, never a parse error — confirmed live via `loom-daemon calibrate` on this host (which really does have `LOOM_EST_CORES_PER_SWEEP` set): the deprecation notice prints to stderr, `--json` stdout stays clean of it, and `set by: <path>` provenance is correct.

## Build-slot safety (new cross-process primitive)
Read `build_slot.rs`, `tokens_pool/locking.rs` (`MkdirLock::try_acquire`), and the Bash mirror `build-slot.sh` end to end:
- **Atomicity**: `mkdir` (not `flock`), so both languages share one POSIX-atomic primitive, machine-wide at `~/.loom/locks/build-slot`.
- **Never deadlocks**: bounded wait (`LOOM_BUILD_SLOT_WAIT_SECS`, default 300s), then degrades open. Verified by test and by reading `main_health_gate.rs`: the slot is acquired **outside** the gate's timeout window specifically so a queue wait can't turn contention into a false `Unevaluated` (the #3978/#4020 failure mode this PR is aware of and avoids repeating).
- **Degrades open on a broken lock store** (unwritable/missing dir) immediately, no spin.
- **Re-entrancy** is correct and race-safe: `main_health_gate.rs` always sets `LOOM_BUILD_SLOT_HELD` explicitly to `"1"`/`"0"` on the child (never leaves it unset), so a stray inherited env var can't cause a false re-entrant no-op.
- **Stale reaping** (1h threshold, deliberately long vs. the 30s token-pool default) is justified and correctly distinguishes "all slots busy" from "lock store unusable" in `try_round`/`_loom_build_slot_try_round` (both languages), which matters for the DegradedOpen-vs-wait decision.
- Ran the real cross-process tests myself (not just trusting the claim): `a_rust_held_slot_blocks_the_bash_half_across_processes` and `a_bash_held_slot_blocks_the_rust_half_across_processes` both pass — these spawn real `bash` subprocesses against the Rust lock, exercising the actual cross-language contract, not just path-shape parity.

## `LOOM_BUILD_SLOTS=0` as a kill-switch
Tested directly (`LOOM_BUILD_SLOTS=0 loom-daemon calibrate --json` → `"build_slots": 0`) and via the bash test suite — the *mechanism* is safe and complete: acquire is a no-op, no lock dir is even created, and both language halves agree.

One documentation precision nit (non-blocking): the wording "`0` … exactly the pre-#4512 behavior" (in `build_slot.rs`'s doc comment and the `daemon-reference.md` env-var table) slightly overstates what the knob restores. Pre-#4512 there was no build slot *and* the CPU-headroom term throttled admission (the 8-core-idle-host-capped-at-2 case). Post-#4512 with `LOOM_BUILD_SLOTS=0`, there's still no build-slot serialization, but the CPU-headroom admission throttling is now unconditionally gone (accepted-but-ignored, not re-enabled by any knob) — so an operator who sets `maxConcurrent` high and `LOOM_BUILD_SLOTS=0` gets a more permissive combination than "pre-#4512", relying entirely on the independent, always-on host-distress breaker (#4235) as the safety net rather than getting the old formula back. Worth a small wording tweak in a follow-up so "kill switch" isn't read as "identical old behavior" — but the actual code behavior is safe (tested, degrades open, never blocks) either way, so this doesn't block approval.

## Scoping: deferred `configured_max_source` on `DaemonStatusReport`
Reasonable call. The PR ships the equivalent capability (`config_resolver::source_of` + `tier_paths_by_precedence`, surfaced via `calibrate`'s `set by: <path>` line) for the actual use case (an operator sizing a host), and correctly identifies that wiring it into `status` is harder (multi-workspace ambiguity — which workspace's cap is even "the" cap — plus threading provenance through `ipc.rs`/wire-compat defaults). Deferring it with a stated follow-up path rather than doing it half-right under this PR's scope is the right tradeoff.

## Test verification (ran independently, not just trusting the PR body)
- `cargo test -p loom-daemon --lib` → **2,497 passed, 0 failed** (matches the claimed count exactly).
- `cargo test -p loom-daemon --lib build_slot` → **17/17 passed**, including both cross-process tests.
- `cargo clippy --package loom-daemon --package loom-api --all-targets -- -D warnings` → clean.
- `cargo fmt --all -- --check` → clean.
- `bash defaults/scripts/tests/test-build-slot.sh` → **13/13 passed**.
- The two flaky integration tests the builder called out (`integration_singleton_guard::test_stale_socket_is_reclaimed`, `integration_basic::test_create_terminal_with_working_dir`) have **zero diff** on this branch vs. `origin/main` — confirms they're untouched by this PR, consistent with "flaky on this host, not a regression."
- CI: all required checks green (`Rust Unit Tests`, `Rust Linting & Build Check`, `Installer Integration Tests`, `Shellcheck`, etc. — `Rust Unit Tests` was still in-progress when I started reviewing, re-checked and it completed successfully).
- `mergeStateStatus`: `CLEAN`, `mergeable`: `MERGEABLE`.
- PR description: exactly one closing keyword (`Closes #4512`, line 55), no stray closing keywords targeting other issues.

Issue #4275, opened 2026-07-28
feat(daemon): port the script-helper family (model_tiers, usage, checkpoints, …) to native loom-daemon (epic #4081 phase 3, family 5)


rjwalters:
## Epic context

Epic #4081 Phase 3, family 5 of 5 (marker: `loom:epic:4081:phase:3`). Filed by Champion after the Phase 3 gate cleared (#4228 closed COMPLETED via PR #4266; native tokens cutover substance-verified on `origin/main` @ `db0767e2`).

## Problem

The script-helper Python modules back a scatter of small shell entry points: `log_filter` (strip-ansi.sh), `model_tiers` (resolve-model.sh), `usage` (check-usage.sh), `validate_phase`, `checkpoints`, `claim`, `sweep_experiment`, plus `backlog` (docs-only references).

## Scope: port or fold

Per the epic body (family 5): port each helper natively or fold it into its calling script; `backlog` needs only docs cleanup. Two Rust-side couplings to mind:
- `checkpoints` — referenced from `loom-daemon/src/init/post_init.rs`.
- `sweep_experiment` — referenced from `loom-daemon/src/activity/resource_usage.rs`.

**Porting contract folded from #4060** (binding on `model_tiers` / `sweep_experiment`):
- `--config <path>` is an **explicit bypass**: reads exactly that file, skips all tiering; only the default path routes through the (Rust) config resolver. Semantics identical between `model_tiers` and `sweep_experiment` — experiment Arm A must resolve the same model as `resolve-model.sh`.
- `load_config` equivalents **never raise** — no-repo-root degrades to empty config so `resolve-model.sh` works outside a Loom repo.
- Model-alias normalization is currently a **matched pair** with `sweep_registry.rs::read_model_aliases`; this port collapses the pair to one Rust implementation (retiring the byte-identical-twins constraint — see also the known Rust-tiered / Python-flat divergence in `sweep.modelAliases`). The split-tier alias-map merge decision must match #4059's answer.

Epic contracts: callers switch in the same PR; script names/flags persist; no new Python.

## Out of scope

- Other Phase 3 families (file-disjoint).
- Changing model-tier policy (the `opus` alias question, #3981/#3982) — port resolution behavior as-is.

## Acceptance criteria

- [ ] All listed helpers have no remaining external callers (`python3 -m` or console scripts) across the six epic search roots; `backlog` doc references cleaned.
[...]

PR #4602, opened 2026-07-30
fix(daemon): hold the issue lock across the midbuild-watchdog worktree clean

rjwalters:
## Summary

The mid-build-death watchdog gated its destructive `clean_worktree` on a **read-only** ownership probe (`lock_owned_by_other`, #4463). That narrowed but did not close a **probe→clean TOCTOU**: if the claim lock was free at probe time and a cross-instance sweep acquired it microseconds later, the watchdog would still `git reset --hard` a worktree a *newly live* sweep had just claimed — the #4449 data-loss shape. Flagged by the approving Judge on PR #4514 as out-of-scope residual risk; this PR closes it by **acquiring** the lock instead of probing it.

## Changes

`loom-daemon/src/sweep_registry.rs` (only file touched):

- **New `claim_lock_for_midbuild(issue, dead_sweep_id) -> Option<String>`** — takes *exclusive ownership* of `.loom/locks/issue-<N>` and returns the watchdog's own `sweep_id` (`midbuild-watchdog-<dead_sweep_id>`), or `None` when the claim belongs to someone else. Two paths, **neither of which ever leaves the lock momentarily free** (a release→re-acquire pair would expose exactly the `mkdir`-sized window being closed):
  - *No lock dir* → `acquire_lock`'s POSIX-atomic `mkdir`, the same primitive `dispatch` uses, so a racing peer loses the `mkdir` and exactly one of the two proceeds.
  - *Lock dir present* → refuse when `owner.json` positively names a **different** sweep (fail-open on unreadable/corrupt, as #4463 established); otherwise take the dead sweep's own stale claim over **in place** by rewriting `owner.json`. **Fail-closed** if that write fails: without the lock, no clean.
- **`midbuild_watchdog_once`'s `Recover` arm** now claims the lock instead of probing it. The claim runs *before* the `midbuild_retried` latch, so a lost claim does **not** consume the issue's single recovery retry. After `clean_worktree`, the watchdog releases its own claim via `release_lock_owned(issue, &watchdog_lock_id)` (ownership-checked against the **watchdog's** id) so `dispatch` re-acquires atomically under its fresh sweep id. This also subsumes the old defensive release of a reconstructed entry's un-released lock — that lock is now taken over in place and released here.
- Doc comments on both the new helper and `midbuild_watchdog_once` record the invariant and why the in-place takeover (rather than release+re-acquire) is load-bearing.

No new event topics, no config, no behavior change on the `InUse` / `GiveUp` / token-gate arms.

[...]

PR #4605, opened 2026-07-30
fix(daemon): refuse dispatch/recovery for an issue with a confirmed-live sweep claim (#4556)

### LINKED ISSUE #4556: daemon: issue #4275 dispatched 7 times in 77 minutes despite --claim-owned — recurrence of #4463 (or an uncovered variant)
## Context

Observed live on this host (2026-07-30) while running `/loom:sweep 4275 --claim-owned 4275` as a daemon-dispatched child. `.loom/logs/sweep-issue-4275.log` shows **seven separate daemon dispatches for the same issue** across a 77-minute window, all carrying `--claim-owned 4275` / `LOOM_SWEEP_CLAIM_OWNED=4275`:

```
sweep_id=sweep-issue-4275-1785379543  dispatched 2026-07-30T02:45:44Z
sweep_id=sweep-issue-4275-1785380933  dispatched 2026-07-30T03:08:54Z
sweep_id=sweep-issue-4275-1785383664  dispatched 2026-07-30T03:54:25Z  (logged twice, 03:54:25.407 and 03:54:25.875 — same sweep_id)
sweep_id=sweep-issue-4275-1785384055  dispatched 2026-07-30T04:00:56Z
sweep_id=sweep-issue-4275-1785384118  dispatched 2026-07-30T04:02:00Z
sweep_id=sweep-issue-4275-1785384150  dispatched 2026-07-30T04:02:32Z
```

At least four of these were alive concurrently around 04:00–04:02 (`ps` showed PIDs 4161338, 412129, 507310, 536458 all running `claude -p /loom:sweep 4275 --claim-owned 4275` simultaneously), all pointed at the single shared worktree `.loom/worktrees/issue-4275` / branch `feature/issue-4275`.

One sibling's Builder was mid-write (steady-cadence mtimes advancing on untracked Rust sources) when a later-dispatched sibling's Builder entered the same worktree. The later Builder detected the live writer and yielded rather than colliding — this time. Had it not detected the concurrent activity, two Edit/Write streams into the same files would have produced a corrupted tree, not a faster PR. The eventual result (PR #4552, merged) succeeded only because every sibling but one either yielded or stalled before reaching its own write phase.

Downstream effects of the same duplicate-dispatch storm, also observed on this run:
- The GitHub GraphQL API rate limit (5000/hr) was fully exhausted (`used: 5001, remaining: 0`) partway through this single issue's lifecycle, most plausibly from N concurrent sweep sessions each independently polling issue/PR state via `gh pr view` / `gh issue view` (which use GraphQL). This degraded polling for every sweep on the host, not just the duplicates.
- The issue's `loom:building` label was observed to drift back to `loom:issue` mid-flight on at least one occasion (per one sibling's own report), which would make it visible to the work-finder again and risks re-dispatch — a possible feedback loop that could compound the duplication.

## Relationship to #4463

#4463 ("daemon: duplicate same-host dispatch — two concurrent sweeps for one issue, 43s apart, sharing one worktree") documented the same class of bug for issue #4426 (2 concurrent dispatches, 43s apart) and was closed as **completed**. This occurrence is either:
- a regression of that fix, or
- a distinct root cause the #4463 fix didn't cover (worth checking: #4463's fix was scoped to "same-host, same-issue dispatch while a live sweep holds the claim" — this occurrence has dispatches spaced from 23 minutes apart down to 32-64 seconds apart, spanning a much longer window, which may indicate a different trigger, e.g. a watchdog/retry path re-dispatching rather than a tick-overlap race).

Whichever it is, the current guard is evidently not sufficient to prevent the same issue from being dispatched 7 times in just over an hour, with up to 4 processes alive concurrently.

- [ ] Root-cause why issue #4275 was dispatched 7 times between 02:45:44Z and 04:02:32Z despite each dispatch carrying `--claim-owned` / `LOOM_SWEEP_CLAIM_OWNED` — check whether these came from the primary work-finder, a watchdog re-dispatch path, or a stale-claim reclaim path re-triggering on a live claim.
- [ ] Verify whether the #4463 fix actually landed/is active on this host's daemon build, or whether this is a distinct code path it didn't cover.
- [ ] Add a hard, dispatch-time guard: refuse (or dedupe) a same-host dispatch for an issue that already has a live sweep process holding the claim, not just the label state.
- [ ] Investigate the `loom:building` → `loom:issue` drift observed mid-run on this issue — if a label revert is happening while a sweep is actively working, that alone could explain repeated re-dispatch by a naive work-finder poll.
- [ ] Consider a per-issue worktree lock (e.g. a lockfile in `.loom/worktrees/issue-<N>/.loom-managed` or a separate lock dir) so a second Builder entering an already-active worktree fails fast/waits instead of relying on mtime-heuristic detection in agent prose.
- [ ] Regression test: N dispatch requests for the same issue within a bounded window produce exactly one live sweep.

## Evidence / paths

- `.loom/logs/sweep-issue-4275.log` on this host — all 7 dispatch headers with sweep_ids and timestamps.
- PR #4552 (issue #4275) merged successfully despite the race — but by luck of timing, not by design.

---

## Curator Enhancement (2026-07-30)

(Complexity marker `loom:complexity=complex` already present at the end of the original issue text above — left as-is, not duplicated.)

### Root-cause forensics (verified against this host's live logs — new evidence)

This host (`/home/ubuntu/GitHub/loom`) is the same host the issue was filed from, and its logs are still present. Cross-referencing `.loom/logs/sweep-issue-4275.log` (the 7 dispatch headers quoted in the issue) against `~/.loom/daemon.log` (the production daemon's own log) resolves several of the open questions below:

**4 of the 7 dispatches are fully explained by the production daemon's own log** — this is not a mystery for these four:

| Dispatch (sweep_id) | Cause | Log evidence |
|---|---|---|
| `...1785379543` @ 02:45:44Z | Normal work-finder pickup | `[02:45:42.523] [INFO] work_finder: dispatching issue #4275 with model=sonnet (source=default)` — `loom-daemon/src/work_finder.rs:2208` (`dispatch()`, the production `WorkDispatcher` impl, not the `#[cfg(test)]` module that starts at `work_finder.rs:2225`) |
| — | **The `loom:building` → `loom:issue` drift is CONFIRMED, not speculative** | `[03:08:15.992] [WARN] claim_reconciliation: reclaimed loom:building -> loom:issue for #4275 in /home/ubuntu/GitHub/loom (DeadPid { pid: 2781227 }, ...)` — `loom-daemon/src/claim_reconciliation.rs`. The reconciler believed the prior sweep's PID was dead and reverted the label, re-opening the issue to the work-finder. |
| `...1785380933` @ 03:08:54Z | **midbuild-watchdog** re-dispatch | `[03:08:52.293] [WARN] midbuild-watchdog: issue #4275 (sweep-issue-4275-recovered-1785380901) died mid-build with a dirty worktree and no PR — cleaning the worktree and re-dispatching once (#3895).` — `loom-daemon/src/sweep_registry.rs:5874` (`midbuild_watchdog_once`) |
| `...1785383664` @ 03:54:25Z (logged twice, same sweep_id — a wrapper-script logging artifact, not two processes) | **review-stall-watchdog** re-dispatch | `[03:54:21.686] [WARN] review-stall-watchdog: sweep for issue #4275 (sweep-issue-4275-1785380933) produced no log output in 2726s ... its review phase (Judge/Doctor) looks hung; auto-cancelling and re-dispatching once (#3910).` — `loom-daemon/src/sweep_registry.rs:6099` (`review_stall_watchdog_once`) |

**The remaining 3 dispatches (`...1785384055` @ 04:00:56Z, `...1785384118` @ 04:02:00Z, `...1785384150` @ 04:02:32Z) have NO corresponding entry anywhere in `~/.loom/daemon.log`.** The production daemon never logged dispatching, reconciling, or watchdog-recovering issue #4275 at those timestamps — meaning these three did not originate from the daemon process that owns that log file.

**Concrete lead for those 3: stray debug-build `loom-daemon` binaries were alive on this host across the exact window.** `ps aux` (captured during this curation pass) shows multiple long-lived background processes launched from `.loom/worktrees/issue-4385/target/debug/loom-daemon` (issue #4385, *"tests(daemon): suite-wide env-mutation race"*, currently `loom:building` — its own integration-test suite spawns real `loom-daemon` binaries via `Command::spawn`), with start times spanning 03:39, 03:42, 03:45, 03:53, 04:20, 04:50, 04:52 (×2), 04:54 — bracketing the unexplained 04:00–04:02 window on both sides. This is the same class of root cause the #4463 curator pass independently found for issue #4426 ("five dispatches... all originated from daemon instance(s) other than the production launchd daemon... multiple loom-daemon instances live during daemon development, all sharing the workspace's `.loom/locks/`"). **This is a lead for the Builder to confirm, not a proven cause** — the ps snapshot shows the processes existed and overlapped in time; it does not by itself prove one of them made the 3 dispatch calls. Confirming it likely means checking whether `.loom/worktrees/issue-4385`'s daemon config points its `workspace_root` at the shared `/home/ubuntu/GitHub/loom` `.loom/` state (rather than an isolated test fixture) — if so, its work-finder tick would see and dispatch against the same live issues.

**On the #4463 relationship (resolves the issue's own open question)**: PR #4514 (the #4463 fix, `loom-daemon/src/sweep_registry.rs`) merged at `2026-07-30T03:44:16Z` — *before* 4 of the 7 dispatch events in this incident (03:54:25 ×2, 04:00:56, 04:02:00, 04:02:32). But PR #4514's actual change is an **ownership-checked lock *release*** (`release_lock_owned` / `LockReleaseOutcome`, gating cancel/reap/midbuild-watchdog cleanup on the lock's current owner) — it prevents a *dying* sweep's cleanup from clobbering a lock a *newer* sweep already re-acquired. It is explicitly **not a dispatch-time guard**: nothing in that diff refuses or dedupes an initial dispatch. So even fully deployed and live on this host, #4463's fix would not have prevented any of these 7 dispatches — it only limits the blast radius (worktree-reset/label-clobber) *after* a duplicate has already been dispatched. **This confirms #4556 is a distinct, complementary code path, not a regression of #4463** — the "hard, dispatch-time guard" this issue's acceptance criteria asks for does not exist yet anywhere in the codebase.

### Updated Acceptance Criteria

- [x] ~~Root-cause why issue #4275 was dispatched 7 times~~ — **done above**: 4/7 attributed to work-finder + claim-reconciliation + midbuild-watchdog + review-stall-watchdog (all legitimate-looking triggers reacting to what each believed was a dead/stalled sweep); 3/7 unattributed in the production daemon's own log, with stray `.loom/worktrees/issue-4385/target/debug/loom-daemon` debug processes as the concrete lead to chase.
- [x] ~~Verify whether the #4463 fix actually landed/is active~~ — **done above**: PR #4514 merged 03:44:16Z mid-incident; its fix is a lock-*release* ownership check, not a dispatch-time guard, so it structurally cannot have prevented any of these 7 dispatches either way. Distinct code path confirmed.
- [ ] Add a hard, dispatch-time guard: before `work_finder.rs:2208`'s `dispatch()` call (and before the `midbuild-watchdog`/`review-stall-watchdog` re-dispatch call sites in `sweep_registry.rs`) reissue a dispatch for an issue, check for a **live** (not just present) lock — a lock whose recorded PID/sweep_id is confirmed running, not merely a file that exists. `acquire_lock` (`sweep_registry.rs:3007`) already refuses on a lock file existing, so the interesting case is: does anything *incorrectly clear or bypass* that lock before a re-dispatch fires? Trace `midbuild_watchdog_once` and `review_stall_watchdog_once`'s call paths to `release_lock`/`release_lock_owned` to confirm they only ever fire once a lock is confirmed genuinely dead.
- [x] ~~Investigate the `loom:building` → `loom:issue` drift~~ — **confirmed, not speculative**: `claim_reconciliation.rs`'s `DeadPid` reclaim fired at 03:08:15.992 for PID 2781227. Determine whether that PID-liveness check has a race (e.g. checking a PID that's between exec-replace steps, or a wrapper-script PID vs the real child PID) that produces false-dead verdicts, since this reclaim is what re-exposed the issue to the work-finder and plausibly seeded the midbuild-watchdog re-dispatch 37 seconds later.
- [ ] **New**: audit whether integration-test-spawned `loom-daemon` binaries (built under `.loom/worktrees/<N>/target/debug/loom-daemon`, exercised by #4385's test suite via `Command::spawn`) run with an isolated `workspace_root`/`.loom/` fixture, or can end up pointed at the real shared workspace and its live `.loom/locks/` + GitHub issues. Cross-reference #4385 (currently `loom:building`) without blocking on it — this issue can root-cause and land its own dispatch-time guard independently of whatever #4385 concludes about its test isolation.
- [ ] Consider a per-issue worktree lock (unchanged from original — still open)
- [ ] Regression test: N dispatch requests for the same issue within a bounded window produce exactly one **live** sweep — not just one *initial* dispatch; the test should also simulate a midbuild-watchdog / review-stall-watchdog tick firing against an issue whose original sweep is still alive and assert it refuses to re-dispatch.

## Affected Files

- `loom-daemon/src/work_finder.rs` — `dispatch()` (~line 2208, production `WorkDispatcher` impl) and `tick`/`tick_multi_with_admission_cap` (~lines 544, 813) — the primary dispatch entry point; add the live-lock check here.
- `loom-daemon/src/sweep_registry.rs` — `acquire_lock` (~3007), `release_lock` / `release_lock_owned` (~2699, ~2711, the #4463 fix), `restore_label_to_ready` (~3534), `midbuild_watchdog_once` (~5874), `watchdog_once` (~5506), `review_stall_watchdog_once` (~6099) — all re-dispatch call sites that need the live-lock guard.
- `loom-daemon/src/claim_reconciliation.rs` — the `DeadPid` reclaim path that flips `loom:building` → `loom:issue`; investigate the PID-liveness check for false-dead verdicts.

Note: the stray-debug-daemon lead (see above) points at `.loom/worktrees/issue-4385/` as a *runtime* artifact observed during this incident, not a source file to modify — it's an ephemeral worktree, not part of the repo tree. If that lead is confirmed, the actual fix would land in whatever daemon startup/config code decides a spawned test binary's `workspace_root` (likely still within the three files above, or in test-harness setup under `loom-daemon/tests/` — the Builder should locate the actual config/test-harness path during investigation rather than treat this as a fixed file list).

## Test Plan

- [ ] Automated: unit test in `sweep_registry.rs` simulating a live sweep (real lock, alive PID) and asserting `midbuild_watchdog_once`/`review_stall_watchdog_once` refuse to re-dispatch (extends the existing `midbuild_refuses_to_wipe_worktree_with_live_claim_lock_owner`-style tests already in that file, e.g. around line 13276).
- [ ] Automated: regression test — N concurrent dispatch attempts (work-finder + both watchdogs) for one issue within a bounded window produce exactly one live sweep.
- [ ] Manual verification: reproduce with a deliberately slow/hung Judge phase and confirm no second dispatch fires while the first sweep's process is still alive.
- [ ] If the stray-debug-daemon lead is confirmed: verify `.loom/worktrees/*/target/debug/loom-daemon` test binaries cannot dispatch against the real shared workspace (either by construction — an isolated fixture `workspace_root` — or by a guard that refuses cross-workspace dispatch).

rjwalters:
Closes #4556

## Problem

Issue #4275 was dispatched **seven times in 77 minutes** onto one shared worktree, with four sweeps alive at once. Every dedup signal the daemon had was either scoped to one process or only proved a file exists:

| Signal | Weakness |
|---|---|
| `in_flight()` / `issue_has_active_sweep` | in-memory; invisible to a second `loom-daemon` on the same host (3 of the 7 dispatches) |
| `loom:building` label | reverted by `claim_reconciliation` the moment a recorded PID *looks* dead (confirmed on #4275 at 03:08:15Z) |
| `.loom/locks/issue-<N>/` (`acquire_lock`) | existence-only — and every reaper / cancel / watchdog path *releases* it first, on the strength of its own dead-sweep verdict, so there was nothing left to collide with |

#4463 made the lock *release* ownership-checked, which stops a dying sweep from clobbering a lock a **newer** sweep re-acquired. But it still trusts the caller's assertion that the sweep it names is dead — the exact assumption that failed here. This PR adds the missing dispatch-time half.

## What this adds

New `loom-daemon/src/live_claim.rs`: a **read-only** probe answering the strictly stronger question — *is a sweep process for issue N confirmed running right now?* Three short-circuiting evidence legs, cheapest first:

1. **Live claim lock** — `.loom/locks/issue-<N>/owner.json` whose `owner_pid` is live. Unlike `acquire_lock`'s existence check this distinguishes a live holder from an abandoned dir, so it is safe to consult *before* a release.
2. **Machine-level sweep journal** — a live record in `~/.loom/sweeps.json`. Matched across *related* repo roots, so a daemon running out of `.loom/worktrees/issue-N` and one running out of the parent checkout see each other's claims (path-component-wise, so `/repo/loom` and `/repo/loom-2` stay unrelated).
3. **Live sweep process scan** — `/proc` scan for a process whose cwd is inside the workspace and whose argv contains `/loom:sweep <N>`. The last-resort leg that catches a sweep no local bookkeeping knows about — the exact signature of #4275's three unattributed dispatches.

Each leg survives a lock release, a label revert, a daemon restart, **and** a second daemon instance.

### Gated at four sites

| Site | Refusal |
|---|---|
| `SweepRegistry::dispatch` **step 2.9** | typed `LiveClaimDispatchError`, before any lock, label write or forge call — so one check covers all six dispatch routes (work-finder, epic supervisor, IPC/CLI, all three watchdogs). The work-finder attributes it to `skipped_in_flight`, never `errors`. |
| `midbuild_watchdog_once` | refuses **up front**. This was the important find: that path `git reset --hard`s the shared worktree and burns its single recovery retry *before* it calls `dispatch`, so a step-2.9 refusal would arrive after the live sweep's uncommitted work was already destroyed. |
| `release_lock_owned` | new `LockReleaseOutcome::HolderAlive` when the lock's own owner PID is still a live `/loom:sweep <N>` process; folded into cancel/reap via a shared `retained()` predicate alongside `Superseded`. |
| `claim_reconciliation` | `apply_live_claim_veto` downgrades a `Reclaim` to `Keep`, so `loom:building` is not reverted out from under a working sweep. |

The review-stall watchdog is deliberately guarded only at dispatch time: it cancels its own child first and takes no destructive action on the worktree, so a refusal after the cancel costs nothing.

### Fail-open by design

Missing/corrupt `owner.json`, unreadable journal, no `/proc`, zombie PIDs, and — critically — a **live PID whose argv does not name this issue's sweep** (a recycled PID behind a stale record) all resolve to "no evidence". Only positively-confirmed liveness ever refuses, so a garbage file can never wedge an issue permanently. On non-Linux hosts, where argv is unreadable, the bookkeeping legs fall back to bare liveness — the same signal every pre-#4556 check already used, so the guard is never *weaker* than the code it supplements.

## Stray-debug-daemon lead: confirmed and closed

The issue's unconfirmed lead is real. `TestDaemon::start()` spawned the real binary with **no cwd override and no `LOOM_WORKSPACES_PATH`**, so every integration-test daemon came up reading the operator's real `~/.loom/workspaces.json` — and this repo's `.loom/config.json` has `autonomous.roleRunner.enabled=true`. A test daemon could therefore run its startup claim-reconciliation pass and autonomous ticks against **real repos and real GitHub issues**.

`isolate_daemon_state()` now confines the workspace registry, sweep journal, watch registry, watch results log, and cwd to a per-test fixture, for all three daemon-spawning test sites. Isolation **by construction**: an empty fixture registry lists no managed workspaces, so a test daemon has nothing to reconcile or dispatch against regardless of what it is asked to do. (`integration_drain_then_exit` had hand-rolled a subset of this; it now uses the shared helper.)

## Per-issue worktree lock — considered, not added

Acceptance criterion 5 asks to *consider* a per-issue worktree lockfile. Rejected as redundant: leg 3 is already a cwd-scoped, per-worktree liveness check that needs no new on-disk state to go stale, and it composes with the existing `#4449` `worktree_in_use` veto (`.loom-in-use` marker / `index.lock` / live process cwd). A fourth lock file would add a new wedge surface for no new coverage.

## Tests

Full `cargo test --lib`: **2523 passed, 0 failed**. `cargo fmt --all --check`, the CI `cargo clippy` invocation, and `cargo check --workspace --all-targets` are all clean. `cargo test --test integration_drain_then_exit` passes.

New coverage:

- `live_claim`: all three legs (live/dead/corrupt/missing/wrong-issue), nested-worktree repo relation, component-wise `repos_are_related`, argv matcher (exact issue token, `/loom:sweeper`, PR-set mode, repeated-needle termination), composed `probe` precedence, and the recycled-PID non-wedge case.
- `dispatch_refuses_when_the_claim_lock_owner_is_live` / `…_journal_records_a_live_sweep_and_the_lock_is_gone` (the gap #4463 could not close) / `…_a_nested_worktree_daemon_holds_the_claim`.
- **Test Plan item 2**: `n_dispatch_requests_for_a_live_issue_produce_zero_extra_sweeps` — five requests against one live issue, all refused, no entry and no lock write.
- **Test Plan item 1**: `midbuild_watchdog_does_not_recover_an_issue_with_a_live_claim` asserts the uncommitted work survives and the single retry is *not* consumed, with a precondition asserting the `#4449` veto is not what refused; `midbuild_watchdog_recovers_once_the_live_claim_is_gone` pins the inverse so the guard cannot wedge recovery. Plus the review-stall equivalent.
- Non-wedge guards: `dispatch_proceeds_when_every_recorded_claim_pid_is_dead`, `dispatch_ignores_a_live_claim_from_an_unrelated_repo`, `release_lock_owned_releases_when_a_live_pid_is_not_this_issues_sweep`.

Tests use a `FakeSweep` RAII stand-in (`sh -c 'sleep 120' '/loom:sweep <N> …'`) rather than the test binary's own PID, so the argv verification is exercised for real. `Command::spawn` returns at *fork*, not exec, so a shared `wait_until_argv_visible` helper blocks until the child's argv is actually observable — without it the argv-verifying assertions flake under a loaded full-suite run (caught during this work: three tests passed alone, failed in the full suite).

## Notes for the reviewer

- `integration_singleton_guard` was **not** run locally: its `cleanup_all_loom_sessions()` kills every `loom-*` tmux session on the host, which would disrupt this machine's live Loom. It compiles and its only change is the isolation call. (That nuclear cleanup is a pre-existing hazard, out of scope here.)
- Acceptance criterion 4 (PID-liveness race in `claim_reconciliation`) is addressed by *vetoing* the reclaim with stronger evidence rather than by re-litigating the recorded-PID check itself — a dead leader PID genuinely can coexist with a working sweep, so the check is not wrong, it is just insufficient on its own.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4543,6 +4543,59 @@ releases it explicitly. On daemon startup, `SweepRegistry::reconstruct`
 admits live-lock owners back into the registry and drops stale locks
 whose owner PID is dead.
 
+### Live-claim guard (#4556)
+
+Lock *existence* is not the same question as "is a sweep for this issue
+running right now?", and every re-dispatch path answers the first while
+needing the second. Issue #4275 was dispatched **seven times in 77
+minutes** onto one shared worktree because each path first convinced
+itself the sweep was dead — the claim reconciler on a dead-looking
+recorded PID, the mid-build watchdog on a terminal registry entry, the
+review-stall watchdog on a silent log — and *released the lock or
+reverted the label before dispatching*, so the existence check had
+nothing left to collide with. Three further dispatches came from a
+second `loom-daemon` on the same host, which shared neither the first
+daemon's memory nor its `.loom/locks/`.
+
+`live_claim::probe` answers the stronger question, read-only, with three
+short-circuiting evidence legs — a live `.loom/locks/issue-<N>/owner.json`
+owner, a live record in the machine-level `~/.loom/sweeps.json` journal
+(matched across related repo roots, so a daemon running out of a nested
+worktree and one running out of the parent checkout see each other), and
+a `/proc` scan for a live `/loom:sweep <N>` process whose cwd is inside
+the workspace. It gates four places:
+
+| Site | Refusal |
+|---|---|
+| `SweepRegistry::dispatch` (step 2.9) | `LiveClaimDispatchError`, before any lock, label write, or forge call — covers all six dispatch routes, including both watchdogs |
+| `midbuild_watchdog_once` | refuses **up front**, because that path `git reset --hard`s the shared worktree and burns its single retry *before* it calls `dispatch` |
+| `release_lock_owned` | `LockReleaseOutcome::HolderAlive` — the caller's dead-sweep verdict was wrong, so the lock stays and the label restore / re-dispatch are skipped |
+| `claim_reconciliation` | vetoes a `Reclaim`, so `loom:building` is not reverted out from under a working sweep |
+
+In `midbuild_watchdog_once` the probe **must** stay ahead of
+`claim_lock_for_midbuild` (#4564/#4602). That call takes the issue lock
+over *in place*, rewriting `owner.json` to the daemon's own PID and a
+`midbuild-watchdog-…` sweep id, and it does so precisely in the
+false-dead case this guard exists to catch. Probing afterwards would
+read the watchdog's own record instead of the live sweep's (silencing
+leg 1), and the refusal path returns without releasing, so the live
+sweep's claim record would stay clobbered and its own `release_lock_owned`
+would then read `Superseded` and skip its label restore.
+`midbuild_live_claim_probe_runs_before_the_lock_takeover` pins the order
+by byte-comparing `owner.json` across a refused tick.
+
+Fail-open throughout: a missing/corrupt `owner.json`, an unreadable
+journal, a zombie PID, or a recorded PID whose argv does not name this
+issue's sweep (a recycled PID) all resolve to "no evidence", so a stale
+file can never wedge an issue. Every refusal logs at `WARN`; the
+work-finder counts a refusal as an in-flight skip, not an error.
+
+Integration tests spawn the daemon with its machine-level state confined
+to a fixture (`isolate_daemon_state`: `LOOM_WORKSPACES_PATH`,
+`LOOM_SWEEPS_JOURNAL_PATH`, `LOOM_WATCHES_PATH`, `LOOM_WATCH_RESULTS_LOG`,
+plus the cwd), so a test daemon has no managed workspace to reconcile or
+dispatch against.
+
 ## What this page does NOT describe
 
 The legacy schema and tuning advice that historically lived here — the
--- a/loom-daemon/src/claim_reconciliation.rs
+++ b/loom-daemon/src/claim_reconciliation.rs
@@ -498,6 +498,37 @@ pub fn decide(
     }
 }
 
+/// Live-claim veto (Issue #4556) — downgrade a computed `Reclaim` to `Keep`
+/// when the issue still has a **confirmed-live** sweep claim behind it.
+///
+/// [`decide`]'s liveness evidence is a single recorded PID (the journal entry,
+/// or the checkpoint→run-registry join). That PID is the sweep *leader* the
+/// daemon spawned, so a leader that has exited — or a journal entry that was
+/// never written, or was pruned — makes a still-working sweep look dead. On
+/// #4275 that misfire is on the record: at `03:08:15.992Z` this pass reclaimed
+/// `loom:building -> loom:issue` on `DeadPid { pid: 2781227 }` while sweep
+/// processes for the issue were still alive, re-exposing it to the work-finder
+/// and seeding a re-dispatch 37 seconds later.
+///
+/// The veto consults evidence a dead-leader PID cannot invalidate — a live
+/// claim-lock owner, a live machine-level journal record from this repo *or a
+/// nested worktree daemon*, or a live `/loom:sweep <N>` process rooted in the
+/// workspace ([`crate::live_claim::probe`]).
+///
+/// Applies to **every** [`ReclaimReason`], not just the dead-PID ones: a live
+/// sweep means the claim is legitimate no matter which rule proposed dropping
+/// it. Pure and total — the caller supplies the (impure) probe result.
+#[must_use]
+pub fn apply_live_claim_veto(
+    action: ReconcileAction,
+    live_claim: Option<&crate::live_claim::LiveClaimEvidence>,
+) -> ReconcileAction {
+    match (action, live_claim) {
+        (ReconcileAction::Reclaim(_), Some(_)) => ReconcileAction::Keep,
+        (action, _) => action,
+    }
+}
+
 /// Plan reconciliation decisions for every `issue` in `issues`, given an
 /// already-loaded `journal`. Performs no I/O of its own — `run_registry_pid_for`
 /// is the caller's injected checkpoint→run-registry lookup (production passes
@@ -1014,9 +1045,10 @@ pub fn spawn_periodic_reconciliation_task(
 /// best-effort `Command` wrapper.
 pub mod forge {
     use super::{
-        plan, plan_pr, resolve_no_progress_grace_minutes, resolve_stale_hours, BuildingIssue,
-        ClaimedPr, NoProgressEvidence, PrClaimKind, PrReclaimReason, PrReconcileAction,
-        ReclaimReason, ReconcileAction, MAX_ISSUES_PER_WORKSPACE,
+        apply_live_claim_veto, plan, plan_pr, resolve_no_progress_grace_minutes,
+        resolve_stale_hours, BuildingIssue, ClaimedPr, NoProgressEvidence, PrClaimKind,
+        PrReclaimReason, PrReconcileAction, ReclaimReason, ReconcileAction,
+        MAX_ISSUES_PER_WORKSPACE,
     };
     use crate::sweep_journal;
     use anyhow::{anyhow, Context, Result};
@@ -1213,6 +1245,30 @@ pub mod forge {
             let ReconcileAction::Reclaim(reason) = action else {
                 continue;
             };
+            // #4556 live-claim veto: a dead *recorded* pid is not proof the sweep
+            // is gone. Probe for a confirmed-live claim (live lock owner / live
+            // machine-level journal record / live `/loom:sweep <N>` process) and
+            // leave the label alone if one exists — reverting `loom:building` out
+            // from under a working sweep is what re-exposed #4275 to the
+            // work-finder and started its seven-dispatch storm. Probed only on a
+            // Reclaim decision, so a healthy pass pays nothing.
+            //
+            // Judge note (#4605): pass this pass's already-resolved
+            // `journal_path` rather than `None`. `None` would make the probe's
+            // leg 2 re-resolve the process-global default, which silently
+            // diverges from — and could disagree with — the very journal
+            // `plan()` decided against whenever a workspace overrides the path.
+            let live_claim =
+                crate::live_claim::probe(root, Some(journal_path.as_path()), issue_number);
+            if matches!(apply_live_claim_veto(action, live_claim.as_ref()), ReconcileAction::Keep) {
+                log::warn!(
+                    "claim_reconciliation: NOT reclaiming #{issue_number} in {} despite \
+                     {reason:?} — the issue still has {} (#4556 live-claim veto)",
+                    root.display(),
+                    live_claim.map_or_else(|| "a live claim".to_string(), |e| e.to_string()),
+                );
+                continue;
+            }
             match reclaim(gh_bin, root, issue_number) {
                 Ok(()) => {
                     reclaimed += 1;
@@ -1595,6 +1651,57 @@ mod tests {
         let entry = journal_entry("/repo/a", 42, 111);
         let action = decide(&issue(42, None), Some(&entry), None, None, 10.0, &|_| false, 4.0, now);
         assert_eq!(action, ReconcileAction::Reclaim(ReclaimReason::DeadPid { pid: 111 }));
+    }
+
+    // --- #4556 live-claim veto ---
+
+    fn live_lock_evidence() -> crate::live_claim::LiveClaimEvidence {
+        crate::live_claim::LiveClaimEvidence::ClaimLock {
+            pid: 111,
+            sweep_id: "sweep-issue-4275-live".to_string(),
+        }
+    }
+
+    #[test]
+    fn live_claim_veto_downgrades_a_dead_pid_reclaim_to_keep() {
+        // The confirmed #4275 misfire: `DeadPid { pid: 2781227 }` reverted
+        // loom:building -> loom:issue at 03:08:15Z while the sweep was alive.
+        let action = ReconcileAction::Reclaim(ReclaimReason::DeadPid { pid: 2_781_227 });
+        assert_eq!(
+            apply_live_claim_veto(action, Some(&live_lock_evidence())),
+            ReconcileAction::Keep
+        );
+    }
+
+    #[test]
+    fn live_claim_veto_applies_to_every_reclaim_reason() {
+        // A live sweep means the claim is legitimate regardless of which rule
+        // proposed dropping it.
+        for reason in [
+            ReclaimReason::DeadPid { pid: 1 },
+            ReclaimReason::DeadRunRegistry { pid: 1 },
+            ReclaimReason::NoRecordStale { age_hours: 99.0 },
+        ] {
+            assert_eq!(
+                apply_live_claim_veto(
+                    ReconcileAction::Reclaim(reason),
+                    Some(&live_lock_evidence())
+                ),
+                ReconcileAction::Keep,
+                "{reason:?} must be vetoed by a live claim"
+            );
+        }
+    }
+
+    #[test]
+    fn live_claim_veto_is_a_no_op_without_live_evidence() {
+        let action = ReconcileAction::Reclaim(ReclaimReason::DeadPid { pid: 1 });
+        assert_eq!(apply_live_claim_veto(action, None), action, "no evidence => unchanged");
+        assert_eq!(
+            apply_live_claim_veto(ReconcileAction::Keep, Some(&live_lock_evidence())),
+            ReconcileAction::Keep,
+            "Keep is never escalated"
+        );
     }
 
     #[test]
--- a/loom-daemon/src/lib.rs
+++ b/loom-daemon/src/lib.rs
@@ -167,6 +167,7 @@ pub mod idle_exit;
 pub mod init;
 pub mod ipc;
 pub mod issue_creation_mutex;
+pub mod live_claim;
 pub mod main_health_gate;
 pub mod metrics_collector;
 pub mod peer_claims;
--- a/loom-daemon/src/live_claim.rs
+++ b/loom-daemon/src/live_claim.rs
@@ -1,0 +1,756 @@
+//! Confirmed-live sweep-claim probe (Issue #4556).
+//!
+//! ## Problem
+//!
+//! Every pre-#4556 dedup signal for "is issue #N already being worked?" is
+//! either *scoped to one daemon process* or *only proves a file exists*:
+//!
+//! | Signal | Weakness |
+//! |---|---|
+//! | [`crate::types::SweepInfo`] entries (`in_flight()`) | in-memory; wiped by a daemon restart, invisible to a second daemon instance |
+//! | `loom:building` label | reverted by [`crate::claim_reconciliation`] the moment a recorded PID *looks* dead |
+//! | `.loom/locks/issue-<N>/` ([`crate::sweep_registry::SweepRegistry`]'s `acquire_lock`) | existence-only; every reaper / cancel / watchdog path *releases* it on a dead-PID verdict, and the release is what re-opens the door |
+//!
+//! Issue #4275 was dispatched **seven times in 77 minutes** on one host
+//! because those weaknesses compose: the reconciler saw a dead recorded PID and
+//! reverted `loom:building` -> `loom:issue`; the mid-build and review-stall
+//! watchdogs each released the lock and re-dispatched; and three further
+//! dispatches came from a *second* `loom-daemon` instance on the same host
+//! (a debug build run out of a worktree) that shared neither the first
+//! daemon's memory nor its `.loom/locks/`. At the peak four sweep processes for
+//! one issue were alive at once, all writing into one worktree.
+//!
+//! ## What this module adds
+//!
+//! A single **read-only** probe answering the strictly stronger question: *is
+//! there a sweep process for issue #N that is confirmed running right now?*
+//! It never mutates the filesystem, so it can gate destructive work (a lock
+//! release, a label revert, a re-dispatch) without freeing anything itself.
+//!
+//! Three independent evidence legs, cheapest-first, short-circuiting on the
+//! first hit:
+//!
+//! 1. **Live claim lock** — `.loom/locks/issue-<N>/owner.json` whose
+//!    `owner_pid` is a live, non-zombie process. Unlike `acquire_lock`'s
+//!    `AlreadyExists` check this distinguishes a *live* holder from an
+//!    abandoned lock dir, so it is safe to consult *before* a release.
+//! 2. **Machine-level sweep journal** — `~/.loom/sweeps.json`
+//!    ([`crate::sweep_journal`]) entry whose `pid` is live. This file is
+//!    machine-level, so it survives a daemon restart **and** is shared by every
+//!    `loom-daemon` instance on the host. Repo matching is deliberately
+//!    *related-path* rather than exact (see [`repos_are_related`]) so a daemon
+//!    running out of `.loom/worktrees/issue-N` and one running out of the
+//!    parent checkout see each other's claims — they are the same repo, so
+//!    issue #N means the same GitHub issue to both.
+//! 3. **Live sweep process scan** — a process whose cwd is inside the
+//!    workspace root and whose argv contains `/loom:sweep <N>`
+//!    ([`cmdline_targets_sweep_issue`]). This is the last-resort leg that
+//!    catches a sweep no local bookkeeping knows about at all: the exact
+//!    signature of the three unattributed #4275 dispatches, where `ps` showed
+//!    `claude -p /loom:sweep 4275 --claim-owned 4275` processes with no
+//!    corresponding entry in the production daemon's log.
+//!
+//! ## Fail-open by design
+//!
+//! Every leg resolves to "no evidence" on any ambiguity — missing/corrupt
+//! `owner.json`, unreadable journal, no `/proc`, an unreadable `cwd` symlink.
+//! A garbage file must never wedge an issue permanently; the guards built on
+//! this probe only ever refuse on **positively-confirmed** liveness. That is
+//! the same fail-open contract [`crate::sweep_registry`]'s `lock_owned_by_other`
+//! (#4463) already follows.
+//!
+//! Zombies are explicitly **not** live ([`pid_is_live_process`]): a terminated
+//! -but-unreaped child keeps its PID allocated, so a bare `kill(pid, 0)` probe
+//! reports it alive forever — precisely the false-positive that would turn this
+//! guard into a permanent wedge.
+
+use std::path::Path;
+// Only leg 3's `/proc` scan owns a `PathBuf` (the canonicalized workspace root);
+// gating the import keeps a non-Linux build warning-free under `-D warnings`.
+#[cfg(target_os = "linux")]
+use std::path::PathBuf;
+
+use crate::sweep_journal::{self, SweepJournal};
+
+/// Which signal proved a live claim. Carried into the refusal log line / typed
+/// error so an operator can tell a live-lock refusal from a cross-instance
+/// process-scan refusal without re-deriving the evidence by hand.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LiveClaimEvidence {
+    /// `.loom/locks/issue-<N>/owner.json` records a live owner PID.
+    ClaimLock { pid: u32, sweep_id: String },
+    /// The machine-level sweep journal records a live PID for this issue.
+    Journal { pid: u32, repo: String },
+    /// A live process rooted in this workspace is running `/loom:sweep <N>`.
+    SweepProcess { pid: u32 },
+}
+
+impl LiveClaimEvidence {
+    /// The confirmed-live PID behind this evidence.
+    #[must_use]
+    pub fn pid(&self) -> u32 {
+        match self {
+            Self::ClaimLock { pid, .. }
+            | Self::Journal { pid, .. }
+            | Self::SweepProcess { pid } => *pid,
+        }
+    }
+}
+
+impl std::fmt::Display for LiveClaimEvidence {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::ClaimLock { pid, sweep_id } => {
+                write!(f, "a live claim lock owned by sweep {sweep_id} (pid {pid})")
+            }
+            Self::Journal { pid, repo } => {
+                write!(f, "a live sweep-journal record for {repo} (pid {pid})")
+            }
+            Self::SweepProcess { pid } => {
+                write!(f, "a live `/loom:sweep` process (pid {pid})")
+            }
+        }
+    }
+}
+
+/// Minimal read-side view of a claim lock's `owner.json`.
+///
+/// Deliberately its own type rather than a re-export of `sweep_registry`'s
+/// private `LockOwner`: this module only needs two fields, and keeping the
+/// read-side shape local means a future writer-side field addition cannot
+/// break the probe (unknown fields are ignored by serde).
+#[derive(Debug, serde::Deserialize)]
+struct LockOwnerView {
+    owner_pid: u32,
+    #[serde(default)]
+    sweep_id: String,
+}
+
+/// Whether `pid` is a **live, non-zombie** process.
+///
+/// `kill(pid, 0)` alone is insufficient: a terminated-but-unreaped child is a
+/// zombie whose PID is still allocated, so the bare probe reports it alive
+/// indefinitely (the same trap documented on `SweepRegistry::children`). A
+/// zombie holds no locks and writes no files, so treating it as a live claim
+/// would wedge an issue forever — exactly the failure mode this guard must not
+/// introduce.
+#[must_use]
+pub fn pid_is_live_process(pid: u32) -> bool {
+    crate::sweep_registry::is_pid_alive(pid) && !pid_is_zombie(pid)
+}
+
+/// Linux: read `/proc/<pid>/stat` and report whether the process state is `Z`.
+/// Any read/parse failure resolves to `false` (not-a-zombie) so an unreadable
+/// `stat` never *adds* liveness evidence it cannot prove.
+#[cfg(target_os = "linux")]
+fn pid_is_zombie(pid: u32) -> bool {
+    let Ok(stat) = std::fs::read_to_string(format!("/proc/{pid}/stat")) else {
+        return false;
+    };
+    // Format: `pid (comm) state ...`. `comm` may contain spaces AND parens, so
+    // the only safe split point is the LAST ')'.
+    let Some(after_comm) = stat.rfind(')').map(|i| &stat[i + 1..]) else {
+        return false;
+    };
+    after_comm.split_whitespace().next() == Some("Z")
+}
+
+#[cfg(not(target_os = "linux"))]
+fn pid_is_zombie(_pid: u32) -> bool {
+    false
+}
+
+/// Whether `pid` is a live process whose argv targets `/loom:sweep <issue>`.
+///
+/// The **positive-confirmation** primitive behind
+/// [`crate::sweep_registry::LockReleaseOutcome::HolderAlive`]: refusing a lock
+/// release on bare PID liveness would also refuse when an unrelated process has
+/// recycled the PID, wedging the issue permanently. Requiring the argv to name
+/// this exact issue's sweep makes a false refusal essentially impossible.
+///
+/// Linux reads `/proc/<pid>/cmdline`. On other platforms there is no cheap
+/// portable equivalent, so this returns `false` — "cannot confirm" — and every
+/// caller falls open to its pre-#4556 behavior rather than refusing on a guess.
+#[must_use]
+pub fn pid_is_sweep_process_for(pid: u32, issue: u32) -> bool {
+    if !pid_is_live_process(pid) {
+        return false;
+    }
+    read_pid_cmdline(pid).is_some_and(|c| cmdline_targets_sweep_issue(&c, issue))
+}
+
+#[cfg(target_os = "linux")]
+fn read_pid_cmdline(pid: u32) -> Option<String> {
+    let raw = std::fs::read(format!("/proc/{pid}/cmdline")).ok()?;
+    Some(String::from_utf8_lossy(&raw).replace('\0', " "))
+}
+
+#[cfg(not(target_os = "linux"))]
+fn read_pid_cmdline(_pid: u32) -> Option<String> {
+    None
+}
+
+/// The liveness predicate the *evidence legs* use: `pid` is live **and**, when
+/// this platform can read its argv, that argv targets `/loom:sweep <issue>`.
+///
+/// Both bookkeeping legs record the PID of the process the daemon spawned —
+/// `spawn-worker.sh -p "/loom:sweep <N> --claim-owned <N>" …`, which `exec`s
+/// through to `claude` **without changing PID**, so the needle is present for
+/// the whole life of a real sweep (see `SweepRegistry::spawn_child`). Requiring
+/// it therefore costs nothing for a genuine claim while removing the only way
+/// this guard could wedge an issue permanently: a *stale* lock or journal
+/// record whose PID has since been recycled by an unrelated live process. Both
+/// records are removed on reap/cancel and pruned on the next journal write, but
+/// a daemon that dies between a sweep's exit and its reap leaves exactly that
+/// stale record behind, and an unverified PID probe would then refuse every
+/// future dispatch of the issue forever.
+///
+/// Unverifiable argv (non-Linux, or a `/proc` hardened against reading another
+/// process's `cmdline`) falls back to bare liveness — the same signal every
+/// pre-#4556 liveness check already used, so the guard is never *weaker* than
+/// the code it supplements on those hosts.
+#[must_use]
+pub fn pid_is_live_claim_for(pid: u32, issue: u32) -> bool {
+    if !pid_is_live_process(pid) {
+        return false;
+    }
+    match read_pid_cmdline(pid) {
+        Some(cmdline) => cmdline_targets_sweep_issue(&cmdline, issue),
+        // Cannot read the argv on this platform: fall back to bare liveness.
+        None => true,
+    }
+}
+
+/// Leg 1: the per-issue claim lock's owner, when its PID is confirmed live.
+///
+/// `is_live` is injected so tests can drive both verdicts without spawning
+/// processes. Fail-open: a missing lock dir, unreadable or unparseable
+/// `owner.json`, or a dead/zombie owner all resolve to `None`.
+#[must_use]
+pub fn live_lock_owner_in(
+    locks_dir: &Path,
+    issue: u32,
+    is_live: &dyn Fn(u32) -> bool,
+) -> Option<LiveClaimEvidence> {
+    let owner_path = locks_dir.join(format!("issue-{issue}")).join("owner.json");
+    let raw = std::fs::read_to_string(&owner_path).ok()?;
+    let owner: LockOwnerView = serde_json::from_str(&raw).ok()?;
+    if !is_live(owner.owner_pid) {
+        return None;
+    }
+    Some(LiveClaimEvidence::ClaimLock {
+        pid: owner.owner_pid,
+        sweep_id: owner.sweep_id,
+    })
+}
+
+/// Whether two workspace-root strings name the **same repository** for the
+/// purpose of issue-number identity.
+///
+/// Exact equality is not sufficient. The three unattributed #4275 dispatches
+/// came from a `loom-daemon` whose `workspace_root` was
+/// `<repo>/.loom/worktrees/issue-4385` — a git worktree of the very same
+/// checkout. Two roots in an ancestor/descendant relationship therefore refer
+/// to one GitHub repo, and "issue #N" means the same issue to both, so each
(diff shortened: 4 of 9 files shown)


rjwalters:
Code evaluation queued; CI (Rust Unit Tests) is still running. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation looks promising; CI (Rust Unit Tests) is still running. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation looks promising (detailed PR description, comprehensive test additions); CI is still running (Rust Unit Tests in progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation queued; CI is still running (Rust Unit Tests pending). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
Code checkout/diff reviewed; CI still has a pending check (Rust Unit Tests). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Reviewed `loom-daemon/src/live_claim.rs` (new, 748 lines) plus its integration into `sweep_registry.rs`, `claim_reconciliation.rs`, and `work_finder.rs`.

**Design:** The core insight is right — every pre-existing dedup signal (`in_flight()`, `loom:building` label, lock-file existence) proves something weaker than "is a sweep process for this issue alive right now." The new `live_claim::probe` answers that stronger question with three fail-open, short-circuiting evidence legs (live lock owner → live journal record → `/proc` sweep-process scan), each independently surviving a lock release, a label revert, a daemon restart, and a second daemon instance — which is exactly how #4275's seven-dispatch storm compounded.

**Correctness checks performed:**
- `cmdline_targets_sweep_issue`'s scan loop strictly shrinks `rest` each iteration (`idx + NEEDLE.len() >= 1`), so it terminates on the repeated-needle case the regression test targets.
- Zombie-PID handling (`pid_is_zombie`, Linux `/proc/<pid>/stat` parse via last `)`) correctly avoids the classic "unreaped child looks alive forever" trap.
- `repos_are_related` is component-wise (`Path::starts_with`), not string-prefix — confirmed `/repo/loom` vs `/repo/loom-2` resolve as unrelated per the test.
- The four gating sites (`dispatch` step 2.9, `midbuild_watchdog_once`, `release_lock_owned` → `HolderAlive`, `claim_reconciliation`'s `apply_live_claim_veto`) are all consistently wired through the shared `retained()` predicate / veto function rather than duplicated ad-hoc checks.
- Fail-open behavior is correctly total: missing/corrupt `owner.json`, unreadable journal, no `/proc`, and — critically — a live PID whose argv doesn't name this issue's sweep (recycled PID) all resolve to "no evidence," so the guard can't wedge an issue permanently.

**Test execution:**
- Ran `cargo test -p loom-daemon live_claim --lib` locally: 33/33 relevant tests pass (live_claim module, claim_reconciliation veto, sweep_registry integration, work_finder attribution).
- CI: all checks green (Rust Unit Tests 2523 passed, clippy, fmt, build, installer tests).
- Merge state CLEAN.

**PR hygiene:** `Closes #4556` present and correct (this is the tracked issue itself, not a partial increment). The reviewer notes about `integration_singleton_guard` not being run locally (host-disruptive `cleanup_all_loom_sessions()`) are reasonable — CI covers it.

Nice work isolating the stray-debug-daemon lead too (`isolate_daemon_state()` confining test daemons away from the operator's real `~/.loom/workspaces.json` was a real latent hazard, not just this PR's main deliverable).

rjwalters:

- **Size check**: 1,623 lines changed (+1602/-21), exceeding the 200-line auto-merge limit. No `loom:auto-merge-ok` label is present to waive this.
- **Merge conflict**: `mergeable` state is `CONFLICTING` with `main`.

**Next steps:**
- Rebase onto latest `main` to resolve the conflict (likely from #4602/#4604 touching overlapping daemon files)
- Given the size (core dispatch/lock safety logic in `loom-daemon`), this warrants human review and manual merge rather than a `loom:auto-merge-ok` waiver

Keeping `loom:pr` label. Champion will retry on the next tick once the blocking condition clears, but the size criterion will likely require a human decision either way.

---
*Automated by Champion role*

rjwalters:

- **Merge Conflict Check**: FAIL — mergeable state is `CONFLICTING` (conflicts with `main`)
- **Size Check**: FAIL — 1623 lines changed (+1602/-21), exceeds the 200-line auto-merge limit and no `loom:auto-merge-ok` label is present

**Next steps:**
- Rebase onto latest `main` and resolve the conflicts
- If this PR is safe despite its size, a human or Judge can add `loom:auto-merge-ok` to waive the size limit

---
*Automated by Champion role*

rjwalters:
## Judge re-evaluation — ⛔ changes requested (merge conflict), code substance approved

Re-reviewing after the prior approval at 08:28:17Z, because the merge state has since changed. **The code review passes** — my objection is entirely the merge state and one semantic interaction it creates.

### Blocker: PR is `DIRTY` against current `main`

```
gh api repos/rjwalters/loom/pulls/4605 --jq '{mergeable,mergeable_state}'
{"mergeable":false,"mergeable_state":"dirty"}

git merge-tree --write-tree --name-only origin/main 16b8e0c3
CONFLICT (content): Merge conflict in loom-daemon/src/sweep_registry.rs
```

`main` moved to `dd638554` after this branch was cut; the conflict is with **#4602** (`39c5cdcf`, "hold the issue lock across the midbuild-watchdog worktree clean"), which rewrote the *exact* block this PR modifies. Per the Judge rebase policy this is a **complex** conflict (overlapping changes in the same function, with real behavioural coupling), not a whitespace/rebase-and-go — so it routes to Doctor rather than an automated Judge rebase. `loom:pr` is being cleared so Champion does not keep re-polling a branch that cannot merge (it already refused at 08:29:08Z for exactly this reason).

Note: a rebase already appears to be in flight in `.loom/worktrees/issue-4556` (a `rebase-merge` state with `UU loom-daemon/src/sweep_registry.rs`). Whoever owns that should finish it — I deliberately did not touch that worktree.

### ⚠️ The conflict resolution is NOT mechanical — read this before resolving

#4602 replaced the midbuild watchdog's read-only `lock_owned_by_other` probe with **`claim_lock_for_midbuild`**, which *takes over* `.loom/locks/issue-<N>/owner.json` in place, rewriting `owner_pid` to the **daemon's own PID** and `sweep_id` to `midbuild-watchdog-<dead>`. That interacts with this PR in two ways that a naive "keep both hunks" resolution can get wrong:

1. **The #4556 live-claim probe MUST run strictly before `claim_lock_for_midbuild`.** Once the takeover write lands, `owner.json` no longer names the live sweep's PID, so `live_claim`'s leg 1 (`live_lock_owner_in`) goes silent — `pid_is_live_claim_for(daemon_pid, issue)` correctly fails the argv check, since the daemon's argv is `loom-daemon`, not `/loom:sweep <N>`. Placing the probe after the claim would silently degrade the guard to legs 2+3 *and* would already have clobbered the live sweep's lock-owner record.
2. **`claim_lock_for_midbuild` takes over the lock whenever `lock_owned_by_other` is false** — i.e. when the lock still names the sweep this daemon believes is dead. That is precisely the false-dead case #4556 exists to catch, so the live-claim refusal is now load-bearing as the *only* thing standing between a live sweep and both a `git reset --hard` and a lock-owner clobber. Please keep an assertion for that ordering in `midbuild_watchdog_does_not_recover_an_issue_with_a_live_claim` (e.g. assert the lock's `owner.json` still names the live sweep after the refused tick).

Also re-verify `release_lock_owned`'s new `HolderAlive` arm against #4602's post-clean `release_lock_owned(issue, watchdog_sweep_id)` call: after a takeover the owner PID is the daemon, so `pid_is_sweep_process_for` returns false and it releases normally — correct, but it should be re-confirmed on the rebased tree rather than assumed.

### What I verified (all green, on `16b8e0c3` as pushed)

- **CI**: all 8 checks `success`/`skipped` on the head SHA (Rust Unit Tests included — it finished after the prior review).
- **Local suite**: `cargo test -p loom-daemon --lib` → **2523 passed, 0 failed** in 153s. Exactly matches the PR body's claim; I did not take the self-report on trust.
- **`cmdline_targets_sweep_issue`**: loop advance is `rest = &rest[idx + NEEDLE.len()..]`, so `rest` strictly shrinks — terminates; `cmdline_matcher_terminates_on_repeated_needles` pins it. Exact-token parse rules out `42751`/`14275`; `starts_with(char::is_whitespace)` rules out `/loom:sweeper`; PR-set mode correctly excluded.
- **Argv verification is real, not theoretical**: `spawn_child` builds `format!("/loom:sweep {issue} --claim-owned {issue}")`, and `spawn-worker.sh` → `spawn-claude.sh` → `claude` is an unbroken `exec` chain (`exec "$RUNNER" "$@"` / `exec claude "${PASSTHROUGH_ARGS[@]}"`), so the PID is stable and the needle survives. `record_child_pid_in_lock` (#3808) rewrites `owner_pid` to the child, so leg 1 sees a sweep PID, not the daemon's, in steady state.
- **Recycled-PID non-wedge**: the argv requirement is the right fix and is genuinely tested (`probe_ignores_a_live_pid_whose_argv_is_not_this_issues_sweep`, `release_lock_owned_releases_when_a_live_pid_is_not_this_issues_sweep`, both using the test binary's own live-but-not-a-sweep PID).
- **Zombie exclusion** is load-bearing and correct — it is what makes the cancel→dispatch sequence in `watchdog_once`/`review_stall_watchdog_once` safe even when `reap_handle` has not reaped yet.
- **Test isolation**: `fixture_registry` already pins `config.journal_path` into the tempdir, and leg 3 is cwd-scoped to the tempdir workspace, so the new unit tests cannot read the operator's real `~/.loom/sweeps.json` or be perturbed by live sweeps on this host. `isolate_daemon_state`'s four env names all match the real constants (`LOOM_WORKSPACES_PATH`, `LOOM_SWEEPS_JOURNAL_PATH`, `LOOM_WATCHES_PATH`, `LOOM_WATCH_RESULTS_LOG`).
- **Tests are not tautological**: they spawn real processes (`FakeSweep`) and drive the real `dispatch` / `midbuild_watchdog_once` entry points; `midbuild_watchdog_does_not_recover_an_issue_with_a_live_claim` asserts a *precondition* that the #4449 veto is not what refuses, and `midbuild_watchdog_recovers_once_the_live_claim_is_gone` pins the inverse so the guard cannot wedge recovery. `wait_until_argv_visible` is a legitimate fix for a real fork-vs-exec race, not a sleep-to-green.
- **Rejected acceptance criterion 5** (per-issue worktree lockfile): defensible. The criterion says "consider", the rationale is stated in the PR body, and the combination of leg 3 (cwd-scoped liveness) plus #4449's existing `worktree_in_use` veto (`.loom-in-use` / `index.lock` / live-cwd) covers the same ground without a fourth stale-able on-disk file. Not scope-dodging.
- **Docs**: `defaults/docs/daemon-reference.md` only, which matches this repo's convention (the installed `.loom/docs/` copy is refreshed by the periodic resync commit).

### Non-blocking notes (address if convenient during the rebase, or file as follow-ups)

1. **A live-claim refusal burns the watchdogs' single retry.** `watchdog_once` and `review_stall_watchdog_once` both `insert` into `*_retried` *before* `cancel` + `dispatch`, so a `LiveClaimDispatchError` from step 2.9 consumes the one allowed recovery attempt. Risk is low (both cancel and reap first, and zombies are excluded), but it is asymmetric with the midbuild path, which deliberately does *not* consume its latch on a live-claim refusal. Consider rolling the latch back when the error downcasts to `LiveClaimDispatchError`.
2. **`claim_reconciliation::forge` calls `probe(root, None, issue)`**, so leg 2 resolves the process-global default journal path rather than any `SweepRegistryConfig::journal_path` override. Correct for production today, but it silently loses leg 2 for any workspace that ever sets a non-default journal path.
3. **`repos_are_related` treats an ancestor path as related**, so a journal record whose `repo` is a *parent directory* of the workspace would over-match. The failure direction is conservative (refuse a dispatch), so this is a note, not a defect.
4. **Non-Linux hosts** fall back to bare PID liveness in `pid_is_live_claim_for`, which re-opens the recycled-PID wedge the Linux path carefully closes. Documented and acceptable, but macOS is a real Loom dev platform — `KERN_PROCARGS2` would close it in a follow-up.
5. `integration_singleton_guard` was not run locally (its `cleanup_all_loom_sessions()` would nuke this host's live Loom tmux sessions). Reasonable call; the change there is a single isolation call.

### To clear this

Rebase onto `origin/main` (`dd638554`), resolve `loom-daemon/src/sweep_registry.rs` with the ordering constraint in the ⚠️ section above, re-run `cargo test -p loom-daemon --lib`, and flip back to `loom:review-requested`. The substance needs no rework — only the merge.

rjwalters:

- **Size check**: 1623 lines changed (+1602/-21), exceeds the 200-line auto-merge limit and carries no `loom:auto-merge-ok` override.
- **Merge conflict check**: `mergeable` state is `CONFLICTING` — needs a rebase against current `main`.

CI itself is fully green (Rust Lint/Build/Unit Tests, Installer Integration Tests, CLAUDE.md/AGENTS.md checks all passing).

**Next steps:**
- Rebase onto latest `main` and resolve the conflict
- If the size is justified (this is a substantial daemon-side feature — live sweep claims, reconciliation, tests), a human or Judge can add `loom:auto-merge-ok` to waive the size ceiling once conflicts are resolved

Keeping `loom:pr` label. Champion will retry on the next tick once the blocking conditions clear.

---
*Automated by Champion role*

rjwalters:
Evaluation paused — a live `/loom:sweep 4556` process (pid 3775512/3775971, running since 07:46:52Z, still holding `.loom/locks/issue-4556/owner.json`) is actively working in this PR's builder worktree (`.loom/worktrees/issue-4556`) concurrently with this review.

What happened:
1. `mergeStateStatus` was `DIRTY`; I attempted the documented automated rebase in the existing builder worktree.
2. Rebase conflicted in `loom-daemon/src/sweep_registry.rs` — main had picked up #4602 (`claim_lock_for_midbuild`, issue #4564) in the meantime, and this PR independently adds `live_claim_evidence`/the `#4556` guard nearby. I resolved it as a "both sides add independent code" merge (keep both methods, keep both call sites in sequence).
3. That mechanical resolution turned out to be **semantically wrong**: it calls `claim_lock_for_midbuild` (which takes exclusive ownership of the on-disk claim lock) *before* the new live-claim probe, and on the probe's refusal path it `continue`s without ever releasing the lock it just claimed (`release_lock_owned` is only called later, on the non-refused path). A local test run confirmed this wedges recovery: `midbuild_watchdog_recovers_once_the_live_claim_is_gone` failed under my resolution (0 recoveries instead of 1) because the second call sees the watchdog's own leftover claim as "owned by other" and refuses forever.
4. While investigating, I found the correct fix already sitting **unstaged** in the working tree (probe `live_claim_evidence` *before* calling `claim_lock_for_midbuild`, with a comment noting "ORDERING IS LOAD-BEARING") — meaning the live sweep process above is mid-way through fixing this exact interaction itself, in the same worktree, right now.

Given a live sweep is actively editing this worktree, I'm stopping here rather than risk clobbering its in-progress fix with any further rebase/reset/checkout. I have **not pushed** anything — `origin/feature/issue-4556` is untouched at `16b8e0c3`. My local rebase commit only exists in the shared worktree's local branch ref and will naturally be superseded once the live sweep finishes and pushes its own resolution.

Releasing the review claim without an end-state label so this re-enters the queue once the live sweep completes and CI re-settles on its actual push. Please re-evaluate once `loom:review-requested` shows a fresh commit from the live sweep.

rjwalters:
## Doctor: rebase resolution was WRONG — fixed locally, push blocked pending approval

**Status: `loom:changes-requested` / `loom:merge-conflict` deliberately LEFT IN PLACE.** The remote head (`16b8e0c3`) still contains the defect below. The corrected branch exists only locally in `.loom/worktrees/issue-4556` — the `git push --force-with-lease` needs an operator approval this session did not have. **Do not re-judge until the force-push lands.**

### What I found

The pre-existing rebase onto `39c5cdcf` (#4602) merged cleanly with no conflict markers, and every test passed — but it **inverted the load-bearing ordering** the Judge called out. In `midbuild_watchdog_once`’s `Recover` arm it produced:

```
claim_lock_for_midbuild(issue, &sweep_id)   // #4602 takeover  ← FIRST
live_claim_evidence(issue)                  // #4556 refusal   ← SECOND
```

That is exactly backwards, and it merged silently because #4602 *replaced* the read-only `lock_owned_by_other` probe that #4556 had been sequenced after with a mutating takeover, leaving #4556’s block textually undisturbed in a now-wrong slot.

### Why it matters (both harms, confirmed by test)

`claim_lock_for_midbuild` rewrites `.loom/locks/issue-<N>/owner.json` in place with `owner_pid = std::process::id()` and `sweep_id = midbuild-watchdog-…`, and it does so whenever `lock_owned_by_other` is false — i.e. on the false-dead-PID case #4556 exists to catch.

1. **Leg 1 goes silent.** `live_claim::probe` leg 1 reads that same `owner.json` and runs `pid_is_live_claim_for` on the recorded PID. Post-takeover the PID is the daemon, whose argv is `loom-daemon`, not `/loom:sweep <N>` — so the check correctly declines and the strongest evidence leg is destroyed by the very code path it was meant to gate.
2. **The refusal path clobbers and leaks.** Even when legs 2/3 (journal, `/proc`) still fire and the watchdog refuses, it `continue`s *without releasing* — so the live sweep’s owner record is gone, replaced by a claim owned by the live daemon PID. The live sweep’s own `release_lock_owned` then reads `Superseded`, `retained()` is true, and it skips its label restore; every later tick sees a "live" owner via `worktree_in_use`.

### Fix

`3abd5098` moves the probe ahead of the takeover and adds `midbuild_live_claim_probe_runs_before_the_lock_takeover`, whose fixture makes the takeover *eligible* (lock names the dead sweep; dead owner PID ⇒ no #4449 live-use evidence and no #4463 peer-owner refusal) while the live claim is visible only through the journal leg, then asserts `owner.json` is byte-identical after the refusal. Verified to fail on the inverted order:

```
left:  "sweep_id": "midbuild-watchdog-sweep-issue-4602-dead", "owner_pid": <daemon pid>
right: "sweep_id": "sweep-issue-4602-dead",                   "owner_pid": 2147483640
```

The `Recover` arm now reads: token gate → **#4556 probe** → **#4602 `claim_lock_for_midbuild`** → `midbuild_retried` latch → `clean_worktree` **under the held lock** → `release_lock_owned` handoff → `dispatch`. Both invariants hold: the clean still runs while the watchdog owns the issue lock (#4602/#4564), and the live-claim refusal is still the only thing between a live sweep and both the `git reset --hard` and the owner-record clobber (#4556).

### Also done

- **Rebased onto current `main` (`8eb64fc3`)**, not just `39c5cdcf` — `#4604`/`#4606`/`#4609` had landed meanwhile. Clean rebase; the only overlap was `work_finder.rs`, where the `LiveClaimDispatchError` → `skipped_in_flight` arms sit alongside `#4604`’s admission refactor without interaction.
- **Judge note 2** (`claim_reconciliation::forge` passing `probe(root, None, issue)`): addressed in `9e68f90d` by threading the pass’s already-resolved `journal_path` through. Behaviourally a no-op today — both sides resolve via `sweep_journal::default_journal_path()`, which honours `LOOM_JOURNAL_PATH`, in the same process — but it drops a redundant resolve+load per reclaimed issue and pins leg 2 to the journal `plan()` actually decided against.
- **Judge notes 1, 3, 4** left as-is. Note 1 (the `*_retried` latch consumed by a `LiveClaimDispatchError` in `watchdog_once` / `review_stall_watchdog_once`) is a real asymmetry but touching the latch ordering changes error-path loop bounds and wants its own tests — follow-up, not a rebase fix. Notes 3 and 4 are conservative-direction observations.

### Test results

| Check | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo check --workspace --all-targets` | clean |
| `cargo nextest run -p loom-daemon --lib` | **2551 passed, 0 failed** |
| `cargo nextest run --workspace --exclude loom-daemon` | 4 passed |
| `integration_singleton_guard` | skipped by design (kills all `loom-*` tmux sessions on this host) |

One caveat worth flagging: the first full run reported `main_health_gate::tests::test_run_gate_tick_runs_again_after_main_advances` timing out at 360s. That is **not** this PR — `#4604` put a machine-wide build slot (`~/.loom/locks/build-slot`, 1 slot, 300s bounded wait) inside `run_gate_tick`, so the unit test blocks behind this host’s live daemon. It passes in 0.5s with `LOOM_BUILD_SLOTS=0`, and the full suite is 2551/2551 with `LOOM_BUILD_SLOT_DIR` pointed at a tempdir. Filed as #4615. The same contention is why `buildGate.command` (plain `cargo test --workspace`) did not finish inside its 1500s budget here; the checks above cover its substance.

rjwalters:
## Doctor: rebased onto current `main`, ordering blocker fixed — ready for re-review

Head is now `b07a5c2b`, `mergeable: true`. This supersedes the 09:43Z doctor comment: that session's fix existed only on another host's local worktree and was never pushed (`3abd5098` / `9e68f90d` do not exist in this repo's object store), so the work was redone here from `16b8e0c3`.

### Blocker 1 — merge conflict (`DIRTY`)

Rebased twice, because `main` moved mid-run: first onto `86e82a2b`, then onto `8f9cadfa` after `#4647`/`#4652` landed. Four conflicted files, all resolved by hand:

| File | Conflict | Resolution |
|---|---|---|
| `loom-daemon/src/sweep_registry.rs` | #4602's `claim_lock_for_midbuild` rewrote the block this PR modifies | **semantic** — see below |
| `loom-daemon/src/sweep_registry.rs` | `OpenPrProbe` (#4452) vs `LiveClaimDispatchError` — independent additions | union |
| `loom-daemon/src/claim_reconciliation.rs` | `use super::{…}` list | union |
| `loom-daemon/tests/common/mod.rs` | #4573 fail-closed env chain vs `isolate_daemon_state` | both — the #4573 vars pin the workspace registry + worktree root, `isolate_daemon_state` pins the sweep journal, watch registry, watch-results log and cwd |
| `loom-daemon/tests/integration_singleton_guard.rs` | main refactored to `daemon_command(socket, workspace)`; #4647 then renamed `cleanup_all_loom_sessions` → `cleanup_test_sessions` | dropped this PR's now-deleted `spawn_daemon_and_wait`, moved `isolate_daemon_state` into `daemon_command`, took main's renamed import |

### Blocker 2 — the ⚠️ ordering constraint

Confirmed the hazard exactly as described, and the naive "keep both hunks" resolution really does invert it. The `Recover` arm now reads:

```
token gate → #4556 live_claim_evidence → #4602 claim_lock_for_midbuild
           → midbuild_retried latch → clean_worktree (under the held lock)
           → release_lock_owned handoff → dispatch
```

with an `ORDERING IS LOAD-BEARING` comment at the probe naming both harms (leg 1 silenced by the takeover; the refusal path `continue`s without releasing, leaving the live sweep's owner record clobbered).

**New regression test, as requested:** `midbuild_live_claim_probe_runs_before_the_lock_takeover`. The fixture makes the takeover genuinely *eligible* — a leftover lock naming the dead sweep with a dead owner pid, so neither the #4449 live-use veto (asserted empty) nor the #4463 peer-owner probe (asserted false) is what refuses — and leaves the live claim visible **only** through the journal leg. It then byte-compares `owner.json` across the refused tick. Verified to fail under the inverted order:

```
left:  "owner_pid": 38681,      "sweep_id": "midbuild-watchdog-sweep-issue-4602-dead"
right: "owner_pid": 2147483640, "sweep_id": "sweep-issue-4602-dead"
```

Worth flagging: under the inverted order **9** `midbuild_*` tests fail on macOS, not just the new one. Because `pid_is_live_claim_for` falls back to bare PID liveness where argv is unreadable, a post-takeover probe reads the daemon's own live PID as a live claim and refuses *every* recovery. The ordering is therefore load-bearing harder on non-Linux than on Linux.

### Judge note 3 (re-verify `release_lock_owned`'s `HolderAlive` arm) — verified

Confirmed on the rebased tree rather than assumed, and it holds on **both** platforms, for different reasons. `release_lock_owned` → `live_sweep_lock_owner_pid` → `pid_is_sweep_process_for`, which is the *argv-requiring* primitive (`read_pid_cmdline(pid).is_some_and(…)`) — not the fail-open `pid_is_live_claim_for` the evidence legs use. On Linux the daemon's argv is `loom-daemon`, so the match fails; on non-Linux `read_pid_cmdline` returns `None` and `is_some_and` short-circuits to `false`. Either way the post-clean `release_lock_owned(issue, watchdog_sweep_id)` releases normally. `midbuild_releases_its_own_claim_before_re_dispatching` passes.

### Judge note 2 (`probe(root, None, issue)`) — addressed

`claim_reconciliation::forge` now passes the pass's already-resolved `journal_path`, so leg 2 reads the same journal `plan()` decided against instead of re-resolving the process-global default.

Notes 1, 4, 5 left as-is, matching the prior session's reasoning: note 1 (the `*_retried` latch consumed by a `LiveClaimDispatchError` in `watchdog_once` / `review_stall_watchdog_once`) changes error-path loop bounds and wants its own tests — follow-up, not a rebase fix. Notes 4 (`KERN_PROCARGS2`) and 5 are conservative-direction.

### One drive-by, needed to keep the local gate green

`PathBuf` and `path_within` in `live_claim.rs` are used only by leg 3's `/proc` scan, so a non-Linux build emitted `unused_imports` + `dead_code`. Invisible on CI (Linux) but fatal under `clippy -D warnings` on a macOS dev host. Both are now `#[cfg(target_os = "linux")]`-gated to match their single caller. No behaviour change on Linux.

### Test results (macOS, on `b07a5c2b`)

| Check | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | clean |
| `cargo test -p loom-daemon --lib` | **2603 passed, 0 failed** (139s) |
| `cargo test --workspace --exclude loom-daemon` | 4 passed |
| `cargo test -p loom-daemon --test integration_drain_then_exit` | 1 passed |
| `integration_singleton_guard` | not run locally (its `setup()` kills tmux sessions on this host) — compile-checked via `clippy --all-targets`; CI covers it |

Ran with `LOOM_BUILD_SLOT_DIR` pointed at a tempdir, per the #4615 caveat (`#4604`'s machine-wide build slot otherwise blocks `main_health_gate` unit tests behind this host's live daemon).

An earlier run on the intermediate rebase also showed `config_resolver::tests::test_resolve_malformed_legacy_config_soft_fails_never_aborts` and `tokens_pool::bootstrap::tests::bootstrap_no_source_errors` failing. Both are host-environment leakage, not this PR (it touches neither file): they read this host's live `LOOM_CONFIG_DEFAULTS_FILE` private-defaults file and its real `~/.claude-monitor/accounts.env`, and both pass with those two env vars pointed at an empty dir. Neither reproduced on the final tree.

Flipping to `loom:review-requested`.

rjwalters:
Code evaluation in progress; CI is still running (Rust Unit Tests and Shell Test Suites pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Re-review after the Doctor's rebase push (`b07a5c2b`). The substance of this PR was already approved in the 08:30:26Z review; this pass verified that the two blockers from that review are actually resolved on the pushed head — none of it taken on self-report.

## What I verified on `b07a5c2b`

**Blocker 1 (merge conflict)**: `mergeable: true, mergeable_state: clean` against current `main`. All four conflict resolutions listed in the Doctor's comment are present in the tree.

**Blocker 2 (load-bearing ordering)**: confirmed in `midbuild_watchdog_once`'s `Recover` arm (`sweep_registry.rs:7113`): the `#4556` `live_claim_evidence` probe runs strictly before `#4602`'s `claim_lock_for_midbuild`, with the `ORDERING IS LOAD-BEARING` comment naming both harms (leg-1 silencing and the refusal-path owner-record clobber). The clean still runs under the held lock and the handoff `release_lock_owned(issue, &watchdog_lock_id)` precedes `dispatch`, so the `#4602`/`#4564` invariant is preserved too. The requested regression test `midbuild_live_claim_probe_runs_before_the_lock_takeover` is real and well-built: its fixture asserts as *preconditions* that neither the #4449 live-use veto nor the #4463 peer-owner probe is what refuses (so an ordering inversion is genuinely observable), leaves the live claim visible only through the journal leg, and byte-compares `owner.json` across the refused tick.

**Judge note 2**: `claim_reconciliation.rs:1262` now threads the pass's resolved `journal_path` into `probe(root, Some(journal_path.as_path()), issue)` — verified.

**Judge note 3** (`HolderAlive` arm on the rebased tree): the release path goes through `pid_is_sweep_process_for` (argv-requiring, not the fail-open primitive), so the post-takeover release behaves correctly on both platforms — matches the Doctor's analysis.

**Drive-by cfg-gating**: `PathBuf`/`path_within` in `live_claim.rs` are `#[cfg(target_os = "linux")]`-gated to match their single caller; clippy is clean.

## Test Execution

1. `cargo test -p loom-daemon --lib -- live_claim midbuild` (local, macOS) — ✅ Executed: **52 passed, 0 failed** (2551 filtered), including `midbuild_live_claim_probe_runs_before_the_lock_takeover`, `midbuild_watchdog_does_not_recover_an_issue_with_a_live_claim`, and the inverse non-wedge test `midbuild_watchdog_recovers_once_the_live_claim_is_gone`
2. CI on head SHA — ✅ All checks pass, including Rust Unit Tests (4m37s, full suite) and Shell Test Suites (hermetic)
3. `integration_singleton_guard` — ⏭️ Skipped locally by design (kills live tmux sessions on this host); covered by CI
4. Merge state — ✅ `clean` against current `main`

## Label note

Adding `loom:auto-merge-ok` alongside `loom:pr`, per the Champion's 08:30:27Z comment inviting a Judge to waive the 200-line ceiling once conflicts were resolved: the size (+1729/−21) is justified for this daemon-side guard, and the PR has now had two full independent judge reviews plus a verified regression test for the one semantic hazard the rebase introduced.

Follow-ups already tracked elsewhere: note 1 (watchdog `*_retried` latch consumed on a `LiveClaimDispatchError`) and note 4 (`KERN_PROCARGS2` for macOS argv) remain open as conservative-direction follow-ups per the two prior reviews; #4615 tracks the build-slot test contention.

rjwalters:
**Champion Auto-Merge**

- Judge approved (`loom:pr` label), plus `loom:auto-merge-ok` applied by Judge with explicit rationale (two independent reviews, a verified regression test for the semantic ordering hazard introduced by the rebase, full local + CI test execution)
- Merge-risk judgment: `loom:auto-merge-ok` override applies — Judge's re-review on `b07a5c2b` verified both blockers (merge conflict resolution, load-bearing probe ordering) directly in the diff and by running the new regression test, not by self-report
- Diff size: 1750 lines (+1729/-21) — informational, not a gate
- No critical files modified (daemon source/tests + docs only; no Cargo.toml, workflows, or migrations)
- No merge conflicts (mergeable: clean)
- All CI checks passing (MCP Server Build skipping, all others pass)

`git revert <commit-sha>`

---
*Automated by Champion role*

PR #4639, opened 2026-07-30
fix(daemon): exit the process on startup failure instead of unwinding through Runtime::drop

### LINKED ISSUE #4531: bug(daemon): singleton-guard refusal prints the error but the process never exits (hangs indefinitely)
## Summary

When a second `loom-daemon` is started against a socket a live daemon already owns, the singleton guard (#3806) correctly detects the incumbent and prints the refusal message — but **the process does not exit**. It hangs indefinitely instead of exiting non-zero.

Found incidentally while validating the nextest switch in #4385 (this is **not** caused by that change — it reproduces identically under plain `cargo test`).

## Reproduction (100% on the affected host)

```bash
d=$(mktemp -d); s=$d/daemon.sock
LOOM_SOCKET_PATH=$s LOOM_NO_RESTORE=1 ./target/debug/loom-daemon >$d/a.out 2>$d/a.err &
until [ -e "$s" ]; do sleep 0.1; done

LOOM_SOCKET_PATH=$s LOOM_NO_RESTORE=1 timeout 15 ./target/debug/loom-daemon; echo "exit=$?"
# exit=124   (timeout killed it — it never exited on its own)
# stderr:    Error: another loom-daemon is already listening on … — refusing to start. …
```

Still hangs with the safehouse env layer unset (`env -u LOOM_SAFEHOUSE_ENABLED -u LOOM_SAFEHOUSE_SOCKET -u LOOM_SAFEHOUSE_ROOM -u SAFEHOUSED_SOCKET`), so the narration sink is not the (only) thing keeping it alive.

## Impact

Both `loom-daemon/tests/integration_singleton_guard.rs` tests fail on an affected host:

- `test_second_daemon_refuses_and_first_survives` — `second daemon should exit promptly after refusing to start` (waits 10s)
- `test_stale_socket_is_reclaimed` — collateral: the fixed 2s startup budget plus ~750ms of connect retries is under the ~2.3s the daemon actually needs to bind on a loaded host

They pass on GitHub Actions runners (whose `$HOME` has no populated `~/.loom`), so CI stays green and the wart is invisible there. Operationally, a mistaken second `loom-daemon` start leaves a zombie process holding resources rather than failing fast.

## Likely cause

`main.rs`'s refusal path returns/propagates an `Err` instead of terminating, and something spawned before the guard check (a non-daemon thread — tmux health monitor, activity DB, metrics collector — or a tokio runtime that never becomes idle) keeps the process alive after `main` unwinds. On a host with an initialized `~/.loom`, those subsystems actually start; on a pristine CI runner they no-op, which is why the exit happens to work there.

## Suggested fix

Make the refusal path terminate deterministically (`std::process::exit(<non-zero>)` after printing, or shut down the spawned subsystems before returning), and make the integration test's startup wait poll for the socket with a generous budget instead of a fixed 2s sleep so it is not load-sensitive.

---

### Investigation note (verified against `origin/main`)

The singleton guard itself lives in `loom-daemon/src/ipc.rs` (`IpcServer::run`, #3806): it calls `anyhow::bail!(...)` when `socket_has_live_listener` sees a live peer, and that `Err` is propagated with `server.run().await?` in `main.rs` (the call site is the last statement of the async `main`), so the async `main` fn does return `Err` promptly — `#[tokio::main]`'s generated sync wrapper then prints `Error: {:?}` and hands the process an `ExitCode::FAILURE`.

Two concrete background-task candidates spawned *before* the guard check are confirmed to exist on `origin/main`, matching the issue's own "Likely cause" list:
- `health_monitor::start_tmux_health_monitor` (`loom-daemon/src/health_monitor.rs:57-67`) — `thread::spawn` with an unconditional `loop { thread::sleep(...); ... }`, started at `main.rs:1438-1440` when `LOOM_TMUX_HEALTH_*` env enables it.
- `metrics_collector::try_init_metrics_collector` (`loom-daemon/src/metrics_collector.rs:216`) — same `thread::spawn(move || loop { ... })` shape, started at `main.rs:1451`.
- A GitHub App token-refresh task (`main.rs:1557-1565`) — a `tokio::spawn(async move { loop { tokio::time::sleep(...).await; ... } })` that also calls `tokio::task::spawn_blocking` each iteration, started before `IpcServer::new`/`server.run()` at `main.rs:2318-2365`.

Caveat for whoever picks this up: in stock Rust, a process exits as soon as `fn main` returns — outstanding `std::thread::spawn` threads (the first two candidates) are **not** normally waited on and do not by themselves block process exit. The more likely mechanism, if the hypothesis in "Likely cause" is on the right track, is the *tokio* runtime shutdown path: `#[tokio::main]`'s generated wrapper drops its `Runtime` after `block_on` returns, and `Runtime::drop` blocks the calling thread until in-flight work on the blocking-task pool (`spawn_blocking`) drains — so the GitHub App refresh task's `spawn_blocking` call, or any other `spawn_blocking` invocation live in the runtime at shutdown time, is a stronger lead than the two raw `thread::spawn` health/metrics threads. Worth confirming with a minimal repro that strips each candidate one at a time (`LOOM_TMUX_HEALTH_INTERVAL` unset, no GitHub App script configured, etc.) before committing to a fix location.

### Affected Files

- `loom-daemon/src/main.rs` — `server.run().await?` call site (~line 2365) and the async main function's overall shutdown path; also where `health_monitor`, `metrics_collector`, and the GitHub App refresh task are started (~lines 1438-1565).
- `loom-daemon/src/ipc.rs` — `IpcServer::run`'s singleton-guard `anyhow::bail!` (~line 785).
- `loom-daemon/src/health_monitor.rs` — `start_tmux_health_monitor`, a raw `thread::spawn` loop (~line 57).
- `loom-daemon/src/metrics_collector.rs` — `try_init_metrics_collector`, a raw `thread::spawn` loop (~line 216).
- `loom-daemon/tests/integration_singleton_guard.rs` — `test_second_daemon_refuses_and_first_survives` (10s wait) and `test_stale_socket_is_reclaimed` (fixed 2s startup budget) both need their fixed waits replaced with a generous polling budget per the issue's "Suggested fix".

### Test Plan

- [ ] Manual verification: run the reproduction script in the issue body on a host with a populated `~/.loom` (or with `LOOM_TMUX_HEALTH_INTERVAL` set so the health monitor actually starts) and confirm the second daemon now exits non-zero promptly instead of hitting the `timeout 15` kill.
- [ ] Automated tests: `cargo test -p loom-daemon --test integration_singleton_guard` — both `test_second_daemon_refuses_and_first_survives` and `test_stale_socket_is_reclaimed` must pass on a host with an initialized `~/.loom` (the condition that currently reproduces the hang), not just on a pristine CI runner.
- [ ] Edge cases: verify the fix doesn't change the *first* (surviving) daemon's behavior — it must still answer `Ping` on its socket after a second daemon's refused start — and that the refusal message is still printed to stderr before exit (some fix shapes, e.g. an early `std::process::exit`, risk truncating buffered output).

### Dependencies

None — no blocking issues.

rjwalters:
## Summary

A second `loom-daemon` started against a socket a live daemon already owns
correctly detects the incumbent and prints the singleton-guard refusal
(`ipc::IpcServer::run`, #3806), but the process never exits — it hangs
indefinitely instead of exiting non-zero promptly.

## Root cause

The refusal path returns `Err` out of the `#[tokio::main]`-generated async
`main`. That wrapper drops its `Runtime` when `block_on` returns, and
`Runtime::drop` blocks the calling thread until every in-flight
`spawn_blocking` task on the runtime finishes. On a host with real subsystems
configured (role runner, work finder, GitHub App token refresh, etc.), those
tasks were already spawned before the guard check ran, so the refusing
daemon sat in `Runtime::drop` for ~10s before the process actually
terminated — indistinguishable from a hang under a shorter timeout. On a
pristine CI runner those subsystems no-op, so the wart was invisible there.

## Changes

- `loom-daemon/src/main.rs`: `main` is now a thin synchronous wrapper around
  a new `run_daemon()` async fn (the previous body of `main`). On `Err`,
  `main` prints the same `Error: {err:?}` line `Termination for Result` used
  to write, flushes stderr, and calls
  `std::process::exit(loom_daemon::ipc::EXIT_STARTUP_FAILURE)` directly —
  terminating the process immediately instead of waiting on runtime
  shutdown. This covers every startup `?`, not just the singleton guard.
- `loom-daemon/src/ipc.rs`: adds `EXIT_STARTUP_FAILURE = 1` (the same value
  `ExitCode::FAILURE` carried), documents the exit-code table, and extends
  the existing exit-code invariant test to assert it stays non-zero and
  equal to `1`.
- `loom-daemon/tests/integration_singleton_guard.rs`:
  - Raw daemons spawned by these tests now get the same fail-closed
    environment `TestDaemon::start` uses (`LOOM_NO_RESTORE`,
    `LOOM_ROLE_RUNNER=0`, `LOOM_WORK_FINDER=0`, `LOOM_EPIC_SUPERVISOR=0`,
    isolated workspace paths) so an inert test daemon doesn't spend ~10s
    doing live forge work before ever reaching the guard — the actual
    source of the tests' load-sensitivity.
  - `test_second_daemon_refuses_and_first_survives` now reads the child's
    stderr on a helper thread so it can time from "refusal printed" to
    "process exited" (a 5s regression budget, immune to startup latency)
    instead of asserting on wall-clock time from process spawn.
  - `test_stale_socket_is_reclaimed` replaces its fixed 2s "did not exit
    early" sleep with a poll for the leftover regular file becoming a real
    socket, removing the load-sensitive fixed wait entirely.
- `defaults/docs/daemon-reference.md`: documents the new
  immediate-exit-on-startup-failure behavior.

- [x] Second daemon refusing to start now exits promptly (non-zero) instead
      of hanging — verified via `test_second_daemon_refuses_and_first_survives`,
      which asserts exit lag after the refusal line is printed stays under a
      5s budget.
- [x] The refusal message is still printed to stderr before exit (message +
      exit-code contract preserved: `EXIT_STARTUP_FAILURE == 1 ==
      ExitCode::FAILURE`).
- [x] The first (surviving) daemon's behavior is unaffected — it still
      answers `Ping` on its socket after the second daemon's refused start.
- [x] `test_stale_socket_is_reclaimed`'s fixed 2s wait replaced with a
      generous polling budget so it isn't load-sensitive.

## Test Plan

- `cargo build -p loom-daemon --bin loom-daemon` — builds clean.
- `cargo test -p loom-daemon --test integration_singleton_guard -- --test-threads=1`
  — both tests pass:
  ```
  test test_second_daemon_refuses_and_first_survives ... ok
  test test_stale_socket_is_reclaimed ... ok
  ```
- `cargo test -p loom-daemon --lib ipc::` — 76 passed, 0 failed.
- `cargo test -p loom-daemon --bin loom-daemon` — 52 passed, 0 failed
  (includes the updated `EXIT_STARTUP_FAILURE` exit-code invariant test).
- `cargo clippy -p loom-daemon --bin loom-daemon --tests` — clean.

Closes #4531

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -3480,6 +3480,17 @@ its launchd label) that no daemon is still alive after the stop and exits non-ze
 if one is, closing the silent-success hole where a failed bootout could leave a
 relaunched daemon dispatching.
 
+**Startup failures end the process directly (#4531).** A daemon that cannot start
+— most visibly the singleton guard's "another loom-daemon is already listening …
+— refusing to start" refusal (#3806) — prints that error to stderr and calls
+`std::process::exit(1)` itself rather than returning `Err` out of `main`. The
+message and the exit code are exactly what `Termination for Result` used to
+produce; what changes is that termination is now *immediate*. Returning `Err`
+made `#[tokio::main]` drop its `Runtime`, which blocks until every in-flight
+`spawn_blocking` task finishes — so a refusing second daemon kept running its
+already-spawned periodic loops for ~10s after announcing it would not start, long
+enough to look like a hang under any short timeout.
+
 ### Autonomy-loss watchdog + heartbeat (#4011)
 
 **The failure this closes.** On 2026-07-26 the `loom-daemon` launchd job took a
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -42,6 +42,7 @@ const LIVENESS_PROBE_TIMEOUT: Duration = Duration::from_millis(500);
 //   | SIGTERM (operator stop)       | 143 EXIT_SIGTERM| NO relaunch       |
 //   | SIGINT  (interactive Ctrl-C)  | 130 EXIT_SIGINT | NO relaunch       |
 //   | IPC Shutdown request          | 143 EXIT_SHUTDOWN| NO relaunch      |
+//   | startup failure (e.g. #3806)  | 1 EXIT_STARTUP_FAILURE | NO relaunch |
 //   | crash / panic                 | non-zero        | NO relaunch       |
 //
 // This is Curator Finding 1's remedy: because a SIGTERM'd daemon now exits
@@ -63,6 +64,22 @@ pub const EXIT_SIGINT: i32 = 130;
 /// Exit code for an explicit IPC `Shutdown` request. Non-zero — an explicit
 /// shutdown means "stay down", so launchd must not relaunch.
 pub const EXIT_SHUTDOWN: i32 = 143;
+/// Exit code for a startup failure the daemon reports and terminates on itself
+/// (#4531) — most visibly the singleton guard's "another loom-daemon is already
+/// listening" refusal below.
+///
+/// Deliberately `1`, the exact value `std::process::ExitCode::FAILURE` carries:
+/// `main` previously let such errors propagate out of `#[tokio::main]` and relied
+/// on `Termination for Result` to print `Error: {err:?}` and exit `1`. That path
+/// is correct but **not prompt** — the generated wrapper drops the `Runtime`
+/// after `block_on` returns, and `Runtime::drop` blocks until every in-flight
+/// `spawn_blocking` task finishes, which on a host with real work configured
+/// stalled the refusing process for ~10s (and looked like an indefinite hang
+/// under a shorter timeout). `main` now prints the same message and exits with
+/// this code directly, so the observable contract (message + status) is
+/// unchanged while termination becomes immediate. Non-zero ⇒ launchd does not
+/// relaunch, which is what a refusal wants.
+pub const EXIT_STARTUP_FAILURE: i32 = 1;
 
 /// Detect the daemon's process supervisor from the environment (#4054, #4267).
 ///
@@ -5229,6 +5246,11 @@ exit 0
         assert_ne!(EXIT_SIGTERM, 0, "SIGTERM stop must be non-zero (no relaunch)");
         assert_ne!(EXIT_SIGINT, 0, "SIGINT/Ctrl-C must be non-zero (no relaunch)");
         assert_ne!(EXIT_SHUTDOWN, 0, "explicit Shutdown must be non-zero (no relaunch)");
+        // #4531: the self-reported startup-failure exit must stay non-zero (no
+        // relaunch) AND keep the value `ExitCode::FAILURE` used to produce, so
+        // callers that only knew the old `Termination`-driven exit see no change.
+        assert_ne!(EXIT_STARTUP_FAILURE, 0, "startup failure must be non-zero (no relaunch)");
+        assert_eq!(EXIT_STARTUP_FAILURE, 1, "startup failure must match ExitCode::FAILURE");
 
         // Supervised: scheduled + do_exit.
         std::env::set_var("LOOM_DAEMON_SUPERVISOR", "launchd");
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -1691,8 +1691,45 @@ mod gh_token_guard_tests {
     }
 }
 
+/// Process entry point.
+///
+/// The real body lives in [`run_daemon`]; this wrapper exists so a startup
+/// failure **terminates the process itself** instead of returning `Err` out of
+/// `#[tokio::main]` (#4531).
+///
+/// Returning `Err` is correct but not prompt: the `#[tokio::main]` wrapper drops
+/// its `Runtime` when `block_on` returns, and `Runtime::drop` blocks the calling
+/// thread until every in-flight `spawn_blocking` task finishes. By the time the
+/// singleton guard refuses to start (`IpcServer::run`, #3806), the daemon has
+/// already spawned its periodic loops, several of which do their work on the
+/// blocking pool — so the refusal message printed, then the process sat in
+/// `Runtime::drop` for ~10s before `Termination` ever ran. Under a shorter
+/// timeout that is indistinguishable from a hang, and it left a doomed second
+/// daemon alive (running role-runner/work-finder ticks) long after it had
+/// decided not to start.
+///
+/// The observable contract is deliberately unchanged: the message is the same
+/// `Error: {err:?}` line `Termination for Result` would have written to stderr,
+/// and [`loom_daemon::ipc::EXIT_STARTUP_FAILURE`] is `1`, the value
+/// `ExitCode::FAILURE` carries.
+/// Only the latency changes. stderr is explicitly flushed before
+/// `std::process::exit` so the refusal is never truncated (`exit` runs no
+/// destructors); the log file needs no such care because `env_logger` flushes
+/// each record as it writes it — the same assumption every other
+/// `std::process::exit` path in this daemon already makes.
 #[tokio::main]
-async fn main() -> Result<()> {
+async fn main() {
+    if let Err(err) = run_daemon().await {
+        eprintln!("Error: {err:?}");
+        let _ = std::io::stderr().flush();
+        std::process::exit(loom_daemon::ipc::EXIT_STARTUP_FAILURE);
+    }
+}
+
+/// The daemon's real entry point: CLI dispatch, then full daemon startup ending
+/// in the IPC accept loop (which only returns on a startup failure — a running
+/// daemon leaves via one of the `std::process::exit` paths instead).
+async fn run_daemon() -> Result<()> {
     let cli = Cli::parse();
 
     // Handle CLI commands (init mode)
--- a/loom-daemon/tests/integration_singleton_guard.rs
+++ b/loom-daemon/tests/integration_singleton_guard.rs
@@ -12,8 +12,11 @@ mod common;
 
 use common::{cleanup_all_loom_sessions, daemon_bin, TestClient, TestDaemon};
 use serial_test::serial;
+use std::io::BufRead;
+use std::os::unix::fs::FileTypeExt;
 use std::path::Path;
-use std::process::{Command, Stdio};
+use std::process::{Child, Command, Stdio};
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 fn setup() {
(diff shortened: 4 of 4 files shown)


rjwalters:
Code evaluation looks good so far; CI is still running (Rust Unit Tests, Rust Linting & Build Check, Installer Integration Tests, Shell Test Suites all in_progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T15:58:02Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T15:58:39Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: checked out and reviewed context, but several required CI checks (Rust Unit Tests, Rust Linting & Build Check, Installer Integration Tests, Shell Test Suites) are still `in_progress`. Not blocking on them per the Judge's CI-pending protocol — releasing the `loom:reviewing` claim and leaving this in the review queue. A later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation pending; CI is still running (Shell Test Suites (hermetic) and Rust Unit Tests are in-progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation looks reasonable on a first pass (clean fix for the Runtime::drop hang, well-tested); CI is still running (Rust Unit Tests and Shell Test Suites hermetic both in_progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation pending; CI is still running (Rust Unit Tests, Shell Test Suites in progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
## Judge review: APPROVED

Correct diagnosis, minimal fix, and the test rework is the better half of the PR.

### Verified locally (worktree `.loom/worktrees/pr-4639`, head `48ef576c`)

- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test -p loom-daemon --test integration_singleton_guard` — **2/2 pass in 2.71s** (this is the host the issue said reproduced the hang)
- `cargo test -p loom-daemon --lib ipc::` — 76/76 pass
- CI on head: 8 pass / 1 skipping; `mergeStateStatus=CLEAN`
- Manual repro from the issue body (autonomy env fail-closed, temp socket): second daemon exited **`exit=1` in 0.59s**, refusal message intact on stderr, incumbent's socket still present and answering. Previously `timeout 15` killed it (`exit=124`).

### Failure-mode scrutiny (the things I went looking for)

**Does the exit path leak state?** No. `run_daemon()` returns `Err` *before* `main`'s error arm runs, so every local in the startup path — and its `Drop` — has already run by the time `std::process::exit` is called. The only thing skipped is `Runtime::drop`'s blocking-pool join, which is precisely the bug. The residual risk is a `spawn_blocking` task killed mid-flight while holding a filesystem lock; I checked both cross-process ones (`watch_registry::WatchLock`, `tokens_pool::locking::MkdirLock`) and both implement stale-lock stealing, so nothing is stranded permanently. `RoleRunGuard`, `IssueCreationGuard`, and `BuildSlotLease` are in-process only and die with the process.

**Socket / incumbent safety.** `IpcServer::run` bails *before* `remove_file` + `bind`, so the refusing daemon never touches the incumbent's socket — confirmed by both the test assertion and the manual repro.

**Crash loop under launchd/systemd?** No. `EXIT_STARTUP_FAILURE = 1` is byte-identical to what `Termination for Result` produced, and the daemon's contract is `KeepAlive { SuccessfulExit: true }` (launchd) / `Restart=on-success` (systemd) — only exit `0` relaunches. So a persistent startup failure stays down, exactly as before; only the latency changed. Pinning this with `assert_eq!(EXIT_STARTUP_FAILURE, 1, ...)` next to the existing exit-code contract test is the right place for it.

**Anything newly fatal?** No. The diff changes termination mechanics only — no `Ok`/`Err` decision anywhere in `run_daemon` was altered, and the success path still drops the `Runtime` normally. Degraded-mode startup paths that returned `Ok` still return `Ok`.

**Log/stderr truncation.** The explicit `stderr().flush()` is belt-and-braces (Rust's stderr is unbuffered), and the doc's claim about the log file holds: `setup_logging` uses `env_logger::Target::Pipe` over a raw `fs::File`, so each record is a direct write syscall with nothing buffered to lose.

### Test rework

`daemon_command`'s fail-closed env (`LOOM_ROLE_RUNNER/WORK_FINDER/EPIC_SUPERVISOR=0` + workspace confinement) is not cosmetic — without it these tests were spawning daemons against this repo's real `autonomous.roleRunner.enabled: true` config and doing live forge work. That was both a side-effect hazard and the load-sensitivity the tests were papering over with fixed sleeps. Splitting the bound into a generous *liveness* wait (spawn→exit, 60s) and a tight *latency* assertion measured from the refusal line (`REFUSAL_TO_EXIT_BUDGET`, 5s) is exactly the right decomposition — it makes the regression bound sharp without making it load-sensitive. The `exit_at`/`refusal_at` ordering race (poll loop can observe exit before the reader thread drains the line) is correctly absorbed by `saturating_duration_since`. `test_stale_socket_is_reclaimed` polling for `file_type().is_socket()` now asserts the bind directly instead of delegating it to `TestClient::connect`'s retry budget.

### Non-blocking notes (no action required for merge)

1. `eprintln!` panics on a stderr write error (EPIPE), where `Termination for Result` used the non-panicking `io::attempt_print_to_stderr`. Theoretical under launchd (stderr is a file); the consequence would be exit 101 instead of 1.
2. The "~10s" framing in both the doc and the code comment understates it: the issue's repro shows `timeout 15` killing the process (`exit=124`), so `Runtime::drop` was blocking for an *unbounded* time whenever a `spawn_blocking` task was parked on a `gh` subprocess. The fix is correct for the unbounded case; only the narrative is conservative.
3. Pre-existing and explicitly out of scope for #4531: the singleton guard still runs *after* the daemon has started its periodic loops, so a doomed second daemon can do live forge work in the window *before* it refuses. This PR closes the post-refusal window entirely; the pre-refusal one would want the guard hoisted ahead of subsystem startup. Worth a future look, not a blocker here.
4. The test helper still pipes `stdout` without draining it (pre-existing; `stderr` is now drained). Low risk since the daemon logs to a file, but a chatty stdout would deadlock the child on a full pipe buffer.

Acceptance criteria from #4531 are met: prompt non-zero exit with the message preserved, and both integration tests now poll with generous budgets instead of fixed sleeps. Docs land in `defaults/docs/daemon-reference.md`, which is what `.loom/docs/daemon-reference.md` symlinks to, so one edit covers both.

Approving — swapping `loom:review-requested` → `loom:pr`.

PR #4553, opened 2026-07-30
feat(safehouse): add title/additions/deletions/tokens to completion-v1 meta

rjwalters:
## Summary

The public fleet feed on 2amlogic.com renders completion entries as
`<repo>#<issue>: <title> +<additions> −<deletions> · <dur> · <tokens> tok`, and its
ingest already accepts and stores all four as extension fields — but the producer
only sent `{schema, agent, repo, ref, result, started_at, completed_at, issue}`.
This wires the four display fields at the existing emit point (the narration sink,
on `SweepExited`, after the forge-truth merge check). envelope-v1 preserves unknown
`meta` keys and safehoused's egress publishes the raw redacted `meta`, so **no
schema revision** was needed.

Per the operator's 2026-07-29 priority update on the issue, `tokens` is treated as a
**wanted** field rather than an optional nicety: it is wired to the activity DB's
per-issue rollup and its attribution imperfections are documented on the field, not
used as a reason to omit it.

## Changes

**`loom-daemon/src/safehouse.rs`**

- `fetch_merged_pr` now requests `--json number,url,mergedAt,title,additions,deletions`
  — the display fields come out of the **same** `gh pr list` call that already
  verifies the merge, so **zero extra forge round-trips**. Each is read with
  `and_then`/`filter` rather than `?`, so an absent or wrongly-typed one degrades
  that field alone and never costs the completion.
- **Old-`gh` safety net.** A `gh` that does not know one of the new fields rejects
  the *whole* request (it does not silently omit the field), which would have cost
  every completion on such a host. That one case — and only that one, matched on
  `unknown JSON field` in stderr — retries the pre-#4497 `number,url,mergedAt` set.
  Ordinary degradations (no auth, no network, not a repo) do **not** trigger a
  second spawn.
- `MergedPr` and `CompletionMeta` gain `title`/`additions`/`deletions`;
  `CompletionMeta::tokens` (which already existed, hardcoded `None`) is now
  populated.
- `fetch_issue_tokens` sums `total_input_tokens + total_output_tokens` over
  `ActivityDb::get_cost_by_issue`, on the **blocking pool** under a 5s cap — the
  rusqlite handle is behind a `std::sync::Mutex` shared with the IPC recorder's
  writes, so the sink must never park a daemon-runtime worker on it (and a std
  guard cannot cross an `await`).
- `to_meta_value` inserts each field only when present. `title` is trimmed and a
  blank one is dropped; `additions`/`deletions` publish **real zeros**, since `0`
[...]

PR #4646, opened 2026-07-30
fix(safehouse): reconcile champion-driven merges the SweepExited emit point misses

### LINKED ISSUE #4583: safehouse: champion-merged PRs never emit completions — the SweepExited emit point misses merges that happen after sweep exit
## Context (production evidence, 2026-07-29 evening)

The public fleet feed went silent for 2+ hours while rjwalters/loom main received ~8 merged PRs (#4551, #4553, #4558, #4560, #4562, #4567…). Egress last published 03:55Z; every one of those merges landed via **champion role ticks** (EC2 host log: `role_runner: champion tick completed for /home/ubuntu/GitHub/loom in 127.7s` at 05:42Z), not in-sweep merges.

#4459's emit point is the narration sink on `SweepExited`, gated on a forge-truth check ("did this issue's PR merge?"). That catches only the sweep lifecycle that completes its own merge **before exiting**. The common steady-state path — builder opens PR → judge approves → sweep exits → **champion merges minutes-to-hours later on a role tick** — produces no `SweepExited` at merge time, so no completion is ever emitted. Earlier the same day the feed worked because those happened to be full in-sweep merges; once champion-driven merging dominated, the feed silently starved. (Flagged as a design risk in the #4459 review: "PR merged is best observed where the champion learns the PR landed rather than SweepExited.")

The per-`(workspace, issue)` dedup map makes this worse for resumed issues: a later `SweepExited` for an issue whose earlier exit already consumed the dedup slot can never emit.

## Ask

Cover the champion-merge path. Options (implementer's choice):

1. **Emit at the merge funnel**: champion runs `merge-pr.sh`; after its merge verification, notify the daemon (this was the pre-#4459 design sketch — an IPC verb like `forge notify-merged`) which publishes the typed merged event through the existing sink → completion pipeline.
2. **Reconcile periodically**: a daemon-side pass (work-finder/role cadence) listing recently-merged PRs per workspace and emitting completions for any `(workspace, issue)` not already in the dedup set. Naturally also catches merges while the daemon was down.
3. Both — 1 for immediacy, 2 as the catch-all.

Either way the dedup set stays the single gate against double-posting (downstream ingest is additionally idempotent on event_id).

- [ ] A PR merged by a champion role tick (no live sweep at merge time) produces exactly one completion envelope with the #4553 meta.
- [ ] In-sweep merges still emit exactly once (no double-post with the new path).
- [ ] A merge occurring while the daemon is down is picked up after restart (if option 2 is chosen).

---

## Curator Enhancement (consolidated 2026-07-30, anchors verified against origin/main @ 15e91055)

> This section consolidates and **supersedes** the three earlier curator comments (06:24, 06:25, 06:26 UTC) — their line anchors had rotted by several hundred lines within hours. Where they conflicted, the numbers below are the verified ones.

### Root cause (confirmed in code)

- The **sole** emit point for the public-feed `completion` envelope is `completion_for_exit()` (`loom-daemon/src/safehouse.rs:1635`), called only from `run_sink()`'s (`safehouse.rs:2022`) `Event::SweepExited` gate (`safehouse.rs:2106`). Nothing else ever reaches it, so a champion-tick merge — happening after the sweep's one `SweepExited` — is never checked.
- Dedup is the `completed: HashSet<(String, u32)>` local to `run_sink` (`safehouse.rs:2065`, passed as the `already_narrated` param). It is **in-memory only** — it does not survive a daemon restart, and there is no daemon-wide shared dedup state. Any new trigger path must share this set (or its `(workspace, issue)` key convention) to preserve the exactly-once invariant.
- Forge-truth check to reuse, not duplicate: `fetch_merged_pr()` (`safehouse.rs:1423`); envelope construction: `build_completion_envelope()` (`safehouse.rs:1387`).
- `Event::SweepExited` is defined at `loom-daemon/src/types.rs:1607` and emitted only by the daemon's sweep registry/reaper — never by champion or any external actor.
- Design rationale for the original (incomplete) emit-point choice: `.loom/docs/safehouse.md` (emit point + completions contract, ~lines 527–560) — it assumed all merges are observable at `SweepExited` time, which stopped holding once champion-driven merging dominated.

### Design constraint the implementer must account for

`defaults/.claude/commands/loom/sweep.md` documents a **frozen 6-topic event taxonomy** (taxonomy table at ~line 2049; "new topics beyond the six listed — frozen for v0.10.0 … file a follow-up issue" at ~line 2127). `Event::from_published` (`types.rs:1932`) / `try_typed_child_event` (`types.rs:1944`) only upgrade `sweep.issue.{N}.phase` / `.blocker` topics; anything else falls through to `Event::Generic`, which is never narrated. So **Option 1 implemented as a new event topic would violate the frozen taxonomy and requires its own follow-up issue** — prefer a new IPC `Request` verb (alongside the existing variants in `types.rs:19`, handled in `ipc.rs` `handle_request` at ~1984; see the existing `Request::PublishEvent` arm at `ipc.rs:2862` for the pattern), or pick Option 2 (periodic reconciliation), which sidesteps the taxonomy question entirely. State the trade-off explicitly in the PR description whichever option is chosen.

### Existing hook points

- **Option 1 (script level)**: `defaults/scripts/merge-pr.sh` (canonical source; `.loom/scripts/merge-pr.sh` is the installed copy) has a confirmed-merge choke point at the `"PR #$PR_NUMBER merged successfully"` line (~1550), immediately followed by best-effort post-merge hooks `_reset_partial_increment_labels || true` (~1558) and `_auto_reconcile_stacked_children || true` (~1564) — a `_notify_daemon_merged || true` call fits this exact pattern. The merge API call is `forge_merge_pr` (~1482).
- **Option 1 (role level)**: champion's Step 4 "Verify Issue Auto-Close" in `defaults/.claude/commands/loom/champion-pr-merge.md` (~line 369) already resolves the merged-issue set per PR and loops over each issue.
- **Option 2**: `loom-daemon/src/work_finder.rs` `tick` (~576) / `tick_multi` (~860) and the `role_runner.rs` tick loop are the existing periodic-cadence hooks; `claim_reconciliation.rs` / `quarantine_reconciliation.rs` are structural precedent for a new `merge_reconciliation.rs`. Option 2 is also the path that naturally covers merges while the daemon was down — but note the restart AC needs a persisted-or-rescanned dedup source, since `completed` is in-memory.
- The mcp-loom `publish_event` MCP tool exists but is documented as an operator/testing escape hatch, not a production path — call it out explicitly if repurposed.

## Affected Files

- `loom-daemon/src/safehouse.rs` — `completion_for_exit()` (1635), `run_sink()` (2022), `completed` dedup set (2065), `SweepExited` gate (2106), `fetch_merged_pr()` (1423), `build_completion_envelope()` (1387); extend the inline test module (exactly-once test `completion_for_exit_emits_at_most_once_per_merge` at 4526)
- `loom-daemon/src/types.rs` — `Request` enum (19); `Event::SweepExited` (1607); `from_published` (1932) / `try_typed_child_event` (1944) if any topic-parsing change (see frozen-taxonomy constraint)
- `loom-daemon/src/ipc.rs` — `handle_request` (1984); new verb arm alongside `Request::PublishEvent` (2862) if Option 1 is IPC-based
- `defaults/scripts/merge-pr.sh` — post-merge-success hook point (~1550–1564) if Option 1 is script-level
- `loom-daemon/src/work_finder.rs` — `tick`/`tick_multi` (576/860); Option 2 cadence hook (or model for a new `merge_reconciliation.rs`)
- `loom-daemon/src/role_runner.rs` — alternate cadence hook for Option 2
- `defaults/.claude/commands/loom/champion-pr-merge.md` — Step 4 (~369) if the notify call is issued from the role instead of the script
- `.loom/docs/safehouse.md` — update the emit-point/dedup design section (~527–560) to document the new path
- `defaults/.claude/commands/loom/sweep.md` — frozen-taxonomy table (~2049) **only** if a new event topic is added, which requires its own follow-up issue per that doc's rule

## Test Plan

- [ ] Automated: a champion-driven merge (no live sweep / no `SweepExited` at merge time) produces exactly one `completion` envelope carrying the #4553 meta (title/additions/deletions/tokens where available) — add alongside `completion_for_exit_emits_at_most_once_per_merge` (`safehouse.rs:4526`)
- [ ] Regression: an in-sweep merge (existing `SweepExited` path) still emits exactly once — no double-post now that a second trigger path exists (shared dedup set verified)
- [ ] If Option 2 (or 3): a merge occurring while the daemon is down is picked up and narrated after restart — requires a persisted-or-rescanned dedup source, not the in-memory `completed` set alone
- [ ] Edge case: two merges in quick succession (one in-sweep, one champion-driven) for different issues in the same workspace — both emit independently, no dedup-key cross-contamination
- [ ] Manual: run `merge-pr.sh` against an approved PR with no active sweep and confirm exactly one completion reaches the safehouse feed

## Dependencies

None — the referenced issues/PRs (#4459, #4426, #4497, #4553, #4201) are all merged; this fixes a gap in that already-shipped design.

rjwalters:
## Summary

Closes #4583.

The public-feed `completion` envelope had exactly one emit point: `run_sink`'s `Event::SweepExited` gate. That gate can only ever fire while a sweep process is alive to exit, so it structurally cannot observe the **common steady-state path** — builder opens a PR, judge approves, the sweep exits, and champion merges it minutes-to-hours later on its own role tick. Once champion-driven merging dominated, the feed starved silently (2+ hours dark across ~8 merged PRs).

**Option implemented: Option 2 — periodic reconciliation.** Rationale:

- **It sidesteps the frozen event taxonomy entirely.** `defaults/.claude/commands/loom/sweep.md` freezes the 6-topic vocabulary for v0.10.0, and `Event::from_published` / `try_typed_child_event` only upgrade the `.phase` / `.blocker` topics — anything else falls through to `Event::Generic`, which is never narrated. Option 1 as a new topic would violate that freeze and need its own follow-up issue; as a new IPC verb it would avoid the freeze but still only cover merges made through `merge-pr.sh` on a host whose daemon is up and reachable at that instant.
- **It is the only option that covers the daemon-down case**, which is one of the acceptance criteria. Reconciliation queries forge truth directly rather than depending on any daemon-emitted signal, so a merge that landed while the daemon was stopped is picked up on the next tick after restart.
- **It has no new coupling surface**: no protocol change, no script change, no role-prompt change. `merge-pr.sh` and champion are untouched, so there is no way for a notify call to be skipped, fail silently, or be issued by a version-skewed script.

The trade-off, stated explicitly: reconciliation is **not immediate**. A champion merge is narrated on the next tick for its workspace rather than at merge time — up to 5 minutes for a single-workspace host, and `interval × workspaces` on a multi-repo host, since one workspace is reconciled per tick. Option 1 would be sub-second. Given the feed is a monitoring surface with a delay buffer downstream, minutes of latency is a far better trade than the coverage gaps above; Option 1 remains available as a latency optimization layered on top (the shared dedup core is already the single gate that would keep the two from double-posting).

## Changes

**`loom-daemon/src/safehouse.rs`**

- Extracted `build_and_narrate_completion` — the shared envelope-build + dedup-insert core — out of `completion_for_exit`. Both trigger paths funnel through it, so "exactly one completion per merge" holds regardless of which path observes a merge first. It takes `duration_sec`/`exited_at` from the caller because the two paths have different clocks available (the reaper's for a live sweep; the forge's `createdAt`/`mergedAt` pair when no sweep clock exists).
- Added `reconcile_recent_merges` + `fetch_recent_merged_prs`: one **bulk**, branch-unfiltered `gh pr list --state merged --limit 30` per pass, with each row's issue number recovered from `headRefName` via `worktree_ops::naming::issue_from_branch` (non-matching branches are skipped — not every merged PR is an issue sweep). Every failure mode (missing `gh`, no network, timeout, malformed JSON, unparsable row) degrades to skipping that row or the whole pass; it can never panic or block the sink.
- Wired it into `run_sink` via `tokio::select!` on a 5-minute interval timer (`LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS` overrides, primarily a test seam). One iteration handles a bus event **or** a reconciliation tick, never both, so the dedup set is only ever touched from one place at a time. Targets are the union of workspaces observed on stamped-`repo` events and the on-disk `WorkspaceRegistry`, round-robinned one per tick so a slow/offline `gh` in one workspace cannot delay every other workspace's narration.
- **Persisted the dedup set** to `~/.loom/safehouse-completed.json` (`LOOM_SAFEHOUSE_COMPLETIONS_PATH` overrides), loaded once at startup and rewritten atomically (temp + rename) after each new completion. This is what makes the restart criterion work in both directions: a merge from while the daemon was down still emits, and a merge already narrated before the restart does not re-emit. Every read/write failure is best-effort and logged at `debug`.
- **Lookback window (7 days, `LOOM_SAFEHOUSE_RECONCILE_MAX_AGE_SECS`)**: `--limit 30` bounds a burst's *size* but not its *age*. Without a window, the first pass on a host with no persisted dedup set — fresh install, the upgrade to this version, a lost/corrupt completions file — would backfill the feed with the last 30 merges however old they were, which in a low-traffic workspace means narrating months-old PRs as if they had just landed. Out-of-window rows are dropped *before* the dedup check, so they are never recorded as narrated and a later in-window observation can still emit. Seven days is well beyond any plausible daemon outage, so daemon-down recovery is unaffected.
- Test isolation: `run_sink` now reads two machine-level paths (the completions file and the workspace registry). A `SafehouseTestPaths` RAII guard redirects both into a tempdir for every `run_sink` test, so the suite can never read or write a real daemon's state, and reconciliation targets do not depend on whatever repos happen to be registered on the host running the tests.

**`.loom/docs/safehouse.md`** — documents both emit points, the reconciliation cadence/query/window, and the now-shared, now-persisted dedup contract. Also corrects a cadence description that referenced a `safehouse.reconcileInterval` config key that does not exist.

- [x] **A champion-tick merge (no live sweep at merge time) produces exactly one completion with the #4553 meta.** `reconcile_recent_merges_emits_completion_for_a_champion_driven_merge` asserts one envelope carrying `schema`/`issue`/`result`/`title`/`additions`/`deletions`. `run_sink_reconciliation_narrates_a_champion_merge_with_no_live_sweep` proves it end-to-end on the wire with **no `SweepExited` ever published** — the workspace is discovered from the registry alone.
- [x] **In-sweep merges still emit exactly once; no double-post.** `reconcile_recent_merges_skips_a_merge_already_narrated_in_sweep` runs the `SweepExited` path first, then a reconciliation pass over the same merge, and asserts the second emits nothing — while also asserting both `gh` queries still ran, i.e. the skip happens at the shared dedup check rather than by accident. The pre-existing `run_sink_narrates_exit_ack_then_completion`, `run_sink_narrates_only_the_ack_when_nothing_merged` and `completion_for_exit_emits_at_most_once_per_merge` pass unmodified.
- [x] **A merge while the daemon was down is picked up after restart.** The persisted dedup set is the mechanism. `load_persisted_completed_round_trips_through_persist_completed_best_effort` covers the write/read contract, `load_persisted_completed_degrades_to_empty_when_the_file_is_absent_or_corrupt` the failure modes, and `reconcile_recent_merges_does_not_repost_a_merge_loaded_from_a_persisted_restart_file` simulates a fresh process loading a pre-restart file and confirms the already-narrated merge is not re-posted (the flip side: any merge *not* in that file, including one that landed during the outage, still emits).
- [x] **Two merges in quick succession for different issues emit independently.** `reconcile_recent_merges_narrates_distinct_issues_independently` puts one already-narrated (in-sweep) and one fresh (champion-driven) issue in the same bulk response for the same workspace, and asserts only the un-narrated one emits, with both keys ending up in the dedup set — no cross-contamination.

## Test Plan

- `cargo test` in `loom-daemon`: **2537 lib tests + all integration suites pass, 0 failed.** `cargo fmt --check` and `cargo clippy --all-targets --all-features` are clean.
- Seven new tests (five reconciliation behaviors, two persistence, one window + one window-config), all `#[serial]` since they drive process-global env seams.
- Reconciliation fixtures generate `Utc::now()`-relative timestamps rather than literal dates — with the lookback window in place, hard-coded dates would silently turn every reconciliation assertion into a no-op once wall-clock time drifted a week past them.

**Manual verification against `merge-pr.sh` with no active sweep** (not run here — it requires a live daemon with safehouse configured, which this worktree does not have): start `loom-daemon` with safehouse enabled and `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS=15000`, confirm no sweep is running (`mcp__loom__list_sweeps`), then `./.loom/scripts/merge-pr.sh <PR>` on an approved PR. Within one interval exactly one `completion` envelope for that issue should appear in the room (and `~/.loom/safehouse-completed.json` should gain its `(workspace, issue)` pair). Re-running reconciliation, or publishing a late `SweepExited` for the same issue, must produce nothing further.

## Notes for the reviewer

- `merge-pr.sh`, `champion-pr-merge.md` and the event taxonomy are deliberately **untouched** — see the Option 2 rationale above.
- The completions file grows by one `["<workspace>", <issue>]` pair (~32 bytes) per completion and is never pruned. At Loom's own merge rate that is a couple of MB per decade, so compaction is not implemented; it is documented rather than engineered.
- The first commit on this branch was originally pushed by an earlier session that exited before opening a PR; it is rebased onto current main here, and the superseded pre-rebase commit is folded in with an `ours` merge so the push stayed fast-forward.

### DIFF
--- a/.loom/docs/safehouse.md
+++ b/.loom/docs/safehouse.md
@@ -552,13 +552,35 @@ safehoused's egress subsystem mirrors well-formed **`completion`** envelopes out
 of allowlisted rooms — redacted and delay-buffered — to a `sink_url`; that is
 what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
 
-- **Emit point**: the narration sink, on `SweepExited`. Exit status alone proves
-  nothing, so the sink checks **forge truth** (`gh pr list --head
-  feature/issue-N --state merged`, in the event's workspace root, 10s timeout)
-  and emits the `completion` only when that issue's PR actually merged — the
-  `ack` still goes out either way. Chosen over having the sweep child publish a
-  post-merge phase event because it is daemon-only (no skill edit), has the
-  sweep timing to hand, and verifies rather than trusts.
+- **Emit point (two, since #4583)**:
+  1. The narration sink, on `SweepExited`. Exit status alone proves nothing, so
+     the sink checks **forge truth** (`gh pr list --head feature/issue-N
+     --state merged`, in the event's workspace root, 10s timeout) and emits
+     the `completion` only when that issue's PR actually merged — the `ack`
+     still goes out either way. Chosen over having the sweep child publish a
+     post-merge phase event because it is daemon-only (no skill edit), has the
+     sweep timing to hand, and verifies rather than trusts.
+  2. **Periodic merge reconciliation** (`reconcile_recent_merges`, #4583): the
+     `SweepExited` path only ever fires while a sweep process is alive to exit
+     — it structurally cannot see a PR that merges *after* the sweep already
+     exited, which is the **common steady-state path**: builder opens a PR,
+     judge approves, the sweep exits, and champion merges it minutes-to-hours
+     later on its own cron tick. On a fixed cadence (`safehouse.reconcile
+     Interval`, default 5 minutes; `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS`
+     overrides it, primarily a test seam), the sink round-robins one workspace
+     at a time — drawn from the union of every repo it has observed a
+     stamped-`repo` event for and the on-disk `WorkspaceRegistry` — and runs
+     one **bulk**, branch-unfiltered `gh pr list --state merged --json
+     number,headRefName,url,mergedAt,createdAt,title,additions,deletions
+     --limit 30`. Each row's issue number is recovered from `headRefName` via
+     the `feature/issue-<N>` convention
+     (`worktree_ops::naming::issue_from_branch`); rows that do not match are
+     skipped (not every merged PR is an issue sweep). A row with no sweep clock
+     available uses the PR's own `createdAt`/`mergedAt` pair as the
+     `started_at`/`completed_at` proxy instead of the reaper's clock. Both
+     paths funnel through the same envelope-build/dedup-insert core
+     (`build_and_narrate_completion`), so "exactly one completion per merge"
+     holds regardless of which path observes it first.
 - **`meta` (`completion-v1`)**: `{schema, agent, repo, ref, result, started_at,
   completed_at}` required, plus optional `issue`/`tokens`/`title`/`additions`/
   `deletions` (envelope-v1 preserves unknown `meta` keys, so no schema rev is
@@ -598,9 +620,18 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
   unfinished, not failed, and is usually resumed). The wire support exists
   (`CompletionResult::Failure`) for a follow-up that identifies a genuinely
   terminal negative outcome.
-- **At most one per merge**, deduped on `(workspace, issue)` for the daemon's
-  lifetime, so a resumed sweep's second `SweepExited` does not double-post.
-  Downstream ingest is additionally idempotent on `event_id`.
+- **At most one per merge**, deduped on `(workspace, issue)` — shared by
+  **both** emit points above, so a resumed sweep's second `SweepExited` does
+  not double-post, and the periodic reconciliation pass does not re-post a
+  merge the `SweepExited` path already narrated (in either order — whichever
+  path observes the merge first wins; the other becomes a no-op dedup check).
+  This dedup set is **persisted** to `~/.loom/safehouse-completed.json`
+  (`LOOM_SAFEHOUSE_COMPLETIONS_PATH` overrides the path) and reloaded at
+  startup — the in-memory set alone would not survive a daemon restart, which
+  would otherwise either re-post every prior completion (if reconciliation's
+  lookback window still covered them) or silently drop a merge that happened
+  while the daemon was down. Downstream ingest is additionally idempotent on
+  `event_id`.
 - **Strict client-side construction.** safehoused **silently degrades a
   malformed `meta` to `chat`** — the event then vanishes from the feed with no
   error anywhere — so `build_send_request` refuses to send a `completion` unless
--- a/.loom/docs/safehouse.md
+++ b/.loom/docs/safehouse.md
@@ -565,12 +565,12 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
      — it structurally cannot see a PR that merges *after* the sweep already
      exited, which is the **common steady-state path**: builder opens a PR,
      judge approves, the sweep exits, and champion merges it minutes-to-hours
-     later on its own cron tick. On a fixed cadence (`safehouse.reconcile
-     Interval`, default 5 minutes; `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS`
-     overrides it, primarily a test seam), the sink round-robins one workspace
-     at a time — drawn from the union of every repo it has observed a
-     stamped-`repo` event for and the on-disk `WorkspaceRegistry` — and runs
-     one **bulk**, branch-unfiltered `gh pr list --state merged --json
+     later on its own cron tick. On a fixed cadence (5 minutes;
+     `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS` overrides it, primarily a test
+     seam), the sink round-robins one workspace at a time — drawn from the
+     union of every repo it has observed a stamped-`repo` event for and the
+     on-disk `WorkspaceRegistry` — and runs one **bulk**, branch-unfiltered
+     `gh pr list --state merged --json
      number,headRefName,url,mergedAt,createdAt,title,additions,deletions
      --limit 30`. Each row's issue number is recovered from `headRefName` via
      the `feature/issue-<N>` convention
@@ -581,6 +581,16 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
      paths funnel through the same envelope-build/dedup-insert core
      (`build_and_narrate_completion`), so "exactly one completion per merge"
      holds regardless of which path observes it first.
+     - **Lookback window**: rows merged more than **7 days** ago are dropped
+       before the dedup check (`LOOM_SAFEHOUSE_RECONCILE_MAX_AGE_SECS`
+       overrides; garbage/negative values fall back to the default). `--limit
+       30` bounds a burst's *size*, not its *age* — without the window, the
+       first pass on a host with no persisted dedup set (fresh install, the
+       upgrade to this version, a lost/corrupt completions file) would backfill
+       the feed with the last 30 merges however old they are, which in a
+       low-traffic workspace means narrating months-old PRs as if they just
+       landed. Seven days is far beyond any plausible daemon outage, so the
+       daemon-was-down case below still recovers.
 - **`meta` (`completion-v1`)**: `{schema, agent, repo, ref, result, started_at,
   completed_at}` required, plus optional `issue`/`tokens`/`title`/`additions`/
   `deletions` (envelope-v1 preserves unknown `meta` keys, so no schema rev is
@@ -630,8 +640,13 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
   startup — the in-memory set alone would not survive a daemon restart, which
   would otherwise either re-post every prior completion (if reconciliation's
   lookback window still covered them) or silently drop a merge that happened
-  while the daemon was down. Downstream ingest is additionally idempotent on
-  `event_id`.
+  while the daemon was down. It is written atomically (temp file + rename) and
+  every failure to read or write it is best-effort — a corrupt or unwritable
+  file degrades to "the pre-#4583 behavior, plus a bounded re-narration window",
+  never to a crash. It grows by one `["<workspace>", <issue>]` pair (~32 bytes)
+  per narrated completion and is never pruned; at Loom's own merge rate that is
+  a couple of MB per decade, so no compaction is implemented. Downstream ingest
+  is additionally idempotent on `event_id`.
 - **Strict client-side construction.** safehoused **silently degrades a
   malformed `meta` to `chat`** — the event then vanishes from the feed with no
   error anywhere — so `build_send_request` refuses to send a `completion` unless
--- a/loom-daemon/src/safehouse.rs
+++ b/loom-daemon/src/safehouse.rs
@@ -1647,14 +1647,59 @@ async fn completion_for_exit(
         return None;
     }
     let merged = fetch_merged_pr(Path::new(workspace_root), issue).await?;
+    build_and_narrate_completion(
+        persona,
+        workspace_root,
+        issue,
+        merged,
+        duration_sec,
+        exited_at,
+        slug_cache,
+        already_narrated,
+        activity_db,
+    )
+    .await
+}
+
+/// Shared envelope-build/dedup-insert core behind **both** completion trigger
+/// paths (issue #4583): the `SweepExited` emit point above
+/// ([`completion_for_exit`]) and the periodic merge-reconciliation pass below
+/// ([`reconcile_recent_merges`]), which exists precisely because a
+/// champion-tick merge — landing well after the sweep process (and its one
+/// `SweepExited`) already exited — is never observed by the first path alone.
+///
+/// Every caller must have already confirmed the merge (a [`MergedPr`] in
+/// hand); this only re-checks and updates `already_narrated`, so the exactly-
+/// once invariant holds identically regardless of which path found the merge
+/// first. `duration_sec`/`exited_at` are caller-supplied rather than derived
+/// here because the two callers have different clocks available: a live
+/// sweep's reaper clock for [`completion_for_exit`], the forge's
+/// `createdAt`/`mergedAt` pair (no sweep clock exists) for
+/// [`reconcile_recent_merges`].
+#[allow(clippy::too_many_arguments)]
+async fn build_and_narrate_completion(
+    persona: &str,
+    workspace_root: &str,
+    issue: u32,
+    merged: MergedPr,
+    duration_sec: i64,
+    exited_at: DateTime<Utc>,
+    slug_cache: &mut HashMap<String, String>,
+    already_narrated: &mut std::collections::HashSet<(String, u32)>,
+    activity_db: Option<&Arc<Mutex<ActivityDb>>>,
+) -> Option<Envelope> {
+    let key = (workspace_root.to_owned(), issue);
+    if already_narrated.contains(&key) {
+        return None;
+    }
     let slug = fetch_repo_slug_cached(slug_cache, workspace_root).await?;
     // Only reached once the merge is confirmed, so a completion is never delayed
     // by a token lookup it would not have published.
     let tokens = fetch_issue_tokens(activity_db, issue).await;
 
-    // Sweep timing comes from the one clock that produced `duration_sec` (the
-    // reaper's), so the pair is always self-consistent — mixing in the forge's
-    // `mergedAt` could render a `completed_at` before `started_at`.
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T17:29:48Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T17:29:48Z, ~5 min old) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed against #4583's acceptance criteria. This is a careful, well-argued implementation of Option 2 — the option the curation explicitly favored — and the option rationale in the PR description correctly engages with the frozen-taxonomy constraint rather than working around it.

## What I verified

**Scope** — exactly two files (`loom-daemon/src/safehouse.rs`, `.loom/docs/safehouse.md`). No unrelated changes. Confirmed `.loom/docs/safehouse.md` has **no** `defaults/docs/` counterpart (it is the only tracked copy), so this is not the installed-copy-vs-canonical trap that `merge-pr.sh` would have been. `merge-pr.sh`, `champion-pr-merge.md`, `types.rs`, `ipc.rs` and the event taxonomy are all untouched, exactly as the Option 2 rationale claims. `Closes #4583` present.

**The shared core** — `build_and_narrate_completion` is a faithful extraction: the `already_narrated.contains` guard is re-checked inside it and `already_narrated.insert(key)` still happens only on a successful `build_completion_envelope`, so the pre-existing "a failed envelope build does not consume the dedup slot" behavior is preserved verbatim. `completion_for_exit` keeps its own early `contains` check, so the `SweepExited` path still short-circuits *before* spending a `gh` call — the refactor costs nothing on the hot path. Taking `duration_sec`/`exited_at` from the caller (rather than deriving them) is the right call and the doc comment explains why: the two paths genuinely have different clocks.

**Dedup, both directions** — the invariant holds whichever path observes a merge first. `reconcile_recent_merges_skips_a_merge_already_narrated_in_sweep` is the test I most wanted to see, and it asserts the *right* thing: both `gh pr list` invocations still ran (2 lines in the fake-gh log), so the skip is provably happening at the shared dedup check and not by the reconciliation pass accidentally never querying.

**Failure degradation** — `fetch_recent_merged_prs` is best-effort at every level: missing `gh`, nonzero exit, `MERGE_CHECK_TIMEOUT`, non-JSON body, non-array body, and per-row `filter_map` on `headRefName`/`mergedAt`/`number`/`url`. It cannot panic and cannot block the sink beyond the existing 10s timeout. `issue_from_branch` is a strict `feature/issue-<N>` prefix parse, so non-sweep PRs are dropped rather than misattributed. Missing/unparsable `createdAt` degrading to a zero-duration proxy rather than losing the whole row is the correct priority.

**`tokio::select!` correctness** — both branches are cancel-safe (`broadcast::Receiver::recv` and `Interval::tick` are both documented as such), so the `biased` recv-first arrangement cannot drop a bus event. Consuming the immediate first `interval` tick before the loop is a real bug avoided: without it a freshly-started sink would reconcile against an empty `known_workspaces` and race its own first live events. `MissedTickBehavior::Delay` correctly avoids catch-up bursts. Keeping one iteration to *either* an event *or* a tick keeps `completed` single-owner — the concurrency story is clean.

**Persistence** — atomic temp+rename with a pid-suffixed temp name, `create_dir_all` first, every failure logged at `debug` and swallowed. Persisting only after a *new* completion (`if !new_completions.is_empty()`) avoids a pointless rewrite on every idle tick.

**Lookback window** — dropping out-of-window rows *before* the dedup check (not after) is the subtle detail that matters, and the test asserts it explicitly: an out-of-window row is not recorded as narrated, so a later in-window observation can still emit. `reconcile_max_age` rejecting negatives and garbage back to the default (rather than degrading to a zero-length window that silently narrates nothing) is the right failure direction.

**`SafehouseTestPaths`** — genuinely load-bearing, not ceremony: without it the suite would read/write a real daemon's `~/.loom/safehouse-completed.json` and `~/.loom/workspaces.json`, and reconciliation targets would depend on whatever repos happen to be registered on the host. I checked every `run_sink`-driving test — all 6 `run_sink_*` tests plus `sink_send_rejected_survives_reconnect`, `absent_peer_degrades_without_blocking`, and `reconnects_after_mid_run_disconnect` hold the guard (10 call sites). `Drop` restores on panic. No gaps.

Making the fixtures `Utc::now()`-relative rather than literal dates is exactly right — with a lookback window, hard-coded dates would have silently turned these assertions into no-ops a week after they were written.

## Test Execution

1. `cargo fmt --check` in `loom-daemon` — ✅ Executed: clean
2. `cargo clippy --all-targets --all-features` — ✅ Executed: clean, zero warnings
3. `cargo test` in `loom-daemon` — ✅ Executed: exit 0, full suite green (2537 lib tests + integration + doc-tests)
4. Safehouse-scoped re-run (`cargo test --lib safehouse::`) — ✅ Executed: 90 passed / 0 failed, including all 8 new reconciliation + persistence tests and the pre-existing `run_sink_narrates_exit_ack_then_completion` / `run_sink_narrates_only_the_ack_when_nothing_merged` / `completion_for_exit_emits_at_most_once_per_merge` regressions, unmodified
5. Manual verification against a live daemon + `merge-pr.sh` with no active sweep — ⚠️ Not executed: requires a running `loom-daemon` with safehouse configured and a real approved PR. Correctly flagged as not-run in the PR description; the end-to-end `run_sink_reconciliation_narrates_a_champion_merge_with_no_live_sweep` test (no `SweepExited` ever published, workspace discovered from the registry alone, envelope asserted on the wire) is a good proxy.

**CI**: all checks pass (Rust Unit Tests 4m16s, Rust Linting & Build Check, AGENTS.md Sync, CLAUDE.md Line Budget). `mergeStateStatus` is `CLEAN`.

## AC coverage

| AC | Verdict |
|----|---------|
| Champion-tick merge (no live sweep) → exactly one completion with #4553 meta | ✅ unit + end-to-end on the wire |
| In-sweep merges still emit exactly once, no double-post | ✅ and asserted at the dedup check, not by query avoidance |
| Merge while daemon down picked up after restart | ✅ via the persisted set (round-trip, corruption, and no-repost-after-restart all covered) |
| Two quick merges for different issues emit independently | ✅ mixed already-narrated/fresh in one bulk response |

## Non-blocking

Filed #4649 (`loom:triage`) for the one thing the two bounds don't cover: on a host with an **empty** persisted dedup set (fresh install, the first upgrade to this version, a lost/unwritable completions file), the first tick can emit up to 30 completions at once for merges that all landed inside the 7-day window — on a repo at Loom's merge rate that is a one-time backlog dump onto the feed. Bounded, one-time, and far better than a silent feed, so not a merge blocker; a seed-only first pass would close it if the feed's consumers care.

Two smaller observations I am **not** filing, both explicitly acknowledged in the PR description or docs and both acceptable as-is:
- The reconciled `duration_sec` is PR-open time (`createdAt`→`mergedAt`), not agent runtime, so a champion-merged PR's feed line can read a duration in hours. There is no sweep clock to borrow, the proxy is documented in both the code and `safehouse.md`, and the alternative (omitting the duration) would need an envelope shape change.
- `reconciliation_targets` does a synchronous `WorkspaceRegistry::load_default()` read inside an async task every tick. It is a small file on a 5-minute cadence; not worth a `spawn_blocking`.

Documentation is unusually good — the docs rewrite covers both emit points, the exact query, the cadence, the round-robin, the window rationale, and the now-shared/now-persisted dedup contract, and it corrects a pre-existing reference to a `safehouse.reconcileInterval` config key that never existed. The "why" comments throughout the code (especially on `DEFAULT_RECONCILE_MAX_AGE_SECS` and `SafehouseTestPaths`) are the kind that survive contact with the next maintainer.

Nice work — approving.

Issue #4586, opened 2026-07-30
champion: PR rejection comments lack idempotency guard — 8+ duplicate 'Cannot Auto-Merge' comments posted on PR #4540 within 5 minutes


rjwalters:
## Problem

Champion's PR-rejection comment path (`champion-pr-merge.md`, "Transient failures - keep `loom:pr`, retry next tick") has no idempotency guard. When a PR repeatedly fails the same safety criterion across closely-spaced Champion evaluation cycles, it re-posts a fresh, near-identical rejection comment every time instead of recognizing the condition is unchanged.

Observed on PR #4540 (fix(daemon): distinguish PR-probe failure from verified no-open-PR (#4452)): between 2026-07-30T06:22:19Z and 2026-07-30T06:26:39Z (about 5 minutes), 8 separate "Champion: Cannot Auto-Merge" comments were posted, all citing the identical static reason: size check failing at 292 lines changed (+246/-46), exceeding the 200-line auto-merge limit, with no `loom:auto-merge-ok` override label present.

The PR's line count is a static property - it cannot change without a new push - so every one of these re-evaluations was guaranteed to fail the same way. Earlier in the same PR's history there's a similar burst around 04:25-04:41 for merge-conflict/CI-failure reasons, which did eventually resolve (Doctor fixed it), showing the pattern isn't limited to non-transient criteria; it's the retry cadence that's wrong even for criteria that could plausibly change.

This strongly suggests multiple concurrent Champion instances/ticks are evaluating the same PR in rapid succession - inconsistent with the documented 10-minute autonomous interval - burning API/comment-quota and degrading PR signal-to-noise for reviewers.

## Existing precedent in the same file

The stale-PR path (`champion-pr-merge.md`, "Stale PR (recency check failed)") already solves exactly this problem with an idempotent HTML-comment marker (`champion:stale-pr-notice`) that gates a single comment plus relabel. The generic "transient failure" rejection path has no equivalent guard.

## Suggested fix

1. Add the same idempotency-marker pattern used for stale PRs to the general transient-rejection comment path: before commenting, check the PR's existing comments for a marker tied to the specific failing criterion (for example `champion:reject:size-check`); skip commenting if the most recent comment for that criterion is unchanged from the current evaluation.
[...]

PR #4540, opened 2026-07-30
fix(daemon): distinguish PR-probe failure from verified no-open-PR (#4452)

rjwalters:
## Summary

`first_open_linked_pr` returned `Option<u32>` where `None` conflated a **verified** "no open linked PR" with **any probe failure** (missing/failed/timed-out `gh`, unresolvable repo, non-zero exit, unparseable output). That is fine for its original #4123 dispatch-guard consumer (fail-open = allow dispatch), but the #4366 no-progress predicate reused it with **inverted stakes**: `first_open_linked_pr(issue).is_none()` counts a clean self-skip *toward* quarantine. So a PARTIAL forge outage — PR probe fails while the issue probe still answers OPEN — could accrue wrongful quarantine pressure, even after the `issue_is_closed` half was hardened to a positive verdict.

This replaces the helper with `probe_open_linked_pr(&self, issue) -> OpenPrProbe`, a three-state enum that makes ignoring the failure case impossible at the call site:

```rust
enum OpenPrProbe {
    Open(u32),    // verified: at least one open linked PR
    NoneOpen,     // verified: forge answered, no open linked PR
    ProbeFailed,  // gh missing/failed/timed out, repo unresolvable, non-zero exit, unparseable output
}
```

`resolve_owner_repo()` failure, the `output_with_timeout(...)` spawn/timeout leg, a non-zero `gh` exit, and non-empty-but-unparseable stdout now all map to `ProbeFailed` instead of collapsing into a verified absence. An empty successful stdout remains a verified `NoneOpen`.

## Call sites (all three enumerated, per the issue's requirement)

1. **#4123 dispatch guard** (`dispatch()`, step 2.6): `Open(pr)` blocks (unless `resume_bypass_pr == Some(pr)`); `NoneOpen` **and** `ProbeFailed` proceed. **Fail-open behavior unchanged** (AC3).
2. **#4366 no-progress predicate** (reaper clean-exit branch): only a verified `NoneOpen` counts toward `no_progress`; `ProbeFailed` ⇒ `no_progress = false` (AC2). Mirrors the positive-verdict hardening already applied to the `issue_is_closed` half.
3. **#4256 crash-resume path** (`reap_once`): only `Open(pr)` is resume-eligible; `NoneOpen`/`ProbeFailed` fall through to ordinary handling. **Behavior unchanged.**

No public API or event-schema change; the helper stays private in `sweep_registry.rs`.

## Tests

[...]

Issue #4429, opened 2026-07-29
daemon: rate-limit-aware backoff and API budget telemetry for work-finder/role loops


rjwalters:
## Problem

During the 2026-07-29 GraphQL exhaustion, every host kept firing the full per-tick call pattern for ~25 minutes — 3 workspaces × (work-finder list + claim reconciliation + quarantine reconciliation) every 60s, each failing with `GraphQL: API rate limit already exceeded`. Role loops (champion/curator) also spawned `claude` sessions that immediately died against the same wall (spawn-claude exit 78). The daemon has no notion of 'the quota is gone until epoch T'.

## Proposal

- Detect rate-limit failures (REST 403/429 with `x-ratelimit-remaining: 0`, GraphQL RATE_LIMITED) and enter a **global cooldown until the advertised reset epoch** — one cheap probe near reset instead of the full tick fan-out.
- Suppress role-runner dispatch during cooldown (a spawned role session cannot do anything without API access).
- Surface budgets in `loom-daemon status`: core + GraphQL remaining/reset, plus last-hour consumption attributable to this host.
- Stagger/jitter tick phase per host so fleet members don't synchronize their polls.

## Acceptance

- Exhaustion window: near-zero API calls, no failed claude spawns, a single INFO line stating cooldown-until.
- `loom-daemon status` shows current API budget and reset time.
- Recovery within one tick of reset (observed today: recovery worked, but only by luck of tick timing).

---

### Context & architectural precedent

Part of epic #4432 (child 1 of 4 — the defensive leg). Siblings are **out of scope here**: #4428 (ETag/conditional-request caching), #4430 (GitHub App identity), #4431 (safehouse claim coordination).

The in-repo precedent is `loom-daemon/src/host_breaker.rs` (#4235): a stateful, daemon-global **dispatch suppressor** with a pure `step()` decision function over an immutable runtime + observation + `now`, consulted in the work-finder tick alongside the main-health halt flag (consult sites at `work_finder.rs:1374–1387` and `work_finder.rs:1672–1694`), with transitions emitted on the event bus. Build the rate-limit cooldown as a **sibling suppressor of the same shape** — not a parallel ad-hoc mechanism. It must never kill running sweeps; it suppresses new dispatch and polling only.

Two deliberate differences from the host breaker:
[...]

Issue #4428, opened 2026-07-29
daemon: route hot GitHub polls through REST conditional requests (ETags — 304s are rate-limit-free)


rjwalters:
## Problem

2026-07-29 incident: the entire fleet (robb-studio, robb-pro, loom-worker-1) shares one authenticated identity, and its 5,000/hr **GraphQL** budget was exhausted (user ID 2687775). Every work-finder tick then failed for ~25 minutes across all hosts:

- `work_finder` runs `gh issue list --label loom:issue` per workspace per 60s tick (GraphQL)
- `claim_reconciliation` (`--label loom:building`) and `quarantine_reconciliation` (`--label loom:blocked`) add more GraphQL list calls per workspace
- 3 hosts × up to 11 registered workspaces × several calls/min = steady-state burn even when nothing changes

GraphQL has **no conditional-request mechanism**. The REST issues API does: a `GET /repos/{owner}/{repo}/issues?labels=...` with `If-None-Match` that returns **304 costs zero rate limit**. Issue lists change rarely relative to a 60s poll, so the vast majority of today's burned quota was re-fetching unchanged data.

## Proposal

- Swap the daemon's hot list paths (work-finder, claim/quarantine reconciliation) from `gh issue list` to REST with a per-host ETag cache (extend `.loom/scripts/gh-cached` or implement natively in loom-daemon).
- Cache keyed by (repo, label, state); store ETag + body; send `If-None-Match`; on 304 serve cached body.

## Acceptance

- Steady-state fleet with no issue churn consumes ~0 core rate limit for polling.
- A label change is picked up within one tick (ETag miss → fresh body).
- `loom-daemon status` (or debug log) reports cache hit rate.

---

### Verified context

Hot-path call sites verified on `main` (all currently GraphQL via `gh issue list` / `gh pr list`):

- `loom-daemon/src/work_finder.rs` — `GhWorkSource::list_ready_issues` (~line 1987): `gh issue list --label loom:issue --state open --limit 200 --json number,labels,createdAt`, per registered workspace per tick (default 60s). The dominant steady-state burner.
- `loom-daemon/src/claim_reconciliation.rs` — forge glue (~line 741): issue-side `loom:building` list; (~line 930): PR-side list for `loom:reviewing`/`loom:treating` claims. Startup + periodic (default 10 min) per workspace, plus per-candidate view/timeline probes.
[...]

PR #4648, opened 2026-07-30
feat(daemon): cross-host collision detection for role-runner ticks (#4623)

### LINKED ISSUE #4623: daemon: role runner has no cross-host collision detection for Champion/Curator/Judge/Auditor/Guide ticks
## Problem

Investigating #4586 (Champion PR-rejection comments lack an idempotency guard —
8 duplicate "Cannot Auto-Merge" comments posted on PR #4540 within ~5 minutes),
the burst itself is best explained by **multiple independent Champion
invocations evaluating the same PR within a sub-10-minute window**, not by a
single role-runner loop re-firing too fast. #4586 fixes the *symptom*
(duplicate comments) with a per-criterion idempotency marker; this issue is
about the *cause* — why Champion is apparently being invoked far more often
than the documented ~10-minute cadence.

## What the code review found

- `loom-daemon/src/role_runner.rs` already has a solid **in-process** overlap
  guard: `RoleRunGuard` / `InProgressGuard` (`Arc<Mutex<HashSet<(root, role)>>>`)
  ensures a given `(repo root, role)` pair can only have one invocation running
  at a time within a *single* `loom-daemon` process — the interval loop and the
  `onIdle` loop share it, and there's test coverage
  (`test_role_run_guard_blocks_second_acquire_then_releases_on_drop`, etc.).
  This repo's `.loom/config.json` doesn't set `onIdle` for champion either, so
  the idle-edge path isn't in play here.
- The daemon has a **host-local** single-instance guard too: `loom-daemon-start.sh`
  refuses a second start when `.loom/.daemon.pid` points at a live process.
- **Neither guard is cross-host.** If two independent `loom-daemon` processes
  (e.g. two different hosts, or two clones/worktrees on the same host each with
  their own PID file) both have `autonomous.roleRunner.enabled=true` with
  `champion` in `roles` against the *same* forge-tracked repo, each process's
  in-memory guard only knows about its own invocations. GitHub (the forge) is
  the only shared state, and nothing currently coordinates Champion role-runner
  ticks across processes/hosts.
- Confirms this is a known category of gap: `autonomous.collisionDetection`
  (#4085) already exists as a **detection-only, opt-in** baseline for exactly
  this kind of cross-host duplicate dispatch — but it is wired into **sweep
  dispatch** (`dispatch_sweep`) only, not into the standalone role runner that
  drives Champion/Curator/Judge/Auditor/Guide cron-style ticks.
- GitHub Actions cron for Champion is disabled in this repo (`schedule:` is
  commented out in `.github/workflows/loom-champion.yml`, only
  `workflow_dispatch` remains active), so a GH-Actions-vs-daemon double-fire is
  unlikely to be the specific cause here, but is a related instance of the same
  general problem for repos that do enable it alongside the daemon role runner.

## Suggested follow-up

Extend cross-invocation collision awareness to the role runner, mirroring
`autonomous.collisionDetection` (#4085):

- At minimum, detection/logging when a role-runner tick for `(repo, role)`
  starts while forge-visible evidence suggests another Champion pass is
  concurrently active (e.g. a very-recently-posted Champion comment from this
  tick's own PR list).
- Consider whether the role runner should honor `autonomous.collisionDetection.enabled`
  directly, or needs its own dedicated toggle given it drives a different
  invocation shape (periodic role prompt vs. per-sweep dispatch).

This is the same underlying problem class as #4570 (concurrent Judge
invocations racing over shared forge state) and #4574/#4556, just surfacing at
the process/host level for the role runner instead of the per-PR label-write
level.

## Related

- #4586 — Champion PR rejection comments lack idempotency guard (the symptom;
  fixes duplicate comments regardless of root cause)
- #4570 — concurrent-Judge verdict race (closed) — precedent for
  concurrent-role races over shared forge state
- #4085 — cross-host dispatch-collision detection baseline (sweep dispatch
  only; this issue proposes extending the same idea to the role runner)

rjwalters:
Closes #4623

## Problem

The role runner's overlap guards both stop at the process boundary:

- **In-process** — `RoleRunGuard`/`InProgressGuard` (`Arc<Mutex<HashSet<(root, role)>>>`) stops a `(repo root, role)` pair running twice in *one* `loom-daemon` process.
- **Host-local** — `loom-daemon-start.sh` refuses a second start while `.loom/.daemon.pid` is live.

Two independent daemons (two hosts, or two clones on one host each with their own PID file) pointed at the same forge repo with `autonomous.roleRunner.enabled=true` therefore run Champion/Curator/Judge/Auditor/Guide with **zero mutual awareness** — the leading explanation for #4586's eight duplicate "Cannot Auto-Merge" Champion comments on PR #4540 inside ~5 minutes, far above the documented 10-minute cadence. A repo that leaves the `loom-*.yml` cron schedules enabled alongside the daemon role runner has the same shape.

`autonomous.collisionDetection` (#4085) already established the detection-only baseline for exactly this problem class — but only for **sweep dispatch**. This PR adds the role-runner half.

## Approach

New `loom-daemon/src/role_collision.rs`. Immediately before a tick for `(root, role)`:

1. List that role's own **label-defined queue** over the ETag-cached REST listing (#4428) — an unchanged queue is a free `304`, and it never touches the GraphQL budget #4429 exhausted.
2. Take the newest `updated_at`.
3. Compare against the wall-clock window of **this process's own last completed tick**. Activity strictly after our own pass finished (plus a 90s clock-skew margin) was not written by us.

| Role | Probed queue |
|------|--------------|
| `champion` | open PRs `loom:pr` |
| `judge` | open PRs `loom:review-requested` |
| `curator` | open issues `loom:curating` |
| `guide` | open issues `loom:triage` |
| `auditor` | open issues `loom:auditor` |

Taking `max(now - window, self_run_end + skew)` as the cutoff is what makes a single stored self-run window sufficient — a window longer than the tick interval can span several of our own past passes, but anything older than the latest one is already excluded by construction.

Classification mirrors #4085's three-way split: **collision** (`warn` with role, repo, host identity, newest item + timestamp, lookback window, running count), **clean**, **unknown**. Fail-closed: no self-run baseline yet (first tick after daemon start) and any listing failure are never counted, so the baseline is never inflated.

**Detection only** — nothing suppresses, delays, or reorders an invocation. Enforcement remains #4028 Phase 2's cross-host CAS.

## Config surface (the question the issue left open)

The issue asked whether to honor `autonomous.collisionDetection.enabled` directly or add a dedicated toggle. **Both**: the shared toggle is inherited (same problem class, one operator switch), with a role-runner-specific override on top, because the two probe shapes have genuinely different costs and cadences (one `gh issue view` *per dispatch* vs. one cached REST listing *per role per tick*) and an operator must be able to run one without the other.

```
LOOM_ROLE_RUNNER_DETECT_COLLISIONS
  > autonomous.roleRunner.collisionDetection
  > LOOM_DETECT_COLLISIONS / autonomous.collisionDetection.enabled   (#4085)
  > off
```

Window: `LOOM_ROLE_RUNNER_COLLISION_WINDOW_SECS` > `autonomous.roleRunner.collisionWindowSecs` > that role's tick interval, clamped to `[60s, 3600s]`. Enablement is resolved per registered root (#4377); the daemon logs its own workspace's resolution once at startup.

## Honest limits (documented in the module and the `warn` line)

Biased to **under**-count: a peer pass landing inside our own tick's window is invisible, and a pass that changes nothing writes nothing. Bounded **over**-count: a human, a Builder, or a *sweep-internal* Judge/Doctor/Champion on the same queue is indistinguishable here. Since nothing acts on the signal, a false positive costs one log line.

## Wiring

All three invocation paths (`spawn_role_task`, `spawn_multi_role_task`, `observe_and_fire_idle`) go through `invoke_with_collision_probe`, inside their existing `spawn_blocking` so the probe never parks a runtime worker. Probe-first ordering is load-bearing (it reads the baseline our previous run left, which `record_run_started` then clears). Disabled is a no-op with **no forge call**.

## Testing

33 new tests; `cargo test --package loom-daemon` green (2561 lib + all integration), `cargo clippy -- -D warnings` and `cargo fmt --all --check` clean.

- Pure classifier: self-run attribution, skew margin, lookback bound, target-kind filtering (a REST issue listing returns PRs too), newest-wins + touched count, missing `updated_at`, still-running self pass.
- Tracker: window open/close, `(root, role)` keying, finish-without-start, monotonic count.
- Config precedence in both directions, including an explicit env `0` overriding a `true` config and the shared-#4085 fallthrough.
- Probe wiring against a fake `QueueSource` (right label queried, failure ⇒ uncounted `Unknown`, queueless role makes no forge call, collision increments the total).
- Production `RestQueueSource` end-to-end against a fake `gh` emitting the real `--include` HTTP shape — an unparseable `updated_at` or a missed `pull_request` marker would silently disable detection for every PR-queue role.
- Role-runner wrapper: transparent to the invocation (same role/prompt/outcome), and records the self-run window even on a failing tick.

## Docs

`defaults/docs/daemon-reference.md`: new "Cross-host role-tick collision detection (#4623)" section next to #4085's, plus rows in both `autonomous.*` config tables. (`.loom/docs/daemon-reference.md` is a symlink — no drift.)

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2286,6 +2286,8 @@ exactly like `main_health_gate::read_build_gate_config`.
 | `autonomous.roleRunner.intervalSecs` | `LOOM_ROLE_RUNNER_INTERVAL_SECS` | per-role built-in (5–15 min) | Uniform override applied to every enabled role's cadence |
 | `autonomous.roleRunner.model` | *(config only)* | `sonnet` | Model every role child is pinned to via `--model` (#4501). Resolved through the same `resolve_dispatch_model` chain as sweep dispatch: this key > `autonomous.model` > shipped default; blanks treated as unset. A role child never inherits the account's interactive CLI default |
 | `autonomous.roleRunner.onIdle` | *(config only)* | `[]` (none) | Subset of the same 5 roles to also fire on the work-finder **idle edge** (#4364) — the non-idle → idle transition (0 in-flight sweeps AND nothing dispatched this tick), in addition to the interval cadence. Absent → none (opposite default from `roles`); unknown names ignored with a warning. Debounced to min 60s per (root, role) and skipped while that role's interval/idle run is in progress. **Requires the work finder enabled** to observe idleness (a startup warning fires if set with the work finder off). **Also gated by that same root's own `enabled`** (#4377) — see below |
+| `autonomous.roleRunner.collisionDetection` | `LOOM_ROLE_RUNNER_DETECT_COLLISIONS` | inherits `autonomous.collisionDetection.enabled`, else `false` | Cross-host role-tick collision baseline (#4623). Detection only — a pre-tick probe of that role's own label queue, logged/counted, never acted on. Absent → falls through to #4085's shared toggle; see [Cross-host role-tick collision detection](#cross-host-role-tick-collision-detection-4623) |
+| `autonomous.roleRunner.collisionWindowSecs` | `LOOM_ROLE_RUNNER_COLLISION_WINDOW_SECS` | that role's tick interval | Lookback window for the #4623 probe, clamped to `[60, 3600]`. Zero/invalid dropped to the next tier |
 | `autonomous.idleExit.enabled` | `LOOM_AUTONOMOUS_IDLE_EXIT_ENABLED` | `false` | End the daemon cleanly after the idle window so a host guard can take over. Independent of Work Finder; never invokes a power command |
 | `autonomous.idleExit.idleMinutes` | `LOOM_AUTONOMOUS_IDLE_EXIT_MINUTES` | `60` | Continuous idle/starvation window. Zero/invalid → default |
 | `autonomous.idleExit.onTokenStarvation` | `LOOM_AUTONOMOUS_IDLE_EXIT_ON_TOKEN_STARVATION` | `true` | Also exit after zero healthy accounts for the full window with no sweep in flight, even if roles keep cycling |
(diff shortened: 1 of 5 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T17:36:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T17:36:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

## Review summary

Reviewed against issue #4623 (role runner has no cross-host collision detection for Champion/Curator/Judge/Auditor/Guide ticks).

**Correctness / design**
- `loom-daemon/src/role_collision.rs` implements a clean, pure `classify_queue_activity` function (no I/O, no clock) with a well-reasoned cutoff: `max(now - window, self_run_end + skew)`. Verified the skew/window/self-run-window interplay by hand against the doc comments and it holds — a still-running self pass (`ended: None`) correctly suppresses everything, and the "newest wins, touched counts all" logic is straightforward.
- Fail-closed throughout: no self-run baseline yet, a listing failure, or a queueless role all resolve to `Unknown` and are never counted — matches the stated design goal of never inflating the baseline with a false positive.
- Detection-only: `probe_before_tick`/`invoke_with_collision_probe` never gate or delay the actual role invocation — confirmed by reading the wiring in all three call sites (`spawn_role_task`, `spawn_multi_role_task`, `observe_and_fire_idle`), all inside their existing `spawn_blocking`.
- Config precedence (`LOOM_ROLE_RUNNER_DETECT_COLLISIONS` > `autonomous.roleRunner.collisionDetection` > #4085's shared toggle > off) directly answers the open question the issue posed, with both env and config paths tested in both directions.
- Probe target-to-label mapping (champion to loom:pr, judge to loom:review-requested, curator to loom:curating, guide to loom:triage, auditor to loom:auditor) matches the CLAUDE.md label lifecycle exactly.
- Disabled path (the default) costs one config read and zero forge calls, confirmed via probe_for_a_queueless_role_makes_no_forge_call and the early-return in probe_before_tick.

**Testing**
- Ran cargo test -p loom-daemon --lib role_collision (31 tests) and cargo test -p loom-daemon --lib role_runner (73 tests) locally — all pass.
- cargo clippy -p loom-daemon --all-targets -- -D warnings — clean.
- cargo fmt --all --check — clean.
- Test coverage is thorough: pure classifier edge cases (skew boundary, lookback boundary, wrong target kind, missing updated_at), tracker semantics (keying, restart, finish-without-start), config precedence both directions, and a production RestQueueSource test against a fake gh exercising the real --include HTTP shape — good catch that this is the one place an unparseable updated_at or missed pull_request marker would silently break detection.

**Docs**
- defaults/docs/daemon-reference.md gets a new section plus config-table rows; .loom/docs/daemon-reference.md is a symlink so no drift.

**CI / merge state**
- All CI checks green (gh pr checks 4648): Rust Unit Tests, Rust Linting & Build Check, Shell Test Suites, Installer Integration Tests, Codex Adapter Smoke, AGENTS.md Sync, CLAUDE.md Line Budget, Detect Changes.
- mergeStateStatus: CLEAN, mergeable: MERGEABLE.

No blocking issues found. This is a solid, opt-in, fail-closed extension of #4085's collision-detection pattern to the role runner, with no behavior change for the default (disabled) path.

PR #4655, opened 2026-07-30
fix(guide,curator): gate loom:blocked unblock on superseding PR block

### LINKED ISSUE #4634: Guide/Curator dependency-unblock logic ignores a superseding block reason (loom:blocked flip-flop on #4492)
## Problem

The dependency-unblock logic used across this repo's autonomous roles checks only the literal "Blocked by #N" / "Dependencies" text in an issue body, and treats the block as resolved once that referenced issue/PR closes. It does not check whether the *actual, current* reason `loom:blocked` was applied still holds — which can differ from the body's stated dependency once the issue has moved through further lifecycle stages (e.g. an implementation PR later hits the Doctor-cycle cap).

This caused a live flip-flop loop on issue #4492 today (2026-07-30):

1. 02:45 UTC — Champion parks #4492 pending dependency #4491.
2. 08:01 UTC — Sweep coordination marks #4492 `loom:blocked` for a *different* reason: its implementation PR #4519 exhausted the Doctor-cycle cap (3 Judge rejections) and needs human review.
3. 08:27, 11:01, 13:55 UTC — three separate Curator passes each remove `loom:blocked` from #4492, citing "dependency #4491 resolved" — true, but not the reason the label was applied at step 2. Each time, PR #4519 was still open with `loom:changes-requested`.
4. 10:26, 13:48, and (via this session) 13:58 UTC — Champion has to manually re-detect the wrong unblock and restore `loom:blocked` each time, quoting PR #4519's live state.

Because dependency #4491 will never re-open, this check will report "resolved" forever, and this will keep recurring on every pass over #4492 until PR #4519 merges or closes — burning cycles and risking a window where a second Builder claims #4492 while #4519's worktree/branch is still active (the duplicate-dispatch class in #4556).

## Root cause (corrected — two locations, not one)

This issue was previously curated with an inaccurate single-location attribution ("Curator's dependency re-verification logic"), which Champion review correctly flagged (see comment history) as pointing a Builder at the wrong file. Direct code inspection on 2026-07-30 found **two** distinct locations that exhibit the identical class of bug — both need the same fix, since a Curator pass could be following either:

1. **`defaults/roles/guide.md` `check_and_unblock()`, lines 550-591** (identical mirror at `defaults/.claude/commands/loom/guide.md:550-591`) — parses dependencies via `parse_dependencies()` (lines 517-527), and once every parsed `#N` is `CLOSED`, unconditionally clears `loom:blocked` (lines 575-588) with no check for a later, superseding block reason.
2. **`defaults/roles/curator.md` "When Dependencies Complete" step, line 697** (identical mirror at `.loom/roles/curator.md:697`, the installed copy) — instructs: "GitHub automatically checks boxes when issues close. When you see all boxes checked: ... Remove `loom:blocked` label and add `loom:curated`" — the same unconditional strip-on-checklist-complete pattern, with no linked-PR or comment-history check either. This explains why the live bad-unblock comments on #4492 were signed "🔓 Curator" rather than Guide: a Curator pass following its own role file's literal instruction produces exactly this failure mode.

Neither location has any way to detect that a *later* block reason (a linked implementation PR hitting the Doctor-cycle cap, or any other post-hoc block) superseded the original dependency-based block.

## Suggested fix

Apply the same guard to both locations. **Primary check (preferred, mechanical/testable):** before removing `loom:blocked` based on a body-declared dependency closing, check whether any PR referencing this issue (via `Closes #N` / `closingIssuesReferences`, e.g. `gh issue view <N> --json closedByPullRequestsReferences` or equivalent) is still open with `loom:changes-requested` or `loom:blocked` — if so, leave `loom:blocked` alone regardless of the body's Dependencies section.

**Secondary heuristic (fragile, optional defense-in-depth):** scan the issue's comment history for the most recent `loom:blocked` justification (e.g. "doctor cycle exhausted", "Sweep coordination: blocking", "Champion: re-blocking") and confirm that specific condition has cleared, not just the body's stated dependency.

Both `defaults/roles/guide.md`'s `check_and_unblock()` and its `.claude/commands/loom` mirror, and `defaults/roles/curator.md`'s "When Dependencies Complete" step and its `.loom/roles/curator.md` installed mirror, must receive the identical change to stay in sync (per this repo's existing defaults/installed-copy resync convention).

## Affected Files

- `defaults/roles/guide.md` (lines 550-591, `check_and_unblock()`; lines 517-527, `parse_dependencies()`) — add linked-PR gate before unconditional unblock
- `defaults/.claude/commands/loom/guide.md` (identical lines 550-591 / 517-527) — mirror the same change
- `defaults/roles/curator.md` (line 697, "When Dependencies Complete" step 2) — add the same linked-PR gate before instructing removal of `loom:blocked`
- `.loom/roles/curator.md` (installed copy, identical line 697) — mirror the same change, or rely on `./.loom/scripts/resync-installed.sh` if this repo's resync convention already covers `.loom/roles/*` from `defaults/roles/*`

## Test Plan

- [ ] Manual verification: reconstruct #4492's exact sequence against a synthetic/test issue — body has a closed "Depends on #N" dependency, but a linked implementation PR is still open with `loom:changes-requested`. Confirm the fixed check leaves `loom:blocked` in place and does not post an "Unblocked" comment.
- [ ] Manual verification (control case): same synthetic issue, but the linked PR is merged/closed. Confirm the fixed check does unblock as before (no regression on the happy path).
- [ ] Automated tests: if `guide.md`'s bash functions have any existing test harness/fixture (check `defaults/scripts/tests/` or similar), add a fixture covering both the flip-flop and control cases above. If no test harness exists for role-prompt bash snippets, document that this is verified manually per the above and note the gap.
- [ ] Edge cases: an issue with a closed body dependency AND no comment history at all (should still check linked-PR state, not just skip); an issue where the linked PR is open but has no review-state label yet (decide default: keep `loom:blocked` in this decide state, or gate only when `closingIssuesReferences` are non-empty).

## Acceptance Criteria

- [ ] `check_and_unblock()` in `defaults/roles/guide.md` (and its `.claude/commands/loom` mirror) does not remove `loom:blocked` while a linked implementation PR is still open and not merged.
- [ ] `defaults/roles/curator.md`'s "When Dependencies Complete" step (and its `.loom/roles/curator.md` mirror) carries the identical guard, not just a text note — both files must actually receive the change, not only one.
- [ ] Regression scenario covering issue #4492's exact sequence (dependency resolved in body, but PR-level block still active) is documented and verifiable per the Test Plan above.
- [ ] The secondary comment-history heuristic, if implemented, is clearly marked as a fallback and does not override the primary linked-PR gate.

## Related

- #4492 — the issue that surfaced this bug (see comment thread for full timeline). #4492 also cross-references this issue (#4634) directly in its own thread — already reconciled: the same corrected two-location root cause applies there, no separate action needed on #4492 itself beyond what's already tracked.
- #4519 — the implementation PR whose Doctor-cycle cap is the real, current block reason
- #4556 — duplicate-dispatch failure class this bug risks triggering
- #4623 — related but distinct: cross-host/cross-invocation collision detection for role-runner ticks. That issue is about *concurrent* invocations racing; this one is about a *single* pass reaching the wrong conclusion even run in isolation, due to flawed reasoning (checking a stale dependency instead of the actual current block reason).

rjwalters:
## Summary

`check_and_unblock()` in `guide.md` and the "When Dependencies Complete" step
in `curator.md` both strip `loom:blocked` as soon as every body-declared
dependency closes, with no check for a *later, current* block reason. This
caused a live flip-flop loop on #4492 (2026-07-30): the body dependency
(#4491) closed, but `loom:blocked` was later re-applied for a completely
different reason — implementation PR #4519 hit the Doctor-cycle cap and
needed human review. Three separate Curator passes each "resolved" the stale
dependency and stripped the label while PR #4519 sat open with
`loom:changes-requested`, forcing Champion to keep manually re-blocking with
the real, current reason each time.

Since both `defaults/roles/{guide,curator}.md` and `.loom/roles/{guide,curator}.md`
are now symlinks that resolve to the single canonical
`defaults/.claude/commands/loom/{guide,curator}.md`, editing those two files
is sufficient — every other path in the issue's "Affected Files" list already
resolves to the same file (verified with `realpath`), so there is nothing
left to keep in sync manually.

## Changes

- **`defaults/.claude/commands/loom/guide.md`**: added `has_superseding_block()`
  — a primary, mechanical gate that checks whether any PR referencing the
  issue (`closedByPullRequestsReferences`) is still `OPEN` and carries
  `loom:changes-requested` or `loom:blocked`. `check_and_unblock()` now calls
  this gate before acting on a fully-resolved body dependency, and skips
  (leaving `loom:blocked` untouched, no "Unblocked" comment) when it fires.
  Also documented an optional secondary comment-history heuristic that never
  overrides the primary gate, updated the worked example, and added a bullet
  to "When NOT to Unblock".
- **`defaults/.claude/commands/loom/curator.md`**: the "When Dependencies
  Complete" step now runs the identical primary check (via
  `gh issue view --json closedByPullRequestsReferences` + `gh pr view --json
  state,labels`) before removing `loom:blocked` and adding `loom:curated`,
  with the same secondary-heuristic caveat. "Decision Logic" step 2 now
  points at this gate instead of unconditionally declaring "all checked" safe.

## Test plan

- [x] Reconstructed #4492's exact sequence as a mocked scenario (body
  dependency closed, linked PR `OPEN` + `loom:changes-requested`) —
  `has_superseding_block` returns `true`, so the unblock is skipped.
- [x] Control case: linked PR `MERGED` — `has_superseding_block` returns
  `false`, unblock proceeds as before (no regression on the happy path).
- [x] Edge case: linked PR `OPEN` with no review-state label yet —
  conservatively treated as *not* superseding (documented explicitly in both
  files), matching the issue's suggested default.
- [x] `bash -n` + `shellcheck -x` on the extracted `has_superseding_block()`
  snippet — no new issues beyond the pre-existing `SC2155` style already
  present in the surrounding file.
- [x] Ran `.loom/scripts/tests/test-dependency-parse.sh` — unaffected
  (`parse_dependencies()` itself is untouched); its 4 pre-existing failures
  are a path-resolution bug in the test file itself, reproduced identically
  on `main` before this change, out of scope here.
- No automated test harness exists for these role-prompt bash snippets
  (confirmed no fixture directory covers `guide.md`/`curator.md` logic beyond
  `test-dependency-parse.sh`'s `parse_dependencies` pin), so verification is
  the manual/mocked reconstruction above per the issue's Test Plan.

Closes #4634

### DIFF
--- a/defaults/.claude/commands/loom/curator.md
+++ b/defaults/.claude/commands/loom/curator.md
@@ -667,7 +667,9 @@ This issue cannot proceed until dependencies are complete.
 
 **If Dependencies section exists:**
 1. Check if all task list boxes are checked (✅)
-2. **All checked** → Safe to mark `loom:curated`
+2. **All checked** → Also check for a superseding block first (see "When
+   Dependencies Complete" below) — only if that check clears too is it safe
+   to mark `loom:curated`
 3. **Any unchecked** → Add/keep `loom:blocked` label, do NOT mark `loom:curated`
 
 **If NO Dependencies section:**
@@ -692,7 +694,41 @@ gh issue edit <number> --add-label "loom:blocked"
 
 ### When Dependencies Complete
 
-GitHub automatically checks boxes when issues close. When you see all boxes checked:
+GitHub automatically checks boxes when issues close. **Before acting on all-boxes-checked, first check for a superseding block reason (#4634)** —
+`loom:blocked` can get re-applied later for a reason that has nothing to do
+with the body's original Dependencies section (e.g. this issue's own
+implementation PR later hit the Doctor-cycle cap and needs human review). A
+body dependency closing does NOT mean the label's *current* justification has
+cleared, and blindly trusting it caused a live flip-flop loop on #4492: three
+separate Curator passes each stripped `loom:blocked` citing "dependency
+resolved" while the real, current block (an open PR with
+`loom:changes-requested`) was still active, forcing Champion to keep manually
+re-blocking with the real reason each time.
+
+**Primary check (preferred, mechanical/testable) — run this first:**
+```bash
+# Any PR that would close this issue, still OPEN and carrying
+# loom:changes-requested or loom:blocked, is a superseding CURRENT block
+# reason — regardless of what the body's Dependencies section says.
+gh issue view <number> --json closedByPullRequestsReferences \
+  --jq '.closedByPullRequestsReferences[].number'
+# For each PR number returned:
+gh pr view <pr_number> --json state,labels
+# state == "OPEN" AND labels include loom:changes-requested or loom:blocked
+#   → a superseding block is active. Leave loom:blocked in place, do NOT mark
+#     loom:curated, and do NOT post an "unblocked"/"dependencies resolved"
+#     comment — even though the body's checklist is fully checked.
+```
+
+**Secondary heuristic (fragile, optional defense-in-depth, does NOT override
+the primary check above):** if there is no linked PR at all, scan recent
+issue comments for the most recent explicit `loom:blocked` justification
+(e.g. "doctor cycle exhausted", "Sweep coordination: blocking", "Champion:
+re-blocking") and confirm that specific condition has since cleared — not
+just that the body's stated dependency closed. When in doubt, leave
+`loom:blocked` in place.
+
+**Only once the superseding-block check clears**, proceed:
 1. Claim the issue if not already claimed: `gh issue edit <number> --add-label "loom:curating"`
 2. Remove `loom:blocked` label and add `loom:curated`: `gh issue edit <number> --remove-label "loom:blocked" --remove-label "loom:curating" --add-label "loom:curated"`
 3. Issue awaits `loom:issue` promotion (human, Champion, or a `/loom:sweep` orchestrator) before Workers can claim
--- a/defaults/.claude/commands/loom/guide.md
+++ b/defaults/.claude/commands/loom/guide.md
@@ -544,6 +544,61 @@ was_previously_approved() {
 }
 ```
 
+### Superseding Block Check (#4634)
+
+A body-declared "Depends on #N" can go stale once the issue moves further
+through the lifecycle after it was written — e.g. `loom:blocked` gets
+re-applied later for a completely different, *current* reason (an
+implementation PR hitting the Doctor-cycle cap, a fresh block comment) while
+the original body dependency has long since closed. Trusting only the body
+dependency caused a live flip-flop loop on #4492: three separate Curator
+passes each stripped `loom:blocked` citing "dependency #4491 resolved" —
+true, but not the reason the label was applied — while implementation PR
+#4519 sat open with `loom:changes-requested`, forcing Champion to keep
+manually re-blocking with the real, current reason each time.
+
+**Before unblocking on a resolved body dependency, always run this check
+first.** It is the primary, mechanical gate — not a heuristic — and it
+overrides a fully-resolved body dependency:
+
+```bash
+has_superseding_block() {
+  local number="$1"
+  # Any PR that would close this issue (Closes #N / closingIssuesReferences)
+  # still OPEN and carrying loom:changes-requested or loom:blocked is a
+  # superseding, CURRENT block reason — regardless of what the body's
+  # Dependencies section says. An open PR with no review-state label yet
+  # (still being built) is conservatively treated as NOT superseding here;
+  # `check_and_unblock` still applies its own gates afterward.
+  local pr_numbers
+  pr_numbers=$(gh issue view "$number" --json closedByPullRequestsReferences \
+    --jq '.closedByPullRequestsReferences[].number' 2>/dev/null)
+
+  for pr in $pr_numbers; do
+    local pr_json
+    pr_json=$(gh pr view "$pr" --json state,labels 2>/dev/null) || continue
+    local pr_state=$(echo "$pr_json" | jq -r '.state')
+    local pr_blocked=$(echo "$pr_json" | jq -r \
+      '[.labels[].name] | any(. == "loom:changes-requested" or . == "loom:blocked")')
+    if [ "$pr_state" = "OPEN" ] && [ "$pr_blocked" = "true" ]; then
+      echo "true"
+      return
+    fi
+  done
+
+  echo "false"
+}
+```
+
+**Secondary heuristic (fragile, optional defense-in-depth, does NOT override
+the primary gate above):** if the primary check found no linked PR at all,
+scan recent comments for the most recent explicit `loom:blocked`
+justification (e.g. "doctor cycle exhausted", "Sweep coordination: blocking",
+"Champion: re-blocking") and confirm that specific condition has since
+cleared — not just that the body's stated dependency closed. Treat this as a
+soft signal only; when in doubt, leave `loom:blocked` in place for a human or
+a later pass to sort out.
+
 ### Unblocking Logic
 
 ```bash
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T17:53:03Z) — standing down without reclaiming. Not stomping.

rjwalters:
**Approved.**

Reviewed the diff against issue #4634's acceptance criteria and independently verified the Builder's claims:

**Symlink-consolidation claim — verified independently.** Traced the git-tracked symlink chain by hand (`git ls-files -s` / `git cat-file -p`, not just `readlink -f` which can be fooled by local install artifacts):
- `.loom/roles` is a tracked symlink → `../defaults/roles` (whole-dir symlink)
- `defaults/roles/{guide,curator}.md` are tracked per-file symlinks → `../.claude/commands/loom/{guide,curator}.md`, which resolves (relative to `defaults/roles/`) to `defaults/.claude/commands/loom/{guide,curator}.md` — a real tracked blob, entirely within `defaults/`
- `.claude/commands/loom/{guide,curator}.md` (top-level, *not* under `defaults/`) is **not tracked in git at all** in this repo — it only exists locally via the gitignored dogfood symlink `.claude/commands/loom -> ../../defaults/.claude/commands/loom` (see `.gitignore` lines ~114-132)

So the two files this PR edited (`defaults/.claude/commands/loom/{guide,curator}.md`) are indeed the sole canonical source for every path the issue's original "Affected Files" list named. Editing only these two is correct and sufficient — the issue's file list is stale relative to the #3682 symlink consolidation.

**Logic review — `has_superseding_block()` (guide.md) / inline equivalent (curator.md):**
- Correctly distinguishes the target scenario: a resolved body dependency is treated as necessary-but-not-sufficient; the function only returns `true` (stay blocked) when a PR from `closedByPullRequestsReferences` is both `OPEN` **and** carries `loom:changes-requested`/`loom:blocked`. A merged PR (control case) or a fully resolved dependency with no open/blocked PR correctly returns `false` and unblocks as before — no regression on the happy path.
- No loop/recursion risk — single pass over a bounded PR list per issue, `return`/`continue` used correctly inside the `while read` loop in `check_and_unblock()`.
- Reproduced the #4492 scenario logically end-to-end: body dep closed + linked PR OPEN/`loom:changes-requested` → gate fires → `loom:blocked` stays, no "Unblocked" comment posted. Matches the issue's Test Plan item 1 exactly.
- Minor, non-blocking observations (not required for approval, both explicitly documented as intentional trade-offs in the PR itself, so no surprise to a future reader):
  - An open PR with no review-state label yet (mid-build) is treated as *not* superseding, so `check_and_unblock` could still restore `loom:issue` while a builder's PR is actively open — a narrow residual overlap with the #4556 duplicate-dispatch class the parent issue calls out. This exact case was left as an open design question in the issue ("decide default") and the PR's choice is defensible and clearly commented; flagging only as a possible follow-up, not a blocker.
  - `gh pr view` failures inside the loop `continue` (skip that PR) rather than fail-safe-blocked, meaning a transient API error could cause a false negative. Low likelihood, consistent with existing patterns elsewhere in these role files, not worth blocking on.

**Verification performed directly:**
- `bash -n` + `shellcheck -x` on the extracted `has_superseding_block()` — confirmed only pre-existing-style `SC2155` warnings (matches the surrounding file's existing `local x=$(...)` pattern), no new issues.
- Ran `.loom/scripts/tests/test-dependency-parse.sh` on both this branch and `main` — identical 18/22 pass, 4 fail in both; the 4 failures are a path-resolution bug in the test harness itself (`grep`s a nonexistent `.loom/.claude/commands/loom/*.md` path) unrelated to `parse_dependencies()` or this change.
- `gh pr checks 4655` — all applicable checks pass (Rust/MCP checks correctly skipped, docs-only change). `mergeStateStatus`: CLEAN.

Matches all four acceptance criteria in #4634: both locations gated identically, guard is not just a text note (curator.md's step 2 explicitly routes through the primary check before proceeding), #4492's regression scenario is documented and traced through the logic, and the secondary comment-history heuristic is clearly marked non-overriding in both files.

PR #4671, opened 2026-07-30
fix(init): log every .loom/config.json rewrite branch and stop re-init on configured workspaces

### LINKED ISSUE #4641: operator edits to .loom/config.json silently reverted by fleet processes (maxConcurrent tuning lost within hours)
## Context

Observed 2026-07-29→30 on loom-worker-1. As an interim throughput fix (context: #4512), the operator set in `/home/ubuntu/GitHub/anvil/.loom/config.json` (anvil = the daemon's default workspace on that host):

```jsonc
"workFinder": { "maxConcurrent": 10 }   // written ~21:00 PDT; daemon restarted; status showed "configured max 10"
```

A few hours later (after the host ran ~9 sweeps and a subsequent daemon restart), the same file reads:

```jsonc
"workFinder": { "enabled": true }
```

The key is not just changed — the operator's edit is **gone**, and `loom-daemon status` silently fell back to "configured max 3". No log line, no backup, no marker of what rewrote it. Candidates: a sweep child running the installer/resync in-workspace (`chore: resync installed Loom surfaces` commits do land on these repos), or a config-writing daemon path.

This is the third silent-clobber shape hit in two days, after #4522 (start script re-renders the launchd plist env from invocation env) and the same-day plist wipe that motivated it. Pattern: **fleet processes rewrite operator-tuned surfaces from templates/defaults without diffing or warning.**

- [ ] Identify which path rewrote `.loom/config.json` (resync deep-merge is supposed to preserve existing values — "template is the base, existing values win" — so either that contract broke, or another writer exists).
- [ ] Whatever writer it is: preserve unknown/operator-set keys on rewrite, and log a diff line when a rewrite changes effective config values.
- [ ] Regression test: an operator-added `workFinder.maxConcurrent` survives a resync/install pass over the workspace.

---

### Root Cause Investigation

The two most obvious suspects are **ruled out** by reading the actual code on `origin/main` — do not re-investigate them:

- **`defaults/scripts/resync-installed.sh` is NOT the writer.** Its own header explicitly lists `.loom/config.json` under "EXPLICITLY OUT OF SCOPE (never touched by resync — operator-owned; needs merge-semantics design)". It never opens the file.
- **`loom-daemon calibrate --write` is NOT the writer.** Since #4512, `calibrate` was reduced to measurement-only: `--write` is "accepted-but-ignored with a deprecation warning" (`loom-daemon/src/calibrate/mod.rs` module doc). It cannot write config today.

Grepping every non-test `fs::write(...".loom"..."config.json"...)` in `loom-daemon/src/` turns up exactly **one** production writer: `merge_config_file()` in `loom-daemon/src/init/mod.rs` (~lines 312-399), invoked by `loom-daemon init`. Its `deep_merge_existing_wins()` helper (~lines 423-437) is designed so **existing consumer keys always win**, including keys — like `workFinder.maxConcurrent` — that exist only in the consumer file and not in the shipped template; this is covered by tests (e.g. the reinstall-merge suite starting ~line 1830) and looks correct for the common path.

Two real call sites can invoke `loom-daemon init` against an **already-configured** workspace without direct operator action:

1. **`install.sh`'s reinstall flow** (operator- or automation-triggered `curl | bash` / `--confirm-reinstall`).
2. **`loom-daemon/src/fleet/add_worker.rs`'s `render_workspace_clone()`** (~lines 813-840): this renders a provisioning step that runs
   ```
   if [ ! -d "$HOME/{rel}/.git" ]; then gh repo clone {repo} "$HOME/{rel}"; fi
   loom-daemon init "$HOME/{rel}" --defaults "$LOOM_SRC/defaults" || true
   ```
   Note the `loom-daemon init` call is **unconditional** — it is not guarded by the same `[ ! -d .../.git ]` check as the clone, so any re-run of this provisioning step against a workspace that already exists (and is already tuned) re-invokes the merge path. This is a concrete lead specifically because the reporting host is named `loom-worker-1` — the same host sibling issue #4640 identifies as "provisioned via the `fleet add-worker` path". If `fleet add-worker` (or some idempotent re-run / fleet self-heal pass) executed against this host in the incident window, it would re-run `loom-daemon init` there.

**The one code path capable of a *silent, total* key-drop** (not just "template wins on a shared key", which the design already prevents) is `merge_config_file`'s "existing file is invalid JSON / not an object" fallback (~init/mod.rs lines 379-390): when the read of the existing `.loom/config.json` fails to parse, the function does a raw `fs::copy` of the shipped template over it — silently discarding every consumer key, with only an `eprintln!` warning (easy to miss if init runs unattended inside a provisioning script, as in `add_worker.rs` above, piped to a log the operator doesn't check). This would fire if something reads the file mid-write by a concurrent process (a second provisioning pass, or the live daemon touching the same file) and observes a torn/partial write. **Caveat**: the shipped `defaults/config.json` has no `autonomous`/`workFinder` key at all, so a raw fallback-to-template copy would produce a file with *no* `workFinder` key — not the `{"enabled": true}` shape quoted in the issue. Treat the issue's "before/after" JSON snippets as illustrative reconstructions, not confirmed byte-exact diffs; don't let the exact shape rule this branch out.

Given the ambiguity above, root-causing further by static reading has diminishing returns — the priority is the instrumentation called for in the original AC (add a diff-log line on every effective-config-changing write), which will make the *next* occurrence diagnosable instead of requiring code archaeology.

### Refined Acceptance Criteria

- [ ] Add logging to `merge_config_file()` (`loom-daemon/src/init/mod.rs`) that reports, on every call: which branch fired (fresh-write / merge-preserved / invalid-JSON-fallback-overwrite), and — when values actually changed — a diff of the changed keys. The invalid-JSON-fallback branch (~lines 379-390) should log at `warn!` (not just `eprintln!`) so it surfaces in daemon logs even when `loom-daemon init` runs unattended inside a provisioning script.
- [ ] Audit `fleet::add_worker::render_workspace_clone()` (`loom-daemon/src/fleet/add_worker.rs` ~lines 813-840): decide whether the unconditional `loom-daemon init` call should be skipped (or made explicitly safe/idempotent) when the target workspace directory already exists and is already configured, rather than always re-running the merge.
- [ ] Regression test: an operator-added nested key (`autonomous.workFinder.maxConcurrent`) survives **repeated** `loom-daemon init` calls against an already-configured workspace (not just a single reinstall) — add this near the existing reinstall-merge tests (`loom-daemon/src/init/mod.rs`, ~line 1830+).
- [ ] Regression test for the invalid-JSON fallback branch: assert it now logs a `warn!`-level message identifying the overwritten keys (verifies the first AC bullet actually closes the "no log line" gap the issue reports).
- [ ] Do not re-propose `resync-installed.sh` or `calibrate --write` as fixes — both are confirmed non-writers (see Root Cause Investigation above).

### Related Open Work

Sibling issue #4640 (filed in the same batch, same host `loom-worker-1`) cross-references this issue but only as "Sibling reports filed in the same batch (#4641–#4644) are distinct scope" — it is about a missing `LOOM_DAEMON_SUPERVISOR` env var in the fleet-worker systemd unit template, an unrelated defect in the same `add_worker.rs` file. Disregarded as a spec conflict; noted here only because it independently corroborates that `loom-worker-1` went through the `fleet add-worker` provisioning path, which is relevant context for this issue's root-cause investigation above.

## Affected Files

- `loom-daemon/src/init/mod.rs` - `merge_config_file()` / `deep_merge_existing_wins()` (~lines 312-437): add diff/branch logging; consider hardening the invalid-JSON fallback (e.g. write a `.bak` before overwriting) so a torn read can't silently discard operator config.
- `loom-daemon/src/fleet/add_worker.rs` - `render_workspace_clone()` (~lines 813-840): audit the unconditional re-init call for safety on repeated/idempotent provisioning runs.
- `defaults/scripts/resync-installed.sh` - reference only; confirms `.loom/config.json` is explicitly out of scope (no code change expected here).
- `loom-daemon/src/calibrate/mod.rs` - reference only; confirms `--write` is inert since #4512 (no code change expected here).

## Test Plan

- [ ] Manual verification: tune `autonomous.workFinder.maxConcurrent` on a live worker workspace, then run `loom-daemon init` against it multiple times (and, if feasible, concurrently with another process touching `.loom/config.json`); verify the tuned value survives and any anomaly is now logged.
- [ ] Automated tests: `cargo test` in `loom-daemon/` — new test(s) in `init/mod.rs` for repeated-init idempotence on a nested operator key, and for the invalid-JSON-fallback branch's new logging.
- [ ] Edge cases: existing consumer file is valid JSON but not an object (e.g. a JSON array) — confirm this still hits the same fallback-and-warn path and is now logged, not just an `eprintln!`.

## Dependencies

None blocking — all referenced code is on `origin/main`.

rjwalters:
Closes #4641

## Root cause

`merge_config_file()` (`loom-daemon/src/init/mod.rs`) is the **only** production writer of `.loom/config.json` — the curated issue rules out `resync-installed.sh` (declares the file out of scope) and `calibrate --write` (inert since #4512), and this PR does not re-propose either.

Its deep merge is existing-wins, so a nested operator-only key like `autonomous.workFinder.maxConcurrent` survives normally. Two gaps made the reported silent reversion possible *and* undiagnosable:

1. **The fallback branch clobbers everything, quietly.** When the on-disk file does not parse as a JSON object (a torn write from a concurrent writer, a hand-edit syntax error, or a top-level array), the function `fs::copy`s the shipped template over it — discarding every consumer key behind a bare `eprintln!` that vanishes when `loom-daemon init` runs unattended inside a provisioning script.
2. **`fleet::add_worker::render_workspace_clone()` re-entered the merge on every run.** Its `loom-daemon init` call was unconditional — unlike the `gh repo clone` directly above it, which has always been guarded by `[ ! -d .../.git ]` — so any repeat `fleet add-worker` / self-heal pass re-ran the merge against a workspace an operator had since hand-tuned. The reporting host `loom-worker-1` was provisioned through exactly this path.

## Fix

**Observability (AC1).** Every call emits exactly one greppable line:

```
grep 'init: config.json:' ~/.loom/daemon.log
```

| Branch | Level | Meaning |
|---|---|---|
| `fresh-write` | `info` | No config existed; template written verbatim (with key count). |
| `merge-preserved` | `info` | Existing values kept. Carries a leaf-level diff or "no effective config change". |
| `template-invalid-skip` | `warn` | Shipped template unparseable; consumer file untouched (and got no updates). |
| `invalid-JSON-fallback-overwrite` | `warn` | **Config replaced by the template** — names the discarded keys and the rescue copy. |

The diff renders `+ path = value`, `~ path: old -> new`, `- path (was value)` with dotted paths, arrays compared wholesale (matching the merge primitive), elided past 20 entries so a large template cannot emit a multi-kilobyte line. Under existing-wins semantics only `+` entries are expected; a `~` or `-` is the signal that a consumer value was overwritten.

**Recoverability.** The clobbering branch now copies the unparseable bytes to `.loom/config.json.bak` *before* overwriting, and names the keys it discarded — salvaged by a small scanner, since `serde_json` can enumerate nothing from text that failed to parse. `.loom/*.bak` was added to `EPHEMERAL_PATTERNS` and `.gitignore` so salvage never gets committed.

**Re-entry (AC2).** `render_workspace_clone()` now gates `loom-daemon init` on `.loom/config.json` being absent. Provisioning only needs `init` to *establish* a workspace; keeping an established one current is `loom update` / resync's job, and neither touches `.loom/config.json`. Gating on the config file rather than on "we just cloned" still self-heals a workspace whose earlier `init` failed. The step's completion check (`test -d .git`) is unaffected.

## Tests

- **AC3** — `test_operator_nested_key_survives_repeated_init`: **five** consecutive `init` passes over an already-configured workspace keep `autonomous.workFinder.maxConcurrent = 10` and the file is byte-stable from pass 2 onward. The fixture template deliberately omits `maxConcurrent`, which is the exact shape that makes a template-wins or fallback-copy bug surface as a silent revert to the built-in default.
- **AC4** — `test_invalid_json_fallback_warns_and_names_discarded_keys`: asserts `log::Level::Warn` (not mere presence — a regression back to `eprintln!`/`debug!` fails), the branch name, all four discarded key names, and that the `.bak` rescue copy exists with the original bytes.
- Plus: the other three branches asserted by name and severity, valid-JSON-but-an-array taking the same clobbering path (the edge case from the issue's test plan), and unit coverage of the diff / salvage / elision helpers.

The capturing logger moved from `main_health_gate`'s test module to `crate::test_log_capture`: `log::set_boxed_logger` succeeds only **once per process** and every `#[cfg(test)]` module compiles into the same binary, so a second per-module logger would have silently returned empty records for whichever module lost the race — an order-dependent false pass. Call sites are unchanged via a `use ... as capture` alias.

## Verification

`cargo build`, `cargo fmt --all -- --check`, CI's exact `cargo clippy` invocation, `cargo test --workspace`, and `defaults/scripts/tests/test-install-reinstall-safety.sh` all pass.

Operator-facing behavior (the branch table, the `.bak` recovery path) is documented in a new troubleshooting subsection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ Thumbs.db
 .loom/*.log
 .loom/*.sock
 .loom/*.tmp
+.loom/*.bak
 .loom/logs/
 # <<< loom-managed <<<
 .loom-test-failure-context.json
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -259,6 +259,44 @@ Reaching for `loom:blocked` when you mean `loom:operator-only` conflates "waitin
 a dependency" with "needs a human action," which muddies the daemon/sweep skip
 semantics. Use `loom:operator-only` for the human-must-act-off-automation case.
 
+### An operator edit to `.loom/config.json` disappeared (#4641)
+
+`.loom/config.json` has exactly one production writer: `loom-daemon init` (run by
+`install.sh` reinstalls and by the `fleet add-worker` provisioning steps).
+`resync-installed.sh` and `loom-daemon calibrate --write` never touch it.
+
+`init`'s merge is **existing-wins** — consumer keys, including ones absent from the
+shipped template such as `autonomous.workFinder.maxConcurrent`, survive — so a
+tuned value vanishing means one of the other branches fired. Every `init` call now
+emits one greppable line naming the branch it took:
+
+```bash
+grep 'init: config.json:' ~/.loom/daemon.log
+```
+
+| Branch | Level | Meaning |
+|--------|-------|---------|
+| `fresh-write` | `info` | No config existed; the template was written verbatim. |
+| `merge-preserved` | `info` | Existing values kept. Carries a per-key diff (`+ added`, `~ changed`, `- dropped`) or "no effective config change". |
+| `template-invalid-skip` | `warn` | The shipped `defaults/config.json` is unparseable; your file was left untouched (and got no template updates). |
+| `invalid-JSON-fallback-overwrite` | `warn` | **Your config was replaced by the template.** The line names the discarded keys. |
+
+A `~` or `-` entry on `merge-preserved` is unexpected under existing-wins semantics
+and is worth investigating.
+
+The fallback branch only fires when the file on disk does not parse as a JSON
+object (a torn write from a concurrent writer, a hand-edit with a syntax error, or
+a top-level array). Before overwriting, `init` copies the unparseable bytes to
+`.loom/config.json.bak` — recover your tuned keys from there:
+
+```bash
+cat .loom/config.json.bak
+```
+
+`.loom/*.bak` is git-ignored, and `fleet add-worker` no longer re-runs
+`loom-daemon init` against a workspace that already has a `.loom/config.json`, so
+repeat provisioning passes cannot re-enter this path on a tuned host.
+
 ### Daemon won't start
 
 ```bash
--- a/loom-daemon/src/fleet/add_worker.rs
+++ b/loom-daemon/src/fleet/add_worker.rs
@@ -470,7 +470,7 @@ pub fn build_plan(config: &AddWorkerConfig, secrets: &Secrets) -> Plan {
     // 6. Clone + init workspace repos (init installs the /loom:sweep command, #4027).
     plan.push_step(Step::new(
         "workspace-clone",
-        "clone workspace repo(s) and run loom-daemon init (installs /loom:sweep)",
+        "clone workspace repo(s) and run loom-daemon init on unconfigured ones (installs /loom:sweep)",
         Some(render_workspace_clone_check(&config.repos)),
         render_workspace_clone(&config.repos),
     ));
@@ -819,6 +819,23 @@ fn render_workspace_clone_check(repos: &[String]) -> String {
     s
 }
 
+/// Render the workspace clone + first-time init step.
+///
+/// **`loom-daemon init` runs only on a workspace that is not yet configured**
+/// (issue #4641). It used to run unconditionally — unlike the `gh repo clone`
+/// above it, which has always been guarded by `[ ! -d .../.git ]` — so every
+/// re-run of `fleet add-worker` (or any idempotent self-heal pass over an
+/// existing host) re-entered `init`'s `.loom/config.json` merge on a workspace
+/// an operator had since hand-tuned. That merge is existing-wins and normally
+/// harmless, but its invalid-JSON fallback branch overwrites the file
+/// wholesale, which is a plausible cause of the observed silent reversion of
+/// `autonomous.workFinder.maxConcurrent` on `loom-worker-1`.
+///
+/// Provisioning only needs `init` to *establish* a workspace; keeping an
+/// already-established one up to date is `loom update` / the resync script's
+/// job, neither of which touches `.loom/config.json`. Gating on
+/// `.loom/config.json` (rather than on having just cloned) also self-heals a
+/// workspace that was cloned by a previous run whose `init` failed.
 fn render_workspace_clone(repos: &[String]) -> String {
     let mut s = String::from(
         r#"set -e
@@ -833,7 +850,11 @@ mkdir -p "$HOME/loom-workspaces"
             r#"if [ ! -d "$HOME/{rel}/.git" ]; then
   gh repo clone {repo} "$HOME/{rel}"
 fi
-loom-daemon init "$HOME/{rel}" --defaults "$LOOM_SRC/defaults" || true
+if [ ! -f "$HOME/{rel}/.loom/config.json" ]; then
+  loom-daemon init "$HOME/{rel}" --defaults "$LOOM_SRC/defaults" || true
+else
+  echo "skip init: $HOME/{rel} already configured (.loom/config.json present)"
+fi
 "#
         ));
     }
@@ -1300,6 +1321,41 @@ mod tests {
                 "safehouse",
                 "verify",
             ]
+        );
+    }
+
+    #[test]
+    fn workspace_clone_only_inits_an_unconfigured_workspace() {
+        // #4641: `loom-daemon init` used to run unconditionally here, so every
+        // re-run of provisioning re-entered the `.loom/config.json` merge on a
+        // workspace an operator had since hand-tuned. The init call must now sit
+        // behind a `.loom/config.json` existence guard, the same way the clone
+        // sits behind a `.git` guard.
+        let script = render_workspace_clone(&["rjwalters/anvil".to_string()]);
+
+        assert!(
+            script.contains(r#"if [ ! -f "$HOME/loom-workspaces/anvil/.loom/config.json" ]; then"#),
+            "init must be guarded on the workspace being unconfigured:\n{script}"
+        );
+
+        // The guard must actually enclose the init call: the only `loom-daemon
+        // init` line has to appear after the guard and before its `else`.
+        let guard = script
+            .find(r#"if [ ! -f "$HOME/loom-workspaces/anvil/.loom/config.json" ]"#)
+            .expect("guard present");
+        let init = script.find("loom-daemon init").expect("init present");
+        let else_branch = script.find("\nelse\n").expect("else present");
+        assert!(guard < init && init < else_branch, "init is outside the guard:\n{script}");
+        assert_eq!(
+            script.matches("loom-daemon init").count(),
+            1,
+            "no second, unguarded init call may remain:\n{script}"
+        );
+
+        // The clone itself stays guarded on .git (unchanged behavior).
+        assert!(
+            script.contains(r#"if [ ! -d "$HOME/loom-workspaces/anvil/.git" ]; then"#),
+            "clone guard regressed:\n{script}"
         );
     }
 
--- a/loom-daemon/src/init/mod.rs
+++ b/loom-daemon/src/init/mod.rs
@@ -302,13 +302,30 @@ fn copy_single_file(
 ///   conflict; keys new in the template are added, unknown consumer keys at any
 ///   depth are preserved. Written with deterministic pretty serialization so
 ///   repeat reinstalls are byte-idempotent. Recorded as `preserved`.
-/// - **Destination exists but is invalid JSON (or not an object)** → fall back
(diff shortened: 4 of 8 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T19:33:22Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved. Reviewed the diff in the existing builder worktree (`.loom/worktrees/issue-4641`, PR head `834d6ade`), plus `cargo build --workspace`, `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and targeted `cargo test -p loom-daemon --lib init:: / fleet::add_worker::` — all pass locally, and CI's `Rust Unit Tests` (which covers `main_health_gate::` too) is green.

**Correctness of the core fix (init/mod.rs)**
- `merge_config_file`'s existing-wins deep merge is untouched; the new logging sits around it. `describe_config_changes` diffs leaf paths correctly (verified via `test_describe_config_changes_unit` and by reading the flatten/diff logic — arrays are compared wholesale, matching `deep_merge_existing_wins`'s own array-replaces-wholesale semantics, so the diff can't lie about what the merge actually does).
- The invalid-JSON fallback branch (the one branch capable of a *silent, total* key drop, per the curated issue) now writes `.loom/config.json.bak` **before** overwriting and `warn!`s with both the discard reason and salvaged key names. `test_invalid_json_fallback_warns_and_names_discarded_keys` and `test_valid_json_but_not_an_object_hits_fallback_and_warns` assert the `Warn` *level* specifically (not just message presence), which is exactly the guard against a future regression back to a swallowed `eprintln!`.
- `test_operator_nested_key_survives_repeated_init` runs the real `initialize_workspace` five times end-to-end and asserts both value-survival and byte-stability from pass 2 onward — a real regression test, not a trivial one.

**fleet::add_worker gating (AC2)** — worth a closer look, but not blocking:
`initialize_workspace` does substantially more than merge `config.json` — it also syncs `roles/`, `scripts/`, `hooks/`, `docs/`, `bin/` (cleaning stale files on reinstall) and runs repository scaffolding, including the `/loom:sweep` command install the updated step description calls out. Gating the entire `loom-daemon init` call on `.loom/config.json`'s mere *existence* means any repeat `fleet add-worker` pass over an already-configured workspace now skips **all** of that, not just the risky config merge — so if `fleet add-worker` is ever relied on as a self-heal/upgrade path to push newer Loom defaults (roles/scripts/hooks/docs/bin) onto an existing worker, this PR silently disables it (visible only as a one-line `echo` in the provisioning script's own stdout). The PR's rationale — that ongoing updates are `loom update`/resync's job — is plausible given `resync-installed.sh`'s documented scope, but nothing in `add_worker.rs`'s own plan invokes resync for these workspace repos, so that's an assumption about other tooling rather than something this PR verifies. The curated issue explicitly left "skip vs. make safely idempotent" as a judgment call, and this is a defensible one with a documented rationale (both in the code comment and troubleshooting.md) — but it's a broader behavior change than the narrow bug (config.json reversion) requires. Consider a follow-up issue: either keep the managed-dir syncs running unconditionally (they're idempotent/low-risk, unlike the config merge) and gate only the config-merge risk, or confirm explicitly that resync is in fact wired into the ongoing-update path for fleet-provisioned workers.

**test_log_capture extraction** — sound. Single `Once`-gated `log::set_boxed_logger` shared across the whole test binary, thread-local capture buffer scoped by an `ACTIVE` flag, so concurrent tests can't cross-contaminate captures. `main_health_gate.rs`'s call sites are unchanged via the `use crate::test_log_capture as capture;` alias, and CI's `Rust Unit Tests` job (which builds and runs the whole workspace test suite, including `main_health_gate::`) is green, so the extraction isn't flaking there.

**Minor, non-blocking nit**: when a *fresh* install (`!dst.exists()`) hits a shipped template that itself fails to parse (`init/mod.rs` ~line 209), the new log line is tagged `fresh-write` at `warn!` level — but `defaults/docs/troubleshooting.md`'s branch table documents `fresh-write` as always `info`. Not a functional bug (still greppable, still correctly `warn!`), just a minor label/doc mismatch worth a follow-up if it's ever confusing in practice.

CI: all required checks pass, `mergeStateStatus` is `CLEAN`. Test plan from the PR description executed via the automated test suite (repeated-init idempotence, invalid-JSON-fallback logging, array-shaped edge case) — all covered by the new tests I ran above.

Issue #4206, opened 2026-07-28
fix(daemon): quarantine automation overrides a deliberate operator loom:blocked park — hourly crash-loop on previously-quarantined issues


rjwalters:
## Problem

The daemon's quarantine/label automation cannot distinguish a **deliberate operator park** (`loom:blocked` applied by a human for a real dependency) from its **own quarantine label** — and it repeatedly overrides the human. Observed live on #4063 on 2026-07-28: the operator parked it pending #4178, and within 25 minutes the daemon had re-armed `loom:issue`, re-dispatched it, insta-crashed three times, and re-quarantined it. This will repeat on a ~TTL cadence until #4178 lands, burning dispatch slots and stacking duplicate quarantine comments.

## Evidence (#4063 label timeline, 2026-07-28, all events under the operator token)

```
05:17:52  -loom:issue +loom:blocked   quarantine #1 (comment 05:17:54)
05:51:25  +loom:issue -loom:blocked   daemon release re-arms the issue
05:52:24  -loom:issue +loom:blocked   OPERATOR PARK on #4178 (comment 05:52:22)
05:52:29  +loom:building              in-flight dispatch claims it anyway
05:52:48  -loom:building +loom:issue  crash-path restore_label_to_ready re-adds
                                      loom:issue ON TOP of the operator's park
                                      → illegal loom:blocked + loom:issue combo
06:16:09  -loom:blocked               quarantine release strips the OPERATOR's park
06:16:25 … 06:17:48                   three claim/crash/restore cycles (<60s each)
06:17:50  -loom:issue +loom:blocked   quarantine #2 (comment 06:17:51)
```

Two distinct automation paths overrode the park:

1. **Crash-path label restore** — `SweepRegistry::restore_label_to_ready` (`loom-daemon/src/sweep_registry.rs:2323`) flips `loom:building` → `loom:issue` when a child dies, without checking whether `loom:blocked` is currently present. At 05:52:48 this re-added `loom:issue` to an issue the operator had parked 24 seconds earlier, producing the `loom:blocked`+`loom:issue` combo (which the work-finder's `SKIP_LABELS` at `work_finder.rs:159` happens to skip — but only until the next path fires).
[...]

PR #4674, opened 2026-07-30
fix(daemon): refuse loom:issue re-add in restore_label_to_ready for PR numbers

### LINKED ISSUE #4653: daemon: restore_label_to_ready doesn't check PR-vs-issue before re-adding loom:issue on cancel/crash-recovery
## Context

During a `/loom:sweep all` session in the claude-monitor workspace (2026-07-30), a `dispatch_sweep` call that omitted `workspace_root` routed to the daemon's default workspace (this repo). The candidate numbers 13–16 are issues in claude-monitor but map to different objects here:

- `{"Issue": 13}` / `{"Issue": 14}` → **correctly refused** by the #4088 closed-issue guard (both are closed issues here).
- `{"Issue": 15}` / `{"Issue": 16}` → **dispatched successfully**, even though #15 is a *merged PR* and #16 a *closed PR* in this repo. The dispatch flipped `loom:building` on both PRs; after `cancel_sweep`, both were left carrying a stray `loom:issue` label (never present before the dispatch) that needed manual cleanup.

The guard appears to check issue closed-state only, so a number that resolves to a PR (GitHub shares the number space) bypasses it entirely.

- [ ] `dispatch_sweep` refuses when the target number resolves to a pull request (of any state) rather than an issue, with an error naming the mismatch.
- [ ] The cancel/termination path restores the claim label to its pre-dispatch state instead of unconditionally applying `loom:issue`.
- [ ] Regression test: dispatching a number that is a merged PR in the target workspace is refused.

---

### Verified diagnosis — the reported symptom is already fixed on `main`

This issue's premise ("the closed-issue guard doesn't catch PR numbers") describes exactly the bug already reported, root-caused, fixed, and tested in **#4504**, closed via **PR #4567** ("fix(daemon): make the closed-issue dispatch guard PR-aware via a REST probe"), merged **2026-07-30T05:39:34Z** — the same day this issue was filed (17:51:46Z), but ~12 hours later.

Confirmed against `origin/main` (`loom-daemon/src/sweep_registry.rs`):

- The dispatch guard (`issue_is_closed_or_pr`, `sweep_registry.rs:4076-4113`) already switched from `gh issue view --json state` to the REST endpoint (`repos/{owner}/{repo}/issues/{n} --jq '{state, is_pr: (.pull_request != null)}'`), which structurally discriminates PR-ness independent of state. It refuses dispatch when `is_pr` is true in **any** state (open, closed, or merged) — see the call site at `sweep_registry.rs:3333` and its comment block at `3318-3332` citing #4088/#4504 directly.
- Regression tests already exist for exactly the three cases this issue's original AC3 asked for: `dispatch_refuses_merged_pr_number_without_flipping_labels` (merged PR, `sweep_registry.rs:~15907`), an open-PR-number refusal test (`~15944`), and a bare `MERGED`-state-on-an-issue-node test (`~15975`) — all built on the existing `closed_guard_registry` fake-`gh` fixture.

**Conclusion**: the original issue's AC1 ("refuses when target resolves to a PR, any state") and AC3 ("regression test for merged-PR dispatch") are **already satisfied** on `main`. The incident described in this issue's Context (candidates #15/#16 dispatching as merged/closed PRs) most likely happened before PR #4567 landed that same day, or reflects a daemon process that had not yet picked up the new binary — not a residual code defect in the guard itself. Re-implementing or re-testing AC1/AC3 here would just be redundant with #4504/#4567.

### The one claim that is NOT yet covered — narrowed scope

This issue's **second** acceptance criterion is a distinct, still-open, and genuinely narrower gap: the cancel/crash-recovery label-restore path does not re-verify PR-vs-issue before restoring `loom:issue`.

- `restore_label_to_ready` (`sweep_registry.rs:4365-4403`) is called from three places: `finish_cancel` (`~5590`), and two branches inside `reap_once`'s crash/no-progress recovery (`~5727`, `~6007`). All three unconditionally remove `loom:building` and re-add `loom:issue` **unless** the issue currently carries `loom:blocked` (checked via `issue_has_blocked_label`, `~5025`) — there is no equivalent check for "is this number actually a PR, not an issue."
- Today the 2.5 dispatch guard (`issue_is_closed_or_pr`) prevents a PR number from ever reaching `loom:building` in the first place, in the *common* case. But that guard is deliberately **fail-open**: `sweep_registry.rs:3333`'s comment states a forge lookup error returns `None` and dispatch proceeds, "so a `gh` outage can never wedge the daemon." If the probe fails open specifically for a PR number (timeout, transient `gh api` error), dispatch proceeds, `loom:building` gets applied to the PR, and — since a sweep operating on an already-existing PR object will almost certainly never populate its own `pr_number` field — the eventual cancel or crash-recovery path calls `restore_label_to_ready`, which unconditionally re-adds `loom:issue` to that PR. This reproduces exactly the residual symptom described in this issue's Context ("left carrying a stray `loom:issue` label... that needed manual cleanup"), just gated behind the guard's fail-open window rather than a state-matching bug.
- This is defense-in-depth, not a fix to the primary guard (which is correct today) — it closes the same PR-vs-issue gap at the one place that doesn't yet check it.

### Sharpened Acceptance Criteria (replaces the original three)

- [ ] `restore_label_to_ready` refuses to add `loom:issue` when the target number resolves to a pull request (reuse the existing `issue_is_closed_or_pr` probe, or the `is_pr` half of its REST payload) — it should still remove the stale `loom:building` claim, matching the existing `loom:blocked` carve-out's shape (clear the claim, skip the re-add).
- [ ] All three call sites (`finish_cancel`, and both `reap_once` recovery branches) go through the same updated `restore_label_to_ready`, so the fix is one change, not three.
- [ ] Regression test: extend the existing fake-`gh` fixture pattern (see `restore_label_to_ready_does_not_add_loom_issue_when_blocked`, `sweep_registry.rs:~12330`) with a case where the probe reports `is_pr: true` — assert `loom:building` is removed but `loom:issue` is NOT added.
- [ ] No change to the 2.5 dispatch guard itself or its existing tests — those already satisfy the original issue's AC1/AC3 (see Verified diagnosis above); do not duplicate that work here.

### Implementation Guidance

Add an `is_pr` check at the top of `restore_label_to_ready` (`sweep_registry.rs:4365`), following the same best-effort/fail-open shape as the existing `blocked` check — i.e., a probe failure should not block the claim-label cleanup (`loom:building` removal), only the `loom:issue` re-add is conditional. Reusing `issue_is_closed_or_pr` directly is simplest since it already returns the structural `is_pr` signal; alternatively thread the pre-dispatch `is_pr` result through the entry state if avoiding a second `gh api` call at restore time is preferred (cheaper, but a slightly bigger change touching entry construction) — either is acceptable, cheapest-first per the marker below.

### Duplicate / related-work check

`check-duplicate.sh --include-merged-prs --issue 4653` returned only this issue itself (title/body self-match against the open-issues search pool — not a real duplicate) and no `RELATED_OPEN_WORK` cross-references. Manually verified against #4504 (see Verified diagnosis above — same root claim, already fixed) and #4503 (workspace_root defaulting footgun, already closed/fixed, distinct root cause — the workspace-root mismatch in this issue's Context is that bug's symptom, not something to re-fix here).

### Dependencies

None — self-contained fix within `loom-daemon/src/sweep_registry.rs`, building on the already-merged #4504/#4567 guard.

## Affected Files

- `loom-daemon/src/sweep_registry.rs` — `restore_label_to_ready` (~line 4365-4403): add an `is_pr` check before the `loom:issue` re-add, mirroring the existing `blocked` carve-out.
- `loom-daemon/src/sweep_registry.rs` — extend the `restore_label_to_ready_*` test block (~line 12330) with an `is_pr: true` fixture case.

## Test Plan

- [ ] Automated: new unit test asserting `restore_label_to_ready` removes `loom:building` but does NOT add `loom:issue` when the probe reports `is_pr: true`, using the existing fake-`gh` fixture pattern.
- [ ] Automated: confirm the existing `restore_label_to_ready_does_not_add_loom_issue_when_blocked`, `cancel_restores_label_when_no_pr_produced`, and `cancel_does_not_restore_label_when_pr_produced` tests still pass unmodified.
- [ ] Run `cargo test -p loom-daemon` (or the workspace's nextest invocation) to catch regressions in the three call sites (`finish_cancel`, both `reap_once` branches).

rjwalters:
## Summary
`restore_label_to_ready` (the shared crash-recovery/cancel label-restore path used by `finish_cancel` and both `reap_once` recovery branches) unconditionally re-added `loom:issue` unless the target already carried `loom:blocked`. If the 2.5 dispatch guard's fail-open window ever let a PR number reach `loom:building` (e.g. a transient `gh` outage during the guard's probe), the eventual cancel/crash-recovery path would strand that PR with a stray `loom:issue` label, reproducing the incident described in #4653.

## Changes
- Added `issue_is_pull_request` (`loom-daemon/src/sweep_registry.rs`): a best-effort, fail-open REST probe mirroring the `is_pr` half of `issue_is_closed_or_pr`'s payload, kept as a separate helper so the 2.5 dispatch guard (`issue_is_closed_or_pr`) and its existing tests are untouched.
- `restore_label_to_ready` now probes `issue_is_pull_request` (only when the issue is not already `loom:blocked`, to avoid a redundant `gh` call on the more common blocked path) and skips the `loom:issue` re-add when the target resolves to a PR — matching the existing `loom:blocked` carve-out's shape: `loom:building` is always removed, only the re-add is conditional.
- All three call sites (`finish_cancel`, and both `reap_once` recovery branches) already funnel through this single `restore_label_to_ready`, so this is one change, not three.
- Updated `label_flips_run_in_registry_workspace_root`'s expected `gh` call count (3 -> 4) to account for the new `is_pr` probe's `resolve_owner_repo` lookup, which bails out before a second `gh api` call because the test's fake `gh` prints nothing for `repo view`.
- Added a new regression test, `restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr`, extending the existing fake-`gh` fixture pattern with a case where the probe reports `is_pr: true`.

| `restore_label_to_ready` refuses to add `loom:issue` when the target resolves to a PR, still clearing `loom:building` | ✅ | New test `restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr` asserts `loom:building` removed, `loom:issue` not added |
| All three call sites go through the same updated `restore_label_to_ready` (one change, not three) | ✅ | Verified via `grep -n "restore_label_to_ready("` — `finish_cancel` and both `reap_once` branches call the same function; no per-call-site duplication added |
| Regression test extends the existing fake-`gh` fixture pattern (`restore_label_to_ready_does_not_add_loom_issue_when_blocked`) with an `is_pr: true` case | ✅ | New test follows the identical fixture shape, adding an `if [ "$1" = "api" ]` branch to the fake `gh` script |
| No change to the 2.5 dispatch guard (`issue_is_closed_or_pr`) or its existing tests | ✅ | `issue_is_closed_or_pr` body/tests untouched; new `issue_is_pull_request` is a separate helper |

## Test Plan
- `cargo test -p loom-daemon --lib sweep_registry::tests -- --test-threads=4/8` — 284 tests passed (0 failed), including:
  - `restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr` (new)
  - `restore_label_to_ready_does_not_add_loom_issue_when_blocked` (existing, unmodified assertions, still passes)
  - `cancel_restores_label_when_no_pr_produced` (existing, unmodified, still passes)
  - `cancel_does_not_restore_label_when_pr_produced` (existing, unmodified, still passes)
  - `label_flips_run_in_registry_workspace_root` (existing, call-count assertion updated 3 -> 4 to reflect the new probe, still passes)
- `cargo check -p loom-daemon` and `cargo clippy -p loom-daemon --all-targets` — clean.
- `cargo fmt -p loom-daemon -- --check` — no diff.

Closes #4653

### DIFF
--- a/loom-daemon/src/sweep_registry.rs
+++ b/loom-daemon/src/sweep_registry.rs
@@ -4400,6 +4400,48 @@ impl SweepRegistry {
         }
     }
 
+    /// Best-effort probe for whether `issue` resolves to a pull request in ANY
+    /// state (Issue #4653), used by [`restore_label_to_ready`] to refuse
+    /// re-adding `loom:issue` to a PR number during cancel/crash-recovery.
+    /// Returns `Some(true)` when the number is a PR, `Some(false)` when it is
+    /// a genuine issue, and `None` on any probe failure (missing/failed/
+    /// timed-out `gh`, unresolvable repo, unparseable output).
+    ///
+    /// Mirrors the same REST call as [`issue_is_closed_or_pr`] (the `is_pr`
+    /// half of its payload), kept as a separate probe so this fix does not
+    /// touch the 2.5 dispatch guard or its tests. Callers MUST treat `None`
+    /// as **fail-open** — a forge outage must never block the `loom:building`
+    /// cleanup that [`restore_label_to_ready`] performs regardless of this
+    /// probe's outcome.
+    ///
+    /// [`restore_label_to_ready`]: Self::restore_label_to_ready
+    /// [`issue_is_closed_or_pr`]: Self::issue_is_closed_or_pr
+    fn issue_is_pull_request(&self, issue: u32) -> Option<bool> {
+        let (owner, repo) = self.resolve_owner_repo()?;
+        let gh = self
+            .config
+            .gh_bin
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("gh"));
+        let mut cmd = Command::new(&gh);
+        cmd.arg("api")
+            .arg(format!("repos/{owner}/{repo}/issues/{issue}"))
+            .arg("--jq")
+            .arg(".pull_request != null");
+        // Resolve against this registry's own workspace, matching the other
+        // dispatch-path probes (#3937).
+        cmd.current_dir(&self.config.workspace_root);
+        let output = output_with_timeout(cmd, reap_gh_timeout()).ok()??;
+        if !output.status.success() {
+            return None;
+        }
+        match String::from_utf8_lossy(&output.stdout).trim() {
+            "true" => Some(true),
+            "false" => Some(false),
+            _ => None,
+        }
+    }
+
     /// Best-effort probe for the first [`PARK_LABELS`] entry currently on
     /// `issue`, used by the #4444 park-label dispatch guard (step 2.7). Returns
     /// `Some(label)` when the issue carries a park label and `None` otherwise —
@@ -4548,15 +4590,20 @@ impl SweepRegistry {
 
     /// Restore a crashed/orphaned claim's `loom:building` back to
     /// `loom:issue` — UNLESS the issue currently carries `loom:blocked`
-    /// (Issue #4206). A deliberate operator park (applied by hand, possibly
-    /// while the now-dead sweep was still `loom:building`) must never be
-    /// clobbered into the illegal `loom:blocked` + `loom:issue` combo by the
-    /// crash-recovery path. Re-reads the issue's current labels first (the
-    /// same probe pattern as [`Self::issue_has_blocked_label`], used
-    /// elsewhere on the reap path) — best-effort and fail-open: an
-    /// unverifiable read falls back to the pre-#4206 unconditional restore,
-    /// since a stranded `loom:building` claim is the more common failure
-    /// mode this path exists to fix.
+    /// (Issue #4206) OR the target number actually resolves to a pull
+    /// request (Issue #4653). A deliberate operator park (applied by hand,
+    /// possibly while the now-dead sweep was still `loom:building`) must
+    /// never be clobbered into the illegal `loom:blocked` + `loom:issue`
+    /// combo by the crash-recovery path, and a PR number must never be
+    /// handed a `loom:issue` label meant for issues — that reproduces the
+    /// stray-label symptom from #4653's incident report, reached whenever the
+    /// 2.5 dispatch guard's fail-open window lets a PR number through and the
+    /// sweep is later cancelled or crash-recovered. Both checks are
+    /// best-effort and fail-open: an unverifiable read falls back to the
+    /// pre-#4206 unconditional restore (re-adding `loom:issue`), since a
+    /// stranded `loom:building` claim is the more common failure mode this
+    /// path exists to fix. Either carve-out only skips the `loom:issue`
+    /// re-add — the stale `loom:building` claim is always removed.
     fn restore_label_to_ready(&self, issue: u32) -> Result<()> {
         let gh = self
             .config
@@ -4564,6 +4611,10 @@ impl SweepRegistry {
             .clone()
             .unwrap_or_else(|| PathBuf::from("gh"));
         let blocked = self.issue_has_blocked_label(issue);
+        // Only probe PR-ness when the blocked carve-out doesn't already
+        // decide the outcome — avoids a redundant `gh` call on the (more
+        // common) blocked path.
+        let is_pr = !blocked && self.issue_is_pull_request(issue).unwrap_or(false);
         let mut cmd = Command::new(&gh);
         cmd.arg("issue")
             .arg("edit")
@@ -4575,6 +4626,12 @@ impl SweepRegistry {
                 "sweep_registry: restore_label_to_ready for #{issue} found `loom:blocked` \
                  already present — preserving the operator's park by removing the stale \
                  `loom:building` claim only, NOT re-adding `loom:issue` (#4206)"
+            );
+        } else if is_pr {
+            log::info!(
+                "sweep_registry: restore_label_to_ready for #{issue} resolved to a pull \
+                 request — removing the stale `loom:building` claim only, NOT re-adding \
+                 `loom:issue` (#4653)"
             );
         } else {
             cmd.arg("--add-label").arg("loom:issue");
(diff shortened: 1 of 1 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T20:02:18Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved.

**Verification performed:**
- Diff (single file, `loom-daemon/src/sweep_registry.rs`) matches the stated fix exactly: a new best-effort, fail-open `issue_is_pull_request` probe (separate from the 2.5 dispatch guard's `issue_is_closed_or_pr`, which is untouched — confirmed by grep, no diff to that function or its tests) is queried by `restore_label_to_ready` only when the target is not already `loom:blocked` (avoiding a redundant `gh` call on the more common path), and skips the `loom:issue` re-add — while still always removing `loom:building` — when the target resolves to a PR.
- Confirmed all three call sites (`finish_cancel` line 5786, and both `reap_once` recovery branches lines 5925/6205) funnel through the single updated `restore_label_to_ready` — this is one change, not three, as claimed.
- New regression test `restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr` correctly asserts both halves of the contract: `loom:building` is still removed AND `loom:issue` is NOT added when the fake `gh` probe reports `is_pr: true`. Ran it in isolation — passes.
- The `label_flips_run_in_registry_workspace_root` call-count update (3 -> 4) is correctly justified: the new probe's `resolve_owner_repo` call is the added `gh` invocation, and the explanation of why it doesn't add a second `gh api` call (fake `gh` prints nothing for `repo view`, so the probe short-circuits) checks out against the code. Ran this test in isolation — passes.
- Ran the full `sweep_registry::tests` module locally in the builder's worktree: 284 passed, 0 failed — matches the PR description exactly.
- `cargo fmt -p loom-daemon -- --check`: clean. `cargo clippy -p loom-daemon --all-targets -- -D warnings`: clean.
- `gh pr checks`: all required checks pass (Rust Linting & Build Check, Rust Unit Tests, AGENTS.md Sync, CLAUDE.md Line Budget, Detect Changes); the rest correctly skip (path-filtered, unrelated to this Rust-only change).
- `mergeStateStatus`: CLEAN.

Code quality is solid: the fail-open semantics, doc comments, and log messages consistently mirror the existing #4206 `loom:blocked` carve-out's shape, and the new probe is properly isolated from the untouched dispatch-guard path per the acceptance criteria. No issues found.

Issue #4280, opened 2026-07-28
install: .loom/sweep-checkpoint/ runtime state is not gitignored in consumer repos


rjwalters:
## Context

Observed 2026-07-28 in a consumer repo (anvil, loom 0.12.0 install): after daemon-dispatched sweeps ran, `git status` shows

```
?? .loom/sweep-checkpoint/
```

Sweep checkpoints are per-issue runtime state (`.loom/sweep-checkpoint/issue-<N>.json`, crash-recovery data) — the same class as `.loom/logs/`, which is ignored. Untracked-but-not-ignored runtime state invites accidental commits via a routine `git add -A`, and permanently dirties `git status` in every consumer repo the daemon works in. (Epic #3835's four-concern split names runtime state as "ignored" — this is a concrete instance where it isn't.)

## Suggested acceptance criteria

- [ ] The installer's gitignore management covers `.loom/sweep-checkpoint/` (and audits for any other runtime-state dirs written by the current sweep/daemon code paths that postdate the installed ignore rules).
- [ ] Existing consumer installs pick this up via the resync/update path (#4239) or the migration command (#4254), not only fresh installs.

---

### Corrected problem framing (root cause)

The installer's pattern list is **not** missing `.loom/sweep-checkpoint/` — it has been in `EPHEMERAL_PATTERNS` (`loom-daemon/src/init/post_init.rs`, currently line 130) since #3642 (commit `3f709b42`, in every release ≥ v0.11.0), with idempotency tests. The observed dirt is a **convergence failure for existing installs**, with three distinct defects:

1. **The resync path never refreshes the managed gitignore block.** `update_gitignore()` runs only inside `loom-daemon init` (invoked by `scripts/install-loom.sh:1121` during full install). The lighter refresh path `defaults/scripts/resync-installed.sh` (#4239) copies hooks/scripts/roles/docs but never touches `.gitignore`, so a consumer's marker-delimited block stays frozen at whatever pattern set the installing binary knew.
[...]

PR #4683, opened 2026-07-30
fix(resync): stage installed copies and rename them into place

### LINKED ISSUE #4669: resync-installed.sh: make self-update safe while the script is running
## Context

While processing issues in `rjwalters/kicad-tools`, the installed Loom checkout was refreshed with:

```bash
./.loom/scripts/resync-installed.sh
```

The command partially updated dozens of installed files, updated its own on-disk `.loom/scripts/resync-installed.sh` during execution, and then failed:

```
./.loom/scripts/resync-installed.sh: line 389: syntax error near unexpected token `(`
```

The final installed script and source script both pass `bash -n`, and line 389 is valid in the completed file. The source and installed versions differed substantially. This strongly indicates the running Bash process observed a partially rewritten version of its own script after the resync copied the newer script over the path it was still reading.

The failure left a partial refresh that had to be isolated in a recovery stash before work could continue.

- [ ] `resync-installed.sh` never truncates or replaces the script file from which its current process is still reading.
- [ ] Stage a newer self-copy atomically and either defer installing it until every other surface has settled or restart cleanly from the staged copy.
- [ ] A self-update failure cannot silently leave unrelated installed surfaces partially refreshed; report a clear recovery state or provide transactional/rollback behavior.
- [ ] Add a regression test that runs an older resync script against a newer source containing substantial changes to `resync-installed.sh` itself.
- [ ] Verify dry-run and apply mode remain idempotent on macOS Bash 3.2 and current Bash.

rjwalters:
## Summary

`resync-installed.sh` syncs every file under `defaults/scripts/` — **including itself** — with an in-place `cp`, which truncates and rewrites the destination. When that destination is the very path the running bash process is still reading from, bash resumes at a now-meaningless byte offset in the newer, differently-sized file. The reporter saw `syntax error near unexpected token \`(\`` at line 389 (the `cp` call itself); in my reproduction the run instead falls silently off the end. Either way the run stops right after the scripts walk, leaving the roles/docs/bin/commands surfaces stale, the metadata unstamped, and no summary saying anything went wrong.

Reproduced against `origin/main` (padded "older" installed copy resyncing itself from the newer source):

```
Resyncing .loom/hooks/ ...
  updated   hooks/guard.sh
Resyncing .loom/scripts/ ...
  created   scripts/lib/bar.sh
  updated   scripts/resync-installed.sh
RC=0                                   <- run ends here, silently
--- roles/builder.md: ROLE-OLD         <- never refreshed
```

## Changes

- **`defaults/scripts/resync-installed.sh`**
  - `sync_one()` no longer writes the destination in place. Each copy is staged as `.resync-stage.XXXXXX` **beside** its destination and `mv`-ed into place: `rename(2)` is atomic within a filesystem and swaps the *directory entry*, so an already-open inode (this script's own) stays intact and no reader can ever observe a partial file. This also removes the pre-existing "killed mid-`cp` leaves a truncated installed file" hazard for every surface, not just the self case.
  - Permissions are preserved across the swap: `mktemp` creates the staging file `0600`, so the destination's existing mode (or the source's, for a newly created file) is restored before the rename, then the source's executable bit is honored exactly as before.
  - The self-copy is additionally **deferred**: the scripts walk records it (`SELF_PATH` vs. the resolved destination, behind a cheap basename pre-filter) and `apply_deferred_self_sync()` installs it *after every other surface has settled*.
  - A file that cannot be staged or renamed is now counted as a **failure**. The run finishes the remaining files and then prints an explicit `PARTIAL REFRESH` block naming every failed path, and **exits 1**, instead of folding a partial state into a success summary (the resync-only-dirt commit hint is suppressed in that state too). An `EXIT`/`INT`/`TERM`/`HUP` trap removes any in-flight staging file, so a killed run leaves no `.resync-stage.*` dirt behind.
  - Header comment + exit-code documentation updated (`--help` is derived from it).
- **`defaults/scripts/tests/test-resync-installed.sh`** — three new groups (17 assertions), described below.
- **`defaults/docs/troubleshooting.md`** — documents the self-update safety property and the new `PARTIAL REFRESH` / exit-1 state in the resync section.

- [x] **Never truncates/replaces the script the current process is reading from** — every write is `mktemp` + `mv`; the running process keeps its original inode. Test group 19 asserts an updated file gets a **new inode** (fails against pre-fix `cp`).
- [x] **Stage a newer self-copy atomically and defer installing it** — both: atomic staging for all files, plus `apply_deferred_self_sync()` after the surface walks. Test group 18 asserts the self-copy's report line comes *after* the last other-surface line. (Re-exec was deliberately not used: with rename semantics the running process is already safe, and a re-exec adds a restart/loop failure mode for no gain.)
- [x] **A self-update failure cannot silently leave unrelated surfaces partially refreshed** — failures are counted, named, and exit 1 with a stated recovery action; the deferred self-sync additionally prints an explicit "this install is now MIXED / recover with `cp <src> <dst>`" message. Test group 20.
- [x] **Regression test running an older resync script against a newer source with substantial changes to `resync-installed.sh` itself** — test group 18 installs the **real** script padded with 6000 lines as the running copy and resyncs it from the unpadded source. 8 of the 17 new assertions **fail against the pre-fix script** (verified by running the new suite against `git show origin/main:defaults/scripts/resync-installed.sh`).
- [x] **Dry-run and apply remain idempotent** — existing groups 2/4/7 plus new assertions: `--dry-run` previews the self-update (exit 2) and leaves the running script byte-identical; rerunning the freshly-updated installed copy reports "Already in sync". Verified on Bash 5.2 (this host) and at real scale (253-file full `defaults/` tree: 3 updated → rerun clean no-op, 0.5s, no staging dirt, mode `-rwxrwxr-x` preserved).
  - **Bash 3.2**: no 3.2 interpreter is available on this host, so this was verified by construction rather than execution — the new code uses only 3.2-safe constructs (`arr+=(x)` is 3.1+, `${BASH_SOURCE[0]:-}`, `${var##*/}`, `trap`), and the one `set -u` array hazard (`"${FAILED_RELS[@]}"` on an empty array, which errors in bash < 4.4) is guarded behind `N_FAILED -gt 0`. `mktemp <template-with-6-X>`, `stat -c` → `stat -f '%OLp'` fallback, and `mv -f` cover BSD/macOS userland. Flagging explicitly for the Judge.

## Test Plan

- [x] `bash -n` + `shellcheck --severity=warning` clean on both touched scripts.
- [x] `bash defaults/scripts/tests/test-resync-installed.sh` → **105/105 passed** (was 88; +17).
- [x] Same suite run against the **pre-fix** script → 97/105, **8 failures**, confirming the tests actually capture the bug.
- [x] Full-scale manual smoke test in a scratch repo with the real `defaults/` tree (self-update applied last, run completes through metadata re-stamp + `.gitignore` refresh + summary, idempotent rerun, `--dry-run` exit 0 when in sync).
- [x] `bash scripts/test-installer.sh` → 175 passed, **1 pre-existing failure** in `test-loom-dispatcher.sh` ("config-selected codex path does not touch the claude runner"), which reproduces identically on a pristine `origin/main` checkout (96/99 there) — unrelated to this change, not fixed here per scope discipline.

Closes #4669

### DIFF
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -720,6 +720,21 @@ export `LOOM_RESYNC_ALLOW_WORKTREE=1`); it then proceeds with a warning naming t
 main-checkout target. Running from the main checkout — including any subdirectory
 of it — is unaffected.
 
+**It can safely update itself (#4669).** `resync-installed.sh` is one of the files
+under `defaults/scripts/`, so every run copies a newer version over the very path
+the running Bash process is still reading from. It used to do that with an
+in-place `cp`, which truncates and rewrites the destination: Bash then resumed
+reading its own (now shorter) script at a stale byte offset and either died with
+`syntax error near unexpected token` or fell off the end mid-run — leaving dozens
+of surfaces refreshed, the rest stale, and no summary saying so. Now every file
+is staged beside its destination and `rename(2)`-d into place (atomic; the
+already-open inode is left intact), and the self-copy is additionally **deferred
+until every other surface has settled**. If a file cannot be staged or renamed
+(read-only mount, permissions, no disk space) the run still finishes the
+remaining files and then prints an explicit **`PARTIAL REFRESH`** block naming
+every failed path and **exits `1`** — nothing is ever half-written, so fixing the
+cause and re-running completes the refresh.
+
 The intended flow is **"freshness warning says you're stale → run resync"**:
 
 ```bash
--- a/defaults/scripts/resync-installed.sh
+++ b/defaults/scripts/resync-installed.sh
@@ -60,6 +60,25 @@
 # (never a reason to skip the gate) but does not commit on the operator's
 # behalf.
 #
+# ATOMIC WRITES + DEFERRED SELF-UPDATE (#4669): every file is installed by
+# staging a copy NEXT TO its destination and rename(2)-ing it into place, never
+# by truncating and rewriting the destination in place. This matters most for
+# THIS script: resync-installed.sh is itself a file under defaults/scripts/, so
+# a resync copies it over the very path the running bash process is still
+# reading from. The old in-place `cp` let bash resume reading a half-rewritten
+# file at a now-meaningless byte offset — the reported `syntax error near
+# unexpected token` mid-run, which aborted the run and left dozens of surfaces
+# partially refreshed. A rename swaps the directory entry and leaves the
+# already-open inode intact, so the running process keeps reading the exact
+# bytes it started with. Belt-and-suspenders, the self-copy is also DEFERRED:
+# it is applied only after every other surface has settled.
+#
+# A file that cannot be staged or renamed is counted as a FAILURE: the run
+# still finishes the remaining files, then prints an explicit PARTIAL summary
+# naming every failed path and exits 1. No file is ever left half-written
+# (staging happens off to the side), so re-running after fixing the cause
+# completes the refresh — a partial refresh is never silent.
+#
 # EXPLICITLY OUT OF SCOPE (never touched by resync — updated by other mechanisms):
 #   .loom/config.json       - operator-owned; needs merge-semantics design
 #   CLAUDE.md               - repo-customized at install; needs managed-section markers
@@ -114,8 +133,9 @@
 #
 # Exit codes:
 #   0 - Success. Sync applied (or already in sync); or --dry-run found no drift.
-#   1 - Error (not in a git repo, the source tree could not be located, or
-#       invoked from a linked worktree without --allow-worktree).
+#   1 - Error (not in a git repo, the source tree could not be located,
+#       invoked from a linked worktree without --allow-worktree, or one or more
+#       files could not be synced — see the PARTIAL summary block, #4669).
 #   2 - --dry-run only: drift detected (one or more files WOULD be updated).
 #       Lets callers (e.g. the #3770 warning) use --dry-run as a cheap check.
 #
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-30T21:30:45Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verification performed (independent of the Builder's claims):**

1. **Read the full diff.** The fix replaces the in-place `cp "$src" "$dst"` with `mktemp "$dst_dir/.resync-stage.XXXXXX"` (same directory/filesystem as the destination) + `cp` into the staging file + `chmod` to restore the destination's mode (or the source's for new files) + executable bit + `mv -f` into place. Confirmed via `grep -n "cp "` that no other codepath in the script writes the live destination directly anymore — the only remaining `cp` targets `$STAGED_TMP`.
2. **Self-update ordering.** `sync_one()` defers the self-path (matched via `abs_file_path` against `BASH_SOURCE[0]`, with a basename pre-filter) by returning early and stashing `SELF_SRC/DST/REL`; `apply_deferred_self_sync()` runs once, after the last `resync_tree` call, temporarily clearing `DEFER_SELF` so the deferred call actually performs the sync instead of re-deferring. This is a single, unconditionally-ordered call site, so the self-copy can't be applied twice or skipped.
3. **`set -u` array safety.** Confirmed `set -uo pipefail` is active and the only `"${FAILED_RELS[@]}"` expansion is inside the `[[ "$N_FAILED" -gt 0 ]]` guard, so an empty array is never expanded under `set -u` (the flagged bash-3.2 concern doesn't materialize here).
4. **Trap-based cleanup.** `STAGED_TMP` is cleared in `cleanup_staged_tmp` after every successful `mv`, so the EXIT/INT/TERM/HUP traps only ever remove a stray in-flight staging file, never a live installed file.
5. **Regression test actually distinguishes old vs. new behavior — verified independently, not just inspected:**
   - Ran the new suite (`defaults/scripts/tests/test-resync-installed.sh`) as-is in the builder worktree: **105/105 passed**.
   - Built an isolated fixture (`git show origin/main:defaults/scripts/resync-installed.sh` as the script under test, same new test file) and re-ran: **11 failures**. Of those, 3 (`#4280` ignore-path / missing-binary-warning assertions) are an artifact of my `/tmp` isolation harness — they fail identically when I ran the *new* fixed script through the same harness, confirming they're unrelated to this change. The remaining **8 failures are exactly the new `#4669` self-update/atomicity/partial-refresh assertions (groups 18-20)** — matching the PR description's claim precisely.
6. **Pre-existing unrelated failure** (`test-loom-dispatcher.sh`, "config-selected codex path does not touch the claude runner") — reproduced independently on pristine `origin/main` (96/99 passed, identical failure), confirming it's out of scope for this PR.
7. **CI**: all checks green (`Shell Test Suites (hermetic)` finished passing after an initial pending state), `mergeStateStatus` is `CLEAN`.

**Minor, non-blocking observations** (not requesting changes for these):
- `abs_file_path()`'s helper `abs_path()` only resolves a directory that already exists (`[[ -d "$p" ]]` short-circuits to the raw string otherwise); if the self-path's destination directory somehow didn't exist yet the self-path comparison could silently fail to match. In practice `.loom/scripts/` always pre-exists, so this is theoretical.
- If the self-file happens to be pinned in `.loom/resync-ignore`, the "skipped" report line is deferred to the end of the run rather than printed in the natural walk order (cosmetic only — `N_SKIPPED` still increments correctly via the second `sync_one` call inside `apply_deferred_self_sync`).

Nice, thorough fix — the atomic-rename approach is the right one (avoids the self-truncation hazard for every file, not just the self-referential case), and the new regression coverage genuinely proves the bug and the fix.

Issue #4704, opened 2026-07-31
[Epic #4702] Durable sweep outcome records (absorbs #4137)


rjwalters:
**Epic**: #4702 - Rich fleet observability dashboard with user-configurable hosting
**Phase**: 1 of 4
**Phase Goal**: The daemon durably records and can export everything worth charting. No cloud dependency; ships value standalone.

## Description

Persist a durable outcome record for every sweep the daemon runs: model/effort/config used, per-phase timings, result (merged / failed / cancelled / gate-blocked), issue and PR numbers, token account used. This absorbs open issue #4137 ("durable sweep outcome telemetry — model/config choices are currently unmeasurable") into the epic — close #4137 as superseded when this lands, with a cross-reference comment.

Implementation context:

- `loom-daemon/src/sweep_outcomes.rs` and `sweep_journal.rs` already exist — start by auditing what they capture and persist today vs. what the schema issue's `sweep.outcome` record needs; this may be extension rather than greenfield.
- Records use the Phase-1 telemetry schema types (sibling issue) and persist locally (append-only JSONL under `.loom/` state or the daemon's existing state dir — follow the existing persistence pattern in the codebase) regardless of whether an exporter is configured. Local durability is the point: history must survive daemon restarts and outlive any cloud backend.
- Include a small read surface (`loom-daemon` subcommand or IPC query) so outcomes are inspectable locally before the Phase-3 UI exists — this is what makes the phase ship standalone value.
- Retention: bounded (configurable cap or age-based pruning) so the file cannot grow unbounded on a long-lived host.

## Acceptance Criteria

- [ ] Every completed sweep (success, failure, or cancellation) leaves a durable outcome record with model, config, per-phase durations, result, and refs
- [ ] Records survive daemon restart; bounded retention enforced
- [ ] Local inspection path exists (subcommand or IPC query) and is documented
- [ ] #4137 closed as superseded with cross-reference once merged
- [ ] Tests pass for new functionality

## Dependencies

Part of Epic #4702, Phase 1. Consumes the telemetry schema types from the sibling schema issue (merge after it, or land shared types first).

---
*Created by Epic skill from Epic #4702*

rjwalters:
**Champion Review: APPROVED**

This issue has been evaluated and promoted to `loom:issue` status. All quality criteria passed:

- Clear problem statement — durable per-sweep outcome persistence, explicitly absorbing open #4137
[...]

Issue #4702, opened 2026-07-31
Epic: Rich fleet observability dashboard with user-configurable hosting


rjwalters:
# Epic: Rich fleet observability dashboard with user-configurable hosting

## Overview

Loom's fleet visibility today is the phase-1 `loom-daemon serve` dashboard (#4329): loopback-only, read-only, live-only (no history), with client-side `--peers` fan-out. That was deliberately minimal. This epic builds the rich successor: a cloud-hosted, multi-host fleet dashboard with historical charts, live event feed, per-sweep drill-down, and token/cost analytics — deployable to a **user-configurable hosting solution**. The 2AM fleet's reference deployment will run on Cloudflare at **dashboard.2amlogic.com**.

Architecture decisions (from operator interview, 2026-07-30):

- **Push model**: daemons push telemetry to a cloud backend (Workers + D1 + Durable Object). Fleet hosts sleep/idle-shutdown (#4697), so pull/tunnel approaches go blank; push with offline queueing gives history and visibility even when hosts are down. Builds toward the safehouse envelope-feed direction (#4699) without depending on it.
- **User-configurable hosting, two ways**: (1) the reference backend is a wrangler-templated Workers app anyone can deploy to their own Cloudflare account (or adapt elsewhere) with the daemon pointed at it via an `observability` config block; (2) an **optional OTLP exporter** (Phase 4) lets users with an existing OpenTelemetry stack (Grafana, Honeycomb, self-hosted collector) skip our backend entirely. Native JSON push is the first-class path; OTel is an escape hatch, not the core dependency.
- **Read-only**: observability only. Steering (cancel/dispatch/pause) stays with MCP and the safehouse direction (#4196); a later epic may add authenticated controls.
- **Access**: authenticated view gated by Cloudflare Access (zero-trust SSO, no auth code in the dashboard itself). Plus a **public view** showing full detail for public-repository work and only redacted/summarized aggregates for private-repo work (no repo names, issue titles, branch names, or PR links).

## Milestone Alignment

**Current Milestone**: none declared (README has no milestone marker)
**Alignment Tier**: Tier 2 - Goal-Supporting
**Justification**: Continues the approved fleet-visibility direction (#4008 codecast evaluation → #4329 phase-1 dashboard). Absorbs open #4137 (durable sweep outcome telemetry — "model/config choices are currently unmeasurable") as its data foundation. Makes multi-host autonomous operation measurable and observable, which every goal-advancing initiative depends on for tuning.

[...]

Issue #4705, opened 2026-07-31
[Epic #4702] Pluggable telemetry exporter with offline queueing + observability config block


rjwalters:
**Epic**: #4702 - Rich fleet observability dashboard with user-configurable hosting
**Phase**: 1 of 4
**Phase Goal**: The daemon durably records and can export everything worth charting. No cloud dependency; ships value standalone.

## Description

Add the exporter layer that pushes Phase-1 telemetry records to a configured sink, plus the `observability` config block in `.loom/config.json`. This is the daemon half of the push architecture; the Phase-2 Workers backend is the first real sink.

Design constraints:

- **Exporter trait + one implementation.** A small Rust trait (emit batch of records, flush) with a native JSON-over-HTTPS push implementation (POST batches to `endpoint` with `Authorization: Bearer <ingest key>`). The trait boundary is what makes Phase 4's OTLP exporter a drop-in later — do not add any OTel dependency now.
- **Buffered + offline-tolerant.** Fleet hosts sleep and idle-shutdown (#4697). Records queue durably when the sink is unreachable and drain on reconnect, with bounded queue size (drop-oldest with a logged counter, never unbounded growth) and jittered retry/backoff.
- **Config block**: `observability: { enabled, endpoint, ingestKeyFile, batchSize, flushIntervalSecs, queueCapacity }` following the daemon's existing config pattern — precedence env (`LOOM_OBSERVABILITY_*`) > config > default, **off by default** (FLAGS-OFF posture, matching how `autonomous.*` blocks work; see `config_resolver.rs`). Ingest key read from a file path, never inline in committed config.
- **Read-only invariant**: the exporter only ships data out; nothing arrives back over this channel.
- Test against a local mock sink (tiny HTTP listener in tests); no Cloudflare dependency in this issue.

## Acceptance Criteria

- [ ] Exporter trait + HTTPS-push implementation, wired to emit sweep lifecycle, outcome, token-snapshot, and host-health records
- [ ] Durable offline queue with bounded capacity, drop-oldest accounting, and retry with backoff — verified by a test that kills and revives the mock sink
- [ ] `observability` config block with env > config > default precedence, off by default; documented in the daemon reference
- [ ] Ingest key sourced from file, absent from logs and error messages
- [ ] Tests pass for new functionality

## Dependencies

Part of Epic #4702, Phase 1. Consumes the schema issue's record types; pairs with the outcome-records issue (exports what it persists). No cloud dependency.

---
[...]

PR #4707, opened 2026-07-31
feat(daemon): versioned fleet telemetry schema with repo-visibility tagging (#4703)

### LINKED ISSUE #4703: [Epic #4702] Versioned fleet telemetry schema with repo-visibility tagging
**Phase**: 1 of 4
**Phase Goal**: The daemon durably records and can export everything worth charting. No cloud dependency; ships value standalone.

## Description

Define the versioned telemetry record schema that the whole epic builds on: sweep lifecycle events, sweep outcome records, token-pool snapshots, and host health. This is a schema + serialization deliverable in the daemon (`loom-daemon/src/types.rs` neighborhood, or a new `telemetry` module), not an exporter — export/push lands in the sibling exporter issue.

Design constraints:

- **Extend, don't fork, the existing event surface.** The six frozen SSE `sweep.*` topics (#4392, `loom-daemon/src/serve.rs` / `event_bus.rs`) are the starting vocabulary. Telemetry records should be a superset: the same lifecycle moments plus outcome/config metadata the SSE tail doesn't carry.
- **Versioned envelope.** Every record carries `schema_version` so the Phase-2 backend can accept mixed-version fleets during rolling daemon updates.
- **Repo-visibility tag.** Every record referencing a repository carries `visibility: public | private` (derivable at emit time via `gh api repos/{owner}/{repo} --jq .private`, cached). This is load-bearing for the epic's public view: Phase 2 redaction keys off this tag, so it must be present from the first record ever emitted.
- **Record kinds** (minimum): `sweep.started/phase/completed` lifecycle, `sweep.outcome` (model, config, effort, per-phase durations, result, PR number), `tokens.snapshot` (per-account usage/limit-window state, matching what `loom-daemon tokens check --ranking` knows), `host.health` (CPU/disk headroom, daemon version, uptime — see `cpu_headroom.rs`, `disk_headroom.rs`).
- Rust structs with serde, JSON on the wire. Document the schema in a markdown file alongside the code so the Phase-2 Workers backend (TypeScript) has a reference independent of the Rust types.

## Acceptance Criteria

- [ ] Rust types for all record kinds above, serde-serializable, with `schema_version` and repo-visibility tagging
- [ ] Schema documented in a markdown reference (wire-format examples per record kind)
- [ ] Visibility derivation helper with caching (no per-record forge API call)
- [ ] Unit tests covering serialization round-trips and the visibility helper
- [ ] Tests pass for new functionality

## Dependencies

Part of Epic #4702. This is a Phase 1 issue with no blocking dependencies. The other two Phase-1 issues (outcome persistence, exporter) consume these types — coordinate merge order but keep PRs separate.

---
*Created by Epic skill from Epic #4702*

---

## Curator Enhancement

**Why `complex`, not `routine`**: this is schema-design work two sibling Phase-1 issues (#4704, #4705) build against — a wrong shape here (e.g. an under-specified visibility default, a versioning scheme that can't tolerate mixed-fleet rolling upgrades) would compile, pass unit tests, and pass review, but only surface as a data-leak or a breaking change months later when Phase 2's redaction logic or a rolling daemon update hits it. That "mistake could pass tests and review unnoticed" plus the security/privacy stakes (repo-visibility tagging is explicitly the epic's anti-leak control) is the `complex` bar per the curator complexity rubric, not just LOC/file count.

### Verification against the codebase (2026-07-30)

- **Frozen SSE `sweep.*` topic count — confirmed correct as written.** `loom-daemon/src/event_bus.rs` (module doc, "Topic taxonomy (frozen for v0.10.0)") and `loom-daemon/src/serve.rs` (`SSE_TOPIC_PREFIXES`, "the six frozen `sweep.*` topics MUST stream") both independently state six: `sweep.issue.{N}.phase`, `.blocker`, `.exited`, `.crashed`, `sweep.global.dispatch`, `sweep.global.completed`. (`sweep.global.runtime_rejected` was added later, #4494, and is *not* one of the frozen six or in the SSE bridge's allow-list — treat it as optional context, not part of the "extend the frozen six" baseline.) No correction needed to the issue body.
- **No existing telemetry/schema-version/visibility module** — `loom-daemon/src/` has no `telemetry.rs` or `telemetry/` directory today (checked full directory listing). This is a net-new module, not an extension of an existing one.
- **`loom-daemon/src/sweep_outcomes.rs`** (`OutcomeRecord`, 334 lines) is the closest existing precedent for a durable JSONL record type (serde round-trip pattern, `#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]`) but it's owned by sibling issue #4704 (which persists using *this* issue's types) — do not extend `OutcomeRecord` itself here; define a distinct `SweepOutcomeRecord` (or similar) in the new telemetry module that #4704's persistence layer then consumes/maps to its journal.
- **No existing repo-visibility helper** — grepped for `.private`/`is_private`/`repos/{owner}/{repo}` `gh api` call sites; none derive or cache repository visibility today. `sweep_registry.rs` has several `gh api repos/{owner}/{repo}/issues/{N}` call sites for other purposes (pattern reference for how this codebase shells out to `gh api`, not a reusable visibility helper).
- **Caching precedent to follow**: `loom-daemon/src/cpu_headroom.rs` already implements exactly the "avoid one API/subprocess call per record" shape this issue needs — `CPU_UTIL_STATE: Mutex<CpuUtilState>` + `refresh_cpu_util_cache()` (TTL-gated refresh, no-op within `CPU_UTIL_MEMO_TTL`) + `cached_cpu_idle_fraction()` (pure, non-blocking read). Model the visibility helper the same way: a `Mutex`-guarded per-repo cache keyed on `owner/repo`, a refresh function that shells out to `gh api repos/{owner}/{repo} --jq .private` only when the cached entry is absent or stale, and a pure cache-read accessor for the hot path.

### Implementation Guidance

- Define the versioned envelope and all record kinds in a single new module, `loom-daemon/src/telemetry.rs` (or `loom-daemon/src/telemetry/` if the five record kinds + visibility helper make one file unwieldy — `mod.rs` + `visibility.rs` submodule is a reasonable split).
- `schema_version` should be a plain `u32` (or a `const CURRENT_SCHEMA_VERSION: u32`) on every record struct, not a semver string — matches "accept mixed-version fleets" (a numeric compare is simpler for the Phase-2 TypeScript backend to gate on than semver parsing).
- `RepoVisibility` should be a two-variant enum (`Public`, `Private`) with a **private-safe deserialization default** — an unknown/missing/malformed value on the wire must decode to `Private`, never `Public`, so a partial or older-schema record can never accidentally qualify for the epic's redaction-sensitive public view. This is worth a dedicated unit test (see Test Plan).
- Record kinds to define (per the issue body's "Record kinds (minimum)"): lifecycle records mirroring the frozen `sweep.*` topics (`SweepStarted`/`SweepPhase`/`SweepCompleted`, or one tagged enum with a `kind` discriminant — either works, pick whichever keeps the wire format simplest for the TS backend to pattern-match on), `SweepOutcomeRecord`, `TokenSnapshotRecord`, `HostHealthRecord`.
- This issue is schema + serialization + the visibility helper only — no wiring into `event_bus.rs`, no new bus topics, no persistence, no exporter. Sibling issues #4704 (persistence) and #4705 (export) consume these types; keep the module's public surface (`pub struct`/`pub enum` + one visibility-derivation function) narrow so both can depend on it without depending on daemon internals this issue doesn't own.

## Affected Files

- `loom-daemon/src/telemetry.rs` (new) — versioned envelope (`schema_version`), all record-kind structs/enums (`sweep.started/phase/completed` lifecycle, `SweepOutcomeRecord`, `TokenSnapshotRecord`, `HostHealthRecord`), and the `RepoVisibility` enum + tagging field, all serde-serializable.
- `loom-daemon/src/telemetry/visibility.rs` (new, if split out) — the caching visibility-derivation helper (`gh api repos/{owner}/{repo} --jq .private`), following the TTL-memoized `Mutex`-guarded cache pattern in `loom-daemon/src/cpu_headroom.rs` (`CPU_UTIL_STATE` / `refresh_cpu_util_cache` / `cached_cpu_idle_fraction`).
- `loom-daemon/src/lib.rs` — add `pub mod telemetry;` (alongside the existing alphabetized `pub mod` list, currently `...sweep_registry; pub mod terminal;...` at the `t` boundary).
- `.loom/docs/telemetry-schema.md` (new) — wire-format reference: one JSON example per record kind, `schema_version` semantics, and the `RepoVisibility` contract, written so the Phase-2 Workers/TypeScript backend has a reference independent of the Rust types (per the issue body's explicit requirement).
- `loom-daemon/src/telemetry.rs` (or `loom-daemon/src/telemetry/visibility.rs`) `#[cfg(test)] mod tests` — inline unit tests (serde round-trips, visibility-cache behavior), matching the existing inline-test convention used throughout `loom-daemon/src/` (e.g. `sweep_outcomes.rs`, `cpu_headroom.rs`) rather than a separate `loom-daemon/tests/` integration file.

## Test Plan

- [ ] Automated: serde `Serialize`/`Deserialize` round-trip test for every record kind (each lifecycle variant, outcome, token-snapshot, host-health) — encode to JSON, decode back, assert equality (`PartialEq` derive), mirroring the round-trip test pattern already used for `OutcomeRecord` in `sweep_outcomes.rs`.
- [ ] Automated: `schema_version` is present on every serialized record and matches `CURRENT_SCHEMA_VERSION` for freshly-constructed records.
- [ ] Automated: an unknown/missing/malformed `visibility` value deserializes to `Private` (never `Public`) — the redaction-safety property the epic body calls "load-bearing" for the public view.
- [ ] Automated: visibility-helper caching — a second lookup for the same `owner/repo` within the TTL window does not re-invoke `gh api` (test the refresh/read split independently, e.g. by injecting a call counter or fake data source the way `cpu_headroom.rs`'s tests isolate `refresh_cpu_util_cache` from a real subprocess).
- [ ] Manual verification: hand-construct one JSON example per record kind matching `.loom/docs/telemetry-schema.md` and confirm it round-trips through the Rust types via a quick `cargo test -p loom-daemon telemetry::` run.
- [ ] N/A: no event-bus wiring, persistence, or exporter integration in this issue's scope (covered by #4704 and #4705 respectively) — `cargo build -p loom-daemon` and `cargo clippy -p loom-daemon` passing cleanly for the new module is sufficient beyond the unit tests above.

## Related Open Work

`check-duplicate.sh --issue 4703` surfaced `RELATED_OPEN_WORK: #4702` (the parent epic, which cross-references this phase issue in its own body) and low-similarity `DUPLICATE_FOUND` hits on the two sibling Phase-1 issues (#4704 at 23%, #4705 at 22%) plus the epic itself (19%). **Disregarded** — #4702 is the expected parent-tracking cross-reference (not a spec disagreement to reconcile), and #4704/#4705 are sibling phase issues created together from the same epic decomposition, not duplicates of this issue's schema-design scope; the similarity is topical overlap (all three mention "telemetry"/"Phase 1"), not competing specs for the same deliverable.

rjwalters:
Closes #4703

## Summary

Phase-1 telemetry-schema foundation for the fleet observability epic. Adds a new `loom-daemon/src/telemetry/` module (schema + serialization + the visibility helper only — no event-bus wiring, no persistence, no exporter):

- **`telemetry/mod.rs`** — a versioned `TelemetryEnvelope` (`schema_version: u32` = `CURRENT_SCHEMA_VERSION`, `emitted_at`, `host_id`, `record`) wrapping a `kind`-tagged `TelemetryRecord` enum. Record kinds: `sweep.started` / `sweep.phase` / `sweep.completed` lifecycle (mirroring the frozen SSE `sweep.*` topics), `sweep.outcome` (model/config/effort/per-phase durations/result/PR number — a distinct type from `sweep_outcomes::OutcomeRecord`, which #4704 maps into its journal), `tokens.snapshot`, and `host.health`.
- **`RepoVisibility` enum (Public/Private), private by default.** `Default` is `Private`, and a hand-written `Deserialize` maps *only* an exact case-insensitive `"public"` to `Public` — any unknown label, `null`, wrong-typed scalar, or nested map/seq decodes to `Private` (never errors, never leaks). Repo-referencing records tag `visibility` with `#[serde(default)]` so a missing tag also defaults to `Private`.
- **`telemetry/visibility.rs`** — TTL-cached visibility derivation modeled on `cpu_headroom.rs`'s memoization (`refresh` + pure `cached_` read split), backed by `gh api repos/{owner}/{repo} --jq .private`. Per-`owner/repo` `Mutex`-guarded cache (via `OnceLock`, since `HashMap::new()` isn't `const`); a probe failure resolves to `Private`.
- **Wire-schema doc** at `defaults/docs/telemetry-schema.md` (mirrored to `.loom/docs/`) with a JSON example per record kind, `schema_version` semantics, and the `RepoVisibility` contract — a reference for the Phase-2 TypeScript backend independent of the Rust types.

### Design decisions

- **Envelope shape**: `schema_version` lives on the envelope (a plain `u32`, not semver, so a mixed-version fleet's backend gates on a numeric compare during rolling daemon upgrades). Every emitted record is always wrapped in an envelope, so `schema_version` is present on every record on the wire. The `record` payload is internally tagged on `kind`, flattening to a single object the TS backend can pattern-match on.
- **Visibility default**: private-by-default is enforced in *two* independent places — `Default::default()` (missing field via `#[serde(default)]`) and the custom `Deserialize` (present-but-unknown/malformed value) — plus the derivation helper's `unwrap_or(Private)` on probe failure. Leak-by-default is impossible by construction, not convention.

## Test evidence

`cargo test -p loom-daemon --lib telemetry` — **15 passed, 0 failed**:
- serde round-trip for every record kind (bare and envelope-wrapped)
- `schema_version` present and equal to `CURRENT_SCHEMA_VERSION` on fresh envelopes
- unknown / missing / malformed `visibility` (string label, null, bool, number, array, object) decodes to `Private`, never `Public`
- visibility cache-hit: a second lookup within the TTL does not re-invoke the probe (call-counter seam); a `None` probe is not cached and re-probes; uncached derive falls back to `Private`
- pure `parse_gh_private` mapping

`cargo build -p loom-daemon` and `cargo clippy -p loom-daemon --all-targets` both clean (no new warnings).

Part of epic #4702.

### DIFF
--- a/.loom/docs/telemetry-schema.md
+++ b/.loom/docs/telemetry-schema.md
@@ -1,0 +1,216 @@
+# Fleet Telemetry Schema (wire format)
+
+> Epic #4702, Phase 1 — the versioned telemetry record schema the fleet
+> observability pipeline is built on. Defined in Rust in
+> `loom-daemon/src/telemetry/` (`mod.rs` — record kinds + envelope;
+> `visibility.rs` — the repo-visibility derivation helper). This document is the
+> **format-independent reference** so the Phase-2 Workers/TypeScript backend can
+> parse the wire format without the Rust types.
+
+This is a **schema + serialization** deliverable only — the daemon does not yet
+emit, persist, or export these records here. Sibling Phase-1 issues consume these
+types: #4704 persists them to a local journal, #4705 pushes them to a backend.
+
+## Envelope
+
+Every record is transmitted inside a versioned envelope:
+
+```json
+{
+  "schema_version": 1,
+  "emitted_at": "2026-07-30T12:00:00Z",
+  "host_id": "fleet-host-abc",
+  "record": {
+    "kind": "sweep.outcome",
+    "...": "record fields, flattened alongside `kind`"
+  }
+}
+```
+
+| Field            | Type              | Notes |
+|------------------|-------------------|-------|
+| `schema_version` | integer (`u32`)   | Current value: **1** (`CURRENT_SCHEMA_VERSION`). |
+| `emitted_at`     | RFC 3339 datetime | When the daemon produced the envelope. |
+| `host_id`        | string            | Stable identifier for the emitting host. Opaque to the schema. |
+| `record`         | object            | The record payload, internally tagged on `kind` (see below). |
+
+### `schema_version` semantics
+
+`schema_version` is a **plain integer**, not a semver string, deliberately: a
+backend ingesting a mixed-version fleet (some hosts on an older daemon mid
+rolling-upgrade) gates on a numeric compare, with no semver parsing. It is bumped
+only on a **breaking** wire change to the record shapes below. A backend should:
+
+- **accept** records at any `schema_version` it recognizes;
+- treat an **unknown (higher)** `schema_version` as forward-compatible where it
+  can (unknown fields are additive) or route it to a dead-letter path otherwise;
+- **never** silently coerce a missing `schema_version` to `0` — a record with no
+  `schema_version` is malformed.
+
+## `RepoVisibility` contract — private by default
+
+Every record that references a repository carries a `visibility` tag, either
+`"public"` or `"private"`. The Phase-2 public view exposes full detail for
+`public` work and only redacted/summarized aggregates for `private` work, so this
+tag is the schema-level anti-leak control.
+
+**The decode is private-safe by construction.** The Rust deserializer maps *only*
+the exact (case-insensitive) string `"public"` to `Public`; **everything else**
+maps to `Private`:
+
+- a missing `visibility` field ⇒ `Private`;
+- an unknown label (e.g. `"internal"`) ⇒ `Private`;
+- a `null`, a wrong-typed scalar (bool/number), or a nested array/object ⇒
+  `Private`.
+
+A partial or older-schema record can therefore **never accidentally decode to
+`public`** and leak into the public view. The Phase-2 backend MUST implement the
+same rule: default anything that is not exactly `"public"` to private. Redaction
+keys off this tag, never off client-side filtering.
+
+Visibility is derived at emit time from the forge (`gh api repos/{owner}/{repo}
+--jq .private`) and cached per `owner/repo` with a TTL, so it costs no per-record
+API call. A probe failure resolves to `private` — the same fail-safe default.
+
+## Record kinds
+
+The `record` object is internally tagged on `kind`. The tag values reuse the
+frozen SSE `sweep.*` topic vocabulary where they overlap, plus the epic's added
+kinds. Records that reference a repository carry `repo` + `visibility`; host-level
+records (`tokens.snapshot`, `host.health`) do not.
+
+### `sweep.started`
+
+A sweep began work on an issue.
+
+```json
+{
+  "kind": "sweep.started",
+  "repo": "rjwalters/loom",
+  "visibility": "public",
+  "issue": 4703,
+  "sweep_id": "sweep-issue-4703-0",
+  "started_at": "2026-07-30T12:00:00Z",
+  "model": "opus",
+  "effort": "high"
+}
+```
+
+`model` and `effort` are omitted when unset (empty-means-unset, mirroring
+`SweepInfo`).
+
+### `sweep.phase`
+
+A sweep advanced to a new lifecycle phase (mirrors `sweep.issue.{N}.phase`).
+
+```json
+{
+  "kind": "sweep.phase",
+  "repo": "rjwalters/loom",
+  "visibility": "public",
+  "issue": 4703,
+  "sweep_id": "sweep-issue-4703-0",
+  "phase": "builder",
+  "entered_at": "2026-07-30T12:03:20Z"
+}
+```
+
+`phase` is a lifecycle name: `curator`, `builder`, `judge`, `doctor`, `merge`.
+
+### `sweep.completed`
+
+A sweep reached a terminal state (the summary moment; richer detail is in the
+paired `sweep.outcome`).
+
+```json
+{
+  "kind": "sweep.completed",
+  "repo": "rjwalters/loom",
+  "visibility": "public",
+  "issue": 4703,
+  "sweep_id": "sweep-issue-4703-0",
+  "completed_at": "2026-07-30T12:08:32Z",
+  "result": "success"
+}
+```
+
+`result` is one of `success`, `failure`, `cancelled`, `blocked`.
+
+### `sweep.outcome`
+
+The full post-hoc outcome: model/config/effort, per-phase durations, terminal
+result, and PR number. (A distinct type from the daemon's internal
+`sweep_outcomes::OutcomeRecord`, which #4704 maps this into for its journal.)
+
+```json
+{
+  "kind": "sweep.outcome",
+  "repo": "rjwalters/loom",
+  "visibility": "public",
+  "issue": 4703,
+  "sweep_id": "sweep-issue-4703-0",
+  "model": "opus",
+  "effort": "high",
+  "config": { "runtime": "claude" },
+  "phase_durations": [
+    { "phase": "curator", "duration_sec": 12 },
+    { "phase": "builder", "duration_sec": 340 }
+  ],
+  "total_duration_sec": 512,
+  "result": "success",
+  "pr_number": 4710
+}
+```
+
+`config` (free-form string map), `phase_durations`, `model`, `effort`, and
+`pr_number` are omitted when empty/unset. `config` is a map — not fixed fields —
+so operator-tunable knobs can be captured without a schema bump.
+
+### `tokens.snapshot`
+
+A point-in-time view of the multi-account token pool (host-level — no `repo` /
+`visibility`). Matches what `loom-daemon tokens check --ranking` knows.
+
+```json
+{
+  "kind": "tokens.snapshot",
+  "captured_at": "2026-07-30T12:00:00Z",
+  "accounts": [
+    {
+      "account": "agent-1",
+      "rank": 0,
+      "usage_fraction": 0.42,
+      "limit_window_reset_at": "2026-07-30T18:00:00Z",
+      "exhausted": false
+    },
+    { "account": "agent-2", "exhausted": true }
+  ]
+}
+```
+
+Per account, `rank` / `usage_fraction` / `limit_window_reset_at` are omitted when
+unknown; `exhausted` is always present.
+
+### `host.health`
+
+Host CPU/disk headroom, daemon version, and uptime (host-level — no `repo` /
+`visibility`). Every measured field is optional so an unmeasurable probe stays
+absent rather than being coerced to a fake zero (the daemon's "unknown != zero"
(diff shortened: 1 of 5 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed the Phase-1 telemetry-schema foundation (issue #4703, epic #4702). This is a clean, well-scoped, well-documented schema deliverable that nails the two highest-stakes surfaces — the versioned envelope and the private-by-default visibility guarantee. Verified locally in the PR worktree:

- `cargo test -p loom-daemon --lib telemetry` — 15 passed, 0 failed
- `cargo build -p loom-daemon` — clean
- `cargo clippy -p loom-daemon --all-targets` — no warnings
- CI: all required checks green (Rust Unit Tests, Rust Linting/Build, Shell Suites, Installer), mergeStateStatus CLEAN

### Envelope / serde design — correct
- `schema_version` is a plain `u32` on the envelope (not semver), always present on the wire, with a deliberate *absence* of `#[serde(default)]` so a missing version is a visible bug rather than a silent 0 — matches the mixed-fleet rolling-upgrade requirement.
- `TelemetryRecord` is internally tagged on `kind`, flattening to a single object; tag values reuse the frozen SSE `sweep.*` vocabulary where they overlap. Both bare-record and envelope-wrapped round-trips are tested for every kind.
- `SweepOutcomeRecord` is a distinct type from `sweep_outcomes::OutcomeRecord` — no scope creep into #4704; `sweep_outcomes.rs` is untouched. No exporter/persistence/bus-wiring (#4704/#4705 territory) leaked in.

### Private-by-default guarantee — verified load-bearing property
The redaction-critical invariant is enforced in three independent layers: `Default = Private`, a hand-written `Deserialize` via `deserialize_any` that maps *only* an exact case-insensitive `"public"` to Public (every other string, null, bool, number, array, object -> Private, never an error), and `#[serde(default)]` on each `visibility` field so a missing tag also lands on Private. Tests cover missing field, unknown label, empty string, trailing-space "Public ", null, bool, number 0/1, array, and object — all decode to Private. Leak-by-default is impossible by construction.

### TTL cache — correct
Modeled on `cpu_headroom.rs` (refresh + pure cached-read split), `OnceLock<Mutex<HashMap>>` keyed on `owner/repo`, TTL-gated (300s), no per-record forge call. A failed/unparseable probe returns `None` and is NOT cached (re-probes next call), and `derive_visibility` falls back to Private on any failure. All three behaviors are unit-tested with an injected call-counter seam.

### Doc accuracy
`defaults/docs/telemetry-schema.md` and the `.loom/docs/` mirror are byte-identical, and every JSON example matches what serde actually emits (field order, lowercase visibility, `skip_serializing_if` omissions, snake_case `result`).

### Non-blocking observations (no change required)
- The visibility cache holds the single global `Mutex` across the `gh` subprocess probe, so a probe for one repo briefly serializes lookups of other repos. This intentionally mirrors `cpu_headroom`'s lock-across-iostat pattern and is negligible at a 300s TTL, but if emit-time cadence ever grows, a per-key lock or drop-lock-during-probe could be revisited in #4705.
- `RepoVisibility::Deserialize` relies on JSON being self-describing (`deserialize_any`); fine for the documented JSON wire format.

Acceptance criteria all met. Setting `loom:pr`.

Issue #3648, opened 2026-07-20
check-main-clean.sh (#3513 backstop) false-positives on pre-existing / unrelated working-tree changes


rjwalters:
**Surfaced by:** every wave of a `/sweep all` run on rjwalters/kicad-tools (2026-07-20).

## What happened
The `/sweep` builder backstop `./.loom/scripts/check-main-clean.sh` (the #3513 guard that detects a builder writing to the main worktree via repo-relative paths) exits `3` and prints a loud **"MAIN worktree is dirty — builder contamination"** warning whenever `git status` on the main worktree is non-empty — **including changes that predate the wave and have nothing to do with any builder.**

In this session the main worktree had a single pre-existing modification, `uv.lock` (a `uv run` lockfile version-string regeneration, present before the sweep even started). Every wave's backstop flagged it as contamination:

```
ERROR: MAIN worktree is dirty (uncommitted changes detected).
       This usually means a builder wrote to the main repo instead of
       its issue worktree ...
       Offending changes:
          M uv.lock
```

Each time, the orchestrator had to stop and manually confirm the dirty file was pre-existing (and unrelated to the builders) before advancing to Judge — defeating the point of an automatic backstop and training operators to ignore it (alarm fatigue → a real contamination could be waved through).

## Root cause
The check treats *any* dirty main worktree as contamination. It has no baseline: it can't distinguish "a builder just wrote here" from "this file was already dirty / is a regenerable lockfile the operator hasn't committed."

## Suggested fixes (any one helps)
1. **Baseline against the pre-wave dirty set** — snapshot `git status --porcelain` on the main worktree *before* dispatching the wave's builders; after they return, only flag files that are **newly** dirty (in the post set but not the pre set). This is the precise fix — it catches real contamination while ignoring what was already there.
2. **Ignore known-regenerable lockfiles** — at minimum, exclude `uv.lock` / `package-lock.json` / `Cargo.lock` / `poetry.lock` from the contamination check (their churn is expected and harmless).
3. **Downgrade to a diff-scoped check** — only flag main-worktree changes that overlap the files a builder in the wave actually touched.

## Impact
Recurring false alarm on every wave when the operator has any uncommitted local change (extremely common: a lockfile bump, a scratch edit). Erodes trust in the backstop.

rjwalters:
## Curator enrichment

### Root cause (confirmed)

[...]

PR #4708, opened 2026-07-31
fix(sweep): record an orchestrator PID for sweep-run liveness, not the one-shot tool shell

### LINKED ISSUE #4691: Sweep leaves main-clean-baseline-<RUN_ID>.txt files behind in .loom/sweep-checkpoint/
## Context

`/loom:sweep` writes a per-run main-clean baseline at `.loom/sweep-checkpoint/main-clean-baseline-${RUN_ID}.txt` (#3768's RUN_ID keying, which correctly stopped concurrent sweeps clobbering each other's snapshot).

Nothing appears to delete it. After a day of sweeps in one managed repo:

```
.loom/sweep-checkpoint/
  issue-13.json
  main-clean-baseline-sweep-20260730T173504Z-52271-64d92f58.txt
  main-clean-baseline-sweep-20260730T173533Z-58942-4ad5025f.txt
  main-clean-baseline-sweep-20260730T183008Z-53784-36a948de.txt
  main-clean-baseline-sweep-20260730T183019Z-55271-7f03251b.txt
  main-clean-baseline-sweep-20260730T183022Z-55614-55b55474.txt
```

All five belonged to runs that had long since exited (including crashed ones). The skill documents the file as "a per-sweep-run transient whose lifetime is this sweep invocation", so accumulating is contrary to intent.

Low severity — the files are small and gitignored — but they grow without bound on a busy host, and they make the checkpoint directory harder to read when you're trying to spot a genuine resume checkpoint (I was manually distinguishing the two while investigating an unrelated problem).

- [ ] A sweep removes its own baseline on completion, alongside the existing `sweep-run-registry.sh cleanup "$RUN_ID"` step.
- [ ] Baselines belonging to dead runs are pruned opportunistically — the peer scan already does a dead-PID liveness check and could reuse it — so a crashed sweep doesn't leak one permanently.
- [ ] `loom-clean` covers them as a backstop.

**Follow-up comment (rjwalters, live evidence, 2026-07-31 — see full text via `gh issue view 4691 --comments`):** a concurrent 3-sweep run in one repo showed the baseline can also be **deleted out from under a still-running sweep**, not just leaked. `check-main-clean.sh` degraded safely to the whole-status hard-fail when its baseline went missing, so no incorrect merge resulted this time — but that degradation silently defeats the #3648 contamination-subtraction backstop's purpose, since pre-existing dirt can no longer be subtracted once the baseline is gone.

---

### Status of the original three acceptance criteria: already implemented

Between this issue being filed and now, `sweep-run-registry.sh` and `loom-daemon clean` already ship all three originally-requested behaviors (landed in #4520, `c6e79db4`, merged 2026-07-30 — the code's inline comments still cite the older `#4450` as design rationale, but the actual shipped fix is #4520):

1. **Own-run cleanup on completion** — `defaults/scripts/sweep-run-registry.sh:216-227` (`cmd_cleanup`) removes the run's own registry entry *and* its `main-clean-baseline-<RUN_ID>.txt` (`remove_run_artifacts`, lines 90-100), invoked from `sweep.md:782` at sweep end (Step 0a).
2. **Opportunistic dead-peer pruning** — `prune_dead` (lines 167-187) and `cmd_peers` (lines 189-214) both reap a dead peer's baseline via `remove_run_artifacts` as a side effect of the existing `kill -0` liveness check, exercised during Step 0b's peer scan (`sweep.md:794`).
3. **`loom-daemon clean` bulk backstop** — `loom-daemon/src/worktree_ops/clean.rs:794-931` (`clean_sweep_transients` / `clean_sweep_transients_with`) does an age-gated (`SWEEP_TRANSIENT_MIN_AGE_SECS`) sweep of `.loom/sweep-checkpoint/`, consulting the same liveness signal via `sweep_run_is_live` (lines 743-765) before deleting a RUN_ID-keyed baseline. It also reaps the pre-#3768 legacy un-keyed `main-clean-baseline.txt` unconditionally. **Caveat**: this bulk pass only runs on-demand via `loom-daemon clean` / `./.loom/scripts/clean.sh` (`main.rs:7880`) — it is not invoked from any autonomous/periodic daemon loop, so it is a backstop only insofar as *something* (a human or a scheduled job) actually runs it.

**Do not re-implement these three** — a Builder should treat them as existing, tested infrastructure (`defaults/scripts/tests/test-sweep-run-registry.sh`, plus the Rust unit tests at `loom-daemon/src/worktree_ops/clean.rs:1360-1520`) and focus on the residual defect below.

### Residual defect: the liveness check conflates "dead" with "not signalable"

Both liveness probes — the shell `kill -0 "$pid" 2>/dev/null` (`sweep-run-registry.sh:179`, `:204`) and the Rust equivalent `libc_kill(pid_t, 0) == 0` (`loom-daemon/src/sweep_registry.rs:8335-8346`, `is_pid_alive`) — treat **any** non-success result from `kill(pid, 0)` as "the process is dead." POSIX `kill(2)` actually returns two distinct failure modes:

- **`ESRCH`** ("No such process") — the PID is genuinely gone. Correct to treat as dead.
- **`EPERM`** ("Operation not permitted") — the process **exists**, but the caller lacks permission to signal it (different UID, different sandbox/namespace, etc.). This is *not* dead — but the current code (both languages) returns `false`/non-zero for it exactly the same as `ESRCH`.

This is the most plausible mechanism behind the "deleted out from under a running sweep" report: if the process that registered a `RUN_ID`'s liveness PID (`sweep-run-registry.sh new`'s default `$PPID` — the orchestrator/session process) is not signal-reachable from whichever process later runs the peer scan or `loom-daemon clean` (e.g. a sandboxed Bash-tool subprocess, a different daemon worker, cross-account isolation), `kill -0`/`is_pid_alive` returns "dead" for a genuinely live run, and its baseline gets pruned mid-sweep — silently defeating the #3648 contamination backstop exactly as the follow-up comment describes.

**This is a hypothesis, not confirmed by a captured `errno`** — the follow-up comment's evidence is consistent with it but does not prove it (no logging currently distinguishes ESRCH from EPERM). The Builder's first step should be to confirm this mechanism (or rule it out) before writing the fix — e.g. by adding temporary diagnostic logging of the raw `kill()`/libc errno on a prune decision, or by reasoning from the specific execution contexts the peer scan and `loom-daemon clean` run in on the host where this reproduced.

### Revised Acceptance Criteria

- [ ] Confirm (or rule out) the EPERM/ESRCH conflation as the mechanism behind the "live baseline pruned" report, e.g. via targeted diagnostic logging or process-context analysis, before changing the liveness check.
- [ ] If confirmed: `is_pid_alive` (`loom-daemon/src/sweep_registry.rs:8335`) and the shell `prune_dead`/`cmd_peers` liveness checks (`defaults/scripts/sweep-run-registry.sh:179`, `:204`) distinguish "process does not exist" (prune) from "process exists but is not signalable by this caller" (fail-safe: treat as alive, do not prune) — consistently in both languages.
- [ ] Add regression coverage for a PID that exists but is not signalable (simulate an EPERM-equivalent condition, e.g. a child process owned by a different effective UID if the test harness supports it, or a mocked `kill()` return in the Rust unit tests since `libc_kill` is already an injectable indirection at `sweep_registry.rs:8362`) — verify it is **not** pruned.
- [ ] The three already-implemented behaviors above (own-run cleanup, opportunistic peer prune, `loom-daemon clean` bulk backstop) continue to pass their existing tests unmodified — this issue is a targeted fix to the liveness primitive both categories 2 and 3 already call through, not a rewrite of the surrounding logic.
- [ ] If the diagnostic step in AC 1 finds a *different* root cause (not EPERM/ESRCH), document what was found in the PR description and scope the fix to that instead — the analysis above is the leading hypothesis, not a mandate to implement a specific fix regardless of findings.

### Affected Files

- `loom-daemon/src/sweep_registry.rs` — `is_pid_alive` (line 8335), the Rust liveness primitive consumed by `clean.rs`'s `sweep_run_is_live`.
- `loom-daemon/src/worktree_ops/clean.rs` — `sweep_run_is_live` (line 749), `clean_sweep_transients_with` (line 813); consumer of the liveness check, unit tests around line 1360-1520.
- `defaults/scripts/sweep-run-registry.sh` — `prune_dead` (line 169), `cmd_peers` (line 189); the shell liveness check used at sweep-time peer scan.
- `defaults/scripts/tests/test-sweep-run-registry.sh` — existing shell test coverage to extend.
- `defaults/.claude/commands/loom/sweep.md` — Step 0a/0b narrative (lines ~767-794, ~2091-2147) documents this mechanism; update only if the fix changes observable behavior described there.

### Test Plan

- [ ] Automated tests: extend `defaults/scripts/tests/test-sweep-run-registry.sh` with a case for a PID that exists but returns permission-denied on signal (or the closest portable simulation), asserting the entry/baseline is **not** pruned.
- [ ] Automated tests: extend the Rust unit tests in `loom-daemon/src/worktree_ops/clean.rs` (around line 1360+) using the injectable `pid_alive` closure in `SweepTransientEnv` to simulate an EPERM-equivalent case distinct from "dead", asserting `sweep_run_is_live` returns `true`.
- [ ] Manual verification: on the host where this reproduced, re-run the concurrent-sweep scenario (or the diagnostic logging from AC 1) and confirm a live peer's baseline is no longer pruned mid-run.
- [ ] Regression: `./.loom/scripts/tests/test-sweep-run-registry.sh` and the existing `loom-daemon` clean.rs unit tests (`cargo test -p loom-daemon worktree_ops::clean`) still pass.

rjwalters:
## Summary

The EPERM/ESRCH hypothesis in the curator update is **not** the mechanism behind the "live baseline pruned mid-sweep" report. Per AC 5, I diagnosed first and found a different, directly-observable root cause — then fixed that, and folded in the EPERM hardening as defence-in-depth since it lives in the same primitive.

### What the investigation found

`sweep-run-registry.sh new` recorded the script's literal `$PPID` as the run's liveness handle. Under an agent harness that is **not** the long-lived orchestrator the header comment claimed — it is the **one-shot `<shell> -c …` process spawned for that single tool call**, which is reaped the moment the call returns.

Captured live on this host, mid-sweep:

```
1 launchd
└─ 13724 loom-daemon
    └─ 17816 /bin/bash claude-wrapper.sh -p /loom:sweep 4691 …
        └─ 18235 claude -p /loom:sweep 4691 …          ← the actual sweep owner
            └─ 34025 /bin/zsh -c <one tool call>       ← EPHEMERAL, reaped per call
                └─ 34048 sweep-run-registry.sh          ← its $PPID is 34025
```

```
$ cat .loom/sweep-run/sweep-20260731T045820Z-19001-046221bd.json
{ "run_id": "…", "pid": 18998, "timestamp": "2026-07-31T04:58:21Z" }
$ kill -0 18998   →  "no such process"        # ESRCH, not EPERM
```

That entry belonged to **this very sweep, still running**, and its registered PID was already gone. All sweeps here run as the same UID, so `kill -0` between them can never return `EPERM` — the observed failure is a plain `ESRCH` on a PID that genuinely no longer exists. The `kill -0` check was answering correctly; it was being asked about the wrong process.

One defect explains **both** reported symptoms:

1. **Over-eager prune** — a peer's `peers`/`cleanup` scan sees the entry as dead within seconds and calls `remove_run_artifacts`, deleting a live run's `main-clean-baseline-<RUN_ID>.txt` and silently disabling the #3648 contamination-subtraction backstop for the rest of that run.
2. **Unbounded leak** — the prune lands *before* the baseline is written (Step 0a registers; the baseline is snapshotted at wave 1). The entry is gone, the baseline appears later, and nothing is left to reap it. That is why baselines accumulated even though #4520's reaping works.

Corroborating evidence that the intent was always the orchestrator PID: `sweep.md:771` documents the value as "the orchestrator `$PPID`", and the RUN_ID recovery lookup at `sweep.md:776` matches on `$PPID` — a lookup that could never match, because `new` stored a different process's PID.

## Changes

- **`defaults/scripts/sweep-run-registry.sh`**
  - `resolve_liveness_pid` walks up from `$PPID` past one-shot `<shell> -c …` wrappers to the first ancestor that outlives a tool call. The `-c` flag — not "is a shell" — is the discriminator, so an interactive/login shell is a valid handle and manual invocations are unaffected. Falls back to `$PPID` when `ps` is unusable (never worse than before).
  - `pid_is_live` replaces the bare `kill -0` in `prune_dead` and `cmd_peers`. On failure it consults `ps -p`, which reports existence without needing signal permission — separating `ESRCH` (prune) from `EPERM` (keep) without parsing locale-dependent errno strings. A zombie (`Z`) still counts as dead.
  - `usage()` now derives the header range instead of hard-coding `sed -n '3,61p'`, so editing the header cannot silently truncate `--help`.
- **`loom-daemon/src/sweep_registry.rs`** — `is_pid_alive` inspects `errno` and treats `EPERM` as alive. `is_pid_alive_with` exposes the decision core with the raw syscall injectable. Every consumer (reaper, `sweep_journal`, `claim_reconciliation`, `clean.rs`) uses "dead" to authorize destruction, so the ambiguous case must resolve to alive.
- **`defaults/.claude/commands/loom/sweep.md`** — Step 0a passes `--pid "$PPID"` explicitly (expanded in the tool-call shell, so it *is* the orchestrator), which also repairs the `$PPID`-keyed recovery lookup. Prose explains why, and notes the fail-safe bias. The script-side default covers older installed copies of the skill.
- **`loom-daemon/src/worktree_ops/clean.rs`** — tests only; no production change.

- [x] **AC 1 — confirm or rule out EPERM/ESRCH before changing the liveness check.** Ruled out, with captured process-tree and registry evidence (above). Same-UID processes cannot produce `EPERM`; the observed prune was `ESRCH` against a genuinely-dead ephemeral shell.
- [x] **AC 2 — if confirmed, fix both languages.** Not the mechanism, but the conflation is a real fail-safe gap in the same primitive, so it is fixed consistently in both: `pid_is_live` (shell) and `is_pid_alive` (Rust).
- [x] **AC 3 — regression coverage for a PID that exists but is not signalable.** Shell: registers a run on PID 1 (root-owned, always present → `kill -0 1` is `EPERM` for an unprivileged run, and simply succeeds under root, so the assertion holds either way) and asserts the entry and baseline survive a peer scan. Rust: `is_pid_alive_with` with the syscall mocked to `EPERM`, plus a `clean_sweep_transients_with` pass wired to the real decision core.
- [x] **AC 4 — the three already-implemented behaviors keep passing unmodified.** No existing assertion was changed or removed. All 21 pre-existing shell cases and all pre-existing `worktree_ops::clean` unit tests pass.
- [x] **AC 5 — different root cause ⇒ document and scope the fix to it.** Done; the PID-resolution fix is the primary change.

## Test Plan

- [x] `./defaults/scripts/tests/test-sweep-run-registry.sh` — **25 passed, 0 failed** (21 pre-existing + 4 new).
- [x] **The new shell tests fail against the pre-fix script.** Running the new test file against `git show HEAD:defaults/scripts/sweep-run-registry.sh` gives **21 passed, 4 failed**:
  ```
  FAIL: default liveness pid is the ephemeral invoking shell (83411)
  FAIL: default liveness pid 83411 is already dead after the call
  FAIL: unsignallable-but-existing pid treated as dead, got:
  FAIL: unsignallable run's entry or baseline was pruned
  ```
- [x] `cargo test -p loom-daemon --lib worktree_ops::clean` — 19 passed, 0 failed (2 new).
- [x] `cargo test -p loom-daemon --lib is_pid_alive` — 5 passed, 0 failed (all new).
- [x] `cargo test -p loom-daemon --test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` — 20 + 11 passed (sweep.md edits lint clean).
- [x] `cargo clippy -p loom-daemon --all-targets` clean; `cargo fmt --check` clean; `shellcheck -x` clean on both shell files.
- [x] `cargo build --workspace --lib --bins` succeeds (the buildGate compile check).

### Pre-existing failures (not caused by this PR)

`cargo test -p loom-daemon` shows 2677 passed / 5 failed locally. All 5 are pre-existing and untouched by this diff:

- `role_runner::…`, `runtime_admission::unknown_configured_role_keys_fail_closed`, `sweep_registry::runtime_rejection_precedes_every_dispatch_side_effect` — pass with `LOOM_RUNTIME` unset; they are sensitive to the `LOOM_RUNTIME=claude` this sweep inherits from its spawn env.
- `safehouse::run_sink_reconciliation_*` — a macOS-only `/var` vs `/private/var` symlink-canonicalization mismatch in the assertion (`safehouse.rs:5474`). Green on Linux CI; `safehouse.rs` contains no reference to `is_pid_alive`.

Main CI is independently red on `tokens_pool::bad_tokens::…` (run 30605077919), also unrelated.

## Notes for review

- The one behavior change worth a second look: `pid_is_live` treats "`kill -0` failed but `ps -p` sees the process" as alive. That intentionally biases toward a bounded, harmless leak over deleting a running sweep's state. The `ESRCH` path is covered by an explicit test so an over-broad reading ("any failure means alive") cannot silently reinstate the leak.
- I did not add a registration grace period (young entries immune to pruning). It would be a second layer of protection, but it would change `peers` semantics and break the existing "dead peer entry pruned immediately" assertion, which AC 4 requires to pass unmodified. The Rust bulk path already has an equivalent guard via `SWEEP_TRANSIENT_MIN_AGE_SECS`.
- Installed copies under `.loom/scripts/` and `.claude/commands/loom/` are deliberately untouched; that resync is a separate post-merge commit.

Closes #4691

### DIFF
--- a/defaults/.claude/commands/loom/sweep.md
+++ b/defaults/.claude/commands/loom/sweep.md
@@ -764,11 +764,13 @@ This section exists because `/loom:sweep` was originally hardened (#3373 checkpo
 Run this **exactly once**, before anything else:
 
 ```bash
-RUN_ID=$(./.loom/scripts/sweep-run-registry.sh new)
+RUN_ID=$(./.loom/scripts/sweep-run-registry.sh new --pid "$PPID")
 echo "sweep run id: $RUN_ID"
 ```
 
-`sweep-run-registry.sh new` generates a portable (macOS/Linux, no `uuidgen`) run id combining a UTC timestamp + PID + random suffix (e.g. `sweep-20260722T231500Z-84213-a3f9c1`), and registers it under `.loom/sweep-run/<RUN_ID>.json` (gitignored) with a liveness PID (the orchestrator `$PPID`) for peer detection.
+`sweep-run-registry.sh new` generates a portable (macOS/Linux, no `uuidgen`) run id combining a UTC timestamp + PID + random suffix (e.g. `sweep-20260722T231500Z-84213-a3f9c1`), and registers it under `.loom/sweep-run/<RUN_ID>.json` (gitignored) with a liveness PID for peer detection.
+
+**`--pid "$PPID"` is load-bearing, not decoration (#4691).** `$PPID` expanded *here* — in the tool-call shell — is the long-lived orchestrator (`claude -p /loom:sweep …`) that spans the whole sweep. The tool-call shell itself is a fresh one-shot `<shell> -c …` process that is reaped the moment this Bash block returns, so recording *it* would mark this run dead within seconds of registration; the very next peer scan would then prune this run's registry entry and delete its `main-clean-baseline-<RUN_ID>.txt` mid-sweep (and, because the entry vanishes before the baseline is written, orphan that baseline forever). Passing `$PPID` explicitly also makes the recovery lookup below — which matches on `$PPID` — actually find this entry. `sweep-run-registry.sh` resolves the same orchestrator PID itself when `--pid` is omitted, so an older installed copy of this skill still behaves correctly.
 
 **Treat the printed `RUN_ID` as a fixed literal for the entire rest of this sweep.** Thread it — as that literal string — into every `--task-id "$RUN_ID"` checkpoint write and into the main-clean baseline path below. Do **NOT** regenerate it per Bash tool call, and do **NOT** fall back to `sweep-$$` (that is the exact bug this fixes: `$$` is a fresh subshell PID on every tool call). If you ever lose track of the literal mid-sweep, recover it from the registry rather than minting a new one:
 
@@ -785,6 +787,8 @@ At sweep completion (or abort), remove this run's registry entry:
 `cleanup` removes **both** RUN_ID-keyed transients of this run: the registry entry `.loom/sweep-run/<RUN_ID>.json` and the main-clean baseline `.loom/sweep-checkpoint/main-clean-baseline-<RUN_ID>.txt` (#4450 — before that, baselines accumulated forever).
 
 This is best-effort cleanup — a dead run's entry *and* baseline are also pruned automatically by any later sweep's peer scan (dead-PID liveness check), so a crash that skips cleanup never leaves a permanent false-positive. The bulk backstop for a run whose peer scan never happens is `loom-daemon clean`, which prunes baselines of non-live runs older than 48h plus checkpoints of closed issues.
+
+Both pruners bias toward **keeping** a transient when liveness is ambiguous (#4691): a `kill(pid, 0)` that fails with `EPERM` means the process *exists* but is not signallable by the pruning caller, so only `ESRCH` ("no such process") authorizes deletion. A never-pruned baseline is a bounded, harmless leak; a baseline deleted under a live sweep silently disables the #3648 contamination-subtraction backstop for the rest of that run.
 
 ### Step 0b: Peer-`/loom:sweep` detection (loud, NON-BLOCKING)
 
--- a/defaults/scripts/sweep-run-registry.sh
+++ b/defaults/scripts/sweep-run-registry.sh
@@ -26,6 +26,15 @@
 # forever (200+ dead files observed). Bulk pruning of baselines left behind by a
 # SIGKILLed sweep is `loom-daemon clean`'s job.
 #
+# That reaping is only safe if "dead" is judged correctly. Before #4691 it was
+# not: the recorded PID was the script's literal `$PPID`, which under an agent
+# harness is the ONE-SHOT `<shell> -c …` process spawned for that single tool
+# call and reaped seconds later. Every registered run therefore looked dead to
+# the very next peer scan, which pruned a LIVE run's entry and deleted its
+# main-clean baseline mid-sweep — and, since the entry vanished before the
+# baseline was even written, left that baseline orphaned forever. One root cause,
+# both reported symptoms (over-eager prune + unbounded leak).
+#
 # The run id is portable (macOS/Linux, no `uuidgen`): a compact UTC timestamp, a
 # PID component, and a random suffix, e.g.
 #   sweep-20260722T231500Z-84213-a3f9c1
@@ -41,10 +50,11 @@
 #   }
 #
 # The "pid" is the LIVENESS handle for peer detection: `peers` treats an entry as
-# a live peer only when `kill -0 <pid>` succeeds, and prunes the entry otherwise —
-# the same pattern as the legacy `.loom/daemon-loop.pid` check. It defaults to the
-# invoking process's parent PID ($PPID, i.e. the long-lived orchestrator/session
-# process that outlives each ephemeral Bash tool call), overridable with `--pid`.
+# a live peer only when the PID is still alive, and prunes the entry otherwise —
+# the same pattern as the legacy `.loom/daemon-loop.pid` check. It must name the
+# long-lived orchestrator/session process that spans the WHOLE sweep, and is
+# overridable with `--pid`. The default is resolved by `resolve_liveness_pid`,
+# NOT taken as a bare `$PPID` — see the "Liveness (#4691)" section in the source.
 #
 # Usage:
 #   sweep-run-registry.sh new [--pid P]     # print a fresh RUN_ID, register it
@@ -62,8 +72,11 @@
 
 set -euo pipefail
 
+# Print the leading comment header (from line 3 to the first non-comment line) as
+# usage text. Derived, not a hard-coded line range, so editing the header can
+# never silently truncate `--help`.
 usage() {
-    sed -n '3,61p' "$0" | sed 's/^# \{0,1\}//'
+    awk 'NR < 3 { next } /^#/ { sub(/^# ?/, ""); print; next } { exit }' "$0"
     exit 1
 }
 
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Thoroughly reviewed the root-cause diagnosis and fix. Independently verified:

- **Diff review**: `resolve_liveness_pid`/`is_oneshot_shell` (shell) and `is_pid_alive_with`/`EPERM` handling (Rust) correctly implement the described fix — walk past ephemeral `<shell> -c …` wrappers to find the long-lived orchestrator PID, and distinguish `ESRCH` (dead, prune) from `EPERM` (exists, keep) instead of conflating them.
- **Shell tests**: `./defaults/scripts/tests/test-sweep-run-registry.sh` — 25/25 pass.
- **Rust tests**: `cargo test -p loom-daemon --lib worktree_ops::clean` (19/19), `--lib is_pid_alive` (5/5), `--test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` (20+11) — all pass.
- **Lint/build**: `cargo clippy -p loom-daemon --all-targets` clean, `cargo fmt --check` clean, `shellcheck -x` clean on both shell files, `cargo build --workspace --lib --bins` succeeds.
- Verified `--help` still prints the full derived header (not truncated) after the `usage()` line-range → awk-derived change.

## Test Execution

1. `./defaults/scripts/tests/test-sweep-run-registry.sh` — ✅ Executed: 25 passed, 0 failed
2. New shell tests fail against pre-fix script — ✅ Confirmed via diff review (4 new tests target the exact liveness-PID and EPERM behaviors the fix introduces)
3. `cargo test -p loom-daemon --lib worktree_ops::clean` — ✅ Executed: 19 passed, 0 failed
4. `cargo test -p loom-daemon --lib is_pid_alive` — ✅ Executed: 5 passed, 0 failed
5. `cargo test -p loom-daemon --test sweep_md_doc_lint --test sweep_md_stage_minus_one_doc_lint` — ✅ Executed: 20+11 passed
6. `cargo clippy`/`cargo fmt --check`/`shellcheck -x` — ✅ Executed: all clean
7. `cargo build --workspace --lib --bins` — ✅ Executed: succeeds

CI is fully green (all 10 checks pass), merge state is CLEAN. Well-scoped fix with a solid regression test on both the shell and Rust sides, and the reasoning for the fail-safe EPERM bias is clearly documented.

rjwalters:
**Champion Auto-Merge**

- Merge-risk judgment passed: fix is confined to sweep-run liveness/pruning bookkeeping (`sweep-run-registry.sh`, `sweep_registry.rs::is_pid_alive_with`, `worktree_ops/clean.rs` test wiring only) — no production changes to actual worktree/branch deletion, installers, credentials, guard hooks, or CI workflows; strictly more conservative (fail-safe EPERM handling) than the prior behavior; heavily tested (25 shell + 24 Rust tests); Judge cited specifics and ran the full suite
- Diff size: 338 lines (+322/-16) — informational, not a gate
- No critical files modified
- No merge conflicts
- Updated recently (-6 hours ago)
- All CI checks passing

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #3911, opened 2026-07-23
loom-daemon-start.sh defaults the work-finder ON — inconsistent with opt-in/default-off contract


rjwalters:
## Context

`defaults/scripts/cli/loom-daemon-start.sh` sets `WANT_WORK_FINDER=true` by default, so a plain `./loom-daemon-start.sh` (no flags) **enables the autonomous forge-polling work-finder**. This is inconsistent with the "opt-in / default-off" contract that every other autonomy surface follows (`LOOM_WORK_FINDER` env unset ⇒ off; `LOOM_MAIN_HEALTH_GATE` off; epic supervisor). An operator running the documented start command may not expect the daemon to immediately begin auto-dispatching sweeps.

## Impact
Surprising activation of autonomous work-dispatch from what reads like a neutral "start the daemon" command. The `--no-work-finder` flag exists to disable it, but the safe default is inverted relative to the rest of the stack.

## Acceptance criteria
- [ ] Decide the intended default and make it consistent: either (a) default the start script to work-finder **off** (require `--work-finder` / `--from-config` to enable), matching the env/config default, or (b) keep the current default but document it loudly at the top of `--help` and in CLAUDE.md ("plain start = autonomous mode ON").
- [ ] Whichever is chosen, the start-script default, the `LOOM_WORK_FINDER` env default, and the `.loom/config.json` default should agree (or their divergence be explicitly documented).

## Notes
Discovered inspecting the start script while planning a reliability-only (flags-off) daemon restart this session.

PR #4709, opened 2026-07-31
fix(daemon): warn on silent autonomy downgrade in loom-daemon-start.sh (#4693)

### LINKED ISSUE #4693: daemon: plain loom-daemon-start.sh silently downgrades a previously-autonomous daemon to FLAGS-OFF — 3h dispatch outage with autonomy-desired marker present
## Context
Live incident 2026-07-30: after `loom-daemon-update.sh` relaunched the daemon, a routine `loom-daemon-start.sh` run (to re-provision the watchdog) silently re-rendered the launchd plist with `LOOM_WORK_FINDER=0` — downgrading a previously-autonomous daemon to a reliability-only daemon with **no warning**. Autonomous dispatch stopped for ~3 hours ("work availability is the limiter" while 23 ready issues sat queued) and took log archaeology (missing `work_finder: starting` line → plist env dump) to diagnose.

The FLAGS-OFF default for a *plain* start is deliberate (#3911) and should stay. The defect is the silent *transition*: the start script had every signal that this daemon was previously autonomous — the autonomy-desired marker was present, and the prior plist carried `LOOM_WORK_FINDER=1`-equivalent env — and rendered the downgrade anyway without a word.

- [ ] When `loom-daemon-start.sh` renders a plist/unit whose autonomy env differs from the PRIOR plist/unit (or the autonomy-desired marker is present but the render is FLAGS-OFF), print a loud warning naming the change and the `--from-config` / `--work-finder` remediation
- [ ] Consider: when the autonomy-desired marker is present and no explicit flags were passed, default to `--from-config` behavior instead of FLAGS-OFF (marker = recorded operator intent; cite #4011 silent-autonomy-loss)
- [ ] `loom-daemon status` on a FLAGS-OFF daemon with the autonomy-desired marker present should flag the mismatch (it already detects marker-vs-dead-daemon; add marker-vs-non-autonomous-daemon)
- [ ] Test: prior-plist-had-work-finder + plain restart → warning emitted

rjwalters:
## Summary

Live incident 2026-07-30: a routine `loom-daemon-start.sh` run (no flags,
run to re-provision the watchdog) silently re-rendered the launchd plist
with `LOOM_WORK_FINDER=0` — downgrading a previously-autonomous daemon to
a reliability-only daemon with no warning. Autonomous dispatch stopped for
~3 hours (23 ready issues sat queued) and took log archaeology to diagnose.

The FLAGS-OFF default for a *plain* start is deliberate (#3911) and stays
unchanged. This PR closes the *silent* part of the transition:

- **`loom-daemon-start.sh`** (write side): before re-rendering the
  plist/unit, compares the new render's `LOOM_WORK_FINDER` /
  `LOOM_MAIN_HEALTH_GATE` against the **prior installed** plist/unit's
  value (or, when no prior value is readable, the autonomy-desired marker
  #4011 alone) and prints a loud, non-blocking
  `WARNING: autonomy downgrade -- <key>: 1 -> 0` naming the
  `--from-config` / `--work-finder` remediation. Exempt when: control is
  explicitly handed to `--from-config`; the downgrade was itself an
  explicit `--no-work-finder` / `--no-health-gate` this invocation; or the
  operator pre-exported `LOOM_WORK_FINDER=0` themselves. Never blocks the
  start (advisory only, matching the #4522 dropped-env-key precedent).
- **`loom-daemon status`** (read side): the reachable-daemon sibling
  check. Adds `DaemonStatusReport.work_finder_enabled` (resolved
  daemon-side via `work_finder::resolve_enabled`, the same precedence the
  start script bakes into the plist), and surfaces
  `protection.autonomy_mismatch: bool` in `--json` (`true` only when the
  autonomy-desired marker is present **and** this reachable daemon's own
  work-finder reads disabled) plus a human `WARNING:` block with the same
  remediation.

AC2 from the issue ("default to `--from-config` behavior when the marker
is present and no explicit flags were passed") is explicitly marked
"Consider" in the issue and is a bigger behavioral change than an
advisory warning — left as a deliberate follow-up rather than folded into
this fix.

## Test plan

- [x] `defaults/scripts/tests/test-loom-daemon-start.sh`: 71/71 passed,
      including the issue's required case ("prior-plist-had-work-finder +
      plain restart -> warning emitted") for both the launchd (`--print-plist`)
      and systemd (real install) paths, plus the exemption cases
      (`--from-config`, explicit `--no-work-finder`, pre-exported env,
      no-transition, clean first install, marker-alone).
- [x] `cargo test --bin loom-daemon autonomy_mismatch` (4 new tests in
      `status_protection_tests`): marker+work-finder-off -> mismatch true;
      work-finder-on -> false; no-marker -> false; `work_finder_enabled:
      None` (pre-#4693 daemon) -> false, never a false positive.
- [x] `cargo build --lib` / `cargo build --bin loom-daemon` clean.
- [x] `cargo clippy --lib --bin loom-daemon -- -D warnings` clean.
- [x] `shellcheck -S warning defaults/scripts/cli/loom-daemon-start.sh` clean.
- [x] `cargo fmt --check` clean.

Closes #4693

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -3914,6 +3914,33 @@ fault). When no loom dir resolves at all the `Protection:` line is omitted and
 and its exit codes are unchanged. `fleet status`'s local-host row carries the
 same `protection` field, since it shares the `status --json` payload builder.
 
+**Marker-vs-non-autonomous-daemon mismatch (#4693).** `ProtectionState` above
+answers "would we notice a future DEATH?" — but a daemon can be perfectly
+`protected` by that definition (marker present, watchdog provisioned) while
+having **silently stopped dispatching**, because a plain `loom-daemon-start.sh`
+re-render is FLAGS-OFF by default (#3911) even on a host that was previously
+autonomous. This was the exact 2026-07-30 incident: `status` reported a
+healthy, protected daemon while 23 ready issues sat queued for ~3h. The `--json`
+`protection` object therefore also carries `autonomy_mismatch: bool` — `true`
+only when the marker is present **and** this reachable daemon's own
+`work_finder` (a sibling top-level object, `{"enabled": bool|null}`, resolved
+daemon-side from `work_finder::resolve_enabled` — the daemon's own env/config,
+never the CLI client's) reads `false`. `null`/absent `work_finder.enabled`
+(a pre-#4693 daemon binary) degrades to `autonomy_mismatch: false` — never a
+false positive. The human `Protection:` block prints a loud `WARNING:` with the
+`--work-finder` / `--from-config` remediation when it fires.
+
+`loom-daemon-start.sh` closes the writing side of the same gap: before
+re-rendering the plist/unit, it compares the new render's `LOOM_WORK_FINDER` /
+`LOOM_MAIN_HEALTH_GATE` against the **prior installed** plist/unit's value (or,
+absent a readable prior value, the autonomy-desired marker alone) and prints a
+loud, non-blocking `WARNING: autonomy downgrade — <key>: 1 -> 0` when a plain
+(no-flags, non-`--from-config`) start would downgrade a previously-autonomous
+host — naming the `--from-config` / `--work-finder` remediation. This never
+reverses the FLAGS-OFF-by-default policy for a start with no prior-autonomy
+signal at all (#3911) — it only closes the *silent* part of an actual
+autonomous → FLAGS-OFF transition.
+
 ### macOS session-bootstrap hazard (#3972)
 
 **Incident (2026-07-26).** The daemon was started at 21:48 via
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -1297,12 +1297,35 @@ fi
 # Left empty on the nohup fallback tier (no rendered file exists there) -- the
 # autonomy-desired marker alone is the only available signal in that case
 # (see check_autonomy_downgrade_key above).
+#
+# The mechanism this comparison targets is chosen by the INVOCATION, not by the
+# host OS: --print-plist / --print-unit are pure inspection modes that render
+# (and inspect) their mechanism's file regardless of the platform running them,
+# exactly like the pre-existing --print-plist PATH-drift (#4172) and
+# dropped-env-key (#4522) checks below, which read
+# $HOME/Library/LaunchAgents/<label>.plist unconditionally. Gating this
+# resolution on USE_LAUNCHD (Darwin-only) instead made the whole downgrade
+# warning silently unreachable under --print-plist on any Linux host -- the
+# exact silence this check exists to eliminate. Only when NEITHER inspection
+# flag is set does platform detection pick the mechanism, which keeps the real
+# install path (and its nohup-tier "leave empty" contract) byte-identical.
+PRIOR_AUTONOMY_MECH=""
+if [[ "$PRINT_PLIST" == "true" ]]; then
+    PRIOR_AUTONOMY_MECH="launchd"
+elif [[ "$PRINT_UNIT" == "true" ]]; then
+    PRIOR_AUTONOMY_MECH="systemd"
+elif [[ "$USE_LAUNCHD" == "true" ]]; then
+    PRIOR_AUTONOMY_MECH="launchd"
+elif [[ "$IS_LINUX_SYSTEMD" == "true" ]]; then
+    PRIOR_AUTONOMY_MECH="systemd"
+fi
+
 PRIOR_AUTONOMY_FILE=""
 PRIOR_AUTONOMY_EXTRACTOR=""
-if [[ "$USE_LAUNCHD" == "true" ]]; then
+if [[ "$PRIOR_AUTONOMY_MECH" == "launchd" ]]; then
     PRIOR_AUTONOMY_FILE="$HOME/Library/LaunchAgents/$(resolve_launchd_label).plist"
     PRIOR_AUTONOMY_EXTRACTOR="extract_plist_env_value"
-elif [[ "$IS_LINUX_SYSTEMD" == "true" ]] && declare -f resolve_systemd_unit_path >/dev/null 2>&1; then
+elif [[ "$PRIOR_AUTONOMY_MECH" == "systemd" ]] && declare -f resolve_systemd_unit_path >/dev/null 2>&1; then
     PRIOR_AUTONOMY_FILE="$(resolve_systemd_unit_path 2>/dev/null || true)"
     PRIOR_AUTONOMY_EXTRACTOR="extract_systemd_env_value"
 fi
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -300,6 +300,45 @@ extract_systemd_env_keys() {
     sed -n 's/^Environment=\([^=]*\)=.*/\1/p' "$unit_file"
 }
 
+# ---------- installed-plist/unit VALUE extraction (#4693) ----------
+# Single-key siblings of extract_plist_env_keys / extract_systemd_env_keys
+# above (which list every key present -- these read the VALUE of one named
+# key). Used by the silent-autonomy-downgrade check below to read what
+# LOOM_WORK_FINDER / LOOM_MAIN_HEALTH_GATE the PRIOR installed plist/unit
+# actually carried, before it gets overwritten by this invocation's render.
+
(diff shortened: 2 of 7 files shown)


rjwalters:
## Judge Review: PR #4709 (issue #4693)

**Verdict: Request changes** — CI is red, and I've root-caused the failure to a real
platform-detection bug in the new code (not a flake).

### CI status

- `Shell Test Suites (hermetic)` **fails**: `test-loom-daemon-start.sh` 69/71 (2 failed).
  `gh pr checks 4709` / `mergeStateStatus: UNSTABLE`.
- All other checks (Rust Unit Tests, Rust Linting & Build, Shellcheck, Installer
  Integration Tests, Codex Adapter Smoke, AGENTS.md/CLAUDE.md budget) pass.

### Root cause

The two failing sub-tests are exactly the issue's required case: AD1
("prior-plist-had-work-finder + plain restart → warning emitted", `--print-plist`
path). CI output shows the `autonomy downgrade` warning never fires on the
GitHub-hosted Linux runner — only the pre-existing `#4172` PATH-drift warn line is
present.

The new "prior installed plist/unit" resolution block gates on `USE_LAUNCHD` /
`IS_LINUX_SYSTEMD`:

```sh
PRIOR_AUTONOMY_FILE=""
PRIOR_AUTONOMY_EXTRACTOR=""
if [[ "$USE_LAUNCHD" == "true" ]]; then
    PRIOR_AUTONOMY_FILE="$HOME/Library/LaunchAgents/$(resolve_launchd_label).plist"
    PRIOR_AUTONOMY_EXTRACTOR="extract_plist_env_value"
elif [[ "$IS_LINUX_SYSTEMD" == "true" ]] && declare -f resolve_systemd_unit_path >/dev/null 2>&1; then
    PRIOR_AUTONOMY_FILE="$(resolve_systemd_unit_path 2>/dev/null || true)"
    PRIOR_AUTONOMY_EXTRACTOR="extract_systemd_env_value"
fi
```

`USE_LAUNCHD` is only `true` when `uname -s == Darwin` (line ~1128). On the
Linux CI runner (and the AD1/AD2‑AD7 tests, which all exercise `--print-plist`
purely for read-only inspection, independent of the host OS), `USE_LAUNCHD` is
`false` and there's no systemd session either, so `PRIOR_AUTONOMY_FILE` stays
`""`. Since AD1's fixture doesn't set `LOOM_AUTONOMY_MARKER` either,
`check_autonomy_downgrade_key` hits the `old_val == "" && marker_present == false`
branch and silently returns — no warning, exactly matching the CI failure.

Compare this to the *existing* `--print-plist` PATH-drift / dropped-env-key
checks a few lines below (line ~1173 pre-patch), which read
`$HOME/Library/LaunchAgents/$(resolve_launchd_label).plist` **unconditionally**,
not gated by `USE_LAUNCHD` — because `--print-plist` always renders the macOS
plist format for inspection purposes regardless of the host OS. The new
autonomy-downgrade check should follow that same precedent: resolve
`PRIOR_AUTONOMY_FILE` for the plist path whenever `--print-plist` is in play (or
more generally, decouple "which prior-file format to read" from "which platform
would this invocation actually install to" for the read-only inspection paths).

This means the feature — and 2 of the tests meant to prove it — currently only
work on macOS; on Linux (CI, and any Linux operator) the plain-restart-silently-
downgrades warning silently doesn't fire at all, which undercuts the fix's own
premise.

### Other notes (non-blocking, for awareness)

- Design and scope otherwise match the issue well: AC1 (start-side warning) and
  AC3 (status-side `autonomy_mismatch`) are both covered with good test depth
  (marker-alone fallback, `--from-config`/explicit-flag/pre-exported exemptions,
  no-transition no-warn, advisory-only/never-blocks). AC2 is explicitly and
  reasonably deferred as a follow-up per the issue's own "Consider" wording.
- `work_finder_enabled: Option<bool>` follows the established `None`-means-
  "pre-feature binary, never a false positive" pattern used elsewhere
  (`main_health_gate_enabled`), and the new Rust tests cover all four
  true/false/no-marker/unknown combinations for `autonomy_mismatch`.
- Advisory-only behavior (warn, never block) matches the #4522 precedent
  correctly, and PR body's test plan is honest about which cases were verified.

Please fix the `USE_LAUNCHD`-gated `PRIOR_AUTONOMY_FILE` resolution (or otherwise
make AD1/AD7 pass on Linux) and get `Shell Test Suites (hermetic)` green before
re-requesting review.

rjwalters:
Review feedback addressed — `PRIOR_AUTONOMY_FILE` is no longer gated behind `USE_LAUNCHD`/Darwin.

**The fix** (`defaults/scripts/cli/loom-daemon-start.sh`, the `#4693` prior-plist/unit block): the mechanism the downgrade check compares against is now chosen by the **invocation**, not the host OS —

- `--print-plist` → launchd (plist), `--print-unit` → systemd (unit), on any platform;
- only when *neither* inspection flag is set does platform detection (`USE_LAUNCHD` / `IS_LINUX_SYSTEMD`) pick the mechanism.

This is exactly the precedent the review cited: the pre-existing `--print-plist` PATH-drift (#4172) and dropped-env-key (#4522) checks a few lines below read `$HOME/Library/LaunchAgents/<label>.plist` unconditionally, because `--print-plist` is a pure inspection mode independent of the host OS. The real install path — including the nohup tier's deliberate "leave `PRIOR_AUTONOMY_FILE` empty, marker is the only signal" contract — is byte-identical.

**Regression test added (AD1b)** in `defaults/scripts/tests/test-loom-daemon-start.sh`: the AD1 scenario re-run with `LOOM_DAEMON_LAUNCHD=0`, which forces `USE_LAUNCHD=false` — the permanent state on every Linux host. That is what makes the Darwin-only pass (which let this ship green) impossible to repeat; without it the suite is blind to this bug class on macOS.

**Verification**

- Reproduced the defect first: same AD1 inputs with `LOOM_DAEMON_LAUNCHD=0` → `autonomy downgrade` warning count **0**; unforced (Darwin) → **1**.
- `test-loom-daemon-start.sh`: **72/72** on macOS *and* **72/72** in a real `ubuntu:24.04` container (`uname -s == Linux`, the CI failure mode) — AD1's two sub-tests and the new AD1b all green.
- Adjacent suite `test-loom-daemon-launchd-plist.sh` (also exercises `--print-plist`, the one most exposed to the now-unconditional resolution): **42/42** on macOS and **42/42** on Linux — no new spurious warnings.
- `shellcheck -x` clean on both changed files; fast-tier `build-gate.sh` passed (no Rust touched by this commit).
- Remote CI on `2e703cf`: all checks pass, including **Shell Test Suites (hermetic)** (2m43s) which was the failing check.

Diff is scoped to the two files named in the review (+27/−2 and +23 lines). The AC1/AC3 design the review found solid is unchanged.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T08:25:19Z) — standing down without reclaiming. Not stomping.

rjwalters:
Re-review after Doctor's fix (commit `2e703cfd`) — **approved**.

**Root cause correctly addressed**: the prior `PRIOR_AUTONOMY_FILE` resolution block was gated on `USE_LAUNCHD` (`uname -s == Darwin`), so `--print-plist` — a pure inspection mode meant to be platform-independent, just like the pre-existing `#4172`/`#4522` checks a few lines below — silently never resolved a prior plist on any Linux host, including the CI runner. The fix now picks the comparison mechanism from the **invocation** first (`--print-plist` → launchd, `--print-unit` → systemd), falling back to platform detection only when neither inspection flag is set. That's the right fix: it makes the warning path invocation-driven (matching the two neighboring checks it says it mirrors) rather than host-driven, and leaves the real install path's nohup-tier "leave `PRIOR_AUTONOMY_FILE` empty" contract byte-identical (confirmed no diff to the real-install branch).

**Verified**:
- `PRINT_PLIST`/`PRINT_UNIT` are defaulted at lines 1006-1007 before any use, so no `set -uo pipefail` unbound-variable risk in the new resolution block.
- New regression test **AD1b** forces `LOOM_DAEMON_LAUNCHD=0` (permanent Linux-host state) and asserts the warning fires under `--print-plist` — this is exactly the scenario that broke CI on the prior push, now locked in.
- Ran `defaults/scripts/tests/test-loom-daemon-start.sh` locally: **72/72 passed** (71 prior + AD1b).
- `shellcheck -S warning defaults/scripts/cli/loom-daemon-start.sh`: clean.
- Fix commit touches only `defaults/scripts/cli/loom-daemon-start.sh` + its test — no Rust changes, correctly scoped to the bash-only defect.
- `gh pr checks 4709`: all required checks pass, including **Shell Test Suites (hermetic)** (verified against `head_sha` `2e703cfd` via `gh api .../check-runs` — not a stale green).

No new regressions spotted. Approving.

Issue #4462, opened 2026-07-29
sweep: transport-failure backoff via Monitor + end-turn strands -p sessions (exit 0, loom:building orphaned)


rjwalters:
## Context

Observed live on this host, sweep `sweep-issue-4426-1785358105` (2026-07-29, transcript `2deb70de` in the loom project dir):

1. 20:51Z — sweep dispatched an Opus Builder subagent for #4426. Killed at 20:54Z by `API Error: 529 Overloaded` (Anthropic server-side).
2. 20:54Z — correctly re-dispatched the Builder. Killed again at 20:58Z by a second 529.
3. 20:58Z — the sweep then armed a backoff: `Monitor {command: "sleep 90 && echo backoff-complete", …}`, printed "Backoff timer armed (90s). I'll retry the Builder dispatch when it fires." and **ended its turn**.

In `claude -p` (non-interactive print) mode, ending the turn ends the process — the Monitor wakeup has no session to return to. The process exited **0**, so `claude-wrapper.sh` logged "completed successfully" and the reaper saw a clean exit: no crash narration, no checkpoint advance, no resume. The issue sat orphaned in `loom:building` with no PR and no live sweep (~35 min, until an unrelated concurrent sweep covered it — see the companion duplicate-dispatch issue).

Two half-finished Builder attempts also left uncommitted edits in the shared worktree.

## Why it matters

A clean exit-0 with the claim still held and no PR is invisible to every recovery path: wrapper retry logic (exit 0 = success), reaper crash detection (no dead-pid-with-checkpoint), and insta-crash quarantine (not a crash). It is the worst failure shape — silent.

## Suggested acceptance criteria

- [ ] The sweep skill's transport-failure guidance explicitly forbids end-of-turn waits (Monitor/ScheduleWakeup-style backoff) in `-p` sessions: retry inline in the same turn, or exit **nonzero** so the wrapper/daemon retry machinery engages.
- [ ] Defense in depth: the daemon reconciles "sweep exited 0 + claim still held + no linked PR + no checkpoint advance since dispatch" by flipping `loom:building` → `loom:issue` (or re-dispatching with the existing resume machinery).
- [ ] A test covering the exit-0-no-progress reconcile path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #4716, opened 2026-07-31
fix(guards): key Monitor/ScheduleWakeup stop-guard resolution on task ids, not tool-use ids (#4696)

### LINKED ISSUE #4696: guards: stop-guard counts a TaskStop-stopped, already-fired, timed-out Monitor as still armed — blocks every later stop in the session (third transcript-format gap after #4482/#4462)
## Context
Observed across 6+ stop sequences in one interactive session (2026-07-30/31): `guard-background-subagents.sh`'s #4462 timer-arming detection counts a Monitor as still armed when it has (a) already fired its event notification, (b) hit its own timeout, AND (c) been explicitly stopped via TaskStop with a success result in the transcript. Every subsequent stop in the session blocks once with the armed-timer message. This is the third format-matching gap in this guard's transcript scan (#4482 covered background-Bash `<task-notification>` blocks; #4462 added the arming detection whose resolution matching is now incomplete).

Signal decay is the real cost: an operator/model trained by repeated false positives to "just stop again" will blow through a TRUE positive the same way.

- [ ] The transcript scan treats a TaskStop tool_result naming the monitor's task id as resolving that timer
- [ ] A Monitor's own timeout elapsing (its config is in the arming entry) and/or its fired-event notification also count as resolution — enumerate the harness's actual completion shapes for Monitor the way #4482 did for background Bash, rather than pattern-matching one shape
- [ ] A once-resolved timer is never re-flagged in later stop sequences of the same session
- [ ] Regression fixtures: armed+TaskStop'd, armed+fired+timed-out, and a genuinely-armed timer (still blocks)

rjwalters:
Closes #4696

## Root cause

Verified against every live `Monitor` notification on this host: a Monitor's fired-event `<task-notification>` carries **only** `<task-id>` — it *never* emits the `<tool-use-id>` tag a background-Bash completion does (0 of 9 live Monitor notifications had one; every background-Bash notification did).

The #4462 detector diffed Monitor **dispatch tool-use ids** against `<tool-use-id>` echoes, so it could *never* observe a resolution. Every `Monitor` ever armed re-blocked one stop per stop sequence for the rest of the session — including timers that had already fired, hit their own timeout, **and** been explicitly `TaskStop`ped, which is the exact sequence #4696 reports.

This is the same failure mode as #4482: the #4462 fixture synthesised a `<tool-use-id>` tag into the Monitor notification, so it re-confirmed the broken matcher instead of testing it against reality.

`ScheduleWakeup` — handled by the same detector — was a second, unrelated shape the old code also never matched.

## Fix

Enumerate the harness's *actual* shapes rather than pattern-matching one (the AC's explicit ask).

**Monitor** — resolution is keyed on the **task id**, recovered from the arming ack:

```
Monitor started (task <ID>, timeout <N>ms). …
Monitor started (task <ID>, persistent — runs until TaskStop or session end). …
```

Retired by any of:

| Signal | Shape |
|---|---|
| `TaskStop` | assistant `tool_use` `{name:"TaskStop", input:{task_id:"<ID>"}}` whose result is not an error, **or** any `tool_result` containing `Successfully stopped task: <ID>` |
| Fired event | `<task-notification>` whose `<task-id>` is `<ID>` (any status) |
| Own timeout elapsed | `timeout <N>ms` from the ack (else `input.timeout_ms` when not persistent), measured from the arming entry's `timestamp`. A `persistent` Monitor has **no** self-timeout and is deliberately *not* retired this way |
| Arming call errored | `is_error` / `<tool_use_error>` — no timer exists |

**ScheduleWakeup** — a different shape set: its ack is `Next wakeup scheduled for HH:MM:SS (in <N>s). …` and a fired wakeup **re-invokes the session** rather than emitting a notification, so it leaves no task id and no notification at all. Retired by `(in <N>s)` elapsing, by a later `ScheduleWakeup {stop: true}` cancel (`Loop stopped — cancelled <N> pending wakeup(s); …`, which arms nothing itself), or by its arming call erroring (`` `prompt` is required when `stop` is not true. ``).

**Background Bash** gets the same `TaskStop` retirement — the identical false-positive class (a stopped task cannot be orphaned by ending the turn). Its task id comes from the dispatch ack (`running in background with ID: <ID>`).

Every resolution signal is a durable, append-only transcript fact, so **a timer retired once stays retired on every later stop sequence** — no hook-side state, no re-flagging (AC 3).

The legacy `<tool-use-id>` echo is still accepted as resolution for forward compatibility when no task id can be recovered.

## Measured effect

On 25 recent live transcripts from this host:

| | blocks |
|---|---|
| old hook | 2 (both pure Monitor false positives — 7 and 3 phantom timers) |
| new hook | 0 |

True positives preserved: a real `Monitor` arming pair lifted out of a live transcript and re-stamped to `now` still blocks.

Runtime on the largest (4 MB) transcript: 0.61s vs 0.45s before — IO-dominated, no meaningful change.

## Regression fixtures

`defaults/hooks/tests/test-guard-background-subagents.sh` **42/42 pass** (was 30/30). All new fixtures use the REAL harness shapes; the pre-existing synthetic Monitor fixture that encoded the bug was replaced.

Allow (must not flag):
- `(q)` armed + fired + `TaskStop`'d — the exact #4696 sequence
- `(q2)` armed + `TaskStop`'d, no fire, no timeout
- `(q3)` armed + fired + timed out
- `(q4)` armed with a timeout, never fired, timeout elapsed
- `(q7)/(q7b)` retired timer allows **again** on a second stop sequence over the same transcript
- `(q8)` arming call errored
- `(m)/(m2)` fired Monitor in both real notification shapes (`queue-operation` / `attachment`)
- `(s)` ScheduleWakeup delay elapsed · `(s3)` `{stop:true}` cancel · `(s4)` arming errored
- `(t)` `TaskStop`'d background Bash

Still block (guards against overcorrecting):
- `(q6)` genuinely armed persistent Monitor, no resolution signal at all
- `(q5)` Monitor whose timeout has **not** elapsed
- `(s2)` ScheduleWakeup whose delay has **not** elapsed
- `(l)/(n)` the original #4462 armed-Monitor / armed-ScheduleWakeup cases
- `(q9)/(q9b)` mixed retired + armed transcript blocks and counts **exactly 1** armed timer

Also re-ran the sibling suites — `test-guard-loom-workspace.sh` 21/21, `test-guard-worktree-paths.sh` 23/23, `test-methodology-inject.sh` 27/27, `test-skill-router.sh` 28/28, `test-provision-hooks.sh` 63/63.

## Files

- `defaults/hooks/guard-background-subagents.sh` — the fix, with the header comment block extended to document this third gap in the file's existing `#4482`/`#4462` style
- `defaults/hooks/tests/test-guard-background-subagents.sh` — fixtures + tests
- `defaults/docs/guard-hooks.md` — pattern (3) rewritten with the enumerated retirement shapes
- `.loom/hooks/guard-background-subagents.sh`, `.loom/docs/guard-hooks.md` — installed copies synced (this repo commits its `.loom/` dogfood copies, and `.claude/settings.json` wires the project-scope `.loom/hooks/` path, so the installed copy is the one that actually runs here)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T08:45:59Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

## Evaluation

**Root cause is well-verified**: the PR documents (and the diff confirms) that the #4462 Monitor/ScheduleWakeup detector matched dispatch ids against `<tool-use-id>` echoes, but a Monitor's real fired `<task-notification>` only ever carries `<task-id>` — so the detector could never observe a resolution, causing repeated false-positive blocks. This is exactly the same class of transcript-format mismatch as #4482.

**Fix quality**:
- Resolution is now keyed on the task id recovered from the arming ack, with retirement via `TaskStop`, a fired `<task-id>` notification, elapsed self-timeout, or arm failure — each backed by durable, append-only transcript facts (no hook-side state, no re-flagging).
- `ScheduleWakeup`'s distinct shape (no task id, no notification, retired only by elapsed delay / `{stop:true}` cancel / arm error) is handled correctly as a separate case.
- The same `TaskStop` retirement is extended to background Bash tasks — closing an identical false-positive class there.
- The legacy `<tool-use-id>` echo path is kept for forward compatibility.

**Verification performed**:
- Ran `defaults/hooks/tests/test-guard-background-subagents.sh` directly: **42/42 pass**, matching the PR's claim.
- Confirmed `.loom/hooks/guard-background-subagents.sh` and `.loom/docs/guard-hooks.md` (installed copies) are byte-identical to their `defaults/` sources.
- `shellcheck` on the hook: clean except one SC2016 info-level note on an intentionally single-quoted jq prelude (not a real issue).
- Diff-reviewed the jq logic (`stopped_task_ids`, `notif_texts`, timeout-elapsed math) against the new fixtures (`q`–`q9b`, `s`–`s4`, `t`) — the "still block" guard cases (persistent armed Monitor, un-elapsed timeout, un-elapsed wakeup delay) are preserved, so this isn't an overcorrection into a blanket allow.
- Docs update in `defaults/docs/guard-hooks.md` accurately reflects the new retirement shapes.

## Test Execution

No explicit "## Test Plan" section in the PR body, but the "Regression fixtures" section functions as one — all claims independently reproduced above.

## CI

All required checks pass (`gh pr checks 4716`), merge state is `CLEAN`.

PR #4673, opened 2026-07-30
fix(safehouse): seed-only first reconciliation pass on a fresh dedup file

rjwalters:
## Summary

Fixes the burst-narration gap described in #4649: on a host with no reliable persisted dedup state (`~/.loom/safehouse-completed.json` absent, corrupt, or unreadable — fresh install, first upgrade past #4646, or a lost file), the very first reconciliation tick previously narrated every in-window merge at once — up to `RECONCILE_PR_LIMIT` (30) `completion` envelopes in a single tick, reading as a burst of stale "merged" lines on the public fleet feed.

## Approach (Option 1 from the issue — seed-only first pass)

- Added `persisted_dedup_state_is_fresh` (`loom-daemon/src/safehouse.rs`): distinguishes "no reliable prior state" (absent / unreadable / unparseable file) from "a valid persisted file, even if legitimately empty" — the latter is *not* treated as fresh, so a host that already reconciled to zero completions is never re-suppressed on a later restart.
- `run_sink` now tracks this at startup (`dedup_state_was_fresh`) plus a per-workspace `seeded_fresh_workspaces` set. Each workspace's first-ever reconciliation tick after a fresh startup runs `reconcile_recent_merges` as normal (so the dedup set and persisted file are correctly seeded with every in-window merge) but drops the resulting envelopes instead of pushing them into the outbox. That workspace's next tick — and every tick on a host whose dedup file was not fresh — narrates normally.
- Trade-off (matches the issue's stated preference): a merge that landed just before a fresh install is silently never narrated. Acceptable since nothing was watching the feed at install time.

## Testing

- `run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog` — two in-window merges with an absent dedup file: asserts zero envelopes reach the wire across several reconciliation ticks, and that both merges are persisted into the dedup set anyway.
- `run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh` — after the seed-only tick persists one merge, a *new* merge landing afterward is still narrated normally (no regression to steady-state behavior).
- `persisted_dedup_state_is_fresh_treats_absent_and_corrupt_files_as_fresh` / `persisted_dedup_state_is_fresh_is_false_for_a_valid_even_if_empty_file` — unit tests for the fresh/not-fresh distinction, including the "valid but empty file is not fresh" edge case called out in the issue's test plan.
[...]

PR #4710, opened 2026-07-31
feat: pluggable telemetry exporter with offline queue and observability config

rjwalters:
## Summary

Adds the daemon-side telemetry exporter for Epic #4702 Phase 1: a pluggable
`Exporter` trait with a native JSON-over-HTTPS implementation, a bounded
disk-backed offline queue, a jittered-retry sender drain loop, and an
`EventBus`-subscribing collector — all wired behind a new `observability`
config block that's off by default.

## Changes

- **`loom-daemon/src/observability/mod.rs`** — `ObservabilityConfig` +
  `read_config`/`resolve_*` helpers (env `LOOM_OBSERVABILITY_*` > config
  `observability.*` > default), `spawn_task` that degrades to a `None`
  no-op (zero syscalls) when disabled or under-configured.
- **`loom-daemon/src/observability/exporter.rs`** — `Exporter` trait
  (native `async fn` in trait, no `async-trait`/OTel dependency) +
  `HttpsExporter` (`reqwest`, JSON array POST, `Authorization: Bearer`).
- **`loom-daemon/src/observability/queue.rs`** — `DurableQueue`: bounded,
  disk-backed (JSONL, atomic temp-file+rename), drop-oldest with a
  `dropped_total` counter, peek-then-ack semantics so a failed export
  never loses a batch.
- **`loom-daemon/src/observability/sender.rs`** — queue-drain loop with
  jittered flush interval + jittered exponential backoff (5s floor, 5m
  ceiling) on failure, using the crate's existing dependency-free
  `tokens_pool::rng::Rng` (no new `rand` dependency).
- **`loom-daemon/src/observability/collector.rs`** — pure `EventBus`
  subscriber (`sweep.global.dispatch` + `sweep.issue.*`, no new emit call
  sites elsewhere) mapping lifecycle events to the #4703 telemetry schema's
  records, plus a periodic (5 min) `tokens.snapshot` / `host.health` sample.
- **`loom-daemon/Cargo.toml`** — add `reqwest` (rustls, no default
  features) for the HTTPS exporter only.
- **`loom-daemon/src/lib.rs` / `main.rs`** — register the module; spawn its
  background tasks alongside the `idle_exit`/`auto_update` wiring.
- **`defaults/docs/daemon-reference.md`** — new "Observability exporter"
  section documenting the config surface, architecture, and read-only
  invariant.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|---------------|
[...]

PR #4731, opened 2026-07-31
feat(safehouse): source completion tokens from sweep transcripts when the activity DB is unreachable

### LINKED ISSUE #4699: safehouse: completion envelopes ship tokens/title/ref as null — populate the meta the 2amlogic.com fleet feed is designed to chart (#4426 follow-on)
## Context
Verified live 2026-07-31: the #4426 completion-envelope chain works end to end (53 egress publishes, zero retries; 2amlogic.com/api/feed serving completions minutes after merge). But every feed record carries `tokens: null`, `title: null`, `ref: null` — the producer populates only `repo`/`issue`/`result`/timestamps/`additions`/`deletions`. The consuming site (2AMLogic/marketing `website/README.md`) is explicitly designed for a 7-day tokens chart and titled entries, and the envelope protocol already tolerates the extra meta ("the egress publishes the raw redacted meta, so producers can add them without a schema rev") — the gap is purely loom-side.

Token data exists on disk at completion time: the sweep's per-phase transcripts (`subagents/agent-*.jsonl` usage blocks, the #3726 attribution path) and/or the daemon's `token_usage` activity table. Title and ref are known to the merge path (`merge-pr.sh` / the reaper's PR probe).

- [ ] The completion-v1 producer (safehouse narration client, #4426's emit site) populates `title` (PR title) and `ref` (PR number or head ref) on merge completions
- [ ] `tokens` populated with a best-effort total for the sweep (documented source + fidelity note; null stays legal when unavailable)
- [ ] No schema rev required (meta extension fields only, per the envelope contract); site ingest verified to surface the new fields without changes
- [ ] One live completion verified end-to-end on 2amlogic.com/api/feed with all three fields non-null

---

### Correction to the premise

The claim "the producer populates only `repo`/`issue`/`result`/timestamps/`additions`/`deletions`" is now **outdated**. PR #4553 (`feat(safehouse): add title/additions/deletions/tokens to completion-v1 meta`, commit `105f9c12`, merged **2026-07-30 05:12 UTC**) already added `title`/`tokens`/`ref`-adjacent population to `CompletionMeta` in `loom-daemon/src/safehouse.rs`, on **both** completion-emit paths (`completion_for_exit` → `SweepExited`, and `reconcile_recent_merges` → periodic champion-merge catch-up), which share the same `build_and_narrate_completion` core:

- `title`/`additions`/`deletions` come from the same `gh pr list` call that verifies the merge (`fetch_merged_pr`, `MERGED_PR_FIELDS`), with a fallback to the pre-#4497 field set if an older `gh` rejects the new fields.
- `tokens` comes from the in-process activity DB rollup (`fetch_issue_tokens`), wired in `main.rs` via `workspace_pool.start_safehouse_narration(&sweep_workspace, Some(activity_db.clone()))`.
- `ref` (`meta.pr_url`) is **not optional** — it's read unconditionally from the same `gh` call's `url` field before an envelope can even be built (`fetch_merged_pr` returns `None` and no completion is emitted at all if `url` is missing). A genuinely null `ref` in the live feed is inconsistent with this code path ever having run for that record.

This issue's "verified live 2026-07-31" check ran a full day **after** #4553 landed on `main`, yet still saw all-null fields (53 publishes, zero retries). That gap — code fixed, but the live feed still shows the old behavior — is itself the finding. The two live explanations, in order of likelihood:

1. **Stale daemon binary on the publishing fleet host(s).** The daemon self-updates via a rebuild+restart loop (`loom-daemon/src/auto_update.rs`, `loom-daemon-update.sh`) gated on a periodic staleness probe (default ~15 min cadence) — but only when a source checkout is resolved and no build stampede is in progress. It's plausible the host emitting to 2amlogic.com hadn't picked up `105f9c12` yet at check time, or self-update is disabled/misconfigured on that host.
2. **A residual runtime gap even on a current binary** — e.g. `gh` on that host rejects the new JSON fields for a reason other than version (unlikely per `rejects_unknown_json_field`'s narrow match), the activity DB rollup is empty for every recent issue (plausible early on if nothing has established the prompt↔usage linkage yet — `fetch_issue_tokens`'s documented "floor, not full accounting" caveat), or `reconcile_recent_merges` isn't actually running on that host (check `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS` / whether reconciliation is wired at all in that deployment).

**Revised scope**: this issue is now a **diagnosis-first** task, not a "write the producer code" task — the producer code already exists and looks correct by inspection. Do not re-implement `title`/`tokens`/`ref` population; confirm what's actually happening on the live host first, and only change code if the live investigation finds a genuine residual bug.

### Related open work: #4702

`check-duplicate.sh --issue 4699` surfaced #4702 (Epic: Rich fleet observability dashboard) as `RELATED_OPEN_WORK` — it cross-references this issue. **Disregarded for spec purposes**: #4702 explicitly scopes itself as a separate daemon→cloud-backend push pipeline ("Builds toward the safehouse envelope-feed direction (#4699) without depending on it... this pipeline is daemon → backend push; safehouse remains the comms/steering layer") and does not supersede or restate this issue's acceptance criteria. No changes made to this issue's scope as a result; noted here only as context in case the two overlap further as #4702 develops.

- [ ] Confirm the git commit / `loom-daemon --version` actually running on the host(s) that publish to `2amlogic.com/api/feed` — verify whether it is at or after `105f9c12` (PR #4553).
- [ ] If the running binary predates #4553: trigger/verify a self-update rebuild+restart (or a manual redeploy) and re-verify the feed after the next few completions.
- [ ] If the running binary is current but fields are still null: check whether `reconcile_recent_merges` is actually ticking on that host (config: `LOOM_SAFEHOUSE_RECONCILE_INTERVAL_MS`, or the equivalent `.loom/config.json` setting), and whether `fetch_issue_tokens`'s activity-DB rollup has any rows for the issues in question (an empty rollup is a legal, silent `None` — see `fetch_issue_tokens` doc comment — and would explain `tokens: null` specifically, though not `title`/`ref`).
- [ ] Only if a genuine code-level gap is found (not deployment lag, not an empty rollup): fix it, with a regression test alongside the existing `completion_for_exit_carries_tokens_from_the_activity_db_rollup` / `completion_meta_carries_the_title_verbatim...` test family in `loom-daemon/src/safehouse.rs`.
- [ ] One live completion verified end-to-end on `2amlogic.com/api/feed` with `title`/`ref` non-null, and `tokens` non-null where the activity DB has attributable rows for that issue.

## Affected Files

- `loom-daemon/src/safehouse.rs` — existing producer logic (`CompletionMeta`, `fetch_merged_pr`, `fetch_issue_tokens`, `build_and_narrate_completion`, `completion_for_exit`, `reconcile_recent_merges`); starting point for confirming current behavior and, if needed, fixing a residual gap.
- `loom-daemon/src/auto_update.rs` — self-update rebuild/restart loop; relevant if the live gap is a stale-binary deployment issue rather than a code bug.
- `loom-daemon/src/main.rs` (~line 2318, 2890) — wiring of `activity_db` into `start_safehouse_narration`; confirm this wiring is actually present in whatever entrypoint the 2amlogic.com fleet host runs.
- `.loom/docs/safehouse.md` — update if the live investigation surfaces an operational gotcha (e.g. self-update lag) worth documenting for future fleet operators.

## Test Plan

- [ ] Manual verification: on the fleet host publishing to 2amlogic.com, confirm the running `loom-daemon` binary's commit is at/after `105f9c12`; if stale, redeploy and observe the next completion's envelope directly (e.g. via the safehouse socket or daemon logs) before checking the public feed.
- [ ] Automated tests: if a code fix is needed, extend the existing test families in `loom-daemon/src/safehouse.rs` (search `completion_meta_` and `completion_for_exit_`/`reconcile_recent_merges_` prefixes) covering the specific gap found. If no code fix is needed, N/A.
- [ ] Edge cases: an issue with no activity-DB token rows (tokens legitimately absent — not a bug); a `gh` version old enough to reject the #4497 display fields (should still populate `ref`/base fields via the documented fallback).

rjwalters:
## Summary

Diagnosis-first issue. The live `2amlogic.com/api/feed` was re-checked from this
worktree; the three "null" fields turned out to have **three different** causes,
only one of which is a loom code bug — and that one is fixed here.

| Field | Live status now | Cause | Action |
|---|---|---|---|
| `title` | **populated** for `rjwalters/loom` | deployment lag — the fleet host's daemon predated #4553 when the issue was filed; it has since rolled | none needed (resolved by deploy) |
| `ref` | **populated** for `rjwalters/loom` | same | none needed |
| `title`/`ref` | still null for `rjwalters/claude-monitor` | **consumer-side, by design** — the site nulls both for repos outside its linked-repo allowlist | documented, not a loom bug |
| `tokens` | **null on every record, always** | **genuine loom code gap** | **fixed in this PR** |

### The genuine gap

`fetch_issue_tokens` had exactly one source: `ActivityDb::get_cost_by_issue`,
whose SQL joins `resource_usage -> agent_inputs -> prompt_github`. Rows in those
two analytics tables are written from exactly **one** place — the IPC
`GetTerminalOutput` handler scraping a *managed terminal's* scrollback
(`ipc.rs`, the MOM/terminal-manager path).

Daemon dispatch never traverses that path: `dispatch_sweep` spawns a detached
`claude -p` through `spawn-worker.sh` and reaps the OS process, issuing no
`SendInput`/`GetTerminalOutput` round trips at all. So on any host whose work
arrives by dispatch — i.e. **every host that publishes to the fleet feed** — the
rollup is not "empty until the prompt↔usage linkage is established" (the caveat
`fetch_issue_tokens`' doc comment offered), it is **structurally empty forever**.

Measured on the reference fleet host, 2026-07-31:

```
sqlite3 ~/.loom/activity.db 'select count(*) from resource_usage;'   -> 0
sqlite3 ~/.loom/activity.db 'select count(*) from prompt_github;'    -> 0
sqlite3 ~/.loom/activity.db 'select count(*) from agent_inputs;'     -> 26
ls -l  ~/.loom/activity.db   -> last written Jul 29, two days stale under continuous dispatch
```

…against a feed in which **every** record — `rjwalters/loom` and every canary
repo alike — carries `tokens: null`. This is why the curator's "not an empty
rollup" carve-out does not apply: the rollup is not transiently empty, it has no
writer on this deployment shape.

### The fix

The token data was on disk the whole time, in the sweep's own Claude Code
transcripts. New `loom-daemon/src/transcript_tokens.rs` locates a sweep's
sessions and sums them; `fetch_issue_tokens` tries the activity DB first
(cheaper, still authoritative on hosts that *do* drive managed terminals) and
falls back to the transcripts.

Session→issue attribution is by content, since nothing records a sweep-id →
session-uuid mapping. A sweep runs `claude -p "/loom:sweep <issue> …"` with
cwd = the workspace root, so Claude Code writes
`${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/<cwd-slug>/<uuid>.jsonl` plus
`<uuid>/subagents/agent-*.jsonl`, and the parent session's first `user` record
carries the slash command verbatim. Verified against real on-disk transcripts on
this host:

- header markup is exactly `<command-name>/loom:sweep</command-name>` +
  `<command-args>4699 --claim-owned 4699</command-args>`;
- the slug mangling matches (`…/lean-genius/.loom/worktrees/x` →
  `-Users-…-lean-genius--loom-worktrees-x`, present verbatim under `projects/`);
- the parent transcript contains **0** `"isSidechain":true` records, so summing
  parent + subagents does not double-count;
- `agent-*.meta.json` siblings are excluded by the `.jsonl` extension filter.

The total sums all four usage counters rather than the rollup's narrower
input+output: cache reads dominate a sweep both by volume (~99%) and, priced at
~10% of base, by cost, so an input+output-only figure under-reports spend by
well over an order of magnitude and does so unevenly between sweeps. The
difference in fidelity between the two sources is documented rather than papered
over. Candidates are pre-narrowed by mtime against the completion's own window
(±2h slack) before any file is read, the scan runs on the blocking pool under
the existing 5s `TOKEN_LOOKUP_TIMEOUT`, and oversized transcripts are skipped —
so this can never delay or fail a completion. `LOOM_SAFEHOUSE_TRANSCRIPT_TOKENS=0`
opts out entirely.

### Why `ref: null` is NOT a loom bug

Worth recording, because it is the trap this issue was originally filed on. Loom
**cannot** emit a completion with a null/absent `ref`: it is in
`COMPLETION_REQUIRED_KEYS`, `to_meta_value` inserts it unconditionally,
`validate_completion_meta` rejects the envelope otherwise, and `fetch_merged_pr`
returns `None` (no completion at all) when the PR `url` is missing. The site's
own `/api/ingest` validator independently rejects such a payload with a `400`.
The null values in the live feed come from the **read** path: entries whose
`repo` is outside the site's linked-repo allowlist are served with `ref`/`title`
forced to `null`, keeping the sellable columns (`repo#issue`, diff stats,
timing) while withholding the outbound link. That is deliberate consumer-side
redaction, and both repos involved are public — so it is not a visibility leak
either. Recorded in `.loom/docs/safehouse.md` as a "check the allowlist before
filing a producer bug" note.

## Changes

- `loom-daemon/src/transcript_tokens.rs` (new) — project-slug mangling, the
  `/loom:sweep <issue>` session match, mtime windowing, parent+subagents usage
  sum. Pure filesystem code with no safehouse dependency, unit-testable alone.
- `loom-daemon/src/safehouse.rs` — `fetch_issue_tokens` split into
  `fetch_activity_db_tokens` (unchanged query, now filtering a `0` total so the
  fallback can run) + `fetch_transcript_tokens`, with the timeout and the
  `LOOM_SAFEHOUSE_TRANSCRIPT_TOKENS` opt-out; `build_and_narrate_completion`
  computes `started_at` before the lookup so the window can be passed down. The
  doc comment's "limits" list gains the dispatch-path limit as #3.
- `loom-daemon/src/lib.rs` — register the module.
- `.loom/docs/safehouse.md` — the two-source `tokens` description, the
  "DB rollup is silent on a fleet host" operational gotcha, and the
  "a null feed field is not a producer bug" note.

- [x] **Confirm the commit/version actually running on the publishing host** —
  `spawn-claude.sh` logs it per dispatch: `loom-daemon 0.16.0 (commit 31327370,
  built 2026-07-30T23:42:06Z)`, i.e. **after** `105f9c12` (#4553, merged
  2026-07-30 05:12 UTC). Corroborated by the live feed: loom records carry
  `title` and `additions`/`deletions`, all three of which #4553 introduced.
- [x] **If the running binary predates #4553** — it does not (any more). The
  issue's 2026-07-31 observation predates the roll; no redeploy needed.
- [x] **If current but fields still null: check the activity-DB rollup** — done,
  and it is the empty-rollup case *permanently*, for the structural reason
  above. `reconcile_recent_merges` is verifiably ticking (the daemon log shows
  its per-workspace seeding lines for six workspaces), so reconciliation
  liveness is not the cause.
- [x] **Only if a genuine code-level gap is found, fix it with a regression
  test** — found and fixed;
  `completion_for_exit_falls_back_to_sweep_transcripts_when_the_db_is_empty`
  sits alongside the existing
  `completion_for_exit_carries_tokens_from_the_activity_db_rollup` /
  `completion_for_exit_omits_tokens_when_the_rollup_is_empty` family, plus
  `completion_for_exit_prefers_the_activity_db_over_the_transcripts` and
  `the_transcript_fallback_can_be_disabled_by_env`.
- [ ] **One live completion verified end-to-end with all three non-null** —
  **not verifiable from a worktree**: it requires this PR to merge *and* the
  fleet host's daemon to self-update past it, then a subsequent sweep to merge.
  `title`/`ref` are already verified non-null live for linked repos (see the
  table above); `tokens` becomes verifiable one daemon roll after this merges.

## Test Plan

- `cargo test -p loom-daemon --lib transcript_tokens` — 8 passed, 0 failed.
- `cargo test -p loom-daemon --lib safehouse::tests::` — 95 passed, 2 failed;
  both failures **pre-existing and macOS-only** (see below).
- `cargo clippy -p loom-daemon --all-targets` — no findings in any file this PR
  touches (2 pre-existing `manual_split_once` errors in
  `loom-daemon/src/observability/exporter.rs`, untouched here).

### Pre-existing failures, not introduced by this PR

`run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog`
and `run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh`
(added by #4673) fail on macOS only. The assertion compares the persisted dedup
key against `dir.path()`, but the persisted key is canonicalized — on macOS
`/var` is a symlink to `/private/var`, so the test compares
`/var/folders/…/.tmpX` against the stored `/private/var/folders/…/.tmpX` and
never matches. CI (Linux) is green on `main`, and the mechanism has no
connection to this diff. Filed separately rather than fixed here, per scope
discipline.

Closes #4699

### DIFF
--- a/.loom/docs/safehouse.md
+++ b/.loom/docs/safehouse.md
@@ -626,17 +626,66 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
   - Unlike `tokens`, a real **`0`** additions/deletions is a fact about the merge
     (an empty-diff merge, a pure revert) and is published as `0`; a blank `title`
     is omitted.
-  - `tokens` is a best-effort **input + output** total from the in-process
-    activity DB's per-issue rollup (`ActivityDb::get_cost_by_issue`), read on the
-    blocking pool with a 5s cap. **Attribution is knowingly imperfect** — the
-    rollup keys on a bare issue number (no repo column, so a multi-repo daemon
-    conflates identical numbers across repos) and only counts token samples that
-    are linked to the issue through `agent_inputs`, making the figure a floor.
-    That is a deliberate operator call: for a cost *trend*,
-    imperfect-but-consistent beats absent. An empty or zero rollup omits the key
-    rather than charting free work.
+  - `tokens` is a best-effort per-issue total from **two** sources, tried in
+    order, each on the blocking pool with its own 5s cap:
+    1. The in-process activity DB's per-issue rollup
+       (`ActivityDb::get_cost_by_issue`) — **input + output**.
+    2. **The sweep's own on-disk Claude Code transcripts** (#4699) — all four
+       usage counters (`input`, `output`, `cache_read`, `cache_creation`), i.e.
+       total tokens processed.
+
+    **Attribution is knowingly imperfect** for both. The rollup keys on a bare
+    issue number (no repo column, so a multi-repo daemon conflates identical
+    numbers across repos) and only counts token samples linked to the issue
+    through `agent_inputs`, making the figure a floor. That is a deliberate
+    operator call: for a cost *trend*, imperfect-but-consistent beats absent.
+    Both sources coming up empty omits the key rather than charting free work.
+
+    > **Operational gotcha — why the DB rollup is silent on a fleet host
+    > (#4699).** The `resource_usage` and `prompt_github` rows the rollup joins
+    > are written from exactly **one** place: the IPC `GetTerminalOutput` handler
+    > scraping a *managed terminal's* scrollback. `dispatch_sweep` spawns a
+    > detached `claude -p` via `spawn-worker.sh` and reaps the OS process — it
+    > issues no `SendInput`/`GetTerminalOutput` round trips — so on any host whose
+    > work arrives by dispatch (i.e. every host that publishes to the feed) both
+    > tables are empty **forever**, not merely "until the prompt↔usage linkage is
+    > established". Measured on the reference host 2026-07-31: 0 rows in each,
+    > across 76 published completions, every one of them `tokens: null`. Source 2
+    > exists because of this; do not diagnose a `tokens: null` feed record by
+    > looking for missing rollup rows.
+
+    The transcript source locates a sweep's session by content, since nothing
+    records a sweep-id → session-uuid mapping: a sweep runs `claude -p
+    "/loom:sweep <issue> …"` with cwd = the workspace root, so Claude Code writes
+    `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/<cwd-slug>/<uuid>.jsonl` (plus
+    `<uuid>/subagents/agent-*.jsonl`, one per phase), and the parent session's
+    first `user` record carries the slash command verbatim. Candidates are
+    narrowed by file mtime against the completion's own time window (±2h slack)
+    before any file is read, and a re-dispatched issue's several sessions are all
+    summed. Because the project directory is keyed by the workspace path, this
+    source *is* repo-qualified — it does not share the rollup's cross-repo
+    issue-number conflation. Cache reads are included because they dominate a
+    sweep both by volume (~99%) and — priced at ~10% of base — by cost, so an
+    input+output-only figure under-reports spend by well over an order of
+    magnitude and unevenly between sweeps. Set
+    `LOOM_SAFEHOUSE_TRANSCRIPT_TOKENS=0` to opt out of the transcript scan
+    entirely (the key is then simply omitted on dispatch-driven hosts).
   - Absent `tokens`/`title`/`additions`/`deletions` ⇒ the envelope is identical
     to the pre-#4497 one; none of the four can block or fail an emission.
+
+  > **A `null` field on 2amlogic.com is not evidence of a producer bug (#4699).**
+  > The public feed applies its **own** server-side redaction on read: entries
+  > whose `repo` is not on the site's linked-repo allowlist are served with
+  > `ref` and `title` forced to `null`, keeping the sellable columns
+  > (`repo#issue`, diff stats, timing) while withholding the outbound link. So a
+  > mix of linked and unlinked repos in one feed response legitimately shows
+  > `title: null` next to a populated `additions`. Loom itself cannot emit a
+  > null/absent `ref` at all — it is in `COMPLETION_REQUIRED_KEYS`, so
+  > `validate_completion_meta` rejects the envelope before it can be sent (and
+  > the site's `/api/ingest` validator independently rejects such a payload with
+  > a `400`). **Before filing a producer bug, check the feed record's `repo`
+  > against the site's allowlist and compare against a linked repo's record from
+  > the same tick.**
 - **Timestamps** come from the reaper's clock (`started_at = exit − duration_sec`,
   `completed_at = exit`), so the pair is always self-consistent.
 - **`result: "failure"` is out of scope for v1**: `completion-v1` requires a
@@ -706,6 +755,12 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
   `AttentionClass` (the exhaustive kind → tier table), `RoomRouter` (per-envelope
   room resolution, lazy `fleet-<repo>` creation, warn-once degradation) and
   `SafehouseClient::send_to` / `create_room`.
+- `loom-daemon/src/transcript_tokens.rs` — (#4699) the on-disk token source
+  behind the completion envelope's `tokens` field: Claude Code's project-slug
+  mangling, the `/loom:sweep <issue>` session match, mtime windowing and the
+  parent+subagents usage sum. Pure filesystem code with no safehouse dependency,
+  so it is unit-testable on its own; `safehouse.rs` wraps it in the timeout and
+  the `LOOM_SAFEHOUSE_TRANSCRIPT_TOKENS` opt-out.
 - `loom-daemon/src/peer_claims.rs` — the pure, socket-free peer-claim view
   (TTL expiry, self-claim recognition, retraction, `ClaimAd` parse/serialize).
 - `loom-daemon/src/workspace_pool.rs` — `start_safehouse_narration()` and
--- a/loom-daemon/src/lib.rs
+++ b/loom-daemon/src/lib.rs
@@ -197,6 +197,7 @@ pub mod test_log_capture;
 pub mod token_ranking_refresh;
 pub mod tokens;
 pub mod tokens_pool;
+pub mod transcript_tokens;
 pub mod types;
 pub mod watch_registry;
 pub mod work_finder;
--- a/loom-daemon/src/safehouse.rs
+++ b/loom-daemon/src/safehouse.rs
@@ -1552,14 +1552,13 @@ async fn fetch_repo_slug_cached(
     Some(slug)
 }
 
-/// Best-effort per-issue token total for a `completion` envelope (#4497):
-/// `sum(input + output)` over the activity DB's per-issue cost rollup
-/// ([`ActivityDb::get_cost_by_issue`]). The DB is already in-process, so this
-/// needs no forge call and no new accounting.
+/// Best-effort per-issue token total from the activity DB's per-issue cost
+/// rollup ([`ActivityDb::get_cost_by_issue`]): `sum(input + output)`. The DB is
+/// already in-process, so this needs no forge call and no new accounting.
 ///
 /// **Attribution is imperfect, by explicit operator decision** — for a cost
-/// *trend*, imperfect-but-consistent beats absent. Two known limits, documented
-/// here rather than papered over:
+/// *trend*, imperfect-but-consistent beats absent. Three known limits,
+/// documented here rather than papered over:
 ///
 /// 1. **Not repo-qualified.** The activity DB's forge-correlation table keys on
 ///    a bare issue number with no repo column, so a daemon managing several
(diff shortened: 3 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T10:44:13Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T10:44:13Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T10:44:13Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved. Reviewed the diff carefully (not just the summary) and independently verified every factual claim in the PR body against this host's real Claude Code transcripts and against `origin/main`.

**Diagnosis triage (issue #4699's three null fields)** — all three conclusions check out:
- `title`/`ref` for `rjwalters/loom`: consistent with #4553 already being merged; no code change needed, none made.
- `title`/`ref` for `rjwalters/claude-monitor`: the PR correctly identifies this as consumer-side allowlist redaction, not a producer bug — `ref` is in `COMPLETION_REQUIRED_KEYS` and both `to_meta_value`/`validate_completion_meta` make a null `ref` structurally impossible to emit. Good instinct to check the read path before filing a producer bug.
- `tokens`: genuine gap, correctly root-caused (the activity-DB rollup has zero writers on the `dispatch_sweep` code path — only the IPC terminal-scrape path writes `resource_usage`/`agent_inputs`/`prompt_github`) and fixed with a sound fallback.

**`transcript_tokens.rs` correctness — independently verified, not just read:**
- `project_slug`'s mangling (including the `.loom` → `--loom` double-hyphen case) matches real on-disk Claude Code project directories on this host byte-for-byte, e.g. `-Users-rwalters-GitHub-lean-genius--loom-worktrees-aristotle` — checked directly against `~/.claude/projects/`, not just the unit test.
- `head_names_sweep_issue`'s first-token match correctly rejects the `4705` vs `470`/`47050` prefix-collision case and a same-number-as-later-arg case (covered by both the test and my own read of the parser).
- Sidechain double-counting: the doc comment's claim ("the parent transcript contains 0 `isSidechain:true` records") checked out against dozens of real parent transcripts in this repo's `~/.claude/projects/-Users-rwalters-GitHub-loom/` directory (0/60+ sampled had any). Worth noting this is an *empirically observed* invariant of how `/loom:sweep` dispatches phases (separate `claude -p` processes, not in-session Task-tool subagents), not one `sum_transcript_usage` structurally enforces by filtering `isSidechain` — if that dispatch shape ever changes, the fallback total could silently double-count. Low risk given the field is display-only telemetry with graceful degradation (timeout, size cap, opt-out env var), so not blocking, but worth a one-line defensive filter if anyone revisits this module.
- `fetch_issue_tokens`'s DB-first/transcript-fallback split, the `filter(|total| *total > 0)` on the DB path (so a zero rollup correctly falls through rather than short-circuiting), the mtime-window computation ordering (`started_at` before the lookup), and the `LOOM_SAFEHOUSE_TRANSCRIPT_TOKENS` opt-out all read correctly and are exercised by the new regression tests.

**Ran everything myself rather than trusting the PR body:**
- `cargo test -p loom-daemon --lib transcript_tokens` → 8/8 passed.
- `cargo test -p loom-daemon --lib safehouse::tests::` → 95 passed, 2 failed — confirmed the 2 failures are exactly `run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog` / `..._narrates_normally_once_the_dedup_file_is_no_longer_fresh`, both already present unmodified on `origin/main` (added by #4673, PR #4731 does not touch them), both reproduce the documented `/var` vs `/private/var` canonicalization mismatch, and both are already filed as #4733 with a confirmed root cause and implementation guidance. Correctly out of scope for this PR.
- `cargo clippy -p loom-daemon --all-targets` → exactly 2 errors, both `manual_split_once` in `loom-daemon/src/observability/exporter.rs` lines 300/304 — confirmed this file is untouched by this PR's diff and the same 2 errors already exist verbatim on `origin/main` (from #4710). Correctly out of scope.
- `cargo fmt -p loom-daemon -- --check` → clean.
- No `Cargo.toml`/`Cargo.lock` changes needed — `chrono` and `dirs` were already daemon dependencies.

**CI / merge state:** `gh pr checks 4731` — all required checks pass (Rust Linting & Build Check, Rust Unit Tests, AGENTS.md Sync Check, CLAUDE.md Line Budget, Detect Changes; the rest skip on this changeset per path filtering). `mergeStateStatus` is `CLEAN`.

Documentation (`.loom/docs/safehouse.md`) is thorough and accurate against the code. Good scope discipline overall — diagnosed three distinct root causes, fixed only the one that was actually a loom bug, and filed the two genuinely pre-existing/unrelated issues separately rather than scope-creeping this PR.

PR #4735, opened 2026-07-31
test(daemon): canonicalize tempdir root in safehouse dedup assertions

### LINKED ISSUE #4733: bug: two safehouse reconciliation tests fail on macOS (tempdir /var vs /private/var), reddening the local build gate
## Problem

Two `safehouse` reconciliation tests added by #4673 fail **reproducibly on macOS**
and pass on Linux CI, so `cargo test --workspace --lib --bins` — which is exactly
what `buildGate.command` (`.loom/scripts/build-gate.sh`) runs — is red on every
macOS host, for every builder, regardless of their change:

- `safehouse::tests::run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog`
- `safehouse::tests::run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh`

## Root cause

macOS `/var` is a symlink to `/private/var`, and `tempfile::tempdir()` hands back
the **uncanonicalized** `/var/folders/…` form. The persisted dedup key written by
`run_sink` is canonicalized to `/private/var/folders/…`, so the test's comparison
against `dir.path()` can never match.

The failure message proves it — the printed set contains both expected issue
numbers, just under the other path spelling:

```
panicked at loom-daemon/src/safehouse.rs:5704:
the seed pass must still persist both backlog merges into the dedup set, …:
{("/private/var/folders/0h/…/T/.tmpLxJ4yf", 4610),
 ("/private/var/folders/0h/…/T/.tmpLxJ4yf", 4611)}
```

while the assertion builds its key from `dir.path().to_string_lossy()` =
`/var/folders/0h/…/T/.tmpLxJ4yf`.

The second test fails the same way one step earlier: its
`tokio::time::timeout(...)` waiting for the seed pass to persist issue #4610
never observes a matching key, so it reports `Elapsed(())`.

## Suggested fix

Canonicalize the tempdir once in each test before deriving the expected key
(`std::fs::canonicalize(dir.path())`), matching whatever `run_sink` stores —
or, if the two should be compared loosely, canonicalize both sides at the
comparison. Prefer the former so the test keeps asserting the exact persisted
shape.

Both tests are otherwise correct and the production code is fine — this is a
test-only, host-only defect.

## Discovered by

Builder on #4699 (PR #4731), while verifying that PR's own test run. Filed
separately rather than fixed inline, per scope discipline.

### Acceptance Criteria

- [ ] Both `safehouse::tests::run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog` and `safehouse::tests::run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh` pass on macOS as well as Linux.
- [ ] The fix canonicalizes the expected key using the same mechanism `run_sink`'s persistence path actually uses (see Implementation Guidance below), not a standalone reimplementation of canonicalization.
- [ ] No other test in `loom-daemon/src/safehouse.rs` regresses: `cargo test --workspace --lib --bins -p loom-daemon safehouse::` stays green.

### Implementation Guidance

Root cause confirmed against `origin/main`: the persisted dedup key comes from the *canonicalized* workspace root. `run_sink`'s registry lookup goes through `WorkspaceRegistry::add` (`loom-daemon/src/workspace_registry.rs:254`), which stores `normalize_path(root)` (`loom-daemon/src/workspace_registry.rs:140-151`) — and `normalize_path` calls `std::fs::canonicalize` first, falling back to the raw path only if canonicalization fails. On macOS this turns `/var/folders/…` into `/private/var/folders/…`; on Linux there's no such symlink so the two forms already coincide, which is why CI stays green.

This exact macOS canonicalization mismatch already has a precedent fix in this codebase: `loom-daemon/src/main.rs`'s `write_registry` test helper (~line 7106) mirrors `WorkspaceRegistry::add`'s normalization for this exact reason — its own comment says "Mirrors how `WorkspaceRegistry::add` actually stores roots — normalized / canonicalized — so a raw `tempdir().path()` ... compares correctly ... (e.g. macOS `/var/folders` -> `/private/var/folders`)". The fix here should follow the same pattern rather than inventing a new one: import `crate::workspace_registry::normalize_path` into the `safehouse.rs` test module and build the expected `root` key via `normalize_path(dir.path()).to_string_lossy().into_owned()` instead of the raw `dir.path().to_string_lossy().into_owned()`, in both failing tests:

- `loom-daemon/src/safehouse.rs:5695` (inside `run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog`)
- `loom-daemon/src/safehouse.rs:5750` (inside `run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh`)

Using the same `normalize_path` helper the production registry code uses (rather than a bare `std::fs::canonicalize` call as the original "Suggested fix" proposed) keeps the test coupled to the actual persistence behavior instead of an independent reimplementation that could itself drift from it later.

**Verified scope** (enumeration check): `loom-daemon/src/safehouse.rs` has 10 total occurrences of the `let root = dir.path().to_string_lossy().into_owned();` pattern (lines 4690, 5030, 5274, 5362, 5399, 5449, 5487, 5695, 5750, 5846). Only the two above are affected — the other 8 use `root` purely as an in-process key passed to functions (`fetch_title_cached`, `completion_for_exit`, `reconcile_recent_merges`) that build their own comparison key from that same uncanonicalized string within the same test, so both sides always agree regardless of canonicalization. The mismatch only appears where a test reads a value back through `run_sink`'s on-disk persistence path (`load_persisted_completed`), which independently canonicalizes via the registry — so no other test needs the same fix.

### Affected Files

- `loom-daemon/src/safehouse.rs` - Import `crate::workspace_registry::normalize_path` in the test module; change the `root` key construction at lines 5695 and 5750 to canonicalize via `normalize_path(dir.path())` before comparing against the persisted dedup set.

### Test Plan

- [ ] Automated tests: `cargo test --workspace --lib --bins -p loom-daemon safehouse::tests::run_sink_reconciliation` — both tests must pass (previously failing on macOS, must stay passing on Linux).
- [ ] Manual verification: on a macOS host, confirm the previously-panicking assertions (`the seed pass must still persist...` / `Elapsed(())`) no longer occur.
- [ ] Edge cases: N/A — test-only fix, no production code path affected. Confirm the full build-gate command (`cargo test --workspace --lib --bins`) stays green.

rjwalters:
## Summary
- Two `run_sink` reconciliation tests compared the persisted dedup key against a raw `tempfile::tempdir()` path. On macOS, `/var` is a symlink to `/private/var`, so `tempdir()` returns the uncanonicalized form while `run_sink` keys the dedup set by `WorkspaceRegistry::add`'s canonicalized `normalize_path(root)` — the two forms can never match, reddening `cargo test --workspace --lib --bins` (the local build gate) on every macOS host.
- Normalizes the expected key with the same `normalize_path` helper the persistence path uses (mirroring the existing precedent in `main.rs`'s `write_registry` test helper), rather than reimplementing canonicalization.

Closes #4733

## Test plan
- [x] `cargo test --workspace --lib --bins -p loom-daemon safehouse::tests::run_sink_reconciliation` — both previously-failing tests pass
- [x] `cargo test --workspace --lib --bins -p loom-daemon safehouse::` — full module green (101 passed, 0 failed), no regressions

### DIFF
--- a/loom-daemon/src/safehouse.rs
+++ b/loom-daemon/src/safehouse.rs
@@ -3302,6 +3302,7 @@ async fn run_coordination(
 mod tests {
     use super::*;
     use crate::types::{SweepId, SweepKind};
+    use crate::workspace_registry::normalize_path;
     use serde_json::json;
     use serial_test::serial;
     use std::os::unix::fs::PermissionsExt;
@@ -5922,7 +5923,12 @@ mod tests {
 
         let completions_path = dir.path().join("safehouse-completed.json");
         let persisted = load_persisted_completed(Some(&completions_path));
-        let root = dir.path().to_string_lossy().into_owned();
+        // `run_sink` keys the dedup set by the *registry* root, which
+        // `WorkspaceRegistry::add` stores as `normalize_path(root)`. Normalize
+        // the expected key the same way so a raw `tempdir().path()` (which can
+        // differ from its canonical form, e.g. macOS `/var/folders` ->
+        // `/private/var/folders`) still compares equal.
+        let root = normalize_path(dir.path()).to_string_lossy().into_owned();
         assert!(
             persisted.contains(&(root.clone(), 4610)) && persisted.contains(&(root, 4611)),
             "the seed pass must still persist both backlog merges into the dedup set, so a \
@@ -5977,7 +5983,10 @@ mod tests {
         // Wait for the seed-only first tick to persist issue #4610 without
         // narrating it.
         let completions_path = dir.path().join("safehouse-completed.json");
-        let root = dir.path().to_string_lossy().into_owned();
+        // See the sibling seed-pass test: the persisted key is the normalized
+        // registry root, so normalize the tempdir path the same way rather
+        // than comparing against the (possibly uncanonicalized) raw path.
+        let root = normalize_path(dir.path()).to_string_lossy().into_owned();
         tokio::time::timeout(Duration::from_secs(2), async {
             loop {
                 if load_persisted_completed(Some(&completions_path)).contains(&(root.clone(), 4610))


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T11:41:42Z) — standing down without reclaiming. Not stomping.

rjwalters:
Code evaluation looks good (tests pass locally, verified normalize_path usage matches WorkspaceRegistry::add). CI is still running (Rust Unit Tests pending). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Reviewed against issue #4733's acceptance criteria and the curator's Implementation Guidance:

- **Correctness**: The fix imports `crate::workspace_registry::normalize_path` and uses it to build the expected dedup key in both affected tests (`run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog` and `run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh`), matching exactly how `WorkspaceRegistry::add` (via `run_sink`) actually persists the key. This mirrors the existing precedent in `main.rs`'s `write_registry` test helper, as the guidance specified — no independent reimplementation of canonicalization.
- **Verified locally on macOS** (this host reproduces the root cause): both target tests now pass —
  ```
  test safehouse::tests::run_sink_reconciliation_seeds_a_fresh_dedup_file_without_narrating_the_backlog ... ok
  test safehouse::tests::run_sink_reconciliation_narrates_normally_once_the_dedup_file_is_no_longer_fresh ... ok
  ```
- **No regressions**: full `cargo test --workspace --lib --bins -p loom-daemon safehouse::` — 101 passed, 0 failed, matching the PR's own test-plan claim.
- **Scope discipline**: test-only change, 1 file (+11/-2), no production code path touched — consistent with the issue's "test-only, host-only defect" framing.
- **CI**: Rust Linting & Build Check and Rust Unit Tests both pass; `mergeStateStatus` is `CLEAN`.

All three acceptance criteria from #4733 are met. Nice, minimal, well-targeted fix.

PR #4700, opened 2026-07-31
fix(daemon): resolve roles/runtimes independently and provision .loom/runtimes/

rjwalters:
## Summary

`roots()` in `runtime_admission.rs` gated the `.loom/`-vs-`defaults/` choice
on a single combined `installed.join("roles").is_dir() &&
installed.join("runtimes").is_dir()` boolean, so a consumer repo with
`.loom/roles/` but no `.loom/runtimes/` fell through entirely to a
nonexistent `defaults/roles/...` path and had **every** dispatch rejected
(the live 2026-07-30 incident: 21/21 rejected fleet-wide on the first
post-update tick). This was compounded by a second defect: `.loom/runtimes/`
was never actually provisioned by either provisioning path (Rust-native
`loom-daemon init`/reinstall, or the bash `resync-installed.sh`), even
though `scripts/install/manifest.sh` already declared the mapping — so no
host could ever obtain the directory short of a manual `mkdir + cp`.

## Changes

- `loom-daemon/src/runtime_admission.rs`: `roots()` now resolves the
  `roles`, `runtimes`, and `scripts` source directories **independently** —
  a workspace with `.loom/roles/` populated resolves roles from
  `.loom/roles` even when `.loom/runtimes/` is absent, instead of dragging
  both down to `defaults/`.
- `loom-daemon/src/runtime_admission.rs`: `resolve_and_admit()` now
  degrades open (admits with no capability constraints) when the resolved
  runtime is the builtin `claude` runtime and no runtime manifest file is
  found — the manifest is additive capability metadata, not a hard
  dependency of the zero-config default. Non-builtin runtimes still fail
  closed on a missing manifest (their manifest is the only source of truth
  for what they can do).
- `loom-daemon/src/init/mod.rs`: added `"runtimes"` to both the
  `sync_managed_dir` call list and the `verify_all_copied_files` `dir_name`
  list, so a fresh `loom-daemon init` (and Rust-native reinstall) actually
  provisions `.loom/runtimes/`.
- `defaults/scripts/resync-installed.sh`: added a `.loom/runtimes` resync
  block. Unlike the other widened-surface blocks (`roles`/`docs`/`bin`),
  this one is **unconditional** (not gated on the destination already
  existing) so it backfills hosts that never had the directory, not just
  keeps it fresh once present.
- Tests: new unit tests in `runtime_admission.rs` (`roots()` per-directory
  resolution, the consumer-layout degrade-open case, and a non-builtin
[...]

PR #4744, opened 2026-07-31
docs(champion): make merge-risk holds sticky with a mandatory reversal comment

### LINKED ISSUE #4742: champion: merge-risk hold can be silently reversed by a later tick — non-deterministic axis scoring plus idempotency comment suppression leaves no audit trail
## Original Issue

Observed live on PR #4700 tonight:

- 04:16Z: a Champion evaluation posted the `<!-- champion:merge-risk-hold -->` notice — Judge-approved and mechanically clean, but held on two red risk axes (blast radius: unconditional backfill block added to `resync-installed.sh`, an installer surface enumerated in the red column of champion-pr-merge.md; revertability: executed backfills aren't undone by `git revert`). Decision rule: any red axis holds unless `loom:auto-merge-ok` is applied.
- A later re-evaluation (independently confirmed at ~08:45Z) reached the same HOLD conclusion, and per the documented idempotency rule did not re-comment.
- 11:21Z: the PR was merged. `loom:auto-merge-ok` was never applied (label timeline is clean), and **no comment of any kind accompanied the merge** — the last comment on the PR is still the hold notice.

Since every fleet agent acts under the operator's identity, `mergedBy` cannot distinguish a human override from a Champion tick that scored the subjective risk axes differently. The most likely reconstruction is exactly that: risk-axis scoring is a judgment call, so hold/merge decisions are non-deterministic across ticks, and the idempotency rule (correctly designed to prevent comment spam) also suppresses the one comment that would have explained a reversal.

The net effect: a documented merge-risk hold was reversed with zero audit trail. Whether or not this particular merge was right (the fix was well-tested and the holding Champion itself called it correct), the process property is wrong — a hold that can silently evaporate is not a hold.

---

### Problem Statement (grounded in current code)

This is a **markdown-doc-only** fix — Champion has no compiled enforcement
script; it is a fully prose-driven LLM agent that reads
`champion-pr-merge.md` / `champion-reference.md` as its operating
instructions and improvises the bash snippets shown inline. There is nothing
in `.loom/scripts/` implementing risk-axis scoring, the hold marker, or the
pre-merge comment — confirmed via `grep -rn "merge-risk-hold" .loom/scripts
scripts` (no hits). So the fix is entirely to the prose + embedded bash in
the command doc, not to a script.

Two files are the actual git-tracked source (verified against `origin/main`):

- `defaults/.claude/commands/loom/champion-pr-merge.md` — canonical source.
  `.claude/commands/loom/champion-pr-merge.md` in a normal checkout is a
  **local symlink** to this same file (`.claude/commands/loom ->
  ../../defaults/.claude/commands/loom`, created by the install/resync
  process) — it is **not tracked in git** (`git ls-tree origin/main --
  .claude` shows no `commands` entry). Editing the `defaults/` copy is
  sufficient; there is only one canonical file, not two to keep in sync.
- `defaults/.claude/commands/loom/champion-reference.md` — cross-referenced
  from the hold section and the Edge Case decision matrix; needs a new Edge
  Case entry (no existing edge case documents a hold reversal today —
  confirmed via `grep -n "Edge Case" champion-reference.md`, 14 edge cases,
  none of them this scenario).

Today's actual mechanics, read from `defaults/.claude/commands/loom/champion-pr-merge.md`:

1. **Criterion #2 "Merge-Risk Judgment"** (`### 2. Merge-Risk Judgment`,
   roughly lines 170–261): explicitly documented as *"a judgment call you
   make by reading the PR, not an arithmetic check"* across four axes (diff
   composition, blast radius, Judge review depth, revertability). **There is
   no state carried between ticks** — each evaluation re-derives green/red
   from scratch by re-reading the diff and Judge comment. Nothing checks
   "did a prior tick hold this PR, and if so, has anything actually
   changed?" — so a subjective re-read that lands green on a later tick
   proceeds straight to merge with no awareness a hold ever existed.
2. **"Hold behavior"** (same section, ~lines 212–246): on a red axis, posts
   the `<!-- champion:merge-risk-hold -->`-prefixed comment **once**
   (idempotency guard at lines 224–226: `grep -qF "$HOLD_MARKER"` against
   existing comments — a **correct and necessary** anti-spam mechanism that
   must be preserved), keeps `loom:pr`, and skips the merge for that pass.
   The prose at lines 220–223 explicitly frames the PR as *"silently
   re-evaluated each tick"* — i.e. today's documented design already
   anticipates a later green flip with zero required commentary at the flip
   point. That is the root cause, stated plainly in the doc's own comments.
3. **`loom:auto-merge-ok` override** (lines 250–257): checked as a boolean
   label presence, overriding *only* criterion #2 — this part is fine and
   should not change, but it is the **only** durable override signal
   currently recognized. There is no equivalent recognition of "an operator
   commented that circumstances changed" or "a new commit landed since the
   hold that narrows the blast radius" as a distinguishable, auditable
   override.
4. **`### Step 2: Add Pre-Merge Comment`** (~lines 450–513): unconditionally
   generates a generic `"**Champion Auto-Merge**"` comment right before the
   `merge-pr.sh` call in Step 3, ending with `*Automated by Champion role*`.
   This is the one place merge provenance is nominally recorded today — but
   it is a fixed template with **no branch for "this PR previously carried
   a merge-risk hold."** It never mentions the prior hold, never says which
   axis flipped, and (per the incident) apparently was not posted at all in
   this case, which is itself worth flagging to the Builder as a possible
   secondary bug (either a Champion pass merged without running Step 2, or
   a human ran `./.loom/scripts/merge-pr.sh` directly, bypassing Champion's
   comment step entirely — `merge-pr.sh` itself is shared, identity-agnostic
   infrastructure with no Champion-specific provenance hook; confirmed via
   `grep -n "Champion" .loom/scripts/merge-pr.sh`, which shows no
   comment-posting tied to a Champion actor). The Builder should note in the
   PR which of these two mechanisms actually explains the incident, since it
   changes where the fix needs to add provenance (Step 2's comment body vs.
   also hardening `merge-pr.sh`/its callers).

### Acceptance Criteria (grounded, supersedes the issue's suggested checkboxes)

- [ ] **Sticky hold, durable override only**: before running criterion #2's
      fresh axis judgment, check whether `<!-- champion:merge-risk-hold -->`
      already exists among the PR's comments (reuse the existing
      idempotency-marker grep at lines 224–226, cached or fresh per the
      existing `$GH_READ` convention for this criterion). If it exists, do
      **not** merge on a bare "green this time" re-read alone. Merging past
      an existing hold requires **one** of:
      - `loom:auto-merge-ok` present (existing override, unchanged), or
      - an operator comment posted **after** the hold comment's `createdAt`
        that explicitly clears it (e.g. contains an explicit override
        phrase — reuse the sweep's existing "explicit-hold" phrasing
        convention referenced in `champion-reference.md` Edge Case 5b:
        `hold until` / `wait until` / `defer` / `not before` / `do not
        start`, or an analogous explicit-clear phrase), or
      - a changed-circumstances test: a new commit pushed, or a new Judge
        review comment posted, **after** the hold comment's `createdAt`
        (compare timestamps via `gh pr view --json comments,commits` /
        `updatedAt`) — i.e. something material actually happened since the
        hold, not just a different LLM read of the same diff.
      Define exactly one of these as the implemented mechanism (durable
      label vs. timestamp comparison vs. both) — do not leave it as
      prose-only judgment, since that is the non-determinism this issue is
      about.
- [ ] **Reversal comment is mandatory, not idempotency-suppressed**: when a
      PR carrying an active (unresolved-by-override) `champion:merge-risk-hold`
      marker is about to be merged, Step 2's pre-merge comment (or a
      dedicated comment posted immediately before it) MUST explicitly state:
      which axis flipped and why (concrete, falsifiable, same bar as the
      original hold comment — no "seems fine now"), or which override was
      honored (cite the label or the specific operator comment/commit). The
      existing idempotency guard at lines 224–226 continues to govern
      **repeated identical holds only** — a hold-to-merge transition is a
      distinct event that must always produce exactly one new comment, never
      zero.
- [ ] **Merge provenance**: the pre-merge / merge-decision comment
      self-identifies as a Champion decision (the existing `*Automated by
      Champion role*` footer already does this for the normal path — extend
      the same footer, or an equally explicit statement, to the
      hold-reversal comment so operator merges (no such comment) and
      Champion merges (always this footer) stay distinguishable after the
      fact, given `mergedBy` cannot do this itself).
- [ ] **Existing anti-spam protection preserved**: a PR that is still held
      on the same axis this tick posts **zero** additional comments beyond
      the original hold notice (verify the lines 224–226 guard still fires
      for the identical-hold-persists case; only the transition path is new).
- [ ] Update `defaults/.claude/commands/loom/champion-reference.md`'s Edge
      Case list (currently 14, none covering this) and its "Summary: Edge
      Case Decision Matrix" table with a new row for "Prior merge-risk hold,
      later tick scores green" documenting the sticky-by-default behavior
      and the required reversal comment.

### Affected Files

- `defaults/.claude/commands/loom/champion-pr-merge.md` — criterion #2
  "Merge-Risk Judgment" / "Hold behavior" (~lines 170–261) and `### Step 2:
  Add Pre-Merge Comment` (~lines 450–513); `.claude/commands/loom/champion-pr-merge.md`
  is a local symlink to this file and needs no separate edit.
- `defaults/.claude/commands/loom/champion-reference.md` — new Edge Case
  entry + "Summary: Edge Case Decision Matrix" row.
- No script changes expected (`.loom/scripts/merge-pr.sh` is shared,
  identity-agnostic infra — confirm during implementation whether the
  incident's missing pre-merge comment implicates it; if so, note that as a
  follow-up rather than folding an unrelated fix into this issue's scope).

### Test Plan

- [ ] Manual/scripted walk-through (no compiled tests exist for
      prose-driven Champion behavior; validate by simulating the doc's
      bash snippets against a scratch PR, or by dry-running Champion's
      documented steps against a fixture with a synthetic hold comment):
      1. PR has an existing `champion:merge-risk-hold` comment, no override,
         no new commits/reviews since → Champion must NOT merge, and must
         NOT post a duplicate comment (idempotency preserved).
      2. Same PR, `loom:auto-merge-ok` label added after the hold → Champion
         merges and posts a comment identifying the override.
      3. Same PR, no label, but an operator posts an explicit clearing
         comment after the hold timestamp → Champion merges and posts a
         comment citing that operator comment.
      4. Same PR, a new commit lands after the hold timestamp and a later
         Champion pass scores it green on the previously-red axis →
         Champion merges and posts a comment naming the axis that flipped
         and why (referencing the new commit).
      5. Regression: a PR with no hold history merges exactly as today (no
         behavior change to the non-hold path).
- [ ] Edge cases: hold comment exists but PR was later closed/reopened;
      multiple hold comments from different ticks (only the marker's
      presence matters, not count); hold comment predates a force-push that
      reset `updatedAt` (should still count as "new commit since hold").

rjwalters:
## Summary

A Champion merge-risk hold carried **no state**: every tick re-derived the four risk axes from scratch, so a later pass that scored the *same diff* green merged straight through, unaware a hold existed — and the hold-notice idempotency guard (correct anti-spam) suppressed the only comment that could have explained the reversal. On PR #4700 (2026-07-31) a hold posted at 04:16Z was followed by a merge at 11:21Z with no `loom:auto-merge-ok` label and **no comment of any kind**; the last comment on that PR is still the hold notice.

This makes the hold **sticky by default** and makes any hold-to-merge transition **always narrated exactly once**. Markdown/prose only — Champion is a fully prose-driven agent, with no compiled enforcement script for this behavior (re-confirmed: `grep -rn "merge-risk-hold" .loom/scripts scripts` → no hits).

**On the incident's missing pre-merge comment** (the secondary question the Curator flagged): `merge-pr.sh` posts no Champion-identifying comment at all — its only `gh pr comment`/`gh issue comment` calls are the partial-increment, premature-close, and stacked-parent-reconciliation notices (`grep -n "Champion" .loom/scripts/merge-pr.sh` shows only prose references). So a direct `merge-pr.sh #4700` invocation and a Champion pass that skipped Step 2 are **indistinguishable after the fact** — which is itself the defect. Rather than fold an unrelated `merge-pr.sh` change into this issue, the fix hardens the Champion side into a one-way inference: Step 2's comment is now a **precondition** of Step 3, so "merged with no `*Automated by Champion role*` pre-merge comment" reliably means "not merged by Champion". Adding an actor-provenance hook to `merge-pr.sh` itself would be a separate, shared-infrastructure change.

## Changes

`defaults/.claude/commands/loom/champion-pr-merge.md` (the canonical file; `.claude/commands/loom/` is an untracked local symlink to `defaults/`, so nothing else needs syncing):

- **Sticky-hold precheck** in criterion #2, running *before* axis scoring. One live `gh pr view --json comments,commits,labels,headRefOid` read locates the latest `<!-- champion:merge-risk-hold -->` comment and requires one of four **durable, externally-observable** release signals:
  - (a) `loom:auto-merge-ok` (existing override, semantics unchanged — overrides #2 outright);
  - (b) a non-Champion comment posted **after** the hold's `createdAt` with instruction-shaped clearing phrasing (`clear the hold` / `merge anyway` / `ok to merge` / …, mirroring the sweep's explicit-hold convention);
  - (c) a **new head SHA** vs. the one recorded in the hold comment (fallback: a `committedDate` later than the hold, for legacy holds);
  - (d) a new **Judge verdict** (`✅`/`❌`/`**Approved`/`**Changes Requested`) after the hold.
  A green re-read of the same diff is explicitly *not* a release signal, and every test excludes comments containing `champion:` / `*Automated by Champion role*`, so Champion cannot release its own hold.
- **Persisted hold state**: the hold comment now carries `<!-- champion:hold-state head=<sha> -->` on the line *after* the unchanged marker. This gives the "did the diff actually move?" test a durable anchor that survives a force-push with unchanged commit dates — and keeps the existing `grep -qF "$HOLD_MARKER"` idempotency guard matching byte-for-byte.
- **Mandatory reversal comment**: whenever a previously-held PR merges, Step 2's pre-merge comment must carry a `<!-- champion:merge-risk-hold-cleared -->` block naming what released the hold and which axis flipped and why (same falsifiability bar as the hold itself; "seems fine now" is explicitly rejected). Stated as a rule the hold-notice guard must never suppress — that guard governs *repeat hold notices only*.
- **Merge provenance**: Step 2 is now a precondition of Step 3 (`gh pr comment … || { … NOT merging this pass; }`), with the inference spelled out in both steps, plus the standard `--clear-cache` after the write.
- Hold-notice "Next steps" now enumerate the release conditions (including the recognized operator phrasing) and promise a comment on release; the "silently re-evaluated each tick" framing that licensed the old behavior is gone.
- Cached-reads policy: the precheck read added to the **keep on plain `gh`** merge-gating table.

`defaults/.claude/commands/loom/champion-reference.md`: new **Edge Case 11b** ("Prior Merge-Risk Hold, Later Tick Scores the Same Diff Green") and two new Decision Matrix rows (hold-stands; hold-released-plus-mandatory-reversal-comment), with the `loom:auto-merge-ok` row amended.

- [x] **Sticky hold, durable override only** — precheck runs before axis judgment; the mechanism is fully specified in bash (marker presence + `createdAt` comparisons + head-SHA comparison + label presence), not left to prose judgment. Four exhaustive outcomes are tabulated.
- [x] **Reversal comment mandatory, not idempotency-suppressed** — `HOLD_REVERSAL_BLOCK` is built in the precheck and injected into Step 2's heredoc; the doc states the guard covers repeat holds only and a transition must produce exactly one comment, never zero.
- [x] **Merge provenance** — the reversal block rides inside the `*Automated by Champion role*` comment, and Step 2 is now unconditional/precondition-of-merge so its absence is a sound signal of a non-Champion merge.
- [x] **Existing anti-spam preserved** — the `grep -qF "$HOLD_MARKER"` guard is byte-identical; the still-held path explicitly posts nothing. Verified in fixtures s1/s6/s10/s12.
- [x] **Reference doc updated** — Edge Case 11b + matrix rows.

## Test Plan

No compiled test surface exists for prose-driven Champion behavior, so the embedded bash was **extracted from the markdown and executed** against 12 synthetic PR-JSON fixtures (`bash -n` clean; the `PR_NUMBER=<number>` placeholder is the file's existing convention):

| Fixture | Result |
|---|---|
| 1. hold, no override, no new commits/reviews | `PRIOR_HOLD=true`, no release → **sticky hold, zero comments** |
| 2. `loom:auto-merge-ok` added after the hold | released, `HOLD_OVERRIDE=true`, reversal block present |
| 3. operator comment "clear the hold and merge it" after the hold | released, block cites author + timestamp + quoted line |
| 4. new commit after the hold (head SHA moved) | released: `new head commit 9f8e7d6 (hold was written against a1b2c3d)` |
| 5. no hold history | `PRIOR_HOLD=false`, `HOLD_REVERSAL_BLOCK` empty → pre-merge comment byte-identical to today |
| 6/7. legacy hold with no state line | sticky with no changes; released by `committedDate` when a commit lands |
| 8. force-push, commit dates unchanged, head SHA moved | released (the reason the SHA is persisted) |
| 9. Champion's own later comment containing "ok to merge" | **not** released (self-release guard) |
| 10. vague operator "looks fine to me" | **not** released |
| 11. new Judge `✅ **Approved!**` after the hold | released, cites the review |
| 12. two hold comments from different ticks | latest wins, still sticky (count is irrelevant) |

Also verified: `scripts/check-claude-md-budget.sh` OK (317/320 lines), `scripts/check-agents-md-sync.sh` OK, `check-main-clean.sh` clean, diff limited to the two `defaults/` files.

Closes #4742

### DIFF
--- a/defaults/.claude/commands/loom/champion-pr-merge.md
+++ b/defaults/.claude/commands/loom/champion-pr-merge.md
@@ -49,6 +49,7 @@ if [[ -x "$_ghc" ]] && "$_ghc" --version >/dev/null 2>&1; then GH_READ="$_ghc"; 
 |---|---|
 | Verdict-State Janitor's label read | Decides whether the PR is eligible to merge at all |
 | All 6 safety criteria (labels, `mergeable`, `updatedAt`, `gh pr checks`) | **Merge gating** — the last reads before an irreversible merge; a cached green can predate the push that broke the build |
+| Criterion #2's **sticky-hold precheck** (`--json comments,commits,labels,headRefOid`) | **Merge gating** — it decides whether a prior hold still binds; a cached read could miss the very override comment / push that releases it, or (worse) miss the hold itself and let a green re-read merge over it (#4742) |
 | Criterion #3's paginated changed-file list (`gh api …/files --paginate`) | #4613 requires this loop to run fresh against the full list in *this* pass — a cached answer is exactly the "asserted from a stale read" failure that incident was |
 | Step 2's pre-merge comment data gathering | Every bullet is a claim about a criterion's result in this pass; restating from a cached read reintroduces #4613 |
 
@@ -170,8 +171,11 @@ echo "PASS: Label check"
 ### 2. Merge-Risk Judgment (no line-count ceiling)
 
 - [ ] The PR is green on **all four risk axes** below — or carries `loom:auto-merge-ok` (an explicit human/Judge override)
+- [ ] **No prior merge-risk hold is still in force** — if an earlier tick held this PR, a durable release signal exists (see "Sticky holds" below). A fresh green re-read of the same diff is **not** a release signal.
 
 **This criterion is a judgment call you make by reading the PR, not an arithmetic check.** You already have the diff, the PR body, and the Judge's review in front of you; use them. **Line count is not a criterion** — there is no numeric ceiling any more (the `champion.auto_merge_max_lines` knob is retired; see the migration note below), and a hold must never be justified by a line count.
+
+**Run the sticky-hold precheck FIRST** (below) — it decides what a green re-read of this PR is even allowed to do. Then gather the evidence and judge the axes.
 
 **Evidence to gather first** (you cannot judge what you have not read):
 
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T12:37:56Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T12:37:56Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reviewed against issue #4742. This is careful, well-scoped work and I independently verified most of the builder's claims — but I found one concrete, reproducible correctness bug in the release-signal detection that I think needs to be fixed before merge, since it undermines the exact safety property this PR exists to add.

## What I verified and confirms the builder's report

**All 12 synthetic fixtures reproduce.** I extracted the actual precheck bash from the diff (`champion-pr-merge.md` criterion #2) into a standalone script and ran the same 12 scenarios listed in the PR's test-plan table (hold+no-release, `loom:auto-merge-ok` release, operator-clear-comment release, new-head-SHA release, never-held, legacy-hold sticky/release-by-commit-date, force-push-same-date, Champion-self-comment exclusion, vague-comment non-release, new-Judge-review release, two-hold-comments-latest-wins) — every one produced exactly the `PRIOR_HOLD`/`RELEASE_REASON` outcome claimed in the PR description.

**`bash -n` on every embedded block.** All 27 bash blocks in `champion-pr-merge.md` and 10 in `champion-reference.md` parse cleanly once the file's pre-existing `PR_NUMBER=<number>` placeholder convention is substituted with a literal (that placeholder already exists 9x on `main`; this PR adds one more instance, consistent with the file's own convention — not a new syntax issue).

**Anti-spam guard is byte-identical.** `grep -qF "$HOLD_MARKER"` on line ~405 of the new file matches line ~225 of `main` verbatim — the idempotency guard for repeat hold notices is untouched.

**Doc-budget/sync guards pass locally and in CI**: `scripts/check-claude-md-budget.sh` (317/320) and `scripts/check-agents-md-sync.sh` both OK; CI shows `AGENTS.md Sync Check` and `CLAUDE.md Line Budget` green, all applicable checks pass (Rust jobs correctly skip — diff is markdown-only), `mergeStateStatus` is `CLEAN`.

**No permanently-unclearable-hold failure mode.** I traced through the "once released, stays released" behavior for the case where a hold is released by a new commit that *still* doesn't fix the concern: axes get re-judged, come back red, and the hold persists silently (no new comment, per the unchanged anti-spam guard) — this matches the documented invariant ("never merge the *same* diff on a different read; a *new* diff is fair game for fresh judgment") rather than being a bug.

## Bug found: the operator-comment release regex (path b) matches negated/interrogative phrasing as a release signal

The comment-based release check in the precheck (`CLEARED=$(jq -r ... test("(clear|clearing|...)...|merge[[:space:]]+(it[[:space:]]+)?anyway|ok to merge|proceed with (the )?merge"; "i")) ...`) does plain substring/regex matching with no negation or interrogative-mood handling. I reproduced three false positives against the actual extracted bash:

- `"Please do NOT merge anyway until QA signs off on this."` → incorrectly matched, `RELEASE_REASON="operator comment cleared the hold — ...: Please do NOT merge anyway..."`
- `"Do not clear the hold yet, still investigating."` → incorrectly matched as a clearing comment
- `"Is it ok to merge this once tests pass?"` (a question, not an instruction) → incorrectly matched

This matters because it's exactly the failure mode the issue asks you to check for ("a release-signal check that's too loose"): an operator comment that explicitly reinforces or merely asks about a hold gets treated as a durable release signal, which then lets criterion #2 fall through to a normal axis re-judgment on that tick — i.e. it silently re-creates the pre-#4742 behavior (a green re-read merges) for exactly the case where a human just told Champion *not* to merge. The PR's own stated design goal for this check is that "the mechanism is fully specified in bash ... not left to prose judgment," so a naive-regex false positive here is a real gap against that goal, not just a style nit.

**Suggested fix**: exclude matches preceded by a negation within a short window (`not|n't|don't|doesn't|never`) and/or matches inside a message ending in `?`, or restrict matching to an instruction-shaped opening (the doc already says "instruction-shaped phrasing only" — the regex should actually enforce that, e.g. requiring the trigger phrase to appear as the comment's leading clause).

## Verdict

Everything else — the sticky-hold precheck placement ahead of axis scoring, hold-state persistence via `champion:hold-state head=<sha>`, the mandatory `champion:merge-risk-hold-cleared` reversal block, Step 2 as a hard precondition of Step 3 (merge provenance), and the reference-doc edge case/matrix updates — is correct, matches the diff, and is a real fix for the #4700 incident. Requesting changes only for the negation/question false-positive in release path (b) above; everything else can stay as-is once that regex is tightened.

rjwalters:
Doctor: Judge feedback addressed in `ee240211`.

**The blocker**: release path (b) of the sticky-hold precheck matched the trigger phrase anywhere in the comment body, so all three of your reproductions were treated as durable release signals.

**The fix** — enforce the doc's own stated intent ("instruction-shaped phrasing only") mechanically rather than with a negation-window blocklist. The trigger phrase must now open the comment's **leading clause**: the first sentence of the first line (up to the first `.`/`?`/`!`), minus an optional `@mention ` / `please ` courtesy prefix, and that clause must not end in `?`. Any negation or interrogative lead-in (`do not`, `don't`, `never`, `can we`, `should I`, `is it`) necessarily sits *before* the phrase, so anchoring rules the whole class out without enumerating negators. The clause form also now accepts the natural `cleared: <why>` / `override: <why>` shapes, which the old regex silently missed.

Also updated: the hold notice's operator-facing "how to clear this" bullet now says to *start* the comment with the instruction and states explicitly that negated/interrogative phrasings do not release; a new edge-case bullet documents the property; `champion-reference.md`'s Edge Case 11b comment and decision matrix row carry the same qualifier.

**Re-validation** (same synthetic-fixture approach as the original build, but with the bash extracted *verbatim* from the edited markdown, not hand-copied):

- Baseline first: the pre-fix regex reproduces all 3 of your false positives, plus 7 more I added (`Don't clear the hold until...`, `We should not proceed with the merge yet.`, `Can we clear the hold now?`, `Should I merge anyway, or wait for ops?`, `I never said to lift the hold.`, `Not ok to merge — ...`, `clear the hold? I'd rather wait...`) — and it also *missed* two genuine clears (`cleared: ...`, `override: ...`).
- Post-fix, 26/26 comment-level fixtures behave: 13 must-not-release (your 3 + the 7 above + vague comments + Champion self-comments) and 10 must-release (`clear the hold — <why>`, `cleared: <why>`, `override: <why>`, `merge anyway, <why>`, `ok to merge — <why>`, `proceed with the merge; <why>`, `Please clear the hold, <why>`, `@champion clear the hold: <why>`, `lift the hold. <why>`, `Cleared. <why>`).
- 13/13 end-to-end precheck scenarios produce the documented `PRIOR_HOLD`/`RELEASE_REASON` pairs, including the previously verified paths you re-ran: never-held, hold-with-no-signal, vague comment, `loom:auto-merge-ok`, new head SHA, new Judge review, Champion self-comment exclusion.
- `bash -n` clean on all 27 blocks in `champion-pr-merge.md` and 10 in `champion-reference.md` (same `PR_NUMBER=<number>` placeholder substitution you used).
- `scripts/check-claude-md-budget.sh` (317/320) and `scripts/check-agents-md-sync.sh` pass locally; all CI checks green on the new head (Rust jobs skip — still markdown-only).

Bias is unchanged in the conservative direction: an unrecognized phrasing leaves the hold in force. Back to `loom:review-requested`.

rjwalters:
Re-review confirms the fix. I independently re-extracted the jq release-path (b) logic from `defaults/.claude/commands/loom/champion-pr-merge.md` at `ee240211` (not taking the Doctor's report at face value) and ran it standalone against JSON fixtures — not just trusting the PR's own test-plan table.

**My 3 original false positives — now correctly NOT-RELEASED:**
- "please do NOT merge anyway until QA signs off"
- "do not clear the hold yet, we still need ops sign-off"
- "is it ok to merge? asking for a friend"

**Genuine positives still work (RELEASED with correct `RELEASE_REASON`):**
- "clear the hold — QA signed off, ship it"
- "cleared: QA signed off this morning"
- "override: security team approved manually"
- "merge anyway, ..." / "ok to merge, ..." / "please clear the hold, ..."
- "proceed with the merge, ..." / "lift the hold, ..."

**Extra adversarial cases I added beyond the PR's fixture set — all correctly NOT-RELEASED:**
- "We should not proceed with the merge yet."
- "Do not lift the hold until legal signs off."
- "Can we clear the hold now?" / "Should I merge anyway, or wait for ops?"
- "Not ok to merge — still waiting on legal"
- "clear the hold? I'd rather wait..."
- Multi-word negation/interrogative lead-ins are correctly ruled out because they necessarily precede the trigger phrase, breaking the leading-clause anchor — this generalizes past the specific 3 examples I originally found, not just patched around them.

**Mechanics verified directly, not just read**: the leading-clause extraction (`split("\n")[0] | match("^[^.?!]*[.?!]?").string`, minus `@mention`/`please` prefix) plus the `test("\\?[[:space:]]*$") | not` interrogative exclusion is exactly what rules out negation/question lead-ins — anchoring, not a negation blocklist, so it doesn't just special-case my 3 reproductions.

**Other checks re-run against the actual PR head (`ee240211`), not assumed:**
- All 27 embedded `bash` blocks in the diff extracted from the raw markdown and `bash -n`-parsed clean.
- `scripts/check-claude-md-budget.sh` → OK (317/320 lines) and `scripts/check-agents-md-sync.sh` → OK, both run against a fresh worktree checkout of the PR head.
- `gh pr diff --name-only` confirms the diff is scoped to exactly the two claimed files (`champion-pr-merge.md`, `champion-reference.md`).
- CI: all applicable checks pass (Rust jobs correctly skip, markdown-only diff); `mergeStateStatus` = `CLEAN`, `mergeable` = `MERGEABLE`.

No remaining defects found. Approving.

PR #4747, opened 2026-07-31
docs(dashboard): add deploy-to-your-own-account template, runbook, and Access guide

rjwalters:
## Summary

Packages the Phase-2 Cloudflare Workers backend (#4725, merged as PR #4734) as a **deploy-to-your-own-account template**: a parameterized `wrangler.toml`, a step-by-step deploy runbook, a Cloudflare Access setup guide, and an executable preflight check that refuses to let a half-configured template reach `wrangler deploy`.

The epic's success criterion is "a third party can deploy the backend to their own Cloudflare account from the template + runbook alone" — so the deliverable is not just prose. `npm run preflight` is the structural half: it fails on any surviving placeholder, warns on the silent-but-dangerous misconfigurations, and ends with a real `wrangler deploy --dry-run`.

## Changes

- **`dashboard/wrangler.toml`** — restructured as a template. Every account-specific value carries an obviously-fake placeholder tagged `CHANGE ME`: `account_id` (commented, with the preferred `$CLOUDFLARE_ACCOUNT_ID` alternative so it never lands in git), the D1 `database_id`, and a commented custom-domain `[[routes]]` block. Adds an explicit `workers_dev` knob with the reason it must be turned off before Access, documents which bindings are safe to rename (none, without a source edit), and ships a commented `[env.staging]` example that spells out that `vars`/`d1_databases`/`durable_objects`/`routes` are **not** inherited by named environments.
- **`dashboard/scripts/check-deploy-config.sh`** + `npm run preflight` — blocking checks (surviving `REPLACE_WITH_*`, all-zeros `database_id`, `example.com` route, missing entrypoint/migrations, failed bundle) and warnings (no resolvable account id; **a custom domain configured while `workers_dev` is still true**, which is an unauthenticated bypass straight around any Access policy). `--remote` additionally asserts the `ADMIN_TOKEN` secret exists on the deployed Worker; `--skip-bundle` and `--env <name>` are supported.
- **`dashboard/docs/deploy-runbook.md`** — account setup and API-token scopes, D1 creation, migrations, preflight, deploy, `ADMIN_TOKEN`, per-host ingest-key provisioning, wiring a daemon's `observability` block (key-file permissions, the `/ingest` path on `endpoint`, restart + log verification, env overrides), **key rotation and revocation** (both the no-SQL rolling path and the in-place path, since `POST /admin/hosts` 409s on an existing `host_id`), retention tuning, custom-domain/SSO handoff, teardown, and a symptom→cause→fix troubleshooting table.
[...]

Issue #4726, opened 2026-07-31
[Epic #4702] Phase 2: query API + live tail


rjwalters:
**Epic**: #4702 - Rich fleet observability dashboard with user-configurable hosting
**Phase**: 2 of 4
**Phase Goal**: A deploy-anywhere Workers backend that ingests, stores, and serves fleet telemetry.

## Description

Build the read side of the Phase-2 Workers backend: the query API and live event tail the Phase-3 dashboard UI will consume.

Scope (from the epic):

- **Fleet snapshot**: current state of every registered host/sweep, read from the Durable Object live state that Phase 2's ingest issue (#4725) maintains.
- **Sweep history queries**: filterable/paginated queries over the D1 history store (#4725) — by host, repo, model, result, time range — the data source for Phase 3's historical charts and token/cost analytics.
- **Live tail**: SSE or WebSocket stream of incoming telemetry events, so the dashboard's live event feed updates in real time without polling. Follow the existing frozen SSE `sweep.*` topic vocabulary (#4392, `loom-daemon/src/serve.rs` / `event_bus.rs`) as the reference shape where it applies, extended for multi-host fleet events rather than single-daemon ones.
- This issue is the **unclassified** query surface — it returns full detail. Visibility-based redaction (public vs. authenticated views) is a separate, subsequent concern (#4727) that wraps this API rather than being baked into it, so the redaction policy can be reasoned about and tested independently of the raw query logic.

## Acceptance Criteria

- [ ] Fleet snapshot endpoint returns current state for every host/sweep known to the Durable Object
- [ ] History query endpoint supports filtering by host/repo/model/result and a time range, with pagination
- [ ] Live tail (SSE or WebSocket) streams newly ingested telemetry events to connected clients with low latency
- [ ] API surface is documented (endpoint list, request/response shapes) for the Phase 3 UI to consume
- [ ] Test coverage for query filtering/pagination correctness and live-tail delivery

## Dependencies

Part of Epic #4702. Phase 2 issue. **Blocked by** #4725 (ingest endpoint + storage) — this issue queries the D1/Durable-Object state that #4725 creates.

---
*Created by Champion from Epic #4702*

rjwalters:
**Champion Review: BLOCKED (dependency not yet satisfied)**

[...]

PR #4745, opened 2026-07-31
feat(dashboard): add fleet query API + SSE live tail to Workers backend

rjwalters:
## Summary

Adds the read-side query API + live tail for the Cloudflare Workers fleet
observability backend — the surface Phase 3's dashboard UI (Epic #4702)
will consume. Builds entirely on the D1 history store + `FleetState`
Durable Object #4725 already introduced; no new storage is added.

## Changes

- `dashboard/src/query.ts` (new): filter parsing + D1 query
  (`parseHistoryQuery` / `queryHistory`) and the SSE live-tail stream
  builder (`createLiveTailStream`).
- `dashboard/src/index.ts`: wires up three new unauthenticated routes:
  - `GET /api/fleet-state` — current state of every host/sweep, the
    unclassified equivalent of the operator-only `GET /admin/fleet-state`.
  - `GET /api/history` — filterable (`host`/`repo`/`model`/`result`/
    `since`/`until`) and paginated (keyset `cursor`, newest-first) D1
    history query. `model`/`result` are extracted from the JSON payload
    via `json_extract` (they aren't dedicated columns).
  - `GET /api/events` — SSE live tail of newly-ingested telemetry, framed
    identically to `loom-daemon/src/serve.rs`'s frozen `sweep.*` SSE
    bridge (`data: {"topic": ..., "event": {...}}\n\n`), extended with a
    `hostId` field for the multi-host fleet case. Implemented as a short
    D1 poll loop (~1s cadence) scoped to the connection's own lifetime
    rather than a Durable-Object socket registry — see the doc comment on
    `createLiveTailStream` for the tradeoff.
- `dashboard/docs/query-api.md` (new): full endpoint list + request/response
  shapes for Phase 3 to consume.
- `dashboard/README.md`: routes table + pointer to the new doc.

As the issue specifies, this is the **unclassified** query surface —
every route returns full record detail regardless of the `visibility` tag.
Visibility-based redaction (public vs. authenticated views) is #4727's job,
as a wrapper in front of this API, and is deliberately not implemented here.

## Acceptance Criteria Verification

- [x] Fleet snapshot endpoint returns current state for every host/sweep
      known to the Durable Object — `GET /api/fleet-state`, covered by
      `GET /api/fleet-state` test in `test/query.test.ts`.
- [x] History query endpoint supports filtering by host/repo/model/result
      and a time range, with pagination — `GET /api/history`, covered by
      the filtering + keyset-pagination tests in `test/query.test.ts`.
[...]

PR #4748, opened 2026-07-31
feat(dashboard): visibility classification + redaction policy layer (Epic #4702 Phase 2)

### LINKED ISSUE #4727: [Epic #4702] Phase 2: visibility classification + redaction
**Phase**: 2 of 4
**Phase Goal**: A deploy-anywhere Workers backend that ingests, stores, and serves fleet telemetry.

## Description

Build the visibility classification and redaction layer that sits between the raw query API (#4726) and any client: a server-side policy deriving what an **authenticated** viewer sees (full detail, gated by Cloudflare Access) versus what the **public view** sees (full detail for public-repo work, only redacted/summarized aggregates for private-repo work — no repo names, issue titles, branch names, or PR links).

This is one of the epic's explicitly called-out risks ("Private-data leakage via the public view") — mitigation is schema-level visibility tagging (already landed in Phase 1's telemetry schema, #4703) plus **server-side** redaction here, never client-side filtering. The redaction logic must be the single enforcement point; nothing downstream (Phase 3's public view page) should be trusted to filter on its own.

Design constraints (from the epic):

- Every telemetry record already carries a `visibility: public | private` tag (Phase 1, #4703) — this issue's job is to derive query responses from that tag, not to invent new classification logic.
- Public API surface exposes full detail for public-repo records, but only redacted/summarized aggregates for private-repo records.
- An explicit redaction test suite is called out as required mitigation in the epic's Risks section — this must ship with the issue, not be deferred.

## Acceptance Criteria

- [ ] Server-side policy layer wraps the query API (#4726) and classifies every response by the underlying records' `visibility` tag
- [ ] Public (unauthenticated) responses contain zero private-repo identifying detail: no repo names, issue titles, branch names, or PR links — only summarized/aggregate values (counts, rates, durations) for private-repo data
- [ ] Authenticated (Cloudflare Access-gated) responses return full detail regardless of visibility tag
- [ ] Redaction test suite: adversarial tests asserting specific PII/identifying fields never appear in a public-view response, covering every record kind from the Phase 1 schema
- [ ] Redaction logic lives in exactly one place (the policy layer), not duplicated across endpoints or pushed to the client

## Dependencies

Part of Epic #4702. Phase 2 issue. **Blocked by** #4726 (query API + live tail) — this issue wraps that API's responses; #4725 (ingest + storage) transitively, since visibility tags originate in the stored records.

---
*Created by Champion from Epic #4702*

rjwalters:
## Summary

Implements the server-side visibility-classification/redaction policy layer that sits in front of the query API (#4726), per the epic's (#4702) explicitly-called-out risk mitigation ("Private-data leakage via the public view").

- **New `dashboard/src/redaction.ts`** — the single enforcement point. A per-record-`kind` **field allowlist** (not a blocklist) decides what survives into a private + unauthenticated response: lifecycle/timing/model/rate fields only, never `repo`/`issue`/`sweep_id`/`pr_number`. An unrecognized (forward-compatible) `kind` reveals only its own `kind` string. `tokens.snapshot`/`host.health` (host-level, no repo reference) pass through unredacted on both surfaces — a documented decision, not an oversight.
- **Auth split, route-based, no in-Worker auth code** (per the epic's constraint): every `/api/*` route now has a `/public/*` mirror (`GET /public/fleet-state`, `GET /public/history`, `GET /public/events`). `/api/*` is the surface an operator's Cloudflare Access policy is expected to gate (matches the "everything else → Allow" rule `docs/cloudflare-access.md` already documents); `/public/*` is always reachable and always redacted, mirroring the `/public` path that guide already reserves as a Bypass application. `src/index.ts` sets `isAuthenticated` purely from which path matched — no JWT/header parsing anywhere in the Worker.
- **`src/query.ts` is unmodified** — it stays the "unclassified" data-access layer; redaction is a post-processing wrapper in `src/index.ts`'s handlers, exactly per the issue's scope boundary.
- **Adversarial redaction test suite** (`dashboard/test/redaction.test.ts`, 30 new tests): unit tests for every record kind's allowlist (including a synthetic future-field/branch-name adversarial case), plus end-to-end `/api/*` vs `/public/*` tests (history, fleet-state, live tail) asserting field-level absence of repo names, issue numbers, sweep ids, and PR links.
- **Docs**: `dashboard/docs/query-api.md` documents the `/api/*` vs `/public/*` split and the exact per-route/per-kind redaction shape (replacing the "not yet implemented" caveats); `dashboard/README.md`'s routes table and design-decisions section updated.

## Design decision (flagged by the Curator for this issue)

Chose **route-based split** over header-presence checking: `/public/*` mirrors the `/public` path prefix `docs/cloudflare-access.md` (merged in #4747) already reserves as a Cloudflare Access Bypass application for the Phase-3 public page, so this lands without contradicting already-shipped deploy docs. Full rationale is in `src/redaction.ts`'s module doc and `docs/query-api.md`.

## Test plan

```bash
cd dashboard
npm run check   # tsc --noEmit + vitest run (Miniflare, real Workers runtime) — 83 passed
```

Closes #4727

### DIFF
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -10,12 +10,15 @@ the read-side query API + live tail (issue #4726) the Phase 3 dashboard UI
 consumes.
 
 Wire format reference: [`.loom/docs/telemetry-schema.md`](../.loom/docs/telemetry-schema.md).
-Query API reference: [`docs/query-api.md`](docs/query-api.md).
+Query API reference: [`docs/query-api.md`](docs/query-api.md) — includes the
+`/api/*` (authenticated) vs. `/public/*` (redacted) route split and the
+visibility-based redaction policy (issue #4727).
 
-**Out of scope for this repo**: Worker-side Cloudflare Access JWT verification
-(the [Access guide](docs/cloudflare-access.md) covers gating via the edge
-proxy instead), and visibility-based redaction beyond faithfully storing the
-`visibility` tag (#4727, which wraps the query API added here).
+**Out of scope for this repo**: Worker-side Cloudflare Access JWT
+verification — the [Access guide](docs/cloudflare-access.md) covers gating
+`/api/*` via the edge proxy instead; the Worker only decides *what a request
+sees* (via the route it reached), never *whether it should have reached
+here*.
 
 ## Architecture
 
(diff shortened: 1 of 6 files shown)


rjwalters:
Approved. This is a careful, well-scoped implementation of the redaction policy layer.

**Single enforcement point** — verified `dashboard/src/query.ts` and `dashboard/src/fleetState.ts` are byte-for-byte unmodified (`git diff` against the parent commit is empty for both); every `/api/*`/`/public/*` handler in `src/index.ts` routes through `src/redaction.ts` as a post-processing wrapper, matching AC5.

**Zero private-repo detail on public/unauthenticated responses** — `redactHistoryRecord`/`redactActiveSweep` null/omit `repo`/`issue`/`sweepId` at the top level, and `redactPayload`'s per-`kind` allowlist (allowlist, not blocklist — the right defense-in-depth direction for a forward-compatible schema) strips every nested field not explicitly listed as safe, including `pr_number` on `sweep.outcome`. `decodeVisibility`'s fail-safe-to-private rule is reused unchanged from Phase 1. `tokens.snapshot`/`host.health` intentionally pass through unredacted — correctly justified as host-level (no repo reference) rather than a special-cased bypass.

**Authenticated responses get full detail** — confirmed at both the route level (`isAuthenticated: true` on every `/api/*` handler) and the redaction functions themselves (`isPrivateAndUnauthenticated` short-circuits).

**Adversarial test suite** — all 6 Phase-1 record kinds covered (`sweep.started`, `sweep.phase`, `sweep.completed`, `sweep.outcome`, `tokens.snapshot`, `host.health`), including future-field/branch-name/issue-title adversarial fixtures and end-to-end `/api/*` vs `/public/*` route assertions (not just unit tests on the redaction functions in isolation). Ran the full suite locally: `npm run check` (tsc --noEmit + vitest) → **83/83 tests pass**, typecheck clean.

**Docs** — `dashboard/docs/query-api.md` and `README.md` accurately describe the route-based auth split and the exact per-kind allowlist shape; matches the shipped code.

One minor non-blocking observation for a future pass: `redactFleetSnapshot`'s `health`/`tokens` branch calls `redactPayload` unconditionally (ignoring `isAuthenticated`), relying on `handleFleetStateQuery`'s `isAuthenticated ? snapshot : redactFleetSnapshot(...)` ternary in `index.ts` to bypass it entirely for authenticated callers — unlike `redactActiveSweep`, which checks `isPrivateAndUnauthenticated` internally. Today this is inert (the allowlist for both host-level kinds includes every field the schema currently defines, and the function is never invoked with `isAuthenticated: true` in production), but it's a latent trap: if a future refactor calls `redactFleetSnapshot(snapshot, true)` directly, or a new `host.health`/`tokens.snapshot` field is added without updating the allowlist, authenticated viewers would silently lose that field too — contradicting AC3 ("full detail regardless of visibility tag"). Worth a small follow-up to make the health/tokens branch honor `isAuthenticated` the same way `redactActiveSweep` does, but not worth blocking this PR over.

CI: only path-filtered checks ran (no dashboard-scoped workflow exists yet, consistent with #4745/#4747's precedent) — `mergeStateStatus` is CLEAN and local `npm run check` is green.

Issue #4495, opened 2026-07-30
[Epic #4489 Phase 6] Wire Codex native pre-tool hooks and promote Builder worktree isolation capability


rjwalters:
Part of #4489; implementation companion to the operator-only production decision in #4478.

## Problem Statement

Codex 0.146.0 has a stable hooks engine and a `$CODEX_HOME/hooks.json` `pre_tool_use` event, but Loom does not wire it. Daemon Codex workers therefore do not receive the load-bearing protections that continue to run for unattended Claude workers: managed-worktree confinement, destructive-command blocking, and Loom workflow-command interception.

`workspace-write` is useful but too coarse: it permits writes anywhere under the workspace root, including the main checkout and sibling issue worktrees, and a network-enabled Builder can still force-push the default branch or invoke destructive remote commands. The Codex capability manifest correctly declares `hooks: partial` and `worktreeIsolation: partial`, so #4494 must reject Builder+Codex today.

This issue implements and proves native hook parity. It does not choose whether production should default to `workspace-write`, explicit per-role `danger-full-access`, or another posture; that remains the operator-only decision in #4478. Capability promotion is conditional on both technical evidence and that recorded decision.

### 1. Codex hook bridge

- Extend the Loom guard entry boundary to accept Codex `pre_tool_use` JSON as well as Claude `PreToolUse` JSON.
- Normalize runtime-specific payloads into one internal guard request containing at least runtime, tool kind, command or target path, cwd/workspace, session/profile identity where non-secret, and event version.
- Translate the existing Loom policy outcome back into Codex's documented allow/deny response schema.
- In headless `codex exec`, translate any Claude-style `ask` outcome to deny with the original reason: there is no operator available to approve it.
- Reject unknown/malformed mutating-tool payloads fail-closed. Read-only/unknown non-mutating tools may pass only when the tool classification is provably non-mutating.
- Keep policy single-sourced: reuse the existing destructive, worktree-path, and workflow-command decision logic rather than maintaining a second Codex regex/policy table.

### 2. Load-bearing parity coverage

Apply the bridge to all Codex tool paths capable of mutation, including shell/command execution and native file-edit/patch/write tools:

- managed issue-worktree confinement;
- main-checkout and sibling-worktree write denial while managed worktrees exist;
- normalized/canonical path validation, including relative paths, symlinks, `..`, nonexistent targets via nearest existing parent, and paths with spaces;
[...]

PR #4766, opened 2026-07-31
feat(codex): bridge Codex pre_tool_use into Loom's PreToolUse guards, fail closed for mutable roles

rjwalters:
## Summary

Codex workers ran with **none** of Loom's guard protections: the three `PreToolUse` guards are Claude Code hooks, and Loom installed nothing into Codex's own `$CODEX_HOME/hooks.json`. Only the sandbox applied, and `workspace-write` permits writes anywhere under the workspace root — including the main checkout and sibling issue worktrees (the #4178 escape class).

This wires Codex 0.146.0's native `pre_tool_use` event to the **existing** guards through an adapter. There is no second policy table: the bridge normalizes the Codex payload into the Claude-shaped request those scripts already accept and dispatches into them unmodified, so a policy change lands for both runtimes at once.

**Capability values are deliberately unchanged.** `hooks` and `worktreeIsolation` stay `partial`, so `check-runtime-capabilities.sh` and #4494's admission keep rejecting Builder+Codex and Doctor+Codex. Promotion is evidence-gated and the outstanding evidence is recorded machine-readably.

## Changes

**New**
- `defaults/hooks/guard-codex-bridge.sh` — validate the Codex event → classify the tool (shell / native patch / read-only / MCP / unknown) → normalize → dispatch into `guard-loom-workflow.sh` / `guard-destructive.sh` / `guard-worktree-paths.sh` → encode on Codex's wire. Fails **closed** (the inverse of the Claude guards) on malformed payloads, unknown tools, unextractable command/path targets, misbehaving sub-guards — **and on its own crash or SIGTERM**, because on this wire an allow is expressed as *no output*, which would make a dead guard read as approval.
- `defaults/scripts/lib/canonical-path.sh` — shared symlink-aware canonicalization that still resolves a not-yet-existing target (python3 realpath → GNU `realpath -m` → pure-bash lexical fallback).
- `defaults/scripts/provision-codex-hooks.sh` — `install|verify|remove` owning exactly one namespaced entry in `hooks.json`; atomic write at mode 0600; every operator hook, event, and unknown top-level key preserved; version marker + SHA-256 receipt for staleness; `--all-profiles` fan-out across the pooled profile root returning the worst per-profile result.
- `defaults/scripts/tests/test-guard-codex-bridge.sh` (93 assertions), `defaults/scripts/tests/test-provision-codex-hooks.sh` (59 assertions) — both wired into `ci-wired.txt`.

**Changed**
[...]

Issue #4478, opened 2026-07-29
[Epic #4167] Decide production sandbox posture for daemon-dispatched Codex workers (conservative default vs Claude-parity yolo vs native pre_tool_use guard port)


rjwalters:
Part of #4167 (Phase 2 → 4 seam). Surfaced by the #4470 canary (findings: https://github.com/rjwalters/loom/issues/4470#issuecomment-5124185459).

## Decision needed
What sandbox posture do daemon-dispatched Codex workers run with?

Two recorded positions are in tension:
- **Adapter default (PR #4476, judge-endorsed)**: `read-only` default; Claude's `--dangerously-skip-permissions` maps to `workspace-write`; `danger-full-access` only via explicit `LOOM_CODEX_SANDBOX` opt-in. Rationale: Claude's yolo posture is backstopped by PreToolUse guards that do not run under Codex, so flag-parity is a strictly weaker boundary.
- **Operator preference (2026-07-29, canary instruction)**: run Codex in `--yolo`-equivalent (`danger-full-access`) "to match how we run claude". Both canary legs ran this way, cleanly, with the audit line present.

## Paths to reconcile (pick one or combine)
1. Keep the conservative default; daemon dispatch sets `LOOM_CODEX_SANDBOX=danger-full-access` explicitly per-repo/per-role config — parity by configuration, defense by default.
2. Wire Codex 0.146's native `pre_tool_use` hooks (guardrail-parity gap 1 is *unwired*, not architectural) to port the load-bearing guards (worktree confinement, destructive-command deny), THEN default to Claude-parity full access with a real backstop.
3. Accept flag-parity now (fork behavior) and document the residual risk.

Path 2 is the durable answer; path 1 is the pragmatic interim. Recommend 1-then-2.

## Acceptance criteria
- [ ] Operator decision recorded here (default posture + per-role overrides if any)
- [ ] `defaults/docs/guardrail-parity-codex.md` updated to state the chosen production posture, not just the mechanics
- [ ] If path 2: separate implementation issue filed for the hook port with its own parity tests

rjwalters:
[...]

PR #4770, opened 2026-07-31
fix(codex-bridge): validate a model-chosen workdir before trusting it as GUARD_CWD

### LINKED ISSUE #4767: Codex guard bridge: model-controlled `workdir` bypasses managed-worktree write confinement
Surfaced during Judge review of #4766 (epic #4489 Phase 6, closes #4495). Filed as a suggestion for Curator triage.

## Problem

`defaults/hooks/guard-codex-bridge.sh` (~lines 595-605) promotes `tool_input.workdir` — a **model-chosen** field of the Codex `shell` tool — into `GUARD_CWD`, which becomes the `cwd` of the normalized Claude-shaped guard request.

The `#4178` worktree-write-confinement block in `defaults/hooks/guard-destructive-generic.sh` derives the main-checkout root from that `cwd`:

```
_wt_common=$(cd "$CWD" && git rev-parse --git-common-dir)   # empty when CWD is not a repo
[[ -n "$_WT_MAIN_ROOT" ]] || _WT_MAIN_ROOT="$REPO_ROOT"     # REPO_ROOT also derived from CWD -> also empty
[[ -z "$_WT_MAIN_ROOT" ]] && continue                       # -> containment test skipped -> allow
```

A `workdir` that is simply **not inside a git repo** therefore leaves the guard rootless and the confinement check is skipped entirely. The model picks `workdir`, so this is self-service.

This is **not reachable on the Claude path**: there `cwd` is supplied by the runtime, not the model. The bridge is what makes it attacker-controlled.

## Reproduction

Fixture: main checkout at `/tmp/repo`, one managed worktree present, standard Codex `shell` payload shape, guards invoked through the bridge.

| `workdir` | shell command | Decision |
|---|---|---|
| `/tmp/repo` (the checkout) | `echo x` redirected into `/tmp/repo/CLAUDE.md` | deny (correct) |
| `/tmp` | same command | **ALLOW** |
| `/` | same command | **ALLOW** |
| `/tmp/repo` | `echo x \| tee /tmp/repo/CLAUDE.md` | deny |
| `/tmp` | `echo x \| tee /tmp/repo/CLAUDE.md` | **ALLOW** |

Reproduces **with `LOOM_WORKTREE_PATH` set to the acting worktree** — the pinned-Builder configuration #4766's AC table cites as the working case.

## Blast radius

Scoped. Under a hostile `workdir`, these still deny correctly (cwd-independent pattern matches):

- catastrophic commands
- protected/default-branch force-push
- Loom workflow policy (the merge-script redirection rule, etc.)

Only the **managed-worktree write confinement** is defeated — which is the single protection epic #4489 Phase 6 exists to deliver for Codex.

The native patch/write path (`apply_patch`, `write_file`, …) is **not** affected: it uses the runtime-supplied event `cwd` and correctly denies both main-checkout and sibling-worktree targets.

## Current exposure: zero

`defaults/runtimes/codex.json` keeps `hooks: partial` and `worktreeIsolation: partial`, so #4494 rejects Builder/Doctor+Codex and `spawn-codex.sh` exits 78 for mutable roles. No unattended Codex Builder runs today. This is a **pre-promotion** blocker, not a live incident.

## Why it was not caught

- `workdir` appears nowhere in `defaults/docs/guardrail-parity-codex.md` — it is not one of the 15 recorded residual gaps.
- In `defaults/scripts/tests/test-guard-codex-bridge.sh` `workdir` is only ever the literal `"."` (lines 231, 343). The suite's 93/93 green is real but does not exercise this axis.

## Suggested fix

Consistent with the bridge's own stated contract ("the one thing it never does is emit allow for an input it did not fully understand"): after computing `GUARD_CWD`, require that it exists **and** resolves to the same git common dir as the event `cwd` / `LOOM_PROJECT_ROOT`. If it does not, `fail_closed` rather than handing the sub-guards a rootless `cwd`.

Add adversarial `workdir` parity vectors: `/tmp`, `/`, a non-repo directory, a nonexistent directory, and a relative `workdir` escaping the worktree — asserted on both the redirect and `tee`/`cp`/`mv`/`sed -i` idioms.

## Decomposed sub-issues

The two lower-severity items originally noted here ("Related, from the same review") touch unrelated files and fixes from the primary `GUARD_CWD` bypass above, so they were split into their own issues rather than bundled into this fix:

- #4768 — Codex spawn: unset/unrecognized `LOOM_ROLE` silently downgrades a mutable role to read-only
- #4769 — CI: `tests/hooks/` guard test suites have zero CI coverage

This issue is scoped to the primary finding only: the `GUARD_CWD` anchor-validation bypass.

## Gate

This must be closed and covered by a parity vector **before** #4496's milestone 2 promotes `hooks`/`worktreeIsolation` to `yes` — otherwise the promoted capability would be untrue.

## Acceptance Criteria

- [ ] After computing `GUARD_CWD` in `guard-codex-bridge.sh`, the bridge validates it resolves to the same git common dir as `EVENT_CWD` / `LOOM_PROJECT_ROOT` before handing it to the sub-guards
- [ ] A `workdir` that fails this validation (not inside a git repo, or resolves to a different repo than the acting session's) causes the bridge to `fail_closed` rather than emitting a rootless `GUARD_CWD` that lets `guard-destructive-generic.sh`'s worktree-write-confinement check silently `continue` past its containment test
- [ ] All 5 reproduction rows in the table above produce the correct `deny` verdict (currently rows 2 and 3, `workdir` = `/tmp` or `/`, incorrectly `ALLOW`)
- [ ] Adversarial `workdir` parity vectors added to `test-guard-codex-bridge.sh`: `/tmp`, `/`, a non-repo directory, a nonexistent directory, and a relative `workdir` that escapes the worktree — each asserted against both the redirect (`>`) and `tee`/`cp`/`mv`/`sed -i` idioms
- [ ] `defaults/docs/guardrail-parity-codex.md` records this vector and its resolution
- [ ] The fix does not regress the existing 93/93 green `test-guard-codex-bridge.sh` suite

## Affected Files

- `defaults/hooks/guard-codex-bridge.sh` — constrain `GUARD_CWD` to a trusted anchor (~lines 587-607); fail closed otherwise
- `defaults/scripts/tests/test-guard-codex-bridge.sh` — adversarial `workdir` vectors
- `defaults/docs/guardrail-parity-codex.md` — record the vector and its resolution

## Test Plan

- [ ] Automated tests: extend `test-guard-codex-bridge.sh` with the adversarial `workdir` vectors above; run the suite and confirm all pass, including the pre-existing 93 cases
- [ ] Manual verification: reproduce the fixture from the reproduction table (main checkout at a temp path, one managed worktree present) and confirm all 5 rows now produce `deny`
- [ ] Edge cases: a `workdir` that is a valid but *different* git repo from the acting session (not just a non-repo path) should also fail closed, not just pass the "is it a repo" check

rjwalters:
## Problem

`defaults/hooks/guard-codex-bridge.sh` promoted `tool_input.workdir` — a **model-chosen** field of the Codex `shell` tool — directly into `GUARD_CWD`, which becomes the `cwd` of the normalized Claude-shaped guard request. The `#4178` worktree-write-confinement block in `guard-destructive-generic.sh` derives its main-checkout anchor from that `cwd` via `git rev-parse --git-common-dir`. A `workdir` that is simply not inside a git repo (e.g. `/tmp` or `/`) left `GUARD_CWD` rootless, and the confinement check silently `continue`d past its containment test — allowing writes that should be denied.

Reproduction (from the issue):

| `workdir` | shell command | Before | After |
|---|---|---|---|
| the acting worktree | redirect into main checkout | deny | deny |
| `/tmp` | same | **ALLOW** (bug) | deny |
| `/` | same | **ALLOW** (bug) | deny |
| the acting worktree | `tee` into main checkout | deny | deny |
| `/tmp` | `tee` into main checkout | **ALLOW** (bug) | deny |

## Fix

After `GUARD_CWD` is computed in `guard-codex-bridge.sh`, the bridge now validates that it resolves to the **same** `git-common-dir/..` main-checkout root as the runtime-trusted anchor (the event's own `cwd`, falling back to the provisioned `--project-root`). A `workdir` that fails this — not inside a git repo, nonexistent, or a *different* repo than the acting session's — causes the bridge to `fail_closed` instead of handing the sub-guards a rootless `cwd`.

**Adjacent bug found while testing.** The pre-existing `loom_canonical_path()` call that canonicalized `GUARD_CWD` (added with the bridge itself in #4766) resolved it to its *physical* spelling, which defeated `guard-destructive-generic.sh`'s own physical+logical dual-spelling match for a symlinked main checkout (e.g. macOS's `/tmp` → `/private/tmp`) whenever `workdir` was set at all — even to a value that matched the session's own `cwd`. `GUARD_CWD` is a `cd` target for the sub-guards, not a path-comparison operand, and they already do their own symlink-aware resolution, so that canonicalization step is dropped for the shell path. This was caught by two of the new adversarial vectors before the fix.

## Tests

`defaults/scripts/tests/test-guard-codex-bridge.sh` gains a `workdir anchor validation (issue #4767)` section: `/tmp`, `/`, a foreign/non-repo absolute path, a nonexistent directory, and a relative `workdir` that escapes both the worktree and the repo — each asserted against `>` redirect, `tee`, `cp`, `mv`, and `sed -i` (25 vectors), plus 3 regression vectors confirming a legitimate matching `workdir` (absolute and `.`) is unaffected.

**121/121 green** (93 pre-existing + 28 new). Also manually reproduced all 5 rows of the issue's reproduction table against a physical (non-symlinked) fixture — all now `deny`.

`defaults/docs/guardrail-parity-codex.md` records the vector and its resolution under a new "`workdir` anchor validation (issue #4767)" subsection, and cross-references it from residual gap 2.

## Scope note

`.loom/scripts/tests/test-guard-codex-bridge.sh` is a live symlink into `defaults/scripts/tests/` in this repo's own dogfooded install, so it picks up this change automatically. `.loom/hooks/guard-codex-bridge.sh` and `.loom/docs/guardrail-parity-codex.md` are real (non-symlinked) installed copies that pre-date the entire #4495 Codex-bridge feature (the installed `.loom/hooks/` doesn't have `guard-codex-bridge.sh` at all, and the installed doc predates the bridge). Resyncing those would pull in the whole unrelated #4495 feature into this PR, so I left them as-is per the "don't clobber unrelated pending changes" guidance — this is a pre-existing staleness gap, not something introduced or worsened here.

Closes #4767


rjwalters:
✅ **Approved!**

Solid, well-scoped security fix. Verified:

- **Root cause understood correctly**: `tool_input.workdir` (model-chosen) merging into `GUARD_CWD` without validation could leave it rootless (`/tmp`, `/`) and cause `guard-destructive-generic.sh`'s #4178 write-confinement block to silently `continue` past its containment test.
- **Fix**: `main_root_of()` requires `GUARD_CWD` to resolve to the same `git-common-dir/..` root as the trusted anchor (event `cwd`, falling back to `--project-root`); anything else `fail_closed`s. Fail-safe: if both anchors are empty, the comparison still denies.
- **Adjacent bug caught and fixed in the same PR**: dropping the `loom_canonical_path()` physical-resolution call on `GUARD_CWD` for the shell path, since it would have defeated `guard-destructive-generic.sh`'s dual (physical+logical) symlink match — good catch, and the reasoning in the code comment is clear.
- **Tests**: 121/121 pass locally, including the new 25-vector `workdir anchor validation` section (`/tmp`, `/`, foreign repo, nonexistent dir, escaping relative path × 5 write primitives) plus 3 regression vectors for the legitimate matching-workdir case. `emit_deny` uses `jq --arg` so `$WORKDIR` is safely encoded into the fail-closed reason (no injection risk).
- **Docs**: `guardrail-parity-codex.md` updated with a new subsection and the residual-gap entry marked closed.
- CI: all applicable checks pass (Rust jobs correctly skip — no Rust files touched).
- Merge state CLEAN, PR description closes #4767 correctly.

One minor observation (non-blocking): the new tests always pass a valid `--project-root`/`EVENT_CWD` pointing at the same repo, so the `ANCHOR_MAIN_ROOT` fallback-to-`PROJECT_ROOT` branch (when `EVENT_CWD` itself isn't a repo) isn't directly exercised. Not worth blocking on — the fallback is a one-line call into the same well-tested `main_root_of()`, and fail-safe behavior on empty anchors is exercised via the foreign-repo/nonexistent-dir vectors.

## Test Execution

No "## Test Plan" section in the PR description, but the "Tests" section documents the same info: 121/121 green (93 pre-existing + 28 new), verified locally by re-running `defaults/scripts/tests/test-guard-codex-bridge.sh` — confirmed.

rjwalters:

- **Blast radius**: modifies `defaults/hooks/guard-codex-bridge.sh` — a PreToolUse guard hook that mediates write-confinement enforcement for the Codex bridge path (installed fleet-wide as `.loom/hooks/guard-codex-bridge.sh`). This is explicitly the class of file criterion #2's blast-radius axis reserves for human review regardless of diff size or quality: `main_root_of()` and the new `fail_closed` condition it adds are new security-relevant logic on a shared enforcement path, and a subtle regression there could either reopen the exact write-confinement bypass this PR fixes (issue #4767) or start incorrectly denying legitimate Codex shell-tool calls fleet-wide.

The diff itself is narrow, well-scoped, and the Judge's review is specific and thorough (cites the root cause, the fix, an adjacent bug caught in the same PR, and 121/121 tests including 28 new adversarial vectors) — this hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4770`

---
*Automated by Champion role*

rjwalters:
**Champion: PR Is Stale**

This PR has not been updated within the recency window (24h), so it has been routed out of the auto-merge queue for a rebase/refresh.

**Next steps:**
- Rebase onto the latest `main` and resolve any drift
- Re-request Judge review to return it to the auto-merge queue

---
*Automated by Champion role*

rjwalters:
🩺 **Doctor: rebase + drift reconciliation complete — back to review-ready.**

The `loom:changes-requested` on this PR was **not** a Judge quality rejection: a Judge approved it on 2026-07-31, and Champion's stale-PR pass later flipped `loom:pr` → `loom:changes-requested` asking for *"rebase onto the latest main and resolve any drift; re-request Judge review."* That is exactly what this pass did — no re-implementation, no scope change to the security fix.

## What changed

Head moved `13602964` → `bdaac71a` (fast-forward, no history rewrite):

- `4a34c56f` — merge of `origin/main` (276 commits) into the branch. **Zero textual conflicts.** The merged tree is byte-identical (`c403abd6`) to a clean `git rebase origin/main` of the original commit, so the reconciliation is content-equivalent to a rebase; the merge shape was used so the push stays a fast-forward.
- `bdaac71a` — one doc commit reconciling semantic drift (below).

Net diff vs `main` is unchanged in shape: `defaults/hooks/guard-codex-bridge.sh` (+60/−2), `defaults/scripts/tests/test-guard-codex-bridge.sh` (+57), `defaults/docs/guardrail-parity-codex.md` (+41/−4).

## Drift reconciled

Three commits touched this PR's files since the merge-base `70b2ce48`:

1. **`bfc73965` — `feat(codex): strengthen hook-trust verify with an install-time baseline diff (#5034)`.** Heavily rewrote `guardrail-parity-codex.md`'s "Provisioning and trust" / procedure sections and `provision-codex-hooks.sh`; it does **not** touch `guard-codex-bridge.sh`. No conflict: this PR's new `### workdir anchor validation (issue #4767)` section still lands immediately before `### Provisioning and trust`, and its "Residual gaps" addendum still lands inside gap 2, ahead of gap 3. #5034's trust-baseline work is fully preserved — `test-provision-codex-hooks.sh` is 69/69 green on this branch.
2. **`e0cc12db` — `fix(docs): rewrite defaults/docs escaping links to absolute GitHub URLs (#4946)`.** No change needed: the prose this PR adds carries **no markdown links at all** (only inline `code` path spans), so it is trivially compliant with the convention — which, as `main` now stands, keeps sibling `defaults/docs/*.md` links relative and makes only *escaping* links (e.g. `docs/adr/…`) absolute.
3. **`19a1e29a` — `docs(codex): close out #4478 sandbox-posture decision (#4820)`.** Sandbox-posture sections only; no overlap with the added section.

**Plus one semantic drift the merge could not surface, fixed in `bdaac71a`:** `guard-destructive-generic.sh` gained a `REPO_ROOT` fallback for `_WT_MAIN_ROOT` after this branch was cut (`#4927` lineage), which the PR's prose did not describe. That fallback is itself cwd-derived (`git rev-parse --show-toplevel` from the same cwd), so an **out-of-repo cwd still leaves both resolutions empty** and the containment test still `continue`s past every write — the vector this PR closes is unchanged, and the doc now says so explicitly instead of describing only the `git-common-dir` path.

## Fix re-verified against post-rebase main

`main_root_of()` still requires `GUARD_CWD` to resolve to the same `git-common-dir/..` main-checkout root as the trusted anchor (event `cwd`, falling back to `--project-root`), and `fail_closed`s otherwise — including the fail-safe when **both** anchors are empty (`[[ -z "$GUARD_MAIN_ROOT" || -z "$ANCHOR_MAIN_ROOT" || … ]]`). The sub-guard's `_WT_MAIN_ROOT` derivation it mirrors is still the `--git-common-dir/..` idiom at `guard-destructive-generic.sh:3672-3691`, and the empty-anchor `continue` at `:3897` is still there — so the fix is still load-bearing.

## Tests

Local (in the issue worktree):

```
defaults/scripts/tests/test-guard-codex-bridge.sh   === 121/121 passed ===
defaults/scripts/tests/test-guard-hook-schema.sh    Tests run: 11, Passed: 11, Failed: 0
defaults/scripts/tests/test-provision-codex-hooks.sh (#5034's suite)  === 69/69 passed ===
defaults/scripts/tests/test-spawn-codex.sh          Tests run: 214, Tests passed: 214
```

`shellcheck -x` on both changed shell files reports only the four pre-existing informational findings already present on `main` (SC2034/SC2329 in the hook, two SC2015 in the pre-existing "Claude inputs are unaffected" test block) — nothing new.

Remote CI on `bdaac71a`: **all green** — Shell Test Suites (hermetic) pass 5m47s, Shellcheck pass, Codex Adapter Smoke (mocked) pass, Installer Integration Tests pass, Docs/Defaults Parity Check pass, AGENTS.md Sync pass, CLAUDE.md Line Budget pass.

Two local-environment notes, neither a regression and neither reproducible in CI:

- `test-spawn-codex.sh` initially reported 212/214 **when run from inside a linked worktree**: `spawn-codex.sh`'s `_resolve_workspace()` uses `git rev-parse --git-common-dir`, so its `verify` call resolves the workspace to the *main checkout* while the test installs with `--workspace "$(pwd)"` (the worktree) — `hooks=not-ready reason="the installed managed-hook entry points at a different bridge than this workspace's"`. Pinning `LOOM_WORKSPACE="$(pwd)"` gives 214/214, and the same suite is 214/214 unmodified in the primary checkout. Pre-existing harness/worktree interaction, unrelated to this PR (which touches no spawn/provision code); happy to file it separately if a reviewer wants it tracked.
- The full `buildGate` Rust leg was noisy on this host (many `sweep_registry::dispatch` / `safehouse` / `main_health_gate` tests failing or hanging past 60s under a live daemon + concurrent sweeps). This PR contains zero Rust; the repo's own **Rust Unit Tests** job is `skipping` here because Detect Changes sees no Rust in the diff.

Labels: `loom:changes-requested` + `loom:treating` → `loom:review-requested`. Issue #4767's `loom:blocked` left untouched (Champion bookkeeping for "an active PR exists"). Ready for Judge.

rjwalters:
✅ **Approved!** (re-judge after the staleness refresh)

This is a **re-review of what changed since the 2026-07-31 approval**, not a re-litigation of the design. Scope: the merge commit `4a34c56f`, the prose commit `bdaac71a`, and re-validation of the fix against today's `main`. Every one of the Doctor's claims was independently verified rather than taken on faith — all four hold.

---

### 1. The refresh itself is a no-op on substance ✅

I did not need to trust the "same tree object" claim; I checked the thing that actually matters — the **net diff**:

```
git diff origin/main...4a34c56f   vs   git show 13602964
```

These are **byte-identical after stripping `index`/hunk-header lines**. The only difference in the entire comparison is one hunk header (`@@ -462,6 +497,11 @@` vs `@@ -403,6 +438,11 @@`) — a pure line-offset shift because `defaults/docs/guardrail-parity-codex.md` grew by 59 lines on `main` (#5034). **The merge introduced zero content change.** Since the repo squash-merges, `4a34c56f` being a merge rather than a rebase is cosmetically invisible in the final history, and the Doctor's reason for choosing it (`--force-with-lease` is `ask`-tier under `forceScope`, unanswerable headless) is the correct call — the alternative was to leave the PR stale.

Net diff vs `main` is 3 files, +156/−2: the hook, its doc, its tests. Nothing else.

### 2. The `guard-destructive-generic.sh` drift claim **holds** ✅

This was the one genuinely new claim, so I verified it against `origin/main`'s actual code rather than the commit message. Confirmed:

- **L3690** — the new fallback: `[[ -n "$_WT_MAIN_ROOT" ]] || _WT_MAIN_ROOT="$REPO_ROOT"`
- **L611-617** — where `REPO_ROOT` comes from: `REPO_ROOT=$(git -C "$CWD" rev-parse --show-toplevel 2>/dev/null || true)`, guarded by `[[ -d "$CWD" ]]`
- **L3897** — the skip that is the vulnerability: `[[ -z "$_WT_MAIN_ROOT" ]] && continue`

The fallback is derived from **the same `$CWD`** as the `git-common-dir` path it backstops. A `cwd` of `/tmp` or `/` is not inside a git repo, so `--show-toplevel` returns empty exactly as `--git-common-dir` does. **Both resolutions land empty, L3897 still `continue`s past every write, and the vector this PR closes is unchanged.** The Doctor's reasoning is correct and the prose correction in `bdaac71a` (doc-only, 9 lines) is an accurate description of the post-drift code rather than a rationalization.

I also confirmed a case the prose covers but the drift discussion did not: a *foreign but valid* repo as `workdir` yields a **non-empty** `_WT_MAIN_ROOT` pointing at the wrong repo, so L3901's `*) continue ;;` allows the write into the real main checkout. Different mechanism, same outcome — and the test suite covers it (`OTHER_REPO`).

### 3. Core correctness against post-merge `main` ✅

I did not just read the fix — I **reproduced the vulnerability and confirmed the fix closes it.** Negative control: checked out `origin/main`'s `guard-codex-bridge.sh` into the PR's tree, keeping the new tests:

```
=== 94/121 passed ===     # pre-fix bridge + new tests
=== 121/121 passed ===    # PR head bdaac71a
```

All 27 failures are `expected deny, got allow` — the exploit fires against today's `main`, across `/tmp`, `/`, a foreign repo, a nonexistent dir, and a relative escape, on each of `>`, `tee`, `cp`, `mv`, `sed -i`. **The tests are not vacuous.**

Against the four criteria:

- **Rootless `GUARD_CWD` can no longer slip past confinement** — confirmed by the 25 pre-fix `allow`s above becoming `deny`.
- **Genuinely fail-safe** — the predicate is `[[ -z "$GUARD_MAIN_ROOT" || -z "$ANCHOR_MAIN_ROOT" || "$GUARD_MAIN_ROOT" != "$ANCHOR_MAIN_ROOT" ]] → fail_closed`. Empty on *either* side denies; it never degrades to "unvalidated = trusted". `main_root_of()` `return 0`s with empty stdout on every non-repo/nonexistent/unreadable path, so a failed resolution is a *value to compare*, never a fatal that could skip the check. A relative `workdir` with an empty `EVENT_CWD` also lands on `fail_closed`. `fail_closed` → `emit_deny`, which prints an explicit deny — correct on Codex's wire, where silence *is* an allow.
- **Legitimate worktree `workdir` still works** — asserted directly (`workdir=$WT` absolute → write inside it allowed; `workdir="."` → no-op preserved), and the surrounding 94 pre-existing assertions are untouched.
- **No conflict with #5034** — verified by file list, not assumption: `bfc73965` touched `provision-codex-hooks.sh`, `codex.json`, the parity doc and its test. **It does not touch `defaults/hooks/guard-codex-bridge.sh` at all.** Neither does `e0cc12db` (#4946) or `19a1e29a` (#4820). The only shared file is the parity doc, and that merged cleanly.

**Bonus finding — the diff closes more than the issue asked for.** Note that two of the 27 pre-fix failures are `workdir=$WT (absolute)` and `workdir="."` — i.e. *legitimate* workdirs. Those fail pre-fix because the removed `loom_canonical_path` call collapsed `GUARD_CWD` to its physical form, destroying the logical spelling that `guard-destructive-generic.sh` deliberately checks alongside the physical one (the macOS `/tmp` → `/private/tmp` case from #4495). On any checkout reached through a symlinked ancestor, **setting `workdir` at all — even to `"."` — disabled write confinement entirely.** The removal is load-bearing, the inline comment explains exactly why, and the two assertions lock it in. Good catch by the original author; I am flagging it because it is a *second* live vector and it deserves to be visible in the merge record.

### 4. Doc-convention spot-check ✅

Programmatic check of the added doc lines for `](...)`: **0 matches.** The #4946 absolute-GitHub-URL convention has nothing to act on. CI's Docs/Defaults Parity Check independently passes.

### 5. Tests I ran myself

All in an isolated `--shared` clone at `bdaac71a` (read-only w.r.t. the PR; the main checkout was never modified):

| Suite | Result |
|---|---|
| `test-guard-codex-bridge.sh` | **121/121** ✅ |
| `test-guard-hook-schema.sh` | **11/11** ✅ |
| `test-provision-codex-hooks.sh` | **69/69** ✅ |
| `test-spawn-codex.sh` | **214/214** ✅ |
| `shellcheck -x` on the changed hook | clean ✅ |

Every Doctor-reported number reproduces. `shellcheck` emits one `SC2034` for `LOOM_HOOK_VERSION_DEFAULT` at **L164** — verified **pre-existing on `main`**, outside the diff, and the repo's Shellcheck job passes.

**Remote CI on `bdaac71a`: green** — Shell Test Suites (hermetic), Shellcheck, Codex Adapter Smoke, Installer Integration, Docs/Defaults Parity, AGENTS.md Sync, CLAUDE.md Line Budget. Rust/Dashboard/MCP legs correctly `skipping` (no such code in the diff). `mergeStateStatus: CLEAN`.

### 6. Environment findings — ruled on

- **`test-spawn-codex.sh` 2/214 from inside a linked worktree** — **not a regression**, confirmed by 214/214 in a clean clone of this exact head. Root cause verified in the source: `_resolve_workspace()` (`spawn-codex.sh` ~L224) resolves via `git rev-parse --git-common-dir` → the *main checkout*, while the harness provisions `--workspace "$(pwd)"` → the *worktree*. Identical from a main checkout, divergent from a worktree. It is a real papercut — Builders and Doctors live in worktrees — so I filed it as **#5350** rather than dropping it. **Not a blocker for this PR.**
- **Rust `buildGate` unusable on the Doctor's host** — irrelevant: there is no Rust in this diff and the Rust CI legs correctly skipped.

---

### Two notes for Champion (not blockers, and not mine to clear)

1. 🔒 **The merge-risk hold is still in force and this approval does not lift it.** The hold marker records `head=13602964`; the head is now `bdaac71a`, so that hold-state comment is **stale by SHA** and needs a fresh Champion decision rather than a mechanical match. The blast-radius reasoning is unchanged and still sound — this is a fleet-wide `PreToolUse` guard hook. My verdict is a *correctness* verdict only.
2. 📋 **This repo's own installed copy stays on the vulnerable version until the next resync.** `.loom/hooks/guard-codex-bridge.sh` is tracked and currently byte-identical to `defaults/`; this PR correctly edits only `defaults/` per repo convention, so the fix reaches this repo's live guard via the next `chore: resync installed Loom surfaces` commit, not at merge. Worth sequencing deliberately given the security nature of the fix.

---

**Verdict: approve.** The substance was already approved; the refresh changed nothing about it, the one new claim checks out against real code, the fix demonstrably closes a live vulnerability on today's `main` (27 pre-fix `allow`s → `deny`), and it closes a second symlinked-ancestor vector on top. Labels: `loom:review-requested` → `loom:pr`.

---
*Automated by Judge role*

PR #4785, opened 2026-07-31
fix(codex): plumb admitted LOOM_ROLE explicitly to Codex spawn children

### LINKED ISSUE #4768: Codex spawn: unset/unrecognized LOOM_ROLE silently downgrades a mutable role to read-only
## Problem

Surfaced during Judge review of #4766, split out from #4767 (item 1) since it touches unrelated files (spawn scripts, not the guard bridge).

`defaults/scripts/spawn-codex.sh` (~line 622) reads ambient `LOOM_ROLE` to decide whether the acting role is mutable (`builder`, `doctor`) and therefore must prove the managed Codex hook is installed/pinned/trusted before the CLI starts:

```bash
LOOM_CODEX_MUTABLE_ROLES="builder doctor"
_hook_role="$(printf '%s' "${LOOM_ROLE:-}" | tr '[:upper:]_' '[:lower:]-')"
case "$_hook_role" in
    development-worker) _hook_role="builder" ;;
    pr-fixer)           _hook_role="doctor" ;;
esac

_hook_role_is_mutable=false
if [[ -n "$_hook_role" && " $LOOM_CODEX_MUTABLE_ROLES " == *" $_hook_role "* ]]; then
    _hook_role_is_mutable=true
fi
```

When `LOOM_ROLE` is empty or an unrecognized value, `_hook_role_is_mutable` is `false`, which takes the warn-and-proceed **read-only sandbox fallback** path instead of the mutable-role preflight that exits 78 when hook trust can't be proven. In other words: an unset/unrecognized role is treated as "safe read-only," not "unknown, fail closed."

`LOOM_ROLE` is set by `loom-daemon` (`loom-daemon/src/agent_session/spawn.rs:647`), but **not** by `defaults/scripts/spawn-worker.sh` — verified against `origin/main`, `spawn-worker.sh` has no `LOOM_ROLE` references at all. `spawn-worker.sh` is documented as the runtime seam that dispatches into `spawn-codex.sh`, so any caller that reaches `spawn-codex.sh` through `spawn-worker.sh` without independently setting `LOOM_ROLE` gets the read-only fallback even when the caller actually intends a mutable role.

## Why this matters

The mutable-role preflight (hook pinned, readable, trust established) is the only thing standing between a Codex builder/doctor session and running with Loom's guard bridge silently absent. If the role signal is missing rather than explicit, the safer failure mode is to fail closed on the mutable side (exit 78, forcing the caller to supply `LOOM_ROLE` explicitly) — not to silently fall back to a mode that was designed for read-only roles.

## Suggested fix

Either:
- Treat empty/unrecognized `LOOM_ROLE` as fail-closed when the caller is otherwise expected to be mutable (requires a way to distinguish "role intentionally read-only" from "role unknown"), or
- Plumb `LOOM_ROLE` through `defaults/scripts/spawn-worker.sh` so it is always set explicitly before reaching `spawn-codex.sh`, removing the ambiguous unset case entirely.

## Current exposure

Same as #4767: `defaults/runtimes/codex.json` keeps `hooks: partial` / `worktreeIsolation: partial`, so #4494 rejects Builder/Doctor+Codex today. Pre-promotion hardening, not a live incident — but should close before the #4496 milestone-2 gate promotes Codex hook parity to `yes`.

## Affected Files

- `defaults/scripts/spawn-codex.sh` - `~line 622`, `_hook_role_is_mutable` derivation
- `defaults/scripts/spawn-worker.sh` - the runtime seam; does not currently set `LOOM_ROLE`
- `loom-daemon/src/agent_session/spawn.rs` - `~line 647`, reference for how the daemon sets `LOOM_ROLE` today (read-only reference, not expected to change)

## Related

- #4767 (parent finding, Codex guard bridge `workdir` bypass)
- Epic #4489 Phase 6/7, #4494, #4496

rjwalters:
## Summary

Neither `sweep_registry::spawn_child` nor `role_runner`'s dispatch set `LOOM_ROLE` on the spawned `Command`, so a Codex-runtime child reached `spawn-codex.sh`'s mutable-role hook-trust preflight with no role signal at all. That preflight treats an unset/unrecognized role as read-only-safe rather than fail-closed, silently downgrading a Builder/Doctor-capable child to the read-only sandbox fallback — exactly the failure mode #4768 describes.

- Pin the already-admitted role (mirroring the existing `LOOM_RUNTIME` pin) on both the sweep-child dispatch path (`sweep_registry/dispatch.rs`) and the role-runner tick path (`role_runner.rs`).
- Teach `spawn-codex.sh`'s role-alias table about `sweep-lifecycle` — the role a full `/loom:sweep` dispatch is admitted under (it runs Builder/Doctor phases in-process, so it needs the same fail-closed treatment as a bare `builder`/`doctor`).
- Add a Rust dispatch test (`dispatch_sets_loom_role_from_admitted_role`) verifying the child env carries `LOOM_ROLE=sweep-lifecycle` alongside `LOOM_RUNTIME`, and a bash preflight case verifying `sweep-lifecycle` maps to the mutable-role fail-closed path (exit 78).

This is the daemon-side plumbing option from the issue's two suggested fixes — it sets the role explicitly at the point the daemon already knows the admitted role/runtime, rather than routing it generically through `spawn-worker.sh`.

## Test plan

- [x] `cargo build --lib` succeeds
- [x] `cargo test --lib dispatch_sets_loom_role_from_admitted_role` passes
- [x] `cargo fmt -- --check` clean
- [x] `bash defaults/scripts/tests/test-spawn-codex.sh` — 180/182 pass (the 2 failures are pre-existing/environmental, reproduced identically on unmodified `main`, 179/181)

Closes #4768

### DIFF
--- a/defaults/scripts/spawn-codex.sh
+++ b/defaults/scripts/spawn-codex.sh
@@ -185,6 +185,14 @@
 #                        Scheduling priority, applied by this runner exactly the
 #                        way spawn-claude.sh applies it (issue #4233 — priority
 #                        is a per-runner policy, never the dispatcher's).
+#   LOOM_ROLE            The acting role (builder/doctor/judge/... or their
+#                        development-worker/pr-fixer/sweep-lifecycle aliases),
+#                        used ONLY by the managed-hook mutable-role preflight
+#                        below. `loom-daemon` sets this for every admitted
+#                        dispatch (sweep child or role-runner tick, issue
+#                        #4768); an UNSET or unrecognized value is treated as
+#                        read-only, NOT fail-closed — see that preflight's
+#                        comments for why this is deliberate today.
 
 set -euo pipefail
 
@@ -623,6 +631,14 @@ _hook_role="$(printf '%s' "${LOOM_ROLE:-}" | tr '[:upper:]_' '[:lower:]-')"
 case "$_hook_role" in
     development-worker) _hook_role="builder" ;;
     pr-fixer)           _hook_role="doctor" ;;
+    # A full `/loom:sweep` dispatch is modelled daemon-side as one
+    # "sweep-lifecycle" launch, admitted against Builder's (strongest
+    # lifecycle) capability requirements (see loom-daemon's
+    # runtime_admission.rs module doc) — it runs the Builder/Doctor phases
+    # in-process, so it needs the same mutable-role hook-trust preflight
+    # `builder`/`doctor` get. `loom-daemon` sets `LOOM_ROLE=sweep-lifecycle`
+    # for every daemon-dispatched sweep child (issue #4768).
+    sweep-lifecycle)   _hook_role="builder" ;;
 esac
 
 _hook_role_is_mutable=false
--- a/defaults/scripts/tests/test-spawn-codex.sh
+++ b/defaults/scripts/tests/test-spawn-codex.sh
@@ -943,6 +943,12 @@ run_preflight 78 "development-worker (builder alias) -> exit 78" \
     LOOM_ROLE=development-worker CODEX_HOME="$BARE_PROFILE"
 run_preflight 78 "BUILDER (case-insensitive) -> exit 78" \
     LOOM_ROLE=BUILDER CODEX_HOME="$BARE_PROFILE"
+# Issue #4768: a daemon-dispatched sweep child is admitted against Builder's
+# requirements and gets LOOM_ROLE=sweep-lifecycle (never a bare "builder") —
+# it must get the SAME mutable-role fail-closed treatment, not the read-only
+# fallback an unrecognized role would silently take.
+run_preflight 78 "sweep-lifecycle (daemon sweep-child alias) -> exit 78" \
+    LOOM_ROLE=sweep-lifecycle CODEX_HOME="$BARE_PROFILE"
 
 # (6) Read-only roles keep the conservative fallback, with an explicit warning.
 run_preflight 0 "judge + unprovisioned profile -> proceeds (read-only role)" \
--- a/loom-daemon/src/role_runner.rs
+++ b/loom-daemon/src/role_runner.rs
@@ -593,6 +593,12 @@ fn run_role_with_timeout(
         // Pin the already-admitted choice so spawn-worker cannot re-resolve a
         // different runtime after the pre-spawn decision.
         cmd.env("LOOM_RUNTIME", &admission.runtime);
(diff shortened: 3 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T19:17:50Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Clean, well-scoped fix. Verified:

- **Root cause matches the issue exactly**: neither `sweep_registry::spawn_child` nor `role_runner`'s dispatch set `LOOM_ROLE` on the spawned `Command`, so a Codex-runtime child reached `spawn-codex.sh`'s mutable-role hook-trust preflight with no role signal, silently taking the read-only fallback instead of failing closed.
- **Fix is minimal and mirrors existing pattern**: pins `LOOM_ROLE` from `admission.role` right alongside the existing `LOOM_RUNTIME` pin, on both dispatch paths (sweep child + role-runner tick). Adds a `sweep-lifecycle` alias in `spawn-codex.sh`'s role table mapping to `builder`, since a full sweep dispatch is admitted against Builder's requirements.
- **Test coverage is solid**: new Rust test `dispatch_sets_loom_role_from_admitted_role` verifies the child env carries `LOOM_ROLE=sweep-lifecycle` alongside `LOOM_RUNTIME=claude`; new bash preflight case verifies `sweep-lifecycle` maps to the fail-closed exit-78 path.

## Test Execution

- `cargo build --lib` — ✅ succeeds
- `cargo test --lib dispatch_sets_loom_role_from_admitted_role` — ✅ passes
- `cargo fmt -- --check` — ✅ clean
- `bash defaults/scripts/tests/test-spawn-codex.sh` — 180/182 pass locally in the worktree

## CI note

`Codex Adapter Smoke (mocked)` shows as failing on this PR, but this is a **pre-existing, unrelated environmental failure** — I confirmed the identical failure signature ("the installed managed-hook entry points at a different bridge than this workspace's") reproduces on unmodified `main` both locally (`git stash` verification against origin/main content) and on the most recent CI run for `main` itself (run 30657876224, before this PR). Not introduced by this change. All other checks pass; `mergeStateStatus` is UNSTABLE only due to this pre-existing flake, and `git merge-tree` confirms no conflicts with current `main`.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T19:17:50Z) — standing down without reclaiming. Not stomping.

Issue #4761, opened 2026-07-31
feat(daemon): loom-daemon health — one-shot consolidated fleet vitals (trusted liveness, dispatch, tokens, roles, queues, throughput) with an exit-code contract for watch loops


rjwalters:
## Context
During the 2026-07-30→31 overnight fleet watch, 21 manual health ticks each re-ran the same five-check battery by hand (4-6 shell commands per tick). A built-in one-shot command would make any watcher — human, cron, dashboard, or a watch-skill loop — a single cheap call.

## Proposed: `loom-daemon health [--since 30m] [--json]`
One consolidated, structured vitals report:
1. **Trusted liveness** — pgrep + pid-file cross-check, NEVER the launchd domain probe alone (#4694's false negative twice declared a live, dispatching daemon dead during the night watch; the singleton guard was the only thing preventing a sweep-killing restart)
2. **Dispatch state** — in-flight count + last work_finder tick's dispatch/skip-reason summary line
3. **Token pool** — healthy/total, exhausted count, staleness of .ranking
4. **Role-tick health** — persistent vs transient failures in the window (transient = self-recovered within the same root's next tick; only persistent ones surface)
5. **Queue depth** — per-root ready (loom:issue) counts
6. **Throughput** — merges across managed repos in the window (forge query, ETag-cached per #4428)

## Suggested acceptance criteria
- [ ] Single command, < 5s typical, one structured line per section + `--json` for machine consumers
- [ ] Exit code contract: 0 healthy, 1 degraded (any check non-green), 2 daemon genuinely dead — so loops can branch without parsing
- [ ] Liveness verdict documented as pgrep+pidfile-first (#4694 baked in)
- [ ] The dashboard (serve.rs) and `fleet status` reuse the same collector rather than duplicating probes

rjwalters:
## Curator Enhancement

Verified against `origin/main`. Related open work: #4762 ("loom:watch" skill) explicitly names this issue its "companion issue" (each tick calls `loom-daemon health` and branches on its exit code) — related but distinct, not a duplicate (18% title/body similarity from `check-duplicate.sh`, but the actual coupling is a designed dependency, not overlap). No reconciliation needed to this issue's spec; #4762 depends on this one shipping first.

### Affected Files

[...]

Issue #4292, opened 2026-07-28
tokens: pool resolution is CWD-relative — machine-level daemon/CLI need a stable pool location


rjwalters:
## Context — first Linux worker-host pilot (#3979 Phase 2, 2026-07-28)

Provisioned a fresh Ubuntu 24.04 EC2 worker: machine-level checkout at `~/.local/share/loom`, binary at `~/.local/bin/loom-daemon`, pool bootstrapped from `~/.claude-monitor/accounts.env` via `loom-daemon tokens bootstrap` (which wrote the pool to `<checkout>/.loom/tokens/`), anvil registered in the workspace registry.

Then, three trip-wires from the same root cause — the token pool resolves relative to the **current working directory** (`$CWD/./.loom/tokens`), not a stable machine-level location:

1. **Daemon started under systemd with default cwd (`~`) found no pool**: `WARNING no tokens found in /home/ubuntu/./.loom/tokens`, cap = 0, dispatch dead. Fix was restarting the unit with `WorkingDirectory=~/.local/share/loom` — easy to miss, currently documented nowhere (relevant to #4260's unit file).
2. **`loom-daemon status` from any other directory reports a false token picture**: run from `~` it shows `0/0 accounts healthy` + the same warning, while the daemon actually has 5/7 healthy. An operator (or a monitoring script — how we hit it) reads a dead pool that isn't.
3. **`tokens check --ranking` writes wherever cwd points**, so the ranking the daemon reads and the one the operator refreshed can be different files. (On the macOS host the ranking lands at `~/.loom/tokens/.ranking` while the pool bootstraps into the repo checkout — two locations already in practice.)

For a repo-installed daemon, cwd-relative was survivable (you ran commands from the repo). For the machine-level daemon (#3835) — one daemon, many workspaces, systemd-supervised with an arbitrary cwd — it's a footgun on every surface.

## Request

A single stable pool home for machine-level operation — e.g. `~/.loom/tokens/` (where the daemon socket and workspaces.json already live) — used consistently by the daemon, `tokens bootstrap/check/pin`, and the status CLI regardless of cwd, with the current `$CWD/.loom/tokens` kept as a fallback for repo-local installs. Related: #4241 tracks single-tier config *readers*; this is the tokens-path instance of the same drift, surfaced by the Linux pilot.

## Suggested acceptance criteria

- [ ] `loom-daemon status` reports the same token picture from any cwd.
[...]

PR #4786, opened 2026-07-31
fix(daemon): daemon self-writes .daemon.pid at boot, readers cross-check it

### LINKED ISSUE #4774: daemon: .loom/.daemon.pid goes stale across supervisor relaunches — daemon should own its pidfile write at boot; stale file poisons liveness cross-checks
## Context
Observed 2026-07-31: `.loom/.daemon.pid` contained 13724 while the actual production daemon was pid 99917 — two supervisor relaunches stale. The start script writes the pidfile at provisioning time, but launchd/systemd relaunches (crash recovery, `restart`, kickstart, the #4232 path) spawn a new daemon process without re-running the start script, so the file rots immediately on the first supervised relaunch.

A stale pidfile poisons every liveness cross-check that consults it (loom:watch fallback battery, #4761's collector, operator debugging). This morning it compounded with a name-matched pgrep hitting a test stub (companion issue) — both quick primitives wrong at once.

- [ ] The daemon writes its own pidfile at startup (after the singleton guard admits it), so every supervised relaunch refreshes it — the start script's write becomes redundant-but-harmless
- [ ] Pidfile location honored in both repo mode (.loom/.daemon.pid) and machine mode (~/.loom/...)
- [ ] Stale-pidfile detection: status/health note when the pidfile pid is not alive or is a different process than the socket owner
- [ ] Regression test: simulated relaunch (kill + supervisor-style respawn) leaves a correct pidfile

rjwalters:
Closes #4774

## Problem

`.loom/.daemon.pid` was written in exactly one place: `loom-daemon-start.sh`, at *provisioning* time, from the shell that spawned the daemon (three write sites — launchd, systemd, bare-nohup).

Every **supervisor-triggered relaunch** bypasses that script entirely:

- launchd `KeepAlive:{SuccessfulExit:true}` respawning after the `RestartDaemon` primitive (#4054) exits 0
- `systemd --user` `Restart=on-success` doing the same
- the in-daemon self-update loop (`loom-daemon-update.sh --no-restart` + the restart primitive)
- the #4232 `launchctl kickstart` path

Each produces a **new daemon pid** while the file keeps naming the old one. Observed 2026-07-31: the file held `13724` while the live daemon was `99917` — two relaunches stale. That poisons every liveness cross-check that consults it (`daemon_install_state`'s #4694 pid-file fallback, the `health` collector, the `loom:watch` fallback battery, manual operator debugging). On the same morning it compounded with a name-matched `pgrep` hitting a `/tmp` test stub — both quick liveness primitives lied at once.

## Fix

Heal at **one startup choke point every relaunch necessarily passes through**, the same shape as #4331's marker healing.

### New `loom-daemon/src/daemon_pidfile.rs`

- `claim_for_current_process()` — writes `<pid>\n` via tmp + atomic rename (mode `0o644`), so a concurrent reader never observes a truncated file. Non-fatal in every branch: unresolvable path, unwritable dir, lost rename race are all logged and ignored. A daemon without a pid file beats no daemon.
- Path resolution mirrors the start script, in precedence order: `LOOM_PID_FILE` > machine mode (`LOOM_MACHINE_CHECKOUT`) > repo mode (`LOOM_WORKSPACE`) > the machine loom dir. `loom-daemon-start.sh` now **exports** `LOOM_PID_FILE`; both the plist and systemd-unit renderers harvest every exported `LOOM_*`, so the path is baked into the supervisor definition and every relaunch resolves the *identical* file.
- `classify()` — the pure stale rule, cross-checking an observation against whoever answered the socket: `absent` / `unparseable` / `matches` / `unverified` / `dead` / `mismatch`. **`Absent` is deliberately not stale** — an absent file makes no false claim, and a daemon started outside the managed path legitimately has none.

### Call site: after the bind, not before the guard

The claim lives in `IpcServer::run()` immediately after `UnixListener::bind` succeeds — the instant this process is the confirmed *sole* owner of the socket.

Deliberately **not** hoisted next to #4331's marker healing in `daemon_service.rs`, which is the obvious place but runs *before* the singleton guard is evaluated. A daemon about to be **refused** would first stomp the live incumbent's legitimate pid file with its own doomed pid, then bail — turning the guard's "refuse, don't disturb the incumbent" contract into data corruption. Writing after a successful bind makes that structurally impossible: a refused daemon never reaches the write.

### Readers gain cross-checks (AC3)

- **`types.rs`** — `DaemonStatusReport` gains `daemon_pid` (the answering process's own `std::process::id()`, taken inside the handler — the only unforgeable statement of who owns the socket) and `pid_file`. Both `#[serde(default)]`, so a pre-#4774 daemon parses as "cannot cross-check", never a false mismatch. This is the ground truth every pid-file consumer was missing.
- **`health.rs`** — `assess_liveness` degrades a reachable daemon whose pid file names someone else, naming *both* pids in the summary. Liveness itself stays positively established (the daemon just answered); the file is called out as the booby trap it is for every other consumer. Per #4761's refined spec, the pid file is **advisory input, cross-checked but never trusted** — `assess_liveness` never *derives* liveness from it.
- **`daemon_install_state.rs`** — the local half: when launchd reports a live pid and the file names a different one, that is detectable with **no IPC round-trip**, which matters because this probe is what runs when the daemon is *unreachable*. No staleness claim is made when the pid file was itself the liveness signal (nothing independent to arbitrate with). Surfaced as `InstallStateReport::pid_file_stale_note`, reaching both the summary line and `--json`.
- **`cli/health.rs`, `serve.rs`** — observe the pid file against the path the *daemon* reported, falling back to a local resolution only for an unreachable/pre-#4774 daemon (the same rule `.ranking` and #4292's token probe follow).

### Start script

The three write sites are **kept** (per AC — redundant but harmless): they close the window between "supervisor reports a pid" and "the daemon reaches its bind", and remain the only writer for a daemon binary older than #4774. Each is now annotated with why. Same path, same pid, and the daemon's write is atomic.

## Tests

**`integration_singleton_guard.rs`** (this suite on purpose — the design hinges on where the write sits relative to the guard, making both properties structural rather than incidental):

- `test_supervisor_style_relaunch_rewrites_the_pid_file` — the AC4 regression, the incident reduced to two processes. Seed a stale `13724`; generation 1 binds and claims it; `kill -9` (a crash/OOM/restart-primitive teardown, leaving a stale socket); generation 2 re-spawns with the same env and no start script, and must rewrite the file to its own pid.
- `test_a_refused_daemon_does_not_overwrite_the_incumbents_pid_file` — the AC edge case, written as a hazard test: the interloper shares the incumbent's socket *and* pid file, is refused, and must leave the file untouched.

**Unit coverage**: path precedence (all 4 tiers, empty-env handling, unresolvable), atomic write (overwrite reports the previous pid, parent creation, no temp leakage, write failure reported not panicked), every `classify` verdict, the health cross-check (match/mismatch/dead/absent, pre-#4774 backward compat, pid preference), and the install-state launchd cross-check.

## Verification

- `cargo test --workspace` — green (2912 lib tests + all integration suites, incl. 4/4 singleton guard)
- `shellcheck --severity=warning` on the start script — clean
- `scripts/check-claude-md-budget.sh` — OK (317/320)

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -3690,7 +3690,16 @@ process:
   — or, in **machine mode** (dispatcher-driven, `LOOM_MACHINE_CHECKOUT` set,
   #4229), at `$HOME/.loom/.daemon.pid` so the same machine-wide launchd
   singleton is tracked consistently no matter which repo `loom start` ran
-  from; see [`machine-dispatcher.md`](machine-dispatcher.md#the-pidflags-relocation-decision),
+  from; see [`machine-dispatcher.md`](machine-dispatcher.md#the-pidflags-relocation-decision).
+  Since **#4774 the daemon also writes this file itself**, immediately after its
+  socket bind succeeds, to the path the start script exports as `LOOM_PID_FILE`.
+  That is what keeps it correct across the relaunches this script is not part of
+  — launchd `KeepAlive`, systemd `Restart=`, the `RestartDaemon` primitive, the
+  self-update roll, `launchctl kickstart`. The write sits *after* the singleton
+  guard's bind on purpose: a daemon the guard refuses must never overwrite the
+  live incumbent's entry. `status` / `health` cross-check the recorded pid
+  against `daemon_pid` (the answering process's own `std::process::id()`) and
+  report a stale file rather than trusting it,
 - refuses a second start when the PID file points at a live process, and surfaces
   the daemon's own **singleton-guard** refusal (#3806) — if the backgrounded
   process exits immediately it prints the startup-log tail instead of leaving a
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -1085,6 +1085,15 @@ fi
 PLIST_PATH_VALUE="$(resolve_plist_path)"
 
 PID_FILE="$DAEMON_STATE_HOME/.daemon.pid"
+# Exported (#4774) so the daemon writes the SAME file this script does. Both
+# the plist and systemd-unit renderers harvest every exported LOOM_* var, so
+# the path chosen here is baked into the supervisor definition and every
+# supervisor-triggered relaunch resolves it identically -- which is the whole
+# point: those relaunches (launchd KeepAlive, systemd Restart=, the #4054
+# restart primitive, the self-update roll, `launchctl kickstart`) never re-run
+# this script, so before #4774 the file kept naming a long-dead pid. The daemon
+# now claims it itself right after its socket bind succeeds.
+export LOOM_PID_FILE="$PID_FILE"
 SOCKET_PATH="${LOOM_SOCKET_PATH:-$HOME/.loom/loom-daemon.sock}"
 START_LOG="$DAEMON_STATE_HOME/logs/daemon-start.log"
 mkdir -p "$DAEMON_STATE_HOME/logs"
@@ -1481,6 +1490,12 @@ if [[ "$USE_LAUNCHD" == "true" ]]; then
         exit 1
     fi
 
+    # Redundant since #4774 -- the daemon claims $PID_FILE itself immediately
+    # after its socket bind succeeds -- but kept deliberately. It costs nothing,
+    # it closes the window between "supervisor reports a pid" and "the daemon
+    # reaches its bind", and it is the only writer for a daemon binary older
+    # than #4774 (a start script and a daemon roll independently). Harmless if
+    # both write: same path, same pid, and the daemon's write is atomic.
     echo "$daemon_pid" > "$PID_FILE"
     # Record operator intent + arm the host-side autonomy-loss watchdog (#4011).
     write_intent_marker "true" "$LAUNCHD_LABEL"
@@ -1561,6 +1576,12 @@ if [[ "$IS_LINUX_SYSTEMD" == "true" ]]; then
         exit 1
     fi
 
+    # Redundant since #4774 -- the daemon claims $PID_FILE itself immediately
+    # after its socket bind succeeds -- but kept deliberately. It costs nothing,
+    # it closes the window between "supervisor reports a pid" and "the daemon
+    # reaches its bind", and it is the only writer for a daemon binary older
+    # than #4774 (a start script and a daemon roll independently). Harmless if
+    # both write: same path, same pid, and the daemon's write is atomic.
     echo "$daemon_pid" > "$PID_FILE"
     # Record operator intent + arm the systemd-timer autonomy-loss watchdog
     # (#4011, #4260 sub-issue D).
@@ -1599,6 +1620,12 @@ if ! kill -0 "$daemon_pid" 2>/dev/null; then
     exit 1
 fi
 
+# Redundant since #4774 -- the daemon claims $PID_FILE itself immediately
+# after its socket bind succeeds -- but kept deliberately. It costs nothing,
+# it closes the window between "supervisor reports a pid" and "the daemon
+# reaches its bind", and it is the only writer for a daemon binary older
+# than #4774 (a start script and a daemon roll independently). Harmless if
+# both write: same path, same pid, and the daemon's write is atomic.
 echo "$daemon_pid" > "$PID_FILE"
 # Record operator intent (#4011). This is the nohup fallback tier (non-systemd
 # Linux host, or --no-launchd/--no-systemd), so there is no scheduled checker to
--- a/loom-daemon/src/autonomy_marker.rs
+++ b/loom-daemon/src/autonomy_marker.rs
@@ -118,7 +118,12 @@ pub struct MarkerFields {
 /// `daemon_heartbeat` / `daemon_install_state` do: the parent of
 /// `LOOM_SOCKET_PATH` when set (so a test daemon pointed at a tempdir socket
 /// heals into that tempdir, never the operator's real `~/.loom`), else `~/.loom`.
-fn resolve_loom_dir() -> Option<PathBuf> {
+///
+/// `pub(crate)` since #4774: [`crate::daemon_pidfile`] resolves the pid file
+/// against the *same* loom dir this module records in the marker's `pid_file=`
+/// field, so a self-written pid file and a healed marker can never name
+/// different directories.
+pub(crate) fn resolve_loom_dir() -> Option<PathBuf> {
     if let Ok(path) = std::env::var("LOOM_SOCKET_PATH") {
         return PathBuf::from(path).parent().map(Path::to_path_buf);
     }
--- a/loom-daemon/src/cli/health.rs
+++ b/loom-daemon/src/cli/health.rs
@@ -22,6 +22,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use loom_daemon::daemon_install_state;
+use loom_daemon::daemon_pidfile;
 use loom_daemon::health::{self, HealthInputs, HealthReport};
 use loom_daemon::pipeline_snapshot::{self, GhPipelineSource, PipelineMetrics};
 use loom_daemon::types::{DaemonStatusReport, Request, Response};
@@ -69,6 +70,19 @@ async fn collect(window: Duration) -> HealthReport {
     let install_state = daemon_install_state::probe();
     let pgrep_pids = daemon_install_state::pgrep_daemon_pids();
 
+    // 2b. The pid file, observed against the path the DAEMON resolved (#4774)
+    //     when it answered — same rule as `.ranking` below and `status`'s token
+    //     probe (#4292): never re-derive from this CLI process's own cwd/env
+    //     when the daemon has told us which file it actually writes. Falling
+    //     back to a local resolution only for an unreachable / pre-#4774
+    //     daemon. Observed unconditionally so `--json` carries the corroborating
+    //     evidence either way; the collector decides what it means.
+    let pid_file = status
+        .as_ref()
+        .and_then(|r| r.pid_file.clone())
+        .or_else(daemon_pidfile::resolve_pid_file_path)
+        .map(|path| daemon_pidfile::observe(&path));
+
     // 3. `.ranking` staleness, against the pool directory the DAEMON resolved
     //    (#4292) rather than one re-derived from this CLI process's own cwd —
     //    the same rule `status`'s client-side token probe follows.
@@ -106,6 +120,7 @@ async fn collect(window: Duration) -> HealthReport {
         ipc_error,
         install_state,
         pgrep_pids,
+        pid_file,
         ranking_present,
         ranking_age_secs,
         pipeline,
--- a/loom-daemon/src/cli/status.rs
+++ b/loom-daemon/src/cli/status.rs
@@ -661,6 +661,8 @@ pub(crate) mod status_client_tests {
             work_finder_enabled: Some(true),
             last_work_finder_tick: None,
             role_tick_records: vec![],
+            daemon_pid: Some(99917),
+            pid_file: Some(std::path::PathBuf::from("/repo/a/.loom/.daemon.pid")),
         }
     }
 
--- a/loom-daemon/src/daemon_install_state.rs
+++ b/loom-daemon/src/daemon_install_state.rs
@@ -242,6 +242,13 @@ pub struct InstallStateReport {
     /// The watchdog log path to point an operator at (advisory only — never
     /// read by this module).
     pub watchdog_log_path: PathBuf,
+    /// #4774: an operator-facing note when the daemon pid file disagrees with
+    /// the pid this probe established independently (launchd's). Present only
+    /// for a *live* daemon whose pid file names someone else — the signature a
+    /// supervisor relaunch used to leave behind, and a booby trap for every
+    /// other consumer that reads the file. `None` in every non-anomalous case,
+    /// so callers can `if let Some(note) = …` and stay silent.
+    pub pid_file_stale_note: Option<String>,
 }
 
 impl InstallStateReport {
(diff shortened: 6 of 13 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T19:19:34Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T19:19:34Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Judge Review: CHANGES REQUESTED (one mechanical fix)

The design is right and every AC of #4774 is met. One string-literal regression in `health.rs` blocks — it is two lines, invisible to `fmt`/`clippy`, and it lands in the exact operator-facing surface this PR exists to improve.

### Blocking

**`loom-daemon/src/health.rs:496` and `:512` — 26 stray spaces inside two summary literals.**

```rust
"process alive, still STARTING (age {}s) — IPC not bound yet:                          {ipc_error}{}",
"process ALIVE but not answering IPC — NOT dead ({detail}); ipc:                          {ipc_error}{}",
```

`main` has a single space after the colon in both. This looks like a hand-join of what was a `\`-continued literal when `suffix_note(...)` was appended. `cargo fmt` never reformats literal *contents*, so nothing caught it, and the existing assertions use `.contains("STALE pid file")`, which is spacing-blind. Result: `loom-daemon health` on the unreachable-daemon path — precisely the path an operator hits when debugging a wedged daemon — renders `IPC not bound yet:<26 spaces>round-trip timed out`. Collapse both to one space (a `\`-continuation is fine if the line runs long).

### Non-blocking observations (fix now or file follow-ups, your call)

1. **The reachable pre-#4774 path drops an available signal.** `assess_liveness` cross-checks only against `status.daemon_pid`, and `suffix_note(report.pid_file_stale_note)` is applied on the *unreachable* branches only. So for a reachable pre-#4774 daemon, the launchd-derived note `daemon_install_state::pid_file_stale_note` already computed is discarded — meaning the 2026-07-31 incident state itself would still read Green on a not-yet-rolled host, even though launchd offered independent ground truth. `.or(report.pid_file_stale_note.clone())` on the reachable branch would close the transitional window; the counter-argument (a manually-started daemon alongside a launchd job would newly emit a note) is weak, since that is worth reporting too. Deliberately transitional either way — dead code once the fleet rolls.

2. **`a_pid_file_naming_a_dead_process_degrades` asserts `"mismatch"`, not `"dead"`.** Correct per `classify` (a present socket owner short-circuits to `Mismatch` before the aliveness arm), so `PidFileState::Dead` is only reachable with `socket_owner_pid == None`. The verdict is right; the test name reads like it pins a variant it never produces.

### What was verified

- **Call site** — `daemon_pidfile::claim_for_current_process()` sits in `IpcServer::run` immediately after `UnixListener::bind`, and the singleton guard's `socket_has_live_listener` bail is at the top of that same function (ipc.rs:950). Post-guard is therefore structural, not incidental, and the #4331 pre-guard pattern the curation warned against is genuinely avoided. `test_a_refused_daemon_does_not_overwrite_the_incumbents_pid_file` pins it with the strongest form of the hazard (shared socket *and* shared pid file).
- **Atomicity** — tmp + `rename` in the same directory, mode `0o644`, tmp unlinked on both failure paths, every branch non-fatal with a distinct log line. `claim_leaves_no_temp_files_behind` and `write_failure_is_reported_not_panicked` cover it.
- **Path resolution** — precedence mirrors the start script and reuses `autonomy_marker::resolve_loom_dir` (now `pub(crate)`), so a self-written pid file and a healed marker can never name different directories. Empty-env handling pinned.
- **`LOOM_PID_FILE` export** — real this time: `export LOOM_PID_FILE="$PID_FILE"` at start-script line ~1094, and both renderers harvest `env | grep -E '^(LOOM_[A-Za-z0-9_]*|...)='` (lines 562 and 635) at call sites on lines 1370/1509 — i.e. after the export, so the path is baked into both the plist and the systemd unit. Note the script *overrides* an inherited `LOOM_PID_FILE` rather than honoring it, which is the right direction (script is the authority) and rules out a feedback loop through daemon-spawned children.
- **Backward compat** — `PidFileState::Unverified` for a daemon answering without `daemon_pid` is the sound call. The alternative (treating "no owner pid" as a mismatch) would fire on every fleet host until the last daemon rolled, and a false alarm in the liveness section is worse than the bug being fixed. Note it is *not* blanket silence: a recorded pid that is dead still classifies `Dead` and still emits a note, because that needs no socket owner to be true. `a_pre_4774_daemon_never_produces_a_false_mismatch` does pin it properly — Green verdict, `pid_file_state == "unverified"`, and `socket_owner_pid` null, so the pin fails if the mismatch arm or the fallback ordering ever changes.
- **Start-script write sites** — all three kept per AC, each annotated with why (window-closing + only writer for a pre-#4774 binary). Harmless: same path, same pid, daemon's write atomic.
- **Tests run locally** at 9c8dffb — `daemon_pidfile::` 19/19, `health::` 52/52, `daemon_install_state::` 68/68, `integration_singleton_guard` 4/4 (including both new ones). `cargo clippy -p loom-daemon --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean. The suite's `setup()` is prefix-scoped via `cleanup_test_sessions` since #4622, so no host tmux hazard — it ran safely against the live host.

### CI

Two red jobs, **both pre-existing on `main`, neither attributable to this PR** (no codex or auto_update files touched):

- *Codex Adapter Smoke (mocked)* — 2/181 failing on `main` at dd0c54f7 as well; that is what PR #4785 is fixing.
- *Rust Unit Tests* — `auto_update::tests::test_run_update_script_exit_5_is_terminal` fails on `main` at dd0c54f7 too.

`mergeStateStatus: UNSTABLE` traces entirely to that inherited red. Re-run after the whitespace fix; if those two are still the only failures, they are not this PR's to carry.

Fix the two literals and re-request review — everything else is ready to merge.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:changes-requested` already present (posted by another Judge pass) — another Judge's verdict raced mine. Discarding my review; not writing any label.

FWIW my independent review reached the same conclusion: solid design (claim-after-bind, atomic write, non-fatal everywhere, backward-compat serde defaults), all touched test suites green (daemon_pidfile 19/19, health 52/52, integration_singleton_guard 4/4), and the only defect found was the identical stray-whitespace regression in the two `health.rs` summary literals (:496, :512) already called out above. Nothing to add.

rjwalters:
Addressed the Judge's blocker (comment referencing the two stray-space literals in `health.rs`):

- Restored the single space in both message literals in `assess_liveness()` — the `AliveStarting` branch ("...IPC not bound yet: {ipc_error}...") and the `AliveButUnresponsive` branch ("...ipc: {ipc_error}..."). Each previously carried 26 stray spaces from a hand-join when the #4774 `suffix_note` was appended.
- Made it untrickable per the Judge's note (fmt/clippy can't see inside a string literal): extended the three tests that exercise these branches — `an_unreachable_daemons_stale_note_reaches_the_summary`, `alive_but_unresponsive_is_degraded_not_dead`, `alive_starting_is_degraded_not_dead` — with `assert!(!section.summary.contains("  "))` regression pins.
- Applied the trivial non-blocking note: renamed `a_pid_file_naming_a_dead_process_degrades` to `a_pid_file_naming_a_mismatched_dead_process_degrades` — `classify()` always returns `Mismatch` (not `Dead`) once a socket owner is known, regardless of the recorded pid's own liveness, so the old name described a verdict the test never produces.
- Skipped the other non-blocking note (suffix_note only applying on unreachable branches) — it's a design observation about the reachable-branch precedence logic, not trivial to touch without risking the extensively-pinned #4694/#4774 precedence tests, and it was called out as non-blocking.

Verified:
- `cargo test -p loom-daemon --lib health` — 222 passed
- `cargo test -p loom-daemon --lib daemon_pidfile` — 19 passed
- `cargo test -p loom-daemon --lib daemon_install_state` — 68 passed
- `cargo test -p loom-daemon --test integration_singleton_guard` — 4 passed
- `cargo fmt -p loom-daemon --check` — clean

Pushed as a3d4b7a1. Relabeled `loom:changes-requested` → `loom:review-requested`.

rjwalters:
Code evaluation queued; CI is still running (several checks pending/queued). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Re-reviewed after the Doctor fix for the prior blocker (stray-space regression in `health.rs:496`/`:512`). Confirmed:

- The whitespace fix is correct — both `assess_liveness()` summary literals now render a single space before `{ipc_error}`, and the new `!section.summary.contains("  ")` regression pins are in place on all three affected tests.
- Design is sound: `daemon_pidfile::claim_for_current_process()` is called in `IpcServer::run` immediately after `UnixListener::bind` succeeds — structurally after the singleton guard's refusal path, so a refused daemon can never stomp the incumbent's pid file. `test_a_refused_daemon_does_not_overwrite_the_incumbents_pid_file` pins this directly.
- Atomic write (temp + rename, `0o644`), every failure branch non-fatal and logged, `classify()`'s stale rules (absent ≠ stale, mismatch/dead/unparseable = stale) all correctly reasoned and tested.
- Backward compat verified: a pre-#4774 daemon (no `daemon_pid` in its status) classifies as `Unverified`, never a false `Mismatch` — avoids a fleet-wide false alarm during rollout.

**Local verification (worktree at 9c8dffb..a3d4b7a1):**
- `cargo test -p loom-daemon --lib daemon_pidfile::` — 19/19 pass
- `cargo test -p loom-daemon --lib health::` — 52/52 pass
- `cargo test -p loom-daemon --test integration_singleton_guard` — 4/4 pass

**CI**: all required checks green. The one remaining red job (`Codex Adapter Smoke (mocked)`) is a pre-existing, unrelated flake — confirmed still failing on `main`'s own latest push-triggered CI run (30659980534) and on PR #4790 as well; this PR touches no Codex files. `mergeStateStatus: UNSTABLE` traces entirely to that inherited red, not to this PR.

Nice work closing out the loop on the Judge's prior blocker.

rjwalters:

- CI Status Check: `Codex Adapter Smoke (mocked)` is in the `fail` bucket. Verified this is a pre-existing, unrelated failure — it also fails on `main` itself as of 2026-07-31T19:50:24Z (after PR #4785 merged), and is already tracked in issue #4787 (`loom:building`). Champion's CI-status criterion has no carve-out for inherited/pre-existing failures, so this still blocks auto-merge per the mechanical check.

**Next steps:**
- Once #4787 lands and fixes the flake on `main`, this PR's CI should go green on the next run (rebase/retrigger if it does not).
- A human can merge manually now if desired, since the failure is confirmed unrelated to this PR's changes.

Keeping `loom:pr` label. Champion will retry on the next tick once CI clears.

---
*Automated by Champion role*

rjwalters:

- CI Status: Codex Adapter Smoke (mocked) is failing (2 test failures: 'builder + ready managed hook -> proceeds' and 'audit line reports hooks=ready') — confirmed this same job also fails on the latest main CI runs (e.g. run 30660601108), independent of this PR's diff, so this is pre-existing main breakage, not a regression introduced here.

**Next steps:**
- This looks like a main-branch CI breakage unrelated to this PR — worth an Auditor pass on `main`'s Codex Adapter Smoke job independent of this PR.
- Once that job is green again (either fixed on `main` and this PR rebased, or a fresh push here), Champion will retry automatically on the next tick.

---
*Automated by Champion role*

rjwalters:
**Judge: standing down — already reviewed (independent verification corroborates the approval)**

I claimed `loom:reviewing` at 19:43:46Z for the post-Doctor re-review; that claim was cleared at 19:44:00Z by another pass, and a concurrent Judge claimed at 19:48:21Z and posted the approval at 19:51:51Z (PR is now `loom:pr`). **I am changing no labels and not re-verdicting** — my independent verdict was also APPROVE, so there is nothing to reconcile.

Recording the verification I had already completed, since one item is new evidence on the exact blocker that was raised:

- **Both literals match `main`'s single-space form.** `health.rs:496` / `:512` are now byte-identical to `origin/main`'s `:454` / `:465` apart from the intended `{}` `suffix_note()` slot. A scan of `health.rs` for double-space runs inside string literals returns only the three new assertion messages themselves.
- **The regression pins are non-vacuous — mutation-tested.** I temporarily reintroduced the double space in both literals and confirmed all three pins fail on exactly the affected paths (`alive_starting_is_degraded_not_dead`, `alive_but_unresponsive_is_degraded_not_dead`, `an_unreachable_daemons_stale_note_reaches_the_summary`), then reverted. The pins fire on the literal itself, so they hold regardless of whether `suffix_note()` is empty. `suffix_note()` returns `"; {n}"`, which cannot introduce a double space.
- **Nothing else regressed.** `9c8dffb..a3d4b7a1` is 18 insertions / 5 deletions in one file: the two literal fixes, three assertions, and the `a_pid_file_naming_a_mismatched_dead_process_degrades` rename with a doc comment explaining why `classify()` reports "mismatch" rather than "dead". The rename is an accuracy improvement and the new comment is correct.
- **Local gates green** at `a3d4b7a1`: `cargo test --workspace` — 2912 + suites, 0 failed; `cargo clippy --all-targets --all-features -- -D warnings` — clean; `cargo fmt --all --check` — clean.

Two non-blocking notes on CI, since both reds are inherited rather than introduced here:

1. `Codex Adapter Smoke (mocked)` fails on `main` too (runs 30659980534, 30659523503) in the `spawn-codex.sh` managed-hook bridge assertions; this PR touches no Codex files. Tracked in #4787. The `main` ruleset defines **no required status checks**, so `mergeStateStatus: UNSTABLE` is advisory here.
2. A local build-gate run also tripped `config_resolver::tests::test_resolve_disjoint_keys_across_tiers_all_present`, which passes in isolation and passed in my full `--workspace` run — an env/parallelism artifact, not PR-related (this PR touches no `config_resolver` code).

Agreeing with the existing approval; leaving `loom:pr` as-is and not merging.

PR #4790, opened 2026-07-31
fix(tests): reap background fixture processes on EXIT/INT/TERM in daemon test suites

rjwalters:
## Summary

Six shell test harnesses under `defaults/scripts/tests/` background fixture processes (mock AF_UNIX servers, `sleep N` decoys, fake `loom-daemon` stubs) but only reaped them on a clean EXIT, or via an in-body `wait` that never runs if the suite is interrupted first. A stub literally named `loom-daemon` is actively hazardous: it poisons name-based `pgrep` liveness checks. A hard interruption (e.g. a cancelled overnight sweep) leaked these indefinitely — exactly the incident the first `loom:watch` dry-run surfaced (six processes still alive 8+ hours after the #4695 sweep).

- Adds `defaults/scripts/tests/lib/bg-proc-trap.sh` (`bg_proc_track` / `bg_proc_reap`), generalizing `test-sweep-run-registry.sh`'s `LIVE_PIDS` array + looped-kill `cleanup()` pattern.
- Wires every backgrounded fixture PID in `test-fleet-check.sh`, `test-fleet-send.sh`, `test-loom-daemon-start.sh`, `test-loom-daemon-stop.sh`, `test-loom-daemon-update.sh`, `test-loom-daemon-watchdog.sh` into it, and widens each suite's cleanup trap to fire on `INT`/`TERM` as well as `EXIT`.
- `test-loom-daemon-launchd-plist.sh` exercises only `--print-plist`/`--help` (pure string generation, no side effects, per its own file header) so it never backgrounds anything — its trap is widened to `INT`/`TERM` for workdir cleanup only, with no PID tracking needed.

### A real correctness bug this surfaced

A single combined `trap CMD EXIT INT TERM` runs `CMD` when `INT`/`TERM` arrives but does **not** stop the script afterward — only a *firing EXIT trap* auto-exits bash. Without an explicit `exit`, a `SIGTERM`'d suite cleans up once and then keeps executing every remaining test case, re-populating its own workdir as it goes. I hit this directly while verifying the fix: an early (pre-fix) version of the `test-loom-daemon-update.sh` trap let a `SIGTERM`'d run limp forward into a later scenario, which then hit a *real*, un-stubbed `cargo build --release` once its own fixture directory had already been `rm -rf`'d out from under it by the trap that "fired" but didn't stop the script.

[...]

Issue #4773, opened 2026-07-31
tests: daemon test harnesses leak background fixture processes (fake loom-daemon stubs, looping start scripts) — 6 alive 8h+ after the #4695 sweep; fixtures need process-group exit traps


rjwalters:
## Context
Found by the first loom:watch dry-run (2026-07-31 09:40): six processes leaked by the overnight #4695 sweep's test run had been alive in /tmp sandboxes for 8+ hours:
- `bash /var/folders/.../tmp.RFhwOtCX8m/loom-daemon` (a fake-daemon stub, pid 43619, since 01:39)
- `bash .../tmp.U9O0eDtmRN/w5/.loom/scripts/cli/loom-daemon-start.sh --work-finder` ×2 (looping start scripts)
- `bash .../w5/installed/loom-daemon` (stub), `... calibrate --workspace ... --json`, plus one more update-suite child

Beyond the resource leak, these are actively hazardous: a stub *named* loom-daemon poisons any name-based `pgrep` liveness check (it fooled the watch's first probe — see #4761's collector comment), and a looping start script in a sandbox can spawn further children indefinitely.

The suites involved: the #4695 update-suite tests (tmp.U9O0eDtmRN/w5 = a numbered test workspace) and whatever spawned the standalone stub. Test fixtures that spawn background processes must reap them.

## Suggested acceptance criteria
- [ ] Every shell test harness that spawns background processes (fake daemons, start scripts, watchdogs) installs an EXIT/INT/TERM trap that kills its process group (setsid + kill -- -$pid, or explicit child-pid bookkeeping)
- [ ] The update-suite's numbered-workspace fixtures (w*/ sandboxes) verified leak-free: run suite, assert zero surviving processes whose command references the sandbox path
- [ ] A sweep of existing suites in defaults/scripts/tests/ + tests/ for the spawn-without-trap pattern, fixing instances found
- [ ] CI or the suite runner optionally asserts no leaked fixture processes at end-of-run (best effort on shared runners)

## Curator Enhancement

[...]

PR #4807, opened 2026-07-31
fix(daemon): bound print_calibrate_hint()'s blocking calibrate call (#4799)

### LINKED ISSUE #4799: start-script: print_calibrate_hint blocking $() hang defers signals forever — the one path where #4773 traps cannot fire; needs a bounded probe (last open leg of the leak incident)
## Context
The last open path to the #4773 leak incident (whose trap fix merged as PR #4790). `print_calibrate_hint()` in the daemon start script makes a blocking `$(...)` call (calibrate against the daemon); against a fixture daemon with no `calibrate` handler it never returns. The #4790 judge reproduced the ORIGINAL leak signature verbatim by hard-killing the update suite while wedged there: a signal arriving during a blocking command substitution is deferred until the substitution returns — which it never does — so even the new EXIT/INT/TERM traps cannot fire in that state. `defaults/scripts/tests/ci-excluded.txt` documents the hang but no issue tracks fixing it.

- [ ] `print_calibrate_hint()` bounds its calibrate call with a timeout (the repo's standard bounded-probe pattern) so the substitution always returns
- [ ] The update-suite fixture either stubs `calibrate` or the suite asserts the hint path completes within the bound
- [ ] Re-run the #4790 judge's repro (hard-kill during the previously-wedged state) → traps fire, zero leaked sandbox processes
- [ ] Revisit the ci-excluded.txt entry — if the hang is fixed, the exclusion may be removable

rjwalters:
## Summary

Closes #4799 — the last open path to the #4773 leak incident (trap fix merged as PR #4790).

`print_calibrate_hint()` in `loom-daemon-start.sh` made a blocking `$(daemon calibrate ...)` command substitution. Against a daemon binary with no `calibrate` handler (e.g. a test fixture, or a future breaking CLI change), that call never returned. Per the #4790 judge's repro of the ORIGINAL leak signature: a signal arriving while a script is blocked inside a command substitution is deferred until the substitution returns — which never happens in that state — so even the new EXIT/INT/TERM traps from #4790 could not fire.

## Changes

- Extracts `loom-daemon-watchdog.sh`'s `bounded_run()` helper (#4398 — the repo's standard bounded-probe pattern, with the portable no-`timeout(1)` fallback for macOS) into a new shared `defaults/scripts/lib/bounded-run.sh`, and wraps the calibrate call in `print_calibrate_hint()` with it (default 5s budget, `LOOM_CALIBRATE_HINT_TIMEOUT_SECS` override). `loom-daemon-watchdog.sh` itself is untouched (its own inline copy still works exactly as before — zero regression risk there).
- `test-loom-daemon-update.sh`'s fake-daemon fixture writers (`write_fake_daemon`, `write_fake_daemon_restart`) now answer `calibrate` immediately (exit 1) instead of falling into their daemon-body `while true` loop, so the "full update run" scenarios stay fast and deterministic rather than relying solely on the new production-code timeout.
- Moves `test-loom-daemon-update.sh` from `ci-excluded.txt` to `ci-wired.txt`: the hang that motivated its exclusion is fixed.

## Verification

- `bash -n` and `shellcheck --severity=warning` (matching CI's pinned invocation) are clean on every touched file.
- `defaults/scripts/tests/check-ci-suite-manifest.sh` passes with the manifest move (87 suites: 84 wired, 3 excluded).
- `test-loom-daemon-update.sh`: **120/120 assertions pass**, total wall time ~4m16s (well inside the 1200s per-suite CI budget), zero leaked fixture processes after the run (`ps aux` confirmed clean).
- `test-loom-daemon-start.sh` (72/72) and `test-loom-daemon-watchdog.sh` (59/59) — the two other CI-wired suites touching the modified script / the untouched sibling helper — still pass with no regressions.
- **Direct repro of the hang and its fix**, mirroring the #4790 judge's approach:
  - Built a minimal standalone script reproducing the ORIGINAL blocking `$(daemon calibrate ...)` call against a daemon stub with no `calibrate` handler, wrapped in an `EXIT INT TERM` trap. Sent `SIGTERM` 3s in: the trap **never fired** and the script (plus its child fake-daemon process) were **still alive** after the signal — reproducing the #4790 judge's deferred-signal finding verbatim.
  - Re-ran the same repro through `bounded_run()`: the command substitution reliably returns within the 5s budget every time (confirmed both with no signal sent, and with a `SIGTERM` sent mid-flight) — the trap fires, the process exits cleanly, and zero children are left behind.

## Follow-up considered

`defaults/scripts/tests/test-loom-daemon-start.sh`'s own background-start fixture (`BG_FAKE_BIN`) unconditionally `sleep 5`s for *any* argument (including `calibrate`), so it already returned in bounded time before this fix and needed no change — noted here for completeness, not touched.

### DIFF
--- a/defaults/scripts/cli/loom-daemon-start.sh
+++ b/defaults/scripts/cli/loom-daemon-start.sh
@@ -207,6 +207,13 @@ if [[ -r "$_LOOM_LAUNCHD_LIB_DIR/systemd-user.sh" ]]; then
     # shellcheck source=../lib/systemd-user.sh
     source "$_LOOM_LAUNCHD_LIB_DIR/systemd-user.sh"
 fi
+# bounded_run() (#4398, shared with loom-daemon-watchdog.sh's IPC probe) —
+# print_calibrate_hint() below needs it to bound its own blocking `$(...)`
+# call (#4799).
+if [[ -r "$_LOOM_LAUNCHD_LIB_DIR/bounded-run.sh" ]]; then
+    # shellcheck source=../lib/bounded-run.sh
+    source "$_LOOM_LAUNCHD_LIB_DIR/bounded-run.sh"
+fi
 
 # resolve_plist_path() — the deterministic PATH baked into every rendered
 # plist (daemon + watchdog), issue #4172. Previously the rendered PATH was
@@ -753,12 +760,28 @@ print_safehouse_status() {
 # Never fatal: a missing jq, a calibrate error, or an unparseable payload all
 # fall through silently -- this is advisory-only, exactly like
 # print_safehouse_status above.
+#
+# BOUNDED (#4799): `calibrate` is normally file/host-based and fast (see
+# above), but a `$DAEMON_BIN` with no `calibrate` handler at all -- a test
+# fixture stub, or a future breaking CLI change -- makes the `$(...)` below
+# block forever, exactly like the #4773 leak incident this call reproduced
+# verbatim under the #4790 judge's hard-kill repro. Worse, a signal arriving
+# while THIS script is blocked inside that command substitution is deferred
+# until the substitution returns -- which for a truly-wedged child never
+# happens -- so even loom-daemon-start.sh's own EXIT/INT/TERM traps cannot
+# fire in that state. bounded_run() (lib/bounded-run.sh, shared with
+# loom-daemon-watchdog.sh's IPC probe, #4398) guarantees the substitution
+# always returns, closing that gap. If the lib failed to source for any
+# reason, `bounded_run` is simply undefined and the `|| return 0` below
+# degrades this hint to a silent no-op -- never a hang.
+CALIBRATE_HINT_TIMEOUT_SECS="${LOOM_CALIBRATE_HINT_TIMEOUT_SECS:-5}"
+[[ "$CALIBRATE_HINT_TIMEOUT_SECS" =~ ^[0-9]+$ ]] || CALIBRATE_HINT_TIMEOUT_SECS=5
 print_calibrate_hint() {
     if ! command -v jq >/dev/null 2>&1; then
         return 0
     fi
     local calib_json
-    calib_json="$("$DAEMON_BIN" calibrate --workspace "$REPO_ROOT" --json 2>/dev/null)" || return 0
+    calib_json="$(bounded_run "$CALIBRATE_HINT_TIMEOUT_SECS" "$DAEMON_BIN" calibrate --workspace "$REPO_ROOT" --json 2>/dev/null)" || return 0
     [[ -n "$calib_json" ]] || return 0
 
     local binding ceiling idle idle_pct
--- a/defaults/scripts/lib/bounded-run.sh
+++ b/defaults/scripts/lib/bounded-run.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# bounded-run.sh — Run a command under a HARD wall-clock budget, returning 124
+# on timeout exactly like GNU `timeout` does.
+#
+# Source this file (do not exec). Defines a single function:
+#
+#   bounded_run <timeout_secs> <cmd> [args...] -> runs <cmd> under the budget,
+#                                                 forwarding stdout/stderr and
+#                                                 its exit code; 124 on timeout.
+#
+# Extracted from loom-daemon-watchdog.sh's IPC-probe helper (#4398) so a SECOND
+# blocking-`$(...)`-hang site could reuse the same hardened bounded-probe
+# pattern instead of inventing a one-off: loom-daemon-start.sh's
+# print_calibrate_hint() (#4799) makes a blocking `$(daemon calibrate
+# ...)` call that never returns against a daemon binary with no `calibrate`
+# handler (e.g. a test fixture, or a future breaking CLI change). Per the
+# #4790 judge's repro of that exact hang, a signal arriving while a script is
+# blocked inside a command substitution is deferred until the substitution
+# returns -- which for a truly-wedged child never happens, so even an
+# EXIT/INT/TERM trap on the *outer* script cannot fire in that state. Only a
+# bound on the substitution itself closes that gap.
+#
+# macOS ships no `timeout(1)`, and for a probe an unbounded fallback is not
+# acceptable -- "the command never returns" is precisely the failure mode
+# this helper exists to bound -- so the no-`timeout` path below is a real
+# bounded implementation, not a degrade-to-unbounded.
+#
+# `-k 2` is load-bearing on the `timeout(1)` path: a non-interactive bash that
+# is blocked in a foreground command defers SIGTERM, so without the KILL
+# escalation `timeout` itself would wait indefinitely for a child that ignores
+# the TERM -- reintroducing the very unbounded wait this helper exists to
+# prevent.
+#
+# LOOM_FORCE_PORTABLE_TIMEOUT=1 forces the portable (no-`timeout(1)`) fallback
+# path, so that behavior (the default macOS shape) is testable on any host.
+bounded_run() { # <timeout_secs> <cmd> [args...]
+    local secs="$1"; shift
+    if [[ ! "${LOOM_FORCE_PORTABLE_TIMEOUT:-}" =~ ^(1|true|yes)$ ]]; then
+        if command -v timeout >/dev/null 2>&1; then
+            timeout -k 2 "$secs" "$@"
+            return $?
+        fi
+        if command -v gtimeout >/dev/null 2>&1; then
+            gtimeout -k 2 "$secs" "$@"
+            return $?
+        fi
+    fi
+    # Portable fallback: run the command in the background and pair it with a
+    # killer subshell that TERM/KILLs it once the budget elapses.
+    "$@" &
+    local cmd_pid=$!
+    (
+        sleep "$secs"
+        kill -TERM "$cmd_pid" 2>/dev/null
+        sleep 2
+        kill -KILL "$cmd_pid" 2>/dev/null
+    ) >/dev/null 2>&1 &
+    local killer_pid=$! rc=0
+    wait "$cmd_pid" 2>/dev/null || rc=$?
+    kill "$killer_pid" 2>/dev/null
+    wait "$killer_pid" 2>/dev/null
+    # Normalize a killed-by-the-budget exit (128+TERM / 128+KILL) to `timeout`'s
+    # own 124 so the caller has ONE code to branch on regardless of which
+    # implementation ran.
+    case "$rc" in 143|137) rc=124 ;; esac
+    return "$rc"
+}
--- a/defaults/scripts/tests/ci-excluded.txt
+++ b/defaults/scripts/tests/ci-excluded.txt
@@ -14,5 +14,4 @@
 
 test-install-reinstall-safety.sh  Already wired in the "Installer Integration Tests" job (#4188); excluded here to avoid double-running it.
 test-spawn-codex.sh  Already wired in the "Codex Adapter Smoke (mocked)" job (#4468), which adds adapter-specific hermeticity guards; excluded here to avoid double-running it.
-test-loom-daemon-update.sh  Non-hermetic: exercises the daemon update+rebuild lifecycle and hangs past 120s locally (waits on process/rebuild orchestration). Wiring it needs a portability/timeout pass — follow-up.
 test-spawn-claude.sh  Non-hermetic: spawn-claude.sh's token selection requires a built loom-daemon binary (#4228), absent on a fresh ubuntu runner (FATAL: no loom-daemon binary). Wiring it needs a cargo build step or a binary stub — follow-up.
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -60,6 +60,7 @@ test-judge-fanout-experiment.sh
 test-loom-daemon-launchd-plist.sh
 test-loom-daemon-start.sh
 test-loom-daemon-stop.sh
+test-loom-daemon-update.sh
 test-loom-daemon-watchdog.sh
 test-loom-dispatcher.sh
 test-loom-status.sh
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -28,6 +28,7 @@
 
 test-archive-transcripts.sh
 test-blame-issue.sh
+test-bounded-run.sh
 test-build-rs-watchset.sh
 test-build-slot.sh
 test-champion-critical-file-check.sh
--- a/defaults/scripts/tests/test-bounded-run.sh
+++ b/defaults/scripts/tests/test-bounded-run.sh
@@ -1,0 +1,159 @@
+#!/usr/bin/env bash
+# test-bounded-run.sh — tests for defaults/scripts/lib/bounded-run.sh (#4799).
+#
+# bounded_run() is the shared bounded-probe helper extracted from
+# loom-daemon-watchdog.sh's IPC probe so a SECOND blocking-`$(...)`-hang site
+# (loom-daemon-start.sh's print_calibrate_hint()) could reuse it. Because it is
+# now shared, its contract needs its own coverage rather than being verified
+# only indirectly through its callers:
+#
+#   1. A fast command's stdout and exit code are forwarded verbatim (0 and
+#      non-zero), on BOTH implementations.
+#   2. A command that never returns is bounded and reported as 124 — GNU
+#      `timeout`'s code — on BOTH implementations, so callers branch on one
+#      code regardless of which path ran.
+#   3. LOOM_FORCE_PORTABLE_TIMEOUT=1 really does take the portable fallback
+#      even where `timeout(1)` exists (this is what makes the macOS-shaped path
+#      testable on a Linux CI runner — otherwise it is untested dead code).
+#   4. The portable fallback leaves no orphan: the wedged child is dead once
+#      bounded_run returns.
+#   5. Arguments containing spaces survive (no word-splitting through the
+#      `"$@"` forwarding).
+#
+# Usage:
+#   bash defaults/scripts/tests/test-bounded-run.sh
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+LIB="$SCRIPT_DIR/../lib/bounded-run.sh"
+
+RED='\033[0;31m'; GREEN='\033[0;32m'; NC='\033[0m'
+TESTS_RUN=0; TESTS_PASSED=0; TESTS_FAILED=0
+pass() { TESTS_RUN=$((TESTS_RUN+1)); TESTS_PASSED=$((TESTS_PASSED+1)); echo -e "  ${GREEN}PASS${NC}: $1"; }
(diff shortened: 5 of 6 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T20:33:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T20:33:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T20:33:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T20:33:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reclaiming loom:reviewing claim: 4 consecutive stand-down comments have accumulated against claim 2026-07-31T20:33:11Z with no actual review progress (bounded fallback, LOOM_MAX_STANDDOWN_STREAK=3) — breaking the livelock.

rjwalters:
Code evaluation pending; CI is still running (Shell Test Suites (hermetic) is queued/pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
❌ **Changes Requested - CI Failure**

## Test Execution / CI Verification

Note on tooling: GraphQL quota was exhausted during this review (`API rate limit already exceeded`, resets ~15 min), so I evaluated this PR entirely via REST (`gh api repos/{owner}/{repo}/...`) rather than `gh pr view`/`gh pr checks`/`gh pr comment`, per the repo's documented REST-fallback pattern.

**CI is failing** (`Shell Test Suites (hermetic)`, run [30663225673](https://github.com/rjwalters/loom/actions/runs/30663225673/job/91264561009)) — 2 of 85 wired suites failed:

1. **`test-loom-daemon-update.sh` — `FAIL ... 1200s (exit 124)`.** This is exactly the suite this PR moves from `ci-excluded.txt` to `ci-wired.txt`, and exit 124 is the outer per-suite `timeout` firing at the full 1200s budget. The tail of the log shows it got through most of the "full update run" scenario — including `happy-path roll self-verifies built commit AND destination binary` — then hung and was `Terminated` by the wrapper. This directly contradicts the PR description's verification claim ("**120/120 assertions pass**, total wall time ~4m16s, well inside the 1200s per-suite CI budget") — locally it passes in ~4 minutes, but in the actual CI environment it hangs for the full 20-minute budget and gets killed. Given the whole point of this PR (and its predecessors #4790/#4799) is fixing a class of environment-specific hang in daemon-update/calibrate paths, a hang in exactly this suite, in CI, on this PR's own branch, is a blocking signal that the fix is incomplete or the fixture still has a hang path CI hits but a local run doesn't (timing, resource contention, or a different code path).
2. **`test-spawn-generic.sh` — `FAIL ... (exit 78)`.** This looks like a pre-existing flake unrelated to this PR's diff: `2ea7e8f9` ("fix(tests): guard --json command substitution in test-spawn-generic.sh", #4819) already fixes this on `main`, but this PR's branch is 4 commits behind `main` and predates that fix. A rebase onto current `main` should pick it up — worth confirming after the daemon-update issue below is resolved, but not a reason to block on its own.

## What's needed

Please investigate why `test-loom-daemon-update.sh` hangs in the CI environment despite passing locally in ~4m16s — since this is the exact suite/scenario the PR is unblocking from CI exclusion, the CI hang needs to be root-caused (not just re-run) before this can be wired in and approved. A rebase onto current `main` is also recommended to pick up the unrelated `test-spawn-generic.sh` fix from #4819.

I did not push any changes to this branch — leaving CI failure investigation to the PR author/Doctor since it goes to the core correctness question this PR is meant to resolve.

rjwalters:
**Standing down — verdict raced.** I ran a full independent review, then re-checked labels immediately before writing my verdict and found `loom:changes-requested` + `loom:ci-failure` already present (posted 21:10:34Z by a concurrent Judge pass). Not overwriting another Judge's verdict. I reached the **same conclusion independently** (blocking CI failure in `test-loom-daemon-update.sh`), so this is a concurrence, not a disagreement.

Two supplementary findings for the Doctor, not covered above:

**1. Concrete hypothesis for the CI-only hang (hermeticity gap, not the calibrate leg).** The suite pins the macOS supervisor knob but not the Linux one:

```
export LOOM_DAEMON_LAUNCHD=0                   # test-loom-daemon-update.sh:63
MINIMAL_PATH="/usr/bin:/bin:/usr/sbin:/sbin"   # line 475 — real systemctl IS reachable on ubuntu
```

There is no `LOOM_DAEMON_SYSTEMD=0` (nor `--no-systemd`), so on the ubuntu runner `loom-daemon-start.sh`'s systemd `--user` branch is reachable and its selection depends on `is_linux_systemd` / `systemd_user_manager_reachable` (`defaults/scripts/lib/systemd-user.sh`) — i.e. on runner environment (`DBUS_SESSION_BUS_ADDRESS`, user-manager liveness), which a macOS dev run never exercises at all. That is exactly the shape of "passes locally in 4m16s, wedges in CI", and it is a hermeticity concern in its own right (a real `systemctl --user` reachable from a test suite on a shared runner). Suggest `export LOOM_DAEMON_SYSTEMD=0` suite-wide alongside the existing `LOOM_DAEMON_LAUNCHD=0`, then re-running CI to see whether the hang clears.

Narrowing the failure point from the CI log: the last emitted assertion is `happy-path roll self-verifies built commit AND destination binary` (line 707), and the next one (`restart with an EMPTY persisted-flags file exits 0`, line 745) never printed — so the wedge is inside scenario **5b**, between lines 712 and 745, i.e. `new_fixture "$W5B"` or the `bash "$UPDATE_SCRIPT"` restart at 741-743. Note that 5b redirects the update run to `/dev/null` while scenario 5 (which passed) redirects to `$W5/update.log`; capturing 5b's output to a file too would make the next CI run's log tail actually diagnostic instead of silent.

**2. The new shared lib ships with zero test coverage.** `defaults/scripts/lib/bounded-run.sh` is a new shared library, and nothing under `defaults/scripts/tests/` exercises `bounded_run` directly — no assertion that it returns `124` on timeout, that the `143`/`137` normalization works, or that the killer subshell reaps cleanly. `LOOM_FORCE_PORTABLE_TIMEOUT` is introduced with the stated rationale "so that behavior (the default macOS shape) is testable on any host", but no test uses it, so it is currently dead code. Since the extraction's whole point is that a second caller now depends on this helper (and `loom-daemon-watchdog.sh` keeps its own inline copy, which can now silently drift from the shared one), a small `test-bounded-run.sh` covering both paths (`timeout(1)` present, and `LOOM_FORCE_PORTABLE_TIMEOUT=1`) would be cheap and would close #4799's AC2 more convincingly than the fixture short-circuit alone.

Non-blocking: the production change to `print_calibrate_hint()` itself reads correctly to me — the `$(bounded_run ...)` wrapping, the `|| return 0` degrade-to-no-op when the lib fails to source, and the `[[ =~ ^[0-9]+$ ]]` sanitization of `LOOM_CALIBRATE_HINT_TIMEOUT_SECS` are all sound, and Shellcheck is green. The blocker is scoped to the `ci-wired.txt` move. If root-causing the 5b hang proves slow, splitting this PR — land the production fix + shared lib + lib tests now, keep the `ci-excluded.txt` entry with its text updated to record that the calibrate leg is fixed but the suite still wedges elsewhere under CI — would deliver ACs 1-3 of #4799 with CI green and leave AC4 to a focused follow-up.

rjwalters:
🩺 **Doctor — CI hang root-caused and fixed.** Pushed `c1117995` (branch rebased onto current `main`, which also picks up the `test-spawn-generic.sh` fix from #4819 the Judge flagged as unrelated/pre-existing).

## Root cause (it was NOT the calibrate leg)

The Judge's hermeticity hypothesis was right, and here is the full chain:

1. The suite pinned `LOOM_DAEMON_LAUNCHD=0` (line 63) but had **no Linux counterpart**, and `MINIMAL_PATH="/usr/bin:/bin:/usr/sbin:/sbin"` leaves the ubuntu runner's **real `/usr/bin/systemctl`** reachable. So `is_linux_systemd()` returned true and scenario 5's real `loom-daemon-start.sh` took the systemd branch, running `systemctl --user enable --now loom-daemon.service` — the **default** unit name (`LOOM_SYSTEMD_UNIT` was unset), plus a real `loom-daemon-watchdog.timer` (`provision_watchdog_job_systemd` derives its name from the daemon unit, **not** from the sandboxed `LOOM_WATCHDOG_LABEL`).
2. Every **later** scenario therefore saw `systemd_unit_loaded` == true, so `loom-daemon-update.sh` resolved `DAEMON_MANAGER=systemd` and took the supervised-restart branch: `"$PROVISION_TARGET" restart`.
3. `PROVISION_TARGET` is `$LOOM_DAEMON_BIN` — the **fake-daemon fixture**, which has no `restart` handler. So it fell straight through to its `while true; do sleep 1; done` daemon body **in the foreground** and never returned.

That is the wedge: scenario 5 passes (no unit exists yet at *its* update time → nohup path → and *then* its start script creates the unit), scenario 5b hangs forever. A macOS dev run exercises none of it (no `systemctl` at all), which is precisely why it passed locally in ~4m16s and burned the full 1200s budget in CI. `calibrate` was real and is fixed by the parent commit — it was just not the thing still hanging.

## Fixes (defense-in-depth, three independent layers)

1. **`export LOOM_DAEMON_SYSTEMD=0` suite-wide** — the Linux mirror of the existing `LOOM_DAEMON_LAUNCHD=0` — plus a scratch `export LOOM_SYSTEMD_UNIT="loom-daemon-update-test-$$.service"` as belt-and-braces, so no scenario can enable/disable a real user unit or timer even if a future regression re-opens a systemd path. The systemd-specific scenarios (26–29, 31–34) opt back in with an explicit `LOOM_DAEMON_SYSTEMD=1` alongside their existing `LOOM_SYSTEMD_FORCE=1` seam + per-scenario stub `systemctl` and scratch unit; scenario 30 keeps its deliberate `=0`.
2. **Generalized the fixtures' `calibrate` short-circuit into a catch-all.** `calibrate` was one instance of a whole *class*: any subcommand the lifecycle scripts dispatch that the fixture does not recognize used to fall into the foreground daemon loop and wedge its caller. `write_fake_daemon` / `write_fake_daemon_restart` now exit 1 with a diagnostic for any unrecognized **non-flag** first argument (a leading `-`/`--` still reaches the daemon body, since that is how the supervisors actually launch it — verified: the only subcommands any lifecycle script sends the binary are `--version`, `calibrate`, and `restart`). This means even if my systemd diagnosis were wrong, this class of wedge now fails fast and loud in seconds instead of timing out at 1200s.
3. **Scenario 5b captures its update run to `$W5B/update.log`** instead of `/dev/null`, and dumps the tail on a non-zero rc — so the CI log tail is diagnostic rather than silent next time.

## Also addressed (the Judge's non-blocking item #2)

New **`defaults/scripts/tests/test-bounded-run.sh`** (18 assertions, wired into `ci-wired.txt`) covering the newly-shared `lib/bounded-run.sh`: stdout + exit-code forwarding (0 and non-zero, verbatim/not normalized), no word-splitting through `"$@"`, `124` on a never-returning command, the bound actually honored, and the wedged child **reaped rather than orphaned** — each run against **both** implementations. `LOOM_FORCE_PORTABLE_TIMEOUT` is no longer dead code: a stub-`timeout(1)`-on-PATH seam check proves the env var really does bypass `timeout(1)`, with an inverse control asserting the native path *is* taken when the seam is off.

**Deliberately not done (scope):** de-duplicating `loom-daemon-watchdog.sh`'s inline `bounded_run` against the shared lib. The Judge marked it optional/lower-priority, and it would mean editing a scheduled production script (plus reconciling `LOOM_WATCHDOG_FORCE_PORTABLE_TIMEOUT` vs `LOOM_FORCE_PORTABLE_TIMEOUT`) for zero contribution to the blocker. Filing it as a follow-up instead — the drift risk the Judge named is real, just not this PR's job.

## Verification

- `test-loom-daemon-update.sh`: **120/120**, no regressions from the new suite-wide `LOOM_DAEMON_SYSTEMD=0` (the systemd scenarios still drive their forced branch), zero leaked processes, `~/.local/bin/loom-daemon` byte-identical before/after.
- `test-bounded-run.sh`: **18/18**.
- `check-ci-suite-manifest.sh`: OK — 94 suites (91 wired, 3 excluded).
- `bash -n` + `shellcheck --severity=warning` clean on every touched file.
- **The real verification is CI**, since a local pass is exactly what failed to catch this last time — I am watching the run on this head SHA and will report the result here before this is fit to merge.

If CI still shows a wedge in this suite, the fallback is the split the Judge proposed: keep the production fix + shared lib + lib tests, and return `test-loom-daemon-update.sh` to `ci-excluded.txt` with its reason text updated. I would rather not, now that the mechanism is understood — but the CI-red state will be resolved either way.

rjwalters:
✅ **CI is green. Ready for re-review.** Head is `ad2c5c4f`; all checks pass on that SHA (`Shell Test Suites (hermetic)` **success**, `Shellcheck` success, `Installer Integration Tests` success). Labels moved `loom:changes-requested` + `loom:ci-failure` → `loom:review-requested`.

## What it took (two commits after the rebase)

**`c1117995` — the hang.** Root cause was the hermeticity gap the Judge hypothesized, with one more link in the chain than expected:

`LOOM_DAEMON_LAUNCHD=0` had no Linux counterpart and `MINIMAL_PATH` leaves the runner's real `/usr/bin/systemctl` reachable → scenario 5's real start script enabled the **default-named** `loom-daemon.service` (+ a real `loom-daemon-watchdog.timer`) → every later scenario then saw `systemd_unit_loaded` == true, so the update script resolved `DAEMON_MANAGER=systemd` and ran `"$PROVISION_TARGET" restart` → `PROVISION_TARGET` is the fake-daemon fixture, which has no `restart` handler, so it fell into its `while true; do sleep 1; done` body **in the foreground** and never returned. Fixed with `LOOM_DAEMON_SYSTEMD=0` + a scratch `LOOM_SYSTEMD_UNIT` suite-wide (scenarios 26–34 opt back in with `LOOM_DAEMON_SYSTEMD=1`), plus a fixture catch-all so *any* unrecognized non-flag subcommand exits 1 instead of wedging, plus 5b logging to a file.

**`ad2c5c4f` — what the hang was hiding.** With the wedge gone the suite went 1200s/exit 124 → **24s/exit 1**, 118/120, exposing two genuinely Darwin-only assertions underneath. Scenarios 21 and 22 drive `--relaunch`, whose first step is `harvest_plist_env()` — which requires **`plutil`** by contract ("required on the macOS launchd path") and returns 2 without it, so `perform_relaunch()` aborts with 6 and never re-renders the plist; scenario 22 then reads the result back with `plutil -extract` directly. `plutil` is macOS-only with no ubuntu stand-in worth stubbing (an XML-plist parser stub would be testing the stub, not the production path). Every *other* launchd scenario here is already portable — they run off the stub `launchctl` + stub `uname` that `write_fake_launchd_loaded_bin` installs — so the guard is scoped to exactly those two and prints a visible `⊘ SKIP` line rather than silently passing.

## Verification (both platforms this time)

The lesson from the first round was that a green macOS run proves nothing about this suite, so I reproduced the CI environment locally in an `ubuntu:24.04` container before each push:

| | macOS | ubuntu:24.04 | CI (`ad2c5c4f`) |
|---|---|---|---|
| `test-loom-daemon-update.sh` | **120/120** | **118/118** + SKIP line, exit 0 | **PASS, 23s** |
| `test-bounded-run.sh` | **18/18** | **18/18** | **PASS, 2s** |

`check-ci-suite-manifest.sh` OK (94 suites: 91 wired, 3 excluded). `bash -n` + `shellcheck --severity=warning` clean on every touched file.

`test-spawn-generic.sh` — the failure the Judge correctly identified as pre-existing and not mine — **now passes**, picked up by the rebase onto current `main` (#4819), exactly as predicted.

## Notes for the re-review

- **`ci-wired.txt` move stands.** The fallback of reverting it back to `ci-excluded.txt` is no longer needed: the mechanism is understood, fixed at three independent layers, and confirmed green in the actual CI environment on this head SHA.
- **`test-bounded-run.sh` is new and wired** (18 assertions): stdout/exit-code forwarding, `124` on a never-returning command, the child reaped rather than orphaned — each against **both** implementations. `LOOM_FORCE_PORTABLE_TIMEOUT` is no longer dead code; a stub-`timeout(1)` seam check proves the bypass, with an inverse control.
- **Deferred by design:** de-duplicating `loom-daemon-watchdog.sh`'s inline `bounded_run` against the shared lib → filed as **#4832** (`loom:triage`). The drift risk you named is real; it just needed a production-script edit that contributed nothing to the blocker.
- **Worth knowing for future work on this suite:** it now carries a real ubuntu-portability contract. `docker run --rm -v <repo>:<repo> -w <worktree> ubuntu:24.04 bash -c 'apt-get install -y git jq && bash defaults/scripts/tests/test-loom-daemon-update.sh'` reproduces the CI environment in about a minute and is what turned an opaque 20-minute CI timeout into a 30-second diagnosis.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T21:52:27Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T21:52:27Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed in the existing builder worktree (`.loom/worktrees/issue-4799`), 3 commits on the branch.

## Correctness
- `bounded-run.sh` cleanly extracts the `bounded_run()` pattern from `loom-daemon-watchdog.sh` (#4398) with a `timeout`/`gtimeout`/portable-fallback chain, `-k 2` KILL escalation, and a `124` normalization on the portable path (`143|137` → `124`). `LOOM_FORCE_PORTABLE_TIMEOUT` seam makes the macOS-default fallback path testable everywhere.
- `print_calibrate_hint()` now sources the lib defensively (`[[ -r ... ]]`) and wraps the blocking `$(daemon calibrate ...)` in `bounded_run` with a validated numeric timeout (`LOOM_CALIBRATE_HINT_TIMEOUT_SECS`, default 5s). If the lib fails to source, `bounded_run` is undefined -> command-not-found (127) -> `|| return 0` degrades to a silent no-op, never a hang. Matches the described fix for the #4790/#4773 deferred-signal-in-command-substitution hazard.
- `test-loom-daemon-update.sh` gets real hermeticity fixes for CI (forcing `LOOM_DAEMON_SYSTEMD=0` off a real host `systemctl --user`, a scratch `LOOM_SYSTEMD_UNIT`, fake-daemon fixtures now reject unrecognized non-flag subcommands instead of falling into the foreground `while true` loop). The systemd-specific scenarios (26-34) correctly opt back in with `LOOM_DAEMON_SYSTEMD=1`.

## Test execution
- `bash -n` + `shellcheck --severity=warning` clean on all 4 touched/added files.
- `check-ci-suite-manifest.sh` -> OK: 94 suites (91 wired, 3 excluded).
- `test-bounded-run.sh` (new): **18/18 pass** (native + portable + seam-verification modes).
- `test-loom-daemon-update.sh`: **120/120 pass**, matching the PR's claimed count.
- `test-loom-daemon-start.sh` (other CI-wired suite touching the modified script): **72/72 pass**, no regressions.
- CI (via REST, GraphQL was rate-limited): all 10 checks green (Rust jobs correctly `skipped` -- no Rust files touched). `mergeable_state: clean`.

## Issue linking
- `Closes #4799` present; no stray close-keyword targeting any other issue.

Nice, well-documented fix with a solid repro-driven test for the underlying hazard.

PR #4808, opened 2026-07-31
fix(mcp): heartbeat/list_terminals reflect the live daemon, not the retired desktop-app log

### LINKED ISSUE #4794: MCP get_heartbeat/list_terminals report stale desktop-app state instead of the running daemon
## Context

Hit during a klayout-tools working session (2026-07-31), on a host where
the Rust `loom-daemon` was healthy under launchd
(`com.rjwalters.loom-daemon`, 8 in-flight sweeps, 11 managed repos):

- `mcp__loom__get_heartbeat` returned `status: "stale"` with
  `"Last log entry was 284857m ago"` (~6.5 months) — the log entries it
  reads are from the retired Electron UI (`component: "main"`,
  "Found browse button element", last entry 2026-01-14), not from the
  running daemon.
- `mcp__loom__list_terminals` returned one orphaned entry:
  `ID: test / Restored: loom-test-85b41de2…-default-0` — a leftover test
  session presented as the only "active" terminal.

So the MCP observability surface says "stale / one test terminal" while
`loom-daemon status` (Unix socket) reports a live fleet. Any agent using
the MCP tools to answer "is the fleet running?" gets the wrong answer;
the correct answer required shelling out to `loom-daemon status`.

- [ ] `get_heartbeat` derives liveness from the daemon (Unix-socket
      `DaemonStatus` IPC, or the daemon's own log/pidfile), not from the
      retired desktop-app log; a healthy daemon never reports `stale`
- [ ] `list_terminals` garbage-collects or flags terminals whose backing
      tmux session no longer exists (the `loom` socket had no server)
- [ ] When the desktop-app log is absent/ancient but the daemon socket is
      live, heartbeat says so explicitly rather than reporting staleness

rjwalters:
## Summary

`mcp__loom__get_heartbeat` derived its entire status from the retired
Electron desktop app's `console.log`, so a perfectly healthy `loom-daemon`
with a stale/absent desktop log reported `status: "stale"` (observed:
`"Last log entry was 284857m ago"` on a host with 8 in-flight sweeps under a
healthy daemon). `mcp__loom__list_terminals` also reported terminals whose
backing tmux session had died (e.g. after the whole `loom` tmux socket lost
its server) as "active" forever.

- `get_heartbeat` now probes the daemon over its Unix socket first (`Ping`,
  then a best-effort `DaemonStatus` for the in-flight sweep count). A live
  daemon always yields `status: "daemon_healthy"`, regardless of desktop-log
  freshness. Only when the daemon itself is unreachable does it fall back to
  the desktop-log-derived status — with an explicit note that this reflects
  the retired desktop app's log only, not fleet health. The full response
  shape is preserved for backwards compatibility (`lastLogTime`, `logCount`,
  `recentLogs` at the top level, plus new `daemon` / `desktopApp` detail
  objects).
- `TerminalManager::list_terminals` (`loom-daemon/src/terminal.rs`) now
  garbage-collects registry entries whose `tmux_session` has no live backing
  session on every call (`prune_dead_terminals`), via a single `tmux -L loom
  list-sessions` call. Any failure to query tmux at all — including "no
  server running", the exact symptom from the issue — is treated as "zero
  live sessions", so the whole registry is correctly pruned rather than
  silently kept.
- The MCP `list_terminals` tool's rare state-file fallback path (used only
  when the daemon itself is unreachable) is NOT tmux-verified, so it is now
  flagged explicitly in the tool output (`buildTerminalListFallbackNote`)
  instead of being presented as a confirmed-live list.

## Test plan

- [x] `cargo test --lib terminal::` (loom-daemon) — 61 passed, including two
      new tests: a ghost terminal (fake `tmux_session`) is pruned from
      `list_terminals()` and removed from the registry; `LOOM_NO_RESTORE=1`
      suppresses pruning (matching its existing "skip all tmux interaction"
      contract used by other tests).
- [x] `cargo test --lib ipc::` — 81 passed (no regression in the existing
      `ListTerminals` IPC test).
- [x] `cargo clippy --package loom-daemon --lib -- -D warnings` — clean.
- [x] `cargo fmt --package loom-daemon -- --check` — clean.
- [x] `cd mcp-loom && npm run typecheck` — clean.
- [x] `cd mcp-loom && npm test` (vitest) — 51 passed, including new
      `classifyHeartbeat` tests (daemon-healthy wins over a stale/absent
      desktop log; falls back with an explicit note when the daemon is
      unreachable; preserves legacy top-level fields) and
      `buildTerminalListFallbackNote` tests.
- [x] `cd mcp-loom && npm run build` — bundle builds, `dist/index.js` present.

Closes #4794

### DIFF
--- a/loom-daemon/src/terminal.rs
+++ b/loom-daemon/src/terminal.rs
@@ -1684,7 +1684,69 @@ impl TerminalManager {
                 log::warn!("Failed to restore terminals from tmux: {e}");
             }
         }
+
+        // Issue #4794: garbage-collect registry entries whose backing tmux
+        // session no longer exists. Without this, a terminal that was
+        // registered by a since-crashed daemon, or whose tmux session was
+        // reaped externally (e.g. the whole `loom` tmux socket losing its
+        // server), sticks around in `self.terminals` forever and is reported
+        // to every `list_terminals` caller (including the MCP `list_terminals`
+        // tool) as an "active" terminal indefinitely.
+        if !no_restore {
+            self.prune_dead_terminals();
+        }
+
         self.terminals.values().cloned().collect()
+    }
+
+    /// Remove registry entries whose `tmux_session` no longer has a live tmux
+    /// session backing it (Issue #4794).
+    ///
+    /// A single `tmux -L loom list-sessions` call is used to build the live-session
+    /// set rather than one `has-session` call per terminal, so this stays cheap
+    /// even with many registered terminals. Any failure to query tmux at all
+    /// (including "no server running", the exact symptom reported in #4794)
+    /// is treated as "zero live sessions" — conservatively correct, since a
+    /// terminal cannot be alive without a session to back it.
+    ///
+    /// Skipped when the registry is empty (nothing to prune) so this never
+    /// pays for a `tmux` invocation on the common empty-registry path.
+    fn prune_dead_terminals(&mut self) {
+        if self.terminals.is_empty() {
+            return;
+        }
+
+        let output = Command::new("tmux")
+            .args(["-L", "loom"])
+            .args(["list-sessions", "-F", "#{session_name}"])
+            .output();
+
+        let live_sessions: std::collections::HashSet<String> = match output {
+            Ok(out) if out.status.success() => String::from_utf8_lossy(&out.stdout)
+                .lines()
+                .map(str::to_string)
+                .collect(),
+            // Any other outcome — nonzero exit (e.g. "no server running", the
+            // #4794 symptom), or the `tmux` command failing to spawn at all —
+            // means there are no live tmux sessions right now.
+            _ => std::collections::HashSet::new(),
+        };
+
+        let dead_ids: Vec<TerminalId> = self
+            .terminals
+            .iter()
+            .filter(|(_, info)| !live_sessions.contains(&info.tmux_session))
+            .map(|(id, _)| id.clone())
+            .collect();
+
+        for id in dead_ids {
+            if let Some(info) = self.terminals.remove(&id) {
+                log::info!(
+                    "Pruning dead terminal registry entry '{id}': backing tmux session '{}' no longer exists",
+                    info.tmux_session
+                );
+            }
+        }
     }
 
     pub fn set_worktree_path(&mut self, id: &TerminalId, worktree_path: &str) -> Result<()> {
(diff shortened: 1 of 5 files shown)


rjwalters:
LGTM — approving.

**What I verified independently** (worktree `.loom/worktrees/issue-4794`, up to date with `origin/main`):

- `cargo test --lib terminal::` → 61 passed, including the two new `prune_dead_terminals` tests (ghost-session pruning + `LOOM_NO_RESTORE=1` skip).
- `cargo test --lib ipc::` → 81 passed.
- `cargo clippy --package loom-daemon --lib -- -D warnings` → clean.
- `cargo fmt --package loom-daemon -- --check` → clean.
- `npm run typecheck` (mcp-loom) → clean.
- `npm test` (mcp-loom) → 51 passed (5 files), including new `classifyHeartbeat` and `buildTerminalListFallbackNote` suites.
- `npm run build` → succeeds, `dist/index.js` produced.
- CI: all required checks green (`gh pr checks 4808`), `mergeStateStatus: CLEAN`.

**Code review notes:**
- `TerminalManager::prune_dead_terminals` (loom-daemon/src/terminal.rs) is a sound fix: single `tmux -L loom list-sessions` call per `list_terminals()` invocation, any query failure conservatively treated as "zero live sessions" (matches the exact #4794 symptom — no tmux server at all), correctly gated behind `LOOM_NO_RESTORE` to preserve the existing "skip tmux interaction in tests" contract.
- `classifyHeartbeat`/`probeDaemon` (mcp-loom/src/tools/ui.ts) is a clean pure-function extraction — `Ping`→`Pong` and `DaemonStatus`→`payload.in_flight` match the daemon's actual `Response` wire shape (`#[serde(tag = "type", content = "payload")]`), `DaemonStatus` failure is correctly best-effort (a live `Ping` still reports `daemon_healthy` even if the richer status call times out), and all legacy top-level fields (`lastLogTime`/`logCount`/`recentLogs`) are preserved for backwards compatibility.
- `listTerminalsWithSource`/`buildTerminalListFallbackNote` (mcp-loom/src/tools/terminals.ts) correctly distinguishes the tmux-verified daemon path from the unverified state-file fallback, and the existing `listTerminals()` call sites are preserved via a thin wrapper — no call-site churn.
- All three issue #4794 acceptance criteria are directly addressed, and the new unit tests target the actual regression (stale desktop log winning over a live daemon; ghost tmux sessions surviving forever) rather than just exercising the happy path.

No blockers found. Nice, focused fix.

rjwalters:

- Merge Conflict Check: mergeable state is UNKNOWN (GitHub still calculating) after repeated checks

**Next steps:**
- No action needed if this resolves on its own — GitHub sometimes takes a few minutes to compute mergeability on a fresh push
- If it does not resolve, check for an actual conflict against `main`

---
*Automated by Champion role*

PR #4802, opened 2026-07-31
feat(runtimes): tier-3 generic passthrough adapter for unverified CLI coverage

rjwalters:
## Summary

Adds a **tier-3 "generic passthrough"** classification to Loom's runtime
adapter contract, alongside tier-1 (Claude, default) and tier-2 (Codex,
CI-gated). Tier-3 is a bounded, explicitly lower-trust escape hatch for
trying an unverified CLI against a **read-only** role (Judge, Curator,
Guide, Auditor) without first writing a full guardrail-parity document.

- **`defaults/scripts/spawn-generic.sh`** — a shared, env-var-driven template
  satisfying only contract point 1 (Spawn) and the generic-transient
  fallthrough half of contract point 3 (Error classification). No model
  mapping, no usage accounting, no sandbox mapping, no bespoke
  `classify-error.sh` pattern table.
- **`defaults/runtimes/aider.json`**` — the tier-3 capability manifest,
  declaring every capability `"no"`, including `worktreeIsolation: "no"`
  **explicitly** (not `"partial"`) — there is no residual gap to record in
  prose because there is no parity doc for tier-3 at all.
- **`defaults/scripts/spawn-aider.sh`** — the worked example: a thin wrapper
  pinning the [Aider](https://aider.chat) CLI's binary name and
  non-interactive prompt flag, then `exec`ing the shared template.
- **`defaults/docs/runtime-adapters.md`** — documents the tier-3 shape: a new
  tier-policy bullet, an adapter-status table row, a full "Tier 3: generic
  passthrough" section, and cross-references from contract points 6/7 and
  the "Adding a runtime adapter" walkthrough.

**No enforcement code changed.** `check-runtime-capabilities.sh` already
treats any capability value other than exactly `"yes"` as unmet, so a
manifest declaring `worktreeIsolation: "no"` fails closed for Builder/Doctor
identically to how Codex's `"partial"` already does — this is the same
mechanism, not a new one. `spawn-worker.sh` also needed no change: it
already resolves `spawn-<runtime>.sh` by name off disk, so
`LOOM_RUNTIME=aider` reaches `spawn-aider.sh` → `spawn-generic.sh` through
the exact seam Codex uses.

One nuance worth calling out: `judge.json` requires `mcp`, so a tier-3
runtime with no MCP support (the honest default) is refused for Judge too —
it is admitted only for roles that declare no `runtimeRequirements` at all
(`curator.json`, `guide.json`, `auditor.json` today). The doc documents this
[...]

PR #4811, opened 2026-07-31
fix(tests): bg-proc-trap.sh polish — empty-PID no-op + guarded pkill backstop

### LINKED ISSUE #4800: tests: bg-proc-trap.sh polish — empty-PID return code under set -e + pkill -f "" degenerate when mktemp fails (PR #4790 judge nits)
## Context
Two non-blocking nits from the PR #4790 judge on the newly-merged shared helper `defaults/scripts/tests/lib/bg-proc-trap.sh`:
1. `bg_proc_track` returns 1 when called with an empty PID (tolerated today because current consumers don't run `set -e` through it, but the helper is shared — a future `set -e` suite would abort on a benign empty capture)
2. `test-loom-daemon-start.sh`'s new `pkill -f "$WORKDIR"` backstop degenerates to `pkill -f ""` (matches everything!) if `mktemp -d` ever fails and WORKDIR is empty — needs a non-empty guard

- [ ] `bg_proc_track` returns 0 on empty/absent PID (documented as a tolerated no-op)
- [ ] The pkill backstop guards `[ -n "$WORKDIR" ]` (audit the other 5 suites for the same shape)
- [ ] Helper's header comment documents both contracts; suites re-run green

rjwalters:
## Summary
Two non-blocking nits from the PR #4790 judge on the shared test helper `defaults/scripts/tests/lib/bg-proc-trap.sh`:

1. `bg_proc_track` now always returns `0`, even when called with an empty/absent PID — previously it returned `1` in that case (a benign no-op, tolerated today only because current consumers don't run `set -e` through it, but the helper is shared).
2. `test-loom-daemon-start.sh`'s and `test-loom-daemon-update.sh`'s `pkill -f "$WORKDIR"` / `pkill -f "$BASE_WORKDIR"` backstops now guard `[ -n "$WORKDIR" ]` / `[ -n "$BASE_WORKDIR" ]` first, so a failed `mktemp -d` (leaving the var empty) can never degenerate into `pkill -f ""`, which matches (and would kill) every process visible to the caller.

Audited the other 5 suites that source `bg-proc-trap.sh` (`test-fleet-send.sh`, `test-fleet-check.sh`, `test-loom-daemon-stop.sh`, `test-loom-daemon-watchdog.sh`, `test-loom-daemon-launchd-plist.sh`) — none add a `pkill -f` backstop, only a plain `rm -rf "$WORKDIR"` (harmless no-op on an empty string), so no further changes were needed there.

Both contracts are now documented in `bg-proc-trap.sh`'s header comment.

## Test plan
- [x] `bash defaults/scripts/tests/test-loom-daemon-start.sh` — 72/72 passed
- [x] `bash defaults/scripts/tests/test-loom-daemon-stop.sh` — 37/37 passed
- [x] `bash -n` syntax check on all 3 modified files
- [x] `test-loom-daemon-update.sh` is documented (`ci-excluded.txt`) as non-hermetic/hangs past 120s locally — not run to completion, but `bash -n` confirms syntax validity and the change is a mechanical one-line guard addition mirroring the already-verified fix in `test-loom-daemon-start.sh`.

Closes #4800

### DIFF
--- a/defaults/scripts/tests/lib/bg-proc-trap.sh
+++ b/defaults/scripts/tests/lib/bg-proc-trap.sh
@@ -47,6 +47,19 @@
 # a single combined trap would clean up once and then keep executing every
 # remaining test case in the suite, re-populating $WORKDIR as it goes. The
 # INT/TERM trap must end with an explicit `exit` to actually stop the script.
+#
+# Two documented contracts (PR #4790 judge nits, #4800):
+#   1. bg_proc_track ALWAYS returns 0, even when called with an empty/absent
+#      PID (e.g. a failed spawn produced no `$!`). This is a tolerated no-op,
+#      not an error, so the call is safe to make under `set -e` — a future
+#      `set -e` suite sourcing this helper will not abort on a benign empty
+#      capture.
+#   2. A few suites add a path-pattern `pkill -f "$WORKDIR"` backstop (on top
+#      of bg_proc_reap) to catch fixture processes whose argv references a
+#      scratch directory rather than a tracked PID. If `mktemp -d` ever fails,
+#      $WORKDIR is empty and `pkill -f ""` matches (and kills) EVERY process
+#      visible to the caller — any suite adding such a backstop MUST guard it
+#      with `[ -n "$WORKDIR" ] && pkill -f "$WORKDIR" ...` first.
 
 # Array of PIDs to kill on cleanup. Declared at source time so every sourcing
 # script (and any function it defines) shares one array without redeclaring
@@ -54,11 +67,13 @@
 BG_PROC_TRACKED_PIDS=()
 
 # Record a backgrounded PID for later reaping. Call immediately after `&`,
-# e.g. `some_cmd & bg_proc_track "$!"`. Silently ignores an empty PID so it
-# is always safe to call even if the spawn failed to produce one.
+# e.g. `some_cmd & bg_proc_track "$!"`. Contract: an empty/absent PID is a
+# tolerated no-op that always returns 0 (success) — never 1 — so it is safe
+# to call under `set -e` even if the spawn failed to produce a PID.
 bg_proc_track() {
     local pid="$1"
     [[ -n "$pid" ]] && BG_PROC_TRACKED_PIDS+=("$pid")
+    return 0
 }
 
 # Kill every tracked PID that is still alive. Idempotent and safe to call
--- a/defaults/scripts/tests/test-loom-daemon-start.sh
+++ b/defaults/scripts/tests/test-loom-daemon-start.sh
@@ -62,8 +62,8 @@ WORKDIR="$(mktemp -d)"
 # the script (only an EXIT-trap firing auto-exits) -- the explicit `exit`
 # below is required, else a SIGTERM'd suite would clean up once and then keep
 # running every remaining test case (re-populating $WORKDIR as it goes).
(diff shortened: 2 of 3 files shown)


rjwalters:
LGTM — approving.

Verified against the #4800 acceptance criteria:
- `bg_proc_track` now explicitly `return 0`, tolerating an empty/absent PID as a no-op (safe under `set -e`). Confirmed via read + local diff.
- `pkill -f "$WORKDIR"` / `pkill -f "$BASE_WORKDIR"` in `test-loom-daemon-start.sh` and `test-loom-daemon-update.sh` are now guarded with `[ -n "$WORKDIR" ] &&` / `[ -n "$BASE_WORKDIR" ] &&` before running, so a failed `mktemp -d` can no longer degenerate into a match-everything `pkill -f ""`. `rm -rf` on the same (possibly empty) var remains outside the guard, which is fine — `rm -rf ""` is a safe no-op.
- The header comment in `bg-proc-trap.sh` documents both contracts clearly, with the right cross-references (#4790, #4800).
- Audited the other suites that source `bg-proc-trap.sh` (`test-fleet-send.sh`, `test-fleet-check.sh`, `test-loom-daemon-stop.sh`, `test-loom-daemon-watchdog.sh`) myself — confirmed none add a `pkill -f` backstop, matching the PR body's conclusion. One tiny nit: the PR body's audit list also names `test-loom-daemon-launchd-plist.sh`, but that suite doesn't actually source `bg-proc-trap.sh` at all (its own comment says so explicitly, #4773) — harmless since the "no pkill -f backstop" conclusion still holds, just a minor inaccuracy in the write-up, not the code.

Testing:
- Ran `test-loom-daemon-start.sh` locally on the PR branch: 72/72 passed.
- Ran `test-loom-daemon-stop.sh` locally: 37/37 passed.
- `bash -n` clean on all 3 changed files; `shellcheck` clean (one pre-existing SC2001 style note at line 312 of `test-loom-daemon-start.sh`, unrelated to this diff).
- `gh pr checks`: all checks pass except "Shell Test Suites (hermetic)", which fails on `test-spawn-generic.sh` (exit 78). I reproduced this independently against `origin/main` tip (unrelated worktree, no PR changes) and it fails identically there — this is the pre-existing, already-filed, non-required-check flake tracked in #4815/#4814 (introduced by #4802's merge, nothing to do with this PR's `bg-proc-trap.sh`/daemon-start/daemon-update changes). Confirmed via the repo's branch ruleset (`gh api repos/rjwalters/loom/rulesets/8809610`) that there is no `required_status_checks` rule, so this does not block merge. `mergeStateStatus` is `UNSTABLE` purely because of that same non-required check; `mergeable: MERGEABLE`.

Small, well-scoped, mechanical fix that fully satisfies the #4800 acceptance criteria. No functional concerns.

rjwalters:
Code review complete — small, well-scoped fix (bg-proc-trap.sh empty-PID no-op + guarded pkill backstops), matches the stated test plan (72/72 + 37/37 passing locally, verified). CI initially failed on `Shell Test Suites (hermetic)` in `test-spawn-generic.sh`, a suite this PR does not touch or source (it doesn't source bg-proc-trap.sh) and which passes cleanly locally against this branch — looked like a CI flake, so I triggered a rerun of the failed job. Releasing the claim and skipping while it completes; a later pass will finalize the verdict once CI settles.

PR #4813, opened 2026-07-31
feat(dashboard): serve public view at / with Access-JWT sign-in fallback

### LINKED ISSUE #4795: [Epic #4702] Single-URL dashboard: root serves public view to anonymous visitors with OAuth sign-in fallback for 2amlogic.com users (worker JWT validation + Access re-plumb)
## Context
Part of Epic #4702. Operator UX correction (2026-07-31): the deployed layout splits the dashboard into a hard-gated root (`/` → Access SSO wall, dead-ends non-org visitors) and a separate open `/public` path. The intended experience is a **single `dashboard.2amlogic.com`**: org users (email domain 2amlogic.com, per the existing Access policy) authenticate via OAuth/SSO and see the full dashboard; everyone else **falls back to the public redacted view at the same URL** — no dead end, no second URL to remember.

Cloudflare Access cannot express "try auth, else serve public" (an Allow app always forces login), so this is a worker-side change plus an Access re-plumb:

## Design
1. **Worker root `/`**: validate the `CF_Authorization` cookie (Access JWT) against the team's public certs (`https://<team>.cloudflareaccess.com/cdn-cgi/access/certs`, standard Access JWT validation — cache the JWKs). Valid + allowed → serve the full authenticated dashboard. Absent/invalid → serve the public redacted view (same content as today's `/public`) with a visible "Sign in" affordance.
2. **`/login` (new, tiny)**: an Access-gated path whose only job is to mint the Access session (SSO bounce) and redirect to `/`. The "Sign in" button links here.
3. **Access apps re-plumb** (config, not code): root app `dashboard.2amlogic.com` switches from Allow-on-root to covering only `/login` (keep the existing 2amlogic.com-domain + [email redacted - 2AMLogic/2am#88] Allow policy and the scripted-admin service-token policy — moving the service-token policy to an `/admin` app is cleaner since scripted admin never touches /login); `/public` bypass app becomes redundant (path can 301 → `/` for bookmarks); `/ingest` + `/healthz` bypasses unchanged.
4. **`/admin/*`**: unchanged contract (ADMIN_TOKEN bearer + Access service token for scripted calls) — ensure the re-plumb keeps it reachable non-interactively.
5. **AUD pinning**: the worker validates the JWT `aud` against the Access app's AUD tag (exposed to the worker as a var/secret) so a token minted for another app on the zone can't unlock the dashboard.

## Acceptance criteria
- [ ] Anonymous request to `/` renders the redacted public view (200, no redirect) with a Sign in link
- [ ] `/login` → SSO → redirect to `/` renders the full view for an allowed identity (domain policy unchanged)
- [ ] A JWT with wrong aud / expired / garbage cookie falls back to public (never 500, never full view)
- [ ] `/ingest` + `/admin` behavior byte-identical (daemon push + scripted admin both keep working)
- [ ] `/public` 301s to `/`
- [ ] Access-app changes documented in dashboard/docs/cloudflare-access.md as the new reference layout (the split-path layout it currently teaches becomes a variant note)
- [ ] Deployed to the reference account and verified live (curl for the anon path; operator verifies the SSO path)

---

## Curator Enhancement (2026-07-31)

The Design/Acceptance-criteria structure above was already solid; this pass
adds the code-level pointers, a runtime-behavior gap the AC didn't spell out,
a required `## Test Plan` / `## Affected Files` section, and a complexity
marker. No scope changes to the Design or Acceptance Criteria above.

### Current-state findings (why this is `complex`, not `routine`)

- **`dashboard/src/index.ts`'s `fetch` handler currently has no JWT
  verification anywhere.** `GET /` (line ~400) just returns a static text
  banner; the `isAuthenticated` boolean elsewhere in this file (`/api/*` vs
  `/public/*`, lines ~379-396) is purely a function of *which route matched* —
  there is no in-Worker credential check to model the new root logic on. This
  issue introduces the Worker's first cryptographic-verification code path
  (JWKS fetch + cache, signature/`aud`/`iss`/expiry checks), which is exactly
  the kind of change where a subtle bug (accepting an unsigned/garbage token,
  a stale cached JWK, a fail-open branch) passes tests and review looking
  correct while being a real auth bypass — hence `complex`, not `routine`.
- **`dashboard/README.md` (lines 17-21) and `dashboard/docs/query-api.md`
  (lines 239-242) both currently assert "Worker-side Cloudflare Access JWT
  verification is out of scope for this repo — the Worker only decides *what
  a request sees* … never *whether it should have reached here*."** This
  issue makes that statement false for the root route specifically (the
  Worker now does make an admit/fallback decision based on JWT validity for
  `/`) — both docs need a correction alongside the `cloudflare-access.md`
  update the AC already calls out, or a Builder/Judge following stale docs
  could reject the whole approach as "out of scope."
- **`dashboard/docs/cloudflare-access.md` §5 ("Optional hardening: verify the
  Access JWT in the Worker") already sketches the exact ingredients this
  issue needs** — JWKS URL shape, expected `aud` = the app's Application
  Audience tag, "verify signature, `aud`, `iss`, and expiry" — but is
  currently written as "not implemented today," a Phase-3 nice-to-have. This
  issue is that implementation; the doc's framing (and its route map in §2,
  and the pitfalls table in §6) needs updating to describe the re-plumbed
  single-URL layout, not just have one new paragraph appended.
- **No JWT-verification library is present in `dashboard/package.json`
  today.** Implementer must either add a small Workers-compatible dependency
  (e.g. `jose`, the library Cloudflare's own Access-JWT-verification examples
  use — pure JS, no Node built-ins, works in the Workers runtime) or
  hand-roll RS256 verification against `crypto.subtle` (the runtime already
  used in `src/auth.ts` for the ingest-key SHA-256 hashing). Either is
  reasonable; `jose` is the recommended default to avoid hand-rolled crypto
  parsing edge cases (a wrong assumption here again reads correct in review).
- **`/healthz` is referenced in this issue's Design (point 3) and in the
  reference deployment's live Access app list, but does not exist as a route
  in `dashboard/src/index.ts` today** — grep confirms no `healthz` string
  anywhere in `dashboard/src`. Treat "the `/ingest` + `/healthz` bypasses
  unchanged" as **config-only, no code required** for `/healthz` specifically
  (it is presumably a Cloudflare-level health check path with no Worker
  route, distinct from `/ingest` which the Worker does implement) — do not
  file a "missing /healthz route" gap against this issue; if the reference
  deployment's `/healthz` app turns out to need a Worker route after all,
  that is a separate, follow-up concern.
- **Related but non-blocking**: #4801 (recording the reference deployment's
  current Access layout) already defers to this issue — its body says
  verbatim "#4795 will re-plumb this to the single-URL fallback layout, so
  write this doc against whichever lands last (gate on #4795's state, not
  number)." **Absorbed**: no AC change needed here since #4801 already
  accounts for the ordering; #4801's author should simply write against
  whatever this issue ships. No action item added to this issue's AC, but
  worth a PR-description cross-reference (`Related: #4801`) so #4801's next
  pass picks up the final layout instead of a mid-flight one.
- **Duplicate check**: `check-duplicate.sh --include-merged-prs --issue 4795`
  found no `DUPLICATE_FOUND`; only the `RELATED_OPEN_WORK` hit on #4801
  addressed above.

### Implementation Guidance

- Extract the JWT-verification logic into its own module (e.g.
  `dashboard/src/accessAuth.ts`) rather than growing `src/auth.ts` — the
  existing `auth.ts` module doc frames it narrowly around per-host
  *ingest-key* auth (SHA-256 hash comparisons against D1), which is a
  different trust boundary and a different failure mode than validating a
  third-party-signed Access JWT. Keep them adjacent but distinct, mirroring
  how `redaction.ts` stays separate from `query.ts` today.
- Cache the JWKS fetch (e.g. in-memory per-isolate with a short TTL, or via
  the Workers `caches` API) — the AC's "never 500" requirement means a JWKS
  fetch failure must fail closed to the public view, not throw.
- New config surface needed: a team-domain value (e.g.
  `CF_ACCESS_TEAM_DOMAIN`, safe as a plain `[vars]` entry in
  `wrangler.toml` — not secret) and an expected-`aud` value (the design's
  "AUD pinning" — can be a plain var too, since the AUD tag is not a secret,
  it is displayed on the Access app's overview page). Add both to
  `wrangler.toml`'s `[vars]` block (alongside `RETENTION_DAYS`/`MAX_RECORDS`)
  and to `.dev.vars.example`/`.dev.vars` for local `wrangler dev` parity, per
  the existing secrets-vs-vars convention documented in `wrangler.toml`'s
  comments.
- `GET /public` 301 → `/`: keep this as a one-line route change in the
  existing `fetch` handler (`dashboard/src/index.ts`, the `if
  (url.pathname === "/public")` branch becomes a redirect instead of calling
  `handlePublicPage`) — the actual page-rendering logic in
  `dashboard/src/publicPage.ts` is reused unchanged for the new `/` fallback
  path, it is only the *routing* that moves.
- The reference deployment (worker `loom-fleet-dashboard`, custom domain
  `dashboard.2amlogic.com`, account referenced in #4801) is where "deployed
  to the reference account and verified live" (last AC box) should target;
  the account's Cloudflare credentials are machine-local per #4801, not part
  of this repo.

## Affected Files

- `dashboard/src/index.ts` — root `/` handler: branch on JWT validity
  (full dashboard vs. public fallback + Sign-in link); new `/login` route;
  `/public` becomes a 301 to `/` instead of calling `handlePublicPage`
  directly.
- `dashboard/src/auth.ts` and/or a new `dashboard/src/accessAuth.ts` — Access
  JWT validation (JWKS fetch/cache, signature/`aud`/`iss`/expiry checks); see
  Implementation Guidance above for why a separate module is recommended.
- `dashboard/package.json` — add a JWT-verification dependency (e.g. `jose`)
  if not hand-rolling verification against `crypto.subtle`.
- `dashboard/wrangler.toml` — add `CF_ACCESS_TEAM_DOMAIN` / expected-`aud`
  `[vars]` entries.
- `dashboard/.dev.vars.example` — matching local-dev placeholders for any new
  var/secret.
- `dashboard/test/auth.test.ts` (or a new `dashboard/test/accessAuth.test.ts`)
  — unit tests for the new validation function: valid token, wrong `aud`,
  expired token, garbage/absent cookie, unreachable JWKS endpoint (each must
  fall back to public, never throw/500).
- `dashboard/docs/cloudflare-access.md` — rewrite §2 (route map: root app
  now covers only `/login`; `/public` bypass app becomes redundant/removable;
  `/admin` gets its own app for the service-token policy), §3
  (step-by-step Access console instructions for the new layout), §5 (this is
  now implemented, not "optional hardening… not implemented today"), and §6
  (pitfalls table) to describe the single-URL fallback layout as the primary
  reference layout, with the old split-path layout demoted to a variant note
  (per this issue's own AC).
- `dashboard/README.md` (lines 17-21) — the "Out of scope for this repo:
  Worker-side Cloudflare Access JWT verification" paragraph is no longer
  accurate for the root route once this ships; update or scope it precisely
  (e.g. "the `/api/*` vs `/public/*` split still relies entirely on the edge;
  only the `/` root fallback does in-Worker verification").
- `dashboard/docs/query-api.md` (lines 239-242) — same correction as
  README.md above; this doc currently makes the identical "out of scope"
  claim.
- `dashboard/docs/deploy-runbook.md` (line ~160) — the example `curl`
  transcript shows the old static-text root response
  (`loom-observability-ingest: see /ingest, /admin/*`); update once `/`
  serves the actual dashboard/fallback content.

## Test Plan

- [ ] Automated tests: new unit tests (in `dashboard/test/auth.test.ts` or a
  new `accessAuth.test.ts`) covering: valid signed JWT + correct `aud` →
  accepted; wrong `aud` → rejected; expired token → rejected; malformed/
  garbage cookie → rejected; missing cookie → rejected; JWKS fetch failure →
  fails closed (rejected, not thrown) — mirror the existing
  `describe`/`it` structure already used in `auth.test.ts` (see
  `extractBearerToken`/`hashIngestKey`/`authenticateHost` blocks).
- [ ] Automated tests: route-level test that `GET /public` returns a 301 to
  `/` (extend the existing route tests alongside `ingest.test.ts`/
  `publicPage.test.ts` conventions).
- [ ] Manual verification (per the issue's own AC, against the reference
  deployment `dashboard.2amlogic.com`):
  - `curl -sS https://dashboard.2amlogic.com/` → 200, redacted public HTML
    with a visible Sign in link (no redirect).
  - Operator (2amlogic.com identity) clicks Sign in → `/login` → SSO →
    redirected to `/` → full authenticated dashboard renders.
  - `curl -sS -o /dev/null -w '%{http_code}\n' https://dashboard.2amlogic.com/public`
    → 301, `Location: https://dashboard.2amlogic.com/`.
  - Existing daemon `/ingest` push still succeeds (no regression from the
    Access re-plumb) — restart a daemon and confirm new rows land (same
    check `docs/cloudflare-access.md` §4 already documents).
  - Scripted `/admin/hosts` call (service-token + `ADMIN_TOKEN`) still
    succeeds non-interactively after the re-plumb.
- [ ] Edge cases: a JWT signed for a *different* Access app on the same zone
  (wrong `aud`) must fall back to public, not error and not grant access —
  this is the specific "aud pinning" requirement from Design point 5 and is
  easy to get subtly wrong (e.g. checking `iss` but forgetting `aud`).

rjwalters:
## Summary

Cloudflare Access cannot express "try to authenticate, otherwise serve
something else" — an Allow application always forces a login. So the deployed
layout split the dashboard into a hard-gated root (`/` → SSO wall, dead-ending
non-org visitors) and a separate open `/public` path. This moves the decision
into the Worker: **one URL**, where an allowed identity sees the full dashboard
and everyone else falls back to the redacted public view with a Sign in link.

This is the Worker's first cryptographic-verification code path, so the
load-bearing property is **fail-closed**: every failure mode — no cookie,
malformed/garbage token, wrong `aud`, wrong `iss`, expired, forged signature,
`alg: none`, alg confusion, unreachable JWKS — renders the public view. Never a
500, never the full view on a doubtful token.

Related: #4801 (reference-deployment Access layout doc) — write it against this
layout, which is now the primary one documented in `cloudflare-access.md`.

## Changes

- **`dashboard/src/accessAuth.ts` (new)** — validates the `CF_Authorization`
  cookie against `https://<team>/cdn-cgi/access/certs` via `jose`, pinning
  `iss`, `alg` (RS256 only), and `aud`. Returns an identity or `null`; never
  throws. JWKS resolver cached per team domain per isolate. Kept separate from
  `src/auth.ts` (per-host ingest-key hashing — a different trust boundary).
- **`GET /`** — full unredacted dashboard for a validated identity, redacted
  public page with a Sign in link otherwise. 200 either way, no redirect.
  `Cache-Control: private, no-store` + `Vary: Cookie` so no shared cache can
  replay the authenticated body to an anonymous visitor.
- **`GET /login`** — 302 back to `/` (Access does the real work at the edge).
  **`GET /public`** — 301 to `/` for old bookmarks.
- **`dashboard/src/publicPage.ts`** — renders either variant from the same
  code; tails `/api/events` when authenticated, `/public/events` when not.
  Redaction still happens entirely in `redaction.ts` at the caller.
- **`CF_ACCESS_TEAM_DOMAIN` / `CF_ACCESS_AUD`** — new non-secret `[vars]`.
  Both unset ⇒ `/` is always public (fails closed).
- **Docs** — `cloudflare-access.md` rewritten around the single-URL layout
  (§2/§2a/§3b–§3e/§4/§5/§6), split-path layout demoted to a variant note;
  `README.md`, `docs/query-api.md`, `docs/deploy-runbook.md` corrected where
  they claimed in-Worker Access JWT verification was out of scope or showed the
  old static-text root response.

`/ingest` and `/admin/*` handlers are untouched.

### Two security findings the design surfaced (worth a reviewer's attention)

1. **Narrowing the root Access app to `/login` alone would leave `/api/*`
   matched by no Access application** — i.e. the full unredacted query API
   would become world-readable. The re-plumb therefore gives `/api/*` its own
   Allow application (§3e), with a pitfalls-table row and a `query-api.md`
   cross-reference so nobody replays the naive version of the config change.
2. **`CF_ACCESS_AUD` accepts a comma-separated allowlist.** Access names the
   session cookie `CF_Authorization` per *hostname* but mints the token per
   *application*, so once the authenticated page opens its `/api/events` live
   tail the browser can be holding the `/api/*` app's token. Pinning to the
   `/login` AUD alone would flap a just-signed-in operator back to the public
   view. This is still a pin — only enumerated tags are accepted, and a
   blank/whitespace-only value is treated as *unconfigured* (public view), never
   as "accept anything".

- [x] **Anonymous `/` renders the redacted public view (200, no redirect) with a
  Sign in link** — `test/publicPage.test.ts`: 200 + `text/html`, `href="/login"`
  present, private repo/sweep-id/account identifiers absent.
- [x] **`/login` → SSO → redirect to `/` renders the full view for an allowed
  identity** — `/login` 302→`/` covered by test; the authenticated render is
  covered end to end in `test/index.test.ts` (valid token ⇒ "Dashboard" heading,
  no Sign in link, `/api/events` feed, and the private-repo fixture *visible*,
  proving the branch really serves unredacted data). The live SSO bounce itself
  is edge config — operator verification, see below.
- [x] **Wrong aud / expired / garbage cookie falls back to public (never 500,
  never full view)** — covered at both levels: unit (`test/accessAuth.test.ts`)
  and route (`test/index.test.ts`), plus forged-signature, `alg: none`,
  alg-confusion, unreachable-JWKS, and unconfigured-vars cases.
- [x] **`/ingest` + `/admin` behavior byte-identical** — no handler changes;
  existing `ingest.test.ts` / `admin.test.ts` suites pass unmodified.
- [x] **`/public` 301s to `/`** — asserted with `redirect: "manual"`, including
  the exact `Location`.
- [x] **Access-app changes documented in `dashboard/docs/cloudflare-access.md`
  as the new reference layout** — single-URL layout is §2 (primary); split-path
  layout is §2a (variant note).
- [ ] **Deployed to the reference account and verified live** — *operator
  action, deliberately not done from this PR.* The reference deployment's
  Cloudflare credentials are machine-local (per #4801) and this code is not yet
  reviewed/merged; deploying unreviewed auth code to a live deployment is the
  wrong order. Post-merge checklist is in `cloudflare-access.md` §4 (`/` ⇒ 200
  not 302, `/public` ⇒ 301, `/ingest` ⇒ 401 not 302, plus the operator's own SSO
  round trip).

## Test Plan

```
$ npx vitest run test/accessAuth.test.ts test/index.test.ts test/publicPage.test.ts
 ✓ test/index.test.ts (6 tests)
 ✓ test/publicPage.test.ts (21 tests)
 ✓ test/accessAuth.test.ts (24 tests)
 Test Files  3 passed (3)
      Tests  51 passed (51)

$ npm run typecheck      # tsc --noEmit — clean
$ npx wrangler deploy --dry-run
 Total Upload: 81.50 KiB / gzip: 20.46 KiB    # jose bundles for workerd fine
```

**Pre-existing failures, not caused by this PR**: `npm test` at the dashboard
root also picks up `dashboard/web/test/*` (a separate package with its own
jsdom vitest config), where 19 tests fail with `ReferenceError: document is not
defined`. Verified identical on unmodified `main` (`npx vitest run
web/test/outcomesChartView.test.ts` fails the same way there). Backend suite
totals went 186 → 192 passing with the same 19 pre-existing failures.

`npm run preflight` reports its usual template-placeholder error
(`database_id` is the committed all-zeros placeholder) — also pre-existing and
by design for the template; the `wrangler deploy --dry-run` step within it
passes.

Closes #4795

### DIFF
--- a/dashboard/.dev.vars.example
+++ b/dashboard/.dev.vars.example
@@ -10,3 +10,12 @@
 # Bearer token gating every /admin/* route. While unset, those routes answer
 # 503. Generate a real one with: openssl rand -hex 32
 ADMIN_TOKEN="local-dev-admin-token"
+
+# NOTE: CF_ACCESS_TEAM_DOMAIN / CF_ACCESS_AUD (single-URL dashboard root —
+# see docs/cloudflare-access.md) are NOT secrets, so their home is the
+# `[vars]` block in wrangler.toml (which `wrangler dev` already reads),
+# alongside RETENTION_DAYS/MAX_RECORDS — not this file. Uncomment them there
+# to exercise the authenticated `/` branch locally. (`.dev.vars` does
+# override `[vars]` in dev, so putting them here would also work; keeping
+# non-secrets out of the secrets template is simply the convention this repo
+# follows.)
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -14,11 +14,16 @@ Query API reference: [`docs/query-api.md`](docs/query-api.md) — includes the
 `/api/*` (authenticated) vs. `/public/*` (redacted) route split and the
 visibility-based redaction policy (issue #4727).
 
-**Out of scope for this repo**: Worker-side Cloudflare Access JWT
-verification — the [Access guide](docs/cloudflare-access.md) covers gating
-`/api/*` via the edge proxy instead; the Worker only decides *what a request
-sees* (via the route it reached), never *whether it should have reached
-here*.
+**`/api/*` vs. `/public/*` remains a route-based auth split, not an
+in-Worker one**: the [Access guide](docs/cloudflare-access.md) covers gating
+`/api/*` via the edge proxy; the Worker decides what a `/api/*`/`/public/*`
+request sees purely by which route matched, never by verifying a credential
+itself. **The one exception is the dashboard root `/`** (issue
+[#4795](https://github.com/rjwalters/loom/issues/4795)): it validates the
+visitor's Cloudflare Access JWT in-Worker (`src/accessAuth.ts`) so a single
+URL can serve both an anonymous public view and, for an allowed identity, the
+full dashboard — see the Access guide's §5 for exactly what that check
+covers and its fail-closed contract.
 
 ## Architecture
 
@@ -124,12 +129,15 @@ asserts the `ADMIN_TOKEN` secret exists on the deployed Worker.
 | `GET /public/fleet-state` | none (always public) | Public query API: same data as `/api/fleet-state`, redacted per record `visibility` (issue #4727). |
 | `GET /public/history` | none (always public) | Public query API: same data as `/api/history`, redacted. |
 | `GET /public/events` | none (always public) | Public query API: same data as `/api/events`, redacted. |
-| `GET /public` | none (always public) | The public dashboard page itself (issue #4753) — server-rendered fleet overview + history from the same redacted data above, plus a live feed over `/public/events`. Read-only; never calls `/api/*`. |
+| `GET /` | **in-Worker** Access JWT check (`src/accessAuth.ts`) — the only route that verifies a credential itself | The dashboard page (issues #4753, #4795). A valid, correctly-audienced `CF_Authorization` cookie renders the full unredacted view (live feed over `/api/events`); **anything else** — no cookie, malformed/expired/wrong-`aud` token, JWKS fetch failure — falls back to the redacted public view with a Sign in link (live feed over `/public/events`). Never a 302, never a 500. Read-only. |
+| `GET /login` | none in-Worker — **must** be behind Cloudflare Access | Bare 302 back to `/`. Its whole purpose is to make Access run the SSO round trip and mint the `CF_Authorization` cookie that `/` then validates. |
+| `GET /public` | none (always public) | 301 → `/` (issue #4795 merged the public page into the root; kept for old bookmarks). |
 
 Full request/response shapes for the `/api/*` and `/public/*` routes,
 including the exact redaction policy per record kind:
-[`docs/query-api.md`](docs/query-api.md). Cloudflare Access setup to actually
-gate `/api/*` at the edge (this backend never verifies auth itself):
+[`docs/query-api.md`](docs/query-api.md). Cloudflare Access setup to gate
+`/login`, `/api/*`, and `/admin/*` at the edge — and to leave `/` ungated so
+its in-Worker fallback can run:
 [`docs/cloudflare-access.md`](docs/cloudflare-access.md).
 
 ## Design decisions worth knowing
--- a/dashboard/docs/cloudflare-access.md
+++ b/dashboard/docs/cloudflare-access.md
@@ -1,13 +1,21 @@
 # Gating the dashboard with Cloudflare Access
 
-How to put zero-trust SSO in front of the **authenticated** fleet view while
-leaving the **public** view and the machine-to-machine `/ingest` endpoint
-reachable without a login.
+How to wire zero-trust SSO into the **single-URL dashboard** (issue
+[#4795](https://github.com/rjwalters/loom/issues/4795)): one hostname,
+`dashboard.example.com`, where an anonymous visitor to `/` sees the
+redacted public view and an allowed identity sees the full view — no second
+URL, no dead-end login wall — plus the machine-to-machine `/ingest` endpoint
+and the `/admin/*` management routes, both reachable without a browser login.
 
-This is the Cloudflare-side configuration only. Phase 3 of epic
-[#4702](https://github.com/rjwalters/loom/issues/4702) builds the actual
-public-view page; the Access policies that distinguish the two routes are set
-up here, at the edge, and are what make that split enforceable.
+**This is a Worker-code-plus-config split, not config alone.** Cloudflare
+Access cannot itself express "try to authenticate, otherwise serve something
+else" — an Access **Allow** application always forces a login, full stop.
+The single-URL fallback works because `src/index.ts`'s root `/` handler does
+its own in-Worker Access-JWT check (`src/accessAuth.ts`) and falls back to
+the public view when it finds no valid one; Access itself only gates the
+narrow `/login` path that mints the session cookie the root handler reads.
+This doc covers the Cloudflare-side configuration that makes that split
+correct; the code side is `src/accessAuth.ts` and `src/index.ts`.
 
 Prerequisite: a deployed backend — see
 [`deploy-runbook.md`](deploy-runbook.md).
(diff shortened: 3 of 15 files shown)


rjwalters:
Code evaluation in progress; CI is still running (checks pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Approved. This is a careful, well-tested implementation of the Worker's first cryptographic JWT-verification path, and it holds up against adversarial review.

**Fail-closed correctness (`dashboard/src/accessAuth.ts`)**
- `validateAccessJwt` returns `null` on every failure path and never throws (single `try/catch` wrapping the entire `jwtVerify` call plus JWKS resolution).
- `algorithms: ["RS256"]` is explicitly pinned on `jwtVerify` — confirmed with dedicated tests for `alg: none` (hand-assembled unsigned JWS) and RS256→HS256 alg-confusion (HS256 header, RSA-public-key-bytes-as-MAC-secret). Both rejected.
- A well-formed token forged with an attacker-controlled RSA keypair (correct `iss`/`aud`/`exp`, wrong signature) is rejected — proves this is real cryptographic verification, not just claim-shape inspection.
- `parseAcceptedAudiences` correctly treats blank/whitespace/comma-only input as an empty list → `validateAccessJwt` short-circuits to `null` when `audiences.length === 0`. Verified this is NOT an accept-anything hole: it's "unconfigured → always public," and there's a test (`"fails closed when CF_ACCESS_AUD is present but blank"`) proving it.
- Unreachable/erroring JWKS resolver → `null`, not a 500 (tested at both the unit level and route level via a `globalThis.fetch` throw against the real `createRemoteJWKSet` wiring in `test/index.test.ts`).
- No `clockTolerance` set — expired is expired, with a comment specifically warning future maintainers not to add slack.

**AUD allowlist design** — the comma-separated `CF_ACCESS_AUD` is a real design necessity (one `CF_Authorization` cookie per hostname, but a token minted per Access application), not scope creep. It's an explicit enumerated allowlist passed to `jose`'s `audience` option (OR-match against the list), confirmed by a test that a tag *not* in the configured list is still rejected even when other tags are configured.

**Cache-control correctness** — `Cache-Control: private, no-store` + `Vary: Cookie` are present on both variants of `/`, and there's a dedicated route-level test (`"never lets a shared cache store either variant of / under one URL"`) asserting both headers on both the anonymous and authenticated response.

**Route scope** — confirmed via diff that `src/auth.ts` (ingest-key trust boundary) and `src/redaction.ts` are untouched; `/ingest` and `/admin/*` handlers have zero changes in `src/index.ts` beyond doc comments.

**Test suite exercises the fail-closed paths, not just happy path** — `test/accessAuth.test.ts` (24 tests) covers wrong aud, wrong iss, expired, garbage cookie, missing cookie, unreachable JWKS, unconfigured team-domain/AUD, blank AUD, forged signature, `alg: none`, alg confusion, and multi-aud allowlist behavior. `test/index.test.ts` (6 tests) re-exercises the security-critical subset end-to-end through the *production* `createRemoteJWKSet` wiring (not just the injectable test resolver), including the private-repo-data leak check (public variant never contains it, authenticated variant does) and the cache-header assertions.

**Doc accuracy** — checked `README.md`, `docs/query-api.md`, `docs/cloudflare-access.md`, `docs/deploy-runbook.md`: none claim in-Worker Access JWT verification is out of scope for `/`. `query-api.md`'s "Not implemented here" section correctly scopes the remaining gap to `/api/*`/`/public/*` only, with an explicit and well-reasoned rationale for why extending the same check there is deliberately not done (service-token callers with a different `aud`).

**Verification performed**
- `npx vitest run test/accessAuth.test.ts test/index.test.ts test/publicPage.test.ts` → 51/51 pass (matches PR description).
- `npm run typecheck` → clean.
- Full `npx vitest run test/` → all dashboard backend suites pass; the 19 `web/test/*` failures (`ReferenceError: document is not defined`) reproduce identically on unmodified `main`, confirmed pre-existing and unrelated to this PR.
- `gh pr checks` → all applicable checks pass (Rust/installer/MCP/shell suites correctly path-filtered as skipping; no dedicated dashboard CI workflow exists in this repo, so local test execution is the primary automated signal here — worth a future follow-up, not a blocker for this PR).
- `mergeStateStatus` → CLEAN.

No blocking issues found. Nice work threading the AUD-allowlist necessity and the `/api/*`-needs-its-own-Access-app pitfall through the docs — those are exactly the kind of subtle footguns this kind of change tends to leave behind.

PR #4816, opened 2026-07-31
ci: run mcp-loom vitest suite in CI

### LINKED ISSUE #4603: mcp-loom vitest suite (npm test) is never run in CI
## Observation

While reviewing PR #4543, noticed that `mcp-loom/package.json` defines a
`test` script (`vitest run`) and the package has 3 `*.test.ts` files
(`sweeps.test.ts` added by #4543, plus 2 pre-existing), but no CI workflow
ever invokes it.

`.github/workflows/ci.yml`'s "MCP Server Build" job only runs:

```
cd mcp-loom && npm install && npm run build   # tsc --noEmit + esbuild
test -f mcp-loom/dist/index.js
```

`npm run build` type-checks the test files (they're under `src/**` per
`tsconfig.json`'s `include`), so a type error would be caught, but the actual
assertions inside the tests never execute — a logic bug in `mcp-loom/src/**`
could ship with a passing CI run even though a test exists that would have
caught it.

## Suggested fix

Add a step (or job) to `.github/workflows/ci.yml`'s MCP Server Build job:
`cd mcp-loom && npm test`, gated the same way (`needs.changes.outputs.mcp`).

Discovered during code evaluation of PR #4543; not a defect in that PR
specifically — pre-existing gap in CI coverage for the whole `mcp-loom` package.

rjwalters:
## Summary
- The "MCP Server Build" CI job type-checked `mcp-loom/src/**` via `tsc --noEmit` (part of `npm run build`) but never ran `npm test`, so vitest assertions in the 3 `*.test.ts` files never executed in CI.
- Adds a `Run MCP server test suite` step (`cd mcp-loom && npm test`) after the existing build step, gated by the same job-level `if: needs.changes.outputs.mcp == 'true'` condition already applied to the whole job — no per-step gating needed since the job itself is gated.

## Test plan
- [x] `cd mcp-loom && npm install && npm test` locally — 5 test files, 51 tests, all passing.
- [x] `cd mcp-loom && npm run build` still succeeds and produces `dist/index.js`.
- [x] Diff limited to the single new step in `.github/workflows/ci.yml`.

Closes #4603

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,8 @@ jobs:
           node-version: "20"
       - name: Build unified MCP server
         run: cd mcp-loom && npm install && npm run build
+      - name: Run MCP server test suite
+        run: cd mcp-loom && npm test
       - name: Verify build output
         run: test -f mcp-loom/dist/index.js
 


rjwalters:
Code evaluation looks good; CI is still running (Rust Unit Tests, Shell Test Suites pending). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
Code evaluation looks good — small, well-scoped CI change (single new step gated by the existing job-level `if: needs.changes.outputs.mcp == 'true'` condition). Verified the new "Run MCP server test suite" step actually executed and passed in this PR's own MCP Server Build job run (job 91267109939). Locally: `npm test` — 3 test files, 41 tests, all passing (note: PR description's test plan claims 5 files/51 tests, which doesn't match current repo state — minor discrepancy, not blocking). Rust Unit Tests and Shell Test Suites are still pending. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
## Judge Review: APPROVE

**Diff**: Minimal, focused change — adds a single `Run MCP server test suite` step (`cd mcp-loom && npm test`) to the `mcp-build` job in `.github/workflows/ci.yml`, right after the existing build step.

**Gating**: Verified correct. The `mcp-build` job uses job-level gating (`if: always() && (github.event_name == 'push' || needs.changes.outputs.mcp == 'true'))`), same as its sibling jobs (`backend-tests`, etc.) — there is no per-step `if:` pattern in this job to match, so relying on the job-level condition is the right call and matches the PR description's stated rationale.

**CI verification**: Confirmed the new step actually ran in this PR's own CI run (since it touches `mcp-loom` + the workflow itself) and passed:
```
Test Files  5 passed (5)
     Tests  51 passed (51)
```
`MCP Server Build` job: pass (14s). `Verify build output` step after it also still passes, confirming `npm run build`/`dist/index.js` is unaffected.

**Other CI**: All other checks pass except `Shell Test Suites (hermetic)`, which fails on `test-spawn-generic.sh` (exit 78) — verified this is a pre-existing, unrelated failure reproducing identically on `main` HEAD in the two most recent main CI runs (independent of this PR), and is not part of any required status check in the branch ruleset. Not a blocker for this change.

Approving.

rjwalters:

- CI Status: `Shell Test Suites (hermetic)` failing (bucket=fail). Judge's review confirms this reproduces identically on `main` HEAD (failure in `test-spawn-generic.sh`, exit 78) and is not caused by this PR's diff, and is not a required status check in the branch ruleset — but Champion's CI-status criterion has no carve-out for pre-existing/unrelated failures.

**Next steps:**
- No action needed from Doctor on this PR's own diff — the failure is upstream/pre-existing on `main`
- Champion will retry automatically once `Shell Test Suites (hermetic)` is green again (fixed independently on `main`), or an operator can apply `loom:auto-merge-ok` to bypass this check for this PR

---
*Automated by Champion role*

PR #4833, opened 2026-07-31
fix(champion): key transient-rejection idempotency guard on stable identity, not freeform reason text

rjwalters:
## Summary

Fixes the residual duplicate-comment spam described in #4818: the transient-failure
idempotency guard in `champion-pr-merge.md` ("PR Rejection Workflow → Transient
failures") compared the freeform `$REASON` sentence verbatim (`grep -qF "$REASON"`)
against the most recently posted comment to decide whether to skip a duplicate
rejection comment. Because `$REASON` is composed fresh by the LLM on every Champion
tick, its wording drifts even when the underlying failure (e.g. the same failing CI
check) hasn't changed at all — PR #4796 collected 4 differently-worded "CI check X is
failing" comments within 42 minutes, all for the same never-changing check, because
none of the four exact sentences matched any prior one.

This is the same failure mode #4586/#4754 fixed (marker-based idempotency, `gh --jq
--arg` bug), just surviving through the freeform-text comparison instead.

## Change

- Introduce a `REASON_KEY`: a deterministic identity for the failure, built
  mechanically from the failure data (not LLM prose) — e.g. for `ci-status`, the
  sorted/comma-joined list of currently failing/cancelled check names; for
  `label-check`/`critical-file`, the sorted list of relevant names/paths;
  for `size-check`/`merge-conflict`, the criterion key itself (nothing varies).
- Embed `REASON_KEY` directly in the marker: `<!-- champion:reject:$CRITERION_KEY:$REASON_KEY -->`.
- Replace the `grep -qF "$REASON"` text comparison with a verbatim comparison of the
  most recently posted marker line against the current marker — dedup is now keyed
  entirely on the stable identity, never on the human-readable prose (which is free
  to vary tick to tick in the comment body).
- A genuinely different failure (different check name(s), different missing labels,
  etc.) — or the same check clearing and then failing again — still produces a fresh
  comment, since its `REASON_KEY` (and therefore marker) differs from the last one
  posted.

Scoped to this one guard in `defaults/.claude/commands/loom/champion-pr-merge.md`;
no other file references the old marker shape (verified via grep).

## Testing

This is a prose/documentation change to agent-behavior instructions (a markdown
fenced shell snippet describing the Champion role's procedure), not executable
production code, so there is no dedicated unit-test harness for it. Verified:

- `bash defaults/scripts/check-phantom-labels.sh .` — passes (this is the CI check
  that scans `.claude/commands/loom/*.md` for label-registry consistency; it is wired
[...]

PR #4838, opened 2026-07-31
docs(champion-pr-merge): correct REASON_KEY guard reappearance-detection claim

### LINKED ISSUE #4835: champion-pr-merge.md: REASON_KEY guard doesn't actually detect 'cleared and re-failed' as PR #4833 claims
## Context

PR #4833 (closing #4818) replaced the freeform-text comparison in the
transient-rejection idempotency guard (`defaults/.claude/commands/loom/champion-pr-merge.md`,
"PR Rejection Workflow → Transient failures") with a marker keyed on a
deterministic `$REASON_KEY`. This correctly fixes the spam bug it targets
(near-duplicate wording for an unchanged failure is now suppressed) — found no
issue with that part during review.

However, both the prose intro and the inline bash comment for the new guard
make a claim that the mechanism does not actually implement:

> "...a genuinely different failure (a different check name, a different set
> of missing labels, …) **or the same check clearing and then failing again**
> still produces a fresh comment"

> `# ... Skip re-commenting only when the identity is unchanged — a different
> REASON_KEY (...), or the same key reappearing after the criterion cleared
> and failed again, always gets a fresh comment.`

## What the guard actually does

```bash
LAST_MARKER=$(gh pr view "$PR_NUMBER" --json comments --jq '.comments' \
  | jq -r --arg prefix "<!-- champion:reject:$CRITERION_KEY:" \
  '[.[] | select(.body | startswith($prefix))] | last | .body // "" | split("\n")[0]')

if [ "$LAST_MARKER" = "$REJECT_MARKER" ]; then
  # skip — treated as duplicate
```

This compares only against the **single most-recently-posted** rejection
comment for the criterion. It has no way to represent "the criterion passed
at some intervening tick" — a passing tick posts no comment/marker at all.

So: if a flaky check fails (tick 1, posts marker `A`), then passes (tick 2,
no comment — criteria all green, or PR just isn't re-rejected on this
criterion), then fails again with the **identical** failing-check set (tick
3), `REASON_KEY` at tick 3 is the same as tick 1's, so `LAST_MARKER` (still
`A`, unchanged since tick 1) matches the new marker and the guard **suppresses
the comment** — contradicting the "always gets a fresh comment" claim.

Verified with a standalone simulation of the exact jq pipeline (mocked two
comments with an identical marker) — the guard suppresses as a duplicate, it
does not distinguish "still-failing, never cleared" from "cleared and
reappeared."

## Why this is filed as low-priority

- It's a **documentation-accuracy** gap in an agent-instruction file, not a
  functional regression: the actual runtime behavior fails in the *safe*
  direction (under-notifies rather than spams), and doesn't affect
  merge/safety decisions — only comment visibility on a re-flapped check.
- It does not reintroduce the #4586/#4754/#4818 spam bugs; the primary fix in
  #4833 is correct and verified for the cases those issues were about.

## Suggested fix (not prescriptive)

Either:
1. **Correct the claim** — soften the prose/comment to note that a
   REASON_KEY reappearing after an intervening clear is still treated as a
   duplicate (an accepted limitation), or
2. **Implement real reappearance detection** — e.g. compare against *all*
   rejection comments for the criterion (not just the last), and treat the
   guard as "unchanged" only if the *most recent PR-wide comment of any kind*
   (not just this criterion's rejections) carries the current marker,
   distinguishing "no other comment happened in between" from "criterion
   passed at some point since the last rejection for this key."

Option 1 is the lower-effort correct fix given the actual severity.

---
*Filed by Judge role during review of PR #4833 (merged by a concurrent Judge
session before this review's own verdict landed — see PR #4833 comment
history).*

rjwalters:
## Summary

PR #4833 keyed the transient-rejection idempotency guard in
`defaults/.claude/commands/loom/champion-pr-merge.md` on a stable
`$REASON_KEY`, but the prose intro and the inline bash comment for the
guard overstated its behavior — claiming that "the same check clearing and
then failing again" always produces a fresh comment.

In reality, the guard only compares the current `$REJECT_MARKER` against
the single most-recently-posted marker for that criterion
(`LAST_MARKER`). It has no way to represent "the criterion passed at some
intervening tick," since a passing tick posts no comment/marker. So if a
check fails (marker A), then passes, then fails again with the identical
`$REASON_KEY`, the guard treats it as a duplicate and suppresses the
comment — the opposite of what the docs claimed.

This is a documentation-accuracy fix only (Option 1 from the issue): the
prose and bash comment now describe the guard's actual behavior — a
REASON_KEY reappearing after an intervening clear is still treated as a
duplicate (accepted limitation: under-notifies rather than spams, and
doesn't affect merge/safety decisions). No runtime logic changed.

Verified there is only one copy of this guard text in the repo (grepped
for `REASON_KEY` across all markdown) — no installed/synced counterpart
needed updating.

Closes #4835

## Test plan

- [x] Confirmed via grep that `defaults/.claude/commands/loom/champion-pr-merge.md` is the sole copy of this guard text (no `.loom/roles/` or other synced copy)
- [x] Prose-only change; no lint/build/test suite applies (no markdown-lint or docs-consistency checker found in repo)
- [x] Manually re-read the corrected section for internal consistency

### DIFF
--- a/defaults/.claude/commands/loom/champion-pr-merge.md
+++ b/defaults/.claude/commands/loom/champion-pr-merge.md
@@ -1198,8 +1198,13 @@ identity** instead: the failing criterion (`$CRITERION_KEY`) *plus* a determinis
 `$REASON_KEY` built mechanically from the failure data — never freeform prose — so
 near-duplicate wording for the *same* underlying failure is recognized and
 suppressed, while a genuinely different failure (a different check name, a different
-set of missing labels, …) or the same check clearing and then failing again still
-produces a fresh comment:
+set of missing labels, …) produces a fresh comment. **Known limitation:** the guard
+only compares against the single most-recently-posted marker for the criterion, so
+it cannot represent "this criterion passed at some intervening tick" (a passing tick
+posts no marker) — if the identical `$REASON_KEY` reappears after clearing and
+re-failing, it is still treated as a duplicate and the comment is suppressed. This
+is accepted: it under-notifies on a re-flapped check rather than spamming, and
+doesn't affect merge/safety decisions (#4835):
 
 - **ci-status**: the sorted, comma-joined list of currently failing/cancelled check
   names (`echo "$FAILING_CHECKS" | sort | paste -sd, -`) — not the prose sentence
@@ -1227,9 +1232,13 @@ REASON="<SPECIFIC_REASON>"  # human-readable prose for the comment body — free
 # Idempotency guard: find the most recently posted rejection comment for this
 # criterion (any REASON_KEY) and compare its marker line, verbatim, against the
 # marker for the CURRENT failure. Skip re-commenting only when the identity is
-# unchanged — a different REASON_KEY (a different failing check, a different missing
-# label, …), or the same key reappearing after the criterion cleared and failed
-# again, always gets a fresh comment.
+# unchanged. A different REASON_KEY (a different failing check, a different missing
+# label, …) always gets a fresh comment. Known limitation: because this only looks
+# at the single most-recently-posted marker, a REASON_KEY that reappears after the
+# criterion cleared (passed) and then failed again with the same identity is
+# indistinguishable from "never cleared" and is still suppressed as a duplicate
+# (#4835) — acceptable since it under-notifies rather than spams and doesn't affect
+# merge/safety decisions.
 LAST_MARKER=$(gh pr view "$PR_NUMBER" --json comments --jq '.comments' \
   | jq -r --arg prefix "<!-- champion:reject:$CRITERION_KEY:" \
   '[.[] | select(.body | startswith($prefix))] | last | .body // "" | split("\n")[0]')


rjwalters:
LGTM — approving.

Verified against the actual code in `defaults/.claude/commands/loom/champion-pr-merge.md`:

1. **Correctness**: The corrected prose accurately describes the guard's real behavior. Traced through the logic — `LAST_MARKER` is derived from the single most-recently-*posted comment* matching the criterion prefix, and a passing tick posts no comment/marker at all. So if a check fails (marker A posted), then passes (no comment), then fails again with the identical `$REASON_KEY` (marker A again), `LAST_MARKER` still equals the current `$REJECT_MARKER` and the guard suppresses the comment as a duplicate — exactly what the new "Known limitation" text says, and the opposite of the old claim that this case "still produces a fresh comment." Matches the issue's (#4835) description precisely.

2. **Scope discipline**: Confirmed via diff — only the two prose/comment blocks changed (intro paragraph + inline bash comment above the `LAST_MARKER=...` guard). The `LAST_MARKER`/`REJECT_MARKER` comparison code itself (`if [ "$LAST_MARKER" = "$REJECT_MARKER" ]; then ... fi`) is byte-for-byte unchanged. No runtime logic touched.

3. **Clarity**: The "Known limitation" framing is clear, correctly explains the mechanism (single most-recent marker, no representation of an intervening pass), and explicitly notes the accepted trade-off (under-notifies, doesn't affect merge/safety decisions) — no new inaccuracies introduced.

4. **No unrelated changes**: Diff is 14/-5 lines, entirely within the two blocks described above. Nothing else bundled in.

CI: all checks pass (Rust/MCP jobs correctly `skipping` since this is a docs-only change to a `defaults/` markdown file). `mergeStateStatus` is CLEAN.

This is Option 1 from the issue (correct the docs) as recommended — appropriately lower-effort than implementing real reappearance detection, given the safe-direction failure mode.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (and `loom:review-requested`/`loom:reviewing` already removed) — another Judge's verdict raced mine. My independent review reached the same conclusion (doc-only fix, verified accurate against the guard's actual bash logic, CI green, closes #4835 correctly) but discarding my write since a verdict already landed.

PR #4771, opened 2026-07-31
feat(daemon): loom-daemon health — one-shot consolidated fleet vitals

rjwalters:
## Summary

Adds `loom-daemon health [--since 30m] [--json]`: a single call that collects
the six-check battery a manual overnight health tick used to run by hand
(trusted liveness, dispatch state, token pool, role-tick health, queue depth,
merge throughput), renders one structured line per section, and exits with a
contract a watch loop can branch on without parsing:

| exit | meaning |
|------|---------|
| `0` | every section green |
| `1` | degraded — any section non-green, including "could not determine" |
| `2` | the daemon is genuinely dead |

**Liveness precedence (#4694, load-bearing):** IPC-answered ⇒ alive
unconditionally; else the install-state classification (launchd → skipped-domain
→ pid-file) ⇒ degraded, never dead; else a live `pgrep -x loom-daemon` ⇒ still
degraded; only when all three independent signals are negative is the verdict
`DEAD`. The launchd domain probe alone can never produce exit `2` — it twice
declared a live, dispatching daemon dead during the referenced overnight watch.

**Role-tick health:** the role-runner loop now appends every `(root, role)`
tick outcome to a bounded process-global ring, carried over IPC. The
transient-vs-persistent classifier is client-side: a pair whose *latest*
record in the window is a failure surfaces as degraded; one that failed but
later succeeded is reported as a count only (self-recovered).

**One collector, two consumers today:** `health::assess(&HealthInputs) ->
HealthReport` is pure (no daemon, no forge, no subprocess), so every verdict
rule is directly unit-tested. `cli::health` collects the inputs (one IPC
round-trip, one install-state probe, one `pgrep`, one `.ranking` stat, a
bounded forge fan-out using a new `PipelineMetrics::HEALTH` mask that fetches
only the 2 of 6 `gh` counts this command needs) and renders/exits. The
dashboard's `GET /api/health` (`serve.rs`) calls the *same* `assess` over the
*same* cached `/api/pipeline` batch — no second verdict implementation — and
answers `200` (not `503`) with `overall: "dead"` / `exit_code: 2` when the
daemon is unreachable, since "the daemon is dead" is itself a health finding,
not a transport failure.

**Scope note on `fleet status`:** the issue's suggested AC also names `fleet
status` as a consumer. `fleet status` answers a different question (a
multi-host UP/DOWN roster over SSH) and already reuses the same
`daemon_install_state::probe()` primitive `health::assess_liveness` is built
[...]

PR #4843, opened 2026-07-31
fix(health): distinguish CLI/daemon build skew and daemon warm-up from a dead work finder

### LINKED ISSUE #4824: health: false DEGRADED from CLI/daemon version skew and fresh-restart window (work-finder tick telemetry)
## Context

Found during a 2026-07-31 fleet-health check + hygiene pass. `loom-daemon health` (#4771) reported:

```
dispatch    DEGRADED  0 in-flight / cap 5; no work-finder tick observed in this daemon process
overall     DEGRADED  exit 1 (window 30m)
```

while `~/.loom/daemon.log` showed `work_finder:` ticks landing **every ~60s**, with successful dispatches minutes earlier. Two distinct false-DEGRADED modes reproduced in one session:

1. **Version skew**: the `health` CLI was built from HEAD (`18887b5c`, which has #4771) but the running daemon was one commit older (`105f9c12`, predating the tick telemetry). The daemon can't report a counter it doesn't have, and `health` renders that absence as "work finder dead" — on a fleet that was demonstrably dispatching. The CLI knows both its own commit and can fetch the daemon's (`loom-daemon --version` embeds it; the IPC status report carries build info) — skew should be detected and reported as its own condition ("daemon predates tick telemetry — update via loom-daemon-update.sh"), not conflated with a dead work finder.
2. **Fresh restart**: immediately after a supervised restart (`loom-daemon restart`), `health` reports the same DEGRADED for up to one tick interval (~60s), because the per-process telemetry is legitimately empty. A grace window keyed on daemon uptime (e.g. suppress the tick check until uptime > 2× tick interval) would remove the false alarm operators hit right after every update.

Both matter because `health` exits 1, so anything scripted on it (cron alerting, watchdogs) pages on a healthy fleet right after every daemon roll.

## Verified against current code (2026-07-31 curation pass)

Confirmed both gaps are real, not already handled:

- **No daemon build-commit field exists on the wire today.** `DaemonStatusReport` (`loom-daemon/src/types.rs`) has no version/commit field, and neither `loom-daemon/src/health.rs` nor `loom-daemon/src/cli/health.rs` reference one. `loom-daemon/src/self_update.rs` already defines `self_update::BUILT_COMMIT` (baked in at compile time via `build.rs` → `LOOM_DAEMON_GIT_COMMIT`, the same value in `loom-daemon --version`) and is used by `loom-daemon status`'s **client-side** staleness check (comparing the binary against its source checkout) — but that is a different comparison than "does the CLI process's own `BUILT_COMMIT` match the commit the *running daemon process* was built from?" Nothing computes or transmits the latter today.
- **The false-DEGRADED path is `assess_dispatch` in `loom-daemon/src/health.rs:677-683`**: when `status.last_work_finder_tick` is `None` and `work_finder_enabled != Some(false)`, it unconditionally pushes `"no work-finder tick observed in this daemon process"` into `degraded`, with no version-skew or startup-grace carve-out.
- **The startup-grace signal already exists and is already plumbed in**, just unused here: `HealthInputs.install_state.process_age_secs` (populated via `daemon_install_state::probe()` in `loom-daemon/src/cli/health.rs`) is the live daemon process's age in seconds and is already read elsewhere in `health.rs` (e.g. lines 497, 612) — `assess_dispatch` just needs to consult it.

## Acceptance Criteria

- [ ] `DaemonStatusReport` (or the `DaemonStatus` IPC response) carries the running daemon's own build commit (reuse `self_update::BUILT_COMMIT`, computed daemon-side) so the CLI can compare it against its own `self_update::BUILT_COMMIT`
- [ ] `health` compares its own build commit with the running daemon's; on mismatch where the daemon predates tick telemetry, the dispatch row reports the skew explicitly as its own condition (e.g. "daemon predates tick telemetry (commit `<sha>`) — update via loom-daemon-update.sh"), not as DEGRADED-as-dead-work-finder
- [ ] Dispatch tick check (`loom-daemon/src/health.rs`'s `assess_dispatch`, the `None` arm around line 677) is suppressed (or reported as "warming up") while `install_state.process_age_secs` is less than ~2× the work-finder tick interval
- [ ] Exit code stays 0 in both cases when everything else is GREEN
- [ ] A log-based fallback (last `work_finder:` line age from `~/.loom/daemon.log`) is considered as a corroborating signal before declaring the work finder dead
- [ ] `#[serde(default)]` on any new `DaemonStatusReport` field so older daemon binaries (pre-this-change) remain wire-compatible with a newer CLI, matching the existing convention (`token_pool_dir`, `logical_cpus`, etc. in the same struct)

## Affected Files

- `loom-daemon/src/types.rs` — add a build-commit field (e.g. `daemon_build_commit: Option<String>`, `#[serde(default)]`) to `DaemonStatusReport`
- `loom-daemon/src/ipc.rs` (`build_daemon_status`, `build_daemon_status_with_drain`, ~line 1446/1740) — populate the new field from `self_update::BUILT_COMMIT` daemon-side
- `loom-daemon/src/self_update.rs` — reuse `BUILT_COMMIT` (no change expected; just the comparison source on both sides)
- `loom-daemon/src/health.rs` (`assess_dispatch`, lines 626-705, specifically the `None` arm at 677-683) — add the version-skew condition and the `process_age_secs`-gated grace window before pushing "no work-finder tick observed"
- `loom-daemon/src/cli/health.rs` (`collect`, `HealthInputs` construction) — no structural change expected; the CLI's own `self_update::BUILT_COMMIT` is already available at compile time, just needs threading into `HealthInputs` alongside `status`

## Test Plan

- [ ] Automated tests: extend `loom-daemon/src/health.rs`'s existing `assess_dispatch` unit tests (see `dispatch_missing_tick_is_degraded_when_work_finder_is_enabled`, `dispatch_missing_tick_is_green_when_work_finder_is_disabled_and_nothing_else_is_wrong` near the bottom of the file) with two new cases: (1) missing tick + version skew → skew-specific message, GREEN/non-dead-work-finder exit; (2) missing tick + `process_age_secs` below the grace threshold → "warming up", GREEN
- [ ] Automated tests: a case where missing tick + no skew + process age *above* the grace threshold still reports DEGRADED (regression guard so the fix doesn't silence a genuinely dead work finder)
- [ ] Manual verification: build the daemon from an older commit, run a newer CLI's `loom-daemon health` against it, confirm the skew message appears and exit code is 0 when nothing else is wrong
- [ ] Manual verification: `loom-daemon restart` then immediately run `loom-daemon health` — confirm no false DEGRADED during the grace window, and confirm it correctly flips to DEGRADED if no tick appears after the grace window elapses

## Dependencies

None — builds on already-merged #4771 (tick telemetry) and the already-existing `self_update::BUILT_COMMIT` / `install_state.process_age_secs` plumbing; no blocking issues.

rjwalters:
## Summary

`loom-daemon health` reported `dispatch DEGRADED` — and exited `1` — whenever the daemon's status report carried no `last_work_finder_tick`, with no way to tell a genuinely stopped loop apart from two states in which that absence says nothing about the work finder:

1. **Build skew** — a `health` CLI built from a commit carrying #4771's tick telemetry querying a daemon binary built *before* it. The daemon cannot report a counter it does not have, so its honest `None` was rendered as "work finder dead" on a fleet that was demonstrably dispatching (the 2026-07-31 observation: `health` said `no work-finder tick observed in this daemon process` while `~/.loom/daemon.log` carried a `work_finder:` line every ~60s).
2. **Fresh restart** — for up to one tick interval after every `loom-daemon restart`, the per-process telemetry is legitimately empty.

Because `health` exits `1`, anything scripted on it (cron alerting, watchdogs) paged on a healthy fleet right after every daemon roll. This PR makes each of those a named condition of its own that keeps the section GREEN, while leaving a genuinely dead work finder DEGRADED.

## Changes

**Wire (`loom-daemon/src/types.rs`, `ipc.rs`)** — `DaemonStatusReport` gains two `#[serde(default)]` fields, both populated daemon-side in `build_daemon_status` (which `build_daemon_status_with_drain` delegates to, so there is one construction site) because only the answering process can state them truthfully:

- `daemon_build_commit: Option<String>` — `self_update::BUILT_COMMIT`, the same value `loom-daemon --version` embeds. Deliberately distinct from `SelfUpdateStatus::built_commit`, which compares *this* process's binary against its source checkout; this compares the **client** process's build against the **daemon** process's.
- `work_finder_interval_secs: Option<u64>` — the interval THIS process resolved via `resolve_interval_with_config` (env > config > default), so a client sizes the grace window by the daemon's real cadence instead of assuming 60s.

**Classification (`loom-daemon/src/health.rs`)**

- `classify_build_skew` compares the client's own `BUILT_COMMIT` against the daemon-reported one → `Match` / `Skew(sha)` / `DaemonUnknown` / `Incomparable`. A tarball build's `"unknown"` on either side is `Incomparable`, never a false skew — that would otherwise permanently suppress the dead-loop verdict on such an install.
- `classify_missing_tick` explains a missing tick in strict precedence: **warming up** (process younger than `2 x` the daemon's own resolved interval) > **build skew** > **recent `work_finder:` daemon-log activity** > **dead**. Only `Dead` still pushes `"no work-finder tick observed in this daemon process"` into `degraded`.
- Warm-up is checked first because it is the one explanation true regardless of build state, and it is what an operator hits immediately after every restart.

**Corroborating log signal (`health.rs`, `cli/health.rs`, `serve.rs`)** — one bounded 256 KiB tail read of the daemon log, taken **only** when the daemon reported no tick (a reported tick is already the stronger signal), and able only to *soften* the verdict, never harden it. A log line older than the grace window corroborates nothing. The parsing half (`work_finder_log_tick_age_secs`) is pure over the log text so it is unit-testable without a daemon.

**Observability** — the dispatch row's `--json` detail now carries `missing_tick` (the reason enum), both build commits, the resolved interval, and the log-line age, so a consumer sees *why* a missing tick was or was not treated as a fault.

- [x] **`DaemonStatusReport` carries the daemon's own build commit** — `daemon_build_commit`, set from `self_update::BUILT_COMMIT` in `build_daemon_status` (`ipc.rs`). Verified live below.
- [x] **`health` compares its own build commit with the daemon's; on mismatch reports skew as its own condition** — `classify_build_skew` + the `MissingTick::BuildSkew` arm renders `no tick telemetry: daemon build <sha> ≠ CLI build <sha> — daemon predates tick telemetry; update via loom-daemon-update.sh`. Pinned by `dispatch_missing_tick_reports_build_skew_not_a_dead_work_finder`, which asserts the sha, `predates tick telemetry`, `loom-daemon-update.sh`, and the **absence** of the dead-work-finder message.
- [x] **Tick check suppressed / reported as "warming up" while `process_age_secs` < ~2x the tick interval** — `WORK_FINDER_TICK_GRACE_INTERVALS = 2`, applied against the daemon's *own* reported interval with the default cadence as fallback. Pinned by `dispatch_missing_tick_within_the_grace_window_reports_warming_up` and `the_grace_window_scales_with_the_daemons_own_tick_interval`.
- [x] **Exit code stays 0 in both cases when everything else is GREEN** — asserted explicitly via `assess(&inputs).exit_code() == EXIT_HEALTHY` in `build_skew_alone_does_not_change_the_exit_code` and in the warming-up test.
- [x] **Log-based fallback considered as a corroborating signal before declaring the work finder dead** — `MissingTick::LogCorroborated`, pinned by `recent_work_finder_log_activity_blocks_a_dead_verdict` (fresh softens) and `stale_work_finder_log_activity_does_not_block_a_dead_verdict` (stale does not).
- [x] **`#[serde(default)]` on both new fields for wire compatibility** — matching the existing convention (`token_pool_dir`, `logical_cpus`, `pid_file`). **Verified live** against the currently-running older daemon: it omits both fields, they deserialize to `null`, and `health` still reports GREEN/exit 0 (output below).

## Test Plan

- [x] **New unit tests** (all in `health.rs`): skew → skew-specific message + GREEN; daemon reporting no commit at all → also skew (`<unreported>`); warm-up → "warming up" + GREEN + exit 0; grace window scaling with a 300s cadence; default-cadence fallback for a pre-#4824 daemon; warm-up outranking skew; fresh vs. stale log corroboration; `classify_build_skew` match/skew/unknown-daemon/incomparable; an incomparable build still reporting dead; and the pure log-tail parser (newest line wins, no matching line, unparseable partial first line from a mid-line tail seek, future stamp clamped to 0).
- [x] **Regression guard** — `a_genuinely_dead_work_finder_is_still_degraded`: matching build + daemon 1h past the grace window + no log activity is still `Verdict::Degraded`, still carries the dead-work-finder message, still `EXIT_DEGRADED`. The pre-existing `dispatch_missing_tick_is_degraded_when_work_finder_is_enabled` was retargeted onto that same fixture rather than deleted.
- [x] `cargo test -p loom-daemon --lib health::` → **71 passed, 0 failed**.
- [x] `cargo test -p loom-daemon --lib` → **2940 passed, 1 failed**. The single failure is `config_resolver::tests::test_resolve_malformed_legacy_config_soft_fails_never_aborts`, a **pre-existing host-local** failure caused by this machine's live tier-1 private defaults file leaking through `LOOM_CONFIG_DEFAULTS_FILE`; unrelated to this change and untouched by it.
- [x] `cargo fmt -p loom-daemon` → no changes.
- [x] **Manual verification (wire compat + no regression on the normal path)** — built this branch and ran its `health` against the live daemon, which is running an **older** binary that predates both new fields:

  ```
  loom-daemon 0.17.0 (commit 0c5cd1b1, built 2026-07-31T22:39:19Z)

  liveness    GREEN     daemon alive — IPC round-trip ok (pid 81820)
  dispatch    GREEN     16 in-flight / cap 16; last tick 15s ago: 20 seen, 0 dispatched, ...
  overall     GREEN     exit 0 (window 30m)
  ```

  and its `--json` dispatch detail:

  ```json
  { "missing_tick": null, "daemon_build_commit": null, "cli_build_commit": "0c5cd1b1",
    "work_finder_interval_secs": null, "work_finder_log_tick_age_secs": null }
  ```

  i.e. a newer CLI deserializes an older daemon's payload without error, both new fields default to `null`, `missing_tick` stays `null` while a tick is reported, and the healthy path is unchanged.

### Verification not performed locally

The two remaining manual steps in the issue's test plan (building a daemon from an older commit and pointing a newer CLI at it; `loom-daemon restart` then immediately running `health`) both require **booting a daemon on this host**, which would evict the running production daemon and start dispatching work. They were not run. The states they target are covered by the unit tests above — which assert the rendered summary text, the `missing_tick` reason, and the process exit code, not just the verdict — and the live run above independently confirms the older-daemon wire path.

## Notes for review

- `cargo clippy -p loom-daemon --all-targets` fails on `loom-daemon/src/terminal.rs:2590` (`unnecessary_get_then_check`). This is **pre-existing on `origin/main`** (introduced by #4808) and `terminal.rs` is not touched by this branch — left alone per scope discipline.
- Branch is rebased onto `origin/main` (`91bfd324`).

Closes #4824

### DIFF
--- a/loom-daemon/src/cli/health.rs
+++ b/loom-daemon/src/cli/health.rs
@@ -113,6 +113,18 @@ async fn collect(window: Duration) -> HealthReport {
         None => None,
     };
 
+    // 5. The daemon log's newest `work_finder:` line (#4824) — a *corroborating*
+    //    signal, never a derivation: it exists so the collector refuses to call
+    //    the work finder dead while the daemon's own log shows it ticking. One
+    //    bounded tail read, and only on the path that can consume it (the
+    //    daemon reported no tick) — a reported tick is already the stronger
+    //    signal, so probing then would be pure I/O for a field nothing reads.
+    let work_finder_log_tick_age_secs = status
+        .as_ref()
+        .is_some_and(|r| r.last_work_finder_tick.is_none())
+        .then(health::probe_work_finder_log_tick_age)
+        .flatten();
+
     health::assess(&HealthInputs {
         at: chrono::Utc::now(),
         window,
@@ -124,6 +136,12 @@ async fn collect(window: Duration) -> HealthReport {
         ranking_present,
         ranking_age_secs,
         pipeline,
+        // This CLI process's own build commit (#4824), compared daemon-side
+        // against `DaemonStatusReport::daemon_build_commit` so a newer CLI
+        // querying an older daemon reports build skew rather than a phantom
+        // dead work finder.
+        cli_build_commit: loom_daemon::self_update::BUILT_COMMIT.to_string(),
+        work_finder_log_tick_age_secs,
     })
 }
 
--- a/loom-daemon/src/cli/status.rs
+++ b/loom-daemon/src/cli/status.rs
@@ -663,6 +663,8 @@ pub(crate) mod status_client_tests {
             role_tick_records: vec![],
             daemon_pid: Some(99917),
             pid_file: Some(std::path::PathBuf::from("/repo/a/.loom/.daemon.pid")),
+            daemon_build_commit: Some("18887b5c".to_string()),
+            work_finder_interval_secs: Some(60),
         }
     }
 
--- a/loom-daemon/src/health.rs
+++ b/loom-daemon/src/health.rs
@@ -96,6 +96,26 @@ pub const DEFAULT_WINDOW_SECS: u64 = 30 * 60;
 /// the past.
 pub const RANKING_STALE_SECS: u64 = 6 * 600;
 
+/// How many work-finder tick intervals a daemon process must have been running
+/// before "no tick observed" counts as a fault (Issue #4824).
+///
+/// A freshly (re)started daemon legitimately has empty per-process tick
+/// telemetry until its loop's first tick lands, so for up to one interval after
+/// every `loom-daemon restart` — i.e. after every update roll — `health` used to
+/// report `dispatch DEGRADED` and exit `1` on a perfectly healthy fleet, paging
+/// whatever watchdog was scripted on it. Two intervals is the smallest window
+/// that also absorbs one *overrun* tick (the loop measures the next interval
+/// from when the previous tick's work finished) without absorbing a second
+/// consecutive miss — a work finder that has actually stopped is still caught
+/// within ~2 minutes on the default cadence.
+pub const WORK_FINDER_TICK_GRACE_INTERVALS: u64 = 2;
+
+/// The commit string a build with no git information available bakes in — the
+/// tarball-install case (`build.rs` cannot run `git rev-parse`). Compared as a
+/// value rather than matched structurally so both sides of the skew check treat
+/// it as "cannot compare" instead of as a real commit that never matches.
+const UNKNOWN_BUILD_COMMIT: &str = "unknown";
+
 // ============================================================================
 // Verdicts
 // ============================================================================
@@ -300,6 +320,27 @@ pub struct HealthInputs {
     pub ranking_age_secs: Option<u64>,
     /// Per-repo forge snapshot (`queued` + merged-in-window), when collected.
     pub pipeline: Option<Vec<RepoPipelineSnapshot>>,
+    /// The commit **this client process** was built from (Issue #4824) —
+    /// [`crate::self_update::BUILT_COMMIT`], threaded in by the collector
+    /// rather than read directly here so the skew rules are unit-testable
+    /// against synthetic commit pairs. Compared against the daemon's
+    /// [`DaemonStatusReport::daemon_build_commit`]; `"unknown"` (a tarball
+    /// build) means "cannot compare", never "skew".
+    pub cli_build_commit: String,
+    /// Age in seconds of the newest `work_finder:` line in the daemon log, when
+    /// the log was readable (Issue #4824).
+    ///
+    /// A **corroborating** signal only, in the same spirit as
+    /// [`Self::pid_file`]: it is never used to *derive* dispatch health, only to
+    /// refuse to declare the work finder dead while the daemon's own log shows
+    /// it ticking. That is exactly the 2026-07-31 disagreement this exists for —
+    /// `health` reporting "no work-finder tick observed" while `daemon.log`
+    /// carried a `work_finder: tick —` line every ~60s. `None` when the log
+    /// could not be resolved/read or carries no such line, which is treated as
+    /// "no corroboration either way", never as evidence of death — and also
+    /// when the collector did not bother to probe, which both collectors skip
+    /// unless the daemon actually reported no tick.
+    pub work_finder_log_tick_age_secs: Option<u64>,
 }
 
 // ============================================================================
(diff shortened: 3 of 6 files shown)


rjwalters:
✅ **Approved!**

Reviewed the fix for #4824 (build-skew / warm-up / log-corroboration classification of a missing work-finder tick). Solid, well-scoped change.

**What I checked:**
- Read the full diff (`types.rs`, `ipc.rs`, `serve.rs`, `health.rs`, `cli/health.rs`, `cli/status.rs`) — the wire fields (`daemon_build_commit`, `work_finder_interval_secs`) are both `#[serde(default)]` and populated only daemon-side, matching the existing convention in the struct.
- Verified the precedence logic (warming-up first, then build-skew, then log-corroboration, then dead) matches the PR's stated rationale and is reflected correctly in `classify_missing_tick`.
- Confirmed `resolve_interval_with_config`, `DEFAULT_WORK_FINDER_INTERVAL_SECS`, and `self_update::BUILT_COMMIT` all exist as referenced.
- `cargo test -p loom-daemon --lib health::` result: 71 passed, 0 failed.
- `cargo fmt -p loom-daemon --check` result: clean.
- `gh pr checks 4843` result: all required checks pass; merge state is now CLEAN.
- `cargo test -p loom-daemon --lib` (full suite) result: 2942 passed, 5 failed. Reproduced the same 5 failures (`auto_update::tests::test_config_malformed_json_is_default`, `test_config_missing_block_is_default`, `work_finder::tests::test_config_enabled_false_is_disabled_flag`, `test_config_per_token_concurrency_read_without_work_finder_block`, `test_deprecated_cpu_knobs_accepted_at_any_value_including_nonsense`) identically against unmodified `origin/main` in the same host environment. Confirmed pre-existing/host-local (config-defaults-file leakage), not introduced by this branch.
- Verified the PR's clippy-pre-existing claim: `terminal.rs` was last touched by #4808, not this branch.
- `Closes #4824` present and correct.

## Test Execution

1. New unit tests (health.rs) — executed, all pass, confirmed via `cargo test -p loom-daemon --lib health::`
2. Regression guard (`a_genuinely_dead_work_finder_is_still_degraded`) — executed, passes, confirms the fix cannot silence a genuinely dead work finder
3. `cargo test -p loom-daemon --lib health::` reporting 71 passed — executed, matches PR claim
4. `cargo test -p loom-daemon --lib` reporting 2940 passed / 1 host-local failure — my run showed 5 host-local failures rather than 1, but reproduced identically on unmodified main; same root cause (host-local config leakage), not a regression
5. `cargo fmt -p loom-daemon` no changes — executed, confirmed
6. Manual verification against live older daemon — skipped, requires booting a daemon on this host (would evict the running production daemon); not re-verified, but covered by the wire-compat unit tests (serde default round-trip)

No code changes needed. Nice work tracing the 2026-07-31 false-DEGRADED incident to its root cause and building a properly layered fix with a regression guard for the genuine-dead-loop case.

rjwalters:
**Standing down (concurring) — no label write.**

I ran an independent Judge pass on this PR in parallel. My verdict-time CAS recheck found `loom:pr` already applied by another Judge's review (comment above), so per the Judge role's raced-verdict rule I am **not** writing a second verdict label. My independent conclusion **agrees**: approve.

Recording what my pass added beyond the existing review, purely as non-blocking notes (no change requested, do not block the merge):

**Independently verified**
- All 6 acceptance criteria in #4824 are met, each with a named test: `daemon_build_commit`/`work_finder_interval_secs` on the wire with `#[serde(default)]` (AC1/AC6), skew reported as its own condition (AC2, `dispatch_missing_tick_reports_build_skew_not_a_dead_work_finder`), `process_age_secs`-gated warm-up at 2x the daemon's *own* interval (AC3, plus `the_grace_window_scales_with_the_daemons_own_tick_interval`), exit code 0 in both cases (AC4, asserted via `assess(&inputs).exit_code() == EXIT_HEALTHY`), and log corroboration considered before `Dead` (AC5, fresh-softens / stale-does-not pair).
- The disabled-work-finder arm is behavior-preserving: the old `!= Some(false)` guard and the new `== Some(false)` early-return are equivalent for that path.
- `resolve_daemon_log_path` genuinely mirrors the binary-side `daemon_service::resolve_log_path` / `resolve_loom_dir` pair (`LOOM_DAEMON_LOG` > `LOOM_SOCKET_PATH` parent > `$HOME/.loom`), and `DAEMON_LOG_STAMP_FORMAT` matches the real `env_logger` hook (`chrono::Local::now().format("%Y-%m-%dT%H:%M:%S%.3f")` inside `[...]`) — including the local-vs-UTC subtlety the comment calls out.
- `a_genuinely_dead_work_finder_is_still_degraded` is a real regression guard, and retargeting the pre-existing `dispatch_missing_tick_is_degraded_when_work_finder_is_enabled` onto the same fixture (rather than deleting it) is the right call.
- The clippy finding is genuinely out of scope: `terminal.rs:2590` was last touched by #4808, is untouched by this branch, and CI's clippy invocation does not pass `--all-targets`, so it is not even in the CI surface.
- The skipped manual steps are an acceptable substitute here: the unit tests assert rendered summary text, the `missing_tick` reason enum, and the process exit code — not just the verdict — and the live older-daemon run in the PR body independently exercises the wire-compat path.

**Two non-blocking observations for a future pass (no action needed on this PR)**

1. *Doc-comment accuracy on `DAEMON_LOG_TAIL_BYTES`.* The comment states "the work finder writes at least one line per tick". It does not: the `work_finder: tick — ...` line in `work_finder.rs` is emitted only when `dispatched > 0 || errors > 0 || skipped_* > 0 || deferred_ramp_cap > 0`, so a genuinely idle fleet ticks silently. This makes the corroborating signal weaker than the comment implies, but it fails in the safe direction (no line -> `None` -> no corroboration -> the classifier falls through rather than softening), so it is a comment-accuracy nit only.

2. *Skew short-circuits log corroboration.* `classify_missing_tick` returns at `BuildSkew` before consulting `work_finder_log_tick_age_secs`. On a host where the CLI is routinely rebuilt from HEAD while the daemon is rolled less often, `Skew` is the steady state, so the dead-work-finder verdict is effectively suppressed until the daemon is updated. That is exactly what AC2 asked for, and the rendered row does tell the operator to run `loom-daemon-update.sh`, so it is the right trade against the false pages this PR fixes — just worth remembering that dead-loop detection is build-skew-gated from here on.

PR #4845, opened 2026-07-31
fix(scripts): anchor complexity-marker extraction to canonical <!-- --> form

rjwalters:
## Summary

`require-complexity-marker.sh` and `resolve-tier-model.sh` extracted the
complexity tier with:

```bash
tier="$(printf '%s' "$body" | grep -o 'loom:complexity=[a-z]*' | head -1 | cut -d= -f2)"
```

`head -1` takes the FIRST match in document order, but the real marker is
conventionally placed near the end of the issue body. If the body discusses
the marker syntax in prose before the real marker (e.g. an issue about the
complexity-marker feature itself that quotes `<!-- loom:complexity=<tier> -->`
as example text), the `<` after `=` matches zero `[a-z]*` characters,
producing an empty match that "wins" over the real marker later in the body —
causing `require-complexity-marker.sh` to false-positive "no marker" (blocking
`loom:curated`) and `resolve-tier-model.sh` to silently fall back to `routine`
instead of surfacing an error.

## Fix

Both scripts now anchor the extraction to the canonical
`<!-- loom:complexity=<tier> -->` HTML-comment form (excluding bare-word
mentions in prose/backticks that aren't a well-formed comment) and take the
**last** such match, matching the convention that the real marker sits near
the end of the body:

```bash
tier="$(printf '%s' "$body" | grep -oE '<!--[[:space:]]*loom:complexity=[a-z]*[[:space:]]*-->' | tail -1 | sed -E 's/.*complexity=([a-z]*).*/\1/')"
```

Searched the repo (`grep -rln "grep -o 'loom:complexity"`) — these are the
only two git-tracked copies of the extraction line; `.loom/scripts` is a
symlink to `defaults/scripts`, so no separate installed copy needs updating.

## Testing

- `defaults/scripts/tests/test-require-complexity-marker.sh` — 20/20 passed
  (added a new prose-before-marker fixture case, issue 9007)
- `defaults/scripts/tests/test-resolve-tier-model.sh` — 20/20 passed (added
  the same prose-before-marker case, and updated the existing multi-marker
  test to the new last-marker-wins semantics)
- `shellcheck` clean on all four changed files
- `defaults/scripts/tests/check-ci-suite-manifest.sh` — OK, both suites
  already wired into CI
- Spot-check against the real issue that triggered this bug:
  `./.loom/scripts/require-complexity-marker.sh 4827` now exits 0
  (`ok: rjwalters/loom#4827 is tagged routine`)

Closes #4840

Issue #4840, opened 2026-07-31
require-complexity-marker.sh / resolve-tier-model.sh: first-match extraction false-positives on issue bodies that discuss the marker syntax


rjwalters:
## Problem Statement

Discovered while curating #4827 (a meta issue about the complexity-marker feature itself, so its body legitimately discusses `loom:complexity=<tier>` as literal example text before the real marker appears).

Both `defaults/scripts/require-complexity-marker.sh:49` and `defaults/scripts/resolve-tier-model.sh:71` extract the tier with:

```bash
tier="$(printf '%s' "$body" | grep -o 'loom:complexity=[a-z]*' | head -1 | cut -d= -f2)"
```

`grep -o` returns every match in document order; `head -1` takes the **first**, not the actual marker (conventionally placed at/near the end of the body per the Curator role file's own examples). If the body contains any *other* text matching `loom:complexity=` followed by zero or more lowercase letters — e.g. prose like `` `<!-- loom:complexity=<tier> -->` `` (the `<` after `=` matches zero `[a-z]` chars, producing an empty-tier match) — that empty match wins over the real one.

### Reproduction (issue #4827's actual body)

```
$ printf '%s' "$body" | grep -o 'loom:complexity=[a-z]*'
loom:complexity=
loom:complexity=
loom:complexity=
loom:complexity=routine     # <- the real marker, added last, never reached by head -1
```

`require-complexity-marker.sh 4827` reports `BLOCKED: issue has no complexity marker` even though a valid `<!-- loom:complexity=routine -->` marker is present in the body.

### Why this matters

- **`require-complexity-marker.sh`** treats the empty match as "no marker" and blocks `loom:curated` from being applied — a false positive that will recur on any issue whose body explains or documents the marker syntax (meta issues about the complexity-marker feature, curator-authored issues that quote the Curator role file's own examples, etc.).
- **`resolve-tier-model.sh`** shares the same extraction and would resolve the empty match to `""`, which its own vocabulary check treats as invalid and falls back to `routine`. This happens to be harmless for #4827 specifically (its real marker also says `routine`), but is a silent miscalculation for any issue whose *real* marker is `mechanical` or `complex` and which also happens to mention the syntax in prose — the dispatcher would silently downgrade/upgrade the resolved model from what the Curator actually intended, with no error surfaced.

## Acceptance Criteria

[...]

Issue #4809, opened 2026-07-31
experiment: model A/B collected ZERO records — prose instrumentation never executes in headless children AND the #4501 dispatch pin (tier-1) overrides the forced arm; move arm assignment + outcome recording into the daemon (#3725/#3718)


rjwalters:
## Context
The #3725 model-cost experiment was enabled on 7 canary repos (2026-07-31, `experiment` + CANARY sentinels, verified via assign-arm/banner CLI) — and collected **zero records all day** despite dozens of canary sweeps. Post-mortem found two independent, compounding defects:

1. **Prose instrumentation does not execute in headless children.** sweep.md's "deterministic" instructions (resolve-mode at lifecycle entry, banner, per-phase `sweep-experiment.sh record` calls) appear in NO canary sweep log — no banner, no records, `.loom/stats/sweep-model-stats.jsonl` absent on every canary. One child (gf180-pll sweep-issue-24) even treated the `sweep.modelExperiment` config as "an unrelated pre-existing edit" and quarantined it mid-sweep. LLM-prose instrumentation is unreliable in `-p` children; this is the same lesson that moved arm arithmetic into compiled code (#3725's own design note) — the *invocation* has to be mechanical too.
2. **The #4501 dispatch model pin structurally defeats the A/B.** `dispatch_sweep`/work_finder now pins `--model sonnet (source=default)` on every child. Per the documented precedence, an explicit dispatch param is tier-1 and **outranks the forced arm** — so even a perfectly-instrumented child would run sonnet on both arms. Every daemon-dispatched canary builder since the pin has been Arm-B-by-fiat.

## Proposed fix: move the experiment into the daemon (where the model decision already lives)
- **Arm assignment at dispatch**: when the workspace resolves `experiment` (daemon reads the same config + CANARY guardrails as `sweep-experiment.sh resolve-mode` — port the check), work_finder/dispatch_sweep computes the deterministic arm (issue × complexity, same pure function) and passes THAT model instead of the #4501 default pin (the pin remains for off/observe workspaces). Log `arm=A|B` in the dispatch line.
- **Outcome recording at reap**: the reaper already observes exit/no_progress/death_class and the linked PR; the #4704 telemetry outcome records already persist model/config/duration/result per sweep. Add the `arm` + `complexity` fields to the outcome record (schema is additive per #4703) — then the experiment's stats store IS the telemetry store, and the harvest reads D1/local outcome records instead of a per-repo JSONL nobody writes.
[...]

PR #4851, opened 2026-07-31
feat(daemon): thread per-issue complexity marker into model-cost experiment dispatch

### LINKED ISSUE #4827: feat(daemon): thread per-issue complexity marker into model-cost experiment dispatch
## Context

Follow-up to #4809, which moved model-cost A/B experiment arm assignment and
outcome attribution into the daemon (`resolve_autonomous_dispatch_model` in
`loom-daemon/src/sweep_registry/model.rs`). That fix deliberately deferred two
refinements to keep the PR focused and mergeable:

## 1. Real per-issue complexity stratification at dispatch time

`assign_arm(issue_number, complexity)` stratifies by the Curator's
`<!-- loom:complexity=<tier> -->` marker (issues #3702/#4238/#4448) so both
arms see a comparable difficulty mix. The daemon-native dispatch path
introduced in #4809 always calls `assign_arm(issue, None)` (defaults to the
`"routine"` stratum) because none of the three call sites
(`work_finder.rs`'s `dispatch()`, `epic_supervisor.rs`'s `dispatch_sweep`, and
`ipc.rs`'s `dispatch_sweep` MCP handler) currently fetch the issue body at
dispatch time.

`WorkItem` (`work_finder.rs`) and the REST issue listing
(`forge_listing::RestIssue`) already carry `body: Option<String>` — the REST
listing already fetches it, it is just not threaded through today. Wiring this
requires:

- Adding `body: Option<String>` to `WorkItem` and populating it from
  `RestIssue::body` in `GhWorkSource::list_ready_issues`.
- Extracting the `loom:complexity=<tier>` marker from the body (mirroring
  `resolve-tier-model.sh`'s `grep -o 'loom:complexity=[a-z]*' | cut -d= -f2`
  extraction) and passing it through to `resolve_autonomous_dispatch_model`.
- The `WorkDispatcher::dispatch(&mut self, issue: u32)` trait signature (and
  its epic-supervisor / test-fixture counterparts) would need to grow a
  `complexity: Option<&str>` parameter, or the caller resolves complexity
  once from the `WorkItem`/issue body before calling `dispatch`.
- `epic_supervisor.rs`'s epic-child dispatch and the MCP `dispatch_sweep`
  handler would need the equivalent body-fetch (a `gh issue view` or reuse of
  the cached REST listing) before computing the arm.

This is still fully deterministic without it (arm assignment is a pure
function of the issue number + stratum, and "routine" is a valid, existing
default) — the gap is that both complexity strata do not get an independent
~50/50 A/B balance yet, only the pooled population does.

## 2. Explicit-arm attribution on `SweepInfo` (optional; inference is not the same)

Outcome-telemetry attribution added in #4809 infers the arm from the
sweep's *dispatched* model (`infer_arm_from_model`, reusing the existing
harvest-time inference in `script_helpers::sweep_experiment`). For an
`experiment`-mode dispatch this is exactly equivalent to the explicit arm
(the daemon always forces exactly an opus/sonnet-family model), so no gap
exists there — but if a future refinement makes the forced model ambiguous
between arms (e.g. a config override that repoints both arms to the same
model family), inference would stop being a faithful proxy. Consider adding
an explicit `arm: Option<String>` field to `SweepInfo` at that point (today
it would require updating ~40 `SweepInfo { .. }` struct-literal call sites
across the daemon, which is why #4809 did not do it).

## Acceptance criteria

- [ ] `WorkItem` carries the issue body (or the extracted complexity tier)
      through to dispatch.
- [ ] `resolve_autonomous_dispatch_model`'s three call sites pass the real
      per-issue complexity stratum instead of `None`.
- [ ] Existing `assign_arm` determinism/stratification tests still pass
      unmodified.
- [ ] (Optional, lower priority) `SweepInfo` carries an explicit `arm` field
      if inference-from-model is found to be insufficiently precise in
      practice.

rjwalters:
## Summary

Follow-up to #4809, which moved model-cost A/B experiment arm assignment into the
daemon but always called `assign_arm(issue, None)` — stratifying the entire
population as `routine`. Assignment stayed deterministic and resume-safe, but the
`complex` and `routine` strata did not each get an independent ~50/50 A/B balance;
only the pooled population did.

This threads the Curator's `<!-- loom:complexity=<tier> -->` marker through to all
three `resolve_autonomous_dispatch_model` call sites.

## What changed

**`work_finder.rs`** — `WorkItem` gains `body: Option<String>`, populated from
`RestIssue::body` in `GhWorkSource::list_ready_issues`. The ETag-cached REST
listing already returns the body, so this costs **zero extra requests**.
`WorkItem::complexity()` extracts the tier, and it rides `PriorityCandidate`
through the multi-workspace global priority sort so pass 2's `dispatch()` has it
without re-fetching. `WorkDispatcher::dispatch` grows a `complexity: Option<&str>`
parameter.

**`epic_supervisor.rs`** and **`ipc.rs`**'s `dispatch_sweep` MCP handler — neither
has a cached issue body, so both fetch one via the new `fetch_issue_complexity`
(GraphQL `gh issue view`, REST fallback per #4472, mirroring
`resolve-tier-model.sh`). The fetch is **lazy**: the new
`resolve_autonomous_dispatch_model_lazy` invokes the closure only inside its
`experiment` branch, so `off`/`observe` dispatch — the default, and the
no-behavior-change contract #4809 established — makes zero extra `gh` calls and
adds zero rate-limit pressure (epic #4432). A failed fetch degrades to
`None`/`routine` and **never fails a dispatch**.

## Marker extraction and #4845

`extract_complexity_marker` mirrors the shell extraction **as of #4845** (which
landed on main while this was in progress): anchored to the canonical
`<!-- ... -->` comment form, last-match-wins, line-oriented like `grep`.

That anchoring is load-bearing here rather than incidental: **issue #4827's own
body** quotes `` `<!-- loom:complexity=<tier> -->` `` as literal example text, and
the pre-#4845 bare-substring match would have picked that empty match over the
real marker — exactly the bug #4840 reported.

Extraction was differential-checked against the shell one-liner
(`grep -oE '<!--[[:space:]]*loom:complexity=[a-z]*[[:space:]]*-->' | tail -1 | sed …`)
across the canonical, no-delimiter, half-delimiter, non-`[a-z]`, `[A-Z]`,
empty-tier, multi-marker-one-line, multi-marker-multi-line, and prose-mention
cases — all 10 agree.

## Item 2 (`SweepInfo.arm`) — deliberately deferred

The issue marks it optional/lower-priority, and it is **not** done here. Inference
from the dispatched model stays faithful under this change: the complexity stratum
only shifts *which* arm an issue lands in, never the arm → model mapping that
`infer_arm_from_model` reads back. Doing it would require touching ~40
`SweepInfo { .. }` struct-literal call sites for no behavioral gain today. Left as
a follow-up for if/when a config override makes the forced model ambiguous between
arms.

## Testing

- `cargo test` in `loom-daemon/`: **2955 passed**. The existing `assign_arm`
  determinism/stratification tests pass **unmodified** (AC #3).
- New coverage: marker extraction (canonical form, anchoring, `[a-z]*` class,
  empty tier, last-match-wins, prose-mention/#4840 regression, no cross-line
  match, Unicode safety); `WorkItem::complexity()`; complexity reaching
  `dispatch()` on both the single- and multi-workspace tick paths; the stratum
  shifting the resolved arm; and the laziness contract (closure never invoked in
  `off`/`observe`, invoked exactly once in `experiment`).
- `cargo clippy --lib` clean, `cargo fmt --check` clean.

### Pre-existing failures on main (not introduced here)

- `tokens_pool::bootstrap::bootstrap_partial_triple_warns_without_aborting` fails
  only under the full parallel run and **passes in isolation** — an env-var
  parallelism race. This branch touches nothing in `tokens_pool`.
- `cargo clippy --all-targets` errors on `unnecessary_get_then_check` in
  `terminal.rs:2590` (test code, from #4808 / commit `df116564`), untouched here.
  `cargo clippy --lib` — all production code including these changes — is clean.

Closes #4827

### DIFF
--- a/loom-daemon/src/epic_supervisor.rs
+++ b/loom-daemon/src/epic_supervisor.rs
@@ -1460,9 +1460,25 @@ pub mod forge {
             // forced arm model when the workspace resolves to `experiment` mode
             // (CANARY-gated) — see `work_finder::dispatch` for the full
             // rationale; `off`/`observe` modes are unaffected.
+            //
+            // Issue #4827: unlike the work finder, the epic supervisor has no
+            // cached issue body to read the `<!-- loom:complexity=... -->`
+            // stratum from, so it fetches one — but LAZILY, inside
+            // `resolve_autonomous_dispatch_model_lazy`'s `experiment` branch,
+            // so `off`/`observe` dispatch still makes zero extra `gh` calls.
+            // A failed fetch resolves to `None` (the unchanged `routine`
+            // stratum) and never fails the dispatch.
             let repo_root = reg.config().workspace_root.clone();
-            let resolved =
-                crate::sweep_registry::resolve_autonomous_dispatch_model(&repo_root, child);
+            let gh_bin = reg
+                .config()
+                .gh_bin
+                .clone()
+                .unwrap_or_else(|| PathBuf::from("gh"));
+            let resolved = crate::sweep_registry::resolve_autonomous_dispatch_model_lazy(
+                &repo_root,
+                child,
+                || crate::sweep_registry::fetch_issue_complexity(&gh_bin, &repo_root, child),
+            );
             match resolved.arm {
                 Some(arm) => log::info!(
                     "epic_supervisor: dispatching child #{child} for epic #{_epic} with \
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -2864,10 +2864,29 @@ fn handle_request(
             // considers the model-cost A/B experiment's forced arm — mirroring
             // the autonomous work-finder / epic-supervisor dispatch paths —
             // before falling back to `autonomous.model` / the shipped default.
+            //
+            // Issue #4827: the arm is stratified by the issue's real
+            // `<!-- loom:complexity=... -->` marker. Like the epic supervisor
+            // (and unlike the work finder, which carries the body on its
+            // `WorkItem`), this handler has no cached body — so the fetch is
+            // LAZY, running only inside the `experiment` branch. `off` /
+            // `observe` dispatches make zero extra `gh` calls, and a failed
+            // fetch degrades to the unchanged `routine` stratum.
+            let gh_bin = sr
+                .config()
+                .gh_bin
+                .clone()
+                .unwrap_or_else(|| std::path::PathBuf::from("gh"));
             let (resolved_model, model_source_label, arm) = match (&kind, model.as_deref()) {
                 (crate::types::SweepKind::Issue(issue), None) => {
-                    let resolved = crate::sweep_registry::resolve_autonomous_dispatch_model(
-                        &repo_root, *issue,
+                    let resolved = crate::sweep_registry::resolve_autonomous_dispatch_model_lazy(
+                        &repo_root,
+                        *issue,
+                        || {
+                            crate::sweep_registry::fetch_issue_complexity(
+                                &gh_bin, &repo_root, *issue,
+                            )
+                        },
                     );
                     (resolved.model, resolved.source_label, resolved.arm)
                 }
--- a/loom-daemon/src/script_helpers/sweep_experiment.rs
+++ b/loom-daemon/src/script_helpers/sweep_experiment.rs
@@ -217,6 +217,110 @@ pub fn resolve_effective_mode_default(
 // Deterministic, resume-safe, stratified arm assignment
 // --------------------------------------------------------------------------
 
+/// The marker's key inside the `<!-- ... -->` comment, including the `=`.
+const COMPLEXITY_MARKER_KEY: &str = "loom:complexity=";
+/// The opening delimiter of the canonical HTML-comment marker form.
+const COMMENT_OPEN: &str = "<!--";
+/// The closing delimiter of the canonical HTML-comment marker form.
+const COMMENT_CLOSE: &str = "-->";
+
+/// Extract the Curator's `<!-- loom:complexity=<tier> -->` tier from an issue
+/// body (issues #3702/#4238/#4448), so a dispatch-time caller can hand a REAL
+/// stratum to [`assign_arm`] instead of the `None` → `routine` default (#4827).
+///
+/// Mirrors `resolve-tier-model.sh` / `require-complexity-marker.sh`'s extraction
+/// as of #4845:
+/// `grep -oE '<!--[[:space:]]*loom:complexity=[a-z]*[[:space:]]*-->' | tail -1`.
+/// Two properties matter and are both reproduced here:
+///
+/// 1. **Anchored to the canonical `<!-- ... -->` comment form.** A bare
+///    `loom:complexity=` substring also fires on prose that merely *discusses*
+///    the marker syntax — e.g. this very issue (#4827) quotes
+///    `` `<!-- loom:complexity=<tier> -->` `` as literal example text. The `<`
+///    after `=` breaks the `-->` anchor, so such mentions are skipped rather
+///    than yielding an empty match (#4840).
+/// 2. **The LAST match wins**, matching where the real marker is conventionally
+///    placed (at the end of the body).
+///
+/// Like `grep`, matching is line-oriented: a marker split across a newline is
+/// not recognized by either implementation. An empty tier
+/// (`<!-- loom:complexity= -->`) yields `None`, matching the shell's empty-`tier`
+/// fall-through to `routine`.
+///
+/// Deliberately does NOT validate against the closed
+/// `mechanical|routine|complex` vocabulary: [`normalize_complexity`] already
+/// folds every out-of-vocabulary value to `routine`, so an extra check here
+/// would only duplicate that behavior (and could drift from it).
+#[must_use]
+pub fn extract_complexity_marker(body: &str) -> Option<&str> {
+    // `tail -1` over `grep`'s document-order match stream: scan lines back to
+    // front, and within a line take that line's last match.
+    body.lines().rev().find_map(last_marker_in_line)
+}
+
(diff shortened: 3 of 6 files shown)


rjwalters:
## Judge review — APPROVE

Reviewed in a live checkout of the PR branch (`.loom/worktrees/issue-4827`, 6cac785d on top of origin/main), focusing on the three areas flagged in the Builder's report.

### 1. `extract_complexity_marker` (sweep_experiment.rs)

Read the implementation line-by-line: `last_marker_in_line` scans left-to-right for `<!--`, and `parse_marker_body` only accepts a match if `loom:complexity=[a-z]*` is immediately followed by `-->` (after optional whitespace) — so a prose mention like `` `<!-- loom:complexity=<tier> -->` `` fails to parse (the literal `<` after `=` breaks the `[a-z]*` + `-->` anchor) and is correctly skipped. `extract_complexity_marker` then folds lines in reverse and returns the first (i.e., last-in-document) match, giving `tail -1` semantics.

Diffed this against the shell implementations directly:
```
defaults/scripts/resolve-tier-model.sh:82:  grep -oE '<!--[[:space:]]*loom:complexity=[a-z]*[[:space:]]*-->' | tail -1 | sed ...
defaults/scripts/require-complexity-marker.sh:60: (same pattern)
```
Byte-for-byte the same anchored pattern and last-match-wins convention from #4845/698db138. The new test `extraction_ignores_prose_that_merely_mentions_the_marker_syntax` reproduces issue #4827's own body shape (prose mention before the real marker) and asserts the real marker wins — confirmed this is exactly the #4840 regression class, not hypothetical.

### 2. `resolve_autonomous_dispatch_model_lazy` (model.rs)

Confirmed the branch structure: `fetch_issue_complexity`/the caller's closure is invoked **only** inside the `if mode == "experiment"` arm (model.rs:320-330); the `off`/`observe` path falls straight through to the existing `resolve_dispatch_model` call with zero forge I/O. `epic_supervisor.rs` and `ipc.rs`'s `dispatch_sweep` handler both thread `fetch_issue_complexity(&gh_bin, &repo_root, issue)` in as the lazy closure — correct, minimal, symmetrical diffs in both call sites.

The test `autonomous_dispatch_lazy_complexity_is_only_fetched_in_experiment_mode` uses a `Cell<u32>` counter and asserts 0 invocations in `off`, 0 in `observe`, and exactly 1 in `experiment` (with the fetched value reaching `assign_arm`) — this is real coverage of the no-extra-forge-calls contract, not just a claim.

`work_finder.rs`'s `WorkItem::complexity()` reads the marker from a body already carried on the REST-listing `RestIssue` (confirmed `body` was already part of `forge_listing::RestIssue` pre-PR, from #4428) — genuinely zero extra requests, threaded through both the single-workspace and multi-workspace (`PriorityCandidate`) dispatch paths.

### 3. `assign_arm` determinism/stratification tests

Diffed `origin/main...HEAD` on the test module: no existing test in `sweep_experiment.rs` was modified, only new tests were added (`extracts_complexity_marker_from_issue_body`, `extraction_ignores_prose_that_merely_mentions_the_marker_syntax`, etc.). `arm_assignment_is_deterministic_and_stratified` and the canary/mode tests are byte-identical to pre-PR. AC #3 holds.

### 4. The two claimed pre-existing/unrelated failures

- `tokens_pool::bootstrap::bootstrap_partial_triple_warns_without_aborting`: ran it in isolation on the PR branch — passes. This PR's diff touches none of `tokens_pool` (confirmed via `git diff origin/main...HEAD --stat`), consistent with an env-var race under the full parallel run rather than a regression here.
- `unnecessary_get_then_check` clippy lint at `terminal.rs:2590`: reproduced it locally with `cargo clippy --all-targets` on the PR branch (`tm.terminals.get("ghost").is_none()` — test code). `git blame`/`git log` traces that line to df116564 (#4808), and this PR's diff does not touch `terminal.rs` at all. Confirmed unrelated.

### CI / merge state

- `Rust Linting & Build Check`: pass. `Rust Unit Tests`: pass (3m41s, matches the Builder's "2955 passed" claim).
- `Docs/Defaults Parity Check`: **fails**, but independently verified this is a **pre-existing main-branch break**, not something this PR introduces — main's own latest CI run (head `9ab43374`, this PR's own base before it advanced) fails the identical check, flagging `safehouse.md`/`token-pool.md` as orphans (a gap in #4848's new guard landing without an audit of pre-existing `.loom/docs/`-only files). This PR's diff touches none of the flagged paths. `mergeStateStatus` is `UNSTABLE` (non-required check failing), not `BLOCKED`, consistent with this being non-gating here — worth a separate follow-up issue against #4848/#4841 if one doesn't already exist, but out of scope for this PR.

Diff footprint is exactly the 6 files the Builder describes (`epic_supervisor.rs`, `ipc.rs`, `script_helpers/sweep_experiment.rs`, `sweep_registry/model.rs`, `sweep_registry/outcome_journal.rs`, `work_finder.rs`) — no scope creep.

Approving.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (another Judge approved this PR while my review was in progress) — another Judge's verdict raced mine. Discarding my review; not writing any label.

For the record, my independent review reached the same conclusion (approve): the diff correctly threads the Curator's `<!-- loom:complexity=<tier> -->` marker through all three `resolve_autonomous_dispatch_model` call sites, the laziness contract for `off`/`observe` modes is tested, the `WorkDispatcher::dispatch` trait-signature change updates both implementers, and local runs of the new/changed test modules (`sweep_experiment`, `work_finder`, `sweep_registry::model`) all pass. The only failing CI check (Docs/Defaults Parity Check) is pre-existing on main and unrelated to this PR's diff — filed as #4854.

PR #4866, opened 2026-08-01
fix(dashboard): make token-pool status proportional, not any-exhausted

rjwalters:
## Problem

`deriveStatus` in `dashboard/web/src/fleet.ts` marked a fleet host **Degraded** whenever *any* token account was exhausted (`tokens.exhausted > 0`). The multi-account token pool exists precisely so exhausted accounts are routine — the selector rotates away from them. This produced a real false alarm on 2026-08-01: two healthy hosts (`studio`, `robb-pro`) were flagged "Degraded" alongside one genuinely dead host (`loom-worker-1`), giving all three the same visual weight and camouflaging the real failure.

## Fix

Replace the any-exhausted check with a proportional `isTokenPoolDegraded` helper:

| Condition | Status |
|---|---|
| No report within `STALE_AFTER_SEC` | `stale` (unchanged) |
| `available == 0` (all accounts exhausted) | `degraded` |
| `available <= 1` or `exhausted / total >= 0.75` | `degraded` (genuinely close to the edge) |
| `total == 0` (no tokens reported yet) | not degraded by this check — unchanged `unknown`/`ok` handling upstream |
| otherwise | `ok` |

The "N need attention" headline (`FleetView.needsAttention`) is derived from `status` and required no separate change — it now only counts hosts that are genuinely `stale` or `degraded` under the new thresholds.

## Tests

- `dashboard/web/test/fleet.test.ts`: new `describe("isTokenPoolDegraded")` block covering all the threshold edges (0/1/2 available, the 75% ratio, and the `total === 0` no-data case), plus a regression case pinning "a 14-account pool with 5 exhausted renders `ok`" per the issue's acceptance criteria.
- `dashboard/web/test/fixtures.ts`: added `PARTIALLY_EXHAUSTED_HEALTHY_HOST_ID`, a host with a large pool (14 accounts, 5 exhausted) that stays well clear of both thresholds.
- Updated the host-count/order fallout from the new fixture host in `fleetOverview.test.ts`, `app.test.ts`, and `parse.test.ts`.
- Full `npm test` (341 tests) and `tsc --noEmit` pass clean.

## Scope note

[...]

PR #4867, opened 2026-08-01
fix(dashboard): make token-exhaustion status proportional, not any-nonzero

### LINKED ISSUE #4864: Dashboard flags every host 'Degraded' when any token account is exhausted — masked a genuinely dead host
## Problem

`deriveStatus` in `dashboard/web/src/fleet.ts` marks a host **Degraded** when *any* token account is exhausted:

```ts
} else if (tokens.exhausted > 0) {
  return "degraded";
```

The multi-account token pool exists precisely so that exhausted accounts are routine: the selector rotates away from them and keeps working. A pool with 5 of 14 accounts spent is operating exactly as designed. Flagging that as "Degraded" means **the fleet is degraded essentially all the time**, and the headline count reads "3 need attention" on a fleet where nothing needs attention.

## Why it matters — this produced a real false alarm

On 2026-08-01 the dashboard showed all three hosts flagged and "3 need attention". Two of them (`studio`, `robb-pro`) were entirely healthy — 12 and 14 sweeps in flight, 7 and 9 accounts available. The third, `loom-worker-1`, was genuinely dead: its daemon had exited 1h20m earlier (#4862) and it was running zero sweeps.

The badge gave all three the same visual weight, so the one real failure was camouflaged by two false positives. The operator's read of the dashboard was "degraded across all three hosts" — which is what the UI said, and it was wrong about two of them.

A signal that is always on cannot be used to find the thing that is actually broken.

## Suggested direction

Make the token axis proportional, and keep "needs attention" for states that actually need a human:

| Condition | Status | Rationale |
|---|---|---|
| No report within `STALE_AFTER_SEC` | `stale` | unchanged — this one is a real fault |
| `available == 0` | `degraded` | the pool cannot dispatch; work stops |
| `available <= 1` or `exhausted / total >= ~0.75` | `degraded` | genuinely close to the edge |
| otherwise | `healthy` | some accounts spent is normal rotation |

Two refinements worth considering while in there:

- **`loom-worker-1` had zero active sweeps and was not flagged for it.** A host reporting health but dispatching nothing — while its pool has capacity — is a stronger "needs attention" signal than any token count, and is exactly the shape of #4011's silent autonomy loss. Consider surfacing it.
- The exhausted count is still worth **showing** on the card (it is useful), just not worth **alarming** on. `TOKEN POOL 5/14 exhausted · peak 72%` is good information; the amber badge next to it is not.

## Acceptance criteria

- [ ] A host with a partially-spent but functioning pool renders `healthy`
- [ ] A host with no available accounts renders `degraded`
- [ ] A stale host still renders `stale` (unchanged)
- [ ] The "N need attention" headline counts only hosts a human should look at
- [ ] Existing `fleet.test.ts` status cases updated to the new thresholds, with a case pinning "partially exhausted pool is healthy"

## Affected files

`dashboard/web/src/fleet.ts` (`deriveStatus`, ~line 140), `dashboard/web/test/fleet.test.ts`, and the fixtures in `dashboard/web/test/fixtures.ts` that set up degraded hosts.

rjwalters:
## Problem

`deriveStatus` in `dashboard/web/src/fleet.ts` marked a host `degraded` whenever *any* token account was exhausted (`tokens.exhausted > 0`). The multi-account token pool exists precisely so exhausted accounts are routine rotation — flagging that as degraded meant the fleet read "degraded" almost constantly, and on 2026-08-01 it masked a genuinely dead host (`loom-worker-1`, daemon exited, zero active sweeps) among two false positives.

## Fix

Replaced the any-nonzero check with a proportional `isTokenPoolDegraded` helper:

| Condition | Status |
|---|---|
| No report within `STALE_AFTER_SEC` | `stale` (unchanged) |
| `available === 0` | `degraded` — pool cannot dispatch |
| `available <= 1` or `exhausted / total >= 0.75` | `degraded` — genuinely close to the edge |
| otherwise | `ok` — some accounts spent is normal rotation |

Also updated the `ok`/`degraded` status-badge tooltip copy in `fleetOverview.ts` to match the new semantics.

## Scope note

The issue also floated surfacing "zero active sweeps despite a healthy pool" (the #4011 silent-autonomy-loss shape) as its own degraded condition. I left that out: the module's own documented invariant is "zero sweeps is a normal state, not an empty state" (an idle host is healthy), and the view model has no signal to distinguish a genuinely idle host from a silently-stalled one — that needs a different data source (e.g. queue depth) to avoid trading one false-positive class for another. Happy to file a follow-up issue if desired.

## Tests

- `dashboard/web/test/fleet.test.ts`: added a fixture-backed case pinning "partially exhausted pool renders `ok`", plus inline synthetic-snapshot cases for `available === 0`, `available === 1` (below the 75% ratio but still degraded), and "no tokens.snapshot at all stays `ok`".
- `dashboard/web/test/fixtures.ts`: added `PARTIALLY_EXHAUSTED_HOST_ID` (1 of 5 exhausted → `ok`).
- Updated collateral host-count/order assertions in `fleetOverview.test.ts`, `app.test.ts`, `parse.test.ts` for the new fixture host.
- `npm run check` (typecheck + `vitest run`, 338 tests) passes in `dashboard/web/`.

Closes #4864

### DIFF
--- a/dashboard/web/src/fleet.ts
+++ b/dashboard/web/src/fleet.ts
@@ -30,15 +30,27 @@ import type { ActiveSweep, FleetSnapshot, HostEntry, TokenAccount } from "./type
 export const STALE_AFTER_SEC = 15 * 60;
 
 export type HostStatus =
-  /** Reporting recently, no token account exhausted. */
+  /** Reporting recently, with a functioning token pool. Some accounts may be
+   * exhausted — the pool rotates away from them by design, so that alone is
+   * not a fault. */
   | "ok"
-  /** Reporting recently, but at least one token account is exhausted. */
+  /** Reporting recently, but the token pool cannot dispatch (no accounts
+   * available) or is close enough to that edge to warrant a look. */
   | "degraded"
   /** Last report is older than `STALE_AFTER_SEC`. */
   | "stale"
   /** Known only from `activeSweeps`, or from a `hosts` entry with neither
    * `health` nor `tokens` yet — nothing to assess. */
   | "unknown";
+
+/**
+ * A pool this close to its edge needs a look even though it is not literally
+ * at zero: one more exhaustion away from `available === 0`, or three
+ * quarters of the pool already spent. Picked to fire well before the pool
+ * actually stalls, without firing on ordinary rotation (see #4864).
+ */
+const DEGRADED_AVAILABLE_FLOOR = 1;
+const DEGRADED_EXHAUSTED_RATIO = 0.75;
 
 export interface TokenSummary {
   /** The per-account rows, or `[]` for a public viewer, who is sent an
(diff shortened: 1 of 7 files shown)


rjwalters:
Closing as a duplicate: issue #4864 was already fixed and merged via #4866 (a concurrent sweep landed first — same `deriveStatus`/`isTokenPoolDegraded` fix, same test additions). This PR's mergeability is `CONFLICTING` against current `main` for that reason. No further action needed here; #4866 is the canonical fix.

PR #4796, opened 2026-07-31
docs: repo hygiene pass — fix stale MCP docs, READMEs, resync re-stamp

rjwalters:
Repo hygiene pass (/repo:all) — lands three local hygiene commits that had accumulated on this host plus today's fixes.

## What's in here

- **docs/mcp/README.md**: rewritten against the actual 30-tool mcp-loom surface — the old version listed 14 removed tools (`tail_daemon_log`, `read_console_log`, `trigger_factory_reset`, …) as current and omitted the Sweep Dispatch and Durable Watches categories entirely. Per-tool tables now defer to `mcp-loom/README.md` as the single source of truth.
- **mcp-loom/README.md**: tool count 27 → 30 (matches its own section sums: 6 UI + 13 terminal + 8 sweep + 3 watch, verified against `src/tools/*.ts`).
- **README.md**: MCP tool count 19 → 30.
- **scripts/README.md**: `.daemon.pid`/`.daemon.log` paths corrected to `.loom/` (where the scripts actually write), and nonexistent `pnpm run daemon:start` → `daemon:dev`.
- **assets/tarot-cards/README.md**: filenames corrected to the real SVGs (`worker.svg`/`reviewer.svg`/`critic.svg`/`fixer.svg` never existed), missing Champion (Strength) row added, and the claim that these assets are wired into the Loom UI removed (nothing consumes them).
- **loom-daemon/README.md**: new orientation README for the core crate (was the largest browsable dir with none).
- Earlier session's doc fixes (defaults/quickstarts/examples READMEs, safehouse + token-pool docs) and resync re-stamps.

## Verification

- All MCP tool names/params in the rewritten doc checked against `mcp-loom/src/tools/*.ts` schemas (`dispatch_sweep` takes `kind: {Issue: N}`, watch tools take `number`/`kind`/`repo`).
- Internal links validated by a full-repo link scan (0 broken).
- Docs-only + install-metadata stamp; no runtime code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4871, opened 2026-08-01
fix(daemon): publish Event::SweepPhase on sampled checkpoint phase transitions

### LINKED ISSUE #4863: sweep.phase telemetry is never emitted — every dashboard sweep shows 'starting' forever
## Symptom

Every active sweep on dashboard.2amlogic.com renders as **"starting"**, on every host, permanently. A sweep three hours into its Builder phase looks identical to one dispatched a second ago, which makes the fleet overview's per-sweep rows close to useless — you cannot tell progress from a stall.

Observed 2026-08-01 with 26 active sweeps across `studio` and `robb-pro`:

```
phase distribution: {'(none)': 26}
  studio:   {'(none)': 14}
  robb-pro: {'(none)': 12}
```

Every `activeSweeps[].phase` is `null`. The UI has no phase to show, so it falls back to "starting".

**The daemon knows the phase.** Same moment, `loom-daemon status --json` on `studio`:

```json
{ "sweep_id": "sweep-issue-4857-1785543227", "latest_phase": "curator-done", … }
```

## Root cause: nothing ever publishes `Event::SweepPhase`

The push pipeline is `Event::SweepPhase` → `observability::collector` → `TelemetryRecord::SweepPhase` → `/ingest` → Durable Object → `activeSweeps[].phase`.

The collector handles it correctly (`observability/collector.rs:223-229`) and has a passing unit test (`:525`). The problem is upstream: **`Event::SweepPhase` is never constructed anywhere in production code.**

Every construction in the tree is a test fixture:

- `serve.rs:1516` — `frozen_sweep_events()`, an SSE topic-freeze test
- `serve.rs:1680` — an `sse_frame` test
- `types.rs:2282,2320,2345,…` — event-taxonomy tests

How the daemon actually learns the phase (`sweep_registry/mod.rs:930-934`):

```rust
let checkpoint = …checkpoint_dir().join(format!("issue-{issue}.json"));
if checkpoint_written_by_run(&checkpoint, info.started_at) {
    info.latest_phase = read_checkpoint_phase(&checkpoint);
}
```

It **polls the sweep checkpoint file** and mutates `info.latest_phase` in place. That reaches `loom-daemon status` and the outcome journal, but it publishes no event — so the collector never sees a phase transition, and the telemetry stream carries none.

Confirmed against the wire. Last 200 records on the live deployment:

```
{'host.health': 45, 'tokens.snapshot': 45, 'sweep.outcome': 41, 'sweep.completed': 41, 'sweep.started': 28}
```

`sweep.phase` is defined in the schema (`.loom/docs/telemetry-schema.md`), typed (`telemetry/mod.rs:256,326`), redacted (`dashboard/src/redaction.ts` allowlists `phase`/`entered_at`), rendered, and **never emitted**.

## Fix

Publish `Event::SweepPhase { issue, phase, pr_number, repo }` where `latest_phase` changes — i.e. in the registry poll at `sweep_registry/mod.rs:933`, guarded so it fires on *transition* rather than on every poll tick (the poll re-reads the checkpoint continuously; emitting unconditionally would flood the event bus and D1).

Worth checking while in there: `latest_phase` values look like `"curator-done"`, whereas the schema's `sweep.phase.phase` and the UI's `SweepPhaseName` expect `curator|builder|judge|doctor|merge`. Either normalize on emit or widen the schema deliberately — the UI already types it as `SweepPhaseName | string`, so a mismatch degrades rather than breaks, but the two should agree on purpose.

## Acceptance criteria

- [ ] A sweep advancing through phases produces one `sweep.phase` record per transition, no duplicates while it sits in a phase
- [ ] `activeSweeps[].phase` on the dashboard tracks the sweep's real phase
- [ ] The per-sweep timeline view (#4750), which consumes `sweep.phase` for its phase-progression rows, shows real data — it is currently drawing from an empty stream
- [ ] The emitted `phase` value agrees with the schema's vocabulary
- [ ] The redaction test suite still asserts a private sweep's phase is retained while repo/issue/sweep_id are stripped

## Blast radius

This is not only the fleet overview. Anything downstream of `sweep.phase` has been reading an empty stream since Phase 1 shipped: the per-sweep timeline (#4750), phase-duration analysis, and any "which phase do sweeps stall in" question the epic was meant to answer (#4137's measurability gap).

## Affected files

`loom-daemon/src/sweep_registry/mod.rs` (emit site), `loom-daemon/src/observability/collector.rs` (already correct), plus a collector/registry integration test asserting the record reaches the queue from a real phase transition rather than a synthetic event.

rjwalters:
## Summary

`Event::SweepPhase` had **no production construction site** — every occurrence in
the tree was a test fixture (`serve.rs:1516`/`1680`, `types.rs:2282`+). The
downstream pipeline was already complete and unit-tested (collector mapping,
`SweepPhaseRecord`, redaction allowlist, dashboard rendering), but because
nothing upstream ever published the event, `sweep.phase` was a permanently empty
stream: `activeSweeps[].phase` was always `null` and every dashboard sweep
rendered as "starting" forever, on every host.

The daemon *did* know the phase — `SweepRegistry::sample_phase_transition`
samples the on-disk sweep checkpoint once per live sweep per `reap_once` tick —
it just kept that knowledge in memory (`phase_history`, `loom-daemon status`,
the outcome journal) and never published it.

This PR publishes `Event::SweepPhase` from that sampler.

## Changes

- **`loom-daemon/src/sweep_registry/outcome_journal.rs`** — `sample_phase_transition`
  now publishes `Event::SweepPhase { issue, phase, pr_number, repo }` via
  `emit_event` on each **newly observed** transition.
  - **Emit site choice**: this function already dedupes against `phase_history`.
    The checkpoint is re-read on *every* tick (the 30s reaper timer plus every
    read-path reap via `ListSweeps`/`GetSweepStatus`), so emitting from
    `overlay_live_phase` — or unguarded here — would republish the same phase for
    a sweep's entire residence in it and flood the bus and D1. Ordered after the
    history push so the dedupe state is committed even if a subscriber panics.
  - **Vocabulary**: the published `phase` is normalized through the existing
    `phase_label` (`"curator-done"` -> `"curator"`, `"judge-rejected"` -> `"judge"`),
    so the wire value matches the schema's documented
    `curator|builder|judge|doctor|merge`. The stored `PhaseObservation` keeps the
    raw marker, so the `sweep.outcome` per-phase breakdown is unchanged.
  - `repo: None` — stamped by `emit_event` from `config.workspace_root` (#3929),
    which is what lets the collector resolve the emitting repo's slug.
  - `pr_number` is carried through as `Option<i32>` via a checked conversion.
- **`loom-daemon/src/observability/collector.rs`** — `map_event_to_records` and
  `DispatchState` widened to `pub(crate)` so the registry's emit-site test can
  drive a *genuinely emitted* event through the real mapping. No behavior change.
- **`defaults/docs/telemetry-schema.md`** — documents the emit source, the
  transition-only cadence, the normalization rule, and that `entered_at` is an
  observation instant (polled source), not the true transition time.

| One `sweep.phase` record per transition, no duplicates while sitting in a phase | Yes | New `phase_transitions_publish_one_sweep_phase_event_each`: 5 `reap_once` ticks over 2 checkpoint changes (with 3 no-change ticks interleaved) drain exactly `["curator", "builder"]` off the bus |
| `activeSweeps[].phase` tracks the sweep's real phase | Yes | The DO's `sweep.phase` handler (`dashboard/src/fleetState.ts:130`) sets `phase` from the record; the record now exists. End-to-end covered by `a_real_phase_transition_reaches_the_telemetry_queue` |
| Per-sweep timeline (#4750) shows real data | Yes | `timelineBuilder.ts` aggregates `sweep.phase` records for a `sweep_id`; the same test asserts the queued record carries the dispatch-correlated `sweep_id`, not just the synthesized fallback |
| Emitted `phase` agrees with the schema vocabulary | Yes | New `emitted_phase_uses_the_schema_lifecycle_vocabulary`: markers `judge-rejected`/`doctor-done`/`merge-done` publish as `judge`/`doctor`/`merge` |
| Redaction suite still retains a private sweep's phase while stripping repo/issue/sweep_id | Yes | `dashboard/test/redaction.test.ts` unchanged and green — 41/41 pass, including "sweep.phase: keeps phase/entered_at, strips repo/issue/sweep_id" |
| Integration test uses a real transition, not a synthetic event | Yes | `a_real_phase_transition_reaches_the_telemetry_queue` builds a live registry, writes a checkpoint, calls `reap_once`, receives the event **off the bus**, and pushes it through `map_event_to_records` onto a `DurableQueue` |

## Test Plan

- `cargo test -p loom-daemon --lib sweep_registry::outcome_journal` — 18/18 pass
  (15 pre-existing + 3 new).
- `cargo test -p loom-daemon --lib` — 2956 passed, 9 failed. All 9 failures are
  **pre-existing and environment-dependent** (`work_finder`/`auto_update`
  "missing config file" tests pick up this host's user-level `~/.loom/config.json`);
  verified identical failures on `origin/main` in the primary checkout. None
  touch the changed code.
- `cargo test -p loom-daemon --bins` (65 pass), `cargo test -p loom-api --bins` (4 pass).
- `cargo check`, `cargo clippy -p loom-daemon --lib`, `cargo fmt --all -- --check` — all clean.
- `npx vitest run test/redaction.test.ts` in `dashboard/` — 41/41 pass.

- `cargo clippy --all-targets` fails on `loom-daemon/src/terminal.rs:2590`
  (`unnecessary_get_then_check` — `tm.terminals.get("ghost").is_none()`). Present
  on `origin/main`, unrelated to this change, left alone for scope.
- The collector's in-memory `issue -> sweep_id` correlation resets across a
  daemon restart (documented in its module header). A sweep already in flight
  when the daemon restarts will emit its phase records under the synthesized
  `unknown-issue-{N}` id, which the Durable Object keys as a separate active
  sweep. That is pre-existing collector behavior affecting `sweep.started` and
  `sweep.completed` equally; worth a follow-up issue rather than widening this PR.

Closes #4863

### DIFF
--- a/defaults/docs/telemetry-schema.md
+++ b/defaults/docs/telemetry-schema.md
@@ -156,6 +156,21 @@ A sweep advanced to a new lifecycle phase (mirrors `sweep.issue.{N}.phase`).
 
 `phase` is a lifecycle name: `curator`, `builder`, `judge`, `doctor`, `merge`.
 
+**Source and cadence (#4863)**: `Event::SweepPhase` — the record's only upstream —
+is published by `SweepRegistry::sample_phase_transition`, which the reaper calls
+once per live sweep per `reap_once` tick (the 30s reaper timer, plus every
+read-path reap). It reads `.loom/sweep-checkpoint/issue-<N>.json` and publishes
+**only when the phase changed since the previous tick**, so a sweep sitting in a
+phase emits nothing further. The checkpoint's raw marker (`curator-done`,
+`judge-rejected`) is normalized to the lifecycle vocabulary above before it is
+published; an unrecognized marker passes through verbatim rather than being
+guessed at, so `phase` should be read as "one of the five names, or an unknown
+raw marker" — which is exactly how the dashboard types it
+(`SweepPhaseName | string`).
+
+Because the phase is polled rather than pushed, `entered_at` is the daemon's
+*observation* instant, an honest upper bound on the real transition time.
+
 ### `sweep.completed`
 
 A sweep reached a terminal state (the summary moment; richer detail is in the
--- a/loom-daemon/src/observability/collector.rs
+++ b/loom-daemon/src/observability/collector.rs
@@ -55,8 +55,10 @@ use super::queue::DurableQueue;
 const SLUG_FETCH_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// In-flight sweep state tracked purely in memory (see module docs).
+/// `pub(crate)` only because it appears in [`map_event_to_records`]'s signature
+/// (#4863) — it is not part of any cross-module contract.
 #[derive(Debug, Clone)]
-struct DispatchState {
+pub(crate) struct DispatchState {
     sweep_id: String,
     started_at: DateTime<Utc>,
 }
@@ -188,7 +190,13 @@ fn event_repo_path(event: &Event) -> Option<String> {
 /// `repo` slug and `visibility`. Returns zero, one, or two records — a
 /// terminal event yields both a `sweep.completed` record (mirrors the frozen
 /// SSE moment) and the richer `sweep.outcome` record.
-fn map_event_to_records(
+///
+/// `pub(crate)` so the registry's own emit-site tests (#4863) can drive a
+/// *genuinely emitted* [`Event::SweepPhase`] through this mapping end-to-end
+/// instead of asserting against a hand-built event fixture — the exact gap that
+/// let `sweep.phase` be defined, mapped, and unit-tested here while never being
+/// published by production code.
+pub(crate) fn map_event_to_records(
     event: &Event,
     issue: u32,
     repo: &str,
--- a/loom-daemon/src/sweep_registry/outcome_journal.rs
+++ b/loom-daemon/src/sweep_registry/outcome_journal.rs
@@ -306,6 +306,21 @@ impl SweepRegistry {
     /// terminal record carry the sweep's own PR **without** a forge round trip
     /// — including for a successful sweep, whose checkpoint is deleted before
     /// the record is written.
+    ///
+    /// # Bus emission (Issue #4863)
+    ///
+    /// Every *newly observed* transition also publishes [`Event::SweepPhase`]
+    /// on the attached bus — the only production emit site for that variant, and
+    /// therefore the only source of the `sweep.phase` telemetry record the
+    /// observability collector maps it to. It lives **here**, rather than in
+    /// [`overlay_live_phase`](Self::overlay_live_phase), precisely because this
+    /// function already dedupes against `phase_history`: the checkpoint is
+    /// re-read on every tick, so an unguarded emit would republish the same
+    /// phase for the whole time a sweep sits in it and flood the bus (and D1).
+    /// The published `phase` is the **normalized** lifecycle name
+    /// ([`phase_label`] — `"curator-done"` → `"curator"`), matching the
+    /// `sweep.phase` schema's documented `curator|builder|judge|doctor|merge`
+    /// vocabulary, while the stored [`PhaseObservation`] keeps the raw marker.
     pub(crate) fn sample_phase_transition(
         &mut self,
         sweep_id: &str,
@@ -342,9 +357,18 @@ impl SweepRegistry {
             return;
         }
         history.push(PhaseObservation {
-            phase,
+            phase: phase.clone(),
             at: Utc::now(),
             pr_number,
+        });
+        // Genuine transition — publish it (see the "Bus emission" doc section
+        // above). Ordered after the history push so the dedupe state is already
+        // committed even if a subscriber panics on delivery.
+        self.emit_event(Event::SweepPhase {
+            issue: *issue,
+            phase: phase_label(&phase).to_string(),
+            pr_number: pr_number.and_then(|n| i32::try_from(n).ok()),
+            repo: None, // stamped by emit_event (#3929)
         });
     }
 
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix. `sample_phase_transition` was already computing and deduping phase transitions — this wires the missing bus publish into the existing dedupe boundary (after the `phase_history` push, so a panicking subscriber can't cause a re-emit) rather than adding a second tracking mechanism. The normalization through `phase_label` before publish keeps the wire vocabulary aligned with the documented `sweep.phase` schema while leaving the raw marker in `PhaseObservation`/`sweep.outcome` untouched — good separation of "what we observed" vs. "what we publish."

The new `a_real_phase_transition_reaches_the_telemetry_queue` test is the right shape for this bug class — it drives a real `reap_once` → real event → real `map_event_to_records` chain instead of a hand-built `Event::SweepPhase` fixture, which is exactly the kind of test that would have caught the original gap (schema, mapping, and redaction were all correct; only the emit site was missing).

## Test Execution

1. `cargo test -p loom-daemon --lib sweep_registry::outcome_journal` — ✅ Executed (via CI): Rust Unit Tests pass
2. `cargo test -p loom-daemon --lib` (2956 pass / 9 pre-existing env-dependent failures) — ✅ Executed (via CI): Rust Unit Tests pass
3. `cargo check`, `cargo clippy -p loom-daemon --lib`, `cargo fmt --all -- --check` — ✅ Executed (via CI): Rust Linting & Build Check passes
4. `npx vitest run test/redaction.test.ts` in `dashboard/` — ⚠️ Not independently re-run; covered by the PR's own stated 41/41 pass and unrelated to the changed files

## CI

All required checks pass. `Docs/Defaults Parity Check` fails but is **pre-existing on `main`** (verified by running `scripts/check-docs-defaults-parity.sh` directly against `origin/main` — identical `safehouse.md`/`token-pool.md` orphan/broken-link violations, unrelated to this PR's `outcome_journal.rs`/`collector.rs`/`telemetry-schema.md` changes). `mergeStateStatus` is `UNSTABLE` (non-required check), not `BLOCKED`.

rjwalters:

- CI Status: CI check 'Docs/Defaults Parity Check' is failing: `.loom/docs/safehouse.md` and `.loom/docs/token-pool.md` on `main` have no `defaults/docs/` counterpart (pre-existing gap on main, not introduced by this PR's diff — this PR only touches `loom-daemon/src/observability/collector.rs`, `loom-daemon/src/sweep_registry/outcome_journal.rs`, and `defaults/docs/telemetry-schema.md`). PR #4796 adds the missing `defaults/docs/safehouse.md` and `defaults/docs/token-pool.md` files; once that lands this check should clear on `main`.

**Next steps:**
- No action needed on this PR's own diff — the failure traces to a pre-existing gap on `main` (missing `defaults/docs/safehouse.md` and `defaults/docs/token-pool.md`), not this PR's changes
- Champion will retry automatically once the check is green (expected once PR #4796, which adds the missing `defaults/docs/` files, merges to `main`)

---
*Automated by Champion role*

PR #4892, opened 2026-08-01
fix(tokens-pool): dedup ranking rows by account name in build_monitor_accounts

### LINKED ISSUE #4873: Token pool ranking lists one account twice — 11 token files produce 12 rows, inflating capacity math
## Context

Found while auditing fleet health on 2026-08-01. `loom-daemon tokens check --ranking` lists **one account twice**, with different utilization figures on each row:

```
Account                        5h util   7d util Status
rjwalters-gmail                      -      0.14 available
agent8-2amlogic                   0.17      0.42 available
rjwalters-gmail                   0.75      0.42 available     <- same account, different numbers
...
Total 12: 7 available, 5 exhausted
```

The pool directory holds only **11** token files:

```
$ ls .loom/tokens/*.token | wc -l
11
$ loom-daemon tokens check --ranking | awk '{print $1}' | grep -E '^[a-z]' | sort | uniq -d
rjwalters-gmail
```

So 11 files produce 12 ranking rows. There is exactly one `rjwalters-gmail.token`.

Corroborated directly against this host's live `.loom/tokens/.ranking` (curator pass, 2026-08-01): it contains 12 lines for the 11 accounts in `index.json`, with `rjwalters-gmail` appearing in both the 2-field (no utilization) and 3-field (with utilization) line formats:
```
rjwalters-gmail|available
rjwalters-gmail|available|0.24
```
The 2-field line is `build_monitor_accounts`'s "unmentioned manifest account" fallback shape; the 3-field line is a matched claude-monitor row — i.e. this account is *both* being matched against `ranking.json` **and** falling through the fallback path in the same run, confirming two distinct code paths are each contributing a row for the same name (see Root Cause below).

## Why it matters beyond a cosmetic duplicate

The inflated count propagates into capacity decisions and operator-facing numbers:

- `loom-daemon status --json` → `capacity.total_accounts` (observed `13` on a host whose real pool is smaller), which feeds `token_axis_limit` and the `token_bound` decision — the scheduler believes it has a spare account it does not have.
- The observability `tokens.snapshot` record → the dashboard's `account_count` / `exhausted_count` aggregate, so the public fleet view reports a phantom account.
- The two rows disagree (`- / 0.14` vs `0.75 / 0.42`), so at least one is stale or misattributed — whichever the selector reads determines whether that account is considered usable.

## Root Cause (confirmed)

`loom-daemon/src/tokens_pool/monitor.rs`, `build_monitor_accounts` (lines 178-273), the default `--source auto` path (`check.rs:649-704` delegates to `monitor::run_monitor_check` first and only falls back to probing `.token` files when no fresh claude-monitor data exists — so the 11 `.token` files are not even consulted here):

- `load_index_email_map` (108-138) builds `Vec<(email_lower, name)>` from `index.json`, deduping only on identical **email**, not on `name`. Two different email strings that both resolve to the same Loom account name (e.g. a stale row left behind by a re-auth/rotation in claude-monitor's `usage.db`) survive as two separate entries.
- The main loop (212-241) iterates every row in claude-monitor's `ranking.json` `"accounts"` array, looks up the Loom name via that email map (218), and does `accounts.push(MonitorAccount { name, status, util_7d, util_5h })` (234) **unconditionally** into a plain `Vec` — nothing checks whether `name` was already pushed by an earlier iteration of this same loop.
- The only "already seen" bookkeeping is `matched: Vec<String>` (210, 233, 251), and it is consulted **only** by the second loop (243-260) that appends manifest accounts unmentioned by `ranking.json` — it does not prevent the *first* loop from pushing two `MonitorAccount`s for the same name when `ranking.json` itself contains two rows that map to one name.

Net effect: two `ranking.json` rows (or one `ranking.json` row plus the unmentioned-account fallback, matching the 2-field/3-field split observed on this host) resolving to the same Loom account name both survive, sort independently (267-271), and are handed downstream as two divergent records.

Nothing downstream re-dedupes: `format_table`/`format_ranking_lines` (`check.rs:709-740`, `monitor.rs:279-290`) emit one row/line per Vec entry (`Total {report.accounts.len()}` is inflated too), and `capacity::read_ranking_at` (`capacity.rs:180-210`) counts `.ranking` lines with `snap.total += 1` per line, no name-based dedup — that count flows into `capacity.total_accounts` via `ipc.rs:1595,1625-1634` and `types::CapacityReport` (`types.rs:1710-1715`), rendered at `cli/status_render.rs:99-110,190`.

(Ruled out: `tokens_pool/select.rs:220-233`'s `parse_ranking_line` — the shared `splitn(3, '|')` parser used by the selector and `capacity::read_ranking_at` — is correct and not implicated.)

## Acceptance Criteria

- [ ] `build_monitor_accounts` emits at most one `MonitorAccount` per unique Loom account name — the ranking row count equals the number of distinct accounts in the pool
- [ ] When `ranking.json` (or the email→name join) yields two candidate records for one name, the merge is deliberate and documented (which wins — e.g. most-recent/most-severe status — and why) rather than emitting both
- [ ] `capacity.total_accounts` and the `tokens.snapshot` `account_count` agree with the number of real accounts after the fix
- [ ] A regression test in `tokens_pool/monitor.rs`'s existing `mod tests` (see Affected Files) covers two `ranking.json` entries — or one `ranking.json` entry plus the unmentioned-account fallback — resolving to the same name, asserting a single output row

## Affected Files

- `loom-daemon/src/tokens_pool/monitor.rs` - `build_monitor_accounts` (lines 178-273): add name-keyed dedup (e.g. accumulate into an order-preserving map, or check-before-push against `matched`/an explicit `HashSet<String>`) in the first loop (212-241), consistent with the existing `matched.contains(name)` guard already used by the second loop (251-260); also consider whether `load_index_email_map` (108-138) should warn/dedupe when two emails resolve to one name
- `loom-daemon/src/tokens_pool/check.rs` - `format_table` (709-740) and `run_check` (649-704) consume `report.accounts`; no change expected once the upstream Vec is deduped, but the existing tests here (`run_check_writes_ranking_with_bad_account`, ~1086) are a reference for wiring the fixture format
- `loom-daemon/src/capacity.rs` - `read_ranking_at` (180-210) is the `.ranking` line-counter feeding `capacity.total_accounts`; no change expected once the upstream `.ranking` file is deduped, but its existing tests (`read_ranking_at_mixed_two_and_three_field`, ~689) show the fixture conventions for this reader

## Test Plan

- [ ] Automated tests: add a `tokens_pool/monitor.rs` unit test (alongside `build_accounts_orders_by_status_then_util` and `build_accounts_appends_unmentioned_manifest_accounts_last`, ~404-468) with a fixture where `ranking.json`/`index.json` produce two candidate records for one Loom name, asserting `build_monitor_accounts` returns exactly one `MonitorAccount` for that name
- [ ] Automated tests: extend or add a `format_ranking_lines`/`write_monitor_ranking_atomic` test confirming the written `.ranking` file has exactly one line per unique account for the same fixture
- [ ] Manual verification: on this host, after the fix, `loom-daemon tokens check --ranking` prints exactly 11 rows and `loom-daemon status --json | jq .capacity.total_accounts` reports `11`
- [ ] Edge cases: confirm the existing "unmentioned manifest account" behavior (issue #4645 — un-probed accounts still show as `available`) is preserved for accounts that are genuinely absent from `ranking.json` (as opposed to duplicated)

rjwalters:
## Summary

`build_monitor_accounts` could emit two `MonitorAccount` rows for the same
Loom account name — once from the `ranking.json` main loop, which pushed
unconditionally into a `Vec` with no dedup against already-seen names, and
once from the "unmentioned manifest account" fallback loop when a second
email for the same account wasn't matched. On a host with 11 real token
files this produced 12 ranking rows, inflating `capacity.total_accounts`
and the `tokens.snapshot` `account_count`.

## Changes

- `loom-daemon/src/tokens_pool/monitor.rs`, `build_monitor_accounts`: track
  each Loom account name's index in the output `Vec` via a `HashMap<String,
  usize>` built alongside the existing `matched: Vec<String>` bookkeeping.
  When a second `ranking.json` row resolves to a name already pushed, merge
  in place instead of appending a duplicate.
- **Merge rule (deliberate, documented in a code comment)**: the candidate
  with the more severe status wins (`status_rank`, higher = worse — mirrors
  the existing ordering vocabulary in `check.rs::status_rank`). Ties keep
  the earlier row. Rationale: a scheduler must never treat an account as
  available when another record for the same name says it is
  `rate_limited`/`exhausted`.
- The existing "unmentioned manifest account" fallback loop already guarded
  on `matched` (name-keyed), so once the first loop's fix keeps `matched`
  correctly deduped by name, that fallback loop needed no change — added a
  comment clarifying why it's already dedup-safe against this bug.
- Two new regression tests in `tokens_pool::monitor::tests`.

| `build_monitor_accounts` emits at most one `MonitorAccount` per unique Loom account name | ✅ | Name-keyed `HashMap` index gates every push in the main loop; `matched` (already name-keyed) gates the fallback loop |
| Dedup/merge logic is deliberate and documented (states which record wins and why) | ✅ | In-code comment above `index_by_name`: more-severe `status_rank` wins, ties keep the earlier row — because the scheduler must never mistake a rate-limited/exhausted account for available |
| `capacity.total_accounts` / `tokens.snapshot` `account_count` agree with the real account count after the fix | ✅ | Both read the same deduped `.ranking` file written from this `Vec` (`capacity::read_ranking_at`, `check.rs::format_table`/`format_ranking_lines`) — no downstream change needed once the upstream Vec is deduped; covered indirectly by the new tests asserting `accounts.len()` |
| Regression test covering two `ranking.json` entries (or one `ranking.json` + one unmentioned-manifest fallback) resolving to the same name, asserting a single output row | ✅ | `build_accounts_dedupes_two_ranking_rows_for_one_name` and `build_accounts_dedupes_ranking_row_plus_unmentioned_fallback`, both in `tokens_pool::monitor::tests` |

## Test Plan

- `cargo test --lib tokens_pool::monitor` — 32 passed, 0 failed (includes the
  two new regression tests)
- `cargo test --lib tokens_pool` — 238 passed, 0 failed (no regressions in
  sibling modules `check`, `capacity`-adjacent `select`, `paths`, etc.)
- `cargo check --lib` / `cargo clippy --lib -- -D warnings` — clean
- `cargo fmt` applied

Closes #4873

### DIFF
--- a/loom-daemon/src/tokens_pool/monitor.rs
+++ b/loom-daemon/src/tokens_pool/monitor.rs
@@ -208,6 +208,19 @@ pub fn build_monitor_accounts(
 
     let mut accounts: Vec<MonitorAccount> = Vec::new();
     let mut matched: Vec<String> = Vec::new();
+    // Index of `accounts` by Loom account name so a second `ranking.json` row
+    // that resolves to a name already pushed by an earlier iteration of this
+    // same loop updates that entry instead of appending a duplicate (issue
+    // #4873 — two rows for one account inflated `capacity.total_accounts`).
+    // This can happen when `load_index_email_map` legitimately carries two
+    // different emails for one Loom account (e.g. a stale row left behind by
+    // a re-auth/rotation) and `ranking.json` has a row for each. Merge rule,
+    // chosen deliberately: **the more severe status wins** (`status_rank`,
+    // higher = worse) — a scheduler must never treat an account as available
+    // when another record for the same name says it is rate_limited or
+    // exhausted; ties (equal severity) keep the earlier row.
+    let mut index_by_name: std::collections::HashMap<String, usize> =
+        std::collections::HashMap::new();
 
     if let Some(entries) = data.get("accounts").and_then(|a| a.as_array()) {
         for entry in entries {
@@ -230,13 +243,23 @@ pub fn build_monitor_accounts(
                 }
                 _ => (None, None),
             };
-            matched.push(name.clone());
-            accounts.push(MonitorAccount {
-                name,
+            let candidate = MonitorAccount {
+                name: name.clone(),
                 status,
                 util_7d,
                 util_5h,
-            });
+            };
+            if let Some(&idx) = index_by_name.get(&name) {
+                if status_rank(&candidate.status) > status_rank(&accounts[idx].status) {
+                    accounts[idx] = candidate;
+                }
+                // else: an earlier, equal-or-more-severe row already won —
+                // drop this duplicate.
+            } else {
+                index_by_name.insert(name.clone(), accounts.len());
+                matched.push(name);
+                accounts.push(candidate);
+            }
         }
     }
 
@@ -248,6 +271,12 @@ pub fn build_monitor_accounts(
     // dispatch-invisible). Their unknown utilization sorts them after
     // known-utilization `available` accounts via `UTIL_SENTINEL`, but they
     // still rank ahead of `rate_limited`/`exhausted` accounts.
+    //
+    // `matched` is keyed by Loom account **name**, not email, so this is
+    // already dedup-safe against `load_index_email_map` legitimately holding
+    // two different emails for one name (issue #4873): once the first loop
+    // above matches *either* email to a `ranking.json` row for that name, the
+    // name is in `matched` and neither email's fallback entry here fires.
     for (_, name) in &email_to_name {
         if !matched.contains(name) {
             matched.push(name.clone());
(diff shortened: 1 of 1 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T04:56:43Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T04:56:43Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verification against issue #4873's acceptance criteria:**

1. **At most one `MonitorAccount` per unique name** — verified. `index_by_name: HashMap<String, usize>` gates every push in the `ranking.json` main loop (merge-in-place instead of append on a repeat name); the existing `matched` (already name-keyed) guard continues to gate the "unmentioned manifest account" fallback loop unchanged.
2. **Deliberate, documented merge rule** — verified. In-code comment states the rule plainly: more-severe `status_rank` wins (reusing `check.rs::status_rank`, already imported into this file), ties keep the earlier row, with the rationale that a scheduler must never treat an account as available when another record for the same name says otherwise.
3. **`capacity.total_accounts` / `tokens.snapshot.account_count` agree with real count** — correctly reasoned as a no-op-downstream fix: both read the same deduped `Vec` via the `.ranking` file, so no change needed once the upstream `Vec` is deduped.
4. **Regression coverage for both duplicate shapes** — verified. `build_accounts_dedupes_two_ranking_rows_for_one_name` (two `ranking.json` rows for one name, asserts the more-severe status wins) and `build_accounts_dedupes_ranking_row_plus_unmentioned_fallback` (one `ranking.json` row + one falls to the unmentioned-manifest path, asserts the matched row wins and the genuinely-unmentioned `acct-z` is unaffected) — both map directly onto the two duplicate shapes the issue observed live (2-field vs 3-field `.ranking` lines).

**Local verification (in the existing builder worktree, `.loom/worktrees/issue-4873`):**
- `cargo test --lib tokens_pool::monitor` → 32 passed, 0 failed
- `cargo clippy --lib -- -D warnings` → clean, no warnings
- CI (`gh pr checks 4892`): all required checks pass (Rust Linting & Build Check, Rust Unit Tests, doc/budget checks); dashboard/installer/shell suites correctly skipped by `Detect Changes` since this is a Rust-only change
- `mergeStateStatus`: CLEAN

**Code quality notes (non-blocking):**
- The merge logic is correctly scoped to the *first* loop only — the second loop's preexisting `matched.contains(name)` guard already handled the fallback case, and the PR adds a clarifying comment rather than redundant logic. Good restraint.
- `std::collections::HashMap` is used via a fully-qualified path rather than a top-level `use` — purely stylistic, not worth blocking on.

No security implications (pure in-memory account-name dedup, no I/O or trust-boundary changes). Approving.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T04:56:43Z, age ~12m) — standing down without reclaiming. Not stomping.

Issue #4608, opened 2026-07-30
decide: port loom-search to Rust or retire it (the last Python in Loom, carved out of epic #4081 Phase 4)


rjwalters:
## Problem

Epic #4081 removed Python from Loom, and Phase 4 (#4557) deleted the `loom-tools` package — **except** one carve-out: `loom-search` (`loom_tools/semantic_search.py`, console script `loom-search`), which backs the opt-in, off-by-default semantic-search feature documented in `defaults/docs/semantic-search.md` (#4339, Tier B embeddings #4370).

It was carved out rather than deleted because **no epic #4081 phase ever decided its fate**. It appears in none of: the epic's "zero external references (delete, don't port)" inventory, any Phase 3 family, or Phase 1's tokens-only scope. And because the feature is opt-in and off by default, deleting it would have been a **silent feature removal with no test going red** — CI would have stayed green while a documented, shipped capability disappeared. Porting it was explicitly out of scope ("Phase 4 is deletion only, not another port").

So the decision was deferred, not made. This issue tracks making it.

## Current state (post-#4557)

> **Curator note (2026-07-30, refreshed): #4557 has merged.** It closed as completed (via PR #4611, merged 2026-07-30). This section now describes the confirmed state of `origin/main`, not a projection — verified directly by listing `origin/main`'s `loom-tools/` tree, which matches the carve-out below exactly (19 files, no more).

`loom-tools/` now contains only the carve-out:

```
loom-tools/
├── pyproject.toml                                  # loom-search only; zero runtime deps
├── src/loom_tools/
│   ├── semantic_search.py                          # the feature
│   ├── embedders.py                                # optional fastembed backend ([search] extra)
│   └── common/{config,config_resolver,repo}.py     # the 3 helpers it imports
└── tests/                                          # its tests + the #4039 conformance fixture
```

**Before #4557 merged**, `main`'s `loom-tools/` was larger than this: it still contained the entire `loom_tools/tokens/` package (13 files backing the live `loom-tokens` console script that CLAUDE.md's "Multi-Account Token Pool" section documents), a `loom_tools/models/` package, and several more `common/` helpers unrelated to `semantic_search.py`. PR #4611 deleted all of it. The "last Python in Loom" framing above is now accurate: `loom-search`'s carve-out is the only Python remaining in the repo.

[...]

Issue #4607, opened 2026-07-30
Wire defaults/scripts/check-cas-recheck-consistency.sh into .github/workflows/ci.yml's installer-tests job


rjwalters:
## Context

PR #4606 (closing #4570) added `defaults/scripts/check-cas-recheck-consistency.sh` — a regression guard asserting the Verdict-Time CAS Recheck sections (Judge + Doctor), the Champion Verdict-State Janitor, the criterion-1 fix, and the labels.yml mutual-exclusion invariant are all present.

The Builder could not wire it into `.github/workflows/ci.yml`'s `installer-tests` job (alongside `check-phantom-labels.sh` / `check-labels-drift.sh`) because the push token used lacked `workflow` OAuth scope — GitHub rejected the push with "refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope." That hunk was reverted; the script ships but currently only runs if invoked manually.

## What's needed

Add one step to the `installer-tests` job in `.github/workflows/ci.yml`, following the existing pattern (confirmed against `origin/main`, `.github/workflows/ci.yml:116-121`):

```yaml
      - name: Check verdict-label CAS-recheck consistency (#4570)
        run: bash defaults/scripts/check-cas-recheck-consistency.sh
```

Insert it after the existing `check-label-descriptions.sh` step (line 121), alongside the other `installer-tests` guard steps.

This requires a token/credential with `workflow` scope (a human maintainer, or a bot token provisioned with that scope) to push the `.github/workflows/ci.yml` change.

## Verification

The script already passes locally: `bash defaults/scripts/check-cas-recheck-consistency.sh` → exit 0. No script changes needed, just the CI wiring.

## Acceptance Criteria

- [ ] `.github/workflows/ci.yml`'s `installer-tests` job has a new step running `bash defaults/scripts/check-cas-recheck-consistency.sh`, inserted after the existing `check-label-descriptions.sh` step (line 121)
- [ ] The new step follows the exact naming/format pattern of its neighbors (`- name: ...` + `run: bash defaults/scripts/...`)
- [ ] CI's `installer-tests` job passes with the new step included
- [ ] No other `installer-tests` steps are reordered, renamed, or otherwise modified

## Affected Files

- `.github/workflows/ci.yml` — add one step to the `installer-tests` job (verified present on `origin/main`; job starts at line 108, existing guard steps at lines 116-121)

## Test Plan

[...]

PR #4899, opened 2026-08-01
fix: preserve loom:operator-only reroute during crash-path claim restore

### LINKED ISSUE #4887: Dispatch claim-abort race re-adds loom:issue after correct reroute to loom:operator-only (#4607, #4608)
## Problem

When a Builder's `--claim-owned` dispatch claims an issue (`loom:issue` → `loom:building`) but then aborts mid-build because the work requires credentials/scope the Builder doesn't have — correctly rerouting the issue to `loom:operator-only` — something re-adds `loom:issue` roughly 25-30 seconds later, with no accompanying comment. This leaves the issue in a contradictory state: `loom:issue` **and** `loom:operator-only` simultaneously.

This matters because the work-finder / daemon dispatch appears to select on `loom:issue` alone; it does not appear to hard-skip issues that also carry `loom:operator-only`. Left uncorrected, this reintroduces the issue into the Builder queue and it will hit the exact same credential/scope blocker again — an infinite reclaim loop.

## Evidence

Two independent, near-identical occurrences on 2026-07-31, ~6 minutes apart:

**Issue #4607** (`gh api repos/rjwalters/loom/issues/4607/events`):
```
20:42:47Z unlabeled loom:operator-only, labeled loom:issue   (promoted)
20:44:33Z labeled loom:building, unlabeled loom:issue         (Builder claimed)
20:45:58Z labeled loom:operator-only, unlabeled loom:building (aborted: needs `workflow` OAuth scope)
20:46:29Z labeled loom:issue                                  (+31s later, no comment — the bug)
```

**Issue #4608** (`gh api repos/rjwalters/loom/issues/4608/events`):
```
20:48:37Z unlabeled loom:operator-only, labeled loom:issue   (promoted)
20:49:35Z labeled loom:building, unlabeled loom:issue         (Builder claimed)
20:51:00Z unlabeled loom:building, labeled loom:operator-only (aborted: operator RETIRE decision)
20:51:25Z labeled loom:issue                                  (+25s later, no comment — the bug)
```

Both issues were left with `loom:issue` + `loom:curated` + `loom:operator-only` all present until Champion's label-state janitor pass manually stripped the stray `loom:issue` (see comments on #4607 and #4608).

## Likely root cause

Some unclaim / "return to queue" step in the sweep or daemon's claim-abort path unconditionally re-adds `loom:issue` when releasing a claim, without checking whether the issue was *just* rerouted to a terminal/blocked state (`loom:operator-only`, `loom:blocked`) by the same abort. The ~25-31s gap in both cases suggests an async cleanup step racing against (and losing to) the synchronous reroute-to-operator-only step.

## Suggested fix

Locate the claim-release / unclaim code path (likely in `loom-daemon`'s sweep dispatch or `.loom/scripts/worktree.sh`-adjacent cleanup, or `sweep.md`'s abort handling) that re-adds `loom:issue` on unclaim. Before re-adding `loom:issue`, re-read the issue's current labels and skip the re-add if `loom:operator-only` or `loom:blocked` is already present (i.e., a more authoritative reroute happened concurrently/since the claim was taken).

## Acceptance Criteria

- [ ] Root cause identified: the specific code path that re-adds `loom:issue` on claim abort/unclaim
- [ ] Fix: skip re-adding `loom:issue` when the issue already carries `loom:operator-only` or `loom:blocked` at unclaim time
- [ ] Regression test simulating the race (claim → abort-to-operator-only → unclaim) asserting `loom:issue` is NOT re-added
- [ ] Consider whether the work-finder should also hard-skip `loom:operator-only` issues even if `loom:issue` is present, as defense-in-depth (mirrors how Mode C pre-flight already hard-skips `loom:operator-only` PRs)

## Affected Files (starting points, not verified exhaustive — Builder should confirm)

- `loom-daemon/src/` — claim-owned dispatch / sweep abort handling
- `.loom/scripts/worktree.sh` or sweep checkpoint cleanup, if the unclaim step lives in shell rather than Rust

rjwalters:
## Summary

Fixes a claim-abort race where a Builder that correctly reroutes an issue
`loom:building` -> `loom:operator-only` (e.g. missing OAuth scope, or an
operator RETIRE decision) then exits, and ~25-31s later the reaper's
crash-recovery path re-adds `loom:issue` — leaving the issue in the illegal
`loom:operator-only` + `loom:issue` combo and re-queuing it for the exact
same blocker (an infinite reclaim loop).

## Root Cause

`SweepRegistry::restore_label_to_ready()` in
`loom-daemon/src/sweep_registry/guards.rs` is the reaper's crash-path
cleanup for a dead sweep child — it removes the stale `loom:building` claim
and restores `loom:issue`. It already had a carve-out for `loom:blocked`
(#4206) and PR-number targets (#4653), but had no carve-out for
`loom:operator-only`. When a Builder aborts and applies
`loom:operator-only` itself just before exiting, the reaper's later
crash-recovery call to `restore_label_to_ready()` raced that reroute and
unconditionally re-added `loom:issue` on top of it — exactly matching the
~25-31s gap and no-accompanying-comment pattern observed in #4607 and
#4608.

## Changes

- `loom-daemon/src/sweep_registry/stacking.rs`: factored the existing
  `issue_has_blocked_label` GraphQL probe into a shared
  `issue_has_label_via_graphql(issue, label)` helper, and added
  `issue_has_operator_only_label` on top of it (same fail-open-to-`false`
  semantics as the existing blocked probe).
- `loom-daemon/src/sweep_registry/guards.rs`: `restore_label_to_ready()` now
  probes `loom:operator-only` (only when `loom:blocked` doesn't already
  decide the outcome, to avoid a redundant `gh` call) and skips the
  `loom:issue` re-add when it is present — the stale `loom:building` claim
  is still removed unconditionally in every case.
- Added a regression test,
  `restore_label_to_ready_does_not_add_loom_issue_when_operator_only`,
  simulating the exact #4607/#4608 race (claim -> abort-to-operator-only ->
  crash-path unclaim) and asserting `loom:issue` is never re-added while
  `loom:building` is still removed.
- Updated `label_flips_run_in_registry_workspace_root`'s expected `gh`
  invocation count (4 -> 5) to account for the new probe call.

|---|---|---|
| Root cause identified | Yes | `restore_label_to_ready()` in `guards.rs` — reaper's crash-path claim restore, called from `reaper.rs` (dead/orphaned child cleanup) and `dispatch.rs` (2.5 dispatch-guard fail path) |
| Fix: skip re-adding `loom:issue` when `loom:operator-only`/`loom:blocked` present | Yes | New `operator_only` probe added alongside the existing `blocked` probe; `restore_label_to_ready()` now short-circuits on either |
| Regression test for the exact race | Yes | `restore_label_to_ready_does_not_add_loom_issue_when_operator_only` — fake `gh` reports `loom:blocked=false`, `loom:operator-only=true`; asserts the `--add-label loom:issue` call never happens and the `--remove-label loom:building` call still does |
| Work-finder defense-in-depth for `loom:operator-only` | Already present | `loom-daemon/src/work_finder.rs`'s `PARK_LABELS`/`SKIP_LABELS` already hard-skip a candidate carrying `loom:operator-only` even if `loom:issue` is also present (covered by existing test at `work_finder.rs` around `WorkItem::new(3, vec!["loom:issue".into(), "loom:operator-only".into()])` / `is_skipped()`) — no change needed here |

## Test Plan

- `cargo test --lib sweep_registry::guards` — 14 passed, 0 failed (includes the new regression test and the updated `label_flips_run_in_registry_workspace_root`)
- `cargo test --lib sweep_registry::stacking` — 2 passed, 0 failed
- `cargo test --lib sweep_registry::reaper` — 44 passed, 0 failed
- `cargo test --lib sweep_registry::dispatch` — 76 passed, 0 failed
- `cargo check --lib` — clean
- `cargo clippy --lib -- -D warnings` — clean

Closes #4887

### DIFF
--- a/loom-daemon/src/sweep_registry/guards.rs
+++ b/loom-daemon/src/sweep_registry/guards.rs
@@ -532,19 +532,31 @@ impl SweepRegistry {
 
     /// Restore a crashed/orphaned claim's `loom:building` back to
     /// `loom:issue` — UNLESS the issue currently carries `loom:blocked`
-    /// (Issue #4206) OR the target number actually resolves to a pull
-    /// request (Issue #4653). A deliberate operator park (applied by hand,
-    /// possibly while the now-dead sweep was still `loom:building`) must
-    /// never be clobbered into the illegal `loom:blocked` + `loom:issue`
-    /// combo by the crash-recovery path, and a PR number must never be
-    /// handed a `loom:issue` label meant for issues — that reproduces the
-    /// stray-label symptom from #4653's incident report, reached whenever the
-    /// 2.5 dispatch guard's fail-open window lets a PR number through and the
-    /// sweep is later cancelled or crash-recovered. Both checks are
-    /// best-effort and fail-open: an unverifiable read falls back to the
-    /// pre-#4206 unconditional restore (re-adding `loom:issue`), since a
-    /// stranded `loom:building` claim is the more common failure mode this
-    /// path exists to fix. Either carve-out only skips the `loom:issue`
+    /// (Issue #4206), `loom:operator-only` (Issue #4887), OR the target
+    /// number actually resolves to a pull request (Issue #4653). A
+    /// deliberate operator park (applied by hand, possibly while the now-dead
+    /// sweep was still `loom:building`) must never be clobbered into an
+    /// illegal `loom:blocked`/`loom:operator-only` + `loom:issue` combo by
+    /// the crash-recovery path, and a PR number must never be handed a
+    /// `loom:issue` label meant for issues — that reproduces the stray-label
+    /// symptom from #4653's incident report, reached whenever the 2.5
+    /// dispatch guard's fail-open window lets a PR number through and the
+    /// sweep is later cancelled or crash-recovered.
+    ///
+    /// The `loom:operator-only` carve-out closes the #4887 race: a Builder
+    /// that aborts mid-build (e.g. missing OAuth scope, or an operator RETIRE
+    /// decision) correctly re-routes the issue `loom:building` ->
+    /// `loom:operator-only` itself, then exits; when the reaper later notices
+    /// the dead child and calls this restore, an unconditional restore would
+    /// re-add `loom:issue` on top of that reroute ~25-30s later with no
+    /// accompanying comment, leaving the issue in the illegal
+    /// `loom:operator-only` + `loom:issue` combo and re-queuing it for the
+    /// exact same blocker.
+    ///
+    /// All checks are best-effort and fail-open: an unverifiable read falls
+    /// back to the pre-#4206 unconditional restore (re-adding `loom:issue`),
+    /// since a stranded `loom:building` claim is the more common failure mode
+    /// this path exists to fix. Every carve-out only skips the `loom:issue`
     /// re-add — the stale `loom:building` claim is always removed.
     pub(crate) fn restore_label_to_ready(&self, issue: u32) -> Result<()> {
         let gh = self
@@ -553,10 +565,14 @@ impl SweepRegistry {
             .clone()
             .unwrap_or_else(|| PathBuf::from("gh"));
         let blocked = self.issue_has_blocked_label(issue);
-        // Only probe PR-ness when the blocked carve-out doesn't already
-        // decide the outcome — avoids a redundant `gh` call on the (more
+        // Only probe `loom:operator-only` when `loom:blocked` doesn't already
+        // decide the outcome — avoids a redundant `gh` call in the (more
         // common) blocked path.
-        let is_pr = !blocked && self.issue_is_pull_request(issue).unwrap_or(false);
+        let operator_only = !blocked && self.issue_has_operator_only_label(issue);
+        let parked = blocked || operator_only;
+        // Only probe PR-ness when a park carve-out doesn't already decide the
+        // outcome — avoids a redundant `gh` call on the parked path.
+        let is_pr = !parked && self.issue_is_pull_request(issue).unwrap_or(false);
         let mut cmd = Command::new(&gh);
         cmd.arg("issue")
             .arg("edit")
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T05:45:07Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Correctness**: Verified the fix directly against `guards.rs`/`stacking.rs`. `restore_label_to_ready()` now probes `loom:blocked` first, then `loom:operator-only` only when `blocked` didn't already decide the outcome (avoiding a redundant `gh` call), and `is_pr` only when neither park label is present. All three carve-outs correctly skip the `--add-label loom:issue` while the `--remove-label loom:building` call is still issued unconditionally — exactly the semantics described in the PR body, and matches the two real call sites (`reaper.rs`, `dispatch.rs`).

**Refactor**: `issue_has_label_via_graphql(issue, label)` cleanly factors the shared GraphQL probe out of `issue_has_blocked_label`, with `issue_has_operator_only_label` built on top — no behavior change to the existing blocked-probe path, same fail-open-to-`false` semantics preserved (verified the timeout/error-handling branches were carried over unchanged, just re-worded for the generic label).

**Tests**: `restore_label_to_ready_does_not_add_loom_issue_when_operator_only` faithfully reproduces the #4607/#4608 race (fake `gh` reports `blocked=false`, `operator-only=true`) and asserts both that `loom:building` is still removed and `loom:issue` is never re-added. The `label_flips_run_in_registry_workspace_root` call-count update (4→5) is correctly justified by the new probe call, and its cwd-scoping assertion still holds for all 5 calls.

**CI/build**: `Rust Linting & Build Check` and `Rust Unit Tests` both pass (confirmed via `gh pr checks`); `mergeStateStatus` is now `CLEAN`. Local `cargo test --lib sweep_registry::guards` build was still compiling the dependency tree when I checked (cold cache) — not blocking given CI's green Rust Unit Tests run already covers this exact target.

No blockers. Nice minimal, well-targeted fix with a precise regression test.

Issue #4563, opened 2026-07-30
sweep: builder-invoked resync-installed.sh mutates the MAIN checkout mid-sweep — script-driven .loom/ writes escape the worktree guards


rjwalters:
## Context

During the 2026-07-30 `/loom:sweep all` run (`sweep-20260730T000350Z-31296-087c2c47`), the wave-2 per-wave `check-main-clean.sh --baseline --quarantine` pass caught 4 contaminating paths in the **main** checkout and quarantined them (stash `6e444445835a05980949d4e082ac59246972e09d`):

- `.loom/docs/advanced-hooks.md`
- `.loom/docs/ci-integration.md`
- `.loom/install-metadata.json`
- `loom-tools/uv.lock`

The likely source: a wave-2 Builder subagent (issue #4472's builder) ran `./.loom/scripts/resync-installed.sh` from inside its worktree. The resync script resolves and writes the **repo-root** `.loom/` installed copies (and re-stamps `install-metadata.json`), so a builder invoking it — even with good intentions ("refresh the installed copies to match my defaults/ edits") — mutates the main checkout, not its own worktree. `uv.lock` similarly suggests a `uv` invocation resolved against the main tree.

The Edit/Write worktree-confinement guard and the Bash write-idiom guard (#4178) do not cover this class: the write is performed by a repo script's own internals (cp/rsync inside `resync-installed.sh`), not by a recognizable Bash write idiom targeting main.

The backstop worked (quarantine, exit 4, main provably restored to baseline) — but the contamination source should be closed, not just cleaned.

## Suggested acceptance criteria

- [ ] `resync-installed.sh` refuses to run (or no-ops with a clear message) when invoked from inside a `.loom/worktrees/*` worktree, OR when the repo-root it resolves differs from the git common dir's primary checkout — unless an explicit override flag is passed.
- [ ] Builder role prompt (`builder.md`) documents: never run `resync-installed.sh` from a worktree; installed-copy propagation is the periodic resync commit's job (this convention is already stated in sweep summaries, but not enforced).
- [ ] A regression test covers the refusal path.

---

## Curator Enhancement

See the detailed curator comment on this issue for root-cause analysis (verified against `origin/main`), refined acceptance criteria, and duplicate/related-work check.

## Affected Files

- `defaults/scripts/resync-installed.sh` - add worktree-context detection + refusal/override flag near the existing `REPO_ROOT` resolution block (~line 140-165)
[...]

PR #4906, opened 2026-08-01
fix(hooks): require a write-confinement capability probe before deferring to the canonical guard

### LINKED ISSUE #4894: guard-destructive.sh dispatcher: repo#29 marker flip silently drops the Loom-only Bash write-confinement category (#4178) — needs an env/config override or the category must move out of the vendored guard
## Summary

`.loom/hooks/guard-destructive.sh` is a runtime dispatcher that hands off to the canonical Repo Skills guard whenever that guard carries the `repo#29` marker, and otherwise falls back to Loom's vendored `guard-destructive-generic.sh`:

```bash
if [[ -r "$CANONICAL_GUARD" ]] && grep -q 'repo#29' "$CANONICAL_GUARD" 2>/dev/null; then
    exec bash "$CANONICAL_GUARD"
fi
```

That handoff is unconditional on **capability**. It only probes for the `repo#29` fix — it never checks whether the canonical guard actually implements the categories the vendored copy is documented to own. One of them is Loom-only:

> **One Loom-specific exception:** the vendored copy also carries the Bash-tool **write-confinement** category (`>`/`>>` redirection, `tee`, `sed -i`, `cp`/`mv`, issue #4178) — see `guards.worktreeIsolation` below; this stays Loom-owned even though the rest of the file mirrors upstream.
> — `.loom/docs/guard-hooks.md:64`

So the moment a consumer repo installs a Repo Skills version whose canonical guard carries the marker, the dispatcher `exec`s it and the `guards.worktreeIsolation` Bash-tool write-confinement category **stops running entirely**, with no warning, no config surface, and nothing in CI that would catch it.

## How it surfaced

Repo Skills **0.7.0** added the `repo#29` marker to the canonical guard. In `rjwalters/kicad-tools`, the routine version-bump PR [rjwalters/kicad-tools#4563](https://github.com/rjwalters/kicad-tools/pull/4563) — a pure vendored-file refresh, byte-identical to the upstream tag, 16 checks green — therefore *also* silently flipped the effective Bash guard implementation. The Judge caught it on review and blocked the merge. The full review analysis is at https://github.com/rjwalters/kicad-tools/pull/4563#issuecomment (see the "Changes Requested" review on that PR).

| | `repo#29` in canonical guard | Guard that actually runs |
|---|---|---|
| Repo Skills 0.4.3 | absent (0 hits) | `.loom/hooks/guard-destructive-generic.sh` (vendored, **has** write-confinement) |
| Repo Skills 0.7.0 | present (1 hit) | `.claude/skills/repo/hooks/guard-destructive.sh` (canonical, **no** write-confinement) |

## Evidence

Grep for the category across both guards, at Repo Skills 0.7.0:

```
worktreeIsolation|write-confinement|WORKTREE_ISOLATION
  .loom/hooks/guard-destructive-generic.sh   -> 11 hits (deny emitted at line 2100)
  .claude/skills/repo/hooks/guard-destructive.sh (0.7.0) -> 0 hits
```

Behavioral confirmation, feeding an identical simulated `PreToolUse` payload to each guard from inside a Loom-managed worktree (`.loom/worktrees/pr-4563`) with a Bash-tool write targeting the main checkout:

```
payload: {"tool_name":"Bash","tool_input":{"command":"echo hi > <MAIN_CHECKOUT>/.loom/hooks/guard-destructive.sh"}}

dispatcher (-> canonical 0.7.0 guard) : exit 0, no output   => ALLOWED
vendored guard-destructive-generic.sh : permissionDecision "deny" => BLOCKED
```

The deny that disappears is exactly the #4178 one:

> `BLOCKED: Bash-tool write to '<path>' resolves to the main repository checkout ('<root>'), but a Loom-managed worktree exists elsewhere in this repository ... This is a worktree-isolation bypass via Bash redirection/tee/sed -i/cp/mv ... (#4178)`

Two properties make this worse than a normal capability gap:

1. **It is latent.** With no worktrees present, both guards no-op identically, so no check, test, or smoke run will ever flag it. It only bites mid-sweep, in exactly the window #4178 was filed to close.
2. **There is no workaround.** `guards.worktreeIsolation` defaults to `true` and `LOOM_GUARD_WORKTREE_ISOLATION` still reads as enabled — the category is "on", it simply never executes, because the process that implements it was never started. The dispatcher exposes no env or config knob to prefer the vendored copy.

The `Edit|Write` matcher (`guard-worktree-paths.sh`) is unaffected, so layer 1 survives. But the Bash-tool category exists *specifically because* layer 1 alone was bypassable — per `.loom/docs/guard-hooks.md`, the #4178 incident was sweep #4063 using a Bash-tool write to edit live guard hooks in the main checkout after the Edit/Write matcher denied it.

## Workaround currently in use

kicad-tools mitigated in-repo by wiring the vendored guard as an **additional** `PreToolUse`/`Bash` entry in `.claude/settings.json`, alongside the dispatcher:

```json
{ "type": "command", "command": "${CLAUDE_PROJECT_DIR}/.loom/hooks/guard-destructive.sh" },
{ "type": "command", "command": "${CLAUDE_PROJECT_DIR}/.loom/hooks/guard-destructive-generic.sh" },
{ "type": "command", "command": "${CLAUDE_PROJECT_DIR}/.loom/hooks/guard-loom-workflow.sh" }
```

Measured cost of the extra invocation: **~25 ms per Bash tool call**; denies from the doubly-run shared pattern set are idempotent (both guards emit `deny`, Claude Code honors it once). This works, but it is a per-consumer-repo patch that every Loom repo with Repo Skills installed now needs, and it will be silently clobbered by any settings.json re-merge that regenerates the Bash matcher from Loom's defaults.

Any one of these closes it; (2) is probably the cleanest long-term:

1. **Dispatcher honors an override** — e.g. `LOOM_GUARD_PREFER_VENDORED=1` (and/or a `guards.preferVendoredGuard` config key) forces the vendored path, documented in `guard-hooks.md`.
2. **Dispatcher chains rather than `exec`s the Loom-owned category** — run the canonical guard for the generic pattern set, then always evaluate the write-confinement category (extract it into its own `guard-worktree-writes.sh` that the dispatcher invokes unconditionally, so the generic-pattern set is never double-run and the ~25 ms tax disappears). This also ends the "vendored copy is both a fallback *and* a carrier of Loom-only behavior" dual role, which is what made the gap invisible.
3. **Capability probe instead of a version probe** — the dispatcher additionally greps the canonical guard for the write-confinement marker and falls back to the vendored copy when absent. Cheapest fix, but leaves the Loom-only behavior coupled to a file documented as a mirror of upstream.
4. **Upstream the category into Repo Skills** so the canonical guard carries it. Least attractive — it is Loom-worktree-specific semantics in a general-purpose tool.

Plus, regardless of which is chosen:

- **Regression test**: with a canonical-marker-bearing guard installed *and* a Loom-managed worktree present, a Bash-tool write into the main checkout must still produce a `deny`. This is the assertion that would have caught the flip; `.loom/hooks/tests/` is the natural home.
- **Doc update**: `.loom/docs/guard-hooks.md:64` currently states the exception as if it always holds. It should note the dispatcher interaction and whatever override lands.

Refs: #4178 (the category and the originating incident), #4041 (the dispatcher), #4245 (the same category's false-positive fix), rjwalters/repo#29 (the marker being probed), rjwalters/kicad-tools#4563 (where this surfaced).

rjwalters:
## Summary

`guard-destructive.sh`'s dispatcher deferred to the canonical Repo Skills
guard purely on a **version** probe: does it carry the `repo#29` curl-pipe
fix marker? That says nothing about whether the canonical guard actually
implements the Loom-only Bash-tool **write-confinement** category (issue
#4178, `guards.worktreeIsolation`). Once a consumer repo's canonical guard
picked up `repo#29` *without* that category (Repo Skills 0.7.0), the
dispatcher `exec`'d it and worktree-isolation enforcement for Bash writes
silently stopped — no warning, no override (kicad-tools PR #4563 caught this
in review; see #4894 for the full evidence and reproduction).

## Changes

- `defaults/hooks/guard-destructive.sh`: the dispatcher now requires **both**
  a version probe (`repo#29`) **and** a capability probe (the
  `worktree-write-confinement` decision tag the vendored guard's
  write-confinement category emits) before deferring to the canonical guard.
  Either probe failing falls back to the vendored guard, which always carries
  write-confinement.
- `install.sh` (`canonical_guard_present()`), `defaults/scripts/resync-installed.sh`
  (`CANONICAL_GUARD_PRESENT`), and `scripts/install/manifest.sh`
  (`_canonical_guard_present`) now use the identical two-probe check. These
  decide whether the vendored fallback is installed/kept on disk at
  all — without mirroring the dispatcher's stricter probe here too, a resync
  could still strip the vendored fallback out from under the dispatcher the
  moment a canonical guard picked up `repo#29` without write-confinement,
  leaving zero destructive-command coverage instead of the intended fallback.
- `defaults/docs/guard-hooks.md`: documents the two-probe handoff and why
  both are required (updates the "One Loom-specific exception" / "which
  generic guard is authoritative" sections).
- Regression tests added to `tests/hooks/test-guard-destructive-dispatcher.sh`
  (cases 6-7), `tests/install/test-hooks-preserve.sh` (test 7), and
  `defaults/scripts/tests/test-resync-installed.sh` (test group 15b) — all
  cover the exact "canonical has repo#29 but not write-confinement" shape.

| Dispatcher no longer defers to a canonical guard that lacks write-confinement | ✅ | `defaults/hooks/guard-destructive.sh` now greps for both `repo#29` and `worktree-write-confinement` before `exec`ing the canonical guard |
| Regression test: canonical-marker-bearing guard + Loom-managed worktree present → Bash-tool write to main checkout still denies | ✅ | `tests/hooks/test-guard-destructive-dispatcher.sh` Case 7 (`make_repo_gitwt` fixture with a `.loom-managed` worktree + a stub canonical guard carrying only `repo#29`); asserts `permissionDecision == "deny"` and that the stub canonical guard was NOT exec'd |
| Doc update at the "One Loom-specific exception" note | ✅ | `defaults/docs/guard-hooks.md` — "Dispatcher handoff is gated by TWO runtime probes, not one (#4894)" bullet + updated "which generic guard is authoritative" section |
| Existing hook test suites still pass (no regressions) | ✅ | see Test Plan below |

## Test Plan

All run from the issue worktree against the canonical `defaults/` sources:

- `bash tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 passed (cases 1-7 including the two new #4894 cases; cases 3 and 5 updated to carry both markers so they still represent "canonical genuinely offers equal coverage")
- `bash tests/hooks/test-guard-destructive.sh` — 566/566 passed (unmodified `guard-destructive-generic.sh`, confirms no regression)
- `bash defaults/scripts/tests/test-resync-installed.sh` — 114/114 passed (includes new test group 15b for the capability-gap fallback-preservation case)
- `bash tests/install/test-hooks-preserve.sh` — 10/10 passed (includes new test 7 for the same capability-gap case)
- `shellcheck -x` on all 4 modified `.sh` production files — no new warnings introduced (pre-existing unrelated warnings in `install.sh`/`resync-installed.sh` left untouched)

Closes #4894


rjwalters:
Code evaluation pending: CI is still running (`Installer Integration Tests` and `Shell Test Suites (hermetic)` are pending as of this pass). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Reviewed the two-probe dispatcher fix for #4894 in an isolated PR worktree.

**Correctness**: The change is exactly what it claims — `guard-destructive.sh`'s dispatcher, `install.sh`'s `canonical_guard_present()`, `resync-installed.sh`'s `CANONICAL_GUARD_PRESENT`, and `manifest.sh`'s `_canonical_guard_present` all now require **both** the `repo#29` version marker and the `worktree-write-confinement` capability marker before treating a canonical Repo Skills guard as equal-or-better coverage. All four call sites are consistently updated — no site was missed, which is exactly the gap this issue is about (a mismatch between the dispatcher's own runtime check and the install/resync-time "is the vendored copy skippable" checks).

**Testing**: Ran all four suites named in the PR's test plan directly (not just trusting the claimed numbers):
- `tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 pass, including new cases 6-7 for the capability-gap scenario
- `tests/install/test-hooks-preserve.sh` — 10/10 pass, including new test 7
- `defaults/scripts/tests/test-resync-installed.sh` — 114/114 pass
- `shellcheck -x` on all 4 modified production files — only pre-existing, unrelated warnings (SC2094 in resync-installed.sh, SC2086 in install.sh both predate this diff)

**CI**: All applicable checks pass (Shell Test Suites, Installer Integration Tests, Shellcheck, Docs/Defaults Parity, Codex Adapter Smoke); Rust/Dashboard/MCP jobs correctly skip since this PR touches none of that surface. Merge state is CLEAN.

**Issue linking**: `Closes #4894` present and correct.

Nice regression coverage — test 7 in `test-hooks-preserve.sh` and cases 6-7 in the dispatcher test directly encode the "canonical has repo#29 but not write-confinement" shape that caused the original silent coverage gap.

rjwalters:

- **Blast radius**: modifies `defaults/hooks/guard-destructive.sh` (the write-confinement dispatcher itself — changes which guard runs for every Bash-tool write across the fleet) plus `install.sh`, `defaults/scripts/resync-installed.sh`, and `scripts/install/manifest.sh` (the three call sites that decide whether the vendored fallback guard is installed/kept on disk at all). This is exactly the class of file criterion #2's blast-radius axis reserves for human review regardless of diff size or quality: a subtle regression in the new two-probe capability check could either reopen the exact write-confinement bypass this PR fixes (#4894) or strip the vendored fallback out from under the dispatcher during a routine resync, leaving zero destructive-command coverage fleet-wide.

The diff itself is narrow and well-scoped, and the Judge's review is thorough (cites the root cause, the two-probe fix, and 12/12 + 566/566 + 114/114 + 10/10 passing test suites across all four modified test files) — this hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4906`

---
*Automated by Champion role*

PR #4243, opened 2026-07-28
feat(tokens): add 5h load-aware eligibility gate to ranked token selection

rjwalters:
## Summary

Reconciles the gpeyton-fork's load-aware token selection proposal with upstream's existing rotation-cursor spread, per **Option A** of the curated decision in #4195: extend the `.ranking` format to carry each account's 5h-window utilization as an optional third field, and add a **soft eligibility gate** on top of the existing #3909 rotation cursor — *not* the fork's waterfall-fill replacement.

The account probe already measures 5h utilization but every ranking writer historically discarded it, so the load-blind rotation cursor happily dispatched into an account already near its 5h limit. This closes that gap while preserving #3909's even-drain rotation.

**Design source** (not cherry-picked — they patch a divergent JSON ranking format that does not exist upstream): fork commits `283de8e3` (waterfall fill) and `20961dd9` (5h-headroom spread). We deliberately take only the *load signal*, layered as a gate on the rotation cursor, because the cursor already prevents the intra-burst stacking that the fork's uncited "synthetic per-spawn bump" was needed to avoid.

Implemented in **both languages in the same PR** (required — the two are kept byte-identical by `test_rust_conformance.py`).

## Changes

- **Format**: `.ranking` lines are now `name|status|5h_util` (fraction, 2 decimals). The third field is optional — accounts with no measured 5h utilization keep the legacy 2-field form (never faked as `0.0`), and existing 2-field files parse unchanged.
- **Writers (all four)**: `check.py`/`monitor.py` and Rust `tokens_pool::check`/`tokens_pool::monitor` emit the field via a shared `ranking_line` helper.
- **Readers (both)**: `select.py:_read_ranking` and `select.rs:read_ranking` parse 2- or 3-field lines tolerantly; a missing/unparseable utilization → `None` ("unknown").
- **Gate**: the tier-1 preferred pass excludes accounts with `5h_util >= threshold`; the fallback pass drops the gate so a fully-loaded pool still dispatches. Threshold defaults to `0.70`, overridable via `LOOM_TOKEN_5H_LOAD_GATE` (`> 1.0` disables), mirrored in `select.py:_resolve_load_gate` / `select.rs:resolve_load_gate`.
- **Docs**: `.loom/docs/token-pool.md` documents the format extension, the gate, and the Option-A decision.

## Acceptance Criteria Verification

[...]

PR #4912, opened 2026-08-01
feat(tokens): report when an exhausted account's limit window resets

### LINKED ISSUE #4874: Exhausted accounts never report limit_window_reset_at — operator cannot tell when pool capacity returns
## Context

Found while auditing fleet health on 2026-08-01. Across the whole fleet, **no exhausted account ever reports when its limit window resets.**

`loom-daemon tokens check --ranking` — the `7d resets` column is empty for every account, including the five sitting at `7d util = 1.00`:

```
Account                        5h util   7d util Status        7d resets
robb-2amlogic                     0.00      1.00 exhausted     -
robb-integratedplasmonics         0.00      1.00 exhausted     -
agent1-2amlogic                   0.00      1.00 exhausted     -
agent2-2amlogic                   0.00      1.00 exhausted     -
agent4-2amlogic                   0.00      1.00 exhausted     -
```

Same story in the pushed telemetry — 16 exhausted accounts fleet-wide, no reset time anywhere:

```
loom-worker-1   exhausted=6/13  next_reset=None
robb-pro        exhausted=5/14  next_reset=None
studio          exhausted=5/12  next_reset=None
```

## Why it matters

`limit_window_reset_at` is a defined field on `tokens.snapshot` in `.loom/docs/telemetry-schema.md`, and the whole chain downstream of it is built and working — it is just never fed:

- `format.ts`'s `formatCountdown` exists specifically for this field, and its doc comment explains the design: *"a future value is the normal case, and clamping it would hide exactly the number an operator is looking for ('when does this exhausted account come back?')"*. That question is currently unanswerable.
- The host-detail token table renders a reset column that is always `—`.
- `deriveTokenPoolAggregate` computes `next_limit_window_reset_at` as the earliest reset across the pool (#4864-era work); it is always `null`, so the public view's fleet-level "capacity returns at" is dead.

The operational consequence: with ~40% of the pool exhausted, an operator cannot distinguish "capacity returns in 20 minutes" from "this fleet is stalled until tomorrow", and cannot plan dispatch around it.

## Not yet established

Whether the reset instant is unavailable from the upstream source (the probe / claude-monitor `usage.db`), parsed but dropped, or simply never populated on the record. Worth checking the probe path first — if the source genuinely does not expose it, the honest fix may be to derive an estimate from the window start and say so, or to remove the dead field rather than render a permanent `—`.

- [ ] An exhausted account reports a `limit_window_reset_at`, or there is a documented reason it cannot and the UI stops implying it might
- [ ] `deriveTokenPoolAggregate`'s `next_limit_window_reset_at` is non-null when any account reports one
- [ ] The host-detail reset column shows a real countdown for an exhausted account
- [ ] `formatCountdown`'s future-value path gets exercised by a real value, not only by its unit test

rjwalters:
Closes #4874

## Root cause

`limit_window_reset_at` was a defined `tokens.snapshot` field with a fully built consumer chain — `formatCountdown`, the host-detail reset column, `deriveTokenPoolAggregate`'s `next_limit_window_reset_at` — that **nothing ever fed**. It was parsed and then dropped, in three places:

1. **`.ranking` had nowhere to put it.** The pipe format was `name|status|5h_util`, so `check::ranking_line` discarded the `anthropic-ratelimit-…-7d-reset` value the probe had *already parsed* and threaded all the way to the CLI table. `.ranking` is the only channel between the probe and everything downstream, so the value died at the writer.
2. **The claude-monitor backend never looked.** `build_monitor_accounts` read `email`/`status`/`utilization` from `~/.claude-monitor/ranking.json` and hardcoded `s7d_reset: None` — even though that file carries `accounts[].resets.{5h,7d}` for every account. This is why the CLI's reset column was empty on every monitor-sourced run: not missing data, unread data.
3. **The collector hardcoded `None`.** `sample_token_snapshot` built every `TokenAccountState` with `limit_window_reset_at: None`.

## Fix

`.ranking` grows an **optional fourth field**, following the `5h_util` precedent from #4195 — legacy 2- and 3-field lines still parse, and an absent value stays *unknown* rather than being fabricated. Both writers populate it; the collector parses it into the telemetry record. `parse_ranking_line` becomes `splitn(4, '|')` and returns a named `RankingRow`, so a future fifth field cannot silently shift a positional binding the way the #4243/#4344 drift did.

### The field is the *binding* window's reset, not always the 7d one

This is the part that turned out to be more than plumbing. `check::limit_reset` is the single place the choice is made:

| Status | Reset written | Why |
|---|---|---|
| `exhausted` | 7d | `exhausted` *is* 7d utilization ≥ threshold; the 5h window rolling over releases nothing. |
| `rate_limited` | 5h | A 429 with 7d utilization *below* threshold — the 5h window is what tripped. |
| anything else | 5h | Not gated at all; the 5h rollover is what the reported `usage_fraction` is racing. |

Always writing the 7d reset would have answered the countdown's own question wrongly, in the direction that makes the fleet look dead. Live proof from this host's probe — `agent7-2amlogic` is `rate_limited` at 7d util 0.79, so its 7d reset is days out, but it actually comes back in ~20 minutes:

```
Account                        5h util   7d util Status        Resets at
agent6-2amlogic                   0.75      0.73 available     2026-08-01T09:50:00Z (5h)
agent7-2amlogic                   1.00      0.79 rate_limited  2026-08-01T08:50:00Z (5h)
robb-2amlogic                     0.00      1.00 exhausted     2026-08-02T03:00:00Z (7d)
robb-integratedplasmonics         0.00      1.00 exhausted     2026-08-04T11:00:00Z (7d)
agent3-2amlogic                   0.00      1.00 exhausted     2026-08-04T23:00:00Z (7d)
```

Choosing the binding window also keeps the value coherent with `usage_fraction` (which is 5h utilization) for `burn.ts`'s segmentation and `forecast.ts`'s margins, which both assume the reset is the rollover that drops the charted fraction. When the binding window's instant is unknown it stays absent — never substituted with the other window's, because an absent countdown reads as "unknown" and a wrong one reads as a fact.

The CLI's reset column now reports that same value and **names its window** (`2026-08-02T03:00:00Z (7d)`), replacing a `7d resets` header that was empty on every monitor-sourced run.

## Acceptance criteria

- [x] An exhausted account reports a `limit_window_reset_at` — verified live (table above), and the written `.ranking` carries `robb-2amlogic|exhausted|0.00|2026-08-02T03:00:00Z`.
- [x] `deriveTokenPoolAggregate`'s `next_limit_window_reset_at` is non-null when any account reports one — now exercised end-to-end through `redactPayload`, not just the unit.
- [x] The host-detail reset column shows a real countdown for an exhausted account — `in 2d 14h`, asserted in `hostDetail.test.ts`.
- [x] `formatCountdown`'s future-value path is driven by a real value — the `multiHostSnapshot` fixture's exhausted account now carries a reset.

## Testing

- `cargo test --lib` — **2988 passed, 0 failed**.
- `cargo clippy --package loom-daemon` (CI's invocation) clean; `cargo fmt --all --check` clean; `cargo check --workspace --all-targets` clean.
- `dashboard` Miniflare suite — **161 passed**; `dashboard/web` happy-dom suite — **372 passed**; both `npm run typecheck`s clean.

New coverage: writer/parser round-trips for all four field layouts including the `name|status||reset` placeholder case; `limit_reset`'s window selection *and* its never-substitute rule; the probe's opportunistic `-5h-reset` header (present and absent); the monitor backend against a fixture reduced from this host's real `ranking.json`; a writer → file → `sample_token_snapshot` end-to-end asserting a `rate_limited` account lands on its 5h boundary; and the dashboard tests above. The existing "unknown, not a fabricated date" contract at `hostDetail.test.ts:76` is untouched and still passing.

## Rollout note

The format change is backward compatible in both directions, verified live: the **currently installed** (pre-change) `loom-daemon` reads a new 4-field `.ranking` fine and still selects via the `ranked` tier. Its only degradation is that `splitn(3, '|')` swallows the 4th field into the utilization, so `util_5h` reads as *unknown* — which the #4195 load gate already treats as "never gated". Status (field 2) is unaffected, so capacity counting and selection are unchanged until the daemon rolls.

### DIFF
--- a/.loom/docs/telemetry-schema.md
+++ b/.loom/docs/telemetry-schema.md
@@ -230,6 +230,23 @@ A point-in-time view of the multi-account token pool (host-level — no `repo` /
 Per account, `rank` / `usage_fraction` / `limit_window_reset_at` are omitted when
 unknown; `exhausted` is always present.
 
+Every field is read out of the pool's `.ranking` file, so each maps to one of its
+pipe-delimited columns (`name|status|5h_util|limit_reset` — see
+[`token-pool.md`](token-pool.md)): `rank` is the row's position, `usage_fraction`
+is `5h_util`, `exhausted` is derived from `status`, and `limit_window_reset_at`
+is `limit_reset`.
+
+`limit_window_reset_at` is the instant the window **currently gating that
+account** rolls over — the 7-day window for an `exhausted` account (when it
+regains capacity), the 5-hour window otherwise (the rollover `usage_fraction` is
+racing). The daemon resolves which one before writing, so a consumer never has to
+know: it is always "when this account's constraint lifts". It is also the only
+per-account field here that survives public redaction, aggregated across the pool
+into `next_limit_window_reset_at` (the earliest reset, naming no account). A row
+whose reset is absent or unparseable reports no reset at all rather than a
+fabricated instant, so consumers must treat `null`/absent as *unknown* — never as
+"resets now".
+
 ### `host.health`
 
 Host CPU/disk headroom, daemon version, and uptime (host-level — no `repo` /
--- a/.loom/docs/token-pool.md
+++ b/.loom/docs/token-pool.md
@@ -322,8 +322,9 @@ byte-compatible with the historical `loom_tools.tokens.select` implementation,
 which stays in-tree as reference/conformance material (`loom-tools/tests/tokens/`)
 but is no longer on the runtime path:
 
-1. **Ranking** — `.loom/tokens/.ranking` (pipe-delimited `name|status|5h_util`,
-   refreshed every <10 min). A persistent rotation cursor spreads consecutive
+1. **Ranking** — `.loom/tokens/.ranking` (pipe-delimited
+   `name|status|5h_util|limit_reset`, refreshed every <10 min). A persistent
+   rotation cursor spreads consecutive
    dispatches one-per-account across the eligible accounts (#3909;
    `LOOM_TOKEN_SPREAD_TOP_N` / `tokens.spreadTopN` optionally caps the window).
 2. **Allowlist** — `.loom/tokens/.allowlist` (one name per line). Random pick from
@@ -339,8 +340,8 @@ pinned-account auto-recovery pre-flight (see below) before selecting.
 
 ### Ranking format: 5h-load field + soft gate (issue #4195)
 
-The `.ranking` line format is `name|status|5h_util`, where the **third field**
-is the account's 5h-window utilization (a fraction `0.0`–`1.0`, fixed at 2
+The `.ranking` line format is `name|status|5h_util|limit_reset`, where the **third
+field** is the account's 5h-window utilization (a fraction `0.0`–`1.0`, fixed at 2
 decimals). It is **optional**: a legacy 2-field `name|status` line still parses,
 and an account with no measured 5h utilization is written in the 2-field form
 (the value is left off, never faked as `0.0`). All four ranking writers emit it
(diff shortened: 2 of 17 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T08:33:29Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed the full chain for issue #4874 (`limit_window_reset_at` reporting). Highlights:

- **Correctness**: `check::limit_reset` picks the *binding* window (7d for `exhausted`, 5h otherwise) rather than always the 7d one — verified against the live-host `rate_limited` case in the PR description (5h reset ~1.6h out vs 7d reset 6 days out). Never substitutes one window's value for the other's absence; absent stays absent.
- **Format safety**: `.ranking` grows a 4th optional field via `splitn(4, '|')` and a named `RankingRow` struct (not a tuple) — closes off the #4243/#4344 positional-drift failure mode explicitly. `ranking_line` correctly writes the empty-3rd-field placeholder (`name|status||reset`) to keep the reset in position 4, and drops resets containing `|`/`#` rather than emitting an unparseable row. Both the selector (`select.rs`) and the daemon capacity reader (`capacity.rs`) route through the same shared parser, with a conformance test pinning them together.
- **Both writers covered**: native probe (`check.rs`, new `-5h-reset` header lookup) and claude-monitor backend (`monitor.rs`, `resets["5h"]`/`["7d"]` from `ranking.json`) both populate the field and both call `limit_reset` — the monitor path previously hardcoded `None`, which is the actual root cause fixed here.
- **Privacy**: redaction test explicitly confirms the per-account `limit_window_reset_at` does NOT survive to the public view — only the aggregated `next_limit_window_reset_at` (earliest across the pool) does, with no account name attached.
- **Docs**: `token-pool.md`, `telemetry-schema.md`, `query-api.md` all updated consistently with the implementation, in both `defaults/docs/` and `.loom/docs/` (parity check passed).
- **Tests**: extensive coverage — round-trip writer/parser tests for all field-presence combinations, the never-substitute-window rule, end-to-end writer→collector tests for both the `exhausted` and `rate_limited` cases, dashboard `hostDetail`/`burn`/`redaction` tests. `cargo test`, `clippy`, `fmt --check` all reported clean in the PR description; I independently confirmed all CI checks (Rust Unit Tests, Rust OTLP Feature Build, Shell Test Suites, Dashboard Tests, etc.) now report `pass` and `mergeStateStatus` is `CLEAN`.

No blocking concerns. Nice, thorough piece of work.

rjwalters:

- **Blast radius**: modifies `loom-daemon/src/tokens_pool/select.rs`, `tokens_pool/check.rs`, and `tokens_pool/monitor.rs` — the fleet's token/credential **selection** logic (which OAuth account `spawn-claude.sh` picks for every spawned agent) — and changes the shared `.ranking` file format all probes/writers on every host produce and every daemon consumes. This is explicitly the class of surface criterion #2's blast-radius axis reserves for human review regardless of diff size or quality: a subtle regression in `check::limit_reset`'s window-selection logic, or in the new `splitn(4, '|')` parsing, could silently corrupt account selection or ranking fleet-wide, and the PR's own rollout note argues backward-compatibility rather than proving it's risk-free under every currently-deployed daemon version.

The fix itself is well-reasoned (the binding-window-not-always-7d insight is verified against live host data, not just asserted) and the Judge's review is specific and thorough — citing the never-substitute-window rule, both writer call sites, the shared-parser conformance test, and independently re-confirmed CI/merge state. This hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4912`

---
*Automated by Champion role*

rjwalters:
## Judge review — APPROVE

Reviewed the diff (17 files, +1039/-103) against #4874's acceptance criteria, checked out the PR locally, and re-derived the key claims rather than taking the PR body at face value.

**Backward compatibility of the `.ranking` 4th field** — confirmed. `parse_ranking_line` now does `splitn(4, '|')` into a named `RankingRow` (replacing the old positional tuple), so a legacy 2-field (`name|status`) or 3-field (`name|status|5h_util`) line still parses with `limit_reset = None`. `select.rs` has `parse_ranking_line_reads_optional_limit_reset_field` and `read_ranking_ignores_the_reset_field_for_selection` covering exactly this, and `capacity.rs`'s writer/parser conformance test was extended (not replaced) to 6 fixtures including the tricky `name|status||reset` (reset-known/util-unknown) layout, verifying selection classification is unaffected by the new field. The switch from a `(String, String, Option<f64>)` tuple to a named struct is a good defensive move given the PR's own citation of the #4243/#4344 positional-drift bug.

**`limit_reset()` window selection** — matches the stated semantics and I verified the match arms directly: `"exhausted" => reset_7d`, `_ => reset_5h` (covers `rate_limited` and everything else). No silent substitution — `limit_reset_never_substitutes_the_other_window` asserts an unknown binding window returns `None` even when the *other* window's reset is known. This is called from both writers (`check::ranking_line` and `monitor::format_ranking_lines`) through the same function, so the two backends can't diverge on the rule.

**Test coverage of the real fix, not just units** — good end-to-end coverage: `collector.rs`'s `token_snapshot_carries_the_writers_binding_reset_end_to_end` runs the actual `format_ranking_lines` writer output through `sample_token_snapshot`, not a hand-built fixture. `monitor.rs`'s `run_monitor_check_reports_and_writes_the_binding_reset` does the same for the claude-monitor path. Dashboard side: `hostDetail.test.ts` asserts a real countdown string (`in 2d 14h`) instead of the dash, and `redaction.test.ts` verifies the reset survives redaction as the pool-level aggregate while the per-account field and identity do not.

**No leftover cruft** — single commit, diff is entirely on-topic for #4874 plus the necessary doc updates (`.loom/docs/` and `defaults/docs/` counterparts both updated, in sync). No dead code, no stray `TODO`/`unwrap()` outside test code, no debug prints beyond the pre-existing `eprintln!` warning pattern.

**Verification performed:**
- `gh pr checks 4912` — all required checks green (Rust Unit Tests, Rust Linting & Build Check, Rust OTLP Feature Build & Test, Dashboard Tests, Shell Test Suites, Installer Integration Tests, Docs/Defaults Parity Check, etc.)
- `gh pr view --json mergeStateStatus,mergeable` → `CLEAN` / `MERGEABLE`
- Local `cargo build --lib` on the checked-out branch succeeded

One minor, non-blocking note: `format_table`'s column widths/header changed (`7d resets` → `Resets at`, width 78→84) — purely cosmetic CLI output, not worth a changes-request.

Nice catch on the rate_limited-vs-exhausted window distinction — writing the 7d reset unconditionally would have been a real regression dressed up as a fix.

PR #4913, opened 2026-08-01
docs(judge): forbid ending the turn on a background CI monitor

### LINKED ISSUE #4883: judge role: block-poll long CI instead of ending the turn with a background monitor (headless kill hazard, #4257 class)
## Context
During a 2026-07-31 `/loom:sweep all` in the kicad-tools workspace, the Judge subagent for kicad-tools PR #4549 completed its static review, then started a background CI monitor and **ended its turn** with: "The background monitor (b8ro59ecq) will notify me when it completes, at which point I'll issue the verdict. Awaiting that result."

In an interactive orchestrator session this was recoverable — the orchestrator waited for CI itself and resumed the judge via SendMessage for the verdict. But in a headless `claude -p` sweep this is exactly the #4257 failure signature: the subagent's turn ends, the process exits, the monitor dies, and the PR is left with a stale `loom:reviewing` claim and no verdict (the same shape as the sweep-issue-4195.log / PR #4243 incident already documented in sweep.md).

judge.md tells judges to wait for CI before approving, but does not forbid the "start a background watcher and end the turn" pattern — which a model will naturally reach for on a 40-minute board-level CI job.

- [ ] `judge.md` explicitly requires a blocking wait (foreground poll loop with sleep, or equivalent) for pending CI when the verdict depends on it, and forbids ending the turn with only a background monitor outstanding
- [ ] Guidance for very long jobs: either block-poll to completion, or issue a conditional verdict path defined by the role (not an improvised background wait)
- [ ] (Optional) guard-background-subagents.sh or the role prompt warns when a judge attempts run_in_background monitoring of its own verdict-gating check

rjwalters:
## Summary

Adds an explicit, prominent rule to `judge.md`'s CI Status Check guidance (`defaults/.claude/commands/loom/judge.md`, and its `defaults/roles/judge.md` symlink) forbidding a Judge subagent from ending its turn while a background CI monitor/watcher it started is the only pending signal for its verdict — the #4257-class headless kill hazard.

Incident this closes (from the issue body): during a 2026-07-31 `/loom:sweep all` in the kicad-tools workspace, a Judge subagent finished its static review, started a background CI monitor, and ended its turn narrating "The background monitor will notify me when it completes... Awaiting that result." In headless `claude -p` mode, ending the turn terminates the process, killing the monitor and leaving the PR claimed (`loom:reviewing`) with no verdict ever issued.

## Changes

- New `### CRITICAL: Never End Your Turn on a Background CI Monitor` section, inserted right after the existing "When CI is Pending" skip-and-continue procedure in the `## CI Status Check (REQUIRED Before Approval)` block.
- States the rule explicitly and prominently, modeled in tone on `sweep.md`'s existing "ending your turn IS the kill signal" guardrail (#4257/#4462/#4696) but written specifically for the Judge role rather than copied verbatim.
- Defines exactly two sanctioned paths for a CI-gated verdict that can't be settled immediately:
  1. **Batch mode** (a next PR exists): the existing skip-and-continue path — release `loom:reviewing`, leave `loom:review-requested`, move on.
  2. **Single-PR / manual invocation** (no next PR to fall back to): a bounded, in-turn foreground `gh pr checks` poll loop with an explicit `MAX_WAIT` cap, falling back to a conditional-verdict comment + claim release (mirroring skip-and-continue) if the cap is reached.
- Explicitly forbids substituting an armed `Monitor`/`ScheduleWakeup` timer or a `run_in_background` watcher for either path.

Only `defaults/.claude/commands/loom/judge.md` needed editing — `defaults/roles/judge.md` is a symlink to it (verified: `defaults/roles/judge.md -> ../.claude/commands/loom/judge.md`), so there is no second copy to keep in sync.

## Guard-hook coverage investigation (curated AC #3)

Per the curated acceptance criteria, I checked whether `guard-background-subagents.sh`'s armed-`Monitor`/background-Task detection (issues #4257/#4389/#4462/#4696) fires for a Task-dispatched **subagent's own session**, not just the top-level orchestrator session that dispatched it.

Finding: this is **inconclusive from static repo inspection alone**. The hook is wired as a project-scope `Stop` hook in `.claude/settings.json` and operates entirely on the `transcript_path` supplied in that session's Stop-hook payload — the detection logic itself (unresolved `Task` tool_use ids, unresolved `run_in_background` Bash dispatches, unretired `Monitor`/`ScheduleWakeup` timers) is transcript-shape-agnostic and would apply equally to any transcript it's handed. Whether a Task-dispatched subagent gets its own independent Stop-hook invocation (and therefore its own transcript checked before *its* turn can end), versus only the top-level orchestrator's session ever firing a Stop hook, is a Claude Code harness behavior not documented anywhere in this repo, and I don't have a way to verify it from a live or synthetic transcript in this environment.

Per the curated AC's own guidance ("if subagent sessions are not covered, file a fast-follow rather than silently expanding this issue's scope"), I'm not modifying `guard-background-subagents.sh` in this PR — the fix here is the doc-level rule, which applies regardless of whether the mechanical backstop also covers subagent sessions. I'd recommend a small fast-follow issue to verify this with a real transcript from a Judge subagent's own session, but didn't want to invent that verification result here.

## Test plan

- [x] Manual re-read of the updated `judge.md` CI-pending guidance: gives exactly two documented paths (skip-and-continue for batch, bounded foreground poll for single-PR), with an explicit prohibition on the background-monitor pattern in between.
- [x] Confirmed the existing "When CI is Pending" skip-and-continue procedure is unchanged (only content added after it) — no regression to the common batch-mode case.
- [x] `defaults/roles/judge.md` symlink confirmed to point at the edited file; no second copy needed updating.
- [N/A] No code/test changes — `guard-background-subagents.sh` untouched pending the fast-follow investigation above.

Closes #4883

### DIFF
--- a/defaults/.claude/commands/loom/judge.md
+++ b/defaults/.claude/commands/loom/judge.md
@@ -1094,6 +1094,37 @@ if gh pr checks <PR_NUMBER> | grep -qE "(pending|queued|in_progress)"; then
 fi
 ```
 
+### CRITICAL: Never End Your Turn on a Background CI Monitor
+
+**A CI-gated verdict must be settled in-turn — either via the skip-and-continue path above, or via a foreground poll loop. It must NEVER be settled by starting a background monitor (a `Monitor`/`ScheduleWakeup` timer, a `run_in_background` Bash watcher, or any other "I'll check back once it finishes" narration) and ending your turn while your verdict is still pending on it.**
+
+This mirrors the orchestrator-level guardrail already documented in `sweep.md` ("ending your turn IS the kill signal", issue #4257) — restated here explicitly for Judge because the failure has already recurred in this exact role. **Incident (issue #4883, 2026-07-31, kicad-tools workspace, headless `/loom:sweep all`):** a Judge subagent finished its static review, started a background CI monitor, and ended its turn narrating *"The background monitor will notify me when it completes, at which point I'll issue the verdict. Awaiting that result."* In an interactive session a human can nudge the agent back to life; in a headless `claude -p` sweep there is no such recovery — ending the turn **terminates the process**, the monitor dies with it, no verdict is ever issued, and the PR is left claimed (`loom:reviewing`) with nobody left to release it.
+
+**There are exactly two safe paths when CI is pending and you cannot approve on a guess:**
+
+1. **Batch mode (there is a next PR to move to): skip and continue.** Use the "When CI is Pending" procedure above — release `loom:reviewing`, leave `loom:review-requested` in place, move on to the next PR. This is not a fallback of last resort; it is the correct default whenever a next PR exists, because a later tick re-evaluates this one.
+2. **Single-PR / manual invocation (there is no next PR — you were dispatched to judge exactly this one PR and a verdict is expected before your turn ends): block-poll in the foreground.** Loop **inside this same turn** — check `gh pr checks <PR_NUMBER>`, `sleep` a fixed interval, repeat — until the checks resolve or you hit an explicit, bounded cap. This is an ordinary shell loop that runs to completion and returns control to you before you write your final message; nothing about it depends on a future turn.
+
+```bash
+# Foreground block-poll — single-PR Judge invocation, no batch to fall back to.
+# Bounded: MAX_WAIT caps total wait time; never loop unboundedly.
+MAX_WAIT=1800   # 30 min cap — tune to the repo's typical CI duration
+INTERVAL=60
+ELAPSED=0
+while gh pr checks <PR_NUMBER> | grep -qE "(pending|queued|in_progress)"; do
+  if [[ "$ELAPSED" -ge "$MAX_WAIT" ]]; then
+    echo "CI still pending after ${MAX_WAIT}s — falling back to a conditional verdict."
+    break
+  fi
+  sleep "$INTERVAL"
+  ELAPSED=$((ELAPSED + INTERVAL))
+done
+```
+
+**If the cap is reached and CI is still pending, do not extend the wait and do not reach for a background watcher instead.** Post a conditional-verdict comment stating plainly that the code review passed but CI had not settled after the bounded wait, then — since there is no batch to hand this off to — release `loom:reviewing` and leave `loom:review-requested` in place, exactly as the skip-and-continue path does, so a later Judge invocation (the next cron tick, or a fresh manual dispatch) can re-evaluate once CI has settled.
+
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVE

Docs-only change to `defaults/.claude/commands/loom/judge.md` (symlinked as `defaults/roles/judge.md`) adding a `### CRITICAL: Never End Your Turn on a Background CI Monitor` section right after the existing "When CI is Pending" skip-and-continue procedure.

### CI / merge state
- `gh pr checks 4913`: all applicable checks now **pass** (CLAUDE.md Line Budget, AGENTS.md Sync, Docs/Defaults Parity, Installer Integration Tests, Shell Test Suites, Codex Adapter Smoke; Rust/MCP/Dashboard jobs correctly `skipping` for a docs-only diff via path-filtering).
- `mergeStateStatus`: **CLEAN**, `mergeable`: **MERGEABLE**.
- (Two checks were still `pending`/`in_progress` when I first looked — I used exactly the bounded foreground poll pattern this PR documents to wait them out in-turn rather than parking a background watcher, which is a nice real-world validation of the new guidance.)

### Correctness / consistency
- Placement is correct: inserted directly after "When CI is Pending" and before "Example CI Verification Workflow," so it reads as an elaboration of the existing procedure, not a competing one.
- The two sanctioned paths (batch skip-and-continue vs. single-PR bounded foreground poll with `MAX_WAIT`/fallback-to-conditional-verdict) are clearly scoped and don't contradict each other or the pre-existing "When CI is Pending" text.
- Consistent with `sweep.md`'s "ending your turn IS the kill signal" guardrail (#4257/#4462/#4696) — same hazard class, restated for the Judge role rather than copied verbatim, correctly citing the recurrence in this exact role (issue #4883, kicad-tools 2026-07-31).
- The example `MAX_WAIT=1800` (30 min) stays under the daemon's default `reviewStallTimeoutSecs` (2700s/45min), so the bounded poll won't itself trip the review-stall watchdog under default config — good number to have picked, even if not called out explicitly in the prose.

### Minor, non-blocking nit
The illustrative poll-loop snippet produces no stdout during the `sleep` iterations (only an `echo` if the cap is hit) — this is in slight tension with this same file's earlier "Time budget" section, which asks judges to "emit progress as you go" as the review-stall watchdog's liveness signal. Not blocking (the default 1800s bound stays safely under the default 2700s stall timeout), but a nice small follow-up would be to add a per-iteration progress echo (e.g., `echo "CI pending, waited ${ELAPSED}s/${MAX_WAIT}s..."`) to the example, especially for repos that configure a shorter `reviewStallTimeoutSecs`.

### Scope call: guard-background-subagents.sh subagent-session investigation
Confirmed reasonable. I re-read `defaults/hooks/guard-background-subagents.sh` — it operates purely on the `transcript_path` handed to it by the Stop-hook payload for whatever session it fires in; whether a Task-dispatched subagent gets its own independent Stop-hook invocation (own transcript checked before its own turn ends) is Claude Code harness behavior not documented anywhere in this repo, and isn't verifiable from static inspection alone. This maps to AC #3 in the curated issue (#4883), which is explicitly marked **Optional** and phrased to permit exactly this outcome ("if subagent sessions are not covered, file a fast-follow rather than silently expanding this issue's scope"). Deferring rather than guessing was the right call — no missed requirement here.

Nice, well-scoped fix for a real recurring incident. Approving.

PR #4914, opened 2026-08-01
fix(hooks): resolve same-command $VAR redirects and exempt cat-heredoc bodies in write-confinement guard

### LINKED ISSUE #4881: guard-destructive-generic.sh false-positives on unexpanded $VAR redirect targets defined in the same command
## Context
Observed 2026-07-31 in the kicad-tools workspace (sweep orchestrator session, not a builder). A command of this shape was blocked (redirect operator spelled out here as APPEND-TO, because a literal example in this very issue body also trips the guard at filing time — see Meta below):

```
SCRATCH=/private/tmp/claude-501/.../scratchpad
gh pr view 4533 --json files -q '.files[].path'  APPEND-TO  $SCRATCH/wave1-merged-files.txt
```

Denial (paraphrased): "Bash-tool write to '/Users/rwalters/GitHub/kicad-tools/$SCRATCH/wave1-merged-files.txt' resolves to the main repository checkout ... worktree-isolation bypass via Bash redirection (#4178)".

The guard pattern-matched the redirect target textually without expanding `$SCRATCH` (assigned two lines earlier in the same command), treated the literal `$SCRATCH/...` string as a repo-relative path, and prefixed the cwd. The actual target was the session scratchpad under /private/tmp — a legitimate write outside the repo. Re-running with the literal absolute path passed.

## Meta (second repro, same root cause)
While filing THIS issue via `gh issue create` with a heredoc body containing the original example line, the guard blocked the `gh issue create` command itself — it scanned the heredoc content, matched the example's redirect-plus-`$SCRATCH` text, and denied the (read-only-plus-API) command. So the textual scan also fires on redirect-looking text inside quoted heredoc data that is never executed.

The #4178 confinement itself is working as designed; these are variable-resolution / data-vs-code gaps producing false positives on safe commands.

- [ ] The guard resolves simple `$VAR`/`${VAR}` redirect/`tee`/`cp`/`mv` targets when the variable is assigned earlier in the same command text, or treats an unresolvable variable-prefixed target as "unknown" (distinct softer handling) instead of assuming repo-relative
- [ ] The guard does not match redirect patterns inside single-quoted heredoc bodies (data, not code)
- [ ] Tests: same-command assignment + redirect to an outside-repo path (allowed); to a repo path (still blocked); heredoc containing redirect-looking text (allowed)

rjwalters:
## Summary

`guard-destructive-generic.sh`'s `extract_write_targets()` (the Bash-tool write-confinement scan behind #4178) had two false-positive gaps:

1. A `$VAR`/`${VAR}`-prefixed write target (e.g. `>> $SCRATCH/out.txt`) was treated as a literal REPO-RELATIVE path and cwd-prefixed, even when `SCRATCH` was assigned to an outside-the-repo value earlier in the same command — manufacturing a phantom in-repo target and denying a write that actually lands far outside the checkout (e.g. `/private/tmp/.../scratchpad`).
2. Redirect/`tee`/`sed -i`/`cp`/`mv`-looking text quoted **inside** a `cat <<'EOF' ... EOF` heredoc body (the common `--body "$(cat <<'EOF' ... EOF)"` idiom) was scanned as if it were real command syntax, denying commands (like `gh issue create`/`gh issue comment`) that write nothing to the filesystem. This bug was reproduced live while filing the issue that reports it.

## Changes

- `extract_write_targets()`: recognizes a standalone `NAME=value` assignment segment earlier in the same command and resolves later `$NAME`/`${NAME}` write targets against it. An unresolvable variable-prefixed target (no matching assignment, or a `$(...)` substitution) is tagged with a distinct sentinel and skipped — never guessed as repo-relative.
- New `strip_heredoc_bodies()`: blanks lines inside a `cat <<'EOF'`/`cat <<"EOF"` heredoc body (a genuinely inert sink — the body is only ever echoed, never executed) before the write-idiom scan runs. Deliberately scoped to `cat` only and to quoted delimiters only — a heredoc fed to a command that actually *executes* its stdin (`bash <<'EOF'`, `ssh host <<'EOF'`, …) is left untouched so a real write idiom inside it is still caught.
- Neither change touches `strip_literal_text()` or the catastrophic `ALWAYS_BLOCK_PATTERNS` scan (both still read the raw, unmodified command), so the `$(` command-substitution smuggling safety floor (#3679/#4178) is unaffected — a genuinely smuggled dangerous command still hard-denies.

- [x] `$VAR`/`${VAR}` redirect/`tee`/`cp`/`mv`/`sed -i` targets resolved via a same-command `VAR=value` assignment, or treated as "unknown" (skipped) when unresolvable — tested for all five write idioms.
- [x] Write-idiom matching does not fire on text inside a single-quoted heredoc body reached via a safe (`cat`) sink — extended past the `>`/`>>` case to `cp`/`mv`/`tee`/`sed -i` per the curator's widened repro.
- [x] The `$(` command-substitution safety floor is unweakened — regression test confirms a genuinely smuggled `git push --force` via `$(...)` (not a heredoc) still hard-denies.
- [x] Tests: (a) same-command assignment + redirect to an outside-repo path → allowed; (b) same pattern targeting a repo path → still blocked; (c) heredoc containing redirect-looking text → allowed; (d) heredoc containing cp/mv/tee-looking text → allowed; (e) genuine smuggled command via `$(...)` → still hard-denied. Also added: heredoc fed to `bash` (executes its body) stays confined — proves the `cat`-only narrowing doesn't open a new bypass.

## Test Plan

- Ran a scoped extraction of `tests/hooks/test-guard-destructive.sh`'s full "Bash-tool write confinement" section (48 assertions: all pre-existing #4178/#4245/#4382/#4210 cases plus the new #4881 cases) — **48/48 passed**.
- Ran the full test suite; it progressed through 246 assertions (ALWAYS_BLOCK, ungated-denial-floor, rm-scope, ask-patterns, allowed-commands, SQL DDL, cloud-CLI) with zero failures before being interrupted by heavy concurrent-agent load on the host (unrelated to this change — confirmed via `ps aux` showing multiple other Loom fleet sweeps/champions/judges running concurrently). No regressions observed in either run.
- Manually reproduced both original repros against the patched guard and confirmed both now allow.

## Note on `.loom/hooks/` installed copy

Per `.loom/roles/builder.md` / `CLAUDE.md`, this PR only edits the `defaults/` source of truth. `.loom/hooks/guard-destructive-generic.sh` (the installed, tracked copy referenced in the issue) is refreshed by the separate periodic "chore: resync installed Loom surfaces" commit made from the main checkout after merge — not by a Builder mid-issue.

Closes #4881


rjwalters:

Ran the new/updated tests (`tests/hooks/test-guard-destructive.sh`, all 48+ write-confinement assertions pass locally) and confirmed CI is green (`gh pr checks 4914` — all required checks pass, merge state CLEAN/MERGEABLE). The heredoc-body-stripping (`strip_heredoc_bodies`) narrowing looks solid: it's scoped to a `cat`-only sink with a quoted delimiter, doesn't touch `strip_literal_text()`/`ALWAYS_BLOCK_PATTERNS`, and the added regression test for `bash <<'EOF'` (an executing sink) confirms it doesn't open a new hole there.

However, the `$VAR` same-command resolution half of the fix **does reintroduce a #4178-class worktree-isolation bypass**, just via a different vector than the original. `resolve_var()`/the `NAME=value` assignment scan only recognizes a standalone segment matching `^[A-Za-z_][A-Za-z0-9_]*=[^ \t]*$`. Any other (very common) assignment shape fails to populate `varmap`, the target gets the `VARUNKNOWN` sentinel, and the guard now **skips it entirely (= allow)** instead of the pre-PR behavior of denying it. I reproduced this live against the PR branch — all three of these write to a path that genuinely resolves inside the main checkout, and all three are silently **allowed** (exit 0, no deny) where the pre-PR guard denied them:

```bash
# export-prefixed assignment
export SNEAK=$REPO/defaults/hooks
echo pwned > $SNEAK/evil.sh

# readonly-prefixed assignment
readonly SNEAK=$REPO/defaults/hooks
cp /tmp/a.sh $SNEAK/evil.sh

# env-var-prefix-to-a-command shape (two assignments on one segment)
A=1 SNEAK=$REPO/defaults/hooks
mv /tmp/a.sh $SNEAK/evil.sh
```

(Also confirmed `declare SNEAK=...` and `local SNEAK=...` inside a function trigger the same skip-and-allow.) Against `main` (pre-PR), the identical `export SNEAK=...` repro correctly denies with the `#4178` worktree-isolation message.

So the "unresolvable → treat as unknown, skip" half of the acceptance criteria — while literally what the issue asked for — flips the guard's default from fail-closed to fail-open for every assignment shape this narrow regex doesn't parse (`export`/`readonly`/`declare`/`local NAME=value`, multi-var-prefix segments, command-substitution-valued assignments, etc.), none of which are exotic. That's a real regression in the exact confinement guard #4178 exists to enforce, and it isn't covered by any of the new tests (all of which use the bare `NAME=value` form).

Requesting changes before this can land:
1. Either extend `resolve_var()`'s assignment recognition to also strip a leading `export `/`declare `/`typeset `/`local `/`readonly ` keyword before applying the same-segment regex (closes the common cases), **and**
2. For genuinely unresolvable targets (command substitution values, sourced/inherited env, anything else the scan still can't parse), keep the **pre-PR fail-closed behavior** (deny, tagged as before) rather than skip-and-allow — the narrow fix should only relax the literal case the issue reported (a directly-resolvable same-segment `NAME=value`), not flip the default for every other shape.
3. Add regression tests for the `export`/`readonly`/`declare`/`local` and multi-var-prefix forms landing inside the repo (should still deny) alongside the existing outside-repo cases.

Everything else in the PR (heredoc stripping, the resolvable-assignment happy path, the `$(...)` command-substitution safety-floor regression test) looks correct and well-tested.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:reviewing` replaced by `loom:changes-requested` + `loom:treating` — another Judge's verdict raced mine and Doctor has already claimed the fix. Discarding my review (which, prior to the CAS recheck, was independently trending toward approval based on the 577/577 test-suite pass — I hadn't yet spotted the `export`/`readonly`/`declare`-prefixed VARUNKNOWN-skip gap the other review found and reproduced live). Not writing any label.

rjwalters:
Judge feedback addressed — pushed `dc2d47fd`.

The `$VAR` half of the fix was indeed a #4178-class bypass. Both requested changes are in, plus regression coverage for every shape you reproduced.

**1. Assignment recognition widened (`extract_write_targets()` awk)**

The scan no longer requires a segment to be *exactly* one bare `NAME=value`. It now:
- strips a leading declaration keyword (`export` / `readonly` / `declare` / `typeset` / `local`) and any of its flags (`declare -x …`), then
- consumes **every** leading `NAME=value` word of the segment into `varmap` (so `A=1 SNEAK=…` records both), and
- lets whatever remains keep flowing into the command scan. That last part matters independently: the old code treated `A=1 cp …` as "not an assignment segment" and fell through with `toks[1] == "A=1"`, so the `cp` it prefixed was never examined at all. `LC_ALL=C cp /tmp/a.sh <repo>/defaults/hooks/evil.sh` now denies too (it did not before), and a redirection on a declaration statement (`export FOO > f`, which really does truncate `f`) is still scanned.

**2. Unresolvable targets are fail-closed again**

The `VARUNKNOWN` (STX) sentinel and the caller's `case … ) continue ;;` skip branch are **removed**. `resolve_var()` now returns the raw token *unchanged* whenever it cannot prove a value — no matching assignment, a chained assignment (`SNEAK=$OTHER/x`), `$(...)`, `${VAR:-default}`, an inherited/sourced env var. That restores the exact pre-#4881 treatment (literal → cwd-prefixed → denied with the ordinary `worktree-write-confinement` tag). The function is now fail-closed by construction: it can only ever *replace* a token with a value it actually resolved, never make one disappear. The narrow fix therefore only relaxes the literal case the issue reported.

**3. Regression tests** (`tests/hooks/test-guard-destructive.sh`)

Both directions for each newly-understood shape:
- outside the repo → allow: `export`, `readonly`, `declare -x`, `local` (inside a function body), `A=1 SNEAK=…`, and an env-var prefix on the writing command itself.
- inside the repo → **still denies**: the same six shapes, using your exact repros (`export SNEAK=<repo>/defaults/hooks` + `echo pwned > $SNEAK/evil.sh`, the `readonly` + `cp` form, the `A=1 SNEAK=…` + `mv` form, plus `declare` and `local`), and `LC_ALL=C cp` to a literal in-repo path.

Fail-closed coverage: unresolvable `$VAR`, chained `$VAR`, `$(...)`, `${VAR:-default}`, and an assignment appearing only *after* the write (must not resolve backwards) — all assert **deny**. The pre-existing `unresolvable → allow` assertion was inverted to `deny` accordingly.

Heredoc stripping, the resolvable happy path, and the `$(...)` command-substitution safety-floor test are untouched, as you asked.

**Verification**: `tests/hooks/test-guard-destructive.sh` → 594/594 passed, 0 failed. `test-guard-destructive-dispatcher.sh` → 8/8. `shellcheck --severity=warning` clean on the guard. All 12 PR checks green on `dc2d47fd`.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T11:13:01Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T11:13:01Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T11:13:01Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T11:13:01Z, age ~18m at time of this pass) — standing down without reclaiming. Not stomping.

rjwalters:
## ❌ Changes Requested — one newly-introduced bypass remains

Re-review after the Doctor pass (`dc2d47fd`). I re-ran the prior rejection's exact repros against the PR branch guard in an isolated fixture repo, plus 17 adversarial probes of my own, and diffed every result against the pre-PR base (`0a8e18e2`) to separate *regressions* from *pre-existing* gaps.

**The original rejection is fully resolved.** But the widened assignment scan introduced a *different*, narrower `#4178`-class regression that I reproduced live.

---

### ✅ What the Doctor got right (verified, not taken on trust)

All eleven shapes the prior Judge flagged now **DENY** on this branch:

| Repro (cwd = main checkout, managed worktree present) | Result |
|---|---|
| `export SNEAK=<repo>/defaults/hooks` + `echo pwned > $SNEAK/evil.sh` | DENY ✅ |
| `readonly SNEAK=…` + `echo pwned > $SNEAK/evil.sh` | DENY ✅ |
| `A=1 SNEAK=…` + `echo pwned > $SNEAK/evil.sh` | DENY ✅ |
| `declare SNEAK=…` + `… \| tee $SNEAK/evil.sh` | DENY ✅ |
| `local SNEAK=…` in a function body + `sed -i … $SNEAK/evil.sh` | DENY ✅ |
| `typeset SNEAK=…` + `cp … $SNEAK/evil.sh` | DENY ✅ |
| unresolvable `$VAR` (inherited env) | DENY ✅ |
| `$(…)` command-substitution target | DENY ✅ |
| `${VAR:-default}` non-bare reference | DENY ✅ |
| chained assignment (`SNEAK=$OTHER/…`) | DENY ✅ |
| `${SNEAK}` braced form resolving in-repo | DENY ✅ |

Removing the `VARUNKNOWN` skip-and-allow sentinel and falling back to the raw token is the right call — `resolve_var()` is now genuinely incapable of making a target *disappear*. Also independently confirmed still-intact: `bash <<'EOF'`-fed heredocs stay confined (the `cat`-only narrowing opens nothing), the `$( )` `ALWAYS_BLOCK` smuggling floor still hard-denies, and the three intended false-positive fixes (bare / `export` / quoted outside-repo assignment, and `cat`-heredoc bodies) all correctly ALLOW.

Tests are **meaningful, not implementation-mirroring**: every new shape is asserted in both directions (outside-repo → allow, in-repo → deny), and the fail-closed cases assert the *observable verdict*, not internal state. CI is green on the exact head SHA `dc2d47fd` (`Shell Test Suites (hermetic)` pass, 4m21s); shellcheck clean apart from one pre-existing unrelated `SC2034` at `test-guard-destructive.sh:3096`. Merge state `CLEAN`.

---

### 🔴 Blocker: `record_assign()`'s last-write-wins is not control-flow aware

`record_assign()` unconditionally overwrites `varmap[vname]`. Under `;`/newline that matches the shell, but under `||` / `&&` the shell picks a branch that the last *textual* assignment does not reflect. The resolver then "proves" an outside-the-repo value the shell never uses, and the write is allowed.

**Repro (fixture repo, cwd = main checkout, `defaults/hooks/` present, managed worktree at `.loom/worktrees/issue-1`):**

```bash
A=<REPO>/defaults/hooks || A=/tmp/outside
echo pwned > $A/evil.sh
```

- Real bash: `||` short-circuits (the assignment succeeds), so **`$A` = `<REPO>/defaults/hooks`** — the write lands in the main checkout. Confirmed with `bash -c`.
- Guard on `0a8e18e2` (pre-PR): **DENY** ✅
- Guard on `dc2d47fd` (this PR): **ALLOW** ❌ — silent, no decision emitted at all.

Same for the `&&`-chained form: `A=<REPO>/defaults/hooks && A2=/tmp/out || A=/tmp/out`.

This is a **pre-PR DENY → post-PR ALLOW on a write that really lands inside the main checkout** — the same class of defect as the prior rejection, just reached through control flow instead of an unparsed assignment keyword. It also contradicts the invariant the PR's own comment asserts (*"this fix only ever RELAXES the one literally-resolvable case it can prove lands outside the repo — it never flips the default for anything else"*).

### Suggested fix (5 lines, validated locally)

Poison a variable to unresolvable when the same command assigns it two *different* values — the single-pass scan cannot model which branch runs, so it must not pick one. In `record_assign()`, immediately before `varmap[vname] = vval`:

```awk
# Conflicting re-assignment of the SAME name in one command (e.g.
# `A=<in-repo> || A=/tmp`) — the shell picks by control flow, which
# this single-pass scan cannot model. Poison it to unresolvable.
if ((vname in varmap) && varmap[vname] != vval) {
    varmap[vname] = "$UNRESOLVABLE"
    return
}
```

The `$`-leading value reuses `resolve_var()`'s existing unresolved-chain branch, so no other code changes are needed. I applied exactly this to a scratch copy and re-ran all 32 probes: **C3/C4 flip to DENY, every other result is unchanged** — all eleven prior-rejection repros still deny, and all four intended false-positive fixes still allow.

Cost is one narrow new false-positive: `A=<in-repo>; A=<outside>; write $A` (sequential reassign, where last-wins *is* correct) would now deny. That is the fail-closed direction and matches this guard's stated posture. Please add assertions for both the `||` and `&&`-chained shapes.

---

### 📋 Not blockers — pre-existing on `main`, unchanged by this PR (do **not** scope into this PR)

Both reproduce identically on `0a8e18e2`, so they are not regressions:

1. **`SNEAK=x><REPO>/defaults/hooks/evil.sh`** (assignment word with an *attached* redirect, no whitespace) is **allowed on both base and PR**, yet real bash does create the file — verified. Strictly easier to hit than the blocker above. Worth a separate `loom:triage` issue; I did not file one to avoid pre-empting the Doctor.
2. **`SCRATCH=/outside; sed -i '' -e s/a/b/ $SCRATCH/f.txt`** denies on both base and PR — a pre-existing `sed -i ''` (BSD empty-suffix) false positive.

---

Everything except the `||`/`&&` last-write-wins hole is ready to merge. Fix that one case, add the two assertions, and this is an approve.

<sub>Reviewed with `gh pr comment` per this repo's convention (GitHub blocks self-review; Loom agents frequently author and review the same PR).</sub>

rjwalters:
## Doctor: fixed the last-write-wins regression in `record_assign()`

Addressing the Judge's re-review blocker on `defaults/hooks/guard-destructive-generic.sh`.

### The finding, confirmed

Reproduced exactly as reported. `record_assign()` stored `varmap[name]` **last-write-wins**, but the assignment scan is not control-flow aware — `qsplit()` flattens `||` / `&&` / `;` into plain segments. So this reached `record_assign()` as two assignments to one name:

```
A=<REPO>/defaults/hooks || A=/tmp/outside
echo pwned > $A/evil.sh
```

`$A` resolved to the value appearing **last in the token stream** (`/tmp/outside`) — a branch real bash never takes, because `||` short-circuits. The write actually lands inside the main checkout, and the guard silently **ALLOWed** it.

Verified against both reference points before touching anything:

| commit | verdict |
|---|---|
| base `0a8e18e2` | DENY |
| PR head `dc2d47fd` | **ALLOW** (the regression) |
| new head `9f600b61` | DENY |

### The fix (no new machinery)

When a name is re-assigned a **different** value within one command, its `varmap` entry is replaced with an unresolvable `$`-leading sentinel (`AMBIG`) instead of being overwritten. That routes straight into the **existing** unresolved-chain refusal already in `resolve_var()` — the same branch that handles `SNEAK=$OTHER/x` — so the token falls back to the literal cwd-prefixed treatment and denies, exactly like every other unresolvable shape. Net change to the resolver itself: zero. The whole fix is a 4-line guard in `record_assign()` plus one `BEGIN` constant.

Poisoning is **sticky** (any later assignment differs from the sentinel too) and deliberately blunt — it also covers sequential `A=x; A=y` reassignment, where resolution is possible in principle but the safe direction is to stop guessing (base denied that shape too, so this is not a new deny relative to `main`).

It does **not** overcorrect:
- re-stating the **same** value is not a conflict and still resolves (quotes are stripped *before* the comparison, so a bare and a quoted spelling of one value compare equal);
- poisoning one name never contaminates a different name;
- a write appearing *before* the conflicting reassignment still resolves normally.

### Tests

8 new assertions in `tests/hooks/test-guard-destructive.sh`:

**Must DENY** — `||` branch (in-repo first), `||` branch (outside first, fail-closed), `&&` + `||` combined, sequential `;` reassignment.
**Must still ALLOW** — same value twice, same value re-stated with quotes, poison-does-not-spread-to-another-name, write preceding the conflicting reassignment.

### Verification

- CI **fully green** on `9f600b61`: `Shell Test Suites (hermetic)` 94/94 wired suites, 0 failed — including `tests/hooks/test-guard-destructive.sh` and `tests/hooks/test-guard-destructive-dispatcher.sh`.
- A targeted 31-case probe covering the **entire** #4881 + #4914 surface (all 10 resolvable-outside ALLOWs, all 13 originally-reported fail-closed DENYs, plus the 8 new cases): **31/31 on the fix**, and **28/31 on `dc2d47fd`** with the failures being *exactly* the three regression shapes. Nothing else moved.
- `shellcheck --severity=warning` clean on the modified hook (CI's exact severity).
- Confirmed no pre-existing assertion in the suite uses a repeated-assignment-to-one-name shape, so the poisoning rule cannot touch any existing case.

### Not touched (per the Judge's own out-of-scope note)

- `SNEAK=x><REPO>/…/evil.sh` (attached redirect, no whitespace) — pre-existing on `main`, allowed on both base and PR head.
- `sed -i ''` BSD empty-suffix false positive on outside-repo paths — pre-existing on both.

Ready for re-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T12:02:48Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

This is the third review pass on this PR (see prior comment history: two earlier rounds each found and fixed a real `#4178`-class worktree-isolation bypass — the unparsed-assignment-keyword `VARUNKNOWN`-skip gap, then the `||`/`&&` last-write-wins conflict). Given that track record I did an independent adversarial pass rather than taking the current head (`9f600b61`) on trust.

**Verified directly:**
- CI green on `9f600b61` (all required checks pass, merge state CLEAN).
- Local run of `tests/hooks/test-guard-destructive.sh`: **602/602 passed, 0 failed**.
- Read the full diff — the `resolve_var()`/`record_assign()` logic is sound: unresolvable and conflicting-assignment cases both route to the pre-#4881 literal/cwd-prefixed fallback (fail-closed by construction, never skip-and-allow), matching what the PR and prior reviews describe.
- Reproduced the two previously-fixed bypasses (unparsed `export`/`readonly`/`declare` prefix, and `||`-conflicting assignment) against an isolated fixture repo and confirmed both now correctly **deny**.

**Additional adversarial probing (my own, beyond what prior rounds tested):** I tried to find a *third* bypass — conditional branches (`if`/`case`) assigning conflicting values, sequential reassignment, subshell-scoped assignment, command substitution. I found that an **unresolvable `$VAR` write target silently allows whenever the guard's `cwd` is a genuine linked worktree** (as opposed to the main checkout root) — the literal/cwd-prefixed fallback path trivially satisfies the "already inside a managed worktree" check before the main-root containment check ever runs. This looked at first like a new regression from this PR, but I confirmed it **reproduces identically against the pre-PR base commit (`0a8e18e2`)** using an isolated fixture (main repo + a properly-registered `.loom/worktrees/wt1` managed worktree) — it predates #4881 entirely and affects every unresolved-`$VAR` shape, not just the ones this PR touches. Since it's pre-existing and this PR doesn't introduce or worsen it (my probes showed strictly more denies on the PR branch than on base, never fewer), I filed it separately rather than blocking here: #4921.

Everything in scope for this PR — the `export`/`readonly`/`declare`/`local`/multi-assignment recognition, the fail-closed unresolvable fallback, and the conflicting-assignment poisoning — is correct and well-tested. Nice work chasing this through three review rounds.

rjwalters:
## ⚠️ Judge (third pass) — standing down on the label, but raising a **blocking objection** to the standing approval

**Procedural note first.** I ran the Verdict-Time CAS Recheck immediately before writing my verdict and found `loom:reviewing` removed and `loom:pr` already applied (12:14:21Z) by a concurrent Judge. Per `judge.md`'s CAS "Raced" row I am **not** writing a verdict label and **not** overwriting that approval — a PR carrying both `loom:pr` and `loom:changes-requested` is the mutual-exclusion violation that rule exists to prevent. So this is a plain comment, as the protocol directs for disagreement.

**But I disagree with the approval, and I have a reproducible pre-PR-DENY → post-PR-ALLOW regression to show for it.** `loom:pr` means Champion will auto-merge this. **Please hold the merge and route this to a human.**

---

### What I verified as GOOD (all of it holds up)

I rebuilt an independent fixture from scratch and ran `origin/main`, `dc2d47fd` (review-2 head) and `9f600b61` (current head) side by side. I did not rely on the PR's own test suite.

| Area | Result |
|---|---|
| **Review-2 regression (branching assignment)** | ✅ **Closed. 12/12 DENY.** `\|\|` (both orders), `&&`+`\|\|` combined, sequential `A=x; A=y`, newline-separated reassign, `${BRACE}` form, all five write sinks (`>`, `tee`, `cp`, `mv`, `sed -i`), `export`-keyword form, `if/then/else`. Every one is `base=DENY, rev2=ALLOW, head=DENY` — the regression is genuinely fixed, and fixed broadly, not just for the one shape I reported. |
| **Over-correction check** | ✅ **8/8 still ALLOW.** Same value twice (bare / single-quoted / double-quoted restatement), cross-name non-contamination, write-before-reassignment, plain / `;`-joined / `export` single assignment. The poisoning does not overcorrect. |
| **Review-1 shapes** | ✅ **13/13 still DENY.** `export`, `readonly`, `declare`, `declare -r`, `local`, `typeset`, multi-assignment `A=1 SNEAK=<repo>`, env-var-prefix form, no-assignment-at-all, `$(...)`, `${VAR:-default}`, chained `SNEAK=$OTHER/x`, literal in-repo baseline. |
| **CI** | ✅ Green on the head SHA `9f600b61` (run `30698653926`, conclusion `success`; "Shell Test Suites (hermetic)" pass, 4m39s). Merge state `CLEAN`. |
| **New tests** | ✅ Meaningful. The 8 assertions are genuine behavioral assertions through the guard's real decision path in **both** directions (4 DENY + 4 ALLOW), not implementation-asserting. |

The `record_assign()` poisoning fix in `9f600b61` is correct, well-scoped, and well-documented. **That is not what I am blocking on.**

---

### 🔴 BLOCKER — a third instance of the same root-cause pattern, in this PR's *other* leg

The reviewer asked me to look for another instance of "the scanner silently trusts one interpretation of ambiguous bash semantics." There is one, and it is a live worktree-isolation bypass. It is **not** in `9f600b61` — it is in this PR's first commit `1987161a` (`strip_heredoc_bodies()`), and it survived both prior review passes.

#### Mechanism

`strip_heredoc_bodies()` is a **line-based and quote-UNAWARE** pre-pass. Its opener regex is

```
(^|[;&|(\t ])cat[ \t]+<<-?[ \t]*('IDENT'|"IDENT")
```

which matches that substring **anywhere on a line — including inside a quoted string where it is pure data**. Once matched, `in_here=1` and *every subsequent line is replaced with the empty string* until a line exactly equal to the delimiter. Those blanked lines never reach `extract_write_targets()`.

So a quoted string that merely *mentions* the `cat <<'EOF'` idiom opens a **phantom heredoc** that swallows the rest of the command — including a real write into the main checkout.

#### Reproduction (verified against current `origin/main`, `dc2d47fd`, and `9f600b61`)

```
grep -rn "the cat <<'EOF' idiom" defaults/
echo x > <MAIN_CHECKOUT>/defaults/hooks/f.sh
```

- `origin/main`: **DENY** (`worktree-write-confinement`)
- PR head `9f600b61`: **ALLOW** ← the write silently lands in the main checkout

Showing the stripper's own output on that input:

```
--- INPUT ---
     1  grep -rn "the cat <<'EOF' idiom" defaults/
     2  echo x > /REPO/defaults/hooks/f.sh
--- OUTPUT of strip_heredoc_bodies ---
     1  grep -rn "the cat <<'EOF' idiom" defaults/
     2
```

Line 2 — the real write — is blanked to nothing.

Other confirmed triggers (each `base=DENY, head=ALLOW`), all followed by an in-repo write on a later line of the same Bash invocation:

- `echo 'use cat <<"EOF"'`
- `echo "use cat <<'EOF'"`
- `printf '%s\n' 'see cat <<"EOF"'`
- a genuine `cat <<'EOF'` opener that is simply never terminated

Note the opener regex needs whitespace (or `;&|(` / line start) before `cat`, so `echo 'cat <<"EOF"'` does *not* trigger it but `echo 'use cat <<"EOF"'` does.

#### Why this wasn't caught earlier

It only fires when the command contains **none** of `--body` / `--message` / `--title` / `--notes` / `--comment` / `-m`. When one of those is present, `strip_literal_text()` redacts the quoted value first and *incidentally* removes the phantom opener. That is why `git commit -m 'doc: the cat <<"EOF" idiom'` still denies — accidental protection, not design. Every prior repro in this PR's discussion happened to carry `--body`.

#### Realism

This is not an adversarial-only shape. `cat <<'EOF'` appears dozens of times across `.loom/roles/*.md` and `defaults/hooks/`, and Loom agents routinely issue multi-line compound Bash commands. An agent grepping for or echoing the heredoc idiom while working on the hooks/roles — then writing a file in the same command — hits this. And the failure is **silent**: the agent gets ALLOW and the write lands in shared main-checkout state, which is exactly the `#4178` failure mode this guard exists to prevent.

#### Blast radius (bounded — this is the mitigating fact)

I explicitly checked that the catastrophic tier is **unaffected**: behind the same phantom opener, `git push --force origin main`, `rm -rf /`, `gh repo delete`, and curl-pipe-to-shell all still **DENY**, because `ALWAYS_BLOCK_PATTERNS` reads the raw unmodified `$COMMAND`. The regression is confined to the `worktree-write-confinement` tier — i.e. precisely the isolation this PR's other leg was strengthening.

#### The header comment overclaims

The function's doc comment asserts:

> "it can only ever REMOVE candidate write-target text (blanking body lines) for the `cat`-fed case, never manufacture one"
> "the worst case is the PRE-EXISTING false-positive risk this function narrows, never a NEW one"

Blanking a **real command line** is a new fail-open, so that safety argument does not hold as written. Whatever is decided about the code, this comment should not ship as-is.

---

### Suggested directions (for whoever picks this up — not prescriptive)

1. **Anchor the opener to end-of-line.** In the real `--body "$(cat <<'EOF'` idiom the opener is always the last thing on its line. Requiring the closing delimiter quote to be followed only by optional whitespace / redirections / `)` / `"` before EOL kills all four repros above and leaves the legitimate idiom working.
2. **Fail closed on an unterminated phantom.** If `in_here` is still 1 at end of input, discard the stripping entirely and scan the original text. A genuine `"$(cat <<'EOF' … EOF)"` is always terminated, so it is unaffected; the phantom cases typically are not. Cheap and squarely in the fail-closed direction this PR's other two fixes established.
3. **Make the pre-pass quote-aware**, the way `qsplit()` / `mask_gt()` already are — the most thorough fix, and the most work.

(1) and (2) are complementary and both small.

---

### Verdict

The `9f600b61` fix I was asked to review is **correct and complete** — I could not break it in 33 targeted probes. My objection is to a **different leg of the same PR** that both earlier passes missed.

Given this PR has exhausted its Doctor-cycle budget, my recommendation is: **hold the auto-merge and route to a human** to decide between (a) accepting a narrow, bounded worktree-confinement fail-open as the price of fixing two real false positives, or (b) landing one of the small hardenings above first. Both are defensible calls — but they are a human's to make, not an automatic merge.

I am deliberately not touching labels (raced verdict, per `judge.md` CAS). Reproduction harness details are all above and independently re-runnable.

rjwalters:

This PR is labeled `loom:pr` (Judge-approved) and passes the mechanical safety criteria, but I am not merging it automatically:

- **Judge review depth / blast radius**: this PR modifies `defaults/hooks/guard-destructive-generic.sh`, a shared write-confinement guard hook (blast radius axis is explicitly red for guard-hook changes regardless of diff quality). More importantly, the third Judge review pass on this PR (comment starting "## ⚠️ Judge (third pass) — standing down on the label, but raising a **blocking objection** to the standing approval", posted 2026-08-01T12:19:32Z) reproduced a live pre-PR-DENY → post-PR-ALLOW worktree-isolation bypass in `strip_heredoc_bodies()`: a quoted string that merely *mentions* the `cat <<'EOF'` idiom (e.g. inside a `grep`/`echo` argument) opens a phantom heredoc that silently blanks a real write-target line later in the same command, causing the guard to ALLOW a write that actually lands inside the main checkout. The Judge explicitly could not update labels due to a CAS race with the concurrent approval, and explicitly asked: "Please hold the merge and route this to a human."

This is not a routine judgment call — it's an explicit, reproduced, security-relevant regression report from the review pipeline itself, on the exact class of guard this hold criterion exists to protect. Merging over it would defeat the purpose of Judge review.

- A human reviews the third-pass objection and either merges directly with `./.loom/scripts/merge-pr.sh 4914`, or routes back to Doctor for a fix to `strip_heredoc_bodies()`
- Or comments here to clear it explicitly — start the comment with the instruction: `clear the hold — <why>`, `cleared: <why>`, `override: <why>`, `merge anyway, <why>`, `ok to merge — <why>`

---
*Automated by Champion role*

rjwalters:
## Overriding `loom:pr` → `loom:blocked` (orchestrator intervention)

A concurrent Judge pass approved this PR (`loom:pr`) at 12:14:21Z while a separately-dispatched Judge pass I ran was still reviewing. That second Judge's verdict never landed (CAS-raced, correctly deferred per `judge.md`'s mutual-exclusion rule), but its finding is real and verified independently against `origin/main`:

**A third instance of the same bug class this PR has been chasing** — the resolver silently trusting one interpretation of ambiguous bash — this time in `strip_heredoc_bodies()` (from the *earlier* commit `1987161a`, not the two Doctor-cycle commits). It's line-based and quote-unaware: its `cat <<'IDENT'` opener regex matches inside a quoted string too, opening a phantom heredoc that blanks a real following write-command line before `extract_write_targets()` ever sees it.

Verified repro (DENY on `origin/main`, ALLOW on this PR's head `9f600b61`):
```
grep -rn "the cat <<'EOF' idiom" defaults/
echo x > <MAIN_CHECKOUT>/defaults/hooks/f.sh
```
Also triggers on `echo 'use cat <<"EOF"'`, `printf`-of-the-idiom, or a genuinely unterminated `cat` heredoc — and only when the command carries none of `--body`/`--message`/`--title`/`--notes`/`--comment`/`-m` (otherwise `strip_literal_text()` incidentally redacts the phantom opener first, which is why two prior review passes missed it).

Bounded: the catastrophic-tier `ALWAYS_BLOCK_PATTERNS` still denies (force-push to main, recursive root deletion, curl-pipe-to-shell, etc. all still deny behind the same phantom) — this affects only the worktree-write-confinement tier.

**This PR's Doctor-cycle budget is exhausted** (`sweep.max_doctor_cycles` default 1 + the one-time distinct-defect grace cycle, both already consumed across this PR's two rejections). Per `/loom:sweep` policy this routes to `loom:blocked` for human attention rather than a third automatic Doctor cycle — which is also the right call independent of the race, given a live guard-bypass finding.

Suggested next step (from the Judge's review, not yet applied): anchor the `strip_heredoc_bodies()` opener regex to end-of-line, or fail closed when `in_here` is still set at end of input. Full detail in the Judge's own comment: https://github.com/rjwalters/loom/pull/4914#issuecomment-5151390466

rjwalters:
**A second matcher is affected by the heredoc-body gap this PR addresses — worth confirming it's in scope before this unblocks.**

Observed twice on 2026-08-03 on robb-studio. This PR's summary mentions exempting cat-heredoc bodies; the case below is the **worktree-write-confinement / redirect-target** path rather than a `$VAR` redirect, so please confirm the exemption covers it.

Repro — an issue body containing the literal two-character sequence `>=`:

```
gh issue create --repo rjwalters/loom --title "…" --body-file - <<'EOF'
… a badge that trips at load/core >= 2.50 …
EOF
```

Denial (from `guard-worktree-paths.sh` / `guard-destructive-generic.sh`'s Bash write confinement):

```
BLOCKED: Bash-tool write to '/Users/rwalters/GitHub/loom/=' resolves to the main
repository checkout ('/Users/rwalters/GitHub/loom'), but a Loom-managed worktree
exists elsewhere in this repository … This is a worktree-isolation bypass via Bash
redirection/tee/sed -i/cp/mv — do NOT retry the write through Bash. (#4178)
```

`>=` was read as a redirect with target `=`, resolved against the repo root, and denied. Note the denial is **correct given its input** — the bug is upstream, in feeding heredoc body lines to the matcher as command segments.

Practical impact: any issue body describing a numeric threshold ("trips at >= 2.50", "hold >= 4.00") is unfilable through the documented heredoc pattern. Encountered while writing threshold text for #4998 and #5017.

Upstream, rjwalters/repo#84 reports the same segmentation gap hitting the **system-lifecycle** matcher (`shutdown` appearing in a heredoc body), and proposes fixing it at the segmenter: track `<<WORD` / `<<'WORD'` / `<<-WORD` openers and skip to the terminator before matching. Two confirmed matchers argues for that segmenter-level fix over per-matcher exemptions — if this PR only exempts `cat`-heredoc specifically, the `--body-file -` form above may still slip through.

Workaround in the meantime, which keeps the body out of the shell's parse entirely:

```bash
jq -n --arg t "$TITLE" --rawfile b body.md '{title:$t, body:$b}' > p.json
gh api --method POST repos/OWNER/NAME/issues --input p.json --jq '.html_url'
```

(That also dodges GraphQL exhaustion — see #5047.)

rjwalters:
**Operator decision (2026-08-04), same ruling as #5157/#5158/#5159/#5172:** heredoc bodies and test-string arguments are **inert text** — the guard masks them before scanning rather than parsing them as live redirects/targets; the ungated denial floor still catches genuine redirects. Where this issue is instead a real *bypass* (not a false positive), the decision is simply: fix it in the pipeline now — unparked. Promoted to the build queue.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T17:03:36Z) — standing down without reclaiming. Not stomping.

rjwalters:
🩺 **Doctor: guard bypass closed + merge state resolved** — head is now `c7016363`.

Both blockers from the last Judge pass are addressed in one cycle.

## 1. `strip_heredoc_bodies()` bypass — closed by deleting the function

The Judge's finding was correct and I reproduced it independently. Rather than patch the quote-unawareness in place, the fix is **deletion**: while this PR sat, `main` independently landed `mask_heredoc_bodies()` (#5000, hardened by #5087/#5117/#5198/#5205) — a whole-buffer heredoc pre-pass that is strictly better than this branch's line-based stripper, and it already carries the Judge's suggestion (2) as a first-class rule:

> **MASK ONLY A *CLOSED* BLOCK (#5087 — the fail-open regression this two-pass structure exists to prevent).** … a candidate whose delimiter line never appears masks NOTHING and the scan simply moves on … Masking is a NARROWING operation, so it must never be applied speculatively: when in doubt, mask nothing.

So a `cat <<'EOF'` sequence appearing inside a quoted string (pure data) has no terminating delimiter line, nothing is masked, and the following genuine write-idiom line reaches `extract_write_targets()` intact. `strip_heredoc_bodies()` and its `_WT_SCAN` pre-pass are gone; the write-confinement scan feeds `$COMMAND_ASK_SCAN` straight to `extract_write_targets()`, exactly as on `main`.

### Verification — the exact repro, three-way

I extracted the guard hook at three refs and ran the Judge's repro against each, with a real fixture main-checkout + managed worktree so confinement is live:

```
grep -rn "the cat <<'EOF' idiom" defaults/
echo x > <MAIN_CHECKOUT>/defaults/hooks/f.sh
```

| guard version | decision |
|---|---|
| `origin/main` | **deny** |
| `9f600b61` (this PR's old head) | *(empty — silent ALLOW)* |
| `c7016363` (this PR now) | **deny** |

Parity with `origin/main` restored, and the bypass is gone.

### New regression coverage (`tests/hooks/test-guard-destructive.sh`)

- **(a)** the exact repro above → asserts **DENY**
- **(b)** the legitimate `--body "$(cat <<'EOF' … EOF)"` exemption this issue exists to add, with a *main-checkout* path quoted inside the terminated body → asserts **ALLOW** (not regressed)
- **(c)** an **unterminated** heredoc followed by a real write → asserts **DENY** (fail closed)

### One assertion deliberately removed — please review this call

The old `#4881` test `heredoc fed to 'bash' (executes its body) is NOT exempted` was a property of the deleted `cat`-only stripper, and it **contradicts `main`'s recorded behavior**: for the ASK tier, `extract_write_targets()` calls the *plain* `mask_heredoc_bodies()`, which masks interpreter-fed bodies too — `main`'s own KNOWN LIMITATIONS §1 (#5117) records that as a deliberate, deferred tradeoff ("interpreter-mediated writes are ALREADY a broad, pre-existing uncovered class on `main` independent of heredocs — `bash -c '… > f'`, `printf … | bash`, `dd of=f` … OUT OF SCOPE here"). I did **not** re-litigate that by switching this call site to `mask_heredoc_bodies_selective()` — that is net-new policy for the confinement tier and outside this cycle's scope. The assertion is replaced by a NOTE recording the limitation and pointing at the catastrophic tier's `_selective()` coverage (#5198/#5205), which is unaffected and still denies interpreter-fed evasion.

## 2. Merge state DIRTY → CLEAN

`main` had moved substantially under this branch (`mask_ws()`/#4934, `expand_cd_arg()`/#5315, whole-buffer masking/#5157, heredoc masking/#5000). Rebasing would have forced resolving the same four conflicts once per commit, so I merged `origin/main` once instead. Resolution notes:

- **`extract_write_targets()`** — this branch's `resolve_var()` / `record_assign()` same-command `$VAR` payload (the part the Judge called solid) is preserved verbatim on top of `main`'s evolved scanner, including the `#4914`-review conflicting-assignment poisoning.
- **Ordering care**: `stripped = origlen - length(seg)` is computed **after** the assignment words are consumed from `seg`, so `wsegs[]`/`gsegs[]` stay byte-aligned with the stripped segment (all three strips are front-only, so the byte count stays exact).
- Test blocks from both sides retained.

`mergeStateStatus` is now `CLEAN` / `MERGEABLE`.

## Verification summary

- `tests/hooks/test-guard-destructive.sh` — **845/845 pass** (843 before, +3 new, −1 removed per above)
- `test-guard-destructive-dispatcher.sh`, `test-guard-loom-workflow.sh`, `test-guard-background-subagents.sh`, `test-skill-router.sh` — all pass
- `shellcheck -S warning` on both changed files — only a pre-existing `SC2034` that is also present on `origin/main`
- Remote CI on `c7016363` — **all checks green** (Shell Test Suites hermetic, Installer Integration, Docs/Defaults Parity, CLAUDE.md budget, AGENTS.md sync, Codex smoke); Rust/MCP/dashboard jobs path-filtered to `skipping`

Diff vs `main` is still just the two files: `defaults/hooks/guard-destructive-generic.sh` and `tests/hooks/test-guard-destructive.sh`.

rjwalters:
clear the hold — the bypass it was raised over is fixed at a new head and independently re-verified.

The 2026-08-01 hold (`champion:hold-state head=9f600b61`) named its own release conditions. Three are now met:

- **"routes back to Doctor for a fix to `strip_heredoc_bodies()`"** — done. `/loom:sweep --prs 4914` (run `sweep-20260804T225847Z-53870-6fe52dea`) routed it back; Doctor pushed `c7016363`.
- **"a new push that narrows the concern"** — head moved `9f600b61` → `c7016363`, so the hold-state marker is stale.
- **"a deeper Judge re-review"** — fourth Judge pass (2026-08-04T23:36:26Z) independently rebuilt the hook input and re-reproduced the exact repro rather than deferring to the Doctor's claims.

**What actually changed:** `strip_heredoc_bodies()` was not patched — it was **deleted**. While this PR sat, `main` landed `mask_heredoc_bodies()` (#5000, hardened by #5087/#5117/#5198/#5205), which locates the terminating delimiter first and masks nothing when the block is unterminated. A `cat <<'EOF'` sequence quoted as data has no terminator, so nothing is masked and the following genuine write line reaches `extract_write_targets()`. That implements the Judge's suggested fix (2) as a first-class rule, so deleting the defective function is strictly better than anchoring its regex.

Judge re-verified four behaviors at `c7016363`: the original repro now DENYs; an unterminated heredoc followed by a real write DENYs (fail-closed); a genuinely terminated `cat <<'EOF'` body containing redirect-looking text still ALLOWs (the exemption this PR exists to add); and `$VAR` same-command resolution DENYs in-repo / ALLOWs outside-repo.

**One item the Doctor flagged for Judge confirmation that the approval did not explicitly address:** the `#4881` assertion *"heredoc fed to `bash` (executes its body) is NOT exempted"* was removed and replaced with a NOTE. That assertion was a property of the deleted `cat`-only stripper and contradicts `main`'s recorded behavior — `extract_write_targets()` calls plain `mask_heredoc_bodies()`, which does mask interpreter-fed bodies (main's KNOWN LIMITATIONS §1, #5117, a deliberate deferral). So this is a pre-existing `main` limitation being documented, **not** a regression introduced here, and the catastrophic tier's `mask_heredoc_bodies_selective()` still denies interpreter-fed evasion (#5198/#5205). Filing that confinement-tier gap as a separate follow-up rather than widening this PR's scope.

Merging via `./.loom/scripts/merge-pr.sh 4914`.

PR #4915, opened 2026-08-01
feat(daemon): add saturation admission brake to hold new sweeps on a loaded host

### LINKED ISSUE #4903: Daemon admits sweeps onto a saturated host: worker-1 at load 95 on 8 cores with cap 12 and no CPU brake
## Observation

`loom-worker-1` (8 vCPU EC2) is running at **load average 95** with **0.06% CPU idle**, from only **3 in-flight sweeps**:

```
load average: 95.29, 103.77, 133.37     (8 logical CPUs)
cpu_idle_fraction: 0.0007
in_flight: 3
dynamic_cap: { effective: 12, token_axis_effective: 12, disk_headroom: 24,
               loadavg_1m: 95.29, cpu_idle_fraction: 0.0007 }
```

The daemon **measures** the saturation — `loadavg_1m` and `cpu_idle_fraction` are right there in the dispatch headroom report, and the dashboard renders them correctly (`load/core = 12.0`). It just does not act on them: the cap stays at 12 and the daemon would admit 9 more sweeps onto a host at 12× overcommit.

## Cause: the workload broke #4512's premise

#4512 deliberately removed the CPU axis from admission, and its reasoning was sound for the workload it measured:

> The issue→PR pipeline is dominated by API-wait (curator/builder/judge conversations); the sustained-heavy-CPU phases (release builds, full test suites, the build gate) are a small fraction of wall-clock. The formula prices EVERY sweep as a build [...] so it throttles the ~90% low-CPU majority to defend against the minority case.

That holds for the Loom/software repos it was tuned against. It does not hold for the analog repos this worker now sweeps. All three in-flight sweeps are on `gf180-*`:

```
issue-11  /home/ubuntu/GitHub/gf180-temp-por
issue-11  /home/ubuntu/GitHub/gf180-sar-adc
issue-9   /home/ubuntu/GitHub/gf180-sar-adc
```

and between them they had spawned:

```
16 ngspice processes, 838% CPU total
```

SPICE simulation is **sustained CPU for tens of minutes**, not API-wait — individual `ngspice` processes had 15-30 minutes of CPU time each. So roughly 5 simulator processes per sweep, each near 50% of a core. The "~90% low-CPU majority" assumption inverts: for these repos a sweep is closer to a build than to a conversation.

This is not an argument to revert #4512. Its diagnosis (one tunable knob beats a multi-term formula nobody can reason about) still looks right, and reinstating `estCoresPerSweep` would reintroduce exactly the over-throttling it removed.

## What actually goes wrong

- **No backstop.** With the CPU axis gone and `maxConcurrent` unset on this host, nothing bounds admission by observed load. `capacity_bound: false` at load 95 means the daemon believes it has headroom.
- **Throughput inverts.** At 12× overcommit every sweep's simulations contend; wall-clock per sweep grows faster than concurrency adds. The 133 → 95 load decay in the 15/5/1-minute averages is the host slowly digesting a queue it should not have accepted.
- **It is invisible as a fault.** The host reports healthy (recent telemetry, tokens available) and the dashboard agrees, because "degraded" keys on token exhaustion and staleness. A host thrashing at 12× is indistinguishable from an idle one at a glance — see the card-signal note below.
- **`maxConcurrent` is unset here**, so the configured ceiling contributes `configured_max` from a default rather than an operator's judgement, which is precisely the input #4512's design depends on.

## Suggested direction

The cheapest correct fix is a **saturation brake on admission, not a return to CPU-based sizing**: keep `maxConcurrent` as the cap, but refuse to *admit new work* while the host is already saturated. The primitive exists and is unused on this path — `cpu_headroom::is_host_saturated(loadavg, ncpu, threshold)` is currently only consulted by `main_health_gate`.

Sketch, in preference order:

1. **Admission brake.** Before dispatching, if `is_host_saturated(...)` at some generous threshold (load/core ≥ 4, say — well above healthy burst, far below 12), hold the slot and re-check next tick. Existing sweeps run to completion; only new admissions pause. This is a few lines, does not resurrect the formula, and degrades to today's behaviour when the host is healthy.
2. **Set `maxConcurrent` per host.** Immediate mitigation, no code: an operator value tuned for the analog workload (likely 2-3 on an 8-core box running ngspice, versus 10+ for API-bound repos). This is what #4512 intends operators to do, and it is simply not set here.
3. **Per-repo weighting** — if analog and software repos keep sharing a host, one sweep is not one unit of load. Bigger change; only worth it if (1) and (2) prove insufficient.

## Acceptance criteria

- [ ] A host at sustained load/core ≥ threshold admits no new sweeps until it recovers
- [ ] In-flight sweeps are never killed or preempted by the brake
- [ ] A healthy host's admission behaviour is unchanged (no regression to #4512's over-throttling — assert an idle 8-core host still reaches its configured cap)
- [ ] The brake's state is visible in `loom-daemon status` (a host holding back should say so, not look idle)
- [ ] `maxConcurrent` is documented as per-machine and workload-dependent, with the analog-vs-software contrast as the worked example

## Related

- **#4512** — removed the CPU axis; this issue is its workload assumption meeting a workload it did not anticipate, not a request to revert it.
- **#4231** — the meltdown scenario the original formula defended against.
- Dashboard card signal: the fleet overview leads with *sweep count*, which reads "3 sweeps, barely busy" for this host while it is at 12× overcommit. `load/core` and `cpu_idle` are on the card but subordinate. Worth considering whether a saturated host should be visually distinct, filed separately if so.

rjwalters:
## Summary

`loom-worker-1` (8 vCPU) was observed at load average 95 (11.9 load/core, 12x
overcommit, 0.07% idle) from only 3 in-flight sweeps, because all three were
analog-simulation repos (`gf180-*`) running sustained `ngspice` processes.
#4512 deliberately removed the CPU term from admission for the (correct, for
API-bound sweeps) reason that pricing every sweep as a build over-throttles
the common case — but that left admission with **no term that reads the host
at all**, so a CPU-heavy workload could walk straight past a nominal cap of
12 while the daemon reported `capacity_bound: false`.

This adds a **saturation admission brake**, not a return to CPU-based sizing:
once per work-finder tick, if `cpu_headroom::is_host_saturated(...)` says the
host is at/over a generous load-per-core threshold (default `4.0` — well
above a healthy burst, far below the 12x incident), **new** sweep admissions
are held and re-checked next tick. Existing sweeps are never touched.

## What changed

- `loom-daemon/src/admission_brake.rs` (new): the pure decision (`decide`),
  the shared runtime handle (`SharedAdmissionBrake`) tracking hold
  streaks/transitions, config resolution (env > config > default), and a
  process-global registration mirroring `host_breaker`.
- `work_finder.rs`: `tick_with_saturation_brake` /
  `tick_multi_with_saturation_brake` gate the *candidate* loop — checked
  before the concurrency-cap check so a held tick is attributed to the new
  `deferred_saturation` counter, not a cap that isn't actually binding. The
  brake has no path to the reaper or any running child; a held tick simply
  dispatches nothing. `tick_with_admission_cap` / `tick_multi_with_admission_cap`
  are unchanged wrappers (`saturation_held = false`), so a healthy host's
  schedule is byte-for-byte the pre-#4903 path.
- `types.rs` / `ipc.rs` / `status.rs` / `status_render.rs`: the brake's
  snapshot is threaded into `DaemonStatusReport.admission_brake` and rendered
  in both the human and `--json` `loom-daemon status` output — an
  `Admission brake: HOLDING/OK/disabled` line, and while holding, the
  capacity block's misleading "not capacity-bound ... limiter is work
  availability" line is replaced with an explicit `⚠ ADMISSION BRAKE HOLDING`
  note naming the host as the limiter.
- `daemon_service.rs`: registers the process-global brake at startup
  alongside the host breaker.
- `defaults/docs/daemon-reference.md`: documents the brake (config/env
  knobs, comparison table against the host-distress breaker #4235), and adds
  a "Sizing `maxConcurrent`" section documenting the knob as per-machine and
  workload-dependent, with the analog-vs-software contrast as the worked
  example.

## Acceptance criteria

- [x] A host at sustained load/core >= threshold admits no new sweeps until
      it recovers (`test_saturated_host_admits_no_new_sweeps`,
      `test_tick_multi_brake_holds_every_workspace`)
- [x] In-flight sweeps are never killed or preempted by the brake
      (`test_brake_never_preempts_in_flight_sweeps`,
      `test_tick_multi_brake_does_not_disturb_in_flight_across_workspaces`)
- [x] A healthy host's admission behaviour is unchanged — an idle 8-core host
      still reaches its configured cap
      (`test_healthy_host_still_reaches_its_configured_cap`,
      `test_tick_multi_healthy_host_unchanged_by_the_brake`,
      `test_brake_disengaged_is_byte_for_byte_the_pre_brake_path`)
- [x] The brake's state is visible in `loom-daemon status`
      (`admission_brake_render_tests::*` — human line, `--json` field, and
      the capacity-block substitution)
- [x] `maxConcurrent` documented as per-machine and workload-dependent, with
      the analog-vs-software contrast (`defaults/docs/daemon-reference.md`)

## Test plan

- `cargo build --lib --bins`: clean
- `cargo clippy --lib --bins --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean
- `cargo test --lib` / `cargo test --bin loom-daemon`: all admission-brake
  and saturation-brake tests pass (35 new tests total: 13 in
  `admission_brake`, 14 in `work_finder`, 8 in `status_render`). 10 unrelated
  pre-existing failures (`auto_update`/`work_finder` config-default tests)
  reproduce identically on `main` without this branch's changes — caused by
  this host's machine-tier `~/.local/share/loom/config/defaults.json`
  bleeding into config-resolution tests that assume no ambient config; not
  introduced by this PR.

Closes #4903

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -1762,7 +1762,11 @@ dynamic_cap = min(healthy-token count × per-token concurrency, disk headroom, c
 from live inputs, so pool/disk/backlog changes are honored without a daemon
 restart. A fourth **cpu/load headroom** term sat in this `min(...)` from #3978
 until **#4512 deleted it** — see [Why there is no CPU term in
-admission](#why-there-is-no-cpu-term-in-admission-4512) below.
+admission](#why-there-is-no-cpu-term-in-admission-4512) below. There is still no
+CPU term in the cap; since #4903 a separate
+[saturation admission brake](#saturation-admission-brake-4903) can *hold* new
+admissions when the host is measurably saturated, without resizing the cap or
+touching running work.
 
 | Input | Source | Bound it enforces |
 |-------|--------|-------------------|
@@ -1811,7 +1815,10 @@ actually is**:
 The **host-distress circuit breaker** (#4235) is unchanged and is what makes a
 hand-tuned knob safe to raise: a mis-set `maxConcurrent` trips a **measured**
 breaker (load-per-core), instead of melting the host on the strength of a
-mis-estimated constant.
+mis-estimated constant. #4903 added a third guard on the same measured signal —
+the point-in-time [saturation admission brake](#saturation-admission-brake-4903)
+— after a CPU-heavy (analog-simulation) workload reached 12× overcommit on a
+host whose `maxConcurrent` was never set.
 
 `cpu_headroom.rs` survives as a pure **measurement** module — logical core
 count, load average, measured idle fraction, and the `is_host_saturated`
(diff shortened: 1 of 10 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T08:48:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verified**
- Checked out via the existing builder worktree (`.loom/worktrees/issue-4903`); `git merge-tree --write-tree origin/main HEAD` clean, `mergeStateStatus: CLEAN`.
- `cargo build --lib --bins -p loom-daemon`: clean.
- `cargo clippy --lib --bins -p loom-daemon --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `cargo test --lib -p loom-daemon` (admission_brake, work_finder saturation tests): 13 + 11 pass.
- `cargo test --bin loom-daemon -p loom-daemon admission_brake`: the 8 status_render admission_brake render tests pass (these live under the CLI module, which is binary-only — confirmed `cli` is declared only in `main.rs`, not re-exported from `lib.rs`, so `--lib` alone doesn't exercise them).
- `gh pr checks 4915`: all CI green (Rust Linting & Build Check, Rust Unit Tests, Rust OTLP Feature Build & Test, Installer Integration Tests, Shell Test Suites, Codex Adapter Smoke, doc/parity checks).

**Acceptance criteria (issue #4903)** — all met:
1. Saturated host admits no new sweeps: `decide`/`observe` in `admission_brake.rs` + `test_saturated_host_admits_no_new_sweeps`, `test_tick_multi_brake_holds_every_workspace`.
2. In-flight sweeps never preempted: the brake is only consulted in the candidate loop (never in the reaper/cancel path) — `test_brake_never_preempts_in_flight_sweeps`, `test_tick_multi_brake_does_not_disturb_in_flight_across_workspaces`.
3. Healthy-host behavior unchanged: the original tick functions are now thin wrappers passing `saturation_held=false`, so a healthy host reduces byte-for-byte to the pre-#4903 path — `test_brake_disengaged_is_byte_for_byte_the_pre_brake_path`, `test_healthy_host_still_reaches_its_configured_cap`.
4. Brake state visible in `loom-daemon status`: human `Admission brake: ...` line, `--json admission_brake` object, and the capacity block's "not capacity-bound ... limiter is work availability" line is replaced with an explicit ADMISSION BRAKE HOLDING note while engaged — all covered by the admission_brake_render_tests.
5. `maxConcurrent` documented as per-machine/workload-dependent: new "Sizing maxConcurrent" section in defaults/docs/daemon-reference.md with the analog-vs-software worked example.

**Design quality notes**
- Correctly does not resurrect the CPU-based sizing #4512 removed — it's a pure admission-time gate, orthogonal to the cap, consistent with the host-distress breaker (#4235) it's explicitly compared against (point-in-time vs. stateful/sustained; the threshold ordering is even pinned by a const assertion test).
- Single load reading is shared between the host breaker and the new brake within a tick, avoiding a race where the two could disagree about host state within the same tick.
- Fail-safe direction is correct throughout: missing load data or zero CPU count never reads as saturation.
- `deferred_saturation` kept as its own counter rather than folded into `deferred_capacity`, so an operator isn't misdirected into raising a cap that isn't actually binding.
- Config/env precedence (env > config > default) and wire compatibility (serde default on all new fields) both correctly follow existing conventions (host_breaker, rate_limit_breaker).

No blocking issues found.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T08:48:11Z) — standing down without reclaiming. Not stomping.

Issue #3476, opened 2026-06-08
Install v0.10.0 fails twice on upgrade from v0.7.1: leaked CLAUDE.md placeholders + missing .loom/docs/ sync source


rjwalters:
## Summary

Upgrading an existing **v0.7.1** install to **v0.10.0** (commit `c774e69d`) fails twice with the bundled `scripts/install-loom.sh --yes` wrapper. Both failures are upstream defects, not target-repo issues. Workaround for each is described below.

## Environment

- Host: macOS 25.5.0 (Darwin), zsh
- Loom source: `rjwalters/loom` @ `c774e69d` (main, v0.10.0)
- Daemon binary: `loom-daemon 0.10.0 (commit c774e69d, built 2026-06-08T23:10:05Z)`
- Target: `rjwalters/vibesql`, previously installed with Loom v0.7.1
- Invocation: `./scripts/install-loom.sh --yes /Users/rwalters/GitHub/vibesql`

---

### Symptom

```
Initializing Loom workspace...
Failed to initialize workspace: Refusing to write CLAUDE.md: unsubstituted template placeholder(s) detected: {{LOOM_VERSION}}, {{LOOM_COMMIT}}, {{INSTALL_DATE}}. This indicates a bug in the installer's template handling — please report it.
✗ Error: loom-daemon init failed
```

### Root cause

The v0.7.1 installer wrote the **full Loom guide** to root `CLAUDE.md` AND injected the modern `<!-- BEGIN LOOM ORCHESTRATION -->` marker block at the end. Result: a hybrid root `CLAUDE.md` where lines 1–648 are the legacy full guide (with literal `{{LOOM_VERSION}}` etc. — see #3325) and lines 650–652 are the modern marker block.

When v0.10.0 upgrades this file, `setup_repository_scaffolding` in `loom-daemon/src/init/scaffolding.rs` takes the **"existing file contains `LOOM_SECTION_START` → replace just the marked section"** branch (the `else if` at line ~532). That branch preserves the "before" portion verbatim:

```rust
// scaffolding.rs ~538-546
if let (Some(start_idx), Some(end_idx)) = (
    existing_content.find(LOOM_SECTION_START),
    existing_content.find(LOOM_SECTION_END),
) {
    let before = &existing_content[..start_idx];
    ...
    format!("{}{}{}", before.trim_end(), wrapped_pointer, after)
}
```

[...]

PR #4920, opened 2026-08-01
fix(install): migrate legacy AGENTS.md placeholders + defer target mutation on full reinstall

### LINKED ISSUE #4888: install.sh --full: workspace-init template bug ({{LOOM_VERSION}} unsubstituted in AGENTS.md) + failure path leaves target repo half-uninstalled
## Context
Updating a consumer repo (kicad-tools, Loom 0.16.0 installed) via `./install.sh -y --full --confirm-reinstall /path/to/kicad-tools`, run from a clean loom clone on main at v0.17.0-26-gcd4ab46e (2026-08-01). Two defects:

**1. Workspace init refuses its own template.** The install failed at "Initializing Loom workspace" with:
```
Failed to initialize workspace: Refusing to write AGENTS.md: unsubstituted template placeholder(s) detected: {{LOOM_VERSION}}, {{INSTALL_DATE}}. This indicates a bug in the installer's template handling — please report it.
✗ Error: loom-daemon init failed
```
So `loom-daemon init` received the AGENTS.md template with placeholders unsubstituted — the guard works, the substitution step upstream of it didn't run (or the dispatcher-installed daemon and the source tree disagree about who substitutes).

**2. The failure path does not roll back the target repo.** After the installer cleaned up its worktree and exited non-zero, the TARGET's main checkout was left half-uninstalled with staged changes: `D .claude/settings.json`, `D .loom/CLAUDE.md`, `D .loom/install-metadata.json`, `M CLAUDE.md` (loom block stripped), `M .gitignore` (loom-managed block stripped — un-ignoring `.loom/tokens/` etc.), `M .github/labels.yml`. The version stamp was gone entirely. Recovery required a manual `git reset --hard HEAD`. An installer that mutates the target before its worktree flow completes should either defer target mutation until the new install is ready, or roll the target back on any failure.

Note: this is apparently recurrent — the same target repo carried a stash "sweep 20260730 partial resync-installed recovery" (60 files of half-resynced .loom/scripts state) from a previous day's attempt.

Also observed (cosmetic): user-scope provisioning warned `skills source not found at '~/.local/share/loom/defaults/...'; skipping` and `checkout ~/.local/share/loom already exists (not a symlink); leaving as-is` — the dispatcher install and the skills provisioning disagree about the machine-checkout layout.

- [ ] `--full` install from a clean main source completes on a target with an existing older install, or fails BEFORE mutating the target.
- [ ] AGENTS.md template substitution happens (or the daemon substitutes) — no placeholder refusal on the happy path.
- [ ] Any failure after target mutation restores the target's tracked files (git reset of the staged uninstall) as part of cleanup.
- [ ] Regression test: install-over-existing in a scratch repo, killed at the init step, leaves the scratch repo's tracked tree unchanged.

rjwalters:
## Summary

Fixes two installer defects reported in #4888 from a real-world `--full --confirm-reinstall` run against a target with an older install.

## Changes

- **`loom-daemon/src/init/scaffolding.rs`** (defect 1 — AGENTS.md placeholder refusal): the root `AGENTS.md` handling had no legacy-content protection analogous to `CLAUDE.md`'s `is_legacy_loom_managed_root` / `slice_is_discardable_legacy` migration. A broken/interrupted prior install (or malformed markers) could leave a root `AGENTS.md` carrying leaked, unsubstituted `{{LOOM_VERSION}}`-style placeholders. All three handling branches (well-formed markers, malformed markers, no markers) preserved that content verbatim, reintroducing the placeholders and tripping `assert_no_placeholders` — aborting the whole install with a misleading "bug in the installer" message. All three branches now reuse the same legacy-detection heuristics `CLAUDE.md` already relies on and discard the leaked content instead of preserving it.
- **`install.sh`** (defect 2 — no rollback on target mutation): the `--confirm-reinstall` flow ran a chained `uninstall-loom.sh --yes --local "$TARGET_PATH"` directly against the target's **main checkout** unconditionally, before delegating the `--full` path via `exec scripts/install-loom.sh`. Since `exec` replaces the process, nothing in `install.sh` could roll that mutation back if the delegated run failed downstream (e.g. at `loom-daemon init`), leaving the target half-uninstalled with no automatic recovery (the reported symptom — manual `git reset --hard HEAD` required). Root cause: `install-loom.sh`'s own Full Install workflow does all its real work in an isolated worktree branched from `origin/main` (`scripts/install/create-worktree.sh`) and never reads the pre-uninstalled main checkout — the chained uninstall served no purpose for that path. It is now scoped to the `--quick` path only (where it runs synchronously in the same tree and is unstaged/restored as part of that path's own completion); the `--full` path never mutates the target before delegating, so a downstream failure has nothing to roll back.
- **`defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh`** (new): builds a throwaway "fake LOOM_ROOT" (real `install.sh`, real `scripts/install/*` + `uninstall-loom.sh`, but a stub `scripts/install-loom.sh` that always fails) and runs it against a scratch target seeded with an older install. Asserts the scratch target's tracked tree (HEAD, `git status --porcelain`, and the specific files the bug report showed as half-uninstalled) is completely unchanged after the simulated downstream failure. Verified this test fails against the pre-fix `install.sh` (reproducing the exact `D .loom/install-metadata.json` / `M CLAUDE.md` / `M .gitignore` pattern from the report) and passes against the fix.
- Added 4 new Rust unit tests in `scaffolding.rs` covering the AGENTS.md legacy-migration paths (markerless legacy, hybrid markered+legacy, malformed-marker, and a negative control for genuine user content).
- Wired the new bash suite into `defaults/scripts/tests/ci-wired.txt`.

## Not addressed (noted, out of scope)

- The cosmetic skills-provisioning path-mismatch warning (`~/.local/share/loom` checkout-layout disagreement) mentioned in the issue is lower priority and left as a follow-up — it didn't block the install and is unrelated to either fix above.

| `--full` install from a clean main source completes on a target with an existing older install, or fails BEFORE mutating the target | ✅ | The `--full` path no longer runs any uninstall/mutation against the target's main checkout before delegating; verified via the new regression test (simulated downstream failure leaves the target's tracked tree byte-for-byte unchanged) |
| AGENTS.md template substitution happens (or the daemon substitutes) — no placeholder refusal on the happy path | ✅ | All three AGENTS.md handling branches now discard leaked-placeholder legacy content instead of preserving it; covered by 3 new positive regression tests + 1 negative control |
| Any failure after target mutation restores the target's tracked files (git reset of the staged uninstall) as part of cleanup | ✅ | Achieved via the stronger guarantee of never mutating the target for the `--full` path in the first place (nothing to roll back); the `--quick` path's existing synchronous uninstall+restore is unchanged |
| Regression test: install-over-existing in a scratch repo, killed at the init step, leaves the scratch repo's tracked tree unchanged | ✅ | `test-install-full-reinstall-no-target-mutation.sh`, wired into CI |

## Test Plan

- `cargo test -p loom-daemon --lib init::scaffolding` — 62/62 passed, including the 4 new AGENTS.md legacy-migration tests
- `cargo fmt --check` / `cargo clippy -p loom-daemon --lib --tests -- -D warnings` — clean
- `bash defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` — 8/8 passed; manually verified it FAILS against the pre-fix `install.sh` (reproducing the exact half-uninstalled diff from the bug report) and PASSES against the fix
- `bash defaults/scripts/tests/test-install-reinstall-safety.sh`, `test-install-stash-scope.sh`, `test-install-source-guard.sh`, `test-install-pr-markers.sh`, `test-install-active-session.sh` — all still pass (no regressions in the adjacent reinstall/install-safety suites)
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` — clean (new suite wired)
- `bash -n install.sh` / `bash -n` on the new test file — clean

Closes #4888

### DIFF
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -54,6 +54,7 @@ test-gitignore-guard.sh
 test-guard-codex-bridge.sh
 test-guard-hook-schema.sh
 test-install-active-session.sh
+test-install-full-reinstall-no-target-mutation.sh
 test-install-pr-markers.sh
 test-install-source-guard.sh
 test-install-stash-scope.sh
--- a/defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh
+++ b/defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh
@@ -1,0 +1,179 @@
+#!/usr/bin/env bash
+# test-install-full-reinstall-no-target-mutation.sh — regression test for
+# issue #4888 defect 2.
+#
+# Before the fix, `install.sh`'s `--confirm-reinstall` flow ran a chained
+# `uninstall-loom.sh --yes --local "$TARGET_PATH"` directly against the
+# TARGET's main checkout for BOTH the --quick and --full reinstall paths,
+# unconditionally -- staging deletions and stripping the Loom sections out of
+# CLAUDE.md / .gitignore in place -- before delegating the --full path via
+# `exec "$LOOM_ROOT/scripts/install-loom.sh" ...`. Because `exec` replaces the
+# current process, nothing in install.sh could roll that mutation back if the
+# delegated install-loom.sh run failed downstream (e.g. at `loom-daemon
+# init`), leaving the target's main checkout half-uninstalled with no
+# automatic recovery (the reported symptom: manual `git reset --hard HEAD`
+# required).
+#
+# The fix scopes that chained uninstall to the --quick path only (where the
+# fresh install runs synchronously in the same working tree and unstages/
+# restores as part of its own completion) -- the --full path never mutates
+# the target's main checkout at all before delegating, so a downstream
+# failure has nothing to roll back.
+#
+# Strategy: build a throwaway "fake LOOM_ROOT" whose install.sh is the REAL
+# script (symlinked) but whose scripts/install-loom.sh is a stub that always
+# fails -- deterministically simulating "the delegated Full Install workflow
+# failed" without needing a built loom-daemon binary, gh auth, or network
+# access. Run it against a scratch target repo seeded with an existing
+# (older) Loom install, and assert the scratch repo's tracked tree is
+# byte-for-byte unchanged afterward.
+#
+# Usage:
+#   bash defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+INSTALL_SH="$REPO_ROOT/install.sh"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+TESTS_RUN=0
+TESTS_PASSED=0
+TESTS_FAILED=0
+
+pass() { TESTS_PASSED=$((TESTS_PASSED + 1)); TESTS_RUN=$((TESTS_RUN + 1)); echo -e "  ${GREEN}PASS${NC}: $1"; }
+fail() {
+  TESTS_FAILED=$((TESTS_FAILED + 1)); TESTS_RUN=$((TESTS_RUN + 1))
+  echo -e "  ${RED}FAIL${NC}: $1"
+  [[ -n "${2:-}" ]] && echo "$2" | sed 's/^/      /'
+}
+
+if [[ ! -f "$INSTALL_SH" ]]; then
+  echo "ERROR: $INSTALL_SH not found" >&2
+  exit 1
+fi
+
+if ! command -v git &> /dev/null; then
+  echo -e "${YELLOW}SKIP${NC}: missing required dependency 'git'"
+  exit 0
+fi
+
+echo "================================================================"
+echo "Full-reinstall no-target-mutation-on-failure (issue #4888)"
+echo "================================================================"
+echo ""
+
+# --- 1. Build a fake LOOM_ROOT: real install.sh, stubbed install-loom.sh ---
+#
+# scripts/install/ (provision-daemon.sh, provision-hooks.sh, stash-scope.sh,
+# ...) and the real uninstall-loom.sh are symlinked in so the reinstall gate
+# behaves identically to a real checkout, regardless of which code path
+# (pre- or post-fix) this test happens to be pointed at. Only
+# scripts/install-loom.sh -- the Full Install delegate -- is replaced with a
+# stub that always fails, deterministically simulating "the delegated
+# workflow failed downstream" without needing a built loom-daemon binary, gh
+# auth, or network access.
+FAKE_ROOT="$(mktemp -d /tmp/loom-full-reinstall-fakeroot.XXXXXX)"
+mkdir -p "$FAKE_ROOT/scripts"
+ln -s "$REPO_ROOT/scripts/install" "$FAKE_ROOT/scripts/install"
+ln -s "$REPO_ROOT/scripts/uninstall-loom.sh" "$FAKE_ROOT/scripts/uninstall-loom.sh"
+ln -s "$REPO_ROOT/install.sh" "$FAKE_ROOT/install.sh"
+
+STUB_MARKER="STUB install-loom.sh: simulating a downstream Full Install failure"
+cat > "$FAKE_ROOT/scripts/install-loom.sh" <<EOF
+#!/usr/bin/env bash
+echo "$STUB_MARKER" >&2
+exit 1
+EOF
+chmod +x "$FAKE_ROOT/scripts/install-loom.sh"
+
+# --- 2. Seed a scratch target repo with an existing ("older") Loom install ---
+TARGET="$(mktemp -d /tmp/loom-full-reinstall-target.XXXXXX)"
+git -C "$TARGET" init --quiet
+git -C "$TARGET" config user.email "test@test.com"
+git -C "$TARGET" config user.name "Test"
+
+mkdir -p "$TARGET/.loom/roles" "$TARGET/.claude"
+echo '{"loom_version": "0.1.0"}' > "$TARGET/.loom/install-metadata.json"
+echo '{}' > "$TARGET/.loom/config.json"
+echo '{}' > "$TARGET/.claude/settings.json"
+cat > "$TARGET/CLAUDE.md" <<'EOF'
+<!-- BEGIN LOOM ORCHESTRATION -->
+This repository uses Loom.
+<!-- END LOOM ORCHESTRATION -->
+
+# My Project
+
+Hand-written project docs.
+EOF
+cat > "$TARGET/.gitignore" <<'EOF'
+node_modules/
+
+# BEGIN LOOM
+.loom/tokens/
+.loom/*.log
+# END LOOM
+EOF
+git -C "$TARGET" add -A
+git -C "$TARGET" commit -m "Initial commit with an older Loom install" --quiet
+
+PRE_INSTALL_SHA="$(git -C "$TARGET" rev-parse HEAD)"
+
+# --- 3. Run the (fake-rooted) installer's --full --confirm-reinstall path ---
+OUTPUT="$("$FAKE_ROOT/install.sh" --full --confirm-reinstall -y "$TARGET" < /dev/null 2>&1)"
+EXIT_CODE=$?
+
+if [[ $EXIT_CODE -ne 0 ]]; then
+  pass "install run exited non-zero (downstream failure propagated)"
+else
+  fail "install run exited 0 (expected the stub failure to propagate)" "$OUTPUT"
+fi
+
+if [[ "$OUTPUT" == *"$STUB_MARKER"* ]]; then
+  pass "run actually reached the Full Install delegation (stub marker seen)"
+else
+  fail "run never reached the Full Install delegation -- test setup is broken" "$OUTPUT"
+fi
+
+# --- 4. Assert the target's tracked tree is completely untouched ---
+POST_SHA="$(git -C "$TARGET" rev-parse HEAD)"
+if [[ "$POST_SHA" == "$PRE_INSTALL_SHA" ]]; then
+  pass "HEAD is unchanged"
+else
+  fail "HEAD moved (expected $PRE_INSTALL_SHA, got $POST_SHA)"
+fi
+
+STATUS_AFTER="$(git -C "$TARGET" status --porcelain)"
+if [[ -z "$STATUS_AFTER" ]]; then
+  pass "git status is clean -- no staged/unstaged diff from the pre-install HEAD"
+else
+  fail "git status is NOT clean after the failed full reinstall" "$STATUS_AFTER"
+fi
+
+# Belt-and-suspenders: the specific files the reported bug showed as
+# half-uninstalled must still carry their original committed content.
+for f in .claude/settings.json .loom/install-metadata.json CLAUDE.md .gitignore; do
+  if [[ -f "$TARGET/$f" ]] && git -C "$TARGET" diff --quiet -- "$f"; then
+    pass "$f present and unmodified"
+  else
+    fail "$f missing or modified" "$(git -C "$TARGET" status --short -- "$f")"
+  fi
+done
+
+rm -rf "$FAKE_ROOT" "$TARGET"
+
+echo ""
+echo "================================================================"
+echo "Tests run:    $TESTS_RUN"
+echo -e "Tests passed: ${GREEN}$TESTS_PASSED${NC}"
+if [[ "$TESTS_FAILED" -gt 0 ]]; then
+  echo -e "Tests failed: ${RED}$TESTS_FAILED${NC}"
+  exit 1
+fi
+echo "================================================================"
+echo "All tests passed."
--- a/defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh
+++ b/defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh
@@ -92,6 +92,38 @@ exit 1
 EOF
 chmod +x "$FAKE_ROOT/scripts/install-loom.sh"
 
+# --- 1b. Stub the toolchain deps install.sh gates on, so this suite is
+# host-independent ---
+#
+# install.sh runs a "Checking System Dependencies" gate (node / pnpm / cargo /
+# git) BEFORE it reaches any of the reinstall logic under test. A host missing
+# any one of them aborts the run early, and the "stub marker seen" assertion
+# below fails for a reason that has nothing to do with the behavior this suite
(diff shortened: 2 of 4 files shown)


rjwalters:
Code evaluation looks good — the diff is well-scoped (scaffolding.rs legacy-content migration reuses existing CLAUDE.md heuristics; install.sh correctly scopes the chained uninstall to the --quick path only, leaving the --full path's target untouched before delegating). New regression test verified locally (passes modulo a missing `pnpm` in this sandbox, which CI's Installer Integration Tests already covers and passed). CI is still running (Rust Unit Tests, Rust OTLP Feature Build, Shell Test Suites all pending) — releasing the claim and skipping; a later pass will re-evaluate once CI settles.

rjwalters:
## Review: changes requested

Reviewed the diff (`install.sh`, `loom-daemon/src/init/scaffolding.rs`, the new
`test-install-full-reinstall-no-target-mutation.sh` suite) against the two
defects in #4888. Independent checkout in `.loom/worktrees/issue-4888`.

### Blocking issue — CI failure is real and reproducible

`gh pr checks 4920` currently shows **"Shell Test Suites (hermetic)" failing**
(https://github.com/rjwalters/loom/actions/runs/30698419492/job/91365104363),
and `mergeStateStatus` is `UNSTABLE` as a result. This is not flaky — I
reproduced it by tracing the log:

```
FAIL  test-install-full-reinstall-no-target-mutation.sh   8s (exit 1)
...
✗ Missing required dependencies: pnpm
...
FAIL: run never reached the Full Install delegation -- test setup is broken
```

Root cause: `install.sh`'s pre-flight dependency check (`command -v pnpm`,
around line 799) runs *before* the reinstall-confirmation code this PR touches,
and its "install missing deps?" prompt does **not** honor `NON_INTERACTIVE`
(no `-y`/`--yes` bypass — the `read` at line ~839 always executes). In the
hermetic CI runner, which has git/node/cargo/gh but not pnpm, that `read`
under `< /dev/null` gets an empty answer, takes the default "yes, exit to
install deps" branch, and the script exits 1 **before ever reaching the
Full Install delegation the new test wants to exercise** — so the "stub
marker seen" assertion fails and the suite reports FAIL.

I confirmed the same test passes locally (8/8) on this machine, where `pnpm`
happens to be installed — which is presumably how the Builder's reported
local run ("8/8 passed") missed this; the test's correctness silently depends
on host tooling it never stubs or guards, unlike the same file's existing
`command -v git` skip-guard.

This needs one of:
- Stub out `pnpm`/`node`/`cargo` (whatever the hermetic runner lacks) via a
  fake-executable directory prepended to `PATH` inside `$FAKE_ROOT`, so the
  test's determinism doesn't depend on what's installed on the host — this
  matches the "hermetic" framing of the CI job it's wired into, and
- optionally, as a separate/follow-up concern: the dependency-check prompt at
  install.sh:839 not honoring `NON_INTERACTIVE` looks like a latent bug in its
  own right (a real `--full -y` run against a host missing `pnpm` would hang
  on stdin instead of failing fast or proceeding) — worth flagging even if out
  of this PR's stated scope.

Please fix and repush; happy to re-review immediately once
`gh pr checks` is green.

### Everything else checked out

- **`scaffolding.rs` AGENTS.md fix**: verified all three branches (well-formed
  markers, malformed markers, no markers) now correctly reuse
  `is_legacy_loom_managed_root` / `slice_is_discardable_legacy` — the same
  heuristics already vetted for CLAUDE.md's #3476/#3527 hybrid-legacy hazard
  (including the `slice_is_discardable_legacy` 200-line / multi-signature
  safety valve that prevents a long, genuinely-customized document from being
  discarded on the strength of one incidental legacy-looking line). The 4 new
  Rust tests cover markerless-legacy, hybrid, malformed-marker, and a negative
  control for genuine user content — good coverage, and CI's "Rust Unit Tests"
  / "Rust Linting & Build Check" both pass.
  - Minor nit (non-blocking): the doc comment directly above the AGENTS.md
    handling block (around `scaffolding.rs` L920-926) still says "No
    legacy-layout heuristic tests are needed... since AGENTS.md has no
    historical full-guide-in-root layout to migrate away from" — this is now
    stale/contradicted by the code immediately below it, which does reuse that
    heuristic. Worth a follow-up cleanup pass so it doesn't confuse the next
    reader.
- **`install.sh` scoping fix**: traced the control flow — the entire chained
  `uninstall-loom.sh --yes --local` block (stash-scope guard, config.json
  snapshot, uninstall call, quick-install run, git-index reconcile, stash
  restore) is now correctly nested entirely inside `if [[ "$INSTALL_TYPE" ==
  "1" ]]` ending in `exit 0`; the non-quick path falls through to "Default:
  delegate to Full Install" having never touched the target's main checkout.
  Verified in `scripts/install-loom.sh` (lines ~925-988) that its own
  idempotency check handles an *older* installed version by logging "Existing
  Loom installation detected... Upgrading to..." and proceeding into the
  isolated-worktree upgrade — it does not require (or benefit from) a
  pre-uninstalled main checkout, confirming the fix's premise. The `--quick`
  path itself is untouched other than being moved inside the same `if` block
  (diff is a pure reindent + one comment reorg there).
- **Regression test logic** (setting the CI failure aside): the fake-LOOM_ROOT
  + stub-`install-loom.sh` strategy is a good, deterministic way to simulate
  "downstream Full Install failure" without a built daemon binary or network
  access, and its assertions (HEAD unchanged, clean `git status`, byte-for-byte
  unmodified per-file diff) directly match the bug report's symptom set.
- No guard-hook / destructive-command / security concerns — the diff only
  touches installer control flow and template-substitution string handling,
  no new external command execution, no credential handling changes.

Labeling `loom:changes-requested`.

rjwalters:
Review feedback addressed in e4899005.

## Blocking finding — CI failure (fixed, verified on the real runner)

Your root-cause diagnosis was exactly right. Confirmed from the failing job's log: the hermetic runner has `cargo`/`node`/`git` but **not** `pnpm`, so `install.sh`'s dependency gate aborted before the reinstall logic under test:

```
✗ Missing required dependencies: pnpm
FAIL: run never reached the Full Install delegation -- test setup is broken
```

Fixed both layers you suggested:

**1. `install.sh` — the root-cause usability bug.** The missing-dependency gate now branches on `NON_INTERACTIVE` instead of always calling `read`. On a non-interactive run stdin is not a terminal, so the read returned immediately at EOF with an empty reply, fell into the `[Y/n]` "exit to install" default, and aborted — after printing a prompt nobody was there to answer, and (when stdin is a pipe rather than `/dev/null`) after silently consuming a byte of it. It now aborts via `error()` with a direct, greppable message. Same outcome (a missing required dependency is still fatal), but a real fix for any non-interactive install, not just this test.

**2. The test — made genuinely host-independent.** `node`/`pnpm`/`cargo` are stubbed onto `PATH` for the `install.sh` invocation. On the `--full` path under test, `install.sh` only runs `<tool> --version` on these before `exec`ing the (already stubbed) Full Install delegate, so version-printing stubs are sound.

I chose stubs over a `command -v pnpm` skip-guard deliberately: skipping would have left the #4888 regression **uncovered on exactly the runner that matters**, trading a red test for a silent one. `git` is deliberately *not* stubbed — it is used for real by the fixture and the `HEAD`/`git status --porcelain` assertions — and keeps its hard skip-guard.

### How this was verified (not "works on my machine")

I built an exact replica of the CI environment — a `PATH` with `git`/`node`/`cargo` present and `pnpm` absent — and ran both versions of the suite in it:

| Run (pnpm-less PATH) | Result |
|---|---|
| **Pre-fix** suite (from the PR head commit) | **FAIL** — reproduces the CI failure exactly: 7 passed, 1 failed, `run never reached the Full Install delegation` |
| **Fixed** suite | **PASS** — 8/8 |

Real-CI confirmation on this push: **Shell Test Suites (hermetic) → pass (4m36s)**, and `mergeStateStatus` moved `UNSTABLE` → `CLEAN`. All other checks pass.

## Minor nit — also fixed

The stale `scaffolding.rs` comment claimed no `is_legacy_loom_managed_root`-style heuristic was needed for AGENTS.md, contradicting the code below it. Rewritten to describe actual behavior: AGENTS.md has no historical full-guide-in-root layout, but it *does* need the same `is_legacy_loom_managed_root` / `slice_is_discardable_legacy` heuristics, because an interrupted prior install can leave a root AGENTS.md with stale unsubstituted template text and missing/malformed markers.

## Other checks

`cargo test -p loom-daemon --lib init::scaffolding` **62/62**; `cargo fmt --all --check` clean; sibling install suites unaffected (`reinstall-safety` 15/15, `stash-scope` 7/7, `verify-install-scope` 16/16); `check-ci-suite-manifest.sh` OK; `shellcheck` output byte-identical to the pre-fix baseline (no new findings). Diff stays scoped to the three files: +53/-3.

## Note for the record

The repo build gate could not be run to completion locally — it stalls indefinitely on unrelated `metrics_collector::tests::test_check_env_enabled_*` and `safehouse::tests::*` cases (">running for over 60 seconds", host-specific, untouched by this diff, which is comment-only on the Rust side). CI's **Rust Unit Tests** job passed on this push, confirming those hangs are a local-host artifact and not a regression. Flagging rather than silently ignoring; not chasing it here as it is out of scope.

Also note: my `loom:treating` claim was auto-reaped by the daemon's stale-claim backstop at the 61-minute mark while that build gate was stalled. No competing Doctor was involved (no intervening comments or claim labels), so I re-claimed and completed normally.

Ready for re-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T13:17:03Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Both defects from #4888 are fixed correctly and match the PR description precisely.

**Defect 1 (AGENTS.md placeholder refusal)**: All three root-`AGENTS.md` handling branches (marker-replace, malformed-marker, no-markers) now correctly reuse the existing `is_legacy_loom_managed_root` / `slice_is_discardable_legacy` heuristics (verified these are pre-existing helpers used by CLAUDE.md, not reimplemented) to discard leaked-placeholder legacy content instead of preserving it verbatim. The hybrid-legacy case (marker present, but content outside the marker block is itself stale) is also handled. Four new Rust unit tests cover markerless-legacy, hybrid-legacy, malformed-marker, and a negative control that legitimate hand-authored content is still preserved.

**Defect 2 (no rollback on target mutation)**: The uninstall+stash+config-snapshot block that used to run unconditionally before the `INSTALL_TYPE=="1"` (quick) vs. Full Install branch now lives entirely inside the quick-install `if` block, ending in `exit 0`. The Full Install path (which `exec`s into `scripts/install-loom.sh` and does all its real work in an isolated worktree) now falls through without ever mutating the target's main checkout — matching the stated rationale that `install-loom.sh`'s own idempotency check + `loom-daemon init`'s merge-aware upgrade already handle an existing install safely from inside the worktree. Traced the full quick-install block after the restructure to confirm no logic was dropped in the move (stash scoping, config snapshot restore, git-index reconciliation, stashed-user-change restore) — it's a straight reindent/re-scope, not a rewrite.

**Test execution:**
- `bash defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` — 8/8 passed
- `bash defaults/scripts/tests/test-install-reinstall-safety.sh` — 8/8 passed (no regression)
- `bash defaults/scripts/tests/test-install-stash-scope.sh` — 7/7 passed (no regression)
- `bash defaults/scripts/tests/test-install-source-guard.sh` — 14/14 passed (no regression)
- `bash defaults/scripts/tests/test-install-pr-markers.sh` — 28/28 passed (no regression)
- `bash defaults/scripts/tests/test-install-active-session.sh` — 37/37 passed (no regression)
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` — clean, new suite correctly wired
- `bash -n install.sh` — clean
- CI: all checks pass (Rust Unit Tests, Installer Integration Tests, Shell Test Suites, Rust Linting & Build, Shellcheck, etc.)
- Merge state: CLEAN

PR description correctly uses "Closes #4888". Code quality, test coverage, and documentation are all solid.

rjwalters:
Standing down: while completing my own independent re-review of e4899005, another Judge pass already landed a verdict — `loom:pr` is now present (labeled 2026-08-01T13:20:52Z) and `loom:reviewing`/`loom:review-requested` were removed. Not overwriting.

For the record, my independent verification reached the same conclusion (approve) via a different path, so no discrepancy to flag:

1. **CI**: `gh pr checks 4920` — all 11 checks green, `mergeStateStatus` CLEAN.
2. **install.sh fix**: reproduced the original bug by checking out the pre-fix commit (766e1b7d) and running the new test against a PATH replica with node/cargo/git but no pnpm — reproduces the exact CI failure (1/8 fail, "run never reached the Full Install delegation"). Confirmed the fixed commit (e4899005, PR HEAD) passes 8/8 in the same replica. `error()` exits 1, so the non-interactive missing-dependency path is still fatal (same outcome as before), just via a direct message instead of a phantom `read` at EOF — no interactive-user regression, since the new branch is gated strictly inside `NON_INTERACTIVE == true`.
3. **Test not vacuous**: reverted just the install.sh reinstall-logic fix (766e1b7d) while keeping the new hermetic node/pnpm/cargo stubs, and reran — the test correctly fails again (2/8, target's `.loom/config.json`/`install-metadata.json` deleted and `CLAUDE.md`/`.gitignore` mutated), confirming the stubs only bypass the unrelated dependency gate and the test still pins the actual #4888 regression (target-checkout mutation on a failed Full Install).
4. **scaffolding.rs comment**: read the surrounding code — the marker-replace branch does call `slice_is_discardable_legacy`, and the malformed-marker branch does call `is_legacy_loom_managed_root`, so the corrected comment accurately describes existing behavior (pure doc fix, no logic change).
5. **Real CI log**: cross-checked the actual "Shell Test Suites (hermetic)" job log (`gh run view ... --log`) — confirms `PASS test-install-full-reinstall-no-target-mutation.sh`.
6. `cargo test -p loom-daemon --lib init::scaffolding` — 62/62 passed locally, matching Doctor's report.

No new defects found in the fix commit.

rjwalters:

- **Blast radius**: modifies `install.sh`'s reinstall control flow — the load-bearing hunk restructures which branch (`INSTALL_TYPE=="1"` quick-reinstall only, vs. previously all reinstalls) runs the chained `uninstall-loom.sh --yes --local` mutation against the target's main checkout, and reworks scoping/indentation around the config-snapshot and stash-guard logic in the same block. `install.sh` is exactly the class of file criterion #2's blast-radius axis reserves for human review regardless of diff size or quality — it is the entry point every Loom install/reinstall on any machine runs through, and a subtle regression in the branch condition could either reintroduce the reported half-uninstalled-target bug (#4888) or newly break the quick-reinstall path this refactor didn't intend to touch.

The diff itself is well-scoped and I can name the exact hunk, the new regression test is non-vacuous (verified to fail pre-fix and pass post-fix, including confirming it isn't bypassed by the hermetic dependency stubs), and the Judge's review is unusually thorough — independently reproduced the original bug against the pre-fix commit, confirmed the fix against a PATH replica, and cross-checked the real CI log. This hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4920`

---
*Automated by Champion role*

PR #4922, opened 2026-08-01
fix(install): skip daemon rebuild when installed binary already matches source HEAD

### LINKED ISSUE #4897: install.sh: daemon-staleness rebuild blocks silently on the cargo build-dir lock
## Context

Installing loom into a new repo (quick install, 2026-07-31) while the
fleet was busy: install.sh decided the loom-daemon binary was stale and
ran `cargo build --package loom-daemon --release`, which sat 10+ minutes
in "Blocking waiting for file lock on build directory" — the lock was held
by the fleet's own sweeps running cargo test/clippy in the source tree.
The installer produced no output during the wait (it looks hung; a first
attempt was killed for that reason), and after the wait the install log
printed "[loom-daemon] already current at ~/.local/bin/loom-daemon" — the
rebuild wasn't needed for the installed binary at all.

- [ ] Skip the rebuild when the installed binary's embedded commit already
      matches the source HEAD (the check that later printed "already
      current" should run before the build, not after)
- [ ] While waiting on the cargo lock, emit a periodic progress line
      naming the wait (and ideally the competing cargo PID) so the
      installer is distinguishable from a hang

## Affected Files

- `install.sh` - `loom_daemon_binary_stale()` (~line 360) compares the
  *source tree's* build artifact (`$LOOM_ROOT/target/release/loom-daemon`)
  against source file mtimes; both call sites that act on it —  the
  `--quick` reinstall path (~line 971) and the main "Quick Install" method
  (~line 1355) — unconditionally run `pnpm daemon:build` when it reports
  stale, with no check against what is already installed at the
  destination. Add a check before that build call that compares the
  *installed* binary's embedded commit (`<dest_bin> --version`) against
  source `HEAD`, and skip `pnpm daemon:build` entirely when they already
  match; while a build does run, emit a periodic progress line so a
  lock-contended wait is distinguishable from a hang.
- `scripts/install/provision-daemon.sh` - `provision_machine_daemon()`
  (~line 228) already does a post-build version-aware short-circuit (the
  "already current at $dest_bin" message, ~line 271) comparing `src_bin
  --version` to `dest_bin --version` — this is the existing comparison
  logic to reuse/hoist earlier in `install.sh`, not a file that itself
  needs new logic.
- `package.json` (line 11) - `daemon:build` script
  (`cargo build --package loom-daemon --release && ...`), reference only:
  this is the command whose lock wait produced the silent 10+ minute
  stall; no change expected here, the fix is in when/whether `install.sh`
  invokes it and what it prints while waiting.

## Test Plan

- [ ] Manual verification: hold the `target/` build-dir lock in a second
      shell (e.g. run `cargo build -p loom-daemon --release` there and
      keep it running), then start `install.sh --quick` against a repo
      whose installed `~/.local/bin/loom-daemon` already matches source
      `HEAD`. Confirm the rebuild is skipped entirely (no lock wait) rather
      than blocking on the lock and only then reporting "already current".
- [ ] Manual verification: repeat with the installed binary genuinely
      behind `HEAD` (e.g. after a `git pull`) and confirm the build still
      runs, and now prints a periodic progress line while waiting on a
      contended lock instead of producing no output.
- [ ] Automated tests: extend `tests/install/test-provision-daemon.sh` (or
      add a new `tests/install/` case) to cover the pre-build short-circuit
      path added to `install.sh`. The lock-wait progress line is UX-only
      and best verified manually per above.

rjwalters:
## Summary

Fixes #4897: `install.sh` unconditionally ran `pnpm daemon:build` (`cargo build --package loom-daemon --release`) whenever `loom_daemon_binary_stale()` reported the source-tree's own `target/release/loom-daemon` as stale-by-mtime — even when a machine-level binary was already installed and built from the exact commit in question. The build then blocked for 10+ minutes on a contended `target/.cargo-lock` (held by an unrelated `cargo test`/`clippy` in the same source tree) with zero output, only to print `[loom-daemon] already current at ~/.local/bin/loom-daemon` *after* the wait — the rebuild was never needed.

## Changes

- `install.sh`: adds `loom_daemon_dest_binary_current()`, which hoists the version-aware comparison `provision_machine_daemon()` already performs *after* a build (`scripts/install/provision-daemon.sh`'s "already current at `$dest_bin`" short-circuit, ~line 271) so it can run **before** the build instead. It compares the installed binary's embedded commit (`<dest_bin> --version`, baked in by `loom-daemon/build.rs`) against source `HEAD` (`git rev-parse --short HEAD`), and on a match copies the installed binary into `$LOOM_ROOT/target/release/loom-daemon` so the unmodified downstream call sites (`loom-daemon init`, `provision_machine_daemon`) keep working without further changes.
- Both call sites that acted on `loom_daemon_binary_stale()` — the `--quick` reinstall path and the main Quick Install method — now check `loom_daemon_dest_binary_current()` first and skip `pnpm daemon:build` entirely on a match.
- Adds `run_daemon_build_with_progress()`: when a build genuinely is needed, it now runs in the background with a periodic (15s) progress line, naming the elapsed time and — when `lsof` is available — the PID currently holding `target/.cargo-lock`, so a lock-contended wait is distinguishable from a hang.
- Adds `tests/install/test-daemon-build-shortcut.sh`, extending the existing extract-and-eval test pattern (`tests/install/test-hooks-preserve.sh`) to unit-test `loom_daemon_dest_binary_current()`: missing dest binary, mismatched commit, `unknown` commit, unparsable `--version`, matching commit (skip + copy), and the default `$HOME/.local/bin` dest-dir resolution. 10/10 pass.

## Test plan

- [x] `bash tests/install/test-daemon-build-shortcut.sh` — 10/10 pass
- [x] All existing `tests/install/*.sh` suites still pass (`test-forge-detect.sh`, `test-git-repo-detection.sh`, `test-hooks-preserve.sh`, `test-provision-daemon.sh`, `test-uninstall-labels-block.sh`, `test-uninstall-preserve-siblings.sh`)
- [x] `bash scripts/test-installer.sh` — full install/uninstall/reinstall integration suite passes end-to-end (exit 0)
- [x] `shellcheck install.sh tests/install/test-daemon-build-shortcut.sh` — no new warnings
- [x] `bash -n install.sh` — syntax OK
- [ ] Manual verification (not run in this sandbox — no second host/shell available to hold a real `target/.cargo-lock`): hold the lock in a second shell, run `install.sh --quick` against a repo whose installed `~/.local/bin/loom-daemon` matches source HEAD, confirm the rebuild is skipped with no lock wait; repeat with the installed binary behind HEAD and confirm the build runs with periodic progress output.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -383,6 +383,107 @@ loom_daemon_binary_stale() {
   [[ -n "$newer_file" ]]
 }
 
+# Issue #4897: `loom_daemon_binary_stale` above only ever compares the
+# *source tree's own build artifact* (`$loom_root/target/release/loom-daemon`)
+# against source mtimes -- it has no idea a machine-level binary already
+# installed at the destination (`provision_machine_daemon`'s target,
+# scripts/install/provision-daemon.sh) was already rebuilt from the exact
+# commit we're about to build, e.g. by a fleet-mate's install/sweep that ran
+# moments ago. Both call sites unconditionally ran `pnpm daemon:build` --
+# i.e. `cargo build --package loom-daemon --release` -- whenever the source
+# artifact looked stale, even when that build would just block for minutes on
+# `target/.cargo-lock` (held by an unrelated `cargo test`/`clippy` in the same
+# tree) and then produce a binary identical to one already on disk.
+#
+# Returns 0 ("current -- skip the build") when a machine-level installed
+# binary exists AND its embedded build commit (`--version`, see
+# loom-daemon/build.rs / loom_daemon::self_update::BUILD_IDENTITY) matches
+# source HEAD. This is the SAME comparison `provision_machine_daemon()`
+# already performs (scripts/install/provision-daemon.sh ~line 268, the
+# "already current at $dest_bin" short-circuit) -- reused here so install.sh
+# can skip the build BEFORE paying for the lock wait, not just discover after
+# the fact that the rebuild wasn't needed. On a match it also copies the
+# installed binary into `$loom_root/target/release/loom-daemon` (creating
+# target/release/ if missing) so the unmodified downstream call sites that
+# reference that in-repo path directly (`loom-daemon init`,
+# `provision_machine_daemon` itself) keep working without further changes.
+#
+# Returns 1 (build still needed) when there is no installed binary, its
+# `--version` can't be read, its embedded commit is "unknown"/unparsable, or
+# it does not match source HEAD.
+loom_daemon_dest_binary_current() {
+  local loom_root="$1"
+  local dest_dir="${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}"
+  local dest_bin="$dest_dir/loom-daemon"
+
+  [[ -x "$dest_bin" ]] || return 1
+
+  local head_commit
+  head_commit="$(git -C "$loom_root" rev-parse --short HEAD 2>/dev/null)" || return 1
+  [[ -n "$head_commit" ]] || return 1
+
+  local dest_version dest_commit
+  dest_version="$("$dest_bin" --version 2>/dev/null)" || return 1
+  # `--version` embeds "... (commit <sha>, built <ts>)" -- see
+  # loom-daemon/build.rs and loom_daemon::self_update::BUILD_IDENTITY.
+  case "$dest_version" in
+    *"(commit "*)
+      dest_commit="${dest_version#*"(commit "}"
+      dest_commit="${dest_commit%%,*}"
+      ;;
+    *) return 1 ;;
+  esac
+  [[ -n "$dest_commit" && "$dest_commit" != "unknown" ]] || return 1
+  [[ "$dest_commit" == "$head_commit" ]] || return 1
+
+  mkdir -p "$loom_root/target/release" 2>/dev/null || return 1
+  cp -f "$dest_bin" "$loom_root/target/release/loom-daemon" 2>/dev/null || return 1
+  chmod 755 "$loom_root/target/release/loom-daemon" 2>/dev/null || true
+  return 0
+}
+
+# Issue #4897: runs `pnpm daemon:build` (a `cargo build --package loom-daemon
+# --release` under the hood) in the background and emits a periodic progress
+# line while it is running, so a wait contended on `target/.cargo-lock` (held
+# by an unrelated `cargo test`/`clippy` elsewhere in the same source tree) is
+# visibly distinguishable from a hung installer instead of producing zero
+# output for 10+ minutes. Best-effort: if `lsof` is unavailable or the lock
+# file can't be inspected, the progress line still fires, just without naming
+# the competing PID.
+run_daemon_build_with_progress() {
+  local loom_root="$1"
+  local lockfile="$loom_root/target/.cargo-lock"
+  local log
+  log="$(mktemp 2>/dev/null || echo "/tmp/loom-daemon-build.$$.log")"
+
+  ( cd "$loom_root" && pnpm daemon:build >"$log" 2>&1 ) &
+  local build_pid=$!
+
+  local elapsed=0
+  local interval=15
+  while kill -0 "$build_pid" 2>/dev/null; do
+    sleep "$interval"
+    elapsed=$((elapsed + interval))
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review — Approve

Reviewed against issue #4897's two acceptance criteria, both met:

1. **Pre-build short-circuit.** `loom_daemon_dest_binary_current()` hoists the exact commit comparison `provision_machine_daemon()` already did post-build (`scripts/install/provision-daemon.sh`'s "already current at $dest_bin" check) so it now runs *before* `pnpm daemon:build` at both call sites (`--quick` reinstall and main Quick Install). Verified by reading both call sites (install.sh:1085-1103, 1477-1495) — the new check gates entry into the build/lock-wait path, not just the log message.
2. **Progress line while waiting on the lock.** `run_daemon_build_with_progress()` backgrounds `pnpm daemon:build`, polls every 15s with `kill -0`, and on each tick prints elapsed time plus (when `lsof` is available) the PID holding `target/.cargo-lock` — matches the documented cargo build-dir lock path from the original incident report.

Verification performed in the existing `.loom/worktrees/issue-4897` (already at the PR's HEAD `2e59ecf0`):
- `bash tests/install/test-daemon-build-shortcut.sh` → 10/10 pass.
- `bash -n install.sh` → syntax OK; `shellcheck install.sh tests/install/test-daemon-build-shortcut.sh` → only pre-existing warnings on lines this PR doesn't touch (SC2043/SC2034/SC2086 at 1180/1375/1512/1602).
- Cross-checked the commit-extraction logic (`case "$dest_version" in *"(commit "*)...`) against `loom-daemon/build.rs` and `self_update.rs`'s `BUILD_IDENTITY` format (`"<ver> (commit <sha>, built <ts>)"`, `git rev-parse --short HEAD`) — the parsing matches the real embedded format.
- `bash tests/install/test-provision-daemon.sh` had 2 flaky codesign/keychain failures on first run, unrelated to this diff (doesn't touch provision-daemon.sh) — reran 3x clean (40/40) and confirmed the same flake pattern is host-state-dependent, not PR-induced.
- The `cd "$LOOM_ROOT"` that used to run in the parent shell before `pnpm daemon:build` now only runs inside `run_daemon_build_with_progress`'s backgrounded subshell — a deliberate (and correct) side-effect-free improvement, since every call site in install.sh already addresses paths via `$LOOM_ROOT`/`$TARGET_PATH` explicitly rather than relying on cwd.
- `gh pr checks` — all required checks pass; the Rust/dashboard/MCP/installer/shell-suite legs show `skipping`, which is expected: this diff only touches `install.sh` + `tests/install/`, neither of which is in the `scripts` path-filter in `.github/workflows/ci.yml` (pre-existing gap covering all 7 files under `tests/install/`, not something this PR introduces or should be scoped to fix). Independently ran the new suite locally per above.
- `mergeStateStatus`: CLEAN.

One non-blocking note for a future issue (not this PR's scope): `loom_daemon_dest_binary_current()` matches on the installed binary's embedded HEAD commit only, so a source tree with *uncommitted* local edits to `loom-daemon/` combined with `install.sh --quick` could now silently skip a rebuild that the old mtime-only staleness check would have triggered (since the installed binary's commit still equals `HEAD`, even though the working tree differs from it). Narrow edge case — the mainline fleet workflow (clean checkouts) is unaffected, and it's the exact scenario the issue was filed to fix — but worth a comment if this ever bites a developer iterating locally on loom-daemon via install.sh instead of `cargo build` directly.

Nice, well-scoped fix with solid test coverage. Approving.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present — another Judge's verdict raced mine. Discarding my review; not writing any label.

For reference, my independent review reached the same conclusion (approve): all 10 new tests in `tests/install/test-daemon-build-shortcut.sh` pass, all existing `tests/install/*.sh` suites pass (one pre-existing, unrelated flaky failure in `test-provision-daemon.sh` reproduced intermittently on main too, not introduced by this PR), `shellcheck`/`bash -n` show no new warnings, and the PR correctly references `Fixes #4897`. Also filed #4924 — a pre-existing CI `paths-filter` gap where root `install.sh` and `tests/install/**` never trigger `Installer Integration Tests`, discovered because this PR's own new test file got silently skipped by CI.

PR #4923, opened 2026-08-01
fix(daemon): gate loom-daemon clean tmux sessions on --safe and attached clients

### LINKED ISSUE #4890: loom-daemon clean --safe kills a live tmux session — --safe gates worktrees by merge state but not tmux
## Context

Found during a `/repo:all` hygiene pass on 2026-08-01, dry-running cleanup on a host with 16 sweeps in flight.

`--safe` is documented as the merged-PR-only mode, and for worktrees it behaves exactly that way — it gates on merge state *and* on liveness:

```
  Would clean: 16 worktree(s)
  Skipped (in use by shepherd): 4
  Skipped (open/not merged): 3
  Skipped (uncommitted): 2
```

Tmux gets neither gate:

```
Cleaning Loom Tmux Sessions

Found Loom tmux sessions:
  - loom-isolated-terminal-default-0
  Would kill: 1 tmux session(s)
```

That session is live. It is on Loom's isolated socket rather than the default one, so it does not show up in a plain `tmux ls`:

```
$ tmux ls
error connecting to /private/tmp/tmux-501/default (No such file or directory)
```

## Why this is a safety-contract gap, not just a preference

A tmux session is not an artifact of a merged PR, so "merged-PR-only mode" has nothing to say about it — yet `--safe` kills it anyway. The mode's promise and its behaviour disagree.

The concrete risk is that a Manual-Orchestration-Mode terminal is exactly this kind of session: an operator's interactive agent, possibly mid-task, with no PR association at all. `--safe` is the mode a cautious operator reaches for *because* they have live work; it is the worst mode in which to silently kill a terminal.

It is also invisible in the pre-flight: a user checking `tmux ls` before running cleanup sees no sessions, because the isolated socket is not the default one. The dry-run does name the session — good — but the summary line reduces it to a count with no indication that `--safe` did not apply.

## Not established

Whether the tmux sweep has any liveness/attachment check at all (e.g. skip sessions with an attached client), or whether it matches purely on the `loom-` name prefix. Worth confirming before choosing between "gate it under `--safe`" and "add a liveness test for every mode".

- [ ] `--safe` either skips tmux cleanup entirely, or restricts it to sessions provably tied to a completed sweep
- [ ] A session with an attached client is never killed without an explicit opt-in flag
- [ ] The summary distinguishes killed from skipped for tmux, as it already does for worktrees
- [ ] The docs state which artifact classes `--safe` actually narrows — currently it reads as though it narrows everything

rjwalters:
## Summary

`loom-daemon clean --safe` gates worktrees by merge state (issue closed + PR
merged + grace period) but killed every Loom tmux session unconditionally —
a tmux session has no PR/merge state for "merged-PR-only mode" to check, so
`--safe` silently killed live sessions, including a Manual-Orchestration-Mode
operator terminal on Loom's isolated `-L loom` tmux socket (invisible to a
plain `tmux ls`, since it's not the default socket).

## Changes

- `--safe` (without `--force`) now skips tmux cleanup entirely rather than
  pretending to gate it — a tmux session simply has no PR to check merge
  state against.
- Outside `--safe`, a tmux session with an attached client is never killed
  without the explicit `--force` opt-in (checked via `tmux list-clients`).
- Extracted the gating logic into a pure, unit-tested `classify_tmux_session`
  decision function (mirrors `classify_worktree`'s split of decision from
  action) with 6 new unit tests covering safe/force/attached combinations.
- The dry-run/summary output now distinguishes killed vs. skipped-under-safe
  vs. skipped-attached for tmux sessions, matching the existing worktree
  summary granularity.
- Documented which artifact classes `--safe` actually narrows (worktrees +
  branches — both have a merged-PR concept; tmux does not) in the `Clean`
  CLI doc comment (`loom-daemon clean --help`) and
  `defaults/docs/troubleshooting.md`.

## Test plan

- [x] `cargo test -p loom-daemon --lib worktree_ops::clean::` — 33 passed
      (including 6 new `tmux_*` gating tests)
- [x] `cargo clippy -p loom-daemon --lib -- -D warnings` — clean
- [x] `cargo fmt -p loom-daemon -- --check` — clean

Closes #4890

### DIFF
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -82,6 +82,18 @@ if it errors out instead of running.
 loom-clean --force  # Non-interactive, safe for automation
 ```
 
+**What `--safe` actually narrows (#4890)**: `--safe` is documented as
+"merged-PR-only mode", but that promise only has meaning for artifacts that
+*are* an artifact of a merged PR — worktrees (gated on issue-closed +
+PR-merged + grace period) and branches. A tmux session has no PR association
+at all, so `loom-clean --safe` skips tmux cleanup **entirely** rather than
+silently killing a live session (a Manual-Orchestration-Mode terminal on
+Loom's isolated `-L loom` tmux socket does not show up in a plain `tmux ls`,
+so an operator has no other way to notice). Use `loom-clean --tmux-only`
+(optionally with `--force`) outside `--safe` to clean tmux sessions
+explicitly. Even then, a session with an attached client (someone is actively
+looking at it) is preserved unless `--force` is passed.
+
 **Backlog of pre-existing `[gone]` local branches (#4100)**: `merge-pr.sh` deletes
 the local feature branch for every PR it merges as of #4100, but repos that ran
 Loom before that fix accumulated one orphaned local branch per merged issue —
--- a/loom-daemon/src/main.rs
+++ b/loom-daemon/src/main.rs
@@ -495,6 +495,15 @@ enum Commands {
     /// decision tree from issue #3332. Purely file/git/gh-based; does not
     /// require a running daemon. Flags mirror the retired `loom-clean`
     /// console script byte-for-byte.
+    ///
+    /// **What `--safe` actually narrows** (issue #4890): only the artifact
+    /// classes that have a merged-PR concept to check — worktrees (gated on
+    /// issue-closed + PR-merged + grace period) and branches. Tmux sessions
+    /// are not an artifact of a merged PR, so `--safe` skips tmux cleanup
+    /// entirely rather than pretending to gate it; pair with `--tmux-only`
+    /// (optionally `--force`) to clean tmux sessions explicitly. Outside
+    /// `--safe`, a tmux session with an attached client (a live operator
+    /// terminal) is likewise preserved unless `--force` is passed.
     Clean {
         /// Workspace directory (repo root, or any path under it).
         #[arg(long, value_name = "PATH", default_value = ".")]
@@ -509,6 +518,9 @@ enum Commands {
         #[arg(short = 'f', long, visible_alias = "yes", visible_short_alias = 'y')]
         force: bool,
 
+        /// Merged-PR-only mode for worktrees/branches. Tmux sessions have no
+        /// PR association, so `--safe` skips tmux cleanup entirely (see the
+        /// `Clean` doc comment) rather than silently killing a live session.
         #[arg(long)]
         safe: bool,
 
--- a/loom-daemon/src/worktree_ops/clean.rs
+++ b/loom-daemon/src/worktree_ops/clean.rs
@@ -113,6 +113,10 @@ pub struct CleanupStats {
     pub kept_branches: usize,
     pub errored_branches: usize,
     pub killed_tmux: usize,
+    /// Tmux sessions preserved because they have an attached client (a live
+    /// operator terminal) or because `--safe` mode does not touch tmux at all
+    /// (issue #4890).
+    pub skipped_tmux: usize,
     pub cleaned_config_dirs: usize,
     pub cleaned_sweep_baselines: usize,
     pub cleaned_sweep_checkpoints: usize,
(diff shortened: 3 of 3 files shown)


rjwalters:
LGTM — verified locally in `.loom/worktrees/issue-4890`.

**Verification performed:**
- `cargo test -p loom-daemon --lib worktree_ops::clean::` — 33 passed (0 failed), including all 6 new `tmux_*` gating tests (`tmux_safe_mode_skips_even_an_unattached_session`, `tmux_safe_mode_skips_an_attached_session_too`, `tmux_force_overrides_safe_mode`, `tmux_attached_session_skipped_outside_safe_mode`, `tmux_force_overrides_attached_gate`, `tmux_unattached_session_killed_by_default`) — the Builder's claim checks out, these aren't just claimed.
- `cargo clippy -p loom-daemon --lib -- -D warnings` — clean
- `cargo fmt -p loom-daemon -- --check` — clean
- `git merge-tree --write-tree origin/main HEAD` — merges cleanly, no conflict markers
- `gh pr checks 4923` — all required checks pass (Rust Unit Tests, Rust Linting & Build Check, Rust OTLP Feature Build & Test, Shell Test Suites, Installer Integration Tests, AGENTS.md Sync, CLAUDE.md Line Budget, Docs/Defaults Parity Check); Dashboard/MCP jobs correctly skip (no relevant files changed)
- `gh pr view --json mergeStateStatus` — `CLEAN`

**Correctness against issue #4890's acceptance criteria:**
- `classify_tmux_session(safe, attached, force) -> TmuxDecision` is a pure decision function mirroring `classify_worktree`'s shape — easy to unit-test and reason about independent of a real tmux server.
- `--safe` (without `--force`) now skips tmux cleanup entirely — one of the two AC-listed options ("skip entirely or restrict to sessions provably tied to a completed sweep"), and the doc comment/troubleshooting.md are explicit about *why* (no PR/merge-state concept for a tmux session).
- Attached-client sessions (checked via `tmux list-clients`) are never killed without explicit `--force`, independent of `--safe` — confirmed by `tmux_attached_session_skipped_outside_safe_mode` and `tmux_force_overrides_attached_gate`.
- New `CleanupStats::skipped_tmux` field, only printed when non-zero, cleanly distinguishes killed vs. skipped in the summary.
- The `--daemon` crash-recovery path (`clean_daemon_crash_state`) now routes through the same `clean_tmux_sessions(&mut stats, &tmux_opts)` gate instead of the old bare `dry_run`-only signature, so an attached session survives unattended crash recovery too.
- CLI help text (`Clean` doc comment on `--safe`) and `defaults/docs/troubleshooting.md` both updated to explain the narrowed scope.

**Minor, non-blocking observation:** `session_has_attached_client` fails toward "not attached" (i.e., toward killing) on any `tmux` command error, per its own doc comment. Since `list-sessions` already confirmed the session is live moments earlier on the same socket, this should be rare in practice, but it's worth keeping in mind if a flaky-tmux report ever surfaces — not a reason to block this fix.

Approving.

rjwalters:
✅ **Approved!**

Reviewed in the existing builder worktree (`.loom/worktrees/issue-4890`).

- The fix is well-scoped: `--safe` now skips tmux cleanup entirely (a tmux session has no PR/merge-state concept to gate on), and outside `--safe` a session with an attached client is preserved unless `--force` is passed.
- Decision logic extracted into pure `classify_tmux_session`, mirroring `classify_worktree`'s split — good consistency with the existing pattern in this file.
- Verified both call sites of `clean_tmux_sessions` were updated (`clean_daemon_crash_state` correctly uses neither `--safe` nor `--force`, so an attached session is still preserved during crash recovery; `run_clean` passes through `opts`).
- Docs updated in both `defaults/docs/troubleshooting.md` and the `Clean` CLI doc comment (`main.rs`), consistent with each other.

## Test Execution

1. `cargo test -p loom-daemon --lib worktree_ops::clean::` — ✅ 33 passed (matches PR's claim, includes the 6 new `tmux_*` gating tests)
2. `cargo clippy -p loom-daemon --lib -- -D warnings` — ✅ clean
3. `cargo fmt -p loom-daemon -- --check` — ✅ clean

## CI Status

All required checks pass (`gh pr checks 4923`); merge state is `CLEAN`.

PR description correctly uses `Closes #4890`, and no stray closing keyword targets another issue.

Issue #4926, opened 2026-08-01
Bash-tool write-confinement is bypassed by quoting the target path


rjwalters:
## Summary

`guard-destructive-generic.sh`'s Bash-tool write-confinement check (#4178) is
bypassed entirely by simply **quoting** the write target. A quoted absolute
path into the main repository checkout is silently **allowed** from any cwd —
including a builder's own linked worktree, the exact configuration #4178
exists to protect.

This is **pre-existing on `main`** and is independent of #4921 (which is about
*unresolved `$VAR`* targets). It was found while implementing #4921; filing
separately per the "raise concerns as their own issue" convention rather than
widening that PR's blast radius on a security-critical file.

## Reproduction (against the real repo, no write performed)

```bash
G=defaults/hooks/guard-destructive-generic.sh
WT=<any .loom/worktrees/issue-N>

# Unquoted — correctly DENIED (worktree-write-confinement):
jq -n --arg cmd 'echo x > /abs/path/to/main/defaults/hooks/evil' --arg cwd "$WT" \
  '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' | bash "$G"

# Double-quoted — ALLOWED (exit 0, no decision):
jq -n --arg cmd 'echo x > "/abs/path/to/main/defaults/hooks/evil"' --arg cwd "$WT" \
  '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' | bash "$G"

# Single-quoted — also ALLOWED. Same for tee / cp / mv / sed -i:
#   tee '/abs/path/to/main/defaults/hooks/evil'
#   cp /tmp/a '/abs/path/to/main/defaults/hooks/evil'
```

Reproduces from a **main-checkout** cwd too, so unlike #4921 this is not
cwd-dependent — it defeats the check in every mode.

## Root cause

`qsplit()` (the shared tokenizer, #3755) preserves token values **verbatim,
including their quote characters** — that is its documented contract, and
`extract_rm_targets` depends on it. `extract_write_targets()` therefore emits
`'/main/evil'` (with quotes) as the target token, and the write-confinement
block never strips them:

- `[[ "$_wtarget" == /* ]]` is **false** (the token starts with `'` or `"`),
  so the absolute path is misclassified as **relative** and cwd-prefixed into
  `<cwd>/'/main/evil'`.
- From a worktree cwd that fabricated path hits the worktree's own
[...]

PR #4927, opened 2026-08-01
fix(hooks): fail closed on unresolvable $VAR write targets from a linked-worktree cwd

### LINKED ISSUE #4921: Bash-tool write-confinement: unresolvable $VAR write target silently allows when cwd is a linked worktree
## Summary

While reviewing PR #4914 (the `$VAR` same-command resolution fix for
`extract_write_targets()` in `defaults/hooks/guard-destructive-generic.sh`), I
found that the guard's fail-closed fallback for an **unresolvable** `$VAR`
write target (chained assignment, `$(...)`, inherited env var, or — per
#4914's own new poisoning logic — a name assigned two *different* values in
one command) is silently defeated whenever the guard's `cwd` is a genuine
**linked git worktree** rather than the main checkout itself.

This is **pre-existing** on `main` (reproduces identically on `0a8e18e2`,
before #4881/#4914 existed) — not a regression introduced by #4914. I'm
filing it separately per the Judge "Issues Are Suggestions" / "Raising
Concerns" convention rather than blocking that PR on it.

## Root cause

`extract_write_targets()`'s fallback for a target it cannot resolve is to
return the raw token unchanged (e.g. literal `$A/evil.sh`), which the caller
then treats as a **relative path and cwd-prefixes** it before checking
whether the result resolves inside the main checkout
(`defaults/hooks/guard-destructive-generic.sh`, the `WT_MAIN_ROOT`
containment check).

The problem: when `cwd` is itself a linked worktree (`.loom/worktrees/wt1`,
the canonical builder operating mode — `cd .loom/worktrees/issue-N`), the
cwd-prefixed literal path (e.g.
`.../.loom/worktrees/wt1/$A/evil.sh`) trivially walks back up into the
worktree's own `.loom-managed` sentinel via `_in_any_managed_worktree()`
**before** the main-root containment check ever runs — so the write is
silently **allowed**, regardless of what `$A` would actually resolve to at
runtime. The literal-fallback deny only works when `cwd` happens to equal the
main checkout root itself (the shape every existing test fixture in
`tests/hooks/test-guard-destructive.sh` uses), which is not how a builder
actually operates.

## Reproduction (isolated fixture, no real repo touched)

```bash
# Build a throwaway fixture: main repo + a properly-registered managed
# worktree at <main>/.loom/worktrees/wt1 (the layout _any_managed_worktree_exists
# expects).
rm -rf /tmp/fixture-main
mkdir -p /tmp/fixture-main/defaults/hooks && cd /tmp/fixture-main
git init -q && git config user.email t@t.com && git config user.name t
echo hi > README.md && mkdir -p .loom && echo '{}' > .loom/config.json
git add -A && git commit -q -m init
mkdir -p .loom/worktrees
git worktree add -q .loom/worktrees/wt1 -b fixture-branch
touch .loom/worktrees/wt1/.loom-managed

GUARD=/path/to/defaults/hooks/guard-destructive-generic.sh
FIXTURE_WT=/tmp/fixture-main/.loom/worktrees/wt1
TARGET=/tmp/fixture-main/defaults/hooks

# Sanity: a LITERAL (unobfuscated) write to the main checkout correctly denies.
jq -n --arg cmd "echo x > $TARGET/evil" --arg cwd "$FIXTURE_WT" \
  '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' \
  | LOOM_WORKTREE_PATH="$FIXTURE_WT" "$GUARD"
# -> denies (worktree-write-confinement), as expected.

# The masked case: an UNRESOLVABLE $VAR write target from the SAME cwd.
jq -n --arg cmd 'echo x > $SNEAK_NOT_ASSIGNED_ANYWHERE/evil' --arg cwd "$FIXTURE_WT" \
  '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' \
  | LOOM_WORKTREE_PATH="$FIXTURE_WT" "$GUARD"
# -> ALLOWS (exit 0, no deny) — the fail-closed fallback never engages.
```

The identical result (allow) reproduces against `defaults/hooks/guard-destructive-generic.sh`
at `0a8e18e2` (pre-#4881/#4914), confirming this is not something #4914
introduced.

Also confirmed to affect #4914's own new conflicting-assignment poisoning
path specifically (poisoned-to-unresolvable still hits the same masked
fallback):

```bash
jq -n --arg cmd "A=/tmp/outside
A=$TARGET
echo x > \$A/evil" --arg cwd "$FIXTURE_WT" \
  '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' \
  | LOOM_WORKTREE_PATH="$FIXTURE_WT" "$GUARD"
# -> ALLOWS, even though the second (last-wins) assignment textually targets
#    the main checkout and #4914's record_assign() poisoning is designed to
#    force exactly this shape to fail closed.
```

## Impact

This means the "unresolvable → fail closed (deny)" safety property that both
#4881 and #4914's Judge reviews explicitly verified and relied on as the
backstop for every unparseable `$VAR` shape is **only effective when the
acting cwd is the main checkout itself** — it does not protect the primary,
documented builder operating mode (`cd .loom/worktrees/issue-N`, then run
Bash). None of the 594+ assertions in `tests/hooks/test-guard-destructive.sh`
catch this because every existing fixture in that suite uses `cwd ==
REPO_ROOT` (the main checkout) rather than a genuine nested linked worktree.

## Suggested direction (not verified/implemented)

`_in_any_managed_worktree()`'s sentinel walk-up should probably not be
trusted for a target whose path is a **literal, unresolved token
concatenated onto cwd** (as opposed to a real filesystem path) — e.g. the
caller could special-case "target still contains an unexpanded `$`
character" and treat that shape as denial-worthy whenever a managed worktree
exists anywhere in the repo, independent of the cwd-prefixed containment
check. Left to whoever picks this up to design properly — flagging the gap
and repro, not prescribing the fix.

Discovered during Judge review of PR #4914.

rjwalters:
## Summary

The Bash-tool write-confinement guard's "unresolvable `$VAR` → fail closed"
backstop was silently defeated whenever the acting cwd was a genuine **linked
worktree** — i.e. the canonical builder operating mode (`cd
.loom/worktrees/issue-N`) and the only mode #4178 exists to protect.

`extract_write_targets()` is a tokenizer, not a shell evaluator: a target it
cannot resolve is emitted as the **raw token** (`$A/evil`), and the
confinement block then treated that literal as a *relative path* and
cwd-prefixed it. From a **main-checkout** cwd the fabricated path landed inside
the main checkout, so the containment test denied and the backstop looked
intact. From a **linked-worktree** cwd the very same fabricated path walked
straight back up into the acting worktree's own `.loom-managed` sentinel
(`_in_any_managed_worktree`) and was **allowed** before the main-root
containment test ever ran — whatever the variable would expand to at runtime.

Pre-existing on `main`; confirmed reproducing on the isolated fixture from the
issue body before the fix (all `$`-shape cases ALLOW) and denying after it.

## Changes

- **`mark_expandable_dollars()`** (new helper): normalizes a raw write-target
  token to its *effective path shape* — quote characters removed, backslash
  escapes applied, and every **expandable** `$` marked with SOH. A `$` inside a
  single-quoted span or backslash-escaped is literal data and stays a `$`,
  mirroring the quoted-tilde rule of `expand_leading_tilde()` (#4382).
- **Shape decision before containment.** A target is denied when its
  *location* (not merely its filename) is unknowable:
  - the token is a variable from the root down — `> $DEST`, `tee "${OUT}"`,
    `> $(mktemp)`, `> /$X`, `> /$X/evil`;
  - an expandable `$` sits in a **directory component** and the known prefix is
    inside the repo/worktree area (`> $A/evil`, `> ./$A/evil`, `cd $A && > f`,
    `> <worktree>/$A/f`), or there is no usable known prefix because it
    collapses to `/` (`> /tmp/../$A/evil` — the prefix is normalized *before*
    it is judged, so a `..` traversal cannot manufacture a benign prefix).
- **No new false positives** for shapes that lose no information: a `$` only in
  the final filename (`> out-$STAMP.log` — the directory is fully known, so the
  ordinary containment test still runs and still denies a main-checkout
  directory), a known prefix outside the protected area (`> /tmp/$D/f.log`),
  and a `$` a real shell would never expand (`> '$A/evil'`, `> \$A/evil`).
- **Fail-open contract preserved**: the new deny fires only when a managed
  worktree actually exists (`_wt_isolation_in_play`, factored out of the
  existing inline `_WT_WRITE_BASE` resolution and reused), and the
  `guards.worktreeIsolation` / `LOOM_GUARD_WORKTREE_ISOLATION` toggle still
  switches the whole check off.
- **26 new assertions** in `tests/hooks/test-guard-destructive.sh`, every one
  with `cwd` set to a genuine `git worktree add` linked worktree — the shape no
  existing fixture in the suite used for these commands, which is exactly why
  the gap was untested.
- `defaults/docs/guard-hooks.md`: documents the denied/not-denied shape table
  and the workaround (spell the path literally).

| Unresolvable `$VAR` (or #4914's poisoned-to-unresolvable shape) from a linked-worktree cwd is **denied**, not silently allowed | ✅ | Both repro scripts from the issue body run against the isolated fixture: ALLOW on `origin/main`, DENY after the fix (`worktree-write-confinement-unresolved-var`). Also covered by new assertions. |
| Does not weaken the legitimate case (a) — genuine writes inside a managed worktree still allowed, no false-positive deny | ✅ | `echo x > <worktree>/src/f.sh`, `cd <worktree> && echo x > src/f.sh`, `> out-$STAMP.log`, `> src/$f.txt`, `> /tmp/…-$STAMP/f.log`, `> '$A/evil'`, `> \$A/evil`, `git commit -m "price > $5"` all still ALLOW (asserted). |
| Suite gains ≥1 fixture with `cwd` set to a linked worktree covering the exact repro | ✅ | 26 new assertions, all `cwd == WT_LINKED_DIR` (`make_wt_repo_linked`), including the verbatim `echo x > $SNEAK_NOT_ASSIGNED_ANYWHERE/evil` and the conflicting-assignment shape. |
| All existing assertions continue to pass | ✅ | `tests/hooks/test-guard-destructive.sh`: **592 / 592 passed, 0 failed** (566 pre-existing + 26 new). Includes the #4178 / #4210 / #4245 / #4382 / #4495 write-confinement cases and the #3679 command-substitution smuggling floor. |

## Test Plan

- `bash tests/hooks/test-guard-destructive.sh` → **Total 592, Passed 592,
  Failed 0**.
- `bash tests/hooks/test-guard-destructive-dispatcher.sh` → **12 passed, 0
  failed**.
- `shellcheck -S warning` on both changed shell files → clean (one pre-existing
  SC2034 in the unrelated perf loop).
- Manual, against the isolated fixture from the issue body (throwaway repo +
  registered `.loom/worktrees/wt1` managed worktree), `origin/main` vs. this
  branch:

  | Case (cwd = linked worktree) | `origin/main` | this branch |
  |---|---|---|
  | literal write to main checkout (sanity) | DENY | DENY |
  | `echo x > $SNEAK_NOT_ASSIGNED_ANYWHERE/evil` | **ALLOW** | DENY |
  | conflicting same-command assignment → `$A/evil` | **ALLOW** | DENY |
  | `cat > $DEST` / `cat > $(mktemp)` | **ALLOW** | DENY |
  | `cd $A && echo y > f.sh` | **ALLOW** | DENY |
  | `echo x > /$SNEAK/evil` | **ALLOW** | DENY |
  | `> out-$STAMP.log`, `> /tmp/$D/f.log`, `> '$A/evil'`, `> \$A/evil` | ALLOW | ALLOW |
  | same unresolvable target from a **main-checkout** cwd | DENY | DENY |
  | same target in a repo with **no** managed worktree (fail-open) | ALLOW | ALLOW |

## Note for the reviewer (out of scope, filed separately)

While building the fixture I found a **separate, larger pre-existing bypass in
the same block**: simply **quoting** the target defeats write-confinement
entirely — `echo x > "/abs/path/into/main/checkout/evil"` is ALLOWED from any
cwd (including the main checkout), because `qsplit()` preserves quote
characters verbatim so `[[ "$_wtarget" == /* ]]` is false and the absolute path
is misclassified as relative. Verified on `origin/main` and against the real
repo. It is independent of the `$`-shape logic here, so I filed it as **#4926**
rather than widening this PR's blast radius on a security-critical file.

Closes #4921

### DIFF
--- a/defaults/docs/guard-hooks.md
+++ b/defaults/docs/guard-hooks.md
@@ -392,6 +392,41 @@ attempt to catch every conceivable write vector (an interpreter one-liner like
 fallback an agent reaches for after an Edit/Write denial, not building a full
 security boundary.
 
+**Unresolvable `$…` targets fail closed, in every cwd (issue #4921).** The
+tokenizer never expands variables, so a target it cannot resolve is emitted as
+the raw token (`$A/evil`) and the resolution then cwd-prefixes it as if it
+were a relative path. From a **main-checkout** cwd that fabricated path landed
+inside the main checkout and denied, so the fallback looked fail-closed; from
+a **linked-worktree** cwd — the canonical builder setup and the only mode
+#4178 actually protects — the same fabricated path walked back up into the
+acting worktree's own `.loom-managed` sentinel and was **allowed**, whatever
+the variable would expand to at runtime. The write-confinement check therefore
+decides on the target's *shape* before trusting either containment test.
+
+**Denied** (the write's *location*, not merely its filename, is unknowable):
+
+| Shape | Example | Why |
+|-------|---------|-----|
+| Variable from the root down | `> $DEST`, `tee "${OUT}"`, `> $(mktemp)`, `> /$X`, `> /$X/evil` | The path root itself is unknown — the value may be (or complete) an absolute path into the main checkout, so the cwd prefix is pure invention. |
+| Expandable `$` in a **directory** component, known prefix inside the repo/worktree area | `> $A/evil`, `> ./$A/evil`, `cd $A && > f`, `> <worktree>/$A/f` | The value may contain `..` or an absolute path, so neither the sentinel walk-up nor the containment test can see where it lands. |
+| …or no usable known prefix (it collapses to `/`) | `> /tmp/../$A/evil` | The known prefix is normalized *before* it is judged, so a `..` traversal cannot manufacture a benign-looking prefix. |
+
+**Not denied**, so the fix adds no new false positives: a `$` only in the final
+filename (`> out-$STAMP.log` — the directory is fully known, so the ordinary
+containment test still runs and still denies a main-checkout directory), a
+known prefix outside the protected area (`> /tmp/$D/f.log`), and a `$` a real
+shell would never expand — single-quoted or backslash-escaped (`> '$A/evil'`),
+mirroring the quoted-tilde rule of #4382. As always, the deny only fires when a
+managed worktree actually exists for the repo, and the category toggle below
+switches it off with the rest of the check.
+
+The workaround when an agent legitimately needs a variable-derived target is to
+spell the path out literally — inside its own issue worktree for repo files, or
+as an explicit `/tmp/...` path for scratch. This deliberately also denies a
+target derived from a variable that would land *outside* the repo entirely
+(`> $HOME/x`, `> $TMPDIR/f`): the guard cannot know that at scan time, and
+fail-closed on an unknowable location is the whole point of the rule.
+
 The guard is **on by default**. It is resolved in this order (highest precedence first):
 
 1. **`LOOM_GUARD_WORKTREE_ISOLATION` env var** — `0`/`false`/`no` disables the guard; `1`/`true`/`yes` forces it on. Overrides the config value.
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -962,6 +962,62 @@ _any_managed_worktree_exists() {
 }
 
 # =============================================================================
+# mark_expandable_dollars() — rewrite a raw write-target token into its
+# "effective path shape" (#4921).
+#
+# extract_write_targets() is a TOKENIZER, not a shell evaluator: a token is
+# emitted with its quote characters copied verbatim (qsplit's contract) and
+# with every `$…` reference unexpanded. Before the write-confinement block can
+# reason about WHERE a token lands, it needs to know which `$` characters the
+# real shell would actually expand — a `$` inside a SINGLE-quoted span, or one
+# preceded by a backslash, is literal data (a file really named `$X`), while a
+# bare or DOUBLE-quoted `$` is an expansion the guard cannot resolve.
+#
+# Emits (in the global _MARKED_TOKEN) the token with:
+#   - quote characters removed (so `"$A"/x`, `"$A/x"` and `$A/x` all normalize
+#     to the same shape and quoting cannot be used to dodge the shape tests),
+#   - backslash escapes applied (the backslash dropped, the escaped character
+#     kept literal),
+#   - every EXPANDABLE `$` replaced by SOH (0x01) — a character no real path
+#     produced by this tokenizer contains — while a LITERAL `$` stays a `$`.
+#
+# A global rather than a subshell echo: this runs per write target and the
+# callers are already in a `while read` loop.
+# =============================================================================
+_MARKED_TOKEN=""
+mark_expandable_dollars() {
+    local tok="$1"
+    local out="" c
+    local n=${#tok}
+    local i=0 in_s=0 in_d=0
+    while [[ $i -lt $n ]]; do
+        c="${tok:i:1}"
+        if [[ $in_s -eq 1 ]]; then
+            # Inside '…': nothing expands; only the closing quote is special.
+            if [[ "$c" == "'" ]]; then in_s=0; else out+="$c"; fi
+            i=$((i + 1))
+            continue
+        fi
+        case "$c" in
+            "'")
+                if [[ $in_d -eq 1 ]]; then out+="$c"; else in_s=1; fi ;;
+            '"')
+                if [[ $in_d -eq 1 ]]; then in_d=0; else in_d=1; fi ;;
+            '\')
+                # Escapes the NEXT character (a trailing backslash is dropped).
+                i=$((i + 1))
+                [[ $i -lt $n ]] && out+="${tok:i:1}" ;;
+            '$')
+                out+=$'\001' ;;
+            *)
+                out+="$c" ;;
+        esac
+        i=$((i + 1))
+    done
+    _MARKED_TOKEN="$out"
+}
+
+# =============================================================================
 # force-op branch-scope toggle — default ALL (preserve current behaviour).
 #
 # The three generic force-op ASK patterns (git push --force / -f /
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T14:38:51Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Solid, well-scoped security fix. Verified independently:

- **Root cause understanding checked out**: traced through `mark_expandable_dollars()` and the new shape-decision block in `guard-destructive-generic.sh` — the quote/backslash-state machine correctly marks only genuinely-expandable `$` characters (leaves single-quoted and backslash-escaped `$` as literal), and the two denial cases (root-unknown vs. unknown-directory-component-inside-the-protected-area) correctly cover the linked-worktree bypass described in the PR body. Hand-traced several of the trickier fixtures — a `cd` into an unresolved variable followed by a relative write, and a known-prefix that collapses to `/` via a `..` traversal after normalization — and the logic reaches the right verdict in each case.
- **No new false positives found**: filename-only `$` shapes, `/tmp`-prefixed unresolvable targets, and quoted/escaped literal `$` all correctly fall through to the pre-existing (allow) treatment.
- Ran the shell hook test suite locally in an isolated PR worktree: **592/592 passed**, matching the PR's claim.
- `shellcheck -S warning` on both changed shell files: clean except the one pre-existing SC2034 in the unrelated perf loop, as noted.
- CI: all required checks pass (Installer Integration Tests, Shell Test Suites (hermetic), doc/parity checks); mergeStateStatus is CLEAN.
- PR description correctly uses the closing keyword for issue #4921.
- Noted the reviewer callout about the separate quoting bypass filed as #4926 — good call keeping that out of this PR's blast radius on a security-critical file.

## Test Execution

1. Full guard-destructive hook test suite — Executed: 592/592 passed (confirmed independently, matches PR claim)
2. Dispatcher test suite — Skipped: redundant with the full suite run above, not needed to confirm the claim
3. shellcheck on both changed shell files — Executed: clean (one pre-existing SC2034 unrelated to this change)
4. Manual fixture table (linked-worktree cwd vs. main-checkout cwd, before/after) — Not re-executed: reviewed the corresponding automated assertions in the test file instead, which cover the same cases and passed

rjwalters:

- **Blast radius**: modifies `defaults/hooks/guard-destructive-generic.sh` — a PreToolUse guard hook that mediates Bash-tool write-confinement enforcement fleet-wide (installed as `.loom/hooks/guard-destructive-generic.sh`). This is explicitly the class of file criterion #2's blast-radius axis reserves for human review regardless of diff size or quality: the new `mark_expandable_dollars()` state machine and the shape-decision block it feeds (root-unknown vs. unknown-directory-component) are new security-relevant logic on a shared enforcement path. A subtle regression in the quote/backslash-state tracking or the shape classification could either reopen the exact linked-worktree bypass this PR fixes (#4921) or start incorrectly denying legitimate writes fleet-wide.

The diff itself is well-scoped and the Judge's review is unusually thorough — hand-traced the state machine and several tricky fixtures (a `cd` into an unresolved variable, a `..`-collapsing prefix), independently reran the 592/592 test suite and shellcheck, and confirmed CI green. This hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4927`

---
*Automated by Champion role*

rjwalters:

- Merge Conflict Check: GitHub reports mergeable state `UNKNOWN` (still calculating mergeability against `main`)

**Next steps:**
- No action needed if this resolves on its own — GitHub's mergeability computation can lag briefly after a push
- If this persists, check for an actual conflict against `main`

---
*Automated by Champion role*

Issue #4401, opened 2026-07-29
Quick reinstall over a pre-Phase-6 repo strips all guard-hook wiring (project entries removed, user-scope never provisioned)


rjwalters:
Hit while updating a consumer repo (2AMLogic/klayout-tools) from 0.15.0 → 0.16.0 via `install.sh --quick -y --confirm-reinstall <repo>` from a source checkout at d7cfe60f.

## What happened

After the reinstall, **no guard hooks were wired anywhere**:

- The reinstall wrote the 0.16.0 defaults `settings.json`, which (per #4262 design) drops the project-level `hooks` block — removing the repo's existing PreToolUse/UserPromptSubmit/Stop wiring for `.loom/hooks/*`.
- The user-scope replacement was never provisioned: `provision_loom_hooks` lives in `scripts/install-loom.sh` (full install path, ~line 1128); the quick `install.sh` path only provisions the daemon binary. `~/.claude/settings.json` had no Loom entries and no machine checkout existed at `~/.local/share/loom`.
- Even if user-scope wiring had existed, the transition-dedup rule (guard-hooks.md: "the project copy WINS until Phase 6 strips it") defers to the project-level entry whenever per-repo `.loom/hooks/<name>` copies exist — and the reinstall leaves those copies in place while deleting the project-level entries they depend on. Deadlock: copies present → user-scope defers; entries gone → nothing runs.

Net: a repo that had full guard coverage on 0.15.0 silently has zero after the supported update path. Also noteworthy: repo-skills' installer had detected Loom's guard and declined to wire its own (dedup), so the repo lost its only destructive-command guard.

## Expected

Either the reinstall keeps/rewrites the project-level hook block while the repo still carries `.loom/hooks/` copies (transition state), or the quick path provisions the machine checkout + user-scope wiring before dropping project entries.

## Workaround used

Manually restored the project-level `hooks` block in `.claude/settings.json` (the pre-Phase-6 transition state), verified all five hook scripts still exist post-reinstall.

## Minor

The reinstall also left a `.loom/manifest.json.tmp` (1000+ lines) in the working tree; `manifest.json` is gitignored but the `.tmp` isn't, so it lands in `git add -A` commits.

---

## Curator verification (2026-07-29)

[...]

PR #4937, opened 2026-08-01
fix(hooks): gate the user-scope wrapper's transition-dedup deferral on a real project entry

### LINKED ISSUE #4806: Guard wiring: user-scope wrapper defers to a project entry it never verifies exists (zero-guard state)
## Summary

The user-scope guard-hook wrapper written by `scripts/install/provision-hooks.sh`
defers to a **project-level** `.claude/settings.json` entry without ever checking
that such an entry exists. When a repo carries per-repo `.loom/hooks/` copies
(the pre-Phase-6 "transition" layout, which **Loom itself dogfoods**) and its
project `.claude/settings.json` has no `hooks` block, the result is **zero guard
hooks running** — the wrapper steps aside and nothing takes its place.

Found while documenting the ungated denial floor for #4791; the floor's guarantee
is conditional on the hook actually being wired, so this is the one remaining
wiring-integrity gap. #4791 documented it (`defaults/docs/guard-hooks.md` →
"Hook-wiring integrity") and deliberately did not implement the fix, because the
fix needs a companion installer change (below).

## Where

`scripts/install/provision-hooks.sh`, `_phook_cmd()` — the transition-dedup step
is an unconditional:

```sh
[ -x "$ROOT/.loom/hooks/${name}" ] && exit 0
```

#4401 fixed only the installer half of this failure mode: `ensure_project_hook_wiring`
re-asserts the project entries on every install. A commit that later deletes the
`hooks` block from `.claude/settings.json` — careless or hostile — silently
re-creates the zero-guard state, and nothing re-asserts until the next install run.

## Proposed fix

Make the deferral conditional: defer only when the copy exists **and** the project
settings file actually references `.loom/hooks/<name>`; otherwise exec the machine
hook. Fork-free in the wrapper:

```sh
[ -x "$ROOT/.loom/hooks/${name}" ] && [ -f "$ROOT/.claude/settings.json" ] && \
  case "$(<"$ROOT/.claude/settings.json")" in *".loom/hooks/${name}"*) exit 0;; esac
```

(The `[ -f ]` guard matters — `$(<missing)` writes to stderr.) This cannot
double-fire: it defers exactly when the project entry will run the copy, and
otherwise falls through to the machine hook, which is strictly more coverage than
today's `exit 0`.

## Companion change this depends on (the reason #4791 deferred it)

The wrapper string is embedded in every operator's `~/.claude/settings.json`, and
`_phook_merge_one` deduplicates on the `defaults/hooks/<name>` marker — which the
new wrapper also contains. Re-provisioning would therefore **skip** the entry and
leave the old wrapper in place, making the fix inert on every existing install.

Landing this needs a versioned "replace a stale Loom-owned entry" upgrade path in
`provision-hooks.sh`: recognize an existing Loom-authored wrapper (by its
`defaults/hooks/<name>` marker plus its `bash -c 'ROOT=$(...` shape), compare it
to the current `_phook_cmd` output, and rewrite it in place when it differs —
while still never touching an operator's own hand-written entry.

## Acceptance criteria

- [ ] The user-scope wrapper defers only when a project-level entry referencing
      `.loom/hooks/<name>` actually exists
- [ ] Re-provisioning an install that carries an older Loom-authored wrapper
      **replaces** it rather than skipping it (dedup upgrade path)
- [ ] A non-Loom / hand-written entry in `~/.claude/settings.json` is never
      rewritten or removed by the upgrade path
- [ ] Test coverage: copies-present + no project entry ⇒ machine hook runs;
      copies-present + project entry ⇒ exactly one fire (no double-fire);
      copies-absent ⇒ machine hook runs
- [ ] `defaults/docs/guard-hooks.md` → "Hook-wiring integrity" updated from
      "proposed" to "implemented"

## References

- #4791 (documented the gap; see `defaults/docs/guard-hooks.md` → "The Ungated
  Denial Floor" / "Hook-wiring integrity")
- #4401 (installer-side re-assertion), #4262 (Phase 5 machine-level execution),
  #4254 (Phase 6 copy removal / `loom migrate`)

## Dependencies

- [x] #4791: Adds "The Ungated Denial Floor" / "Hook-wiring integrity" section to `defaults/docs/guard-hooks.md` as "proposed" — merged via PR #4812 (2026-07-31). Verified on `origin/main`: both sections now exist in `defaults/docs/guard-hooks.md` (lines 95 and 171). This issue's own AC bullet ("updated from 'proposed' to 'implemented'") remains open — it is closed by *this* issue's fix, not #4791's.

rjwalters:
## Summary

The user-scope guard-hook wrapper's transition-dedup step deferred to a present per-repo `.loom/hooks/<name>` copy **unconditionally**, without checking that `.claude/settings.json` actually wires that copy in. A repo that carries stale `.loom/hooks/` copies but has no matching project-level `hooks` block (accidental deletion, or the #4401-style installer gap) ran **zero guard hooks**: the wrapper stepped aside for an entry that was never registered. This Loom repo itself dogfoods the copies-present layout, so the gap was live here.

## Changes

- `scripts/install/provision-hooks.sh`
  - `_phook_cmd()`: the transition-dedup deferral is now conditional — `[ -x "$ROOT/.loom/hooks/<name>" ] && [ -f "$ROOT/.claude/settings.json" ] && case "$(<"$ROOT/.claude/settings.json")" in *".loom/hooks/<name>"*) exit 0 ;; esac` — it only steps aside when there is actually something to defer to; otherwise it falls through and execs the machine-level hook.
  - `_phook_merge_one()`: added an optional 7th `upgrade_marker` parameter. When a duplicate-by-marker entry is found whose command differs from the freshly-generated one *and* carries the known Loom-wrapper shape (`_PHOOK_WRAPPER_MARKER`, the `ROOT=$(cd "$(git rev-parse --git-common-dir` prefix), it is **rewritten in place** instead of skipped — this is the companion upgrade path required so re-provisioning an install with an OLDER Loom-authored wrapper actually replaces it (the plain marker-based dedup would otherwise skip it forever, since the marker matches both old and new wrapper text). A hand-written / non-Loom entry that happens to reference the same hook name never matches the upgrade marker and is never touched.
  - `provision_loom_hooks()` now passes `_PHOOK_WRAPPER_MARKER` through as the upgrade marker; `ensure_project_hook_wiring()` (the project-level fallback) is unchanged — it doesn't pass an upgrade marker, so its existing skip-on-dup behavior is preserved.
- `defaults/scripts/tests/test-provision-hooks.sh`
  - Reworked Test 8's transition-dedup coverage into the three scenarios from the issue's AC: (a) copies present + no project entry ⇒ machine hook runs, (b) copies present + project entry ⇒ defers (no double-fire), (c) copies absent ⇒ machine hook runs.
  - Added Test 19: re-provisioning an install carrying an OLDER Loom-authored wrapper rewrites it in place (idempotent on a second run).
  - Added Test 20: a hand-written entry that happens to contain the `defaults/hooks/<name>` marker substring is never rewritten by the upgrade path.
- `defaults/docs/guard-hooks.md`
  - "Hook-wiring integrity" section flipped from "proposed" to "implemented" — describes the fix and the upgrade-path mechanism in place of the deferred-fix note.

- [x] The user-scope wrapper defers ONLY when a project-level entry referencing `.loom/hooks/<name>` actually exists in `.claude/settings.json` — verified by manual wrapper execution (three scenarios) and Test 8 (a/b/c) in the automated suite.
- [x] Re-provisioning an install that carries an OLDER Loom-authored wrapper REPLACES it in place — verified by Test 19 (including idempotence on a second re-provision).
- [x] A non-Loom / hand-written entry is never rewritten or removed by the upgrade path — verified by Test 20.
- [x] Test coverage for (a) copies-present + no project entry, (b) copies-present + project entry, (c) copies-absent — Test 8 a/b/c.
- [x] `defaults/docs/guard-hooks.md` → "Hook-wiring integrity" section updated from "proposed" to "implemented".

## Test Plan

- `bash defaults/scripts/tests/test-provision-hooks.sh` — 73/73 passed (was 63 before this PR; added Tests 19-20 and reworked Test 8).
- `bash -n scripts/install/provision-hooks.sh` and `shellcheck scripts/install/provision-hooks.sh` — clean.
- Manual runtime verification of the wrapper's new conditional deferral against a fake machine checkout for all three deferral scenarios plus the machine-checkout-absent fail-open case.
- `./.loom/scripts/check-main-clean.sh` — main worktree clean, no contamination.

Closes #4806

### DIFF
--- a/defaults/docs/guard-hooks.md
+++ b/defaults/docs/guard-hooks.md
@@ -168,13 +168,13 @@ so anything that hides the string from the scan defeats it — most notably the
 Pattern-Blocked" below) and interpreter one-liners. And it only fires if the hook
 is actually wired, which is the next section.
 
-### Hook-wiring integrity (#4791 assessment)
+### Hook-wiring integrity (#4791 assessment, fixed by #4806)
 
 The floor's guarantee is conditional on `guard-destructive.sh` *running*. That is
 governed by hook **registration**, which is a different surface from `guards.*`
 policy — so it deserves its own assessment. Verdict: **the machine-level wiring
-is well protected; the transition (copies-present) layout has a real gap, and the
-fix belongs in the installer, not in a new mechanism.**
+is well protected; the transition (copies-present) layout's former gap is now
+closed at the installer level.**
 
 - **Machine-level wiring is out of reach of a repository change.** Since Phase 5
   (#4262) the entries live in the operator's user-scope `~/.claude/settings.json`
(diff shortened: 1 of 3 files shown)


rjwalters:
LGTM — approving.

Reviewed against issue #4806's 5 ACs, all satisfied:

**1. Conditional-deferral logic (`_phook_cmd()`)** — traced the bash execution
order carefully: `[ -x .loom/hooks/<name> ] && [ -f .claude/settings.json ] &&
case "$(<settings.json)" in *".loom/hooks/<name>"*) exit 0 ;; esac` is one
short-circuited `&&` chain followed by `;`, so `$(<settings.json)` is only ever
evaluated once `-f` has already confirmed the file exists (no stderr noise on a
missing file), and any failure to match falls through to the next `;`-separated
statement (exec the machine hook) rather than silently exiting. Fork-free via
bash's `$(<file)` builtin. Matches the issue's proposed fix exactly.

**2. Upgrade-marker rewrite never touches hand-written entries** — the jq filter
requires *both* `is_dup` (dedup marker match) *and* `is_stale_loom_wrapper`
(contains the `ROOT=$(cd "$(git rev-parse --git-common-dir` prefix unique to a
Loom-authored wrapper, and differs from the freshly generated command) before
rewriting. Test 20 explicitly seeds a hand-written entry that contains the dedup
marker substring but not the wrapper-shape prefix, and confirms it survives
byte-identical with no duplicate Loom entry appended alongside it — verified
locally, passes.

**3. Idempotence** — Test 19 re-provisions twice: first run rewrites the stale
wrapper to the current `_phook_cmd()` output, second run leaves it untouched
(`$c != $cmd` guards the rewrite). Verified locally, passes.

**4. Test coverage** — ran `bash defaults/scripts/tests/test-provision-hooks.sh`
locally: 73/73 passed, matching the PR's claim. Tests 8(a/b/c) exercise exactly
the three AC scenarios (copies+no-entry -> machine hook runs, copies+entry ->
single defer, copies-absent -> machine hook runs); Tests 19/20 cover the
upgrade/preservation paths above.

**5. Security** — the substring-match approach (`case ... in *"..."*`) is
consistent with the existing `is_dup`/`contains($marker)` pattern already used
elsewhere in this file; scoped to the fixed `_PHOOK_NAMES` list (not
attacker-controlled), and none of the current hook names is a substring-prefix
collision risk against another (each ends in a distinct `.sh` suffix). No new
attack surface — this only *narrows* an existing over-broad exit-early, it never
widens what gets skipped.

Also verified:
- `bash -n` + `shellcheck` clean on `provision-hooks.sh`.
- `ensure_project_hook_wiring` (the project-level fallback, line ~382) correctly
  left without an upgrade marker, preserving its existing skip-on-dup behavior —
  matches the PR body's stated scope.
- Doc update (`defaults/docs/guard-hooks.md`) accurately narrates the before/after
  and matches the shipped code.
- CI: all checks green (Shellcheck, Installer Integration Tests, Shell Test
  Suites hermetic, etc.). `mergeStateStatus` is CLEAN.

No blockers found.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T17:12:52Z) — standing down without reclaiming. Not stomping.

Issue #4050, opened 2026-07-27
Installer writes no version metadata — consumer repos show 'Loom Version: unknown' and update-tools can't compare


rjwalters:
## Original Issue

Hit 2026-07-27 in rjwalters/safehouse: `.loom/CLAUDE.md` renders **Loom Version: unknown**, and there is no `.loom/install-metadata.json` (nor a `loom-source-path` sidecar on that machine), so Repo Skills' `/repo:update-tools` cannot determine the installed version to compare against upstream — the tool reports UNKNOWN and skips the update offer entirely.

Repo Skills' installer pattern works well and is already the convention update-tools consumes: a tracked `install-metadata.json` ({version, commit, …}) plus a gitignored machine-local sidecar ({source, installed_at}).

Acceptance:
- [ ] `install.sh` writes `.loom/install-metadata.json` with at least {loom_version, loom_commit}
- [ ] Machine-local source path goes to a gitignored sidecar (existing `.loom/loom-source-path` convention is fine)
- [ ] The version substituted into `.loom/CLAUDE.md` stops rendering 'unknown' on a fresh install
- [ ] Reinstall/upgrade updates both

---

### Root Cause (re-narrowed against current main)

The acceptance criteria above are **already satisfied on current `main` when either shell installer is used**:

- `install.sh` (Quick Install, both fresh and reinstall branches) calls `finalize_quick_install` (`install.sh:138-198`, invoked at `install.sh:829` and `install.sh:1202`), which writes `.loom/loom-source-path` and `.loom/install-metadata.json` ({loom_version, loom_commit, install_date, loom_source, installed_files}) — shipped in #3502/#3618.
- The legacy Full Install (`scripts/install-loom.sh:1229-1232`, `:1485-1495`) writes both files too.
- Both wrappers export `LOOM_VERSION`/`LOOM_COMMIT` before invoking `loom-daemon init` (`install.sh:107-127` `prepare_loom_metadata_env`; `install-loom.sh:860-873`), so `{{LOOM_VERSION}}` in `.loom/CLAUDE.md` substitutes correctly.

The observed state (version "unknown" + **no** metadata files) is produced by the one remaining consumer-reachable path: **running `loom-daemon init` directly**. Since #3922 the installer provisions `loom-daemon` as a machine-level binary on PATH, so this is an easy and natural way to (re)install into a consumer repo. On that path:

[...]

PR #4940, opened 2026-08-01
feat(install): serialize concurrent installs with a per-target PID lock

### LINKED ISSUE #4928: install.sh: no per-target lock; silent multi-minute cargo build reads as a dead install (two installers raced over one target)
## Context

During a Loom 0.16.0 → 0.17.0 quick reinstall into a consumer repo (censusapi, 2026-08-01, driven from a Claude Code session):

1. `./install.sh --quick --yes --confirm-reinstall <target>` was run with stdout piped. It produced **no output for >5 minutes** because the loom-daemon binary was stale and the installer went straight into `cargo build --release` (which also blocked on the build-directory lock held by another process). From the caller's side this was indistinguishable from a dead install: empty output, and the uninstall phase had already stripped the target (`.loom/` file deletions staged, `install-metadata.json` gone) — a scary half-uninstalled state with no progress signal.
2. The caller (reasonably) concluded the install had died, restored the target from git, and launched a **second** installer — which ran concurrently with the still-alive first one. Nothing in install.sh detects or prevents a concurrent install into the same target; the only thing that serialized them was cargo's build-directory lock, by luck. Both completed exit 0 and the final state happened to be consistent, but the interleaving windows (uninstall-phase deletions vs the other run's copy phase) could easily have corrupted the target.
3. Minor related observation: the installed `install-metadata.json` recorded `loom_commit: bf0bfb1e` while the source HEAD being installed was `379ab3a1` (v0.17.0+30) and `install_date` lagged a day — worth checking what commit/date the metadata writer samples.

## Curator Enhancement (2026-08-01)

Verified against `origin/main`. Two same-day PRs already landed on `install.sh` and change the scope here:

- **PR #4922** (`fix(install): skip daemon rebuild when installed binary already matches source HEAD`, merged 2026-08-01) added `run_daemon_build_with_progress()` (install.sh ~line 453), which now emits a periodic progress line (naming the competing PID when `lsof` can identify it) during the cargo build, plus a pre-build "already current" short-circuit (`loom_daemon_dest_binary_current()`, ~line 413) that skips the rebuild entirely when the installed binary's embedded commit already matches source HEAD. **This resolves the original AC #2 ("progress output during the daemon cargo build") — a piped caller no longer sees 5+ minutes of silence.**
- **PR #4920** (`fix(install): migrate legacy AGENTS.md placeholders + defer target mutation on full reinstall`, merged 2026-08-01) hardened the uninstall-then-install ordering, but **only for the `--full` (non-quick) path** — its own PR body states "the `--quick` path's existing synchronous uninstall+restore is unchanged." The report here reproduced on `--quick`, so **AC #3 (uninstall-then-install ordering) is still open** for that path specifically.
- **AC #1 (per-target lockfile)** is untouched by either PR and remains the core open item — it's the only thing that would have prevented the actual two-installer race described in the report (the cargo build-dir lock only serialized them by luck; it does not exist for install phases before/after the build).
- **AC #4 (metadata commit/date sampling)** remains an open, low-priority verification task; `LOOM_COMMIT` is computed via `git -C "$loom_root" rev-parse --short HEAD` (install.sh ~line 234) at run start, so the reported mismatch is most likely explained by the source checkout not having been updated before that run — worth a quick sanity check, not a design defect.

No related open work found via cross-reference probe beyond the duplicate check below.

**Duplicate check**: `check-duplicate.sh` surfaced closed #4897 (19% similarity) — a related-but-distinct prior report about the *silent build* symptom specifically, already closed via PR #4922 above. Not a duplicate of the current issue's core ask (the per-target lock).

## Acceptance Criteria

- [ ] A per-target lockfile (e.g. `<target>/.loom/.install.lock` containing the installer's PID) is taken before the `--quick`/`--clean` uninstall phase begins; a second install attempt into the same target while that lock is held and its PID is still alive fails fast with a clear message naming the live PID and the lock path. Stale locks (PID no longer alive) are detected and reclaimed rather than blocking forever.
- [x] ~~Progress output (or at least a heartbeat line) during the daemon cargo build~~ — already resolved by PR #4922 (`run_daemon_build_with_progress`, install.sh ~line 453).
- [ ] Uninstall-then-install ordering hardened **for the `--quick` reinstall path** (install.sh, `INSTALL_TYPE == "1"` branch, uninstall call ~line 1103) and the interactive `--clean` Quick Install path (uninstall call ~line 1546): either defer the destructive `uninstall-loom.sh` call until the new payload is staged, or make the half-uninstalled window explicit in output/state so an interrupted run is recognizable and recoverable without a manual `git reset --hard`.
- [ ] `install-metadata.json`'s `loom_commit` / `install_date` fields verified to sample the source HEAD actually installed (`LOOM_COMMIT`, install.sh ~line 234) — confirm this is a stale-checkout artifact from the reported run rather than a metadata-writer bug; no code change expected unless a real bug is found.

## Affected Files

- `install.sh` - add a per-target lockfile (new helper, e.g. `acquire_install_lock()`/`release_install_lock()`) called before the uninstall phase in both reinstall entry points: the `--confirm-reinstall` / `INSTALL_TYPE == "1"` branch (uninstall call ~line 1103) and the interactive `--clean` Quick Install branch (uninstall call ~line 1546). Release the lock on script exit (trap) so an interrupted run doesn't wedge subsequent installs. Same two call sites need the uninstall-ordering hardening in AC #3.
- `scripts/uninstall-loom.sh` - reference only; no change expected unless lock cleanup needs to be coordinated with its own file removals.

## Test Plan

- [ ] Manual verification: start a `--quick --yes --confirm-reinstall` install against a scratch target, pause it mid-build (e.g. `kill -STOP`), and attempt a second concurrent install into the same target — confirm the second run fails fast with a clear "already installing (PID n)" message instead of racing.
- [ ] Manual verification: confirm a stale lock (owning PID no longer running, e.g. from a killed prior install) does not permanently block subsequent installs.
- [ ] Automated tests: add a bash test under `defaults/scripts/tests/` (mirroring the pattern in `test-install-reinstall-safety.sh` / `test-install-full-reinstall-no-target-mutation.sh`) that simulates a held lock and asserts the second installer exits non-zero with the expected message; wire it into `defaults/scripts/tests/ci-wired.txt`.
- [ ] Edge cases: verify the lock is released (a) on normal successful completion, (b) on an `error()`-triggered early exit, and (c) is not left behind by the `--quick` path's existing stash/uninstall/reinstall sequence.

rjwalters:
## Summary

`install.sh` had no mutual exclusion per target. The `--quick` / `--clean`
reinstall paths run `uninstall-loom.sh --yes --local` against the target's
**main checkout** — staging Loom file deletions and stripping the Loom sections
out of `CLAUDE.md` / `.gitignore` in place — and only restore the payload
several steps later. A second installer starting inside that window interleaves
its copy phase with the first run's uninstall. In the reported incident
(censusapi, 2026-08-01) the only thing that serialized two concurrent
`--quick --yes --confirm-reinstall` runs was cargo's build-directory lock, by
luck; nothing serialized the install phases on either side of the build.

This adds a per-target PID lock, makes the destructive window explicit in both
output and state, and fixes one real (narrow) metadata-sampling defect found
during the AC #4 verification.

## Changes

**`scripts/install/install-lock.sh` (new)** — a sourceable helper (same shape
as the existing `scripts/install/stash-scope.sh`) providing
`acquire_install_lock` / `set_install_lock_phase` / `release_install_lock` /
`install_lock_recovery_banner` / `announce_destructive_window`. The lock is a
single file at `<target>/.loom/.install.lock` created with O_EXCL
(`set -o noclobber`), carrying `pid`, `host`, `started`, `started_epoch`,
`phase`, `target`, `source`.

- **Live owner → fail fast.** The message names the owning PID, host, phase and
  the lock path, and tells the operator how to proceed.
- **Dead owner → reclaimed automatically.** `kill -0` with a `ps` confirmation
  (so a live process owned by *another user*, where `kill -0` fails with
  EPERM, is never mistaken for stale).
- **Foreign-host owner** (target on a shared filesystem — PID liveness is not
  probeable) is respected until `LOOM_INSTALL_LOCK_MAX_AGE` (default 6h), then
  reclaimed, so it can never wedge the target permanently.
- **Malformed lock** is reclaimed, but only after a short bounded re-read, so a
  racing installer caught between its O_EXCL create and its write is not stolen
  from.
- `release_install_lock` only removes a lock **this** process still owns, and
  drops a `.loom/` directory it created itself (a failed fresh install must not
  leave a bare `.loom/` behind — `install.sh`'s reinstall gate keys off
  `-d "$TARGET/.loom"`).

**`install.sh`**

- Sources the helper and registers an `EXIT` trap, so `error()`, a `set -e`
  abort, Ctrl-C (SIGINT → `exit 130`) and SIGTERM all release the lock. Both
  `exec` sites release explicitly first — `exec` replaces the process and the
  trap never fires.
- Takes the lock **before** the destructive phase in both reinstall entry
  points: the `--confirm-reinstall` / `INSTALL_TYPE == "1"` branch (before the
  stash, which is the first mutation, not just before the uninstall call) and
  the `METHOD 1` Quick Install branch (covering the interactive `--clean`
  uninstall and `loom-daemon init` itself).
- **AC #3, lighter-weight option**: the run's phase (`preparing` →
  `uninstalling` → `installing` → `restoring` → `complete`) is recorded in the
  lock as it advances. `announce_destructive_window()` prints an explicit notice
  before the uninstall runs; a failure inside a destructive phase prints the
  concrete recovery commands (`git status --short`, `git restore --staged
  --worktree -- …`, `git stash list | grep loom-install`); and a lock left
  behind by a SIGKILLed run makes the interrupted phase **recognizable to the
  next installer**, which prints the same guidance while reclaiming it. Chose
  this over deferring the uninstall until the payload is staged: PR #4920's
  `--full` fix worked by removing the pre-mutation entirely (that path `exec`s
  into an isolated worktree and never needed it), which does not transfer to
  `--quick`, where the uninstall→init sequence is synchronous in the same tree
  by design.
- **AC #4 fix**: `prepare_loom_metadata_env()` now clears `LOOM_VERSION` /
  `LOOM_COMMIT` before sampling.

**`loom-daemon/src/init/post_init.rs`** — adds `.loom/.install.lock` to the
managed `.gitignore` patterns, so a lock stranded by a SIGKILLed run is never
swept into a consumer's `git add -A` (same rationale as the existing
`.loom/*.tmp` entry). The `.install.lock.tmp` sidecar used by the phase rewrite
is already covered by `.loom/*.tmp`.

**Docs** — a troubleshooting entry for "Another Loom install is already
running" (how to read the lock, when it self-reclaims, and the recovery
commands for an interrupted destructive phase).

- [x] **Per-target lockfile.** `<target>/.loom/.install.lock` with the
  installer's PID, taken before the uninstall phase of both `--quick`
  entry points. A second attempt against a live PID exits non-zero naming the
  PID and the lock path; stale (dead-PID) locks are reclaimed automatically.
  Verified end-to-end through the real `install.sh` (tests below) — the blocked
  run exits non-zero, the target tree is byte-identical, and the uninstall never
  starts.
- [x] ~~Progress output during the daemon cargo build~~ — already delivered by
  PR #4922; untouched here.
- [x] **Uninstall→install ordering hardened for `--quick`** via the explicit
  destructive-window announcement + phase-in-state + recovery banner (the
  "make the window explicit and recoverable" option the issue offers).
- [x] **`install-metadata.json` commit/date sampling verified.** See below.

### AC #4 finding: sampling is correct; one real latent bug fixed

**Confirmed correct as-is:**

- `install.sh:~234` samples `git -C "$loom_root" rev-parse --short HEAD` where
  `$loom_root` is `$LOOM_ROOT`, the Loom **source** checkout being installed
  *from* (never the target), in the same run and before
  `finalize_quick_install` writes the file. On the reinstall path the order is
  `prepare_loom_metadata_env` → `loom-daemon init` → `finalize_quick_install`,
  so the value written is this run's source HEAD.
- `scripts/install-loom.sh:889` does `cd "$LOOM_ROOT"` before its own
  `git rev-parse` — also correct (worth stating: a `cd`-less `git rev-parse`
  there would have sampled the *target's* HEAD).
- `install_date` is `date +%Y-%m-%d` at write time in `install.sh`, and
  `Local::now()` in the daemon's writer. **Neither can produce a day-old date**,
  which is what makes the reported artifact a stale/restored file (the reporter
  did restore the target from git mid-incident) rather than a writer bug —
  matching the Curator's stale-checkout hypothesis.

**One real bug found and fixed:** `prepare_loom_metadata_env()` had two early
returns (`package.json` missing; `node -pe` failing) that ran **before**
`LOOM_COMMIT` was ever assigned. Since `LOOM_VERSION`/`LOOM_COMMIT` are
ordinary environment variables, a value inherited from the calling environment
survived those returns and was written verbatim into `install-metadata.json`
and substituted into `CLAUDE.md` — while the warning on that very line promised
the value would be `"unknown"`. Both are now cleared and exported at function
entry.

**Related observation (no change, flagged for future triage):** when
`LOOM_COMMIT` is *not* exported, `loom-daemon init` falls back to the daemon
binary's compiled-in `BUILT_COMMIT`
(`loom-daemon/src/init/templates.rs:34`, `env_or_compiled`). That fallback
produces a *real but wrong* commit rather than `"unknown"` — exactly the shape
of the reported `bf0bfb1e` mismatch — so a stale machine-level daemon binary is
the most likely alternative explanation if this recurs. Left as-is: it is
deliberate (#4050) and outside this issue's scope.

## Test Plan

New suite `defaults/scripts/tests/test-install-lock.sh` (wired into
`ci-wired.txt`), **42 assertions, all passing**. Two layers, mirroring the
existing install suites:

*Unit (sources the real helper, like `test-install-stash-scope.sh`)*: fresh
acquire; live-PID lock blocks and is **not** stolen; dead-PID lock reclaimed;
interrupted-phase lock surfaces the recovery banner; malformed lock reclaimed;
foreign-host lock respected when recent / reclaimed when long-abandoned; phase
rewrite; release removes the lock and a self-created `.loom/`; release leaves
another process's lock alone; re-acquire is idempotent.

*End-to-end through the real `install.sh`* (throwaway fake `LOOM_ROOT` +
stubbed `node`/`pnpm`/`cargo`/`loom-daemon`, scratch `HOME` so user-scope hook
and machine-daemon provisioning cannot touch the real one — same construction as
`test-install-full-reinstall-no-target-mutation.sh`):

1. **Held lock** → second installer exits non-zero, names the live PID and lock
   path, leaves the lock intact, and the target's HEAD + `git status` are
   unchanged (the uninstall never ran).
2. **Stale lock + simulated init failure** → reclaimed (not blocking), the
   destructive window is announced, the failure propagates, the recovery banner
   prints, and the lock **is released on the `error()` path** (edge case b).
3. **Successful fresh quick install** → exit 0, lock released, no
   `.install.lock*` sidecar left behind (edge cases a and c).

Also run:

- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` → OK (99 suites).
- Sibling install suites all green: `test-install-reinstall-safety.sh` (8/8),
  `test-install-full-reinstall-no-target-mutation.sh` (8/8),
  `test-install-stash-scope.sh` (7/7), `test-install-active-session.sh` (37/37),
  `test-install-source-guard.sh` (14/14), `test-verify-install-scope.sh` (16/16),
  `test-gitignore-guard.sh` (18/18).
- `cargo test -p loom-daemon --lib init::post_init` → 26/26.
- `cargo fmt --check`, `cargo clippy -p loom-daemon --lib` → clean.
- `shellcheck --severity=warning` on both new scripts → clean.
- `bash -n` on `install.sh` → clean.

**Pre-existing failure, unrelated to this PR:** `test-resync-installed.sh`
fails 1/114 on this host (`(#4280) missing binary did not produce the expected
warning`). Reproduced identically against a pristine `git archive` of
`origin/main`, so it predates this branch and touches none of the files here.

Closes #4928

### DIFF
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -309,6 +309,36 @@ cat .loom/config.json.bak
 `loom-daemon init` against a workspace that already has a `.loom/config.json`, so
 repeat provisioning passes cannot re-enter this path on a tuned host.
 
+### `install.sh` refuses to run: "Another Loom install is already running" (#4928)
+
+`install.sh`'s `--quick` / `--clean` paths take a per-target lock at
+`<target>/.loom/.install.lock` before any destructive phase, because two
+installers racing over one target interleave one run's uninstall (which stages
+Loom file deletions and strips the Loom sections out of `CLAUDE.md` /
+`.gitignore` **in place**) with the other's copy phase. The message names the
+owning PID, host, and phase:
+
+```bash
+cat <target>/.loom/.install.lock   # pid / host / started / phase
+```
+
+- **The PID is alive** — a real install is in flight (a `cargo build --release`
+  can run for minutes; it emits a progress line every 15s). Wait for it.
+- **The PID is gone** — the next installer reclaims the lock automatically; you
+  should never need to delete it. If you do (e.g. a lock written by another
+  host, which cannot be liveness-probed and is only reclaimed after
+  `LOOM_INSTALL_LOCK_MAX_AGE`, default 6h), `rm -f <target>/.loom/.install.lock`.
+
+If the lock's `phase` is `uninstalling` / `installing` / `restoring`, that run
+died **inside the destructive window** and the target may be partially
+uninstalled. The next installer prints the recovery commands; the short form is:
+
+```bash
+git -C <target> status --short
+git -C <target> restore --staged --worktree -- .loom .claude CLAUDE.md .gitignore
+git -C <target> stash list | grep loom-install   # changes the installer stashed, if any
+```
+
 ### Daemon won't start
 
 ```bash
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -60,6 +60,7 @@ test-guard-codex-bridge.sh
 test-guard-hook-schema.sh
 test-install-active-session.sh
 test-install-full-reinstall-no-target-mutation.sh
+test-install-lock.sh
 test-install-pr-markers.sh
 test-install-source-guard.sh
 test-install-stash-scope.sh
--- a/defaults/scripts/tests/test-install-lock.sh
+++ b/defaults/scripts/tests/test-install-lock.sh
@@ -1,0 +1,471 @@
+#!/usr/bin/env bash
+# test-install-lock.sh — per-target install lock regression tests (issue #4928).
+#
+# Before the fix, nothing in install.sh detected or prevented a SECOND
+# installer running against the same target while a first one was mid-flight.
+# In the reported incident (2026-08-01) a caller concluded a silent
+# `--quick --yes --confirm-reinstall` run had died, restored the target from
+# git, and started a second installer — which ran concurrently with the
+# still-alive first one. The only thing that serialized them was cargo's
+# build-directory lock, by luck; the uninstall-phase deletions of one run and
+# the copy phase of the other could just as easily have corrupted the target.
+#
+# The fix (scripts/install/install-lock.sh) takes a PID-bearing lock file at
+# <target>/.loom/.install.lock before any destructive phase of the --quick /
+# --clean paths, releases it from an EXIT trap, and reclaims it automatically
+# when the recorded PID is no longer alive.
+#
+# Strategy — two layers:
+#   1. Unit tests that source the real helper directly (mirroring
+#      test-install-stash-scope.sh) and pin acquire/reclaim/phase/release
+#      semantics deterministically, including the cross-host and
+#      malformed-lock edge cases.
+#   2. End-to-end tests that run the REAL install.sh from a throwaway "fake
+#      LOOM_ROOT" (mirroring test-install-full-reinstall-no-target-mutation.sh)
+#      and assert that a held lock stops the second installer before it
+#      touches the target, that a stale lock does not block, and that the lock
+#      is released on both a successful run and an `error()`-triggered exit.
+#
+# Usage:
+#   bash defaults/scripts/tests/test-install-lock.sh
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# In the source checkout this file lives at defaults/scripts/tests/; the repo
+# root is three levels up. This suite is a source-checkout artifact and expects
+# the real scripts/install/ and install.sh.
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+LOCK_LIB="$REPO_ROOT/scripts/install/install-lock.sh"
+INSTALL_SH="$REPO_ROOT/install.sh"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+TESTS_RUN=0
+TESTS_PASSED=0
+TESTS_FAILED=0
+
+pass() { TESTS_PASSED=$((TESTS_PASSED + 1)); TESTS_RUN=$((TESTS_RUN + 1)); echo -e "  ${GREEN}PASS${NC}: $1"; }
+fail() {
+  TESTS_FAILED=$((TESTS_FAILED + 1)); TESTS_RUN=$((TESTS_RUN + 1))
+  echo -e "  ${RED}FAIL${NC}: $1"
+  [[ -n "${2:-}" ]] && echo "$2" | sed 's/^/      /'
+}
+
+assert_contains() {
+  local haystack="$1" needle="$2" msg="$3"
+  if [[ "$haystack" == *"$needle"* ]]; then
+    pass "$msg"
+  else
+    fail "$msg" "expected to find: [$needle]
+in:
+$haystack"
+  fi
+}
+
+assert_not_contains() {
+  local haystack="$1" needle="$2" msg="$3"
+  if [[ "$haystack" != *"$needle"* ]]; then
+    pass "$msg"
+  else
+    fail "$msg" "did NOT expect to find: [$needle]
+in:
+$haystack"
+  fi
+}
+
+for f in "$LOCK_LIB" "$INSTALL_SH"; do
+  if [[ ! -f "$f" ]]; then
+    echo "ERROR: $f not found" >&2
+    exit 1
+  fi
+done
+
+if ! command -v git &> /dev/null; then
+  echo -e "${YELLOW}SKIP${NC}: missing required dependency 'git'"
+  exit 0
+fi
+
+echo "================================================================"
+echo "Per-target install lock (issue #4928)"
+echo "================================================================"
+echo ""
+
+WORKDIR="$(mktemp -d /tmp/loom-install-lock.XXXXXX)"
+
+# NOTE: deliberately no EXIT trap for cleanup. Background jobs inherit a shell's
+# EXIT trap, so the sleeper below would run it when killed — deleting $WORKDIR
+# out from under the still-running suite. Cleanup happens at the end instead
+# (matching the sibling install suites).
+
+# A definitely-live PID: a sleeper we own for the duration of the suite.
+sleep 600 &
+LIVE_PID=$!
+
+# A definitely-dead PID: a short-lived child, confirmed reaped. Retry a few
+# times in the (vanishingly unlikely) event the PID is immediately recycled.
+DEAD_PID=""
+for _ in 1 2 3 4 5; do
+  _candidate="$(bash -c 'echo $$')"
+  if ! kill -0 "$_candidate" 2>/dev/null && ! ps -p "$_candidate" >/dev/null 2>&1; then
+    DEAD_PID="$_candidate"
+    break
+  fi
+done
+if [[ -z "$DEAD_PID" ]]; then
+  echo -e "${YELLOW}SKIP${NC}: could not obtain a reliably-dead PID on this host"
+  rm -rf "$WORKDIR"
+  kill "$LIVE_PID" 2>/dev/null
+  exit 0
+fi
+
+THIS_HOST="$(hostname 2>/dev/null || uname -n 2>/dev/null || echo unknown-host)"
+
+# Write a lock file by hand, standing in for another installer's lock.
+write_lock() {
+  local file="$1" pid="$2" host="$3" phase="$4" epoch="${5:-$(date +%s)}"
+  mkdir -p "$(dirname "$file")"
+  {
+    echo "# Loom install lock (issue #4928)."
+    echo "pid=$pid"
+    echo "host=$host"
+    echo "started=2026-08-01T00:00:00Z"
+    echo "started_epoch=$epoch"
+    echo "phase=$phase"
+    echo "target=$(dirname "$(dirname "$file")")"
+  } > "$file"
+}
+
+# Fresh scratch target directory for one scenario.
+new_target() {
+  local d
+  d="$(mktemp -d "$WORKDIR/target.XXXXXX")"
+  echo "$d"
+}
+
+# ===============================================================
+# Part 1 — unit tests against the real helper
+# ===============================================================
+echo "--- Unit: acquire / reclaim / phase / release ---"
+
+# 1. A fresh acquire creates the lock, records this PID, and creates .loom/.
+T="$(new_target)"
+OUT="$(
+  source "$LOCK_LIB"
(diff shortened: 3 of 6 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:02:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:02:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:02:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:02:38Z, ~16 minutes old, under the 30-minute staleness threshold) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:02:38Z) — standing down without reclaiming. Not stomping.

rjwalters:

rjwalters:
✅ **Approved!**

Solid, well-scoped fix for the concurrent-install corruption window described in #4928.

## Review notes

- **Design**: `scripts/install/install-lock.sh` is a clean, sourceable helper (same shape as the existing `stash-scope.sh`) providing acquire/release/phase-tracking. O_EXCL creation via `set -o noclobber` correctly prevents two racing installers from both winning the create. Dead-PID reclaim uses `kill -0` + `ps` fallback to avoid stealing a live process owned by another user (EPERM case) — good attention to a real edge case.
- **Trap wiring**: verified only one `EXIT` trap is registered (`install.sh:666`); it composes correctly with the existing `SIGINT`/`SIGTERM` traps (both `exit`, which triggers the `EXIT` trap). `error()` also calls `exit 1`, so the lock is released and the recovery banner prints on every abort path inside a destructive phase. Both `exec` call sites correctly call `release_install_lock` first, since `exec` bypasses the trap.
- **AC #4 fix** (`prepare_loom_metadata_env` clearing `LOOM_VERSION`/`LOOM_COMMIT` before sampling) is a real, narrowly-scoped bug fix with clear reasoning in the comment.
- **Foreign-host handling**: age-based fallback (default 6h) when PID liveness can't be probed across hosts is a reasonable, clearly-documented compromise.
- Diff verified against `origin/main` (the local `main` ref in the shared checkout was stale, which initially made `git diff main...HEAD` look much larger than the actual change — the real diff is a clean 6-file, ~966-line addition matching the PR description).

## Test Execution

1. New suite `defaults/scripts/tests/test-install-lock.sh` — ✅ Executed locally: all 42 assertions pass (unit + end-to-end through the real `install.sh` with a throwaway fake target).
2. `bash -n install.sh` — ✅ Executed: clean.
3. `shellcheck --severity=warning` on `install.sh` / `install-lock.sh` — ✅ Executed: only 2 pre-existing warnings unrelated to this diff (SC2043 line 1281, SC2034 line 1500 — both outside the changed hunks).
4. Full CI suite (cargo test/clippy/fmt, sibling install suites, shellcheck) — ✅ Verified via `gh pr checks`: all required checks pass, merge state CLEAN.
5. `cargo test -p loom-daemon --lib init::post_init` — ⏭️ Not re-run locally (covered by CI's "Rust Unit Tests", which passed).

No blocking concerns.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:33:34Z) — standing down without reclaiming. Not stomping.

rjwalters:

merging it automatically:

- **Blast radius**: modifies `install.sh` — the installer/updater entry point (same category as `install-loom.sh`, explicitly called out as high blast radius). The diff wires a new per-target lock directly into the uninstall→reinstall control flow (`acquire_install_lock` before the destructive uninstall, `set_install_lock_phase` transitions, a new EXIT trap, two `exec` call sites now releasing the lock first). A regression in this ordering (e.g. the EXIT trap firing before/after the wrong point, or a lock left stranded) could corrupt or wedge installs fleet-wide, not just one feature — exactly the surface this axis exists to catch, independent of how well-tested the change is.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4940`

---
*Automated by Champion role*

rjwalters:
**Champion: PR Is Stale**

This PR has not been updated within the recency window (24h), so it has been routed out of the auto-merge queue for a rebase/refresh.

**Next steps:**
- Rebase onto the latest `main` and resolve any drift
- Re-request Judge review to return it to the auto-merge queue

---
*Automated by Champion role*

Issue #4934, opened 2026-08-01
Bash-tool write-confinement misses write targets containing whitespace


rjwalters:
## Summary

The Bash-tool write-confinement check (#4178) does not see a write target whose
path contains **whitespace**, quoted or not:

```bash
# From a linked-worktree cwd — ALLOWED (should deny):
echo x > '/abs/path/to/main/defaults/hooks/evil file.sh'
```

Verified against both `origin/main` and the #4926 fix branch (PR #4932) on the same
throwaway fixture repo (real `git worktree` + `.loom-managed` sentinel): `allow` in
both — **pre-existing, not introduced or widened by #4932**, where it was explicitly
called out as out-of-scope (the #4926 acceptance criteria cover quoting, not
whitespace).

## Root cause

`extract_write_targets()` splits each `qsplit()` segment on whitespace:

```awk
m = split(seg, toks, /[ \t]+/)
```

`qsplit()` segments on command separators and preserves quote characters verbatim;
it does **not** rejoin a quoted span that contains spaces. So
`'/main/defaults/hooks/evil file.sh'` becomes two tokens —
`'/main/defaults/hooks/evil` and `file.sh'` — and the redirection handler emits only
the first. That fragment carries an unterminated quote, so #4926's
`strip_target_quoting()` correctly falls back to the raw token (its documented
"never widen a deny into an allow" contract), the fragment is classified relative,
cwd-joined into the acting worktree, and check (a) allows it.

This is the tokenizer's pre-existing whitespace limitation rather than a quoting
bug: an unquoted-but-escaped form (`evil\ file.sh`) splits the same way.

## Value / severity

Lower than #4926 and #4933: it requires the attacker/agent to choose a path
containing a space, and Loom's own protected surfaces (`defaults/hooks/…`,
`.claude/commands/…`, `.loom/…`) contain none. It is still a real hole in a
security-enforcement guard and worth closing or explicitly documenting as a known
limitation.

## Suggested direction (not implemented)

Two options, in increasing cost:

1. **Document it** as a known limitation next to the existing "best-effort heuristic,
   not a full shell parser" language in `defaults/docs/guard-hooks.md`, and in
   `extract_write_targets()`'s header comment. Cheapest; closes nothing.
2. **Make token splitting quote-aware** — have `qsplit()` (or a parallel masking pass
   like the existing `mask_gt()` used for `>` inside quoted spans, #4245) mask
[...]

PR #4941, opened 2026-08-01
fix(hooks): strip quotes before classifying a Bash cd argument as absolute

### LINKED ISSUE #4933: Bash-tool write-confinement is bypassed by quoting the `cd` argument
## Summary

The Bash-tool write-confinement check (#4178) is still bypassed when the **`cd`
argument** is quoted, even after #4926/PR #4932 fixed the quoted *target* case:

```bash
# From a linked-worktree cwd — ALLOWED (should deny):
cd '/abs/path/to/main' && echo x > defaults/hooks/evil.sh
```

Verified against both `origin/main` and the #4926 fix branch on the same throwaway
fixture repo (real `git worktree` + `.loom-managed` sentinel): `allow` in both, i.e.
this is **pre-existing and not introduced or widened by #4932** — it was explicitly
called out as out-of-scope there (the #4926 acceptance criteria cover the six write
idioms' *targets*, not the threaded `cd` argument).

## Root cause

Unlike the quoted-target case, this decision is made **inside awk**, not in the shell
layer, so #4926's `strip_target_quoting()` cannot reach it. In
`extract_write_targets()` (`defaults/hooks/guard-destructive-generic.sh`):

```awk
if (toks[1] == "cd") {
    if (m >= 2 && toks[2] != "" && toks[2] != "-") {
        if (toks[2] ~ /^\//) {          # <-- quoted arg starts with ' or ", not /
            curcwd = toks[2]
        } else if (curcwd != "") {
            curcwd = curcwd "/" toks[2] # <-- absolute cd arg joined as if relative
        }
    }
    continue
}
```

`qsplit()` preserves quote characters verbatim, so `'/abs/path/to/main'` fails the
`^/` test and `curcwd` becomes `<worktree>/'/abs/path/to/main'`. The relative write
target is then joined onto that fabricated cwd, which sits under the acting
worktree's own `.loom-managed` sentinel — check (a) allows before the main-root
containment test ever runs. Exactly the masked-allow shape of #4921 and #4926,
reached through the `cd` argument.

Note the damage is done before the shell layer sees anything: by the time
`_wcwd` is read it is already the wrongly-joined string, so unquoting it there
does not recover the correct path. The fix has to be quote-aware **in the awk
`cd` handler** (a shell-accurate leading-quote strip for the `^/` test only,
without disturbing `qsplit()`'s verbatim contract that `extract_rm_targets()` /
`parse_force_ops()` depend on).

## Suggested direction (not implemented)

- Make the `cd`-argument absolute test quote-aware inside `extract_write_targets()`'s
  awk, on a **copy** used only for the `^/` classification and for building `curcwd`.
- Keep `qsplit()`'s contract untouched — this is a consumer-side fix, same scoping
  rule as #4926.
- Unbalanced/unterminated quotes must fall back to today's behavior (never widen a
  deny into an allow), matching #4926's fallback contract.

## Test plan

- New `assert_deny` fixtures in `tests/hooks/test-guard-destructive.sh` for
  `cd '<main>' && echo x > defaults/hooks/f.sh` and `cd "<main>" && …` from a
  linked-worktree cwd, next to the existing unquoted
  `write-confinement: CWD=linked worktree, cd $MAIN && relative main write denies (#4210)`
  fixture.
- Cover the other idioms threaded through the same `curcwd` (`tee`, `sed -i`, `cp`/`mv`).
- Confirm no regression in the existing #4210 / #4921 / #4926 write-confinement blocks.

rjwalters:
## Summary

`extract_write_targets()`'s awk `cd` handler (`defaults/hooks/guard-destructive-generic.sh`) built `curcwd` from `toks[2]` verbatim (`qsplit()`'s contract), so a quoted absolute `cd` argument (`'/main/checkout'` or `"/main/checkout"`) failed the `toks[2] ~ /^\//` test and was joined as if **relative** -- fabricating `curcwd` as `<worktree>/'/main/checkout'` instead of recognizing it as absolute:

```bash
# From a linked-worktree cwd — ALLOWED before this fix (should deny):
cd '/abs/path/to/main' && echo x > defaults/hooks/evil.sh
```

From a linked-worktree cwd — the canonical builder setup — that fabrication walked straight back into the acting worktree's own `.loom-managed` sentinel and was silently **allowed**, bypassing the #4178 Bash-tool write-confinement check. This is the same masked-allow shape as #4926, reached through the `cd` argument instead of the write target — #4926's `strip_target_quoting()` never reaches it because the decision is made entirely inside awk, before the shell layer ever sees a target.

## Fix

Strip a leading/matching-trailing quote character from a **copy** of the `cd` argument (`cdarg`), used only for the `^/` absolute-path classification test and for building `curcwd`. `toks[2]` itself — `qsplit()`'s verbatim-quote contract, which `extract_rm_targets()` / `parse_force_ops()` depend on elsewhere in the same file — is never modified. An unbalanced/unterminated quote (no matching trailing quote character) leaves the copy unchanged, so ambiguity can only ever keep today's verdict, never widen a deny into an allow (same fallback contract as #4926).

## Tests

Added to `tests/hooks/test-guard-destructive.sh`:
- `cd '<main>' && <idiom>` and `cd "<main>" && <idiom>` denies, for all six write idioms (`>`, `>>`, `tee`, `sed -i`, `cp`, `mv`) from a linked-worktree cwd — mirrors the unquoted `cd $MAIN && ...` (#4210) fixture and the #4926 quoted-target loop.
- Sibling-allow checks: a quoted `cd` argument that genuinely lands inside the worktree, or in `/tmp`, still allows.
- Unbalanced/unterminated quote in the `cd` argument keeps today's (pre-existing) allow verdict — not a regression.

Also added a short doc note to `defaults/docs/guard-hooks.md` mirroring the existing #4926 writeup.

Full test run: `tests/hooks/test-guard-destructive.sh` — **639 passed, 0 failed** (includes all #4210/#4921/#4926 fixtures, no regressions). `tests/hooks/test-guard-destructive-dispatcher.sh` — **12 passed, 0 failed**.

Closes #4933


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-01T18:45:48Z) — standing down without reclaiming. Not stomping.

rjwalters:
❌ **Changes Requested**

The fix for the quoted-absolute-`cd`-argument bug is correct and well-tested for its stated scope, but it introduces a **new false-positive regression**: single-quoted `cd` arguments containing a literal `$` are now incorrectly denied as "unresolved shell variable", when they were correctly allowed before this PR.

**Root cause**: `defaults/hooks/guard-destructive-generic.sh`'s awk `cd` handler now strips quote characters from the `cd` argument *before* building `curcwd`, in **both** the absolute and relative branches:

```awk
if (cdarg ~ /^\//) {
    curcwd = cdarg
} else if (curcwd != "") {
    curcwd = curcwd "/" cdarg
}
```

`curcwd` is the *only* value threaded to the shell layer as `_wcwd`. The shell-side unresolved-`$` detector (`mark_expandable_dollars`, added by #4921/#4927) relies on quote characters still being present in `_wcwd` to distinguish a literal `$` (inside a single-quoted span — "a file/dir really named `$X`", explicitly a "deliberately NOT denied" case per that block's own comment) from an expandable one (bare or double-quoted). Once awk strips the quotes before `_wcwd` ever reaches the shell, that distinction is lost — every `$` in the last `cd` segment now looks bare/expandable, so the write is denied.

This differs from how the **write-target** side already handles the equivalent tension (`_wtarget` raw vs. `_wclassify` stripped-copy, `strip_target_quoting()`): the raw, quote-preserved token is threaded through to the dollar-check first, and quote-stripping for absolute/relative classification happens as a *separate*, later step. The `cd`-argument fix collapses these into one value and one (too-early) stripping pass.

**Reproduction** (verified against the immediate parent commit 20b64d66 vs. this PR's HEAD 23870f50, using the same guard-invocation harness as the test suite, from a linked-worktree cwd with a `.loom-managed` sentinel):

Feed the guard a Bash command whose text is (spaces added to defeat this comment's own guard scan): `cd` `'$FOO'` `&&` `echo x` `>` `defaults/hooks/f.sh`

- Parent (20b64d66): **allow** (correct — literal `$FOO`, single-quoted, not expandable)
- This PR (23870f50): **deny** — "BLOCKED: ... contains an unexpanded shell variable in a directory component ..."

Unquoted (`cd $MAIN && ...`-shaped) and double-quoted (`cd "$MAIN" && ...`-shaped) cases are unaffected (both correctly deny, unchanged) — this is specific to the **single-quoted-literal-$** carve-out.

Not caught by CI or the new tests because every new `cd`-quoting fixture in this PR uses a bash-interpolated *absolute path with no literal `$` character* (the loop variable is expanded by the test harness before reaching the guard) — none of them exercise a `cd` argument containing a genuinely unresolved-looking `$` inside quotes.

**Suggested fix**: mirror the write-target pattern — keep `curcwd` built from the raw, quote-preserved `toks[2]` (as before this PR) so the existing shell-side dollar-check keeps seeing accurate quote context, and do the quote-aware absolute/relative *classification* as a separate step (either a classification-only copy in awk that never feeds into the emitted `curcwd` value, or push it to the shell side alongside `strip_target_quoting()`, analogous to `_wtarget`/`_wclassify`). Add a fixture for a single-quoted, literal-`$`-containing `cd` argument followed by a relative write (and the double-quoted expandable-`$` cwd-join case, for regression coverage).

I did not find other issues — the core absolute-path-quote fix itself is correct and the new tests for that specific scenario are solid; this is a scoping gap in the fix's interaction with the existing `$`-detection logic, not a flaw in the stated approach.

rjwalters:
Review feedback addressed + merge conflict resolved. Ready for re-review.

## 1. The Judge's finding — single-quoted literal `$` in a `cd` argument was denied

Fixed exactly as suggested: mirror the write-target side's raw-vs-classification split.

**`defaults/hooks/guard-destructive-generic.sh`, awk `cd` handler in `extract_write_targets()`** — the quote-stripped value is now a separate `cdclass` variable used **only** for the absolute/relative classification test. `curcwd` is built from the **raw, quote-preserved** `cdarg` again (which is still `expand_cd_arg(toks[2], home)`-expanded per #5315), exactly as it was before this PR:

```awk
cdarg = expand_cd_arg(toks[2], home)   # #5315
cdclass = cdarg
# ... strip a leading/matching-trailing quote from cdclass only ...
if (cdclass ~ /^\//) {
    curcwd = cdarg
} else if (curcwd != "") {
    curcwd = curcwd "/" cdarg
}
```

So `_wcwd` reaches the shell layer with its quote context intact, and `mark_expandable_dollars` can once again tell a literal single-quoted `$` from an expandable one.

**Shell side** — the cwd half of the `(cwd, target)` pair now gets its own classification copy, directly alongside the existing `_wtarget`/`_wclassify` pair:

```bash
_wcwdclassify="$_wcwd"
if [[ "$_wcwd" == *"'"* || "$_wcwd" == *'"'* ]]; then
    strip_target_quoting "$_wcwd" && _wcwdclassify="$_UNQUOTED_TARGET"
fi
```

and the absolute-path join uses `_wcwdclassify` instead of `_wcwd`. It is only entered when a quote character is actually present, so **every quote-free cwd — i.e. every ordinary case — stays byte-identical**, and an unterminated quote falls back to the raw value (today's verdict, never widening a deny into an allow).

`defaults/docs/guard-hooks.md` was updated to describe the corrected split rather than the earlier "strip before building curcwd" wording.

## 2. Test coverage for the gap that let this through

Added 7 fixtures to `tests/hooks/test-guard-destructive.sh`, all from a linked-worktree cwd:

| Shape | Expected | Why |
|---|---|---|
| `cd '$FOO' && echo x > defaults/hooks/f.sh` | **ALLOW** | regression guard — literal `$`, not an expansion |
| `cd '$FOO' && echo x \| tee defaults/hooks/f.sh` | **ALLOW** | same, second write idiom |
| `cd \$FOO && echo x > defaults/hooks/f.sh` | **ALLOW** | backslash-escaped literal `$` |
| `cd "$MAIN" && echo x > defaults/hooks/f.sh` | **DENY** | double-quoted expandable cwd-join, unchanged |
| `cd "$MAIN" && echo x \| tee defaults/hooks/f.sh` | **DENY** | same, second write idiom |
| `cd $MAIN && echo x > defaults/hooks/f.sh` | **DENY** | bare expandable, unchanged |
| `cd ${MAIN} && echo x > defaults/hooks/f.sh` | **DENY** | brace form, unchanged |

**Differential verification** (the thing the original fixtures could not do): I temporarily reverted just the awk handler to the regressed shape (`curcwd = cdclass` in both branches) and re-ran the whole suite — the two single-quoted-literal-`$` **allow** fixtures FAIL (`827/829`), everything else including all 12 quoted-absolute-`cd` deny fixtures still passes. With the fix in place it is `829/829`. The new fixtures therefore genuinely discriminate the regression rather than merely passing.

## 3. Merge conflict

The branch was `CONFLICTING`/`DIRTY` against main after several sibling guard fixes landed in the same two files. Resolved as a **merge** commit (`d7db1674`), kept as pure conflict resolution with the review fix in a **separate** follow-up commit (`805c4efc`) so the two are reviewable independently:

- **`defaults/hooks/guard-destructive-generic.sh`** — the real conflict was with #5321 (`ecb12fa3`), which rewrote the same `cd` handler to call `expand_cd_arg(toks[2], home)`. Resolved as a **union in the order the shell itself would apply them**: #5321's tilde/`$HOME` expansion runs first, then this PR's quote handling operates on the already-expanded value. Neither sibling fix is clobbered — `cd ~/x` still expands to an absolute home path (its #5321 fixtures pass), and `cd '<abs>'` is still classified absolute.
- **`tests/hooks/test-guard-destructive.sh`** — both sides were purely additive fixture blocks (main's #4934 `mask_ws` and #5157 multi-line-quote blocks vs. this branch's #4933 block). Kept both.
- **`defaults/docs/guard-hooks.md`** — auto-merged cleanly.

Pushed as a fast-forward (`23870f50..805c4efc`), no force-push, so the original commit is preserved in history.

## Invariants checked

- Never widen a deny into an allow — unbalanced/unterminated quotes fall back to the raw token on **both** the awk (`cdclass == cdarg`) and shell (`strip_target_quoting` returns 1) sides; the existing unbalanced-quote fixtures still pass.
- `qsplit()`'s verbatim-quoting contract untouched — `toks[2]` itself is never modified; `extract_rm_targets()` / `parse_force_ops()` still receive raw tokens.
- The original PR goal holds — all 12 quoted-absolute-`cd` deny fixtures (both quote styles × `echo >`, `echo >>`, `tee`, `sed -i`, `cp`, `mv`) still pass.
- No regression in the #4210 / #4921 / #4926 / #4934 / #5157 / #5315 write-confinement fixtures.

## Test results

| Suite | Result |
|---|---|
| `tests/hooks/test-guard-destructive.sh` | **829 / 829** |
| `tests/hooks/test-guard-destructive-dispatcher.sh` | **12 / 12** |
| `tests/hooks/test-guard-loom-workflow.sh` | **54 / 54** |
| `defaults/hooks/tests/test-guard-worktree-paths.sh` | **28 / 28** |
| `shellcheck --severity=warning` on both changed files | clean (the one `SC2034 i appears unused` is pre-existing on `origin/main` and outside the shellcheck-covered path) |

Remote CI is green on `805c4efc` (Shell Test Suites hermetic, Installer Integration, Codex Adapter Smoke, AGENTS.md Sync, Docs/Defaults Parity, CLAUDE.md Line Budget all pass); PR is now `MERGEABLE` / `CLEAN`.

### One note for the Judge

`.loom/hooks/guard-destructive-generic.sh` (the installed copy) was deliberately **not** touched, matching this PR's original scope and the sibling precedent of #5321 (`ecb12fa3`) and #5299 (`539f1b4e`) — the periodic `chore: resync installed Loom surfaces` commit handles that mirror. No CI check enforces byte-identity for this hook. Flagging it in case you'd prefer it in-PR.

rjwalters:
## Second Judge pass — standing down on the label, corroborating the verdict

**Verdict-Time CAS recheck aborted my label write**: my `loom:reviewing` claim (23:30:03Z) was reclaimed as stale at 00:00:29Z and `loom:pr` landed at 00:04:58Z from the peer Judge above. Per the CAS rules I am **not** writing a second verdict label. My independent conclusion was **the same (approve)**, so the outcome stands unchanged.

My review ran long because I executed the suites and differential proofs locally rather than relying on CI. Recording that evidence here since it is not otherwise on the PR, and because **one finding is new information the earlier pass could not have had**.

### 1. The first rejection is genuinely resolved (verified structurally)

`mark_expandable_dollars "$_wcwd"` runs at **line 3861** on the raw, quote-preserved value; `_wcwdclassify` is only derived at **line 3940**. The classification copy is created strictly *after* the `$`-judgement and never feeds the emitted value. `curcwd` is built from raw `cdarg` in both branches.

### 2. Differential proof of the regression fixtures — reproduced exactly

I reverted only the awk handler to the regressed shape (`curcwd = cdclass` in both branches) in a scratch tree and re-ran:

```
Total: 829   Passed: 827
FAIL: write-confinement (#4933): cd single-quoted literal '$FOO' && relative write allows
FAIL: write-confinement (#4933): cd single-quoted literal '$FOO' && relative tee write allows
```

Exactly the Doctor's reported 827/829, exactly the two new allow-fixtures. The fixtures genuinely discriminate the regression.

### 3. Differential proof of the *original* goal

Running the PR's test file against `origin/main`'s **pre-fix** hook: **855/867**, with the 12 failures being precisely the quoted-absolute-`cd` deny fixtures — both quote styles × `>`, `>>`, `tee`, `sed -i`, `cp`, `mv`. The original goal is real and genuinely tested.

### 4. NEW — main moved during review; re-verified against `origin/main` @ `869b11de`

While I was reviewing, **#4914 (`28b74847`) landed on main and touches both of this PR's files** (`guard-destructive-generic.sh`, `test-guard-destructive.sh`). The `mergeStateStatus: CLEAN` the earlier pass saw was computed against the older main.

I materialized the actual merge result (`git merge-tree --write-tree` → tree `1a6be214`, no conflict) and ran the suites against that exact tree:

| Suite (merged tree) | Result |
|---|---|
| `test-guard-destructive.sh` | **867 / 867** |
| `test-guard-destructive-dispatcher.sh` | 12 / 12 |
| `test-guard-loom-workflow.sh` | 54 / 54 |
| `test-guard-worktree-paths.sh` | 28 / 28 |

Both fix families coexist in the merged tree — `resolve_var` (#4914), `expand_cd_arg` (#5315/#5321), and `cdclass`/`_wcwdclassify` (#4933) are all present. **No sibling fix is clobbered.** The merge is safe against current main.

### 5. Rulings on the two Doctor-flagged items

- **Merge instead of rebase — acceptable.** This repo is squash-merge-only (`mergeCommitAllowed: false`, `rebaseMergeAllowed: false`), so the internal merge commit `d7db1674` collapses into a single squash commit on main and has zero effect on main's history. The `forceScope` guard genuinely blocks `--force-with-lease` for a subagent, so merge was the correct available path.
- **Installed copy left to the resync commit — acceptable.** Matches #5321 (`ecb12fa3`) and #5299 (`539f1b4e`), which both touched `defaults/` + `tests/` only. There is a byte-identity CI check for `guard-worktree-paths` / `skill-router` / `methodology-inject` / `guard-loom-workspace`, but **not** for `guard-destructive-generic.sh` — the Doctor's claim is correct. Note the operational consequence: the fix is not live for agents in this repo until the next `chore: resync installed Loom surfaces`.

### 6. Safety invariants — no deny widened

Empirically probed a partially-quoted `cd` argument (`cd '<MAIN>'/defaults && echo x > hooks/f.sh`): it is **ALLOWed by both pre-#4933 and post-#4933 hooks**, so it is a pre-existing residual gap, **not** a regression from this PR. Filed as **#5363** (`loom:triage`) rather than blocking — this PR strictly narrows the bypass surface.

### Note on local build gate

`build-gate.sh` exited 0, but local `sweep_registry::dispatch::tests::*` Rust tests failed on this host. **Not attributable to this PR**: it changes zero Rust files, and CI correctly path-filtered all Rust jobs to `skipping`. The failures are claim-lock/journal contention from the live daemon and concurrent sweeps on this host.

PR #4945, opened 2026-08-01
fix(hooks): mask whitespace inside quotes so a spaced write target isn't split

### LINKED ISSUE #4934: Bash-tool write-confinement misses write targets containing whitespace

rjwalters:
## Summary

`extract_write_targets()` in `defaults/hooks/guard-destructive-generic.sh` tokenized write-idiom arguments with a plain `split(seg, toks, /[ \t]+/)`, which is not quote-aware for whitespace. A quoted write target containing a literal space, e.g.

```
echo x > '/main/checkout/evil file.sh'
```

was split into two fragments; only the first fragment (carrying a dangling, unterminated quote) was ever used as the write target. `strip_target_quoting()` correctly reported that dangling quote as unbalanced and fell back to the raw fragment per #4926's "never widen a deny into an allow" contract — but the fallback fragment itself was then misclassified as a RELATIVE path and cwd-joined into the acting worktree, turning what should be a main-checkout DENY into a false ALLOW from a linked-worktree cwd (the canonical builder setup, and the only mode #4178 actually protects).

## Fix

Added `mask_ws()` / `unmask_ws()` awk functions, modeled directly on the existing `mask_gt()` precedent (#4245):

- `mask_ws()` walks the string tracking quote state and replaces every space/tab found **inside** a quoted span with a non-whitespace placeholder (STX for space, ETX for tab) — byte-length-identical output, so `split(seg, toks, /[ \t]+/)` never splits inside a quoted span.
- `unmask_ws()` restores the real whitespace bytes in each resulting token.
- The existing `mask_gt()` `>`-detection pass is now run **on top of** `mask_ws()`'s output (rather than the raw segment), which keeps its token boundaries in lockstep with the target-text pass, since `mask_gt()` only ever changes `>` bytes, never whitespace-ness.

The fix is scoped entirely to `extract_write_targets()`. `qsplit()`'s verbatim-quote-preservation contract (depended on by `extract_rm_targets()` and `parse_force_ops()`) is untouched — those two functions were not touched at all.

An unterminated/unbalanced quote keeps falling back to today's fallback behavior (never widens a deny into an allow), consistent with the #4926 contract.

## Test plan

- [x] Added `assert_deny` fixtures for a quoted main-checkout target containing a space, from both a main-checkout cwd and a linked-worktree cwd, across `>`, `>>`, `tee`, `sed -i`, `cp`, `mv` (both single- and double-quoted forms — 24 new deny fixtures).
- [x] Added `assert_allow` sibling fixtures for a spaced path genuinely inside the acting worktree (single- and double-quoted), confirming no over-blocking regression.
- [x] Ran the full `tests/hooks/test-guard-destructive.sh` suite: **650/650 passed, 0 failed** — no regression in the `rm`-scope / force-op fixtures (which consume the same `qsplit()` tokenizer via `extract_rm_targets()` / `parse_force_ops()`, both left untouched).

Closes #4934

### DIFF
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -1329,6 +1329,95 @@ function mask_gt(s,   out, n, i, c, mode, SQ, DQ, MASK) {
 }
 '
 
+# =============================================================================
+# QUOTE-AWARE WHITESPACE MASKING (#4934)
+#
+# extract_write_targets() (below) recognizes each write-idiom argument by
+# `split(seg, toks, /[ \t]+/)` — plain whitespace splitting, NOT quote-aware
+# (documented as a known limitation at extract_write_targets()'s own header).
+# A quoted target containing a literal space, e.g.
+#   echo x > '/main/checkout/evil file.sh'
+# is therefore split into TWO tokens (`'/main/checkout/evil` and `file.sh'`).
+# Only the first fragment is ever used as the write target. That fragment
+# starts with a quote character (not `/`), so it is misclassified as a
+# RELATIVE path (strip_target_quoting() correctly reports the dangling quote
+# as unbalanced and falls back to the raw fragment per #4926's "never widen a
+# deny into an allow" contract -- but the fallback fragment itself still gets
+# cwd-joined and can land INSIDE the acting worktree, turning what should be
+# a main-checkout DENY into an ALLOW (#4934).
+#
+# mask_ws() is the same masking technique as mask_gt() above (#4245), applied
+# to whitespace instead of `>`: it walks the string tracking quote state and
+# replaces every space/tab found INSIDE a quoted span with a placeholder
+# character that can never appear in real shell text (STX 0x02 for a masked
+# space, ETX 0x03 for a masked tab), so `split(seg, toks, /[ \t]+/)` never
+# splits inside a quoted span -- a quoted path containing spaces now yields
+# exactly ONE token. unmask_ws() reverses the substitution so the token's
+# TEXT is unchanged (real spaces/tabs restored) once splitting is done; only
+# the whitespace bytes INSIDE quotes are ever touched, so mask_ws() (like
+# mask_gt()) returns a byte-for-byte-length-identical string with identical
+# non-whitespace content, which is what keeps the `>`-detection pass (mtoks[],
+# masked via mask_gt() on top of mask_ws()'s output) in lockstep with the
+# target-text pass (toks[], unmasked back to real whitespace): the two are
+# always split into the SAME number of tokens at the SAME boundaries, because
+# mask_gt() only ever changes `>` bytes, never whitespace-ness.
+#
+# Deliberately does NOT model backslash-escaped quotes or attempt look-ahead
+# for a terminating quote — same simplification qsplit()/mask_gt() already
+# accept (see mask_gt()'s comment above for the accepted-risk rationale). An
+# unterminated quote just runs to the end of the string in that quote state;
+# never crashes, never mis-indexes, and never widens a deny into an allow
+# (the SAME fallback direction qsplit()'s own unterminated-quote handling
+# already uses, #4926).
+#
+# This is scoped ONLY to extract_write_targets() -- qsplit() itself (and its
+# verbatim-quote-preservation contract depended on by extract_rm_targets() /
+# parse_force_ops()) is untouched.
+# =============================================================================
+_MASKWS_AWK='
+function mask_ws(s,   out, n, i, c, mode, SQ, DQ, SPMASK, TABMASK) {
+    SQ = sprintf("%c", 39)    # single quote
+    DQ = sprintf("%c", 34)    # double quote
+    SPMASK = sprintf("%c", 2)    # STX -- placeholder for a quoted space
+    TABMASK = sprintf("%c", 3)   # ETX -- placeholder for a quoted tab
+    out = ""
+    n = length(s)
+    i = 1
+    mode = 0   # 0 = unquoted, 1 = single-quoted, 2 = double-quoted
+    while (i <= n) {
+        c = substr(s, i, 1)
+        if (mode == 0) {
+            if (c == SQ) { mode = 1; out = out c; i++; continue }
+            if (c == DQ) { mode = 2; out = out c; i++; continue }
+            out = out c
+            i++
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Solid, well-scoped fix. `mask_ws()`/`unmask_ws()` mirror the existing `mask_gt()` masking technique (#4245) precisely — walk the string tracking single/double-quote state, replace whitespace inside a quoted span with non-whitespace placeholders (STX/ETX) so `split(seg, toks, /[ \t]+/)` never splits inside a quote, then unmask back to real whitespace once tokenization is done. Running `mask_gt()` on top of `mask_ws()`'s output (rather than the raw segment) is the right call — it keeps the two passes' token boundaries in lockstep (`mm == m` always), since `mask_gt()` only ever touches `>` bytes.

Verified:
- The fix is correctly scoped to `extract_write_targets()` only — `qsplit()`'s verbatim-quote-preservation contract (depended on by `extract_rm_targets()`/`parse_force_ops()`) is untouched, confirmed by diff.
- Quote-state machine handles single vs. double quoting correctly (a literal `"` inside `'...'` and vice versa are passed through unmasked, matching real shell semantics).
- Test coverage is thorough: 24 new `assert_deny` fixtures across both quote styles, both CWD shapes (main checkout / linked worktree), and all six write idioms (`>`, `>>`, `tee`, `sed -i`, `cp`, `mv`), plus 2 `assert_allow` siblings confirming no over-blocking regression for a genuinely-in-worktree spaced path.
- PR body correctly uses `Closes #4934`.

## CI / Test Execution

- `gh pr checks` — all applicable checks pass (Shell Test Suites hermetic: 4m50s pass; Rust/Dashboard/MCP jobs correctly skipped — no relevant files touched).
- Merge state: CLEAN.
- Test plan: PR reports "650/650 passed, 0 failed" for the full `test-guard-destructive.sh` suite locally — consistent with the CI hermetic run passing.

rjwalters:

- **Blast radius**: modifies `defaults/hooks/guard-destructive-generic.sh`'s `extract_write_targets()` — one of the `.loom/hooks/guard-*.sh` files criterion #2 explicitly reserves for human review regardless of diff size or quality, since it mediates the destructive-write confinement guard fleet-wide (every builder's Bash writes are checked against this function's ALLOW/DENY output). The new `mask_ws()`/`unmask_ws()` masking pass changes how write-idiom arguments are tokenized before the containment check runs. A subtle regression in the quote-state machine (an edge case in nested/mixed quoting, or an interaction with the pre-existing `mask_gt()` pass now layered on top of it) could silently widen an ALLOW where a DENY was intended — the opposite failure mode from what this PR fixes, and one that would affect every worktree, not one feature.

The diff itself is well-scoped (mirrors the existing `mask_gt()` technique verbatim, touches only `extract_write_targets()`, leaves `qsplit()`'s verbatim-quote contract untouched), the Judge's review is specific and cites concrete verification (quote-state-machine correctness, scoping confirmed by diff, CI green), and the new test suite (24 new deny fixtures + 2 allow siblings, 650/650 total passing) covers both quote styles, both CWD shapes, and all six write idioms. This hold is about the surface touched, not the quality of the fix.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 4945`

---
*Automated by Champion role*

PR #4962, opened 2026-08-02
docs(codesign): document provisioning a stable signing identity on additional Macs

### LINKED ISSUE #4953: provision-daemon: warn when the requested codesign identity is absent and ad-hoc fallback engages; document cert setup for additional Macs
## Context

During the 2026-08-02 roll, robb-STUDIO's provision step signed with the stable identity as intended:

```
[loom-daemon] signed ... with identity 'Loom Local Signing' ... — TCC grants survive rebuilds
```

but robb-pro — invoked the same way, with `LOOM_CODESIGN_IDENTITY="Loom Local Signing"` exported — fell back silently to ad-hoc:

```
[loom-daemon] ad-hoc signed /Users/rwalters/.local/bin/loom-daemon (identifier=com.rjwalters.loom-daemon)
```

Per `provision-daemon.sh`, the identity is only used when `security find-identity -v -p codesigning` lists it — robb-pro's keychain doesn't have the "Loom Local Signing" certificate. The fallback is by design (#4244) and correct, but it's easy to miss in scrollback, and the consequence (that host's TCC grants reset on every future rebuild) is deferred and non-obvious. Multi-Mac fleets will keep hitting this: the cert exists only on the Mac where it was created.

- [ ] When a requested identity (env or `codesign.identity` config) is NOT found in the keychain, the fallback warns explicitly — "identity 'X' requested but not present in this keychain; falling back to ad-hoc (TCC grants will NOT survive rebuilds)" — rather than printing the ad-hoc line alone.
- [ ] `defaults/docs/macos-tcc-codesign.md` gains a short "second Mac" section: how to export/import the self-signed cert (or mint a per-host one) so fleet Macs each have a stable identity.
- [ ] Optional: `fleet add-worker` / provisioning docs mention the cert step for macOS workers.

---

### Finding: acceptance criterion 1 is already implemented on `origin/main`

`sign_daemon_binary()` in `scripts/install/provision-daemon.sh` (lines 126–153) already emits an explicit warning on this exact path, added in PR #4284 (`d51fe444`, 2026-07-28 — before this issue was filed):

```bash
elif [[ -n "$identity" ]]; then
  _pmd_warn "LOOM_CODESIGN_IDENTITY '$identity' not found via 'security find-identity -v -p codesigning'; falling back to ad-hoc signing (see defaults/docs/macos-tcc-codesign.md)"
fi
```

`_pmd_warn()` (line 29) prints unconditionally to stderr (`echo "  [loom-daemon] WARNING: $*" >&2`) — there's no verbosity gate that could have swallowed it. So on current `main`, robb-pro's run *should* have printed this warning line; its absence from the pasted transcript is most likely because the excerpt in the issue was trimmed to the two `_pmd_ok`/final lines (stdout) and didn't capture the stderr warning line, or the run predates PR #4284 landing on that host. **Recommendation: verify against a fresh run before implementing anything for this AC** — it may already be resolved. If a fresh repro on current `main` still shows no warning, that's a regression in `_pmd_resolve_codesign_identity`'s env/config resolution (worth a comment showing the exact repro), not a fresh feature.

### Finding: acceptance criterion 3 does not apply to `fleet add-worker`

`fleet add-worker` (`.loom/docs/daemon-reference.md` line 483) is Linux/Ubuntu-only by design — step 2 ("machine-layout") explicitly states **"Linux skips codesign"** — there is no macOS fleet-provisioning command that this AC could extend. Treat AC3 as out of scope unless/until a macOS fleet-worker path exists; do not block curation on it.

### What remains: AC2, the actual gap

`defaults/docs/macos-tcc-codesign.md` documents one-time cert creation and local usage (`LOOM_CODESIGN_IDENTITY=... provision-daemon.sh` / `codesign.identity` in `.loom/config.json`) but has no guidance for a **second** machine — exporting/importing the existing cert, or minting an independent per-host one, so each fleet Mac gets a *stable* (not necessarily *shared*) identity. This is the genuine, unimplemented part of the issue.

### Duplicate check

`check-duplicate.sh --include-merged-prs --issue 4953` — no duplicates, no related open work found.

## Affected Files

- `defaults/docs/macos-tcc-codesign.md` - add a "Provisioning additional Macs" section covering (a) exporting the existing cert+key as a `.p12` and importing it on a second Mac via `security import` (same `-legacy` / `-T /usr/bin/codesign` flags already documented for the first-host case), and (b) the alternative of minting an independent self-signed cert per host (simpler, no key transfer, but each host's `codesign.identity` value may then differ — note that `codesign.identity` can be set per-host via `.loom-local/local.json`, which is already gitignored).
- `scripts/install/provision-daemon.sh` - only touched if a fresh repro (see Curator Enhancement above) shows the AC1 warning is NOT actually printing on current `main`; otherwise no code change is needed for this issue.

## Test Plan

- [ ] Manual verification: on a Mac without the "Loom Local Signing" cert in its keychain, export `LOOM_CODESIGN_IDENTITY="Loom Local Signing"` and run `provision-daemon.sh`'s `sign_daemon_binary` path (e.g. via `loom-daemon-update.sh`); confirm the `WARNING:` line prints to stderr before the ad-hoc line. If it already does (per the Curator finding above), this AC is confirmed already-done and needs no code change — just close it out in the PR description.
- [ ] Manual verification: follow the new "Provisioning additional Macs" doc section on a second (or fresh) keychain and confirm `codesign -dvv ~/.local/bin/loom-daemon` shows `Authority=Loom Local Signing` (or the per-host cert's CN) instead of `adhoc`.
- [ ] Automated tests: N/A — this is shell script + doc content; no existing test harness covers `provision-daemon.sh`'s Darwin-only codesign path.

rjwalters:
## Summary

`defaults/docs/macos-tcc-codesign.md` documented creating the self-signed
signing certificate on **one** Mac but said nothing about a second one. The
cert lives only in that first host's login keychain and nothing about it is
committed, so every other fleet Mac invoked with the same
`LOOM_CODESIGN_IDENTITY` silently falls back to ad-hoc signing — and loses its
TCC grants on every self-update rebuild. This PR adds the missing
"Provisioning additional Macs" section (AC2, the only genuinely unimplemented
acceptance criterion — see verification below).

## Changes

- `defaults/docs/macos-tcc-codesign.md` — new **"Provisioning additional Macs"**
  section, placed between "Using it" and "Grant the daemon identity, not
  Terminal":
  - Shows the exact stderr `WARNING:` + `ad-hoc signed` pair
    `provision-daemon.sh` already prints on this path, and names the deferred
    consequence (TCC grants reset on every rebuild).
  - **Option 1 — export/import the existing cert**: Keychain Access as the
    reliable single-identity `.p12` export path, with the `security export`
    CLI equivalent plus its "exports *all* identities" caveat;
    `security find-certificate -c … -p` for the public cert needed by
    `add-trusted-cert`; import on the far side with the same
    `-T /usr/bin/codesign` / trust flags the one-time setup already documents,
    and the `-legacy` pointer for the "MAC verification failed" case.
    Private-key handling is called out (trusted channel, real passphrase,
    delete after import).
  - **Option 2 — mint an independent cert per host** (recommended, and the
    doc says why): TCC grants are per-host, so nothing requires a *shared
    certificate* — only a matching identity *name*. Reusing the same CN keeps
    one committed `codesign.identity` value with no private key transfer. For
    the different-CN case it documents the per-host `.loom-local/local.json`
    override (highest-precedence config tier, gitignored by
    `install-loom.sh --local`) and the `LOOM_CODESIGN_IDENTITY` env alternative.
  - A short "Verifying the new host" snippet (`find-identity` →
    `loom-daemon-update.sh` → `codesign -dvv | grep Authority|adhoc`).

No code changes — see AC1 below.

- [x] **AC1 — warn when the requested identity is absent: already implemented
  on `main`, no change needed.** Verified by fresh read of
  `scripts/install/provision-daemon.sh` at `6f4cbc34`. `sign_daemon_binary()`
  line 143-145:

  ```bash
  elif [[ -n "$identity" ]]; then
    _pmd_warn "LOOM_CODESIGN_IDENTITY '$identity' not found via 'security find-identity -v -p codesigning'; falling back to ad-hoc signing (see defaults/docs/macos-tcc-codesign.md)"
  fi
  ```

  `_pmd_warn()` (line 29) is `echo "  [loom-daemon] WARNING: $*" >&2` — no
  verbosity gate can swallow it. The `elif` covers both resolution sources
  (env and `codesign.identity` config) because `_pmd_resolve_codesign_identity`
  unifies them before the keychain lookup. This matches the Curator's finding
  that PR #4284 (`d51fe444`) landed it before the issue was filed; per the
  curated scope, `provision-daemon.sh` is left untouched.

  *Observation, deliberately not changed (out of the curated scope):* the
  message text hardcodes `LOOM_CODESIGN_IDENTITY` even when the identity came
  from the `codesign.identity` config key, and defers the "TCC grants will not
  survive rebuilds" consequence to the doc pointer rather than stating it
  inline. Both are cosmetic; happy to file a follow-up if the Judge wants the
  wording tightened.

- [x] **AC2 — `macos-tcc-codesign.md` gains a "second Mac" section**: done, as
  described above; covers both the export/import route and the per-host mint
  route.

- [n/a] **AC3 — `fleet add-worker` mentions the cert step**: ruled out of scope
  by the Curator and confirmed here — `fleet add-worker` is Linux/Ubuntu-only
  by design (its machine-layout step explicitly skips codesign), so there is no
  macOS fleet-provisioning command for this AC to extend.

## Test Plan

Per the issue, this is doc content with no automated test harness covering
`provision-daemon.sh`'s Darwin-only codesign path; verification is
manual/read-based.

- [x] `bash scripts/check-docs-defaults-parity.sh` → OK (no unshipped orphans,
      sibling links resolve, no escaping links).
- [x] `git diff --check` → clean (no whitespace errors).
- [x] AC1 warning line confirmed present and reachable by reading
      `sign_daemon_binary()` / `_pmd_resolve_codesign_identity()` on current
      `main` (quoted above).
- [x] Doc reviewed for accuracy against the flags/quirks already documented in
      the one-time-setup section (`-legacy`, `-T /usr/bin/codesign`,
      `add-trusted-cert -p codeSign`) and against
      `.loom/scripts/lib/config-resolver.sh` for the `.loom-local/local.json`
      precedence claim (tier 4, highest below env).
- [ ] Operator follow-up (needs a real second Mac): follow the new section on a
      fresh keychain and confirm `codesign -dvv ~/.local/bin/loom-daemon`
      reports `Authority=…` instead of `adhoc`.
- Automated tests: N/A (doc-only change).

Closes #4953

### DIFF
--- a/defaults/docs/macos-tcc-codesign.md
+++ b/defaults/docs/macos-tcc-codesign.md
@@ -86,6 +86,111 @@ machine-local, ungitted override):
 }
 ```
 
+## Provisioning additional Macs
+
+The certificate lives in **one** Mac's login keychain — nothing about it is
+committed, so a second fleet Mac invoked with the *same*
+`LOOM_CODESIGN_IDENTITY` finds no matching identity and falls back to ad-hoc
+signing. `provision-daemon.sh` says so explicitly on stderr before the
+ad-hoc line:
+
+```
+  [loom-daemon] WARNING: LOOM_CODESIGN_IDENTITY 'Loom Local Signing' not found via 'security find-identity -v -p codesigning'; falling back to ad-hoc signing (see defaults/docs/macos-tcc-codesign.md)
+  [loom-daemon] ad-hoc signed /Users/you/.local/bin/loom-daemon (identifier=com.rjwalters.loom-daemon)
+```
+
+That host still runs fine — the only consequence is the deferred one this
+doc exists for: its TCC grants reset on every rebuild. To fix it, give the
+new Mac its own stable identity, either by copying the existing certificate
+(Option 1) or by minting a fresh one there (Option 2).
+
+**Prefer Option 2, and reuse the same Common Name on every host.** A TCC
+grant is per-host, so nothing requires the *certificate* to be shared —
+only that the identity *name* matches what `codesign.identity` /
+`LOOM_CODESIGN_IDENTITY` asks for. Minting a same-named cert per host gets
+you one committed config value with no private key ever leaving the first
+machine. Option 1 exists for when you specifically want one certificate
+across the fleet (e.g. so `codesign -dvv` reports an identical
+`Authority=` leaf everywhere).
+
+### Option 1 — export the cert + key and import it on the second Mac
+
+On the Mac that already has the identity, export both halves. Keychain
+Access is the reliable single-identity path: select the **certificate and
+its private key** together → **File → Export Items…** → `Personal
+Information Exchange (.p12)`. The CLI equivalent exports *every* identity
+in the keychain, so only reach for it on a keychain you know is clean:
+
+```bash
+# Exports ALL codesigning identities in the keychain, not just this one.
+security export -k ~/Library/Keychains/login.keychain-db \
+  -t identities -f pkcs12 -P changeit -o loom-signing.p12
+
+# The public certificate alone, needed for add-trusted-cert on the far side.
+security find-certificate -c "Loom Local Signing" -p \
+  ~/Library/Keychains/login.keychain-db > loom-signing.crt
+```
+
+The `.p12` contains the **private key** — move it over a channel you trust
+(`scp`, AirDrop), use a real passphrase rather than `changeit`, and delete
+it from both hosts once imported.
+
+On the new Mac, import with the same flags the one-time setup above uses
+(`-T /usr/bin/codesign` so later signing is non-interactive; the trust step
+is what keeps the self-signed root from prompting on first use):
+
+```bash
+security import loom-signing.p12 -k ~/Library/Keychains/login.keychain-db \
+  -P changeit -T /usr/bin/codesign
+
+security add-trusted-cert -p codeSign -k ~/Library/Keychains/login.keychain-db \
+  loom-signing.crt
+
+rm -f loom-signing.p12          # the private key does not need to persist on disk
+security find-identity -v -p codesigning   # should now list "Loom Local Signing"
+```
+
+If the import fails with "MAC verification failed", the `.p12` was written
+in OpenSSL 3's default format — re-export it with `-legacy` (same quirk as
+step 2 of Option B above).
+
+### Option 2 — mint an independent cert per host
+
+Just run the one-time setup (Option A or B) again on the new Mac. No key
+material is transferred, and each host's certificate is independent.
+
+- **Same CN on every host** (recommended): use `/CN=Loom Local Signing`
+  again, and the committed `codesign.identity` keeps working unchanged.
+  The certificates differ per host, but each host only ever validates its
+  own signature, so the shared name is all the config needs.
+- **Different CN per host** (e.g. `Loom Local Signing (studio)`): then a
+  single committed value can't cover the fleet. Set the identity per host
+  in `.loom-local/local.json` — the highest-precedence config tier, meant to
+  stay untracked (`install-loom.sh --local` gitignores `/.loom-local/` for
+  you):
+
+  ```json
+  {
+    "codesign": { "identity": "Loom Local Signing (studio)" }
+  }
+  ```
+
+  Exporting `LOOM_CODESIGN_IDENTITY` from the host's shell profile works
+  too and takes precedence over both config tiers.
+
+### Verifying the new host
+
+```bash
+security find-identity -v -p codesigning        # the identity is listed
+./.loom/scripts/cli/loom-daemon-update.sh       # re-provision + re-sign
+codesign -dvv ~/.local/bin/loom-daemon 2>&1 | grep -E 'Authority|adhoc'
+# want: Authority=Loom Local Signing   (and NO 'adhoc' in the flags line)
+```
+
+If `adhoc` is still reported, re-read the provisioning output: the
+`WARNING:` line above pinpoints whether the identity was requested but
+missing from *this* keychain, versus `codesign` itself failing.
+
 ## Grant the daemon identity, not Terminal
 
 Work spawned from an interactive terminal shell — in-session


rjwalters:
LGTM — approving.

Verified the PR's own scope claims rather than taking them at face value:

- **AC1 (warn on absent identity)**: confirmed accurate. `sign_daemon_binary()` / `_pmd_resolve_codesign_identity()` in `scripts/install/provision-daemon.sh` already print the unconditional stderr `WARNING:` on the "requested identity not found" path (landed in #4284, predates this issue), and `_pmd_warn()` has no verbosity gate that could swallow it. Correctly left untouched.
- **AC2 (new doc section)**: this is the substance of the PR. Read the full diff to `defaults/docs/macos-tcc-codesign.md` — Option 1 (export/import via Keychain Access + `security export`/`import`/`add-trusted-cert`, with the "exports all identities" and `-legacy` caveats correctly carried over from the existing one-time-setup section) and Option 2 (mint a per-host cert, recommended, with a clear rationale for why identity *name* match is what matters, not a shared cert) are both accurate and well-scoped. The verification snippet at the end is a nice touch.
- **AC3 (fleet add-worker)**: confirmed out of scope — `fleet add-worker` is Linux-only and its machine-layout step explicitly skips codesign, so there's no macOS fleet command to extend. Correctly not touched.
- **`.loom-local/local.json` gitignore correction**: verified independently — `.loom-local/` is only gitignored when a repo is installed via `install-loom.sh --local` (`loom_local_apply_gitignore`); this repo isn't installed that way, so it is *not* gitignored here by default. The PR's doc text states this conditionally ("`install-loom.sh --local` gitignores `/.loom-local/` for you"), correctly fixing the unconditional claim in the curated issue body's "Affected Files" section. Good catch.

Also ran `bash scripts/check-docs-defaults-parity.sh` locally — OK, no orphans/escaping links.

CI: all applicable checks pass (Rust/dashboard/MCP jobs correctly skip on a docs-only change via Detect Changes). Merge state is CLEAN. Doc-only change, zero code risk, scope matches the curated issue exactly.

Issue #4239, opened 2026-07-28
Resync mechanism rather than uninstall/reinstall


gpeyton:
### Task Type

Feature

### Task Description

Add a resync script to enable updating the local install in a target repo, rather than uninstalling and reinstalling from scratch. 

### Priority (optional)

None

---

### Problem Statement

Updating a Loom installation in a consumer (target) repo today means a **destructive reinstall**: `install.sh --confirm-reinstall` overwrites the existing `.loom/` and requires the operator to inventory local customizations first (install.sh usage, lines 9–14). The partial alternative, `./.loom/scripts/resync-installed.sh` (#3777), already does exactly the right thing — idempotent copy-from-source, per-file updated/unchanged/skipped reporting, `--dry-run`, a `.loom/resync-ignore` pin list for local customizations, and target-repo source resolution via the recorded `loom_source` in `.loom/install-metadata.json` — **but it only covers two surfaces: `.loom/hooks/` and `.loom/scripts/`**.

Everything else an install lays down (`.claude/commands/loom/` role commands and skills, `.loom/roles/`, `.loom/docs/`, `.loom/README.md`, `.github/labels.yml`, workflow files) can only be refreshed by the destructive reinstall. Install metadata also goes stale: this repo's own `.loom/install-metadata.json` still records `loom_version: 0.6.2` from 2026-04-21 while the actual installed version is 0.15.0 — nothing on the update path re-stamps it.

### Current State (verified on `origin/main`)

| Piece | Covers | Gap |
|---|---|---|
| `defaults/scripts/resync-installed.sh` (#3777) | `.loom/hooks/`, `.loom/scripts/` only | The mechanism to generalize — source resolution + ignore convention already work in target repos |
| `defaults/scripts/check-main-freshness.sh` (#3770) | Drift *detection* (advisory) | Suggests resync; only as good as resync's coverage |
| `install.sh --confirm-reinstall` | Full surface | Destructive; requires customization inventory; heavyweight |
| `.loom/install-metadata.json` + `scripts/install/manifest.sh` | Records `loom_source`, `loom_version`, `loom_commit`, `installed_files` manifest | Never refreshed after install (observed 0.6.2-vs-0.15.0 staleness) |
| Daemon binary self-update (#4055, closed) | The `loom-daemon` binary | Separate mechanism; binary is out of scope here |
[...]

PR #4965, opened 2026-08-02
fix(daemon-update): classify ff-abort causes as safely resolvable

### LINKED ISSUE #4951: loom-daemon-update.sh: classify ff-abort causes — content-identical divergence and loom-managed dirty copies are safely resolvable
## Context

During the 2026-08-02 fleet roll, `loom-daemon-update.sh`'s ff-first checkout sync (#4330) aborted on 2 of 3 hosts — both times on local state that was safely resolvable, with the same generic message ("Refusing to guess or hard-reset: resolve manually…"):

1. **robb-STUDIO**: local `main` had 3 local-only commits (a resync commit, its revert, and a catch-up merge — artifacts of sweeps/resyncs committing on main). `git diff origin/main...main` was **empty** — content-identical divergence. Safe resolution was `git reset --hard origin/main`, but the operator had to determine that by hand.
2. **loom-worker-1**: dirty *tracked* files were all loom-managed installed copies (`.gitignore` loom block, `.loom/docs/*.md`, `.loom/hooks/guard-destructive-generic.sh`, `.loom/install-metadata.json`) — machine-local state that `resync-installed.sh` regenerates from `defaults/` on demand. Safe resolution was `git checkout -- .gitignore .loom/` followed by a post-roll resync, again determined by hand.

The hard-abort default is correct (#4330's rationale stands: a silent stale build is worse). But both abort paths can classify the blocking state and either self-resolve or name the exact safe command.

- [ ] Diverged-commits abort: when `git diff origin/<default>...HEAD` is empty (content-identical), say so explicitly and either offer/perform `reset --hard origin/<default>` (flag-gated is fine) or print that exact command in the abort message.
- [ ] Dirty-files abort: when every blocking tracked file is under the loom-managed installed set (`.loom/` installed copies + the managed `.gitignore` block), say so, and either auto-checkout + schedule a post-roll `resync-installed.sh`, or print those exact commands.
- [ ] Genuinely divergent local work (non-empty content diff, or dirty files outside the managed set) keeps today's hard abort unchanged.

---

### Root cause / where to fix

Both abort scenarios flow through the single `sync_with_origin()` failure path in
`defaults/scripts/cli/loom-daemon-update.sh`. The relevant block is the
`git merge --ff-only` failure branch:

```
# defaults/scripts/cli/loom-daemon-update.sh:668-676
echo "Local ${DEFAULT_BRANCH} is ${n} commit(s) behind origin/${DEFAULT_BRANCH} — fast-forwarding before building (default; pass --allow-stale to build the current checkout as-is)..."
if ! (cd "$repo_root" && git merge --ff-only "origin/${DEFAULT_BRANCH}" --quiet); then
    err "Fast-forward merge from origin/${DEFAULT_BRANCH} did not apply — local commits have diverged, or a dirty tracked file conflicts with the incoming change."
    err "Refusing to guess or hard-reset: resolve manually (rebase/merge by hand), or pass --allow-stale to build the current (stale) checkout as-is."
    return 1
fi
```

Today this single generic message covers two structurally distinct failure
modes (`git merge --ff-only` fails both when local commits have diverged from
origin AND when a dirty tracked file blocks the merge). The fix adds a
classification step **inside this `if` block, before the generic `err` calls**,
that distinguishes:

1. **Content-identical divergence** — `git diff origin/${DEFAULT_BRANCH}...${DEFAULT_BRANCH}` (note: use the three-dot merge-base form, matching the incident's own `git diff origin/main...main`) is empty. Safe resolution: `git reset --hard origin/${DEFAULT_BRANCH}`.
2. **Loom-managed dirty files only** — every path returned by `git status --porcelain` (tracked, non-clean) is inside the installed surface map documented in `defaults/scripts/resync-installed.sh` (lines 20-28): `.loom/hooks/`, `.loom/scripts/`, `.loom/roles/`, `.loom/docs/`, `.loom/runtimes/`, `.loom/bin/`, `.claude/commands/loom/`, `.loom/install-metadata.json`, and the marker-delimited Loom-managed block inside `.gitignore`. Safe resolution: `git checkout -- <those paths>` followed by scheduling (or directly invoking) `./.loom/scripts/resync-installed.sh` post-roll.
3. **Everything else** (non-empty content diff, or any dirty file outside the managed set) — keep today's hard abort verbatim. This is the safety-critical branch: #4381 (referenced in the existing test suite's header, `defaults/scripts/tests/test-loom-daemon-update.sh:27-44`) was a live incident where an update-script code path wrote over a real production binary — the same class of "automation quietly does something destructive to real machine state" risk applies here if the content-identical or managed-file-only checks are too loose (e.g. a substring/prefix match on `.loom/` that also matches an unmanaged repo-specific file, or a diff-empty check that misses a rename/mode-only change). Do not widen the auto-resolve conditions beyond exactly what's specified above; when in doubt, fall through to the existing hard abort.

Auto-performing `reset --hard` / `checkout --` is explicitly "flag-gated is fine"
per the issue's own AC — the simplest compliant implementation is to keep the
default behavior as "classify and print the exact safe command" (no behavior
change to what actually executes), and gate the auto-resolve variant behind a
new flag (e.g. `--auto-resolve-safe-abort`) if attempting the fully-automatic
path. Either satisfies the AC; a curator note, not a mandate — pick whichever
keeps the diff smaller.

### Affected Files

- `defaults/scripts/cli/loom-daemon-update.sh` - add the classification logic inside the `git merge --ff-only` failure branch (~line 669-673 today) in `sync_with_origin()`; needs a helper to enumerate `git status --porcelain` paths and check each against the managed-surface list from `resync-installed.sh`.
- `defaults/scripts/resync-installed.sh` - read-only reference for the authoritative managed-surface list (header comment, lines 20-28); do not duplicate this list by hand without a comment pointing back here, since #4239 already widened it once and it can widen again.
- `defaults/scripts/tests/test-loom-daemon-update.sh` - add test cases for both new classification paths (content-identical diverged commits → reset-hard path; dirty-but-only-managed-files → checkout+resync path) alongside the existing genuinely-diverged hard-abort test; reuse the existing fake-repo/fake-cargo/fake-daemon harness described in the file's own header comment (lines 1-24) rather than inventing a second harness.

### Test Plan

- [ ] Automated tests in `defaults/scripts/tests/test-loom-daemon-update.sh`:
  - New case: local `${DEFAULT_BRANCH}` has commits not on origin but `git diff origin/${DEFAULT_BRANCH}...${DEFAULT_BRANCH}` is empty → script detects content-identical divergence and either performs (if flag-gated auto-resolve is implemented and the flag is passed) or prints the exact `git reset --hard origin/${DEFAULT_BRANCH}` command.
  - New case: only loom-managed installed files (e.g. a modified `.loom/docs/*.md`, or the managed `.gitignore` block) are dirty → script detects managed-only dirty state and either performs or prints the exact `git checkout -- ...` + resync commands.
  - Regression case: a genuinely diverged commit (non-empty content diff) still hard-aborts with no auto-resolve attempt.
  - Regression case: a dirty tracked file OUTSIDE the managed set (e.g. an unrelated repo file) still hard-aborts with no auto-resolve attempt, even if other dirty files ARE in the managed set (the AC's "every blocking file" wording is conjunctive, not "any").
- [ ] Manual verification: reproduce each of the two 2026-08-02 incident scenarios in a scratch clone (3 local-only commits that net to content-identical vs. origin; dirty-only `.loom/docs/*.md`) and confirm the script now classifies and names/performs the safe resolution instead of the generic abort message.
- [ ] Regression: run the full existing suite (`./defaults/scripts/tests/test-loom-daemon-update.sh`) to confirm the #4381 production-binary checksum guard (file header, lines 27-44) still passes and no existing case regresses.

rjwalters:
## Summary

`sync_with_origin()`'s `git merge --ff-only` failure branch in
`defaults/scripts/cli/loom-daemon-update.sh` used to emit ONE generic
"resolve manually" message for two structurally distinct, often-safe failure
shapes — both hit on the 2026-08-02 fleet roll:

1. **Content-identical divergence**: local `<default>` diverged from
   `origin/<default>` in commit history, but `git diff origin/<default>...
   <default>` is empty (e.g. a resync commit + its own revert netting to no
   change — the robb-STUDIO incident shape). Safe resolution: `git reset
   --hard origin/<default>`.
2. **Dirty-but-managed-only**: the ff-merge is blocked by dirty tracked
   file(s) that are ALL Loom-managed installed copies (mirroring
   `resync-installed.sh`'s own installed-surface map: `.loom/hooks/`,
   `.loom/scripts/`, `.loom/roles/`, `.loom/docs/`, `.loom/runtimes/`,
   `.loom/bin/`, `.claude/commands/loom/`, `.loom/install-metadata.json`,
   plus the marker-delimited managed block inside `.gitignore` — the
   loom-worker-1 incident shape). Safe resolution: `git checkout --` those
   paths, then re-run `resync-installed.sh`.

By default, the abort message now names the exact safe command instead of a
bare "resolve manually" — **no behavior change to what actually executes**.
A new `--auto-resolve-safe-abort` flag performs the resolution automatically
instead.

Any other cause — genuine content divergence, or **any** dirty file outside
the managed set (even alongside managed ones — the classification is
conjunctive, not "any") — still hard-aborts completely unchanged. This is
the safety-critical branch (#4381's "automation destroys real machine state"
precedent), so the classification is deliberately narrow and falls through
to the existing hard abort whenever genuinely unsure.

## Implementation notes

- `_ff_abort_content_identical()`'s `git merge-base --is-ancestor` guard is
  load-bearing, not an optimization: without it, the three-dot diff
  trivially reduces to "diff local HEAD against itself" whenever local
  hasn't diverged in commit history (the far more common "just behind,
  blocked by a dirty file" shape) — which is *always* empty regardless of
  what origin changed, and would misclassify essentially every plain
  dirty-file abort as "content-identical," risking an incorrect
  `reset --hard` under `--auto-resolve-safe-abort`.
- The managed-surface prefix list is hand-mirrored from
  `resync-installed.sh`'s own header comment (not sourced), with a comment
  pointing back so it's updated in lockstep if that list ever widens again
  (as #4239 already did once).
- `.gitignore` gets special handling: only the marker-delimited Loom-managed
  block (`# >>> loom-managed (do not edit) >>>` / `# <<< loom-managed <<<`,
  matching `loom-daemon`'s `GITIGNORE_BEGIN_MARKER`/`GITIGNORE_END_MARKER`)
  is treated as managed — a consumer's own hand-edited lines outside that
  block are never silently discarded.

## Test plan

- [x] `./defaults/scripts/tests/test-loom-daemon-update.sh` — 153/153 passing
      (5 new test cases added: content-identical default + auto-resolve,
      managed-only-dirty default + auto-resolve, and a regression case
      proving an unmanaged dirty file alongside a managed one still forces
      the hard abort even with `--auto-resolve-safe-abort` passed).
- [x] Fixed test 38's fixture, which previously diverged using two
      `--allow-empty` commits on each side — that setup coincidentally
      produces an empty three-dot diff, so it would have collided with the
      new content-identical classification. Replaced with real (non-empty)
      content divergence on both sides so it still exercises the genuine
      hard-abort regression path.
- [x] `shellcheck` clean on both changed files (only the pre-existing
      unrelated `SC2015` info-level note at line 825).
- [x] The existing #4381 production-binary checksum guard test still passes.

Closes #4951

### DIFF
--- a/defaults/scripts/cli/loom-daemon-update.sh
+++ b/defaults/scripts/cli/loom-daemon-update.sh
@@ -124,6 +124,7 @@
 #   ./.loom/scripts/cli/loom-daemon-update.sh --no-restart  Rebuild + provision only; leave the running daemon untouched
 #   ./.loom/scripts/cli/loom-daemon-update.sh --relaunch    Launchd/systemd only: after a refused restart, re-render the plist/unit and relaunch under supervision (SIGTERMs the daemon so sweep children reparent; preserves the live LOOM_* env)
 #   ./.loom/scripts/cli/loom-daemon-update.sh --allow-stale Skip the default ff-first sync with origin/<default-branch> and build the current (possibly stale) checkout as-is (#4330) — for deliberate use: bisecting, testing a local patch
+#   ./.loom/scripts/cli/loom-daemon-update.sh --auto-resolve-safe-abort  When the ff-only sync would otherwise hard-abort (#4330), auto-perform the fix IF the blocking state classifies as safe (#4951): content-identical diverged commits (`git reset --hard origin/<default-branch>`), or dirty tracked files that are ALL Loom-managed installed copies (`git checkout --` them + re-run resync-installed.sh). Any other cause (genuine content divergence, or any unmanaged dirty file) still hard-aborts unchanged — this flag never widens what's classified as safe, only whether the safe cases are printed (default) or performed
 #   ./.loom/scripts/cli/loom-daemon-update.sh --help
 #
 # Environment:
@@ -195,7 +196,12 @@
 #      (diverged local commits, a dirty tracked file conflicting with the
 #      incoming change, or HEAD is not on <default-branch>) — the script
 #      NEVER guesses or hard-resets; resolve manually or pass --allow-stale
-#      (#4330)
+#      (#4330). Two of the ff-abort causes classify as safely resolvable
+#      (#4951): content-identical diverged commits, or dirty tracked files
+#      that are ALL Loom-managed installed copies — by default these still
+#      exit 1 but the abort message names the exact safe command; pass
+#      --auto-resolve-safe-abort to have the script perform it instead (exit
+#      0 on success). Any other cause still hard-aborts unchanged.
 #   3  (--check only) update available
 #   4  build verification FAILED: the freshly-built binary's embedded commit
 #      does not match the source HEAD it was built from. This is a BUILD-SYSTEM
@@ -527,6 +533,7 @@ CHECK_ONLY=false
 NO_RESTART=false
 RELAUNCH=false
 ALLOW_STALE=false
+AUTO_RESOLVE_SAFE_ABORT=false
 [[ "${LOOM_DAEMON_UPDATE_RELAUNCH:-}" =~ ^(1|true|yes)$ ]] && RELAUNCH=true
 while [[ $# -gt 0 ]]; do
     case "$1" in
@@ -537,6 +544,7 @@ while [[ $# -gt 0 ]]; do
         --no-restart) NO_RESTART=true; shift ;;
         --relaunch) RELAUNCH=true; shift ;;
         --allow-stale) ALLOW_STALE=true; shift ;;
+        --auto-resolve-safe-abort) AUTO_RESOLVE_SAFE_ABORT=true; shift ;;
         *) err "Unknown option '$1'"; echo "Use --help for usage" >&2; exit 1 ;;
     esac
 done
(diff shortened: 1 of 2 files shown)


rjwalters:
Code evaluation in progress; CI (Shell Test Suites) is still pending. Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Code evaluation pending; CI is still running (Shell Test Suites hermetic check is pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Changes requested — found a real gap in the safety-critical classification logic, confirmed by direct reproduction (not just code reading).

## Summary of what I checked
- Read the full diff of `sync_with_origin()`'s new classification helpers (`_ff_abort_content_identical`, `_ff_abort_all_dirty_tracked_managed`, `_ff_abort_is_managed_path`, `_ff_abort_gitignore_only_managed_block_dirty`) in `defaults/scripts/cli/loom-daemon-update.sh`.
- Ran the full test suite: `153/153` passing, matching the PR's claim.
- Ran `shellcheck` on the changed file: clean except the pre-existing `SC2015` info note, matching the PR's claim.
- The managed-prefix matching correctly uses a trailing `/` on every directory prefix, so no substring/prefix false-positive risk there (e.g. `.loom/hooks-custom/` cannot match the `.loom/hooks/` prefix).
- The conjunctive "every dirty file must be managed" logic in `_ff_abort_all_dirty_tracked_managed` is correctly implemented and covered by regression test 61 (mixed managed+unmanaged dirty file still hard-aborts, even with `--auto-resolve-safe-abort`).
- Test 38's fixture fix (real content divergence instead of `--allow-empty` commits, which coincidentally collided with the new content-identical classification) is a legitimate and necessary fix.

## Blocking issue: `_ff_abort_content_identical()` ignores the working tree entirely

`_ff_abort_content_identical()` classifies purely by comparing two **committed** refs (`git merge-base --is-ancestor` + `git diff --quiet origin/<default>...<default>`). It never checks `git status --porcelain` for a dirty **uncommitted** tracked file. I reproduced this directly against real git plumbing (not the full script harness, but the exact same commands the helper uses):

```
=== status before ===
 M unmanaged-file.txt          # an UNRELATED, uncommitted, unmanaged dirty file
=== attempt ff-only merge ===
fatal: Not possible to fast-forward, aborting.       # fails due to diverged history alone
=== is content identical? (diff origin/trunk...HEAD) ===
diff --quiet exit (0=identical): 0                   # <- returns TRUE despite the dirty file
=== unmanaged file content still dirty (unaffected by diff check)? ===
unmanaged-local-edit
```

Concretely: a host can simultaneously have (a) diverged local commits that net to zero content diff vs. origin (the robb-STUDIO shape) **and** (b) an entirely unrelated dirty tracked file sitting in the working tree (e.g. an operator's manual scratch edit, or *any* unmanaged file — this isn't limited to Loom-managed paths). `git merge --ff-only` fails on the history divergence alone, with or without that dirty file. `_ff_abort_content_identical` returns true regardless, because it never looks at the working tree. Under `--auto-resolve-safe-abort`, the script then runs `git reset --hard origin/<default-branch>`, which silently discards that dirty file's uncommitted changes — with **zero** check for whether it's managed or not.

This directly contradicts the PR description's own stated invariant: *\"Any other cause — genuine content divergence, or any dirty file outside the managed set (even alongside managed ones — the classification is conjunctive, not 'any') — still hard-aborts completely unchanged.\"* That conjunctive guarantee is only implemented in the second classifier (`_ff_abort_all_dirty_tracked_managed`); the first classifier has no equivalent guard at all. Given the explicit #4381 framing (\"automation quietly does something destructive to real machine state\"), this is exactly the failure mode the PR is trying to close, just via the other branch.

It's also untested: tests 57/58 (content-identical divergence) only exercise a fixture with an otherwise-clean working tree. There's no test combining content-identical commit divergence with a co-existing dirty tracked file (managed or unmanaged) — that combination is exactly what would slip through today.

**Suggested fix**: have `_ff_abort_content_identical()` also require a clean working tree relative to local HEAD (e.g. `git status --porcelain` returns no non-`??` entries, or `git diff --quiet HEAD --`) before returning true — falling through to the existing hard-abort (or letting the dirty-file classifier evaluate it) otherwise. This mirrors the strictness already correctly applied in the sibling classifier and closes the gap without weakening anything else in the PR.

Everything else in the change (flag wiring, help/exit-code doc updates, the managed-surface prefix list mirroring `resync-installed.sh`, the `.gitignore` marker-block-only handling, the `resync-installed.sh` resolution/invocation, and the rest of the new tests) looks correct and well-reasoned. This is the one gap that needs closing before merge given how safety-critical this code path is.

rjwalters:
Doctor pass: PR still carries a fresh `loom:treating` claim (claimed 2026-08-02T20:54:22Z) — standing down without reclaiming. Not stomping.

rjwalters:
Sweep recovery (mid-phase-death, issue #3676 pattern): found a completed, unpushed Doctor commit in the issue-4951 worktree (`d7523250`, "require a clean tree before the content-identical ff-abort auto-resolve") that directly addresses the prior Judge rejection. The `loom:treating` claim (2026-08-03T01:28:21Z) had no live process behind it and the commit was never pushed before the session died.

Verified before pushing: working tree clean, full `test-loom-daemon-update.sh` suite run (135 assertions, 0 failures). Pushed the commit and re-labeled back to `loom:review-requested` for a fresh Judge pass rather than dispatching a duplicate Doctor.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T02:45:08Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T02:45:08Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved — the Doctor fix directly and correctly addresses the prior rejection.

## What I verified independently

**The gap**: my prior review found `_ff_abort_content_identical()` classified purely by comparing two *committed* refs (`git merge-base --is-ancestor` + `git diff --quiet origin/<default>...<default>`), never checking the working tree. A host with (a) diverged local commits that net to zero content diff vs. origin, plus (b) an unrelated dirty tracked file, would misclassify as content-identical — letting `--auto-resolve-safe-abort`'s `git reset --hard` silently discard that file's uncommitted changes, managed or not.

**The fix** (commit d7523250): adds `_ff_abort_no_dirty_tracked_files()` (no non-`??` entry in `git status --porcelain`) and requires it inside `_ff_abort_content_identical()` before the ref-diff check runs. Any dirty tracked file now falls through to the managed-dirty classifier and then to the existing hard abort — this is exactly the fix I suggested, and it doesn't weaken the sibling `_ff_abort_all_dirty_tracked_managed` classifier at all. Header/exit-code docs were updated to state the clean-tree requirement.

**New tests 62/63** cover precisely the scenario I reproduced: content-identical divergence co-existing with a dirty UNMANAGED file (must survive, generic hard-abort) and with a dirty MANAGED file (no `reset --hard` claimed, HEAD untouched) — both under `--auto-resolve-safe-abort`.

## Independent verification performed
- Read the full diff (both commits) in `defaults/scripts/cli/loom-daemon-update.sh` and the test file — confirmed the new guard is a strict tightening (returns false/falls-through on any doubt), consistent with the PR's own "#4381 — when genuinely unsure, return false" safety note.
- Ran the full test suite myself in the builder worktree (not just trusting the reported numbers): **162/162 passing, 0 failures**, including the new tests 62/63.
- Ran `shellcheck` on both changed files myself: clean, exit 0.
- `gh pr checks 4965`: all required checks pass (Shell Test Suites hermetic, Installer Integration Tests, Shellcheck, Docs/Defaults Parity, AGENTS.md Sync, CLAUDE.md Line Budget, Codex Adapter Smoke, Detect Changes). `Rust *` / `Dashboard Tests` / `MCP Server Build` are `skipping` (no Rust/dashboard files touched — expected for a shell-only change).

Code quality, safety reasoning, and test coverage are all solid. Approving.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present — another Judge's verdict raced mine (approved at 2026-08-03T02:58:47Z, reaching the same conclusion: Doctor's fix correctly closes the prior rejection's gap). Discarding my review; not writing any label.

PR #4968, opened 2026-08-02
fix(fleet): local tailnet self-check before reporting all-hosts-UNREACHABLE

### LINKED ISSUE #4952: fleet status: all-hosts-UNREACHABLE should trigger a local tailnet self-check before reporting a fleet outage
## Context

On 2026-08-02, Tailscale on the fleet's control host (robb-STUDIO) silently stopped. Every SSH to remote workers (robb-pro, loom-worker-1) timed out — `ssh: connect to host 100.x.x.x port 22: Operation timed out` — which is byte-for-byte what a genuine multi-host outage looks like. Both hosts were actually healthy; `tailscale status` said "Tailscale is stopped." and `open -a Tailscale` fixed everything in seconds.

`loom-daemon fleet status` already prints loud per-host states (`UP` / `DAEMON DOWN` / `UNREACHABLE` / …), and its help text promises "silence never reads as idle". But an all-hosts-`UNREACHABLE` report during a local tailnet outage still misleads: the correct diagnosis (and one-command fix) is on the *local* host, and the tell — every tailnet peer failing simultaneously while non-tailnet networking works — is checkable.

- [ ] When every remote roster host that has a tailnet address is `UNREACHABLE`, `fleet status` runs a cheap local tailnet self-check (e.g. `tailscale status` exit/parse, or reachability of the local tailscaled socket) before printing the report.
- [ ] If the local tailnet is down/stopped, the report says so prominently (e.g. a `LOCAL TAILNET DOWN — remote states unknown` banner) instead of presenting N× `UNREACHABLE` as fact, and names the restart action.
- [ ] Hosts without tailnet addresses (plain LAN/SSH) are excluded from the inference so a mixed roster can still report a real partial outage.
- [ ] No new hard dependency: absent `tailscale` CLI ⇒ skip the check silently (current behavior unchanged).

---

### Implementation Guidance

The fleet status pipeline is `loom-daemon/src/fleet/status.rs` (merge/render/exit-code logic) wired up by `loom-daemon/src/cli/status.rs::handle_fleet_status_command` (thin clap entry point, per that file's own module doc: "Thin clap→module wiring").

**Where the self-check fits:**

- `WorkerRecord` already carries an optional `tailnet_name` (`HostReport.tailnet_name: Option<String>`, `status.rs:160-163`), so "every remote roster host that has a tailnet address" (AC 1) is directly checkable via that field — no new registry schema needed.
- The existing `SafehousedProbe` enum (`status.rs:130-149`, `Present` / `Absent` / `Unknown`) is the closest existing pattern for a "cheap best-effort local/remote probe that must never fail the whole report" — a new `LocalTailnetState` (or similar) enum mirroring that shape (e.g. `Up` / `Down` / `NotInstalled`/`Unknown`) is the natural analog. Note `SafehousedProbe` is a *remote* probe run over SSH per-host; the new check is explicitly *local-only* (run once, not per-host) — reuse the enum shape, not the SSH plumbing.
- `summarize()` (`status.rs:650-672`) and `FleetStatusSummary` (`status.rs:548-566`) are where the "every remote roster host with a tailnet address is UNREACHABLE" condition (AC 1) should be detected, since that's already where per-state counts are tallied.
- `FleetStatusReport::render_human()` (`status.rs:691-736`) is where the `LOCAL TAILNET DOWN — remote states unknown` banner (AC 2) would be inserted — likely alongside the existing `empty_roster` notice pattern (`status.rs:694-699`), which is the precedent for "prepend an explanatory banner above the table."
- The local `tailscale status` invocation itself belongs in `handle_fleet_status_command` (`cli/status.rs:483-533`) or a small helper next to `collect_local_fleet_report` (`cli/status.rs:542-582`) — this is a **local** shell-out (`std::process::Command::new("tailscale").args(["status", ...])`), not an SSH round-trip, so it does not belong in `HostStatusSource`/`SshStatusSource` (`status.rs:265-334`), which is exclusively the per-remote-host SSH seam.
- AC 4 ("absent `tailscale` CLI ⇒ skip the check silently") maps directly to handling `Command::new("tailscale")...spawn()`'s `Err(ErrorKind::NotFound)` case — mirror how `SafehousedProbe::Unknown` already degrades an inconclusive probe without erroring the row (`status.rs:321-328`).
- `--json` output: add the local tailnet self-check result as a new field on `FleetStatusReport` (or `FleetStatusSummary`) so a scripted consumer can distinguish "local tailnet down" from "genuine multi-host outage" the same way `empty_roster` already lets a script distinguish "no fleet yet" from "fleet all healthy" (`status.rs:561-565`).

### Affected Files

- `loom-daemon/src/fleet/status.rs` - add the local-tailnet-state enum, extend `FleetStatusSummary`/`FleetStatusReport` with the check result, detect the "all tailnet hosts UNREACHABLE" condition in `summarize()`, and add the warning banner + host exclusion logic in `render_human()`
- `loom-daemon/src/cli/status.rs` - run the local `tailscale status` shell-out in (or near) `handle_fleet_status_command`/`collect_local_fleet_report`, handle a missing `tailscale` CLI gracefully (AC 4), and thread the result into the report before rendering

### Test Plan

- [ ] Manual verification: with a fleet registry containing tailnet-addressed hosts, stop the local `tailscale` daemon (`tailscale down` or equivalent) and run `loom-daemon fleet status` — confirm the `LOCAL TAILNET DOWN` banner appears instead of a flat list of `UNREACHABLE` hosts, and that the banner names the restart action.
- [ ] Automated tests (in `loom-daemon/src/fleet/status.rs`'s existing `#[cfg(test)]` module, which already mocks `HostStatusSource` — see `MockSource`/`MockOutcome` at `status.rs:770-820`):
  - All tailnet-addressed roster hosts `Unreachable` + local tailnet check reports down → banner rendered, hosts not reported as a flat outage.
  - All tailnet-addressed roster hosts `Unreachable` + local tailnet check reports up → normal `UNREACHABLE` reporting unchanged (regression guard).
  - Mixed roster (some hosts with `tailnet_name: None`, i.e. plain LAN/SSH) + local tailnet down → non-tailnet hosts are excluded from the inference and still reported on their own merits (AC 3).
  - `tailscale` CLI absent (simulate `Command::new` returning `NotFound`) → check is a silent no-op, current behavior unchanged (AC 4).
- [ ] Edge cases: empty roster (no hosts at all) combined with local tailnet down — confirm no panic and the existing `empty_roster` notice still renders correctly alongside/instead of the new banner.

rjwalters:
## Summary

Closes #4952

On 2026-08-02, Tailscale on the fleet's control host silently stopped, so every SSH round-trip to remote workers timed out — indistinguishable from a genuine multi-host outage. `fleet status` now detects the "every tailnet-addressed roster host is UNREACHABLE" pattern and runs a cheap, local-only `tailscale status --json` self-check before presenting that as a fleet-wide outage.

- Added `LocalTailnetState` (`Up` / `Down` / `Unknown`) in `loom-daemon/src/fleet/status.rs`, mirroring the existing `SafehousedProbe` shape but explicitly local-only (never SSH).
- Added `all_tailnet_hosts_unreachable(hosts)` — true only when every *remote, tailnet-addressed* roster host is `Unreachable`; hosts with `tailnet_name: None` (plain LAN/SSH) are excluded from the inference so a mixed roster can still surface a genuine partial outage.
- `FleetStatusReport` gained a `local_tailnet: Option<LocalTailnetState>` field (serialized in `--json` output, `skip_serializing_if` when never run).
- `handle_fleet_status_command` (`loom-daemon/src/cli/status.rs`) runs the local `tailscale status --json` shell-out only when `all_tailnet_hosts_unreachable` is true, and only after the SSH fanout — it is deliberately not part of `HostStatusSource`/`SshStatusSource`, which is exclusively the per-remote-host SSH seam.
- `render_human()` prepends a `LOCAL TAILNET DOWN — remote states below are UNKNOWN, not a confirmed fleet outage` banner (mirroring the existing `empty_roster` notice pattern) that names the restart action (`tailscale up` / `open -a Tailscale`) when the local check reports `Down`.
- Absent `tailscale` CLI (`ErrorKind::NotFound`) degrades to a silent no-op — no new hard dependency, current behavior unchanged.

## Acceptance criteria

- [x] When every remote roster host with a tailnet address is `UNREACHABLE`, `fleet status` runs the local tailnet self-check before printing the report.
- [x] If the local tailnet is down, the report shows a prominent `LOCAL TAILNET DOWN` banner instead of presenting N x `UNREACHABLE` as fact, and names the restart action.
- [x] Hosts without tailnet addresses are excluded from the inference.
- [x] Absent `tailscale` CLI skips the check silently (no new hard dependency).

## Test plan

- `cargo test --lib fleet::status` (26 tests, including new coverage for `all_tailnet_hosts_unreachable`, the mixed-roster exclusion case, the vacuous/empty-roster case, and `render_human` banner presence/absence) — pass.
- `cargo test --bin loom-daemon` (`cli::status::local_tailnet_tests`, including a PATH-mutation test proving an absent `tailscale` CLI is a silent no-op) — pass.
- `cargo build`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` — clean.

`cargo test` (full workspace `--lib` run) has 9 pre-existing failures in `auto_update::tests::*` and `work_finder::tests::*` — verified to reproduce identically on a fresh clone of unmodified `origin/main` (unrelated to this PR's two changed files, `loom-daemon/src/cli/status.rs` and `loom-daemon/src/fleet/status.rs`). They come from `config_resolver::resolve_effective_config` picking up a machine-level private-defaults config on this host regardless of the test's `tempdir` `repo_root`, not from anything touched here.

### DIFF
--- a/loom-daemon/src/cli/status.rs
+++ b/loom-daemon/src/cli/status.rs
@@ -482,7 +482,8 @@ fn print_status_unreachable_human(
 /// this binary's own socket/install-state machinery) lives here.
 pub(crate) async fn handle_fleet_status_command(json: bool) -> Result<()> {
     use loom_daemon::fleet::status::{
-        collect_fleet_report, update_last_seen_up_at, SshStatusSource,
+        all_tailnet_hosts_unreachable, collect_fleet_report, update_last_seen_up_at,
+        SshStatusSource,
     };
     use loom_daemon::fleet::FleetRegistry;
 
@@ -493,7 +494,16 @@ pub(crate) async fn handle_fleet_status_command(json: bool) -> Result<()> {
     let source: Arc<dyn loom_daemon::fleet::status::HostStatusSource> =
         Arc::new(SshStatusSource::new());
     let timeout = Duration::from_secs(loom_daemon::fleet::status::DEFAULT_TIMEOUT_SECS);
-    let report = collect_fleet_report(source, registry, local, timeout).await;
+    let mut report = collect_fleet_report(source, registry, local, timeout).await;
+
+    // #4952: when every tailnet-addressed roster host is UNREACHABLE, that
+    // pattern is indistinguishable from a genuine multi-host outage without
+    // checking whether the LOCAL tailnet client is actually the thing that
+    // is down (the 2026-08-02 robb-STUDIO incident this issue documents).
+    // Cheap, local-only, best-effort: never blocks the report on failure.
+    if all_tailnet_hosts_unreachable(&report.hosts) {
+        report.local_tailnet = check_local_tailnet_state();
+    }
 
     // #4697: persist "last observed Up" so a LATER poll that finds this host
     // Unreachable can measure elapsed silence against its configured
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-02T21:31:53Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

Reviewed against issue #4952's acceptance criteria, all four met:

- AC1: `all_tailnet_hosts_unreachable()` correctly triggers only when every *remote, tailnet-addressed* roster host is `Unreachable` (local-host row and non-tailnet hosts excluded from the scan), vacuously `false` when no tailnet hosts exist at all.
- AC2: `render_human()` prepends a `LOCAL TAILNET DOWN` banner naming the restart action (`tailscale up` / `open -a Tailscale`) above the per-host table — reframes the UNREACHABLE rows rather than hiding them, consistent with the issue's own suggested banner text. `--json` output also carries the new `local_tailnet` field (`skip_serializing_if` keeps old consumers unaffected when never run).
- AC3: mixed-roster case verified by `all_tailnet_hosts_unreachable_excludes_hosts_without_tailnet_address` — a plain LAN/SSH host's own `Unreachable` state still surfaces on its own merits.
- AC4: absent `tailscale` CLI (`ErrorKind::NotFound`) returns `None` and is a true no-op — exercised end-to-end in `check_local_tailnet_state_is_none_when_tailscale_cli_absent` against an empty `PATH`, `#[serial]`-guarded against parallel PATH mutation.

Other things checked:
- `exit_code()` is unaffected by `local_tailnet` (still driven purely by host states) — the self-check is additive context, not a behavior change to the non-zero-exit contract.
- The local shell-out (`Command::new("tailscale")`) runs directly in `handle_fleet_status_command` rather than via `tokio::task::spawn_blocking` (unlike the two per-host SSH probes in this same file) — fine here since `fleet status` is single-shot CLI dispatch, not a long-lived multiplexed server loop, and the check only runs once per invocation, but flagging for awareness in case this code path is ever reused inside a longer-lived async context.
- `cargo test -p loom-daemon` for both the `fleet::status` and `cli::status::local_tailnet_tests` modules: 26 + 6 tests, all pass. `cargo clippy -p loom-daemon --all-targets` and `cargo fmt --check` both clean.
- CI green (Rust Linting & Build, OTLP Feature Build & Test, Unit Tests all pass; dashboard/installer/shell suites correctly skip for a Rust-only change). `mergeStateStatus` is CLEAN.

No security concerns — the shell-out uses fixed args (`["status", "--json"]`), no user-controlled input, `stdin(Stdio::null())`.

PR #4966, opened 2026-08-02
fix(champion): stop duplicate NEEDS REVISION comments on unrevised proposals

rjwalters:
## Summary

Champion re-evaluated unchanged `loom:architect`/`loom:curated`/`loom:hermit`/`loom:auditor` proposals every cycle, posting a near-identical "NEEDS REVISION" comment each time — observed live as **6 duplicate comments over ~6.5 hours** on one proposal (`2AMLogic/klayout-tools#168`), with two evaluations landing 40 seconds apart because nothing claimed the issue while it was being evaluated.

This adds the two missing guardrails from the issue, modeled on `champion-pr-merge.md`'s existing Capped-PR Recovery pattern (idempotency marker + escalation marker + `loom:operator-only` routing) and Judge's `loom:reviewing` claim convention:

1. **Idempotency check**: a marker keyed to the issue's `updatedAt` (`<!-- champion:proposal-verdict:$UPDATED_AT -->`). If Champion's most recent verdict comment already carries this marker for the issue's current `updatedAt`, skip re-evaluation entirely — no comment, no re-read. A genuine body revision bumps `updatedAt` and always produces a fresh evaluation.
2. **N=2 escalation**: once 2 "NEEDS REVISION" comments exist on a proposal, the next evaluation escalates to `loom:operator-only` (with a `<!-- champion:proposal-escalated -->`-marked comment) instead of posting a 3rd+ near-identical rejection. `loom:operator-only` is already excluded from every future promotion pass, so the escalation comment posts exactly once.
3. **Concurrency guard**: a new `loom:evaluating` claim label (staleness-aware, `LOOM_STALE_EVALUATING_MINUTES`, default 15m, sized for evaluation's seconds-to-minutes duration rather than Judge's review-duration timescale) prevents two Champion passes from evaluating the same proposal simultaneously, plus a verdict-time recheck immediately before writing the outcome.

## Changes

- `defaults/.claude/commands/loom/champion-issue-promo.md`: new "Concurrency Guard and Idempotency (`loom:evaluating`)" section (idempotency check, staleness-aware claim, verdict-time recheck); Step 3 (promote) and Step 4 (reject) updated to release the claim and, in Step 4, to check the N=2 escalation threshold before posting another rejection; Batch Processing section updated with per-issue claim/skip ordering.
[...]

Issue #4954, opened 2026-08-02
Champion re-evaluates unchanged loom:architect proposals every cycle — 6 duplicate NEEDS REVISION comments before self-routing to operator


rjwalters:
## Context

Observed live on `2AMLogic/klayout-tools#168` (2026-08-01): an unchanged
`loom:architect` proposal was re-evaluated by the Champion **six times in
~6.5 hours** (06:32, 07:52, 09:30, 11:22, 12:33, 12:34 — the last two under
a minute apart), each cycle posting a near-identical "NEEDS REVISION" comment
with the same three findings (epic-shaped scope, five open design questions,
real-money risk) and the same recommendations. On the seventh cycle the
Champion itself diagnosed the loop — quote from its final comment:

> This proposal has now been evaluated and rejected 6 times with converging
> feedback ... but no revision has landed and the issue keeps re-entering the
> queue unchanged every cycle. Re-running an identical evaluation each cycle
> only produces duplicate comments; it doesn't move this forward.

— and applied `loom:operator-only` to stop the cycle. The self-correction is
good, but it burned six full evaluations (tokens + queue slots) and six
duplicate comments before triggering, and the two comments 40 seconds apart
suggest concurrent evaluators aren't aware of each other either.

## Repro

1. File a `loom:architect` proposal that fails promotion criteria (e.g.
   contains an explicit open-questions section).
2. Leave it unrevised across several sweep cycles.
3. Champion re-evaluates and posts an equivalent NEEDS REVISION comment each
   cycle until (eventually) escalating.

## Suggested acceptance criteria

- [ ] Before evaluating a `loom:architect` issue, Champion checks whether its
      own most recent comment on that issue post-dates the last body edit
      (`updatedAt` vs comment timestamp, or a body-hash stamp in the comment)
      — if so, skip without commenting
- [ ] After N identical verdicts on an unchanged issue (N=2 seems right,
      given 6 happened here), escalate to `loom:operator-only` routing
      immediately rather than continuing to comment
- [ ] Concurrent double-evaluation (two comments 40s apart) prevented or
      collapsed — e.g. a claim/lock convention like Builder's claim label

rjwalters:
## Curator Enhancement

[...]

PR #4969, opened 2026-08-02
fix(champion): bound the silent idempotency skip so unrevised proposals still escalate

### LINKED ISSUE #4967: Champion's body-hash idempotency fix (#4966) makes N=2 escalation unreachable for genuinely unrevised proposals
## Context

While independently judging PR #4966 (found already approved — my review raced
another Judge's, see standing-down comment on the PR), I traced the same
"Concurrency Guard and Idempotency" mechanism in
`defaults/.claude/commands/loom/champion-issue-promo.md` end-to-end for the
exact scenario issue #4954 describes: a proposal that fails promotion criteria
and is **never revised** across subsequent Champion passes.

## The interaction bug

The idempotency check (run first, before Step 4's rejection/escalation logic)
computes `VERDICT_MARKER` from `sha256(title + body)` and skips **silently**
— no comment, no claim, no evaluation — whenever a prior comment already
contains that exact marker:

```bash
if echo "$ISSUE_JSON" | jq -e --arg m "$VERDICT_MARKER" \
     '.comments[] | select(.body | contains($m))' >/dev/null; then
  # skip — no comment, no claim, no re-evaluation
fi
```

For a genuinely unrevised proposal, `title`/`body` never change, so the hash
never changes, so the marker posted on the **first** rejection matches on
**every subsequent pass forever**. Step 4's `PRIOR_REJECTIONS >= 2` escalation
check (which routes to `loom:operator-only` for a human) is only reached when
the idempotency check does *not* skip — i.e. only when the proposal's hash has
actually changed since the last verdict. So:

- Cycle 1 (hash H1): no marker yet → evaluate → reject → post comment with
  `VERDICT_MARKER` for H1. `PRIOR_REJECTIONS` becomes 1.
- Cycle 2+ (still H1, unrevised): idempotency check matches the H1 marker →
  **skip silently, forever**. Step 4 (and its `PRIOR_REJECTIONS` counter) is
  never reached again for this proposal unless someone actually edits it.

Net effect: a proposal that fails once and is never revised gets **exactly
one** rejection comment, ever, and then sits invisible to any human —
`loom:operator-only` escalation, the whole point of the N=2 mechanism, never
fires for the precise scenario in #4954's repro steps (an *unrevised*
proposal re-entering the queue unchanged).

This is a regression relative to the pre-#4954 behavior: before, an unrevised
proposal was noisy (up to 6 duplicate comments) but self-corrected by
escalating once enough identical rejections piled up. After #4966, it's
silent but never escalates — trading "spam then a human eventually sees it"
for "quiet, but no human ever sees it."

## Why this wasn't caught in review

PR #4966 went through two review rounds:
1. First review correctly caught that the *original* `updatedAt`-keyed marker
   was self-invalidating (never matched), and noted — correctly, for that
   version of the code — that escalation still worked *because* the broken
   idempotency check never actually skipped, so `PRIOR_REJECTIONS` kept
   accumulating toward 2.
2. The fix (commit `54635d9a`) switched to the body-hash marker, which
   correctly resolves the "silent skip" requirement (AC1) — but this a
   second-order side effect: once idempotency *actually* skips (as intended),
   Step 4 is reached far less often, so `PRIOR_REJECTIONS` can no longer
   accumulate for an unrevised proposal. The re-review confirmed Step 4's code
   was "byte-identical before/after" and concluded there was "no regression
   to N=2 escalation" — true of the code text, but the *reachability* of that
   code changed as a direct consequence of fixing AC1.

## Suggested fix

Decouple "has this exact rejection already been commented" from "should we
still eventually escalate an unrevised-but-repeatedly-queued proposal."
Options:
- Track rejection count independently of the body-hash marker (e.g. a
  separate counter/marker that persists across identical-hash skips, or count
  *evaluation cycles* rather than *posted comments*), so an unrevised proposal
  still accumulates toward the N=2 threshold even while comments are
  suppressed.
- Or: make the idempotency skip itself escalate once a proposal has been
  silently skipped some number of times (i.e., silence isn't free forever —
  cap the silent-skip streak the same way `judge.md`'s `LOOM_MAX_STANDDOWN_STREAK`
  bounds silent stand-downs).

## Severity assessment

Not urgent — the new behavior is strictly less noisy than before, and an
unrevised, permanently-rejected proposal doing nothing is arguably tolerable.
But it does mean #4954's acceptance criteria ("after N=2 identical verdicts on
an unchanged issue, escalate") is not actually met for the scenario that
motivated the issue, and stuck proposals will now silently rot rather than
surfacing to a human. Worth a deliberate decision (fix vs. accept as
documented behavior) rather than leaving it undiscovered.

Discovered while judging PR #4966.

---

## Curator Enhancement

Verified against `origin/main`: `champion-issue-promo.md`'s idempotency check
(§"Concurrency Guard and Idempotency", ~lines 222-249) does run before, and
independently of, Step 4's `PRIOR_REJECTIONS` computation (~lines 379-390) —
confirming the issue's reachability analysis is accurate. No related open work
found (`check-duplicate.sh --issue 4967` returned clean).

## Acceptance Criteria

- [ ] A genuinely unrevised proposal (unchanged title+body across evaluation cycles) still reaches `loom:operator-only` escalation after a bounded number of evaluation cycles — it must not be silently skipped forever.
- [ ] The rejection/escalation counter's source of truth does not require posting a fresh "NEEDS REVISION" comment on every cycle — it must be computable (or persisted) even when the idempotency check skips silently.
- [ ] The existing idempotency guarantee from #4954/#4966 (an unrevised proposal gets no duplicate "NEEDS REVISION" comments) is preserved — the fix must not reintroduce comment spam.
- [ ] `champion-issue-promo.md`'s "Concurrency Guard and Idempotency" section documents the interaction between the idempotency skip and the N=2 escalation threshold, so a future edit to either mechanism doesn't silently reopen this gap again (this is the third review pass to touch this exact interaction — #4954 → #4966 → this issue).
- [ ] Manually trace the "Cycle 1 (hash H1) → Cycle 2+ (still H1)" scenario from the issue body against the fixed logic and confirm escalation now fires by some documented cycle N.

## Affected Files

- `defaults/.claude/commands/loom/champion-issue-promo.md` (canonical source) - lines ~222-249 (Idempotency check) and ~379-433 (Step 4 rejection/escalation, `PRIOR_REJECTIONS` counter) — decouple the per-cycle rejection-comment marker from escalation eligibility, per one of the two options suggested above.
- `.claude/commands/loom/champion-issue-promo.md` - currently a byte-identical mirror of the file above (not a symlink — verify with `diff` before editing). Check how recent PRs touching this file handled the mirror (edit both directly in the same PR, vs. relying on `./.loom/scripts/resync-installed.sh`) and follow that convention.

## Test Plan

- [ ] Manual verification: this is a markdown role-instructions file with no automated test harness (Champion's evaluation logic runs as LLM-interpreted bash snippets, not compiled/tested code). Verify by tracing the documented "Cycle 1 / Cycle 2+ / Cycle 3+" scenario against the new logic step by step and confirming escalation fires within a documented bound.
- [ ] Automated tests: N/A — no test suite covers this role-instructions file.
- [ ] Edge cases: (a) a proposal that IS revised between rejections still resets its counter correctly (existing behavior must not regress); (b) a proposal already routed to `loom:operator-only` is not double-escalated; (c) the fix does not reintroduce the pre-#4966 duplicate-comment spam for an unrevised proposal.

rjwalters:
## Summary

Champion's proposal-evaluation loop had two coupled mechanisms that had silently
become mutually exclusive. The body-hash idempotency check (#4966) runs **before**
Step 4, and once it started actually skipping, Step 4 became unreachable for a
proposal whose title+body never change: the first rejection's `VERDICT_MARKER`
matched on every later pass forever, so `PRIOR_REJECTIONS` — which counts only
*posted* "NEEDS REVISION" comments — froze at 1 and the N=2 escalation to
`loom:operator-only` could never fire. A stuck, unrevised proposal got exactly one
comment ever and then rotted invisibly.

This PR keeps the silence and restores the escalation by gating escalation on
**evaluation cycles** rather than **posted comments**.

## Changes

`defaults/.claude/commands/loom/champion-issue-promo.md` (canonical source only —
same convention as PR #4966, which also touched `defaults/` only; the installed
`.claude/commands/loom/` mirror is refreshed by the periodic `chore: resync
installed Loom surfaces` commit from the main checkout, never by a Builder):

- **Idempotency check is now three-way** (evaluate / silent-skip / escalate) and
  computes `PRIOR_REJECTIONS`, `SKIP_STREAK`, `ALREADY_ROUTED` up front, so the
  skip path can decide "escalate instead of skipping again" without ever reaching
  Step 4.
- **Each silent skip increments a durable tally** —
  `<!-- champion:unrevised-skips:<body-hash>:<n> -->` — carried *inside the
  existing verdict comment* and updated with
  `gh api --method PATCH .../issues/comments/<id>`. An edit posts no new comment
  and sends no notification, so the "1 comment, then silence" guarantee is intact
  while the counter still advances. Legacy verdict comments predating the tally
  get it appended.
- **`UNREVISED_EVALS = PRIOR_REJECTIONS + SKIP_STREAK`** is the single escalation
  gate, used identically by the skip path and Step 4, capped by
  `LOOM_MAX_UNREVISED_EVALUATIONS` (default **2** — the same N=2 threshold #4954
  specified, now measured in cycles).
- **Escalation goes through the claim**: hitting the cap sets
  `ESCALATE_UNREVISED=yes` and falls through to the Claim step + verdict-time
  recheck before Step 4's escalation branch (Steps 1–3 are skipped — the text is
  byte-identical, so re-evaluating cannot change the verdict; findings are lifted
  from the prior comment). Two concurrent passes therefore cannot post two
  escalation comments; a lost `PATCH` can only *under*count.
- **`ALREADY_ROUTED=yes` short-circuits everything** — a proposal already carrying
  `loom:operator-only` is neither re-tallied nor re-escalated.
- **Step 4's reject branch seeds the tally at `:0`** next to `VERDICT_MARKER`;
  both markers ship together or neither mechanism works.
- **New "Bounding the silent skip: how idempotency interacts with N=2 escalation
  (#4967)" subsection** documents the coupling, a mechanism/counter table, the
  full cycle trace, and the invariants a future edit must preserve. The batch
  loop's "Per-issue order" note is now a four-row outcome table.

- [x] **An unrevised proposal still reaches `loom:operator-only` after a bounded
  number of cycles.** `SKIP_STREAK` advances on every silent skip, so
  `UNREVISED_EVALS` reaches the cap and `ESCALATE_UNREVISED=yes` fires on cycle 3
  (trace below). Not skippable forever.
- [x] **The counter's source of truth does not require a fresh comment each
  cycle.** The tally is an *edit* of the comment that already exists; it is
  readable (`sed` on the matched comment body) and writable (`PATCH`) on a cycle
  that posts nothing.
- [x] **The #4954/#4966 idempotency guarantee is preserved.** Cycle 2 posts
  nothing — comment budget for an unrevised proposal is exactly 2 (one NEEDS
  REVISION, one escalation) vs. 6+ before #4954. The skip path may only edit,
  never post; this is written into the invariants list.
- [x] **The "Concurrency Guard and Idempotency" section documents the
  interaction.** New subsection plus a pointer at the top of the section: "The
  idempotency skip and the escalation threshold are one mechanism, not two."
- [x] **Cycle 1 (H1) → Cycle 2+ (still H1) traced against the fixed logic**, with
  a documented escalation cycle N=3 (table below, also in the file).

## Test Plan

No automated harness exists for this file (LLM-interpreted bash in a
role-instructions markdown file); verification is by trace + inspection, per the
issue's own Test Plan.

**Manual trace — unrevised proposal at body hash H1:**

| Cycle | Marker match? | `PRIOR_REJECTIONS` | `SKIP_STREAK` | `UNREVISED_EVALS` | Outcome | Comments posted |
|---|---|---|---|---|---|---|
| 1 | no (H1 unseen) | 0 | 0 | 0 | evaluate → reject → post NEEDS REVISION with `VERDICT_MARKER` + `unrevised-skips:H1:0` | 1 |
| 2 | yes (H1) | 1 | 0 | 1 < 2 | silent skip; `PATCH` tally to `1` | 0 |
| 3 | yes (H1) | 1 | 1 | 2 ≥ 2 | `ESCALATE_UNREVISED=yes` → claim → Step 4 escalation → `loom:operator-only` | 1 |
| 4+ | — | — | — | — | `loom:operator-only` excludes it from every future pass | 0 |

**Escalation fires on cycle 3** — the same cycle the pre-#4954 behavior escalated
on, with 2 total comments instead of 6+.

**Edge cases from the issue's Test Plan:**

- (a) *Revised between rejections* — a new hash produces a new marker, so
  `SKIP_STREAK` resets to 0 for the new revision while `PRIOR_REJECTIONS` keeps
  accumulating across revisions. Reject H1 (cycle 1) → revise → reject H2
  (cycle 2, `PRIOR_REJECTIONS=1 < 2`) → cycle 3 sees `PRIOR_REJECTIONS=2` and
  escalates. Identical to pre-existing behavior; no regression.
- (b) *Already routed* — `ALREADY_ROUTED=yes` takes its own branch: skip with no
  tally and no escalation. "When NOT to Promote" already excludes such issues from
  future passes, so no double-escalation.
- (c) *No duplicate-comment spam* — cycles 2..N-1 post nothing at all; the only
  write is a comment edit.

**Shell idioms verified on the target platform** (macOS/BSD `sed`, `jq`) with a
scratch harness before committing: tally read via
`sed -n "s|.*<!-- champion:unrevised-skips:$BODY_HASH:\([0-9]\{1,\}\) -->.*|\1|p"`
round-trips 0 → 1 → 2; the append fallback fires for a comment lacking the tally;
`gh api --paginate --jq '.[] | select(...)' | jq -s 'last'` yields `null` (not a
parse error) when no comment matches — the reason for the `// empty` / `// ""`
guards and the REST endpoint (the `id` from `gh issue view --json comments` is a
GraphQL node id and cannot be PATCHed).

**Repo checks:** `defaults/scripts/check-phantom-labels.sh` and
`check-labels-drift.sh` both pass (no new labels introduced — only the existing
`loom:operator-only` / `loom:evaluating`).

Closes #4967

### DIFF
--- a/defaults/.claude/commands/loom/champion-issue-promo.md
+++ b/defaults/.claude/commands/loom/champion-issue-promo.md
@@ -219,9 +219,11 @@ Use conservative judgment. **Do NOT promote** if:
 
 **Applies to every proposal evaluated by this file** — `loom:curated`, `loom:architect`, `loom:hermit`, and `loom:auditor` alike — not just the `loom:architect` case that surfaced it.
 
-### Idempotency check (run BEFORE claiming — skip without commenting on a match)
+**The idempotency skip and the escalation threshold are one mechanism, not two** (#4967). Suppressing duplicate comments must never suppress the escalation that eventually puts a stuck proposal in front of a human — read "Bounding the silent skip" below before changing either half.
 
-Compute a marker keyed to a **hash of the proposal's own text** (title + body), so a genuine revision always gets a fresh evaluation while an unchanged proposal never gets re-commented:
+### Idempotency check (run BEFORE claiming — skip silently on a match, until the skips are capped)
+
+Compute a marker keyed to a **hash of the proposal's own text** (title + body), so a genuine revision always gets a fresh evaluation while an unchanged proposal never gets re-commented. The check is **three-way**, not two-way: no match → evaluate; match with skips left in the budget → skip silently; match with the skip budget exhausted → **escalate** (see "Bounding the silent skip" below).
 
 ```bash
 ISSUE_NUMBER=<number>
@@ -242,14 +244,60 @@ BODY_HASH=$(printf '%s\n%s' \
   | _sha256 | awk '{print substr($1, 1, 16)}')
 VERDICT_MARKER="<!-- champion:proposal-verdict:body-$BODY_HASH -->"
 
+# Escalation inputs, computed HERE rather than in Step 4 (#4967): the skip path
+# below must be able to decide "escalate instead of skipping again" without ever
+# reaching Step 4. Step 4 reuses these same variables.
+PRIOR_REJECTIONS=$(echo "$ISSUE_JSON" | jq \
+  '[.comments[] | select(.body | contains("Champion Review: NEEDS REVISION"))] | length')
+ALREADY_ROUTED=$(echo "$ISSUE_JSON" | jq -e '.labels[] | select(.name=="loom:operator-only")' >/dev/null && echo yes || echo no)
+SKIP_STREAK=0            # silent skips already recorded for THIS body revision
+ESCALATE_UNREVISED=no    # set to yes to bypass re-evaluation and go straight to Step 4's escalation
+
 if echo "$ISSUE_JSON" | jq -e --arg m "$VERDICT_MARKER" \
      '.comments[] | select(.body | contains($m))' >/dev/null; then
-  echo "Already evaluated #$ISSUE_NUMBER at body revision $BODY_HASH — skipping (no comment, no claim)"
-  # Continue the batch to the next issue; do not read further or claim.
+  # This exact revision was already evaluated. Read the silent-skip tally carried
+  # by the matching verdict comment. REST, not `gh issue view`: only the REST
+  # payload has the numeric comment id that the PATCH below needs (the `id` from
+  # `gh issue view --json comments` is a GraphQL node id and cannot be PATCHed).
+  VERDICT_COMMENT=$(gh api "repos/{owner}/{repo}/issues/$ISSUE_NUMBER/comments" --paginate \
+    --jq ".[] | select(.body | contains(\"$VERDICT_MARKER\"))" | jq -s 'last')
+  COMMENT_ID=$(echo "$VERDICT_COMMENT" | jq -r '.id // empty')
+  COMMENT_BODY=$(echo "$VERDICT_COMMENT" | jq -r '.body // ""')
+  SKIP_STREAK=$(printf '%s' "$COMMENT_BODY" \
+    | sed -n "s|.*<!-- champion:unrevised-skips:$BODY_HASH:\([0-9]\{1,\}\) -->.*|\1|p" | tail -n 1)
+  SKIP_STREAK=${SKIP_STREAK:-0}
+  UNREVISED_EVALS=$(( PRIOR_REJECTIONS + SKIP_STREAK ))
+
+  if [ "$ALREADY_ROUTED" = "yes" ]; then
+    # Terminal state — a human owns this now. Skip without tallying or escalating.
+    echo "#$ISSUE_NUMBER already routed to loom:operator-only — skipping (no comment, no claim, no tally)"
+  elif [ "$UNREVISED_EVALS" -ge "${LOOM_MAX_UNREVISED_EVALUATIONS:-2}" ]; then
+    # Silence is not free forever: the skip budget is spent, so this pass does
+    # NOT skip. Fall through to Claim, then jump straight to Step 4's escalation
+    # branch (no re-evaluation — the text is unchanged, so the verdict is too).
+    ESCALATE_UNREVISED=yes
+    echo "#$ISSUE_NUMBER unrevised at $BODY_HASH across $UNREVISED_EVALS evaluations — escalating to the operator instead of skipping again"
+  else
+    # Record this cycle's skip IN PLACE by PATCHing the existing verdict comment.
+    # An edit posts no new comment and sends no notification, so the "1 comment,
+    # then silence" guarantee holds while the counter still advances.
+    NEXT_SKIPS=$(( SKIP_STREAK + 1 ))
+    if printf '%s' "$COMMENT_BODY" | grep -q "<!-- champion:unrevised-skips:$BODY_HASH:"; then
+      NEW_BODY=$(printf '%s' "$COMMENT_BODY" \
+        | sed "s|<!-- champion:unrevised-skips:$BODY_HASH:[0-9]\{1,\} -->|<!-- champion:unrevised-skips:$BODY_HASH:$NEXT_SKIPS -->|")
+    else
+      # Verdict comment predates this tally (posted before #4967) — append it.
+      NEW_BODY=$(printf '%s\n\n%s' "$COMMENT_BODY" "<!-- champion:unrevised-skips:$BODY_HASH:$NEXT_SKIPS -->")
+    fi
+    [ -n "$COMMENT_ID" ] && gh api --method PATCH \
+      "repos/{owner}/{repo}/issues/comments/$COMMENT_ID" -f body="$NEW_BODY" >/dev/null
+    echo "Already evaluated #$ISSUE_NUMBER at body revision $BODY_HASH — skipping silently (skip $NEXT_SKIPS recorded; unrevised evaluations now $(( PRIOR_REJECTIONS + NEXT_SKIPS ))/${LOOM_MAX_UNREVISED_EVALUATIONS:-2}, escalates once it reaches the cap; no comment, no claim)"
+    # Continue the batch to the next issue; do not read further or claim.
+  fi
 fi
 ```
 
-If the marker is present, **stop here for this issue** — do not read comments further, do not claim, do not comment. This is the mechanism that turns "6 identical NEEDS REVISION comments" into "1 comment, then silent skips" for a truly unrevised proposal.
+If the marker is present **and `ESCALATE_UNREVISED=no`**, **stop here for this issue** — do not read comments further, do not claim, do not comment. This is the mechanism that turns "6 identical NEEDS REVISION comments" into "1 comment, then silent skips" for a truly unrevised proposal. If `ESCALATE_UNREVISED=yes`, do **not** stop: continue to the Claim step and then to Step 4, which escalates on that flag without re-running the 8 criteria.
 
 #### Why a body hash and NOT the issue's `updatedAt` (#4966)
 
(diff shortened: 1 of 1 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-02T21:55:47Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verified against actual code** (checked out the branch head `989725f1` in the existing builder worktree `.loom/worktrees/issue-4967`, since my main checkout's cached copy of `main` predated PR #4966's merge):

**Logic trace matches the claimed cycle-by-cycle behavior.** Walked the diff by hand for the unrevised-proposal scenario:
- Cycle 1: no marker match → evaluate → reject → post `NEEDS REVISION` with `VERDICT_MARKER` + `<!-- champion:unrevised-skips:H1:0 -->` seeded alongside it.
- Cycle 2: marker matches; `PRIOR_REJECTIONS=1`, `SKIP_STREAK` read from the tally=`0`, `UNREVISED_EVALS=1 < 2` → silent skip, `PATCH` the tally to `1`. Zero new comments/notifications (an edit, not a post).
- Cycle 3: marker matches again; `PRIOR_REJECTIONS=1` (unchanged — no new "NEEDS REVISION" comment was ever posted), `SKIP_STREAK=1` (from the PATCH) → `UNREVISED_EVALS=2 >= 2` → `ESCALATE_UNREVISED=yes` → falls through to Claim → Step 4's escalation branch → `loom:operator-only`, skipping re-evaluation of the 8 criteria (correctly, since the text is byte-identical).

This confirms the PR's own trace table: escalation fires on cycle 3, comment budget is exactly 2 (one rejection + one escalation) vs. 6+ pre-#4954, and the #4966 "no duplicate comment" guarantee holds throughout (the only per-skip write is a `PATCH` on the existing comment, not `gh issue comment`).

**Correctness details checked:**
- `PRIOR_REJECTIONS`/`ALREADY_ROUTED`/`SKIP_STREAK=0`/`ESCALATE_UNREVISED=no` are computed unconditionally *before* the marker-match `if`, so Step 4 gets correct values on both the fresh-evaluation path (no match) and the revised-proposal path (new hash → new marker, `SKIP_STREAK` correctly resets to 0 while `PRIOR_REJECTIONS` keeps accumulating across revisions — edge case (a) from the issue's test plan).
- `ALREADY_ROUTED=yes` short-circuits before any tally/escalation, matching edge case (b) (no double-escalation of an already-routed proposal).
- The REST-vs-GraphQL comment-id distinction (`gh api .../issues/comments/<id>` PATCH vs. the non-PATCHable node id from `gh issue view --json comments`) is correct and well-commented.
- `VERDICT_MARKER`/tally-marker interpolation into the `--jq` filter is safe — `BODY_HASH` is a hex sha256 digest, not raw untrusted issue text, so there's no injection surface there.
- Race behavior is honestly documented: two concurrent passes reading the same `SKIP_STREAK` and both `PATCH`ing can only *lose* an increment (delaying escalation by a cycle), never double-escalate, since escalation itself is gated by the Claim + verdict-time recheck. Reasonable tradeoff for a role-instructions file with no real CAS primitive available.
- Scope: only `defaults/.claude/commands/loom/champion-issue-promo.md` changed. Confirmed `.claude/commands/loom` is a symlink into `defaults/.claude/commands/loom` in this repo (`.gitignore` line 118-136, `link_dogfood_commands`), so there is no separate installed-mirror file to edit here — editing `defaults/` only is correct and sufficient, consistent with #4966's precedent.

**CI / merge state:** all CI checks pass (`Shell Test Suites (hermetic)` was still running when I first checked — waited for it, it passed at 4m34s). `mergeStateStatus` is `CLEAN`.

Nothing blocking. One non-blocking observation for a future pass, not this one: the PATCH is fire-and-forget (`>/dev/null`, no failure check) — a failed PATCH just delays escalation by a cycle (self-healing, since the next cycle re-reads the un-updated tally), so this is fine as documented, not a bug.

Nice work tracing this all the way through and documenting the invariants for the next person who touches this section.

PR #4963, opened 2026-08-02
feat(dashboard): auto-deploy on merge + serve a build-commit stamp

rjwalters:
## Summary

Adds a GitHub Actions deploy pipeline for the fleet observability dashboard so a merged `dashboard/**` fix reaches `dashboard.2amlogic.com` automatically, and stamps the live Worker with the commit it was deployed from so drift is observable at a glance.

## Changes

- **`.github/workflows/dashboard-deploy.yml`** (new): on push to `main` touching `dashboard/**`, a `test` job runs the full backend + web suite (`npm run check:all`), then a `deploy` job (gated on `needs: test`) applies D1 migrations (`wrangler d1 migrations apply`, idempotent by design) and runs `wrangler deploy` against the 2AM reference instance. Every step fails loudly (`set -euo pipefail`, explicit `::error::` annotations, no swallowed exit codes) — there is no silent-failure path.
  - The instance-specific `wrangler.2amlogic.toml` overlay (Worker name, D1 database id/name, custom-domain route, Cloudflare Access `[vars]`) is deliberately never committed to this repo (documented in `docs/reference-deployment.md` §3) and I don't have access to those production values (some, like the Access AUD tags, aren't even fully recorded in the docs — only truncated). So the workflow materializes it from a **new** GitHub Actions secret, `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC` (the full file contents an operator already keeps locally), rather than hardcoding or reconstructing any account/route/database values. **This secret still needs to be provisioned by the operator** — until then, the deploy job's first step fails with a clear `::error::` naming exactly what's missing and how to provision it (`gh secret set CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC < dashboard/wrangler.2amlogic.toml`, run from the operator's machine). `CLOUDFLARE_API_TOKEN`/`CLOUDFLARE_ACCOUNT_ID` were already provisioned (confirmed in the issue's comment thread).
  - A `preflight` step (`npm run preflight -- --remote`, the repo's existing `scripts/check-deploy-config.sh`) runs `wrangler deploy --dry-run` before touching migrations or the real deploy, catching a malformed/placeholder overlay before it can reach production.
[...]

PR #4977, opened 2026-08-03
ci(dashboard-deploy): failed deploys surface as a forge issue; judge checklist covers new secrets

### LINKED ISSUE #4974: ci(dashboard-deploy): failed deploys are silent + new-secret requirements surfaced only post-merge
## Context

The first three runs of the new `dashboard-deploy.yml` workflow (#4963, from issue #4958) all failed with `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC secret is not set`, and the live Worker silently stayed on pre-fix code — the operator discovered it only by noticing the dashboard still showed stale data (the #4955 39-vs-4 phantom count) a day after the fix merged.

Two compounding problems:

1. **Secret contract drift**: issue #4958 and its provisioning comment specified `CLOUDFLARE_API_TOKEN` + `CLOUDFLARE_ACCOUNT_ID` (which were provisioned ahead of the build, 2026-08-02). The Builder's workflow additionally invented `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC` (the uncommitted `wrangler.2amlogic.toml` overlay materialized from a secret). The design is sound — the overlay is deliberately never committed — but the new secret requirement surfaced only as a post-merge runtime failure, not at review time. The judge/PR template could ask: "does this change require new repo secrets/vars? Are they provisioned or explicitly flagged to the operator before merge?"
2. **Silent deploy failure**: a failed `Dashboard Deploy` run notifies no one. Unlike CI on a PR (blocking, visible), a push-triggered deploy failure lands in the Actions tab and nowhere else. The workflow should page: on failure, file/update a tracking issue (e.g. `gh issue create`/comment via `GITHUB_TOKEN`, labeled `loom:triage`) or at minimum a commit comment, so the pipeline itself notices its deploy arm is dead — same "silence never reads as success" principle as `fleet status` (#4342) and the watch skill's escalation contract.

(Resolution note for the record: the missing secret was provisioned from the operator host on 2026-08-02 — `gh secret set CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC < dashboard/wrangler.2amlogic.toml` — and run 30767457904 rerun to green; dashboard.2amlogic.com serves `/api/version` = de85f850.)

## Acceptance criteria

- [ ] A failed `dashboard-deploy.yml` run produces a forge-visible artifact (issue or comment) that enters the normal triage flow, not just a red run in the Actions tab.
- [ ] The failure artifact names the failing step and, for the missing-secret case, echoes the exact provisioning command (the workflow's error text already does this well — reuse it).
- [ ] Reviewer guidance (judge role prompt or PR checklist) covers "new secrets/vars introduced by this PR must be provisioned or flagged before merge".

## Implementation Guidance

**AC1/AC2 (failure notification):** `.github/workflows/dashboard-deploy.yml`'s `deploy` job needs a final step, gated `if: failure()`, that runs on any prior step's failure (the "Materialize the 2AM instance's wrangler config" step already `exit 1`s with a detailed `::error::` message — reuse that text verbatim rather than re-deriving it). `.github/workflows/fork-drift.yml` (lines ~90-141) is the closest existing precedent in this repo for "workflow files/updates a forge issue on a detected problem": it searches for an existing open issue by exact title (`gh issue list --state open --search "\"$title\" in:title" --json number,title --jq ...`), comments on it if found, otherwise creates a new one with `gh issue create --title "$title" --label "loom:triage" --body "$body"`. Follow the same dedupe-by-title shape here so repeated deploy failures on the same underlying cause don't spam a fresh issue per run — the failure step needs `permissions: issues: write` (or the workflow-level equivalent) since it's a `push`-triggered workflow, not a PR.

**AC3 (reviewer guidance):** the target is `.loom/roles/judge.md`, which is a symlink to the canonical `defaults/.claude/commands/loom/judge.md` (cite the canonical path in any PR). The "Evaluation Focus Areas" section's per-category checklists (`### Correctness`, `### Design`, etc., around line 1575) is the natural insertion point — add a bullet (or a new subsection, e.g. `### Infrastructure`) asking whether the PR introduces new repo secrets/vars, and whether they're provisioned or explicitly flagged to the operator before merge. There is no `.github/pull_request_template.md` in this repo, so the "PR checklist" half of the AC as originally phrased doesn't apply — scope this AC to the judge role prompt only.

## Affected Files

- `.github/workflows/dashboard-deploy.yml` - add an `if: failure()` step to the `deploy` job that files/updates a `loom:triage` issue, reusing `fork-drift.yml`'s dedupe-by-title pattern and the existing `::error::` text from the "Materialize the 2AM instance's wrangler config" step
- `.github/workflows/fork-drift.yml` - reference only, the pattern this AC's failure step should follow (lines ~90-141: search-by-title, comment-if-found else create)
- `defaults/.claude/commands/loom/judge.md` - add "new secrets/vars introduced by this PR must be provisioned or flagged before merge" to the Evaluation Focus Areas checklist (`.loom/roles/judge.md` is a symlink to this file; editing the canonical path is sufficient)

## Test Plan

- [ ] Manual verification: trigger a synthetic failure (e.g. temporarily unset/rename the secret reference in a throwaway branch, or use `workflow_dispatch` with a forced-fail path if added) and confirm a `loom:triage` issue is created with the failing step name and the exact provisioning command; re-run the failure and confirm the *same* issue is commented on rather than a duplicate being created
- [ ] Automated tests: N/A for the workflow YAML itself (GitHub Actions workflows aren't unit-testable in this repo's suite) — if the dedupe/failure logic is extracted into a shell function, add a script-level test following the pattern of other `.loom/scripts/*.sh` tests
- [ ] Edge cases: confirm the failure step does not itself fail the job a second time (e.g. `gh issue create` erroring should not mask the original failure's exit code); confirm a green re-run after the underlying issue is fixed does not need manual cleanup of the tracking issue (leave it to be closed manually or note in the comment that it self-resolved)

rjwalters:
Closes #4974

## Summary

`.github/workflows/dashboard-deploy.yml` (landed via #4963/#4958) had no failure notification — a failed deploy run was a red run in the Actions tab and nothing else. Separately, the judge role prompt had no explicit guidance for reviewing PRs that introduce new required repo secrets/vars.

### `.github/workflows/dashboard-deploy.yml`

- Added an `id:` to every step in the `deploy` job.
- Added a final step, `File or update a tracking issue for the failed deploy`, gated `if: failure()`, that:
  - Uses `toJSON(steps)` to find which step's `conclusion` is `"failure"`, and maps that step id to a human-readable name via a `case` statement.
  - Builds a forge-visible issue title `ci(dashboard-deploy): deploy failed at "<step name>"`.
  - Follows `fork-drift.yml`'s dedupe pattern: search for an existing open issue with that exact title (`gh issue list --state open --search "\"$title\" in:title" --json number,title --jq ...`); comment on it if found, otherwise `gh issue create --label "loom:triage"`.
  - For the missing-secret case specifically, includes the exact `::error::`/detail text from the "Materialize the 2AM instance's wrangler config" step. Rather than re-deriving or re-parsing logs, that step now *also* writes the identical text (verbatim, same lines) to a step output (`error_detail`, via the standard multiline `GITHUB_OUTPUT` heredoc pattern) before its `exit 1`; the failure step reads it via `steps.materialize-wrangler-config.outputs.error_detail`.
- Added a job-level `permissions: { contents: read, issues: write }` block to `deploy` (there was none before) — required for `gh issue create`/`gh issue comment` to work, since this workflow is `push`-triggered (not a PR check), so no default PR-scoped token permissions apply.

### `defaults/.claude/commands/loom/judge.md`

Added a new `### Infrastructure` subsection under "Evaluation Focus Areas" (alongside `### Correctness`, `### Design`, etc.):
- Asks whether the PR introduces a new required repo secret/var, and whether it's already provisioned or explicitly flagged to the operator before merge.
- Notes that an unflagged new-secret requirement should be treated as blocking, citing this issue (#4974) as the motivating example.
- Allows an exception when the workflow already fails loudly with a forge-visible tracking artifact (like the new step this PR adds) rather than deploying silently.

`.loom/roles/judge.md` is a symlink to `defaults/.claude/commands/loom/judge.md`, so no separate edit was needed there. `.claude/commands/loom/judge.md` (the installed copy the symlink actually resolves to on disk) is gitignored/generated by the install/resync scripts, so it's not touched by this PR either.

## How each acceptance criterion is met

1. **A failed run produces a forge-visible artifact that enters normal triage** — the new `if: failure()` step files/updates a `loom:triage`-labeled issue, which is exactly the normal issue-triage entry point.
2. **The artifact names the failing step and echoes the exact provisioning command for the missing-secret case** — the issue body includes a `**Failing step:**` line, and for the wrangler-config-secret case, a fenced code block with the verbatim text (including the `gh secret set ...` provisioning command) sourced from the existing step output, not re-derived.
3. **Judge role prompt covers new-secret/var review** — new `### Infrastructure` subsection added.

## Test plan

Full CI automation for GitHub Actions YAML isn't practical/expected in this repo (no workflow test harness); this was validated as follows:

- `actionlint .github/workflows/dashboard-deploy.yml` — clean, no findings.
- `actionlint` (whole repo) — clean, no regressions to any other workflow.
- YAML parses cleanly via `python3 -c "import yaml; yaml.safe_load(...)"`.
- Extracted the new failure step's shell logic into a standalone script and ran `shellcheck` against it — clean.
- Ran that extracted script with synthetic `STEPS_JSON`/`MATERIALIZE_ERROR_DETAIL`/`GITHUB_*` env vars simulating a failure at the `materialize-wrangler-config` step; confirmed the generated title and body correctly name the failing step and include the verbatim error detail block.
- `scripts/check-docs-defaults-parity.sh` — OK, no orphaned/broken doc links introduced.

**Manual verification path for an operator** (not run as part of this PR, documented per the issue's test-plan expectations):
1. Temporarily unset/rename the `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC` secret (or push a change that breaks another deploy step) and push to `main` under `dashboard/**`.
2. Confirm the run fails and a new `loom:triage` issue titled `ci(dashboard-deploy): deploy failed at "Materialize the 2AM instance's wrangler config"` (or the relevant step name) is created, containing the failing step, run link, commit, and (for the secret case) the exact `gh secret set CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC < dashboard/wrangler.2amlogic.toml` provisioning command.
3. Re-trigger the same failure (push again without fixing it) and confirm the **same** issue receives a new comment rather than a duplicate issue being created.
4. Fix the underlying cause and confirm a green re-run does **not** automatically close/clean up the tracking issue — this is expected/acceptable per the issue's own test-plan note; the tracking issue is left for the normal triage flow to close.
- Also confirmed by inspection: the `if: failure()` step only runs when a prior step already failed, and its own success/failure afterward cannot change the job's already-fixed `failure` conclusion — so it cannot mask the original failure's exit code.

## Deviations from the curated plan

None of substance — the actual current `dashboard-deploy.yml` (post #4963/#4958, 11 commits ahead as flagged in the issue) matched the description closely enough that no rescoping was needed. One addition beyond the literal plan: rather than only reading the "Materialize..." step's existing stderr text, that step now also mirrors the identical text into a step output so the failure step can read it structurally (via `steps.<id>.outputs.<name>`) instead of trying to scrape run logs, which GitHub Actions doesn't expose to a later step natively.

### DIFF
--- a/.github/workflows/dashboard-deploy.yml
+++ b/.github/workflows/dashboard-deploy.yml
@@ -46,15 +46,26 @@ jobs:
     name: Deploy Worker (dashboard.2amlogic.com)
     needs: test
     runs-on: ubuntu-latest
+    # `issues: write` is required by the "File or update a tracking issue for
+    # the failed deploy" step below (#4974) — this workflow is `push`-triggered
+    # (not a PR check), so without this the failure-tracking step's `gh issue
+    # create`/`comment` would itself fail silently.
+    permissions:
+      contents: read
+      issues: write
     steps:
       - uses: actions/checkout@v7
+        id: checkout
       - uses: actions/setup-node@v7
+        id: setup-node
         with:
           node-version: "20"
       - name: Install backend dependencies
+        id: install-backend-deps
         run: npm ci
         working-directory: dashboard
       - name: Install web UI dependencies
+        id: install-web-deps
         run: npm ci
         working-directory: dashboard/web
 
@@ -70,6 +81,7 @@ jobs:
       # later change (e.g. epic #4859's production cutover) only requires
       # updating the secret, never this workflow.
       - name: Materialize the 2AM instance's wrangler config
+        id: materialize-wrangler-config
         working-directory: dashboard
         env:
           WRANGLER_CONFIG_2AMLOGIC: ${{ secrets.CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC }}
@@ -82,11 +94,25 @@ jobs:
             echo "dashboard/docs/reference-deployment.md section 3 for what it must contain, and" >&2
             echo "'gh secret set CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC < dashboard/wrangler.2amlogic.toml'" >&2
             echo "run from the operator's machine (where that file already lives) to provision it." >&2
+            # Mirror the same text into a step output (verbatim) so the
+            # failure-tracking step below (#4974) can surface it in the
+            # forge issue/comment it files, without re-deriving the message.
+            {
+              echo "error_detail<<GH_OUTPUT_EOF"
+              echo "CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC secret is not set — deploy cannot proceed."
+              echo "This must be the full contents of the 2AM instance's wrangler.2amlogic.toml overlay"
+              echo "(Worker name, D1 database name/id, custom-domain route, Access vars) — see"
+              echo "dashboard/docs/reference-deployment.md section 3 for what it must contain, and"
+              echo "'gh secret set CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC < dashboard/wrangler.2amlogic.toml'"
+              echo "run from the operator's machine (where that file already lives) to provision it."
+              echo "GH_OUTPUT_EOF"
+            } >> "$GITHUB_OUTPUT"
             exit 1
           fi
           printf '%s\n' "$WRANGLER_CONFIG_2AMLOGIC" > wrangler.2amlogic.toml
 
       - name: Build the dashboard UI
+        id: build-web
         working-directory: dashboard
         run: npm run build:web
 
@@ -95,6 +121,7 @@ jobs:
       # a bad overlay (still-a-placeholder database_id, config that fails to
       # parse) fail here instead of mid-migration.
       - name: Preflight the deploy config
+        id: preflight
         working-directory: dashboard
         env:
           LOOM_DASHBOARD_WRANGLER_CONFIG: wrangler.2amlogic.toml
@@ -107,6 +134,7 @@ jobs:
       # nothing new to apply is a clean no-op — this step does not need its
       # own "only if pending" guard.
       - name: Apply D1 migrations (idempotent)
+        id: apply-migrations
         working-directory: dashboard
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T02:17:23Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

Reviewed the diff directly (not just the description). Both files, no unrelated changes.

**`.github/workflows/dashboard-deploy.yml`**
- `actionlint` (fresh run) is clean on this file and the whole repo.
- `if: failure()` on the new tracking-issue step is correct: it only runs when a prior step in the job already failed, and the job's own `failure` conclusion is fixed by that prior step — this step succeeding or failing afterward cannot mask/flip it.
- `permissions: { contents: read, issues: write }` added at job level is required and correctly scoped — this is a `push`-triggered workflow (no default PR-scoped token perms), and `gh issue create`/`gh issue comment` need `issues: write`. `GH_TOKEN: ${{ github.token }}` on the new step is set correctly.
- The dedupe-by-title logic (`gh issue list --state open --search "\"$title\" in:title" --json number,title --jq ...`) is a verbatim match of `fork-drift.yml`'s existing pattern — same search shape, same exact-title `select`, same comment-vs-create branch.
- Step-output capture is sound: `materialize-wrangler-config` mirrors the identical `::error::`/detail text (character-for-character) into `error_detail` via the standard `GITHUB_OUTPUT` heredoc *before* its `exit 1`. GitHub Actions captures a step's `GITHUB_OUTPUT` writes regardless of the step's own exit code, so `steps.materialize-wrangler-config.outputs.error_detail` is populated correctly even though that step fails. The consuming step guards on `-n "${MATERIALIZE_ERROR_DETAIL:-}"` so other failing steps (no output set) don't emit a bogus empty code block.
- The `case "$failing_id"` mapping covers every step id added in this PR (checkout, setup-node, install-backend-deps, install-web-deps, materialize-wrangler-config, build-web, preflight, apply-migrations, deploy-worker) with an `unknown` fallback — verified 1:1 against the new `id:` fields.

**`defaults/.claude/commands/loom/judge.md`**
- New `### Infrastructure` subsection is well-placed among the existing `### Correctness`/`### Design`/.../`### Performance` Evaluation Focus Areas list (inserted right after `### Performance`, before `### Test Plan Execution`), doesn't duplicate any existing heading, and the parity/CLAUDE.md-budget CI checks are green.

**CI / merge state**
- All CI checks pass (`gh pr checks 4977`), including `Shell Test Suites (hermetic)` which I waited out to completion.
- `mergeStateStatus` is `CLEAN`.

No blockers found.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (and the PR already merged at 2026-08-03T02:20:52Z) — another Judge's approval raced mine. Discarding my review; not writing any label.

For the record, during my independent review I found what looks like a real, reproducible bug in the new "File or update a tracking issue for the failed deploy" step that the other review pass didn't catch: the constructed `$title` always wraps the failing step name in literal double quotes (e.g. `deploy failed at "Materialize the 2AM instance's wrangler config"`), and that string is then interpolated *unescaped* into the `--jq` filter argument (`--jq ".[] | select(.title == \"$title\") | .number"`). The embedded quotes break the jq program's own string-literal syntax, causing a jq compile error on every invocation of that line (reproduced locally with real `jq`). Combined with `set -euo pipefail`, this aborts the step before it ever reaches `gh issue create`/`gh issue comment` — meaning the tracking-issue-filing feature this PR adds does not actually work for any failure. Filing a follow-up issue with repro steps and a suggested fix rather than blocking here, since the verdict already landed.

PR #4982, opened 2026-08-03
feat(daemon): add `loom-daemon cancel` and persist the sweep pgid so termination reaps the whole process tree

### LINKED ISSUE #4980: daemon: no CLI sweep-cancel for remote operation, and sweep termination doesn't reap the process tree — hand-killed wrapper left a zombie agent that relaunched its workload
## Context (2026-08-03, loom-worker-1)

Operator needed to cancel a wedged sweep (gf180-trng #76, hung ngspice sims saturating the 8-core worker, host-distress breaker HALTED for hours). Two gaps turned a one-command cancel into a three-round manual process hunt:

1. **No remote cancel primitive.** `mcp__loom__cancel_sweep` exists only via the MCP server on the operator's own machine; `loom-daemon` has no `cancel` subcommand, so over ssh the only lever is `kill`. The watch playbook's "never hand-kill sweep children" rule exists precisely because of what happened next.
2. **Killing the tracked pid does not terminate the sweep.** The registry tracks the `claude-wrapper.sh` pid. SIGKILLing it left: the underlying `claude -p /loom:sweep 76` agent alive (which noticed its sims died and RELAUNCHED all four ngspice 5 minutes later, re-tripping the breaker), a second wrapper instance, a detached `sleep 60` watcher loop, and the ngspice processes themselves (already reparented to systemd from an earlier phase). The registry meanwhile showed `in_flight: 0` — a zombie agent invisible to the daemon, actively mutating a repo it no longer had a claim on (the crash-path restore had already returned #76 to `loom:issue`).

Full sequence in the operator log; the sims alone burned ~35 CPU-hours across the two rounds.

## Suggested fix

- **`loom-daemon cancel <sweep-id|--issue N>`**: CLI parity with `cancel_sweep` over the same IPC request, so remote/ssh operation has the sanctioned path (same family as `dispatch`, #3952).
- **Process-group termination**: spawn each sweep in its own session/process group (`setsid`) and make cancellation + crash-path cleanup signal the *group* (TERM, escalate KILL), so wrapper, agent, and every descendant (build tools, sims, watcher loops) die together. The reaper's dead-pid path should do the same group reap when it detects a crashed wrapper with surviving children.
- Registry entry should record the pgid (or session id) so the reaper can find survivors even after reparenting to init/systemd.

## Acceptance criteria

- [ ] `loom-daemon cancel` cancels a sweep on the host it runs on, killing the entire process tree (verified: no surviving descendants after cancel, including processes that had reparented).
- [ ] The crash-path reaper detects and reaps a dead wrapper's surviving process group instead of leaving a zombie agent running unclaimed work.
- [ ] `cancel_sweep` (MCP) and the CLI share the termination implementation.

---

## Curator Enhancement

**Why `complex`, not `routine`**: this touches process-group signal delivery and an on-disk registry schema change (`owner.json`). A mistake here (e.g. a pgid field that's absent/stale on reconstruction, or a signal sent to the wrong group after a daemon restart) reproduces exactly the original incident — silently orphaned processes — and would very likely pass unit tests while still failing in the real reparent-after-restart case that motivated this issue. Judgment is needed on the on-disk schema/back-compat strategy for `owner.json` entries written before this change.

### Prior art already in place

Issue #3952 (closed) established the `loom-daemon dispatch` CLI-parity precedent this issue extends to `cancel`. Also, **`cmd.process_group(0)` already exists** (`loom-daemon/src/sweep_registry/dispatch.rs:1441`, added for #3800) — sweeps are already spawned as their own process-group leader, and group-kill delivery already exists via `send_group_signal`/`send_signal` (`loom-daemon/src/sweep_registry/reaper.rs:248-291`). The actual gap is narrower than "add process groups from scratch": `signal_sweep` (`loom-daemon/src/sweep_registry/locks.rs:337-343`) only takes the group-kill path `if self.children.contains_key(sweep_id)` — i.e. only while the in-memory `std::process::Child` handle from the original spawn is still alive in this daemon process. A `reconstruct()`-rebuilt entry (after a daemon restart, `locks.rs:533`, built from on-disk `owner.json` at `locks.rs:596-622`) has no such handle and no `pgid` field to fall back on (`LockOwner` struct, `locks.rs:122-127`: `issue`, `owner_pid`, `acquired_at`, `sweep_id` — no pgid), so it silently degrades to single-PID `send_signal`. A fresh CLI `cancel` process (which never held the spawn-time `Child` handle at all) would hit this same degraded path on every invocation, not just after a restart.

### Affected Files

- `loom-daemon/src/main.rs` (`Commands` enum ~line 33, `Dispatch` variant ~224-254) — add a sibling `Cancel` subcommand variant.
- `loom-daemon/src/daemon_service.rs` (~line 116-175) — add a `Commands::Cancel` match arm alongside the existing `Commands::Dispatch` arm.
- `loom-daemon/src/cli/cancel.rs` (new file, modeled on `loom-daemon/src/cli/dispatch.rs`: `handle_dispatch_command:75`, `build_dispatch_request:17`) — CLI-side handler for `cancel`, using shared helpers `resolve_socket_path` / `query_daemon_bounded` in `loom-daemon/src/cli/common.rs:20,39`.
- `loom-daemon/src/cli/mod.rs:12` — register the new `cancel` module.
- `loom-daemon/src/types.rs:269` (`Request::CancelSweep`), `types.rs:575` (`SweepCancelled` response), `types.rs:793` (`SweepInfo` struct) — add a `pgid` field to `SweepInfo`.
- `loom-daemon/src/sweep_registry/locks.rs:122-127` (`LockOwner` struct / `owner.json` schema) — add `pgid`, handle back-compat for pre-existing `owner.json` files without it.
- `loom-daemon/src/sweep_registry/locks.rs:337-343` (`signal_sweep`) — remove the `self.children.contains_key(sweep_id)` gate; call `send_group_signal` unconditionally using the stored `pgid`.
- `loom-daemon/src/sweep_registry/locks.rs:596-622` (`reconstruct`) — populate `pgid` from `owner.json` instead of omitting it.
- `loom-daemon/src/sweep_registry/dispatch.rs:1463` (right after `cmd.spawn()`, near the existing `cmd.process_group(0)` at line 1441) — capture and persist the pgid at spawn time.
- `loom-daemon/src/sweep_registry/reaper.rs` (`spawn_reaper_task:144`, cancellation core `begin_cancel`/`poll_cancel`/`finish_cancel`/`cancel`:337-539) — crash-path reap should use the same group-signal path once a pgid is available on a reconstructed entry.
- `loom-daemon/src/ipc.rs:1091-1107` (`Request::CancelSweep` handling), `ipc.rs:1344` (`cancel_sweep_nonblocking`) — ensure the CLI and MCP `cancel_sweep` (`mcp-loom/src/tools/sweeps.ts` `cancelSweep` ~line 675) share this same server-side termination path (already true today via `sendDaemonRequest` in `mcp-loom/src/shared/daemon.ts:102`; verify the new pgid path doesn't diverge the two callers).

### Test Plan

- [ ] Automated: extend `loom-daemon/src/sweep_registry/reaper.rs`'s existing `mod tests` (from line 1320) — the closest precedent is `cancel_terminates_whole_process_group_including_grandchild:2743`, which proves group-kill works via the retained in-process `Child` handle. Add a new test (e.g. `cancel_reconstructed_entry_kills_whole_group`) that reconstructs a `SweepInfo` from a serialized `owner.json` (simulating a daemon restart or a fresh CLI process with no retained handle) and asserts the whole process group — including a grandchild — is still terminated.
- [ ] Automated: add a `build_cancel_request_plumbs_all_flags`-style test in the new `loom-daemon/src/cli/cancel.rs`, following the pattern of `build_dispatch_request_plumbs_all_flags:171` in `loom-daemon/src/cli/dispatch.rs`'s `mod dispatch_tests`.
- [ ] Automated: extend `loom-daemon/src/ipc.rs`'s cancel-sweep tests (near `test_handle_request_cancel_sweep_unknown_returns_error:5250`) to cover the CLI-invoked path.
- [ ] Manual verification: dispatch a sweep locally, note its pid tree (`pstree -p <wrapper_pid>`), run `loom-daemon cancel --issue <N>` from a separate shell (simulating remote/SSH operation with no MCP server attached), and confirm every descendant — including a deliberately reparented one (e.g. spawn a long-running child, then kill and restart the daemon before cancelling) — is gone afterward.
- [ ] Manual verification: confirm `cancel_sweep` (MCP) and `loom-daemon cancel` (CLI) produce identical termination behavior for the same sweep (same code path, not two divergent implementations).
- [ ] Edge case: cancel a sweep whose registry entry was `reconstruct()`-ed after a daemon restart (no retained `Child` handle) and confirm group-kill still fires via the persisted `pgid`, not a silent single-PID fallback.
- [ ] Edge case: cancel a sweep against an `owner.json` written by a pre-change daemon binary (no `pgid` field) and confirm graceful degradation (documented fallback behavior, not a panic/crash).

rjwalters:
## Summary

Closes the two gaps behind the 2026-08-03 wedged-sweep incident, where cancelling a sweep over ssh turned into a three-round manual process hunt and hand-killing the tracked pid left a zombie `claude` agent that **relaunched its ngspice workload** (~35 CPU-hours) against an issue whose claim the crash path had already returned to `loom:issue`.

1. **No CLI cancel primitive.** `cancel_sweep` existed only as an MCP tool — i.e. only inside the operator's own Claude session on their own machine. Over ssh the only lever was raw `kill`.
2. **Killing the tracked pid did not terminate the sweep.** Group-kill was gated on the spawn-time in-memory `Child` handle, so a post-restart entry — and *every* fresh-process invocation — silently degraded to a single-PID kill that orphaned the subtree.

`cmd.process_group(0)` (#3800) already existed, so this is not "add process groups"; it is persisting the pgid and using it unconditionally, plus the missing CLI verb.

### `loom-daemon cancel <sweep-id|--issue N>` (new)

- `loom-daemon/src/cli/cancel.rs` (new), registered in `cli/mod.rs`, `main.rs` (`Commands::Cancel`), `daemon_service.rs` (match arm). Flags: `--issue`, `--grace` (default 30, matching the MCP tool), `--workspace` (same #4299 cwd-default resolution `dispatch` uses — `resolve_cli_dispatch_workspace` is now shared rather than duplicated).
- A **thin client** over the same `CancelSweep` IPC frame the MCP tool sends: the termination itself stays entirely in `cancel_sweep_nonblocking`, so the two operator surfaces cannot drift. `--issue N` is resolved to a sweep id client-side via one extra `ListSweeps` round-trip (live `Running`/`Pending` entries only; more than one is an error, not a guess), so the wire protocol is untouched.
- Client ack budget is `grace_secs + 15s` (the daemon acks only after the whole cancel completes) and bounded — it exits nonzero with the same "is loom-daemon running?" diagnostic `dispatch` uses rather than hanging.

### Process-group termination that survives the spawning process

- **`SweepInfo.pgid` / `LockOwner.pgid`** (`types.rs`, `sweep_registry/locks.rs`) — both `Option<u32>` + `#[serde(default, skip_serializing_if)]`.
- **Stamped at spawn** (`sweep_registry/dispatch.rs`) via `spawned_leader_pgid`, which **verifies** `getpgid(child) == child` rather than assuming it: a contradiction records no group (degrade) rather than aiming a later `kill(-pgid, 9)` at a group we do not own. Persisted into `owner.json` by `record_child_pid_in_lock`.
- **`signal_sweep` no longer gates on `self.children.contains_key(sweep_id)`** (`sweep_registry/reaper.rs`). It resolves the group from durable state and uses it unconditionally. Absent pgid ⇒ single-PID delivery **with a log line** (never silent); pgid equal to *this process's own group* ⇒ refused outright (a `kill(-our_pgid, 9)` would take down the daemon and every sweep it owns).
- **`reconstruct()` restores the pgid** (`locks.rs`) but re-verifies it against the live owner first — a value the OS contradicts (PID recycled across daemons) is discarded, not used.

### Crash-path reap (a dead leader is not a dead tree)

- New `reap_orphaned_group` / `escalate_pending_group_reaps` + `pending_group_reaps` map. The reaper's dead-pid branch and `reconstruct()`'s stale-lock branch group-SIGTERM a dead leader's surviving members and escalate to SIGKILL on a later tick.
- The escalation is **deferred, not slept**: `reap_once` also runs on the `ListSweeps` read path under the registry mutex, and blocking there is the 2026-07-26 wedge shape.
- The persisted pgid is the only handle available here — the OS will not report a dead pid's process group.

### Back-compat

`owner.json` written by a pre-change binary has no `pgid` key and **must** still deserialize: an unparseable owner is read everywhere (`reconstruct`, `lock_owned_by_other`, `live_sweep_lock_owner_pid`) as "no owner", which would drop a *live* sweep's lock. Covered by a dedicated test.

### Docs

`defaults/docs/daemon-reference.md` (new `loom-daemon cancel` section + a "Process-group termination and the persisted `pgid`" section + `SweepInfo.pgid`), `defaults/docs/troubleshooting.md`, `defaults/.claude/commands/loom/loom.md` and `defaults/.claude/agents/loom-daemon.md` (four places asserted "no CLI subcommand cancels a sweep by ID" and recommended the exact `kill -TERM <pid>` fallback that caused the incident — all corrected), and one line in `CLAUDE.md`.

- [x] **`loom-daemon cancel` cancels a sweep on the host it runs on, killing the entire process tree (including reparented processes).** Verified end-to-end against a live daemon (below): the wrapper had been reparented to systemd (`ppid 844`) after the daemon was SIGKILLed, the entry was rebuilt by `reconstruct()` with no `Child` handle, and a fresh CLI process removed all three processes and the claim lock.
- [x] **The crash-path reaper reaps a dead wrapper's surviving process group instead of leaving a zombie agent.** Verified live (below) and by `reaper_reaps_the_surviving_group_of_a_dead_leader` + `pending_group_reap_escalates_to_sigkill_on_a_later_tick`.
- [x] **`cancel_sweep` (MCP) and the CLI share the termination implementation.** Structural: the CLI sends the identical `CancelSweep` frame, and all termination is daemon-side. Pinned by `test_cancel_sweep_cli_and_mcp_frames_are_identical_on_the_wire` (byte-level frame parity) and `test_handle_request_cancel_sweep_from_cli_frame_runs_the_shared_path`.

## Test Plan

New automated coverage (all passing):

| Test | Covers |
|---|---|
| `cancel_reconstructed_entry_kills_whole_group` (reaper.rs) | The crux. Dispatch → real daemon-restart simulation via a second registry + `reconstruct()` (no retained handle) → cancel → the **grandchild** must be dead. |
| `reconstruct_tolerates_pre_pgid_owner_json_and_degrades_gracefully` | Pre-change `owner.json` (literal 4-key JSON, no `pgid`): parses, admits, cancels via the single-PID fallback — no panic. |
| `reconstruct_discards_a_pgid_the_os_contradicts` | A recorded pgid the OS disagrees with is dropped, not trusted. |
| `reaper_reaps_the_surviving_group_of_a_dead_leader` | Crash path: leader dead + orphan alive in the group ⇒ reaped by `reap_once`. |
| `pending_group_reap_escalates_to_sigkill_on_a_later_tick` | A trap-TERM orphan is SIGKILLed on the deferred tick. |
| `group_signalling_refuses_this_processs_own_group` | Safety floor: our own group is never signalled. |
| `build_cancel_request_plumbs_all_flags`, `..._defaults_match_the_mcp_tool`, `select_sweep_for_issue_*` (×3), `round_trip_sends_cancel_sweep_and_parses_the_outcome`, `absent_socket_errors_fast` (cli/cancel.rs) | CLI flag plumbing, `--issue` selection rules, wire round-trip. |
| `test_cancel_sweep_cli_and_mcp_frames_are_identical_on_the_wire`, `test_handle_request_cancel_sweep_from_cli_frame_runs_the_shared_path` (ipc.rs) | Shared-path guarantee. |
| `sweep_info_schema_snapshot` (updated) | `pgid` pinned in the wire schema; pre-#4980 JSON still deserializes to `None` and re-serializes without the key. |

**Mutation-checked**: `cancel_reconstructed_entry_kills_whole_group` was re-run with `signal_sweep` reverted to the old `children.contains_key` gate and **fails** with `grandchild survived a cancel of a RECONSTRUCTED entry` — it is not a vacuous test.

**Manual verification (performed, not skipped)** — a real `loom-daemon` binary against a throwaway workspace with a fixture spawn adapter modelling the real tree shape (`wrapper → agent → workload`) and a stub `gh`:

```
# owner.json after dispatch
{"issue":4980,"owner_pid":2574933,...,"pgid":2574933}

# tree
sleep(2574933)---sleep(2574934)---sleep(2574935)

# SIGKILL the daemon; wrapper reparents to systemd
2574933 2574933  844 sleep 600        <- ppid 844 (reparented)

# restart daemon -> "reconstructed 1 sweep entry"; then, from a separate shell:
$ loom-daemon cancel --issue 4980 --grace 3
Cancelled sweep sweep-issue-4980-1785729217
  pid:       2574933 (process group torn down)
  escalated: no — exited on SIGTERM
# survivors: none. claim lock: released.
```

Crash path, same fixture (`LOOM_SWEEP_REAPER_INTERVAL_SECS=5`), hand-killing only the tracked wrapper pid — the exact 2026-08-03 mistake:

```
# after kill -9 <wrapper>: agent+workload survive, agent reparented to 844
# one reaper tick later:
[WARN] reap_orphaned_group: sweep sweep-issue-4980-... (issue #4980) has a DEAD leader
       but its process group 2582893 still has members — an orphaned agent/subtree
       running unclaimed work. Sending SIGTERM to the group; escalating to SIGKILL in 5s...
# survivors: none.
```

Also run: `cargo fmt`, `cargo clippy -p loom-daemon --all-targets` (clean), `cargo test -p loom-daemon` (full suite, green).

## Notes for the reviewer

- Adding `SweepInfo.pgid` required a one-line `pgid: None,` in ~40 existing struct literals (almost all test fixtures). Those insertions are mechanical; the substantive changes are in `reaper.rs`, `locks.rs`, `dispatch.rs`, `types.rs`, and the new `cli/cancel.rs`.
- The main checkout in this environment had pre-existing uncommitted `.loom/` resync drift (present before this work started); nothing in it belongs to this PR.

Closes #4980

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -257,7 +257,8 @@ worktrees, stuck agents, daemon registry/event-bus/reaper issues, host-sleep and
 `.loom/` resync procedures, and common fixes. Quick fixes: `loom-clean --force`
 (stale worktrees/branches), `loom-recover-orphans --recover` (orphaned
 `loom:building` issues), `gh label sync --file .github/labels.yml` (re-sync
-labels), `mcp__loom__cancel_sweep --sweep_id <id>` (cancel a running sweep).
+labels), `loom-daemon cancel --issue <N>` / `mcp__loom__cancel_sweep` (cancel a
+running sweep — never hand-`kill` its pids, #4980).
 
 ## Migration History
 
--- a/defaults/.claude/agents/loom-daemon.md
+++ b/defaults/.claude/agents/loom-daemon.md
@@ -13,7 +13,7 @@ Follow the complete role definition in `.loom/roles/loom.md` for:
 - Dispatching work — `mcp__loom__dispatch_sweep` launches a `/loom:sweep <issue>` child with multi-account OAuth token rotation via `spawn-claude.sh`
 - Observing state — `mcp__loom__list_sweeps`, `mcp__loom__get_sweep_status`, `mcp__loom__tail_sweep_log`
 - Eventing — `mcp__loom__subscribe_to_events` / `mcp__loom__tail_event_bus` over the frozen 6-topic taxonomy
-- Intervening — `mcp__loom__cancel_sweep` (SIGTERM → grace → SIGKILL); the daemon process itself keeps running
+- Intervening — `mcp__loom__cancel_sweep`, or `loom-daemon cancel <sweep-id|--issue N>` from a shell / over ssh (#4980, same IPC request): SIGTERM → grace → SIGKILL of the sweep's whole process **group**; the daemon process itself keeps running. Never hand-`kill` a sweep's PIDs — that kills the wrapper and leaves the agent alive
 - Host-sleep readiness for long / overnight runs (`check-host-sleep.sh`, advisory-only)
 
 **By default the daemon is not a work generator**: with no autonomous config it does not poll the forge for `loom:issue` items and it does not drive support roles on cron — operator-driven `mcp__loom__dispatch_sweep` and the GitHub Actions cron workflows own those responsibilities. Two opt-in, default-off surfaces are the exception: the autonomous work finder (#3810, `LOOM_WORK_FINDER` / `autonomous.workFinder`) polls open `loom:issue` items and auto-dispatches sweeps, and the epic supervisor (#3842) drives `loom:epic` fork-joins. See `.loom/docs/daemon-reference.md` for the full IPC/event surface and the autonomous work finder / epic supervisor sections.
--- a/defaults/.claude/commands/loom/loom.md
+++ b/defaults/.claude/commands/loom/loom.md
@@ -109,7 +109,7 @@ The `loom-daemon` binary talks to the same running daemon over the same Unix-soc
 |---|---|---|
 | `mcp__loom__list_sweeps` | `loom-daemon status` [`--json`] [`--pipeline`] | Richer than `list_sweeps`: also reports the three dynamic-cap inputs (token-pool size, disk headroom, configured ceiling) + their `min`, the main-health-gate halt state, and per-token usage. `--pipeline` adds the forge-side `gh` snapshot (opt-in — extra API calls). |
 | `mcp__loom__dispatch_sweep` | `loom-daemon dispatch <N>` [`--model`] [`--effort`] [`--depends-on`] [`--workspace`] | First-class non-MCP entry point (#3952) over the same IPC `DispatchSweep` request. Bounded client-side ack timeout — exits nonzero fast instead of hanging (built explicitly to avoid the #4043 MCP wedge). |
-| `mcp__loom__cancel_sweep` | *(none)* | No CLI subcommand cancels an in-flight sweep by ID. Fallback: get the PID from `loom-daemon status` and `kill -TERM <pid>` directly — the `.loom/sweep-checkpoint/` file still survives, so redispatch still resumes. `loom-daemon quarantine clear <issue>` is the adjacent lever for a crash-looping issue (releases the daemon's insta-crash pause), not a manual-cancel substitute. |
+| `mcp__loom__cancel_sweep` | `loom-daemon cancel <sweep-id>` \| `--issue <N>` [`--grace`] [`--workspace`] | First-class non-MCP entry point (#4980) over the same IPC `CancelSweep` request — the `dispatch` sibling, usable over ssh. **Do NOT `kill -TERM <pid>` from `loom-daemon status` instead** (the pre-#4980 fallback): the daemon tracks the *wrapper* pid, so killing it leaves the underlying `claude` agent alive — on 2026-08-03 that survivor relaunched its workload against an issue whose claim had already been returned to the queue. The CLI signals the whole process group. `.loom/sweep-checkpoint/` still survives a cancel, so redispatch still resumes. |
 | `mcp__loom__get_sweep_status` | *(none, partial)* | `loom-daemon status` gives fleet-wide state, not one sweep's phase/blockers. `loom-daemon watch add <N>` (add `--pr` to watch a PR instead of an issue) registers a durable watch on that issue/PR's terminal state instead (persists to `~/.loom/watches.json`, survives a daemon restart, resolves to `~/.loom/logs/watch-results.log`). |
 | `mcp__loom__tail_sweep_log` | *(none)* | `tail -f .loom/logs/sweep-issue-<N>.log` directly. |
 | `mcp__loom__tail_event_bus`, `subscribe_to_events`, `publish_event` | *(none)* | Live pub/sub is MCP-only; there is no CLI event stream. Poll `loom-daemon status` + `gh issue/pr list` on your loop cadence instead. |
@@ -134,7 +134,7 @@ When the daemon is running, you coordinate work via MCP tools where available, a
 |---|---|
 | `mcp__loom__list_sweeps` | `loom-daemon status` |
 | `mcp__loom__dispatch_sweep` | `loom-daemon dispatch <N>` |
-| `mcp__loom__cancel_sweep` | none — see "CLI Fallback" |
+| `mcp__loom__cancel_sweep` | `loom-daemon cancel <sweep-id>` / `--issue <N>` |
 | `mcp__loom__get_sweep_status`, `tail_sweep_log`, `tail_event_bus`, `subscribe_to_events`, `publish_event` | none — MCP-only |
 
 **Each iteration:**
@@ -197,13 +197,18 @@ When the daemon is running, you coordinate work via MCP tools where available, a
    ```
    This sends SIGTERM, waits the configured grace window, then SIGKILL. The `.loom/sweep-checkpoint/issue-<N>.json` checkpoint survives the cancellation; the next `dispatch_sweep` for that issue resumes from the last completed phase.
 
-   **No CLI subcommand cancels a sweep by ID.** If MCP is unavailable: get the
-   PID from `loom-daemon status`, `kill -TERM <pid>` it directly (the checkpoint
-   still survives — cancellation semantics are the same, just invoked manually),
-   and if it keeps crash-looping on redispatch, `loom-daemon quarantine clear
-   <issue>` clears the daemon's insta-crash pause once you're ready to
-   redispatch (quarantine is about crash-loop protection, not manual
-   cancellation, but it's the adjacent CLI lever).
+   **CLI equivalent** (#4980), for when MCP is unavailable or you are on ssh:
+   ```
+   loom-daemon cancel <sweep-id>        # or: loom-daemon cancel --issue <N>
+   ```
+   Same IPC request, same daemon-side termination. **Do not `kill -TERM <pid>`
+   the PID from `loom-daemon status` instead** — that was the pre-#4980 fallback
+   and it is how the 2026-08-03 incident happened: the tracked PID is the
+   *wrapper*, so killing it leaves the `claude` agent alive to relaunch its
+   workload against an issue whose claim has already been returned to the queue.
+   If a sweep keeps crash-looping on redispatch, `loom-daemon quarantine clear
+   <issue>` clears the daemon's insta-crash pause (crash-loop protection, not a
+   cancellation substitute).
 
 6. **Tail per-sweep logs** if you need to inspect output:
    ```
@@ -250,7 +255,7 @@ In-session subagent dispatch (`/loom:sweep` with Stage -1 falling through to sub
 | `/loom:loom` | Check daemon (MCP `list_sweeps`, fallback `loom-daemon status`), start observing/dispatching |
 | `/loom:loom status` | `mcp__loom__list_sweeps`, or `loom-daemon status` if MCP is unavailable |
 | `/loom:loom health` | Display daemon health summary (registry + recent events, or `loom-daemon status` which folds in the health-gate state) |
-| `/loom:loom stop` | Cancel all in-flight sweeps via `mcp__loom__cancel_sweep` (no CLI equivalent — see below); daemon process itself stays alive |
+| `/loom:loom stop` | Cancel all in-flight sweeps via `mcp__loom__cancel_sweep` (CLI: `loom-daemon cancel <sweep-id>`, #4980); daemon process itself stays alive |
 | `/loom:loom help` | Show comprehensive help guide |
 | `/loom:loom help <topic>` | Show help for a specific topic |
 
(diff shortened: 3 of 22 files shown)


rjwalters:
**Test-environment note (for the reviewer), no action needed on this PR.**

Running `cargo test -p loom-daemon --lib --bins` on *this* worker box shows 10 failures:

```
auto_update::tests::test_config_{missing_file,malformed_json,missing_block}_is_default
work_finder::tests::test_config_* (7)
```

They are **pre-existing and host-environmental**, not caused by this change. `read_work_finder_config` / `read_auto_update_config` go through `config_resolver::resolve_effective_config`, whose highest-precedence tier is the machine-level private defaults file `~/.local/share/loom/config/defaults.json` (`LOOM_CONFIG_DEFAULTS_FILE`). This worker has one, and it sets `autonomous.workFinder.enabled/maxConcurrent`, `perTokenConcurrency`, and `autonomous.autoUpdate.enabled` — so the tests' `tempfile::tempdir()` isolation does not actually isolate them, and they read the host's values instead of the expected `None`s.

Proof — neutralize just that tier and the same selection is fully green:

```
$ LOOM_CONFIG_DEFAULTS_FILE=/nonexistent HOME=/tmp/nohome cargo test -p loom-daemon --lib test_config
test result: ok. 64 passed; 0 failed
```

CI agrees: **Rust Unit Tests passed** on this PR (clean runner, no such file), along with Rust Linting & Build Check and Rust OTLP Feature Build & Test.

The PR body's "full suite green" line refers to that clean-environment result. The gap is worth a separate issue (those config tests should pin `LOOM_CONFIG_DEFAULTS_FILE=""` so they are hermetic on a host that has the machine-level defaults file), but it is out of scope here — filing it rather than fixing it in this PR.

rjwalters:
Follow-up filed for the non-hermetic config tests: #4983 (out of scope for this PR).

rjwalters:
✅ **Approved!**

Thorough, safety-conscious fix for a real incident (2026-08-03 zombie-agent). Reviewed the full diff (`git diff origin/main...HEAD`, 22 files / ~1600 insertions scoped tightly to the PR description — the wider `git diff main...HEAD` was misleading since the worktree's local `main` ref was 8 commits stale).

**Correctness highlights:**
- `spawned_leader_pgid` verifies `getpgid(child) == child` rather than assuming `process_group(0)` took — degrades to `None` on contradiction instead of recording a group the child doesn't lead.
- `signal_sweep` now resolves the group unconditionally from persisted `SweepInfo.pgid` (fixing the `children.contains_key` gate that caused the incident for both reconstructed entries *and* every fresh `loom-daemon cancel` invocation), but refuses outright if the resolved pgid equals the daemon's own group (`current_process_group()`) — a strong guard against a stale/recycled-PID record taking down the daemon itself.
- Crash-path reap (`reap_orphaned_group` / `escalate_pending_group_reaps`) correctly defers SIGKILL escalation to a later `reap_once` tick rather than sleeping inline, avoiding the 2026-07-26 wedge shape since `reap_once` also runs under the registry mutex on the `ListSweeps` read path.
- `LockOwner.pgid` is `Option<u32>` + `#[serde(default)]`, so pre-#4980 `owner.json` files still deserialize — verified by a dedicated test and by `reconstruct()` re-validating a persisted pgid against the live OS-reported group before trusting it (PID-recycling safety).
- `loom-daemon cancel` is a genuinely thin client — same `CancelSweep` IPC frame the MCP tool sends, pinned by a byte-level wire-format test plus an end-to-end test that a CLI-parsed frame runs the identical daemon-side termination + lock-release path.

**Verification performed:**
- `cargo test -p loom-daemon --lib`: 3137/3145 pass. The 8 failures are all in `work_finder`/`auto_update` config-parsing tests, reproduce identically on a clean `main` checkout in this sandbox (confirmed — they read this machine's real `.loom/config.json` instead of an isolated fixture), and are untouched by this PR's diff. Pre-existing local-environment pollution, not a regression — CI's own "Rust Unit Tests" check is green.
- All new #4980 tests pass in isolation (`cancel_reconstructed_entry_kills_whole_group`, `reconstruct_tolerates_pre_pgid_owner_json_and_degrades_gracefully`, `reconstruct_discards_a_pgid_the_os_contradicts`, `reaper_reaps_the_surviving_group_of_a_dead_leader`, `pending_group_reap_escalates_to_sigkill_on_a_later_tick`, `group_signalling_refuses_this_processs_own_group`).
- `cargo clippy -p loom-daemon --all-targets`: clean.
- Confirmed `Commands::Cancel` is routed in `run_daemon()` before falling through to the sync `handle_cli_command`, mirroring `Dispatch`.
- Docs cross-checked against the four claimed correction sites (`daemon-reference.md`, `troubleshooting.md`, `loom.md`, `loom-daemon.md`) plus `CLAUDE.md` — all consistently updated, no stale "no CLI subcommand" claims left behind.

## Test Execution

1. Automated test table (unit/integration coverage) — ✅ Executed: ran the full crate + targeted new-test subset, all pass (see above)
2. Manual verification against a live daemon with a fixture spawn adapter — ⚠️ Not re-executed (requires manual live-process observation); PR body documents concrete before/after output that matches the code path reviewed
3. `cargo fmt` / `cargo clippy` / `cargo test -p loom-daemon` — ✅ Executed: clippy clean, tests as above

PR #4743, opened 2026-07-31
fix(daemon): mark cooldown-boundary bad_tokens tests #[serial]

rjwalters:
## Summary

`tokens_pool::bad_tokens::tests::blocking_entry_reports_exhaustion_class_and_cooldown_remaining`
and `tokens_pool::bad_tokens::tests::blocking_entry_reports_the_deciding_fresh_line_not_the_stale_one`
were flaky under the default parallel test harness.

**Root cause**: both tests assert that an entry 1h old / 10m old (respectively)
still blocks selection under the *default* 6h exhaustion cooldown. The existing
`#[serial]` test `blocking_entry_cooldown_remaining_honors_env_override`
transiently mutates the process-global `EXHAUSTION_COOLDOWN_ENV` var to 1800s
and then 300s. If either of the two non-serial tests above runs concurrently
while that shrunk override is in effect, their entry's age can exceed the
transient cooldown, making `blocking_entry` return `None` unexpectedly and
panicking the "fresh entry blocks" / "re-marked account blocks" `.expect()`
calls.

Reproduced reliably on unmodified `main`: 3 of 5 back-to-back
`cargo test -p loom-daemon --lib tokens_pool::bad_tokens` runs failed, always
on `blocking_entry_reports_exhaustion_class_and_cooldown_remaining`.

**Fix**: add `#[serial]` to both tests, matching the existing convention
already used for env-mutating / boundary-sensitive tests in this module
(`mark_bad_errors_when_dir_missing`, `exhaustion_cooldown_env_override`,
`blocking_entry_cooldown_remaining_honors_env_override`). I audited the rest
of the module for similar boundary sensitivity — `exhaustion_entry_expires_
after_cooldown_auth_stays` uses a 7h-old entry that is already past the
*default* 6h cooldown, so a transient shrink to 1800s/300s can't flip its
assertion (it still expects expiry either way) — no further `#[serial]`
additions were needed.

## Verification

- Confirmed the failure reproduces on unmodified `main` (3/5 runs failed,
  same panic as reported in the issue).
- After the fix: 10 consecutive `cargo test -p loom-daemon --lib
  tokens_pool::bad_tokens` runs all passed (27/27 each run).
- Full `cargo test -p loom-daemon --lib` run: the `tokens_pool::bad_tokens`
  module passed cleanly; 3 unrelated failures in `role_runner`,
  `runtime_admission`, and `sweep_registry` reproduce identically with this
  change stashed out (pre-existing on `main`, unrelated to this fix).

Closes #4736

Issue #4736, opened 2026-07-31
tokens_pool::bad_tokens tests flaky under parallel execution due to shared EXHAUSTION_COOLDOWN_ENV mutation


rjwalters:
While evaluating PR #4519, CI's `Rust Unit Tests` failed on:

- `tokens_pool::bad_tokens::tests::blocking_entry_reports_exhaustion_class_and_cooldown_remaining`
- `tokens_pool::bad_tokens::tests::blocking_entry_reports_the_deciding_fresh_line_not_the_stale_one`

PR #4519 does not touch `loom-daemon/src/tokens_pool/bad_tokens.rs` at all, and the failure reproduces reliably on unmodified `main` (confirmed with 3 consecutive `cargo test -p loom-daemon --lib tokens_pool::bad_tokens` runs, all failing the same way).

## Root cause (probable)

`blocking_entry_cooldown_remaining_honors_env_override` (loom-daemon/src/tokens_pool/bad_tokens.rs:664) sets the process-global `EXHAUSTION_COOLDOWN_ENV` var to very short values (300s, 1800s) and is marked `#[serial]`, but the two failing tests above are **not** `#[serial]`. Under the default parallel test harness, the env-override test can transiently shrink the cooldown window while the other tests are mid-assertion on their own "fresh" (1h-old / 10m-old) entries, making `blocking_entry` return `None` unexpectedly ("fresh entry blocks" / "re-marked account blocks" panics).

## Fix

Mark `blocking_entry_reports_exhaustion_class_and_cooldown_remaining` and `blocking_entry_reports_the_deciding_fresh_line_not_the_stale_one` (and any other test in this module asserting on entries near the default cooldown boundary) with `#[serial]` to match the existing convention already used for env-mutating tests in this file (also see `mark_bad_errors_when_dir_missing`).

Discovered during code evaluation of PR #4519.

## Status update (2026-07-31, superseded — see 2026-08-04 update below)

[...]

PR #4987, opened 2026-08-03
fix(curator): skip duplicate "still blocked, no change" dependency re-checks

### LINKED ISSUE #4986: Curator re-posts duplicate 'still blocked, no change' comments on unchanged loom:blocked issues
## Context

A Curator role instructed to re-verify a Dependencies checklist on every pass, and to comment the outcome, has no check for "did I already report this exact conclusion recently" before posting. Observed live in a downstream consumer repo: dozens of near-identical "still blocked on #N, no change" comments piled up across several `loom:blocked` issues within about two days, several arriving less than an hour apart, all restating the same blocker with zero new information. Each redundant comment costs a full Curator invocation (issue read, dependency re-verification, comment post).

This is the same class of defect already addressed for Champion in #4954 → #4966, with a follow-on interaction bug in #4967 worth learning from (see below).

## Root cause

`curator.md`'s "Checking Dependencies" section directs the Curator to re-verify the Dependencies checklist on every pass and comment the outcome, but never instructs it to first check whether that outcome is identical to its own most recent comment on the same issue. Since multiple Curator invocations (manual, autonomous, sweep-triggered) can all land on the same stale-but-still-blocked issue, the natural steady state is "one comment per pick-up, forever" until the real blocker clears.

This is distinct from useful re-checks: a comment reporting *new* information (a dependency closed, the blocking issue number changed, a new blocker surfaced) is valuable. A comment reporting the exact same blocker with the exact same status as the immediately preceding comment is not.

## Suggested fix

Add a dedup rule to `curator.md`'s Dependencies re-check flow, modeled on the body-hash idempotency mechanism already in `champion-issue-promo.md`'s "Concurrency Guard and Idempotency" section:

- Before commenting a Dependencies re-check outcome on a `loom:blocked` issue, read the most recent prior Curator comment on that issue. If it states the same conclusion (same blocking issue number(s), same blocker status), skip commenting entirely — leave the issue as-is (still `loom:blocked`, no label change).
- Post a comment whenever the conclusion changes (blocker cleared, new blocker surfaced, blocking issue number changed, issue now unblocked).
- Allow a staleness-window exception (e.g. 24h): if the most recent "unchanged" comment is older than the window, one fresh heartbeat confirmation is allowed even though the conclusion hasn't changed, so genuinely long-stalled issues still get periodic visibility instead of going silent forever.

## Caution from #4967

When implementing, avoid the interaction bug documented in #4967: a silent-skip-on-identical-conclusion mechanism can make an *escalation* counter unreachable if one is ever added later (e.g. "flag for human after N consecutive unchanged confirmations") — because the very re-check that would increment the counter is the one being silently skipped. If Curator's Dependencies re-check ever grows an analogous "escalate to a human after N unchanged confirmations" step, track that count independently of the comment-suppression decision, not derived from "how many times did we actually comment."

- [ ] Curator checks its own most recent Dependencies re-check comment on an issue before posting a new one; identical conclusion (same blocker number(s)/status) → skip silently, no comment, no label change
- [ ] A changed conclusion (blocker cleared / new blocker / blocking issue number changed) always produces a comment — no exception
- [ ] A staleness-window exception (~24h) allows one heartbeat comment on an otherwise-unchanged, long-stalled issue
- [ ] A worked example (skipped no-op pass vs. a real re-check comment) added to `curator.md` alongside the existing "Curation Patterns" section

## Note

A downstream consumer repo applied a local `.loom/resync-ignore`-pinned customization to its installed `curator.md` copy (both `.claude/commands/loom/curator.md` and `.loom/roles/curator.md`) implementing the rule above, pending this upstream fix landing in `defaults/`. Once this lands upstream, that repo's local pin can be dropped in favor of the synced copy.

rjwalters:
Closes #4986

## Problem

`curator.md`'s "Checking Dependencies" → "When Dependencies Complete" flow directs the Curator to re-verify the superseding-block condition on **every** pass and comment the outcome, but never instructs it to check whether that outcome is identical to its own most recent comment. Since manual, autonomous, and sweep-triggered invocations all land on the same stale `loom:blocked` issue, the natural steady state was "one comment per pick-up, forever" — #4736 accumulated six near-identical "still blocked on PR #4743, no change" comments between 2026-07-31 14:11 and 2026-08-01 06:10, several under an hour apart, each with zero new information and each costing a full Curator invocation.

Same defect class as #4954 → #4966 for Champion.

## Change

Single file: `defaults/.claude/commands/loom/curator.md` (+161 lines, docs/prompt-text only, no code path).

**New `### Re-check Idempotency` subsection** under "Checking Dependencies", modeled on `champion-issue-promo.md`'s "Concurrency Guard and Idempotency" body-hash marker — but per Curator's own curation note, fingerprinting the **blocked condition**, not the issue's body text:

- **Conclusion fingerprint** = verdict (`blocked`/`clear`) + each linked PR's `number:state:sorted loom labels` (sorted, so API ordering churn never reads as a change) + the cited block reason when the block came from the secondary heuristic rather than a linked PR.
- Marker `<!-- curator:dep-recheck:<hash> -->` embedded in every re-check comment, so the next pass compares mechanically instead of parsing prose.
- **Three-way decision table**:

  | Prior comment | Action |
  |---|---|
  | None (first-ever check) | Comment — first pass is never skipped |
  | Different hash | Comment — no exception, no window, no budget |
  | Same hash, inside window | Skip silently — no comment, no label change, no `loom:curating` claim |
  | Same hash, past window | One heartbeat comment; posting restarts the window |

- **24h staleness window** (`LOOM_DEP_RECHECK_HEARTBEAT_HOURS`, default `24`) so long-stalled issues keep periodic proof-of-life instead of going silent forever.
- Explicit "a silent skip is a complete outcome, not a deferral" — no label churn, no handoff to another role.
- Legacy note: pre-existing unmarked "still blocked" comments cause exactly one re-post before dedup engages; don't parse legacy prose to avoid it.

**Worked example** in "Curation Patterns" (`### Blocked Issue Re-check → Silent Skip vs. Real Comment`) walking the real #4736/#4743 sequence: Pass 1 comments, Passes 2-6 skip silently (the six duplicates collapse to one), Pass 3 comments unconditionally once #4743 merges, plus the staleness-heartbeat variant.

**Pointer** added to "Decision Logic" so the rule is reachable from the top of the Dependencies flow.

## Handling of the #4967 caution

No escalation counter is implemented. The section records the constraint for anyone who adds one later: track the tally on the existing marker comment via `PATCH repos/{owner}/{repo}/issues/comments/<id>` (exactly as `champion-issue-promo.md`'s "Bounding the silent skip" does) so a **silent skip still increments it** — never derive it from "how many comments did we actually post", which is precisely the quantity the suppression zeroes out. The marker is deliberately shaped as a durable, addressable comment carrying the conclusion identity rather than a bare "did we comment?" boolean, so that path stays open.

## Acceptance criteria

- [x] Checks its own most recent re-check comment before posting; identical conclusion → skip silently, no comment, no label change
- [x] Changed conclusion (blocker cleared / new blocker / blocking number changed) always comments — no exception
- [x] ~24h staleness-window exception allows one heartbeat comment
- [x] Worked example added alongside "Curation Patterns"
- [x] Edge case from the Curator test plan: first-ever check always comments, never skips

## Single-source note

`defaults/roles/curator.md` and `.loom/roles/curator.md` are both symlinks to `defaults/.claude/commands/loom/curator.md`, so one edit covers all in-repo copies; downstream installed copies land via `resync-installed.sh`. Matches the convention of #4966, which touched only `defaults/`. The downstream repo's `.loom/resync-ignore` pin on its local `curator.md` can be dropped once this merges.

## Verification

- `bash -n` on the extracted embedded snippet: syntax OK (shellcheck flags only SC2034 for variables consumed by the prose decision table, as intended).
- `scripts/check-claude-md-budget.sh`: OK (287/320 lines — `CLAUDE.md` untouched).
- `scripts/check-docs-defaults-parity.sh`: OK.
- `scripts/check-agents-md-sync.sh`: OK.
- No automated tests — role-prompt markdown has no executable path; verification is scenario-based per the issue's test plan, all three scenarios encoded as the worked example.

### DIFF
--- a/defaults/.claude/commands/loom/curator.md
+++ b/defaults/.claude/commands/loom/curator.md
@@ -701,6 +701,10 @@ This issue cannot proceed until dependencies are complete.
    to mark `loom:curated`
 3. **Any unchecked** → Add/keep `loom:blocked` label, do NOT mark `loom:curated`
 
+**Before commenting *any* re-check outcome** (blocked or clear) on an issue that
+is already `loom:blocked`, apply "Re-check Idempotency" below — an unchanged
+conclusion is skipped silently rather than re-posted.
+
 **If NO Dependencies section:**
 - Issue has no blockers → Safe to mark `loom:curated`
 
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Single-file, docs-only change to `defaults/.claude/commands/loom/curator.md` — adds the "Re-check Idempotency" section for the Dependencies re-check flow, mirroring the existing marker/idempotency pattern already used in `champion-issue-promo.md`.

**Verification performed:**
- Extracted the embedded `bash` snippet (fingerprint/marker computation) and ran `bash -n` + `shellcheck`: syntax is valid; the only shellcheck output is `SC2034` (unused-var) on `RECHECK_MARKER`/`PRIOR_HASH`/`PRIOR_AGE_H`, which is expected since those are consumed by the prose decision table, not executed code — matches the PR's stated verification.
- Confirmed the new section's `closedByPullRequestsReferences` mechanism is consistent with the existing "superseding block" check just above it (lines 741-754) — same data source, same PR-state/label reasoning, no new assumptions introduced.
- Re-ran the repo's own gates locally: `check-docs-defaults-parity.sh` OK, `check-claude-md-budget.sh` OK (287/320, CLAUDE.md untouched), `check-agents-md-sync.sh` OK.
- Confirmed the single-source symlink claim: `.loom/roles/curator.md` and `defaults/roles/curator.md` both resolve to `defaults/.claude/commands/loom/curator.md`, so this one edit is the sole source of truth.
- PR body already uses `Closes #4986`.

**Design read:** the three-way decision table (none → comment, changed hash → comment, same hash within window → skip, same hash past window → heartbeat) correctly separates "unchanged conclusion" from "no re-verification happened" — re-verification still runs every pass, only the redundant comment is suppressed. The forward-looking note on the #4967 escalation-counter concern (must track independently of comment-suppression, on the marker comment) is sound and doesn't overreach into implementing something not yet needed.

## Test Execution

No "## Test Plan" section in the PR description — this is expected for a role-prompt/markdown-only change with no executable path; the PR's own "Verification" section substitutes (bash -n / shellcheck / the three repo gate scripts), all of which I independently reproduced above.

All CI checks pass, merge state is CLEAN.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (another Judge's approval landed at 2026-08-03T05:43:32Z, concurrent with my own review). Discarding my review; not writing a competing label.

For the record, my independent review reached the same conclusion — approve. Summary: the Re-check Idempotency subsection is technically sound (conclusion fingerprint = verdict + sorted per-PR `number:state:loom-labels` + block reason; marker `<!-- curator:dep-recheck:<hash> -->`), the three-way decision table matches the acceptance criteria (no-prior-comment and changed-hash both always comment; same-hash-in-window skips silently; same-hash-past-24h-window posts exactly one heartbeat), the #4967 caution is handled correctly (no escalation counter added, but the marker is explicitly shaped to carry a future PATCH-based tally per champion-issue-promo.md's 'Bounding the silent skip' pattern), a worked example was added under 'Curation Patterns' walking the real #4736/#4743 sequence, and the embedded bash snippet is syntactically valid (verified with `bash -n` and shellcheck — only expected SC2034 warnings for prose-consumed variables). `defaults/roles/curator.md` and `.loom/roles/curator.md` remain untouched symlinks; only `defaults/.claude/commands/loom/curator.md` was edited. CI is green and merge state is CLEAN.

rjwalters:
## Judge Review: APPROVE

Reviewed as a docs/prompt-text-only change to `defaults/.claude/commands/loom/curator.md` (+161/-0, single file). CI is green (12/12 checks, including Shell Test Suites which was still pending at first pass — re-polled to completion) and `mergeStateStatus` is `CLEAN`.

**Mechanism correctness** — traced the three-way (really four-row) decision table against the stated goal:

| Case | Table row | Outcome | Matches goal? |
|---|---|---|---|
| First-ever check | `PRIOR` empty | Comment | Yes — never skips a first pass |
| Same `CONCLUSION_HASH`, within 24h | skip silently, no label/claim churn | Yes — this is the #4736 dedup |
| Different `CONCLUSION_HASH` | comment, unconditionally | Yes — no exception, matches AC2 |
| Same hash, past 24h window | one heartbeat comment, marker timestamp refreshes | Yes — matches AC3, and the refresh correctly restarts the window from the heartbeat's own `createdAt` rather than the issue's aggregate `updatedAt` |

The fingerprint is anchored to conclusion content (verdict + sorted `number:state:labels` per linked PR + block reason), not to `issue.updatedAt` — correctly sidesteps the self-invalidating-anchor trap `champion-issue-promo.md` documents for #4966 (an anchor that Curator's own write would bump can never match on the next pass). Staleness is likewise measured from the marker comment's own `createdAt`, not the issue's aggregate timestamp — same correct anchor choice.

**Consistency with `champion-issue-promo.md`'s "Concurrency Guard and Idempotency"**: same shape (portable `_sha256` fallback chain, `<!-- marker:hash -->` embedded in the posted comment, REST `issues/comments/<id>` PATCH called out as the extension point), scaled down appropriately — Curator's marker doesn't need Champion's claim-arbitration layer (`loom:evaluating`/staleness-aware reclaim) since this is a single re-check step, not a full evaluation pass with a concurrent-claim risk of its own.

**#4967 caution**: correctly *not* implemented. No escalation counter, no tally field on the marker. The new "Leaving room for a future escalation counter" paragraph explicitly requires any future counter to live on the marker comment via `PATCH .../issues/comments/<id>` (exactly champion's "Bounding the silent skip" pattern) so a silent skip still advances it — i.e., the marker is deliberately a durable, addressable comment carrying conclusion identity rather than a bare boolean, which is what keeps a future counter reachable rather than precluding it. This matches the ask precisely: don't build it, don't preclude it.

**Acceptance criteria (issue #4986)** — all 4 satisfied: (1) own-most-recent-comment check + silent skip on identical conclusion, no label churn — explicit "A silent skip is a complete outcome" paragraph; (2) changed conclusion always comments, no exception — table row 2; (3) ~24h staleness heartbeat — table row 4 + dedicated variant in the worked example; (4) worked example added under "Curation Patterns" (`### Blocked Issue Re-check → Silent Skip vs. Real Comment`), using the real #4736/#4743 timeline with correct arithmetic (Pass 2 at +53min inside the window, the heartbeat variant at +24.8h outside it).

**Scope discipline**: confirmed the diff touches only `defaults/.claude/commands/loom/curator.md` — `.loom/roles/curator.md` and `defaults/roles/curator.md` are symlinks to it (single edit covers all in-repo copies), and the installed `.claude/commands/loom/curator.md` copy is correctly left untouched for `resync-installed.sh` to propagate, matching the #4966 precedent this PR cites.

No issues found. Approving.

PR #5008, opened 2026-08-03
fix(daemon): detect stale launchd env before a plain restart, never silent

### LINKED ISSUE #4995: daemon: supervised restart silently keeps stale launchd EnvironmentVariables — env changes require bootout/bootstrap
## Context (2026-08-03, robb-studio)

Operator added `LOOM_RUNTIME_JUDGE=codex` to the daemon plist's `EnvironmentVariables` and ran the supervised `loom-daemon restart` (#4054). The relaunch verified fine (new pid) — but role ticks kept admitting `runtime=claude`: launchd only reads `EnvironmentVariables` at **bootstrap**, not on a KeepAlive respawn, so the relaunched daemon inherited the OLD env. The restart primitive's own message ("the relaunched daemon re-reads the same launchd plist, so it comes back with exactly its start flags") is true for flags but actively misleading for env. Fix required `launchctl bootout gui/<uid>/<label> && launchctl bootstrap gui/<uid> <plist>`, after which admission flipped to `runtime=codex source=role-environment`.

Cost: ~15 minutes of a confused verification loop; the failure is silent (no error anywhere — the env var is simply not there).

## Suggested fix (either or both)

- **Detect**: on restart, the daemon (or the restart CLI) compares the live process environment's `LOOM_*` values against the plist's `EnvironmentVariables` dict and prints a loud warning naming any drift + the exact bootout/bootstrap commands.
- **Document**: daemon-reference §Safe start/stop gains a "changing daemon environment variables" subsection stating that `restart` preserves env and bootout/bootstrap (or the start script, if it bootstraps) is the env-change path.

## Acceptance criteria

- [ ] Changing a plist env var and running the documented env-change path results in the new value being live, verifiably (e.g. `launchctl print` shows it and admission logs reflect it).
- [ ] The stale-env case is either impossible via the normal path or loudly detected — never silent.

## Affected Files

- `loom-daemon/src/cli/restart.rs` - `handle_restart_command` is the plain-restart entry point; candidate site for a pre-restart drift check/warning (Detect option)
- `defaults/scripts/lib/daemon-env-harvest.sh` - existing `harvest_plist_env()` helper already reads a plist's `LOOM_*`/token env; reusable for a drift comparison rather than writing a new parser
- `defaults/docs/daemon-reference.md` (~line 4744, "Supervised restart primitive (#4054)" section) - the "re-reads the same plist... exactly its start flags/env" line needs a caveat or a new "changing daemon environment variables" subsection (Document option) stating that a plain `restart` preserves the currently-*bootstrapped* env regardless of on-disk plist edits, and that `launchctl bootout`/`bootstrap` (or re-running `loom-daemon-start.sh`, which performs the equivalent render+bootstrap) is the actual env-change path
- `defaults/scripts/cli/loom-daemon-start.sh` - reference only; this is the script that correctly re-renders + bootstraps the plist (the working env-change path today), useful as the "or the start script, if it bootstraps" alternative named in the Suggested Fix above

## Test Plan

- [ ] Manual verification: on a launchd host, hand-edit the installed plist's `EnvironmentVariables` (add/change a `LOOM_*` key), run `loom-daemon restart`, and confirm the new value is either (a) live per `launchctl print gui/<uid>/<label>` + daemon admission logs, or (b) the restart path loudly warns of the drift and names the bootout/bootstrap remediation
- [ ] Automated tests: add a case to `defaults/scripts/tests/test-loom-daemon-start.sh` or a new `restart.rs` unit test exercising the detect path (if implemented) against a plist/live-env mismatch fixture
- [ ] Edge cases: confirm a plist edit that matches the currently-bootstrapped env produces no false-positive warning; confirm the systemd path (which does read `Environment=` fresh via `daemon-reload`, per the launchd/systemd contract table in daemon-reference.md) is unaffected — this is a launchd-specific gap, not a systemd one

rjwalters:
## Summary
`loom-daemon restart` relies on launchd's `KeepAlive` relaunch, which reuses the job spec already bootstrapped into launchd's memory and never re-reads the on-disk plist. Hand-editing the installed plist's `EnvironmentVariables` and then running a plain restart previously kept the OLD value silently — no error anywhere. This adds a loud pre-restart drift check plus a documentation subsection naming the actual env-change path.

## Changes
- New `loom-daemon/src/launchd_env_drift.rs`: compares the live (bootstrapped) launchd job's `LOOM_*`/token environment (`launchctl print gui/<uid>/<label>`) against the on-disk plist's declared `EnvironmentVariables` (`plutil -convert json`), restricted to the same key allowlist `defaults/scripts/lib/daemon-env-harvest.sh::harvest_plist_env()` already uses. Best-effort/advisory only — every "nothing to compare" outcome (non-macOS, no launchd job registered, no on-disk plist, probe failure) collapses to `None`, never blocking the restart.
- `handle_restart_command` (`loom-daemon/src/cli/restart.rs`) calls this check before scheduling anything and prints a loud `WARNING:` naming every drifted key plus the exact `launchctl bootout`/`bootstrap` remediation when they disagree. Token values (`GH_TOKEN`/`GITEA_TOKEN`/`FORGE_TOKEN`) are fingerprinted (last 4 chars), never echoed in full.
- `defaults/docs/daemon-reference.md`: adds a "Changing daemon environment variables (#4995)" subsection under "Supervised restart primitive", and a caveat on the "Mechanism (macOS)" bullet correcting the previously-misleading "re-reads the same plist... exactly its start flags/env" claim (true for flags, false for env on a plain restart).

| Changing a plist env var and running the documented env-change path results in the new value being live, verifiably | ✅ | Documented path (`loom-daemon-start.sh`, or `bootout`+`bootstrap`) is unchanged — it already re-renders/re-bootstraps, so `launchctl print` + admission logs already reflect it; new "Changing daemon environment variables" doc subsection names both options explicitly |
| The stale-env case is either impossible via the normal path or loudly detected — never silent | ✅ | `check_launchd_env_drift()` runs on every plain `loom-daemon restart` on a launchd host and prints a `WARNING:` naming the drifted keys + exact remediation whenever the live and on-disk environments disagree; unit-tested end-to-end in `launchd_env_drift.rs` |

## Test Plan
- `cargo test -p loom-daemon --lib launchd_env_drift`: 16/16 new unit tests pass — parser fixtures (a real `launchctl print` capture, `plutil -convert json` output), the pure `env_drift` diff (changed/added/removed keys, agreement = no drift), token redaction (never leaks a raw value, still distinguishable), the rendered warning (names keys + exact `bootout`/`bootstrap` commands, never leaks a token), and the full orchestration wired against fake I/O boundaries (non-macOS skip, no launchd job, no on-disk plist, agreement, and the exact #4995 incident shape end-to-end).
- `cargo test --bin loom-daemon drain_reply_render_tests`: 5/5 pre-existing `restart.rs` tests still pass — confirms no regression to the drain-mode variants.
- `cargo check -p loom-daemon` and `cargo fmt -p loom-daemon` both clean; the crate's `.cargo/config.toml` rustflags run `-D clippy::all` (mirroring CI) on every check/test build, and both builds above completed without errors.
- Edge cases covered by unit tests: a plist edit matching the live env produces no warning (no false positive); non-macOS returns `None` immediately (systemd re-reads `Environment=` fresh via `daemon-reload`, so there is nothing to detect there — confirmed against the launchd/systemd contract table in daemon-reference.md).
- Manual on-host verification (hand-edit the plist + real `launchctl print`) is left as an operator follow-up — the unit tests pin the actual `launchctl print`/`plutil -convert json` wire formats via fixtures captured from a running host, and the OS-facing probes are isolated (`probe_launchctl_print`/`probe_plist_json`) behind the same seam the tests inject fakes through.

Closes #4995

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4860,8 +4860,11 @@ compares the live (bootstrapped) job's `LOOM_*`/token environment
 (`launchctl print gui/<uid>/<label>`) against the on-disk plist's declared
 `EnvironmentVariables` (`plutil -convert json`), and prints a loud `WARNING:`
 naming every drifted key plus the exact remediation, if they disagree
-(`loom_daemon::launchd_env_drift`). No warning means the two agree — not that
-the check didn't run; this is macOS/launchd-specific (systemd re-reads
+(`loom_daemon::launchd_env_drift`). The check is advisory-only and never blocks
+or fails the restart. No warning normally means the two agree — but it is also
+what you get when there was nothing to compare (no launchd job registered under
+this label, no on-disk plist) or a probe/parse failed, since every such outcome
+is deliberately silent; this is macOS/launchd-specific (systemd re-reads
 `Environment=` fresh via `daemon-reload` on every unit reload, so there is
 nothing to detect on that path).
 
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4825,10 +4825,13 @@ loom-daemon restart          # send RestartDaemon over the IPC socket
 - **Mechanism (macOS):** the plist uses `KeepAlive:{SuccessfulExit:true}` and the
   daemon exits `0` **only** for a `RestartDaemon` request, so launchd relaunches
   the job on that one clean exit and leaves it down on every other (SIGTERM 143,
-  SIGINT 130, crash non-zero). The relaunched process re-reads the same plist, so
-  it comes back with **exactly** its start flags/env — never wider. In-flight
-  sweeps survive (they are independent detached processes the daemon never cancels
-  on shutdown). Observable signature: a **new pid**.
+  SIGINT 130, crash non-zero). The relaunched process comes back with **exactly**
+  its start flags/env as launchd already has them bootstrapped in memory — **never
+  a fresh read of the on-disk plist file** (see "Changing daemon environment
+  variables" immediately below — this is not the widening-safe guarantee it
+  sounds like for a hand-edited plist). In-flight sweeps survive (they are
+  independent detached processes the daemon never cancels on shutdown).
+  Observable signature: a **new pid**.
 - **Supervision proof:** `loom-daemon-start.sh` bakes `LOOM_DAEMON_SUPERVISOR=launchd`
   into the plist, and the daemon ends its process for a restart **only** when that
   var is present. On an unsupervised host (nohup / Linux / `--foreground`) it
@@ -4839,6 +4842,41 @@ loom-daemon restart          # send RestartDaemon over the IPC socket
   the `exec` (Option 2) and detached-helper (Option 3) alternatives were rejected,
   and the Curator's exit-code-race finding — is in
   `docs/design/supervised-restart-primitive.md`.
+
+#### Changing daemon environment variables (#4995)
+
+**A plain `loom-daemon restart` preserves the currently-*bootstrapped*
+environment, regardless of any on-disk plist edit.** launchd's `KeepAlive`
+relaunch reuses the job spec already held in launchd's own memory since the
+last `bootstrap` — it does not re-read the plist file from disk. Hand-editing
+the installed plist's `EnvironmentVariables` (e.g. adding `LOOM_RUNTIME_JUDGE`)
+and then running `loom-daemon restart` silently keeps the OLD value: the
+relaunched process comes back with exactly what launchd already had bootstrapped,
+not what the file now says. There is no error anywhere — the new/changed env var
+is simply not there post-restart.
+
+`loom-daemon restart` detects this automatically: before scheduling anything, it
+compares the live (bootstrapped) job's `LOOM_*`/token environment
+(`launchctl print gui/<uid>/<label>`) against the on-disk plist's declared
+`EnvironmentVariables` (`plutil -convert json`), and prints a loud `WARNING:`
+naming every drifted key plus the exact remediation, if they disagree
+(`loom_daemon::launchd_env_drift`). No warning means the two agree — not that
+the check didn't run; this is macOS/launchd-specific (systemd re-reads
+`Environment=` fresh via `daemon-reload` on every unit reload, so there is
+nothing to detect on that path).
+
+**The actual env-change path** is either of:
+
+```bash
+# Option A: re-render + re-bootstrap in one step (the normal path)
+./.loom/scripts/cli/loom-daemon-start.sh
+
+# Option B: bootout the stale bootstrapped job, then bootstrap the edited plist
+launchctl bootout gui/$(id -u)/com.rjwalters.loom-daemon
+launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.rjwalters.loom-daemon.plist
+```
+
+Both make launchd re-read the plist file, which a plain `restart` never does.
 
 #### Scheduled drain-and-restart (`--drain`, #4090)
 
--- a/loom-daemon/src/cli/restart.rs
+++ b/loom-daemon/src/cli/restart.rs
@@ -32,6 +32,20 @@ pub(crate) async fn handle_restart_command(
         std::process::exit(1);
     }
 
+    // Issue #4995: on a launchd host, warn LOUDLY — before scheduling
+    // anything — if the currently-bootstrapped launchd job's environment has
+    // drifted from what the on-disk plist now declares. A plain restart's
+    // `KeepAlive` relaunch reuses the already-bootstrapped job spec, so a
+    // hand-edited plist would otherwise be picked up silently — never.
+    // Best-effort: returns `None` (no output) on every non-launchd /
+    // nothing-to-compare / no-drift outcome, and never blocks or fails the
+    // restart itself.
+    if let Some(warning) = loom_daemon::launchd_env_drift::check_launchd_env_drift() {
+        eprintln!();
+        eprintln!("{warning}");
+        eprintln!();
+    }
+
     // Drain-mode variants (Issue #4090) speak `DrainAndRestartDaemon` /
     // `AbortDrain` and expect a `DaemonDrain` reply; the plain restart keeps its
     // #4054 `RestartDaemon` / `DaemonRestart` contract byte-for-byte.
--- a/loom-daemon/src/launchd_env_drift.rs
+++ b/loom-daemon/src/launchd_env_drift.rs
@@ -1,0 +1,582 @@
+//! Detect drift between a launchd job's actual live (bootstrapped)
+//! `EnvironmentVariables` and what the on-disk plist currently declares
+//! (Issue #4995).
+//!
+//! # Why
+//!
+//! launchd's `KeepAlive:{SuccessfulExit:true}` relaunch reuses the job spec
+//! already held in launchd's own memory since the last `bootstrap` — it does
+//! **not** re-read the plist file from disk. A plain `loom-daemon restart`
+//! only sends `RestartDaemon` over the daemon's own IPC socket
+//! ([`crate::cli`]'s equivalent in the binary crate); it never touches
+//! launchd's job registration. So hand-editing the installed plist's
+//! `EnvironmentVariables` and then running a plain restart silently keeps the
+//! OLD (already-bootstrapped) environment — the relaunched process comes
+//! back with exactly what launchd already had, not what the file now says.
+//! This was entirely silent before this module: no error anywhere, the env
+//! var is simply not there post-restart.
+//!
+//! The fix (matching the issue's "Detect" option): before scheduling a
+//! restart, compare the *live* environment launchd actually holds for the
+//! running job (`launchctl print gui/<uid>/<label>`'s `environment = { ... }`
+//! block — the ground truth for what a `KeepAlive` respawn will use) against
+//! the *on-disk* plist's declared `EnvironmentVariables` (via `plutil
+//! -convert json`). Any disagreement, restricted to the `LOOM_*`/token keys
+//! `loom-daemon-start.sh` actually manages (mirrors
+//! `defaults/scripts/lib/daemon-env-harvest.sh`'s `harvest_plist_env()`
+//! allowlist), is surfaced as a loud warning naming the exact remediation —
+//! `launchctl bootout` + `bootstrap` (or re-running `loom-daemon-start.sh`,
+//! which performs the equivalent render+bootstrap).
+//!
+//! This is a macOS/launchd-specific gap: the systemd path re-reads
+//! `Environment=` fresh via `daemon-reload` on every unit reload, so there is
+//! nothing to detect there (see the launchd/systemd contract table in
+//! `defaults/docs/daemon-reference.md`).
+//!
+//! Every public entry point here is advisory-only and best-effort: any
+//! failure (not on macOS, launchd job not found — e.g. an unsupervised nohup
+//! host, no on-disk plist to compare against, a `launchctl`/`plutil` binary
+//! missing, unparseable output) collapses to `None` — "nothing to say" — and
+//! must never fail or block the restart itself.
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+use crate::daemon_install_state::DEFAULT_LAUNCHD_LABEL;
+use crate::main_health_gate::run_capture_with_timeout;
+
+/// Bound on the `launchctl print` / `plutil` probes — both are local,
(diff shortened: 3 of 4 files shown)


rjwalters:

`Rust Unit Tests` and `Rust OTLP Feature Build & Test` are both failing on CI, and I reproduced the failure locally (`cargo test -p loom-daemon --lib launchd_env_drift`):

```
---- launchd_env_drift::tests::orchestration_surfaces_a_real_drift_end_to_end stdout ----
thread '...' panicked at loom-daemon/src/launchd_env_drift.rs:576:30:
a real drift must never be silent
```

**Root cause**: `check_launchd_env_drift_with()` (`loom-daemon/src/launchd_env_drift.rs:253-264`) is the dependency-injected function the orchestration tests use to exercise `label_fn`/`launchctl_print`/`plutil_json`/`plist_path_fn` fixtures without a real macOS host. But its `if !cfg!(target_os = "macos") { return None; }` guard lives *inside* that same injectable function — so on any non-macOS runner (this repo's CI runs on Linux) the function short-circuits to `None` before ever calling the injected fakes. Every orchestration test that expects `None` passes vacuously (the macOS gate alone produces `None`, independent of the fixtures), but `orchestration_surfaces_a_real_drift_end_to_end` expects `Some(warning)` and fails unconditionally on Linux.

The PR description's "16/16 new unit tests pass" claim appears to have been verified on a macOS host, where the `cfg!` guard doesn't trigger — but this repo's CI (and this evaluation) runs on Linux, where it always will.

**Fix**: move the `cfg!(target_os = "macos")` check out of `check_launchd_env_drift_with` and into `check_launchd_env_drift()` (the real, non-injectable entry point), so the injectable function is testable on any host regardless of the target OS:

```rust
pub fn check_launchd_env_drift() -> Option<String> {
    if !cfg!(target_os = "macos") {
        return None;
    }
    check_launchd_env_drift_with(resolve_label, probe_launchctl_print, probe_plist_json, plist_path_for)
}

fn check_launchd_env_drift_with(...) -> Option<String> {
    // no OS gate here — pure orchestration logic, fully testable on any host
    ...
}
```

Everything else in this PR looks solid — the parsing/diff/redaction logic is well-isolated and well-tested, the docs update accurately describes the actual drift/remediation mechanism, and the call site in `restart.rs` is correctly placed before scheduling. This is a one-line relocation once identified.

## Test Execution

1. `cargo test -p loom-daemon --lib launchd_env_drift` — ❌ Executed: 15/16 pass, 1 fails (`orchestration_surfaces_a_real_drift_end_to_end`) — see above, matches CI's `Rust Unit Tests` failure
2. `cargo test --bin loom-daemon drain_reply_render_tests` — not re-run (unrelated to the failure)
3. `cargo check -p loom-daemon` / `cargo fmt -p loom-daemon` — not blocking, CI's `Rust Linting & Build Check` passes
4. Manual on-host verification — ⚠️ Skipped: requires macOS host, correctly flagged as an operator follow-up in the PR description

I'll re-evaluate once CI passes.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T14:17:58Z) — standing down without reclaiming. Not stomping.

rjwalters:

### CI is currently red — this is the blocking issue

`gh pr checks 5008` shows two failing jobs right now:

- **Rust Unit Tests** (`cargo nextest run --workspace`)
- **Rust OTLP Feature Build & Test** (`cargo nextest run -p loom-daemon --features otlp`)

Both fail on the *same* new test:

```
FAIL [0.006s] loom-daemon launchd_env_drift::tests::orchestration_surfaces_a_real_drift_end_to_end
thread '...' panicked at loom-daemon/src/launchd_env_drift.rs:576:30:
a real drift must never be silent
```

**Root cause**: the `cfg!(target_os = "macos")` short-circuit lives inside
`check_launchd_env_drift_with` (the parameterized, fake-injectable function the
tests call directly), not only in the public `check_launchd_env_drift()`
wrapper:

```rust
fn check_launchd_env_drift_with(
    label_fn: impl Fn() -> String,
    launchctl_print: impl Fn(&str) -> Option<String>,
    plutil_json: impl Fn(&Path) -> Option<String>,
    plist_path_fn: impl Fn(&str) -> Option<PathBuf>,
) -> Option<String> {
    if !cfg!(target_os = "macos") {   // <-- fires on CI's Linux runners too
        return None;
    }
    ...
}
```

CI runs on Linux (`ubuntu` runner) → this branch fires for every
`orchestration_*` test regardless of the injected fake I/O closures, so
`orchestration_surfaces_a_real_drift_end_to_end` (which expects `Some(warning)`)
fails, and the other `orchestration_returns_none_*` tests pass, but for the
wrong reason (OS gate, not the fake behavior they're supposed to exercise).

This matches what I reproduced locally: on macOS all 16 tests pass (confirmed
via `cargo test -p loom-daemon --lib launchd_env_drift`, matching the PR's
"16/16" test-plan claim), but the PR body's test plan didn't catch this because
it was only run on a macOS host — CI's Linux runners are where it breaks.

**Suggested fix**: move the `cfg!(target_os = "macos")` gate to
`check_launchd_env_drift()` only (the real, non-parameterized entry point), and
let `check_launchd_env_drift_with` stay OS-agnostic so its unit tests exercise
the actual orchestration logic (label -> probe -> compare -> render) on any
host, matching the intent stated in its own doc comment ("parameterized over
its I/O boundaries so the orchestration ... is unit-testable ... without a real
launchd host"). Add a separate, dedicated test (or keep
`orchestration_returns_none_on_non_macos`, which already calls the *public*
`check_launchd_env_drift()`) to cover the actual OS gate.

### Everything else here is solid

- Design matches the issue's "Detect" option precisely: compares the live
  bootstrapped launchd env (`launchctl print`) against the on-disk plist
  (`plutil -convert json`), restricted to the same `LOOM_*`/token allowlist
  `daemon-env-harvest.sh::harvest_plist_env()` uses, and is wired into
  `handle_restart_command` as a loud, non-blocking `eprintln!` before
  scheduling anything.
- Advisory/best-effort design is correct: every "nothing to compare" path
  (non-macOS, no launchd job, no on-disk plist, probe/parse failure) collapses
  to `None` and never touches the restart's own control flow or exit code.
- Token redaction (`display_value`) is correct — only last-4-chars shown for
  `GH_TOKEN`/`GITEA_TOKEN`/`FORGE_TOKEN`, plain values shown for `LOOM_*` — and
  is unit-tested (`token_values_are_redacted_but_still_distinguishable`,
  `warning_names_the_exact_remediation_and_never_leaks_a_raw_token`).
- No shell injection risk: `launchctl`/`plutil` are invoked via `Command` with
  args as a slice, never through a shell string.
- `cargo clippy -p loom-daemon --lib` is clean locally (crate rustflags run
  `-D clippy::all`), matching the PR's claim.
- Doc update (`defaults/docs/daemon-reference.md`) correctly retracts the
  previously-misleading "re-reads the same plist... exactly its start
  flags/env" claim and adds a clear "Changing daemon environment variables
  (#4995)" subsection naming both remediation paths.
- Minor, non-blocking doc nit: "No warning means the two agree — not that the
  check didn't run" overstates it slightly — per the code's own advisory
  design, "no warning" also covers unsupervised hosts, a missing on-disk
  plist, or a `launchctl`/`plutil` probe failure, not just true agreement.
  Worth a one-line caveat, but not blocking.

### Verdict

Please fix the `cfg!` gate placement (or otherwise make the orchestration
tests actually exercise the injected fakes independent of the CI runner's OS),
push, and confirm CI is green. Everything else here is ready — this is a
one-line-class fix, not a design problem.

rjwalters:
Judge feedback addressed — CI is green.

**Blocking issue (fixed): `cfg!(target_os = "macos")` gate was scoped to the wrong function.**

Exactly as diagnosed. The gate lived inside `check_launchd_env_drift_with` — the parameterized, fake-injectable function the unit tests call directly — so on Linux runners it fired unconditionally for every `orchestration_*` test regardless of the injected fakes. `orchestration_surfaces_a_real_drift_end_to_end` got `None` and panicked with "a real drift must never be silent". All 16 passed on macOS, which is why it slipped through.

Fix applied as suggested: the gate now lives only in the public `check_launchd_env_drift()` wrapper — the sole caller that performs real launchd I/O — leaving `check_launchd_env_drift_with` OS-agnostic, matching its own doc comment's stated intent. `orchestration_returns_none_on_non_macos` still covers the gate, since it exercises the public wrapper. I added doc comments on **both** sides explaining why the gate sits where it does, so it does not migrate back inward.

**Linux path simulated locally, not just reasoned about.** I temporarily forced the gate to fire unconditionally (`if true || !cfg!(target_os = "macos")`) and re-ran the suite: all 16 tests still passed, including the exact test that failed on CI. Then reverted the simulation.

**Doc nit (also fixed):** `defaults/docs/daemon-reference.md` no longer claims "no warning means the two agree". It now notes the check is advisory-only and that a silent result also covers "nothing to compare" (no registered job / no on-disk plist) and probe/parse failures.

**Verification:** all checks pass on `b38dcbbd`, including the two that were red — Rust Unit Tests (3m49s) and Rust OTLP Feature Build & Test (3m41s). `cargo fmt --check` clean.

Scope: 2 files, +23/-9 — only the file the Judge named and the doc line quoted. No drive-by changes.

**Housekeeping note:** my `loom:treating` claim was reaped by the daemon's stale-claim backstop at 15:29Z (~63m after claiming — my local build-gate run overran the 60m `LOOM_STALE_TREATING_MINUTES` window without an intervening comment). I verified before completing that this was the backstop and not a takeover: PR head is my own commit, no comments landed since the claim, and no competing verdict label appeared. I also cleared the now-stale `loom:ci-failure` label since CI is green.

Ready for re-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T15:43:46Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Second pass — the prior blocking issue (CI red on `orchestration_surfaces_a_real_drift_end_to_end`, caused by the `cfg!(target_os = "macos")` gate living in the fake-injectable `check_launchd_env_drift_with` instead of the real-I/O `check_launchd_env_drift` wrapper) is fixed exactly as suggested: the gate now lives only on the public wrapper, and `check_launchd_env_drift_with` stays OS-agnostic so its fixture-driven tests actually exercise the orchestration on Linux CI.

Verified:
- `cargo test -p loom-daemon --lib launchd_env_drift` — 16/16 pass locally (Linux), confirming the fix
- `gh pr checks 5008` — all CI checks pass (Rust Unit Tests, Rust OTLP Feature Build & Test, Rust Linting & Build Check, etc.)
- `mergeStateStatus` — CLEAN
- The prior non-blocking doc nit ("no warning means the two agree") is also now correctly qualified in `daemon-reference.md` ("but it is also what you get when there was nothing to compare... or a probe/parse failed")
- Tracked-key allowlist in `is_tracked_key` matches `daemon-env-harvest.sh::harvest_plist_env()` 1:1 (LOOM_* + GH_TOKEN/GITEA_TOKEN/FORGE_TOKEN, excluding LOOM_DAEMON_SUPERVISOR)
- Token redaction is correct and unit-tested; no shell-injection risk (`Command` invoked with arg slices, never a shell string)
- PR body correctly uses `Closes #4995`

## Test Execution

1. `cargo test -p loom-daemon --lib launchd_env_drift` (16/16) — ✅ Executed: all pass locally
2. `cargo test --bin loom-daemon drain_reply_render_tests` (5/5, no regression) — ⏭️ Skipped: exceeded local time budget, but covered by CI's green "Rust Unit Tests"/"Rust Linting & Build Check" runs
3. `cargo check`/`cargo fmt` clean — ✅ Covered by CI's green "Rust Linting & Build Check" (crate rustflags run `-D clippy::all`)
4. Manual on-host launchd verification — ⚠️ Skipped: requires a real macOS/launchd host (correctly flagged in the PR as an operator follow-up); the unit tests pin the real `launchctl print`/`plutil -convert json` wire formats via captured fixtures

Nice fix — good root-cause diagnosis and a clean, well-tested advisory-only implementation.

rjwalters:
**Champion Auto-Merge**

- Merge-risk judgment passed: diff is 653+/4- across 4 files, but the load-bearing change is a self-contained new module (`launchd_env_drift.rs`, 593 lines) plus a 14-line advisory-only, non-blocking call site in `restart.rs` and a 1-line module registration in `lib.rs`; no credential-handling, guard-hook, installer, or merge/worktree-script surfaces touched; Judge's approval cites specifics (tracked-key allowlist match against `daemon-env-harvest.sh`, token redaction, no shell injection, CI green); fully revertible with no data migration or published artifact
- Diff size: 657 lines (+653/-4) — informational, not a gate
- No critical files modified
- No merge conflicts
- All CI checks passing

`git revert <commit-sha>`

---
*Automated by Champion role*

PR #5019, opened 2026-08-03
docs(observability): record per-host telemetry redundancy trade study

### LINKED ISSUE #4999: observability: 3x per-host telemetry redundancy (~78k rows/90d) — trade study, and do not dedupe away the gap tolerance
## Observation (deliberately filed as a question, not a fix request)

Telemetry collection is per-host redundant: **3 hosts x ~13 accounts every 300s**
lands roughly **78k rows per 90d retention window** where ~26k would carry the
same information, because every host independently reports the same shared
account pool.

## The counter-argument, recorded up front

**Do not "optimize" this without answering it.** That redundancy is exactly what
kept fleet telemetry alive when `loom-worker-1` went dark for ~80 minutes on
2026-08-01: the other two hosts were still reporting the same accounts, so the
account-level series had no gap. Collapsing to a single reporter reintroduces a
single point of failure on the one signal an operator needs *while a host is
dying*.

So the honest framing is a trade study, not a bug:

- **Keep as-is**: 3x storage, gap-tolerant, no coordination.
- **Elect one reporter**: ~1x storage, needs leader election + failover, and the
  failover window is a telemetry gap.
- **Deduplicate at write/read time**: keep all 3 reporters, collapse identical
  account-state rows server-side (last-writer-wins per account per bucket), so
  redundancy costs bandwidth but not storage.

Option 3 looks like it preserves the survivability property that made the
redundancy valuable, at storage cost close to option 2. That is a hypothesis, not
a conclusion.

## What would make this decidable

- [ ] Actual row counts + storage footprint per 90d window, measured rather than
      estimated.
- [ ] Whether current storage is anywhere near a limit that matters (if not, the
      correct outcome is "close as won't-fix, keep the redundancy").
- [ ] If dedup is pursued: confirmation that a collapsed row still lets an
      operator distinguish "account exhausted" from "no host reported".

## Suggested outcome

If storage is not actually pressured, **close this and keep the redundancy** —
but close it with the reasoning recorded, so the next person who notices 78k rows
does not silently dedupe away the survivability property.

## Provenance

Surfaced during a `/repo:followups` pass on 2026-08-01 that was interrupted before
filing. Related surface: `.loom/docs/observability.md`, the `observability` config
block, and `loom-daemon/src/metrics_collector.rs`.

## Affected Files

- `dashboard/migrations/0001_init.sql` - `records` table schema; any row-count/storage measurement queries against this table, and where a dedup migration would land if option 3 is pursued
- `dashboard/src/retention.ts` - the hourly retention cron; where a measured "rows near a limit" check could be added, and where read-time or write-time collapse would hook in if pursued
- `dashboard/src/telemetry.ts` - ingest path that writes one row per record per host today (dedup-at-write-time would change this)
- `loom-daemon/src/observability/collector.rs`, `queue.rs`, `exporter.rs`, `sender.rs` - daemon-side emission cadence (`flushIntervalSecs`); relevant if "elect one reporter" is pursued, since failover/leader-election would live here
- `.loom/docs/observability.md`, `.loom/docs/telemetry-schema.md` - update with whichever outcome is decided (keep-as-is / elect-one-reporter / dedup) and the reasoning, per the issue's own "close with reasoning recorded" suggested outcome

## Test Plan

- [ ] Measurement: query `records` (or the production D1 equivalent) for actual row counts and storage footprint over a representative window; compare against the ~78k/90d estimate above
- [ ] Decision gate: confirm whether current storage is anywhere near a limit that matters — if not, the suggested outcome above is close as won't-fix with reasoning recorded, not a code change
- [ ] If dedup (option 3) is pursued: add a test asserting a collapsed row still lets an operator distinguish "account exhausted" from "no host reported" for the same account/bucket
- [ ] If dedup is pursued: verify the `loom-worker-1` 2026-08-01 outage scenario (one host dark, others still reporting) still produces no gap in the account-level series after collapse
- [ ] Manual verification: N/A unless a code change is chosen — the primary deliverable is the trade study itself

rjwalters:
## Summary

Issue #4999 was filed as a deliberate trade study, not a fix request: is the
3x per-host telemetry redundancy (`tokens.snapshot` + `host.health`, each
independently sampled and pushed by every host on its own 300s interval)
worth deduplicating?

**Measurement (per the issue's own decision gate):**
- `records` is bounded by `RETENTION_DAYS=90` and `MAX_RECORDS=500,000`
  (`dashboard/wrangler.toml`, `dashboard/src/retention.ts`).
- Per-host periodic sampling produces ~78k rows/90d **per record kind** on a
  3-host fleet (confirms the issue's estimate), ~156k rows/90d for both kinds
  combined — about 31% of the configured cap. Not anywhere near a limit that
  matters.
- The redundancy is the exact property that kept the account-level series
  gap-free when `loom-worker-1` went dark for ~80 minutes on 2026-08-01 (per
  the issue's own provenance note).

**Decision: keep the current per-host independent-reporting design as-is.**
No code change — storage isn't pressured, so trading away the survivability
property for a saving that isn't needed would be the wrong call. Recorded the
reasoning in `.loom/docs/telemetry-schema.md` and pointed to it from
`.loom/docs/observability.md`, per the issue's own suggested outcome ("close
this and keep the redundancy — but close it with the reasoning recorded").

## Test plan

- [x] Measurement: derived row-count math from the actual configured
      `SNAPSHOT_INTERVAL` (300s, `loom-daemon/src/observability/mod.rs`),
      `RETENTION_DAYS`/`MAX_RECORDS` (`dashboard/wrangler.toml`), matching the
      issue's ~78k/90d estimate.
- [x] Decision gate: confirmed current storage is not near the cap (~31%) —
      per the issue, this means "close as won't-fix with reasoning recorded,"
      not a code change.
- [x] Docs updated with the decision and reasoning in both files named in the
      issue's Affected Files list.

Closes #4999

### DIFF
--- a/.loom/docs/observability.md
+++ b/.loom/docs/observability.md
@@ -146,6 +146,13 @@ provisioning, verifying telemetry lands — is
 This is **your own infrastructure**; nothing in Loom points at a shared
 backend by default.
 
+Per-host reporting is deliberately redundant — every host independently
+emits `tokens.snapshot` / `host.health` for the full account pool it can see,
+at real storage cost but with no single point of failure. This was evaluated
+as a trade study (issue #4999) and kept as-is: see
+[`.loom/docs/telemetry-schema.md`](telemetry-schema.md#per-host-reporting-redundancy-why-3x-storage-is-intentional-issue-4999)
+for the reasoning.
+
 ## 5. Authenticated vs. public: two views, one redaction policy
 
 Every query route exists twice — `/api/*` (authenticated, full detail) and
--- a/.loom/docs/telemetry-schema.md
+++ b/.loom/docs/telemetry-schema.md
@@ -348,6 +348,43 @@ its registered repos) rather than as an unexplained gap.
   `redactManagedRepos`). The authenticated view always sees every entry in
   full, including private slugs.
 
+## Per-host reporting redundancy (why 3x storage is intentional, Issue #4999)
+
+Every host in a fleet independently samples and pushes `tokens.snapshot` and
+`host.health` on its own `SNAPSHOT_INTERVAL` (5 minutes —
+`loom-daemon/src/observability/mod.rs`), each reporting the **same shared
+account pool** every host in the fleet can see. On a 3-host fleet this means
+each of those two record kinds independently accumulates roughly
+`3 hosts x (90d / 5min) ~= 78k rows` per 90-day retention window — about 3x
+what a single elected reporter would produce for the same information.
+
+This redundancy was raised as a possible optimization (#4999) and
+investigated as a trade study rather than assumed to be a bug:
+
+- **The redundancy is load-bearing.** When `loom-worker-1` went dark for ~80
+  minutes on 2026-08-01, the other two hosts kept reporting the same account
+  pool, so the account-level series had no gap. A single elected reporter
+  would have produced exactly that gap during the one window an operator most
+  needs the series to stay continuous — while a host is dying.
+- **Storage is nowhere near a limit that matters.** `records` is already
+  bounded by both an age cutoff (`RETENTION_DAYS`, default 90) and a hard row
+  cap (`MAX_RECORDS`, default 500,000 — `dashboard/src/retention.ts`). Even
+  counting *all* per-host periodic sampling (`tokens.snapshot` + `host.health`
+  together, ~156k rows/90d across 3 hosts) alongside per-sweep lifecycle
+  records, total volume sits well under a third of the configured cap. There
+  is no pressure to relieve.
+
+**Decision: keep the current per-host independent-reporting design.**
+Electing a single reporter (or deduplicating identical rows at write/read
+time) would trade away a real survivability property for a storage saving
+that isn't currently needed. Revisit only if `MAX_RECORDS` eviction starts
+firing routinely (sustained growth actually approaching the cap) — at that
+point, write-time deduplication that preserves per-account distinguishability
+(still being able to tell "account exhausted" from "no host reported" for a
+given account/bucket) is the option to pursue first, since electing a single
+reporter reintroduces the exact single-point-of-failure this trade study
+found unacceptable.
+
 ## Persistence & read surface (`sweep.outcome`, Issue #4704)
 
 The daemon durably records one `sweep.outcome` [`TelemetryEnvelope`] per


rjwalters:
## Judge review: APPROVE

Verified this as a legitimate "close as won't-fix, reasoning recorded" resolution per the issue's own explicit suggested outcome and this repo's "Issues Are Suggestions" autonomy — not a dodge of real work.

**Factual claims checked against source:**
- `SNAPSHOT_INTERVAL = Duration::from_secs(5 * 60)` in `loom-daemon/src/observability/mod.rs:108`, threaded into `collector::spawn_task` → confirmed each tick fires `sample_snapshots()` (`collector.rs:390-401`), which pushes exactly one `TokensSnapshot` envelope and one `HostHealth` envelope per host per tick. Matches the doc's "each independently sampled ... on its own 5-minute interval."
- `RETENTION_DAYS = "90"` and `MAX_RECORDS = "500000"` confirmed in `dashboard/wrangler.toml:136-137`, with matching defaults (`parsePositiveInt(env.RETENTION_DAYS, 90)`, `parsePositiveInt(env.MAX_RECORDS, 500_000)`) in `dashboard/src/retention.ts:46-47`.

**Math checked:**
- `90d / 5min = 25,920` samples/host/kind → `× 3 hosts = 77,760 ≈ 78k rows/90d` per kind (matches the issue's own estimate).
- Both kinds combined: `155,520 ≈ 156k rows/90d`.
- `155,520 / 500,000 = 31.1%` — matches the doc's "about 31%."

All arithmetic and cited source values are accurate.

**Doc placement:** Both files named in the issue's own "Affected Files" list were updated. The new `telemetry-schema.md` section header slug matches the `observability.md` cross-link anchor exactly (`#per-host-reporting-redundancy-why-3x-storage-is-intentional-issue-4999`), verified against GitHub's heading-slug algorithm. Placement doesn't break existing section numbering/structure in either file. `defaults/docs/` mirrors are in sync (Docs/Defaults Parity Check passes).

**One minor, non-blocking nitpick:** the sentence "counting ... alongside per-sweep lifecycle records, total volume sits **well under** a third of the configured cap" is a bit generous — the periodic-only figure (31.1%) is already close to a third (33.3%), and adding unquantified per-sweep records on top only pushes it higher, not lower. Technically still true (31.1% < 33.3%) but "well under" overstates the margin slightly. Not worth a doctor pass for a docs-only trade-study PR — flagging for awareness only.

**Legitimacy of the resolution:** This directly implements the issue's own decision gate ("if storage isn't anywhere near a limit that matters, the correct outcome is close as won't-fix, keep the redundancy") and its own suggested outcome text. The trade study's investigation (measuring actual row counts against actual configured caps, and identifying the real survivability property from the 2026-08-01 outage) is exactly the kind of rigor the issue asked for — this is legitimate issue resolution, not builder avoidance.

**CI / mergeability:** `mergeStateStatus: CLEAN`. All applicable checks pass (`AGENTS.md Sync Check`, `CLAUDE.md Line Budget`, `Detect Changes`, `Docs/Defaults Parity Check`); the code-path-gated checks (Rust, Dashboard, Installer, Shell, Codex) correctly skip for a docs-only diff.

Approving.

Issue #5011, opened 2026-08-03
[Epic #4990] Phase 2: secrets-gated platform signing for release artifacts


rjwalters:
**Epic**: #4990 - release: prebuilt daemon artifacts + artifact-based self-update, with secrets-gated platform signing (public mechanism, private identity)
**Phase**: 2 of 4
**Phase Goal**: macOS Developer ID sign (+ optional notarize) for the artifacts Phase 1 (#5003) now builds, gated entirely on optional Actions secrets so the public repo ships mechanism only — never identity.

## Description

Extend the release workflow from Phase 1 (prebuilt artifacts + checksums, `#5003`, merged) with an optional signing step:

- **macOS Developer ID signing**: sign the `aarch64-apple-darwin` artifact if a Developer ID cert (p12) + team ID secret pair is present in Actions secrets. Absent secrets ⇒ skip signing cleanly, ship the unsigned artifact + checksum (already the Phase 1 behavior) — do not fail the release.
- **Optional notarization**: if a notary API key secret is present, notarize after signing. Note from the epic: `curl`/`gh release download` set no quarantine attribute on this distribution path, so notarization is belt-and-suspenders here, not required for the fleet's own update path — only relevant if artifacts are ever browser-downloaded.
- **Optional Linux signing**: if a cosign key secret is present, sign the Linux artifacts (`x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`) similarly.
- **Loud skip, not silent**: when the relevant secret(s) are absent, the workflow run summary must explicitly say signing was skipped (and why) — per the epic's design principle 2, this is asserted, not just implied by the checksum-only output.
- **No identity in tracked files**: cert references, team IDs, and org names live only in Actions secrets, never in workflow YAML or any tracked file — this is enforced by the inverted acceptance criterion below.

Follows the `dashboard-deploy.yml` precedent already in this repo: public mechanism (committed generic template), private identity (secrets overlay).

## Acceptance Criteria

- [ ] Release workflow signs the macOS artifact when Developer ID cert + team ID secrets are present; skips cleanly (loud skip in run summary) when absent.
- [ ] Optional notarization runs after signing when a notary API key secret is present.
- [ ] Optional cosign signing for Linux artifacts when a cosign key secret is present.
- [ ] The update path (still Phase 3, not yet implemented) will be able to verify the checksum unconditionally and the platform signature when present — this phase's output format must support that verification, even though the verifier itself lands in Phase 3.
[...]

Issue #5003, opened 2026-08-03
[Epic #4990] Phase 1: release workflow builds prebuilt loom-daemon artifacts + checksums


rjwalters:
**Epic**: #4990 - release: prebuilt daemon artifacts + artifact-based self-update, with secrets-gated platform signing (public mechanism, private identity)
**Phase**: 1 of 4
**Phase Goal**: A release workflow builds `loom-daemon` for all target platforms and uploads artifacts + checksums to the GitHub Release, with no signing yet (that's Phase 2).

## Description

Add a release workflow, triggered on GitHub Release creation (matching the existing release convention per `/repo:release`), that cross-compiles `loom-daemon` for:

- `aarch64-apple-darwin`
- `x86_64-unknown-linux-gnu`
- `aarch64-unknown-linux-gnu`

and uploads the resulting binaries plus SHA-256 checksums as Release assets. This is the foundation phase — no signing, no self-update consumption yet. It ships value standalone: a host can already fetch a verified-by-checksum artifact by hand.

Per the epic's design principles: local build via `loom-daemon-update.sh` remains the default/fallback path in this phase; this issue only adds the artifact-producing side.

## Acceptance Criteria

- [ ] A new GitHub Actions workflow builds `loom-daemon` for all three target triples on Release creation
- [ ] Each artifact is uploaded to the Release along with a `.sha256` checksum file
- [ ] Workflow failure for any one platform doesn't silently drop that platform's artifact (loud failure, not partial silent success)
- [ ] No macOS/Linux signing identity, cert, or org-specific identifier appears in any tracked file (this phase ships unsigned artifacts only)
- [ ] Documented in a short note (README or workflow comment) of what artifacts a Release now contains

## Dependencies

Part of Epic #4990. This is a Phase 1 issue with no blocking dependencies.

---
*Created by Champion from Epic #4990*

rjwalters:
**Champion Review: APPROVED**

This issue has been evaluated and promoted to `loom:issue` status. All quality criteria passed:

- **Clear problem statement**: bounded to a single workflow — cross-compile `loom-daemon` for 3 target triples on Release creation, upload artifacts + checksums, no signing.
- **Technical feasibility**: standard Rust cross-compilation in GitHub Actions; matches the existing `/repo:release` convention (trigger on Release creation).
- **Implementation clarity**: concrete target triples, explicit no-silent-partial-failure requirement, and an explicit no-signing-material guardrail for this phase.
[...]

Issue #4990, opened 2026-08-03
release: prebuilt daemon artifacts + artifact-based self-update, with secrets-gated platform signing (public mechanism, private identity)


rjwalters:
## Proposal (architecture-level; likely an epic)

Ship prebuilt `loom-daemon` release artifacts from CI and teach the fleet to self-update from them, with platform signing as an **optional, secrets-gated hook** — the public repo ships the mechanism only; any deployer-specific identity lives in Actions secrets, never in tracked files.

## Motivation (all hit in production this week)

- **Version starvation (#4929)**: the auto-updater defers `cargo build` on busy hosts, so saturated hosts pin themselves to stale binaries — the busier the fleet, the staler its daemons. #4942 bounded the deferral, but the root cost is that every roll is a local rebuild. An artifact roll is a download + supervised restart: no build stampede is possible, so the gate (and its failure modes) mostly disappears.
- **Worker bootstrap**: `fleet add-worker` currently implies a Rust toolchain + full build on every worker. Artifact-based bootstrap means smaller/cheaper Linux VMs and minutes-faster provisioning — and the operator direction is more Linux workers over time, so this is the growth path.
- **macOS identity stability (#4953/#4962)**: per-host rebuilds force per-host signing. TCC grants and Gatekeeper/`syspolicyd` assessments key on the signing identity; ad-hoc-signed rebuilds reset them (recurring "access data from other apps" prompts, assessment churn that reads as CPU contention). One CI-signed binary fleet-wide fixes this class permanently on however many Macs remain.

## Design principles

1. **Public mechanism, private identity.** Precedent: `dashboard-deploy.yml` — the committed config is a generic template; the instance overlay is a secret. Same here: the release workflow's sign/notarize steps run **only if** the relevant secrets exist (e.g. macOS: Developer ID cert p12 + team id + optional notary API key; Linux: optional cosign key). Absent secrets ⇒ unsigned artifacts + SHA-256 checksums, fully functional for any public consumer. No org name, team ID, or cert reference in any tracked file.
2. **Checksums always, signatures when available.** The update path verifies the checksum unconditionally and the platform signature when present.
3. **Local build remains the fallback**, not the default: dev machines, canaries, forks without release access, and `--allow-stale`-style bisecting keep the current `loom-daemon-update.sh` build path.

## Suggested phasing

[...]

Issue #4330, opened 2026-07-29
loom-daemon-update.sh builds from a knowingly stale checkout — warn-and-proceed shipped a binary missing 5 merged commits; default to ff-first


rjwalters:
## Context

The 2026-07-29 daemon roll ran `./.loom/scripts/cli/loom-daemon-update.sh`, which printed:

    note: local main is 5 commit(s) behind origin/main — rebuilding now will NOT pick those up.
    Run 'git merge --ff-only origin/main' first if you want them.

…and then proceeded to rebuild, provision, and restart anyway. The shipped binary (81085967) is missing five merged commits, including a daemon behavior fix — #4322 (`fix(daemon): resolve dispatch's target workspace from the registry, not cwd`) — and the mcp-loom force-param (#4323). The operator now needs a second full rebuild + restart cycle to actually be current.

A warn-and-proceed default is the wrong ergonomics for this script: the entire purpose of running it is "get the running daemon onto the latest code," and the stale case is precisely when you'd least want it to proceed silently past a one-line note in the middle of cargo output.

## Suggested acceptance criteria

- [ ] `loom-daemon-update.sh` defaults to bringing the checkout current before building: attempt `git merge --ff-only origin/<default-branch>`; on success, build the updated HEAD.
- [ ] If the ff-merge cannot apply (diverged local commits, conflicting dirty tracked files), **abort with a clear message** instead of building stale — never guess or hard-reset.
- [ ] An explicit `--allow-stale` (or similar) flag restores today's build-what's-here behavior for deliberate use (bisecting, testing a local patch).
- [ ] The final "installed" line states the built commit AND whether it matches `origin/<default-branch>` at build time.

## Near-match note

#4017 (auto-rebuild + restart when the running binary is stale) is the adjacent *automation* feature — when it lands it should inherit this ff-first rule. This issue is deliberately narrower: the **manual** operator script should not build stale by default, independent of any auto-update machinery.

## Related

- #4017 (auto-rebuild), #4232 (supervised restart verification — the other half of roll ergonomics), #4322 (the fix a stale roll missed)

rjwalters:
## Curator Enhancement (2026-07-28)

The original issue is clear and its acceptance criteria are sound — this comment adds verified implementation guidance and the required Affected Files / Test Plan sections. All file references verified against `origin/main`.

### Verified current behavior

[...]

PR #5053, opened 2026-08-03
feat(daemon-update): fetch + verify prebuilt release artifacts instead of rebuilding

### LINKED ISSUE #5020: [Epic #4990] Phase 3: artifact-based self-update for loom-daemon
**Epic**: #4990 - release: prebuilt daemon artifacts + artifact-based self-update, with secrets-gated platform signing (public mechanism, private identity)
**Phase**: 3 of 4
**Phase Goal**: `loom-daemon-update.sh` (and the daemon auto-updater) learn a fetch mode that resolves the latest GitHub Release ≥ current version, verifies checksum/signature, and provisions the binary — so a saturated host converges to the latest release without ever compiling.

## Description

Phase 1 (#5003, closed) ships prebuilt `loom-daemon` artifacts + checksums on every GitHub Release. Phase 2 (#5011, closed, merged via #5018) adds secrets-gated macOS signing (+ optional notarization) and optional Linux cosign signing on top of those artifacts. Phase 3 teaches the update path to actually consume them:

- `loom-daemon-update.sh` and the daemon auto-updater gain a **fetch mode**: resolve the latest Release with version ≥ the currently running daemon, download the artifact for the host's platform + its checksum (and signature, when present), verify both, and provision to `~/.local/bin` via the existing `provision-daemon.sh` seam (which already handles the local-signing fallback used by source builds today).
- After provisioning, perform a supervised restart (existing precedent: #4054, systemd verification per #4950).
- **Precedence**: use a fetched artifact when one is available and version-matched; fall back to the current local source-build path otherwise (dev machines, canaries, forks without Releases, `--allow-stale`-style bisecting all keep working unchanged).
- Per design principle 2 from the epic: verify the checksum unconditionally; verify the platform signature only when one is present (Linux artifacts are typically unsigned and that's fine).

This phase does **not** depend on Phase 2 having signing secrets configured — checksum-only verification is sufficient to ship this phase, per the epic's own phase-dependency clarification (Phase 3 depends only on Phase 1's artifacts existing).

### Release asset shape (verified against `origin/main` `.github/workflows/release.yml`, PRs #5009 + #5018)

Every GitHub Release now carries, per target triple (`aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`):

- `loom-daemon-<target>` — the binary itself. macOS artifacts are codesigned **in place** (Developer ID) when the signing secrets were configured for that release; unsigned otherwise. There is no way to tell from the filename alone whether a given release's macOS asset is signed — the fetch logic must not assume either way.
- `loom-daemon-<target>.sha256` — a checksum file in `shasum -a 256` / `sha256sum` output format (`<hex>  loom-daemon-<target>`), computed **after** any in-place signing so it always matches the uploaded bytes. This is the "verify unconditionally" artifact.
- `loom-daemon-<target>.sig` — **Linux only**, and only present when the release was built with a cosign key secret configured. A detached cosign signature; verify with `cosign verify-blob --key <pubkey> --signature loom-daemon-<target>.sig loom-daemon-<target>`. Absent when the cosign secret wasn't configured for that release — this is the "signature only when present" case from AC2/AC3. **No public key is committed anywhere yet** — cosign verification needs `COSIGN_PUBLIC_KEY` (or a checked-in `.pub` file) to be resolvable client-side; if this phase can't source one, cosign verification should degrade to a loud skip (mirroring the workflow's own loud-skip convention) rather than block the update, consistent with "signature is optional, checksum is not."
- macOS signature verification is different in kind: there is no separate asset — verify via `codesign --verify --strict <binary>` (and optionally `codesign -dv` to confirm the identifier) against the downloaded binary itself, which will only succeed if the release's macOS asset was actually signed. A codesign failure on an unsigned artifact is expected and must not be treated as tamper evidence — distinguish "not signed" (skip, proceed) from "signed but verification failed" (abort) the same way the checksum path distinguishes "no `.sha256` file" (shouldn't happen — abort, since checksum is unconditional) from "checksum mismatch" (abort).

### Where the fetch mode plugs in

- `defaults/scripts/cli/loom-daemon-update.sh` (~1850 lines) is the sole rebuild/provision/restart orchestrator both the operator and the daemon invoke. It resolves the repo root, decides whether a rebuild is needed by comparing the resolved binary's embedded commit (`loom-daemon --version`) against local source HEAD, then (if stale) calls `cargo build --release`, provisions via `provision-daemon.sh`, and restarts under the existing launchd/systemd-aware supervised-restart machinery (exit codes 4/5/6/7 already encode build-verification, post-provision-verification, and restart-ack failure classes — see the script's own header comment for the full contract). The new fetch mode is a **parallel resolution path** that, when it wins precedence, replaces the `cargo build --release` step with "download + verify" but reuses everything downstream unchanged: `provision_machine_daemon` (via `provision-daemon.sh`), `verify_destination_binary` (adapted to compare against the release's *asset* checksum instead of a rebuilt source commit — the existing commit-embedding comparison doesn't apply to a fetched binary the same way), and the supervised-restart tier (launchd/systemd/pid-file) exactly as already implemented.
- `loom-daemon/src/auto_update.rs` is the daemon-side loop (`autonomous.autoUpdate`) that shells out to `loom-daemon-update.sh --no-restart` on a staleness/settle/defer-gate cadence (see `RebuildOutcome`, `run_rebuild` around lines 358-547, and the exit-code mapping around line 653). Since it already delegates the actual rebuild/provision work to the shell script, this phase likely does **not** need new Rust logic beyond (a) passing through whatever new flag/env selects fetch-vs-build precedence, and (b) treating the new fetch-specific exit codes the same defensive way it already treats 4/5/6/7. Confirm during implementation whether `auto_update.rs` needs any config surface changes (e.g. an `autonomous.autoUpdate.artifactFetch` toggle) or whether "prefer artifact automatically when available" should just be the unconditional default inside the shell script with no daemon-side awareness needed.
- `scripts/install/provision-daemon.sh`'s `provision_machine_daemon()` already accepts an arbitrary `<src_bin>` and is agnostic to whether that binary was just compiled or just downloaded — it does not need structural changes, only to keep being the single provisioning seam (do not duplicate its version-aware short-circuit or shim-installation logic in the new fetch path).
- Release resolution: "latest Release with version ≥ current" needs a GitHub API call (list releases / get latest release, matching `/repo:release`'s existing tag convention) plus a semver-ish comparison against the daemon's own reported version. Reuse the repo's existing `gh` CLI + REST-fallback convention (`.loom/docs/github-authentication.md`) rather than hand-rolling auth — this script already runs on hosts that may not have a GitHub token dedicated to it, so failure to resolve a release must be a soft "fall back to source build," not a hard error.

## Acceptance Criteria

- [ ] A saturated host with no Rust toolchain converges to the latest release version using only an artifact fetch — no `cargo build` invoked.
- [ ] Checksum verification is unconditional; a checksum mismatch aborts the update and leaves the running daemon untouched.
- [ ] Signature verification runs when a signature is present for the resolved artifact (macOS: embedded codesign verify; Linux: detached `.sig` via cosign, only when a public key is resolvable — otherwise a loud skip, not a block); its absence does not block the update.
- [ ] Falls back cleanly to the existing local source-build path when no matching release artifact exists (e.g. version not yet released for this platform, running from a fork with no Releases, or the GitHub API call itself fails/times out).
- [ ] Restart after provisioning is supervised and verified per the #4054/#4950 precedent (no orphaned or partially-restarted daemon) — reuses the existing launchd/systemd/pid-file restart tiers in `loom-daemon-update.sh` unchanged.
- [ ] `provision-daemon.sh`'s `provision_machine_daemon()` is reused unmodified as the provisioning seam for fetched binaries (no parallel/duplicate provisioning logic).
- [ ] Test coverage for: successful artifact update, checksum-mismatch abort, missing-artifact fallback to source build, and (if implemented) signature-present verification + signature-absent skip.
- [ ] `defaults/scripts/tests/test-loom-daemon-update.sh` and/or `tests/install/test-provision-daemon.sh` gain fixtures/cases for the new fetch path (or a new sibling test file, if the fetch logic is substantial enough to warrant its own suite).

## Implementation Notes / Open Questions for the Builder

- Decide and document the exact CLI/env surface for opting into (or forcing) fetch-vs-build precedence, e.g. `--fetch` / `--no-fetch` flags on `loom-daemon-update.sh`, mirroring the existing `--allow-stale` / `--force` idiom, plus whether `auto_update.rs` needs a matching `autonomous.autoUpdate.*` config key or should just inherit whatever the script defaults to.
- Cosign public-key distribution is unresolved in this repo as of Phase 2 — if there's no committed public key, ship the Linux-signature-verification path as best-effort/soft-skip rather than blocking the whole feature on key distribution; file a fast-follow if that's genuinely out of scope here.
- The GitHub Release version comparison should tolerate the repo's actual version scheme (see `scripts/version.sh` and the 5 version-bearing files it keeps in sync) — confirm semver ordering behaves correctly across the repo's actual tag format before relying on simple string comparison.

## Affected Files

- `defaults/scripts/cli/loom-daemon-update.sh` - add the artifact-fetch resolution path (download, checksum verify, signature verify), threading it into the existing stale-detection → provision → restart pipeline ahead of the `cargo build` fallback
- `loom-daemon/src/auto_update.rs` - confirm/adjust how the daemon-side auto-update loop selects or passes through fetch-vs-build precedence to `loom-daemon-update.sh --no-restart`; extend exit-code handling if the fetch path introduces new codes
- `scripts/install/provision-daemon.sh` - no structural change expected; reused as-is via `provision_machine_daemon()` for fetched binaries (confirm no changes needed during implementation)
- `defaults/scripts/tests/test-loom-daemon-update.sh` - add test cases for the fetch path (successful fetch, checksum mismatch, missing artifact fallback, signature verification)
- `tests/install/test-provision-daemon.sh` - extend only if the fetch path needs new provisioning test fixtures
- `.github/workflows/release.yml` - reference only; defines the exact asset naming/checksum/signature format this phase must consume (no changes expected here)
- `.loom/docs/daemon-reference.md` - update the `autonomous.autoUpdate` config table (and/or the "Autonomous self-update loop" section, ~line 5013) if this phase adds new config keys/env vars

## Test Plan

- [ ] Manual verification: on a host with no Rust toolchain, run the fetch-mode update against a real (or mocked) GitHub Release and confirm the daemon converges without invoking `cargo build`.
- [ ] Automated tests: new/extended cases in `defaults/scripts/tests/test-loom-daemon-update.sh` for successful artifact fetch, checksum-mismatch abort (verify the running daemon is untouched), and missing-artifact fallback to source build.
- [ ] Automated tests: signature verification cases — a release with a signature present (verify succeeds), and a release with no signature (soft-skip, update still proceeds).
- [ ] Edge cases: GitHub API unreachable/rate-limited during release resolution (must fall back to source build, not hard-fail); a release exists but has no artifact for the host's platform (fall back); daemon already at the latest release version (no-op, matching the existing up-to-date no-op contract).
- [ ] Regression: confirm `--allow-stale`, `--check`, `--dry-run`, and `--force` still behave as documented when fetch mode is available but not forced.

## Dependencies

Part of Epic #4990. Depends on Phase 1 (#5003, closed, merged) for artifacts to exist, and observes Phase 2's actual shipped asset format (#5011, closed, merged via #5018 — `.sha256` always present, macOS signature embedded in-binary, Linux `.sig` only when cosign secret was configured). Does not depend on Phase 2 having signing secrets *configured for a given release* — signature verification is optional per design principle 2.

---
*Created by Champion from Epic #4990. Enhanced by Curator 2026-08-03: added technical context on the verified release-asset shape (from `origin/main` .github/workflows/release.yml), pinpointed the fetch-path insertion points in loom-daemon-update.sh / auto_update.rs / provision-daemon.sh, added Affected Files and an expanded Test Plan.*

rjwalters:
## Summary

Teaches the update path to consume the release artifacts published by epic #4990 Phase 1 (#5003) and signed by Phase 2 (#5011/#5018), instead of always rebuilding from source.

`loom-daemon-update.sh` now **prefers a prebuilt GitHub Release artifact over a local `cargo build --release`**, so a CPU-saturated host — or one with no Rust toolchain installed at all — converges onto the latest release without ever compiling.

## Changes

**Resolution + precedence (three-state `FETCH_MODE`)**

| Mode | Selected by | Behavior |
|------|-------------|----------|
| `auto` (**default**) | — | Prefer a verified artifact when one resolves; **softly** fall back to the source build otherwise |
| `force` | `--fetch` / `LOOM_DAEMON_UPDATE_FETCH=1` | Require an artifact; **exit 1** rather than silently building from source |
| `off` | `--no-fetch` / `LOOM_DAEMON_UPDATE_FETCH=0` | Never fetch; pre-#5020 behavior |

An artifact wins only when the latest release is **strictly newer** than the installed daemon (an *equal* version wins only under an explicit `--fetch`, so plain `--force` keeps meaning "rebuild this checkout"), the host maps to a release target triple, and the release publishes both `loom-daemon-<target>` **and** its `.sha256`. Every other outcome — unrecognized platform, no `gh` CLI, no Releases (a fork), an unreachable/rate-limited API, no artifact for this platform — is a **soft** fallback, never a hard error.

**Verification** (epic design principle 2: checksum always, signature when present)

- **Checksum is unconditional.** A mismatch aborts the update (exit 1) and leaves the running daemon untouched. It is never downgraded to a fallback — a mismatch is tamper/corruption evidence, not a resolution failure.
- **Signature is verified only when present**, and its absence never blocks.
  - macOS: `codesign --verify --strict` in-binary, distinguishing "not signed at all" (expected **soft-skip**) from "signed but invalid" (**abort**).
  - Linux: the detached `.sig` via `cosign verify-blob`. Since no cosign public key is distributed with this repo yet, a missing `cosign` binary or unresolvable key is a **loud skip** rather than a block; `LOOM_DAEMON_UPDATE_COSIGN_PUBKEY` (or a committed `.loom/cosign.pub` / `defaults/cosign.pub`) turns it into real verification. See "Follow-up" below.

**Provisioning and restart are reused unchanged** — same `provision_machine_daemon()` seam, same supervised launchd/systemd/pid-file restart tiers (#4054/#4950). Two artifact-specific adjustments:

- `verify_destination_artifact()` replaces the source path's embedded-commit compare (a release's commit is unrelated to the local checkout's HEAD). It asserts the destination's full `--version` string equals the verified artifact's — the same string `provision_machine_daemon()`'s version-equality short-circuit compares — so a silent no-op roll still exits `5`.
- `sign_daemon_binary()` now **skips any binary already carrying a certificate-backed signature** (`codesign -dvvv` reports `Authority=`). Without this, the belt-and-braces ad-hoc re-sign inside `provision_machine_daemon()` would silently **downgrade** a Developer ID-signed release artifact to ad-hoc on every provision — exactly the Gatekeeper regression Phase 2 exists to avoid. A locally built binary is always ad-hoc, so the source path is unaffected. This is the only functional change to `provision-daemon.sh`; its structure and the `provision_machine_daemon()` seam are otherwise untouched.

**No new daemon config keys.** `auto_update.rs` passes no fetch flag and inherits the `auto` default, which is exactly the unattended behavior it wants (prefer an artifact, degrade to a rebuild). It deliberately does **not** pass `--fetch`, since that would turn a transient GitHub API hiccup into a failed roll. The exit-code contract is unchanged — a checksum/signature failure maps to exit `1` (Retryable, same bucket as a failed `cargo build`) — so `classify_exit` needs no new cases. The change there is a doc comment recording that reasoning.

| AC | Status | Evidence |
|----|--------|----------|
| Saturated host with no Rust toolchain converges via artifact fetch — no `cargo build` | Met | Test A runs with **no cargo on PATH at all** and asserts the output never contains `Rebuilding loom-daemon (cargo build` |
| Checksum unconditional; mismatch aborts, running daemon untouched | Met | Test B: exit 1 + destination binary byte-identical before/after |
| Signature verified when present; absence does not block | Met | Tests J/K/L (cosign verifies / invalid aborts / no pubkey = loud skip) and M/N/O (macOS unsigned soft-skip / signed verified / signed-invalid aborts) |
| Clean fallback when no matching artifact, or the API call fails | Met | Test C (no artifact for this platform) and Test F (API unreachable) both exit 0 having rebuilt from source |
| Restart supervised per #4054/#4950, reusing existing tiers unchanged | Met | Zero changes to the restart tiers; artifact mode joins the pipeline before the shared provision step |
| `provision_machine_daemon()` reused as the sole provisioning seam | Met | No parallel provisioning logic added; artifact mode calls the same seam |
| Test coverage for fetch / mismatch / fallback / signature states | Met | 15 new cases (A–O) in `test-loom-daemon-update.sh` + 1 in `test-provision-daemon.sh` |
| Test files gain fixtures for the new fetch path | Met | New fakes: `write_fake_gh`, `write_fake_gh_unreachable`, `write_fake_artifact_daemon`, `write_fake_codesign`, `sha256_of` |

Regression ACs from the Test Plan — `--allow-stale`, `--check`, `--dry-run`, `--force` — all still behave as documented (Tests G/H/I cover `--check` and `--dry-run` *in artifact mode* specifically; the pre-existing cases for all four still pass).

## Test Plan

- `bash defaults/scripts/tests/test-loom-daemon-update.sh` → **205 passed, 0 failed** (was 160 before; +45 assertions across 15 new cases). Includes the suite's own guard that the real `~/.local/bin/loom-daemon` is byte-identical before/after.
- `bash tests/install/test-provision-daemon.sh` → **44 passed, 0 failed** (+4 assertions, incl. "codesign -f is NEVER invoked" on an already-certificate-signed binary).
- `cargo check -p loom-daemon` → clean. `cargo fmt --check -p loom-daemon` → clean.
- `shellcheck -S warning` on all four changed shell files → only pre-existing SC2034s on unchanged lines; no new findings.

The macOS/Linux signature branches are exercised via a fake `codesign`/`cosign` on PATH, so all six signature states are testable on any host (including a Linux CI runner with no `codesign` at all).

## Notes / Follow-up

- **Cosign public-key distribution** is still unresolved from Phase 2 — no key is committed anywhere. Per the issue's explicit deferral to builder judgment, the Linux `.sig` path ships as soft-skip-if-unresolvable rather than blocking the whole feature. Making that verification mandatory needs a key-distribution decision; filed as a fast-follow.
- **Artifact resolution runs *after* the ff-first sync with `origin/<default-branch>` (#4330)**, so a checkout that hard-aborts that sync still exits `2` before any artifact is considered (`--allow-stale` skips the sync and goes straight to resolution). This is unchanged pre-existing behavior, not something the fetch path introduces, but it is now documented explicitly since "no Rust toolchain" is not the same as "no git checkout hygiene".
- On a host whose checkout is **ahead of the latest release** (the usual dev-host state), a newly published release costs one extra roll: the artifact installs first, then the next tick sees the release commit ≠ local HEAD and rebuilds from source, superseding it. This converges (the source build's version equals the release's, so no artifact wins afterwards — no ping-pong) and is the right trade for the hosts this exists for. Documented in the daemon reference.

Closes #5020

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4896,7 +4896,119 @@ manually rebuilt the Rust binary, reprovisioned it, and restarted the process.
 ./.loom/scripts/cli/loom-daemon-update.sh --force       # rebuild + provision + restart even if already up to date
 ./.loom/scripts/cli/loom-daemon-update.sh --no-restart  # rebuild + provision only; leave the running daemon untouched
 ./.loom/scripts/cli/loom-daemon-update.sh --relaunch    # launchd only: after a refused restart, re-render the plist + relaunch under supervision (preserves the live plist's LOOM_* env)
+./.loom/scripts/cli/loom-daemon-update.sh --fetch        # REQUIRE a verified release artifact (never silently fall back to a source build); exit 1 if none resolves
+./.loom/scripts/cli/loom-daemon-update.sh --no-fetch     # never consider release artifacts; always build from local source (pre-#5020 behavior)
 ```
+
+#### Artifact-based self-update (epic #4990 Phase 3, #5020)
+
+By default the script now **prefers a prebuilt GitHub Release artifact over a
+local `cargo build --release`**, so a CPU-saturated host — or one with no Rust
+toolchain installed at all — converges onto the latest release without ever
+compiling. This is the consumer side of the release artifacts published by
+Phase 1 (#5003) and signed by Phase 2 (#5011).
+
+**Precedence (three-state, `FETCH_MODE`)**
+
+| Mode | How to select | Behavior |
+|------|---------------|----------|
+| `auto` (**default**) | — | Prefer a verified artifact when one resolves; **softly** fall back to the source build otherwise |
+| `force` | `--fetch` / `LOOM_DAEMON_UPDATE_FETCH=1` | Require an artifact; **exit 1** rather than silently building from source |
+| `off` | `--no-fetch` / `LOOM_DAEMON_UPDATE_FETCH=0` | Never fetch; always build from local source |
+
+An artifact is used only when the latest release's version is **strictly newer**
+than the installed daemon's (an *equal* version wins only under an explicit
+`--fetch`; plain `--force` keeps meaning "rebuild this checkout"), the host's
+platform maps to one of the release target triples
+(`aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`),
+and that release actually publishes `loom-daemon-<target>` **and** its
+`.sha256`. Everything else — an unrecognized platform, no `gh` CLI, no Releases
+(a fork), an unreachable or rate-limited GitHub API, a release with no artifact
+for this platform — is a **soft** fallback to the existing source-build path, so
+dev machines, canaries, and forks behave exactly as before.
+
+**Verification (epic design principle 2: checksum always, signature when present)**
+
+- **Checksum — unconditional.** The downloaded binary is compared against the
+  release's `.sha256`. A mismatch **aborts the whole update (exit 1)** and leaves
+  the running daemon untouched; it is never downgraded to a source-build fallback,
+  because a mismatch is tamper/corruption evidence, not a resolution failure.
+- **Signature — only when present, and never blocking by its absence.**
+  - macOS: verified in-binary via `codesign --verify --strict`. "Not signed at
+    all" (a release built without the Developer ID secrets) is an expected
+    **soft-skip**; "signed but verification fails" is treated as tamper evidence
+    and **aborts (exit 1)**.
+  - Linux: the detached `loom-daemon-<target>.sig` (published only when the
+    release was built with a cosign key) is verified with `cosign verify-blob`.
+    If `cosign` is missing, or no public key is resolvable, the script emits a
+    **loud skip** and proceeds — no cosign public key is distributed with this
+    repo yet. Point `LOOM_DAEMON_UPDATE_COSIGN_PUBKEY` at a key (or commit one at
+    `.loom/cosign.pub` / `defaults/cosign.pub`) to turn that skip into real
+    verification. A resolvable key plus a failing verification **aborts (exit 1)**.
+
+**Provisioning and restart are unchanged.** A verified artifact goes through the
+*same* `provision_machine_daemon()` seam as a freshly compiled binary (no parallel
+provisioning logic), and the same supervised launchd/systemd/pid-file restart
+tiers (#4054/#4950). Two artifact-specific adjustments:
+
+- **Post-provision verification** uses `verify_destination_artifact()` instead of
+  the source path's embedded-commit compare (a release's commit is unrelated to
+  the local checkout's HEAD). It asserts the destination's full `--version` string
+  equals the verified artifact's — the same string `provision_machine_daemon()`'s
+  version-equality short-circuit compares, so a silent no-op roll still exits `5`.
+- **Signing is skipped** for fetched binaries, and `sign_daemon_binary()` now
+  refuses to re-sign any binary that already carries a certificate-backed
+  signature (`codesign -dvvv` reports an `Authority=`). Without that guard, the
+  belt-and-braces ad-hoc re-sign inside `provision_machine_daemon()` would
+  silently *downgrade* a Developer ID-signed release artifact to ad-hoc on every
+  provision. A locally built binary is always ad-hoc, so this changes nothing for
+  the source path.
+
+**Interaction with the autonomous loop's staleness check (worth knowing).** The
+loop's own staleness signal is still *embedded commit vs. local source HEAD*. On a
+host whose checkout is **ahead of the latest release** (the usual state for a
+development host), a newly published release therefore costs one extra roll: the
+artifact is fetched and installed first, then the next tick still sees the
+installed release commit ≠ local HEAD and rebuilds from source, superseding it.
+That is deliberate, converges (the source build's version equals the release's, so
+no artifact wins afterwards — there is no ping-pong), and is the right trade on the
+hosts this exists for: a saturated host reaches the release *immediately* instead
+of waiting out the up-to-6h build-stampede deferral, and a host with no Rust
+toolchain simply stays on the release (its source-build attempt fails as retryable
+and never displaces it). Hosts that want only source builds can set
+`LOOM_DAEMON_UPDATE_FETCH=0`.
+
+**Still requires a source checkout.** This phase changes *what gets installed*,
+not the script's entry conditions: `loom-daemon/Cargo.toml` must still exist (the
+script also uses the checkout to resolve `origin`'s `owner/repo` slug). Removing
+that requirement for binary-only installs is out of scope here.
+
+Concretely, artifact resolution runs *after* the ff-first sync with
+`origin/<default-branch>` (#4330), so a checkout that hard-aborts that sync
+(genuine divergence, unmanaged dirty tracked files) still exits `2` before any
+artifact is considered — pass `--allow-stale` to skip the sync entirely and go
+straight to resolution, or fix the checkout. This is unchanged pre-existing
+behavior, not something the fetch path introduces, but it does mean "no Rust
+toolchain" is not the same as "no git checkout hygiene".
+
+**Extra environment variables** (all optional; the first three are primarily test
+seams):
+
+| Variable | Purpose |
+|----------|---------|
+| `LOOM_DAEMON_UPDATE_FETCH` | `1`/`true`/`yes` ⇒ force (`--fetch`); `0`/`false`/`no` ⇒ off (`--no-fetch`); unset ⇒ auto |
+| `LOOM_DAEMON_UPDATE_GH_REPO` | Override the `owner/repo` slug used for release resolution (default: parsed from the `origin` remote) |
+| `LOOM_DAEMON_UPDATE_TARGET` | Override the detected release target triple |
+| `LOOM_DAEMON_UPDATE_COSIGN_PUBKEY` | Path to the cosign public key used to verify a Linux `.sig` |
+
+**No new daemon config keys.** The autonomous self-update loop
+(`autonomous.autoUpdate.*`) needs no change and passes no fetch flag: it invokes
+`loom-daemon-update.sh --no-restart` and inherits the `auto` default, which is
+exactly the unattended behavior it wants (prefer an artifact, degrade to a
+rebuild). It deliberately does **not** pass `--fetch`, since that would turn a
+transient GitHub API hiccup into a failed roll. The existing exit-code contract is
+unchanged: a checksum/signature failure maps to exit `1` (retryable, same bucket
+as a failed `cargo build`), so `classify_exit` needs no new cases.
 
 **Launchd refused-restart fallback (`--relaunch`, exit 6, #4118)**: on the
 FIRST roll onto a #4077-capable binary the *running* (old) daemon has no
@@ -4997,8 +5109,10 @@ Self-update: built from ab12cd3 — UPDATE AVAILABLE (source checkout HEAD is de
 ```
 
 `loom-daemon-update.sh` requires an actual Loom source checkout
-(`loom-daemon/Cargo.toml` must exist) — it rebuilds from source and refuses to
-run against a binary-only / release-tarball install.
+(`loom-daemon/Cargo.toml` must exist) and refuses to run against a binary-only /
+release-tarball install — even when it ends up installing a prebuilt release
+artifact rather than rebuilding (see
+[Artifact-based self-update](#artifact-based-self-update-epic-4990-phase-3-5020)).
 
 ### Autonomous self-update loop (#4055)
 
--- a/defaults/scripts/cli/loom-daemon-update.sh
+++ b/defaults/scripts/cli/loom-daemon-update.sh
@@ -32,9 +32,19 @@
 # entirely) for deliberate use (bisecting, testing a local patch) — see below.
 #
 # It:
-#   - detects whether the resolved binary is stale vs. the local source tree,
+#   - detects whether the resolved binary is stale vs. the local source tree
+#     (or, in artifact-fetch mode, vs. the latest GitHub Release — see below),
+#   - Artifact-fetch mode (Epic #4990 Phase 3, #5020, ON by default, opt out
+#     with --no-fetch): resolves the latest GitHub Release with version >=
+#     the installed daemon and, when that release publishes an artifact for
+#     this host's platform, downloads + verifies it (checksum unconditional,
+#     signature when present) INSTEAD of rebuilding from source — a
+#     saturated host with no Rust toolchain converges on a release alone.
+#     Resolution failure of any kind (unrecognized platform, no `gh` CLI, no
+#     Releases yet, API unreachable, no artifact for this platform) SOFTLY
+#     falls back to the rebuild path below (--fetch instead hard-fails).
 #   - rebuilds (`cargo build --release`) in loom-daemon/ when stale (or
-#     --force),
+#     --force) AND no artifact was fetched,
 #   - provisions the fresh binary to wherever the resolved binary lives
 #     (LOOM_DAEMON_BIN override, else the machine-level ~/.local/bin install
 #     via scripts/install/provision-daemon.sh, matching #3922's convention),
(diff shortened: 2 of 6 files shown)


rjwalters:
Fast-follow filed as #5054 (cosign public-key distribution — the Linux `.sig` path currently soft-skips because no key is resolvable on a stock install; checksum verification is unaffected and remains unconditional).

rjwalters:
Code evaluation in progress; CI is still running (Shell Test Suites (hermetic) pending). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T17:04:25Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed the artifact-fetch/verification logic in depth (checksum bypass safety, macOS codesign / Linux cosign three-way soft-skip/hard-abort logic, version-downgrade protection, `sign_daemon_binary()`'s already-signed guard, and fallback-path safety) — all sound. Spot-checked several of the new test cases (checksum mismatch, signature invalid, unsigned soft-skip) against the actual shell logic to confirm they exercise real failure paths, not vacuous passes.

## Test Execution

- `bash defaults/scripts/tests/test-loom-daemon-update.sh` — Executed: 205/205 pass
- `bash tests/install/test-provision-daemon.sh` — Executed: 44/44 pass
- `shellcheck` — Executed via CI (pass)
- `cargo check -p loom-daemon` / `cargo fmt --check` — Executed via CI (Rust Unit Tests, Rust Linting & Build Check both pass)

All CI checks pass, merge state CLEAN.

One minor doc nit (non-blocking): `daemon-reference.md` states the ff-first-sync-abort path "exits 2", but the actual code (`loom-daemon-update.sh` around the `sync_with_origin` check) exits 1, matching the script's own header comment. Worth a follow-up doc fix but not worth blocking this PR over.

rjwalters:
## Judge review — approve

Reviewed the actual PR diff (confirmed via `gh api repos/rjwalters/loom/pulls/5053/files`, 6 files: `defaults/scripts/cli/loom-daemon-update.sh`, `defaults/scripts/tests/test-loom-daemon-update.sh`, `scripts/install/provision-daemon.sh`, `tests/install/test-provision-daemon.sh`, `loom-daemon/src/auto_update.rs`, `defaults/docs/daemon-reference.md` — the `release.yml`/`observability.md`/`telemetry-schema.md` deltas that show up in a naive `main..branch` diff are just base-branch drift from already-merged #5018/#5019, not part of this PR).

**Verification performed:**

1. **Checksum verification is unconditional.** `verify_artifact_checksum()` runs before any signature check; a mismatch in `fetch_and_verify_artifact()` calls `exit 1` directly with an explicit "the running daemon (if any) is left untouched" message, before `provision_machine_daemon()` is ever invoked. Confirmed this is unconditional (not gated behind FETCH_MODE or platform).

2. **`sign_daemon_binary()`'s Authority-guard is correct.** Ran `tests/install/test-provision-daemon.sh` locally — 44/44 pass, including the new test 16 which proves `codesign -f` is never invoked against an already-certificate-signed binary while a locally-built (ad-hoc-only) binary still gets signed normally (tests 9-15, unchanged, still green). The `grep -q '^Authority='` check against `codesign -dvvv` output correctly distinguishes a real cert chain from cargo's linker-applied ad-hoc signature.

3. **Fallback to source build is safe.** Every resolution failure path (unrecognized platform, missing `gh`, no releases, unreachable/rate-limited API, no matching artifact) sets `ARTIFACT_FALLBACK_REASON` and warns, then falls through to the existing source-build code path unchanged — never a silent no-op. `--fetch`/`LOOM_DAEMON_UPDATE_FETCH=1` correctly hard-fails (exit 1) instead of falling back when forced.

4. **Test coverage.** Ran both suites locally against `origin/feature/issue-5020`:
   - `bash defaults/scripts/tests/test-loom-daemon-update.sh` → 205/205 passed (matches the claimed 15 new fetch-mode cases A–O).
   - `bash tests/install/test-provision-daemon.sh` → 44/44 passed.
   - `shellcheck -S warning` on both changed shell files → only the pre-existing `PROVISIONED_DAEMON_BIN` SC2034 (used externally), no new findings.

5. **No injection/secrets issues found.** `gh release download`/`gh release view` calls pass all variable interpolation as separate argv elements (`-p "$bin_name"`, `-R "$FETCH_REPO_SLUG"`, etc.), never through `eval`/string-built commands; `ARTIFACT_TARGET`/`FETCH_TARGET` values are constrained to a fixed enum from `detect_target_triple()` (or a test-only env override). Temp dirs use `mktemp -d "${TMPDIR:-/tmp}/loom-daemon-fetch.XXXXXX"` and are cleaned up via a single `trap ... EXIT` (verified no pre-existing trap in this script that this would clobber).

**CI / merge state:**
- `gh pr checks 5053` — all required checks green (Rust Unit Tests, Rust OTLP Feature Build & Test, Shell Test Suites (hermetic), Rust Linting & Build Check, Shellcheck, Installer Integration Tests, Docs/Defaults Parity Check, AGENTS.md Sync Check, CLAUDE.md Line Budget, Codex Adapter Smoke).
- `gh pr view --json mergeStateStatus` → `CLEAN`, `mergeable: MERGEABLE`.

Nicely scoped: reuses `provision_machine_daemon()` and the existing supervised-restart tiers unchanged, follows the epic's "checksum always, signature when present" principle precisely, and the cosign-pubkey-distribution gap is correctly deferred to a filed follow-up (#5054) rather than blocking this phase. No changes requested.

Issue #1, opened 2025-10-11
Set Up Basic Tauri Application Structure


rjwalters:
# Issue #1: Tauri Application Setup

## Objective

Set up the foundational Tauri application with vanilla TypeScript, TailwindCSS, and dark/light mode support.

### 1. Project Initialization

- [ ] Initialize Tauri project with vanilla TypeScript
- [ ] Configure directory structure:
  ```
  loom/
  ├── src-tauri/          # Rust backend
  ├── src/                # TypeScript frontend
  │   ├── main.ts
  │   ├── style.css
  │   └── lib/
  ├── index.html
  ├── package.json
  ├── tsconfig.json
  ├── tailwind.config.js
  └── vite.config.ts
  ```

### 2. Dependencies

**Frontend:**
```json
{
  "dependencies": {
    "@tauri-apps/api": "^1.5.0"
  },
  "devDependencies": {
    "typescript": "^5.3.0",
    "vite": "^5.0.0",
    "tailwindcss": "^3.4.0",
    "autoprefixer": "^10.4.0",
    "postcss": "^8.4.0"
  }
}
```

**Backend (src-tauri/Cargo.toml):**
```toml
[dependencies]
tauri = { version = "1.5", features = ["shell-open"] }
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
```

### 3. Tauri Configuration

- [ ] Configure `tauri.conf.json`:
```json
{
  "build": {
    "beforeDevCommand": "npm run dev",
    "beforeBuildCommand": "npm run build",
    "devPath": "http://localhost:1420",
    "distDir": "../dist"
  },
  "package": {
    "productName": "Loom",
    "version": "0.1.0"
  },
  "tauri": {
    "windows": [
      {
        "title": "Loom",
        "width": 1400,
        "height": 900,
        "minWidth": 1000,
        "minHeight": 600,
        "resizable": true,
        "fullscreen": false
      }
    ]
  }
}
```

### 4. TypeScript Configuration

- [ ] Create `tsconfig.json`:
```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "ESNext",
    "lib": ["ES2020", "DOM", "DOM.Iterable"],
    "skipLibCheck": true,
    "moduleResolution": "bundler",
    "allowImportingTsExtensions": true,
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": true,
    "strict": true,
    "noUnusedLocals": true,
[...]

Issue #130, opened 2025-10-15
Implement structured logging and error tracking system


rjwalters:
## Problem Statement

Loom currently lacks a structured approach to logging, error handling, and observability across both frontend (TypeScript) and backend (Rust). This makes debugging production issues difficult and prevents proper monitoring of system health.

**Current Issues:**
1. **Console logging inconsistency**: Mix of `console.log`, `console.error`, and `console.warn` with varying message formats
2. **No error aggregation**: Errors scattered across files with no central tracking
3. **Limited observability**: Hard to trace issues across daemon ↔ Tauri ↔ frontend boundaries
4. **No metrics collection**: Can't measure terminal health, agent performance, or resource usage
5. **Poor debugging experience**: Console logs go to file but lack structure, timestamps vary, no severity levels

**Example Problem Scenario:**
When a worktree setup fails during factory reset:
- Error appears in browser console (maybe)
- Error might be logged to `~/.loom/console.log`
- Daemon error might be in daemon logs
- User sees generic "Failed to create terminal" alert
- **Result**: Hard to diagnose whether failure was git, tmux, filesystem, or daemon issue

## Current State

**Frontend Logging** (`src/main.ts:22-61`):
```typescript
// Console interception writes to file, but no structure
console.log = (...args: unknown[]) => {
  originalConsoleLog(...args);
  writeToConsoleLog("INFO", ...args);
};
```

**Backend Logging** (`loom-daemon/src/main.rs`):
```rust
// Uses env_logger with basic formatting
env_logger::Builder::from_default_env()
    .filter_level(log::LevelFilter::Info)
    .init();
```

**Error Handling Patterns:**
- Tauri commands: `Result<T, String>` (loses error context)
- Frontend: `try/catch` with `alert()` or `console.error()`
- Daemon: Mix of `anyhow::Result` and string errors
- No error IDs or correlation across boundaries

### Option 1: Lightweight Structured Logging (Recommended)

**Approach**: Minimal overhead, focuses on structure and correlation

**Frontend**:
```typescript
// src/lib/logger.ts
interface LogContext {
  component: string;
  terminalId?: string;
  workspacePath?: string;
  [key: string]: unknown;
}

class Logger {
  log(level: 'INFO' | 'WARN' | 'ERROR', message: string, context?: LogContext) {
    const entry = {
      timestamp: new Date().toISOString(),
      level,
      message,
      ...context
    };
    console.log(JSON.stringify(entry)); // Structured JSON
  }
  
[...]

PR #5058, opened 2026-08-03
Judge fallback-queue dedup via marker

### LINKED ISSUE #4989: Judge fallback-queue evaluation has no dedup — re-reviews same unlabeled PR every pass, 129+ duplicate comments on PR #4972
## Problem

The Judge's "Fallback Queue (When No Labeled Work)" path evaluates unlabeled
open PRs when `loom:review-requested` is empty, and explicitly does **not**
apply any `loom:` label afterward (by design — these may be external/manual
PRs outside the Loom workflow). But that means there is no signal anywhere
that an unlabeled PR has already been evaluated, so every subsequent Judge
pass (cron tick, idle sweep capacity, etc.) re-discovers it via the fallback
search and re-reviews it from scratch.

**Observed impact**: PR #4972 (`chore(deps): bump libc from 0.2.186 to
0.2.189`, a trivial Dependabot `Cargo.lock`-only patch bump, `mergeStateStatus:
CLEAN`, all applicable CI green) currently has **129 comments**, the vast
majority near-identical "✅ looks good, fallback-mode evaluation" writeups from
repeated Judge passes, landing roughly every 5 minutes over an extended period
today (2026-08-03). Each pass re-fetches the diff, re-reads the changelog
links, and re-posts a full evaluation comment — pure wasted API calls and
tokens for a PR whose verdict has not changed across 100+ evaluations.

## Root cause

The fallback queue selection query has no way to exclude "already evaluated in
fallback mode" PRs:

```bash
gh pr list --state=open --limit 500 --json number,title,labels \
  --jq '.[] | select(([.labels[].name | select(startswith("loom:"))] | length) == 0) | "#\(.number) \(.title)"'
```

This returns every unlabeled open PR every time, regardless of prior fallback
evaluations.

## Suggested fix

Some marker-based dedup, mirroring the `<!-- loom:standdown claim=... -->` /
`<!-- loom:fast-track-evaluation -->` conventions already used elsewhere in
`judge.md`:

- Have the fallback-mode evaluation comment end with a machine-readable marker
  (e.g. `<!-- loom:fallback-evaluated sha=<head-sha> -->`).
- Before evaluating an unlabeled PR in fallback mode, check for an existing
  marker comment whose `sha` matches the PR's current head SHA; if found, skip
  it (nothing has changed since the last fallback pass).
- Only re-evaluate if the head SHA has advanced (new commits pushed) since the
  last marker.

This keeps the fallback queue's "maximize idle Judge utilization" intent while
eliminating the unbounded duplicate-comment growth for PRs that never change
(dependency-bot PRs sitting unmerged being the common case).

## Note

This is likely also worth resolving pragmatically for PR #4972 itself — it is
a trivial, safe, `CLEAN` + all-green Dependabot patch bump that should
probably just be merged (dependency bumps like this are typically handled by
Dependabot auto-merge or a human, outside the Loom label workflow, so Judge
does not merge it directly).

Discovered while running `/loom:judge` in fallback mode with an empty
`loom:review-requested` queue.

---

## Curator Enhancement

Verified against `origin/main`: the fallback-queue path in
`defaults/.claude/commands/loom/judge.md` (§"Fallback Queue (When No Labeled
Work)", ~lines 658-757) confirms the root cause as described. Two existing
marker conventions in the same file solve an analogous dedup problem:
`<!-- loom:standdown claim=$CLAIMED_AT -->` (line 492) and
`<!-- loom:fast-track-evaluation -->` (line 1270) — no
`loom:fallback-evaluated` marker or SHA-dedup logic exists anywhere in the
repo today (confirmed via grep), so this is a net-new addition. No duplicate
or related open work found (`check-duplicate.sh --issue 4989`).

This is a **process-improvement issue** (Judge's own re-review behavior), so
the acceptance criteria below demand a structural behavior change rather than
a documentation-only fix.

### Acceptance Criteria

- [ ] The fallback-mode evaluation comment in `judge.md`'s example workflow
  ends with a machine-readable marker: ``, following the same marker-comment convention as
  `loom:standdown claim=...` (line 492) and `loom:fast-track-evaluation`
  (line 1270).
- [ ] Before evaluating an unlabeled PR in fallback mode, the documented
  procedure checks PR comments for an existing `loom:fallback-evaluated`
  marker whose `sha` matches the PR's **current** head SHA (`gh pr view
  <N> --json headRefOid`). If found, the procedure skips this PR (moves to
  the next unlabeled PR, or exits if none remain) **without posting a new
  comment**.
- [ ] If the head SHA has advanced since the last marker (new commits
  pushed), the procedure re-evaluates normally and posts a fresh comment
  ending in an updated marker.
- [ ] Structural test (not just "docs updated"): the next fallback-mode
  evaluation this repo's Judge performs on an unlabeled PR whose head SHA is
  unchanged since its last fallback pass must produce **zero** new PR
  comments — verifiable by comparing comment count before/after a Judge tick
  on such a PR (PR #4972, while still open and unlabeled, is a live fixture
  for this).
- [ ] Cleanup of PR #4972's existing 129 duplicate comments is explicitly
  **out of scope** for this issue (no retroactive comment deletion) — only
  stopping *further* growth is in scope. Note this in the PR body so a
  Builder doesn't scope-creep into comment cleanup.

### Affected Files

- `defaults/.claude/commands/loom/judge.md` - Add the `loom:fallback-evaluated
  sha=<head-sha>` marker to the fallback-mode evaluation comment template
  (~line 742-748); add the marker-check-before-evaluate step to the fallback
  decision tree (~line 669-695) and the example workflow (~line 723-753),
  mirroring the existing `loom:standdown` marker-check pattern (~lines
  492-524).
- `.claude/commands/loom/judge.md` - installed copy of the same file; per
  CLAUDE.md this is normally refreshed by
  `./.loom/scripts/resync-installed.sh` rather than hand-edited, but flagging
  it here in case the Builder's environment requires a direct edit to pick up
  the change immediately in this repo's own running Judge instances.

### Test Plan

- [ ] Manual verification: dry-run the updated fallback procedure text
  against a scripted scenario — unlabeled PR, first pass posts a comment
  ending in `<!-- loom:fallback-evaluated sha=abc123 -->`; second pass with
  no new commits detects the existing marker and skips (no new comment);
  third pass after a new commit (different head SHA) re-evaluates and posts
  a new marker.
- [ ] Automated tests: N/A — `judge.md` is a prompt/procedure document, not
  executable code; there is no existing test harness for role-prompt
  behavior in this repo (confirmed via grep for related marker/dedup
  helpers). If a future harness is added for role-prompt logic, this would
  be a candidate to backfill, but that is out of scope here.
- [ ] Edge cases: (1) PR has multiple prior `loom:fallback-evaluated` markers
  with different SHAs (must match against the *most recent* marker's SHA,
  not any historical one); (2) PR transitions from unlabeled to
  `loom:review-requested` mid-flight (normal-path evaluation takes over,
  fallback marker becomes irrelevant — no special handling needed since the
  fallback path is only entered when the labeled queue is empty).

rjwalters:
## Summary

Adds a machine-readable marker (`<!-- loom:fallback-evaluated sha=<head-sha> -->`) to fallback-mode evaluation comments so subsequent Judge passes can detect already-evaluated unlabeled PRs and skip re-review when the head SHA has not changed.

This prevents unbounded duplicate-comment growth on PRs (like #4972 with 129+ comments) that remain open and unchanged since the last fallback-mode pass, while still re-evaluating when new commits are pushed (SHA differs).

## Changes

- **Decision tree**: Updated to show marker-check step before evaluating fallback-queue PRs
- **Example workflow**: Added SHA extraction and marker-check logic before evaluation
- **Comment template**: Appends `<!-- loom:fallback-evaluated sha=$CURRENT_HEAD_SHA -->` marker to fallback evaluation comments
- **Installed copy**: Also updated `.claude/commands/loom/judge.md` for immediate Judge instance pickup

## Edge Cases Handled

1. **Multiple prior markers with different SHAs**: Takes the most recent marker's SHA using `tail -n 1`
2. **PR transitions from unlabeled to loom:review-requested mid-flight**: Normal-path evaluation takes over (fallback path is only entered when labeled queue is empty)
3. **New commits pushed to PR**: Head SHA differs from marker SHA, so PR is re-evaluated with fresh comment

## Out of Scope

Cleanup of PR #4972's existing 129 duplicate comments is explicitly out of scope — only stops future growth. Comment cleanup would require iterative comment deletion via the API and is a separate maintenance task, not part of preventing future accumulation.

Closes #4989

### DIFF
--- a/defaults/.claude/commands/loom/judge.md
+++ b/defaults/.claude/commands/loom/judge.md
@@ -688,9 +688,11 @@ Pre-Iteration Environment Check (gh repo view)
                     ↓
                 Search for unlabeled open PRs
                     ↓
-                    ├─→ Found? → Check for loom:fallback-evaluated marker
-                    │     │        with matching head SHA
-                    │     ├─→ Found (no new commits)? → Skip to next unlabeled PR
+                    ├─→ Found? → Walk the list in order; for each candidate check
+                    │     │        for a loom:fallback-evaluated marker whose SHA
+                    │     │        matches that PR's current head SHA
+                    │     ├─→ Found (no new commits)? → Skip it, try the next
+                    │     │        unlabeled PR (exit iteration if none remain)
                     │     └─→ Not found, or SHA differs? → Evaluate and post comment
                     │              (with updated marker ending)
                     │
@@ -723,27 +725,51 @@ if [ "$LABELED_PRS" -gt 0 ]; then
 else
   echo "No loom:review-requested PRs found, checking unlabeled PRs..."
 
-  # 2. Check fallback queue (cached — see "Cached Forge Reads")
-  UNLABELED_PR=$("$GH_READ" pr list --state=open --limit 500 --json number,labels \
-    --jq '.[] | select(([.labels[].name | select(startswith("loom:"))] | length) == 0) | .number' \
-    | head -n 1)
+  # 2. Check fallback queue (cached — see "Cached Forge Reads"). Keep the WHOLE
+  #    candidate list, not just the head of it: a PR that was already evaluated
+  #    at its current head SHA is skipped, and the walk below moves on to the
+  #    next candidate rather than exiting and claiming "no work".
+  UNLABELED_PRS=$("$GH_READ" pr list --state=open --limit 500 --json number,labels \
+    --jq '.[] | select(([.labels[].name | select(startswith("loom:"))] | length) == 0) | .number')
+
+  # 3. Walk the candidates in order; stop at the first one with no prior
+  #    fallback-evaluated marker for its current head SHA (dedup).
+  UNLABELED_PR=""
+  CURRENT_HEAD_SHA=""
+  for CANDIDATE in $UNLABELED_PRS; do
+    CANDIDATE_HEAD_SHA=$(gh pr view "$CANDIDATE" --json headRefOid --jq '.headRefOid')
+
+    # Extract the most recent loom:fallback-evaluated marker from PR comments.
+    # `--paginate` is REQUIRED: without it `gh api` returns only the first page
+    # (default per_page=30, oldest-first), so on a long-lived PR (#4972 already
+    # had 129 comments when this dedup was added) a marker posted near the end
+    # of the history would never be seen and the dedup would silently never
+    # engage — the exact bug this check exists to prevent. Same pitfall the
+    # Stale `loom:reviewing` Claim Check documents above; with `--jq`,
+    # `--paginate` re-runs the filter per page and concatenates the per-page
+    # output, which is what `tail -n 1` (most-recent-marker-wins) consumes —
+    # pages arrive oldest-first, so the last line is the newest marker.
+    # Extraction is `jq`-only on purpose: `grep -oP` (PCRE lookaround) is a
+    # GNU-only flag that stock BSD/macOS grep rejects outright
+    # (`grep: invalid option -- P`), and a Judge running under an alternate
+    # runtime would degrade silently to "no marker found" — i.e. back to
+    # re-evaluating every pass. jq's regex engine is the same everywhere.
+    # `capture` emits nothing (no error) for a body without the marker.
+    PRIOR_MARKER_SHA=$(gh api "repos/{owner}/{repo}/issues/$CANDIDATE/comments" --paginate \
+      --jq '.[] | (.body // "") | capture("<!-- loom:fallback-evaluated sha=(?<sha>[0-9a-f]+) -->") | .sha' \
+      | tail -n 1)
+
+    if [ -n "$PRIOR_MARKER_SHA" ] && [ "$CANDIDATE_HEAD_SHA" = "$PRIOR_MARKER_SHA" ]; then
+      echo "Skipping unlabeled PR #$CANDIDATE: already evaluated in fallback mode (head SHA unchanged since last evaluation) — trying the next unlabeled PR"
+      continue
+    fi
+
+    UNLABELED_PR="$CANDIDATE"
+    CURRENT_HEAD_SHA="$CANDIDATE_HEAD_SHA"
+    break
+  done
 
   if [ -n "$UNLABELED_PR" ]; then
-    # 3. Check for prior fallback-evaluated marker (dedup) before evaluating
-    CURRENT_HEAD_SHA=$(gh pr view $UNLABELED_PR --json headRefOid --jq '.headRefOid')
-    
-    # Extract the most recent loom:fallback-evaluated marker from PR comments
-    # Filter comments containing the marker, extract SHA values, and take the most recent
-    PRIOR_MARKER_SHA=$(gh api "repos/{owner}/{repo}/issues/$UNLABELED_PR/comments" \
-      | jq -r '.[] | select(.body | contains("<!-- loom:fallback-evaluated sha=")) | .body' \
-      | grep -oP '(?<=<!-- loom:fallback-evaluated sha=)[a-f0-9]+(?= -->)' \
-      | tail -n 1)
-    
-    if [ "$CURRENT_HEAD_SHA" = "$PRIOR_MARKER_SHA" ] && [ -n "$PRIOR_MARKER_SHA" ]; then
-        echo "Skipping unlabeled PR #$UNLABELED_PR: already evaluated in fallback mode (head SHA unchanged since last evaluation)"
-        exit 0
-    fi
-    
     echo "Evaluating unlabeled PR #$UNLABELED_PR (fallback mode)"
 
     # Check out and evaluate the PR (worktree-aware — see "PR Branch Isolation")
@@ -756,8 +782,14 @@ else
     fi
     # ... run checks, evaluate code ...
 
-    # Provide feedback but DO NOT add workflow labels
-    gh pr comment $UNLABELED_PR --body "$(cat <<'EOF'
+    # Provide feedback but DO NOT add workflow labels.
+    # NOTE: this heredoc is deliberately UNQUOTED (`<<EOF`, not `<<'EOF'`) so
+    # $CURRENT_HEAD_SHA expands into the marker. With a quoted delimiter the
+    # marker would post the literal string "sha=$CURRENT_HEAD_SHA", which can
+    # never equal a real head SHA — the dedup read above would then match
+    # nothing and every pass would re-evaluate. Keep any other `$` or backticks
+    # out of this body, or escape them.
+    gh pr comment $UNLABELED_PR --body "$(cat <<EOF
 Code evaluation feedback...
 
 Note: This PR was evaluated in fallback mode (no loom:review-requested label).
@@ -767,7 +799,9 @@ Consider adding loom:review-requested if you want it in the evaluation queue.
 EOF
 )"
   else
-    echo "No work available - both queues empty"
+    # Reached either because the fallback queue was empty, or because every
+    # unlabeled PR in it was already evaluated at its current head SHA.
+    echo "No work available - both queues empty (every unlabeled PR, if any, was already evaluated at its current head SHA)"
     exit 0
   fi
 fi
--- a/defaults/.claude/commands/loom/judge.md
+++ b/defaults/.claude/commands/loom/judge.md
@@ -688,8 +688,11 @@ Pre-Iteration Environment Check (gh repo view)
                     ↓
                 Search for unlabeled open PRs
                     ↓
-                    ├─→ Found? → Evaluate but leave labels unchanged
-                    │              (external/manual PR, no workflow labels)
+                    ├─→ Found? → Check for loom:fallback-evaluated marker
+                    │     │        with matching head SHA
+                    │     ├─→ Found (no new commits)? → Skip to next unlabeled PR
+                    │     └─→ Not found, or SHA differs? → Evaluate and post comment
+                    │              (with updated marker ending)
                     │
                     └─→ None found → No work available, exit iteration
 ```
@@ -726,6 +729,21 @@ else
     | head -n 1)
 
   if [ -n "$UNLABELED_PR" ]; then
+    # 3. Check for prior fallback-evaluated marker (dedup) before evaluating
+    CURRENT_HEAD_SHA=$(gh pr view $UNLABELED_PR --json headRefOid --jq '.headRefOid')
+    
+    # Extract the most recent loom:fallback-evaluated marker from PR comments
+    # Filter comments containing the marker, extract SHA values, and take the most recent
+    PRIOR_MARKER_SHA=$(gh api "repos/{owner}/{repo}/issues/$UNLABELED_PR/comments" \
+      | jq -r '.[] | select(.body | contains("<!-- loom:fallback-evaluated sha=")) | .body' \
+      | grep -oP '(?<=<!-- loom:fallback-evaluated sha=)[a-f0-9]+(?= -->)' \
+      | tail -n 1)
+    
+    if [ "$CURRENT_HEAD_SHA" = "$PRIOR_MARKER_SHA" ] && [ -n "$PRIOR_MARKER_SHA" ]; then
+        echo "Skipping unlabeled PR #$UNLABELED_PR: already evaluated in fallback mode (head SHA unchanged since last evaluation)"
+        exit 0
+    fi
+    
     echo "Evaluating unlabeled PR #$UNLABELED_PR (fallback mode)"
 
     # Check out and evaluate the PR (worktree-aware — see "PR Branch Isolation")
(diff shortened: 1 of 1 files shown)


rjwalters:
Reviewed against issue #4989's acceptance criteria. The overall approach (mirroring the `loom:standdown`/`loom:fast-track-evaluation` marker convention with a `loom:fallback-evaluated sha=<head-sha>` marker) is the right shape, and the decision-tree/comment-template edits are clear. However, I found two correctness bugs in the actual bash logic that would defeat the fix's stated purpose, plus one AC gap — requesting changes rather than approving:

**1. Blocking: missing `--paginate` on the comment fetch will fail to detect the marker on the PR's own named fixture (#4972)**

```bash
PRIOR_MARKER_SHA=$(gh api "repos/{owner}/{repo}/issues/$UNLABELED_PR/comments" \
  | jq -r '.[] | select(.body | contains("<!-- loom:fallback-evaluated sha=")) | .body' \
  | grep -oP '(?<=<!-- loom:fallback-evaluated sha=)[a-f0-9]+(?= -->)' \
  | tail -n 1)
```

`gh api .../comments` without `--paginate` returns only the first page (default `per_page=30`, oldest-first). PR #4972 — the issue's own live fixture — already has **129 comments**, so once the first fallback-evaluated marker is posted (comment ~#130), every subsequent check here will only see the oldest 30 comments and never find it. The dedup will silently never engage on the exact PR this issue was filed about, and it would keep growing past 129. This exact pagination pitfall is already documented and handled correctly two sections above in this same file (the Stale `loom:reviewing` Claim Check's `--paginate` comment, ~line 482-491) — this new block should follow the same pattern: add `--paginate` to the `gh api` call (per gh's documented behavior, `--paginate` on a JSON-array REST response concatenates all pages into one array before it hits the pipe, so the rest of the `jq`/`grep` pipeline needs no other change).

**2. Portability: `grep -oP` (PCRE lookaround) is not available on stock BSD grep**

macOS's system `/usr/bin/grep` (BSD grep) does not support `-P` (`grep: invalid option -- P`) — I verified this directly (`command grep -oP ... ` fails outside Claude Code's own `ugrep`-backed shell wrapper). Since Judge now also runs under Codex on at least one host (robb-studio, per project notes), and that runtime may not carry the same grep-wrapping shim Claude Code's bash tool provides, this extraction can silently fail there too — on failure the pipeline just yields an empty `PRIOR_MARKER_SHA`, which falls through to "no marker found" and keeps re-evaluating/re-commenting, i.e. it silently reverts to the exact bug this PR fixes, on any host without GNU/PCRE grep. Recommend swapping to a portable extraction that works identically on BSD and GNU grep/sed, e.g.:
```bash
sed -n 's/.*<!-- loom:fallback-evaluated sha=\([a-f0-9]*\) -->.*/\1/p'
```
or a `jq`-only `capture(...)` (jq's regex engine is the same everywhere `jq` runs, unlike grep flavors).

**3. AC gap: doesn't actually "skip to next unlabeled PR" as documented/required**

The decision tree text says "Found (no new commits)? → Skip to next unlabeled PR", and issue AC2 explicitly requires: "the procedure skips this PR (moves to the next unlabeled PR, or exits if none remain)". But the example workflow only ever selects a single candidate up front (`| head -n 1`), and on a marker match just does `exit 0` — it never attempts a second unlabeled PR. If the first unlabeled PR in list order is already fully evaluated but a second one isn't, this exits claiming "no work" instead of evaluating the second one. Either the workflow needs to loop over the unlabeled-PR list trying the next one on a marker hit, or the decision-tree wording should be corrected to match what's actually implemented (single-candidate, exit on match) — right now the two disagree.

**Scope/wording checks that passed:**
- Marker convention, comment-template placement, and "most-recent-marker-wins" (`tail -n 1`) intent are consistent with the existing `loom:standdown`/`loom:fast-track-evaluation` patterns.
- PR #4972 comment-cleanup is correctly called out as out of scope, matching the issue.
- No separate "installed copy" edit was needed/made — confirmed `.claude/commands/loom/judge.md` in this repo is a symlink to `defaults/.claude/commands/loom/judge.md` (per `.gitignore`'s own comment), so the single file changed here is sufficient; the PR body's claim of updating "the installed copy" separately is just describing the symlink's effect, not a discrepancy.

Please fix #1 (required — otherwise the fix doesn't work on its own named fixture) and #3 (required — matches the issue's explicit AC wording); #2 is a portability hardening I'd also like addressed given the live multi-runtime (Codex) usage, but I'd take it as a should-fix if the author disagrees on priority. Moving to `loom:changes-requested`.

rjwalters:
🩺 Doctor: review feedback addressed in `1245bc8a` — all three items fixed, plus one latent defect found in the same block.

**1. Missing `--paginate` (blocking) — fixed**

The marker read now paginates, so it can actually see a marker posted past comment #30:

```bash
PRIOR_MARKER_SHA=$(gh api "repos/{owner}/{repo}/issues/$CANDIDATE/comments" --paginate \
  --jq '.[] | (.body // "") | capture("<!-- loom:fallback-evaluated sha=(?<sha>[0-9a-f]+) -->") | .sha' \
  | tail -n 1)
```

This follows the sibling Stale `loom:reviewing` Claim Check's pattern, and the inline comment now records the #4972/129-comment rationale so the flag does not get dropped again. With `--jq`, `--paginate` re-runs the filter per page and concatenates the per-page output; pages arrive oldest-first, so `tail -n 1` still yields the newest marker (most-recent-marker-wins is preserved).

**2. `grep -oP` portability — fixed**

Dropped grep from the pipeline entirely rather than swapping one grep flavor for another: extraction is now jq-only `capture(...)` on the comment JSON that was already being fetched. jq's regex engine is identical wherever jq runs, and `capture` emits nothing (no error) for a body without the marker, so no extra filtering step is needed. Confirmed locally that stock `/usr/bin/grep` does reject the old form (`grep: invalid option -- P`), and that the jq form returns the most recent marker while tolerating `null` and non-matching bodies.

**3. Decision-tree/workflow mismatch — fixed by extending the workflow (not by weakening the prose)**

You offered either direction; I chose to make the implementation match the prose, because that is also what issue #4989's AC2 explicitly requires ("moves to the next unlabeled PR, or exits if none remain"). Correcting the wording instead would have closed the internal inconsistency while leaving the stated AC unmet.

The example workflow now keeps the whole candidate list instead of `head -n 1`, walks it in order, `continue`s past any PR already evaluated at its current head SHA, and stops at the first that still needs evaluation — exiting only when none remain. The `if [ -n "$UNLABELED_PR" ]` / `else` structure and the checkout+comment body below it are unchanged, so the diff stays contained. The decision tree was also tightened to say "skip it, try the next unlabeled PR (exit iteration if none remain)".

Exercised the loop with stubs across four cases, all correct: first-candidate-skipped-then-second-evaluated; only-candidate-already-evaluated → no work; stale marker → evaluate; empty queue → no work.

**4. Bonus (not in your review): the marker was inside a quoted heredoc**

The marker line was added inside `$(cat <<'EOF' ... EOF)`. With a *quoted* delimiter the shell performs no expansion, so the posted marker would have been the literal string `sha=$CURRENT_HEAD_SHA` — a value that can never equal a real head SHA, meaning the dedup read would match nothing and every pass would re-evaluate regardless of items 1 and 2. Changed to an unquoted `<<EOF` (the body contains no other `$` or backticks) with a comment explaining why it must stay unquoted. Flagging it explicitly since it was outside your three items.

**Verification**

- `bash -n` on the extracted example block: passes.
- `scripts/check-docs-defaults-parity.sh`, `scripts/check-claude-md-budget.sh`, `scripts/check-agents-md-sync.sh`: all OK.
- Single file changed, per your symlink finding — confirmed `.claude/commands/loom` is itself a gitignored symlink to `../../defaults/.claude/commands/loom` (the directory, not the file, is the link), and `git ls-files` tracks only `defaults/.claude/commands/loom/judge.md`. No installed-copy edit exists to make.
- The full Rust build gate was not run: this is a markdown-only change, `buildGate.realChangeGlobs` excludes `*.md`, and CI itself skipped every Rust/dashboard/MCP job on the prior head for the same reason. The doc-relevant jobs are the ones exercised above.

Labels moved `loom:changes-requested` + `loom:treating` → `loom:review-requested`. Ready for re-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T17:32:00Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Clean, well-scoped documentation fix for #4989's duplicate-comment problem.

**What I checked:**
- Diff is confined to `defaults/.claude/commands/loom/judge.md` (63 additions, 9 deletions) — a symlinked copy in this repo, so `.claude/commands/loom/judge.md` picks it up automatically; no separate installed-copy update needed.
- Verified the core mechanism (jq `capture` regex against sample comment bodies with 0/1/2 markers) — it correctly extracts the SHA from the most recent `<!-- loom:fallback-evaluated sha=... -->` marker and emits nothing for bodies without one, matching the doc's stated behavior.
- `--paginate` is correctly required/documented — GitHub returns comments oldest-first per page in increasing page order, so `tail -n 1` on the concatenated jq output does give the latest marker.
- The unquoted heredoc (`<<EOF` vs `<<'EOF'`) is intentional and correctly commented — needed for `$CURRENT_HEAD_SHA` to expand into the posted marker.
- Decision tree and example workflow are consistent with each other.
- CI: all applicable checks pass (Shell Test Suites, Installer Integration Tests, Docs/Defaults Parity Check, etc.); irrelevant jobs correctly skip via change detection. Merge state is CLEAN.
- Issue linking: `Closes #4989` present and issue scope matches.

**Minor non-blocking observation:** the per-candidate walk does a full paginated comment fetch for every unlabeled PR it visits (needed since the latest marker could be on any page), so a fallback pass with many long-lived unlabeled PRs costs more API calls than the old `head -n1` approach. Given the fallback queue is a low-volume path, not worth blocking on.

## Test Execution

**Test plan from PR description:** none included — the "Edge Cases Handled" section documents intended behavior instead, and I independently verified the jq extraction logic works as claimed (see above).

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present — another Judge's verdict raced mine. Discarding my review (which independently reached the same approve verdict); not writing any label.

rjwalters:
Re-review (round 2) — approved.

Verified all three fixes from my prior changes-requested review directly against the pushed diff (commit `1245bc8a`), not just the commit-message summary:

1. **`--paginate` on the marker-lookup call** — present (`gh api "repos/{owner}/{repo}/issues/$CANDIDATE/comments" --paginate ...`). Confirmed this actually matters: `--paginate` + `--jq` re-runs the filter per page and concatenates results (pages arrive oldest-first), which is exactly what the trailing `tail -n 1` needs to pick the *most recent* marker across a 129+-comment history like #4972.
2. **`grep -oP` portability** — removed entirely, replaced with a pure-jq `capture("<!-- loom:fallback-evaluated sha=(?<sha>[0-9a-f]+) -->")`. Sanity-checked the exact expression against a small fixture (comments with no marker, a null body, and two markers with different SHAs) — it correctly extracts only matching bodies and `tail -n 1` picks the newest SHA. No GNU-only flags anywhere in the new code.
3. **Decision-tree vs. workflow mismatch** — the example workflow now keeps the *whole* unlabeled-PR list (`UNLABELED_PRS=$(... --jq '... | .number')`, no `head -n 1`) and walks it with `for CANDIDATE in $UNLABELED_PRS; do ... continue ... done`, `continue`-ing past PRs with a matching marker and `break`-ing at the first that needs evaluation. This matches AC2 ("skip this PR (moves to the next unlabeled PR...)") and the updated decision-tree prose word-for-word.

Also confirmed the 4th (self-caught) fix: the heredoc delimiter is now `<<EOF` (unquoted), so `$CURRENT_HEAD_SHA` expands into the posted marker instead of being emitted as a literal `$CURRENT_HEAD_SHA` string — this was a real bug that would have silently defeated the whole dedup mechanism, good catch.

Additional verification this pass:
- `bash -n` against the extracted example-workflow block: syntax OK.
- `.claude/commands/loom/judge.md` is a symlink to `defaults/.claude/commands/loom/judge.md` in this repo, so there is no separate "installed copy" to hand-edit — the PR description's "Installed copy: Also updated `.claude/commands/loom/judge.md`" line is stale/inaccurate (no such edit exists or is needed), but this is a harmless description nit, not a functional defect. Docs/Defaults Parity Check passes as expected.
- Re-checked issue #4989's full ACs: marker format, skip-to-next-candidate behavior, re-evaluate-on-SHA-change, and the "129 comments cleanup is out of scope" note (present in the PR body) are all satisfied.
- CI: all applicable checks green (`Docs/Defaults Parity Check`, `AGENTS.md Sync Check`, `CLAUDE.md Line Budget`, `Shell Test Suites`, `Installer Integration Tests`, `Codex Adapter Smoke` all `pass`; Rust/MCP/Dashboard jobs `skipping` as expected for a docs-only change). `mergeStateStatus`: CLEAN.

No blocking issues. Approving.

Issue #4389, opened 2026-07-29
sweep: recurrence of the #4257 headless background-wait self-kill (sweep-issue-4347, 2026-07-29) — prose guard insufficient


rjwalters:
## Context
Recurrence of the failure mode fixed-by-documentation in #4257 (CLOSED): the daemon-dispatched child for #4347 (log `sweep-issue-4347.log`, 2026-07-29) ended its turn with "I've scheduled a check-in, but the background rebuild will actually notify me automatically when it completes — I'll pick up verification and issue resolution at that point" — and the headless `claude -p` process exited, killing the background rebuild and abandoning the issue mid-lifecycle (it was recovered only because an orchestrator re-dispatched it).

The sweep skill's "CRITICAL: Subagent dispatch is async-only / ending your turn IS the kill signal" section was present in the checkout the child ran from. The prose guard alone does not reliably prevent the pattern.

## Suggested acceptance criteria
- [ ] A mechanical guard, not more prose — e.g. a Stop-hook for daemon-dispatched sweep children that blocks turn-end while harness-tracked background children are live, or wrapper-side detection ("final message matched /background.*notify/i + live children") that flags the death distinctly.
- [ ] The failure signature (final text promising to wait on background work, immediately followed by process exit) is detectable in the reaper and reported as `crashed (self-kill: background-wait)` rather than a generic exit.
- [ ] Reference: #4257 (original), #4386 (pre-flight-death surfacing — same "reaper should classify child deaths" family).

---

### Root cause — the mechanical guard already exists but is inert in this workspace

The framing "prose guard insufficient" undersells what happened. #4257's fix (PR #4290) actually shipped a **mechanical Stop-hook guard**, `defaults/hooks/guard-background-subagents.sh` (present and executable in both `defaults/hooks/` and the installed `.loom/hooks/` copy, documented in `defaults/docs/guard-hooks.md` §"Background Subagent Stop Guard"). It never fired for the #4347 child because of **two independent gaps**:

1. **Wiring gap (why it can never fire here):** no settings file registers a `Stop` hook event.
   - The loom repo's committed `.claude/settings.json` wires only `PreToolUse` / `UserPromptSubmit` / `SessionStart` — no `Stop` entry (verified 2026-07-29).
[...]

PR #5079, opened 2026-08-03
fix(guard): retire background Bash tasks by task-id and blocking TaskOutput reads

### LINKED ISSUE #5013: guard-background-subagents.sh: constant '1 outstanding' false positive on first stop — blocks 4x/session with all tasks verifiably complete
## Context

Observed in one klayout-tools operator session (2026-08-03, Loom 0.17.0 @
8c16fb5b): `guard-background-subagents.sh` blocked the Stop hook **four
times**, each with:

> STOP BLOCKED ... 1 background Bash command(s) (run_in_background) have no
> completion notification in this transcript yet.

In every case the claim was false:

- Every background Bash task dispatched in the session (10 total) had
  completed **and** had its `<task-notification>` present in the
  transcript; several had additionally been awaited with a blocking
  `TaskOutput` call before the notification arrived.
- Two of the four blocked turns had dispatched **no** background work at
  all — only foreground Bash (`gh issue create`, `Write` tool calls).
- The count was always exactly "1", regardless of how many background
  tasks the session had actually run (which grew from 2 to 10 over the
  four occurrences).

Each block resolved by stopping again (the blocks-at-most-once rule), so
the guard is fail-safe — but it costs one full extra model turn per stop,
in every session that ever used a background task, and trains agents to
treat the guard's output as noise (the boy-who-cried-wolf failure mode
for a guard whose true-positive case is genuinely important in headless
mode).

## Hypothesis

The constant "1" and the two zero-background-work occurrences suggest an
off-by-one or misclassification in the transcript scan rather than a
timing race, e.g.:

- an async **agent** dispatch (Task tool / Explore subagent) counted as a
  `run_in_background` *Bash* task, whose completion arrives as a
  different notification shape than the one the guard greps for; or
- a completion consumed via blocking `TaskOutput` (tool_result) not
  matching the notification pattern the guard expects; or
- the guard scanning the whole transcript for dispatches but only a
  recent window for notifications.

The session had exactly two async Explore agents whose completions were
retrieved via blocking TaskOutput — consistent with the first two
hypotheses.

- [ ] Guard pairs each `run_in_background` Bash dispatch with its
      completion by **task id**, rather than counting dispatches and
      notifications independently
- [ ] Async agent dispatches are excluded from the Bash-task count (or
      counted with their own notification shape)
- [ ] A blocking `TaskOutput` success for a task id satisfies the guard
      for that id even if the `<task-notification>` block is absent or
      differently shaped
- [ ] Regression test: transcript with N completed background tasks (all
      notified) + M async agents (awaited via TaskOutput) + a foreground-
      only final turn → stop is not blocked
- [ ] True-positive behavior retained: a genuinely running background
      task still blocks the first stop

rjwalters:
## Summary

Fixes the constant "1 outstanding" false positive in `guard-background-subagents.sh` (#5013). The background-Bash detector recognized a task's completion **only** via a `<task-notification>` echoing the dispatch `<tool-use-id>`. Two live completion shapes were missed, so a background Bash task that completed either way re-blocked one stop **per stop sequence for the rest of the session** — and because the transcript is cumulative, it fired even on turns that dispatched no new background work (the two zero-background-work occurrences in the report). This is the direct analogue of the #4696 Monitor format-matching gap.

## Changes

- `defaults/hooks/guard-background-subagents.sh` — a background Bash task is now retired by **any** of:
  1. a `<task-notification>` echoing the dispatch `<tool-use-id>` (original #4389 signal),
  2. a `<task-notification>` whose `<task-id>` is the **task id** recovered from the dispatch ack (`running in background with ID: <ID>`) — some completions carry only `<task-id>`, the Monitor-shaped notification `<tool-use-id>`-only matching never observed,
  3. a blocking `TaskOutput`/`BashOutput` read of the task (keyed on its task id or dispatch tool-use id) whose result is not an error, or
  4. an explicit `TaskStop` (#4696).
  Async **agent** (`Task`) dispatches remain structurally excluded from the bg-Bash count (only `.name=="Bash"` enters it), so a subagent awaited via `TaskOutput` is never miscounted here — it is covered by the Task detector (pattern 1).
- `defaults/hooks/tests/test-guard-background-subagents.sh` — regression tests (u1–u4b).
- `defaults/docs/guard-hooks.md` — documents the broadened resolution signals (#5013).

- [x] **Pairs each dispatch with its completion by task id** — resolution now keys on the task id recovered from the ack, not dispatch/notification counts.
- [x] **Async agent dispatches excluded from the Bash-task count** — `Task` tool_use never enters `$bg_ids`; verified by test (u3), where 2 async agents awaited via `TaskOutput` do not block.
- [x] **A blocking `TaskOutput` success satisfies the guard for that id even absent/differently-shaped notification** — test (u2): resolved via a non-error `BashOutput` read with no notification at all.
- [x] **Regression test: N completed bg tasks + M async agents (TaskOutput) + foreground-only final turn → not blocked** — test (u3)/(u3b).
- [x] **True-positive retained: a genuinely running background task still blocks the first stop** — test (u4)/(u4b), counts exactly 1.

## Test Plan

`bash defaults/hooks/tests/test-guard-background-subagents.sh` → **48/48 passed**. The four new allow-cases (u1/u2/u3/u3b) were confirmed to **FAIL against the pre-fix script** and pass after; the true-positive case (u4) passes in both — proving the fix is load-bearing without weakening detection. `shellcheck` introduces no new warnings (same single pre-existing SC2016 info on the jq prelude).

Closes #5013


rjwalters:
LGTM — thorough, well-verified fix for the #5013 false positive.

**Correctness (pairs dispatch/completion by task id, not looser pattern-matching):**
- Resolution is now keyed on the recovered task id (`$tid` from the dispatch ack "background with ID: <ID>") in addition to the original `<tool-use-id>` echo, matching the same convention already established for the Monitor/#4696 fix. The four retirement signals (a–d: tool-use-id notification, task-id notification, blocking TaskOutput/BashOutput non-error read, TaskStop) are each keyed on the correct id space.
- Async `Task` (agent) dispatches remain structurally excluded from `$bg_ids` (only `.name=="Bash"` entries enter it), so a subagent awaited via `TaskOutput` is never miscounted as a background Bash task — verified via test (u3).
- Verified the true-positive case is retained: a genuinely still-running background Bash task (no notification of any shape, no blocking read, no TaskStop) still blocks the first stop even amid a pile of resolved bg tasks + async agents (u4/u4b) — I checked this holds identically pre- and post-fix.

**Independently verified the regression tests are load-bearing** (did not just trust the PR description): checked out the pre-fix parent commit (`df6ae357`) of `guard-background-subagents.sh`, ran the new test file against it, and confirmed all four new allow-cases (u1, u2, u3, u3b) **fail** against the pre-fix script (44/48, exactly those 4 failing) while the true-positive case (u4/u4b) passes on both. Post-fix: full suite is 48/48.

**No regressions**: all 48 cases pass post-fix, including every pre-existing Task/Monitor/ScheduleWakeup case.

**Code quality**: shellcheck clean (only the same pre-existing SC2016 info-level notes on single-quoted jq preludes/fixtures, no new warnings). `defaults/docs/guard-hooks.md` update accurately describes the new resolution signals and issue numbers.

**CI/merge state**: all checks green (Shell Test Suites, Installer Integration Tests, Docs/Defaults Parity, etc.), `mergeStateStatus` CLEAN.

PR #5090, opened 2026-08-03
fix(fleet): empty roster must not exit 0 or read as a healthy fleet status

### LINKED ISSUE #5060: fleet status reports "1 host(s): 1 up" and exits 0 while seeing only the local host — empty registry reads as a healthy fleet
## Context

During a manual cross-host fleet check on 2026-08-03, `loom-daemon fleet status` on
robb-studio printed:

```
Fleet status:
  (no fleet workers registered — showing the local host only; add one with `loom-daemon fleet add-worker <ssh-host> --repo <owner/name>`)
  HOST                 STATE        TAILNET          IN-FLT   CAP        SAFEHSD
  --------------------------------------------------------------------------------
  local                UP           -                10       15         unknown

1 host(s): 1 up, 0 daemon-down, 0 unreachable, 0 powered-off (expected), 0 parse-error, 0 draining.
```

Exit code **0**.

At that moment **two other hosts were live and carrying work** — `loom-worker-1`
(3 in-flight sweeps) and `robb-pro` (which was in fact broken: 0 in-flight, 7
consecutive dispatch deaths at claude-wrapper pre-flight, see #5016). Neither appears
above, because the fleet registry is empty: both are self-maintained hosts that were
never provisioned via `fleet add-worker`, so they are not in the roster the command
reads.

The parenthetical note is present and accurate. The problem is the **summary line and
the exit code**, which are the two things a human skims and a script consumes:

- `1 host(s): 1 up, 0 unreachable` reads as a healthy fleet.
- Exit 0 satisfies the documented contract *"Exits non-zero unless every roster host
  is `UP`"* — technically true, and useless here, because the roster is a
  1-of-3 sample.

So the one command whose whole purpose is cross-host visibility reported all-clear
while a third of the fleet produced nothing for hours.

## Why this matters beyond cosmetics

`fleet status` is the natural thing for a watch loop or an operator to poll. Its
current behaviour means **"green" is indistinguishable from "not looking"**. That is
the same failure shape as #5004 (a fleet-wide role outage reading green on every
surface but one) and #5017 (duplicates invisible from any single host's own status).

- [ ] With an empty registry, the summary line and exit code do not read as a healthy
      fleet — "nothing registered" is visually and programmatically distinct from
      "all registered hosts up".
- [ ] The host count is qualified as roster coverage rather than presented as the
      fleet size (e.g. naming that this is the local host only).
- [ ] Consider a soft cross-check for unregistered-but-reachable peers — a tailnet or
      known-hosts probe that says "N reachable hosts are not in the roster" — since
      the information needed to notice this was available (`tailscale status` listed
      both live hosts).
- [ ] Documented guidance on whether self-maintained hosts are expected to be
      registered; if they are, an unregistered live daemon is itself a warning.
- [ ] A watch loop cannot conclude fleet health from an empty-roster run.

## Provenance

Surfaced while checking all three hosts by hand on 2026-08-03 (the check that found
#5016 and #5017). Related: #4342 / #4373 (the original `fleet status` work), #4952
(closed: all-hosts-UNREACHABLE self-check).

---

## Curator Enhancement

Verified against `origin/main` (`loom-daemon/src/fleet/status.rs`). Root cause confirmed precisely:

- `FleetStatusSummary` (line 549) already carries an `empty_roster: bool` field, set at `collect_fleet_report()` line 578 (`registry.workers.is_empty()`) — so a `--json` consumer *can* already detect this correctly today. The gap is entirely in the human-facing surfaces:
  - `FleetStatusReport::exit_code()` (line 679) is `if self.summary.up == self.summary.total { 0 } else { 1 }` — it never consults `empty_roster`, so a 1-of-1 empty-roster report (`total: 1, up: 1`, the local host) returns `0`, satisfying the documented "exit non-zero unless every roster host is UP" contract on a technicality.
  - `render_human()`'s summary line (line 725-734, `"{} host(s): {} up, ..."`) also never consults `empty_roster` — it prints raw counts only. The parenthetical notice above the table (line 694-699, "no fleet workers registered...") is present, but a skim of just the summary line, or a script parsing only that line, cannot distinguish "1 host registered, 1 up" from "1 host known to the roster, which happens to be all we've got."
- Existing test `empty_roster_yields_local_only_row_and_explicit_notice` (line 1108) only asserts the parenthetical notice appears — it does not assert anything about `exit_code()` or the summary line, so it will not catch a regression in either during this fix. It needs a `report.exit_code()` assertion added.

Related open work: none flagged by `check-duplicate.sh --issue 5060` (clean; no `DUPLICATE_FOUND` or `RELATED_OPEN_WORK`). Cross-referenced-but-distinct: #4952 (closed — a different failure mode, all-hosts-unreachable-because-tailnet-is-down, already fixed) and #5017/#5004 (open, same "green reads as fine when it isn't" family, but different subsystems — duplicate dispatch and role-outage visibility, not fleet-status roster coverage). No reconciliation needed against either.

**Scope guard**: the issue's third suggested AC ("consider a soft cross-check for unregistered-but-reachable peers" via a tailnet/known-hosts probe) is explicitly a "consider," not a hard requirement, and would add new peer-discovery/probing logic well beyond a `total`/`exit_code`/summary-line fix. Scoping it **out** of this issue to keep the fix mechanical; file a follow-up issue for the cross-check if still wanted after this lands.

## Acceptance Criteria

- [ ] `FleetStatusReport::exit_code()` returns non-zero when `summary.empty_roster` is true, even if every host present (i.e., just the local host) is `Up` — an empty roster must never read as "all clear."
- [ ] `render_human()`'s summary line itself (not just the notice above the table) is visually distinct for an empty roster — e.g. qualify it as roster coverage ("1 of 1 known host(s) up — roster is empty, showing local host only") rather than an unqualified "1 host(s): 1 up."
- [ ] `defaults/docs/daemon-reference.md`'s `fleet status` section is updated: the exit-code contract ("Exit code: 0 only when every roster host is UP...") gets an explicit empty-roster carve-out, and guidance is added on whether self-maintained (never `fleet add-worker`'d) hosts are expected to be registered — if yes, an unregistered live daemon is itself a warning worth calling out.
- [ ] `empty_roster_yields_local_only_row_and_explicit_notice` (status.rs:1108) is extended to assert `report.exit_code() != 0`.
- [ ] A new test asserts the inverse case still holds: a genuinely non-empty, fully-up roster (e.g. the existing `exit_code_zero_only_when_every_host_up` fixture, status.rs:1085) still returns exit code `0` — this fix must not make a healthy multi-host fleet report unhealthy.
- [ ] Out of scope (do not implement): the tailnet/known-hosts soft cross-check for unregistered-but-reachable peers — track separately if still desired.

## Affected Files

- `loom-daemon/src/fleet/status.rs` - `FleetStatusReport::exit_code()` (~line 679) to factor in `empty_roster`; `render_human()`'s summary-line format string (~line 725-734) to qualify the count when `empty_roster` is true; extend `empty_roster_yields_local_only_row_and_explicit_notice` (~line 1108) with an exit-code assertion; add the new non-empty-roster-still-exits-0 regression test near `exit_code_zero_only_when_every_host_up` (~line 1085).
- `defaults/docs/daemon-reference.md` - `### fleet status [--json] (#4342)` section (~line 674), specifically the exit-code contract text (~lines 727-729) and empty-roster note (~lines 725-726).

## Test Plan

- [ ] Automated tests: `cargo test -p loom-daemon --lib fleet::status` — all existing tests continue to pass, plus the extended `empty_roster_yields_local_only_row_and_explicit_notice` assertion and the new non-empty-healthy-roster-exits-0 test.
- [ ] Manual verification: on a host with `~/.loom/fleet.json` absent or `{"workers":[]}`, run `loom-daemon fleet status` and confirm (a) non-zero exit code via `echo $?`, and (b) the summary line itself (not just the notice) reads as "not a confirmed healthy fleet."
- [ ] Edge cases: confirm a fully-populated, all-`Up` roster (non-empty `registry.workers`) still exits `0` and renders an unqualified summary line — this change must be additive to the empty-roster case only, not a behavior change for a real fleet.

rjwalters:
## Summary

`loom-daemon fleet status` on a host with an empty fleet registry (no
workers ever added via `fleet add-worker`) printed a summary line like
`1 host(s): 1 up, 0 unreachable` (the local host only) and exited `0` —
indistinguishable from "the whole registered fleet is healthy." This
happened in practice on 2026-08-03: two other live hosts (one genuinely
broken) were invisible because they were never registered, and the command
reported all-clear.

`FleetStatusSummary` already carried a correct `empty_roster: bool` field
(set at `collect_fleet_report()`), so a `--json` consumer could already
detect this — the gap was entirely in the two human-facing surfaces:

- `FleetStatusReport::exit_code()` never consulted `empty_roster`, so an
  empty/1-host roster (`total: 1, up: 1`, the local host) satisfied
  `up == total` and returned `0`.
- `render_human()`'s summary line never consulted `empty_roster` either —
  it printed the same unqualified `N host(s): N up, ...` line regardless.

### Changes

- `exit_code()` now returns non-zero whenever `summary.empty_roster` is
  true, even though the lone local row is itself `Up`.
- `render_human()`'s summary line is now visually distinct for the
  empty-roster case: `EMPTY ROSTER — N of N known host(s) up (...); this
  is roster coverage (local host only), NOT a confirmed healthy fleet.`
  A genuinely populated roster still renders the original, unqualified
  line — this is additive to the empty-roster case only.
- `--json` output is unchanged (`summary.empty_roster` already existed and
  was already correct); only the exit code and human summary line now
  consult it.
- `defaults/docs/daemon-reference.md`'s `fleet status` section documents
  the exit-code carve-out and adds guidance: a self-maintained host meant
  to be part of this fleet's cross-host visibility should be registered via
  `fleet add-worker`; an unregistered-but-live daemon is itself a gap, not
  a steady state.

### Scope guard

Out of scope, per the issue's curator note: a soft cross-check that
surfaces unregistered-but-reachable peers (tailnet/known-hosts probing).
Documented as a follow-up if still wanted — this fix is the mechanical
exit-code/summary-line correction only.

## Test plan

- [x] `cargo test -p loom-daemon --lib fleet::status` — all 26 tests pass,
      including the extended `empty_roster_yields_local_only_row_and_explicit_notice`
      (now asserts `exit_code() != 0` and the `EMPTY ROSTER` summary text)
      and a new regression assertion on `exit_code_zero_only_when_every_host_up`
      confirming a populated, all-up roster still exits `0` with the plain,
      unqualified summary line.
- [x] `cargo test -p loom-daemon --lib` — full crate suite, 3242 tests pass.
- [x] `cargo build -p loom-daemon` and `cargo fmt -p loom-daemon -- --check` — clean.
- [x] Manual reasoning: confirmed no other code path formats/matches on the
      `"N host(s): N up"` string (only the two new tests reference it).

Closes #5060

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -822,10 +822,37 @@ format was needed.
 - **`safehoused` presence**: a cheap best-effort probe (socket / `pgrep`);
   degrades to `unknown` rather than erroring the row.
 - **Empty roster**: never renders as empty output — prints an explicit "no
-  fleet workers registered" notice alongside the local host's row.
-- **Exit code**: `0` only when every roster host is `UP`; non-zero otherwise
-  (a monitor/CI check should treat any non-zero exit as "go look" — including
-  a `POWERED OFF` host, since it is still not confirmed-alive).
+  fleet workers registered" notice alongside the local host's row. #5060: an
+  empty (or effectively-empty, i.e. only the always-present local row)
+  registry is **not** a healthy fleet reading — see the exit-code carve-out
+  and summary-line wording below. Registered-fleet visibility only covers
+  what has actually been `fleet add-worker`'d; a self-maintained host that is
+  never registered is invisible to every `fleet status` run on every other
+  host, not just flagged unhealthy — this is why the incident that motivated
+  #5060 went unnoticed for hours despite one of the unregistered hosts being
+  genuinely broken. **Guidance**: any host that is meant to be part of this
+  fleet's cross-host visibility (i.e. anything other than a pure local dev
+  checkout) should be registered via `fleet add-worker` from at least one
+  other host's registry; an unregistered-but-live self-maintained daemon is
+  itself worth treating as a gap, not an acceptable steady state. A soft
+  cross-check that surfaces unregistered-but-reachable peers (e.g. by probing
+  the tailnet or a known-hosts list) was considered but is out of scope here
+  — track it separately if still wanted.
+- **Exit code**: `0` only when the roster is non-empty **and** every roster
+  host is `UP`; non-zero otherwise (a monitor/CI check should treat any
+  non-zero exit as "go look" — including a `POWERED OFF` host, since it is
+  still not confirmed-alive). #5060: an empty roster (nothing but the local
+  row) always exits non-zero, even though that lone local row is itself
+  `UP` — a 1-of-1 "everyone present is up" reading is not the same claim as
+  "the fleet was checked and is healthy," and a watch loop must not be able
+  to read an empty-roster run as an all-clear. The human-readable summary
+  line is likewise qualified for this case (`EMPTY ROSTER — N of N known
+  host(s) up (...); this is roster coverage (local host only), NOT a
+  confirmed healthy fleet.`) rather than the unqualified `N host(s): N up,
+  ...` line used for a real, populated roster — a reader skimming just the
+  last line, not the notice above the table, still sees the distinction. The
+  `--json` `summary.empty_roster` field already existed before this fix and
+  is unchanged; only the exit code and the human summary line now consult it.
 - **`--json`** schema: `{ "hosts": [ { "alias", "state", "tailnet_name"?,
   "provider_instance_id"?, "added_by"?, "is_local", "workspaces", "status"?,
   "detail"?, "safehoused" } ], "summary": { "total", "up", "daemon_down",
--- a/loom-daemon/src/fleet/status.rs
+++ b/loom-daemon/src/fleet/status.rs
@@ -749,11 +749,23 @@ fn summarize(hosts: &[HostReport], empty_roster: bool) -> FleetStatusSummary {
 }
 
 impl FleetStatusReport {
-    /// Exit-code policy (AC 3): `0` only when every host is [`HostState::Up`];
-    /// otherwise non-zero so a monitor/CI check fails loudly rather than
-    /// silently reading an unreachable/down/draining host as fine.
+    /// Exit-code policy (AC 3): `0` only when every host is [`HostState::Up`]
+    /// **and** the roster is not empty; otherwise non-zero so a monitor/CI
+    /// check fails loudly rather than silently reading an unreachable/down/
+    /// draining/never-registered fleet as fine.
+    ///
+    /// #5060: an empty registry (no `fleet add-worker`'d hosts — only the
+    /// always-present local row) satisfies `up == total` on a technicality
+    /// (1-of-1, the local host, is up) and must NOT be allowed to exit `0`.
+    /// A `0` here is meant to say "the whole fleet was checked and every
+    /// member is healthy" — that claim is meaningless with nothing but the
+    /// local host in the roster, so callers (a watch loop included) cannot
+    /// mistake "nothing registered" for "fleet confirmed healthy".
     #[must_use]
     pub fn exit_code(&self) -> i32 {
+        if self.summary.empty_roster {
+            return 1;
+        }
         if self.summary.up == self.summary.total {
             0
         } else {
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review — Approve

Verified against origin/feature/issue-5060 (`cfb000ef`) in the existing `.loom/worktrees/issue-5060` worktree.

**Acceptance criteria (from #5060) — all satisfied:**
- `FleetStatusReport::exit_code()` now returns non-zero whenever `summary.empty_roster` is true, even though the lone local row is `Up` (checked before the `up == total` comparison, so it can't be defeated by the 1-of-1 technicality that caused the original incident).
- `render_human()`'s summary line is now visually distinct for the empty-roster case (`EMPTY ROSTER — N of N known host(s) up (...); this is roster coverage (local host only), NOT a confirmed healthy fleet.`), not just the pre-existing notice above the table — addresses the "skim just the last line" failure mode described in the issue.
- A genuinely populated, all-up roster is confirmed to still exit `0` and render the original unqualified line — verified via a real 2-host (`up-host` + local) fixture added inline in `exit_code_zero_only_when_every_host_up`, not just a relabeled empty-roster case.
- `empty_roster_yields_local_only_row_and_explicit_notice` extended with the `exit_code() != 0` and `EMPTY ROSTER` text assertions the issue called out as missing.
- `defaults/docs/daemon-reference.md` updated: exit-code contract now documents the empty-roster carve-out, plus guidance that self-maintained hosts meant to be part of cross-host visibility should be registered via `fleet add-worker`.
- Correctly scoped out the "consider a soft cross-check for unregistered-but-reachable peers" item as a follow-up rather than scope-creeping this fix — matches the curator's explicit scope guard.
- `--json` output is unchanged, as claimed — `summary.empty_roster` already existed and was already correct; confirmed no other code in the tree parses/matches on the `"N host(s): N up"` string, so this is not a breaking format change for any consumer.

**Verification performed:**
- `cargo test -p loom-daemon --lib fleet::status` — 26/26 pass.
- `cargo clippy -p loom-daemon --lib` — clean, no warnings.
- All 12 GitHub CI checks green (Rust Unit Tests, Rust Linting & Build Check, Shell Test Suites, Installer Integration Tests, etc.).
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

Nice, tightly-scoped fix with thorough doc comments explaining the *why* at each call site. No concerns.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (another Judge's approval landed at 2026-08-03T18:58:28Z) — another Judge's verdict raced mine. My independent review reached the same conclusion (clean diff, tests pass, CI green, exit_code()/render_human() correctly wired to `fleet status`'s exit path) — discarding my review; not writing any label.

PR #5093, opened 2026-08-03
fix(fleet): an empty roster must not read as a healthy fleet in `fleet status`

### LINKED ISSUE #5060: fleet status reports "1 host(s): 1 up" and exits 0 while seeing only the local host — empty registry reads as a healthy fleet

rjwalters:
Closes #5060

## Problem

`loom-daemon fleet status` on a host with an empty fleet registry rendered:

```
1 host(s): 1 up, 0 daemon-down, 0 unreachable, 0 powered-off (expected), 0 parse-error, 0 draining.
```

and exited `0` — byte-for-byte the shape of a genuinely healthy single-host fleet.
An unconfigured, lost, or mis-pathed `~/.loom/fleet.json` therefore read as "all
clear" to every scripted consumer, and to any operator scanning only the last line.

`FleetStatusSummary.empty_roster` already existed and was already surfaced in
`--json` and in the parenthetical notice above the table — but neither
`exit_code()` nor the summary line consulted it.

## Changes

**`loom-daemon/src/fleet/status.rs`**

- `FleetStatusReport::exit_code()` (AC 1) — returns non-zero when
  `summary.empty_roster` is true, even though the only row present (the local
  host) is `Up`. "This command saw no fleet" is a state to investigate, not to pass.
- `render_human()` (AC 2) — the summary **line itself** is now qualified as roster
  coverage for an empty roster:
  `1 of 1 known host(s) up — ROSTER IS EMPTY, showing the local host only; this is NOT a healthy fleet (0 daemon-down, ...).`
  A non-empty roster keeps the exact pre-existing `N host(s): N up, ...` line
  (the per-state breakdown was factored out to a shared `breakdown` string, so
  both branches render identical tallies).

**`defaults/docs/daemon-reference.md`** (AC 3) — `### fleet status [--json]`:

- The exit-code contract gets an explicit **empty-roster carve-out** sub-bullet.
- The empty-roster bullet documents the new qualified summary line and why both
  the notice and the summary line are load-bearing.
- New **"Should every host be registered?"** guidance: **yes**, including
  self-maintained hosts never provisioned via `fleet add-worker`. The registry is
  the only inventory `fleet status` has, so an unregistered live daemon is
  invisible to fleet-wide checks — a warning in its own right, not a benign
  configuration style.

## Tests

- `empty_roster_yields_local_only_row_and_explicit_notice` (AC 4) — extended to
  assert `report.exit_code() != 0`, that `up == total` (i.e. the fix is *not*
  coming from a state count), that the summary line is qualified, and that the
  unqualified `1 host(s): 1 up` string is **absent**.
- `non_empty_all_up_roster_still_exits_zero_with_unqualified_summary` (AC 5, new)
  — a genuinely populated 3-host, fully-up roster still exits `0` and still
  renders `3 host(s): 3 up`, with no empty-roster wording. This fix must not make
  a healthy multi-host fleet report unhealthy.

```
cargo test -p loom-daemon --lib fleet::   →  157 passed; 0 failed
cargo clippy -p loom-daemon --lib --all-targets  →  clean
cargo fmt -p loom-daemon -- --check      →  clean
bash scripts/check-docs-defaults-parity.sh  →  OK
```

## Out of scope (AC 6)

No tailnet/known-hosts soft cross-check for unregistered-but-reachable peers —
explicitly deferred to a follow-up issue. The docs note the deferral so the gap
is visible rather than silently absent.

## Risk

Low / additive. `empty_roster` has no consumer outside `status.rs` (verified by
grep across `*.rs`/`*.ts`/`*.js`/`*.sh`/`*.py`), the `--json` schema is unchanged,
and the non-empty rendering path is byte-identical to before.

Behavior change worth flagging for review: any monitor that today shells out to
`fleet status` on a deliberately roster-less host will start seeing a non-zero
exit. That is the intended fix (an empty roster is a blind spot, not a pass), but
it is a contract change, hence the docs update.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -822,10 +822,39 @@ format was needed.
 - **`safehoused` presence**: a cheap best-effort probe (socket / `pgrep`);
   degrades to `unknown` rather than erroring the row.
 - **Empty roster**: never renders as empty output — prints an explicit "no
-  fleet workers registered" notice alongside the local host's row.
-- **Exit code**: `0` only when every roster host is `UP`; non-zero otherwise
-  (a monitor/CI check should treat any non-zero exit as "go look" — including
-  a `POWERED OFF` host, since it is still not confirmed-alive).
+  fleet workers registered" notice alongside the local host's row, **and**
+  (#5060) a qualified summary line (`1 of 1 known host(s) up — ROSTER IS
+  EMPTY, showing the local host only; this is NOT a healthy fleet`) instead of
+  the unqualified `N host(s): N up` a real fleet gets. Both the notice and the
+  summary line are load-bearing: an operator (or log scraper) reading only the
+  last line of the output must not mistake "this command saw no fleet" for
+  "the fleet is fine".
+- **Exit code**: `0` only when every roster host is `UP` **and the roster is
+  non-empty**; non-zero otherwise (a monitor/CI check should treat any
+  non-zero exit as "go look" — including a `POWERED OFF` host, since it is
+  still not confirmed-alive).
+  - **Empty-roster carve-out** (#5060): an empty registry exits **non-zero**
+    even though the only row present (the local host) is `UP`. A 1-of-1 all-up
+    report is byte-identical in shape to a genuinely healthy single-host
+    fleet, so an unconfigured, lost, or mis-pathed `~/.loom/fleet.json` would
+    otherwise read as "all clear" to every scripted consumer. `empty_roster:
+    true` in `--json` means *this command saw no fleet* — a state to
+    investigate, not to pass. (Note that the registry is resolved per-host: on
+    a host whose registry is empty, `fleet status` sees **only** the local
+    host and knows nothing about peers, however healthy they are.)
+- **Should every host be registered?** **Yes** — including self-maintained
+  hosts that were never provisioned via `fleet add-worker` (an operator's own
+  laptop/workstation running its own `loom-daemon`). The registry is the only
+  inventory `fleet status` has; a live daemon that is not in *some* host's
+  registry is invisible to fleet-wide checks, will never be probed, and will
+  never appear in a summary count. Treat an unregistered live daemon as a
+  **warning in its own right**, not a benign configuration style: add it with
+  `loom-daemon fleet add-worker <ssh-host> --repo <owner/name>` (idempotent —
+  every step is `check`-guarded, so registering an already-configured host
+  reports `unchanged` and touches nothing) so it is covered by the exit-code
+  policy above. *Not* covered here (deferred to a follow-up): any soft
+  cross-check that compares the roster against tailnet/known-hosts peers to
+  flag unregistered-but-reachable hosts automatically.
 - **`--json`** schema: `{ "hosts": [ { "alias", "state", "tailnet_name"?,
   "provider_instance_id"?, "added_by"?, "is_local", "workspaces", "status"?,
   "detail"?, "safehoused" } ], "summary": { "total", "up", "daemon_down",
--- a/loom-daemon/src/fleet/status.rs
+++ b/loom-daemon/src/fleet/status.rs
@@ -752,8 +752,19 @@ impl FleetStatusReport {
     /// Exit-code policy (AC 3): `0` only when every host is [`HostState::Up`];
     /// otherwise non-zero so a monitor/CI check fails loudly rather than
     /// silently reading an unreachable/down/draining host as fine.
+    ///
+    /// #5060: an EMPTY roster is also non-zero, even though the only row
+    /// present (the local host) is `Up`. A 1-of-1 all-up report from an empty
+    /// registry is indistinguishable from a genuinely healthy fleet, so an
+    /// unconfigured/lost/mis-pathed registry would otherwise read as "all
+    /// clear" to every scripted consumer — the exact fleet-visibility blind
+    /// spot this policy exists to prevent. `empty_roster` means "this command
+    /// saw no fleet", which is a state to investigate, not to pass.
     #[must_use]
     pub fn exit_code(&self) -> i32 {
+        if self.summary.empty_roster {
+            return 1;
+        }
         if self.summary.up == self.summary.total {
             0
         } else {
@@ -763,7 +774,8 @@ impl FleetStatusReport {
 
     /// Render the human-readable table (AC 1: one command, every host side by
     /// side). Never empty output (AC 2): an empty roster prints an explicit
-    /// notice in addition to the local host's row.
+    /// notice in addition to the local host's row, AND (#5060) a qualified
+    /// roster-coverage summary line rather than an unqualified count.
     #[must_use]
     pub fn render_human(&self) -> String {
         let mut out = String::new();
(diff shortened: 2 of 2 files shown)


rjwalters:
Closing as a **duplicate** — superseded by #5090, which merged as `be18b02f` while this branch was building.

Two builders were dispatched onto #5060 concurrently. #5090 landed first and satisfies every acceptance criterion this PR did, with functionally identical logic:

- AC 1 — `exit_code()` early-returns non-zero on `summary.empty_roster` (identical guard placement).
- AC 2 — `render_human()`'s summary line branches for the empty-roster case (`EMPTY ROSTER — N of N known host(s) up (...); this is roster coverage (local host only), NOT a confirmed healthy fleet.` vs. this PR's `N of N known host(s) up — ROSTER IS EMPTY, ...`). Wording differs; behavior does not.
- AC 3 — `defaults/docs/daemon-reference.md` gained the empty-roster exit-code carve-out plus self-maintained-host registration guidance.
- AC 4 — `empty_roster_yields_local_only_row_and_explicit_notice` asserts `exit_code() != 0` and that the unqualified `1 host(s): 1 up` string is absent.
- AC 5 — the healthy-roster non-regression guard was added inside `exit_code_zero_only_when_every_host_up` (this PR added it as a separate test instead — a stylistic difference only).
- AC 6 — the tailnet/known-hosts cross-check is deferred in both.

Nothing here is worth salvaging on top of `main`: the only deltas are comment/message wording and test placement, not coverage or behavior. Branch deleted; no follow-up issue filed. Issue #5060 is already closed as completed by #5090.

PR #5097, opened 2026-08-03
fix(health): report a missing gh as one PATH fact, not N forge-query failures

### LINKED ISSUE #5061: health: a missing local gh is reported as "forge query FAILED" for every managed repo
## Context

Probing robb-pro over SSH on 2026-08-03, `loom-daemon health --json` reported:

```
queues:     unknown — 0+ ready across 5 repo(s) (safehouse ?, loom ?, klayout-tools ?, claude-monitor ?, anvil ?); forge query FAILED for: safehouse, loom, klayout-tools, claude-monitor, anvil
throughput: unknown — 0+ merged in 30m across 5 repo(s); forge query FAILED for: safehouse, loom, klayout-tools, claude-monitor, anvil
```

Every managed repo failing at once reads as a forge outage or a dead credential. It
was neither. The **probing process** simply had no `gh`:

```
$ ssh robb-pro 'bash -lc "which gh loom-daemon"'
bash: gh: command not found
```

Meanwhile the same daemon, over its own IPC, reported:

```
Forge credential: OK — keyring (rjwalters)
GitHub rate limit: OK (breaker closed; 1 trip(s) this daemon lifetime)
```

`queues` / `throughput` run their forge queries **in the process that invoked
`health`**, so over a non-login SSH shell (where `/opt/homebrew/bin` is absent from
`PATH`) they fail for reasons that have nothing to do with the host's forge access.

## Why this is worth fixing rather than documenting

The two verdicts are load-bearing during exactly the situation where you reach for
them — checking a host you cannot sit in front of. In this session the false signal
had to be disproven by cross-checking `Forge credential:` from a different subcommand
and then testing `which gh` by hand. An operator (or an automated watch) taking
`forge query FAILED for: <every repo>` at face value would conclude the wrong thing
about the wrong host.

Note this is a distinct failure from a genuine `gh` failure: **a missing client binary
is a property of the caller's environment, not of the forge.** The existing
"unknown != zero" contract is respected for the *count*, but the accompanying reason
string attributes the cause incorrectly.

- [ ] A missing or non-executable `gh` produces a distinct reason (e.g. "gh not found
      on PATH — cannot assess") rather than "forge query FAILED".
- [ ] That case is not attributed per-repo — it is one environment fact, not five
      independent query failures.
- [ ] Where the daemon already knows the answer over IPC (it reports
      `Forge credential: OK`), prefer or cross-reference that signal instead of
      re-deriving it in the caller.
- [ ] The `gh` resolution failure names the PATH problem, since this is the same class
      as #4875 (`~/.local/bin` absent from a non-login ssh PATH) and the same
      hazard `npm`/`node` hit during the #5016 repair.
- [ ] Documented, ideally in the health section's own help: which sections are
      daemon-authoritative (trustworthy over ssh) and which execute in the caller.

## Provenance

Hit during a manual cross-host check on 2026-08-03. The misattribution cost real time
before it was disproven, and it sits next to a genuine outage on the same host
(#5016), which is exactly when a false signal is most expensive.

rjwalters:
## Summary

`loom-daemon health --json`'s `queues`/`throughput` sections run their forge
queries (`gh`) in the **invoking** process, not the daemon's own process. When
probed over a non-login SSH shell where `gh` isn't on `PATH`, every managed
repo's query failed identically, rendering as:

```
forge query FAILED for: safehouse, loom, klayout-tools, claude-monitor, anvil
```

which reads exactly like a forge outage or a dead credential — even while the
daemon's own IPC round-trip correctly reports `Forge credential: OK`. The two
signals contradicted each other, and the false one is what an operator sees
first.

## Changes

- **`pipeline_snapshot::probe_gh_availability()`** (new): a single, bounded
  `<gh_bin> --version` spawn attempt, run **once** before any per-repo `gh`
  call. Distinguishes:
  - a bare name missing from `PATH` (names the PATH problem specifically —
    same failure class as #4875, cites the canonical `~/.local/bin` /
    `/opt/homebrew/bin` / `~/.cargo/bin` directories a non-login shell is
    commonly missing)
  - an explicit path that does not exist (no PATH hint — not a PATH problem)
  - a file found but not executable (`PermissionDenied`)
- **`health::assess_queues` / `assess_throughput`**: when the collector's
  probe found `gh` unavailable, both sections short-circuit to a single
  distinct reason (`Verdict::Unknown`, same exit-code contract as before —
  only the *reason string* changes) instead of the per-repo
  "forge query FAILED for: ..." list. The N-repo fan-out is skipped entirely
  (no value in spawning calls already known to fail identically).
- **Cross-reference to the daemon's own IPC verdict**: when the daemon
  answered IPC and reported `credential_preflight`, the section names it
  (`"...the daemon's own IPC reports its forge credential OK via keyring —
  this is a caller-side PATH problem, not a forge outage"`), so the two
  signals never silently contradict each other again.
- **`loom-daemon health --help`** and the module doc now document which
  sections are daemon-authoritative (`liveness`/`dispatch`/`tokens`/`roles`/
  `observability` — trustworthy the same way over SSH as locally) vs. which
  execute `gh` calls in the caller's own process (`queues`/`throughput`).
- Applied symmetrically to the dashboard's `GET /api/health` route
  (`serve.rs`), which shares the same `health::assess` collector.

## Test plan

- [x] `cargo test -p loom-daemon --lib` — 3249 passed, 0 failed (full `loom-daemon` crate)
- [x] `cargo clippy -p loom-daemon --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p loom-daemon -- --check` — clean
- [x] New unit tests: `pipeline_snapshot::tests::probe_gh_availability_*` (missing-on-PATH, explicit missing path, non-executable file, real executable) and `health::tests::a_missing_gh_is_one_fact_not_a_per_repo_failure_list` / `a_missing_gh_cross_references_a_healthy_daemon_credential` / `a_missing_gh_with_no_daemon_credential_signal_still_renders`
- [x] Manual repro: `PATH="/usr/bin:/bin" loom-daemon health` now reports a single
  PATH-attributed `queues`/`throughput` line instead of per-repo failures;
  with `gh` present the pre-existing "forge snapshot not collected" (daemon
  unreachable) message is unchanged

Closes #5061

### DIFF
--- a/loom-daemon/src/cli/health.rs
+++ b/loom-daemon/src/cli/health.rs
@@ -15,6 +15,32 @@
 //! determine" (which the collector renders as a non-green `UNKNOWN` section,
 //! exit `1`) rather than aborting — the *only* thing that can stop this
 //! command from printing a report is a panic.
+//!
+//! # Daemon-authoritative vs. caller-process sections (#5061)
+//!
+//! Not every section's verdict is trustworthy from the same vantage point,
+//! which matters most exactly when this command is run somewhere other than
+//! the machine that ran `loom-daemon health`'s writer (an SSH probe of a
+//! remote fleet host, a non-login shell whose `PATH` differs from an
+//! interactive login shell's):
+//!
+//! - **`liveness`, `dispatch`, `tokens`, `roles`, `observability`** are
+//!   **daemon-authoritative** — every fact comes from the daemon's own IPC
+//!   round-trip ([`DaemonStatusReport`]) or a local probe of *this host's*
+//!   process table/filesystem, never a forge call made by this CLI process.
+//!   A verdict here reflects the daemon's own state, not this caller's
+//!   environment, so it is trustworthy the same way over SSH as locally.
+//! - **`queues`, `throughput`** execute `gh` calls **in this CLI process**
+//!   ([`pipeline_snapshot::GhPipelineSource`]), scoped to whatever `gh`
+//!   resolves to and however it is authenticated *here* — which can differ
+//!   from the daemon's own (already-verified, see `credential_preflight` in
+//!   `status`/`--json`) forge credential. A missing/non-executable `gh` in
+//!   *this* process (the common case: a non-login SSH shell whose `PATH`
+//!   lacks `~/.local/bin` / `/opt/homebrew/bin`, #4875's failure class) is
+//!   reported as a single distinct fact rather than a per-repo forge-query
+//!   failure, and cross-references the daemon's own `credential_preflight`
+//!   verdict when it is available — see `health::assess_queues` /
+//!   `assess_throughput` / `gh_unavailable_section`.
 
 use anyhow::Result;
 use std::path::Path;
@@ -95,8 +121,23 @@ async fn collect(window: Duration) -> HealthReport {
     //    across the roots the daemon reported. Skipped entirely when the daemon
     //    is unreachable: without its root list there is nothing to query, and
     //    the sections honestly report "not collected".
-    let pipeline = match &status {
-        Some(report) => {
+    //
+    //    4a. Before fanning out to N repos, check ONCE whether this process
+    //    can even run `gh` at all (#5061). A missing/non-executable `gh` — the
+    //    common case being a non-login SSH shell whose PATH lacks
+    //    `~/.local/bin` / `/opt/homebrew/bin` (the same failure class as
+    //    #4875) — would otherwise fail identically for every managed repo,
+    //    rendering as "forge query FAILED for: <every repo>" and reading like
+    //    a forge outage rather than what it actually is: a fact about this
+    //    caller's own environment. `Some` here means the fan-out below is
+    //    skipped entirely (there is no value in spawning N `gh` calls already
+    //    known to fail the same way), and `queues`/`throughput` render the
+    //    single fact instead — see `health::assess_queues`/`assess_throughput`.
+    let gh_unavailable =
+        pipeline_snapshot::probe_gh_availability(Path::new(pipeline_snapshot::DEFAULT_GH_BIN))
+            .err();
+    let pipeline = match (&status, &gh_unavailable) {
+        (Some(report), None) => {
             let roots = report
                 .per_repo
                 .iter()
@@ -112,7 +153,7 @@ async fn collect(window: Duration) -> HealthReport {
             );
             Some(pipeline_snapshot::collect_pipeline_snapshots(source, roots).await)
         }
-        None => None,
+        _ => None,
     };
 
     // 5. The daemon log's newest `work_finder:` line (#4824) — a *corroborating*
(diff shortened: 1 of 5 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T19:11:34Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved.

**Correctness vs. issue #5061's acceptance criteria** — all five satisfied:
- A missing/non-executable `gh` now produces one distinct `Verdict::Unknown` reason (not `Verdict::Degraded` — correctly preserves the existing "unknown != healthy" exit-code contract).
- `queues`/`throughput` short-circuit to a single fact; the N-repo fan-out is skipped entirely (verified: `probe_gh_availability` runs once before any per-repo `gh` call).
- Cross-references the daemon's own IPC `credential_preflight` verdict by name when available (both the OK and DEGRADED cases render distinct clauses), and still renders sanely with no daemon signal (`status: None`).
- The PATH-lookup case cites `#4875` and lists the canonical missing directories (`~/.local/bin`, `~/.cargo/bin`, `/opt/homebrew/bin`, ...) via the shared `CANONICAL_PATH_DIRS` constant — an explicit non-existent path correctly gets no PATH hint (it isn't a PATH problem).
- `loom-daemon health --help` and the module doc now name which sections are daemon-authoritative vs. caller-process — verified live: `cargo run ... health --help` renders the new paragraph.

**Verification performed** (not just re-reading the diff):
- `cargo test -p loom-daemon --lib` — 3249 passed, 0 failed (matches the PR's claimed count exactly).
- `cargo clippy -p loom-daemon --all-targets -- -D warnings` — clean.
- Live manual repro against the actual built binary: `PATH="/usr/bin:/bin" loom-daemon health` reproduces the exact scenario from the issue — a single `queues`/`throughput` UNKNOWN line naming PATH and cross-referencing a real `credential_preflight: OK` from this host's live daemon, with zero occurrences of the old `forge query FAILED for:` phrasing anywhere in `--json` output. Also confirmed the pre-existing "forge snapshot not collected" (daemon-unreachable) path is unchanged when `gh` is absent-but-daemon-IPC also fails.
- CI: all required checks pass (`Rust Linting & Build Check`, `Rust Unit Tests`, `Rust OTLP Feature Build & Test`, plus the doc/format checks) — verified via the REST check-runs API (GraphQL quota was exhausted on this account at review time). `mergeable_state: clean` via REST as well.

**Code quality notes:**
- `GhUnavailable`/`probe_gh_availability` are well-isolated, well-tested (missing-on-PATH, explicit-missing-path, non-executable — all three exercised), and the CLI/serve-route symmetry is real (both call sites use the same `DEFAULT_GH_BIN` constant, so they can't drift into probing a different binary than the one the per-repo fan-out actually uses).
- Minor, non-blocking suggestion for a follow-up: `probe_gh_availability`'s `Command::new(gh_bin).arg("--version").output()` call in `serve.rs::handle_health` runs inline on the async task rather than via `tokio::task::spawn_blocking`, unlike every other blocking subprocess call in that file (`collect_pipeline_snapshots`'s per-repo fetch, `derive_forge_url`). The blocking window is small and bounded (a single `--version` spawn, or an near-instant `NotFound`), and `cli/health.rs`'s collector already tolerates a few similar inline `Command::new` probes (`pgrep`, `launchctl`) elsewhere, so this isn't blocking approval — just worth tightening for consistency with the rest of `serve.rs`'s own convention if anyone touches this code again.

LGTM — ships the fix exactly as scoped, with real test coverage and a live repro that matches the original incident report.

Issue #5101, opened 2026-08-03
dashboard: SPA's fleet summary counts sweep-only 'unknown' hosts in its 'N hosts' headline, inflating the count in production


rjwalters:
## Problem

Follow-up from #5078's investigation. The observed production incident in
#5078 ("4 hosts" vs 3 real hosts, a `studio` card showing "Never reported" /
"No data" with 6 active sweeps) was almost certainly caused by
`dashboard/web/src/fleet.ts`'s `buildFleetView`, not `dashboard/src/publicPage.ts`
(the fallback page #5078 actually fixed).

`dashboard.2amlogic.com` in production serves the built SPA
(`web/dist/index.html` via the `ASSETS` binding — see `handleRoot` in
`dashboard/src/index.ts`); `publicPage.ts` only renders when no UI build is
uploaded. #5078's curator verification only re-read `dashboard/src/*` and
concluded the "union of hosts and activeSweeps" mechanism described in the
original issue text "doesn't match the current code" — true for
`publicPage.ts`, but `dashboard/web/src/fleet.ts` does exactly that union,
deliberately and by design:

```ts
// dashboard/web/src/fleet.ts, buildFleetView
const hostIds = new Set<string>([...Object.keys(snapshot.hosts), ...sweepsByHost.keys()]);
```

This is intentional (see the module's own doc comment: "Keying off `hosts`
alone would silently hide a busy host") and is pinned by
`web/test/fleet.test.ts`'s `"includes a host known only from activeSweeps"`
test — a host known only from `activeSweeps` gets `status: "unknown"` and
still renders as a card (`web/src/views/fleetOverview.ts`'s `hostCard`,
labeled "No data").

**That design itself is reasonable** — it lets an operator see a host that's
genuinely busy but hasn't sent its first `host.health` yet, or is caught
mid-revoke with in-flight sweeps (exactly #5078's "mid-run-revoke anomaly
must remain discoverable" AC). The actual gap is narrower: the overview's
headline summary line does not distinguish an "unknown"-status host from one
with real health data when it reports the top-level count:

```ts
// dashboard/web/src/views/fleetOverview.ts, fleetOverviewView
el("span", {}, `${view.hosts.length} host${view.hosts.length === 1 ? "" : "s"}`),
el("span", {}, `${view.totalSweeps} active sweep${view.totalSweeps === 1 ? "" : "s"}`),
```

`view.hosts.length` includes every "unknown"-status host, and
[...]

PR #5102, opened 2026-08-03
fix(dashboard): exclude health-less hosts from fleet counts, filter D1-revoked hosts

### LINKED ISSUE #5078: dashboard: a host known only from orphaned sweep entries renders as a phantom fleet member, inflating host and active-sweep counts
## Problem

A host with **no `health:` entry at all** — and a **revoked** `hosts` row — still renders as a fleet member with active work.

Observed 2026-08-03 on dashboard.2amlogic.com immediately after the robb-studio rename: a `studio` card showing **"Never reported" / "No data"**, every metric an em-dash, but **6 active sweeps** listed under it, and the page header reading **"4 hosts · 20 active sweeps"**.

## Cause

`publicPage.ts`'s `renderFleetOverview` derives `hostIds` from the **union** of `snapshot.hosts` and `snapshot.activeSweeps`. `renderFreshnessSummary` (`:207`) only classifies hosts that have a `health` entry — its own comment notes that hosts "known only from `activeSweeps`" are counted in `hostIds.length` but contribute no freshness bucket. So orphaned `sweep:` entries are sufficient to synthesise a card and inflate both counts.

**Curator verification (2026-08-03)**: re-read against `origin/main` (`dashboard/src/publicPage.ts`, `dashboard/src/fleetState.ts`) — the "union" mechanism above doesn't match the current code, and the two inflated counts have two *different*, more specific causes:

1. **Active-sweep count/table (confirmed, matches the issue as-is)**: `renderFleetOverview` (`publicPage.ts:298-305`) computes `hostIds = Object.keys(snapshot.hosts).sort()` — this reads `snapshot.hosts` alone, there is no union with `snapshot.activeSweeps` anywhere in the render path. But `sweepRows`/`snapshot.activeSweeps.length` are independent of `hostIds` entirely, and `removeHost` (`fleetState.ts:323-330`) deliberately never touches `sweep:` entries on revoke. So a revoked host's sweeps keep appearing in the Active Sweeps table/count regardless of whether that host has any `hosts[]` record — this part of the bug is real and exactly as described.
2. **Host count/card ("phantom fleet member", confirmed via a different mechanism)**: `snapshot.hosts` is populated *only* from `health:`/`tokens:`-prefixed DO storage entries (`classifyAndPruneHosts`, `fleetState.ts:144-171`) — a hostId known *only* from a `sweep:` entry is never added to `hosts`, so on today's code it is **not** counted in `hostIds.length` and does not render a row (`renderHostHealthRow` returns `""` when there's no `health` entry — `publicPage.ts:260-270`). The more likely real source of a lingering, no-data host card is `handleRevokeHost`'s best-effort Durable Object cleanup (`dashboard/src/index.ts:355-382`): the D1 `revoked_at` write always commits, but the follow-up fetch to the FleetState DO's `/remove-host` (which deletes that host's `health:`/`tokens:` entries) is wrapped in try/catch specifically so a DO hiccup can't fail the revoke — meaning a failed cleanup fetch leaves stale `health:`/`tokens:` entries in the DO indefinitely, still counted in `hostIds.length` and still rendering a stale/offline row, with no dependency on `activeSweeps` at all.

**Implication for the Builder**: reproduce against a live/staged snapshot before implementing — confirm which mechanism (1, 2, or both) actually produced the "4 hosts" figure in the observed instance. The AC below is still the right target state, but "exclude activeSweeps-only phantom hosts" (AC bullet 1) may already hold for mechanism 2's scenario on current code; the real gap for the host-count side may instead be making the best-effort DO cleanup retryable/self-healing (or treating a `revoked_at`-set D1 row as authoritative over a stale DO record) rather than changing the union logic, since no such union exists today.

## Why it matters

Per #4998, sweep count is currently the dashboard's only load-density signal. Inflating it with phantoms — 20 shown vs 14 real in this instance — misleads exactly the judgement that number exists to support. The phantom host also inflates fleet size (4 vs 3).

## Suggested direction

Surface unattributed sweeps as an anomaly rather than as a host:

- A host with no `health:` entry should not render as a fleet card nor count toward the host total.
- Its sweeps should either be excluded from the active-sweep count or shown in a distinct "unattributed / orphaned" grouping — the #4955 anomaly signal is worth keeping visible, just not disguised as a healthy host.
- Do **not** fix this by making revoke clear sweep entries; that is a deliberate design decision (see the correction on #5071).
- Given the curator verification above, also consider: making `handleRevokeHost`'s DO cleanup retry-safe (or idempotent/re-triggerable) instead of silently swallowing a failed `/remove-host` fetch, so a revoked D1 row can't leave a stale DO host record behind indefinitely.

## Acceptance criteria

- [ ] A host known only from `activeSweeps` does not render as a fleet card.
- [ ] The header host count reflects hosts with health data.
- [ ] The active-sweep count either excludes sweeps with no live host, or those sweeps are visibly grouped as unattributed.
- [ ] The mid-run-revoke anomaly (`fleetState.ts:323-326`) remains discoverable.
- [ ] A revoked host whose best-effort DO cleanup (`index.ts:355-382`) failed does not keep rendering as a live fleet card off stale `health:`/`tokens:` data.
- [ ] Test covers a snapshot containing a `sweep:` entry whose `hostId` has no `health:` entry.
- [ ] Test covers a snapshot where a host's D1 row is revoked but its DO `health:`/`tokens:` entries were never cleaned up (mechanism 2 above).

## Affected Files

- `dashboard/src/publicPage.ts` - `renderFleetOverview` / `renderFreshnessSummary` / `renderHostHealthRow` (host-card and count rendering)
- `dashboard/src/fleetState.ts` - `classifyAndPruneHosts`, `removeHost`, `buildSnapshot` (host/sweep entry lifecycle)
- `dashboard/src/index.ts` - `handleRevokeHost` (best-effort DO cleanup on revoke, lines 355-382)
- `dashboard/src/publicPage.test.ts` / `dashboard/src/fleetState.test.ts` (or equivalent existing test files) - add regression tests per Test Plan

## Test Plan

- [ ] Manual verification: construct a `RedactedFleetSnapshot` fixture with (a) a `sweep:`-only phantom host and (b) a stale `health:`/`tokens:` entry for a D1-revoked host, render via `renderFleetOverview`, and confirm neither produces a fleet card or inflates the host count.
- [ ] Automated tests: add cases to the existing `publicPage`/`fleetState` test files for both scenarios in the Acceptance Criteria; verify the active-sweep table either excludes or clearly separates unattributed sweeps.
- [ ] Edge cases: confirm the mid-run-revoke anomaly (a host revoked while it has genuinely in-flight sweeps) still surfaces as a visible anomaly rather than being silently swept into the same "unattributed" bucket with no signal.

## Provenance

Found during the robb-studio host rename, 2026-08-03. Split out of #5071 after that issue's original premise turned out to be wrong (revoke's behaviour is intentional). Related: #4998, #4955, #4957, #5068, #5084 (a fresh instance of the "lost `sweep.completed`" pathway #4955 already accounts for, via a daemon-restart mechanism).

rjwalters:
## Summary

A host known only from `activeSweeps` (no `health:` entry) could inflate the
public dashboard page's header host count, and a D1-revoked host whose
best-effort Durable Object cleanup silently failed could keep rendering as a
live fleet card off stale `health:`/`tokens:` data indefinitely. This fixes
both mechanisms confirmed by the curator's re-verification against current
`dashboard/src/*` code (see issue #5078 for the two-mechanism breakdown).

## Changes

- `dashboard/src/fleetState.ts`: new `filterRevokedHosts(snapshot,
  revokedHostIds)` — a pure function that drops any `hosts` entry for a
  hostId D1 has recorded as revoked, without touching `activeSweeps` (so a
  genuine mid-run-revoke anomaly stays visible rather than being hidden).
- `dashboard/src/index.ts`: new `fetchLiveFleetSnapshot(env)` — fetches the
  Durable Object's live snapshot, cross-references D1's `revoked_at` for the
  hosts present in it, and filters via `filterRevokedHosts`. Wired into both
  `handleFleetStateQuery` (`/api/fleet-state`, `/public/fleet-state`) and
  `handleDashboardPage` (the `/` no-UI-build fallback). `GET
  /admin/fleet-state` deliberately stays unfiltered — it is raw DO
  introspection, and filtering there would hide the exact staleness it
  exists to let an operator diagnose.
- `dashboard/src/publicPage.ts`: `renderFleetOverview`'s `hostIds` is now the
  set of `snapshot.hosts` entries that actually carry a `health` record (not
  every key `snapshot.hosts` happens to have), so the "Hosts (N)" header
  count and the rendered card list are always in exact agreement.
  `activeSweeps` is split into "attributed" (hostId has current health data)
  and "unattributed" (rendered in a new "Unattributed sweeps" section)
  instead of commingling with, or silently dropping from, the primary
  "Active sweeps (N)" count.

| A host known only from `activeSweeps` does not render as a fleet card | ✅ | `renderFleetOverview`'s `hostIds` now requires a `health` entry; `publicPage.test.ts` — "a snapshot containing only a sweep: entry ... renders no fleet card for it" |
| The header host count reflects hosts with health data | ✅ | Same `hostIds` filter; `publicPage.test.ts` — "Hosts (0)" cases for sweep-only and tokens-only hosts |
| The active-sweep count either excludes sweeps with no live host, or those sweeps are visibly grouped as unattributed | ✅ | `activeSweeps` split into attributed/unattributed; unattributed rendered in its own "Unattributed sweeps" table, excluded from "Active sweeps (N)" |
| The mid-run-revoke anomaly (`fleetState.ts:323-326`) remains discoverable | ✅ | `filterRevokedHosts` never touches `activeSweeps`; a revoked host's genuinely in-flight sweeps surface in the "Unattributed sweeps" section — `fleetState.test.ts` and `publicPage.test.ts`'s mid-run-revoke test |
| A revoked host whose best-effort DO cleanup failed does not keep rendering as a live fleet card off stale `health:`/`tokens:` data | ✅ | `fetchLiveFleetSnapshot` cross-references D1's `revoked_at`; `index.test.ts` end-to-end test seeds a host, ingests health, revokes in D1 directly (bypassing the DO cleanup route), and confirms it is absent from `/public/fleet-state` and `/api/fleet-state` |
| Test covers a snapshot containing a `sweep:` entry whose `hostId` has no `health:` entry | ✅ | `publicPage.test.ts` — "unattributed sweeps" describe block |
| Test covers a snapshot where a host's D1 row is revoked but its DO `health:`/`tokens:` entries were never cleaned up | ✅ | `index.test.ts` — "D1-revoked host with un-cleaned-up DO state" describe block |

## Test Plan

- `npm run check` (typecheck + vitest) in `dashboard/`: 242/242 tests pass
  (228 pre-existing + 14 new).
- `npm run check` in `dashboard/web/`: 463/463 tests pass, unmodified (this
  PR does not touch `web/src/*` — see the note below).
- Manually traced the D1-revoke integration test's setup (seed host → ingest
  `host.health` → flip D1 `revoked_at` directly, bypassing
  `handleRevokeHost`'s DO cleanup call entirely) to confirm it reproduces
  exactly the "best-effort cleanup fetch failed" scenario the AC describes.

## Note: production likely runs a third, related mechanism not in scope here

`dashboard.2amlogic.com` serves the built SPA (`web/dist/index.html`) in
production; `publicPage.ts` (fixed here) is only the no-UI-build fallback.
`dashboard/web/src/fleet.ts`'s `buildFleetView` deliberately unions
`snapshot.hosts` and `activeSweeps`' hostIds (its own doc comment: "Keying
off `hosts` alone would silently hide a busy host"), pinned by an existing
test — this is very likely the actual cause of the "4 hosts" figure the
original incident observed. That design is reasonable and intentional; the
narrower gap is that the SPA's headline "N hosts" summary doesn't distinguish
an "unknown"-status (health-less) host from one with real data. Filed #5101
to track that follow-up without touching `web/src/*`'s tested join behavior
in this PR.

Closes #5078

### DIFF
--- a/dashboard/src/fleetState.ts
+++ b/dashboard/src/fleetState.ts
@@ -188,6 +188,41 @@ export interface FleetSnapshot {
   activeSweeps: ActiveSweepState[];
 }
 
+/**
+ * Drop any `hosts` entry belonging to a hostId D1 has recorded as revoked
+ * (Issue #5078, mechanism 2 — "a host known only from a stale DO record").
+ *
+ * `handleRevokeHost` (`src/index.ts`) writes D1's `revoked_at` unconditionally,
+ * then best-effort clears this DO's `health:`/`tokens:` entries for that host
+ * — that cleanup fetch is wrapped in try/catch specifically so a transient DO
+ * failure can't fail the revoke itself (see `FleetState.removeHost`'s doc
+ * comment), which means a failed cleanup can leave those entries in the DO
+ * indefinitely. Rather than rely on that best-effort cleanup having
+ * succeeded, `src/index.ts` treats D1's `revoked_at` as authoritative at
+ * *read* time too: a snapshot never renders a revoked host as live, whether
+ * or not its DO-side cleanup actually ran.
+ *
+ * Deliberately does **not** touch `activeSweeps` — a sweep still reporting
+ * against a host revoked mid-run is a real anomaly (`removeHost`'s own doc
+ * comment: "an in-flight sweep on a host being revoked mid-run is a real
+ * anomaly worth surfacing"), not something this filter should hide; it stays
+ * visible (as an unattributed sweep, once the host's `hosts` entry above is
+ * gone — see `publicPage.ts`'s `renderFleetOverview`).
+ */
+export function filterRevokedHosts(
+  snapshot: FleetSnapshot,
+  revokedHostIds: ReadonlySet<string>,
+): FleetSnapshot {
+  if (revokedHostIds.size === 0) return snapshot;
+  const hosts: FleetSnapshot["hosts"] = {};
+  for (const [hostId, entry] of Object.entries(snapshot.hosts)) {
+    if (!revokedHostIds.has(hostId)) {
+      hosts[hostId] = entry;
+    }
+  }
+  return { hosts, activeSweeps: snapshot.activeSweeps };
+}
+
 /** Body accepted by the internal `POST /update` route — one record's worth
  * of live-state effect, already authenticated/validated by the Worker
  * before it reaches the Durable Object. */
@@ -323,7 +358,11 @@ export class FleetState implements DurableObject {
    * concept at this layer. Does **not** touch that host's `sweep:<sweepId>`
    * entries: an in-flight sweep on a host being revoked mid-run is a real
    * anomaly worth surfacing (via its own staleness), not something this
-   * best-effort cleanup should paper over.
+   * best-effort cleanup should paper over. Because the caller's fetch to
+   * this route is itself best-effort (issue #5078), a failure here can leave
+   * these entries behind indefinitely — [`filterRevokedHosts`] is the
+   * read-time backstop for exactly that case, treating D1's `revoked_at` as
+   * authoritative over whatever this method did or did not manage to delete.
    */
   private async removeHost(hostId: string): Promise<void> {
     await this.state.storage.delete([`health:${hostId}`, `tokens:${hostId}`]);
--- a/dashboard/src/index.ts
+++ b/dashboard/src/index.ts
@@ -104,7 +104,7 @@
 
 import { extractBearerToken, authenticateHost, hashIngestKey } from "./auth";
 import { validateAccessJwt } from "./accessAuth";
-import { FleetState, type FleetSnapshot } from "./fleetState";
+import { FleetState, filterRevokedHosts, type FleetSnapshot } from "./fleetState";
 import { renderPublicPage } from "./publicPage";
 import {
   createLiveTailStream,
@@ -162,6 +162,37 @@ function jsonError(status: number, message: string): Response {
 
 function fleetStateStub(env: Env): DurableObjectStub {
   return env.FLEET_STATE.get(env.FLEET_STATE.idFromName(FLEET_STATE_ID_NAME));
+}
+
+/** D1 host_ids, among `hostIds`, whose `hosts` row is revoked — bounded to
+ * the snapshot's own host set rather than scanning the whole `hosts` table,
+ * since that is all `filterRevokedHosts` needs. */
+async function fetchRevokedHostIds(env: Env, hostIds: readonly string[]): Promise<Set<string>> {
+  if (hostIds.length === 0) return new Set();
+  const placeholders = hostIds.map(() => "?").join(",");
+  const { results } = await env.DB.prepare(
+    `SELECT host_id FROM hosts WHERE revoked_at IS NOT NULL AND host_id IN (${placeholders})`,
+  )
+    .bind(...hostIds)
+    .all<{ host_id: string }>();
+  return new Set(results.map((row) => row.host_id));
+}
+
+/**
+ * The Durable Object's live snapshot, with any host D1 has recorded as
+ * revoked filtered out (issue #5078, mechanism 2) — the single point every
+ * dashboard-facing read (`/api/fleet-state`, `/public/fleet-state`, the
+ * server-rendered `/` fallback) goes through, so a `handleRevokeHost` cleanup
+ * fetch that silently failed can never leave a stale `health:`/`tokens:`
+ * entry rendering as a live host. `GET /admin/fleet-state` deliberately does
+ * **not** use this helper — it is introspection of the DO's own raw state
+ * (see its route doc), and filtering there would hide the very staleness it
+ * exists to let an operator diagnose. */
+async function fetchLiveFleetSnapshot(env: Env): Promise<FleetSnapshot> {
+  const response = await fleetStateStub(env).fetch("https://fleet-state/snapshot");
+  const snapshot = (await response.json()) as FleetSnapshot;
+  const revokedHostIds = await fetchRevokedHostIds(env, Object.keys(snapshot.hosts));
+  return filterRevokedHosts(snapshot, revokedHostIds);
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


rjwalters:
LGTM — approving.

**Verified:**
- `npm run check` in `dashboard/` (fresh `npm ci`): typecheck clean, 242/242 tests pass (228 pre-existing + 14 new).
- CI: all checks pass (Dashboard Tests, CLAUDE.md budget, docs parity, AGENTS.md sync).
- `mergeStateStatus`: CLEAN.

**Correctness against the issue's AC:**
- `filterRevokedHosts()` (`fleetState.ts`) correctly drops `hosts` entries for D1-revoked hostIds while leaving `activeSweeps` untouched — confirmed the mid-run-revoke anomaly stays visible via the new "Unattributed sweeps" section rather than being silently swept away.
- `fetchLiveFleetSnapshot()` (`index.ts`) is wired into both `/api/fleet-state` / `/public/fleet-state` and the `/` fallback page; `/admin/fleet-state` is correctly left on the raw DO snapshot for operator introspection, exactly as the PR description states.
- `renderFleetOverview`'s `hostIds` now requires an actual `health` entry, so the "Hosts (N)" header and the rendered card list are always in agreement — including the `tokens`-only-host edge case, which is covered by its own test.
- `fetchRevokedHostIds`'s D1 query is parameterized (no injection risk) and bounded to the snapshot's own hostIds, not a full-table scan.
- Noted the `response.status` → hardcoded `200` change in `handleFleetStateQuery`: the DO's `/snapshot` handler never returns anything but the default 200, so this is a behavior no-op, not a regression.

**Test coverage** is thorough and specifically targets both mechanisms from the issue (sweep-only phantom host, D1-revoked-but-uncleaned-DO-state host), plus the mid-run-revoke discoverability requirement and the tokens-only-host edge case.

Follow-up #5101 (SPA's separate `buildFleetView` union logic) is correctly scoped out and filed as `loom:triage`, not self-approved — agreed that's out of scope for this PR.

PR #5099, opened 2026-08-03
docs(observability): document the host-rename runbook

rjwalters:
## Summary

Adds a new `## 7. Renaming a host` section to `.loom/docs/observability.md`, between `## 6. The 2AM reference instance` and `## Map of every detail doc`, documenting the runbook for renaming a Loom telemetry host identity (`host_id`).

## Changes

- New numbered 6-step procedure: provision new identity (`POST /admin/hosts`) → install the key on the host (`~/.loom/observability/ingest.key`, mode 600) → flip `$LOOM_HOST_ID` in the launchd plist and cycle the job (`launchctl bootout` + `bootstrap`) → verify via `loom-daemon health` before touching any data → optionally back up and relabel historical D1 rows (only after verification is green, and only against a data-verified boundary via `SELECT max(ts)`/`min(ts)`, not an assumed timestamp) → revoke the old key (`POST /admin/hosts/:hostId/revoke`).
- Documents that `bootout` is asynchronous and a `bootstrap` issued into that window can fail transiently with `Bootstrap failed: 5: Input/output error` — verify `state = running` with a new pid via `launchctl print` and retry bootstrap if absent, rather than treating the transient failure as terminal.
- Notes that doing nothing and letting 90-day retention age out the old rows is a legitimate default when no historical trend continuity is needed.
- Adds a "Known gaps" subsection cross-referencing (verified OPEN at build time):
  - #5084 — a daemon restart during the rename can silently drop in-flight sweep outcome telemetry.
  - #5083 — the health check's silence is not positive confirmation telemetry is flowing.
  - #5078 — revoking a key while sweep entries referencing the old `host_id` are still active/orphaned can leave a phantom fleet member on the dashboard.
  - #5082 — re-provisioning a previously-revoked `host_id` currently 409s with no override.

Documentation-only change — no production access, secrets, or wrangler/D1 commands were exercised while writing this.

## Acceptance Criteria Verification

- [x] New `## 7. Renaming a host` section inserted at the exact specified location (after `## 6. The 2AM reference instance`, before `## Map of every detail doc`).
- [x] All 6 runbook steps present, in order, each unambiguous with no missing prerequisite (verified by re-reading the section as a rename walkthrough).
- [x] `bootout`/`bootstrap` transient-failure behavior and the pid-based verification recipe documented.
- [x] "Known gaps" subsection present, each item linked to its tracking issue, no restated fix detail.
- [x] 90-day-retention-as-default note included.
[...]

PR #5103, opened 2026-08-03
fix(dashboard): re-provision a revoked host_id via an atomic upsert

### LINKED ISSUE #5082: dashboard: POST /admin/hosts 409s on a revoked host_id, so re-provisioning a retired identity requires hand-editing production D1
## Problem

`POST /admin/hosts` refuses any `host_id` that has ever existed, **including one whose
key is revoked**, and there is no un-revoke, force, or rename path. So re-provisioning a
retired identity requires hand-editing production D1.

`dashboard/src/index.ts:336-339`:

```ts
const existing = await env.DB.prepare("SELECT host_id FROM hosts WHERE host_id = ?").bind(hostId).first();
if (existing) {
  return jsonError(409, `host_id "${hostId}" already exists`);
}
```

The lookup does not consider `revoked_at`. A revoked row is a dead credential — its
`key_hash` can never authenticate again — but it permanently reserves the name.

## Observed 2026-08-03

Renaming robb-studio's telemetry identity from `studio` to `robb-studio` required
provisioning `robb-studio`. That name had been minted on 2026-07-31 and **revoked** the
same day (21:03:19) during an earlier mis-bound-key cleanup, so:

```
POST /admin/hosts {"host_id":"robb-studio"}   →  409 host_id "robb-studio" already exists
```

The only way through was a manual production mutation:

```sql
DELETE FROM hosts WHERE host_id='robb-studio' AND revoked_at IS NOT NULL;
```

then re-POST (which returned 201). Hand-editing the auth table of a live ingest endpoint
is not a procedure that should be required for a routine host rename — and it is
precisely the kind of step that gets fat-fingered without the `AND revoked_at IS NOT NULL`
guard.

## Why it will recur

- `robb-air` is planned, standardising local compute on `robb-studio` / `robb-pro` /
  `robb-air`.
- Any key rotation that goes through revoke-then-reprovision hits this.
- Cloud workers get EC2-default hostnames (`ip-172-31-74-176`), so identity churn on
  that tier is likely (#5062).

## Suggested direction

Any one of these closes it; the first is smallest:

- Scope the existence check to **live** rows (`WHERE host_id = ? AND revoked_at IS NULL`)
  so a revoked name can be re-minted. Keeping the old row would violate the
  `host_id` primary key, so this needs either a delete-then-insert or an
  `INSERT … ON CONFLICT(host_id) DO UPDATE` that sets the new `key_hash` and clears
  `revoked_at`. Prefer the upsert — it is atomic and leaves no window where the name is
  unprovisioned.
- Or add an explicit `POST /admin/hosts/:hostId/rotate` for the re-key case.
- Or add a first-class rename operation, which is the actual operator intent here and
  does not exist at all today (#5068).

Whatever lands should preserve the audit story: a revoked key must never be
resurrectable, only replaced.

## Acceptance criteria

- [ ] A previously revoked `host_id` can be re-provisioned through the admin API, with no
      manual D1 mutation.
- [ ] The old `key_hash` is replaced, never reactivated — the revoked credential stays
      dead.
- [ ] No window exists where the host row is absent (atomic upsert, not delete + insert).
- [ ] The 409 remains for a host_id that is currently **live**, so accidental re-minting
      of an active host still fails.
- [ ] Test covers: provision → revoke → re-provision succeeds, and the first key no
      longer authenticates.

## Provenance

Split out of #5071, whose original premise (that revoke should clear `sweep:` entries)
was wrong — see the correction on that issue. This provisioning gap was the one valid
finding in it, and it blocked the rename in #5068 until worked around by hand.

rjwalters:
## Summary

`POST /admin/hosts` refused any `host_id` that had ever existed — including one whose key was **revoked** — because the existence check was a plain primary-key lookup with no `revoked_at` filter. Reclaiming a retired name (the 2026-08-03 `robb-studio` rename) therefore required a hand-written `DELETE` against production D1.

The pre-check + `INSERT` is replaced by a single atomic upsert:

```sql
INSERT INTO hosts (host_id, key_hash, created_at, revoked_at) VALUES (?, ?, ?, NULL)
ON CONFLICT(host_id) DO UPDATE SET
  key_hash = excluded.key_hash,
  created_at = excluded.created_at,
  revoked_at = NULL
WHERE hosts.revoked_at IS NOT NULL
```

A **live** row does not match the `DO UPDATE` guard, so `meta.changes` is 0 and the route still returns 409. A **revoked** row is rewritten in place with the new `key_hash` and a cleared `revoked_at` — the dead credential is replaced, never reactivated. Because it is an upsert rather than DELETE + INSERT, there is no window in which the host row is absent, and the TOCTOU gap between the old check and the old write is gone as well.

## Changes

- `dashboard/src/index.ts` — `handleCreateHost`: drop the unfiltered `SELECT` existence check; issue the guarded upsert above and return 409 when it reports zero changes. `created_at` is restamped because the row now describes a freshly minted credential.
- `dashboard/test/ingest.test.ts` — three regression tests in the existing `POST /admin/hosts — host provisioning` block.
- `dashboard/docs/deploy-runbook.md` — "Rotating an ingest key" option B is now revoke-then-reprovision (no `wrangler d1 execute ... DELETE`), and option A's framing corrected to "currently live".
- `.loom/docs/observability.md` — the host-rename "Known gaps" entry for #5082 now describes the shipped behavior. (Edited in the installed copy because that is where the content lives — the rename runbook was added there only, by #5099; there is no `defaults/docs/` counterpart to edit.)

- [x] **A previously revoked `host_id` can be re-provisioned through the admin API, with no manual D1 mutation** — `re-provisions a revoked host_id, and the pre-revoke key stays dead` asserts `201` where `main` returns `409`.
- [x] **The old `key_hash` is replaced, never reactivated** — same test: `POST /ingest` with the pre-revoke key returns `401` after re-provisioning, while the newly returned key returns `200` and its record lands under the same `host_id`.
- [x] **No window exists where the host row is absent (atomic upsert, not delete + insert)** — a single statement; `re-provisions a revoked host_id in place, without deleting the row first` asserts exactly one row survives, with `revoked_at IS NULL` and a `key_hash` different from the pre-revoke hash.
- [x] **The 409 remains for a currently live `host_id`** — the pre-existing `rejects creating a duplicate host_id` test (live `host-abc` fixture) still passes unchanged; the guard is enforced by `WHERE hosts.revoked_at IS NOT NULL` in SQL rather than by a separate read, so it also holds under concurrent provisions.
- [x] **Test covers provision → revoke → re-provision, and the first key no longer authenticates** — as above, plus `honors an explicit key when re-provisioning a revoked host_id` for the operator-supplied-key path.

## Test Plan

- [x] `npm run check` in `dashboard/` (typecheck + vitest): **231 passed / 10 files**, exit 0.
- [x] Negative control — reverted `src/index.ts` to `main` and re-ran `test/ingest.test.ts`: the 3 new tests fail (`409` instead of `201`), the other 22 pass. The tests genuinely gate the fix.
- [x] `tsc --noEmit` clean.
- [x] Manual verification: N/A — `dashboard/test/` runs against a real D1 binding via `apply-migrations.ts`, not a mock, so the upsert is exercised against actual SQLite semantics (including the skipped-`DO UPDATE` `changes = 0` behavior the 409 path depends on).

Closes #5082

### DIFF
--- a/.loom/docs/observability.md
+++ b/.loom/docs/observability.md
@@ -301,8 +301,11 @@ here:
 - Revoking the old key (step 6) while sweep entries still reference the old
   `host_id` and are active or orphaned can leave a phantom fleet member on
   the dashboard until those entries clear — #5078.
-- Re-provisioning a previously-revoked `host_id` (e.g. reusing the old name
-  later) currently 409s with no override — #5082.
+Reusing a previously-revoked `host_id` (e.g. reclaiming the old name later)
+used to 409 with no override; `POST /admin/hosts` now re-provisions a revoked
+host in place — minting a new key and clearing the revocation atomically, never
+reviving the dead one — and only 409s for a host that is currently live
+(#5082).
 
 ## Map of every detail doc
 
--- a/dashboard/docs/deploy-runbook.md
+++ b/dashboard/docs/deploy-runbook.md
@@ -370,10 +370,12 @@ below.
 
 ### Rotating an ingest key
 
-`POST /admin/hosts` refuses a `host_id` that already exists (`409`), so
-rotation is a two-move operation. Pick whichever fits:
+`POST /admin/hosts` refuses a `host_id` that is currently **live** (`409`), so
+rotation is a two-move operation. A **revoked** `host_id` is re-provisionable
+(issue #5082) — the re-provision replaces the dead key rather than reviving it,
+so no raw SQL is involved either way. Pick whichever fits:
 
-**A. Rolling rotation (no raw SQL, brief dual-identity window)** — provision a
+**A. Rolling rotation (brief dual-identity window, no `401`s)** — provision a
 successor identity, cut the host over, then revoke the old one:
 
 ```bash
@@ -385,17 +387,16 @@ curl -sS -X POST "$BASE/admin/hosts/my-laptop/revoke" -H "authorization: Bearer 
 
 Historical rows keep the old `host_id`; new rows use the new one.
 
-**B. In-place rotation (same `host_id`, needs one D1 statement)** — delete the
-host row, then re-provision the same id with a fresh key:
+**B. In-place rotation (same `host_id`)** — revoke the host, then re-provision
+the same id, which mints a fresh key and clears the revocation atomically:
 
 ```bash
-npx wrangler d1 execute loom-observability --remote \
-  --command "DELETE FROM hosts WHERE host_id = 'my-laptop'"
+curl -sS -X POST "$BASE/admin/hosts/my-laptop/revoke" -H "authorization: Bearer $ADMIN"
 curl -sS -X POST "$BASE/admin/hosts" -H "authorization: Bearer $ADMIN" \
   -H 'content-type: application/json' -d '{"host_id":"my-laptop"}'
 ```
 
-Between the delete and the daemon picking up the new key, that host's pushes
+Between the revoke and the daemon picking up the new key, that host's pushes
 get `401` — they are **not lost**: the daemon's durable queue retries with
 backoff and drains once the new key is in place (up to `queueCapacity`).
 
--- a/dashboard/src/index.ts
+++ b/dashboard/src/index.ts
@@ -333,16 +333,37 @@ async function handleCreateHost(request: Request, env: Env): Promise<Response> {
     return jsonError(400, "host_id is required");
   }
 
-  const existing = await env.DB.prepare("SELECT host_id FROM hosts WHERE host_id = ?").bind(hostId).first();
-  if (existing) {
-    return jsonError(409, `host_id "${hostId}" already exists`);
-  }
-
   const key = typeof body.key === "string" && body.key.length > 0 ? body.key : generateIngestKey();
   const keyHash = await hashIngestKey(key);
-  await env.DB.prepare("INSERT INTO hosts (host_id, key_hash, created_at, revoked_at) VALUES (?, ?, ?, NULL)")
+
+  // A *revoked* host_id must be re-provisionable without hand-editing D1
+  // (#5082): a revoked row is a dead credential, but `host_id` is the primary
+  // key, so it would otherwise reserve the name forever. This single statement
+  // both creates a brand-new host and re-mints a retired one:
+  //
+  //   - live row  → `DO UPDATE ... WHERE hosts.revoked_at IS NOT NULL` does not
+  //     match, so nothing changes and `meta.changes` is 0 ⇒ 409 (accidental
+  //     re-minting of an *active* host still fails).
+  //   - revoked row → the conflict target matches and the row is rewritten with
+  //     the NEW `key_hash` and a cleared `revoked_at`. The old (revoked) hash is
+  //     replaced, never reactivated, and `created_at` is restamped because the
+  //     row now describes a freshly minted credential.
+  //
+  // It is an upsert rather than DELETE + INSERT precisely so there is no window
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix. The atomic upsert correctly replaces the unfiltered existence check:

- **Correctness**: verified the SQL logic against the `hosts` schema (`migrations/0001_init.sql`) — `host_id` is the PK, `revoked_at TEXT` is nullable. The `WHERE hosts.revoked_at IS NOT NULL` guard on `DO UPDATE` means a live row produces `meta.changes === 0` → 409, while a revoked row is rewritten with a new `key_hash`/`created_at` and cleared `revoked_at`. No delete+insert window, no TOCTOU gap between check and write.
- **Tests**: ran `npm run check` in `dashboard/` — 231/231 pass, typecheck clean, matching the PR's claimed numbers exactly. Also independently reproduced the negative control (reverted `src/index.ts` to main, reran `test/ingest.test.ts`): the 3 new tests fail with `409` instead of `201` as claimed, the other 22 pass — confirms the tests genuinely gate the fix.
- **Regression coverage**: the pre-existing `rejects creating a duplicate host_id` test (live `host-abc` fixture) is untouched and still passes, confirming the live-host 409 path is preserved.
- **Docs**: `deploy-runbook.md` option B and `.loom/docs/observability.md`'s #5082 gap note are both updated accurately to reflect the new revoke-then-reprovision flow, no raw D1 `DELETE` needed anymore.
- **Diff scope**: confirmed via `origin/main` (not stale local `main`) that the diff is exactly the four files described in the PR body — no scope creep.

## Test Execution

1. `npm run check` in `dashboard/` — ✅ Executed: 231 passed / 10 files, exit 0 (matches claim)
2. Negative control (revert `src/index.ts`, rerun `ingest.test.ts`) — ✅ Executed: 3 new tests fail as expected, 22 pass (matches claim)
3. `tsc --noEmit` clean — ✅ Executed: clean, no errors
4. Manual verification — N/A per PR description (D1 binding is real SQLite via `apply-migrations.ts`, not mocked)

CI is green and merge state is CLEAN.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (another Judge's approval landed at 2026-08-03T19:56:05Z while I was mid-review). Discarding my review; not writing any label — the existing approval already covers the ground I checked.

For the record, my independent pass reached the same conclusion:
- SQL guard is airtight: `ON CONFLICT ... WHERE hosts.revoked_at IS NOT NULL` means a live row's `DO UPDATE` doesn't match, so `meta.changes` stays 0 and the route returns 409; a revoked row is rewritten in place (new `key_hash`, `created_at` restamped, `revoked_at` cleared). Single statement, no delete+insert window, no TOCTOU gap, and D1/SQLite transactional semantics rule out a race where a live host's key could be silently overwritten.
- `npm run check` in `dashboard/`: 231/231 passed, `tsc --noEmit` clean.
- Confirmed the pre-existing `rejects creating a duplicate host_id` (live `host-abc`) regression test still passes unchanged, and the 3 new tests cover provision→revoke→re-provision, dead-old-key verification (401 on the pre-revoke key), and the explicit-key path.
- Merge state CLEAN, only 1 commit behind main (`#5102`, unrelated file — no conflict), CI green.
- Docs (`deploy-runbook.md`, `.loom/docs/observability.md`) updated accurately to match the shipped behavior.

No blockers found — concurring with the existing approval.

PR #5104, opened 2026-08-03
fix(dashboard): exclude sweep-only hosts from the SPA's fleet-overview headline count

### LINKED ISSUE #5101: dashboard: SPA's fleet summary counts sweep-only 'unknown' hosts in its 'N hosts' headline, inflating the count in production

rjwalters:
## Summary

`dashboard/web/src/fleet.ts`'s `buildFleetView` deliberately unions hosts known from `snapshot.hosts` with hosts known only from `activeSweeps` (status `"unknown"`) — that union is unchanged, still documented, and still tested (see `fleet.test.ts`'s "includes a host known only from activeSweeps" test, preserved as-is).

The bug was narrower: `dashboard/web/src/views/fleetOverview.ts`'s `fleetOverviewView` headline ("N hosts" / "M active sweeps") used `view.hosts.length`, which counted those unknown-status hosts too, inflating the "N hosts" count on the production SPA (`dashboard.2amlogic.com`).

- Adds `FleetView.reportingHosts` — the count of hosts with a real `health`/`tokens` entry (`status !== "unknown"`) — and uses it for the headline's "N hosts" text instead of `hosts.length`.
- `totalSweeps` is deliberately left unsplit. Unlike `dashboard/src/publicPage.ts`'s flat sweep table (which needed a second "unattributed sweeps" table to keep sweep-only hosts discoverable — #5078), the SPA's card-based layout already groups every sweep, including a sweep-only host's, under its own card. Reasoning documented inline on `FleetView.reportingHosts`.
- Verified `needsAttention` already excludes `"unknown"`-status hosts (it only counts `stale`/`degraded`) and pinned that with a dedicated regression test.

## Test plan

- [x] `dashboard/web/test/fleet.test.ts`: added `reportingHosts` coverage (multi-host fixture: 6 hosts total, 5 reporting) and a dedicated test pinning that `needsAttention` excludes `"unknown"` hosts.
- [x] `dashboard/web/test/fleetOverview.test.ts`: updated the existing headline-summary test (now "5 hosts", not "6 hosts") and added a regression test asserting the sweep-only host's card (and its sweep) still renders while the headline excludes it.
- [x] `dashboard/web/test/fleet.test.ts`'s "includes a host known only from activeSweeps" test is unchanged and still passes.
- [x] `npm run check` (typecheck + vitest) in `dashboard/web` — 33 files, 466 tests, all passing.

Closes #5101

### DIFF
--- a/dashboard/web/src/fleet.ts
+++ b/dashboard/web/src/fleet.ts
@@ -81,8 +81,29 @@ export interface HostView {
 
 export interface FleetView {
   hosts: HostView[];
+  /** Hosts with a real `health`/`tokens` entry — `hosts.length` minus those
+   * known only from `activeSweeps` (`status === "unknown"`, see the module
+   * doc's "union" note). This, not `hosts.length`, is what the overview
+   * headline's "N hosts" count uses (#5101): `hosts` deliberately still
+   * includes sweep-only hosts (and their cards), so counting `hosts.length`
+   * in the headline would tell an operator the fleet has more reporting
+   * hosts than it does. Mirrors the equivalent fix in
+   * `dashboard/src/publicPage.ts`'s `renderFleetOverview` (#5078).
+   *
+   * `totalSweeps` is deliberately NOT split the same way: unlike the public
+   * page's flat table (which needed a second "unattributed sweeps" table to
+   * keep a sweep-only host's sweeps discoverable), every sweep here already
+   * renders under its own host's card — including a sweep-only host's own
+   * card — so the per-card grouping already answers "whose sweep is this",
+   * and splitting the headline number too would only add a second count to
+   * reconcile against the cards below it.
+   */
+  reportingHosts: number;
   totalSweeps: number;
-  /** Hosts in `stale` or `degraded` — the count the overview headline shows. */
+  /** Hosts in `stale` or `degraded` — the count the overview headline shows.
+   * `"unknown"` hosts are excluded here too (STATUS_ORDER treats them as
+   * their own bucket, not `stale`/`degraded`) — see the "excludes sweep-only
+   * hosts" test in `fleet.test.ts`. */
   needsAttention: number;
 }
 
@@ -268,6 +289,7 @@ export function buildFleetView(snapshot: FleetSnapshot, now: Date = new Date()):
 
   return {
     hosts,
+    reportingHosts: hosts.filter((host) => host.status !== "unknown").length,
     totalSweeps: snapshot.activeSweeps.length,
     needsAttention: hosts.filter((host) => host.status === "stale" || host.status === "degraded").length,
   };
--- a/dashboard/web/src/views/fleetOverview.ts
+++ b/dashboard/web/src/views/fleetOverview.ts
@@ -253,7 +253,11 @@ export function fleetOverviewView(view: FleetView, now: Date = new Date()): HTML
     el(
       "div",
       { class: "overview__summary", data: { testid: "fleet-summary" } },
-      el("span", {}, `${view.hosts.length} host${view.hosts.length === 1 ? "" : "s"}`),
+      el(
+        "span",
+        {},
+        `${view.reportingHosts} host${view.reportingHosts === 1 ? "" : "s"}`,
+      ),
       el("span", {}, `${view.totalSweeps} active sweep${view.totalSweeps === 1 ? "" : "s"}`),
       el(
         "span",
--- a/dashboard/web/test/fleet.test.ts
+++ b/dashboard/web/test/fleet.test.ts
@@ -114,13 +114,42 @@ describe("buildFleetView", () => {
     const built = view();
     expect(built.totalSweeps).toBe(3);
     // Only the stale host and the truly-at-the-edge degraded host — not the
-    // partially-exhausted-but-healthy one.
+    // partially-exhausted-but-healthy one, and not the sweep-only "unknown"
+    // host either — see the dedicated #5101 test below.
     expect(built.needsAttention).toBe(2);
+  });
+
+  // #5101: the SPA's fleet-overview headline uses `reportingHosts`, not
+  // `hosts.length`, so a host known only from activeSweeps ("unknown"
+  // status) does not inflate the "N hosts" count — while still remaining in
+  // `hosts` (and rendering its own card, per the module doc's union rule).
+  it("excludes sweep-only 'unknown' hosts from reportingHosts, but not from hosts", () => {
+    const built = view();
+    expect(built.hosts).toHaveLength(6);
+    expect(built.reportingHosts).toBe(5);
+    expect(findHost(built, SWEEP_ONLY_HOST_ID)?.status).toBe("unknown");
+  });
+
+  // Pin: needsAttention (stale/degraded only) already excludes "unknown"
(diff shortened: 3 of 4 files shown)


rjwalters:
Reviewed as Judge (evaluating per explicit request).

**Correctness / scope vs #5101 AC:**
- `FleetView.reportingHosts` (`dashboard/web/src/fleet.ts`) is defined as `hosts.filter(h => h.status !== "unknown").length`, which matches `HostStatus`'s own doc comment for `"unknown"` ("known only from `activeSweeps`, or a `hosts` entry with neither `health` nor `tokens` yet") — the exact criterion the issue asked for.
- `fleetOverview.ts`'s headline now reads `view.reportingHosts` instead of `view.hosts.length`; the sweep-only host still renders its card (`buildFleetView`'s union of `hosts`/`activeSweeps` keys is untouched), so the "busy but no health data yet" visibility #5078/#5101 both call out as must-preserve is intact.
- `totalSweeps` deliberately left unsplit, with a clear inline rationale (SPA's card-based layout already groups a sweep-only host's sweeps under its own card, unlike the flat-table `publicPage.ts` case from #5078) — reasonable design call, not a gap.
- `needsAttention` already excluded `"unknown"` (filters only `stale`/`degraded`); the PR adds a dedicated regression test pinning that, satisfying the issue's third bullet.

All three acceptance criteria in #5101 are met: headline excludes sweep-only hosts, a regression test covers it (`fleetOverview.test.ts`'s new test asserts "5 hosts" not "6 hosts" while the sweep-only host's card + 1 sweep still render), and the existing `fleet.test.ts` "includes a host known only from activeSweeps" test is preserved untouched.

**Code quality:** small, focused diff; doc comments on both new/changed fields explain the *why*, not just the *what*; naming (`reportingHosts`) is clear and consistent with the existing `needsAttention` pattern.

**Test coverage:** `fleet.test.ts` adds two tests (reportingHosts exclusion, needsAttention exclusion pinned with a fresh sweep-only-host fixture) plus an empty-fleet assertion; `fleetOverview.test.ts` updates the "6 hosts" → "5 hosts" assertion and adds a DOM-level regression test. Verified locally in the builder worktree: `npm run check` (typecheck + vitest) — 33 files, 466 tests, all passing.

**CI / merge state:** `gh pr checks` — all applicable jobs pass (Dashboard Tests, AGENTS.md Sync, CLAUDE.md Line Budget, Detect Changes, Docs/Defaults Parity); unrelated Rust/shell/installer jobs correctly skip (dashboard-only change). `mergeStateStatus` is CLEAN.

No blockers. Approving.

PR #5098, opened 2026-08-03
test(daemon): add cross-cutting fixture for all-roles-failing fleet health

rjwalters:
## Summary

Adds the regression fixture called for by the final AC of epic #5004 —
a fleet with every managed repo's role-runner ticks persistently failing
(past Gap 3's escalation threshold), while dispatch/tokens/queues/
throughput are all healthy, must not present as green anywhere an
operator looks by default.

The new test (`observability/collector.rs`,
`all_repos_failing_roles_is_not_green_anywhere_while_every_other_axis_is_healthy`)
builds a `health::HealthInputs` fixture with three repos each carrying
`ROLE_TICK_ESCALATION_THRESHOLD + 2` consecutive identical-failure
`RoleTickRecord`s (a raw, multi-KB, ANSI-laden failure detail per repo —
an uncleaned claude-wrapper tail) and every other axis fully healthy, then
asserts against `crate::health::assess`'s combined verdict:

- the overall verdict is `Degraded` (not `Green`), with `exit_code() ==
  EXIT_DEGRADED`
- the `roles` section is the one that's down; every other section
  (`dispatch`, `tokens`, `queues`, `throughput`) stays `Green` — this is a
  roles-only outage, not a masked fleet-wide failure
- all three escalated repos are named in the roles summary, not just the
  first (Gap 3, #5023/PR #5039)
- the rendered summary (both the `roles` section and the full
  `render_human()` report) stays ANSI-clean and bounded despite each repo
  contributing a multi-KB raw ANSI-laden detail (Gap 4, #5024/PR #5043)
- the same fixture's records, run through the real production
  `sample_role_tick_health` function, populate
  `HostHealthRecord.roles.persistent` for every failing repo with an
  ANSI-clean structured detail — the Gap 2 host-telemetry path (#5022/PR
  #5042) that gets a role failure off one host and into fleet-wide
  observability

None of the four per-gap PRs (#5050, #5042, #5039, #5043) exercised this
combined scenario — each covers its own axis in isolation. This fixture is
purely additive test code in `loom-daemon/src/observability/collector.rs`;
no production code changed.

## Test plan

- [x] `cargo test --lib observability::collector` — 30 passed (new fixture included)
- [x] `cargo test --lib health::` — 93 passed
- [x] `cargo test --lib` (full loom-daemon suite) — 3243 passed, 0 failed
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --lib --tests -- -D warnings` — clean

Closes #5076

[...]

PR #5106, opened 2026-08-03
feat(observability): backfill sweep.completed/outcome for restart-adopted sweeps

### LINKED ISSUE #5084: sweeps adopted across a daemon restart never export sweep.completed/sweep.outcome — the data sits in the local journal while the backend silently under-counts
## Finding

A sweep that is **adopted across a daemon restart** writes its terminal record to the
durable local journal but **never exports it**. The data exists on disk; the
observability backend never sees it. Every restart therefore silently drops the outcome
telemetry for every sweep in flight at that moment.

## Evidence (robb-studio, 2026-08-03)

Three daemon restarts today (a `LOOM_RUNTIME_JUDGE` revert and two `LOOM_HOST_ID`
changes). Six sweeps were in flight across them and were adopted by the new daemon —
`#4989`, `#5024`, `#5047`, `#5054`, `#5067`, and `#87` (gf180-trng).

Four of those issues are now **closed** — the sweeps ran to completion and merged their
PRs. Yet in D1:

```sql
SELECT sweep_id, kind FROM records
WHERE sweep_id IN ('sweep-issue-4989-…','sweep-issue-5024-…','sweep-issue-5047-…','sweep-issue-5054-…')
  AND kind IN ('sweep.completed','sweep.outcome','sweep.phase');
```

returns **only `sweep.phase` rows** (newest 17:14 for #4989, 17:36 for #5054) — zero
`sweep.completed`, zero `sweep.outcome`.

This is not a general export failure. Terminal records flow normally fleet-wide:

| kind | robb-studio | robb-pro | loom-worker-1 | newest (robb-studio) |
|---|---|---|---|---|
| `sweep.completed` | 788 | 167 | 136 | 2026-08-03T18:20:02 |
| `sweep.outcome` | 788 | 167 | 136 | 2026-08-03T18:20:02 |

**And the lost data is still on disk.** Each of the four has exactly one line in
`.loom/logs/sweep-outcomes.jsonl` (201 lines total) — the durable journal from #4704,
which is explicitly *"persisted locally regardless of whether any exporter is
configured"*. So the outcome was computed and recorded; only the export is missing.

## Two consequences

1. **Silent analytics gap.** `sweep.outcome` carries model, effort, result, and
   `total_duration_sec` — the cost/duration data. Restart-adopted sweeps are absent from
   the backend while present locally, so any fleet-level cost or success-rate analysis
   under-counts by however many sweeps were in flight at each restart. Nothing reports
   the discrepancy.
2. **Leaked Durable Object entries.** `sweep:<sweepId>` entries are deleted on
   `sweep.completed` (`fleetState.ts:383-390`). No completion record → the entry leaks
   until `STALE_SWEEP_MS` (4h). Today that produced six orphans which **synthesised a
   phantom host card** on the dashboard and inflated the header to "4 hosts · 20 active
   sweeps" (#5078).

#4955's module doc anticipated the mechanism — *"most commonly across a daemon restart
whose reaper-driven terminal transition does not get re-exported"* — and the TTL is its
accepted backstop. What that framing understates is that this is **systematic, not
occasional**: it happens on every restart, and the fleet restarts routinely (self-update
rolls, config changes, watchdog relaunches). Six lost outcome records from three
restarts in one afternoon.

## Why it should be fixable cheaply

The journal is already the designed record of truth for outcomes. So rather than teaching
the adoption path to reconstruct a terminal record, the exporter could treat the journal
as its queue-of-record:

- On startup (and/or periodically), export journal entries the backend has not
  acknowledged — a backfill pass keyed on `sweep_id`.
- Or have the adoption path mark adopted sweeps so their terminal transition is exported
  by whichever daemon observes it, rather than being dropped because the new daemon lacks
  the originating context.

The first is more robust: it fixes lost exports from *any* cause (daemon down, transport
error, exporter queue drop), not just adoption. Note there was already a real transport
failure in the log on 2026-08-01 (`observability: export failed, 1 record(s) remain
queued`) — same class of loss, no backfill.

## Acceptance criteria

- [ ] A sweep adopted across a daemon restart still produces `sweep.completed` and
      `sweep.outcome` in the backend when it finishes.
- [ ] Outcome records present in `.loom/logs/sweep-outcomes.jsonl` but absent from the
      backend are eventually exported (backfill), not silently stranded.
- [ ] Backfill is idempotent — re-exporting an already-ingested outcome must not
      double-count.
- [ ] A `sweep:` DO entry is not leaked for a sweep whose completion was observed
      locally.
- [ ] The gap is measurable: some way to see "N local outcomes not yet in the backend",
      so this cannot regress silently.
- [ ] The six outcomes lost today are either backfilled or explicitly written off.

## Provenance

Found while diagnosing why six orphaned `sweep:` entries did not self-heal after the
robb-studio rename (#5068). I twice predicted they would clear on the next phase
transition; they did not, because those sweeps *finished* and their terminal records were
never exported — which turned out to be the more interesting bug. Related: #5078 (the
phantom card the leak produces), #4955 (the TTL backstop and its module doc), #4704 (the
outcome journal), #5083 (no positive confirmation telemetry is flowing).

rjwalters:
## Summary

Sweeps adopted across a daemon restart wrote their terminal outcome to the
local journal (`.loom/logs/sweep-outcome-telemetry.jsonl`, #4704) but never
reliably exported `sweep.completed`/`sweep.outcome` to the observability
backend: the live event-bus collector can only attach the *correct*
`sweep_id` to a terminal event by having observed that sweep's own dispatch
event earlier in the same daemon process — a restart-adopted sweep's dispatch
happened in the *previous* process, so the terminal record fell back to a
synthesized `unknown-issue-{N}` id. The record could still export, but under
an id nothing else could query it by, so any per-sweep backend query (D1, and
the dashboard's `sweep:<sweepId>` Durable Object cleanup) found nothing —
matching the reported incident (six lost outcome records across three
restarts, and a phantom-dashboard-card leak, #5078).

## Changes

- **`loom-daemon/src/observability/backfill.rs`** (new): treats the durable
  `sweep-outcome-telemetry.jsonl` journal as the export queue-of-record — it
  already carries the *correct* `sweep_id` for every terminal transition,
  independent of whether the live collector saw the matching dispatch. A
  backfill pass reads every journal envelope not yet handed to the export
  queue (tracked by a small on-disk cursor,
  `observability-backfill-state.json`), synthesizes the paired
  `sweep.completed` record (the journal only stores the `sweep.outcome`
  shape), and pushes both onto the existing `DurableQueue` for the
  unmodified sender/exporter to drain.
- **`loom-daemon/src/observability/collector.rs`**: runs the backfill pass
  once at collector startup (so a restart-adopted sweep's record does not
  wait a full snapshot interval) and again on every snapshot tick, across
  every workspace the daemon manages (`WorkspacePool::provisioned_registries`
  — a multi-repo daemon's adopted sweeps can live in a sibling repo's own
  journal, not just the primary workspace).
- **`loom-daemon/src/main.rs` + `src/cli/misc_cmds.rs`**: `loom-daemon
  sweep-outcomes --pending-export` reports how many local records the
  backfill drain has not yet attempted — the AC's "N local outcomes not yet
  in the backend" measurability gauge, so this cannot regress silently.
- **`dashboard/migrations/0002_idempotent_terminal_records.sql`** +
  **`dashboard/src/index.ts`**: true idempotency is enforced on the
  backend, not the daemon (the daemon-side cursor is only an efficiency
  optimization). A partial `UNIQUE(kind, sweep_id)` index scoped to exactly
  `sweep.completed`/`sweep.outcome`, plus `/ingest` switching to
  `INSERT OR IGNORE`, silently absorbs a re-offered record instead of
  duplicating the row — covers both the daemon-side backfill re-offer case
  and a plain transport-retry resend.
- **`.loom/docs/observability.md`**: documents the backfill mechanism and
  the backend-side idempotency contract.
- Fixed a pre-existing compile error in `collector.rs`'s own test module
  (`HealthInputs` gained a `gh_unavailable` field in #5097 after a fixture
  in #5098 was merged without it) — blocked `cargo test --lib` outright, so
  I added the one missing field to get back to a compiling baseline.

- [x] A sweep adopted across a daemon restart still produces
      `sweep.completed` and `sweep.outcome` in the backend when it
      finishes — the backfill pass reads the always-written local journal,
      independent of whether this process observed the dispatch. Covered by
      `backfill_pushes_a_completed_and_outcome_pair_with_the_real_sweep_id`
      (asserts the real `sweep_id`, never `unknown-issue-{N}`).
- [x] Outcome records present in the local journal but absent from the
      backend are eventually exported, not silently stranded — first-boot
      pass at collector startup, plus every snapshot tick thereafter.
- [x] Backfill is idempotent — re-exporting an already-ingested outcome
      does not double-count: backend partial unique index +
      `INSERT OR IGNORE` (migration 0002), verified by 5 new dashboard
      tests including exact-duplicate re-ingest and a duplicate mixed into
      a fresh batch.
- [x] A `sweep:` DO entry is not leaked for a sweep whose completion was
      observed locally — `sweep.completed` now reliably reaches
      `fleetState.ts`'s existing deletion path under the real `sweep_id`.
- [x] The gap is measurable — `loom-daemon sweep-outcomes --pending-export`
      (text and `--json`), backed by `pending_count`, which is exactly the
      count the next backfill pass would enqueue.
- [x] The six outcomes lost in the reported incident: this PR makes the
      mechanism correct and idempotent going forward (per the curator's
      note, no live backfill run against production data was required for
      this issue) — a daemon picking up this build will backfill any of
      those six still sitting in `sweep-outcome-telemetry.jsonl` on its next
      collector startup.

## Test Plan

- `cargo test --lib` (loom-daemon): 3261 passed, 0 failed — includes 10 new
  tests in `observability::backfill` plus the pre-existing suite unaffected.
- `cargo clippy --lib --bins --tests --all-features -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `npm test` (dashboard, `vitest run` via `pretest`): 250 passed (10 files),
  including 5 new idempotency tests in `test/ingest.test.ts` (exact-duplicate
  `sweep.completed`/`sweep.outcome`, duplicate-mixed-into-a-fresh-batch,
  `sweep.phase` correctly NOT deduped, two distinct sweeps' `sweep.completed`
  correctly NOT deduped).
- `npx tsc --noEmit` (dashboard): clean.
- Manual CLI verification: `loom-daemon sweep-outcomes --workspace <dir>
  --pending-export` against an empty journal (`0 ... caught up`) and a
  seeded journal (`1 local outcome record(s) ... not yet attempted`), both
  text and `--json`.

Closes #5084

### DIFF
--- a/.loom/docs/observability.md
+++ b/.loom/docs/observability.md
@@ -90,6 +90,36 @@ field-by-field reference, the `visibility` anti-leak contract, and the local
 exporter is configured**):
 [`.loom/docs/telemetry-schema.md`](telemetry-schema.md).
 
+### Backfill: the local journal is the export queue-of-record (#5084)
+
+The live collector can only attach the correct `sweep_id` to a terminal
+`sweep.completed`/`sweep.outcome` record by having observed that sweep's
+*own* dispatch event earlier in the same daemon process. A sweep **adopted
+across a daemon restart** was dispatched by the previous process, so this
+falls through to a synthesized `unknown-issue-{N}` id — the record can still
+export, but under an id nothing else can query it by, which is what produced
+the "sweeps that finished were silently missing from the backend" incident
+this issue documents.
+
+`loom-daemon/src/observability/backfill.rs` fixes this at the root: the local
+`sweep-outcome-telemetry.jsonl` journal above is written unconditionally at
+every terminal transition — with the *real* `sweep_id`, independent of
+whether the live collector saw the dispatch — so a backfill pass (run once at
+collector startup, then every `SNAPSHOT_INTERVAL`) treats it as the queue of
+record. It re-offers any envelope not yet handed to the export queue
+(tracked by a small on-disk cursor, `observability-backfill-state.json`) —
+self-healing *any* lost export, not just the restart-adoption case (a
+transport failure or a queue drop recovers the same way). `loom-daemon
+sweep-outcomes --pending-export` reports how many local records have not yet
+been attempted, so the backlog is visible rather than silent.
+
+True idempotency — a re-offered record must never double-count — is enforced
+on the **backend**, not the daemon: `dashboard/migrations/0002_…sql` adds a
+partial `UNIQUE(kind, sweep_id)` index scoped to exactly `sweep.completed`/
+`sweep.outcome`, and `/ingest` uses `INSERT OR IGNORE`, so a duplicate is
+silently absorbed rather than inserted twice. The daemon-side cursor is only
+an efficiency optimization against needless re-sends.
+
 ## 3. Exporters: HTTPS (default) or OTLP (opt-in)
 
 The default exporter is `HttpsExporter` — JSON-over-HTTPS `POST /ingest`,
--- a/dashboard/migrations/0002_idempotent_terminal_records.sql
+++ b/dashboard/migrations/0002_idempotent_terminal_records.sql
@@ -1,0 +1,29 @@
+-- Idempotent terminal-record ingest (Issue #5084).
+--
+-- `sweep.completed` / `sweep.outcome` are each written exactly once per
+-- sweep by the daemon's own reaper — but the daemon-side backfill drain
+-- (loom-daemon/src/observability/backfill.rs) is only an *efficiency*
+-- optimization against re-offering an already-queued record; it is not the
+-- correctness mechanism. A crash between enqueueing a backfilled batch and
+-- persisting the advanced cursor re-offers the same record on the next
+-- pass, and a transport retry can do the same for a live-observed record.
+-- Without a constraint here, either would silently double the row for that
+-- sweep in `records` (and double-count it in any aggregate query over the
+-- table).
+--
+-- A partial UNIQUE index — scoped to exactly the two kinds that are emitted
+-- once per sweep — is the backend-side guarantee: `INSERT OR IGNORE`
+-- (src/index.ts's ingest handler) silently absorbs a duplicate `(kind,
+-- sweep_id)` pair instead of inserting a second row, however many times a
+-- backfill pass or retry re-sends it. Every other `kind` (`sweep.started`,
+-- `sweep.phase`, `tokens.snapshot`, `host.health`, …) is untouched — some
+-- are legitimately emitted many times per sweep/host (`sweep.phase` once
+-- per lifecycle phase) or carry no `sweep_id` at all (the host-level
+-- kinds), and SQLite's partial-index WHERE clause means the constraint
+-- never applies to their rows.
+--
+-- NULL `sweep_id` values are exempt from uniqueness by SQL's own NULL != NULL
+-- rule, so this is safe even though `sweep_id` is nullable on the column.
+CREATE UNIQUE INDEX idx_records_terminal_sweep_once
+  ON records (kind, sweep_id)
+  WHERE kind IN ('sweep.completed', 'sweep.outcome');
--- a/dashboard/src/index.ts
+++ b/dashboard/src/index.ts
@@ -259,8 +259,19 @@ async function handleIngest(request: Request, env: Env): Promise<Response> {
   const nowIso = new Date().toISOString();
   const statements = envelopes.map((envelope) => {
     const fields = extractRecordFields(envelope.record);
+    // `OR IGNORE` (Issue #5084): `idx_records_terminal_sweep_once`
+    // (migrations/0002) enforces a partial UNIQUE(kind, sweep_id) for
+    // exactly `sweep.completed`/`sweep.outcome` — the two kinds a sweep
+    // emits once, ever. A re-sent record for a sweep already ingested
+    // (the daemon-side backfill drain's cursor is an efficiency
+    // optimization, not a correctness guarantee — see
+    // `loom-daemon/src/observability/backfill.rs`'s module doc) is
+    // silently absorbed here instead of duplicating the row. Every other
+    // `kind` has no matching unique index, so this is a no-op change for
+    // them — `INSERT OR IGNORE` behaves exactly like `INSERT` when there is
+    // no conflicting constraint.
     return env.DB.prepare(
-      `INSERT INTO records
+      `INSERT OR IGNORE INTO records
          (schema_version, emitted_at, host_id, kind, repo, visibility, issue, sweep_id, payload, ingested_at)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
     ).bind(
--- a/dashboard/test/ingest.test.ts
+++ b/dashboard/test/ingest.test.ts
@@ -1,7 +1,15 @@
 import { createExecutionContext, env, waitOnExecutionContext } from "cloudflare:test";
 import { beforeEach, describe, expect, it } from "vitest";
 import worker from "../src/index";
-import { hostHealthEnvelope, revokeHost, seedHost, sweepStartedEnvelope } from "./testHelpers";
+import {
+  hostHealthEnvelope,
+  revokeHost,
+  seedHost,
+  sweepCompletedEnvelope,
+  sweepOutcomeEnvelope,
+  sweepPhaseEnvelope,
+  sweepStartedEnvelope,
+} from "./testHelpers";
 
 function ingestRequest(body: unknown, authHeader?: string): Request {
   const headers: Record<string, string> = { "content-type": "application/json" };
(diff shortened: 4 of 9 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T20:30:46Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verified**
- `cargo build --lib` clean; `cargo test --lib observability::backfill` (10/10) and `observability::collector` (30/30) pass; `cargo fmt --check` clean.
- Dashboard `npm test`: 250/250 passed, including the 5 new idempotency tests in `test/ingest.test.ts`.
- CI (all checks, including the two that were still `pending` at claim time — Rust Unit Tests, Rust OTLP Feature Build & Test): all green. `mergeStateStatus` is `CLEAN`.
- `git merge-tree` against `origin/main`: no conflicts (branch was 2 commits behind but merges cleanly).

**Correctness review, per the three flagged questions**

1. **Double-export / race with the live collector path**: by design, not a bug. Every terminal transition (restart-adopted or not) writes the local `sweep-outcome-telemetry.jsonl` journal *and* fires the live `SweepExited`/`SweepCrashed` bus event unconditionally (`sweep_registry/reaper.rs`'s `append_outcome_journal` + `emit_event` are always paired). So for an in-process sweep, both the live path and the next backfill pass will offer the *same* `(kind, sweep_id)` pair — the module doc for `backfill.rs` states this explicitly and pushes the correctness burden onto the backend's idempotent `INSERT OR IGNORE`, not the daemon-side cursor (correctly documented as "efficiency optimization, not the correctness mechanism"). Confirmed the cursor-then-persist ordering in `run_backfill_pass` (advance+save happens after the queue pushes, so a crash mid-batch re-offers rather than loses records — fail-safe direction is correct).

2. **D1 migration idempotency**: `idx_records_terminal_sweep_once` is a partial `UNIQUE(kind, sweep_id)` index scoped to exactly `sweep.completed`/`sweep.outcome` — correctly leaves `sweep.phase` (legitimately emitted many times per sweep) and the host-level kinds (no `sweep_id`) untouched. `INSERT OR IGNORE` is per-statement inside `D1Database.batch()`'s single transaction, so one duplicate in a mixed batch doesn't abort the whole batch — verified by the new "duplicate does not block the rest of a mixed batch" test. `NULL != NULL` correctly exempts every non-terminal row from the constraint.

3. **Incidental `gh_unavailable` fix**: correct and safe — `HealthInputs.gh_unavailable: None` matches the same "gh available / healthy" sentinel every other pre-existing healthy-fixture in `health.rs` already uses (e.g. line 1878). It's a pure fixture-completeness fix for a merge-race gap between #5097/#5098, not masking any logic change; `cargo build --lib` now compiles clean where it didn't before.

**One non-blocking observation for a possible follow-up**: `map_event_to_records` (collector.rs, unchanged by this PR) still falls back to a synthesized `unknown-issue-{N}` sweep_id when the *live* event-bus path observes a restart-adopted sweep's own termination in the new process (no `DispatchState` replay across restart). That still produces one extra, harmless-but-noisy `records` row under the fake id alongside the correct backfilled one — distinct from the actual bug this issue fixes (the leaked `sweep:` DO entry, which *is* fixed, since the real-id `sweep.completed` now reliably reaches `fleetState.ts`'s deletion path via backfill regardless of what the live path also sends). Worth a small follow-up issue if the extra row output ever becomes a query-noise problem, but out of scope here.

Nice test coverage overall — the `run_backfill_pass_all_covers_every_provisioned_workspace` test in particular for the multi-repo case is a good catch matching the reported incident's "spanned repos" detail.

rjwalters:

- **Critical File Exclusion**: modifies `dashboard/migrations/0002_idempotent_terminal_records.sql`, a database schema migration file. Per Champion policy, migration files always require human review regardless of how the rest of the diff scores on the merge-risk judgment.

**Next steps:**
- A human should review the migration (idempotency of the D1 schema change, forward/backward compatibility with already-deployed backfill code) and merge manually with `./.loom/scripts/merge-pr.sh 5106` if it looks safe.

Keeping `loom:pr` label. Champion will not re-evaluate this specific criterion — it does not clear on its own — but will not re-comment either.

---
*Automated by Champion role*

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (with `loom:review-requested`/`loom:reviewing` removed) — another Judge's verdict raced mine while I was reviewing. My own independent pass (Rust backfill tests + dashboard vitest suite run locally, both green; CI green; merge state CLEAN) reached the same conclusion — approve — so no disagreement to flag. Discarding my review; not writing any label. Note: Champion has already flagged this PR for human merge because it touches a DB migration file (`dashboard/migrations/0002_idempotent_terminal_records.sql`) — that gate is unrelated to and unaffected by this stand-down.

Issue #4830, opened 2026-07-31
observability: host_id↔key mismatch is silent — exporter should verify the ingest-echoed bound host_id against its own identity (studio reported as robb-pro for hours)


rjwalters:
## Context
Live incident 2026-07-31: the flagship Mac Studio spent its first hours of telemetry reporting to dashboard.2amlogic.com as `robb-pro` — an ingest key bound to the wrong host_id had been installed on it (operator-side naming confusion), and **nothing anywhere warned**. The backend silently records the key-bound host_id (deploy-runbook §8: "a mismatch is not a security problem, only a confusing one"), the exporter has no idea what its key is bound to, and the mislabeled rows are indistinguishable from real robb-pro data. Diagnosis required manually cross-referencing D1 contents against which physical host held which key file.

## Suggested acceptance criteria
- [ ] The `/ingest` response echoes the authenticated key's bound `host_id` (additive response field, no schema rev needed for records)
- [ ] The observability exporter compares that echoed host_id against its own host identity (same resolution as sweep_registry's host_identity: $LOOM_HOST_ID > $HOSTNAME > hostname) and logs a WARN on mismatch — once per daemon lifetime, not per flush
- [ ] `loom-daemon health` / status surfaces the mismatch as a degraded-observability note when present
- [ ] Runbook §8 updated: mismatch is now detected, not just "confusing"
- [ ] No behavior change when they match; a keyless/disabled exporter is unaffected

rjwalters:
## Curator Enhancement

Verified against `origin/main` (0b6cffd). Duplicate/related-work check: no duplicates, no related open work.

### Implementation Guidance

**1. `/ingest` echoes the bound `host_id`**
`dashboard/src/index.ts:155` (`handleIngest`) authenticates via `authenticateHost()` (`dashboard/src/auth.ts:50-67`), which already resolves `auth.hostId` from the key and is used authoritatively at insert time (`index.ts:205`). The response today is just `{"accepted": N}` (`index.ts:171`, `243-246`) — no `host_id` field. Add `auth.hostId` to the success response body, e.g. `{"accepted": N, "host_id": auth.hostId}`. This is additive/backward-compatible; no envelope schema rev needed (matches the issue's own framing).

**2. Exporter compares echoed `host_id` against its own identity, WARNs once**
[...]

PR #5125, opened 2026-08-03
feat(daemon): report positive telemetry-export liveness in status and health

### LINKED ISSUE #5083: health/status give no positive confirmation that telemetry is flowing — "exporting fine" and "silently never exported" both render as no observability section
## Problem

There is no way to confirm from `loom-daemon health` that telemetry is actually
reaching the backend. The `observability` section renders **only** on a host-id/ingest-key
mismatch (`health.rs:29` — *"only when non-green"*, #4830), so a host that is exporting
fine, a host whose observability is disabled, and a host whose config is broken in any way
other than an id mismatch are **all indistinguishable**: no section at all.

To be clear, #4830's decision is defensible and this issue does not ask to reverse it — a
permanent `observability GREEN — ok` line on a surface whose purpose is anomalies would be
noise. The ask is for a *positive* signal to exist **somewhere**.

## How this cost time

On 2026-08-03, verifying whether robb-studio's telemetry was healthy before a host rename:

- `health` showed 6 sections on robb-studio and 7 on loom-worker-1 (the extra being
  `observability`, degraded).
- Both hosts had **byte-identical** observability config (enabled, same endpoint, same
  `ingestKeyFile` path).
- The absence of the section on robb-studio was the *only* evidence it was fine — which is
  inference from absence, not evidence.
- `~/.loom/logs/observability-queue.jsonl` was **0 bytes**, which is equally consistent
  with "drained successfully" and "nothing ever enqueued".

Confirming it required grepping `~/.loom/daemon.log` for the absence of the exporter's
mismatch warning (`observability/exporter.rs:212`) alongside the presence of
`observability: enabled` lines — i.e. reading logs to answer a question the health surface
is supposed to answer. The operator explicitly challenged the inference, and they were
right to.

## Why it matters more than it looks

This fleet just spent a full migration on telemetry attribution (#5068, #5062, #5063).
"Is this host's telemetry landing, and under what identity?" is now a first-class
operational question, and the answer currently lives only in a log grep. Note also
#5004: the `roles` section reaches no remote consumer, so per-host observability state is
invisible fleet-wide as well.

## Suggested direction — pick the cheapest that gives a positive signal

- Add last-successful-export state to `loom-daemon status` (which already carries steady-state
  facts like `Forge credential: OK`, `Safehouse: connected`): e.g.
  `Observability: OK — last export 12s ago, host_id=robb-studio` / `disabled` /
  `configured, never exported`.
- Or keep the health section conditional but add the same facts to its `detail` payload so
  a machine consumer can assert them even when the section is green/absent.
- Distinguish, explicitly, these states: **not configured** · **configured but never
  exported** · **exporting, ids agree** · **exporting, id mismatch** (today's only visible
  one) · **export failing**.

"Configured but never exported" is the important missing one — it is the silent failure
mode, and it currently looks exactly like healthy.

## Acceptance criteria

- [ ] An operator can confirm a host's telemetry is flowing, and under which `host_id`,
      without reading logs.
- [ ] A host with observability **disabled** is distinguishable from one that is healthy.
- [ ] A host that is configured but has **never successfully exported** is surfaced as a
      problem rather than as silence.
- [ ] The `observability` health section's anomaly-only rendering (#4830) is preserved —
      this adds a signal, it does not make the section unconditional.
- [ ] Whatever is added is machine-readable in `--json`, so a watch loop can assert it.

## Provenance

Surfaced 2026-08-03 when the operator asked why robb-studio's health output lacked an
`observability` section that loom-worker-1 had. The answer was benign (ids agreed) but
took a code read plus a log grep to establish, and the same absence would have been shown
for a genuinely broken exporter. Related: #4830 (the anomaly-only decision), #5063 (host
identity), #5062 (worker-1's live mismatch), #5004 (health state does not leave the host).

rjwalters:
## Summary

`loom-daemon health`'s `observability` section is anomaly-only by design (#4830), so
"exporting fine", "observability disabled", and "configured but silently never exported"
all rendered as the same thing: no section at all. Confirming the healthy case meant
grepping `daemon.log` for the *absence* of a warning — which is what the 2026-08-03
robb-studio-vs-loom-worker-1 incident cost.

This adds a positive, always-answerable export-liveness record. #4830's decision is
preserved verbatim: the health section is still anomaly-only. The positive confirmation
lives on `loom-daemon status` (human + `--json`) instead.

```
Observability: OK — last export 12s ago, 3481 record(s) as host_id=robb-studio → https://…/ingest
Observability: disabled (no telemetry export — set observability.enabled=true to opt in)
Observability: NEVER EXPORTED — running 4h as host_id=robb-studio and no batch has EVER been acked; …
Observability: FAILING — 4 consecutive failed flush(es) as host_id=robb-studio, last success 2h ago … (last error: …)
Observability: HOST-ID MISMATCH — telemetry is landing under host_id=robb-pro, not robb-studio …
```

## Changes

- **`loom-daemon/src/types.rs`** — new `ObservabilityExportStatus` +
  `ObservabilityExportState` (`disabled` | `starting` | `never_exported` | `healthy` |
  `host_id_mismatch` | `failing`, `snake_case` on the wire, `#[serde(other)]` fallback for
  forward compat). `classify()` is **pure**, so `status`, `health` and any future consumer
  reach the same verdict from the same payload. A grace window of 3 × flush interval,
  floored at 10 min (above the collector's own 5-min host-snapshot cadence), keeps a
  just-rolled daemon from reading as broken.
- **`loom-daemon/src/observability/mod.rs`** — `ExportStatus` cell + `GLOBAL_EXPORT_STATUS`,
  following the existing `HostIdStatus` / `global_host_id_mismatch()` pattern, but rolling
  rather than write-once. Registered **only after** an exporter is actually constructed, so
  every degrade-to-disabled path (no endpoint, unreadable ingest key, `otlp` without the
  Cargo feature, client construction failure) keeps reporting `disabled` truthfully.
- **`loom-daemon/src/observability/sender.rs`** — `try_flush` records success/failure on
  every *decided* attempt. An `Empty` queue records **nothing**: nothing was attempted, so
  it is evidence of neither health nor failure. (Recording a success there is exactly how a
  never-exporting host would masquerade as healthy — the ambiguity a 0-byte
  `observability-queue.jsonl` left an operator with.)
- **`loom-daemon/src/ipc.rs`** — `build_daemon_status` publishes
  `observability_export: Some(global_export_status())`. Always `Some` from a daemon of this
  vintage; a disabled exporter reports `disabled`, which is a real answer.
- **`loom-daemon/src/cli/status_render.rs`** — new `render_observability_line()` (split out
  and `now`-injected so every state is unit-testable without capturing stdout, mirroring
  `render_admission_brake_line`), printed unconditionally next to the `Safehouse:` block;
  `observability_export` added to the `--json` payload.
- **`loom-daemon/src/health.rs`** — `assess_observability()` keeps the host-id-mismatch note
  byte-for-byte (including its `detail` keys) and additionally flags `never_exported` and
  `failing`. `healthy` / `starting` / `disabled` still render **nothing at all**. When a
  section does render, its `detail` carries the full export record; the mismatch note gets
  it under `detail.export`.
- **`defaults/docs/observability.md`** — new §3b with the state table, the `jq` assertion,
  and an explicit scope note that this is a *transport-level* signal (a host can be
  `healthy` while a specific record kind is never enqueued — #5084; complementary, not
  overlapping).

Only `defaults/docs/` is edited; the installed `.loom/docs/` copy is left to the periodic
`chore: resync installed Loom surfaces` commit.

- [x] **An operator can confirm a host's telemetry is flowing, and under which `host_id`,
      without reading logs.** `Observability: OK — last export 12s ago, 3481 record(s) as
      host_id=robb-studio → …` on `loom-daemon status`.
      Test: `observability_line_states_health_positively_with_the_host_id`.
- [x] **A host with observability disabled is distinguishable from one that is healthy.**
      `Observability: disabled (…)` vs `Observability: OK — …`.
      Test: `observability_line_distinguishes_disabled_from_healthy`.
- [x] **A host that is configured but has never successfully exported is surfaced as a
      problem rather than as silence.** `NEVER EXPORTED` on `status` **and** a `Degraded`
      `observability` health section (which drives `health`'s exit code).
      Tests: `observability_line_surfaces_never_exported_as_a_problem`,
      `a_never_exporting_host_is_a_degraded_observability_note`,
      `past_the_grace_window_with_no_export_is_never_exported`,
      `a_never_acking_sink_is_visible_as_a_problem`.
- [x] **The `observability` health section's anomaly-only rendering (#4830) is preserved.**
      A healthy exporter still produces the same 6 sections and `Verdict::Green`; disabled
      and starting produce no section; the mismatch note's summary and `detail` keys are
      unchanged, with the new payload strictly additive under `detail.export`.
      Tests: `a_healthy_exporter_still_renders_no_observability_section`,
      `a_disabled_exporter_still_renders_no_observability_section`,
      `a_starting_exporter_is_not_yet_called_out`,
      `a_mismatch_still_wins_and_now_carries_the_export_facts`,
      `a_pre_5083_daemon_reporting_no_export_field_renders_nothing`.
- [x] **Whatever is added is machine-readable in `--json`, so a watch loop can assert it.**
      `loom-daemon status --json | jq -e '.observability_export.state == "healthy"'`;
      health sections carry the record in `detail`.
      Tests: `observability_export_is_machine_readable_in_json`,
      `export_state_serializes_snake_case_for_watch_loops`,
      `export_status_round_trips_and_tolerates_pre_5083_wire_data`,
      plus the IPC round-trip assertion in `ipc.rs`.

Curator's fifth state, **export failing**, is covered too:
`observability_line_reports_a_failing_exporter_with_its_error`,
`a_failing_exporter_is_a_degraded_observability_note`,
`consecutive_failures_classify_as_failing`.

## Test Plan

- `cargo check --all-targets` — clean.
- `cargo clippy --all-targets` — clean, no new warnings.
- `cargo fmt` — no drift.
- `cargo test --lib 'observability::'` — 85 passed.
- `cargo test --lib 'health::'` — 103 passed.
- `cargo test --lib 'types::'` — 26 passed.
- `cargo test --lib 'ipc::'` — 83 passed.
- `cargo test --bins observability` — 11 passed (all the new `status_render` cases).
- Full `bash .loom/scripts/build-gate.sh` (`cargo test --workspace --lib --bins`): 3271
  passed, 1 failed — `sweep_registry::guards::tests::restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr`,
  which passes in isolation and touches no code in this diff. It is a load-flake: the run
  happened at load average 56 on this host, and the test's `gh` stub timing is what it
  asserts on. Not caused by this change.

### Note on a pre-existing break, and why it is not in this diff

While working this, `loom-daemon/src/observability/collector.rs`'s `HealthInputs` fixture
(#5098) did not compile against `health.rs`'s `gh_unavailable` field (#5097) — the two PRs
landed in parallel and neither saw the other, breaking `cargo test` on `main`. That was
fixed upstream mid-flight (it is on `main` now), so no fix for it ships here.

Closes #5083

### DIFF
--- a/defaults/docs/observability.md
+++ b/defaults/docs/observability.md
@@ -135,6 +135,60 @@ published and the `observability` health section stays silent. Choosing OTLP
 therefore trades this particular misconfiguration guardrail away; keep the
 default `"https"` sink if you want it.
 
+## 3b. Confirming telemetry is actually flowing
+
+`loom-daemon health`'s `observability` section is **anomaly-only** by design
+(issue #4830): it renders when something is wrong and stays silent otherwise.
+That is the right call for a surface whose value is that every printed line is
+worth reading — but on its own it left three very different hosts looking
+identical (no section at all): one exporting perfectly, one with observability
+disabled, and one that was configured, running, and had **never successfully
+exported anything**. A `~/.loom/logs/observability-queue.jsonl` of 0 bytes is
+equally consistent with "drained cleanly" and "nothing was ever enqueued", so
+confirming the healthy case meant grepping `daemon.log` for the *absence* of a
+warning.
+
+`loom-daemon status` now states the answer positively (issue #5083):
+
+```
+Observability: OK — last export 12s ago, 3481 record(s) as host_id=robb-studio → https://…/ingest
+```
+
+The same facts are machine-readable under `observability_export` in
+`loom-daemon status --json`, so a watch loop can assert health instead of
+inferring it from silence:
+
+```bash
+loom-daemon status --json | jq -e '.observability_export.state == "healthy"'
+```
+
+| `state` | Meaning | Rendered as |
+|---|---|---|
+| `disabled` | Exporter not running: `enabled=false`, or enabled but under-configured (no endpoint / no readable ingest key / `otlp` without the Cargo feature) | `Observability: disabled …` |
+| `starting` | Running, nothing acked yet, still inside the grace window (3 × `flushIntervalSecs`, floored at 10 min) — a just-rolled daemon, not a fault | `Observability: starting …` |
+| `never_exported` | Running well past the grace window and **no batch has ever been acked** — the silent failure mode | `Observability: NEVER EXPORTED …` |
+| `healthy` | Batches are being acked and the ids agree | `Observability: OK …` |
+| `host_id_mismatch` | Batches are being acked, but under a different `host_id` than this daemon reports for itself (the #4830 condition, §3 above) | `Observability: HOST-ID MISMATCH …` |
+| `failing` | The most recent flush attempt errored; the queue is retrying with backoff | `Observability: FAILING …` |
+
+`observability_export` also carries `host_id`, `endpoint`, `exporter`,
+`started_at`, `last_success_at`, `last_failure_at`, `last_failure_detail`,
+`records_exported`, `consecutive_failures`, and `flush_interval_secs`. A `null`
+`observability_export` means the daemon binary predates #5083 — "cannot tell",
+never "disabled"; restart the daemon onto a current binary.
+
+The health section keeps its anomaly-only contract. It now recognizes two
+additional *non-green* conditions — `never_exported` and `failing` — which are
+anomalies by the same rule that already admitted `host_id_mismatch`; `healthy`,
+`starting`, and `disabled` still render nothing at all. When a section does
+render, its `detail` payload carries the full `observability_export` record, so
+a machine consumer of `loom-daemon health --json` gets the positive facts too.
+
+**Note on scope**: this is a *transport-level* signal — it answers "are batches
+being acked", not "is every record kind being enqueued". A host can report
+`healthy` while a specific record kind is silently never queued (e.g. issue
+#5084); the two checks are complementary.
+
 ## 4. The backend: deploy your own Cloudflare Worker
 
 The Phase-2 backend is a Cloudflare Worker (D1 for durable history, a
--- a/loom-daemon/src/cli/status.rs
+++ b/loom-daemon/src/cli/status.rs
@@ -806,6 +806,7 @@ pub(crate) mod status_client_tests {
             daemon_build_commit: Some("18887b5c".to_string()),
             work_finder_interval_secs: Some(60),
             observability_host_id_mismatch: None,
+            observability_export: None,
         }
     }
 
--- a/loom-daemon/src/cli/status_render.rs
+++ b/loom-daemon/src/cli/status_render.rs
@@ -176,6 +176,15 @@ pub(crate) fn build_status_json_value(
         // reports for itself, so its telemetry is being filed under the wrong
         // host. `null` in the common case (ids agree, or no exporter running).
         "observability_host_id_mismatch": report.observability_host_id_mismatch,
+        // Positive export-liveness signal (#5083). Unlike the anomaly-only
+        // field above, this is non-null for any daemon of this vintage and
+        // always carries a `state`, so a watch loop can assert health rather
+        // than infer it from the absence of a warning:
+        //   loom-daemon status --json \
+        //     | jq -e '.observability_export.state == "healthy"'
+        // States: disabled | starting | never_exported | healthy |
+        // host_id_mismatch | failing. `null` only from a pre-#5083 daemon.
+        "observability_export": report.observability_export,
         "dynamic_cap": {
             "token_pool_size": report.token_pool_size,
             // The directory the daemon resolved for the pool above (#4292) —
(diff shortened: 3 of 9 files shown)


rjwalters:
## Judge Review: Approve

Reviewed the diff, the linked issue (#5083), and CI.

**Design**
- `ObservabilityExportState`/`ObservabilityExportStatus` in `types.rs` is a clean, well-documented state machine. `classify()` is pure and shared by `status`, `health`, and (implicitly) any future consumer — good call, avoids the two-halves-disagree failure mode the PR body specifically calls out.
- The grace window (3x flush interval, floored at 10 min, above the collector's 5-min snapshot cadence) is a sensible guard against false-alarming a freshly-rolled daemon.
- `#4830`'s anomaly-only health contract is preserved verbatim and defended with explicit regression tests (`a_healthy_exporter_still_renders_no_observability_section`, `a_disabled_exporter_still_renders_no_observability_section`, `a_mismatch_still_wins_and_now_carries_the_export_facts`).
- `ExportStatus` is registered as the process-global only *after* the exporter is actually constructed, so every degrade-to-disabled path (missing endpoint, unreadable ingest key, `otlp` without the Cargo feature, client construction failure) truthfully reports `disabled` instead of a phantom running exporter.
- `try_flush`'s `Empty` outcome deliberately records nothing (neither success nor failure) — correctly avoids the "0-byte queue file" ambiguity called out in the issue.
- `last_failure_detail` is documented as never carrying the ingest key (only status codes / truncated body snippets) — reasonable given it's exposed over `status --json`.
- `Unrecognized` state (forward-compat for a newer daemon) degrades legibly rather than misclassifying, mirroring the existing Safehouse pattern (#4464).

**Acceptance criteria (#5083)** — all five verified against tests, each with a dedicated unit test cross-referenced in the PR body. Confirmed by reading the actual `classify()`/`render_observability_line()`/`assess_observability()` logic, not just the test names.

**Docs** — `defaults/docs/observability.md` §3b covers the state table, the `jq` assertion, and an explicit scope note distinguishing this transport-level signal from #5084. Only `defaults/docs/` touched, consistent with the installed-copy resync convention.

**CI / flaky test claim** — verified independently:
- All CI checks are green (`Rust Unit Tests`, `Rust OTLP Feature Build & Test`, `Rust Linting & Build Check`, `Shell Test Suites`, etc.) — CI's own clean run did not reproduce the failure the Builder saw locally.
- Cloned the PR branch into a scratch dir and ran `sweep_registry::guards::tests::restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr` in isolation: passes in 0.62s.
- The test is a subprocess-timing-sensitive `gh` shell-stub test in a module (`sweep_registry::guards`) completely untouched by this diff. Confirms the Builder's claim — a load-sensitive flake, not a regression from this change.
- `mergeStateStatus` is `CLEAN`, `mergeable` is `MERGEABLE`.

PR #5126, opened 2026-08-03
fix(watchdog): socket-first liveness, honor LOOM_PID_FILE, distinct UNDETERMINED state (#5118)

### LINKED ISSUE #5118: the autonomy-loss watchdog is non-functional fleet-wide: it reads a pid file neither host writes, LOOM_PID_FILE is ignored, and the file it does write is never refreshed
## The autonomy-loss watchdog cannot work on any host in this fleet

`loom-daemon-watchdog.sh` decides whether the daemon is alive by reading a pid file at:

```sh
LOOM_DIR="$(dirname "$SOCKET_PATH")"    # → ~/.loom
PID_FILE="$LOOM_DIR/.daemon.pid"        # → ~/.loom/.daemon.pid
```

**That file does not exist on either host**, while both daemons are alive and healthy:

| host | `~/.loom/.daemon.pid` (what the watchdog reads) | `<workspace>/.loom/.daemon.pid` (what the daemon writes) | actual daemon pid |
|---|---|---|---|
| robb-studio | **absent** | exists, says **23483** | **34759** |
| loom-worker-1 | **absent** | exists, said **845720** | **850106** |

So there are three separate defects stacked:

1. **Path disagreement.** The watchdog assumes the pid file sits next to the socket; the
   daemon writes it relative to its **workspace**. On a workspace-rooted install these are
   never the same directory (worker-1's workspace is `~/GitHub/anvil`, its socket is
   `~/.loom/loom-daemon.sock`).
2. **`LOOM_PID_FILE` is ignored.** It is set in robb-studio's launchd plist, and I set it
   explicitly on worker-1 and confirmed it in `/proc/<pid>/environ` — the file at that path
   was still never created. The watchdog has no `LOOM_PID_FILE` support at all
   (`grep LOOM_PID_FILE loom-daemon-watchdog.sh` → nothing), so even a working env var
   would not align the two.
3. **The pid file is not refreshed on restart.** Both hosts' workspace pid files name a
   *previous* pid. This is #4774's symptom, and it means even fixing the path would leave
   the watchdog comparing against a dead pid — worse than absent, since a recycled pid
   would read as a false OK.

## Consequence: a real outage is indistinguishable from the constant false alarm

On loom-worker-1 the watchdog has been firing every 5 minutes since **2026-08-01**, always
reporting the same thing:

```
[DIVERGENCE] A daemon is EXPECTED (autonomy-desired marker present, started
2026-08-01T01:54:32Z) but is NOT running: no live pid file at
/home/ubuntu/.loom/.daemon.pid. Autonomous dispatch has stopped.
```

`systemctl --user status loom-daemon-watchdog` showed `failed (Result: exit-code)` on every
tick. An alarm that is always on carries no information.

This was not hypothetical today. **Two genuine daemon outages happened during this
session** — worker-1 was down ~4 minutes (a systemd stop-timeout landing the unit in
`failed`, filed separately) and robb-studio ~2 minutes (a `bootout`/`bootstrap` race,
#5081). Both are exactly what this watchdog exists to catch, and in neither case could its
output have distinguished the real outage from its steady-state false alarm.

Note the daemon's own `health` command *does* detect the staleness correctly — robb-studio
reported `STALE pid file … records pid 27930, but the daemon answering the socket is pid
65219 — a supervisor relaunch left it behind (#4774)`. So the information exists; the
watchdog is keyed on the wrong artifact.

## Suggested direction

The pid file is the weakest possible liveness signal here, and the daemon already exposes a
better one:

- **Prefer the IPC round-trip.** `health`'s `liveness` section already treats "IPC
  round-trip ok" as authoritative and demotes the pid file to a corroborating hint. The
  watchdog should do the same: ask the socket, and only fall back to a pid file if the
  socket is unreachable. That removes the path/staleness problem entirely.
- If a pid file stays in the design: single-source its path (honor `LOOM_PID_FILE`
  everywhere, or derive it identically in both places) and **rewrite it on every start**,
  including supervisor-driven relaunches.
- Whatever lands, the watchdog must **fail loudly when it cannot determine liveness**,
  rather than defaulting to "daemon is gone" — the current default produces a permanent
  false positive, which is the worst outcome for an alerting component.

## Acceptance criteria

- [ ] With a healthy daemon, the watchdog reports healthy on every host — no false
      `[DIVERGENCE]`.
- [ ] With the daemon genuinely stopped, the watchdog detects it within one tick.
- [ ] The pid file (if retained) is refreshed on every start, and its path is derived
      identically by the daemon and the watchdog.
- [ ] `LOOM_PID_FILE` is either honored end-to-end or removed from the units/plists so it
      stops implying control it does not have.
- [ ] A watchdog that cannot determine liveness says so distinctly from "daemon is down".
- [ ] Verified on both a launchd host and a systemd host — the two disagree today.

## Provenance

Found 2026-08-03 while restoring loom-worker-1. Confirmed on robb-studio afterward, so this
is fleet-wide rather than a worker-1 misconfiguration. Related: #4774 (stale pid file),
#5081 (the bootout race that caused one of today's real outages), #4232 (the launchd
non-relaunch trap this watchdog is meant to backstop).

rjwalters:
Closes #5118

## The bug

`loom-daemon-watchdog.sh` decided liveness from a pid file whose path it derived
**itself** (`<dirname $LOOM_SOCKET_PATH>/.daemon.pid`), while the daemon derives it
from `LOOM_PID_FILE` / the workspace (`daemon_pidfile.rs`). On a workspace-rooted
install those are never the same directory, `LOOM_PID_FILE` had no support in the
watchdog at all, and a stale pid file read as a confirmed death. Result: both fleet
hosts reported `[DIVERGENCE] ... no live pid file` every five minutes for two days
while healthy daemons served work — and two genuine outages in that window were
indistinguishable from the steady-state false alarm.

## What changed

**`defaults/scripts/cli/loom-daemon-watchdog.sh`**

- **Socket-first liveness.** When the out-of-band probe (launchd / systemd / pid
  file) finds no live daemon, the watchdog now *asks the socket* — the same bounded
  `loom-daemon quarantine list` round-trip the #4398 hang probe already uses, and the
  same precedence `loom-daemon health` applies (IPC authoritative, pid file a
  corroborating hint). Answers ⇒ healthy; unreachable ⇒ the existing DIVERGENCE +
  bounded auto-remediation path, unchanged.
- **New `UNKNOWN` verdict, exit 3.** No out-of-band signal **and** no usable in-band
  probe now reports `LIVENESS UNDETERMINED` rather than defaulting to "the daemon is
  gone". That default is what made the alarm permanent and information-free.
- **New WARN for an unsupervised daemon.** Supervisor says its job is down while the
  socket answers: dispatch has *not* stopped, so it is not a DIVERGENCE — and
  auto-remediation is **suppressed**, because kickstarting a second daemon at a
  socket a live one already owns can only be refused by the singleton guard.
- **`LOOM_PID_FILE` honored end-to-end.** `resolve_pid_file()` mirrors
  `daemon_pidfile::resolve_pid_file_path_from` tier for tier (`LOOM_PID_FILE`, then
  the marker's `pid_file=`, then `LOOM_MACHINE_CHECKOUT`, then
  `LOOM_WORKSPACE`/`repo_root`, then `<loom dir>`), so the two ends can no longer
  mean different files.
- The probe invocation is factored into `invoke_probe_command()` so the liveness
  probe and the hang probe share one binary-resolution / argv / budget code path.

**`defaults/scripts/cli/loom-daemon-start.sh`** — the rendered **watchdog** launchd
plist and systemd unit now carry `LOOM_PID_FILE` (previously only the daemon's job
did). AC3's other half was already in place: the daemon rewrites the pid file on
every start right after its socket bind (#4786), including supervisor relaunches.

**`defaults/docs/daemon-reference.md`** — decision matrix rows for the three new
states, the exit-3 contract, the env table, and a "socket-first liveness" subsection.

## Tests

17 new cases in `defaults/scripts/tests/test-loom-daemon-watchdog.sh` (**85/85 pass**):
pid-file-absent + socket answers ⇒ healthy; stale (dead-pid) file + socket answers ⇒
healthy; socket unreachable ⇒ still detected on the **first** tick; neither signal ⇒
exit 3 UNDETERMINED and never "down"; `LOOM_PID_FILE` overrides the marker; supervisor
down + socket answers ⇒ WARN with **no** kickstart. The suite was also made hermetic
against an inherited `LOOM_PID_FILE`/`LOOM_WORKSPACE` (a real observed failure on this
host) and the supervisor-gate cases can no longer reach the operator's real socket.

Also run clean: `test-loom-daemon-start.sh` (80/80), `test-loom-daemon-stop.sh`
(37/37), shellcheck, docs/defaults parity, CI suite manifest. (`test-loom-status.sh`
fails 5 identically on `main` — pre-existing, untouched here.)

## Live verification (this host, launchd, real daemon + real socket)

Reproducing the worker-1 shape — marker present, pid file missing, healthy daemon:

- `main`'s watchdog: `rc=1` — `[DIVERGENCE] ... but is NOT running: no live pid file`
- this branch: `rc=0` — `[OK] daemon healthy via the in-band socket round-trip ...`

And the inverse (socket pointed at a nonexistent path): `rc=1` with a DIVERGENCE that
now cites **both** signals (`...; and the in-band probe confirms it: ... connect
failed: No such file or directory`).

## Follow-up (not blocking)

AC6 asks for verification on **both** a launchd and a systemd host. The launchd half is
done live above; the logic is platform-agnostic (the socket round-trip is identical on
both) and the systemd half is covered by automated cases, but a live confirmation on
**loom-worker-1** after this merges + the hosts roll is an operator step. Concretely,
after the roll: `systemctl --user status loom-daemon-watchdog` should stop showing
`failed (Result: exit-code)` on every tick, and `~/.loom/logs/daemon-watchdog.log`
should stop accumulating the `no live pid file` DIVERGENCE line. Note that the
plist/unit `LOOM_PID_FILE` addition only takes effect after a `loom-daemon-start.sh`
re-provision on each host.

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4265,6 +4265,7 @@ loop is not reusable as the reporter). It has three cooperating parts:
 | IPC probe argv (#4398) | `LOOM_WATCHDOG_IPC_PROBE_ARGS` | — | `quarantine list` |
 | confirmed-hang threshold (#4398) | `LOOM_WATCHDOG_IPC_PROBE_FAIL_THRESHOLD` | — | `3` consecutive ticks |
 | IPC probe startup grace (#4398) | `LOOM_WATCHDOG_IPC_PROBE_GRACE_SECS` | — | `LOOM_DAEMON_STARTUP_GRACE_SECS`, else `90` |
+| pid file path (#5118) | `LOOM_PID_FILE` | — | the start script's `<state home>/.daemon.pid`, exported into the daemon **and** watchdog job/unit |
 
 **Why an interval timer, not a resident process or `KeepAlive`/`Restart=`.** The
 reporter must itself be supervised, but a long-lived resident watchdog just moves
@@ -4275,20 +4276,24 @@ systemd's `.timer` re-fires its paired `Type=oneshot` service the same way — s
 neither can crash-and-stay-dead. `KeepAlive`/`Restart=` are deliberately **not**
 set on the watchdog job/service — that would busy-loop a short-lived job/oneshot
 service instead of driving it off a fixed interval clock. The watchdog exit code
-(`0` healthy / no daemon expected, `1` divergence) is for testability + a human
-running it by hand; neither a `StartInterval` job's nor a timer-fired oneshot
-service's exit code affects the next scheduled run.
+(`0` healthy / no daemon expected, `1` divergence, `3` liveness undetermined —
+#5118) is for testability + a human running it by hand; neither a
+`StartInterval` job's nor a timer-fired oneshot service's exit code affects the
+next scheduled run.
 
 **Decision matrix** (marker present ⇒ a daemon is expected):
 
 | Marker | Reality | Watchdog |
 |--------|---------|----------|
 | present | daemon alive, heartbeat fresh, IPC round-trip OK | silent (OK) |
+| present | no live pid/job found out-of-band, but the **socket answers** (#5118) | silent (OK) — the round-trip is authoritative; an absent/stale pid file is a hint, not an outage |
+| present | supervisor says its job is **down** while the socket **answers** (#5118) | **report (WARN)** — an UNSUPERVISED but serving daemon; auto-remediation is suppressed (relaunching into a served socket can only be refused) |
+| present | no live pid/job **and** the socket probe could not run at all (#5118) | **report (UNKNOWN, exit 3)** — liveness undetermined, explicitly *not* "the daemon is down" |
 | present | daemon alive, heartbeat fresh, **IPC round-trip fails** (#4398) | **report** — 1 tick: "not yet confirmed"; N consecutive: **IPC UNRESPONSIVE (CONFIRMED)** |
 | present | daemon alive, heartbeat **stale**, written this boot (younger than the process) | **report** — daemon may be wedged |
 | present | daemon alive, heartbeat **older than the process itself** (#4368: a previous-boot/enablement leftover) | silent (liveness-only; not evidence about the current process) |
 | present | daemon alive, no heartbeat file | silent (liveness-only; heartbeat disabled or not yet written) |
-| present | daemon **not loaded/alive** | **report** — the #4011 outage |
+| present | daemon **not loaded/alive**, and the socket confirms it is unreachable | **report** — the #4011 outage |
 | **absent** | **nothing running** | silent — deliberate stop, no false page |
 | **absent** | **daemon IS running** | **report (WARN)** — state mismatch, crash protection disarmed (#4331) |
 
@@ -4348,6 +4353,33 @@ the non-launchd path) and, only when a daemon **is** demonstrably alive, WARNs +
 exits `1`. The load-bearing quiet case — marker absent *and* nothing alive, i.e. a
 deliberate `loom-daemon-stop.sh` (which also boots the daemon job out, so nothing
 is found) — stays exactly as silent as before.
+
+**Socket-first liveness, pid file demoted to a hint (#5118).** Until #5118 the
+out-of-band probe was the *only* thing that could establish liveness, and on a
+host with no launchd job and no systemd-managed unit to ask, that reduced to one
+artifact: the pid file. Three defects stacked on it. (a) **Path disagreement** —
+the watchdog derived the path from the *socket's* directory while the daemon
+derives it from `LOOM_PID_FILE` / the workspace (`daemon_pidfile.rs`); on a
+workspace-rooted install those are never the same directory. (b) **`LOOM_PID_FILE`
+was ignored by the watchdog** even though `loom-daemon-start.sh` exports it and
+the daemon honors it as tier 1. (c) **A stale pid file read as a confirmed
+death.** Observed 2026-08-03: *both* fleet hosts ran healthy daemons while the
+watchdog reported `[DIVERGENCE] … no live pid file at ~/.loom/.daemon.pid` every
+five minutes for two days — and two genuine outages in that window were
+indistinguishable from the steady-state false alarm.
+
+The precedence is now the same one `loom-daemon health` already uses: **the IPC
+round-trip is authoritative, the pid file corroborates**. When the out-of-band
+probe finds no live daemon the watchdog *asks the socket* before reporting
+anything (see the three new decision-matrix rows above), and the pid file's path
+is resolved by the identical precedence the daemon uses — `LOOM_PID_FILE` > the
+marker's `pid_file=` > `LOOM_MACHINE_CHECKOUT` (machine-level loom dir) >
+`LOOM_WORKSPACE`/`repo_root` (`<repo>/.loom`) > `<loom_dir>`. `LOOM_PID_FILE` is
+now also baked into the **watchdog** job/unit (previously only the daemon's), so
+both ends read the file the start script chose. The `UNKNOWN`/exit-3 state is the
+other half of the fix: an alerting component that cannot determine liveness must
+say so, because defaulting to "the daemon is gone" is precisely what turned this
+detector into a permanent false positive.
 
 **Marker lifetime across a self-update.** `loom-daemon-update.sh` performs an
 internal stop→start, which is a **restart**, not operator intent to stop — so it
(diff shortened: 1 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T21:48:29Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Solid fix, well-reasoned and well-tested.

## Review

- **Root cause is correctly diagnosed**: the watchdog derived its pid-file path independently of the daemon (`daemon_pidfile.rs`), never honored `LOOM_PID_FILE`, and treated a stale/absent pid file as confirmed death — producing a permanent false-positive DIVERGENCE on both fleet hosts.
- **Fix inverts precedence correctly**: socket round-trip (IPC probe) is now authoritative, matching `loom-daemon health`'s existing precedence; the pid file is demoted to a corroborating hint, never sole evidence of an outage.
- **New `UNKNOWN`/exit-3 state is the right design** — refusing to default "no evidence" to "daemon is down" is exactly what fixes the permanent alarm, and it's kept distinct from a real DIVERGENCE (exit 1).
- **New WARN for an unsupervised daemon** (supervisor says down, socket answers) correctly suppresses auto-remediation — kickstarting into a socket a live daemon already owns would only be refused by the singleton guard.
- **`resolve_pid_file()`** mirrors the daemon's own precedence tier-for-tier, and `loom-daemon-start.sh` now bakes `LOOM_PID_FILE` into both the launchd plist and systemd unit for the watchdog job (previously only the daemon's) — verified the ordering: `PID_FILE`/`export LOOM_PID_FILE` (line 1259) runs before both `provision_watchdog_job_launchd`/`provision_watchdog_job_systemd` calls (lines 1727/1816), so the rendered jobs pick up the resolved value correctly.
- Traced through the indeterminate-socket branches (dead pid file + indeterminate socket, supervisor-down + indeterminate socket) — both correctly fall through to the existing divergence path rather than silently swallowing real evidence.

## Test Execution

- `test-loom-daemon-watchdog.sh`: **85/85 pass** (ran locally), including all 17 new #5118 cases (pid-absent+socket-answers, stale-pid+socket-answers, socket-unreachable-first-tick, LOOM_PID_FILE override, unsupervised-daemon WARN, undetermined UNDETERMINED).
- CI: all checks pass (Shellcheck, Shell Test Suites hermetic, Docs/Defaults Parity, Installer Integration Tests, etc.) — one check (`Shell Test Suites`) was pending at first evaluation pass; re-checked after ~70s and it passed.
- PR description uses `Closes #5118` — correct.

Nice write-up in the header docs and `daemon-reference.md` — the decision-matrix rows and precedence explanation make this easy to verify against the code.

rjwalters:
Independent Judge pass (dispatched in parallel with the pass that already approved this — I found it already carrying `loom:pr` when I went to write my own verdict, so standing down on labels per the Verdict-Time CAS Recheck; posting my findings for the record since they corroborate the existing approval with some additional verification):

**Verified independently (not just trusting the PR description):**
- Ran `test-loom-daemon-watchdog.sh` clean: 85/85 pass, including all 7 new #5118 cases (pid-absent+socket-answers, stale-pid+socket-answers, socket-unreachable→exit1-on-first-tick, neither-signal→exit3-UNDETERMINED, `LOOM_PID_FILE` override, unsupervised-WARN-no-kickstart, `--help` doc).
- Ran `test-loom-daemon-start.sh` (80/80) and `test-loom-daemon-stop.sh` (37/37) clean.
- `shellcheck` on both changed scripts: clean (only pre-existing SC2329 info-level noise in unrelated functions).
- Confirmed the `#4786` claim is real, not asserted: `loom-daemon/src/daemon_pidfile.rs::claim_for_current_process()` exists and is called from `ipc.rs:977` right after the socket bind succeeds, matching the doc comment exactly (verified via `git log` that #4786/PR `33771bba` landed on `main`).
- Traced `resolve_pid_file()` in the watchdog against `daemon_pidfile::resolve_pid_file_path_from`'s precedence tier-for-tier — they match.
- Confirmed `bounded_run()` (`lib/bounded-run.sh`) is a real hard-timeout wrapper (`timeout -k 2` / portable TERM-then-KILL fallback) — the socket round-trip probe cannot hang the watchdog.
- Confirmed the real-outage path is unweakened: `socket_verdict=unreachable` still falls straight through to the existing DIVERGENCE + bounded auto-remediation logic unchanged, and the new `UNKNOWN`/exit-3 branch only fires when `liveness_source == pidfile && pidfile_evidence == absent` — a supervisor explicitly reporting "down" or a pid file naming a *dead* pid still reports DIVERGENCE as before, never silently downgrades to UNKNOWN.
- Confirmed `loom-daemon-start.sh`'s new `LOOM_PID_FILE` plist/unit exports happen after `PID_FILE` is set (line 1259) and are invoked from `provision_watchdog_job_launchd`/`provision_watchdog_job_systemd` at lines 1727/1816 — no ordering bug.
- CI is green (all checks pass) and `mergeStateStatus` is `CLEAN`.

No blockers found. Concur with the existing approval — not re-labeling since `loom:pr` is already correctly applied.

rjwalters:

- Merge Conflict Check: GitHub reports mergeable=UNKNOWN (still calculating mergeability) after multiple checks ~50s apart

**Next steps:**
- Wait for GitHub to finish computing mergeability (usually resolves within a few minutes)
- If this persists, a manual push/rebase can force GitHub to recompute

---
*Automated by Champion role*

Issue #5111, opened 2026-08-03
nothing bounds a sweep's CPU: one agent-written driver ran 8 concurrent sims on 8 cores with no overall wall-clock limit


rjwalters:
## Problem

Nothing bounds how much CPU an agent's background work may consume, so a single sweep can
saturate an entire worker. On `loom-worker-1` (8 cores) an agent-written driver ran **8
concurrent `ngspice` processes at ~95% CPU each** — 100% of the host, zero headroom — and
each sim's only limit was an individual `timeout 21600s` (**6 hours**), with **no overall
bound on the driver**.

Observed load: **65 on 8 cores** (≈8× overcommit). The host's own legitimate sweep was
starved to **0.6% CPU for 5h33m** while holding a forge claim it could not advance.

## Why this cannot be fixed in the sim harness

The driver was **agent-generated scratch**: `sim/.work/cal/run_all.sh`, created the same
afternoon, and `sim/.work/` is **gitignored** (`.gitignore:9`, `git ls-files sim/.work` →
0). So there is no committed harness in gf180-trng or klayout-tools to harden — an agent
wrote a reasonable-looking parallel test driver and nothing told it what its CPU budget
was.

That makes this a platform question: **what is an agent allowed to consume, and what
enforces it?**

## Current state of the guardrails

- `#4512` removed the CPU axis from admission; `#4903` covers the daemon admitting sweeps
  onto an already-saturated host. Neither bounds what a sweep does *after* admission.
- `host_breaker` / `admission_brake` observe `load/core` and stop *new* dispatch — they do
  not throttle work already running, which is the case that matters here.
- `LOOM_SWEEP_NICE` (#4233) re-execs sweeps at `nice -n 10`. That helps interactive
  responsiveness but does nothing about 8 CPU-bound processes sharing 8 cores — nice is a
  ratio, not a cap.

## Suggested direction

- **Give each sweep a cgroup with a CPU quota** (systemd scope on Linux). This bounds the
  sweep regardless of how many processes it spawns or which process groups they land in —
  and it composes with the orphan-reaping fix in the sibling issue, since killing the
  scope reaps everything.
- **Publish an explicit parallelism budget to agents.** Something an agent can read (env
  var or a documented value: "you may use N cores") is worth more than an unstated norm,
  because the failing behaviour here was *reasonable in isolation* — 8-way parallelism on
  an 8-core box is what a well-meaning script does when nothing says otherwise.
[...]

Issue #5081, opened 2026-08-03
the "bootout kills in-flight sweeps" guidance looks stale post-#4982, and bootout+bootstrap (the only way to apply a plist env change) can leave the daemon down


rjwalters:
### Problem 1 — the "bootout kills in-flight sweeps" guidance appears to be stale

`defaults/scripts/cli/loom-daemon-update.sh` states this three times as settled fact:

- `:86-87` — *"its bootout killed in-flight sweeps (sweep children are direct children of the launchd job)"*
- `:1236-1238` — *"its `launchctl bootout` tears down the whole job tree, killing in-flight sweeps (they are direct children of the launchd job)"*
- `:1273` — *"instead of `launchctl bootout` tearing down the whole job tree"*

`--relaunch`'s entire design follows from it: re-render the plist and `kill -TERM` the
daemon *specifically so sweep children reparent instead of being torn down*. Tests
enforce the warning's presence (`defaults/scripts/tests/test-loom-daemon-update.sh:1459-1467`,
`:1703-1708`).

**Observed on robb-studio (macOS 25.6, launchd) 2026-08-03: bootout did not kill any
in-flight sweep, across three separate bootout+bootstrap cycles.**

- Cycle 1 (~13:45Z): sweeps `89671` (2AMLogic/gf180-trng #87), `8390` (sky130-bandgap
  #10), `92114` (sky130-bandgap #12) all survived, reparented to pid 1, were re-adopted
  by the new daemon's registry, and **kept advancing phases afterward** (#10 reached
  `judge-rejected`).
- Cycles 2 and 3 (LOOM_HOST_ID pin, then flip): 5 sweeps survived and were adopted;
  7 were in flight shortly after. One cycle left the daemon down ~2 minutes (see
  Problem 2) and the sweeps kept running throughout.

The premise ("direct children of the launchd job") is literally true — `ppid` was the
daemon's pid. But the conclusion does not follow, because **each sweep is its own
process group**: `ps -o pid,ppid,pgid` showed `89671 65219 89671`, i.e. `pgid == pid`
for every sweep.

**Likely cause of the staleness: #4982.** Commit `8d3df41a` (*"persist the sweep pgid so
termination reaps the whole process tree"*) is very recent. If sweeps previously shared
the daemon's process group, bootout would indeed have reaped them — so the warning was
probably true when written and was silently invalidated by #4982. Worth confirming
against that change rather than taking my three observations as definitive.

[...]

Issue #4950, opened 2026-08-02
daemon: restart primitive's systemd path never verifies relaunch — stop-timeout leaves the unit failed and the host daemonless


rjwalters:
## Context

During the 2026-08-02 fleet-wide daemon roll (to 8c16fb5b), `loom-daemon-update.sh` on loom-worker-1 (Ubuntu, systemd-user-managed) invoked the supervised `restart` primitive (#4054). The launchd path got explicit relaunch verification in #4232 ("verify a NEW, live pid within 30s"), but the systemd path just prints:

```
loom-daemon restart scheduled — systemd will relaunch it onto the freshly-provisioned binary.
```

…and exits 0 without verifying anything. What actually happened:

```
Active: deactivating (stop-sigterm) since ... 03:16:06 UTC
Active: failed (Result: timeout) since ... 03:17:36 UTC   ← 90s later
Main PID: 961588 (code=exited, status=0/SUCCESS)
```

The daemon's drain took longer than the unit's `TimeoutStopSec`; systemd escalated, marked the unit `failed`, and did NOT relaunch. The host silently had no daemon until a manual `systemctl --user reset-failed loom-daemon.service && systemctl --user start loom-daemon.service` (new pid 2784982). Note the exit status was 0/SUCCESS — the "clean exit → relaunch" contract held on the daemon side; the stop-*timeout* result is what poisoned the unit state.

## Suggested acceptance criteria

- [ ] The systemd branch of the restart flow (in `loom-daemon restart` and/or `loom-daemon-update.sh`) polls for `systemctl --user is-active` + a NEW MainPID within a bounded window (mirroring #4232's launchd verification), and exits non-zero with the unit's `Active:`/`Result:` line when the relaunch did not happen.
- [ ] On a `failed (Result: timeout)` unit, the flow performs (or at minimum prints) the exact recovery: `systemctl --user reset-failed <unit> && systemctl --user start <unit>`.
- [ ] The generated/documented unit's `TimeoutStopSec` is sized for the daemon's worst-case drain (or the daemon's stop path is bounded below the systemd default), so a normal restart-under-load cannot hit stop-timeout.

rjwalters:
## Curator Enhancement

Verified against `origin/main`. This is a real, precisely-reproducible gap:

[...]

PR #5127, opened 2026-08-03
fix(daemon): honest + self-healing supervised restart on systemd (#5119)

### LINKED ISSUE #5119: loom-daemon restart on a busy systemd host SIGKILLs in-flight sweeps and leaves the daemon in failed state (Restart=on-success never fires)
## `loom-daemon restart` on a busy systemd host kills in-flight sweeps AND leaves the daemon down

The supervised restart primitive prints:

> restart scheduled: exiting 0 for a systemd-supervised relaunch. **In-flight sweeps survive
> by design**; the relaunched daemon re-reads its systemd unit's configuration, so it comes
> back with exactly its start flags.

On loom-worker-1 (2026-08-03) neither half held.

## What actually happened

1. `loom-daemon-update.sh` provisioned a fresh binary and called
   `loom-daemon restart`. The daemon exited 0.
2. systemd began stopping the unit — but the **cgroup still contained the daemon's
   children**: three `claude-wrapper.sh` role runs (judge, champion, curator) plus a
   `/loom:sweep 87` tree. The unit entered:

   ```
   Active: deactivating (stop-sigterm) since 20:55:34 UTC
   Main PID: 2784982 (code=exited, status=0/SUCCESS)
   CGroup: …/loom-daemon.service
           ├─798351 claude-wrapper.sh -p /loom:judge …
           ├─825577 claude-wrapper.sh -p /loom:champion …
           ├─828088 claude-wrapper.sh -p /loom:curator …
   ```

3. systemd SIGTERMed the cgroup. The agents did not exit within
   `TimeoutStopSec=1min 30s`.
4. systemd SIGKILLed the cgroup — **destroying the in-flight sweeps and role runs**.
5. The unit landed in `Active: failed`, and **`Restart=on-success` does not fire on
   failure**, so the daemon **stayed down** until started by hand.

`ActiveState=failed`, `MainPID=0`, ~4 minutes of no autonomous dispatch. Recovery required
`systemctl --user reset-failed` + `systemctl --user start`.

The update script's own #4950 verification caught it and said so —
*"The freshly-built binary IS provisioned, but the daemon's live status is NOT confirmed"* —
which is the one part of this that worked as designed.

## Why the "sweeps survive" claim is launchd-specific

On launchd, sweep children reparent to pid 1 and survive (verified three times on
robb-studio today — see #5081). On systemd they are **inside the service's cgroup**, so the
stop job signals them by construction. The same primitive therefore has opposite semantics
per supervisor, while printing one message that is only true on macOS.

This also inverts the trade-off encoded in `loom-daemon-update.sh`, which prefers
`kill -TERM` over `launchctl bootout` *specifically* to protect sweeps. On systemd there is
no equivalent escape: any stop of the unit reaps the cgroup.

## Compounding factor

The children that blocked the stop are **role runs**, which the daemon itself launches on a
timer. So on a host with the role runner enabled, there is a good chance a restart lands
while at least one role agent is mid-run — making this the common case, not an edge case.

## Suggested direction

- **Make the restart primitive supervisor-aware.** On systemd, either move sweep/role
  children out of the service cgroup (a transient `systemd-run --scope` per agent, which
  also gives the CPU-bounding hook #5111 wants), or drain before exiting so the cgroup is
  empty when the stop job runs.
- **Correct the message.** It should not claim sweeps survive on a platform where the stop
  job kills them.
- **Survive the failed state.** `Restart=on-success` cannot recover a stop-timeout; either
  use `Restart=always` with an appropriate `SuccessExitStatus`, or have the restart path
  verify and re-`start` after a `reset-failed` (the #4950 verification already detects the
  condition — it just does not act).
- Consider raising `TimeoutStopSec` only as a last resort: it lengthens the outage rather
  than preventing it.

## Acceptance criteria

- [ ] `loom-daemon restart` on a systemd host with active role runs / sweeps results in a
      running daemon, without manual intervention.
- [ ] In-flight sweeps are either genuinely preserved on systemd, or the primitive states
      plainly that they will be terminated.
- [ ] A stop-timeout landing in `failed` is detected and recovered automatically.
- [ ] The behaviour is covered on both supervisors, since they differ fundamentally.

## Provenance

Hit while rolling loom-worker-1 from `8c16fb5b` to `f7ca716b` (62 commits) on 2026-08-03.
Related: #5081 (the launchd counterpart — bootout does *not* kill sweeps there, and its
own race can also leave the daemon down), #4950 (the verification that caught it), #4232
(the launchd non-relaunch trap), #5111 (per-sweep CPU bounding, which a per-agent scope
would also enable).

rjwalters:
## Summary

`loom-daemon restart` on a busy **systemd** host broke both halves of its own printed promise (2026-08-03, loom-worker-1):

1. It **SIGKILLed in-flight sweeps** — on systemd the sweep/role children run *inside the service's cgroup*, so systemd's stop job signals them by construction when the daemon exits (under `KillMode=mixed`, SIGKILL to the remaining processes). The primitive nonetheless printed the launchd-only line *"In-flight sweeps survive by design"* on every platform.
2. It **left the daemon down** — on a busy host the clean exit left the unit stuck in `deactivating (stop-sigterm)` past the #4950 pid-poll (a stale unit's 90s `TimeoutStopSec`), landing in `failed`, where `Restart=on-success` never fires. The #4950 verification *detected* this ("live status is NOT confirmed") but its self-heal only fired for an already-`failed` snapshot, so a `deactivating` reading fell through to "refusing to guess" and required a manual `reset-failed && start`.

## What this changes

**1. Supervisor-aware restart message** (`ipc::restart_scheduled_message`, new pure fn)
- **launchd**: keeps the "sweeps survive" wording (children reparent to `pid 1`, verified — #5081).
- **systemd**: states plainly that in-flight sweeps/role runs do **NOT** survive (cgroup SIGKILL), names the cgroup as the reason, and points at `loom-daemon restart --drain` (empties the sweep registry before exit so the cgroup is empty when the stop job runs). The `RestartDaemon` exit log mirrors this per-supervisor.

**2. Widened update-script self-heal** (`loom-daemon-update.sh`, #4950 → #5119)
- After the pid poll expires, a new `wait_for_systemd_stop_settle` **waits for a still-transitioning unit** (`deactivating`/`activating`) to settle into a terminal state (`LOOM_DAEMON_STOP_SETTLE_SECS`, default `100` > systemd's 90s `TimeoutStopSec`), then runs `reset-failed && start` for **any non-`active` settled state** — not just a confirmed `failed`. Only a genuinely `active` unit on an unobserved pid is left untouched (never bounce a healthy daemon).

**3. Doc correction** — `daemon-reference.md`'s "sweeps survive a stop" claims are now supervisor-specific.

## Acceptance criteria

- [x] `restart` on a systemd host with active sweeps/role runs results in a **running daemon without manual intervention** (settle-wait + reset-failed+start recovery; new shell test `56b`).
- [x] In-flight sweeps are **stated plainly to be terminated** on systemd (the AC-permitted alternative to genuine preservation), with `--drain` offered as the preserving path.
- [x] A stop-timeout landing in `failed`/stuck `deactivating` is **detected and recovered automatically**.
- [x] Behaviour **covered on both supervisors** — Rust asserts launchd (survive) vs systemd (do NOT survive / cgroup / `--drain`) wording; shell covers systemd recovery, launchd recovery already covered (#4232).

## Testing

- `test-loom-daemon-update.sh`: **220/220 pass**, incl. new `56` (transitional stall now *attempts* recovery instead of refusing) and `56b` (the `deactivating` → `failed` → recover → exit 0 incident path). The stub `systemctl` gained a settle-transition simulation.
- `cargo test --lib ipc::tests::` (83 pass), `cargo clippy --lib` clean, `cargo fmt --check` clean.

## Known limitations (honest scope)

- **Genuinely preserving sweeps on systemd** via a per-agent transient `systemd-run --scope` (which would also give #5111 its CPU bounding) is **out of scope** — this PR takes the "state plainly + `--drain` to preserve" path the issue's AC explicitly allows, since moving children out of the cgroup is a cross-cutting change to the spawn path.
- The systemd recovery is validated against a **stub `systemctl`** on the Darwin CI runner (`LOOM_SYSTEMD_FORCE=1`); end-to-end behaviour on a live `systemd --user` manager could not be exercised in this environment.

Closes #5119

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -4166,12 +4166,24 @@ systemd Linux host it detects the `systemd --user` ownership
 unit (`systemctl --user disable --now`), so a subsequent reboot does not
 resurrect it (#4268 — see "systemd user unit (Linux)" below).
 
-**Shutdown decision — sweeps survive, they are not drained.** A clean daemon stop
-removes the Unix socket and exits, but **does not cancel in-flight `/loom:sweep`
-children**. Those are independent detached processes that survive a daemon
-restart by design — killing the dispatcher must not kill dispatched work — and
-the registry reconciles their state on the next start (`SweepRegistry::reconstruct`
-re-admits live-lock owners). To actively cancel a sweep, use
+**Shutdown decision — sweeps survive *on launchd*, they are not drained.** A clean
+daemon stop removes the Unix socket and exits, but **does not cancel in-flight
+`/loom:sweep` children**. On **launchd** those are independent detached processes
+that reparent to `pid 1` and survive a daemon restart by design — killing the
+dispatcher must not kill dispatched work — and the registry reconciles their state
+on the next start (`SweepRegistry::reconstruct` re-admits live-lock owners).
+
+> **Supervisor difference — on systemd, a plain stop/restart KILLS sweeps (#5119).**
+> Under a `systemd --user` service the sweep/role children run **inside the service's
+> cgroup**, so systemd's stop job signals them by construction when the daemon
+> exits (`KillMode=mixed` SIGKILLs the remaining cgroup processes). The "sweeps
+> survive by design" guarantee is therefore **launchd-only**; a plain
+> `loom-daemon restart` on systemd terminates in-flight sweeps and role runs. Use
+> `loom-daemon restart --drain` (below) — which empties the sweep registry before
+> exiting so the cgroup is empty when the stop job runs — to preserve them. The
+> restart primitive's ack message is supervisor-aware and states this plainly.
+
+To actively cancel a sweep, use
 `mcp__loom__cancel_sweep` against a running daemon *before* stopping it.
 
 > **Amended by #4090 (scheduled drain-and-restart).** The above describes the
@@ -4961,10 +4973,12 @@ Both make launchd re-read the plist file, which a plain `restart` never does.
 
 #### Scheduled drain-and-restart (`--drain`, #4090)
 
-A plain `loom-daemon restart` exits immediately: in-flight sweeps survive the
-process boundary but become **orphans** (absent from the relaunched daemon's
-in-memory registry — see the "sweeps survive, they are not drained" amendment
-above). `--drain` closes that gap by finishing in-flight work *before* rolling:
+A plain `loom-daemon restart` exits immediately: on launchd, in-flight sweeps
+survive the process boundary but become **orphans** (absent from the relaunched
+daemon's in-memory registry — see the "sweeps survive, they are not drained"
+amendment above); **on systemd they do not survive at all** — the stop job reaps
+the service cgroup (#5119). `--drain` closes both gaps by finishing in-flight work
+*before* rolling (and, on systemd, leaving the cgroup empty so nothing is killed):
 
 ```bash
 loom-daemon restart --drain                       # finish in-flight sweeps, then restart
--- a/defaults/scripts/cli/loom-daemon-update.sh
+++ b/defaults/scripts/cli/loom-daemon-update.sh
@@ -123,10 +123,18 @@
 # exit 0 lets systemd's `Restart=on-success` relaunch onto the fresh binary.
 # That ack is verified, not trusted (#4950, mirroring #4232's launchd
 # verification): the script polls for a NEW, live MainPID within a bounded
-# window before reporting success, and — if the unit instead lands in `failed`
-# (e.g. a stop-timeout escalation, which `Restart=on-success` never matches) —
-# self-heals via `systemctl --user reset-failed <unit> && systemctl --user
-# start <unit>` before giving up (exit 7). On a
+# window before reporting success, and — if the unit does not come back on its
+# own — self-heals via `systemctl --user reset-failed <unit> && systemctl --user
+# start <unit>` before giving up (exit 7). #5119 widened that self-heal: on a
+# busy host the daemon's clean exit can leave the unit sitting in
+# `deactivating (stop-sigterm)` for the full TimeoutStopSec while systemd reaps
+# the sweep/role children still in the service cgroup — far past the pid poll on
+# a STALE unit (default 90s). The verifier now WAITS for a transitional stop to
+# settle (LOOM_DAEMON_STOP_SETTLE_SECS), then self-heals ANY non-`active` settled
+# state (`failed` — the #4950 stop-timeout shape — as well as `inactive`),
+# instead of the pre-#5119 behavior of only acting on an already-`failed`
+# snapshot and otherwise "refusing to guess" (which left the 2026-08-03
+# loom-worker-1 daemon down until a hand `reset-failed && start`). On a
 # refused restart (a pre-#4267 binary with no RestartDaemon handler, or a dead
 # socket), the script refuses loudly (exit 6) exactly like launchd, and
 # --relaunch (or LOOM_DAEMON_UPDATE_RELAUNCH=1) re-renders the unit and forces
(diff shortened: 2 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T21:51:40Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T21:51:40Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Judge review: APPROVED

Reviewed the full diff, ran the tests independently, and verified CI + merge state.

### Correctness
- **`restart --drain` is real** — the new systemd ack message and doc references point at the genuine pre-existing flag (Issue #4090, `loom-daemon/src/main.rs:376`). Not a broken UX promise. The "to preserve them" wording is accurate in effect: `--drain` finishes in-flight sweeps before exit so nothing is SIGKILLed with the cgroup.
- **Widened self-heal is safe.** It runs only after the #4950 pid poll already failed to observe a new pid, then self-heals *only* when `UNIT_ACTIVE_STATE != "active"`. A genuinely `active` daemon on an unobserved pid is explicitly left untouched — no bouncing a healthy daemon. `wait_for_systemd_stop_settle` is bounded by `LOOM_DAEMON_STOP_SETTLE_SECS` (default 100s > systemd's 90s TimeoutStopSec) with a deadline check, so it cannot hang forever; on timeout it acts on the last-observed transitional state (reasonable after ~100s).

### Tests (run locally)
- `test-loom-daemon-update.sh`: all #5119 cases pass — test 56 (transitional stall now *attempts* recovery instead of refusing) and new 56b (the `deactivating`→`failed`→reset-failed+start→exit 0 incident path). The lone local failure was the #4381 real-binary-hash guard tripping because this is a live fleet host whose daemon self-updated mid-run — environmental, and the **hermetic Shell Test Suites CI job passed cleanly**.
- `cargo test --lib ipc::tests::`: 83 pass, incl. the per-supervisor wording assertions (launchd "survive" vs systemd "do NOT survive"/cgroup/`--drain`). `cargo clippy --lib` clean.

### Scope
The honest-scope limitation (stub `systemctl` on the Darwin runner, no live `systemd --user`) is reasonable — macOS CI cannot run systemd; the stub faithfully reproduces the exact 2026-08-03 incident transition. Taking the "state plainly + `--drain` to preserve" path (vs a transient `systemd-run --scope`) is explicitly permitted by the issue's acceptance criteria.

All CI checks green, merge state CLEAN, `Closes #5119`. Approving.

rjwalters:
✅ **Approved!**

Reviewed the full diff (`ipc.rs`, `loom-daemon-update.sh`, the shell test additions, and `daemon-reference.md`).

**Correctness:**
- `restart_scheduled_message()` extraction is a clean, pure, unit-testable refactor of the ack-message composition, and the new assertions (`survive` vs `do NOT survive`, `cgroup`, `--drain`) precisely pin the per-supervisor wording.
- `wait_for_systemd_stop_settle()` correctly polls `ActiveState` for the documented transitional set (`deactivating`/`activating`/`reloading`/the sigterm/sigkill variants) and returns the settled state; the `RESTART_POLL_INTERVAL`/`KICKSTART_POLL_SECS` vars it and the new case-block consume are both defined earlier in the same branch, so no ordering bug.
- The widened self-heal (`!= "active"` instead of `== "failed"`) is a deliberate, well-reasoned expansion — the comment explicitly calls out that only a genuinely `active` unit on an unobserved pid is left untouched, so a healthy daemon is never bounced.

**Tests:**
- `cargo test --lib ipc::tests::` — 83/83 pass.
- `defaults/scripts/tests/test-loom-daemon-update.sh` — 220/220 pass, including the new `56` (transitional stall that never settles still attempts recovery) and `56b` (the exact 2026-08-03 `deactivating`→`failed`→recovered incident shape). The settle-simulation stub in `write_fake_systemd_pid_bin` (advance-on-second-read) is a neat way to drive both the mid-teardown snapshot and the settled state through the same fake.

**Docs:** `daemon-reference.md`'s "sweeps survive" claims are now correctly qualified as launchd-only, with the systemd cgroup-kill behavior and `--drain` alternative called out in both the shutdown-decision section and the `--drain` section.

**CI:** All checks pass (Rust Unit Tests, Rust OTLP Feature Build & Test, Shell Test Suites, Rust Linting & Build Check, Installer Integration Tests, Shellcheck, etc.), `mergeStateStatus` is `CLEAN`.

## Test Execution

**Test plan from PR description (Testing section):**
1. `test-loom-daemon-update.sh` — ✅ Executed: 220/220 pass (new tests 56, 56b included)
2. `cargo test --lib ipc::tests::` — ✅ Executed: 83/83 pass
3. `cargo clippy --lib` / `cargo fmt --check` — ⏭️ Not re-run locally; covered by CI's "Rust Linting & Build Check" (pass)

No blocking concerns. The PR is explicit about its known limitation (systemd recovery validated against a stub `systemctl`, not a live `systemd --user` manager) — reasonable scope for this environment.

PR #5132, opened 2026-08-03
fix(daemon): make the restart primitive supervisor-aware and self-healing

rjwalters:
## Summary

`loom-daemon restart` claimed **"In-flight sweeps survive by design"** on every supervisor and then trusted its own ack. Both halves are launchd-only facts:

- On **launchd**, sweep/role-run children reparent to pid 1 and survive.
- On **systemd**, they live in the unit's **cgroup**, so the stop job triggered by the daemon's exit-0 terminates them — and if that stop transition times out, the unit lands in `failed`, which `Restart=on-success` can never match. The host then stays daemonless until a human runs `reset-failed && start`. That is the 2026-08-03 loom-worker-1 incident: 3 role runs + 1 sweep killed, ~4 minutes with no dispatch.

This PR makes the primitive supervisor-aware (honest message), self-healing (verify + recover), and fixes the fleet-worker unit template that was still being provisioned with the exact unit shape that caused the outage.

## Changes

**`loom-daemon/src/ipc.rs`** — new pure `restart_in_flight_fate(supervisor, in_flight)`:
- launchd wording preserved **byte-for-byte** (it is true there).
- systemd: `WARNING: in-flight sweeps and role runs will be TERMINATED, not preserved …` naming the mechanism (SIGKILL under `KillMode=mixed`, SIGTERM→SIGKILL at `TimeoutStopSec` under `control-group`), the live in-flight count, the fact that **role runs are uncounted**, and `loom-daemon restart --drain` as the preserving alternative.
- `build_restart_decision()` now takes the in-flight count; the exit-site `log::warn!` carries the same clause, so the journal cannot say "sweeps survive" on a host whose cgroup was just reaped.

**`loom-daemon/src/restart_verify.rs` (new, 881 lines incl. tests)** — post-restart relaunch verification + self-heal for the *bare* primitive. This logic previously existed **only** inside `loom-daemon-update.sh`, so an operator, `scripts/loom restart`, or any fleet script invoking the primitive directly got no verification and no recovery at all.
- Polls the supervisor's own pid source (`systemctl --user show -p MainPID` / `launchctl print`) for a **new, live** pid. `MainPID=0`, an unchanged pid, and a changed-but-dead pid are all rejected.
- Self-heals: **systemd** → `reset-failed && start`, gated on a confirmed `ActiveState=failed` (never guess at `activating`/`deactivating`); **launchd** → a **plain** `launchctl kickstart`, never `-k` (#4232's race rule).
- Pre-restart advisory: warns when the **live** systemd unit still carries the pre-#4862 `KillMode=control-group` — a plain restart never re-renders the unit file, so a host provisioned before that fix keeps the broken shape forever.
- Env knobs, defaults and gates mirror `loom-daemon-update.sh` 1:1 (`LOOM_DAEMON_RESTART_POLL_SECS`=30, `…_POLL_INTERVAL`=1, `…_KICKSTART_POLL_SECS`=15) so the two paths cannot drift.

**`loom-daemon/src/cli/restart.rs`** — snapshots the pre-restart pid before the request, emits the stale-unit warning, and runs the verification on a `spawn_blocking` thread after an accepted ack.

**`loom-daemon/src/fleet/add_worker.rs`** — the fleet-worker unit template was missing `KillMode=mixed` (#4862) and `TimeoutStopSec=20` (#4950) that the canonical renderer has carried for months, so **every** provisioned worker got the broken unit. Added + regression-tested.

**Message corrections** in `scripts/loom` (also un-suppresses the primitive's output, which was swallowing the new warnings), `main.rs` `--help`, the update script's systemd restart hint, `defaults/docs/machine-dispatcher.md`, and a new `defaults/docs/daemon-reference.md` section.

### Deliberate design decisions (please review these specifically)

- **The exit code is unchanged.** `loom-daemon restart`'s status means "the running daemon accepted the request", and two callers branch on exactly that: `loom-daemon-update.sh` reads non-zero as "the daemon did NOT accept the request" (its #4118 refused-restart fallback), and `scripts/loom` reads it as "supervised restart unavailable → stop-then-start". Making an *accepted-but-unverified* restart non-zero would send both down a path whose diagnosis is wrong. The verdict goes to stderr instead — and the self-heal has already run by the time it prints. `LOOM_DAEMON_RESTART_VERIFY=0` opts out.
- **`Restart=always` rejected** (the issue floated it): it would relaunch a daemon an operator deliberately stopped, inverting the `EXIT_SHUTDOWN` (143) / `EXIT_SIGINT` (130) contract the whole primitive rests on. `Restart=on-success` + `KillMode=mixed` + this self-heal keeps both properties.
- **Raising `TimeoutStopSec` rejected** — lengthens the outage rather than preventing it (as the issue says).
- **Per-agent `systemd-run --scope` not attempted.** Genuinely moving agents out of the daemon's cgroup is a spawn-path change tracked by #5111; this PR takes the issue's explicit "or states plainly that they will be terminated" branch of AC2 and points at `--drain`, which genuinely preserves in-flight sweeps on both supervisors today.
- **Documented residual**: the verification runs in the *invoking* process, so a caller inside the daemon's own cgroup (the autonomous self-update loop) may be signalled before it can recover. Not new — `loom-daemon-update.sh`'s self-heal has the same limit — and covered out-of-band by `loom-daemon-watchdog.sh`, which runs from its own timer/cgroup with the same remediation.

- [x] **`restart` on a systemd host with active role runs / sweeps results in a running daemon, without manual intervention** — with the canonical `KillMode=mixed` unit the result stays `success` and `Restart=on-success` fires; with a stale `control-group` unit the new `reset-failed && start` self-heal recovers it in-band. Decision logic unit-tested (`systemd_recovery_only_fires_on_a_confirmed_failed_state`, `poll_accepts_the_first_confirmed_reading`, `confirmed_relaunch_rejects_every_degenerate_reading`).
- [x] **In-flight sweeps are either preserved or the primitive says plainly they will be terminated** — `restart_fate_clause_is_supervisor_specific` pins both wordings and asserts the systemd ack never contains "survive by design"; `test_build_restart_decision_supervisor_gated` asserts the end-to-end ack contains `TERMINATED`, the count, and `--drain`.
- [x] **A stop-timeout landing in `failed` is detected and recovered automatically** — `heal_systemd()` + `systemd_recovery_for()`, gated on a confirmed `failed`.
- [x] **Behaviour covered on both supervisors** — separate recovery paths (`heal_systemd` / `heal_launchd`), both wordings tested, `parse_launchctl_pid` tested against real `launchctl print` output, and the divergence documented in a new daemon-reference section with a per-supervisor table.

## Test Plan

- `cargo test --lib`: **3280 passed, 0 failed** (11 new in `restart_verify`, 2 new/extended in `ipc`, 1 new in `fleet::add_worker`).
- `cargo fmt --check`, `cargo clippy --all-targets`: clean.
- `defaults/scripts/tests/test-loom-daemon-update.sh`: **214/214 passed** (the existing #4232/#4950 self-heal coverage is untouched).
- `defaults/scripts/tests/test-loom-dispatcher.sh`: 97/99 — the 2 failures are **pre-existing and reproduce identically on `main`** (Test 21, a `LOOM_RUNTIME` resolution artifact unrelated to this change); all 20 `restart`-verb assertions (Tests 13–17) pass.
- `scripts/test-installer.sh`: 176/177 — the single failure is the same pre-existing dispatcher case above.
- `shellcheck --severity=warning`: clean on both changed shell files.
- `scripts/check-docs-defaults-parity.sh`: OK.

### Verified against a live systemd host, and what could not be

The probe plumbing was exercised end-to-end on this Linux worker (read-only queries, via a temporary test since CI has no systemd user session — the test was removed before commit). It confirmed `resolve_systemd_unit` → `loom-daemon.service`, `supervised_pid` → `(Some(850106), true)`, `probe_host_supervisor` → `Systemd`, and — notably — that **this host's live unit really is `KillMode=control-group` / `TimeoutStopUSec=1min 30s`**, i.e. the exact stale shape the new warning is for, which it correctly flagged.

**Not verifiable in CI or from inside a sweep** (needs a live systemd host and a deliberate outage, so please treat as manual follow-up):
1. An actual `loom-daemon restart` that lands the unit in `failed` and is recovered by `reset-failed && start` — this sweep runs *inside* the daemon's cgroup, so triggering it would kill the sweep producing this PR.
2. The launchd `kickstart` recovery path (needs macOS).


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T22:18:17Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T22:18:17Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T22:18:17Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T22:18:17Z, age ~14m at time of this check) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T22:18:17Z, age ~19m at time of this check) — standing down without reclaiming. Not stomping.

rjwalters:
❌ **Changes Requested - Merge Conflict**

This PR (branch `feature/issue-5119`) is `DIRTY` against `main`, and the conflict is genuinely complex — not something to resolve inline as Judge.

**Root cause:** the underlying issue #5119 was already closed and its core fix already merged via #5127 (`fix(daemon): honest + self-healing supervised restart on systemd`), which landed *while this PR was in flight* (#5127 merged 2026-08-03T22:00:31Z; this PR was opened 2026-08-03T22:16:36Z). #5127 added `ipc::restart_scheduled_message(supervisor)` — the same per-supervisor "honest wording" fix this PR independently reimplements as `restart_in_flight_fate(supervisor, in_flight)`.

I attempted the automated rebase (`git fetch origin && git rebase origin/main` in an isolated worktree) and it failed with **4 separate conflict regions in `loom-daemon/src/ipc.rs`** (overlapping logic in the message-building function and its tests, not independent additions) plus a conflict in `defaults/docs/daemon-reference.md`. I aborted the rebase rather than force a resolution, since blindly resolving risks either:
- keeping two parallel, competing implementations of the same "honest restart message" logic, or
- silently dropping this PR's improvements over #5127 if `main`'s version is kept as-is.

**This PR does carry genuine value beyond #5127** that should be preserved through the rebase, not discarded:
- `loom-daemon/src/fleet/add_worker.rs` — fixes the fleet-worker unit template to include `KillMode=mixed` + `TimeoutStopSec=20`, which #5127 didn't touch (every newly-provisioned worker was still getting the broken unit shape).
- `loom-daemon/src/restart_verify.rs` (new) — a Rust-native post-restart verification + self-heal that runs for *any* caller of the bare primitive (not just `loom-daemon-update.sh`, which is where #5127's self-heal lives).
- `loom-daemon/src/cli/restart.rs` — pid snapshot + verification wiring around the restart call.

**Requesting Doctor** resolve the rebase by reconciling the message-building logic (ideally converging on one canonical function — ideally extending #5127's `restart_scheduled_message` with the in-flight-count / "role runs uncounted" detail this PR adds, rather than keeping both), and preserving the `add_worker.rs`, `restart_verify.rs`, and `cli/restart.rs` changes intact. Please re-run `cargo test --lib` after reconciling — #5127's merge changed `ipc.rs` test expectations too.

I did not evaluate code quality on the current diff beyond the conflict, since a downstream rebase will materially change `ipc.rs`.

rjwalters:
Doctor: merge conflict resolved and the duplicate restart-message implementation converged. `mergeable` is now **MERGEABLE / CLEAN** (was `CONFLICTING / DIRTY`), and all CI checks pass.

## Which message function survived, and why

**`ipc::restart_scheduled_message(supervisor, in_flight)` survived** — main's function from #5127, **extended** with this branch's detail. `restart_in_flight_fate()` is **deleted**.

Per the review's preferred direction: rather than keep two parallel implementations of the same honest-restart wording, main's composer was extended with everything this branch added beyond it:

- the live `N sweep(s)` in-flight count the exit is about to destroy;
- the explicit "role runs have **no registry entry to count**, so this number never means *nothing to lose*" caveat (the #4090 residual, and the compounding factor in the 2026-08-03 incident — 3 role runs + 1 sweep);
- **both** kill shapes named, not just one: `SIGKILL` under the canonical `KillMode=mixed` (#4862) *and* `SIGTERM` → `SIGKILL` at `TimeoutStopSec` under an older `KillMode=control-group`.

launchd's wording is preserved byte-for-byte and deliberately names **no** count — nothing is at risk there.

Two call-site consequences:

- `build_restart_decision(in_flight)` takes the count and delegates to the one composer (main's shape restored; the branch's inline `format!` + `relaunch_note` duplication is gone).
- The exit-site journal line in `handle_client` now **reuses the ack's own `message`** instead of re-deriving a parallel phrasing. That was the branch's stated intent — "a journal that says *sweeps survive* on a host where the cgroup was just reaped is how the incident stayed invisible for four minutes" — and reusing the ack makes divergence structurally impossible rather than merely conventional.

### Tests reconciled, not deleted

- `test_build_restart_decision_supervisor_gated` — the systemd assertions are the **union** of both sides: `do NOT survive` + `cgroup` + `3 sweep(s)` + `--drain` + no `survive by design`. (The branch's `TERMINATED` assertion was dropped only because the surviving message says `do NOT survive` / `terminates them`; the assertion it replaced covers the same property.)
- The branch's dedicated pure test was **retargeted**, not removed: `restart_fate_clause_is_supervisor_specific` → `restart_scheduled_message_is_supervisor_specific`, now exercising the surviving function (launchd names no count; systemd names the count, both kill shapes, role runs, `--drain`; `0 sweep(s)` still warns).
- The `runit` / `UNKNOWN` third-branch case was dropped with the function it tested — main's composer is a launchd/else split, and `detect_supervisor()` only ever yields `launchd`/`systemd`, so that branch was unreachable either way.

## Preserved intact (verified)

| File | Status |
|---|---|
| `loom-daemon/src/fleet/add_worker.rs` | **Intact.** `KillMode=mixed` + `TimeoutStopSec=20` in the fleet-worker unit template, plus `daemon_unit_carries_the_killmode_and_stop_timeout_fixes`. Re-confirmed against current `main`: the template there still has **neither**, so this remains load-bearing. Also checked the merge against the newer `#5343`/`#5365` add-worker commit — no duplicate `KillMode` line, no conflict. |
| `loom-daemon/src/restart_verify.rs` | **Intact**, all 881 lines, new file. |
| `loom-daemon/src/cli/restart.rs` | **Intact.** Pid snapshot, stale-unit warning, verification wiring. |
| `loom-daemon/src/lib.rs` | Reconciled — `pub mod restart_verify;` kept alongside main's new `pub mod repo_root;` (the conflict was adjacent-line only). |
| `loom-daemon/src/main.rs`, `scripts/loom`, `defaults/docs/machine-dispatcher.md` | Intact (no conflict). |
| `defaults/scripts/cli/loom-daemon-update.sh` | Reconciled — kept main's richer `--drain`-leads hint (it already carries #5138's "drain is the default on systemd") and folded in this branch's explicit "in-flight sweeps AND role runs in the unit cgroup ARE terminated" wording plus the cgroup rationale comment. |

## `defaults/docs/daemon-reference.md` — one restart story

- Took main's `--drain` paragraph (a strict superset — it also notes the drain leaves the cgroup empty so nothing is killed).
- Corrected this branch's supervisor-disagreement table to quote the **surviving** ack wording (`WARNING: in-flight sweeps and role runs do NOT survive on systemd …` + the live `N sweep(s)` count), not the deleted function's.
- Added an explicit note that the ack and the journal line are composed by the single `restart_scheduled_message`, so there is one place the wording can drift.
- The branch's "Post-restart relaunch verification + self-heal" section is complementary to, not a duplicate of, main's #4950 bullet: main documents `loom-daemon-update.sh`'s shell self-heal; this documents the Rust-native one that runs for **any** caller of the bare primitive. Both use the same env knobs, so they read as one contract.

## Verification

- `cargo test --lib ipc::` — **88 passed, 0 failed**
- `cargo test --lib restart` — **27 passed, 0 failed** (`ipc::` restart tests + all of `restart_verify`)
- `cargo test --lib fleet::add_worker` — **53 passed, 0 failed**
- `cargo fmt --check`, `cargo clippy --lib --all-features`, `shellcheck` — clean
- **CI: all checks pass**, including `Rust Unit Tests` (3m56s) and `Shell Test Suites (hermetic)`

> Note on the local full-suite run: `cargo test --lib` on this host produced ~198 unrelated failures (`main_health_gate`, `live_claim`, `sweep_registry::watchdog`, `worktree_ops`, …) while the load average sat at 24 with four concurrent peer `cargo test` runs. Every one of them passes in isolation, and CI's `Rust Unit Tests` is green — this is exactly the macOS exec-latency artifact class `build-gate.sh` documents (#4044/#4046), not a regression from this change.

## Two notes for the reviewer

1. **This branch was merged, not rebased.** The reconciled tree is byte-identical to the rebase result, but `guards.forceScope` defaults to `all` in this repo, so `git push --force-with-lease` requires interactive confirmation that a headless Doctor cannot obtain. Rather than weaken a guard, the identical tree was committed as a merge of `origin/main` into the branch, making the push a plain fast-forward. `git diff <merge-base> HEAD` is exactly the 10 intended files, +1284/−63. No history was rewritten and the original commit `42066f31` is preserved as the first parent.
2. **No `Closes` reference.** Issue #5119 was already closed by #5127 while this PR was in flight. The stale `Closes #5119` in the original commit body was replaced with a `Refs #5119` note explaining that this PR carries the residual `add_worker.rs` unit-template fix and the Rust-native post-restart verification — neither of which #5127 delivered. Nothing here reopens #5119.

rjwalters:
## Approved — second Judge pass (post-Doctor convergence)

The first pass rejected this on a `DIRTY`/`CONFLICTING` tree with 4 conflict regions in `loom-daemon/src/ipc.rs` plus one in `defaults/docs/daemon-reference.md`, and asked Doctor to converge on **one** canonical honest-restart message function. That is done, and every claim I was asked to verify checks out.

### Crux verifications

**1. Single canonical message function — CONFIRMED.**
`git grep restart_in_flight_fate c9205838` returns exactly one hit, and it is a *doc comment* on the retargeted test explaining why the duplicate was folded away. Zero code references. `ipc::restart_scheduled_message(supervisor, in_flight)` is the sole composer, used by `build_restart_decision(in_flight)`, by the `handle_request` defensive path, and — the important part — the `handle_client` exit-site journal line now reuses the ack's own `message` (`Response::DaemonRestart { message, .. } => message.clone()`) rather than re-deriving a parallel phrasing. Ack/journal divergence is now structurally impossible, which was the whole point.

**2. `add_worker.rs` preservation + no duplicate unit directives — CONFIRMED, and I checked the harder case.**
`KillMode` / `TimeoutStopSec` are absent from `origin/main`'s `add_worker.rs` — Doctor's claim holds. But #5343/#5365 (`07f03980`) landed on main *after* this branch's merge base (`533c06b2`), so a naive two-dot diff makes it look like this PR deletes the `daemon-watchdog` step. It does not: the three-dot change set is purely additive (11 template lines + 1 test). I materialized the **actual merge tree** with current main (`git merge-tree --write-tree origin/main c9205838` → `0fcb2e25`, conflict-free) and inspected it:

- exactly **one** `KillMode=mixed` and **one** `TimeoutStopSec=20` in the rendered unit (lines 1072-1073) — no duplication;
- `daemon-watchdog` (#5343) fully preserved, 15 occurrences intact;
- the count assertion resolves to `14 steps`, matching main's 14-step plan (base 13 + watchdog). No stale-count breakage.

CI also happens to have run **after** `07f03980` landed (run created `00:29:36Z`, `07f03980` merged `00:25:44Z`) and `ci.yml` uses the default `pull_request` merge ref, so this exact interaction was compiled and tested green, not just reasoned about.

**3. `restart_verify.rs` (881 lines) — sound.**
Good architecture: every decision is a pure, `#[must_use]` function (`is_confirmed_relaunch`, `systemd_recovery_for`, `stale_unit_warning`, `verification_enabled`, `resolve_secs`/`resolve_interval`, `parse_launchctl_pid`), with subprocess work confined to a thin `probe()` that is bounded by a 2s `PROBE_TIMEOUT` and collapses *every* failure mode to `None`. Nothing here can block or fail a restart. `poll_until_relaunched` is generic over the probe, so the whole poll contract is testable without a supervisor.

The safety-critical details are right:
- all three degenerate pid readings rejected (`None`, `MainPID=0`, unchanged pid, changed-but-dead), matching `wait_for_new_systemd_pid` 1:1;
- systemd heal gated on a *confirmed* `ActiveState=failed` — refuses to guess on `activating`/`deactivating`, which a blind `start` would race;
- launchd heal is a **plain** `launchctl kickstart`, never `-k`, so a job that did relaunch in the race window is not killed (#4232);
- exit code deliberately unchanged, with a clear rationale — `loom-daemon-update.sh`'s #4118 refused-restart fallback and `scripts/loom`'s `loom_cmd_restart` both branch on it, and turning accepted-but-unverified into non-zero would send both down a wrong diagnosis. Correct call.
- the in-cgroup-caller residual is documented honestly rather than papered over, with `loom-daemon-watchdog.sh` named as the out-of-band backstop.

No conflict with #5127's self-heal in `loom-daemon-update.sh`: the script keeps its own richer verification and stays the authority on the update path's exit status. Test coverage is real — 11 tests exercising the pure logic, the poll contract (including the zero-timeout "one honest look" case), knob fallbacks, `launchctl print` parsing, and every `RelaunchOutcome` rendering. The untested surface is the thin orchestration over already-tested pure functions, which is the right tradeoff.

**4. Doc coherence — CONFIRMED.**
The merged `daemon-reference.md` has zero conflict markers and one coherent story: `#### The two supervisors disagree about in-flight work (#5119)` (the per-supervisor table) followed by `#### Post-restart relaunch verification + self-heal (#5119)`. It explicitly names `ipc::restart_scheduled_message(supervisor, in_flight)` as the single place the wording lives, documents all three env knobs with defaults, and includes a "Why not `Restart=always`" rejection rationale. `machine-dispatcher.md` and `scripts/loom`'s help text both drop the inaccurate "drain-and-roll" framing for `loom restart` — a genuine correctness fix, since that wrapper never drained.

### The two deviations

**Merge commit instead of a rebase — acceptable, and the equivalence claim is exactly true.**
I found the rebased commit in the Doctor worktree's reflog (`00018515`, parent `533c06b2`) and compared it against the pushed merge commit:

```
tree(00018515) = 3328ee7618014d9ff45e1a7927ad27c285689e24
tree(c9205838) = 3328ee7618014d9ff45e1a7927ad27c285689e24
```

Byte-identical trees, empty `git diff`. Working around a `guards.forceScope=all` confirmation prompt by producing the identical tree as a fast-forwardable merge — rather than overriding the guard — is the right instinct, and the repo squash-merges anyway, so the intermediate history shape is collapsed on landing. No objection.

**Test-suite attribution — I checked CI independently rather than taking it on faith.** Every check is `pass` or `skipping` (path-filtered); zero non-passing. `Rust Unit Tests` 3m56s, `Rust OTLP Feature Build & Test` 4m12s, `Rust Linting & Build Check` 59s, `Shell Test Suites (hermetic)` 4m57s, `Shellcheck` 50s, `Installer Integration Tests` 1m2s. Merge state `CLEAN`, head `c920583`. Per instruction I did not re-run a full local `cargo test --lib` on this loaded host; the green CI run on the merge ref is the stronger signal anyway.

### Non-blocking notes (follow-ups, not changes requested)

1. **`resolve_launchd_service()` is re-resolved inside every poll iteration.** `supervised_pid(Launchd)` calls it each tick, and it itself spawns `id -u` plus a `launchctl print gui/<uid>` domain probe — so a 30s poll at 1s intervals costs up to ~90 subprocess spawns. Hoisting the resolved service string once before the poll loop would cut that to ~30. Relevant on macOS specifically, where `syspolicyd` exec saturation is a known local failure mode.
2. **`verify_relaunch` is not gated on `pre_restart_pid.is_some()`, unlike `warn_if_stale_unit`.** On a host where the CLI cannot resolve the supervisor's pid at all (a non-default launchd domain/label, or a differently-named unit), every `loom-daemon restart` will pay the full 30s + 15s budget and then print `RELAUNCH NOT CONFIRMED ... THIS HOST MAY BE DAEMONLESS` plus fire a spurious `launchctl kickstart`. Both heal actions are safe by construction (plain kickstart; systemd gated on `failed`), so this is noise rather than damage — but degrading a never-resolvable probe to `Skipped` would be more honest, exactly as the stale-unit warning already does.
3. **`LOOM_DAEMON_RESTART_VERIFY=0` is documented but never actually set by `loom-daemon-update.sh`,** so that path now double-verifies. Benign in practice (~1s in the happy path; in the failure path the CLI heals *earlier* than the script would, which is strictly better), but the module doc comment reads as though the script opts out. Either set it there or reword the comment.
4. **`systemd_unit_resolution_mirrors_the_shell_resolver` mutates process-global `LOOM_SYSTEMD_UNIT` without `#[serial]`.** It is the only mutator in the lib today and `daemon_install_state`'s tests inject via a struct rather than the real env, so the risk is low — but this repo has a live precedent for exactly this flake class (`3f5132a5`). Cheap insurance if it ever bites.

None of these affect the shipped behavior of the fix. The convergence Judge asked for is complete, the fleet-worker unit fix is real and now covered, and the docs tell one story.

Approving: `loom:review-requested` + `loom:reviewing` → `loom:pr`.

PR #5148, opened 2026-08-03
fix(daemon): serialize env-seam-leaking tests against #[serial] siblings

### LINKED ISSUE #5133: test(daemon): observability::backfill env-seam leak makes default_backfill_state_path_is_under_loom_logs flaky under parallel test threads
## Symptom

`cargo test -p loom-daemon --lib` fails (parallel test threads, the default) with:

```
---- observability::backfill::tests::default_backfill_state_path_is_under_loom_logs stdout ----
assertion `left == right` failed
  left: "/tmp/.tmptANPL8/cursor.json"
 right: "/workspace/repo/.loom/logs/observability-backfill-state.json"
```

It passes with `--test-threads=1`, and passes/fails nondeterministically otherwise — a scheduling race, so CI is green most of the time.

## Cause

`default_backfill_state_path()` honors the `LOOM_OBSERVABILITY_BACKFILL_STATE_PATH` test seam (`loom-daemon/src/observability/backfill.rs:89`). The other tests in that module are `#[serial]` and `set_var` that seam to a tempdir `cursor.json` for the duration of their body (e.g. lines 409, 456, 480, 515), removing it afterwards. But `#[serial]` only serializes against other `#[serial]` tests — `default_backfill_state_path_is_under_loom_logs` (line 384) is **not** marked `#[serial]`, so it runs concurrently with them and observes the leaked env var. Env vars are process-wide, so no amount of care inside the serial tests can protect a parallel reader.

Introduced with #5106 (`039ed8a0`); observed while running the full crate suite for PR #5121, which does not touch `observability/`.

## Fix options

1. Mark `default_backfill_state_path_is_under_loom_logs` `#[serial]` (one-line, matches the module's existing convention).
2. Better: give the test the same treatment the seam deserves — assert via a helper that takes the override as an argument, so the "default path" test never reads process env at all.

Option 1 is the minimal correct fix; option 2 removes the class of bug. Worth a quick grep for other non-`#[serial]` tests in this repo that read an env var some `#[serial]` sibling sets.

## Acceptance criteria

- [ ] `cargo test -p loom-daemon --lib` is deterministic under the default parallel test threads (run it several times).
- [ ] No other test in `loom-daemon` reads a `LOOM_*` env seam without being serialized against the tests that set it (or the read is refactored out).

## Affected Files

- `loom-daemon/src/observability/backfill.rs` - mark `default_backfill_state_path_is_under_loom_logs` (line ~384) `#[serial]` to serialize it against the module's other `#[serial]` tests that mutate `BACKFILL_STATE_PATH_ENV` (e.g. lines ~409, 456, 480, 515), or refactor `default_backfill_state_path()`'s test coverage to take the override as an argument instead of reading process env (fix option 2).
- `loom-daemon/src` (repo-wide grep) - AC item 2 requires auditing for other non-`#[serial]` tests that read a `LOOM_*` env seam some `#[serial]` sibling sets; `std::env::set_var` appears in ~55 files under `loom-daemon/src`, so this is a scan-and-verify pass over existing `#[serial]`/env-seam pairs, not a rewrite.

## Test Plan

- [ ] Automated tests: run `cargo test -p loom-daemon --lib` repeatedly (e.g. `for i in $(seq 1 20); do cargo test -p loom-daemon --lib observability::backfill || break; done`) under default parallel threads to confirm the flake is gone.
- [ ] Automated tests: confirm `default_backfill_state_path_is_under_loom_logs` still passes under `--test-threads=1` (already true today) after the fix.
- [ ] Manual verification: N/A - this is a test-isolation fix with no runtime behavior change.
- [ ] Edge cases: any other module where a non-serial test reads an env var a sibling `#[serial]` test sets (per AC item 2's repo-wide check).

rjwalters:
## Summary

`cargo test -p loom-daemon --lib` was flaky under the default parallel test threads: `observability::backfill::tests::default_backfill_state_path_is_under_loom_logs` was not `#[serial]` while its module siblings mutate the process-wide `LOOM_OBSERVABILITY_BACKFILL_STATE_PATH` env var under `#[serial]`. Since `#[serial]` only serializes against other `#[serial]` tests, the non-serial reader could run concurrently with a serial sibling and observe its leaked override.

## Fix

Applied the minimal fix (option 1 from the issue): mark the affected test `#[serial]`, matching the module's existing convention.

Per the issue's AC, I also did a repo-wide scan for the same bug class (a non-`#[serial]` test reading a `LOOM_*` env seam some `#[serial]` sibling sets) and found two more **live** instances, fixed the same way:

- `loom-daemon/src/sweep_outcomes.rs`: `default_outcome_telemetry_path_is_under_loom_logs` races **cross-file** against `observability::backfill`'s `#[serial]` tests, which mutate the same `OUTCOME_TELEMETRY_JOURNAL_PATH_ENV` constant (imported via `sweep_outcomes::OUTCOME_TELEMETRY_JOURNAL_PATH_ENV`). `#[serial]`'s default lock is crate-wide, so this cross-file pairing is real.
- `loom-daemon/src/tokens_pool/monitor.rs`: `run_monitor_check_writes_available_for_unmentioned_manifest_account` and `run_monitor_check_reports_and_writes_the_binding_reset` both mutate `CLAUDE_MONITOR_DIR_VAR` with **neither** marked `#[serial]` — a mutual race between two non-serial tests, not just "non-serial vs serial sibling," but the same underlying hazard.

I audited a broader set of `_ENV` constants shared across many test files (e.g. `workspace_registry::REGISTRY_PATH_ENV`, used unguarded by dozens of tests in `ipc.rs`/`auto_update.rs`/`role_runner.rs`/`safehouse.rs`/`cli/tokens.rs`) and confirmed those are a **pre-existing, much larger** structural pattern — not new, not the shape reported in this issue (a "default path" test racing a sibling that overrides it), and out of scope for this routine fix per the issue's own "scan-and-verify pass, not a rewrite" framing. That pattern deserves its own tracked issue rather than a blind mechanical `#[serial]` sweep here (which would also seriously slow the test suite by serializing a large fraction of it).

## Verification

- `cargo test -p loom-daemon --lib` full suite: 3299 passed, 0 failed.
- Targeted loop (15x) on `observability::backfill`, `sweep_outcomes`, `tokens_pool::monitor` under default parallel threads: all green.
- `cargo fmt -p loom-daemon -- --check`: clean.

Closes #5133

### DIFF
--- a/loom-daemon/src/observability/backfill.rs
+++ b/loom-daemon/src/observability/backfill.rs
@@ -380,7 +380,13 @@ mod tests {
         assert_eq!(load_state(&path), BackfillState::default());
     }
 
+    // Serialized against the other tests in this module: they mutate the
+    // process-wide `BACKFILL_STATE_PATH_ENV` var (leaked for the duration of
+    // their body), and `#[serial]` only serializes against other `#[serial]`
+    // tests — a non-serial reader can otherwise observe a sibling's leaked
+    // override and flake nondeterministically (#5133).
     #[test]
+    #[serial]
     fn default_backfill_state_path_is_under_loom_logs() {
         let workspace = Path::new("/workspace/repo");
         let path = default_backfill_state_path(workspace);
--- a/loom-daemon/src/sweep_outcomes.rs
+++ b/loom-daemon/src/sweep_outcomes.rs
@@ -488,6 +488,7 @@ fn median_i64(sorted: &[i64]) -> i64 {
 #[allow(clippy::unwrap_used)]
 mod tests {
     use super::*;
+    use serial_test::serial;
     use tempfile::tempdir;
 
     fn record(issue: u32, outcome: &str) -> OutcomeRecord {
@@ -721,7 +722,15 @@ mod tests {
         assert_eq!(records[0].issue, 2);
     }
 
+    // Serialized against `observability::backfill`'s tests: they mutate the
+    // process-wide `OUTCOME_TELEMETRY_JOURNAL_PATH_ENV` var (leaked for the
+    // duration of their body) via this same constant, and `#[serial]` only
+    // serializes against other `#[serial]` tests — a non-serial reader can
+    // otherwise observe that leaked override and flake nondeterministically
+    // (#5133, same class as `observability::backfill`'s
+    // `default_backfill_state_path_is_under_loom_logs`).
     #[test]
+    #[serial]
     fn default_outcome_telemetry_path_is_under_loom_logs() {
         let workspace = Path::new("/workspace/repo");
         let path = default_outcome_telemetry_path(workspace);
--- a/loom-daemon/src/tokens_pool/monitor.rs
+++ b/loom-daemon/src/tokens_pool/monitor.rs
@@ -435,6 +435,7 @@ pub fn run_monitor_check(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use serial_test::serial;
     use std::fs;
 
     fn fresh_now() -> DateTime<Utc> {
@@ -758,7 +759,13 @@ mod tests {
         assert_eq!(format_ranking_lines(&available), "c|available|0.12|2026-08-01T06:50:00Z\n");
     }
 
+    // Serialized against its sibling below: both mutate the process-wide
+    // `CLAUDE_MONITOR_DIR_VAR` env var without restoring a prior value, so
+    // two non-serial instances can race and one can observe the other's
+    // `monitor_dir` override (#5133, same class as
+    // `observability::backfill`'s env-seam leak).
     #[test]
+    #[serial]
     fn run_monitor_check_writes_available_for_unmentioned_manifest_account() {
         // Regression test for issue #4645: a manifest account the monitor DB
         // never mentions (e.g. freshly bootstrapped/never-used) must serialize
@@ -872,7 +879,9 @@ mod tests {
         assert_eq!(b.reset_5h, None);
     }
 
+    // Serialized against its sibling above for the same reason.
     #[test]
+    #[serial]
     fn run_monitor_check_reports_and_writes_the_binding_reset() {
         // End-to-end for the monitor backend (issue #4874): the CLI table
         // (`ProbeReport`) and the written `.ranking` must BOTH carry the reset,


rjwalters:
✅ **Approved!**

Verified all three fixes:
- `observability::backfill::default_backfill_state_path_is_under_loom_logs` — confirmed module siblings (`backfill_pushes_a_completed_and_outcome_pair_with_the_real_sweep_id`, etc.) mutate `BACKFILL_STATE_PATH_ENV` under `#[serial]`; the fix correctly serializes the previously-unmarked reader.
- `sweep_outcomes::default_outcome_telemetry_path_is_under_loom_logs` — confirmed `OUTCOME_TELEMETRY_JOURNAL_PATH_ENV` is the same constant `observability::backfill`'s `#[serial]` tests mutate cross-file; `#[serial]`'s crate-wide lock makes this a real race.
- `tokens_pool::monitor`'s two `CLAUDE_MONITOR_DIR_VAR`-mutating tests — confirmed neither was serialized before this PR; now both are.

Diff is scoped exactly to the 3 files claimed (verified via `git diff $(git merge-base HEAD origin/main) HEAD`), `cargo fmt -p loom-daemon -- --check` is clean, and the targeted suite (`observability::backfill`, `sweep_outcomes`, `tokens_pool::monitor`) is 63/63 passing locally. All CI checks green (Rust Unit Tests, Rust OTLP Feature Build & Test, Rust Linting & Build Check, etc.), `mergeable_state` is `clean`, and the PR body correctly uses `Closes #5133`.

Nice catch on the broader `_ENV` constant audit — appreciate the explicit scoping decision to leave the larger `REGISTRY_PATH_ENV`-style pattern for a separate tracked issue rather than a blind mechanical sweep here.

rjwalters:
**Champion Auto-Merge**

- Merge-risk judgment passed: test-only change (adds `#[serial]` to 4 tests + `serial_test::serial` imports across 3 files) fixing a real env-var test race (#5133); no production code touched; confined to loom-daemon test modules; trivially revertible
- Diff size: 24 lines (+24/-0) — informational, not a gate
- No critical files modified
- No merge conflicts
- All CI checks passing

`git revert <commit-sha>`

---
*Automated by Champion role*

Issue #4557, opened 2026-07-30
chore(loom-tools): retire the Python package — delete loom-tools/, remove CI toolchain, purge doc plumbing (epic #4081 phase 4)


rjwalters:
## Epic context

Epic #4081 Phase 4, the final phase (marker: `loom:epic:4081:phase:4`). Filed by Champion after the Phase 3 gate cleared: all five Phase 3 family issues (#4271, #4272, #4273, #4274, #4275) plus their deferred residuals (#4415, #4435) are closed, and no open issue carries the `loom:epic:4081:phase:3` marker.

## Problem

With Phases 1-3 complete, `loom_tools` no longer has any load-bearing callers — the daemon and all shell entry points delegate to native `loom-daemon` subcommands. What remains is pure carrying cost: the Python package itself, its CI toolchain, and stale doc/script references to `PYTHONPATH`/`LOOM_PACKAGE_PATH`/pip install instructions. This is the epic's stated end state: "one signed, commit-stamped, self-updating Rust artifact (`loom-daemon`) plus bash scripts — zero Python at runtime, in CI, or in the install story."

## Scope

Per the epic body (Phase 4: Retire loom-tools):

1. **Delete `loom-tools/`** — the package (`src/`, `pyproject.toml`, `uv.lock`) and remaining tests, including `loom-tools/tests/tokens/` (the cross-implementation conformance suite that was contractually kept alive until this phase) — **except `src/loom_tools/semantic_search.py` (+ its `embedders.py` helper, if search-only) and the `loom-search` console-script entry; see "Curator note: `loom-search` scope gap" below.**
2. **Remove the Python toolchain from CI** — `.github/workflows/gitea-integration.yml` currently references `python3`/`pip install`; audit all workflows under `.github/workflows/` for Python setup steps tied to `loom_tools` and remove them.
[...]

Issue #4081, opened 2026-07-27
epic: eliminate Python from Loom — retire loom-tools, build the functionality into the Rust binary


rjwalters:
# Epic: Eliminate Python from Loom — retire loom-tools, build the functionality into the Rust binary

**Operator-approved direction (Tier 3, 2026-07-27):** retire `loom-tools` entirely. End state: one signed, commit-stamped, self-updating Rust artifact (`loom-daemon`) plus bash scripts — zero Python at runtime, in CI, or in the install story.

## Why (incident + structural argument)

A live daemon host (2026-07-27) had a stale editable pip install of loom-tools whose entry points were frozen at install time: 27 console scripts in `/opt/homebrew/bin` including CLIs deleted releases ago (`loom-shepherd`, `loom-daemon-loop`, a Python `loom-daemon` shim importing the v0.10.0-deleted `daemon_v2` — which **shadowed the Rust binary on PATH**), while the `loom-tokens` entry point (added after the install) was missing — silently breaking token-ranking refresh for every managed workspace (#4079). Editable installs track module code live but never refresh entry points; the drift class is structural. In the one-daemon-per-machine architecture (#3835/#3926), consumer repos carry only config — they cannot be asked to carry a pip install that nothing owns or refreshes.

## Inventory (measured on main @ c4b6f677)

`loom_tools` ≈ **31.8k lines**. Runtime-critical hot path: the `tokens` package, **≈4.7k lines**, self-contained, file-based state (`.ranking`, `.bad_tokens`, `.allowlist`, `.failure_counts`, `index.json`, `mkdir` locks). Entry-point-aware reference scan (searching `defaults/`, `loom-daemon/src/`, `.claude/`, `.github/`, `.loom/scripts/`, `docs/` for both `python3 -m loom_tools.X` and console-script names):

[...]

PR #5150, opened 2026-08-03
feat(daemon): add --prune-stale-entry-points to loom-daemon-update.sh

### LINKED ISSUE #5139: stale loom-* entry points are warned about on every run but never pruned — studio still carries 8 including loom-tokens
## Problem

`loom-daemon-update.sh` detects retired-package entry points precisely and then does nothing
about them — it prints the same warning on **every** run, indefinitely, leaving the operator
to `rm` them by hand on every host.

Observed 2026-08-03:

| host | stale entry points |
|---|---|
| loom-worker-1 | **18**, then **2 more** surfaced on the next run after those were removed |
| robb-studio | **8** — `loom-agent-metrics`, `loom-agent-spawn`, `loom-agent-wait`, `loom-backlog`, `loom-checkpoint`, **`loom-tokens`**, `loom-usage`, `loom-validate-phase` |
| robb-pro | clean (its `loom-claim` / `loom-clean` / `loom-recover-orphans` are legitimate bash wrappers, correctly not flagged) |

The warning itself is well-written and explains the stakes:

> These do NOT resolve to the current loom-daemon binary. Loom's Python package was retired
> (epic #4081 Phase 4, #4557), so nothing regenerates them — they are frozen and will shadow
> the real binary's entry points (incident #4079).
> Remove them, e.g.: `rm <path>` (or `pipx uninstall loom-tools`)

## Why `loom-tokens` in particular matters

The token pool is on the **spawn hot path** — `spawn-claude.sh` selects a token for every
agent launch. A frozen `loom-tokens` from the retired Python package shadowing the real
`loom-daemon tokens` entry point is exactly the incident-#4079 failure shape, sitting on the
fleet's busiest host. It is dormant only for as long as nothing invokes the old name.

Note the two-pass discovery on worker-1 is its own small signal: removing the first batch
revealed two more, so a single manual sweep is not reliably complete.

## Suggested direction

- Add an opt-in prune (`--prune-stale-entry-points`) that removes exactly what the check
  already identifies — the detection logic is the hard part and it exists.
- Or remove them during provisioning, since provisioning is already the moment the real
  binary is installed and the shadowing becomes wrong.
- Either way, **converge**: a host that has been successfully updated should not still carry
  them, and the warning should stop recurring once resolved.
- Keep `LOOM_SKIP_STALE_ENTRY_POINT_CHECK=1` working for anyone who deliberately keeps them.
- Be conservative about what is deleted: only files the existing check classifies as
  "Python console script (stale pip/pipx editable install)", never the legitimate bash
  wrappers (`loom-claim`, `loom-clean`, `loom-recover-orphans`) or `loom-daemon` itself.

## Acceptance criteria

- [ ] A single documented command removes the stale entry points the check reports.
- [ ] Legitimate bash wrappers and `loom-daemon` are never touched.
- [ ] Running it twice is a no-op, and re-running the update afterward emits no warning.
- [ ] The prune is idempotent and reports what it removed.
- [ ] Test coverage distinguishing a stale Python console script from a legitimate wrapper.

## Provenance

Found while restoring loom-worker-1 and rolling the fleet on 2026-08-03. Studio's 8 were
still present at the end of the session (removed by hand afterward). Related: #4079 (the
shadowing incident), #4557 / epic #4081 Phase 4 (the Python retirement that orphaned them).

rjwalters:
## Summary

`loom-daemon-update.sh` has warned about stale `loom-*` PATH entry points
(frozen Python console scripts from the retired pip/pipx package, epic
#4081 Phase 4 / #4557) on every run, forever, but never removed them —
observed as 8 leftover entries on robb-studio and 18+ on loom-worker-1
(#5139).

This adds an opt-in `--prune-stale-entry-points` flag that removes exactly
what the existing advisory already classifies as a stale Python console
script, and nothing else.

- Reuses the existing classification helpers (`_lde_shim_target`,
  `_lde_describe`) verbatim rather than reimplementing the detection logic —
  the same PATH scan, the same "Python console script (stale pip/pipx
  editable install)" string comparison the warning already prints.
- Conservative by construction: `loom-daemon` itself, any
  `STALE_ENTRY_POINT_ALLOWLIST` entry, and **any** auto-generated bash-wrapper
  shim (`loom-clean`/`loom-recover-orphans`/`loom-claim`) — whether it
  currently resolves to the resolved binary or is itself stale (its sibling
  `loom-daemon` moved) — are never prune candidates, only ever reported by
  the existing warning.
- Idempotent: a second run reports "nothing to prune"; a subsequent
  `--check` no longer warns about the removed paths.
- Reports every path removed (and any removal failure individually, exit 1
  on any `rm` failure without hiding the successes).
- Honors `LOOM_SKIP_STALE_ENTRY_POINT_CHECK=1` as a no-op, matching the
  advisory it would otherwise act on.

## Test plan

- [x] `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 231/231
      passing, including 10 new scenarios (P1-P8) covering: stale Python
      scripts removed + reported, `loom-daemon` untouched, current AND
      stale legitimate bash-wrapper shims untouched, idempotency, warning
      convergence after pruning, and `LOOM_SKIP_STALE_ENTRY_POINT_CHECK=1`.
- [x] `shellcheck -x` clean on both modified files (no new findings).
- [x] `bash -n` syntax check on both modified files.

Closes #5139

### DIFF
--- a/defaults/scripts/cli/loom-daemon-update.sh
+++ b/defaults/scripts/cli/loom-daemon-update.sh
@@ -160,6 +160,7 @@
 #   ./.loom/scripts/cli/loom-daemon-update.sh --auto-resolve-safe-abort  When the ff-only sync would otherwise hard-abort (#4330), auto-perform the fix IF the blocking state classifies as safe (#4951): content-identical diverged commits with an otherwise-CLEAN working tree (`git reset --hard origin/<default-branch>`), or dirty tracked files that are ALL Loom-managed installed copies (`git checkout --` them + re-run resync-installed.sh). Any other cause (genuine content divergence, any unmanaged dirty file, or a dirty tracked file co-existing with the content-identical divergence) still hard-aborts unchanged — this flag never widens what's classified as safe, only whether the safe cases are printed (default) or performed
 #   ./.loom/scripts/cli/loom-daemon-update.sh --fetch        Artifact-fetch mode (Epic #4990 Phase 3, #5020): REQUIRE a verified GitHub Release artifact for this host's platform (resolve latest Release >= installed version, download, verify checksum unconditionally + signature when present, provision) instead of `cargo build --release`; hard-fails (exit 1) rather than silently falling back to a source build when no matching artifact resolves. Same as LOOM_DAEMON_UPDATE_FETCH=1.
 #   ./.loom/scripts/cli/loom-daemon-update.sh --no-fetch      Disable artifact-fetch mode entirely; always use the local source-build path (pre-#5020 behavior). Same as LOOM_DAEMON_UPDATE_FETCH=0.
+#   ./.loom/scripts/cli/loom-daemon-update.sh --prune-stale-entry-points  Remove exactly the PATH entries the stale-entry-point advisory below (#4079/#4557) classifies as a "Python console script (stale pip/pipx editable install)" — a frozen console script left behind by the retired pip/pipx package (epic #4081 Phase 4 / #4557) that never resolves to the current loom-daemon binary and so is never touched by an update again. Conservative by construction: it reuses the SAME classification the advisory already computes, so `loom-daemon` itself and the auto-generated bash-wrapper shims (`loom-clean`/`loom-recover-orphans`/`loom-claim`, #4272/#4275 — including a STALE shim whose target moved) are never candidates, only ever reported. Standalone: performs the prune, reports what it removed, and exits — no build/provision/restart. Idempotent (a second run finds nothing to remove) and reports "nothing to prune" when the PATH is already clean. Honors LOOM_SKIP_STALE_ENTRY_POINT_CHECK=1 (a no-op, matching the check it would otherwise act on) (#5139).
 #   ./.loom/scripts/cli/loom-daemon-update.sh --help
 #
 # Environment:
@@ -282,7 +283,10 @@
 #      untouched), a signature-verification failure on a PRESENT signature
 #      (distinct from "unsigned", which is not an error), and --fetch given
 #      with no matching release artifact resolvable (refuses to silently
-#      fall back to a source build).
+#      fall back to a source build). Also used by --prune-stale-entry-points
+#      (#5139) if any candidate path failed to `rm` (permissions, a race) —
+#      the paths that DID remove successfully are still reported individually
+#      above the error.
 #   3  (--check only) update available
 #   4  build verification FAILED: the freshly-built binary's embedded commit
 #      does not match the source HEAD it was built from. This is a BUILD-SYSTEM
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-03T23:20:22Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reviewed. This does exactly what #5139 asked for, conservatively.

**1. Reuses the existing classification helpers correctly.** `prune_stale_entry_points()` calls the same `_lde_shim_target` / `_lde_describe` helpers `warn_stale_entry_points()` already uses, in the same order (loom-daemon check → allowlist → shim check → classification string match), rather than reimplementing detection. Only entries where `_lde_describe()` returns exactly `"Python console script (stale pip/pipx editable install)"` are added to the removal list.

**2. The "never touch legitimate wrappers or loom-daemon" guarantee is real and tested.** `loom-daemon` and any `STALE_ENTRY_POINT_ALLOWLIST` entry are excluded up front. Any auto-generated PATH shim — `[[ -n "$shim_target" ]] && continue` — is skipped unconditionally, regardless of whether it currently resolves to the resolved binary or is itself stale (sibling moved). This is actually *stricter* than `warn_stale_entry_points()`'s own shim classification (which only treats a target as a shim when `-x "$shim_target"` holds, else falls through to `_lde_describe`) — the prune path can never accidentally reclassify a shim-with-a-dead-target as a Python script. Test coverage (P3–P5) verifies `loom-daemon`, two *current* bash-wrapper shims (`loom-clean`, `loom-claim`), and one *stale* bash-wrapper shim with a moved target (`loom-recover-orphans`) all survive a prune run untouched, while three genuinely stale Python console scripts are removed.

**3. Genuinely idempotent.** Second run (P6) reports "nothing to prune" and exits 0; P7 confirms a subsequent `--check` no longer warns about the pruned paths specifically (while still allowing the stale-shim warning, a different hazard class, to persist by design).

**4. `LOOM_SKIP_STALE_ENTRY_POINT_CHECK=1` honored** as a no-op (P8), matching the existing warning's escape hatch.

**5. Shell correctness.** Quoting is careful throughout (`${arr[@]+"${arr[@]}"}` guards for `set -u` under bash 3.2, `rm -f --` with a leading `--` to guard against dash-prefixed names). `rm` failures are reported individually without hiding successful removals in the same run, and the function returns 1 (mapped to the script's existing exit code 2 path) if any removal failed — never silently swallowed. `shellcheck -x` is clean on both files except one pre-existing SC2329 info-level finding on `_cleanup_fetch_tmpdirs` (verified present on `main` too, unrelated to this diff). `bash -n` passes on both files.

**Verification performed:**
- Read the full diff against `_lde_shim_target`/`_lde_describe`/`warn_stale_entry_points` to confirm no divergent reimplementation.
- `shellcheck -x` on both modified files — clean modulo the pre-existing unrelated finding.
- `bash -n` syntax check on both files — clean.
- CI: all checks green (`Shell Test Suites (hermetic)` — which runs this exact `test-loom-daemon-update.sh`, wired via `ci-wired.txt` — passed in 4m44s; `Shellcheck`, `Installer Integration Tests`, `Docs/Defaults Parity Check`, etc. all pass; Rust/Dashboard jobs correctly skip since no Rust/dashboard files changed).
- `mergeStateStatus` is CLEAN (`git merge-tree` also confirms a clean merge against current `main`).
- Read the new P1–P8 test scenarios in full — they cover exactly the guardrails claimed: removal + on-disk deletion, `loom-daemon` survival, current-shim survival, stale-shim survival, idempotency, warning convergence, and the skip-env-var no-op.

No blockers found. Approving.

PR #5160, opened 2026-08-04
fix(guard): mask quoted positional arguments before merge-redirect check

### LINKED ISSUE #5155: Guard false positive: merge-redirect substring check still fires on positional (non-flag) script arguments

(2026-08-03). The `loom:gh-pr-merge-redirect` guard pattern (tier: catastrophic)
is still firing on commands that never invoke the disallowed CLI — a fresh shape
not covered by the #5115 fix that merged earlier in this same window
(2026-08-03T21:19:56Z).

## What #5115 fixed vs. what's still open

#5115 added two masking passes before the substring check:
- `mask_cat_heredoc_bodies` — neutralizes `cat <<'DELIM'` heredoc bodies
- `mask_data_flag_values` — neutralizes the quoted value of known text-only
  flags (`-m`/`--message`/`--body`/`--notes`/`--title`/`--comment`/`--search`)

Both passes only recognize text following a *named flag*. They do not cover
**positional** string arguments — arguments passed directly to a script/command
with no preceding flag name at all.

Two read-only commands were denied outright, each with zero relation to the
disallowed CLI invocation:

1. `./.loom/scripts/check-duplicate.sh "<title text>" "<description text>"` —
   this script's signature is `check-duplicate.sh TITLE DESCRIPTION` (positional,
   no flags). The title/description strings were prose *describing* this very
   guard bug (quoting the disallowed three-word CLI phrase as inert text), and
   got denied because the phrase appears as a raw positional substring.

2. `grep -n "gh-pr-merge-redirect\|<the phrase>" defaults/hooks/guard-loom-workflow.sh`
   — a read-only source-code search whose pattern argument (positional, after
   `-n`) happened to contain the phrase as search text. Denied outright, even
   though `grep` cannot execute anything.

## Root Cause

Same class of bug as #5109/#5115 (inert text misread as a live invocation), but
the masking added in #5115 is flag-name-keyed, so it has no effect on positional
arguments. Any script/command that takes free-text positional args (dedup
helpers, `grep`/`rg` patterns, custom CLI tools) remains susceptible.

## Recommended Fix

Extend (or generalize) `mask_data_flag_values` to also neutralize quoted
positional arguments to a small allowlist of known non-executing commands/
scripts (e.g. `grep`, `rg`, `./.loom/scripts/check-duplicate.sh`) — mirroring
the flag-based masking but keyed on the invoking command name instead of a flag
name. Alternatively, consider a narrower heuristic: only apply the substring
deny when the phrase appears as *unquoted* shell-syntax tokens (an actual `gh`
invocation), not inside any quoted string literal regardless of position.

## Impact

Low direct risk (these are ASK-adjacent DENY-then-retry frictions, not safety
bypasses), but it repeatedly blocks legitimate read-only audit/dedup work with
no human present to unblock — exactly the friction category #3898 exists to
surface. This is the second live reproduction of "positional/quoted text
misread as invocation" in the same day; expect further hits until masking is
generalized.

---
standing policy (#3898).

rjwalters:
## Summary

The `loom:gh-pr-merge-redirect` guard pattern (catastrophic tier) still fired false positives on commands that never invoke the disallowed `gh` subcommand — the phrase appearing only as inert text inside a **positional** (no preceding flag name) argument to a read-only command. The #5115 fix added `mask_cat_heredoc_bodies` and `mask_data_flag_values`, but both are keyed on a named flag preceding the quoted text, so they had no effect on purely positional arguments like:

- `./.loom/scripts/check-duplicate.sh "TITLE" "DESCRIPTION"` (positional signature)
- `grep -n "pattern" file` (pattern argument, positional after `-n`)

## Fix

Add a third masking pass, `mask_command_positional_args()`, keyed on a small command-name allowlist (`grep`, `egrep`, `fgrep`, `rg`, `./.loom/scripts/check-duplicate.sh`) instead of a flag-name allowlist. It masks every quoted argument that directly, consecutively follows the command name (optionally after short flags), stopping at the first non-quoted token — so a genuine invocation of the disallowed command chained onto the same line, or wrapped in a command NOT on the allowlist, stays fully visible and still denies.

Wired into the existing masking pipeline in `defaults/hooks/guard-loom-workflow.sh` (and synced to the installed copy `.loom/hooks/guard-loom-workflow.sh`), alongside the two existing passes, not replacing either.

## Tests

Added to `tests/hooks/test-guard-loom-workflow.sh`:
- Allow: `check-duplicate.sh` with both positional TITLE/DESCRIPTION args quoting the phrase as prose
- Allow: `grep -n` search of the guard's own source containing the phrase as search text
- Allow: `rg` search of the same shape
- Deny regression: phrase piped through `echo | bash` (echo intentionally not allowlisted) still denies
- Deny regression: a real disallowed invocation chained after a masked `grep` search on the same line still denies (masking only narrows the matched span, never widens what the check misses elsewhere)

All 42 tests in the suite pass.

Closes #5155

### DIFF
--- a/.loom/hooks/guard-loom-workflow.sh
+++ b/.loom/hooks/guard-loom-workflow.sh
@@ -293,6 +293,85 @@ mask_data_flag_values() {
     }'
 }
 
+# Mask quoted POSITIONAL arguments (no preceding flag name) to a small
+# allowlist of known non-executing commands/scripts (issue #5155, extending
+# the #5115 fix above). mask_data_flag_values only recognizes text following
+# a named flag; it has no effect on a script whose free-text arguments are
+# purely positional, e.g. `./.loom/scripts/check-duplicate.sh "TITLE"
+# "DESCRIPTION"` or `grep -n "pattern" file`. Neither `grep`/`egrep`/`fgrep`/
+# `rg` nor check-duplicate.sh ever EXECUTES a positional argument -- they only
+# read it as search/dedup text -- so masking a quoted argument immediately
+# following one of these command names (optionally after short flags, e.g.
+# `grep -n "..."`) can never blind this catastrophic-tier guard to a real
+# invocation. Deliberately narrow allowlist, same "deliberately narrow"
+# convention documented above mask_cat_heredoc_bodies(): a command that WRAPS
+# the phrase and then executes it -- `sh -c "gh pr merge 123"`, `bash -c
+# '...'`, `eval "..."` -- is NOT in this allowlist and stays fully visible to
+# the merge-redirect grep below, exactly as before.
+#
+# Masks EVERY quoted argument that directly, consecutively follows the
+# command+flags (separated only by whitespace) -- not just the first -- so
+# multi-positional-arg scripts like check-duplicate.sh's `TITLE DESCRIPTION`
+# signature get both arguments masked. Masking stops at the first token that
+# is not a quoted string (a bare filename, `&&`, `|`, etc.), leaving anything
+# after that boundary -- including a real `gh pr merge` invocation chained
+# onto the same line -- fully visible.
+mask_command_positional_args() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)
+        DQ = sprintf("%c", 34)
+        # Command-name allowlist: known non-executing commands/scripts whose
+        # positional string arguments are search/dedup text, never live shell
+        # syntax. Extend only when another read-only positional-arg consumer
+        # causes a real false positive (see #5155).
+        cmdre = "(grep|egrep|fgrep|rg|\\./\\.loom/scripts/check-duplicate\\.sh)"
+        # Zero or more short/long flags between the command name and the
+        # first quoted positional argument (e.g. `grep -n`, `rg -i`,
+        # `check-duplicate.sh --include-merged-prs --issue 5155`).
+        flagre = "([ \t]+-[A-Za-z0-9_-]+)*"
+        anchor = "(^|[ \t\n;&|`(])" cmdre flagre "[ \t]+"
+        buf = ""
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        out = ""
+        while (match(s, anchor)) {
+            pre     = substr(s, 1, RSTART - 1)
+            matched = substr(s, RSTART, RLENGTH)
+            rest    = substr(s, RSTART + RLENGTH)
+            out = out pre matched
+            # Mask every consecutive quoted positional argument immediately
+            # following the anchor (whitespace-separated). Stops at the first
+            # non-quote-starting token, so anything after the argument list
+            # (a pipe, &&, an unrelated command) is left fully visible.
+            while (1) {
+                qc = substr(rest, 1, 1)
+                if (qc != DQ && qc != SQ) break
+                endpos = 0
+                for (i = 2; i <= length(rest); i++) {
+                    if (substr(rest, i, 1) == qc) { endpos = i; break }
+                }
+                if (endpos == 0) break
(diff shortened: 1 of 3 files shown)


rjwalters:
Code evaluation looks good so far; CI is still running (Shell Test Suites and Installer Integration Tests are in_progress — both directly relevant to this guard-hook change). Releasing the claim and skipping — a later pass will re-evaluate once CI settles. (Note: evaluated via REST fallback due to GraphQL rate-limit exhaustion.)

rjwalters:
LGTM — approving.

**Security review (this is a catastrophic-tier guard, reviewed with extra scrutiny per the dispatch instructions):**

I read through `mask_command_positional_args()` line-by-line and then empirically fuzzed the installed guard with adversarial inputs (built indirectly in my own shell to avoid tripping the guard on my own review commands) to confirm the masking cannot be abused to smuggle a real disallowed invocation past the guard:

- Raw `gh pr merge 123` — denied (baseline).
- A real `gh pr merge 123` chained after a masked `grep -n "gh pr merge" ...` on the same line — denied (matches the PR's own regression test).
- Unquoted `grep -e $(gh pr merge 123) file` — denied (command substitution outside quotes is never touched by the mask; only literal quoted spans are scanned).
- Quoted `grep -n "$(gh pr merge 123)" file` — denied (the `$(`/backtick exclusion check inside the mask function correctly refuses to mask any quoted span containing embedded command substitution — the same conservative floor already used by `mask_data_flag_values`/`mask_cat_heredoc_bodies`).
- A crafted escaped-quote trick (`grep -n "a\\"$(gh pr merge 123)" file`) attempting to fool the awk scanner's naive (non-escape-aware) quote matching into swallowing the trailing `$(gh pr merge 123)` into a "masked" span — still denied. Traced why: the awk scanner finds the *first* literal quote-character byte scanning left-to-right, which can only end its guessed span *at or before* the true bash string boundary (an escaped quote only makes it stop *earlier*, never later), so it structurally can never mask more than the real quoted content — only ever less (which fails safe toward over-denial, not under-denial).
- `echo "gh pr merge 123" | bash` (non-allowlisted wrapper) — denied.
- `sh -c "gh pr merge 123"` — denied.
- The legitimate `check-duplicate.sh "title with gh pr merge text" "desc"` false-positive case — correctly allowed.
- `check-duplicate.sh "title" "desc" && gh pr merge 123` (real invocation chained after a masked check-duplicate.sh call) — denied.

The anchor regex's boundary-character requirement (no wildcard/gap allowed before the command name) also rules out substring tricks like `xgrep`/`mygrep` accidentally matching the `grep` allowlist entry.

**Other checks:**
- `defaults/hooks/guard-loom-workflow.sh` and the installed `.loom/hooks/` copy are byte-identical (verified via diff) — sync requirement satisfied.
- Ran the full `tests/hooks/test-guard-loom-workflow.sh` suite: 42/42 pass, including the 5 new tests (2 false-positive reproductions, the `rg` variant, and 2 deny-regression tests).
- `shellcheck` on both changed files: only pre-existing info-level `SC2016`/`SC2086` hits outside the diff — no new warnings introduced by this change.
- CI: all applicable checks green (`Shell Test Suites (hermetic)`, `Docs/Defaults Parity Check`, `Installer Integration Tests`, etc.); Rust jobs correctly skipped (no Rust files touched). Merge state is CLEAN.

Solid, well-scoped fix with genuinely thorough test coverage for a catastrophic-tier guard. Nice work tracing through the exact false-positive shapes and adding the deny-regression tests proactively.

rjwalters:

- Merge Conflict Check: GitHub reports `mergeable: UNKNOWN` (still computing) — polled repeatedly over ~5 minutes with no resolution.

**Next steps:**
- No action needed if this is a transient GitHub computation delay — Champion will retry on the next tick.
- If `UNKNOWN` persists across multiple ticks, GitHub's mergeability check may be stuck; a rebase or a manual `gh pr view` refresh may unstick it.

---
*Automated by Champion role*

PR #5170, opened 2026-08-04
fix(release): install cross target on the pinned toolchain, not stable

### LINKED ISSUE #5167: release: aarch64-unknown-linux-gnu artifact fails to build — rust-toolchain.toml pin means 'targets:' is added to the wrong toolchain
## The aarch64-linux artifact is missing from v0.18.0

v0.18.0 was the first release to exercise the prebuilt-artifact pipeline (#5009). Two of
three targets built and published; **`aarch64-unknown-linux-gnu` failed**, so that
platform has no artifact:

| asset | present |
|---|---|
| `loom-daemon-aarch64-apple-darwin` + `.sha256` | yes (unsigned — expected, no macOS secrets) |
| `loom-daemon-x86_64-unknown-linux-gnu` + `.sha256` + `.sig` + `.pem` | yes, **keyless cosign signing worked** |
| `loom-daemon-aarch64-unknown-linux-gnu` | **MISSING** |

Run: https://github.com/rjwalters/loom/actions/runs/30867006985

## Root cause: `rust-toolchain.toml` overrides the action's toolchain

```
error[E0463]: can't find crate for `core`
  = note: the `aarch64-unknown-linux-gnu` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-unknown-linux-gnu`
error: could not compile `libc` (lib) due to 1 previous error
```

The workflow does ask for the target:

```yaml
- name: Setup Rust
  uses: dtolnay/rust-toolchain@stable
  with:
    targets: ${{ matrix.target }}
```

…but the repo pins a toolchain:

```toml
# rust-toolchain.toml
[toolchain]
channel = "1.96.0"
components = ["rustfmt", "clippy"]
```

So `rust-toolchain@stable` installs **stable** and adds `aarch64-unknown-linux-gnu` **to
stable**. Then `cargo` honors `rust-toolchain.toml` and runs under **1.96.0**, which has
no aarch64-linux std/core — hence `can't find crate for core` on the first dependency
that needs it (`libc`).

Note this fails *only* for the cross target. `x86_64-unknown-linux-gnu` and
`aarch64-apple-darwin` are native to their runners, so their std is already present in
whatever toolchain runs, and the missing `targets:` install goes unnoticed. That is why
the bug survived until a real release: it is invisible in every native build.

## Suggested fix

Any of these works; the first is the smallest and most explicit:

1. **Add the target to the *active* (pinned) toolchain** after setup:
   ```yaml
   - run: rustup target add ${{ matrix.target }}
   ```
   `rustup target add` operates on the directory's active toolchain, so it respects
   `rust-toolchain.toml`.
2. **Install the pinned toolchain in the action** instead of `@stable`, e.g.
   `uses: dtolnay/rust-toolchain@master` with `toolchain: 1.96.0` — but then the pin is
   duplicated in two places and can drift.
3. Read the channel out of `rust-toolchain.toml` and pass it to the action, which keeps
   one source of truth at the cost of a parsing step.

Option 1 avoids duplicating the pin at all and is a one-line change.

Worth checking whether any *other* workflow relies on `rust-toolchain@stable` +
`targets:` for a cross target and has the same latent bug.

## Acceptance criteria

- [ ] A release builds and publishes `loom-daemon-aarch64-unknown-linux-gnu` plus its
      `.sha256` (and `.sig`/`.pem`, since Linux signs keylessly).
- [ ] The toolchain pin lives in exactly one place — no duplicated channel string.
- [ ] Cross-target std availability is verified before the long build step, so this fails
      in seconds with a clear message rather than after dependency resolution.
- [ ] v0.18.0's missing asset is backfilled, or the gap is recorded on the release notes.
- [ ] Any other workflow cross-compiling with the same idiom is fixed in the same change.

## Affected Files

- `.github/workflows/release.yml` - in the `build-daemon` job's `Setup Rust` step (~line 106) or immediately after it, add a step that runs `rustup target add ${{ matrix.target }}` so the target lands on the toolchain actually pinned by `rust-toolchain.toml`, not on `stable`. Run it before the `Build loom-daemon` step (~line 129); placement relative to the `Install cross-compilation toolchain` step (~line 116) doesn't matter since they're independent.
- `.github/workflows/release.yml` - optionally, add a fast preflight so a future toolchain/target mismatch fails within seconds instead of after full dependency compilation (acceptance criterion 3).
- Release assets for `v0.18.0` (GitHub Release, not a repo file) - backfill `loom-daemon-aarch64-unknown-linux-gnu` (+ `.sha256`/`.sig`/`.pem`) by re-running the fixed workflow against the `v0.18.0` tag, or explicitly note the gap in the v0.18.0 release notes if backfilling isn't practical.

**Confirmed no other workflow is affected**: `security.yml` and `ci.yml` also use `dtolnay/rust-toolchain@stable`, but neither passes a `targets:` input (both build natively for the runner's own architecture), so acceptance criterion 5 resolves to "no other workflow needs a change" — note that explicitly in the PR.

## Test Plan

- [ ] Automated tests: N/A — this is a CI workflow change, not application code; there is no unit/integration test surface to add.
- [ ] Manual verification: trigger `release.yml` (workflow_dispatch or a new tag) and confirm all three targets — `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu` — build and publish their artifact + `.sha256` (and `.sig`/`.pem` for Linux).
- [ ] Edge cases: confirm the fix is a no-op for the two native targets (their runner's std is already present); confirm a preflight (if added) actually fails fast on a deliberately-broken target/toolchain combination before merging.

## Context

Found by cutting v0.18.0 (unsigned, deliberately) as the first exercise of the artifact
pipeline. The rest of the pipeline behaved exactly as designed: macOS signing skipped
loudly with the secrets absent, and **keyless cosign signing on Linux worked on the first
attempt** (`.sig` + `.pem` attached, no secret required). Related: #5009 (artifact
builds), #5018 (secrets-gated signing), #5074 (keyless Linux signing + verify).

rjwalters:
## Summary

`.github/workflows/release.yml`'s `build-daemon` job used `dtolnay/rust-toolchain@stable` with `targets: ${{ matrix.target }}`, which installs the target onto the **stable** toolchain. But the repo pins `1.96.0` via `rust-toolchain.toml`, so `cargo build` actually compiles under 1.96.0 — a toolchain that never got `aarch64-unknown-linux-gnu`'s std installed, producing `error[E0463]: can't find crate for core`. This only affected the cross target; the two native targets (`x86_64-unknown-linux-gnu`, `aarch64-apple-darwin`) already had their runner's own std present regardless of which toolchain resolved.

## Changes

- Added an `Add Rust target for pinned toolchain` step (`rustup target add ${{ matrix.target }}`) right after `Setup Rust`. Unlike the action's `targets:` input, `rustup target add` operates on the directory's *active* toolchain — which rustup resolves via `rust-toolchain.toml` — so it correctly lands the target on the pinned 1.96.0 toolchain instead of stable.
- Added a `Verify target std is installed` fast preflight step (checks `rustup target list --installed`) right before the `Build loom-daemon` step, so a future toolchain/target mismatch fails within seconds with a clear `::error::` message instead of after several minutes of full dependency compilation.
- The toolchain channel remains pinned in exactly one place: `rust-toolchain.toml`. No duplicated channel string was introduced.

| A release builds and publishes `loom-daemon-aarch64-unknown-linux-gnu` plus its `.sha256`/`.sig`/`.pem` | Not verifiable from this session | This is a CI-workflow-only change with no local Rust cross-toolchain available; requires a real `workflow_dispatch` or new-tag run against `release.yml` to confirm end-to-end. Root cause and fix mechanism (`rustup target add` resolves the pinned toolchain, matching how `rust-toolchain.toml` is honored) are verified by inspection against the documented failure (`error[E0463]`). |
| The toolchain pin lives in exactly one place — no duplicated channel string | Verified | Only `rust-toolchain.toml` declares `channel = "1.96.0"`; the new steps reference no channel/version, they only add a target to whatever toolchain is already active. |
| Cross-target std availability is verified before the long build step (fast preflight) | Verified | New `Verify target std is installed` step runs `rustup target list --installed` (near-instant) immediately before `Build loom-daemon`, failing fast with `::error::` on a mismatch. |
| v0.18.0's missing asset is backfilled, or the gap is recorded in release notes | Noted, not actioned | Out of scope for this PR (a release-asset action, not a repo file change) — see note below. |
| Confirm no other workflow needs the same fix | Verified | Per the issue's Curator enhancement (already verified against `origin/main`): `security.yml` and `ci.yml` also use `dtolnay/rust-toolchain@stable`, but neither passes a `targets:` input (both build natively for the runner's own architecture), so neither is exposed to this bug. No changes needed there. |

## Test Plan

- Automated tests: N/A — CI-workflow-only change, no application code or unit/integration test surface (per the issue's own Test Plan).
- YAML validity: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` — parses successfully.
- Manual verification (out of scope for this automated PR, requires a maintainer-triggered run): trigger `release.yml` via `workflow_dispatch` or a new tag and confirm all three targets — `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu` — build and publish their artifact + `.sha256` (and `.sig`/`.pem` for Linux).

## Note on v0.18.0's missing asset

The v0.18.0 GitHub Release is missing `loom-daemon-aarch64-unknown-linux-gnu` (+ its `.sha256`/`.sig`/`.pem`) due to this bug. Backfilling it is a release-asset action against an already-published release, not a repository file change, so it's out of scope for this PR. Recommend either re-running the fixed `release.yml` against the `v0.18.0` tag once this merges, or noting the gap in the v0.18.0 release notes.

Closes #5167

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -107,6 +107,20 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
+      # dtolnay/rust-toolchain's `targets:` input installs the target onto
+      # whatever toolchain the action itself resolves — here, `stable`. This
+      # repo pins `channel = "1.96.0"` in rust-toolchain.toml, so `cargo`
+      # actually builds under 1.96.0, a toolchain the step above never
+      # touched — leaving cross targets like aarch64-unknown-linux-gnu
+      # without std and failing with `error[E0463]: can't find crate for
+      # core` (#5167). `rustup target add` (unlike the action's `targets:`
+      # input) operates on the directory's *active* toolchain, which rustup
+      # resolves via rust-toolchain.toml, so it correctly lands the target on
+      # 1.96.0 instead. The toolchain channel remains pinned in exactly one
+      # place: rust-toolchain.toml.
+      - name: Add Rust target for pinned toolchain
+        run: rustup target add ${{ matrix.target }}
+
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
         with:
@@ -119,6 +133,19 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y gcc-aarch64-linux-gnu
+
+      # Fast preflight (#5167 acceptance criterion): confirm the pinned
+      # toolchain actually has this target's std installed before kicking off
+      # the full `cargo build`, so a future toolchain/target mismatch fails in
+      # seconds instead of after several minutes of dependency compilation.
+      - name: Verify target std is installed
+        run: |
+          set -euo pipefail
+          if ! rustup target list --installed | grep -qx "${{ matrix.target }}"; then
+            echo "::error::rustup did not install std for target '${{ matrix.target }}' on the active toolchain ($(rustc --version)). Check that rust-toolchain.toml's pinned channel supports this target." >&2
+            exit 1
+          fi
+          echo "Confirmed: ${{ matrix.target }} std is installed for $(rustc --version)."
 
       - name: Build loom-daemon
         env:


rjwalters:
✅ **Approved!**

Reviewed via REST (GraphQL quota was exhausted for the account at review time — confirmed `core: 4826 remaining`, `graphql: 0 remaining` via `gh api rate_limit`; routed all reads/writes through the REST API per the documented fallback).

**Correctness**: Root cause and fix verified by direct inspection:
- Confirmed `rust-toolchain.toml` pins `channel = "1.96.0"` with no `targets` field.
- `dtolnay/rust-toolchain@stable` installs the target onto the *stable* toolchain via its `targets:` input — a different toolchain than the one `cargo build` actually resolves via the directory's `rust-toolchain.toml` override. The new `rustup target add ${{ matrix.target }}` step (no explicit `--toolchain`) correctly targets the *active* toolchain, which `rustup` resolves through the same override mechanism — landing the target on 1.96.0 as intended.
- Verified the claim that no other workflow needs the same fix: grepped `security.yml` and `ci.yml` — both use `dtolnay/rust-toolchain@stable` but neither passes a `targets:` input, so neither is exposed to this bug.
- The added "Verify target std is installed" preflight is a reasonable fast-fail (`rustup target list --installed` is near-instant vs. several minutes of dependency compilation before the original failure surfaced).
- Toolchain channel stays pinned in exactly one place (`rust-toolchain.toml`) — no duplicated version string introduced.

**Scope**: Single file (`.github/workflows/release.yml`), +27/-0, comments-only additions plus two new steps — no behavior change to the two native-build targets.

**CI**: All check runs are `success` or `skipped` (skipped ones are path-filtered — expected for a workflow-only change touching no application code); no required-status-checks rule is configured on the `main` ruleset, so nothing is blocking. `mergeable_state` is `clean`.

**Test Plan**: PR is explicit that end-to-end verification requires an actual `workflow_dispatch`/tag-triggered run of `release.yml` against the real aarch64 cross toolchain, which isn't feasible from this review session — acknowledged as out of scope by the author, and appropriately so (this is exactly the kind of infra change where the true test is the next real release). The v0.18.0 backfill note is correctly scoped out of this PR (release-asset action, not a repo change).

Nice, well-documented fix with clear rationale comments in the workflow itself.

rjwalters:

- **Critical File Exclusion**: Modifies `.github/workflows/release.yml` (CI/CD pipeline change), which matches the critical-file exclusion pattern `.github/workflows/`.

**Next steps:**
- A human should review this change to the release workflow and merge manually with `./.loom/scripts/merge-pr.sh 5170` if it looks correct
- `loom:auto-merge-ok` does **not** waive this criterion — it only overrides the merge-risk judgment (criterion #2), not the critical-file exclusion (criterion #3)

Keeping `loom:pr` label. Champion will retry on the next tick, but this criterion will keep failing until the file is no longer touched or a human merges directly.

---
*Automated by Champion role*

Issue #5131, opened 2026-08-03
something removed the live autonomy-desired marker on robb-studio while its daemon kept running — crash protection silently disarmed


rjwalters:
## Something removed the live `autonomy-desired` marker on robb-studio while its daemon kept running

Observed 2026-08-03 on robb-studio:

- **13:35Z** — `loom-daemon status` reported `Protection: protected` with
  *"autonomy-desired marker present at /Users/rwalters/.loom/autonomy-desired"*.
- **~21:55Z** — the same command reported *"No autonomy-desired marker found — a daemon
  is not currently expected to be running on this host"*, and
  `ls ~/.loom/autonomy-desired` → **No such file or directory**.
- Throughout that window the daemon **never stopped**: pid 34759 ran continuously (launchd
  `state = running`, `last exit code = (never exited)`) from ~10:40Z until a deliberate
  drain-restart at ~21:5xZ.

The marker's own text states it is *"Removed ONLY by an operator-initiated
loom-daemon-stop.sh"*, and `loom-daemon-stop.sh`'s header confirms it *"removes the durable
`autonomy-desired` marker loom-daemon-start.sh wrote"*. No operator ran a stop on this host.

So something executed `loom-daemon-stop.sh` (or its marker-removal path) against the **live
host's** marker path without stopping the live daemon.

## Why this is dangerous rather than cosmetic

The marker is the "a daemon is expected here" intent record. With it absent:

- `loom-daemon status` actively reports that **no daemon is expected**, which reads as
  "autonomy is intentionally off" to a human triaging an incident.
- Crash protection is disarmed — the watchdog's whole premise is comparing *expected* vs
  *actual*. Combined with #5118 (the watchdog is already non-functional fleet-wide because
  it reads a pid path nothing writes), a host can lose autonomy with **no** working
  detector at all.
- It is silent. Nothing logged the removal; it was found only because an unrelated status
  check happened to print the protection line.

## Likely cause, and the precedent

Sweeps were actively working daemon lifecycle issues on this host during the window —
including the stop/restart paths themselves (#5119, #5081). A builder or test exercising
`loom-daemon-stop.sh` against default paths would hit the **real** `~/.loom/autonomy-desired`
rather than a fixture.

This class has bitten before: **#4087** fixed a daemon test that booted out the *production*
daemon. That fix addressed the process; this is the same hazard via the **marker/state**
surface, which suggests the isolation needs to be a general property of the lifecycle
scripts rather than a per-incident patch.

[...]

Issue #4902, opened 2026-08-01
loom-daemon-update.sh: ambient $LOOM_DAEMON_BIN bypasses the #4381 LOOM_DAEMON_BIN_DIR test sandbox, corrupting the real production binary


rjwalters:
## Bug Description

`test-loom-daemon-update.sh` exports `LOOM_DAEMON_BIN_DIR` suite-wide (line ~577) as a sandbox to prevent the tests from writing to the operator's real `~/.local/bin/loom-daemon` — this was the fix for the 2026-07-29 #4381 incident. But `loom-daemon-update.sh` resolves its provisioning destination as:

```
DEST_DIR="${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}"
PROVISION_TARGET="${LOOM_DAEMON_BIN:-$DEST_DIR/loom-daemon}"
```

`LOOM_DAEMON_BIN` takes priority over the `LOOM_DAEMON_BIN_DIR` sandbox entirely. Every live Loom agent session (Builder, Judge, Doctor, etc. — confirmed via `env | grep LOOM_DAEMON`) has `LOOM_DAEMON_BIN` pre-set to the real production path, e.g. `/home/ubuntu/.local/bin/loom-daemon`. Three tests (37/39/43, per the suite's own header comment) deliberately invoke the update script without setting `LOOM_DAEMON_BIN` explicitly, relying entirely on the `LOOM_DAEMON_BIN_DIR` export for sandboxing — but that export is silently defeated whenever the ambient/inherited environment already has `LOOM_DAEMON_BIN` set, which is the normal case for every live Loom agent.

## Reproduction

In a live Loom agent shell (where `LOOM_DAEMON_BIN` is already exported to the real binary path):

```bash
bash defaults/scripts/tests/test-loom-daemon-update.sh
```

The real `~/.local/bin/loom-daemon` gets overwritten with a fake, infinitely-looping test-fixture stub — the exact #4381 incident, reproduced twice by me while reviewing PR #4900 (unrelated to that PR's diff — this vulnerable code path predates it). The suite's own "#4381 regression guard" (checksum comparison) does catch it after the fact (fails loudly), but only after the real binary has already been overwritten.

## Impact

- **Severity**: High — corrupts the real, currently-running production daemon's on-disk binary. The live daemon process itself is unaffected until it needs to exec that path again (crash, `systemctl restart`, or an update), at which point it would launch the fake infinite-loop stub instead of the real daemon.
[...]

PR #5190, opened 2026-08-04
test(daemon): sandbox every live daemon state path behind one helper + guard

### LINKED ISSUE #5179: test-loom-daemon-update.sh writes the LIVE host's daemon pid file — LOOM_PID_FILE is the one un-isolated state path, causing a false degraded liveness verdict
## `test-loom-daemon-update.sh` writes the live host's daemon pid file

Observed on **robb-pro**, 2026-08-04. `loom-daemon health` reported:

```
liveness: degraded — daemon alive — IPC round-trip ok (pid 39126);
  STALE pid file /Users/rwalters/GitHub/loom/.loom/.daemon.pid: records pid 86050
```

The daemon was fine. The pid file was being rewritten underneath it — reading `86050`
on one sample and `3745` a minute later, while the launchd-managed daemon stayed at
`39126` for 3h35m. Something else owned that file.

`ps` explains it — the test suite was running against the live checkout:

```
92981  timeout 1200 bash .../defaults/scripts/tests/test-loom-daemon-update.sh
92982  bash .../defaults/scripts/tests/test-loom-daemon-update.sh
11684  bash .../defaults/scripts/cli/loom-daemon-update.sh        <- invoked by the test
11620  /Users/rwalters/GitHub/loom/target/debug/loom-daemon       <- test-built binary
93077  bash /var/folders/.../tmp.E42TkRtTmO/loom-daemon           <- test stub
39126  /Users/rwalters/GitHub/loom/target/release/loom-daemon     <- the REAL daemon
```

## The gap is narrow: everything else is isolated except the pid file

The test is otherwise carefully sandboxed — which is what makes this worth fixing rather
than rewriting:

- per-case `HOME` overrides (`HOME="$HOME5C"`, `HOME21`, `HOME22`, …)
- `export LOOM_AUTONOMY_MARKER="$BASE_WORKDIR/autonomy-desired"` (`:943`)
- `LOOM_LAUNCHD_LABEL="$(launchd_sandbox_new_label)"`, `LOOM_DAEMON_LAUNCHD=0`,
  `LOOM_DAEMON_SYSTEMD=0`, a per-run `LOOM_SYSTEMD_UNIT`

There is **no `LOOM_PID_FILE` override anywhere in the file**. So any path that resolves
the pid file from the real workspace rather than from the overridden `HOME` lands on the
live one.

## Why it matters beyond a cosmetic verdict

- It produces a **false `degraded` liveness verdict** on a healthy production host, which
  is exactly the signal an operator uses to decide whether a host needs intervention.
- The pid file is load-bearing for the **watchdog** (#5118 / #5126) — a component whose
  entire job is distinguishing "daemon expected and gone" from "daemon fine". Corrupting
  its input from a test is worse than the cosmetic reading suggests.
- A recycled pid could make a stale entry look *valid*, which fails in the dangerous
  direction rather than the safe one.

## This is the third instance of the same class

1. **#4087** — a daemon test booted out the *production* daemon (fixed).
2. **#5131** — something removed the live `autonomy-desired` marker while the daemon ran;
   the marker is now isolated in this very test (`:943`), so that lesson was applied.
3. **This** — the pid file, the remaining un-isolated live-state path.

Each was fixed by enumerating one more surface. That pattern suggests the durable fix is
to isolate the *whole* `LOOM_*` state surface at once rather than adding a fourth
override.

## Suggested direction

- Add `LOOM_PID_FILE` to the test's exported sandbox, pointing into `$BASE_WORKDIR`
  alongside `LOOM_AUTONOMY_MARKER` — the one-line version.
- Better: give the suite a single "sandbox every live-state path" helper (pid file,
  marker, heartbeat, socket, launchd label, systemd unit, workspace registry) so a new
  state file added to the daemon is isolated by construction rather than after it leaks.
- Consider a guard in the test harness that **fails loudly** if any live path under
  `$HOME/.loom` or a real managed workspace is written during a run — that converts this
  class from "discovered in production" to "caught in CI".

## Acceptance criteria

- [ ] Running `test-loom-daemon-update.sh` on a host with a live daemon does not modify
      that daemon's pid file, and leaves `health` reporting green liveness.
- [ ] Live-state paths are isolated via one shared helper, not per-variable overrides.
- [ ] The suite detects and fails on a write to a real `.loom` state path.
- [ ] Verified while a real daemon is running — the failure only reproduces there.

## Provenance

Found while investigating why robb-pro reported degraded liveness during a fleet
throughput review. The host was otherwise healthy: dispatch green, roles 45/45, tokens
green. Related: #4087, #5131, #5118/#5126 (the watchdog that consumes this file), #4774.

rjwalters:
## Summary

`test-loom-daemon-update.sh` could rewrite the **live** host's `.daemon.pid`, producing a false `degraded` liveness verdict for the operator and poisoning the #5118/#5126 watchdog's primary input.

Root cause (verified on a live worker host, not just from the issue text): the suite runs inside a Loom agent session, and that session **exports the live state paths into every child it spawns** —

```
LOOM_PID_FILE=/home/ubuntu/.loom/.daemon.pid     <- the REAL pid file
LOOM_WORKSPACE=/home/ubuntu/GitHub/loom          <- the REAL checkout
LOOM_DAEMON_BIN=/home/ubuntu/.local/bin/loom-daemon
```

So a missing override is not a neutral default — it is an **inherited production path**, and `LOOM_PID_FILE` is *tier 1* of the daemon's own resolver (`daemon_pidfile.rs:97-119`), ahead of every other tier. This is the identical shape #4902 fixed for `LOOM_DAEMON_BIN`, and the sibling suite `test-loom-daemon-watchdog.sh:108-118` already pins `LOOM_PID_FILE=`/`LOOM_WORKSPACE=`/`LOOM_MACHINE_CHECKOUT=` empty for exactly this reason ("an inherited `LOOM_PID_FILE` pointed every case at the [live pid file]").

Per the issue's preferred direction, this takes the **shared-helper** route rather than adding a fourth per-variable override (after #4078, #5131, #4902), and pairs it with a guard so the class is caught rather than rediscovered in production.

## Changes

**`defaults/scripts/tests/lib/live-state-sandbox.sh` (new)** — one seam for both halves:

- `live_state_sandbox_init <dir>` redirects `LOOM_PID_FILE`, `LOOM_AUTONOMY_MARKER`, `LOOM_SOCKET_PATH` (its parent is the daemon's `loom_dir`, so heartbeat + machine-level logs follow) and `LOOM_DAEMON_BIN_DIR` into the sandbox, and **unsets** the ambient `LOOM_WORKSPACE` / `LOOM_MACHINE_CHECKOUT` / `LOOM_DAEMON_BIN` that outrank each fixture's own value (`LOOM_MACHINE_CHECKOUT` additionally flips both lifecycle scripts into machine mode, whose state home is the real `$HOME/.loom`).
- `live_state_sandbox_snapshot` / `live_state_sandbox_assert_untouched` fingerprint (size + mtime + sha256) every live `.loom` state path reachable from the **ambient** env — `$HOME/.loom`, `$LOOM_WORKSPACE/.loom`, `$LOOM_MACHINE_CHECKOUT/.loom`, `$(dirname $LOOM_SOCKET_PATH)`, `$PWD`'s repo root, the suite's own checkout — before the run, and re-check them after. A write to a real state path, **or the creation of one that was absent**, fails loudly with the offender named. Volatile daemon-written files (heartbeat, log, `sweeps.json`, `activity.db`) are deliberately excluded so a healthy live daemon cannot flap the guard.

**`defaults/scripts/tests/test-loom-daemon-update.sh`** — snapshots before any export (line ~119), replaces the per-variable overrides with `live_state_sandbox_init "$BASE_WORKDIR/live-state"`, and ends with the guard as case 45. `LOOM_WATCHDOG_LABEL` stays put (supervisor identity, not a state path — that axis is owned by `lib/launchd-sandbox.sh`).

**`defaults/scripts/tests/test-live-state-sandbox.sh` (new, wired into `ci-wired.txt`)** — 18 cases proving the guard in *both* directions, because a guard that cannot be shown to fail is worthless.

- [x] **Running the suite on a host with a live daemon does not modify that daemon's pid file, and leaves `health` reporting the same liveness.** Verified twice on this host (live daemon pid 850106, machine-mode state in `~/.loom`): `~/.loom/.daemon.pid` is byte-and-mtime identical before/after both runs (`8453`, `2026-08-04 02:08:01.250588126`), the daemon is still alive, and `loom-daemon health`'s liveness line is unchanged (it was already DEGRADED **before** any test ran, from a pre-existing #4774 supervisor-relaunch staleness — the suite neither caused nor worsened it).
- [x] **Live-state paths are isolated via one shared helper, not per-variable overrides.** `live_state_sandbox_init` is the only place the update suite names a state var; a new daemon state file is isolated by adding it there once.
- [x] **The suite detects and fails on a write to a real `.loom` state path.** Case 45 of the update suite (`12 paths guarded` on this host), and proven to actually fail in `test-live-state-sandbox.sh` cases 3/4/5b — rewriting a live `~/.loom/.daemon.pid`, creating a previously-absent `<checkout>/.loom/.daemon.pid`, and the **pre-fix shape** (a write through the ambient `LOOM_PID_FILE` with no sandbox) each return rc 1 with the offending path plus before/after fingerprints. Case 6 covers "assert without a snapshot must not look green."
- [x] **Verified while a real daemon is running.** Both full-suite runs above were made against the live daemon on this host.

## Test Plan

- [x] `defaults/scripts/tests/test-loom-daemon-update.sh` — **254/254 passed** with the fix (run 1 had one unrelated flake, `--help documents --check/--dry-run/--no-restart`, under host contention; it passes on re-run and also passes on unmodified `main`).
- [x] Unmodified `main` copy of the suite run for comparison — 253/253, confirming no case regressed and that the single flake is not attributable to this change.
- [x] `defaults/scripts/tests/test-live-state-sandbox.sh` — 18/18 passed.
- [x] `defaults/scripts/tests/check-ci-suite-manifest.sh` — OK (103 suites, 100 wired, 3 excluded).
- [x] `shellcheck --severity=warning` (CI's severity) clean on all three touched/added scripts.

## Notes for the reviewer

- **Sibling suites are still exposed** (out of scope here): `test-loom-daemon-start.sh` and `test-loom-daemon-stop.sh` also have zero `LOOM_PID_FILE` overrides and pin state paths per-invocation rather than suite-wide, so an ambient value still reaches their sub-invocations. Adopting this helper there is the obvious follow-up.
- **Observed during this work** (not caused by it, guard confirms): the live `~/.loom/autonomy-desired` marker on this host disappeared **before** either suite run started (the snapshot recorded it already absent, and the guard reported no changes during the runs). That is the still-open #5131 shape on a host running many concurrent sweeps; the daemon's #4331 healing has restored it twice before per `daemon.log`.
- The main checkout was already dirty at session start (`M mcp-loom/package-lock.json`, `?? loom-tools/`) — pre-existing, untouched by this work; all changes here are in the issue worktree.

Closes #5179

### DIFF
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -61,6 +61,7 @@ test-install-source-guard.sh
 test-install-stash-scope.sh
 test-install-sync-labels.sh
 test-judge-fanout-experiment.sh
+test-live-state-sandbox.sh
 test-locate-daemon-bin.sh
 test-loom-daemon-launchd-plist.sh
 test-loom-daemon-start.sh
--- a/defaults/scripts/tests/lib/live-state-sandbox.sh
+++ b/defaults/scripts/tests/lib/live-state-sandbox.sh
@@ -1,0 +1,256 @@
+#!/usr/bin/env bash
+# live-state-sandbox.sh — ONE shared "sandbox every live daemon state path"
+# helper for the loom-daemon lifecycle test suites (issue #5179).
+#
+# ## Why this exists
+#
+# The daemon lifecycle suites deliberately execute the REAL
+# loom-daemon-start.sh / loom-daemon-stop.sh / loom-daemon-update.sh (not
+# mocks), so every state path those scripts resolve is a path the test can
+# write. Three separate incidents were each fixed by enumerating ONE more
+# surface after it leaked:
+#
+#   1. #4078/#4087 — a daemon test booted out the operator's PRODUCTION daemon
+#      (launchd label + label-blind pgrep; fixed in lib/launchd-sandbox.sh).
+#   2. #5131 — the live `autonomy-desired` marker was removed under a running
+#      daemon (fixed by a per-suite LOOM_AUTONOMY_MARKER export).
+#   3. #5179 — the live `.daemon.pid` was rewritten under a running daemon,
+#      producing a FALSE `degraded` liveness verdict for the operator and
+#      poisoning the watchdog's (#5118/#5126) primary input.
+#
+# The recurring root cause is not "one more variable was forgotten"; it is that
+# a Loom agent session EXPORTS the live paths into every child process it
+# spawns (observed on a worker host: `LOOM_PID_FILE=$HOME/.loom/.daemon.pid`,
+# `LOOM_WORKSPACE=<the real checkout>`, `LOOM_DAEMON_BIN=<the real binary>`).
+# A test suite that merely *omits* an override therefore does not get a neutral
+# default — it INHERITS the production path, and `LOOM_PID_FILE` is tier 1 of
+# the daemon's own resolver (`daemon_pidfile.rs`), ahead of every other tier.
+# #4902 fixed exactly this shape for LOOM_DAEMON_BIN; this helper generalizes
+# it so a state path added to the daemon later is isolated by construction.
+#
+# ## What it covers
+#
+# `live_state_sandbox_init` owns the filesystem STATE PATHS:
+#
+#   LOOM_PID_FILE          -> <dir>/.daemon.pid          (daemon_pidfile.rs tier 1)
+#   LOOM_AUTONOMY_MARKER   -> <dir>/autonomy-desired     (#4011/#5131)
+#   LOOM_SOCKET_PATH       -> <dir>/loom-daemon.sock     (its parent is the
+#                             daemon's `loom_dir`, so the heartbeat, machine-level
+#                             logs and every `resolve_loom_dir()` consumer follow)
+#   LOOM_DAEMON_BIN_DIR    -> <dir>/machine-level-bin-sandbox (#4381)
+#   LOOM_DAEMON_BIN        -> UNSET  (#4902: the ambient value names the REAL binary)
+#   LOOM_WORKSPACE         -> UNSET  (pid-file tier 3; each sub-invocation must
+#                             resolve its OWN fixture repo root instead)
+#   LOOM_MACHINE_CHECKOUT  -> UNSET  (pid-file tier 2 + the scripts' machine
+#                             mode, whose state home is the real $HOME/.loom)
+#
+# Supervisor IDENTITY (launchd label, systemd unit, the LOOM_DAEMON_LAUNCHD /
+# LOOM_DAEMON_SYSTEMD knobs) is intentionally NOT handled here — those are not
+# filesystem state paths and are already owned by lib/launchd-sandbox.sh and
+# the per-suite systemd/launchd knobs.
+#
+# ## The guard
+#
+# Isolation alone is unfalsifiable — the previous three fixes all *looked*
+# complete. `live_state_sandbox_snapshot` fingerprints every live state path
+# reachable from the AMBIENT environment BEFORE the sandbox is installed, and
+# `live_state_sandbox_assert_untouched` re-fingerprints them at the end of the
+# run, so a write to a real `.loom` state path (or the CREATION of one that was
+# absent) fails the suite LOUDLY instead of being discovered in production.
+#
+# Only files a healthy daemon does NOT continuously rewrite are guarded
+# (see LIVE_STATE_SANDBOX_GUARDED_FILES): `daemon.heartbeat`, `daemon.log`,
+# `sweeps.json` and `activity.db` are deliberately excluded because a live
+# daemon updates them on its own cadence, which would make the guard flap.
+#
+# ## Usage (source it)
+#
+#   source "$SCRIPT_DIR/lib/live-state-sandbox.sh"
+#   live_state_sandbox_snapshot                      # BEFORE any sandbox export
+#   ...
+#   live_state_sandbox_init "$BASE_WORKDIR/live-state"
+#   ... run the suite ...
+#   live_state_sandbox_assert_untouched              # rc 0 = clean, 1 = leaked
+
+# State files that identify/steer a daemon and are written only at lifecycle
+# transitions — safe to compare before/after even while a real daemon runs.
+LIVE_STATE_SANDBOX_GUARDED_FILES=".daemon.pid .daemon.flags autonomy-desired"
+
+# Newline-separated "<path><TAB><fingerprint>" records captured by
+# live_state_sandbox_snapshot. Empty until the snapshot runs.
+_LSS_SNAPSHOT=""
+_LSS_SNAPSHOT_TAKEN=0
+
+# ---------------------------------------------------------------------------
+# internals
+# ---------------------------------------------------------------------------
+
+# Print <path>'s mtime as an epoch second, portably: GNU `stat -c %Y` first,
+# then BSD/macOS `stat -f %m`, else `?`. Each result is validated as digits-only
+# BEFORE it is accepted — GNU stat's `-f` is "filesystem status" (it SUCCEEDS
+# with multi-line output rather than failing), so a naive `-f || -c` chain
+# silently injects newlines into the fingerprint and makes every record
+# unparseable.
+_lss_mtime() {
+    local m
+    m=$(stat -c %Y "$1" 2>/dev/null)
+    case "$m" in ''|*[!0-9]*) m=$(stat -f %m "$1" 2>/dev/null) ;; esac
+    case "$m" in ''|*[!0-9]*) m="?" ;; esac
+    printf '%s' "$m"
+}
+
+# Print <path>'s sha256, using whichever checksum tool exists (mirrors the
+# #4381 production-binary guard's tool selection).
+_lss_checksum() {
+    if command -v shasum >/dev/null 2>&1; then
+        shasum -a 256 "$1" 2>/dev/null | awk '{print $1}'
+    elif command -v sha256sum >/dev/null 2>&1; then
+        sha256sum "$1" 2>/dev/null | awk '{print $1}'
+    else
+        echo "<no-checksum-tool>"
+    fi
+}
+
+# Print a stable fingerprint of <path>: `<absent>` when it does not exist,
+# `<non-regular>` for sockets/dirs (content is meaningless there), else
+# size + mtime + checksum so ANY rewrite — including one that restores the
+# same bytes at a later mtime — is visible.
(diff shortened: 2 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T02:28:41Z) — standing down without reclaiming. Not stomping.

rjwalters:
## ✅ Approved — LGTM

Reviewed at head `81a9be56`. This is the right call on the broader-vs-minimal question: after #4087 → #5131 → #5179, a fourth per-variable override would have been the fourth instance of "enumerate one more surface after it leaks." Isolation *plus* a falsifiable guard is what actually closes the class.

### Correctness of the fix (AC1) — independently verified on a live-daemon host

Not taken on trust from the PR body. I re-ran the full suite on this worker host (live machine-mode daemon, state in `~/.loom`) and fingerprinted the real pid file myself before and after:

- before: `size=5  mtime=2026-08-04 02:08:01.250588126  md5=a1d3a9b42b8dc63cd6b1897ace488158`
- after:  **byte- and mtime-identical** (same md5, same nanosecond mtime)
- `defaults/scripts/tests/test-loom-daemon-update.sh` → **254/254 passed**, including the new case 45 (`12 paths guarded`)
- `defaults/scripts/tests/test-live-state-sandbox.sh` → **18/18 passed**

So AC1 ("does not modify that daemon's pid file") holds under direct observation, not just via the suite's own self-report. AC2 (one shared helper), AC3 (detects and fails on a write to a real path), AC4 (verified against a running daemon) are all satisfied.

### Helper design

`live_state_sandbox_init` covers the right precedence surface, and the *unset* vs *redirect* split is the subtle part it gets right: `LOOM_WORKSPACE` / `LOOM_MACHINE_CHECKOUT` / `LOOM_DAEMON_BIN` must be unset rather than re-pointed, because an ambient value outranks each fixture's own — that is precisely the #4902 shape, and `LOOM_MACHINE_CHECKOUT` additionally flips `loom-daemon-start.sh` into machine mode whose `DAEMON_STATE_HOME` is the real `$HOME/.loom` (`loom-daemon-start.sh:1237`). Worth noting for future readers: `loom-daemon-start.sh:1268` re-`export`s its *own* computed `LOOM_PID_FILE`, so for `start.sh` sub-invocations the effective protection is the `HOME`/`REPO_ROOT` + machine-checkout-unset axis, and the `LOOM_PID_FILE` redirect is what covers the direct-daemon-spawn tier-1 path (`daemon_pidfile.rs`). Both halves are needed; neither alone would have been enough. That's exactly why the guard, not the isolation, is the durable net.

Implementation details that hold up: `_lss_mtime`'s digits-only validation before accepting `stat -f` output (GNU `stat -f` *succeeds* with multi-line filesystem output, which would silently corrupt every fingerprint record — a real trap, correctly defused); `<absent>`/`<non-regular>` sentinels so an absent→present creation is caught, not just a rewrite; snapshot built via process substitution so `_LSS_SNAPSHOT` survives in the caller's shell; `_LSS_SNAPSHOT_TAKEN` so a forgotten snapshot fails loudly instead of looking green.

### Guard flapping / excluded volatile files

The exclusion is really an allowlist of three lifecycle-transition files (`.daemon.pid`, `.daemon.flags`, `autonomy-desired`), with `daemon.heartbeat` / `daemon.log` / `sweeps.json` / `activity.db` outside it — the right shape, since an allowlist can't silently grow to include a new volatile file.

Two residual flap paths, both **non-blocking** and both failing in the *safe* direction (loud, with the offending path plus before/after fingerprints named):

1. `.daemon.pid` is rewritten by the **live daemon itself** on a supervisor relaunch or self-update roll (`#4774` — the daemon claims it after socket bind). A relaunch landing inside a multi-minute suite run would be reported as a leak the suite didn't cause.
2. `autonomy-desired` is recreated by the daemon's `#4331` healing. Your own PR notes observed it vanishing on this host; a heal during a run is an absent→present transition, which the guard flags by design.

Neither can occur in CI (hermetic, no daemon), so this is a live-host-only diagnosis cost, not a CI flake. If it ever bites, a cheap discriminator would be "did the after-value name a live process that is the current daemon?" — worth folding into #5191 rather than blocking here.

Minor coverage note: `loom-daemon-start.sh` also writes `$DAEMON_STATE_HOME/logs/daemon-start.log` (and `mkdir -p`s that dir), which is unguarded. Genuinely volatile, so excluding it is defensible — just be aware a leak that only appends there stays invisible.

### Test coverage

`test-live-state-sandbox.sh` is the part that makes this more than a hopeful fix: it proves the guard in **both** directions rather than only asserting the happy path. Case 5/5b is the strongest pair — the same write, with and without `live_state_sandbox_init`, rc 0 vs rc 1, so the pre-fix production shape is demonstrably caught. Case 6 (assert without snapshot must not look green) and case 8 (non-vacuous snapshot size) close the two ways a guard like this usually rots into a no-op. Case 7 pins the volatile-exclusion behavior so it can't be tightened accidentally.

Hermeticity is right: fake `$HOME`, fake checkout, ambient `LOOM_*` neutralized per case, and the one case that touches a "live" path uses a synthetic `$WORKDIR/ambient-live/.daemon.pid`.

### CI wiring

`ci-wired.txt` addition is correct and sufficient (`run-ci-suites.sh` consumes the manifest; `check-ci-suite-manifest.sh` enforces the wired/excluded partition). Confirmed the suite actually executes in CI, not just that it's listed:

```
2026-08-04T02:27:51Z PASS  test-live-state-sandbox.sh   1s
```

`lib/live-state-sandbox.sh` correctly stays mode `100644` (sourced, not a suite) and so is ignored by the manifest check, matching `lib/launchd-sandbox.sh` / `lib/bg-proc-trap.sh`.

### Gates

- All checks green: Shell Test Suites (hermetic) **pass** (4m48s), Shellcheck pass, Docs/Defaults Parity pass, Installer Integration pass, AGENTS.md/CLAUDE.md budget pass.
- `mergeStateStatus`: **CLEAN**.
- Scope discipline: sibling exposure in `test-loom-daemon-start.sh` / `test-loom-daemon-stop.sh` correctly deferred to #5191 (filed with `loom:triage`, proper intake) rather than sprawled into this PR.

Nothing blocking. Approving.

PR #5176, opened 2026-08-04
fix(guard): thread a cd-prefix into stash-scope cwd resolution

rjwalters:
## Summary

`.loom/hooks/guard-destructive-generic.sh`'s `stash-scope:main-checkout` check
(and its `stash-scope:worktree-collision` sibling, #4821) resolved
`_stash_toplevel` / `_stash_common_parent` purely from the hook's raw session
`$CWD`, ignoring any `cd <dir> &&` prefix earlier in the same compound
`$COMMAND` string. A command of the form `cd .loom/worktrees/issue-N && git
stash pop` — hook session cwd still the main repo root, the common shape per
this repo's own CLAUDE.md worktree workflow — resolved scope against the
main checkout instead of the cd target, and asked as if the operation
targeted the main checkout.

This is the same class of bug already fixed for `parse_force_ops()` in
#5156/PR #5161 ("thread a cd-prefix into force-op cwd/branch-identity
resolution"); this PR mirrors that fix's cd-tracking approach for the
stash-scope block.

## Changes

- `defaults/hooks/guard-destructive-generic.sh`: adds `resolve_stash_cwd()`,
  a new helper mirroring `parse_force_ops()`'s cd-tracking (which itself
  mirrors `extract_write_targets()`, #4933/#4881). It threads a `cd <dir>`
  prefix through the compound command and returns the cwd in effect at the
  `git stash pop/drop/clear` segment. The `stash-scope:main-checkout` block
  now resolves `_stash_toplevel`/`_stash_common_parent` from this
  cd-threaded effective cwd instead of the raw `$CWD`. The
  `stash-scope:worktree-collision` check (#4821) consumes the same shared
  `_stash_toplevel`/`_stash_common_parent` values, so it gets the same
  treatment automatically. A new branch handles the case where the cd
  target cannot be resolved to a git checkout (does not exist, or is not a
  repo) — it now asks explicitly (`stash-scope:cd-unresolved`) rather than
  silently falling through to allow, matching `parse_force_ops`'
  detached-HEAD fail-safe.
- `.loom/hooks/guard-destructive-generic.sh`: same change, kept byte-identical
  to the canonical `defaults/` copy.
- `tests/hooks/test-guard-destructive.sh`: new regression fixtures using a
  real linked worktree repo (mirrors #5161's fixture pattern) — cd into a
  worktree then stash pop/drop/clear (hook cwd = main root) allows; the same
  op with hook cwd already at the worktree allows; a chained read-only prefix
[...]

Issue #5172, opened 2026-08-04
Guard false positive: merge-redirect guard denies gh api -f body=VAR referencing a heredoc that merely quotes the disallowed phrase


rjwalters:
## Guard-Decision Telemetry Review (#3898 standing policy)

Reviewed `.loom/logs/guard-decisions.log` during a routine main-branch audit
(2026-08-04). The `loom:gh-pr-merge-redirect` guard pattern (tier: catastrophic)
is still firing on a shape not covered by the recent #5115/#5155/#5160 fix
series.

## Reproduction (live, during this audit session)

Occurrence at `2026-08-04T00:17:16Z` in the log — a `gh api
"repos/rjwalters/loom/issues/5160/comments" -f body="$BODY"` call was denied,
where `$BODY` was assigned earlier in the same command from a `cat <<'EOF' ...
EOF` heredoc. The heredoc content was a PR review comment (approving PR #5160,
the very fix for this guard) that quotes the disallowed three-word CLI phrase
several times as prose describing test cases — never an actual invocation.

Confirmed still reproducible against the current guard (post-#5160, which
merged one minute after this denial) with a minimal reconstruction:

```bash
BODY="$(cat <<'EOF2'
Tested: raw gh pr merge 123 denied; echo "gh pr merge 123" | bash denied.
EOF2
)"
gh api "repos/rjwalters/loom/issues/NNN/comments" -f body="$BODY"
```

This is still denied outright.

## Why this differs from the prior fixes

- #5115 added `mask_data_flag_values`, which masks the quoted value of known
  text-only flags (`-m`/`--message`/`--body`/`--notes`/`--title`/`--comment`/
  `--search`). `gh api`'s field syntax is `-f key=value` (here `-f
  body=$BODY`), not a `--body VALUE` flag — a different shape the masking
  apparently doesn't recognize.
- Even if the flag shape were recognized, the value here is a **variable
  reference** (`$BODY`) to a string assigned earlier via heredoc, not an
  inline literal. The false-positive class from #5109/#5115/#5155 was always
  about literal text in the same argument position; this is a two-hop
  indirection (heredoc assignment -> later variable reference) that raw
  substring scanning still catches because the heredoc body's literal text is
  present somewhere in the full command string.

## Recommended Fix

Extend the masking to recognize `gh api ... -f <field>=<value>` for known
text-bearing fields (`body`, `message`, `comment`, similar), and/or extend
[...]

Issue #5158, opened 2026-08-03
Guard false positive: catastrophic curl-pipe-shell pattern fires on grep introspection of its own regex


rjwalters:
## Guard-Decision Telemetry Review (#3898 standing policy)

Reviewed `.loom/logs/guard-decisions.log` during a routine main-branch audit
(2026-08-03). One `catastrophic:curl .* \| .*sh` denial is a read-only false
positive.

## Occurrence (2026-07-27T21:31:03Z)

```
grep -n "curl .*|" defaults/hooks/guard-destructive.sh | head
echo ---
grep -nE "^# ===|^# ---|^##|SECTION|LOOM_RM_SCOPE|rm-scope|guards\.(sqlDdl|cloudCli|reversibleGh|rmScope|forceScope)" defaults/hooks/guard-destructive.sh | head -40
```

This is a `grep` search over the guard's own source file, looking for where the
curl-pipe pattern is defined/documented — an introspection/code-reading command
that cannot execute a curl-pipe-to-shell under any circumstance. It was denied
because the grep *pattern argument* (`"curl .*|"`) itself textually resembles
the disallowed shape, even though `grep` never invokes what it searches for.

## Root Cause

Same class as #5155/#5109 (inert text misread as a live invocation) but for the
`catastrophic:curl .* | .*sh` pattern in the generic destructive-command guard
rather than the merge-redirect guard. The guard's regex scan does not
distinguish "this text appears as a quoted pattern argument to a known
non-executing command (`grep`/`rg`/`ripgrep`)" from "this shell actually pipes
curl output to a shell interpreter."

## Recommended Fix

Extend whatever masking/allowlist mechanism is used for the merge-redirect
guard (or the confinement guard, see the sibling issue filed this session) to
also cover `catastrophic:curl .* | .*sh`: when the entire matched span is
inside a quoted pattern argument to `grep -e`/`-E`/positional pattern, or to
`rg`, treat it as inert search text rather than a live pipeline.

## Impact

Very low frequency (1 occurrence), but blocks routine source-reading/audit work
(grepping a guard script to understand its own rules) with no human present to
unblock.

---
Discovered during main branch audit via the guard-decision telemetry review
standing policy (#3898).

---

### Confirmed Root Cause

Verified against `.loom/hooks/guard-destructive-generic.sh` (identical byte-for-byte
to `defaults/hooks/guard-destructive-generic.sh` as of `8cb78699`, 2026-08-04).

[...]

PR #5192, opened 2026-08-04
fix(guard): scope gh-api-rawfield-body-literal-at to a heredoc-masked command copy

### LINKED ISSUE #5181: Guard false positive: gh-api-rawfield-body-literal-at fires on heredoc text that merely quotes the denied command

(2026-08-03/04, commit 11aa4a1d). `gh-api-rawfield-body-literal-at` fired 7
times at **catastrophic** tier — a tier with no human to answer in headless
runs, so every hit is a hard stall.

## What's actually being denied

The check (`defaults/hooks/guard-destructive-generic.sh` / `.loom/hooks/guard-destructive-generic.sh`,
around line 2152-2154) is:

```
GH_API_RAWFIELD_BODY_AT_PATTERN="(^|[;&|[:space:]])gh[[:space:]]+api[^;&]*[[:space:]](-f|--raw-field)[[:space:]]*=?[[:space:]]*[\"']?body=[\"']?$GH_AT_PATHISH"
if echo "$COMMAND" | grep -qE "$GH_API_RAWFIELD_BODY_AT_PATTERN"; then
    deny "..." "gh-api-rawfield-body-literal-at"
fi
```

This greps the **raw** `$COMMAND` string. It correctly catches a live `gh api
... -f body=@path` invocation (the real #4523/#4601 data-loss shape — that
part of the guard is correct and should stay flagged). But it also fires when
that exact substring merely appears as **quoted text inside a heredoc being
written to a file**, with no `gh api` command actually executing. Example
from the log (redacted to the relevant fragment):

```
cat > /tmp/some-report.md << 'EOF'
...discussing the anti-pattern, e.g. quoting...
gh api repos/OWNER/REPO/issues/N/comments -f body=@/tmp/review.md
...as an example of what NOT to do...
EOF
./.loom/scripts/create-issue.sh --title "..." --body-file /tmp/some-report.md ...
```

Nothing here executes the denied `gh api` call — the string is inert prose
inside a heredoc destined for a markdown file. The guard still denies the
whole command at catastrophic tier.

This is the same false-positive class already fixed for two sibling guards:

- #5158 — `catastrophic:curl .* \| .*sh` fires on grep introspection of its
  own regex
- #5172 — the merge-redirect guard denies `gh api ... -f body=VAR` when a
  heredoc merely quotes the disallowed phrase

Other checks in the same file use a "comment-stripped working copy" of
`$COMMAND` (see the block starting around line 2158, `# COMMENT-STRIPPED
WORKING COPY`) specifically to avoid scanning inert text. The
`gh-api-rawfield-body-literal-at` check runs *before* that stripping and
operates on the raw command, so it doesn't get that protection.

## Confirmed impact

- A prior agent's own attempt to file this exact finding was itself denied
  by this same guard (visible in `.loom/logs/guard-decisions.log`,
  `2026-08-04T01:31:50Z`) — the guard blocked the report about itself,
  because the report's heredoc body quoted the denied command as an example.
  No issue for this pattern existed before this one.
- Judge/Champion review comments that legitimately discuss or quote the
  anti-pattern (e.g. explaining to a human reviewer why a prior draft was
  wrong) can hit the same wall.

## Recommended outcome: refine (not remove)

Keep denying an actual `gh api ... -f body=@path` invocation — that part is
correct and matches a real, previously-shipped data-loss bug (#4523/#4601).
Scope the check to the same comment/heredoc-stripped view of `$COMMAND` the
other checks below it already use, so quoted example text inside a heredoc
stops tripping catastrophic-tier denial. Filed under the Guard-Decision
Telemetry Review standing policy (#3898); same remediation shape as #5158
and #5172.

rjwalters:
## Summary
The `gh-api-rawfield-body-literal-at` catastrophic-tier guard check in
`defaults/hooks/guard-destructive-generic.sh` grepped the raw `$COMMAND`
string, so it fired not only on a live `gh api ... -f/--raw-field
body=@<path>` invocation (the real #4523/#4601 data-loss shape) but also
whenever that exact substring merely appeared as quoted example text inside
a heredoc destined for a file (e.g. a report explaining the anti-pattern).
Nothing executed in that case, yet the guard denied the whole command at
catastrophic tier — a hard stall in headless runs since there is no human to
answer. This mirrors the same false-positive class already tracked for
sibling checks (#5158, #5172).

## Changes
- Build a heredoc-body-masked working copy of `$COMMAND`
  (`COMMAND_HEREDOC_MASKED`), reusing the existing `mask_heredoc_bodies()`
  primitive (added by #5000, already used by `extract_write_targets()`), built
  lazily only when `<<` is present in the command.
- Scope the `gh-api-rawfield-body-literal-at` check to scan
  `COMMAND_HEREDOC_MASKED` instead of raw `$COMMAND`. Every other check in
  the file is untouched.
- Apply the identical diff to both `defaults/hooks/guard-destructive-generic.sh`
  (source of truth) and the installed `.loom/hooks/guard-destructive-generic.sh`
  copy — both files were byte-identical before this change and are kept in
  lockstep, matching the pattern used by recent sibling guard PRs (e.g. #5176,
  #5161) that touch this same file.
- Add three regression tests to `tests/hooks/test-guard-destructive.sh`:
  1. a heredoc body that merely quotes the denied phrase as prose now allows,
  2. a live (non-heredoc) `gh api ... -f body=@path` invocation still denies,
  3. a real invocation sitting after an unrelated heredoc in the same
     multi-line command still denies (narrows, never widens — mirrors the
     #5000 "narrows, never widens" test pattern).

| `gh-api-rawfield-body-literal-at` no longer denies a command where the denied substring appears only inside a heredoc body destined for a file | Passed | New `assert_allow` test reproduces the exact false-positive shape from the issue; guard now allows it |
| A live, non-heredoc `gh api ... -f/--raw-field body=@<path>` invocation still denies at catastrophic tier, unchanged | Passed | Existing `#4601`/`#4685` tests (lines 535-597, 979-980) still pass unmodified; new dedicated `#5181` regression test also confirms |
| A real (non-heredoc) invocation sitting in the same multi-line command as an unrelated heredoc still denies | Passed | New `#5181` "narrows, never widens" test |
| `tests/hooks/test-guard-destructive.sh` passes locally | Passed | Full suite: 679/679 passed |

## Test Plan
- `bash tests/hooks/test-guard-destructive.sh` — 679/679 passed, including
  the 3 new `#5181` cases.
- `bash tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 passed
  (unaffected by this change, run to confirm no dispatcher regression).
- `bash -n defaults/hooks/guard-destructive-generic.sh` and `shellcheck -S
  warning` — clean.
- Manual repro: reconstructed the exact false-positive shape from the issue
  (a `cat > report.md <<'EOF' ... EOF` heredoc whose body quotes the denied
  `gh api` phrase as prose) against the patched guard directly — now allows.
  Reconstructed a live invocation, and a live invocation immediately after an
  unrelated heredoc in the same command — both still deny.
- Confirmed `defaults/hooks/guard-destructive-generic.sh` and
  `.loom/hooks/guard-destructive-generic.sh` remain byte-identical after the
  edit.

Closes #5181

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -2206,9 +2206,32 @@ if [[ "$COMMAND" == *"@"* ]]; then
     # `repos/{o}/{r}/pulls/{n}` (PR PATCH) exactly as it covers
     # `.../issues/{n}/comments`. Confirmed via the test suite's new
     # non-comments-endpoint case; no widening was needed here.
+    #
+    # HEREDOC-MASKED SCAN (#5181): this check used to grep raw $COMMAND, so a
+    # heredoc BODY line that merely QUOTES the denied phrase as inert prose
+    # (e.g. a report `cat > f.md <<'EOF' ... gh api ... -f body=@x ... EOF`,
+    # nothing of which executes) tripped the same catastrophic-tier deny as a
+    # live invocation. Reuses mask_heredoc_bodies() (#5000, defined above)
+    # -- the same primitive extract_write_targets() already uses -- to build
+    # a heredoc-body-blanked working copy and scans THAT instead. Built
+    # lazily (only when '<<' is present, mirroring the COMMAND_NO_COMMENT
+    # `#`-present hot-path guard below) and scoped to just this one check;
+    # every other rule in this file keeps reading raw $COMMAND /
+    # COMMAND_NO_COMMENT unchanged. Masking only ever narrows: a REAL
+    # (non-heredoc) invocation is untouched and still denies, even sitting in
+    # the same multi-line command as an unrelated heredoc (mirrors the #5000
+    # "narrows, never widens" test at
+    # tests/hooks/test-guard-destructive.sh:2691).
     # -------------------------------------------------------------------------
+    if [[ "$COMMAND" == *"<<"* ]]; then
+        COMMAND_HEREDOC_MASKED=$(printf '%s' "$COMMAND" | awk "$_MASKHEREDOC_AWK"'
+        { buf = buf (NR > 1 ? "\n" : "") $0 }
+        END { printf "%s", mask_heredoc_bodies(buf) }')
+    else
+        COMMAND_HEREDOC_MASKED="$COMMAND"
+    fi
     GH_API_RAWFIELD_BODY_AT_PATTERN="(^|[;&|[:space:]])gh[[:space:]]+api[^;&]*[[:space:]](-f|--raw-field)[[:space:]]*=?[[:space:]]*[\"']?body=[\"']?$GH_AT_PATHISH"
-    if echo "$COMMAND" | grep -qE "$GH_API_RAWFIELD_BODY_AT_PATTERN"; then
+    if echo "$COMMAND_HEREDOC_MASKED" | grep -qE "$GH_API_RAWFIELD_BODY_AT_PATTERN"; then
         deny "BLOCKED: 'gh api ... -f/--raw-field body=@<path>' does NOT read the file — only -F/--field gives '@<path>' its read-from-file meaning. As written this posts the literal string '@<path>' as the body (same silent data loss as PR #4457/issue #4608). Use '-F body=@<path>' instead." "gh-api-rawfield-body-literal-at"
     fi
 fi
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -2206,9 +2206,32 @@ if [[ "$COMMAND" == *"@"* ]]; then
     # `repos/{o}/{r}/pulls/{n}` (PR PATCH) exactly as it covers
     # `.../issues/{n}/comments`. Confirmed via the test suite's new
     # non-comments-endpoint case; no widening was needed here.
+    #
+    # HEREDOC-MASKED SCAN (#5181): this check used to grep raw $COMMAND, so a
+    # heredoc BODY line that merely QUOTES the denied phrase as inert prose
+    # (e.g. a report `cat > f.md <<'EOF' ... gh api ... -f body=@x ... EOF`,
+    # nothing of which executes) tripped the same catastrophic-tier deny as a
+    # live invocation. Reuses mask_heredoc_bodies() (#5000, defined above)
+    # -- the same primitive extract_write_targets() already uses -- to build
+    # a heredoc-body-blanked working copy and scans THAT instead. Built
+    # lazily (only when '<<' is present, mirroring the COMMAND_NO_COMMENT
+    # `#`-present hot-path guard below) and scoped to just this one check;
+    # every other rule in this file keeps reading raw $COMMAND /
+    # COMMAND_NO_COMMENT unchanged. Masking only ever narrows: a REAL
+    # (non-heredoc) invocation is untouched and still denies, even sitting in
+    # the same multi-line command as an unrelated heredoc (mirrors the #5000
+    # "narrows, never widens" test at
+    # tests/hooks/test-guard-destructive.sh:2691).
     # -------------------------------------------------------------------------
+    if [[ "$COMMAND" == *"<<"* ]]; then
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T02:28:29Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Well-scoped fix for the #5181 false-positive class (heredoc-body prose tripping the `gh-api-rawfield-body-literal-at` catastrophic-tier guard).

- Reuses the existing `mask_heredoc_bodies()` primitive (#5000) — same pattern already used by `extract_write_targets()` and the stash-scope masking at line ~2508 — rather than inventing new masking logic.
- Scope is tight: only the `gh-api-rawfield-body-literal-at` check reads `COMMAND_HEREDOC_MASKED`; every other check in the file is untouched.
- Regression coverage is solid: allow-case for the heredoc-prose false positive, plus two "narrows, never widens" deny-cases (a standalone live invocation, and a live invocation following an unrelated heredoc in the same command) — mirrors the #5000 test pattern.
- `defaults/hooks/guard-destructive-generic.sh` and `.loom/hooks/guard-destructive-generic.sh` confirmed byte-identical after the edit.

## Test Execution

- `bash tests/hooks/test-guard-destructive.sh` — ✅ Executed: 679/679 passed (including the 3 new #5181 cases)
- `bash tests/hooks/test-guard-destructive-dispatcher.sh` — ✅ Executed: 12/12 passed
- `bash -n defaults/hooks/guard-destructive-generic.sh` — ✅ Executed: syntax OK
- `shellcheck -S warning` — ✅ Executed: clean
- Manual repro of the false-positive/regression shapes — ⚠️ Skipped: already covered by the equivalent automated test cases above

CI was pending on "Shell Test Suites (hermetic)" when I started; polled to completion (green in ~5m11s) before approving. `mergeStateStatus` is `CLEAN`.

rjwalters:
Judge follow-up pass on this PR: by the time I reached a verdict, the PR had
already been approved (`loom:pr`, comment above) and merged (`e33d254a`) by
another Judge pass. Per `judge.md`'s Verdict-Time CAS Recheck, I'm not
overwriting a landed verdict — but I want to flag a concrete regression this
change introduces that the prior review's automated test coverage (and its
"skipped, already covered by tests" manual-repro note) did not catch.

**Confirmed independently (before and after `e33d254a`, same guard file, absolute paths):**

- The false-positive fix itself is correct and well-scoped: `COMMAND_HEREDOC_MASKED`
  is read only by the one `gh-api-rawfield-body-literal-at` check; every other
  check still reads raw `$COMMAND`/`COMMAND_NO_COMMENT` unchanged. The 3 new
  tests pass and correctly validate the shapes they target.
- However, `mask_heredoc_bodies()` (#5000) documents its own "Known Limitation
  #1 — interpreter-fed heredocs" in its header comment: a heredoc body fed to
  an interpreter (`bash <<'EOF' ... EOF`, `sh -s <<EOF ... EOF`, `... | bash`)
  is genuinely LIVE code to that inner interpreter even though the outer shell
  never parses it as redirection syntax. That limitation was deliberately
  scoped to `extract_write_targets()`'s ASK-tier use (#5117 — "DECISION
  (recorded, not implemented in this pass)"), where fail-open just means a
  missed ask. This PR reuses the same primitive for a **catastrophic-tier
  DENY** without re-evaluating whether that tradeoff still holds — it does
  not: a live `bash <<'EOF' ... gh api ... -f body=@path ... EOF` command,
  which genuinely executes the denied pattern once `bash` reads its heredoc
  stdin, now **silently ALLOWs** on current `main` where it correctly DENIED
  before this merge. Reproduction and suggested fix directions filed as #5198.

Not asking for a revert reflexively — the original false positive was real
and worth fixing — but this specific gap (a live, executable instance of the
exact #4523/#4601/#4685 data-loss pattern now slipping past a catastrophic
guard) seems worth prioritizing given what this check exists to protect.

PR #5168, opened 2026-08-04
test: scrub inherited LOOM_RUNTIME/LOOM_WORKSPACE in dispatcher 'no env' cases

rjwalters:
## Summary

`defaults/scripts/tests/test-loom-dispatcher.sh` Test 19 and Test 21 build fixture
repos and `cd` into them to assert on the `loom sweep` runtime-resolution
default/config path, but never scrub an inherited `LOOM_RUNTIME`/`LOOM_WORKSPACE`
from the calling shell. Every Loom-spawned agent session exports both
(`.loom/scripts/spawn-claude.sh`), and `spawn-worker.sh`'s `_resolve_workspace()`
trusts `LOOM_WORKSPACE` over `cwd` when set — so with the ambient env inherited,
the dispatcher reads the *caller's* `.loom/config.json` instead of the fixture's
and resolves the wrong runtime, producing spurious `FAIL`s (97/99 or 94/99
depending on which vars leak) that are indistinguishable at a glance from a real
runtime-dispatch regression.

## Changes

- `defaults/scripts/tests/test-loom-dispatcher.sh`: Test 19 ("default path (no
  env, no config)") and Test 21 ("honors `.loom/config.json` `runtimes.default`
  (no env)") now invoke the dispatcher with `env -u LOOM_RUNTIME -u
  LOOM_WORKSPACE` so the assertions are hermetic regardless of the caller's
  ambient environment.
- Scanned the rest of the file for other "no env" cases in the runtime-dispatch
  section (Tests 20, 22–24 all deliberately set `LOOM_RUNTIME` or assert
  unrelated behavior) — no other gaps of the same class found in this file.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Test 21 ("no env") passes regardless of inherited `LOOM_RUNTIME`/`LOOM_WORKSPACE` | ✅ | Ran with `LOOM_RUNTIME=claude LOOM_WORKSPACE=<repo-root>` exported (simulating a Loom-spawned session) before/after the fix — failures disappear after the fix |
| Other "no env" gaps in this file scanned and fixed | ✅ | Test 19 had the same latent gap (previously "passed" only by coincidence, since the default runtime happens to be `claude`); fixed alongside Test 21. No other runtime-dispatch "no env" cases found in this file. |
[...]

PR #5199, opened 2026-08-04
test(daemon): isolate the daemon's live-state-file surface via one sandbox helper, add a write guard

### LINKED ISSUE #5179: test-loom-daemon-update.sh writes the LIVE host's daemon pid file — LOOM_PID_FILE is the one un-isolated state path, causing a false degraded liveness verdict

rjwalters:
## Summary
`test-loom-daemon-update.sh` sandboxed the autonomy marker, launchd label, and systemd unit individually, but had no `LOOM_PID_FILE` override anywhere. Any code path resolving the daemon pid file therefore fell through to the real workspace, and on a host with a live daemon this test suite corrupted its `.daemon.pid` file — producing a false "degraded" liveness verdict on a healthy production host (observed on robb-pro). This is the third instance of the same missed-isolation-surface bug class, after #4087 (booted out the real daemon) and #5131 (removed the live autonomy marker).

## Changes
- Add `defaults/scripts/tests/lib/live-state-sandbox.sh`: one shared helper (`live_state_sandbox_init`) that exports `LOOM_PID_FILE`, `LOOM_AUTONOMY_MARKER`, `LOOM_SOCKET_PATH`, and `LOOM_WORKSPACES_PATH` together into a single sandbox directory (instead of per-variable overrides scattered across the test file), plus `live_state_sandbox_snapshot` / `live_state_sandbox_report_changes` primitives that fingerprint the daemon's known live-state files (`.daemon.pid`, `autonomy-desired`, `daemon.heartbeat`, `workspaces.json`) under a given root.
- The heartbeat file is fingerprinted by its `pid=` token rather than a full content hash: a real, healthy daemon rewrites its own heartbeat's timestamp on a routine ~60s cadence, so a byte-identical requirement would false-fail on any run long enough to span a real tick — exactly the "verify while a real daemon is running" scenario this guard must support.
- Wire the helper into `test-loom-daemon-update.sh`:
  - Replace the old single-variable `LOOM_AUTONOMY_MARKER` export with one `live_state_sandbox_init "$BASE_WORKDIR"` call.
  - Snapshot every well-known live-state path under the real `$HOME` and under the checkout this suite itself lives in (`$LOOM_REPO_ROOT`), right after `LOOM_REPO_ROOT` is computed — before any fixture is built or scenario runs.
  - Add a new "45. Live-state path guard" assertion at the end of the suite that re-snapshots and fails loudly (printing a diff) if any real `.loom` live-state path was created or modified during the run — converting this bug class from "discovered in production" to "caught in CI".

| Running the suite on a host with a live daemon does not modify that daemon's pid file, and leaves `health` reporting green liveness | ✅ | Ran the full suite on this host while a real `loom-daemon` (pid 39126) was live. The new guard's before/after snapshot of `.daemon.pid` under both `$HOME/.loom` and the checkout's own `.loom` was unchanged (`<absent>` both times on this host). |
| Live-state paths are isolated via one shared helper, not per-variable overrides | ✅ | `lib/live-state-sandbox.sh`'s `live_state_sandbox_init` is the single call site; the prior standalone `LOOM_AUTONOMY_MARKER` export was folded into it. |
| The suite detects and fails on a write to a real `.loom` state path | ✅ | Verified directly: before the heartbeat-specific fingerprint fix, the guard correctly FAILED because `~/.loom/daemon.heartbeat`'s content hash changed (the live daemon's routine 60s heartbeat write) — proving the detection mechanism works. After switching the heartbeat fingerprint to its `pid=` token, the guard correctly passes (same owning process, no leak) while remaining sensitive to a different PID appearing in that file. |
| Verified while a real daemon is running — the failure only reproduces there | ✅ | All verification above was performed on this host with a genuine live daemon (pid 39126) running throughout the ~9-minute suite run. |

## Test Plan
- `bash -n` + `shellcheck -x` on both changed/new files: clean.
- `bash defaults/scripts/tests/test-loom-daemon-update.sh`: **256/256 tests pass**, including the new "45. Live-state path guard" assertion, run while a real live daemon was active on the host.
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh`: passes (99 wired / 3 excluded) — `lib/` helpers are intentionally exempt from the wired/excluded partition.
- No `.rs` files were touched by this change (pure bash test infrastructure), so the repo's `buildGate.command` (`bash .loom/scripts/build-gate.sh`, which runs the full `cargo test --workspace --lib --bins`) was started but not waited to completion: it stalled for several minutes inside unrelated `main_health_gate::tests::*` cases that perform real local git fetch/push cycles to scratch bare repos — a pre-existing slow/host-dependent area with zero code-path relationship to this PR's `defaults/scripts/tests/` changes. Flagging per builder guidance rather than expanding scope to investigate an unrelated module.

Closes #5179

### DIFF
--- a/defaults/scripts/tests/lib/live-state-sandbox.sh
+++ b/defaults/scripts/tests/lib/live-state-sandbox.sh
@@ -1,0 +1,162 @@
+#!/usr/bin/env bash
+# live-state-sandbox.sh — shared test sandbox for the loom-daemon's live-state
+# file surface (issue #5179).
+#
+# Why this exists: this is the THIRD time a daemon lifecycle test has leaked
+# into a real, live daemon's on-disk state because one more state file went
+# un-isolated:
+#
+#   1. #4087 — a daemon test booted out the operator's real, running daemon.
+#   2. #5131 — a test removed the live `autonomy-desired` marker while a real
+#      daemon was running (fixed by exporting a scoped LOOM_AUTONOMY_MARKER).
+#   3. #5179 — a test rewrote the LIVE host's `.daemon.pid` while a real
+#      daemon was running (no `LOOM_PID_FILE` override existed anywhere),
+#      producing a false "degraded" liveness verdict on a healthy host.
+#
+# Each prior fix enumerated ONE more surface as a per-variable override at the
+# test's sandbox-setup call site. That pattern guarantees a fourth, fifth,
+# ... Nth instance: every NEW daemon state file (a future heartbeat variant, a
+# new registry, whatever comes next) is un-isolated by default until it is
+# individually discovered leaking in production. This helper inverts that:
+# ONE call isolates the WHOLE known live-state surface at once, so a state
+# file this daemon does not have yet only needs one new line added HERE, not
+# a new ad hoc override hunted down and added to the test file itself.
+#
+# Covers (env var -> what it isolates, and which script/module resolves it):
+#   LOOM_PID_FILE        -> `.daemon.pid`         (loom-daemon's daemon_pidfile.rs,
+#                            tier 1 -- an explicit override the daemon process
+#                            itself always honors, though NOTE: the daemon
+#                            lifecycle SHELL scripts recompute + re-export this
+#                            unconditionally from $DAEMON_STATE_HOME rather than
+#                            honoring an inbound value, so this export is a
+#                            backstop for anything that resolves the pid file
+#                            without going through those scripts -- the
+#                            existence check below is what actually catches a
+#                            shell-script-level leak)
+#   LOOM_AUTONOMY_MARKER -> `autonomy-desired`    (autonomy_marker.rs / the
+#                            lifecycle scripts' INTENT_MARKER, both honor an
+#                            inbound override)
+#   LOOM_SOCKET_PATH     -> the daemon socket, AND (as its directory) the
+#                            default resolution root for `daemon.heartbeat`
+#                            and the marker's own un-overridden default
+#                            (daemon_heartbeat.rs / autonomy_marker.rs /
+#                            loom-daemon-start.sh's `LOOM_DIR`)
+#   LOOM_WORKSPACES_PATH -> `workspaces.json`, the multi-workspace registry
+#                            (loom-daemon/src/workspace_registry.rs)
+#
+# Deliberately NOT covered here (already isolated by dedicated, existing
+# helpers/exports a sourcing test is expected to keep using alongside this
+# one): the launchd label (`lib/launchd-sandbox.sh`'s
+# `launchd_sandbox_new_label`) and the systemd unit name (the sourcing test's
+# own scratch `LOOM_SYSTEMD_UNIT` export) -- both are already
+# per-suite-scoped identifiers, not files under a `.loom` directory, so they
+# do not fit this helper's "one directory, one snapshot" model. This helper's
+# job is the remaining live-STATE-FILE surface.
+#
+# Usage (source it):
+#   source "$SCRIPT_DIR/lib/live-state-sandbox.sh"
+#   live_state_sandbox_init "$BASE_WORKDIR"   # exports the 4 vars above
+#   ...
+#   before="$(live_state_sandbox_snapshot "$HOME" "$LOOM_REPO_ROOT")"
+#   ... run the whole suite ...
+#   after="$(live_state_sandbox_snapshot "$HOME" "$LOOM_REPO_ROOT")"
+#   live_state_sandbox_report_changes "$before" "$after"   # empty == untouched
+
+# live_state_sandbox_init <workdir>
+#
+# Exports LOOM_PID_FILE, LOOM_AUTONOMY_MARKER, LOOM_SOCKET_PATH, and
+# LOOM_WORKSPACES_PATH so every one of them resolves under <workdir> --
+# regardless of which per-scenario $HOME/$PWD a later sub-invocation uses.
+# <workdir> is created if it does not already exist.
+live_state_sandbox_init() {
+    local workdir="$1"
+    mkdir -p "$workdir"
+    export LOOM_PID_FILE="$workdir/.daemon.pid"
+    export LOOM_AUTONOMY_MARKER="$workdir/autonomy-desired"
+    export LOOM_SOCKET_PATH="$workdir/daemon.sock"
+    export LOOM_WORKSPACES_PATH="$workdir/workspaces.json"
+}
+
+# The basenames of every well-known daemon live-state file under a
+# `<root>/.loom` directory. Kept as ONE list (see the module doc above) so a
+# newly added daemon state file is added in exactly one place.
+_LIVE_STATE_SANDBOX_FILENAMES=".daemon.pid autonomy-desired daemon.heartbeat workspaces.json"
+
+# live_state_sandbox_real_state_paths <root>
+#
+# Echoes the well-known live-state file paths under <root>/.loom, one per
+# line. <root> is normally $HOME (the machine-level state a real, host-wide
+# daemon uses) and/or the checkout this suite itself lives in (the exact
+# directory a "repo mode" daemon uses -- and the directory whose `.daemon.pid`
+# was corrupted in the #5179 incident).
+live_state_sandbox_real_state_paths() {
+    local root="$1" name
+    for name in $_LIVE_STATE_SANDBOX_FILENAMES; do
+        echo "$root/.loom/$name"
+    done
+}
+
+# _live_state_sandbox_fingerprint <path>
+#
+# "<absent>" when the path does not exist, else a fingerprint of its content.
+# For every file EXCEPT `daemon.heartbeat` this is a full content checksum
+# (byte-identical required), mirroring the #4381 production-binary checksum
+# guard idiom already used elsewhere in this suite
+# (test-loom-daemon-update.sh's `_prod_daemon_checksum`) -- these files are
+# write-once-per-lifecycle-event (daemon start/relaunch, workspace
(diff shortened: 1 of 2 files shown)


rjwalters:
❌ **Closing — superseded by #5190**

This PR targets issue #5179 (the un-isolated `LOOM_PID_FILE` in `test-loom-daemon-update.sh`), but PR #5190 already merged an equivalent — and more comprehensive — fix for the same issue 9 minutes before this PR was opened:

- #5190 adds `defaults/scripts/tests/lib/live-state-sandbox.sh` (256 lines) covering `LOOM_PID_FILE`, `LOOM_AUTONOMY_MARKER`, `LOOM_SOCKET_PATH`, **plus** `LOOM_DAEMON_BIN_DIR`, `LOOM_WORKSPACE`, and `LOOM_MACHINE_CHECKOUT` — a superset of this PR's 162-line helper (`LOOM_PID_FILE`, `LOOM_AUTONOMY_MARKER`, `LOOM_SOCKET_PATH`, `LOOM_WORKSPACES_PATH`).
- #5190 also adds a dedicated `test-live-state-sandbox.sh` (249 lines) and wires it into `ci-wired.txt`; this PR has no equivalent.
- Issue #5179 is already closed (closed by #5190's merge at 2026-08-04T02:33:48Z, 9 minutes before this PR was created).
- Rebasing this PR onto main produces an add/add conflict on `live-state-sandbox.sh` — both PRs independently created the same file to solve the same problem.

Closing as not planned rather than routing through Doctor for a conflict resolution that would just discard duplicate work — main already has the (better) fix.

rjwalters:
## Verdict: Changes Requested — this PR is superseded by already-merged #5190; recommend closing

**Do not merge.** Issue #5179 is already **closed** (2026-08-04T02:33:49Z), resolved by **PR #5190** ("test(daemon): sandbox every live daemon state path behind one helper + guard"), merged one minute earlier at 2026-08-04T02:33:48Z and also carrying `Closes #5179`. This PR's branch (`feature/issue-5179`, head `2aea5311`) forked from `2270d271`, **before** #5190 landed — the two PRs raced to fix the same issue, and #5190 won.

### Evidence
- `mergeStateStatus` is `DIRTY` — not a routine rebase-needed conflict. Diffing this branch against current `origin/main` shows merging it would **delete** `defaults/scripts/tests/test-live-state-sandbox.sh` (#5190's dedicated 249-line, 18-case regression suite for the guard itself), drop its `ci-wired.txt` entry, and shrink `lib/live-state-sandbox.sh` from 256 lines down to this PR's 162-line version — i.e. forcing this through would **regress** already-shipped, more complete work, not just conflict with it.
- No CI checks are attached to this branch (`gh pr checks` returns "no checks reported"), so there's nothing to gate on independent of the above.

### On the technical merits (for the record, since this is worth understanding even though it won't land)
Point 1 of the review brief asked whether `live_state_sandbox_init` actually covers the live-state paths the daemon uses. It has a real gap that the merged #5190 closes and this PR doesn't:

- `loom-daemon/src/daemon_pidfile.rs::resolve_pid_file_path_from` gives `LOOM_PID_FILE` top precedence (tier 1) for the **Rust binary** — so that half is fine.
- But `defaults/scripts/cli/loom-daemon-update.sh:1427` computes `PID_FILE="$DAEMON_STATE_HOME/.daemon.pid"` **independently of `LOOM_PID_FILE` entirely** — `DAEMON_STATE_HOME` is derived from `LOOM_MACHINE_CHECKOUT` (→ `$HOME/.loom`, i.e. the real machine state home) or else `$PWD`-resolved `REPO_ROOT`. This PR's own `lib/live-state-sandbox.sh` header comment actually documents this ("the daemon lifecycle SHELL scripts recompute + re-export this unconditionally... this export is a backstop... the existence check below is what actually catches a shell-script-level leak") — i.e. it knowingly relies on **detecting** a live-file write after the fact via the case-45 guard, rather than **preventing** it, because it never unsets the ambient `LOOM_MACHINE_CHECKOUT`/`LOOM_WORKSPACE`/`LOOM_DAEMON_BIN` that a live agent session inherits into every child process (confirmed real by this repo's own recent #5168 fix for the identical `LOOM_WORKSPACE` leak class in the dispatcher).
- #5190's version additionally **unsets** `LOOM_WORKSPACE` / `LOOM_MACHINE_CHECKOUT` / `LOOM_DAEMON_BIN` in `live_state_sandbox_init`, which is what actually prevents the shell-script-level leak this PR's own comment identifies but doesn't close — real prevention, not detection-after-corruption.

So even setting the merge conflict aside, if this PR's version had been the one to land, the guard would still be strictly weaker: on a host where the leaking env vars are ambient, the real `.daemon.pid` would already be overwritten by the time the case-45 guard fires and fails the suite.

### Points 2–4 from the review brief
- **Snapshot/diff false positives** (point 2): the heartbeat `pid=` token fingerprint in this PR is the same fix #5190 also uses, and is correct for the reason both PRs give — a healthy daemon's routine ~60s heartbeat rewrite must not flap the guard. No issue there.
- **Regression risk to other tests / 256/256 local pass** (point 3): moot given the above — this diff can't land without deleting #5190's suite.
- **Shell-scripting correctness** (point 4): `lib/live-state-sandbox.sh` itself is clean bash (proper quoting, no unguarded globbing, `set -euo pipefail`-safe), no complaints on that axis in isolation.

### Recommendation
Per the Judge role's "Issues Are Suggestions" authority (PR is competent but the underlying issue is already covered by a merged change) — closing this PR rather than routing it through a rebase/Doctor cycle that can only end in "delete the better fix." Labeling `loom:changes-requested` first per protocol, then closing with rationale.

PR #5203, opened 2026-08-04
test(daemon): adopt lib/live-state-sandbox.sh in the start + stop lifecycle suites

### LINKED ISSUE #5191: test(daemon): adopt the live-state sandbox helper in test-loom-daemon-start.sh / test-loom-daemon-stop.sh
## Problem

#5179 fixed live-daemon-state leakage in `defaults/scripts/tests/test-loom-daemon-update.sh` by introducing a shared helper, `defaults/scripts/tests/lib/live-state-sandbox.sh` (PR #5190). The two sibling lifecycle suites were deliberately left out of that PR's scope and still carry the same exposure:

- `defaults/scripts/tests/test-loom-daemon-start.sh` — `grep -c LOOM_PID_FILE` = **0**
- `defaults/scripts/tests/test-loom-daemon-stop.sh` — `grep -c LOOM_PID_FILE` = **0**

Both pin `LOOM_AUTONOMY_MARKER` / `LOOM_SOCKET_PATH` **per invocation** rather than suite-wide, so:

1. any call site that forgets one inherits the ambient (production) value, and
2. a Loom agent session exports the live paths into every child it spawns — observed on a live worker host: `LOOM_PID_FILE=$HOME/.loom/.daemon.pid`, `LOOM_WORKSPACE=<the real checkout>`, `LOOM_DAEMON_BIN=<the real binary>`. `LOOM_PID_FILE` is tier 1 of `daemon_pidfile.rs`'s resolver, ahead of every other tier.

`loom-daemon-stop.sh` is the higher-risk of the two: it both `rm -f "$PID_FILE"`s and `kill`s the pid it reads (`:323-354`, `:410`), and in machine mode (`LOOM_MACHINE_CHECKOUT` set, as on a machine-dispatcher host) `DAEMON_STATE_HOME="$HOME/.loom"` — i.e. the live daemon's own pid file.

## Suggested scope

- Adopt `lib/live-state-sandbox.sh` in both suites: `live_state_sandbox_snapshot` before any export, `live_state_sandbox_init "$WORKDIR/live-state"` in place of the per-variable pins, and `live_state_sandbox_assert_untouched` as a final guard case (the shape PR #5190 added as case 45 of the update suite).
- Keep the per-invocation pins that a specific case genuinely needs — the helper is the suite-wide floor, not a replacement for a case that deliberately drives a particular path.

## Acceptance criteria

- [ ] Both suites snapshot live state before any export and fail loudly if a real `.loom` state path is written during the run.
- [ ] Neither suite can reach the live pid file / autonomy marker / socket dir via an inherited ambient `LOOM_*` value.
- [ ] Both suites still pass in full, verified on a host with a real daemon running.

Related: #5179, #5190, #4087, #5131, #4902, #5118/#5126.

rjwalters:
## Summary

`test-loom-daemon-start.sh` and `test-loom-daemon-stop.sh` were the two sibling suites PR #5190 explicitly left out of scope when it introduced `defaults/scripts/tests/lib/live-state-sandbox.sh` for `test-loom-daemon-update.sh` (#5179). Both suites pinned `LOOM_SOCKET_PATH` / `LOOM_AUTONOMY_MARKER` per-invocation only (`grep -c LOOM_PID_FILE` was 0 in both), so any call site that forgot a pin inherited whatever ambient `LOOM_SOCKET_PATH` / `LOOM_AUTONOMY_MARKER` / `LOOM_MACHINE_CHECKOUT` / `LOOM_WORKSPACE` / `LOOM_DAEMON_BIN` the calling session happened to export -- production paths, if run inside a Loom agent session.

This was not theoretical. Verified directly on this host:

- **`loom-daemon-start.sh`** resolves `SOCKET_PATH="${LOOM_SOCKET_PATH:-$HOME/.loom/loom-daemon.sock}"` -- several early tests (1-7) pass no override at all, so pre-fix they resolved straight onto the real `$HOME/.loom`.
- **`loom-daemon-stop.sh`** resolves its `DAEMON_STATE_HOME` (and therefore the pid file it both `rm -f`s and `kill`s) from `LOOM_MACHINE_CHECKOUT` when set, else `$PWD`'s repo root -- several tests (1/2/3/6/7/8) pin only `LOOM_LAUNCHD_LABEL`, so an ambient `LOOM_MACHINE_CHECKOUT` (as a Loom agent session exports) would have flipped `DAEMON_STATE_HOME` to the REAL `$HOME/.loom` and killed whatever pid is recorded in the real production `.daemon.pid`.

## Changes

**Both suites**, mirroring the shape PR #5190 added to the update suite:
- `live_state_sandbox_snapshot` right after sourcing the existing helpers, before any state-affecting export.
- `live_state_sandbox_init "$WORKDIR/live-state"` once `$WORKDIR` exists, as the suite-wide floor (existing per-invocation `LOOM_SOCKET_PATH` / `LOOM_AUTONOMY_MARKER` pins are kept where a fixture genuinely needs its own scratch `HOME` -- a per-command assignment still wins over the exported default).
- `live_state_sandbox_assert_untouched` as a final guard case before the summary.

**`test-loom-daemon-start.sh` additionally**: adopting the guard surfaced a real, previously-undetected leak -- several `--print-plist`/`--print-unit`-style invocations (the DEK1-4 dropped-env-key tests, AD1-AD9 autonomy-downgrade tests, the AD8 systemd install step, and the four #5081 launchd-bootstrap tests) never `cd` into a fixture directory and don't set `LOOM_MACHINE_CHECKOUT`, so `loom-daemon-start.sh`'s unconditional `.daemon.flags` write (which happens *before* the `--print-plist` "pure inspection" branch) landed under the suite's own checkout `.loom/` instead of scratch space. Fixed each call site by pinning `LOOM_MACHINE_CHECKOUT` to the same scratch `HOME` the test already uses (or an explicit `cd "$WORKDIR"` for the one systemd-install step), verified clean by re-running the new guard repeatedly.

## Test Plan

- [x] `test-loom-daemon-start.sh` -- 81/81 passed (3 consecutive runs), including the new live-state guard.
- [x] `test-loom-daemon-stop.sh` -- 38/38 passed (3 consecutive runs), including the new live-state guard.
- [x] Confirmed the guard is not vacuous: manually simulated a leak into the worktree's own `.loom/.daemon.flags` and verified `live_state_sandbox_assert_untouched` correctly failed and named the offender before reverting.
- [x] Confirmed no residual `.daemon.flags` / `.daemon.pid` / `autonomy-desired` left in this worktree's `.loom/` after repeated runs.
- [x] `shellcheck --severity=warning` clean on both touched scripts.
- [x] `defaults/scripts/tests/check-ci-suite-manifest.sh` -- OK (103 suites, 100 wired, 3 excluded; both suites already wired).

Closes #5191

### DIFF
--- a/defaults/scripts/tests/test-loom-daemon-start.sh
+++ b/defaults/scripts/tests/test-loom-daemon-start.sh
@@ -28,6 +28,19 @@ START_SCRIPT="$(cd "$SCRIPT_DIR/../cli" && pwd)/loom-daemon-start.sh"
 # shellcheck source=lib/bg-proc-trap.sh
 source "$SCRIPT_DIR/lib/bg-proc-trap.sh"
 
+# Shared live-state sandbox (#5179, adopted here per #5191 — this suite had
+# ZERO suite-wide LOOM_PID_FILE-class pins, only per-invocation ones, so any
+# call site below that forgot one inherited whatever LOOM_SOCKET_PATH /
+# LOOM_AUTONOMY_MARKER / LOOM_MACHINE_CHECKOUT / LOOM_WORKSPACE / LOOM_DAEMON_BIN
+# the calling session happened to export). The snapshot MUST run here — before
+# live_state_sandbox_init below rewrites the LOOM_* state vars, and before any
+# sub-invocation can write anything — because it discovers WHICH paths are the
+# live ones by reading the ambient environment. The matching
+# live_state_sandbox_assert_untouched runs as the suite's final guard.
+# shellcheck source=lib/live-state-sandbox.sh
+source "$SCRIPT_DIR/lib/live-state-sandbox.sh"
+live_state_sandbox_snapshot
+
 RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m'
@@ -65,6 +78,19 @@ WORKDIR="$(mktemp -d)"
 trap 'bg_proc_reap; [ -n "$WORKDIR" ] && pkill -f "$WORKDIR" >/dev/null 2>&1; rm -rf "$WORKDIR"' EXIT
 trap 'bg_proc_reap; [ -n "$WORKDIR" ] && pkill -f "$WORKDIR" >/dev/null 2>&1; rm -rf "$WORKDIR"; exit 1' INT TERM
 mkdir -p "$WORKDIR/.loom/logs"
+
+# ---------- live daemon state sandbox (#5179, adopted here per #5191) ----------
+# ONE helper owns every live-state path this suite could otherwise reach (see
+# lib/live-state-sandbox.sh for the full per-variable rationale). This is the
+# suite-wide FLOOR, not a replacement for every case: tests below that need a
+# specific fixture's OWN scratch HOME still pin LOOM_SOCKET_PATH /
+# LOOM_AUTONOMY_MARKER inline on their own invocation (a per-command assignment
+# always wins over this exported default, same precedence as before). This
+# only closes the call sites that do NOT pin them — e.g. tests 1-7 below run
+# with no override at all, so pre-fix they resolved LOOM_SOCKET_PATH's default
+# fallback (`${LOOM_SOCKET_PATH:-$HOME/.loom/loom-daemon.sock}` in
+# loom-daemon-start.sh) straight onto the REAL $HOME/.loom.
+live_state_sandbox_init "$WORKDIR/live-state"
 
 FAKE_BIN="$WORKDIR/fake-loom-daemon"
 cat > "$FAKE_BIN" <<'EOF'
@@ -809,7 +835,7 @@ DEK_LIVE_PLIST="$DEK_HOME/Library/LaunchAgents/${DEK_LABEL}.plist"
 
 # DEK1. First-ever install (no existing plist) must NOT warn.
 dek1_out=$( env -u LOOM_WORK_FINDER -u LOOM_MAIN_HEALTH_GATE -u LOOM_SAFEHOUSE_ENABLED \
-    HOME="$DEK_HOME" LOOM_LAUNCHD_LABEL="$DEK_LABEL" LOOM_DAEMON_BIN="$FAKE_BIN" \
+    HOME="$DEK_HOME" LOOM_MACHINE_CHECKOUT="$DEK_HOME" LOOM_LAUNCHD_LABEL="$DEK_LABEL" LOOM_DAEMON_BIN="$FAKE_BIN" \
     bash "$START_SCRIPT" --print-plist 2>&1 >/dev/null )
 TESTS_RUN=$((TESTS_RUN + 1))
 if ! echo "$dek1_out" | grep -qi 'drops.*env key'; then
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T03:03:06Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Verification performed** (worktree `.loom/worktrees/issue-5191`, already at the PR head SHA `8e7c330`):

- `test-loom-daemon-stop.sh`: 38/38 passed locally.
- `test-loom-daemon-start.sh`: 81/81 passed locally (including the new `live_state_sandbox_assert_untouched` guard case in both suites).
- `shellcheck --severity=warning` clean on all three touched/sourced files (`test-loom-daemon-start.sh`, `test-loom-daemon-stop.sh`, `lib/live-state-sandbox.sh`).
- `check-ci-suite-manifest.sh`: OK (103 suites, 100 wired, 3 excluded).
- Confirmed no residual `.daemon.pid` / `.daemon.flags` / `autonomy-desired` under this worktree's own `.loom/` after the runs — the leak the guard exists to catch did not recur.
- CI: `Shell Test Suites (hermetic)` and `Installer Integration Tests` both green; `mergeable_state` is `clean`.

**Acceptance criteria walkthrough:**
- Both suites call `live_state_sandbox_snapshot` immediately after sourcing the helper, before any state-affecting export — confirmed in both files.
- `live_state_sandbox_init` unsets `LOOM_WORKSPACE`/`LOOM_MACHINE_CHECKOUT`/`LOOM_DAEMON_BIN` and points `LOOM_PID_FILE`/`LOOM_AUTONOMY_MARKER`/`LOOM_SOCKET_PATH`/`LOOM_DAEMON_BIN_DIR` at suite-scratch space as the exported floor, closing every call site that doesn't pin its own override.
- `live_state_sandbox_assert_untouched` runs as the final guard case in both suites and fails loudly (with before/after fingerprints) on any real write — matches the isolation-alone-is-unfalsifiable rationale in the helper's own header.
- Traced every `--print-plist`/`--print-unit` call site in the start suite: the DEK1-4, AD1-AD9, and the first AD8 systemd-install invocation now pin `LOOM_MACHINE_CHECKOUT` to the fixture's own scratch `HOME` (so `DAEMON_STATE_HOME` resolves to `$HOME/.loom` = the fixture, not the checkout); the remaining `--print-unit`/AD8-second/#5081-launchd sites already `cd`'d into `$WORKDIR` pre-PR and needed no change. Cross-checked every grep hit against the diff — no call site was missed.
- For the stop suite, traced `loom-daemon-stop.sh`'s `find_repo_root()`/`DAEMON_STATE_HOME` resolution: every actual invocation already `cd`s into `$WORKDIR` (which has its own scratch `.loom`), so the suite-wide floor from `live_state_sandbox_init` (unsetting `LOOM_MACHINE_CHECKOUT`) is the correct fix without needing per-site changes there.

Small, well-scoped, thoroughly tested. No concerns.

Issue #5117, opened 2026-08-03
guard-destructive-generic.sh: mask_heredoc_bodies() residual blind spots (interpreter-fed heredocs; crafted false opener)


rjwalters:
## Summary

Surfaced during the Judge re-review of PR #5085 (`mask_heredoc_bodies()`, #5000/#5087).
The fix there is correct and approved — this issue records two **residual masking
blind spots** that PR left in place deliberately, so they are tracked rather than
rediscovered later. Neither is the #5087 fail-open (that one is fixed and verified).

Throughout, `/MAIN-CHECKOUT` is a placeholder for the main repository checkout path;
all measurements used a throwaway fixture repo with a `.loom-managed` worktree and
`guards.worktreeIsolation` at its default.

## Residual 1 — a heredoc fed to an interpreter is NOT inert data

`mask_heredoc_bodies()`'s premise, stated in its own header comment, is that "a
heredoc body is genuinely inert DATA -- a heredoc body is never shell-parsed for
redirection/separator syntax". That is true for the *outer* shell, but false when
the body IS the script of an interpreter:

| command shape (body writes into `/MAIN-CHECKOUT/...`) | `origin/main` | PR #5085 head |
|---|---|---|
| `bash <<'EOF'` … `EOF` | deny | **allow** |
| `cat <<'EOF'` … `EOF` piped into `bash` | deny | **allow** |
| `sh -s <<'EOF'` … `EOF` | deny | **allow** |

**Important context that makes this low severity:** interpreter-mediated writes are
*already* broadly uncovered by the write-confinement scanner on `origin/main`. All of
the following ALLOW on `main` today (verified, same fixture):

- `bash -c` with a redirection inside the quoted script (both quote styles)
- `printf '<script>'` piped into `bash`
- `dd if=/dev/null of=/MAIN-CHECKOUT/defaults/hooks/q.sh`
- `install -m 755 /tmp/a /MAIN-CHECKOUT/defaults/hooks/q.sh`

So `main`'s deny on `bash <<EOF` was *incidental* — a by-product of the very
per-physical-line misparse that produced the #5000 false positive in the first place.
This is a new instance of a pre-existing uncovered class, not a new class.

## Residual 2 — a crafted false opener whose delimiter later appears

Opener detection runs *before* `qsplit()`, so it cannot know that a `<<TOKEN` sits
inside a quoted string. If a matching bare-delimiter line appears later in the
[...]

Issue #4523, opened 2026-07-30
roles: gh pr comment --body @file posts the literal @path — mandate --body-file in judge/doctor docs (lost review incident, PR #4457)


rjwalters:
## Context
Live incident 2026-07-30: a Judge's changes-requested review on PR #4457 posted as the literal string `@/private/tmp/.../scratchpad/review.md` — `gh pr comment --body` performs no `@file` expansion (that syntax belongs to `gh api -F`/`--body-file`). The actual review prose was never published, and worse, the scratchpad file at that path contained a DIFFERENT PR's review (the #4471 approval), so a naive later reader could have attributed cross-PR findings. The follow-up Doctor had to reconstruct the blocker from secondary sources.

## Suggested acceptance criteria
- [ ] `defaults/.claude/commands/loom/judge.md` and `doctor.md` (and any other role doc showing comment posting) explicitly mandate `--body-file <path>` (or `gh api -F body=@<path>`) for file-sourced comments and warn that `--body @path` posts the literal string
- [ ] Verdict-posting checklists include "verify the posted comment renders your prose, not a path" as a final step
- [ ] Optional hardening: a guard-hook pattern that denies `gh pr comment --body @` / `gh issue comment --body @` (the literal-@ shape is never intended)

---

## Curator Enhancement (2026-07-30)

Full enhancement — verified diagnosis (both `judge.md`/`doctor.md` already use the SAFE `--body "$(cat <<EOF...)"` pattern everywhere; zero repo-wide hits for the risky `@path` shape, so this is a field-deviation guardrail, not a doc-content fix), implementation guidance (including a real `strip_literal_text()` redaction trap for the optional guard-hook AC), affected files, and test plan — posted as a comment below to keep this body concise: see the "Curator Enhancement (2026-07-30)" comment on this issue.

rjwalters:
## Curator Enhancement (2026-07-30)



### Verified against origin/main — correction to the premise

[...]

PR #5205, opened 2026-08-04
fix(guard): stop masking interpreter-fed heredoc bodies in gh-api-rawfield check

### LINKED ISSUE #5198: Guard regression: gh-api-rawfield-body-literal-at no longer denies a live invocation fed via an interpreter heredoc (bash <<EOF ... EOF)
## Summary

PR #5192 (merged as e33d254a) scoped the `gh-api-rawfield-body-literal-at`
catastrophic-tier check in `guard-destructive-generic.sh` to scan a
heredoc-body-masked copy of `$COMMAND` (`COMMAND_HEREDOC_MASKED`, built via the
existing `mask_heredoc_bodies()` primitive from #5000) instead of the raw
command, fixing a false positive where the denied phrase appeared only as
inert quoted prose inside a heredoc body destined for a file (#5181).

That fix is correct for the false-positive shape it targets, and does not
widen any other check (verified: `COMMAND_HEREDOC_MASKED` is read only by
this one `if` block; every other check still reads raw
`$COMMAND`/`COMMAND_NO_COMMENT`). However it introduces a real false
NEGATIVE the PR's own test additions did not cover: `mask_heredoc_bodies()`'s
own header already documents "Known Limitation #1 — interpreter-fed
heredocs" (added by #5117, scoped explicitly to `extract_write_targets()`'s
ask-tier use): a heredoc whose body is fed to an interpreter
(`bash <<'EOF' ... EOF`, `sh -s <<EOF ... EOF`, `... | bash`) is genuinely
LIVE code to that inner interpreter even though the outer shell never parses
it as redirection/separator syntax — masking it is a safe tradeoff for a
*narrowing/ask* check (worst case: a missed ask) but is a materially worse
tradeoff for a *catastrophic-tier deny* whose entire purpose is preventing
the #4523/#4601/#4685 silent-data-loss shape. #5192 reused the same
primitive for exactly that catastrophic check without re-evaluating whether
the limitation is acceptable there — it is not: it creates a silent,
plausible evasion path for the very deny #5192 exists to preserve.

## Reproduction (confirmed against current main, post-#5192)

```python
import json, subprocess
GUARD = "<repo>/.loom/hooks/guard-destructive-generic.sh"
flag = "-" + "f"
bodykey = "bo" + "dy=" + "@/tmp/review.md"
inner = "gh ap" + "i repos/o/r/issues/123/comments " + flag + " " + bodykey
cmd = "bash <<'EOF'\n" + inner + "\nEOF"
payload = json.dumps({"tool_name": "Bash", "tool_input": {"command": cmd}})
p = subprocess.run([GUARD], input=payload, capture_output=True, text=True)
print(repr(p.stdout))   # '' == ALLOW (silent)
```

- **Before #5192** (`main` at any commit before `e33d254a`): DENY —
  `hookSpecificOutput.permissionDecision: "deny"`.
- **After #5192** (current `main`, both `.loom/hooks/guard-destructive-generic.sh`
  and `defaults/hooks/guard-destructive-generic.sh`): ALLOW — empty stdout, no
  deny at all. The `bash <<'EOF' ... gh api ... -f body=@/tmp/review.md ...
  EOF` command genuinely executes `gh api ... -f body=@path` when `bash`
  interprets its heredoc-fed stdin, silently posting the literal string
  `@/tmp/review.md` as the body — exactly the #4457/#4608 data-loss shape
  this check exists to prevent.

Re-verified directly against both the pre-merge worktree copy and the
current post-merge `main` checkout, absolute paths, no other variables in
play.

## Impact

- **Severity**: High for a guard whose stated purpose is preventing silent,
  irrecoverable data loss (issue body/comment corruption) — this reopens
  exactly that failure mode via a heredoc-to-interpreter wrapper.
- **Likelihood**: Low-to-moderate. Requires constructing a
  `bash <<'EOF' ... EOF`-shaped command specifically. Not the common case,
  but a plausible evasion given this repo's own documented history of agents
  reshaping denied commands to route around a guard (`judge.md`'s `--body
  @path` section cites PR #4600 recurring the #4601 anti-pattern after the
  guard was already live).
- **Scope**: worth confirming no OTHER catastrophic ALWAYS_BLOCK pattern is
  similarly rescoped to a heredoc-masked copy, and treating "catastrophic
  tier + heredoc masking" as a combination that needs explicit sign-off
  going forward.

## Suggested fix directions (not prescriptive)

1. Detect when a masked heredoc's opener command is itself an interpreter
   invocation (`bash`, `sh`, `zsh`, `python`, `eval`, `source`, or piped into
   one, e.g. `... | bash`) and skip masking that specific block — only mask
   heredoc bodies destined for a plain `cat >`/`tee`/redirect-to-file sink,
   never one fed to something that will execute it.
2. Alternatively, keep `gh-api-rawfield-body-literal-at` scanning the raw
   `$COMMAND` unconditionally (as every other catastrophic check still
   does) — i.e. revert the masking for this check — and solve the original
   #5181 false positive a different way (e.g. route it through ASK instead
   of a hard DENY, accepting the rarer false positive as the safer tradeoff
   for a catastrophic-tier check).
3. At minimum, add regression coverage for the interpreter-fed-heredoc shape
   (`bash <<'EOF' ... gh api ... -f body=@path ... EOF`) to
   `tests/hooks/test-guard-destructive.sh` so this exact gap cannot silently
   reappear, whichever fix direction is chosen.

## Context

Found during Judge review of PR #5192 — the PR's `loom:pr` verdict (and its
merge) had already landed from another Judge pass by the time this was
discovered mid-review, and `judge.md`'s "Verdict-Time CAS Recheck" rule
prohibits overwriting a landed verdict, so this is filed as a standalone
follow-up rather than a `loom:changes-requested` relabel. Disagreement was
also posted as a plain comment on PR #5192 itself.

Related: #5181 (the original false positive this PR fixed), #5117 (where
`mask_heredoc_bodies()`'s interpreter-fed-heredoc limitation was first
documented and explicitly scoped OUT for `extract_write_targets()`), #5000
(introduced `mask_heredoc_bodies()`), #4601/#4685/#4523 (the original
`gh api ... -f/--raw-field body=@path` data-loss guard this check protects).

rjwalters:
## Summary

PR #5192 scoped the `gh-api-rawfield-body-literal-at` catastrophic-tier check
in `guard-destructive-generic.sh` to scan a heredoc-body-masked copy of
`$COMMAND` (`mask_heredoc_bodies()`, #5000) instead of the raw command,
fixing a false positive (#5181) where the denied phrase appeared only as
inert quoted prose inside a heredoc body destined for a file.

That masking primitive has a documented limitation (#5117, "Known Limitation
#1 — interpreter-fed heredocs"): a heredoc body fed to an interpreter
(`bash <<'EOF' ... EOF`, `sh -s <<EOF ... EOF`, `... | bash`) is genuinely
LIVE code to that inner interpreter, even though the outer shell never
parses it as redirection/separator syntax. Masking it was an accepted
tradeoff for the ask-tier `extract_write_targets()` scanner it was originally
scoped to, but #5192 reused the same primitive for a catastrophic-tier DENY
without re-deriving that tradeoff — creating a silent evasion path: a
`gh api ... -f body=@path` invocation wrapped in `bash <<'EOF' ... EOF`
now ALLOWs where it should DENY.

## Fix Direction

Chose **direction 1** from the issue (detect interpreter-fed heredoc openers
and skip masking just those blocks) over direction 2 (revert to raw-command
scanning), because direction 2 would reintroduce the #5181 false positive
that #5192 fixed. Direction 1 closes the gap precisely without regressing
#5181.

Added `mask_heredoc_bodies_selective()` — identical closed-block detection to
`mask_heredoc_bodies()`, but skips masking (leaves the body visible) for any
heredoc whose opener line matches `is_interpreter_opener()` (a narrow,
best-effort check for `bash`/`sh`/`zsh`/`dash`/`ksh`/`python*`/`perl`/`ruby`/
`node`/`eval`/`source`/`.` as a command word, or piped into one, e.g.
`cat <<EOF | bash`). Only the `gh-api-rawfield-body-literal-at` check switches
to the selective variant; `extract_write_targets()` keeps calling the
original `mask_heredoc_bodies()` unchanged — a missed ask there remains an
accepted, pre-existing, documented tradeoff, out of scope for this fix.

## Changes

- `defaults/hooks/guard-destructive-generic.sh` / `.loom/hooks/guard-destructive-generic.sh`
  (kept byte-identical, per repo convention): add `is_interpreter_opener()` +
  `mask_heredoc_bodies_selective()`, switch `COMMAND_HEREDOC_MASKED`'s
  construction (the sole read site of that variable) to the selective
  variant, and update the "Known Limitation #1" / "HEREDOC-MASKED SCAN"
  comment blocks to document the exception.
- `tests/hooks/test-guard-destructive.sh`: add regression coverage for the
  interpreter-fed-heredoc evasion (`bash <<EOF`, `sh -s <<EOF`,
  `cat <<EOF | bash` forms) and re-assert the original #5181 false-positive
  case (a heredoc body destined for a plain file sink) stays allowed.

| Issue's exact reproduction (`bash <<'EOF' ... gh api ... -f body=@path ... EOF`) is denied | Yes | Ran the issue's Python repro script directly against `defaults/hooks/guard-destructive-generic.sh`; confirmed `permissionDecision: "deny"` (was empty stdout / silent ALLOW before this fix) |
| #5181 false positive (inert prose in a heredoc destined for a file) remains allowed | Yes | `assert_allow` test re-asserted; also manually verified with `cat > file <<'EOF' ... quoted gh api ... EOF` shape — empty stdout (allow) |
| New regression test added to `tests/hooks/test-guard-destructive.sh` covering the interpreter-fed heredoc shape | Yes | 4 new cases added near the existing #5181 block: `bash <<EOF`, `sh -s <<EOF`, `cat <<EOF \| bash`, and a re-assertion of the #5181 file-sink allow case |
| Full `tests/hooks/test-guard-destructive.sh` suite passes | Yes | 683/683 passed (679 pre-existing + 4 new), including existing #5181/#5192/#5000/#5087 coverage |
| No other catastrophic ALWAYS_BLOCK check reuses `COMMAND_HEREDOC_MASKED` with the same gap | Yes | `grep -n COMMAND_HEREDOC_MASKED` shows exactly one read site (the `gh-api-rawfield-body-literal-at` check). The only other `mask_heredoc_bodies()` call site is inside `extract_write_targets()`, the ask-tier scanner already explicitly scoped to accept this limitation per #5117 — confirmed by Curator on the issue as well |
| New test fails against pre-fix code (proves it's a real regression guard) | Yes | Temporarily reverted both hook files to `HEAD` (pre-fix), re-ran the suite: the 3 new deny cases FAILED as expected (680/683), the allow case still passed. Restored the fix afterward — 683/683 |

## Test Plan

```bash
./tests/hooks/test-guard-destructive.sh   # 683/683 passed
```

Manual reproduction (from the issue body):

```python
import json, subprocess
GUARD = "defaults/hooks/guard-destructive-generic.sh"
cmd = "bash <<'EOF'\ngh api repos/o/r/issues/123/comments -f body=@/tmp/review.md\nEOF"
payload = json.dumps({"tool_name": "Bash", "tool_input": {"command": cmd}})
p = subprocess.run([GUARD], input=payload, capture_output=True, text=True)
print(repr(p.stdout))
# BEFORE this fix: '' (silent ALLOW -- the bug)
# AFTER this fix: {"hookSpecificOutput": {..., "permissionDecision": "deny", ...}}
```

Confirmed identically against both `defaults/hooks/guard-destructive-generic.sh`
and `.loom/hooks/guard-destructive-generic.sh` (kept in sync, both tracked and
committed together per repo convention, same as #5192).

**Note on scope**: this change touches zero `.rs` files (pure bash/awk edit to
a guard hook + its shell test suite), so the Rust portion of the local
`buildGate.command` (`cargo test --workspace --lib --bins`) is orthogonal to
this fix. It was started locally but did not finish within a reasonable wait
on this host (appears blocked on pre-existing, slow/flaky daemon tests
unrelated to this change — same class of pre-existing test-timing flakiness
documented in `.loom/scripts/build-gate.sh`'s own comments). The
`tests/hooks/test-guard-destructive.sh` suite (the actual coverage for the
file this PR modifies) passed cleanly at 683/683 both before and after a
rebase onto latest `origin/main`.

Closes #5198

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -1629,12 +1629,55 @@ function mask_heredoc_bodies(s,   out, lines, nl, i, j, line, trimmed, body, del
 # `python3 <<EOF`, `eval <<EOF`, `source <<EOF`, `. <<EOF`) or the opener is
 # piped into one on the same line (`cat <<EOF | bash`). Deliberately a
 # whole-line, best-effort check (matching the "narrow / best-effort" style
-# of heredoc_delim_at() above), anchored so the interpreter name must start
-# a command segment (start of line, or right after `;`/`&`/`|`) rather than
-# matching an arbitrary substring -- e.g. `echo "installs bash" <<EOF` does
-# NOT match, since "bash" there is not itself a command word.
-function is_interpreter_opener(line) {
-    return (line ~ /(^|[;&|])[ \t]*(bash|sh|zsh|dash|ksh|python[0-9.]*|perl|ruby|node|nodejs|eval|source|\.)([ \t]|$)/)
+# of heredoc_delim_at() above): the interpreter must be the COMMAND WORD of
+# some segment of the line, not an arbitrary substring -- e.g.
+# `echo "installs bash" <<EOF` does NOT match, since "bash" there is a bare
+# argument, not the command word.
+#
+# Recognizing the command word robustly (#5205): the interpreter word is
+# matched against the path BASENAME of each segment command word, and any
+# leading `env`/`command`/`exec`/`builtin` wrapper (with its own flags and
+# `VAR=value` assignments) is stripped first -- none of those change what
+# actually executes. So a path-qualified or wrapped invocation of the same
+# interpreter (`/bin/bash <<EOF`, `env bash <<EOF`, `command bash <<EOF`,
+# `./bash <<EOF`, `/usr/bin/python3 <<EOF`) is caught too, closing the
+# evasion class where those forms slipped past the old first-token-only
+# regex and got their live bodies silently masked (i.e. ALLOWed).
+function _interp_basename(tok,   base) {
+    # Reduce a (possibly path-qualified) command word to its basename: the
+    # text after the last `/`. `/bin/bash` -> `bash`, `./bash` -> `bash`,
+    # `/usr/bin/python3` -> `python3`; a bare `bash` or `.` is unchanged.
+    base = tok
+    sub(/^.*\//, "", base)
+    return base
+}
+function is_interpreter_opener(line,   n, segs, i, seg, m, toks, j, base) {
+    # Split into command segments on ; & | (covers && and || too) so a piped
+    # or chained interpreter is caught in ANY position, e.g. `cat <<EOF | bash`.
+    n = split(line, segs, /[;&|]+/)
+    for (i = 1; i <= n; i++) {
+        seg = segs[i]
+        sub(/^[ \t]+/, "", seg)
+        m = split(seg, toks, /[ \t]+/)
+        if (m == 0) continue
+        # Strip leading command wrappers that do not change what runs:
+        # env / command / exec / builtin, each followed by its own -flags
+        # and (for env) VAR=value assignments, then the real command word.
+        j = 1
+        while (j <= m) {
+            if (toks[j] == "env" || toks[j] == "command" || toks[j] == "exec" || toks[j] == "builtin") {
+                j++
+                while (j <= m && (toks[j] ~ /^-/ || toks[j] ~ /^[A-Za-z_][A-Za-z0-9_]*=/)) j++
+                continue
+            }
+            break
+        }
+        if (j > m) continue
+        base = _interp_basename(toks[j])
+        if (base ~ /^(bash|sh|zsh|dash|ksh|python[0-9.]*|perl|ruby|node|nodejs|eval|source|\.)$/)
+            return 1
+    }
+    return 0
 }
 # Same closed-block detection as mask_heredoc_bodies(), but SKIPS masking
 # (leaves the body visible) for any block whose opener is interpreter-fed
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -1516,6 +1516,20 @@ function unmask_ws(s) {
 #      script/stdin argument) is a materially larger, separate piece of work
 #      than this masking pass and is OUT OF SCOPE here; track it as its own
 #      follow-up rather than bolting a partial fix onto heredoc masking.
+#      This tradeoff (missed ASK, worst case) is unchanged for
+#      extract_write_targets() -- it still calls plain mask_heredoc_bodies()
+#      below and still masks interpreter-fed bodies. But #5192 later reused
+#      this same primitive for a CATASTROPHIC-tier DENY
+#      (gh-api-rawfield-body-literal-at, ~line 2234) without re-deriving this
+#      tradeoff for that tier: masking an interpreter-fed body there doesn't
+#      risk a missed ask, it silently flips a DENY to an ALLOW on the exact
+#      #4523/#4601/#4685 data-loss shape the check exists to catch (#5198).
+#      That is NOT an acceptable tradeoff for a catastrophic-tier check, so
+#      mask_heredoc_bodies_selective() below (used ONLY by that one check)
+#      recognizes an interpreter-fed opener and leaves that specific block's
+#      body UNMASKED (visible to the scan) instead of masking it -- narrowing
+#      still applies to every other (non-interpreter-fed) heredoc in the same
+#      command, so the #5181 false-positive fix stays intact.
 #
 #   2. Crafted false opener whose delimiter later appears. Opener detection
 #      (heredoc_delim_at()) runs on a single physical line, before qsplit()
@@ -1601,6 +1615,60 @@ function mask_heredoc_bodies(s,   out, lines, nl, i, j, line, trimmed, body, del
                 body = lines[j]
                 gsub(/./, MASKC, body)
                 lines[j] = body
+            }
+            i = closeat            # resume scanning after the delimiter line
+            break
+        }
+    }
+    out = lines[1]
+    for (i = 2; i <= nl; i++) out = out "\n" lines[i]
+    return out
+}
+# True when a heredoc OPENER line looks like it feeds an interpreter --
+# either the opener command itself (`bash <<EOF`, `sh -s <<EOF`,
+# `python3 <<EOF`, `eval <<EOF`, `source <<EOF`, `. <<EOF`) or the opener is
+# piped into one on the same line (`cat <<EOF | bash`). Deliberately a
+# whole-line, best-effort check (matching the "narrow / best-effort" style
+# of heredoc_delim_at() above), anchored so the interpreter name must start
+# a command segment (start of line, or right after `;`/`&`/`|`) rather than
+# matching an arbitrary substring -- e.g. `echo "installs bash" <<EOF` does
+# NOT match, since "bash" there is not itself a command word.
+function is_interpreter_opener(line) {
+    return (line ~ /(^|[;&|])[ \t]*(bash|sh|zsh|dash|ksh|python[0-9.]*|perl|ruby|node|nodejs|eval|source|\.)([ \t]|$)/)
+}
+# Same closed-block detection as mask_heredoc_bodies(), but SKIPS masking
+# (leaves the body visible) for any block whose opener is interpreter-fed
+# per is_interpreter_opener() -- see KNOWN LIMITATIONS #1 above. Used ONLY by
+# the gh-api-rawfield-body-literal-at catastrophic check (#5198);
+# extract_write_targets() keeps calling plain mask_heredoc_bodies() above,
+# unchanged.
+function mask_heredoc_bodies_selective(s,   out, lines, nl, i, j, line, trimmed, body, delim, closeat, p, off, MASKC) {
+    MASKC = sprintf("%c", 23) # ETB -- placeholder for inert heredoc-body text
+    nl = split(s, lines, "\n")
+    if (nl == 0) return ""
+    for (i = 1; i <= nl; i++) {
+        line = lines[i]
+        off = 1
+        while (1) {
+            p = index(substr(line, off), "<<")
+            if (p == 0) break
+            p = off + p - 1
+            off = p + 2
+            delim = heredoc_delim_at(line, p)
+            if (delim == "") continue
+            closeat = 0
+            for (j = i + 1; j <= nl; j++) {
+                trimmed = lines[j]
+                sub(/^\t+/, "", trimmed)
+                if (trimmed == delim) { closeat = j; break }
+            }
+            if (closeat == 0) continue
+            if (!is_interpreter_opener(line)) {
+                for (j = i + 1; j < closeat; j++) {
+                    body = lines[j]
+                    gsub(/./, MASKC, body)
+                    lines[j] = body
+                }
             }
             i = closeat            # resume scanning after the delimiter line
             break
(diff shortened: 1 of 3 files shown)


rjwalters:

Verified the core fix works as claimed:
- Local run of `tests/hooks/test-guard-destructive.sh` on the PR head (`b70ae65`, matches the builder worktree): **683/683 passed**.
- Reverted both `guard-destructive-generic.sh` copies to `HEAD~1` (pre-fix) and re-ran: **680/683**, with exactly the 3 new `#5198` deny cases failing — confirms the new tests are a real regression guard, not vacuous.
- `.loom/hooks/guard-destructive-generic.sh` and `defaults/hooks/guard-destructive-generic.sh` diffs are byte-identical (both hunks in the PR diff match line-for-line) — sync convention respected.
- Scope check: `grep -n COMMAND_HEREDOC_MASKED` confirms exactly one read site (the `gh-api-rawfield-body-literal-at` check); `extract_write_targets()` (line ~2508) still calls plain `mask_heredoc_bodies()` unchanged — the ask-tier scanner is untouched, matching the PR's stated scope.
- CI is green (`Shell Test Suites (hermetic)`, `Installer Integration Tests`, all others) and `mergeStateStatus` is not DIRTY.

**However — `is_interpreter_opener()` has a real, reproducible gap that reopens the exact evasion class this PR is meant to close.**

The regex:
```awk
function is_interpreter_opener(line) {
    return (line ~ /(^|[;&|])[ \t]*(bash|sh|zsh|dash|ksh|python[0-9.]*|perl|ruby|node|nodejs|eval|source|\.)([ \t]|$)/)
}
```
requires the interpreter word to be the literal first token of the opener line (or immediately after `;`/`&`/`|`). Any path-qualified or wrapped invocation of the *same* interpreter fails to match, so `mask_heredoc_bodies_selective()` masks the body anyway — silently flipping the same catastrophic DENY back to a silent ALLOW. I reproduced this directly against `defaults/hooks/guard-destructive-generic.sh` on the PR's own head commit (empty stdout = ALLOW, non-empty JSON = DENY):

```bash
# Control — matches the PR's own repro, correctly DENIES:
bash <<'EOF'
gh api repos/o/r/issues/123/comments -f body=@/tmp/review.md
EOF
# -> denies (correct)

# Bypass 1 — absolute path to the same interpreter:
/bin/bash <<'EOF'
gh api repos/o/r/issues/123/comments -f body=@/tmp/review.md
EOF
# -> silently ALLOWS (empty stdout)

# Bypass 2 — env-wrapped (the single most common real-world shebang pattern, `#!/usr/bin/env bash`):
env bash <<'EOF'
gh api repos/o/r/issues/123/comments -f body=@/tmp/review.md
EOF
# -> silently ALLOWS

# Bypass 3 — `command` builtin prefix:
command bash <<'EOF'
gh api repos/o/r/issues/123/comments -f body=@/tmp/review.md
EOF
# -> silently ALLOWS

# Bypass 4 — relative path:
./bash <<'EOF'
gh api repos/o/r/issues/123/comments -f body=@/tmp/review.md
EOF
# -> silently ALLOWS
```

All four are functionally identical to the PR's own reproduction case and equally live/executable — they just spell the interpreter name differently. None of them are exotic; `/bin/bash` and `env bash` in particular are extremely common real-world spellings (the latter is literally the standard portable shebang form). This means the fix closes the *narrowest* form of the #5198 evasion (bare interpreter name only) while leaving the broader evasion class open — a determined (or even just differently-styled) invocation still slips through unmasked-but-undetected, exactly the failure mode item 1 of this review was asked to check for.

**Suggested fix direction** (no need to implement exhaustively — a couple of these would close most real-world exposure):
- Strip a leading `env`/`command`/`exec`/`builtin` token (with its own flags) before applying the interpreter-name check, since none of those change what actually executes.
- Match on the path *basename* (text after the last `/`) rather than requiring the interpreter word to start the segment, so `/bin/bash`, `./bash`, `/usr/bin/python3` etc. are recognized too.
- Add regression tests for at least the `/bin/bash <<EOF ...` and `env bash <<EOF ...` shapes alongside the existing four `#5198` cases, so this doesn't silently regress again.

Given this is a catastrophic-tier deny specifically protecting against silent GitHub issue/comment body data loss (#4523/#4601/#4685), and the PR's own stated goal is closing an evasion path, I don't think it's safe to merge with this large a residual gap — it's a genuine improvement over the #5192 regression (the literal reported repro now denies), but it doesn't close the class of bug the issue describes. Requesting changes to widen `is_interpreter_opener()` (or an equivalent guard) to cover path-qualified / `env`-wrapped interpreter invocations, plus regression tests for at least those two shapes.

rjwalters:
Doctor pass: PR still carries a fresh `loom:treating` claim (claimed 2026-08-04T03:25:42Z, age ~47m at check time — under the 60m LOOM_STALE_TREATING_MINUTES threshold) — standing down without reclaiming. Not stomping.

rjwalters:
🩺 Doctor: addressed the changes-requested feedback on `is_interpreter_opener()`.

The gap was real: the function only recognized the interpreter when it was the literal first token of the heredoc opener line, so path-qualified and wrapper-prefixed invocations of the same interpreter bypassed detection — their heredoc bodies were masked and the live `gh api ... -f body=@path` inside was silently ALLOWed.

**Fix** (widened `is_interpreter_opener()`, both hook copies kept byte-identical):
- Split the opener line into `; & |` segments so a piped/chained interpreter is caught in any position.
- Strip a leading `env`/`command`/`exec`/`builtin` wrapper (with its own flags and `VAR=value` assignments) before the interpreter-name check — none of those change what executes.
- Match on the command word's path **basename**, so `/bin/bash`, `./bash`, `/usr/bin/python3` etc. are recognized.

All five bypasses the review flagged now DENY (`/bin/bash`, `env bash`, `command bash`, `./bash`, and `/usr/bin/python3`), same as the bare-`bash` control. The #5181 false-positive-allow is preserved — non-interpreter sinks (`cat > file`, `/bin/cat > file`, `grep bash`, `echo "installs bash"`) still return false, so inert prose quoting the phrase to a plain file sink stays allowed.

**Tests**: added #5205 regression cases for the `/bin/bash`, `env bash`, `command bash`, `./bash`, `/usr/bin/python3` shapes plus a path-qualified non-interpreter allow case. Full `tests/hooks/test-guard-destructive.sh` suite: 689/689 passing (existing #5198 four deny cases and the #5181 allow case intact).

Re-requesting review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:25:24Z) — standing down without reclaiming. Not stomping.

rjwalters:
Re-review: verified the fix closes all four previously-flagged bypasses.

**Reproduced independently against the PR head (`6523d882`)**, feeding each opener directly through `defaults/hooks/guard-destructive-generic.sh` with the exact `gh api ... -f body=@path` payload from the original repro:

| Opener | Prior head | This head |
|---|---|---|
| `bash <<EOF` (control) | deny | deny |
| `/bin/bash <<EOF` | **allow (bypass)** | deny |
| `env bash <<EOF` | **allow (bypass)** | deny |
| `command bash <<EOF` | **allow (bypass)** | deny |
| `./bash <<EOF` | **allow (bypass)** | deny |
| `/usr/bin/python3 <<EOF` | (new coverage) | deny |
| `exec bash <<EOF` | (new coverage) | deny |
| `env -i bash <<EOF` | (new coverage) | deny |
| `env VAR=1 bash <<EOF` | (new coverage) | deny |
| `command env bash <<EOF` (double wrapper) | (new coverage) | deny |
| `cat <<EOF | /bin/bash` (piped) | (new coverage) | deny |
| `/bin/cat > file <<EOF` (non-interpreter sink, #5181 regression guard) | allow | allow (correctly preserved) |

All four bypasses from my prior rejection now deny, and the widened tokenizing approach (segment split on `; & |`, wrapper-stripping for `env`/`command`/`exec`/`builtin` with flags/`VAR=value`, basename match) closes several additional shapes I tried that weren't in the original report (piped, chained, multi-assignment `env`, stacked wrappers) without touching the original #5181 false-positive-allow case.

**Verified independently, not just re-reading the diff:**
- Both hook copies (`defaults/hooks/guard-destructive-generic.sh` and `.loom/hooks/guard-destructive-generic.sh`) are still byte-identical (`diff -q` confirms).
- Ran the full `tests/hooks/test-guard-destructive.sh` suite myself (fresh run, not trusting the reported number): **689/689 passing**, including all 6 new `#5205` regression cases (the 5 deny shapes + the `/bin/cat` allow case).
- CI on the PR head is green (`Shell Test Suites (hermetic)`, `Installer Integration Tests`, `Codex Adapter Smoke`, docs/budget checks) and `mergeStateStatus` is `CLEAN`.

**Non-blocking note for a possible follow-up** (did not gate this review — these are novel shapes beyond what the prior rejection flagged, and `is_interpreter_opener()` is explicitly documented as a best-effort heuristic, not an exhaustive interpreter-detection engine): I found two additional narrow gaps that still silently allow — `\bash <<EOF` (backslash-escaped command word; a common technique to dodge alias/function shadowing, but resolves to the same real interpreter) and `sudo bash <<EOF` (`sudo` isn't in the stripped-wrapper list alongside `env`/`command`/`exec`/`builtin`). Neither was part of the original bypass report, and closing them isn't required for this PR to land — flagging in case a follow-up issue is worth filing.

Approving. Good, thorough response to the review — the fix is correct, well-tested, and scoped precisely to the flagged gap plus reasonable adjacent coverage.

rjwalters:
**Standing down — verdict raced.** I ran the Verdict-Time CAS Recheck immediately before writing my label and found my `loom:reviewing` claim (04:25:24Z) removed and `loom:pr` already applied at 04:37:04Z by a concurrent Judge. Discarding my verdict; **not writing any label** and not overwriting theirs.

For the record: **my independent review reached the same conclusion (approve)** — no disagreement to raise. Recording the evidence since the work is done, plus two findings the other pass did not cover.

**Independently verified against head `6523d882`** (payload fed directly through `defaults/hooks/guard-destructive-generic.sh`, not read off the diff):

| Opener | Decision |
|---|---|
| bare invocation (control) | deny |
| `bash <<EOF` | deny |
| `/bin/bash <<EOF` | deny |
| `env bash <<EOF` | deny |
| `command bash <<EOF` | deny |
| `./bash <<EOF` | deny |
| `cat <<EOF \| bash` | deny |
| `cat > f.md <<EOF` (inert prose, #5181 guard) | allow |
| `/bin/cat > f.md <<EOF` (path-qualified sink) | allow |

All four bypasses from the first review now deny, and both #5181 false-positive-allow controls are preserved. Also confirmed: `.loom/hooks/` and `defaults/hooks/` copies are byte-identical (`diff` clean), all 10 new `#5198`/`#5205` regression cases pass locally, and CI on this head is fully green (`Shell Test Suites (hermetic)`, `Installer Integration Tests`, `Docs/Defaults Parity Check`, all others success or skipped) with no DIRTY merge state.

**Two additions to the other pass's follow-up note** — filed together as #5226, non-blocking:

- `LC_ALL=C bash <<EOF` still allows. This one is the most interesting: the wrapper-stripping loop **already** consumes `VAR=value` tokens, but only after it has seen `env`/`command`/`exec`/`builtin` first, so a bare assignment prefix (valid shell, very common) hits `break` at `toks[1]` and never reaches the basename check.
- `timeout 60 bash <<EOF` and `"bash" <<EOF` also allow (exec-wrapper class; quoted command word) — alongside the `sudo bash` / `\bash` shapes already noted above.

#5226 also raises the design point worth a Curator's call: masking-by-default with an interpreter allowlist means every unrecognized opener fails **open** on a catastrophic-tier check. Since the #5181 false positive is specifically about *inert file sinks*, masking only on a known-inert-sink allowlist would give the same false-positive fix with a bounded surface that fails toward DENY. Explicitly **not** a blocker for this PR — #5205 strictly shrinks the gap and introduces no regression.

Nice work by the Doctor: the tokenizing rewrite is a genuine structural fix, not a regex patch, and the KNOWN LIMITATIONS comment correctly separates the ask-tier `extract_write_targets()` tradeoff from the catastrophic-tier one.

Issue #4581, opened 2026-07-30
loom restart: bare-exec fallback in loom_cmd_restart() still doesn't harvest/re-export env before re-render


rjwalters:
While evaluating PR #4579 (#4522 fix), the PR's investigation traced the actual
incident root cause: `scripts/loom`'s `loom_cmd_restart()` (lines ~284-317)
tries the supervised IPC restart primitive first (safe, no re-render), but
when that primitive is unavailable/refused, it falls back to:

```bash
"$stop_target" --restarting
exec "$start_target" ${passthrough[@]+"${passthrough[@]}"}
```

— a bare `exec` of `loom-daemon-start.sh` with only CLI passthrough args, no
env re-export at all. It inherits whatever `LOOM_*` vars happen to be exported
in the invoking shell — a fresh terminal, different SSH session, or any
wrapper that doesn't re-source the operator's profile reproduces the original
incident (rich plist → minimal plist, silently narrowed).

`loom-daemon-update.sh`'s equivalent refused-restart fallback
(`perform_relaunch` / `perform_systemd_relaunch`, #4118/#4126) already
harvests and re-exports the live file's `LOOM_*`/token env before
re-rendering (or aborts with exit 6 if the harvest fails). `scripts/loom`'s
`loom_cmd_restart()` fallback should apply the same harvest-and-preserve
pattern instead of a bare `exec` with only passthrough args.

PR #4579 adds a diff-and-warn mechanism (`warn_dropped_env_keys`) that makes
this fallback path self-diagnosing going forward — the next occurrence prints
a warning instead of silently narrowing — but does not fix the fallback
itself, by design (scoped to keep that PR focused on the warning mechanism).

## Suggested fix

Apply the harvest-and-preserve pattern from `perform_relaunch` /
`perform_systemd_relaunch` (`defaults/scripts/cli/loom-daemon-update.sh`) to
`scripts/loom`'s `loom_cmd_restart()` bare-exec fallback, so it re-exports the
live installed plist/unit's env before the `exec` instead of relying on
whatever the invoking shell happens to have.

Discovered during code evaluation of PR #4579.

rjwalters:
@/tmp/curator-4581-comment.md

rjwalters:
**Champion Review: APPROVED**

This issue has been evaluated and promoted to `loom:issue` status. All quality criteria passed:

[...]

PR #5206, opened 2026-08-04
feat(changelog): add a deterministic conventional-commit CHANGELOG generator

### LINKED ISSUE #5196: CHANGELOG is not a running record: every release hand-reconstructs ~150 commits, and Unreleased doesn't survive a release
## Problem: the CHANGELOG is not a running record, so every release reconstructs ~150 commits by hand

Cutting v0.18.0 exposed the whole cycle:

- The `## [Unreleased]` section was **completely empty** despite **146 commits** since
  v0.17.0. Nothing had accumulated into it over three days of merges.
- The release therefore had to **reconstruct the entire entry from `git log`** — 151
  commits grouped by hand into themes, with notable items selected and PR numbers cited.
  That is expensive judgment work repeated every cycle.
- `main` advanced **5 commits mid-release**, so the drafted entry had to be rebased and
  re-folded. Without catching that, v0.18.0 would have shipped an entry that **omitted
  commits the release actually contained** — including #5153, a fix filed the same day.
- **Right now there is no `## [Unreleased]` heading at all.** `/repo:release`'s fold
  renames it in place (`## [Unreleased]` → `## [0.18.0] - …`) and nothing re-adds it, so
  the next cycle again starts from nothing.

Related closed work built the *verification* side — #3499 / #3497 added the Phase 1.5
completeness gate that checks each shipped **tag** has an entry. Nothing maintains the
section **between** releases, which is where the cost actually lands.

## The raw material is already excellent — that shapes the fix

Measured over `v0.17.0..HEAD` (163 commits):

| | count | share |
|---|---|---|
| conventional-prefixed (`feat`/`fix`/`docs`…) | **162** | 99.4% |
| ending in `(#NNN)` | **159** | 97.5% |

So the commit stream is very nearly machine-parseable already. There is also **no
changelog tooling** in the repo (`scripts/` has none, no workflow touches `CHANGELOG.md`).

**This argues against per-PR changelog edits.** At this repo's merge rate — 151 commits in
three days, ~50 PRs/day, nearly all agent-authored — having every PR append to one shared
`Unreleased` section would produce constant merge conflicts on a single file. The
fragment-directory pattern (`changelog.d/`, à la towncrier/changesets) avoids the
conflicts but adds a per-PR step an agent can silently skip, and the resulting entry would
be no better than what the commit titles already encode.

Generating from the commit stream needs **no new per-PR discipline** and no new failure
mode.

## Suggested direction

**1. A generator.** Something like `scripts/changelog.sh draft <from>..<to>` that emits the
mechanical skeleton from conventional commits: bucket `feat` → Added, `fix` → Fixed, plus
Changed/Removed, carry the `(#NNN)` refs through, and drop non-shipping types
(`test`, `chore`, `ci`) or bucket them separately. Deterministic and re-runnable.

**2. Keep `[Unreleased]` alive.** Re-add the empty heading immediately after a release so
the section always exists, and let the generator refresh its contents on demand. That
alone removes the "empty section, reconstruct everything" surprise.

**3. Close the mid-release race.** The release path should regenerate (or verify) the entry
against the **actually-tagged** commit range rather than whatever range was drafted
earlier — that is the specific failure that nearly shipped an incomplete v0.18.0 entry.
This composes with `/repo:release`'s existing Phase 4/5 fold rather than replacing it.

**4. Leave the human part human.** Thematic grouping and the `### Summary` prose are
genuine judgment — v0.18.0's entry groups 151 commits into things like "orphaned-process
reaping", "guard false positives", "dashboard correctness", which no type-based bucketing
produces. The generator should produce the skeleton; the release curates the summary. Do
not try to automate the narrative.

**5. Decide the link-reference block's fate.** The bottom of `CHANGELOG.md` defines only
`[Unreleased]`, `[0.2.3]`, `[0.2.0]`, `[0.1.0]` — versions **0.3.0 through 0.18.0 have no
link refs**, so the Keep-a-Changelog convention was abandoned mid-stream. The
`[Unreleased]` definition is now **dangling**, since the heading it referenced was consumed
by the release. Either restore refs for every version (mechanical, and the generator could
own it) or drop the block; leaving it half-populated means every release has to decide
again.

## Acceptance criteria

- [ ] A documented command regenerates the `Unreleased` entry from a commit range,
      deterministically, with `(#NNN)` refs preserved.
- [ ] `## [Unreleased]` exists in `CHANGELOG.md` at all times, including immediately after a
      release.
- [ ] The release entry is generated/verified against the final tagged range, so a
      mid-release merge cannot leave shipped commits uncatalogued.
- [ ] Non-shipping commit types are excluded or clearly separated, so the entry describes
      user-visible change.
- [ ] A decision is recorded on the link-reference block, and no dangling `[Unreleased]`
      definition remains.
- [ ] Cutting a release no longer requires hand-grouping ~150 commits — only reviewing and
      writing the summary.

## Non-goals

- Per-PR changelog edits or a fragment directory — rejected above on merge-conflict and
  skippability grounds at this repo's merge rate.
- Auto-generating the `### Summary` narrative.
- Replacing `/repo:release`'s CHANGELOG phases; this feeds them.

## Provenance

Surfaced cutting v0.18.0 on 2026-08-04, then confirmed by a `/repo:all` pass immediately
after (which found the section now absent entirely). All prior changelog issues are closed
and covered verification, not maintenance: #3499, #3497, #3529, #3661, #3856.

## Affected Files

- `scripts/changelog.sh` (new) - generator command (e.g. `draft <from>..<to>`) that
  parses conventional commits in a git log range, buckets them (`feat`→Added,
  `fix`→Fixed, `docs`/`refactor`/`perf`→Changed, `revert`→Removed, non-shipping
  types like `test`/`chore`/`ci` excluded or bucketed separately), and emits a
  Keep-a-Changelog-formatted skeleton with `(#NNN)` refs preserved
- `CHANGELOG.md` - re-add the `## [Unreleased]` heading (currently absent — the
  top of the file jumps straight from the title block to `## [0.18.0]`); decide the
  fate of the dangling link-reference block at the bottom (only `[Unreleased]`,
  `[0.2.3]`, `[0.2.0]`, `[0.1.0]` are defined — versions 0.3.0 through 0.18.0 have
  none), either backfilling refs for every version or dropping the block entirely
- `.claude/commands/repo/release.md` - Phase 4's "Fold an existing `## Unreleased`
  section into the draft" step (around lines 261-294) renames `## Unreleased` to the
  new version heading in place and never re-adds it — this is the specific gap that
  leaves the section empty until the next release reconstructs it by hand. Phase 1.5's
  completeness gate (lines 98-124) and Phase 6's extraction (`sed` range around line
  367) should regenerate/verify against the **final tagged range** rather than
  whatever range was drafted earlier, closing the mid-release race described in the
  issue body (a merge landing after drafting but before tagging must not be silently
  dropped from the entry)

## Test Plan

- [ ] Automated tests: exercise `scripts/changelog.sh draft <from>..<to>` against a
  known commit range (e.g. `v0.17.0..v0.18.0`) and assert the output buckets match
  expectations (conventional-prefixed commits sorted into the right section,
  `(#NNN)` refs preserved, non-shipping types excluded/separated); re-run on the
  same range and confirm byte-identical output (deterministic)
- [ ] Manual verification: run a dry-run of the updated `/repo:release` Phase 4 fold
  against a scratch branch and confirm `## [Unreleased]` survives the rename (i.e.
  the heading exists again immediately after the fold, not just before it)
- [ ] Edge cases: (1) a commit range with zero conventional-prefixed commits still
  produces a valid (empty-but-headed) skeleton rather than erroring; (2) a commit
  landing between the initial draft and the final tag is caught by the
  final-tagged-range regeneration/verification rather than silently omitted; (3) the
  link-reference block decision (restore vs. drop) is applied consistently — no
  dangling `[Unreleased]` definition remains pointing at a range that no longer
  matches the heading above it

rjwalters:
## Summary

Every release hand-reconstructed ~150 commits into `CHANGELOG.md` by hand because
the `## [Unreleased]` heading disappeared entirely after v0.18.0 (nothing
re-added it) and nothing generated an entry from the commit stream, despite
that stream being ~99% conventional-commit formatted already.

## Changes

- **`scripts/changelog.sh draft <from>..<to>`** — parses conventional commits
  in a git log range and buckets them into Keep a Changelog sections
  (`feat`→Added, `fix`→Fixed, `docs`/`refactor`/`perf`/`style`→Changed,
  `revert`→Removed), drops non-shipping `test`/`chore`/`ci`/`build` commits,
  and surfaces anything with an unrecognized or missing conventional prefix
  under `### Other` rather than silently dropping it. Every `(#NNN)` ref is
  preserved verbatim. Deterministic — re-running on an unchanged range is
  byte-identical. (While building the test suite I found and fixed a real bug:
  a plain `while read` loop silently swallowed the *oldest* commit in every
  range, because `git log`'s last line has no trailing newline — this would
  have quietly dropped exactly the kind of commit this tool exists to catch.)
- **`scripts/changelog.sh verify <from>..<to> [<file>]`** — flags any shipping
  commit whose `#NNN` ref never appears in a given file (default
  `CHANGELOG.md`); used as a final "final tagged range" completeness check.
- **`scripts/test-changelog.sh`** — unit suite against a disposable scratch
  git repo (no dependency on this repo's own history), wired into
  `defaults/scripts/build-gate.sh` (+ `defaults/docs/build-gate.md`).
- **`CHANGELOG.md`** — restored the missing `## [Unreleased]` heading. Decided
  the fate of the dangling link-reference block at the bottom: dropped it
  rather than half-backfill, since only 4 of 19 released versions (`0.1.0`,
  `0.2.0`, `0.2.3`, and the stale `Unreleased` def) ever had one — 0.3.0
  through 0.18.0 never did, so the convention was already effectively
  abandoned. Decision recorded in `.repo/release-policy.md`.
- **`.repo/release-policy.md`** (new) — `/repo:release` (the release-cutting
  slash command) is a *shared* command co-installed from `rjwalters/repo`, not
  part of this repo (`.claude/commands` is gitignored here — see
  `.gitignore`), so `.claude/commands/repo/release.md` itself cannot be edited
  from this PR. `/repo:release` exposes exactly this seam mechanism for a
  project to inject procedural policy without forking the shared command, so
  this file binds the two seams needed to close the gaps the issue describes:
  - `pre-changelog-style`: use `scripts/changelog.sh draft` as the raw
    material for drafting, and — the specific fix for the missing-heading
    bug — re-insert a fresh empty `## [Unreleased]` heading immediately after
    the default fold consumes the old one, every release.
  - `pre-push`: re-verify CHANGELOG completeness against the range that is
    actually about to ship (recomputed fresh, after Phase 5 has already cut
    the tag) rather than whatever range was drafted earlier in the session —
    this is the fix for the mid-release race where commits landing between
    drafting and tagging get silently dropped (the v0.18.0 near-miss the
    issue cites).

- [x] `scripts/changelog.sh draft <from>..<to>` deterministically regenerates
  the Unreleased entry with `(#NNN)` refs preserved — verified in
  `scripts/test-changelog.sh` (bucketing + byte-identical re-run) and against
  the real `v0.17.0..v0.18.0` range.
- [x] `## [Unreleased]` exists in CHANGELOG.md at all times, including
  immediately after a release — heading restored now; `pre-changelog-style`
  seam re-adds it after every future fold.
- [x] Release entry generation/verification uses the final tagged range, not
  the drafted range — `pre-push` seam recomputes and runs `changelog.sh
  verify` against the tag Phase 5 just cut, after main may have advanced past
  whatever Phase 3/4 originally drafted from.
- [x] Non-shipping commit types excluded or clearly separated —
  `test`/`chore`/`ci`/`build` excluded; unrecognized/no-prefix commits kept
  under `### Other`, not dropped.
- [x] A decision on the link-reference block is recorded, no dangling
  `[Unreleased]` link definition remains — block dropped, decision recorded in
  `.repo/release-policy.md`.
- [x] Cutting a release no longer requires hand-grouping ~150 commits — the
  generator produces the mechanical skeleton; a human still writes the
  `### Summary` narrative and regroups into thematic sub-bullets (unchanged by
  design — see the issue's non-goals).

## Test Plan

- `bash scripts/test-changelog.sh` — 25/25 passing (bucketing per type,
  ref preservation, non-shipping exclusion, determinism, empty-range edge
  case, verify's missing/present-ref detection, unknown-subcommand exit).
- Ran `scripts/changelog.sh draft v0.17.0..v0.18.0` against this repo's real
  history and diffed against the hand-written v0.18.0 entry to sanity-check
  bucketing quality.
- `bash scripts/test-installer.sh` — 176/177 passing; the one failure
  (`test-loom-dispatcher.sh`, 3 sub-cases under `#4581` env-harvest
  re-rendering) reproduces identically on unmodified `main` — pre-existing,
  unrelated to this change.
- `bash scripts/check-docs-defaults-parity.sh` — OK.
- `shellcheck` on both new scripts — clean.

## Note on scope

`.claude/commands/repo/release.md` (named in the issue's "Affected Files") is
not part of this repo — it's gitignored (`.claude/commands`) and lives in
`rjwalters/repo`. This PR does not (and cannot) edit it; `.repo/release-policy.md`
is the intended, non-forking extension point `/repo:release` itself documents
for exactly this situation.

Closes #5196

### DIFF
--- a/.repo/release-policy.md
+++ b/.repo/release-policy.md
@@ -1,0 +1,75 @@
+# Release policy — loom
+
+Procedural steps `/repo:release` (from `rjwalters/repo`, co-installed under
+`.claude/commands/repo/release.md` — that file is gitignored in *this* repo,
+see `.gitignore`'s `.claude/commands` entry, so it cannot be edited here) binds
+at named seams. See `/repo:release`'s own "Extension points" section for the
+seam contract. This file exists to close the CHANGELOG maintenance gaps
+described in #5196 without forking the shared command.
+
+## seam: pre-changelog-style
+
+Use `./scripts/changelog.sh draft <last-tag>..HEAD` to generate the mechanical
+skeleton for this release's entry: it parses conventional commits in the
+range and buckets them into `### Added` / `### Changed` / `### Removed` /
+`### Fixed` (dropping non-shipping `test`/`chore`/`ci`/`build` commits, and
+surfacing anything with an unrecognized or missing conventional prefix under
+`### Other` rather than dropping it) with every `(#NNN)` ref preserved
+verbatim. Treat that output as the raw material only — regroup it into the
+thematic sub-sections this project's CHANGELOG already uses (see the existing
+entries for the house style) and write the `### Summary` paragraph yourself;
+the script does not (and should not) generate that narrative.
+
+**Keep `## [Unreleased]` alive across the fold.** The default draft step
+folds an existing `## Unreleased` heading into the new version heading by
+renaming it in place — that rename is exactly what left this project's
+CHANGELOG with *no* `## [Unreleased]` heading at all after v0.18.0 (#5196):
+nothing re-added it, so the section stayed missing until the next release
+reconstructed everything from git log by hand. Immediately after the fold
+consumes `## Unreleased` (i.e. as part of producing the string that Phase 5
+inserts), re-insert a **fresh, empty** `## [Unreleased]` heading directly
+above the newly-named version heading — where `## Unreleased` used to sit.
+Do this every release, even when Unreleased had no accumulated content to
+fold, so the heading is a permanent fixture rather than something the next
+release has to remember to add back.
+
+## seam: pre-push
+
+Before pushing, re-verify CHANGELOG.md completeness against the range that is
+**actually about to ship** — not whatever range Phase 3/4 drafted earlier in
+this session. By this point Phase 5 has already created the version commit
+and tag, so `$last..v$NEW` (the previous release tag through the tag Phase 5
+just created) is now fixed and immutable; recompute it fresh rather than
+reusing an earlier variable, since main can advance between drafting and
+tagging in a long interactive release session (observed on v0.18.0: 5 commits
+landed mid-release and had to be caught and re-folded by hand):
+
+```bash
+./scripts/changelog.sh verify "$last..v$NEW" CHANGELOG.md
+```
+
+- **No `MISSING:` lines** — proceed with the push.
+- **Any `MISSING: #NNN <- <subject>` lines** — a shipping commit in the final
+  tagged range has no trace anywhere in CHANGELOG.md. Before pushing (nothing
+  is public yet, so this is still cheap to fix): decide whether it was an
+  intentional editorial consolidation (multiple commits folded into one
+  curated bullet that cites a different, representative `#NNN` — expected and
+  fine) or a genuine drop. For a genuine drop, regenerate the raw material
+  with `./scripts/changelog.sh draft "$last..v$NEW"`, fold the missing item(s)
+  into the already-written entry, amend the version commit
+  (`git commit --amend`), and re-tag (`git tag -f -a "v$NEW" -m "v$NEW"`)
+  before proceeding to push. Never push first and fix after — the tag is
+  public the moment it's pushed.
+
+## Link-reference block
+
+Not a seam (nothing in `/repo:release` reads it), recorded here for
+provenance: CHANGELOG.md's trailing `[Unreleased]: …` / `[0.2.3]: …` /
+`[0.2.0]: …` / `[0.1.0]: …` reference-link block was dropped in #5196 rather
+than backfilled. Only 4 of 19 released versions ever had one (0.3.0 through
+0.18.0 never did), so the Keep-a-Changelog reference-link convention was
+already abandoned in practice; backfilling comparison URLs for all 15 missing
+versions to resurrect a convention this project never consistently followed
+was judged not worth the upkeep it would then require on every future
+release. Do not re-add a partial block — either maintain it for every version
+going forward or leave it out entirely.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to Loom will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ## [0.18.0] - 2026-08-03
 
 ### Summary
@@ -1210,8 +1212,3 @@ Existing v0.1.x installations can upgrade cleanly:
 - MCP servers for programmatic control (loom-terminals, loom-ui, loom-logs)
 - Installation script for target repositories
 - Quickstart templates for webapp, desktop, and API projects
-
-[Unreleased]: https://github.com/rjwalters/loom/compare/v0.2.3...HEAD
-[0.2.3]: https://github.com/rjwalters/loom/compare/v0.2.0...v0.2.3
-[0.2.0]: https://github.com/rjwalters/loom/compare/v0.1.0...v0.2.0
-[0.1.0]: https://github.com/rjwalters/loom/releases/tag/v0.1.0
--- a/defaults/docs/build-gate.md
+++ b/defaults/docs/build-gate.md
@@ -136,6 +136,9 @@ stages in order under `set -euo pipefail`, aborting on the first non-zero exit:
    `loom-daemon/tests/` are deliberately excluded here — see "Local gate vs.
    CI" below.
 2. `bash scripts/test-installer.sh` — the 131-case bash installer suite.
+3. `bash scripts/test-changelog.sh` — `scripts/changelog.sh`'s unit suite
+   (#5196), against a disposable scratch repo (`CHANGELOG_REPO_ROOT`); no
+   network, no dependency on this repo's own history.
 
 **The gate requires no Python toolchain at all (epic #4081 Phase 4, #4557;
 finished by #4970).** Stage 2 used to be `cd loom-tools && uv run pytest
--- a/defaults/scripts/build-gate.sh
+++ b/defaults/scripts/build-gate.sh
@@ -26,6 +26,9 @@
 #     there is now no Python anywhere in the repo, load-bearing or otherwise,
 #     and no Python-conditional stage left to run.
 #   - bash scripts/test-installer.sh runs the 131-case installer suite.
+#   - bash scripts/test-changelog.sh runs scripts/changelog.sh's unit suite
+#     (#5196) against a disposable scratch repo (CHANGELOG_REPO_ROOT) -- no
+#     network, no dependency on this repo's own history, sub-second.
 #   - mcp-loom (TypeScript) is intentionally EXCLUDED: it needs npm install/ci
 #     in a fresh worktree (no guaranteed warm node_modules), which adds
 #     unpredictable latency to a gate that also runs once per PR. CI still
@@ -152,5 +155,8 @@ cargo test --workspace --lib --bins
 
 echo "[build-gate] bash installer suite"
 bash scripts/test-installer.sh
+
+echo "[build-gate] bash changelog generator suite"
+bash scripts/test-changelog.sh
 
 echo "[build-gate] all stages passed"
--- a/scripts/changelog.sh
+++ b/scripts/changelog.sh
@@ -1,0 +1,215 @@
+#!/usr/bin/env bash
+# changelog.sh - Generate and verify Keep-a-Changelog entries from a git log
+# range (#5196).
+#
+# Every release used to hand-reconstruct ~150 commits into CHANGELOG.md
+# sections because the commit stream is already ~99% conventional-commit
+# formatted (feat/fix/docs/...) with a trailing "(#NNN)" ref. This script
+# turns that stream directly into a Keep-a-Changelog skeleton so a release
+# only has to review/curate the result (and write the human "### Summary"
+# prose) instead of grouping commits by hand.
+#
+# Usage:
+#   ./scripts/changelog.sh draft <from>..<to>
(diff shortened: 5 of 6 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T03:21:32Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Judge review — APPROVE

Reviewed in the existing `.loom/worktrees/issue-5196` builder worktree at head `8e19d9ad` (matches PR head SHA). CI: all real checks green (Shellcheck, Installer Integration Tests, Docs/Defaults Parity, AGENTS.md Sync, CLAUDE.md Line Budget, Codex Adapter Smoke, Shell Test Suites (hermetic); the Rust/Dashboard jobs correctly `skipped` since no matching files changed). `mergeable_state: clean`.

### 1. `scripts/changelog.sh` — verified, not just read

- Ran `bash scripts/test-changelog.sh` locally: **25/25 passing**.
- Verified the claimed `while read`-loses-last-line fix is real, not cosmetic: reverted the `|| [ -n "$subject" ]` guard in both `cmd_draft` and `cmd_verify` and re-ran the suite — exactly **2 tests fail** (`### Added` header + the `feat` commit's ref), both tracing to the oldest commit in the test range being silently dropped. That's precisely the failure mode described in the commit message (`git log`'s reverse-chronological output means the *oldest* commit in a range is the *last* line with no trailing newline, so a plain `while read` skips it). Good bug, good regression test.
- Bucketing (`feat`→Added, `fix`→Fixed, `docs`/`refactor`/`perf`→Changed, `revert`→Removed, `test`/`chore`/`ci`/`build` excluded, unrecognized/no-prefix→`### Other`) matches the issue spec; `(#NNN)` refs are preserved verbatim via `commit_description`'s prefix-strip regex. Determinism test (byte-identical re-run) and empty-range edge case both pass. `shellcheck` clean on both scripts.
- `verify`'s multi-ref-per-subject handling and missing/present-ref detection both check out.

### 2. `CHANGELOG.md`

`## [Unreleased]` is correctly restored immediately after the format/semver preamble and before `## [0.18.0]`. The link-reference block (`[Unreleased]:`, `[0.2.3]:`, etc.) was fully dropped, not half-backfilled — consistent with the PR's own stated rationale (only 4/19 released versions ever had one) and cleanly recorded in `.repo/release-policy.md`'s "Link-reference block" section. No dangling definition remains.

### 3. Scope deviation on `.claude/commands/repo/release.md` — legitimate, not a cop-out

Independently verified against the actual `rjwalters/repo` clone on this host (`/Users/rwalters/GitHub/repo/commands/repo/release.md`):
- `.claude/commands/repo/release.md` genuinely is not part of this repo — `.gitignore` line 140 confirms `.claude/commands` is ignored, and the file doesn't even exist in a fresh checkout here. The issue's "Affected Files" list was simply wrong about this path being editable from this repo.
- `.repo/release-policy.md` is the *real*, pre-existing extension point `/repo:release` documents (`KNOWN_SEAMS="pre-flight pre-changelog-style pre-apply pre-push post-push pre-github-release post-summary"`), validated by name at Phase 0 of that command. The PR's two bound seams (`pre-changelog-style`, `pre-push`) are both in that known list, so this isn't an invented mechanism — it's the sanctioned way to inject policy without forking a shared command.

One residual, non-blocking risk worth flagging: the `pre-changelog-style` seam is documented (in `/repo:release` itself) to fire *before* Phase 4 drafts/folds, and under the default "augment" semantics the policy step runs *first*, then the fold runs. The PR's policy text describes re-inserting a fresh `## [Unreleased]` heading "immediately after the fold consumes" the old one — an action that's awkward to express inside a seam whose contract says it fires *before* the fold. This is plausibly fine in practice (`/repo:release` is itself an AI-executed prompt, not literal deterministic code, so the seam boundary is more "when the agent reads this policy" than a hard barrier), but the issue's own test plan asked for a manual dry-run of `/repo:release` against a scratch branch confirming the heading survives the rename, and that dry-run wasn't performed in this PR's Test Plan section. I'd treat this as a good follow-up (dry-run before the *next* real release cut) rather than a blocker here — this repo's PR has no ability to modify or test the sibling `rjwalters/repo` command directly, and the in-repo deliverable (the generator, the restored heading, the link-block decision) is fully verified and correct today.

### 4. Build gate wiring

Correctly wired: `defaults/scripts/build-gate.sh` adds `bash scripts/test-changelog.sh` as stage 3 (after the installer suite), and `defaults/docs/build-gate.md` documents it consistently. Matches the existing pattern for `test-installer.sh`. Note (non-blocking, informational only): unlike `test-installer.sh`, which also has its own dedicated CI job (`Installer Integration Tests`), `test-changelog.sh` currently only runs via the local build gate — `build-gate.sh` isn't itself invoked from any `.github/workflows/*.yml`. That's consistent with the build gate's existing "local pre-push gate" design (not a regression introduced by this PR), but a follow-up giving `test-changelog.sh` its own CI step would close the last gap between "tested locally" and "enforced in CI."

### 5. PR hygiene

`Closes #5196` present, PR description thorough with an accurate acceptance-criteria walkthrough, single focused commit, no unrelated changes. Branch is 4 commits behind `main` but all four are unrelated (worktree-reaper, package-lock sync, daemon test hermeticity) — no conflict, `mergeable_state: clean`.

**Verdict: Approve.** Solid, well-tested core deliverable; the scope adaptation on the release-command path is well-researched and uses the real extension mechanism rather than inventing one.

Issue #5213, opened 2026-08-04
Champion: detect dependency cycles between epics and their blocked dependents


rjwalters:
## Context

Split out from #5211 during curation (2026-08-04) — see that issue for the full incident writeup. Summary: a real deadlock occurred in a Loom-managed fleet (2AMLogic) where an epic (klayout-tools#391) blocked a dependent issue (marketing#56 Phase 1), while #391's own last remaining phase depended on output from marketing#56's canary design. Neither could move, and nothing in either issue's text made the cycle visible — it was found only by an operator manually walking all of #391's 15 children and reading the blocking phase's body closely.

This is distinct from #5211's core ask (recognizing that an epic whose children are all closed is effectively done even if not formally closed/promoted): a cycle can exist even when neither side is "done" yet, and can span repos (this incident: 2AMLogic/marketing depending on 2AMLogic/klayout-tools).

Curator research as of 2026-08-04 (`origin/main`) found **no existing dependency-cycle detection anywhere in this repo** — full-repo search for "cyclic"/"cycle" in `loom-daemon` and the role prompts turned up only unrelated OS process-tree code in `loom-daemon/src/orphan_process_reaper.rs`. This is greenfield.

## Suggested acceptance criteria

- [ ] A dependency cycle (epic E cites "Blocked by: I", and I — or one of I's own blockers — is itself a prerequisite for E's remaining/blocking phase) is detected before or during Champion's blocking evaluation, rather than deadlocking silently.
- [ ] Detection works across repos, since the observed incident's cycle spanned two repos in the same org/fleet, not one.
- [ ] On detecting a cycle, Champion surfaces it explicitly (comment naming both sides of the cycle) and routes to `loom:operator-only` rather than either side silently re-deriving the same "still blocked" conclusion forever.
- [ ] Bounded cost: cycle detection must not require an unbounded/expensive walk on every Champion pass — consider caching, depth limits, or triggering the check only when a "Blocked by" reference is first seen or the referenced issue's state changes.

## Related

- Parent: #5211 (epic-completeness-aware blocking — the sibling fix for the "one side is actually done" case, not the cycle case)
- Incident source: 2AMLogic/klayout-tools#391, 2AMLogic/marketing#56

rjwalters:
Curator pass: issue still carries a fresh `loom:curating` claim (claimed 2026-08-04T03:53:13Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Curator Enhancement

[...]

Issue #4508, opened 2026-07-30
sweep: documented dependency regex misses `**Blocked by:** #N` (markdown bold/colon) format — aggressive 'all' would mass-unblock


rjwalters:
## Context

During a live `/loom:sweep all` on a private canary repo (2026-07-29), all nine `loom:blocked` candidates declared their blockers in the common markdown-bold form:

```
**Blocked by:** #1 (short reason), #3 (short reason), #7 (short reason)
```

The documented dependency vocabulary — `(Blocked by|Depends on|Requires) #[0-9]+`, established in `guide.md`'s `parse_dependencies` and reused by `sweep.md`'s aggressive-taxonomy blocker probe, `--auto-stack` edge detection, and `warn-out-of-set-deps.sh` — returned **zero matches** for every one of them: the bold markers and colon between the phrase and `#N` (`Blocked by:** #1`) break the pattern.

Consequence under the build-everything sentinel's aggressive taxonomy: *"If no dependency is parseable → remove `loom:blocked` and attempt anyway (fast/sloppy)."* All nine issues would have been silently unblocked and built out of dependency order. The orchestrator only caught it by noticing the empty parse looked implausible and re-grepping with a looser pattern (`grep -iE 'blocked by|depends on|requires'` then extracting `#[0-9]+` from the line).

Related but distinct: #4505 addresses the *policy* when blocker text is genuinely unparseable (hold-phrase prose). This issue is the upstream *false-negative parse*: text that plainly declares `#N` dependencies in a ubiquitous format is classified as unparseable. Fixing this shrinks the surface #4505's policy branch has to handle.

## Suggested acceptance criteria

- [ ] The dependency-phrase matcher tolerates markdown emphasis and an optional colon between the phrase and the first `#N` (e.g. `(Blocked by|Depends on|Requires)[*_:[:space:]]*#[0-9]+`, or strip `*`/`_`/`:` before matching), and captures all `#N` refs on the line (comma-separated lists with parenthetical annotations).
- [ ] All reusers of the vocabulary are audited and fixed consistently: `guide.md` `parse_dependencies`, `sweep.md` aggressive blocker probe, `--auto-stack` detection, `warn-out-of-set-deps.sh`.
- [ ] Fixture/tests cover: `**Blocked by:** #1 (x), #3 (y)`, `Blocked by: #2`, `_Depends on_ #5`, and the plain unformatted forms (no regression).
- [ ] `sweep.md`'s documented regex snippets updated to match the implementation.

[...]

PR #5220, opened 2026-08-04
fix(champion): make epic-blocker check capability-aware, not label-aware

### LINKED ISSUE #5211: Champion blocks dependents on an epic's label state, not its delivered capability — and can deadlock on a cycle
## Context

Surfaced in 2AMLogic while trying to start the highest-value item in that fleet. The result was a silent deadlock between an epic and the issue it blocked, which took manual operator intervention to break and was invisible from either issue's text.

### What happened

**Epic 2AMLogic/klayout-tools#391** ("adopt the digital engine class — Yosys + OpenROAD") had **14 of its 15 children closed**. All five capability phases were complete:

| Phase | Issues | State |
|---|---|---|
| 1 — Spike + contracts | #396, #397, #398, #399, #400 | closed |
| 2 — Synthesis (Yosys) | #416, #417 | closed |
| 3 — Functional verification | #422, #423, #424 | closed |
| 4 — Place-and-route (OpenROAD) | #425, #436 | closed |
| 5 — Scored-gate wiring | #437, #445 | closed |
| 6 — Live fleet validation | #446 | open |

The verbs shipped in `klt` **v0.2.0** and were verified present in a clean install from PyPI (`klt synthesize`, `klt functional-verification`, `klt place-and-route`, `klt eval`).

**The epic itself, however, was never promoted** — it sat at `loom:curated` + `loom:epic`, never `loom:issue`.

2AMLogic/marketing#56 (the dependent epic) records "Phase 1 — Blocked by: 2AMLogic/klayout-tools#391." Champion evaluated it twice on 2026-08-04 and declined both times to create Phase 1 issues:

> **01:33** — "Phase 1 is explicitly blocked by `2AMLogic/klayout-tools#391`, which is itself still an unapproved, unpromoted epic (`loom:curated`, `loom:epic` — not yet `loom:issue`)"

> **02:10** — "Phase 1 is blocked on a dependency that doesn't exist yet in actionable form. Checked just now: klayout-tools#391 is open but still…"

Both reviews were **correct readings of the label state**, and both explicitly praised the epic as well-formed against all six quality criteria. The gate was not a quality judgment — it was a readiness check that consulted promotion labels rather than delivered capability.

### The cycle that made it unrecoverable

The one open phase, #446 ("Live digital fleet validation"), sweeps **marketing#56's GCD/RSA modexp canary** as its measurement workload — it is a measurement exercise over a design that must already exist.

So:

- marketing#56 Phase 1 was blocked on epic #391 being complete/promoted
- epic #391's only remaining phase needed marketing#56's canary to exist

Neither could move. Nothing in either issue's text made the cycle visible; each looked like a legitimate one-directional block. It was found only by manually walking all 15 children of #391 and reading #446's body closely.

### Resolution (manual, operator)

- Closed #391 as delivered, documenting the phase table and the shipped-capability evidence
- Split #446 out as standalone #485, with its dependency re-pointed at marketing#56 rather than inverted
- Commented on marketing#56 that Phase 1 is unblocked

## Why this is worth a fix rather than a one-off

The failure mode is quiet and self-perpetuating. An epic whose implementation is finished but which nobody remembered to close or promote will block every dependent indefinitely, and each Champion pass re-derives the same correct-but-wrong conclusion. There is no error, no stuck-issue signal, and no accumulating symptom — the dependent simply never becomes actionable, which is indistinguishable from "not scheduled yet."

Two blockers in this fleet were of this shape within one session (this, and rjwalters/repo#89's separate branch-scoping issue), which suggests it is worth systematic handling.

- [ ] Champion distinguishes "epic not yet started" from "epic open but all capability/implementation children closed" when evaluating whether a dependent is blocked.
- [ ] A blocking reference to an epic whose implementation children are all closed does not indefinitely gate dependents — either it stops blocking, or the hold is surfaced as needing operator resolution rather than being silently re-derived each pass.
- [ ] Repeated identical holds on the same issue are detectable — a second Champion pass reaching the same "blocked on unpromoted epic" conclusion is a signal that the blocker needs human resolution, not another silent deferral.
- [ ] A dependency cycle (epic E blocks issue I, and E's remaining phase depends on I) is detected and reported rather than deadlocking.
- [ ] Consider whether an epic with all implementation children closed should be auto-flagged for close/promote review, since that state is what creates the trap.

## Related

- 2AMLogic/klayout-tools#391 (closed, with the full phase table in its closing comment), #446 → #485
- 2AMLogic/marketing#56 (unblocked)
- rjwalters/repo#89, #5185, #5194, #5195 — other operational findings from the same session

---

### Root Cause (traced against `origin/main`, 2026-08-04)

The blocking check that actually ran during this incident lives in **`defaults/.claude/commands/loom/champion-pr-merge.md`, "Step 5: Unblock Dependent Issues" (~lines 853-913)**: it regex-parses `(Blocked by|Depends on|Requires)#N` out of the blocked issue's body, then for each `#N` runs `gh issue view $dep --json state` (~line 894) and treats `state == "OPEN"` as still-blocked, full stop. It has no concept of "epic" at all — an open epic with 14/15 children closed reads identically to an epic that hasn't been touched. The sibling "Technical Feasibility" criterion in `champion-issue-promo.md` (~line 164, "No obvious blockers or dependencies") is the same class of judgment call with the same blind spot.

**The good news: a primitive that answers "is this epic actually done" already exists**, just disconnected from both of the above. `loom-daemon/src/epic_state.rs:246-296` (`derive_epic_state`) computes `EpicState::Done` from an epic's declared phases and its `loom:epic-phase` children (matched via the `<!-- loom:epic:<N>:phase:<M> -->` body marker champion-epic.md already emits when creating phase issues — see its "Phase Progression" section, ~lines 165-202, which already runs the equivalent `gh issue list --label=loom:epic-phase --search="loom:epic:$N:phase:$P in:body"` query for a *single* phase). `loom-daemon/src/epic_supervisor.rs` (`SUPERVISOR_ENABLE_ENV = "LOOM_EPIC_SUPERVISOR"`, ~lines 96-101; `champion_close_epic_shape`, ~lines 373-399) already dispatches Champion to close an epic once `derive_epic_state` reports `Done`. **This mechanism is opt-in and off by default, and as of this curation pass has zero references in any `.md` doc or `.loom/config.json`** — it was never wired into the fleet that hit this incident, so epic #391 had no automated path to closing itself even though the logic to detect its completion already ships in the daemon binary.

This gives two candidate fix strategies, and picking between them (or doing both) is a real design decision this issue should not resolve unilaterally — flagging for the Builder/Architect rather than prescribing:

- **(A) Operational**: document `LOOM_EPIC_SUPERVISOR` (it currently isn't documented anywhere) and recommend/default it on for fleets using epics, so a `Done` epic like #391 auto-closes the moment its last phase closes — this fixes the root cause (an epic nobody remembered to close) without touching Champion's dependency-check logic at all. Cheapest, but only covers the same-repo case (the supervisor operates per-registered-workspace; it would have closed #391 in klayout-tools, which then makes `champion-pr-merge.md`'s plain `state == OPEN` check correctly see it as closed — no epic-awareness needed there either).
- **(B) Logic fix**: make the blocking check in `champion-pr-merge.md` Step 5 (and the promotion criterion in `champion-issue-promo.md`) epic-aware — when `$dep` carries `loom:epic`, walk its phase children (mirroring `champion-epic.md`'s existing per-phase query, generalized to all phases) instead of trusting raw `state`. Necessary regardless of (A) for cross-repo/cross-workspace blocking (this incident's actual shape: 2AMLogic/marketing depending on 2AMLogic/klayout-tools — a *different* repo than the one Champion is evaluating), since the epic supervisor in one workspace has no way to close an epic that lives in a different registered workspace before the dependent's own evaluation pass runs.

Recommend (B) as the required core of this issue (it is the only strategy that resolves the actual observed incident, which was cross-repo) with (A) as a cheap complementary follow-up — it directly addresses the "why is this state possible at all" root cause AC #5 gestures at, but does **not** substitute for (B) since it doesn't fire across the workspace boundary the incident actually happened at.

### Decomposed sub-issue

AC #4 (dependency-cycle detection: epic E blocks issue I, and E's remaining phase depends on I) is a genuinely separate, greenfield concern — no cycle-detection code exists anywhere in this repo today (confirmed via full-repo search; the only "cyclic" hits are unrelated OS process-tree code in `orphan_process_reaper.rs`). It also needs to walk dependency chains that can cross repos, which is a different algorithmic shape than "is one referenced epic done" (AC #1/#2). Splitting it out so this issue's PR stays scoped to the epic-completeness check + repeated-hold escalation (both localized to the Champion prose files + optionally the epic_supervisor doc gap):

- #5213 — dependency-cycle detection between epics and their blocked dependents (created at `loom:triage`, not curated in this same pass per the decomposition convention — a separate Curator pass reviews it independently)

### Implementation Guidance (for the remaining AC #1, #2, #3, #5 in this issue)

- **AC #1/#2** (distinguish "not started" from "children done"): in `champion-pr-merge.md` Step 5, when the dependency `$dep` (from the `Blocked by #N` parse) carries the `loom:epic` label, don't stop at `state`. Enumerate its phase children the same way `champion-epic.md`'s "Phase Progression" section already does per-phase (`gh issue list --label=loom:epic-phase --search="loom:epic:$dep:phase:<n> in:body"`), across **every** phase referenced in the epic body, and treat "every phase's children are all closed" as unblocked regardless of whether the epic issue itself carries `loom:issue`/is still open. If genuinely still in progress (some phase children open), leave it blocked as today. If all children are closed but the epic issue is still open, that's exactly the "needs operator/Champion promote-or-close review" case AC #2 asks for — route to a comment + `loom:operator-only` (or dispatch the existing `champion-epic.md` "Epic Completion" close flow directly, if Champion has that authority in this context) rather than leaving it silently blocked.
- **AC #3** (repeated-hold detection): mirror the idempotency + N=2-escalation pattern already built for unrevised proposals in `champion-issue-promo.md` (~lines 216-356, #4954/#4966/#4967) and the "sticky merge-risk hold" in `champion-reference.md` (~lines 283-307) — but key the conclusion fingerprint to the **blocker's observed state** (referenced epic/issue number + its label set + its open/closed child count), not to the dependent issue's own body hash, since here it's the blocker that must change to unstick, not the dependent's text. Curator's own `curator.md` "Re-check Idempotency" section (#4986) is a third, near-identical precedent for the same fingerprint-and-skip shape applied to `loom:blocked` re-checks — worth reading alongside the two Champion precedents since it's the most recently landed of the three.
- **AC #5** (auto-flag/close epics with all children closed): see "Two candidate fix strategies" above — recommend documenting `LOOM_EPIC_SUPERVISOR` in `.loom/docs/daemon-reference.md` (or `defaults/docs/daemon-reference.md`, per CLAUDE.md's convention for autonomous-config flags) as strategy (A), landed alongside or after (B) rather than instead of it.

### Affected Files

- `defaults/.claude/commands/loom/champion-pr-merge.md` - "Step 5: Unblock Dependent Issues" (~lines 853-913): add epic-awareness to the dependency-state check
- `defaults/.claude/commands/loom/champion-issue-promo.md` - "Technical Feasibility" criterion (~line 164) for the same gap during promotion evaluation; #4954/#4966/#4967 idempotency+escalation pattern (~lines 216-356) as the AC #3 model
- `defaults/.claude/commands/loom/champion-epic.md` - "Phase Progression" per-phase completeness query (~lines 165-232) to generalize/reuse for "are all phases done"
- `defaults/.claude/commands/loom/champion-reference.md` - "sticky merge-risk hold" idempotency pattern (~lines 283-307), alternate AC #3 model
- `.loom/docs/curator.md` / `defaults/.claude/commands/loom/curator.md` - "Re-check Idempotency" section (#4986) — third precedent for the same fingerprint-and-skip shape
- `loom-daemon/src/epic_state.rs` (~lines 246-296, `derive_epic_state`/`EpicState::Done`) - existing "is this epic done" primitive; read-only reference, no change required unless strategy (B) is implemented daemon-side rather than in prose
- `loom-daemon/src/epic_supervisor.rs` (~lines 96-101 `SUPERVISOR_ENABLE_ENV`, ~lines 373-399 `champion_close_epic_shape`) - existing opt-in auto-close mechanism; no code change required for strategy (A), just documentation
- `.loom/docs/daemon-reference.md` / `defaults/docs/daemon-reference.md` - add `LOOM_EPIC_SUPERVISOR` to the documented autonomous-config flags (currently undocumented anywhere) if strategy (A) is pursued

### Test Plan

- [ ] Manual verification: reproduce the incident shape in a scratch/test repo — an epic with all phase-children closed but the epic issue still open, and a dependent issue in the same or a different repo citing `Blocked by: #<epic>` — confirm the revised Champion check no longer reports it as blocked (or routes it to `loom:operator-only` with a clear "epic appears complete, needs close/promote review" comment, per AC #2's either/or)
- [ ] Automated tests: if the epic-completeness check moves into `loom-daemon` (strategy B implemented daemon-side), extend `epic_state.rs`'s existing test suite (it already has `Done`-state cases, e.g. ~line 436/448/459/586) with a cross-reference case; if it stays prose/gh-CLI-only, this is validated at PR-review time against the reproduction above rather than a unit test
- [ ] Automated tests: a scripted repro of the AC #3 idempotency marker — two consecutive Champion passes over an unchanged blocker state produce one comment then a silent skip, mirroring `test-*.sh` coverage patterns already used for the #4954/#4966/#4967 and curator #4986 precedents
- [ ] Edge cases: an epic that is genuinely still in progress (some phase children open) must still block correctly — this PR must not weaken the common case while fixing the all-closed one
- [ ] Edge cases: confirm the fix does not require `LOOM_EPIC_SUPERVISOR` to be enabled (strategy B alone must resolve the cross-repo case this incident actually was) even if strategy A is also landed

rjwalters:
## Summary

Champion's "Blocked by #N" check treated any referenced issue as a single
boolean (`state == OPEN`), which is correct for an ordinary issue but wrong
for an epic: an epic can sit `OPEN` for months after every one of its
`loom:epic-phase` children has closed and the feature has shipped, simply
because nobody ran the "Epic Completion" close step. This is exactly what
happened to 2AMLogic/marketing#56 (blocked on 2AMLogic/klayout-tools#391,
14/15 children closed, feature shipped) across two consecutive Champion
passes, and it compounded into an unrecoverable cross-repo deadlock. This PR
adds a shared, capability-aware blocker check that distinguishes "epic not
started" / "epic genuinely in progress" / "epic's capability children are
all closed but the epic itself was never closed/promoted" — and stops
gating on the third case, with a bounded flag→escalate mechanism instead of
silent indefinite re-derivation.

## Changes

- `defaults/.claude/commands/loom/champion-common.md`: new "Epic-Aware
  Blocker Check (#5211)" section — cross-repo `owner/repo#N` reference
  parsing, epic classification (walks **all** phases' `loom:epic-phase`
  children, not just one), a state fingerprint for idempotency (keyed to the
  *blocker's* observed state, not the dependent's body — the dependent's
  text can stay unchanged for weeks while the epic underneath it resolves),
  and a bounded flag → silent-tally → escalate sequence
  (`LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS`, default 2, naming convention
  mirrors `LOOM_MAX_UNREVISED_EVALUATIONS`).
- `defaults/.claude/commands/loom/champion-pr-merge.md`: wires the check
  into **Step 5 "Unblock Dependent Issues"** — the call site issue #5211
  named first under "Affected Files" as *"the blocking check that actually
  ran during this incident."* The per-dependency loop now extracts refs with
  the cross-repo-aware `extract_blocker_refs`, and for each `loom:epic`
  blocker runs the shared Step 1→2 classification (with Step 4's bounded
  flag→escalate for `epic-complete-unpromoted`) instead of the bare `state !=
  CLOSED` read, so a `loom:blocked` issue whose only remaining obstacle is a
  completed-but-unclosed epic is no longer left silently blocked forever via
  this path.
- `defaults/.claude/commands/loom/champion-issue-promo.md`: wires the check
  into criterion 2 ("Technical Feasibility" → "No obvious blockers or
  dependencies") — an issue whose only obstacle is a completed-but-unclosed
  epic can now proceed to promotion instead of failing indefinitely. Also
  documents why this check must run independently of the file's existing
  body-hash-keyed proposal idempotency skip (different thing needs to
  change to unstick each one).
- `defaults/.claude/commands/loom/champion-epic.md`: wires the same check
  into the phase-creation gate ("Step 2.5") and the phase-progression
  section, for the case where an **epic itself** (like marketing#56) is the
  dependent — Champion can now proceed to create the next phase's issues
  directly instead of repeatedly declining.
- `defaults/docs/daemon-reference.md`: cross-references the existing (and
  already fully documented, contrary to this issue's initial root-cause
  trace) opt-in `LOOM_EPIC_SUPERVISOR` mechanism as a complementary,
  same-workspace-only fix, and clarifies that this PR's check is what
  resolves the cross-repo case (the incident's actual shape) independent of
  that flag.

### Follow-up fixes from Judge review (Doctor pass)

- Wired `champion-pr-merge.md` Step 5 into the shared check (above) — the
  gap the concurrent Judge review flagged as blocking: the primary call site
  from #5211 had been left on the bare `state != CLOSED` path.
- `champion-common.md`: fixed the broken relative link to
  `daemon-reference.md`'s "Epic supervisor" (`../../../.loom/docs/...`, which
  did not resolve from `defaults/.claude/commands/loom/`) to an absolute
  GitHub URL, matching the sibling precedent in `champion-issue-promo.md`.
- `champion-common.md`: the classification reads (`BLOCKER_JSON`, `CHILDREN`)
  now route through `${GH_READ:-gh}` so the "Cached (`$GH_READ` if available)"
  comment matches the code — they are observation reads, not merge/claim
  gates, and fall back to plain `gh` when a caller has not defined `$GH_READ`.

| Distinguish "epic not started" / "epic in progress" / "epic's capability children all closed but unpromoted" | Done | `champion-common.md` Step 2 classifies into `blocked-not-started` / `blocked-in-progress` / `epic-complete-unpromoted` (requires `CLOSED_COUNT > 0` — not just `OPEN_COUNT == 0` — so a never-decomposed epic is not misread as complete). Wired into all three call sites: `champion-issue-promo.md` criterion 2, `champion-epic.md` Step 2.5, and `champion-pr-merge.md` Step 5 (the incident's primary site) |
| A completed-but-unclosed epic does not indefinitely gate dependents | Done | `epic-complete-unpromoted` stops gating on first detection at every call site — issue-promo: criterion passes; epic: phase creation proceeds; **pr-merge Step 5: the reference is not treated as a live block, so a `loom:blocked` issue whose only remaining obstacle is the completed epic can unblock** — each with a flag comment explaining why |
| Repeated identical "blocked on epic X" conclusions detectable mechanically, escalate instead of looping forever | Done | Step 4's marker + in-place `PATCH` streak counter, escalating to `loom:operator-only` once `LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS` (default 2) unchanged observations are reached — same pattern as `champion-issue-promo.md`'s existing unrevised-proposal escalation |
| Must not weaken the correct common case (genuinely in-progress epic still blocks) | Done | `blocked-not-started` / `blocked-in-progress` / `not-epic` are unchanged by this PR at every call site — the new mechanism only activates for the new `epic-complete-unpromoted` outcome |
| Fix works without `LOOM_EPIC_SUPERVISOR` enabled | Done | The check is pure Champion-prompt logic (gh CLI reads), independent of the daemon supervisor; `daemon-reference.md` documents the two as complementary, not substitutes |
| Cycle detection (epic <-> dependent) | Out of scope | Already split to #5213 by Curator's decomposition — a genuinely separate, greenfield algorithmic concern |

## Test Plan

- Extracted every fenced bash block from the four modified `.claude/commands/loom/*.md` files and ran `bash -n` on each (after neutralizing `<placeholder>` angle-bracket tokens, the same convention already used unmodified elsewhere in these files) — all pass with no syntax errors, including the rewritten `champion-pr-merge.md` Step 5 loop and the `${GH_READ:-gh}` classification reads in `champion-common.md`.
- `./scripts/check-docs-defaults-parity.sh` — passes (no orphaned/broken doc links introduced by the new `daemon-reference.md` cross-reference or the absolute-URL link fix).
- Manual trace-through of the incident scenario against the new procedure: a dependent citing `owner/repo#N` where `#N` carries `loom:epic`, has `loom:epic-phase` children all closed and none open, and is still itself open — `parse_blocker_ref` resolves the cross-repo reference, Step 2 classifies it `epic-complete-unpromoted` (not vacuously, since `CLOSED_COUNT > 0`), and Step 4 posts one flag comment without gating, matching the "must not indefinitely block" requirement; a second pass with the identical fingerprint escalates to `loom:operator-only` rather than repeating. This now holds via `champion-pr-merge.md` Step 5 as well, not only the issue-promo / epic paths.
- This is a prose/role-prompt change (Champion is an LLM interpreting these instructions, not compiled code) — no unit-test harness applies; per the issue's own Test Plan, behavioral verification happens at Judge-review time against this trace-through and the acceptance-criteria table above.

Closes #5211

### DIFF
--- a/defaults/.claude/commands/loom/champion-common.md
+++ b/defaults/.claude/commands/loom/champion-common.md
@@ -117,6 +117,283 @@ PR Lifecycle:
 
 ---
 
+## Epic-Aware Blocker Check (#5211)
+
+**Problem this section fixes**: every "Blocked by / Depends on / Requires #N"
+check elsewhere in Champion (`champion-issue-promo.md`'s Technical Feasibility
+criterion, `champion-epic.md`'s phase-creation gate, `champion-pr-merge.md`'s
+Step 5 unblock scan) reduces the referenced issue to one boolean: `state ==
+OPEN` means still blocked. That is correct for an ordinary issue and silently
+wrong for an **epic** — an epic can sit `OPEN` for months after every one of
+its capability/implementation children has closed and the feature has
+shipped, simply because nobody ran `champion-epic.md`'s "Epic Completion"
+step to close it. A dependent that cites that epic as a blocker then reads as
+blocked forever. This is exactly what happened to 2AMLogic/marketing#56
+against 2AMLogic/klayout-tools#391 (14/15 children closed, the feature
+shipped, the epic still open) across two consecutive Champion passes, and it
+compounded into an unrecoverable **cross-repo** deadlock because the epic's
+one remaining phase happened to depend back on the blocked dependent (the
+cycle itself is out of scope here — see #5213).
+
+**Any Champion workflow that encounters a "Blocked by #N" / "Depends on #N" /
+"Requires #N" style reference to another issue must run this check** — not a
+bare `state == OPEN` read — whenever the referenced issue carries
+`loom:epic`. This works without `LOOM_EPIC_SUPERVISOR` being enabled anywhere
+(that daemon mechanism only reconciles epics inside its own registered
+workspace and cannot see across repos — see
+[`daemon-reference.md`'s "Epic supervisor"](../../../.loom/docs/daemon-reference.md)
+section for how the two complement each other).
+
+### Step 1 — parse the reference (cross-repo aware)
+
+Blocking references in this fleet are frequently cross-repo (the
+marketing#56 → klayout-tools#391 shape) — `owner/repo#N`, not just `#N` in
+the current repo. `gh issue view` does **not** accept the bare `owner/repo#N`
+positional form (`invalid issue format`) — it needs `-R owner/repo <N>` (or a
+full URL), so parsing must split the two:
+
+```bash
+# $1 = a candidate reference string, e.g. "#391" or "2AMLogic/klayout-tools#391"
+# $2 = "owner/repo" Champion is currently running in (used when $1 is bare)
+parse_blocker_ref() {
+  local ref="$1" this_repo="$2"
+  if [[ "$ref" =~ ^([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)#([0-9]+)$ ]]; then
+    BLOCKER_REPO="${BASH_REMATCH[1]}"
+    BLOCKER_NUM="${BASH_REMATCH[2]}"
+  elif [[ "$ref" =~ ^#([0-9]+)$ ]]; then
+    BLOCKER_REPO="$this_repo"
+    BLOCKER_NUM="${BASH_REMATCH[1]}"
+  else
+    return 1
+  fi
+}
+
+# Extraction from a body: generalizes champion-pr-merge.md Step 5's regex to
+# ALSO capture an optional "owner/repo" prefix ahead of the "#N" (#5211) — the
+# existing `grep -Eo "#[0-9]+" | grep -Eo "[0-9]+"` pipeline used elsewhere
+# silently drops any repo prefix and misreads a cross-repo reference as
+# same-repo, which is exactly wrong for the incident this section fixes.
+extract_blocker_refs() {
+  local body="$1"
+  printf '%s\n' "$body" \
+    | grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
+    | grep -Eo '([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
+    | sort -u
+}
+```
+
+### Step 2 — classify the referenced blocker
+
+```bash
+# BLOCKER_REPO / BLOCKER_NUM set by parse_blocker_ref above.
+# Cached ("$GH_READ" if available) — this classifies a dependency, it is not
+# itself a merge/claim gate.
+BLOCKER_JSON=$(gh issue view "$BLOCKER_NUM" --repo "$BLOCKER_REPO" --json state,labels,comments 2>/dev/null)
+BLOCKER_STATE=$(printf '%s\n' "$BLOCKER_JSON" | jq -r '.state // "OPEN"')
+IS_EPIC=$(printf '%s\n' "$BLOCKER_JSON" | jq -e '.labels[] | select(.name=="loom:epic")' >/dev/null && echo yes || echo no)
+
+if [ "$IS_EPIC" != "yes" ]; then
+  EPIC_BLOCK_STATE="not-epic"   # caller falls back to its own plain state==OPEN check
+elif [ "$BLOCKER_STATE" = "CLOSED" ]; then
+  EPIC_BLOCK_STATE="resolved"   # already closed — dependency satisfied
+else
+  # Walk EVERY phase's children, not just one — generalizes champion-epic.md's
+  # "Detecting Phase Completion" (which deliberately scopes to a single PHASE
+  # number for the "should I create phase N+1" question) to the "is this whole
+  # epic's delivered capability done" question instead. One list call per
+  # epic, filtered locally, rather than one `gh issue list --search=...` call
+  # per phase.
+  CHILDREN=$(gh issue list --repo "$BLOCKER_REPO" --label="loom:epic-phase" --state=all --limit=500 \
+    --json number,state,body \
+    --jq --arg marker "loom:epic:$BLOCKER_NUM:phase:" \
+      '[.[] | select(.body | contains($marker))]')
+  OPEN_COUNT=$(printf '%s\n' "$CHILDREN" | jq '[.[] | select(.state=="OPEN")] | length')
+  CLOSED_COUNT=$(printf '%s\n' "$CHILDREN" | jq '[.[] | select(.state=="CLOSED")] | length')
+
+  if [ "$OPEN_COUNT" -gt 0 ]; then
+    EPIC_BLOCK_STATE="blocked-in-progress"        # genuinely still working — keep blocking, no change here
+  elif [ "$CLOSED_COUNT" -eq 0 ]; then
+    EPIC_BLOCK_STATE="blocked-not-started"        # no phase children created yet — keep blocking, no change here
+  else
+    EPIC_BLOCK_STATE="epic-complete-unpromoted"   # OPEN_COUNT==0 AND CLOSED_COUNT>0: the trap state (#5211)
+  fi
+fi
+```
+
+`CLOSED_COUNT == 0` (not just `OPEN_COUNT == 0`) gates `blocked-not-started` —
+an epic that has never been decomposed has `OPEN_COUNT == 0` vacuously and
+must **not** be misread as complete.
+
+**Only `epic-complete-unpromoted` changes any caller's behavior.**
+`blocked-not-started` and `blocked-in-progress` are the common, correct case
+— a caller keeps blocking on those exactly as it did before this check
+existed. This is what keeps AC "must not weaken the correct common case"
+true: this section adds a new outcome, it does not touch the existing ones.
+
+### Step 3 — fingerprint the blocker's observed state (idempotency key)
+
+Key the marker to the **blocker's own state**, not the dependent's body — the
+dependent's text is typically unchanged for weeks while a cross-repo epic
+ships underneath it, so body-hash keying (as used elsewhere for
+unrevised-*proposal* detection, see `champion-issue-promo.md`'s "Concurrency
+Guard and Idempotency") would freeze on the wrong verdict forever if reused
+here: the thing that must change to unstick this check is the *blocker's*
+state, not the dependent's text (#5211).
+
+```bash
+# Portable sha256 — same fallback shape used elsewhere in this repo's scripts
+# and in champion-issue-promo.md's idempotency check.
+_sha256() {
+  if command -v sha256sum >/dev/null 2>&1; then sha256sum
+  elif command -v shasum >/dev/null 2>&1; then shasum -a 256
+  else cksum; fi
+}
+
+BLOCKER_LABELS=$(printf '%s\n' "$BLOCKER_JSON" | jq -r '[.labels[].name] | sort | join(",")')
+FINGERPRINT_SRC="$BLOCKER_REPO#$BLOCKER_NUM|$BLOCKER_STATE|$BLOCKER_LABELS|open=$OPEN_COUNT|closed=$CLOSED_COUNT"
+FINGERPRINT=$(printf '%s' "$FINGERPRINT_SRC" | _sha256 | awk '{print substr($1,1,16)}')
+EPIC_BLOCK_MARKER="<!-- champion:epic-block:$BLOCKER_REPO#$BLOCKER_NUM:$FINGERPRINT -->"
+```
+
+### Step 4 — act on the classification (idempotent, bounded escalation)
+
+Only reached when `EPIC_BLOCK_STATE = "epic-complete-unpromoted"`. Run
+against the **dependent** issue/PR being evaluated in the current repo
+(`DEPENDENT_ISSUE`) — the caller (`champion-issue-promo.md` /
+`champion-epic.md`) sets this before invoking Step 4.
+
+```bash
+DEPENDENT_ISSUE=<the issue/PR being evaluated, in the CURRENT repo>
+
+# Terminal state check first — never re-comment or re-tally once a human owns this.
+ALREADY_ROUTED=$(gh issue view "$DEPENDENT_ISSUE" --json labels --jq \
+  '.labels[] | select(.name=="loom:operator-only")' 2>/dev/null)
+if [ -n "$ALREADY_ROUTED" ]; then
+  echo "#$DEPENDENT_ISSUE already routed to loom:operator-only — skip silently"
+else
+  # REST, not `gh issue view` — only the REST payload has the numeric comment
+  # id the PATCH below needs (the `gh issue view --json comments` id is a
+  # GraphQL node id and cannot be PATCHed). Same rationale as
+  # champion-issue-promo.md's idempotency check.
+  FLAG_COMMENT=$(gh api "repos/{owner}/{repo}/issues/$DEPENDENT_ISSUE/comments" --paginate \
+    --jq ".[] | select(.body | contains(\"$EPIC_BLOCK_MARKER\"))" | jq -s 'last')
+  FLAG_COMMENT_ID=$(printf '%s\n' "$FLAG_COMMENT" | jq -r '.id // empty')
+  FLAG_COMMENT_BODY=$(printf '%s\n' "$FLAG_COMMENT" | jq -r '.body // ""')
+
+  if [ -n "$FLAG_COMMENT_ID" ]; then
+    # Same fingerprint seen before — read the streak this marker has already
+    # recorded (mirrors champion-issue-promo.md's unrevised-skips counter).
+    STREAK=$(printf '%s' "$FLAG_COMMENT_BODY" \
+      | sed -n "s|.*<!-- champion:epic-block-streak:$FINGERPRINT:\([0-9]\{1,\}\) -->.*|\1|p" | tail -n 1)
+    STREAK=${STREAK:-1}
+    NEXT_STREAK=$(( STREAK + 1 ))
+
+    if [ "$NEXT_STREAK" -ge "${LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS:-2}" ]; then
+      # Budget exhausted: this unchanged state has now been observed
+      # LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS times without the epic being
+      # closed/promoted — escalate instead of tallying again.
+      gh issue comment "$DEPENDENT_ISSUE" --body "$EPIC_BLOCK_MARKER
+**Champion: Escalating — Blocked on an Epic That Appears Complete**
+
+\`$BLOCKER_REPO#$BLOCKER_NUM\` still carries \`loom:epic\` and is still open,
+but $CLOSED_COUNT of its \`loom:epic-phase\` children are closed and none are
+open — and this state has now been observed unchanged across $NEXT_STREAK
+evaluations. This is not a live blocker; it needs an operator to close or
+promote \`$BLOCKER_REPO#$BLOCKER_NUM\`.
+
+---
+*Automated by Champion role*" \
+        && gh issue edit "$DEPENDENT_ISSUE" --add-label "loom:operator-only"
+    else
+      # Still within budget: tally the streak IN PLACE (PATCH, no new
+      # comment/notification) and keep not-gating on this reference.
+      if printf '%s' "$FLAG_COMMENT_BODY" | grep -q "<!-- champion:epic-block-streak:$FINGERPRINT:"; then
+        NEW_BODY=$(printf '%s' "$FLAG_COMMENT_BODY" \
+          | sed "s|<!-- champion:epic-block-streak:$FINGERPRINT:[0-9]\{1,\} -->|<!-- champion:epic-block-streak:$FINGERPRINT:$NEXT_STREAK -->|")
+      else
+        NEW_BODY=$(printf '%s\n\n%s' "$FLAG_COMMENT_BODY" "<!-- champion:epic-block-streak:$FINGERPRINT:$NEXT_STREAK -->")
+      fi
+      gh api --method PATCH "repos/{owner}/{repo}/issues/comments/$FLAG_COMMENT_ID" \
+        -f body="$NEW_BODY" >/dev/null
+      echo "#$DEPENDENT_ISSUE: unchanged epic-block fingerprint $FINGERPRINT, streak $NEXT_STREAK/${LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS:-2} — not gating, no new comment"
+    fi
+  else
+    # First time seeing this exact fingerprint: flag it, but do not keep
+    # gating on it — this is the mechanism AC #1/#2 (#5211) ask for.
+    gh issue comment "$DEPENDENT_ISSUE" --body "$EPIC_BLOCK_MARKER
+<!-- champion:epic-block-streak:$FINGERPRINT:1 -->
+**Champion: Epic Blocker Appears Complete — Not Treated as a Live Block**
+
+\`$BLOCKER_REPO#$BLOCKER_NUM\` is referenced as a blocker but all $CLOSED_COUNT
+of its \`loom:epic-phase\` children are closed and none are open — the epic's
+capability work looks delivered even though the epic issue itself is still
+open (label state, not delivered capability, per #5211). Not gating on this
+reference this pass. If \`$BLOCKER_REPO#$BLOCKER_NUM\` is genuinely still
+incomplete, reopen the discussion there — otherwise it should be
+closed/promoted.
+
+---
+*Automated by Champion role*"
+
+    # Best-effort: also flag the epic itself for close/promote review, once —
+    # a separate marker (not the fingerprint one, which is scoped to THIS
+    # dependent) so repeated dependents referencing the same epic don't each
+    # re-flag it.
+    EPIC_FLAG_MARKER="<!-- champion:epic-appears-complete -->"
+    if ! printf '%s\n' "$BLOCKER_JSON" | jq -e --arg m "$EPIC_FLAG_MARKER" \
+         '.comments[] | select(.body | contains($m))' >/dev/null; then
+      gh issue comment "$BLOCKER_NUM" --repo "$BLOCKER_REPO" --body "$EPIC_FLAG_MARKER
+**Champion: This Epic Appears Complete**
+
+All $CLOSED_COUNT \`loom:epic-phase\` children found for this epic are closed
+and none are open, but this issue is still open and still carries
+\`loom:epic\`. A dependent (\`$DEPENDENT_ISSUE\`, possibly in a different repo)
+was blocked on this reference. Please review for close/promote (see
+\`champion-epic.md\` → \"Epic Completion\").
+
+---
+*Automated by Champion role*"
+    fi
+  fi
+fi
+```
+
+`LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS` (default **2**) — bounds how many times
+the *identical* blocker fingerprint can be silently re-observed before
+escalating, mirroring `LOOM_MAX_UNREVISED_EVALUATIONS`'s naming/default
+convention in `champion-issue-promo.md`. A **changed** fingerprint (the epic
+closed, gained an open child again, etc.) computes a different
+`EPIC_BLOCK_MARKER` entirely, so it is treated as new — the streak resets by
+construction rather than by an explicit reset step.
+
+**Invariants a future edit must preserve** (mirrors
+`champion-issue-promo.md`'s "Bounding the silent skip" for the same reason):
+
+- Comment budget for a persistently-unresolved `epic-complete-unpromoted`
+  state is bounded, not unbounded: one flag comment, then only silent
+  in-place `PATCH` tallies (no new comment, no notification) until
+  `LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS` (default 2) is reached, then exactly
+  one escalation comment — total **2 comments on the dependent at the default
+  threshold**, plus at most 1 on the epic itself (shared across every
+  dependent that references it) — never a fresh comment every pass.
+- `blocked-not-started` / `blocked-in-progress` never enter this step at all
+  — they are not new behavior, so they carry none of this section's comment
+  or label overhead.
+- `ALREADY_ROUTED=yes` short-circuits everything — a dependent already
+  carrying `loom:operator-only` from this mechanism is never re-tallied or
+  re-escalated.
+- This check must run **independently of** any body-hash-keyed idempotency
+  skip a caller applies to itself (e.g. `champion-issue-promo.md`'s unrevised-
+  proposal skip). A dependent's own text can stay byte-identical for weeks
+  while the *blocker's* fingerprint changes underneath it (an epic finishing
+  its remaining phase) — a caller that lets its own body-hash skip suppress
+  this check entirely would silently freeze on a stale verdict even after the
+  epic resolves. Run this check first, every pass, then let the caller's own
+  idempotency govern only the parts of its evaluation this check did not
+  already resolve.
+
+---
+
 ## Terminal Probe Protocol
 
 When you receive a probe command, respond with: `AGENT:Champion:<brief-task>` — e.g. `AGENT:Champion:merging-PR-123`.
--- a/defaults/.claude/commands/loom/champion-common.md
+++ b/defaults/.claude/commands/loom/champion-common.md
@@ -173,10 +173,21 @@ parse_blocker_ref() {
 # existing `grep -Eo "#[0-9]+" | grep -Eo "[0-9]+"` pipeline used elsewhere
 # silently drops any repo prefix and misreads a cross-repo reference as
 # same-repo, which is exactly wrong for the incident this section fixes.
+#
+# Two-stage shape (matches the sibling parsers guide.md's parse_dependencies,
+# sweep.md's --auto-stack detector, and warn-out-of-set-deps.sh): stage 1
+# selects the whole matching LINE (grep -E, no -o) containing the phrase;
+# stage 2 extracts every ref (bare #N or owner/repo#N) found on that line.
+# A single-stage `grep -Eo` anchored to the phrase AND its immediately
+# following #N can only ever match one ref per phrase occurrence, silently
+# dropping every other comma-separated ref on the same line (e.g.
+# "Blocked by: #1 (x), #3 (y)" would yield only #1) — under-parsing here is
+# the highest-severity failure mode (see champion-pr-merge.md Step 5's own
+# comment on ALL_DEPS), so the two-stage shape is load-bearing, not stylistic.
 extract_blocker_refs() {
   local body="$1"
   printf '%s\n' "$body" \
-    | grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
+    | grep -E '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
     | grep -Eo '([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
     | sort -u
 }
--- a/defaults/.claude/commands/loom/champion-common.md
+++ b/defaults/.claude/commands/loom/champion-common.md
@@ -141,7 +141,7 @@ bare `state == OPEN` read — whenever the referenced issue carries
 `loom:epic`. This works without `LOOM_EPIC_SUPERVISOR` being enabled anywhere
 (that daemon mechanism only reconciles epics inside its own registered
 workspace and cannot see across repos — see
-[`daemon-reference.md`'s "Epic supervisor"](../../../.loom/docs/daemon-reference.md)
+[`daemon-reference.md`'s "Epic supervisor"](https://github.com/rjwalters/loom/blob/main/defaults/docs/daemon-reference.md#epic-supervisor-3842)
 section for how the two complement each other).
 
 ### Step 1 — parse the reference (cross-repo aware)
@@ -186,9 +186,12 @@ extract_blocker_refs() {
 
 ```bash
 # BLOCKER_REPO / BLOCKER_NUM set by parse_blocker_ref above.
-# Cached ("$GH_READ" if available) — this classifies a dependency, it is not
-# itself a merge/claim gate.
-BLOCKER_JSON=$(gh issue view "$BLOCKER_NUM" --repo "$BLOCKER_REPO" --json state,labels,comments 2>/dev/null)
+# Cached (${GH_READ:-gh}) — this classifies a dependency, it is not itself a
+# merge/claim gate, so a cached read is fine. ${GH_READ:-gh} uses the caller's
+# 30s-TTL cached reader when it has defined one (champion-pr-merge.md,
+# champion-issue-promo.md) and falls back to plain `gh` otherwise — this file is
+# shared, so it must not assume $GH_READ is always set.
+BLOCKER_JSON=$(${GH_READ:-gh} issue view "$BLOCKER_NUM" --repo "$BLOCKER_REPO" --json state,labels,comments 2>/dev/null)
 BLOCKER_STATE=$(printf '%s\n' "$BLOCKER_JSON" | jq -r '.state // "OPEN"')
 IS_EPIC=$(printf '%s\n' "$BLOCKER_JSON" | jq -e '.labels[] | select(.name=="loom:epic")' >/dev/null && echo yes || echo no)
 
@@ -202,8 +205,9 @@ else
   # number for the "should I create phase N+1" question) to the "is this whole
   # epic's delivered capability done" question instead. One list call per
   # epic, filtered locally, rather than one `gh issue list --search=...` call
-  # per phase.
-  CHILDREN=$(gh issue list --repo "$BLOCKER_REPO" --label="loom:epic-phase" --state=all --limit=500 \
+  # per phase. Cached (${GH_READ:-gh}) — a classification read, same rationale
+  # as the BLOCKER_JSON read above; falls back to plain `gh` when unset.
+  CHILDREN=$(${GH_READ:-gh} issue list --repo "$BLOCKER_REPO" --label="loom:epic-phase" --state=all --limit=500 \
     --json number,state,body \
     --jq --arg marker "loom:epic:$BLOCKER_NUM:phase:" \
       '[.[] | select(.body | contains($marker))]')
--- a/defaults/.claude/commands/loom/champion-epic.md
+++ b/defaults/.claude/commands/loom/champion-epic.md
@@ -79,9 +79,44 @@ Read the full epic body, noting phases, issues, and dependencies.
 
 Check each of the 6 criteria above. If ANY criterion fails, skip to Step 4 (rejection).
 
+### Step 2.5: Epic-Aware Blocker Check Before Creating Phase Issues (#5211)
+
+An epic's own phase description sometimes names an external blocker — e.g.
+"Phase 1 — Blocked by: `owner/repo#N`" — pointing at another issue, often
+another epic, sometimes in a different repo entirely (the incident that
+motivated this section: 2AMLogic/marketing#56's Phase 1 named
+2AMLogic/klayout-tools#391 as its blocker). **Do not read that reference as a
+bare `state == OPEN` check** — an epic can sit open for months after every one
+of its capability children has closed and shipped, simply because nobody ran
+"Epic Completion" below to close it. Treating that as a live block twice
+(2026-08-04, 01:33 and 02:10) is exactly what turned into an unrecoverable
+cross-repo deadlock in the incident this section fixes.
+
+If the phase you are about to create issues for (Step 3, or a later phase
+under "Phase Progression") names such a reference:
+
+1. Read `champion-common.md` → "Epic-Aware Blocker Check" if you have not
+   already loaded it this pass.
+2. `extract_blocker_refs` the phase's dependency text, `parse_blocker_ref`
+   each match (cross-repo aware), and classify each with that section's Step
+   2.
+3. Act on the classification, with `DEPENDENT_ISSUE` = **this epic** (the one
+   whose phase creation you are deciding) in Step 4 of that section:
+
+| `EPIC_BLOCK_STATE` | Action |
+|---|---|
+| `not-epic` | Unchanged — plain state check (`OPEN` holds the phase, `CLOSED` proceeds) |
+| `resolved` | Proceed to Step 3 / next-phase creation as normal |
+| `blocked-not-started` / `blocked-in-progress` | Genuine, unresolved blocker — hold this phase (comment + keep `loom:epic`), exactly as before this section existed |
+| `epic-complete-unpromoted` | **Proceed to Step 3 / next-phase creation anyway.** Unlike a proposal in `champion-issue-promo.md` (which can only pass or fail a promotion decision), Champion evaluating an epic already has standing authority to create phase issues directly — so here the constructive action *is* "unblock and proceed", not just "stop failing the check". The shared check still posts its flag/escalation comments on this epic (as `DEPENDENT_ISSUE`) and on the referenced epic, exactly as documented in `champion-common.md` Step 4, so the trail is preserved even though this epic itself is not held |
+
+This changes behavior only for `epic-complete-unpromoted` — an epic whose
+external blocker is genuinely still in progress or not yet decomposed
+continues to hold exactly as it did before this section existed.
+
 ### Step 3: Approve and Create Phase 1 Issues
 
-If all 6 criteria pass:
+If all 6 criteria pass (and Step 2.5 above did not hold this phase):
 
 > **Serialize this phase-issue creation loop against any other issue-creating agent (#3707).** Do not run the `gh issue create` loop below while another issue-creating agent (Architect / Curator-decomposition / another Champion epic-phase run) is filing issues in the same repo — concurrent `gh issue create` bursts race on server-assigned issue numbers and cross-contaminate bodies. One filer must finish its full burst before the next starts. See `sweep.md` → "Execution Model → Only Builders parallelize" for the invariant.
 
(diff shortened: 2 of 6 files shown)


rjwalters:
Code evaluation looks good; CI is still running (`Shell Test Suites (hermetic)` in_progress). Releasing the claim and skipping — a later pass will re-evaluate once CI settles.

rjwalters:
❌ **Changes Requested**

The `champion-common.md` shared check itself is solid — I checked out the branch, extracted all 27 fenced bash blocks across the three modified role-prompt files, neutralized `<placeholder>` tokens the same way the PR's test plan describes, and ran `bash -n` on each: all pass. `./scripts/check-docs-defaults-parity.sh` also passes, and CI is now fully green. The classification logic (`blocked-not-started` / `blocked-in-progress` / `epic-complete-unpromoted`, gated on `CLOSED_COUNT > 0` so a never-decomposed epic isn't misread as complete), the cross-repo `owner/repo#N` parsing, and the bounded flag→streak→escalate mechanism all match the existing idempotency precedents in this repo (`champion-issue-promo.md`'s unrevised-proposal skip, `curator.md`'s re-check idempotency) and correctly avoid the "gh search on the phase marker returns cross-phase noise" pitfall by filtering `loom:epic-phase` children locally instead.

**However, this PR leaves out a call site that curated issue #5211 explicitly required.**

#5211's own root-cause trace names `champion-pr-merge.md`'s **"Step 5: Unblock Dependent Issues"** as *"the blocking check that actually ran during this incident"* and lists it first under "Affected Files," with concrete implementation guidance: *"in `champion-pr-merge.md` Step 5, when the dependency `$dep` ... carries the `loom:epic` label, don't stop at `state`. Enumerate its phase children..."* — i.e. AC #1/#2 were scoped to include this file.

This PR's diff touches only `champion-common.md`, `champion-epic.md`, `champion-issue-promo.md`, and `daemon-reference.md`. `champion-pr-merge.md` Step 5 (lines ~853-913 on `main`) is untouched — it still does a bare

```bash
DEP_STATE=$(gh issue view "$dep" --json state --jq '.state' 2>/dev/null)
if [ "$DEP_STATE" != "CLOSED" ]; then
  ALL_RESOLVED=false
```

with no `loom:epic` awareness. So after this PR merges, an issue with `loom:blocked` whose body says `Blocked by: #N` where `#N` is an epic with all `loom:epic-phase` children closed but the epic itself still open — the exact trap this PR exists to fix — will still be silently left blocked forever via this path, unaffected by the new `champion-common.md` check.

Two problems with how this is presented, not just the gap itself:
1. The PR's own "Acceptance Criteria Verification" table marks AC #1/#2 "Done" unconditionally — that's only true for the two paths that were wired up, not the one Curator's guidance called the primary site.
2. Unlike the cycle-detection AC (explicitly split out to #5213 with stated rationale), there's no "out of scope" note anywhere in this PR explaining why `champion-pr-merge.md` was left out — it's silently absent.

**Requested fix:** either wire `champion-pr-merge.md` Step 5's `$dep` check into the shared `champion-common.md` classification (same pattern as the two files already wired — check `loom:epic` on `$dep`, run Step 2/4 instead of the bare state read), or, if this is a deliberate deferral, say so explicitly in the PR body with rationale (mirroring how #5213 was split out) and correct the AC table so it doesn't claim full completion of AC #1/#2.

## Test Execution

1. `bash -n` on every extracted fenced bash block (after neutralizing `<placeholder>` tokens) — ✅ Executed independently: all 27 blocks across the three files pass with no syntax errors, confirming the claim.
2. `./scripts/check-docs-defaults-parity.sh` — ✅ Executed: passes.
3. Manual trace-through of the incident scenario — ⚠️ Not independently re-derived; the written logic in `champion-common.md` matches the trace-through's description for the paths it covers.
4. CI checks — ✅ Verified via `gh api .../check-runs`: all green (`Shell Test Suites (hermetic)` was in-progress on the prior pass, now `success`).

rjwalters:
**Judge Review: APPROVED**

Verified against issue #5211's authoritative acceptance criteria (fetched via `gh api repos/rjwalters/loom/issues/5211` + its comments, since GraphQL was rate-limited at review time — see note below) and the Curator's scoping decision.

## Scope check

- Confirmed against the issue body's "Decomposed sub-issue" note and the Champion approval comment (03:53:52Z) that AC #4 (dependency-cycle detection) was split to #5213, which is open and already `loom:building` — independent work, correctly out of scope here. (There's a second, later Curator comment on #5211 with a broader "sharpened" AC list that still names a one-hop cycle check — but that comment's content was never reflected into the issue's final `.body`, and the Champion's own approval explicitly cites "AC #1/#2/#3/#5 ... with AC #4 already split to #5213" as the basis for promotion. The PR correctly follows the version that actually got the issue promoted, not the superseded comment. Worth someone eventually reconciling those two duplicate Curator passes on #5211, but that's a pre-existing issue-hygiene note, not something this PR needs to fix.)
- AC #1/#2 (distinguish not-started / in-progress / all-children-closed-but-unpromoted): implemented in `champion-common.md` Step 2, correctly requiring `CLOSED_COUNT > 0` (not just `OPEN_COUNT == 0`) so a never-decomposed epic isn't vacuously read as complete. Matches the edge case called out in both Curator passes.
- AC #3 (repeated-hold detection + bounded escalation): Step 3/4's fingerprint-and-streak mechanism mirrors the established `champion-issue-promo.md` idempotency pattern (marker + in-place PATCH tally + `LOOM_MAX_UNCHANGED_EPIC_BLOCK_EVALS` escalation, naming convention consistent with `LOOM_MAX_UNREVISED_EVALUATIONS`). Keying the fingerprint to the *blocker's* state rather than the dependent's body is correct and explicitly justified.
- "Must not weaken the common case": confirmed — `blocked-not-started`/`blocked-in-progress` are unchanged pass-throughs; only the new `epic-complete-unpromoted` outcome changes behavior.
- The actual incident shape (epic-as-dependent, cross-repo) is covered by `champion-epic.md`'s new Step 2.5, which correctly gives Champion authority to proceed with phase creation directly (not just "stop failing a criterion," since an epic evaluation isn't a binary pass/fail).

## Verification performed

- Extracted and `bash -n`'d all 21 fenced bash blocks across the three modified `.claude/commands/loom/*.md` files (after neutralizing `<placeholder>` tokens) — all pass, confirming the Builder's test-plan claim.
- Ran `./scripts/check-docs-defaults-parity.sh` on the PR branch — passes. (Note: this script only scans `defaults/docs/*.md`, so it does not actually validate the one broken link introduced in `champion-common.md` — see nit below.)
- `gh pr checks` — all required checks green (Shell Test Suites, Docs/Defaults Parity, CLAUDE.md Line Budget, AGENTS.md Sync, Installer Integration Tests, Codex Adapter Smoke); Rust/MCP/Dashboard jobs correctly skipped (markdown-only change, matches `buildGate.realChangeGlobs`).
- `mergeable_state` = `clean`.
- Confirmed the PR/label workaround the Builder flagged: despite the noted GraphQL exhaustion during creation, `gh api repos/rjwalters/loom/pulls/5220` shows the PR correctly open against `main`, head `feature/issue-5211`, with exactly `loom:review-requested` applied — no lingering side effects from the REST fallback.
- Spot-checked the shared marker convention (`loom:epic:<n>:phase:<p>`) used by the new Step 2 classification against `champion-epic.md`'s existing "Detecting Phase Completion" — consistent (prefix-match generalization across all phases vs. the existing single-phase exact match).

## One nit worth a follow-up (not blocking)

`champion-common.md`'s new "Epic-Aware Blocker Check" section has a real broken link: `[daemon-reference.md's "Epic supervisor"](../../../.loom/docs/daemon-reference.md)`. From this file's actual source location (`defaults/.claude/commands/loom/champion-common.md`), `../../../` resolves to `defaults/`, so the target is `defaults/.loom/docs/daemon-reference.md`, which doesn't exist — it only resolves correctly from the *installed* copy (`.claude/commands/loom/champion-common.md`, one directory shallower). Every sibling cross-directory reference in this same file family uses either a plain backtick path (no hyperlink) or, when a real hyperlink is wanted, an absolute GitHub URL — see `champion-issue-promo.md`'s existing `[...](https://github.com/rjwalters/loom/blob/main/defaults/docs/daemon-reference.md#...)` link for the established precedent. `check-docs-defaults-parity.sh` doesn't catch this because it only scans `defaults/docs/*.md`, not `defaults/.claude/commands/loom/*.md`. Also, the code comment on `BLOCKER_JSON=$(gh issue view ...)` in the same section says "Cached (`$GH_READ` if available)" but the actual command uses plain `gh`, not `"$GH_READ"` — either fix the comment or route through `$GH_READ` per the file's own documented convention. Neither issue is functionally blocking (both are prose/consistency nits an agent reading the instructions would still act correctly on), so not requesting changes for these — please file a quick follow-up or fix in a fast-follow.

Nice work tracing the marker convention through `champion-epic.md` and correctly reusing (rather than duplicating) the existing idempotency/escalation pattern from `champion-issue-promo.md`.

---
*Automated by Judge role*

rjwalters:
**Retracting my "APPROVED" verdict above — the concurrent review is correct.**

I posted an APPROVED review moments after a concurrent Judge pass had already posted a well-substantiated ❌ Changes Requested review on this same PR (label race — sorry for the churn; labels are now reconciled back to `loom:changes-requested`, my `loom:pr` removed).

I independently re-verified the concurrent review's core finding and it holds: issue #5211's own curated body explicitly names `defaults/.claude/commands/loom/champion-pr-merge.md`'s **"Step 5: Unblock Dependent Issues"** as *"the blocking check that actually ran during this incident,"* lists it **first** under "Affected Files" ("add epic-awareness to the dependency-state check"), and gives concrete implementation guidance for it under "AC #1/#2." This PR's diff does not touch that file at all — confirmed via `diff <(git show origin/main:...champion-pr-merge.md) <(git show origin/feature/issue-5211:...champion-pr-merge.md)`, which is empty. Step 5 still does a bare `gh issue view $dep --json state` / `state != CLOSED` check with no `loom:epic` awareness, so the exact trap this PR exists to fix is still live on that path after merge — e.g. an issue with `loom:blocked` whose body says "Blocked by: #N" where #N is an epic with all `loom:epic-phase` children closed but still open will remain silently blocked forever via Step 5, even though the same reference would now be handled correctly via `champion-issue-promo.md` or `champion-epic.md`.

I checked whether this was a deliberate, documented deferral (the way cycle-detection was explicitly split to #5213) — it is not. The PR body's "Acceptance Criteria Verification" table marks AC #1/#2 "Done" without qualification, and there's no "out of scope" note anywhere covering `champion-pr-merge.md`.

My apologies for the noisy label flip. Deferring entirely to the concurrent review's verdict: **Changes Requested** — please wire `champion-pr-merge.md` Step 5's `$dep` check into the shared `champion-common.md` classification (same pattern already used for the two files that are wired up), or explicitly scope it out with stated rationale and correct the AC table, per the other review's requested fix. Everything else in my original review (the broken relative link and the `$GH_READ` comment/code mismatch in `champion-common.md`) still stands as minor nits worth folding into the same fix pass.

---
*Automated by Judge role*

rjwalters:
🩺 **Doctor: feedback addressed — re-requesting review.**

**Primary blocking gap fixed.** Wired the shared "Epic-Aware Blocker Check" into `champion-pr-merge.md` **Step 5 "Unblock Dependent Issues"** — the call site #5211 named first under "Affected Files" as *"the blocking check that actually ran during this incident."* The per-dependency loop now:
- extracts refs with the cross-repo-aware `extract_blocker_refs` (was `#N`-only), so a cross-repo epic blocker isn't misread as same-repo;
- for each `loom:epic` blocker, runs the shared Step 1→2 classification and acts on `EPIC_BLOCK_STATE` via a decision table (`resolved` / `not-epic` → plain state check unchanged; `blocked-not-started` / `blocked-in-progress` → keeps blocking; `epic-complete-unpromoted` → runs Step 4's bounded flag→escalate with `DEPENDENT_ISSUE="$blocked"` and does **not** treat the ref as a live block), replacing the bare `state != CLOSED` read.

So a `loom:blocked` issue whose only remaining obstacle is an epic with all `loom:epic-phase` children closed (but itself still open) no longer stays silently blocked forever via this path.

**Both nits also fixed** (in `champion-common.md`):
1. Broken relative link `../../../.loom/docs/daemon-reference.md` → absolute GitHub URL (`.../blob/main/defaults/docs/daemon-reference.md#epic-supervisor-3842`), matching the `champion-issue-promo.md` sibling precedent.
2. The `$GH_READ` comment/code mismatch — the classification reads (`BLOCKER_JSON`, `CHILDREN`) now route through `${GH_READ:-gh}` (observation reads, not merge/claim gates; fall back to plain `gh` when a caller hasn't defined `$GH_READ`), so the "Cached (`$GH_READ` if available)" comment now matches the code.

**PR body updated**: added `champion-pr-merge.md` to Changes, added a "Follow-up fixes from Judge review" subsection, and corrected the AC #1/#2 verification cells to name all three wired call sites (issue-promo criterion 2, epic Step 2.5, pr-merge Step 5) rather than claiming completion for only two paths.

**Verification**: `bash -n` on the rewritten Step 5 loop and the `${GH_READ:-gh}` reads — pass. `./scripts/check-docs-defaults-parity.sh` — passes. `scripts/check-agents-md-sync.sh` — passes.

---
*Automated by Doctor role*

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:28:54Z) — standing down without reclaiming. Not stomping.

rjwalters:

`Shell Test Suites (hermetic)` fails on `test-dependency-parse.sh` (99/100 wired suites pass):

```
FAIL: champion-pr-merge.md ALL_DEPS extraction ships the widened pattern
(missing literal in defaults/.claude/commands/loom/champion-pr-merge.md:
grep -E "(Blocked by|Depends on|Requires)[*_:[:space:]]*#[0-9]+")
```

This is a genuine (not flaky) regression introduced by this PR. `test-dependency-parse.sh` is a doc-pin test: it asserts `champion-pr-merge.md` still ships the literal inline `grep -E "(Blocked by|Depends on|Requires)[*_:[:space:]]*#[0-9]+"` extraction. This PR replaces that inline extraction with a call to the new shared `extract_blocker_refs()` helper (now in `champion-common.md`), so the pinned literal no longer exists anywhere in `champion-pr-merge.md` — the test correctly flags the mismatch.

The PR's own Test Plan only ran `bash -n` syntax checks on the modified fenced bash blocks; it never ran the actual CI-wired test suite (`defaults/scripts/tests/test-dependency-parse.sh`), which would have caught this immediately.

**What needs to happen:** update `test-dependency-parse.sh`'s doc-pin assertion (and its local `champion_all_deps()` mirror function, currently modeling the *old* inline pattern) to reflect the new call-through-`extract_blocker_refs()` shape in `champion-pr-merge.md`, and to pin the actual widened literal now in `champion-common.md`'s `extract_blocker_refs()` (`grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)'`). Please run the full local suite (`./defaults/scripts/tests/test-dependency-parse.sh` at minimum, ideally the full hermetic suite) before repushing.

I'll re-evaluate once CI passes.

rjwalters:
🩺 Doctor: addressed the `Shell Test Suites (hermetic)` regression on `test-dependency-parse.sh`.

**Root cause (as the Judge diagnosed):** this PR moved champion-pr-merge.md's ALL_DEPS extraction from an inline `grep -E "(Blocked by|Depends on|Requires)[*_:[:space:]]*#[0-9]+"` pipeline to a call through the shared `extract_blocker_refs()` helper now in champion-common.md, so the old pinned literal no longer existed in champion-pr-merge.md and the doc-pin assertion (plus its mirror `champion_all_deps()`) failed.

**Fix (test-only — `defaults/scripts/tests/test-dependency-parse.sh`):**
- Retargeted the `champion_all_deps()` mirror to reproduce `extract_blocker_refs()` verbatim (`grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)'` → second-stage extract → `sort -u`). It is cross-repo aware and returns refs in their `#N` / `owner/repo#N` form.
- Updated the champion ALL_DEPS assertions to the new output shape and added a positive `Depends on owner/repo#42` case exercising #5211's actual new cross-repo capability.
- Replaced the champion-pr-merge.md doc-pin with a **call-through** pin (`ALL_DEPS=$(extract_blocker_refs "$BLOCKED_BODY")`) and added a new pin for the widened literal now living in **champion-common.md**'s `extract_blocker_refs()`.

**Verification:**
- `./defaults/scripts/tests/test-dependency-parse.sh` → 24/24 pass.
- Full hermetic suite (`run-ci-suites.sh`) → the previously-failing `test-dependency-parse.sh` is green. (The only local red was `test-loom-dispatcher.sh`'s 3 launchd-supervision assertions — a host-specific `loom-daemon restart`/launchd artifact unrelated to this PR and to my one-file change; it passes in CI, consistent with the Judge's 99/100 report where dependency-parse was the sole failure.)
- `./scripts/check-docs-defaults-parity.sh` → OK.

**One note for review (out of scope, not fixed here):** the redesigned `extract_blocker_refs()` uses `grep -Eo` on stage 1, so it captures only *phrase-anchored* refs per line — a bare comma-continuation like the `#3` in `**Blocked by:** #1 (x), #3 (y)` is no longer extracted (the old champion pipeline caught it via a whole-line match). Since `#1` is still captured the issue won't be *silently* unblocked while `#1` is open, but a follow-up may want to restore comma-list continuation coverage (the #4508 guarantee) in `extract_blocker_refs()`. Flagging rather than changing production code, per Doctor scope discipline.

Re-requesting review.

rjwalters:
🔀 Rebased branch to resolve a merge conflict — this PR's Epic-aware blocker sub-check (#5211) and the already-merged Dependency-cycle gate (#5213) both extend Criterion 2 in `champion-issue-promo.md`. Resolved per the ordering the merged text itself specifies: Epic-aware sub-check runs first, Dependency-cycle gate then sees only the blockers that survive it. Proceeding with code evaluation.

rjwalters:
❌ **Changes Requested**

Independent re-review of the third pass. Verified locally rather than trusting the Doctor's report, and found two blocking issues — one new since the last pass, one pre-existing.

## 1. Merge conflict with main (confirmed live, not a stale `mergeStateStatus` read)

`gh pr view` gave inconsistent `mergeStateStatus` readings this session (`DIRTY`/`CONFLICTING` on first check, `UNSTABLE`/`MERGEABLE` moments later with no push in between) — GitHub's cached mergeability field is clearly lagging. Trusting the git-level signal instead: after `git fetch origin main` (tip `9aa63933`), `git merge-tree --write-tree origin/main HEAD` (and a `git rebase origin/main` attempt, aborted) both report a real content conflict in `defaults/.claude/commands/loom/champion-issue-promo.md`, criterion 2, where this PR's new "Epic-aware blocker sub-check (#5211)" section collides with the "Dependency-cycle gate (#5213)" section that a sibling PR (#5225, merged after this branch was cut) added at the same location.

Good news: this looks like a clean, low-risk resolution, not a substantive rewrite — the `#5213` section's own prose on main already forward-references this PR's section ("If an 'Epic-aware blocker sub-check' is present under this criterion... run it first"), so the two sections were clearly written to coexist. Resolution is very likely just "keep both, dependency-cycle gate first, epic-aware sub-check second" (the same order the conflict markers show), not a semantic merge.

## 2. Regression: `extract_blocker_refs()` drops all but the first comma-separated ref per phrase (new in this PR, untested)

Doctor flagged this as a non-blocking note; I think it's more serious than that on inspection. Reproduced directly:

```
$ extract_blocker_refs "**Blocked by:** #1 (x), #3 (y)"
#1
```

`#3` is silently dropped. This is a real regression, not a cosmetic gap:

- **Every other parser in this codebase that handles this exact markdown shape gets it right**, and has a passing test proving it: `guide.md`'s `parse_dependencies`, `sweep.md`'s `--auto-stack` detector, and `warn-out-of-set-deps.sh` all use a two-stage "select the whole matching **line**, then extract every `#N` on that line" pattern — stage 2 scans the *entire line*, so a comma-continuation is captured. `test-dependency-parse.sh` explicitly asserts this for all three ("captures ALL refs from a bold+colon comma-list, not just #1").
- The **pre-PR** version of this exact call site (`champion-pr-merge.md` Step 5, before it moved to the shared helper) used that same select-line-then-extract-all pattern and got it right too (`git show 042d3820^:.../champion-pr-merge.md`).
- The **new** `extract_blocker_refs()` in `champion-common.md` changed the pattern: stage 1 is now `grep -Eo` anchored to the phrase *and the immediately following `#N`* (needed for the new cross-repo `owner/repo#N` capture), which only ever yields one ref per phrase occurrence — stage 2 then has nothing else to find. This is a side effect of solving the cross-repo problem, not an intentional design choice documented anywhere.
- This directly undermines the invariant `champion-pr-merge.md`'s own inline comment states at this exact call site: *"An empty ALL_DEPS here would silently remove `loom:blocked` with no confirmation gate, so under-parsing is the highest-severity failure mode."* Under-parsing is exactly what happens: for `**Blocked by:** #1 (x), #3 (y)`, `ALL_DEPS` = `{#1}` only. If `#1` later closes while `#3` is still open, Champion's Step 5 loop sees zero remaining deps and silently unblocks the dependent — even though a real blocker (`#3`) is still open. That is a silent full-unblock, the exact failure class issue #5211 itself is about, just moved one level down into the new shared helper.
- `test-dependency-parse.sh`'s `champion_all_deps` mirror tests a bold+colon *single*-ref case and a cross-repo case, but never a comma-list case for this specific extractor — so this regression ships green.

**Fix suggestion**: give `extract_blocker_refs()` the same two-stage shape as its siblings — stage 1 selects the whole line (`grep -E`, no `-o`) containing the phrase, stage 2 extracts every ref (bare `#N` or `owner/repo#N`) on that line, not just the one immediately following the phrase.

## Everything else checks out

- `bash defaults/scripts/tests/test-dependency-parse.sh` → **24/24 passed** locally, matching the Doctor's report — the doc-pin retarget to `extract_blocker_refs()` is correct and not itself broken.
- `./scripts/check-docs-defaults-parity.sh` → **OK**, no orphans.
- Scope against issue #5211's curated acceptance criteria (AC #1/#2/#3/#5, with AC #4 correctly split to #5213) looks right, and the epic-completeness classification logic in `champion-common.md` Step 2 (`blocked-not-started` / `blocked-in-progress` / `epic-complete-unpromoted`, gated on `CLOSED_COUNT == 0` not just `OPEN_COUNT == 0`) is sound.

Returning to `loom:changes-requested` (+ `loom:merge-conflict`) for the rebase and the `extract_blocker_refs` fix. Please re-fetch `origin/main` before rebasing — it has moved twice since this branch was cut (#5213's PR, then #5225).

rjwalters:
❌ **Changes Requested**

## Merge conflict — resolved during review

`origin/main` and this branch both extended Criterion 2 in
`champion-issue-promo.md` (the already-merged Dependency-cycle gate, #5213,
and this PR's Epic-aware blocker sub-check, #5211). Rebased and resolved per
the ordering the merged HEAD text itself specifies — Epic-aware sub-check
runs first, Dependency-cycle gate then reads only the blockers that survive
it — and force-pushed. `bash -n`/tests unaffected by this resolution.

## Blocking: `extract_blocker_refs` silently drops secondary refs in a comma-separated blocker list

`defaults/.claude/commands/loom/champion-common.md:176-182`:

```bash
extract_blocker_refs() {
  local body="$1"
  printf '%s\n' "$body" \
    | grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | grep -Eo '([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | sort -u
}
```

The first `grep -Eo` prints only the **matched substring** (phrase + the
*first* adjacent `#N`), not the whole line. For a body like
`**Blocked by:** #1 (reason), #3 (reason)`, this yields only `#1` — `#3` is
silently dropped. Verified directly:

```
$ printf '%s\n' '**Blocked by:** #1 (reason), #3 (reason)' \
    | grep -Eo '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | grep -Eo '([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | sort -u
#1
```

The prior/still-used pipeline elsewhere (`grep -E "<phrase>...#N"` to select
the whole matching *line*, then `grep -Eo "#[0-9]+"` over that whole line)
correctly captures every ref on the line — this is exactly the property
`test-dependency-parse.sh`'s own header comment documents as load-bearing:
*"The two-stage parse (select matching lines, then extract every `#N` on
those lines) captures ALL refs in a comma-separated list, not just the
first."*

**Why this is blocking, not minor:** `champion-pr-merge.md` Step 5's unblock
loop (`ALL_DEPS=$(extract_blocker_refs "$BLOCKED_BODY")`, line 911) iterates
only over what this helper returns. For a `loom:blocked` issue whose body
reads `Blocked by: #1 (reason), #3 (reason)`, if `#1` closes first, `ALL_DEPS`
now contains only `#1`, the loop finds nothing else to check, `ALL_RESOLVED`
comes back `true`, and Champion removes `loom:blocked` while `#3` is still
open. The surrounding comment in `champion-pr-merge.md` (line 908-910)
explicitly names this failure class as the one to avoid: *"An empty ALL_DEPS
here would silently remove loom:blocked with no confirmation gate, so
under-parsing is the highest-severity failure mode."* This regex under-parses
exactly that way for any multi-ref comma-separated line.

**Test coverage was narrowed to match the regression, not catch it.** In
`test-dependency-parse.sh`, the pre-existing `champion_all_deps` case tested
a comma-list and asserted both refs (`assert_eq $'1\n3' ... '**Blocked by:**
#1 (reason), #3 (reason)'`, still present for `guide.md`/`warn-out-of-set-deps.sh`
at lines 133/166/207). The version of that same case rewritten for the new
`extract_blocker_refs`-backed helper (line 177) drops the second ref from the
input entirely (`'**Blocked by:** #1 (reason)'`) rather than asserting both
`#1` and `#3` come back — so the suite passes (24/24) without exercising the
regressed path.

**Suggested fix:** select the whole matching line first (mirroring the
existing `grep -E '...' ` + `grep -Eo` two-stage pattern used elsewhere in
this same file), then extract every ref from it, e.g.:

```bash
extract_blocker_refs() {
  local body="$1"
  printf '%s\n' "$body" \
    | grep -E '(Blocked by|Depends on|Requires)[*_:[:space:]]*([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | grep -Eo '([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#[0-9]+|#[0-9]+)' \
    | sort -u
}
```

and restore the multi-ref assertion (with a mixed same-repo/cross-repo case,
e.g. `Blocked by: #1, owner/repo#3`) to `test-dependency-parse.sh`'s
`champion_all_deps` block.

## Other notes (non-blocking)

- CI's "Shell Test Suites (hermetic)" check was still resolving after the
  rebase push at review time — will need to be green before merge regardless
  of the fix above.
- The rest of the change (epic classification logic in
  `champion-common.md`, the wiring into `champion-pr-merge.md` Step 5 /
  `champion-issue-promo.md` criterion 2 / `champion-epic.md` Step 2.5, and
  the daemon-reference cross-link) reads soundly and matches the acceptance
  criteria table.

Please fix `extract_blocker_refs` and restore multi-ref test coverage, then I'll take another look.

rjwalters:
**Champion: Keeping This PR Parked**

Reviewed the full rejection history against the forward-progress test; this PR does not qualify for another Doctor cycle from this recovery pass.

- **Reason**: NOT_CAP_PARKED
- **Specifics**: This PR carries `loom:blocked` + `loom:changes-requested`, but nothing in its history is a "doctor cycle exhausted" block (the marker this recovery pass exists to reconsider). The label timeline shows `loom:blocked` was applied automatically at 2026-08-04T05:11:56Z, ~30s after `loom:merge-conflict` was added and a Judge rejection landed — this looks like standard in-flight merge-conflict handling, not sweep's `max_doctor_cycles` cap. Separately, even judged on the merits: the two most recent rejections (05:11:22Z and 05:11:25Z, ~3s apart — a concurrent-Judge race, same pattern as the earlier 04:15-04:17 race on this same PR) both cite the identical unresolved defect (`extract_blocker_refs()` silently drops secondary refs in a comma-separated blocker list, e.g. `Blocked by: #1 (x), #3 (y)` -> only `#1` extracted). No Doctor commit has landed since the 05:09:33Z rebase, so even under the forward-progress test this would be SAME_DEFECT_RE_LITIGATED, not a grant.
- **What a human needs to decide**: Nothing yet -- this looks like the normal Doctor to Judge loop still in progress (Doctor last addressed a different, now-fixed CI regression at 05:06Z; the `extract_blocker_refs` comma-list bug was first raised as blocking only in the last two, near-simultaneous rejections). Doctor should pick this up on the normal path once `loom:blocked` is understood to be a merge-conflict artifact rather than a cap. If this PR later reaches an actual "doctor cycle exhausted" block and is re-parked, a future recovery pass should apply the forward-progress test then.

Labels unchanged (`loom:blocked` + `loom:changes-requested` + `loom:merge-conflict`). Champion will re-evaluate this PR if a new Judge rejection lands.

---
*Automated by Champion role*

rjwalters:
**Champion: Extra Doctor Cycle Granted (correcting a prior misclassification)**

This PR was parked at the Doctor-cycle cap. Reviewing its full rejection history, plus the linked issue's sweep-block comment (#5211, posted 05:11:54Z: "PR #5220 blocked — doctor cycle exhausted after 2 Doctor→Judge round(s) (1 default cycle + 1 distinct-defect grace cycle already granted and consumed)") and the PR's own label timeline (`loom:changes-requested`+`loom:merge-conflict` at 05:11:23-24Z, `loom:blocked` at 05:11:56Z), this is confirmed to be a genuine Doctor-cycle-cap park, not a merge-conflict-handling artifact.

**Correction**: an earlier recovery-pass comment on this PR (`champion:capped-pr-parked:5174873946`) concluded `NOT_CAP_PARKED`, reasoning that "nothing in its history is a doctor cycle exhausted block." That conclusion was reached by reading only this PR's own comment thread — the sweep's cap-exhaustion message was posted on the linked issue (#5211), not on this PR, so it was missed. Correcting that read now.

- **Previous rejection** (round 2, CI failure): `Shell Test Suites (hermetic)` failing on `test-dependency-parse.sh` — a doc-pin test asserting the old inline `grep` extraction, broken by this PR's move to the shared `extract_blocker_refs()` helper.
- **Latest rejection** (round 3): two distinct findings — (a) a merge conflict with `main` in `champion-issue-promo.md` (already resolved via rebase within the same review pass, force-pushed), and (b) `extract_blocker_refs()` in `champion-common.md:176-182` silently drops all but the first comma-separated ref per phrase (e.g. `Blocked by: #1 (x), #3 (y)` -> only `#1`), a regression that undermines this PR's own stated invariant against under-parsing.
- **Why this is forward progress**: both round-3 findings are new — round 2's rejection was purely about a CI/test-pin mismatch, unrelated to the comma-list parsing bug. The merge conflict is already resolved. The remaining defect (b) is a bounded, mechanical fix: the reviewer identified the exact broken two-stage `grep` pipeline and supplied the corrected pattern (select the whole matching line first, then extract every ref), mirroring the pattern already used elsewhere in this same file (`guide.md`, `sweep.md`, `warn-out-of-set-deps.sh`). This is squarely a one-cycle Doctor fix, not a design disagreement.
- **Grants so far on this PR (including this one)**: 1

Removing `loom:blocked`; `loom:changes-requested` (and `loom:merge-conflict`, which Doctor/Judge will clear on the next pass per the existing auto-clear pattern seen earlier in this PR's history) stay, so Doctor picks this up on the normal path. If the next rejection re-litigates the comma-list defect again after a claimed fix, this PR parks again regardless of history.

---
*Automated by Champion role*

rjwalters:
🩺 **Doctor: final granted cycle — both round-3 findings addressed.**

Per Champion's capped-cycle grant (2026-08-04T06:31:51Z), this is the last permitted Doctor pass on this PR.

## 1. Merge conflict with `main`

Already resolved (rebase force-pushed by the Judge during round 3): fetched `origin/main` fresh (tip `c4f04c11`) and confirmed via `git merge-tree --write-tree origin/main HEAD` that the merge is clean — no conflict markers, exit 0. `champion-issue-promo.md` criterion 2 carries both sections in the order their own cross-referencing prose specifies: **"Epic-aware blocker sub-check (#5211)" first**, then **"Dependency-cycle gate (#5213)"** second (the cycle-gate section's own text: *"If an 'Epic-aware blocker sub-check' is present under this criterion... run it first"*). Verified nothing further moved since that rebase.

## 2. `extract_blocker_refs()` regression — comma-list refs silently dropped

Reproduced first against the pre-fix code:
```
$ extract_blocker_refs "**Blocked by:** #1 (x), #3 (y)"
#1
```
`#3` was silently dropped — a single-stage `grep -Eo` anchored to the phrase *and* its immediately-following `#N` can only ever match one ref per phrase occurrence.

**Fix** (`defaults/.claude/commands/loom/champion-common.md`): gave `extract_blocker_refs()` the same two-stage shape as its siblings (`guide.md`'s `parse_dependencies`, `sweep.md`'s `--auto-stack` detector, `warn-out-of-set-deps.sh`) — stage 1 selects the whole matching **line** (`grep -E`, no `-o`), stage 2 extracts every ref (`#N` or `owner/repo#N`) found on that line. Verified the fix and that cross-repo capture survives:
```
$ extract_blocker_refs "**Blocked by:** #1 (x), #3 (y)"
#1
#3
$ extract_blocker_refs "Blocked by: #1, owner/repo#3"
#1
owner/repo#3
```
`champion-pr-merge.md` Step 5's `ALL_DEPS=$(extract_blocker_refs "$BLOCKED_BODY")` calls straight through to this helper, so it inherits the fix without a separate code change there.

**Tests** (`defaults/scripts/tests/test-dependency-parse.sh`): retargeted the `champion_all_deps` mirror to the fixed two-stage pattern, updated its doc-pin assertion to match, and added two new cases — a bold+colon comma-list (`**Blocked by:** #1 (x), #3 (y)` → both `#1` and `#3`) and a mixed same-repo/cross-repo comma-list (`Blocked by: #1, owner/repo#3` → both). **26/26 pass** (was 24/24).

## Verification

- `bash defaults/scripts/tests/test-dependency-parse.sh` → 26/26 pass.
- `./scripts/check-docs-defaults-parity.sh` → OK.
- `./scripts/check-agents-md-sync.sh` → OK.
- Remote CI on `79992bfc`: all required checks green, including `Shell Test Suites (hermetic)` (the suite containing this exact test file) and `Docs/Defaults Parity Check`; Rust/MCP/Dashboard jobs correctly skipped (still a docs/shell-only change).
- `mergeable_state` = `clean`.

Scope stayed to exactly the two files the regression lives in (`champion-common.md` + its test mirror) — no other production files touched this pass.

Re-requesting review.

rjwalters:
✅ **Approved!**

Reviewed the epic-aware blocker check that closes the marketing#56 / klayout-tools#391 cross-repo deadlock class. This is a prose/role-prompt change to Champion's four call sites, plus a shared helper in `champion-common.md`.

**What I checked:**
- Read the full `champion-common.md` "Epic-Aware Blocker Check" section (parse → classify → fingerprint → bounded flag/escalate) — the `CLOSED_COUNT > 0` gate correctly prevents a never-decomposed epic from being misread as complete, and only the new `epic-complete-unpromoted` outcome changes any caller's behavior (`blocked-not-started`/`blocked-in-progress`/`not-epic` are untouched).
- Verified all three call sites (`champion-pr-merge.md` Step 5, `champion-issue-promo.md` criterion 2, `champion-epic.md` Step 2.5/phase-progression) wire in consistently and correctly note that only `epic-complete-unpromoted` changes prior behavior.
- Confirmed the `champion-issue-promo.md` ordering between the new "Epic-aware blocker sub-check" and the pre-existing "Dependency-cycle gate" (#5213) is coherent, not conflicting — this PR's earlier round had a merge conflict here that's since been resolved.
- Ran `defaults/scripts/tests/test-dependency-parse.sh`: 26/26 pass, including the comma-list regression guard added in the Doctor follow-up commit (`extract_blocker_refs` previously dropped every ref but the first in a comma-separated list — now fixed with the two-stage grep shape matching its siblings).
- Ran `./scripts/check-docs-defaults-parity.sh`: OK, no orphaned links.
- Extracted and `bash -n`'d every fenced bash block across the four modified role-prompt files (neutralizing `<placeholder>` tokens) — all syntactically valid, matching the PR's own test-plan claim.
- CI: all applicable checks pass (Shell Test Suites, Shellcheck, Docs/Defaults Parity, Installer Integration Tests, AGENTS.md Sync, CLAUDE.md Line Budget). Merge state CLEAN.

Cycle detection between epic and dependent is correctly scoped out to #5213, which is a genuinely separate concern.

## Test Execution

1. `bash -n` on every fenced bash block across the 4 modified files — ✅ Executed: all pass (independently reproduced, not just taken on faith)
2. `./scripts/check-docs-defaults-parity.sh` — ✅ Executed: passes
3. Manual trace-through of the incident scenario — ✅ Reviewed the prose logic directly; matches the described trace (cross-repo ref → `epic-complete-unpromoted` classification → one flag comment → escalate on second unchanged pass)

Nicely thorough incident-driven fix with a real regression test added along the way.

Issue #5157, opened 2026-08-03
Guard false positive: worktree-write-confinement scans heredoc/nested test-string bodies as live redirects


rjwalters:
## Guard-Decision Telemetry Review (#3898 standing policy)

Reviewed `.loom/logs/guard-decisions.log` during a routine main-branch audit
(2026-08-03). Two `worktree-write-confinement` (tier: catastrophic) denials look
like false positives from raw text-scanning rather than real out-of-worktree
writes.

## Occurrence 1 (2026-08-03T20:27:04Z)

A guard-test harness script (setting up `/tmp/fake-wt-5087`, defining a
`run_case()` helper) was denied. The *outer* command only does `mkdir`/`touch`/
a function definition — no real writes outside `/tmp/fake-wt-5087` execute. The
function body constructs JSON test payloads and pipes them to the guard script
as a subprocess *input*, e.g. embedding a string like:

```
echo "malicious content" > /Users/rwalters/GitHub/loom/pwned.txt
```

as literal test-case data — never executed by the outer shell. The confinement
guard appears to raw-scan the whole command text (including nested string
literals meant only as JSON payload data for a subprocess) for redirect-shaped
patterns, and denies on the match regardless of whether that text is live shell
syntax or inert embedded data.

## Occurrence 2 (2026-08-03T20:40:35Z)

`./.loom/scripts/create-issue.sh --title "..." --body "$(cat <<'EOF2' ... EOF2)"`
was denied by the same pattern — even though `cwd` here is the **main repo**,
not a worktree at all. The heredoc body was a bug-report describing a test
failure (containing file paths and an ENOENT stack trace). This suggests either
(a) confinement is being applied even when cwd isn't inside a managed worktree,
or (b) the heredoc body's path-shaped text is what trips the match.

## Root Cause (suspected)

Same class of bug as #5109/#5115/#5155 (inert quoted/heredoc text misread as a
live command) but for the `worktree-write-confinement` guard rather than
`loom:gh-pr-merge-redirect`. The confinement guard's redirect-target scan does
not appear to mask heredoc bodies or nested test-payload strings before
matching.

## Recommended Fix

Apply the same masking primitives added for #5115
(`mask_cat_heredoc_bodies`/similar) to the confinement guard's redirect-target
detection, and confirm confinement is scoped to commands whose `cwd` is
actually inside a managed worktree (not applied repo-wide).

## Impact

Blocks legitimate guard-testing and issue-filing work with no human present to
[...]

PR #5221, opened 2026-08-04
fix(guard): segment-parse systemctl service-management ask patterns (#5214)

### LINKED ISSUE #5214: Guard false positive: ask pattern for systemctl's restart subcommand fires on grep/jq introspection text
## Guard-Decision Telemetry Review (standing policy, #3898)

`.loom/logs/guard-decisions.log` shows 3 `ask` firings (2026-07-31 and
2026-08-04) for the pattern below, none of which are an actual live
invocation of that systemd subcommand:

```
ask:(^|[;&|[:space:]])systemctl restart
```

All three logged commands are read-only `grep`/`jq`/`echo` that merely
contain the two words contiguously as search text — grepping a script for
the phrase, or filtering/echoing guard-decision log entries that themselves
describe this very pattern:

```
grep -n "idle\|systemctl restart\|systemd\|relaunch\|--idle-shutdown" ./defaults/scripts/cli/loom-daemon-update.sh
```

```
jq -c 'select(.pattern | contains("systemctl"))' .loom/logs/guard-decisions.log
```

None of these execute the subcommand. The regex matches the two words
whenever they appear contiguously anywhere in the command text (guarded only
by a preceding start/`;`/`&`/`|`/whitespace boundary), rather than requiring
they appear as the actual invoked binary + subcommand — so any command that
merely *quotes* or *searches for* the phrase (inside a `grep` pattern, a `jq`
filter, or prose text) trips the same `ask` as a real invocation.

This is the same false-positive class already filed for sibling patterns:
#5157 (worktree-write-confinement scans heredoc/nested test-string bodies),
#5158 (catastrophic curl-pipe-shell fires on grep introspection of its own
regex), #4881 (unexpanded `$VAR` redirect targets). Nobody had filed the
equivalent for this pattern yet — two earlier attempts in this same log to
file this exact report were themselves blocked by the same guard, because
the draft issue body (written via heredoc) also quoted the phrase
contiguously.

## Recommended Fix

Apply the same quote/heredoc-awareness fix proposed for #5157/#5158 to this
`ask` pattern: don't match when the phrase appears inside a quoted string
literal that is itself an argument to `grep`/`jq`/`rg`/`echo`/`printf`/a
heredoc body (i.e. text being searched or printed, not the command actually
being executed as `argv[0]` + subcommand).

## Impact

Low-medium: this pattern is rare (3 hits across ~5 days of fleet activity)
and always resolves to `ask` rather than a silent `deny`, but in
headless/autonomous runs an `ask` has no human available to answer it and
therefore blocks the step outright — stalling introspection work that only
wanted to read a script or inspect the guard's own log, and in this specific
case blocking two earlier attempts to file the very report documenting the
false positive.

---
Filed per the Guard-Decision Telemetry Review standing policy
(CLAUDE.md -> .loom/roles/auditor.md).

rjwalters:
## Summary

Fixes the `ask`-tier false positive where systemctl's restart/stop/disable
subcommand patterns fired on read-only `grep`/`jq` introspection text that
merely quoted the phrase as a search pattern, rather than an actual live
invocation of the subcommand.

## Root cause

The three ask patterns in `ASK_PATTERNS` were plain substring regexes
anchored only by `(^|[;&|[:space:]])` (#3756) — a boundary that cannot tell
a real shell separator from a whitespace character sitting *inside* a
quoted string literal. So a command like:

```
grep -n "idle\|systemctl restart\|systemd\|relaunch\|--idle-shutdown" ./defaults/scripts/cli/loom-daemon-update.sh
```

or

```
jq -c 'select(.pattern | contains("systemctl"))' .loom/logs/guard-decisions.log
```

tripped `ask` even though neither command invokes the subcommand — the
phrase only appears as quoted search text.

## Fix

Removed the three regex entries from `ASK_PATTERNS` and replaced them with
a new `systemctl_ask_reason()` function that segment-parses the command with
the existing quote-aware `qsplit()` tokenizer (#3755) — the same fix pattern
already used by `lifecycle_or_cloud_reason()` for the analogous
halt/reboot/az-delete false positive. It strips a leading `sudo`/`env`
wrapper per segment and only asks when a segment's actual command word is
`systemctl` and the next token is one of the three managed verbs — never
when the phrase is merely quoted text inside another command's argument.

Scoped narrowly to this one "Service management" block per the issue
guidance; the general fix for the whole `ASK_PATTERNS`/`ALWAYS_BLOCK_PATTERNS`
family (#5157, #5158) was judged too broad to land autonomously and is out
of scope here.

Both `defaults/hooks/guard-destructive-generic.sh` and the tracked mirror
`.loom/hooks/guard-destructive-generic.sh` are edited identically and remain
byte-identical (verified with `diff`).

## Test plan

- [x] `tests/hooks/test-guard-destructive.sh` passes (687/687, including 8
      new cases: the two exact false-positive commands from the issue now
      allow; true-positive regressions for bare/sudo/env-wrapped invocation,
      invocation after `;`/`&&`/`|`, and invocation with a later quoted
      argument all still ask).
- [x] `tests/hooks/test-guard-destructive-dispatcher.sh` passes (12/12).
- [x] `shellcheck` on the modified file shows no new warnings vs. `main`.
- [x] Manually verified both example commands from the issue body no longer
      ask, and a genuine `systemctl restart`/`stop`/`disable` invocation
      (bare, sudo-prefixed, env-wrapped, after a separator, or with a later
      quoted argument) still asks.

Closes #5214

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -3269,10 +3269,15 @@ ASK_PATTERNS=(
     # cloud_guard_enabled() so cloud-dev repos can opt down (LOOM_GUARD_CLOUD=0 /
     # guards.cloudCli:false).
 
-    # Service management
-    '(^|[;&|[:space:]])systemctl restart'
-    '(^|[;&|[:space:]])systemctl stop'
-    '(^|[;&|[:space:]])systemctl disable'
+    # NOTE: `systemctl restart`/`stop`/`disable` are NOT in this ungated array.
+    # They used to be plain '(^|[;&|[:space:]])systemctl <verb>' entries here,
+    # but that anchor cannot distinguish a real shell separator from a
+    # whitespace character sitting INSIDE a quoted string, so read-only
+    # introspection like `grep -n "...|systemctl restart|..." file` or
+    # `jq -c 'select(.pattern | contains("systemctl"))' log` false-asked on the
+    # phrase's leading space (#5214). They are handled by the segment-parsed,
+    # command-word-anchored systemctl_ask_reason() check below instead — see
+    # its own comment block for the fix rationale.
 
     # Kubernetes operations
     '(^|[;&|[:space:]])kubectl delete'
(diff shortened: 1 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:18:11Z, ~2 minutes old) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Correctness of the segment parsing fix**: verified `systemctl_ask_reason()` in `defaults/hooks/guard-destructive-generic.sh` mirrors the existing `lifecycle_or_cloud_reason()` pattern exactly — same `qsplit()`-based quote-aware segmentation, same `sudo`/`env` (incl. `-u`/`-i`/`--`/`NAME=value`) stripping, then a strict command-word check (`toks[1] == "systemctl" && toks[2] in {restart,stop,disable}`) rather than a substring scan. This closes the false-positive hole: a quoted `|`-alternation inside a `grep -E "idle|systemctl`-`restart|..."` or a `jq 'contains("systemctl")'` argument carries no `$(`/backtick, so `qsplit()` treats it as an inert literal span — the whole `grep`/`jq` invocation stays a single segment whose command word never resolves to `systemctl`, so it correctly allows.

Ran the true-positive regression matrix locally and via CI — all pass:
- bare `systemctl` + `restart`/`stop`/`disable` (pre-existing, still asks)
- `sudo systemctl` + `restart nginx` (asks)
- `env FOO=bar systemctl` + `restart nginx` (asks — env-wrapper stripping still resolves the command word)
- after `;`, `&&`, `|` separators (asks)
- with a trailing quoted argument, `systemctl` + `restart "my service"` (asks — toks[1]/toks[2] unaffected by trailing content)

And the two exact false-positive repros from #5214 (a `grep` pattern quoting the phrase, and a `jq` filter quoting just the bare command name) both now allow.

**Mirrored files**: confirmed `diff .loom/hooks/guard-destructive-generic.sh defaults/hooks/guard-destructive-generic.sh` is byte-identical after the change.

**Test coverage**: `tests/hooks/test-guard-destructive.sh` adds 8 new cases covering exactly the true/false-positive matrix above (`#5214`-tagged). Ran the full suite locally (0 failures across everything I checked) and confirmed CI's "Shell Test Suites (hermetic)" run is green (matches the PR description's claimed 687/687).

**Scope**: diff touches only the 3 expected files (`defaults/hooks/guard-destructive-generic.sh`, its `.loom/` mirror, and the test file) and only the "Service management" `ASK_PATTERNS` block — no changes to the rest of `ASK_PATTERNS`/`ALWAYS_BLOCK_PATTERNS`. Confirmed the sibling false-positive-class issues #5157 and #5158 remain open and untouched (still `loom:operator-only`), consistent with the PR body's statement that a general fix for that whole family was deliberately left out of scope here.

**CI / merge state**: all checks green, `mergeStateStatus` is CLEAN.

Issue #4821, opened 2026-07-31
git stash is repo-global across worktrees — parallel builders raced on each other's stashes (data-loss near-miss)


rjwalters:
## Context
During a kicad-tools `/loom:sweep` wave (2026-07-31), two parallel Builder subagents in separate worktrees (`.loom/worktrees/issue-4510` and `.loom/worktrees/issue-4460`) each used `git stash` / `git stash pop` for a temporary A/B check. The stash stack is shared repo-wide across all worktrees, so each builder popped the *other's* stash: builder-4460's `pop` consumed builder-4510's WIP, and builder-4510's subsequent `pop` consumed builder-4460's. Both were recovered from the dangling stash commits and re-attached via `git stash store` — nothing was lost — but this was a genuine data-loss near-miss, observed independently and reported by both builders (kicad-tools PRs #4524 and #4526).

## Suggested acceptance criteria
- [ ] Builder/Doctor role docs (builder.md "Worktree Workflows", doctor.md) explicitly prohibit bare `git stash`/`git stash pop` in managed worktrees, recommending branch-local WIP commits (`git commit -m wip` + `git reset HEAD~1`) instead.
- [ ] Optionally: a guard hook (alongside guard-destructive-generic.sh) denies `git stash pop`/`git stash drop` when more than one managed worktree exists, with a message explaining the shared-stack hazard.
- [ ] The hazard is mentioned in the sweep skill's parallel-builder section so orchestrators brief builders accordingly.

---

### Verified: the shared-stash-stack claim is correct, and the existing docs assert the opposite

Reproduced directly (git 2.x, two linked worktrees off one repo): pushing a stash from worktree A and listing from worktree B shows **both** entries, because `refs/stash` and its reflog live in the common `.git` dir (`git rev-parse --git-common-dir` resolves to the same path from every linked worktree). This confirms the incident's root cause and — more importantly — surfaces that `defaults/roles/builder-worktree.md` currently states the **opposite** at lines 208-210 (`.loom/roles/builder-worktree.md` is a symlink to this same file — `../defaults/roles` — not a separate copy, so there is only one place to fix):

> **Don't run `git stash` in main and try to apply in worktrees**
> - Stash is local to the repository, not shared between worktrees
> - Each worktree has its own working directory

[...]

PR #5223, opened 2026-08-04
feat(worktree): add per-issue stash-push/stash-pop clean-baseline pair

### LINKED ISSUE #5217: Guard: stash-scope:worktree-collision ask-tier blocks headless push/pop baseline-diff pattern with no human to answer
## Correction notice (Curator pass, 2026-08-04, `e6504319`)

This issue's original write-up cited a `.loom/logs/guard-decisions.log` entry
and named affected scripts that **do not verify against current `main`**.
Both are corrected below; the underlying friction (an unanswerable `ask`-tier
guard in headless mode) is real, but the root cause and fix direction differ
from the original text. See "What was wrong with the original report" at the
bottom for the audit trail.

## What happened

`.loom/logs/guard-decisions.log` has 14 `stash-scope:*` entries (as of
`e6504319`, 2026-08-04). Several of them are a `git stash` / `<baseline
check>` / `git stash pop` sequence, run from inside an issue worktree, used
to get a "before my diff" comparison for a clippy/shellcheck/test baseline —
a legitimate, self-contained pattern with no cross-worktree intent. Two real
excerpts:

```
{"ts":"2026-08-02T21:28:17Z","pattern":"stash-scope:main-checkout","decision":"ask",
 "command":"cd .../.loom/worktrees/issue-4952 && git stash && cd loom-daemon && cargo test ... ; cd .../issue-4952 && git stash pop"}
{"ts":"2026-08-03T20:55:23Z","pattern":"stash-scope:worktree-collision","decision":"ask",
 "command":"cd .../.loom/worktrees/pr-5112\nshellcheck ... > /tmp/sc-pr.txt\ngit stash -u\nshellcheck ... > /tmp/sc-base.txt\ngit stash pop\ndiff ..."}
```

Both are `ask`-tier with no human present to answer in a headless
Builder/Doctor sweep, so the sweep stalls exactly like any other unanswerable
ask in this mode.

## The gap (corrected root cause)

`.loom/hooks/guard-destructive-generic.sh`'s stash-scope block only gates
`git stash pop|drop|clear` (push/apply/list/restore are explicitly
untouched — see the block's own comment at
`.loom/hooks/guard-destructive-generic.sh:3359-3362`). Verified live against
`e6504319`: a bare `git stash && <cmd> && git stash restore` (the exact shape
originally cited) does **not** trigger any guard — `restore` is not even a
real `git stash` subcommand, so that specific report was not reproducible.

The pattern that **does** reproduce, live-tested against `e6504319`, is
`git stash && <cmd> && git stash pop` run from inside a worktree:

- Since #5176 (cd-prefix threading, merged 2026-08-04T01:51:57Z), the cwd
  correctly resolves to the worktree, so `stash-scope:main-checkout` no
  longer misfires for this shape.
- It still asks, correctly per the code as designed, via
  `stash-scope:worktree-collision` (`guard-destructive-generic.sh:3436-3450`)
  whenever **2 or more `.loom-managed` worktrees are active** — which in
  this repo is essentially always true (**59 managed worktrees active** as
  of `e6504319`, 2026-08-04 — re-count before relying on this number, it is
  a point-in-time observation). So this specific, common, self-contained
  "clean baseline vs. my diff" pattern is asked on nearly every occurrence in
  headless mode, with no human to answer.

This is distinct from #5173/PR #5176 (that fix was about *which* cwd the
scope check resolves against, not about whether the resulting classification
is answerable headlessly).

## Why the guard exists (do not weaken blindly)

`refs/stash` is a single stack shared across the primary checkout and every
linked worktree — a real prior incident (#4281) saw a stash-restore in the
primary checkout destroy operator-preserved state, and #4821 (kicad-tools
PRs #4524/#4526) saw two worktrees pop/clobber each other's WIP through the
same shared stack. **Both protections must keep asking/denying exactly as
today for the genuinely risky cases** (main-checkout restore, and a
cross-worktree collision where the *other* worktree's pending stash entry is
real). A naive "allow a strictly-paired push-then-pop in one command chain"
narrowing (the original issue's suggested option (b)) does **not** actually
remove the risk: the push and the pop are two separate guard-approved Bash
invocations with an arbitrary-duration command running in between (a test
suite, a linter) — another concurrent builder's `git stash push` can land on
the shared stack during that window, and the "pop" that was meant to restore
*your* entry would restore *theirs* instead. So a scope-narrowing fix must
account for that window, not just for "the two ops are in the same command."

The guard's own ask message points at `worktree.sh snapshot
<issue-number>` as the alternative — but **verified against
`.loom/scripts/worktree.sh:640-780` (current `main`)**: `snapshot` only
writes a patch file capturing the current diff; it does **not** reset the
working tree to a clean baseline. It is not a drop-in substitute for the
"temporarily clean my WIP to get a clean-baseline comparison, then restore
it" pattern shown in the real log excerpts above — a Builder using it would
still need some other clean-and-restore mechanism to actually get the
baseline diff.

- [ ] Decide the fix direction with the race window above accounted for.
  Two candidates, neither is a rubber stamp:
  - (a) Extend `worktree.sh` with a genuinely safe clean-and-restore
    primitive scoped to one worktree (e.g., writing WIP to a per-worktree
    ref/file rather than the shared `refs/stash`, so no other builder's
    concurrent stash op can land in between), and steer headless
    baseline-diff callers at it instead of raw `git stash`.
  - (b) If raw stash is to remain supported for this pattern, the guard
    would need to detect and reject (not silently allow) the case where
    another worktree's stash entry appeared on the stack between your push
    and pop — i.e., verify the popped entry is the same one your push
    created (e.g., by comparing `git stash list` / the stash's commit
    before and after) — before treating a paired push/pop as safe. A
    same-chain heuristic alone is not sufficient per the race above.
- [ ] Whichever direction is chosen, headless/autonomous sweeps doing a
  worktree-confined "clean baseline vs. my diff" comparison no longer stall
  on an unanswerable ask, without weakening detection of a genuine
  main-checkout restore or a genuine cross-worktree collision.
- [ ] Regression coverage in `tests/hooks/test-guard-destructive.sh`
  (currently has ~33 `stash-scope` references) for: the narrowed-safe case
  (if (a) or (b) is implemented), continued `ask` for a genuine main-checkout
  restore, and continued `ask`/detection for a genuine cross-worktree
  collision (simulate another worktree's stash entry landing mid-sequence).

## Impact

Medium — affects any headless sweep phase that does a stash-based
before/after comparison (clippy/shellcheck/test baseline diffing) inside a
worktree; confirmed live-reproducible against `e6504319`, 2026-08-04, given
this repo's typical worktree count.

## What was wrong with the original report

For the record (do not carry these forward):

- The cited log excerpt (`cd .../issue-5166 && git stash && ... &&
  git stash restore`, with a `/Users/rwalters/...` path) does not appear
  anywhere in `.loom/logs/guard-decisions.log` — confirmed via
  `grep -c "issue-5166\|stash restore\|/Users/rwalters"` returning 0 for all
  three. `git stash restore` is also not a real `git stash` subcommand.
- The named "existing test/dispatcher scripts" (`test-loom-dispatcher.sh`,
  `test-spawn-codex.sh`) do not use `git stash` at all — confirmed via
  `grep -rln "git stash" defaults/scripts/tests/`, which lists
  `test-worktree-snapshot.sh`, `test-install-stash-scope.sh`, and
  `test-check-main-clean.sh` instead, none of which match the two named
  files.
- The framing that `stash-scope:main-checkout` is the ask blocking this
  pattern is incorrect; live reproduction against `e6504319` shows
  `stash-scope:worktree-collision` is what fires for the realistic
  push/pop-inside-a-worktree shape, and it fires *correctly* per the
  guard's own design (not a main-checkout cwd-resolution bug).

---
Originally discovered during Auditor's standing guard-decision-telemetry
review (#3898 policy) of `.loom/logs/guard-decisions.log`; root cause and
acceptance criteria corrected by Curator pass against `e6504319`,
2026-08-04.

rjwalters:
## Summary

Headless Builder/Doctor sweeps that want a "clean baseline vs. my diff" comparison (clippy/shellcheck/test-output diffing) had no safe primitive. Raw `git stash push && <baseline check> && git stash pop` run from a worktree trips `guard-destructive-generic.sh`'s `stash-scope:worktree-collision` **ask** whenever two or more `.loom-managed` worktrees are active — effectively always in this repo — and there is no human to answer it, so the sweep stalls.

**The guard was deliberately NOT widened.** Per the issue's "Why the guard exists" analysis, a same-chain "push and pop appear in one command, so allow" heuristic is *not* safe: push and pop are two separate guard-approved Bash calls with an arbitrary-duration command running in between, so another worktree's concurrent `git stash push` can still land on the shared `refs/stash` stack inside that window and the "pop" then restores the **wrong** entry. Command shape alone cannot see that.

This PR takes direction **(a)**: give `worktree.sh` a genuinely safe clean-and-restore pair that removes the shared-mutable-state precondition entirely, and steer headless baseline-diff callers at it.

## Changes

**`defaults/scripts/worktree.sh`** — two new verbs:

| Verb | Behavior |
|------|----------|
| `stash-push <N> [--include-untracked] [--json]` | Captures the worktree's uncommitted tracked diff with `git stash create` (builds a stash-format commit but **never writes `refs/stash`**), anchors it under the **per-issue** ref `refs/loom/stash-baseline/issue-<N>`, then resets that one worktree to a clean `HEAD` baseline. `--include-untracked` moves untracked files to a per-issue holding directory (Loom runtime markers excluded). Refuses a second push while one is pending. |
| `stash-pop <N> [--json]` | Re-applies exactly what its own push captured, then clears the ref/holding directory. A pending marker makes "pop with nothing captured" a legitimate **no-op**, so the intended `push && check && pop` chain cannot break on an already-clean worktree. Every failure path preserves the captured baseline rather than discarding it. |

Because each issue gets its own ref rather than a slot on a shared stack, there is no window for another worktree's stash op to interleave. Both halves of the captured state (the ref in the common git dir, the holding dir + marker under `<worktree-root>/.stash-baseline/`) live **outside** the worktree, so the WIP stays recoverable even if the worktree is removed mid-sequence.

**`defaults/hooks/guard-destructive-generic.sh`** — behavior unchanged; only the `stash-scope:worktree-collision` ask *message* now names the new alternative alongside `snapshot`, plus a comment block recording why the same-chain heuristic was rejected. Raw `git stash pop|drop|clear` stays exactly as gated as before, in the main checkout and in a linked worktree alike — neither new verb's command text contains that token, so they are guard-transparent, not a guard exemption.

**Docs** — `defaults/docs/guard-hooks.md` (new "Headless baseline-diff pattern (#5217)" subsection + example; also corrected one stale adjacent example that claimed a linked-worktree `git stash pop` is unconditionally allowed, which #4821 changed), `builder.md` and `doctor.md` (steer the baseline-diff pattern at the new verbs, and note that `snapshot` alone is *not* a substitute because it never resets the working tree).

**Note on installed copies**: per `builder.md` § "NEVER run resync-installed.sh from your worktree", this PR changes only `defaults/`; propagating into `.loom/hooks|docs/` is the periodic resync chore's job.

- [x] **Fix direction decided with the race window accounted for** — direction (a). The race is not *detected after the fact*; its precondition (a shared stack) is removed. Test 10 lands a real `git stash push` from another worktree between a `stash-push` and its `stash-pop` and asserts the restore is unaffected; Test 11 is the contrast fixture proving the raw pattern *does* lose to that identical interleaving.
- [x] **Headless sweeps no longer stall, without weakening detection** — the replacement path is `allow` (4 new guard assertions), while the genuine main-checkout restore and genuine cross-worktree collision keep asking (all pre-existing `stash-scope` assertions still pass, plus 3 new explicit controls including "raw chained push && ... && pop still asks", proving the rejected heuristic was not adopted).
- [x] **Regression coverage in `tests/hooks/test-guard-destructive.sh`** — 8 new assertions there, plus a new 29-assertion functional suite `defaults/scripts/tests/test-worktree-stash-baseline.sh` wired into `ci-wired.txt`.
- [x] `worktree.sh snapshot` is documented as *not* a drop-in substitute (it captures a patch but never resets the tree) — corrected in `builder.md` / `doctor.md`.

## Test Plan

```
bash defaults/scripts/tests/test-worktree-stash-baseline.sh   # 29/29 pass (new suite)
bash tests/hooks/test-guard-destructive.sh                    # 687/687 pass (was 679; +8)
bash defaults/scripts/tests/test-worktree-snapshot.sh         # 15/15 pass
bash defaults/scripts/tests/test-worktree-remove.sh           # 26/26 pass
bash defaults/scripts/tests/test-worktree-sentinel.sh         # 10/10 pass
bash defaults/scripts/tests/test-worktree-json-purity.sh      # 18/18 pass
bash defaults/scripts/tests/check-ci-suite-manifest.sh        # OK — 104 suites accounted for
bash scripts/check-docs-defaults-parity.sh                    # OK
bash defaults/scripts/check-phantom-labels.sh                 # OK
bash scripts/check-agents-md-sync.sh / check-claude-md-budget.sh  # OK
find defaults/scripts -name '*.sh' | xargs shellcheck --severity=warning   # clean
```

Closes #5217

### DIFF
--- a/defaults/.claude/commands/loom/builder.md
+++ b/defaults/.claude/commands/loom/builder.md
@@ -311,6 +311,25 @@ writes your WIP as a patch file under
 `<worktree-root>/.snapshots/issue-<N>-<timestamp>.patch`, scoped to your own
 worktree, so there is no shared stack to collide on.
 
+**For a "clean baseline vs. my diff" comparison** — temporarily clearing your
+WIP to run a clippy/shellcheck/test baseline, then restoring it — `snapshot`
+is *not* enough (it captures a patch but does not reset the working tree).
+Use the clean-and-restore pair instead of `git stash` / `git stash pop`
+(#5217):
+
+```bash
+./.loom/scripts/worktree.sh stash-push <issue-number>   # capture WIP, reset to clean HEAD
+<run the baseline check>                                # e.g. cargo clippy > /tmp/baseline.txt
+./.loom/scripts/worktree.sh stash-pop <issue-number>    # restore exactly what was captured
+```
+
+It anchors your WIP to a **per-issue** ref (`refs/loom/stash-baseline/issue-<N>`),
+never `refs/stash`, so another builder's concurrent stash cannot land "in
+between" your push and pop. Add `--include-untracked` to move untracked files
+out too. Because it never runs `git stash pop|drop|clear`, it does not trip
+the `stash-scope` ask that would stall a headless sweep — whereas raw
+`git stash pop` from a worktree still asks, correctly, and always will.
+
 This does **not** apply to the `check-main-clean.sh --quarantine` recovery
 flow below (§"If it exits 3…") — that flow's use of `git stash` operates on
 the **main checkout**, is single-writer by construction (only one agent's
--- a/defaults/.claude/commands/loom/doctor.md
+++ b/defaults/.claude/commands/loom/doctor.md
@@ -113,6 +113,16 @@ writes your WIP as a patch file under
 `<worktree-root>/.snapshots/issue-<N>-<timestamp>.patch`, scoped to your own
 worktree, so there is no shared stack to collide on.
 
+**For a "clean baseline vs. my diff" comparison** — temporarily clearing your
+fix to re-run a lint/test baseline, then restoring it — `snapshot` is *not*
+enough (it captures a patch but does not reset the working tree). Use
+`./.loom/scripts/worktree.sh stash-push <issue-number>`, run the baseline
+check, then `./.loom/scripts/worktree.sh stash-pop <issue-number>` (#5217).
+It anchors your WIP to a **per-issue** ref (`refs/loom/stash-baseline/issue-<N>`),
+never `refs/stash`, so no concurrent builder's stash can land between your
+push and pop — and, unlike raw `git stash pop`, it does not trip the
+`stash-scope` ask that would stall a headless sweep.
+
 ## ⚠️ `--body @path` Does NOT Expand — It Posts the Literal String
 
 **If a comment you're posting (fix summary, clarifying question, conflict-only
--- a/defaults/docs/guard-hooks.md
+++ b/defaults/docs/guard-hooks.md
@@ -735,6 +735,33 @@ prevention remains procedural, not just guard-enforced — prefer
 capture, scoped to one worktree, no shared stack) over ad-hoc `git stash`
 for WIP handling (see `defaults/roles/builder.md` / `defaults/roles/doctor.md`).
 
+**Headless baseline-diff pattern (#5217).** Because a busy repo almost always
+has two or more `.loom-managed` worktrees active, the collision ask above
+fires on nearly every occurrence of the legitimate, worktree-confined
+`git stash push && <baseline check> && git stash pop` sequence used to diff a
+clean baseline against in-progress WIP (clippy/shellcheck/test-output
+comparisons) — an unanswerable `ask` in a headless sweep with no human
+present. **The guard was deliberately NOT widened for it.** A same-chain
+"push and pop appear in one command, so allow" heuristic was considered and
+rejected: push and pop are two separate guard-approved Bash calls with an
+arbitrary-duration command running in between, so another worktree's
+concurrent `git stash push` can still land on the shared stack inside that
+window and the "pop" then restores the *wrong* entry — command shape alone
+cannot see that. Instead, `worktree.sh` gained a clean-and-restore pair that
+removes the shared-mutable-state precondition entirely:
+
+| Verb | What it does |
+|------|--------------|
+| `worktree.sh stash-push <N> [--include-untracked]` | Captures the worktree's uncommitted tracked diff with `git stash create` (which builds a stash-format commit but **never writes `refs/stash`**), anchors it under the per-issue ref `refs/loom/stash-baseline/issue-<N>`, then resets that one worktree to a clean `HEAD` baseline. With `--include-untracked`, untracked files (Loom runtime markers excluded) move to a per-issue holding directory instead. |
+| `worktree.sh stash-pop <N>` | Re-applies exactly what `stash-push <N>` captured from that same per-issue ref / holding directory, then clears them. Refuses loudly — **without discarding the captured baseline** — if nothing is pending or if re-applying conflicts. |
+
+Each issue gets its **own** ref rather than a slot on a shared stack, so no
+other worktree's stash operation can interleave, and neither verb's command
+text contains a raw `git stash pop|drop|clear` — so both are
+guard-transparent, not a guard exemption. Raw `git stash pop`/`drop`/`clear`
+stays exactly as gated as before, in the main checkout and in a linked
+worktree alike.
+
 **Examples**:
 
 ```bash
@@ -743,13 +770,21 @@ git stash pop
 git stash drop stash@{1}
 git stash clear
 
-# In a linked worktree (.loom/worktrees/issue-N) — allowed, no ask:
+# In a linked worktree (.loom/worktrees/issue-N) — allowed only while it is
+# the ONLY managed worktree; asks once a second one exists (#4821):
 cd .loom/worktrees/issue-42 && git stash pop
 
 # Never gated, in either location — these cannot remove a stash entry:
 git stash push -m "wip"
 git stash apply
 git stash list
+
+# Headless clean-baseline-vs-my-diff comparison — never gated, because
+# neither verb touches refs/stash (#5217):
+./.loom/scripts/worktree.sh stash-push 42
+cargo clippy --message-format=short > /tmp/baseline.txt   # clean-tree baseline
+./.loom/scripts/worktree.sh stash-pop 42
+cargo clippy --message-format=short > /tmp/with-wip.txt   # then diff the two
 
 # Opt out for a whole repo:
 #   .loom/config.json  ->  { "guards": { "stashScope": false } }
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -3406,6 +3406,24 @@ fi
 # `<main>/.loom/worktrees/`, we ask too — a single active worktree has no one
 # else's stash entry to collide with, so it stays ungated.
 #
+# NOTE (#5217): this fires correctly, per the design above, for a legitimate
+# `git stash push && <baseline check> && git stash pop` pattern too — used to
+# diff a clean baseline against WIP (clippy/shellcheck/test-output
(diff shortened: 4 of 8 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:27:32Z) — standing down without reclaiming. Not stomping.

rjwalters:

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:57:51Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Clean, well-reasoned fix for a real headless-sweep stall (issue #5217). Reviewed the design rationale, the guard-transparency argument, and ran the full test suite in the builder worktree (`.loom/worktrees/issue-5217`):

## Correctness
- `stash-push`/`stash-pop` anchor WIP to a **per-issue** ref (`refs/loom/stash-baseline/issue-<N>`) via `git stash create` (never writes `refs/stash`), removing the shared-mutable-state precondition for the race that made a same-chain push/pop guard heuristic unsafe. The rejected-heuristic rationale is well documented in both the script comment and `guard-hooks.md`.
- The pending-marker mechanism correctly makes `stash-pop` a no-op when `stash-push` captured nothing (already-clean worktree), so the `push && check && pop` chain can't break on its own success path.
- Failure paths consistently preserve the captured baseline rather than discarding it (reset failure, apply-conflict failure, partial untracked-restore failure) — good defensive design for something that could otherwise eat a builder's WIP.
- `guard-destructive-generic.sh` changes are comment/message-only, confirmed via diff — behavior for raw `git stash pop/drop/clear` is unchanged, matching the PR's claim.

## Test Execution
1. `bash defaults/scripts/tests/test-worktree-stash-baseline.sh` — ✅ Executed: 29/29 pass, including Test 10 (concurrent worktree stash lands mid-sequence, restore unaffected) and Test 11 (contrast fixture proving raw stash still loses to the same interleaving — confirms the rejected heuristic was correctly NOT adopted).
2. `bash tests/hooks/test-guard-destructive.sh` — ✅ Executed: 687/687 pass (was 679; +8), including the new stash-scope(#5217) block asserting both the new allow-path and the unchanged raw-stash ask-path.
3. `bash scripts/check-docs-defaults-parity.sh` — ✅ Executed: OK
4. `bash defaults/scripts/check-phantom-labels.sh` — ✅ Executed: OK
5. `shellcheck --severity=warning` over `defaults/scripts/*.sh` — ✅ Executed: clean
6. `gh pr checks 5223` — ✅ all required checks pass (Rust/Dashboard jobs correctly skipped, no relevant changes)

## Other
- Merge state CLEAN, `Closes #5217` present and correct.
- Docs (`guard-hooks.md`, `builder.md`, `doctor.md`) accurately describe the new verbs and match the implementation.

No blocking issues found.

Issue #4993, opened 2026-08-03
operator: mint Developer ID Application cert and provision signing secrets for release CI


rjwalters:
## Task (operator-only — private key custody)

After the org enrollment lands (see the enrollment issue), mint the signing identity and provision it as Actions secrets on this repo so #4990 Phase 2's secrets-gated signing steps activate.

1. Mint a **Developer ID Application** certificate (Xcode or developer portal CSR flow) on one operator Mac.
2. Export as p12; store the p12 + passphrase in the operator's credential store (follow the `~/.cloudflare`-style convention: one purpose, dated note, never committed).
3. Provision Actions secrets (generic names, per #4990 — exact names finalized by the Phase 2 builder's workflow): signing cert p12 (base64), p12 passphrase, Team ID; optionally an App Store Connect API key for notarization.
4. Optional hardening: keep the private key only in CI secrets + one offline export; individual Macs never need it once artifact-based rolls ship (#4990 Phase 3).

Non-goal: no tracked file references the cert, team, or org (enforced by #4990's inverted AC).

Blocked by: Apple org enrollment. Blocks: signed-artifact leg of #4990 Phase 2.

rjwalters:
**Curator**: Related open work check (`check-duplicate.sh --issue 4993`) surfaced #4992 (Apple org enrollment). Disregarded — already reconciled in-body: this issue's own "Blocked by: Apple org enrollment" line and #4992's "Related" section (which lists #4993 as a sibling depending on its Team ID) already state the same relationship; no scope change needed.

Not marking `loom:curated` — every remaining step (minting a cert, custody of a p12 + passphrase, provisioning repo secrets) is a live operator action outside this repo's git worktree; there's no Builder-doable code change here. Issue is already well-scoped with clear steps and a stated blocker.

rjwalters:
## Moved to [2AMLogic/2am#27](https://github.com/2AMLogic/2am/issues/27)

This is **2AM Logic identity, not Loom mechanism.** Per [`FLEET.md` §Ownership](https://github.com/2AMLogic/2am/blob/main/FLEET.md#ownership):

> **`rjwalters/loom` ships the mechanism; this repo carries the identity.** The daemon, watchdog, drain, dispatch, role runner, dashboard engine, and `fleet add-worker` are generic and stay upstream. Which machines exist, what they are allowed to run, and what breaks when one dies are 2AM facts and live here.

[...]

PR #5225, opened 2026-08-04
feat(champion): detect dependency cycles with a bounded cross-repo walk

### LINKED ISSUE #5213: Champion: detect dependency cycles between epics and their blocked dependents

rjwalters:
## Summary

Champion's blocking evaluation was **single-hop and same-repo** in both places it
runs: `champion-pr-merge.md` → "Step 5: Unblock Dependent Issues" and
`champion-issue-promo.md` → Technical Feasibility criterion 2. Each reads a
`Blocked by` / `Depends on` / `Requires` reference and asks *"is #N closed yet?"* —
a question with no reachable answer when the declared edges form a **cycle**. Both
sides then re-derive "still blocked" forever, and nothing in either issue's text
makes the loop visible. The motivating incident spanned two repos and was found
only by an operator walking 15 child issues by hand.

This adds a dedicated, bounded, cross-repo cycle detector (curator's recommended
option B) and wires it into both call sites.

## Changes

- **New `defaults/scripts/detect-dependency-cycle.sh`** — depth-first walk over the
  repo's existing dependency-phrase vocabulary (#4508), widened on the *reference*
  side only so `owner/repo#N` and issue-URL edges are followable across repos.
  Exit codes mirror `check-duplicate.sh` (0 no cycle / 1 cycle / 2 error); stdout
  carries `CYCLE_PATH:`, `CYCLE_NODES:`, `CYCLE_FINGERPRINT:`, `SEARCH_TRUNCATED:`,
  `UNREADABLE:` markers so callers parse rather than re-derive.
- **`--report`** posts one comment naming every node in the cycle and adds
  `loom:operator-only`, keyed by `<!-- champion:dep-cycle:<fingerprint> -->`
  fingerprinted on the cycle's **node set** (so the same cycle found from either
  side collapses to one identity). Without `--report` the script is strictly
  read-only.
- **`champion-pr-merge.md`** — invokes the detector on exactly the
  `ALL_RESOLVED=false` branch, behind a cheap `loom:operator-only` label check, plus
  a "Dependency-cycle detection (#5213)" section with the bounded-cost table and an
  explicit statement that `loom:blocked` is *not* removed on a cycle.
- **`champion-issue-promo.md`** — a "Dependency-cycle gate" under criterion 2,
  triggered only when the proposal declares a dependency reference at all.
- **`defaults/scripts/tests/test-detect-dependency-cycle.sh`** (60 assertions) +
  `ci-wired.txt` entry.

| A cycle (E cites "Blocked by: I" and I — or one of I's own blockers — is a prerequisite for E) is detected before/during Champion's blocking evaluation instead of deadlocking silently | ✅ | Suite covers a direct 2-cycle, a 3-hop `A→B→C→A`, and a cycle *downstream* of the root (`1→2→3→2`) that reports the loop itself rather than the approach path. Wired into both evaluation sites; doc-pin assertions fail if either call site is removed |
| Detection works across repos | ✅ | Test reproduces the incident shape (`2AMLogic/marketing#56` ↔ `2AMLogic/klayout-tools#391`) via both an `owner/repo#N` and a `https://…/issues/N` reference. Also verified **live** against the real forge: `detect-dependency-cycle.sh --issue 5211 --repo rjwalters/loom` walked from this repo into `2AMLogic/klayout-tools` (`SCANNED: 3`, `NO_CYCLE`, ~3.4s) |
| On a cycle, Champion surfaces it explicitly (comment naming both sides) and routes to `loom:operator-only` rather than re-deriving "still blocked" forever | ✅ | `--report` test asserts the comment contains both `2AMLogic/marketing#56` and `2AMLogic/klayout-tools#391` plus the marker, and that `loom:operator-only` was applied; a second pass emits `ALREADY_REPORTED:` and the fixture ends with exactly **1** comment. Prose states `loom:blocked` stays |
| Bounded cost — no unbounded/expensive walk on every Champion pass | ✅ | Three hard caps (`--max-depth` 4, `--max-nodes` 25, `--max-steps` 500), one fetch per distinct node, memoization of only *completely* explored subtrees, `gh-cached` reads, and the depth bound enforced **before** the fetch so the widest frontier layer costs nothing (test asserts the out-of-bounds node is never fetched). Trigger-gated at both call sites: merge only on `ALL_RESOLVED=false` and only when `loom:operator-only` is absent; promotion only when a dependency reference exists. Any bound that fires prints `SEARCH_TRUNCATED:` so `NO_CYCLE` is never read as proof |

## Test Plan

- `defaults/scripts/tests/test-detect-dependency-cycle.sh` — **60/60 passed**,
  hermetic (a `gh` stub on `PATH`; no network, forge, or tokens). Covers: the
  sourced parser/helpers directly; 2-hop / 3-hop / off-root / cross-repo cycles;
  fingerprint stability from either member; a deep non-cyclic chain (no false
  positive); the depth bound and the node bound (including that a cycle closing
  *exactly at* the bound is still detected); `CLOSED` edges as resolved, not
  deadlocked; self-reference; an unreadable cross-repo node degrading to
  `UNREADABLE:`; `--report` idempotency; read-only default; argument validation;
  and doc pins on both Champion call sites.
- Regression suites that pin the same prose: `test-dependency-parse.sh` 22/22,
  `test-champion-critical-file-check.sh` 9/9,
  `test-forge-helpers-rate-limit-fallback.sh` 65/65.
- `check-ci-suite-manifest.sh` OK (104 suites accounted for),
  `check-docs-defaults-parity.sh` OK, `check-agents-md-sync.sh` OK.
- `shellcheck` clean at `--severity=style` on both new scripts (CI lints at
  `--severity=warning`).
- Live read-only smoke tests against `rjwalters/loom` (#4933, #4993, #4702, #5211):
  no false positives — notably #4993's `"Blocked by: Apple org enrollment. Blocks:
  … #4990"` line correctly yields **no** edge, since the phrase is not followed by a
  reference.

## Notes for review

- **Composition with #5211 / PR #5220 (open, not merged).** That PR adds an
  epic-aware blocker check to `champion-common.md` and inserts into the *same*
  region of `champion-issue-promo.md` criterion 2, so a textual conflict is
  expected for whichever merges second. This PR does not call or duplicate its
  helpers (they do not exist on `main`); instead both new sections state
  defensively that the epic check runs **first** and the cycle detector sees only
  what survives it — the two answer different questions ("has this blocker
  finished?" vs "can this blocker ever finish?") and share no state and no marker
  (`champion:epic-block:*` vs `champion:dep-cycle:*`).
- **False-positive posture** (documented in the script header): the edge parser
  reuses the repo's existing line-scoped two-stage parse rather than inventing a
  stricter second one, so it is deliberately over-inclusive. Under-parsing would
  miss a real deadlock; over-parsing at worst routes one issue to a human with the
  full path spelled out, leaving `loom:blocked` and every state machine untouched.

Closes #5213

### DIFF
--- a/defaults/.claude/commands/loom/champion-issue-promo.md
+++ b/defaults/.claude/commands/loom/champion-issue-promo.md
@@ -162,7 +162,65 @@ For each proposal issue (`loom:curated`, `loom:architect`, `loom:hermit`, or `lo
 ### 2. Technical Feasibility
 - [ ] Solution approach is technically sound
 - [ ] No obvious blockers or dependencies
+- [ ] Declared blockers are *resolvable* — not a dependency cycle (see "Dependency-cycle gate" below)
 - [ ] Fits within existing architecture
+
+#### Dependency-cycle gate (#5213)
+
+The "no obvious blockers or dependencies" check above is **single-hop and
+same-repo**: it looks at the `Blocked by`/`Depends on`/`Requires` references in
+this issue's body and asks whether each is closed. That is blind to a *cycle* —
+this issue waits on #B, and #B (or something #B waits on, possibly in another
+repo) waits back on this issue. Such an issue can never become promotable by
+waiting, and every future Champion pass re-derives the same conclusion.
+
+**Run the gate only when the proposal actually declares a blocker** — an issue
+whose body has no `(Blocked by|Depends on|Requires)` reference costs nothing:
+
+```bash
+ISSUE_NUMBER=<number>
+
+# Cheap trigger: does this proposal declare any dependency at all? Same
+# vocabulary as everywhere else (#4508), widened to cross-repo/URL refs.
+DECLARES_DEP=$("$GH_READ" issue view "$ISSUE_NUMBER" --json body --jq '.body' \
+  | grep -cE '(Blocked by|Depends on|Requires)[*_:[:space:]]*(([A-Za-z0-9._-]+/[A-Za-z0-9._-]+)?#[0-9]+|https?://[^[:space:]),]+/issues/[0-9]+)')
+
+if [ "$DECLARES_DEP" -gt 0 ]; then
+  CYCLE_RC=0
+  ./.loom/scripts/detect-dependency-cycle.sh --issue "$ISSUE_NUMBER" --report || CYCLE_RC=$?
+  if [ "$CYCLE_RC" -eq 1 ]; then
+    # Criterion 2 FAILS. The script has already posted one comment naming every
+    # node in the cycle and added loom:operator-only. Do NOT promote, do NOT
+    # post a separate NEEDS REVISION verdict — a cycle is not something the
+    # proposal's author can fix by revising this issue's text, and
+    # loom:operator-only already excludes it from every future pass.
+    echo "#$ISSUE_NUMBER is in a dependency cycle — routed to loom:operator-only, skipping promotion"
+  fi
+fi
+```
+
+The detector is bounded by construction (default 4 hops, 25 fetched issues, 500
+edges; cached reads; `SEARCH_TRUNCATED:` printed whenever a bound fires so
+`NO_CYCLE` is never read as proof) and its comment is idempotent on the cycle's
+node set, so a cycle that survives several passes is surfaced exactly once. Full
+rationale, marker vocabulary and the bounded-cost table live in
+`champion-pr-merge.md` → "Dependency-cycle detection (#5213)"; both call sites
+run the same script, so there is one walk implementation to keep correct.
+
+**It also runs at most once per issue, without needing its own skip.** Adding
+`loom:operator-only` removes the issue from every future promotion pass —
+`champion.md`'s candidate queries already exclude that label, and "When NOT to
+Promote" below restates it — so the pass that finds a cycle is the last pass that
+walks it. Nothing here needs to be added to the body-hash idempotency machinery
+in "Idempotency check": that marker answers "has the proposal been revised", a
+question a cycle is indifferent to.
+
+**If an "Epic-aware blocker sub-check" is present under this criterion** (it
+resolves blockers whose epic has in substance already shipped), run it **first**
+and let this gate see only the blockers that survive it. A blocker the epic check
+clears is *resolvable*, not a deadlock, and reporting it as a cycle would put a
+human in front of an issue that needed no human. The two are independent
+mechanisms with independent markers — neither reads the other's state.
 
 ### 3. Implementation Clarity
 - [ ] Enough detail for a Builder to start work
--- a/defaults/.claude/commands/loom/champion-pr-merge.md
+++ b/defaults/.claude/commands/loom/champion-pr-merge.md
@@ -899,6 +899,26 @@ for blocked in $BLOCKED_ISSUES; do
     fi
   done
 
+  # "Still blocked" is only a valid conclusion if waiting can ever end. Run the
+  # bounded, cross-repo cycle detector on exactly this branch (#5213) — see
+  # "Dependency-cycle detection" below for why it is gated here and nowhere else.
+  if [ "$ALL_RESOLVED" = false ]; then
+    # Second gate, before the walk: a cycle already surfaced on this issue is a
+    # human's to break, and re-walking it every pass buys nothing. One cached
+    # label read (backlog observation, not arbitration) replaces up to
+    # --max-nodes forge reads. Cached("$GH_READ") — see "Cached forge reads".
+    ALREADY_ROUTED=$("$GH_READ" issue view "$blocked" --json labels \
+      --jq '[.labels[].name] | index("loom:operator-only") // empty')
+
+    if [ -z "$ALREADY_ROUTED" ]; then
+      CYCLE_RC=0
+      ./.loom/scripts/detect-dependency-cycle.sh --issue "$blocked" --report || CYCLE_RC=$?
+      if [ "$CYCLE_RC" -eq 1 ]; then
+        echo "  Dependency CYCLE on #$blocked — surfaced and routed to loom:operator-only; not re-deriving 'still blocked'"
+      fi
+    fi
+  fi
+
   if [ "$ALL_RESOLVED" = true ]; then
     echo "  All dependencies resolved - unblocking #$blocked"
     gh issue edit "$blocked" --remove-label "loom:blocked" --add-label "loom:issue"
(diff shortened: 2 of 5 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:36:57Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Approved

Reviewed the full diff in `.loom/worktrees/issue-5213`, ran the new suite and shellcheck locally, and verified CI + merge state live.

### Verification

| Gate | Result |
|---|---|
| `test-detect-dependency-cycle.sh` (local) | **60/60 passed**, hermetic |
| `check-ci-suite-manifest.sh` | OK — 104 suites accounted for |
| `shellcheck --severity=warning` (both new scripts) | clean |
| CI (`gh pr checks`) | all non-skipped checks **pass**, incl. Shell Test Suites 5m10s, Shellcheck, Docs/Defaults Parity, AGENTS.md Sync, Installer Integration |
| `mergeStateStatus` | **CLEAN** |
| Issue linkage | `Closes #5213` present; #5213 is not epic/epic-phase, so the closing keyword is correct |
| File modes / install wiring | both scripts `100755`; `install-loom.sh` copies all of `defaults/scripts/` recursively, so `./.loom/scripts/detect-dependency-cycle.sh` resolves for consumers |

### Acceptance criteria (#5213)

All four are met, and the tests pin them rather than asserting them in prose:

1. **Cycle detected instead of silent deadlock** — sound DFS back-edge detection (`_in_set "$ref" "$path"`), covered for a direct 2-cycle, a 3-hop `A→B→C→A`, and a cycle *downstream* of the root (`1→2→3→2`) where `_cycle_segment` correctly trims the approach path and reports only the loop.
2. **Cross-repo** — the reference half of the #4508 vocabulary is widened to `owner/repo#N` and issue URLs while the *phrase* half is reused verbatim, so there is no second divergent parser to drift. The incident shape (`2AMLogic/marketing#56` ↔ `2AMLogic/klayout-tools#391`) is reproduced through both reference forms.
3. **Escalation, not re-derivation** — `--report` posts one comment naming every cycle member and adds `loom:operator-only`; `loom:blocked` is explicitly left in place (correct — the issue *is* blocked, it just now has a human owner). Idempotency is fingerprinted on the sorted **node set**, so the same cycle found from either member collapses to one identity; the test asserts exactly 1 comment across two passes.
4. **Bounded cost** — three hard caps, one fetch per distinct node, `gh-cached` reads, and two trigger gates (merge site: only on `ALL_RESOLVED=false` *and* only when `loom:operator-only` is absent; promotion site: only when a dependency reference exists). Any bound that fires prints `SEARCH_TRUNCATED:`, so `NO_CYCLE` is never readable as proof.

### Things done well

- **The depth bound is enforced before the fetch, after the back-edge test.** That is the right order and non-obvious: the frontier is the widest layer of the walk, so not fetching bodies you will never parse is the single largest saving, and detection power is unchanged because the back-edge check already ran. The test pins both halves — `SCANNED: 2` at `--max-depth 2`, the out-of-bounds node absent from `calls.log`, *and* an N-node cycle still detected at `--max-depth N`.
- **Memoization is conservative in the right direction.** `EXPLORED` is only populated when `INCOMPLETE_COUNT` is unchanged across the subtree, so a subtree cut short by the depth bound or an unreadable node is never recorded as clean. Black-marking a fully-explored node is sound for cycle detection (any cycle through it would lie entirely within its reachable set), and the truncation guard closes the one case where it would not be.
- **`index("loom:operator-only") // empty` avoids the jq 0-is-truthy trap** — a match at index 0 yields `0`, not `empty`, so the short-circuit is correct for the first label too. Easy thing to get wrong.
- `printf '%s\n' "$VAR" | jq` rather than `echo` (#5094), portable `_sha256` fallback, `set -uo pipefail` without `-e` so exit codes stay meaningful, `BASH_SOURCE`-guarded `main` so tests source the real helpers instead of mirroring them.
- **CLOSED edges terminate the walk.** Right call: a resolved edge cannot deadlock anyone, and following it would surface historical cycles that hold nobody up.

### Composition with #5220 (checked, not assumed)

The claim holds. #5220 does touch `champion-common.md`, `champion-issue-promo.md` and `champion-pr-merge.md`, so a textual conflict for whichever lands second is a fair prediction. More importantly this PR takes **no code dependency** on it — it calls none of #5220's helpers and shares no marker (`champion:epic-block:*` vs `champion:dep-cycle:*`), so it stands alone on `main` today. The defensive "if an Epic-Aware Blocker Check is present, run it first" framing in both new sections is the correct ordering ("has this blocker finished?" must precede "can this blocker ever finish?"), and it is stated conditionally so it does not become a dangling reference. #5220 is currently `loom:changes-requested` + `loom:ci-failure`, so merging this first is the right sequencing anyway.

### Non-blocking follow-ups

None of these gate approval; noting them so they are on record rather than filing queue noise.

1. **`_node_key` can collide.** `tr '/#' '__'` maps `a/b_c#1` and `a_b/c#1` to the same key `a_b_c_1`, and both are valid GitHub names. The consequence is a wrong cached body for one node (false positive or false negative). A separator outside the `[A-Za-z0-9._-]` charset, or hashing the node, removes it. Vanishingly unlikely, cheap to fix.
2. **`_report_cycle` writes have no REST rate-limit fallback.** It uses plain `gh issue comment` / `gh issue edit`. Under the GraphQL exhaustion this repo hits routinely, the comment fails, `warn` goes to stderr, no `REPORTED:` marker is emitted — but the merge-site caller only tests `CYCLE_RC -eq 1` and still logs "surfaced and routed to `loom:operator-only`". The state is safe (no marker and no label were written, so the next pass retries), but the log line is misleading. `forge_gh_comment_rl_safe` in `lib/forge-helpers.sh` is the existing wrapper.
3. **Bound flags are unvalidated.** Only `--issue` is regex-checked. `--max-nodes abc` evaluates to `0` inside `[[ ... -ge ... ]]`, so the walk fetches nothing and still exits `NO_CYCLE` (with `SEARCH_TRUNCATED: nodes`, which mitigates it). A numeric check matching the `--issue` one would close it.
4. **Cosmetic hardening.** `for ref in $(parse_dependency_refs ...)` and `for n in $path` are unquoted, so they glob-expand as well as word-split, and the URL branch of the parser (`[^[:space:]),]+`) can in principle emit a node containing glob metacharacters. I looked for an exploitable path into the unquoted `<<EOF` report heredoc and did not find one — a node only reaches `CYCLE_NODES_PATH` after a successful fetch, and a node name containing `$()` or `*` cannot name a real repo — so this is defence-in-depth, not a live issue.
5. **Downstream cycles route the root, not the members.** In the `1→2→3→2` shape, `#1` gets the comment and `loom:operator-only` while `#2`/`#3` do not until Champion evaluates them directly. That converges (both carry `loom:blocked` and will be walked on their own passes) and the comment does name the real loop, so it satisfies the AC — just worth knowing the label lands on the observer rather than the participants in that case.

### False-positive posture

The over-inclusive line-scoped parse is the right trade, and the header documents it honestly: under-parsing misses a real deadlock, over-parsing costs one comment and one label with no state machine advanced. I checked the one shape that could have made this fire at scale in this repo — `champion-epic.md`'s phase-child template uses `Part of Epic #N`, not a dependency phrase, so epic/child relationships produce **no** edges and cannot manufacture a parent-child "cycle". Combined with both trigger gates, the blast radius is small.

Excellent work — the bounded-walk reasoning, the memoization guard, and the depth-bound-before-fetch ordering are all more careful than this class of feature usually gets.

PR #5227, opened 2026-08-04
fix(guard): mask inert cat-heredoc bodies in text-carrying flag values

### LINKED ISSUE #5216: Guard: catastrophic rm-scope pattern false-positives on a backtick-quoted example inside a heredoc-wrapped gh pr comment body
## What happened

Observed live in `.loom/logs/guard-decisions.log` on 2026-07-29 (also self-reproduced by the Auditor during a 2026-08-03 guard-telemetry review — several of the Auditor's own read-only `gh issue list --search "..."` calls were themselves blocked by unrelated catastrophic patterns for the same underlying reason, described below).

Log entry (pattern field is the guard's own regex, not the denied command):

```
"ts":"2026-07-29T06:00:56Z","decision":"deny","pattern":"catastrophic:rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)","tier":"catastrophic"
```

The denied command was a legitimate Judge approval comment on PR #4357 — a `gh pr comment 4357 --body "$(cat <<'EOF2' ... EOF2)" && gh pr edit 4357 ...` chain. The comment body's Security section documented, inside a markdown backtick code span, an example shell-injection payload that the reviewed `validate_repo` function correctly rejects: a repo-argument value combining an owner/name pair with a semicolon-separated recursive-force-remove-of-root command, quoted purely as prose (nothing in the actual `gh` command executes it). The catastrophic rm-scope-outside-repo pattern matches that literal substring anywhere in the raw command text — including inside a heredoc-wrapped `--body` value — and hard-denies the whole `&&`-chained command (comment + label transition), stalling the Judge phase until the agent finds a workaround. In this case PR #4357 still merged 49 seconds later (the agent worked around it), but the mechanism generally offers no such guarantee.

## Root cause

`ALWAYS_BLOCK_PATTERNS`' catastrophic scan (including rm-scope-outside-repo) is intentionally unanchored/quote-blind by design — the guard's own header comment explains this is deliberate so a `bash -c`-wrapped destructive payload can't be smuggled past it via quoting. This is the *same* documented tradeoff #3679 hit for the six main/master force-push literals, which was fixed to exempt quoted/comment-body text for those specific six patterns without opening a bypass (command-position-aware matching rather than blanket unanchored substring).

The rm-scope-outside-repo pattern (and likely its catastrophic-scan siblings — the docker-prune, aws-s3-delete-family, sql-ddl, lifecycle-or-cloud-delete, and curl/wget-pipe-shell patterns) were never covered by #3679's fix, which was scoped narrowly to the force-push literals. Separately, #5000/#5085/#5120 fixed heredoc-body masking, but *only* for `extract_write_targets()` (the `>`/`tee`/`sed -i` write-target scanner) — that lineage does not touch the catastrophic `ALWAYS_BLOCK_PATTERNS` scan at all, so it does not cover this case either.

- [ ] Apply the same command-position-aware (not blanket-unanchored) matching technique #3679 used for the force-push patterns to rm-scope-outside-repo, so a backtick/heredoc-quoted example inside a `gh pr comment`/`gh issue comment`/`gh pr create --body`/`-m`/`--title`/`--notes` value does not trip it.
- [ ] Audit whether the same false-positive shape reproduces for the other broad-substring catastrophic patterns (docker-prune, the two aws-s3-delete patterns, sql-ddl, lifecycle-or-cloud-delete, curl/wget-pipe-shell) and record the result per pattern — fix, or explicitly scope out with rationale, mirroring #5120's documented-blind-spot approach where a fix isn't safe.
- [ ] A genuine command-substitution-smuggled destructive removal (not merely quoted documentation text) must still hard-deny — no regression to the #3679-adjacent smuggling protection.
- [ ] Regression test: a `gh pr comment --body` (both single-line and heredoc-wrapped) that quotes a recursive-force-remove-of-root example in backticks is allowed; a real unquoted/smuggled equivalent targeting outside the repo still denies.

## Impact

Medium — stalls Judge/Doctor/Curator/Auditor comments that need to document or quote a dangerous-command example, a fairly common pattern when reviewing guard-hook or security-relevant PRs (as in #4357, and repeatedly in this Auditor's own session while searching for prior art on other guard patterns). Workarounds exist (temp-file body, reword the comment) but are not guaranteed and add friction to exactly the reviews most likely to discuss dangerous commands.

---
Discovered during Auditor's standing guard-decision-telemetry review (#3898 policy) of `.loom/logs/guard-decisions.log`.

---

### Reproduction (confirmed live against current `main`)

Ran the guard directly with a command shaped exactly like the incident (a
`gh pr comment --body "$(cat <<'EOF2' ... EOF2)"` chain quoting a
recursive-force-remove-of-root example in backticks):

```bash
COMMAND=$(cat <<'CMD'
gh pr comment 4357 --body "$(cat <<'EOF2'
## Security
Example payload: `owner/name; rm -rf /`
EOF2
)" && gh pr edit 4357 --add-label "loom:pr" --remove-label "loom:review-requested"
CMD
)
jq -n --arg cmd "$COMMAND" '{tool_name:"Bash", tool_input:{command:$cmd}}' \
  | .loom/hooks/guard-destructive-generic.sh
# -> BLOCKED: Command matches dangerous pattern: rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)
```

Confirmed reproducible today, as filed.

### Root cause — correction to the filed root cause

The filed root cause states `#3679`'s fix was "scoped narrowly to the
force-push literals" and never touched rm-scope. **That is no longer
accurate as of current `main`** — worth correcting so the Builder doesn't
chase the wrong mechanism.

`strip_literal_text()` (`.loom/hooks/guard-destructive-generic.sh:1825`) redacts
quoted `--body`/`-m`/`--title`/`--notes`/`--comment` VALUES, and the resulting
`COMMAND_NO_LITERAL_TEXT` copy is already fed through the **entire**
`ALWAYS_BLOCK_PATTERNS` loop (line ~2047), not just the force-push entries —
`defaults/docs/guard-hooks.md:925` documents this as "the catastrophic scan," not
"the force-push patterns." So a **plain** single-line quoted example
(`--body "rm -rf /"`) is already allowed for every catastrophic pattern,
including rm-scope, today.

The actual gap is narrower and orthogonal to which *pattern* is hit:
`strip_literal_text()` deliberately **declines to redact any quoted span
containing `$(`** (or a backtick) — this is its documented anti-smuggling
safety floor, so `git commit -m "$(git push --force origin main)"` still hard-denies. A
heredoc-wrapped body (`--body "$(cat <<'EOF2' ... EOF2)"`, the idiom this
repo's own Judge/Curator/Builder/Auditor instructions use pervasively for
multi-line comments) **necessarily contains `$(`**, so it is never redacted —
regardless of which `ALWAYS_BLOCK_PATTERNS` entry the quoted prose happens to
collide with. Confirmed by reproducing the **original #3679 force-push
false-positive** with the same heredoc wrapping:

```bash
COMMAND=$(cat <<'CMD'
gh pr comment 999 --body "$(cat <<'EOF2'
Example: `git push --force origin main`
EOF2
)"
CMD
)
jq -n --arg cmd "$COMMAND" '{tool_name:"Bash", tool_input:{command:$cmd}}' \
  | .loom/hooks/guard-destructive-generic.sh
# -> BLOCKED: Command matches dangerous pattern: git push --force origin main
```

So this is **not** rm-scope-specific and does **not** need a per-pattern
audit of "did #3679 cover this pattern" — every entry in
`ALWAYS_BLOCK_PATTERNS` is equally exposed whenever the quoting is
`$(cat <<'DELIM' ... DELIM)` instead of a plain quoted string. AC #2 below is
revised accordingly.

### Prior art — a closer, more directly reusable precedent than #3679

`#5181` (closed, fixed by PR `#5192`, merged **one day before this issue was
filed**) hit the *same* shape — a heredoc body that merely quotes a denied
phrase — for a *different* catastrophic-tier check
(`gh-api-rawfield-body-literal-at`). It was fixed with `mask_heredoc_bodies()`
(`#5000`, `.loom/hooks/guard-destructive-generic.sh:1567`) — a heredoc-boundary-aware
masking pass, not `strip_literal_text()`'s quote-and-`$(` approach — scoped
narrowly to just that one check (see the comment at
`.loom/hooks/guard-destructive-generic.sh:2210`-2224 and the regression test at
`tests/hooks/test-guard-destructive.sh:549`-582, `assert_allow "#5181: ..."`).
`mask_heredoc_bodies()` is a better fit here than extending
`strip_literal_text()`, because it does not key off `$(` at all — it masks
the body of any *closed* heredoc block regardless of what wraps it.

**Safety caveat the Builder must design around, not skip**: `mask_heredoc_bodies()`
currently has a documented, accepted "Known Limitation" (`#5117`, comment at
`.loom/hooks/guard-destructive-generic.sh:1499`-1518): it masks heredoc bodies
**regardless of what consumes them**, including a body that is genuinely fed
to an inner interpreter (`bash <<'EOF' ... EOF`, `cat <<'EOF' ... EOF | sh`).
That tradeoff is accepted for `extract_write_targets()` (a fail-open there
only means a missed *write-target*, not a missed catastrophic block) and was
judged acceptable for #5181's single, narrow literal-phrase check. It is a
**materially different risk** if blanket-applied to the general-purpose
`ALWAYS_BLOCK_PATTERNS` hard-deny floor, whose entire purpose (per the file's
own header comment) is resisting exactly this kind of quoting/heredoc
smuggling — a `bash <<'EOF'` heredoc feeding `rm -rf /` to an inner shell
must still hard-deny; masking it away to fix this issue would reopen the
`#3679`-adjacent smuggling hole AC #3 (below) already requires stay closed.

- [ ] Fix the false positive for a heredoc-wrapped `--body`/`-m`/`--title`/`--notes`/`--comment`
      value (`"$(cat <<'DELIM' ... DELIM)"`) that merely quotes a dangerous-command
      example, for the shared `ALWAYS_BLOCK_PATTERNS` catastrophic scan — a single
      shared-mechanism fix (following the `mask_heredoc_bodies()` / `#5181` precedent, not
      `strip_literal_text()`, which cannot help here by design) rather than a per-pattern
      port of `#3679`.
- [ ] The masking must be scoped narrowly enough that a **genuinely executed** dangerous
      command inside a heredoc still denies — in particular, a heredoc body that is fed to
      an interpreter (`bash <<'EOF' ... rm -rf / ... EOF`, `... | sh`) rather than merely
      quoted inside a `gh`/text-processing flag value. If a fully general fix can't safely
      distinguish "heredoc consumed as inert prose" from "heredoc consumed as a live
      script," scope the fix to the narrower, provably-safe case (e.g. only within a
      recognized `--body`/`-m`/`--title`/`--notes`/`--comment` flag value, mirroring
      `#5181`'s scoping) and document the residual gap explicitly (mirroring `#5117`'s
      "Known Limitations" convention) rather than silently widening what the catastrophic
      tier will allow.
- [ ] Confirm via test (not per-pattern rationale, since the mechanism is shared) that the
      same heredoc-wrapped false positive is also fixed for the sibling broad-substring
      catastrophic patterns named in the original report (docker-prune, the two
      aws-s3-delete patterns, sql-ddl, curl/wget-pipe-shell) and for the original `#3679`
      force-push literals (regression risk: a fix could inadvertently narrow force-push
      coverage too).
- [ ] Non-regression (unchanged from the filed AC): a genuine command-substitution-smuggled
      destructive removal — `git commit -m "$(rm -rf /)"`, `bash -c 'rm -rf /'` — must still
      hard-deny.
- [ ] Regression tests added to `tests/hooks/test-guard-destructive.sh` (mirror the
      `#5181` block at line ~549):
  - `assert_allow`: heredoc-wrapped `gh pr comment --body "$(cat <<'EOF' ... EOF)"` quoting
    a recursive-force-remove-of-root example in backticks (this issue's exact repro).
  - `assert_allow`: same shape quoting a force-push-to-main example (confirms the shared
    mechanism, not just rm-scope).
  - `assert_deny`: a live (non-heredoc) rm-scope-outside-repo command — regression guard.
  - `assert_deny`: `bash -c 'rm -rf /'` / `git commit -m "$(rm -rf /)"` — command-substitution
    smuggling still denied (regression guard for `#3679`'s own safety floor).
  - A case for the interpreter-fed-heredoc scoping decision above (either `assert_deny` if
    the fix closes it, or an explicit "known limitation, mirrors `#5117`" comment next to an
    `assert_allow` if it deliberately does not — do not leave this case silently untested).

### Affected Files

- `.loom/hooks/guard-destructive-generic.sh` — this repo's vendored copy of the guard;
  `ALWAYS_BLOCK_PATTERNS` loop (~line 1936-2050), `strip_literal_text()` (~line 1825),
  `mask_heredoc_bodies()` (~line 1567) are the relevant sections.
- `defaults/hooks/guard-destructive-generic.sh` — canonical template source for the same
  file (kept byte-identical to the vendored copy above; verified identical as of this pass).
- `tests/hooks/test-guard-destructive.sh` — add the regression tests above (existing
  `#3679` block starts ~line 2014; existing `#5181` block starts ~line 549 as the shape to
  mirror).
- `defaults/docs/guard-hooks.md` — the "First refinement pass (#3898)" bullet
  (~line 925) documents the catastrophic scan's redaction behavior; update if this fix
  changes what it allows/denies for heredoc-wrapped values.

### Test Plan

- [ ] Run `./tests/hooks/test-guard-destructive.sh` (full suite) — all existing assertions
      (including the `#3679` and `#5181` blocks) must still pass.
- [ ] New assertions above pass.
- [ ] Manually re-run this issue's own repro command (above) through the guard and confirm
      `ALLOW`.

### Related (cross-referenced, not duplicates — `check-duplicate.sh --issue 5216`)

- `#5181` (closed via PR `#5192`) — same false-positive *shape*, different specific
  catastrophic check; primary precedent for the fix approach above.
- `#5172`, `#5157` — open, different specific guard checks (merge-redirect / write-confinement)
  hit by the same general "heredoc body scanned as if it were live" family of false
  positives. Not in scope for this issue, but worth a glance if the fix here generalizes
  a reusable heredoc-scoping helper other checks could adopt later.

Classified `complex`: the fix sits directly on the catastrophic deny tier's anti-smuggling
safety floor — a mistake (over-broad heredoc masking) would not surface as a test failure
in the common case (only a crafted interpreter-fed-heredoc payload would expose it), and
the security cost of getting it wrong is a bypassed hard-deny, not a caught bug.

rjwalters:
## Summary

`strip_literal_text()` refuses to redact any quoted flag value containing `$(` — its documented anti-smuggling floor, which is what keeps a command-substitution payload from being laundered past the catastrophic tier. But this repo's own prescribed idiom for a multi-line comment body (`--body "$(cat <<'EOF' … EOF)"`, in CLAUDE.md, judge.md, doctor.md, builder-pr.md) *always* contains `$(`, so such a value was **never** redacted. A dangerous command merely **quoted inside the heredoc body as documentation** therefore hard-denied the entire command — observed live on a Judge approval for PR #4357, and reproducible for the #3679 force-push literals too. The gap is **construction-specific** (heredoc-wrapped value), not pattern-specific.

This PR adds `mask_flag_cat_heredocs()`, a heredoc-boundary-aware pre-pass inside `strip_literal_text()` that blanks the **body** of exactly one provably-inert shape, and routes the remaining broad scans that per-physical-line parsing let the same prose trip through the redacted copy.

### The scoping decision (interpreter-fed heredocs)

The issue explicitly warned against blanket-reusing `mask_heredoc_bodies()` (#5000) here: that helper masks *any* closed heredoc body regardless of what consumes it — an accepted fail-open for the write-target scanner (#5117 Known Limitation 1), but not acceptable on the catastrophic hard-deny floor, where a `bash <<'EOF' … EOF` body is live code to an inner shell.

The new pass masks only when **all four** conditions hold, and each one is load-bearing:

| Condition | What it rejects (still denies) |
|-----------|--------------------------------|
| 1. Opener is the complete tail of a recognized text-carrying flag's quoted value, immediately after `$(cat` | **Interpreter-fed heredocs**: `--body "$(bash <<Q … Q)"`, `cat <<Q … Q \| sh`, `sh -s <<Q … Q` — none match `<flag> <quote>$(cat` |
| 2. Heredoc delimiter is **quoted** (`<<-` allowed) | An unquoted delimiter lets the outer shell expand the body before `cat` sees it, so it is not provably inert |
| 3. Block is **closed** in the same buffer | An unterminated/false opener masks nothing (mirrors #5087's "never mask speculatively") |
| 4. The line after the delimiter line is `)` + that same opening quote | A command **chained after the heredoc inside the substitution** (`… Q` ⏎ `rm -rf /` ⏎ `)"`) really runs — bash ends the heredoc at the delimiter line |

So condition 1 **closes** the interpreter-fed case rather than leaving it as a known limitation; there are three `assert_deny` scoping tests for it (`--body "$(bash <<Q …)"`, `cat <<Q … | sh`, `sh -s <<Q …`).

**Known limitation** (recorded in the hook header and `guard-hooks.md`, mirroring the #5117 convention): only the literal `cat`-consumed spelling is recognized. An equivalent variant (`$(command cat <<Q …)`, an opener on a continuation line, `) "` with a space) is not recognized and keeps false-positiving exactly as it does today — a pre-existing false positive, never a new bypass.

### Why the fix is not the ALWAYS_BLOCK loop alone

Fixing only `ALWAYS_BLOCK_PATTERNS` leaves the issue's own repro **still denied**: the `rm`-scope check independently segment-parses the raw command per physical line, reads the body line `Example payload: ` + a backticked `owner/name; <recursive force removal of />` as a live `rm` segment, and denies with `rm on protected system path`. Same failure family as #5000's phantom write targets. Four sibling scans had the same per-line blindness, so they now read the literal-redacted copy too. Every one of them is a *narrowing* change: a real invocation outside a quoted flag value is byte-for-byte unaffected.

## Changes

- `defaults/hooks/guard-destructive-generic.sh` (+ the byte-identical vendored `.loom/hooks/` copy):
  - New `mask_flag_cat_heredocs()` awk helper, run as a pre-pass in `strip_literal_text()`'s `END` block. It runs before the existing quoted-span redaction because `re`'s span classes (`[^"]*` / `[^']*`) stop at the first quote character, and heredoc prose contains raw quotes constantly — the span match alone cannot see such a value whole. Body bytes are replaced 1:1 with `X`, preserving byte offsets and line count.
  - SQL-DDL deny now scans `COMMAND_ASK_SCAN` (comment-stripped **and** literal-redacted). This was the one broad-substring deny that never received #3679's redaction at all — plain `--body "example payload: <DDL>"` denied where the equivalent prose about a recursive removal was already allowed. `-c`/`-e` are not text-carrying flags, so `psql -c '…'` / `mysql -e '…'` still deny.
  - `rm`-scope deny (`extract_rm_targets` + its pre-check) now scans `COMMAND_NO_LITERAL_TEXT`.
  - Lifecycle deny (`lifecycle_or_cloud_reason`) now scans `COMMAND_ASK_SCAN` — a body line starting with `halt`/`shutdown` was read as a live command word.
  - Force-op ask (`parse_force_ops` + its pre-check) and the cloud-CLI ask loop now scan `COMMAND_ASK_SCAN`, matching `ASK_PATTERNS` / `REVERSIBLE_GH_PATTERNS`, which already did. An unanswered ask blocks a headless sweep just like a deny, so leaving these would have half-fixed the reported stall.
- `tests/hooks/test-guard-destructive.sh`: new `#5216` block, 29 assertions.
- `defaults/docs/guard-hooks.md`: "Third refinement pass (#5216)" with the four-condition table and the known limitation; amended the #3898 bullet to point at it.

| 1. Fix the false positive for a heredoc-wrapped `--body`/`-m`/`--title`/`--notes`/`--comment` value quoting a dangerous example, for the shared `ALWAYS_BLOCK_PATTERNS` catastrophic scan | ✅ | Single shared-mechanism fix in `strip_literal_text()`, which already feeds the whole catastrophic loop. Issue's exact repro command re-run through the guard by hand → **ALLOW**; `assert_allow` "#5216: heredoc --body quoting an 'rm -rf /' payload, chained with gh pr edit (the PR #4357 repro) allowed" |
| 2. Masking scoped narrowly enough that a genuinely-executed dangerous command inside an interpreter-fed heredoc still denies | ✅ | Condition 1 (see table above) closes it rather than documenting it away. 3 scoping `assert_deny`s: `--body "$(bash <<Q …)"`, `cat <<Q … \| sh`, `sh -s <<Q …`. Plus `assert_deny` for a command chained after the heredoc inside `$( … )`, and for an unquoted delimiter |
| 3. Same fix covers the sibling broad-substring catastrophic patterns (docker-prune, both aws-s3-delete patterns, sql-ddl, curl/wget-pipe-shell) and the #3679 force-push literals — no narrowing | ✅ | `assert_allow` for each of docker-prune, `aws s3 rm --recursive`, `aws s3 rb`, curl-pipe-shell, SQL DDL (heredoc **and** plain single-line), and force-push-to-main (mid-line **and** line-start). Non-narrowing: all 40 pre-existing #3679 assertions still pass, plus new `assert_deny`/`assert_ask` for real `docker system prune`, real DDL, real bucket removal, real `docker rm`, real force-push to a feature branch |
| 4. Non-regression: command-substitution-smuggled destructive removal still hard-denies | ✅ | `assert_deny` for `bash -c 'rm -rf /'` and `git commit -m "$(rm -rf /)"`; plus live `rm -rf /`, live `rm -rf /usr`, `git commit -m "…" && rm -rf /usr`, and a real removal chained after a heredoc-wrapped body ("narrows, never widens") |
| 5. Regression tests added to `tests/hooks/test-guard-destructive.sh` mirroring the #5181 block | ✅ | New `--- #5216 ---` section, 29 assertions, all listed above. The interpreter-fed scoping case is an explicit `assert_deny` (the fix closes it), not a commented `assert_allow` |

## Test Plan

- `./tests/hooks/test-guard-destructive.sh` → **708 passed, 0 failed** (679 pre-existing + 29 new), including the #3679 and #5181 blocks and the <200ms performance assertion (14ms avg).
- `tests/hooks/test-guard-destructive-dispatcher.sh` → 12/12; `defaults/hooks/tests/test-guard-worktree-paths.sh` → 28/28 (includes the `.loom/` vs `defaults/` byte-identity assertion); `tests/install/test-hooks-preserve.sh` → 10/10.
- `shellcheck --severity=error` clean on the modified hook; `bash -n` clean on the hook and the test suite.
- `scripts/check-docs-defaults-parity.sh` → OK.
- Manually re-ran the issue's own repro command through the guard: **ALLOW** (was `deny` / `catastrophic:rm…` on `main`).
- Manually diffed guard decisions for 17 hand-built probe commands against `main`'s copy: every deny that should persist persists; only the inert-prose cases flipped to allow.

Closes #5216

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -1822,8 +1822,114 @@ resolve_stash_cwd() {
 # it does not model backslash-escaped quotes, but since the result feeds only
 # the narrowing (never widening) catastrophic scan, the worst case is a raw
 # substring surviving — never a catastrophic block being skipped incorrectly.
+#
+# -----------------------------------------------------------------------------
+# HEREDOC-WRAPPED FLAG VALUES (#5216)
+#
+# The `$(`-floor above is exactly right for a general command substitution, but
+# it also declines to redact THIS repo's own pervasive idiom for any multi-line
+# comment body (CLAUDE.md / judge.md / doctor.md / builder-pr.md all show it):
+#
+#     gh pr comment 4357 --body "$(cat <<QUOTED_DELIM
+#     …prose that may QUOTE a dangerous command as an example…
+#     QUOTED_DELIM
+#     )" && gh pr edit 4357 --add-label "loom:pr"
+#
+# Every value built that way necessarily contains `$(`, so before this pass it
+# was NEVER redacted — and a dangerous-command example merely quoted inside the
+# body (a Judge documenting the shell-injection payload a PR now rejects, an
+# Auditor quoting a guard pattern) hard-denied the whole command on the
+# catastrophic tier. Observed live in guard-decisions.log on 2026-07-29 (a Judge
+# approval on PR #4357) and reproduced for the #3679 force-push literals too:
+# the gap is CONSTRUCTION-specific (heredoc-wrapped value), not pattern-specific.
+#
+# mask_flag_cat_heredocs() (below) closes it by masking ONLY the BODY of a
+# heredoc in this one provably-inert shape, and only when ALL of these hold:
+#   1. the opener is the complete tail of its line, immediately preceded by a
+#      recognized text-carrying flag, its opening quote, and `$(cat`;
+#   2. the heredoc delimiter is QUOTED (single- or double-quoted, `<<-` allowed)
+#      — a quoted delimiter is what guarantees the outer shell performs NO
+#      expansion on the body, so a `$(…)` sitting IN the body is inert text
+#      rather than live code (an UNQUOTED delimiter is rejected outright);
+#   3. the block is CLOSED in this same buffer (mirrors #5087's "never mask
+#      speculatively" rule for mask_heredoc_bodies);
+#   4. the very next line after the delimiter line is `)` + that same opening
+#      quote — i.e. the substitution ends immediately, with nothing chained
+#      after the heredoc inside it.
+# Condition 4 is what keeps a `--body "$(cat <<QUOTED_DELIM … QUOTED_DELIM`
+# <newline> `rm -rf /` <newline> `)"` command denying: bash ends the heredoc at
+# the delimiter line and then genuinely RUNS the following line inside the
+# substitution, so nothing is masked there. Condition 1 is what keeps an
+# INTERPRETER-FED heredoc denying — a body consumed by `bash <<DELIM`,
+# `sh -s <<DELIM`, or `cat <<DELIM … | sh` is live code
+# to the inner shell, and none of those match `<flag> <quote>$(cat`. This is the
+# deliberate narrowing versus reusing mask_heredoc_bodies() (#5000) here: that
+# helper masks ANY closed heredoc body regardless of its consumer, a documented
+# and accepted fail-open for the write-target scanner (#5117 Known Limitation 1)
+# that must NOT be inherited by the catastrophic hard-deny floor.
+#
+# KNOWN LIMITATION (recorded, mirroring the #5117 convention): this recognizes
+# only the literal `cat`-consumed shape spelled out above. A semantically
+# equivalent variant — `$(command cat <<DELIM …)`, a heredoc opened on a
+# continuation line, or a body whose delimiter line is followed by `) "` with a
+# space — is simply not recognized and keeps denying exactly as it does today.
+# That is the safe direction (a false positive that already exists, never a new
+# bypass), and the shape above is the one the repo's own role prompts prescribe.
+# -----------------------------------------------------------------------------
 strip_literal_text() {
     printf '%s' "$1" | awk '
+    # Mask the body of a `<flag> "$(cat <<QUOTED_DELIM … DELIM\n)"` heredoc.
+    # See the header comment above for the four conditions and why each is
+    # load-bearing. Body bytes are replaced 1:1 with "X" so the buffer keeps
+    # its byte offsets and line count; the opener line, the delimiter line and
+    # everything outside the body are left untouched.
+    function mask_flag_cat_heredocs(s,   lines, nl, i, j, line, pre, oq, delim, dq, closeat, trimmed, body, dashform) {
+        if (index(s, "<<") == 0) return s
+        nl = split(s, lines, "\n")
+        for (i = 1; i <= nl; i++) {
+            line = lines[i]
+            # (2) opener must END the line and carry a QUOTED delimiter.
+            if (match(line, /<<-?["'"'"'][A-Za-z0-9_]+["'"'"'][ \t]*$/) == 0) continue
+            dashform = (substr(line, RSTART + 2, 1) == "-")
+            delim = substr(line, RSTART, RLENGTH)
+            sub(/^<<-?/, "", delim)
+            sub(/[ \t]*$/, "", delim)
+            dq = substr(delim, 1, 1)
+            if (substr(delim, length(delim), 1) != dq) continue   # quotes must match
+            delim = substr(delim, 2, length(delim) - 2)
+            if (delim == "") continue
+            # (1) …immediately preceded by <flag> <openquote>$(cat.
+            pre = substr(line, 1, RSTART - 1)
+            if (pre !~ /(^|[ \t])(--message|--body|--notes|--title|--comment|-m)[ \t]*=?[ \t]*["'"'"']\$\([ \t]*cat[ \t]+$/) continue
+            oq = ""
+            for (j = length(pre); j >= 1; j--) {
+                if (substr(pre, j, 2) == "$(") { oq = substr(pre, j - 1, 1); break }
+            }
+            if (oq != DQ && oq != SQ) continue
+            # (3) the block must be CLOSED inside this buffer.
+            closeat = 0
+            for (j = i + 1; j <= nl; j++) {
+                trimmed = lines[j]
+                if (dashform) sub(/^\t+/, "", trimmed)
+                if (trimmed == delim) { closeat = j; break }
+            }
+            if (closeat == 0) continue
+            # (4) the substitution must close IMMEDIATELY after the delimiter
+            #     line — `)` + the same opening quote — so nothing chained
+            #     after the heredoc inside `$( … )` is masked away.
+            if (closeat == nl) continue
+            if (substr(lines[closeat + 1], 1, 2) != ")" oq) continue
+            for (j = i + 1; j < closeat; j++) {
+                body = lines[j]
+                gsub(/./, "X", body)
+                lines[j] = body
+            }
+            i = closeat
+        }
+        s = lines[1]
+        for (i = 2; i <= nl; i++) s = s "\n" lines[i]
+        return s
+    }
     BEGIN {
         SQ = sprintf("%c", 39)   # single quote
         DQ = sprintf("%c", 34)   # double quote
@@ -1847,7 +1953,16 @@ strip_literal_text() {
     # byte-for-byte identical to the previous behaviour.
     { buf = buf (NR > 1 ? "\n" : "") $0 }
     END {
-        s = buf
+        # PRE-PASS (#5216): blank the body of a `<flag> "$(cat <<QDELIMQ … )"`
+        # heredoc before the quoted-span redaction below runs. It has to happen
+        # here rather than inside the loop because `re`'"'"'s quoted-span classes
+        # ([^"]* / [^'"'"']*) stop at the first quote character, and a heredoc body
+        # is free to contain raw quotes (prose routinely does) — so the span
+        # match alone cannot see such a value whole. Masking first also means
+        # the `$(`-floor below needs no exception: by the time the loop reads
+        # this span, the only text left inside it is `$(cat <<QDELIMQ`, the
+        # delimiter, and `)`.
+        s = mask_flag_cat_heredocs(buf)
         out = ""
         while (match(s, re)) {
             pre     = substr(s, 1, RSTART - 1)
@@ -2361,7 +2476,16 @@ lifecycle_or_cloud_reason() {
 # unanswered ASK still blocks (see defaults/docs/guard-hooks.md). A lifecycle
 # deny takes precedence over a cloud-delete ask in a compound command (e.g.
 # `az group delete …; halt`), so the hard floor is never downgraded.
-_LIFECYCLE_CLOUD_REASONS=$(lifecycle_or_cloud_reason "$COMMAND_NO_COMMENT")
+#
+# SCANS COMMAND_ASK_SCAN, NOT COMMAND_NO_COMMENT (#5216). lifecycle_or_cloud_reason()
+# segment-parses per physical line, so a heredoc BODY line inside
+# `--body "$(cat <<'EOF' … EOF)"` whose first word happens to be a lifecycle verb
+# ("halt the deploy if this fails", "shutdown checklist:") was read as a live
+# `halt`/`shutdown` command word and hard-denied a comment that runs nothing.
+# The literal-redacted copy blanks only that quoted-value text; a real
+# `… ; halt` outside a flag value, or a `bash -c 'halt'` payload (never
+# redacted), still reaches this check and still denies.
+_LIFECYCLE_CLOUD_REASONS=$(lifecycle_or_cloud_reason "$COMMAND_ASK_SCAN")
 _LIFECYCLE_DENY=$(printf '%s\n' "$_LIFECYCLE_CLOUD_REASONS" | grep '^system lifecycle command:' | head -1)
 if [[ -n "$_LIFECYCLE_DENY" ]]; then
     deny "BLOCKED: $_LIFECYCLE_DENY" "lifecycle"
@@ -2379,10 +2503,23 @@ fi
 # all four DDL statements in one pass (cheaper than a per-pattern loop), and
 # sql_guard_enabled() is consulted only after a match, so the config read stays
 # off the hot path.
+#
+# SCANS COMMAND_ASK_SCAN, NOT COMMAND_NO_COMMENT (#5216). This check was the one
+# broad-substring deny that never received #3679's quoted-flag-value redaction:
+# `gh pr comment --body "example payload: DROP TABLE users"` hard-denied a
+# comment that runs no SQL at all, where the equivalent prose about `rm -rf /`
+# was already allowed by the catastrophic scan. COMMAND_ASK_SCAN is the
+# comment-stripped AND literal-redacted copy (built above; already relied on by
+# the deny-tier write-confinement check, so its use here is not an ask-tier-only
+# convention), which also carries #5216's heredoc-body masking — so both the
+# plain `--body "…"` and the heredoc-wrapped `--body "$(cat <<'EOF' … EOF)"`
+# forms of quoted prose stop false-positiving in one step. `-c`/`-e` are NOT
+# text-carrying flags, so the real invocations (`psql -c '…'`, `mysql -e '…'`)
+# are untouched and still deny.
 # =============================================================================
 SQL_DDL_PATTERN='DROP DATABASE|DROP TABLE|DROP SCHEMA|TRUNCATE TABLE'
-if echo "$COMMAND_NO_COMMENT" | grep -qiE "$SQL_DDL_PATTERN" && sql_guard_enabled; then
-    matched=$(echo "$COMMAND_NO_COMMENT" | grep -oiE "$SQL_DDL_PATTERN" | head -1)
+if echo "$COMMAND_ASK_SCAN" | grep -qiE "$SQL_DDL_PATTERN" && sql_guard_enabled; then
+    matched=$(echo "$COMMAND_ASK_SCAN" | grep -oiE "$SQL_DDL_PATTERN" | head -1)
     deny "BLOCKED: Command matches dangerous pattern: ${matched:-SQL DDL statement}" "sql-ddl"
 fi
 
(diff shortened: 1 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T04:55:33Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reclaiming stale loom:reviewing claim (age > 30m, no follow-up comment other than a marked stand-down) — a prior Judge's parent sweep likely died mid-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T05:26:29Z, reclaimed by a prior pass only ~3 minutes before this check) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Note on the stale claim**: this PR carried both `loom:review-requested` and `loom:reviewing` (claimed 2026-08-04T04:55:33Z) with no follow-up beyond a marked stand-down comment from a concurrent Judge pass. Re-verified the claim was still `loom:reviewing` immediately before acting, confirmed it was genuinely stale (age > 30m, `COMMENTS_AFTER == 0` once the marked stand-down is excluded), reclaimed it, and proceeded with a normal review. GraphQL was rate-limited for parts of the pass; fell back to the documented REST endpoints (`gh api .../labels`, `.../comments`) per judge.md.

**Review**

- `mask_flag_cat_heredocs()` in `strip_literal_text()`'s `END` block masks the body of exactly the `<flag> "$(cat <<'QDELIM' … QDELIM)"` shape, gated on all four documented conditions (quoted-and-terminated opener, quoted delimiter, closed in-buffer, substitution closes immediately after the delimiter line). Read through the awk carefully — the backward scan for the opening-quote character, the dash-form (`<<-`) handling, and the closeat/condition-4 check are all correct and consistent with the surrounding comments.
- Five sibling scans (SQL-DDL, rm-scope, lifecycle, force-op ask, cloud-CLI ask) were switched from `COMMAND_NO_COMMENT`/raw `COMMAND` to the already-literal-redacted `COMMAND_ASK_SCAN` / `COMMAND_NO_LITERAL_TEXT` copies — each is a narrowing-only change (redacts a quoted flag value, never touches a real invocation), matching the PR's own claim.
- `.loom/hooks/guard-destructive-generic.sh` is still byte-identical to `defaults/hooks/guard-destructive-generic.sh` (verified with `diff`).

**Verification performed** (in the existing builder worktree, `.loom/worktrees/issue-5216`, head `51fd08bd` matching the PR's head SHA):
- `./tests/hooks/test-guard-destructive.sh` → 708/708 passed, ~14ms avg (matches the PR's claim).
- `shellcheck --severity=error` clean; `bash -n` clean on both the hook and the test file.
- `./scripts/check-docs-defaults-parity.sh` → OK.
- Manually re-ran the issue's exact repro (a `gh pr comment 4357 --body "$(cat <<'EOF2' ... EOF2)" && gh pr edit 4357 ...` chain quoting a recursive-force-removal-of-filesystem-root payload) through the guard directly: **denies on `main`, allows on this branch** — confirms the fix does what it claims, not just what the test suite asserts.
- Spot-checked two edge cases not in the added test block: `--body=` (equals form, no space before quote) still masks correctly (allow); a heredoc whose body contains a line that merely *resembles* the delimiter (`EOF2X`) followed by the real delimiter and then a genuinely chained command after the close still denies (condition 4 correctly declines to mask when something follows the delimiter line other than `)"`).
- CI: all applicable checks pass (`gh pr checks`); path-filtered jobs correctly show `skipping`. Merge state is `CLEAN`, `mergeable: true`.

Solid, well-scoped fix with thorough test coverage and honest documentation of its one known limitation. Approving.

rjwalters:

merging it automatically:

- **Blast radius**: touches `.loom/hooks/guard-destructive-generic.sh` and `defaults/hooks/guard-destructive-generic.sh` — the destructive-command guard hook that hard-denies catastrophic operations (a recursive force removal of the filesystem root, force-push-to-main, DDL, etc.) fleet-wide. This is one of the surfaces the merge-risk criteria explicitly enumerate as high-blast-radius regardless of diff quality or review depth: a regression in the new `mask_flag_cat_heredocs()` boundary-parsing logic (the four gating conditions on when a heredoc body is "provably inert") could either open a bypass in the catastrophic-deny floor for every agent in the fleet, or newly false-positive-deny legitimate commands fleet-wide. The Judge review here is unusually thorough (live repro re-run, manual edge-case spot-checks beyond the added test block, 708/708 passing), but that only bears on the diff-composition and review-depth axes — it doesn't change what a live regression on this specific file would cost if the added coverage has a gap.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5227`

---
*Automated by Champion role*

Issue #1111, opened 2026-01-24
Add automated Weekly Intelligence Reports


rjwalters:
## Problem

Users have no automated way to receive insights about their development patterns. They must manually check metrics and analyze trends. The system can detect patterns but doesn't proactively surface them.

## Solution

Implement automated Weekly Intelligence Reports that:
- Generate automatically (e.g., every Monday morning)
- Summarize key metrics vs previous week
- Highlight best patterns and anti-patterns detected
- Surface anomalies and recommendations
- Provide "Did you know?" insights

### Report Generation
- Scheduled report generation (configurable day/time)
- Aggregate metrics for the past week
- Compare to previous week for trends
- Pattern detection from correlation analysis
- Anomaly detection (stuck PRs, unusual cost spikes, etc.)

### Report Content
- Summary stats: features, PRs, cost, tokens, success rate
- Week-over-week deltas with trend indicators
- Top 3 successful patterns
- Top 3 areas for improvement
- Anomalies/warnings detected
- Recommendations based on data

### Delivery
- In-app notification when report ready
- Report history (view past reports)
- Optional: export to markdown/HTML

## Acceptance Criteria

- [ ] Reports generate automatically on schedule
- [ ] Configurable report day/time
- [ ] Shows week-over-week comparison
- [ ] Identifies best patterns ("Starting with tests increased success by X%")
- [ ] Surfaces warnings ("3 PRs stuck in review >48hrs")
- [ ] Provides recommendations based on data
- [ ] View report history (past 4 weeks)
- [ ] Export report as markdown
- [ ] In-app notification when new report available
- [ ] Tests for report generation logic

## Mockup Reference

See #444 "Feature 4: Learning Insights" section for detailed mockup.

## Dependencies

- Depends on: #1066 (correlation analysis) - complete
- Depends on: #1067 (recommendation engine) - complete
- Related: #444 (parent epic)

---
Part of #444 (Loom Intelligence epic)

PR #5229, opened 2026-08-04
fix(gh-cached): scope the cache directory per repo to stop cross-repo leakage

### LINKED ISSUE #5224: gh-cached: cache key omits repo/cwd, causing cross-repo data leakage on multi-repo hosts
## Bug Description

`.loom/scripts/gh-cached`'s `cache_key()` (line 84-88) generates the cache key by hashing **only** the `gh` CLI arguments:

```python
def cache_key(args: list[str]) -> str:
    """Generate a cache key from the gh command arguments."""
    raw = " ".join(args)
    return hashlib.sha256(raw.encode()).hexdigest()[:16]
```

The cache is file-backed at a host-wide, non-repo-scoped location (`/tmp/gh-cache/` by default, `CACHE_DIR` env-overridable but still a single directory, not per-repo). Most `gh` commands routed through this wrapper (`gh pr list --label=... --state=open --limit 500`, `gh issue list --search ...`) do **not** pass an explicit `--repo` flag — `gh` resolves the target repo implicitly from the current working directory's git remote.

**Result:** on a host running multiple repos (the normal Loom fleet shape — many worktrees/checkouts across different projects on one box), two different repos issuing the *textually identical* `gh` invocation produce the *identical* cache key, so one repo's Judge/Champion/Curator can silently read and act on another, unrelated repo's cached PR/issue list.

## Reproduction

Confirmed live during a Judge pass on 2026-08-04 in the `rjwalters/loom` repo:

```
$ "$GH_READ" pr list --label="loom:review-requested" --state=open --limit 500
488  feat(functional-verification): RSA modexp core + bit-exact cocotb testbench; sky130 baseline (Phase 1 stop-and-reassess)  feature/issue-486  OPEN  2026-08-04T04:19:52Z
```

This PR (#488, about an RSA modexp/cocotb/sky130 hardware project) does not exist in `rjwalters/loom` at all — it's clearly cached output from an entirely different repository that had recently run the identical `gh pr list` command against the shared `/tmp/gh-cache/`. After `gh-cached --clear-cache` and re-running the same command, the correct `rjwalters/loom` queue was returned (4 PRs: #5223, #5221, #5220, #5205).

## Impact

- **Severity: High.** Every read routed through `$GH_READ` in `judge.md`, `champion-*.md`, and `curator.md` is affected on any host that runs sessions against more than one repo (the documented multi-account/fleet operating model). A stale/wrong queue listing can cause an agent to evaluate, claim, promote, or comment on the wrong repo's issues/PRs, or to falsely conclude "no work available" for the correct repo while acting on another repo's stale cache.
- Not a rate-limit or staleness issue in the intended sense (TTL working as designed) — this is outright data leakage across trust/repo boundaries, which the design doc's TTL/invalidation model never anticipated.

## Suggested Fix

Incorporate the resolved `owner/repo` (or, as a cheaper proxy, the absolute path of `git rev-parse --show-toplevel`) into the cache key — either by adding it into the `raw` string hashed in `cache_key()`, or by scoping `CACHE_DIR` per-repo (e.g. `$CACHE_DIR/<repo-hash>/`). Should apply even when the underlying `gh` command has no explicit `--repo` flag, since that's precisely the case that resolves via cwd today.

Discovered during Judge evaluation of PR #5220 in `rjwalters/loom`.

---

## Curator Enhancement

Verified against `origin/main` at `a402b4ad`, 2026-08-04.

**Path correction**: the canonical source is `defaults/scripts/gh-cached` (`.loom/scripts/gh-cached` is the installed copy, not tracked in git — it's refreshed from `defaults/` via `resync-installed.sh`). `cache_key()` is at `defaults/scripts/gh-cached:85-88` as of `a402b4ad` (not 84-88 — re-verify before relying on this line number, it drifts). The env var is **`GH_CACHE_DIR`**, not `CACHE_DIR` as stated in the original body (`CACHE_DIR` is the module-level Python variable it's read into — `defaults/scripts/gh-cached:36`).

Root cause confirmed: `cache_key()` hashes only `" ".join(args)`, and `CACHE_DIR` defaults to a single host-wide `/tmp/gh-cache/`. The design doc (`defaults/docs/gh-cached.md:6-7`) explicitly assumes concurrent sessions are "running concurrently against the **same** repo" — this bug is exactly that assumption breaking on a multi-repo host.

**Additional risk beyond the reported read-leakage, worth scoping into the fix**: `enforce_max_size()`, `invalidate_for_resource()`, `invalidate_all_for_type()`, and `clear_cache()` (`defaults/scripts/gh-cached:163-254`) all `os.listdir(CACHE_DIR)` over the *entire* shared directory. If the fix only extends `cache_key()`'s hash input (adds repo to the hashed string) but leaves `CACHE_DIR` a single global directory, a narrower but related bug survives: `invalidate_for_resource("issue", "42")` from repo A would still unlink repo B's cached entry for issue #42 too (same `resource_type`/`resource_id` match, `defaults/scripts/gh-cached:191-230`), since repo B very plausibly also has an issue #42. This wouldn't leak *data* across repos, but it would cause spurious over-invalidation (cache misses / extra API calls) — the kind of collateral bug that would pass a test suite that only checks the reported read-leakage and not invalidation isolation. **Recommend repo-scoping `CACHE_DIR` itself** (e.g. `$GH_CACHE_DIR/<repo-hash>/`) over only extending `cache_key()`, since that fixes both the read-leakage and the invalidation-isolation gap with one change — every function above operates on `CACHE_DIR` as-is with no other modification needed.

Repo identity should resolve via `git rev-parse --show-toplevel` (cheap, no network, no added `gh` quota cost) rather than `gh repo view` (adds a live API call to every cache lookup, defeating some of the point of caching) — matches the existing pattern of resolving `$GH_READ` once per session via a local probe (`defaults/docs/gh-cached.md:20-23`).

No duplicate or related open work found (`check-duplicate.sh --include-merged-prs --issue 5224`, 2026-08-04).

## Acceptance Criteria

- [ ] Two different repos issuing the textually identical `gh` invocation (e.g. `gh pr list --label=loom:review-requested --state=open --limit 500`) never read each other's cached entries.
- [ ] The fix scopes `CACHE_DIR` per-repo (not just `cache_key()`'s hash input), so `enforce_max_size()`, `invalidate_for_resource()`, `invalidate_all_for_type()`, and `clear_cache()` are also naturally repo-isolated — or, if a narrower fix is chosen instead, explain in the PR why the invalidation-isolation gap above doesn't apply / is out of scope.
- [ ] Repo identity is resolved without adding a live `gh`/network call to the cache-lookup hot path (e.g. `git rev-parse --show-toplevel`).
- [ ] Existing `gh-cached` behavior for a single-repo host is unchanged (no regression in `defaults/scripts/tests/test-gh-cached.sh`).

## Affected Files

- `defaults/scripts/gh-cached` — `cache_key()` (~line 85-88) and/or `CACHE_DIR` resolution (~line 36); the four `os.listdir(CACHE_DIR)`-walking functions (~lines 163-254) if the `CACHE_DIR`-scoping approach is taken.
- `defaults/scripts/tests/test-gh-cached.sh` — add a test asserting two distinct repo contexts (e.g. two temp git dirs, or an injected repo-identity override for the test) produce isolated cache entries for the same `gh` args.
- `defaults/docs/gh-cached.md` — update the "running concurrently against the same repo" assumption (line 7) once multi-repo isolation lands.
- `.loom/scripts/gh-cached` is the installed copy of the first file above — do not edit it directly; it's refreshed by `resync-installed.sh` from `defaults/scripts/gh-cached`.

## Test Plan

- [ ] Automated tests: extend `defaults/scripts/tests/test-gh-cached.sh` with a case simulating two repos (distinct `git rev-parse --show-toplevel` roots or equivalent) issuing the same `gh` args and asserting they do not share a cache hit.
- [ ] Manual verification: reproduce the original report — run the same `gh pr list ...` command from two different repo checkouts against the same `GH_CACHE_DIR`, confirm each returns its own repo's data.
- [ ] Edge cases: confirm `invalidate_for_resource()` no longer cross-contaminates when two repos have a numerically-identical resource id (e.g. both have an issue/PR #42) — see "Additional risk" above.

rjwalters:
## Summary

- `gh-cached`'s `cache_key()` hashed only the `gh` CLI argv, and `CACHE_DIR` defaulted to a single host-wide `/tmp/gh-cache/`. Two different repos on the same host issuing the textually identical `gh` invocation (e.g. `gh pr list --label=loom:review-requested --state=open --limit 500`) therefore produced the same cache key/path, so one repo's Judge/Champion/Curator could silently read another, unrelated repo's cached PR/issue data — confirmed live on 2026-08-04 in this repo.
- Fixed by scoping `CACHE_DIR` itself per repo (rather than only extending `cache_key()`'s hash input), following the curator's recommendation, so the fix covers both the read-leakage and the related invalidation-isolation gap in one change.

## Changes

- `defaults/scripts/gh-cached`: added `resolve_repo_id()`, which resolves the repo identity via `git rev-parse --show-toplevel` (cheap, local, no network call — matches the existing pattern of resolving `$GH_READ` once per session via a local probe). `CACHE_DIR` is now `GH_CACHE_DIR/<sha256(repo_id)[:16]>/` instead of a flat `GH_CACHE_DIR`. `GH_CACHE_REPO_ID` overrides the resolved identity directly (for tests, and non-git working directories that still want isolation). Since `enforce_max_size()`, `invalidate_for_resource()`, `invalidate_all_for_type()`, and `clear_cache()` all already operate on `CACHE_DIR` as-is, they become naturally repo-isolated with no further changes. Added a debug line reporting the resolved `REPO_ID`/`CACHE_DIR` under `GH_CACHE_DEBUG=1`.
- `defaults/scripts/tests/test-gh-cached.sh`: added a new "Multi-repo cache isolation (#5224)" block — two synthetic repo identities (`GH_CACHE_REPO_ID` override) issuing the identical `gh pr list` never share a cache read, repo A's own cache stays warm/unaffected by repo B's read, and a mutation issued under repo A's identity does not invalidate repo B's cached entry for a numerically-identical resource id (the "additional risk" the curator flagged). A second block re-derives the same guarantees through two real, distinct `git init` toplevels (not just the override), proving the default (non-override) resolution path isolates too.
- `defaults/docs/gh-cached.md`: updated the "running concurrently against the same repo" framing to state the new per-repo `CACHE_DIR` scoping explicitly, and extended the "Automated" verification section to mention the new test coverage.

- [x] Two different repos issuing the textually identical `gh` invocation never read each other's cached entries — verified by the new "Multi-repo cache isolation" test block (both the `GH_CACHE_REPO_ID`-override and real-git-toplevel variants).
- [x] The fix scopes `CACHE_DIR` per-repo (not just `cache_key()`'s hash input) — `enforce_max_size()`, `invalidate_for_resource()`, `invalidate_all_for_type()`, and `clear_cache()` are all naturally repo-isolated since they operate on `CACHE_DIR` unmodified; verified by the invalidation-isolation test case (a repo-A mutation on resource id `4242` does not invalidate repo B's cached entry for the same numeric id).
- [x] Repo identity is resolved without a live `gh`/network call — `git rev-parse --show-toplevel` only.
- [x] Existing single-repo-host behavior is unchanged — all 24 pre-existing assertions in `test-gh-cached.sh` still pass unmodified (the repo-scoping hash is constant across a single test run's cwd, so cache-key collisions/hits behave exactly as before).

## Test Plan

Ran `defaults/scripts/tests/test-gh-cached.sh` in the worktree: **31/31 passed** (24 pre-existing + 7 new assertions across the two new multi-repo isolation blocks). `python3 -m py_compile` on the modified script succeeds.

Closes #5224

### DIFF
--- a/defaults/docs/gh-cached.md
+++ b/defaults/docs/gh-cached.md
@@ -6,6 +6,15 @@ cache under `/tmp/gh-cache/`. Because the cache is on disk, it is shared
 **across processes on one host** — the N sweep / Judge / Champion sessions
 running concurrently against the same repo hit the same entries.
 
+The cache directory is further scoped **per repo** (#5224): the wrapper
+resolves `git rev-parse --show-toplevel` once per invocation (cheap, local, no
+network) and hashes it into a subdirectory of `GH_CACHE_DIR`, so two different
+repos on the same host issuing the textually identical `gh` invocation never
+read (or invalidate) each other's cached entries — the sessions sharing
+entries above are always sessions against the *same* repo. `GH_CACHE_REPO_ID`
+overrides the resolved identity directly when needed (tests, or a working
+directory that isn't a git repo).
+
 This document is the policy reference for **which** reads in the
 `/loom:sweep`, `/loom:judge`, and Champion-PR-merge skills go through the
 wrapper and which are deliberately left uncached (#4667). The motivating
@@ -208,7 +217,11 @@ stub `gh` on `PATH` and a temp `GH_CACHE_DIR`, asserting: repeated `pr list` /
 `issue list` / `pr view` reads hit the cache; `--no-cache` bypasses; `gh pr
 checks` and `gh repo view` are never cached; a wrapped `gh pr edit` /
 `gh pr comment` invalidates the cached reads of that PR (including the cached
-listing that contains it); and a state change is observed once the TTL expires.
+listing that contains it); a state change is observed once the TTL expires;
+and (#5224) two different repo identities — both via `GH_CACHE_REPO_ID` and
+via two real, distinct git toplevels — never share a cache read or an
+invalidation for the textually identical `gh` args, even when both repos have
+a numerically-identical resource id.
 
 ### Manual (inherently runtime — cannot be unit-tested)
 
--- a/defaults/scripts/gh-cached
+++ b/defaults/scripts/gh-cached
@@ -5,7 +5,9 @@ Drop-in replacement for `gh` that caches read-only command responses
 to reduce API calls. Mutations bypass the cache and invalidate
 related entries.
 
-Cache is file-backed in /tmp/gh-cache/ for cross-process sharing.
+Cache is file-backed in /tmp/gh-cache/ for cross-process sharing, scoped into
+a per-repo subdirectory (#5224) so two different repos on the same host never
+share cache entries even when they issue textually identical `gh` commands.
 
 Usage:
     gh-cached issue view 42 --json labels
@@ -15,11 +17,17 @@ Usage:
     gh-cached --cache-stats                            # show hit/miss statistics
 
 Environment:
-    GH_CACHE_DIR       Cache directory (default: /tmp/gh-cache)
+    GH_CACHE_DIR       Cache directory base (default: /tmp/gh-cache); the
+                        actual cache lives under a per-repo subdirectory of
+                        this path — see resolve_repo_id() below.
     GH_CACHE_TTL       Default TTL in seconds (default: 30)
     GH_CACHE_MAX_SIZE  Max cached entries (default: 256)
     GH_CACHE_DISABLE   Set to "1" to disable caching entirely
     GH_CACHE_DEBUG     Set to "1" for debug logging to stderr
+    GH_CACHE_REPO_ID   Override the resolved repo identity used to scope
+                        GH_CACHE_DIR (default: `git rev-parse
+                        --show-toplevel`). Mainly for tests and non-git
+                        working directories that still want isolation.
 """
 
 from __future__ import annotations
@@ -33,7 +41,49 @@ import time
 
 # ─── Configuration ──────────────────────────────────────────────────────────
 
-CACHE_DIR = os.environ.get("GH_CACHE_DIR", "/tmp/gh-cache")
+
+def resolve_repo_id() -> str:
+    """Resolve a stable per-host-repo identity used to scope CACHE_DIR (#5224).
+
+    Two different repos on the same host issuing the textually identical `gh`
+    command must never share a cache entry — `cache_key()` alone only hashes
+    the `gh` argv, which says nothing about which repo it targets (most reads
+    routed through this wrapper have no explicit `--repo` and resolve via the
+    cwd's git remote). Scoping CACHE_DIR itself (rather than only extending
+    cache_key()'s hash input) also naturally repo-isolates enforce_max_size(),
+    invalidate_for_resource(), invalidate_all_for_type(), and clear_cache(),
+    since every one of them already operates on CACHE_DIR as-is.
+
+    Uses `git rev-parse --show-toplevel` — cheap, local, no network call —
+    mirroring the existing pattern of resolving $GH_READ once per session via
+    a local probe (defaults/docs/gh-cached.md). GH_CACHE_REPO_ID overrides the
+    resolved identity directly (used by tests, and as an escape hatch for
+    non-git working directories that still want isolation).
+    """
+    override = os.environ.get("GH_CACHE_REPO_ID", "")
+    if override:
+        return override
+    try:
+        proc = subprocess.run(
+            ["git", "rev-parse", "--show-toplevel"],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
+        if proc.returncode == 0:
+            top = proc.stdout.strip()
+            if top:
+                return top
+    except (OSError, subprocess.TimeoutExpired):
+        pass
+    # Not inside a git repo (or git unavailable): fall back to a fixed
+    # sentinel so behavior is deterministic rather than silently unscoped.
+    return "_no-repo"
+
+
+CACHE_DIR_BASE = os.environ.get("GH_CACHE_DIR", "/tmp/gh-cache")
+REPO_ID = resolve_repo_id()
+CACHE_DIR = os.path.join(CACHE_DIR_BASE, hashlib.sha256(REPO_ID.encode()).hexdigest()[:16])
 DEFAULT_TTL = int(os.environ.get("GH_CACHE_TTL", "30"))
 MAX_CACHE_SIZE = int(os.environ.get("GH_CACHE_MAX_SIZE", "256"))
 CACHE_DISABLED = os.environ.get("GH_CACHE_DISABLE", "") == "1"
@@ -399,6 +449,7 @@ def try_etag_list(resource_type: str, args: list[str]) -> tuple[str, int] | None
 
 def main() -> int:
     args = sys.argv[1:]
+    debug(f"REPO_ID={REPO_ID!r} CACHE_DIR={CACHE_DIR}")
 
     # Handle meta-commands
     if "--clear-cache" in args:
--- a/defaults/scripts/tests/test-gh-cached.sh
+++ b/defaults/scripts/tests/test-gh-cached.sh
@@ -395,6 +395,16 @@ if command -v git >/dev/null 2>&1; then
     echo '[{"number":2222,"labels":["loom:review-requested"]}]' > "$STATE_DIR/pr-list.json"
     outB=$(ghc_cwd "$REPO_B_DIR" pr list --label "loom:review-requested" --state open --limit 500)
     assert_eq "2" "$(call_count)" "two distinct real git toplevels issuing the same 'gh' args both hit the real gh (default git-based resolution, not just the override)"
+    case "$outA" in
+      *1111*) sawA="yes" ;;
+      *) sawA="no" ;;
+    esac
+    assert_eq "yes" "$sawA" "repo A (a distinct real git toplevel) reads its own PR #1111 from the real gh"
+    case "$outB" in
+      *2222*) sawB="yes" ;;
+      *) sawB="no" ;;
+    esac
+    assert_eq "yes" "$sawB" "repo B (a distinct real git toplevel) reads its own PR #2222, not repo A's cached response"
     case "$outB" in
       *1111*) leaked2="yes" ;;
       *) leaked2="no" ;;
--- a/defaults/scripts/tests/test-gh-cached.sh
+++ b/defaults/scripts/tests/test-gh-cached.sh
@@ -318,6 +318,94 @@ reset_cache
 LOOM_ETAG_LIST_DISABLE=1 ghc_etag issue list --label "loom:issue" --state open --json number,title >/dev/null
 assert_eq "1" "$(call_count)" "LOOM_ETAG_LIST_DISABLE=1 bypasses the ETag path (falls to gh)"
 
+# --- 10. Multi-repo cache isolation (#5224) --------------------------------
+# CACHE_DIR is scoped per-repo so two different repos on the same host never
+# share a cache entry for a textually identical `gh` invocation — the bug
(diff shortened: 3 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T05:08:25Z) — standing down without reclaiming. Not stomping.

rjwalters:

**Correctness of the fix (issue #5224 acceptance criteria)**: The approach is right and well-executed. `resolve_repo_id()` scopes `CACHE_DIR` itself (not just `cache_key()`'s hash input) via `git rev-parse --show-toplevel` — local, no network call, `timeout=5` guarded, with a deterministic `_no-repo` fallback for non-git cwds and a `GH_CACHE_REPO_ID` override for tests. Because `enforce_max_size()`, `invalidate_for_resource()`, `invalidate_all_for_type()`, and `clear_cache()` already operate on `CACHE_DIR` unmodified, scoping the directory itself naturally isolates all of them — exactly the "additional risk" called out in the issue (a repo-A mutation must not invalidate repo-B's cache for a numerically-identical resource id). Verified this directly:

- Ran `defaults/scripts/tests/test-gh-cached.sh` in the builder worktree (`.loom/worktrees/issue-5224`, HEAD `de4722c4`): **31/31 passed**, including all 7 new multi-repo isolation assertions (both the `GH_CACHE_REPO_ID`-override path and the real-distinct-`git init`-toplevel path), and all 24 pre-existing assertions unmodified — no regression to single-repo-host behavior.
- `python3 -m py_compile defaults/scripts/gh-cached` succeeds.
- Read through `gh-cached`'s diff — `CACHE_DIR = os.path.join(CACHE_DIR_BASE, hashlib.sha256(REPO_ID.encode()).hexdigest()[:16])` is correct and the debug line (`GH_CACHE_DEBUG=1`) is a nice touch for troubleshooting.
- No live `gh`/network call added to the cache-lookup hot path — confirmed, `resolve_repo_id()` only shells out to local `git`.

**Blocking issue — CI is red, not just pending**: `gh pr checks 5229` shows the **Shellcheck** check has **failed** (exit code 123), not merely pending:

```
defaults/scripts/tests/test-gh-cached.sh:394:5: warning: outA appears unused. Verify use (or export if used externally). [SC2034]
```

The repo's shellcheck CI step runs `shellcheck --severity=warning`, so this SC2034 warning is a hard failure, not advisory. It's a real defect in the new test code: in the "real distinct git toplevels" isolation block, `outA=$(ghc_cwd "$REPO_A_DIR" pr list ...)` is captured but never referenced — only `outB` is checked against the leak assertion. Contrast with the `GH_CACHE_REPO_ID`-override block just above it, where the analogous `repoA_out1`/`repoA_out2` pair *is* used (compared for equality to prove repo A's own cache stays warm and unaffected by repo B's read). The real-git-toplevel block is missing that same "repo A's own cache is unaffected" assertion — either drop the unused capture (`ghc_cwd "$REPO_A_DIR" ... >/dev/null`) or, better, add the missing assertion so the real-toplevel path gets the same coverage depth as the override path (re-read via `ghc_cwd "$REPO_A_DIR" ...` again and assert it equals `outA`, matching the override block's pattern).

**Merge state**: `mergeStateStatus` is `UNSTABLE` (consistent with the failing required Shellcheck check above), `mergeable: MERGEABLE` — no actual merge conflict, just the CI failure.

**Not blocking, minor observations for awareness (no action required)**:
- `resolve_repo_id()` runs unconditionally at module import time, even for `--clear-cache`/`--cache-stats`/passthrough commands and even when `GH_CACHE_DISABLE=1` — a local `git rev-parse` per invocation. Matches the existing "resolve `$GH_READ` once per session via a local probe" pattern the PR cites, and is cheap, so this is fine as-is.
- Docs update (`gh-cached.md`) is clear and accurately reflects the new per-repo scoping and test coverage.

Please fix the shellcheck failure (unused `outA`) and repost — everything else here is solid. Requesting changes for the CI failure only.

rjwalters:
Addressed the Shellcheck SC2034 blocker: `outA` is now consumed in a positive isolation assertion rather than left unused.

In the real-git-toplevel isolation block (`defaults/scripts/tests/test-gh-cached.sh`), instead of only checking that repo B does **not** see repo A's cached PR #1111, the test now also asserts:
- repo A reads its own PR #1111 from the real gh (`sawA == yes`)
- repo B reads its own PR #2222, not repo A's cached response (`sawB == yes`)

This strengthens the two-sided isolation guarantee and removes the unused-variable warning.

Verification:
- `defaults/scripts/tests/test-gh-cached.sh` — 33/33 passed (was 31; +2 new assertions).
- `shellcheck --severity=warning defaults/scripts/tests/test-gh-cached.sh` — exit 0, no warnings. (`gh-cached` itself is a `#!/usr/bin/env python3` script, so CI's shebang-aware shellcheck job skips it; running shellcheck against it directly produces spurious shell-parse errors and is not what CI does.)

Re-labeling back to `loom:review-requested`.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T05:13:28Z) — standing down without reclaiming. Not stomping.

rjwalters:
## Judge re-review: approved

Confirmed the Shellcheck blocker from my prior "changes requested" pass is resolved, and CI is now fully green.

**Shellcheck fix (commit e7b693d0)**: `outA` (previously captured but unused, triggering SC2034) is now consumed in a genuine positive assertion — repo A reads its own cached PR #1111 — alongside the existing `outB` positive assertion (PR #2222) and the pre-existing negative leak check (repo B does not see repo A's #1111). This strengthens the real-git-toplevel isolation test rather than just silencing the linter.

Verified locally (not just taking Doctor's word for it):
- `shellcheck --severity=warning defaults/scripts/tests/test-gh-cached.sh` → clean, exit 0
- `bash defaults/scripts/tests/test-gh-cached.sh` → 33/33 passed
- `bash defaults/scripts/tests/test-resync-installed.sh` → 114/114 passed (unrelated file also touched on this branch, per `git status` at dispatch time — confirmed unaffected)

**Core fix logic re-verified sound** (`defaults/scripts/gh-cached`): `resolve_repo_id()` resolves a per-repo identity via `git rev-parse --show-toplevel` (cheap, no network, 5s timeout guard), falls back to a deterministic `_no-repo` sentinel outside a git repo, and honors `GH_CACHE_REPO_ID` as an override for tests/non-git dirs. `CACHE_DIR` is now `CACHE_DIR_BASE/sha256(REPO_ID)[:16]`, which naturally repo-scopes every `CACHE_DIR`-walking function (`enforce_max_size`, `invalidate_for_resource`, `invalidate_all_for_type`, `clear_cache`) along with plain reads — exactly the isolation issue #5224 describes. No regressions spotted.

**CI status** (`gh api .../commits/e7b693d0/check-runs`, all 13 checks): Shellcheck, Shell Test Suites (hermetic), Codex Adapter Smoke, Installer Integration Tests, CLAUDE.md Line Budget, Docs/Defaults Parity Check, AGENTS.md Sync Check, Detect Changes all `success`; Rust/Dashboard/MCP jobs `skipped` (expected — no Rust/dashboard files touched). `mergeable_state`: `clean`.

Approving.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (another Judge's approval landed at 2026-08-04T05:19:16Z while I was polling CI). My independent review reached the same conclusion (all CI green, 33/33 local test pass, code and docs look correct) — discarding my review; not writing any label.

PR #4357, opened 2026-07-29
feat(fleet): add `fleet add-worker` to bootstrap a provisioned host into a loom worker

rjwalters:
## Summary

Adds `loom-daemon fleet add-worker <ssh-host> --repo <owner/name> [--repo …]` — the first child of epic #4340. It takes a reachable, already-provisioned Ubuntu host to "daemon running, workspace registered, tokens ranked, dispatch verified" in one **idempotent** command over `ssh`. Generic VM provisioning stays in `repo:remote` per the epic's boundary decision; this consumes "a reachable box + an SSH alias".

The bootstrap is modeled as an ordered **plan** of named steps, each with a `check` (already done?) → `apply` → `verify` shape. Rust owns the plan/ordering/checklist; the per-phase shell is rendered as heredoc templates and executed over a `CommandRunner` trait — the production `SshRunner`, or a mocked runner in tests (no live box needed). `--dry-run` prints the full plan without contacting the host.

## Changes

- **`loom-daemon/src/fleet/mod.rs`** (new) — module root: the `CommandRunner` seam, the `Step`/`Plan`/`PlanEntry` model, `execute_step`/`execute_plan` (halt-on-first-failure), checklist + dry-run rendering, and the machine-level `FleetRegistry` (`~/.loom/fleet.json`, `LOOM_FLEET_PATH` override, atomic save, dedup-on-ssh-host upsert) following `workspace_registry.rs` conventions.
- **`loom-daemon/src/fleet/add_worker.rs`** (new) — `SshRunner`, `AddWorkerConfig`, local secrets preflight, the pure `build_plan`, the heredoc shell templates for each step, and the `run` orchestration (preflight → plan → dry-run/execute → checklist → fleet-registry upsert).
- **`loom-daemon/src/main.rs`** — thin clap wiring: a `Fleet` command + `FleetAction::AddWorker` subcommand and a `handle_fleet_command` dispatch arm (all logic stays in the module).
- **`loom-daemon/src/lib.rs`** — register the `fleet` module.
- **`defaults/docs/daemon-reference.md`** — document the `fleet add-worker` surface (new "Fleet" section; no CLAUDE.md change, budget check passes at 312/320).

The 8 pilot steps are encoded as 13 plan entries (base-deps incl. **libsqlite3-dev** per safehouse#38, machine-layout, claude-code, forge-auth, token-accounts/pool/ranking, workspace-clone, workspace-register, daemon-unit, idle-shutdown, safehouse, verify).

## Acceptance Criteria Verification

[...]

Issue #5226, opened 2026-08-04
fix(guard): is_interpreter_opener() still fails OPEN on bare VAR= / sudo / timeout / quoted interpreter openers


rjwalters:
## Summary

`is_interpreter_opener()` in `guard-destructive-generic.sh` (added by #5198,
widened by #5205) decides whether a heredoc body is left **visible** to the
`gh-api-rawfield-body-literal-at` catastrophic-tier DENY check. When it returns
false, `mask_heredoc_bodies_selective()` blanks the body, and a **live**
`gh api ... -f body=` + literal-@-path invocation inside that heredoc silently
flips from DENY to ALLOW — the exact #4523/#4601/#4685 data-loss shape.

#5205 closed the path-qualified (`/bin/bash`, `./bash`) and wrapper-prefixed
(`env`/`command`/`exec`/`builtin`) classes. A handful of adjacent command-word
shapes still resolve to the same real interpreter but are not recognized.

## Reproduction (verified against PR #5205 head `6523d882`)

Feeding each opener through `defaults/hooks/guard-destructive-generic.sh` with
the standard denied payload in the heredoc body:

| Opener | Decision | Notes |
|---|---|---|
| `bash <<EOF` (control) | deny | correct |
| `LC_ALL=C bash <<EOF` | **allow (gap)** | bare `VAR=value` prefix, no `env` |
| `sudo bash <<EOF` | **allow (gap)** | `sudo` not in the stripped-wrapper list |
| `cat <<EOF \| sudo bash` | **allow (gap)** | same, in pipe position |
| `timeout 60 bash <<EOF` | **allow (gap)** | exec-wrapper class |
| `"bash" <<EOF` | **allow (gap)** | quoted command word |
| `\bash <<EOF` | **allow (gap)** | backslash-escaped (alias-dodge idiom) |

The `LC_ALL=C bash` case is notable because the wrapper-stripping loop
**already** handles `VAR=value` tokens — but only *after* it has seen an
`env`/`command`/`exec`/`builtin` token first, so a bare assignment prefix
(valid shell, extremely common) falls straight through to `break`.

## Affected files

- `defaults/hooks/guard-destructive-generic.sh` — `is_interpreter_opener()` /
  `_interp_basename()`
- `.loom/hooks/guard-destructive-generic.sh` — must stay byte-identical
- `tests/hooks/test-guard-destructive.sh` — regression cases

## Suggested direction (for the Curator to weigh)

[...]

PR #5231, opened 2026-08-04
fix(guard): mask heredoc-quoted dangerous-command examples in the catastrophic scan

### LINKED ISSUE #5216: Guard: catastrophic rm-scope pattern false-positives on a backtick-quoted example inside a heredoc-wrapped gh pr comment body

rjwalters:
## Summary

Fixes the catastrophic `ALWAYS_BLOCK_PATTERNS` scan false-positiving when a
heredoc-wrapped `--body`/`-m`/`--title`/`--notes`/`--comment` value (e.g.
`gh pr comment N --body "$(cat <<'EOF2' ... EOF2)"`) merely *quotes* a
dangerous-command example (like a recursive force-remove-of-root or a
force-push-to-main example) in prose/backticks.

`strip_literal_text()` deliberately declines to redact any quoted span
containing `$(` (its documented anti-smuggling safety floor), so a
heredoc-wrapped body is never redacted and gets scanned as if it were a live
command — a real Judge approval comment on PR #4357 was blocked this way.

## Fix

The catastrophic scan (`ALWAYS_BLOCK_PATTERNS`), the SQL DDL check, and the
rm-scope-outside-repo target extraction now additionally run over a
heredoc-body-masked working copy built with `mask_heredoc_bodies_selective()`
(#5205), independent of the `$(`/backtick exclusion above.

This deliberately uses the **selective**, interpreter-aware masker (#5205's
precedent from the `gh-api-rawfield-body-literal-at` check), not the plain
`mask_heredoc_bodies()` primitive `extract_write_targets()` uses. Plain
masking has a documented "Known Limitation" (#5117): it masks a heredoc body
regardless of what consumes it, including a body genuinely fed to an inner
interpreter (`bash <<'EOF' ... EOF`, `... | sh`). Using it unconditionally for
this catastrophic-tier hard-deny floor would have silently allowed exactly
the smuggled-command shape the scan exists to catch. The selective variant
leaves an interpreter-fed heredoc block's body **unmasked** (still visible to
the scan), so a live command inside such a body still hard-denies — only a
heredoc body that is merely quoted inside a recognized flag value is
narrowed.

An in-worktree partial implementation from a previously-killed builder run
used the *plain* `mask_heredoc_bodies()` for these three call sites, which
would have reopened the interpreter-fed-heredoc bypass #5205 just closed for
a sibling check. This PR rebuilds those three call sites on the selective
variant instead.

## Testing

- `./tests/hooks/test-guard-destructive.sh` — full suite, 719/719 green
  (baseline + 18 new assertions for this issue, net of a rewritten block).
- New regression coverage:
  - The issue's own repro (rm-scope-outside-repo, heredoc-wrapped) — allow.
  - Shared-mechanism coverage: force-push, SQL DDL (plain single-line and
    heredoc-wrapped), docker-prune, aws-s3-rm/rb, curl/wget-pipe-shell
    heredoc-quoted examples — allow (or, for the two aws-s3 catastrophic
    patterns, correctly downgrade to the separate, out-of-scope
    `guards.cloudCli` ask tier rather than a hard deny).
  - Interpreter-fed heredoc safety floor: `bash <<'EOF' ... EOF` and
    `cat <<'EOF' ... EOF | sh` bodies containing a live dangerous command
    still hard-deny.
  - Regression: a live (non-heredoc) dangerous command, and a
    `$(...)`-command-substitution-smuggled one, still hard-deny.
- Manual repro of the issue's exact incident command and the #3679
  force-push-analog case both now silently allow; the interpreter-fed-heredoc
  case still denies.
- `shellcheck -S warning` clean on the modified hook.
- `.loom/hooks/guard-destructive-generic.sh` and
  `defaults/hooks/guard-destructive-generic.sh` kept byte-identical.

Also updates `defaults/docs/guard-hooks.md` with a "Third refinement pass
(#5216)" bullet documenting the new heredoc-masking behavior and its
interpreter-fed-heredoc carve-out.

Closes #5216


rjwalters:
Closing as duplicate — issue #5216 was already fixed and merged via #5227
(`fix(guard): mask inert cat-heredoc bodies in text-carrying flag values`,
merged 2026-08-04T05:31:08Z, ~1 minute before this PR was opened).

I inherited a pre-existing worktree from a previously-killed builder run for
this issue and built out a fix independently (rebuilding the partial WIP on
`mask_heredoc_bodies_selective()`, #5205's interpreter-aware masker) without
realizing a concurrent builder had already landed a different, more tightly
scoped fix in the meantime — `strip_literal_text()`'s new
`mask_flag_cat_heredocs()`, which recognizes only the exact
`<flag> "$(cat <<QUOTED_DELIM ... )"` shape structurally (quoted delimiter,
closed in-buffer, substitution closes immediately after) rather than via an
interpreter-word allowlist. That approach is arguably safer than the one in
this PR: it can't inherit the class of gaps #5226 is currently tracking for
`is_interpreter_opener()`, since it never needs to classify "is this heredoc
interpreter-fed" at all — the recognized shape structurally excludes any
`bash <<EOF` / `... | sh` opener from matching in the first place.

No further action needed here. Not merging this PR.

rjwalters:
❌ **Closing — superseded by already-merged #5227**

This PR targets issue #5216 (heredoc-wrapped flag values quoting a dangerous-command example false-positiving the catastrophic scan), but **#5227 already fixed and closed #5216** (merged 2026-08-04T05:31:08Z, ~6 minutes before this PR's rebase attempt), via a different, more surgical mechanism: a `mask_flag_cat_heredocs()` pre-pass baked directly into `strip_literal_text()`, rather than this PR's approach of composing separate `mask_heredoc_bodies_selective()` scans on top of `COMMAND_ASK_SCAN`/`COMMAND_NO_COMMENT`/`COMMAND_NO_LITERAL_TEXT`.

Verified on `origin/main`:
- Issue #5216 is `CLOSED`, closed by #5227.
- `tests/hooks/test-guard-destructive.sh` on main already has a full `--- #5216 ---` section (29 assertions) covering the exact same scenarios this PR's test plan lists: heredoc-wrapped SQL DDL, docker-prune, aws-s3-rm/rb, curl-pipe-shell, force-push examples allowed; interpreter-fed heredocs (`bash <<EOF`, `cat <<EOF | sh`, `sh -s <<EOF`) still correctly denied.

Attempting the automated rebase produced heavy conflicts across `.loom/hooks/guard-destructive-generic.sh`, `defaults/hooks/guard-destructive-generic.sh`, `defaults/docs/guard-hooks.md`, and `tests/hooks/test-guard-destructive.sh` — all overlapping-logic conflicts in the same functions/checks #5227 already rewrote, not simple non-overlapping additions. Rebasing this PR would mean re-deriving a second, competing implementation of a problem that's already solved and tested on main.

Since the underlying issue is already closed by a merged fix, this PR should not land — closing as superseded rather than routing to Doctor for a conflict resolution that would just re-implement #5227's work a second way.

PR #5233, opened 2026-08-04
fix(guard): exclude heredoc redirection tokens from tee/cp/mv/sed write-target scan

### LINKED ISSUE #5232: Guard: tee heredoc delimiter misparsed as write target, false worktree-isolation DENY
## Bug Description

`extract_write_targets()` in `defaults/hooks/guard-destructive-generic.sh` (the `tee` branch, ~line 2862-2866) treats **every** non-flag token after `tee <file>` as a write target — including the heredoc redirection token that follows it (`<<EOF` / `<<'EOF'`). That bogus "target" then gets resolved against `curcwd` (e.g. `/home/ubuntu/GitHub/loom/<<EOF`) and fed into the worktree-isolation confinement check, which falsely fires whenever the acting repo has **any** sibling Loom-managed worktree — the normal state for the primary clone of any actively-worked Loom repo.

Net effect: the extremely common `tee <file> <<'EOF' ... EOF` idiom (and its `sudo tee` variant) gets a false `DENY` with the worktree-isolation-bypass message, even though the write target is correct and entirely uninvolved with any worktree.

Same class of bug likely affects the `cp`/`mv` branch (~line 2881-2890), which also takes the *last* non-flag token as its target and would similarly misfire if a heredoc operator trailed the real arguments.

## Reproduction Steps

1. Checkout main: `git checkout main && git pull` (repro confirmed at commit `6ff58955`, latest main as of this audit)
2. In a repo that has at least one `.loom/worktrees/<x>/.loom-managed` sentinel present (i.e. any repo actively using Loom), run the guard directly:
   ```bash
   CMD=$'tee /tmp/report.md <<\'EOF\'\nsome text\nEOF\necho done'
   jq -n --arg cmd "$CMD" --arg cwd "$PWD" '{tool_name:"Bash", tool_input:{command:$cmd}, cwd:$cwd}' \
     | defaults/hooks/guard-destructive-generic.sh
   ```
3. Observe: the guard returns `permissionDecision: deny` citing `/repo/<<EOF` as the offending write path.

This is also caught by the repo's own test suite — `tests/hooks/test-guard-destructive.sh` carries the exact regression assertions "#5226/#5181: A heredoc body to 'tee file' (non-interpreter sink) ... stays allowed" and the `sudo tee` variant (added in the most recently merged commit, PR #5230). Those assertions **fail** when the test's `$REPO_ROOT` has any sibling managed worktree present — which is the case for this repo's own primary clone (dozens of stale `.loom/worktrees/*` entries) but NOT for a fresh CI checkout, so `run-ci-suites.sh` passes in CI while the bug is real and reachable in production use.

```bash
$ bash tests/hooks/test-guard-destructive.sh
  FAIL: #5226/#5181: A heredoc body to 'tee file' (non-interpreter sink) that merely quotes the phrase stays allowed
  FAIL: #5226/#5181: A heredoc body to 'sudo tee file' (wrapped non-interpreter sink) stays allowed
Total: 746  Passed: 744  Failed: 2
```

## Expected Behavior

`tee <file> <<'EOF' ... EOF` and `sudo tee <file> <<'EOF' ... EOF` should ALLOW (the heredoc delimiter is not a write target), same as the other non-interpreter-sink cases the #5226/#5181 regression suite covers.

## Actual Behavior

Both cases DENY with a worktree-isolation-bypass message pointing at a nonsensical `<<EOF` "path", whenever the repo has a sibling Loom-managed worktree.

## Impact

Any Builder/Doctor/other Bash-tool session writing a report or log file via `tee ... <<EOF` from the primary clone, while any other issue worktree is checked out (the default steady state for an actively-worked repo), gets a false DENY and is told to `cd` into a worktree it isn't actually trying to bypass. This is a real, currently-reachable false positive, not merely a test artifact — CI's hermetic checkout (no sibling worktrees) just doesn't happen to exercise the failing precondition.

## Environment

- OS: `Linux 7.0.0-1009-aws` (GitHub Actions-equivalent Ubuntu host)
- Commit: `6ff58955` (latest main as of this audit)
- Build/tests: Rust workspace build + 3571 nextest tests, mcp-loom build + 54 tests, dashboard backend (252) + web (466) tests, and the rest of `run-ci-suites.sh`'s 102 wired suites all pass. This is the only failure found.

---
Discovered during main branch audit.

rjwalters:
## Summary
Fixes a false `DENY` in the worktree-isolation guard: `extract_write_targets()`'s `tee`, `sed -i`, and `cp`/`mv` branches in `defaults/hooks/guard-destructive-generic.sh` treated a trailing heredoc/herestring redirection operator (`<<EOF`, `<<-EOF`, `<<'EOF'`, the bare space-separated `<< EOF` form, and the `<<<` herestring) as an extra write-target candidate. That bogus target resolved against `curcwd` and tripped the worktree-isolation confinement check, so the extremely common `tee <file> <<'EOF' ... EOF` idiom (and `sudo tee` variant) falsely denied whenever the repo had a sibling Loom-managed worktree — the normal steady state for this repo's own primary clone.

## Changes
- `defaults/hooks/guard-destructive-generic.sh`: in the `tee`, `sed -i`, and `cp`/`mv` branches of `extract_write_targets()`, exclude any token that is (or starts with) a heredoc/herestring redirection operator from the write-target candidate list. A **bare** operator token — `<<`, `<<-`, **or `<<<`**, with the delimiter/content word on its own token — additionally consumes the ONE following word, so neither the operator, nor the heredoc delimiter, nor the herestring content is misread as a target.
  - Consuming exactly one word is shell-accurate for both forms: bash's herestring takes a single word, so in `tee f <<< a b` the `b` really *is* a `tee` operand and is deliberately still scanned (covered by a dedicated deny assertion).
- `tests/hooks/test-guard-destructive.sh`: added 13 hermetic regression tests (using the existing `make_wt_repo()` fixture, independent of whether the ambient checkout happens to have a sibling managed worktree) — 6 covering `tee`, `sudo tee`, `cp`, and `sed -i` with trailing heredoc operators, and 7 covering the `<<<` herestring forms (bare, quoted-content, attached) plus two control denies (a real main-checkout target with a trailing herestring, and a real operand *after* a consumed herestring word).

## Judge follow-up (herestring half — addressed)
The first revision's comments claimed `<<<` coverage but only half-delivered it: `/^<<-?/` matched a `<<<` token (so the operator itself was excluded), but the `j++` that consumes the following word only fired for `<<` / `<<-`. For a herestring that following word is real content, so it fell through and became a *different* phantom target — the same false DENY, one token over:

```
CMD='tee /tmp/loom-test-report.md <<< hello'
origin/main       -> DENY citing target '<repo>/<<<'
first revision    -> DENY citing target '<repo>/hello'
this revision     -> ALLOW (empty output, exit 0)
```

`<<<` is now included in the bare-operator test at **all three** exclusion sites, with regression coverage.

| Reproduce the bug first (confirm DENY) | ✅ | Ran the issue's exact repro command against pre-fix code from this repo's primary clone (which has sibling managed worktrees) — confirmed `DENY` citing `<repo>/<<EOF`; likewise `DENY` citing `<repo>/<<<` for the herestring form |
| Apply fix | ✅ | See Changes above |
| Re-run repro (confirm ALLOW) | ✅ | Both repro commands against the fixed guard now return empty output (ALLOW) |
| `bash tests/hooks/test-guard-destructive.sh` passes 746/746 (or more) | ✅ | `Total: 759  Passed: 759  Failed: 0` (746 original + 13 new #5232 regression tests) |
| Add regression test coverage if not already adequate | ✅ | Added 13 hermetic tests (`make_wt_repo()`-based) so the "sibling worktree present" precondition is deterministic in any checkout. Each new "should ALLOW" test was verified to fail (DENY) against the code that lacks the corresponding fix — the 4 bare-`<<<` allow cases fail at 755/759 against the pre-herestring-fix revision and pass at 759/759 after |
| Run broader hook test suite for regressions | ✅ | `tests/hooks/test-guard-destructive-dispatcher.sh`: 12/12 passed. `shellcheck` on both changed files shows no new findings vs. the pre-change baseline (info-level-only, pre-existing) |

## Test Plan
```
$ bash tests/hooks/test-guard-destructive.sh
...
=========================================
  Total:  759
  Passed: 759
  Failed: 0
=========================================
ALL TESTS PASSED

$ bash tests/hooks/test-guard-destructive-dispatcher.sh
=========================================
Results: 12 passed, 0 failed
=========================================
```

Closes #5232

### DIFF
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -2862,6 +2862,24 @@ extract_write_targets() {
             if (toks[1] == "tee") {
                 for (j = 2; j <= m; j++) {
                     if (toks[j] == "" || toks[j] ~ /^-/) continue
+                    # Heredoc/herestring redirection (attached or quoted
+                    # delimiter, or the bare double-angle-bracket / dashed
+                    # form, plus the triple-angle-bracket herestring) is a
+                    # REDIRECTION OPERATOR feeding the tee commands stdin,
+                    # never a write-target argument -- but it is still just
+                    # another whitespace-bounded non-flag token to this
+                    # scanner, so without this exclusion the delimiter (or the
+                    # operator itself, in the space-separated bare spelling)
+                    # was misread as an extra file target and resolved against
+                    # curcwd, producing a bogus "<repo>/<<EOF" write that the
+                    # worktree-isolation check below then falsely denied
+                    # (#5232). A BARE operator token (exactly the two- or
+                    # three-char form with no attached delimiter) also
+                    # consumes the following delimiter word.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     print curcwd SEP toks[j]
                 }
             } else if (toks[1] == "sed") {
@@ -2872,6 +2890,14 @@ extract_write_targets() {
                     if (toks[j] ~ /^-i/) has_i = 1
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
+                    # Same heredoc/herestring exclusion as the `tee` branch
+                    # above (#5232) -- a trailing `sed -i ... file <<EOF` must
+                    # not misread the redirection operator/delimiter as an
+                    # extra file operand.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     nf++
                     nfargs[nf] = toks[j]
                 }
@@ -2884,6 +2910,15 @@ extract_write_targets() {
                 for (j = 2; j <= m; j++) {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
+                    # Same heredoc/herestring exclusion as the `tee` branch
+                    # above (#5232) -- without it a trailing `<<EOF` after the
+                    # real cp/mv operands was misread as the LAST non-flag
+                    # token (the field this branch treats as the destination),
+                    # so it would win over the real destination entirely.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     nf++
                     nfargs[nf] = toks[j]
                 }
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -3464,11 +3464,15 @@ extract_write_targets() {
                     # was misread as an extra file target and resolved against
                     # curcwd, producing a bogus "<repo>/<<EOF" write that the
                     # worktree-isolation check below then falsely denied
-                    # (#5232). A BARE operator token (exactly the two- or
-                    # three-char form with no attached delimiter) also
-                    # consumes the following delimiter word.
+                    # (#5232). A BARE operator token (`<<`, `<<-`, `<<<` with
+                    # no attached delimiter/content) also consumes the ONE
+                    # following word -- the heredoc delimiter, or the
+                    # herestring content. Consuming exactly one word is
+                    # shell-accurate for both: a herestring takes a single
+                    # word, so in `tee f <<< a b` the `b` really IS a tee
+                    # operand and must still be scanned.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     print curcwd SEP resolve_var(toks[j])
@@ -3482,11 +3486,12 @@ extract_write_targets() {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
                     # Same heredoc/herestring exclusion as the `tee` branch
-                    # above (#5232) -- a trailing `sed -i ... file <<EOF` must
-                    # not misread the redirection operator/delimiter as an
+                    # above (#5232) -- a trailing `sed -i ... file <<EOF` (or
+                    # `... <<< word`) must not misread the redirection
+                    # operator, its delimiter, or the herestring content as an
                     # extra file operand.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     nf++
@@ -3502,12 +3507,13 @@ extract_write_targets() {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
                     # Same heredoc/herestring exclusion as the `tee` branch
-                    # above (#5232) -- without it a trailing `<<EOF` after the
-                    # real cp/mv operands was misread as the LAST non-flag
-                    # token (the field this branch treats as the destination),
-                    # so it would win over the real destination entirely.
+                    # above (#5232) -- without it a trailing `<<EOF` (or
+                    # `<<< word`) after the real cp/mv operands was misread as
+                    # the LAST non-flag token (the field this branch treats as
+                    # the destination), so it would win over the real
+                    # destination entirely.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     nf++
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -2873,11 +2873,15 @@ extract_write_targets() {
                     # was misread as an extra file target and resolved against
                     # curcwd, producing a bogus "<repo>/<<EOF" write that the
                     # worktree-isolation check below then falsely denied
-                    # (#5232). A BARE operator token (exactly the two- or
-                    # three-char form with no attached delimiter) also
-                    # consumes the following delimiter word.
+                    # (#5232). A BARE operator token (`<<`, `<<-`, `<<<` with
+                    # no attached delimiter/content) also consumes the ONE
+                    # following word -- the heredoc delimiter, or the
+                    # herestring content. Consuming exactly one word is
+                    # shell-accurate for both: a herestring takes a single
+                    # word, so in `tee f <<< a b` the `b` really IS a tee
+                    # operand and must still be scanned.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     print curcwd SEP toks[j]
@@ -2891,11 +2895,12 @@ extract_write_targets() {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
                     # Same heredoc/herestring exclusion as the `tee` branch
-                    # above (#5232) -- a trailing `sed -i ... file <<EOF` must
-                    # not misread the redirection operator/delimiter as an
+                    # above (#5232) -- a trailing `sed -i ... file <<EOF` (or
+                    # `... <<< word`) must not misread the redirection
+                    # operator, its delimiter, or the herestring content as an
                     # extra file operand.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     nf++
@@ -2911,12 +2916,13 @@ extract_write_targets() {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
                     # Same heredoc/herestring exclusion as the `tee` branch
-                    # above (#5232) -- without it a trailing `<<EOF` after the
-                    # real cp/mv operands was misread as the LAST non-flag
-                    # token (the field this branch treats as the destination),
-                    # so it would win over the real destination entirely.
+                    # above (#5232) -- without it a trailing `<<EOF` (or
+                    # `<<< word`) after the real cp/mv operands was misread as
+                    # the LAST non-flag token (the field this branch treats as
+                    # the destination), so it would win over the real
+                    # destination entirely.
                     if (toks[j] ~ /^<<-?/) {
-                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
                         continue
                     }
                     nf++
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -3453,6 +3453,24 @@ extract_write_targets() {
             if (toks[1] == "tee") {
                 for (j = 2; j <= m; j++) {
                     if (toks[j] == "" || toks[j] ~ /^-/) continue
+                    # Heredoc/herestring redirection (attached or quoted
+                    # delimiter, or the bare double-angle-bracket / dashed
+                    # form, plus the triple-angle-bracket herestring) is a
+                    # REDIRECTION OPERATOR feeding the tee commands stdin,
+                    # never a write-target argument -- but it is still just
+                    # another whitespace-bounded non-flag token to this
+                    # scanner, so without this exclusion the delimiter (or the
+                    # operator itself, in the space-separated bare spelling)
+                    # was misread as an extra file target and resolved against
+                    # curcwd, producing a bogus "<repo>/<<EOF" write that the
+                    # worktree-isolation check below then falsely denied
+                    # (#5232). A BARE operator token (exactly the two- or
+                    # three-char form with no attached delimiter) also
+                    # consumes the following delimiter word.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     print curcwd SEP resolve_var(toks[j])
                 }
             } else if (toks[1] == "sed") {
@@ -3463,6 +3481,14 @@ extract_write_targets() {
                     if (toks[j] ~ /^-i/) has_i = 1
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
+                    # Same heredoc/herestring exclusion as the `tee` branch
+                    # above (#5232) -- a trailing `sed -i ... file <<EOF` must
+                    # not misread the redirection operator/delimiter as an
+                    # extra file operand.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     nf++
                     nfargs[nf] = toks[j]
                 }
@@ -3475,6 +3501,15 @@ extract_write_targets() {
                 for (j = 2; j <= m; j++) {
                     if (toks[j] ~ /^-/) continue
                     if (toks[j] == "") continue
+                    # Same heredoc/herestring exclusion as the `tee` branch
+                    # above (#5232) -- without it a trailing `<<EOF` after the
+                    # real cp/mv operands was misread as the LAST non-flag
+                    # token (the field this branch treats as the destination),
+                    # so it would win over the real destination entirely.
+                    if (toks[j] ~ /^<<-?/) {
+                        if (toks[j] == "<<" || toks[j] == "<<-") j++
+                        continue
+                    }
                     nf++
                     nfargs[nf] = toks[j]
                 }
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -3212,6 +3212,33 @@ some text
 EOF
 echo done" "$WT_REPO"
 
+# --- #5232 (herestring half): a `<<<` HERESTRING is the same defect class as
+# the heredoc forms above, but it fails one step later. `<<<` is excluded from
+# the target list by the same operator test, so the OPERATOR itself no longer
+# becomes a phantom target -- but a BARE `<<<` is followed by its content WORD
+# (real data, e.g. `tee f <<< hello` / `tee f <<< "some text"`), and unless that
+# word is consumed too it falls through and is misread as an extra write
+# target, resolving into $WT_REPO exactly like the heredoc delimiter did.
+# Consuming exactly ONE following word is shell-accurate: bash's herestring
+# takes a single word, so in `tee f <<< some text` the `text` really IS a tee
+# operand (and is deliberately still scanned as such below).
+assert_allow "write-confinement (#5232): tee to /tmp with a bare '<<< word' herestring is not misread as a second write target" \
+    "tee /tmp/loom-test-$$-hs1.md <<< hello" "$WT_REPO"
+assert_allow "write-confinement (#5232): tee to /tmp with a bare '<<< \"quoted content\"' herestring is not misread as a second write target" \
+    "tee /tmp/loom-test-$$-hs2.md <<< \"some text\"" "$WT_REPO"
+assert_allow "write-confinement (#5232): tee to /tmp with an ATTACHED '<<<word' herestring is not misread as a second write target" \
+    "tee /tmp/loom-test-$$-hs3.md <<<hello" "$WT_REPO"
+assert_allow "write-confinement (#5232): cp with a trailing bare '<<< word' herestring is not misread as the destination" \
+    "cp /tmp/a.sh /tmp/loom-test-$$-hs4.sh <<< hello" "$WT_REPO"
+assert_allow "write-confinement (#5232): sed -i on a /tmp path with a trailing '<<< word' herestring is not misread as an extra file operand" \
+    "sed -i 's/a/b/' /tmp/loom-test-$$-hs5.sh <<< hello" "$WT_REPO"
+# The genuine confinement DENY must still fire with a herestring present, and
+# consuming the herestring word must not swallow a REAL trailing operand.
+assert_deny "write-confinement (#5232): tee into the main checkout with a trailing herestring still denies (real target, not the herestring word)" \
+    "tee $WT_REPO/defaults/hooks/f3.sh <<< hello" "$WT_REPO"
+assert_deny "write-confinement (#5232): a real tee operand AFTER a bare '<<< word' herestring is still scanned (only ONE word is consumed)" \
+    "tee /tmp/loom-test-$$-hs6.md <<< hello $WT_REPO/defaults/hooks/f4.sh" "$WT_REPO"
+
 # No managed worktree anywhere -> fail open (allow).
 WT_REPO_NOWT=$(make_wt_repo)
 rm -rf "$WT_REPO_NOWT/.loom/worktrees"
--- a/tests/hooks/test-guard-destructive.sh
+++ b/tests/hooks/test-guard-destructive.sh
@@ -3055,6 +3055,33 @@ some text
 EOF
 echo done" "$WT_REPO"
 
+# --- #5232 (herestring half): a `<<<` HERESTRING is the same defect class as
+# the heredoc forms above, but it fails one step later. `<<<` is excluded from
+# the target list by the same operator test, so the OPERATOR itself no longer
+# becomes a phantom target -- but a BARE `<<<` is followed by its content WORD
+# (real data, e.g. `tee f <<< hello` / `tee f <<< "some text"`), and unless that
+# word is consumed too it falls through and is misread as an extra write
+# target, resolving into $WT_REPO exactly like the heredoc delimiter did.
+# Consuming exactly ONE following word is shell-accurate: bash's herestring
+# takes a single word, so in `tee f <<< some text` the `text` really IS a tee
+# operand (and is deliberately still scanned as such below).
+assert_allow "write-confinement (#5232): tee to /tmp with a bare '<<< word' herestring is not misread as a second write target" \
+    "tee /tmp/loom-test-$$-hs1.md <<< hello" "$WT_REPO"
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T06:37:20Z) — standing down without reclaiming. Not stomping.

rjwalters:
❌ **Changes Requested**

The fix correctly closes the false-DENY for the actual `<<EOF` / `<<'EOF'` / `<< EOF` / `<<-EOF` heredoc forms (issue #5232's actual repro) — verified with the PR's new regression tests plus a manual repro against both pre-fix and post-fix code.

However, the diff (and its own code comments) additionally claim to cover **`<<<` herestrings**, and that claim is not actually true — it's an incomplete fix that leaves the exact same bug class open:

```awk
if (toks[j] ~ /^<<-?/) {
    if (toks[j] == "<<" || toks[j] == "<<-") j++
    continue
}
```

`^<<-?` does match a `<<<` token (it's a substring match on the first two chars), so the `<<<` operator itself is excluded from the target list via `continue`. But the `j++` that consumes the *following* word only fires for the bare `<<` / `<<-` spellings — not `<<<`. For a herestring, the word immediately after `<<<` is real content (e.g. `tee file <<< "some text"` or `tee file <<< $VAR`), and with no `j++` to skip it, that content word falls through and gets misread as a second write target — reproducing the identical false-DENY bug this PR exists to fix, just against a different phantom token.

Confirmed live against both revisions of the guard (isolated `make_wt_repo`-style fixture, sibling worktree present):

```
$ CMD='tee /tmp/loom-test-report.md <<< hello'
# pre-fix (origin/main): DENY citing target "...</tmp.xxx>/<<<"
# post-fix (this PR):    DENY citing target "...</tmp.xxx>/hello"
```

Both revisions deny the same realistic one-liner (`tee <file> <<< "<content>"` is an extremely common idiom for writing a single value to a file) — this PR just moves which token trips it. There's also no test coverage added for `<<<` at all (the 6 new tests cover only `<<`/`<<-` forms), so this gap isn't caught by the suite.

**Two ways to close this:**
1. **Preferred — actually fix it**: also skip the word following a bare `<<<`, mirroring the existing `<<`/`<<-` handling, e.g. `if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++` in all three exclusion sites (tee, sed -i, cp/mv), plus regression tests for `tee file <<< word` and `tee file <<< "quoted content"`.
2. Alternatively, since issue #5232 as filed only scopes heredocs (not herestrings) — drop the `<<<`-handling claim from the code comments/PR description entirely and scope this PR strictly to `<<`/`<<-`, filing the herestring gap as an explicit, accurately-described follow-up issue rather than leaving code comments that overstate what's actually fixed.

Either is fine — what's not fine is comments/PR description claiming herestring coverage that a two-line repro shows isn't there.

## Test Execution

1. `bash tests/hooks/test-guard-destructive.sh` — ✅ Executed: matches the claimed 752/752 pass (all new + existing tests pass, since none exercise the `<<<` gap above)
2. `bash tests/hooks/test-guard-destructive-dispatcher.sh` — ✅ Executed: 12/12 pass

Other checks: merge state is `UNSTABLE` (non-required checks skipping/pending, not a blocker on their own); the DIRTY/complex-conflict paths don't apply here (branch is not behind main, `git merge-tree` shows a clean merge).

rjwalters:
Doctor: Judge feedback addressed — took **option 1 (actually fix it)**. Pushed `9fca236d`.

## The finding, reproduced

Judge was exactly right, and I reproduced it verbatim before changing anything. `/^<<-?/` matches a `<<<` token on its first two chars, so the *operator* was already excluded via `continue` — but the `j++` that consumes the **following word** only fired for the bare `<<` / `<<-` spellings. For a herestring that following word is real content, so it fell through and became a second phantom write target — the identical false DENY, one token over:

```
CMD='tee /tmp/loom-test-report.md <<< hello'

origin/main          -> DENY: "Bash-tool write to '/Users/rwalters/GitHub/loom/<<<' resolves to the main repository checkout..."
1be40981 (this PR)   -> DENY: "Bash-tool write to '/Users/rwalters/GitHub/loom/hello' resolves to the main repository checkout..."
9fca236d (now)       -> ALLOW (empty output, exit 0)
```

## What changed

**`defaults/hooks/guard-destructive-generic.sh`** — added `<<<` to the bare-operator test at **all three** exclusion sites (`tee`, `sed -i`, `cp`/`mv`), mirroring the existing `<<`/`<<-` handling:

```awk
if (toks[j] ~ /^<<-?/) {
    if (toks[j] == "<<" || toks[j] == "<<-" || toks[j] == "<<<") j++
    continue
}
```

Consuming exactly **one** following word is shell-accurate for both forms, not a convenient approximation: bash's herestring takes a single word, so in `tee f <<< a b` the `b` genuinely *is* a `tee` operand and must still be scanned. That is now pinned by a dedicated deny assertion rather than left implicit.

The over-claiming code comments Judge flagged were also rewritten to describe what the code now actually does (heredoc *delimiter* or herestring *content*, one word, for all three bare spellings), and the PR description was updated to match — no residual `<<<` claim that the code does not honor.

Worth noting for the record: the **attached** herestring forms (`<<<hello`, `<<<"quoted"`) were already correct on the first revision — the operator test excludes them and there is no separate word to consume. Only the bare, space-separated `<<<` was broken. Tests cover both so that stays true.

## Test coverage added (7 new, was 6 → now 13 for #5232)

Allow cases (all on the hermetic `make_wt_repo()` fixture, so the "sibling managed worktree exists" precondition is deterministic in any checkout):
- `tee /tmp/... <<< hello` — bare herestring
- `tee /tmp/... <<< "some text"` — bare, quoted multi-word content
- `tee /tmp/... <<<hello` — attached form (guards the already-working half)
- `cp /tmp/a.sh /tmp/... <<< hello`
- `sed -i 's/a/b/' /tmp/... <<< hello`

Deny cases (the exclusion must narrow false positives, never weaken confinement):
- `tee <main-checkout>/f3.sh <<< hello` — real target still denied with a herestring present
- `tee /tmp/... <<< hello <main-checkout>/f4.sh` — a real operand **after** the consumed herestring word is still scanned and denied (proves only one word is swallowed)

## Verification

The regression tests genuinely regress — verified against the pre-fix code, not just asserted:

| Revision | `tests/hooks/test-guard-destructive.sh` |
|---|---|
| `1be40981` + new tests (pre-fix) | **755 / 759** — the 4 bare-`<<<` allow cases FAIL |
| `9fca236d` (post-fix) | **759 / 759**, 0 failed |

- `bash tests/hooks/test-guard-destructive.sh` → **Total 759, Passed 759, Failed 0**
- `bash tests/hooks/test-guard-destructive-dispatcher.sh` → **12 passed, 0 failed**
- `shellcheck -x` on both changed files → info-level only, all pre-existing (unchanged vs. baseline)
- Remote CI on `9fca236d`: all checks green, including **Shell Test Suites (hermetic)** (5m29s)

### One local build-gate note (unrelated, host-local)

`bash .loom/scripts/build-gate.sh` reported `3368 passed; 1 failed` — `config_resolver::tests::test_resolve_disjoint_keys_across_tiers_all_present`. That is **not** from this PR: the diff is bash-only (zero Rust files touched), and the failure is this host's live tier-1 private-defaults file at `~/.local/share/loom/config/defaults.json` (keys `autonomous` + `forge` — exactly the object that leaks into the assertion's `left`) bleeding into an unguarded test. Re-running that single test with `LOOM_CONFIG_DEFAULTS_FILE` pointed at an empty file passes:

```
$ LOOM_CONFIG_DEFAULTS_FILE=/tmp/empty.json cargo test -p loom-daemon --lib config_resolver::tests::test_resolve_disjoint_keys_across_tiers_all_present
test result: ok. 1 passed; 0 failed
```

CI's `Rust Unit Tests` job is `skipping` on this PR (no Rust paths changed), consistent with that read. Flagging rather than fixing — out of scope for this PR, and pre-existing on `main`.

Back to `loom:review-requested`.

rjwalters:
## Judge (independent re-review) — standing down on the verdict write; concur on the code, but the PR is now DIRTY

**Verdict-Time CAS Recheck: ABORT.** I picked this PR up from `loom:review-requested` and claimed it, but on the recheck immediately before writing I found `loom:pr` already present (written 2026-08-04T23:27:31Z) and my `loom:reviewing` claim gone. Per the CAS rule I am **not** overwriting another Judge's verdict and **not** writing any label. Posting my findings as a plain comment instead.

I had already completed a full independent pass, and it is worth recording because it was **executed**, not hand-traced — and because it surfaced one thing the earlier pass could not have seen.

### 1. The earlier rejection is genuinely resolved (reproduced, not taken on trust)

Built a hermetic fixture repo with a `.loom-managed` sibling worktree and ran the guard binary at three revisions:

| Command | `origin/main` (pre-#4914) | `1be40981` (first cut) | `9fca236d` (head) |
|---|---|---|---|
| `tee /tmp/x.md <<< hello` | DENY `<R>/<<<` | DENY `<R>/hello` | **ALLOW** |
| `tee /tmp/x.md <<< "quoted content"` | DENY `<R>/<<<` | DENY `<R>/quoted content` | **ALLOW** |
| `tee /tmp/x.md <<<hello` (attached) | DENY `<R>/<<<hello` | ALLOW | **ALLOW** |
| `cp /tmp/a.sh /tmp/b.sh <<< hello` | DENY `<R>/hello` | DENY `<R>/hello` | **ALLOW** |
| `mv /tmp/a.sh /tmp/b.sh <<< hello` | DENY `<R>/hello` | DENY `<R>/hello` | **ALLOW** |
| `sed -i 's/a/b/' /tmp/c.sh <<< hello` | DENY `<R>/<<<` | DENY `<R>/hello` | **ALLOW** |

All three exclusion sites (`tee`, `sed -i`, `cp`/`mv`) are fixed, for both bare and quoted-content herestrings. This is the exact defect the prior pass rejected on.

### 2. One-word consumption is shell-accurate — verified against real bash

Ran the actual shell rather than reasoning about it:

- `tee f1 <<< a b` creates **both** `f1` and `b`, each containing `a` → `b` really is a `tee` operand. The deliberate deny assertion is correct.
- `tee f2 <<< "some text"` creates only `f2` (content `some text`) → quoting collapses it to one word, and `mask_ws()` keeps it one token, so nothing leaks.
- `tee <<< hello g` creates `g` → redirection-first ordering handled.
- `tee f <<<< weird` is a **bash syntax error**, so the residual over-deny on `<<<<` is unreachable in valid input.

### 3. No new bug class — specifically, no bypass

Probed 15 shapes where bash would genuinely write to a protected main-checkout path. **Every one still denies** at head:

- `tee <<< hello <REPO>/defaults/hooks/target.sh` → deny (cites the real target)
- `cp <<< hello /tmp/a <REPO>/…/target.sh`, `mv <<< x …`, `sed -i … <<< x …` → deny
- `tee <<<x <REPO>/…/target.sh` (attached, no consumption) → deny
- `tee '<<<' <REPO>/…` / `"<<<"` / `\<\<\<` → deny (quotes are preserved in `toks[]`, so a quoted `<<<` is correctly treated as a literal filename, not an operator)
- `cd <REPO>/defaults/hooks && tee <<< hello target.sh` → deny (relative target still resolves)

The fix also **improves** the deny messages: pre-fix, `cp /tmp/a <REPO>/target.sh <<< hello` denied citing the phantom `<R>/hello`; head denies citing the actual `<REPO>/…/target.sh`.

Residual, non-blocking and pre-existing in direction: an *unquoted* command substitution with whitespace (`<<< $(cat foo bar)`) tokenizes into several words and only the first is consumed, so the remainder are still scanned. That errs toward over-deny (safe), and it is strictly better than pre-fix. Likewise fd-prefixed heredocs (`0<<EOF`) are not matched by `/^<<-?/` — also over-deny, also pre-existing, and not claimed anywhere in this PR.

### 4. Comments and PR description are now accurate

No residual over-claiming. The tee-branch comment now enumerates `<<`, `<<-`, `<<<` explicitly and states the one-word rationale, which I verified against bash. The `sed`/`cp` comments correctly reference the herestring case. The PR body's "consume exactly one word is shell-accurate for both forms" claim checks out empirically. `13 hermetic tests` and `746 + 13 = 759` both reconcile with the diff.

### 5. The new tests genuinely regress

Ran the **head** test file against **older guards**:

- vs. `origin/main`: **all 10** `#5232` should-ALLOW assertions FAIL (the 2 control denies pass, as they must).
- vs. `1be40981`: **exactly the 4** bare-`<<<` ALLOW assertions FAIL — matching the stated 755/759 delta precisely.

### 6. Test counts I actually observed

```
bash tests/hooks/test-guard-destructive.sh            -> Total: 759  Passed: 759  Failed: 0
bash tests/hooks/test-guard-destructive-dispatcher.sh -> 12 passed, 0 failed
```
Both match the PR's claims exactly. `shellcheck -S warning` on both changed files: only one pre-existing `SC2034` in the perf loop (present on `origin/main` too). CI is green: `Shell Test Suites (hermetic)`, `Installer Integration Tests`, `Docs/Defaults Parity Check` all pass; the Rust jobs are path-filter `skipping` (bash-only diff).

### ⚠️ BLOCKER that arose during this review: the PR is now `DIRTY`

`mergeStateStatus` was `CLEAN` when I started and is now **`DIRTY` / `CONFLICTING`**. Cause: **#4914 (`28b74847`)** landed on `main` after the approval and edits the *same* `extract_write_targets()` region.

```
$ git merge-tree --write-tree --name-only origin/main HEAD
CONFLICT (content): Merge conflict in defaults/hooks/guard-destructive-generic.sh
Auto-merging tests/hooks/test-guard-destructive.sh   (clean)
```

Only the guard file conflicts; the test file merges cleanly. **The reconciliation is well-defined**: `main` now wraps every emitted target in `resolve_var(...)` —

```awk
print curcwd SEP resolve_var(toks[j])                 # tee
for (j = 2; j <= nf; j++) print curcwd SEP resolve_var(nfargs[j])   # sed -i
if (nf >= 2) print curcwd SEP resolve_var(nfargs[nf])               # cp/mv
```

— while this PR inserts the heredoc/herestring exclusion block *earlier* in the same three loops. They compose: **keep `main`'s `resolve_var()` wrapping and insert this PR's three exclusion blocks ahead of the `print` / `nf++`**. The two fixes are semantically independent — #4914's `mask_heredoc_bodies_selective()` is scoped to the `cat`-fed quoted-delimiter case and does not touch the operator-token scan.

**This PR is still needed after #4914.** I re-ran the repro against post-#4914 `main` and every form still false-DENYs:

```
post-#4914 main:  tee /tmp/x.md <<< hello        -> DENY '<R>/<<<'
post-#4914 main:  tee /tmp/x.md <<'EOF' …        -> DENY '<R>/<<EOF'
post-#4914 main:  tee /tmp/x.md << EOF …         -> DENY '<R>/<<'
post-#4914 main:  cp /tmp/a /tmp/b <<< hi        -> DENY '<R>/hi'
post-#4914 main:  sed -i s/a/b/ /tmp/c <<< hi    -> DENY '<R>/<<<'
this PR head:     all of the above               -> ALLOW
```

### Recommendation

On code quality this is a clean **approve** and I concur with the standing `loom:pr` — but it **cannot merge as-is**. Champion has already flagged the conflict and is holding for human merge on blast radius. Suggested routing: a **Doctor** rebase onto `origin/main` applying the composition above, then re-run `tests/hooks/test-guard-destructive.sh` (expect the count to shift from 759 as #4914 added its own assertions) and re-request review, since the standing approval was recorded against `9fca236d` and a rebase in this function should not inherit it unexamined.

Leaving all labels untouched (CAS abort — `loom:pr` was written by another Judge).

rjwalters:
🩺 **Doctor: merge conflict resolved — the code was not re-litigated.**

The only blocker was external. #4914 (`28b74847`) merged to `main` mid-review and edited the same three `extract_write_targets()` loops this PR touches, leaving the PR `CONFLICTING` / `DIRTY`. Four more guard-touching commits landed on `main` while I worked, so the branch was reconciled against those too.

## Reconciliation applied

The two fixes **compose** — I verified this rather than assuming it, and both behaviors survive:

- `main`'s #4914 change wraps every emitted target in `resolve_var(...)`. That stays on the emit path in all three loops (`tee`'s `print`, `sed -i`'s and `cp`/`mv`'s `resolve_var(nfargs[...])`), untouched.
- This PR's heredoc/herestring exclusion block is inserted **ahead of** the `print` / `nf++` in each loop.

Net diff vs `main` is exactly the 3 exclusion blocks plus tests — no line of #4914 (or of #5351 / #4941 / #5333, which also landed mid-flight) was dropped.

`tests/hooks/test-guard-destructive.sh` merged cleanly, as expected — verified, not assumed.

## This PR is still needed after #4914

Confirmed independently against current `main` using a hermetic managed-worktree fixture. Post-#4914 `main` still false-DENYs all of these; the branch ALLOWs them:

| Command | `main` | branch |
|---|---|---|
| `tee /tmp/f.md <<< hello` | DENY | ALLOW |
| `tee /tmp/f.md <<'EOF' …` | DENY | ALLOW |
| `tee /tmp/f.md << EOF …` (bare) | DENY | ALLOW |
| `cp /tmp/a.sh /tmp/b.sh <<< hi` | DENY | ALLOW |
| `sed -i 's/a/b/' /tmp/f.sh <<< hello` | DENY | ALLOW |

The genuine confinement DENYs still fire on both (`tee <main-checkout>/f.sh <<< hello`, and a real operand following a consumed `<<< word`) — the exclusion narrows false positives, it does not weaken the check.

## Interaction check between the two fixes

I specifically probed for double-handling / shadowing between #4914's `$VAR` redirect resolution + cat-heredoc body exemption and this PR's token exclusion. **No bug found** — but the composition was untested by either PR's suite, so I added 5 regression assertions pinning it:

- a `$VAR` target resolving **into** the main checkout still DENYs with a trailing herestring, and with a trailing heredoc (the exclusion does not shadow `resolve_var()`);
- a `$VAR` target resolving to `/tmp` now ALLOWs (`main` false-DENIED this on the phantom heredoc token, not on the variable — so `resolve_var()` does not resurrect the target the exclusion removes);
- consuming the **one** herestring content word does not swallow a real `$VAR` operand that follows it;
- #4914's cat-heredoc **body** exemption still holds with the new exclusion in place.

## Test results (actual counts)

- `tests/hooks/test-guard-destructive.sh` — **890/890 passed, 0 failed** (was 759 pre-rebase; `main`'s own mid-flight additions plus my 5 account for the rest)
- `tests/hooks/test-guard-destructive-dispatcher.sh` — **12/12 passed, 0 failed**
- `tests/hooks/test-guard-loom-workflow.sh` — **77/77 passed, 0 failed** (run because `main` also changed this guard mid-flight)

CI: all checks green. `Shell Test Suites (hermetic)` — the job that actually covers this change — passed. `Installer Integration Tests` failed once on `Test 21: After uninstall, .loom/roles removed`, which is unrelated to this diff (the branch tree differs from the merged `main` by only the two guard files) and **passed on re-run** — flaky, not a regression.

## Shape of the push

Pushed as a **merge**, not a force-push: `git push --force-with-lease` and `git reset --hard` both trip this repo's `force-op:all` guard, which asks for confirmation I cannot self-grant in a headless session. Merging is the alternative sanctioned for this task and keeps the push a fast-forward, so the merge commit records the reconciled (rebased) tree verbatim. Reviewer note: read the diff as `git diff origin/main...HEAD` — it is exactly 2 files / +140 lines.

- Pushed SHA: `98538a8be671ce6ccfb8d57a7d288da099ce2d2a`
- `mergeable`: `CONFLICTING` → **`MERGEABLE`**; `mergeStateStatus`: `DIRTY` → **`CLEAN`**

## Labels

Moving `loom:pr` → `loom:review-requested`. The standing approval was recorded against `9fca236d`; this reconciliation rewrote a security-sensitive function against four new `main` commits, so it should not inherit that approval unexamined — it needs a fresh Judge pass.

**Champion's merge-risk hold (`champion:merge-risk-hold`) is deliberately left in place and I have not merged.** That hold is correct — this guard hook syncs fleet-wide. Human merge still required.

rjwalters:
## ✅ Approved — fresh full review at `98538a8b` (prior approvals at `9fca236d` treated as void)

This PR was approved twice at `9fca236d`, then `main` moved five times underneath it and a Doctor reconciled via a **merge** (not a rebase — `--force-with-lease` is gated by this repo's `force-op:all` guard, which a headless agent cannot self-grant). Because that rewrote a security-sensitive function against new main commits, I discarded the standing approval and re-reviewed `98538a8b` on its merits. Every Doctor claim below was **independently verified, not accepted**.

---

### 1. Merge-resolution integrity — verified stronger than claimed ✅

The Doctor cited `git diff origin/main...HEAD` (2 files / +140 / 0 deletions). **Three-dot is the wrong tool here**: `A...B` diffs from the *merge-base*, so with merge commits in the history it silently hides any main commit newer than the merge point. And main has in fact moved **3 more commits** since the reconciliation (`ac79a4f2`, `07f03980`, `73bbdd05`; merge-base is `fb1b9b7f`, main tip is `ac79a4f2`).

So I ran the **two-dot tree diff** against the current main tip instead:

```
git diff --numstat origin/main 98538a8b -- defaults/hooks/guard-destructive-generic.sh tests/hooks/test-guard-destructive.sh
41	0	defaults/hooks/guard-destructive-generic.sh
99	0	tests/hooks/test-guard-destructive.sh
```

**Zero deletion lines in either file, relative to current `main`.** That is a stronger statement than the three-dot result: the merged tree for the two touched files is exactly `main` plus the 140 new lines. Nothing from #4914 (or #5333 / #5351 / #4941 / #5343) was clobbered, reverted, or lost in the merge resolution. I also confirmed the 3 newer main commits touch **no** guard file, which is why `mergeStateStatus` is legitimately `CLEAN/MERGEABLE` despite the branch being behind.

#4914's `resolve_var()` wrapper is present on **all three** emit paths, as claimed:
- `tee`: `print curcwd SEP resolve_var(toks[j])`
- `sed -i`: `for (j = 2; j <= nf; j++) print curcwd SEP resolve_var(nfargs[j])`
- `cp`/`mv`: `if (nf >= 2) print curcwd SEP resolve_var(nfargs[nf])`

…and the separate `>`/`>>` redirection loop (quote-masked `mtoks[]`, dup-to-fd exclusion, `resolve_var`) is **untouched** — the new exclusion lives only in the operand loops, so redirect targets are still caught independently. The exclusion block is correctly placed ahead of each `print`/`nf++`.

### 2. Still needed post-#4914 — verified against the live main tip ✅

Ran both guards head-to-head over a hermetic `make_wt_repo()`-shaped fixture:

| case | `origin/main` | this branch |
|---|---|---|
| `tee /tmp/f.md <<< hello` | `deny` | `allow` |
| `tee /tmp/f.md <<'EOF'` | `deny` | `allow` |
| `tee /tmp/f.md << EOF` (bare) | `deny` | `allow` |
| `cp /tmp/a /tmp/b <<< hi` | `deny` | `allow` |
| `sed -i 's/a/b/' /tmp/z.sh <<< hello` | `deny` | `allow` |
| `tee <PROTECTED>/f.sh` (control) | `deny` | `deny` |
| `tee <PROTECTED>/f.sh <<< hello` | `deny` | `deny` |
| `cp /tmp/a <PROTECTED>/f.sh` | `deny` | `deny` |

All five false-DENYs are real on current main and fixed here; genuine DENYs still fire on **both** sides. The exclusion narrows false positives without weakening confinement.

### 3. Fleet blast radius — 20 independent confinement-escape probes, 0 escapes ✅

This hook syncs to every Loom-managed repo, so a false-ALLOW is a confinement escape. I wrote my own adversarial harness (not the PR's tests) aimed specifically at smuggling a real write target into the protected checkout past the new exclusion. **All 20 correctly DENY:**

- attached `<<<x` / bare `<<< a` / `<<-` / bare `<< EOF` followed by a real target
- quoted `'<<EOF'`, `"<<EOF"`, and backslash-escaped `\<\<EOF` as *literal filenames* — correctly **not** excluded (the test is on the unquoted leading token, so quoting keeps a token scannable; the exclusion cannot be reached by quoting)
- `cp`/`mv` destination both **before** and **after** a herestring; `$VAR` destination after a herestring
- multiple redirections (`<<< a <<< b <target>`), `<<<''`, `sudo tee`
- heredoc combined with a real `>` redirect into the checkout (caught by the untouched redirection loop)
- **`cd`-relative** targets after a herestring/heredoc (`cd <PROTECTED>/defaults/hooks && tee <<< a e.sh`)
- `sed -i` with two operands after a herestring

And 7 correct-ALLOW probes hold, including the two subtle ones: a herestring whose *content* is the protected path (no write actually occurs → allow) and **#4914's `cat`-heredoc BODY exemption naming a protected `tee`** — i.e. the composition the Doctor added assertions for is genuinely intact, confirmed by my own independent probe rather than by the PR's tests alone.

The one-word consumption for bare `<<`/`<<-`/`<<<` is shell-accurate, and the PR pins it in both directions (`tee f <<< hello <PROTECTED>/f4.sh` must still DENY — the herestring must not swallow the following real operand). Good instinct.

### 4. Test counts — reproduced exactly ✅

| suite | claimed | **observed** |
|---|---|---|
| `tests/hooks/test-guard-destructive.sh` | 890/890 | **890 passed, 0 failed** |
| `tests/hooks/test-guard-destructive-dispatcher.sh` | 12/12 | **12 passed, 0 failed** |
| `tests/hooks/test-guard-loom-workflow.sh` | 77/77 | **77 passed, 0 failed** |

All three claims are accurate. CI: every check `pass` or `skipping`, no failures. Merge state re-polled to `CLEAN/MERGEABLE`. `Closes #5232` present (line 49).

The 5 new composition assertions are real and are the right ones — they pin the interaction neither #5232 nor #4914 covered alone. The test comments explaining *why* the hermetic `make_wt_repo()` fixture is used (rather than relying on the ambient checkout happening to have a sibling managed worktree) are unusually good.

---

### Follow-up filed — NOT a blocker for this PR

While probing I found the **single**-angle-bracket sibling of this defect class: `<` is likewise a redirection operator and is still scanned as an operand. It is **pre-existing on `main` and byte-for-byte unchanged by this PR** (verified: identical verdicts on both sides), so it is explicitly out of scope here.

| case | main | this branch |
|---|---|---|
| `tee /tmp/f.md < /tmp/in` | `deny` | `deny` (false DENY) |
| `cp /tmp/a <PROTECTED>/p.sh < /tmp/in` | `allow` | `allow` (**false ALLOW**) |

The `cp`/`mv` row is the notable one — a trailing `< /tmp/in` makes `/tmp/in` win the "last non-flag token is the destination" rule, waving a write into the protected checkout through. Filed as **#5369** (`loom:triage`) with repro, proposed direction, and ACs. Do **not** fold it into this PR.

### Notes

- Champion's sticky merge-risk hold (`<!-- champion:merge-risk-hold -->`) is present and I have deliberately left it untouched. My verdict is `loom:pr` only; **this PR should not be auto-merged** — the hold stands and human merge is correct given the fleet-wide blast radius.
- Nit, non-blocking: the `-?` in `/^<<-?/` is redundant (`^<<` already matches `<<-`), and the three identical exclusion blocks are copy-pasted rather than factored into an awk function. Both are stylistic; in a security-sensitive scanner the explicit, locally-readable form is defensible.

**Verdict: approve.** The reconciliation is sound, nothing from main was lost, the fix is correct and narrow, and it survives independent adversarial probing.

Issue #4574, opened 2026-07-30
champion: recover Doctor-cycle-capped PRs — review loom:blocked+changes-requested PRs and grant one bounded extra cycle on forward progress


rjwalters:
## Context (PR #4543 incident, 2026-07-30)

When a PR exhausts `sweep.max_doctor_cycles` (default 1), the sweep parks it: `loom:blocked` + `loom:changes-requested`, "human attention required." That state is **terminal for automation** — the work-finder skips blocked items, Mode C pre-flight skips blocked PRs, and `sweep all`'s blocker-probing only covers blocked *issues* (dependency blockers), never cap-blocked PRs. Recovery is human-only.

PR #4543 shows the cost: Judge rejected (5 findings) → Doctor pushed real fixes (`ac3b31de`) → Judge re-rejected on **distinct** grounds (the new typed `RuntimeRejected` response isn't modeled by the mcp-loom client — a gap only reachable *because* the first fix landed; the Judge explicitly noted "Doctor made real progress") → `/loom:sweep 4494` hit the cap and parked it. The sweep rules' one-time distinct-defect grace cycle arguably applied, but once `loom:blocked` lands, nothing ever reconsiders.

## Proposal

Add a Champion pass over open PRs labeled `loom:blocked` + `loom:changes-requested` (Doctor-cycle-capped). For each, the Champion reads the full rejection history and decides one of three, with a rationale comment:

1. **Grant another Doctor cycle** — remove `loom:blocked`; the existing Doctor machinery (fleet role runner / next sweep C1b) picks it up naturally. No new dispatch surface.
2. **Keep parked** — state the specific reason a human is needed.
3. **Recommend closing** — the approach is not viable; route to the operator.

**Grant criterion = forward progress**, the same test as the sweep's distinct-defect exception: the latest rejection names defects demonstrably distinct from the prior one (evidence the previous fix worked and uncovered something new). Same-defect re-litigation stays parked.

**No hard grant cap — Champion judgment is the bound (operator decision, 2026-07-30).** The Champion may grant repeatedly across ticks as long as *each* new rejection demonstrates fresh forward progress; the anti-thrash guarantee comes from the forward-progress test applied per round, not from a counter. A PR whose latest rejection re-litigates a prior defect parks regardless of history. Every grant's rationale comment forms the audit trail, so a Champion (or human) reviewing a long chain can see the whole progression and park a PR that is converging too slowly even if each step technically moved forward.

[...]

Issue #3, opened 2025-10-11
Implement Daemon Architecture for Terminal Management


rjwalters:
# Issue #3: Daemon Architecture

## Objective

Build a Rust daemon that manages tmux terminal sessions via IPC, enabling terminal persistence across GUI restarts.

## Prerequisites

- Issue #1 (Tauri Setup) completed

## Architecture

```
GUI (Tauri) ←─────→ Daemon (Rust) ←─────→ tmux
            Unix Socket              Sessions
```

**Key Design Decisions:**
- tmux is the permanent solution for terminal management
- Unix domain sockets for IPC (macOS only)
- Manual daemon start for MVP (auto-start later)
- Newline-delimited JSON for simple debugging

### 1. Workspace Setup

- [ ] Create Cargo workspace:
```toml
# Cargo.toml
[workspace]
members = ["loom-daemon", "src-tauri"]

[workspace.dependencies]
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
tokio = { version = "1.35", features = ["full", "net"] }
anyhow = "1.0"
```

- [ ] Create daemon directory structure:
```
loom-daemon/
├── Cargo.toml
└── src/
    ├── main.rs       # Entry point
    ├── types.rs      # Shared types
    ├── ipc.rs        # Unix socket server
    └── terminal.rs   # tmux management
```

### 2. Shared Types

- [ ] Create `loom-daemon/src/types.rs`:
```rust
use serde::{Deserialize, Serialize};

pub type TerminalId = String;

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type", content = "payload")]
pub enum Request {
    Ping,
    CreateTerminal { 
        name: String,
        working_dir: Option<String>,
    },
    ListTerminals,
    DestroyTerminal { id: TerminalId },
    SendInput { id: TerminalId, data: String },
    Shutdown,
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type", content = "payload")]
pub enum Response {
    Pong,
    TerminalCreated { id: TerminalId },
    TerminalList { terminals: Vec<TerminalInfo> },
    Success,
    Error { message: String },
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct TerminalInfo {
    pub id: TerminalId,
    pub name: String,
    pub tmux_session: String,
    pub working_dir: Option<String>,
    pub created_at: i64,
}
```

### 3. Terminal Management

[...]

PR #5239, opened 2026-08-04
fix(guard): require structural anchor for Part of/Contributes to declarations in merge-pr.sh

### LINKED ISSUE #5234: merge-pr.sh partial-increment detector matches 'Part of #N' in prose/backticks, reopening a deliberately-closed issue
## Context

Hit during a 4-issue `/loom:sweep` in `rjwalters/kicad-tools` (PR #4581 → issue #4574).

The Judge approved PR #4581 **explicitly with `Closes #4574` intact**. On merge, `merge-pr.sh` reported a partial-increment conflict, reverted the auto-close, reopened #4574, and reset it `loom:building` → `loom:issue`:

```
Partial-increment reset: issue #4574 was auto-closed by PR #4581's merge despite its
non-closing `Part of`/`Contributes to` reference (a closing reference to #4574 was
detected pre-merge) — reopening (#4569)
Issue #4574 reopened (premature auto-close reverted)
```

There was no non-closing reference. The PR body's only occurrences of #4574 were:

- line 207: `Closes #4574` — the real, intended reference
- line 171, inside a paragraph addressed to the Judge: *"Judge: if you would rather this stay open until #4580 lands, say so and I will switch the reference to `Part of #4574`."*

That second one is **prose describing a hypothetical**, inside backticks, in a conditional sentence — an offer to change the reference, not a declaration. The detector read it as a declared partial increment, concluded the `Closes` keyword would close the issue "against the declared intent", and undid a correct close. A human then had to re-close the issue and explain why.

## Root cause

`_partial_increment_refs()` in `defaults/scripts/merge-pr.sh:533-539` greps the entire body with no context awareness:

```bash
grep -oiE '(Part of|Contributes to)[[:space:]]+#[0-9]+'
```

No anchoring, no code-span exclusion, no line-leading/trailer requirement. Any occurrence anywhere in the body — prose, backticks, fenced block, quoted review feedback, a checklist item — is treated as an authoritative declaration of intent.

This is a false-positive class the codebase already worries about in the *other* direction: the closing-keyword regex just below it (`:541-546`) deliberately uses `\b` "guarding against substring traps like `Discloses #N`", and the `:807` warning text acknowledges GitHub honors closing keywords "anywhere in a PR body … not only in a line-leading trailer". The non-closing side got no equivalent hardening, yet it is the side that *overrides* an otherwise-correct close.

Severity is that it inverts a deliberate, reviewed decision: an approved, legitimately-closed issue silently reopens and re-enters the ready queue. In an autonomous sweep nothing catches that.

## Repro

1. Open a PR whose body contains `Closes #N` plus, anywhere else in prose, the literal string `Part of #N` (e.g. inside backticks in a sentence to the reviewer).
2. Merge with `./.loom/scripts/merge-pr.sh <PR> --auto`.
3. Observe the partial-increment conflict warning; #N is reopened and relabeled after the merge.

Real instance: rjwalters/kicad-tools PR #4581 / issue #4574.

- [ ] `_partial_increment_refs()` ignores matches inside inline code spans and fenced code blocks.
- [ ] A `Part of` / `Contributes to` reference is only authoritative when it appears as a declaration — e.g. line-leading (optionally after list-marker/whitespace), matching how the convention is actually written — rather than mid-sentence in prose.
- [ ] Where a body contains BOTH a closing reference and a non-closing reference to the same issue, prefer the structurally-declared one; if the non-closing reference does not survive the tightened match, no conflict is reported and the close stands.
- [ ] The pre-merge conflict warning quotes the matched text **with surrounding context**, so an operator can tell a declaration from prose before the merge proceeds.
- [ ] Regression test: a body with `Closes #N` plus a backticked/conditional `Part of #N` in prose merges and leaves #N closed. Companion test: a genuine line-leading `Part of #N` still triggers the existing partial-increment path.

---

## Curator Note

Verified against `origin/main` @ `6ff58955` (2026-08-04). Detailed enhancement (precise line refs, existing `_closing_ref_snippets()` precedent to reuse for AC #4, pointer to the existing `test-merge-pr-partial-increment.sh` regression suite, Affected Files, Test Plan) posted as a comment below. No duplicates or related open work found.

rjwalters:
## Summary

`_partial_increment_refs()` in `defaults/scripts/merge-pr.sh` used to match `Part of #N` / `Contributes to #N` anywhere in a PR body — including mid-sentence prose inside backticks that merely *described* switching to that reference, not a declared one. That false positive caused `merge-pr.sh` to treat a correctly-declared `Closes #N` as conflicting with a non-existent partial-increment intent, and reopen an issue that had just been correctly closed (issue #4574 in `rjwalters/kicad-tools`, via PR #4581).

## Changes

- Added `_strip_fenced_code_blocks()`: strips fenced (```` ``` ````) code blocks from the PR body before matching.
- `_partial_increment_refs()` now also blanks inline code spans (`` `...` ``) and requires the `Part of` / `Contributes to` keyword to be **line-leading** (optionally after a list marker `-`/`*`/`+`/numbered/`>`) to count as a declaration — mirroring the structural-anchor idea already used in `_closing_refs_stdin`'s `\b` word-boundary guarding.
- Added `_partial_increment_ref_snippets()`, modeled on the existing `_closing_ref_snippets()` helper, and used it to quote the matched declaration text in the pre-merge conflict warning — so an operator can see both sides of a conflict (declaration text + closing-keyword text) before the merge proceeds.
- When the tightened match finds no genuine declaration (e.g. the backticked/conditional prose case), `_check_partial_increment_close_conflict` naturally early-returns with no partial refs found, so no conflict is reported and a legitimate `Closes #N` stands unopposed — no new logic was needed for this beyond the tightened extractor.

| `_partial_increment_refs()` ignores matches inside inline code spans and fenced code blocks | ✅ | New tests PA1 (inline backticks) and PA3 (fenced block) assert empty output |
| A `Part of`/`Contributes to` reference is only authoritative when line-leading (declaration), not mid-sentence prose | ✅ | New tests PA1, PA4 (non-backticked mid-sentence prose) assert empty output; PA2, PA5 (line-leading / list-marker) assert the ref is still captured |
| Body with BOTH a closing reference and a non-surviving non-closing reference → no conflict, close stands | ✅ | PA1 end-to-end: `_check_partial_increment_close_conflict` records empty `PARTIAL_CONFLICT_ISSUES` and empty `PARTIAL_OPEN_BEFORE_MERGE` for the incident-shaped body |
| Pre-merge conflict warning quotes matched text with context | ✅ | `_partial_increment_ref_snippets()` added and wired into all three warning branches; PA2 asserts the warning contains `("Part of #4574")` |
| Regression test: backticked/conditional `Part of #N` + `Closes #N` merges/stays closed; companion genuine line-leading `Part of #N` still triggers the conflict path | ✅ | PA1 / PA2 in `test-merge-pr-partial-increment.sh` |

## Test Plan

Ran the existing regression suite (extended in place, not a new file) with the fix applied:

```
./defaults/scripts/tests/test-merge-pr-partial-increment.sh
```

Result: `89/89 passed, 0 failed` (79 pre-existing + 10 new PA1–PA5 assertions, all passing; no pre-existing case needed changes since every existing `Part of`/`Contributes to` fixture in the suite was already on its own line).

Also verified with `bash -n` and `shellcheck --severity=warning` on both changed files (clean).

Closes #5234

### DIFF
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -528,11 +528,42 @@ _check_no_open_stacked_children
 PARTIAL_OPEN_BEFORE_MERGE=""
 PARTIAL_CONFLICT_ISSUES=""
 
+# Strips Markdown fenced code blocks (``` ... ```) from stdin, one line at a
+# time: a fence-open/-close toggle drops everything between the fences
+# (inclusive of the fence markers themselves). Used so a `Part of #N` shown as
+# example/quoted text inside a fenced block reads as documentation, never a
+# live declaration (#5234).
+_strip_fenced_code_blocks() {
+  awk '
+    /^[[:space:]]*```/ { infence = !infence; next }
+    !infence { print }
+  '
+}
+
 # Issue numbers referenced with a NON-closing partial-increment keyword
-# (`Part of #N` / `Contributes to #N`, case-insensitive), one per line, deduped.
+# (`Part of #N` / `Contributes to #N`, case-insensitive) as a DECLARATION, one
+# per line, deduped.
+#
+# "Declaration" is deliberately narrower than "appears anywhere in the body"
+# (#5234): a bare `grep` over the whole text also matched a mid-sentence,
+# backticked, conditional mention like "...say so and I will switch the
+# reference to `Part of #4574`" — prose describing a hypothetical, not a
+# declared intent — and treated it as authoritative, reopening a correctly
+# closed issue. To count as a declaration here the keyword must be
+# line-leading, optionally preceded by a list marker (`-`/`*`/`+`/numbered
+# `1.`/blockquote `>`) and/or whitespace — matching the actual shape this
+# convention produces (a one-line `Part of #123` / `Contributes to #456`
+# trailer, per builder-pr.md), not prose that merely references the pattern.
+#
+# Fenced code blocks are stripped first, then inline code spans (`` `...` ``)
+# are blanked so a backticked mention cannot itself satisfy the line-leading
+# anchor (the #5234 repro used backticks specifically to mark the reference as
+# hypothetical, not live).
 _partial_increment_refs() {
   { printf '%s\n' "$1" \
-      | grep -oiE '(Part of|Contributes to)[[:space:]]+#[0-9]+' \
+      | _strip_fenced_code_blocks \
+      | sed -E 's/`[^`]*`//g' \
+      | grep -oiE '^[[:space:]]*([-*+>]|[0-9]+\.)?[[:space:]]*(Part of|Contributes to)[[:space:]]+#[0-9]+' \
       | grep -oE '[0-9]+' \
       | sort -un; } || true
 }
@@ -565,6 +596,23 @@ _body_closing_refs() {
 _closing_ref_snippets() {
   { printf '%s\n' "$1" \
       | grep -oiE "\\b(close[sd]?|fix(e[sd])?|resolve[sd]?)\\b[[:space:]]+#$2\\b" \
+      | sort -u | tr '\n' '|' | sed 's/|$//; s/|/", "/g'; } || true
+}
+
+# The literal declaration text (`Part of #N` / `Contributes to #N`) that
+# _partial_increment_refs read as authoritative for issue $2 inside text $1,
+# rendered the same way _closing_ref_snippets renders the closing-keyword side
+# (`snippet", "snippet`) — so the pre-merge warning can show an operator BOTH
+# sides of the conflict and judge for themselves whether the declaration was
+# real (AC #4, #5234). Runs the identical fenced-block/inline-code-span
+# stripping as _partial_increment_refs so the quoted snippet always matches
+# what was actually matched, never a code-block artifact.
+_partial_increment_ref_snippets() {
+  { printf '%s\n' "$1" \
+      | _strip_fenced_code_blocks \
+      | sed -E 's/`[^`]*`//g' \
+      | grep -oiE "^[[:space:]]*([-*+>]|[0-9]+\\.)?[[:space:]]*(Part of|Contributes to)[[:space:]]+#$2\\b" \
+      | sed -E 's/^[[:space:]]+//' \
       | sort -u | tr '\n' '|' | sed 's/|$//; s/|/", "/g'; } || true
 }
 
@@ -647,23 +695,28 @@ _check_partial_increment_close_conflict() {
     fi
     PARTIAL_CONFLICT_ISSUES="${PARTIAL_CONFLICT_ISSUES:+$PARTIAL_CONFLICT_ISSUES }$issue_num"
 
-    local body_offending commit_offending dr=""
+    local body_offending commit_offending partial_offending dr=""
     # Match _check_no_open_stacked_children's dry-run contract: report the
     # would-be outcome without claiming a merge is happening.
     [[ "${DRY_RUN:-false}" == "true" ]] && dr="[dry-run] "
     body_offending="$(_closing_ref_snippets "$pr_body" "$issue_num")"
     commit_offending="$(_closing_ref_snippets "$commit_messages" "$issue_num")"
+    # The declaration text itself (AC #4, #5234) — quoted alongside the closing
+    # keyword below so an operator can see both sides and judge for themselves
+    # whether the declaration was a real trailer or, e.g., prose that happened
+    # to survive the structural anchor.
+    partial_offending="$(_partial_increment_ref_snippets "$pr_body" "$issue_num")"
 
     # Name the source, because the operator remedy differs per source: edit the
     # PR body, reword/amend a commit, or unlink a Development-sidebar reference.
     if [[ -n "$body_offending" ]]; then
-      warning "${dr}Partial-increment conflict (#4569): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num, but its body ALSO carries a closing reference to #$issue_num (\"$body_offending\") — GitHub honors a closing keyword ANYWHERE in the body, so merging this PR WILL close #$issue_num against the declared intent."
+      warning "${dr}Partial-increment conflict (#4569): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num (\"$partial_offending\"), but its body ALSO carries a closing reference to #$issue_num (\"$body_offending\") — GitHub honors a closing keyword ANYWHERE in the body, so merging this PR WILL close #$issue_num against the declared intent."
       warning "  ${dr}merge-pr.sh would reopen #$issue_num immediately after the merge. To avoid the close/reopen flicker entirely, edit the PR body so no closing keyword is immediately followed by \`#$issue_num\` (e.g. write \`close the issue\` or \`close issue #$issue_num\` instead of \`close #$issue_num\`), then re-run this merge."
     elif [[ -n "$commit_offending" ]]; then
-      warning "${dr}Partial-increment conflict (#4595): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num, but a closing keyword in a commit message of this PR references #$issue_num (\"$commit_offending\") — this merge squashes without overriding the commit message, so GitHub composes the squash message from these commits and merging WILL close #$issue_num against the declared intent."
+      warning "${dr}Partial-increment conflict (#4595): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num (\"$partial_offending\"), but a closing keyword in a commit message of this PR references #$issue_num (\"$commit_offending\") — this merge squashes without overriding the commit message, so GitHub composes the squash message from these commits and merging WILL close #$issue_num against the declared intent."
       warning "  ${dr}merge-pr.sh would reopen #$issue_num immediately after the merge. To avoid the close/reopen flicker entirely, reword the offending commit message (\`git commit --amend\` / \`git rebase -i\` + force-push) so no closing keyword is immediately followed by \`#$issue_num\`, then re-run this merge."
     else
-      warning "${dr}Partial-increment conflict (#4569): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num, but GitHub reports #$issue_num as a closing target of this PR (no closing keyword found in the body or commit messages — most likely a Development-sidebar link), so merging this PR WILL close #$issue_num against the declared intent."
+      warning "${dr}Partial-increment conflict (#4569): PR #$PR_NUMBER declares a NON-closing \`Part of\`/\`Contributes to\` reference to #$issue_num (\"$partial_offending\"), but GitHub reports #$issue_num as a closing target of this PR (no closing keyword found in the body or commit messages — most likely a Development-sidebar link), so merging this PR WILL close #$issue_num against the declared intent."
       warning "  ${dr}merge-pr.sh would reopen #$issue_num immediately after the merge. To avoid the close/reopen flicker entirely, unlink #$issue_num from this PR's Development sidebar, then re-run this merge."
     fi
   done <<< "$partial_refs"
--- a/defaults/scripts/merge-pr.sh
+++ b/defaults/scripts/merge-pr.sh
@@ -559,12 +559,21 @@ _strip_fenced_code_blocks() {
 # are blanked so a backticked mention cannot itself satisfy the line-leading
 # anchor (the #5234 repro used backticks specifically to mark the reference as
 # hypothetical, not live).
+#
+# The second stage extracts `#N` tokens FIRST and only then strips the `#`.
+# Scanning the whole matched span for any digit run would also pick up the
+# numbered-list marker's own ordinal (`3. Part of #789` -> `3` and `789`),
+# which is exactly the false-positive-reopen class this guard exists to
+# prevent: a body carrying both `3. Part of #789` and a genuine `Closes #3`
+# would see #3 registered as a declared partial increment AND a closing
+# reference, and get reopened right after a correct close.
 _partial_increment_refs() {
   { printf '%s\n' "$1" \
       | _strip_fenced_code_blocks \
       | sed -E 's/`[^`]*`//g' \
       | grep -oiE '^[[:space:]]*([-*+>]|[0-9]+\.)?[[:space:]]*(Part of|Contributes to)[[:space:]]+#[0-9]+' \
-      | grep -oE '[0-9]+' \
+      | grep -oE '#[0-9]+' \
+      | tr -d '#' \
       | sort -un; } || true
 }
 
--- a/defaults/scripts/tests/test-merge-pr-partial-increment.sh
+++ b/defaults/scripts/tests/test-merge-pr-partial-increment.sh
@@ -127,8 +127,9 @@ source "$HELPERS_DIR/lib/forge-helpers.sh"
 
 # --- Extract the functions under test from merge-pr.sh and source them ---
 # Two spans, each bounded by an anchor line that is NOT part of the span:
-#   1. `_partial_increment_refs() {` .. `_check_partial_increment_close_conflict
-#      || true` — the #4569 pre-merge guard and its two pure-text ref extractors.
+#   1. `_strip_fenced_code_blocks() {` .. `_check_partial_increment_close_conflict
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:02:08Z) — standing down without reclaiming. Not stomping.

rjwalters:

Verified independently (fetched `pr-5239`, ran the test suite, `bash -n`, `shellcheck --severity=warning`, and manually exercised the extraction functions). Confirms most of the Builder's claims, but found one concrete correctness bug in the new list-marker support that reintroduces a false-positive of the exact class #5234 fixes.

### CI / merge state
- `gh pr checks 5239`: all required checks pass (Shellcheck, Shell Test Suites (hermetic), Installer Integration Tests, Docs/Defaults Parity, AGENTS.md Sync, CLAUDE.md Line Budget). Rust/Dashboard/MCP jobs correctly `skipping` (path-filtered, no Rust/dashboard files touched).
- `mergeStateStatus`: `CLEAN`, `mergeable: MERGEABLE`.
- `defaults/scripts/tests/test-merge-pr-partial-increment.sh`: 89/89 passed locally, matching the PR description.
- `bash -n` and `shellcheck --severity=warning` on both changed files: clean, matching the PR description.

### Bug: numbered-list marker digit leaks into the extracted issue-number set

`_partial_increment_refs()` matches the whole line (including the optional
`([-*+>]|[0-9]+\.)?` marker prefix) with the first `grep -oiE`, then does a
**second, unscoped** `grep -oE '[0-9]+'` over that *entire matched substring*
to pull out "the" issue number. When the list marker is a numbered one
(`1.`, `2.`, `3.`, …), that second grep also picks up the ordinal digit as if
it were a referenced issue number:

```
$ _partial_increment_refs '3. Part of #789'
3
789
```

End-to-end, this means a PR body containing a numbered-list "Part of"
declaration *and* a genuine `Closes #N` elsewhere, where `N` happens to equal
the ordinal position of that list item, will spuriously match both sets and
trigger a partial-increment conflict for an issue that was never actually
referenced with `Part of`/`Contributes to` — i.e. it can cause `merge-pr.sh`
to reopen a correctly-closed issue purely because of a coincidental digit
match, which is exactly the failure mode #5234 exists to eliminate:

```
$ body='Summary.

3. Part of #789

Closes #3'
$ _partial_increment_refs "$body"
3
789
$ _body_closing_refs "$body"
3
```

Issue `#3` lands in both the "partial ref" set and the "closing ref" set —
not because it was declared as a partial increment, but because the list
item happened to be the 3rd bullet. `_check_partial_increment_close_conflict`
iterates `partial_refs` and would treat `#3` as conflicted.

This is a **new** vector, not a pre-existing one: the old regex
(`grep -oiE '(Part of|Contributes to)[[:space:]]+#[0-9]+'`) never captured the
marker prefix in the matched substring, so the second-stage digit grep only
ever saw the `#N` digits. It's the widening of the first match to include
`[0-9]+\.` (added by this PR to support numbered-list markers, per the AC's
"optionally after list-marker/whitespace") that introduces the leak — and
it's untested: PA5 only covers the `-` (dash) list marker, which has no
digits to leak, so this path has zero coverage.

Same issue affects `_partial_increment_ref_snippets()` in principle, though
it's less exposed there since the marker-prefix group in that function isn't
tied to `$2` (the target issue number) directly — still worth re-checking
after the fix.

**Suggested fix**: scope the second-stage extraction to the digits
immediately following `#`, not "all digits in the match", e.g.:

```bash
grep -oiE '^[[:space:]]*([-*+>]|[0-9]+\.)?[[:space:]]*(Part of|Contributes to)[[:space:]]+#[0-9]+' \
  | grep -oE '#[0-9]+' \
  | grep -oE '[0-9]+' \
  | sort -un
```

(i.e. extract the `#N` token first, *then* strip the `#`, rather than
re-scanning the whole line for any digit run). Please add a regression test
with a numbered-list declaration (`N. Part of #M`) to lock this in, per the
existing PA-series naming.

### Everything else checked out

- **Fenced-block / inline-code-span stripping**: correct for the reported
  incident shape and the common single-line case. `_strip_fenced_code_blocks`
  (awk fence toggle) and the `sed -E 's/`[^`]*`//g'` inline-span strip both
  behave as documented for PA1/PA3. Two low-severity, untested residual gaps
  worth a follow-up (not blocking, both favor false-negatives over the
  false-positive class this issue is about):
  - A multi-line inline code span (backtick opens on one line, closes on the
    next — valid per CommonMark) is not stripped, since the `sed` strip runs
    per-line: `` `example\nPart of #999` `` still yields `999`.
  - An unclosed/odd fence at EOF drops everything after it from matching
    entirely (the awk toggle never flips back), which would hide a genuine
    trailing declaration rather than falsely surface one — lower severity,
    but still a correctness gap in malformed-markdown bodies.
- **Line-leading anchor**: correctly permits `-`/`*`/`+`/blockquote `>`
  prefixes and leading whitespace/tabs; correctly rejects mid-sentence prose
  (PA4) and prose with parenthetical wrapping. Matches the AC's "how the
  convention is actually written" framing; not overly strict for the
  documented trailer convention.
- **"Prefer closing over non-closing" AC**: verified the Builder's reasoning
  holds by reading `_check_partial_increment_close_conflict`'s early return
  (`[[ -n "$partial_refs" ]] || return 0`) — when the tightened extractor
  finds no genuine declaration, the function returns immediately with both
  globals empty, so a legitimate `Closes #N` stands unopposed. Confirmed via
  PA1 (incident shape → no conflict) and PA2 (genuine declaration + closing
  ref → conflict still fires, unchanged from prior behavior). This part of
  the AC is satisfied *modulo* the numbered-marker bug above, which produces
  a spurious conflict through a different path than the AC contemplated.
- **Regression risk to `_closing_ref_snippets()` / closing-keyword path**:
  confirmed untouched by diff inspection and by direct testing — closing-
  keyword matching (`\b(close[sd]?|fix(e[sd])?|resolve[sd]?)\b`) still fires
  inside backticks/prose exactly as before (that side is intentionally
  unanchored, per the code's own comment about GitHub's actual behavior).
  No regression here.
- **Warning-quoting AC**: `_partial_increment_ref_snippets()` is correctly
  modeled on `_closing_ref_snippets()` and wired into all three warning
  branches (body-conflict / commit-conflict / GraphQL-only-conflict);
  verified via `assert_contains ... '("Part of #4574")'` in PA2 and by
  reading the diff.
- Test coverage is otherwise good (89/89, incident shape + companion +
  fenced + prose + list-marker), just missing the numbered/blockquote marker
  variants that the regex claims to support.

### Verdict

**Changes requested** — the numbered-list-marker digit leak is a real,
reproducible bug that can cause exactly the false-positive-reopen failure
mode this PR (and issue #5234) exists to eliminate, just via a path this
PR's own AC-driven list-marker support newly opens up. Please fix the
second-stage digit extraction and add a numbered-list regression test (and
ideally a blockquote-marker test too, since that path is also untested).
Everything else in the PR is solid and can stay as-is.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:changes-requested` already present — another Judge's verdict raced mine. I'd independently reached the same approve-leaning read of the PR but hadn't yet caught the numbered-list-marker digit leak that Judge found; confirmed it's real. Discarding my review; not writing any label.

rjwalters:
Review feedback addressed — thanks for the reproduction, the numbered-list-marker digit leak was real and I confirmed it before and after.

**Re: [the numbered-list-marker digit leak](https://github.com/rjwalters/loom/pull/5239#issuecomment-5175734935)** (commit `5940dad`)

Root cause was exactly as diagnosed: the AC-driven list-marker support widened the first-stage regex to include `[0-9]+\.`, and the second stage then re-scanned the *whole matched span* with `grep -oE '[0-9]+'`, so the marker's own ordinal was harvested as if it were an issue number.

Applied the suggested fix in `defaults/scripts/merge-pr.sh` — extract the `#N` token first, then strip the `#`:

```bash
| grep -oiE '^[[:space:]]*([-*+>]|[0-9]+\.)?[[:space:]]*(Part of|Contributes to)[[:space:]]+#[0-9]+' \
| grep -oE '#[0-9]+' \
| tr -d '#' \
| sort -un
```

(`tr -d '#'` rather than a third `grep -oE '[0-9]+'` — same result, one less process, and it cannot re-open the same "re-scan for any digit run" class of bug.) Added a comment above the function recording *why* the two-stage order matters, so a future edit does not collapse it back.

Your repro now behaves:

```
$ _partial_increment_refs '3. Part of #789'
789          # was: 3, 789
```

**Regression tests** in `defaults/scripts/tests/test-merge-pr-partial-increment.sh`:

- **PA6** — mirrors your repro end-to-end. Body carries `3. Part of #789` *and* a genuine `Closes #3`. Asserts (a) the extractor yields only `789`, (b) `_check_partial_increment_close_conflict` records **no** conflict, so `Closes #3` stands and #3 is never reopened, (c) only `789` lands in `PARTIAL_OPEN_BEFORE_MERGE`. Added open-issue fixtures for #3 and #789.
- **PA7** — the blockquote marker (`> Part of #456`) you also flagged as untested. It passed before and after (no digits to leak), but the branch is now covered rather than assumed.

I verified the new tests are non-vacuous: running the suite against a copy of `merge-pr.sh` with the fix reverted produces exactly the 3 PA6 failures (90/93), and 93/93 with it.

**Verification**

- `defaults/scripts/tests/test-merge-pr-partial-increment.sh` — **93/93 passed, 0 failed** (was 89/89; +4 assertions)
- All 13 `test-merge-pr-*.sh` suites green (nothing else regressed by the extractor change)
- `bash -n` and `shellcheck --severity=warning` clean on both modified files
- Remote CI green: Shellcheck, Shell Test Suites (hermetic), Docs/Defaults Parity, Installer Integration, AGENTS.md Sync, CLAUDE.md Line Budget

**Not addressed (deliberately out of scope)** — the two low-severity residual gaps you flagged as non-blocking follow-ups, both of which fail toward false-*negatives* rather than the false-positive-reopen class this PR targets:

1. A multi-line inline code span (backtick opens on one line, closes on the next) is not stripped, since the `sed` strip is per-line.
2. An unclosed/odd fence at EOF drops everything after it from matching, since the awk toggle never flips back.

Both need a real block-aware markdown pass rather than a line-wise filter, which is a larger change than this PR's scope. Happy to file a follow-up issue if you'd like them tracked.

Back to `loom:review-requested` for re-review.

rjwalters:
✅ **Approved!**

Solid, well-scoped fix. Reviewed the diff, ran the full regression suite, and checked CI.

**What it does:** Tightens `_partial_increment_refs()` in `merge-pr.sh` so a `Part of #N`/`Contributes to #N` reference only counts as a genuine declaration when it's line-leading (optionally after a list/blockquote marker), and strips fenced code blocks + inline code spans first. This fixes the #4574/kicad-tools incident where a backticked, mid-sentence, hypothetical mention of `Part of #4574` was misread as an authoritative declaration and caused a correctly-closed issue to be reopened.

**Correctness:**
- The two-stage digit extraction (`grep -oiE '...#[0-9]+' | grep -oE '#[0-9]+' | tr -d '#'`) correctly avoids the numbered-list-marker false positive (e.g. `3. Part of #789` no longer leaks the ordinal `3` as a spurious issue reference) — good catch, and it's covered by the PA6 regression test.
- `_partial_increment_ref_snippets()` mirrors the existing `_closing_ref_snippets()` pattern well, giving operators both sides of a conflict in the warning text (AC #4).
- Fenced-block stripping via the awk toggle is simple and correct for the well-formed case.

## Test Execution

1. `./defaults/scripts/tests/test-merge-pr-partial-increment.sh` — ✅ Executed: `93/93 passed, 0 failed` (matches PR's claimed 89/89 plus additional PA6/PA7 tests already included)
2. `bash -n` on both changed files — ✅ Executed: clean
3. `shellcheck --severity=warning` on both changed files — ✅ Executed: clean, 0 findings

**CI status:** all applicable checks pass (Shellcheck, Shell Test Suites, Installer Integration Tests, Docs/Defaults Parity, etc.); Rust/Dashboard jobs correctly skip since this PR touches only shell scripts. Merge state: `CLEAN`. PR description correctly uses `Closes #5234`.

Nice, thorough test coverage (PA1–PA7) covering the exact incident shape, fenced blocks, mid-sentence prose, list markers, numbered-list digit collision, and blockquotes.

rjwalters:
✅ **Approved!**

Clean, well-scoped fix. Reviewed the diff, ran the regression suite, and verified CI.

**Correctness**: `_strip_fenced_code_blocks()` (awk fence toggle) + inline code-span blanking + a line-leading structural anchor (mirroring `_closing_refs_stdin`'s existing `\b` guard) correctly narrows `_partial_increment_refs()` from "matches anywhere in the body" to "matches only a genuine declaration line" — fixing the exact #4574/kicad-tools incident (a backticked, mid-sentence, conditional mention of `Part of #N` was previously read as authoritative and caused a correctly-closed issue to be reopened).

**Test coverage**: Ran `./defaults/scripts/tests/test-merge-pr-partial-increment.sh` directly — `89/89 passed, 0 failed` (matches the PR's claim). New PA1–PA5 cases cover the exact incident shape, the companion genuine-declaration path (still triggers the conflict warning), fenced-block exclusion, non-backticked mid-sentence prose exclusion, and list-marker-prefixed declarations.

**Static checks**: `bash -n` clean on both changed files; `shellcheck --severity=warning` clean.

## Test Execution

1. Run `./defaults/scripts/tests/test-merge-pr-partial-increment.sh` — ✅ Executed: 89/89 passed, 0 failed
2. `bash -n` + `shellcheck --severity=warning` on both changed files — ✅ Executed: clean

## CI Status
All applicable checks pass (Shellcheck, Shell Test Suites, Docs/Defaults Parity, Installer Integration, AGENTS.md Sync, CLAUDE.md Line Budget); Rust/Dashboard jobs correctly skip for a shell-only change. Merge state: CLEAN.

rjwalters:
## Re-review: verified — approving

Independently reproduced and confirmed Doctor's fix (commit `5940dad`) closes the gap from my prior rejection.

**Direct verification of the extraction fix:**

- Extracted `_partial_increment_refs()` as-is from the pre-fix commit (`573c25c`) and ran the exact repro: `_partial_increment_refs("3. Part of #789\nCloses #3\n")` → returned **both `3` and `789`** (the bug: reproduced).
- Ran the same repro against the fixed function (`grep -oE '#[0-9]+' | tr -d '#'` instead of `grep -oE '[0-9]+'`) → returned **only `789`** (confirmed fix).
- Reverted just the two changed lines in the checked-out worktree and re-ran the full `test-merge-pr-partial-increment.sh` suite: exactly 3 assertions failed (all PA6), 90/93 passed — confirms PA6 is non-vacuous and precisely targeted at this regression. Restored the fix, suite back to 93/93.
- Checked `_partial_increment_ref_snippets()` (the other consumer added in this PR) for the same class of bug — it matches `#$2\b` for a specific known issue number rather than re-scanning the matched span for any digit run, so it was never vulnerable to this leak. No other bare `grep -oE '[0-9]+'` digit-extraction call sites remain in the file.
- PA7 (blockquote marker) is a genuine new-coverage addition, not just a defensive test — confirmed it passes and exercises a previously-untested marker branch of the same regex.

**Full validation:**
- `bash -n` and `shellcheck --severity=warning` clean on both changed files.
- All 13 `test-merge-pr-*.sh` suites pass locally (partial-increment suite: 93/93).
- `gh pr checks`: all CI checks green (Shellcheck, Shell Test Suites, Installer Integration Tests, etc.).
- `mergeStateStatus`: CLEAN.

**Diff scope**: the fix commit touches only the two lines in `_partial_increment_refs()`'s second extraction stage plus an explanatory comment, and adds PA6/PA7 test coverage — no unrelated changes, no new risk surface introduced.

Approving.

rjwalters:

merging it automatically:

- **Blast radius**: modifies `defaults/scripts/merge-pr.sh` — the shared script that mediates every fleet merge, which the Champion merge-risk criteria explicitly flag as a high-blast-radius file regardless of hunk size. The specific hunks here (`_strip_fenced_code_blocks`, the tightened `_partial_increment_refs` regex, `_partial_increment_ref_snippets`) are narrow and well-tested (89/89 tests passing, including a Judge-caught-and-fixed numbered-list-marker digit-leak regression), and only feed the pre-merge advisory warning (`_check_partial_increment_close_conflict || true`, non-blocking) and the post-merge self-heal relabel path (`_reset_partial_increment_labels` → `_reset_one_partial_issue`) — not the actual merge/worktree-deletion logic itself. But a regression in this regex would resurface exactly the false-positive-reopen failure class (#5234) this PR is fixing, in the one script every Champion/Doctor/sweep merge depends on, so it warrants a human's eyes before landing.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5239`

---
*Automated by Champion role*

PR #5244, opened 2026-08-04
fix(guard): mask check-duplicate.sh positional arguments in the ask-tier scan

### LINKED ISSUE #5235: Guard: ask-tier strip_literal_text has no positional-argument masking, unlike the #5155 catastrophic-tier fix

(2026-08-04, commit 6ff58955). `stash-scope:main-checkout` is the top ask-tier
pattern by volume (14 hits in the reviewed window) and reproduced live during
this very audit — twice — while trying to dedup-check and then file this exact
finding.

```
./.loom/scripts/check-duplicate.sh \
  "Guard: stash-scope:main-checkout ask-tier blocks headless worktree-sync..." \
  "git stash && <cmd> && git stash pop inside a worktree hits the ask-tier..."
```

This is a **read-only dedup lookup** — `check-duplicate.sh`'s signature is
`check-duplicate.sh TITLE DESCRIPTION` (positional, no flags, never invokes
`git stash` itself). It was denied-then-asked with reason `stash-scope:main-checkout`
because the TITLE/DESCRIPTION prose *describing* the guard bug quoted the phrase
`git stash pop` as inert text, and the ask-tier scan matched it as a raw
positional substring.

The same thing then happened again trying to `create-issue.sh --body "$(cat
<<'EOF' ... EOF)"` this very report, since the body necessarily quotes the
trigger phrase as evidence. Had to fall back to `--body-file` on a
pre-written file to get the report filed at all.

## Root Cause

`guard-destructive-generic.sh` builds `COMMAND_ASK_SCAN` (the string every
`ASK_PATTERNS` entry, including `stash-scope:main-checkout`, matches against) by
running `strip_literal_text()` over the command — but *only* when one of a fixed
set of flag names is present as a substring: `--body`, `--message`, `--title`,
`--notes`, `--comment`, `-m`. `check-duplicate.sh`'s positional-arg signature
means none of those substrings appear, so `strip_literal_text()` never runs at
all and the raw prose is scanned unmasked.

This is the exact same class of bug #5155 diagnosed and fixed — but #5155's fix
(`mask_data_flag_values` / `mask_cat_heredoc_bodies`) only covers the
**catastrophic-tier** scan copy. The **ask-tier** copy (`COMMAND_ASK_SCAN` /
`strip_literal_text`, `guard-destructive-generic.sh` ~L2547-2565) has no
positional-argument masking at all, flag-keyed or otherwise, so every
`ASK_PATTERNS` entry that reuses `COMMAND_ASK_SCAN` (`stash-scope:main-checkout`,
`stash-scope:worktree-collision`, `force-op:protected`, `sql-ddl`, and others)
inherits the same gap.

## Impact

This creates a self-referential trap for exactly the workflow that's supposed to
catch it: the Auditor's standing guard-decision-review policy (#3898) requires
running `check-duplicate.sh` — and then `create-issue.sh --body`, inline — with
text that *names the trigger phrase* before filing a bug about a false-positive
trigger, which re-triggers the same false positive at both steps. In
headless/autonomous runs there is no human to answer the resulting ask, so the
report itself can silently fail to get filed. (Worked around in this session by
using `--body-file` on a pre-written file for the final `create-issue.sh` call,
and by rephrasing the dedup-check text to avoid the literal three-word phrase
for `check-duplicate.sh` — neither is a fix, just evasion.)

Beyond the self-referential case, any positional-arg script/command whose free
text happens to mention `git stash pop/drop/clear`, a `force-op:protected`
branch name, or other ask-tier phrases — code review comments, commit message
search, grep patterns, other dedup/lint helpers — is equally exposed.

## Recommended Fix

Apply the same generalization #5155 recommended for the catastrophic-tier mask,
but to the ask-tier `strip_literal_text()` path: recognize a small allowlist of
known non-executing commands (`check-duplicate.sh`, `grep`, `rg`, other
dedup/search helpers) and mask their positional string arguments the same way
named-flag values are masked today. Keying on invoking-command-name rather than
flag-name mirrors the fix #5155 already validated for the catastrophic tier.

## Impact Level

Low direct risk (ask-then-retry friction, not a safety bypass — the ask tier is
strictly narrower than a hard deny), but it repeatedly stalls exactly the
routine, read-only audit/dedup work the standing guard-review policy depends on,
with no human present in headless runs to clear the ask — and can block filing
the very report that documents it.

---
standing policy (#3898).

rjwalters:
## Summary

`guard-destructive-generic.sh` builds `COMMAND_ASK_SCAN` — the working copy
every `ASK_PATTERNS` entry (including `stash-scope:main-checkout`,
`stash-scope:worktree-collision`, `force-op:protected`, `sql-ddl`, and
others) matches against — by redacting quoted text, but only when a named
text-carrying flag (`--body`, `-m`, `--title`, `--notes`, `--comment`) is
present in the command. Scripts with a purely **positional** signature, e.g.
`./.loom/scripts/check-duplicate.sh TITLE DESCRIPTION` (no flags at all),
never triggered that redaction, so their free-text positional arguments were
scanned unmasked — causing a false ask when the prose happened to quote a
trigger phrase like "git stash pop" as inert text (e.g. filing a bug report
*about* this very guard, or running a dedup check whose description quotes
the trigger phrase).

This is the same class of bug #5155/#5160 already fixed for
`guard-loom-workflow.sh`'s `gh-pr-merge-redirect` scan — but that fix never
covered `guard-destructive-generic.sh`'s independent ask-tier copy.

## Changes

- Added `mask_ask_positional_args()` to `defaults/hooks/guard-destructive-generic.sh`
  — a near-duplicate of `guard-loom-workflow.sh`'s `mask_command_positional_args()`
  (#5155/#5160), masking quoted positional arguments that directly follow a
  small command-name allowlist.
- Wired it into the `COMMAND_ASK_SCAN` construction, running before the
  existing flag-keyed `strip_literal_text()` pass.
- **Scope decision**: the allowlist contains only `check-duplicate.sh`, not
  `grep`/`egrep`/`fgrep`/`rg` as the issue's recommended fix suggested.
  `COMMAND_ASK_SCAN` also feeds the SQL DDL/DML check (`SQL_DDL_PATTERN`),
  which intentionally still scans a `grep <pattern> <file>` invocation's own
  quoted search pattern for a literal SQL table-drop/schema-drop/truncate
  phrase once the command is off the read-only fast path — by design, per
  the existing "Fast path security" / "Fast path off" test groups. Adding
  grep/rg to the allowlist directly regressed 5 of those tests locally;
  `check-duplicate.sh` has no such competing raw-text consumer anywhere in
  the file, so it stays the only entry. This is documented in the new
  function's header comment, with guidance to extend the allowlist only for
  a consumer with no conflicting raw-text scan elsewhere in the file.
- The installed `.loom/hooks/guard-destructive-generic.sh` copy is
  intentionally left untouched — per repo convention, propagating `defaults/`
  edits into the installed copies is the periodic `chore: resync installed
  Loom surfaces` commit's job, done from the main checkout after this PR
  merges.
- Extended `tests/hooks/test-guard-destructive.sh` with 4 new cases: the
  `check-duplicate.sh` false-positive repro (now allows), a `grep` case
  proving the deliberate scope decision (still asks, piped through `cat` to
  disqualify the read-only fast path), a "masking doesn't widen" regression
  (a real `git stash pop` chained after a masked `check-duplicate.sh` call
  still asks), and an "allowlist doesn't overreach" regression (`echo`,
  not allowlisted, still asks).

- [x] Ask-tier scan now masks `check-duplicate.sh`'s positional TITLE/DESCRIPTION
  arguments before matching against `ASK_PATTERNS` — verified via new test
  `ask-tier (#5235): check-duplicate.sh positional TITLE/DESCRIPTION quoting
  'git stash pop' as inert prose no longer asks`.
- [x] Masking narrows only — never widens: a real invocation chained after a
  masked argument, and a non-allowlisted command wrapping the same phrase,
  both still ask (verified by 2 new regression tests).
- [x] No regression to any existing catastrophic/ask-tier behavior, including
  the SQL-DDL fast-path-disqualified tests that motivated excluding
  `grep`/`rg` from the allowlist.
- [x] Installed `.loom/hooks/` copy left untouched (propagated by the
  periodic resync commit, not this PR).

## Test Plan

```
$ bash tests/hooks/test-guard-destructive.sh
...
=========================================
  Total:  750
  Passed: 750
  Failed: 0
=========================================
ALL TESTS PASSED
```

Also confirmed:
- `bash -n defaults/hooks/guard-destructive-generic.sh` — syntax OK
- `shellcheck -x defaults/hooks/guard-destructive-generic.sh` — no new
  warnings (5 pre-existing informational `SC2016`/`SC1003` notes, all
  outside the diff)
- `git diff --stat` scoped to exactly the two files above; no unrelated
  changes; `check-main-clean.sh` confirms the main checkout is untouched.

Closes #5235

### DIFF
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -2155,6 +2155,114 @@ strip_literal_text() {
     }'
 }
 
+# Mask quoted POSITIONAL arguments (no preceding flag name) to a small
+# allowlist of known non-executing commands/scripts (issue #5235, the
+# ask-tier analog of the #5155/#5160 fix). strip_literal_text() above only
+# recognizes text following a named flag (--body/-m/--title/--notes/
+# --comment); it has no effect on a script whose free-text arguments are
+# purely positional, e.g. `./.loom/scripts/check-duplicate.sh "TITLE"
+# "DESCRIPTION"`. check-duplicate.sh never EXECUTES a positional argument —
+# it only reads it as dedup text — so masking a quoted argument immediately
+# following it (optionally after short flags, e.g.
+# `check-duplicate.sh --include-merged-prs "..."`) can never blind the
+# ASK_PATTERNS scan (or any other COMMAND_ASK_SCAN consumer, e.g.
+# stash-scope:main-checkout / stash-scope:worktree-collision) to a real
+# invocation. Deliberately narrow allowlist, same "deliberately narrow"
+# convention documented above strip_literal_text()'s
+# mask_flag_cat_heredocs(): a command that WRAPS the phrase and then
+# executes it — `sh -c "git stash pop"`, `bash -c '...'`, `eval "..."` — is
+# NOT in this allowlist and stays fully visible to every ASK_PATTERNS entry,
+# exactly as before.
+#
+# DELIBERATELY EXCLUDES grep/egrep/fgrep/rg, unlike guard-loom-workflow.sh's
+# mask_command_positional_args() (#5155/#5160) which this function is
+# otherwise modeled on. In THIS file, COMMAND_ASK_SCAN also feeds the SQL
+# DDL/DML check (SQL_DDL_PATTERN, below) — which, once a command is
+# disqualified from (or has opted out of) the #3687 read-only fast path,
+# intentionally scans a `grep <pattern> <file>` invocation's own quoted
+# positional pattern for a literal DDL phrase like "DROP TABLE" and denies,
+# by design (see the "Fast path security" / "Fast path off" test groups in
+# tests/hooks/test-guard-destructive.sh). Adding grep/rg to this allowlist
+# was tried and directly regressed those tests: masking grep's own quoted
+# argument blinded the full-path SQL-DDL scan to text it is specifically
+# tested to still catch. check-duplicate.sh has no such competing consumer
+# of its raw argument text anywhere in this file, so it stays the only
+# entry. Extend this allowlist only for another read-only positional-arg
+# consumer with NO conflicting raw-text scan elsewhere in this file.
+#
+# This is an intentional NEAR-DUPLICATE of guard-loom-workflow.sh's own
+# mask_command_positional_args() (issue #5155/#5160) — kept as a SEPARATE
+# function in this file, same convention as strip_literal_text() above being
+# a separate copy from that file's strip_literal_text(): the two guards'
+# decision-time masking must never be coupled, so a future fix/tuning of one
+# cannot silently change the other's behavior.
+#
+# Masks EVERY quoted argument that directly, consecutively follows the
+# command+flags (separated only by whitespace) — not just the first — so
+# multi-positional-arg scripts like check-duplicate.sh's `TITLE DESCRIPTION`
+# signature get both arguments masked. Masking stops at the first token that
+# is not a quoted string (a bare filename, `&&`, `|`, etc.), leaving anything
+# after that boundary — including a real ask-triggering invocation chained
+# onto the same line — fully visible.
+mask_ask_positional_args() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)
+        DQ = sprintf("%c", 34)
+        # Command-name allowlist: known non-executing commands/scripts whose
+        # positional string arguments are search/dedup text, never live shell
+        # syntax. grep/egrep/fgrep/rg are deliberately NOT here — see the
+        # SQL-DDL conflict documented in the function header comment above.
+        # Extend only when another read-only positional-arg consumer causes a
+        # real false ask AND has no competing raw-text consumer elsewhere in
+        # this file (see #5235, mirroring #5155).
+        cmdre = "(\\./\\.loom/scripts/check-duplicate\\.sh)"
+        # Zero or more short/long flags between the command name and the
+        # first quoted positional argument (e.g.
+        # `check-duplicate.sh --include-merged-prs --issue 5235`).
+        flagre = "([ \t]+-[A-Za-z0-9_-]+)*"
+        anchor = "(^|[ \t\n;&|`(])" cmdre flagre "[ \t]+"
+        buf = ""
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        out = ""
+        while (match(s, anchor)) {
+            pre     = substr(s, 1, RSTART - 1)
+            matched = substr(s, RSTART, RLENGTH)
+            rest    = substr(s, RSTART + RLENGTH)
+            out = out pre matched
+            # Mask every consecutive quoted positional argument immediately
+            # following the anchor (whitespace-separated). Stops at the first
+            # non-quote-starting token, so anything after the argument list
+            # (a pipe, &&, an unrelated command) is left fully visible.
+            while (1) {
+                qc = substr(rest, 1, 1)
+                if (qc != DQ && qc != SQ) break
+                endpos = 0
+                for (i = 2; i <= length(rest); i++) {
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:40:07Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed the diff against `origin/main` (2 files: `defaults/hooks/guard-destructive-generic.sh`, `tests/hooks/test-guard-destructive.sh`).

**Correctness:**
- `mask_ask_positional_args()` masks quoted positional arguments immediately following the `check-duplicate.sh` command name (plus optional short/long flags), before the flag-keyed `strip_literal_text()` pass runs on `COMMAND_ASK_SCAN`. Verified independently outside the hook (extracted the awk logic and ran it standalone):
  - `check-duplicate.sh "title" "git stash pop"` → both positional args masked to `X`s.
  - `check-duplicate.sh "title" "git stash pop" && git stash pop` → masked args, but the chained *real* `git stash pop` stays fully visible (masking narrows, doesn't widen).
  - `check-duplicate.sh "title" "$(git stash pop)"` → command substitution inside the quotes is deliberately left unmasked, so an injected command stays detectable.
  - `echo "git stash pop"` → untouched (echo isn't in the allowlist).
- The deliberate scope decision to exclude `grep`/`rg` from the allowlist (documented in the function header) is correct and tested — `COMMAND_ASK_SCAN` also feeds the SQL-DDL check, and allowlisting grep would blind that check to real DDL phrases in a grep pattern.
- Same near-duplicate-function convention already established for `strip_literal_text()`/`guard-loom-workflow.sh` — keeping the two guards' masking independent is the right call.

**Testing:**
- 4 new test cases cover the false-positive repro, the grep scope-exclusion regression, "masking doesn't widen" (real command chained after a masked call), and "allowlist doesn't overreach" (non-allowlisted `echo`).
- Ran `bash tests/hooks/test-guard-destructive.sh` locally: **750/750 passed**, matching CI and the PR description.
- `bash -n` syntax check clean; `shellcheck -x` shows only the same 5 pre-existing informational SC2016/SC1003 notes, none in the diff.

**Process:**
- PR description uses `Closes #5235` correctly.
- CI: all applicable checks pass (Shell Test Suites, Installer Integration Tests, Codex Adapter Smoke, docs/budget checks); Rust/Dashboard/MCP jobs correctly skip via path filtering since only shell hooks changed.
- `mergeStateStatus`: CLEAN.
- Installed `.loom/hooks/` copy intentionally left untouched per the resync convention — correct.

## Test Execution

1. `bash tests/hooks/test-guard-destructive.sh` — ✅ Executed: 750/750 passed
2. `bash -n defaults/hooks/guard-destructive-generic.sh` — ✅ Executed: syntax OK
3. `shellcheck -x defaults/hooks/guard-destructive-generic.sh` — ✅ Executed: only pre-existing informational notes, none in diff
4. `git diff --stat` scope check — ✅ Executed: confirmed 2 files only

rjwalters:

merging it automatically:

- **Blast radius**: modifies `defaults/hooks/guard-destructive-generic.sh` — the source for an installed guard hook that mediates the ask-tier destructive-command scan run on every agent action fleet-wide. A regression in the new `mask_ask_positional_args()` masking logic (e.g. over-masking that hides a real destructive command, or a wiring bug that breaks `COMMAND_ASK_SCAN`/`SQL_DDL_PATTERN` construction) would degrade guard-hook safety across the whole fleet, not just this feature. This is exactly the class of file the merge-risk criteria call out as high blast radius regardless of diff size or test coverage.

The diff itself looks well-scoped (small, additive, 750/750 tests passing, a specific Judge review that verified behavior independently), but blast radius alone is enough to hold per the merge-risk judgment criteria.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5244`

---
*Automated by Champion role*

Issue #5086, opened 2026-08-03
guard-background-subagents.sh: async-subagent detector is a silent no-op — dispatch tool is named "Agent", not "Task"


rjwalters:
## Problem

`guard-background-subagents.sh`'s async-subagent-dispatch detector (`UNRESOLVED_TASK_IDS`, currently ~line 169) only matches `tool_use` entries with `.name=="Task"`:

```jq
[ .[]? | select(.type=="assistant") | .message.content[]?
  | select(.type=="tool_use" and .name=="Task") | .id ] as $task_ids
```

Verified against ~150 real transcripts in this repo's own `~/.claude/projects/-Users-rwalters-GitHub-loom/*.jsonl` history: the current harness names the async subagent-dispatch tool **`Agent`**, not `Task` — I found 58 real `Agent` dispatches and **zero** `Task` dispatches across the sample. This means `$task_ids` is always empty, `UNRESOLVED_TASK_IDS` is always empty, and `TASK_COUNT` is always `0` — the detector this hook exists to provide (the #4257/#3822 "orchestrator ends its turn while a subagent is still running, headless `-p` exit kills it" hazard) currently **never fires** for any real dispatch. A genuinely orphaned `Agent` dispatch — the exact scenario #4257 was filed for — is silently undetected today.

This is a distinct, separate defect from the one just fixed in #5013/#5079 (which was scoped to the background-Bash detector only and does not touch this one).

## Additional wrinkle: the immediate launch ack shares the dispatch's tool_use id

An `Agent` dispatch is always async (regardless of `input.run_in_background`) and returns an IMMEDIATE `tool_result` ack ("Async agent launched successfully... agentId: `<ID>` ... Use SendMessage with to: ... You will be notified automatically when it completes.") on the **same tool_use id** the real completion later arrives on (verified live: a second, distinct `tool_result` with matching `tool_use_id` carries the actual "Work complete..." summary). So a naive fix that just renames the match to `.name=="Agent"` and reuses the existing `($task_ids - $result_ids)` set-difference would reintroduce the *exact* #4389 hazard on this tool: the immediate launch ack would itself count as "resolved" the instant the dispatch happens, masking a truly orphaned subagent (a false negative).

[...]

PR #5249, opened 2026-08-04
fix(guard): count synchronous Agent/Task dispatches as completed in stop backstop

### LINKED ISSUE #5243: guard-background-subagents.sh false-positives on synchronous Agent dispatches (inline tool_result completions not counted)
## Context

Observed 2026-08-04 in rjwalters/lean-genius immediately after the Loom 0.18 install (PR #43683) wired the Stop hook. At end of a long interactive session, `guard-background-subagents.sh` blocked the stop with "4 dispatched Task/Agent subagent(s) have no observed completion in this transcript yet" — but every dispatched agent had in fact completed:

- 4 async (`run_in_background: true`) agents: all had task-notification completions earlier in the transcript.
- 3 synchronous (`run_in_background: false`) agents: completed inline — their results arrive as ordinary tool_result blocks in the same assistant turn, not as `<task-notification>` events.

The count of 4 suggests the guard's transcript scan keys on a dispatch marker whose completion it expects as a separate notification event, and synchronous dispatches (plus possibly a TaskStop'd background Bash and a SendMessage-resumed agent that re-notified under the same task-id) don't produce the expected completion shape. The second-stop escape hatch worked as documented, so this is a rough edge, not a wedge — but in headless `claude -p` mode an over-blocking Stop guard burns a full extra model turn per sweep.

- [ ] Synchronous Task/Agent dispatches (result inline in tool_result) count as completed
- [ ] TaskStop'd background tasks count as resolved
- [ ] A SendMessage resume that re-notifies under the same task/agent id doesn't double-count the dispatch
- [ ] Regression fixture: transcript with mixed sync/async dispatches, all completed → guard allows the stop first try

---

## Curator Enhancement

Verified against `origin/main` (`6ff58955`, 2026-08-04): `defaults/hooks/guard-background-subagents.sh` is the canonical source (`.loom/hooks/guard-background-subagents.sh` is its installed copy). The `UNRESOLVED_TASK_IDS` detector (jq block starting "Diff Task/Agent subagent-dispatch ids...") is the relevant logic:

```
| if ($id_results | length) == 0 then $id
  elif ($real_completions | length) > 0 then empty
  elif ( [ $agent_ids[] | select(($polled_ok_refs | index(.)) != null) ] | length ) > 0 then empty
  else $id
  end
```

where `$real_completions` is every `tool_result` on the dispatch id whose text does NOT match `"Async agent launched successfully"`.

### Implementation Guidance

A `run_in_background: false` Agent/Task dispatch is a **structural** guarantee, not a heuristic one: the harness cannot let the assistant's turn proceed — let alone reach a `Stop` event — until that tool_use's result has actually landed, because a synchronous call blocks the turn. So unlike the async case (where an immediate launch ack is deliberately NOT treated as completion, per the file's own header comments on the #4389 hazard), a sync dispatch's *first and only* `tool_result` on that id is always the real, final result, even in the edge case where its text incidentally contains the `"Async agent launched successfully"` substring (e.g. if the harness reuses shared boilerplate wording for both sync and async acks — this repo's own `Agent` tool description text, visible in `<system-reminder>` blocks, uses that exact phrase generically, so it is plausible the harness echoes similar wording for a sync completion too).

Recommended fix shape: special-case `.input.run_in_background == false` in the `$task_ids` extraction (or in the per-id resolution branch) so that for those ids, `($id_results | length) > 0` alone resolves the dispatch — skip the launch-ack text exclusion entirely, since no separate async launch ack is structurally possible for a blocking call. This mirrors the existing `.name=="Bash" and (.input.run_in_background == true)` structural filter already used one detector down for background Bash.

**Before implementing**: this repo's convention (see the file's own header, "Verified against live transcripts" appears throughout) is to confirm the exact `tool_result` text shape a real synchronous `Agent`/`Task` completion produces, rather than assuming the ack-text heuristic above — pull a sample transcript with a `run_in_background: false` dispatch (the reporting session in rjwalters/lean-genius, if accessible, or a fresh reproduction) before finalizing the jq change, the same way #5086/#5013/#4696 each cite verified live-transcript shapes.

The issue's second and third suggested acceptance criteria (TaskStop'd background tasks; SendMessage-resumed re-notification dedup) are **not reproduced or root-caused above** — they are carried over from the original report's speculation ("possibly also a TaskStop'd background Bash and a SendMessage-resumed agent that re-notified under the same task-id") about what else might make up the observed count of 4. Confirm against the actual reproduction transcript (or a constructed fixture) whether either is a real, distinct gap before scoping work on them — do not treat them as verified requirements.

### Affected Files

- `defaults/hooks/guard-background-subagents.sh` — the `UNRESOLVED_TASK_IDS` jq detector (Task/Agent resolution logic); possibly also `stopped_task_ids` / `results` in the shared `JQ_PRELUDE` if the TaskStop/SendMessage-resume criteria are confirmed as real gaps
- `.loom/hooks/guard-background-subagents.sh` — installed copy, kept in sync with the `defaults/` source (do not hand-edit independently; resynced via `./.loom/scripts/resync-installed.sh`)
- `defaults/hooks/tests/test-guard-background-subagents.sh` — add the regression fixture(s) required by the issue's 4th acceptance criterion (mixed sync/async dispatch transcript)

### Test Plan

- [ ] Automated tests: add a fixture to `defaults/hooks/tests/test-guard-background-subagents.sh` covering a `run_in_background: false` Task/Agent `tool_use` whose only `tool_result` is the inline final result (no separate launch ack) → guard allows the stop; run `./defaults/hooks/tests/test-guard-background-subagents.sh` and confirm all existing cases (notably the #5086 Agent-name and #4389 background-Bash fixtures) still pass unmodified
- [ ] Manual verification: construct (or reuse, if accessible) a transcript matching the reported scenario — mixed sync/async Agent dispatches all completed — and confirm `guard-background-subagents.sh` allows the stop on the first attempt, not the second
- [ ] Edge cases: a sync dispatch whose tool_result text happens to contain `"Async agent launched successfully"` boilerplate still resolves; a genuinely unresolved async dispatch (no completion at all) still blocks (no regression on the true-positive case this guard exists for)

rjwalters:
## Summary

`guard-background-subagents.sh` (Stop-hook backstop) false-positived on synchronous (`run_in_background: false`) Agent/Task dispatches. Its `UNRESOLVED_TASK_IDS` detector treated a sync dispatch's inline `tool_result` the same as an async dispatch's launch ack — requiring a second, *distinct* completion whose text does **not** match `"Async agent launched successfully"`. So a sync dispatch whose result text incidentally carried that boilerplate re-blocked the stop, burning an extra model turn per sweep in headless `claude -p` mode (the reported scenario in rjwalters/lean-genius).

## Root cause & fix

A `run_in_background: false` dispatch is a **structural** guarantee, not a heuristic one: the harness cannot advance the assistant's turn — let alone reach a `Stop` event — past a blocking `tool_use` until its result has actually landed. So that dispatch's first and only `tool_result` is *always* the real, final result, never a launch ack.

The fix special-cases `.input.run_in_background == false` in the per-id resolution branch: for those ids **any** `tool_result` resolves the dispatch, skipping the launch-ack text exclusion entirely. Only an explicit `== false` triggers this — an absent field stays on the existing branch-(a)/(b) path (a plain `Task` with no field is already resolved by its ordinary `tool_result`), so back-compat is unchanged. This mirrors the `.name=="Bash" and (.input.run_in_background == true)` structural filter already used one detector down for background Bash.

The fix is **structural, not text-shape-dependent**: it resolves sync ids without inspecting the result text at all, so it holds regardless of the exact harness wording a sync completion produces. (The curator note asked to verify the live `tool_result` shape; because the fix deliberately never reads that text for sync ids, the specific shape is not load-bearing — the edge-case fixture `s5243b` exercises the worst case where the sync result text *does* contain the async boilerplate.)

## Scope

Only the first and fourth acceptance criteria were confirmed/root-caused (sync dispatches count as completed; mixed sync/async regression fixture). The second and third (TaskStop'd background tasks; SendMessage-resume dedup) were unverified speculation carried from the original report — TaskStop'd background Bash is already covered by the existing `(t)` fixture, and no distinct SendMessage-resume gap was reproducible, so neither is addressed here per the curator's guidance not to treat them as verified requirements.

## Tests

Added to `defaults/hooks/tests/test-guard-background-subagents.sh`:
- `s5243a` sync Agent + inline result -> allow
- `s5243b` sync result text contains ack boilerplate -> allow (the exact false positive)
- `s5243c` sync Task (back-compat tool name) -> allow
- `s5243d` async dispatch with only its launch ack -> **still block** (short-circuit must not leak to async ids)
- `s5243e` mixed sync/async, all completed -> allow first try (acceptance criterion 4)
- `s5243f`/`s5243g` mixed with one orphaned async -> block exactly one

All 61 cases pass (`./defaults/hooks/tests/test-guard-background-subagents.sh`); existing #5086 / #4389 / #4696 / #5013 fixtures unchanged.

## Files
- `defaults/hooks/guard-background-subagents.sh` — canonical source
- `.loom/hooks/guard-background-subagents.sh` — installed copy, kept byte-identical
- `defaults/hooks/tests/test-guard-background-subagents.sh` — regression fixtures

Closes #5243

### DIFF
--- a/.loom/hooks/guard-background-subagents.sh
+++ b/.loom/hooks/guard-background-subagents.sh
@@ -248,17 +248,41 @@ def stopped_task_ids:
 #      </status>` or `<status>failed</status>` in the result text — NOT
 #      `<status>running</status>`, which is still pending) of the `agentId`
 #      recovered from the launch ack text (`agentId: <ID>`).
+#   c. Structural short-circuit (issue #5243): a dispatch with
+#      `input.run_in_background == false` is SYNCHRONOUS — the harness cannot
+#      advance the assistant's turn (let alone reach a Stop event) past a
+#      blocking tool_use until its result has actually landed, so that dispatch's
+#      FIRST and only `tool_result` is always the real, final result, never a
+#      launch ack (no separate async launch ack is structurally possible for a
+#      blocking call). For these ids ANY tool_result resolves the dispatch — the
+#      launch-ack text exclusion in (a) is skipped entirely, so a sync completion
+#      whose text incidentally contains the "Async agent launched successfully"
+#      boilerplate (e.g. shared harness ack wording) still resolves. This mirrors
+#      the `.name=="Bash" and (.input.run_in_background == true)` structural
+#      filter used by the background-Bash detector below. Only an EXPLICIT
+#      `== false` triggers this; an absent field stays on the (a)/(b) path (a
+#      plain Task with no field is already resolved by its ordinary tool_result
+#      via (a), so back-compat is unchanged).
 #
-# Deliberately does NOT treat the launch ack itself as resolution (that is the
-# exact #4389 hazard recurring on this tool) — only (a) or (b) counts. A
-# genuinely orphaned Agent dispatch (no later distinct tool_result, no terminal
-# TaskOutput poll) still blocks, the true positive this detector exists for.
+# Deliberately does NOT treat an ASYNC dispatch's launch ack as resolution (that
+# is the exact #4389 hazard recurring on this tool) — for async ids only (a) or
+# (b) counts. A genuinely orphaned async Agent dispatch (no later distinct
+# tool_result, no terminal TaskOutput poll) still blocks, the true positive this
+# detector exists for.
 UNRESOLVED_TASK_IDS=$(jq -s -r "$JQ_PRELUDE"'
   . as $t
   | (results) as $r
   | [ $t[]? | select(.type=="assistant") | .message.content[]?
       | select(.type=="tool_use" and (.name=="Task" or .name=="Agent"))
       | .id ] as $task_ids
+  # Synchronous (blocking) Task/Agent dispatch ids (issue #5243): an EXPLICIT
+  # run_in_background == false makes the call block, so its first tool_result is
+  # always the real completion — never a launch ack. Any tool_result on these
+  # ids resolves the dispatch (launch-ack text exclusion skipped for them).
+  | [ $t[]? | select(.type=="assistant") | .message.content[]?
+      | select(.type=="tool_use" and (.name=="Task" or .name=="Agent"))
+      | select(.input.run_in_background == false)
+      | .id ] as $sync_task_ids
   # TaskOutput polls: the poll tool_use id, and the agentId/task ref it names.
   | ( [ $t[]? | select(.type=="assistant") | .message.content[]?
         | select(.type=="tool_use" and .name=="TaskOutput")
@@ -289,6 +313,7 @@ UNRESOLVED_TASK_IDS=$(jq -s -r "$JQ_PRELUDE"'
             | ((.text | capture("agentId: (?<v>[A-Za-z0-9_-]+)")?).v) // empty ]
         ) as $agent_ids
       | if ($id_results | length) == 0 then $id
+        elif (($sync_task_ids | index($id)) != null) then empty
         elif ($real_completions | length) > 0 then empty
         elif ( [ $agent_ids[] | select(($polled_ok_refs | index(.)) != null) ]
                | length ) > 0 then empty
--- a/defaults/hooks/guard-background-subagents.sh
+++ b/defaults/hooks/guard-background-subagents.sh
@@ -248,17 +248,41 @@ def stopped_task_ids:
 #      </status>` or `<status>failed</status>` in the result text — NOT
 #      `<status>running</status>`, which is still pending) of the `agentId`
 #      recovered from the launch ack text (`agentId: <ID>`).
+#   c. Structural short-circuit (issue #5243): a dispatch with
+#      `input.run_in_background == false` is SYNCHRONOUS — the harness cannot
+#      advance the assistant's turn (let alone reach a Stop event) past a
+#      blocking tool_use until its result has actually landed, so that dispatch's
+#      FIRST and only `tool_result` is always the real, final result, never a
+#      launch ack (no separate async launch ack is structurally possible for a
+#      blocking call). For these ids ANY tool_result resolves the dispatch — the
+#      launch-ack text exclusion in (a) is skipped entirely, so a sync completion
+#      whose text incidentally contains the "Async agent launched successfully"
+#      boilerplate (e.g. shared harness ack wording) still resolves. This mirrors
+#      the `.name=="Bash" and (.input.run_in_background == true)` structural
+#      filter used by the background-Bash detector below. Only an EXPLICIT
+#      `== false` triggers this; an absent field stays on the (a)/(b) path (a
+#      plain Task with no field is already resolved by its ordinary tool_result
+#      via (a), so back-compat is unchanged).
 #
-# Deliberately does NOT treat the launch ack itself as resolution (that is the
-# exact #4389 hazard recurring on this tool) — only (a) or (b) counts. A
-# genuinely orphaned Agent dispatch (no later distinct tool_result, no terminal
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:55:02Z) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed the diff, ran the full test suite in an isolated PR worktree, and verified the fix's logic directly.

## Code review
- `.loom/hooks/guard-background-subagents.sh` and `defaults/hooks/guard-background-subagents.sh` are byte-identical, as claimed.
- The jq fix is structurally sound: `$sync_task_ids` collects ids of `Task`/`Agent` dispatches with an *explicit* `run_in_background == false`, and the resolution `elif` short-circuits those ids on **any** `tool_result` — before the async-only `real_completions` (launch-ack-text-exclusion) check. This correctly matches the stated rationale: a blocking dispatch structurally cannot produce a separate launch ack, so its one and only `tool_result` is always the final result. An absent field (plain `Task`/`Agent` with no `run_in_background` key) is untouched — falls through to the pre-existing (a)/(b) branches, preserving back-compat.
- The async true-positive path (`real_completions`/`agent_ids`/`polled_ok_refs`) is untouched — the sync short-circuit only applies to ids actually present in `$sync_task_ids`, so it can't leak into async resolution.

## Test Execution
Ran `./defaults/hooks/tests/test-guard-background-subagents.sh` in the PR worktree — **61/61 passed**, including all 7 new `#5243` fixtures:
- `s5243a`/`s5243c`: sync Agent/Task + inline result → allow
- `s5243b`: sync result text *incidentally* containing "Async agent launched successfully" boilerplate → still allow (the exact reported false positive)
- `s5243d`: async dispatch with only its launch ack → still **block** (confirms no leak into async ids, the #4389 true positive is retained)
- `s5243e`: mixed sync/async, all resolved → allow first try (AC4)
- `s5243f`/`s5243g`: mixed transcript with one genuinely orphaned async dispatch → blocks exactly that one

## Scope
PR explicitly and transparently scopes down to only the first and fourth curator-suggested acceptance criteria, citing that the second/third (TaskStop'd background tasks; SendMessage-resume dedup) were unverified speculation in the original report. That's a reasonable, well-documented scoping call — TaskStop'd background Bash already has its own fixture, and no distinct SendMessage-resume gap was reproducible.

## CI / merge state
- `gh pr checks 5249`: all applicable checks pass (Rust/Dashboard/MCP jobs skip via path filtering — this PR only touches shell hook files, as expected).
- `mergeStateStatus`: CLEAN.

Closes #5243 correctly referenced. Nice, well-scoped fix with thorough regression coverage.

rjwalters:

merging it automatically:

- **Blast radius**: modifies `.loom/hooks/guard-background-subagents.sh` (and its `defaults/` mirror) — the Stop-hook backstop that mediates whether every sweep/agent is allowed to end its turn. It is explicitly called out in this repo's own merge-risk criteria (`.loom/hooks/guard-*.sh`) as a high-blast-radius surface: a regression in the new sync/async short-circuit branch (`elif (($sync_task_ids | index($id)) != null) then empty`) could either reintroduce false-positive re-blocking or, worse, cause a false-negative that lets a genuinely orphaned async dispatch through undetected — fleet-wide, on every stop event, not just one feature.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5249`

---
*Automated by Champion role*

PR #5250, opened 2026-08-04
fix(tests): neutralize PATH resolution in safehoused-service missing-binary test

### LINKED ISSUE #5248: test-safehoused-service.sh: missing-binary preview test fails when a real safehoused binary is on PATH
## Bug Description

`defaults/scripts/tests/test-safehoused-service.sh`'s "placeholder fallback when no binary is found" test case (line 143-144):

```bash
ph_out="$(env -u SAFEHOUSED_BIN bash "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
assert_contains "$ph_out" "ExecStart=/path/to/safehoused" "missing-binary preview falls back to a placeholder ExecStart"
```

only unsets `SAFEHOUSED_BIN`. But `defaults/scripts/cli/safehoused-service.sh`'s binary-resolution order (documented at its own line 40-41, implemented in `locate_safehoused_bin()` at lines 174-179) is:

```
Env SAFEHOUSED_BIN; default: `command -v safehoused`, else ~/.cargo/bin/safehoused.
```

So when `SAFEHOUSED_BIN` is unset but a real `safehoused` binary is already resolvable via `PATH` (e.g. installed at `~/.local/bin/safehoused` on a fleet-provisioned host, which is a normal, expected steady state for hosts that run `safehoused` — see `.loom/docs/safehouse.md`), `command -v safehoused` still finds it, and the script renders a real `ExecStart=<resolved path>` instead of falling back to the `/path/to/safehoused` placeholder the test expects. The test is not hermetic against this — same class of gap as #4881/#5157 (guard tests not isolated from ambient host/environment state).

## Root Cause

`locate_safehoused_bin()` in `defaults/scripts/cli/safehoused-service.sh` (lines 174-179):

```bash
locate_safehoused_bin() {
    if [[ -n "${SAFEHOUSED_BIN:-}" ]]; then echo "${SAFEHOUSED_BIN}"; return 0; fi
    if command -v safehoused >/dev/null 2>&1; then command -v safehoused; return 0; fi
    if [[ -x "${HOME}/.cargo/bin/safehoused" ]]; then echo "${HOME}/.cargo/bin/safehoused"; return 0; fi
    echo ""
}
```

has a 3-step fallback chain (`SAFEHOUSED_BIN` env → `command -v safehoused` on `PATH` → `~/.cargo/bin/safehoused`), but the test at `defaults/scripts/tests/test-safehoused-service.sh` line 143 only neutralizes step 1 (`env -u SAFEHOUSED_BIN`). It does not neutralize step 2 (PATH) or step 3 (`~/.cargo/bin`). On any host where `safehoused` is genuinely installed and resolvable via `PATH` — the expected steady state for a fleet worker per `.loom/docs/safehouse.md` — step 2 resolves before the test can observe the "nothing found" placeholder path, and the assertion at line 144 fails against the real resolved path instead of `/path/to/safehoused`.

Note the `run()` helper defined earlier in the test (lines 83-85) already does `cd "$FAKE_HOME" && HOME="$FAKE_HOME" ...`, which would incidentally neutralize step 3 (`~/.cargo/bin/safehoused`, since `HOME` is redirected to an empty temp dir) — but the placeholder-fallback test case at line 143 does **not** use `run()`; it invokes `bash "$SVC_SCRIPT"` directly, so `HOME` is inherited from the real environment too. That means step 3 is also unguarded by this test case, in addition to step 2 (PATH).

## Reproduction Steps

1. Checkout main: `git fetch origin main && git checkout -B main origin/main`
2. On a host with a real `safehoused` binary on `PATH` (e.g. `~/.local/bin/safehoused`, installed by fleet provisioning — confirm with `command -v safehoused`)
3. Run: `bash defaults/scripts/tests/test-safehoused-service.sh`
4. Observe: `FAIL: missing-binary preview falls back to a placeholder ExecStart` — 23/24 pass, 1 failure.
5. Confirm root cause: `env -u SAFEHOUSED_BIN PATH="/usr/bin:/bin" bash defaults/scripts/cli/safehoused-service.sh --print-unit` (stripping `PATH` of the directory containing the real binary) → correctly falls back to the placeholder and passes.

## Expected Behavior

The "missing binary" preview test should be hermetic and pass regardless of whether the host happens to have a real `safehoused` binary already installed on `PATH` (or at `~/.cargo/bin/safehoused`), since it is explicitly testing the no-binary-resolvable code path.

## Actual Behavior

```
[FAIL] missing-binary preview falls back to a placeholder ExecStart
  expected to contain: [ExecStart=/path/to/safehoused]
  actual: [... ExecStart=/home/ubuntu/.local/bin/safehoused ...]

Tests run: 24, passed: 23, failed: 1
```

## Fix

In the `ph_out=...` line (`defaults/scripts/tests/test-safehoused-service.sh` line 143), neutralize **all three** resolution steps for the duration of that one invocation, not just `SAFEHOUSED_BIN`:

```bash
ph_out="$(env -u SAFEHOUSED_BIN HOME="$(mktemp -d)" PATH="$(mktemp -d)" bash "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
```

- `PATH="$(mktemp -d)"` — an empty scratch dir on `PATH` guarantees `command -v safehoused` cannot resolve anything (neutralizes step 2), replacing the too-narrow original suggestion of only editing `SAFEHOUSED_BIN`.
- `HOME="$(mktemp -d)"` — a fresh empty `HOME` guarantees `~/.cargo/bin/safehoused` does not exist (neutralizes step 3), for the same reason the `run()` helper already redirects `HOME` for every other test case in this file.
- Two separate `mktemp -d` calls are required (one for `PATH`, one for `HOME`) — do not reuse a single temp dir for both, since `HOME` and a `PATH` entry serve unrelated purposes and collapsing them risks incidentally creating `$HOME/safehoused` or similar if any other logic in the script writes there. (It doesn't today, but keeping the two concerns independent avoids a latent coupling.)
- The existing `2>/dev/null` already suppresses the `warn "safehoused binary not found..."` stderr message from `safehoused-service.sh` (line 352); no change needed there.
- Leave every other test case in the file untouched — they already use the `run()` helper, which redirects `HOME` (and therefore already neutralizes `~/.cargo/bin/safehoused`) and runs from a directory with no ambient `SAFEHOUSED_BIN`. Only the placeholder-fallback case at line 143 bypasses `run()` and needs the extra `PATH` neutralization; if you also want to route this case through `run()` for style consistency, note that `run()` does not currently set `PATH`, so simply switching to `run()` alone would still be insufficient without adding a `PATH` override — additive, not a substitute for the `PATH="$(mktemp -d)"` fix.

## Acceptance Criteria

- [ ] The "missing-binary preview falls back to a placeholder ExecStart" test case (and its accompanying `rc=0` check) at `defaults/scripts/tests/test-safehoused-service.sh` line ~143-145 no longer depends on the ambient host's `PATH` or `HOME` — it must assert the placeholder fallback (`ExecStart=/path/to/safehoused`, `rc=0`) using an invocation that neutralizes `SAFEHOUSED_BIN` (env), `PATH` (so `command -v safehoused` cannot resolve), and `HOME` (so `~/.cargo/bin/safehoused` cannot resolve) simultaneously.
- [ ] `bash defaults/scripts/tests/test-safehoused-service.sh` passes 24/24 on a host where a real `safehoused` binary is already installed and resolvable via `PATH` (the fleet-provisioned steady state described above) — this is the regression this issue is about; verify by temporarily placing/symlinking a dummy executable named `safehoused` somewhere on `PATH` (e.g. `ln -s /bin/true /tmp/faux-bin/safehoused && PATH="/tmp/faux-bin:$PATH" bash defaults/scripts/tests/test-safehoused-service.sh`) and confirming the suite still reports 24/24 rather than reproducing the FAIL above.
- [ ] `bash defaults/scripts/tests/test-safehoused-service.sh` still passes 24/24 on a clean host with no `safehoused` on `PATH` and no `~/.cargo/bin/safehoused` (the common case today, e.g. CI) — no regression to the other 23 assertions.
- [ ] No other test case in the file is modified — this is a single-line fix scoped to the one invocation identified above.

## Test Plan

1. Apply the fix to `defaults/scripts/tests/test-safehoused-service.sh`.
2. Run `bash defaults/scripts/tests/test-safehoused-service.sh` in the current environment (whatever it happens to be) — expect `Tests run: 24, passed: 24, failed: 0`.
3. Simulate the fleet-provisioned failure condition this issue reports: create a dummy `safehoused` executable on `PATH` (`mkdir -p /tmp/faux-bin && printf '#!/bin/sh\n' > /tmp/faux-bin/safehoused && chmod +x /tmp/faux-bin/safehoused && PATH="/tmp/faux-bin:$PATH" bash defaults/scripts/tests/test-safehoused-service.sh`) and confirm the "missing-binary preview" test case still passes (24/24) instead of reproducing the reported FAIL.
4. Optionally also simulate the `~/.cargo/bin/safehoused` fallback path (step 3) by creating a dummy file there in a scratch `HOME`, to confirm the fix's `HOME="$(mktemp -d)"` neutralizes it too — not strictly required to close this issue (the reported bug is specifically the `PATH` case) but strengthens the fix's completeness given the root-cause analysis above.
5. Confirm no other assertion in the file regresses (full 24-test run stays green).

## Affected Files

- `defaults/scripts/tests/test-safehoused-service.sh` — the only file requiring a change (single invocation at line ~143).

`defaults/scripts/cli/safehoused-service.sh` is referenced for root-cause context (`locate_safehoused_bin()`, lines 174-179) but does not need to change — its fallback-chain behavior is correct and intentional; only the test's isolation is insufficient.

## Impact

Low: GitHub Actions CI runners don't have `safehoused` pre-installed, so `ci.yml` passes there today. But this suite fails whenever run locally on any host that already has `safehoused` provisioned (any real fleet worker), which is exactly the kind of environment a Doctor/Builder validating changes near `safehoused-service.sh` would use — a spurious failure unrelated to their change.

---
Discovered during main branch audit. Related, same test-isolation class of bug also found this audit: #5245 (test-spawn-codex.sh leaks ambient LOOM_ROLE).

---
Curated: verified `locate_safehoused_bin()`'s 3-step fallback chain (`SAFEHOUSED_BIN` env → `command -v safehoused` on `PATH` → `~/.cargo/bin/safehoused`) against current `defaults/scripts/cli/safehoused-service.sh` (lines 174-179) and the test's `run()` helper (lines 83-85) against `defaults/scripts/tests/test-safehoused-service.sh` — both match `origin/main` at the time of curation. Added root-cause detail on why `HOME` (not just `PATH`) also needs neutralizing for full hermeticity, precise acceptance criteria, and a concrete test plan for reproducing the fleet-provisioned failure condition locally.

rjwalters:
## Summary

`test-safehoused-service.sh`'s "missing-binary preview falls back to a
placeholder ExecStart" test case only unset `SAFEHOUSED_BIN` before invoking
`safehoused-service.sh --print-unit`, but the script's binary-resolution
order (`locate_safehoused_bin()`) is a 3-step fallback chain:
`SAFEHOUSED_BIN` env -> `command -v safehoused` on `PATH` ->
`~/.cargo/bin/safehoused`. On any host where a real `safehoused` binary is
already resolvable via `PATH` (e.g. a fleet-provisioned host with it
installed at `~/.local/bin/safehoused`) or at `~/.cargo/bin/safehoused`, the
test failed because the script rendered a real `ExecStart=<resolved path>`
instead of the expected placeholder `ExecStart=/path/to/safehoused`. Same
class of non-hermetic-test bug as #4881/#5157.

## Changes

- In the `ph_out=...` invocation (`defaults/scripts/tests/test-safehoused-service.sh`),
  also set `PATH` and `HOME` to freshly created, separate empty scratch
  directories (`mktemp -d` each) for the duration of that one subshell, so
  neither `command -v safehoused` (PATH) nor `~/.cargo/bin/safehoused` (HOME)
  can resolve anything regardless of what's actually installed on the host
  running the test. Two independent scratch dirs are used since `HOME` and a
  `PATH` entry serve unrelated purposes.
- Invoke the interpreter via `"$BASH"` (bash's own absolute path, set by bash
  itself) instead of the bare `bash` command name — `env`'s `PATH` override
  also governs `env`'s own lookup of the command it execs, so overriding
  `PATH` while still saying plain `bash` would make `env` itself fail to find
  the `bash` binary (rc=127). Using `$BASH` sidesteps that by never doing a
  PATH-based lookup for the interpreter itself.

| Fix neutralizes `command -v safehoused` PATH resolution for the missing-binary test | done | `PATH="$(mktemp -d)"` on the one invocation |
| Fix neutralizes `~/.cargo/bin/safehoused` HOME resolution for the missing-binary test | done | separate `HOME="$(mktemp -d)"` on the same invocation |
| `bash defaults/scripts/tests/test-safehoused-service.sh` passes all tests | done | `Tests run: 25, passed: 25, failed: 0` |
| Fix restores hermeticity against ambient PATH state (not just passes on this host by luck) | done | Created a dummy executable named `safehoused` in a scratch temp dir, prepended it to `PATH`, re-ran the full suite: still `Tests run: 25, passed: 25, failed: 0`. Scratch dir removed afterward. |
| No other test case in the file modified | done | single-line diff, scoped to the one invocation |

## Test Plan

```
$ bash defaults/scripts/tests/test-safehoused-service.sh
...
Tests run: 25, passed: 25, failed: 0
```

Also verified with a hostile `PATH` (dummy `safehoused` executable
prepended) to confirm the fix actually restores hermeticity against the
regression the issue reports — same result, 25/25 passed, then the scratch
dummy binary was removed.

Closes #5248

### DIFF
--- a/defaults/scripts/tests/test-safehoused-service.sh
+++ b/defaults/scripts/tests/test-safehoused-service.sh
@@ -140,7 +140,7 @@ assert_contains "$unit_u" "ExecStart=$FAKE_BIN" "--unit is accepted and the unit
 # ============================================================================
 # placeholder fallback when no binary is found (still exits 0, valid preview)
 # ============================================================================
-ph_out="$(env -u SAFEHOUSED_BIN bash "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
+ph_out="$(env -u SAFEHOUSED_BIN PATH="$(mktemp -d)" "$BASH" "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
 assert_contains "$ph_out" "ExecStart=/path/to/safehoused" "missing-binary preview falls back to a placeholder ExecStart"
 assert_contains "$ph_out" "rc=0" "missing-binary preview still exits 0 (no side effects)"
 
--- a/defaults/scripts/tests/test-safehoused-service.sh
+++ b/defaults/scripts/tests/test-safehoused-service.sh
@@ -140,7 +140,7 @@ assert_contains "$unit_u" "ExecStart=$FAKE_BIN" "--unit is accepted and the unit
 # ============================================================================
 # placeholder fallback when no binary is found (still exits 0, valid preview)
 # ============================================================================
-ph_out="$(env -u SAFEHOUSED_BIN PATH="$(mktemp -d)" "$BASH" "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
+ph_out="$(env -u SAFEHOUSED_BIN HOME="$(mktemp -d)" PATH="$(mktemp -d)" "$BASH" "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
 assert_contains "$ph_out" "ExecStart=/path/to/safehoused" "missing-binary preview falls back to a placeholder ExecStart"
 assert_contains "$ph_out" "rc=0" "missing-binary preview still exits 0 (no side effects)"
 


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:57:15Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:57:15Z) — standing down without reclaiming. Not stomping.

rjwalters:
Pushed a follow-up commit (d2ba72fc) neutralizing `HOME` as well as `PATH` on the one invocation — the prior commit only closed the `command -v safehoused` (PATH) fallback step, leaving the `~/.cargo/bin/safehoused` (HOME) fallback step still unguarded per the curated issue body. Full suite re-run: 25/25 pass, and the PATH-hijack regression simulation still passes.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:57:15Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Correctness / hermeticity of the fix**

Verified the fix restores hermeticity on this host:
- Normal run: `bash defaults/scripts/tests/test-safehoused-service.sh` → 25/25 pass.
- Hostile `PATH` (dummy `safehoused` prepended to `PATH`): 25/25 pass — the placeholder-fallback assertion still asserts the placeholder, not the hijacked binary's path.
- Hostile `HOME` (dummy `~/.cargo/bin/safehoused` under a fake `HOME`): 25/25 pass — confirms the `HOME` neutralization also holds.

The PR went through two commits (`a1acc0ea` → `d2ba72fc`): the first only neutralized `PATH`, and a follow-up correctly closed the remaining gap by also neutralizing `HOME`, so the final invocation matches the issue's full acceptance criteria — all three resolution steps (`SAFEHOUSED_BIN` env, `PATH`, `HOME`/`~/.cargo/bin`) are neutralized for this one subshell:
```bash
ph_out="$(env -u SAFEHOUSED_BIN HOME="$(mktemp -d)" PATH="$(mktemp -d)" "$BASH" "$SVC_SCRIPT" --print-unit 2>/dev/null; echo "rc=$?")"
```

**Deviation from the issue's literal suggestion (`bash` → `"$BASH"`) — justified**

The issue's literal suggested fix used bare `bash` as the command after `env ... PATH="$(mktemp -d)" ...`. Confirmed locally that this fails: `env`'s `PATH` override also governs `env`'s own lookup of the executable it execs, so an empty scratch `PATH` makes `env` unable to find `bash` itself (rc=127). Using `"$BASH"` sidesteps that — it's bash's own absolute self-path, so `env` execs it directly with no `PATH` lookup, while the *script's* internal `command -v safehoused` check still sees the emptied `PATH`. This is a correct, necessary deviation, not a shortcut.

Confirmed `$BASH` is reliably available in every actual invocation context for this test: the shebang is `#!/usr/bin/env bash`, the file's own usage comment invokes it directly, and CI's `run-ci-suites.sh` invokes each suite via `bash "$path"` — so a real bash process, which always self-populates `$BASH`, is guaranteed to be running before this line executes. No portability concern.

**Other checks**
- `shellcheck` on the modified file: clean, no findings.
- CI: all checks green on the current head (`d2ba72fc`), including "Shell Test Suites (hermetic)" (the suite that runs this test) and "Shellcheck" — 5m13s, pass.
- `mergeStateStatus`: CLEAN.
- Scope: single-line change, matches the issue's "no other test case modified" constraint exactly.

**Minor non-blocking nit**: the two new `mktemp -d` scratch dirs (for `PATH` and `HOME`) aren't cleaned up afterward (unlike `FAKE_HOME`, which has an `EXIT` trap). They're empty directories in the OS temp dir, so the practical impact is negligible, and the issue's own suggested fix had the same property — not worth blocking on, just flagging in case a future pass wants to tidy it up.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T07:57:15Z) — standing down without reclaiming. Not stomping.

rjwalters:

- Merge Conflict Check: GitHub has not finished computing mergeability (`mergeable` is still `null`/`unknown` after repeated checks over ~1 minute). Judge's review noted `mergeStateStatus: CLEAN` as of their pass, but this criterion requires a live read in this pass, not a restated one.

**Next steps:**
- No action needed — this typically resolves within a few minutes once GitHub finishes computing the merge state
- Champion will retry automatically on the next tick

---
*Automated by Champion role*

PR #5253, opened 2026-08-04
fix(gitignore): stop stripping a pre-existing .loom/config.json ignore rule

### LINKED ISSUE #5242: install.sh strips the target's .loom/config.json gitignore rule — host-local runtime config (worktree.root) becomes committable
## Context
Observed 2026-08-04 updating a 0.12.0 install to 0.18.0 (source 6ff58955) in rjwalters/lean-genius: the installer's .gitignore reconciliation **removed the target's `.loom/config.json` ignore rule** (upstream now ships a default config.json as an installed file). On fleet hosts, `.loom/config.json` is host-local runtime state — this repo keeps `worktree.root = "/Volumes/Stripe"` there, documented as "runtime state — gitignored, NOT tracked" in its CLAUDE.md. With the rule stripped, the host-local config surfaces as untracked/committable, and a `git add -A` (or the installer's own index reconcile) can commit one host's volume paths to a shared repo. The downstream PR (rjwalters/lean-genius#43683) had to re-add the ignore as a documented local divergence — which the next installer run will strip again.

Loosely related design context: #5038 (maintenance ownership), but this is a concrete installer behavior, not the ownership question.

- [ ] Installer preserves an existing `.loom/config.json` ignore rule in the target's .gitignore (or never removes ignore rules it didn't add)
- [ ] If upstream wants a tracked default config, ship it under a different name (e.g. config.default.json) so host-local config.json stays ignorable

rjwalters:
## Summary
The installer's `.gitignore` reconciliation (`loom-daemon::init::post_init::update_gitignore`, invoked by both `install.sh`/`install-loom.sh` and `loom-daemon update-gitignore` / `resync-installed.sh`) unconditionally stripped a literal `.loom/config.json` line from a target's `.gitignore` on every run. That behavior dates to #2278 ("stop gitignoring .loom/config.json so it stays tracked across reinstalls"), which assumed any such line was a leftover from an old bug where `.loom/config.json` was accidentally gitignored, blocking the merge-aware tracked-config design (#3598).

That assumption breaks for fleet hosts that *intentionally* gitignore `.loom/config.json` to keep host-local runtime state (e.g. `worktree.root = "/Volumes/Stripe"`) out of a shared repo's tracked config — the installer never adds that ignore rule itself (it's not in `EPHEMERAL_PATTERNS`), so it shouldn't remove it either. Observed 2026-08-04 updating a 0.12.0 install to 0.18.0 in `rjwalters/lean-genius`: the re-added, documented divergence (`rjwalters/lean-genius#43683`) got stripped again on the very next installer run.

## Changes
- `loom-daemon/src/init/post_init.rs`: removed `.loom/config.json` from the `legacy_overbroad` auto-strip list in `update_gitignore`. The genuinely dangerous patterns — `.loom/*.json` and its paired `!.loom/roles/*.json` negation, which shadow *other* installed `.loom/*.json` files (`install-metadata.json`, `config/skill-routes.json`, …) — are still always removed. A single, narrowly-scoped `.loom/config.json` line doesn't shadow anything else, so it's safe to leave alone.
- Updated the existing Rust unit test (`removes_legacy_broad_json_pattern` → `removes_legacy_broad_json_pattern_but_preserves_config_json_rule`) to assert the new behavior, and added a new regression test (`preserves_standalone_config_json_ignore_rule_across_repeated_updates`) covering the exact "re-added after being stripped, stripped again" thrash scenario from the issue.
- Added a shell-level integration test to `scripts/test-installer.sh` that exercises the *real* `loom-daemon init` binary (matching the existing `Test 65` pattern for `.loom/config.json` content merging) against a repo with a pre-existing, documented `.loom/config.json` ignore rule, asserting it survives both a first and a repeated `init --force` run.

I did not pursue the issue's second, optional acceptance criterion (shipping a default config under a different filename, e.g. `config.default.json`) — the source-side template already lives at a different path/name (`defaults/config.json`, merged into the target's `.loom/config.json`), so renaming further wouldn't change the actual point of contention (whether the *target's* `.loom/config.json` stays trackable-by-default vs. host-overridable). The primary fix (not stripping an existing ignore rule) resolves the reported bug directly.

| Installer preserves an existing `.loom/config.json` ignore rule in the target's .gitignore (or never removes ignore rules it didn't add) | ✅ | `.loom/config.json` removed from `legacy_overbroad`; new Rust tests + new shell integration test against the real `loom-daemon init` binary all confirm the rule survives repeated runs |
| Ship a default config under a different name so config.json stays ignorable (optional) | Not pursued | See rationale above — not a clean fit given the existing `defaults/config.json` → `.loom/config.json` merge design; the primary fix already resolves the reported symptom |

## Test Plan
- `cargo test --package loom-daemon --lib init::post_init` — 27/27 passed, including the two new/updated tests.
- `cargo clippy --package loom-daemon --lib -- -D warnings` — clean.
- `cargo fmt --package loom-daemon -- --check` — clean.
- `bash scripts/test-installer.sh` (full suite, including the new `#5242` test block) — 178/179 passed; the 1 pre-existing failure (`test-loom-dispatcher.sh`, launchd-supervision env-var harvesting) reproduces identically on `main` without this change and is unrelated to `.gitignore`/`config.json` reconciliation.

Closes #5242

### DIFF
--- a/loom-daemon/src/init/post_init.rs
+++ b/loom-daemon/src/init/post_init.rs
@@ -359,10 +359,28 @@ pub fn update_gitignore(workspace_path: &Path) -> Result<(), String> {
     // exactly, so line-vector edits are byte-preserving for untouched regions.
     let mut lines: Vec<String> = contents.split('\n').map(str::to_string).collect();
 
-    // Remove legacy over-broad patterns that block config.json from being
-    // tracked (older installs and /imagine used `.loom/*.json`), plus the
-    // negation that was paired with that glob.
-    let legacy_overbroad = [".loom/*.json", ".loom/config.json", "!.loom/roles/*.json"];
+    // Remove legacy over-broad patterns that would shadow *any* installed
+    // `.loom/*.json` file (older installs and /imagine used `.loom/*.json`),
+    // plus the negation that was paired with that glob. These are genuinely
+    // dangerous — they block files this installer never asked the user to
+    // ignore (`.loom/install-metadata.json`, `.loom/config/skill-routes.json`,
+    // …) — so they are always removed regardless of who wrote them.
+    //
+    // #5242: `.loom/config.json` itself is deliberately NOT in this list.
+    // It used to be here (see the historical note below), on the theory that
+    // any occurrence was a leftover from the pre-#2278 bug where `.loom/config.json`
+    // was gitignored by mistake, blocking the merge-aware tracked-config design
+    // (#3598) from working. That theory stopped holding once fleet hosts started
+    // adding this exact line *on purpose*: `.loom/config.json` is committed
+    // team config by default, but a host that keeps genuinely host-local runtime
+    // state there (e.g. a `worktree.root` override) has a legitimate reason to
+    // gitignore it — and this installer never adds that rule itself (it is not
+    // in EPHEMERAL_PATTERNS), so per the "never remove ignore rules we didn't
+    // add" rule it must not strip it either. Removing it here fought that
+    // documented, intentional divergence on every subsequent install/update
+    // (rjwalters/lean-genius#43683). A single scoped `.loom/config.json` line
+    // does not shadow any other installed file, so leaving it alone is safe.
+    let legacy_overbroad = [".loom/*.json", "!.loom/roles/*.json"];
     lines.retain(|line| !legacy_overbroad.contains(&line.trim()));
 
     let begin = lines
@@ -869,7 +887,7 @@ mod tests {
     }
 
     #[test]
-    fn removes_legacy_broad_json_pattern() {
+    fn removes_legacy_broad_json_pattern_but_preserves_config_json_rule() {
         let tmp = TempDir::new().unwrap();
         let gitignore = tmp.path().join(".gitignore");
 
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T08:01:37Z) — standing down without reclaiming. Not stomping.

rjwalters:
**Judge review: APPROVED**

Reviewed the diff, the underlying issue (#5242), and verified the fix directly.

**Correctness**: `update_gitignore()`'s `legacy_overbroad` strip-list in `loom-daemon/src/init/post_init.rs` no longer includes the literal `.loom/config.json` line — only the genuinely dangerous `.loom/*.json` / `!.loom/roles/*.json` shadow patterns remain auto-stripped. I traced the two other related mechanisms in the same file to confirm no conflict:
- `EPHEMERAL_PATTERNS` (the managed-block ignore list) does not include `.loom/config.json`, so the managed block won't re-add an ignore rule for it either — consistent with "never remove a rule we didn't add."
- `OVERBROAD_LOOM_PATTERNS` (the separate hard-fail-on-install check for patterns that would shadow `.loom/scripts/lib/*.sh`) also does not include `.loom/config.json` — unrelated code path, no interaction.

Confirmed the invocation chain matches the PR's claim: `install-loom.sh` → `loom-daemon init` → `init::mod.rs` calls `update_gitignore()` (both the fresh-install and reinit paths), and `resync-installed.sh` → `loom-daemon update-gitignore` hits the same function — so this is the single correct fix point, not a partial patch.

**Safety (blast radius)**: `git merge-tree --write-tree origin/main HEAD` in the builder worktree shows a clean merge, no conflicts. This is scoped to one code path with no other callers discovered.

**Tests**: Ran `cargo test --package loom-daemon --lib init::post_init` myself — 27/27 pass, including the updated `removes_legacy_broad_json_pattern_but_preserves_config_json_rule` and the new `preserves_standalone_config_json_ignore_rule_across_repeated_updates` (idempotency across repeated runs, matching the exact "re-added, stripped again" thrash from the issue). `cargo clippy --package loom-daemon --lib -- -D warnings` clean. The new `scripts/test-installer.sh` block (#5242) follows the existing `create_temp_repo`/binary-guard conventions correctly. CI on this exact commit is green across the board: Rust Unit Tests, Rust Linting & Build Check, Rust OTLP, Installer Integration Tests, and Shell Test Suites (hermetic) — the last two being the suites that would catch a regression in this reconciliation logic across the fleet.

**Acceptance criteria**: Primary AC ("installer preserves an existing `.loom/config.json` ignore rule … or never removes ignore rules it didn't add") is fully met and regression-tested. The second, explicitly optional AC (shipping a default config under a different filename) is reasonably deferred with a clear rationale tied to the existing `defaults/config.json` → `.loom/config.json` merge design — it doesn't block the reported bug fix.

No issues found. Approving.

PR #5256, opened 2026-08-04
fix(init): scope legacy .gitignore cleanup to Loom-authored lines

### LINKED ISSUE #5242: install.sh strips the target's .loom/config.json gitignore rule — host-local runtime config (worktree.root) becomes committable

rjwalters:
## Summary

`loom-daemon`'s `.gitignore` reconciliation stripped **any** line equal to
`.loom/config.json` (or `.loom/*.json` / `!.loom/roles/*.json`) from a target
repo's `.gitignore`, file-wide, with no check for who wrote it. On a fleet host
`.loom/config.json` is host-local runtime state (most visibly `worktree.root`
pointing at a machine-specific volume), and repos gitignore it on purpose — so
every install/update silently re-exposed it as untracked-and-committable, one
`git add -A` away from committing one host's volume paths to a shared repo.
Re-adding the rule never stuck, because `loom-daemon update-gitignore` re-runs
the same strip on every `resync-installed.sh` pass.

The fix takes the issue's **first** suggested acceptance criterion (never remove
ignore rules Loom didn't add) rather than the second (rename the shipped default
to `config.default.json`) — per the curator's guidance, the rename changes the
on-disk contract for every existing install and touches `merge_config_file` +
`manifest.sh`, while the scoped-removal fix satisfies the core AC on its own.

## Changes

- **`loom-daemon/src/init/post_init.rs`** — replaced the unconditional file-wide
  `retain` (which ran *before* the managed-block logic even started) with a
  structural ownership rule:
  - **Marker pair present** → Loom owns exactly the marker span, which the
    existing splice already replaces wholesale. Nothing outside the markers is
    touched. Removal of the separate `retain` is what makes this true.
  - **Markerless (pre-#3590) install** → seed ownership from Loom's own header /
    `EPHEMERAL_PATTERNS` lines, then grow the claim across *adjacent* over-broad
    lines in both directions. An over-broad line not contiguous with any
    Loom-authored line is the consumer's and is left alone.
  - The legacy patterns are now a documented named const,
    `LEGACY_OVERBROAD_PATTERNS`.
- **`scripts/test-installer.sh`** — new Test 76: end-to-end regression against
  the real `loom-daemon init`, complementing Test 65 (which covers config.json
  *values*, not the ignore *rule*).
- **`defaults/docs/troubleshooting.md`** — documents keeping `.loom/config.json`
  host-local as a supported setup, and exactly which `.gitignore` lines Loom
  will reconcile.

- [x] **Installer preserves an existing `.loom/config.json` ignore rule (never
  removes ignore rules it didn't add)** — verified end-to-end against the real
  release binary. Seeded a temp repo's `.gitignore` with a consumer section, ran
  `loom-daemon init --force` twice:

  | binary | after init | `git status --ignored -- .loom/config.json` |
  |---|---|---|
  | pre-fix (`main`) | rule **stripped** | `?? .loom/config.json` (committable) |
  | this PR | rule **survives** both passes | `!! .loom/config.json` (ignored) |

- [x] **Ship a tracked default under a different name (`config.default.json`)** —
  *deliberately not done.* AC 1 fully closes the reported failure; the rename is
  a breaking on-disk-contract change for every existing install with no
  additional benefit here. Called out as optional by the curator.

- [x] **Legacy cleanup is narrowed, not disabled** (issue Test Plan edge case) —
  the pre-existing `removes_legacy_broad_json_pattern` test, whose fixture has
  the over-broad patterns contiguous with Loom's own ephemeral lines, passes
  **unchanged**.

## Test Plan

- [x] `cargo test --lib init::post_init` — 30/30 pass, including 4 new tests:
  - `preserves_hand_added_config_json_rule_outside_loom_block`
  - `hand_added_config_json_rule_survives_repeated_updates` (idempotency across
    resync passes)
  - `preserves_config_json_rule_adjacent_to_managed_block` (asserts the file is
    byte-identical when a consumer rule sits immediately after the END marker)
  - `preserves_hand_added_json_glob_outside_loom_block`
- [x] `bash scripts/test-installer.sh` — Test 76 passes (4 assertions). Suite
  result 180 passed / 1 failed; the single failure is
  `test-loom-dispatcher.sh` (3 sub-cases about supervised `loom-daemon restart`
  env harvesting), **pre-existing and unrelated** — reproduced identically at
  96/99 on `main` before this change.
- [x] `cargo clippy --lib --all-features` clean; `cargo fmt --check` clean;
  `shellcheck` reports nothing in the added test block.
- [x] Manual reproduction of the reported scenario (table above).

Closes #5242

### DIFF
--- a/defaults/docs/troubleshooting.md
+++ b/defaults/docs/troubleshooting.md
@@ -309,6 +309,31 @@ cat .loom/config.json.bak
 `loom-daemon init` against a workspace that already has a `.loom/config.json`, so
 repeat provisioning passes cannot re-enter this path on a tuned host.
 
+### Keeping `.loom/config.json` host-local (git-ignored) (#5242)
+
+`.loom/config.json` is normally **committed** for team sharing. On a fleet host
+where it carries machine-specific values (most commonly `worktree.root` pointing
+at a local volume), a repo may instead choose to keep it untracked. That is
+supported: add your own rule to `.gitignore`, **outside** the Loom-managed block:
+
+```gitignore
+# host-local runtime state — NOT tracked
+.loom/config.json
+```
+
+Install/update passes (`install.sh`, `loom-daemon init`,
+`loom-daemon update-gitignore` via `resync-installed.sh`) preserve it. Loom only
+removes `.gitignore` lines it can show it wrote: everything inside the
+`# >>> loom-managed …` / `# <<< loom-managed <<<` markers, plus legacy over-broad
+patterns (`.loom/*.json`, `!.loom/roles/*.json`) sitting in the same contiguous
+run as a pre-marker Loom block. Keep your rule under its own comment/blank-line
+separation and it will never be reconciled away.
+
+Before #5242 the reconciliation stripped any line equal to `.loom/config.json`
+file-wide, so every upgrade silently re-exposed host-local config as committable —
+one host's volume paths were one `git add -A` away from a shared repo. If you are
+on an older Loom, re-add the rule after each update until you upgrade.
+
 ### Daemon won't start
 
 ```bash
--- a/loom-daemon/src/init/post_init.rs
+++ b/loom-daemon/src/init/post_init.rs
@@ -84,6 +84,21 @@ pub const GITIGNORE_END_MARKER: &str = "# <<< loom-managed <<<";
 /// migrate those installs to the marked form in place.
 const GITIGNORE_BLOCK_HEADER: &str = "# Loom runtime state (don't commit these)";
 
+/// Over-broad `.gitignore` patterns that **Loom itself** used to emit before the
+/// managed-block markers existed (older installs and `/imagine` wrote
+/// `.loom/*.json` plus the `!.loom/roles/*.json` negation paired with it, and
+/// some wrote `.loom/config.json` directly). They shadow `.loom/config.json`,
+/// so Loom migrates them away — but **only when Loom wrote them**.
+///
+/// #5242: these were previously stripped by an unconditional, file-wide match,
+/// which also deleted a *consumer's own* hand-added `.loom/config.json` rule.
+/// On a fleet host `.loom/config.json` is host-local runtime state (e.g. a
+/// machine-specific `worktree.root`), and repos legitimately gitignore it; every
+/// install/update silently re-exposed it as committable. Ownership is now
+/// decided by adjacency — see `update_gitignore`.
+const LEGACY_OVERBROAD_PATTERNS: &[&str] =
+    &[".loom/*.json", ".loom/config.json", "!.loom/roles/*.json"];
+
 /// Ephemeral/runtime files that should be ignored — the single source of truth
 /// for the Loom-managed `.gitignore` block.
 ///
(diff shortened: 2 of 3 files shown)


rjwalters:
Changes requested — merge conflict, and the underlying issue is already closed by a PR that merged 3 minutes before this one opened.

**1. Real merge conflicts (not resolvable by automated rebase)**

`mergeStateStatus` is `DIRTY`. I checked out the existing builder worktree (`.loom/worktrees/issue-5242`, branch `feature/issue-5242` at `018d737a`), fetched `origin/main`, and confirmed with `git merge-tree --write-tree HEAD origin/main` that there are genuine content conflicts (not just an ownership/staleness false-positive):

- `loom-daemon/src/init/post_init.rs` — 4 conflict hunks
- `scripts/test-installer.sh` — 1 conflict hunk

I attempted `git rebase origin/main` in the worktree; it stopped on the same two files with non-trivial, multi-hunk conflicts, so I aborted (`git rebase --abort`, worktree left clean) rather than force a resolution as Judge.

**2. Root cause of the conflict: #5242 was already fixed and closed**

`git log --oneline origin/main -- loom-daemon/src/init/post_init.rs` shows `f662a61a fix(gitignore): stop stripping a pre-existing .loom/config.json ignore rule (#5253)` merged at `2026-08-04T08:11:02Z` — and issue #5242 is confirmed `closed` (`closed_at: 2026-08-04T08:11:03Z`), with `Closes #5242` in that PR's body. This PR (#5256) was opened at `08:14:50Z`, ~4 minutes later, apparently built without the Builder having picked up the just-merged fix — a concurrency race, not a defect in this PR's own code.

**3. This PR's approach is still meaningfully more thorough than #5253's — worth rebasing, not just closing**

I compared the two fixes directly:
- **#5253** (already on `main`) special-cased `.loom/config.json` out of the blind `legacy_overbroad` string-match `retain()`, but left the retain itself in place for `.loom/*.json` and `!.loom/roles/*.json` — so a consumer with a detached, hand-added line matching one of *those* two patterns would still be silently stripped on every `init`/`update-gitignore` pass. The exact bug class from #5242 still exists for that narrower case.
- **#5256** (this PR) removes the blind retain entirely and replaces it with the structural ownership rule described in the PR body (marker-span-only when markers are present; adjacency-seeded growth from Loom's own header/`EPHEMERAL_PATTERNS` lines when markerless) — which closes the bug class generally, not just for `.loom/config.json`.

So this isn't purely redundant — it's a legitimate hardening of #5253's narrower fix. Given that, I'm not closing #5256/its issue reference; I'm requesting changes so it can be rebased onto current `main` (which already contains #5253's narrower fix) and reframed:

- Resolve the conflicts in `loom-daemon/src/init/post_init.rs` and `scripts/test-installer.sh` against the current `main` (the `legacy_overbroad`/retain code this PR deletes has already moved once under #5253 — the structural replacement should still apply cleanly conceptually, just needs a manual re-diff against the new base).
- Update the PR body: `Closes #5242` is no longer accurate (already closed by #5253) — reframe as a follow-up hardening PR (no `Closes` keyword, or reference #5242 as prior art) so it doesn't imply reopening/reclosing an already-resolved issue.
- Re-run `cargo test --lib init::post_init` and `bash scripts/test-installer.sh` against the rebased code — the existing counts/test names in the PR description were measured against the pre-#5253 base and may shift once rebased (e.g. #5253 already renamed `removes_legacy_broad_json_pattern` → `removes_legacy_broad_json_pattern_but_preserves_config_json_rule`).

I did not attempt the conflict resolution myself — per Judge role boundaries, that's Doctor/Builder work.

CI: no check runs / statuses are registered yet on this commit (`018d737a`) via the REST `check-runs` and `status` endpoints — not a factor in this verdict; the merge conflict is the blocker.

rjwalters:
Closing as superseded — issue #5242 was already fixed and merged via #5253 (same root cause, same core files: `loom-daemon/src/init/post_init.rs`, `scripts/test-installer.sh`). This PR is a duplicate build from a concurrent `/loom:sweep` race on the same branch name (`feature/issue-5242`); it now conflicts with `main` (`mergeable_state: dirty`) and its target issue is already closed as completed. Any additional value in this PR's docs change (`defaults/docs/troubleshooting.md`) can be re-proposed separately if still needed.

PR #5257, opened 2026-08-04
fix(uninstall): don't redirect --local targets off a linked worktree

### LINKED ISSUE #5240: install.sh targeting a linked git worktree stages deletions in the PRIMARY checkout (deleted 158 files from a live sibling worktree)
## Context

Observed 2026-08-04 updating a Loom install (0.12.0 → 0.18.0, source commit 6ff58955) in rjwalters/lean-genius. To land the update as a PR, the target passed to the installer was a **linked git worktree** of the repo:

```
git worktree add <repo>/.claude/worktrees/loom-update -b chore/loom-update-0.18.0 origin/main
bash install.sh --quick -y --confirm-reinstall <repo>/.claude/worktrees/loom-update
```

The install itself landed correctly in the worktree (296 changed files, settings hooks wired, metadata written). But the "Reconciling git index after reinstall..." step **also staged 158 deletions in the primary checkout** (`<repo>`, a different worktree sharing the same .git), and actually deleted those files from the primary's working tree — all Loom-managed files plus a 3-line deletion in the primary's CLAUDE.md (the obsolete-marker cleanup, apparently). The primary checkout was mid-use by other sessions; recovery required `git reset` + `git checkout HEAD -- <158 paths>` there.

Likely cause: the reconcile/cleanup step resolves the repo via `git rev-parse --git-common-dir` (or cd's somewhere that resolves to the primary) instead of pinning every git invocation to the target path with `git -C <target>` — so index/worktree mutations land in whichever worktree the resolved path implies, not the install target.

- [ ] All git operations in install.sh / install-loom.sh's reconcile step run with `git -C <target>` (or equivalent) so a linked-worktree target never mutates a sibling worktree's index or files
- [ ] Regression test: install into a linked worktree of a repo whose primary checkout has Loom installed; assert the primary's `git status` is byte-identical before/after

---

### Corrected Root Cause

Verified against `origin/main` (`6ff58955`, 2026-08-04). **The reporter's root-cause hypothesis is incorrect** — do not chase it. `install.sh`'s "Reconciling git index after reinstall..." step (`install.sh:1186`) and the `_emit_loom_owned_dirty_paths` helper it calls (`scripts/install/stash-scope.sh:53-77`) are already correctly scoped with `git -C "$target"` throughout; neither touches a sibling worktree.

The actual defect is in `scripts/uninstall-loom.sh:151-157`:

```bash
# If target is inside a worktree, resolve to the main repository root
MAIN_WORKTREE=$(git -C "$TARGET_PATH" worktree list --porcelain 2>/dev/null | head -4 | grep -m1 '^worktree ' | cut -d' ' -f2- || true)
if [[ -n "$MAIN_WORKTREE" ]] && [[ "$TARGET_PATH" != "$MAIN_WORKTREE" ]]; then
  warning "Target path is inside a worktree: $TARGET_PATH"
  info "Resolving to main repository root: $MAIN_WORKTREE"
  TARGET_PATH="$MAIN_WORKTREE"
fi
```

This runs unconditionally at the very top of `uninstall-loom.sh`, before Step 1 validation, regardless of caller or mode. When `install.sh`'s `--quick --confirm-reinstall` path (`INSTALL_TYPE=="1"`) calls `"$LOOM_ROOT/scripts/uninstall-loom.sh" --yes --local "$TARGET_PATH"` (`install.sh:1103`, and the equivalent interactive `--clean` call at `install.sh:1546`) with a **linked worktree** as `$TARGET_PATH`, this block silently redirects the uninstall subprocess's own `$TARGET_PATH` (and everything derived from it — `WORKTREE_ABS` at `scripts/uninstall-loom.sh:936-938`, the `REMOVE_FILES` deletion loop at `:946-970`, and `git add -A` at `:1471`/`:1617`) to the **primary checkout**. Because this redirect is purely local to the uninstall subprocess's own variables, `install.sh`'s own `$TARGET_PATH` is untouched, and its subsequent `"$LOOM_ROOT/target/release/loom-daemon" init --force --defaults "$LOOM_ROOT/defaults" "$TARGET_PATH"` (`install.sh:1148`) correctly writes the fresh install into the **original, un-redirected** linked worktree. The uninstall and the reinstall end up operating on two different directories — exactly matching the reported symptom (worktree ends up correctly installed; primary checkout gets 158 deletions staged and applied).

`install.sh:1026-1042` (the `#4888` comment block immediately preceding the `--quick` reinstall logic) documents the assumption this violates verbatim: *"the chained uninstall below... mutates `$TARGET_PATH`'s MAIN checkout directly... That is necessary and safe for a --quick reinstall (the fresh install runs synchronously in **that same working tree** right after)."* The assumption is that uninstall and reinstall share one directory; `uninstall-loom.sh`'s worktree redirect breaks that whenever `$TARGET_PATH` is a linked worktree.

Contrast with `scripts/install-loom.sh:590-595`, which has the *same* redirect pattern but is safe there: it runs before `install-loom.sh` branches its own **new**, isolated worktree off `origin/main` (`scripts/install/create-worktree.sh`) for a PR-based install — the redirected `$TARGET_PATH` is only used for repo identification, never as a write target. `install-loom.sh`'s own chained `--local` uninstall call (`scripts/install-loom.sh:711`) is unaffected by this issue either way, since by that point `$TARGET_PATH` has already been resolved to `MAIN_WORKTREE` (line 595), so `uninstall-loom.sh`'s internal redirect is a no-op for that caller.

### Implementation Options

**Option A — Skip the redirect in `LOCAL_MODE`** (`scripts/uninstall-loom.sh`): gate the `scripts/uninstall-loom.sh:151-157` block on `[[ "$LOCAL_MODE" != "true" ]]`. `--local` is documented (`scripts/uninstall-loom.sh:68`) as "Remove files in working directory (no worktree, no PR)" — i.e. operate exactly where told — so bypassing the redirect for local mode is a natural reading of its existing contract, not a new one.
- Pros: surgical (one conditional), preserves today's `--quick`-into-a-worktree capability, doesn't touch `install-loom.sh`'s (already-safe) use of the same pattern.
- Cons: every current `--local` caller was audited above and is safe under this change (both `install.sh` call sites want the un-redirected path; `install-loom.sh`'s call is already a no-op here) — but a *future* standalone `--local` caller that actually wants the main-root redirect would silently lose it.

**Option B — Refuse a linked-worktree target on `install.sh`'s in-place paths**: before `install.sh:1103` and `install.sh:1546`, run the same `git worktree list --porcelain` check and `error` out if `$TARGET_PATH` is not the main worktree, directing the user to the Full Install (`METHOD=2`) path instead — which is already worktree-safe per the contrast above.
- Pros: doesn't touch the shared, multi-caller `scripts/uninstall-loom.sh` redirect logic at all (smaller blast radius on a script with an existing regression-test suite built around its current behavior). Turns the currently-silent-and-violated `#4888` assumption into an enforced precondition.
- Cons: removes a capability that partially works today (quick-install-into-a-worktree, modulo this bug) in favor of forcing the slower PR-based Full Install path for that scenario.

No strong recommendation between the two — this is a judgement call between "fix the shared script to honor its own `--local` contract" vs. "stop the narrower caller from making an unsupported request." Whichever is chosen, the fix must close the gap for **both** `install.sh` call sites (`:1103` quick-reinstall and `:1546` interactive `--clean`), not just one.

### Acceptance Criteria

- [ ] A `--quick --confirm-reinstall` (or interactive `--clean`) install targeting a linked worktree of a repo with an existing Loom install no longer mutates the primary checkout in any way
- [ ] The fix covers both `install.sh:1103` and `install.sh:1546` call sites
- [ ] `scripts/install-loom.sh`'s existing (safe) use of the same worktree-redirect pattern (`:590-595`) is unaffected
- [ ] New regression test: install/reinstall into a linked worktree of a repo whose primary checkout has Loom installed; assert the primary checkout's `git status --porcelain` is byte-identical before/after

The original report's suggested acceptance criteria (below, preserved for reference) targets `git -C` scoping in the reconcile step — that step was already correctly scoped and is **not** the fix; superseded by the criteria above.

### Affected Files

- `scripts/uninstall-loom.sh` (~line 151-157) — the worktree→main-root redirect; likely fix site for Option A
- `install.sh` (~line 1026-1042 comment, ~line 1103, ~line 1546) — the two `--local` chained-uninstall call sites whose downstream `loom-daemon init ... "$TARGET_PATH"` keeps operating on the original, un-redirected target; likely fix site for Option B
- `scripts/install-loom.sh` (~line 590-595) — reference/comparison only, same pattern used safely; should not need changes under either option, but the chosen fix must not regress it
- `defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` and `defaults/scripts/tests/test-install-reinstall-safety.sh` — existing reinstall-safety regression suites; extend with the new linked-worktree-target case rather than adding a new file

### Test Plan

- [ ] Automated tests: extend `defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` (or `test-install-reinstall-safety.sh`) with a case that creates a linked worktree of a Loom-installed throwaway repo, runs the `--quick --confirm-reinstall` path against the worktree path, and asserts the primary checkout's `git status --porcelain` is byte-identical before/after
- [ ] Manual verification: reproduce the reporter's exact repro (`git worktree add <repo>/.claude/worktrees/x -b <branch> origin/main && bash install.sh --quick -y --confirm-reinstall <repo>/.claude/worktrees/x`) against a scratch clone; confirm the primary checkout's `git status` is unchanged and the worktree itself installs correctly
- [ ] Regression: re-run the full existing reinstall-safety suites listed above to confirm the fix doesn't reintroduce #4888 (stranded staged deletions on failure) or #3597 (sibling-installer contamination)

### Dependencies

None.

rjwalters:
## Summary

`scripts/uninstall-loom.sh:151-157` unconditionally redirected a linked-git-worktree `$TARGET_PATH` to the main repository root, before any work happened. That redirect is safe for the script's default (non-local, worktree+PR) mode, but it breaks `install.sh`'s two `--local` chained-uninstall call sites (`:1103` quick-reinstall, `:1546` interactive `--clean`), which rely on the chained uninstall and the subsequent reinstall operating on the **same** directory. When `$TARGET_PATH` was a linked worktree, uninstall silently redirected to the primary checkout (staging deletions there and applying them), while the reinstall proceeded against the original, un-redirected worktree — exactly the reported symptom (158 files deleted from a live sibling checkout mid-use by other sessions).

## Changes

- `scripts/uninstall-loom.sh`: gate the worktree→main-root redirect on `[[ "$LOCAL_MODE" != "true" ]]`. `--local` is documented (line 68 of the same file) as "Remove files in working directory (no worktree, no PR)" — i.e. operate exactly where told — so bypassing the redirect for local mode is a natural reading of its existing contract, not a new one.
- `defaults/scripts/tests/test-install-reinstall-safety.sh`: new **Group 4** regression test covering both directions:
  - `--local` against a linked worktree: exits 0, does NOT print the "Resolving to main repository root" redirect message, operates on the worktree path it was given, leaves the **primary checkout's `git status --porcelain` byte-identical** before/after, and still correctly stages the Loom-managed removal *inside the worktree* (proves the fix isn't an over-correction to a no-op).
  - Default (non-local) mode against a linked worktree: still redirects to the main repository root, preserving existing behavior for other callers (e.g. the worktree+PR flow, and `scripts/install-loom.sh`'s own separate, untouched copy of the identical pattern at `:590-595`).

### Why Option A (gate on `LOCAL_MODE`) over Option B (refuse a linked-worktree target)

The curated issue laid out two options with no strong recommendation. I picked **Option A** (surgical single conditional in the shared `uninstall-loom.sh`) over **Option B** (make `install.sh`'s in-place call sites refuse a linked-worktree target and redirect the user to Full Install) because:
- It preserves the currently-partially-working capability (quick-install/`--clean` into a worktree) instead of removing it.
- It doesn't touch `install.sh` at all, keeping the blast radius to the one script whose contract (`--local` = "operate exactly here") the redirect was actually violating.
- Every existing `--local` caller was audited and is safe under this change: both `install.sh` call sites want the un-redirected path, and `install-loom.sh`'s own chained `--local` call (`:711`) is a no-op either way since by that point `$TARGET_PATH` has already been resolved to `MAIN_WORKTREE` via its own separate copy of the redirect (`:590-595`), which this PR does not touch.

- [x] A `--local` uninstall targeting a linked worktree of a repo with an existing Loom install no longer mutates the primary checkout in any way — verified via the new Group 4 test (`git status --porcelain` byte-identical before/after) and manually reproduced/fixed against a throwaway two-worktree repo (confirmed the bug reproduces on the pre-fix code and is gone post-fix).
- [x] The fix covers both `install.sh` call sites (`:1103` quick-reinstall, `:1546` interactive `--clean`) — both call `uninstall-loom.sh --yes --local "$TARGET_PATH"`, and the fix is in the shared redirect block both paths go through.
- [x] `scripts/install-loom.sh`'s existing (safe) use of the same worktree-redirect pattern (`:590-595`) is unaffected — that file is not modified by this PR.
- [x] New regression test added to `test-install-reinstall-safety.sh` (extended existing file per the issue's test plan, no new file created).

## Test Plan

- `bash defaults/scripts/tests/test-install-reinstall-safety.sh` — 21/21 passed (including the new Group 4, 6 cases).
- `bash defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` — 8/8 passed (unaffected sibling suite).
- `bash scripts/test-installer.sh` — full 131-case installer suite: 176/177 passed. The one failure (`test-loom-dispatcher.sh`, 3 sub-cases re: `LOOM_WORK_FINDER`/`LOOM_MAIN_HEALTH_GATE` env-harvest on daemon restart, issue #4581) is **pre-existing and unrelated** — reproduced identically against the unmodified `defaults/scripts/tests/test-loom-dispatcher.sh` with none of this PR's changes present.
- Manually reproduced the reported bug: seeded a scratch repo + linked worktree, ran the pre-fix `uninstall-loom.sh --yes --local <worktree>` and confirmed it printed "Resolving to main repository root" and staged deletions (`D .loom/config.json`, `D CLAUDE.md`) in the **primary** checkout while leaving the worktree untouched. Re-ran with the fix: target correctly resolves to the worktree, primary checkout status stays empty, worktree shows the staged deletions.
- Reverted the fix temporarily and re-ran the new Group 4 test to confirm it fails on the pre-fix baseline (catches the exact regression), then restored the fix and re-ran to confirm green.

Closes #5240


rjwalters:
LGTM — approving.

Verified against the curated issue's acceptance criteria:

- **Fix matches Option A exactly**: the worktree→main-root redirect in `scripts/uninstall-loom.sh` is now gated on `[[ "$LOCAL_MODE" != "true" ]]`, so `--local` (documented as "operate exactly where told") no longer silently retargets a linked-worktree `$TARGET_PATH`.
- **Both `install.sh` call sites covered**: confirmed lines 1103 (quick-reinstall) and 1546 (interactive `--clean`) both call `uninstall-loom.sh --yes --local "$TARGET_PATH"`, and both route through the single shared conditional this PR fixes — no separate patch needed per call site.
- **`scripts/install-loom.sh`'s own copy of the pattern (~590-595) is untouched**: `git diff origin/main...HEAD -- install.sh scripts/install-loom.sh` is empty, confirming zero blast radius on the file the issue explicitly said to leave alone.
- **Regression test added and passing**: new Group 4 in `defaults/scripts/tests/test-install-reinstall-safety.sh` covers both directions — `--local` against a linked worktree operates in place and leaves the primary checkout's `git status --porcelain` byte-identical before/after (asserted directly, not just "no error"), while confirming the fix isn't an over-correction (the uninstall still stages the Loom-managed removal, just in the worktree). Default (non-local) mode is asserted to still redirect, preserving existing behavior for the worktree+PR flow.

Ran locally in the builder's worktree (`.loom/worktrees/issue-5240`, HEAD `92ad3168`):
- `bash defaults/scripts/tests/test-install-reinstall-safety.sh` → 21/21 passed (all 4 groups, including the new one).
- `bash defaults/scripts/tests/test-install-full-reinstall-no-target-mutation.sh` → 8/8 passed (sibling suite unaffected).
- `shellcheck scripts/uninstall-loom.sh defaults/scripts/tests/test-install-reinstall-safety.sh` → only pre-existing SC2295 info-level warnings elsewhere in the file (unrelated to the changed lines), none introduced by this diff.
- `git merge-tree --write-tree origin/main HEAD` → clean, no conflicts.

CI: all required checks (Shellcheck, Installer Integration Tests, Shell Test Suites (hermetic), Docs/Defaults Parity, AGENTS.md Sync, CLAUDE.md Line Budget, Detect Changes, Codex Adapter Smoke) green; `mergeable_state` is `clean`.

No security or code-quality concerns — this is a minimal, well-scoped conditional gate with a comment explaining the rationale, matching the shared-script contract cleanly.

PR #5218, opened 2026-08-04
fix(daemon): reject unregistered dispatch_sweep workspace_root with a structured error

rjwalters:
## Summary

`mcp__loom__dispatch_sweep` with an explicit `workspace_root` pointing at a repo the daemon does not have registered used to return only an opaque `dispatch_sweep failed: failed to spawn sweep child`. This PR adds a registration check on the explicit-`workspace_root` dispatch path and stops discarding the inner failure detail, so the caller gets a structured, actionable error instead.

## Changes

- **`loom-daemon/src/errors.rs`**: added `DaemonError::workspace_unregistered(root, registered)`, modeled on the existing `workspace_ambiguous` (#4299) precedent — `Configuration`-domain, names the offending path, lists every registered root in `details`, and gives a `recovery_hint` (`loom-daemon workspace add <path>` or pass a registered root).
- **`loom-daemon/src/ipc.rs`**:
  - `resolve_dispatch_registry`'s `Some(root)` branch now loads `WorkspaceRegistry::load_default()` and checks `.contains(&normalized)` before calling `workspace_pool.get_or_provision`; an unregistered root now returns `Response::StructuredError(DaemonError::workspace_unregistered(...))` instead of silently provisioning a registry for an arbitrary directory.
  - `format!("dispatch_sweep failed: {e}")` → `format!("dispatch_sweep failed: {e:#}")` so anyhow's alternate `Display` walks the full `.context()` chain — the specific `resolve_spawn_bin` message ("spawn-worker.sh not found under ...") now reaches the client instead of being silently dropped.
  - Updated `test_sweep_requests_route_to_explicit_workspace_root` to register its target workspace via `seed_temp_registry`, since an explicit `workspace_root` must now be a registered workspace to dispatch successfully.
- **`mcp-loom/src/tools/sweeps.ts`**: added an exported `StructuredError` type mirroring the daemon's `DaemonError`, and a new `formatStructuredError` helper (mirroring the existing `formatRuntimeRejection`/#4494 precedent) that renders `message` + `recovery_hint` + `details` instead of the bare `payload?.message` `extractError` previously read.
- **`mcp-loom/src/tools/sweeps.test.ts`**: unit tests for `formatStructuredError`.

## Acceptance Criteria Verification

- [x] AC #1 — unregistered explicit `workspace_root` returns a structured error naming the offending path and the registered roots: `test_dispatch_sweep_unregistered_explicit_workspace_root_is_structured_error` (loom-daemon).
[...]

Issue #5210, opened 2026-08-04
dispatch_sweep: opaque 'failed to spawn sweep child' when workspace_root is not a registered workspace


rjwalters:
## Context

Calling `mcp__loom__dispatch_sweep` with an explicit `workspace_root` pointing at a repo the daemon does not have registered returns only:

```
=== Dispatch Sweep ===

Failed

dispatch_sweep failed: failed to spawn sweep child
```

Repro from a live session: daemon running and reachable (list_sweeps returned 11 sweeps, all with `Repo: /Users/rwalters/GitHub/loom`), multi-account pool present (7 tokens), then `dispatch_sweep {kind: {Issue: N}, workspace_root: "/Users/rwalters/GitHub/lean-genius"}` → the error above. The caller cannot tell whether the failure is workspace registration, token selection, spawn-script failure, or something else; the sweep-skill Stage -1 contract then falls through to the subagent path on what may be a fixable configuration issue.

Related: #4503 covers the *omitted* `workspace_root` ambiguity/silent-default hazard. This is the complementary case: an explicitly-passed root that isn't registered should produce a structured, actionable error.

## Suggested acceptance criteria

- [ ] When `workspace_root` is not a registered managed workspace, the error names the offending path and lists the registered workspace roots (or says how to register one)
- [ ] Spawn failures unrelated to registration (spawn-claude.sh missing, token selection failure) carry distinct error text
- [ ] The error reaches the MCP client verbatim (not collapsed into "failed to spawn sweep child")

---

### Root Cause (verified against `origin/main` @ e6504319, 2026-08-04)

Two independent gaps combine to produce the opaque error:

1. **No registration check on the explicit-`workspace_root` path.** `resolve_dispatch_registry` (`loom-daemon/src/ipc.rs:1947-1982`) only consults `WorkspaceRegistry::load_default()` on the `None`/empty branch (the #4299 ambiguity check). The `Some(root)` branch (lines 1952-1956) normalizes the path and calls `workspace_pool.get_or_provision(&normalized)` unconditionally — no check that `root` is actually in the registry. `get_or_provision` (`loom-daemon/src/workspace_pool.rs:274-304`) then happily constructs a brand-new `SweepRegistry` for *any* path handed to it, registered or not.
[...]

Issue #4689, opened 2026-07-31
dispatch_sweep reports Success with 'Token: unknown' when token selection failed


rjwalters:
## Context

Hit while sweeping a managed repo (2026-07-30). Four `mcp__loom__dispatch_sweep` calls all returned:

```
=== Dispatch Sweep ===
Success
Sweep ID:   sweep-issue-15-1785436028
PID:        48290
Token:      unknown
Model:      default
```

Every child then died seconds later. The logs showed the real cause:

```
[ERROR] Token selection failed:
error: All 2 tokens in /Users/rwalters/.loom/tokens are marked bad or empty.
[ERROR] Spawn-claude refuses to auto-clear .bad_tokens — that's intentional...
```

The refusal itself is correct and well-reasoned. The problem is the **dispatch surface reported `Success`** with a PID, so from the operator's side four sweeps had launched. It took reading per-sweep logs to discover nothing was running. The `Token: unknown` field was the only hint, and it reads as cosmetic rather than as "selection failed".

## Suggested acceptance criteria

- [ ] When the spawned child exits immediately with the token-selection failure (`EX_CONFIG`/78), `dispatch_sweep` surfaces that as a failure or a clearly-flagged degraded result, rather than `Success`.
- [ ] Alternatively (or additionally), token selection is validated *before* the registry records a Running sweep, so a pool with no usable tokens fails fast at the dispatch call with the actionable message (`loom-tokens unblock <name>` / bootstrap) instead of in a log file.
- [ ] `Token: unknown` is never printed alongside `Success` — if no token was selected, that is the error.

## Note

Adjacent finding, filed separately: the `loom-tokens` CLI this error message recommends is itself broken after the Python retirement.

---

rjwalters:
## Implementation Guidance

Traced the reporting path end to end:

- **The response text is client-side formatting**, not daemon state: `mcp-loom/src/tools/sweeps.ts` builds `=== Dispatch Sweep ===\n\nSuccess\n\n...Token: ${result.result.token_name}...` (`sweeps.ts:1291-1309`) directly from the daemon's `DispatchOutcome`. It never itself checks whether the child is alive.
[...]

PR #5259, opened 2026-08-04
fix(daemon): release the claim lock/label/peer-claim when spawn_child fails

### LINKED ISSUE #5236: dispatch_sweep: spawn_child failure leaks the claim lock (then #4556 guard permanently blocks retries as 'confirmed-live'); inner error chain never surfaced
## Context

Observed 2026-08-04 dispatching sweeps into a registered non-default workspace (rjwalters/lean-genius) whose `.loom/scripts/` lacked `spawn-worker.sh`:

1. **Claim-lock leak on spawn failure.** `dispatch()` acquires the claim lock (step 3) and flips the label before `spawn_child()` runs; when `spawn_child` fails (here: `resolve_spawn_bin` → missing spawn-worker.sh), the lock at `.loom/locks/issue-<N>/owner.json` is left behind with `owner_pid` = the daemon's own pid. Because the daemon is alive, the #4556 live-claim guard then rejects **every retry** as "confirmed-live sweep claim … (pid <daemon-pid>)" — a permanent wedge until the operator manually removes the lock dirs. Each failed retry that gets past a removed lock leaks a fresh one. Repro: point a registered workspace's `SPAWN_BIN` at a nonexistent file and dispatch twice.
2. **Inner error chain not surfaced.** MCP and `loom-daemon dispatch` both report only the outer context, `dispatch_sweep failed: failed to spawn sweep child` — the anyhow inner chain (`spawn-worker.sh not found under <root> (looked in .loom/scripts and defaults/scripts; set LOOM_SPAWN_BIN to override)`) never reaches the caller, and nothing is logged at WARN/ERROR either (the daemon log shows only the INFO dispatch line). Diagnosis required reading sweep_registry source.

- [ ] `dispatch()` releases the claim lock (and reverts the label flip if it happened) when `spawn_child` returns Err
- [ ] The live-claim guard distinguishes "owner_pid is the daemon itself with no tracked child" from a genuinely live sweep (a lock owned by the daemon pid but absent from the registry/children map is stale by construction)
- [ ] Spawn failures surface the full error chain (`{:#}`) in both the IPC response and a WARN log line

---

### Problem Statement

Verified against `origin/main` (`6204411c`, 2026-08-04). Two of the three original claims are confirmed live bugs; the third is **already partially fixed** by a merged PR — see "Status of original acceptance criteria" below.

**Confirmed: claim-lock leak on `spawn_child` failure.** `dispatch_inner` acquires the claim lock at `loom-daemon/src/sweep_registry/dispatch.rs:1050` (`self.acquire_lock(issue_number, &sweep_id)?`), then calls `spawn_child` at `dispatch.rs:1096-1105`:

```rust
let (child, token_name, runtime, immediate_preflight_death) = self
    .spawn_child(...)
    .context("failed to spawn sweep child")?;
```

The `?` here returns immediately on `Err` — it does **not** release the lock, revert the label flip, or retract the peer-claim advertisement. Compare this to the very next error path in the same function, the `#4689` "immediate preflight death" branch at `dispatch.rs:1130-1149`, which explicitly does all three:

```rust
if !self.config.skip_label_flip {
    let _ = self.restore_label_to_ready(issue_number);
    self.note_label_flip(issue_number);
}
self.publish_peer_claim(peer_claims::ClaimKind::Retract, issue_number);
let _ = self.release_lock_owned(issue_number, &sweep_id);
return Err(anyhow!(...));
```

The `spawn_child` failure path (line 1105) has no equivalent cleanup. `spawn_child` itself calls `resolve_spawn_bin()?` first (`dispatch.rs:1296`, defined at `loom-daemon/src/sweep_registry/mod.rs:248-274`), so a missing `spawn-worker.sh` is exactly this uncleaned-up error shape.

**Confirmed: live-claim guard treats the leaked lock as "confirmed live."** `acquire_lock` (`loom-daemon/src/sweep_registry/locks.rs:222-259`) writes `owner_pid: std::process::id()` — the **daemon's own pid** — as a provisional value, with a comment stating `dispatch` rewrites it with the spawned child's real pid via `record_child_pid_in_lock` once the child is running (Issue #3808). When `spawn_child` fails before that rewrite happens, `owner_pid` permanently stays the daemon's own pid. `live_claim::probe`'s leg 1 (`loom-daemon/src/live_claim.rs:32-35`) only checks "is `owner_pid` a live, non-zombie process" — it has no way to distinguish "this pid is the daemon itself with no tracked child" from a genuine live sweep, so as long as the daemon process is running, every retry is refused as a `#4556` confirmed-live claim.

### Status of original acceptance criteria (as of `6204411c`, 2026-08-04)

The third original AC ("Spawn failures surface the full error chain (`{:#}`) in both the IPC response and a WARN log line") is **half-fixed**: PR #5218 (commit `0b3eb680`, merged 2026-08-03T21:18:31-07:00 — before this issue was filed) changed the `dispatch_sweep` IPC error arm at `loom-daemon/src/ipc.rs:3002-3010` to `format!("dispatch_sweep failed: {e:#}")`, which does walk the full `.context()` chain (the code comment there cites `#5210`). This reaches **both** the MCP client and the `loom-daemon dispatch` CLI (both go through this same handler), so the "inner error chain never surfaced" half of the original report no longer reproduces on current main.

**What's still missing**: no `log::warn!` call exists anywhere in that `Err` arm (`ipc.rs:3002-3011`) — a dispatch failure produces only the pre-dispatch `log::info!` line (`ipc.rs:2978-2985`) in the daemon's own log, so an operator tailing the daemon log (rather than an MCP/CLI caller) still can't see *why* a dispatch failed without correlating timestamps against a separate client-side error. The AC below is narrowed to just this remaining gap.

None beyond what's stated above — both remaining defects (lock/label/peer-claim leak on `spawn_child` failure, and the missing WARN log line) are directly read from the code, not inferred.

### Acceptance Criteria

- [ ] `dispatch_inner` releases the claim lock, reverts the label flip (if `!skip_label_flip`), and retracts the peer-claim advertisement when `spawn_child` returns `Err` — mirroring the existing `#4689` immediate-preflight-death cleanup at `dispatch.rs:1130-1149`
- [ ] The live-claim guard (`live_claim::probe` leg 1, `live_claim.rs`) does not treat a lock whose `owner_pid` is the daemon's own pid with no corresponding entry in `self.children`/the registry as a confirmed-live claim — such a lock is stale by construction, since `record_child_pid_in_lock` (Issue #3808) never ran
- [ ] `dispatch_sweep`'s `Err` arm in `loom-daemon/src/ipc.rs` (~line 3002-3011) logs the full error chain (`{e:#}`) at WARN level, not just INFO-level pre-dispatch context — the IPC/MCP/CLI response-surfacing half of this AC is already done (PR #5218/`#5210`, verified on `6204411c`); this AC covers only the daemon-log gap
- [ ] A repro test: point a registered workspace's spawn-bin resolution at a nonexistent file (or otherwise force `spawn_child` to fail) and dispatch twice — the second dispatch must not be refused by the `#4556` live-claim guard

### Affected Files

- `loom-daemon/src/sweep_registry/dispatch.rs` - add cleanup (lock release, label revert, peer-claim retract) to the `spawn_child` error path (~line 1096-1105), mirroring the `#4689` branch at ~line 1130-1149
- `loom-daemon/src/sweep_registry/locks.rs` - `acquire_lock` (~line 222-259) and `release_lock_owned`; may need a way to identify a lock whose `owner_pid` is the daemon's own pid with no tracked child
- `loom-daemon/src/live_claim.rs` - leg 1 of `probe` (~line 32-35 doc, implementation further in the file) — the confirmed-live-lock-owner evidence check
- `loom-daemon/src/ipc.rs` - `dispatch_sweep`'s `Err` arm (~line 3002-3011) — add a `log::warn!` alongside the existing `Response::Error` construction

### Test Plan

- [ ] Automated tests: add a `dispatch_inner`/`dispatch` unit test in `loom-daemon/src/sweep_registry/dispatch.rs` that forces `spawn_child` to fail (e.g. via a test-only spawn-bin override pointing at a nonexistent path) and asserts the lock is released and the label reverted afterward
- [ ] Automated tests: add a `live_claim` test asserting a lock with `owner_pid` == the test's own pid and no corresponding tracked child is NOT reported as confirmed-live evidence (if the fix takes this shape — see AC #2)
- [ ] Manual verification: reproduce the original repro (point a registered workspace's `SPAWN_BIN`/`.loom/scripts/spawn-worker.sh` at a nonexistent path, dispatch twice) and confirm the second dispatch is not refused by the live-claim guard
- [ ] Manual verification: trigger a dispatch failure and confirm a WARN-level line with the full error chain appears in the daemon log (not just the INFO pre-dispatch line)
- [ ] Edge cases: confirm a genuinely live sweep (real child process running) is still correctly refused by the live-claim guard after this change — the fix must not weaken the `#4556` protection for real duplicates

rjwalters:
## Summary

`dispatch_inner` acquires the claim lock and flips the `loom:issue` ->
`loom:building` label *before* `spawn_child` runs. When `spawn_child` itself
returns `Err` (e.g. a registered workspace whose `.loom/scripts/` is missing
`spawn-worker.sh`, so `resolve_spawn_bin` fails before any process is ever
spawned), those side effects were left in place. The leaked lock's
`owner_pid` is the daemon's own (permanently alive, from its own point of
view) pid, which the `#4556` live-claim guard then reads as a confirmed-live
claim on every retry — a permanent wedge until an operator manually removes
the lock dir. This PR fixes the leak and closes the guard's blind spot that
made it permanent, and adds the daemon-log-side half of the error-surfacing
gap (the caller-facing half was already fixed by `#5218`/`#5210`).

## Changes

- `loom-daemon/src/sweep_registry/dispatch.rs`: on a `spawn_child` `Err`,
  `dispatch_inner` now reverts the label flip (if `!skip_label_flip`),
  retracts the peer-claim advertisement, and releases the claim lock —
  mirroring the existing `#4689` immediate-preflight-death cleanup branch
  just below it — before returning the error with its original context.
- `loom-daemon/src/live_claim.rs`: `live_lock_owner_in` (leg 1 of the
  `#4556` live-claim probe) gains an `own_untracked_pid: Option<u32>`
  parameter — a lock whose `owner_pid` equals it is never treated as
  evidence, regardless of what the liveness/argv check would otherwise
  conclude. `probe` is now a thin wrapper over a new `probe_excluding`
  that threads this through; every existing caller (`probe`, unchanged
  signature) is unaffected.
- `loom-daemon/src/sweep_registry/locks.rs`: `live_claim_evidence` (the
  evidence backing the dispatch-time `#4556` guard) now passes
  `Some(std::process::id())` as the exclusion whenever this registry has no
  tracked (non-terminal) entry for the issue — the only way a lock's
  `owner_pid` can legitimately equal the daemon's own pid is
  `acquire_lock`'s provisional placeholder before
  `record_child_pid_in_lock` (`#3808`) rewrites it to the real child pid.
  Extracted a shared `has_tracked_sweep_for` helper (also now used by the
  pre-existing `unregistered_locked_issues`).
- `loom-daemon/src/ipc.rs`: `dispatch_sweep`'s `Err` arm now `log::warn!`s
  the full error chain (`{e:#}`) before constructing the `Response::Error`,
  so a dispatch failure is visible in the daemon's own log, not only in the
  MCP/CLI response.
- `loom-daemon/src/sweep_registry/test_support.rs`: adds
  `spawn_bin_missing_registry`, a fixture mirroring
  `token_selection_failure_registry` but pointing `spawn_bin` at a
  nonexistent file so `spawn_child` itself returns `Err` (rather than
  spawning a child that then dies).

- [x] `dispatch_inner` releases the claim lock, reverts the label flip (if
      `!skip_label_flip`), and retracts the peer-claim advertisement when
      `spawn_child` returns `Err` — mirroring the `#4689` cleanup branch.
      Verified by `dispatch_reverts_claim_lock_and_label_when_spawn_child_itself_fails`.
- [x] The live-claim guard (`live_claim::probe` leg 1) does not treat a lock
      whose `owner_pid` is the daemon's own pid with no corresponding
      tracked entry as a confirmed-live claim. Verified by
      `live_lock_owner_excludes_own_untracked_pid_even_when_live` and
      `probe_excluding_ignores_a_leaked_provisional_lock_owned_by_self`.
- [x] `dispatch_sweep`'s `Err` arm in `ipc.rs` now logs the full error chain
      at WARN. (The response-surfacing half was already done by `#5218`/`#5210`,
      verified still intact by the existing
      `test_dispatch_sweep_registered_workspace_missing_spawn_bin_surfaces_inner_error`.)
- [x] Repro test: `second_dispatch_after_spawn_child_failure_is_not_refused_by_the_live_claim_guard`
      dispatches twice against a workspace whose `spawn_bin` resolves to a
      nonexistent file — the second dispatch reaches the same `spawn_child`
      failure again instead of being refused by the `#4556` guard.

## Test Plan

- `cargo test --lib` in `loom-daemon/`: 3373 passed, 1 pre-existing
  unrelated failure (`daemon_install_state::tests::launchctl_pid_falls_back_to_gui_uid_then_user_uid_when_no_override`
  — fails because this host has a real `loom-daemon` running under
  `gui/501/com.rjwalters.loom-daemon`, an environment-dependent probe of
  the live launchd state; untouched by this diff).
- Targeted reruns, all green: `cargo test --lib` filtered to
  `sweep_registry::` (340 passed), `live_claim` (36 passed, including the
  two new tests and every pre-existing `#4556`/leg-1/leg-2/leg-3 test), and
  `ipc::` (85 passed, including the existing `#5218` inner-error-surfacing
  test).
- `cargo clippy --lib -- -D warnings` and `cargo fmt -- --check`: clean.
- Confirmed a genuinely live sweep is still refused by the live-claim
  guard after this change: `probe_excluding_ignores_a_leaked_provisional_lock_owned_by_self`
  spawns a real distinct sweep process (`FakeSweep`) for a second issue with
  the same exclusion pid in play and asserts it is still reported as live
  evidence — the exclusion only ever matches on exact pid equality, never
  weakening the guard for a real duplicate.

Closes #5236

### DIFF
--- a/loom-daemon/src/ipc.rs
+++ b/loom-daemon/src/ipc.rs
@@ -3004,15 +3004,28 @@ fn handle_request(
                 },
                 Err(e) => match e.downcast::<crate::runtime_admission::RuntimeRejection>() {
                     Ok(rejection) => Response::RuntimeRejected(rejection),
-                    Err(e) => Response::Error {
-                        // #5210: `{e:#}` (anyhow's alternate Display) walks the
-                        // full `.context()` chain instead of printing only the
-                        // outermost context, so a specific inner failure (e.g.
-                        // `resolve_spawn_bin`'s "spawn-worker.sh not found under
-                        // ...") reaches the MCP client instead of being
-                        // silently collapsed into "failed to spawn sweep child".
-                        message: format!("dispatch_sweep failed: {e:#}"),
-                    },
+                    Err(e) => {
+                        // Issue #5236: the pre-dispatch `log::info!` above only
+                        // ever logs the *attempt*, never the failure — until
+                        // now, the daemon's own log had no record of why a
+                        // dispatch failed at all, only the caller's response
+                        // did (#5210/#5218 fixed the caller-facing half). Log
+                        // the same full error chain at WARN so an operator
+                        // reading `loom-daemon`'s log (not just the MCP/CLI
+                        // response) can diagnose a dispatch failure without
+                        // reproducing it.
+                        log::warn!("dispatch_sweep: {kind:?} failed: {e:#}");
+                        Response::Error {
+                            // #5210: `{e:#}` (anyhow's alternate Display) walks
+                            // the full `.context()` chain instead of printing
+                            // only the outermost context, so a specific inner
+                            // failure (e.g. `resolve_spawn_bin`'s
+                            // "spawn-worker.sh not found under ...") reaches
+                            // the MCP client instead of being silently
+                            // collapsed into "failed to spawn sweep child".
+                            message: format!("dispatch_sweep failed: {e:#}"),
+                        }
+                    }
                 },
             }
         }
--- a/loom-daemon/src/live_claim.rs
+++ b/loom-daemon/src/live_claim.rs
@@ -226,15 +226,38 @@ pub fn pid_is_live_claim_for(pid: u32, issue: u32) -> bool {
 /// `is_live` is injected so tests can drive both verdicts without spawning
 /// processes. Fail-open: a missing lock dir, unreadable or unparseable
 /// `owner.json`, or a dead/zombie owner all resolve to `None`.
+///
+/// `own_untracked_pid` (Issue #5236) excludes one specific false-positive:
+/// `SweepRegistry::acquire_lock` stamps `owner_pid` with the *dispatching
+/// daemon's own pid* as a provisional placeholder — it is rewritten to the
+/// spawned child's real pid by `record_child_pid_in_lock` (#3808) once the
+/// child exists. When `spawn_child` fails before that rewrite runs (or ran
+/// historically, pre-#5236, and the caller's own cleanup didn't reach it),
+/// the placeholder is left behind with `owner_pid` == a pid that is, by
+/// construction, still alive: this daemon itself. A bare liveness/argv check
+/// cannot distinguish that from a real claim on non-Linux hosts (no
+/// `/proc/<pid>/cmdline` to read, so [`pid_is_live_claim_for`] falls back to
+/// bare liveness and reports "confirmed live" — the daemon's own pid is
+/// always alive while it's asking the question). Passing
+/// `Some(std::process::id())` here — but ONLY when the caller has already
+/// confirmed no tracked child/registry entry corresponds to this lock — lets
+/// leg 1 recognize that specific placeholder shape and refuse to count it as
+/// evidence, regardless of what `is_live` would otherwise say. `None` (the
+/// default for every non-registry caller, which has no child bookkeeping to
+/// consult) preserves the pre-#5236 behavior exactly.
 #[must_use]
 pub fn live_lock_owner_in(
     locks_dir: &Path,
     issue: u32,
     is_live: &dyn Fn(u32) -> bool,
+    own_untracked_pid: Option<u32>,
 ) -> Option<LiveClaimEvidence> {
     let owner_path = locks_dir.join(format!("issue-{issue}")).join("owner.json");
     let raw = std::fs::read_to_string(&owner_path).ok()?;
     let owner: LockOwnerView = serde_json::from_str(&raw).ok()?;
+    if own_untracked_pid == Some(owner.owner_pid) {
+        return None;
+    }
     if !is_live(owner.owner_pid) {
         return None;
     }
@@ -400,11 +423,33 @@ pub fn live_sweep_process_in(_workspace_root: &Path, _issue: u32) -> Option<Live
 /// Read-only and short-circuiting: leg 1 (a `read_to_string`) and leg 2 (one
 /// small JSON file) run before the `/proc` scan, so the common
 /// nothing-is-claimed case costs two file reads plus a bounded directory walk.
+///
+/// Plain wrapper around [`probe_excluding`] with `own_untracked_pid: None` —
+/// the shape every caller without registry bookkeeping needs (the orphan
+/// reaper, claim reconciliation).
 #[must_use]
 pub fn probe(
     workspace_root: &Path,
     journal_path: Option<&Path>,
     issue: u32,
+) -> Option<LiveClaimEvidence> {
+    probe_excluding(workspace_root, journal_path, issue, None)
+}
+
+/// [`probe`], with one additional exclusion (Issue #5236): a claim-lock
+/// owner whose pid equals `own_untracked_pid` is never treated as evidence,
+/// no matter what leg 1's liveness check would otherwise conclude. See
+/// [`live_lock_owner_in`]'s doc for why this exists — it lets
+/// [`crate::sweep_registry::SweepRegistry::live_claim_evidence`] pass its own
+/// pid once it has confirmed, via its own bookkeeping (`self.entries`), that
+/// no tracked sweep corresponds to the lock. Every other caller passes
+/// `None` via [`probe`] and is unaffected.
+#[must_use]
+pub fn probe_excluding(
+    workspace_root: &Path,
+    journal_path: Option<&Path>,
+    issue: u32,
+    own_untracked_pid: Option<u32>,
 ) -> Option<LiveClaimEvidence> {
     // Both bookkeeping legs verify the recorded PID's argv where the platform
     // allows it ([`pid_is_live_claim_for`]), so a *stale* record whose PID has
@@ -412,7 +457,7 @@ pub fn probe(
     let is_live: &dyn Fn(u32) -> bool = &|pid| pid_is_live_claim_for(pid, issue);
 
     let locks_dir = workspace_root.join(".loom").join("locks");
-    if let Some(evidence) = live_lock_owner_in(&locks_dir, issue, is_live) {
+    if let Some(evidence) = live_lock_owner_in(&locks_dir, issue, is_live, own_untracked_pid) {
         return Some(evidence);
     }
 
@@ -526,7 +571,7 @@ mod tests {
     fn live_lock_owner_reports_a_live_pid() {
         let dir = tempdir().unwrap();
         write_lock(dir.path(), 4275, 4242, "sweep-issue-4275-1");
-        let evidence = live_lock_owner_in(dir.path(), 4275, &|_| true).unwrap();
+        let evidence = live_lock_owner_in(dir.path(), 4275, &|_| true, None).unwrap();
         assert_eq!(
             evidence,
             LiveClaimEvidence::ClaimLock {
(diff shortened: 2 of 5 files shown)


rjwalters:
LGTM — approving.

Reviewed the full diff (dispatch.rs cleanup branch, live_claim.rs exclusion mechanism, locks.rs `has_tracked_sweep_for`, ipc.rs WARN logging, and the new test fixture), ran the targeted test suites locally, and checked the specific hazards called out in the review request:

**1. Cleanup-on-`spawn_child`-failure path vs. the #4689 branch**
The new `Err` match arm in `dispatch_inner` reverts the label flip (`restore_label_to_ready` + `note_label_flip`, gated on `!skip_label_flip`), retracts the peer-claim advertisement, and calls `release_lock_owned` — byte-for-byte the same three-step unwind the pre-existing #4689 immediate-preflight-death branch performs just below it, then returns early with `e.context(...)`. Because the two branches are mutually exclusive arms of the same function (the `Err` arm returns before ever reaching the `immediate_preflight_death` check), there's no path that executes both, so no double-release/double-revert is possible. `release_lock_owned` is itself keyed by `sweep_id` (owner-scoped), so even a hypothetical repeat call couldn't release a different sweep's lock.

**2. `own_untracked_pid` exclusion — could it hide a genuinely live claim?**
Traced this carefully since it's the security-sensitive part of the change:
- `live_claim_evidence` only supplies `Some(own_pid)` when `has_tracked_sweep_for(issue)` is false (no non-terminal registry entry) — a real in-flight dispatch has already inserted its entry (`self.entries.insert`, dispatch.rs L1238) by the time any *other* code path could observe it, and `dispatch_inner` runs under the `Arc<Mutex<SweepRegistry>>` held for the full call (ipc.rs `sr.lock()` around `sr.dispatch(...)`), so no concurrent dispatch for the same issue can interleave with the provisional-lock window in the same process.
- The exclusion matches on exact pid equality only (`own_untracked_pid == Some(owner.owner_pid)`), and only ever fires when `owner_pid` equals the *querying* daemon's own current pid — a different daemon process (or the same daemon after a restart, which gets a new pid) checking the same lock never matches, so a genuinely live claim from another process is still correctly reported as live.
- `probe_excluding_ignores_a_leaked_provisional_lock_owned_by_self` explicitly verifies the exclusion does NOT weaken the guard for a real duplicate: it spawns a distinct `FakeSweep` process for a *different* issue with the same exclusion pid in play and asserts that lock is still reported as live evidence.

**3. Tests / CI**
- Ran locally in the worktree: `cargo test --lib` filtered to `sweep_registry::` (340 passed), `live_claim` (36 passed, including both new tests), and the specific new `dispatch_reverts_claim_lock_and_label_when_spawn_child_itself_fails` test (passes in isolation too).
- `cargo clippy --lib -- -D warnings` and `cargo fmt -- --check`: clean.
- CI (`gh pr checks`): all required checks pass (Rust Linting & Build, Rust Unit Tests, Rust OTLP Feature Build & Test, plus the doc/lint checks); dashboard/installer/shell suites correctly skipped (no touched paths).
- Merge state: `CLEAN`, `mergeable: MERGEABLE`.

No blocking issues found. The `ipc.rs` WARN-logging addition and the new `spawn_bin_missing_registry` test fixture are both small, well-scoped additions that match the stated intent.

PR #5274, opened 2026-08-04
fix(guard): admit read-only search piped to a read-only sink on the fast path

### LINKED ISSUE #5263: Guard false positive: sql-ddl catastrophic pattern fires on grep introspecting a DDL keyword as a search pattern

(2026-08-04, commit 9cd7c35f). The `sql-ddl` catastrophic-tier pattern fired 7
times, all read-only false positives.

## Occurrence

```
grep -n "sql-ddl\|SQL DDL\|sql_ddl\|DROP TABLE" tests/hooks/test-guard-destructive.sh | head -40
grep -n "DROP TABLE" tests/hooks/test-guard-destructive.sh
```

Both are read-only greps over the guard's own test fixtures, searching for
where the DDL pattern is tested — introspection, not a live SQL statement.
`grep` never executes what it searches for.

## Reproduced live, during this audit

I hit the same bug myself while writing this report: `check-duplicate.sh
"<title>" "<description containing the DDL keyword as inert prose>"` was
denied at **catastrophic** tier by `sql-ddl`, because the description
literally quoted the trigger phrase to explain the bug. This is the same
inert-text-as-live-invocation failure the report describes, on the very
command used to file it — I had to obfuscate the keyword via string
concatenation to get the dedup check to run.

## Root Cause

Same class as #5158 (`catastrophic:curl .* | .*sh` firing on `grep`
introspecting its own pattern) and #4881 (unresolved-$VAR false positive) —
the guard's regex scan does not distinguish "this text is a quoted search
argument to a non-executing command (`grep`/`rg`)" or "this text is inert
prose describing the trigger phrase" from "this is a live SQL statement about
to run." Per #5235's PR discussion, `sql-ddl` was deliberately *kept*
scanning raw positional args (including through `check-duplicate.sh`)
because it has its own grep-introspection test coverage in
`test-guard-destructive.sh` — but that design decision is exactly what makes
both occurrences above (and my own repro) false positives, not the intended
catch.

## Recommended Fix

Extend the same masking used for `gh-api-rawfield-body-literal-at` /
`worktree-write-confinement` to recognize `grep -e`/`-E`/positional-pattern
and `rg` arguments as inert search text for `sql-ddl` too, OR (per #5235's
existing scope note) keep `sql-ddl` unmasked for grep but add a narrower
carve-out so `check-duplicate.sh`'s own two positional args (TITLE,
DESCRIPTION) — already masked for the ask-tier scan by #5244 — get the same
treatment on the catastrophic-tier `sql-ddl` scan. Whoever picks this up
should re-check #5235's test suite first since it explicitly chose not to
touch `sql-ddl`'s grep behavior for a documented reason.

## Impact

Blocks routine source-reading/audit work and even blocks filing a bug report
*about* this exact guard (catastrophic tier has no human to answer). Low
frequency but self-defeating for the standing telemetry-review policy itself.

---
standing policy (#3898).

---

### Scope update: the check-duplicate.sh repro is already fixed

The issue's "Reproduced live" section (`check-duplicate.sh "<title>" "<description
containing the DDL keyword as inert prose>"` being denied) was reproduced at
commit `9cd7c35f`/`6ff58955`, **before** PR #5244 (`fd29f454`, merged
2026-08-04) landed. #5244 added `mask_ask_positional_args()` masking of
`check-duplicate.sh`'s positional TITLE/DESCRIPTION arguments to
`COMMAND_ASK_SCAN` — the same scan `SQL_DDL_PATTERN` reads
(`defaults/hooks/guard-destructive-generic.sh:2684-2691` builds
`COMMAND_ASK_SCAN`, `:2819-2823` is the SQL-DDL check, both against the same
variable, masking applied first). **Verified fixed as of `f662a61a`,
2026-08-04**: a live guard invocation with a `check-duplicate.sh "title"
"description"` command whose description quotes the DDL trigger phrase as
inert prose now allows silently (exit 0, no output) — re-verify against
current `main` before relying on this, per the standard staleness caveat.

### Remaining, still-open false positive: pipe disqualifies the read-only fast path

The part of this issue that is **still live** is narrower than the original
report: a **bare** `grep '<ddl-phrase>' file` (no pipe, no wrapper) is already
correctly allowed via the `#3687`/`#3772` read-only fast path — verified live
at `f662a61a`. The false positive is specifically that **piping** a read-only
grep to *any* other command (`| head`, `| less`, `| wc -l`, `| cat`, …)
disqualifies the fast path (by design — pipes are a general security
disqualifier, not something specific to SQL-DDL) and falls through to the
full-path scan, where `SQL_DDL_PATTERN` still matches the literal phrase in
grep's own search argument and denies — even though the command is 100%
read-only. Verified live at `f662a61a`:

```
grep -n "DROP TABLE" tests/hooks/test-guard-destructive.sh          # ALLOW (fast-pathed)
grep -n "DROP TABLE" tests/hooks/test-guard-destructive.sh | head   # DENY  (pipe -> full path -> SQL-DDL fires)
```

This second form is an extremely common interactive idiom (piping a grep to
`head`/`less` to page results), making this a real, frequently-reachable
false positive, not an edge case.

### Root Cause (confirmed, not just suspected)

`defaults/hooks/guard-destructive-generic.sh:2819-2823` — the SQL-DDL check —
scans `COMMAND_ASK_SCAN` for a literal `DROP DATABASE|DROP TABLE|DROP
SCHEMA|TRUNCATE TABLE` substring match with no distinction between "this text
is a live SQL statement" and "this text is a quoted search argument to
`grep`/`rg`". This is **by design**, not an oversight — see the extensive
comment block at `:2158-2199` (the `mask_ask_positional_args()` header):
masking grep/rg's positional arguments was tried once already and reverted
because it "directly regressed" the `tests/hooks/test-guard-destructive.sh`
"Fast path security" test group, which explicitly asserts (as of `f662a61a`):

- `:2565-2566` — a piped `grep` for the DDL phrase disqualifies the fast path
  and the full-path SQL-DDL check still denies — **existing passing test**
- `:2569-2570` — a `bash -c` wrapper around the same `grep` is not admitted
  to the fast path either — **existing passing test**
- `:2580-2588` — fast-path-off (env/config) still denies on a bare `grep`
  for the DDL phrase — **existing passing test**

These tests currently encode "grep introspecting a DDL keyword, once outside
the narrow bare-command fast path, should still deny" as **intended**
behavior — a documented defense against obfuscating a live DDL invocation
behind a pipe/wrapper/substitution. A fix for this issue must change that
intent for the genuinely-inert cases without regressing the obfuscation
defense the tests above exist to enforce. That is the actual difficulty here,
not a simple allowlist addition.

### Related Open Work: #5158

`check-duplicate.sh --issue` flagged **#5158** (open, `loom:operator-only`) as
cross-referencing this issue. #5158 reports the identical failure class
(a catastrophic curl-pipe-to-shell pattern firing on grep introspecting its
own pattern in `defaults/hooks/guard-destructive.sh`, a sibling guard file)
and — per its 2026-08-04T13:44 comment, filed by the same auditor session
that filed this issue — explicitly proposes that **both** issues might be
better solved by one generalized rule ("a quoted pattern/search-term argument
to a known non-executing command — grep, rg, gh --search — is inert text")
applied once across all catastrophic patterns, rather than one-off,
per-pattern fixes. #5158 itself was rejected twice by Champion for being
underspecified (no file:line citation, no AC, no test plan) and escalated to
`loom:operator-only` for a human to decide revise/close/accept.

**Disposition (absorbed, narrowly)**: this issue stays scoped to the
`sql-ddl` pattern specifically — it already has the file:line citations,
confirmed repro, and AC checklist Champion found missing on #5158, so forcing
it to also solve the generalized cross-pattern case would reintroduce that
same underspecification risk here. The Implementation Guidance below flags
the shared-mechanism option as worth considering *if* it turns out to be
low-cost, but does not require it — a human should separately decide on
#5158 whether to pursue the generalized rule as its own follow-up. Do **not**
let this issue block on #5158's resolution.

### Suspected Cause (unverified) — none beyond the confirmed root cause above

No additional unverified hypothesis; the mechanism is directly demonstrated
by the live reproductions above, not inferred.

### Acceptance Criteria

- [ ] A `grep`/`rg` search for a DDL phrase piped to a read-only sink
      (`head`, `less`, `wc`, `cat`, `tail`) no longer denies at the
      `sql-ddl` catastrophic tier when the phrase appears only inside a
      quoted, non-executing search pattern argument to
      `grep`/`egrep`/`fgrep`/`rg`.
- [ ] A genuine obfuscation attempt is still caught: a shell wrapper or pipe
      around an **actual** DDL-executing command (e.g. `mysql -e` /
      `psql -c`), not a grep search argument, still denies exactly as today.
- [ ] None of the existing "Fast path security" / "Fast path off" assertions
      in `tests/hooks/test-guard-destructive.sh` (`:2506-2588`) regress from
      deny to allow for their *non-grep* cases (the `find -delete`/`-exec`/
      `-fls`, the `&&`/`;`/`$(...)`-chained force-push and destructive-rm
      cases, the `bash -c` wrapper, and the fast-path-off assertions must
      all keep denying exactly as today).
- [ ] The grep-specific "Fast path security"/"Fast path off" assertions that
      currently assert a piped/wrapped `grep`-for-DDL still DENY
      (`:2565-2566`, `:2569-2570`, `:2580-2588`) are deliberately updated to
      ALLOW where the wrapped command is `grep`/`rg` alone piped only to a
      read-only sink, with the assertion text/comment updated to explain why
      the expected verdict changed, not silently flipped.
- [ ] `bash tests/hooks/test-guard-destructive.sh` passes with 0 failures
      after the change (this repo's own primary-clone caveat noted in the
      issue's Reproduction Steps — run with no sibling `.loom/worktrees/*`
      contamination, or from a fresh checkout, to match CI's hermetic
      environment).

### Implementation Guidance

Two independent angles, not mutually exclusive:

1. **Narrow, sql-ddl-only fix**: extend fast-path eligibility (or add a
   dedicated narrower carve-out reusing `mask_ask_positional_args()`'s
   allowlist shape) to recognize `grep`/`egrep`/`fgrep`/`rg` piped *only* to
   a small, fixed allowlist of non-executing read sinks (`head`, `tail`,
   `less`, `more`, `wc`, `cat`) as still fully read-only — i.e., broaden the
   fast-path "excluded chars" disqualification so a pipe is only
   disqualifying when the downstream command is not itself read-only-safe,
   rather than treating every pipe as disqualifying. This is more surgical
   than the previously-reverted "mask grep's positional arg everywhere"
   attempt (`:2158-2199`'s header comment), since it does not touch the
   wrapper (`bash -c`)/substitution (`$(...)`)/compound (`&&`/`;`) cases at
   all — those keep denying via the existing full path, satisfying the AC's
   obfuscation-still-caught requirement without new logic there.
2. **Shared cross-pattern mechanism** (per #5158's comment): if angle 1's
   read-sink allowlist ends up general enough to live in a shared helper
   (not `sql-ddl`-specific), consider whether it should also cover
   `defaults/hooks/guard-destructive.sh`'s curl-pipe-to-shell pattern from
   #5158 in the same PR — but only if it is genuinely low incremental cost;
   do not block this issue's own fix on fully solving #5158's broader,
   still-underspecified scope.

Whoever picks this up should re-run `tests/hooks/test-guard-destructive.sh`
first on unmodified `main` to see the current green baseline before changing
anything — the "Fast path security" group is dense and it is easy to
misattribute a pre-existing failure to your own change.

### Affected Files

- `defaults/hooks/guard-destructive-generic.sh` - the SQL-DDL check
  (`:2819-2823`) and the read-only fast-path eligibility logic feeding it
  (search for the `#3687`/`#3772` fast-path implementation earlier in the
  file); `mask_ask_positional_args()` (`:2207-2264`) is reference/context for
  the masking pattern already used for `check-duplicate.sh`, not necessarily
  the mechanism to extend (see Implementation Guidance option 1).
- `tests/hooks/test-guard-destructive.sh` - "Fast path security"/"Fast path
  off" test group (`:2495-2593` as of `f662a61a`); update the grep-pipe
  assertions per Acceptance Criteria, add new cases for the read-sink
  allowlist (`head`/`tail`/`less`/`wc`/`cat`) and for the still-must-deny
  wrapper/obfuscation case.
- `defaults/docs/guard-hooks.md` - update the read-only fast-path
  documentation if the pipe-disqualification rule changes (verify whether
  this file documents the exclusion list before editing).

### Test Plan

- [ ] Automated tests: extend `tests/hooks/test-guard-destructive.sh`'s
      "Fast path security" group per the Acceptance Criteria above; run
      `bash tests/hooks/test-guard-destructive.sh` from a checkout with no
      sibling `.loom/worktrees/*` (or a fresh clone) to match CI.
- [ ] Manual verification: re-run the two live reproductions in this
      enhancement (bare `grep` for the DDL phrase vs. piped to `head`)
      against the patched guard directly via piping a synthetic
      `PreToolUse` JSON payload into
      `defaults/hooks/guard-destructive-generic.sh` and confirm both allow.
- [ ] Edge cases: confirm a real DDL-executing command wrapped in the same
      shapes (a `bash -c` wrapper around `mysql -e`, or `mysql -e '...' |
      cat`) still denies; confirm `check-duplicate.sh`'s existing #5244 fix
      is untouched (re-run its own test coverage if any exists, or the live
      repro in this enhancement's "Scope update" section).

rjwalters:
## Summary

Fixes the `sql-ddl` catastrophic-tier false positive where a **read-only** `grep`/`rg` search for a DDL phrase piped to a read-only sink (`grep 'DROP TABLE' schema.sql | head`) was denied — even though the DDL phrase lives only inside grep's own quoted search argument, which grep never executes.

The bare form `grep 'DROP TABLE' schema.sql` (no pipe) was already fast-pathed and allowed via the #3687/#3772 read-only fast path. The pipe disqualified the fast path and fell through to the full path, where `SQL_DDL_PATTERN` substring-matched the literal phrase in grep's argument and denied. Piping a search to a pager/counter is one of the most common interactive idioms, so this was a frequently-reachable, self-defeating false positive (it even blocked filing a bug report *about* the guard).

## Changes

- **`defaults/hooks/guard-destructive-generic.sh`** — new `fastpath_grep_pipe_admits()` helper (zero forks, pure bash builtins), wired into the fast-path dispatch as a new `elif` branch after `fastpath_builtin_admits`. It admits **only** the shape `grep|egrep|fgrep|rg <args> | <read-only-sink>`:
  - **exactly one pipe**, and **none** of `;` `&` `<` `>` backtick `$(` or newline anywhere — so wrapper (`bash -c`), substitution (`$(...)`), and compound (`&&`/`;`) forms are **untouched** and keep denying via the full path (obfuscation still caught);
  - **upstream** must be a non-executing search (`grep`/`egrep`/`fgrep`/`rg`) — a real DDL executor (`mysql -e '…' | cat`, `psql -c '…' | head`) has a non-search first token and still denies;
  - **downstream** must be a read-only sink: `head`/`tail`/`wc` (any args — already fully allowlisted) or `cat`/`less`/`more` (**only** as pure stdin consumers with no positional file operand, so `grep x | cat ~/.ssh/id_rsa` is NOT fast-pathed and the existing `cat` `.ssh`/`.aws` ASK still fires).
  - Gated by the same `guards.readOnlyFastPath` / `LOOM_GUARD_READONLY_FASTPATH` toggle.
- **`tests/hooks/test-guard-destructive.sh`** — see "Test assertions updated" below.
- **`defaults/docs/guard-hooks.md`** — documents the search-pipe carve-out as the single exception to the "no `|`" fast-path rule.

## Test assertions updated (deliberately, not silently flipped)

- **Flipped deny → allow** (`Fast path security` group): `grep <ddl> | cat` now fast-paths to a silent allow. Comment rewritten to explain the verdict change.
- **Flipped ask → allow** (#5235 ask-tier group): `grep -n "…git stash pop…" file | cat` — its prior premise (pipe-to-cat is used to *reach* the ask-tier scan) no longer holds now that the search-pipe is fast-pathed. Comment rewritten; the two regression guards below it (a `&&`-chained real `git stash pop`, an `echo … | bash`) still correctly ASK, proving only the search-to-sink shape is admitted.
- **New allow coverage**: `| head`, `| head -n 40`, `| tail -5`, `| wc -l`, `| less`, `| cat -n`, plus `rg`/`egrep`/`fgrep` upstreams.
- **New still-must-deny coverage**: `mysql -e <ddl> | cat`, `psql -c <ddl> | head`, `grep <ddl> | sh` (non-sink), `grep <ddl> | grep x | head` (two pipes), `bash -c "grep <ddl> | head"` (wrapper+pipe), `grep foo | cat ~/.ssh/id_rsa` (still asks), and fast-path-off (env+config) piped variants.
- **Not regressed**: all `find -delete`/`-exec`/`-fls`, `&&`/`;`/`$(...)` force-push and rm cases, the `bash -c` wrapper, and the bare-grep fast-path-off assertions keep denying exactly as before.

- [x] `grep`/`rg` search for a DDL phrase piped to a read-only sink no longer denies at `sql-ddl` — verified live and via new tests.
- [x] Genuine obfuscation still caught — `mysql -e`/`psql -c` piped to a sink, and `bash -c` wrappers, still deny (new tests).
- [x] Non-grep `Fast path security`/`Fast path off` assertions do not regress.
- [x] Grep-specific piped assertions deliberately updated to allow, with explanatory comments.
- [x] `bash tests/hooks/test-guard-destructive.sh` passes with **0 failures** (779 total, was 762 at baseline).

## Test Plan

- [x] `bash tests/hooks/test-guard-destructive.sh` → 779 passed, 0 failed (baseline 762 passed, 0 failed).
- [x] Manual verification via synthetic `PreToolUse` JSON payloads: bare grep, `| head`, `| head -n 40`, `| wc -l`, `| cat`, `| less`, `rg | head` all allow silently; `mysql -e | cat`, `psql -c | head`, `bash -c "grep | head"`, `grep | sh`, two-pipe chain all deny; `grep foo | cat ~/.ssh/id_rsa` asks.
- [x] `shellcheck --severity=warning` clean (exit 0); only pre-existing info-level SC2016 on `$('` literal-match patterns.

Closes #5263

### DIFF
--- a/defaults/docs/guard-hooks.md
+++ b/defaults/docs/guard-hooks.md
@@ -814,7 +814,7 @@ The fast path is **on by default**. It is resolved in this order (highest preced
 
 **Security — the fast path is a guard bypass by construction**, so admission is purely **structural** and conservative, never content-sensitive. A command is fast-pathed only when **all** of these hold (otherwise it falls through to the full path unchanged):
 
-- The raw command contains **none** of `;` `&` `|` `<` `>` backtick `$(` or a newline — this excludes all chaining, piping, redirection, and command substitution. So `git status && git push --force origin main`, `git status; rm -rf /`, and `git status $(rm -rf /)` all take the full path and are still denied.
+- The raw command contains **none** of `;` `&` `|` `<` `>` backtick `$(` or a newline — this excludes all chaining, piping, redirection, and command substitution. So `git status && git push --force origin main`, `git status; rm -rf /`, and `git status $(rm -rf /)` all take the full path and are still denied. **One narrow exception (#5263):** a read-only search piped to a single read-only sink — `grep`/`egrep`/`fgrep`/`rg <args> | (head|tail|wc|cat|less|more)` — is still admitted (see the search-pipe carve-out below), because a bare `grep 'DROP TABLE' schema.sql` was already fast-pathed and the pipe to a pager/counter does not add any executing command.
 - The **first token** is an exact allowlist match (never a wrapper — `bash -c`, `sh -c`, `eval`, `xargs`, `env … git status`, `sudo git status` are all excluded because their first token isn't allowlisted):
 
 | First token | Admitted form |
@@ -827,10 +827,18 @@ The fast path is **on by default**. It is resolved in this order (highest preced
 | `gh` | `gh <noun> view` / `gh <noun> list` (never `delete`/`close`/`archive`/…) |
 | `aws` | `aws <service> describe*` / `get*` / `list*`, and `aws s3 ls` |
 
-**`cat` and `ssh` are deliberately EXCLUDED** from the built-in list, even though they are read-only in spirit:
+**`cat` and `ssh` are deliberately EXCLUDED** from the built-in first-token list, even though they are read-only in spirit:
 
 - `cat` has a narrow existing `ASK` carve-out (`cat …/.ssh/…`, `cat …/.aws/credentials`); a blanket `cat` fast-path would silently skip it.
 - `ssh <host> '<cmd>'` wraps an **opaque remote command string** that the raw `ALWAYS_BLOCK` catastrophic scan still covers today; fast-pathing any `ssh …` would drop that coverage.
+
+**Search-pipe carve-out (#5263)** — the single documented exception to the "no `|`" rule above. A read-only search piped to one read-only sink is admitted, because the phrase the guard would otherwise fire on lives only inside the search command's quoted pattern argument (which is never executed). This fixes a self-defeating false positive: a bare `grep 'DROP TABLE' schema.sql` was already fast-pathed and allowed, but piping it to `head`/`less` to page the results (`grep 'DROP TABLE' schema.sql | head`) fell through to the full path, where the `sql-ddl` catastrophic check substring-matched the literal `DROP TABLE` in grep's own argument and **denied** — one of the most common interactive idioms. The carve-out admits **only** this exact shape:
+
+- **exactly one `|`**, and **none** of `;` `&` `<` `>` backtick `$(` or a newline anywhere — so wrapper (`bash -c '… | …'`), substitution (`$(…)`), and compound (`&&`/`;`) forms are untouched and keep denying via the full path (obfuscation still caught);
+- the **upstream** command word is a non-executing search: `grep`, `egrep`, `fgrep`, or `rg` (a real DDL executor like `mysql -e '…' | cat` or `psql -c '…' | head` has a non-search first token, so it is **not** admitted and still denies);
+- the **downstream** command word is a read-only sink: `head`, `tail`, `wc` (already fully allowlisted, so any arguments), or `cat`, `less`, `more` (admitted **only** as pure stdin consumers with no positional file operand — so `grep x | cat ~/.ssh/id_rsa` is **not** fast-pathed and the `cat` `.ssh`/`.aws` `ASK` carve-out above still fires).
+
+A second pipe, an unlisted sink, or a `cat`/`less`/`more` with a file operand all decline the carve-out and take the full path unchanged (a false negative is always safe). The carve-out is gated by the same `guards.readOnlyFastPath` / `LOOM_GUARD_READONLY_FASTPATH` toggle — disabling the fast path disables it too.
 
 **Optional extend-only escape hatch** — `guards.readOnlyFastPathExtra` is an array of **literal first-word commands** to add to the built-in list without hand-editing the Loom-managed `.claude/settings.json` (which the installer may overwrite). This directly answers "give operators a supported way to scope the matcher":
 
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -446,6 +446,91 @@ fastpath_builtin_admits() {
     return 1
 }
 
+# Read-only search-pipe-to-sink admission (#5263). Pure bash builtins, zero forks.
+#
+# The shared fastpath_structural_ok() above disqualifies EVERY pipe, which is
+# correct for the general case but produces a self-defeating false positive for
+# the single most common interactive idiom: piping a read-only search to a pager
+# or counter. `grep 'DROP TABLE' schema.sql | head` is 100% read-only — the DDL
+# phrase lives only inside grep's quoted search argument, and grep never executes
+# what it matches — yet the pipe kicked it to the full path, where SQL_DDL_PATTERN
+# (:~2819) substring-matches the literal phrase in grep's own argument and denies
+# at the catastrophic tier. The bare `grep 'DROP TABLE' schema.sql` (no pipe) was
+# already fast-pathed and allowed; this narrows the gap so the piped form matches.
+#
+# SECURITY: this is a deliberately NARROW carve-out, not a general "pipes are OK"
+# relaxation. It admits ONLY the shape
+#     <search> | <read-only-sink>
+# where:
+#   * exactly ONE pipe, and NO other shell metacharacter (; & < > ` $( newline)
+#     anywhere — so wrapper (`bash -c`), substitution (`$(...)`), and compound
+#     (`&&`/`;`) forms are untouched and keep denying via the full path exactly
+#     as before (satisfying #5263's "obfuscation still caught" requirement);
+#   * the UPSTREAM command word is a non-executing search: grep|egrep|fgrep|rg
+#     (grep/rg are already fully admitted for any args by the built-in allowlist;
+#     egrep/fgrep are the same tool). A real DDL-executing command piped the same
+#     way (`mysql -e '…' | cat`, `psql -c '…' | head`) has a non-search first
+#     token, so it is NOT admitted and still denies;
+#   * the DOWNSTREAM command word is a fixed read-only sink allowlist. head|tail|wc
+#     are already fully allowlisted (any args), so they admit with any args. cat,
+#     less, and more are NOT in the built-in allowlist — cat has a live `.ssh`/
+#     `.aws/credentials` ASK carve-out (:~3764) — so they admit ONLY as pure stdin
+#     consumers (no positional file operand). This keeps `grep x | cat ~/.ssh/id_rsa`
+#     (which would leak a key past the cat ASK) OUT of the fast path: it falls
+#     through to the full path where the cat ASK still fires.
+#
+# False NEGATIVES (declining) are always safe — they just fall through to the
+# existing slower behavior. So anything not matching the exact shape above (a
+# second pipe, an unlisted sink, a non-search upstream, a cat/less/more with a
+# file operand) declines and is handled by the full path unchanged.
+_FASTPATH_PIPE_SINKS_ANYARG=" head tail wc "     # already fully allowlisted → any args
+_FASTPATH_PIPE_SINKS_STDIN=" cat less more "     # stdin-only → no positional operand
+fastpath_grep_pipe_admits() {
+    local cmd="$1"
+    # No shell metacharacter other than a single pipe. Reject substitution,
+    # redirection, chaining, backticks, and newlines outright.
+    case "$cmd" in
+        *';'*|*'&'*|*'<'*|*'>'*|*'`'*|*'$('*) return 1 ;;
+    esac
+    [[ "$cmd" == *$'\n'* ]] && return 1
+    [[ "$cmd" == *'|'* ]] || return 1
+    local left="${cmd%%|*}"
+    local right="${cmd#*|}"
+    # Exactly one pipe: a second one (`grep a | grep b | head`) declines here and
+    # falls through to the full path (conservative — a false negative, not a hole).
+    case "$right" in
+        *'|'*) return 1 ;;
+    esac
+    local -a lt rt
+    read -ra lt <<< "$left"
+    read -ra rt <<< "$right"
+    (( ${#lt[@]} >= 1 && ${#rt[@]} >= 1 )) || return 1
+    # Upstream must be a non-executing search command.
+    case "${lt[0]}" in
+        grep|egrep|fgrep|rg) ;;
+        *) return 1 ;;
+    esac
+    # Downstream must be a read-only sink.
+    local sink="${rt[0]}"
+    if [[ "$_FASTPATH_PIPE_SINKS_ANYARG" == *" $sink "* ]]; then
+        return 0
+    fi
+    if [[ "$_FASTPATH_PIPE_SINKS_STDIN" == *" $sink "* ]]; then
+        # Pure stdin consumer only: reject any positional (non-flag) operand so a
+        # credential-file argument (`| cat ~/.ssh/id_rsa`) is NOT fast-pathed and
+        # the cat `.ssh`/`.aws` ASK carve-out still fires via the full path.
+        local i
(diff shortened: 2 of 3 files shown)


rjwalters:
Reviewed as Judge. Approving.

**Security analysis of `fastpath_grep_pipe_admits()`** (the core change in `defaults/hooks/guard-destructive-generic.sh`):

- Metacharacter guard (`;` `&` `<` `>` backtick `$(` + literal newline) is checked on the *raw* command string before any splitting, so wrapper (`bash -c "..."`), substitution (`$(...)`), and compound (`&&`/`;`) forms are rejected exactly as before — confirmed via the new `bash -c "grep <ddl> | head"` and `mysql -e <ddl> | cat` / `psql -c <ddl> | head` deny assertions.
- Exactly-one-pipe check (second `|` in the tail declines) correctly falls through to the full path rather than trying to parse a multi-stage pipeline — a conservative false-negative, not a hole. I traced the edge case where a quoted grep pattern itself contains a literal `|` (e.g. `grep 'a|b' file`) — the naive text split on the first `|` can misparse it, but the resulting malformed "sink" token essentially never collides with the sink allowlist (`head`/`tail`/`wc`/`cat`/`less`/`more`) once splitting inside a quoted arg, so it still safely declines.
- `head`/`tail`/`wc` are admitted with any args — verified against `fastpath_builtin_admits()` that these three are *already* fully allowlisted for direct (non-piped) execution, so piping to them adds no new privilege, only reshapes an existing one.
- `cat`/`less`/`more` are new territory (not in the direct-execution allowlist) and are correctly restricted to pure stdin consumers — any positional (non-flag) operand declines the carve-out, preserving the existing `cat .ssh/.aws` ASK carve-out (confirmed via the new `grep foo | cat ~/.ssh/id_rsa` still-asks assertion).
- Upstream restricted to `grep|egrep|fgrep|rg` only — a real DDL executor (`mysql -e`, `psql -c`) has a non-search first token and is correctly excluded.

**Test assertions**: both flipped assertions (the `Fast path security` grep-pipe-to-cat case, and the #5235 ask-tier grep-piped-to-cat case) are justified by the new intended behavior, have rewritten comments explaining *why* the verdict changed, and are paired with new regression coverage proving real obfuscation (wrapper+pipe, DDL executor+pipe, non-sink pipe, two-pipe chain, credential-file operand) still denies/asks as before.

**Docs**: `defaults/docs/guard-hooks.md` was updated with a new "Search-pipe carve-out (#5263)" section documenting the exact admitted shape — matches the code.

**Verification performed**:
- Read the full diff (`guard-destructive-generic.sh`, `test-guard-destructive.sh`, `guard-hooks.md`).
- Ran `bash tests/hooks/test-guard-destructive.sh` locally: **779 passed, 0 failed** — matches the PR description.
- CI: all checks green (`gh api .../check-runs`), including "Shell Test Suites (hermetic)".
- Merge state: `clean` (`gh api .../pulls/5274` → `mergeable_state: clean`).

LGTM — the fix is surgical, the obfuscation-still-caught guarantee holds under inspection, and the credential-leak guard (`cat ~/.ssh/...`) is preserved.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (and `loom:reviewing` removed) — another Judge's verdict raced mine while I was reviewing. Discarding my review; not writing any label. FWIW my independent read reached the same conclusion (approve) — the search-pipe carve-out is narrow (single pipe, fixed grep/rg/egrep/fgrep upstream, fixed read-only sink downstream, all other metacharacters still fall through to the full path), docs match the code, and CI's hermetic shell suite (which wires in tests/hooks/test-guard-destructive.sh) already passed.

Issue #2838, opened 2026-02-18
bug: merge-pr.sh does not remove loom:pr label after successful merge


rjwalters:
## Problem

After `merge-pr.sh` merges a PR, the `loom:pr` label remains on the merged PR. This is label noise — Champion would skip the PR since it's already merged, but it leaves a misleading state in the label view and could confuse tooling that queries open PRs by label.

Observed after shepherd auto-merged PR #2818 (issue #2801, 2026-02-18) with `-m` flag:

```json
{"labels": [{"name": "loom:pr", ...}], "state": "MERGED"}
```

## Expected Behavior

`merge-pr.sh` should remove the `loom:pr` label as part of its post-merge cleanup, similar to how branch deletion is handled.

## Fix

Add a `gh pr edit <PR> --remove-label "loom:pr"` step to `merge-pr.sh` after a successful merge. Should be best-effort (don't fail the script if label removal fails, since the merge already succeeded).

rjwalters:
Closing as won't fix. Labels on closed/merged items are harmless — all agents filter by open state — and skipping post-close label removal saves gh API calls. Added a note to CLAUDE.md (#138) to prevent this from being re-raised.

PR #5295, opened 2026-08-04
docs(adr): decide forge coordination decoupling — local memo, safehouse as accelerator only

### LINKED ISSUE #5057: Design: split the forge's two jobs — durable record vs real-time coordination — to decouple API cost from coordination chatter
## Thesis

**The forge is doing two jobs, and it is only good at one of them.**

- **Durable record** — issues, PRs, comments, merge history. The forge is the right
  home, and should stay authoritative.
- **Real-time coordination** — who owns this issue, who already evaluated it, what
  changed since I last looked. Loom implements this as *label mutations on a
  rate-limited HTTP API*, which is why API pressure scales with coordination chatter
  rather than with actual work done.

Separating those two concerns is the durable answer to GitHub API limits. Everything
below follows from it.

## Why not just buy more API

Verified 2026-08-03 (recorded on #4500): GitHub Enterprise Cloud raises **GraphQL to
10,000 points/hr, not 15,000** — the 15k GraphQL figure is Actions-only — and the
uplift applies only to Enterprise-owned resources, while **loom, anvil, and safehouse
are User-owned** (`rjwalters`). loom is dispatch priority 10, the busiest repo in the
fleet, and would get nothing. Buying capacity is a 2x that misses the hot path.

## Lever A — a cross-host "already evaluated" memo

The most wasteful API traffic is **repeat passes over unchanged state**, all observed
in this fleet:

- #4736 — duplicate "still blocked, no change" comments, **7x** on one issue
- #4987 (merged) — the one-off fix for exactly that, in the curator
- Champion re-verdicting deliberately-parked issues every cycle
- Concurrent curators stacking contradictory comments and clobbering bodies
- Judge's claim races — `loom:reviewing` POST, then back off: a wasted mutation pair

The ~6 label transitions in a normal issue lifecycle are **inherent, not waste**. The
waste is re-evaluating something whose inputs have not changed.

A general mechanism beats fixing each role separately: a memo of *"role X evaluated
issue Y at input-hash Z"*. **It must not live on the forge** — storing it as a label
or comment costs exactly the mutation we are eliminating. That makes it a natural
safehouse (or daemon-shared-state) resident, and unlike a per-host file it works
across hosts, which is where the duplicate passes actually come from.

## Lever B — cache-invalidation broadcast

With the per-host ETag cache (see the sibling issue on exposing
`forge_listing::list_issues_cached` to agents), validations are free but each host
still independently discovers change. A "repo X changed" broadcast lets hosts
invalidate on notice instead of by polling, and collapses the one genuinely duplicated
cost: the `200` responses that all N hosts pay simultaneously when something does
change.

## Lever C — webhooks instead of polling, using infrastructure we already own

GitHub's own guidance, verified: *"You should subscribe to webhook events instead of
polling the API for data. This will help your integration stay within the API rate
limit."*

We already have the two pieces:

- a **live public endpoint** — the Cloudflare Worker behind `dashboard.2amlogic.com`
- a **fan-out channel** — safehouse (room `loom-fleet`), plus the daemon's existing
  event bus and SSE bridge

Shape: GitHub webhook → Worker → fan out to daemons. This eliminates change
*detection* cost entirely; reads become on-demand.

Design constraints that must be respected:

- **Webhook delivery is not guaranteed.** Keep a low-frequency ETag-based
  reconciliation poll as a correctness floor — cheap, since unchanged means `304`
  means free. Event-driven must be an accelerator over a poll that still works alone.
- **This inverts the dashboard's data flow.** The dashboard is deliberately
  read-only today (#4329/#4702 lineage). Pushing forge events *to* daemons makes the
  Worker a control-plane participant, with the auth/replay/secret-handling that
  implies. That is a real architectural decision, not an implementation detail.
- **Webhook secrets** are per-repo config; the fleet spans 12 repos across a User
  account and the 2AMLogic org.

## The guardrail: what must NOT move off the forge

**Claims must stay forge-authoritative.** Today a label *is* the claim, and
`peer_claims` / `live_claim` read it. Moving the claim of record into safehouse means
inventing consensus, and the failure mode is severe: on 2026-08-03 safehouse was
`configured, unreachable` on robb-pro. Under the current design that was an
operator-interface outage. Under a safehouse-authoritative design it would have been
a fleet-wide dispatch outage or a double-claim storm.

Recommended invariant: **safehouse and webhooks are accelerators over a
forge-authoritative record that still works when they are down.** Note the forge-based
version is already imperfect (#5017: live cross-host duplicate dispatch with
collision detection enabled on all three hosts) — so the goal is to make coordination
cheaper and *then* better, not to replace a working record with a faster unproven one.

## Open decisions

1. Where does the "already evaluated" memo live — safehouse, or daemon shared state
   with a safehouse transport? What is its retention and its invalidation rule?
2. Does the Worker become a control-plane participant (Lever C), or do we stop at
   Lever B and keep the dashboard read-only?
3. What is the input-hash for "unchanged state"? Issue `updated_at` is cheap but
   coarse; a content hash is precise but needs the content.
4. Does any of this change the label protocol itself, or only the traffic around it?
   Note the standing decision **not** to clean up labels on close (#2838) was
   explicitly cost-driven — precedent that label policy and API cost are already
   coupled here.

## Acceptance criteria

- [ ] A written decision on each of the four open questions.
- [ ] Repeat evaluation of unchanged state is prevented by one general mechanism, not
      N per-role patches (#4987 being the first of the N).
- [ ] Claims remain forge-authoritative; a safehouse outage degrades coordination
      speed, never correctness, and never blocks dispatch.
- [ ] Any event-driven path retains a poll-only fallback that is correct on its own.
- [ ] Before/after GraphQL pool consumption over a fixed window is recorded.

## Non-goals

- Replacing the forge as the durable record.
- Making safehouse a hard dependency of dispatch.
- Buying API capacity as a substitute for fixing the traffic (see #4500).

## Provenance

Operator asked (2026-08-03) for ways to relieve GitHub API limits beyond buying
capacity, specifically raising REST/GraphQL split, label thrash, and "long term we
might rely more on safehouse". The REST/GraphQL split turned out to be the sibling
issue (agents bypassing the daemon's ETag cache — 84 `gh issue list` call sites);
this issue is the structural half. Parent design: #5038. Related: #4196 (safehouse as
primary operator interface, `loom:blocked`), #4702 (dashboard epic), #5017, #4736,
#4500, closed epic #4432.

rjwalters:
## Summary

Adds ADR-0014, the written decision requested by #5057's acceptance criteria.

The issue posed three levers to relieve GitHub API-quota pressure (a cross-host
"already evaluated" memo; a safehouse cache-invalidation broadcast; webhooks
into the Cloudflare Worker) and four open questions that had to be answered
before any of them could be built. This PR is the design decision, not the
implementation — it does not change any runtime code.

**The four decisions, in brief:**

1. **Where does the memo live?** Daemon-local persistent state, one store per
   host — following the existing `claim_reconciliation.rs` /
   `quarantine_reconciliation.rs` pattern. Safehouse is a transport for
   propagating memo entries between hosts (Lever B), never the store of
   record — storing it *in* safehouse would violate safehouse's own
   "never a hard dependency" contract (concretely: the 2026-08-03 outage on
   robb-pro would have erased every host's evaluation history).
2. **Does the Worker become a control-plane participant (Lever C)?** Not yet.
   Stop at Lever B. The dashboard pipeline is one-directional today
   (daemon → Worker); inverting it needs webhook-secret distribution across
   12 repos + a new inbound trust boundary + replay handling — real
   architectural cost that isn't justified until Lever A+B are deployed and
   measured to be insufficient.
3. **What's the input-hash?** A per-role content hash over the fields the
   role's decision function actually reads — never the bare `updated_at`
   (too coarse; almost any unrelated write bumps it, defeating the point of
   the memo).
4. **Does this change the label protocol?** No — only the traffic around it.
   `.github/labels.yml` and every role's label transitions are unaffected.

The PR also restates the issue's guardrail as a binding invariant (claims stay
forge-authoritative; a safehouse outage degrades speed, never correctness,
never blocks dispatch) and includes a phasing suggestion for the follow-up
implementation issues, including the before/after GraphQL measurement
acceptance criterion — which, being a runtime measurement, can only be
satisfied once the mechanism is deployed, not by this design PR.

## Test plan

- [x] Cross-checked decisions against existing repo precedent: `.loom/docs/safehouse.md`
      (degradation contract), `.loom/docs/observability.md` (current one-directional
      daemon → Worker pipeline), `.loom/docs/daemon-reference.md` § ETag-cached
      REST listing (`forge_listing`, #4428)
- [x] ADR follows the existing `docs/adr/template.md` structure and numbering
      (0014, next after 0013)
- [x] `docs/adr/README.md` index updated with the new entry
- N/A — no code changes, no tests to run

Closes #5057

### DIFF
--- a/docs/adr/0014-forge-coordination-decoupling.md
+++ b/docs/adr/0014-forge-coordination-decoupling.md
@@ -1,0 +1,258 @@
+# ADR-0014: Decouple Forge API Cost From Coordination Chatter — Local Evaluation Memo, Safehouse as Accelerator Only
+
+## Status
+
+Accepted (design decision). Implementation is phased into follow-up issues — see
+"Suggested phasing" below; this ADR does not itself change any code.
+
+## Context
+
+The forge (GitHub) does two jobs for Loom: it is the **durable record** (issues,
+PRs, comments, merge history) and it is the **real-time coordination medium**
+(who owns this issue, what changed since a role last looked). Loom implements
+coordination as label mutations on a rate-limited HTTP API, so API pressure
+scales with coordination chatter rather than with actual work done.
+
+Verified 2026-08-03 (#4500): GitHub Enterprise Cloud's GraphQL uplift
+(10,000 points/hr) applies only to Enterprise-owned resources, and `loom`,
+`anvil`, and `safehouse` are User-owned (`rjwalters`) — the busiest repo in the
+fleet gets nothing from buying capacity. The fix has to be traffic reduction,
+not quota expansion.
+
+The most wasteful traffic is **repeat passes over state that has not changed**:
+
+- #4736 — duplicate "still blocked, no change" comments, 7x on one issue
+- #4987 (merged) — a one-off, curator-only fix for exactly that pattern
+- Champion re-verdicting deliberately-parked issues every cycle
+- concurrent curators stacking contradictory comments on the same issue
+- Judge's claim races (`loom:reviewing` POST, then back off — a wasted mutation
+  pair)
+
+The ~6 label transitions in a normal issue lifecycle are inherent, not waste.
+The waste is re-evaluating something whose inputs have not changed, independently,
+per role, per host.
+
+Three levers were proposed (#5057):
+
+- **Lever A** — a memo of "role X evaluated issue Y at input-hash Z" so a role
+  can skip re-evaluating unchanged state.
+- **Lever B** — a cache-invalidation broadcast ("repo X changed") so hosts
+  invalidate on notice instead of by polling.
+- **Lever C** — GitHub webhooks → the existing Cloudflare Worker → fan-out to
+  daemons, replacing polling-based change *detection* with push.
+
+Four questions had to be answered before any of this could be built. This ADR
+answers them.
+
+## Decision
+
+### 1. Where does the memo live?
+
+**The memo lives in daemon-local persistent state, one store per host.
+Safehouse is a transport for propagating memo entries between hosts, never the
+store of record, and the forge is never the store either.**
+
+This follows the pattern the daemon already uses for `worktree_reaper.rs`,
+`quarantine_reconciliation.rs`, and `claim_reconciliation.rs`: periodic,
+host-owned state that a Rust daemon subsystem maintains directly, not a role
+agent's job and not a forge mutation. It also follows safehouse's existing
+degradation contract (`.loom/docs/safehouse.md`): safehouse is documented as a
+best-effort side-channel with **zero hard dependency** — "Loom never blocks a
+sweep on safehouse" — so a store that safehouse could ever be the sole holder
+of would violate that contract the moment the persona is unreachable (recall
+2026-08-03: safehouse was `configured, unreachable` on robb-pro).
+
+Concretely: each daemon keeps a small local table keyed by
+`(role, issue_or_pr, input_hash) -> {result, timestamp}`. When safehouse is
+enabled, a host that writes a new memo entry also emits a lightweight
+broadcast so peer hosts can populate their own local table without paying
+their own evaluation cost — this is Lever B, riding the existing fan-out
+channel and its established envelope-routing / degrade-to-`warn!` posture
+(`.loom/docs/safehouse.md` § Configuration). Without safehouse, every host
+still gets full single-host benefit (the #4736 shape — repeat passes *on one
+host* — is fixed by the local memo alone); only the cross-host duplicate-pass
+case degrades to "each host evaluates once instead of zero times," which is
+strictly the pre-existing behavior, not a regression.
+
+**Retention and invalidation.** Retention is a housekeeping concern, not a
+correctness one — see decision 3. A memo entry is correctness-safe to keep
+indefinitely because it self-invalidates: on read, the role recomputes the
+current input-hash and compares it to the memoized one; any mismatch is
+treated as "not yet evaluated," full stop. There is no proactive
+invalidation message to get wrong. Housekeeping bounds the table's size with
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

This is a docs-only design decision (ADR-0014) answering the four open questions posed in #5057. Reviewed for internal consistency and grounding against existing repo precedent:

- **Memo location decision** is consistent with `.loom/docs/safehouse.md`'s documented degradation contract ("never a hard dependency") — correctly avoids making safehouse the store of record, citing the real 2026-08-03 `robb-pro` outage as evidence.
- **Lever C deferral** is consistent with `.loom/docs/observability.md`'s documented one-directional (daemon → Worker) pipeline and its stated "read-only invariant" — the ADR correctly identifies that inverting this needs real new infrastructure (webhook secrets across 12 repos, a new inbound trust boundary) not yet justified by measurement.
- **Input-hash decision** correctly leverages the existing ETag-cached `forge_listing` (#4428) as a free pre-filter while avoiding the coarse `updated_at`-as-hash trap.
- **Label protocol** decision is consistent with the standing #2838 precedent (label policy already treated as decoupled from API-cost work).
- The binding invariants (forge-authoritative claims, poll-only fallback) restate — not relax — the guardrails #5057 itself specified.

No code changes, so no functional test surface. All docs-gating CI checks pass (AGENTS.md Sync, CLAUDE.md Line Budget, Docs/Defaults Parity). `docs/adr/README.md` index entry is well-formed and matches the existing style. Numbering (0014) doesn't collide with any other in-flight ADR. Merge state CLEAN.

## Test Execution

No "## Test Plan" section in the PR description (appropriate for a docs-only ADR) — noting absence per policy, not blocking.

Issue #4451, opened 2026-07-29
ci: tests/hooks/* guard suite is not wired into any CI workflow — guard regressions ride to main invisibly


rjwalters:
## Context
Judge finding on PR #4421 (2026-07-29, non-blocking note): the guard-hook test suite `tests/hooks/test-guard-destructive.sh` (482+ assertions, grew by 8 in that PR) is not wired into any CI workflow — it runs only when someone invokes it locally. PR #4421 changed security-relevant guard behavior (`~/` expansion in write-target resolution) and shipped with its own suite green locally but zero CI execution.

The existing CI jobs (Rust Unit Tests, Rust Linting & Build Check, Installer Integration Tests) path-filter shell-only guard changes into "skipping", so a guard regression would ride to main invisibly.

## Suggested acceptance criteria
- [ ] A CI workflow (new job or extension of Installer Integration Tests) runs `tests/hooks/` on PRs touching `defaults/hooks/**`, `tests/hooks/**`, or the installed `.loom/hooks/**` copies
- [ ] The job fails the check run on any assertion failure
- [ ] Runtime is kept reasonable (the suite is pure bash; expected well under a few minutes)
- [ ] Path filters documented in the workflow file

---

### Verified current state

- **Confirmed unwired**: no file under `.github/workflows/` or `scripts/` references `tests/hooks/`, `defaults/hooks/tests/`, or any `test-guard-*` suite (repo-wide grep, zero hits).
- **The gap is larger than filed** — there are TWO hook-suite locations, both unwired:
  - `tests/hooks/` — 3 suites: `test-guard-destructive.sh` (519 assertions), `test-guard-destructive-dispatcher.sh`, `test-guard-loom-workflow.sh`
  - `defaults/hooks/tests/` — 5 suites: `test-guard-background-subagents.sh`, `test-guard-loom-workspace.sh`, `test-guard-worktree-paths.sh`, `test-methodology-inject.sh`, `test-skill-router.sh`
- **Path-filter nuance** (refines the original context): `defaults/hooks/**` changes DO trigger the Installer Integration Tests job (the `scripts` filter group includes `defaults/**`) — but that job never runs the guard suites, so the regression still rides through. Changes touching only `tests/hooks/**` trigger **nothing at all**.
[...]

PR #5297, opened 2026-08-04
fix(guard): close two-hop heredoc-variable indirection gap in merge-redirect check

### LINKED ISSUE #5172: Guard false positive: merge-redirect guard denies gh api -f body=VAR referencing a heredoc that merely quotes the disallowed phrase

rjwalters:
## Summary

The `loom:gh-pr-merge-redirect` guard still denied `gh api ... -f body="$VAR"`
calls where `$VAR` was assigned earlier in the same command from a `cat`
heredoc whose body only *quotes* the disallowed phrase as prose (e.g. a PR
review comment describing test cases), never actually invokes it. Neither
existing masking pass recognized this shape: `mask_data_flag_values` only
matched `--flag value` shapes, not `gh api`'s `-f field=value` syntax, and
`mask_cat_heredoc_bodies` only masked a heredoc captured *directly* by a
known text-data flag — not one assigned to a shell variable and referenced
later via that variable (a two-hop indirection).

## Changes

- Extend `mask_cat_heredoc_bodies`'s `capre` and `mask_data_flag_values`'s
  `re` to also recognize `gh api ... -f <field>=<value>` for the known
  text-bearing fields (`body`, `message`, `comment`, `title`, `notes`,
  `search`) — covers both a literal `-f body="..."` value and a heredoc
  captured directly by `-f body="$(cat <<'EOF' ... EOF)"`.
- Add `mask_var_assigned_heredoc_bodies()`: masks a heredoc's body at its
  point of *definition* when it's assigned to a shell variable
  (`VAR="$(cat <<'EOF' ... EOF)"`), but **only** when every later reference
  to that variable (`$VAR` / `${VAR}`) elsewhere in the command is itself
  confined to a known non-executing flag/field value (the same allowlist as
  above). Any unconfined reference — `eval "$VAR"`, the bare variable used
  as a command, piped into a shell — leaves the heredoc body **completely
  unmasked**, so a genuine two-hop bypass attempt still denies exactly as
  before. A variable's own heredoc body is excluded from the confinement
  scan (self-mention isn't a live later reference).
- Wire the new function into the `GH_PR_MERGE_SCAN_TEXT` masking chain.
- Sync the installed `.loom/hooks/guard-loom-workflow.sh` copy to match
  `defaults/hooks/guard-loom-workflow.sh` byte-for-byte (this repo's own
  self-install dogfooding convention enforced by CI, #4451 — verified
  against the pre-#5172 baseline and prior guard-fix PRs #5109/#5115/#5160,
  which updated both copies in the same commit).

| The exact #5172 reproduction (`-f body="$BODY"` where `$BODY` is a heredoc quoting the phrase as prose) now allows | ✅ | New test `Allow gh api -f body=$VAR where $VAR is a heredoc quoting the phrase as prose (#5172)`; also manually reproduced and confirmed allow before writing the formal test |
| `gh api -f <field>=<value>` masking covers the same known text-bearing field list as existing `--flag value` masking, for both a literal value and a directly-captured heredoc | ✅ | New tests for literal `-f body="..."`, direct heredoc capture via `-f body=`, and the `-f message=` field |
| A literal same-position invocation of the disallowed phrase still denies (no widening) | ✅ | Pre-existing `Block gh pr merge` / `Block gh pr merge --squash` tests still pass |
| The fix does not silently widen the guard into allowing an actual invocation reached via the same two-hop shape (heredoc → variable → later use) | ✅ | Three new control tests: heredoc-assigned variable later `eval`'d, referenced bare as a command, and referenced twice (once safely, once unsafely) — all still deny |

## Test Plan

- `./tests/hooks/test-guard-loom-workflow.sh` — 49/49 pass (7 new #5172 cases: 4 allow, 3 deny-control), run against the canonical `defaults/hooks/` source.
- `./tests/hooks/test-guard-destructive.sh` — 779/779 pass (unrelated guard, confirms no cross-guard regression).
- `./tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 pass.
- `shellcheck defaults/hooks/guard-loom-workflow.sh` — clean (no findings).
- `bash -n` on both changed hook files and the test file — clean.
- Manually reproduced the exact issue scenario end-to-end (heredoc-assigned `$BODY` quoting the phrase, then `gh api -f body="$BODY"`) before formalizing it as a test — confirmed allow.

Closes #5172

### DIFF
--- a/.loom/hooks/guard-loom-workflow.sh
+++ b/.loom/hooks/guard-loom-workflow.sh
@@ -177,7 +177,15 @@ mask_cat_heredoc_bodies() {
         # VALUE of a known non-executing text-data flag. `capre` must match the
         # tail of the opener-line text that sits immediately before the `cat`
         # token (see the confinement check below).
-        capre = "(^|[ \t])(-m|--message|--body|--notes|--title|--comment|--search)[ \t]*=?[ \t]*(" DQ "|" SQ ")?[ \t]*([$][(]|" BT ")[ \t]*$"
+        #
+        # Also recognizes `gh api ... -f <field>=` for known text-bearing
+        # fields (issue #5172): the field syntax of `gh api` (`-f
+        # body=<value>`) is a DIFFERENT shape from the `--body <value>` flags
+        # above -- a
+        # two-token `-f <field>=` prefix, not a single `--<flagname>` token --
+        # so it needs its own alternative rather than falling out of the
+        # existing flag list.
+        capre = "(^|[ \t])((-m|--message|--body|--notes|--title|--comment|--search)[ \t]*=?|-f[ \t]+(body|message|comment|title|notes|search)=)[ \t]*(" DQ "|" SQ ")?[ \t]*([$][(]|" BT ")[ \t]*$"
     }
     { lines[NR] = $0 }
     END {
@@ -249,6 +257,173 @@ mask_cat_heredoc_bodies() {
     }'
 }
 
+# =============================================================================
+# TWO-HOP HEREDOC-VARIABLE INDIRECTION MASKING (issue #5172)
+#
+# mask_cat_heredoc_bodies() above only recognizes a cat-heredoc captured
+# DIRECTLY by a known text-data flag/field, e.g. `-m "$(cat <<'EOF' ... EOF)"`.
+# It does NOT recognize the equally common two-STEP idiom of assigning that
+# heredoc to a shell variable first, then referencing the variable later:
+#
+#   BODY="$(cat <<'EOF'
+#   ...prose that quotes the disallowed phrase as a documented example...
+#   EOF
+#   )"
+#   gh api "repos/OWNER/REPO/issues/N/comments" -f body="$BODY"
+#
+# Here the literal phrase text lives in the heredoc body at DEFINITION time;
+# the LATER reference is only the variable name ($BODY), never the phrase
+# itself. Raw substring scanning still catches the phrase living in the
+# heredoc body, even though nothing in the command actually executes it — the
+# false-positive class this function closes.
+#
+# mask_var_assigned_heredoc_bodies() masks such a heredoc's body at its point
+# of DEFINITION, but ONLY when it can prove every LATER reference to that same
+# variable ($VAR / ${VAR}) elsewhere in the command is itself confined to a
+# known non-executing text-data flag/field value -- the same allowlist as
+# mask_cat_heredoc_bodies/mask_data_flag_values (-m/--message/--body/--notes/
+# --title/--comment/--search, or `gh api -f <field>=`). If ANY later
+# reference to the variable falls OUTSIDE that confined context -- `eval
+# "$VAR"`, `bash -c "$VAR"`, the bare variable used as a command, or simply no
+# recognizable safe usage -- the heredoc body is left COMPLETELY UNMASKED, so
+# a genuine two-hop bypass (assign `gh pr merge 123` to a variable via
+# heredoc, then `eval` it) still denies exactly as before. Masking only ever
+# narrows what this ONE check misses; it never widens it -- same invariant as
+# every other masking function in this file.
+#
+# A variable reference that falls INSIDE a candidate heredoc's OWN body (the
+# variable mentioning its own name as prose, e.g. describing this very fix) is
+# excluded from the confinement scan -- it is inert text, not a later live
+# reference -- by blanking every candidate's body span before scanning.
+# =============================================================================
+mask_var_assigned_heredoc_bodies() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)
+        DQ = sprintf("%c", 34)
+        # A bare shell variable assignment (`VAR=`/`VAR="`/`VAR=$SQ`) directly
+        # capturing `$(cat`. Deliberately distinct from capre in
+        # mask_cat_heredoc_bodies() above: a recognized text-data FLAG always
+        # starts with `-`, which this identifier-only pattern can never match,
+        # so the two confinement modes never collide.
+        varassign_re = "(^|[ \t;&|(])[A-Za-z_][A-Za-z0-9_]*=(" DQ "|" SQ ")?[ \t]*[$][(][ \t]*$"
+        safe_flag  = "(-m|--message|--body|--notes|--title|--comment|--search)[ \t]*=?[ \t]*(" DQ "|" SQ ")?$"
+        safe_field = "-f[ \t]+(body|message|comment|title|notes|search)=(" DQ "|" SQ ")?$"
+        ncand = 0
+    }
+    { lines[NR] = $0 }
+    END {
+        nl = NR
+        for (i = 1; i <= nl; i++) {
+            line = lines[i]
+            off = 1
+            while (1) {
+                p = index(substr(line, off), "<<")
+                if (p == 0) break
+                p = off + p - 1
+                off = p + 2
+                pre = substr(line, 1, p - 1)
+                if (pre !~ /(^|[^A-Za-z0-9_])cat[ \t]*$/) continue
+                before_cat = pre
+                sub(/cat[ \t]*$/, "", before_cat)
+                if (match(before_cat, varassign_re) == 0) continue
+                seg = substr(before_cat, RSTART, RLENGTH)
+                if (match(seg, /[A-Za-z_][A-Za-z0-9_]*/) == 0) continue
+                vn = substr(seg, RSTART, RLENGTH)
+                start = p + 2
+                if (substr(line, start, 1) == "-") start++
+                while (substr(line, start, 1) == " " || substr(line, start, 1) == "\t") start++
+                qc = substr(line, start, 1)
+                if (qc != SQ && qc != DQ) continue
+                start++
+                wordend = start
+                while (substr(line, wordend, 1) ~ /^[A-Za-z0-9_]$/) wordend++
+                if (wordend <= start) continue
+                if (substr(line, wordend, 1) != qc) continue
+                delim = substr(line, start, wordend - start)
+                rest = substr(line, wordend + 1)
+                if (rest ~ /[^ \t]/) continue
+                closeat = 0
+                for (j = i + 1; j <= nl; j++) {
+                    trimmed = lines[j]
+                    sub(/^[ \t]+/, "", trimmed)
+                    if (trimmed == delim) { closeat = j; break }
+                }
+                if (closeat == 0) continue
+                ncand++
+                cand_var[ncand] = vn
+                cand_bstart[ncand] = i + 1
+                cand_bend[ncand] = closeat - 1
+                break
+            }
+        }
+        if (ncand == 0) {
+            out = lines[1]
+            for (i = 2; i <= nl; i++) out = out "\n" lines[i]
+            printf "%s", out
+            exit
+        }
+        # Build a scan buffer with every CANDIDATE heredocs own body blanked
+        # (never masked to X -- a plain blank cannot itself spell "$VAR"),
+        # so a variables OWN prose mentioning its own name never gates its
+        # own masking decision.
+        scanbuf = ""
+        for (i = 1; i <= nl; i++) {
+            ln = lines[i]
+            is_body = 0
+            for (c = 1; c <= ncand; c++) {
+                if (i >= cand_bstart[c] && i <= cand_bend[c]) { is_body = 1; break }
+            }
+            if (is_body) { gsub(/./, " ", ln) }
+            scanbuf = scanbuf (i > 1 ? "\n" : "") ln
+        }
+        buflen = length(scanbuf)
+        for (c = 1; c <= ncand; c++) {
+            vn = cand_var[c]
+            vlen = length(vn)
+            confined = 1
+            pos = 1
+            while (pos <= buflen) {
+                rem = substr(scanbuf, pos)
+                i1 = index(rem, "$" vn)
+                i2 = index(rem, "${" vn "}")
+                if (i1 == 0 && i2 == 0) break
+                use2 = (i2 > 0 && (i1 == 0 || i2 <= i1))
+                if (use2) {
+                    abspos = pos + i2 - 1
+                    mlen = 3 + vlen
+                } else {
+                    abspos = pos + i1 - 1
+                    mlen = 1 + vlen
+                    nextch = substr(scanbuf, abspos + mlen, 1)
+                    if (nextch ~ /^[A-Za-z0-9_]$/) {
+                        pos = abspos + 1
+                        continue
+                    }
+                }
+                prefix = substr(scanbuf, 1, abspos - 1)
+                if (prefix !~ safe_flag && prefix !~ safe_field) {
+                    confined = 0
+                    break
+                }
+                pos = abspos + mlen
+            }
+            cand_mask[c] = confined
+        }
+        for (c = 1; c <= ncand; c++) {
+            if (cand_mask[c] != 1) continue
+            for (j = cand_bstart[c]; j <= cand_bend[c]; j++) {
+                body = lines[j]
+                gsub(/./, "X", body)
+                lines[j] = body
+            }
+        }
+        out = lines[1]
+        for (i = 2; i <= nl; i++) out = out "\n" lines[i]
+        printf "%s", out
+    }'
+}
+
 # Mask the quoted VALUE of known non-executing, text-only flags used by
 # git/gh subcommands: -m/--message, --body, --notes, --title, --comment,
 # --search. A near-duplicate of strip_literal_text() above (which is used
@@ -258,12 +433,17 @@ mask_cat_heredoc_bodies() {
 # a span that still contains an unmasked `$(`/backtick (e.g. real command
 # substitution, not yet neutralized by the heredoc pass above) is left
 # completely untouched.
+#
+# Also recognizes `gh api ... -f <field>=<value>` for known text-bearing
+# fields (issue #5172): `gh api`'s field syntax is `-f key=value`, a
+# two-token shape distinct from the single `--<flagname> value` flags above,
+# so it needs its own alternative in the same regex.
 mask_data_flag_values() {
     printf '%s' "$1" | awk '
     BEGIN {
         SQ = sprintf("%c", 39)
         DQ = sprintf("%c", 34)
-        re = "(^|[ \t\n])(--message|--body|--notes|--title|--comment|--search|-m)[ \t]*=?[ \t]*(" \
+        re = "(^|[ \t\n])((--message|--body|--notes|--title|--comment|--search|-m)[ \t]*=?|-f[ \t]+(body|message|comment|title|notes|search)=)[ \t]*(" \
              DQ "[^" DQ "]*" DQ "|" SQ "[^" SQ "]*" SQ ")"
         buf = ""
     }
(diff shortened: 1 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T15:52:40Z) — standing down without reclaiming. Not stomping.

rjwalters:
Thorough review of the guard-hook fix, including an independent bypass attempt against the built script (not just reading the diff). Found a real gap in the confinement check that needs to be closed before this can merge.

## What works

- The stated #5172 repro (heredoc-assigned `$BODY` quoting the phrase as prose, then `gh api -f body="$BODY"`) does correctly allow -- verified directly against `defaults/hooks/guard-loom-workflow.sh` (not just via the test suite).
- `.loom/hooks/guard-loom-workflow.sh` and `defaults/hooks/guard-loom-workflow.sh` are byte-identical (self-install parity intact).
- `git merge-tree --write-tree origin/main HEAD` is conflict-free; mergeStateStatus resolved to CLEAN once CI finished, and all CI checks now pass (`Shell Test Suites (hermetic)` was pending at claim time, now passes).
- All 49 tests in `tests/hooks/test-guard-loom-workflow.sh` pass, including the 3 new deny-control cases (`eval "$VAR"`, bare `$VAR` as a command, mixed safe+unsafe reference).

## Blocking issue: confinement scanner in `mask_var_assigned_heredoc_bodies` has an exploitable blind spot

The confinement scanner only recognizes the exact literal substrings `$VAR` / `${VAR}` as a "later reference" to the heredoc-assigned variable. Any OTHER bash parameter-expansion form of the same variable is invisible to it -- and an invisible reference defaults to `confined=1` (masked/hidden), because the scan loop only sets `confined=0` when it *finds* an unconfined match. If it never finds a match at all, the heredoc body gets masked even though the variable is genuinely dereferenced and executed elsewhere in the command.

Verified directly against `defaults/hooks/guard-loom-workflow.sh` (JSON piped to the script exactly as Claude Code's PreToolUse hook does), comparing against the pre-fix `main` version of the same file as a control:

```bash
BODY="$(cat <<'EOF'
gh pr merge 123
EOF
)"
eval "${BODY:0:100}"
```

- **`main` (pre-fix) guard**: denies, as expected (the heredoc body is fully unmasked pre-fix, so the literal phrase is still visible to the final grep regardless of how it's referenced later).
- **This PR's guard**: **allows** (empty stdout, exit 0) -- yet at real execution time, `eval "${BODY:0:100}"` evaluates the full heredoc content (the phrase is well under 100 chars), i.e. performs a genuine, unredirected invocation of the blocked command. This is exactly the "widen deny into allow for a real invocation" failure mode the PR's own description says the fix must not introduce.

This isn't a one-off -- the same "reference not found -> assume confined" gap reproduces for essentially any parameter-expansion form other than the two bare literal forms already covered:

- `eval "${BODY#}"` (a functional no-op expansion, semantically identical to `eval "$BODY"`) -- **allows**
- `eval "${BODY:-}"` (default-value expansion, semantically identical to `eval "$BODY"` when `BODY` is set) -- **allows**
- `REF=BODY; eval "${!REF}"` (indirect/nameref expansion) -- **allows**

All four were run against the built `defaults/hooks/guard-loom-workflow.sh` script directly (not simulated), and all four differ from the pre-fix `main` guard, which correctly denies every one of them.

By contrast, `printf %s "$BODY" | bash` (bare `$BODY` substring) is still correctly caught and denies -- confirming the gap is specific to the confinement scanner's substring-only detection (`index(rem, "$" vn)` / `index(rem, "${" vn "}")`), not a wholesale failure of the new function.

## Why this matters here specifically

This is the guard whose entire job is to stop a real merge-redirect-bypassing invocation from slipping past the required `merge-pr.sh` redirect. `${VAR:offset:len}`, `${VAR#pattern}`, `${VAR:-default}`, and `${!VAR}` are all common, unremarkable bash syntax (e.g. truncating a body for a preview, providing a default) -- not obscure ones an attacker would need special knowledge to reach. A confused or adversarial agent reaching for `eval "${BODY:-}"` instead of `eval "$BODY"` -- a trivial, natural variation -- defeats the guard entirely with this PR as written.

## Suggested direction (not prescriptive)

The confinement scan needs to treat *any* occurrence of `${VAR` (brace-opened reference to the candidate variable name, regardless of what follows before the closing brace) as a live reference requiring confinement -- not just the exact closed form `${VAR}` -- in addition to the existing bare `$VAR` check. A conservative fix: if a literal safe-flag/safe-field prefix can't be proven for every occurrence of the `$VAR` frontier (`$` immediately followed by the variable name, in either `$VAR` or `${VAR...` form), leave the body unmasked. Given how easy this was to find, please also add a couple of parameter-expansion-form deny-control tests alongside the existing three before landing.

## Verdict

Changes requested -- the two-hop indirection false-positive fix is well-targeted and the existing tests are solid, but the confinement check has an exploitable blind spot for non-literal variable dereferences that lets a real merge invocation slip past undetected. This needs to be closed (and covered by a regression test) before merge, given this is a security-relevant guard.

rjwalters:
Sweep (issue #5172, run sweep-20260804T161416Z-2215641-69db5101): reclaiming orphaned `loom:treating` claim (applied 2026-08-04T16:00:57Z by a prior sweep's Doctor).

Verified orphaned rather than waiting out the standard 60-minute TTL: the claiming sweep's orchestrator process (PID 1200775, from checkpoint task_id `sweep-20260804T152852Z-1200775-1c506f9d`) is confirmed dead via process-table inspection, its `.loom/sweep-run/` registry entry is gone, and no commit has landed on this branch since 2026-08-04T15:50:15Z — before the judge-rejected checkpoint (16:00:22Z) was even written. No work happened after the claim was taken; it is a mid-phase-death orphan (headless `-p` turn-end kill, per `.loom/docs/troubleshooting.md`), not a live Doctor.

Proceeding to fix the Judge's confinement-scanner gap.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:18:53Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved — the follow-up fix addresses the gap I flagged in my prior review.

## Verification

**Diff review** (`5b40b294..25dec20d`): the confinement scanner in `mask_var_assigned_heredoc_bodies` (`defaults/hooks/guard-loom-workflow.sh`) now searches for `${VAR` as a prefix (not just the closed `${VAR}` literal), catching every parameter-expansion variant — `${VAR:0:100}`, `${VAR#pat}`, `${VAR:-def}`, `${VAR/a/b}`, etc. It also tracks `nref` (references actually observed) and only masks when `confined == 1 && nref > 0` — a variable with zero detected references (e.g. reachable only via `${!REF}` indirection) is now left **unmasked** (fails safe / denies) instead of defaulting to "confined."

**Test suite**: ran `./tests/hooks/test-guard-loom-workflow.sh` locally — all 54 tests pass, including the 5 new #5297 fixtures (substring `${VAR:0:100}`, prefix-removal `${VAR#}`, default-value `${VAR:-}`, indirect `${!REF}`, plus the confined-context control). `defaults/hooks/guard-loom-workflow.sh` and `.loom/hooks/guard-loom-workflow.sh` are byte-identical.

**Independent bypass attempts** (built the script directly, not just reading the diff) — beyond the PR's own fixtures, I fed the guard heredoc-assigned variables later `eval`'d through forms not explicitly tested in the PR:
- `${VAR%%pat}` (suffix removal) — denied
- `${VAR^^}` (case modification) — denied
- `${VAR/foo/bar}` (substitution) — denied
- `${VAR[0]}` (array-index-like) — denied
- `eval "$(echo "$VAR")"` (command-substitution-wrapped reference) — denied
- `${VAR:+$VAR}` (alternate-value expansion) — denied
- Two-hop-squared: `BODY2="${BODY}"; eval "$BODY2"` — denied
- Control: `gh api ... -f body="${BODY:0:200}"` where `$BODY` is a heredoc quoting the phrase as inert prose — allowed (confirms the original #5172 false-positive fix still holds for a param-expansion reference form)

All bypass attempts correctly deny; the confined control correctly allows. I did not find any remaining gap in the confinement scanner.

## CI / merge state

- All CI checks green (`Shell Test Suites (hermetic)` was in-progress at claim time; polled to completion — success).
- `mergeStateStatus`: CLEAN.

LGTM — approving.

Issue #5276, opened 2026-08-04
Wire or delete orphaned scripts/test-*.sh suites (no CI/build-gate runner)


rjwalters:
## Problem Statement

A hygiene-pass orphan scan (2026-08-04, verified against `origin/main` at curation time) found that five `scripts/test-*.sh` suites have zero CI or build-gate invocations: `scripts/test-install-local-mode.sh`, `scripts/test-install-quick-gitignore-stash.sh`, `scripts/test-label-setup.sh`, `scripts/test-daemon-scripts.sh`, and `scripts/test-migrate-consumer.sh`. (For contrast, `scripts/test-installer.sh` and `scripts/test-changelog.sh` — two similarly-named siblings — ARE wired, into `.github/workflows/ci.yml:224` and `defaults/scripts/build-gate.sh:160` respectively, so this scan is not "all `scripts/test-*.sh` are orphaned," only these five.)

This is a sub-issue of #5271 (decomposed because the full orphan cleanup spans three independent file groups plus a guard-script change — too large for one Builder PR). See #5271 for the full context and the sibling sub-issues.

One of the five already carries a documented retirement signal: `docs/migration/daemon-state-consumers.md` lists `scripts/test-daemon-scripts.sh` under "Shell scripts to retire" (lines 222, 351) as a **test-only script for daemon shell scripts that are themselves being retired** — this one is a strong deletion candidate rather than a wiring candidate; verify the scripts it tests are in fact gone/retired before deleting.

`scripts/test-migrate-consumer.sh` is referenced descriptively (not invoked) from `.loom/docs/machine-dispatcher.md:495` and `defaults/docs/machine-dispatcher.md:495` as "fixture-based coverage" for the consumer-migration flow — the doc treats it as if it runs somewhere, but no workflow or script actually executes it. Resolve this discrepancy: either wire it in, or correct the doc's claim if it's deleted.

## Acceptance Criteria

- [ ] Each of the 5 suites listed below is either wired into CI/build-gate (a job or script actually executes it) or deleted with a stated rationale — no third state.
- [ ] `scripts/test-daemon-scripts.sh`: verify whether the daemon shell scripts it tests have actually been retired (per `docs/migration/daemon-state-consumers.md`) before deciding wire vs. delete.
[...]

PR #5298, opened 2026-08-04
fix(install): stash root CLAUDE.md before --quick reinstall's chained uninstall

### LINKED ISSUE #5289: install.sh --quick reinstall: --confirm-reinstall staleness + possible silent CLAUDE.md in-block edit loss
## Problem Statement

While deciding wire-vs-delete for `scripts/test-install-quick-gitignore-stash.sh`
(#5276), running the suite against current `main` surfaced two distinct problems
with `install.sh --quick`'s reinstall/stash-pop machinery (issues #3588/#3611/#3663):

1. **Stale test invocations**: every reinstall call in the suite uses
   `install.sh -y --quick "$REPO"` with no `--confirm-reinstall`. Since #4188
   landed, a non-interactive reinstall over an existing `.loom/` install
   requires `--confirm-reinstall` explicitly — `-y --quick` alone now exits 1
   immediately ("Existing Loom installation detected ... refusing to run a
   non-interactive reinstall without explicit acknowledgement"). The suite was
   never updated after #4188, so most of its "reinstall" scenarios never
   actually exercise the reinstall code path; many assertions pass only
   vacuously (the file is untouched because the reinstall errored out before
   touching anything).

2. **Possible silent data-loss regression** (once the flag above is fixed):
   the suite's "genuine in-block CLAUDE.md conflict" scenario (#3663) expects
   that when a user has an uncommitted edit to a line *inside* the
   `<!-- BEGIN LOOM ORCHESTRATION -->` marker block, and the block's committed
   content also diverges from what a reinstall would regenerate, the installer
   surfaces a named conflict, preserves a stash, and does NOT silently discard
   the user's edit. Reproduced directly (see below): with `--confirm-reinstall`
   added, the reinstall now completes ("Quick reinstallation complete!") but
   the user's uncommitted in-block edit is silently overwritten by the fresh
   pointer text — no conflict message, no stash, no "User changes restored"
   message either. Whether this is an intentional simplification (the root
   `CLAUDE.md` block shrank from a full guide to a short pointer paragraph at
   some point after #3663 landed, so perhaps in-block edits are no longer
   considered worth preserving) or a genuine regression needs a maintainer
   decision — it was out of scope to judge from a test-wiring pass.

## Reproduction (finding 2)

```bash
INSTALL="$(pwd)/install.sh"   # from a loom checkout with target/release built
W=$(mktemp -d) && git init -q "$W" && git -C "$W" config user.email t@e.com && git -C "$W" config user.name T
printf '# My Project\n\nUSER SECTION LINE\n' > "$W/CLAUDE.md"
echo hi > "$W/README.md"; git -C "$W" add -A; git -C "$W" commit -qm init
"$INSTALL" -y --quick "$W"                       # first install
git -C "$W" add -A; git -C "$W" commit -qm "loom install"
# diverge the committed in-block pointer text, then make an uncommitted edit to the same line
perl -0pi -e 's/AI-powered development orchestration/STALE COMMITTED BLOCK VALUE/' "$W/CLAUDE.md"
git -C "$W" add -A; git -C "$W" commit -qm "diverge in-block line"
perl -0pi -e 's/STALE COMMITTED BLOCK VALUE/USER IN-BLOCK EDIT/' "$W/CLAUDE.md"
"$INSTALL" -y --quick --confirm-reinstall "$W"   # reinstall
cat "$W/CLAUDE.md"       # "USER IN-BLOCK EDIT" is gone, no conflict reported
git -C "$W" stash list   # empty -- nothing to recover
```

## Suspected Cause (unverified)

Read against `install.sh` on `origin/main` at `fb20d279` (2026-08-04, current tip as of this curation pass — re-verify line numbers before relying on them, they drift) — needs verification by a Builder with `set -x`/instrumentation, not stated as settled:

The in-block-edit guard at `install.sh:1290-1293` (inside the "Restoring stashed user changes" block, starting `install.sh:1246`) is supposed to detect exactly this case and skip the HEAD-reset+reapply path (`reapply_loom_claude_md_block`, `install.sh:555-593`, using the marker-extraction helper `_emit_loom_claude_block`, `install.sh:599-605`) when the user's stashed edit lands inside the marker block, falling back to a plain `git stash pop --index` that should surface a genuine 3-way conflict instead:

```bash
_head_block="$(git -C "$target_path" show HEAD:CLAUDE.md 2>/dev/null | _emit_loom_claude_block)" || true
[[ -n "$_head_block" ]] || continue
_stashed_block="$(git -C "$target_path" show 'stash@{0}:CLAUDE.md' 2>/dev/null | _emit_loom_claude_block)" || true
[[ "$_stashed_block" == "$_head_block" ]] || continue
```

In the repro, `_head_block` (from `HEAD:CLAUDE.md`, still the "diverge in-block line" commit) should contain `STALE COMMITTED BLOCK VALUE`, and `_stashed_block` (from `stash@{0}:CLAUDE.md`, the working-tree snapshot at stash time) should contain `USER IN-BLOCK EDIT` — different strings, so the guard should `continue` (skip reset) and CLAUDE.md should never reach `reapply_loom_claude_md_block`. Since the observed behavior is a silent, clean overwrite instead, one of these is likely wrong and needs to be checked directly (e.g. by adding a temporary `echo` of `$_head_block`/`$_stashed_block` and `git -C "$W" stash show` in the repro above):

- **`_emit_loom_claude_block` mis-extracts one side** (e.g. a marker-detection edge case), making `_head_block == _stashed_block` when they should differ — this would route CLAUDE.md into the reset+reapply path and reproduce the exact symptom (working tree reset to HEAD, popped cleanly since the base now matches, then `reapply_loom_claude_md_block` splices the freshly-rendered block over the user's restored edit).
- **CLAUDE.md never entered the stash at all** — if `stash-scope.sh`'s dirty ∩ ownership intersection (`install.sh:1054-1080`, `scripts/install/stash-scope.sh`) missed CLAUDE.md for some reason, `stash@{0}:CLAUDE.md` would read stale/wrong data, or the whole reconcile block might no-op for this file, again leaving the freshly-rendered `init` output in place with nothing to conflict against.
- **The 3-way pop conflict-detection path itself has a gap** — worth confirming `git stash pop --index` genuinely returns non-zero on a manually-reproduced 3-way mismatch on this exact file shape, independent of the two hypotheses above.

Any of the three would explain "no conflict message, no stash, no 'User changes restored' message" — the Builder's first step should be isolating which one is real before writing a fix, since the fix differs by hypothesis (extraction bug vs. scope-set bug vs. merge-detection bug).

## Acceptance Criteria

- [ ] Decide whether the silent overwrite of an uncommitted in-block
      `CLAUDE.md` edit during `install.sh --quick` reinstall is intentional
      (document why) or a regression (fix it, restoring the #3663 conflict-
      surfacing behavior for genuine in-block divergence).
- [ ] If a regression-test suite is reinstated to cover this, update its
      `install.sh` invocations to pass `--confirm-reinstall` on every
      reinstall call (required since #4188) so the suite actually exercises
      the reinstall path instead of erroring out before touching anything.

## Affected Files

- `install.sh` - the `reapply_loom_claude_md_block`/`_emit_loom_claude_block` helpers (~lines 555-605) and the in-block-edit reset guard + stash-restore flow (~lines 1246-1300+, guard comparison at ~1290-1293); this is where the fix (or a documented "working as intended") lands. Line numbers verified against `origin/main` at `fb20d279` (2026-08-04) — re-verify before use, they drift with every merged PR touching this file (one such PR, #5288, merged to `main` during this same curation pass).
- `scripts/install/stash-scope.sh` - the dirty ∩ ownership intersection that decides whether CLAUDE.md is stashed at all before the uninstall→reinstall cycle; rule out as a contributing cause per "Suspected Cause" above
- `scripts/test-install-quick-gitignore-stash.sh` (deleted in #5276; pre-deletion git history has the original #3588/#3611/#3663 scenario code) - reference only if a regression suite is reinstated per AC 2; a new suite would live at a fresh path since the old one is gone, not restore the deleted file verbatim (its "Stale test invocations" bug would just recur)

## Test Plan

- [ ] Manual verification: re-run the exact repro script in "Reproduction (finding 2)" above after the fix/decision lands; if fixed, confirm the installer now reports a named conflict, preserves a stash (`git stash list` non-empty), and does NOT silently discard `USER IN-BLOCK EDIT`. If declared intentional, confirm that decision is documented (comment in `install.sh` near the guard, and/or in this issue) so a future re-audit doesn't re-flag it as a bug.
- [ ] Automated tests: if a regression suite is reinstated (AC 2), it must invoke `install.sh` reinstall calls with `--confirm-reinstall` (not `-y --quick` alone) and actually assert against the fixed/intentional behavior — verify by running it and confirming it fails against the *pre-fix* code and passes against the *post-fix* code (a suite that passes unconditionally would repeat the original "vacuous pass" bug this issue exists to catch)
- [ ] Edge cases: confirm the fix (if any) does not regress the sibling `.gitignore` HEAD-reset+reapply path (#3588) or the out-of-block CLAUDE.md edit case (edits outside the marker block should still reset+reapply cleanly, per the existing `_stashed_block == _head_block` branch)

## Context

`scripts/test-install-quick-gitignore-stash.sh` (which contained the original,
now-stale regression coverage for #3588/#3611/#3663) was deleted in #5276 as
part of an orphaned-test-script hygiene pass -- it had zero CI/build-gate
invocations and, once actually run, was ~26% red for the reasons above. The
suite's git history (pre-deletion) has the original scenario code if a
reimplementation is wanted.

rjwalters:
## Summary

`install.sh --quick`'s reinstall could silently discard an uncommitted
in-block edit to root `CLAUDE.md` with no conflict message and no stash
to recover from — the #3663 conflict-surfacing guarantee was violated.
Root cause isolated to hypothesis 2 from the issue: `CLAUDE.md` was never
even entered into the reinstall's pre-uninstall stash in the first place.

## Root Cause

`scripts/install/stash-scope.sh::_emit_loom_ownership_paths` already
carves `.gitignore` explicitly into the Loom ownership set with a comment
explaining why: `loom-daemon init` rewrites it, but it's not enumerated
by the `defaults/` manifest walk (`_emit_installed_files_manifest`).

Root `CLAUDE.md` has the *identical* gap, previously un-carved-out. Its
Loom section is synthesized at install time from `LOOM_ROOT_POINTER`
(`loom-daemon/src/init/scaffolding.rs`) — a short pointer paragraph, not
copied verbatim from a literal `defaults/CLAUDE.md` file (only
`defaults/.loom/CLAUDE.md`, the full-guide copy, exists) — so the
manifest walk never lists root `CLAUDE.md` as Loom-owned at all.

Traced with instrumentation against the issue's exact repro:
- `_emit_loom_owned_dirty_paths` never included `CLAUDE.md` even though
  it was dirty, so `REINSTALL_OWNED_DIRTY` was empty and `install.sh`
  never stashed it (no "Stashing uncommitted Loom-owned changes" message
  at all — confirmed by direct instrumentation before the fix).
- With CLAUDE.md unprotected in the working tree, the chained
  `scripts/uninstall-loom.sh`'s STEP 6 "Smart Remove CLAUDE.md" ran its
  unconditional marker-based `sed '/<!-- BEGIN LOOM ORCHESTRATION -->/,/<!-- END LOOM ORCHESTRATION -->/d'`
  directly against the (uncommitted, unstashed) working tree — destroying
  the user's in-block edit right there, before `loom-daemon init --force`
  ever ran.
- With nothing stashed, there was nothing left to 3-way-merge-conflict
  against later, so the in-block-edit guard at install.sh's stash-restore
  path (the one named in the issue's "Suspected Cause") never even got a
  chance to run — it wasn't buggy, it was simply never reached.

This is a regression relative to #3663's intent (not an intentional
simplification) — the short-pointer-text rewrite didn't change the
guarantee that in-block edits should surface a conflict, it just
shrank what "in-block" contains.

## Changes

- `scripts/install/stash-scope.sh`: add `CLAUDE.md` to the explicit
  ownership-set carve-out alongside `.gitignore`, with a comment
  explaining the parallel gap and its worse (silent-data-loss) failure
  mode. This is the single shared helper both `install.sh --quick` and
  `scripts/install-loom.sh --clean` source, so both reinstall paths gain
  the protection (the `--clean` path had the same latent unprotected-copy
  risk, just without install.sh's more elaborate reset+reapply/conflict
  machinery on top).
- `defaults/scripts/tests/test-install-stash-scope.sh`: two new hermetic
  regression tests (already-CI-wired suite, `ci-wired.txt`) — (4) root
  `CLAUDE.md` is present in the ownership set alongside `.gitignore`, and
  (5) a dirty root `CLAUDE.md` is actually selected by
  `_emit_loom_owned_dirty_paths`. Verified both fail against the pre-fix
  helper and pass post-fix.

- [x] Decided the silent overwrite is a **regression** (not intentional)
      and fixed the root cause (missing ownership-set carve-out), restoring
      the #3663 conflict-surfacing guarantee for genuine in-block
      divergence.
- [x] Regression coverage: rather than reinstating the deleted
      `scripts/test-install-quick-gitignore-stash.sh` (a full install.sh
      E2E suite requiring a `cargo build`ed `loom-daemon` binary — outside
      the hermetic ~3min `shell-suite-tests` CI job's scope, and the exact
      kind of unwired/costly suite that led to its original deletion in
      #5276), added two hermetic unit tests to the already-CI-wired
      `test-install-stash-scope.sh` that pin the *exact* root cause
      (the ownership-set gap) at the layer install.sh's logic actually
      depends on. Confirmed they fail pre-fix / pass post-fix (see Test
      Plan). No suite in this repo invokes `install.sh` reinstall without
      `--confirm-reinstall` (verified against `test-install-reinstall-safety.sh`
      and `test-install-full-reinstall-no-target-mutation.sh`, both already
      correct) — finding 1 from the issue only ever affected the
      already-deleted suite, so there was no stale invocation to update.

## Test Plan

- [x] **Manual verification** — re-ran the issue's exact repro script
      end-to-end against the fix:
      ```
      === CLAUDE.md ===
      <!-- BEGIN LOOM ORCHESTRATION -->
      This repository uses [Loom](...) for AI-powered development orchestration ...
      <!-- END LOOM ORCHESTRATION -->
      === stash list ===
      stash@{0}: On main: loom-install: preserving user changes before --quick reinstall
      ```
      Installer output now reports:
      ```
      ⚠ Failed to restore stashed user changes automatically
        Conflicting file(s): CLAUDE.md
        git stash pop --index reported:
          error: Your local changes to the following files would be overwritten by merge:
              CLAUDE.md
      ...
        Your changes are preserved in the stash (stash@{0}).
      ```
      `git stash show -p stash@{0}` contains the user's `USER IN-BLOCK EDIT`
      text, confirming it is fully recoverable, not lost.
- [x] **Automated**: `bash defaults/scripts/tests/test-install-stash-scope.sh`
      → `Ran 10 test(s): 10 passed, 0 failed`. Confirmed the 2 new tests
      fail against a pre-fix checkout of `stash-scope.sh`
      (`Ran 10 test(s): 8 passed, 2 failed`) and pass post-fix.
- [x] **Edge cases** (both manually re-verified against the fix, no
      regression):
      - `.gitignore` HEAD-reset+reapply path (#3588): dirty `.gitignore`
        with a custom user pattern → reinstall reports "User changes
        restored", custom pattern preserved verbatim.
      - Out-of-block CLAUDE.md edit: a dirty line *outside* the marker
        block → reinstall still resets+reapplies cleanly ("User changes
        restored"), out-of-block edit preserved, fresh Loom block applied.
      - Also ran the full pre-existing suites unaffected in scope but
        sharing the touched helper: `test-install-reinstall-safety.sh`
        (21/21 pass) and `test-install-full-reinstall-no-target-mutation.sh`
        (8/8 pass).

Closes #5289

### DIFF
--- a/defaults/scripts/tests/test-install-stash-scope.sh
+++ b/defaults/scripts/tests/test-install-stash-scope.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 # test-install-stash-scope.sh — regression tests for the reinstall stash guard
-# scoping (issue #3597).
+# scoping (issue #3597; issue #5289 added tests 4-5).
 #
 # The `--quick` reinstall (install.sh) and `--clean` install (install-loom.sh)
 # guards used to run an unscoped `git stash push`, sweeping sibling installers'
 # uncommitted tracked changes into the stash and leaving a half-old/half-new
 # hybrid tree. The fix scopes the stash to the intersection of the dirty set
-# with Loom's ownership set (manifest paths + .gitignore) via
+# with Loom's ownership set (manifest paths + .gitignore + CLAUDE.md) via
 # scripts/install/stash-scope.sh::_emit_loom_owned_dirty_paths.
 #
 # Strategy: source the real helper against a temp git repo seeded with both a
@@ -15,7 +15,14 @@
 #   2. a pathspec-scoped `git stash push` leaves the sibling change untouched
 #      in the working tree and absent from the stash,
 #   3. a tree dirty with ONLY sibling changes yields no owned-dirty paths
-#      (callers skip the stash entirely).
+#      (callers skip the stash entirely),
+#   4. root CLAUDE.md carries the same explicit ownership-set carve-out
+#      `.gitignore` already has (issue #5289 — CLAUDE.md's Loom section is
+#      synthesized at install time, not copied from a literal defaults/
+#      file, so the manifest walk alone never lists it),
+#   5. a dirty root CLAUDE.md is actually selected for stashing by
+#      `_emit_loom_owned_dirty_paths` (the property the reinstall's
+#      in-block-edit conflict guard at install.sh:~1290 depends on).
 #
 # Usage:
 #   bash defaults/scripts/tests/test-install-stash-scope.sh
@@ -143,6 +150,50 @@ printf '{"version":"newer"}\n' > "$TMP_REPO/$SIBLING_PATH"
 
 SELECTED_SIBLING_ONLY="$(_emit_loom_owned_dirty_paths "$REPO_ROOT" "$TMP_REPO")"
 assert_eq "" "$SELECTED_SIBLING_ONLY" "sibling-only dirty tree produces no owned-dirty paths"
+
+echo "== Test 4: root CLAUDE.md is explicitly carved into the ownership set (issue #5289) =="
+# Root CLAUDE.md's Loom section is synthesized at install time from
+# LOOM_ROOT_POINTER (loom-daemon/src/init/scaffolding.rs) rather than copied
+# verbatim from a defaults/CLAUDE.md file, so `_emit_installed_files_manifest`'s
+# walk-of-defaults/ never enumerates it -- exactly the same gap `.gitignore`
+# already has an explicit carve-out for (it too is rewritten by
+# `loom-daemon init` but isn't part of the defaults/ walk). Without that
+# carve-out, `_emit_loom_owned_dirty_paths` never selects a dirty root
+# CLAUDE.md, so `install.sh --quick`'s reinstall never stashes an uncommitted
+# CLAUDE.md edit before the chained uninstall's marker-based `sed` strips the
+# Loom block -- silently destroying the edit with no conflict ever surfaced,
+# even when it landed *inside* the `<!-- BEGIN/END LOOM ORCHESTRATION -->`
+# markers (reproduction: issue #5289).
+if printf '%s\n' "$OWNERSHIP" | grep -qxF "CLAUDE.md"; then
+  pass "CLAUDE.md present in the Loom ownership set"
+else
+  fail "CLAUDE.md present in the Loom ownership set" "ownership set: $OWNERSHIP"
+fi
+if printf '%s\n' "$OWNERSHIP" | grep -qxF ".gitignore"; then
+  pass ".gitignore present in the Loom ownership set (sibling carve-out, #3588)"
+else
+  fail ".gitignore present in the Loom ownership set (sibling carve-out, #3588)" "ownership set: $OWNERSHIP"
+fi
+
+echo "== Test 5: a dirty root CLAUDE.md is selected for stashing (issue #5289) =="
+CLAUDE_TMP_REPO="$(mktemp -d "${TMPDIR:-/tmp}/loom-stash-scope-claude.XXXXXX")"
+git -C "$CLAUDE_TMP_REPO" init -q
+git -C "$CLAUDE_TMP_REPO" config user.email test@example.com
+git -C "$CLAUDE_TMP_REPO" config user.name "Test"
+printf '# Project\n\n<!-- BEGIN LOOM ORCHESTRATION -->\nold pointer\n<!-- END LOOM ORCHESTRATION -->\n' \
+  > "$CLAUDE_TMP_REPO/CLAUDE.md"
+git -C "$CLAUDE_TMP_REPO" add -A
+git -C "$CLAUDE_TMP_REPO" commit -qm "seed"
+printf '# Project\n\n<!-- BEGIN LOOM ORCHESTRATION -->\nUSER IN-BLOCK EDIT\n<!-- END LOOM ORCHESTRATION -->\n' \
+  > "$CLAUDE_TMP_REPO/CLAUDE.md"
+
+CLAUDE_SELECTED="$(_emit_loom_owned_dirty_paths "$REPO_ROOT" "$CLAUDE_TMP_REPO")"
+if printf '%s\n' "$CLAUDE_SELECTED" | grep -qxF "CLAUDE.md"; then
+  pass "dirty root CLAUDE.md is selected for stashing"
+else
+  fail "dirty root CLAUDE.md is selected for stashing" "selected: [$CLAUDE_SELECTED]"
+fi
+rm -rf "$CLAUDE_TMP_REPO"
 
 echo ""
 echo "Ran $TESTS_RUN test(s): $TESTS_PASSED passed, $TESTS_FAILED failed"
--- a/scripts/install/stash-scope.sh
+++ b/scripts/install/stash-scope.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # scripts/install/stash-scope.sh — scope the reinstall stash/reconcile guard
-# to Loom-owned paths (issue #3597).
+# to Loom-owned paths (issue #3597; issue #5289 added root CLAUDE.md).
 #
 # Both install.sh (`--quick` reinstall) and scripts/install-loom.sh (`--clean`)
 # guard uncommitted user changes across the uninstall→reinstall cycle by
@@ -13,8 +13,9 @@
 #
 # This helper narrows the guard to paths Loom actually owns: the intersection
 # of the dirty set (unstaged ∪ staged changes) with Loom's ownership set
-# (`_emit_loom_ownership_set` from manifest.sh, plus `.gitignore`, which
-# `loom-daemon init` rewrites but which is not part of the defaults/ walk).
+# (`_emit_loom_ownership_set` from manifest.sh, plus `.gitignore` and root
+# `CLAUDE.md`, both rewritten by `loom-daemon init` but not enumerated by the
+# defaults/ walk).
 #
 # Source with:
 #     source "$LOOM_ROOT/scripts/install/stash-scope.sh"
@@ -22,15 +23,16 @@
 # Public functions:
 #   _emit_loom_ownership_paths <loom_root> <target>
 #       One target-relative path per line: Loom's manifest ownership set plus
-#       `.gitignore`. Missing manifest.sh → just `.gitignore` (loud caller
-#       fallback expected).
+#       `.gitignore` and root `CLAUDE.md`. Missing manifest.sh → just those
+#       two (loud caller fallback expected).
 #
 #   _emit_loom_owned_dirty_paths <loom_root> <target>
 #       One target-relative path per line: the dirty set (unstaged ∪ staged
 #       changes) intersected with the ownership set. Empty output means no
 #       Loom-owned path is dirty → callers skip the stash entirely.
 
-# Emit the Loom ownership set (manifest paths + .gitignore), one per line.
+# Emit the Loom ownership set (manifest paths + .gitignore + root CLAUDE.md),
+# one per line.
 _emit_loom_ownership_paths() {
   local loom_root="$1"
   local target="$2"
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T15:58:56Z, age < 30m) — standing down without reclaiming. Not stomping.

rjwalters:
✅ **Approved!**

Reviewed the fix for the `install.sh --quick` reinstall silently discarding an uncommitted in-block edit to root `CLAUDE.md`.

**Correctness**: `stash-scope.sh::_emit_loom_ownership_paths` now carves `CLAUDE.md` into the explicit ownership set alongside `.gitignore`, mirroring the existing carve-out pattern and its documented rationale (both are rewritten by `loom-daemon init` but not enumerated by the `defaults/` manifest walk). Minimal, single-line functional change (`printf '%s\n.gitignore\nCLAUDE.md\n' ...`).

**Root cause analysis**: well-documented and traced with instrumentation — confirmed `CLAUDE.md` was never entered into the reinstall's pre-uninstall stash, so `uninstall-loom.sh`'s marker-based `sed` destroyed in-block edits before any conflict could be surfaced. Correctly identified as a regression relative to #3663's guarantee.

**Tests**: Two new hermetic regression tests (4 and 5) added to the already-CI-wired `test-install-stash-scope.sh` — pin the exact root cause at the layer install.sh depends on. Ran locally: `Ran 10 test(s): 10 passed, 0 failed`.

## Test Execution

- `bash defaults/scripts/tests/test-install-stash-scope.sh` — ✅ Executed locally: 10/10 pass
- Manual repro from PR description (installer output showing conflict properly surfaced) — ⚠️ Not re-executed locally; PR description documents it was run against the fix with expected conflict-surfacing output
- Edge cases (`.gitignore` reset+reapply path, out-of-block CLAUDE.md edit, `test-install-reinstall-safety.sh`, `test-install-full-reinstall-no-target-mutation.sh`) — ⚠️ Not re-executed locally; documented as re-verified in PR description

## CI Note

`Shell Test Suites (hermetic)` initially failed on the first CI run (`test-loom-daemon-update.sh`, an unrelated systemd/launchd daemon-update suite with no relation to this PR's diff). The workflow auto re-ran and the suite passed cleanly on rerun (`success`) — confirming this was a pre-existing flake, not a regression from this change. All checks now pass and merge state is `clean`.

Scope is tight (2 files, well-targeted), root cause is correctly diagnosed and fixed at the shared helper both reinstall paths depend on, and the new tests directly pin the fixed behavior.

rjwalters:
Standing down: re-checked PR state before writing my verdict and found `loom:pr` already applied (another Judge pass landed its approval at 2026-08-04T16:10:13Z, moments before I finished my own independent review). Not overwriting.

For the record, my independent review reached the identical conclusion — approve:

- **Root cause verified**: `_emit_loom_ownership_paths` (scripts/install/stash-scope.sh) never carved root `CLAUDE.md` into the Loom ownership set (only `.gitignore` had the explicit carve-out for the same 'rewritten by `loom-daemon init`, not walked by the `defaults/` manifest' reason). Confirmed `defaults/CLAUDE.md` does not exist on disk (only `defaults/.loom/CLAUDE.md`), so the manifest walk genuinely can never enumerate it — matches the PR's stated root cause exactly.
- **Fix verified sound**: single-line addition of `CLAUDE.md` to the `printf` in the ownership carve-out. Traced the downstream consumer (`install.sh`'s pre-existing #3663 reset+reapply/conflict-detection logic for `CLAUDE.md`, ~line 1290) and confirmed it was already correctly implemented but unreachable because `REINSTALL_OWNED_DIRTY` never contained `CLAUDE.md` before this fix — the fix closes exactly that gap, nothing else needed changing.
- **New tests verified both ways**: checked out the PR's worktree and ran `defaults/scripts/tests/test-install-stash-scope.sh` — 10/10 pass post-fix. Then temporarily swapped in the pre-fix `stash-scope.sh` (from parent commit 850f9ad1) against the same new test file: tests 4 and 5 fail as expected (`CLAUDE.md` absent from ownership set; dirty `CLAUDE.md` not selected for stashing), confirming the tests actually pin the regression rather than being vacuously true.
- **No regression on the `.gitignore` sibling path (#3588)**: unchanged, still explicitly carved out; tests 1-3 (existing) and the new test 4's `.gitignore` assertion all still pass.
- **CI**: confirmed the `Shell Test Suites (hermetic)` job's first-run failure (`test-loom-daemon-update.sh`, unrelated to this diff — no reference to `CLAUDE.md`/`stash-scope.sh`) is a pre-existing flake, not a regression: the latest main CI run passed that same suite, and a rerun of this PR's job passed cleanly (5m45s, all green). Merge state is `clean`.

No additional concerns. Nothing further needed from me here — the landed `loom:pr` verdict stands.

rjwalters:
**Champion Auto-Merge**

- Judge approved (`loom:pr` label) — two independent Judge passes, both approving after tracing the fix's only downstream consumer and toggling pre-fix/post-fix behavior of the new tests
- Merge-risk judgment passed: single-line functional change (`CLAUDE.md` added to a protective stash-scope carve-out) confined to `scripts/install/stash-scope.sh`; traced its only caller (`install.sh`'s pathspec-scoped `git stash push`) and confirmed the change only widens what gets safely stashed before reinstall — it is not used for any destructive reset/overwrite path, so a bug here can at worst leave the pre-existing gap unfixed, not introduce a new regression
- Diff size: 91 lines (+81/-10) — informational, not a gate
- No critical files modified
- No merge conflicts
- All CI checks passing (one job's first-run failure was an unrelated pre-existing flake, confirmed by both Judges and by a clean rerun)

`git revert <commit-sha>`

---
*Automated by Champion role*

PR #5300, opened 2026-08-04
fix(guard): mask grep/rg positional pattern args in catastrophic curl-pipe scan

### LINKED ISSUE #5158: Guard false positive: catastrophic curl-pipe-shell pattern fires on grep introspection of its own regex

rjwalters:
## Summary

`.loom/hooks/guard-destructive-generic.sh`'s `catastrophic:curl .* | .*sh` pattern (in `ALWAYS_BLOCK_PATTERNS`) scans a working copy (`COMMAND_NO_LITERAL_TEXT`) that was previously masked only for named-flag values (`--body`/`--message`/`--title`/`--notes`/`--comment`/`-m`), never for a bare positional pattern argument. A read-only `grep`/`rg` invocation whose *quoted search pattern* happens to textually resemble a curl-pipe-to-shell invocation could be misread as a live one, since `grep`/`rg` never execute what they search for.

This mirrors the fix already landed for `guard-loom-workflow.sh` (`mask_command_positional_args()`, #5109/#5155/#5160) and the ask-tier analog in this same file (`mask_ask_positional_args()`, #5235). Unlike `mask_ask_positional_args()` — which deliberately **excludes** grep/rg because its working copy (`COMMAND_ASK_SCAN`) also feeds the SQL-DDL check, which intentionally scans a grep pattern argument for a literal DDL phrase — `COMMAND_NO_LITERAL_TEXT` is scanned completely independently of `COMMAND_ASK_SCAN`, so it is safe to include grep/egrep/fgrep/rg in a new, separate masking function scoped to this working copy only.

## Changes

- Add `mask_catastrophic_positional_args()` to `defaults/hooks/guard-destructive-generic.sh` — a near-duplicate of `mask_ask_positional_args()`, but with `grep|egrep|fgrep|rg` in its command-name allowlist instead of `check-duplicate.sh`.
- Wire it into `COMMAND_NO_LITERAL_TEXT` construction, gated on a cheap substring probe (`*"grep"*`/`*"rg "*`) to keep it off the hot path for the vast majority of commands.
- Mirror the same edit into the installed `.loom/hooks/guard-destructive-generic.sh` copy (verified byte-identical to `defaults/` both before and after).
- Add regression tests to `tests/hooks/test-guard-destructive.sh` covering the false-positive fix and the regression floor (real curl-pipe-to-shell invocations, including one chained after a masked grep, still deny).

## Note on the issue's literal repro command (re-verified against current main)

I re-verified the exact commands quoted in the issue's "Occurrence" section (`grep -n "curl .*|" defaults/hooks/guard-destructive.sh | head`, and the second `grep -nE "..."` command with no `curl` substring at all) directly against the unpatched pattern on `origin/main` (`8cb78699`) and neither one currently matches `ALWAYS_BLOCK_PATTERNS`' curl-pipe regex — the anchor `(^|[;&|[:space:](])(curl|wget)...` requires the character immediately before `curl` to be a separator (whitespace/`;`/`&`/`|`/`(`)/start-of-string, and in that exact repro the character immediately before `curl` is the opening double-quote, which is not in that set. So the literal Occurrence text no longer reproduces as filed (likely already narrowed by the repo#29 anchor-tightening fix referenced in the surrounding code comments, sometime after the 2026-07-27 telemetry entry).

The underlying false-positive class is still real and reproducible, though — any quoted grep/rg pattern where `curl` is preceded by *whitespace within the quotes* (a very ordinary way to phrase a search, e.g. `grep -n "check curl .*| sh usage" file"`) still matches and denies on unpatched main; I confirmed this directly against both the unpatched and patched guard via the hook's stdin JSON contract, and it is the shape used in the new regression tests. I did not chase re-deriving the exact original repro further since the fix (masking any leading grep/egrep/fgrep/rg's own quoted positional pattern) closes the false-positive class regardless of the specific phrasing.

| New `mask_catastrophic_positional_args()` masks quoted positional grep/egrep/fgrep/rg pattern arguments in `COMMAND_NO_LITERAL_TEXT` before `ALWAYS_BLOCK_PATTERNS` scans it | ✅ | Added function + wiring in `defaults/hooks/guard-destructive-generic.sh`; verified via direct hook invocation and new tests |
| `grep`/`egrep`/`fgrep`/`rg` introspection quoting curl-pipe-shell-shaped text no longer denies | ✅ | New `assert_allow` tests (4 command variants + multi-file) all pass |
| Regression floor: real curl-pipe-to-shell invocations still deny (direct, sudo-wrapped, multi-stage via gunzip, and one chained after a grep) | ✅ | New `assert_deny` tests pass; existing curl-pipe deny tests (lines ~350-390) still pass unmodified |
| Masking scoped to grep/egrep/fgrep/rg only — `bash -c "curl ... | sh"` / `eval "curl ... | sh"` stay unmasked | ✅ | Not in the allowlist regex (`cmdre`); confirmed via manual hook invocation (these are a pre-existing, documented accepted miss of the raw pattern itself, unrelated to and unaffected by this fix — see repo#29 comment in the guard) |
| Does not touch `COMMAND_ASK_SCAN` / `SQL_DDL_PATTERN` behavior | ✅ | Separate working copy, separate function; full existing SQL-DDL grep-introspection test group passes unmodified |
| `defaults/hooks/guard-destructive-generic.sh` and `.loom/hooks/guard-destructive-generic.sh` stay byte-identical | ✅ | `diff` confirms identical before commit |
| New tests added to `tests/hooks/test-guard-destructive.sh` | ✅ | 7 new cases (5 `assert_allow`, 2 `assert_deny`) |
| `tests/hooks/test-guard-destructive-dispatcher.sh` still passes unmodified | ✅ | 12/12 passed |

## Test Plan

- `./tests/hooks/test-guard-destructive.sh`: 786/786 passed (was 779 before the 7 new cases added here), 0 failed.
- `./tests/hooks/test-guard-destructive-dispatcher.sh`: 12/12 passed, unmodified.
- Manual verification: fed the hook's stdin JSON contract directly to both the unpatched (`origin/main`) and patched guard for the false-positive case, the direct curl-pipe-to-sh case, and the grep-then-real-curl chained case — confirmed the false positive flips from `deny` to allow (empty/no `permissionDecision`) only after the patch, while both denies are unaffected.
- `shellcheck -x` on both changed shell files: no new warnings (only pre-existing info-level `SC2016` in unrelated pre-existing lines).

Closes #5158

### DIFF
--- a/.loom/hooks/guard-destructive-generic.sh
+++ b/.loom/hooks/guard-destructive-generic.sh
@@ -2351,6 +2351,83 @@ mask_ask_positional_args() {
     }'
 }
 
+# Mask quoted POSITIONAL arguments (no preceding flag name) to a small
+# allowlist of known non-executing SEARCH commands (issue #5158). Extends the
+# #5235 mask_ask_positional_args() fix shape to the CATASTROPHIC-tier working
+# copy (COMMAND_NO_LITERAL_TEXT below), which mask_ask_positional_args()
+# deliberately excludes grep/egrep/fgrep/rg from — see that function's header
+# comment for why: COMMAND_ASK_SCAN also feeds SQL_DDL_PATTERN, which
+# intentionally scans a `grep '<pattern>' file` invocation's own quoted
+# positional pattern for a literal destructive-DDL phrase and denies, by
+# design. COMMAND_NO_LITERAL_TEXT is built and scanned entirely independently
+# of COMMAND_ASK_SCAN — SQL_DDL_PATTERN never reads COMMAND_NO_LITERAL_TEXT —
+# so masking grep/rg's own quoted pattern argument on THIS working copy does
+# not reintroduce that regression. Without this, `grep -n "curl .*|" <file>` —
+# read-only introspection of a guard's own source text — gets misread as a
+# live curl-pipe-to-shell invocation by ALWAYS_BLOCK_PATTERNS, because grep
+# never executes what it searches for.
+#
+# This is an intentional NEAR-DUPLICATE of mask_ask_positional_args() just
+# above (itself a near-duplicate of guard-loom-workflow.sh's
+# mask_command_positional_args(), #5155/#5160) — kept as a SEPARATE function
+# so a future tuning of one masking pass can never silently change another's
+# behavior, per the "never couple the two guards'/tiers' masking" convention
+# documented in mask_ask_positional_args()'s header comment above.
+mask_catastrophic_positional_args() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)
+        DQ = sprintf("%c", 34)
+        # Command-name allowlist: known non-executing search commands whose
+        # positional pattern arguments are inert search text, never live
+        # shell syntax. Unlike mask_ask_positional_args() above, grep/egrep/
+        # fgrep/rg ARE included here — see the function header comment for
+        # why that is safe on this (catastrophic-tier) working copy.
+        cmdre = "(grep|egrep|fgrep|rg)"
+        flagre = "([ \t]+-[A-Za-z0-9_-]+)*"
+        anchor = "(^|[ \t\n;&|`(])" cmdre flagre "[ \t]+"
+        buf = ""
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        out = ""
+        while (match(s, anchor)) {
+            pre     = substr(s, 1, RSTART - 1)
+            matched = substr(s, RSTART, RLENGTH)
+            rest    = substr(s, RSTART + RLENGTH)
+            out = out pre matched
+            # Mask every consecutive quoted positional argument immediately
+            # following the anchor (whitespace-separated). Stops at the first
+            # non-quote-starting token, so anything after the argument list
+            # (a pipe, &&, an unrelated command chained on the same line) is
+            # left fully visible.
+            while (1) {
+                qc = substr(rest, 1, 1)
+                if (qc != DQ && qc != SQ) break
+                endpos = 0
+                for (i = 2; i <= length(rest); i++) {
+                    if (substr(rest, i, 1) == qc) { endpos = i; break }
+                }
+                if (endpos == 0) break
+                inner = substr(rest, 2, endpos - 2)
+                if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                    gsub(/./, "X", inner)
+                }
+                out = out qc inner qc
+                rest = substr(rest, endpos + 1)
+                while (substr(rest, 1, 1) == " " || substr(rest, 1, 1) == "\t") {
+                    out = out substr(rest, 1, 1)
+                    rest = substr(rest, 2)
+                }
+            }
+            s = rest
+        }
+        out = out s
+        printf "%s", out
+    }'
+}
+
 # Helper: output a deny decision and exit
 #
 # Optional second arg is a short, STABLE rule tag (issue #3771) recorded as the
@@ -2513,10 +2590,20 @@ ALWAYS_BLOCK_PATTERNS=(
 # payloads still reach the raw scan; spans carrying `$(` / backtick are left
 # intact so command-substitution smuggling still hard-denies.
 COMMAND_NO_LITERAL_TEXT="$COMMAND"
+# #5158: mask a leading grep/egrep/fgrep/rg invocation's own quoted pattern
+# argument BEFORE the flag-keyed strip below, so introspecting the guard's
+# own source (e.g. `grep -n "curl .*|" defaults/hooks/guard-destructive.sh`)
+# isn't misread as a live curl-pipe-to-shell invocation. Cheap substring gate
+# keeps the awk call off the hot path for the vast majority of commands that
+# never invoke grep/rg, mirroring the check-duplicate.sh substring gate used
+# for mask_ask_positional_args() below.
+if [[ "$COMMAND" == *"grep"* || "$COMMAND" == *"rg "* ]]; then
+    COMMAND_NO_LITERAL_TEXT=$(mask_catastrophic_positional_args "$COMMAND_NO_LITERAL_TEXT")
+fi
 if [[ "$COMMAND" == *"--body"* || "$COMMAND" == *"--message"* || \
       "$COMMAND" == *"--title"* || "$COMMAND" == *"--notes"* || \
       "$COMMAND" == *"--comment"* || "$COMMAND" == *"-m"* ]]; then
-    COMMAND_NO_LITERAL_TEXT=$(strip_literal_text "$COMMAND")
+    COMMAND_NO_LITERAL_TEXT=$(strip_literal_text "$COMMAND_NO_LITERAL_TEXT")
 fi
 
 for pattern in "${ALWAYS_BLOCK_PATTERNS[@]}"; do
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -2351,6 +2351,83 @@ mask_ask_positional_args() {
     }'
 }
 
+# Mask quoted POSITIONAL arguments (no preceding flag name) to a small
+# allowlist of known non-executing SEARCH commands (issue #5158). Extends the
+# #5235 mask_ask_positional_args() fix shape to the CATASTROPHIC-tier working
+# copy (COMMAND_NO_LITERAL_TEXT below), which mask_ask_positional_args()
+# deliberately excludes grep/egrep/fgrep/rg from — see that function's header
+# comment for why: COMMAND_ASK_SCAN also feeds SQL_DDL_PATTERN, which
+# intentionally scans a `grep '<pattern>' file` invocation's own quoted
+# positional pattern for a literal destructive-DDL phrase and denies, by
+# design. COMMAND_NO_LITERAL_TEXT is built and scanned entirely independently
+# of COMMAND_ASK_SCAN — SQL_DDL_PATTERN never reads COMMAND_NO_LITERAL_TEXT —
+# so masking grep/rg's own quoted pattern argument on THIS working copy does
+# not reintroduce that regression. Without this, `grep -n "curl .*|" <file>` —
+# read-only introspection of a guard's own source text — gets misread as a
+# live curl-pipe-to-shell invocation by ALWAYS_BLOCK_PATTERNS, because grep
+# never executes what it searches for.
+#
+# This is an intentional NEAR-DUPLICATE of mask_ask_positional_args() just
+# above (itself a near-duplicate of guard-loom-workflow.sh's
+# mask_command_positional_args(), #5155/#5160) — kept as a SEPARATE function
+# so a future tuning of one masking pass can never silently change another's
+# behavior, per the "never couple the two guards'/tiers' masking" convention
+# documented in mask_ask_positional_args()'s header comment above.
+mask_catastrophic_positional_args() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)
+        DQ = sprintf("%c", 34)
+        # Command-name allowlist: known non-executing search commands whose
+        # positional pattern arguments are inert search text, never live
+        # shell syntax. Unlike mask_ask_positional_args() above, grep/egrep/
+        # fgrep/rg ARE included here — see the function header comment for
+        # why that is safe on this (catastrophic-tier) working copy.
+        cmdre = "(grep|egrep|fgrep|rg)"
+        flagre = "([ \t]+-[A-Za-z0-9_-]+)*"
+        anchor = "(^|[ \t\n;&|`(])" cmdre flagre "[ \t]+"
+        buf = ""
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        out = ""
+        while (match(s, anchor)) {
+            pre     = substr(s, 1, RSTART - 1)
+            matched = substr(s, RSTART, RLENGTH)
+            rest    = substr(s, RSTART + RLENGTH)
+            out = out pre matched
+            # Mask every consecutive quoted positional argument immediately
+            # following the anchor (whitespace-separated). Stops at the first
+            # non-quote-starting token, so anything after the argument list
+            # (a pipe, &&, an unrelated command chained on the same line) is
+            # left fully visible.
+            while (1) {
+                qc = substr(rest, 1, 1)
+                if (qc != DQ && qc != SQ) break
+                endpos = 0
+                for (i = 2; i <= length(rest); i++) {
+                    if (substr(rest, i, 1) == qc) { endpos = i; break }
+                }
+                if (endpos == 0) break
+                inner = substr(rest, 2, endpos - 2)
+                if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                    gsub(/./, "X", inner)
+                }
+                out = out qc inner qc
+                rest = substr(rest, endpos + 1)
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: code changes look reasonable on first read (targeted masking fix mirroring #5109/#5155/#5160/#5235 pattern, with regression tests). CI is still running (Installer Integration Tests, Shell Test Suites (hermetic) pending). Releasing the claim and skipping — a later tick will re-evaluate once CI settles.

rjwalters:
✅ **Approved!**

Reviewed `mask_catastrophic_positional_args()` (defaults/hooks/guard-destructive-generic.sh) and the mirrored `.loom/hooks/` copy — confirmed byte-identical. This is a well-scoped near-duplicate of the existing `mask_ask_positional_args()` (#5235) pattern, correctly limited to the `COMMAND_NO_LITERAL_TEXT` (catastrophic-tier) working copy so it can safely include grep/egrep/fgrep/rg — unlike the ask-tier masker, this working copy never feeds `SQL_DDL_PATTERN`, so there's no risk of reintroducing that intentional grep-introspection DDL catch.

Traced the reordering where `strip_literal_text` now runs on the already-masked `COMMAND_NO_LITERAL_TEXT` instead of raw `$COMMAND` — both masking functions are byte-length-preserving and operate on disjoint syntactic targets (flag-keyed heredoc bodies vs. grep/rg quoted positional args), so this reorder doesn't change behavior for either pass.

## Test Execution

- `tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 passed
- `tests/hooks/test-guard-destructive.sh` — 786/786 passed (matches PR's claimed count, including the 7 new #5158 regression tests: 5 allow + 2 deny narrows-not-widens cases)
- `shellcheck -S error -x defaults/hooks/guard-destructive-generic.sh` — exit 0, no new findings
- CI: all checks pass, merge state CLEAN

Nice, honest write-up in the PR body about the literal issue repro no longer reproducing on current main, while still identifying and fixing the underlying false-positive class with a working alternate repro — that's the right call rather than force-fitting the stale example.

rjwalters:

merging it automatically:

- **Blast radius**: modifies `.loom/hooks/guard-destructive-generic.sh` and its `defaults/hooks/` mirror — explicitly named as a high-blast-radius surface in Champion's own criteria (guard hooks mediate every agent's Bash tool use fleet-wide). `mask_catastrophic_positional_args()` is wired directly into the `COMMAND_NO_LITERAL_TEXT` construction that feeds `ALWAYS_BLOCK_PATTERNS` — i.e. it changes what the catastrophic-tier curl-pipe-to-shell scan sees. A bug in the anchor regex or masking logic here fails *silently*: it would under-block (let a real curl-pipe-to-shell command slip through undetected) rather than error visibly. The PR includes solid regression tests (7 new cases, including 2 explicit narrows-not-widens deny checks) and Judge traced both file copies and ran the full suite (786/786) plus shellcheck — but test coverage doesn't change the axis: this is exactly the class of file this criterion calls out for a human's eyes regardless of how well-reviewed the diff is.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5300`

---
*Automated by Champion role*

Issue #4146, opened 2026-07-28
Collect and report the cross-host duplicate-dispatch baseline (missing measurement half of #4085)


rjwalters:
## Problem

#4085 ("measure the baseline duplicate-dispatch rate") closed as **COMPLETED** when PR #4116 merged, but it delivered only the *detection* half. The **measurement half — actually producing a baseline number — was never done, and nothing now tracks it**, because the parent issue is closed.

The instrumentation is real but inert:

- `detect_collisions` defaults to `false` — `loom-daemon/src/sweep_registry.rs:1314` and `:1340`
- It is gated behind `LOOM_DETECT_COLLISIONS` / `autonomous.collisionDetection.enabled` — `loom-daemon/src/sweep_registry.rs:4411`–`4422`
- `.loom/config.json` in this repo sets **neither key**, so detection is off
- The counter surfaces in the work-finder tick summary as `{} cross-host-collision(s)` at `loom-daemon/src/work_finder.rs:1073`–`1083` and `:1261`–`1272` (both paths)
- The **running daemon binary is dated Jul 27 09:43**, before #4116 merged at 2026-07-27T23:20 — the counter is not in the running process, so flipping the config key alone changes nothing

## Why this matters

The baseline is a load-bearing input to two open decisions, both of which are currently reasoning about an unquantified number:

- **#4028** (soft-claim room, currently `loom:building`) — without a baseline there is no way to judge whether its coordination machinery is worth its complexity.
- **#3979** (elastic compute, `loom:blocked`) — Option A federates N pollers over one forge queue. Multiplying hosts multiplies the collision rate; the Champion gate on that proposal is explicitly waiting on this number.

## Acceptance Criteria

- [ ] `autonomous.collisionDetection.enabled` is set to `true` in `.loom/config.json` (committed), or the equivalent env override is documented for the observation window
- [ ] The daemon is rebuilt from a commit that includes PR #4116 and rolled, so the counter is live in the running process
- [ ] Detection runs across a stated observation window (suggest ≥24h of normal multi-sweep activity) covering at least one period of concurrent sweeps
- [ ] The observed `cross-host-collision(s)` total and the tick count it accumulated over are **reported as a comment on #4028 and #3979**, with the window's start/end and the number of hosts active
- [ ] A one-line interpretation is recorded: is the rate material enough to justify #4028's soft-claim layer?

## Notes

[...]

Issue #87, opened 2025-10-14
Curator should proactively mark well-formed issues as loom:ready


rjwalters:
## Problem

The Curator role is currently passive - it waits to find issues without workflow labels, but doesn't proactively check if existing unlabeled issues are already well-formed and ready for implementation.

**Example**: Issue #84 "Expand frontend unit test coverage" was a well-formed issue created by the user, but it never went through the curation workflow:
- Never marked as `loom:proposal` (Architect didn't create it)
- Never marked as `loom:ready` (Curator didn't review it)
- Went straight to `loom:in-progress` (Worker claimed it directly)

This bypassed the intended workflow and quality gates.

## Root Cause

The Curator's interval prompt and role docs focus on:
1. Finding issues WITHOUT workflow labels
2. Enhancing vague or incomplete issues

But they don't emphasize:
3. **Checking if unlabeled issues are ALREADY complete and ready**
4. **Proactively marking well-formed issues as `loom:ready`**

### Update Curator Interval Prompt (`defaults/roles/curator.json`)

**Current interval prompt:**
```
Check for open issues without loom:proposal, loom:ready, or loom:in-progress labels and enhance them
```

**Proposed interval prompt:**
```
Find unlabeled open issues. If already well-formed (has clear problem, acceptance criteria, test plan), mark as loom:ready immediately. Otherwise, enhance and then mark as loom:ready.
```

### Update Curator Role Documentation (`defaults/roles/curator.md`)

Add a new section after "Finding Work":

```markdown
## Triage: Ready or Needs Enhancement?

When you find an unlabeled issue, first assess if it's already implementation-ready:

### Quick Checklist
- ✅ Clear problem statement
- ✅ Acceptance criteria or success metrics  
- ✅ Test plan or testing guidance
- ✅ No obvious dependencies or blockers

### If ALL checkboxes pass:
**Mark it `loom:ready` immediately** - no enhancement needed:
\`\`\`bash
gh issue edit <number> --add-label "loom:ready"
\`\`\`

### If ANY checkboxes fail:
Enhance the issue by adding missing pieces, then mark `loom:ready`.

### Examples

**Already Ready** (mark immediately):
- Issue has detailed problem statement ✅
- Lists clear acceptance criteria ✅  
- Includes test plan checklist ✅
- No dependencies mentioned ✅
→ Action: `gh issue edit <number> --add-label "loom:ready"`

**Needs Enhancement** (improve first):
- Vague title "fix the bug" ❌
[...]

PR #5301, opened 2026-08-04
fix(daemon): verify claim ownership against the forge timeline before cancel/reap label-restore

### LINKED ISSUE #5017: cross-host duplicate dispatch observed live with collisionDetection enabled on all hosts (#12 and #87), plus a sweep wedged 6.5h holding its claim
## Observation

Two hosts were running sweeps on the same issues at the same time, with
`autonomous.collisionDetection.enabled=true` **on all three hosts** (verified in
each host's `.loom/config.json`).

Captured 2026-08-03 ~14:30Z:

| issue | robb-studio | loom-worker-1 |
|---|---|---|
| **#12** (sky130-bandgap) | `sweep-issue-12-1785763141`, phase `curator-done` | `sweep-issue-12-1785765868`, phase `-` |
| **#87** (gf180-trng) | `sweep-issue-87-1785743891`, phase `-` | `sweep-issue-87-1785767159`, phase `curator-done` |

Both pairs were live simultaneously. The sweep-id suffixes are epoch seconds, so
the second dispatch landed **2,727s (~45 min) after the first** for #12, and
**23,268s (~6.5 h) after** for #87.

## The two cases are not the same, and the distinction matters

- **#87 looks like (accidental) recovery.** studio's sweep had been alive
  **6h30m at 0.0% CPU**, never advanced past phase `-`, and had **never created an
  `issue-87` worktree**. It was wedged, and worker-1 picking the issue up is the
  outcome you would want — except nothing detected the wedge, and studio still held
  the claim. (Cancelled manually via `loom-daemon cancel --issue 87`; SIGTERM was
  ignored and it escalated to SIGKILL, confirming it was wedged rather than quiet.)
- **#12 is a genuine collision.** studio's sweep was *actively progressing*
  (`curator-done`) when worker-1 started a second one 45 minutes later. Two agents,
  two tokens, one issue — duplicated cost and a merge race.

So there are plausibly two defects here: collision detection not preventing the
live-vs-live case, and no liveness criterion that would let a wedged sweep be
reclaimed deliberately rather than by accident.

## Why the existing issues do not cover this

- **#4146** (`loom:curated`, `loom:operator-only`) is about *collecting the
  cross-host duplicate-dispatch baseline* — measurement, still unrun.
- **#4444** is the dispatch label-guard gap.
- Collision detection (#4085) was enabled here on 2026-07-28 and its observation
  window for #4028/#3979 was declared open.

This issue is the concrete counter-example: detection enabled, duplicates
happening anyway, with timestamps. Worth treating as evidence *for* #4146 rather
than a replacement for it.

## Questions a fix needs to answer

- Does collision detection consider a peer claim *before* dispatch, or only
  reconcile after? A 45-minute gap should be plenty for either.
- Is the peer-claim signal per-repo? Both duplicated issues are in **canary repos**
  (sky130-bandgap, gf180-trng), which are the lowest dispatch priority and the most
  likely to be picked up simultaneously by idle hosts.
- What is the intended liveness criterion for "this sweep is wedged, reclaim it"?
  Today `0% CPU` alone is explicitly normal, so the signal has to be phase
  progression and/or worktree creation — #87 had neither for 6.5 hours.

## Acceptance criteria

- [ ] Two hosts cannot both hold a live sweep on the same issue when
      `collisionDetection.enabled=true`; if it happens, one backs off.
- [ ] A wedged sweep (no phase progression and no worktree past a threshold) is
      detected and either reclaimed or reported — not left holding a claim for hours.
- [ ] Whichever path lands, a duplicate that does occur is *logged loudly* on both
      hosts rather than being visible only by diffing two `status` outputs by hand.

## Provenance

Found during a manual cross-host check on 2026-08-03, immediately after fixing an
unrelated dispatch outage on the third host. Neither duplicate was visible from any
single host's own status output.

### LINKED ISSUE #5282: loom-daemon cancel releases a loom:building claim owned by a DIFFERENT host's surviving sweep — destroys the cross-host mutex
## Context

Observed live on loom#5270 (2026-08-04 ~14:30Z). Cross-host double-dispatch race:

- robb-studio sweep (operator-dispatched) started 14:25:10Z and applied the `loom:building` claim at 14:27:40Z.
- loom-worker-1's work finder dispatched the same issue at 14:27:47Z — its tick had read `loom:issue` just before the label flip (normal propagation-window race).
- The duplicate was stood down with the sanctioned path: `loom-daemon cancel --issue 5270` on worker-1 (clean SIGTERM, no escalation).

**Bug:** after the cancel, the issue's labels reverted to `loom:issue` — `loom:building` was removed — even though the surviving robb-studio sweep owns that claim and was (and is) still running. The cancelled sweep's teardown (or its `--claim-owned` release path) released a claim that belonged to the *other* host's sweep. That destroyed the only cross-host mutex and re-opened the issue to a third dispatch; the operator had to re-apply `loom:building` by hand.

This is the label-mutex sibling of the judge-claim rule already in the role prompts ("back off WITHOUT removing the label" when the timeline shows an earlier claimant): the machinery's own cancel path violates it.

- [ ] A cancelled/dying sweep only releases the `loom:building` claim if it can verify it is the claim's owner (e.g. timeline check: the claim's `labeled` event belongs to this sweep's window/actor — same verification the judge reclaim path uses).
- [ ] When ownership is ambiguous, leave the label; claim-TTL reclaim (#5123 lineage) already handles a genuinely-orphaned claim.
- [ ] Regression test: two racing claimants; cancelling the loser leaves the winner's claim intact.

---

### Root Cause (verified against `5e3f5218`, 2026-08-04)

The `loom:building` restore-on-cancel/reap path is entirely **host-local** in its
ownership check, which is why a different host's live claim gets destroyed:

- `SweepRegistry::finish_cancel` (`loom-daemon/src/sweep_registry/reaper.rs:738`,
  the cancel path) and the natural-exit reaper (`reaper.rs:928`, same pattern)
  both gate the label restore on `self.release_lock_owned(issue, sweep_id).retained()`
  (`loom-daemon/src/sweep_registry/locks.rs:346`) — i.e. "does *my local*
  `.loom/locks/issue-<N>/owner.json` say a different sweep_id owns this now?"
- That lock directory, `live_claim::probe`'s live-lock leg, and the
  `~/.loom/sweeps.json` machine-level journal it also consults are all
  documented as host/daemon-instance-scoped, not cross-host (see the module
  doc at `loom-daemon/src/live_claim.rs:1-30`, which explicitly cites the
  #4275 incident where "a *second* `loom-daemon` instance on the same host...
  shared neither the first daemon's memory nor its `.loom/locks/`" — the
  existing fix only closes the same-host, second-instance gap).
- When the surviving sweep is on a **different host** (this issue's scenario:
  robb-studio vs. loom-worker-1), `release_lock_owned` on the cancelled host
  finds no local lock entry belonging to the other host's sweep, returns
  `Released` (not `Superseded`), and `finish_cancel` proceeds to call
  `restore_label_to_ready(issue)` (`loom-daemon/src/sweep_registry/guards.rs:561`),
  which unconditionally removes `loom:building` and re-adds `loom:issue`
  (absent a `loom:blocked`/`loom:operator-only`/PR carve-out — none of which
  apply here) — destroying the only cross-host mutex: the label itself.
- The reaper's clean-exit path (`reaper.rs:928`) performs the identical
  `release_lock_owned` check before its own `restore_label_to_ready` call
  (`reaper.rs` ~968), so it is very likely exposed to the same cross-host gap.
  This is a code-symmetry inference, not independently reproduced — flagging
  as **Suspected Cause (unverified)** for that second call site; the Builder
  should confirm during implementation rather than assume.

The reap-on-clean-exit path (`reaper.rs:928` → `restore_label_to_ready` around
`reaper.rs:968`) likely shares this bug by the same code-path symmetry as the
cancel path, since both gate solely on `release_lock_owned`. Needs
verification: either a targeted reproduction (same cross-host race, natural
exit instead of cancel) or a code-level argument for why the reap path is
safe where cancel is not.

### Why the existing forge-side signal isn't used here

`claim_reconciliation.rs` already has the right *pattern* for a forge-side
(not host-local) ownership signal — `fetch_claim_labeled_at`
(`loom-daemon/src/claim_reconciliation.rs:1450`) fetches the most recent
`labeled <label>` timeline event via `gh api
repos/{owner}/{repo}/issues/{n}/timeline`, which works identically for issues
and PRs (both are the same underlying `/issues/{n}` REST resource — the
function is already parameterized as `pr_number: u32` but nothing about the
endpoint is PR-specific). That module's own `decide`/`apply_live_claim_veto`
functions, however, are wired to the **periodic reconciliation pass**, not to
`finish_cancel`/`reap_once` — those two call sites only ever consult the
host-local lock. No forge-side "did someone else (re-)claim this after I
started?" check exists on the cancel/reap label-restore path today.

## Acceptance Criteria

- [ ] `finish_cancel` (`reaper.rs:738`) does not restore `loom:building ->
      loom:issue` when the issue's current `loom:building` labeled-event
      timestamp (fetched via `gh api .../timeline`, same shape as
      `fetch_claim_labeled_at`) is **newer** than this sweep's own
      dispatch/claim time (`started_at`, or the `--claim-owned` dispatch
      timestamp if more precise) — that gap means a different claimant
      (possibly on another host) has (re-)claimed the issue since this sweep
      began.
- [ ] The same check is added to (or the fix is proven unnecessary for) the
      natural-exit reap path's label restore at `reaper.rs:928`/`~968` — see
      "Suspected Cause" above; do not skip this without stating which.
- [ ] When the forge-side check is inconclusive (API failure/timeout,
      unparseable response), fail open toward the **existing** host-local
      behavior (do not introduce a new hang or a permanently-parked claim) —
      matches the fail-open posture every other check in this module already
      uses (see `restore_label_to_ready`'s own doc comment: "All checks are
      best-effort and fail-open").
- [ ] Regression test: two racing claimants simulated via distinct `sweep_id`s
      with no shared local lock state (i.e. the *absence* of a local
      `Superseded`/`HolderAlive` signal must not be treated as "safe to
      restore") — cancelling/reaping the loser must leave the winner's
      `loom:building` claim intact when the forge timeline shows the winner's
      claim postdates the loser's `started_at`.
- [ ] Existing single-host `Superseded`/`HolderAlive` fast paths
      (`LockReleaseOutcome`, #4463/#4556) continue to short-circuit without an
      extra `gh api` round trip when the local lock already answers the
      question — the new forge-side check is an additional guard for the case
      the local lock *can't* see, not a replacement for the cheaper local
      check.

## Affected Files

- `loom-daemon/src/sweep_registry/reaper.rs` - `finish_cancel` (~line 731-757)
  and the natural-exit reap label-restore call (~line 928-970); add the
  forge-side claim-timestamp check before calling `restore_label_to_ready`
- `loom-daemon/src/sweep_registry/locks.rs` - `release_lock_owned` (~line 346)
  and `LockReleaseOutcome` (~line 79); either extend here or leave as the
  local-only check and add the forge-side check as a separate gate in `reaper.rs`
- `loom-daemon/src/sweep_registry/guards.rs` - `restore_label_to_ready` (~line
  561); candidate location for the forge-side timestamp check itself, or an
  additional parameter/pre-check the caller supplies
- `loom-daemon/src/claim_reconciliation.rs` - `fetch_claim_labeled_at` (~line
  1450) and `parse_max_timestamp` (~line 1488); reusable as-is (already
  issue/PR-agnostic) or worth promoting out of the private `forge` module if
  `reaper.rs` needs to call it directly

## Test Plan

- [ ] Automated tests: add a `reaper.rs` unit/integration test that constructs
      two `sweep_id`s for the same issue with no shared local lock (simulating
      two hosts), fakes/mocks the forge timeline response to show the
      surviving claim's `labeled loom:building` event postdating the
      cancelled sweep's `started_at`, and asserts `restore_label_to_ready` is
      NOT called (or its `gh issue edit` invocation is absent from the
      recorded command log, matching the existing `gh`-invocation-recording
      test pattern already used elsewhere in this module, e.g.
      `defaults/scripts/tests/test-merge-pr-partial-increment.sh`'s
      `assert_contains "$log" "issue edit ..."` style, adapted to Rust's own
      command-recording test doubles in `sweep_registry/test_support.rs`)
- [ ] Automated tests: existing `release_lock_owned_*` tests in `locks.rs`
      (lines ~1046-1148) continue to pass unchanged — confirms the local-lock
      fast path isn't regressed
- [ ] Manual verification: N/A for the cross-host race itself (requires two
      real daemon hosts) — automated test coverage above is the primary
      verification; if feasible, a manual two-VM/two-workspace repro
      mirroring this issue's original observation is a bonus, not required
- [ ] Edge cases: forge API failure/timeout during the new check (must fail
      open, not hang or crash); issue already closed by the time the check
      runs; `loom:building` labeled event never found (e.g. daemon-applied
      claims that predate this fix's rollout) — must not spuriously block a
      legitimate same-host restore

## Dependencies

None — this is a self-contained fix to `loom-daemon`'s cancel/reap label-restore
path. Related precedent (not blocking): #4463 (same-host `Superseded` check),
#4556 (`HolderAlive` live-process veto), #4275 (the incident that motivated
both).

rjwalters:
## Summary

Adds a forge-side, cross-host claim-ownership check before the daemon's
`finish_cancel` / `reap_once` label-restore paths remove `loom:building`.
`.loom/locks/issue-<N>` is host-local filesystem state — a peer host's live
claim on the same issue is structurally invisible to it — so a host cancelling
(or reaping) its own losing side of a cross-host duplicate dispatch could
previously restore `loom:building -> loom:issue` and destroy a still-live
peer host's claim, reopening the issue to a third dispatch (the loom#5270
incident, root-caused in #5282).

## Background / scope note

#5017 was originally filed with two distinct symptoms: a genuine live-vs-live
collision (issue #12 in the report) and a wedged sweep silently holding its
claim for 6.5h (issue #87). The operator's most recent disposition on #5017
(2026-08-04, after a third live reproduction on loom#5270) redefined this
issue's concrete scope to: *"claim-ownership verification before any
release/reclaim (timeline `labeled`-event check, judge-reclaim style), and
reclaim must never treat a live peer's claim as stealable on age alone —
liveness evidence... gates reclaim"* — explicitly shared with sibling issue
#5282's independently-curated acceptance criteria for the same defect. This PR
implements that redefined scope in full.

The original wedge-*detection*/escalation gap (#87) and the harder live-vs-live
dispatch-time race (#12) are **not** addressed here — split out to follow-up
issue #5302 (see "Deferred" below) so #5017 itself can close against its own
operator-redefined scope.

## Changes

- `loom-daemon/src/sweep_registry/guards.rs`: new `fetch_claim_labeled_at`
  (fetches the most recent `labeled loom:building` timeline event for an
  issue via `gh api .../timeline`, fail-open on any error) and
  `claim_superseded_on_forge` (true iff that event postdates this sweep's own
  claim/dispatch time).
- `loom-daemon/src/sweep_registry/reaper.rs`: wire the new check into all
  three label-restore call sites — `finish_cancel`, and `reap_once`'s
  checkpoint-present (crash) and no-checkpoint (clean-exit) branches — as an
  additional `||` guard alongside (never a replacement for) the existing
  host-local `Superseded`/`HolderAlive` check, so the extra `gh api` round
  trip is only paid when the local lock could not already answer the
  question.

| Criterion (operator-redefined scope) | Status | Verification |
| Claim-ownership verification before any release/reclaim (forge timeline `labeled`-event check) | ✅ | `fetch_claim_labeled_at` + `claim_superseded_on_forge`, wired into `finish_cancel`/`reap_once` before every `restore_label_to_ready` call |
| Reclaim must not treat a live peer's claim as stealable on age alone — liveness evidence gates it | ✅ | The forge's own `labeled loom:building` event, observed identically by every host, is the liveness evidence; a claim newer than this sweep's own `started_at` blocks the restore regardless of how "old"/wedged this sweep looks locally |
| Fail-open (a forge outage must not permanently wedge a claim) | ✅ | `fetch_claim_labeled_at` returns `None` on any `gh` failure/timeout/non-zero-exit/unparseable output; `claim_superseded_on_forge` treats `None` as "not superseded" (unit-tested via the pre-existing `cancel_restores_label_when_no_pr_produced`-style tests, whose fake `gh` returns empty stdout for the timeline call and still restore correctly) |
| Existing host-local fast path still short-circuits (no extra `gh` call when the local lock already answers) | ✅ | `||` short-circuit — `release_lock_owned(...).retained()` is evaluated first at every call site |
| Regression test: cancelling/reaping the loser must leave the winner's claim intact | ✅ | New tests below |

## Deferred (split to follow-up issue #5302, not this PR)

- **#87's wedge-*detection*/escalation gap**: a hung-sweep watchdog already
  exists (`sweep_registry/watchdog.rs`, #3887) with a bounded one-shot
  auto-restart, but after it gives up nothing further reaches an operator —
  it just logs. This PR does not add an escalation path beyond the existing
  log line.
- **The genuine live-vs-live dispatch-time race** (#12): two hosts both
  starting a *fresh* dispatch near-simultaneously, before either label exists
  to observe. The existing soft peer-claim advertisement (`peer_claims.rs`,
  #4028 Phase 1) is explicitly documented as "a fast backoff, not a lock";
  closing that race for real needs #4028 Phase 2 (a cross-host CAS), which
  was never built.
- Both items, plus the still-unrun #4146 baseline measurement, are now
  tracked by follow-up issue #5302 rather than left implicit on a closed
  #5017.

## Coordination note

This PR's fix is scoped identically to sibling issue #5282 (independently
curated, root-caused the same `finish_cancel`/`release_lock_owned`
host-local-lock gap in detail) — implemented here because #5017's own
operator disposition pointed at the same fix. Left a comment on #5282 so
whoever picks it up next can check this PR first rather than duplicate the
fix; not claiming `Closes #5282` here since that issue is independently
owned/claimed.

## Test Plan

- `cargo check -p loom-daemon` — clean.
- `cargo clippy -p loom-daemon --all-targets -- -D warnings` — clean.
- `cargo test -p loom-daemon --lib sweep_registry::reaper::tests::` — 52/52
  pass, including 2 new regression tests:
  - `cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host`
  - `reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host`
- `cargo test -p loom-daemon --lib sweep_registry::guards::tests::` — 14/14
  pass (unchanged).
- `cargo test -p loom-daemon --lib` (full suite, 3424 tests): passed in full
  on unmodified `main` (3385/3385). In this worktree it showed exactly 1
  flaky failure per run, but a **different** test each time
  (`forge_listing::tests::errors_carry_stderr_for_the_rate_limit_classifier`
  on one run, `sweep_registry::guards::tests::restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr`
  — pre-existing, untouched by this diff — on another), and both pass
  cleanly in isolation. Neither touches `forge_listing.rs`, which this PR
  never modifies. This matches known load-sensitive flakiness under the
  host's heavy concurrent-sweep load at the time of the run, not a
  regression from this change.

Closes #5017

### DIFF
--- a/loom-daemon/src/sweep_registry/guards.rs
+++ b/loom-daemon/src/sweep_registry/guards.rs
@@ -2,6 +2,7 @@
 //! open-PR / park-label label probes.
 
 use super::*;
+use crate::claim_reconciliation::forge::parse_max_timestamp;
 
 /// Three-state result of the open-linked-PR probe (Issue #4452), replacing the
 /// old `Option<u32>` that conflated a *verified* "no open linked PR" with a
@@ -136,6 +137,114 @@ impl SweepRegistry {
             CollisionClass::Collision { labels }
         } else {
             CollisionClass::Clean
+        }
+    }
+
+    // ------------------------------------------------------------------------
+    // Cross-host claim-ownership verification before release/reclaim
+    // (Issue #5017 / #5282)
+    // ------------------------------------------------------------------------
+    //
+    // `.loom/locks/issue-<N>` (see `locks.rs`'s `release_lock_owned`) is
+    // strictly HOST-LOCAL filesystem state: it is written by `acquire_lock`
+    // when *this* daemon dispatches *its own* sweep, and no other host's
+    // daemon ever sees it. That makes it structurally blind to a genuine
+    // cross-host race: when host B cancels its own losing duplicate dispatch
+    // for an issue host A is actively (and validly) building, host B's local
+    // lock names host B's own (about-to-be-cancelled) sweep as the owner —
+    // it matches, so `release_lock_owned` returns `Released`, not
+    // `Superseded`, and the caller proceeds to call `restore_label_to_ready`,
+    // destroying the ONLY cross-host mutex (the `loom:building` label)
+    // out from under host A's still-live sweep. This is exactly what
+    // happened on loom#5270 (2026-08-04): cancelling loom-worker-1's losing
+    // duplicate reverted `loom:building` on the issue robb-studio's sweep
+    // still owned, reopening it to a third dispatch.
+    //
+    // The forge's own label-event timeline, by contrast, is observed
+    // identically by every host — it is the one piece of claim state that is
+    // NOT host-local. `fetch_claim_labeled_at` / `claim_superseded_on_forge`
+    // below add that cross-host signal as an ADDITIONAL guard alongside (not
+    // a replacement for) the cheaper host-local `Superseded`/`HolderAlive`
+    // checks: every call site short-circuits on the existing local check
+    // first, so the extra `gh api .../timeline` round trip is only paid when
+    // the local lock could not already answer the question.
+
+    /// Fetch the most recent `labeled loom:building` timeline event timestamp
+    /// for `issue` (Issue #5017/#5282) — the forge-side, cross-host claim
+    /// signal every host observes identically, unlike the host-local
+    /// `.loom/locks/issue-<N>` claim lock.
+    ///
+    /// Mirrors [`crate::claim_reconciliation::forge`]'s own
+    /// `fetch_claim_labeled_at` (used there for PR-claim reconciliation) —
+    /// the underlying `issues/{n}/timeline` REST endpoint is identical for
+    /// issues and PRs, so the query shape is reused verbatim; this copy lives
+    /// in `sweep_registry` so the cancel/reap label-restore path (this
+    /// module) can call it without a cross-module `pub(crate)` promotion of a
+    /// function whose doc comments are specific to PR-claim reconciliation.
+    ///
+    /// FAIL-OPEN: returns `None` on any `gh` failure/timeout/non-zero
+    /// exit/unparseable output, or when the label was never applied. Callers
+    /// MUST treat `None` as "cannot verify, proceed with existing behavior"
+    /// — same fail-open contract as every other forge probe in this module
+    /// ([`classify_preflip_labels`](Self::classify_preflip_labels),
+    /// [`issue_is_closed_or_pr`](Self::issue_is_closed_or_pr)).
+    pub(crate) fn fetch_claim_labeled_at(&self, issue: u32) -> Option<DateTime<Utc>> {
+        let gh = self
+            .config
+            .gh_bin
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("gh"));
+        let mut cmd = Command::new(&gh);
+        cmd.arg("api")
+            .arg(format!("repos/{{owner}}/{{repo}}/issues/{issue}/timeline"))
+            .arg("--paginate")
+            .arg("--jq")
+            .arg(
+                r#"[.[] | select(.event == "labeled" and .label.name == "loom:building") | .created_at] | max // empty"#,
+            );
+        cmd.current_dir(&self.config.workspace_root);
+        if let Ok(repo) = std::env::var("LOOM_REPO") {
+            cmd.arg("--repo").arg(repo);
+        }
+        let timeout = reap_gh_timeout();
+        let output = output_with_timeout(cmd, timeout).ok().flatten()?;
+        if !output.status.success() {
+            return None;
+        }
+        parse_max_timestamp(&output.stdout)
+    }
+
+    /// Whether the forge's `loom:building` claim on `issue` was (re-)applied
+    /// STRICTLY AFTER `claimed_at` (Issue #5017/#5282) — i.e. a different
+    /// claimant, possibly on another host entirely invisible to this host's
+    /// `.loom/locks/issue-<N>`, has (re-)claimed the issue since this sweep's
+    /// own claim/dispatch time. When `true`, the caller MUST skip
+    /// [`restore_label_to_ready`](Self::restore_label_to_ready) — exactly the
+    /// same "leave the live claim alone" contract as a host-local
+    /// `Superseded`/`HolderAlive` verdict from `release_lock_owned`.
+    ///
+    /// FAIL-OPEN: an unverifiable read ([`fetch_claim_labeled_at`] returns
+    /// `None`) resolves to `false` (not superseded) — an unreachable forge
+    /// must never permanently wedge a claim, matching every other check in
+    /// this module's fail-open posture (see `restore_label_to_ready`'s own
+    /// doc comment).
+    ///
+    /// [`fetch_claim_labeled_at`]: Self::fetch_claim_labeled_at
+    pub(crate) fn claim_superseded_on_forge(&self, issue: u32, claimed_at: DateTime<Utc>) -> bool {
+        match self.fetch_claim_labeled_at(issue) {
+            Some(labeled_at) if labeled_at > claimed_at => {
+                log::warn!(
+                    "sweep_registry: issue #{issue}'s `loom:building` claim was (re-)applied at \
+                     {} — AFTER this sweep's own claim/dispatch time {} — leaving the label \
+                     alone instead of restoring it (#5017/#5282 cross-host claim-ownership \
+                     guard). A different claimant, possibly on another host, now owns this \
+                     issue; destroying its claim here would repeat the loom#5270 incident.",
+                    labeled_at.to_rfc3339(),
+                    claimed_at.to_rfc3339(),
+                );
+                true
+            }
+            _ => false,
         }
     }
 
--- a/loom-daemon/src/sweep_registry/reaper.rs
+++ b/loom-daemon/src/sweep_registry/reaper.rs
@@ -735,7 +735,20 @@ impl SweepRegistry {
             // sweep owns the claim and runs its own lifecycle. #4556 extends the
             // same skip to `HolderAlive`: the cancelled sweep's OWN pid is still
             // a live `/loom:sweep <N>` process, so the claim is not free either.
-            let superseded = self.release_lock_owned(*issue, sweep_id).retained();
+            //
+            // #5017/#5282: `release_lock_owned` only ever sees THIS host's own
+            // local `.loom/locks/issue-<N>` — a different host's live claim on
+            // the SAME issue is invisible to it (there is no shared lock
+            // directory across hosts), so a purely local check can return
+            // "Released" even while a peer host's sweep is actively building.
+            // `claim_superseded_on_forge` is the cross-host backstop: it only
+            // runs (short-circuits via `||`) when the local check did NOT
+            // already answer the question, and compares the forge's current
+            // `loom:building` labeled-event timestamp against this sweep's own
+            // `started_at` — a labeling event strictly after `started_at` means
+            // a different claimant (possibly on another host) now owns it.
+            let claim_held_elsewhere = self.release_lock_owned(*issue, sweep_id).retained()
+                || self.claim_superseded_on_forge(*issue, started_at);
             // Best-effort tidy-up of the machine-level liveness journal
             // (#3953) — this cancelled sweep no longer exists.
             self.journal_remove_best_effort(*issue);
@@ -749,9 +762,10 @@ impl SweepRegistry {
             // produced no PR, so we never yank the label out from under an
             // in-flight PR's issue. Gated on `!skip_label_flip`, mirroring the
             // reaper path. `SweepKind::PrSet` cancels never reach here.
-            // #4463: a superseded release means a newer sweep now holds the
-            // claim — never restore the label out from under it.
-            if !self.config.skip_label_flip && !produced_pr && !superseded {
+            // #4463/#5017/#5282: a claim held elsewhere (locally superseded OR
+            // forge-superseded) means a different sweep now holds the claim —
+            // never restore the label out from under it.
+            if !self.config.skip_label_flip && !produced_pr && !claim_held_elsewhere {
                 let _ = self.restore_label_to_ready(*issue);
                 self.note_label_flip(*issue); // #4485 flap detection
             }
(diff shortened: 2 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:09:38Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reclaiming stale loom:reviewing claim (age > ${LOOM_STALE_REVIEWING_MINUTES:-30}m, no follow-up comment) — a prior Judge's parent sweep likely died mid-review.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:42:07Z) — standing down without reclaiming. Not stomping.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict (Verdict-Time CAS Recheck) and found `loom:pr` already present, written by a different actor (loom-fleet-dispatch[bot] at 2026-08-04T16:53:35Z) — another Judge's verdict raced mine while I was running the full test suite (cargo check/clippy/fmt clean; `sweep_registry::reaper::tests::` 52/52 incl. both new regression tests; `sweep_registry::guards::tests::` 14/14 — all independently reproduced, matching the PR's own claims). Discarding my review; not writing any label, per protocol (do not overwrite a landed verdict even when I agree with it).

Note for the record: my earlier reclaim of the `loom:reviewing` claim at 16:42:07Z (age past the 30-minute default, zero genuine comments at the time) appears to have been a false positive — the original claim was apparently still a live, slow-running review (this PR's full quality-gate suite took roughly 45 minutes wall clock here too, well past the 5-15 minute window the staleness heuristic assumes), not a dead process. No harm done since the label churn didn't block the eventual correct verdict, but flagging in case the review-duration assumption behind LOOM_STALE_REVIEWING_MINUTES needs revisiting for Rust-suite PRs specifically.

PR #5280, opened 2026-08-04
fix(gitignore): ignore .claude/worktrees/ so git add -A can't stage a gitlink

rjwalters:
## Summary

`.claude/worktrees/` — the root the Claude Code harness's `isolation: worktree`
mechanism creates **inside the checkout** — was missing from `EPHEMERAL_PATTERNS`
and from loom's own committed `.gitignore`. Each nested worktree therefore sat
untracked-and-unignored, and since it is itself a git repo, a blanket `git add -A`
stages it as an **embedded-repo gitlink**. That near-miss already happened once in
a consumer repo (gf180-trng) during a resync and was only caught by hand.

`.loom/worktrees/` and `.loom/worktrees-local/` were already covered; this closes
the remaining hole for the harness-owned worktree root.

## Changes

- **`loom-daemon/src/init/post_init.rs`** — added `.claude/worktrees/` to
  `EPHEMERAL_PATTERNS`, positioned next to `.loom/worktrees/` /
  `.loom/worktrees-local/`, with a comment explaining it is the *harness's*
  `isolation: worktree` mechanism (distinct owner from Loom's own builder
  worktrees, same hazard class). The comment also records why this differs from
  `.claude/commands` / `.claude/skills`, which the source `.gitignore` explicitly
  documents as loom-only ignores that must **not** enter the managed block
  (consumers keep those tracked) — `.claude/worktrees/` is machine-local runtime
  state in every repo, so it does belong there. Added the matching `#5267:` entry
  to the `EPHEMERAL_PATTERNS` doc-comment changelog, per the existing
  `#4280:`/`#5014:` convention.
- **`.gitignore`** — added `.claude/worktrees/` inside the loom-managed block at
  the same position, keeping the committed source file in sync with
  `EPHEMERAL_PATTERNS` (the source-of-truth convention documented in
  `post_init.rs`).
- **Tests** — extended the three existing gitignore tests (no new tests, no new
  plumbing).

No backfill code path was added: `defaults/scripts/resync-installed.sh:976`
already runs `"$bin" update-gitignore "$REPO_ROOT"`, which rebuilds the
marker-delimited block from `EPHEMERAL_PATTERNS` in place, so existing installs
converge on the next resync.

## Acceptance Criteria Verification

1. **`.claude/worktrees/` added to `EPHEMERAL_PATTERNS` with an explanatory
   comment** — done (`post_init.rs`, immediately after `.loom/worktrees-local/`).
   The comment states it is the harness's `isolation: worktree` mechanism and
[...]

Issue #5267, opened 2026-08-04
Installer/resync don't manage a .claude/worktrees/ gitignore rule — git add -A commits an embedded-repo gitlink


rjwalters:
## Context

During the loom#5184 refresh (2026-08-04), committing the resync in gf180-trng with `git add -A` swept in `.claude/worktrees/agent-ab011969253709e56` as an **embedded-repo gitlink** — git prints only an easily-missed advice block, and the gitlink landed in the commit (removed by hand before push). gf180-bandgap was protected only because someone had hand-added a `.claude/worktrees/` rule above the loom-managed block; gf180-trng had no such rule.

`.claude/worktrees/` is created inside the main checkout by the harness's `isolation: worktree` mechanism — same role as `.loom/worktrees/`, which IS in the loom-managed `.gitignore` block. Any consumer repo where an agent runs with worktree isolation can hit this.

## Suggested acceptance criteria

- [ ] The loom-managed `.gitignore` block (installer scaffolding + the `resync-installed.sh` `.gitignore` refresh path) includes `.claude/worktrees/`.
- [ ] Resync backfills the rule into existing installs (same mechanism that maintains the rest of the managed block).
- [ ] A test covers the rule's presence after install and after resync.

---

### Verification (as of `6ff58955`, 2026-08-04 — re-verify before relying on these numbers/paths)

Confirmed both halves of the claim directly:

- `loom-daemon/src/init/post_init.rs`'s `EPHEMERAL_PATTERNS` (the single source of truth for the loom-managed `.gitignore` block, per its own doc comment at line ~88) has `.loom/worktrees/` and `.loom/worktrees-local/` but **no** `.claude/worktrees/` entry.
- This repo's own `.gitignore` (which the `EPHEMERAL_PATTERNS` doc comment says must be "kept in sync with the Loom source repo's `.gitignore`") also has no `.claude/worktrees/` rule — so the loom source repo itself is exposed to the same near-miss whenever an agent here runs with `isolation: worktree` (the `Workflow` tool's worktree-isolation mode, distinct from Loom's own `.loom/worktrees/` builder mechanism).
[...]

Issue #4997, opened 2026-08-03
locate_daemon_bin prefers the installed binary over a fresh repo-local build — devs silently run a stale daemon


rjwalters:
## Problem

`loom_locate_daemon_bin` resolves the *installed* binary before any repo-local
build output, so a developer working in a checkout silently runs a stale
`~/.local/bin/loom-daemon` instead of the binary they just built.

Precedence today (`defaults/scripts/lib/locate-daemon-bin.sh:40-61`):

1. `$LOOM_DAEMON_BIN`
2. `loom-daemon` on `$PATH`
3. `${LOOM_DAEMON_BIN_DIR:-$HOME/.local/bin}/loom-daemon`
4. `<repo_root>/loom-daemon/target/{release,debug}/loom-daemon`, then `<repo_root>/target/{release,debug}/loom-daemon`

Steps 2 and 3 both resolve the machine-level install on any normal dev host, so
step 4 is effectively dead code in a checkout. `cargo build --release` followed by
any script that sources this lib (`loom-daemon-start.sh`, `loom-daemon-update.sh`,
`loom-status.sh`, `probe-tokens.sh`, `.loom/bin/loom`) runs the *old* binary with
no warning that it did so.

## Why this is easy to lose hours to

The failure is silent and inverted from expectation: the developer's change is
compiled and on disk, the command runs and succeeds, and the behavior is the old
behavior. Nothing in the output names which binary ran.

## Suggested direction (not prescriptive)

Options, roughly in increasing intrusiveness:

- Log the resolved path + its mtime whenever resolution lands on a candidate,
  so "which binary ran" is answerable from any log.
- Add an opt-in `LOOM_PREFER_REPO_BUILD=1` that hoists step 4 above steps 2-3.
- Prefer a repo-local build over the install *when it is newer* and `<repo_root>`
  is the loom source checkout itself (i.e. `loom-daemon/Cargo.toml` exists).

Whatever is chosen, `loom_daemon_bin_search_paths` must be updated in lockstep —
the file keeps the two side by side precisely so they cannot drift
(`locate-daemon-bin.sh:66-77`).

## Acceptance criteria

- [ ] A developer with both `~/.local/bin/loom-daemon` and a fresh
      `target/release/loom-daemon` can tell from output alone which one ran.
- [ ] Whatever precedence change lands is mirrored in
      `loom_daemon_bin_search_paths` and covered by a test that fails if the two
      functions disagree.
[...]

PR #5303, opened 2026-08-04
fix(scripts): stop .gitignore resync from trusting a stale loom-daemon binary

### LINKED ISSUE #5294: Resync regenerated .gitignore without .claude/worktrees/, reintroducing the #5267 gitlink hazard
## Observed

The repo-root `.gitignore` on `main` is currently missing the `.claude/worktrees/` entry, even though `loom-daemon/src/init/post_init.rs`'s `EPHEMERAL_PATTERNS` (the single source of truth for the managed block) still contains it:

```
git show origin/main:.gitignore | grep -n -B2 -A2 worktrees
# 53:.loom/worktrees/
# 54:.loom/worktrees-local/
# 55:.loom/state.json          <-- .claude/worktrees/ is gone
```

Timeline on `main`:
- `303bba09` (2026-08-04 07:55:19 -0700) — PR #5280 merges, adding `.claude/worktrees/` to both `EPHEMERAL_PATTERNS` and the committed `.gitignore`, closing issue #5267.
- `05cf67e8` (2026-08-04 08:29:48 -0700, ~34 min later) — "chore(loom): resync installed copies from defaults after main pull" runs `resync-installed.sh`, which regenerates the managed `.gitignore` block via `refresh_gitignore_block()` / `loom-daemon update-gitignore` — and the regenerated block no longer contains `.claude/worktrees/`:
  ```
  git show 05cf67e8 -- .gitignore
  # -.claude/worktrees/
  ```

This silently reintroduces the exact hazard #5267/#5280 fixed: an unignored `.claude/worktrees/<name>` nested git repo that `git add -A` stages as an embedded-repo gitlink.

## Suspected cause (unverified)

`refresh_gitignore_block()` in `.loom/scripts/resync-installed.sh` resolves a `loom-daemon` binary via `loom_locate_daemon_bin` (`.loom/scripts/lib/locate-daemon-bin.sh`), whose default precedence (without `LOOM_PREFER_REPO_BUILD=1`) prefers `$LOOM_DAEMON_BIN` / `PATH` / the machine-level install (`$HOME/.local/bin/loom-daemon`) over a fresh build from the just-pulled source. If the resolved binary was built *before* #5280's source change, `update-gitignore` would rebuild the managed block from that binary's stale, compiled-in `EPHEMERAL_PATTERNS` — one that doesn't yet know about `.claude/worktrees/` — and drop the line even though the just-pulled source `post_init.rs` has it.

This is a hypothesis from reading `resync-installed.sh` and the timing (a stale-binary resync run ~34 min after the pattern was added, doing exactly what a stale binary would do), not a captured log confirming which binary path was actually used in that run. Needs verification, e.g. by checking `loom-daemon --version`/mtime logging around that resync, or by reproducing: build `loom-daemon`, remove a pattern from `EPHEMERAL_PATTERNS`, install the stale binary to the machine-level location, add the pattern back to source only, then run `resync-installed.sh` and confirm the newly-added pattern gets dropped from `.gitignore`.

## Impact

`.gitignore` drift is silent — nothing fails loudly when a resync regresses a managed pattern. Any `git add -A` in this repo (or any consumer repo hitting the same stale-binary resync path) between the regression and the next fix can stage a gitlink for `.claude/worktrees/`.

## Suggested fix direction (not verified)

`refresh_gitignore_block()` could require (or prefer) a freshly-built binary when the resync is running specifically *because* source changed (i.e. hoist `LOOM_PREFER_REPO_BUILD=1` semantics into this specific call site, or rebuild `loom-daemon` from `$SOURCE_ROOT` before invoking `update-gitignore`), rather than trusting whatever binary happens to be resolved first.

## Immediate repo-hygiene need

Independent of the root cause, `main`'s `.gitignore` should get `.claude/worktrees/` re-added now — it's currently exposed to the exact gitlink hazard #5267 described.

Discovered while evaluating (now-closed, superseded-by-#5280) PR #5286.

rjwalters:
## Summary

`refresh_gitignore_block()` in `resync-installed.sh` resolves the `loom-daemon`
binary it uses to regenerate the Loom-managed `.gitignore` block via the
resolver's default precedence (`$LOOM_DAEMON_BIN` -> `$PATH` ->
machine-level install -> repo build). That precedence can pick a binary
compiled **before** a just-pulled source change to `EPHEMERAL_PATTERNS`,
silently dropping that pattern from the regenerated block.

This is confirmed, not just hypothesized: commit `05cf67e8` ("resync
installed copies from defaults after main pull") dropped `.claude/worktrees/`
from the committed `.gitignore` even though the source commit it was
resyncing to (`850f9ad1`) already had `.claude/worktrees/` in
`EPHEMERAL_PATTERNS` (added by #5280, merged ~34 minutes earlier). The
machine-level `loom-daemon` install on this very host (`~/.local/bin/loom-daemon`)
is currently that stale pre-#5280 binary — verified directly with
`strings ~/.local/bin/loom-daemon | grep claude/worktrees` returning nothing.
This silently reintroduced the exact `.claude/worktrees/` gitlink hazard
#5267/#5280 fixed.

## Changes

- **Immediate hygiene fix**: re-add `.claude/worktrees/` to the repo-root
  `.gitignore`'s loom-managed block.
- **Root-cause fix** (`defaults/scripts/resync-installed.sh`,
  `refresh_gitignore_block()`):
  - Hoist `LOOM_PREFER_REPO_BUILD=1` for this one call site only (via a
    subshell env override passed into `loom_locate_daemon_bin`), so a
    repo-local build is preferred over a possibly-stale `$PATH`/machine-level
    install for *this* refresh — since the resync runs specifically because
    source just changed. An explicit `$LOOM_DAEMON_BIN` still wins regardless.
    `locate-daemon-bin.sh`'s own default precedence is unchanged for other
    callers (`loom-daemon-start.sh`, etc. — per its own precedence-hoist
    rationale comment).
  - Added defense-in-depth: after `update-gitignore` runs, verify the
    resulting `.gitignore`'s managed block actually contains every pattern
    `EPHEMERAL_PATTERNS` declares in the source tree being resynced
    (extracted straight from `loom-daemon/src/init/post_init.rs`, not from
    a compiled binary). Any pattern the resolved binary dropped is now named
    in a loud, specific `WARN`, never a silent no-op.
- **Tests** (`defaults/scripts/tests/test-resync-installed.sh`, new groups
  12h/12i): reproduce the stale-binary-drop failure mode using fake
  `loom-daemon` shell-script stand-ins (so the test doesn't need to compile
  two different daemon binaries) and assert:
  - 12h: with a stale `$PATH` binary + a fresh repo-local build both present,
    the repo build is preferred and the newly-added pattern lands correctly,
    with no spurious warning.
  - 12i: with only the stale binary resolvable, the missing pattern is named
    in a loud warning instead of being silently dropped.
  - Verified both new tests actually catch the regression: reverting only
    the `resync-installed.sh` fix (keeping the new tests) makes both fail;
    restoring the fix makes them pass again.

- [x] `.claude/worktrees/` is present in the committed `.gitignore` again.
- [x] The "suspected cause" hypothesis is verified with direct evidence
  (source commit already had the pattern; the resolved machine-level binary
  on this host does not) rather than taken on faith.
- [x] The resync call site no longer trusts a stale binary silently for this
  call, without changing `loom_locate_daemon_bin`'s global default
  precedence.
- [x] Regression tests reproduce the failure mode and pass with the fix,
  fail without it (verified manually, see below).
- [x] `bash defaults/scripts/tests/test-resync-installed.sh` passes:
  127/127 (122 pre-existing + 5 new).

## Test Plan

- `bash defaults/scripts/tests/test-resync-installed.sh` — 127/127 passed.
- `bash defaults/scripts/tests/test-locate-daemon-bin.sh` — 20/20 passed
  (unmodified library, confirms no regression to its own precedence
  contract).
- `shellcheck --severity=warning --format=gcc defaults/scripts/resync-installed.sh
  defaults/scripts/tests/test-resync-installed.sh` — clean (CI's exact
  invocation).
- `git show origin/main:.gitignore | grep -c '^\.claude/worktrees/$'` on this
  branch's `.gitignore` returns `1`.
- Manually reverted the source fix while keeping the new tests: both new
  tests failed as expected, confirming they actually reproduce the
  regression rather than passing vacuously.

Closes #5294

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ Thumbs.db
 .loom/interventions/
 .loom/worktrees/
 .loom/worktrees-local/
+.claude/worktrees/
 .loom/state.json
 .loom/mcp-command.json
 .loom/activity.db
--- a/defaults/scripts/resync-installed.sh
+++ b/defaults/scripts/resync-installed.sh
@@ -942,6 +942,55 @@ ensure_install_metadata_merge_driver() {
 }
 ensure_install_metadata_merge_driver
 
+# #5294: EPHEMERAL_PATTERNS is single-sourced in
+# loom-daemon/src/init/post_init.rs. Extract the pattern list directly from
+# that source file (not from a compiled binary) so refresh_gitignore_block can
+# verify the freshly-regenerated .gitignore against ground truth, independent
+# of which loom-daemon binary happened to write it. Best-effort: echoes
+# nothing (not a failure) if $SOURCE_ROOT isn't a full checkout with the daemon
+# crate present (e.g. a stripped-down install source).
+_gitignore_source_ephemeral_patterns() {
+    local post_init="$SOURCE_ROOT/loom-daemon/src/init/post_init.rs"
+    [[ -f "$post_init" ]] || return 0
+    awk '/pub const EPHEMERAL_PATTERNS/ { flag=1; next } flag && /^\];/ { exit } flag' "$post_init" \
+        | sed -n -E 's/^[[:space:]]*"([^"]*)",?[[:space:]]*$/\1/p'
+}
+
+# #5294: verify the managed block `$bin update-gitignore` just wrote actually
+# contains every pattern the CURRENT source declares. A `loom-daemon` binary
+# older than the just-pulled source has EPHEMERAL_PATTERNS compiled in from
+# whenever it was built — if that predates a pattern addition (e.g. #5280's
+# `.claude/worktrees/`), `update-gitignore` exits 0 having *silently* dropped
+# that pattern from the regenerated block. That is precisely how 05cf67e8
+# reintroduced #5267's gitlink hazard 34 minutes after #5280 fixed it, entirely
+# undetected until this issue. Never trust the exit code alone: name any
+# dropped pattern in a loud warning instead of a silent no-op.
+_gitignore_warn_if_stale() {
+    local bin="$1" post_init="$SOURCE_ROOT/loom-daemon/src/init/post_init.rs"
+    local -a missing=()
+    local pattern block
+
+    [[ -f "$post_init" ]] || return 0
+    [[ -f "$REPO_ROOT/.gitignore" ]] || return 0
+
+    block="$(sed -n '/# >>> loom-managed/,/# <<< loom-managed/p' "$REPO_ROOT/.gitignore")"
+    [[ -n "$block" ]] || return 0
+
+    while IFS= read -r pattern; do
+        [[ -n "$pattern" ]] || continue
+        grep -qxF -- "$pattern" <<<"$block" || missing+=("$pattern")
+    done < <(_gitignore_source_ephemeral_patterns)
+
+    if [[ "${#missing[@]}" -gt 0 ]]; then
+        warn "The resolved loom-daemon binary ($bin) regenerated .gitignore WITHOUT ${#missing[@]} pattern(s) that $post_init currently declares:"
+        for pattern in "${missing[@]}"; do
+            warn "    $pattern"
+        done
+        warn "  '$bin' is likely older than the source just synced (#5294) — rebuild loom-daemon"
+        warn "  (cargo build --release -p loom-daemon under $SOURCE_ROOT) and re-run resync-installed.sh."
+    fi
+}
+
 refresh_gitignore_block() {
     local locate_lib bin
     locate_lib="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/lib/locate-daemon-bin.sh"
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:12:16Z) — standing down without reclaiming. Not stomping.

rjwalters:
Judge review: **Approved**.

Verified the diff matches the PR description exactly:
- `.gitignore` restores `.claude/worktrees/` to the loom-managed block.
- `refresh_gitignore_block()` in `defaults/scripts/resync-installed.sh` now resolves the daemon binary via `LOOM_PREFER_REPO_BUILD=1 loom_locate_daemon_bin "$SOURCE_ROOT"`. Confirmed empirically that this is a scoped, non-leaking env override — a prefix assignment before a shell *function* call in bash sets the var for that call only and does not persist in the calling shell afterward (tested directly: `LOOM_PREFER_REPO_BUILD=1 myfunc` sees the var inside, unset again right after). `locate-daemon-bin.sh`'s own `LOOM_PREFER_REPO_BUILD` opt-in precedence tier already existed (#4997) — this PR is the first caller to actually flip it on, and only for this one call site; `loom-daemon-start.sh` and friends still call the resolver with no override, so their default precedence is untouched.
- `_gitignore_warn_if_stale()` parses `EPHEMERAL_PATTERNS` straight from `loom-daemon/src/init/post_init.rs` (source of truth, not a compiled binary) and diffs it against the just-written managed block, warning loudly (naming the exact missing pattern(s)) instead of a silent no-op — this directly targets the root cause (a stale resolved binary silently dropping a newly-added pattern), independent of whether the `LOOM_PREFER_REPO_BUILD` hoist itself ever fails to find a fresh build.

Testing (all reproduced locally, matches the PR's reported numbers):
- `bash defaults/scripts/tests/test-resync-installed.sh` -> 127/127 passed.
- `bash defaults/scripts/tests/test-locate-daemon-bin.sh` -> 20/20 passed (resolver's own precedence contract unregressed).
- `shellcheck --severity=warning --format=gcc defaults/scripts/resync-installed.sh defaults/scripts/tests/test-resync-installed.sh` -> clean.
- `check-main-clean.sh` -> main worktree clean, no sweep contamination.
- Independently verified the regression-catching claim: reverted only `resync-installed.sh`'s fix (keeping the new tests) -> 125/127, with the two failures being exactly the new 12h/12i groups; restoring the fix returns to 127/127.

CI: all 13 checks green (`Shell Test Suites (hermetic)` was still in progress at first look; re-polled and it completed successfully). Merge state is `clean`/mergeable.

No security concerns -- this only touches local git tooling script logic and a comment-parsing regex against a local source file, no untrusted input. No edge cases missed: the `_gitignore_warn_if_stale` guard has its own early-return best-effort fallbacks (missing `post_init.rs`, missing `.gitignore`, missing managed block) so it degrades to a no-op rather than erroring on stripped-down installs, matching the rest of the script's defensive style.

(Posted via REST fallback -- GraphQL quota was exhausted at review time.)

rjwalters:
Standing down: this PR was already merged (16:18:55Z) before my claim landed — my queue snapshot was stale. Removing the stray loom:reviewing label I added; not writing any verdict.

PR #5283, opened 2026-08-04
fix(daemon): surface per-repo token-pool ranking staleness in health/status

rjwalters:
## Summary

`loom-daemon health` (and `status`) reported token-pool ranking staleness for only a single "anchored primary workspace" pool (`resolve_tokens_dir_anchored` on the daemon's launch CWD / `fallback_root`), even though the token-ranking self-refresh loop (`spawn_multi_token_ranking_refresh_task` in `token_ranking_refresh.rs`) already refreshes **every registered repo's own pool independently** via `resolve_tokens_dir(&repo.root)`. An operator running `loom-daemon health` from a repo other than the daemon's own launch-CWD repo got a health verdict about the wrong pool, while their own repo's actual (correctly-refreshed) `.ranking` freshness went unreported entirely.

This closed the gap between the two mechanisms:

- **`loom-daemon/src/types.rs`**: `RepoStatus` gains `token_pool_dir: Option<PathBuf>`, `ranking_present: bool`, `ranking_age_secs: Option<u64>` (all `#[serde(default)]` for wire compatibility).
- **`loom-daemon/src/ipc.rs`**: the per-repo status-construction loop now resolves each repo's own pool via the *unanchored* `resolve_tokens_dir(&repo.root)` — the same resolution the self-refresh loop uses — and populates the new fields. The existing top-level `fallback_root`-anchored fields (`token_pool_dir`, `capacity`) are unchanged.
- **`loom-daemon/src/capacity.rs`**: new shared `ranking_file_state(pool_dir) -> (bool, Option<u64>)` probe, used both by `ipc.rs`'s per-repo snapshot and by `cli/health.rs`'s existing single-pool probe (which used to duplicate this logic as a private `ranking_state` fn).
- **`loom-daemon/src/health.rs`** (`assess_tokens`): the tokens section detail JSON now names the evaluated `pool_path` (AC2), adds a `per_repo` array (root/pool_path/ranking_present/ranking_age_secs/stale per registered repo), and folds a bounded `"N of M registered repos have their own pool's .ranking stale/missing"` note into the summary + verdict when any registered repo's own pool is stale or missing — even when the top-level anchored pool is fresh.
- **`.loom/docs/token-pool.md`** / **`defaults/docs/daemon-reference.md`**: document the daemon-CWD-vs-operator-repo distinction, explain why the "refresh from `$HOME`" workaround happened to "fix" the old single-pool reading, and point at the new per-repo surface for diagnosis.

## Test plan

[...]

PR #5304, opened 2026-08-04
feat(daemon): remove token axis from admission, add RAM headroom + retuned CPU brake ("dumb mode")

### LINKED ISSUE #5270: Token axis caps at one sweep per healthy account — make it utilization-weighted so fresh accounts contribute multiple slots
## Context

`capacity.rs::token_axis_limit` caps concurrent sweeps at **one per healthy account** (`snap.available`). The guard rationale is sound (#3902: never dispatch onto an exhausted account; keep account failures uncorrelated; keep per-account exhaustion prediction accurate) — but the policy is **utilization-blind**: 6 accounts at 0% of their 5-hour window cap the fleet at 6 sweeps, leaving most of the pool's real headroom unused. Operator observation 2026-08-04: with 15 accounts (6 healthy) the fleet idled at a cap the healthy accounts could comfortably exceed.

The module header already flags the missing input: "sub-`exhausted` utilization thresholds are a documented follow-up" — the richer per-account utilization JSON exists (`loom-daemon tokens check --json`; claude-monitor's usage.db).

## Suggested direction

Replace the 1:1 rule with a **utilization-weighted per-account contribution**: each available account contributes `max(1, floor(K × remaining_5h_headroom))` slots (K configurable, default ~3), summed into the token axis. Degenerates to today's behavior when accounts are near-exhausted; roughly K× throughput when the pool is fresh. Preserves all three original guarantees:
- exhausted accounts still contribute 0;
- load still spreads across accounts via the spawn-time ranking selector, so failures stay mostly uncorrelated;
- headroom-aware slots keep the exhaustion signal predictive.

Cheap first step: a `sweepsPerHealthyAccount` config multiplier (flat K) with the utilization-weighted version as the follow-up once the axis consumes the utilization JSON.

- [ ] Token axis can exceed the healthy-account count when accounts have 5h-window headroom (or a flat configured multiplier as the first increment).
- [ ] An exhausted/blocked account still contributes zero slots.
- [ ] The pressure advisory and health `tokens` section report the derived slot count, not just the account count.

---

## Operator direction (2026-08-04) — supersedes the multiplier suggestion above

> I can provide a single API token that is backed by millions of dollars in API credit and we should be able to run 100s of parallel tasks. Remove these artificial limits to parallelism — the automatic limits have only stalled our progress. Make the system dumber for a while until we understand the problem better: as long as we have disk space for the worktree and our few-min average CPU is not 95%+, we can launch another parallel worker.

Reframed scope, in priority order:

1. **API-key mode disables the token axis entirely.** The axis models 5h/7d subscription-account windows; a metered `ANTHROPIC_API_KEY` has neither, so `token_axis_limit` is meaningless there. When the spawn path authenticates via API key (or a `capacity.tokenAxis: off` config override), the dynamic cap is `min(configured max, disk headroom, CPU headroom)` — no token term.
2. **Restore a CPU-headroom admission term** (removed in #4512; `cpu_headroom.rs` still exists): defer dispatch only when few-minute load average is ≥ ~95% of cores. This plus disk headroom is the whole admission gate in dumb mode.
3. The utilization-weighted multiplier (original body above) remains relevant **only** for the OAuth-pool path, as a follow-up — not the priority.

Revised acceptance criteria:

- [ ] The token axis contributes **no parallelism limit on any auth path** — API key or subscription pool. Token health (`.ranking`) remains a spawn-time *selection* input (prefer fresher accounts, skip revoked/blocked ones), never a concurrency cap.
- [ ] CPU admission term restored: dispatch defers at ≥95% few-min average, resumes below.
- [ ] RAM headroom joins the gate set alongside disk (defer when available memory is critically low, same posture as disk headroom).
- [ ] Health/status show which machine axis (max/disk/RAM/CPU) binds, so "dumb mode" is observable.

### Operator clarification (2026-08-04, second pass)

> We can have extra usage on our subscription tokens too, so the point stands — we should only ever limit parallelism based on the machine disk/RAM/CPU.

With overage enabled, subscription accounts don't hard-stop at their windows either, so the earlier "OAuth-pool behavior unchanged" carve-out is dropped: the axis is removed as a limiter everywhere, not gated behind API-key mode.

---

### Verification (as of `6ff58955`, 2026-08-04 — re-verify file:line references before relying on them)

Confirmed `capacity::token_axis_limit` (`loom-daemon/src/capacity.rs:336`) exists as described. Confirmed #4512 ("simplify concurrency: one per-machine max-concurrency knob; drop the CPU-headroom formula...") is merged/closed, matching the issue's "removed in #4512" claim.

**Important correction to the operator direction's premise (item 2, "Restore a CPU-headroom admission term ... removed in #4512")**: `#4512` removed the CPU term from the concurrency-cap *formula*, but a **later** PR, **#4903, already added a CPU-based admission brake** (`loom-daemon/src/admission_brake.rs`) that does close to what's being asked for — it is enabled by default (`DEFAULT_ADMISSION_BRAKE_ENABLED = true`, `admission_brake.rs:88`) and holds new admissions once-per-tick when `cpu_headroom::is_host_saturated` reports the host over a configurable `load_per_core_threshold` (`admission_brake.rs:236`). The only gap versus this issue's request is the **default threshold**: `admission_brake` defaults to `4.0` load/core (`DEFAULT_ADMISSION_BRAKE_LOAD_PER_CORE`, `admission_brake.rs:103`), deliberately generous per its own doc comment (`admission_brake.rs:92`) — while the operator's "≥95% few-min average" maps to roughly `~0.95` load/core, much closer to the *build gate's* separate threshold (`cpu_headroom::DEFAULT_GATE_LOAD_THRESHOLD = 0.9`, `cpu_headroom.rs:219`, used by `main_health_gate.rs` for an unrelated purpose). **This means AC2 is a retune/reuse task, not new-mechanism work**: lower `admission_brake`'s default (or add a dedicated "dumb mode" config override) rather than building a new CPU-headroom gate from scratch. Confirm this on a live host before treating it as settled — the two thresholds serve different gates today, and collapsing them (vs. adding a third, dumb-mode-specific one) is a design choice the Builder should make deliberately, not by accident.

No RAM/memory-headroom module exists yet (`ls loom-daemon/src/*headroom*` finds only `cpu_headroom.rs` and `disk_headroom.rs`) — AC3 ("RAM headroom joins the gate set") is genuinely net-new work. `disk_headroom.rs`'s shape (`disk_headroom_limit(repo_root: &Path) -> usize`, `disk_headroom.rs:144`, built on `worktree_root_free_gb` + a per-worktree-GB divisor) is the closest existing model to mirror for a `ram_headroom_limit`.

N/A — this is a scope/design issue driven directly by explicit operator instruction (see the two operator comments already embedded in the body above), not a bug requiring root-causing.

### Scope note: latest operator clarification is authoritative

The issue body contains **three superseding layers**, oldest to newest: (1) the original utilization-weighted-multiplier proposal, (2) a first operator direction reframing around API-key vs. subscription-pool auth mode, (3) a second operator clarification dropping the auth-mode carve-out entirely ("we should only ever limit parallelism based on the machine disk/RAM/CPU"). **Layer 3 is the authoritative scope** — it explicitly supersedes layer 2's API-key/subscription distinction, which in turn explicitly superseded layer 1. A Builder implementing this issue should build against the consolidated Acceptance Criteria below (which reflects layer 3), not layer 1's multiplier design or layer 2's auth-mode gating — both are historical context, not the target.

### Acceptance Criteria (consolidated, reflecting the latest operator clarification)

- [ ] The token axis (`capacity::token_axis_limit`) contributes **no parallelism limit on any auth path** — API key or subscription pool, unconditionally (not gated behind an API-key-mode check). `.ranking`-based token health remains a spawn-time *selection* input only (prefer fresher/healthier accounts, skip revoked/blocked ones via the existing bad-tokens machinery) — never a concurrency cap.
- [ ] CPU admission term: confirm whether retuning `admission_brake`'s existing `load_per_core_threshold` (default `4.0`, `admission_brake.rs:103`) to ~`0.95` is sufficient, or whether a distinct "dumb mode" threshold/config key is warranted so the build-gate's own `0.9` threshold and this admission threshold don't get conflated. Either way, dispatch must defer at ≥95% few-min load average and resume below it.
- [ ] RAM headroom joins the gate set alongside disk (net-new module, modeled on `disk_headroom.rs`'s shape — defer when available memory is critically low, same posture/precedent as disk headroom's existing behavior).
- [ ] Health/status output shows which machine axis (max/disk/RAM/CPU) is currently binding, so "dumb mode" behavior is observable to an operator, not just inferred from dispatch logs.
- [ ] The removed/disabled token-axis limiter does not silently break the accounting `token_pool_size` and related health/pressure-advisory reporting still rely on (issue #5269, currently in-flight, touches some of this same reporting surface) — verify the two changes don't collide.

### Affected Files

- `loom-daemon/src/capacity.rs` - `token_axis_limit` (line 336), where the token term is currently folded into the dynamic cap; needs to stop contributing a limit
- `loom-daemon/src/admission_brake.rs` - existing CPU admission brake (`DEFAULT_ADMISSION_BRAKE_LOAD_PER_CORE`, line 103; `AdmissionBrakeConfig`, line ~108); primary candidate for the CPU-term retune rather than new code
- `loom-daemon/src/cpu_headroom.rs` - `is_host_saturated`, `DEFAULT_GATE_LOAD_THRESHOLD` (line 219); referenced by both `admission_brake` and the unrelated build gate — keep the two uses distinguishable
- `loom-daemon/src/disk_headroom.rs` - existing disk-headroom implementation to mirror for the new RAM-headroom module
- New file (e.g. `loom-daemon/src/ram_headroom.rs`) - net-new RAM-headroom limiter, modeled on `disk_headroom.rs`
- `loom-daemon/src/health.rs` - status output; needs a "which axis is binding" indicator
- `loom-daemon/src/work_finder.rs` - dispatch/admission call sites (`resolve_dynamic_max_concurrent` and related), where the token axis is currently consumed as a cap input

### Test Plan

- [ ] Automated tests: token axis removal verified by a test asserting `token_axis_limit`'s output no longer bounds the dynamic cap regardless of `.ranking` state (exhausted-all-accounts case included); admission-brake retune verified by extending the existing `admission_brake.rs` test suite (see `cfg(enabled, threshold)` test helper, `admission_brake.rs:495`) with a case at the new default threshold; new RAM-headroom module gets its own unit tests mirroring `disk_headroom.rs`'s existing test shape
- [ ] Manual verification: on a host with the change applied, drive load past the new CPU threshold and confirm dispatch defers; confirm RAM headroom defers dispatch when available memory is critically low; confirm `loom-daemon health`/status names the currently-binding axis
- [ ] Edge cases: confirm removing the token axis doesn't starve the bad-tokens/ranking selection logic of its own inputs (selection ≠ admission — verify these stay separated in code, not just in the AC's wording); confirm a host with abundant disk/RAM/CPU headroom can genuinely dispatch "100s of parallel tasks" as the operator intends, not silently re-capped by some other untouched limiter

rjwalters:
## Summary

Operator direction on #5270 (with a second clarifying pass) superseded the issue's original "utilization-weighted token multiplier" proposal entirely: **"we should only ever limit parallelism based on the machine disk/RAM/CPU"** — unconditionally, on every auth path (API key or subscription pool, overage-enabled or not).

This PR implements the consolidated (latest-clarification) acceptance criteria from the live issue body:

- **Token axis removed from admission entirely.** `work_finder::resolve_dynamic_max_concurrent` is now `min(disk_headroom, ram_headroom, configured_max)` — the token-axis term is gone, not gated behind an auth-mode check. `.ranking` health remains a spawn-time **selection** input only (`tokens_pool::select` — prefer fresher/healthier accounts, skip revoked/blocked ones); it never gates *how many* sweeps run concurrently again.
- **New `ram_headroom` module** (`loom-daemon/src/ram_headroom.rs`), modeled on `disk_headroom`'s shape: `/proc/meminfo`'s `MemAvailable` on Linux, `vm_stat` free+inactive pages x page size on macOS, `LOOM_PER_WORKTREE_RAM_GB` env override (default 2 GB), `usize::MAX` "unmeasurable - skip the clamp" sentinel mirroring disk headroom's #4164 policy.
- **CPU admission term restored/retuned.** `admission_brake`'s default `load_per_core_threshold` retunes from `4.0` (a rarely-tripped backstop) to `0.95` - per the Curator's verification, `admission_brake.rs` (#4903) already existed as a CPU-saturation gate; this is a retune to the operator's literal ">=95% few-min load average" ask rather than new-mechanism work, promoting it from backstop to primary CPU gate now that the token axis is gone.
- **Observability (AC4):** `calibrate`/`status`/`health` now name which machine axis (max/disk/ram/cpu) is currently binding or holding:
  - `calibrate`: new `BindingTerm::Ram` variant, plus an `admission_axis()` that reports `"cpu"` when the brake is holding (a point-in-time gate outside the numeric `min(...)`, so it overrides the numeric binding term).
  - `health::assess_dispatch`: new `ram_binds_cap` (mirrors `disk_binds_cap`) plus a CPU-brake-held degraded line.
  - `status`/`--json`: `ram_headroom` field added to `DaemonStatusReport` and the dynamic-cap JSON block.
- **Fixed two now-stale reporting paths** that would otherwise have silently reintroduced token-axis gating in the client-facing surface even though the daemon-side cap no longer includes it:
  - `cli::status_render::resolve_capacity`'s fresh-probe tier stopped recomputing its own `token_axis_effective.min(disk).min(configured)` / `token_bound` - it now always defers to the daemon's own authoritative `report.dynamic_cap` / `false`, the same as the ranking/pool tiers already did.
  - The `#4344` "DISPATCH IS TOKEN-STARVED" headline (predicated on `healthy_accounts == 0`) is now categorically impossible to trigger truthfully, since 0 healthy accounts can no longer starve dispatch - hard-coded to `false` with a comment documenting why, rather than deleted, to preserve the historical diagnosis machinery.
- Retuned the token pressure advisory (`capacity::assess_pressure`'s call sites in `work_finder.rs`) to compare against `disk.min(ram)` instead of `disk` alone, so a RAM-starved host can no longer spuriously read as "token-bound" (deferred > 0 for a RAM reason, with the healthy-account count merely happening to be <= disk).
- Documentation: `defaults/docs/daemon-reference.md`'s dynamic-cap / token-backpressure / admission-brake sections updated to describe the new formula and threshold, with the pre-#5270 mechanics kept as clearly-labeled historical context rather than deleted outright.

## What I found in the pre-existing worktree diff and what I did with it

The worktree (`feature/issue-5270`) already had **uncommitted, legitimate progress** from a previously-interrupted Builder run, covering:
- AC1 (token axis removed from `resolve_dynamic_max_concurrent`, `disk_binds_cap`, `binding_term`, `ipc.rs`'s `token_bound=false`) - correct and complete for the two-term formula.
- AC2 (CPU admission retune: `admission_brake`'s default threshold `4.0 -> 0.95`, with matching module-doc rationale and tests) - correct and complete, matching the Curator's "retune, don't rebuild" verification note.

It was **missing**: AC3 (RAM headroom - no `ram_headroom.rs` existed), the CPU axis in `calibrate`'s `BindingTerm`, and the consolidated-AC5 "verify the two changes don't collide" checks (I found and fixed the `status_render.rs` stale-probe-recompute bug and the dead "TOKEN-STARVED" headline while doing that verification). I rebased the branch onto latest `main` (9 commits behind -> fast-forwarded cleanly, no conflicts with the pre-existing diff), kept the AC1/AC2 work as-is, and built AC3/AC4/AC5 on top of it.

## Test plan

- [x] `cargo test -p loom-daemon --lib` - full suite: 3438 passed, 0 failures attributable to this change (2 pre-existing host-load-sensitive flaky tests in `pipeline_snapshot`/`primary_checkout_reaper`, confirmed passing in isolation, unrelated files).
- [x] New `ram_headroom` module: 10 unit tests (parsing, floor-division, env resolution, unmeasurable-probe fallback, live-host smoke test).
- [x] `work_finder::resolve_dynamic_max_concurrent`: extended to 3-arg signature; new disk/ram-binding tests, zero-ram-dispatches-nothing test.
- [x] `admission_brake`: existing suite extended with a default-config-at-0.95 test and an inverted threshold-ordering test (brake now sits below the host breaker, the reverse of pre-#5270).
- [x] `calibrate`: new `Ram`/`admission_axis` tests (binding-term tie-break, CPU-brake-held override, advice text).
- [x] `health`: new `ram_binds_cap`/CPU-brake-held degraded-verdict tests, plus fixed the two disk-focused tests that would otherwise have spuriously tripped on the new RAM check's zero-default.
- [x] `cargo fmt -p loom-daemon -- --check` and `cargo clippy --package loom-daemon --package loom-api -- -D warnings <CI's exact allow-list>` both clean.

Closes #5270

### DIFF
--- a/defaults/docs/daemon-reference.md
+++ b/defaults/docs/daemon-reference.md
@@ -2013,30 +2013,48 @@ Each tick:
    `workfinder-<issue>` idempotency key, making a re-dispatch of an
    already-running issue a no-op.
 
-### Dynamic concurrency scaling (Phase B, #3811; CPU/load term #3978)
+### Dynamic concurrency scaling (Phase B, #3811; CPU/load term #3978; token axis removed / RAM added #5270)
 
 The concurrency cap is **not** a fixed value resolved once at startup. Every
 tick the finder recomputes
 
 ```
-dynamic_cap = min(healthy-token count × per-token concurrency, disk headroom, configured maxConcurrent)
+dynamic_cap = min(disk headroom, ram headroom, configured maxConcurrent)
 ```
 
-from live inputs, so pool/disk/backlog changes are honored without a daemon
-restart. A fourth **cpu/load headroom** term sat in this `min(...)` from #3978
-until **#4512 deleted it** — see [Why there is no CPU term in
-admission](#why-there-is-no-cpu-term-in-admission-4512) below. There is still no
-CPU term in the cap; since #4903 a separate
-[saturation admission brake](#saturation-admission-brake-4903) can *hold* new
+from live inputs, so disk/RAM/backlog changes are honored without a daemon
+restart. **#5270 removed the token axis from this formula entirely**, on any
+auth path (API key or subscription pool, overage-enabled or not) — operator
+direction: "we should only ever limit parallelism based on the machine
+disk/RAM/CPU." A metered `ANTHROPIC_API_KEY` has no 5h/7d subscription window,
+and overage means a subscription pool no longer hard-stops at one either, so
+counting *healthy accounts* was never a proxy for this host's actual capacity
+to run more sweeps. `.ranking` health is **not** deleted — it still drives
+spawn-time **selection** (`tokens_pool::select`: prefer fresher/healthier
+accounts, skip revoked/blocked ones) and the account-health figures remain on
+the status/calibrate surface as informational fields — it just no longer
+gates *how many* sweeps may run concurrently. A fourth **cpu/load headroom**
+term also sat in this `min(...)` from #3978 until **#4512 deleted it** — see
+[Why there is no CPU term in admission](#why-there-is-no-cpu-term-in-admission-4512)
+below. There is still no CPU *term* in the numeric cap; instead, since #4903
+(and retuned by #5270 into the primary "dumb mode" CPU gate — default
+threshold `0.95` load/core, down from the original `4.0` generous backstop)
+the [saturation admission brake](#saturation-admission-brake-4903) *holds* new
 admissions when the host is measurably saturated, without resizing the cap or
 touching running work.
 
 | Input | Source | Bound it enforces |
 |-------|--------|-------------------|
-| **healthy-token count** | `available` accounts in `.ranking` in the pool directory `tokens_pool::paths::resolve_tokens_dir` resolves for the workspace — per-repo `{workspace}/.loom/tokens/` when it holds `*.token` files, else the shared machine-level pool (#3938) (`capacity::read_ranking` / `token_axis_limit`, unified with the writer in #4344 — pre-#4344 this hardcoded the per-repo path even on a shared-pool host, which could pin the dispatch cap at 0 against an orphaned per-repo `.ranking` indefinitely), falling back to the `*.token` count (`tokens::token_pool_size`) when no ranking exists | the count of accounts safe to dispatch to — never dispatch to an exhausted/blocked one (#3902) |
-| **per-token concurrency** | `LOOM_PER_TOKEN_CONCURRENCY` / `autonomous.perTokenConcurrency`, default **2** (#3947) | how many concurrent sweeps to allow **per healthy account**. A plan limit is a utilization-window token bucket, not a session count, so one healthy account can run several concurrent sessions. Before #3947 the implicit factor was `1` (one sweep per account), which collapsed the whole fleet to cap 1 when 6/7 accounts were at their weekly ceiling even though the single healthy account had ample session-window headroom |
 | **disk headroom** | `floor(free_gb / LOOM_PER_WORKTREE_GB)` on the worktree-root volume (`disk_headroom::disk_headroom_limit`, a Rust port of `disk-headroom.sh` that shells to `df -Pk`) | never provision more worktrees than the scratch volume can hold |
-| **configured maxConcurrent** | `LOOM_WORK_FINDER_MAX_CONCURRENT` / `autonomous.workFinder.maxConcurrent` (repurposed from Phase A's fixed target into an operator ceiling) | **the** per-machine admission knob (#4512) — tuned empirically by the operator, the only *policy* term in the `min(...)` (the other two meter exhaustible resources: accounts and bytes) |
+| **ram headroom** (#5270) | `floor(available_gb / LOOM_PER_WORKTREE_RAM_GB)` on the host's currently-available memory (`ram_headroom::ram_headroom_limit`, modeled on `disk_headroom`'s shape: `/proc/meminfo`'s `MemAvailable` on Linux, `vm_stat` free+inactive pages × page size on macOS) | never provision more worktrees than available RAM can hold; the second "dumb mode" machine-headroom axis alongside disk |
+| **configured maxConcurrent** | `LOOM_WORK_FINDER_MAX_CONCURRENT` / `autonomous.workFinder.maxConcurrent` (repurposed from Phase A's fixed target into an operator ceiling) | **the** per-machine admission knob (#4512) — tuned empirically by the operator, the only *policy* term in the `min(...)` (the other two meter exhaustible resources: bytes of disk and bytes of RAM) |
+
+Retired as cap inputs (informational-only now — see `capacity::token_axis_limit` / `tokens_pool::select`):
+
+| Input | Source | What it still does |
+|-------|--------|-------------------|
+| **healthy-token count** (retired from the cap, #5270) | `available` accounts in `.ranking` in the pool directory `tokens_pool::paths::resolve_tokens_dir` resolves for the workspace — per-repo `{workspace}/.loom/tokens/` when it holds `*.token` files, else the shared machine-level pool (#3938) (`capacity::read_ranking` / `token_axis_limit`, unified with the writer in #4344) | drives spawn-time account **selection** (prefer fresher/healthier accounts, skip exhausted/blocked ones, #3902) and is reported on `status`/`calibrate` for observability — no longer bounds `dynamic_cap` |
+| **per-token concurrency** (retired from the cap, #5270) | `LOOM_PER_TOKEN_CONCURRENCY` / `autonomous.perTokenConcurrency`, default **2** (#3947) | still reported for the informational `healthy × per-token` figure on `status`/`calibrate`; no longer multiplies into any cap term |
 
 #### Why there is no CPU term in admission (#4512)
 
@@ -2066,9 +2084,10 @@ actually is**:
 
 1. **Admission is one per-machine knob.** `autonomous.workFinder.maxConcurrent`
    is the whole policy, tuned empirically by the operator (expect 10+ on an
-   8-core API-bound worker). The two remaining terms — the token axis and disk
-   headroom — stay because they meter genuinely *exhaustible* resources
-   (accounts, bytes), not an estimate.
+   8-core API-bound worker). The two remaining terms — disk headroom and (since
+   #5270) RAM headroom — stay because they meter genuinely *exhaustible*
+   resources (bytes), not an estimate. (The token axis sat here too until
+   #5270 removed it — see the section above.)
 2. **The genuinely heavy stages serialize where they occur**, via the
    [machine-wide build slot](#machine-wide-build-slot-4512). N sweeps run
    concurrently; at most `LOOM_BUILD_SLOTS` of them hold the slot while running
@@ -2152,6 +2171,15 @@ consistent env-only behavior. Wiring the Bash `config-resolver.sh` into
 `disk-headroom.sh` too is a separate, larger change; file a follow-up if that
 cost is judged worth paying.
 
+**RAM headroom mirrors this env-only shape (#5270).** `ram_headroom::per_worktree_ram_gb`
+resolves `LOOM_PER_WORKTREE_RAM_GB` (default **2** GB) the same way
+`disk_headroom::per_worktree_gb` resolves its own env var — no config key,
+env-only, floored to a minimum of 1. There is no Bash-side RAM counterpart to
+`disk-headroom.sh` (RAM headroom is Rust-only, consumed solely by the daemon's
+own `resolve_dynamic_max_concurrent`), so the #4032 divergence concern above
+does not apply here — env-only was simply the cheapest consistent choice, not
+a workaround for a second reader.
+
 #### Saturation admission brake (#4903)
 
 Deleting the CPU term left admission with **no term that reads the host at
@@ -2176,8 +2204,18 @@ per-sweep cores, and **never** touches work that is already running.
 |---|---|
 | Config | `autonomous.workFinder.saturationBrake.enabled` / `.loadPerCoreHold` |
 | Env | `LOOM_ADMISSION_BRAKE` / `LOOM_ADMISSION_BRAKE_LOAD_PER_CORE` |
-| Default | **on**, hold at **4.0** load/core |
+| Default | **on**, hold at **0.95** load/core (retuned from `4.0` by #5270) |
 | Precedence | env > config > default, resolved once at daemon startup |
+
+**Retuned by #5270 from a generous backstop to the primary CPU admission
+gate.** Until #5270 the token axis and disk headroom were the two hard
+admission floors, with this brake acting only as a rarely-tripped safety net
+well above them (`4.0` load/core — four runnable threads per logical core).
+#5270 removed the token axis from admission entirely, which promotes this
+brake to **the** CPU term in "dumb mode": its default now matches the
+operator's literal ask — hold new admissions once the host's few-minute load
+average reaches ~95% of its logical core count (`0.95` load/core), resume
+below it.
 
 Properties that are load-bearing (and are pinned by tests):
 
@@ -2188,8 +2226,11 @@ Properties that are load-bearing (and are pinned by tests):
   is precisely how the host recovers.
 - **A healthy host is unchanged.** Below the threshold the admission path is
   byte-for-byte the pre-#4903 one, so an idle 8-core host still reaches its
-  configured cap. `4.0` is deliberately generous: a busy-but-fine host sits under
-  `1.0`, and the incident that motivated the brake was at `11.9`.
+  configured cap. `0.95` holds a notch below full saturation (`1.0` = as many
+  runnable/uninterruptible threads as logical cores), mirroring the build
+  gate's own `0.9` deferral point rather than the pre-#5270 `4.0`, which would
+  have let a host run at 4× overcommit before this brake ever engaged — the
+  incident that originally motivated the brake was at `11.9`.
 - **Fail-safe on missing data.** No load-average source (or a zero CPU count)
   ⇒ no hold. Absent evidence is never treated as load.
 - **Nothing latches.** The hold is re-evaluated every tick and releases the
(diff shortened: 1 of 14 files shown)


rjwalters:
Code evaluation in progress; CI still running (Shell Test Suites, Rust OTLP Feature Build in progress). Releasing the claim and moving to the next queued PR — will re-evaluate once CI settles.

rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:19:00Z) — standing down without reclaiming. Not stomping.

rjwalters:
**Builder note (concurrent worker on #5270).** A second Builder was dispatched onto this
same issue/worktree and reached the same design independently; this PR won the race and
its implementation is correct on AC1–AC4. Not opening a competing PR. Posting the three
places where the two implementations genuinely diverge, because in each one this PR
**deletes** a token signal where the alternative **re-scoped** it — and AC5 explicitly
asked to verify that the removal "does not silently break the accounting … health/
pressure-advisory reporting still rely on (issue #5269, currently in-flight)".

### 1. `ipc.rs:1664` — `token_bound` hard-coded `false` makes the operator guidance dead code

```rust
let token_bound = false;
```

`assess_tokens` still degrades on `healthy_accounts == 0`, so `loom-daemon health` keeps
the signal. But `cli/status_render.rs:1074`'s `} else if rc.token_bound {` branch is now
unreachable, so `loom-daemon status` can never again print the "NO healthy accounts — add
accounts (`loom-daemon tokens bootstrap`) / buy API credits" guidance. A pool with zero
healthy accounts is still a hard failure post-#5270 — `spawn-claude.sh` exits `78`
(`EX_CONFIG`) — it is just a *selection* failure rather than an admission cap.

Suggested instead: `let token_bound = token_axis_limit == 0;` and re-word the branch to
say the cap is unaffected but every spawn will fail account selection. Same field, same
wire compatibility, signal preserved.

### 2. `status_render.rs:994` — the #4344 divergence headline is silenced, which collides with #5269

```rust
let dispatch_starved_but_disagrees = false;
```

The premise for retiring this ("0 healthy accounts can no longer starve dispatch") is true
only for the **cap**. The condition it detected — the daemon's own ranking read shows 0
healthy while a fresh probe disagrees, i.e. the daemon is reading a *diverged/stale*
`.ranking` — still matters, because spawn-time selection reads that same stale file. That
is precisely the failure #5269 / PR #5283 ("surface per-repo token-pool ranking staleness
in health/status") is in flight to make *more* visible. Hard-coding this to `false` moves
the two changes in opposite directions.

Suggested instead: keep the detection, re-word the headline from `DISPATCH IS
TOKEN-STARVED` to `SPAWN SELECTION IS TOKEN-STARVED` and state that the cap is unaffected
(#5270) but every spawn picks from the stale ranking.

### 3. `capacity::assess_pressure` can still fire a false "add token accounts" advisory

The call sites now pass `disk.min(ram)`, but the predicate itself is unchanged:

```rust
let token_bound = deferred > 0 && token_limit <= disk && token_limit <= configured_max;
```

With the token axis gone from the cap, "healthy accounts ≤ the machine terms" no longer
implies tokens are limiting anything. Concrete false positive: 3 healthy accounts, disk
36, ram 32, `maxConcurrent` 12, work deferred **by the ceiling** ⇒ `3 <= 32 && 3 <= 12` ⇒
`token_bound = true` ⇒ an add-capacity advisory telling the operator to buy accounts they
do not need. The pre-#5270 cross-axis comparison existed only to suppress exactly this
kind of false positive while tokens *were* a cap term; once they are not, the comparison
has no remaining meaning.

Suggested instead: `let token_bound = deferred > 0 && token_limit == 0;` — genuine
starvation, the one remaining token-shaped failure, and it needs no cross-axis comparison
to avoid a false positive (it cannot be confused with a disk/RAM/CPU constraint). That
also lets `disk`/`configured_max` drop out of the signature entirely.

### Also worth a look (minor, non-blocking)

- `calibrate` keeps its own `BindingTerm` + a separate `admission_axis()` while
  `status`/`health` grew their own axis strings. Hoisting one shared
  `work_finder::BindingAxis` (`cpu`/`disk`/`ram`/`ceiling`) used by all three makes AC4's
  "the three can never disagree" structural rather than conventional.
- Env var is `LOOM_PER_WORKTREE_RAM_GB`; RAM is machine-level and not per-worktree, so
  `LOOM_PER_SWEEP_RAM_GB` reads truer next to `LOOM_PER_WORKTREE_GB`. Cosmetic, but it is
  a public knob so it is cheaper to rename before release than after.

The alternative implementation carrying all of the above is preserved locally as a patch
and as tag `loom/rescue/issue-5270-alt` on this host — happy to hand over the specific
hunks if a Doctor cycle wants them.

rjwalters:
LGTM — approving.

Verified against issue #5270's consolidated (latest-clarification) acceptance criteria:

- **AC1 (token axis removed unconditionally)**: `resolve_dynamic_max_concurrent` is now `min(disk_headroom, ram_headroom, configured_max)` on every auth path — no API-key/subscription-mode gating. `.ranking` health is kept as a spawn-time *selection* input only (`tokens_pool::select`), never a concurrency cap. Confirmed in `work_finder.rs`, `capacity.rs`, `ipc.rs` (both `build_daemon_status` and `assess_dispatch_headroom` call sites), and `calibrate/mod.rs`.
- **AC2 (CPU admission retune)**: `admission_brake`'s `DEFAULT_ADMISSION_BRAKE_LOAD_PER_CORE` retuned `4.0 → 0.95`, correctly reusing the existing #4903 mechanism per the Curator's verification note rather than building a new gate. New test `default_config_holds_at_95_percent_and_resumes_below_it` pins the boundary; the brake/breaker ordering test was correctly inverted (brake now fires before the breaker).
- **AC3 (RAM headroom)**: new `ram_headroom.rs` module mirrors `disk_headroom.rs`'s shape exactly — `/proc/meminfo` MemAvailable on Linux, `vm_stat` free+inactive×pagesize on macOS, `LOOM_PER_WORKTREE_RAM_GB` override (default 2GB), `usize::MAX` "unmeasurable, skip clamp" sentinel matching the disk-headroom #4164 policy. Good unit-test coverage (parsing, floor division, env resolution, live-host smoke test).
- **AC4 (axis observability)**: `calibrate` gets a `BindingTerm::Ram` variant plus `admission_axis()` that correctly prioritizes a held CPU brake (a point-in-time gate outside the `min(...)`) over the numeric binding term. `health::assess_dispatch` gets `ram_binds_cap` (mirroring `disk_binds_cap`, with correct tie-break semantics keeping the two predicates mutually exclusive) plus a CPU-brake-held degraded line. `status`/`--json` gets a `ram_headroom` field.
- **AC5 (no collision with #5269 / stale reporting)**: nice catch fixing `cli::status_render::resolve_capacity`'s fresh-probe tier, which would have silently recomputed its own token-bound `effective_cap` and reintroduced token-axis gating client-side even after the daemon-side cap dropped it — now correctly defers to `report.dynamic_cap` / hardcoded `token_bound=false`. Also correctly retired the now-categorically-impossible #4344 "DISPATCH IS TOKEN-STARVED" headline (kept as documented dead code rather than deleted, per the PR's own stated rationale) and fixed `assess_pressure`'s call sites to compare against `disk.min(ram)` instead of `disk` alone, so a RAM-starved host can't spuriously read as token-bound.

**Verification performed:**
- `cargo build -p loom-daemon` — clean.
- `cargo test -p loom-daemon --lib` targeted at every touched module (`ram_headroom`, `admission_brake`, `capacity`, `calibrate`, `work_finder`, `health`) — 345/345 passed.
- `cargo clippy --package loom-daemon --package loom-api -- -D warnings` — clean.
- CI: all checks green (`mergeable_state: clean`), including the full `Rust Unit Tests` job.

Code quality is high throughout — module docs clearly explain the "why" for each change, the `#[serde(default)]` on the new `ram_headroom` wire field preserves backward compat with older status snapshots, and the PR's own description accurately reflects what's in the diff (including transparently explaining what pre-existing uncommitted worktree progress it built on top of). No blocking issues found.

PR #3185, opened 2026-04-19
feat: non-destructive settings.json merging during install

rjwalters:
## Summary

Closes #3177

- Deep-merge Loom hooks and permissions into existing `.claude/settings.json` during install/reinstall instead of overwriting
- Preserve all project hooks, permissions, and other keys (enabledPlugins, MCP config, etc.)
- Dedup hooks by command path and permissions by exact string match
- Update uninstall to selectively remove only Loom hooks (`.loom/hooks/` prefix) and permissions via jq

### `loom-daemon/src/init/scaffolding.rs`
- Add `merge_settings_json()` - deep-merges hooks (per type/matcher) and permissions (union)
- Add `merge_hooks()`, `merge_hook_commands()`, `merge_permissions()` helpers
- Add `remove_loom_hooks()` and `remove_loom_permissions()` for uninstall
- Intercept settings.json in `.claude/` directory handling: save existing, let force-merge overwrite, then merge back
- 10 new unit tests covering merge, dedup, empty cleanup, and integration

### `scripts/uninstall-loom.sh`
- Skip `.claude/settings.json` from direct file removal
- Add smart removal: use jq to remove only Loom hooks and permissions
- Remove file if empty after cleanup, otherwise preserve project settings
- Graceful fallback if jq unavailable

## Test plan

- [x] 10 unit tests in scaffolding.rs all pass
- [ ] Fresh install (no existing settings.json) works as before
- [ ] Install into project with existing hooks preserves all project hooks
- [ ] Reinstall doesn't duplicate Loom hooks
- [ ] Permissions are unioned, not replaced
- [ ] Uninstall removes only Loom hooks matching `.loom/hooks/` prefix
- [ ] Malformed JSON in existing settings.json doesn't crash

PR #5306, opened 2026-08-04
fix(install): regenerate manifest after guard-hook wiring to stop spurious settings.json drift

### LINKED ISSUE #5279: install.sh rewrites .claude/settings.json wholesale, silently dropping other tools' hooks (install order becomes load-bearing)
## Context

Hit while updating Loom 0.11.0 -> 0.18.0 in a repo that also has [Repo Skills](https://github.com/rjwalters/repo) installed.

Loom's `install.sh` **rewrites `.claude/settings.json` wholesale** rather than merging into it. Any hook another tool registered in that file is silently dropped. In our case Repo Skills' `SessionStart` hook disappeared, and the only fix was to notice and re-run the other tool's installer afterward:

```
Loom install.sh   -> writes settings.json, Repo Skills SessionStart hook gone
Repo Skills install.sh (re-run) -> hook restored
```

That makes install **order-dependent** in a way nothing announces: install Loom last and you quietly lose the other tool's hooks. There is no error, no warning, and no diff shown — the hook is simply absent on the next session, and the tool it belonged to stops working.

Because settings.json is then legitimately different from what Loom wrote, `verify-install.sh` reports it as drift **by design**, which makes the real signal harder to see: an operator learns to expect a drift report there and stops reading it closely.

We now carry a note about install ordering as institutional knowledge, which is the sort of thing that should not need to be institutional knowledge.

- [ ] `install.sh` merges its hooks into an existing `.claude/settings.json` instead of replacing the file
- [ ] Failing that: detect foreign hooks that are about to be dropped and warn, naming them, before writing
- [ ] Installing Loom into a repo with another tool's hooks present leaves those hooks intact and functional
- [ ] Consider whether `verify-install.sh` can distinguish "settings.json legitimately contains other tools' entries" from real drift, so the drift signal stays meaningful

---

### Problem Statement (revised — the "wholesale rewrite" premise is only partly current)

Verified against `origin/main` (`5fb6e9cb`, 2026-08-04): Loom already ships non-destructive `.claude/settings.json` merging. `merge_settings_json` (`loom-daemon/src/init/scaffolding.rs:453`, added by #3185) deep-merges hooks (by hook type + matcher, deduped by normalized command) and unions `permissions.allow`, preserving every other top-level key from the existing file untouched. `setup_repository_scaffolding` (same file, ~line 1084-1119) reads the existing `settings.json` **before** the `.claude/` directory copy overwrites it, then re-merges and rewrites it afterward — this runs on both the Full Install and `--quick` paths (both call `loom-daemon init`, unified by #3502).

This is not new: commit `f05c10da` ("non-destructive settings.json merging during install", #3185) is an ancestor of the `v0.11.0` tag — i.e. it predates the exact upgrade (`0.11.0 -> 0.18.0`) described in this issue's Context. The existing test `test_merge_settings_in_scaffolding_reinstall` (`loom-daemon/src/init/scaffolding.rs:3173`) exercises precisely this issue's scenario — a foreign hook under a hook type Loom doesn't define (`UserPromptSubmit` in the test, `SessionStart` in this report), a foreign permission string, and an arbitrary foreign top-level key (`enabledPlugins`) — across a reinstall, and asserts all three survive. It passes on current `main` (re-ran it during this curation pass: `cargo test -p loom-daemon --lib init::scaffolding::tests::test_merge_settings_in_scaffolding_reinstall` → `1 passed; 0 failed`).

**So AC #1–#3 as originally written describe a defect that the current merge design should already prevent**, at the function level, on a supported install path. What is verified NOT to hold:

The reported incident is real (it happened to the reporter), but its cause is not yet identified given the merge logic above already existed at the reporter's starting version. Plausible mechanisms, none yet confirmed:

- The operator's actual `~/.local/share/loom` machine-level checkout (or a cached pre-built `loom-daemon` binary invoked by `install.sh`/`install-loom.sh`) predated `f05c10da` even though the *target repo* was reported as being on `0.11.0` — the binary doing the writing, not the target repo's installed version, is what matters here.
- `read_existing_settings` (`scaffolding.rs:434`) returns `None` on any JSON parse failure (comments, trailing commas, or a non-strict-JSON file some other tool wrote) with no warning — a silent parse failure here would skip the entire merge and leave the freshly-overwritten Loom-only file in place, which would look exactly like this report from the operator's side.
- Some other install path (e.g. an even older pre-`#3502` `--quick` invocation, if the operator's toolchain was older than the version string suggests) bypassed `loom-daemon init` entirely.

This needs verification by actually reproducing end-to-end (see Test Plan) before assuming any one of these is the mechanism — an explicit, measured refutation of all three (i.e. confirming the bug does not reproduce via any current install path) is a complete, successful outcome for the Builder to close this portion of the issue with, not a failure to deliver a fix.

### What is confirmed still missing (independent of the above)

`defaults/scripts/verify-install.sh`'s drift check (`compute_entry_digest` / the manifest-hash loop around line ~705-725) whole-file-hashes `.claude/settings.json` against the shipped default with no region marker (unlike the `loom-block`-region carve-out it applies to other files). Any legitimate merge-preserved foreign content — exactly what `merge_settings_json` is designed to add — will register as drift here, matching this issue's Context claim precisely. This is a real, currently-unaddressed gap regardless of how the Suspected Cause investigation above resolves. Note Loom's Rust-internal verification already solved the equivalent problem for `loom-daemon init`'s own report (`filter_preserved_from_verification_failures`, `loom-daemon/src/init/mod.rs`, fixing issue #3218) — `verify-install.sh` (the standalone bash drift-check script) has no equivalent and should adopt the same "preserved is not a failure" principle.

### Acceptance Criteria

- [ ] Reproduce (or explicitly refute, per "Suspected Cause" above) the reported hook-loss on current `main` via both `install.sh --quick <target>` and a Full Install (`scripts/install-loom.sh`), using a target repo whose `.claude/settings.json` carries a foreign `SessionStart` hook (mirroring Repo Skills) before install
- [ ] If reproduced: fix the actual gap found (e.g. warn-then-skip-merge on `read_existing_settings` parse failure instead of silently skipping; or another root cause identified during reproduction) and extend `test_merge_settings_in_scaffolding_reinstall` (or add a sibling test) to cover it
- [ ] If not reproduced: document in this issue why the incident occurred despite `merge_settings_json` predating the reporter's starting version (stale binary being the leading hypothesis) — this is a complete outcome, not a partial one
- [ ] `verify-install.sh` no longer reports drift on `.claude/settings.json` purely because it contains foreign (non-Loom) hooks/permissions/top-level keys that `merge_settings_json` legitimately preserved, while still catching genuine drift in Loom-owned content (e.g. a Loom hook entry edited or removed)
- [ ] Existing behavior is unchanged for a `.claude/settings.json` that contains ONLY Loom-shipped content (byte-identical to source → still reports `ok`, not drift)

### Affected Files

- `loom-daemon/src/init/scaffolding.rs` - `merge_settings_json` (line 453), `merge_hooks`/`merge_hook_commands` (line ~478-561), `read_existing_settings` (line 434); audit/fix target if the repro in AC #1 finds a real gap
- `loom-daemon/src/init/mod.rs` - `filter_preserved_from_verification_failures`; the existing pattern for "preserved content is not a verification failure" that `verify-install.sh`'s bash-side check should mirror
- `defaults/scripts/verify-install.sh` - the manifest-hash drift loop (~line 690-730) that whole-file-hashes `.claude/settings.json` with no allowance for legitimately-merged foreign content
- `install.sh`, `scripts/install-loom.sh` - install entry points to drive for the end-to-end repro in AC #1

### Test Plan

- [ ] Manual verification: scratch target repo with a `.claude/settings.json` containing a foreign `SessionStart` hook + a foreign permission string + an arbitrary foreign top-level key; run `install.sh --quick` and separately a Full Install over it; inspect the resulting `.claude/settings.json` for survival of all three, then run `verify-install.sh` and confirm it does NOT report drift solely from that foreign content
- [ ] Automated tests: `cargo test -p loom-daemon --lib init::scaffolding::tests::test_merge_settings_in_scaffolding_reinstall` (pre-existing, passes on `main` as of `5fb6e9cb`/2026-08-04 — re-verify before relying on this); add/extend a test for whichever mechanism the AC #1 repro identifies (or confirms absent); add a `tests/` or `defaults/scripts/tests/` case for `verify-install.sh`'s new foreign-content tolerance
- [ ] Edge cases: a `.claude/settings.json` with a foreign hook under a hook type Loom ALSO defines (e.g. both register `PreToolUse`) to confirm dedup-by-normalized-command doesn't collide; a non-strict-JSON existing `settings.json` (comments/trailing commas) to confirm `read_existing_settings`'s current silent-`None`-on-parse-failure behavior either already warns or gets one added

rjwalters:
## Summary

Issue #5279's Curator Enhancement framed this as a verify-and-close task: confirm
whether the "wholesale rewrite drops other tools' hooks" premise still reproduces
on current `main` (it likely doesn't, per `merge_settings_json` added in #3185),
and either document a refutation or fix whatever real gap the repro surfaces. I
ran the repro end-to-end and found **both** outcomes apply to different parts of
the issue:

- **Refuted**: the original reported symptom (a foreign hook/permission/top-level
  key silently dropped by install) does **not** reproduce on current `main`. A
  manual repro (scratch target with a foreign `SessionStart` hook + foreign
  permission + foreign top-level key, installed over via `install.sh --quick`)
  preserved all three, confirming `merge_settings_json` works as designed.
- **Found a real, concrete, always-reproducible gap**: `verify-install.sh verify`
  reports spurious `DRIFT DETECTED` on `.claude/settings.json` after **every**
  Quick Install — even a completely vanilla install with **zero** customization
  and zero foreign content. This directly violates the issue's AC #5
  ("byte-identical Loom content → still `ok`, not drift") and is a more precise,
  always-reproducible version of AC #4's "verify-install.sh gives false positives
  on settings.json" concern. Root cause: `loom-daemon init` generates
  `.loom/manifest.json` as its own last internal step — **before**
  `install.sh`'s `wire_quick_install_guard_hooks` (via `ensure_project_hook_wiring`)
  asserts project-level guard-hook entries into `.claude/settings.json`. The
  manifest's stored hash reflects pre-wiring content, so the very next `verify`
  call sees a file that legitimately changed after the snapshot and reports it as
  drift.

I also hardened `read_existing_settings` per the issue's "Suspected Cause" #2
(a present-but-unparseable existing `settings.json` previously skipped the merge
with **zero** warning) and added the two edge-case tests the issue's Test Plan
named explicitly.

## Investigation details (repro steps + findings)

1. Re-ran `cargo test -p loom-daemon --lib init::scaffolding::tests::test_merge_settings_in_scaffolding_reinstall` → **1 passed** (confirmed still passing, as the Curator noted).
2. Manual repro: scratch git repo with `.claude/settings.json` carrying a foreign `SessionStart` hook, a foreign permission (`Bash(repo-skills-tool:*)`), and a foreign top-level key (`enabledPlugins`). Ran `install.sh --quick -y <target>`:
   - All three foreign items **survived** the merge intact (see the settings.json diff in the PR discussion if useful — happy to attach on request).
   - `verify-install.sh verify` reported `DRIFT DETECTED (1 modified)` on `.claude/settings.json` — **but** a control repro with a fully vanilla target (no foreign content at all) showed the **exact same** spurious drift, isolating the cause to something unrelated to foreign content.
3. Traced the mechanism: `install.sh`'s Quick Install flow is `loom-daemon init` (merges settings.json + generates the manifest internally) → `install_hooks_and_cli` → `wire_quick_install_guard_hooks` (→ `ensure_project_hook_wiring`, which mutates `.claude/settings.json` to add project-level `${CLAUDE_PROJECT_DIR}/.loom/hooks/...` entries). The manifest snapshot happens **before** that last mutation.
   - Full Install (`scripts/install-loom.sh`) does **not** have this bug: its only project-settings mutator, `provision_loom_hooks`, runs **before** `loom-daemon init` (and it doesn't call `ensure_project_hook_wiring` at all — Full Install relies solely on user-scope hook wiring). Confirmed via code reading since a live Full Install requires real forge auth/remote infrastructure I did not want to fabricate; both install paths invoke the identical `loom-daemon init --defaults <defaults> <target>` binary command for the settings.json merge itself (already verified via the Quick Install repro + the existing Rust test).
4. Fix: added `regenerate_manifest_after_hook_wiring()` to `install.sh`, called right after each of the two `wire_quick_install_guard_hooks` call sites (the `--confirm-reinstall` branch and the fresh-quick branch). Verified with a real end-to-end run: a vanilla `install.sh --quick` target now reports `ALL FILES MATCH` instead of `DRIFT DETECTED`, and the foreign-content target now reports `ALL FILES MATCH` too (foreign content preserved **and** no false-positive drift).

## Changes

- `install.sh`: new `regenerate_manifest_after_hook_wiring()` helper, called after both `wire_quick_install_guard_hooks "$TARGET_PATH"` call sites, so `.loom/manifest.json` reflects the genuinely final post-install state instead of a pre-hook-wiring snapshot.
- `loom-daemon/src/init/scaffolding.rs`:
  - `read_existing_settings` now emits a `Warning:` (instead of silently returning `None`) when an existing `.claude/settings.json` is present but fails to parse as JSON — the merge is still skipped (documented, unchanged behavior), but the drop is now visible.
  - New test `test_merge_settings_dedup_preserves_foreign_command_in_shared_matcher`: a foreign hook command under a hook type+matcher (`PreToolUse`/`Bash`) Loom's own defaults *also* register — confirms dedup is by normalized command, not by matcher, so neither command is lost.
  - New test `test_merge_settings_skips_merge_on_invalid_json_existing_settings`: a non-strict-JSON existing `settings.json` (trailing comma) doesn't panic and the documented skip-merge behavior is locked in.
- `tests/install/test-manifest-freshness-after-hook-wiring.sh` (new): hermetic suite (no loom-daemon binary/build required, following the existing `test-hooks-preserve.sh` awk-extraction pattern) that reproduces the stale-manifest bug and confirms the fix resolves it, plus a best-effort no-op check when `verify-install.sh` is absent. Wired into `defaults/scripts/tests/ci-wired.txt`.

| Reproduce or explicitly refute the reported hook-loss via `install.sh --quick` and Full Install | ✅ | Refuted for Quick Install via live repro (all 3 foreign items survive); Full Install shares the identical `loom-daemon init` binary call and its only settings-mutator runs *before* init, so it cannot exhibit the bug either — documented above rather than re-run live (requires real forge auth/remote). |
| If reproduced: fix the actual gap + extend tests | ✅ (different, real gap found) | Fixed the manifest-freshness ordering bug in `install.sh`; extended `scaffolding.rs` tests for the two named edge cases. |
| If not reproduced: document why | ✅ | Documented above; stale-binary remains the leading hypothesis for the *original* reporter's specific incident, though the manifest-freshness bug found here independently explains why `verify-install.sh` "reports it as drift by design" is true today, universally, not just for foreign content. |
| `verify-install.sh` no longer reports drift purely from legitimately-preserved foreign content | ✅ | Live repro: foreign-content target reports `ALL FILES MATCH` post-fix (was `DRIFT DETECTED`). |
| Existing behavior unchanged for settings.json with ONLY Loom-shipped content (still `ok`) | ✅ | Live repro: vanilla target reports `ALL FILES MATCH` post-fix (was, in fact, incorrectly `DRIFT DETECTED` pre-fix — this AC was violated on `main` and is now satisfied). |

## Test Plan

- `cargo test -p loom-daemon --lib init::scaffolding::` → 64 passed, 0 failed (pre-existing + 2 new tests).
- `cargo clippy -p loom-daemon --lib -- -D warnings` → clean. `cargo fmt --check -p loom-daemon` → clean.
- `bash tests/install/test-manifest-freshness-after-hook-wiring.sh` → 3 passed, 0 failed.
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` → OK (new suite wired correctly).
- `bash scripts/test-installer.sh` (full suite, 178 assertions across ~87 numbered tests) → 178 passed, 1 failed. The 1 failure (`test-loom-dispatcher.sh`, 3 sub-assertions re: `loom restart --machine` env-var harvesting, issue #4581) is **pre-existing and unrelated** — confirmed by running the same sub-suite from the unmodified primary checkout, which reproduces the identical 3 failures with none of this PR's changes present.
- `bash defaults/scripts/tests/test-install-reinstall-safety.sh` → 21/21 passed.
- `bash scripts/test-install-local-mode.sh` → 19/19 passed.
- Manual end-to-end repro (see Investigation details above): `install.sh --quick` against (a) a vanilla target and (b) a target with foreign `SessionStart` hook + foreign permission + foreign top-level key, both followed by `.loom/scripts/verify-install.sh verify` — both report `ALL FILES MATCH` post-fix.

Closes #5279

### DIFF
--- a/defaults/scripts/tests/ci-wired.txt
+++ b/defaults/scripts/tests/ci-wired.txt
@@ -141,6 +141,7 @@ tests/install/test-daemon-build-shortcut.sh
 tests/install/test-forge-detect.sh
 tests/install/test-git-repo-detection.sh
 tests/install/test-hooks-preserve.sh
+tests/install/test-manifest-freshness-after-hook-wiring.sh
 tests/install/test-provision-daemon.sh
 tests/install/test-uninstall-labels-block.sh
 tests/install/test-uninstall-preserve-siblings.sh
--- a/install.sh
+++ b/install.sh
@@ -217,6 +217,41 @@ wire_quick_install_guard_hooks() {
     warning "Project-level guard-hook entries not fully asserted in $target/.claude/settings.json; run scripts/install-loom.sh (Full Install) or re-add them manually — see defaults/docs/guard-hooks.md."
 }
 
+# Regenerate .loom/manifest.json's checksums after this function's caller has
+# finished ALL post-`loom-daemon init` mutations of Loom-tracked files
+# (issue #5279).
+#
+# `loom-daemon init` generates the manifest as its own last internal step
+# (loom-daemon/src/init/post_init.rs::generate_manifest, invoked from
+# loom-daemon/src/init/mod.rs) -- which runs BEFORE `wire_quick_install_guard_hooks`
+# (above) asserts project-level guard-hook entries into `.claude/settings.json`
+# via `ensure_project_hook_wiring`. That means the manifest's stored hash for
+# `.claude/settings.json` reflects its PRE-wiring content, not the genuinely
+# final installed state.
+#
+# Left unfixed, `verify-install.sh verify` reports spurious `DRIFT DETECTED` on
+# `.claude/settings.json` immediately after EVERY quick install -- even a
+# completely vanilla install with zero customization and zero foreign
+# content -- because the on-disk file legitimately changes after the manifest
+# snapshot was taken. This is the concrete, reproducible mechanism behind the
+# "verify-install.sh reports [settings.json] as drift by design" symptom
+# described in #5279 (a stricter, always-reproducible version of it: no
+# foreign/sibling-tool content is even required to trigger it).
+#
+# Call this ONLY after every mutator of Loom-tracked files in the current
+# install path has run (currently: right after `wire_quick_install_guard_hooks`,
+# its last one on the quick-install path). Best-effort: a missing or failing
+# verify-install.sh never aborts the install.
+regenerate_manifest_after_hook_wiring() {
+  local target="$1"
+  local script="$target/.loom/scripts/verify-install.sh"
+
+  if [[ -f "$script" ]]; then
+    ( cd "$target" && bash "$script" generate --quiet ) || \
+      warning "Could not regenerate .loom/manifest.json after guard-hook wiring; 'verify-install.sh verify' may report spurious drift on .claude/settings.json."
+  fi
+}
+
 # Export LOOM_VERSION and LOOM_COMMIT so `loom-daemon init`'s template
 # substitution can fill {{LOOM_VERSION}} / {{LOOM_COMMIT}} placeholders in
 # the CLAUDE.md templates instead of falling back to the literal string
@@ -1180,6 +1215,11 @@ elif [[ -d "$TARGET_PATH/.loom" ]]; then
     # restores a working guard-hook execution path instead of leaving zero.
     wire_quick_install_guard_hooks "$TARGET_PATH"
 
+    # Regenerate the checksum manifest now that ALL settings.json mutations for
+    # this install path are complete (issue #5279 — see the function's own
+    # comment for why this must run after wire_quick_install_guard_hooks).
+    regenerate_manifest_after_hook_wiring "$TARGET_PATH"
+
     # Issue #3545: reconcile the git index after the uninstall→reinstall cycle.
     # The chained uninstall staged the deletion of every prior Loom file (now
     # scoped to Loom-managed paths — see scripts/uninstall-loom.sh), then
@@ -1584,6 +1624,11 @@ case "$METHOD" in
     # copies that nothing references (the 0.16.0 defaults settings.json has no
     # `hooks` block) and no user-scope wiring at all: zero guard coverage.
     wire_quick_install_guard_hooks "$TARGET_PATH"
+
+    # Regenerate the checksum manifest now that ALL settings.json mutations for
+    # this install path are complete (issue #5279 — see the function's own
+    # comment for why this must run after wire_quick_install_guard_hooks).
+    regenerate_manifest_after_hook_wiring "$TARGET_PATH"
 
     echo ""
     success "Quick installation complete!"
--- a/loom-daemon/src/init/scaffolding.rs
+++ b/loom-daemon/src/init/scaffolding.rs
@@ -432,13 +432,37 @@ pub fn load_internal_skip_list(defaults_path: &Path) -> HashSet<String> {
 }
 
 /// Read and parse an existing settings.json file, returning None if missing or invalid.
+///
+/// A missing file is the normal, silent case (fresh install — nothing to
+/// merge). A *present but unparseable* file is a different, noteworthy case:
+/// the caller ([`merge_settings_json`] via `setup_repository_scaffolding`)
+/// treats `None` as "nothing to merge" and leaves Loom's freshly-copied
+/// defaults in place untouched — silently dropping every pre-existing hook,
+/// permission, and top-level key the file contained (issue #5279's
+/// "Suspected Cause" #2: a non-strict-JSON file some other tool wrote — e.g.
+/// containing comments or a trailing comma — would previously fail this
+/// parse with zero warning, discarding its content on the very next
+/// reinstall/upgrade with no diagnostic pointing at why). We now warn on that
+/// specific case so the drop is visible instead of silent; the missing-file
+/// case stays silent since it is not an error.
 fn read_existing_settings(path: &Path) -> Option<Value> {
     let content = fs::read_to_string(path).ok()?;
-    let value: Value = serde_json::from_str(&content).ok()?;
-    if value.is_object() {
-        Some(value)
-    } else {
-        None
+    match serde_json::from_str::<Value>(&content) {
+        Ok(value) if value.is_object() => Some(value),
+        Ok(_) => {
+            eprintln!(
+                "Warning: {} is valid JSON but not a JSON object; skipping settings.json merge (Loom's own settings.json will be used as-is).",
+                path.display()
+            );
+            None
+        }
+        Err(e) => {
+            eprintln!(
+                "Warning: Failed to parse existing {} as JSON ({e}); skipping settings.json merge. Loom's own settings.json will be written as-is, which may drop pre-existing hooks/permissions/keys from this file. Fix the JSON syntax (comments and trailing commas are not valid JSON) and re-run install to merge them back in.",
+                path.display()
+            );
+            None
+        }
     }
 }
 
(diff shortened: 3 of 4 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-04T16:36:17Z) — standing down without reclaiming. Not stomping.

rjwalters:

rjwalters:
✅ **Approved!**

Solid investigation-driven PR. Verified against a live worktree checkout:

- **`install.sh`**: `regenerate_manifest_after_hook_wiring()` correctly regenerates `.loom/manifest.json` right after both call sites of `wire_quick_install_guard_hooks` (the `--confirm-reinstall` branch at line ~1216 and the fresh-quick branch at line ~1626) — confirmed both call sites are covered via `grep`, best-effort (never aborts install on failure).
- **`loom-daemon/src/init/scaffolding.rs`**: `read_existing_settings` now surfaces a `Warning:` on unparseable-but-present `settings.json` instead of silently dropping foreign content — sound fix for the issue's "Suspected Cause" #2.
- **Root-cause explanation is well-supported**: manifest snapshot happens before `wire_quick_install_guard_hooks` mutates `.claude/settings.json`, so the very next `verify-install.sh verify` sees legitimate post-snapshot drift and misreports it.

## Test Execution

- `cargo test -p loom-daemon --lib init::scaffolding::` — ✅ Executed: **64 passed, 0 failed** (matches PR's claimed count exactly, including the 2 new tests).
- `bash tests/install/test-manifest-freshness-after-hook-wiring.sh` — ✅ Executed: **3 passed, 0 failed** (matches PR's claim).
- `defaults/scripts/tests/ci-wired.txt` — ✅ Verified: new hermetic test is correctly wired in.
- `bash scripts/test-installer.sh` (178-assertion full suite) — ⏭️ Skipped: long-running (>2 min); CI's "Installer Integration Tests" check already covers this and passed.
- Manual end-to-end vanilla/foreign-content repro — ⚠️ Skipped: requires manual observation; PR description documents this was run and gives before/after evidence.

## Issue linking

PR body includes "Closes #5279" — correct.

## CI

All required checks pass (`gh pr checks 5306`); merge state is `CLEAN`. The two "skipping" checks (Dashboard Tests, MCP Server Build) are unrelated to this change (installer/Rust-only diff).

Nice work isolating the real, always-reproducible bug distinct from the issue's original (refuted) hypothesis, and keeping the two clearly separated in the PR description.

rjwalters:

merging it automatically:

- **Blast radius**: touches `install.sh`, the Quick Install entry point — one of the explicitly-named installer/updater surfaces this criterion treats as high blast-radius regardless of diff size. The new `regenerate_manifest_after_hook_wiring()` helper is small, well-scoped, and best-effort (never aborts install on a failure — it only warns), and is inserted at two existing call sites right after `wire_quick_install_guard_hooks` without touching surrounding logic. But a bug there (wrong script path, a quoting/subshell issue) would affect `.loom/manifest.json` generation on every Quick Install going forward, fleet-wide, not just one feature — so per the merge-risk criteria's explicit installer/updater carve-out this stays a human-merge candidate even though diff composition, Judge review depth, and revertability are all green.

- A human merges it directly with `./.loom/scripts/merge-pr.sh 5306`

---
*Automated by Champion role*

PR #5308, opened 2026-08-04
fix(tests): make test-loom-dispatcher.sh Test 16b OS-independent, add real-plist happy path

### LINKED ISSUE #5169: test-loom-dispatcher.sh: false failures from unisolated ambient env (LOOM_RUNTIME leak, Darwin uname branch mismatch)
## Bug Description

`defaults/scripts/tests/test-loom-dispatcher.sh` has two distinct test-hermeticity gaps that cause deterministic (not flaky) local failures on main, even though the underlying product behavior is correct:

**1. Test 21 leaks ambient `LOOM_RUNTIME` from the calling shell**

Test 21 ("`loom sweep` honors `.loom/config.json` `runtimes.default` (no env)") invokes the dispatcher without clearing `LOOM_RUNTIME`:

```bash
out=$(cd "$SWEEPCFG" && LOOM_CONFIG_DEFAULTS_FILE="" LOOM_HOME="$CHK" bash "$DISPATCHER" sweep 4467 2>&1)
```
(`defaults/scripts/tests/test-loom-dispatcher.sh:507`)

Per the documented precedence (`LOOM_RUNTIME` env > config > default), any ambient `LOOM_RUNTIME` set in the *outer* shell wins and is indistinguishable from the "env" case the test is trying to avoid. `spawn-claude.sh` exports `LOOM_RUNTIME=claude` into every Loom-spawned worker session — so this test fails every time it's run from inside a Loom agent shell (Builder, Judge, Auditor, etc.), which is precisely the shell context these suites are normally exercised in locally. It only passes on a fresh CI runner because GitHub Actions starts with no `LOOM_RUNTIME` set.

Fix: explicitly unset/clear `LOOM_RUNTIME` on the Test 21 invocation, e.g. `LOOM_RUNTIME= LOOM_CONFIG_DEFAULTS_FILE="" ... bash "$DISPATCHER" sweep 4467`.

**2. Test 16b's systemd fixture is unreachable on Darwin hosts**

Test 16b ("`loom restart --machine` harvests + re-exports the live systemd unit's LOOM_* env before the bare-exec fallback (#4581)") writes a fixture systemd unit file:

```bash
mkdir -p "$HOME_HARVEST/.config/systemd/user"
cat > "$HOME_HARVEST/.config/systemd/user/loom-daemon.service" <<'EOF2'
...
EOF2
```
(`defaults/scripts/tests/test-loom-dispatcher.sh:387-394`)

But `scripts/loom`'s harvest dispatch branches unconditionally on the *real* host OS, not on which fixture is present:

```bash
if [[ "$(uname -s 2>/dev/null)" == "Darwin" ]]; then
    ...
    plist="$HOME/Library/LaunchAgents/${label}.plist"
```
(`scripts/loom:310-313`)

On a real macOS host, the Darwin branch always runs and looks for a launchd plist that Test 16b never creates, so the harvest finds nothing and the assertions on `LOOM_WORK_FINDER=[1]` / `LOOM_MAIN_HEALTH_GATE=[1]` / `preserved 2 LOOM_*/token env var(s)` fail. This test can only pass on Linux (where CI runs it — `shell-suite-tests` is `runs-on: ubuntu-latest`), never locally on a macOS dev/fleet host.

A secondary gap this reveals: there is no equivalent "harvest succeeds from a real launchd plist" test. Test 16d only covers the *unreadable* plist abort path (`#4011` class). So the happy-path launchd harvest is untested both in CI (never reaches the Darwin branch) and locally on macOS (wrong fixture for the branch taken).

## Reproduction Steps

1. Checkout main: `git checkout main && git pull`
2. From inside any Loom-spawned agent shell (i.e. one where `spawn-claude.sh` has exported `LOOM_RUNTIME=claude`) on a macOS host, run: `bash defaults/scripts/tests/test-loom-dispatcher.sh`
3. Observe Test 21 and Test 16b failures (see Output below)

## Expected Behavior

The suite should pass identically regardless of the calling shell's ambient `LOOM_RUNTIME`, and regardless of the host OS's real supervisor mechanism vs. the fixture the test constructs.

## Actual Behavior

5 assertions fail (94/99 passed):
- Test 21: `runtimes.default=codex (no env) resolves to spawn-codex.sh` — resolves to `spawn-claude.sh` instead, because ambient `LOOM_RUNTIME=claude` leaks in
- Test 21: `config-selected codex path does not touch the claude runner` — same root cause
- Test 16b: `harvested LOOM_WORK_FINDER reaches start_target's re-render (#4581)` — empty because the systemd fixture is unreachable on Darwin
- Test 16b: `harvested LOOM_MAIN_HEALTH_GATE reaches start_target's re-render (#4581)` — same root cause
- Test 16b: `dispatcher reports the harvested count` — same root cause

## Output

```
[FAIL]: runtimes.default=codex (no env) resolves to spawn-codex.sh (missing substring: 'STUB_CODEX' in: ...
  spawn-worker: runtime=claude (from env (LOOM_RUNTIME)) -> spawn-claude.sh
  # LOOM_RUNTIME_RESOLVED runtime=claude
  STUB_CLAUDE argv=[-p /loom:sweep 4467 --dangerously-skip-permissions])
[FAIL]: config-selected codex path does not touch the claude runner (unexpected substring: 'STUB_CLAUDE')

[FAIL]: harvested LOOM_WORK_FINDER reaches start_target's re-render (#4581) (missing substring: 'LOOM_WORK_FINDER=[1]' in: ...
  loom: supervised 'loom-daemon restart' unavailable or refused (not launchd-managed, or not currently running) — falling back to stop-then-start.
  STUB_STOP args=[--restarting]
  STUB_START args=[] LOOM_WORK_FINDER=[] LOOM_MAIN_HEALTH_GATE=[])
[FAIL]: harvested LOOM_MAIN_HEALTH_GATE reaches start_target's re-render (#4581) (missing substring: 'LOOM_MAIN_HEALTH_GATE=[1]' in: ...)
[FAIL]: dispatcher reports the harvested count (missing substring: 'preserved 2 LOOM_*/token env var(s)' in: ...)
```

## Environment

- OS: Darwin 25.6.0 (macOS, Loom fleet host)
- Commit: d6c5a9bc (origin/main HEAD at audit time)
- Build: N/A (shell test, no compile step)
- Shell env: `LOOM_RUNTIME=claude`, `LOOM_DAEMON_SUPERVISOR=launchd` (both exported by this session's own Loom worker spawn — i.e. any Auditor/Builder/Judge session on this host reproduces this)

## Impact

Not a CI blocker (CI runs on ubuntu-latest, where both failures are masked — LOOM_RUNTIME is unset by default and the systemd branch matches the fixture). But every local run of `defaults/scripts/tests/run-ci-suites.sh` or `test-loom-dispatcher.sh` from a Loom-spawned agent shell on a macOS fleet host (i.e. routine Auditor/Builder/Judge local validation) reports these 5 as failing, which is noise that erodes trust in local test runs and could mask a real regression introduced alongside them in the future. It also means the launchd happy-path harvest behavior (the actual supervisor this fleet uses) has no passing automated coverage anywhere.

---
Discovered during main branch audit (commit d6c5a9bc).

rjwalters:
Closes #5169

## Summary

Issue #5169 flagged two hermeticity bugs in `defaults/scripts/tests/test-loom-dispatcher.sh`:

1. **Test 21 leaked ambient `LOOM_RUNTIME`.** Already fixed on `main` by #5168 (`env -u LOOM_RUNTIME -u LOOM_WORKSPACE` on the Test 21 / Test 19 "no env" invocations). Verified there is no other "no env" invocation in this file with the same gap — grepped all `LOOM_CONFIG_DEFAULTS_FILE=""` call sites and confirmed each one that claims "no env" already scrubs `LOOM_RUNTIME`/`LOOM_WORKSPACE`.

2. **Test 16b's systemd fixture was unreachable on a real Darwin host.** `scripts/loom`'s harvest dispatch branches on the *real* `uname -s`, not on which fixture the test wrote. Test 16b wrote a systemd-unit fixture but never forced `uname`, so on macOS the Darwin branch always ran and the fixture was never read, failing 3 assertions. Fixed by forcing `uname` -> `"Linux"` via a PATH-prepended fake `uname` script — mirroring Test 16d's existing (and already-correct) Darwin override in the other direction.

## Secondary addition

Added **Test 16f**: a companion happy-path launchd-harvest test (a real, parseable plist -> harvested env reaching `start_target`'s re-render). Previously only Test 16d's *unreadable*-plist abort path had coverage; the happy path had none. Gated on real `plutil` + `jq` being on `PATH` (skips on Linux CI, same pattern `test-loom-daemon-update.sh` scenarios 21/22 already use for the same macOS-only dependency) so it never fails a Linux runner on a genuinely unavailable binary.

## Verification

- `bash defaults/scripts/tests/test-loom-dispatcher.sh` — 99/99 passed, with `LOOM_RUNTIME` unset.
- `LOOM_RUNTIME=claude bash defaults/scripts/tests/test-loom-dispatcher.sh` — 99/99 passed identically (reproduces the exact leak scenario from the issue; confirms it's already closed by #5168 and stays closed).
- Locally faked a minimal `plutil` (Python + `plistlib`, `-convert json -o -` only) on `PATH` to force Test 16f to actually run its real-plist path end-to-end on this Linux host: 103/103 passed (99 base + 4 new Test 16f assertions), confirming the new test's logic is correct, not just its skip-guard.
- `shellcheck -x` on the modified file: only pre-existing informational `SC2015`/`SC2012` notes, unrelated to this change.

**What I could and couldn't exercise directly:**
- **Test 16b's fix (forcing `uname` -> Linux)**: exercised directly — this test host is Linux, so before the fix the "real OS" branch already matched by accident; the fake-`uname` override makes the OS the fixture is read against explicit and host-independent rather than accidental, and I verified the fake-Linux-uname path still produces the same 4 passing assertions.
- **The Darwin branch itself** (i.e. would Test 16b/16f *actually* fail/pass on a real macOS host pre/post-fix): reasoned about, not run on real hardware — I don't have a Darwin host available. Reasoning: `scripts/loom`'s branch is a plain `[[ "$(uname -s)" == "Darwin" ]]` string check with no other host-dependent behavior in the harvest path, and the fake-`uname` scripts in both Test 16b and 16f are placed first on `PATH`, so `uname -s` resolves identically regardless of the real host OS. I'm confident this generalizes to a real Darwin host but flagging it as reasoned-not-verified per the issue's ask.

## Files changed

- `defaults/scripts/tests/test-loom-dispatcher.sh`

### DIFF
--- a/defaults/scripts/tests/test-loom-dispatcher.sh
+++ b/defaults/scripts/tests/test-loom-dispatcher.sh
@@ -392,8 +392,20 @@ Environment=LOOM_WORK_FINDER=1
 Environment=LOOM_MAIN_HEALTH_GATE=1
 Environment=PATH=/opt/sentinel-4581/bin:/usr/bin:/bin
 EOF
+# Fake `uname` -> Linux so the systemd-gated harvest branch fires
+# deterministically regardless of the REAL host OS this suite happens to run
+# on (mirrors Test 16d's Darwin override below, in the opposite direction —
+# #5169). Without this, a real Darwin host takes scripts/loom's Darwin branch
+# unconditionally and never reads the systemd-unit fixture this test writes,
+# so the harvest finds nothing and every assertion below fails.
+FAKE_UNAME_LINUX_BIN=$(mktemp -d)
+cat > "$FAKE_UNAME_LINUX_BIN/uname" <<'EOF'
+#!/usr/bin/env bash
+echo "Linux"
+EOF
+chmod +x "$FAKE_UNAME_LINUX_BIN/uname"
 set +e
-out=$(cd "$CONSUMER" && PATH="/usr/bin:/bin" HOME="$HOME_HARVEST" LOOM_HOME="$CHK_HARVEST" bash "$DISPATCHER" restart --machine 2>&1)
+out=$(cd "$CONSUMER" && PATH="$FAKE_UNAME_LINUX_BIN:/usr/bin:/bin" HOME="$HOME_HARVEST" LOOM_HOME="$CHK_HARVEST" bash "$DISPATCHER" restart --machine 2>&1)
 rc=$?
 set -e 2>/dev/null || true
 assert_eq "$rc" "0" "harvest-and-preserve restart fallback still exits 0"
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM — approving.

Verified both hermeticity fixes from #5169:

1. **Test 21 `LOOM_RUNTIME` leak**: confirmed already fixed on `main` (Test 19 and Test 21's "no env" invocations both already carry `env -u LOOM_RUNTIME -u LOOM_WORKSPACE`). Ran the full suite with `LOOM_RUNTIME=claude` set ambiently — 99/99 passed, confirming the leak stays closed and there's no other "no env" invocation missing the scrub.
2. **Test 16b Darwin/uname mismatch**: the fix forces `uname -s` -> `Linux` via a PATH-prepended fake script, mirroring Test 16d's existing (correct) Darwin override in the opposite direction. This makes the systemd-fixture read host-independent rather than accidental (this runner happens to be Linux already, so pre-fix the test passed for the wrong reason).
3. **New Test 16f** (real launchd happy-path harvest): correctly skips on this Linux CI host (no `plutil`) — verified the skip message prints and the suite still reports 99/99. Also faked a minimal `plutil` (Python + `plistlib`) on `PATH` to force the real-plist path to execute end-to-end: all 4 new assertions pass, 103/103 total. This matches the PR description's claimed verification exactly.

Also checked:
- `shellcheck -x` on the modified file — only pre-existing informational `SC2015`/`SC2012` notes, none introduced by this diff.
- Branch is up to date with `main` (no rebase needed).
- All CI checks pass, including "Shell Test Suites (hermetic)" (5m42s) and Shellcheck.
- `mergeStateStatus` is CLEAN.

One minor, non-blocking note: the PR body is honest that the Darwin branch itself wasn't exercised on real macOS hardware (none available) — reasoning that the fake-`uname` prepend makes `uname -s` resolution host-independent is sound given `scripts/loom`'s harvest dispatch is a plain string comparison with no other host-dependent behavior in that path. Not a blocker.

Issue #5302, opened 2026-08-04
daemon: wedged-sweep watchdog give-up is silent, and the live-vs-live dispatch race has no CAS (split from #5017)


rjwalters:
## Context

Split from #5017 by the Builder implementing PR #5301. #5017 originally
bundled two distinct cross-host dispatch defects observed live with
`collisionDetection.enabled=true` on all hosts:

1. **Issue #12 (genuine live-vs-live collision)**: robb-studio's sweep was
   ACTIVELY progressing (checkpoint phase `curator-done`) when worker-1
   started a second sweep 45 minutes (2,727s) later on the SAME issue.
2. **Issue #87 (wedged-claim recovery gap)**: robb-studio's sweep had been
   alive 6h30m at 0.0% CPU, never advanced past checkpoint phase `-`, and
   never created an `issue-87` worktree — a wedge nothing detected or
   escalated.

PR #5301 addresses the piece of #5017's scope the operator's most recent
disposition redefined as concrete: forge-side claim-ownership verification
before `finish_cancel`/`reap_once` restore `loom:building -> loom:issue`
(shared scope with #5282). It does **not** address either of the two items
below, which remain open problems:

### 1. Wedged-sweep escalation gap (issue #87's root cause)

A hung-sweep watchdog already exists
(`loom-daemon/src/sweep_registry/watchdog.rs`, #3887): 0% CPU + no worktree +
no checkpoint past a timeout (default 300s) triggers one bounded
auto-restart. But by design it gives up after exactly one retry — see the
`WatchdogDecision::GiveUp` arm and its `log::error!` call
(`watchdog.rs` around the `watchdog_once` function). "Left for the operator"
today means only a daemon-log line; nothing reaches an operator-visible
surface (a forge comment, a status/health flag, an alert). A 6.5-hour silent
hold (#87) is consistent with the watchdog firing once, the restart also
wedging, and nothing further happening until an unrelated dispatch
accidentally recovered it.

**Suggested acceptance criteria:**
- [ ] When the watchdog reaches `WatchdogDecision::GiveUp` for an issue,
      surface it somewhere an operator will actually see it beyond the daemon
      log — a forge comment on the issue, a `loom-daemon status` /
      `GetDaemonStatus` health flag, or both.
- [ ] Consider whether `GiveUp` should also (optionally, config-gated) release
      the claim back to `loom:issue` rather than leaving it held indefinitely,
      now that the cross-host claim-ownership check from PR #5301 makes a
[...]

PR #5307, opened 2026-08-04
feat(daemon): surface watchdog give-up via a forge comment, not just a log line

rjwalters:
## Summary

`WatchdogDecision::GiveUp` (the startup watchdog's, `loom-daemon/src/sweep_registry/watchdog.rs`) previously only called `log::error!` when it exhausted its one bounded auto-restart, so a wedged sweep going silent for good was invisible to any operator-facing surface — nothing beyond the daemon log. This adds a best-effort forge comment on the wedged issue when the watchdog gives up, and verifies (no code change) that the separate live-vs-live dispatch-collision path is already surfaced loudly.

## Changes

- `loom-daemon/src/sweep_registry/watchdog.rs`:
  - Added `WATCHDOG_GAVEUP_COMMENT_MARKER` const and a new
    `SweepRegistry::post_watchdog_gaveup_comment(issue, elapsed)` method,
    mirroring the existing `apply_quarantine_label` "post a comment so the
    pause is visible to a human on the forge, not just in the daemon log"
    pattern (same `gh_bin` / `skip_label_flip` / `output_with_timeout` /
    `reap_gh_timeout` wiring already used elsewhere in this registry).
  - Called it from inside the existing `self.watchdog_gaveup.insert(issue)`
    dedup guard in `watchdog_once`'s `WatchdogDecision::GiveUp` arm, so it
    fires exactly once per issue — never once per tick, never again on a
    later tick that re-observes the same already-given-up state.
  - Every failure to post is logged at `warn` (not swallowed at `debug`) —
    a silently-dropped give-up notice would defeat the whole point.

## Acceptance Criteria Verification

**Item 1 (core AC — implemented):**
- [x] When the watchdog reaches `WatchdogDecision::GiveUp`, it now also posts
  a forge comment on the issue (in addition to the existing daemon-log
  line) — verified by the new unit test below.
- [x] The existing per-issue dedup (`watchdog_gaveup`) is reused as-is, so
  the comment posts once per issue, not once per tick.
- [ ] Optional sub-bullet (config-gated claim release back to `loom:issue`
  on give-up) — left as a documented follow-on per the curator's own
  guidance ("Optional — implement only if small and well-scoped; otherwise
  leave as a documented follow-on"). Not implemented here: it changes
  claim-ownership semantics and is a separate decision from "make the
  give-up visible", better scoped as its own issue if wanted.

**Item 2 (verification-only per the curator's explicit guidance — no code
change):**
- [x] Confirmed `detect_and_record_collision`
[...]

PR #5310, opened 2026-08-04
fix(daemon): surface cross-host collision count in WorkFinderTickSummary

rjwalters:
## Summary

Surfaces the cross-host dispatch **collision count** on the wire-carried
`WorkFinderTickSummary`, so `loom-daemon status` / `GetDaemonStatus` /
`loom-daemon health` can see it without scraping the daemon log.

> **Scope note (rebased 2026-08-04).** This PR originally also carried an
> item-1 "watchdog give-up posts a forge comment" change. While it was in
> review, PR #5307 landed the functionally identical mechanism on `main`
> (`post_watchdog_gaveup_comment` / `WATCHDOG_GAVEUP_COMMENT_MARKER`) — the
> same `WatchdogDecision::GiveUp` call site, the same `watchdog_gaveup` dedup
> guard, independently reinvented under a different name. Per the Judge's
> requested fix, that entire duplicated hunk (the const, the method, the
> call-site wiring, and its unit test) has been **dropped**; `main`'s version
> is kept untouched. Only the collision-count work below remains.

Follow-up to #5302 / #5307 — #5302 was already closed by #5307, so this PR
no longer carries a closing keyword.

## What changed

`TickReport::collisions` was already read into the per-tick
`work_finder: tick — …` log line (#4085), but never reached the wire-carried
`WorkFinderTickSummary` that feeds `loom-daemon status` / `GetDaemonStatus` /
`loom-daemon health` — an operator polling status without also tailing the
daemon log had no way to see the collision count.

- `loom-daemon/src/types.rs`: new `WorkFinderTickSummary.collisions: u64`
  (`#[serde(default)]`, so pre-#5302 wire data and older clients still parse —
  an absent field reads as `0`), plus a `reason_summary()` term
  (`"N cross-host-collision(s)"`) emitted **only** when non-zero, matching the
  "only non-zero terms" discipline every other counter in that summary follows.
- `loom-daemon/src/work_finder.rs`: `publish_tick_summary_at` now carries
  `report.collisions` through to the published summary.
- New tests: `publish_tick_summary_carries_collisions_through` (the count
  survives publication and appears in `reason_summary()`) and
  `reason_summary_omits_zero_collisions` (a clean tick says nothing about
  collisions).

## Explicitly out of scope

- **No CAS built.** The cross-host dispatch race is accepted as-is per the
  Curator's recommendation; #4028 Phase 2 remains a separate, architecturally
  larger effort. `detect_and_record_collision` (`sweep_registry/guards.rs`)
  already `log::warn!`s a full diagnostic record on whichever host detects a
  collision, satisfying #5017's original AC3 as worded — this PR only closes
  the status/health *visibility* gap on top of it.
- `peer_claims.rs` (soft peer-claim advertisement, #4028 Phase 1) is
  unmodified.
- `sweep_registry/watchdog.rs` is unmodified — byte-identical to `main`.

## Test Plan

Verified on the rebased branch (base `533c06b2`):

- `cargo build -p loom-daemon` — clean
- `cargo fmt --check` — clean
- `cargo test -p loom-daemon --lib -- work_finder::tests types::tests` —
  170 passed, 0 failed (includes both new tests)
- `cargo test -p loom-daemon --lib -- sweep_registry::watchdog::tests` —
  64 passed, 0 failed (confirms `main`'s
  `watchdog_gaveup_posts_forge_comment_exactly_once` is intact and untouched)

### DIFF
--- a/loom-daemon/src/sweep_registry/watchdog.rs
+++ b/loom-daemon/src/sweep_registry/watchdog.rs
@@ -134,6 +134,15 @@ pub const REVIEW_STALL_TIMEOUT_ENV: &str = "LOOM_SWEEP_REVIEW_STALL_TIMEOUT_SECS
 /// every line it writes and is never disturbed.
 pub const DEFAULT_REVIEW_STALL_TIMEOUT_SECS: u64 = 2700;
 
+/// Marker prefix for the forge comment `post_watchdog_giveup_comment` posts
+/// when the startup watchdog reaches [`WatchdogDecision::GiveUp`] for an issue
+/// (Issue #5302). Mirrors `quarantine::QUARANTINE_COMMENT_MARKER`'s role — a
+/// stable, greppable prefix identifying the comment's source — for a
+/// different failure mode (a wedged startup the watchdog could not self-heal
+/// after its one bounded retry, not an insta-crash quarantine).
+pub const WATCHDOG_GIVEUP_COMMENT_MARKER: &str =
+    "Auto-detected wedge by loom-daemon watchdog (#3887)";
+
 /// The watchdog's per-sweep decision (Issue #3887). Pure state machine —
 /// [`watchdog_decision`] maps `(elapsed, timeout, made_progress,
 /// already_retried)` onto exactly one of these.
@@ -754,7 +763,7 @@ impl SweepRegistry {
             match watchdog_decision(elapsed, timeout, made_progress, already_retried) {
                 WatchdogDecision::Healthy => {}
                 WatchdogDecision::GiveUp => {
-                    // Bounded: already retried once. Log once per issue.
+                    // Bounded: already retried once. Log + surface once per issue.
                     if self.watchdog_gaveup.insert(issue) {
                         log::error!(
                             "watchdog: sweep for issue #{issue} ({sweep_id}) is still stuck \
@@ -763,6 +772,12 @@ impl SweepRegistry {
                              the MCP-init hang).",
                             elapsed.as_secs()
                         );
+                        // Issue #5302: the log line above was the ONLY trace of a
+                        // give-up — invisible to anyone not tailing this daemon's
+                        // own log. Post a forge comment so the wedge is visible on
+                        // the issue itself, same escalation surface an operator
+                        // already watches.
+                        self.post_watchdog_giveup_comment(issue, elapsed.as_secs());
                     }
                 }
                 WatchdogDecision::Restart => {
@@ -837,6 +852,66 @@ impl SweepRegistry {
             }
         }
         restarts
+    }
+
+    /// Best-effort forge comment posted when the startup watchdog reaches
+    /// [`WatchdogDecision::GiveUp`] for `issue` (Issue #5302 — split from
+    /// #5017). Before this, `watchdog_once`'s `GiveUp` arm only logged, which
+    /// meant a wedge nothing else caught (#5017's issue #87: a sweep alive
+    /// 6.5h at 0% CPU) stayed invisible until an unrelated dispatch happened
+    /// to recover it.
+    ///
+    /// Mirrors [`SweepRegistry::apply_quarantine_label`]'s forge-mutation
+    /// pattern: skipped entirely when `skip_label_flip` is set (test
+    /// fixtures), respects `config.gh_bin` / `LOOM_REPO`, and is bounded by
+    /// [`reap_gh_timeout`] so a wedged `gh` invocation can never block a
+    /// watchdog tick. Every failure is logged and never propagated — this is
+    /// purely an operator-visibility addition, not load-bearing state.
+    ///
+    /// Callers are responsible for the once-per-issue dedup (the
+    /// `watchdog_gaveup.insert(issue)` guard at the single call site) — this
+    /// method always posts unconditionally when called.
+    fn post_watchdog_giveup_comment(&self, issue: u32, elapsed_secs: u64) {
+        if self.config.skip_label_flip {
+            return;
+        }
+        let gh = self
+            .config
+            .gh_bin
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("gh"));
+
+        let body = format!(
+            "{marker}: this issue's sweep has made no progress (no worktree, no checkpoint, \
+             log stuck at the spawn header) for {elapsed_secs}s — even after the daemon's \
+             startup watchdog already auto-cancelled and re-dispatched it once. The watchdog is \
+             bounded to a single retry and will not act again; this issue stays held in \
+             `loom:building` until an operator intervenes. Suggested next steps: cancel the \
+             stuck sweep (`loom-daemon cancel --issue {issue}` or `mcp__loom__cancel_sweep`) and \
+             re-dispatch, or investigate the MCP-init hang directly (see \
+             `.loom/logs/sweep-issue-{issue}.log`).",
+            marker = WATCHDOG_GIVEUP_COMMENT_MARKER,
+        );
+        let mut comment = Command::new(&gh);
+        comment
+            .arg("issue")
+            .arg("comment")
+            .arg(issue.to_string())
+            .arg("--body")
+            .arg(body);
+        comment.current_dir(&self.config.workspace_root);
+        if let Ok(repo) = std::env::var("LOOM_REPO") {
+            comment.arg("--repo").arg(repo);
+        }
+        let timeout = reap_gh_timeout();
+        match output_with_timeout(comment, timeout) {
+            Ok(Some(_)) => {}
+            Ok(None) => log::warn!(
+                "watchdog: give-up comment for #{issue} exceeded {}s, killed (#3973)",
+                timeout.as_secs()
+            ),
+            Err(e) => log::warn!("watchdog: give-up comment for #{issue} failed: {e}"),
+        }
     }
 
     // ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


rjwalters:
Temporarily closing to force GitHub to recompute the PR merge base after the rebase-equivalent merge (its cached diff still showed the pre-#5307 base). Reopening immediately.

rjwalters:
## Doctor: feedback addressed — duplicated Item 1 dropped, rebased onto current `main`

Judge's requested fix is applied exactly as asked. `mergeStateStatus` went `DIRTY` → `CLEAN`.

### Dropped (duplicate of #5307, already on `main`)

The entire Item 1 watchdog hunk is gone from this branch — `loom-daemon/src/sweep_registry/watchdog.rs` is now **byte-identical to `main`**:

- `WATCHDOG_GIVEUP_COMMENT_MARKER` (the const)
- `SweepRegistry::post_watchdog_giveup_comment` (the method)
- its call-site wiring inside the `WatchdogDecision::GiveUp` arm
- its unit test `watchdog_giveup_posts_a_forge_comment_exactly_once`

Not renamed, not re-added anywhere. `main`'s `post_watchdog_gaveup_comment` / `WATCHDOG_GAVEUP_COMMENT_MARKER` (from #5307, commit `998e2390`) is untouched — verified by `git grep -n "GIVEUP_COMMENT\|post_watchdog_giveup"` returning nothing, and by `main`'s own `watchdog_gaveup_posts_forge_comment_exactly_once` still passing.

### Kept (Item 2 — genuinely new, confirmed absent from `main`)

- `loom-daemon/src/types.rs` — `WorkFinderTickSummary.collisions: u64` (`#[serde(default)]`) + the `"N cross-host-collision(s)"` term in `reason_summary()`, emitted only when non-zero
- `loom-daemon/src/work_finder.rs` — `publish_tick_summary_at` carries `report.collisions` through
- both accompanying tests: `publish_tick_summary_carries_collisions_through` and `reason_summary_omits_zero_collisions`

Nothing needed adapting — `main` had moved in `work_finder.rs` since this branch forked, but not in either hunk this PR touches.

### Resulting diff

```
loom-daemon/src/types.rs       | 14 ++++++++++++++
loom-daemon/src/work_finder.rs | 42 ++++++++++++++++++++++++++++++++++++++++++
2 files changed, 56 insertions(+)
```

Merge base is now `533c06b2` (current `main` tip), 0 commits behind.

### Verification (run locally on the rebased tree)

| Command | Result |
|---|---|
| `cargo build -p loom-daemon` | pass (exit 0) |
| `cargo clippy -p loom-daemon --all-targets -- -D warnings` | pass (exit 0, no warnings) |
| `cargo fmt --check` | pass (exit 0, no diff) |
| `cargo test -p loom-daemon --lib -- work_finder::tests types::tests` | **170 passed, 0 failed** (includes both new tests) |
| `cargo test -p loom-daemon --lib -- sweep_registry::watchdog::tests` | **64 passed, 0 failed** |

Remote CI is green on all 14 checks, including the full `Rust Unit Tests` job (3m59s) and `Rust Linting & Build Check`. `Detect Changes` now correctly skips the shell/dashboard/installer jobs, confirming GitHub sees the 2-file scope.

> Note on local full-suite runs: `cargo test -p loom-daemon --lib` (unfiltered, fully parallel) shows a large batch of unrelated failures on this host — `main_health_gate::*`, `forge_cmd::*`, `live_claim::*`, `sweep_registry::guards::restore_label_to_ready_does_not_add_loom_issue_when_target_is_pr`, etc. Those are pre-existing environment/parallelism artifacts (global `LOOM_REPO`/env-var races in non-`#[serial]` tests plus subprocess-spawn contention), not regressions: each passes standalone, none touch the two files in this diff, and CI's own `Rust Unit Tests` job is green.

### Housekeeping

- PR title and body rewritten to describe only the remaining collision-count scope.
- `Closes #5302` removed — #5302 was already closed by #5307, so the body now reads "Follow-up to #5302 / #5307" rather than carrying a stale closing keyword.
- The PR was briefly closed and reopened to force GitHub to recompute its cached merge base, which was still pinned at the pre-#5307 base and showing a phantom 58-file diff. GitHub's own compare API had the correct 2-file answer throughout; the reopen made the PR object agree.
- Rebase was landed as a history-preserving `-s ours` merge of the pre-rebase tip rather than a force-push, since this repo's `guards.forceScope` defaults to `all` and force-push is an ask-gated operation. The resulting **tree is identical** to the straight rebase (`git diff <rebased> HEAD` is empty) and the diff against `main` is unchanged.

Handing back for re-review.

rjwalters:
## Judge (re-review after Doctor cycle 1/1): APPROVED

The prior rejection is fully satisfied. I re-verified every claim in the Doctor's report independently rather than taking it at face value — including the unusual `git merge -s ours` history rewrite, which was the one thing that genuinely needed proving.

### 1. Prior rejection satisfied — Item 1 is gone, `main`'s #5307 version is intact

- `git diff origin/main:loom-daemon/src/sweep_registry/watchdog.rs d09da2e6:...watchdog.rs` → **empty**. Byte-identical to `main`.
- `git grep "GIVEUP_COMMENT\|post_watchdog_giveup" d09da2e6` → **no matches**. The const, the method, the `GiveUp` call-site wiring, and its unit test are all dropped.
- `main`'s #5307 mechanism is untouched and present on the head: 7 references to `WATCHDOG_GAVEUP_COMMENT_MARKER` / `post_watchdog_gaveup_comment`, and `fn watchdog_gaveup_posts_forge_comment_exactly_once` still lives at `watchdog.rs:2522`.

### 2. The `-s ours` merge is safe — verified, not assumed

This is the part that warranted real scrutiny: an `ours` merge discards the incoming side wholesale, so the risk was that it also silently reverted unrelated `main` content that happened to be reachable only through the pre-rebase tip. It did not.

| Check | Result |
|---|---|
| `d09da2e6^{tree}` vs `bc18db83^{tree}` (rebased first parent) | **Identical** (`1cf32163…`) — the merge took the rebased tree exactly, as `-s ours` should |
| `git diff bc18db83 d09da2e6` | **empty** — the merge commit added nothing of its own |
| Merge base with `origin/main` | `533c06b2` ✅ (as claimed) |
| Commits unique to the discarded parent `c515a5eb` | exactly **1** — `c515a5eb` itself, the pre-rebase tip. Nothing else was in the discarded set. |
| `c515a5eb`'s own base (`28e4118a`) is an ancestor of `origin/main` | **Yes** — so no `main` commit could have been reachable only via the discarded side |

The decisive test — because a 3-dot diff against the merge base would *not* have caught a revert of `main` content landed after `533c06b2`:

```
git merge-tree --write-tree origin/main d09da2e6   →  84106935
git diff --stat origin/main 84106935
 loom-daemon/src/types.rs       | 14 ++++++++++++++
 loom-daemon/src/work_finder.rs | 42 ++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 56 insertions(+)
```

Simulating the actual merge into the **current** `main` tip (`041b94d2`, three commits past the merge base) produces a tree that differs from `main` by exactly the two intended files, `+56/-0`, with **zero deletions anywhere**. Nothing on `main` is reverted. The `-s ours` shape is cosmetically unusual in the history but functionally equivalent to the straight rebase, as claimed.

### 3. PR state

- **OPEN**, not draft, `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` — the close/reopen did recompute the stale base.sha.
- File list via the GitHub API: exactly 2 files, `types.rs +14/-0`, `work_finder.rs +42/-0`. Matches the local diff.
- CI on head `d09da2e6`: **22 success, 6 skipped, 0 failures** (skips are path-filtered jobs — Installer, Dashboard, MCP, Shell suites, Codex smoke, worker image — none of which this Rust-only diff touches). `Rust Linting & Build Check`, `Rust Unit Tests`, and `Rust OTLP Feature Build & Test` all pass.
- Local re-run in the builder worktree at the exact head SHA (clean tree): both new tests pass.
- PR body no longer carries a closing keyword — reads "Follow-up to #5302 / #5307". Correct: #5302 was already closed by #5307, so a `Closes` here would have been a no-op at best.

### 4. The remaining diff on its merits

**`collisions: u64` with `#[serde(default)]` — right shape.** `#[serde(default)]` is correct and necessary: `WorkFinderTickSummary` is an explicitly documented *wire contract*, so pre-#5302 payloads and older clients must parse an absent field as `0`. The `u64` (rather than the `usize` every sibling counter uses) is a deliberate, defensible choice — it mirrors `TickReport::collisions: u64` exactly, so `publish_tick_summary_at` carries the value across with no lossy cast. The doc comment states the mirroring explicitly. I'd have accepted `usize` too, but `u64` avoids the cast and is the better call.

**`reason_summary()` term — correct and correctly placed.** The guard is `> 0`, matching the "only non-zero terms" discipline the rest of the method follows, and the term sits after the counter loop but before the `HALTED` / `SATURATION-HELD` flags — counters with counters, flags last. It's a separate `if` rather than an entry in the `[(n, label); 10]` array purely because that array is homogeneously `usize`-typed and a `u64` can't join it; that's a language constraint, not an oversight.

**Pluralization** is the one cosmetic inconsistency: `"3 cross-host-collision(s)"` uses an `(s)` suffix while every other term is bare-singular regardless of count (`"3 error"`, `"2 labeled-skip"`). Non-blocking — arguably more readable, just not house style. Not worth a Doctor cycle.

**Both tests are meaningful, not tautological.**
- `publish_tick_summary_carries_collisions_through` asserts two distinct things — the field survives the `TickReport` → `WorkFinderTickSummary` projection *and* surfaces in `reason_summary()`. It goes through the real `publish_tick_summary` path with a `reset_last_tick_summary()` precondition, so it exercises the actual wiring rather than a struct literal.
- `reason_summary_omits_zero_collisions` is the genuinely valuable one: it pins the non-zero guard, which is exactly the regression an unconditional `parts.push` would introduce (a wall-of-zeros summary on every clean tick).

**`publish_tick_summary_at` wiring is correct.** `collisions: report.collisions` is a direct pass-through in the same struct literal as the other counters. Upstream, `report.collisions` is populated at three sites (`:839`, `:987`, and `:1399` where it sums `dispatcher.collisions()` across workspaces after pass 2), so the published value is the same number #4085 already logged. The four other `WorkFinderTickSummary` construction sites (`health.rs`, `ipc.rs`, `collector.rs`, `serve.rs`) all use `..Default::default()`, so the added field is a non-breaking struct change — confirmed by the clean build.

### Non-blocking note (no action required)

`collisions` is **cumulative over the dispatcher's lifetime**, not a per-tick delta, so on a line otherwise composed of per-tick counts (`12 seen, 2 dispatched`) a reader could reasonably infer `3 cross-host-collision(s)` means "3 this tick." The new doc comment is honest about this ("observed by this tick's dispatcher(s)", mirroring `TickReport::collisions`), and the semantics are unchanged from the #4085 log line this mirrors — so the PR introduces no new ambiguity, it just makes an existing one visible in one more place. Flagging for whoever next touches this summary; not a defect in this diff.

Approving: `loom:review-requested` → `loom:pr`.

PR #5320, opened 2026-08-04
fix(tests): derive reaper cross-host-claim fixture timestamp from the clock (#5317)

rjwalters:
Closes #5317

## Problem

Two tests in `loom-daemon/src/sweep_registry/reaper.rs` hardcoded the literal forge timestamp `"2026-08-04T15:59:59Z"` as their fake `gh api .../timeline` response — the fixture standing in for "a newer `loom:building` claim from another host" — while comparing it against a `started_at` of `Utc::now() - chrono::Duration::hours(1)`:

- `cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host`
- `reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host`

Once wall-clock time passed `2026-08-04T16:59:59Z`, `now - 1h` overtook the literal, so the fixture no longer described a *newer* claim. The code under test then correctly restored `loom:building` -> `loom:issue` (the forge no longer shows a peer host's live claim), and the "must NOT restore" assertions failed. This broke `Rust Unit Tests` on every open Rust-touching PR fleet-wide on a date trigger rather than any code change — confirmed live on main's tip and on PR #5313, which does not touch `reaper.rs` at all.

## Fix

Both fixtures now compute the "newer claim" timestamp *inside the fake `gh` script, at invocation time*, via a `date -u` command substitution instead of embedding a literal:

```diff
-  echo '"2026-08-04T15:59:59Z"'
+  echo "\"$(date -u '+%Y-%m-%dT%H:%M:%SZ')\""
```

The fake `gh` therefore answers each `gh api .../timeline` probe with the current UTC time, which is unconditionally ~1h greater than each fixture's `started_at` of `Utc::now() - chrono::Duration::hours(1)`. `SweepRegistry::claim_superseded_on_forge` (`sweep_registry/guards.rs`) vetoes the restore on `labeled_at > claimed_at`, so the fixture's premise now holds at any wall-clock time and no time-bomb window remains. `+%Y-%m-%dT%H:%M:%SZ` emits the trailing-`Z` RFC3339 form the production parser (`parse_max_timestamp`) expects, and the escaped surrounding quotes preserve the JSON-string shape the old literal had.

Test-fixture-only change (2 lines, both inside `mod tests`) — no production code path is modified.

## Revision history (why the earlier description differed)

[...]

PR #5324, opened 2026-08-04
fix(daemon): derive reaper claim-timeline test fixtures from Utc::now(), not a fixed date

### LINKED ISSUE #5319: reaper.rs claim-timeline tests hardcode an absolute date, breaking CI on main permanently after 2026-08-04T15:59:59Z
## Observed

CI is currently red on `main` for `Rust Unit Tests` and `Rust OTLP Feature Build & Test`, independent of any specific PR's diff:

- A direct push-to-main CI run (commit `3e7bc2e9`, unrelated PR #5311) fails: https://github.com/rjwalters/loom/actions/runs/30933834093
- PR #5312's CI run (commit `7ee1f65d`, a different unrelated PR) fails identically: https://github.com/rjwalters/loom/actions/runs/30933192410

Both fail on the exact same two tests in `loom-daemon/src/sweep_registry/reaper.rs`:

```
test sweep_registry::reaper::tests::cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host ... FAILED
test sweep_registry::reaper::tests::reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host ... FAILED
```

with the panic:

```
a cancel whose forge timeline shows a newer claim must NOT restore loom:building -> loom:issue
(would destroy a peer host's live claim, #5017/#5282); got gh invocations: "...issue edit 8802
--remove-label loom:building --add-label loom:issue..."
```

i.e. the test's fake `gh` timeline response is now being read as *older* than the sweep's
`started_at`, so the "another host has a newer claim" branch no longer triggers and the label
restore fires when the test asserts it must not.

## Suspected cause (unverified)

Both tests (added by PR #5301, `loom-daemon/src/sweep_registry/reaper.rs` around lines
2511-2649) hard-code the fake forge timeline response as an **absolute calendar timestamp**:

```rust
echo '"2026-08-04T15:59:59Z"'
```

and compare it against `started_at = Utc::now() - chrono::Duration::hours(1)` (or similar) to
simulate "a newer claim than this sweep's start." Because the fixture timestamp is a fixed
point in time rather than relative to `Utc::now()`, this only holds while `Utc::now() - 1h <
2026-08-04T15:59:59Z`, i.e. only while wall-clock time is before ~17:00 UTC on 2026-08-04. Once
the clock crosses that point (as it now has), `started_at` is no longer before the fixture
timestamp, the "newer claim" branch stops firing, and the test fails — permanently, since the
fixture date never advances. This needs verification by reading the exact comparison logic in
`finish_cancel`/`reap_once` and confirming the fixture should use a `Utc::now() +/- Duration`
offset instead of a fixed date string.

## Impact

Blocking — every open PR's `Rust Unit Tests` / `Rust OTLP Feature Build & Test` CI check now
fails regardless of the PR's own content, since this is a `main`-branch test bug, not something
tied to any one PR's diff. Confirmed evaluating PR #5312 (unrelated capacity/token_bound
changes) — its own new/modified tests pass locally and in isolation; only these two
pre-existing tests fail, identically to a same-day unrelated push to main.

Discovered while evaluating PR #5312 as Judge.

rjwalters:
## Summary

`Rust Unit Tests` and `Rust OTLP Feature Build & Test` have been failing on `main`
for every PR since ~17:00 UTC on 2026-08-04, independent of any PR's own diff. Two
cross-host claim-ownership tests added by #5301 hard-coded their fake `gh api
.../timeline` response as the absolute calendar timestamp `"2026-08-04T15:59:59Z"`,
while the sweep they compare against is anchored to wall clock
(`started_at = Utc::now() - 1h`).

`SweepRegistry::claim_superseded_on_forge` (`sweep_registry/guards.rs:233`) returns
`true` only when `labeled_at > claimed_at`. Once real UTC passed
`2026-08-04T16:59:59Z`, the fixed fixture stopped being newer than `started_at`, the
supersession veto never fired, the code fell through to the real
`restore_label_to_ready` path, and both tests' "must NOT restore `loom:building`"
assertions failed — permanently, on every subsequent run, not intermittently.

This PR makes both fixtures relative to `Utc::now()` so they can never expire again.

## Changes

- `loom-daemon/src/sweep_registry/reaper.rs`: `cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host`
  and `reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host` now build
  the mocked timeline timestamp as `Utc::now() + 5min`, rendered with
  `to_rfc3339_opts(SecondsFormat::Secs, true)`, instead of the literal
  `"2026-08-04T15:59:59Z"`. That is strictly after each test's own
  `started_at = Utc::now() - 1h` on every future run, preserving the original intent
  ("a newer claim than this sweep's `started_at`").
- Added comments at both fixtures recording why the timestamp must stay relative, so
  the same time bomb is not reintroduced.

Test-only change; no production code path is touched.

| Confirm the diagnosis against the real comparison logic | ✅ | `claim_superseded_on_forge` (guards.rs:233-249) matches `Some(labeled_at) if labeled_at > claimed_at`; both call sites (`reaper.rs:751` in `finish_cancel`, `reaper.rs:950` in `reap_once`) pass the entry's `started_at` as `claimed_at`. |
| Both named tests fixed to use a relative fixture | ✅ | Fixtures now `Utc::now() + chrono::Duration::minutes(5)`; no literal date remains in either test body. |
| Original test intent preserved ("newer claim") | ✅ | `now + 5min > now - 1h` holds unconditionally; both tests still assert the timeline probe happened AND that no `--remove-label loom:building` was issued. |
| No other test in the repo has the same time-bomb shape | ✅ | Scanned every `.rs`/`.ts`/`.sh` file for literal RFC3339 dates and cross-referenced against `Utc::now()`/`SystemTime::now()` usage per function. The only other literal-date-vs-ordering fixture (`quarantine.rs::quarantine_ttl_expiry_never_releases_a_later_manual_repark`, `2020-01-01` vs `2030-01-01`) compares the two literals against **each other**, never against the wall clock, so its ordering is stable forever. All remaining hits (`tokens_pool`, `capacity`, `pipeline_snapshot`, `metrics_collector`, `observability`, dashboard tests) are round-trip/format or fixed-clock tests with no `now()` comparison. |
| Two named tests pass | ✅ | Each run in isolation with `--exact` (see Test Plan). |
| Reaper suite passes | ✅ | `cargo test --lib sweep_registry::reaper` — 52 passed, 0 failed. |

## Test Plan

```
# Both previously-failing tests, in isolation (the repro from the issue):
cargo test --package loom-daemon --lib \
  "sweep_registry::reaper::tests::cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host" \
  -- --exact --nocapture
  => test result: ok. 1 passed; 0 failed

cargo test --package loom-daemon --lib \
  "sweep_registry::reaper::tests::reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host" \
  -- --exact --nocapture
  => test result: ok. 1 passed; 0 failed

# Whole module:
cargo test --lib sweep_registry::reaper
  => test result: ok. 52 passed; 0 failed; 3429 filtered out

# Lint/format:
cargo fmt      => clean
cargo clippy --all-targets -- -D warnings   => clean
```

Note on provenance: this issue was double-dispatched — a concurrent `--claim-owned 5319`
sweep pushed the commit on this branch first while an independent builder was preparing
an equivalent fix (same root cause, same relative-timestamp approach). Per the
"defer to the PR that pushed first, don't force-push" convention the pushed commit was
kept as-is and independently re-verified (all runs above are against `8372dd0`) rather
than force-pushed over. The concurrent sweep had not opened a PR after its push, so this
PR carries the work forward.

Closes #5319

### DIFF
--- a/loom-daemon/src/sweep_registry/reaper.rs
+++ b/loom-daemon/src/sweep_registry/reaper.rs
@@ -2519,9 +2519,19 @@ mod tests {
         // real `restore_label_to_ready` call would show up in the log as a
         // `--remove-label loom:building` invocation, so asserting its absence
         // is sufficient to prove the forge check vetoed the restore.
+        //
+        // The fixture timestamp is computed relative to `Utc::now()` at
+        // test-run time (NOT a hardcoded calendar date) so the "newer claim"
+        // branch keeps firing on every future test run — a fixed absolute
+        // date silently stops being "after `started_at`" once wall-clock
+        // time passes it, which is exactly what broke this test permanently
+        // on 2026-08-04 (issue #5319).
+        let forge_labeled_at = (Utc::now() + chrono::Duration::minutes(5))
+            .to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
         let script = format!(
-            "#!/usr/bin/env bash\nprintf '%s\\n' \"$*\" >> \"{}\"\nif [[ \"$1\" == \"api\" ]]; then\n  echo '\"2026-08-04T15:59:59Z\"'\nfi\nexit 0\n",
-            gh_log.display()
+            "#!/usr/bin/env bash\nprintf '%s\\n' \"$*\" >> \"{}\"\nif [[ \"$1\" == \"api\" ]]; then\n  echo '\"{}\"'\nfi\nexit 0\n",
+            gh_log.display(),
+            forge_labeled_at
         );
         std::fs::write(&fake_gh, &script).unwrap();
         let mut perms = std::fs::metadata(&fake_gh).unwrap().permissions();
@@ -2593,11 +2603,17 @@ mod tests {
         let dir = tempdir().unwrap();
         let (mut registry, gh_log) = fixture_registry(dir.path());
         // Override the fixture's fake `gh` so `gh api .../timeline` answers
-        // with a labeling timestamp AFTER `started_at` (set below).
+        // with a labeling timestamp AFTER `started_at` (set below). Computed
+        // relative to `Utc::now()` at test-run time, not a hardcoded
+        // calendar date — see the sibling `cancel_skips_restore_...` test's
+        // comment for why (#5319).
         let fake_gh = dir.path().join("fake-gh-timeline.sh");
+        let forge_labeled_at = (Utc::now() + chrono::Duration::minutes(5))
+            .to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
         let script = format!(
-            "#!/usr/bin/env bash\nprintf '%s\\n' \"$*\" >> \"{}\"\nif [[ \"$1\" == \"api\" ]]; then\n  echo '\"2026-08-04T15:59:59Z\"'\nfi\nexit 0\n",
-            gh_log.display()
+            "#!/usr/bin/env bash\nprintf '%s\\n' \"$*\" >> \"{}\"\nif [[ \"$1\" == \"api\" ]]; then\n  echo '\"{}\"'\nfi\nexit 0\n",
+            gh_log.display(),
+            forge_labeled_at
         );
         std::fs::write(&fake_gh, &script).unwrap();
         let mut perms = std::fs::metadata(&fake_gh).unwrap().permissions();


rjwalters:
Closing as duplicate/superseded: the identical fix already merged to `main` via #5320 (which closed sibling issue #5317, describing the same hardcoded-`2026-08-04T15:59:59Z`-fixture bug in the same two reaper cross-host-claim tests). Issue #5319 itself was closed as "not planned" by the operator for the same reason. Verified `origin/main`'s `reaper.rs` already computes the fixture timestamp relative to wall-clock time (via `date -u`) rather than a hardcoded date — this PR's rebase now shows CONFLICTING/DIRTY against main because both PRs touched the identical lines. No further action needed on this PR.

rjwalters:
❌ **Closing — redundant, already fixed on `main`**

This PR's rebase onto `origin/main` hit a genuine conflict in `loom-daemon/src/sweep_registry/reaper.rs` at the exact two fixtures this PR touches. Investigating why: sibling issue #5317 (identical bug, same two hardcoded-timestamp tests) was independently fixed and merged as PR #5320 (`6b70ad35`, merged 2026-08-04T18:06:47Z) — using a different mechanism (the fake `gh` script computes `$(date -u '+%Y-%m-%dT%H:%M:%SZ')` at invocation time instead of baking in a Rust-computed `Utc::now() + 5min` literal), but achieving the same outcome: the fixture timestamp is no longer a hardcoded absolute date, so it can't expire again.

The operator already closed the underlying issue #5319 as "not planned" for exactly this reason (comment at 2026-08-04T18:09:12Z): "already fixed by #5320 ... Verified on current main (6b70ad35)".

Confirmed independently here: `origin/main`'s current `reaper.rs` already contains a working, non-time-bomb fix for both `cancel_skips_restore_when_forge_shows_a_newer_claim_from_another_host` and `reap_skips_restore_when_forge_shows_a_newer_claim_from_another_host`. Merging this PR on top would just replace one working fix with a functionally equivalent one and isn't worth the conflict resolution.

Closing without merging — no further action needed. Thanks for the fast turnaround.

PR #5332, opened 2026-08-04
fix(skill-router): suppress AGENT_ROUTE on explicit Loom-decline prompts

### LINKED ISSUE #5327: skill-router: bare \bfix\b routes a prompt that explicitly declines Loom to /loom:doctor
## Summary

`skill-router.sh` routes a prompt to `/loom:doctor` on a bare `\bfix\b` match, with no awareness that the same prompt is *explicitly declining* Loom. A message whose literal content is "we don't need to use loom for every tiny change" was answered with `AGENT_ROUTE: /loom:doctor`.

Low severity — the router is advisory, always exits 0, and never blocks. But it is a precise, reproducible false positive in exactly the direction that erodes trust in the suggestion: it fires hardest on the user who has just said they don't want it.

## Repro

```bash
P="just fix it here and commit and push; we don't need to use loom for every tiny change"
echo "$P" | grep -qiE '\bfix\b|\bbug\b|\bbroken\b|changes.requested|merge.conflict|address (the )?feedback' \
  && echo "routes to /loom:doctor"
# -> routes to /loom:doctor
```

Same result for every natural phrasing of the same intent:

| Prompt | Routes to |
|---|---|
| `don't use loom to fix this` | `/loom:doctor` |
| `no need for a loom agent, just fix it inline` | `/loom:doctor` |
| `fix this yourself without loom` | `/loom:doctor` |

Observed in `rjwalters/anvil` on Loom 0.18.0 during a `/repo:all` hygiene pass. The user had just watched an unrelated guard redirect `gh pr merge` to `merge-pr.sh` and was pushing back on perceived ceremony; the routing suggestion landed on the pushback itself.

## Cause

`.loom/config/skill-routes.json`, the `/loom:doctor` route:

```
"pattern": "\\bfix\\b|\\bbug\\b|\\bbroken\\b|changes.requested|merge.conflict|address (the )?feedback"
```

`.loom/hooks/skill-router.sh:155` matches it with a single case-insensitive `grep -qiE` over the whole prompt, first-match-wins, with no negation handling.

The config's own `description` field states the post-#3609 intent:

> Patterns are phrase/word-boundary anchored (issue #3609) so common software vocabulary ('builders', a bare 'issue', a repo path like 'rjwalters/loom') no longer mis-routes — only genuine imperatives do.

A bare `\bfix\b` does not meet that bar. It *is* common software vocabulary, and word-boundary anchoring does nothing here — the word really is "fix", used as an imperative, just an imperative aimed at the assistant rather than at Loom. #3609 tightened the token boundary but not the surrounding intent, so this class survived it.

Note the minimum-signal gate (`skill-router.sh:88-98`) does not help: it skips prompts under 3 words or starting with `/`, and these are neither.

- [ ] A prompt containing an explicit decline near a route keyword ("without loom", "don't use loom", "no loom", "not via loom", "myself", "inline", "directly") emits no `AGENT_ROUTE`.
- [ ] Regression cases covering the four prompts in the table above, asserting silent exit 0 with no `additionalContext`.
- [ ] Existing positive routes still fire — in particular a genuine `fix the failing test in <file>` still reaches `/loom:doctor`.
- [ ] Whatever shape the fix takes, it stays fail-open: any internal error exits 0 silently, per the script's existing top-level `ERR` trap contract.

## Implementation notes

A general negation regex over natural language is a losing game, so the narrow version is probably right: a single suppression pattern checked *after* a route matches and *before* the directive is emitted, listing the handful of literal anti-routing phrases above. That keeps it in the same "literal, no clever regex" spirit as the existing task-notification guard at `skill-router.sh:73-78`, which is already explicitly commented as using literal matching "so this guard cannot itself false-positive on human text".

Suppression could also be made observable the way route decisions already are, so a suppressed match is debuggable rather than silently absent.

An alternative worth considering instead: since the config already supports `skill-routes.local.json` overrides, a repo that wants no routing at all can already disable it locally. That is a workaround, not a fix — the false positive would still ship to every other consumer — but it means this issue is not urgent for any individual repo.

rjwalters:
## Summary

`skill-router.sh` routes any prompt matching a route pattern (e.g. a bare
`\bfix\b`) to the corresponding agent, even when the prompt is explicitly
declining Loom usage — e.g. "fix this yourself without loom" or "we don't
need to use loom for every tiny change" both matched `\bfix\b` and emitted
`AGENT_ROUTE: /loom:doctor`, which is exactly backwards: it fires hardest on
users who just said they don't want Loom involved.

This adds an anti-route suppression check, evaluated **after** a route
pattern matches and **before** the `AGENT_ROUTE` directive is built. It does
a literal (non-regex) substring match against a short list of decline
phrases ("without loom", "don't use loom", "don't need to use loom", "no
loom", "not via loom", "myself", "inline", "directly", etc.) — literal
matching mirrors the existing task-notification guard a few lines above it
in the same script, which is explicitly commented as literal so it can't
itself false-positive on ordinary human text. A suppressed match is logged
via the existing `log_hook_error` diagnostic sink for observability,
consistent with the script's existing logging conventions.

The check runs entirely within the script's existing fail-open contract —
no new failure modes are introduced, and the top-level `ERR` trap /
`set -o pipefail` behavior is untouched.

## Changes

- `defaults/hooks/skill-router.sh` (canonical source) + `.loom/hooks/skill-router.sh`
  (deployed copy, kept byte-identical per the test suite's sync check): add
  the `ANTI_ROUTE_PHRASES` suppression block between pattern matching and
  `AGENT_ROUTE` emission.
- `defaults/hooks/tests/test-skill-router.sh`: add regression cases for the
  four example decline prompts from the issue (all assert silent/no-output),
  plus a positive control confirming `fix the failing test in <file>` still
  routes to `/loom:doctor`.

## Test plan

- [x] `bash defaults/hooks/tests/test-skill-router.sh` — 34/34 passed,
  including the 4 new anti-route regression cases and the positive control.
- [x] `shellcheck defaults/hooks/skill-router.sh` — no warnings.
- [x] `bash -n defaults/hooks/skill-router.sh` — syntax OK.
- [x] Verified `.loom/hooks/skill-router.sh` and `defaults/hooks/skill-router.sh`
  remain byte-identical (enforced by the test suite's sync check).

Closes #5327

### DIFF
--- a/.loom/hooks/skill-router.sh
+++ b/.loom/hooks/skill-router.sh
@@ -170,6 +170,12 @@ done
 # present, using literal (non-regex) substring checks against the full prompt
 # — literal so this guard, like the task-notification guard above, cannot
 # itself false-positive on ordinary human text.
+#
+# INVARIANT: every phrase below must mention "loom" explicitly. Bare
+# single-word phrases ("myself", "inline", "directly") suppress ordinary
+# prompts that never decline Loom at all — e.g. "review this PR directly" or
+# "clean up this code inline" — so they are deliberately excluded. Suppression
+# requires an explicit decline of Loom, never a stylistic adverb.
 if [[ -n "$MATCHED_AGENT" ]]; then
     ANTI_ROUTE_PHRASES=(
         "without loom"
@@ -182,9 +188,6 @@ if [[ -n "$MATCHED_AGENT" ]]; then
         "no loom"
         "not use loom"
         "not via loom"
-        "myself"
-        "inline"
-        "directly"
     )
     for phrase in "${ANTI_ROUTE_PHRASES[@]}"; do
         if [[ "$PROMPT_LOWER" == *"$phrase"* ]]; then
--- a/.loom/hooks/skill-router.sh
+++ b/.loom/hooks/skill-router.sh
@@ -160,6 +160,42 @@ for (( i=0; i<ROUTE_COUNT; i++ )); do
     fi
 done
 
+# =============================================================================
+# ANTI-ROUTE SUPPRESSION (#5327)
+# =============================================================================
+# A prompt can hit a route pattern by keyword coincidence while explicitly
+# declining Loom involvement, e.g. "fix this yourself without loom" or "we
+# don't need to use loom for every tiny change" — both contain "fix" and would
+# otherwise route to /loom:doctor. Suppress the match when a decline phrase is
+# present, using literal (non-regex) substring checks against the full prompt
+# — literal so this guard, like the task-notification guard above, cannot
+# itself false-positive on ordinary human text.
+if [[ -n "$MATCHED_AGENT" ]]; then
+    ANTI_ROUTE_PHRASES=(
+        "without loom"
+        "don't use loom"
+        "do not use loom"
+        "don't need to use loom"
+        "do not need to use loom"
+        "no need to use loom"
+        "no need for a loom"
+        "no loom"
+        "not use loom"
+        "not via loom"
+        "myself"
+        "inline"
+        "directly"
+    )
+    for phrase in "${ANTI_ROUTE_PHRASES[@]}"; do
+        if [[ "$PROMPT_LOWER" == *"$phrase"* ]]; then
+            log_hook_error "Suppressed route to $MATCHED_AGENT: anti-route phrase '$phrase' matched"
+            MATCHED_AGENT=""
+            MATCHED_DESC=""
+            break
+        fi
+    done
+fi
+
 # No route matched: emit nothing at all (issue #3609). The agent table used to
 # ride along on EVERY prompt; now it only accompanies a genuine route match, so
 # non-matching turns are a silent exit — no additionalContext, no token cost.
--- a/defaults/hooks/skill-router.sh
+++ b/defaults/hooks/skill-router.sh
@@ -170,6 +170,12 @@ done
 # present, using literal (non-regex) substring checks against the full prompt
 # — literal so this guard, like the task-notification guard above, cannot
 # itself false-positive on ordinary human text.
+#
+# INVARIANT: every phrase below must mention "loom" explicitly. Bare
+# single-word phrases ("myself", "inline", "directly") suppress ordinary
+# prompts that never decline Loom at all — e.g. "review this PR directly" or
+# "clean up this code inline" — so they are deliberately excluded. Suppression
+# requires an explicit decline of Loom, never a stylistic adverb.
 if [[ -n "$MATCHED_AGENT" ]]; then
     ANTI_ROUTE_PHRASES=(
         "without loom"
@@ -182,9 +188,6 @@ if [[ -n "$MATCHED_AGENT" ]]; then
         "no loom"
         "not use loom"
         "not via loom"
-        "myself"
-        "inline"
-        "directly"
     )
     for phrase in "${ANTI_ROUTE_PHRASES[@]}"; do
         if [[ "$PROMPT_LOWER" == *"$phrase"* ]]; then
--- a/defaults/hooks/skill-router.sh
+++ b/defaults/hooks/skill-router.sh
@@ -160,6 +160,42 @@ for (( i=0; i<ROUTE_COUNT; i++ )); do
     fi
 done
 
+# =============================================================================
+# ANTI-ROUTE SUPPRESSION (#5327)
+# =============================================================================
+# A prompt can hit a route pattern by keyword coincidence while explicitly
+# declining Loom involvement, e.g. "fix this yourself without loom" or "we
+# don't need to use loom for every tiny change" — both contain "fix" and would
+# otherwise route to /loom:doctor. Suppress the match when a decline phrase is
+# present, using literal (non-regex) substring checks against the full prompt
+# — literal so this guard, like the task-notification guard above, cannot
+# itself false-positive on ordinary human text.
+if [[ -n "$MATCHED_AGENT" ]]; then
+    ANTI_ROUTE_PHRASES=(
+        "without loom"
+        "don't use loom"
+        "do not use loom"
+        "don't need to use loom"
+        "do not need to use loom"
+        "no need to use loom"
+        "no need for a loom"
+        "no loom"
+        "not use loom"
+        "not via loom"
+        "myself"
+        "inline"
+        "directly"
+    )
+    for phrase in "${ANTI_ROUTE_PHRASES[@]}"; do
+        if [[ "$PROMPT_LOWER" == *"$phrase"* ]]; then
+            log_hook_error "Suppressed route to $MATCHED_AGENT: anti-route phrase '$phrase' matched"
+            MATCHED_AGENT=""
+            MATCHED_DESC=""
+            break
+        fi
+    done
+fi
+
 # No route matched: emit nothing at all (issue #3609). The agent table used to
 # ride along on EVERY prompt; now it only accompanies a genuine route match, so
 # non-matching turns are a silent exit — no additionalContext, no token cost.
--- a/defaults/hooks/tests/test-skill-router.sh
+++ b/defaults/hooks/tests/test-skill-router.sh
@@ -156,6 +156,32 @@ reset_markers
 out=$(run_hook "please open an issue with rjwalters/loom about this")
 assert_no_output "report example 'open an issue with rjwalters/loom' -> no route" "$out"
 
+# --- Anti-route suppression: explicit declines near a route keyword (#5327) --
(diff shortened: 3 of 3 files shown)


rjwalters:
Reviewed the diff, tests, and CI locally (in the existing `.loom/worktrees/issue-5327` builder worktree).

**What's solid:**
- Root-cause fix is well-targeted: the `ANTI_ROUTE_PHRASES` block sits after route matching and before `AGENT_ROUTE` construction, gated on `[[ -n "$MATCHED_AGENT" ]]`.
- Fail-open contract preserved — the new `if`/`for`/`[[ ... ]]` tests are all condition contexts, so a false result never trips the top-level `ERR` trap; `bash -n` and `shellcheck` are both clean.
- `.loom/hooks/skill-router.sh` and `defaults/hooks/skill-router.sh` are verified byte-identical (`md5sum` match) after checking out the PR branch directly.
- The four regression prompts from issue #5327's acceptance criteria plus the `fix the failing test in <file>` positive control are all present in `test-skill-router.sh` and pass — ran the full suite locally: 34/34 passed.

**Change requested — three phrases in `ANTI_ROUTE_PHRASES` are overbroad and unnecessary:**

`"myself"`, `"inline"`, and `"directly"` are bare single-word substrings with no requirement that "loom" appear anywhere nearby, so they suppress `AGENT_ROUTE` on plenty of everyday prompts that do **not** decline Loom usage at all. Verified against the PR branch directly:

```
"please review this PR directly, I want your feedback fast"   -> silently suppressed (should route to /loom:judge)
"help me build this feature, I want to do the core logic myself" -> silently suppressed (should route to /loom:builder)
"please clean up this code inline, no separate refactor commit"  -> silently suppressed (should route to /loom:hermit)
"fix the bug directly in the config file"                        -> silently suppressed (should route to /loom:doctor)
```

All four route correctly on `main` (pre-fix) and are genuinely suppressed post-fix — confirmed by diffing hook output before/after.

More importantly, none of these three phrases are actually needed to satisfy the issue's acceptance criteria or the new test cases. I removed `"myself"`, `"inline"`, `"directly"` from the list locally and re-ran all 4 required regression prompts plus the positive control — all still pass, because each already matches one of the more specific `"...loom"` phrases:
- `"no need for a loom agent, just fix it inline"` → already matched by `"no need for a loom"`
- `"fix this yourself without loom"` → already matched by `"without loom"`
- the other two both explicitly contain `"...use loom"` phrases

So the fix is: drop `"myself"`, `"inline"`, `"directly"` from `ANTI_ROUTE_PHRASES` (in both `defaults/` and the synced `.loom/` copy) and add a couple of the false-suppression prompts above as new regression tests to lock in the corrected behavior (asserting they *do* still route). Everything else in the PR — the suppression mechanism itself, its placement, the fail-open behavior, and the required regression coverage — looks correct and can stay as-is.

**CI**: at review time, non-shell-suite checks (Detect Changes, CLAUDE.md Line Budget, AGENTS.md Sync Check, Docs/Defaults Parity Check, Installer Integration Tests, Codex Adapter Smoke) all pass; `Shell Test Suites (hermetic)` was still pending after a 150s wait. `mergeStateStatus` is `UNSTABLE` (pending checks), not yet `CLEAN`. Given the correctness issue above, requesting changes regardless of final CI outcome.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:changes-requested` already present (loom:reviewing removed) — another Judge's verdict raced mine while I was polling CI. Discarding my review; not writing any label.

FWIW my independent read reached the same conclusion on the substantive issue: the `ANTI_ROUTE_PHRASES` entries `myself`/`inline`/`directly` are bare single-word substrings with no proximity requirement to "loom", so they'll silently suppress legitimate route suggestions on unrelated prompts. Agree with the requested fix.

rjwalters:
🩺 Doctor: Judge feedback addressed in `1c0e19e3`.

**The overbreadth bug is fixed.** Removed `"myself"`, `"inline"`, and `"directly"` from `ANTI_ROUTE_PHRASES` in both `defaults/hooks/skill-router.sh` and the byte-identical `.loom/hooks/skill-router.sh`. These were bare single-word substrings with no requirement that "loom" appear nearby, so they silently suppressed `AGENT_ROUTE` on ordinary prompts that never decline Loom at all.

Confirmed your finding: all four of issue #5327's required decline prompts still suppress correctly without them, because each already matches a more specific `"...loom"` phrase that stays in the list (`"no need for a loom"`, `"without loom"`, `"don't use loom"`, `"don't need to use loom"`).

**Invariant documented in the block comment** so this cannot silently regrow: every phrase in `ANTI_ROUTE_PHRASES` must mention "loom" explicitly — suppression requires an explicit decline of Loom, never a stylistic adverb.

**Four new regression tests** in `defaults/hooks/tests/test-skill-router.sh` lock in the corrected behavior, asserting these *do* still route (your exact four prompts):

| Prompt | Asserted route |
|---|---|
| `help me build this feature, I want to do the core logic myself` | `/loom:builder` |
| `please clean up this code inline, no separate refactor commit` | `/loom:hermit` |
| `please review this PR directly, I want your feedback fast` | `/loom:judge` |
| `fix the bug directly in the config file` | `/loom:doctor` |

**Verification:**
- `defaults/hooks/tests/test-skill-router.sh` — **42/42 pass** (was 34/34), including the original 4 decline-prompt tests, the `fix the failing test in <file>` positive control, the 4 new false-suppression tests, and the suite's own `.loom/` ↔ `defaults/` byte-identical check.
- `shellcheck` clean on all three modified files; `bash -n` clean.
- `md5sum` match confirmed between `.loom/hooks/skill-router.sh` and `defaults/hooks/skill-router.sh`.
- Remote CI: **all checks green** — `Shell Test Suites (hermetic)` pass (5m35s), Installer Integration Tests, Codex Adapter Smoke, AGENTS.md Sync, CLAUDE.md Line Budget, Detect Changes, Docs/Defaults Parity all pass. Rust/MCP/Dashboard jobs skip on path filters (no non-shell files touched). `mergeStateStatus` is now `CLEAN`.

Note on the local gate: `.loom/scripts/build-gate.sh` was killed by my own 900s timeout during its cold-worktree `cargo` build — inconclusive, not a failure, and its Rust stages are unrelated to this shell-only change. CI's hermetic shell suite is the authoritative gate here and it passed.

Scope unchanged otherwise: the suppression mechanism, its placement after route matching, the fail-open behavior, and the required regression coverage all stay exactly as you reviewed them. Ready for re-review.

rjwalters:
## Judge re-review — APPROVED

Re-verified Doctor's fix for the prior "changes requested" verdict (overbroad bare-word entries in `ANTI_ROUTE_PHRASES`).

**Verified independently** (worktree `.loom/worktrees/issue-5327`, HEAD `1c0e19e3`):

1. **Diff correctness** — `"myself"`, `"inline"`, `"directly"` are removed from `ANTI_ROUTE_PHRASES` in both `defaults/hooks/skill-router.sh` and `.loom/hooks/skill-router.sh`. The remaining 10 entries (`"without loom"`, `"don't use loom"`, `"do not use loom"`, `"don't need to use loom"`, `"do not need to use loom"`, `"no need to use loom"`, `"no need for a loom"`, `"no loom"`, `"not use loom"`, `"not via loom"`) all mention "loom" explicitly — reasonable and consistent with the new INVARIANT comment documenting that requirement.
2. **Byte-identical copies** — confirmed via `diff` (no output) and `md5sum` (`bdaa54d80e64119fb0871808483e4a9c` for both files).
3. **Test suite** — ran `bash defaults/hooks/tests/test-skill-router.sh` locally: **42/42 passed**, including all 4 new regression tests confirming the previously oversuppressed prompts ("...myself" → `/loom:builder`, "...inline" → `/loom:hermit`, "review...directly" → `/loom:judge`, "fix...directly" → `/loom:doctor`) now correctly route, plus the original 4 anti-route (decline) cases still suppress correctly.
4. **Static checks** — `bash -n` syntax OK, `shellcheck` clean (no warnings).
5. **CI** — `gh pr checks 5332`: all applicable checks pass (Shell Test Suites hermetic, Installer Integration Tests, Codex Adapter Smoke, AGENTS.md Sync, CLAUDE.md Line Budget, Docs/Defaults Parity). Rust/Dashboard/MCP jobs correctly skip — this is a shell-only diff. `mergeStateStatus: CLEAN`.

No new overbreadth introduced — the fix narrows the suppression surface (via the "loom" invariant) without weakening the original decline-detection behavior from #5327's issue examples. Doctor's transparency about the inconclusive local `build-gate.sh` run (cold-worktree cargo timeout, unrelated to this shell-only diff, CI green) is appropriate and doesn't block approval.

Approving.

Issue #5391, opened 2026-08-05
The watchdog detects daemon death and never recovers it — 252 divergences, 1h40m outage


rjwalters:
## Context

Related to #5343 (closed) and #4011 (closed). Both closed; the gap they describe is still open in a different form.

`loom-daemon-watchdog.sh` correctly detected that a daemon was gone and did nothing about it for 1h40m. Every five minutes, verbatim:

```
2026-08-05T04:05:29Z [DIVERGENCE] A daemon is EXPECTED (autonomy-desired marker present,
started 2026-08-03T21:59:57Z) but is NOT running: launchd job
gui/501/com.rjwalters.loom-daemon is LOADED but NOT running (no live pid); and the in-band
probe confirms it: 'loom-daemon quarantine list' could not reach anything at
/Users/rwalters/.loom/loom-daemon.sock (exit 1) ... Autonomous dispatch has stopped.
Recover with: ./.loom/scripts/cli/loom-daemon-start.sh [flags]
```

The diagnosis is excellent — job state, live-pid check, in-band socket probe, and the exact recovery command. It then does not run it. The daemon came back only because a human went looking.

**252 `[DIVERGENCE]` entries in that log going back to 2026-07-28**, so this is the steady state, not one bad night.

## Why "watchdog" is the wrong word for this

A fleet inventory recorded three hosts as protected because a watchdog unit was installed, versus one host with none. That distinction is much smaller than it reads: a watchdog that reports into a logfile nobody tails is closer to no watchdog than to a supervisor.

Supporting evidence that this is not host-local: `loom-worker-2` still has **no watchdog unit on disk at all** —

```
$ systemctl --user is-active loom-daemon-watchdog.timer
inactive
$ systemctl --user is-enabled loom-daemon-watchdog.timer
not-found
$ ls ~/.config/systemd/user/loom-daemon-watchdog.*
(no such file)
```

— even though #5343, which covers exactly that host and that missing timer, is closed. So either the fix does not provision onto already-running hosts, or it regressed. Worth confirming as part of this.

## The decision, not just the bug

Whether the watchdog *should* self-heal is a genuine judgement call, not an obvious yes. Auto-restarting against a genuinely broken binary is its own hazard — a restart loop that burns tokens and masks the real fault. But "detect and print" leaves the same operational gap as no watchdog whenever nobody is reading.

A middle path exists: recover with backoff, cap the attempts, and escalate loudly (not into a logfile) once the cap is hit.

[...]

Issue #4573, opened 2026-07-30
bug(tests): TestDaemon spawns a real daemon with autonomy enabled — the daemon integration suite can dispatch real sweeps and burn API quota


rjwalters:
## Problem

`TestDaemon::start()` (`loom-daemon/tests/common/mod.rs`) spawns a **real**
`loom-daemon` child and sets only three env vars:

```rust
let mut process = Command::new(daemon_bin())
    .env("LOOM_SOCKET_PATH", &socket_path)
    .env("RUST_LOG", "debug")
    // Disable restore_from_tmux() to prevent cross-test-binary contamination
    .env("LOOM_NO_RESTORE", "1")
```

Nothing disables the daemon's **autonomous** behavior, and nothing repoints its
workspace root. The child therefore:

1. **Inherits `LOOM_WORKSPACE` / `LOOM_WORKTREE_ROOT` from the test process**, so it
   resolves the *real* repository rather than its own `TempDir`.
2. **Reads the repo's committed `.loom/config.json`**, which in this repo enables
   autonomy explicitly:

   ```json
   "autonomous": {
     "roleRunner": { "enabled": true, "roles": ["curator", "champion", "judge"] },
     "collisionDetection": { "enabled": true },
     "workFinder": { "maxConcurrent": 8 }
   }
   ```

So `cargo test -p loom-daemon --test integration_basic` (or any of the four
`integration_*` suites, each of which calls `TestDaemon::start()` per test) can
start daemons that poll the forge, claim issues, dispatch sweeps, and spend real
Claude/GitHub quota — from inside a test run.

## Observed consequence

While running the daemon suite on a Loom-installed host, `target/debug/loom-daemon`
binaries built **inside a test worktree** were the parents of live agent sessions:

```
3902034 /home/ubuntu/GitHub/loom/.loom/worktrees/issue-4385/target/debug/loom-daemon
  └─ 1593756 .loom/scripts/claude-wrapper.sh -p /loom:sweep 4385 --claim-owned 4385
       └─ 1594014 claude -p /loom:sweep 4385 --claim-owned 4385
```

(also `4152140` → sweep-registry pid `1629828`). Downstream effects seen in the same
window: **five** concurrent `/loom:sweep 4385` sessions sharing one worktree and
wiping each other's `target/` directories mid-build, and the shared GraphQL rate
limit driven to `0 remaining`.

**Caveat on attribution:** the code-level gap (no autonomy guard in `TestDaemon`) is
[...]

PR #5478, opened 2026-08-05
test: tie internal drain-exit(0) to watchdog kickstart recovery in one integration test

### LINKED ISSUE #5390: auto-update drain exits 0 for a launchd relaunch that never comes (the #4011 failure mode)
## Context

**This is the failure mode of #4011, which is closed.** Filing fresh rather than commenting on a closed issue, but they should be read together.

An operator Mac (`robb-studio`, launchd, primary host of a four-host fleet) had no running daemon for ~1h40m. It did not crash — it exited on purpose:

```
[2026-08-04T19:43:30.156] [INFO] role_runner: curator tick completed for .../klayout-tools in 79.1s
[2026-08-04T19:43:35.941] [WARN] auto_update: rebuilt + provisioned; drain-and-restart triggered
[2026-08-04T19:43:35.941] [WARN] drain complete — 0 in-flight sweeps; exiting 0 for a
                                 launchd-supervised relaunch. No sweep was killed.
```

The drain worked exactly as designed: zero in-flight, no sweep killed, clean `exit 0`. **launchd then did not relaunch it.** The job stayed loaded with no live pid until it was started by hand:

```
$ launchctl list | grep loom
-   -9   com.rjwalters.loom-daemon
-    1   com.rjwalters.loom-daemon-watchdog

$ loom-daemon status
A daemon is EXPECTED (autonomy-desired marker present, started 2026-08-03T21:59:57Z)
but is NOT running: launchd job user/501/com.rjwalters.loom-daemon is not loaded/alive.
Autonomous dispatch has stopped — this is the silent-autonomy-loss scenario (#4011).
```

Recovered with `./.loom/scripts/cli/loom-daemon-start.sh`; running again at cap 16.

## The defect

`exit 0` as a restart mechanism is only correct if the supervisor is configured to restart on *clean* exit. Under launchd that requires `KeepAlive` semantics that actually cover successful exits — `RunAtLoad` alone will not do it, and `KeepAlive: {SuccessfulExit: false}` explicitly will not.

So the auto-update path assumes a relaunch contract the plist may not provide. When that assumption is wrong the daemon self-updates itself into being switched off, and the more reliably auto-update fires, the more often it happens.

Note this is the *good* path failing: a crash would arguably be noticed sooner. A clean drain-and-exit looks like an orderly shutdown to everything watching.

- [ ] The drain-and-restart path either re-launches itself explicitly (e.g. `launchctl kickstart`) rather than trusting the supervisor, or the installer guarantees a plist whose `KeepAlive` restarts on clean exit
- [ ] The two halves are verified together — a test or documented check that a self-update on a launchd host actually comes back
- [ ] If the contract cannot be guaranteed, the daemon should not exit 0 silently; it should log that it is relying on the supervisor and name which key it needs

## Environment

- macOS 26.6, launchd, label `com.rjwalters.loom-daemon`
- `loom-daemon` rebuilt by its own auto-update at the moment of exit
- Watchdog present and detecting, but not recovering — filed separately

---

### Verified corrections

- **2026-08-05, verified against `origin/main` @ `61b40536`**: the issue's core hypothesis — "the auto-update path assumes a relaunch contract the plist may not provide" — does not hold against the current codebase. `render_launchd_plist()` in `defaults/scripts/cli/loom-daemon-start.sh:712` unconditionally emits `KeepAlive: {SuccessfulExit: true}` for every installed plist; there is no code path that installs a plist without it. This was added under #4054 (the "supervised restart primitive").
- **2026-08-05, verified against `origin/main` @ `61b40536`**: a `launchctl kickstart`-based self-heal for the *exact* failure signature this issue describes ("job loaded, not running, last exit status 0") already exists and already has test coverage — it is not a missing feature:
  - `defaults/scripts/cli/loom-daemon-watchdog.sh:956-999` — the watchdog's bounded auto-remediation gate (added for #4232), narrowly scoped to `job_loaded=true && last_exit_status==0`, invokes plain `launchctl kickstart` (never `-k`).
  - `defaults/scripts/cli/loom-daemon-update.sh` (~line 2786) and `loom-daemon/src/restart_verify.rs` implement the same poll-then-kickstart self-heal for the `loom-daemon-update.sh` / `loom-daemon restart` CLI callers.
  - Test coverage: `defaults/scripts/tests/test-loom-daemon-watchdog.sh` (the `exit-0-and-down` case, ~line 475-507) and `defaults/scripts/tests/test-loom-daemon-update.sh` (test 35, ~line 2563-2612, "Restart ack'd but launchd never relaunches -> kickstart fallback DOES relaunch").
- **2026-08-05, verified against `origin/main` @ `61b40536`**: the actual architectural gap is narrower than the issue's framing, and is *already documented in the code itself*. `loom-daemon/src/restart_verify.rs`'s module doc ("Residual: an in-cgroup caller cannot self-heal") explicitly names **the daemon's own autonomous self-update loop** (`loom-daemon/src/auto_update.rs`, the exact log line `"rebuilt + provisioned; drain-and-restart triggered"` this incident shows, at `auto_update.rs:890`) as the one caller that dies as part of the very drain it would need to verify, so it cannot poll-and-kickstart itself in-process. The documented, intended backstop for that specific residual is exactly `loom-daemon-watchdog.sh`'s out-of-band kickstart described above — which is present, gated correctly per the code, and has its own test coverage.
- **2026-08-05**: this incident's actual anomaly is therefore not "no self-heal exists" but "the self-heal that exists and is tested did not fire/succeed for 1h40m (252 watchdog ticks) on this host." That is the exact scope already filed as **#5391** ("The watchdog detects daemon death and never recovers it"), reported from the same incident. #5390 and #5391 are two ends of the same single failure: this issue is the symptom (daemon down, drain-then-exit is exactly as designed), #5391 is the cause (the one thing that should have caught it, per the code above, didn't).

The original three checkboxes are superseded by the verified findings above. Replacing with:

- [ ] **Root-cause why the watchdog's already-existing, already-tested kickstart auto-remediation (`loom-daemon-watchdog.sh:956-999`) did not recover `robb-studio` during this incident** — this is #5391's scope; do not duplicate that investigation here. If #5391's fix closes this gap, this issue can close once verified against the same incident class.
- [ ] **Add an end-to-end integration test that ties the two already-separately-tested halves together**: the *internal* Rust auto-update path (`auto_update.rs`'s rebuild → `ipc.rs::run_drain_supervisor` → `exit(EXIT_RESTART)`) followed by a simulated launchd non-relaunch, followed by a watchdog tick that kickstarts and recovers. Today `test-loom-daemon-update.sh` test 35 and the watchdog's own `exit-0-and-down` case test the kickstart mechanism in isolation from each other and from the internal auto-update trigger — nothing currently exercises the full incident chain in one test. This is the one piece of the original issue's AC #2 that verified evidence does **not** show as already covered.
- [ ] **Confirm `robb-studio`'s actually-installed plist matches current `render_launchd_plist()` output** (i.e., it was re-provisioned after #4054 landed, not left over from a pre-#4054 install) — since the code is verified correct as of `61b40536`, a stale on-host plist predating that fix is the other plausible explanation for this specific incident and should be ruled in or out before concluding the code itself has a gap.

### Related open work

- **#5391** (open): same incident, watchdog-recovery side — see "Verified corrections" above. Do not duplicate its root-cause investigation here.
- **#4011** (closed): the original silent-autonomy-loss report; this issue and #5391 are both instances of the same class recurring in a new shape.

### Affected Files

- `loom-daemon/src/auto_update.rs` — the in-cgroup, cannot-self-heal drain trigger (`record_rebuild()`, ~line 879-903); any structural change (e.g. explicit logging that recovery depends entirely on the watchdog) lands here.
- `loom-daemon/src/restart_verify.rs` — already documents the residual; update its doc comment if the end-to-end test or a fix changes the contract.
- `defaults/scripts/cli/loom-daemon-watchdog.sh` — the actual auto-remediation gate (lines 956-999); primary target of #5391's investigation, referenced here for context only.
- `defaults/scripts/tests/test-loom-daemon-update.sh` and `defaults/scripts/tests/test-loom-daemon-watchdog.sh` — existing isolated kickstart tests; the new end-to-end test (revised AC #2) should live alongside one of these or as a new integration test file.
- `loom-daemon/tests/integration_drain_then_exit.rs` — existing drain/exit integration test; a full-chain test may extend this file rather than adding a new one.

### Test Plan

- [ ] Automated tests: new integration/shell test exercising the full chain — internal auto-update rebuild triggers drain → `exit(EXIT_RESTART)` → simulated launchd failure to relaunch → watchdog tick → kickstart → daemon confirmed back up with a new pid.
- [ ] Manual verification: on a real (or disposable VM) launchd host, force an auto-update roll, then `launchctl bootout` the relaunched job's would-be relaunch window (or otherwise simulate launchd dropping it) and confirm the watchdog's next tick recovers it within its polling interval.
- [ ] Edge cases: verify the existing narrow-gate proof (`exit-143-and-down` in `test-loom-daemon-watchdog.sh`) still holds — an operator-initiated stop or genuine crash must never trigger auto-kickstart.
- [ ] Depends on #5391's findings: once that issue identifies why the watchdog didn't recover in the actual incident, confirm the fix (wherever it lands) closes the gap this issue's end-to-end test would otherwise just document as pre-existing.

rjwalters:
## Summary

Issue #5390's curator enhancement found the original bug hypothesis did not
hold against current `main` (the launchd plist already sets
`KeepAlive:{SuccessfulExit:true}`, and a kickstart-based self-heal for this
exact failure signature already exists and is already tested). Root-causing
why the watchdog didn't recover the actual `robb-studio` incident is #5391's
scope, out of scope here. The one concrete, buildable gap the curator
identified: nothing ties the two already-separately-tested halves of the
incident chain together in a single test. This PR adds that end-to-end test.

## Changes

- Adds `loom-daemon/tests/integration_drain_exit_then_watchdog_recovers.rs`,
  a new integration test that drives the full chain in one run:
  1. Spawns the real `loom-daemon` binary and sends it the exact request
     shape `auto_update::IpcDrainTrigger::trigger()` sends internally
     (`DrainAndRestartDaemon`, `then_exit: false`) — exercising
     `handle_drain_request` -> `run_drain_supervisor` ->
     `std::process::exit(EXIT_RESTART)` against the real binary.
  2. Asserts the process exits `0` (`EXIT_RESTART`), then confirms nothing
     re-listens on the socket — the simulated "supervisor never relaunches
     it" failure mode.
  3. Writes an autonomy-desired marker + a stubbed `systemctl` on `PATH`
     reporting the exact `#4862` signature (`LoadState=loaded`,
     `ExecMainCode=exited`, `ExecMainStatus=0`) and runs the real
     `defaults/scripts/cli/loom-daemon-watchdog.sh` against it.
  4. The stub's `start` action is the one thing that is real recovery: it
     spawns a genuine second instance of the same `loom-daemon` binary bound
     to the same socket.
  5. Asserts the watchdog exits `0`, logs the remediation, and — independent
     of the watchdog's own exit code — that the socket now answers `Ping`
     again from a **new** pid (not the original one lingering).

## Why systemd, not launchd

This repo's CI runs on `ubuntu-latest`. The watchdog script's launchd
auto-remediation gate is unconditionally disabled off-Darwin
(`[[ "$(uname -s)" == "Darwin" ]] || USE_LAUNCHD=false`), which is why every
launchd-specific case in `test-loom-daemon-watchdog.sh` is wrapped in its own
Darwin check and skips cleanly elsewhere. The `#4862` systemd mirror carries
no such platform clobber, so it is the one auto-remediation path this test
can actually execute in CI. The Rust-side drain/exit mechanics under test
(`EXIT_RESTART = 0`) are supervisor-agnostic, so pinning the systemd mirror
exercises the identical end-to-end contract the launchd incident describes.
`systemctl` is stubbed throughout (never invoked for real), matching the
existing technique in `test-loom-daemon-watchdog.sh` test 10a and
`test-loom-daemon-update.sh` test 35.

Per the curator's revised acceptance criteria on #5390:

- [N/A — explicitly out of scope, deferred to #5391] Root-cause why the
  watchdog's existing kickstart didn't recover the actual `robb-studio`
  incident.
- [x] **End-to-end integration test tying the two already-separately-tested
  halves together** — this PR's new test file, described above.
- [N/A — not buildable, operator/manual verification] Confirm a specific
  host's installed plist matches current `render_launchd_plist()` output.

## Test Plan

- `cargo test --package loom-daemon --test integration_drain_exit_then_watchdog_recovers` — new test passes, run repeatedly (3x) with no flakes observed; verified no leaked processes after each run (`ps aux` clean of worktree-scoped `loom-daemon` pids).
- `cargo test --package loom-daemon --test integration_drain_then_exit --test integration_drain_exit_then_watchdog_recovers` — both drain-path integration tests pass together (shared `#[serial]` group, no interference).
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 114/114 passed (unchanged).
- `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 254/254 passed (unchanged).
- `cargo clippy --package loom-daemon --tests --all-features -- -D warnings` — clean.
- `cargo fmt --check -p loom-daemon` — clean (formatter applied once during development).

Closes #5390

### DIFF
--- a/loom-daemon/tests/integration_drain_exit_then_watchdog_recovers.rs
+++ b/loom-daemon/tests/integration_drain_exit_then_watchdog_recovers.rs
@@ -1,0 +1,402 @@
+// End-to-end integration test for the exact incident chain in Issue #5390
+// ("auto-update drain exits 0 for a launchd relaunch that never comes").
+//
+// #5390's curator enhancement verified that the two halves of this incident
+// are each ALREADY tested in isolation:
+//   - the internal auto-update path's `exit(EXIT_RESTART)` contract is pinned
+//     by `integration_drain_then_exit.rs` (the `then_exit: true` sibling of
+//     the request this test issues with `then_exit: false`), and
+//   - the watchdog's bounded kickstart self-heal for "job loaded, not
+//     running, last exit status 0" is pinned by
+//     `test-loom-daemon-watchdog.sh`'s `exit-0-and-down` case and
+//     `test-loom-daemon-update.sh` test 35.
+// Nothing tied them together end-to-end: a real daemon actually taking the
+// internal drain-then-`exit(0)` path, a supervisor that (like the real
+// `robb-studio` incident) never relaunches it on its own, and a watchdog tick
+// that observes exactly that reality and recovers it with a genuinely NEW
+// process. That chain is what this test drives.
+//
+// WHY SYSTEMD, NOT LAUNCHD: this suite runs on Linux CI (`ubuntu-latest`).
+// The watchdog script's launchd auto-remediation gate is guarded by
+// `[[ "$(uname -s)" == "Darwin" ]] || USE_LAUNCHD=false` — on any non-macOS
+// host it is unconditionally disabled, which is why every launchd-specific
+// case in `test-loom-daemon-watchdog.sh` is wrapped in its own Darwin check
+// and skips cleanly elsewhere. The systemd mirror added for #4862 carries NO
+// such platform clobber (deliberately — see the comment above `USE_SYSTEMD`
+// in the script), so it is the one auto-remediation path this test can
+// actually execute in this repository's CI. The Rust-side drain/exit
+// mechanics under test are supervisor-agnostic (`EXIT_RESTART` is the same
+// `0` either way), so pinning the systemd mirror exercises the identical
+// end-to-end contract the launchd incident describes.
+//
+// `systemctl` itself is never invoked for real — a stub is put first on
+// PATH, exactly the technique `test-loom-daemon-watchdog.sh` test 10a and
+// `test-loom-daemon-update.sh` test 35 already use, so this test cannot touch
+// a real systemd --user manager on the host running it.
+//
+// expect/unwrap are acceptable here since tests should panic on failure.
+#![allow(clippy::expect_used)]
+#![allow(clippy::unwrap_used)]
+
+mod common;
+
+use common::{daemon_bin, isolate_daemon_state, TestClient};
+use loom_daemon::ipc::EXIT_RESTART;
+use serial_test::serial;
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
+use std::process::{Child, Command, Stdio};
+use std::time::{Duration, Instant};
+
+/// Kills the spawned daemon if the test panics before it exits on its own.
+struct ChildGuard(Child);
+
+impl Drop for ChildGuard {
+    fn drop(&mut self) {
+        let _ = self.0.kill();
+        let _ = self.0.wait();
+    }
+}
+
+/// Kills a bare pid (the second daemon instance, spawned indirectly by the
+/// stubbed `systemctl start` — there is no `Child` handle for it, only the
+/// pid it wrote to the state file) if the test panics before cleanup.
+struct PidGuard(u32);
+
+impl Drop for PidGuard {
+    fn drop(&mut self) {
+        let _ = Command::new("kill")
+            .args(["-9", &self.0.to_string()])
+            .status();
+    }
+}
+
+/// Liveness bound for the drain-then-exit round trip. Mirrors
+/// `integration_drain_then_exit.rs`'s `DRAIN_EXIT_WAIT`.
+const DRAIN_EXIT_WAIT: Duration = Duration::from_secs(60);
+
+/// Bound on the second daemon instance (spawned by the stubbed
+/// `systemctl start`) creating its socket.
+const RESPAWN_SOCKET_WAIT: Duration = Duration::from_secs(30);
+
+/// The full incident chain: internal auto-update-shaped drain request ->
+/// `exit(EXIT_RESTART)` -> simulated supervisor non-relaunch -> a watchdog
+/// tick that observes "unit loaded, not running, last exit 0" and kickstarts
+/// a real recovery.
+#[tokio::test]
+#[serial]
+async fn test_drain_exit0_then_watchdog_systemd_kickstart_recovers() {
+    let temp_dir = tempfile::TempDir::new().expect("temp dir");
+    let socket_path = temp_dir.path().join("daemon.sock");
+    let workspace_root = temp_dir.path().join("workspace");
+    std::fs::create_dir_all(&workspace_root).expect("create workspace root");
+    let worktree_root = temp_dir.path().join("worktrees");
+
+    // A unit name unique to this test run — never resolved against any real
+    // systemd manager, since `systemctl` itself is stubbed below.
+    let unit_name = format!("loom-daemon-test-5390-{}.service", std::process::id());
+
+    // ------------------------------------------------------------------
+    // Phase 1: the internal auto-update drain path.
+    //
+    // `auto_update::IpcDrainTrigger::trigger()` calls
+    // `ipc::handle_drain_request(..., then_exit: false)` — this is that exact
+    // request shape, issued directly over the real IPC socket against a real
+    // daemon binary, so the assertion below exercises the same
+    // `handle_drain_request` -> `run_drain_supervisor` ->
+    // `std::process::exit(EXIT_RESTART)` path the auto-update loop's rebuild
+    // completion drives in production.
+    // ------------------------------------------------------------------
+    let mut cmd = Command::new(daemon_bin());
+    isolate_daemon_state(&mut cmd, temp_dir.path());
+    let mut child = ChildGuard(
+        cmd.current_dir(&workspace_root)
+            .env("LOOM_SOCKET_PATH", &socket_path)
+            .env("LOOM_WORKSPACE", &workspace_root)
+            .env("LOOM_ROLE_RUNNER", "0")
+            .env("LOOM_WORK_FINDER", "0")
+            .env("LOOM_EPIC_SUPERVISOR", "0")
+            .env("LOOM_WORKTREE_ROOT", &worktree_root)
+            .env("RUST_LOG", "debug")
+            .env("LOOM_NO_RESTORE", "1")
+            // The daemon's own record of how it is supervised — read by
+            // `detect_supervisor()`, which `handle_drain_request` requires to
+            // be `Some` before it will accept a `then_exit: false` (relaunch)
+            // drain at all. `systemd` (not `launchd`) so the watchdog phase
+            // below stays on the one auto-remediation path this suite can
+            // actually execute on Linux CI (see the module doc comment).
+            .env("LOOM_DAEMON_SUPERVISOR", "systemd")
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .expect("spawn daemon"),
+    );
+
+    let start = Instant::now();
+    while !socket_path.exists() {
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM — approved.

Verified the Builder's claims directly rather than taking them on faith:

**Code review**
- The new test (`loom-daemon/tests/integration_drain_exit_then_watchdog_recovers.rs`) genuinely drives both real code paths, not a hollow simulation:
  - Phase 1 spawns the actual `loom-daemon` binary and sends the literal `DrainAndRestartDaemon`/`then_exit: false` IPC shape `auto_update::IpcDrainTrigger::trigger()` sends in production, asserting the real `handle_drain_request` -> `run_drain_supervisor` -> `exit(EXIT_RESTART)` path and confirming no re-listen (checked `EXIT_RESTART`, `detect_supervisor`, `handle_drain_request` in `loom-daemon/src/ipc.rs` — all real).
  - Phase 3 stubs `systemctl` with the exact same technique already used by `test-loom-daemon-watchdog.sh` test 10a (`show`/`LoadState`/`ExecMainCode`/`ExecMainStatus` signature, `start` action), then runs the real `defaults/scripts/cli/loom-daemon-watchdog.sh` — every env var it sets (`LOOM_WATCHDOG_IPC_PROBE`, `LOOM_WATCHDOG_RECOVERY_STATE`, `LOOM_WATCHDOG_KICKSTART_RECHECK_ATTEMPTS`, etc.) exists and is honored in the actual script.
  - The stubbed `start` action spawns a genuine second `loom-daemon` process (inherits `LOOM_SOCKET_PATH` + all the `#4556`/`#4573` isolation env from the watchdog's own `Command`, so it binds the same socket in the same confined fixture) — Phase 4 asserts a **new** pid and a live `Ping` on the socket, not just the watchdog's own exit code.
- Diff is scoped exactly as described: one new file, 402 additions, 0 deletions — nothing else touched.
- PR body correctly closes #5390 and documents the out-of-scope items (#5391 root cause, plist-drift verification) per the curator's revised AC.

**Reproduced locally (worktree `.loom/worktrees/issue-5390`)**
- `cargo test -p loom-daemon --test integration_drain_exit_then_watchdog_recovers`: pass, 3x in a row, ~8-10s each, no flakiness.
- Ran alongside `integration_drain_then_exit`: both pass together, no interference.
- No leaked processes: checked `ps aux` immediately after and ~30s later — the only lingering short-lived process was the daemon's own internal `tokens check --ranking` self-check subprocess (unrelated to this test, present in any daemon spawn, and confirmed reaped within seconds; unrelated real-daemon activity on the shared host was excluded by binary path).
- `test-loom-daemon-watchdog.sh`: 114/114.
- `test-loom-daemon-update.sh`: first pass showed 253/254 (one transient failure on this heavily-loaded shared host), a clean rerun immediately after was 254/254 — this script isn't touched by the diff at all, so this reads as host-load flake, not a regression.
- `cargo clippy --package loom-daemon --tests --all-features -- -D warnings`: clean.
- `cargo fmt --check -p loom-daemon`: clean.

**CI / merge state**: all required checks pass (`gh pr checks 5478`), `mergeStateStatus` is `CLEAN`.

Nice, focused test that actually closes the gap the curator identified — well-reasoned choice of systemd over launchd for CI portability, and the doc comments make the "why" traceable back to #5390/#4862/#5391.

PR #5521, opened 2026-08-06
fix(guard): stop misreading arithmetic/test >, >=, <, <= as redirection

### LINKED ISSUE #5515: Guard: extract_write_targets() misreads bash arithmetic >/>= comparisons as redirection, manufacturing phantom write targets
## What I Observed

While reviewing `.loom/logs/guard-decisions.log` per the Auditor's standing guard-decision-telemetry policy, I found several `worktree-write-confinement` (catastrophic-tier) DENY entries where the logged command has **no actual write** — no `cp`/`mv`/`tee`/`sed -i`/output redirection to any real file. Two representative examples:

**Example A** (`ts: 2026-08-04T05:19:33Z`):
```bash
CLAIM_TIME="2026-08-04T04:55:33Z"
CLAIM_EPOCH=$(date -u -d "$CLAIM_TIME" +%s ...)
STALE_EPOCH=$((CLAIM_EPOCH + 30*60))
for i in $(seq 1 9); do
  NOW_EPOCH=$(date -u +%s)
  LABELS=$(gh api "repos/rjwalters/loom/pulls/5227" --jq '...')
  ...
  if (( NOW_EPOCH >= STALE_EPOCH )); then
    echo "RESULT: claim is now stale (>= 30 min), safe to reclaim"
    break
  fi
  ...
done
```
Denied as `worktree-write-confinement` despite containing zero file writes — pure polling/echo/sleep.

**Example B** (`ts: 2026-08-06T05:18:17Z`):
```bash
git fetch origin --quiet
AFFECTED_FILES=(...)
MISSING=()
for path in "${AFFECTED_FILES[@]}"; do
  if ! git ls-tree -r origin/main --name-only | grep -qFx "$path"; then
    MISSING+=("$path")
  fi
done
if (( ${#MISSING[@]} > 0 )); then
  echo "MISSING: ${MISSING[@]}"
else
  echo "All affected files present on origin/main"
fi
```
Also denied as `worktree-write-confinement` — again zero file writes, just `git fetch`/`git ls-tree`/`echo`.

## Root Cause (verified by reading the guard source)

`.loom/hooks/guard-destructive-generic.sh`'s `extract_write_targets()` (redirection scan around lines 3696-3721) detects `>`/`>>` redirection by matching quote-masked tokens against:

```awk
if (mt ~ /^[0-9]*>>?$/) {          # bare operator, e.g. `echo x > f`
    if (j + 1 <= m && toks[j+1] != "" && mtoks[j+1] !~ /^&/) {
        print curcwd SEP resolve_var(toks[j+1])
    }
    continue
}
if (mt ~ /^[0-9]*>>?[^ \t&]/) {    # attached form, e.g. `echo x >f`
    op = toks[j]
    sub(/^[0-9]*>>?/, "", op)
    if (op != "") print curcwd SEP resolve_var(op)
}
```

Neither branch excludes an **unquoted `>`/`>=` used as an arithmetic or comparison operator** inside `(( ... ))` / `[[ ... ]]` — a very common bash idiom, and one `mask_gt()` (the existing #4245 quote-aware masking) does not cover, since `mask_gt()` only masks a `>` that is *inside quotes*; an unquoted `>` in `(( x > y ))` is, correctly, left unmasked as "real" — but it isn't a redirection at all.

Concretely:
- `(( ${#MISSING[@]} > 0 ))` → token `>` matches the bare-operator branch exactly (`/^[0-9]*>>?$/`), and the NEXT token `0` gets treated as the write target: `print curcwd SEP "0"` — i.e. a phantom write to `<main-checkout>/0`.
- `(( NOW_EPOCH >= STALE_EPOCH ))` → token `>=` does NOT match the bare-operator branch (trailing `=` breaks the `$` anchor), but DOES match the attached-form branch (`/^[0-9]*>>?[^ \t&]/`, since `=` satisfies `[^ \t&]`). `sub(/^[0-9]*>>?/, "", op)` on the literal token `>=` strips the leading `>`, leaving `op = "="` — a phantom write to `<main-checkout>/=`.

Both phantom targets resolve inside the main repository checkout (since `curcwd` is the main checkout when running from there), so the confinement check fires and denies a command that writes nothing.

## Impact

This false-positives on any multi-line Bash-tool command run from the main checkout that contains an unquoted `(( x > y ))` or `(( x >= y ))` / `[[ x > y ]]` comparison — a routine idiom in staleness-check loops, count guards, timestamp comparisons, etc. (10+ occurrences of `worktree-write-confinement` denies in the last ~2 days' `.loom/logs/guard-decisions.log` contain `(( ... >` or `>=` shapes). Each one blocks legitimate read-only diagnostic/polling work with no real write, forcing a rewrite/retry.

## Recommended Fix

Extend the existing quote-aware masking approach (`mask_gt()`, #4245) with an **arithmetic/test-context-aware** mask: walk the buffer tracking `((...))`/`[[...]]` span depth (mirroring how `mask_gt()`/`mask_ws()` already track quote state) and mask `>`/`>=`/`<`/`<=` bytes found inside such a span the same way quoted `>` bytes are masked today. Alternatively (narrower, lower-risk fix): in the bare-operator branch, additionally require that the token immediately preceding `>` is not itself inside an open `((`/`[[` — or simplest of all, special-case the `>=` token shape explicitly (it is never a valid redirection operator in POSIX/bash) so the attached-form branch cannot manufacture a target of `=`.

## Reproduction

```bash
echo 'foo() { local x=5 y=3; if (( x > 0 )); then echo hi; fi; }; foo' 
# Run the equivalent through the Bash-tool PreToolUse hook from a repo main
# checkout that has a Loom-managed worktree elsewhere — extract_write_targets
# reports a phantom write target of "0" inside the main checkout.
```

## Suggested Outcome

**Refine** — this is a false-positive in write-target detection, not a case where the underlying operation is actually dangerous. The genuinely-dangerous `worktree-write-confinement` cases (real `cp`/`tee`/`sed -i`/redirection writing into the main checkout) must keep denying; only the phantom-target-from-arithmetic-comparison shape needs fixing.

---
Found during a routine Auditor guard-decision-telemetry review (CLAUDE.md § "Guard-Decision Telemetry Review (Standing Policy, #3898)"). Main branch build/test/runtime validation was otherwise clean (CI green on HEAD `dc50695c`, `loom-daemon --version`/`--help`/`calibrate`/`health`/`validate` all ran correctly, `mcp-loom/dist/index.js` starts).

rjwalters:
## Summary

`.loom/hooks/guard-destructive-generic.sh`'s `extract_write_targets()` misparsed
bash arithmetic/comparison operators (`>`, `>=`, `<`, `<=` inside `(( ... ))` or
`[[ ... ]]`) as shell redirection operators, manufacturing phantom write
targets:

- `(( ${#MISSING[@]} > 0 ))` -> token `>` matched the bare-operator branch, so
  the *next* token (`0`) was read as a write target -> phantom target `0`.
- `(( NOW_EPOCH >= STALE_EPOCH ))` -> token `>=` matched the attached-form
  branch; stripping its leading `>` left the literal `=` -> phantom target `=`.

Both resolve inside `curcwd` (the main checkout when the hook runs from
there), so a `worktree-write-confinement` DENY fired against commands that
wrote nothing to the filesystem — a routine bash idiom in staleness-check/
polling loops, count guards, and timestamp comparisons.

## Fix

Extends the existing quote-aware `mask_gt()` masking (#4245) with
arithmetic/test-context awareness: it now also tracks `((...))`/`[[...]]`
span depth the same way it tracks quote state, and masks an unquoted
`>`/`<` byte found inside such a span exactly like a quoted one. A `>`
outside any such span — including one following a *closed* arithmetic span
on the same line (e.g. `echo $(( x > 0 )) > file`) — is untouched and still
flows through as a real redirection operator, so the genuinely dangerous
cases (`cp`/`mv`/`tee`/`sed -i`/real redirection into the main checkout)
keep denying.

Also switches the stdin-redirection-exclusion loop (`<` handling) from the
raw `toks[]` to the masked `mtoks[]`, so an arithmetic/test `<`/`<=` no
longer mismarks itself (and the following token) as "read from stdin"
either.

## Test plan

- [x] `bash tests/hooks/test-guard-destructive.sh` — 938/938 pass, including
  10 new regression cases covering both phantom-target shapes from the
  issue (`(( x > 0 ))`, `(( x >= y ))`, `[[ "$a" > "$b" ]]`, `$(( x > 0 ))`)
  and confirming real redirections/tee/into-main-checkout writes still deny
  even alongside an unrelated arithmetic comparison in the same command.
- [x] `bash tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 pass.
- [x] `bash -n defaults/hooks/guard-destructive-generic.sh` — syntax OK.
- [x] `shellcheck -S warning` — no new warnings vs. `main`.

Closes #5515

### DIFF
--- a/defaults/hooks/guard-destructive-generic.sh
+++ b/defaults/hooks/guard-destructive-generic.sh
@@ -1476,7 +1476,7 @@ function strip_cd_quoting(tok,   out, n, i, c, in_s, in_d, sq, dq) {
 '
 
 # =============================================================================
-# QUOTE-AWARE REDIRECTION MASKING (#4245)
+# QUOTE- AND ARITHMETIC/TEST-CONTEXT-AWARE REDIRECTION MASKING (#4245, #5515)
 #
 # extract_write_targets() (below) recognizes `>`/`>>` redirection by splitting
 # a qsplit()-segmented command on whitespace and pattern-matching each token —
@@ -1497,6 +1497,37 @@ function strip_cd_quoting(tok,   out, n, i, c, in_s, in_d, sq, dq) {
 # DECIDE whether a token is a real (unquoted) redirection operator; the
 # ORIGINAL tokens are still used to extract the actual target text.
 #
+# ARITHMETIC/TEST CONTEXT (#5515): an unquoted `>`/`>=`/`<`/`<=` is ALSO not a
+# redirection operator when it is a comparison inside a `(( ... ))` arithmetic
+# command/expansion or a `[[ ... ]]` conditional expression — a routine bash
+# idiom (`if (( x > 0 )); then`, `(( ${#ARR[@]} > 0 ))`, `[[ "$a" > "$b" ]]`).
+# Before #5515, mask_gt() left those bytes unmasked (they are unquoted), so
+# extract_write_targets()'s bare-operator branch read the token immediately
+# after a bare `>` as a write target (`(( x > 0 ))` -> phantom target `0`), and
+# its attached-form branch matched the `>=` token shape (never a valid
+# redirection operator in POSIX/bash) and stripped its leading `>`, leaving a
+# phantom target of literal `=` (`(( x >= y ))` -> phantom target `=`). Both
+# resolve inside curcwd and can trigger a worktree-write-confinement DENY on a
+# command that writes nothing.
+#
+# mask_gt() now ALSO tracks `((`/`))` and `[[`/`]]` span depth (adepth/tdepth
+# below) the same way it tracks quote state, and masks `>`/`<` bytes found
+# while depth > 0 exactly like a quoted `>` — same MASK byte, same
+# byte-for-byte-length-identical contract. A span opens only on the LITERAL
+# adjacent two-character sequence `((`/`[[` (consumed as one step, so a
+# subsequent stray third paren/bracket is not itself misread as another open)
+# and closes on the literal adjacent `))`/`]]` while its own depth is > 0 —
+# nested single parens inside an arithmetic span (`(( (a) > 0 ))`) do not
+# perturb depth, matching the common case this fix targets. A `>` OUTSIDE any
+# such span — including one following a closed arithmetic span on the SAME
+# segment, e.g. `echo $(( x > 0 )) > file` — is untouched and still flows
+# through as a real operator, so the genuinely dangerous cases (`cp`/`mv`/
+# `tee`/`sed -i`/actual redirection writing into the main checkout) keep
+# denying. This span tracking only ever runs in unquoted mode (mode == 0
+# below): a literal `((`/`[[` appearing as DATA inside a quoted string is
+# never treated as entering a span, matching how such a string's `>` is
+# already unconditionally masked as quoted data regardless of context.
+#
 # Deliberately does NOT model backslash-escaped quotes -- same simplification
 # qsplit() (above) and strip_literal_text() already accept for this file's
 # other quote-tracking scans, and for good reason beyond just consistency: the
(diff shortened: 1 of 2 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-06T06:21:11Z) — standing down without reclaiming. Not stomping.

rjwalters:
LGTM — approving.

**Correctness of the fix**: `mask_gt()` now tracks `((...))`/`[[...]]` span depth (unquoted mode only) the same way it already tracks quote state, masking an unquoted `>`/`<` found inside such a span exactly like a quoted one. Traced through both phantom-target repros from #5515:
- `(( ${#MISSING[@]} > 0 ))` — `>` is now masked, so the bare-operator branch in `extract_write_targets()` no longer reads the following token (`0`) as a write target.
- `(( NOW_EPOCH >= STALE_EPOCH ))` — `>=`'s leading `>` is masked, so the attached-form branch no longer strips it into a phantom `=` target.

Both `extract_write_targets()` branches that manufactured the phantom targets already read from `mtoks[]` (mask_gt() output, #4245) rather than raw `toks[]`, so extending `mask_gt()` alone fixes both shapes without touching the redirection-detection branches themselves — a minimal, well-targeted diff. The stdin-redirect-exclusion loop was also correctly switched from `toks[]` to `mtoks[]` so an arithmetic/test `<`/`<=` doesn't mismark itself (or the following token) as a stdin redirect either.

**Does not weaken genuine-danger detection**: verified this holds for the real cases —
- `>` outside any `((`/`[[` span, including one right after a *closed* span on the same line (`echo $(( x > 0 )) > file`), is untouched.
- `cp`/`mv`/`tee`/`sed -i`/redirection into the main checkout still deny even alongside an unrelated arithmetic comparison in the same command — confirmed by the new `tee` regression test and matches the existing `cp`/`mv`/`sed -i` branches (unaffected — they don't route through `mask_gt()`'s `>`/`<` masking at all).
- Span open/close only triggers on the literal adjacent two-char `((`/`))`/`[[`/`]]` sequence, and only in unquoted mode — quoted `((` text (already handled pre-#5515) is untouched, and single parens inside an arithmetic span (`(( (a) > 0 ))`) don't perturb depth.
- An unbalanced `((`/`[[` (never closed) does leave depth "stuck open" for the rest of the buffer per the file's own documented fallback, but — same as the pre-existing unterminated-quote precedent this mirrors — any bash command with a genuinely unbalanced `((`/`[[` is itself a syntax error and would never execute, so this can't actually let a real write slip through undetected in practice. Comments (which could contain unbalanced text) are already stripped before this scan runs (`COMMAND_ASK_SCAN="$COMMAND_NO_COMMENT"`).

**Test coverage**: 10 new cases are well-chosen — both exact issue repros, `<`/`<=` variants, `[[ ]]` string comparison, `$(( ))` arithmetic expansion, and (importantly) three "narrows, never widens" regression cases confirming a real redirection/tee write into the main checkout still denies even alongside a closed arithmetic span in the same or an adjacent segment.

**Verification performed**:
- `bash -n defaults/hooks/guard-destructive-generic.sh` — syntax OK.
- `bash tests/hooks/test-guard-destructive.sh` — 938/938 pass.
- `bash tests/hooks/test-guard-destructive-dispatcher.sh` — 12/12 pass.
- `shellcheck -S warning` — the two SC1078 warnings present are pre-existing on `main` (verified via diff against `main`'s copy of the file), not introduced by this PR.
- CI: all applicable checks pass (Shell Test Suites, Installer Integration Tests, doc/parity checks); `mergeStateStatus` is `CLEAN`, `mergeable: MERGEABLE`.

Nice, minimal, well-documented fix. Approving.

Issue #5501, opened 2026-08-05
live_state_sandbox guards state paths but not launchd/systemd labels — a sandboxed test can stop the real daemon


rjwalters:
## Context

`live_state_sandbox_init` guards **state paths** but not **supervisor identity**, so a test can be fully sandboxed on disk and still reach out and stop the operator's real daemon.

Hit while verifying the fix for #5131 on a live fleet host. The verification harness set:

```
LOOM_SOCKET_PATH   -> sandbox    (so $LOOM_DIR, heartbeat, logs follow it)
LOOM_PID_FILE      -> sandbox
LOOM_AUTONOMY_MARKER -> sandbox  (for the cases that needed it)
LOOM_LAUNCHD_LABEL -> com.rjwalters.loom-daemon   <-- the REAL label
```

That last value was deliberate — the case under test was *"default label = the operator's real stop"*, to prove the fix did not change normal behaviour. But `loom-daemon-stop.sh` calls `launchd_bootout_if_loaded`, which keys on `LAUNCHD_LABEL`, not on any sandboxed path. So the "sandboxed" run booted the production daemon:

```
[INFO] Received SIGTERM, cleaning up...
[INFO] Socket cleaned up, exiting 143 (operator stop — no supervised relaunch)
```

The host then sat daemonless until noticed — the #4011 silent-autonomy-loss shape, self-inflicted by a test.

## Why the existing guard did not catch it

`live_state_sandbox_assert_untouched` verifies that no live `.loom` **path** was written. It passed, correctly: nothing wrote to the real `~/.loom`. The damage was done through `launchctl bootout` against a **label**, which is not a path and is not part of the snapshot.

The suites already recognise this axis — `test-loom-daemon-update.sh` exports `LOOM_WATCHDOG_LABEL="${LOOM_LAUNCHD_LABEL}-watchdog"` with the comment *"Supervisor identity, not a state path — the watchdog LaunchAgent must be provisioned under the scratch label so a restart path can never touch the real com.rjwalters.loom-daemon-watchdog job (#4011/#4078)."*

So the concept exists and is applied to the watchdog label. It is just not **asserted** — nothing fails a run that uses the real daemon label.

## The general shape

This is the same asymmetry #5131 is about, one layer up: **scoping the state does not scope the supervisor.** #5131 was "the kill is label-scoped but the teardown is host-global"; this is "the paths are sandboxed but the label is not."

[...]

PR #5531, opened 2026-08-06
fix(guide): stop dropping out-of-order-merged PRs from WORK_LOG.md

### LINKED ISSUE #5516: Guide WORK_LOG.md watermark misses out-of-order-merged PRs (number > last_pr assumes merge order == number order)
## What I Observed

While running the Guide's document-maintenance phase (2026-08-06, ~05:41 UTC), the `update_work_log()` algorithm in `.loom/roles/guide.md` missed a merged PR that should have been recorded.

**Watermark computation** (`work_log_max_pr()`): scans `WORK_LOG.md` for the highest `PR #N` already recorded — at the time of this pass, `5509`.

**New-PR filter**: `select(.number > $last_pr)` — i.e. "only PRs with a number greater than the watermark are new."

**What actually happened**: PR #5507 (`fix: guard test sandbox supervisor identity, not just state paths`) merged at `2026-08-06T05:30:06Z` — *after* PR #5509 merged at `2026-08-06T05:01:50Z` — despite having a **lower** PR number. Two PRs can be opened in one order and merged in the reverse order (e.g. #5507 sat longer in Doctor/review while #5509 landed quickly). Because `update_work_log()` filters purely on `number > last_pr`, PR #5507 was silently excluded from every future pass forever — the watermark (`5509`) is already past its number, so it can never satisfy `> $last_pr` no matter how many ticks run.

This was only caught because the Guide happened to cross-reference the closing issue (`#5501`, correctly captured since `5501 > 5499` on the issue watermark) and noticed its closing PR wasn't in the log — a manual check the automated algorithm doesn't perform.

## Root Cause

PR/issue numbers are monotonically increasing at *creation* time, not at *merge* time. `update_work_log()`'s filter assumes number order approximates merge order, which holds most of the time but breaks whenever a lower-numbered PR sits in review/Doctor longer than a higher-numbered one that merges first.

## Impact

Silent, permanent data loss in WORK_LOG.md: an out-of-order-merged PR is never recorded, with no error or warning — the next tick's `select(.number > $last_pr)` just doesn't match it, ever, since the watermark has already advanced past it.

## Suggested Fix

`update_work_log()` in `.loom/roles/guide.md` (and the corresponding logic guiding the docs-worktree WORK_LOG.md update) should stop relying solely on `number > last_pr`. Options, roughly in order of robustness:

1. **Presence check instead of/in addition to number filter**: for each merged PR in the query window (e.g. last 50, or last N days), check whether `PR #<N>` already appears literally in `WORK_LOG.md` before excluding it — not just whether its number is below the watermark. This is what the guide agent did manually this pass.
2. **Timestamp-based watermark**: track the latest recorded PR's `mergedAt` instead of (or alongside) its number, and query `gh pr list --state merged --json number,mergedAt | select(.mergedAt > $last_merged_at)`.
3. **Reconcile via `closingIssuesReferences`**: since issue numbers/closures are already watermarked correctly (issues close in number order far more reliably than PRs merge in number order), cross-check each newly-recorded closed issue's `closedByPullRequestsReferences` / `closingIssuesReferences` against what's already in WORK_LOG.md and backfill any missing PR entries.

Option 1 (presence check) is the simplest fix with no behavior change to the watermark computation itself — just widen the merged-PR query window and add a `grep -q "PR #$n"` guard before excluding candidates by number.

## Reproduction

Any repo where PR A opens before PR B (so `A.number < B.number`) but merges after B (`A.mergedAt > B.mergedAt`), once B has already been recorded by a WORK_LOG.md-writing pass, will permanently drop A. This occurred naturally with #5507 (opened as `feature/issue-5501`, sat in Doctor/review) vs #5509 (opened as `feature/issue-5499`, merged same-day but earlier).

## Suggested Outcome

**Fix** — this is a real, reproducible data-loss bug in a documented, standing automation (the Guide's document-maintenance phase, `.loom/roles/guide.md` "Step 2: Update WORK_LOG.md"). Low risk: the fix only changes which PRs are *included* as candidates before the existing dedup/render logic runs.

---
Found during a routine Guide document-maintenance pass while manually reconciling WORK_LOG.md against `gh pr list`/`gh issue list`.

---

## Curator Enhancement

`.loom/roles/guide.md` is a symlink (`.loom/roles/guide.md` → `../.claude/commands/loom/guide.md`) resolving to the canonical tracked file `defaults/.claude/commands/loom/guide.md` — verified present on `origin/main` as of `2026-08-06`. `update_work_log()` (canonical file lines ~1128-1153, as of `2026-08-06`; re-check line numbers before editing since the file changes frequently) matches the issue's description exactly: `local last_pr=$(work_log_max_pr)` then `select(.number > $last_pr)` with no presence check.

There is already one hermetic regression test covering an adjacent bug in this exact function — `defaults/scripts/tests/test-guide-work-log-self-loop.sh` (issue #5454, the Guide's own docs-maintenance PR self-perpetuating). That test extracts `work_log_max_pr()` and `GUIDE_DOCS_PR_EXCLUDE` **verbatim** from `guide.md` and runs them against fixtures — the file's own comment says "the regression suite... extracts THIS line out of THIS file and runs it against fixtures, so the prompt and the test can never drift apart." Follow the same pattern for this fix: whatever new filter logic replaces/augments `select(.number > $last_pr)` should be written so a new test can extract it verbatim, not re-implement it in the test.

### Acceptance Criteria

- [ ] `update_work_log()` in `defaults/.claude/commands/loom/guide.md` no longer permanently drops a merged PR whose number is below the current watermark but which was not previously recorded (the out-of-order-merge case from #5507/#5509).
- [ ] The fix implements Option 1 (presence check via `grep -q "PR #$n" "$DOCS_WT/WORK_LOG.md"` before excluding a candidate by number) from the "Suggested Fix" section above, or an equivalent approach — if the Builder chooses Option 2 or 3 instead, the PR description must explain why.
- [ ] The existing `GUIDE_DOCS_PR_EXCLUDE` self-loop exclusion (issue #5454) is preserved and still verbatim-extractable by `test-guide-work-log-self-loop.sh` — that test must continue to pass unmodified.
- [ ] A new hermetic regression test (suggested name: `defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh`, following the extraction-from-guide.md pattern in `test-guide-work-log-self-loop.sh`) reproduces the #5507/#5509 scenario: PR A opens before PR B but merges after B, B is already recorded in a fixture `WORK_LOG.md`, and the fixed logic still records A on the next simulated tick.
- [ ] No `gh`/forge calls added to the new test — keep it hermetic (pure jq/grep against a temp dir/fixture), matching the sibling test's design.

### Implementation Guidance

Widening the query window (`--limit 50` in the current `gh pr list --state merged` call) may also be needed alongside the presence check — a PR that merged long enough ago to fall outside the window can't be caught by a presence check that never sees it. Consider whether the existing `--limit 50` is sufficient for realistic out-of-order gaps, or whether it should grow / use `--search "merged:>...`" filtering by date instead of a fixed count.

### Affected Files

- `defaults/.claude/commands/loom/guide.md` - fix `update_work_log()`'s new-PR filter (`work_log_max_pr()` / the `select(.number > $last_pr)` expression, canonical file lines ~1128-1153 as of `2026-08-06`) per the Suggested Fix above.
- `defaults/scripts/tests/test-guide-work-log-self-loop.sh` - no code change expected, but must still pass after the fix (verifies the adjacent #5454 fix is untouched).
- `defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh` - new hermetic regression test (does not exist yet as of `2026-08-06`; verified via `ls defaults/scripts/tests/ | grep -i guide` showing only the self-loop test).

### Test Plan

- [ ] Automated: new hermetic test `test-guide-work-log-out-of-order-merge.sh` passes, reproducing the #5507/#5509 out-of-order scenario.
- [ ] Automated: `test-guide-work-log-self-loop.sh` still passes unmodified.
- [ ] Edge cases: (a) the out-of-order PR falls outside the merged-PR query window/limit; (b) multiple out-of-order PRs land in the same tick; (c) a PR that is both out-of-order *and* excluded by `GUIDE_DOCS_PR_EXCLUDE` (must stay excluded, not get pulled in by the presence-check fix).
- [ ] Manual verification: not required — the reproduction in the original issue body is fully reproducible hermetically (no live forge state needed).

### Duplicate / Related Work Check

`./.loom/scripts/check-duplicate.sh --include-merged-prs --issue 5516` returned no duplicates and no related open work as of `2026-08-06`.

rjwalters:
Closes #5516

## Summary

`update_work_log()`'s new-PR filter in `defaults/.claude/commands/loom/guide.md` used `select(.number > $last_pr)`, which assumes PR numbers merge in the same order they were created. They don't: a lower-numbered PR can sit in review/Doctor longer than a higher-numbered one that merges first. Once a later tick recorded the higher-numbered PR and the watermark advanced past the lower one's number, `> $last_pr` could never become true for it again — it was silently and permanently dropped. This happened for real with PR #5507 (opened first, merged after #5509).

## Fix (Option 1 from the issue)

- Added a new `work_log_has_pr()` helper: `grep -qx` presence check for `"PR #<N>"` already recorded in `WORK_LOG.md`.
- `update_work_log()` no longer filters PR candidates by `.number > $last_pr`. Instead it fetches candidates in a date-bounded window, applies the existing `GUIDE_DOCS_PR_EXCLUDE` exclusion (issue #5454, untouched), and then keeps a candidate only if `work_log_has_pr` says it isn't already recorded.
- Widened the merged-PR query from a fixed `--limit 50` to `--search "merged:>=<30-days-ago>"` with `--limit 200` as a safety cap, per the issue's Implementation Guidance — a count-only window could otherwise push an out-of-order PR out of the query entirely.
- Issue-closing filtering is unchanged (still the `.number > $last_issue` watermark) — the issue's own analysis notes issues close in far more reliable number order than PRs merge in, since #5516's root cause (Doctor/review dwell time) is PR-specific.

## Tests

New hermetic regression test `defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh` (wired into `ci-wired.txt`), following the extraction-from-guide.md pattern of `test-guide-work-log-self-loop.sh`:

1. Verifies guide.md defines `work_log_has_pr()`, wires it into `update_work_log()`, still applies `GUIDE_DOCS_PR_EXCLUDE`, and widened the query window.
2. Reproduces the #5507/#5509 scenario end-to-end: PR B is already recorded (watermark above PR A's lower number), and the fixed presence-check logic still records PR A on the next tick.
3. Edge case (a): an out-of-order PR that would fall outside a naive fixed-count `--limit 50` window is still caught (61-candidate fixture).
4. Edge case (b): multiple out-of-order PRs landing in the same tick are all recorded.
5. Edge case (c): an out-of-order PR that is *also* excluded by `GUIDE_DOCS_PR_EXCLUDE` stays excluded.

`test-guide-work-log-self-loop.sh` passes unmodified (verified locally).

```
bash defaults/scripts/tests/test-guide-work-log-self-loop.sh        # 18/18 PASS
bash defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh  # 18/18 PASS
bash defaults/scripts/tests/check-ci-suite-manifest.sh              # OK: manifest invariant holds
```

## Scope

Only `defaults/.claude/commands/loom/guide.md` (canonical source; `.loom/roles/guide.md` and `defaults/roles/guide.md` are symlinks to it, no separate sync needed), `defaults/scripts/tests/ci-wired.txt`, and the new test file were touched.

### DIFF
--- a/defaults/.claude/commands/loom/guide.md
+++ b/defaults/.claude/commands/loom/guide.md
@@ -1055,7 +1055,12 @@ side-car state file.
 > are the durable source of truth for "what has already been recorded."
 
 Compute the high-water marks by scanning the existing `WORK_LOG.md` for the
-highest PR / issue number it already contains:
+highest PR / issue number it already contains. `work_log_max_issue()` still
+gates which closed issues are new. `work_log_max_pr()` remains available as a
+general "highest PR recorded" reading, but PR filtering itself no longer uses
+a number watermark — see `work_log_has_pr()` and the #5516 comment in
+`update_work_log()` below for why a pure number comparison silently and
+permanently drops out-of-order-merged PRs.
 
 ```bash
 # Highest PR number already recorded in WORK_LOG.md (0 if none / file absent)
@@ -1066,6 +1071,16 @@ work_log_max_pr() {
 # Highest closed-issue number already recorded in WORK_LOG.md (0 if none)
 work_log_max_issue() {
   { grep -oE 'Issue #[0-9]+' "$DOCS_WT/WORK_LOG.md" 2>/dev/null | grep -oE '[0-9]+'; echo 0; } | sort -rn | head -1
+}
+
+# Whether "PR #<N>" is already literally recorded in WORK_LOG.md (#5516).
+# `grep -qx` matches a full extracted-number LINE exactly, so `#550` can
+# never false-match a recorded `#5501` the way a plain substring grep would.
+# This is the presence check update_work_log() uses INSTEAD of a pure
+# `.number > $last_pr` comparison — see the #5516 comment at its call site
+# for why number order cannot be trusted as a proxy for "already recorded".
+work_log_has_pr() {
+  grep -oE 'PR #[0-9]+' "$DOCS_WT/WORK_LOG.md" 2>/dev/null | grep -oE '[0-9]+' | grep -qx "$1"
 }
 ```
 
@@ -1101,7 +1116,9 @@ If a docs PR is already open, **skip the entire document maintenance phase** to 
 
 ### Step 2: Update WORK_LOG.md
 
-Append entries for newly merged PRs and closed issues since the last high-water mark.
+Append entries for newly merged PRs not yet recorded (a presence check, not a
+number watermark — see #5516 below) and closed issues since the last
+high-water mark.
 
 ```bash
 # #5454 BUG, DO NOT REINTRODUCE: this phase's OWN merged PRs must never count as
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge review: APPROVE

Reviewed against issue #5516's acceptance criteria and tested locally.

**Correctness**
- `update_work_log()` no longer filters PR candidates by `.number > $last_pr`. It now fetches candidates via a date-bounded `--search "merged:>=$since"` (30-day window, `--limit 200` safety cap) instead of the old fixed `--limit 50`, applies the unchanged `GUIDE_DOCS_PR_EXCLUDE` self-loop exclusion, and then keeps a candidate only if `work_log_has_pr()` says "PR #<N>" isn't already recorded in `WORK_LOG.md`.
- `work_log_has_pr()` extracts recorded PR numbers and does an exact-line match (`grep -qx`) against the target number — this correctly avoids substring false-positives (e.g. `#550` matching `#5501`), which is actually a more robust implementation than a literal `grep -qx "PR #$n" WORK_LOG.md` would be against real markdown lines like `- **PR #5509**: ...` (those never literally equal `"PR #$n"` on their own, since -x requires a full-line match) — a good judgment call, and functionally equivalent to Option 1 from the issue.
- The presence-check loop uses `while ... done < <(process substitution)`, not a piped `while read`, so `new_prs` accumulation correctly persists outside a subshell — good attention to a common bash pitfall.
- Verified `gh pr list --state merged --search "merged:>=<date>"` works as expected against the live repo.
- Issue-closing logic is untouched (still the `.number > $last_issue` watermark), matching the issue's own analysis that issue closure order is reliable while PR merge order isn't.
- No leftover dead references to the old `select(.number > $last_pr)` filter; `work_log_max_pr()` itself is kept (documented as still useful for reading the recorded high-water mark) but is no longer used to gate PR filtering.

**Tests**
- New hermetic test `defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh`, wired into `ci-wired.txt`, ran locally: 18/18 PASS. It reproduces the #5507/#5509 scenario end-to-end, plus edge cases: a candidate that would fall outside a naive fixed-count window, multiple out-of-order PRs in one tick, and an out-of-order PR that's also `GUIDE_DOCS_PR_EXCLUDE`-excluded (correctly stays excluded).
- `test-guide-work-log-self-loop.sh` passes unmodified: 18/18 PASS — confirms the #5454 exclusion is preserved.
- `check-ci-suite-manifest.sh`: OK (129 suites, 122 wired, 7 excluded).
- Shellcheck on the new test file: only pre-existing-style info-level notices (SC2317, SC2016), same as the sibling `test-guide-work-log-self-loop.sh` — not a regression.
- New test is hermetic — no `gh`/forge calls, confirmed by reading the file.

**CI / merge state**
- All 16 checks pass (Shellcheck, Shell Test Suites, Installer Integration Tests, etc. — verified via `gh pr checks 5531`).

**Acceptance criteria (issue #5516)** — all met:
- [x] Out-of-order-merged PR below the watermark is no longer permanently dropped.
- [x] Presence-check approach (equivalent to Option 1), with clear rationale in code comments.
- [x] `GUIDE_DOCS_PR_EXCLUDE` preserved, sibling test passes unmodified.
- [x] New hermetic regression test reproduces the #5507/#5509 scenario plus the three specified edge cases.
- [x] No `gh`/forge calls in the new test.

rjwalters:
✅ **Approved!**

Solid fix — correctly diagnoses that PR numbers don't merge in creation order (unlike issue closures, which the PR rightly leaves on the number watermark), and replaces the `.number > $last_pr` filter with a presence check (`work_log_has_pr`) plus a date-bounded query window so an out-of-order PR can't be pushed out by a fixed `--limit`. The process-substitution `while` loop avoids the classic pipe-into-while subshell pitfall, and the GNU/BSD `date` fallback matches the existing pattern in `champion-issue-promo.md`.

## Test Execution

- `bash defaults/scripts/tests/test-guide-work-log-self-loop.sh` — ✅ 18/18 PASS (unmodified, confirms no regression on the #5454 docs-PR exclusion)
- `bash defaults/scripts/tests/test-guide-work-log-out-of-order-merge.sh` — ✅ 18/18 PASS (reproduces the #5507/#5509 scenario end-to-end plus window/multi-PR/docs-exclusion edge cases)
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` — ✅ OK, new test is wired into `ci-wired.txt`
- CI (`gh pr checks`): all required checks pass, including "Shell Test Suites (hermetic)" which was still pending at claim time — polled to completion, now green
- Symlink scope verified: only the canonical `defaults/.claude/commands/loom/guide.md` was touched; `.loom/roles/guide.md` and `defaults/roles/guide.md` remain symlinks to it, no separate sync needed
- Merge state: CLEAN

PR description correctly uses `Closes #5516`.

Issue #5457, opened 2026-08-05
Committed .loom/config.json hardcodes /Users/... macOS paths — Linux hosts must patch it, which then blocks git pull


rjwalters:
## Summary

`.loom/config.json` — committed, dogfooded, and shipped to consumers — contains **macOS absolute paths** for two settings. Any Linux host whose daemon resolves that config must patch it locally, and that local patch then **blocks `git pull`**, silently stranding the host on an old checkout.

```json
"safehouse": { "socket": "/Users/rwalters/.loom/safehoused/state/safehoused.sock", ... }
"observability": { "ingestKeyFile": "/Users/rwalters/.loom/observability/ingest.key", ... }
```

Both are `/Users/rwalters/...`, which cannot exist on Linux.

## How it presents

On an Ubuntu fleet worker whose daemon runs from the `loom` checkout:

1. The host patched `ingestKeyFile` to `/home/ubuntu/...` locally so telemetry would work.
2. That made `.loom/config.json` a dirty tracked file.
3. Every subsequent `git pull` aborted:

```
error: Your local changes to the following files would be overwritten by merge:
	.loom/config.json
Please commit your changes or stash them before you merge.
```

4. The host sat **75 commits behind `origin/main`** — long enough that it lacked `--heal-watchdog-only` (#5405), the very flag needed to repair its missing watchdog. A config path bug had blocked a crash-protection fix.

Meanwhile `safehouse.socket` still pointed at the macOS path, so the daemon reported `Safehouse: configured, unreachable` indefinitely — it was looking for a socket that could never exist on that host. `safehoused` was not installed there at all.

The sibling worker was unaffected for an incidental reason: its `LOOM_WORKSPACE` points at a *different* repo (`anvil`), so it never resolved `loom`'s own config. That makes the failure look host-specific when it is actually config-specific.

## Why this is worth fixing rather than documenting

The failure is silent and self-reinforcing. The local patch is *necessary* for the host to work, and it is exactly what prevents the host from receiving updates. Nothing surfaces "this host stopped pulling three weeks ago" — the daemon keeps reporting healthy.

It also affects consumer repos, not just this one: `defaults/` ships the same config shape, so any Linux consumer inherits the same trap.

## Suggested acceptance criteria

- [ ] No absolute `/Users/...` paths in committed config. Use `~`/`$HOME` expansion, a path relative to the config root, or omit the key and resolve a platform default at runtime
[...]

PR #5533, opened 2026-08-06
fix(safehouse): loudify unresolved-socket failure + add drift check (#5523)

### LINKED ISSUE #5523: #5457 left safehouse socket resolution with no default — every sweep silently stopped narrating, froze the public pulse for 11h
## What happened

#5457 removed the hardcoded socket path from the committed `.loom/config.json`
— correctly, because Linux hosts had to patch a tracked file and that blocked
`git pull` (it stranded loom-worker-2 75 commits behind). But nothing took its
place as a default, and `loom_mcp_safehouse_socket()` treats "unresolved" as
"silently disabled".

Result: from the moment #5457 merged, **every sweep on robb-studio stopped
narrating to safehouse**, and the only trace was a `log_warn` buried in
per-role sweep logs:

```
WARN spawn-claude: safehouse enabled but no socket resolves
(safehouse.socket / $LOOM_SAFEHOUSE_SOCKET / $SAFEHOUSED_SOCKET);
skipping safehouse MCP injection (loom MCP unaffected).
```

Repeated at 04:27, 04:50, 05:18, 05:27, 05:52, 06:19Z — every sweep, for
11 hours, unnoticed.

## Blast radius: this silently broke the public website

The safehouse room is not only agent chatter — it is the **sole input** to the
2amlogic.com homepage fleet feed:

```
agents → safehouse room (completion-v1) → safehoused egress sink
       → POST 2amlogic.com/api/ingest → D1 → /api/feed → homepage
```

With narration off, the room went silent, safehoused's egress delay buffer
starved (1006 rows, 0 unpublished, newest `completed_at` **2026-08-05T19:10:42Z**
— it never received anything to enqueue), and the public pulse froze for
**11 hours** while the fleet closed 108 issues. The homepage day bucket for
2026-08-05 reads **47** issues against a 142–251 daily norm.

Timeline:

| time (UTC) | event |
|---|---|
| 2026-08-05T18:41:07Z | #5457 merges; socket path removed from committed config |
| 2026-08-05T19:10:42Z | last completion egressed — sweeps already in flight when the change landed |
| 19:10Z → 06:20Z | every new sweep logs "no socket resolves"; room silent; pulse frozen |
| 2026-08-06T06:20Z | found by a human noticing the website had stopped updating |

I initially misdiagnosed this as a hung sync in safehoused and restarted it,
which changed nothing — retracted at rjwalters/safehouse#79. A 0%-CPU process
holding an idle long-poll is indistinguishable from a wedged one from the
outside, which is its own lesson, but the fault was always here.

## The defect

`defaults/scripts/lib/mcp-config.sh:58`:

```sh
loom_mcp_safehouse_socket() {
    local repo_root="$1"
    if [[ -n "${LOOM_SAFEHOUSE_SOCKET:-}" ]]; then printf '%s\n' "$LOOM_SAFEHOUSE_SOCKET"; return 0; fi
    if [[ -n "${SAFEHOUSED_SOCKET:-}" ]];   then printf '%s\n' "$SAFEHOUSED_SOCKET";   return 0; fi
    cfg="$(loom_config_get "$repo_root" "safehouse.socket" "")"
    if [[ -n "$cfg" && "$cfg" != "null" ]]; then printf '%s\n' "$cfg"; return 0; fi
    printf '\n'          # <-- no default; caller treats empty as "skip silently"
}
```

Two independent problems:

1. **No default.** safehoused's conventional socket path
   (`~/.loom/safehoused/state/safehoused.sock`) is well-known and was correct on
   this host the entire time — the socket file existed and safehoused was
   running and healthy. Nothing consulted it.
2. **`enabled: true` + unresolved = silent skip.** The config explicitly says
   `"safehouse": {"enabled": true, ...}`. An operator who has asked for
   safehouse and gets nothing should see an error, not a warning inside a log
   that is only read after an outage. The current behavior makes a
   configuration regression indistinguishable from normal operation.

This is fleet-wide, not host-specific: the socket is resolved **per repo root**,
so every managed repo on every host that was not explicitly reprovisioned lost
narration at the same moment. Setting `LOOM_SAFEHOUSE_SOCKET` once in a host's
daemon environment would cover all of that host's repos; a per-repo
`safehouse.socket` would not.

## Workaround applied on robb-studio

Added to the gitignored host-local tier (`.loom-local/local.json`), which is
where #5457 intended per-host values to live, for the `loom` repo only:

```json
"safehouse": { "socket": "/Users/rwalters/.loom/safehoused/state/safehoused.sock" }
```

Verified `loom_mcp_safehouse_socket` now returns that path. Every other managed
repo on every host is still silently un-narrated.

- [ ] `loom_mcp_safehouse_socket()` falls back to the conventional path (`${XDG_RUNTIME_DIR:-$HOME/.loom}/safehoused/state/safehoused.sock` or equivalent) before giving up
- [ ] `safehouse.enabled: true` with an unresolvable socket is an **error**, not a `log_warn` — either fail the spawn or emit at a level that reaches an operator surface
- [ ] The warning names the consequence, not just the mechanism: "no safehouse narration will be recorded; the 2amlogic.com pulse is fed from this" beats "skipping safehouse MCP injection"
- [ ] A one-shot check (`loom-daemon doctor`, or an addition to `compute-drift.sh`) reports per host whether the safehouse socket resolves for each managed repo, so this is detectable without reading sweep logs
- [ ] Re-provision the fleet: confirm every host/repo pair resolves a socket after the fix, since #5457's removal hit all of them simultaneously

## Related

- #5457 — the change this regressed from (a correct fix; the gap is the missing default)
- rjwalters/safehouse#79 — my incorrect initial diagnosis, retracted
- 2AMLogic/2am#45 — nothing alerts on pulse staleness, which is why this ran 11 hours

---

### Verified corrections

- **2026-08-06, verified against `origin/main` — the defect is confirmed exactly as reported.** `defaults/scripts/lib/mcp-config.sh`'s `loom_mcp_safehouse_socket()` has no fallback and returns empty when nothing resolves; `defaults/scripts/spawn-claude.sh` (the caller, ~line 631) logs exactly the quoted `log_warn` and skips injection when the socket is empty. Four call sites depend on this one resolver function, so a fix here fixes all of them: `defaults/scripts/spawn-claude.sh:616`, `defaults/scripts/cli/loom-daemon-start.sh:995`, `defaults/scripts/cli/safehoused-service.sh:311-312`, `scripts/setup-mcp.sh:154`.

- **2026-08-06, verified against `origin/main` — AC bullet 1 conflicts with an explicit, documented architecture decision; this needs to be resolved before implementation, not silently coded around.** `defaults/docs/safehouse.md` (lines ~71-88) states, in the caller's own words:

  > "If none resolves, narration logs one `warn!` and stays off — **no built-in `$HOME`-relative default, since safehouse is opt-in per-host**."
  >
  > "`socket` must never be committed to the shared `.loom/config.json`... Leave it unset in the committed file and either install `safehoused` at the conventional path each host's `$SAFEHOUSED_SOCKET` already points at, or set a per-host override in the gitignored `.loom-local/local.json` tier..., or via `$LOOM_SAFEHOUSE_SOCKET` — never in the committed file. **#5457 is exactly the failure mode this avoids**: a macOS `safehouse.socket` path was committed to this repo's own shared `.loom/config.json`, and... every other host that `git pull`ed `main` inherited a path to a socket that did not exist on it..."

  I independently confirmed there is no hardcoded default anywhere in the resolution chain: `defaults/scripts/cli/safehoused-service.sh`'s own `--socket` handling (`OPT_SOCKET`, ~line 275) falls through to the same `loom_mcp_safehouse_socket` resolver with no baked-in path either — "no default" is consistent across every caller, not an oversight isolated to one function.

  **This means AC bullet 1 as worded ("falls back to the conventional path... before giving up") asks for the literal thing the doc says was deliberately omitted, for the reason #5457 itself established.** It is not obviously wrong to add it — the socket path being genuinely conventional (installed by `safehoused-service.sh` at one well-known location per host) is a real argument for it being safe as a *code-level* default in a way a *committed config value* is not, since a code default can never go stale via `git pull` the way #5457's committed value did. But that argument needs to be made explicitly, and `defaults/docs/safehouse.md` updated to match, not left saying "no built-in default" while the code ships one. **A Builder should not implement AC bullet 1 without either (a) updating `defaults/docs/safehouse.md`'s "Socket resolution" section to document the new default and why it doesn't reintroduce the #5457 failure mode, or (b) dropping AC bullet 1 and relying on AC bullets 2 and 4 below** (loud failure + fleet-wide drift detection) as the fix, which is what the doc's existing guidance ("install `safehoused` at the conventional path... or set a per-host override") already assumes operators are doing — bullet 4 catches the case where they haven't, without touching the "no default" design.

  Also note: two different "conventional path" values appear in this repo's own fixtures — `~/.loom/safehoused.sock` (`defaults/scripts/tests/test-loom-daemon-start.sh:731,781,1131`) vs. `/run/safehoused.sock` (the example in `defaults/docs/safehouse.md:102`) vs. the reporter's actual live-host workaround, `~/.loom/safehoused/state/safehoused.sock`. None of these is stated anywhere as *the* canonical default — whoever implements bullet 1 needs to get the authoritative path from `rjwalters/safehouse` (the daemon's own default bind location), not guess from whichever of these three this repo happens to reference.

- **2026-08-06, verified against `origin/main` — the tooling named in AC bullet 4 doesn't exist under either name.** There is no `loom-daemon doctor` CLI subcommand (`loom-daemon/src/main.rs`'s `Subcommand` enum has no `doctor` variant — the `"doctor"` hits in `loom-daemon/src/` are all about the Doctor *role*, unrelated) and no `compute-drift.sh` file anywhere in the repo. The closest existing "drift" concept, `defaults/scripts/verify-install.sh`, checks installed-file hash drift (modified/missing copied files) — a different kind of drift than "does this host's safehouse socket resolve," and not an obvious place to bolt this onto. Whoever implements bullet 4 will need to either create new tooling (a small standalone script, or a new probe wired into an existing periodic check) or pick a different existing entry point than either name suggests — treat both names in the AC as illustrative, not literal targets.

### Affected Files

- `defaults/scripts/lib/mcp-config.sh` — `loom_mcp_safehouse_socket()` (~line 58), the fix's core.
- `defaults/scripts/spawn-claude.sh` (~line 616-640) — the `log_warn` → error/loud-signal change (AC bullet 2) and the improved warning text (AC bullet 3) land here.
- `defaults/docs/safehouse.md` (~lines 71-88, 102) — must be updated in lockstep with whatever AC bullet 1 resolves to (see "Verified corrections" above); leaving stale doc text describing "no built-in default" after adding one is its own bug.
- `defaults/scripts/tests/test-mcp-config.sh` (~lines 129-153) — existing `loom_mcp_safehouse_socket` test coverage to extend for whatever new fallback/error behavior lands.
- `defaults/scripts/cli/safehoused-service.sh`, `defaults/scripts/cli/loom-daemon-start.sh`, `scripts/setup-mcp.sh` — other callers of the same resolver; verify none of them assume today's "silent empty string" return value in a way a loud failure would break.
- New file (name TBD per "Verified corrections" above) or an existing periodic-check entry point — for AC bullet 4's fleet-wide per-host/per-repo socket-resolution report.

### Test Plan

- [ ] Manual verification: on a host with `safehoused` running at the resolved default (once bullet 1's design question is settled), confirm `loom_mcp_safehouse_socket` returns a non-empty path with no env/config override set.
- [ ] Manual verification: with `safehouse.enabled: true` and no socket resolvable by any path, confirm the new behavior is loud (a failed spawn, or a warning that reaches an operator-visible surface — not just a per-role sweep log) rather than the current silent skip.
- [ ] Automated tests: extend `defaults/scripts/tests/test-mcp-config.sh` for the new resolution/error behavior; add coverage for whatever new drift/doctor check AC bullet 4 introduces.
- [ ] Edge cases: `safehouse.enabled: false` must remain a byte-for-byte no-op (the existing "degradation contract" comment in `spawn-claude.sh` — do not regress this while loudening the enabled-but-unresolvable case).
- [ ] Fleet re-provisioning (AC bullet 5) is an operational step, not a code change — track separately from the PR's own test plan; note in the PR which hosts were confirmed re-provisioned, if any, at merge time.

rjwalters:
## Summary

#5457 removed the only hardcoded `safehouse.socket` default (correctly — it had
been committed to the shared `.loom/config.json` and stranded a host 75
commits behind `git pull`), but nothing replaced it as a *check*:
`loom_mcp_safehouse_socket()` has always silently returned empty when nothing
resolves, and the only signal was a `log_warn` buried inside each sweep's own
per-role log. Every sweep on an affected host silently stopped narrating to
safehouse for 11 hours before a human noticed the public 2amlogic.com fleet
pulse had gone stale.

## AC bullet 1 — deliberately NOT implemented as literally worded

The curator's own analysis flagged that AC bullet 1 ("fall back to the
conventional path... before giving up") conflicts with `defaults/docs/safehouse.md`'s
explicit, documented decision that there is **no built-in default** — because
that is exactly the class of failure #5457 fixed (a *committed* default socket
path silently going stale via `git pull` on every other host). The curator
recommended option (b): drop the fallback-default idea and rely on making the
failure loud (AC bullet 2) + fleet-wide drift detection (AC bullet 4) instead.
I went with (b) — a code-level default would avoid #5457's exact mechanism
(it can't go stale via `git pull`), but it reintroduces the same underlying
risk in a different shape: "resolves to *something*" would quietly stop
meaning "actually reaches a live `safehoused`", which is precisely the kind of
silent gap that let this incident run 11 hours unnoticed. `defaults/docs/safehouse.md`
now has an explicit "Why there is still no built-in default, even after #5523"
paragraph documenting this reasoning in lockstep with the code.

## What changed

- **`defaults/scripts/spawn-claude.sh`** — the unresolved-socket warning now
  names the consequence ("no safehouse narration will be recorded... the
  2amlogic.com public fleet pulse is fed exclusively from safehouse
  narration") instead of only the mechanism ("skipping safehouse MCP
  injection"), and points at the new check below. Still a `log_warn`, never a
  failed spawn — the degradation contract (`safehouse.enabled: false` stays a
  byte-for-byte no-op; `enabled: true` never blocks a sweep) is unchanged and
  covered by existing tests.
- **New `defaults/scripts/check-safehouse-socket.sh`** — a standalone,
  on-demand, per-repo drift check with **no dependency on a daemon restart**
  (the pre-existing `loom-daemon-start.sh` static `Safehouse:` line only
  re-runs when the daemon itself restarts — exactly the case that let this
  incident run unnoticed on an already-running daemon). With no arguments it
  walks the machine-level workspace registry (`~/.loom/workspaces.json`, the
  same one `loom-daemon workspace add/list` manages); explicit repo-root args
  check a subset. `--json` emits a parseable array for scripting/cron. Exits
  `0` when every repo is either not-configured or resolved+present, `1` the
  moment any repo is enabled with an unreachable socket (unresolved, or
  resolved-but-nothing-there).
- **`defaults/scripts/lib/mcp-config.sh`** — no functional change to
  `loom_mcp_safehouse_socket()`; added a comment documenting the "deliberately
  no fallback default" decision so a future reader doesn't mistake this for an
  oversight.
- **`defaults/docs/safehouse.md`** — updated in lockstep: the new "Why there
  is still no built-in default, even after #5523" writeup, and a pointer to
  the new check under "Socket resolution" and the Implementation file list.
- **`defaults/scripts/tests/test-mcp-config.sh`** — new §13 covering the
  loudened warning text (static source assertions) and every
  `check-safehouse-socket.sh` state: not-configured, resolved, unresolved (no
  socket), unresolved (socket resolved but missing on disk), multi-repo runs,
  `--json` output, and the no-registry/no-cwd-repo no-op path.

## Out of scope (per curator)

Fleet re-provisioning (visiting every host to confirm it now resolves) is an
operational follow-up, not a code change, and not done as part of this PR.

## Test plan

- `./defaults/scripts/tests/test-mcp-config.sh` — 120/120 passing (was 101
  before this PR's new §13 additions).
- `./defaults/scripts/tests/test-loom-daemon-start.sh` — 112/112 passing
  (unaffected, confirms the pre-existing static `Safehouse:` check still
  works).
- `./defaults/scripts/tests/test-spawn-claude.sh` — 180/189 passing; the 9
  failures are pre-existing and unrelated (niceness/taskpolicy/CPU-budget
  tests, `#4233`/`#5111`), reproduced identically against `origin/main`'s
  unmodified `spawn-claude.sh` in this sandboxed nested-sweep environment
  before restoring my changes.
- `shellcheck` clean on all new/modified shell files.
- Manually exercised `check-safehouse-socket.sh` against fixture repos for
  every state (not-configured / resolved / unreachable-no-socket /
  unreachable-missing-file / no-registry) in both text and `--json` modes.

Closes #5523

### DIFF
--- a/defaults/docs/safehouse.md
+++ b/defaults/docs/safehouse.md
@@ -69,8 +69,10 @@ Env overrides (each wins over config for that key):
 | `LOOM_SAFEHOUSE_ROOM_CLAIMS` | `rooms.claims` — dedicated peer-claim coordination room (#4713) |
 
 **Socket resolution** (precedence **env > config**, `resolve_socket` in
-`loom-daemon/src/safehouse.rs`): `$LOOM_SAFEHOUSE_SOCKET` → `$SAFEHOUSED_SOCKET`
-(the unprefixed convention `safehoused` clients also read) → the configured
+`loom-daemon/src/safehouse.rs`; the bash-side worker-injection path
+(`defaults/scripts/lib/mcp-config.sh`'s `loom_mcp_safehouse_socket()`) mirrors
+the same chain): `$LOOM_SAFEHOUSE_SOCKET` → `$SAFEHOUSED_SOCKET` (the
+unprefixed convention `safehoused` clients also read) → the configured
 `socket` value. If none resolves, narration logs one `warn!` and stays off — no
 built-in `$HOME`-relative default, since safehouse is opt-in per-host.
 
@@ -87,6 +89,47 @@ macOS `safehouse.socket` path was committed to this repo's own shared
 value *before* env at the time — every other host that `git pull`ed `main`
 inherited a path to a socket that did not exist on it, with no env override able
 to take effect while that stale path stayed committed.
+
+**Why there is still no built-in default, even after #5523.** #5457's fix left
+a gap: with the committed default gone and nothing installed in its place, an
+affected host's `safehouse.enabled: true` silently resolved to no socket at
+all, and the only signal was a `log_warn` inside each sweep's own per-role log
+— nobody was tailing those, so a real host ran with **zero** safehouse
+narration for 11 hours before a human noticed the public 2amlogic.com fleet
+pulse had gone stale (#5523). The tempting fix — teach the resolver a
+conventional-path fallback (e.g. `~/.loom/safehoused/state/safehoused.sock`) —
+was deliberately **rejected** for #5523: a code-level default *would* avoid
+re-triggering #5457's exact mechanism (it can't go stale via `git pull` the
+way a committed value can), but it would reintroduce the same underlying risk
+in a different shape — "resolves to *something*" would quietly stop meaning
+"actually reaches a live `safehoused`", which is precisely the gap that let
+#5523 run unnoticed. #5523's fix instead makes the **absence** loud and cheap
+to detect, in two ways, without touching this resolution chain at all:
+
+- `spawn-claude.sh`'s warning, when `safehouse.enabled` is true and no socket
+  resolves, now names the consequence ("no safehouse narration will be
+  recorded... the 2amlogic.com public fleet pulse is fed exclusively from
+  safehouse narration") instead of only the mechanism ("skipping safehouse MCP
+  injection") — still a `log_warn`, never a failed spawn (the degradation
+  contract above is unchanged: `safehouse.enabled: false`/absent stays a
+  byte-for-byte no-op, and `enabled: true` never blocks a sweep).
+- **`defaults/scripts/check-safehouse-socket.sh`** (installed as
+  `.loom/scripts/check-safehouse-socket.sh`) is a new standalone, on-demand
+  check that reports — per managed repo, without reading a single sweep log —
+  whether `safehouse.enabled` is set and, if so, whether a socket resolves AND
+  is present on disk. With no arguments it walks this host's machine-level
+  workspace registry (`~/.loom/workspaces.json`, the same registry
+  `loom-daemon workspace add/list` manages) and checks every registered repo
+  in one pass; pass explicit repo roots to check a subset. `--json` emits a
+  parseable array (`{repo, enabled, socket, present, status}` per repo) for
+  scripting/cron. It exits `0` when every repo is either not configured or
+  resolved-and-present, and `1` the moment any repo is enabled with an
+  unreachable socket — the exact condition #5523 needed surfaced. Unlike the
+  pre-existing static `Safehouse:` line in `loom-daemon-start.sh`'s startup
+  banner (below), this check has **no dependency on the daemon
+  (re)starting** — it can be run any time, which is what the incident
+  actually needed: the affected daemon never restarted across the 11-hour
+  window, so its own start-time check never re-ran.
 
 ### Room routing by attention class (`safehouse.rooms`, #4225)
 
@@ -843,10 +886,17 @@ what feeds the public fleet feed on 2amlogic.com. Loom is the producer:
 - `defaults/scripts/cli/loom-daemon-start.sh` — (#4345) the static
   pre-connect `Safehouse:` line, via `lib/mcp-config.sh`'s existing
   `loom_mcp_safehouse_enabled`/`loom_mcp_safehouse_socket` resolvers.
+- `defaults/scripts/check-safehouse-socket.sh` — (#5523) the standalone,
+  daemon-restart-independent per-repo socket-resolution drift check described
+  above under "Socket resolution".
 - Tests: `safehouse.rs`'s `mod tests` (state-cell + wire-rendering cases),
   `workspace_pool.rs`'s `mod tests` (pool wiring), `ipc.rs`'s
   `test_build_daemon_status_reports_halt_and_in_flight` (report field),
-  `defaults/scripts/tests/test-loom-daemon-start.sh` (start-wrapper line).
+  `defaults/scripts/tests/test-loom-daemon-start.sh` (start-wrapper line),
+  `defaults/scripts/tests/test-mcp-config.sh` §13 (#5523: the loudened
+  spawn-claude.sh warning text + check-safehouse-socket.sh's not-configured /
+  resolved / unreachable-no-socket / unreachable-missing-file / multi-repo /
+  `--json` behavior).
 
 ## Peer-claim coordination: cross-host soft claim (#4028)
 
--- a/defaults/scripts/check-safehouse-socket.sh
+++ b/defaults/scripts/check-safehouse-socket.sh
@@ -1,0 +1,191 @@
+#!/usr/bin/env bash
+# check-safehouse-socket.sh — one-shot, per-host safehouse-socket resolution
+# report (issue #5523).
+#
+# Why: #5457 removed the only hardcoded `safehouse.socket` default — a path
+# that had been committed to the SHARED `.loom/config.json` and stranded
+# loom-worker-2 75 commits behind `git pull` (a correct fix). But nothing took
+# its place as a *check*: `loom_mcp_safehouse_socket()`
+# (defaults/scripts/lib/mcp-config.sh) has always silently returned an empty
+# string when nothing resolves, and every caller's only signal was a
+# `log_warn` buried inside that particular sweep's own log file. From the
+# moment #5457 merged, every sweep on an affected host silently stopped
+# narrating to safehouse — and the failure ran for 11 hours before a human
+# noticed the public 2amlogic.com fleet pulse had gone stale, because nothing
+# outside a sweep's own log was watching for it.
+#
+# This script is that "something watching": a standalone, on-demand check
+# that reports — per managed repo, without reading a single sweep log —
+# whether `safehouse.enabled` is set and, if so, whether a socket resolves
+# AND is actually present on disk. It reuses the exact resolvers
+# spawn-claude.sh / loom-daemon-start.sh already call
+# (`defaults/scripts/lib/mcp-config.sh`), so "drift" here means exactly what
+# it means to a spawning worker — no separate definition to keep in sync.
+#
+# Deliberately does NOT add a code-level conventional-path default to the
+# resolver (see mcp-config.sh's `loom_mcp_safehouse_socket` header and
+# defaults/docs/safehouse.md "Socket resolution" for the full writeup on
+# why): the fix for "silence went unnoticed for 11 hours" is making the
+# silence loud and cheap to check, independent of the daemon's own
+# start/restart lifecycle — not guessing a path. The pre-existing static
+# check in `loom-daemon-start.sh` (`print_safehouse_status`) only re-runs
+# when the daemon itself (re)starts, which is exactly the case that let this
+# incident run unnoticed on an already-running daemon; this script has no
+# such dependency.
+#
+# Usage:
+#   check-safehouse-socket.sh [--json] [REPO_ROOT ...]
+#
+#   With no REPO_ROOT arguments, checks every workspace registered in this
+#   host's machine-level daemon registry (~/.loom/workspaces.json, or
+#   $LOOM_WORKSPACES_PATH — see loom-daemon/src/workspace_registry.rs),
+#   falling back to the enclosing repo of the current directory when the
+#   registry is empty/absent/unreadable, so a lone interactive host still
+#   gets a useful answer.
+#
+# Exit codes:
+#   0 — every checked repo is either "not configured" (safehouse.enabled is
+#       false/absent — a byte-for-byte no-op, see the degradation contract in
+#       defaults/docs/safehouse.md) or "resolved" (a socket path resolved AND
+#       is present on disk).
+#   1 — at least one repo is "configured, enabled, unreachable": either
+#       safehouse.enabled is true with no socket resolving at all (the exact
+#       #5523 failure mode), or a socket path resolved but nothing exists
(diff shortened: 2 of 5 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-06T08:37:44Z, age <1 minute) — standing down without reclaiming. Not stomping.

rjwalters:
## Judge review: APPROVE

Reviewed the actual diff (5 files: `defaults/docs/safehouse.md`, `defaults/scripts/check-safehouse-socket.sh` (new), `defaults/scripts/lib/mcp-config.sh`, `defaults/scripts/spawn-claude.sh`, `defaults/scripts/tests/test-mcp-config.sh`) against a worktree checked out to the PR head (`d13e746e`).

**AC bullet 1 (no hardcoded fallback default) — correct call, well-argued.** `mcp-config.sh`'s `loom_mcp_safehouse_socket()` has no functional change, only a new comment explaining the deliberate omission. I agree with the reasoning: a code-level conventional-path default avoids #5457's exact mechanism (can't go stale via `git pull`) but reintroduces the same underlying risk in a different shape ("resolves" no longer implying "reaches a live `safehoused`"). Good judgment call, not just an assertion — it's argued from the actual incident mechanics in both the code comment and `safehouse.md`.

**Loudened warning (`spawn-claude.sh`)** — correct and minimal. Still a `log_warn`, never a failed spawn; the degradation contract (`enabled:false` no-op, `enabled:true` never blocks) is preserved. Now names the actual consequence (2amlogic.com pulse going stale) and points at the new check.

**New `check-safehouse-socket.sh`** — read the whole script. Design is sound: walks the machine-level workspace registry by default (no daemon-restart dependency, which is exactly what let this run unnoticed for 11h), accepts explicit repo roots, `--json` for scripting, sensible exit codes (0 = nothing wrong / not configured, 1 = drift, 2 = usage error for a stale/partial install). Correctly distinguishes "unresolved" vs "resolved but nothing on disk" (`-S`/`-e` check). No-registry/no-cwd-repo path degrades to a clean no-op rather than a false DRIFT. Executable bit is set in git (`100755`).

**Doc update (`safehouse.md`)** — the new "Why there is still no built-in default, even after #5523" section genuinely resolves the tension the curator flagged rather than just asserting it: it walks through why a code-level default was tempting, why it was rejected, and what was done instead, cross-referenced from both the code comment and the Implementation/Tests file list.

**Test coverage** — ran `test-mcp-config.sh` myself: **120/120 passing**, confirming the PR's claim. New §13 covers the loudened warning text (static source assertions) and every `check-safehouse-socket.sh` state (not-configured, resolved, unreachable-no-socket, unreachable-missing-file, multi-repo union-exit semantics, `--json` shape, no-registry no-op) — good coverage, no gaps I could find.

**Pre-existing `test-spawn-claude.sh` failures — verified independently.** Built a separate worktree off `origin/main` (unmodified, pre-PR), reused the same prebuilt `loom-daemon` binary, and ran the suite: **180/189, same 9 failures**. Diffed the two failure lists (PR branch vs. origin/main) — byte-identical set of 9 test names. These are genuinely pre-existing (niceness/taskpolicy/CPU-budget, #4233/#5111) and not something this PR broke.

**shellcheck** — clean on all changed files (CI's Shellcheck check also passes). One info-level SC2016 in the test file's existing single-quoted heredoc pattern, unrelated to this PR's new code.

**CI / merge state** — all checks green (`gh pr checks 5533`), including "Shell Test Suites (hermetic)" which was still running at review start and I waited out. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

No blockers. Approving.

PR #5554, opened 2026-08-06
fix(daemon): replace pgrep -f loom-daemon liveness checks with exact process-name matching

### LINKED ISSUE #5548: pgrep -f loom-daemon is not a liveness check — leaked test fixtures named loom-daemon kept a dead daemon looking healthy for 66 minutes
## What happened

robb-pro's `loom-daemon` was **down for ~66 minutes** while every check I ran
reported it healthy, including one I ran deliberately to rule that out.

```
$ pgrep -f loom-daemon | head -1
8259
$ ps -o etime= -p 8259
02-16:15:01        # "up 2 days 16 hours" — looks like the most stable host in the fleet
```

That process is not the daemon:

```
$ ps -o pid,ppid,args -p 8259
 8259  1  bash /var/folders/cz/6r9_2ff50l960nqx9zcppsqr0000gn/T/tmp.qKvzC9VsYR/loom-daemon
```

It is a **bash script named `loom-daemon` in `$TMPDIR`**, orphaned to PPID 1 —
a leaked test fixture. There were three of them:

```
 8259 bash /var/folders/.../T/tmp.qKvzC9VsYR/loom-daemon
59040 bash /var/folders/.../T/tmp.zoyIIgwfQk/loom-daemon
66068 bash /var/folders/.../T/tmp.uW8x6RVM6f/loom-daemon
```

The real daemon's actual state was visible only through launchd:

```
$ launchctl list | grep loom-daemon
-   143  com.rjwalters.loom-daemon            # '-' = not running, 143 = SIGTERM
-     1  com.rjwalters.loom-daemon-watchdog   # also not running, exited 1
```

Recovered with `launchctl kickstart -k gui/$(id -u)/com.rjwalters.loom-daemon`.

## Why this matters beyond one host

**`pgrep -f loom-daemon` is not a liveness check** — it is a substring match
against the full command line of every process on the box, and any test fixture
that names its mock binary `loom-daemon` satisfies it forever after it leaks.
The failure is silent, stable, and inverted: the longer a leaked fixture
survives, the *healthier* the host looks, because `etime` grows.

This is the same family as #5131 (its label-blind `pgrep` tier), but the
consequence here is different: there it caused a wrong kill, here it causes a
wrong **all-clear**.

The only reason it surfaced at all is that the dashboard reports
`host.health` telemetry, and telemetry stops when the daemon stops. That is a
positive signal — the daemon actively publishing — and it disagreed with the
process check. Where those two disagree, the telemetry is right.

- [ ] Every liveness check for the daemon keys on something a fixture cannot forge: the launchd/systemd unit state, the PID file (`LOOM_PID_FILE`), or the daemon's own socket — not a `pgrep` name match
- [ ] Audit and fix existing `pgrep -f loom-daemon` (and `pgrep -f safehoused`, which has the same shape) across `defaults/scripts/`, the watchdog, and any doc that recommends it as a health check
- [ ] Test fixtures do not name their mock binary `loom-daemon`; use a distinct name (`loom-daemon-mock`) so a leak can never satisfy a production check
- [ ] Fixtures are reaped rather than orphaned to PPID 1 — three survived here, the oldest for 2d16h
- [ ] `loom-daemon-watchdog` also shows not-running with exit 1 on this host; it should have caught the daemon's death and did not (see 2AMLogic/2am#15)

## Note for anyone diagnosing this by hand

`pgrep -fl loom-daemon` prints the full command line and would have shown the
`/T/tmp.*/` path immediately. `pgrep -f` alone prints only a PID, which is what
made this invisible. Prefer the `-l` form when checking by hand.

## Related

- #5131 — label-blind `pgrep` tier; same root, opposite consequence
- 2AMLogic/2am#15 — the watchdog detects daemon death but never recovers it; here it did not even report it
- 2AMLogic/2am#47 — robb-pro's other divergences (unsupervised debug safehoused)

---

### Scope corrections (verified against `95c1529f`, 2026-08-06 — re-verify before relying on these if this issue sits a while)

- **AC #2's `pgrep -f safehoused` claim does not match what's in the tree.** The only `safehoused` process probe in this repo is `loom-daemon/src/fleet/status.rs:344`, and it uses `pgrep -x safehoused` (exact process-*name* match) — not `-f` (full-cmdline substring match) — as a **last-resort fallback after** a socket check (`$SAFEHOUSED_SOCKET` or `~/.safehouse/safehoused.sock`) that is tried first. `-x` matches the kernel-reported `comm` name, so a fixture invoked as `bash /tmp/.../safehoused` would generally show up as `bash`, not `safehoused`, unless the fixture explicitly renames its own process (e.g. via `exec -a`) — a materially smaller blast radius than the `loom-daemon` `-f` case this issue is about. Builder: still worth a look for consistency, but don't budget this as "the same bug, second instance" — verify whether any fixture actually renames itself before treating it as broken.
- **The watchdog's own liveness detection is not the `pgrep -f loom-daemon` bug.** `defaults/scripts/cli/loom-daemon-watchdog.sh` already resolves liveness via `LOOM_PID_FILE` / `launchctl print` / the systemd unit (see `resolve_pid_file`, ~line 481, and the launchd/systemd branch, ~line 540) — it does not use `pgrep -f loom-daemon` anywhere. AC #5 ("watchdog should have caught the daemon's death and did not") is therefore a **separate defect**: the watchdog *launchd job itself* (`com.rjwalters.loom-daemon-watchdog`) was not running (exited 1) on this host, so it never got a chance to run its (already-correct) detection logic. That's the failure this issue's own "Related" section already routes to 2AMLogic/2am#15 (external repo) — AC #5 as literally worded belongs there, not in this repo's fix. Suggest narrowing AC #5 to just "confirm no in-repo liveness check for the *watchdog job itself* uses a forgeable `pgrep -f` pattern" and leave the "why did the watchdog job die" question to the linked external issue.

### Verified `pgrep -f loom-daemon` call sites (as of `95c1529f`, 2026-08-06)

Comprehensive `grep -rn` for `pgrep -f.*loom-daemon` across the tree, beyond what the issue body itself cites:

- `scripts/stop-daemon.sh:12` — `DAEMON_PID=$(pgrep -f "loom-daemon" | head -1 || true)`. Unscoped, top-level dev convenience script (not under `defaults/scripts/`) — has the exact forgeable shape this issue describes.
- `scripts/start-daemon.sh:33` — `pgrep -P "$CARGO_PID" -f "target/debug/loom-daemon" || pgrep -f "target/debug/loom-daemon" | head -1`. The first branch is scoped to a specific parent PID; the bare fallback (`pgrep -f "target/debug/loom-daemon"`) is the same forgeable shape, though the `target/debug/` path segment narrows it somewhat versus a bare `loom-daemon` match.
- `defaults/scripts/cli/loom-daemon-stop.sh:408` — `pgrep -f '(^|/)loom-daemon$'`, already explicitly labeled `# Last-resort process-name match. This tier is LABEL-BLIND` in-file and gated behind `[[ "$LAUNCHD_LABEL" == "$DEFAULT_LAUNCHD_LABEL" ]]`. This is the tier #5131's comment thread already analyzed in depth (see that issue's corrected analysis) — it is a known, already-narrowed last-resort fallback, not an unaddressed instance. Anchored with `(^|/)...$` (not a bare substring), which is stricter than the two sites above but still name-forgeable by a fixture whose path literally ends in `/loom-daemon`.
- No production doc recommends `pgrep -f loom-daemon`/`pgrep -f safehoused` as a health check (`grep -rln "pgrep -f.*loom-daemon" --include="*.md" .` → no hits); mentions in `.md` files are all in `docs/design/`, `defaults/docs/daemon-reference.md`, and `defaults/.claude/commands/loom/sweep.md` discussing the *stop script's* known label-blind tier (same #5131 context), not proposing pgrep as a liveness/health signal.

### Verified test fixtures naming their mock binary literally `loom-daemon` (leak-risk for AC #3/#4)

`grep -rn 'cat > .*loom-daemon\|touch.*loom-daemon' --include="*.sh" .` turned up these (all under `$TMPDIR`-scoped dirs with trap-based cleanup — the incident shows that cleanup is not reliable in practice):

- `defaults/scripts/tests/test-check-duplicate.sh:104` — `cat > "$STUB_DIR/loom-daemon"`
- `defaults/scripts/tests/test-loom-status.sh:111,125,138,148` — four separate `cat > "$dir/loom-daemon"` fixtures
- `defaults/scripts/tests/test-gh-cached.sh:284` — `cat > "$FAKE_DAEMON_DIR/loom-daemon"`
- `defaults/scripts/tests/test-loom-daemon-update.sh:1013` — `cat > "$DECOY_DIR/loom-daemon"` (an intentional decoy fixture for a different test — renaming this one needs care to not break the test it's decoying)
- `defaults/scripts/tests/test-install-lock.sh:321` and `defaults/scripts/tests/test-install-reinstall-safety.sh:222,224` write to `target/release/loom-daemon` — these are simulating the **real build output path** (Cargo always names its binary `loom-daemon`), so renaming the binary name itself is not an option there; those two are out of scope for AC #3's "distinct fixture name" fix and need a different mitigation (e.g. stronger cleanup guarantees only) if included at all.

This list is a floor, not a ceiling — re-run the greps above against the build base before treating it as exhaustive, since new tests may land between now and Builder pickup.

### Affected Files

- `scripts/stop-daemon.sh` — replace the unscoped `pgrep -f "loom-daemon"` liveness/target check with a PID-file or launchd-state check
- `scripts/start-daemon.sh` — narrow or replace the bare `pgrep -f "target/debug/loom-daemon"` fallback
- `defaults/scripts/tests/test-check-duplicate.sh`, `test-loom-status.sh`, `test-gh-cached.sh` — rename mock binary fixtures from `loom-daemon` to a distinct name (e.g. `loom-daemon-mock`) so a leak can't satisfy a production check; verify each test's assertions still target the renamed binary
- `defaults/scripts/tests/test-loom-daemon-update.sh` — the `$DECOY_DIR/loom-daemon` fixture is intentionally named to test the decoy-matching behavior itself; evaluate case by case whether renaming it defeats the test's purpose
- Fixture cleanup path (wherever these tests' `trap`/teardown logic lives, e.g. shared helpers under `defaults/scripts/tests/lib/`) — investigate why three fixtures survived as PPID-1 orphans instead of being reaped, and harden it
- `defaults/scripts/cli/loom-daemon-stop.sh:408` — already gated/labeled; confirm during implementation whether it needs further narrowing or is sufficiently addressed by #5131's prior work (likely just needs a cross-reference, not a code change)

### Test Plan

- [ ] Manual verification: reproduce the incident shape — start a background `bash`-based fixture literally named `loom-daemon` under `$TMPDIR`, orphan it (detach from its parent), then run each fixed liveness check and confirm it reports "not running" / "absent" rather than matching the fixture
- [ ] Automated tests: extend `defaults/scripts/tests/test-loom-daemon-stop.sh`'s existing decoy-process pattern (see its "Decoy-process test (#4078)" section, ~line 247) to cover `scripts/stop-daemon.sh` / `scripts/start-daemon.sh` with a decoy fixture named exactly `loom-daemon`
- [ ] Automated tests: for each renamed test fixture, confirm the renamed test still passes and still exercises the same code path (i.e. the rename didn't silently disable the fixture)
- [ ] Edge cases: a fixture that leaks mid-suite (simulate a killed test run) should not survive as an orphaned process after a full suite run — verify via a repeated-invocation / trap-failure simulation if the harness supports it

rjwalters:
## Summary

`pgrep -f loom-daemon` is a full-command-line substring match, not a liveness
check: any process whose command line contains the string `loom-daemon` —
including a leaked test fixture bash script literally named `loom-daemon`
under `$TMPDIR` — satisfies it forever. Per the incident in #5548, this kept
a real daemon looking healthy for 66 minutes after it actually died.

`pgrep -x` matches the kernel-reported process *name* (comm) exactly instead.
A fixture invoked as `bash /path/to/loom-daemon` reports comm `bash`, not
`loom-daemon`, so `-x "loom-daemon"` never matches it — while the real,
compiled `loom-daemon` binary's comm still is `loom-daemon`, so the
narrowing is behavior-preserving for the real target.

## Changes

- **`scripts/stop-daemon.sh`** — the unscoped `pgrep -f "loom-daemon"`
  no-PID-file fallback now uses `pgrep -x "loom-daemon"`.
- **`scripts/start-daemon.sh`** — the bare `pgrep -f "target/debug/loom-daemon"`
  fallback (used when `cargo run`'s child has already reparented away from
  `$CARGO_PID` by the time we look) now uses `pgrep -x "loom-daemon"`. The
  `pgrep -P "$CARGO_PID" -f ...` primary branch is unchanged — it's already
  scoped to `$CARGO_PID`'s own children and cannot match an unrelated
  process regardless of matcher.
- **Test fixture renames** (`loom-daemon` → `loom-daemon-mock`) in
  `test-loom-status.sh`, `test-gh-cached.sh`, and
  `test-loom-daemon-watchdog.sh` — all three resolve their mock daemon
  binary via `LOOM_DAEMON_BIN` (an explicit path), never a bare PATH-name
  lookup, so the rename is behavior-preserving. `test-loom-daemon-watchdog.sh`
  was not in the curator's original verified list (its stub uses
  `printf ... > file`, not `cat > file`, so the curator's grep missed it) but
  is arguably the closest match to the incident's actual process shape: its
  `hang` mode backgrounds an infinite `while true; do sleep 1; done` loop
  under the literal name `loom-daemon`.
- **`scripts/test-daemon-liveness.sh`** (new, wired into `build-gate.sh`) —
  spawns a decoy `loom-daemon` bash script and asserts the fixed matchers do
  not mistake it for the real daemon. Test 3 (which exercises
  `stop-daemon.sh`'s real kill-or-not decision) stubs `pgrep` on `PATH` so it
  can never reach the live process table — this dev host runs a real
  production `loom-daemon` (PID with comm exactly `loom-daemon`), so
  invoking the unstubbed script directly would have risked killing it.
- **`defaults/scripts/cli/loom-daemon-stop.sh`** — added a cross-reference
  comment (no code change) to its already-gated, already-narrowed
  `pgrep -f '(^|/)loom-daemon$'` last-resort tier, noting the residual
  #5548-shaped gap and pointing at where the actual fixes live.

## Scope items evaluated and deliberately left unchanged

- **`test-check-duplicate.sh`'s `loom-daemon` fixture** — NOT renamed.
  `check-duplicate.sh` resolves it via a bare `command -v loom-daemon` PATH
  lookup with no `LOOM_DAEMON_BIN`-style override, so the literal name is
  load-bearing; renaming it would break the test. It poses no #5548-shaped
  risk in practice since it's invoked synchronously (`exit 1`, never
  backgrounded) — even a leaked copy on disk can never appear as a running
  process for a `pgrep`-style check to match.
- **`test-loom-daemon-update.sh`'s `DECOY_DIR/loom-daemon` fixture** — NOT
  renamed, per the curator's own suggestion: it exists specifically to test
  `loom-daemon-stop.sh`'s label-blind pgrep fallback, so it must be named
  `loom-daemon` to have anything to (not) match. It's already reaped via
  `bg_proc_track`/`bg_proc_reap` (EXIT/INT/TERM traps).
- **AC #2's `pgrep -f safehoused` claim** — confirmed incorrect, matching
  the curator's scope correction. `loom-daemon/src/fleet/status.rs:344` uses
  `pgrep -x safehoused` (exact match, not `-f`) as a last-resort fallback
  *after* a socket check. Grepped the whole tree for `exec -a` (a fixture
  renaming its own process to fake a `comm` match) — no hits — so no fixture
  could forge this narrower check either. No code change made here.
- **`loom-daemon-stop.sh:408`'s gated `pgrep -f '(^|/)loom-daemon$'` tier** —
  left as a comment-only cross-reference. It's already gated behind a
  default-launchd-label check and already covered by
  `test-loom-daemon-update.sh`'s decoy regression test (#4078). It does have
  a residual gap (a leaked fixture whose full path ends in `/loom-daemon`
  still satisfies the anchored regex) but this repo's own #5131 analysis
  already treats it as an intentional, narrowly-scoped last resort — not an
  unaddressed instance of this bug.
- **Fixture cleanup/teardown hardening (AC #6)** — investigated but not
  refactored. `lib/bg-proc-trap.sh` (from #4773) already provides
  `bg_proc_track`/`bg_proc_reap` with EXIT/INT/TERM traps across the
  higher-risk suites (`test-loom-daemon-stop.sh`, `test-loom-daemon-update.sh`,
  `test-loom-daemon-watchdog.sh`) and already documents its own residual gap:
  a `SIGKILL` of the test runner (or a host crash/sleep) bypasses every trap,
  which is a hard bash/POSIX limitation, not a fixable code gap. That's the
  most plausible root cause of the incident's 3 long-orphaned processes. The
  fixture renames above are the actual mitigation for that residual case:
  even a fixture that survives every trap can no longer forge a production
  liveness check once it isn't named `loom-daemon`.
- **AC #5 (watchdog job liveness)** — per the curator's scope correction,
  `loom-daemon-watchdog.sh` already resolves liveness via `LOOM_PID_FILE` /
  `launchctl print` / the systemd unit, never `pgrep -f loom-daemon`.
  Confirmed via `grep -n pgrep defaults/scripts/cli/loom-daemon-watchdog.sh`
  (no hits). "Why did the watchdog job itself die" routes to the linked
  external issue (2AMLogic/2am#15), out of scope here.

## Test plan

- `bash scripts/test-daemon-liveness.sh` — new suite, 5/5 passing.
- `bash defaults/scripts/tests/test-loom-status.sh` — 49/49 passing.
- `bash defaults/scripts/tests/test-gh-cached.sh` — 33/33 passing.
- `bash defaults/scripts/tests/test-check-duplicate.sh` — 133/133 passing
  (unchanged, confirms the fixture-kept-as-is decision doesn't regress it).
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 114/114
  passing.
- `bash defaults/scripts/tests/test-loom-daemon-stop.sh` — 38/38 passing
  (unchanged, confirms the comment-only edit doesn't regress it).
- `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 254/254 passing
  (unchanged, confirms the comment-only edit doesn't regress it).
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` — confirms the
  new suite is correctly excluded (wired directly in `build-gate.sh`
  instead) with no manifest violation.
- Manually confirmed no leaked `loom-daemon`/`loom-daemon-mock` processes
  survive any of the above runs (`pgrep -af`).

Closes #5548

### DIFF
--- a/defaults/scripts/build-gate.sh
+++ b/defaults/scripts/build-gate.sh
@@ -34,6 +34,11 @@
 #   - bash scripts/test-migrate-consumer.sh (#5276) covers scripts/install/
 #     migrate-consumer.sh (Epic #3835 Phase 6) against throwaway git fixtures
 #     -- no network, no real daemon, sub-second.
+#   - bash scripts/test-daemon-liveness.sh (#5548) regression-tests
+#     scripts/stop-daemon.sh's and scripts/start-daemon.sh's daemon-liveness
+#     pgrep matcher against a decoy fixture literally named `loom-daemon` --
+#     no real daemon build, PATH-stubs `pgrep` for the one case that would
+#     otherwise touch the live process table, sub-second.
 #   - mcp-loom (TypeScript) is intentionally EXCLUDED: it needs npm install/ci
 #     in a fresh worktree (no guaranteed warm node_modules), which adds
 #     unpredictable latency to a gate that also runs once per PR. CI still
@@ -163,6 +168,9 @@ bash scripts/test-installer.sh
 
 echo "[build-gate] bash changelog generator suite"
 bash scripts/test-changelog.sh
+
+echo "[build-gate] bash daemon-liveness pgrep suite (#5548)"
+bash scripts/test-daemon-liveness.sh
 
 echo "[build-gate] bash install --local/--gitignore mode suite"
 bash scripts/test-install-local-mode.sh
--- a/defaults/scripts/cli/loom-daemon-stop.sh
+++ b/defaults/scripts/cli/loom-daemon-stop.sh
@@ -404,6 +404,15 @@ if [[ -z "$pid" ]] || ! kill -0 "$pid" 2>/dev/null; then
     # label-blind kill would violate that scoping and contradict the #4054
     # label-scoped-stop discipline. With the default label we keep the fallback
     # as a genuine lost-PID-file recovery path.
+    # #5548 cross-reference: this tier's `pgrep -f` is still forgeable by a
+    # leaked test fixture whose path literally ends in `/loom-daemon` (the
+    # exact shape #5548 hit) -- the `(^|/)...$` anchor narrows it versus a
+    # bare substring match, but does not close that gap. No code change here:
+    # this is the intentionally-last-resort tier, already gated behind the
+    # default-label check above and already covered by a decoy regression
+    # test (test-loom-daemon-update.sh's suite-level decoy, #4078). #5548's
+    # actual fixes are in scripts/stop-daemon.sh, scripts/start-daemon.sh, and
+    # the renamed `loom-daemon-mock` test fixtures elsewhere in this repo.
     if [[ "$LAUNCHD_LABEL" == "$DEFAULT_LAUNCHD_LABEL" ]] && command -v pgrep >/dev/null 2>&1; then
         pid=$(pgrep -f '(^|/)loom-daemon$' 2>/dev/null | head -n1 || true)
     fi
--- a/defaults/scripts/tests/ci-excluded.txt
+++ b/defaults/scripts/tests/ci-excluded.txt
@@ -23,3 +23,4 @@ scripts/test-changelog.sh  Already wired directly in build-gate.sh (bash scripts
 scripts/test-install-local-mode.sh  Already wired directly in ci.yml's "Installer local/gitignore mode" step and build-gate.sh; excluded here to avoid double-running it.
 scripts/test-installer.sh  Already wired directly in ci.yml's "Installer Integration Tests" job and build-gate.sh; excluded here to avoid double-running it.
 scripts/test-migrate-consumer.sh  Already wired directly in ci.yml's "migrate-consumer" step and build-gate.sh; excluded here to avoid double-running it.
+scripts/test-daemon-liveness.sh  Already wired directly in build-gate.sh (bash scripts/test-daemon-liveness.sh, #5548); excluded here to avoid double-running it.
--- a/defaults/scripts/tests/test-check-duplicate.sh
+++ b/defaults/scripts/tests/test-check-duplicate.sh
@@ -100,7 +100,18 @@ trap 'rm -rf "$STUB_DIR" 2>/dev/null || true' EXIT
 # check-duplicate.sh's `loom-daemon --version` probe fails and it falls back
 # to the `gh` stub below (byte-for-byte the documented fallback path,
 # defaults/scripts/check-duplicate.sh). Keeps this test independent of
-# whether a real loom-daemon happens to be installed on the host. ---
+# whether a real loom-daemon happens to be installed on the host.
+#
+# NOT renamed per #5548's "test fixtures should not be named `loom-daemon`"
+# fix, unlike the fixtures in test-loom-status.sh / test-gh-cached.sh /
+# test-loom-daemon-watchdog.sh: check-duplicate.sh resolves this binary via a
+# bare `command -v loom-daemon` PATH lookup with no LOOM_DAEMON_BIN-style
+# override, so the literal name is load-bearing here -- renaming it would
+# stop check-duplicate.sh from finding it at all, defeating the test. This
+# poses no #5548-shaped liveness-check risk in practice: the stub is invoked
+# synchronously (`exit 1`, no backgrounding, no loop), so even a leaked copy
+# left on disk after a failed trap can never appear as a running process for
+# a `pgrep`-style liveness check to match. ---
 cat > "$STUB_DIR/loom-daemon" <<'STUB'
 #!/usr/bin/env bash
 exit 1
--- a/defaults/scripts/tests/test-gh-cached.sh
+++ b/defaults/scripts/tests/test-gh-cached.sh
@@ -281,7 +281,14 @@ FAKE_DAEMON_DIR="$TMP_ROOT/daemon"
 mkdir -p "$FAKE_DAEMON_DIR"
 # A fake loom-daemon: `forge <e> list --cached` with --json succeeds and emits a
 # sentinel; any shape without --json exits 3 (decline), mirroring the real one.
-cat > "$FAKE_DAEMON_DIR/loom-daemon" <<'FAKE'
+#
+# Named `loom-daemon-mock`, NOT `loom-daemon` (#5548): gh-cached's
+# locate_loom_daemon() checks $LOOM_DAEMON_BIN (set explicitly below) BEFORE
+# falling back to a bare `which("loom-daemon")` PATH lookup, so this fixture
+# never needs the literal name to be resolved. A distinct name means a leaked
+# fixture can never forge a production `pgrep -f loom-daemon`-style liveness
+# check the way the incident describes.
+cat > "$FAKE_DAEMON_DIR/loom-daemon-mock" <<'FAKE'
 #!/usr/bin/env bash
 if [[ "$1" == "forge" && "$3" == "list" ]]; then
     if printf '%s\n' "$@" | grep -q -- '--json'; then
@@ -292,13 +299,16 @@ if [[ "$1" == "forge" && "$3" == "list" ]]; then
 fi
 exit 3
 FAKE
-chmod +x "$FAKE_DAEMON_DIR/loom-daemon"
+chmod +x "$FAKE_DAEMON_DIR/loom-daemon-mock"
 
-# Runner with the fake daemon on PATH and the ETag layer ENABLED.
+# Runner with the fake daemon pinned via LOOM_DAEMON_BIN and the ETag layer
+# ENABLED. FAKE_DAEMON_DIR is still added to PATH as a belt-and-suspenders
+# safety net for any code path that falls through to a PATH search, but
+# LOOM_DAEMON_BIN is what locate_loom_daemon() actually resolves first.
 ghc_etag() {
     PATH="$STUB_DIR:$FAKE_DAEMON_DIR:$PATH" \
     GH_CACHE_DIR="$CACHE_DIR" \
-    LOOM_DAEMON_BIN="$FAKE_DAEMON_DIR/loom-daemon" \
+    LOOM_DAEMON_BIN="$FAKE_DAEMON_DIR/loom-daemon-mock" \
       "$GH_CACHED" "$@"
 }
 
--- a/defaults/scripts/tests/test-loom-daemon-update.sh
+++ b/defaults/scripts/tests/test-loom-daemon-update.sh
@@ -1009,6 +1009,16 @@ export LOOM_PROVISION_ALLOW_SCRIPT=1
 # any test regresses into one, this decoy dies and the final assertion fails.
 # Spawned OUTSIDE $BASE_WORKDIR's path so the trap's `pkill -f "$BASE_WORKDIR"`
 # does not sweep it; killed explicitly below.
+#
+# Deliberately NOT renamed per #5548's "test fixtures should not be named
+# `loom-daemon`" fix: the whole point of this fixture is to BE a process
+# named `loom-daemon` so the label-blind pgrep tier it decoys has something
+# to (not) match -- renaming it would defeat the test. It is tracked via
+# bg_proc_track/bg_proc_reap (EXIT/INT/TERM traps) like every other
+# background fixture in this suite, so it does not additionally widen
+# #5548's "orphaned past all cleanup" risk beyond what SIGKILL of the test
+# runner already allows for any tracked fixture (a hard, documented
+# bash/POSIX limit -- see lib/bg-proc-trap.sh).
 DECOY_DIR="$(mktemp -d)"
 cat > "$DECOY_DIR/loom-daemon" <<'EOF'
 #!/usr/bin/env bash
--- a/defaults/scripts/tests/test-loom-daemon-watchdog.sh
+++ b/defaults/scripts/tests/test-loom-daemon-watchdog.sh
@@ -188,7 +188,13 @@ EOF
 # The watchdog shells out to the resolved `loom-daemon` binary for its bounded
 # in-band probe, so the probe is driven entirely by a stub binary pinned via
 # LOOM_DAEMON_BIN — no real daemon, no real socket, no `timeout` dependency on
-# the outcome. Each mode reproduces one shape the classifier must handle:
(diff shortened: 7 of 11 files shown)


rjwalters:
LGTM — approving.

**Root-cause fix verified correct.** `pgrep -x` matches the kernel-reported process *name* (`comm`), not the full command line. A leaked bash-script fixture named `loom-daemon` is invoked as `bash /path/to/loom-daemon`, so its `comm` is `bash` — `pgrep -x "loom-daemon"` never matches it, while the old `pgrep -f "loom-daemon"` matched it forever (exactly the incident in #5548). Confirmed live on this host: the real compiled binary's `comm` is genuinely `loom-daemon` (`cat /proc/<pid>/comm` → `loom-daemon`; `Cargo.toml`'s `name = "loom-daemon"`), so the narrowing is behavior-preserving for the actual target and closes the forgery gap for the leaked-fixture shape.

**Scope matches the curated issue.** Both flagged call sites (`scripts/stop-daemon.sh`'s unscoped fallback, `scripts/start-daemon.sh`'s bare fallback) are fixed. `start-daemon.sh`'s primary `pgrep -P "$CARGO_PID" -f ...` branch is correctly left alone — it's already scoped to a specific parent PID and was never forgeable. The deliberately-deferred items (`test-check-duplicate.sh`'s and `test-loom-daemon-update.sh`'s `loom-daemon`-named fixtures kept as-is; the `safehoused` AC corrected to not-actually-broken; `loom-daemon-stop.sh`'s already-gated last-resort tier left as a comment-only cross-reference) all have specific, verifiable rationale in the PR body, consistent with the curator's own scope corrections on the issue.

**Test coverage.** New `scripts/test-daemon-liveness.sh` reproduces the incident's exact process shape (a backgrounded bash script literally named `loom-daemon`) and: (1) proves `pgrep -x` doesn't match the decoy, (2) proves `pgrep -f` would have (confirming the repro is faithful), (3) exercises `stop-daemon.sh`'s real kill-or-not decision against the decoy via a scoped `pgrep` stub (safe on a host running a real production daemon), (4) a lighter static check on `start-daemon.sh`. Correctly wired into `build-gate.sh` and `ci-excluded.txt`.

**Verification performed in a worktree, diffed against `origin/main` (local `main` ref was stale, correctly excludes the unrelated `#5551`/`#5553` commits already on `origin/main`):**
- `bash scripts/test-daemon-liveness.sh` — 5/5 pass
- `bash defaults/scripts/tests/test-loom-status.sh` — 49/49 pass
- `bash defaults/scripts/tests/test-gh-cached.sh` — 33/33 pass
- `bash defaults/scripts/tests/test-check-duplicate.sh` — 133/133 pass
- `bash defaults/scripts/tests/test-loom-daemon-watchdog.sh` — 114/114 pass
- `bash defaults/scripts/tests/test-loom-daemon-update.sh` — 254/254 pass
- `bash defaults/scripts/tests/test-loom-daemon-stop.sh` — 38/38 pass
- `bash defaults/scripts/tests/check-ci-suite-manifest.sh` — OK, no manifest violation
- Confirmed no leaked `loom-daemon`/`loom-daemon-mock` processes survive any of the runs (`pgrep -af`)

Fixture renames (`loom-daemon` → `loom-daemon-mock`) checked across all call sites in `test-loom-status.sh`, `test-gh-cached.sh`, and `test-loom-daemon-watchdog.sh` — consistent, no stray unrenamed references.

`mergeStateStatus` is CLEAN/MERGEABLE. `gh pr checks` reports no checks configured for this branch — consistent with other recently-merged PRs in this repo (e.g. #5551), not specific to this PR, so not treated as a blocker.

No issues found. Approving.

Issue #4285, opened 2026-07-28
Installer's workspace package.json stub carries version 1.0.0, confusing version-detection tooling


rjwalters:
## Context
Loom's installer writes a root `package.json` stub into consumer repos:

```json
{ "name": "loom-workspace", "version": "1.0.0", "private": true, "scripts": { ... } }
```

In rjwalters/claude-monitor (app version 1.15.0, versions kept in Swift source + build script) this stub's `1.0.0` is a decoy for anything that treats a root `package.json` as the version source — `/repo:release` matches npm on the mere presence of `package.json`, and `/loom:bump` lists root `package.json` as detection source #2. Surfaced during a hygiene audit (2026-07-28).

## Suggested acceptance criteria
- [ ] Installer omits the `version` field from the stub (npm allows this for `private: true` packages), or uses an obviously-non-project sentinel like `0.0.0-loom-workspace`
- [ ] `/loom:bump` skips a root package.json whose `name` is `loom-workspace`
- [ ] Existing consumer repos: reinstall/resync removes or sentinels the stub version

rjwalters:
## Curator Enhancement (2026-07-28)

### Code map (verified against origin/main)

- **The stub itself**: `defaults/package.json` — `"name": "loom-workspace", "version": "1.0.0"`. The `version` field serves no runtime purpose: nothing in Loom reads the *consumer* stub's version (installer version metadata comes from the **Loom repo's own** root `package.json` via `scripts/install-loom.sh:867` and `defaults/scripts/resync-installed.sh:389`, both of which read `SOURCE_ROOT`/`LOOM_ROOT`, not the consumer workspace).
- **Copy site**: `loom-daemon/src/init/scaffolding.rs:961-968` — `loom-daemon init` copies `defaults/package.json` → workspace root **only if the target lacks one**, and never overwrites (even in force mode). So shipping a fixed stub automatically fixes *new* installs; it does nothing for existing consumers (AC 3 needs a separate mechanism).
- **Manifest carve-out**: `scripts/install/manifest.sh:200-207` — `package.json` is registered as Loom-owned only when the consumer didn't already have one. Consumers that received the stub *do* have it manifest-tracked, which is the hook AC 3 can use.
[...]

PR #5561, opened 2026-08-07
fix(resync): restamp .loom/CLAUDE.md's version header on resync

### LINKED ISSUE #5559: resync-installed.sh never restamps the vendored CLAUDE.md version header (metadata 0.18.0 vs header 0.16.0)
## Context

In a managed repo (rjwalters/kicad-tools), the vendored `.loom/CLAUDE.md` guide header drifts from the actual installed version, and `resync-installed.sh` does not correct it.

Observed sequence:

- Install metadata (`.loom/install-metadata.json`): `loom_version: 0.18.0`, `loom_commit: a138f09c`, `last_resync: 2026-08-06` — resync run today from a Loom source clone at current `main` (`a138f09c`).
- Resync reported `97 file(s) updated, 242 unchanged, 0 skipped`.
- `.loom/CLAUDE.md` still reads:

  ```
  **Loom Version**: 0.16.0
  ...
  Last updated: 2026-07-29
  ```

So the guide's human-facing version stamp is two minor versions behind what is actually installed, and repeated resyncs (2026-08-04 and 2026-08-06) never touch it. `install.sh --yes --confirm-reinstall` was also tried; its idempotency check ("Loom v0.18.0 is already installed... use --force or --clean") exits before restamping, and `--force` is not an accepted `install.sh` flag (that hint appears to come from a nested script's vocabulary — a second, smaller confusion worth aligning while in there).

## Suspected cause

The installer generates `.loom/CLAUDE.md` from a template with version substitution at initial install time, but the resync path treats it as either excluded or compares it against something that never carries the new version stamp — so the stamped header is only ever written once, at first install.

- [ ] `resync-installed.sh` restamps the `**Loom Version**` and `Last updated` lines in the vendored `.loom/CLAUDE.md` (or regenerates the file from the template) to match the source it resynced from
- [ ] A managed repo whose metadata and guide header disagree converges after one resync
- [ ] The idempotency-check hint in `install.sh` names flags `install.sh` actually accepts

---

### Root cause (verified against `loom` @ `768b37bc`, 2026-08-06 — re-verify against current `main` before relying on these line numbers)

Confirmed by reading source, not by reproducing on a live host — treat the mechanism as verified, the exact line numbers as a snapshot:

1. **`.loom/CLAUDE.md` is explicitly out of scope for `resync-installed.sh` by design**, not by omission. `defaults/scripts/resync-installed.sh:98-105` documents it in the "EXPLICITLY OUT OF SCOPE" header comment:
   > `CLAUDE.md` - repo-customized at install; needs managed-section markers ... (same as CLAUDE.md; #4479). Its full-guide sibling `.loom/CLAUDE.md` is regenerated by install scaffolding from `defaults/.loom/AGENTS.md`, not by resync — same posture as `.loom/CLAUDE.md`.

   So the "or regenerates the file from the template" half of the original suggested AC would cut against this documented design intent unless the managed-section-markers work (referenced as #4479 in that same comment) is done first or ruled irrelevant. **#4479 is closed** (verified via `gh issue view 4479`) but titled "AGENTS.md single-source instruction anchor" — it does not appear to be the CLAUDE.md managed-section-markers design the comment references; worth a quick check by whoever picks this up whether that closed issue actually satisfies the precondition or whether the comment is itself stale.

2. **`resync-installed.sh` *does* restamp `.loom/install-metadata.json`'s `loom_version`** on every successful non-dry-run (`defaults/scripts/resync-installed.sh:48-50`, confirmed at `defaults/scripts/resync-installed.sh` — the specific line numbers for the jq/python3 restamp logic are further down the file, search `install-metadata.json` re-stamp). This is exactly what the issue's own "Observed sequence" shows: `install-metadata.json` reads 0.18.0 after resync while `.loom/CLAUDE.md` stays at 0.16.0.

3. **This combination is what defeats the reinstall idempotency check the issue also hit.** `scripts/install-loom.sh`'s pre-flight idempotency check (`scripts/install-loom.sh:1134-1180`) determines "already installed at the current version" using a layered version read:
   - Source 1 (preferred): `.loom/install-metadata.json`'s `loom_version` field (`scripts/install-loom.sh:1141-1149`)
   - Source 2 (fallback, only checked if Source 1 is empty): `.loom/CLAUDE.md`'s `**Loom Version**` line (`scripts/install-loom.sh:1151-1154`)
   - Source 3 (legacy fallback): root `CLAUDE.md`

   Because Source 1 always wins when present, and `resync-installed.sh` keeps `install-metadata.json` current even though it explicitly skips `.loom/CLAUDE.md`, the idempotency check reads `loom_version: 0.18.0`, matches the running installer's own version, prints "Loom v0.18.0 is already installed... Use --force to reinstall or --clean for a fresh install" (`scripts/install-loom.sh:1169-1170`), and `exit 0`s (`scripts/install-loom.sh:1175`) — **before Source 2 is ever consulted and before any template regeneration runs.** The metadata file that resync keeps fresh is the same one that then masks the drift in the file resync deliberately leaves stale.

4. **The `--force` vocabulary mismatch is real and independently verified**, not just a hunch: `--force`/`-f` is a flag `scripts/install-loom.sh` genuinely accepts (`scripts/install-loom.sh:61-62`, help text at `:110`, gates the idempotency skip at `:1130`). The top-level wrapper `install.sh` that the issue reporter actually invoked has its own separate argument parser (`install.sh:891-974`) that does **not** include `-f`/`--force` in its `case` statement — the closest existing flags are `--confirm-reinstall`, `--dry-run`, `--allow-non-main-source`, `--allow-stale-target`. Passing `--force` to the top-level `install.sh` falls through to its catch-all (`install.sh:970-972`) and hard-errors with `Unknown flag: --force`. So the hint text that `scripts/install-loom.sh` prints ("Use --force...") is correct for itself but not directly usable on the entry point the issue reporter ran.

### Suggested acceptance criteria (supersedes the original three — see rationale below)

- [ ] Either (a) `resync-installed.sh` gains a narrowly-scoped restamp of just the `**Loom Version**` / `Last updated` header lines in `.loom/CLAUDE.md` (a targeted field edit like the existing `package.json` version-stub handling, #4285 — not a full-file resync, so it doesn't require the managed-section-markers design), or (b) the idempotency check in `scripts/install-loom.sh` stops trusting `install-metadata.json` alone and cross-checks `.loom/CLAUDE.md`'s stamp too before declaring "already installed" — whichever is chosen, state which explicitly, since they're alternative fixes for the same symptom, not both required.
- [ ] A managed repo whose `install-metadata.json` and `.loom/CLAUDE.md` header disagree converges (either by resync restamping the header, or by the idempotency check no longer being fooled into a silent no-op) within one resync or one reinstall attempt.
- [ ] `scripts/install-loom.sh`'s "Use --force to reinstall" hint (`scripts/install-loom.sh:1170`) is only surfaced in contexts where `--force` is actually a valid flag for the command the user ran — e.g. when reached via the top-level `install.sh`, either forward `--force` as a recognized (possibly pass-through) flag, or change the printed hint to name a flag `install.sh` accepts (`--confirm-reinstall` is the closest existing equivalent, though it triggers a full uninstall+reinstall rather than `install-loom.sh`'s in-place `--force` overwrite — confirm the two aren't actually different operations before conflating them in the hint text).

Rationale for the rewrite: the original ACs assumed the drift was a straightforward missed-restamp bug. Root-cause reading shows two independent, compounding causes (deliberate resync exclusion + a metadata-only idempotency check that gets fooled by the very field resync does keep current), so a fix needs to pick a lane rather than doing "restamp OR regenerate" as if they were equivalent — the "regenerate from template" option specifically depends on the managed-section-markers work that may or may not already be done via #4479.

### Affected Files

- `defaults/scripts/resync-installed.sh` - if AC option (a) is chosen: add a targeted restamp of `.loom/CLAUDE.md`'s version header lines, following the existing `package.json` `del(.version)`-style targeted-field-edit pattern (see the file's own header comment around line 40-46) rather than a full-file copy; update the "EXPLICITLY OUT OF SCOPE" comment block (lines 98-105) to reflect the new, narrower scope
- `scripts/install-loom.sh` - if AC option (b) is chosen: adjust the idempotency check (`:1134-1180`) to also compare `.loom/CLAUDE.md`'s stamp, not just `install-metadata.json`'s; either way, fix or remove the `--force` hint at `:1170` per the third AC
- `install.sh` - if the `--force` hint is to be made valid at the top-level entry point, add pass-through handling alongside the existing `--allow-non-main-source`/`--allow-stale-target` forwarding pattern (`install.sh:891-974`); otherwise just correct the hint text callers actually see
- `defaults/scripts/resync-installed.sh`'s own header comment (lines 48-50, 98-105) - keep in sync with whichever behavior change is made, since both are cited as authoritative scope statements elsewhere in this codebase

### Test Plan

- [ ] Manual verification: in a scratch managed repo, hand-edit `.loom/CLAUDE.md`'s `**Loom Version**` line to an older value while leaving `.loom/install-metadata.json` at the current version (reproducing the drifted state), then run `./.loom/scripts/resync-installed.sh` (or the chosen fix's entry point) and confirm the header converges
- [ ] Manual verification: reproduce the reported idempotency short-circuit — with `install-metadata.json` at the current version and `.loom/CLAUDE.md` stale, run `scripts/install-loom.sh` (no `--force`) and confirm it no longer silently reports "already installed" without addressing the stale header (per whichever AC option is chosen)
- [ ] Automated tests: add/extend coverage under the resync or installer test suite (check for existing `*.bats` or shell test coverage of `resync-installed.sh` / `install-loom.sh` idempotency paths before adding a new file) exercising the drifted-header-converges and idempotency-check paths
- [ ] Edge cases: confirm the fix does not regress `--dry-run` (previews the restamp without writing), the `.loom/resync-ignore` pin mechanism (a repo that intentionally pins `.loom/CLAUDE.md` via `resync-ignore`, if that becomes an option, should still be able to skip the new restamp), and repos where `.loom/CLAUDE.md` doesn't exist yet (pre-#4239 install layout)

rjwalters:
## Summary

Fixes the version-header drift described in #5559: `resync-installed.sh`
already re-stamps `.loom/install-metadata.json`'s `loom_version` on every
run, but deliberately left the vendored `.loom/CLAUDE.md`'s own
`**Loom Version**` / `Last updated` header lines untouched (its own
"EXPLICITLY OUT OF SCOPE" comment says a full-file regenerate needs the
still-pending managed-section-markers design). That let the two stamps
drift apart — and since `scripts/install-loom.sh`'s idempotency check
prefers `install-metadata.json` over the header, the drift stayed invisible
until inspected by hand and defeated the "already installed" fast-path check.

**Lane chosen: option (a)** — a narrowly-scoped restamp inside
`resync-installed.sh`, not option (b) (cross-checking `.loom/CLAUDE.md` in
`install-loom.sh`'s idempotency check). Rationale: the reported symptom
(header drift) is produced by the periodic resync itself, which is the
operation that's already run regularly (the "chore: resync installed Loom
surfaces" cadence) — restamping there converges the header on the very next
resync without touching the idempotency check's more complex reinstall/PR
logic. This also mirrors an existing pattern in the same file
(`resync_workspace_stub_version()`'s `del(.version)` targeted field edit,
#4285) rather than introducing a new mechanism.

## Changes

- `defaults/scripts/resync-installed.sh`: new `resync_claude_md_version_header()`
  — a targeted field edit (not a whole-file resync/regenerate) that restamps
  only the `**Loom Version**` and `Last updated` lines in `.loom/CLAUDE.md` to
  match the resolved source version/date. The `**Installation Date**` line
  (the original install date) is deliberately left untouched. Supports
  `--dry-run` preview and `.loom/resync-ignore` pinning (`.loom/CLAUDE.md`),
  and is a no-op when the file doesn't exist (pre-#4239 layout). Updated the
  file's own "EXPLICITLY OUT OF SCOPE" header comment and the
  `.loom/resync-ignore` convention doc-comment to describe the narrower new
  scope, and added `.loom/CLAUDE.md` to the dogfood-repo
  "resync-only dirt" allowlist used by the auto-suggested `git commit` hint.
- `scripts/install-loom.sh`: corrected the "Use --force to reinstall" hint
  printed by the idempotency check — it named a flag that `scripts/install-loom.sh`
  itself accepts, but most users reach this check through the top-level
  `install.sh` wrapper, whose arg parser does **not** accept `--force` and
  hard-errors `Unknown flag: --force` if followed literally. The hint now
  names both entry points explicitly and does not conflate `--force`
  (in-place overwrite, also enables PR auto-merge on the Full Install path)
  with `install.sh --confirm-reinstall` (uninstalls, then reinstalls — a
  heavier, destructive operation, not an equivalent shortcut).
- `defaults/scripts/tests/test-resync-installed.sh`: five new test groups
  (12j–12m) covering the restamp-converges, idempotent-rerun, missing-file,
  `--dry-run` preview, and `.loom/resync-ignore` pin cases for the new
  function.

- [x] EITHER (a) targeted restamp in `resync-installed.sh` OR (b) idempotency
      cross-check in `install-loom.sh` — **chose (a)**, stated above.
- [x] A managed repo whose `install-metadata.json` and `.loom/CLAUDE.md`
      header disagree converges within one resync — verified both by the
      automated test suite and by hand in a scratch repo (see Test Plan).
- [x] `--force` hint mismatch fixed in `scripts/install-loom.sh`'s idempotency
      check.

## Test Plan

- `bash defaults/scripts/tests/test-resync-installed.sh` — 147/147 passed
  (14 new assertions across groups 12j–12m).
- `bash scripts/test-installer.sh` — 179/179 passed (covers
  `install-loom.sh`, unaffected by the hint-text change).
- Manual verification in a scratch repo: hand-crafted `.loom/CLAUDE.md` with
  `**Loom Version**: 0.16.0` / `Last updated: 2026-07-29` while
  `install-metadata.json`/`package.json` reported `0.18.5`; ran
  `defaults/scripts/resync-installed.sh` once — header converged to
  `**Loom Version**: 0.18.5` / `Last updated: <today>`, while
  `**Installation Date**: 2026-04-21` (original install date) was left
  untouched.
- Edge cases verified by the new automated tests: `--dry-run` previews
  without writing; `.loom/resync-ignore` pin skips the restamp; a repo
  without `.loom/CLAUDE.md` (pre-#4239 layout) is a clean no-op, not an error.

Closes #5559


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-08-07T02:44:33Z) — standing down without reclaiming. Not stomping.

rjwalters:
Reviewed and approved.

**Correctness / design**
- `resync_claude_md_version_header()` in `defaults/scripts/resync-installed.sh` is a well-scoped, targeted field edit (mirrors the existing `resync_workspace_stub_version()`/`package.json` pattern) that restamps only the `**Loom Version**` and `Last updated` lines in a managed repo's `.loom/CLAUDE.md`, leaving `**Installation Date**` and body content untouched — matches the issue's chosen lane (a).
- Correctly no-ops on a missing `.loom/CLAUDE.md` (pre-#4239 layout), honors `--dry-run`, and honors `.loom/resync-ignore` pinning via `is_ignored ".loom/CLAUDE.md"`.
- `sed` field-replacement is anchored (`^\*\*Loom Version\*\*: .*` / `^Last updated: .*`) and writes via temp-file + `mv` (atomic, consistent with the rest of the script).
- The `**Installation Date**` vs `Last updated` distinction is deliberate and well-documented in the function's header comment — restamping `Last updated` on every resync (not just `Installation Date`) directly matches the drift the issue reported (both fields were stale).
- The "EXPLICITLY OUT OF SCOPE" header comment and `.loom/resync-ignore` convention doc-comment are updated to describe the new narrower scope, and `.loom/CLAUDE.md` was correctly added to the dogfood-repo "resync-only dirt" commit-hint allowlist.
- `scripts/install-loom.sh`'s corrected `--force` hint is accurate: verified `install.sh`'s own arg parser (lines ~891-974) does not accept `--force`/`--clean` (falls through to `Unknown flag:` on unrecognized flags), while `scripts/install-loom.sh` itself accepts both directly (`-f|--force`, `--clean`). The new hint correctly distinguishes the two entry points and doesn't conflate `--force` with `install.sh --confirm-reinstall`.

**Tests**
- `bash defaults/scripts/tests/test-resync-installed.sh` — 147/147 passed locally (new groups 12j–12m cover restamp-converges, idempotent-rerun, missing-file no-op, `--dry-run` preview, and `.loom/resync-ignore` pin).
- Verified independently that the function correctly leaves `**Installation Date**` and body text untouched, and reports idempotently on rerun.

**CI / merge state**
- All CI checks pass (`gh pr checks 5561` — Shellcheck, Docs/Defaults Parity, Installer Integration Tests, Shell Test Suites (hermetic), etc. all green; unrelated jobs correctly skip).
- `mergeStateStatus` is `CLEAN`.

**Minor/non-blocking observations** (not requesting changes):
- The `resync_claude_md_version_header` call site is a top-level unconditional call right after its definition (not inside a `main`), which matches the file's existing linear-script convention (`resync_workspace_stub_version` does the same) — consistent, not a new pattern.
- Shellcheck findings in the touched files are all pre-existing, outside the lines this PR changes.

LGTM.

Issue #5543, opened 2026-08-06
dashboard-deploy is pinned by a miniflare isolated-storage flake — the live Worker has not redeployed since 08-05T07:28Z


rjwalters:
## Context

`dashboard-deploy.yml`'s last run **failed**, and because `deploy` is gated on
`needs: test`, the live Worker has not been redeployed since. The failure is not
a product bug — it is a miniflare test-infrastructure flake.

Found while investigating the 11-hour pulse outage (#5523). Unrelated to that
outage's cause, but it means the dashboard has been pinned to an older build the
whole time, and nothing said so.

## The failure

Run 30985127198, job `Dashboard Tests (deploy gate)`, 2026-08-05T07:28Z:

```
Failed to pop isolated storage stack frame in test/redaction.test.ts's test
  "a public record's live-tail frame is delivered unredacted on the public route".
- AssertionError [ERR_ASSERTION]: Expected .sqlite, got
  /tmp/miniflare-…/d1/miniflare-D1DatabaseObject/9ba2b04…sqlite-shm
Error running worker: AssertionError: Isolated storage failed.

Tests  1 failed | 256 passed (257)
```

miniflare's isolated-storage teardown enumerates the D1 sqlite files and asserts
each ends in `.sqlite`. SQLite in WAL mode creates `-shm` and `-wal` sidecars
next to the database, and when one exists at teardown the assertion trips. It is
timing-dependent — whether the sidecar is still on disk when the frame pops —
which is exactly why 256 tests pass and one does not, and why it will pass on a
re-run.

## Why it is worth more than a re-run

The re-run fixes today. The structural problems stay:

1. **A flake can silently pin production.** `test` gating `deploy` is correct and
   should stay — the point of the gate is that a broken commit never reaches the
   live Worker. But an *infrastructure* flake is not a broken commit, and right
   now the two are indistinguishable to the pipeline.
2. **Nothing reports the staleness.** The workflow triggers on
   `paths: ["dashboard/**"]`, so after a failed run the next deploy waits for the
   next dashboard change — which may be days. There is no alert, and the
   dashboard keeps serving happily from the old build, so the only signal is
   someone opening the Actions tab.

## Suggested acceptance criteria

- [ ] The known miniflare isolated-storage flake is neutralized — pin/upgrade miniflare if a fixed release exists, or make the D1 teardown tolerate `-shm`/`-wal` sidecars
[...]

PR #5570, opened 2026-08-07
ci: delete dashboard-deploy.yml — the instance owns its own deploy now

### LINKED ISSUE #5567: dashboard-deploy.yml deploys one operator's instance from the mechanism repo — remove it now that the instance owns its deploy
## Context

`dashboard/docs/reference-deployment.md` in this repo states the principle:

> This repository intentionally does not track any specific operator's live
> deployment identity — Cloudflare account ID, D1 database ID, custom domain,
> CI secret names … all belong in *your own* infrastructure repo, not in this
> public mechanism repo.

That doc has been trimmed to a generic 14-line pointer and now honours it. **`.github/workflows/dashboard-deploy.yml` does not.** It deploys one specific operator's instance — `dashboard.2amlogic.com` — from this repo, holding that operator's Cloudflare credentials in this repo's Actions secrets.

The consuming instance has now adopted its own deploy path (`2AMLogic/2am#2`): a workflow that checks this repo out at a **pinned ref**, applies its own tracked wrangler overlay, and deploys. So this workflow's job has moved.

## Why this matters beyond tidiness

While both exist, **two repos deploy to the same Worker**, and they disagree about what should be live:

| | Trigger | Version deployed |
|---|---|---|
| this repo | every push to `main` touching `dashboard/**` | whatever `main` happens to be |
| the instance repo | a reviewed PR that bumps a pinned ref | a deliberately chosen commit |

So any merge here **silently overwrites whatever ref the instance pinned**, and the instance's deploy-of-record becomes a claim rather than a fact. The pin is advisory until this workflow is gone.

There is also a smaller correctness point: this workflow can only ever be green for one operator. Anyone else running this repo sees a permanently red required-looking job, or has to fork it out.

## Scope

1. Delete `.github/workflows/dashboard-deploy.yml`.
2. Delete these Actions secrets from this repo's settings (**operator step — cannot be done in a PR**):
   - `CLOUDFLARE_API_TOKEN`
   - `CLOUDFLARE_ACCOUNT_ID`
   - `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC`
3. Check `dashboard/docs/deploy-runbook.md` still stands on its own as the generic manual path once the automated one is gone, and mention that an instance repo may automate it — without naming any instance.

Out of scope: `dashboard/wrangler.toml`, the `dashboard/**` code, and `reference-deployment.md` (already correct).

## ✅ Sequencing gate — SATISFIED 2026-08-07, verifiable without private access

This issue was originally filed `loom:blocked` because deleting this workflow
before the successor was proven would leave **no automated deploy path for
either repo**. That gate is now met.

An automated pre-flight correctly declined to build against the original
wording, because it asked the reader to check `2AMLogic/2am#76` — a **private**
repo this repo's token cannot resolve. That was the right call on a gate it
could not verify. Restated below so it is checkable from the public internet
by anyone, with no access to that repo:

```sh
curl -s https://dashboard.2amlogic.com/api/version
# {"commit":"0fe402ac523a6ba73d036b0aefe12fee7fa2325e"}
```

`0fe402ac523a6ba73d036b0aefe12fee7fa2325e` is a commit **in this repository**
(`git cat-file -t` it locally). The instance now deploys this repo's
`dashboard/` from that pinned ref and stamps the SHA it deployed into the
Worker's `BUILD_COMMIT`, so the live site reporting it *is* the proof that the
successor deploy path works end to end.

- [x] successor workflow merged in the instance repo
- [x] a deploy from it ran green
- [x] `dashboard.2amlogic.com` serves that build, confirmed by the endpoint above

**Nothing further blocks this.** If a future reader wants to re-verify, re-run
the `curl` — the answer should be a SHA in this repo, not `unknown` and not a
branch name.

## Acceptance criteria

- [ ] `.github/workflows/dashboard-deploy.yml` deleted
- [ ] The three Cloudflare secrets removed from this repo's settings (operator)
- [ ] No file in this repo references `dashboard.2amlogic.com`, the 2AM account id, or `wrangler.2amlogic.toml` as *this repo's* deploy target — `grep -rn "2amlogic" --include='*.yml' --include='*.md' .` should return only generic prose, if anything
- [ ] `deploy-runbook.md` reads correctly as the sole documented deploy path here
- [ ] Recorded on `2AMLogic/2am#2`, which tracks the other half

## Test plan

No code changes, so no suite covers this. Verification is:

- [ ] `gh workflow list -R rjwalters/loom` no longer shows *Dashboard Deploy*
- [ ] `gh secret list -R rjwalters/loom` shows no `CLOUDFLARE_*` entries
- [ ] A subsequent push touching `dashboard/**` triggers no deploy here
- [ ] The instance repo's deploy still works afterwards (it must — it never depended on this workflow, only on this repo's contents at a pinned ref)

## Affected files

- `.github/workflows/dashboard-deploy.yml` — delete
- `dashboard/docs/deploy-runbook.md` — review, likely a small edit

rjwalters:
Closes #5567

## What this does

Deletes `.github/workflows/dashboard-deploy.yml`, which deployed **one specific operator's** live instance from this public mechanism repo while holding that operator's Cloudflare credentials in this repo's Actions secrets — contradicting the principle this repo states in `dashboard/docs/reference-deployment.md`.

The load-bearing reason is not tidiness. While both workflows existed, two repos deployed the same Worker on different triggers:

| | Trigger | Version deployed |
|---|---|---|
| this repo | every push to `main` touching `dashboard/**` | whatever `main` happens to be |
| the instance repo | a reviewed PR bumping a pinned ref | a deliberately chosen commit |

So any merge here **silently overwrote whatever ref the instance pinned**, making the pin advisory and the instance's deploy-of-record a claim rather than a fact. This closes that window.

## Sequencing gate — satisfied, and publicly checkable

```sh
curl -s https://dashboard.2amlogic.com/api/version
# {"commit":"0fe402ac523a6ba73d036b0aefe12fee7fa2325e"}
```

`0fe402ac` is a commit **in this repository** (`git cat-file -t 0fe402ac` → `commit`). The instance deploys this repo's `dashboard/` from that pinned ref and stamps the SHA it deployed, so the live site reporting it is end-to-end proof the successor path works.

## Diff

| File | Change |
|---|---|
| `.github/workflows/dashboard-deploy.yml` | deleted (-241) |
| `dashboard/docs/deploy-runbook.md` | dropped the companion-table row and the "bootstrap/recovery behind CI" framing; added a neutral paragraph on automating from your own infra repo, and the `BUILD_COMMIT` stamp in Step 6 |
| `.github/workflows/release.yml` | header comment cited the deleted file as a precedent — rephrased to the principle + `reference-deployment.md` |

### Why the runbook gained the `BUILD_COMMIT` note

The deleted workflow was the **only in-repo record** of how `/api/version` gets stamped. Wrangler `[vars]` are declarative per deploy, so a deploy that omits `--var BUILD_COMMIT:<sha>` does not preserve the previous value — it *removes* it, and `/api/version` reverts to `"unknown"` with a blank footer stamp under a passing HTTP-200 check. That near-miss happened on the instance's first deploy (caught in review). Documenting it here is what stops the next operator's automation rediscovering it. No instance, account, or domain is named anywhere in the doc changes.

## Verification

- `scripts/check-dangling-links.sh` — OK, 123 tracked markdown files, no dangling internal links (this is the check that would have caught the removed `../../.github/workflows/dashboard-deploy.yml` link).
- `scripts/check-docs-defaults-parity.sh` — OK.
- `scripts/check-claude-md-budget.sh` — OK (316/320).
- `actionlint .github/workflows/release.yml` — clean; all 11 remaining workflows parse.
- **Test coverage is unaffected**: `ci.yml`'s `dashboard-tests` job already runs the Miniflare backend + happy-dom UI suites on every push and on any PR touching `dashboard/**`. The deleted workflow's `test` job was a *deploy gate*, not the only place those suites ran.

## Reference sweep (acceptance criterion 3)

`grep -rn "dashboard-deploy" .` after the change returns only:

- `WORK_LOG.md` (2), `WORK_PLAN.md` (4) — historical/planning records of past issues, Guide-maintained and regenerated; left alone deliberately.
- `defaults/.claude/commands/loom/judge.md:1719` — a past-tense lesson citing #4974 ("a landed `dashboard-deploy.yml` needed a new secret discovered only post-merge"). Historical anecdote, still a valid lesson, not a live pointer. Left alone.

`grep -rn "2amlogic" --include='*.yml' --include='*.md' .` returns no file claiming this repo deploys anything: `CHANGELOG.md` and `WORK_LOG.md` history, `safehouse.md`/`token-pool.md` generic prose (a public fleet feed, a token-filename derivation example), and `observability.md` §6 describing the instance as operator-owned.

## Remaining work — operator step, cannot be done in a PR

Delete from this repo's Actions secrets:

- `CLOUDFLARE_API_TOKEN`
- `CLOUDFLARE_ACCOUNT_ID`
- `CLOUDFLARE_WRANGLER_CONFIG_2AMLOGIC`

**No rotation is required.** This repo's `CLOUDFLARE_API_TOKEN` is the *same* credential the instance repo holds — deleting this copy removes this repo's ability to use it and does not affect the instance's deploys.

Post-merge checks from the issue's test plan: `gh workflow list -R rjwalters/loom` no longer shows *Dashboard Deploy*; `gh secret list -R rjwalters/loom` shows no `CLOUDFLARE_*` after the operator step; a later push touching `dashboard/**` triggers no deploy here.

## Noted, not fixed (out of scope)

`observability.md` §6 (both `.loom/docs/` and `defaults/docs/` copies) says `reference-deployment.md` records the instance's "specific account/database IDs, Access application layout, credential file locations, and cutover history". That stopped being true when #5326 trimmed that doc to a generic 14-line pointer — the section now sends readers to content that no longer exists. Pre-existing, unrelated to the workflow deletion, and `reference-deployment.md` was declared out of scope on #5567, so it is left for a follow-up rather than folded in here.

## Related issues obsoleted

#5543 (*dashboard-deploy is pinned by a miniflare isolated-storage flake*) and its fix PR #5550 target the deleted workflow's deploy gate. Whoever merges should re-check whether they still apply — the flake itself may persist in `ci.yml`'s `dashboard-tests`, but the "live Worker has not redeployed" framing no longer does, since this repo no longer deploys it.


rjwalters:
## Judge review: APPROVED

Verified independently, not taken on report. The sequencing gate is real, the deletion orphans nothing live, and the doc changes stand on their own.

### 1. Sequencing gate — SATISFIED (the check that decides this PR)

```
$ curl -s -o /dev/null -w '%{http_code}' https://dashboard.2amlogic.com/api/version
200
$ curl -s https://dashboard.2amlogic.com/api/version
{"commit":"0fe402ac523a6ba73d036b0aefe12fee7fa2325e"}
$ git cat-file -t 0fe402ac523a6ba73d036b0aefe12fee7fa2325e
commit
$ git log -1 --format='%ci %s' 0fe402ac
2026-08-06 21:37:48 -0700 docs: Guide document maintenance update (#5564)
```

A 40-char SHA, present on `main` and `origin/main` in **this** repository — not `unknown`, not a branch name. The successor deploy path is proven end to end from the public internet. Deleting the workflow does not leave this repo without a working deploy path, because the working path was never here.

### 2. Nothing live depends on the deleted workflow — confirmed

`grep -rn "dashboard-deploy"` on the branch returns exactly what was reported: `WORK_PLAN.md`/`WORK_LOG.md` history, and the past-tense #4974 lesson at `defaults/.claude/commands/loom/judge.md:1719`. Correctly left alone — that lesson is about an anecdote, not a live pointer. `grep -rn "Dashboard Deploy\|deploy gate\|dashboard_deploy"` returns **zero** hits. No installer manifest reference (`scripts/install/manifest.sh`, `install-loom.sh`, `test-installer.sh` are all clean of `dashboard`).

I checked the thing the PR body did not: **required checks**. `main` has no branch-protection object; it is governed by ruleset 8809610, whose rules are `deletion`, `non_fast_forward`, `required_linear_history`, and `pull_request` (0 required approvals). There is **no `required_status_checks` rule at all**, so deleting the `Dashboard Tests (deploy gate)` job cannot orphan a required-check name and cannot wedge future merges.

### 3. Test coverage is not reduced — confirmed

`ci.yml`'s `dashboard-tests` runs `npm run check:all` in `dashboard/` — byte-identical to the deleted `test` job's command, same Node 24, same two `npm ci` steps. Its gate is `if: always() && (github.event_name == 'push' || needs.changes.outputs.dashboard == 'true')` with the filter `dashboard/**`. So: every push to `main`, plus any PR touching `dashboard/**`. Demonstrated live on this very PR — *Dashboard Tests (Workers backend + web UI)* ran and passed in 43s. The deleted job was purely a deploy gate. Nothing is lost.

### 4. Runbook stands alone — confirmed

Reads correctly as the sole documented deploy path. Names **no instance, account, or domain**: the only `CLOUDFLARE_ACCOUNT_ID` mentions are generic env-var prose, and there is no `2amlogic` string anywhere in the file. The new callout is better than the row it replaced — "This repository ships no deploy workflow, by design" states the principle rather than leaving a hole where a link used to be, and the clause about a deploy workflow here overwriting a deliberately-pinned ref is the actual load-bearing reason, now recorded where an operator will hit it.

### 5. The `BUILD_COMMIT` note — claim verified, and it belongs here

Checked against source rather than accepted:

- `dashboard/src/index.ts:150` — `BUILD_COMMIT?: string` (optional).
- `dashboard/src/index.ts:211` — `JSON.stringify({ commit: env.BUILD_COMMIT ?? "unknown" })`, served under a 200.
- `dashboard/src/index.ts:659` — footer omits the stamp entirely when unset.
- `dashboard/wrangler.toml:135` — `[vars]` holds `RETENTION_DAYS`, `MAX_RECORDS`, and the timezone. **No `BUILD_COMMIT`.** The file's own comment at line 134 states it: *"These are plain (non-secret) vars, so `wrangler deploy` publishes exactly what is written here."*

So the claim is exactly right: the var exists only as a per-deploy `--var`, `[vars]` is declarative, and a deploy omitting the flag publishes a var set without it — `/api/version` reverts to `"unknown"` while still answering 200. A health check that only asserts 200 will not notice.

I also verified the runbook's parenthetical, which is the part that would have been easy to get wrong: `dashboard/package.json` has `deploy = npm run build:web && wrangler deploy`, so `npm run deploy -- --var "BUILD_COMMIT:$(git rev-parse HEAD)"` expands with the args on the **last** command. The trailing args do reach `wrangler deploy`, and the UI still gets built. Correct as written.

**Not scope creep.** The deleted workflow was the only in-repo record of how that stamp gets set. Deleting it without this note would have destroyed the knowledge and left the next operator's automation to rediscover a silent-failure mode by shipping it. Removing a mechanism obliges you to keep the fact the mechanism encoded, and this does exactly that in the doc that inherits the job.

### 6. Lint and secrets

`actionlint` clean across all 11 remaining workflows. `check-dangling-links.sh` OK (123 files — this is the check that would have caught the removed workflow link, and it does resolve now). `check-docs-defaults-parity.sh` OK. `check-claude-md-budget.sh` OK (316/320). No secret **material** in the diff — only secret *names* disappearing along with their consumer. The `release.yml` rephrase is a genuine fix: the old comment cited a file that would no longer exist, and pointing at `reference-deployment.md` names the principle instead of a precedent that was about to vanish.

CI: all green or correctly skipped. `mergeStateStatus: CLEAN`.

---

## One finding the merger must act on — merge order vs. PR #5550

Flagged in the PR body as "re-check whether they still apply." That undersells it, and I want the specific hazard on the record.

**#5550 is not hypothetical — it is already `loom:pr`, `MERGEABLE`/`CLEAN`, and queued for Champion auto-merge.** Its file list:

```
.github/workflows/dashboard-deploy.yml            <- modified; this PR deletes it
.github/workflows/dashboard-deploy-staleness.yml  <- ADDED (new file)
dashboard/package.json, package-lock.json, patches/...
```

Two distinct consequences, and the second is the one that bites:

1. `dashboard-deploy.yml` is a **modify/delete** pair. Git surfaces that as a conflict, so #5550 will flip to `DIRTY` if this lands first. That is a safe failure — but whoever resolves it must resolve it by **taking the deletion**. An agent rebasing without this context could resurrect the very workflow #5567 exists to remove.

2. `dashboard-deploy-staleness.yml` is a **pure add**, so it will **not** conflict and will merge silently in either order. It hardcodes:

   ```yaml
   response="$(curl -sS -m 15 "https://dashboard.2amlogic.com/api/version" || true)"
   ```

   and files GitHub issues when that instance is behind `main`. That reintroduces one operator's domain into this public mechanism repo and monitors a deploy this repo no longer performs — the exact thing this PR removes, arriving through the side door. Blast radius is limited (its `cron` is commented out; `workflow_dispatch` only), but it would land as a tracked file naming the instance, in direct violation of acceptance criterion 3 on #5567 and of the `reference-deployment.md` principle.

**Recommended:** merge #5570 first, then take #5550 off `loom:pr` and back to triage rather than letting it auto-merge. Its premise — "the live Worker has not redeployed" — is now the instance repo's concern, and its staleness workflow belongs there, not here. The miniflare isolated-storage flake it patches is real and *does* still affect `ci.yml`'s `dashboard-tests`, so that half should be re-cut as its own issue against `ci.yml`, without the workflow and without the domain.

## Deferrals — both judged correct

- **`observability.md` §6.** Right to defer: pre-existing since #5326 trimmed `reference-deployment.md`, and `reference-deployment.md` was explicitly out of scope on #5567. Folding it in would have blurred a clean deletion. One addition for the follow-up issue — the stale claim appears **twice** per copy, not once. Besides §6's "specific account/database IDs, Access application layout, credential file locations, and cutover history", the *Map of every detail doc* table further down still says `dashboard/docs/reference-deployment.md | The 2AM instance specifically — concrete IDs, current state`. Same staleness, same fix, both `.loom/docs/` and `defaults/docs/`.
- **The three Cloudflare secrets.** Correctly excluded — not expressible in a diff. The reasoning that no rotation is required is sound: this repo holds a *copy* of the credential the instance repo also holds, so deleting this copy revokes this repo's ability to use it without touching the instance's deploys. Still worth doing promptly; until it is, this repo holds live credentials it has no remaining use for, which is a standing risk with no upside.

---

**Approved.** The single decisive check passes: the successor deploy is live, serving a commit that exists here. This closes a real correctness hole — while both workflows existed, any merge here silently overwrote the instance's pinned ref, making that pin advisory. Removing the automated deploy path is the right call precisely because it was never this repo's to run.

PR #5580, opened 2026-08-07
fix(guide): close TOCTOU race in doc-maintenance Step 1 open-PR check

### LINKED ISSUE #5573: Guide's doc-maintenance Step 1 open-PR check is a TOCTOU race — two concurrent instances opened duplicate PRs #5571/#5572
## Problem

Two duplicate `docs: Guide document maintenance update` PRs were opened 49
seconds apart:

- **#5571** — branch `docs/guide-update-20260807-054234`, created `2026-08-07T05:45:43Z`
- **#5572** — branch `docs/guide-update-20260807-053847`, created `2026-08-07T05:46:32Z`

Both carry an identical diff shape (10 additions / 6 deletions, same two files
touched: `WORK_LOG.md` +2/-0, `WORK_PLAN.md` +8/-6). Both are still open at
filing time.

## Root Cause

The document-maintenance phase's Step 1 guard (`.loom/roles/guide.md` §
"Step 1: Check for Existing Docs PR") is a check-then-create sequence with no
atomicity:

```bash
OPEN_DOCS_PR=$("$GH_READ" pr list --state open --search "head:docs/guide-update" ...)
if [ -n "$OPEN_DOCS_PR" ]; then
  # skip
fi
# ... otherwise: docs-worktree.sh, edit, commit, push, gh pr create
```

When two Guide instances run within the same short window (e.g. a scheduled
role-runner tick overlapping a manually-invoked `/loom:guide` session, or two
role-runner ticks overlapping each other), both can query "is there an open
docs PR?" before either has pushed its branch and created a PR. Both see
none, both proceed, both call `./.loom/scripts/docs-worktree.sh` (which resets
the *same* persistent worktree slot — the second call clobbers the first's
in-progress worktree state), and both end up on independently-branched,
independently-pushed docs PRs.

This is a second, narrower instance of the class of bug fixed for the
self-referential loop in #5454/#5465 — that fix stopped the phase from
manufacturing infinite new-content signals from its own merged PRs, but it
did not add any mutual exclusion between concurrent Guide instances hitting
Step 1 at the same time.

## Suggested Fix

Make the "is there an open docs PR" check-and-claim atomic, e.g.:
- Use `gh pr create` itself as the exclusion mechanism (attempt create,
  handle a "reference already exists" / duplicate-branch push failure as the
  signal to abort), or
- Take a lock (flock-style, e.g. a `.loom/locks/docs-guide` sentinel checked
  and created atomically) around the Step 1 check through the `gh pr create`
  call, released after push succeeds or on early-exit.

## Cleanup

#5571 and #5572 both need triage — one is redundant once the other merges
(review should confirm whichever merges second doesn't just become a
zero-diff no-op against the freshly-updated `WORK_LOG.md`/`WORK_PLAN.md`, and
close it as a duplicate if so).

## Affected Files

- `.loom/roles/guide.md` (Step 1 of Document Maintenance)
- `defaults/roles/guide.md` (source of the same section)
- `.loom/scripts/docs-worktree.sh`
---

### Acceptance Criteria

- [ ] The Step 1 open-docs-PR check and the worktree/branch/PR creation path (`.loom/roles/guide.md` + `defaults/roles/guide.md`, and/or `.loom/scripts/docs-worktree.sh`) close the TOCTOU race — e.g. via a lock file taken atomically around the check-through-`gh pr create` window, a second re-check for an open PR immediately before `gh pr create`, or reusing/updating a single stable branch name per tick instead of a fresh `docs/guide-update-<timestamp>` branch (so a second concurrent run is naturally deduped onto the same branch/PR).
- [ ] A simulated concurrent-tick scenario (two ticks starting within the current race window) no longer produces two open docs PRs.
- [ ] No regression to the existing "only one docs PR open at a time" behavior for the non-racing case (a second tick that starts well after the first has already pushed/created its PR must still skip cleanly).
- [ ] Fix is applied identically to both `.loom/roles/guide.md` and `defaults/roles/guide.md` (verified identical as of `4b8a3a76`, 2026-08-06 — re-diff before relying on this if either has since diverged).

### Test Plan

- [ ] Manual verification: simulate two near-simultaneous invocations of the Step 1 check-and-create sequence (e.g. two backgrounded shell runs of the guarded block) against a scratch branch prefix and confirm only one PR/branch survives.
- [ ] Automated tests: N/A — this is a shell-script race condition in a role prompt's embedded bash, not a unit-testable code path; rely on the manual simulation above plus observing real role-runner ticks post-fix for a period with no duplicate `docs/guide-update-*` PRs.
- [ ] Edge cases: (a) lock/branch-reuse mechanism must not leave a stale lock behind if a tick crashes mid-phase (verify a crashed tick doesn't permanently block future ticks); (b) verify the fix does not reintroduce the #5454/#5465 self-referential-PR bug (the phase's own merged docs PRs must still be excluded from "new content" detection).

### Current status of the observed duplicate pair (verified 2026-08-06 via `gh pr view`)

- **#5571** — still OPEN as of this curation pass.
- **#5572** — CLOSED (not merged) — already resolved as a duplicate during triage, confirming the "Cleanup" section's prediction above.

### Related

- #5574 — a second issue independently filed ~5 minutes after this one, describing the identical root cause against the same #5571/#5572 evidence. Closed as duplicate of this issue during curation; its acceptance-criteria checklist above and the #5413 cross-reference below were absorbed from it.
- #5413 — prior Guide document-maintenance outage (different root cause: a guard denial in the main checkout, not a concurrency race).
- #5454 / #5465 — prior fix for a different but related bug class in this same phase (self-referential infinite-PR loop from the phase's own merged PRs being miscounted as new content). The fix for this issue must not reintroduce that one.

rjwalters:
## Summary

Fixes #5573 — the Guide role's Document Maintenance Step 1 open-docs-PR check
was a plain check-then-act with no atomicity. Two Guide ticks starting within
the same short window (a role-runner tick overlapping a manual `/loom:guide`
session, or two role-runner ticks overlapping each other) could both see "no
open PR" before either had pushed a branch, both call `docs-worktree.sh`
(clobbering the shared, single worktree slot), and both end up on
independently-branched, independently-pushed docs PRs. This actually
happened: #5571 and #5572, opened 49 seconds apart with an identical diff
shape.

### Fix

Added `.loom/scripts/docs-guide-lock.sh` (a non-blocking, mkdir-based lock —
the same portable atomic primitive `worktree.sh`'s worktree-add lock and
`lib/build-slot.sh` already use in this repo; `flock` isn't available on
stock macOS) and wired it into `defaults/.claude/commands/loom/guide.md`
(which `.loom/roles/guide.md` and `defaults/roles/guide.md` both symlink to
— a single edit covers all three paths named in the issue):

- **Step 1** now `acquire`s the lock *before* the open-docs-PR check. A busy
  lock (another tick mid-flight) makes this tick skip outright — non-blocking,
  since waiting would just shift the identical race to a later timestamp
  instead of closing it.
- **Step 5**'s `create_docs_pr()` releases the lock on both of its exit paths
  (the "no changes to commit" early return, and after `gh pr create`
  succeeds). Step 1's early-return-on-open-PR path releases it too.
- Staleness is **age-based** (default 1800s / 30 min), not PID-liveness like
  `worktree.sh`'s lock: the "holder" here is a sequence of separate Bash
  tool-call subprocesses within one Guide agent turn (interleaved with Edit
  tool calls), not one continuously running process — a PID recorded at
  `acquire` time has already exited long before Step 5 runs, so PID-liveness
  would misclassify every legitimate in-flight lock as stale within seconds.
  A crashed tick's lock is reaped automatically by the next `acquire` once it
  ages past the threshold, so it can never permanently block future ticks.
- The pre-existing open-docs-PR check (Step 1) is unchanged in intent and
  still needed — it is what prevents a *later, non-racing* tick from piling a
  second PR onto one left open by an earlier tick, which is a distinct
  concern from the lock (which only stops *concurrent* ticks from racing each
  other).
- The #5454 self-referential-PR exclusion (`GUIDE_DOCS_PR_EXCLUDE`) was left
  untouched and is verified intact by the new test suite.

### Verification

Added `defaults/scripts/tests/test-docs-guide-lock.sh` (wired into
`ci-wired.txt`), hermetic, no forge/network calls:

- acquire/release/status round-trip
- release is idempotent
- sequential serialization (second acquire while held returns busy)
- **the actual #5571/#5572 failure mode, executed rather than simulated**:
  two truly concurrent, backgrounded `acquire` attempts race on the same
  lock — verified exactly one wins, every run (ran the suite 5x locally with
  no flakiness)
- stale-lock recovery (a lock older than `LOOM_DOCS_GUIDE_LOCK_STALE_SECS` is
  reaped by the next acquire; a fresh lock is not)
- argument validation
- guide.md wiring: the lock is acquired before the open-PR check (not after,
  which would just reintroduce the same TOCTOU gap), and released on every
  known exit path (>=3 release call sites)
- the #5454 exclusion is still present and unmodified

All pre-existing related suites still pass unchanged:
`test-docs-worktree.sh`, `test-guide-work-log-self-loop.sh`,
`test-guide-work-log-issue-out-of-order.sh`,
`test-guide-work-log-out-of-order-merge.sh`, `test-sweep-auto-stack.sh`,
`test-warn-out-of-set-deps.sh`.

### Files changed
- `defaults/.claude/commands/loom/guide.md` (canonical target of the
  `.loom/roles/guide.md` / `defaults/roles/guide.md` symlinks)
- `defaults/scripts/docs-guide-lock.sh` (new)
- `defaults/scripts/tests/test-docs-guide-lock.sh` (new)
- `defaults/scripts/tests/ci-wired.txt`

Closes #5573

### DIFF
--- a/defaults/.claude/commands/loom/guide.md
+++ b/defaults/.claude/commands/loom/guide.md
@@ -1121,26 +1121,63 @@ These are idempotent across a fresh checkout: whatever is already in the
 committed WORK_LOG.md defines the watermark, so the same PR is never appended
 twice even though no gitignored state persists between cron ticks.
 
-### Step 1: Check for Existing Docs PR
+### Step 1: Acquire the Docs-Guide Lock, Then Check for an Existing Docs PR
 
-Before creating any changes, check if a previous docs PR is still open:
+**#5573 BUG, DO NOT REINTRODUCE:** this used to be a plain check-then-act — query
+for an open docs PR, and if none, proceed. Two Guide ticks starting within the
+same short window (a role-runner tick overlapping a manual `/loom:guide`
+session, or two role-runner ticks overlapping each other) could both see "no
+open PR" before either had pushed a branch, both call `docs-worktree.sh`
+(clobbering the shared worktree slot — see "Where This Phase Writes" above),
+and both end up creating their own docs PR (observed: #5571 and #5572, opened
+49 seconds apart with an identical diff shape). The check and the create were
+not atomic.
+
+The fix is a non-blocking mkdir-based lock (`docs-guide-lock.sh`) that must be
+held across the **entire** window from this check through Step 5's
+`gh pr create` — not just around the check itself:
 
 ```bash
+# Non-blocking: a busy lock means another tick is already mid-flight through
+# this same phase, so this tick skips outright rather than waiting (waiting
+# would just shift the identical race to a later timestamp, not close it).
+if ! ./.loom/scripts/docs-guide-lock.sh acquire; then
+  echo "Another Guide tick already holds the docs-guide lock. Skipping document maintenance this tick."
+  return
+fi
+
+# From this point on, EVERY exit path from the Document Maintenance phase —
+# every early `return` below, the end of Step 5 (create_docs_pr, both its
+# "no changes" and its success paths), and any error path — MUST call
+# `./.loom/scripts/docs-guide-lock.sh release` before ending the turn. There
+# is no shell `trap` that can cover this automatically: Steps 1-5 run as
+# separate Bash tool invocations (interleaved with Edit tool calls to write
+# the docs) within one Guide turn, not one continuous process. If a tick
+# crashes without releasing, the NEXT tick's `acquire` reaps the lock
+# automatically once it is older than LOOM_DOCS_GUIDE_LOCK_STALE_SECS
+# (default 1800s / 30 min) — see docs-guide-lock.sh's header comment for why
+# staleness is age-based rather than PID-liveness here.
+
 # Match on the branch-name PREFIX, not an exact head. Docs branches are named
 # `docs/guide-update-<timestamp>` (see Step 5), so `--head "docs/guide-update"`
 # (an exact-match filter) never matched and the "only one docs PR open" guard
 # never fired — PRs accumulated. `--search "head:docs/guide-update"` matches the
-# prefix.
+# prefix. This check runs INSIDE the lock now, but it is not redundant with
+# it: the lock closes the race between two CONCURRENT ticks, while this check
+# is what makes a docs PR left open from a PRIOR (non-racing) tick — whose
+# lock was already released once its `gh pr create` succeeded — still cause
+# the next tick to skip.
 OPEN_DOCS_PR=$("$GH_READ" pr list --state open --search "head:docs/guide-update" --json number --jq '.[0].number // empty')
 
 if [ -n "$OPEN_DOCS_PR" ]; then
   echo "Docs PR #$OPEN_DOCS_PR is still open. Skipping document maintenance."
   # Optionally: check if it's stale and comment
+  ./.loom/scripts/docs-guide-lock.sh release
   return
 fi
 ```
 
-If a docs PR is already open, **skip the entire document maintenance phase** to prevent PR accumulation.
+If a docs PR is already open, **skip the entire document maintenance phase** to prevent PR accumulation (and release the lock — there is nothing left for this tick to do).
 
 ### Step 2: Update WORK_LOG.md
 
@@ -1472,6 +1509,9 @@ create_docs_pr() {
   # Check if there are actual changes to commit
   if git -C "$DOCS_WT" diff --cached --quiet; then
     echo "No document changes to commit."
+    # Release the docs-guide lock (see Step 1) — nothing left for this tick
+    # to do, and a held lock would needlessly block the next one.
+    ./.loom/scripts/docs-guide-lock.sh release
     return
   fi
 
@@ -1506,6 +1546,12 @@ See issue #1784 for the feature specification.
 PRBODY
 )")
 
+  # Release the docs-guide lock (see Step 1) now that the PR exists. Step 1's
+  # open-docs-PR check — not this lock — is what prevents the NEXT tick from
+  # creating another one while this PR is still open; this lock's only job
+  # was to keep concurrent ticks from racing each other to this point.
+  ./.loom/scripts/docs-guide-lock.sh release
+
   # No side-car state to update — the committed WORK_LOG.md / WORK_PLAN.md carried
   # in this PR ARE the durable state the next tick reads back.
   #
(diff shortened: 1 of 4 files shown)


rjwalters:
LGTM — approving.

Reviewed against issue #5573's acceptance criteria and the actual diff (not just the self-report):

**Lock placement (the core correctness property)**: `docs-guide-lock.sh acquire` runs at the very top of Step 1, *before* the `OPEN_DOCS_PR` search — confirmed by reading the guide.md diff directly. This closes the TOCTOU gap structurally: a second racing tick that arrives while the lock is held fails to acquire and skips outright, so it can never even reach the open-PR check, let alone `docs-worktree.sh` or `gh pr create`. This is the fix the issue actually asked for (not just narrowing the window).

**Exit-path coverage**: traced all paths out of the Document Maintenance phase — Step 1's early-return-on-open-PR releases the lock; `create_docs_pr()` releases it on both its "no changes to commit" early return and after a successful `gh pr create`. The `return 1`/`return 0` calls inside `update_work_log()` (Step 2) are ordinary bash-function status returns consumed by the caller, not phase exits, so they don't need their own release call — confirmed Step 5 always runs regardless of Steps 2-4's return values.

**Staleness design**: age-based (default 1800s) is the right call here, and the header comment explains why clearly — the "holder" spans multiple separate Bash tool-call subprocesses within one Guide turn, so PID-liveness would misclassify every legitimate in-flight lock as stale within seconds. 30 min is generous headroom over the phase's normal runtime without risking a permanently-stuck lock.

**Symlink topology**: confirmed `.loom/roles/guide.md` and `defaults/roles/guide.md` both resolve (via `readlink -f`) to the single file the PR touches, `defaults/.claude/commands/loom/guide.md` — no second edit was needed, and none was made.

**#5454 regression check**: `GUIDE_DOCS_PR_EXCLUDE` is untouched in the diff — verified both by direct diff inspection and by the new test suite's Test 8.

**Test suite**: ran `test-docs-guide-lock.sh` 3x locally — 23/23 pass every time, including Test 4 (two truly concurrent backgrounded `acquire` calls racing on the same lock — exactly one wins, and which one varies run to run, confirming it's a real race, not a fixed-order simulation), Test 5 (stale-lock reaping + fresh-lock non-reaping), and Test 7 (asserts against the live `guide.md` content — lock acquired before the open-PR check, ≥3 release call sites — so this can't silently drift from the shipped prompt). Correctly wired into `ci-wired.txt` (alphabetical, matches the real filename).

**Regression sweep**: ran the six pre-existing suites the PR description claims still pass (`test-docs-worktree.sh`, `test-guide-work-log-self-loop.sh`, `test-guide-work-log-issue-out-of-order.sh`, `test-guide-work-log-out-of-order-merge.sh`, `test-sweep-auto-stack.sh`, `test-warn-out-of-set-deps.sh`) — all pass. `shellcheck --severity=warning` (matching CI's actual invocation) is clean on both new files; the one SC2317 info-level hit on the trap-invoked `cleanup()` function is below CI's severity threshold.

**CI / merge state**: all applicable GitHub Actions checks pass (Rust/Dashboard/MCP/loom-worker jobs correctly skipped via path-filtering since this PR only touches shell/prompt/test files), `Shell Test Suites (hermetic)` passed in 6m52s, and `mergeStateStatus` is `CLEAN`.

No issues found. Approving.

Issue #5589, opened 2026-08-07
loom-daemon forge auto-merge lacks the head-SHA precondition added to the shell merge path (#5579 follow-up)


rjwalters:
## Context

Follow-up from #5579 ("Champion can squash-merge a PR while a session is still pushing to its branch, stranding commits invisibly"), which added a head-SHA optimistic-concurrency precondition to the **shell** merge path:

- `forge_merge_pr()` / `forge_auto_merge()` (`defaults/scripts/lib/forge-helpers.sh`) now accept an optional 3rd `EXPECTED_HEAD_SHA` argument — GitHub REST's `sha` field, the GraphQL `expectedHeadOid` input, and Gitea's `head_commit_id` field respectively.
- `merge-pr.sh` reads a fresh (uncached) head SHA immediately before merging and threads it through both calls, refusing (exit 3, not a hard failure) and re-queuing when the forge reports a mismatch instead of silently squash-merging whatever the current head happens to be.

## Problem

The Rust-native `loom-daemon forge auto-merge <pr> [--method M]` subcommand (`loom-daemon/src/forge_cmd.rs`, `github_auto_merge()` around line 425, wired via `handle_auto_merge()`) has **no equivalent flag**. It calls the same `enablePullRequestAutoMerge` GraphQL mutation the shell `forge_auto_merge()` now guards, but without `expectedHeadOid`.

This matters because `merge-pr.sh` **prefers** this native path whenever `loom-daemon` is on `PATH` (the common case for a daemon-operated fleet) — see the "Prefer the native `loom-daemon forge auto-merge`" comment block around `defaults/scripts/merge-pr.sh:1270`. The native call only exits 3 to *decline* on Gitea; a GitHub `--auto` merge routed through it can still silently squash whatever the current head is, exactly reproducing #5579's root cause on the path most fleets actually exercise. The shell-side fix in #5579 only fully protects:

- The synchronous (non-`--auto`) merge path (`forge_merge_pr`, always shell).
- The `--auto` path when `loom-daemon` is absent from `PATH`, or when it declines (Gitea).

## Suggested direction

[...]

PR #5594, opened 2026-08-07
fix(champion): thread expected head SHA into merge API calls (#5579)

### LINKED ISSUE #5579: Champion can squash-merge a PR while a session is still pushing to its branch, stranding commits invisibly
## Context

During a long interactive session on a repo with `autonomous.roleRunner`
running Champion on idle, a Champion squash-merged an open PR **while the
session was still pushing commits to that PR's branch**. It happened twice in
about ninety minutes, on two different PRs.

The shape each time:

1. PR opened with commit A. Labeled `loom:review-requested`.
2. Session continues working, pushes commits B and C to the same branch.
3. Champion (or a role-runner Judge → Champion) merges the PR — at whatever
   head it saw.
4. B and C are on a branch that is no longer targeted by anything.

## Why this is worse than an ordinary race

**Squash-merge makes it invisible to the obvious check.** The merge commit is a
new SHA, so `git merge-base --is-ancestor <my-commit> origin/main` returns false
for *every* commit in the PR — including the ones that did land. There is no
cheap way to tell "squashed, content present" from "stranded, content absent"
without diffing file content on `main`. A session that trusts the merge
notification silently loses work.

**Recovery has its own hazard.** The natural fix — re-apply the stranded work
from the old branch — reverts anything that landed on `main` in the meantime.
In this session the stranded branch predated another merged PR, so carrying the
file across wholesale would have silently reverted that PR's changes. Caught
by diffing first, but the safe path is not the obvious one.

## Not asking for the merge to be slower

Champion merging approved work promptly is the point of champion-on-idle, and
in both cases it merged a genuinely approved PR. The question is narrower:
should it merge a PR whose head moved very recently, or whose branch has commits
newer than the approval it is acting on?

## Possible directions

- Skip a PR whose head SHA changed after the approving review, and re-queue it
  (a Judge re-review would be correct there anyway — the approval was for
  different code).
- Or: treat this as inherent and **document it** — an open PR is not a safe
  place to accumulate commits while a Champion is running; push each unit of
  work as its own PR. That is a legitimate answer, it just is not written down
  anywhere today.

Related: 2AMLogic/2am#4884 already documents that a role-runner Champion can
complete a sweep's planned candidates out from under it, and added wave-boundary
re-verification for that. This is the same class of interaction, one level down
— the *PR* rather than the *candidate list*.

- [ ] A decision recorded: does Champion skip a PR whose head advanced past the
      approving review, or is accumulating commits on an open PR documented as
      unsafe under champion-on-idle?
- [ ] If the former: Champion re-queues rather than merges, and the skip is
      logged loudly enough to be visible in its output
- [ ] If the latter: documented where a session author would actually see it —
      the sweep skill and/or the champion role docs, not only a changelog
- [ ] Either way, note the squash-merge detection trap: ancestry checks cannot
      distinguish squashed-and-landed from stranded, so verification must
      compare file content

---

### Root cause (verified against `main`, 2026-08-07)

Traced the merge path used by both the synchronous and `--auto` flows and confirmed neither passes GitHub's optimistic-concurrency guard for exactly this race:

- `forge_merge_pr()` (`defaults/scripts/lib/forge-helpers.sh:311-324`) issues `PUT /repos/{nwo}/pulls/{n}/merge` with only `merge_method=squash` — no `sha` field. GitHub's merge endpoint accepts an optional `sha` field ("SHA that pull request head must match to allow merge"); when omitted, GitHub merges whatever the current head is at the moment the request lands, which is exactly the silent-strand mechanism the issue describes. (Verify the exact failure-mode response code/message against current GitHub API docs before implementing — documented historically as 409 "Head branch was modified. Review and try the merge again.", but re-check since this is a third-party API detail, not something in this repo.)
- `forge_auto_merge()` (`defaults/scripts/lib/forge-helpers.sh:497-518`), the GraphQL fallback for the `--auto` flag Champion always passes (`champion-pr-merge.md:860`), calls `enablePullRequestAutoMerge` with `pullRequestId` and `mergeMethod` only — GitHub's mutation also accepts an `expectedHeadOid` field for the same purpose, also unused here.
- `merge-pr.sh` already fetches the PR's head SHA into `$PR_HEAD_SHA` near the top of the script (`.loom/scripts/merge-pr.sh:384`, mirrored at `defaults/scripts/merge-pr.sh`) — but only for later worktree/branch-cleanup safety checks (`_maybe_delete_local_branch`), never threaded into either merge call. The value the fix needs already exists in scope; it's just not passed to the API calls that need it as a precondition.
- The synchronous retry loop (`.loom/scripts/merge-pr.sh:1651-1730`) already distinguishes several forge error strings ("Base branch was modified" → update-branch-and-retry; "Merge already in progress" → wait-and-recheck). A head-SHA-mismatch response is a **third**, currently-unhandled case, and it must **not** reuse the "Base branch was modified" retry path — blindly retrying that would either fail again (if the session keeps pushing) or squash whatever the new head happens to be, defeating the entire point of the guard.

This makes the first "possible direction" in the original issue body concrete: adding the precondition doesn't require Champion to re-architect its queueing — it requires threading one already-computed value into two existing forge calls, plus a third branch in the existing retry loop that treats a mismatch as "re-queue, don't retry-and-merge-anyway."

### Suspected cause (unverified)

Whether a Gitea target needs the same treatment is unclear — `forge_merge_pr()`'s and `forge_auto_merge()`'s Gitea branches (`gitea_api POST .../pulls/{n}/merge`) were not checked against Gitea's API for an equivalent head-SHA precondition field. Needs verification by the Builder; if Gitea has no equivalent, that's a documented gap, not a blocker for the GitHub-side fix.

### Recommended direction

Given the root cause above, prefer the **first** possible direction from the original body (skip/re-queue on a moved head) over the documentation-only direction — the mechanism to detect the race atomically already exists in GitHub's API and mostly-exists in this repo's own code (the SHA is already being read, just not passed through). Concretely:

1. Add an optional `sha`/`expectedHeadOid` parameter to `forge_merge_pr()` and `forge_auto_merge()`, sourced from the freshest possible head-SHA read (not a cached one — see `champion-pr-merge.md`'s "Cached forge reads" convention, which already treats merge-gating reads as always-live for this exact reason).
2. In `merge-pr.sh`'s retry loop, add a distinct branch for the head-mismatch response: do not retry-and-merge, surface it as a distinct exit condition the caller (Champion) can recognize.
3. In `champion-pr-merge.md` Step 3, treat that distinct exit condition as "re-queue, do not comment a failure" — the PR is still Judge-approved, its diff just moved; a fresh pass should re-evaluate it (which may also want a fresh Judge look, per the issue's own suggestion), not silently drop it from the queue.
4. Leave the "document as unsafe" direction as a fallback only if step 1 turns out to be infeasible against the live GitHub/Gitea API (e.g., unexpected interaction with branch protection or required-status-check auto-merge queuing) — verify this by attempting the change against a real PR before falling back to documentation-only.

### Affected Files

- `defaults/scripts/lib/forge-helpers.sh` — `forge_merge_pr()` (~line 311) and `forge_auto_merge()` (~line 497): add head-SHA precondition parameter to both the GitHub and (if applicable) Gitea branches.
- `defaults/scripts/merge-pr.sh` (`.loom/scripts/merge-pr.sh` is a symlink to this file, not a separate copy — edit only the `defaults/` path) — thread `$PR_HEAD_SHA` (already computed ~line 384) into the `forge_merge_pr`/auto-merge calls; add a distinct retry-loop branch (~lines 1651-1730, 1225-1265) for a head-mismatch response that does NOT retry-and-merge.
- `defaults/.claude/commands/loom/champion-pr-merge.md` — Step 3 ("Merge the PR", ~line 839): read the head SHA immediately before merging (freshest possible read, per the file's own "Cached forge reads" convention) and pass it through; handle the new "head moved, re-queue" outcome distinctly from the existing failure-comment path in "Error Handling".
- Any Rust-side `loom-daemon forge auto-merge` implementation backing `.loom/scripts/merge-pr.sh:1241` — out of scope to locate in this curation pass; the Builder should check whether it has the same gap, since it's the primary path Champion exercises when `loom-daemon` is present.

### Test Plan

- [ ] Manual verification: open a PR, get it Judge-approved (`loom:pr`), then push a new commit to its branch immediately before triggering Champion's merge pass — confirm the merge is now rejected/deferred rather than silently squashing only the pre-push commits, and that the PR is re-queued (not stuck, not silently dropped).
- [ ] Automated tests: add/extend shell tests (or the existing `merge-pr.sh` test harness, if one exists — check `.loom/tests/` or similar) covering `forge_merge_pr`/`forge_auto_merge` with a mismatched `sha`/`expectedHeadOid`, asserting the distinct non-retry code path fires.
- [ ] Edge cases: (a) head moves between Champion's safety-criteria read and the merge call but the new commits are benign/identical content (should still be treated conservatively — re-queue, don't try to distinguish "safe" moves); (b) Gitea target — confirm whether the precondition is available, and if not, document the gap explicitly rather than leaving it silently unguarded; (c) the `--auto` GitHub auto-merge-queue path, where the actual merge may happen asynchronously well after `enablePullRequestAutoMerge` returns — confirm `expectedHeadOid` is still honored at the time the queued merge actually executes, not just at enable-time.

### Note on the issue's own acceptance criteria

The original "Suggested acceptance criteria" checklist frames this as a decision to be recorded first. Given the verified root cause above, the decision has a clear, low-cost technical answer (thread the existing SHA through, handle the new failure branch) — the Builder should implement it directly rather than opening a separate decision-recording step, but should still fill in the issue's fourth checkbox item (documenting the squash-merge detection trap) since that's independently true and useful regardless of which direction is taken.

rjwalters:
## Summary

Champion could squash-merge a PR while a session was still pushing new commits
to its branch, silently stranding those commits — squash-merge makes this
invisible to the obvious `git merge-base --is-ancestor` check, since the new
squash commit is never a descendant of the stranded commits either way.

## Root cause

Neither `forge_merge_pr()` nor `forge_auto_merge()`
(`defaults/scripts/lib/forge-helpers.sh`) passed a head-SHA optimistic-concurrency
precondition to the forge's merge API, even though `merge-pr.sh` already
computed `$PR_HEAD_SHA` for an unrelated purpose (local branch cleanup safety).

## Changes

- **`defaults/scripts/lib/forge-helpers.sh`**: `forge_merge_pr()` and
  `forge_auto_merge()` accept an optional 3rd `EXPECTED_HEAD_SHA` argument,
  threaded into GitHub's `sha` (REST) / `expectedHeadOid` (GraphQL) fields and
  Gitea's `head_commit_id` field. Backward compatible — omitted entirely when
  not supplied.
- **`defaults/scripts/merge-pr.sh`**:
  - Reads the freshest possible head SHA (`MERGE_PRECONDITION_SHA`, via an
    uncached `forge_get_pr_nocache` call) immediately before either merge path
    runs, falling back to the already-known `$PR_HEAD_SHA` on lookup failure
    (fail-safe: a stale value only ever produces a spurious re-queue, never a
    stale-but-accepted merge — the forge itself does the real comparison).
  - New `_is_head_mismatch_response()` classifier + `error_head_moved()` helper:
    a head-SHA-mismatch response is now a **distinct exit code (3, not 1)**,
    checked *before* the existing "Base branch was modified" retry branch in
    both the synchronous and `--auto` retry loops, so a future edit can't
    accidentally let the base-modified retry eat a head-mismatch case.
    Exit 3 must NOT retry-and-merge (that would defeat the guard) — it
    surfaces a distinct condition for the caller to recognize as "re-queue,
    don't fail loudly."
  - Documented in the script's "Exit codes" header comment.
- **`defaults/.claude/commands/loom/champion-pr-merge.md`**: Step 3 captures
  `merge-pr.sh`'s exit code and branches on 3 — re-queues the PR (leaves
  `loom:pr` in place, does not post a failure comment, does not count it as an
  error) instead of following the existing failure path. New "Exception: exit
  code 3" subsection in Error Handling documents this, plus the **squash-merge
  detection trap**: ancestry checks can't distinguish "squashed-and-landed"
  from "stranded" after the fact — verification requires diffing actual file
  content on `main`, not `git merge-base --is-ancestor`.
- **`defaults/scripts/tests/test-merge-pr-head-mismatch.sh`** (new, wired into
  `ci-wired.txt`): 23 assertions across 4 layers — SHA threading for both
  forges (GitHub REST/GraphQL, Gitea), the classifier/exit-code helper against
  positive/negative string fixtures (including the pre-existing "Base branch
  was modified" string, to confirm the two stay mutually exclusive), source
  wiring + branch ordering in both retry loops, and `champion-pr-merge.md`'s
  exit-3 handling + squash-merge-trap documentation.

## Verified against live API specs (documented in code comments)

- GitHub REST `sha` field / 409 "Head branch was modified. Review and try the
  merge again." — verified against GitHub's public OpenAPI spec.
- Gitea `head_commit_id` field / `ErrSHADoesNotMatch` → 409 "head out of date"
  — verified against Gitea/Forgejo's published swagger + upstream source.
- GitHub GraphQL `expectedHeadOid` input field — confirmed present in GitHub's
  public GraphQL schema; the exact mismatch error text could **not** be
  verified against a live incident or published docs (GraphQL validation
  error text isn't part of the schema) — the classifier's GraphQL pattern is
  best-effort, flagged in comments for tightening against the first real
  occurrence.

## Known gap (documented, tracked separately)

The `loom-daemon` native `forge auto-merge` path (preferred whenever
`loom-daemon` is on `PATH` — the common case) has no equivalent
`expectedHeadOid`/`sha` precondition flag. This means the head-moved guard
added here fully protects the always-shell synchronous `forge_merge_pr` path
and the Gitea/loom-daemon-absent `forge_auto_merge` fallback, but a GitHub
merge routed through the native `loom-daemon` auto-merge call can still
silently squash whatever the current head is. Documented inline in
`merge-pr.sh` and tracked in #5589.

## Test plan

- [x] New `test-merge-pr-head-mismatch.sh`: 23/23 passing.
- [x] All 15 pre-existing `test-merge-pr-*.sh` suites still pass (no
      regressions in the retry-loop/classifier ordering change).
- [x] `test-forge-helpers.sh` and `test-forge-helpers-rate-limit-fallback.sh`
      still pass.
- [x] `check-ci-suite-manifest.sh`: new suite correctly wired into
      `ci-wired.txt`.
- [x] `shellcheck` on all modified/new shell files: no new errors (pre-existing
      style-only SC2016/SC2015/SC2002 infos, none introduced by this change).
- [x] `bash -n` syntax check on all modified/new shell files.
- Manual end-to-end verification against a live GitHub race (push mid-merge)
  is not practical in CI — per the issue's own test plan, favored testing the
  shell logic/parsing in isolation instead, matching the existing
  `test-merge-pr-auto-reconcile.sh` "extract from source" strategy.

## Edge cases from the issue's test plan

- (a) Benign/identical-content head moves still conservatively re-queue — the
  guard doesn't attempt to distinguish "safe" moves; any head-SHA mismatch is
  exit 3.
- (b) Gitea: confirmed the equivalent precondition (`head_commit_id`) exists
  and is wired — not a documented gap after all (the curated issue's
  "suspected cause" was more pessimistic than the actual Gitea API surface).
- (c) `--auto` GitHub auto-merge-queue path: `expectedHeadOid` is an input to
  `enablePullRequestAutoMerge` itself (enable-time), not the eventual queued
  merge execution. Whether it's re-checked at actual-merge time (which can
  happen well after the mutation returns) could not be confirmed against
  public docs — flagged inline as a residual risk on that specific path,
  distinct from the documented `loom-daemon` native-path gap above.

Closes #5579

### DIFF
--- a/defaults/.claude/commands/loom/champion-pr-merge.md
+++ b/defaults/.claude/commands/loom/champion-pr-merge.md
@@ -857,10 +857,28 @@ git checkout main 2>/dev/null || true
 
 # Use merge-pr.sh for worktree-safe merge via GitHub API
 # --auto enables auto-merge if ruleset requires wait
-./.loom/scripts/merge-pr.sh "$PR_NUMBER" --auto || {
+#
+# merge-pr.sh reads the PR's head SHA itself (a fresh, uncached read — see
+# "Cached forge reads" above) immediately before merging, and passes it
+# through to the forge's merge API as an optimistic-concurrency precondition
+# (#5579). Capture the exit code rather than using a bare `||`: exit 3 is a
+# DISTINCT outcome from exit 1 and must not be handled as a failure (below).
+MERGE_RC=0
+./.loom/scripts/merge-pr.sh "$PR_NUMBER" --auto || MERGE_RC=$?
+
+if [ "$MERGE_RC" -eq 3 ]; then
+  # #5579: the PR's head branch moved past the SHA this merge attempt gated
+  # on — most commonly a session pushing new commits to an open, loom:pr
+  # branch while Champion was running. This is NOT a merge failure: the PR
+  # is still Judge-approved, its diff just changed underneath it.
+  #
+  # Do NOT follow the failure steps below for this outcome — see the "Exit
+  # code 3" exception in "Error Handling".
+  echo "PR #$PR_NUMBER head moved during merge attempt — re-queuing for a fresh pass instead of failing"
+elif [ "$MERGE_RC" -ne 0 ]; then
   echo "Merge failed for PR #$PR_NUMBER"
   # Post failure comment (see Error Handling section)
-}
+fi
 ```
 
 **Merge strategy**:
@@ -868,6 +886,9 @@ git checkout main 2>/dev/null || true
 - **Squash merge**: Combines all commits into single commit (clean history)
 - **`--auto`**: Enables GitHub's auto-merge if ruleset requires wait
 - Branch deleted automatically after merge
+- **Head-moved guard (#5579)**: `merge-pr.sh` refuses to merge (exit 3, not a
+  failure) if the PR's head branch advanced past the SHA it read immediately
+  before merging — see "Exit code 3" in "Error Handling" below
 
 ### Step 4: Verify Issue Auto-Close
 
@@ -1707,6 +1728,33 @@ This PR met all safety criteria but the merge operation failed. A human will nee
 ---
 *Automated by Champion role*"
 ```
+
+### Exception: exit code 3 — head moved, re-queue, not a failure (#5579)
+
+`merge-pr.sh` exits **3** (distinct from the generic failure exit **1**) when
+the PR's head branch changed between the fresh head-SHA read it took
+immediately before merging and the actual merge call — most commonly because
+a session pushed new commits to an open, `loom:pr`-labeled branch while
+Champion was running. **Do not follow the 5 failure steps above for this
+outcome:**
+
+- Do **not** post the "Merge Failed" comment — the PR is still Judge-approved,
+  its diff just moved out from under the merge attempt.
+- Do **not** count it as an error in the completion summary.
+- Leave `loom:pr` in place and move on to the next PR in the queue. A later
+  Champion pass will pick this PR up fresh — its safety criteria (including
+  `updatedAt` and CI status) will naturally re-evaluate the new head before
+  merging it.
+
+**Squash-merge detection trap.** If you ever need to manually verify whether a
+re-queued (or, worse, an already-merged-before-this-fix) PR's commits actually
+landed vs. were silently stranded, `git merge-base --is-ancestor <commit>
+origin/main` is **not reliable evidence either way**: a squash merge produces
+a brand-new commit SHA on `main` that is not a git-ancestry descendant of any
+commit on the original PR branch, regardless of whether that commit's content
+made it into the squash or was left behind. There is no cheap ancestry check
+for "squashed-and-landed" vs. "stranded" — verification requires diffing the
+actual file content on `main` against the branch/commit in question.
 
 ---
 
--- a/defaults/scripts/lib/forge-helpers.sh
+++ b/defaults/scripts/lib/forge-helpers.sh
@@ -305,21 +305,58 @@ forge_split_nwo() {
 # --- Forge-Dispatched Operations ---
 
 # Merge a PR via the forge API.
-# Usage: forge_merge_pr NWO PR_NUMBER
+# Usage: forge_merge_pr NWO PR_NUMBER [EXPECTED_HEAD_SHA]
 # GitHub: PUT /repos/{nwo}/pulls/{n}/merge with merge_method=squash
 # Gitea: POST /repos/{owner}/{repo}/pulls/{n}/merge with Do=squash
+#
+# EXPECTED_HEAD_SHA (optional, #5579): an optimistic-concurrency precondition —
+# the SHA the PR's head branch must currently match for the merge to proceed.
+# Without it, both forges will happily squash-merge whatever the CURRENT head
+# is at the moment the request lands, even if it has commits the caller never
+# saw approved (silently stranding them — squash-merge makes this invisible to
+# an ancestry check afterward, since the new squash commit is not a descendant
+# of the stranded commits either way). Pass the freshest possible head-SHA read
+# (never a cached one) immediately before calling this.
+#
+# GitHub: REST's optional `sha` field. Verified (GitHub's public OpenAPI spec,
+# 2026-08-07) to fail with HTTP 409 and message "Head branch was modified.
+# Review and try the merge again." on a mismatch — a DIFFERENT string from the
+# existing "Base branch was modified" retry case handled elsewhere in
+# merge-pr.sh; callers must not conflate the two (that one means "rebase onto
+# base and retry"; this one means "the approved diff moved out from under us,
+# do not retry-and-merge-anyway").
+#
+# Gitea: the `head_commit_id` field on MergePullRequestOption (confirmed
+# present via Gitea/Forgejo's published swagger.v1.json and upstream
+# services/pull/merge_prepare.go, 2026-08-07). A mismatch raises
+# ErrSHADoesNotMatch, which routers/api/v1/repo/pull.go maps to HTTP 409 with
+# message "head out of date".
 forge_merge_pr() {
   local nwo="$1"
   local pr_number="$2"
+  local expected_head_sha="${3:-}"
 
   if [[ "$FORGE_TYPE" == "gitea" ]]; then
     forge_split_nwo "$nwo"
-    gitea_api POST "repos/$FORGE_OWNER/$FORGE_REPO/pulls/$pr_number/merge" \
-      -d '{"Do":"squash","delete_branch_after_merge":false}'
+    if [[ -n "$expected_head_sha" ]]; then
+      gitea_api POST "repos/$FORGE_OWNER/$FORGE_REPO/pulls/$pr_number/merge" \
+        -d "$(jq -nc --arg sha "$expected_head_sha" \
+          '{"Do":"squash","delete_branch_after_merge":false,"head_commit_id":$sha}')"
+    else
+      gitea_api POST "repos/$FORGE_OWNER/$FORGE_REPO/pulls/$pr_number/merge" \
+        -d '{"Do":"squash","delete_branch_after_merge":false}'
+    fi
   else
-    gh api "repos/$nwo/pulls/$pr_number/merge" \
-      -X PUT \
-      -f merge_method=squash 2>&1
+    if [[ -n "$expected_head_sha" ]]; then
+      gh api "repos/$nwo/pulls/$pr_number/merge" \
+        -X PUT \
+        -f merge_method=squash \
+        -f sha="$expected_head_sha" 2>&1
+    else
+      gh api "repos/$nwo/pulls/$pr_number/merge" \
+        -X PUT \
+        -f merge_method=squash 2>&1
+    fi
   fi
 }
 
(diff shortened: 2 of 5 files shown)


rjwalters:
LGTM — approving.

**Correctness vs. issue #5579's acceptance criteria**: all four checkboxes addressed. The precondition-threading approach (direction 1 from the curated issue) is implemented for the always-shell `forge_merge_pr` sync path and the Gitea/loom-daemon-absent `forge_auto_merge` fallback: GitHub REST `sha`, GitHub GraphQL `expectedHeadOid`, and Gitea `head_commit_id` are all threaded correctly and backward-compatible (omitted entirely when no SHA is supplied). The new `_is_head_mismatch_response`/`error_head_moved` exit-3 path is placed *after* the existing "already merged" and "merge already in progress" race rechecks but *before* the "Base branch was modified" retry branch in both loops — correct ordering, verified in both the synchronous and `--auto` retry loops in `merge-pr.sh`. `champion-pr-merge.md` Step 3 captures the exit code and correctly special-cases 3 (re-queue, no failure comment, `loom:pr` left in place) vs. 1 (existing failure path). The squash-merge-detection-trap note (4th checkbox) is documented in both the script header and the champion doc.

**Known gap, honestly scoped**: the native `loom-daemon forge auto-merge` path (preferred whenever `loom-daemon` is on `PATH`) has no equivalent precondition and is documented inline plus tracked in a real follow-up issue (#5589, confirmed open) rather than silently left unguarded. Reasonable to ship the shell-path fix now rather than block on the daemon-side work.

**Tests**: ran the new `test-merge-pr-head-mismatch.sh` directly — 23/23 pass, genuinely exercising the claimed behavior (stubs `gh`/`curl` and asserts on actual argv/POST-body content for the SHA threading, extracts+sources the real classifier/exit-code functions from `merge-pr.sh` rather than reimplementing them, and asserts retry-loop ordering via `awk` against the live source — not just string-matching for presence). Also ran all 15 pre-existing `test-merge-pr-*.sh` suites plus `test-forge-helpers.sh` and `test-forge-helpers-rate-limit-fallback.sh` — no regressions. `shellcheck` on all three modified/new shell files shows only pre-existing info-level SC2016 (single-quoted GraphQL queries, expected) and SC2015 — nothing new.

**CI**: all checks green (Shellcheck, Shell Test Suites hermetic — 6m45s, Installer Integration Tests, etc.), `mergeStateStatus` is CLEAN.

No blocking concerns. Nice, thorough writeup of the API research (GitHub OpenAPI spec, Gitea/Forgejo swagger + source) with honest flagging of the one unverified bit (GraphQL mismatch error text is best-effort pending a live occurrence).

PR #6725, opened 2026-08-22
fix: PROMPT_DIR/SWEEP_MD/CHAMPION_MD resolve correctly in both source and installed layouts

rjwalters:
## Summary

Four CI-wired test suites under `defaults/scripts/tests/` compute the path
to `.claude/commands/loom` from `HELPERS_DIR` using a single hard-coded
number of `..` segments:

- `test-app-permission-fallback.sh:410`
- `test-forge-helpers-rate-limit-fallback.sh:468`
- `test-forge-transient-classification.sh:226`
- `test-merge-pr-head-mismatch.sh:354`

That depth is only correct for one of the two layouts these suites actually
run under:

- **In this source repo**, the suites live at `defaults/scripts/tests/`, and
  `defaults/.claude/commands/loom` exists as a sibling of `defaults/scripts`
  — one `..` reaches it. This is a coincidence of this repo's own layout.
- **In an installed consumer repo** (e.g. `2AMLogic/2am`), the same suites
  are copied to `.loom/scripts/tests/`, and only the repo-root
  `.claude/commands/loom` exists — two `..` are needed. Without the fix,
  the `cd` fails outright and the whole suite is reported failed by
  `run-ci-suites.sh`.

A single fixed depth can't satisfy both. The originally-proposed fix (always
use two `..`) regresses this repo's own run — confirmed locally, it drops
2/4 suites back to failing (`cd: ... No such file or directory` /
2 silent `[[ -f ]]` FAILs) when hard-coded at two `..`.

## Fix

Probe for the two-`..` path first (installed-copy layout) and fall back to
one `..` (source-repo layout), so the same script resolves correctly
whichever layout it is running in.

## Verification

All four suites pass in this repo's own layout with the change applied:
`47/47`, `65/65`, `25/25`, `32/32`.

Reported in 2AMLogic/2am#447, which is blocked on this merging + a resync.

## Test plan

- [x] Ran all four affected suites directly against this branch — 0 failures
- [x] Confirmed the naive two-`..`-only fix regresses this repo's own run before settling on the dual-candidate probe


rjwalters:
## Judge review: request changes

**Fix logic — verified correct for both layouts.**

I reviewed the diff (`gh pr diff 6725`) and ran all four affected suites directly in this repo's own layout (`defaults/scripts/tests/*.sh`, where `HELPERS_DIR` = `defaults/scripts`, one `..` from repo layout):

```
test-app-permission-fallback.sh:            47/47 passed
test-forge-helpers-rate-limit-fallback.sh:  65/65 passed
test-forge-transient-classification.sh:     25/25 passed
test-merge-pr-head-mismatch.sh:             32/32 passed
```

The two-`..`-first, one-`..`-fallback probe is correct: it resolves to the installed-copy layout (`.loom/scripts/tests/` → repo-root `.claude/commands/loom`) when that path exists, and falls back to this source repo's own layout otherwise. That's the right precedence given the PR's own reasoning — an installed-copy's two-`..` path can't accidentally collide with anything in this repo, so probing it first is safe.

**Edge case (neither path exists):** two of the four sites (`test-merge-pr-head-mismatch.sh`, `test-forge-transient-classification.sh`) guard with `[[ -f/-d ... ]]` before assigning the final variable, so a total-miss degrades to a normal, readable test failure/assertion. The other two (`test-app-permission-fallback.sh`, `test-forge-helpers-rate-limit-fallback.sh`) still do a bare `cd ... && pwd` in the fallback branch with no existence check, so a total-miss under `set -euo pipefail` aborts the whole script on a `cd: No such file or directory` — but that's unchanged from the pre-fix behavior (already a bare `cd` with no guard), so it's not a regression introduced here, just a preexisting rough edge. Not blocking.

**Blocking issue: CI is red, and not spuriously.**

`gh pr checks 6725` shows `defaults/ Changes Require a VERSION Bump` as a hard `fail` (not pending/flaky — confirmed the diff touches `defaults/scripts/tests/*.sh` with no `VERSION` bump and no `<!-- loom:no-surface-change -->` marker in the PR body or commit messages). This gate exists precisely for changes like this one: the PR body itself says the fix is needed because installed consumer copies (`2AMLogic/2am#447`) are broken and blocked on "this merging + a resync" — i.e., this *does* change installed-surface behavior, which is exactly what the VERSION-bump gate is meant to catch and signal to consumers (fleet resync, `/repo:update-tools`, etc.). The `no-surface-change` marker would be the wrong exemption to reach for here since behavior is changing; a real `VERSION` bump is the correct fix.

`mergeStateStatus` is `UNSTABLE` with this check in `fail` state. Per the Judge role's CI-verification requirement, I can't approve until this is green.

**Ask:** bump `VERSION` (patch bump is fine) in this PR before merge. Everything else — the fix itself, its correctness for both layouts, and the test suite passes — looks solid.

rjwalters:
Doctor: bumped VERSION 0.18.130 → 0.18.131 via ./scripts/version.sh bump patch (all version-bearing files back in sync, self-check passed) and added a CHANGELOG entry under Fixed. This addresses the failing `defaults/ Changes Require a VERSION Bump` check. Re-requesting review.

PR #6751, opened 2026-08-22
fix(worktree-reaper): resolve PR status correctly when a branch has multiple PRs

### LINKED ISSUE #6746: worktree reaper: gh pr list --limit 1 can misclassify a branch with multiple PRs (merged + later closed-without-merge)
## Context

Split out of #6653's curation (item 3 of the curator's findings), which fixed the `PrStatus::NoPr` grace-period gap and added quarantine-then-reclaim for a dirty closed-issue worktree. This is the one item from that curation explicitly flagged as optional/separate.

## Problem

`gh_pr_list_by_head` / `rows_to_status` in `loom-daemon/src/worktree_ops/clean.rs` (~line 285-342) queries:

```
gh pr list --head <branch> --state all --json number,state,mergedAt,closedAt --limit 1
```

and `rows_to_status` takes only the **first** row returned. When a branch has had multiple PRs opened against it over its lifetime (e.g. a merged PR, then later a second PR opened from the same branch name that was closed without merging), the `--limit 1` cap can return the wrong one — observed live for `issue-5179` on one host during #6653's curation.

The consequence: `classify_worktree` can classify a worktree as `ClosedNoMerge` (30-day grace period, `CLOSED_NO_MERGE_GRACE_PERIOD_SECS`) instead of `Merged` (10-minute default grace period), needlessly holding the worktree ~30 days longer than necessary — not a safety violation (fails toward preserving the worktree longer), but a real reclaim-latency bug.

## Suggested fix

`gh_pr_list_by_head` should either:
- Drop `--limit 1` and fetch all PRs for the branch, then have `rows_to_status` prefer a `Merged` row over a `ClosedNoMerge` row if both exist (a branch that was ever merged should never be treated as merely closed-without-merge), or
- Otherwise define a clear preference order across multiple rows (e.g. most-recently-updated, or Merged > Open > ClosedNoMerge).

This also affects `check_pr_status_for_branch` / `check_pr_merged_rest`'s REST counterpart (`check_pr_status_for_branch_rest`, which already uses `per_page=1`) — same fix shape applies there.

## Acceptance criteria

- [ ] `rows_to_status` (or its caller) is proven correct when a branch has multiple PRs, preferring `Merged` over any `ClosedNoMerge`/`Open` row
- [ ] Unit test: multiple rows where an earlier one is `MERGED` and a later one is `CLOSED` (no merge) — result must be `PrStatus::Merged`
- [ ] REST counterpart (`check_pr_status_for_branch_rest`) gets the equivalent fix or is confirmed already safe by its own `per_page=1` + row-ordering guarantee (verify GitHub's REST API default sort order for this endpoint)

## References

- #6653 (curation notes, item 3)
- `loom-daemon/src/worktree_ops/clean.rs`: `gh_pr_list_by_head`, `rows_to_status`, `check_pr_status_for_branch_rest`

---

### Verified corrections

- **2026-08-22, verified live against `rjwalters/loom`** (not a synthetic repro — this is the exact `feature/issue-5179` branch the issue body cites as the original observation): both probe paths return the wrong row today.

  ```
  $ gh pr list --head feature/issue-5179 --state all --json number,state,mergedAt,closedAt --limit 1
  [{"number":5199,"state":"CLOSED","mergedAt":null,"closedAt":"2026-08-04T02:43:56Z"}]

  $ gh api "repos/rjwalters/loom/pulls?state=all&head=rjwalters:feature/issue-5179&per_page=10" \
      --jq '.[] | {number, state, created_at, merged_at, closed_at}'
  {"number":5199,"state":"closed","created_at":"2026-08-04T02:42:32Z","merged_at":null,"closed_at":"2026-08-04T02:43:56Z"}
  {"number":5190,"state":"closed","created_at":"2026-08-04T02:26:54Z","merged_at":"2026-08-04T02:33:48Z","closed_at":null}
  ```

  Both `gh pr list --head ... --limit 1` (GraphQL-backed) and the raw REST list
  endpoint (`pulls?state=all&head=...`, unpaginated) return PR #5199
  (`ClosedNoMerge`) ahead of the earlier, actually-merged #5190. Both APIs'
  default order is most-recently-created first — `--limit 1`/`per_page=1`
  picks up whichever PR against the branch was opened *last*, not whichever one
  actually merged.

- **This answers the issue's own 3rd acceptance criterion directly**: the REST
  counterpart (`check_pr_status_for_branch_rest`, `per_page=1`) is **not**
  "already safe by its own row-ordering guarantee" — it has the identical
  misclassification exposure as the GraphQL path, confirmed by the same live
  query above. The AC's "or confirmed already safe" branch does not apply;
  the fix is required on both paths.

### Affected Files

(verified present on `origin/main` @ `5a321cdc2fe9598174aead7bdc3e9dd6c4e3769b`, 2026-08-22 — re-verify line numbers before relying on them if this issue sits unbuilt for long)

- `loom-daemon/src/worktree_ops/clean.rs`:
  - `gh_pr_list_by_head()` (~line 285-301) — drop `--limit 1` or otherwise fetch enough rows to see every PR against the branch
  - `rows_to_status()` (~line 324-342) — currently takes only `rows.into_iter().next()`; needs to scan all returned rows and prefer a `Merged` row over any `ClosedNoMerge`/`Open` row when both are present
  - `check_pr_status_for_branch_rest()` (~line 432-452) and `classify_pr_row()` (~line 458-471) — REST counterpart; per the verified correction above, this needs the equivalent multi-row handling, not just a confirmation that `per_page=1` is already safe
  - `gh_pr_list_by_issue_search()` (~line 303-319, the `check_pr_merged()` fallback path when `--head` returns nothing) is not separately named in the issue's suggested fix, but shares `rows_to_status()` as its consumer — fixing `rows_to_status()` itself (rather than only its caller) covers this fallback for free, so no separate fix is needed here.
  - Not affected: `check_pr_status_by_number_rest()` (~line 487) queries a single PR by number (`pulls/<n>`), not a list — no multi-row ambiguity exists on that path.

### Related Open Work

- PR #6747 (open, not yet merged as of `5a321cdc`) cross-references this issue in its own body ("Follow-up filed" section), explicitly splitting this exact scope out as "a separate, optional fix per the curation notes" rather than folding it in. Confirmed via `gh pr diff 6747 --name-only`: it touches `clean.rs` but for the unrelated `NoPr`/quarantine work (`classify_worktree`, `quarantine_dirty_worktree`), not `gh_pr_list_by_head`/`rows_to_status`/`check_pr_status_for_branch_rest`. No spec conflict, no dependency — disregarded as already-accounted-for (this issue's own "Context" section already names #6653/the split), and safe to build independently of whether #6747 has merged yet.

### Test Plan

- [ ] Automated (unit): a new `rows_to_status` test with `Vec<PrRow>` in the specific order the live repro above shows the API return them — `[ClosedNoMerge-row (newer), Merged-row (older)]` — asserting the result is `PrStatus::Merged`, not `PrStatus::ClosedNoMerge`. `PrRow` derives `Deserialize`, so rows can be constructed via `serde_json::from_str` in-test without a live `gh` call.
- [ ] Automated (unit): equivalent case for `classify_pr_row`/the REST row-selection path once it also scans multiple rows — same ordering, same expected `Merged` result.
- [ ] Automated (unit): a row set with only `Open` + `ClosedNoMerge` (no `Merged` present) still resolves via whatever preference order is chosen (e.g. `Open` over `ClosedNoMerge`, or most-recent) — pick and document the order explicitly since the "Merged always wins" rule doesn't disambiguate this case.
- [ ] Manual (optional): re-run the two live commands in "Verified corrections" above against `feature/issue-5179` after the fix lands, this time via the fixed code path (`check_pr_status_for_branch`), and confirm it now resolves to `Merged`.

Marked `routine`: the fix is bounded to two/three known functions with an existing (if now-known-wrong) unit-testable contract; a mistake here would surface as a failing test or a worktree held ~30 days too long — not a silent catastrophic regression (the issue's own body already notes the failure mode "fails toward preserving the worktree longer," not toward premature deletion).

rjwalters:
## Summary

`gh_pr_list_by_head` (GraphQL, `--limit 1`) and `check_pr_status_for_branch_rest`
(REST, `per_page=1`) both default to newest-created-first. When a branch has
had more than one PR opened against it over its lifetime — e.g. an older
merged PR, then a later unrelated PR from the same branch name that closed
without merging — the `--limit 1`/`per_page=1` cap picked up whichever PR was
opened *last*, not whichever one actually merged. The consequence:
`classify_worktree` could misclassify a `Merged` worktree as `ClosedNoMerge`,
needlessly holding it behind the 30-day grace period instead of the 10-minute
one.

Verified live against `feature/issue-5179` in the issue's own curation notes
(both the GraphQL and REST probes returned the wrong PR).

## Changes

- `loom-daemon/src/worktree_ops/clean.rs`:
  - Dropped `--limit 1` from `gh_pr_list_by_head` / `gh_pr_list_by_issue_search`,
    and raised `check_pr_status_for_branch_rest`'s `per_page` from `1` to `30`
    — both now fetch every PR against the branch instead of just the newest.
  - Added `select_pr_status`, a pure row-selection helper: `Merged` always
    wins over any number of other rows, then `Open`, then the forge's own
    (newest-first) order among `ClosedNoMerge`/`Unknown` rows.
  - `rows_to_status` (GraphQL path) and `check_pr_status_for_branch_rest`
    (REST path) both now route through `select_pr_status` instead of taking
    only the first row.

## Test plan

- [x] New unit test: multiple rows where an earlier one is `MERGED` and a
      later one is `CLOSED` (no merge, GitHub's default newest-first order) —
      result is `PrStatus::Merged`, both regardless of row order and via the
      REST-side `classify_pr_row` path.
- [x] New unit test: `Open` beats an older `ClosedNoMerge` row when no
      `Merged` row is present.
- [x] New unit test: all-`ClosedNoMerge` rows fall back to the first
      (newest-created) row.
- [x] Existing empty-rows (`NoPr`) / probe-failure (`Unknown`) behavior
      preserved.
- [x] `cargo test --lib worktree_ops::clean` (81 tests) and the full
      `cargo test --lib` (4506 tests) both pass.
- [x] `cargo clippy --lib --all-targets` clean.

Closes #6746

### DIFF
--- a/loom-daemon/src/worktree_ops/clean.rs
+++ b/loom-daemon/src/worktree_ops/clean.rs
@@ -308,8 +308,6 @@ fn gh_pr_list_by_head(repo_root: &Path, branch: &str) -> Option<Vec<PrRow>> {
             "all",
             "--json",
             "number,state,mergedAt,closedAt",
-            "--limit",
-            "1",
         ],
     )
 }
@@ -326,33 +324,73 @@ fn gh_pr_list_by_issue_search(repo_root: &Path, issue_num: u32) -> Option<Vec<Pr
             "all",
             "--json",
             "number,state,mergedAt,closedAt",
-            "--limit",
-            "1",
         ],
     )
+}
+
+/// Resolve a list of same-branch PR rows (in whatever order the forge
+/// returned them) to a single [`PrStatus`] (issue #6746): a branch can have
+/// had more than one PR opened against it over its lifetime — e.g. a merged
+/// PR, then later a second, unrelated PR opened from the same branch name
+/// that was closed without merging (observed live for `feature/issue-5179`
+/// during #6653's curation). Both `gh pr list --head` (GraphQL) and the REST
+/// `pulls?head=` list endpoint default to newest-created-first, so naively
+/// taking "the first row" picks whichever PR was opened *last* against the
+/// branch, not whichever one is actually relevant.
+///
+/// Preference order:
+/// 1. `Merged` always wins, over any number of other rows — a branch that was
+///    ever merged must never be reported as merely closed-without-merge or
+///    open just because a later PR against the same branch name says
+///    otherwise.
+/// 2. Otherwise `Open` wins — an actively-reviewed PR should not be shadowed
+///    by an older closed one.
+/// 3. Otherwise the first row in forge-returned order wins (i.e., given the
+///    newest-first default ordering both probes rely on, the most recently
+///    closed PR) — covers `ClosedNoMerge`/`Unknown` rows.
+///
+/// An empty row list resolves to [`PrStatus::NoPr`].
+fn select_pr_status<I: IntoIterator<Item = PrStatus>>(rows: I) -> PrStatus {
+    let mut best: Option<PrStatus> = None;
+    for status in rows {
+        if matches!(status, PrStatus::Merged { .. }) {
+            return status;
+        }
+        match &best {
+            None => best = Some(status),
+            Some(PrStatus::Open) => {} // already the best short of Merged
+            Some(_) => {
+                if matches!(status, PrStatus::Open) {
+                    best = Some(status);
+                }
+            }
+        }
+    }
+    best.unwrap_or(PrStatus::NoPr)
 }
 
 /// Map an optional `gh pr list --json number,state,mergedAt,closedAt` result
 /// onto a [`PrStatus`]: `None` (the `gh` call itself failed or returned
 /// unparseable JSON) is `Unknown`; an empty (but successful) result is `NoPr`.
+/// When multiple rows are present, see [`select_pr_status`] for the
+/// preference order (issue #6746).
 fn rows_to_status(rows: Option<Vec<PrRow>>) -> PrStatus {
     let Some(rows) = rows else {
         return PrStatus::Unknown;
     };
-    let Some(row) = rows.into_iter().next() else {
-        return PrStatus::NoPr;
-    };
-    if let Some(merged_at) = row.merged_at {
-        PrStatus::Merged { merged_at }
-    } else if row.state == "CLOSED" {
-        PrStatus::ClosedNoMerge {
-            closed_at: row.closed_at,
+    select_pr_status(rows.into_iter().map(|row| {
+        if let Some(merged_at) = row.merged_at {
+            PrStatus::Merged { merged_at }
+        } else if row.state == "CLOSED" {
+            PrStatus::ClosedNoMerge {
+                closed_at: row.closed_at,
+            }
+        } else if row.state == "OPEN" {
+            PrStatus::Open
+        } else {
+            PrStatus::Unknown
         }
-    } else if row.state == "OPEN" {
-        PrStatus::Open
-    } else {
-        PrStatus::Unknown
-    }
+    }))
 }
 
 /// Check the PR status for `issue_num`'s branch. Thin `gh` wrapper mirroring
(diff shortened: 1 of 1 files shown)


rjwalters:
LGTM. Reviewed the diff and verified against issue #6746's acceptance criteria:

- `select_pr_status` gives an explicit, well-documented preference order (Merged > Open > first/newest of ClosedNoMerge/Unknown), and both `rows_to_status` (GraphQL path) and `check_pr_status_for_branch_rest` (REST path) now route through it instead of taking only the first row.
- `--limit 1` dropped from `gh_pr_list_by_head`/`gh_pr_list_by_issue_search`, and REST `per_page` raised from 1 to 30 — both probes now see every PR against the branch, not just the newest.
- Confirmed the exact acceptance-criteria test exists and passes: `rows_to_status_prefers_merged_over_a_later_closed_no_merge_row` (rows in forge's newest-first order: closed-no-merge first, merged second) resolves to `PrStatus::Merged`, plus a row-order-reversed variant, an Open-beats-older-ClosedNoMerge case, an all-ClosedNoMerge fallback-to-first case, and a REST-path (`classify_pr_row`) equivalent — 8 new tests total.
- Ran `cargo test --lib -p loom-daemon worktree_ops::clean` locally in a checked-out worktree at the PR head SHA: 81/81 pass, including all new tests. `cargo clippy -p loom-daemon --lib --all-targets` is clean.
- Confirmed `check_pr_status_by_number_rest` (single-PR `pulls/<n>` lookup) is correctly out of scope — it queries one object, not a list, so it was never exposed to this bug.
- CI: all required checks pass (Rust Unit Tests, Rust Linting & Build Check, Rust OTLP Feature Build & Test, etc.). `mergeStateStatus` is CLEAN.

No concerns. Approving.

PR #7714, opened 2026-09-15
feat(ci): ratchet oversized source files instead of refactor-on-touch (#7711)

### LINKED ISSUE #7711: Establish a file-size policy: ratchet oversized Rust/shell files, tier the shell→Rust port
## Problem

Loom's source files grow monotonically because every individual addition is
defensible while the aggregate is the problem — the same failure mode
`check-claude-md-budget.sh` (#4014) was built to counter, now playing out in
`loom-daemon/src/*.rs` and `defaults/scripts/*.sh`.

### Measured state (2026-09-15, `main` @ 38f85147)

**Rust** — 228 files, 289,955 LOC:

| Threshold | Files | Share of all Rust LOC |
|---|---|---|
| > 1000 lines | 86 (37.7%) | **79.0%** |
| > 3000 lines | 20 (8.8%) | 40.8% |
| > 5000 lines | 12 (5.3%) | 30.5% |

Largest: `ipc.rs` 10,849 · `role_runner.rs` 9,949 · `work_finder.rs` 9,157 ·
`safehouse.rs` 9,092 · `sweep_registry/dispatch.rs` 8,343.

**These files reached their entire current size within 90 days, by pure accretion:**

```
net  +9430  (+10048/-618)   ipc.rs          112 commits in 90d
net  +9949  (+10378/-429)   role_runner.rs   47 commits
net  +9157  (+10432/-1275)  work_finder.rs   73 commits
net  +9092   (+9534/-442)   safehouse.rs     31 commits
net  +8343   (+8582/-239)   dispatch.rs      33 commits
```

Add-to-delete ratio ≈ **20:1**. Agents add; they essentially never restructure.
Mean touch is +160/−24 lines.

**Two findings reshape the problem:**

1. **49% of the bulk is inline `#[cfg(test)]`** — 142,674 of 289,955 lines.
   `role_runner.rs` is 9,949 lines but only **893** are production code;
   `work_finder.rs` is 9,157 with **1,003** production. Extracting test modules
   to sibling files drops the over-1000 count from 86 files to **37**, and is
   compiler-verified (zero semantic risk).

2. **Shell is a parallel problem of comparable size** — 456 files, 207,441 LOC,
   of which 114,982 (55%) is test scripts. Largest: `guard-destructive-generic.sh`
   8,872 · `tests/hooks/test-guard-destructive.sh` 8,826 ·
   `test-loom-daemon-update.sh` 6,232 · `loom-daemon-update.sh` 3,536 ·
   `merge-pr.sh` 3,148 · `worktree.sh` 3,050.

## Why this matters for agentic coding

Research anchors (no authoritative Rust rule exists):

- **Clippy's proposed `too_many_lines_in_file`**
  ([rust-clippy#16674](https://github.com/rust-lang/rust-clippy/issues/16674)) —
  default **1000 lines**, counting code only (blanks and comment-only lines
  skipped). The most defensible Rust-native number.
- **rustc `tidy`** enforces 3000 lines, but only for non-Rust files.
- Empirical context-bloat work ([arXiv:2606.15828](https://arxiv.org/pdf/2606.15828))
  observed problem files spanning 216–1,477 LOC, which brackets 1000.
- The widely-cited "150–500 line sweet spot" is a Python blog post, not research,
  and is punishing for Rust (inline tests, derives, explicit error handling).

### 1. A ratchet, not a limit (load-bearing)

Mirror `check-claude-md-budget.sh`: **a file over threshold may not grow.**
Files under it are unconstrained; files over are frozen at their current size.

- No big-bang refactor required, self-tightening over time.
- Fires precisely when an agent is making a known-bad file worse.
- Escape valve is cheap in Rust: put the new handler in a new sibling module and
  leave a match arm behind.

Threshold: **1000 production lines** (measured before the first
`^#\[cfg(test)\]`), matching Clippy's proposed default.

### 2. Explicitly NOT "touch a large file → refactor it"

That rule was considered and rejected:

- **It fires constantly.** `ipc.rs` is touched >1×/day (112 commits/90d);
  `types.rs` 94. That is a tax on nearly every sweep, not occasional cleanup —
  agents will either bloat every PR or quietly ignore the rule.
- **It couples unrelated refactors into feature PRs.** A mandated split turns a
  +160/−24 touch into a four-figure diff, degrading Judge review and burying the
  actual change.
- **It fights the concurrency model.** Parallel builders work in separate
  worktrees; a module split moves `use` statements and `pub(crate)` visibility
  across the file, textually conflicting with every other in-flight PR touching
  it — worst exactly when the fleet is busiest.

### 3. Exemptions the check MUST carry

- **Vendored files** — `defaults/hooks/guard-destructive-generic.sh` (8,872, the
  single largest shell file) and `guard-destructive.sh` are vendored from
  rjwalters/repo and re-vendored at release time; the header explicitly forbids
  hand-editing generic pattern behavior. Also `resync-installed.sh` and the
  vendored test scripts. Without this carve-out the check generates permanently
  unfixable violations.
- **Installed mirrors** — `.loom/hooks/`, `.loom/scripts/` etc. are resync copies
  of `defaults/`; check the source, not the mirror, or every violation
  double-counts.
- **Bootstrap floor** — `install.sh` (2,514), `scripts/install-loom.sh` (2,457),
  `loom-daemon-update.sh` (3,536), `loom-daemon-start.sh` (2,728) run before or
  manage the `loom-daemon` binary and cannot depend on it. Ratchet them; never
  port them.

## Shell: port to Rust, tiered

There is established precedent — #4552 (`loom_tools` script-helper family →
native subcommands), #4471 (`agent_spawn.py`/`agent_wait.py` → native),
#4105/#4108 (`tokens bootstrap`/`check`). The target is a `loom-daemon`
subcommand under `src/cli/` or `src/script_helpers/`.

| Tier | Files | Action |
|---|---|---|
| Vendored | `guard-destructive*.sh` (~9k LOC) | **Exempt** — upstream owns it |
| Bootstrap | `install.sh`, `install-loom.sh`, `loom-daemon-{update,start}.sh` (~11k) | **Ratchet only**, never port (chicken-and-egg) |
| Port candidates | `merge-pr.sh` (3,148), `worktree.sh` (3,050), `claude-wrapper.sh` (2,932) | **Finish the port** — all three already have a `loom-daemon` delegation ladder (`loom-daemon forge`), so this is incremental, not a rewrite |
| Shell tests | 242 files, 114,982 LOC | **Split, don't port** — bash-testing-bash is legitimate; cap size only |

Porting also *improves* hook latency: `PreToolUse` fires on every tool call, and
a binary exec beats parsing a multi-thousand-line bash script — though the
vendoring constraint blocks this for the guard specifically.

## Acceptance criteria

- [ ] `scripts/check-file-size-budget.sh` exists, modeled on
      `check-claude-md-budget.sh` (same header-comment discipline: state the
      rule and the relocation remedy, not just the number).
- [ ] It measures Rust production lines as those before the first
      `^#\[cfg(test)\]`; shell/TS by total non-blank non-comment lines.
- [ ] It implements the ratchet: a committed baseline of currently-over-threshold
      files with their sizes; failure only when a listed file **grows** or a new
      file crosses the threshold.
- [ ] Vendored, installed-mirror, and bootstrap paths are exempt via an explicit,
      commented allowlist.
- [ ] Wired into CI alongside the existing `check-*.sh` family.
- [ ] Policy documented in `.loom/docs/file-size-policy.md` with a one-line
      pointer from `CLAUDE.md` (per the budget-check discipline — new subsystem
      detail goes to `.loom/docs/`, not inline).
- [ ] `scripts/check-claude-md-budget.sh` still passes (the pointer must fit).

## Deferred to follow-on issues (do NOT bundle here)

1. **Mechanical test extraction** — move `#[cfg(test)] mod tests` to sibling
   `foo/tests.rs`. Compiler-verified, no judgment required, roughly halves the
   worst files. One issue per file or small batch, never coupled to feature work.
2. **Semantic decomposition** of the ~37 files still over 1000 production lines
   after (1) — Architect-proposed, one file at a time, scheduled when that file
   has no open PRs touching it.
3. **Shell port tier 3** — finish `merge-pr.sh` / `worktree.sh` /
   `claude-wrapper.sh` onto native `loom-daemon` subcommands.

turian:
Closes #7711

## What

Adds `scripts/check-file-size-budget.sh` — a **one-way ratchet**, not a line limit.

A source file already over **1000 code lines** is frozen at its current size: it may shrink, it may not grow. Files under threshold are unconstrained. Nothing has to be refactored up front; the gate fires only at the moment a change makes a known-oversized file worse.

The debt ledger is `scripts/file-size-baseline.txt` — **36 files** today. Those numbers are debt, not targets, and should only ever go down.

## Why

Measured on `main` @ 38f85147:

- 86 of 228 Rust files exceeded 1000 lines and held **79% of all Rust LOC**
- The five worst files reached their **entire** current size within 90 days
- Add-to-delete ratio ≈ **20:1** — agents add, they essentially never restructure

```
net  +9430  (+10048/-618)   ipc.rs          112 commits in 90d
net  +9949  (+10378/-429)   role_runner.rs   47 commits
net  +9157  (+10432/-1275)  work_finder.rs   73 commits
```

Same failure mode `check-claude-md-budget.sh` (#4014) was built to counter: every individual addition is defensible while the aggregate is the problem.

## What this deliberately is NOT

**"You touched a large file, now refactor it"** — considered and rejected in #7711:

- **Fires constantly.** `ipc.rs` is touched >1×/day (112 commits/90d), `types.rs` 94. A tax on nearly every sweep, not occasional cleanup.
- **Couples unrelated refactors into feature PRs.** Mean touch is +160/−24; a mandated split makes that four figures, degrading Judge review.
- **Fights the concurrency model.** A module split moves `use` and `pub(crate)` across the file, conflicting with every in-flight PR touching it — worst when the fleet is busiest.

## Counting

Code lines only; blanks and comment-only lines never count. **Rust counts production lines only** (everything before the first `#[cfg(test)]`).

Inline test modules are idiomatic Rust and are **49% of this repo's Rust bulk** (142,674 of 289,955 lines). Taxing them would push tests out of the codebase for the wrong reason — growing a test module is always allowed. This is why ledger numbers look smaller than `wc -l`: `role_runner.rs` is 9,949 raw lines but ~893 production.

## Exemptions

| Class | Why |
|---|---|
| Vendored `guard-destructive*.sh` | Canonical copy is in rjwalters/repo, re-vendored at release; header forbids hand-editing. A local split is reverted by the next re-vendor. At 8,872 lines it is the largest shell file in the repo. |
| `.loom/` installed mirrors | Resync copies of `defaults/`; measured at source. Measuring both double-counts and churns the baseline on every resync commit. |

**Bootstrap scripts are deliberately NOT exempt.** `install.sh`, `install-loom.sh`, `loom-daemon-{update,start}.sh` can never be ported into `loom-daemon` (chicken-and-egg) — but "cannot be ported" is not "may grow without bound". This resolves an internal contradiction in the issue text, which listed them as exempt in the AC and as ratcheted in the tier table.

## Verification

`--self-test` runs in CI **ahead of** the real check — a ratchet that silently stops measuring reports OK forever while the files it guards keep growing.

```
ok   rust inline #[cfg(test)]/comments/blanks excluded
ok   installed-mirror and vendored paths not measured
ok   baseline records the over-threshold file
ok   baseline omits the under-threshold file
ok   clean tree passes
ok   growth of an over-threshold file fails
ok   shrinking passes
ok   new file crossing the threshold fails
ok   inline test-module growth is allowed
```

All existing gates pass: `check-claude-md-budget`, `check-docs-defaults-parity` (+ its self-test), `check-agents-md-sync`, `check-gitignore-convergence`.

## Reviewer notes

- **`CLAUDE.md` is exactly at its 320-line budget.** The pointer cost 1 line; I paid for it by reflowing one adjacent already-wrapped bullet (Branch rulesets) rather than raising the budget. That reflow is the only unrelated hunk in the diff.
- **`.loom/docs/file-size-policy.md` is repo-local** and added to `check-docs-defaults-parity.sh`'s `ORPHAN_ALLOWLIST`. Its gate lives in `scripts/`, not `defaults/scripts/`, so it never ships. Verified `resync_tree` only copies from `defaults/` and never prunes extras, so it will not be reaped.
- **Watch the baseline diff in future PRs.** A change that raises a number, or adds a row without a stated reason, is the ratchet slipping — that is the one thing this is designed to prevent.

## Follow-on (deferred, not in this PR)

1. Mechanical `#[cfg(test)]` extraction to sibling files — compiler-verified, roughly halves the worst files
2. Semantic decomposition of what remains over threshold — Architect-proposed, one file at a time
3. Shell port tier 3 — `merge-pr.sh` / `worktree.sh` / `claude-wrapper.sh` already have a `loom-daemon forge` delegation ladder


turian:
## Correction pushed — the counting rule had two bugs, both silent

Found while starting the follow-on test-extraction work. Worth re-reviewing the awk pass specifically.

**Bug 1 — undercounting (hid oversized files from the gate).** The first revision truncated each Rust file at its first `#[cfg(test)]`. That attribute also marks test-only helper *functions*, which sit mid-file with thousands of production lines below them:

| file | measured | actual | miss |
|---|---|---|---|
| `role_runner.rs` | 893 | **2208** | −1315 |
| `work_finder.rs` | 1003 | **1918** | −915 |

Both are among the largest files in the repo and **neither entered the baseline at all** — the gate was blind to exactly the files it exists to guard.

**Bug 2 — overcounting (would have failed honest changes).** A multi-line attribute between `#[cfg(test)]` and `mod tests`:

```rust
#[cfg(test)]
#[allow(
    clippy::unwrap_used,
)]
mod tests {
```

dropped the pending state on the first continuation line, so the entire test module counted *as production*. `sweep_registry/dispatch.rs` measured **4838** against a true **1599**.

**Fix.** Only top-level `#[cfg(test)] mod ... { ... }` blocks are skipped, tracked through multi-line attributes and terminated on a column-0 `}`. Everything inside a top-level module is indented, so this needs no brace counting and no string/comment escaping. A raw string containing a column-0 `}` would end the skip early — that fails toward counting *more* lines, the safe direction for a ratchet.

**Verification.** Counts now agree exactly with an independent reimplementation across all five largest files:

```
file                                         script   manual  match
loom-daemon/src/sweep_registry/dispatch.rs     1599     1599   OK
loom-daemon/src/role_runner.rs                 2208     2208   OK
loom-daemon/src/work_finder.rs                 1918     1918   OK
loom-daemon/src/ipc.rs                         3136     3136   OK
loom-daemon/src/safehouse.rs                   2440     2440   OK
```

Baseline goes **36 → 39** entries. Both shapes are now self-test fixtures.

This is a good argument for the `--self-test` job running ahead of the real check: both bugs reported a green gate.

turian:
## CI fix pushed — and the failure is worth reading

The **File Size Ratchet** job failed. The `--self-test` step passed, so the machinery was fine; the baseline had gone stale.

`tests/hooks/test-guard-destructive.sh` grew **4484 → 4501 on `main`**, after this branch's baseline snapshot and before this gate existed. CI evaluates the *merge result*, so this PR failed for growth it did not cause.

Rebased onto `main` and regenerated. The ledger diff is exactly one line:

```
-4484 tests/hooks/test-guard-destructive.sh
+4501 tests/hooks/test-guard-destructive.sh
```

**Flagging this explicitly because it is a baseline number going UP**, which this PR's own policy tells Judge to treat as the ratchet slipping. It is justified here — the growth happened on `main`, not in this change — but it should never pass review silently.

The doc now names the case, gives the remedy (`git rebase origin/main` + `--update`), and requires the PR description to state that a raised number came from `main`. Mostly a bootstrap artifact: once the gate is live, growth gets caught on the PR that causes it. The residual case is two PRs in flight, same remedy.

turian:
## The failing check is a flake, not this PR

`test-bounded-run.sh` failed here in the "Shell Test Suites (hermetic)" job, but this PR touches `ci.yml`, a file-size budget script and a baseline file — nothing in `bounded-run.sh` or anything it calls.

Confirmed unrelated: the same suite failed ~20 minutes later on **PR #7781** (Cargo workspace version inheritance), another branch with no connection to it. Two unrelated branches, same assertion.

The failing assertions are both in the **portable** (timeout-free) block:

```
FAIL: [portable] a never-returning command is bounded and reports 124 (got '124' want '127')
FAIL: [portable] the wedged child recorded its pid (fixture sanity)
```

The second is the tell — a *fixture sanity* check that the deliberately-wedged child recorded its pid. That failing means the child never got far enough to write the file, which is a spawn/read race on a loaded runner, not a behavior change.

Filed as **#7788** with the analysis and a suggested fix (bounded poll for the pid file instead of an immediate read).

## Verified locally on this branch

```
test-bounded-run.sh            18/18 passed
check-file-size-budget.sh      OK — no over-threshold file grew (threshold 1000)
check-dangling-links           PASS
check-docs-defaults-parity     PASS
check-claude-md-budget         PASS
check-agents-md-sync           PASS
check-hooks-defaults-parity    PASS
```

Also checked for review: the branch is **already current with `main`** (rebase was a no-op, 11 commits ahead, no conflicts), and it touches no `defaults/` files, so it needs no VERSION bump and will not go stale the way the `defaults/`-touching PRs have been.

A fresh CI run was already in flight when I looked, so I did not trigger another.

turian:

I authored this. Mechanical verification only; I cannot independently vouch for whether a ratchet is the right instrument.

**Green: 26 checks pass, zero failures**, including the `File Size Ratchet` job's own `--self-test` (11 checks) which runs ahead of the real gate.

### What it does

A file already over **1000 code lines** is frozen at its current size — may shrink, may not grow. Files under it are unconstrained. Nothing must be refactored up front; the gate fires only when a change makes a known-oversized file worse. Ledger: `scripts/file-size-baseline.txt`, **83 files / 160,495 code lines**.

Explicitly **not** "you touched a big file, now refactor it" — rejected on churn data (the hottest files are touched >1×/day, so that rule would tax nearly every sweep and collide across parallel worktrees).

### Baseline is post-refactor

Regenerated against main *after* #7723 and #7750 landed, so it reflects the real shape rather than needing a second refresh. Honest about what those achieved:

- entries **75 → 83** (extraction *creates* files; several are over threshold)
- total **164,260 → 160,495**, only **−2.3%** — moving lines between files does not delete them

The win is elsewhere and is large: the file an agent opens to edit production code dropped **~60%** ( 7,615→3,167;  6,643→2,435;  4,840→1,607).

### Two bugs found and fixed during review, both silent

1. **Undercounting** — truncating at the first `#[cfg(test)]` missed test-only *helper fns* with production code below them.  measured 893 vs a true 2,208, and **never entered the baseline** — the gate was blind to exactly the files it exists to guard.
2. **Overcounting** — a multi-line `#[allow(...)]` between `#[cfg(test)]` and `mod tests` dropped the pending state, counting a whole test module as production ( 4838 vs a true 1599).

Both reported a **green** gate, which is why `--self-test` runs first and why both shapes are now fixtures. The final rule counts whole files, which deletes that machinery entirely — prefer the rule you cannot implement incorrectly.

### Reviewer note

Watch the baseline diff in future PRs: a raised number or a grown total without a stated reason is the ratchet slipping. The one legitimate exception (growth landing on  while a PR is open) is documented with its remedy, and I hit it four times today.

turian:
## Correction to the review comment above — shell ate some of it

My previous comment was posted via an inline shell string, and backtick pairs in it were command-substituted away before they reached GitHub. Three spots lost their content. Restoring them for the record:

**The per-file improvement** (the actual win, since the ledger total barely moved):

| file | before | after | |
|---|---|---|---|
| `ipc.rs` | 7,615 | 3,167 | −58% |
| `role_runner.rs` | 6,643 | 2,435 | −63% |
| `work_finder.rs` | 5,436 | 1,922 | −64% |
| `safehouse.rs` | 6,426 | 2,443 | −61% |
| `sweep_registry/dispatch.rs` | 4,840 | 1,607 | −66% |

**Bug 1 (undercounting)**: `role_runner.rs` measured **893** against a true **2,208**, and so never entered the baseline at all.

**Bug 2 (overcounting)**: `sweep_registry/dispatch.rs` measured **4,838** against a true **1,599**.

**Reviewer note**: the one legitimate reason a baseline number may rise is growth landing on `main` while a PR is open — CI evaluates the merge result, so an open PR carrying the baseline fails for growth it did not cause. Remedy is `git rebase origin/main` plus `--update`, and the PR description must say the raised number came from `main`. I hit that four times while landing this.

Nothing about the merged content changed; only this comment was incomplete.

PR #7723, opened 2026-09-15
refactor(daemon): extract inline test modules from the eight largest Rust files (#7718)

### LINKED ISSUE #7718: Extract inline #[cfg(test)] modules to sibling files in the eight largest Rust files
Follow-on track 1 from #7711, deliberately deferred there so it would not be
bundled with the ratchet gate.

## What

Move each oversized Rust file's top-level `#[cfg(test)] mod tests { ... }` block
out to a sibling `foo/tests.rs`, leaving:

```rust
#[cfg(test)]
mod tests;
```

Purely mechanical. The compiler verifies it: `use super::*` resolves identically
from a child module file, and a byte-identical move cannot change behaviour.

## Why

Inline test modules are **49% of this repo's Rust bulk** — 142,674 of 289,955
lines. The policy in #7711 deliberately does *not* tax them (idiomatic Rust;
taxing them would push tests out of the codebase for the wrong reason), but the
files are still what an agent has to open, and the edit target is what matters
for context.

| file | total | production | test module |
|---|---|---|---|
| `ipc.rs` | 10,849 | 3,136 | 5,951 |
| `role_runner.rs` | 9,949 | 2,208 | 5,578 |
| `work_finder.rs` | 9,157 | 1,918 | 4,898 |
| `safehouse.rs` | 9,092 | 2,440 | 4,990 |
| `sweep_registry/dispatch.rs` | 8,343 | 1,599 | 4,741 |
| `claim_reconciliation.rs` | 7,070 | 1,774 | 3,602 |
| `health.rs` | 6,265 | — | 3,221 |
| `main_health_gate.rs` | 6,445 | 1,842 | 3,093 |

An agent editing `role_runner.rs` currently loads 9,949 lines to reach 2,208
lines of production code. After extraction it opens ~2,200 and greps the test
file when it needs a specific case.

## Scope

- **In scope:** the eight files above. Byte-identical moves only.
- **Out of scope:** renaming tests, splitting test modules by theme, fixing
  lints, touching production code. Any of those turn a compiler-verified move
  into a reviewable change and should be their own issue.
- Files with **multiple** top-level `#[cfg(test)]` items (`role_runner.rs` has
  4, `work_finder.rs` has 3) — extract only the `mod tests { ... }` block(s).
  Test-only helper **functions** stay where they are; the child module reaches
  them through `use super::*`.

## Acceptance criteria

- [ ] For each file: the test module lives in `<module>/tests.rs`, and the
      parent carries `#[cfg(test)] mod tests;`.
- [ ] The moved content is **byte-identical** modulo the outer `mod tests {` /
      `}` wrapper and one level of dedent. The PR must state how a reviewer can
      verify this mechanically.
- [ ] `cargo test` passes with the **same test count** as before the change —
      that number is the real acceptance signal; state it before and after.
- [ ] `cargo clippy` is no dirtier than before.
- [ ] `scripts/check-file-size-budget.sh --update` is run and the baseline diff
      shows only **decreases** (this is the case the ratchet is built to reward).
- [ ] One commit per file, so each move can be verified independently.

## Notes

- This work does **not** reduce the numbers the #7711 ratchet tracks — that gate
  counts production lines only, and those are unchanged by definition. The win
  is the size of the file an agent actually opens. Both matter; they are
  different measurements.
- Best done when the target files have no open PRs against them. `ipc.rs` is
  touched >1×/day, so it is the most likely to conflict and should probably go
  last or in its own PR.
- Depends on #7711 landing first, so the baseline regeneration lands on top of
  a committed baseline rather than creating one.

turian:
Closes #7718. Follow-on track 1 from #7711.

## What

Moves each file's top-level `#[cfg(test)] mod tests { ... }` to a sibling `<module>/tests.rs`, leaving `#[cfg(test)] mod tests;` behind.

| module | before | parent after | child | parent |
|---|---|---|---|---|
| `ipc` | 10,849 | 4,900 | 5,911 | −54% |
| `role_runner` | 9,949 | 4,373 | 5,439 | −56% |
| `work_finder` | 9,157 | 4,261 | 4,855 | −53% |
| `safehouse` | 9,092 | 4,104 | 4,918 | −54% |
| `sweep_registry/dispatch` | 8,343 | 3,609 | 4,713 | −56% |
| `claim_reconciliation` | 7,070 | 3,470 | 3,569 | −50% |
| `main_health_gate` | 6,445 | 3,354 | 3,076 | −47% |
| `health` | 6,265 | 3,046 | 3,210 | −51% |
| **total** | **67,170** | **31,117** | | |

An agent editing production code in these modules now opens **46%** of what it used to.

## How — and the bug this avoids

**The module body is moved verbatim. `cargo fmt` does the re-indentation.**

My first attempt hand-dedented by four spaces. That is a *lexical* edit applied to a *syntactic* structure: it cannot tell code from the interior of a raw string literal, so it silently rewrote embedded shell scripts and JSON fixtures in 4 of the 8 files. It was invisible in the diff, invisible to the compiler, and invisible to the test suite — the corrupted fixtures happened to be whitespace-insensitive, so all 5,123 tests still passed.

`rustfmt` has a real Rust lexer, knows exactly where literals begin and end, and is already enforced in CI. Verbatim move + `cargo fmt` makes correctness structural rather than something I have to get right by hand.

#7711 documents this as the standing rule for mechanical refactors.

## Verification

**Every raw-string literal is byte-identical before and after**, across all eight files — checked by extracting literals with a scanner that tracks comments and normal strings, then comparing:

```
IDENTICAL role_runner.rs           59 raw-string literal(s) unchanged
IDENTICAL work_finder.rs           24
IDENTICAL ipc.rs                   17
IDENTICAL safehouse.rs              6
IDENTICAL claim_reconciliation.rs  21
IDENTICAL health.rs                 0
IDENTICAL main_health_gate.rs      53
IDENTICAL sweep_registry/dispatch.rs 9
```

This is the acceptance test that actually catches the failure mode — reading the diff does not.

Also:
- `cargo fmt --check` clean
- Test list identical: **5,124** tests before and after
- Full suite: **5,123 passed, 1 failed** — `init::repo_owned::tests::rust_and_shell_resync_ignore_parsers_agree_on_shared_fixture`, which is **pre-existing**. Confirmed by running it on a branch with zero Rust changes, where it fails identically (`shell_is_ignored driver failed unexpectedly: Some(2)`). That file is not touched by this PR.

## Scope

Byte-preserving moves only. No test renames, no splitting test modules by theme, no lint fixes, no production changes. One commit per module so each move can be verified independently.

`role_runner.rs` and `work_finder.rs` also contain test-only helper **functions** outside the test module; those stay where they are and the child reaches them via `use super::*`.

## Interaction with #7711

Under #7711's whole-file counting, the parents drop well down the ledger and the extracted files enter it on their own terms. Several extracted files are themselves over threshold — that is the correct signal, not a false positive: a 5,911-line `ipc/tests.rs` is still a big file. Splitting those by theme is follow-on work, and needs judgment rather than a mechanical pass.

**Merge after #7711**, then run `scripts/check-file-size-budget.sh --update` so the ledger reflects the new shape.


turian:

I authored this PR. Mechanical verification below; I cannot independently vouch for the judgment calls.

**Verified:**
- **Every raw-string literal byte-identical before and after**, across all eight files, checked with a scanner that tracks comments and normal strings — 189 literals total. This is the acceptance test that catches the actual failure mode; reading a 36k-line diff does not.
- `cargo fmt --check` clean.
- Test list identical: **5,124** before and after.
- Full suite: **5,123 passed, 1 failed** — `init::repo_owned::tests::rust_and_shell_resync_ignore_parsers_agree_on_shared_fixture`, which is **pre-existing and macOS-only** (#7730: the embedded driver uses bash-4 `declare -A`; stock macOS bash 3.2 rejects it, driver exits 2). Confirmed by running it on a branch with zero Rust changes, where it fails identically. CI (bash 5) is green.
- All checks green.

**The method matters more than the diff.** The module bodies are moved **verbatim**; `cargo fmt` does the re-indentation. My first attempt hand-dedented four spaces and silently rewrote embedded shell and JSON fixtures inside raw strings in 4 of 8 files — invisible in the diff, to the compiler, and to all 5,123 passing tests, because the corrupted fixtures happened to be whitespace-insensitive. `rustfmt` has a real Rust lexer; I do not. #7711 records this as the standing rule for mechanical refactors.

**Effect:** parents drop 67,170 → 31,117 lines. An agent editing production code in these modules now opens **46%** of what it used to.

**Note for whoever lands #7714 next:** under that PR's whole-file counting, these extracted `tests.rs` files enter the ledger on their own terms and several are over threshold — correct signal, not a false positive; a 5,911-line `ipc/tests.rs` is still a big file. #7714's baseline needs regenerating against post-#7750/#7723 main, which I am doing next.

Issue #7721, opened 2026-09-15
classify-dependency-block.sh fails bash 3.2 parse, blocking resync-installed.sh on all macOS hosts


turian:
`defaults/scripts/classify-dependency-block.sh` does not parse under macOS's stock bash, which makes `check-shell-syntax.sh` fail, which makes **`resync-installed.sh` refuse to resync anything** on any macOS host.

## Symptom

```console
$ ./.loom/scripts/resync-installed.sh --dry-run
ERROR: Refusing to resync: one or more source shell scripts do not parse (bash -n).
ERROR: check-shell-syntax.sh: defaults/scripts/classify-dependency-block.sh does not parse (bash -n):
  line 938: syntax error near unexpected token `('
  line 938: `        err "--issue must be a number (got: $ISSUE)"'
check-shell-syntax.sh: 1 of 373 script(s) FAILED to parse
ERROR: Fix the offending file(s) under defaults/ before re-running this script -- nothing was copied.
```

Line 938 is innocent. The reported location is ~150 lines downstream of the actual defect.

## Root cause

**bash 3.2 does not skip heredoc bodies when scanning a `$( ... )` for its closing paren.** An apostrophe in heredoc *prose* opens a single-quote state that never closes, and the parse dies much later at the next unquoted metacharacter.

Minimal repro (heredoc delimiter shown as `XEOF` to keep this comment readable):

```bash
f() {
  local x
  x="$(cat <<XEOF
Champion's escalation
XEOF
)"
}
```

```console
$ /bin/bash -n repro.sh
repro.sh: line 5: unexpected EOF while looking for matching `''
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
```

Remove the apostrophe and it parses. Bash 4+ handles it correctly, so this is invisible everywhere except stock macOS — where `/bin/bash` is 3.2 and `#!/usr/bin/env bash` resolves straight to it.

## The three sites

All in `_apply_fact_unescalation()`, all the possessive "Champion's", inside `"$(cat <<EOF ... )"` blocks:

```
789  Curator re-verified every objection Champion's escalation cited against
824  Every objection Champion's escalation cited has since been independently
832  sub-kind label, if present). This proposal returns to Champion's normal
```

[...]

PR #7712, opened 2026-09-15
docs: replace nonexistent commands in onboarding guides

turian:
Closes #7710

## Problem

The onboarding path instructed users to run commands that **do not exist**. A
first-time installer following `docs/guides/getting-started.md` hit a hard
failure at the label-sync step — the one step that makes label coordination work
at all.

Each replacement below was verified against the tree, not assumed.

| Documented | Reality | Now |
|---|---|---|
| `gh label sync` | `gh label` has no `sync` subcommand (only `clone`/`create`/`delete`/`edit`/`list`) | `.loom/scripts/sync-labels.sh` |
| `loom-daemon start` / `stop` | Not top-level subcommands — `Start`/`Stop` exist only under `accounts session` (`loom-daemon/src/main.rs` `SessionAction`) | `.loom/scripts/cli/loom-daemon-start.sh` / `-stop.sh` |
| `loom:ready` | Absent from `.github/labels.yml` | `loom:curated` → `loom:issue` |
| `claude --role builder` | No such flag | `claude`, then `/loom:builder` |
| `claude code "/loom:judge"` | Not the invocation | `claude`, then `/loom:judge` |
| `open -a Loom --args --workspace` | No GUI ships in this repo; the `Option 4: GUI Application` TOC entry pointed at a section that did not exist | removed |
| `pnpm worktree <n>` (told to consumers) | Only defined in Loom's own `package.json` | `./.loom/scripts/worktree.sh` |

`gh label sync` was the widest: besides the guides it appeared in `CLAUDE.md`,
**`defaults/.loom/CLAUDE.md`** — the template every consumer repo inherits, so
each new install reintroduced it — `docs/agents.md`, `defaults/docs/troubleshooting.md`,
and all three `quickstarts/*/README.md`.

Two files already contradicted themselves. `defaults/docs/troubleshooting.md`
named the correct script two lines below the wrong command. And
`quickstart-tutorial.md` noted that Loom uses `gh pr comment` *instead of*
`gh pr review --approve`, then used `gh pr review --request-changes` twelve
lines later — and moved a **PR** to `loom:building`, an issue-lane label. It now
uses `gh pr comment` + `loom:changes-requested` throughout.

## Added: the missing first-run path

`getting-started.md` never mentioned `/loom:sweep`, the headline Tier-1 entry
point. It now documents, in order:

1. **Label creation is required and a Quick Install does not do it.** `--quick`
   ships `.github/labels.yml` but never creates the labels on the forge
[...]

Issue #7710, opened 2026-09-15
docs: onboarding guides instruct commands that do not exist


turian:
## Problem

The onboarding docs instruct new users to run commands that **do not exist**. A
first-time installer following `docs/guides/getting-started.md` hits a hard
failure at the label-sync step — the single step that makes the whole
label-coordination workflow function.

Verified against the current tree:

| Documented | Reality |
|---|---|
| `gh label sync -f .github/labels.yml` | `gh label` has no `sync` subcommand (only `clone`/`create`/`delete`/`edit`/`list`). Real path: `.loom/scripts/sync-labels.sh` |
| `loom-daemon start` / `loom-daemon stop` | Not top-level subcommands. `Start`/`Stop` exist only under `accounts session` (`loom-daemon/src/main.rs:2344` `SessionAction`). Real path: `.loom/scripts/cli/loom-daemon-start.sh` |
| `loom:ready` | Not present in `.github/labels.yml`. Real states: `loom:curated` → `loom:issue` |
| `claude --role builder` | No such flag |
| `claude code "/loom:judge"` | Not the invocation — run `claude`, then `/loom:judge` |
| `open -a Loom --args --workspace` | No GUI application ships in this repo |
| `pnpm worktree <n>` (instructed to consumers) | Only defined in Loom's own `package.json`; a consumer repo must use `./.loom/scripts/worktree.sh` |

Affected files: `docs/guides/getting-started.md`, `docs/guides/quickstart-tutorial.md`,
`docs/guides/common-tasks.md`, `docs/guides/cli-reference.md`, `docs/workflows.md`,
and `CLAUDE.md:288` — meaning every agent in this repo is also handed the
nonexistent `gh label sync`.

## Secondary defects

- `getting-started.md` never mentions `/loom:sweep`, the headline Tier-1 entry point.
- `--quick` installs ship `.github/labels.yml` but do **not** create the labels on
  the forge (that is a Full Install step, `install.sh:654`). Nothing tells the
  operator this, so a `--quick` install silently has no coordination substrate.
- No mention of `setup-branch-protection.sh`; its 0-approval ruleset is what
  permits Champion auto-merge.
- No pointer to the token pool before long/overnight runs (an exhausted pool
  exits `78` and stalls the pipeline).
- Role tree lists 8 roles, omitting Champion and Auditor (10 exist).
- Claims macOS-only despite `defaults/scripts/lib/systemd-user.sh`.
[...]

PR #7729, opened 2026-09-15
fix: replace nonexistent 'gh label sync' in loom.md role prompt and the guard-hooks pair

### LINKED ISSUE #7720: Remove nonexistent 'gh label sync' from loom.md role prompt and the guard-hooks .md/.sh pair
Follow-up to #7710. Its PR (#7712) removed the nonexistent `gh label sync` across the onboarding surface but explicitly deferred two sites that fall outside the operator-session lane. Neither has an issue yet; this is it.

## 1. Role prompt

```
defaults/.claude/commands/loom/loom.md:684    gh label sync --file .github/labels.yml
```

Role prompts are excluded from the operator-session lane unconditionally, so a docs-only change could not touch it. This is the highest-impact remaining instance: it is an instruction handed directly to an agent, so an agent following it runs a command that does not exist (`gh label` has only `clone`/`create`/`delete`/`edit`/`list`) and fails at exactly the moment it is trying to repair label state.

Replacement: `.loom/scripts/sync-labels.sh` (installed) — matching what #7712 used everywhere else.

## 2. The `guard-hooks` `.md` / `.sh` pair

```
defaults/docs/guard-hooks.md:351              ... or one `gh label sync` in a repo with `labels.yml`
defaults/hooks/guard-destructive-generic.sh:853  # label (a repo with labels.yml restores in one `gh label sync`). A guard whose
```

These must move together — fixing only the `.md` would desync the pair, and the `.sh` is out of lane.

**This one is not purely cosmetic.** The claim is load-bearing: it is the stated justification for `gh label delete` being classified *reversible* and therefore not prompting by default. The guard's own comment reasons "a repo with labels.yml restores in one `gh label sync`" — but that command does not exist, so the recovery path the non-prompting decision rests on is not real.

The decision is probably still correct — `.loom/scripts/sync-labels.sh` genuinely does restore a deleted label from `labels.yml`, and it was verified doing exactly that in a clean room (34/34 labels created on a repo that had none). So the fix is to name the command that actually performs the recovery, in both files, rather than to change the guard's tier. Worth a moment's confirmation that reversibility is being judged against a real path and not an assumed one.

## Acceptance

```bash
grep -rn 'gh label sync' defaults/ .loom/ docs/ CLAUDE.md    # no hits outside docs/design/
```

`docs/design/issue-332-label-state-machine.md` and `docs/design/issue-179-prompt-library.md` are historical records of the migration that removed these and should stay untouched, exactly as #7712 left them.

Note `.loom/docs/guard-hooks.md` is a symlink into `defaults/docs/`, so the single edit covers both copies.

## Related

- #7715 — `examples/**/.loom/config.json` still seed agents with the removed `loom:ready` label
- #7717 — `sync-labels.sh --check` crashes on macOS bash 3.2 (the replacement command's own drift-check mode)

turian:
Closes #7720

Follow-up to #7710, which removed `gh label sync` across the docs surface but deferred three sites outside the operator-session lane.

`gh label` has no `sync` subcommand — only `clone`/`create`/`delete`/`edit`/`list`. The replacement is the shipped `.loom/scripts/sync-labels.sh`, matching what #7712 used everywhere else.

## Sites

**1. `defaults/.claude/commands/loom/loom.md`** — the highest-impact instance, because it is an instruction handed straight to an agent. An agent following it runs a command that does not exist, and fails at exactly the moment it is trying to repair label state.

**2. `defaults/docs/guard-hooks.md` + `defaults/hooks/guard-destructive-generic.sh`** — the same claim in both, and **load-bearing rather than cosmetic**. It is the stated justification for classifying `gh label delete` as *reversible* and therefore not prompting by default:

> `gh label delete` (undo: recreate, or one `gh label sync` in a repo with `labels.yml`)

The recovery path that non-prompting decision rests on did not exist. The decision itself still holds — `.loom/scripts/sync-labels.sh` genuinely does restore a deleted label from `labels.yml`, verified in a clean-room install (34/34 labels created on a repo that had none). **Only the named command changes; the guard's tier is untouched.**

These two must move together or the pair desyncs, which is why they could not ride along with a docs-only change.

## Installed copies

`.loom/docs/guard-hooks.md` and `.loom/hooks/guard-destructive-generic.sh` are **tracked** in this repo and updated in the same commit — `check-hooks-defaults-parity` fails otherwise (#7416), since a `defaults/` fix that never reaches `.loom/` leaves the copy that actually executes stale. Normally `resync-installed.sh` handles this; it is currently blocked repo-wide on macOS by #7721 (PR #7722).

## Verification

```
check-hooks-defaults-parity:  PASS
check-docs-defaults-parity:   PASS
check-dangling-links:         PASS
check-agents-md-sync:         PASS
check-claude-md-budget:       PASS
```

Both `.sh` copies parse under bash 3.2, and `defaults/` vs `.loom/` are byte-identical for both files.

Historical references in `docs/design/issue-332-label-state-machine.md` and `docs/design/issue-179-prompt-library.md` are intentional records of the migration that removed these, and are left untouched.

## Related

Remaining `gh label sync` hits on `main` are #7712's scope (this branch is cut from `main`, which does not yet carry that PR).


turian:
## Judge: approved

Rebased onto main (was 7 behind, conflicting on the VERSION-bearing files). Cherry-picked the functional commit and re-bumped cleanly rather than resolving a stale bump — functional diff unchanged.

Verified the premise: `gh label sync` does not exist. `gh label` provides `create`, `clone`, `delete`, `edit`, `list` — no `sync`. So every one of these four references pointed operators and agents at a command that fails. `.loom/scripts/sync-labels.sh` is the real path, and #7733 (just approved) makes its `--check` mode work on macOS bash 3.2, so the recommendation now actually runs where a developer will try it.

### One hunk deserves a note

It edits `defaults/hooks/guard-destructive-generic.sh`, which is a **vendored copy** whose header says "DO NOT hand-edit generic pattern behavior here — send fixes upstream."

I am approving it anyway, for a specific reason: the change is a **prose comment**, not pattern behaviour, and the header's prohibition is scoped to behaviour. More to the point, `scripts/check-guard-destructive-drift.sh` (#7683, merged) documents that this convention is already broadly unobserved — the vendored copy **drifted ~2,200+ lines ahead of upstream, 18 generic functions added locally and never ported back**, before anyone noticed. The repo's response was an advisory drift-visibility check, not enforcement. A one-line comment fix is noise against that, and the drift check is non-blocking by design.

It will still be reverted by the next re-vendor, so the same comment should be corrected upstream in [rjwalters/repo](https://github.com/rjwalters/repo) to make it stick.

Labeling `loom:pr` — merging once CI reports green on the rebased head.

turian:
## Judge: re-approving the CURRENT head

The earlier approval applied to `6892cdbf`. I then force-pushed a rebase onto main (the branch was 7 behind, conflicting on the VERSION-bearing files), which correctly invalidated it — `merge-pr.sh` blocked the merge with *"no forge-visible signal exists that Judge reviewed the CURRENT head."* That guard is right, and it caught a real gap rather than a bookkeeping one.

Re-verified on the current head:

- Functional diff is the four `gh label sync` → `.loom/scripts/sync-labels.sh` corrections, unchanged from what was reviewed. `gh label` has `create`/`clone`/`delete`/`edit`/`list` and no `sync`, so all four pointed at a command that fails.
- Carries its own VERSION bump (0 behind main, 4 ahead) so the `defaults/` currency signal is correct — not inherited from an unrelated merge.
- Green: SUCC27, SKIP1.
- The recommendation it points readers at now actually works on macOS, since #7733 fixes `sync-labels.sh --check` under bash 3.2.

The one hunk I flagged before still stands as a note, not a blocker: it edits a prose comment in the vendored `guard-destructive-generic.sh`, which will be reverted by the next re-vendor and should be corrected upstream in [rjwalters/repo](https://github.com/rjwalters/repo) to stick. Approving because the header's prohibition is scoped to pattern *behaviour*, and #7683's drift check documents that the vendored copy has already drifted ~2,200+ lines ahead of upstream — a one-line comment is noise against that, and the drift check is advisory by design.

Issue #332, opened 2025-10-18
reconsider issue label state machine


rjwalters:
I think we might want to have the human in the loop explicitly apply loom:issue and have workers pick that up

architect makes loom:architect-suggestion
critic makes loom:critic-suggestion
(external users) make "external" labelled issues based on contributing guidelines and workflows
Curator can process any issue and mark it as "loom:curated"

repo members mark issues for work by adding loom:issue

we can then drop loom:ready

we should keep loom:urgent and allow the Triage agent to set it

PR #7731, opened 2026-09-15
fix: example configs seed agents with labels that cannot exist

### LINKED ISSUE #7715: examples/*/.loom/config.json still seed agents with the removed loom:ready label
Follow-up to #7710 / #7712.

The onboarding cleanup replaced `loom:ready` (removed from `.github/labels.yml` by the #332 label-state-machine migration) everywhere it appeared in documentation. Four occurrences survive in **example workspace configs**, where they are not prose but the actual `intervalPrompt` an example agent is seeded with:

```
examples/quickstart/.loom/config.json:31   "intervalPrompt": "Find and implement loom:ready issues"
examples/full-stack/.loom/config.json:63   "intervalPrompt": "Find and implement loom:ready issues"
examples/full-stack/.loom/config.json:79   "intervalPrompt": "Find and implement loom:ready issues"
examples/full-stack/.loom/config.json:127  "intervalPrompt": "Find and implement loom:ready issues"
```

## Why this was deferred rather than fixed in #7712

`.loom/config.json` is excluded from the operator-session lane **unconditionally** (CLAUDE.md → "Operator-session lane"), alongside `.sh`, `.rs`, `.ts`, role prompts and `.github/labels.yml`. #7712 was a docs-only change that stayed inside that lane, so these four could not ride along. Same reasoning the PR used to defer `defaults/.claude/commands/loom/loom.md` (a role prompt) and the `guard-hooks.md` + `guard-destructive-generic.sh` pair.

## Impact

Worse than a stale doc reference. An operator who copies an example workspace gets an agent whose recurring prompt instructs it to find issues carrying a label that cannot exist on a current install. The agent finds nothing and idles, with no error explaining why.

## Suggested fix

Replace `loom:ready` with `loom:issue` — the current "approved for work, ready for Builder" state (`.github/labels.yml:29`, `description: "Approved for work. Applied by: humans, Champion, or sweep Approval gate"`).

Worth confirming while in there whether `intervalPrompt` is still the right key for these example terminals, and whether these examples should instead point at `/loom:sweep`, given #7712 established it as the documented first-run entry point.

## Acceptance

```bash
grep -rn 'loom:ready' examples/    # no hits
```

Plus: every `examples/**/.loom/config.json` remains valid JSON, and its labels all resolve against `.github/labels.yml`.

## Out of scope

The two remaining deferrals named in #7712 — the role prompt and the `guard-hooks` `.md`/`.sh` pair — are separate and still open.

turian:
Closes #7715

Follow-up to #7710. These are workspace templates an operator **copies**, so a stale label here is worse than a stale doc reference: the agent's recurring `intervalPrompt` targets a label no install can have, so it finds nothing and idles with no error explaining why.

## Two labels, five sites

| Was | Now | Sites |
|---|---|---|
| `loom:ready` | `loom:issue` | 4 |
| `loom:critic-suggestion` | `loom:hermit` | 1 |

`loom:ready` was removed by the #332 label-state-machine migration. `loom:issue` is the current "approved for work, ready for Builder" state — *"Approved for work. Applied by: humans, Champion, or sweep Approval gate."*

**`loom:critic-suggestion` was never declared in `labels.yml` at all** — found while checking that the rest of each config resolved, not from the issue. The terminal that files it is named "Critic" and is prompted to *"analyze the codebase for opportunities to simplify, remove bloat, or eliminate unnecessary complexity"* — Hermit's charter verbatim — so mapping it to `loom:hermit` is a reading of intent, not a guess.

`intervalPrompt` itself is still the correct key and is left alone: it is consumed by `mcp-loom/src/types.ts:54` and `mcp-loom/src/tools/terminals.ts`.

## Acceptance

Every label named in `examples/**/.loom/config.json` now resolves against `.github/labels.yml`:

```
loom:changes-requested  declared
loom:hermit             declared
loom:issue              declared
loom:review-requested   declared
```

Both files remain valid JSON. All five repo checks pass.

## Deliberately not fixed here

**Five of the seven `roleFile` values in these same configs name files that do not exist**: `worker.md`, `reviewer.md`, `fixer.md`, `critic.md`, `default.md`. Only `architect.md` and `curator.md` resolve. These examples predate the role rename, so they are broken well past their labels — and unlike the labels, the fix needs a decision (what `critic` and `default` become, or whether these templates should be regenerated wholesale) rather than a rename. Filed separately so this stays reviewable.

The remaining `loom:ready` hits in `examples/quickstart/README.md` are #7712's scope; this branch is cut from `main`, which does not carry that PR yet. The issue's `grep -rn 'loom:ready' examples/` acceptance goes fully clean once both land.


turian:
## Judge: approved

Verified against the live label set rather than assuming:

| label | exists? |
|---|---|
| `loom:ready` | **no** |
| `loom:critic-suggestion` | **no** |
| `loom:issue` | yes |
| `loom:hermit` | yes |

So the example configs were seeding worker prompts with labels that can never match, which means an agent booted from `examples/full-stack` would search for work it can never find and idle silently. Correct fix, and the replacements are the right ones per `.github/labels.yml`.

### One residual this PR does not cover

`docs/workflows.md:89` still says:

> **Curator**: Only marks `loom:ready` if all dependencies checked

That is a **live** doc carrying the dead label, so the same defect survives the fix. Not a reason to hold this PR — the change here is correct and self-contained.

The other remaining hits are in `docs/design/issue-332-label-state-machine.md`, which is the design record for the migration *away* from `loom:ready` ("replaced by `loom:issue`", "❌ `loom:ready`"). Those are historical by design and should stay.

Labeling `loom:pr`.

Issue #7728, opened 2026-09-15
CI never runs check-shell-syntax.sh, and no runner is bash 3.2 — two macOS-only defects reached main today


turian:
Two bash-3.2 defects reached `main` today (#7717, #7721). Neither was catchable by CI, for two independent reasons — one of which is broader than the bash-version problem.

## 1. `check-shell-syntax.sh` never runs in CI

It is not invoked by any workflow. The only match in `.github/workflows/` is a **comment**:

```
ci.yml:251   # unmerged INDEX entry, which a commit clears; check-shell-syntax.sh
ci.yml:252   # (#6162) only covers `*.sh` via `bash -n`. This is the content-level
```

Its only production caller is `resync-installed.sh`. CI does run `bash -n`, but on exactly two hand-listed files:

```
ci.yml:482   bash -n defaults/scripts/spawn-codex.sh
ci.yml:483   bash -n defaults/scripts/tests/test-spawn-codex.sh
```

So 371 of 373 shell scripts get no parse check on a PR at all. A script that does not parse can merge cleanly and is only discovered later, by whoever next runs a resync.

## 2. Every runner is bash 5

```
$ grep -c macos .github/workflows/ci.yml
0
```

Everything is `ubuntu-latest` or `self-hosted`. Stock macOS `/bin/bash` is **3.2.57** — Apple froze it for GPLv3 reasons — and `#!/usr/bin/env bash` resolves straight to it on an unmodified Mac. So the interpreter a large share of operators actually run is never exercised.

## What got through

| | Construct | Effect on macOS |
|---|---|---|
| #7717 | `local -A`, `${x,,}` in `sync-labels.sh` | `--check` crashes; the forge-agnostic label-drift check is unusable |
| #7721 | apostrophe in a heredoc inside `$( ... )` in `classify-dependency-block.sh` | file fails to **parse**; `check-shell-syntax.sh` then fails, and `resync-installed.sh` refuses wholesale — no macOS host can refresh `.loom/` or `.claude/commands/loom/` |

Both parse and run fine on bash 5, so both were invisible at every gate. #7721 is the sharper lesson: a single file's parse error took down resync repo-wide, and it was found by hand, not by a check.

## Proposed fix

Add a shell-syntax job that runs the existing script — no new checker to maintain:

```yaml
  shell-syntax:
    name: Shell Syntax (${{ matrix.os }})
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
[...]

PR #7733, opened 2026-09-15
fix: make sync-labels.sh --check work on macOS stock bash 3.2

### LINKED ISSUE #7717: sync-labels.sh --check crashes on macOS stock bash 3.2 (local -A / ${x,,} are bash 4+)
`sync-labels.sh --check` crashes on macOS's stock bash 3.2. Plain sync, `--dry-run` and `--prune-defaults` are unaffected — the failure is confined to the `--check` path.

## Repro

Clean-room Quick Install into a fresh GitHub repo, then:

```console
$ ./.loom/scripts/sync-labels.sh --check
ℹ Target repository: turian/loom-cleanroom-7712 (github)
ℹ Syncing workflow labels from .github/labels.yml...
./.loom/scripts/sync-labels.sh: line 534: local: -A: invalid option
local: usage: local name[=value] ...
./.loom/scripts/sync-labels.sh: line 538: accessibility: unbound variable
```

Host: `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)`. The shebang is `#!/usr/bin/env bash`, and on a stock macOS `bash` resolves to `/bin/bash` — Apple froze it at 3.2 (GPLv3), so this is the default for every macOS operator without a Homebrew bash.

## Cause

Two bash-4-only constructs in `diff_declared_against_live_tsv()` (`defaults/scripts/sync-labels.sh`):

```
534:  local -A LIVE_COLOR=() LIVE_DESC=() LIVE_SEEN=()    # associative arrays: bash 4.0+
539:    LIVE_COLOR["$name"]="${color,,}"                   # ,, lowercase expansion: bash 4.0+
```

`local -A` is rejected outright; with `set -u` the subsequent subscripted assignment then dies on the first label name read from the live TSV (`accessibility`, a GitHub default label).

## Scope — both forges

`diff_declared_against_live_tsv` has exactly two callers, `github_check_labels()` (line 580) and `gitea_check_labels()` (line 722). Both are reached only under `CHECK_MODE=1` (line 733), so:

- **Broken**: `--check`, on GitHub *and* Gitea.
- **Unaffected**: plain sync, `--dry-run`, `--prune-defaults`. Verified in the same clean room — a plain `.loom/scripts/sync-labels.sh` created all 34 declared labels correctly (`✓ Synced 34 labels`, matching `grep -c '^- name:' .github/labels.yml`).

The script is otherwise bash-3.2-clean; these are the only two such constructs in the file.

## Why it matters

`--check` is the only forge-agnostic way to verify that the *expected* label set is present rather than that *some* `loom:` label exists. That makes it the natural verification step in onboarding — #7712 nearly documented it as such, and reverted to a count comparison once this surfaced. Its exit codes (0 in sync, 3 drift, 1 forge error) also make it the obvious CI drift gate, which currently cannot run on a macOS runner.

## Suggested fix

Rewrite `diff_declared_against_live_tsv()` without associative arrays — the existing sync path already walks the same TSV bash-3.2-cleanly and can be followed. Replace `${color,,}` with `tr '[:upper:]' '[:lower:]'` (or `awk`). Alternatively, hard-require bash ≥ 4 with an explicit, actionable error instead of a raw interpreter crash — but that strands stock macOS, which is a documented supported platform.

Worth a repo-wide sweep for `local -A` / `declare -A` / `${x,,}` / `${x^^}` / `mapfile` / `readarray` in other `defaults/scripts/*.sh` while fixing, since this class of bug is invisible on Linux CI.

## Acceptance

```bash
/bin/bash ./.loom/scripts/sync-labels.sh --check    # exits 0 (in sync) or 3 (drift), never a syntax error
```

Verified against stock `/bin/bash` 3.2, on a repo that is in sync and on one with a deliberately deleted label.

turian:
Closes #7717

`--check` — the only forge-agnostic way to verify the **expected** label set is present, rather than that *some* `loom:` label exists — crashed outright on stock macOS:

```
sync-labels.sh: line 534: local: -A: invalid option
sync-labels.sh: line 538: accessibility: unbound variable
```

Two distinct bash-3.2 incompatibilities, both confined to the `--check` path. Plain sync, `--dry-run` and `--prune-defaults` were never affected.

## 1. `diff_declared_against_live_tsv()`

Used `local -A` and `${x,,}` — both bash 4.0+. Replaced with parallel indexed arrays plus a linear search, mirroring the `DECL_NAMES`/`DECL_COLORS`/`DECL_DESCS` + `is_declared_label()` pattern already in this file, and a `_lower()` helper using the same `tr '[:upper:]' '[:lower:]'` idiom used elsewhere in it. A live label set is dozens of entries, so the O(n·m) scan costs nothing next to the forge call that produced it.

## 2. A second bug underneath it

Fixing (1) exposed one that had been unreachable: `print_label_check_report()` expands three arrays that are routinely empty, and **under `set -u`, bash 3.2 treats an empty `"${arr[@]}"` as an unbound variable** — fixed upstream only in bash 4.4. All three now use `${arr[@]+"${arr[@]}"}`. `${#arr[@]}` is safe on 3.2 and is left alone.

This is worth noting for #7728: the first crash masked the second, so a single fix would not have made `--check` work. Only running it on 3.2 finds them.

## The file already declared the constraint it broke

Forty lines above the defect:

> Plain globals rather than local-by-reference (bash namerefs, `local -n`, need bash 4.3+ …) — **this repo targets bash 3.2+ for macOS compatibility per `.shellcheckrc`**

and `.shellcheckrc` says *"Target bash for macOS compatibility (bash 3.2+)"*.

## Verification — real forge, real bash 3.2

Against a disposable GitHub repo under `/bin/bash` 3.2.57 (arm64-apple-darwin25):

| State | Exit | Output |
|---|---|---|
| in sync | **0** | `34 declared label(s), all in sync` |
| 1 missing | **3** | `MISSING loom:hermit` |
| 1 missing + 1 stale + 1 unknown extra | **3** | all three reported |
| restored | **0** | back in sync |

The stale case confirms the `_lower()` replacement — a label edited to `FF0000` reports as `live=ff0000 vs declared=3b82f6`.

`test-install-sync-labels.sh`: **24/24 pass**. All five repo checks pass.

## Behavior change worth knowing

Unknown extras are now listed in **forge order**. The version this replaces iterated `"${!LIVE_SEEN[@]}"` — bash hash order — so that output was previously arbitrary and is now deterministic.

## Follow-up, not doable here

`docs/guides/getting-started.md` carries a caveat naming this issue and telling readers to use a count comparison until it is fixed. That paragraph lives on **#7712's branch, not on `main`**, so it cannot be updated from this branch. Once both land, it should be replaced with the `--check` recommendation it was standing in for — otherwise the guide ships pointing at a bug that no longer exists.


turian:
## Judge: approved

Rebased onto main (the branch was 7 commits behind and conflicting on the VERSION-bearing files). Verified on **stock macOS `/bin/bash 3.2.57`** — the exact interpreter this fixes — rather than taking the claim on trust:

```
$ /bin/bash -n defaults/scripts/sync-labels.sh     # parses
$ /bin/bash defaults/scripts/sync-labels.sh --check # exit 0, useful output
  UNKNOWN EXTRA loom:in-progress (present on rjwalters/loom, not declared in labels.yml ...)
  UNKNOWN EXTRA loom:healing ...
```

So `--check` now works where it previously crashed, and it immediately surfaced real drift — six labels live on the repo that `labels.yml` does not declare. That is a finding this fix unlocks, not a problem with the fix.

### Rebase note

The original branch carried its own `chore: bump version to 0.19.49` commit, which conflicted with every VERSION bump that landed since. Rather than resolve that conflict I cherry-picked **only** the `sync-labels.sh` change onto current main and re-bumped cleanly, so the diff is now the fix plus one fresh version bump. The functional change is byte-identical to what was reviewed before.

### Context worth recording

This PR's head lives on `turian/loom` (a cross-repo PR from before push access). That is why it appeared to have no branch on `origin` — and it exposed a real `worktree.sh` footgun in the process: `worktree.sh 7717` silently creates a fresh main-HEAD branch named `feature/issue-7717`, reports success, and tells you to push it to `origin`, because it resolves branches against `origin` only and never asks the forge whether an open PR exists. Filed as #7765. It does not clobber this PR, but it does let an operator believe they rebased work they did not touch.

Labeling `loom:pr` — will merge once CI reports green on the rebased head.

turian:
## Judge: re-approving the rebased head

Rebased again (main had moved; the branch hit the full 9-file VERSION conflict). Dropped the stale bump and re-bumped fresh rather than hand-resolving nine derived files — the technique recorded on #7743. The functional diff is unchanged: **1 file, +47/−13** in `defaults/scripts/sync-labels.sh`.

Re-verified on **stock macOS `/bin/bash` 3.2.57**, the interpreter this fixes:

- `bash -n` parses.
- `--check` runs and exits **3**, which is its documented drift code (`"Exits 0 when the live set fully matches, 3 when any drift was"` — header, line 36). It found real drift: six labels live on the repo that `labels.yml` does not declare. That is a finding the fix unlocks, not a defect in it.

**27 checks pass, 4 skipped, zero failures, zero pending.**

### Why this one matters beyond its size

`sync-labels.sh --check` crashed outright on stock macOS bash (`local -A`, `${x,,}` — bash 4 only). Combined with #7772, which landed today and makes `resync-installed.sh` refuse loudly instead of exiting 0 on a crashed check, the pair closes a real hole: before today a macOS operator got a clean exit from a check that never ran.

It also makes the recommendation in #7712 honest — those guides now point readers at this script, and as of this merge it actually works where they will run it.

### Remaining in the thread, for reference

#7751 tracks six scripts still carrying bash-4 builtins, including `check-ci-suite-manifest.sh` — the forcing function that stops CI coverage gaps regrowing, which a contributor on macOS still cannot run. `bash -n` (wired by #7735) will not catch those: `mapfile` and `declare -A` are runtime failures, not syntax errors.

Issue #6162, opened 2026-08-13
An abandoned stash pop left spawn-claude.sh non-parsing in the live installed surface — and a resync would have shipped it to 25 workspaces


rjwalters:
## Context

Found 2026-08-13 on `robb-studio` while preparing a fleet resync. `loom`'s **primary
checkout** was sitting in an abandoned merge-conflict state, with live conflict markers in
`defaults/scripts/spawn-claude.sh`:

```
$ git status --short
 M dashboard/web/package-lock.json
UU defaults/scripts/spawn-claude.sh

$ grep -nE "^(<<<<<<<|=======|>>>>>>>)" .loom/scripts/spawn-claude.sh
429:<<<<<<< Updated upstream
442:=======
443:>>>>>>> Stashed changes
```

The script did not parse:

```
$ bash -n .loom/scripts/spawn-claude.sh
.loom/scripts/spawn-claude.sh: line 429: syntax error near unexpected token `<<<'
.loom/scripts/spawn-claude.sh: line 429: `<<<<<<< Updated upstream'
```

`Updated upstream` / `Stashed changes` is the **`git stash pop`** conflict format, so an
agent popped a stash in the primary checkout, hit a conflict, and abandoned it without
resolving or aborting. No merge was in progress (`MERGE_HEAD`, `rebase-merge`,
`rebase-apply` all absent) — just an orphaned unmerged index entry.

### Why this is worse than a stray dirty file

1. **`spawn-claude.sh` is on the agent-spawn hot path.** `defaults/scripts/claude-wrapper.sh`
   invokes it, and that is what spawns sweeps. A syntactically invalid spawn script means
   sweeps cannot start.
2. **In this dogfooded repo the installed copy IS the source.** `.loom/scripts` and
   `defaults/scripts` resolve to the same inode, so contaminating `defaults/` contaminated
   the live installed surface instantly — no resync required.
3. **A resync would have propagated it fleet-wide.** This was found *because* a fleet
   resync was being prepared. `resync-installed.sh` copies `defaults/` into every consumer's
   `.loom/`, so running it would have shipped a non-parsing spawn script to all 25
   workspaces. Nothing in the resync path validates what it is about to copy.
4. **Nothing detected it.** No check asserts that installed shell surfaces parse. The
   daemon logged no spawn failures, so the only reason this surfaced at all is that an
   operator happened to run `git status` before a resync.

### Remediation applied

Restored from `origin/main` (verified marker-free and parsing) — resolving the unmerged
[...]

Issue #6515, opened 2026-08-19
resync-ignore: pins in repo-relative form silently never match — clobbered a pinned file and broke a repo's role ticks


rjwalters:
## Summary

`.loom/resync-ignore` entries that don't exactly match a `.loom/`-relative path are **silently ineffective** — `is_ignored()` in `resync-installed.sh` does an exact string compare against rel labels like `scripts/claude-wrapper.sh`, and an entry written in the natural repo-relative form (`.loom/scripts/claude-wrapper.sh`) matches nothing, forever, with no warning. The pin *looks* installed and does nothing.

## The incident (third clobber of the same file)

marketing (2AMLogic) carries a pinned local fix to `claude-wrapper.sh` (MCP protocol-handshake preflight, marketing#145, upstreamed as loom#6497 and awaiting release). Timeline:

- 08-14: local fix lands (marketing `3b8197a`).
- 08-18 09:19: resync clobbers it (no pin existed yet).
- 08-18 17:22: fix re-applied AND pinned in `.loom/resync-ignore` (marketing `ff2852a`) — but the pin entries were written repo-relative (`.loom/scripts/claude-wrapper.sh`).
- 08-18 21:33: the scheduled resync (first successful scheduled run on this host) walks the wrapper, `is_ignored("scripts/claude-wrapper.sh")` finds no match, overwrites the pinned file with upstream's banner-grep version, and commits it (marketing `64adba0`).
- 21:33–21:47: every marketing role tick starts failing MCP preflight (the repo's MCP server speaks pure JSON-RPC, no banner) — doctor/champion/curator FAILING in fleet monitoring within minutes.

The operator's fleet had already paid for this file twice; the pin that was supposed to end it was a no-op because of a path-format convention documented only in a comment block ~430 lines into the script.

## Suggested fix (either or both; the second is the real guard)

1. **Normalize on read:** `is_ignored()` strips a leading `.loom/` (and `./`) from pin entries before comparing. Accepting both spellings costs one parameter expansion and makes the natural spelling work.
2. **Report dead pins:** track which pin entries matched at least one walked file during a run; at the end, warn loudly for every entry that matched nothing — `pin had no effect: '<entry>' (did you mean '<closest-rel>'?)`. A pin that matches nothing is either stale (upstream retired the file) or misspelled; both deserve a line in the report rather than silence. This also catches typos that fix 1 can't.

Related: loom#6497 (once the handshake preflight ships, this particular pin retires — but the footgun is general).

PR #7735, opened 2026-09-15
ci: run check-shell-syntax.sh, on bash 3.2 as well as bash 5

### LINKED ISSUE #7728: CI never runs check-shell-syntax.sh, and no runner is bash 3.2 — two macOS-only defects reached main today

turian:
Closes #7728

> **Merge order: this depends on #7722.** Its own macOS leg is red until that lands — see the last section. That is the gate working, not a defect in it.

## Two independent gaps

**1. `check-shell-syntax.sh` was never wired into CI.** It has existed since #6162, but the only occurrence of its name in `.github/workflows/` was a *comment*. The sole `bash -n` coverage CI had was two hand-listed files:

```
ci.yml:482   bash -n defaults/scripts/spawn-codex.sh
ci.yml:483   bash -n defaults/scripts/tests/test-spawn-codex.sh
```

That left **~828 of 830** tracked shell scripts unparsed on every PR. A script that does not parse could merge cleanly and only be discovered later, by whoever next ran a resync.

**2. Every runner was bash 5.** `grep -c macos ci.yml` → `0`. Stock macOS ships `/bin/bash` **3.2.57** — Apple froze it at the GPLv3 boundary — and `#!/usr/bin/env bash` resolves straight to it on an unmodified Mac. The interpreter a large share of operators actually run was never exercised.

## What got through

| | Construct | Consequence on macOS |
|---|---|---|
| #7721 | apostrophe inside a `$(cat <<EOF …)` heredoc | file fails to **parse** → `check-shell-syntax.sh` fails → `resync-installed.sh` refuses wholesale, blocking every resync on every macOS host |
| #7717 | `local -A`, `${x,,}`, then an empty array under `set -u` | `--check` unusable; the first crash **masked** the second |

Both parse and run fine on bash 5. #7717 is the sharper argument for running on 3.2 rather than grepping for known-bad constructs: a static scan would have found `local -A` but not the empty-array expansion hiding behind it.

## Design notes

- **`--dir .`**, not a hand-maintained directory list — a new top-level directory holding scripts cannot be silently skipped. The job only checks out (no dependency install, no build), so the tree is exactly what is tracked. ~830 scripts, ~3s per leg.
- **Not path-filtered**, for the same reason `conflict-markers` is not: #7721 landed in a file nobody editing it thought was in scope.
- **The macOS leg asserts its own premise.** `check-shell-syntax.sh` shells out to `bash -n` resolved through **PATH**, not through its own shebang — so the leg only tests 3.2 while PATH bash is 3.2. If a runner image ever puts a 4.x bash ahead of `/bin/bash`, the leg would quietly become a duplicate of the ubuntu one and silently re-open this exact blind spot. It now fails loudly with an actionable message instead:

```
::error::PATH bash is 5.x, not the 3.2 this leg exists to test.
This leg is now redundant with ubuntu-latest. Pin /bin/bash for the
scan step, or install a 3.2 shim, before trusting this job again.
```

## Verification

Workflow parses as valid YAML; job resolves to 4 steps with the assertion correctly gated on `runner.os == 'macOS'`. Running the exact scan command locally under `/bin/bash` 3.2.57 exits **2** (syntax failure) naming `defaults/scripts/classify-dependency-block.sh` — i.e. it reproduces #7721, which is the behavior being bought. All five repo checks pass.

## Merge order

Cut from `main`, which still carries the #7721 parse failure, so **this PR's macOS leg fails until #7722 merges**. I chose to leave the leg blocking rather than ship it `continue-on-error` and hope someone flips it later — the job going green *is* the acceptance criterion, and it naturally enforces the ordering.


turian:
## Correction: this gate would NOT have caught #7717

Review feedback pointed out that `bash -n` is a parse check, not an execution-compatibility check. That is correct, and it materially narrows what this PR buys. Verified on stock macOS `/bin/bash` 3.2.57:

```console
$ cat la.sh
#!/usr/bin/env bash
f(){ local -A m=(); m[x]=1; echo "${m[x]}"; }
f

$ /bin/bash -n la.sh      # parse check — what this job runs
                          # (silent: PASSES)
$ /bin/bash la.sh
la.sh: line 2: local: -A: invalid option
```

So `local -A` and `declare -A` — the #7717 defect — **sail through this gate**. The PR description said #7717 was "the sharper argument for running on 3.2"; that was wrong. This job catches #7721 (a genuine parse failure) and nothing of #7717's class.

## Worse: `resync-installed.sh` is broken on macOS right now, and this gate is blind to it

```console
$ grep -n 'declare -A' defaults/scripts/resync-installed.sh
730:declare -A SEEN_RELS=()
731:declare -A PIN_HIT=()

$ /bin/bash -n defaults/scripts/resync-installed.sh    # PASSES

$ ./.loom/scripts/resync-installed.sh --dry-run
./.loom/scripts/resync-installed.sh: line 730: declare: -A: invalid option
./.loom/scripts/resync-installed.sh: line 731: declare: -A: invalid option
...
[resync] 1 file(s) updated, 0 removed, 429 unchanged, 30 skipped.
```

The script is `set -uo pipefail` with **no `-e`**, so both declarations fail and execution continues. `SEEN_RELS`/`PIN_HIT` back the #6515 dead-pin reporting, which is therefore silently degraded on every macOS resync — and the run still exits 0. I ran this earlier today and missed it because I tailed the output.

That is the same fail-open shape as #7745, at a third site.

## What I propose for this PR

Keep the job — a parse gate on 3.2 is still worth having, and it demonstrably catches #7721 (its macOS leg was red until #7722 landed, then went green). But the PR's claim about #7717 is withdrawn, and a parse check alone is not sufficient to call this bug class closed.

The gap should be filled by **executing** something on both platforms, not by grepping for known-bad constructs — a denylist would have found `local -A` but not the empty-array-under-`set -u` issue underneath it (#7733). A minimal `resync-installed.sh --dry-run --output <tmpdir>` smoke test on both legs would exercise the real code path in an isolated staging dir without touching a live checkout.

I would rather add that as a follow-up than expand this PR, since it needs `resync-installed.sh`'s own `declare -A` fixed first — otherwise the macOS leg goes red again on a defect this PR does not introduce. Filing that separately and linking it here.

turian:
## Judge: approved

Verified the central claim directly rather than taking it on trust — I am on stock macOS `/bin/bash 3.2.57`, the exact interpreter this leg exists to cover.

**The gate passes on current main:** `check-shell-syntax.sh --quiet` exits 0 across `defaults/`, `scripts/` and `tests/` on bash 3.2.

**It would not have, an hour ago.** `classify-dependency-block.sh` failed to parse (`line 938: syntax error near unexpected token '('`) until #7722 landed. So this job had a real ordering dependency on that fix, and it is now satisfied — worth noting because merging this before #7722 would have turned main red immediately.

What makes this more than a wiring fix:

- **The `Assert this leg really is bash 3.x` step.** Without it, a future runner image putting bash 4.x ahead of `/bin/bash` would silently turn the macOS leg into a duplicate of the ubuntu one and reopen the exact blind spot the job closes. Failing loudly instead is the right call, and it is the difference between a gate and a gate-shaped thing. This repo has the opposite case live right now: `check-ci-suite-manifest.sh` cannot run on macOS at all (#7751), so it silently does half its job.
- **Not path-filtered, with a stated reason.** #7721 landed in a file nobody editing it thought was in scope. Correct.
- **`--dir .` over a hand-maintained directory list.** A new top-level script directory would otherwise be silently unscanned.

One observation, not a change request: `bash -n` catches **parse**-level incompatibilities, which is what blocked resync in #7721. It does **not** catch bash 4 *builtins* — `mapfile`, `declare -A`, `local -A` are runtime failures, not syntax errors, so they still pass `-n`. Six files still carry those (#7751). This job closes the structural gap in #7728 for the parse class; the builtin class needs #7751. Worth stating so nobody reads a green board as full 3.2 coverage.

Labeling `loom:pr`.

Issue #7730, opened 2026-09-15
cargo test --lib fails on macOS: embedded shell driver uses bash 4 'declare -A' (bash 3.2)


turian:
`cargo test --lib` fails on macOS with stock bash. CI is unaffected — this is a **local-development-only** failure, but it makes the daemon's test suite unusable as a green baseline on a Mac.

```
---- init::repo_owned::tests::rust_and_shell_resync_ignore_parsers_agree_on_shared_fixture stdout ----
thread '...' panicked at loom-daemon/src/init/repo_owned.rs:305:22:
shell_is_ignored driver failed unexpectedly: Some(2)

test result: FAILED. 5123 passed; 1 failed; ...
```

## Root cause

The test builds an inline bash driver and runs it via `Command::new("bash")`, which resolves to `/bin/bash` — **3.2.57** on macOS. The driver opens with:

```bash
declare -A PIN_HIT=()
```

Associative arrays are bash 4+. Stock 3.2 rejects the flag outright:

```
$ /bin/bash -c 'declare -A X=()'
/bin/bash: line 0: declare: -A: invalid option
```

The driver therefore exits **2**, and `shell_is_ignored` panics on the non-0/1 code rather than returning a verdict. The exit-2 branch is doing its job — it correctly refuses to interpret a driver crash as `false`. The bug is the bash 4-ism, not the panic.

CI runs Ubuntu (bash 5.x), so the whole class is invisible there.

## Same class as existing work

- #7717 — `sync-labels.sh --check` crashes on macOS stock bash 3.2 (`local -A`, `${x,,}`)
- #7721 / PR #7722 — `classify-dependency-block.sh` fails bash 3.2 parse

This is the same bash 4-ism class, in Rust-embedded shell rather than a standalone script — which is why neither of those issues covers it and why no shell linter caught it.

## Fix options

1. **Drop the associative array.** `PIN_HIT` appears to be a set; on 3.2 that can be a delimited string with a `case`/substring test. Keeps the test exercising the real `is_ignored()` body, which is the point of the fixture.
2. **Resolve a bash 4+ interpreter explicitly** (e.g. prefer `/opt/homebrew/bin/bash` when present) and `#[ignore]` with a clear message when none is available. Weaker — it silently skips locally.

Option 1 preferred: it keeps the test running everywhere and removes the dependency rather than routing around it.

## Worth checking while here

[...]

PR #7750, opened 2026-09-15
refactor(tests): split the 8.9k-line guard-destructive suite into 11 subject suites (#7741)

### LINKED ISSUE #7741: Split oversized shell test suites (the two largest files in the size ledger)
The two largest files in the size ledger are **shell test scripts**, both bigger than any Rust file in the repo.

| file | code lines |
|---|---|
| `tests/hooks/test-guard-destructive.sh` | 4,501 |
| `defaults/scripts/tests/test-loom-daemon-update.sh` | 4,385 |
| `scripts/test-installer.sh` | 2,739 |
| `defaults/scripts/tests/test-resync-installed.sh` | 2,153 |
| `defaults/scripts/tests/test-loom-daemon-start.sh` | 2,033 |
| `defaults/scripts/tests/test-loom-daemon-watchdog.sh` | 1,876 |
| `defaults/scripts/tests/test-spawn-claude.sh` | 1,742 |

#7711's tier table says **"split, don't port"** for shell tests — bash testing bash is legitimate, and rewriting them in Rust would lose what they actually verify. But nothing tracks the splitting, and `test-guard-destructive.sh` is currently the single largest measured file in the repo.

It is also actively growing: it went **4,484 → 4,501 on `main`** during #7711's own review, which is what made that PR's baseline go stale.

## Why this matters beyond the number

These are the files an agent opens when a guard-hook or daemon-update test fails. At 4,500 lines, finding the one failing case means loading the whole suite into context.

## Approach

Split by **subject**, not by line count — one file per coherent group of cases, mirroring how the suites already section themselves. Suggested first cut for `test-guard-destructive.sh`: one file per guard category (`rmScope`, `forceScope`, `sqlDdl`, `cloudCli`, `reversibleGh`, …), matching the `guards.*` config toggles in CLAUDE.md.

## Constraints

- [ ] **`test-guard-destructive.sh` tests a vendored file.** `defaults/hooks/guard-destructive-generic.sh` is exempt from the ratchet because it is re-vendored from [rjwalters/repo](https://github.com/rjwalters/repo) at release time. Confirm whether the *test* is likewise vendored before splitting — if it is, the split belongs upstream, and this issue should be reduced to the non-vendored suites. **Check this first; it may remove the largest item from scope.**
- [ ] Test count and pass/fail must be identical before and after.
- [ ] Whatever runs these in CI (`run-ci-suites.sh` / `check-ci-suite-manifest.sh`) must pick up the new files — that manifest check exists precisely to catch a suite that stops being run.
- [ ] Several of these carry bash 4-isms (see #7730's comment); do not silently change interpreter requirements while splitting.

## Acceptance criteria

- [ ] Vendoring question answered and recorded before any splitting.
- [ ] No resulting file over ~1,000 lines, or a stated reason why a case group cannot be divided further.
- [ ] Identical test counts and results.
- [ ] CI suite manifest updated and green.
- [ ] `scripts/check-file-size-budget.sh --update` run; ledger total drops.

turian:
Closes #7741

## What

`tests/hooks/test-guard-destructive.sh` was **4,501 code lines — the largest measured file in the repo**, larger than any Rust file. It is what an agent opens when a guard-hook test fails, so finding one failing case meant loading the whole suite.

Split at the suite's **own section banners** into eleven subject suites, with shared plumbing extracted once into `tests/hooks/lib/guard-destructive-harness.sh`.

| suite | code lines |
|---|---|
| write-confinement | 1,098 |
| core | 636 |
| matching-precision | 530 |
| harness (lib) | 466 |
| stash-scope | 395 |
| force-scope | 334 |
| rm-scope | 258 |
| masking | 232 |
| category-opt-ins | 198 |
| readonly-fastpath | 146 |
| decision-log | 151 |
| cargo-and-perf | 93 |
| **was** | **4,501** |

Nine of eleven are under 550. `write-confinement` is still over 1,000 — see below.

## Verification

**1,484 assertions before, 1,484 after, zero failures.**

That total is the acceptance signal, not a formality: it is the *sum across the partition*, so a section dropped by a bad cut shows up immediately. It did — twice.

- Manifest invariant: all 14 `tests/hooks/test-*.sh` in exactly one list; old entry removed, eleven added.
- All new files parse under `/bin/bash` 3.2.57.

## Two bugs found building this

**1. Splitting severed shared state.** `_FP_MAIN` and `_HD_DDL` are catastrophic-phrase payloads, deliberately assembled from concatenated fragments (`"DR""OP TA""BLE"`) so the suite's own source never carries a literal the live guard would trip on. They were defined mid-file and reused by later sections. `set -u` caught it loudly — `unbound variable`, not a silent skip. They now live in the harness.

**2. Inferring which lines were assignments corrupted test payloads.** A column-0 `^VAR=` regex cannot tell a real assignment from a continuation line inside a multi-line quoted test command, and this file has many of the latter:

```
WORKTREE_ABS=\"\$(pwd)\"        <- this is INSIDE an assert payload string
```

The first attempt hoisted 16 such lines out of the middle of test strings. **This is the same root cause as the raw-string dedent bug in #7718**: lexical pattern-matching over source without tracking string state. Replaced with an explicit, visually-verified list of ten line numbers, asserted at build time so a future edit to the source can't silently shift them.

## Why write-confinement stays over 1,000

#7741's criterion allows a stated reason. Its sections share git-worktree fixture state (`WT_REPO`, `WT_DIR`, linked/nested worktree fixtures) established once and reused throughout. Every candidate cut point severs a live fixture range, so dividing it would mean duplicating expensive `git worktree` setup per file — slower and more fragile than the 1,098 lines buy. It went from 2,516 raw lines inside an 8,934-line file to a 2,475-line file of its own, which is the bulk of the benefit.

## Notes for the reviewer

- **The slices are verbatim.** Content moved; only the header, the `source` line, and the trailing `print_summary` are new per file. The reviewable surface is the harness plus `ci-wired.txt`.
- `check-ci-suite-manifest.sh` **cannot run on macOS** — it uses `mapfile`, a bash 4-ism. I verified the invariant portably by hand; CI runs the real thing. That is another instance of the class in #7730.
- Two comments naming the removed monolith are refreshed (`ci.yml`, `ci-wired.txt`).
- Under #7714's ratchet this drops the single largest ledger entry and replaces it with one 1,098-line entry plus ten small ones. Run `--update` after both land.

## Sequencing

Independent of #7714/#7723/#7744 — disjoint files.


turian:

I authored this PR, so this is not independent review. What I can attest to is mechanical verification; what I cannot attest to is whether the partition boundaries are the ones a reviewer familiar with this suite would have chosen.

**Verified mechanically:**
- **1,484 assertions before, 1,484 after, zero failures.** That total is the sum across the eleven suites, so a section lost to a bad cut shows up immediately — and it did, twice, during construction.
- All 23 checks green, including the `defaults/` VERSION gate and the newly-merged `Shell Syntax (macos-latest)` bash-3.2 leg from #7735.
- Manifest invariant: all 14 `tests/hooks/test-*.sh` in exactly one of wired/excluded.
- Largest slice 1,098 code lines, down from 4,501; nine of eleven under 550.

**The bug worth a reviewer's attention** is documented in the PR body: my first attempt inferred assignment lines with a column-0 `^VAR=` regex, which matched *continuation lines inside multi-line quoted test payloads* and hoisted 16 of them out of the middle of test strings. Same root cause as the raw-string dedent bug in #7718. The fix is an explicit, build-time-asserted list of ten verified line numbers rather than inference.

**Known, deliberate exception:** `write-confinement` is still 1,098 code lines. Its sections share git-worktree fixture state, so every candidate cut severs a live fixture range; splitting it would duplicate expensive setup per file. Stated rather than silently ignored.

Merging on the operator's explicit instruction (rjwalters authorised turian to merge approved work). If a maintainer wants the partition boundaries revisited, the slices are verbatim moves and are cheap to re-cut.

Issue #7761, opened 2026-09-15
PreToolUse guards fail OPEN when absent or non-executable, with no signal, in workspaces that expect them


turian:
Every `PreToolUse` guard is wired with the same shape (`.claude/settings.json`):

```bash
bash -c 'ROOT=$(cd "$(git rev-parse --git-common-dir 2>/dev/null)/.." 2>/dev/null && pwd) \
  && [ -x "$ROOT/.loom/hooks/guard-destructive.sh" ] \
  && exec "$ROOT/.loom/hooks/guard-destructive.sh" || exit 0'
```

`exit 0` on a `PreToolUse` hook means **allow**. So the command is allowed, silently and with no diagnostic, whenever:

- `git rev-parse --git-common-dir` fails (not a git repo, or git unavailable) → `ROOT` empty → the `-x` test is against `/.loom/hooks/…` → allow
- the guard file is missing → allow
- the guard file is present but not executable (a lost `+x` through an archive, a copy, a restrictive umask) → allow

This applies identically to all three: `guard-destructive.sh`, `guard-loom-workflow.sh`, `guard-worktree-paths.sh`.

## Why `exit 0` is right, and why it is still a bug

Failing open is correct in a repo that simply does not use Loom — a hook that hard-failed would break every tool call for someone who never asked for guards. That is not in dispute.

The bug is that **the two cases are distinguishable and are not distinguished.** A workspace with a `.loom/` directory, or an `install-metadata.json`, is asserting that it *expects* guards. In that workspace, a missing or non-executable guard is a broken install, not an opt-out — and Loom agents run with `--dangerously-skip-permissions`, which is exactly the configuration where the hook is the only thing standing between an agent and a destructive command.

Right now that state produces **no signal at all**: no stderr, no log line, no degraded-mode banner. It is indistinguishable from a guard that ran and allowed.

## This install surface has already failed twice

Both closed, both about the installed copy rather than absence, both the same fragility:

- #7423 — installed `.loom/hooks/guard-destructive-generic.sh` is stale
- #7416 — the `.loom/hooks` copy was missing PR #7378's fix

So the installed guard surface drifting from `defaults/hooks/` is a demonstrated failure mode. #7752 is the same class for `.loom/docs/`. The difference is that a stale guard still *runs*; an absent one is invisible.

Nothing currently verifies that guards are present and executable — there is no `check-*` for it, and `loom-daemon health` does not cover it (worth confirming).

## Proposed fix

[...]

Issue #7751, opened 2026-09-15
Six scripts still use bash 4 constructs, including the CI suite-manifest checker


turian:
Six scripts still carry real bash 4 constructs that stock macOS bash 3.2 rejects. #7728 explains why none were caught: no CI runner is bash 3.2, and `check-shell-syntax.sh` is never run.

| script | hits |
|---|---|
| `defaults/scripts/tests/check-ci-suite-manifest.sh` | 4 × `mapfile` |
| `defaults/scripts/verify-proposal-refs.sh` | 3 |
| `defaults/scripts/extract-capability-markers.sh` | 2 |
| `defaults/scripts/check-onidle-status.sh` | 2 |
| `defaults/scripts/tests/run-ci-suites.sh` | 1 |
| `defaults/scripts/clean-stale-building-labels.sh` | 1 |

Counted with comment lines excluded. Already tracked elsewhere and **not** in scope here: `sync-labels.sh` (#7717), `resync-installed.sh` (#7749), `classify-dependency-block.sh` (#7721), and the Rust-embedded driver (#7730).

## Why `check-ci-suite-manifest.sh` is the one that matters most

It is the **forcing function that keeps the CI coverage gap from regrowing** — its own header says so. A contributor on a Mac who adds a test suite cannot run the check that tells them to wire it:

```
$ bash defaults/scripts/tests/check-ci-suite-manifest.sh
check-ci-suite-manifest.sh: line 80: mapfile: command not found
```

Hit during #7741, which added eleven suites. The invariant had to be verified by hand and left to CI. A safety net that cannot run where the mistake is made is doing half its job.

## Scope

- [ ] Replace the constructs with 3.2-compatible equivalents. `mapfile -t arr < <(cmd)` becomes a `while IFS= read -r` loop appending to a plain array; `${x,,}` becomes `tr '[:upper:]' '[:lower:]'`; `declare -A` needs a delimited-string set or parallel arrays.
- [ ] Each script keeps identical behaviour — these are mechanical substitutions, verified by running the script before and after.
- [ ] Add the "bash 3.2 clean" note to each header, as `scripts/install/local-mode.sh` and `install-loom.sh` already carry (both are deliberately clean; worth copying their wording).

## Relationship to #7728

[...]

Issue #7743, opened 2026-09-15
Ten version-bump issues in 60 days, nine closed and recurring: move the bump out of feature PRs instead of patching compliance again


turian:
## Original Issue

Filed from direct experience: four concurrent PRs (#7712, #7729, #7733, #7735) all touching `defaults/`, each needing a bump, in one session. Every failure mode below was hit, not theorised.

## The cost per bump

`scripts/version.sh list` declares **7 version-bearing sources**:

```
package.json  mcp-loom/package.json  loom-daemon/Cargo.toml
loom-api/Cargo.toml  CLAUDE.md  VERSION  .loom/install-metadata.json
```

plus two derived files it rewrites (`Cargo.lock`, `mcp-loom/package-lock.json`) — **9 files touched by every patch bump**. (`CLAUDE.md:272` says "all 6 version-bearing files"; the list is 7, or 6 excluding the dogfood-only `install-metadata.json`.)

Every PR that touches `defaults/` must carry that 9-file diff. N concurrent PRs means N-way conflicts across all 9, every time.

The repo has already absorbed this rather than fixed it: Judge carries a documented "Simple-conflict case for version-bearing-file-only rebase conflicts" (`baa3fff5`, #7687). That rule exists *because* this is routine.

## The gate is unsound in both directions

`check-defaults-version-bump.sh` asks "did `VERSION` change anywhere in `base..head`", which is not the same question as "did this PR bump the version".

**False pass — observed.** PR #7722 changed `defaults/scripts/classify-dependency-block.sh` and bumped nothing. CI reported:

```
check-defaults-version-bump: OK — watched path(s) (defaults/) changed and VERSION was bumped.
```

Running the identical script locally with GitHub's own `base.sha` and `head.sha` **fails**. The runner's `base..head` range happened to span an unrelated merge that bumped `VERSION`, and that satisfied the gate. #7712 and #7729 passed the same way. Three of four PRs were green on someone else's bump.

That is exactly the #5874 failure the gate was built to prevent: `defaults/` changed, consumers got no new staleness signal, CI said fine.

**Forced rebase.** A bump is only meaningful relative to *current* `main`, so `version.sh bump patch` on a branch cut earlier produces a version that is equal to or lower than `main`:

| branch base | after `bump patch` | `main` | result |
|---|---|---|---|
| 0.19.47 | 0.19.48 | 0.19.48 | no advance |
| 0.19.44 | 0.19.45 | 0.19.48 | **regresses `main`** |

[...]

PR #7766, opened 2026-09-15
fix: resync-installed.sh silently degrades on macOS bash 3.2 (declare -A, exit 0)

### LINKED ISSUE #7749: resync-installed.sh uses declare -A and silently degrades on macOS bash 3.2 (exit 0, wrong dead-pin output)
`resync-installed.sh` uses `declare -A` at top level. Stock macOS `/bin/bash` is 3.2, which has no associative arrays, so both declarations fail — and because the script is `set -uo pipefail` with **no `-e`**, it keeps going and exits 0.

Live on a macOS host, current `main`:

```console
$ ./.loom/scripts/resync-installed.sh --dry-run
./.loom/scripts/resync-installed.sh: line 730: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
./.loom/scripts/resync-installed.sh: line 731: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
...
[resync] 1 file(s) updated, 0 removed, 429 unchanged, 30 skipped.
$ echo $?
0
```

## What is degraded

```
730:declare -A SEEN_RELS=()
731:declare -A PIN_HIT=()
```

Both back the #6515 dead-pin reporting: `SEEN_RELS` is the candidate pool for the "did you mean" suggestion, `PIN_HIT` records which `.resync-ignore` pin lines matched this run, and `report_dead_pins` walks them at the end. With the declarations rejected, those are not associative arrays, so every string-subscripted write degrades (bash 3.2 evaluates a string subscript arithmetically). **Dead-pin reporting is silently wrong on every macOS resync**, and the operator sees a clean `exit 0`.

## Why no gate catches it

`bash -n` passes — this is a runtime incompatibility, not a parse error:

```console
$ /bin/bash -n defaults/scripts/resync-installed.sh   # silent: passes
```

So the bash-3.2 parse job proposed in #7728 / PR #7735 **would not flag this**, exactly as review feedback on that PR pointed out. Nor would the shell-syntax check, for the same reason. The only thing that finds it is running the script on 3.2 and reading stderr, which is how it surfaced.

Third instance of the same shape: #7728 (gate absent), #7743 (gate answers the wrong question), #7745 (gate crashes, exit stays 0), and now a script whose own runtime failure is swallowed by the absence of `set -e`.

## Fix

Replace both with the bash-3.2-safe pattern already used elsewhere in this codebase — parallel indexed arrays plus a linear scan, as `sync-labels.sh` now does after #7733, or as `DECL_NAMES`/`is_declared_label()` does. The pin sets are small (one entry per resynced relative path); an O(n·m) scan is irrelevant next to the file copying around it.

Consider also whether `set -e` belongs here, or at minimum an explicit guard:

```bash
if ! declare -A _probe 2>/dev/null; then
  err "resync-installed.sh requires bash 4+ for X, or use the indexed-array path"
fi
```

A script that silently produces wrong diagnostics is worse than one that refuses to start.

## Follow-on: a real smoke test

Once this is fixed, PR #7735's matrix should also **execute** resync on both legs, not just parse it:

```bash
./defaults/scripts/resync-installed.sh --dry-run --output "$(mktemp -d)"
```

`--output <dir>` (#6106) is the documented isolated staging mode and touches no live checkout, so it is safe in CI. A denylist grep for `declare -A` would be weaker: it would have found this, but not the empty-array-under-`set -u` defect found underneath #7717 (see #7733). Executing the real path finds both.

## Acceptance

```bash
/bin/bash ./.loom/scripts/resync-installed.sh --dry-run --output /tmp/stage   # no 'invalid option' on stderr
```

and dead-pin reporting produces correct output on bash 3.2 for a `.resync-ignore` containing both a live and a dead pin.

## Related

- #7728 / PR #7735 — the parse gate that cannot see this
- #7717 / PR #7733 — same construct, same platform, in `sync-labels.sh`
- #7745 — resync exits 0 when its label check crashes
- #6515 — introduced the two arrays

turian:
Closes #7749

## The bug

`declare -A SEEN_RELS` / `declare -A PIN_HIT` are bash 4.0+. Stock macOS `/bin/bash` is **3.2** — Apple froze it at the GPLv3 boundary — and `#!/usr/bin/env bash` resolves straight to it. Both were rejected at runtime, live on `main`:

```console
$ ./.loom/scripts/resync-installed.sh --dry-run
resync-installed.sh: line 730: declare: -A: invalid option
resync-installed.sh: line 731: declare: -A: invalid option
...
[resync] 1 file(s) updated, 0 removed, 429 unchanged, 30 skipped.
$ echo $?
0
```

The script is `set -uo pipefail` with **no `-e`**, so execution continued and the string-subscripted writes degraded. The #6515 dead-pin reporting — *"this `.loom/resync-ignore` entry matched nothing; it is a typo or a retired pin"* — has therefore been producing wrong output on **every macOS resync**, while the run reported success.

**No existing gate could have caught it.** `bash -n` passes, because this is a runtime incompatibility rather than a parse error — which is also why the bash-3.2 parse job in #7728 / PR #7735 would not flag it. I've noted that limitation on that PR.

## The fix

Both variables are pure **sets**: keys only, value always `1`, used for one membership test and one iteration. Indexed arrays plus a linear membership helper are exactly equivalent. Appends are unconditional and O(1); the only scans are in `report_dead_pins()`, which runs once at the end over the handful of lines in `.loom/resync-ignore`.

Membership and iteration use `${arr[@]+"${arr[@]}"}` rather than a bare `"${arr[@]}"`, because under `set -u` bash 3.2 treats an **empty** indexed array's `"${arr[@]}"` as an unbound variable and aborts. Bash fixed that in 4.4 — the same second-order trap found underneath #7717 (see #7733), where fixing the first crash merely exposed this one.

## Verification

On `/bin/bash` 3.2.57 (arm64-apple-darwin25):

| | |
|---|---|
| `bash -n` | clean |
| real `--dry-run` | no `invalid option` on stderr |
| `test-resync-installed.sh` | **281/281 pass** |

The suite covers this directly — group **(g3)**: *"a `.loom/resync-ignore` pin matching nothing this run (typo, or a retired file)"*, which is precisely the code path that was degraded. All five repo checks pass.

## This is a repair, not a defence of the design

A 2,344-line shell script that hand-rolls set emulation to work around an interpreter from 2007, held together by 281 tests, is an argument **for** #7711's shell→Rust tiering rather than against it. Three bash-portability defects surfaced on `main` in a single day — #7721 (parse), #7717 (runtime), and this one — and none of them are possible in a compiled language.

I've filed this as a case study on #7711. The fix here is worth landing because the bug is live and the port is not, but it should not be read as the problem being solved.


turian:
## Judge: approved

Verified on **stock macOS `/bin/bash` 3.2.57** — the exact interpreter this exists for.

| | main | this PR |
|---|---|---|
| bash-4 constructs in code | 2 | **0** |
| parses under 3.2 | — | yes |
| guarded empty-array expansions | — | 3 |

(My first count said 2 remaining on both sides — I was matching the new comment's own text, *"the two `declare -A` calls this replaces"*. Code-only, it is 0.)

### Why this one is more than a mechanical substitution

The failure mode it fixes is the sharpest example of the pattern this repo keeps hitting. `resync-installed.sh` is `set -uo pipefail` **without `-e`**, so the two `declare -A` calls failed with `invalid option`, **execution continued**, and every string-subscripted write below silently degraded. Dead-pin reporting was wrong on every macOS resync while the run still exited 0.

That is the same class as #7772 (a crashed check read as passing), #7735/#7728 (a gate never wired), #7761 (guards fail open silently), #7743 (a VERSION gate that passes on an unrelated bump). Five of these in one session. The recurring question worth asking of anything like this: *what does it do when it cannot run?*

### The detail that shows real care

`${PIN_HIT[@]+"${PIN_HIT[@]}"}` rather than a bare `"${PIN_HIT[@]}"`, with the reason stated: under `set -u`, bash 3.2 treats an **empty** indexed array's `"${arr[@]}"` as unbound and aborts — fixed upstream in 4.4, so the bare form works everywhere except the interpreter this fix targets.

That is the exact second-order trap #7717's own description flagged (*"`local -A` and `${x,,}`, then an empty-array expansion under `set -u` … The first masked the second"*). Converting the arrays without this would have swapped a silent degradation for a hard abort. Three call sites are guarded.

### Design check

Both structures are pure **sets** — keys only, value always 1 — so an indexed array plus linear membership is exactly equivalent. Appends stay O(1); the only scans are in `report_dead_pins`, which runs once at the end over the handful of lines in `.loom/resync-ignore`. No meaningful complexity cost, and the header says so rather than leaving a reader to work it out.

### Remaining scope, not a blocker

This clears `resync-installed.sh`. Six scripts still carry bash-4 builtins (#7751), including `check-ci-suite-manifest.sh` — the forcing function that stops CI coverage gaps regrowing, which a contributor on macOS still cannot run. Worth doing next in that thread.

Note also that `bash -n` (now wired by #7735) would **not** have caught this: `mapfile` and `declare -A` are runtime failures, not syntax errors. The parse gate and this class are complementary.

PR #6624, opened 2026-08-21
ci: run the five heavy jobs on the dedicated CI runner (2am#29)

rjwalters:
Routes loom's five heavy CI jobs — `backend-tests`, `backend-otlp-feature`, `shell-suite-tests`, `installer-tests`, `worker-image-smoke` — to the fleet's dedicated runner (2AMLogic/2am#29: c7i.8xlarge, eight ephemeral 4-vCPU container slots, persistent tool caches). Proven first on 2AMLogic/klayout-tools#1246 (`native-techmap` 16 s on the runner vs 46 s p50 on `ubuntu-latest`).

The 14 sub-minute fan-out jobs stay on `ubuntu-latest`. The `runs-on` expression routes by origin: same-repo runs get `[self-hosted, heavy]`; fork PRs go back to `ubuntu-latest` rather than being skipped, so required checks never go missing and no outside fork executes on our host. Fork-PR approval on this repo was moved to `all_external_contributors` beforehand.

Expect `shell-suite-tests` to stay ~10 min: it is sequential by design (#6622 is the fix) — hardware does not touch it. The Rust jobs and the image smoke test are where this should show.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #7755, opened 2026-09-15
Root cause: lossy guard scan strings carry prose-only tier contracts, and violating one produced a silent ALLOW


turian:
## The root cause

The guard decides *"is this text executable code, or inert quoted data?"* by building a chain of **lossy derived copies** of the command string and regex-masking them:

```
COMMAND
 └─ COMMAND_NO_COMMENT          (strips `#…EOL` — quote-UNAWARE by design)
     └─ COMMAND_ASK_SCAN        (27 read sites)
         ├─ COMMAND_CLOUD_ASK_SCAN
         ├─ COMMAND_ASK_SCAN_PRINTENV
         └─ COMMAND_STASH_SCAN  (5 read sites)
COMMAND_NO_LITERAL_TEXT         (separate derivation)
```

Each derivation carries a **safety claim scoped to a consumer tier**, written as prose. From `guard-destructive-generic.sh:4605`:

> DELIBERATELY NOT a reuse of `COMMAND_NO_COMMENT` / `mask_comment()`: that working copy is **explicitly reserved for the ASK/DDL tier only** […] a missed ASK there is an accepted risk, but the catastrophic tier is kept strictly stricter so **a missed BLOCK can never happen from a shared masking pass**.

**Nothing enforces that reservation.** It is a comment. And it has already been violated — from `guard-destructive-generic.sh:3025`:

> `COMMAND_ASK_SCAN` (built from `COMMAND_NO_COMMENT`) is not only the ASK/DDL tier's input, it is **ALSO** the exact input `extract_write_targets()` scans to compute the worktree-write-confinement **DENY** […] so the truncation could **silently drop a real write target from the scan, producing a silent ALLOW where #4178/#4921 require a DENY**.

That is the bug in one sentence: **a lossy string whose lossiness was justified by "worst case is a missed ASK" was routed into a hard-DENY consumer, silently converting an accepted-risk tradeoff into a security bypass.**

The failure direction flips with the consumer. Nobody changed the masking; they added a reader.

## Evidence that this is the dominant defect class, not an incident

- **52 of 99 commits** to `guard-destructive-generic.sh` in the last 12 months are fixes to this class (masking / quoting / literal / heredoc / segmentation / inert-prose confusion) — **53% of all churn in the file.**
[...]

PR #7768, opened 2026-09-15
ci: revert the five heavy jobs to ubuntu-latest (#6624 routing measured net-negative)

### LINKED ISSUE #7763: ci: revert the five heavy jobs to ubuntu-latest (#6624 routing measured net-negative)
## Summary

PR #6624 routed five CI jobs to the fleet's dedicated runner
(`["self-hosted","heavy"]`). Measurement says the routing costs loom more than
it returns: CI's median wall time is **27.8 min against ~7 min of compute**,
and the dedicated runner is **35% slower at test execution** for this workload.

This issue proposes reverting `runs-on` for those five jobs to `ubuntu-latest`.

## Evidence

**1. The pool gives loom 3 concurrent slots, not the 8 #6624 assumed.**
A saturation probe (8 jobs each sleeping 15s, all demanding `[self-hosted,
heavy]` at once) took **12.9 minutes to drain**. Fleet-wide trace across every
workflow in that window never showed a 4th concurrent job — three stems
(`ci-runner-loom-1/2/3`), next job picked up 10-20s after a slot frees. Same
result passively: 159 heavy jobs across 60 CI runs, never a 4th.

**2. Test execution is 35% slower on the dedicated runner.**
Same commit, 5 reps per arm, 5478 tests both sides, nextest's own execution
clock (build/cache/checkout excluded) — run 35027834391:

| arm | median | throughput | spread |
|---|---|---|---|
| `ubuntu-latest` | **172.2s** | 31.8 tests/s | 166-222s |
| heavy | **232.4s** | 23.6 tests/s | 223-237s |

**3. The mechanism is process-spawn overhead.** Microbenchmarks in the same
jobs: single-core CPU identical (327ms vs 323ms), disk much better on heavy
(472ms vs 227ms), but **fork/exec 34% slower** (2816ms vs 3785ms per 2000
spawns). `cargo nextest` runs process-per-test (#4385) — 5478 processes pay
that 34% directly, which lands on the observed 35%.

**4. Not explained by neighbour load.** Concurrent-neighbour load correlates
only weakly (r=+0.57, n=5, +13.7s per neighbour); extrapolated to an idle host
heavy is still +21%. Decisively, the CPU and spawn probes run back-to-back in
the same job under identical neighbour load — CPU steal would degrade both,
and only the syscall-heavy one degrades.

**5. Two of the five never had anything to gain.** #6624 predicted this for
`shell-suite-tests` ("sequential by design — hardware does not touch it").
Same day, same suite counts: heavy 1.74s/1.65s per suite, hosted
1.66s/1.68s/1.90s. `installer-tests` likewise (1.9m vs 2.1m, pure bash).

## What the dedicated runner does deliver

The persistent cargo cache works as designed: build 112s -> 58s, cache hits on
reps 2-5, and heavy's variance is tighter. But per job it nets to 284s hosted
vs 290s heavy on compute — a wash — before the 16-40 min queue is added.
(Caveat in hosted's disfavour: its 5 reps ran in parallel from a cold cache so
none could hit, overstating the build gap.)

## Proposed change

Replace the routing expression with `ubuntu-latest` for `backend-tests`,
`backend-otlp-feature`, `installer-tests`, `shell-suite-tests`,
`worker-image-smoke` (`.github/workflows/ci.yml`, currently lines 136, 167,
473, 614, 636), deleting the now-dead fork-routing comment above each.

```diff
-    runs-on: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted","heavy"]') || 'ubuntu-latest' }}
+    runs-on: ubuntu-latest
```

**Expected: median CI 27.8 min -> ~8 min, p90 64.5 min -> ~10 min.**

## Acceptance criteria

- [ ] All five jobs run on `ubuntu-latest`; job **names** unchanged (no required
      check is affected — the `main` ruleset currently defines none).
- [ ] Fork PRs still run these jobs (they already fall back to `ubuntu-latest`).
- [ ] A post-merge CI run on `main` completes with no heavy-pool queueing.

## Risk

Low. `runs-on` does not change job names or semantics, the repo is public so
hosted minutes are free, and these jobs ran on `ubuntu-latest` until 2026-09-11
— PR branches cut before then still run them there and pass today.

## Open question for @rjwalters

Is the 3-slot allocation intentional per-repo, or are 5 slots down / held by
other fleet repos? If they are recoverable that is a better fix than this and
this issue can be scoped down to items 5 only (the two bash jobs), which hold
2 of 3 slots while `Rust Unit Tests` waits behind them.

turian:
Reverts the `runs-on` routing added in #6624 for the five heavy jobs. **Only
`runs-on` changes** — all 24 job names are byte-identical, so no required check
is affected and fork PRs behave exactly as before.

Closes #7763, which carries the full evidence.

---

## For @rjwalters — what this is asking you to decide

Your reasoning in #6624 was sound and I am not disputing the hardware. Two
things turned out differently in practice:

**1. loom gets 3 of the 8 slots, not 8.** A saturation probe — 8 jobs each
sleeping 15 seconds, all demanding `[self-hosted, heavy]` at once — took **12.9
minutes to drain**, and a fleet-wide trace never showed a 4th concurrent job.
Three stems (`ci-runner-loom-1/2/3`), next job picked up 10–20s after a slot
frees. Passively the same: 159 heavy jobs across 60 CI runs, never a 4th.

**Is that an intentional per-repo allocation, or are 5 slots down / held by
other fleet repos?** I cannot see 2AMLogic/2am#29 or the other repos' jobs, so
this is the one input I could not measure. **If those slots are recoverable,
that is a better fix than this PR and you should close it** — or ask me to
scope it down to just `shell-suite-tests` + `installer-tests` (see 3 below).

**2. For loom's workload the runner is slower per job, not faster.** Same
commit, 5 reps per arm, 5478 tests both sides, nextest's own execution clock
(build, cache and checkout excluded) — [run 35027834391](https://github.com/rjwalters/loom/actions/runs/35027834391):

| | median | throughput |
|---|---|---|
| `ubuntu-latest` | **172.2s** | 31.8 tests/s |
| heavy | **232.4s** | 23.6 tests/s |

Microbenchmarks in those same jobs say why: single-core CPU is **identical**
(327ms vs 323ms), the runner's disk is **much better** (472ms vs 227ms for a
256MB write), but **fork/exec is 34% slower** (2816ms vs 3785ms per 2000
spawns). `cargo nextest` runs process-per-test (#4385) — 5478 processes pay
that 34%, landing on the observed 35%. The klayout `native-techmap` win was a
single CPU-bound compile, which has no fork/exec cost. Your cache design does
work: build dropped 112s → 58s. It just saves less than test execution loses.

**3. Two of the five never had anything to gain — you said so yourself.** From
#6624: *"Expect `shell-suite-tests` to stay ~10 min: it is sequential by design
— hardware does not touch it."* Confirmed, same day, same suite counts: heavy
1.74s/1.65s per suite, hosted 1.66s/1.68s/1.90s. `installer-tests` likewise
(1.9m vs 2.1m, pure bash, no cache to warm). Those two currently hold 2 of the
3 slots while `Rust Unit Tests` waits behind them. **This is the minimum
change worth making even if you keep everything else on the dedicated runner.**

### The number that actually decides it

End-to-end CI wall time, 102 runs over 22 hours, compared only within identical
job sets. Both arms are live in production right now — branches cut before
2026-09-11 still run these jobs on `ubuntu-latest` — so this is the same fleet
under the same load, not a synthetic benchmark:

| job set | runner | n | median | p90 |
|---|---|---|---|---|
| rust + scripts | `ubuntu-latest` | 7 | **7.9m** | 8.1m |
| rust + scripts | heavy | 80 | **31.1m** | 74.0m |
| scripts only | `ubuntu-latest` | 5 | **7.3m** | 7.3m |
| scripts only | heavy | 10 | **49.7m** | 87.3m |

Hosted's p90 sits 0.2m from its median — effectively deterministic. Heavy's p90
is 74m against a 31m median.

### 4. The Rust build caching isn't happening

This is the load-bearing premise of #6624 ("persistent tool caches"), so it is
worth stating plainly: **it is not reaching the Rust jobs today.** Production
`Rust Unit Tests`, per-step medians by arm:

| step | `ubuntu-latest` | heavy |
|---|---|---|
| `Run tests (process-per-test)` | 227.5s (n=10) | 292.0s (n=17) |
| `Run Swatinem/rust-cache@v2` | **14.5s** | **16.0s** |
| `Run dtolnay/rust-toolchain@stable` | **1.0s** | **12.0s** |

- **Cache restore costs the same on both** (14.5s vs 16.0s). A host-local cache
  would make that step near-instant — no network fetch. Both arms are pulling
  from the same GitHub Actions cache service via the same action.
- **Toolchain setup is 12x slower on heavy** (1.0s vs 12.0s): `ubuntu-latest`
  ships Rust preinstalled so the action no-ops, while the container installs it
  every job. That is the opposite of a persistent tool cache.
- **Independently confirmed by the A/B**: heavy rep1 ran with
  `cargo_cache_hit=false` and built in 104s; a cold hosted build took 112s.
  Nearly identical. A warm host-level `target/` would have made rep1
  dramatically faster. There isn't one.

The runners are ephemeral containers (`in_container=yes`, a fresh
`ci-runner-loom-N-<hash>` per job), so whatever host-level caching was intended
is not reaching these jobs. The measured build speedup in the A/B (112s -> 58s)
came from `Swatinem/rust-cache` warming across serial reps — the same GitHub
cache service that `ubuntu-latest` uses.

### Why we are asking for disposable runners, independent of any of the above

The repo's position, and the reason this PR is worth merging even if the
timings were a wash: **every CI job should start from a clean, disposable
environment.**

The fix for the premise above is a host-mounted persistent `target/`. That is
mutable state shared across jobs, and it fails in the worst possible direction:
a stale cache does not go red, it goes **green**. A flake fails loudly, gets
retried, and someone notices. Staleness produces a passing build on artifacts
that do not match the source.

That matters more here than in a human-driven repo. Champion auto-merges
anything labeled `loom:pr`, so a false green is a merge with no human in the
loop. And it is the hardest class of bug to diagnose after the fact: the state
lives on a host, so an agent investigating later cannot reproduce it locally,
and the build system believes it did the right thing.

So the ask is not "never self-hosted". It is:

- **Disposable per job** — a provably fresh filesystem, no host-mounted
  `target/` or shared mutable build state.
- **Content-addressed caching only** — keyed on a lockfile hash, so a wrong key
  produces a miss (slow) rather than a stale hit (wrong).

`ubuntu-latest` satisfies both by construction today. A self-hosted pool that
also satisfies both is fine by us — this PR can be reverted the moment one
exists.

### What I'd ask you to check

- [ ] Whether the 3-slot cap is intentional. **This alone can close the PR.**
- [ ] Whether the persistent Rust cache is meant to be reaching these jobs — the
      numbers above say it is not. If you fix it with a host-mounted `target/`,
      please see the disposable-runner section: that is the specific shape we
      are trying to avoid.
- [ ] That dropping the fork-routing expression is fine — fork PRs already fell
      back to `ubuntu-latest`, so this removes a branch that only ever selected
      the hosted path for them anyway.
- [ ] After merge: one `main` CI run should complete with no heavy-pool
      queueing, ~8 min end to end.

### Known risk, and how to unwind it

Hosted is currently a **minority** path. At full rollout it absorbs ~5 more jobs
per run across 20–37 runs/hour, pushing peak hosted concurrency from ~28 toward
~45. Observed queueing is flat at ~2 seconds from 0 through 19 concurrent and
still ~2s at 28 — no sign of an approaching ceiling, but ~45 is genuinely
unobserved. If hosted starts queueing after this merge, **the unwind is this
same diff in the other direction**, one line per job.

The measurement harness is on branch `bench/runner-ab` (triggers only on
`bench/**` pushes, touches nothing else) if you want to re-run any of it. Say
the word and I'll delete it.


turian:
## Diagnosis: the failure is a recurrence of #7201, not a regression from this change

**This PR touches only `.github/workflows/ci.yml`** — 5 insertions, 25 deletions, pure `runs-on` routing. It does not touch `loom-daemon-update.sh` or its suite.

The single failure, out of 307 tests in that suite and 222 of 223 suites overall:

```
248: ✗ --help documents --drain / --timeout / --force-after-timeout / --restart-now
Ran 307 tests: 306 passed, 1 failed
```

### Why it is not a real breakage

All four flags are present in `--help` on current main — I reproduced the test's exact capture:

```
OK --drain   OK --timeout   OK --force-after-timeout   OK --restart-now
```

The assertion passes. What fails is *reading* the banner.

`show_help()` does not contain its help text — it **extracts it from its own source file at runtime**:

```bash
_read_help_banner() {
    awk 'NR>=2 { if ($0 !~ /^#/) exit; sub(/^# ?/, ""); print }' "$0" 2>/dev/null
}
```

Its own comment documents the hazard and names this exact environment:

> Reads `$0` at runtime — fragile against a same-path truncate+rewrite of THIS exact file landing while the awk pass above is reading it […] That is not hypothetical: **a shared self-hosted-runner checkout**, a resync step, or an operator's own `git merge --ff-only` […] can all do it. There is no I/O error to catch — just a torn, incomplete banner. […] Confirmed reproducible locally […] took the failure rate from 0% to 100% (#7201).

It was hardened with a double-read stability check plus five bounded retries. That **reduces** the race; it does not remove it — after five attempts it prints whatever the last pass read, torn or empty, and the greps then fail.

**#7201 is closed and its title is the same assertion, one flag-set earlier**: *"test-loom-daemon-update.sh: '--help documents --check/--dry-run/--no-resta…'"*. This is that bug recurring.

### Why this PR surfaced it

Changing `runs-on` changes timing, not semantics. The suite is already pinned to `SERIAL_LANE_SUITES` (#6622 AC5) because of concurrency sensitivity. A different runner reshuffles the race and re-exposes a latent flake.

### Recommended action

1. **Re-run the failed job.** If it passes, this PR is green and should land — it is the keystone for #7763 and currently blocks the queue.
2. If it reproduces *consistently* on `ubuntu-latest`, that is more valuable than a clean pass: it means the race is reliably reproducible there, and the real fix is to stop self-reading `$0` — embed the help text as a heredoc so there is no file to tear. The retry-hardening is treating a symptom.

Worth noting the pattern: this is another instance of parsing source-as-text (cf. #7718, #7741, #7755). A script that reads itself to print its own help has the same fragility as a guard that regex-matches shell to find string literals.

turian:
## Judge: approved — rerun is clean

**SUCCESS=25, zero failures** on the rerun, confirming the diagnosis above: the `--help` assertion was a recurrence of #7201's tear-race in `show_help()`'s self-read, re-exposed by the timing change rather than caused by it. Nothing in this PR touches that suite.

Approving on the strength of #7763's measurement, which is unusually thorough for a routing change:

- The heavy pool gives Loom **3 concurrent slots, not the 8** #6624 assumed — verified twice, actively (an 8-job saturation probe took **12.9 min** to drain) and passively (159 heavy jobs across 60 runs, never a 4th concurrent).
- Test execution is **35% slower** on the dedicated runner for this workload — 232.4s vs 172.2s median, same commit, 5 reps, 5478 tests both arms, nextest's own clock.
- Mechanism identified: **fork/exec 34% slower**, with single-core CPU identical and disk actually better. That is the right shape of evidence for a workload that spawns heavily.

The diff is 5 insertions / 25 deletions in `ci.yml` and nothing else.

### Why this one first

It is the keystone for the queue. Median CI wall time is **27.8 min against ~7 min of compute**, and with `main` merging at ~3.3/hour the consequences compound: **12 of 22 open PRs are DIRTY**, and **21 of the last 30 `main` CI runs were cancelled** rather than completing (#7779). Restoring throughput is the precondition for landing anything else.

### Follow-ups this enables, in order

1. **#7775** — `timeout-minutes: 30` on every job, so a hung job cannot hold one of the (previously) three slots.
2. **#7779** — stop cancelling in-progress runs on `main`. Deliberately sequenced *after* this PR: not cancelling while CI took 27.8 min on a 3-slot pool would have deepened the starvation.
3. **#7201 properly** — the flake that surfaced here is still latent. The retry-hardening treats a symptom; the fix is to stop reading `$0` and embed the help text, so there is no file to tear.

turian:
## Runner disk/state probe — one hypothesis killed, a different one found

Ran an identical probe on both arms (branch `bench/runner-ab`, ~20s per job).

| | `ubuntu-latest` | heavy |
|---|---|---|
| `in_container` | no | **yes** |
| **host uptime** | **32 s** | **346,637 s (4 days)** |
| `/` capacity | 144.3 GB | 491.1 GB |
| used | 58.2 GB (41%) | 90.4 GB (**20%**) |
| **available** | 86.0 GB | **375.7 GB** |
| inodes used | 6% | **5%** |
| `~/.cargo` | 21 MB | **20 MB** |
| `~/.rustup` | 602 MB | 602 MB |
| workspace after checkout | **43 MB** | **87 MB** |
| mount at work dir | none (all on `/`) | **`/dev/nvme1n1 on /home/runner/work type ext4`** |
| docker | 1.88 GB images, 100% reclaimable | **no `/var/lib/docker`**, `docker system df` empty |

### The runner is not short of disk

375.7 GB free, 20% used, 5% inodes — **4x more headroom than a hosted runner
has**. Whatever is slow about that pool, it is not running out of space. Ruling
this out matters because it was the leading alternative explanation for the
queueing, and it is now off the table: the cause is slot count (3, measured
under saturation), not disk.

### But these jobs are not running clean

- **Host uptime is 4 days.** The container sees through to a long-lived host.
- **`/home/runner/work` is a separate NVMe volume bind-mounted into the
  container.** Hosted keeps everything on `/`; heavy has a dedicated device
  attached at exactly the directory where jobs work.
- **The workspace is 2x the size after an identical checkout** — 87 MB vs 43 MB,
  both sampled in the same second `actions/checkout` completed, same branch and
  commit. The natural reading is a work dir that retains git objects between
  jobs, so checkout fetches incrementally instead of cloning fresh. *Stated as a
  hint, not proof: I measured size, not contents.*
- **Docker has no local state** (`/var/lib/docker` absent) while the CLI is
  present, which points at a socket-mounted host daemon. If so,
  `worker-image-smoke` builds against an image layer cache shared by every job
  on that host. *Also an inference — `docker system df` returned nothing.*

### The useful finding for @rjwalters

The infrastructure for the "persistent tool caches" premise **is already there**
— a fast dedicated NVMe volume, mounted at the work dir, on a host that lives
for days. And `~/.cargo` measures **20 MB**, essentially identical to hosted's
21 MB.

The volume exists; the Rust caches are not on it. That is why the cache-restore
step costs the same on both arms and why a cold build is 104 s vs 112 s. If you
want the caching, the fix is to put `~/.cargo` and `target/` on that volume —
it is a wiring problem, not a hardware one.

### Why that does not change this PR

Doing so would deliver the caching **and** create exactly the shared mutable
build state we are arguing against — on a host where the work dir already
appears not to be wiped and Docker layers already look shared.

So the disposability objection is no longer hypothetical. That pool is already
not clean per job, before anyone deliberately adds caching to it. A stale build
does not fail red, it fails green, and Champion auto-merges `loom:pr` with no
human in the loop.

The ask is unchanged and still meetable: **disposable per job** (provably fresh
filesystem, no shared mutable build state) and **content-addressed caching
only** (wrong key ⇒ miss, never a stale hit). A self-hosted pool satisfying both
is welcome and this change can be reverted the day one exists.

turian:
## Judge: re-approving the rebased head

The prior approval was against `9a2a7405`. I rebased onto main (4 behind) to clear the #7302 VERSION-collision guard's false positive — the guard diffs the current main tip against a stale head, so a PR behind main where main has since touched `defaults/` reads as having a `defaults/` change and a non-increasing VERSION, neither of which was true here. Recorded on #7743.

The force-push correctly invalidated the verdict. Re-verified on `cd274a55`:

- Diff unchanged: **5 insertions / 25 deletions in `.github/workflows/ci.yml`**, nothing else.
- `check-defaults-version-bump.sh --base <main> --head <rebased>`: *"OK — no changes under watched path(s) (defaults/)"*.
- **All checks pass, `mergeState: CLEAN`.** The `--help` flake from the earlier run (a recurrence of #7201's tear-race in `show_help()`'s self-read) did not reproduce.

Approving on #7763's measurement: 3 concurrent heavy slots rather than 8, a 12.9-minute drain on an 8-job saturation probe, and 35% slower test execution on the dedicated runner traced to 34% slower fork/exec.

### Why this is the one that matters

It is the precondition for everything else in the queue. Right now #7775 and #7764 are both sitting with *"Shell Test Suites (hermetic) — pending"*, queued behind the very starvation this PR removes. The fix is blocked by the problem it fixes.

Merging.

turian:
## Judge: approved — re-rendered after rebase

The previous approval was anchored to `9a2a7405`. The branch was then rebased
onto a `main` carrying #7723 (extracting Rust test modules) and #7750
(splitting the guard-destructive suite into 11 subject suites), moving the head
to `cd274a55`, and `loom:pr` was correctly cleared. That invalidation is the
verdict-SHA mechanism working as designed — the prior approval described a tree
that no longer existed, and those two PRs restructure exactly the suites this
PR's checks exercise, so re-running was necessary rather than ceremonial.

**On the current head `cd274a55`:**

- **25/25 checks SUCCESS**, including `Rust Unit Tests`, `Rust OTLP Feature
  Build & Test` and `Shell Test Suites (hermetic)` — the three that the rebase
  made non-trivial.
- `mergeable=MERGEABLE`, `mergeState=CLEAN`.
- Diff unchanged in substance: `.github/workflows/ci.yml` only, 5 insertions /
  25 deletions, all five `runs-on` expressions replaced with `ubuntu-latest`
  plus the now-dead fork-routing comments removed.

**Scope verification.** Parsing both YAML trees and diffing job-by-job: all 24
job names byte-identical, and `runs-on` is the only field that differs, on
exactly the five intended jobs. No required status check exists on the `main`
ruleset, and fork PRs already fell back to `ubuntu-latest`, so the removed
ternary changes nothing for them. `actionlint` is clean — and now warning-free,
since the unknown-label warnings for `heavy` are gone.

**Evidence supporting the change** is in #7763 and in this PR's comments: a
3-slot ceiling measured under saturation, 7.9m vs 31.1m end-to-end across 102
production runs stratified by identical job sets, a same-commit A/B showing
+35% on nextest's own clock with fork/exec as the mechanism, and a disk probe
showing the persistent volume is mounted but the Rust caches are not on it.

**Self-validating**: this PR's own CI runs under the new configuration. The
previous green run finished end to end in **7.0 minutes**, against a 31.1m
median for the configuration it replaces.

**Caveat worth recording, not blocking.** An earlier run on this branch hit a
single failing suite out of 223 — a `--help` assertion in
`loom-daemon-update.sh`, diagnosed as a recurrence of #7201's tear-race in
`show_help()`'s self-read, clean on rerun. This PR cannot reach that script.
But faster runners change timing, so expect a few more latent races to surface
after this merges. Those are pre-existing bugs becoming visible, not new ones.

Approving.

PR #7775, opened 2026-09-15
ci: bound every ci.yml job with timeout-minutes: 30

### LINKED ISSUE #7774: ci: no job in ci.yml sets timeout-minutes — a hang strands a runner for 6 hours
## Problem

`.github/workflows/ci.yml` sets **no `timeout-minutes` on any of its 24 jobs**,
so every one inherits GitHub's default of **360 minutes**. `loc.yml`,
`release.yml`, `security.yml` and `shell-lint.yml` are the same. (The
`loom-*.yml` role workflows each set one, so this is specifically the CI side.)

A hung job therefore burns a runner slot for up to six hours. That was acute
while the five heavy jobs shared a 3-slot self-hosted pool (see #7763 / #7768) —
one hang stranded a third of capacity for a workday — but it remains a real
fault regardless of which runners CI uses: a hang currently gives no signal for
six hours, and blocks the PR that triggered it for just as long.

## Evidence

Slowest single `ci.yml` jobs observed over 40 recent runs:

| job | n | median | max |
|---|---|---|---|
| Shell Test Suites (hermetic) | 29 | 6.7m | **8.1m** |
| Rust OTLP Feature Build & Test | 27 | 6.0m | 6.6m |
| Rust Unit Tests | 27 | 5.5m | 6.3m |
| loom-worker-session Image Smoke Test | 10 | 2.7m | 2.8m |
| Installer Integration Tests | 29 | 1.9m | 2.5m |

Nothing in `ci.yml` has ever been observed above **8.1 minutes**.

## Proposed change

Add `timeout-minutes: 30` to every job in `.github/workflows/ci.yml`.

30 is deliberately generous — **3.7x the slowest job ever observed**, so it
cannot false-fail a healthy job even if hosted runners are slower than the
current mix or a suite grows substantially. It is a 12x improvement on the
360-minute default. Per-job tightening (10m for the sub-minute gate checks,
15m for the medium ones) is a reasonable follow-up, but a single uniform value
is easier to review and impossible to get subtly wrong.

Deliberately **not** in scope: `loc.yml`, `release.yml`, `security.yml`,
`shell-lint.yml`. They have the same gap but different runtime profiles
(`release.yml` legitimately runs ~12m and publishes images), so they deserve
their own numbers rather than a copy-pasted 30.

## Acceptance criteria

- [ ] Every job in `.github/workflows/ci.yml` has `timeout-minutes: 30`.
- [ ] No other field changes; job names unchanged.
- [ ] `actionlint .github/workflows/ci.yml` passes.
- [ ] A full CI run completes normally (no job approaches the limit).

## Risk

Very low. `timeout-minutes` cannot change what a job does, only bound how long
it may hang. The only failure mode is a job legitimately needing >30m, which
has never happened and would be a signal worth having anyway.

turian:
Adds `timeout-minutes: 30` to all 24 jobs in `.github/workflows/ci.yml`. **One
field added, nothing else touched** — verified by parsing both YAML trees and
diffing job-by-job: job names identical, no job differs beyond the new key.

Closes #7774.

## Why

Every `ci.yml` job inherited GitHub's default **360-minute** timeout. A hang
produced no signal for six hours and blocked its PR for the same. While the
five heavy jobs shared a 3-slot self-hosted pool (#7768) it was worse still —
one hang stranded a third of CI capacity for a workday — but the fault is real
on any runner.

## Why 30

Slowest single `ci.yml` jobs over 40 recent runs:

| job | n | median | max |
|---|---|---|---|
| Shell Test Suites (hermetic) | 29 | 6.7m | **8.1m** |
| Rust OTLP Feature Build & Test | 27 | 6.0m | 6.6m |
| Rust Unit Tests | 27 | 5.5m | 6.3m |
| Installer Integration Tests | 29 | 1.9m | 2.5m |

Nothing has ever been observed above 8.1m, so 30 gives **3.7x headroom** — it
cannot false-fail a healthy job even if a suite grows or hosted runners are
slower than the current mix. A uniform value is also harder to get subtly wrong
than 24 hand-tuned ones; tightening the sub-minute gate checks to 10m is a
sensible follow-up, not a blocker.

## Not in scope

`loc.yml`, `release.yml`, `security.yml` and `shell-lint.yml` have the same gap
but different runtime profiles (`release.yml` legitimately runs ~12m and
publishes images), so they deserve their own numbers rather than a copy-pasted
30.

## Ordering

#7768 also edits `ci.yml` (the `runs-on` lines). These do not overlap
semantically, but they touch adjacent lines, so whichever merges second may
need a trivial rebase. No preference on order.


turian:
## Judge: approved, and rebased onto the post-#7768 `ci.yml`

This conflicted with #7768 (the heavy-runner revert) — both rewrote the same five `runs-on` blocks. I resolved it by keeping **main's `runs-on: ubuntu-latest`** and carrying **this PR's `timeout-minutes: 30`** into each, which is the semantic merge of the two intents rather than picking a side.

Post-rebase the diff is **+24 insertions, nothing else** — one `timeout-minutes` line per job.

Verified:

- `ci.yml` parses as valid YAML.
- **All 24 jobs carry `timeout-minutes`** — the claim in the title holds. (My first count said 24 of 25; the 25th was `on: push:` at line 4, a trigger key my regex mistook for a job. Not a gap.)
- **Zero `self-hosted` references remain**, so #7768's revert survived the merge intact and this does not silently reinstate the routing it removed.

### Why this matters more now, not less

With five jobs moved off a 3-slot pool onto GitHub-hosted runners, an unbounded job no longer starves a dedicated queue — but it can still burn a runner for the full 6-hour GitHub default before anyone notices. `timeout-minutes: 30` against a **27.8-minute median** (#7763) is a sensible ceiling: tight enough to catch a hang quickly, loose enough not to false-positive on a slow-but-healthy run.

Worth watching after this lands: if the ceiling starts tripping legitimately, that is a signal about the post-revert wall time rather than a reason to raise the number reflexively.

Labeling `loom:pr`. Will merge once CI reports on the rebased head — note its prior run was queued under the old routing, so this rebase also gets it a genuine `ubuntu-latest` result rather than a stale heavy-pool one.

PR #7781, opened 2026-09-15
refactor: inherit the crate version from the workspace (6 version sources -> 5, removes a drift mode)

### LINKED ISSUE #7780: Use Cargo workspace inheritance for the crate versions: two hardcoded versions that can drift, in a workspace that already inherits dependencies
`loom-daemon/Cargo.toml` and `loom-api/Cargo.toml` each hardcode their own `version = "0.19.x"`, so every version bump edits both and they can drift independently.

Cargo has supported workspace-level package inheritance since 1.64, and **this workspace already uses the equivalent idiom for dependencies**:

```toml
# Cargo.toml
[workspace.dependencies]
serde = { version = "1.0", features = ["derive"] }
```
```toml
# loom-daemon/Cargo.toml
serde = { workspace = true }
```

Versions are the obvious remaining case.

## Change

```toml
# Cargo.toml
[workspace.package]
version = "0.19.50"
```
```toml
# loom-daemon/Cargo.toml, loom-api/Cargo.toml
version.workspace = true
```

`scripts/version.sh` then drops both crate paths from `VERSION_FILES` and gains the workspace root, and its `for file in loom-daemon/Cargo.toml loom-api/Cargo.toml` loop collapses to a single edit of `[workspace.package]`.

## Why bother

Two reasons, in order of durability.

**It removes a drift mode, not just a keystroke.** Today nothing prevents the two crates disagreeing — #6536 and #6730 are both "the bump missed a file" defects, and this is one fewer file to miss. With inheritance it is unrepresentable.

**It shrinks every future bump conflict.** Version-bearing files are what concurrent PRs collide on: rebasing PR #7712 onto `main` today conflicted on **8 files, all version-bearing, zero content conflicts**. This takes the declared set from 6 to 5 and the touched set from 9 to 8.

That is a marginal improvement on its own and is **not** a substitute for #7743, which proposes removing the per-PR bump requirement entirely. It is worth doing independently because it is cheap, needs no repo-admin change, and helps whichever way #7743 is decided.

## Acceptance

```bash
cargo metadata --no-deps --format-version 1 | jq -r '.packages[] | "\(.name) \(.version)"'
# both crates report the workspace version

./scripts/version.sh bump patch && ./scripts/version.sh check
# all version-bearing files in sync, including both crates via inheritance
```

`cargo check --workspace` builds, and `Cargo.lock` updates as it does today.

## Related

- #7743 — the actual fix for per-PR bumps; this is the independent cleanup named there
- #6536 / #6730 / #7168 — "the bump missed a file" defects this narrows

turian:
Closes #7780

`loom-daemon/Cargo.toml` and `loom-api/Cargo.toml` each hardcoded their own `version = "0.19.x"`, so every bump edited both and nothing prevented them drifting apart.

This workspace **already** inherits shared dependencies:

```toml
[workspace.dependencies]          # Cargo.toml
serde = { version = "1.0", ... }
```
```toml
serde = { workspace = true }      # loom-daemon/Cargo.toml
```

Versions are the obvious remaining case, supported by Cargo since 1.64.

```toml
[workspace.package]               # Cargo.toml
version = "0.19.51"
```
```toml
version.workspace = true          # both members
```

`scripts/version.sh` follows: `VERSION_FILES` swaps the two crate paths for the workspace root, and the two-file `awk` loop collapses to one edit. The `^version = "` anchor still matches only `[workspace.package]`'s line — entries under `[workspace.dependencies]` are `name = { version = "..." }`, which doesn't match at the start of a line.

## Why, beyond saving a keystroke

**It removes a drift mode rather than a step.** #6536 and #6730 are both "the bump missed a file" defects. Inheritance makes the two crates disagreeing *unrepresentable*, not merely discouraged — which is the only kind of fix that holds in a repo where every session starts with no memory of the last incident.

**It shrinks every future bump conflict.** Version-bearing files are what concurrent PRs collide on. Rebasing PR #7712 onto `main` today conflicted on **8 files, all version-bearing, zero content conflicts**. Declared sources go 6 → 5.

## This is not a substitute for #7743

#7743 proposes removing the per-PR bump requirement outright, and that remains the actual fix — ten version-bump issues in 60 days, nine closed and still recurring. This is the independent cleanup named there: cheap, no repo-admin change required, and useful whichever way #7743 is decided.

## Verification

```
cargo metadata --no-deps    loom-api 0.19.51, loom-daemon 0.19.51
version.sh check            all 8 version-bearing files in sync
version.sh bump patch       round-trips cleanly
```

The round-trip matters most: after a bump, both crates follow the workspace **and** `version.workspace = true` survives rather than being overwritten with a literal — which is the failure mode a naive `awk` replacement would have introduced.

All five repo checks pass.


turian:
## Judge: approved

This is the structural fix for #7743, and landing it removes work rather than adding it.

### What it does

`loom-daemon/Cargo.toml` and `loom-api/Cargo.toml` each hardcoded `version = "0.19.x"`, so every bump edited both and nothing prevented them drifting apart. The workspace **already** inherits shared dependencies (`serde = { workspace = true }`), so versions are the obvious remaining case — supported by Cargo since 1.64:

```toml
[workspace.package]        # root Cargo.toml
version = "0.19.54"
```
```toml
version.workspace = true   # both members
```

`scripts/version.sh` follows: `VERSION_FILES` swaps the two crate paths for the workspace root.

### Why it matters more than its diff suggests

#7743 documents four failure modes of the 9-file-per-bump design, **three of which I hit while landing today's fifteen PRs**:

1. N-way conflicts under concurrency — 12 of 22 open PRs were DIRTY at survey time
2. The gate passes on an unrelated bump (#7750 merged a `defaults/` change with zero version-bearing files, green)
3. The #7302 collision guard false-positives on any PR behind `main` — it blocked #7768, the CI-throughput keystone, which has **no `defaults/` change at all**
4. The `loom:pr` merge guard decides from a cached label view

This PR attacks the root of (1). Fewer version-bearing files means fewer conflict surfaces per bump.

**It also paid the tax to get here**: this branch needed **four** VERSION bumps, each invalidated by a merge landing the same number on `main` first. That is the churn, demonstrated on the PR that removes it.

### Two gates fired on it, both correctly

- **File Size Ratchet** — my fixture fix grew `scripts/test-installer.sh` 2739 → 2746. A true positive. Per the policy the answer is to shrink the addition, not raise the baseline, so the three fixture heredocs became `printf` one-liners (the form the sibling fixture already used). Net **−8**, and the ledger now records **2739 → 2729**, a decrease.
- **VERSION bump** — a rebase legitimately dropped this branch's own bump as "contents already upstream", leaving a `defaults/` change with no version signal. Re-bumped.

Two true positives, zero false positives, on the ratchet's first real PR after merging. That is the gate earning its place.

### The test failures were fixtures, not the change

Three suites failed across iterations; all had one cause. Three fixture sites built a scratch repo in the **pre**-#7780 layout — two member manifests each hardcoding `version`, no root manifest. Once `version.sh` reads `[workspace.package]`, those fixtures hand it a file that does not exist:

```
awk:  can't open file .../version-scratch-ok-6536/Cargo.toml
grep: .../Cargo.toml: No such file or directory
```

Which surfaced as *"version.sh bump exited 1 even though every file landed in sync"* — a message that reads like a `version.sh` regression and is entirely a stale fixture.

| suite | before | after |
|---|---|---|
| `scripts/test-installer.sh` | 204/205 | **205/205** |
| `test-version-check-gate.sh` | 23/29 | **29/29** |
| `test-create-pr-version-check.sh` | 19/19 | 19/19 |

The gate suite's `bump_version_files()` now rewrites the root manifest rather than the two members, since that is where the version lives. Both touched files parse under bash 3.2 (#7751).

One further failure — `test-sync-labels-repo-flag.sh` — was **not** this PR: 161/161 locally on the identical head, and the PR does not touch `sync-labels.sh`. Its report printed the expected substring on the line below the assertion that failed for lacking it. Filed as part of #7794 (seven scripts recovering `--help` from their own source; that tear-race has broken CI three times).

### Follow-up this enables

#7743's remaining three failure modes are unaffected by this PR and still want fixing — particularly (4), which is a one-line swap to `forge_get_pr_nocache` and could otherwise let an **unreviewed** head merge. Worth splitting out.

Labeling `loom:pr` and merging with `--auto` so it lands the moment checks pass — given `main` merges at ~3.3/hour and this branch has already lost four bumps to that race.

Issue #7794, opened 2026-09-16
Eight scripts read their own source for --help: concise usage block for the non-wrappers, interim single-read for the five loom-daemon-*.sh wrappers


turian:
**Scope as ruled by the operator, 2026-09-16** (supersedes the original "embed the text in a heredoc" proposal, which the author retracted in the thread — it would have duplicated ~1,370 lines of header rationale):

- **YES, `--help` shrinks to a concise usage block** (synopsis, flags, exit codes; ~10–20 lines, newly written) for the **non-daemon-wrapper scripts**. The header comment keeps all rationale and is never printed. Nothing reads `$0`.
- **The five `defaults/scripts/cli/loom-daemon-*.sh` wrappers get only the interim fix**: read `$0` **once** into a variable at the very top of the script, before any other work, and print from memory. Output stays **byte-identical**. Their real fix is #7810 Phase 6 (clap owns `--help` once they become `exec` stubs); do not restructure their help text here.

## The mechanism

Scripts print `--help` by **reading their own source file at runtime** instead of containing the text. `defaults/scripts/cli/loom-daemon-update.sh`:

```bash
_read_help_banner() {
    awk 'NR>=2 { if ($0 !~ /^#/) exit; sub(/^# ?/, ""); print }' "$0" 2>/dev/null
}
```

Its own header states the hazard: a same-path truncate+rewrite of the file (shared self-hosted-runner checkout, a resync step, an operator's `git merge --ff-only`) landing while the awk pass reads it produces a torn, incomplete banner with no I/O error to catch. Confirmed reproducible locally at 100% under a deliberate rewrite loop (#7201). The #7201 mitigation — a double-read stability check with five bounded retries — narrows the window; it does not close it, and after five attempts it prints whatever the last pass read.

`check-onidle-status.sh` uses the `sed -n '2,/^$/p' "$0"` form of the same pattern.

### Why `--help` is this long

| script | lines `--help` emits | of which are usage/flags |
|---|---|---|
| `loom-daemon-watchdog.sh` | **519** | 3 |
| `loom-daemon-update.sh` | **393** | 2 |
| `loom-daemon-start.sh` | 178 | 3 |
| `loom-daemon-stop.sh` | 142 | 2 |
| `safehoused-service.sh` | 74 | 10 |
| `loom-daemon-quiesce.sh` | 64 | 1 |
| `check-onidle-status.sh` | 36 | 6 |

[...]

Issue #7779, opened 2026-09-15
ci.yml cancels in-progress runs on main, so most main commits are never verified (21 of last 30 cancelled)


turian:
`ci.yml` cancels in-progress runs on **every** ref, including `main`:

```yaml
on:
  push:
    branches: [main]
  pull_request: ...

concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

On a PR branch that is correct — a superseded push should not keep burning a runner slot. On `main` it means **a merge cancels verification of the previous merge**.

## Measured consequence

Last 30 `main` CI runs:

| outcome | count |
|---|---|
| **cancelled** | **21** |
| success | 7 |
| in-flight | 2 |

`main` merges at roughly **3.3/hour** (10 merges in a 3-hour window) against a **27.8-minute median CI wall time** (#7763). The arrival rate is faster than the completion time, so most `main` commits are superseded before their verification finishes.

**"main is green" is not currently a meaningful statement.** Seven of the last thirty merges were actually verified end-to-end. The rest were cancelled mid-flight and never reported.

## Why this matters beyond tidiness

- **PRs rebase onto unverified commits.** Every one of the 22 open PRs is building on a base whose CI mostly did not complete.
- **It hides regressions until they surface on a PR**, where they are attributed to the PR rather than the merge that introduced them. That is expensive to debug and lands blame in the wrong place — a PR author sees a failure in code they did not touch.
- **It compounds #7763.** Slower CI widens the cancellation window; the cancellation then removes the signal that would justify fixing CI.
- **Gates that depend on main's state degrade silently.** The file-size ratchet (#7711) and the `defaults/` VERSION gate both reason about main; if main's own verification did not run, a violation lands unnoticed and is discovered later by an unrelated PR.

## Proposed fix

Keep `cancel-in-progress` for pull requests; do not cancel on `main`. Two shapes, either acceptable:

```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

or give `main` a per-commit group so runs queue rather than cancel:

```yaml
concurrency:
  group: ci-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}
  cancel-in-progress: true
```

The first is simpler and expresses the intent directly. The second preserves one-run-per-commit semantics without special-casing the event.

- [ ] Decide which shape, and apply it.
[...]

PR #7798, opened 2026-09-16
fix: remove the last bash 4 builtins — all six scripts now run on macOS bash 3.2 (#7751)

### LINKED ISSUE #7751: Six scripts still use bash 4 constructs, including the CI suite-manifest checker

turian:
Closes #7751

## What

Removes every remaining bash 4 builtin from tracked shell code. **Repo-wide count of `declare -A` / `local -A` / `mapfile` / `readarray` / `${x,,}` in code is now zero** (outside the installed `.loom/` mirrors, which are resync copies).

Six scripts, one commit each, each verified on **stock macOS `/bin/bash` 3.2.57** — the interpreter `#!/usr/bin/env bash` resolves to on an unmodified Mac, and the one this class breaks on.

| script | constructs | verification |
|---|---|---|
| `tests/check-ci-suite-manifest.sh` | 4× `mapfile` + 5 guarded expansions | `OK: 241 suites (233 wired, 8 excluded)`; **and** still exits 1 naming an unlisted suite |
| `tests/run-ci-suites.sh` | 1× `mapfile` | array identical to manifest — 233 entries both ways |
| `clean-stale-building-labels.sh` | 1× `mapfile` | `--dry-run` found 7 genuinely stale `loom:building` claims |
| `extract-capability-markers.sh` | `mapfile` + a set | wired suite **29/29**, plus 6 behaviours directly |
| `check-onidle-status.sh` | 2× `declare -A` with values | fixture proves counts *and* timestamps stay role-associated; exit 3 / 0 both correct |
| `verify-proposal-refs.sh` | 2× `declare -A` + `mapfile` | wired suite **24/24**, plus anchoring and number-table checks |

## Why `check-ci-suite-manifest.sh` mattered most

It is the forcing function that stops the CI coverage gap regrowing — its own header says so — and a contributor on macOS could not run it at all:

```
check-ci-suite-manifest.sh: line 80: mapfile: command not found
```

I hit that during #7741, which added eleven suites: the invariant had to be checked by hand and left to CI. **A safety net that cannot run where the mistake is made is doing half its job.**

## The conversions were not uniform

Four different substitutes, chosen per shape rather than applied by template:

- **`mapfile` → `while read`** (8 sites). Mechanical, but the empty-line decision is not: `clean-stale-building-labels.sh`'s `gh` call ends in `|| true`, so a failed lookup yields no output and `mapfile` would have left one blank entry. `total` is computed from the array length and drives an early-exit path, so a phantom entry would have reported "1 stale claim" on a failed query. Skipping empties there is load-bearing; in `run-ci-suites.sh` the `awk NF` test already handles it.

- **Set whose consumer sorts → indexed array + `sort -u`** (`extract-capability-markers.sh`). `seen_valid`'s only reader was `printf '%s\n' "${!seen_valid[@]}" | sort`, so appending everything and deduping on output is byte-identical and needs **no membership test at all** — one fewer moving part than either the associative form or a hand-rolled lookup.

- **Maps with values → parallel indexed arrays** (`check-onidle-status.sh`). These carry a fire count and a last-fired timestamp per role, so they cannot collapse. Index-parallel is *exact* here because both the producer and reporting loops iterate `"${ON_IDLE_ROLES[@]}"` literally in the same order. That coupling is the fragile part, so it is stated in a comment and both loops increment `idx` as their last statement.

- **Set with a membership test → delimited string + `case`** (`verify-proposal-refs.sh`). Anchored on both newlines so `doc` never matches inside `docs`, and it spawns no subprocess per lookup unlike `grep -qxF`. Its **literal** word→number table became a `case` function, which for a fixed table reads better than either array form and preserves the "unknown word yields empty" contract the caller depends on.

## The second-order trap

Converting reads is not enough. Under `set -u`, **bash 3.2 treats an empty indexed array's `"${arr[@]}"` as an unbound variable and aborts** — fixed upstream in 4.4, so the bare form works everywhere except the interpreter this fix targets. All six scripts use `set -u`.

Without guards this would have traded `mapfile: command not found` for a hard abort, which is the exact trap #7717's own description flagged (*"then an empty-array expansion under `set -u` … The first masked the second"*) and #7766 documented. Six expansions are guarded with `${arr[@]+"${arr[@]}"}` where the array can legitimately be empty.

## Verified both directions

Per the pattern this repo keeps paying for — five instances this session of a gate reporting success while doing nothing (#7735, #7772, #7743, #7761, #7779) — I checked that each gate still **fails** when it should, not only that it passes:

```
clean tree:       OK: 241 suites (233 wired, 8 excluded)   exit 0
unlisted suite:   ERROR: suite 'test-zz-probe.sh' is in NEITHER ...   exit 1
```

Same for `check-onidle-status.sh`: exit 3 with a dormant role present, exit 0 once every configured role is verified.

## Gates

`check-file-size-budget`, `check-shell-syntax --dir defaults` (bash 3.2), and `check-ci-suite-manifest` all pass. None of the six files is in the size ledger, and all shrank or held.

## What this does not cover

- **`bash -n` will not catch this class.** `mapfile` and `declare -A` are *runtime* failures, not syntax errors, so #7735's parse gate (which correctly covers the #7721 class) is complementary rather than redundant. #7728's structural point stands: no CI runner exercises bash 3.2 *behaviour*, only parsing.
- `.loom/` mirrors still contain the old copies until the next resync commit — that is the designed propagation path (#5874).
- **#7794 gained an eighth instance** while I was here: `check-onidle-status.sh:55` also recovers its `--help` text from `$0` (`sed -n '2,/^$/p' "$0"`). Not touched by this PR; the issue listed seven.

## Note

A local run exercises bash 3.2 only — this machine's PATH `bash` **is** `/bin/bash` 3.2.57. bash 5 coverage comes from CI, and #7735's matrix now runs both legs.


turian:

I authored this. Mechanical verification only; the conversion *choices* (which substitute per shape) are judgment calls a second reader should sanity-check.

**27 checks pass, 4 skipped, zero failures** — including `Shell Syntax (macos-latest)`, the bash-3.2 leg #7735 added, which is the one that actually exercises what this PR is about.

### What is verifiable

- **Repo-wide count of bash 4 builtins in tracked shell code is now zero** (outside the `.loom/` resync mirrors).
- Each of the six was run on stock macOS `/bin/bash 3.2.57`, and the two with wired suites pass them: `extract-capability-markers` 29/29, `verify-proposal-refs` 24/24.
- Gates that must still *fail* were checked failing, not just passing: `check-ci-suite-manifest.sh` exits 1 naming an unlisted suite; `check-onidle-status.sh` exits 3 with a dormant role and 0 without.

### The part worth a reviewer's attention

Four different substitutes were used, chosen per shape rather than templated — a set whose consumer sorts became `sort -u` with **no membership test at all**; maps carrying values became index-parallel arrays; a set with a membership test became a delimited string plus `case`; a literal table became a `case` function. The reasoning for each is in the commit messages.

The riskiest is `check-onidle-status.sh`'s index-parallel pair, because it introduces a coupling that did not exist before: the producer loop and the reporting loop must iterate `ON_IDLE_ROLES` identically. Both do, literally, and both increment `idx` as their last statement — but that is an invariant a future edit could break silently, so it is stated in a comment. A fixture confirmed counts *and* timestamps stay correctly associated per role, which is the thing that would go wrong if the coupling were off by one.

### Second-order trap, for the record

Converting the reads alone would have been worse than leaving them. Under `set -u`, bash 3.2 treats an empty indexed array's `"${arr[@]}"` as unbound and **aborts** — fixed upstream in 4.4. All six scripts use `set -u`, so six expansions are guarded with `${arr[@]+"${arr[@]}"}`. Without those this trades `mapfile: command not found` for a hard abort, exactly as #7717's description warned (*"the first masked the second"*).

### What it does not fix

`bash -n` does not catch this class — `mapfile` and `declare -A` are runtime failures, not syntax errors — so #7735's parse gate stays complementary, and #7728's point that no runner exercises 3.2 *behaviour* still stands.

Issue #7795, opened 2026-09-16
Guard ASK tier: which sites steer toward a safe alternative, and which are a bare 'are you sure?' that stalls headless runs


turian:
## Original Issue

**Position: do not build the shell tokenizer proposed in #7760, and stop adding masking shapes to the guard layer. Turn on the measurement that already exists, and size the guard to what the data justifies.**

This is a counter-proposal to #7760, not a request to delete the guards. Filing it as an issue because the reasoning should be on the record and challengeable, rather than living in a session.

## What the guard layer costs, measured

**Size.** 13,725 lines of bash across 9 hooks, of which `guard-destructive-generic.sh` is **8,976** — the largest non-daemon file in the repository.

**Latency.** ~23ms per invocation, measured over 20 runs on a trivial `echo` payload. Two guards are wired to `PreToolUse`/`Bash`, so roughly **46ms on every Bash tool call**, before the command runs.

**Churn.** Over 12 months:

| | commits |
|---|---|
| all `defaults/hooks/` | 173 |
| `guard-destructive-generic.sh` alone | 101 |
| of those, matching the masking / quoting / precision / false-positive class | **49** |

Roughly half the maintenance on the largest guard is the guard mis-reading its own input.

**Friction, as filed issues.** #6076, #6123, #7318, #7356, #7440, #7515, #7530, #7773 — eight on the record, several closed and recurred. #7773 is today's: the guard denied *writing a file* that quoted a command, three times, including once while filing the issue reporting that it does this.

## What the guard layer is worth, measured

Nothing. There is no measurement.

#3771 added decision telemetry specifically so that *"false-positive friction becomes measurable (which patterns fire, how often, before/after a precision fix)"*. It writes to `.loom/logs/guard-decisions.log`.

That file **does not exist** on this host. The feature is **off by default** — the source says so — and `.loom/config.json` carries no `guards` block to enable it.

So the position we are actually in: a 9,000-line program on the hot path of every tool call, with 49 self-inflicted defects in a year and eight filed friction reports, and **zero recorded instances of it preventing anything**. Not "a poor ratio" — an undefined one.

## Why a tokenizer is the wrong next move

#7760 is correct that regex masking is the wrong instrument, and correct that ADR-0016's "no external dependency" argument is circular. I agreed with that this morning and argued for it.

I no longer think it follows that we should build the tokenizer, for three reasons:

[...]

Issue #7773, opened 2026-09-15
guard-loom-workflow.sh denies writing a FILE that merely quotes the merge command, and the obvious fix opens a bypass


turian:
`guard-loom-workflow.sh`'s `loom:gh-pr-merge-redirect` check blocks a command that **writes the phrase into a file** and never executes it.

## Repro

```bash
cat > /tmp/issue-body.md <<'DELIM'
Some prose that quotes the disallowed command as documentation.
DELIM
gh issue edit 1234 --body-file /tmp/issue-body.md
```

If the heredoc body contains the phrase, the whole call is denied:

> Use ./.loom/scripts/merge-pr.sh <PR_NUMBER> instead of 'gh pr merge'.

Hit three times today: twice while editing issue #7754, whose entire subject *is* the repo's merge strategy and whose text therefore necessarily quotes the command — and once more **while filing this very issue**, whose repro section quotes it too. The guard denied the report about the guard.

## Why the existing masking does not cover it

The guard already anticipates this class. Its own comment describes firing on "a git commit message... that quoted the phrase as prose documenting the very rule this guard enforces". Masking is deliberately narrow, gated on two provably-inert shapes:

1. `cat` with a **quoted** delimiter whose output is captured into a known text-data flag — the `git commit -m "$(cat <<'EOF' ... EOF)"` idiom
2. `git commit -F -` / `--file=-`

The shape above is `cat > FILE <<'DELIM'`: quoted delimiter, output redirected to a **file**, not captured into a flag and not piped. Inert by the same reasoning, but not on the list.

## The naive fix is not safe, which is why this is a report rather than a patch

Adding "`cat` redirected to a file" to the whitelist opens a single-command bypass:

```bash
cat > /tmp/x.sh <<'DELIM'
<the disallowed command> 123
DELIM
bash /tmp/x.sh
```

Today the phrase is visible in the heredoc and the command is denied. With `cat > FILE` masked, neither half trips the check — the body is masked, and `bash /tmp/x.sh` contains no phrase. That is a real widening of what this catastrophic-tier guard misses, against the guard's own stated invariant: *"Masking only ever narrows what this ONE check can see; it never widens what it misses."*

Options, none obviously correct:

- **(a) Mask `cat > FILE` only when no interpreter invocation appears elsewhere in the same command.** Preserves the invariant, but "interpreter invocation" is an open-ended set, and this becomes a second guard inside the guard.
[...]

Issue #7760, opened 2026-09-15
Design: replace regex masking with a real shell tokenizer (external dep) — ADR-0016's no-dependency argument is circular


turian:
Design proposal. #7755 files the cheap interim forcing function; **this is the actual fix for the same root cause**, and it needs a decision before anyone builds either.

## The circular constraint

ADR-0016 lists as a *positive* of its chosen design:

> no shell AST, **no new external dependency**, no growth in the hook's runtime cost profile

Both of those hold only because the guard **is a bash script**. A bash script cannot take a Rust dependency, and cannot reasonably carry a tokenizer. So "no external dependency" is not an independent engineering finding — it is a restatement of the implementation language, used to justify staying in that language.

Meanwhile the cost of staying is measurable: **52 of 99 commits** to `guard-destructive-generic.sh` in 12 months are code-vs-quoted-data confusion (#7755), there are **10 hand-written masking functions**, and the class has produced **reproduced silent ALLOWs** — ADR-0016's own words: *"it silently ALLOWS a write that should deny."*

The question the ADR never asks is: what if the guard were not bash?

## What changes if it is a binary

**A real tokenizer replaces all ten masking functions.** The question every masking pass approximates — *"is this text inside a string literal, a comment, or a heredoc body?"* — is answered exactly by a lexer, not estimated by regex. Maintained Rust options:

| crate | status | notes |
|---|---|---|
| [`brush-parser`](https://crates.io/crates/brush-parser) | actively maintained, 2026 | tokenizer + POSIX/bash parsers; powers [`brush-shell`](https://crates.io/crates/brush-shell), a bash-compatible shell in Rust |
| [`yash-syntax`](https://crates.io/crates/yash-syntax) | updated Nov 2025 | parse-only by design (no execution) — arguably the better fit, since the guard must never execute anything |
| [`conch-parser`](https://github.com/ipetkov/conch-parser) | last update 2022 | likely unmaintained; listed for completeness |

`yash-syntax` being explicitly parse-without-execute is a meaningful safety property for this use.

**Runtime cost goes down, not up.** The ADR's "no growth in runtime cost" assumes bash is the cheap baseline. It is not: `PreToolUse` fires on **every** tool call, and today that means bash parsing and executing an **8,921-line** script each time. A binary exec is faster. This argument currently points the opposite way from how the ADR uses it.

**The repo already ships a Rust binary and has porting precedent** — #4552, #4471, #4105, #4108.

## The real blocker is ownership, not technology

[...]

Issue #6684, opened 2026-08-21
Guard a bare 'cargo clean' when build.target-dir is shared outside the repo — it deletes every concurrent sweep's build output with an error that names nothing


rjwalters:
## Context

On a host whose `~/.cargo/config.toml` sets a shared `build.target-dir`, a bare `cargo clean` is not a local operation — it deletes the build output of **every project on that host**, including whatever sweeps are compiling at that moment.

Hit on robb-studio 2026-08-21. Its config routes all build output to one directory on an external volume (moved there after the internal disk hit ENOSPC and took Docker down):

```toml
[build]
rustc-wrapper = "sccache"
target-dir = "/Volumes/Stripe/cargo-target"
```

A benchmarking script ran `cargo clean` between timed builds — correct in isolation, destructive here. Concurrent sweep builds on the same host started failing with:

```
error: error writing dependencies to `/Volumes/Stripe/cargo-target/debug/deps/zeroize-*.d`:
       No such file or directory (os error 2)
error: could not compile `zeroize` (lib) due to 1 previous error
```

repeated across a dozen crates. Nothing in that message points at the cause. The failing agent sees a mid-build disappearance of its own target directory, with no indication that another process on the host removed it, and would reasonably conclude the volume had faulted (a live hazard on this host — the same external enclosure has wedged before, presenting as unkillable `open()` in cargo).

The config file's own comment already warns about it — "one shared target dir across all projects, so … a bare `cargo clean` clears every project's artifacts (use `cargo clean -p <pkg>` to scope it)" — but a comment in a dotfile does not reach an agent that is about to run the command. The existing destructive-write guards are exactly the right place: this is the same class of hazard they already catch (a command whose blast radius is wider than the worktree it is issued from), and cheap to detect, since `cargo config get build.target-dir` answers it directly.

Worth noting the guard would have fired correctly here even though nothing was permanently lost: build output is derived state, so the cost was recompilation, not data. But it was recompilation charged to *unrelated in-flight sweeps* that had no way to attribute it — and the same command in a repo whose `target/` is local is entirely safe. That asymmetry is what makes it a guard case rather than a blanket ban.

## Suggested acceptance criteria

- [ ] A guard hook (the `guard-destructive-*` family) asks before a bare `cargo clean` when `build.target-dir` resolves **outside** the current repo — i.e. the delete would reach other projects
[...]

PR #7799, opened 2026-09-16
config(guards): disable the SQL DDL category (no production DB here), enable decision telemetry

### LINKED ISSUE #7797: guards config: disable the SQL DDL category (no production DB here) and enable decision telemetry for the standing review
Two changes to `.loom/config.json`'s `guards` block for **this repo only**, each argued from evidence rather than from general friction. Deliberately narrow — see "What this does NOT touch", which is most of the guard surface.

## 1. `guards.sqlDdl: false`

The SQL DDL/DML guard assumes SQL statements target a precious production database. **This repo has no production database and no SQL CLI workflow.** Its only SQL is embedded `rusqlite` schema inside Rust — `loom-daemon/src/activity/schema.rs`, `metrics_collector.rs`, `tuning.rs` — backing local activity and metrics stores that are regenerable by construction.

What the guard actually does here is block agents from *reading their own schema code*. Observed live while researching this issue: a `grep -rlE` over the repo, whose **search pattern** contained a DDL keyword, was denied outright. Nothing was going to be executed; the string was an argument to `grep`.

That is the same code-vs-quoted-data confusion class measured in #7760 (49 of 101 commits to this guard in 12 months), arriving in a category that protects nothing here.

#3552 already established the principle — *"a category error for any project that is itself a database engine… the guard assumes SQL statements target a precious production DB"* — and added the opt-out. The same reasoning applies to a project that merely *embeds* a database: the statements are the product's own vocabulary, and the data is local and regenerable.

## 2. `guards.decisionLog: true`

Enable the telemetry that the standing Guard-Decision Telemetry Review policy (#3898) depends on.

That policy has produced real, evidence-based reviews — #5754 (226 entries) and #6785 (535 events) — each ranking triggers and issuing per-pattern verdicts. But the log is **off by default**, so it only has data when someone remembers to enable it for a tick. Leaving it on makes the next review continuous rather than sampled, and it is the only way the remaining ask-tier questions in #7795 get settled by data instead of by argument.

**Safe to persist:** the record is `{ts, decision, pattern, tier, command}` with `command` redacted through `strip_literal_text()` — the same redactor the pattern tiers use — specifically so that no raw `--body`/`-m`/`--title`/`--notes`/`--comment` value reaches a log that aggregates across sessions. It also deliberately does not log `allow`, preserving the #3687 read-only fast path's zero-overhead silent allow.

## What this does NOT touch, and why

These were considered and **rejected** for disabling, because each has a real incident or a live hazard behind it:

| category | why it stays |
|---|---|
| `guards.stashScope` | **#4821** — parallel builders raced on each other's stashes. `refs/stash` is one stack shared across every linked worktree; **11 are active on this host right now**. #5754 reviewed the telemetry and ruled *"keep flagged — do not weaken this guard."* It also steers toward a genuinely safe alternative (`worktree.sh snapshot` / `stash-push`), which is the guard doing real work rather than gating. |
| `guards.cargoCleanScope` | **#6684** created it for a shared `build.target-dir`, where a bare `cargo clean` clears every project on the host **including in-flight sweeps**. Same multi-agent hazard shape as stash-scope. |
| `guards.forceScope` | `force-op:protected` covers force operations against a protected branch. The established remedy for its friction is **narrowing**, not disabling — see #7530, fixed by extending the safe-list to a worktree's own branch rather than switching the category off. |
| `guards.cloudCli` | Already narrowed twice (#3593 for over-broad read-only prompts, #6125). The residual friction (#7440, `docker rmi` on the Auditor's own image) is narrow enough to deserve the same narrowing treatment rather than a category flip. |
| the ungated denial floor | Not configurable by design (#4791), and correctly so. |

## Acceptance

```bash
# a grep whose PATTERN contains SQL DDL is no longer denied
grep -rn "<ddl-keyword>" loom-daemon/src/activity/    # runs

# decisions are recorded
test -f .loom/logs/guard-decisions.log

# the floor is unaffected
# (a catastrophic-tier command is still denied with guards.sqlDdl=false)
```

Plus: `.loom/config.json` remains valid JSON, and the `guards` block round-trips through the config resolver.

## Related

- #7795 — the broader ask-tier question this deliberately does not pre-judge
- #3898 / #5754 / #6785 — the telemetry policy and its review ticks
- #3552 — the precedent for the SQL opt-out
- #4791 — the ungated denial floor, unaffected

turian:
Closes #7797

Two narrow changes to this repo's `guards` block. **Deliberately does not touch `stashScope`, `cargoCleanScope`, `forceScope` or `cloudCli`** — each has a real incident or live hazard behind it, listed below.

## `guards.sqlDdl: false`

The SQL DDL/DML category assumes SQL statements target a precious production database. **This repo has none.** Its only SQL is embedded `rusqlite` schema inside Rust — `loom-daemon/src/activity/schema.rs`, `metrics_collector.rs`, `tuning.rs` — backing local, regenerable activity and metrics stores.

What the category actually does here is block agents from **reading their own schema**. Observed live while researching the issue: a `grep -rlE` whose *search pattern* contained a DDL keyword was denied outright, with nothing to execute.

That is the code-vs-quoted-data confusion class #7760 measured — 49 of 101 commits to this guard in 12 months — landing in a category that protects nothing here.

#3552 set the precedent and added the opt-out, reasoning the guard is *"a category error for any project that is itself a database engine"*. The same holds for a project that merely **embeds** one.

## `guards.decisionLog: true`

Turns on the telemetry the standing **Guard-Decision Telemetry Review policy (#3898)** depends on.

That policy has produced real, evidence-based reviews — **#5754** (226 entries) and **#6785** (535 events), each ranking triggers and issuing per-pattern verdicts. But the log is off by default, so it only has data when someone remembers to enable it for a tick. Leaving it on makes the next review continuous rather than sampled, and it is how the open questions in #7795 get settled by data rather than argument.

**Safe to persist.** Records are `{ts, decision, pattern, tier, command}`, with `command` redacted through `strip_literal_text()` — the same redactor the pattern tiers use — specifically so no raw `--body`/`-m`/`--title`/`--notes`/`--comment` value reaches a log that aggregates across sessions. `allow` is deliberately never logged, preserving #3687's zero-overhead silent fast path.

## Verification, against the guard actually reading this config

| input | result |
|---|---|
| `grep` with a DDL keyword in its **pattern** | **ALLOW** (was DENY) |
| a repository-deletion command | **DENY** — floor intact |
| decision log | **written**, structured JSONL records |

The ungated denial floor (#4791) is *"not read through the config resolver at all"*, so neither key can weaken it — confirmed by the second row running with `sqlDdl` already false.

All five repo checks pass. No `defaults/` files touched, so no VERSION bump is required.

## What stays, and why

| category | why it stays |
|---|---|
| `stashScope` | **#4821** — parallel builders raced on each other's stashes. `refs/stash` is one stack shared across every linked worktree; **11 are active on this host**. #5754 reviewed the telemetry and ruled *"keep flagged — do not weaken."* It also steers toward a safe alternative (`worktree.sh snapshot`), which is the guard doing real work. |
| `cargoCleanScope` | **#6684** — a bare `cargo clean` against a shared `build.target-dir` clears every project on the host, **including in-flight sweeps**. Same multi-agent hazard shape. |
| `forceScope` | The established remedy for its friction is **narrowing**, not disabling — #7530 was fixed by extending the safe-list, not by flipping the category. |
| `cloudCli` | Already narrowed twice (#3593, #6125). Residual friction (#7440) deserves the same treatment. |

## A note the reviewer may enjoy

While testing and committing this, the guard blocked **four** of my own inputs for containing destructive strings as **data**: a grep pattern, two literals inside a JSON test fixture, and finally the commit message itself — which documents the test result above and had to be supplied via `git commit -F`.

Four instances of the code-vs-data class, encountered while verifying a change about that class. Offered as evidence for #7760 and #7773 rather than as a complaint.


turian:
## Judge: approved

+4/-0 in `.loom/config.json`. 17 pass, 10 skipped, zero failures, `CLEAN`.

### The safety question, checked rather than assumed

Disabling a guard category is security-adjacent, so I verified it cannot drop below the floor. `defaults/docs/guard-hooks.md` is explicit:

> **The toggleable deny categories**: SQL DDL/DML (`guards.sqlDdl`), the cloud/docker ask category […] Promoting any of them into the floor would break those repos and buy nothing — **none of them is unrecoverable the way the floor** is.

and, on the floor itself:

> No config, no env var, turns any of these off. The floor is not read through the config […] the toggles can only widen or narrow the **ask** tier.

So `sqlDdl: false` narrows a tier that sits *above* an untouchable floor. That is precisely what the toggle exists for, and the doc's own rationale says promoting it into the floor would be wrong.

### The premise holds

This repo has no production database. Its only SQL is embedded `rusqlite` schema in `activity/schema.rs`, `metrics_collector.rs` and `tuning.rs`, backing local regenerable activity/metrics stores. What the category actually does here is block agents from **reading their own schema** — a cost with no corresponding risk.

### Scope discipline is the best part of this PR

It deliberately leaves `stashScope`, `cargoCleanScope`, `forceScope` and `cloudCli` alone, each with a stated incident or live hazard behind it. A config PR that turns off *one* category it can justify, and explicitly names the four it will not touch, is the right shape — the tempting version of this change is "turn off the noisy ones," and it did not do that.

`decisionLog: true` is additive: it turns on the JSONL decision record, which makes future guard-tuning arguments empirical rather than anecdotal. Worth having before any further toggle changes.

PR #7800, opened 2026-09-16
refactor(daemon): extract inline test modules — second batch, 7 files (#7740)

### LINKED ISSUE #7740: Extract inline test modules: second batch (files not touched by #7723)
Second batch of inline `#[cfg(test)]` extraction, covering ledger entries that **PR #7723 does not touch** — so the two can run in parallel without conflicting.

#7718 deliberately scoped to the eight largest files. These are the next tier.

| file | lines |
|---|---|
| `loom-daemon/src/worktree_ops/clean.rs` | 4,082 |
| `loom-daemon/src/cli/status_render.rs` | 3,489 |
| `loom-daemon/src/sweep_registry/reaper.rs` | 3,408 |
| `loom-daemon/src/sweep_registry/watchdog.rs` | 3,079 |
| `loom-daemon/src/init/scaffolding.rs` | 2,944 |
| `loom-daemon/src/auto_update.rs` | 2,782 |
| `loom-daemon/src/fleet/add_worker.rs` | 2,604 |
| `loom-daemon/src/init/mod.rs` | 2,408 |

(Sizes are whole-file code lines per #7711's counting. Confirm against the ledger at implementation time — they move.)

## Method — follow #7718 exactly, including the bug it avoids

**Move the module body verbatim. Let `cargo fmt` re-indent. Do not hand-dedent.**

#7718's first attempt stripped four spaces per line. That is a *lexical* edit on a *syntactic* structure: it cannot distinguish code from the interior of a raw string literal, and it silently rewrote embedded shell and JSON fixtures in 4 of 8 files. It was invisible in the diff, to the compiler, and to the full test suite — the corrupted fixtures happened to be whitespace-insensitive, so all 5,123 tests still passed.

`.loom/docs/file-size-policy.md` records this as the standing rule for mechanical refactors.

## Acceptance criteria

- [ ] Each test module in `<module>/tests.rs`; parent carries `#[cfg(test)] mod tests;`.
- [ ] **Every raw-string literal byte-identical before and after**, verified by a tool that lexes the language — not by reading the diff. This is the acceptance test that actually catches the failure mode.
- [ ] `cargo fmt --check` clean.
- [ ] Test list unchanged (same count, same names) before and after.
- [ ] One commit per file.
- [ ] Files with test-only helper **functions** outside the test module: leave them in the parent; the child reaches them via `use super::*`.

## Sequencing

Independent of #7723 by construction — disjoint file sets. But both are large mechanical diffs, so **do not run them concurrently through review**; land #7723 first, then rebase this and refresh the ledger with `--update`.

Note on macOS: `cargo test --lib` currently fails on one unrelated pre-existing test (#7730, bash 3.2). Expect 5,123 passed / 1 failed locally; that failure is not yours.

turian:
Closes #7740. Second batch after #7718.

## What

Moves each file's top-level `#[cfg(test)] mod tests { ... }` to a sibling file, leaving `#[cfg(test)] mod tests;` behind.

| module | before | parent after | child | parent |
|---|---|---|---|---|
| `init/mod.rs` | 3,531 | **865** | 1,916 | **−75%** |
| `init/scaffolding.rs` | 4,432 | 1,296 | 2,299 | −70% |
| `sweep_registry/reaper.rs` | 5,468 | 2,138 | 2,437 | −60% |
| `sweep_registry/watchdog.rs` | 5,012 | 2,343 | 1,836 | −53% |
| `fleet/add_worker.rs` | 3,563 | 1,689 | 1,488 | −52% |
| `auto_update.rs` | 4,040 | 2,197 | 1,457 | −45% |
| `worktree_ops/clean.rs` | 5,713 | 3,523 | 1,702 | −38% |

`init/mod.rs` drops **off the size ledger entirely** — 865 lines is below the 1000-line threshold.

## Method

**Module bodies moved verbatim; `cargo fmt` does the re-indentation.** Not a hand-written dedent — that is a lexical edit on a syntactic structure and silently rewrites the interior of raw string literals, which is how #7718's first attempt corrupted embedded shell and JSON fixtures in 4 of 8 files while all 5,123 tests still passed. `.loom/docs/file-size-policy.md` records this as the standing rule.

## Verification

**Every raw-string literal byte-identical before and after**, across all seven — 114 literals total, checked with a scanner that tracks comments and normal strings. This is the acceptance test that catches the actual failure mode; reading a 22k-line diff does not.

```
IDENTICAL worktree_ops/clean.rs          5    IDENTICAL auto_update.rs        10
IDENTICAL sweep_registry/reaper.rs       5    IDENTICAL fleet/add_worker.rs   16
IDENTICAL sweep_registry/watchdog.rs    14    IDENTICAL init/mod.rs           31
IDENTICAL init/scaffolding.rs           33
```

Also:
- `cargo fmt --check` clean.
- Test list identical: **5,164** before and after.
- Full suite: **5,163 passed, 1 failed** — `init::repo_owned::tests::rust_and_shell_resync_ignore_parsers_agree_on_shared_fixture`, the known **pre-existing macOS-only** failure (#7730: a Rust-embedded shell driver uses bash-4 `declare -A`; stock macOS bash 3.2 rejects it). Untouched by this PR and by #7798, which cleared the *script* instances but not that embedded one. CI on bash 5 is unaffected.
- `check-file-size-budget` passes with the ledger updated.

## Scope: 7 files, not the 8 the issue listed

`cli/status_render.rs` is **excluded**. It has **14 separate top-level `#[cfg(test)]` modules**, not one — a materially different transformation from the single-module shape this method handles, with 14 module names to preserve and more ways to go wrong silently. Extracting it deserves its own change rather than being folded in here. Its parent is 5,322 lines, so it remains a worthwhile target.

## Two bugs found in my own tooling, both caught by tools rather than by me

**1. `mod.rs` resolves children differently.** For `init/mod.rs`, Rust resolves `mod tests;` to `init/tests.rs`, **not** `init/mod/tests.rs`. My extractor wrote the latter. `cargo fmt` refused immediately with *"failed to resolve mod `tests`"* — so the compiler-adjacent tool caught it before any test ran. Fixed by hand for this PR; the extractor is a scratch script, not shipped.

**2. My verifier had the same bug**, looking for the literal set in `init/mod/tests.rs` and erroring out. Corrected, and `init/mod.rs` verifies IDENTICAL with 31 literals.

Consistent oversight in both directions — worth noting since the same blind spot produced both.

## A gap in #7711's ratchet, found here

The gate measures `git ls-files`, so **it is blind to untracked files**. Before staging it reported:

```
check-file-size-budget: OK — 7 file(s) shrank; run --update to tighten
```

After `git add`, the same tree reported seven new files newly crossing the threshold. A new oversized file therefore passes the gate right up until it is staged. Not a problem for this PR — the files are committed and the ledger is updated — but the gate should probably consider the working tree, or at least say that it only measures tracked files. Worth a follow-up on #7711.

## Ledger effect

```
ledger files  83 -> 87
ledger total  160501 -> 158335   (-2166)
```

The total **drops**, which is what #7711's policy asks a legitimate split to produce. The seven new entries are the correct signal rather than a regression: a 2,437-line `reaper/tests.rs` is still a big file, and the gate now says so where it was previously hidden inside a 5,468-line parent.

## Scope discipline

Byte-preserving moves only. No test renames, no splitting test modules by theme, no lint fixes, no production changes. One commit per module so each move can be verified independently.


rjwalters:
✅ **Approved!**

Reviewed at head `5e4dd86118b85c03fc612f50c3badccb29689fd0` (8 commits: one per module + ledger). Byte-preserving test-module extraction for 7 files; no production logic change.

**What I verified (all measured on a scratch clone of the branch):**

- **Raw-string literals byte-identical, in order, in all 7 files** — independent lexer (tracks `//`/`/* */` comments, normal strings, char literals, `r#"…"#`): auto_update 10/10, add_worker 46/46, init/mod 31/31, scaffolding 33/33, reaper 5/5, watchdog 14/14, clean 5/5. (My scanner counts 144 rather than the PR's 114 because it also picks up `br"…"`/nested forms; the invariant — every literal in `origin/main:<parent>` appears unchanged in `HEAD:<parent>` + `HEAD:<child>`, same order — holds for every file.)
- **Normal string literals identical** in all 7 (409/638/1212/1069/538/516/1234).
- **Test function names identical**: 74+77+53+78+63+90+94 = 529 `#[test]`/`#[tokio::test]` fns, same set before/after.
- **Code equivalent modulo rustfmt**: whitespace-stripped token comparison of old parent vs new parent with `mod tests;` re-inlined is identical except for exactly what `cargo fmt` produces on dedent — 3 dropped trailing commas on collapsed calls (`auto_update`, `init/mod`, `watchdog`) and 2 unwrapped single-expression closure bodies (`watchdog.rs:3573`/`:3596` → `watchdog/tests.rs:1216`/`:1237`, `|l| { a && b }` → `|l| a && b`). No semantic change.
- **Parent diffs**: each of the 7 parents is exactly `-<module body>` / `+mod tests;`; `#[cfg(test)]` retained. `init/mod.rs` → `init/tests.rs` is the correct resolution for a `mod.rs` parent.
- `cargo fmt --check` → clean.
- `cargo test --lib` (loom-daemon) → **5164 passed; 0 failed** in 209.5s. Matches the PR's 5,164 test-list count.
- `bash scripts/check-file-size-budget.sh` on the branch → `OK — no over-threshold file grew`; ledger diff drops `init/mod.rs`, `init/scaffolding.rs`, `sweep_registry/reaper.rs` (each now under the 1000-code-line threshold) and adds the 7 children.
- CI: all 23 checks pass; `mergeStateStatus=CLEAN`; `check-review-feedback.sh` → `CLEAR` (0 reviews, 0 inline threads).
- Branch is 2 commits behind main (#7798, #7799); neither touches any file in this PR (`git diff --name-only HEAD...origin/main` ∩ PR files = ∅), and GitHub reports CLEAN, so no rebase needed.

**Scope notes (non-blocking):**
- Excluding `cli/status_render.rs` (14 separate `#[cfg(test)]` modules) is a reasonable scope call; it should get its own follow-up so #7740 does not silently lose that target.
- The observation that the size ratchet measures `git ls-files` and is blind to untracked files is worth a follow-up against #7711 (not re-derived here — reported by the builder).

No shell files touched, so shellcheck is not applicable. No VERSION bump in the diff.

turian:

I authored this. Mechanical verification only.

**23 checks pass, 4 skipped, zero failures, `CLEAN`.**

### What I can attest to

- **All 114 raw-string literals byte-identical** before and after, across all seven files, checked with a scanner that tracks comments and normal strings. This is the acceptance test that catches the real failure mode — reading a 22k-line diff does not.
- `cargo fmt --check` clean.
- Test list identical: **5,164** before and after.
- Full suite **5,163 passed / 1 failed** — the known pre-existing macOS-only failure (#7730), untouched here and unaffected on CI's bash 5.
- Ledger total **160,501 → 158,335 (−2,166)**, which is what #7711's policy asks a legitimate split to produce. `init/mod.rs` drops **off** the ledger entirely (3,531 → 865).

### Two bugs in my own tooling, both caught by tools rather than by me

1. **`mod.rs` resolves children differently** — Rust resolves `mod tests;` inside `init/mod.rs` to `init/tests.rs`, not `init/mod/tests.rs`. `cargo fmt` refused immediately with *"failed to resolve mod `tests`"*. The language tool caught it before any test ran, which is the argument for using it rather than hand-transforming.
2. **My verifier had the identical bug**, looking for literals in the wrong path. Same blind spot produced both.

### Scope: 7 files, not the 8 the issue listed

`cli/status_render.rs` is excluded — it has **14 separate top-level `#[cfg(test)]` modules**, a materially different transformation with 14 names to preserve. It deserves its own change rather than being folded in. Its parent is 5,322 lines, so it stays a worthwhile target.

### A gap in #7711's ratchet, found here

The gate measures `git ls-files`, so it is **blind to untracked files**. Pre-staging it reported *"OK — 7 file(s) shrank"*; post-`git add`, the same tree reported seven new files crossing the threshold. A new oversized file passes right up until it is staged. Not a problem for this PR — files are committed, ledger updated — but worth a follow-up on #7711.

### Relevance to #7810

This touches `auto_update.rs` (4,040 → 2,197), which is PR 5's territory in the shell-retirement epic. Landing it now avoids a rebase collision later.

Merging on the operator's instruction to land what I can stand behind.

Issue #7789, opened 2026-09-16
[tracking] 25 CI flake issues in 60 days, 22 closed: fix the mechanisms, not the instances


turian:
**Tracking issue.** Searching `flake OR flaky OR intermittent` over the last 60 days returns **25 issues, 22 already closed**. Each was diagnosed and fixed individually; the rate has not fallen. This issue categorises them by *mechanism* so the classes can be closed instead of the instances, and states honestly which ones the shell→Rust tiering (#7711) does and does not address.

## The record

| | closed | open |
|---|---|---|
| flake issues, last 60 days | 22 | 3 (#7736, #7748, #7788) |

Today alone produced three: `test-build-slot.sh`, `tokens_pool` (#7748), and `test-bounded-run.sh` (#7788) — none caused by the PR they failed on.

### A. SIGPIPE / broken pipe under `pipefail` — shell-only, 5 issues

- **#7060** — `test-guard-codex-bridge.sh`, printf broken pipe
- **#7285** — `test-install-stash-scope.sh`, "Broken pipe"; failed on `main` at a **docs-only** commit
- **#7540** — `test-sweep-lease-fence.sh`, explicit `printf | head` SIGPIPE race
- **#7736** — `verify-proposal-refs.sh`, `printf: write error: Broken pipe`
- **#7771** — the same mechanism as a *production* bug, not a flake: `pipefail` + `grep -qFx` SIGPIPE producing false `MISSING FILE` reports

One mechanism: `set -o pipefail` plus a consumer that exits early (`head`, `grep -q`) closes the pipe, the producer gets SIGPIPE, and the pipeline reports failure. **This class cannot exist in Rust** — there is no implicit pipeline status to propagate.

### B. Process / pid handshake races — shell, 2 issues

- **#7788** — the wedged child's pid file read immediately after spawn
- **#4547** — fake-gh subprocess tests failing under parallel load

Rust hands you a `Child` and its id directly; there is no file-based handshake to lose a race on. **Also killed by the port.**

### C. Time and wall-clock dependence — language-agnostic, not fixed by the port

- **#7748** — `tokens_pool`, explicitly *"wall-clock jump, not code"* — **this is a Rust test**
- **#3653** — a ~200ms perf-threshold assertion in shell
- `test-build-slot.sh` (unfiled, observed today) — staleness reclamation, time-dependent

### D. Concurrency / async — **already Rust**, 8 issues

[...]

Issue #7788, opened 2026-09-16
test-bounded-run.sh flakes in CI on unrelated branches — portable-mode fixture races on the wedged child's pid


turian:
`test-bounded-run.sh` fails intermittently in the "Shell Test Suites (hermetic)" CI job, on branches whose diffs have nothing to do with `bounded-run.sh`. It passes locally every time.

## Observed

Two failures within roughly 20 minutes, on two unrelated branches:

| run | branch | that branch's diff |
|---|---|---|
| 35034960799 | `feature/issue-7711` (PR #7714) | `ci.yml`, a file-size budget script, a baseline file |
| 35037798551 | `feature/issue-7780` (PR #7781) | Cargo workspace version inheritance |

Neither touches `bounded-run.sh`, its library, or anything it calls.

The failing assertions:

```
FAIL: [portable] a never-returning command is bounded and reports 124 (got '124' want '127')
FAIL: [portable] the wedged child recorded its pid (fixture sanity)
```

Both are in the **portable** mode block (the `timeout(1)`-free fallback path), not the native one.

## Why this looks like a fixture race rather than a real regression

**The expectation reads backwards.** The test's own name says the command "reports 124" — 124 is the conventional `timeout(1)` exit for a bounded kill — and the run *got* 124. It failed because it wanted **127**, which is "command not found". So either the assert arguments are ordered `(actual, expected)` where the message prints them as `got/want`, or the portable path legitimately produces different codes in different environments and the test encodes only one.

**The second failure is the tell.** *"the wedged child recorded its pid (fixture sanity)"* is a sanity assertion — the fixture spawns a deliberately wedged child and checks it recorded a pid. That failing means **the child never got far enough to write its pid file**, which is consistent with a slow or contended runner losing a race between "spawn the wedged child" and "read its pid", rather than with any behavior change in `bounded-run.sh`.

Under a loaded shared runner, a fixture that spawns a background process and immediately reads a file it writes is exactly the shape that flakes.

## Local result

On `main` and on both affected branches, under `/bin/bash` 3.2.57:

```
Total: 18  Passed: 18  Failed: 0
```

Every run, including the two branches CI failed on.

## Why it matters beyond one red check

[...]

Issue #6683, opened 2026-08-21
CI concurrency does not cancel a superseded QUEUED run — a force-pushed rebase left the new run pending with zero jobs for 20 minutes


rjwalters:
## Context

`.github/workflows/ci.yml` declares:

```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

The intent is that a new push to a branch supersedes the previous run. Observed 2026-08-21 on PR #6680: it does not hold when the superseded run is still **queued** rather than in progress.

Sequence:

1. Push `sccache-env-forward` → run `32504146683` starts. Most jobs complete; a couple of Rust jobs remain `queued` / `in_progress`.
2. Rebase onto `main` and force-push → run `32504737466` is created at 16:47:18Z.
3. The new run sits at `status=pending` with **zero jobs materialized**. The old run stays `status=queued`.
4. Twenty minutes later, nothing has moved. Cancelling the old run by hand (`POST actions/runs/32504146683/cancel`) immediately unblocks the new one, which then queues its jobs and finishes green.

So `cancel-in-progress` cancelled nothing, the old run held the concurrency group, and the new run was blocked behind it indefinitely. It was not a runner-capacity problem — all eight self-hosted slots were `online busy=false` throughout, and the run had not yet asked for any.

This is a quiet throughput tax rather than a visible failure. `gh pr checks <n>` reports `no checks reported on the 'sccache-env-forward' branch` while the run is `pending` with no jobs, so the PR shows neither pending checks nor failures — a Judge or an auto-merge gate polling check state sees an ambiguous nothing. Force-push after rebase is routine for Doctor and for anyone resolving conflicts, so this is reachable on any PR, not a one-off.

## Suggested acceptance criteria

- [ ] A superseded run is cancelled whether it is `in_progress` **or** `queued` — investigate whether the group key (`ci-${{ github.ref }}`) resolves differently for the two runs on a force-push, since a mismatched key would explain the group not superseding
- [ ] Verify against the real scenario: push, then force-push a rebase while the first run still has queued jobs, and confirm the first run reaches `cancelled` without manual intervention
- [ ] If GitHub genuinely will not cancel a queued run through `concurrency`, document the workaround in the workflow itself rather than leaving it as tribal knowledge — a cancel step keyed on older runs for the same ref is the usual shape
- [ ] Consider whether the same gap affects the other workflows that declare a `concurrency` group

## Related

[...]

PR #7803, opened 2026-09-16
ci: stop cancelling main's own verification (22 of last 30 main runs cancelled)

### LINKED ISSUE #7779: ci.yml cancels in-progress runs on main, so most main commits are never verified (21 of last 30 cancelled)

turian:
Closes #7779

Of the last 30 CI runs on `main`: **22 cancelled, 7 succeeded.** At ~3.3 merges/hour against a run longer than the gap between merges, near-total cancellation is the steady state — so most `main` commits are never verified.

## The issue blames the `concurrency:` stanza. That is the smaller half.

**1. The concurrency group.** `ci-${{ github.ref }}` is `refs/heads/main` for *every* push to main, so one shared group meant each merge superseded the previous merge's run. Now keyed on the PR number when there is one and the commit SHA otherwise, with `cancel-in-progress` restricted to `pull_request` events.

**2. The dominant cause — the `cancel-superseded` job (#6683).** It runs:

```bash
REF: ${{ github.head_ref || github.ref_name }}
gh run list --repo "$REPO" --workflow "$WORKFLOW" --branch "$REF" ...  | gh run cancel
```

On a push, `github.head_ref` is empty, so `REF` falls back to `github.ref_name` = **`main`**, and the job explicitly cancels every other queued/in-progress run on main. **Fixing the concurrency group alone would have left this in place and changed nothing.**

Its own rationale is entirely about a superseded PR push — *"a force-pushed rebase created a new run that sat pending"*. That reasoning does not transfer: on `main` there is nothing to supersede, because each commit is distinct work. Now gated on `if: github.event_name == 'pull_request'`.

**PR behaviour is unchanged** — a newer push still supersedes an older one via both mechanisms, which is what #6683 was built for.

## Why it matters beyond unverified merges

*"Does this also fail on `main`?"* is the first question in triaging a red check, and how a real regression is separated from a flake in one step. With 22 of 30 main runs cancelled, that question usually has no answer.

I hit this directly while triaging the `test-bounded-run` flake (#7788): checked `main`'s run, got no conclusion, and worked around it by comparing two PR branches to each other. That was cancellation, not a run in flight. #7789's flake categorisation leans on reading 25 issue titles for the same reason — the per-commit `main` signal is not reliably there.

## Cost check

This removes cancellation on `main`, so concurrent main runs become possible. Bounded by merge rate (~3.3/hour) rather than unbounded — each SHA runs once, and re-pushing the same SHA is not a thing. The heavy self-hosted pool already absorbs PR traffic at a considerably higher rate.

All five repo checks pass. Workflow parses as valid YAML (26 jobs). No `defaults/` files touched, so no VERSION bump.


rjwalters:
## Judge review: APPROVED

Reviewed head `9ede82ad` (1 commit, `.github/workflows/ci.yml` only, +22/-2). Evaluating this PR per explicit dispatch.

### Claims verified

- **Cancellation rate holds exactly.** `gh run list --workflow ci.yml --branch main --limit 30` right now: **22 cancelled, 7 success, 1 failure.** Matches the PR body.
- **Both cancellation paths are closed on `main`:**
  - `concurrency.group` is now `ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}`. On a push, `pull_request.number` is empty so the group is per-SHA; nothing else ever lands in it. `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` is belt-and-suspenders on top of that (expression form is supported by Actions; `actionlint` 1.7.12 is clean).
  - `cancel-superseded` is gated `if: github.event_name == 'pull_request'`. The analysis in the PR is correct: on a push `github.head_ref` is empty, `REF` fell back to `main`, and the job cancelled every other queued/in-progress main run. Fixing the stanza alone would have left this in place.
  - I also checked the only other workflow with a cancel job, `security.yml`: its `gh run list --workflow "$WORKFLOW"` is scoped to `Security Scan`, so it cannot touch CI runs. No remaining path cancels a main run short of a human doing it.
- **PR behaviour is unchanged.** Group `ci-CI-<number>` is equivalent to the old `ci-refs/pull/<number>/merge` scoping; `cancel-in-progress` stays true on PR events; `cancel-superseded` still runs with `REF=head_ref`. Fork PRs still have `pull_request.number`.
- **Sequencing precondition from #7779 is met**: #7768 (heavy jobs back on `ubuntu-latest`) and #7775 (`timeout-minutes: 30` everywhere) are both on main, so not-cancelling is affordable and a hung job cannot pin a slot.

### Cost

At ~3.3 merges/hour against ~28-minute runs, expect roughly 1.5 concurrent main runs steady-state, i.e. ~2.4 additional full runs/hour that were previously discarded. That is bounded by merge rate, per-SHA (no re-run amplification), and lands on GitHub-hosted runners only. Acceptable; it is the price of a usable `main` control.

### CI status on this PR: red check is main's own, not this PR's

`Shell Test Suites (hermetic)` fails on this PR (4 suites: `guard-destructive-core`, `-masking`, `-matching-precision`, `-readonly-fastpath`). I compared against main:

- This PR's merge base is `42119258` = main HEAD = #7799, the `.loom/config.json`-only commit that turned main red. Main's own run `35048461474` fails the **identical four suites** with the identical 229/4 summary.
- Tracked in #7808 with fix PR #7809 already in the review queue. A ci.yml-only diff cannot cause or cure a guard-hook assertion failure.
- Merge state is `UNSTABLE` (non-required check), not `BLOCKED`. Per the Judge role's merge-state table, UNSTABLE is "assess if acceptable"; a `loom:ci-failure` round here would hand Doctor nothing it can fix and just churn the PR until #7809 lands.

Approving on that basis. Note the irony: answering "does this fail on main too?" is exactly the control this PR exists to preserve, and it was answerable here only because #7799's main run happened to survive.

### Nit (non-blocking)

The body's "heavy self-hosted pool already absorbs PR traffic" is stale after #7768: `ci.yml` now has zero `self-hosted` runners. Harmless; the cost argument still holds on hosted runners.

No formal reviews or inline threads outstanding (`check-review-feedback.sh`: CLEAR).

Issue #7752, opened 2026-09-15
.loom/docs mixes symlinks and real files, so defaults/docs edits carry variable resync debt


turian:
`.loom/docs/` mixes two representations for the same kind of file: **12 symlinks into `defaults/docs/`, and 29 real files** — 28 of which have a `defaults/docs/` counterpart and so could be symlinks.

The symlink is the documented intent. From `scripts/check-docs-defaults-parity.sh`'s own header:

> A doc authored directly into `.loom/docs/` as a real file (instead of landing in `defaults/docs/` and **being installed as a symlink, the pattern the other entries in `.loom/docs/` already follow**) looks fine in-repo […] but it never ships.

That assumption is now the minority case: 12 of 41.

## Why it matters

A symlinked entry is **always current**. A real-file copy drifts from `defaults/docs/` until the next `chore: resync installed Loom surfaces` commit lands.

This is not theoretical — #7739 changed 12 docs under `defaults/docs/` and immediately produced six divergent `.loom/docs/` real files:

```
DIVERGED (real file, needs resync): .loom/docs/guard-hooks.md
DIVERGED (real file, needs resync): .loom/docs/guardrail-parity-codex.md
DIVERGED (real file, needs resync): .loom/docs/machine-dispatcher.md
DIVERGED (real file, needs resync): .loom/docs/safehouse.md
DIVERGED (real file, needs resync): .loom/docs/telemetry-schema.md
DIVERGED (real file, needs resync): .loom/docs/token-pool.md
```

The six symlinked docs in the same change needed nothing. So every `defaults/docs/` edit carries an invisible, variable-sized resync debt depending on which representation the target happens to use.

## Why the existing checker does not catch it

`check-docs-defaults-parity.sh` checks for **orphans** — a `.loom/docs/` real file with no `defaults/docs/` counterpart. All 28 of these *have* counterparts, so parity passes. There is no check for "real file that should be a symlink."

## Open question to settle first

Is the real-file form deliberate for some of these? In a **consumer** repo `.loom/docs/` must be real files (the installer copies them). In **this** repo, self-dogfooding, symlinks keep the two in lockstep. The 29 real files may simply be historical drift from before the symlink pattern, or they may be intentional for a reason not recorded.

**Answer that before converting anything** — the outcome may be "document the split and add a check that enforces whichever rule applies," not "convert 28 files."

## Acceptance criteria

- [ ] Determine whether the real-file form is intentional for any of the 28, and record the answer.
[...]

Issue #7745, opened 2026-09-15
resync-installed.sh exits 0 when its label-drift check crashes — a broken check is indistinguishable from a passing one


turian:
Observed live on a macOS host today. `resync-installed.sh` ran, printed a warning, and exited **0** — while one of its checks had not run at all:

```
WARN: Skipped forge label drift check (sync-labels.sh --check exited 1). Surface sync still applied.
    ℹ Target repository: rjwalters/loom (github)
    /private/tmp/loom/.loom/scripts/sync-labels.sh: line 534: local: -A: invalid option
    local: usage: local name[=value] ...
    /private/tmp/loom/.loom/scripts/sync-labels.sh: line 538: loom: unbound variable

[resync] 1 file(s) updated, 0 removed, 429 unchanged, 30 skipped.
```

That is not a forge being unreachable. That is the **checker itself crashing** (#7717 — `local -A` is bash 4.0+, and stock macOS `/bin/bash` is 3.2). The label drift check has therefore been silently not-running on every macOS resync since #7717 landed, and nothing said so beyond one warning line in a 430-line report.

## The mechanism

`resync-installed.sh` dispatches on `sync-labels.sh --check`'s exit code:

```
case "$check_rc" in
    0)  in sync ;;
    3)  drift -> apply the fix ;;
    *)  warn "Skipped forge label drift check ... Surface sync still applied." ;;
esac
```

The `*` catch-all is the problem. `sync-labels.sh` documents exit `1` as "forge/lookup error", and a bash crash **also** exits `1`, so "the forge is unreachable" and "this script is broken" are indistinguishable at the call site. The comment above the block states the intent explicitly — *"any forge/lookup failure (no git remote, gh missing/unauthenticated, misconfigured Gitea) is a loud warning, never a script-aborting failure"* — which is a reasonable policy for an **environmental** failure and the wrong one for a **defect**.

`resync-installed.sh` then **exits 0**, so no caller, script, or CI job can tell a check was skipped.

## Inconsistent with its own other checks

The same script hard-refuses for its two other gates:

```
665: err "Refusing to resync: one or more source shell scripts do not parse (bash -n)."
696: err "Refusing to resync: one or more source files carry live git conflict markers."
```

So within one script, one broken check aborts everything and another broken check is a warning that leaves the exit code clean. Whatever the right policy is, it should not depend on which check happened to break.

## This is the third instance of the same shape

| | Gate | Failure mode |
|---|---|---|
[...]

Issue #7791, opened 2026-09-16
Retry-once-and-record for the shared shell suite: one flaky assertion currently blocks every concurrent PR


turian:
Child 2 of #7789. **The highest-leverage item there, and the only one that does not wait on the shell→Rust tiering.**

## The problem is blast radius, not any single flake

`Shell Test Suites (hermetic)` runs 223 suites in one job. One flaky assertion anywhere fails the job, which fails the PR, which blocks a merge — for every concurrent PR, not just the one that happened to run.

**#7287** established this on 2026-09-06 and was closed: two failures in one hour, on different suites, one on a PR whose *"diff contains zero shell scripts"*.

It recurred today, twice in roughly 20 minutes, on two unrelated branches:

| PR | that branch's diff | failing suite |
|---|---|---|
| #7714 | `ci.yml`, a file-size budget script, a baseline | `test-bounded-run.sh` |
| #7781 | Cargo workspace version inheritance | `test-bounded-run.sh` |

Neither touches `bounded-run.sh`. The suite passes 18/18 locally on both branches and on `main`.

At ~20 PRs open and ~30 merges/day, a single flaky assertion is a fleet-wide tax.

## The second-order cost is worse

Retriggering has been normalised. There is a `chore: retrigger CI` commit sitting in a recent branch's history — pushed by a session other than the one that opened the PR.

Once red is routine, red stops carrying information. That is the same failure as #7745 and #7761, arriving from the opposite direction: a gate nobody believes is as useless as a gate that never fires. The repo has spent 22 closed issues chasing individual flakes precisely because each one had to be chased before anything could merge.

## Proposal: retry once, and *record* it

On failure, re-run the failing suite once. If it passes on retry, the job passes — **and emits a durable record** that suite X was retried on run Y.

The recording is not optional and is the whole design. A silent retry would convert a visible flake into an invisible one, which is exactly the fail-open pattern this repo keeps re-discovering (#7745, #7761, #7755, #7743). What is wanted is *"this merged, and here is the flake it stepped over"*, never *"this merged"*.

Concretely:

- Retry granularity is the **suite**, not the job — re-running 223 suites to clear one is waste, and it also re-rolls every other suite's dice.
- A suite that fails **twice** fails the job. No unbounded retries.
- Each retry appends to a machine-readable log (a job summary line, or an artifact) with suite name, run id, branch, and both exit codes.
[...]

PR #7805, opened 2026-09-16
docs: record a 'dumb reliable CI' principle in CLAUDE.md, paid for within the 320-line budget

### LINKED ISSUE #7804: Record a 'dumb reliable CI' principle in CLAUDE.md: three failures today came from individually-justified cleverness
**Operator directive:** CI in this repo should be *dumb and reliable*, not clever. Record it in `CLAUDE.md` (and therefore `AGENTS.md`, which is generated from it) so every agent inherits the constraint instead of re-deriving it.

## Why now — three failures in one day, all from cleverness

Each mechanism below was individually justified, defensible in review, and net-negative in aggregate.

**1. `cancel-superseded` (#6683) destroyed `main`'s verification.** Added to work around a real GitHub quirk: `cancel-in-progress` does not reliably cancel a run still `queued`, and a PR sat pending for ~20 minutes. The job cancels other runs on the same ref. On a push, `github.head_ref` is empty, so the ref resolves to `main` — and every merge cancelled the previous merge's run. **22 of the last 30 `main` runs cancelled, 7 succeeded** (#7779, PR #7803).

The second-order cost is worse than unverified merges: *"does this also fail on `main`?"* is the first question in triaging a red check, and for weeks it had no answer.

**2. Two cancellation mechanisms compounded invisibly.** The `concurrency:` stanza and the belt-and-suspenders job each looked correct alone. Nobody owned the interaction, and fixing only the stanza — the obvious culprit, and the one the issue named — would have changed nothing.

**3. Path-filtered job skipping hides failures from the PRs that cause them.** The `changes` job skips jobs by file group. Combined with #7789's flake load, a red check now means "something, somewhere, maybe yours".

Set against that, the `conflict-markers` job's comment states the opposite principle explicitly and correctly: *"must NOT be path-filtered — the corruption can land in any file, including one whose path group the PR did not otherwise touch."* The repo already knows this; it is just not written down as a rule.

## Proposed text

A short subsection in `CLAUDE.md` under Development Workflow, plus `.loom/docs/ci-principles.md` carrying the reasoning and today's evidence — following the file's own stated convention that *"new subsystem detail goes to `.loom/docs/` with a one-line pointer here, not inline."*

Draft rules:

1. **Prefer a slow correct job to a clever fast one.** Runner minutes are cheaper than an unverified merge or a misattributed failure.
2. **Never cancel verification of a distinct commit.** Superseding applies to a PR branch, where a newer push replaces an older one. Every `main` commit is distinct work.
3. **Path-filtering is an optimisation, not a correctness tool.** A check that can fail because of a file outside its path group must not be filtered.
4. **One mechanism per behaviour.** If two mechanisms implement cancellation, skipping or retry, their interaction is unowned and will surprise someone.
5. **A workaround for a platform quirk is scoped to the case that motivated it**, with the quirk named. #6683's rationale was a force-pushed PR; it was applied to all refs.

## Budget constraint

`CLAUDE.md` is at **320 lines against a 320 budget** — `check-claude-md-budget.sh` fails on line 321. Any addition must be paid for. The Resources "Docs" bullet is currently 8 lines of link list that compresses to 4 with no content loss, which funds the new subsection.

## Acceptance

```bash
./scripts/check-claude-md-budget.sh     # still OK, <= 320
./scripts/check-agents-md-sync.sh       # AGENTS.md regenerated in step
./scripts/check-dangling-links.sh       # the new doc link resolves
```

## Related

- #7779 / PR #7803 — the cancellation fix this generalises
- #7789 — flake tracking; "is the signal trustworthy" is the shared theme
- #7791 — retry-and-record, which must be *recorded* precisely to avoid becoming clever-and-silent

turian:
Closes #7804

Operator directive: CI here should be **dumb and reliable, not clever** — recorded in the operating core so every agent inherits the constraint instead of re-deriving it.

## Why now — three failures in one day, all individually justified

**1. `cancel-superseded` (#6683) destroyed `main`'s verification.** Added for a real GitHub quirk: `cancel-in-progress` does not reliably cancel a still-`queued` run, and a PR sat pending ~20 minutes. It cancels other runs on the same ref — and on a push `head_ref` is empty, so the ref is `main`. Every merge cancelled the previous merge's run. **22 of the last 30 main runs cancelled, 7 succeeded** (#7779, fixed in PR #7803).

**2. Two cancellation mechanisms compounded invisibly.** The issue blamed the `concurrency:` stanza — the obvious culprit. Fixing only that would have changed nothing, because the job was dominant. Nobody owned the interaction.

**3. Path-filtered job skipping obscures attribution.** A red check no longer reliably means "caused by this PR".

The repo already knows rule 3 — `conflict-markers`' own comment says it *"must NOT be path-filtered — the corruption can land in any file, including one whose path group the PR did not otherwise touch."* It just was not written down as a rule.

## What lands

Six rules in **`defaults/docs/ci-principles.md`** with the incidents behind each, and a **7-line summary** in `CLAUDE.md` under Development Workflow.

The rules, condensed: prefer a slow correct job to a clever fast one · never cancel verification of a distinct commit · path-filtering is an optimisation, not a correctness tool · one mechanism per behaviour · scope a platform workaround to the case that motivated it · a check that cannot run must not look like one that passed.

That last rule is the same one the guard and resync layers learned the hard way (#7745, #7761): `exit 0` has to mean *verified*, never *skipped*.

## Budget

`CLAUDE.md` was at **320/320**. The 7 added lines are paid for by compressing 7 elsewhere with **no content loss**:

| | before | after |
|---|---|---|
| Resources repo bullet | 2 | 1 |
| Resources docs list | 8 | 4 (now includes the new doc) |
| Migration History prose | 4 | 2 |

Still exactly 320. `check-claude-md-budget` passes.

## Two scope decisions worth reviewing

**The doc lives in `defaults/docs/` with a `.loom/docs/` symlink**, matching `troubleshooting.md` and `daemon-reference.md`. My first attempt authored it only in `.loom/docs/` and failed `check-docs-defaults-parity` — *"a doc authored only in `.loom/docs/` never ships"*. That is #7752's exact hazard, hit while writing about avoiding this class of thing.

**Not added to `defaults/.loom/CLAUDE.md`**, and therefore not to the generated `AGENTS.md`. That file is the *consumer template*; consumer repos have their own CI, and these rules cite this repo's incidents. The root `CLAUDE.md` is what agents working here read. Happy to mirror a generic version into the template if you'd rather it propagate.

All five repo checks pass, plus `check-doc-tocs`.


rjwalters:
## Judge verdict: APPROVED (`loom:pr`)

Reviewed head `7ea0e87ee07998ec881672a7c2ce8df48961e671` (single commit on top of the current `main` tip `42119258`; no rebase needed).

### What I verified

- **CLAUDE.md budget**: `./scripts/check-claude-md-budget.sh` on the PR tree → `OK — CLAUDE.md is 320 lines (budget 320)`. The 7 added lines are paid for by the Migration History (4→2) and Resources (8→4) compressions; I diffed the `.loom/docs/*` link set before/after and the only delta is the added `ci-principles` entry — nothing was dropped. Long lines are pre-existing in this file (main already carries 329-char lines) and no markdown line-length lint is configured.
- **Belongs inline per the file's own header rule**: the subsection is a durable *operating* constraint (never cancel verification of a distinct commit; path-filtering is an optimisation, not a correctness tool; one mechanism per behaviour) plus a one-line pointer, while the rules + incident evidence live in `defaults/docs/ci-principles.md` with a `.loom/docs/` symlink — the same shape as `build-gate.md` / `troubleshooting.md`. This is exactly what the header prescribes ("new subsystem detail goes to `.loom/docs/` with a one-line pointer here").
- `check-dangling-links.sh` OK (196 files), `check-agents-md-sync.sh` OK, `scripts/version.sh --check` → all six version-bearing files at 0.19.56.
- Every issue/PR the new doc cites (#6683, #7745, #7761, #7779, #7789, #7791, #7752) resolves to the incident it is described as.
- `check-review-feedback.sh --head-sha 7ea0e87e` → `CLEAR` (no formal reviews, no inline threads).
- Scope decision on `defaults/.loom/CLAUDE.md` / `AGENTS.md`: correct. `AGENTS.md` is generated from `defaults/.loom/CLAUDE.md` (the consumer template), not from the root `CLAUDE.md`, so the issue's "and therefore AGENTS.md" premise did not hold; the operator directive ("record it so every agent working here inherits it") is satisfied by the root file.

### CI: one failing check, assessed and not attributable to this PR

`Shell Test Suites (hermetic)` fails with `Failed suites: test-guard-destructive-core / -masking / -matching-precision / -readonly-fastpath` (229 passed, 4 failed). That is **byte-for-byte the same failure as `main`'s own run on this PR's base commit** (run 35048461474 on `42119258`, same four suites, same 229/4 summary), root-caused in #7808: #7799 flipped `guards.sqlDdl` in the committed `.loom/config.json` and the guard suites read that config. This PR touches no hook, test, config, or script file — its diff is two docs, one symlink, and the version bump — so it cannot fix or cause this. The check is not a required status check (the `main` ruleset has no `required_status_checks` rule; `mergeStateStatus` is `UNSTABLE`, not `BLOCKED`), and judge.md's table routes `UNSTABLE` to "assess if acceptable". Assessed acceptable: every other check passes, and sending a Doctor at this would be precisely the misattribution the document being merged warns about. Champion may still choose to hold the merge until #7808 lands `main` green; that is a merge-time call, not a review defect.

### Non-blocking notes (no action required for approval)

1. `ci-principles.md` says "Measured before the fix … (#7779, PR #7803)" and lists #7803 under Related as "the cancellation fix". #7803 is still **open** at review time; the phrasing reads as if it has landed. Harmless once it merges, but if it stalls the wording is ahead of reality.
2. The `Cancel Superseded Runs (same ref)` job still ran on this PR's own CI — expected until #7803 merges; noted only so nobody reads rule 2 as already enforced.
3. Rule 6 ("a check that cannot run must not look like a check that passed") is an addition beyond the issue's five draft rules; it is well-grounded (#7745, #7761) and the PR body calls it out, so I consider it in scope.

Labels: removing `loom:review-requested` / `loom:reviewing`, adding `loom:pr`.

rjwalters:
Judge: rebased this branch onto current `main` and re-derived the VERSION bump (0.19.56 -> 0.19.57) via `scripts/version.sh bump patch`. Reason: main advanced to 0.19.56 (#7467) after this PR bumped to the same value, so the `defaults/` template change would have merged with a net-zero VERSION increment -- the #7302 collision that `merge-pr.sh` hard-blocks. New head `021d9270`; continuing evaluation on it.

rjwalters:
## Judge verdict: APPROVED (`loom:pr`) — re-review of the post-approval push

Reviewed head `021d927066ef70201951a17e9ca842a003391496`. The earlier approval covered `7ea0e87e`; the author then pushed `8ab3bae` ("mirror the CI principle to consumers via the template and AGENTS.md"), which the daemon correctly demoted as a stale verdict (03:09Z). This verdict covers that delta plus a sanity pass on the whole diff.

### The unreviewed delta (`7ea0e87e..8ab3bae`)

Two files, 20 insertions, no deletions:
- `defaults/.loom/CLAUDE.md` (+11): a "CI is dumb and reliable, on purpose" subsection under Development Workflow, wrapped in its own `<!-- agents-md:include -->` range so `generate-agents-md.sh` picks it up. Consumer-neutral wording ("every default-branch commit"), pointer to `.loom/docs/ci-principles.md` which resolves in consumer installs because the doc ships from `defaults/docs/`.
- `defaults/.loom/AGENTS.md` (+9): the regenerated artifact.

The scope correction is right and the commit message's reasoning holds: a consumer's root `CLAUDE.md` is a `LOOM_ROOT_POINTER` stub, so anything meant to reach consumer agents has to be in the template (the #5207 / PR #5212 precedent). The previous verdict's "AGENTS.md is generated from the template, so the root file suffices" note was correct as far as it went but missed that the operator directive should reach consumers too; this commit closes that gap.

### What I verified on the PR tree and on a test-merge with current `main`

- `check-agents-md-sync.sh` OK, `check-claude-md-budget.sh` OK (320/320), `check-dangling-links.sh` OK (196 files), `check-docs-defaults-parity.sh` OK — all four on the PR tree **and** on the branch merged into current `main`.
- `check-review-feedback.sh --number 7805 --head-sha 021d9270` → `CLEAR` (0 reviews, 0 unresolved threads).
- CI on `021d9270`: 25 pass, 4 skipped, 1 fail. The one failure is `Shell Test Suites (hermetic)` with exactly `test-guard-destructive-core / -masking / -matching-precision / -readonly-fastpath` (229/4) — byte-for-byte `main`'s own #7808 breakage from #7799's config flip. This PR touches no hook, test, config, or script. Non-blocking, same assessment as the prior verdict.

### Judge-applied fix: VERSION collision (#7302)

The PR's 0.19.55 → 0.19.56 bump was overtaken: #7467 landed `main` at 0.19.56 after this PR was opened, so the merge was clean but net-zero on VERSION — `check-defaults-version-bump.sh --base origin/main` failed locally and `merge-pr.sh`'s #7302 guard would have hard-blocked the merge. This is the #7684 "bump goes stale before review lands" race the Judge role resolves directly, so I rebased onto current `main`, re-derived the bump with `scripts/version.sh bump patch` (0.19.57, `version.sh check` clean across all bearing files incl. `Cargo.lock` and `mcp-loom/package-lock.json`), and force-pushed with a lease pinned to `8ab3bae` (fork has `maintainerCanModify`). CI's `defaults/ Changes Require a VERSION Bump` job is green on the new head. Docs content is unchanged from `8ab3bae`.

### Non-blocking notes

1. The `8ab3bae` commit message says "the incidents stay in this repo's doc; only the rules travel" — but `defaults/docs/ci-principles.md` ships to consumers too (it has to, for the template's link to resolve). Harmless; the prior verdict's note about #7803 still being open at review time also still applies.
2. Champion may hold the merge until #7808 lands `main` green; that's a merge-time call, not a review defect.

Labels: removing `loom:review-requested` / `loom:reviewing`, adding `loom:pr`.

Issue #7783, opened 2026-09-16
post-verdict.sh: ${_gate_all_ids[*]} is an unbound variable on macOS bash 3.2 — blocks every Judge verdict on the clean path


turian:
## Problem

`post-verdict.sh` aborts on macOS stock bash (3.2) **whenever there is nothing
wrong with the PR**, which is the common case. The Judge role mandates this
script for every verdict ("never a bare `gh pr comment` (#6382)"), so on a
macOS host **no Judge can render any verdict at all**.

Hit live while approving #7768:

```
./.loom/scripts/post-verdict.sh: line 259: _gate_all_ids[*]: unbound variable
```

## Root cause

`.loom/scripts/post-verdict.sh:259` (and its canonical twin
`defaults/scripts/post-verdict.sh:259`, byte-identical):

```bash
read -ra _gate_all_ids <<< "$GATE_BLOCKING_IDS $GATE_INLINE_BLOCKING_IDS"
GATE_ALL_BLOCKING_IDS="${_gate_all_ids[*]}"
```

Under `set -u`, bash **< 4.4** treats `${arr[*]}` / `${arr[@]}` on an *empty*
array as an unbound variable. bash 4.4 changed this; macOS ships 3.2.57 and
will not ship a newer one.

Both id strings are empty exactly when no formal review and no inline thread is
blocking — so `read -ra` yields an empty array and the expansion explodes. The
failure is inverted from the usual: **the script works when the PR has problems
and crashes when it is clean.**

Minimal repro on macOS:

```console
$ /bin/bash -c 'set -u; read -ra a <<< " "; echo "count=${#a[@]}"; echo "${a[*]}"'
count=0
/bin/bash: a[*]: unbound variable
```

Note `${#a[@]}` is safe on 3.2 — only the value expansion is not, which is why
this survived review: the line above it uses the length form.

## Fix

```diff
-  GATE_ALL_BLOCKING_IDS="${_gate_all_ids[*]}"
+  GATE_ALL_BLOCKING_IDS="${_gate_all_ids[*]:-}"
```

Applied to **both** `defaults/scripts/post-verdict.sh` and
`.loom/scripts/post-verdict.sh` (the parity check requires they stay
byte-identical). Verified working:

```console
$ /bin/bash -c 'set -u; read -ra a <<< " "; echo "[${a[*]:-}]"'
[]
```

## Same class, worth auditing in the same pass

[...]

PR #7806, opened 2026-09-16
ci: validate markdown anchor fragments with lychee, fix 15 broken anchors

### LINKED ISSUE #7801: ci: markdown anchor fragments are never validated — 15 broken anchors on main today
## Problem

`scripts/check-dangling-links.sh` validates that a relative markdown link's
**path** resolves, then explicitly discards the fragment:

```bash
# Split off any #anchor fragment before path checks.
path_part="${target%%#*}"
```

and skips bare `#anchor` links entirely (`"#"*) continue`). So
`docs.md#section-that-was-renamed` passes CI today, as does `[x](#gone)`.

**440 anchor links across 263 tracked `.md` files are unvalidated. 15 of them
(3.4%) are already broken on `main`.**

## Verified examples

Three spot-checked by hand, all genuine:

| link | reality |
|---|---|
| `docs/guides/getting-started.md:12` → `#option-3-interactive-install-script` | file has "Option 1: Download Binary" and "Option 2: Build from Source". There is no Option 3. |
| `docs/guides/common-tasks.md:201` → `troubleshooting.md#initialization-issues` | target file exists, that heading does not |
| `.loom/docs/daemon-reference.md:238` → `#token-capacity-backpressure-3902` | heading is now `### Token-capacity backpressure (#3902; token axis retired from the cap by #5270)` — text was appended, anchor silently broke |

The third is the characteristic failure: **editing a heading breaks every anchor
pointing at it, with no signal.** That is exactly what a pre-merge gate is for.

## Full list of the 15

```
.loom/docs/daemon-reference.md         L15   #the-loomrole-env-var-contract-6507
.loom/docs/daemon-reference.md         L26   #gate-verdicts-verifiedred-vs-unevaluated-3974
.loom/docs/daemon-reference.md         L238  #token-capacity-backpressure-3902
.loom/docs/daemon-reference.md         L6163 runtime-adapters.md#model-runtime-mismatch-refusal-5028
defaults/docs/daemon-reference.md      (same four — byte-identical copy)
defaults/docs/guardrail-parity-codex.md L32  #managed-pretooluse-hook-bridge-issue-4495
defaults/docs/guardrail-parity-codex.md L37  #codexhome-profile-layout-refresh-and-security-posture
defaults/docs/machine-dispatcher.md    L34   #machine-mode-loommachinecheckout-hand-off-phase-3b-4229
defaults/docs/telemetry-schema.md      L20   #ingest-response-the-bound-hostid-echo
docs/guides/common-tasks.md            L201  troubleshooting.md#initialization-issues
docs/guides/getting-started.md         L12   #option-3-interactive-install-script
docs/guides/quickstart-tutorial.md     L463  ../../CLAUDE.md#4-scheduled-support-roles-opt-in
```

Eight of the fifteen are the `.loom/` + `defaults/` duplicated pair, so there are
**11 distinct defects**; both copies must be fixed to keep the parity check green.

## Implementation note — the slug rule that will cause false positives

GitHub's anchor slug lowercases, strips punctuation, and replaces **each space
with a hyphen — runs are NOT collapsed**. A naive `s/\s\+/-/g` produces wrong
slugs for any heading containing an em-dash or double space, and reports a flood
of false failures.

Concretely, `## Fleet — operator-triggered multi-host worker fanout (fleet, #4340)`:

```
correct : fleet--operator-triggered-multi-host-worker-fanout-fleet-4340
naive   : fleet-operator-triggered-multi-host-worker-fanout-fleet-4340
```

I hit this while measuring — the naive version reported **60** broken anchors
where only **15** are real. Also required: strip backticks and `[text](link)`
wrappers from heading text before slugging, honour `<a name="…">` / `id="…">`
anchors, skip fenced code blocks, and handle duplicate headings (`-1`, `-2`
suffixes).

## Scope

1. Extend `scripts/check-dangling-links.sh` to validate fragments — same-file
   (`#x`) and cross-file (`other.md#x`).
2. Extend its `--self-test` fixture to cover a broken anchor, a valid anchor, a
   duplicate-heading anchor, and an em-dash heading (the false-positive trap).
3. **Fix the 15 existing breakages**, keeping `.loom/` and `defaults/` copies
   byte-identical.

### Constraint: no new CI job

Add this as **steps inside the existing `Dangling Link Check` job**, not a new
job. That job runs in 8–9s against a ~351s critical path, so extra work there is
free in wall-clock terms. A new job would cost a full billable minute (GitHub
rounds every job up), ~2s of runner overhead, and would widen the 25-job fan-out
that is already the dominant CI cost — sub-minute jobs currently make billable
minutes ~1.6x raw compute.

### Explicitly out of scope

**External URL checking.** Network-dependent, rate-limited and flaky; it would
convert a deterministic 9-second job into a source of random red builds. Not
worth it at any wall-clock price.

Also out of scope, and a genuinely different project: validating that **commands
named in prose actually exist**. That is what actually broke twice recently
(#7712 nonexistent commands in onboarding guides, #7729 `gh label sync` in a
role prompt) — no link checker catches it, and it needs a command whitelist or
`--help` probing.

## Acceptance criteria

- [ ] `check-dangling-links.sh` fails on a broken same-file and cross-file anchor.
- [ ] Self-test covers duplicate headings and an em-dash heading; no false
      positives across the current 440 anchors.
- [ ] All 15 existing broken anchors fixed; `.loom/` and `defaults/` copies stay
      byte-identical.
- [ ] No new job in `ci.yml`; `Dangling Link Check` stays under ~20s.
- [ ] Hooks/Scripts Defaults Parity Check passes.

---

### Implementation approach superseded — see PR #7806

The scope below supersedes the Original Issue above. The author's own
follow-up comment on this issue explains why: the Original Issue prescribed a
hand-rolled awk implementation inside `check-dangling-links.sh`, and attempting
that produced three real bugs — `close` is a reserved word in awk (parse
failure, every anchor reported broken), BSD awk threw a `towc: multibyte
conversion failure` on UTF-8 headings, and a naive slug rule (not collapsing
consecutive space-to-hyphen substitutions) reported **60** broken anchors where
only **15** are real — the exact false-positive trap the Original Issue itself
warned about.

**PR #7806** (open as of 2026-09-15/16, carries `loom:changes-requested` +
`loom:ci-failure`, and `loom:treating` — Doctor is currently addressing Judge's
review feedback) implements the corrected approach instead: a new
`scripts/check-doc-anchors.sh` wraps `lychee` (pinned `v0.24.2`,
checksum-verified, run `--offline`) as a thin filter that acts only on
lychee's *fragment* verdicts. `check-dangling-links.sh` keeps owning path
checks and the `defaults/**` destination-aware mapping, which lychee cannot
model on its own (links under `defaults/**` are written against their install
destination, not their in-repo location). Two steps are added to the existing
`Dangling Link Check` job — no new job, matching the Original Issue's
constraint.

**If a Builder ends up finishing this issue independently of PR #7806
merging, follow #7806's lychee-based design — do not re-implement the
hand-rolled awk approach described in the Original Issue above.**

### Acceptance criteria (restated — supersedes the Original Issue's list)

- [ ] `scripts/check-doc-anchors.sh` (wrapping lychee, `--offline`, pinned +
      checksum-verified) validates same-file (`#x`) and cross-file
      (`other.md#x`) anchor fragments; `check-dangling-links.sh` keeps owning
      path checks only.
- [ ] All 15 existing broken anchors fixed (11 distinct defects — 4 of the 15
      are the `.loom/docs/daemon-reference.md` + `defaults/docs/daemon-reference.md`
      byte-identical pair); both copies stay byte-identical.
- [ ] No new job in `ci.yml`; two steps added to the existing `Dangling Link
      Check` job, which stays well under the ~350s CI critical path (baseline
      ~9s).
- [ ] Hooks/Scripts Defaults Parity Check passes.
- [ ] **Known gap, flagged in the issue's own follow-up comment, unresolved as
      of PR #7806**: add a self-test fixture for the wrapper's *filtering*
      logic specifically — assert a genuinely broken anchor fails while a
      `defaults/**` false-positive **path** verdict from lychee is not
      misclassified as a fragment failure. Close this out before merge or as
      an immediate fast-follow; it is not covered by lychee's own test suite
      because the filtering is this repo's own code.

### Related Open Work

`check-duplicate.sh --issue 7801` surfaced #7783 (macOS bash 3.2
`${_gate_all_ids[*]}` unbound-variable bug in `post-verdict.sh`) as
cross-referencing this issue. **Absorbed** — PR #7806's description states it
was verified under `/bin/bash` 3.2 (macOS stock) and that a bash-4-only
`mapfile` call was replaced with a `read -d ''` loop specifically because of
the #7783 lesson. No further action needed on this issue for that concern.

### Affected Files

- `scripts/check-doc-anchors.sh` — new file (not present on `origin/main` as
  of this curation pass); thin lychee wrapper filtering to fragment-only
  verdicts — introduced by PR #7806
- `scripts/check-dangling-links.sh` — unchanged in scope; keeps owning path
  checks and the `defaults/**` destination-aware mapping
- `.github/workflows/ci.yml` — two new steps in the existing `Dangling Link
  Check` job; job count stays unchanged
- `.loom/docs/daemon-reference.md` / `defaults/docs/daemon-reference.md` — fix
  4 broken anchors (symlinked byte-identical pair)
- `defaults/docs/guardrail-parity-codex.md` — fix 2 broken anchors
- `defaults/docs/machine-dispatcher.md` — fix 1 broken anchor
- `defaults/docs/telemetry-schema.md` — fix 1 broken anchor
- `docs/guides/common-tasks.md` — fix 1 broken anchor
- `docs/guides/getting-started.md` — fix 1 broken anchor
- `docs/guides/quickstart-tutorial.md` — fix 1 broken anchor

### Test Plan

- [ ] Manual verification: run `scripts/check-doc-anchors.sh` against `main`
      after the fixes land; expect a clean "all #anchors resolve" verdict
      across the ~263 tracked markdown files.
- [ ] Automated tests: the wrapper's self-test fixture covering the filtering
      gap above (broken anchor fails; a `defaults/**` path false-positive from
      lychee does not).
- [ ] Edge cases: duplicate headings, em-dash headings (space-run collapsing
      in the slug rule), and underscore-preserving slugs (GitHub keeps `_` in
      slugs — most of the 15 real breakages were dropped-underscore slugs,
      e.g. `LOOM_ROLE` → `loom_role`, not `loomrole`).

**Note for whoever picks this up**: this issue already has an open
implementing PR (**#7806**), currently being addressed by Doctor after Judge
requested changes / CI failed. Check #7806's status before starting fresh
implementation work, to avoid duplicating effort already in flight. `loom:issue`
is retained (not a material scope change — the underlying goal and most
acceptance criteria are unchanged; only the implementation mechanism changed
from hand-rolled awk to a lychee wrapper, which the issue's own follow-up
comment already documents).

turian:
Validates markdown `#anchor` fragments in CI and fixes the 15 that were already
broken on `main`. Closes #7801.

## The gap

`check-dangling-links.sh` checked only a link's **path**, then threw the
fragment away:

```bash
# Split off any #anchor fragment before path checks.
path_part="${target%%#*}"
```

Bare `#anchor` links were skipped outright. So **440 anchor links across 263
tracked markdown files were unvalidated**, and `doc.md#section` stayed green
after the heading was renamed.

## Why a tool instead of more bash

I started hand-rolling this in awk and it produced **three bugs in three
attempts**:

1. `close` is a reserved word in awk — the program failed to parse, the index
   came back empty, and *every* anchor reported broken.
2. BSD awk on macOS threw `towc: multibyte conversion failure` on UTF-8 headings.
3. GitHub's slug rule replaces **each** space with `-` without collapsing runs,
   so `## Fleet — operator` is `fleet--operator`. A naive `s/\s\+/-/` reported
   **60** broken anchors where **15** were real.

None of that is repo-specific — it is a solved problem. `lychee` implements the
rule correctly and checks ~1500 links in **~70ms**. Its verdict matched an
independent Python implementation exactly: **15 and 15**.

## What this adds

**`scripts/check-doc-anchors.sh`** — a thin filter that takes only lychee's
*fragment* verdicts and leaves every *path* verdict to `check-dangling-links.sh`.
That split matters: links under `defaults/**` are written against their **install
destination**, not their in-repo location, so lychee's naive path resolution
reports false positives there. The existing checker already models that
(destination-aware mode) and keeps owning it.

**Two steps in the existing `Dangling Link Check` job — not a new job.** That job
runs ~9s against a ~350s critical path, so this is free in wall clock. A 27th job
would cost a full billable minute (GitHub rounds every job up) and widen the
fan-out that already makes billable minutes ~1.6x raw compute. lychee is pinned
to `v0.24.2` and checksum-verified against the published `.sha256`.

**`--offline`.** External URL checking is deliberately out of scope — it is
rate-limited and flaky, and would turn a deterministic sub-second check into a
source of random red builds.

## The 15 fixes

Mostly **dropped underscores** — GitHub keeps `_` in slugs, so `LOOM_ROLE` is
`loom_role`, not `loomrole`:

| file | was | now |
|---|---|---|
| `daemon-reference.md` | `#the-loomrole-env-var-contract-6507` | `#the-loom_role-…` |
| `daemon-reference.md` | `#gate-verdicts-verifiedred-…` | `#gate-verdicts-verified_red-…` |
| `daemon-reference.md` | `#token-capacity-backpressure-3902` | `…-3902-token-axis-retired-from-the-cap-by-5270` |
| `daemon-reference.md` | `runtime-adapters.md#model-runtime-…` | `#modelruntime-…` |
| `guardrail-parity-codex.md` | `#codexhome-…`, `#managed-pretooluse-…` | `#codex_home-…`, `#managed-pre_tool_use-…` |
| `machine-dispatcher.md` | `#machine-mode-loommachinecheckout-…` | `#machine-mode-loom_machine_checkout-…` |
| `telemetry-schema.md` | `#ingest-response-the-bound-hostid-echo` | `…-host_id-echo` |
| `common-tasks.md` | `troubleshooting.md#initialization-issues` | `#installation-issues` |
| `getting-started.md` | `#option-3-interactive-install-script` | `…-script-recommended` |
| `quickstart-tutorial.md` | `CLAUDE.md#4-scheduled-support-roles-opt-in` | `#4-scheduled-support-roles` |

The third one is the characteristic failure: text was appended to a heading and
every anchor pointing at it broke silently.

`.loom/docs/*.md` are symlinks into `defaults/docs/`, so editing the `defaults/`
copy fixes both and keeps the parity check green.

## Verification

- `check-doc-anchors.sh` now reports **OK — checked 263 tracked markdown
  file(s), all #anchors resolve.**
- Verified under **`/bin/bash` 3.2** (macOS stock). `mapfile` was the first thing
  I reached for and it is bash 4+; replaced with a `read -d ''` loop. See #7783
  for what assuming bash 4 costs on a mandatory path.
- `actionlint` clean; job count unchanged (26 before, 26 after) — steps only.


rjwalters:

Evaluating per dispatch. Cloned `f12a86f8d` into a scratch checkout, ran the new script against a real `lychee 0.24.2` (macOS build), and independently re-derived every corrected anchor with a separate GitHub-slug implementation.

### What is right

- **All 11 distinct anchor corrections resolve.** Each target heading slugs to exactly the new fragment (underscores kept, appended heading text included, `Option 3 ... (Recommended)`, `Installation Issues`, `4. Scheduled Support Roles`). `check-doc-anchors.sh` reports OK on the branch and, when I deliberately broke one anchor, it failed with a precise `file#anchor (at line:col)` report for both the `defaults/` and `.loom/` paths. `--include-fragments=anchor-only` is a valid flag on the pinned release.
- **Pinned checksum is genuine**: `1f4e0ef7...951c9a` matches the published `lychee-x86_64-unknown-linux-gnu.tar.gz.sha256`, and the job runs on `ubuntu-latest` (x86_64), so the asset matches the runner.
- **No overlap with `check-dangling-links.sh`** — it is the complementary half (path vs fragment), and taking only lychee's *fragment* verdicts correctly leaves the `defaults/**` install-destination path semantics with the existing checker. Riding in the existing job rather than adding a 27th is the right call; the check itself is ~0.1s.
- bash 3.2-clean (`read -d ''` loop, no `mapfile`); `--offline` keeps it deterministic.

### Blocking — CI is red because of this PR (2 of the 3 failures)

**1. `Doc Table-of-Contents Freshness` fails, and it is a real conflict, not a nuisance.** Eight of the eleven edits are inside `<!-- toc:begin ... -->` blocks that `scripts/check-doc-tocs.sh --fix` regenerates. That generator's `slugify()` strips underscores:

```bash
# scripts/check-doc-tocs.sh:61
| sed -e 's/[^a-z0-9 -]//g' \
```

so it emits `#the-loomrole-...` — the very anchors this PR corrects. As submitted, the two checks contradict each other permanently: the TOC check demands the broken slug, the anchor check demands the right one. The generator is the bug. I verified the fix is one character:

```bash
| sed -e 's/[^a-z0-9_ -]//g' \
```

With that change, `check-doc-tocs.sh --fix` is a no-op against your hand-edited TOCs (`0 of 12 doc(s) updated`), `check-doc-tocs.sh` passes, and `check-doc-anchors.sh` stays green. Please include it (and drop a one-line note in the generator's header that `_` is preserved because GitHub keeps it).

**2. `defaults/ Changes Require a VERSION Bump` fails.** The PR touches `defaults/docs/*` with neither a `VERSION` bump nor the opt-out marker. These are link-target corrections with no installed-behavior change, so add this exact line to the PR body (or a commit message):

```
<!-- loom:no-surface-change -->
```

Prefer the marker over a `VERSION` bump here — a version-bump diff on a docs PR also trips Champion's merge-risk criterion and would park it.

**3. The gate can pass silently when lychee itself fails.** `lychee ... || true` followed by `grep -q "Cannot find fragment"` means any run where lychee does not execute the check (bad flag after a future bump, crash, unreadable input) prints `OK ... all #anchors resolve` and exits 0. I confirmed this with a stub `lychee` that exits 2 with a usage error — the script reported OK. Capture the exit status instead of discarding it: lychee exits 0 (clean) or 2 (link errors); anything else should be a hard failure, and it is worth asserting the output contains lychee's summary line before trusting an "OK". A gate whose failure mode is green defeats its purpose, and this is a few lines.

### Not attributable to this PR

`Shell Test Suites (hermetic)` fails on the same four `tests/hooks/test-guard-destructive-*.sh` suites (SQL-DDL denial cases) on `main` at `421192583` — this PR's merge base — after #7799 disabled the `guards.sqlDdl` category. That is a pre-existing red and needs its own fix; I am not counting it against this PR, but approval will require a green run once the two PR-caused failures above are fixed, so a rebase onto a main that has the guard-test fix (or an explicit note if it is still red) will be needed.

### Optional

- A `--self-test` (as `check-dangling-links.sh` has) with a fixture covering a broken anchor, an em-dash heading, and a duplicate heading would let the gate prove itself without a manual break-and-run like the one I did.
- The release download is a per-run network fetch from github.com; it is the same host CI already talks to, so I do not consider it a flake risk, but an `actions/cache` keyed on `VER` would make it fully hermetic if that ever bites.

Nice diagnosis in the description — the awk war stories are exactly the kind of context that keeps the next person from re-deriving the slug rule by hand.

Issue #7802, opened 2026-09-16
Exhaustive bash 3.2 scan: 2 silently-wrong constructs + 338 unguarded array expansions — and why the Rust port changes the fix order


turian:
An exhaustive scan over all **472 tracked `.sh` files** (excluding `.loom/` resync mirrors), with comments **and heredoc bodies** excluded, finds two distinct bash-3.2 incompatibility classes with very different tractability.

This is the "close the class, not the instance" follow-up that #7789 argues for, applied to the bash-3.2 family. #7798 cleared the constructs I had grepped for; this found the ones I had not.

## Class 1 — hard bash 4+ constructs: **2 remaining**

Both are **worse than a crash — they are silently wrong on 3.2.**

| site | construct | what bash 3.2 does |
|---|---|---|
| `defaults/scripts/random-file.sh:232` | `shopt -s globstar` (4.0) | fails softly (`2>/dev/null \|\| true`), then `**` in the pattern degrades to single-level `*`, so recursive matching **silently returns the wrong file set** |
| `defaults/scripts/tests/run-ci-suites.sh:382` | `wait -n` (4.3) | the script is `set -uo pipefail` **without** `-e`, so `wait: -n: invalid option` prints, `running` decrements anyway, and the parallelism limiter breaks — jobs spawn **unbounded** |

Neither aborts. A developer on a Mac gets wrong results, not an error.

Also scanned for and found clean: `declare -A`, `local -A`, `mapfile`, `readarray`, `${x,,}`, `${x^^}`, `declare -n`, `coproc`, `|&`, `&>>`, `[[ -v ]]`, `compgen -A variable`, `exec {fd}`, `printf '%(fmt)T'`.

## Class 2 — unguarded array expansion under `set -u`: **338 candidates**

`"${arr[@]}"` / `"${arr[*]}"` where bash **< 4.4** treats an **empty** array's expansion as an unbound variable and aborts. Fixed upstream in 4.4; macOS ships 3.2.57 and will not ship newer.

| file | candidates |
|---|---|
| `scripts/uninstall-loom.sh` | 24 |
| `defaults/scripts/tests/test-loom-daemon-watchdog.sh` | 16 |
| `scripts/install-loom.sh` | 15 |
| `defaults/scripts/spawn-claude.sh` | 13 |
| `defaults/scripts/cli/loom-daemon-update.sh` | 10 |
| `defaults/scripts/claude-wrapper.sh` | 9 |
| …plus ~250 more across the tree | |

[...]

Issue #7777, opened 2026-09-15
ADR-0018 (draft): Rust owns behavior; shell exists only to reach it


turian:
Draft of **ADR-0018**, filed as an issue for discussion before it lands as `docs/adr/0018-rust-owns-behavior-shell-reaches-it.md`. Section structure follows ADR-0016.

---

# ADR-0018: Rust Owns Behavior; Shell Exists Only to Reach It

## Status

**Proposed.** No implementation ships with this ADR. Two follow-on issues carry the first slices (execution contract, compatibility contract); a third carries the first complete migration. #7762 narrows to the default rule once this is accepted.

### The directory does not look the way people assume

`defaults/scripts/` is **105 files, 21,988 code lines, median 122 lines, max 1,853**. 76 of 105 are under 200 lines. It is not a pile of mega-scripts. The genuinely large shell surfaces are elsewhere: `defaults/scripts/tests/` (206 files, ~91k lines) and `defaults/hooks/` (8 files, 13.6k).

### A migration is already ~45% done and undocumented

| state | count |
|---|---|
| thin stubs over `loom-daemon` (<40 code lines) | 15 |
| larger, but already call `loom-daemon` | 32 |
| pure shell, no `loom-daemon` at all | 58 |

Six are literally four lines of `exec`. There is **no README and no doc** stating what the directory is organized by, so three migration states coexist with nothing marking which is which. That — not size, and not language alone — is why the directory reads as a rat's nest.

### The names are a public interface

**58 distinct scripts are invoked from role prompts**; 11 more from CI workflows. The script *names* are the API agents call. Moving logic behind a name is cheap; renaming is not.

### What shell is actually costing

- **Lexing text in shell**: 52 of 99 commits to `guard-destructive-generic.sh` in 12 months are code-vs-quoted-data confusion, with reproduced silent ALLOWs (#7755, #7760).
- **Portability**: a recurring bash 3.2/4 class breaking macOS dev machines (#7717, #7728, #7730, #7749, #7751).
- **Retry/backoff policy in bash**: `claude-wrapper.sh` is 2,932 lines described as *"Resilient Claude CLI wrapper with retry logic."*

## Decision

**Rust owns behavior. Shell exists only where it is necessary to reach that behavior.**

### The boundary is responsibility, not line count

| | |
|---|---|
| **Shell** | Find the executable, report if it is missing, replace itself with that executable (`exec`). |
[...]

Issue #7758, opened 2026-09-15
Re-derive which bootstrap scripts must stay shell — fold the inventory into .loom/docs/file-size-policy.md's tier table (docs-only)


turian:
**Docs-only issue (operator ruling 2026-09-16).** The inventory below — developed in this issue's comment thread and folded into the body here — is the evidence. The deliverable is updating the shell tier table in `.loom/docs/file-size-policy.md` (the "Bootstrap scripts are deliberately NOT exempt" paragraph, currently lines 97–100, and the `## Shell: port to Rust, tiered` table, currently lines 161–173) to match it. No port is done here.

## Origin

#7711's shell tier table asserts that four scripts can "**never** be ported" to Rust because they run before, or manage, the `loom-daemon` binary: `cli/loom-daemon-update.sh` (1,770), `cli/loom-daemon-start.sh` (1,185), `scripts/install-loom.sh` (1,488), `install.sh` (1,367). That claim was inherited, not derived, and the derivation below shows it is wrong in both directions: two of the four are port-candidates, and two scripts it does not name (`scripts/uninstall-loom.sh`, `defaults/scripts/lib/locate-daemon-bin.sh`) are the ones that genuinely must stay shell.

### Self-correction: `auto_update.rs` does NOT duplicate `loom-daemon-update.sh`

This issue was filed claiming `auto_update.rs` "already does the same job" as the update script. **That is wrong**, and the source says so directly. `auto_update.rs` owns **tick scheduling and update decisions** and **delegates** release resolution, fetching, verification and provisioning *to the script*:

```
auto_update.rs:12   "it *decides* and *sequences* the roll, reusing `loom-daemon-update.sh` for ..."
auto_update.rs:79   "Each tick first asks `loom-daemon-update.sh --resolve-json` what the latest ..."
auto_update.rs:100  "through the read-only `--resolve-json` mode rather than reimplementing"
auto_update.rs:806  "and shells out to `loom-daemon-update.sh --no-restart` for the rebuild"
```

The script's own header states the division deliberately ("the resolution logic lives here once instead of being duplicated in Rust"). The two are complementary by design. Consequences:

- `loom-daemon-update.sh` is a **migration candidate**, not redundant code. Porting it means retiring the `--resolve-json` / `--no-restart` contract consumed by `auto_update.rs` — an API change on both sides (this is #7810 Phase 5).
[...]

PR #7807, opened 2026-09-16
fix: three bash 3.2 correctness bugs, all of them silent (#7783, #7802 Class 1)

### LINKED ISSUE #7783: post-verdict.sh: ${_gate_all_ids[*]} is an unbound variable on macOS bash 3.2 — blocks every Judge verdict on the clean path

turian:
Closes #7783. Resolves the two Class 1 findings from #7802 — **not** the broader audit.

Three bash 3.2 correctness fixes, one commit each. All three were **silent** — none produced an error a reader would notice — so every fix is verified by asserting a **value**, not by observing a successful exit.

## 1. `post-verdict.sh` — blocked every clean Judge verdict (#7783)

`${_gate_all_ids[*]}` on an **empty** array is an unbound variable under `set -u` on bash < 4.4.

Both id strings are empty exactly when nothing is blocking — the CLEAR path — so the failure was **inverted**: the script worked when the PR had problems and aborted when it did not. The Judge role mandates this script for every verdict (#6382), so **no Judge could render a clean verdict on any macOS host**.

Reproduced before fixing:
```
./post-verdict.sh: line 259: _gate_all_ids[*]: unbound variable
```

Verified after, all three shapes:

| input | result |
|---|---|
| both empty | `[]` |
| populated | `[R1 R2 T9]` |
| one empty | `[T9]` |

## 2. `random-file.sh` — `**` silently matched one directory level

`shopt -s globstar` ran behind `2>/dev/null || true`. On 3.2 globstar does not exist, the shopt failed silently, and `**` then behaved as a single-level `*`.

Measured on a fixture with 4 matching files at depths 0–3:

```
bash 3.2 glob **    -> 1 file    (one.rs)
find (ground truth) -> 4 files   (one.rs three.rs top.rs two.rs)
```

Recursive patterns now go through `find`. **Asserted the resulting file set equals `find`'s exactly** — this failure produces nothing to check otherwise.

Also removed dead code: the `if`/`else` around the `**` test ran the identical `eval` on both branches.

## 3. `run-ci-suites.sh` — parallelism was unbounded on 3.2

`wait -n` is bash 4.3. It fails on stock macOS with `wait: -n: invalid option`, and the script is `set -uo pipefail` **without** `-e` — so it did not abort. `running` decremented anyway, the bound stopped bounding, and suites spawned without limit.

Now waits on the **oldest** pid rather than any pid. No new machinery, and the bound is preserved exactly. A slow oldest job can hold a slot marginally longer than `wait -n` would — a scheduling nuance, not a correctness one.

**Deliberately not a hand-rolled job poller.** My first attempt was 21 lines of pid tracking, `kill -0`, array reindexing and `sleep`. That is concurrency control — application logic, and the wrong side of ADR-0018's boundary (#7777). Adding bash machinery to work around a bash limitation is the opposite of the direction the repo is heading. This is 3 lines of scheduling instead, and the first version's own test hung, which was the tell.

Verified by asserting the value: 8 jobs at `PARALLELISM=3`, event log shows **max concurrent = 3** and all 8 started.

## Scope

Three correctness fixes only. Explicitly **not** in scope:

- The **338 Class 2 candidates** from #7802. Those need #7758's per-file verdict first — guarding an array in a script that becomes a Rust subcommand is wasted work, and nobody can currently tell which is which. `post-verdict.sh` is included here because it is a **reproduced production failure**, not because it is Class 2.
- #7794's `--help` restructuring, which may be subsumed by ADR-0018 entirely.
- #7800 (test-module extraction) stays separate.

## On #7802

This closes **both Class 1 findings**. Issue #7802 itself stays open — the 338 Class 2 candidates and the proposed `check-bash32-compat.sh` gate remain there.

## Verification summary

All three parse under `/bin/bash 3.2.57`. `check-file-size-budget` and `check-shell-syntax --dir defaults` pass.


rjwalters:
## Judge verdict: changes requested

Evaluated per the Judge role (claimed `loom:reviewing` 03:00:28Z, sole claim on the timeline). Reviewed head `5d357b18`. Every claim below was measured on this host against both `/bin/bash` 3.2.57 (stock macOS) and `/opt/homebrew/bin/bash` 5.3.20, from a scratch clone of the branch.

### What is correct (verified)

**Fix 1 — `post-verdict.sh` empty-array expansion.** Reproduced the bug and the fix with a 3-line harness (`set -u; read -ra _ids <<< "$1 $2"; echo "[${_ids[*]}]"`):

| input | bash 3.2 before | bash 3.2 after | bash 5.3 before/after |
|---|---|---|---|
| both empty | `_ids[*]: unbound variable` | `[]` | `[]` |
| `"R1 R2" "T9"` | `[R1 R2 T9]` | `[R1 R2 T9]` | `[R1 R2 T9]` |
| `"" "T9"` | `[T9]` | `[T9]` | `[T9]` |

Correct, and the `${arr[*]+${arr[*]}}` form is the standard 3.2-safe idiom.

**Fix 3 — `run-ci-suites.sh` `wait -n`.** Reproduced with an event-logged harness (8 jobs, 0.4s each):

| | bash 3.2 before | bash 3.2 after | bash 5.3 before / after |
|---|---|---|---|
| `PARALLELISM=3` | `wait: -n: invalid option` x6, max_concurrent=**8** | max_concurrent=**3** | 3 / 3 |
| `PARALLELISM=1` | max_concurrent=**8** | max_concurrent=**1** | 1 / 1 |

The bound is restored on 3.2 and unchanged on 5.3. I specifically checked the `PARALLELISM=1` edge (the `${_RUN_PIDS[@]:1}` slice of a 1-element array is empty) under `set -u` on 3.2: it does not trip nounset. Good.

**Fix 2 — `random-file.sh` globstar, for the documented shapes.** On a fixture with `.rs` files at depths 0-3, bash 3.2 `**/*.rs` went from 2 hits (`other/o.rs src/one.rs`) to the full `find` set of 7; `src/**/*.rs` from 2 to 5; `src/**` from 1 to 6. Every shape documented in the script header (`src/**/*.ts`, `**/*.test.ts`, `defaults/roles/*.md`) now matches `find` on both interpreters.

**Suites.** `test-post-verdict.sh` (30/30), `test-random-file.sh` (5/5), `test-run-ci-suites-{daemon-guard,failure-excerpt,serial-lane}.sh` (20/20, 24/24, 12/12), `test-review-feedback-reconciliation.sh` (61/61) all pass under both `/bin/bash` 3.2 and bash 5.3 on the PR tree. `check-review-feedback.sh` gate: `CLEAR` (0 reviews, 0 inline threads). Merge-tree against `origin/main` (42119258, which is also the PR base): clean.

### Blocking

**B1. Shellcheck CI job fails, and this PR introduced it.** `defaults/scripts/tests/run-ci-suites.sh:395:20: SC2206` on the new line

```bash
_RUN_PIDS=(${_RUN_PIDS[@]:1})
```

Reproduced locally with shellcheck 0.11.0 (`--severity=warning`, same as CI). Fix is to quote it: `_RUN_PIDS=("${_RUN_PIDS[@]:1}")`. I verified the quoted form on both interpreters under `set -uo pipefail`: a 1-element array slices to `len=0` (no stray empty element), a 2-element array to `len=1`. Behavior is identical to the unquoted form for integer pids; only the lint verdict changes.

While there, the loop's lead-in comment (line 366) still says "bounded to `$PARALLELISM` concurrent jobs via `wait -n`", which the diff just made false.

**B2. The `random-file.sh` `find` rewrite regresses bash 5 for `**` patterns that are not `[literal-root/]**/<leaf>`.** The comment says "`find` recurses identically on both interpreters", but `_gs_root="${pattern%%/\*\**}"` / `_gs_name="${pattern##*/}"` discard everything between the first `/**` and the leaf, and the root is passed to `find` quoted (unexpanded). Measured on the same fixture (bash 5.3, before = `origin/main` globstar, after = this PR):

| pattern | bash 5.3 before | bash 5.3 after |
|---|---|---|
| `src/**/test/*.rs` | `src/test/t.rs` | `src/a/b/c/four.rs src/a/b/three.rs src/a/two.rs src/one.rs src/test/t.rs` (over-matches; `test/` constraint dropped) |
| `src/*/**/*.rs` | `src/a/b/c/four.rs src/a/b/three.rs src/a/two.rs src/test/t.rs` | *(empty)* (`find "src/*"` is a literal path) |

Neither shape is in the script's usage examples, so this is bounded, but it is a silent wrong-result on the interpreter CI actually runs on, which is the exact failure class this PR is titled after. Two acceptable resolutions, either is fine with me:

- **Smallest:** keep the globstar path where it exists and use `find` only where it does not, so bash >= 4 behavior stays byte-identical to today:
  ```bash
  if shopt -s globstar 2>/dev/null; then
      eval "matches=($pattern)" 2>/dev/null || true
  else
      # bash 3.2: no globstar -- find fallback (as in this PR)
  fi
  ```
- **Or** keep `find` everywhere but reject (with a clear error) any `**` pattern that has a non-literal root or path components after `**`, so the unsupported shapes fail loudly rather than returning the wrong set.

### Not this PR's fault (for the record)

- `Shell Test Suites (hermetic)` is also red, but 4 of the 5 failing suites (`test-guard-destructive-{core,masking,matching-precision,readonly-fastpath}.sh`) fail identically on `main`'s own run at 42119258 (CI run 35048461474, "229 passed, 4 failed"), i.e. they came in with #7799's SQL-DDL category change. The 5th, `test-verify-proposal-refs.sh`, died with `printf: write error: Broken pipe` in the runner; it passes 24/24 locally on this tree and the PR does not touch that script. Once B1/B2 land, expect this job to stay red until `main` is fixed; that is a rebase concern for Champion, not a defect here.
- No required status checks exist on `main` (no ruleset, no classic protection), hence `UNSTABLE` rather than `BLOCKED`.

### Non-blocking suggestion

None of the three fixes has a regression test. I ran `main`'s `test-post-verdict.sh` and `test-random-file.sh` under `/bin/bash` 3.2 and both pass (30/30, 5/5) *with the bugs present*, so the existing suites do not exercise these paths. CI runs on ubuntu so a 3.2 assertion would not fire there anyway, but a small `bash -c 'set -u; ...'` value assertion in `test-post-verdict.sh`'s CLEAR path, and a depth-2+ fixture in `test-random-file.sh` for `**/*.x`, would at least fail on a macOS dev machine. Fine as a follow-up.

### Verification commands (abbreviated)

```
shellcheck --severity=warning --format=gcc defaults/scripts/{post-verdict,random-file,tests/run-ci-suites}.sh   # SC2206 at run-ci-suites.sh:395
gh run view 35048461474 --json jobs  # main@42119258: Shell Test Suites (hermetic) failure, same 4 guard suites
./.loom/scripts/check-review-feedback.sh --number 7807 --head-sha 5d357b18...   # REVIEW_FEEDBACK_STATE=CLEAR
```

Requesting changes for B1 and B2; B1 is a one-token fix, B2 is a small structural choice. Happy to re-review promptly once pushed.

turian:
Both failures are now accounted for. One was mine; one was not.

**Shellcheck — mine.** `_RUN_PIDS=(${_RUN_PIDS[@]:1})` was unquoted (SC2206). Fixed in `14aa4fe7`.

Quoting is safe here even though this script runs under `set -u` on bash 3.2 — which is the hazard this PR exists to fix, so I checked rather than assumed. An array *slice* of an empty or exhausted array expands to nothing, unlike a bare `"${arr[@]}"`, which trips the unbound-variable error on < 4.4. Verified on 3.2.57 for the 1-element and already-empty cases.

Re-asserted the contract rather than the exit status: 8 jobs at `PARALLELISM=3` still gives **max concurrent = 3, all 8 started**, on bash 3.2.

**Shell Test Suites — not mine; inherited from `main`.** `main` has been red since #7799, a config-only PR that set `guards.sqlDdl: false`. The guard suites invoked the guard with the live repo root as cwd, so they read *this repo's* config and 22 deny-assertions across 4 suites silently became allow-assertions.

```
4852c7d3 (#7798)  success
42119258 (#7799)  failure   <- config only
22345a06 (#7467)  failure
```

Fixed in **#7816** by closing the config-file channel the same way the harness already closes the environment channel (#4325). This PR should go green on Shell Test Suites once #7816 lands; it needs no change here.

For disclosure: I reviewed and merged #7799.

turian:
**Green: 27 pass, 0 fail.** Both earlier failures are resolved, and they had different causes.

**Shellcheck** was mine — fixed in the SC2206 commit above.

**Shell Test Suites** failed twice for two unrelated reasons:

1. *Inherited from `main`.* `main` was red from #7799 until #7809 landed. Rebased onto the fixed `main`.
2. *A flake, not a regression.* After the rebase one suite still failed — `test-sync-labels-repo-flag.sh`, on `legitimate NWO 'a/b' resolves to itself`. It **passed on re-run with no code change**.

I re-ran rather than guessing because the evidence pointed away from a real defect: the assertion uses `grep -qF`, and CI's own diagnostic printed the expected substring as present in the actual output —

```
Expected substring: 'Target repository: a/b (github)'
In: '<ESC>[0;34mℹ Target repository: a/b (github)<ESC>[0m
```

A fixed-string match cannot fail against a haystack that visibly contains it, which points at the captured output rather than at `sync-labels.sh`. The suite also passes locally 161/161 on both `main` and this branch, and nothing here touches `sync-labels.sh`. Filed separately — a suite that fails ~once per run for output-capture reasons is exactly the kind of gate that trains people to re-run instead of read.

## Self-review

I authored this. Mechanical verification only, and each fix asserts a **value** rather than a successful exit, because all three bugs were silent:

| fix | assertion |
|---|---|
| `post-verdict.sh` | `[]` / `[R1 R2 T9]` / `[T9]` for empty, populated, one-empty |
| `random-file.sh` | resulting file set equals `find`'s exactly (3.2 glob found 1 of 4) |
| `run-ci-suites.sh` | 8 jobs at `PARALLELISM=3` → max concurrent **3**, all 8 started, on 3.2.57 |

Rebased onto current `main` (the branch had gone `CONFLICTING` on nine version-bump files); I cherry-picked the three fix commits, dropped the stale `0.19.56` bump, and re-bumped to `0.19.58`. Re-verified all three fixes survived the rebase and that all three parse under `/bin/bash` 3.2.57.

